Issue #8160, opened 2017-10-24
[bug] Uncaught ReferenceError: global is not defined


amitport:
### Bug Report or Feature Request (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.

```
Angular CLI: 1.5.0-rc.3
Node: 8.7.0
OS: win32 x64
Angular: 4.4.6
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router, tsc-wrapped, upgrade

@angular/cdk: 2.0.0-beta.12
@angular/cli: 1.5.0-rc.3
@angular/flex-layout: 2.0.0-beta.9
@angular/material: 2.0.0-beta.12
@angular-devkit/build-optimizer: 0.0.30
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.34
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.0-rc.3
@schematics/angular: 0.0.48
typescript: 2.3.4
webpack: 3.8.1
```

### Repro steps.
create angular/cli project
add dependency and import `socket.io-client@2.0.4`
ng serve --aot
open browser

### The log given by the failure.

```
index.js:10 Uncaught ReferenceError: global is not defined
    at eval (index.js:10)
    at Object.../../../../has-binary2/index.js (vendor.bundle.js:2914)
    at __webpack_require__ (inline.bundle.js:55)
    at eval (index.js:8)
    at Object.../../../../socket.io-parser/index.js (vendor.bundle.js:9569)
    at __webpack_require__ (inline.bundle.js:55)
    at eval (index.js:7)
    at Object.../../../../socket.io-client/lib/index.js (vendor.bundle.js:9527)
    at __webpack_require__ (inline.bundle.js:55)
    at eval (io-service.ts:1)
```

dottodot:
I seeing the same issue as well with `ng-socket-io` but also with `import 'intl';` in `polyfills.ts`

amitport:
- I tried adding `(window as any).global = window;` to `pollyfils.ts`, but it didn't work. 
- reverting to 1.5.0-rc.2 worked

dottodot:
Strange I've reverted to `1.5.0-rc.2` and still get it, no idea what's going on.

amitport:
[...]

PR #8130, opened 2017-10-20
fix(@angular/cli): disable node global

filipesilva:
The node only `global` object had been left in because it used to cause a build size regression with Angular.

This doesn't seem to be the case anymore so it should be removed because it causes problematic interactions with some libraries.

Fix #5804
Supersedes #7931

Issue #5804, opened 2017-04-03
mapboxgl is incompatible with the CLI in a production build


rolandjitsu:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

It seems that using [Mapbox GL](https://github.com/mapbox/mapbox-gl-js) and Angular CLI when targeting production does not work.

Running an app that imports Mapbox with `import * as mapboxgl from 'mapboxgl'` via `ng serve` seems to work just fine. But if you run it via:
```bash
ng serve --open --sourcemap --target=production --output-hashing=all
```
An error is thrown from one of the workers created by Mapbox.

### Versions.

```text
@angular/cli: 1.0.0-rc.4
node: 7.4.0
os: darwin x64
@angular/common: 2.4.10
@angular/compiler: 2.4.10
@angular/core: 2.4.10
@angular/flex-layout: 2.0.0-rc.1
@angular/forms: 2.4.10
@angular/http: 2.4.10
@angular/material: 2.0.0-beta.2
@angular/platform-browser: 2.4.10
@angular/platform-browser-dynamic: 2.4.10
@angular/router: 3.4.10
@angular/cli: 1.0.0-rc.4
@angular/compiler-cli: 2.4.10
@angular/tsc-wrapped: 0.5.2
```

### Repro steps.

1. Scaffold an Angular CLI app
2. Add Mapbox GL dep `"mapbox-gl": "^0.34.0"`
3. Add the typings (`"@types/mapbox-gl": "^0.30.0"`)
4. `import * as mapboxgl from 'mapboxgl'` anywhere in the app (log `mapboxgl` just to see the output)
5. Create a map in one of the components `const map: mapboxgl.Map = new Map(Object.assign({container: element}));` (`element` is the `nativeElement` from `@ViewChild('map', {read: ElementRef}) mapRef: ElementRef`)

### The log given by the failure.

```text
blob:https://ds.infarm.com/e940fbf0-8c73-42bf-a3cd-d5326c57e0f6:1 Uncaught ReferenceError: e is not defined
33../support/isBuffer @ blob:https://ds.infarm.com/e940fbf0-8c73-42bf-a3cd-d5326c57e0f6:1
o @ blob:https://ds.infarm.com/e940fbf0-8c73-42bf-a3cd-d5326c57e0f6:1
[...]

PR #8250, opened 2017-10-30
fix(@angular/cli): Reenable node global

Brocco:


filipesilva:
This PR essentially reverts https://github.com/angular/angular-cli/pull/8130 to fix #8160, but in turn 'un-fixes' #5804.

PR #8255, opened 2017-10-30
New help

Brocco:


packages/@angular/cli/tasks/schematic-get-help-output.ts
```diff
@@ -0,0 +1,62 @@
...
+      project: this.project
+    });
+    return Promise.all([getOptionsTask.run({
+      schematicName: schematicName,
+      collectionName: collectionName,
+    }), nonSchematicOptions])
+    .then(([availableOptions, nonSchematicOptions]: [SchematicAvailableOptions[], any[]]) => {
+      const output: string[] = [];
+      [...(nonSchematicOptions || []), ...availableOptions]
+        .filter(opt => opt.name !== 'name')
```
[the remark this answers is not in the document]
Brocco: that means that things like `--spec` would not be reflected with help, and I don't agree. I filtered out `name` as it is an argument and not an option.

packages/@angular/cli/commands/help.ts
```diff
@@ -67,7 +67,9 @@ const HelpCommand = Command.extend({
           if (commandOptions.short) {
             this.ui.writeLine(command.printShortHelp(commandOptions));
           } else if (command.printDetailedHelp(commandOptions, rawArgs)) {
+            this.ui.writeLine('cool');
```
hansl: ?

packages/@angular/cli/commands/help.ts
```diff
@@ -67,7 +67,9 @@ const HelpCommand = Command.extend({
...
             const result = command.printDetailedHelp(commandOptions, rawArgs);
+            this.ui.writeLine('not cool', result);
```
hansl: ??

PR #8222, opened 2017-10-27
docs: remove output hashing from universal guide example

Toxicable:
To be consistent with usage at the bottom of the guide
closes #8221

PR #8264, opened 2017-10-31
docs(universal): update the build command and output to --output-hashing over --output

beeman:
I was trying to follow the guide when it gave this error:

```
The option '--output' is not registered with the build command. Run `ng build --help` for a list of supported options.
```

Looking further in the doc showed me it should be `--output-hashing=false`


CaerusKaru:
Duplicate of #8222

filipesilva:
Fixed by #8222

PR #8223, opened 2017-10-27
build: update internal Angular dep to 5.0.0

alxhub:

PR #8273, opened 2017-11-01
build: update internal Angular dep to 5.0.0

hansl:
Replaces #8223.


hansl:
Silly CLA bot, Alex and I are both Google employees.

PR #8276, opened 2017-11-01
chore: update codelyzer and add rules

mgechev:
codelyzer@^4.0.0 works with Angular version 5. The PR also introduces
two rules, which we can discuss further:

- `no-on-prefix-output-name` - Follows the style recommending to not
  prefix
  https://angular.io/guide/styleguide#dont-prefix-output-properties
- `angular-whitespace` - Makes sure an expression within interpolation
  is surrounded with whitespace. This rule has a fix as well.

So, in case the code looks like:

```
{{foo + bar}}
```

codelyzer will suggest to add whitespace like:

```
{{ foo + bar }}
```
The fix flag will automatically update the template.

There are few more rules in beta which I will introduce with a PR later.

### DIFF
--- a/tests/e2e/assets/1.0.0-proj/package.json
+++ b/tests/e2e/assets/1.0.0-proj/package.json
@@ -29,7 +29,7 @@
     "@angular/compiler-cli": "^4.0.0",
     "@types/jasmine": "2.5.38",
     "@types/node": "~6.0.60",
-    "codelyzer": "~2.0.0",
+    "codelyzer": "^4.0.0",
     "jasmine-core": "~2.5.2",
     "jasmine-spec-reporter": "~3.2.0",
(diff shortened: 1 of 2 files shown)


filipesilva:
Heya, you actually changed a test that is meant to check that old 1.0.0 projects still work. Can you make this PR in https://github.com/angular/devkit instead? Specifically, in https://github.com/angular/devkit/tree/master/packages/schematics/angular/

mgechev:
:+1:

delasteve:
@mgechev: which way is the rule meant to be defined as, `no-on-prefix-output-name` (in this PR) or `no-output-on-prefix-name` (way you have it defined in Codelyzer)? I'll submit a PR to Codelyzer if you want it the way you have it here.

PR #8290, opened 2017-11-02
fix(build): add `root` option for css-loader

asnowwolf:
So that I can use absolute path (`@import "/app/styles/common"`) to import css files instead of relative path `@import "../../../styles/common"`.

See <https://github.com/webpack-contrib/css-loader#root>.

P.S. My english is not so good, feel free to edit my commit message please.

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/styles.ts
+++ b/packages/@angular/cli/models/webpack-configs/styles.ts
@@ -163,6 +163,7 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
       loader: 'css-loader',
       options: {
         sourceMap: cssSourceMap,
+        root: appRoot,
         importLoaders: 1
       }
     },


clydin:
A root relative URL is akin to an absolute URL and is purposefully not processed by Webpack.
This would also be a fairly large breaking change.

asnowwolf:
@clydin Oh, are there some related issues? does you mean some references to `/assets`

lonr:
@asnowwolf 
see the bottom section of the page:
https://github.com/angular/angular-cli/wiki/stories-global-styles
add
```json
"stylePreprocessorOptions": {
  "includePaths": [
    "app/styles"
  ]
},
```
to `angular-cli.json` `app` object
then `@import "common"` should work
I want to import styles added by npm, so I also added `"../node_modules".

asnowwolf:
@lonr Sorry for my poor english. but `_common.scss` is not the global styles. 

e.g. the common variables should be imported by the component style files.

hansl:
Closing this as it seems to be breaking a lot of our tests and so it's probably better to go back to the drawing board on this fix.

asnowwolf:
Sorry, I did not close this PR in time. Thanks for your help. I learned a lot.

PR #8369, opened 2017-11-06
Add a missing not-supported marker

Tragetaschen:
7d208b73ad06ea4f31bedcca111a45d9561f06ec removed support not only for setting "exclude", but also for overriding "compilerOptions"


hansl:
This is fixed in the new plugin now.

PR #8469, opened 2017-11-13
docs(@angular/cli): fix output path in universal-rendering guide

cre8:
add the correct output path so it matches with the code example


filipesilva:
Heya, can you rebase and name the commit `docs: fix output path in universal-rendering guide` please? That's what's failing the CI.

clydin:
@cre8 you'll need to change the actual commit message not the PR title.

PR #8474, opened 2017-11-14
fix(@angular/cli): sync lint error code behavior with TSLint

clydin:
This synchronizes the return code behavior of `ng lint` and `tslint` wherein lint errors cause a non-zero return value and warnings do not affect the outcome.


filipesilva:
@clydin can you rebase please?

PR #8480, opened 2017-11-14
fix(@angular/cli): throw xi18n errors

### LINKED ISSUE #8065: ng xi18n doesn't report typescript errors
```
- [x] bug report
```

### Versions.
```
Angular CLI: 1.5.0-rc.0 (e)
Node: 6.9.5
OS: win32 x64
Angular: 5.0.0-rc.2

@angular/cli: 1.5.0-rc.0
@angular-devkit/build-optimizer: 0.0.27
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.33
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.0-rc.0
@schematics/angular: 0.0.45
typescript: 2.3.3
webpack-concat-plugin: 1.4.0
webpack-dev-server: 2.7.1
webpack: 3.7.1
```

### Repro steps.
Create a new app with the cli 1.5.0-rc.0 but skip install. Update Angular dependencies to 5.0.0-rc.2, but don't update typescript (keep ~2.3.3).
Update your component to add a `i18n` attribute on some text. For example: `<h1 i18n>Welcome to {{title}}!</h1>`.
Add the following command to your package.json file: `"extract": "ng xi18n -f xlf -l en --of messages.xlf -op src/i18n"` and start it with `npm run extract`.
The app is compiled and... nothing happens. The translations are not extracted.

There is no log and that is the problem. The actual error is:
```sh
Error: The Angular Compiler requires TypeScript >= 2.4.\n    at new AngularCompilerProgram (C:\\www\\i18n-no-cli\\node_modules\\@angular\\compiler-cli\\src\\transformers\\program.js:61:19)\n    at Object.exports.createProgram (C:\\www\\i18n-no-cli\\node_modules\\@angular\\compiler-cli\\src\\transformers\\program.js:650:12)\n    at Object.createProgram (C:\\www\\i18n-no-cli\\node_modules\\@angular\\compiler-cli\\src\\ngtools_api2.js:27:22)\n    at AngularCompilerPlugin._createOrUpdateProgram (C:\\www\\i18n-no-cli\\node_modules\\@ngtools\\webpack\\src\\angular_compiler_plugin.js:234:43)\n    at _donePromise.Promise.resolve.then.then (C:\\www\\i18n-no-cli\\node_modules\\@ngtools\\webpack\\src\\angular_compiler_plugin.js:459:29)
```

But it is not reported. If you go to this line: https://github.com/angular/angular-cli/blob/db9c55c8e27bb1111495ecdbb72b515f7bddaef0/packages/%40angular/cli/tasks/extract-i18n.ts#L39 then you can see that `err` is `null`, but if you log the stats object you will see the errors.
Also the `.catch` method line 53 is called but `err` is `undefined` so nothing happens (I'm not sure if it is related).

### Desired functionality.
It should log the error to the console and fail the extraction instead of looking like everything worked.

ocombe:
Correctly throw xi18n errors (and print warnings) instead of ignoring them.
I've also removed the catch so that the extraction really fails when there is an error.

Fixes #8065

### DIFF
--- a/packages/@angular/cli/tasks/extract-i18n.ts
+++ b/packages/@angular/cli/tasks/extract-i18n.ts
@@ -3,6 +3,8 @@ import * as webpack from 'webpack';
 import { AngularCompilerPlugin } from '@ngtools/webpack';
 import { XI18nWebpackConfig } from '../models/webpack-xi18n-config';
 import { getAppFromConfig } from '../utilities/app-utils';
+import {getWebpackStatsConfig} from '../models/webpack-configs';
+import {statsErrorsToString, statsWarningsToString} from '../utilities/stats';
 
 const Task = require('../ember-cli/lib/models/task');
 const MemoryFS = require('memory-fs');
@@ -34,6 +36,7 @@ export const Extracti18nTask = Task.extend({
 
     const webpackCompiler = webpack(config);
     webpackCompiler.outputFileSystem = new MemoryFS();
+    const statsConfig = getWebpackStatsConfig(runTaskOptions.verbose);
 
     return new Promise((resolve, reject) => {
       const callback: webpack.compiler.CompilerCallback = (err, stats) => {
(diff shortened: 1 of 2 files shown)


packages/@angular/cli/tasks/extract-i18n.ts
```diff
@@ -34,27 +36,32 @@ export const Extracti18nTask = Task.extend({
...
+    const statsConfig = getWebpackStatsConfig(runTaskOptions.verbose);
 
     return new Promise((resolve, reject) => {
       const callback: webpack.compiler.CompilerCallback = (err, stats) => {
         if (err) {
           return reject(err);
         }
 
+        const json = stats.toJson('verbose');
+        if (runTaskOptions.verbose) {
```
filipesilva: I'd prefer we didn't output bundle stats in `xi18n` since those bundles aren't actually generated on disk and we didn't  outputt this information for `xi18n` before. Can you leave just the warning/error logic please?

PR #8488, opened 2017-11-14
misc(payload-size): Upload payload size to angular firebase database

tinayuangao:
- Secure the firebase access token
- Upload compressed/uncompressed payload size to angular-payload-size firebase database
- Path is /payload/cli/$branch/$commit
- With commit message, and timestamp


tinayuangao:
Wrong branch name. Close this PR

PR #8519, opened 2017-11-17
fix(@ngtools/webpack): fix elide removing whole imports on single match

### LINKED ISSUE #8518: v1.5.1 AOT Build - Uncaught ReferenceError: InjectionToken is not defined
### Versions
```
Angular CLI: 1.5.1
Node: 8.6.0
OS: win32 x64
Angular: 5.0.1
... service-worker

@angular/cli: 1.5.1
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.36
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.1
@schematics/angular: 0.1.5
typescript: 2.4.2
webpack: 3.8.1
```

### Repro steps

* `ng serve --aot`

### Observed behavior
```
Uncaught ReferenceError: InjectionToken is not defined
    at eval (logger.service.ts:20)
    at Object.../../../../../src/app/shared/services/logger/logger.service.ts (main.bundle.js:206)
    at __webpack_require__ (inline.bundle.js:713)
    at fn (inline.bundle.js:118)
    at eval (logger.module.ts:2)
    at Object.../../../../../src/app/shared/services/logger/logger.module.ts (main.bundle.js:198)
    at __webpack_require__ (inline.bundle.js:713)
    at fn (inline.bundle.js:118)
    at eval (index.ts:1)
    at Object.../../../../../src/app/shared/services/index.ts (main.bundle.js:190)
```

### Desired behavior

The app builds fine in AOT mode with no error, but when I run it in the browser, it results in the error provided above during bootstrap process. I tried to reproduce this error on CLI `1.5.0` and it works fine. It complains about `InjectionToken` being not defined.

Only happening on `v1.5.1`, works on `v1.5.0`. No other dependency was changed. All I did was upgrade `@angular/cli` from `1.5.0` to `1.5.1`.

hansl:
If an import imports multiple symbols, the previous condition meant that the whole
import statement was removed, instead of only the symbol.

Fixes #8518.

### DIFF
--- a/packages/@ngtools/webpack/src/transformers/elide_imports.ts
+++ b/packages/@ngtools/webpack/src/transformers/elide_imports.ts
@@ -113,7 +113,7 @@ export function elideImports(
     })
     .forEach((symbol) => {
       // Remove the whole declaration if it's a single import.
-      const nodeToRemove = symbol.singleImport ? symbol.importSpec : symbol.importDecl;
+      const nodeToRemove = symbol.singleImport ? symbol.importDecl : symbol.importSpec;
       ops.push(new RemoveNodeOperation(sourceFile, nodeToRemove));
     });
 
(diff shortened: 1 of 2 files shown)


hansl:
I verified that the test was failing BEFORE this fix, now passes.

PR #8589, opened 2017-11-22
Work with typescript files without content

PanayotCankov:
Too bad we have empty typescript files and their content is "" which is a false-y value.
On the other hand:
```
> null != undefined
false
> undefined != undefined
false
> "" != undefined
true
```
I am not sure this is a proper "fix" but please mind it should work with empty string. Usually we won't have empty files but on few occasions we use a mechanism to do a preprocessor kind of stuff that will output some TypeScript code for some builds but have none for others. Then web pack running uglify will strip comments and compress whitespace leading to "".


filipesilva:
There's also the issue of the CLA.

PR #8604, opened 2017-11-22
fix(@ngtools/webpack): do not change user's `LOCALE_ID`

ocombe:
We need to normalize the locale to load i18n locale data files, but we should not change the actual user's locale that we pass to the compiler.

Fixes https://github.com/angular/angular/issues/20340

### DIFF
--- a/packages/@ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/@ngtools/webpack/src/angular_compiler_plugin.ts
@@ -108,6 +108,7 @@ export class AngularCompilerPlugin implements Tapable {
   private _donePromise: Promise<void> | null;
   private _compiler: any = null;
   private _compilation: any = null;
+  private _normalizedLocale: string;
 
   // TypeChecker process.
   private _forkTypeChecker = true;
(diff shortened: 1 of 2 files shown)


packages/@ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -652,9 +654,9 @@ export class AngularCompilerPlugin implements Tapable {
       // If we have a locale, auto import the locale data file.
       // This transform must go before replaceBootstrap because it looks for the entry module
       // import, which will be replaced.
-      if (this._compilerOptions.i18nInLocale) {
+      if (this._compilerOptions._normalizedLocale) {
```
filipesilva: Is `_normalizedLocale` a new option in `AngularCompilerOptions`?

hansl: Seems like he’s reading from the instance but setting in the options.

hansl: Surprised that this passes build.

ocombe: sorry, my mistake, fixing this!

PR #8643, opened 2017-11-27
feat: direct ngc compilation with a webpack compiler host

shlomiassaf:
This PR add's the missing bridge to `ngc` with a wepack compiler host.

Now it is possible to execute a compilation (not bundling) process similar to `ngc -p ...` but with a webpack enabled resource loader, no more gulp process, scss, postcss... just provide a webpack config and a tsconfig.

The process will inline resources when skipping template codegen. Inline is supported for JS files and metadata.json files (including flat bundle)

webpack is not used to run any bundling, just to compile resources on demand.

This is the first step for providing library support, it completes the "angular" gap of dealing with AOT resources.

The PR does not include cli interface, just node API, I guess CLI should go elsewhere using this node API.

Usage is simple:

```js
    require('@ngtools/webpack')
      .compile('webpack.config.js', 'tsconfig.json')
      .then( () => process.exit() )
      .catch( err => {
        process.exit(99);
        console.error(err);
      });
```

No change to the current code base, only additions.

Also, once angular/angular#20130 is added some code can be removed

### DIFF
--- a/packages/@ngtools/webpack/src/index.ts
+++ b/packages/@ngtools/webpack/src/index.ts
@@ -15,6 +15,7 @@ try {
     + e);
 }
 
+export * from './ngc_bridge';
 export * from './plugin';
 export * from './angular_compiler_plugin';
 export * from './extract_i18n_plugin';
--- a/packages/@ngtools/webpack/src/ngc_bridge/index.ts
+++ b/packages/@ngtools/webpack/src/ngc_bridge/index.ts
@@ -1,0 +1,1 @@
+export { compile, CompilationResult } from './run';
--- a/packages/@ngtools/webpack/src/ngc_bridge/inline_metadata.ts
+++ b/packages/@ngtools/webpack/src/ngc_bridge/inline_metadata.ts
@@ -1,3 +1,4 @@
+// @ignoreDep @angular/core
 import * as Path from 'path';
 
 import { Component } from '@angular/core';
--- a/packages/@ngtools/webpack/src/ngc_bridge/inline_metadata.ts
+++ b/packages/@ngtools/webpack/src/ngc_bridge/inline_metadata.ts
@@ -1,0 +1,67 @@
+import * as Path from 'path';
+
+import { Component } from '@angular/core';
+import {
+  ModuleMetadata,
+  MetadataSymbolicCallExpression,
+  isClassMetadata,
+  isMetadataSymbolicCallExpression,
+  isMetadataImportedSymbolReferenceExpression
+} from '@angular/compiler-cli';
+
+
+function hasResources(obj: any): obj is Component {
+  return obj.hasOwnProperty('templateUrl') || obj.hasOwnProperty('styleUrls');
+}
+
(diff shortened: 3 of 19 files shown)


shlomiassaf:
@hansl @filipesilva 

The dependency validation test is failing on 3 packages:

  - `@angular/core`
This is used [here](https://github.com/shlomiassaf/angular-cli/blob/a591dfee08bc336abe30aa8ade8b13643679820b/packages/%40ngtools/webpack/src/ngc_bridge/inline_metadata.ts#L3)
It is only used as type so it doesn't go to runtime.
I can remove it and use any...

  - `@angular/compiler` and `tsickle`
These are used on the same module [here](https://github.com/shlomiassaf/angular-cli/blob/a591dfee08bc336abe30aa8ade8b13643679820b/packages/%40ngtools/webpack/src/ngc_bridge/perform_compile_async.ts#L15-L16)

This is local implementation of `performCompilation` from `@angular/compiler-cli` but async so
webpack can do it's work.
I opened an issue for it  angular/angular#20130 with 2 proposed solutions in my branch (not PRed yet)
If we can push that quickly we can remove this implementation however it means not supporting previous versions of the compiler.

Suggestions?

shlomiassaf:
never mind, just read the deps script and figured out how to ignore them.

shlomiassaf:
never mind, just read the deps test script and found how to mark them to be ignored

hansl:
@shlomiassaf What problem exactly are you trying to fix? What are your requirements? This is just a code dump without actual context, so I'm left scratching my head here.

shlomiassaf:
@hansl TL;DR: This will allow compiling libraries with webpack as resource loader.

Current the ngtools/webpack is only for application development.
When creating libraries we don't need most of it (can't use it actually) but we do need other features within the ngtools/webpack, mainly the resource loaders.

With this PR it is now possible to run `ngc` with angular component that reference SCSS, LESS or anything you have a loader from and use your already existing webpack configuration to build the library.

The PR also adds inlining support so the output of resources will go into `template` and `styles` in both `js` files and `metadata.json` files.

For example, the material repository has a build process that uses `ngc` to compile the components.
Before and after `ngc` they use gulp to: 
  - Identify all resources in all component (templateUrl, styleUrls)
  - compile `css` files from all `scss` files referenced by `styleUrls`.
  - inline all resources found into the source code

With this plugin they don't need that, it will do that.

Also, worth mentioning:
  - They use a different process of CSS to SCSS (gulp) then what they develop with (webpack)
  - The process is complex
  - Its a lot faster with this PR since it's native to the compilation process.


Simply put, we are "teaching" the compiler how to resolve resources through our already existing webpack loader chain.

> This does not run webpack bundling, it will only create a webpack instance for compiling resources.

matheo:
@hansl this sounds like a big step into the library support which discussion you may know.
Thanks for your work @shlomiassaf and @dherges

hansl:
We have a path towards library support for the next major, using Architect and ng-packagr. I'm closing this PR in favor of using that solution. Thanks for your effort.

Issue #7991, opened 2017-10-10
postcss-custom-properties warnings when overriding css variables


Morente5:
### Bug Report
From Issue #7514 
Related to PR #7712 

### Versions.
```
Angular CLI: 1.5.0-rc.0
Node: 8.1.2
OS: win32 x64
Angular: 5.0.0-rc.2
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 1.5.0-rc.0
@angular-devkit/build-optimizer: 0.0.25
@angular-devkit/core: 0.0.17
@angular-devkit/schematics: 0.0.30
@ngtools/json-schema: 1.1.0
@schematics/angular: 0.0.41
typescript: 2.5.3
webpack: 3.7.1
```

### The log given by the failure.
A lot of warnings are being shown to me since #7712 : 
```
WARNING in ./path/to/component/some-component.component.scss
(Emitted value instead of an instance of Error) postcss-custom-properties: path\to\component\some-component.component.scss: Custom property ignored: not scoped to the top-level :root element (some-css-selector { ... --custom-property: ... })
    at Object.emitWarning (path-to-my-project\node_modules\webpack\lib\NormalModule.js:117:16)
    at path-to-my-project\node_modules\postcss-loader\index.js:131:24
    at Array.forEach (native)
    at path-to-my-project\public\node_modules\postcss-loader\index.js:130:31
    at <anonymous>
```

### Repro steps.
when I set:
```css
:root {
  --custom-property: 200; 
}
```
on `styles.scss`,
and I override it in a component, `some-component.scss`
```css
some-css-selector {
  --custom-property: 100; 
}
```

It's working perfectly fine but it's very annoying seeing a warning for each css variable I override.

May that be related to a [`postcss-custom-properties`](https://github.com/postcss/postcss-custom-properties) limitation?

renestalder:
postcss-custom-properties doesn't support variable declarations on CSS selectors other than :root. Only :root is allowed. That's why you get the warning. Mostly this only means that your variable will not be transpiled back for older browsers. So when you want  this to work with older browsers, you have to use :root all the time, so postcss-custom-properties knows what to do with it.

adrm:
Is there a way to silence this warning or noting to the tool that I don't care about transpiling this feature back to older browsers? Thanks.

renestalder:
[...]

PR #8667, opened 2017-11-28
fix(@angular/cli): postCssOption

jycouet:
Fix: https://github.com/angular/angular-cli/issues/7991

Like this, we can turn off the flag `withPostCssWarnings` in `.angular-cli.json`:
```
"defaults": {
  "build": {
    "withPostCssWarnings": false
  }
}
```

The default value is set to `true` so that we have the same behaviour as the current v1.5.x

### DIFF
--- a/tests/e2e/tests/build/with-post-css-warnings.ts
+++ b/tests/e2e/tests/build/with-post-css-warnings.ts
@@ -1,0 +1,43 @@
+import {
+  killAllProcesses,
+  waitForAnyProcessOutputToMatch,
+  execAndWaitForOutputToMatch
+} from '../../utils/process';
+import { appendToFile } from '../../utils/fs';
+import { getGlobalVariable } from '../../utils/env';
+import { updateJsonFile } from '../../utils/project';
+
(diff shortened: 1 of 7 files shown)


tests/e2e/tests/build/with-post-css-warnings.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  killAllProcesses,
+  waitForAnyProcessOutputToMatch,
+  execAndWaitForOutputToMatch
+} from '../../utils/process';
+import { appendToFile } from '../../utils/fs';
+import { getGlobalVariable } from '../../utils/env';
+import { updateJsonFile } from '../../utils/project';
+
+const webpackGoodRegEx = /webpack: Compiled successfully./;
```
YoannBureau: rename to Webpack Success

jycouet: I will do this, but it won't fix failing tests right?

clydin:
Superseded by #8874

PR #8669, opened 2017-11-29
Enable usage of @nrwl/nx on Windows

bryanrideshark:
For usage with `@nrwl/nx`, which nests things more than one level deep, replace('/') is not adequate as it only replaces the first occurrence of '/'. This uses a regex instead, which enables the usage of `@nrwl/nx` on Windows.


bryanrideshark:
I signed it!

bryanrideshark:
I really, really signed it....

bryanrideshark:
@filipesilva , I wasn't sure how to change the commit message on my side of things, so I've submitted another PR.

PR #8681, opened 2017-11-29
fix(@angular/cli): support ejection of service worker apps

alxhub:


packages/@angular/cli/tasks/eject.ts
```diff
@@ -554,6 +554,16 @@ export default Task.extend({
         packageJson['scripts']['pree2e'] = pree2eNpmScript;
         packageJson['scripts']['e2e'] = 'protractor ./protractor.conf.js';
 
+        if (runTaskOptions.environment === 'production' && usesServiceWorker(project.root)
```
filipesilva: I think this is `runTaskOptions.target` instead of `.enviroment`.

packages/@angular/cli/tasks/eject.ts
```diff
@@ -554,6 +554,16 @@ export default Task.extend({
...
+            && !!runTaskOptions.serviceWorker) {
```
filipesilva: I don't think we have a option for service worker proper. It's just a flag in `.angular-cli.json`.

filipesilva: You should be able to get it as `appConfig.serviceWorker`

alxhub: I added both checks.

packages/@angular/cli/tasks/eject.ts
alxhub: Fixed.

PR #8704, opened 2017-12-01
Fixed inline link and inline code

vaidd4:
Apparently the GH Markdown doesn't render well (at all ?) when inside html tags.


vaidd4:
Just signed my CLA.

vaidd4:
I will re-make a better described commit

PR #8706, opened 2017-12-01
docs(@angular/cli): fix inline link and code formatting

vaidd4:
Change markdown syntax for link and inline code for html tags. Apparently the GH Markdown doesn't render well (at all ?) when inside html tags.

### DIFF
--- a/docs/documentation/e2e.md
+++ b/docs/documentation/e2e.md
@@ -23,7 +23,7 @@ Please note that options that are supported by `ng serve` are also supported by 
     <code>--config</code> (aliases: <code>-c</code>)
   </p>
   <p>
-    Use a specific config file. Defaults to the protractor config file in `.angular-cli.json`.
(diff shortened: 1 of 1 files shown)


vaidd4:
As its the commit name that is failing, what should I rename it in ?

clydin:
As this is documentation for the CLI, `docs(@angular/cli): fix inline link and code formatting` would probably be appropriate.  For further details, please see https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#commit.

Note that a new PR is not needed to change the commit message.  Try something similar to the following:
```
git commit --amend -m "<commit message here>"
git push --force
```

vaidd4:
Oh okay, I admit I've read the contribution guidelines for the angular project but not for CLI, my bad.

I fixed the commit message and title of the PR.

PR #8707, opened 2017-12-01
ci: run optionals on master only

filipesilva:


filipesilva:
The conditional doesn't seem to be working, or I have it configured wrong. Opened an issue in https://github.com/travis-ci/travis-ci/issues/8846.

PR #8723, opened 2017-12-02
feat(@angular/cli): specify multiple default collections

tomastrajan:
Hi 😉

This PR is related to [issue](https://github.com/angular/devkit/issues/249) posted on `angular/devkit` repository by @jeffbcross . It looks like the whole implementation of executing schematics belongs to `@angular/cli` itself so that's the reason for creating PR against this repo.

The proposed solution adds possibility to specify additional `collections` property in defaults of `schematics` in `angular-cli.json`. That way app has base collection which defaults to `@schematics/angular` (but could be overridden too) and potential to add unlimited additional collections. Execution priority is that the first collection containing requested schematics wins.
the additional `collections` property should make this solution backwards compatible.

This is initial working implementation. Review of someone who knows that part would be highly appreciated. Also some guidance on how to test this is very welcome.

Check out screenshot of working solutions.

schema1
schema2
schema3

### DIFF
--- a/packages/@angular/cli/commands/generate.ts
+++ b/packages/@angular/cli/commands/generate.ts
@@ -7,6 +7,8 @@ import 'rxjs/add/observable/of';
 import 'rxjs/add/operator/ignoreElements';
 import {
   getCollection,
+  getCollectionNameForSchematicName,
+  getCollectionNames,
   getEngineHost
 } from '../utilities/schematics';
 import { DynamicPathOptions, dynamicPathParser } from '../utilities/dynamic-path-parser';
@@ -17,7 +19,7 @@ import { SchematicAvailableOptions } from '../tasks/schematic-get-options';
 const Command = require('../ember-cli/lib/models/command');
 const SilentError = require('silent-error');
 
-const { cyan, yellow } = chalk;
+const { cyan, yellow, white } = chalk;
 const separatorRegEx = /[\/\\]/g;
 
 
@@ -65,8 +67,10 @@ export default Command.extend({
     '<schematic>'
   ],
 
-  getCollectionName(rawArgs: string[]) {
-    let collectionName = CliConfig.getValue('defaults.schematics.collection');
+  getCollectionName(schematicName: string, rawArgs: string[]) {
+    let collectionName = getCollectionNameForSchematicName(
+      getCollectionNames(), schematicName);
+
     if (rawArgs) {
       const parsedArgs = this.parseArgs(rawArgs, false);
       if (parsedArgs.options.collection) {
@@ -103,7 +107,7 @@ export default Command.extend({
       ui: this.ui,
       project: this.project
     });
-    const collectionName = this.getCollectionName(rawArgs);
+    const collectionName = this.getCollectionName(schematicName, rawArgs);
 
     return getOptionsTask.run({
         schematicName,
@@ -172,7 +176,7 @@ export default Command.extend({
       project: this.project
     });
     const collectionName = commandOptions.collection ||
-      CliConfig.getValue('defaults.schematics.collection');
+      this.getCollectionName(schematicName);
 
     if (collectionName === '@schematics/angular' && schematicName === 'interface' && rawArgs[2]) {
       commandOptions.type = rawArgs[2];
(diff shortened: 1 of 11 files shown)


packages/@angular/cli/tasks/schematic-get-options.ts
```diff
@@ -19,10 +18,7 @@ export interface SchematicAvailableOptions {
 
 export default Task.extend({
   run: function (options: SchematicGetOptions): Promise<SchematicAvailableOptions[]> {
-    const collectionName = options.collectionName ||
-      CliConfig.getValue('defaults.schematics.collection');
```
tomastrajan: No need to retrieve config value from multiple places. Now it's always responsibility of the  caller.

packages/@angular/cli/tasks/init.ts
```diff
@@ -72,15 +72,14 @@ export default Task.extend({
     });
 
     const cwd = this.project.root;
-    const schematicName = CliConfig.fromGlobal().get('defaults.schematics.newApp');
```
tomastrajan: No need to retrieve config value from multiple places. Now it's always responsibility of the caller.

packages/@angular/cli/commands/generate.ts
```diff
@@ -204,16 +207,22 @@ export default Command.extend({
...
     } else {
-      const schematicNames: string[] = engineHost.listSchematics(collection);
       const output: string[] = [];
-      output.push(cyan('Available schematics:'));
-      schematicNames.forEach(schematicName => {
-        output.push(yellow(`    ${schematicName}`));
+      output.push(cyan('Available collections & schematics:'));
+      const collections = getCollectionNames()
+        .map((collectionName: string) => getCollection(collectionName));
+      collections.forEach((collection: any) => {
```
tomastrajan: List schematics per collection

tomastrajan:
@clydin  a bit meta question, how does it work with this repo, is it ok to always mention main guys in PR to let them know something was created or do they check PRs themselves, don't want to spam.

tomastrajan:
Hi @Brocco , I have seen you did PR #8812 concerning schematics and collections, maybe you can have a look on this one too? Its about support for multiple collections per project.

clydin:
@tomastrajan we monitor all the PRs but if you think a PR should be reviewed by a particular team member, there's no harm in mentioning them.

Also, thank you for contributions to the project and spending the time to help make the CLI better.  Bug fixes and required features take priority so please don't feel discouraged if it takes a little while to hear back.

bmayen:
Is this necessary in order to, say, use default Angular schematics and the new ngrx schematics in the same project?

tomastrajan:
@bmayen @hansl No, you can use any collection with passing its name to collection flag like `ng g someCustomSchematics --collection someCustomCollection` which has to be installed in `node_modules`. 

This PR is about being able to just run `ng generate someCustomSchenatics` without the need for collection flag. You could specify included collections in `angular-cli.json` in order of priority. This would be great for authors of single specific schematics so that you dont have to remember all the collection names every time you want to execute them. It would be enough to add them just once.

bmayen:
Certainly much more convenient being able to define, only once in angular-cli.json, all of the collections you'll be using for a project.

I do think more thought needs to go into name collisions. Currently the first one encountered wins. This is fine when the intent is always to override, but we can't assume that. It's possible that two separate collections define `foo`, which are intended for entirely different purposes, and we'd want access to both of them. Otherwise we're assuming that any schematic name in a collection is reserved and will be overridden, potentially accidentally, if used by another collection.

tomastrajan:
I see your point. The way this was intended to work is that you get override behaviour by default but you can always execute for specific collection in the same fashion like befote by passing `--collection` flag.

Besides from what I saw in `ngrx` schematics it seems to be the case that the way this is going is to use `extends` in schematics json declarations to pass through calls to underlying default angular schematics for stuff like component or service.

But this then only works nicely if you use ONE other schematics which delegates all the calls to underlying default. It doesnt help if you want to use more smaller / single purpose schematics from multiple libraries.

bmayen:
> But this then only works nicely if you use ONE other schematics which delegates all the calls to underlying default. It doesnt help if you want to use more smaller / single purpose schematics from multiple libraries.

This would be my main concern. It requires some knowledge of the implementation or behavior could be unexpected.

Tiberriver256:
Just sharing my pre-conceived notion on how this was going to work. When I saw it was prefixed with @schematics, I thought it was going to be similar to how @types is managed by DefintelyTyped.

Package owners create a PR to merge schematics for their package into the main repo which is then auto-published after approval. This would solve the naming conflicts and it would be nice to just see them all available schematics listed in NPM under @schematics but I'm not sure how much overhead it really takes those Microsoft guys to manage it that way. Just a thought.

hansl:
Closing this PR as it is too old and does not merge properly anymore. The work has been tracked locally in https://github.com/angular/angular-cli/issues/12157 (which is a transfer from the issue from devkit).

It's something we want to support but needed to do some cleanup first. Now that 7.0 is out and that cleanup is done, we'll track it in our internal Jira and expect updates on the issue #12157.

Issue #8247, opened 2017-10-30
Angular CLI 1.5.0-rc.6 does not support @angular/service-worker@5.0.0-rc.8


niikoo:
### Bug Report or Feature Request (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.

**ng -v**

Angular CLI: 1.5.0-rc.6
Node: 8.5.0
OS: win32 x64
Angular: 5.0.0-rc.8
... animations, common, compiler, compiler-cli, core, forms
... http, platform-browser, platform-browser-dynamic
... platform-server, router

@angular/cdk: 2.0.0-beta.12
@angular/cli: 1.5.0-rc.6
@angular/material: 2.0.0-beta.12
@angular/service-worker: 5.0.0-rc.8
@angular-devkit/build-optimizer: 0.0.31
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.34
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.0-rc.6
@schematics/angular: 0.0.48
typescript: 2.5.3
webpack: 3.8.1

### Repro steps.

ng build --prod --aot --base-href=.

### The log given by the failure.

**Build fails with error: **

The installed version of @angular/service-worker is 5.0.0-rc.8. This version of the CLI
requires the @angular/service-worker version to satisfy >= 1.0.0-beta.5 < 2.0.0. Please upgrade
your service worker version.
Error: The installed version of @angular/service-worker is 5.0.0-rc.8. This version of the CLI
requires the @angular/service-worker version to satisfy >= 1.0.0-beta.5 < 2.0.0. Please upgrade
your service worker version.

### Desired functionality.

The newest @angular/cli isn't compatible with any of the new @angular/service-worker versions; which it should.

### Mention any other details that might be useful.

The old version of @angular/service-worker is working, 1.0.0-beta.16.

avatsaev:
Same issue here

cgatian:
Perhaps this is by design. Using 1.5 CLI and 5.0 @angular/service-worker also reports this error

csentis:
Just looked at the latest commits and saw that service-worker got pinned @1.0.0-beta.16 for the 1.5.0 release.

A pity :)

Splaktar:
Yep, `@angular/service-worker` support did not make it into Angular CLI `v1.5.0`.

MattiJarvinen-BA:
[...]

PR #8729, opened 2017-12-02
fix(@angular/cli): allow RCs and betas for @angular/service-worker

cexbrayat:
It was not possible to use `5.1.0-rc.1` previously (see https://github.com/angular/angular-cli/issues/8247#issuecomment-348722905),
because of how `semver.satisfies()` works (see https://github.com/npm/node-semver#prerelease-tags).
This relaxes the check by using `semver.gte()`.

Related to #8247

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/production.ts
+++ b/packages/@angular/cli/models/webpack-configs/production.ts
@@ -47,7 +47,7 @@ export function getProdConfig(wco: WebpackConfigOptions) {
     const swVersion = JSON.parse(swPackageJson)['version'];
 
     const isLegacySw = semver.satisfies(swVersion, OLD_SW_VERSION);
-    const isModernSw = semver.satisfies(swVersion, NEW_SW_VERSION);
+    const isModernSw = semver.gte(swVersion, NEW_SW_VERSION);
 
     if (!isLegacySw && !isModernSw) {
       throw new Error(stripIndent`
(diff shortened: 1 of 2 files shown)


cexbrayat:
(build failed for unrelated reasons I think)

hansl:
I'll let @alxhub review, but you should be using `gte()` instead so if they're equals it matches.

packages/@angular/cli/models/webpack-configs/production.ts
```diff
@@ -47,7 +47,7 @@ export function getProdConfig(wco: WebpackConfigOptions) {
     const swVersion = JSON.parse(swPackageJson)['version'];
 
     const isLegacySw = semver.satisfies(swVersion, OLD_SW_VERSION);
-    const isModernSw = semver.satisfies(swVersion, NEW_SW_VERSION);
+    const isModernSw = semver.gt(swVersion, NEW_SW_VERSION);
```
alxhub: Per @hansl, this should be `.gte`. But otherwise, nice!

packages/@angular/cli/utilities/service-worker/index.ts
```diff
@@ -58,7 +58,7 @@ export function usesServiceWorker(projectRoot: string): boolean {
   const swPackageJson = fs.readFileSync(`${swModule}/package.json`).toString();
   const swVersion = JSON.parse(swPackageJson)['version'];
 
-  return semver.satisfies(swVersion, NEW_SW_VERSION);
+  return semver.gt(swVersion, NEW_SW_VERSION);
```
alxhub: Same here.

cexbrayat:
Of course, you're right. Now fixed!

PR #8740, opened 2017-12-05
feat(@angular/cli): add trailing slash to deployUrl when missing

### LINKED ISSUE #7295: deployUrl (ng build) doesn't seem to add a trailing slash
```
- [ ] feature request
```

### Versions.

@angular/cli: 1.2.7
node: 7.10.0
os: darwin x64
@angular/animations: 4.3.3
@angular/cli: 1.2.7
@angular/common: 4.3.3
@angular/compiler: 4.3.3
@angular/core: 4.3.3
@angular/forms: 4.3.3
@angular/http: 4.3.3
@angular/platform-browser: 4.3.3
@angular/platform-browser-dynamic: 4.3.3
@angular/router: 4.3.3
@angular/compiler-cli: 4.3.3
@angular/language-service: 4.3.3

### Repro steps.

`$ ng build --deployUrl=some/path`

Some resources (like font awesome loaded via npm) do not load because they're missing a slash in the middle.
e.g.: `GET 404: http://localhost:8080/some/pathfontawesome-webfont.af7ae505a9eed503f8b8.woff2`

### Desired functionality.

Add a trailing slash to the deployUrl (if needed) when forming the path.
e.g.: `http://localhost:8080/some/path/fontawesome-webfont.af7ae505a9eed503f8b8.woff2`

tomastrajan:
Closes #7295

### DIFF
--- a/packages/@angular/cli/commands/build.ts
+++ b/packages/@angular/cli/commands/build.ts
@@ -239,6 +239,11 @@ const BuildCommand = Command.extend({
       commandOptions.forceTsCommonjs = true;
     }
 
+    // Add trailing slash if missing to prevent https://github.com/angular/angular-cli/issues/7295
+    if (commandOptions.deployUrl && commandOptions.deployUrl.substr(-1) !== '/') {
+      commandOptions.deployUrl += '/';
+    }
+
     const BuildTask = require('../tasks/build').default;
 
     const buildTask = new BuildTask({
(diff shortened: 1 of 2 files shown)


filipesilva:
@clydin can you review please?

clydin:
As currently implemented, the deploy URL option mimics the behavior of webpack's public path option which simply prepends the value provided to resource requests.  So technically this PR would be a breaking change.  However, I'm not sure anyone is currently using the option to prepend directly to the filename of the resource (as opposed to the common case of adjusting the path).

Also of note, the CLI's base href option also would typically require the user to use a trailing slash.  According to the specification, the trailing slash is actually significant to the meaning of the base href value (i.e., `/base/` and `/base` have different behavior within a web browser).

tomastrajan:
@clydin I had similar worries and [asked Webpack guys directly](https://twitter.com/tomastrajan/status/937919027971629056) and got response that it is always safe to append trailing slash when missing on publicPath. Then again I the context in the question was lacking so it's possible that answer didn't account for all use cases.

clydin:
This is something that webpack should probably fix directly but changing it here as well shouldn't cause a problem.

Issue #8292, opened 2017-11-02
Warning order


jsBrian:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x ] feature request
```

### Versions.
```
@angular/cli: 1.4.5
node: 8.7.0
os: win32 x64
@angular/common: 4.4.4
@angular/compiler: 4.4.4
@angular/core: 4.4.4
@angular/forms: 4.4.4
@angular/platform-browser: 4.4.4
@angular/platform-browser-dynamic: 4.4.4
@angular/router: 4.4.4
@angular/cli: 1.4.5
@angular/compiler-cli: 4.4.4
typescript: 2.2.2
```

### Desired functionality.
This is of course cosmetic, but currently in the log we see the build info first, errors after that and warnings are last.
Doesn't it make more sense to move the warning messages?

So, first show the warnings, after that the build info and errors last. 

That way, if your output screen is small, the warnings can't hide potential errors.

m-henderson:
@filipesilva I'm sure you get this all the time, but could you give me a starting point to fixing this issue ? I am a first time contributor to angular and I dont know my way around. thanks

Taha-Di-Nero:
https://github.com/angular/angular-cli/pull/8745

filipesilva:
@m-henderson sorry I didn't see your comment before, by now there's already a fix up.

@Taha-Di-Nero PR fixes this pretty well, we'll get that one on.

filipesilva:
I think whatever the problem was (that showed build info, errors and then warnings) has been fixed since. 

@Taha-Di-Nero had a PR up but I've closed it since (see https://github.com/angular/angular-cli/pull/8745#issuecomment-356313604).

PR #8745, opened 2017-12-05
fix(@angular/cli): show errors after warnings

Taha-Di-Nero:


Taha-Di-Nero:
Of course

Taha-Di-Nero:
@filipesilva why merging is blocked ??

filipesilva:
Sorry it took awhile for me to come back to this PR. I've re-reviewed it and it should be ready. Thanks!

clydin:
It looks like this is moving the warnings before the output stats.  I'm not sure this is a good idea as this can cause any warnings to be easily missed.  Warnings and errors are both diagnostic messages and I think it would make more sense to group them together as such.  Also, errors are already shown after warnings.  So the current order is information, warnings, errors.  This represents an increasing level of severity and therefore shows the most urgent message last; which could be considered the order of most utility.

In regards to the actual underlying concern of too much text being displayed in a small console window, PR #8875 should provide partial relief.  In addition, further work on reducing the verbosity of warnings and errors would also be a fruitful effort in this regard.

filipesilva:
To be honest I misunderstood the original issue (#8292). I thought warnings were actually showing after errors. 

I agree with @clydin here: the warnings are already being shown before the errors so I don't think there's any need to change.

Apologies for that... I should have read read the issue and PR with more attention.

PR #8772, opened 2017-12-07
feat(@angular/cli): display app index or name in process title when s…

### LINKED ISSUE #6888: Display app name in process.title if running specific app
```
- [X] feature request
```

### Versions.

@angular/cli: 1.1.3
node: 8.1.0
os: win32 x64
@angular/animations: 4.2.4
@angular/cli: 1.1.3
@angular/common: 4.2.4
@angular/compiler: 4.2.4
@angular/compiler-cli: 4.2.4
@angular/core: 4.2.4
@angular/forms: 4.2.4
@angular/http: 4.2.4
@angular/platform-browser: 4.2.4
@angular/platform-browser-dynamic: 4.2.4
@angular/platform-server: 4.2.4
@angular/router: 4.2.4
@angular/service-worker: 1.0.0-beta.16

### Repro steps.

ng serve --app 1
Other window: 
ng serve --app 2

Powershell Window title 1: @angular/cli
Powershell Window title 2: @angular/cli

### Desired functionality.

If you're developing multiple apps inside one project, the process title for all running angular/cli processes is set to "@angular/cli". This is used in for example Powershell as the Window title. It would be nice to see the app name or the host name in the window title. This makes it much easier to see which angular-cli instance is running inside that window.

Powershell Window title 1: @angular/cli - appname1
Powershell Window title 2: @angular/cli - appname2

tomastrajan:
…pecified

Closes #6888

This sets process title to `ng - app <index>` when used with `--app 0` 
or to `ng - <app-name>` when used with `--app appName`

### DIFF
--- a/packages/@angular/cli/utilities/app-utils.ts
+++ b/packages/@angular/cli/utilities/app-utils.ts
@@ -14,11 +14,13 @@ export function getAppFromConfig(nameOrIndex?: String) {
     if (nameOrIndex.match(/^[0-9]+$/)) {
       const index = parseInt(nameOrIndex.toString(), 10);
       if (apps[index]) {
+        process.title = `ng - app ${index}`;
         return apps[index];
       }
     } else {
       const filtered = apps.filter((currentApp: any) => currentApp.name === nameOrIndex);
       if (filtered.length > 0) {
+        process.title = `ng - ${filtered[0].name}`;
         return filtered[0];
       }
     }


tomastrajan:
Had tests like 

```javascript
import { killAllProcesses } from '../../../utils/process';
import { ngServe } from '../../../utils/project';

export default function () {
  return Promise.resolve()
    .then(() => ngServe('--app', '0'))
    .then(() => {
      if (!process.title.match(/ng - app 0/)) {
        throw new Error('Process title does not match expected value.');
      }
    })
    .then(() => killAllProcesses(), (err) => { killAllProcesses(); throw err; });
}
```

but removed them because they were failing on Travis CI and appVeyor. The process title was set to `node` in those environments. 

This and another test for checking app by name (not index) were working locally on Win10, cygwin, any idea how we can make them work on CI environments ? @filipesilva  @Brocco 

Besides that, it works as described in original issue. Also, for consideration, the original `ng` process name is not tested either.

filipesilva:
@hansl can you take a look please?

alexeagle:
This seems to have gone stale, do we still need this change?
Sorry @tomastrajan

tomastrajan:
Hi @alexeagle ! As you might noticed this is a rather very old PR from December 2017, it was meant to display app name (instead of app index) in the toolbar of the console when running specific app in a multi app scenario. Angular CLI experience quite some evolution since that time so I don't really know if this is still applicable. Might have a look and try to update it if you think it is worth it :)

Issue #3879, opened 2017-01-06
Using raw-loader inline (import "raw!....") leads to compilation error in ng build


Batch1211:
> Please provide us with the following information:
> ---------------------------------------------------------------

### OS?
> Windows 7, 8 or 10. Linux (which distribution). Mac OSX (Yosemite? El Capitan?)
Mac OSX Sierra

### Versions.
> Please run `ng --version`. If there's nothing outputted, please run in a Terminal: `node --version` and paste the result here:
angular-cli: 1.0.0-beta.24
node: 6.7.0
os: darwin x64
@angular/common: 2.4.1
@angular/compiler: 2.4.1
@angular/core: 2.4.1
@angular/forms: 2.4.1
@angular/http: 2.4.1
@angular/platform-browser: 2.4.1
@angular/platform-browser-dynamic: 2.4.1
@angular/router: 3.4.1
@angular/compiler-cli: 2.4.1

### Repro steps.
> Was this an app that wasn't created using the CLI? What change did you do on your code? etc.

Created a component and tried to load an XML File using the raw-loader of webpack:

`import formXml from 'raw!./form.xml'`

### The log given by the failure.
> Normally this include a stack trace and some more information.

`ERROR in ./src/app/frontend/user/login/login.component.ts
Module build failed: Error: /Applications/XAMPP/xamppfiles/htdocs/tekkl/src/app/frontend/user/login/login.component.ts (4,21): Cannot find module 'raw!./form.xml'.)
    at _checkDiagnostics (/Applications/XAMPP/xamppfiles/htdocs/tekkl/node_modules/@ngtools/webpack/src/loader.js:116:15)
    at /Applications/XAMPP/xamppfiles/htdocs/tekkl/node_modules/@ngtools/webpack/src/loader.js:141:17
 @ ./src/app/frontend/user/user.module.ts 12:0-57
 @ ./src async
 @ ./~/@angular/core/src/linker/system_js_ng_module_factory_loader.js
 @ ./~/@angular/core/src/linker.js
 @ ./~/@angular/core/src/core.js
 @ ./~/@angular/core/index.js
 @ ./src/main.ts
 @ multi main`

### Mention any other details that might be useful.

The XML file is loaded without a problem using `ng serve`. It is during the compilation of `ng build` that this error is thrown.

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

filipesilva:
[...]

PR #8824, opened 2017-12-11
Add ability to load any file as raw.

Stoom:
As there has been no progress on the plugin system for webpack this is a good stop gap to load files of "unsupported types."


Stoom:
I signed it!

filipesilva:
What would be the usage of this functionality?

admosity:
There's a workaround for this to invoke raw-loader:

https://github.com/angular/angular-cli/issues/3879#issuecomment-357324730

I have a separate PR that allows custom configuration:

#9191

filipesilva:
I'm sorry but we're not adding support to use arbitrary loaders at this point. For the next major version of the CLI we're adding an extensibility model that allows you to replace the build system.

PR #8842, opened 2017-12-12
docs(@ngtools/webpack): full file name for mainpath option

marmelin:
To prevent others from using mainPath option with file name like in a "module import" (without extension), so they won't get "Cannot read type ... errors".


marmelin:
I signed it!

filipesilva:
Closing as inactive.

Issue #8783, opened 2017-12-07
An asset cannot be read from a location outside the project configuration  1.6.0


eromano:
### Versions
```
1.6.0
```

### Repro steps

* Create a project
* Configure the .angular-cli.json to Read a file from a parent folder

### Observed behavior

Load file from an external folder now fails because of this check : 

An asset cannot be read from a location outside the project

Introduced by the PR:

https://github.com/angular/angular-cli/commit/64c60313d7c972de870cf5af7c52efbe712e8856#diff-9eb0a75e499bdadd44d6dc62218812b8

could be possible introduce a parameter configuration as allowOutsideOutDir but for reading? something like : 

```
 "allowOutsideReadDir": true 
```
It could be false by default and be enabled by the user only when he knows the security risk behind this kind of decision.

Side note the warning message could be improved indicating the assets failing

### Mention any other details that might be useful (optional)
If the idea above are ok for the project I could create a PR

tleveque23:
Was working on 1.6.4, but since the new 1.6.5 it is failing again!

trendzetter:
Not sure if this is related but it is a somewhat similar https://github.com/angular/angular-cli/issues/8613#issuecomment-360139629

dotansimha:
Same here. tried 1.6.4, 1.6.5, 1.6.6, 1.6.7, 1.6.8 :(

Facius:
Same here on 1.70, worked with 1.50

      "assets": [
        {
          "glob": "**/*",
          "input": "./../../fonts/",
          "output": "./assets",
          "allowOutsideOutDir": true
        }
      ]

lazysergey:
any workarounds for this?

alan-agius4:
Closing as this is working as intended. 

Files cannot be read outside of the workspace root directory as there is also a potential security concern. Especially now that deploying the application is possible via a `ng deploy`.

PR #8852, opened 2017-12-12
feat: allowOutsideReadDir configuration

eromano:
Load file from an external folder now fails because of this check :
An asset cannot be read from a location outside the project
Introduced by the PR:
64c6031#diff-9eb0a75e499bdadd44d6dc62218812b8
With allowOutsideOutDir property asset configuration the user can decide to allow this kind of behaviour:

```json
"assets": [
  {
    "glob": "**/*",
    "input": "/temp-folder/outside-allowed/of/project",
    "output": "../not-dist/some/folder/",
    "allowOutsideReadDir": true
  },
]

```
it is false by default and be enabled by the user only when he knows the security risk behind this kind of decision.

See issue: https://github.com/angular/angular-cli/issues/8783


eromano:
Hi guys this PR is needed to our project https://github.com/Alfresco/alfresco-ng2-components in order to update the angular cli. If you want I am happy to illustrate a possible scenario

tleveque23:
Any news on this one? Seems that only some reviews are missing. This bug is really annoying for our project....

bjornharvold:
Interested in how this will get resolved as well. Wouldn't having a mono repo with common assets in an external directory be a good use case for this feature? In my case, I have Ethereum contracts in one project that needs to be compiled and then used by an admin app and a consumer app. All 3 projects are in the same mono repo at the same level.

hansl:
@eromano Could you add a warning on compilation when the configuration option is present (and error out if it's not)?

The reason this check is there is so that people cannot `ng build && deploy` and end up with files they don't want (or aren't aware of) from, e.g., their home directory. There's a way to copy all your SSH keys for example to the assets directory, and someone who doesn't know better will end up with their CD keys available publicly.

This will be all removed in the soon future (next major), but we can make that fix in 1.7.0.

prashu2705:
Hi,

When will this be merged? I'm in need of this change as well.

Thanks.

eromano:
@hansl I have added the warning. I don't know if this PR still make sense or not.
Sorry for the late response but I didn't notice the mention

SpellChucker:
+1 on this one too, we use a shared folder for assets outside of the project and pull them in through `.angular-cli.json`.

eromano:
@hansl should I rebase it another time? or there are no intention to merge it?

Dumpen:
This would be a greatly appreciated feature. We have a lot of shared assets that will benefit from this.

kafumanto:
We are using sibling shared modules and this missing feature is a problem.

agustintarifa:
Any news about this?

alexeagle:
Sorry this got stale @eromano 
Sounds like this change is still needed? Will discuss with @hansl

alexeagle:
We are okay with accepting a PR to read assets from outside the project. But this needs some work to re-base first (likely just make the change again fresh).

alan-agius4:
Thanks for your PR. Though I'll be closing as per @alexeagle comment. Please re-open the PR when the issues are fixed and is in a mergeable state.

PR #8861, opened 2017-12-13
fix(@angular/cli): xi18n --locale option not affecting "source-language" tag

avilv:


ocombe:
Ah sorry, we've submitted at the same time!
Closing this one since mine (https://github.com/angular/angular-cli/pull/8862) has a better fix + tests, but thanks anyway!

PR #8874, opened 2017-12-14
fix(@angular/cli): remove postcss-custom-properties

### LINKED ISSUE #8289: css warning "variable '--some-var' is undefined"
```
- [ ] feature request
```

### Versions.

Angular CLI: 1.5.0
Node: 8.9.0
OS: win32 x64
Angular: 5.0.0

@angular/cdk: 2.0.0-beta.12
@angular/cli: 1.5.0
@angular/flex-layout: 2.0.0-beta.10
@angular/material: 2.0.0-beta.12
@angular-devkit/build-optimizer: 0.0.32
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.35
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.0
@schematics/angular: 0.1.0
typescript: 2.4.2
webpack: 3.8.1

### Repro steps.

add to global `styles.scss`:
```
:root {
  --some-var: blue;
}
```

add to some component `scss`:
```
.some-class {
  background: var(--some-var);
}
```

`ng serve`

```
NonErrorEmittedError: (Emitted value instead of an instance of Error) postcss-custom-properties:
<my-component>.scss:2:3: variable '--some-var' is undefined and used without a fallback
    at Object.emitWarning (<path>\node_modules\webpack\lib\NormalModule.js:117:16)
    at <path>\node_modules\postcss-loader\index.js:131:24
    at Array.forEach (<anonymous>)
    at <path>\node_modules\postcss-loader\index.js:130:31
    at <anonymous>

```

### Desired functionality.

no warning needed, the component is defined in the global `styles.scss` and everything works

one workaround is the create a `_variables.scss` with all the global `:root { --my-var ... }` and include this files in components. This will make to warning go away but the declarations in `_variables.scss` will be included few times

filipesilva:
We currently use a fallback for CSS custom properties on older browsers. This approach has a few problems:
- does not work if the custom property declaration is not part of the same of the same file that uses it (multiple global stylesheets, component css).
- does not work at all for component CSS in AOT.

@clydin suggested a browserlist based approach for enabling this functionality, but that requires a new feature that we don't have.

Since currently taking advantage of the custom property fallback is flaky even in the best case scenario, and potentially broken in prod (AOT), I think it's better to remove it altogether until we can actually do it right.

Fix #8289

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/styles.ts
+++ b/packages/@angular/cli/models/webpack-configs/styles.ts
@@ -11,7 +11,6 @@ import { CleanCssWebpackPlugin } from '../../plugins/cleancss-webpack-plugin';
 const postcssUrl = require('postcss-url');
 const autoprefixer = require('autoprefixer');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
-const customProperties = require('postcss-custom-properties');
 const postcssImports = require('postcss-import');
 
 /**
@@ -115,14 +114,12 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
         }
       ]),
       autoprefixer(),
-      customProperties({ preserve: true })
     ];
   };
   (postcssPluginCreator as any)[postcssArgs] = {
     variableImports: {
       'autoprefixer': 'autoprefixer',
       'postcss-url': 'postcssUrl',
-      'postcss-custom-properties': 'customProperties',
       'postcss-import': 'postcssImports',
     },
     variables: { minimizeCss, baseHref, deployUrl, projectRoot }
--- a/package-lock.json
+++ b/package-lock.json
@@ -6550,40 +6550,6 @@
         "postcss-value-parser": "3.3.0"
       }
     },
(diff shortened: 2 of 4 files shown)


packages/@angular/cli/models/webpack-configs/styles.ts
```diff
@@ -67,7 +66,7 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
               // If deployUrl contains a scheme, ignore baseHref use deployUrl as is.
               return `${deployUrl.replace(/\/$/, '')}${url}`;
             } else if (baseHref.match(/:\/\//)) {
-              // If baseHref contains a scheme, include it as is.
+              // If baseHref contains a scheme, include it as is.git
```
Brocco: Remove random `git` text.

filipesilva: Urf durf, my bad

byrondover:
👏

PR #8905, opened 2017-12-16
refactor(check-port): return found port

realappie:
This will allow the angular cli to simply run on a different port in case the current port is occupied. This is super way more useful than receiving a warning from the `angular-cli` the port is in use.

### DIFF
--- a/packages/@angular/cli/utilities/check-port.ts
+++ b/packages/@angular/cli/utilities/check-port.ts
@@ -1,22 +1,19 @@
 import * as denodeify from 'denodeify';
 
-const SilentError = require('silent-error');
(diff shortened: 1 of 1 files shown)


clydin:
The current behavior is intentional.  If the user specifies a specific port and it is not available, something is wrong.  This is especially important in test and CI scenarios.

Also, the behavior attempting to be achieved here is already available by passing “0” as the desired port.

realappie:
I’m unsure if the cost outweighs the benefits. How many people want the process to exit when the port is not found versus how many want to simply use the cli to serve their application.

I didn’t know about the port 0 flag, that’s useful. But I personally think the default behavior should have been different.

Issue #8434, opened 2017-11-10
AOT build error Cannot read property 'kind' of undefined


samalexander:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.

Angular CLI: 1.5.0
Node: 8.5.0
OS: darwin x64
Angular: 5.0.1
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cdk: 5.0.0-rc0
@angular/cli: 1.5.0
@angular/flex-layout: 2.0.0-beta.10-4905443
@angular/material: 5.0.0-rc0
@angular-devkit/build-optimizer: 0.0.32
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.35
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.0
@schematics/angular: 0.1.2
typescript: 2.4.2
webpack-bundle-analyzer: 2.9.1
webpack: 3.8.1

### Repro steps.

rm -rf node_modules/
yarn add @angular/cli@latest -D
ng build --aot

### The log given by the failure.

`ERROR in Error: TypeError: Cannot read property 'kind' of undefined
    at nodeCanBeDecorated (/Users/sam/Developer/ng-prime/node_modules/typescript/lib/typescript.js:7805:35)
    at nodeIsDecorated (/Users/sam/Developer/ng-prime/node_modules/typescript/lib/typescript.js:7825:16)
    at Object.nodeOrChildIsDecorated (/Users/sam/Developer/ng-prime/node_modules/typescript/lib/typescript.js:7829:16)
    at isDecoratedClassElement (/Users/sam/Developer/ng-prime/node_modules/typescript/lib/typescript.js:51668:23)
    at isInstanceDecoratedClassElement (/Users/sam/Developer/ng-prime/node_modules/typescript/lib/typescript.js:51659:20)
    at Object.filter (/Users/sam/Developer/ng-prime/node_modules/typescript/lib/typescript.js:1652:25)
    at getDecoratedClassElements (/Users/sam/Developer/ng-prime/node_modules/typescript/lib/typescript.js:51641:23)
    at generateClassElementDecorationExpressions (/Users/sam/Developer/ng-prime/node_modules/typescript/lib/typescript.js:51815:27)
[...]

PR #8928, opened 2017-12-18
fix(@ngtools/webpack): allow typescript dev versions

kevinphelps:
This PR fixes the issue described in https://github.com/angular/angular-cli/issues/8434#issuecomment-352554433.

### DIFF
--- a/packages/@ngtools/webpack/src/index.ts
+++ b/packages/@ngtools/webpack/src/index.ts
@@ -1,12 +1,12 @@
 // @ignoreDep typescript
-import { satisfies } from 'semver';
+import { gte } from 'semver';
 
 // Test if typescript is available. This is a hack. We should be using peerDependencies instead
 // but can't until we split global and local packages.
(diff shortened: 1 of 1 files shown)


packages/@ngtools/webpack/src/index.ts
```diff
@@ -5,7 +5,7 @@ import { satisfies } from 'semver';
 // but can't until we split global and local packages.
 // See https://github.com/angular/angular-cli/issues/8107#issuecomment-338185872
 try {
-  const version = require('typescript').version;
+  const version = require('typescript').version.match(/^[0-9]+\.[0-9]+\.[0-9]+/)[0];
   if (!satisfies(version, '^2.0.2')) {
```
clydin: changing the `satisfies` usage to `gte` should also work and avoid the regex

kevinphelps: That appears to be correct.

kevinphelps: PR updated.

Issue #8937, opened 2017-12-19
SVG inlining with postcss-url not compatible with IE11


RoopeHakulinen:
### Versions
Possibly caused by [this commit](https://github.com/angular/angular-cli/commit/def8f6294a2a9af0468d8d0b2e1fff1774b2fcbb#diff-c9c80a591c7b05af6bb6a3ab25ca6a30) introduced in 1.6.0.

### Repro steps

Add an SVG image in CSS with `url('assets/pic.svg')`. This image needs to be small enough (less than 10kB) so that postcss-url does inline it as it is configured so.

### Observed behavior
Image is not shown.

### Desired behavior

Image should be shown.

### Mention any other details that might be useful (optional)
This is because the base64 encoding is not used and thus the image contains spaces in the produced inline SVG inside the `url('...')`. This again does not work in IE11 thus the images are not shown.

clydin:
Does this happen in both development and production build targets?

RoopeHakulinen:
Only for production as I analyzed it further and it is caused by [this commit](https://github.com/angular/angular-cli/commit/392a3935a757945be68dbde0c4bcc8d5b4aa7bd5) instead of the one I referenced initially. There on lines 80 to 87 is this:

clydin:
Those lines run in both production and development.  However, they appear to have exposed a defect in the css optimizer which is only run in production.  Master has some changes to css optimization.  As a result (and we'll run some tests to ensure), this should already be fixed in master and should be in the next minor version.

filipesilva:
I tested with `@angular/cli@1.6.3` using https://s.cdpn.io/3/kiwi.svg (3kb) and `body { background-image: url('./assets/kiwi.svg') }` in `src/styles.css`. Then I ran `npm run ng -- serve --prod` and opened the page in Internet Explorer 11. It seems to be fine now. 

@RoopeHakulinen let me know if you still see this with `1.6.3` and I will reopen.

RoopeHakulinen:
Nope still not working. I'll see if I can make a small project to reproduce.

Here's what happens
Original SVG:
```
<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
  <defs>
    <style>.cls-1{fill:#0051ba;}</style>
  </defs>
  <title>details-hover</title>
  <rect class="cls-1" x="11" y="11" width="2" height="5"/>
[...]

PR #8954, opened 2017-12-21
feat(@angular/cli): add build flag --inline-asset-max-size - Maximum size of asset to be inlined

oburakevych:

```
```


RoopeHakulinen:
This would really be appreciated because of the problems caused by inlining on production builds. These include IE11 problems (#8937) and the security issue (#8945)

clydin:
There does appear utility in this option or something similar to this.  However, please be aware that issue #8937 is a bug and a fix will be in a release much sooner than a new feature.
Also please note that a strict CSP configuration requirement does not equate to a security issue.

oburakevych:
Some merge issues with this branch. Had to create another PR https://github.com/angular/angular-cli/pull/8967

PR #8967, opened 2017-12-22
feat(@angular/cli): add build flag --inline-asset-max-size Maximum size of asset to inline

oburakevych:

```
```

### DIFF
--- a/docs/documentation/build.md
+++ b/docs/documentation/build.md
@@ -73,15 +73,15 @@ dead code elimination via UglifyJS.
 Both `--dev`/`--target=development` and `--prod`/`--target=production` are 'meta' flags, that set other flags.
 If you do not specify either you will get the `--dev` defaults.
 
-Flag                | `--dev` | `--prod`
----                 | ---     | ---
-`--aot`             | `false` | `true`
-`--environment`     | `dev`   | `prod`
-`--output-hashing`  | `media` | `all`
-`--sourcemaps`      | `true`  | `false`
-`--extract-css`     | `false` | `true`
-`--named-chunks`    | `true`  | `false`
-`--build-optimizer` | `false` | `true` with AOT and Angular 5
+Flag                      | `--dev` | `--prod`
+---                       | ---     | ---
+`--aot`                   | `false` | `true`
+`--environment`           | `dev`   | `prod`
+`--output-hashing`        | `media` | `all`
+`--sourcemaps`            | `true`  | `false`
+`--extract-css`           | `false` | `true`
+`--named-chunks`          | `true`  | `false`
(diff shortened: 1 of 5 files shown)


docs/documentation/build.md
```diff
@@ -100,7 +100,11 @@ code.
 ### CSS resources
 
 Resources in CSS, such as images and fonts, will be copied over automatically as part of a build.
-If a resource is less than 10kb it will also be inlined.
+If a resource is less than 10kb it will also be inlined by default.
+
+By using flaf `--inline-asset-max-size` it's possible to define the max size of the
```
blackholegalaxy: typo "flag"

packages/@angular/cli/models/webpack-configs/styles.ts
```diff
@@ -44,39 +44,60 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
...
-            } else {
-              // Join together base-href, deploy-url and the original URL.
-              // Also dedupe multiple slashes into single ones.
-              return `/${baseHref}/${deployUrl}/${url}`.replace(/\/\/+/g, '/');
-            }
-          }
-        },
-        {
-          // TODO: inline .cur if not supporting IE (use browserslist to check)
-          filter: (asset: any) => !asset.hash && !asset.absolutePath.endsWith('.cur'),
```
clydin: Instead of the extra function (which is breaking the eject command).  Add a condition to the filter here.
Also the semantics of the option values should be changed.  A maximum size of zero (0) means nothing should be inlined.  Infinity would mean everything should be inlined.

tests/e2e/tests/build/styles/inline-asset-max-size.ts
```diff
@@ -0,0 +1,51 @@
...
+      `,
+      'src/app/app.component.css': `
+        h3 { background: url('../assets/small.svg'); }
+        h4 { background: url('../assets/large.png'); }
+      `,
+      'src/assets/small.svg': imgSvg,
+      'src/assets/small-id.svg': imgSvg
+    }))
+    .then(() => copyProjectAsset('images/spectrum.png', './assets/large.png'))
+    .then(() => ng('build', '--extract-css', '--inline-asset-max-size=-1', '--aot'))
```
clydin: Tests for inlining all resources and at an arbitrary non-default value would also be needed.

clydin:
Closing this as infrastructure to more easily support this feature has been added since this PR's creation and a major refactoring of the build system in the next major version will provide a mechanism for the addition of features such as this.

filipvh:
@clydin how would i go about this?

PR #8976, opened 2017-12-22
Pipeable

Brocco:


Brocco:
Dependent upon angular/devkit#345

packages/@angular/cli/tasks/schematic-run.ts
```diff
@@ -8,12 +8,11 @@ import {
...
 } from '@angular-devkit/schematics';
 import { FileSystemHost } from '@angular-devkit/schematics/tools';
-import { Observable } from 'rxjs/Observable';
+import { of as observableOf } from 'rxjs/observable/of';
 import * as path from 'path';
 import chalk from 'chalk';
 import { CliConfig } from '../models/config';
-import 'rxjs/add/operator/concatMap';
-import 'rxjs/add/operator/map';
+import { concat, concatMap, ignoreElements, map } from 'rxjs/operators';
```
CaerusKaru: These should be separate imports: 

```
import { concat } from 'rxjs/operators/concat';
...
```

packages/@ngtools/logger/src/console-logger-stack.spec.ts
```diff
@@ -1,13 +1,13 @@
 import {LogEntry, Logger} from './logger';
 import {ConsoleLoggerStack} from './console-logger-stack';
 import {NullLogger} from './null-logger';
+import {toArray} from 'rxjs/operators';
```
CaerusKaru: This should be an import directly from 'rxjs/operators/toArray'

CDDelta:
Maybe blacklist importing the whole rxjs library by adding this to tslint.json `"import-blacklist": [true, "rxjs", "rxjs/operators"]`

trotyl:
@CDDelta Importing from `rxjs/operators` won't cause any bundle size problem in runtime:

image

Let alone this change is for cli tasks only.

CDDelta:
@trotyl I just thought it would be a good idea since that was what @CaerusKaru was pushing for, but you are right this is more of a style thing than a tree-shaking thing.

packages/@ngtools/logger/src/null-logger.ts
```diff
@@ -1,13 +1,10 @@
-import {Logger} from './logger';
-
-import {Observable} from 'rxjs/Observable';
-
-import 'rxjs/add/observable/empty';
+import { EmptyObservable } from 'rxjs/Observable/EmptyObservable';
```
clydin: Lower case 'o' in `rxjs/observable/EmptyObservable`.
Also rxjs v6 appears to be moving towards using creation helpers over class instantiation.  So may want to go with `import { empty } from 'rxjs/observable/empty';`

packages/@ngtools/logger/src/console-logger-stack.spec.ts
Toxicable: no, importing from `'rxjs/operators'` is fine

CaerusKaru: This is to maintain parity with angular/material2, angular/flex-layout, and angular/preboot, starting with https://github.com/angular/material2/pull/8160. It seems like it's an open question though on that thread, and will probably become irrelevant as RxJS 6 comes closer

Toxicable: Importing from operators does cause a larger bundle currently but from memory that's a bug at the moment and should be fixed. With that in mind I think the more ergonomic approach should be the desired kne

clydin: Please note also that this is a node application not a web application.  Bundle size is irrelevant.

CaerusKaru: I think it's more about setting a precedent for best practices, as a lot of developers look to first-party packages like this for guidance (node app or no), especially when it comes to RxJS. Again, for this at least it seems to be a non-starter, the only reason I'm leaving this comment up is to archive the discussion.

hansl: Best practices for RxJS 6 is to `import {toArray} from 'rxjs'`. I'm personally okay with this as it's likely the whole Angular ecosystem will move towards that over the next quarter or two.

Issue #6646, opened 2017-06-12
Apps inheritance


Orhleil:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Versions.
Node: v6.11.0
NPM: 3.10.10
@angular/cli: 1.1.1
Windows 7 Pro SP1 64bits

### Repro steps.
N/A

### The log given by the failure.
N/A

### Desired functionality.

Hello everyone,

I am currently working on a project that uses the apps feature. My feature suggestion would be to add a way to make an app inherit from another in order to avoid repeating the same configuration countless times when there is just one variation per app.

For instance, I've got several apps that only differ from each other on the assets. As a result, I've got an immense .angular-cli.json file that looks like this:
```javascript
{
  (.......),
  "apps": [
    {
      "name": "app1",
      "root": "src",
      "outDir": "dist",
      "index": "index.html",
      "main": "main.ts",
      "polyfills": "polyfills.ts",
      "test": "test.ts",
      "tsconfig": "../conf/tsconfig.app.json",
      "testTsconfig": "../conf/tsconfig.spec.json",
      "prefix": "app",
      "styles": [
        "../node_modules/some_package1/style.min.css",
        "../node_modules/some_package1/style.min.css",
        "styles.scss"
      ],
      "scripts": [
        "../node_modules/some_package1/script.min.js",
        "../node_modules/some_package2/script.min.js",
        "../node_modules/some_package3/script.min.js",
      ],
      "environmentSource": "environments/environment.ts",
      "environments": {
        "dev": "environments/environment.dev.ts"
        "ci": "environments/environment.ci.ts"
        "prod": "environments/environment.prod.ts"
      },
      "assets": [
        "assets",
        "favicon.ico",
        "specific-asset-app1"
      ]
    },
    {
      "name": "app2",
      "root": "src",
      "outDir": "dist",
      "index": "index.html",
      "main": "main.ts",
      "polyfills": "polyfills.ts",
      "test": "test.ts",
      "tsconfig": "../conf/tsconfig.app.json",
      "testTsconfig": "../conf/tsconfig.spec.json",
      "prefix": "app",
      "styles": [
        "../node_modules/some_package1/style.min.css",
        "../node_modules/some_package1/style.min.css",
        "styles.scss"
      ],
      "scripts": [
        "../node_modules/some_package1/script.min.js",
        "../node_modules/some_package2/script.min.js",
[...]

PR #8993, opened 2017-12-26
feat(@angular/cli): add extends app from another apps by name

KorsaR-ZN:
This feature let you use inheritance of the application configuration.
```
"apps": [
  {
    "name": "base",
    "root": "src",
    "test": "test.ts",
    "tsconfig": "tsconfig.app.json",
    "testTsconfig": "tsconfig.spec.json",
    ...
  },
  {
    "name": "app1",
    "extends": "base",
    "outDir": "dist-app1",
    "prefix": "app1",
    "assets": [
      "assets/app1"
    ]
  },
  {
    "name": "app2",
    "extends": "base",
    "outDir": "dist-app2",
    "prefix": "app2",
    "assets": [
      "assets/app2"
    ]
  },
  ....
],
```
see #6646

### DIFF
--- a/docs/documentation/angular-cli.md
+++ b/docs/documentation/angular-cli.md
@@ -11,6 +11,7 @@
 
 - **apps** (`array`): Properties of the different applications in this project.
   - *name* (`string`): Name of the app.
+  - *extends* (`string|array`): Can be used to extend from another apps by name.
   - *root* (`string`): The root directory of the app.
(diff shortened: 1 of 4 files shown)


filipesilva:
Heya, we're changing the way we handle project configuration in https://github.com/angular/devkit/pull/366. It should take care of these cases.

PR #9015, opened 2017-12-28
docs(README): update `skipCodeGeneration` section

alan-agius4:
Seems like it's still widely used in https://github.com/angular/angular-cli/blob/master/packages/%40ngtools/webpack/src/angular_compiler_plugin.ts


alan-agius4:
Apologies about the commit msg, fixed it!

alan-agius4:
CI is green now :)

filipesilva:
Thanks for updating this 👍

PR #7619, opened 2017-09-07
fix(@ngtools/webpack): fix #7615, check if this.done is defined before delaying resolvers

rezonant:
Resolves #7615

PR #9043, opened 2017-12-30
fix(@ngtools/webpack): redo #7619 in angular_compiler_plugin

nename0:
The issue fixed in #7619, is still in [`angular_compiler_plugin.ts`](https://github.com/angular/angular-cli/blob/5eea7faf5a320e5f18c7dbef79105046819e32ea/packages/%40ngtools/webpack/src/angular_compiler_plugin.ts#L604) 
With my current webpack production setup I keep hitting this problem, that `this.done` is not defined. 
As a quick fix I was able to just edit the file in the node_modules. However as a permanent solution I would like this PR to be merged.


filipesilva:
Thanks for porting this fix over, should be in the next release.

Issue #7968, opened 2017-10-08
ng generate component --module doesn't support module names properly


shadowmint:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report
- [ ] feature request
```

### Versions.
```
$ ng version
    _                      _                 ____ _     ___
   / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
  / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
 / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
/_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
               |___/
@angular/cli: 1.4.5
node: 8.2.1
os: win32 x64
@angular/animations: 4.4.4
@angular/cli: 1.4.5
@angular/common: 4.4.4
@angular/compiler: 4.4.4
@angular/core: 4.4.4
@angular/forms: 4.4.4
@angular/http: 4.4.4
@angular/platform-browser: 4.4.4
@angular/platform-browser-dynamic: 4.4.4
@angular/router: 4.4.4
@angular/compiler-cli: 4.4.4
@angular/language-service: 4.4.4
typescript: 2.3.4
```

### Repro steps.
```
$ ng new
...

$ ng generate module TestMod
  create src/app/test-mod/test-mod.module.ts (191 bytes)

$ ng generate component TestComp --module test-mod.module.ts
Error: Specified module does not exist
Specified module does not exist
```

### The log given by the failure.
```
Error: Specified module does not exist
```

### Desired functionality.
Support a meaningful parameter value; either a path to a file, or a module name.
ie. One of:

```
ng generate component Foo --module FooBar
ng generate component Foo --module ./src/app/foo-bar/foo-bar.module.ts
```

### Mention any other details that might be useful.
Previously this worked (1.3.x): `ng generate component Foo --module Bar`, (although `ng generate component Foo --module FooBar` never did).

However, the wiki (https://github.com/angular/angular-cli/wiki/generate-component) now suggests:

```
ng generate component Foo --module bar.module.ts
```

...but that doesn't work either.

By adding a console log to `node_modules/@schematics/angular/utility/find-module.js` to log the base path, you can see it failing to find the correct path:

```
$  ng generate component FooBarOne --module FooBar
Attempting to resolve using base path: /src/app/FooBar
Error: Specified module does not exist
Specified module does not exist

[...]

PR #9117, opened 2018-01-07
specify where angular-cli reads module name from

Vandivier:
Adding this to the docs would be helpful.

Background:
I renamed the folder containing an initialized module from `app` to `client`. Then I attempted to `ng generate service --module client` I received an error that the module doesn't exist.

So I renamed all the files in the folder. `./client/app.module.ts` -> `./client/client.modules.ts`. Same issue. So I renamed the variables within each file. `AppRoutingModule` -> `ClientRoutingModule`, etc. Issue is still there.

Finally I realized that the name was coming from .angular-cli.json. Let's spare some other folks the time.

### DIFF
--- a/docs/documentation/generate.md
+++ b/docs/documentation/generate.md
@@ -5,6 +5,8 @@
 ## Overview
 `ng generate [name]` generates the specified blueprint
 
+`[name]` is read from the prefix property of an object in the apps array within .angular-cli.json
+
 ## Available blueprints:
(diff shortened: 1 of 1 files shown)


Vandivier:
CLA signed.

Vandivier:
Please cancel. The PRd change is not accurate. See issue [7968](https://github.com/angular/angular-cli/issues/7968).

Issue #9100, opened 2018-01-05
@ngtools/webpack breaks when project has # character on path


dealloc:
### Versions
```
$ node --version
v8.7.0
$ npm --version
5.4.2
$ yarn --version
1.3.2

Running Windows 10 Pro x64
Shell: Git bash
```

### Repro steps

* Clone https://github.com/dealloc/ngtools-webpack-reproduction
* Install dependencies using `yarn install`
* Start application using `yarn start`

### Observed behavior
You will get a blank page with "Loading..." and in the console the following error:
```
Uncaught Error: No NgModule metadata found for 'AppModule'.
    at NgModuleResolver.webpackJsonp../node_modules/@angular/compiler/esm5/compiler.js.NgModuleResolver.resolve (compiler.js:20233)
    at CompileMetadataResolver.webpackJsonp../node_modules/@angular/compiler/esm5/compiler.js.CompileMetadataResolver.getNgModuleMetadata (compiler.js:15186)
    at JitCompiler.webpackJsonp../node_modules/@angular/compiler/esm5/compiler.js.JitCompiler._loadModules (compiler.js:34255)
    at JitCompiler.webpackJsonp../node_modules/@angular/compiler/esm5/compiler.js.JitCompiler._compileModuleAndComponents (compiler.js:34216)
    at JitCompiler.webpackJsonp../node_modules/@angular/compiler/esm5/compiler.js.JitCompiler.compileModuleAsync (compiler.js:34110)
    at CompilerImpl.webpackJsonp../node_modules/@angular/platform-browser-dynamic/esm5/platform-browser-dynamic.js.CompilerImpl.compileModuleAsync (platform-browser-dynamic.js:230)
    at PlatformRef.webpackJsonp../node_modules/@angular/core/esm5/core.js.PlatformRef.bootstrapModule (core.js:5568)
    at Object../src/main.ts (main.ts:11)
    at __webpack_require__ (bootstrap 1550deb7735f43aaf874:712)
    at fn (bootstrap 1550deb7735f43aaf874:117)
    at Object.2 (main.ts:11)
    at __webpack_require__ (bootstrap 1550deb7735f43aaf874:712)
    at webpackJsonpCallback (bootstrap 1550deb7735f43aaf874:25)
    at app.js:1
```

### Desired behavior

The application should run properly and display the text "Hello world!" in a `<h1>` tag.

### Mention any other details that might be useful (optional)
[...]

PR #9140, opened 2018-01-09
fix(@ngtools/webpack): allow # in paths

dealloc:
Possible fix for #9100 by only splitting on # when it's followed by a valid classname


dealloc:
I signed it!

Issue #7625, opened 2017-09-08
Angular CLI compiler requires older Typescript


jtsom:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.
@angular/cli: 1.4.0
node: 8.4.0
os: darwin x64
@angular/animations: 4.3.6
@angular/cdk: 2.0.0-beta.10
@angular/common: 4.3.6
@angular/compiler: 4.3.6
@angular/core: 4.3.6
@angular/forms: 4.3.6
@angular/http: 4.3.6
@angular/material: 2.0.0-beta.10
@angular/platform-browser: 4.3.6
@angular/platform-browser-dynamic: 4.3.6
@angular/router: 4.3.6
@angular/cli: 1.4.0
@angular/compiler-cli: 4.3.6
@angular/language-service: 4.3.6
typescript: 2.5.2

### Repro steps.
Start up app with `ng serve` and this message is displayed:

```
@angular/compiler-cli@4.3.6 requires typescript@'>=2.1.0 <2.4.0' but 2.5.2 was found instead.
Using this version can result in undefined behaviour and difficult to debug problems.

Please run the following command to install a compatible version of TypeScript.

    npm install typescript@'>=2.1.0 <2.4.0'

To disable this warning run "ng set --global warnings.typescriptMismatch=false".
```

kuncevic:
Based on that [commit ](https://github.com/angular/angular-cli/commit/57d7246 )it should even allow 2.5 but I have same issue even with `2.4.2`

karser:
Newly created project with `angular-cli@1.4.0` requires `typescript@~2.3.3`

kondi:
Is this requirement valid? My project looks fine with typescript@2.4.2 and I was planning to go for 2.5.x but now I am not sure if it is a good idea.

kondi:
Anyway it would be good if @angular/compiler-cli defines the correct peer dependency here: https://github.com/angular/angular/blob/4.4.x/packages/compiler-cli/package.json#L17. Not everybody is using angular-cli. I can understand the safety check, but I am not sure the matrix linked by kuncevic should be the correct place to specify the requirements.

jtsom:
There were no indications that this restriction would be implemented. The only thing I updated in my project was the CLI. The @angular/compiler-cli@4.3.6 has no problems previously.

filipesilva:
[...]

PR #9159, opened 2018-01-10
Irritating warning when setting missingTranslation in AotPlugin options

mhoyer:
When using `missingTranslation` in options for `AotPlugin` constructor the plugin will print a confusing warning when angular (and CLI) is already at version 5 or higher:

    The --missing-translation parameter will be ignored because is only with Angular version 4.2.0 or higher.
    If you want to use it, please your Angular version.

Example in `webpack.config.js`:

```javascript
    ...
    const aotOptions = {
        tsConfigPath: './tsconfig.json',
        entryModule: ...
        exclude: [...],
        locale: 'en',
        i18nFile: 'messages.en.xlf',
        i18nFormat: 'xlf',
        missingTranslation: 'warning', // <<< this will produce the warning
    }
    new AotPlugin(aotOptions);
```

### DIFF
--- a/packages/@ngtools/webpack/src/plugin.ts
+++ b/packages/@ngtools/webpack/src/plugin.ts
@@ -295,7 +295,7 @@ export class AotPlugin implements Tapable {
     }
     if (options.hasOwnProperty('missingTranslation')) {
       const [MAJOR, MINOR, PATCH] = VERSION.full.split('.').map((x: string) => parseInt(x, 10));
-      if (MAJOR < 4 || (MINOR == 2 && PATCH < 2)) {
+      if (MAJOR < 4 || (MAJOR == 4 && (MINOR < 2 || (MINOR == 2 && PATCH < 2)))) {
         console.warn((`The --missing-translation parameter will be ignored because it is only `
           + `compatible with Angular version 4.2.0 or higher. If you want to use it, please `
           + `upgrade your Angular version.\n`));


clydin:
It shouldn’t be showing the warning.  However, AngularCompilerPlugin should be used for Angular 5+.  AotPlugin is intended for Angular 2 & 4.

mhoyer:
@clydin you are right. I missed that line in the `README.md`:

> Angular version 5 and up, use AngularCompilerPlugin:

Hence, the whole `AotPlugin` maybe should fail to initialize in general - in case the underlying version of `@angular/compiler-cli` already is 5.0 or higher? Or does is there any use case where it makes sense for someone to keep using `AotPlugin` together with ng5+?

And was it on purpose to keep both versions (`AotPlugin` and new `AngularCompilerPlugin`) in one version branch to be released and published side-by-side? E. g. defining clear `peerDependencies` to a required `@angular/compiler-cli` would make more sense IMHO.

In general it feels strange to [have in-code verification for particular version number](https://github.com/angular/angular-cli/blob/master/packages/%40ngtools/webpack/src/ngtools_api.ts#L153) of peer installed dependencies when there is actually infrastructure for dependency/version management through NPMs `package.json` available.

clydin:
The package is intended to provide support for Angular 2 through 5 (and eventually beyond).  The CLI also supports building Angular 2, 4, and 5 based projects.  So both are required.

filipesilva:
@mhoyer no, there isn't any usecase for using `AotPlugin` with Angular v5+. 

At the moment we can't set peer dependencies on `@angular/compiler-cli` because we use the same global and local CLI package, and that complicates things (see https://github.com/angular/angular-cli/issues/7625#issuecomment-333497314). That's why we have in code verification of versions.
We're moving to separate this though. 

I agree we could have some error being thrown in `AoTPlugin` though. Would you be interested in adding a PR for that?

clydin:
Closing as the package has been moved to the devkit project and Angular 5 is now the minimum version.

PR #9168, opened 2018-01-11
fix(@angular/cli): only force CommonJS if module kind is ES2015

### LINKED ISSUE #9158: 1.6.2+ breaks when using import('module') syntax
### Versions
```
Angular CLI: 1.6.3
Node: 8.7.0
OS: darwin x64
Angular: 5.2.0

@angular/cli: 1.6.3
@angular-devkit/build-optimizer: 0.0.36
@angular-devkit/core: 0.0.22
@angular-devkit/schematics: 0.0.42
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.3
@schematics/angular: 0.1.11
@schematics/schematics: 0.0.11
typescript: 2.4.2
webpack: 3.10.0
```

### Repro steps

* Clone repo https://github.com/devoto13/esnext-repro
* `ng serve --aot`

### Observed behavior

Build fails and TS error is reported:

> ERROR in src/app/app.component.ts(12,5): error TS1323: Dynamic import cannot be used when targeting ECMAScript 2015 modules.

### Desired behavior

No error is reported.

This is a regression in 1.6.2. It worked correctly in 1.6.1. I managed to isolate it to this commit: https://github.com/angular/angular-cli/commit/09177d1.

This commit ensured that `forceTsCommonjs` works for AngularCompilerPlugin, but changing `esnext` module format to `commonjs` leads to the above error. Note this only happens with `target=es2015` because `target=es5` treats `import()` as a normal function call (compiler just passes it through) and than it is picked up by Webpack and treated as ESNext import and everything works because of happy coincidence.

But I really think that CLI should not attempt to replace `esnext` with `commonjs` as it is not a drop in replacement as indicated by the error above.

clydin:
Fixes #9158

### DIFF
--- a/packages/@angular/cli/models/webpack-config.ts
+++ b/packages/@angular/cli/models/webpack-config.ts
@@ -48,6 +48,11 @@ export class NgCliWebpackConfig<T extends BuildOptions = BuildOptions> {
     const supportES2015 = tsConfig.options.target !== projectTs.ScriptTarget.ES3
                         && tsConfig.options.target !== projectTs.ScriptTarget.ES5;
 
+    // Only force if the default module kind is used (ES2015)
+    // Allows user to adjust module kind (to esNext for example)
+    buildOptions.forceTsCommonjs = buildOptions.forceTsCommonjs
+                                 && tsConfig.options.module === projectTs.ModuleKind.ES2015;
+
     this.wco = { projectRoot, buildOptions, appConfig, tsConfig, supportES2015 };
   }
 


packages/@angular/cli/models/webpack-configs/typescript.ts
```diff
@@ -7,6 +7,8 @@ import {
   AngularCompilerPluginOptions,
   PLATFORM
 } from '@ngtools/webpack';
+// @ignoreDep typescript - used only for type information
+import * as ts from 'typescript';
```
devoto13: The usage below does not look like using only for type information. Looks as much value as it can get :)

    options.compilerOptions.module = ts.ModuleKind.CommonJS;

PR #9173, opened 2018-01-12
refactor: add another easter egg

hansl:


packages/@angular/cli/commands/easter-egg.ts
```diff
@@ -32,4 +32,34 @@ const MakeThisAwesomeCommand = Command.extend({
   }
 });
 
-export default MakeThisAwesomeCommand;
+export MakeThisAwesomeCommand;
+
+export const HoustonCommand = Command.extend({
+  name: 'make-this-awesome',
```
aciccarello: should this be `houston`?

Brocco:
At this point adding new easter eggs is not on the priority list. Maybe in the future.

Issue #9051, opened 2018-01-01
@ngtools/webpack Conflict between DefinePlugin and AngularCompilerPlugin


daniel-nagy:
Variables defined by the Webpack `DefinePlugin` are not available to the config object of the `RouterModule`.

### Versions
```
@angular/compiler: 5.1.2
@angular/compiler-cli: 5.1.2
```

### Repro steps

```javascript
// webpack config
new DefinePlugin({
  PRODUCTION: Boolean(env.production)
}),
```

```typescript
// app module
const router = RouterModule.forRoot([...], {
  enableTracing: !PRODUCTION
});

@NgModule({
  imports: [
    router
  ]
})
export class AppModule {

}
```

### Observed behavior
The value of `PRODUCTION` as defined by the `DefinePlugin` is not used and `enableTracing` is always true.

### Desired behavior
I should be able to configure the `RouterModule` using variables defined by the `DefinePlugin`.

### Mention any other details that might be useful (optional)

The following [test.zip](https://github.com/angular/angular-cli/files/1595616/test.zip) file contains a minimal project to reproduce the issue. Download the zip file then run

```
npm install
npm start
```

Open a browser to localhost:3000 and notice that the router events are always logged to the console regardless of the value of `PRODUCTION` as defined in the webpack.config.js file.

Toxicable:
how about using `env.production` instead?
imported from `environment.ts`
more here https://github.com/angular/angular-cli/wiki/stories-application-environments

daniel-nagy:
I’m not using the Angular CLI. Will the AngularCompiler Wepack plugin look for an .angular-cli.json file?

The environment file approach won’t allow for code elimination when minified, unlike the DefinePlugin.

vagonpidarasov:
This is only happens when ENV variables are used in decorators' metadata.

dgp1130:
Duplicate of #9051.

atamansv:
> Duplicate of #9051.

Duplicates itself?

dgp1130:
Apologies, should be duplicate of #9874. Must have copy-pasted from the wrong tab.

vtulin:
Subscribe.
I pass env variables in my webpack to server and Angular AOT client.
Everything is fine on server, but client env are empty {}.

Issue #1656, opened 2016-08-12
[Enhancement] Overridable webpack config


nikolasleblanc:
### Angular-cli@webpack is awesome ###

With webpack on board, angular-cli is looking like a seriously viable toolset for developing and building production ready applications.

### And a default webpack config is a life-saver ###

Webpack's not exactly a breeze. Having default webpack configs ready to go the second a developer creates a project saves them a lot of time and energy.

### But webpack needs to be configurable ###

Once that project's spun up and begins to take form, that developer may want to tweak the webpack config. They may want to add _this_ loader, and _that_ loader, and they may have opinions about how it should be bundled for production.

### You can't please everybody ###

To bloat the base webpack configs, one PR after another, with each individual's particular project need is unsustainable.

### What then? ###

When a developer needs something that the default won't be able to provide, they might be inclined to break off on their own, leaving the benefits of angular-cli aside.

Why not allow the ease of angular-cli to coexist with the configurability of webpack?

#### `webpack-config.js` ####

`webpack-config.js` looks for the following config files, based on the current environment.

- weback-build-_common_.ts
- webpack-build-_development_.ts
- webpack-build-_production_.ts
- webpack-build-_mobile_.ts

Using `angular-cli.json`, the developer could point to environment-specific webpack configuration overrides.

#### `angular-cli.json` ####

`angular-cli.json` is already storing paths for karma configs, protractor configs, and tsconfig, why not let it also store webpack configs? 

```
{
  ...,
  "e2e": {
    "protractor": {
      "config": "config/protractor.conf.js"
    }
  },
  "test": {
    "karma": {
      "config": "config/karma.conf.js"
    }
  },
  "webpack": {
    "development": {
      "config": "config/webpack.dev.conf.js"
    },
    ...
}  
```

If `webpack-config.js` then checked `angular-cli.json` to see:

  - if an override path has been provided, and
  - if a file exists at that path

...it could load the override config instead of the default config.

### Summary ###

This solution offers the developer the option of using default webpack configs to get rolling, as well as the ability to override those defaults once their project begins to show unique requirements, with the smallest possible footprint on the base library.

It also allows the developer to stay in sync with the angular-cli library without having to _eject_ from it, and allows the default configs to continue to stay relevant.

nikolasleblanc:
https://github.com/nikolasleblanc/angular-cli/tree/Fe-Webpack-Overrides

[...]

PR #9191, opened 2018-01-12
feat(@angular/cli): user defined webpack configuration per build target

### LINKED ISSUE #3879: Using raw-loader inline (import "raw!....") leads to compilation error in ng build

admosity:
Fixes #3879 

Similar request here #8824

This PR provides developers a way to add custom webpack configuration to their development and production build targets on a per app basis. The custom webpack configuration is loaded if the user provides the paths to their custom configurations.

We have a usecase where we are automatically generating documentation from various file formats and we needed a way to load files without relying on any backend or heavy frontend rendering. We also did not want to eject from @angular/cli as the generated webpack file is quite unwieldy as well and we lose all the generator benefits of @angular/cli.

Sample repo at:
https://github.com/admosity/angular-cli-custom-webpack

The configuration is added after `getTargetConfig` so these configurations are merged after that.

.angular-cli.json schema is updated with an `extraWebpack` property under an app and accepts `development` and `production` paths to custom webpack files. I believe this functionality should be provided as experimental as it is unknown what the behavior may be when combined with @angular/cli defined webpack configurations. 

From the sample repo:
```json
      "extraWebpack": {
        "production": "../webpack-custom.js",
        "development": "../webpack-custom.js"
      }
```

Below is a sample file that can be loaded (follows webpack config schema). In this usecase, I am loading prerendered markdown to be loaded directly into the app.
```javascript

module.exports = {
  module: {
    rules: [
      {
        test: /\.md$/,
        use: [
          {
            loader: 'raw-loader',
          },
          {
            loader: 'markdown-loader'
          }
        ]
      },
    ]
  },
  plugins: [

  ]
 };
```

I was able to run this successfully, but after fast forwarding to the latest 1.7.0-beta.0 on master, I've been getting: 
`TypeError: querystring.unescape is not a function` being thrown from various files. I think this is due to ambiguity between loading the querystring library or loading node's querystring library.

### DIFF
--- a/packages/@angular/cli/lib/config/schema.json
+++ b/packages/@angular/cli/lib/config/schema.json
@@ -216,6 +216,20 @@
             "description": "Name and corresponding file for environment config.",
             "type": "object",
             "additionalProperties": true
+          },
+          "extraWebpack": {
+            "description": "Experimental support for custom webpack configuration per build target in commonjs format. Paths will be resolved to project root.",
+            "type": "object",
+            "properties": {
+              "development": {
+                "description": "Path to development webpack configuration file.",
+                "type": "string"
+              },
+              "production": {
+                "description": "Path to production webpack configuration file.",
+                "type": "string"
+              }
+            }
           }
         },
         "additionalProperties": false
--- a/packages/@angular/cli/models/webpack-config.ts
+++ b/packages/@angular/cli/models/webpack-config.ts
@@ -1,9 +1,11 @@
 // @ignoreDep typescript - used only for type information
 import * as ts from 'typescript';
+import * as fs from 'fs';
 import { AngularCompilerPlugin } from '@ngtools/webpack';
 import { readTsconfig } from '../utilities/read-tsconfig';
 import { requireProjectModule } from '../utilities/require-project-module';
 const webpackMerge = require('webpack-merge');
+const SilentError = require('silent-error');
 import { CliConfig } from './config';
 import { BuildOptions } from './build-options';
 import {
@@ -17,6 +19,7 @@ import {
   getAotConfig
 } from './webpack-configs';
 import * as path from 'path';
+import { oneLine } from 'common-tags';
 
 export interface WebpackConfigOptions<T extends BuildOptions = BuildOptions> {
   projectRoot: string;
@@ -59,7 +62,8 @@ export class NgCliWebpackConfig<T extends BuildOptions = BuildOptions> {
       getCommonConfig(this.wco),
       platformConfig,
       getStylesConfig(this.wco),
-      this.getTargetConfig(this.wco)
+      this.getTargetConfig(this.wco),
+      this.getTargetConfigCustom(this.wco)
     ];
 
     if (this.wco.appConfig.main || this.wco.appConfig.polyfills) {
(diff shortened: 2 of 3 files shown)


Toxicable:
Doing this goes against the reason for the cli existing.
With this change it would be nearly impossible to swap out webpack for another build system such as parcel or bazel, therefore this should not be merged

admosity:
I don't think this goes against the reason for the cli existing. This would be similar to Ionic's method of offering controlled entrypoints into the build system for advanced configuration: 
https://github.com/ionic-team/ionic-app-scripts. Users can still reap the benefits of the cli while having advanced configuration.

I think this offers a good midway point for users that want to add advanced webpack configuration without having to abandon the cli completely. I think this may be useful for users that are looking to add simple plugins to the setup as well (#9051) while still staying on the cli setup.

It also looks like the approach for supporting bazel is vastly different than the webpack based offering of the cli:
https://github.com/angular/angular-cli/compare/bazel

admosity:
Prior discussion over at #1656

Stoom:
@admosity  the problem is there is still no plugin system to the cli.  People also want to have custom loaders on the webpack.  Are we really going to let fear of a different build system potentially coming into the cli at some time in the future which has no timeline?  If there was a plugin system then great, but there isn't.  Why not add this, mark as experimental and fix it when a better solution comes out?(the agile way)  The argument about the new comer having to know webpack isn't really valid from the issue you've posted about either.  The cli will cover 90% of the use case and an extended webpack will cover the last 10%.

hansl:
We are working on a solution that will solve using a custom webpack configuration, without requiring some angular-cli configuration hacks.

> the problem is there is still no plugin system to the cli.
Engineering is a slow process. Forcing design decisions based on short term needs against long term gains is not the proper solution.

> The cli will cover 90% of the use case and an extended webpack will cover the last 10%.
We've had this discussion many times before, both internally and externally. Our (the Angular team) take on the topic is that if you're part of the 10% (or whatever actual number) of users who need a custom webpack config, there's the eject button. You can still use almost every commands once you're ejected, and instead of `ng build` you can then use `webpack`.

vadjs:
@hansl It's 1.5 years since angular-cli team said "we will add Webpack customization later".
Sure, this problem has very low priority for angular-cli team. But **any solution** is better than **no solution**. Sure, we have `ng eject` option but if you do that it will be painful to update project to new webpack version (as it was in update from webpack 3 to webpack 4).

I have idea how to add minimal customization functionality to angular-cli projects. It's adding `addRules` option to `BuildOptions` in `webpack-configs`. It will cover most of cases when users want to add/overwrite Webpack rules. It's non-ideal but working solution. With this solution you can forget about "your best solution" for next 1.5 years.
@filipesilva If Angular team approve described solution i could realize it. I just need the approvement of this idea.

PR #9215, opened 2018-01-15
fix(@angular/cli): Fixed lint not excluding files when project contained a file or a folder starting with a dot

britvik:


britvik:
I signed it!

PR #9256, opened 2018-01-17
fix(@ngtools/webpack): Allow for usage of Angular v6.*.*

Brocco:


packages/@ngtools/webpack/src/ngtools_api.ts
```diff
@@ -144,8 +144,9 @@ export function CompilerCliIsSupported() {
       + 'Please clean your node_modules and reinstall.');
   }
 
-  // Throw if we're neither 2.3.1 or more, nor 4.x.y, nor 5.x.y.
-  if (!(version.major == '5'
+  // Throw if we're neither 2.3.1 or more, nor 4.x.y, nor 5.x.y, nor 6.x.y.
+  if (!(version.major == '6'
```
Toxicable: Might be a reasonable idea to refactor this into a object like the TS Version checker

Brocco: I agree with you, but this is a quick fix for now.

PR #9299, opened 2018-01-19
fix(@angular/cli): Remove testing log message.

Brocco:


filipesilva:
I feel like we're getting less value out of each eject now though.

PR #9300, opened 2018-01-19
fix(@angular/cli): Allow service-workers package to be installed in parent node_modules

### LINKED ISSUE #8300: serviceWorker = true fails to find @angular/service-worker with yarn workspaces
```
- [ ] feature request
```

### Versions.

Angular CLI 1.5.0
Node: 6.10.3

### Repro steps.

Use yarn workspaces, set serviceWorker = true in angular-cli.json.

Your project is configured with serviceWorker = true, but @angular/service-worker
is not installed. Run `npm install --save-dev @angular/service-worker`
and try again, or run `ng set apps.0.serviceWorker=false` in your .angular-cli.json.

### Desired functionality.

The error shouldn't be fatal for an undetected service worker module. It should probably be a warning and able to be disabled similar to the TypeScript warnings for incorrect versions.

bgotink:
The checks on existence of @angular/service-worker used the node_modules relative to the
project root directly, but those don't exist when using yarn workspaces or lerna. Use
resolve.sync instead, which looks up the tree to parent node_modules folders.

A similar change is required when loading the service worker script from the
@angular/service-worker package. Switch to require.resolve here as well.

Fixes #8300

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/production.ts
+++ b/packages/@angular/cli/models/webpack-configs/production.ts
@@ -9,6 +9,7 @@ import { BundleBudgetPlugin } from '../../plugins/bundle-budget';
 import { StaticAssetPlugin } from '../../plugins/static-asset';
 import { GlobCopyWebpackPlugin } from '../../plugins/glob-copy-webpack-plugin';
 import { WebpackConfigOptions } from '../webpack-config';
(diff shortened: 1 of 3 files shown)


filipesilva:
@alxhub can you review please?

Issue #9229, opened 2018-01-16
bootstrap-local.js is replacing installed NodeJS built-in modules with old versions


paisanousa:
### Versions

```
:~$ node -v
v9.3.0
:~$ npm -v
5.6.0
:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.3 LTS
Release:	16.04
Codename:	xenial

@angular/cli version (following clone/install of @angular/cli detailed in Repro steps):
:~/demo/myapp (master)$ ng -v

    _                      _                 ____ _     ___
   / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
  / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
 / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
/_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
               |___/
    
Angular CLI: local (v1.7.0-beta.0, branch: master)
Node: 9.3.0
OS: linux x64
Angular: 5.2.0
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 1.7.0-beta.0
@angular-devkit/build-optimizer: 0.0.38
@angular-devkit/core: 0.0.25
@angular-devkit/schematics: 0.0.48
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.10.0-beta.0
@schematics/angular: 0.1.13
@schematics/schematics: 0.0.13
typescript: 2.5.3
webpack: 3.10.0
```
### Repro steps

```
Beginning list of global installs:
$ npm list --depth=0 -g
/home/dennis/.nvm/versions/node/v9.3.0/lib
└── npm@5.6.0
```
```
* Step 1 - Clone and globally symlink the angular/angular-cli github repository:

Using the documentation provided at:
https://github.com/angular/angular-cli#development-hints-for-working-on-angular-cli

Clone and link the angular/angular-cli repository:

:~$ git clone https://github.com/angular/angular-cli.git
Cloning into 'angular-cli'...
remote: Counting objects: 26358, done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 26358 (delta 2), reused 2 (delta 0), pack-reused 26350
[...]

PR #9302, opened 2018-01-20
fix(@angular/cli): exclude nodejs built-ins from process.cwd() resolve

paisanousa:
Subsequent to commit dd9e69e, bootstrap-local.js attempts to resolve packages using process.cwd() when @angular/cli is linked locally. The nodejs built-in modules are among the packages being evaluated. Unfortunately, if a built-in module is matched in the local node_modules structure, the replacement is more likely to be an older, or customized, version of that module. Issue #9229 documents the type of error that can be caused by this replacement process.

This pull request refines the replacement logic by excluding nodejs built-ins from process.cwd() resolves during a local bootstrap.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     "@types/source-map": "^0.5.0",
     "@types/webpack": "^3.0.5",
     "@types/webpack-sources": "^0.1.3",
+    "builtin-modules": "2.0.0",
     "conventional-changelog": "1.1.0",
     "dtsgenerator": "^0.9.1",
     "eslint": "^3.11.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "@angular-devkit/schematics": "~0.0.51",
     "@schematics/angular": "~0.1.16",
     "autoprefixer": "^7.2.3",
+    "builtin-modules": "2.0.0",
```
clydin: This should be a dev dependency

paisanousa: Thanks @clydin.  I agree and will make the change.

paisanousa:
@clydin, requested change squash'ed under original commit and pushed.

PR #9321, opened 2018-01-22
feat(@angular/cli): Add update cmd to update angular versions.

Brocco:


packages/@angular/cli/commands/update.ts
```diff
@@ -0,0 +1,48 @@
+const Command = require('../ember-cli/lib/models/command');
+import { UpdateTask } from '../tasks/update';
+
+export interface UpdateOptions {
+  schematic?: boolean;
+}
+
+const UpdateCommand = Command.extend({
+  name: 'update',
+  description: 'Opens the official Angular API documentation for a given keyword.',
```
clydin: bad copy/paste?

packages/@angular/cli/commands/update.ts
```diff
@@ -0,0 +1,48 @@
...
+  ],
+
+  anonymousOptions: [
+    '<schematic>'
+  ],
+
+  run: function(commandOptions: any, rawArgs: Array<string>) {
+    let schematic = rawArgs[0] || '@schematics/package-update:all';
+    if (!schematic.match(/:/)) {
+      schematic = `@schematics/package-update:${schematic}`;
```
clydin: should this be restricted to just the `package-update` collection?  maybe even validated against the list of schematics in the collection as well?  Not sure the enduser should even be aware schematics are even in use.

packages/@angular/cli/tasks/update.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface UpdateTaskOptions {
+  dryRun: boolean;
+  force: boolean;
+}
+
+export const UpdateTask: any = Task.extend({
+  run: function(schematic: string, options: UpdateTaskOptions): Promise<any> {
+    const [collectionName, schematicName] = schematic.split(':');
+
+    const schematicRunTask = new SchematicRunTask({
```
clydin: Could potentially use `SchematicRunTask` directly.  Unless there's more planned.

packages/@angular/cli/commands/update.ts
Brocco: It is not limited to just `@schematics/package-update` but if you want to use another schematic lib, you need to specify the entire schematic name (`collectionName:schematicName`). As for the validation, that error will be handled/thrown by the schematics engine.

packages/@angular/cli/tasks/update.ts
Brocco: Possibly, but that layer is consistent with the other command/task combinations. I'm indifferent about this.

packages/@angular/cli/commands/update.ts
Brocco: Good catch, updating.

PR #9326, opened 2018-01-23
fix(@angular/cli): import all rxjs operators

### LINKED ISSUE #9314: Error in rxjs operators if installed locally with rxjs 5.5.6
May be related to #9196

I'm trying to `generate` `component` deep inside tree
If I remove remove `node_modules` folder and run `ng g c comp-name` it works fine
But if I install node_modules using yarn and run `ng g c comp-name` it fails with following error

```
Error: dryRunSink.commit(...).ignoreElements is not a function
dryRunSink.commit(...).ignoreElements is not a function
```
possibly because I have `rxjs : 5.5.6` in my app
error is coming from https://github.com/angular/angular-cli/blob/d3db0f6c6c2c0970bbea43985a98012629ef3f22/packages/%40angular/cli/tasks/schematic-run.ts#L117

### Versions
```
Angular CLI: 1.6.5
Node: 6.10.1
OS: win32 x64
Angular: 5.2.1

@angular/cdk: 5.1.0
@angular/cli: 1.6.5
@angular/material: 5.1.0
@angular-devkit/build-optimizer: 0.0.41
@angular-devkit/core: 0.0.28
@angular-devkit/schematics: 0.0.51
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.5
@schematics/angular: 0.1.16
typescript: 2.6.2
webpack: 3.10.0
```

### Repro steps

* Install rxjs 5.5.6 
* Install @angula/cli 1.6.5
* try to generate component

### Observed behavior
fails to generate component with error message 

```
Error: dryRunSink.commit(...).ignoreElements is not a function
dryRunSink.commit(...).ignoreElements is not a function
```

### Desired behavior

Should generate component

update package to use lettable `rxjs` operators

hansl:
This is before we actually move to lettable operators, which can take a while.

Fixes #9314

### DIFF
--- a/packages/@angular/cli/tasks/schematic-run.ts
+++ b/packages/@angular/cli/tasks/schematic-run.ts
@@ -12,7 +12,9 @@ import { Observable } from 'rxjs/Observable';
 import * as path from 'path';
 import chalk from 'chalk';
 import { CliConfig } from '../models/config';
+import 'rxjs/add/operator/concat';
 import 'rxjs/add/operator/concatMap';
+import 'rxjs/add/operator/ignoreElements';
 import 'rxjs/add/operator/map';
 import { getCollection, getSchematic } from '../utilities/schematics';
 


deebloo:
@hansl would you accept a PR that was just for porting to lettable/pipeable operators one file at a time?

hansl:
@deebloo we have a PR in flight (#8976) that does that. It's just not in a ready state for now. I think I might be able to just merge that other one if I poke @brocco a bit more ^_^

PR #9344, opened 2018-01-23
feat(@angular/cli): missing bootstrap code should not error

hansl:
It is totally valid now to have a different bootstrapping process. By removing
the error in resolveEntryModuleFromMain we allow the compilation to work.


hansl:
Please note with Ivy this is necessary as there will be no bootstrap.

Issue #6882, opened 2017-07-04
Autoprefixer options without ejecting


veke:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Versions.
```
@angular/cli: 1.2.0
node: 8.1.2
os: darwin x64
@angular/common: 4.2.5
@angular/compiler: 4.2.5
@angular/core: 4.2.5
@angular/forms: 4.2.5
@angular/http: 4.2.5
@angular/platform-browser: 4.2.5
@angular/platform-browser-dynamic: 4.2.5
@angular/router: 4.2.5
@angular/cli: 1.2.0
@angular/compiler-cli: 4.2.5
```

### Repro steps.
Cli v1.2.0 uses v6 postcss autoprefixer. It has issue with [grid support](https://github.com/postcss/autoprefixer/issues/856) 
Autoprefixer v7 has this issue fixed but the css grid support is disabled by default.
In order to get css grid prefixes working, autoprefixer({ grid: true }) is needed.
Is there a way to set options for autoprefixer without ejecting app?

### Desired functionality.
Option in .angular-cli.json to pass options

kylecordes:
Anyone know of a (non-Eject) workaround, until this can be taken care of?

veke:
You can manually add prefixes? :)

kylecordes:
@veke Manually keeping the two different vocabularies for grid settings in sync (to account for not just the prefixes, but the different ways of describing grid spans) between current Grid and the older spec that IE supports, is very unpleasant. Autoprefixer 7 can do this.

julienmartin:
We really need this since angular-cli 1.6.4+ uses Autoprefixer v7 !

florentsuc:
+1

kylecordes:
Prior to the **very** helpful commit above, the following very ugly hack did the job. It could probably be adapted by anyone with an urgent need to adjust autoprefixer options.


https://github.com/OasisDigital/cli-css-grid-demo/blob/master/patch_cli_autoprefixer.sh

PR #9366, opened 2018-01-24
fix(@angular/cli): maintain autoprefixer backward compatibility

clydin:
Autoprefixer 7+ changed the default for `grid` to false which causes different behavior for CLI end users.


florentsuc:
Related to #6882

PR #9373, opened 2018-01-25
feat(@angular/cli): hide error/warning build output stacktraces

clydin:
Internal Webpack stack traces are not very useful for standard build errors.  They are now hidden behind the verbose option.  This drastically reduces the verbiage when build errors or warnings are present.

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/utils.ts
+++ b/packages/@angular/cli/models/webpack-configs/utils.ts
@@ -6,16 +6,19 @@ export const ngAppResolve = (resolvePath: string): string => {
 
 const webpackOutputOptions = {
   colors: true,
-  hash: true,
-  timings: true,
-  chunks: true,
+  hash: true, // required by custom stat output
+  timings: true, // required by custom stat output
+  chunks: true, // required by custom stat output
   chunkModules: false,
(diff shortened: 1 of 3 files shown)


packages/@angular/cli/models/webpack-configs/utils.ts
```diff
@@ -14,16 +14,21 @@ const webpackOutputOptions = {
   modules: false,
   reasons: false,
   warnings: true,
-  assets: false, // listing all assets is very noisy when using assets directories
-  version: false
+  errors: true,
+  assets: true, // listing all assets is very noisy when using assets directories
```
filipesilva: `assets` should remain false.

clydin: It's used here: https://github.com/angular/angular-cli/blob/master/packages/%40angular/cli/utilities/stats.ts#L32

The CLI has actually been using the `verbose` option which enables `assets` by default. https://github.com/angular/angular-cli/blob/master/packages/%40angular/cli/tasks/build.ts#L47

I updated the comments to reflect which options are required by the CLI's custom output formatter.

filipesilva: I see now, thank you for explaining it to me.

PR #9402, opened 2018-01-26
refactor(@angular/cli): exclude builtin modules

paisanousa:
This pull request is a refactor of PR #9302 which was my original submission to resolve issue #9229. Since that time, @hansl and @filipesilva have resolved similar issues (commits ea22589, 6585ddb, 20860cc) with cleaner code that doesn't require the 'builtin-modules' package.

Other than the "!", this is not my contribution; it is @hansl 's.  I would prefer to see his solution implemented at this location.

If this PR needs to be closed and reopened under @hansl, I will be in full agreement.

### DIFF
--- a/lib/bootstrap-local.js
+++ b/lib/bootstrap-local.js
@@ -4,7 +4,6 @@
 const fs = require('fs');
 const path = require('path');
 const ts = require('typescript');
-const builtinModules = require('builtin-modules');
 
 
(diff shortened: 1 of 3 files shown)


paisanousa:
Not sure why,
```
X continuous-integration/travis-ci/pr — The Travis CI build could not complete ...
```
Details are all green.

PR #9405, opened 2018-01-26
fix(@ngtools/webpack): ensure webpack handles module resolution

clydin:
Ensure `package.json` fields take priority in node packages.
Also covers some edge cases where node packages contain TypeScript files.

Added `firebase@4.9.0` as a test case as it exhibits both issues.

target: 1.7.x


filipesilva:
@hansl can you review please?

PR #9410, opened 2018-01-27
fix(@angular/cli): avoid defaulting base-href to an empty string

### LINKED ISSUE #9297: ng serve reloads fail with cli 1.7.0 beta
### Versions
```
Angular CLI: 1.7.0-beta.1
Node: 9.4.0
OS: win32 x64
Angular: 6.0.0-beta.0
... router

@angular/cli: 1.7.0-beta.1
@ngtools/json-schema: 1.1.0
typescript: 2.6.2
webpack: 3.10.0
```

### Repro steps
create app with routing
ng serve, load home page, navigate away from home url
reload page in browser - or make edit and save for auto releoad

### Observed behavior
```
Load fails as it tries to load everything from wrong path (... below represents path that was navigated to)
GET http://localhost:4200/.../inline.bundle.js net::ERR_ABORTED
GET http://localhost:4200/.../polyfills.bundle.js net::ERR_ABORTED
GET http://localhost:4200/.../styles.bundle.js net::ERR_ABORTED
GET http://localhost:4200/.../vendor.bundle.js net::ERR_ABORTED
GET http://localhost:4200/.../main.bundle.js net::ERR_ABORTED
```

### Desired behavior
It should work as it does with cli version < 1.7.0

deployed dist continues to work, its only ng serve that is failing

### LINKED ISSUE #9401: Routing not working properly while using `ng serve` (1.7.0-beta.2)
### Versions
```
Angular CLI: 1.7.0-beta.2 (also happened in beta.1)
Node: 8.9.4
OS: win32 x64
Angular: 5.2.2
```

### Repro steps
* ng new cli-test --routing
* ng g c home
* ng g c about
* edit `app-routing.module.ts` to add the routes for the newly created components:
```
import { HomeComponent } from './home/home.component';
import { AboutComponent } from './about/about.component';

const routes: Routes = [
  { path: 'about', component: AboutComponent },
  { path: '', pathMatch: 'full', component: HomeComponent }
];
```
* edit `app.component.html`:
```
<a [routerLink]="['/']">Home</a>
<a [routerLink]="['/about']">About</a>

<router-outlet></router-outlet>
```
* ng serve
* access localhost:4200
* click on the `About` link. Url becomes `http://localhost:4200/about`
* Refresh the page. Now the HomeComponent is loaded instead of the AboutComponent.

### Observed behavior
After refreshing the page on the `/about` route (or directly accessing the app on `http://localhost:4200/about`), the url turns into `http://localhost:4200/about/` (trailing `/` is added) and the HomeComponent is loaded instead of the AboutComponent. From this point on, any link clicked adds to `/about/` (if we had more than '' and '/about')

clydin:
Fixes #9401
Fixes #9297

### DIFF
--- a/packages/@angular/cli/tasks/serve.ts
+++ b/packages/@angular/cli/tasks/serve.ts
@@ -74,7 +74,7 @@ export default Task.extend({
 
     const serveDefaults = {
       deployUrl: appConfig.deployUrl || '',
-      baseHref: appConfig.baseHref || '',
+      baseHref: appConfig.baseHref,
     };
 
     serveTaskOptions = Object.assign({}, serveDefaults, serveTaskOptions);


sbologescu:
I applied the fix for 9410 manually on my local, and while the behavior improved so that the browser refreshes the proper route on code changes, I noticed that my fontawesome path is still incorrect causing the icons on the page to be broken after a refresh.
For example, navigating from project root to a route like https://localhost:4200/[some_route] will cause fontawesome to be loaded correctly from the basehref (https://localhost:4200/fontawesome[hash]).
However, if you refresh the above route (or change code to cause a refresh), the browser will incorrectly try to retrieve fontawesome from https://localhost:4200/[some_route]/fontawesome[hash]

Thank you and regards,
Seb

PR #9414, opened 2018-01-27
feat(@angular/cli): better budget error messages

cexbrayat:
Improves error messages for budgets.

Before:

     ERROR in budgets, maximum exceeded for total scripts.

After:

    ERROR in budgets, maximum exceeded for total scripts. Budget 500 kB was exceeded by 83.9 kB.

### DIFF
--- a/packages/@angular/cli/plugins/bundle-budget.ts
+++ b/packages/@angular/cli/plugins/bundle-budget.ts
@@ -5,7 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { Budget, calculateBytes, calculateSizes } from '../utilities/bundle-calculator';
+import { Budget, Size, calculateBytes, calculateSizes } from '../utilities/bundle-calculator';
+import { formatSize } from '../utilities/stats';
 
 interface Thresholds {
   maximumWarning?: number;
@@ -34,7 +35,7 @@ export class BundleBudgetPlugin {
       }
 
       budgets.map(budget => {
-        const thresholds = this.calcualte(budget);
+        const thresholds = this.calculate(budget);
         return {
           budget,
           thresholds,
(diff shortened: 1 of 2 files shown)


packages/@angular/cli/plugins/bundle-budget.ts
```diff
@@ -43,54 +43,40 @@ export class BundleBudgetPlugin {
...
       cb();
     });
   }
 
-  private calcualte(budget: Budget): Thresholds {
+  private checkMinimum(threshold: number, size: Size, messages: any) {
+    if (threshold) {
+      if (threshold > size.size) {
+        messages.push(`budgets, minimum exceeded for ${size.label}. `
+        + `Expected ${threshold} but was ${size.size}.`);
```
Brocco: "Expected" reads a lot like a test, which seems a little bit off for me, plus I think that units would be nice here...

Thoughts on changing the text to something like this:

```
`Budget ${threshold} was exceeded by ${size.size - threshold}.`
```

Also, putting units on these values would be helpful see [here](https://github.com/angular/angular-cli/blob/a40e801a954fd4b79e8f042f9e1692971feadf9e/packages/%40angular/cli/utilities/stats.ts#L10-L19).

packages/@angular/cli/plugins/bundle-budget.ts
```diff
@@ -43,54 +43,40 @@ export class BundleBudgetPlugin {
...
+      }
+    }
+  }
+
+  private checkMaximum(threshold: number, size: Size, messages: any) {
+    if (threshold) {
+      if (threshold < size.size) {
+        messages.push(`budgets, maximum exceeded for ${size.label}.`
+        + `Expected ${threshold} but was ${size.size}.`);
```
Brocco: See comment above about the message.

packages/@angular/cli/plugins/bundle-budget.ts
cexbrayat: :+1: I did not dare to make `formatSize` public :)

cexbrayat:
@Brocco Done, if you want to give it a try.

PR #9429, opened 2018-01-29
fix(@angular/cli): fix the version of uglifyjs-webpack-plugin

hansl:
1.1.7 released without any file in it.


filipesilva:
Superseded by https://github.com/angular/angular-cli/pull/9435

Issue #9395, opened 2018-01-26
Deployment to Github Pages issue: Cannot match any routes


alamgirqazi:
### Versions
```
Angular CLI: 1.5.5
Node: 8.9.4
OS: win32 x64
Angular: 5.0.0
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 1.5.5
@angular-devkit/build-optimizer: 0.0.34
@angular-devkit/core: 0.0.22
@angular-devkit/schematics: 0.0.38
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.5
@schematics/angular: 0.1.8
typescript: 2.4.2
webpack: 3.8.1
```
### Error: Uncaught (in promise): Error: Cannot match any routes

I'm trying to deploy my Angular SPA app to github pages using this [/angular-cli/wiki](https://github.com/angular/angular-cli/wiki/stories-github-pages .)

I run this command `ng build --prod --output-path docs --base-href learningangular`

it reroutes my app from https://alamgirqazi.github.io/learningangular to https://alamgirqazi.github.io/learningangular/learningangular/

and I get the above error. I've added more detail here https://stackoverflow.com/questions/48436223/angular-cli-deploy-to-github-pages-issue-error-error-uncaught-in-promise-er# 

Here's the link to the app: https://alamgirqazi.github.io/learningangular

filipesilva:
Both an absolute and relative URL should work, so I'm not sure why the relative doesn't work in your case. That will need some looking into.

denisyilmaz:
in my case it works with leading/trailing slash in the base-href
```ng build --prod --output-path docs --base-href /PROJECT_NAME/```

**Comment by penghuili** [comment]
It's better to change the documentation(https://github.com/angular/angular-cli/wiki/stories-github-pages),
from 
```
ng build --prod --output-path docs --base-href PROJECT_NAME
```
to
```
ng build --prod --output-path docs --base-href /PROJECT_NAME/
```

mgechev:
This is already fixed in the docs on angular.io. Thanks for pointing this out!

PR #9434, opened 2018-01-30
changed base ref from project name to full repository name for deploy to github pages [change in wiki]

alamgirqazi:
Github pages publishing was resolved here https://stackoverflow.com/questions/48436223/angular-cli-deploy-to-github-pages-issue-error-error-uncaught-in-promise-er/48461805#48461805. 

changed baseref from project name to complete repository name.


alamgirqazi:
I signed it!

filipesilva:
Both an absolute and relative URL should work, so I'm not sure why the relative doesn't work in your case.

I don't think we should change the documentation to use the absolute url without investigating what's actually going wrong in https://github.com/angular/angular-cli/issues/9395. For this reason I am closing this PR.

PR #9449, opened 2018-01-31
fix(@angular/cli): guard against scripts without a src attribute

thepian:
If a script is inline it has no src attribute. The build should not
fail on looking for CSS scripts due to inline scripts.

This fixes #9448

PR #9452, opened 2018-01-31
fix(@ngtools/webpack): guard against scripts without a src attribute

thepian:
If a script is inline it has no src attribute. The build should not
fail on looking for CSS scripts due to inline scripts.


filipesilva:
Merged in https://github.com/angular/angular-cli/pull/9449

PR #9521, opened 2018-02-06
MAJOR VERSION ONLY - Size improvements

hansl:
The `node` flag in the config should probably be configurable when the build facade is done.

Although I'd like to see all of those changes in the major version, I know they may end up in a separate PR in the build facade.

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/browser.ts
+++ b/packages/@angular/cli/models/webpack-configs/browser.ts
@@ -80,8 +80,10 @@ export function getBrowserConfig(wco: WebpackConfigOptions) {
   return {
     resolve: {
       mainFields: [
+        'module',
         ...(wco.supportES2015 ? ['es2015'] : []),
-        'browser', 'module', 'main'
+        'browser',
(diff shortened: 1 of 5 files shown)


packages/@angular/cli/models/webpack-configs/browser.ts
```diff
@@ -80,8 +80,10 @@ export function getBrowserConfig(wco: WebpackConfigOptions) {
   return {
     resolve: {
       mainFields: [
+        'module',
```
filipesilva: Something is very broken if `module` is always resolved before the others. This needs to be investigated further.

devoto13: From `@angular/*` packages:

```
  "main": "./bundles/compiler.umd.js",
  "module": "./esm5/compiler.js",
  "es2015": "./esm2015/compiler.js",
```

`es2015` should definitely go first in the `mainFields`, otherwise ES2015 target will use ES5 build of Angular.

PR #9547, opened 2018-02-08
Added tip to make current working directory as angular project

shivarajnaidu:
Will be helpful to people those want to initialize there existing directory as angular project


shivarajnaidu:
I signed it!

Brocco:
This is not something we want to promote, it can cause issues with overwriting files.

PR #9573, opened 2018-02-09
fix(@angular/cli): ensure external component styles are optimized

clydin:
External component stylesheets retain their original filename which caused preprocessor-based component stylesheets to not be optimized in production.

target: 1.7.x

### DIFF
--- a/packages/@angular/cli/plugins/cleancss-webpack-plugin.ts
+++ b/packages/@angular/cli/plugins/cleancss-webpack-plugin.ts
@@ -16,11 +16,19 @@ interface Chunk {
 
 export interface CleanCssWebpackPluginOptions {
   sourceMap: boolean;
+  test: (file: string) => boolean;
 }
 
 export class CleanCssWebpackPlugin {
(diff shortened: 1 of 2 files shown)


packages/@angular/cli/plugins/cleancss-webpack-plugin.ts
```diff
@@ -16,11 +16,21 @@ interface Chunk {
...
+  test: (file: string) => boolean;
 }
 
 export class CleanCssWebpackPlugin {
-
-  constructor(private options: Partial<CleanCssWebpackPluginOptions> = {}) {}
+  private readonly _options: CleanCssWebpackPluginOptions;
+
+  constructor(options: Partial<CleanCssWebpackPluginOptions> = {}) {
+    this._options = {
```
filipesilva: Nit: isn't this the same as
```
    this._options = {
      sourceMap: false,
      test: (file) => file.endsWith('.css'),
      ...options,
    };
```

PR #9582, opened 2018-02-09
docs(@angular/cli): path node module location

MikhaelGerbet:
- Replace font awesome fonts location by the best practice definition


clydin:
The use of the tilde is Webpack specific and Webpack is considered an internal implementation detail of the CLI.  While Webpack is currently the only option for a build engine within the CLI, it may not necessarily be so moving forward.  As a result, the use of relative paths is the generally preferred option.

PR #9608, opened 2018-02-12
feat(@angular/cli): add the add command

hansl:


packages/@angular/cli/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+export default class AddCommand extends Command {
+  readonly name = 'add';
+  readonly description = 'Add support for a library to your project.';
+  scope = CommandScope.inProject;
+  arguments = ['collection'];
+  options: Option[] = [];
+
+  private _rawArgs: any;
+
+  async initializeRaw(args: any): Promise<any> {
```
Brocco: Why not perform this in `Command`?
`private _rawArgs` => `protected rawArgs`
This would reduce the efforts needed by each command.

packages/@angular/cli/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    const availableOptions: SchematicAvailableOptions[] = await getOptionsTask.run({
+      schematicName: 'ng-add',
+      collectionName,
+    });
+    let anonymousOptions: string[] = [];
+
+    if (availableOptions) {
+      const nameOption = availableOptions.filter(opt => opt.name === 'name')[0];
```
Brocco: Do we want to allow for `name` to be specially handled for `add` or would it be best left as another option? This name logic exists for the common name scenario usage of `generate`

`ng add some-lib --name foo`

packages/@angular/cli/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+  options: Option[] = [];
+
+  private _rawArgs: any;
+
+  async initializeRaw(args: any): Promise<any> {
+    this._rawArgs = args.slice();
+    return args;
+  }
+
+  public async parseOptions(collectionName: string): Promise<any> {
```
Brocco: Should not be public, and consider naming it something else to reduce confusion as parseOptions is imported as well. Possible name: `setSchematicOptions`

packages/@angular/cli/commands/add.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.arguments = this.arguments.concat(anonymousOptions);
+    this.options = this.options.concat(availableOptions || []);
+
+    return parseOptions(this._rawArgs, this.options, this.arguments);
+  }
+
+  async run(commandOptions: any) {
+    const collectionName = commandOptions.collection;
+
+    if (!collectionName) {
```
Brocco: This logic should live in the `validate` function.

packages/@angular/cli/models/command-runner.ts
```diff
@@ -131,23 +135,17 @@ function parseOptions(parser: Function,
...
-    const arg = args[index];
+    const arg = commandArguments[index];
     if (arg) {
-      optsWithMappedArgs[arg] = value;
+      parsedOptions[arg] = value;
     }
   });
 
-  delete optsWithMappedArgs._;
+  delete parsedOptions._;
```
Brocco: I've been thinking about this one... we should only delete if all the values have been mapped appropriately? Although you've allowed for the rawArgs to live on, just something to consider.

packages/@angular/cli/models/command.ts
```diff
@@ -14,16 +14,19 @@ export enum CommandScope {
...
     this.logger = logger;
     if (context) {
       this.project = context.project;
       this.ui = context.ui;
     }
   }
 
-  initialize(_options: any): Promise<void> {
-    return Promise.resolve();
+  async initializeRaw(args: any): Promise<any> {
```
Brocco: Like the earlier comment, I think this implementation should preserve the raw options on the base class as a protected property.

packages/@angular/cli/tasks/npm-install.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (packageManager === 'npm') {
+      installArgs.push('--quiet');
+    }
+    if (this.packageName) {
+      try {
+        // Verify if we need to install the package (it might already be there).
+        // If it's available and we shouldn't save, simply return. Nothing to be done.
+        resolve(this.packageName, { checkLocal: true, basedir: this.project.root });
+
+        if (!this.save) {
```
Brocco: `save` is not defined. Either add `save` or remove all references to it.

packages/@angular/cli/commands/add.ts
hansl: Done.

packages/@angular/cli/commands/add.ts
hansl: We should get rid of the `name` logic for the `generate`. I removed it from add.

packages/@angular/cli/commands/add.ts
hansl: Done

packages/@angular/cli/commands/add.ts
hansl: Done.

packages/@angular/cli/models/command-runner.ts
hansl: I'm going to fix the argv before RC, since we have one more week.

packages/@angular/cli/tasks/npm-install.ts
hansl: `save` _is_ defined as an option for the task.

packages/@angular/cli/models/command.ts
hansl: done.

Issue #9611, opened 2018-02-13
Vendor Chunk can't be generated in a Yarn Workspace


shaungrady:
### Versions
```
Angular CLI: 1.6.8
Node: 8.8.0
OS: darwin x64
Angular: 5.2.4
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 1.6.8
@angular-devkit/build-optimizer: 0.0.42
@angular-devkit/core: 0.0.29
@angular-devkit/schematics: 0.0.52
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.8
@schematics/angular: 0.1.17
typescript: 2.5.3
webpack: 3.10.0
```

### Repro steps
Use [this demo repo](https://github.com/shaungrady/ng-cli-yarn-workspaces-bug), or...
* Step 1: Create a Yarn workspace [as outlined here](https://yarnpkg.com/lang/en/docs/workspaces/#toc-how-to-use-it).
* Step 2: Ensure you have an ng-cli workspace.
* Step 3: Run `yarn workspace <workspace-name-here> ng build --vendor-chunk=true`.

### Observed behavior
No vendor chunk will be produced.

### Desired behavior
A vendor chunk to be produced.

### Mention any other details that might be useful (optional)

Looking at the source, the `projectPath` is defined by where `.angular-cli.json` resides, [as seen here](https://github.com/angular/angular-cli/blob/00ca690807a2326b6e8632a9a0ec46587b841a78/packages/%40angular/cli/models/webpack-config.ts#L37).

When `buildOptions.vendorChunk` is set, the `node_modules` path is set as being a child of `projectPath`, [as seen here](https://github.com/angular/angular-cli/blob/00ca690807a2326b6e8632a9a0ec46587b841a78/packages/%40angular/cli/models/webpack-configs/browser.ts#L28).

This behavior is all well and good, except when it comes to Yarn Workspaces. Yarn generates a `node_modules` folder inside each workspace—**_however_**, that `node_modules` directory does not contain package sources. So if ng-cli is inside a workspace, the `node_modules` inside that `projectPath` is essentially empty. Instead, package sources are located at the root of the Yarn Workspaces project, also in a `node_modules` folder.

[...]

PR #9621, opened 2018-02-14
feat(@angular/cli): support Yarn workspaces for vendor chunk

shaungrady:
Resolves https://github.com/angular/angular-cli/issues/9611.

When the Angular CLI project is inside of a Yarn workspace, the vendor chunk can't be generated as the `node_module` directory is at the workspace root, not the workspace where `.angular-cli.json` resides. This PR resolves that by determining if there's a Yarn workspace root. If there is, use that directory's `node_modules`, otherwise, fall back to the original behavior of using the `node_modules` within the `.angular-cli.json` project's directory.

I haven't written a test for this, yet. Not sure how to approach it with the fact that it requires a specific file structure to run the test with. Any guidance on this would be greatly appreciated!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3102,6 +3102,201 @@
         "pinkie-promise": "2.0.1"
       }
     },
+    "find-yarn-workspace-root": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/find-yarn-workspace-root/-/find-yarn-workspace-root-1.0.0.tgz",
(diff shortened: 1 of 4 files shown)


admosity:
Maybe we can make this generic to support lerna as well?

clydin:
Thank you for the contribution.  However, both use cases will be fully supported in CLI v6 with no additional changes required.  v6.0 is currently in the beta phase.  There are no current plans to add this as a feature to 1.7 at this time.

jeffora:
@clydin If this PR is working in 1.7, and 1.7 is broken for anyone wanting to use workspaces at the moment, is it too difficult to release a patch to 1.7 to unblock people?

Upgrading to v6 is going to be a potentially big change for existing project, and forcing a major framework version upgrade just to get a bugfix in the cli tool seems heavy-handed.

PR #9625, opened 2018-02-14
docs: add universal concurrent builds

hmagrini:
The current Universal Story suggests to run both the "browser" and "server" builds sequentially, this means that as applications grow, their build time will increase at a fast rate since the total time is the sum of both builds (although "server" builds are normally faster than the "browser", it's still adds a considerable time overhead).

This PR attempts to add a quick tip to run this builds in parallel. Since the CLI doesn't commit changes outside the output directory (and when it does, with the Universal Story both builds target different directories), it seems that both builds can run simultaneously (we are currently doing this on a large app with 300+ components and 15+ locales).

### DIFF
--- a/docs/documentation/stories/universal-rendering.md
+++ b/docs/documentation/stories/universal-rendering.md
@@ -72,7 +72,7 @@ import {AppComponent} from './app.component';
     // The AppServerModule should import your AppModule followed
     // by the ServerModule from @angular/platform-server.
     AppModule,
-    ServerModule, 
+    ServerModule,
     ModuleMapLoaderModule // <-- *Important* to have lazy-loaded routes work
   ],
   // Since the bootstrapped component is not inherited from your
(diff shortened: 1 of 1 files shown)


Toxicable:
if you want to run them concurrently then you can simple do 
`"build:client-and-server-bundles": "npm run build:browser & npm run build:server"`

hmagrini:
Yes for sure, but the output on the log would get all mixed together, whereas concurrently will identify each output line for each command you are running, which is particularly useful when errors get thrown (which might be different errors for each build target depending on the setup).

Concurrently allows you to customize the log output as well if you need it to be more clear, for example:

#### output from `npm run build:browser & npm run build:server`

default-operator

#### output from `concurrently --names "BROWSER,SERVER" "npm run build:browser" "npm run build:server"`

concurrently-default

(this is how we normally run the build command)

#### output from `concurrently --names "BROSER,SERVER" -c "bgBlue.bold,bgGreen.bold" "npm run build:browser" "npm run build:server"`

screen shot 2018-02-16 at 7 52 27 am

There are other features that might come in handy as applications grow and build scripts get more complex. 

Of course this is just my opinion, and I'm more than happy to edit to just use the `&` operator, but since the concurrency section is more of a "tip" than a guide, I thought it could help to suggest using a tool that is flexible.

Brocco:
The build system is currently being refactored and updated, this will be something that is definitely looked into once that new build system is implemented.

Thanks for putting this together, hopefully you're able to assist once the new build system is in place.

PR #9659, opened 2018-02-16
Adding --history-api-fallback CLI option to `ng serve`

### LINKED ISSUE #9640: Options to disable WebDevServer's historyApiFallback
Currently, `ng serve` will serve up the index page whenever an HTTP 404 error would otherwise occur. There are several instances where this is problematic:
* You've configured the router to use URL fragments (aka "#" hash) and want local development server to mirror production usage -- so that "works on my machine" isn't as possible
* You're trying to serve a new file and you're not sure if it's being bundled/output/copied successfully. It's frustrating when you try to browse around to find the new file and get index.html contents instead

### Repro steps
* Configure your angular app's router to use URL fragments (aka hash, "#")
* Possibly make changes to `.angular-cli.json`
* Open http://localhost:4200/page-that-does-not-exist.html

### Observed behavior
* http://localhost:4200/page-that-does-not-exist.html will serve up the app's index page

### Desired behavior
* I want those 404 Not Found's!

### Notes
While it would be *fantastic* if the cli could inspect the app to determine that the router was configured with [`{ useHash: true}`](https://angular.io/api/router/ExtraOptions#members), I'd also be willing to accept a command-line parameter or some new option in `.angular-cli.json`.

The 404-handling is using WebpackDevServer's [.historyApiFallback option](https://webpack.js.org/configuration/dev-server/#devserver-historyapifallback).

See also: [./packages/@angular/cli/tasks/serve.ts](https://github.com/angular/angular-cli/blob/a40e801a954fd4b79e8f042f9e1692971feadf9e/packages/%40angular/cli/tasks/serve.ts#L214)

eib:
Fixes #9640 by adding a CLI option (defaults to current behavior).

While it would be swell to pull the correct behavior from each app's router configuration, that's a little outside my area of expertise. This doesn't feel important enough to polluting the .angular-cli.json schema over, either. In the end, just a CLI option is enough for my needs.

Feedback/edits welcome.

### DIFF
--- a/docs/documentation/serve.md
+++ b/docs/documentation/serve.md
@@ -59,6 +59,16 @@ All the build Options are available in serve, below are the additional options.
 </details>
 
 <details>
+  <summary>history-api-fallback</summary>
+  <p>
+    <code>--history-api-fallback</code> <em>default value: true</em>
+  </p>
+  <p>
+    Serves the index page instead of HTTP 404 "Not Found" errors (but only when the `Accept` HTTP header includes "html" content-types).
+  </p>
+</details>
+
+<details>
   <summary>open</summary>
   <p>
     <code>--open</code> (aliases: <code>-o</code>) <em>default value: false</em>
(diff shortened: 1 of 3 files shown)


filipesilva:
Hi, we aren't thinking of adding new flags right now, and we are adding an extensibility model in the next major version that will allow you to replace the dev server. For these reasons I am closing this PR.

PR #9673, opened 2018-02-18
Fix a compilation error that occurs when the "paths" option is not set but "baseUrl" is

### LINKED ISSUE #9671: Module resolution fails in v1.7.0 when "baseUrl" is set in tsconfig
### Versions
```
Angular CLI: 1.7.0
Node: 8.9.4
OS: linux x64
Angular: 5.2.5

@angular/cli: 1.7.0
@angular-devkit/build-optimizer: 0.3.1
@angular-devkit/core: 0.3.1
@angular-devkit/schematics: 0.3.1
@ngtools/json-schema: 1.2.0
@ngtools/webpack: 1.10.0
@schematics/angular: 0.3.1
@schematics/package-update: 0.3.1
typescript: 2.6.2
webpack: 3.11.0
```

### Repro steps
* Run `git clone https://github.com/mprahl/angular-cli-9671-reproducer.git`
* Run `cd angular-cli-9671-reproducer && npm install && ng serve`

### Observed behavior
```
ERROR in ./src/app/app.module.ts
Module not found: Error: Can't resolve 'reproducer/hello/hello.component' in '/path/to/angular-cli-9671-reproducer/src/app'
```

### Desired behavior
The imports should work as they did in v1.6.x.

The following commit broke this functionality: https://github.com/angular/angular-cli/commit/29aa274b6721c1399ca2baca18565b21254fb3f0

Running angular-cli with commit https://github.com/angular/angular-cli/commit/29aa274b6721c1399ca2baca18565b21254fb3f0 reverted works as expected.

mprahl:
Fixes #9671

### DIFF
--- a/packages/@ngtools/webpack/src/paths-plugin.ts
+++ b/packages/@ngtools/webpack/src/paths-plugin.ts
@@ -32,23 +32,25 @@ export function resolveWithPaths(
   }
 
   // check if any path mapping rules are relevant
-  const isPathMapped = compilerOptions.paths && Object.keys(compilerOptions.paths)
-    .some(pattern => {
-      // can only contain zero or one
-      const starIndex = pattern.indexOf('*');
-      if (starIndex === -1) {
-        return pattern === request.request;
-      } else if (starIndex === pattern.length - 1) {
-        return request.request.startsWith(pattern.slice(0, -1));
-      } else {
-        const [prefix, suffix] = pattern.split('*');
-        return request.request.startsWith(prefix) && request.request.endsWith(suffix);
-      }
-    });
+  if (compilerOptions.paths) {
+    const isPathMapped = Object.keys(compilerOptions.paths)
+      .some(pattern => {
+        // can only contain zero or one
+        const starIndex = pattern.indexOf('*');
+        if (starIndex === -1) {
+          return pattern === request.request;
+        } else if (starIndex === pattern.length - 1) {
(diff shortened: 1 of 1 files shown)


mprahl:
I signed it!

clydin:
Thank you for the contribution.  However, while this may fix the problem observed within the referenced issue, this will break module resolution in the general case.  The `ts.resolveModuleName` method is unfortunately too course-grained to fully support module resolution in all cases.

mprahl:
@clydin what cases would cause this to break? This works on a new app generated from angular-cli and the app that wouldn't compile with v1.7.0.

clydin:
For a package, `ts.resolveModuleName` will resolve to the `typings`/`types` field of the `package.json`.  While appropriate for transpiling typescript, this is not the case for bundling the actual code for the package.  The location of the actual code can potentially be found at multiple locations depending on the configuration (e.g., fields such as `main`, `browser, `module`, or `es2015`).  As the type definition file can be located at an arbitrary location within the package;  there's no clean method to determine the appropriate source code file from only the path of the type definition file returned.  Factor in deep imports, as well as the no-typings fallback case of using a root `index.ts`/`index.js`, and `ts.resolveModuleName` is not useable for general resolution.  If the intrinsics behind the method were publicly available, it would potentially possible.
Longer-term, the method's use will most likely be removed.

PR #9686, opened 2018-02-19
fix: Respect build assets allowOutsideOutdir

### LINKED ISSUE #9363: allowOutsideOutDir for assets not working again
### Versions
```
Tested on:
1.6.3
1.6.5
Windows 10
```

### Repro steps
* npm install -G @angular/cli
* cd to project with assets that need to be copied outside the project folder. From the .angular-cli.json:
"assets": [{
        "input": "./assets/",
        "glob": "**/*",
	    	"output": "../EcView/WebContent/assets",
        "allowOutsideOutDir": true
      }],

* ng build

### Observed behavior
Build failed:
```
>ng build
An asset cannot be written to a location outside the project.
```

### Desired behavior
The build process copies the assets to the specified folder and completes the compilation of source code.

### Other details
Seems similar to, possible regression of https://github.com/angular/angular-cli/issues/8613

trendzetter:
The old check should be removed so assets are allowed outside the project folder if allowOutsideOutdir is configured.
Fixes #9363

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/common.ts
+++ b/packages/@angular/cli/models/webpack-configs/common.ts
@@ -96,13 +96,6 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
       const outputRelativeOutput = path.relative(absoluteOutputPath, absoluteAssetOutput);
 
       if (outputRelativeOutput.startsWith('..') || path.isAbsolute(outputRelativeOutput)) {
-
-        const projectRelativeOutput = path.relative(projectRoot, absoluteAssetOutput);
-        if (projectRelativeOutput.startsWith('..') || path.isAbsolute(projectRelativeOutput)) {
-          const message = 'An asset cannot be written to a location outside the project.';
-          throw new SilentError(message);
-        }
-
         if (!asset.allowOutsideOutDir) {
           const message = 'An asset cannot be written to a location outside of the output path. '
                         + 'You can override this message by setting the `allowOutsideOutDir` '


trendzetter:
I signed it!

clydin:
That's not an old check. It's intentional and prevents writing to arbitrary locations on the filesystem.

trendzetter:
@clydin: It's checked twice. Once taking into account the optional configuration and the other is not taking into account a valid configuration directive

clydin:
It's not checked twice; they are two different checks.  The first check verifies it is within the `outputPath` for the app.  If it is not, then the second check verifies if it is within the project itself.  Then it checks if assets are allowed to be written outside the `outputPath` (but still within the project due to the previous check).

trendzetter:
I am reading it again because initially I thought this issue was marked as easy and it seemed there are simply too many checks. 
@clydin:  "Also prevent writing outside the project path" is in the code. How is the "project path" defined? is it a configuration flag or simply the directory where angular cli is called?

yuvalbl:
Angular version: 6.0.0-rc.5
Angular-CLI version: 6.0.0-rc.5

I've created a new project and was able to build to a directory outside the project root, with no  allowOutsideOutDir flag

side-note:
If I get it correctly, ***.angular-cli.json*** configuration is now included in ***angular.json***.

Repro steps:
1. create new project.
2. in angular.json change outputPath to: `'../MyDir'`
3. run `ng build`
4. see build output outside the project  root directory

trendzetter:
@yuvalbl It's not merged, is it?

PR #9704, opened 2018-02-20
refactor: remove support for Angular 2 and 4

clydin:


packages/@ngtools/webpack/src/paths-plugin.ts
```diff
@@ -87,76 +81,3 @@ export function resolveWithPaths(
...
 }
-
-export interface PathsPluginOptions {
-  nmf: NormalModuleFactory;
-  tsConfigPath: string;
-  compilerOptions?: ts.CompilerOptions;
-  compilerHost?: ts.CompilerHost;
-}
-
-export class PathsPlugin implements Tapable {
```
hansl: What are we using in Ng5+ for TS path resolution?

tests/e2e/tests/packages/webpack/server-ng5.ts
```diff
@@ -1,18 +1,22 @@
...
 import { createProjectFromAsset } from '../../../utils/assets';
 import { exec } from '../../../utils/process';
-import { expectFileToMatch, rimraf } from '../../../utils/fs';
+import { expectFileToMatch } from '../../../utils/fs';
 
 
 export default function (skipCleaning: () => void) {
+  // This test was broken as it was a copy of the ng2 server test
+  // and did not actually test ng5
+  return;
```
filipesilva: I take it is not working currently and needs further attention in a separate PR, correct?

packages/@ngtools/webpack/src/paths-plugin.ts
clydin: AngularCompilerPlugin uses the `resolveWithPaths` function directly: https://github.com/angular/angular-cli/blob/master/packages/%40ngtools/webpack/src/angular_compiler_plugin.ts#L654

tests/e2e/tests/packages/webpack/server-ng5.ts
clydin: Yes.  I tried to fix up the actual test but it appears that the test asset project is a copy of the ng2 version as well so more work is needed.

Issue #9691, opened 2018-02-20
ajv@^6.0.0 is missing


KieraDOG:
npm WARN ajv-keywords@3.1.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.

### Versions
```
Angular CLI: 1.6.8
Node: 8.9.4
OS: darwin x64
Angular: 4.4.4
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router, tsc-wrapped

@angular/cdk: 2.0.0-beta.12
@angular/cli: 1.6.8
@angular/material: 2.0.0-beta.12
@angular-devkit/build-optimizer: 0.0.42
@angular-devkit/core: 0.0.29
@angular-devkit/schematics: 0.0.52
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.8
@schematics/angular: 0.1.17
typescript: 2.3.4
webpack: 3.10.0
```

### Repro steps

* Step 1

``` npm i -g @angular/cli ```

* Step 2

``` npm WARN ajv-keywords@3.1.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself. ```

### Observed behavior
```
long@LONG-MBP  ~/Developer/ngx   LONG-jenkins-submodule  npm update @angular/cli
npm WARN ajv-keywords@3.1.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.

+ @angular/cli@1.7.0
added 34 packages, removed 139 packages, updated 45 packages and moved 3 packages in 18.928s
```

### Desired behavior

angular installed without any error or warning

agllai:
me too have the same problem 
and when i tried to install ajv with the command :npm install ajv i get this 
npm WARN saveError ENOENT: no such file or directory, open 'C:\WINDOWS\system32\package.json'
npm WARN enoent ENOENT: no such file or directory, open 'C:\WINDOWS\system32\package.json'
npm WARN system32 No description
npm WARN system32 No repository field.
npm WARN system32 No README data
npm WARN system32 No license field.

+ ajv@6.1.1
updated 1 package in 3.14s

Yisus17:
Same issue :S

KellyR-STCU:
I am also seeing this

AianaM:
+1

timbilt:
[...]

PR #9713, opened 2018-02-21
fix(@angular/cli): work around npm peerdep issue

filipesilva:
NPM hoists dependencies with peerDeps of their own, leading to unmet peer dependencies.

This workaround should be safe for now since the only two packages that depend on `ajv-keywords@3.1.0` are `webpack` and `schema-utils`.

See https://github.com/angular/angular-cli/issues/9691#issuecomment-367322703 and https://github.com/npm/npm/issues/19877 for more information.

Fix https://github.com/angular/angular-cli/issues/9691.

### DIFF
--- a/packages/@angular/cli/lib/cli/index.ts
+++ b/packages/@angular/cli/lib/cli/index.ts
@@ -1,3 +1,10 @@
+// TODO: remove this commented AJV require.
+// We don't actually require AJV, but there is a bug with NPM and peer dependencies that is
+// whose workaround is to depend on AJV.
+// See https://github.com/angular/angular-cli/issues/9691#issuecomment-367322703 for details.
+// We need to add a require here to satisfy the dependency checker.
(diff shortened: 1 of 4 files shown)


packages/@angular/cli/lib/cli/index.ts
```diff
@@ -1,3 +1,9 @@
+// We don't actually require AJV, but there is a bug with NPM and peer dependencies that is
+// whose workaround is to depend on AJV.
+// See https://github.com/angular/angular-cli/issues/9691#issuecomment-367322703 for details.
+// We need to add a require here to satisfy the dependency checker.
+// require('ajv');
```
hansl: Could you add a TODO here to make sure we get it in our reminders?

filipesilva: Done

PR #9754, opened 2018-02-23
doc: add deploy URL and base HREF comparison

clydin:


clydin:
updated with a legend and correct scope

PR #9761, opened 2018-02-23
fix(@angular/cli): remove modules only on prod

filipesilva:
See https://github.com/webpack/webpack/issues/6248#issuecomment-368160223


filipesilva:
Closing for now, could not see improvement in benchmarks (see https://github.com/webpack/webpack/issues/6248#issuecomment-368489543).

PR #9787, opened 2018-02-26
fix(@ngtools/webpack): prevent relative request path mapping

### LINKED ISSUE #9727: Build error: Zone js wih @ngtools/webpack `1.10.1`
### Versions
```
@ngtools/webpack: 1.10.1
```

I am using the `@ngtools/webpack` outside of the Angular CLI, and the the latest patch version from 1.10.0 to 1.10.1. I am getting the following error;

```bash
Module not found: Error: Can't resolve 'C:\git\solutions\webpack\buildin\global.js' in 'C:\git\solutions\site\node_modules\zone.js\dist'
```

### Observed behavior
```
It looks the the module resolution or webpack context is not correct.
```

### Desired behavior
No error.

I am thinking it might be related to this: https://github.com/angular/angular-cli/commit/595168eeb251c9928d42139424e05dcb50270b08#diff-c8ee7673e7d5e8741306839a34f0c8fc

I'll try to debug some more.

clydin:
Fixes #9727

### DIFF
--- a/packages/@ngtools/webpack/src/paths-plugin.ts
+++ b/packages/@ngtools/webpack/src/paths-plugin.ts
@@ -25,30 +25,43 @@ export function resolveWithPaths(
     return;
   }
 
+  const originalRequest = request.request.trim();
+
+  // Relative requests are not mapped
+  if (originalRequest.startsWith('.') || originalRequest.startsWith('/')) {
+    callback(null, request);
+    return;
+  }
+
   // check if any path mapping rules are relevant
   const pathMapOptions = [];
   for (const pattern in compilerOptions.paths) {
       // can only contain zero or one
       const starIndex = pattern.indexOf('*');
       if (starIndex === -1) {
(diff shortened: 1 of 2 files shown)


packages/@ngtools/webpack/src/paths-plugin.ts
```diff
@@ -25,30 +25,43 @@ export function resolveWithPaths(
     return;
   }
 
+  const originalRequest = request.request.trim();
+
+  // Relative requests are not mapped
+  if (originalRequest.startsWith('.')) {
```
hansl: or `/`?

clydin: good point.  done.

eromano:
@clydin @hansl @filipesilva  why has been added this limitation in the paths mapping? It sounds like a problem to me when you develop specially a component library.

PR #9805, opened 2018-02-27
feat(@angular/cli): minify inline CSS, JS in prod

valentinxxx:
This PR enables `minifyCSS` and `minifyJS` in `html-webpack-plugin` (prod build).

Angular app preloading screen can be styled in the index.html inside the `<app-root></app-root>` tag (with inline style for performance).

Sample: https://medium.com/@tomastrajan/how-to-style-angular-application-loading-with-angular-cli-like-a-boss-cdd4f5358554

Another example for JS is GTM code in index.html.

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
    <title>Angular</title>
  </head>
  <body>
    <!-- Google Tag Manager (noscript) -->
    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-XXXX" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
    <!-- End Google Tag Manager (noscript) -->
    <app-root>
      <style>
        .pre-loader {
          font-weight: 700;
        }
      </style>
      <div class="pre-loader">Loading...</div>
    </app-root>
    <!-- Google Tag Manager -->
    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
    })(window,document,'script','dataLayer','GTM-XXXX');</script>
    <!-- End Google Tag Manager -->
  </body>
</html>
```

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/browser.ts
+++ b/packages/@angular/cli/models/webpack-configs/browser.ts
@@ -41,7 +41,9 @@ export function getBrowserConfig(wco: WebpackConfigOptions) {
         minify: buildOptions.target === 'production' ? {
           caseSensitive: true,
           collapseWhitespace: true,
-          keepClosingSlash: true
+          keepClosingSlash: true,
+          minifyCSS: true,
+          minifyJS: true
         } : false
     }));
     extraPlugins.push(new BaseHrefWebpackPlugin({


valentinxxx:
I signed it!

filipesilva:
@clydin can you have a look? I think we'd have to overwrite the default minifiers with our own.

alfaproject:
What would be amazing is that we could have something like: `<link rel="stylesheet" type="text/text/x-scss" href="loader.scss" inline>`

And it would compile and inline that. That would be a killer feature. Currently, I have a separate task to generate that file and then inline manually. \:

clydin:
The JS code example is already minified, there is not much more that can be done there.  Additionally, anything that is placed in the index.html file can potentially be pre-minified or otherwise processed.

It's also important to note that inline styles and javascript (more prominently) create a potential security issue and are generally banned in standard CSP rule configurations.  Also, with the advent of HTTP/2,  the use of pre-fetch/pre-load hints, and proper caching configuration, having small stylesheet/JS files designed for early display is quite possible.  The CLI supports generating separate stylesheets that can contain only the single css/scss/etc. source file in additional to the standard global stylesheet bundle.

The index.html file can also be completely generated as a pre-build step.  Allowing for any type of templating/injection/etc. to be performed before the CLI is run.

Typically http servers will also automatically compress (gz/brotli/etc) the files before transfer which will significantly reduce the size of the file on its own as whitespace is really the size driver in this situation.  Especially considering any potential inline `loading` styling will most likely be minimal in size and therefore minifying directly will have limited effect on size reduction outside of whitespace removal.

Factor in the added complexity, requirement provide ongoing support the features, and most likely the need to externalize at least some of the settings for each minifier (as the code/styles in question are completely outside the control of the CLI) and this is most likely not something that the CLI should support at this time.

PR #9812, opened 2018-02-28
refactor(@angular/cli): remove webpack nodeJS globals shimming

clydin:
BREAKING CHANGE:
Libraries/Packages must contain a web version if not web specific.


filipesilva:
This for reference, this PR is a followup to https://github.com/angular/angular-cli/pull/8250.

Issue #9820, opened 2018-02-28
ng build --prod failed - TypeError: Cannot read property '1' of undefined


RafaelFronteira:
It work very well in develop mode (ng serve), but I can't build to production

### Versions
Angular CLI: 1.7.1
Node: 6.13.0
OS: linux x64
Angular: 5.2.6
 @ angular/cdk: 2.0.0-beta.12
 @ angular/cli: 1.7.1
 @ angular-devkit/build-optimizer : 0.3.2
 @ angular-devkit/core: 0.3.2
 @ angular-devkit/schematics : 0.3.2
 @ ngtools/json-schema : 1.1.0
 @ ngtools/webpack : 1.10.1
 @ schematics/angular : 0.3.2
 @ schematics/package-update : 0.3.2
 typescript : 2.5.3
 webpack : 3.11.0

### Repro steps
ng build --prod

### The log given by the failure

10% building modules 6/6 modules 0 active/home/rafael/Documents/bimweb-spa/bimweb-spa/node_modules/clean-css/lib/optimizer/level-2/properties/populate-components.js:30
          throw e;
          ^
TypeError: Cannot read property '1' of undefined    at Object.font [as breakUp] (/home/rafael/Documents/bimweb-spa/bimweb-spa/node_modules/clean-css/lib/optimizer/level-2/break-up.js:308:66)
    at populateComponents (/home/rafael/Documents/bimweb-spa/bimweb-spa/node_modules/clean-css/lib/optimizer/level-2/properties/populate-components.js:17:42)
    at optimizeProperties (/home/rafael/Documents/bimweb-spa/bimweb-spa/node_modules/clean-css/lib/optimizer/level-2/properties/optimize.js:19:3)
    at recursivelyOptimizeProperties (/home/rafael/Documents/bimweb-spa/bimweb-spa/node_modules/clean-css/lib/optimizer/level-2/optimize.js:63:9)
    at level2Optimize (/home/rafael/Documents/bimweb-spa/bimweb-spa/node_modules/clean-css/lib/optimizer/level-2/optimize.js:77:3)
    at optimize (/home/rafael/Documents/bimweb-spa/bimweb-spa/node_modules/clean-css/lib/clean.js:131:5)
    at /home/rafael/Documents/bimweb-spa/bimweb-spa/node_modules/clean-css/lib/clean.js:104:29
[...]

PR #9876, opened 2018-03-06
update @angular-devkit/core version to fix prod build TypeError

hdayacr:
In order to fix: https://github.com/angular/angular-cli/issues/9820 we need to update the @angular-devkit/core module


clydin:
Thank you for the contribution.  However, devkit versions are adjusted during the CLI release process.  1.x is currently fixed to 0.3.x of the devkit.  If there is a fix in 0.4 that should be backported to 0.3 then please open an issue in the devkit repo.

PR #9883, opened 2018-03-06
feat(@angular/cli): use architect

filipesilva:
Fix https://github.com/angular/angular-cli/issues/9973


packages/@angular/cli/commands/build.ts
```diff
@@ -1,293 +1,44 @@
...
-    type: Boolean,
-    description: 'Flag to prevent building an app shell',
-    default: false
-  }
-];
-
-export interface BuildTaskOptions extends BuildOptions {
-  statsJson?: boolean;
+export interface Options {
+  app?: string;
```
hansl: `project`?

hansl: Need to be changed this week, but can be after this PR.

PR #9886, opened 2018-03-06
fix(@ngtools/webpack): support invalidating compiler-host with a directory

clydin:
`invalidate` is called from the Webpack `InputFileSystem.purge` method which can be passed a file or directory.

### DIFF
--- a/packages/@ngtools/webpack/src/lazy_routes.ts
+++ b/packages/@ngtools/webpack/src/lazy_routes.ts
@@ -33,7 +33,7 @@ export function findLazyRoutes(
   if (program) {
     sourceFile = program.getSourceFile(fileName);
   }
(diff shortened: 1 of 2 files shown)


packages/@ngtools/webpack/src/lazy_routes.ts
```diff
@@ -42,7 +42,7 @@ export function findLazyRoutes(
     );
   }
   if (!sourceFile) {
-    throw new Error(`Source file not found: '${fileName}'.`);
+    return {};
```
filipesilva: Shouldn't the error still be thrown here?

clydin: Re-added error.

PR #9891, opened 2018-03-06
feat(@angular/cli): Add config command to replace (get/set).

Brocco:
BREAKING CHANGE: Removing the get and set commands in favor of the new config


packages/@angular/cli/commands/config.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  private updateLintForPrefix(filePath: string, prefix: string): void {
+    if (!fs.existsSync(filePath)) {
+      return;
+    }
+
+    const tsLint = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+
+    const componentLint = tsLint.rules['component-selector'][2];
+    if (componentLint instanceof Array) {
```
clydin: `Array.isArray`? This is old code but why not.

packages/@angular/cli/commands/config.ts
```diff
@@ -0,0 +1,135 @@
...
+      tsLint.rules['component-selector'][2].push(prefix);
```
clydin: Re-use `componentLint` variable?

packages/@angular/cli/commands/config.ts
```diff
@@ -0,0 +1,135 @@
...
+      aliases: ['g'],
+      description: 'Get/set the value in the global configuration (in your home directory).'
+    }
+  ];
+
+  public run(options: ConfigOptions) {
+    const config = options.global ? CliConfig.fromGlobal() : CliConfig.fromProject();
+
+    if (config === null) {
+      throw new SilentError('No config found. If you want to use global configuration, '
```
clydin: Use `logger.error` and return an errorCode (or just return)? does the refactoring support returning a number?  It'd be nice if we could remove the need to use `SilentError` and only deal with exceptions propagating up from here when something unexpected or fatal occurs.

packages/@angular/cli/commands/config.ts
```diff
@@ -0,0 +1,135 @@
...
+      let value: any = options.value;
+      switch (type) {
+        case 'boolean': value = this.asBoolean(options.value); break;
+        case 'number': value = this.asNumber(options.value); break;
+        case 'string': value = options.value; break;
+
+        default: value = this.parseValue(options.value, options.jsonPath);
+      }
+
+      if (options.jsonPath.indexOf('prefix') > 0) {
```
clydin: We probably should strengthen this check at some point as it will catch anything with the word `prefix`.

packages/@angular/cli/commands/config.ts
Brocco: The basis is that failing command runs throw an exception, but what should happen here is that this check should be run during the validate method.

clydin: I suppose that basis is what I'm questioning which is really more systemic than particular to this command.  There's failing commands due to unexpected behavior (bad logic, memory issues, etc.)  and failing commands due to do expected behavior (in this case the config not found or improper value).  The first should definitely give rise to an exception but the second seems like something that could be better handled locally.  Control flow via exception can be highly problematic.  A try/catch in the wrong place and the user never knows something bad happened.  An error is also not necessarily fatal.  So potentially, using the `logger.error` for expected user errors and reserving `logger.fatal` for actual exceptions caught at a higher level could help the user understand the severity of the situation.  As opposed to the current version, where error reporting is fairly inconsistent and confusing.

packages/@angular/cli/commands/config.ts
```diff
@@ -0,0 +1,135 @@
...
+      return true;
+    } else if (raw == 'false' || raw == '' || raw == '0') {
+      return false;
+    } else {
+      throw new SilentError(`Invalid boolean value: "${raw}"`);
+    }
+  }
+
+  private asNumber(raw: string): number {
+    if (Number.isNaN(+raw)) {
```
Toxicable: `isNaN` already coerces to `number`

packages/@angular/cli/commands/config.ts
```diff
@@ -0,0 +1,135 @@
...
+      throw new SilentError(`Invalid number value: "${raw}"`);
+    }
+    return +raw;
```
Toxicable: For everyone's sanity could ya use `Number(raw)` instead

packages/@angular/cli/commands/config.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  private updateLintForPrefix(filePath: string, prefix: string): void {
+    if (!fs.existsSync(filePath)) {
+      return;
+    }
+
+    const tsLint = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+
+    let componentLint = tsLint.rules['component-selector'][2];
```
Toxicable: destructing would be better fit here I think?

packages/@angular/cli/commands/config.ts
Brocco: Done

packages/@angular/cli/commands/config.ts
Brocco: Done

packages/@angular/cli/commands/config.ts
Brocco: Moot point now... creating the local variable did me no good, as I had to write it back to the origin not to the local variable.

PR #9910, opened 2018-03-08
refactor(@angular/cli): remove deploy URL option and CSS special cases

clydin:
Superseded by https://github.com/angular/devkit/pull/505

For reference, a comparison of the deploy URL and base HREF options can be found here: https://github.com/angular/angular-cli/blob/master/docs/design/deployurl-basehref.md

BREAKING CHANGE:

The deploy URL option has been removed.  The base HREF option alone covers a majority of the use cases of the deploy URL option with the added benefit of affecting relative URL assets referenced in component HTML templates.  For more advanced use cases, the Angular `APP_BASE_HREF` token can be used in conjuction to adjust the Angular routing root path.
Also, root-relative CSS resources will no longer be modified when a base HREF is present.  For cases where a relative URL referenced resource should not be processed by the CLI, a caret `^` may be prefixed to the URL of the CSS resources to prevent the CLI from processing or otherwise modifying the URL and resource.


filipesilva:
I think these changes need to be added to the DevKit repo instead, we're in the middle of removing the webpack configs from this one.

changhuixu:
@clydin  Excellent. This is an exciting change.
Currently, webpack public path uses ```--deploy-url``` to load [```lazy loading feature modules```](https://angular.io/guide/lazy-loading-ngmodules).  What will be the "public path" in future? base HREF or APP_BASE_HREF? Or do we need to worry about it at all? Thank you.

clydin:
@changhuixu The HTML base HREF applies to all relative URL requests within the application.  This includes lazy loaded modules.  Webpack internally injects a `<script/>` element into the DOM with `src` attribute referencing the bundle to load as a relative URL.

changhuixu:
Great. Thank you.

PR #9918, opened 2018-03-09
refactor: Move task logic into commands

Brocco:


packages/@angular/cli/models/command-runner.ts
```diff
@@ -9,7 +9,9 @@ import { oneLine } from 'common-tags';
 import { logging } from '@angular-devkit/core';
 import { camelize } from '@angular-devkit/core/src/utils/strings';
 
-const yargsParser = require('yargs-parser');
+// const yargsParser = require('yargs-parser');
```
hansl: Remove commented code.

Brocco: Fixed

PR #9986, opened 2018-03-16
Update build.ts

lioman:
fix typo


clydin:
Thank you for the contribution.  However, the file in question no longer exists within the current code base.

PR #10001, opened 2018-03-18
feat(@ngtools/webpack): always use ngc to discover lazy routes

clydin:
This changes the plugin to always use the `@angular/compiler-cli` TypeScript program but disables all generation of factory code and metadata, as well as template checking, when in JIT mode.  Lazy route discovery is then always performed using the more comprehensive metadata analysis method that was previously only used in AOT mode.  Lazy routes are also now only updated when TS files are changed.  Further optimizations could allow for further limits on updates based on the content of the changed TS files (i.e., presence of Angular metadata, etc.).


clydin:
Move to the devkit project as: https://github.com/angular/devkit/pull/601

Issue #1568, opened 2016-08-05
Feature request: ng serve should select another port if 4200 is already taken


amreladawy:
> ## Feature request
- OS?  
  Windows 7
- Versions. 

```
angular-cli: 1.0.0-beta.11-webpack
node: 4.4.7
os: win32 x64
```
- Repro steps. 

a.   start an angular project with `ng serve`
b. start another angular project with `ng serve` 

The second one won't start as port 4200 is already taken 
`Port 4200 is already in use.`

Ionic cli is switching to another port automatically 
`The port 8100 was taken on the host localhost - using port 8101 instead`

mwawrusch:
Random ports are a bad idea, especially if you use callbacks to login (like facebook). Just provide an easy option in the config (not the ember config) to specify the port.

amreladawy:
I don't think random ports are a good idea as well. I wanted angular-cli to do like ionic-cli. increment the port number until the first available port is found and printed.

bhelzer:
I'd rather have cli configured with my port, otherwise the port of my app would depend on in which order I start my two projects.

amreladawy:
@bhelzer You already have the option to specify the port at your wish

`ng serve --host 0.0.0.0 --port 4201 --live-reload-port 49153`

We are proposing the facility of automatically switch to another port if the port is already taken.

filipesilva:
There is currently a way to get an free port, by using `--port=0`:

```
$ ng serve --port=0
** NG Live Development Server is running on http://localhost:49152. **
# and on another console
$ ng serve --port=0
** NG Live Development Server is running on http://localhost:49153. **
```

manoj30dec:
Thanks @amreladawy ng serve --host 0.0.0.0 --port 4201 --live-reload-port 49153
works for me.

pain0486:
@filipesilva 
Could there be a way to opt into killing the task that is running automatically instead of getting the error? More and more I run into the issue that I closed the window running the app (VSCODE/Powershell/CommandPrompt) or it was closed for me. Then when I go to open it again and run npm start I get Port 4200 is already in use error. 

I used to go into task manager and try to find the correct nodejs.exe and end the process. Now I run this
netstat -ano | findstr :4200 | ConvertFrom-String | ForEach-Object {taskkill /PID $_.P6 /F}
But it would  be nice that if something is already running on that port to just kill it and then run as normal. 

[...]

PR #10012, opened 2018-03-19
docs: add information on special port 0

matthiaslischka:
When port is set to 0 on ng-serve a new free port will be taken.
Was already pointed out by @filipesilva in #1568

BR Matthias


matthiaslischka:
I signed it!

PR #10048, opened 2018-03-22
Health checks and ngtools/webpack move

filipesilva:
…CLI v6


sod:
It moved to https://github.com/angular/devkit/tree/master/packages/ngtools/webpack i guess.

filipesilva:
@sod yes that is the right place. We have already moved it at this point, but needed to delete it from here.

PR #10071, opened 2018-03-26
feat(@angular/cli): update cli config with workspace support

clydin:


sebastien-yoobic:
Hey where can we find info or doc about the workspace config ? We were using custom schematics with ng generate and it does not work anymore. I have seen few commits which changed the way schematics are handled and this seems to be related with workspace.

Thanks !

PR #10101, opened 2018-03-28
fix(@angular/cli): use safer clean-css optimization level

iamchucky:
Target: 1.7.x

### DIFF
--- a/packages/@angular/cli/plugins/cleancss-webpack-plugin.ts
+++ b/packages/@angular/cli/plugins/cleancss-webpack-plugin.ts
@@ -37,9 +37,11 @@ export class CleanCssWebpackPlugin {
 
         const cleancss = new CleanCSS({
           compatibility: 'ie9',
-          // use a safer optimization level
-          // see: https://github.com/jakubpawlowicz/clean-css#optimization-levels
-          level: 1,
+          // disables mergeIntoShorthands due to issue with merged style ordering
+          // see: https://github.com/jakubpawlowicz/clean-css/issues/1022
+          level: {
+            2: { mergeIntoShorthands: false }
+          },
           inline: false,
           returnPromise: true,
           sourceMap: this._options.sourceMap,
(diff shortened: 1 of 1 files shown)


clydin:
Please open an issue following the template guidelines and describing the problem that this would address.

splincode:
Level 1 is more aggressive than 2
Why you usage Level 1?

splincode:
If you are sure of this, you need to write and attach tests

iamchucky:
@clydin please take a look at the issue https://github.com/angular/angular-cli/issues/10186

[Repo to reproduce](https://github.com/iamchucky/angular-cli-clean-css-optimization-issue)

dominikg:
i think disabling all level 2 optimizations is overkill. This sounds like an issue with shorthand merging for non-adjacent properties. [clean-css level2 optimizations](https://github.com/jakubpawlowicz/clean-css#level-2-optimizations) contain some options that read like they control exactly that.

iamchucky:
@dominikg for my use case mentioned in the issue #10186 , yes, setting `mergeIntoShorthands: false` will resolve it.

> Level 2 optimizations operate at rules or multiple properties level, e.g. can remove duplicate rules, remove properties redefined further down a stylesheet, or restructure rules by moving them around.

My original rationale was trying to reduce the surprises of the resulting CSS by using level 1 instead of the more aggressive level 2 as there may be other edge cases that are not just caused by `mergeIntoShorthands`.

I am open to suggestions and would like to know what the team thinks.

clydin:
This appears to be a defect with the `mergeIntoShorthands` optimization.  For cases such as this, we generally will disable the specific rule with a comment explaining why and a link to an issue within the project in question (`clean-css` in this case).  Once the defect is corrected, the minimum version for the package should be increased to the fixed version and disabled rule can then be removed.

@iamchucky Can you update the PR in this regard?

PR #10110, opened 2018-03-29
Fix rc deps

filipesilva:


filipesilva:
@googlebot I can confirm both me and @hansl are ok with our commits being contributed to this project.

PR #10121, opened 2018-03-30
refactor(@angular/cli): detect current project via CWD

clydin:


packages/@angular/cli/utilities/config.ts
```diff
@@ -135,9 +135,26 @@ export function validateWorkspace(json: JsonValue) {
   return true;
 }
 
-export function getProjectByCwd(_workspace: experimental.workspace.Workspace): string | null {
-  // const cwd = process.cwd();
-  // TOOD: Implement project location logic
+export function getProjectByCwd(workspace?: experimental.workspace.Workspace): string | null {
+  if (!workspace) {
```
alfaproject: `workspace = workspace || getWorkspace('local');` if you want to reduce nesting

clydin: Thank you for the review.  In general, I actually prefer the explicitness.  It is also technically slightly more efficient as well (one comparison for the present case as opposed to always two).

alfaproject: I was hoping it short circuits, so should be the same?

clydin: It does short circuit but to do so it needs to do a comparison check on the value.

PR #10200, opened 2018-04-05
PostCSS handling of absolute deployUrl paths

andrewtaylor:
When a `deployUrl` consists of an absolute path, it needs to be prepended to the `outputUrl` since `url.resolve` will just ignore it. The proposed patch simply checks for an absolute path to perform a prepend, falling back to the existing logic otherwise.

Without this patch resources within stylesheets do not get their paths prefixed with the `deployUrl` if for example that `deployUrl` is "/static".

### DIFF
--- a/packages/@angular/cli/plugins/postcss-cli-resources.ts
+++ b/packages/@angular/cli/plugins/postcss-cli-resources.ts
@@ -95,7 +95,11 @@ export default postcss.plugin('postcss-cli-resources', (options: PostcssCliResou
         }
 
         if (deployUrl) {
-          outputUrl = url.resolve(deployUrl, outputUrl);
+          if (deployUrl.startsWith('/')) {
+            outputUrl = deployUrl + outputUrl;
+          } else {
+            outputUrl = url.resolve(deployUrl, outputUrl);
+          }
         }
 
         resourceCache.set(inputUrl, outputUrl);


clydin:
This is operating as intended.  `/static` and `/static/` have two different meanings.  The deployUrl, if used (base href is generally the more appropriate option), should end in a `/` if the last segment represents a directory and not a file.

andrewtaylor:
Setting to `/static/` does fix it. However given the context of a `deployUrl`, it can **only** be a directory. Hence is there not a reasonable expectation that it either warns about a missing slash or adds it itself? It seems other parts of the code do the latter, for example https://github.com/angular/angular-cli/blob/e48d54558f3640920a6adf6793bcf5eb1b9e4070/packages/%40angular/cli/models/webpack-configs/styles.ts#L113.

clydin:
There's a more in-depth discussion on the deployUrl option in this PR: https://github.com/angular/angular-cli/pull/8740.

If the goal is to normalize the deploy URL to always refer to a directory, then similar code to the above linked PR should be placed after this line in the configuration builder: https://github.com/angular/angular-cli/blob/1.7.x/packages/%40angular/cli/models/webpack-config.ts#L41

As an aside, in general, anywhere that someone attempts to use deploy URL, they should be using the base href option instead.  For a comparison, see here: https://github.com/angular/angular-cli/blob/master/docs/design/deployurl-basehref.md

PR #10204, opened 2018-04-06
fix(@angular/cli): allow private schematics on ng add

hansl:
The execution was changed in #10184, but the getOptions() method was forgotten.


filipesilva:
Windows CI is still broken on `ng new`, but I thought that https://github.com/angular/angular-cli/pull/10208 fixed it. @hansl can you rebase on master?

PR #10275, opened 2018-04-10
fix(@angular/cli): dont show Nothing to be done on update

### LINKED ISSUE #10253: Clarify 'nothing to be done' message
### Versions
```
Angular CLI: 6.0.0-rc.2
Node: 9.2.0
OS: linux x64
Angular: 6.0.0-rc.3

@angular/cli: 6.0.0-rc.2
@angular-devkit/architect: 0.5.4
@ngtools/json-schema: 1.2.0
@schematics/update: 0.5.4
typescript: 2.9.0-dev.20180409
webpack: error
```

### Repro steps
Run `ng update` when there is a package listed that needs to be updated

### Observed behavior
```
$ ng update
    We analyzed your package.json, there are some packages to update:
    
      Name                               Version                    Command to update
     ---------------------------------------------------------------------------------------------------------
      @types/node                        6.0.104 -> 9.6.2           npm install @types/node
      jasmine-core                       2.8.0 -> 3.1.0             npm install jasmine-core
      karma-jasmine-html-reporter        0.2.2 -> 1.0.0             npm install karma-jasmine-html-reporter
      protractor                         5.1.2 -> 5.3.1             npm install protractor
      ts-node                            4.1.0 -> 5.0.1             npm install ts-node
    
    
    There might be additional packages that are outdated.
    Or run ng update --all to try to update all at the same time.
    
Nothing to be done.
```

### Desired behavior
The 'Nothing to be done.' message is confusing. 

I recommend either removing it entirely, or changing it to something like 'By default no updates are run automatically, run one of the commands above to start updating'.

hansl:
Fix #10253.

### DIFF
--- a/packages/@angular/cli/commands/update.ts
+++ b/packages/@angular/cli/commands/update.ts
@@ -39,16 +39,13 @@ export default class UpdateCommand extends SchematicCommand {
   }
 
   public async run(options: UpdateOptions) {
-
-    const schematicRunOptions = {
+    return this.runSchematic({
       collectionName: this.collectionName,
       schematicName: this.schematicName,
       schematicOptions: options,
       dryRun: options.dryRun,
       force: false,
-      workingDir: this.project.root,
-    };
-
-    return this.runSchematic(schematicRunOptions);
+      showNothingDone: false,
+    });
   }
 }
--- a/packages/@angular/cli/models/schematic-command.ts
+++ b/packages/@angular/cli/models/schematic-command.ts
@@ -21,6 +21,7 @@ export interface RunSchematicOptions {
   debug?: boolean;
   dryRun: boolean;
   force: boolean;
+  showNothingDone?: boolean;
 }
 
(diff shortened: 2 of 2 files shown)


packages/@angular/cli/commands/update.ts
```diff
@@ -39,16 +39,13 @@ export default class UpdateCommand extends SchematicCommand {
...
   public async run(options: UpdateOptions) {
-
-    const schematicRunOptions = {
+    return this.runSchematic({
       collectionName: this.collectionName,
       schematicName: this.schematicName,
       schematicOptions: options,
       dryRun: options.dryRun,
       force: false,
-      workingDir: this.project.root,
```
filipesilva: Why was `workingDir` removed? It doesn't look related to this fix.

clydin: It's not actually used anywhere in the function.

Issue #10155, opened 2018-04-03
`ng new` installs dependencies before showing created files


filipesilva:
On `6.0.0-rc.0`, creating a new project will first show npm installing dependencies and only after files being created via schematics. This makes it look like `ng new` isn't working until the install is finished.

PR #10278, opened 2018-04-11
fix(@angular/cli): show schematic logging queue every workflow cycle

hansl:


filipesilva:
Changes LGTM, but it has linting errors.

This fixes https://github.com/angular/angular-cli/issues/10155, right?

hansl:
Yes it does. I searched for that particular issue and couldn't find it :)  Thanks!

PR #10338, opened 2018-04-14
fix(@angular/cli): fix xi18n command typo

confraria:
Change the command to extract i18n messages to xi18n instead of xi81n.


confraria:
I signed it!

Issue #730, opened 2016-05-11
trying to use rxjs - getting vendor/rxjs/index.js 404 (Not Found) in browser


alvipeo:
Added to my prj jquery, bootstrap 4, other things and got it working (except for tests - a separate issue). Now, trying to add some state which uses rxjs (Observables, etc) and now getting in a browser (in Chrome, when 'ng serve'):

>>zone.js:101 GET http://localhost:4200/vendor/rxjs/index.js 404 (Not Found)

**Why is it looking for rxjs/index.js ???**

Here's now I added it - folder structure first:
- app
  - shared
    -models  **added**
      - news.model.ts - simple class for model
      - index.ts - (export * from "./news.model";)
- state **added**
      - appState.ts - contains actual code that uses rxjs
      - index.ts - (export * from "./appState";)

and then I just use it in my main component - with import from ... and DI.

Example code for appState.ts:
```
import { Observable, BehaviorSubject } from "rxjs";
import { NewsItem } from "../models";

export interface AppState {
    news: NewsItem[];
}

export class AddNewsAction {
    constructor(public newItem: NewsItem) { }
}

export type Action = AddNewsAction;

export function stateFn(initState: AppState, actions: Observable<Action>): Observable<AppState> {
    const combine = s => ({ news: s[0] });
    const appStateObs: Observable<AppState> = allState(initState.news, actions)
        .map(combine);

    return wrapIntoBehavior(initState, appStateObs);
}

...

```

alvipeo:
Can't really format lists on github, so for clarity - the folder structure:
;; app
;; ;; shared
;; ;; ;; models
;; ;; ;; state

alvipeo:
[The project zip](http://inus.azureedge.net/pub/NewsNotifier.zip?sv=2014-02-14&ss=2016-05-11T21%3A55%3A11Z&se=2016-05-18T21%3A55%3A06Z&sp=r&sr=b&sig=I11kyXECIr0XbY3%2FdXE%2BAucDm7jDjialfoM7wZMEfaY%3D).

Feel free to download.

Should I
```
import { Observable } from "rxjs/Observable";
```
and so on? I tried that but 1) it gets too verbose and 2) this
```
import { map } from "rxjs/operator/map";
```
does not help because I still get the error - map is not a func.

Finally, I tried to import from 'rxjs/Rx' but still getting error for **map**.

BenevidesLecontes:
I'm using rxjs and it's working fine for me. 
[...]

PR #10397, opened 2018-04-19
feat(@angular/cli): build only the projects of a certain type

noelmace:
## Goal feature

Permit to run some architect commands only on projects with a specific projectType.

For example, this will only build the libraries : 
```
ng build --project-type=library
```
angular/devkit#739 would be needed for filtering project names.

## Purpose

Applications and libraries doesn't share the same use-cases : where applications are built (thanks to the build-angular build architect) in order to be served through HTTP, libraries, on the other hand, are bundled by ng-packagr in order to be published to a npm repository. Therefore, deployment / publication processes diverged between the two, so is the CI.

This PR is also a first example of angular/devkit#739 use.

### DIFF
--- a/packages/@angular/cli/commands/build.ts
+++ b/packages/@angular/cli/commands/build.ts
@@ -1,12 +1,15 @@
 import { ArchitectCommand } from '../models/architect-command';
 import { Option, CommandScope } from '../models/command';
 import { Version } from '../upgrade/version';
+import { experimental } from '@angular-devkit/core';
 
 export interface Options {
   project?: string;
   configuration?: string;
+  projectType?: string;
   prod: boolean;
 }
+
 
 export default class BuildCommand extends ArchitectCommand {
   public readonly name = 'build';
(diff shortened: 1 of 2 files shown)


filipesilva:
Heya @noelmace I talked a bit about this problem in https://github.com/angular/devkit/issues/730#issuecomment-383137168, and at the end of my comment I mention this PR and why I think it is not the right approach to this problem at the moment. Let me know what you think.

noelmace:
@filipesilva Thanks @filipesilva for this feedback. We continued to speak (a lot) about this in #730, and I think I see now how we could proceed. Please, see if you can add some feedbacks on https://github.com/angular/devkit/issues/730#issuecomment-384820452.

On the other hand, I still think this option could be relevant for a more restricted use case : we have some set on libraries as a whole, like ``@angular/*``. So, we have to build all of them at the same time, but without building our apps. For that, we can only add some dumb script that just run ``ng build --project liba && ng build --project libb && ....`` (see https://github.com/angular/devkit/issues/730#issuecomment-384821377 or https://github.com/noelmace/ng-admin/blob/master/scripts/publish.ts#L25 for some more sofisticated approach).

For my use case especially, we could also add a --scope option which would permit to just run build targets for projects of a given npm scope /  `projects/` sub-folder. If you think this would be a good feature to add, I could make a PR quickly, but this would require https://github.com/angular/devkit/pull/739. Could you take a look at it please ? Thanks.

filipesilva:
Heya I brought this up with the team and we don't want to do it via filtering through project type. We feel it is a fairly limited way of doing this. 

The need for this sort of thing is still there in the cases you mentioned. I have made https://github.com/angular/angular-cli/issues/11002 to track this. There's other parties with similar needs and we hope to converge on a design that satisfies them.

Issue #10334, opened 2018-04-14
ng server runs ng serve


Toxicable:
### Versions
```
Angular CLI: 6.0.0-rc.4
Node: 8.11.1
OS: linux x64
Angular: 6.0.0-rc.5
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.5.6
@angular-devkit/build-angular     0.5.6
@angular-devkit/build-optimizer   0.5.6
@angular-devkit/core              0.5.6
@angular-devkit/schematics        0.5.6
@angular/cli                      6.0.0-rc.4
@ngtools/json-schema              1.1.0
@ngtools/webpack                  6.0.0-rc.4
@schematics/angular               0.5.6
@schematics/update                0.5.6
rxjs                              6.0.0-uncanny-rc.7
typescript                        2.7.2
webpack                           4.5.0
```

### Repro steps
* ng new my proj
* cd prod 
* yarn ng server

### Observed behavior
`ng serve` is run

### Desired behavior
Nothing should be run.

It's confusing since there's another target called `server`

alan-agius4:
I found this quite confusing as well

PR #10401, opened 2018-04-20
feat(@angular-cli): remove `server` alias for `serve`

alan-agius4:
The alias has been removed as it's quite confusing as one might expect that `ng server` will trigger a server build 

Closes: #10334


filipesilva:
@hansl WDYT?

PR #10404, opened 2018-04-20
fix(@angular/cli): update several config warning messages

alan-agius4:


packages/@angular/cli/lib/init.ts
```diff
@@ -91,7 +91,7 @@ resolve('@angular/cli', { basedir: process.cwd() },
         Your global Angular CLI version (${globalVersion}) is greater than your local
         version (${localVersion}). The local Angular CLI version is used.
 
-        To disable this warning use "ng set --global warnings.versionMismatch=false".
+        To disable this warning use "ng config -g cli.warnings.typescriptMismatch false".
```
alan-agius4: Not sure if this is the correct command to disable the warning

clydin: This should be `cli.warnings.versionMismatch`.   There's only two controllable warnings currently: https://github.com/angular/angular-cli/blob/master/packages/%40angular/cli/lib/config/schema.json#L53

packages/@angular/cli/utilities/check-package-manager.ts
```diff
@@ -19,16 +19,16 @@ export function checkYarnOrCNPM() {
       .then((data: Array<boolean>) => {
         const [isYarnInstalled, isCNPMInstalled] = data;
         if (isYarnInstalled && isCNPMInstalled) {
-          console.log(chalk.yellow('You can `ng set --global packageManager=yarn` '
-            + 'or `ng set --global packageManager=cnpm`.'));
+          console.log(chalk.yellow('You can `ng config -g cli.packageManager \'yarn\'` '
```
clydin: The quotes (single or otherwise) are no longer needed.

For example, `ng config -g cli.packageManager yarn` will work with rc.7+.

alan-agius4:
@clydin requested changed done.

PR #10405, opened 2018-04-20
test: fix unit tests for 'my-lib' is not a known element

alan-agius4:


filipesilva:
I haven't yet fixed this because I think it's a but on devkit. A library shouldn't get an automatic prefix IIRC. 

/cc @Brocco

alan-agius4:
Makes sense. i’ll leave it open the PR until it’s confirmed.

PR #10452, opened 2018-04-24
fix(@angular/cli): Fix library generation bug

Brocco:
This will not pass CI until a devkit release is done that includes this: angular/devkit#766


packages/@angular/cli/utilities/find-up.ts
```diff
@@ -16,7 +18,7 @@ export function findUp(names: string | string[], from: string, stopOnPackageJson
       }
     }
 
-    if (stopOnPackageJson) {
+    if (!isInHomeDir && stopOnPackageJson) {
```
filipesilva: Maybe we should separate these checks. `find-up` itself shouldn't care about home dirs.

PR #10476, opened 2018-04-26
docs(@angular-cli): Clarify the --module option for `generate service`

happycollision:
Add examples for using the `--module` option for `generate service`

I had to hammer at it to figure out how it worked, so I figure that someone else would find this useful.

### DIFF
--- a/docs/documentation/generate/service.md
+++ b/docs/documentation/generate/service.md
@@ -34,6 +34,13 @@
   <p>
     Specifies where the service should be provided.
   </p>
+  <p>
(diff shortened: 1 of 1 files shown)


Brocco:
Thank you for putting together this PR; however, the `module` option is no longer available with the next version of the CLI/schematics. You can see the current options listed [here](https://github.com/angular/devkit/blob/57731deec2cde5b9a6c4e5a5905cfd3b55b94139/packages/schematics/angular/service/schema.json).

Issue #10531, opened 2018-04-30
lint is failing for multiple projects


Tallyb:
### Versions
```
@angular-devkit/architect          0.5.10
@angular-devkit/build-angular      0.5.10
@angular-devkit/build-ng-packagr   0.5.10
@angular-devkit/build-optimizer    0.5.10
@angular-devkit/core               0.5.10
@angular-devkit/schematics         0.5.10
@angular/cli                       6.0.0-rc.7
@ngtools/json-schema               1.1.0
@ngtools/webpack                   6.0.0-rc.8
@schematics/angular                0.5.10
@schematics/update                 0.5.10
ng-packagr                         2.4.2
rxjs                               6.0.0
typescript                         2.7.2
webpack                            4.6.0
```

### Repro steps

run `ng lint`

### Observed behavior
```
Could not determine a single project for the 'lint target.
Error: Could not determine a single project for the 'lint target.
    at LintCommand.getProjectNamesByTarget
```
### Desired behavior
run lint on all defined projects. 

### Mention any other details that might be useful (optional)
I believe it is tracked to this line: 
https://github.com/angular/angular-cli/blob/7f8489c5b328833ef2fe4fccc5e4f749ec6cad06/packages/%40angular/cli/models/architect-command.ts#L20
Multi target is not set to true, although lint is defined as multi target.

krotscheck:
Experiencing the same issue, on RC7, using lint, build, and test. 

```
Could not determine a single project for the 'test target.
```

```
Could not determine a single project for the 'build target.
```

The quote mark before the target name seems suspicious.

PR #10542, opened 2018-04-30
fix(@angular/cli): properly set multi-target command flag

clydin:
Fixes: #10531


shahmirn:
@clydin This doesn't fix the issue with the build command.

build.ts also needs:

this.multiTarget = true;

shahmirn:
This results in an issue where if a 1.x CLI project is updated to 6.x, and then a library is added to it, then the build command fails

I believe it's 'cause when .angular-cli.json is migrated to angular.json, defaultProject is not set.

filipesilva:
@shahmirn I just tested with a new updated project and I think you're right. `defaultProject` does not get set but should.

PR #10557, opened 2018-04-30
test: don't install after generating lib

filipesilva:
It is not needed anymore.


filipesilva:
Superseded by https://github.com/angular/angular-cli/pull/10561

PR #10570, opened 2018-05-01
fix(@angular/cli): do not remove schematics defaults with uppercase

### LINKED ISSUE #10371: Default schematics options with uppercase doesn't work
### Versions
```
cli v6.0.0-rc.5
```

### Repro steps
* `npm install @angular/cli@next`
* `ng new helloworld --inline-template`
* `cd helloworld`
* `ng g generate test`

### Observed behavior
Produce a component with external template.

### Desired behavior
Produce a component with inline template.

I did some investigation:
- `ng g component test --inline-template` works
- following angular/devkit#666, default options are rightly saved in `angular.json`:
```json
"schematics": {
      "inlineTemplate": true
    }
}
```
- adding other options in lowercase like `spec` works:
```json
"schematics": {
      "spec": false
    }
}
```
- adding other options with uppercase letters like `skipImport` doesn't work:
```json
"schematics": {
      "skipImport": true
    }
}
```

So I think the defaults options of `angular.json` with uppercase letters are wrongly transformed/passed to the generation command.

I think it's a quick PR I can do, but I don't know where the problem happen...

cyrilletuzi:
Fixes #10371

@hansl @Brocco @filipesilva @clydin

### DIFF
--- a/packages/@angular/cli/models/schematic-command.ts
+++ b/packages/@angular/cli/models/schematic-command.ts
@@ -7,6 +7,7 @@ import { DryRunEvent, UnsuccessfulWorkflowExecution } from '@angular-devkit/sche
 import { getPackageManager, getDefaultSchematicCollection } from '../utilities/config';
 import { getCollection, getSchematic } from '../utilities/schematics';
 import { getSchematicDefaults } from '../utilities/config';
+import { camelize } from '@angular-devkit/core/src/utils/strings';
 import { take } from 'rxjs/operators';
 import { WorkspaceLoader } from '../models/workspace-loader';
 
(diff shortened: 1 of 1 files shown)


cyrilletuzi:
I don't understand the failure in travis, there's no reason my small change caused this, and I can't reproduce the error on a local project (when doing `ng g component test` with no defaults, the css file is correctly created). So not sure it's related.

Brocco:
Closing this in favor of #10586 due to timing.

Also a cleaner import for the string utilities and an included tests.

Thank you for reporting this and providing this PR which I based my changes off of, totally appreciate it!

PR #10592, opened 2018-05-02
docs: fix doc link, fix hmr configuration

robertbaker:


hansl:
Hi @robertbaker,

2 things; we don't want people to disable warnings by default (HMR is not fully supported), and the link is correct (will point to the proper wiki).

Closing this PR, sorry.

PR #10651, opened 2018-05-04
fix(@angular/cli): remove mangle and compress from server build

### LINKED ISSUE #8616: No apparent way to use `--bundle-dependencies=all` successfully since 1.5.0
Prior to the 1.5.0 release, it was possible to pass `--bundle-dependencies=all --prod` to get an AoT compiled self-contained bundle that worked for server-side rendering (SSR) under Node.js. Since 1.5.0, these flags produce a bundle that, as far as I can tell, can never work because they now rely on the existence of browser DOM APIs that don't exist in Node.

If there is any way to use `--bundle-dependencies=all` successfully in `--prod` build with 1.5.0+, I'd love to know about it.

### Versions
```
Angular CLI: 1.5.4
Node: 6.11.0
OS: win32 x64
Angular: 5.0.2

@angular/cli: 1.5.4
@angular-devkit/core: 0.0.21
@angular-devkit/schematics: 0.0.37
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.4
@schematics/angular: 0.1.7
typescript: 2.4.2
webpack: 3.8.1
```

### Repro steps

* Clone the example repo at https://github.com/SteveSandersonMS/angular-bundledependencies-repro. This is the same as the universal-starter repo, except for [one commit](https://github.com/SteveSandersonMS/angular-bundledependencies-repro/commit/6104ae864b10e2a5d80c95ea836b596d0559c1f8) that gets as close as I can to a scenario where `--bundle-dependencies=all` could work (but still doesn't).
* Run `npm install`
* Observe that it works fine in a non-prod build by running `npm run build:ssr` then `npm run serve:ssr`. If you visit http://localhost:4000, you'll get both server and client rendering.
* Observe that if you build in prod mode (run `npm run build:ssr:prod` then `npm run serve:ssr`), then it does **not** work - you'll get this error:

```
TypeError: StaticInjectorError[InjectionToken DocumentToken]: 
  StaticInjectorError[InjectionToken DocumentToken]: 
    Right-hand side of 'instanceof' is not an object
    at ke (C:\path\angular-bundledependencies-repro\dist\server.js:4106:724114)
    at xe (C:\path\angular-bundledependencies-repro\dist\server.js:4106:723991)
    at lr (C:\path\angular-bundledependencies-repro\dist\server.js:4106:746752)
    at _n.insertToken (C:\path\angular-bundledependencies-repro\dist\server.js:4106:767718)
    at C:\path\angular-bundledependencies-repro\dist\server.js:4106:726646
    at Fe (C:\path\angular-bundledependencies-repro\dist\server.js:4106:726661)
    at I (C:\path\angular-bundledependencies-repro\dist\server.js:4106:721959)
    at Object.parse (C:\path\angular-bundledependencies-repro\dist\server.js:4106:766648)
    at Object.t.createDocument (C:\path\angular-bundledependencies-repro\dist\server.js:4106:1194422)
    at Object.t.createWindow (C:\path\angular-bundledependencies-repro\dist\server.js:4106:1194526)
```

If you trace this into the `server.js` bundle code, what's happening is that it's testing whether `someValue instanceof HTMLElement` (though it's not obvious at first glance because everything is minified and renamed). That code would be OK in a browser, but it's an error in Node, because `HTMLElement` is `undefined`. So the generated bundle isn't going to work for SSR under Node.

* If you want, verify that removing `--bundle-dependencies=all` from lines 22 and 23 in `package.json` is enough to make it work in both dev and prod builds. So the problem is the combination of `--prod` with `--bundle-dependencies=all` - you can use either on their own, but not both together.

### Observed behavior

Fails with the `Right-hand side of 'instanceof' is not an object` error shown above.

### Desired behavior

SSR works as it did in the non-prod build.

The equivalent did work pre-1.5.0 (which was also Angular 4.x).

The reason I'm asking about this is because I'm building new Angular CLI-based ASP.NET templates that will be included in the .NET Core SDK and Visual Studio. We want the template to use `--bundle-dependencies=all` by default because then developers don't have to publish their `node_modules` to production, and attempting to publish 20,000+ files from `node_modules` is extremely time-consuming (and not always reliable) from VS. Plus it tends to hit Windows path-length limitations depending on the deployment target.

hansl:
Using those for server might break the runtime, and since it all runs on the server
we dont need to compress, local name mangling works fine. Also remove the build
optimizer pass on server optimized builds.

Fixes #8616

### DIFF
--- a/packages/@angular/cli/models/webpack-config.ts
+++ b/packages/@angular/cli/models/webpack-config.ts
@@ -37,7 +37,7 @@ export class NgCliWebpackConfig<T extends BuildOptions = BuildOptions> {
     const projectRoot = path.dirname(configPath);
 
     appConfig = this.addAppConfigDefaults(appConfig);
-    buildOptions = this.addTargetDefaults(buildOptions);
+    buildOptions = this.addTargetDefaults(buildOptions, appConfig);
     buildOptions = this.mergeConfigs(buildOptions, appConfig, projectRoot);
 
     const tsconfigPath = path.resolve(projectRoot, appConfig.root, appConfig.tsconfig);
@@ -101,7 +101,7 @@ export class NgCliWebpackConfig<T extends BuildOptions = BuildOptions> {
   }
 
   // Fill in defaults for build targets
-  public addTargetDefaults(buildOptions: T): T {
+  public addTargetDefaults(buildOptions: T, appConfig: any): T {
     const targetDefaults: { [target: string]: Partial<BuildOptions> } = {
       development: {
         environment: 'dev',
@@ -126,8 +126,10 @@ export class NgCliWebpackConfig<T extends BuildOptions = BuildOptions> {
 
     // Use Build Optimizer on prod AOT builds by default when AngularCompilerPlugin is supported.
     const buildOptimizerDefault = {
-      buildOptimizer: buildOptions.target == 'production' && buildOptions.aot !== false
+      buildOptimizer: buildOptions.target == 'production'
+        && buildOptions.aot !== false
         && AngularCompilerPlugin.isSupported()
+        && appConfig.platform != 'server'
     };
 
     merged = Object.assign({}, buildOptimizerDefault, merged);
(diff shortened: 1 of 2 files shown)


hansl:
Build optimizer without uglify compress is not very helpful. Please note that the optimization is not turned off. Build optimizer is a separate plugin that works with uglify.

Memory consumption on server does not matter as much as you'd think, once the files are loaded and compiled. The difference between a 2MB and a 3MB node bundle is minimal. Unless you run on a raspberry pi I presume.

naveedahmed1:
Though this would be a part of 6.0.1 :(

clydin:
It is.  This PR is for the 1.7.x variant of the change.

Issue #10232, opened 2018-04-08
Generated schematic created in unexpected directory


GregOnNet:
### Preface

It seems that the behaviour described below only can be reproduced on Windows machines using `cygwin`.
MINGW seems to work fine.

I was able to reproduce that issue with `6.0.0-rc.1` and `6.0.0-rc.2` of the CLI.

### Versions
```
Angular CLI: 6.0.0-rc.2
Node: 8.9.4
OS: win32 x64
Angular: 6.0.0-rc.3
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 6.0.0-rc.2
@angular-devkit/architect: 0.5.4
@angular-devkit/build-angular: 0.5.4
@angular-devkit/build-optimizer: 0.5.4
@angular-devkit/core: 0.5.4
@angular-devkit/schematics: 0.5.4
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 6.0.0-rc.2.4
@schematics/angular: 0.5.4
@schematics/update: 0.5.4
typescript: 2.7.2
webpack: 4.5.0
```

### Repro steps
* `ng new angularProject`
* `ng m layout` or `ng generate module layout`

### Observed behavior

The schematic is placed in the root directory of the angular project.
Somehow the whole absolute path is recreated inside the project's folder.

![image](https://user-images.githubusercontent.com/444278/38467321-c9fe0cac-3b36-11e8-8944-925958da2a75.png)

The console output indicates that the schematic was created in the right directory.

```
@myMachine /c/workbench/projects/probs
± ng generate module layout
CREATE c/workbench/projects/probs/layout/layout.module.spec.ts (275 bytes)
CREATE c/workbench/projects/probs/layout/layout.module.ts (190 bytes)
```

### Desired behavior

The schematic should be created inside `src/app`.

julianobrasil:
I had problems with Windows too. #10206 worked for me.

hansl:
Hi @GregOnNet,

Because of the low frequency that this happens, I'm not blocking 6.0 final for this issue. I'm keeping it opened and we will fix it in the short future. Thanks!

PR #10778, opened 2018-05-10
fix(@angular/cli): fix PWD in cygwin

filipesilva:
Fix https://github.com/angular/angular-cli/issues/10232


packages/@angular/cli/lib/cli/index.ts
```diff
@@ -41,8 +41,9 @@ function loadCommands() {
 }
 
 export default async function(options: any) {
-  // ensure the environemnt variable for dynamic paths
-  process.env.PWD = path.normalize(process.env.PWD || process.cwd());
+  // Ensure the environment variables for dynamic paths are set.
+  // We can't use the existing process.env.PWD because it can be incorrect in shells like cygwin.
+  process.env.PWD = process.cwd();
```
clydin: This is only used here: https://github.com/angular/angular-cli/blob/master/packages/%40angular/cli/models/schematic-command.ts#L100
If that is changed to use `process.cwd()`, the line here can be removed.  The following line appears completely unused as well.

filipesilva: Done.

filipesilva:
@Brocco done.

PR #10797, opened 2018-05-11
ci: adjust appveyor settings

clydin:


filipesilva:
Note: this is a followup to https://github.com/angular/angular-cli/pull/9996, and in the past we ended up needing to re-add the install every couple of months or so.

Issue #10805, opened 2018-05-11
ng g universal make --clientProject optional


ampgular:
### Versions
```
    _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/


Angular CLI: 6.0.0
Node: 8.11.1
OS: win32 x64
Angular: 6.0.0
... animations, cli, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.6.0
@angular-devkit/build-angular     0.6.0
@angular-devkit/build-optimizer   0.6.0
@angular-devkit/core              0.6.0
@angular-devkit/schematics        0.6.0
@angular/platform-server          6.0.1
@ngtools/webpack                  6.0.0
@schematics/angular               0.6.0
@schematics/update                0.6.0
rxjs                              6.1.0
typescript                        2.7.2
webpack                           4.6.0

### Repro steps

* Step 1; ng n newporject
* Step 2: ng g universal


### Observed behavior
```
'Client app undefined not found'.
```

### Desired behavior
the ng g universal works when adding --clientproject
It would be good that when --clientproject not available schematics will pick de defaultproject in aabgular.json


### Mention any other details that might be useful (optional)
Already Tested locally, only two small changes to do:
in /universal/schema.json delete  "required": [
    "clientProject"
  ]
in /universal/index.ts within function getClientProject
replace 
const  clientProject = workspace.projects[options.clientProject];

with
    if(options.clientProject==undefined)
    {
        options.clientProject = workspace['defaultProject']
    }
 const  clientProject = workspace.projects[options.clientProject]

and that´s all!!

ampgular:
Closed I will post it in @angular-devkit

PR #10806, opened 2018-05-11
release: v6.0.1: make ng g universal --clienproject optional: ISSUE #10805

ampgular:
It would be good that when the option --clientproject is not available 
schematics will pick de defaultproject in aabgular.json


ampgular:
closed I will put it move it to @angular-devkit

PR #10809, opened 2018-05-11
style(@angular/cli): Alphabetized the keys for the angular:component properties object

brianpilati:
style(@angular/cli): Alphabetized the keys for the angular:component properties object

Is there are reason that the keys for the angular:x properties are not alphabetized?

I find it difficult to read when I'm not certain if I should scroll up or down to find a key.

I have alphabetized only the @schematics/angular:component properties' keys to start with on this PR. First, to see if there are objections and second to not make the PR so large that the reviewer can not tell if there is parity.

--Cheers,
Brian

### DIFF
--- a/packages/@angular/cli/lib/config/schema.json
+++ b/packages/@angular/cli/lib/config/schema.json
@@ -74,6 +74,23 @@
         "@schematics/angular:component": {
           "type": "object",
           "properties": {
+            "changeDetection": {
+              "description": "Specifies the change detection strategy.",
+              "enum": ["Default", "OnPush"],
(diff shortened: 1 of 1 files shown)


brianpilati:
I'm not certain what I am missing to cause the failure. Any tips from the gurus will be appreciated.

clydin:
There are three commits in the PR.  Squash the commits into one with a properly formatted commit message.

https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#commit

brianpilati:
Closing due to errors.

PR #10829, opened 2018-05-13
docs: change proxyConfig folder path to root directory

hudasyed:
The directions on line 8 require the creation of the proxy.conf.json file next to the packages.json file which is in the root folder. The error on line 29 pointed the proxyConfig to the src/proxy.conf.json file which does not exist, as that file was placed in the root folder.


hudasyed:
I've signed the CLA.

PR #10870, opened 2018-05-15
docs(@angular/cli): add note to configuration flag

isaacplmann:
Added a note to the configuration flag description in the build, serve
and test wiki pages that mentions that the flag was previously named env


isaacplmann:
Yeah, adding a note there could work.  It might be better to have some centralized place to look up all the flags that changed from 1.7.x to 6.x.  Like an upgrade guide for your npm scripts.

Closing this PR.

PR #10904, opened 2018-05-17
fix(@angular/cli): support codeFrame format with ng lint

maxime1992:
After upgrading to v6 I was not able to run ng lint on my project with the format option set to `codeFrame`.

I believe it's just been forgotten has it's a [valid option on TSLint](https://github.com/palantir/tslint/blob/7199227d683a2c275681a551341c8dfa5d1cdc6e/docs/usage/cli/index.md#cli-usage).


clydin:
Can you change the commit message prefix to a `fix`?

maxime1992:
Sure :+1: done

Issue #10759, opened 2018-05-09
CLI exit code is 0 when quitting for unknown options


delego-wlritchi:
### Versions
```
$ ng version

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 6.0.0
Node: 10.0.0
OS: linux x64
Angular: 6.0.0
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, http, language-service, material, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.6.0
@angular-devkit/build-angular     0.6.0
@angular-devkit/build-optimizer   0.6.0
@angular-devkit/core              0.6.0
@angular-devkit/schematics        0.6.0
@angular/flex-layout              5.0.0-beta.14
@ngtools/webpack                  6.0.0
@schematics/angular               0.6.0
@schematics/update                0.6.0
rxjs                              6.1.0
typescript                        2.7.2
webpack                           4.6.0
```

### Repro steps
Run an Angular CLI command with an unknown option, e.g. `--foo`
```
$ ng build --foo
Unknown option: '--foo'
```
### Observed behavior
The CLI exits (as expected), but with exit code 0 (indicating no error).
```
$ echo $?
0
```

### Desired behavior
The CLI should exit with a nonzero exit code to indicate that the command did not complete successfully. This is particularly important for scripts that call the CLI, especially if these scripts run on a CI server and do not have their output closely monitored. For example, if a CI server is misconfigured and passes an unknown option to `ng test`, the tests will not run at all, but the CI server would incorrectly label the build as passing all tests.

megfault:
When is this fix expected to be merged? This is a rather serious issue for people using CI tools.

I am sorry if this is perceived as impolite. I have no idea of what your time plans are, but I see a fix that is two months old by now, and it's silently letting broken code being deployed in god knows how many repositories of unsuspecting users.

compeek:
[...]

PR #10923, opened 2018-05-18
fix: Change CLI exit code to 1 when quitting for unknown options #10759

jsoref:
This is a proposal based on my vague understanding of what's going on.

One of the various flag changes broke our integration when someone upgraded us to Angular 6.

Our build system assumed that commands would exit 0 when they did their job, or exit non zero when they failed.

Instead, the ng command exited 0, and didn't do its job. Unfortunately, we didn't have enough test coverage to recognize this, and thus a package was created that didn't include its package...


filipesilva:
Actually, `this.logger.fatal` should already cause the process to exit with non-zero code, it's a bug that isn't working. @clydin can you take a look, since you last modified this error logic?

clydin:
Superseded by angular/devkit#1013

PR #10944, opened 2018-05-21
Update proxy.md with Correct Path

jamesqquick:
When adding proxy configuration in angular.json, it appears to be incorrect to keep 'src/' as part of the path to the proxy file.  The documentation says to create a proxy.conf.json file "next to", which I interpret as "in the same directory", the package.json file which resides in the root directory.  Therefore, by including'src/' in the path to the newly proxy file I get an error.  It seems the correct configuration is just 'proxyConfig": "proxy.conf.json"'.


jamesqquick:
I signed it!

clydin:
Thank you for the contribution.  However, this has since been corrected.

PR #10946, opened 2018-05-21
feat(@angular/cli): add eject command

filipesilva:
Blocked by https://github.com/angular/devkit/pull/946 and https://github.com/angular/devkit/pull/947


oluijks:
Is eject coming back soon?

alexeagle:
@IgorMinar we discussed again, we don't plan to add back the `eject` command, instead we will follow the proposal to provide some docs on our wiki and a webpack config or seed repo that shows how to do it.

@filipesilva can you locate the proposal for that?

@StephenFluin as FYI, and to veto if needed

PR #11007, opened 2018-05-25
feat(@angular/cli): support entry component

JLHwung:
See https://github.com/angular/devkit/pull/936


JLHwung:
@clydin Reworded and thanks.

PR #11078, opened 2018-05-31
fix(@angular/cli): change serve fallback config to support Accept header wildcard

davidsmith-accesso:
If a resource request is not present on the server and the Accept header only contains the wildcard / to accept all content, the server responds with a 404 instead of the index.html content.

Testing:
Execute a GET request against any resource not available on the local dev server with the following Accept header:
`Accept: */*`
The server will return a 404 instead of a 200 with the index.html content


davidsmith-accesso:
retest this please

PR #11124, opened 2018-06-06
ci: don't use snapshots on non-master branches

filipesilva:


filipesilva:
This PR is now obsolete.

PR #11128, opened 2018-06-06
fix(@schematics/angular): Fix root path for e2e projects during update

Brocco:


packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -538,7 +538,7 @@ function extractProjectsConfig(
         architect.server = serverTarget;
       }
       const e2eProject: JsonObject = {
-        root: project.root,
+        root: join(project.root as Path, 'e2e'),
```
hansl: normalize paths.

Brocco: The project.root was normalized above, but stored in a JsonObject, I am now reusing the value set above.

PR #11158, opened 2018-06-07
Allow less-loader to use webpack resolver when no includePaths are given

vincentjames501:
- If you look at https://github.com/webpack-contrib/less-loader#less-resolver, it states that
if you provide paths, it will use the less resolver. We should only use the less-resolver when
stylePreprocessorOptions.includePaths are actually given.
- This fixes issues where if in your less file you `@import "~@dependency/less-file";` it actually
resolves correct and when you import something like `@import "some-local-file";` it actually
recompiles when `ng build --watch` is ran.


vincentjames501:
I signed it!

PR #11181, opened 2018-06-09
feat(angular_devkit): stop blocking karma after compilation error

### LINKED ISSUE #11170: Don't block karma on webpack compilation errors
```
- [x] feature request
```

### Area
```
- [x] devkit
- [ ] schematics
```

### Versions

```
$ node --version
v10.1.0
$ npm --version
6.0.1
```

Windows 10

### Repro steps

1. Create a fresh angular project 
1. Open up your karma.conf.js file and set `autoWatch: false`
1. Run karma using `npx karma run` from the `src` directory (in another console, ). Specs should pass. ` Executed 3 of 3 SUCCESS`
1. Now go to the "app.component.ts" file and change `styleUrls: ['./app.component.css']` to `styleUrls: ['']`. 
1. Rerun karma using `npx karma run`. The karma process hangs and is not reporting the error (however, the AngularCompilerPlugin itself does report it).

This can be annoying for users using the Webstorm IDE. It starts angular with `singleRun=false, watch=false`. It uses `karma run` in order to kick of a test run and choose which tests to run. The result is that the tests hang in webstorm:

image

I'm investigating using `ng test` straight up for supporting mutation testing with [Stryker](https://stryker-mutator.io). For this new approach to work, it would help a lot if the normal karma way of reporting errors is used.

-none-

### Desired functionality

The compile error is reported via a karma event. We could use the "normal" error event for this: `'browser_error'`. Otherwise we can create a new one: `'compiler_error'` or something. 

Furthermore, a compiler error should not result in a blocked karma run. Instead, it should finish up the run like normal, using the normal `'run_complete'` event.

If we can fix this issue, we should be able to close https://github.com/angular/angular-cli/issues/8744

nicojs:
This PR removes the blocking of karma in case of compilation errors. This will not influence regular developer workflows (with `ng test` and `ng test --single-run`), but does help plugin creators relying on `karma run` functionality (with `singleRun: false` and `watch: false`).

Fixes #11170

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma-webpack-failure-cb.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma-webpack-failure-cb.ts
@@ -12,12 +12,12 @@
 // Workaround for https://github.com/webpack-contrib/karma-webpack/issues/66
 
 export class KarmaWebpackFailureCb {
-  constructor(private callback: () => void) { }
+  constructor(private callback: (error: string | undefined, errors: string[]) => void) { }
 
   apply(compiler: any): void {
     compiler.hooks.done.tap('KarmaWebpackFailureCb', (stats: any) => {
       if (stats.compilation.errors.length > 0) {
-        this.callback();
+        this.callback(undefined, stats.compilation.errors.map((error: any) => error.message? error.message : error.toString()));
       }
     });
   }
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
@@ -91,8 +91,16 @@ const init: any = (config: any, emitter: any, customFileHandlers: any) => {
     publicPath: '/_karma_webpack_/',
   };
 
-  // Finish Karma run early in case of compilation error.
-  const compilationErrorCb = () => emitter.emit('run_complete', [], { exitCode: 1 });
+  const compilationErrorCb = (error: string | undefined, errors: string[]) => {
+    // Notify potential listeners of the compile error
+    emitter.emit('compile_error', errors);
+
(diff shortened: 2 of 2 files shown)


filipesilva:
LGTM, thanks for improving this! 

By the way, have you found a way to exit karma gracefully in a programmatic way? I haven't really found one that doesn't call `process.exit`.

nicojs:
> By the way, have you found a way to exit karma gracefully in a programmatic way? I haven't really found one that doesn't call process.exit.

No, that isn't possible (that I know of). The only way documented way to stop karma is using `karma.stopper.stop(options, [callback=process.exit])`. This command sends an HTTP request to the karma server (for example: `HTTP GET http://localhost:3456/stop`). 

The callback parameter defaults to `process.exit` so you might think that you can override that behavior. However, this is only the callback received by the *caller*. If you look at the middleware handling the http request, you can see that it always calls `process.kill`:
https://github.com/karma-runner/karma/blob/e8ca4ecefe92bd28494467503b09f3588b787cf7/lib/middleware/stopper.js#L13

In most cases (stryker, ng test, ...) the process hosting the karma server is also the process that would be sending the stop command, so the callback isn't even executed in those cases, as `process.kill` already happened. We designed Stryker around this by sand boxing test runners into separate processes.

Karma deserves some more love from the open source community. Unfortunately, I don't have the time for it.

Btw, when is the next release of the Angular CLI? I can't wait to try it out with Stryker.

filipesilva:
Yeah that's the conclusion I got to as well, just wanted the double check if I was missing anything.

Next release of the CLI and devkit packages should be next week.

nicojs:
> Next release of the CLI and devkit packages should be next week.

I there is a new release: 

image

However, the new code seems to be absent

image

Did something go wrong during release?

nicojs:
I see. It is released under the `@next` tag. Thanks for pointing that out to me @filipesilva, works like a charm!

PR #11196, opened 2018-06-11
feat(@schematics/angular): use es2015 modules overall

filipesilva:
This change helps to keep the behaviour of build/serve/test consistent by using the same import semantics in all of them.


mgol:
Something to check: what about `tsconfig.server.json` when Universal is used? Will it still work correctly with `"module": "es2015"`?

filipesilva:
@mgol good catch, added now.

Issue #8515, opened 2017-11-16
ServiceWorker register generated by cli with base-href don't works


jvitor83:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.
Angular CLI: 1.6.0-beta.0
Node: 8.9.0
OS: win32 x64
Angular: 5.0.1
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router, service-worker

@angular/cli: 1.6.0-beta.0
@angular-devkit/build-optimizer: 0.0.33
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.35
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.0-beta.0
@schematics/angular: 0.1.3
typescript: 2.4.2
webpack: 3.8.1

### Repro steps.
ng new project --service-worker
cd project
ng build --prod --base-href=/project/

then

Publish the dist folder to a web server with a base url /project/
Ex: http://localhost/project/

### The log given by the failure.
Failed to fetch `ngsw-worker.js` at chrome devtools

### Desired functionality.
Fetch `ngsw-worker.js` at chrome devtools and service-worker works!

### Mention any other details that might be useful.
At line 16 of `app.module.ts` generated by cli:
```
    environment.production ? ServiceWorkerModule.register('/ngsw-worker.js') : []
```
the register is starting with `/` which make the fetch start from http://localhost/project/

If is changed to `./` it will works, but i don't test if the first request to a sub-component can confuse this solution, because the first access would be http://localhost/project/somecomponent/, and the register can try to get from http://localhost/project/somecomponent/ngsw-worker.js

My guess is that `ServiceWorkerModule.register` could include the base-path by default, then it could solve this problem definitely.

hansl:
@alxhub Could you investigate/answer this? Seems like using `--base-href` breaks SW.

Ploppy3:
@alxhub We need your help over here!

zmazouzi:
hello @jvitor83 can you help out, 

I have the same problem, I use angular-cli 1.6.7
steps : 
ng new project --service-worker
cd project
ng serve --prod

failed to load service worker to register :4200/ngsw-worker.js:-Infinity

thank you
[
[...]

PR #11131, opened 2018-06-06
refactor: Consolidate adding dependencies

Brocco:

PR #11200, opened 2018-06-11
Fix stuff

gkalpak:
The main fix is:

- fix(@ schematics/angular): allow ServiceWorker to work with baseHref
  (Fixes angular/angular-cli#8515.)

This is a port of angular/devkit#999 from the archived `devkit` repo (minus the stuff already fixed in #11131).
The NGSW parts have been approved by @alxhub [here][1].

[1]: https://github.com/angular/devkit/pull/999#pullrequestreview-126098248

### DIFF
--- a/packages/angular/pwa/pwa/index.ts
+++ b/packages/angular/pwa/pwa/index.ts
@@ -12,7 +12,6 @@ import {
   SchematicsException,
   Tree,
   apply,
-  branchAndMerge,
   chain,
   externalSchematic,
   mergeWith,
@@ -39,15 +38,14 @@ function addServiceWorker(options: PwaOptions): Rule {
 
 function getIndent(text: string): string {
   let indent = '';
-  let hitNonSpace = false;
-  text.split('')
-    .forEach(char => {
-      if (char === ' ' && !hitNonSpace) {
-        indent += ' ';
-      } else {
-        hitNonSpace = true;
-      }
-    }, 0);
+
+  for (const char of text) {
+    if (char === ' ' || char === '\t') {
+      indent += char;
+    } else {
+      break;
+    }
+  }
 
   return indent;
 }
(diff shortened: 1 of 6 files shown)


packages/angular/pwa/pwa/index.ts
```diff
@@ -38,18 +37,8 @@ function addServiceWorker(options: PwaOptions): Rule {
...
-      if (char === ' ' && !hitNonSpace) {
-        indent += ' ';
-      } else {
-        hitNonSpace = true;
-      }
-    }, 0);
-
-  return indent;
+  // This RegExp is guaranteed to match any string (even if it is a zero-length match).
+  return (/^ */.exec(text) as RegExpExecArray)[0];
```
clydin: What about a quick loop instead of the regex (plus tab support):
```
let indent = '';
for (const char of text) {
    if (char === ' ' || char === '\u0009') {
        indent += char;
    } else {
        break;
    }
}
```

packages/angular/pwa/pwa/index.ts
```diff
@@ -70,43 +59,32 @@ function updateIndexFile(options: PwaOptions): Rule {
...
-      .join('\n');
-
-    const bodyTagIndent = getIndent(closingBodyTagLine) + '  ';
-    const itemsToAddToBody
-      = '<noscript>Please enable JavaScript to continue using this application.</noscript>';
-
-    const textToInsertIntoBody = bodyTagIndent + itemsToAddToBody;
+    const bodyIndent = getIndent(lines[closingBodyTagLineIndex]) + '  ';
+    const itemsToAddToBody = [
+      '<noscript>Please enable JavaScript to continue using this application.</noscript>',
```
clydin: Adding this to the actual template `index.html` may make sense since it applies to a standard web application as well.  Thoughts?

packages/angular/pwa/pwa/index.ts
gkalpak: I personally like RegExps, but happy to change it to whatever you prefer (or even keep the original version).
I just happened to go through the files trying to fix something else and made some changes that made sense to me (and hopefully to others).

packages/angular/pwa/pwa/index.ts
gkalpak: True. I guess this was added as poart of `@angular/pwa`, because good PWA practices include having fallback content for when JS is disabled, but this is indeed useful for non-PWA apps as well.
Should Imove it to `index.html`?

packages/angular/pwa/pwa/index.ts
clydin: I'm partial to the more self explanatory nature of the loop above then the other two options.  And it definitely performs better than the first.

packages/angular/pwa/pwa/index.ts
clydin: I think it eventually should but probably should be left to a separate PR.  However, the addition here would need to stay either way to support projects that didn't get created with the new `index.html`.  The code here should be augment now though to check if the `noscript`element is already present and only add if not.  Since someone could have added the element manually as well.

gkalpak: Actually, I just realized that a proper solution should also check whether `link[rel="manifest"]` and `meta[name="theme-color"]` already exists as well. Since this PR is only refactoring that part of the code (no behavior changes), I'd rather leave the checks for another PR.

packages/angular/pwa/pwa/index.ts
gkalpak: To each his own 😛
Happy to change it to a `for` loop.

packages/angular/pwa/pwa/index.ts
clydin: Good point and works for me.

clydin:
@gkalpak looks like there was a rebase error with the last commit.  Its message is `@gkalpak
fixup! refactor(@angular/pwa):`...

PR #11250, opened 2018-06-14
build: fixes needed in google3

alexeagle:
We have an older version of Jasmine that doesn't support generics in this spot

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/action_spec.ts
+++ b/packages/angular_devkit/schematics/src/tree/action_spec.ts
@@ -8,6 +8,7 @@
 import { normalize } from '@angular-devkit/core';
 import { Action, ActionList } from './action';
 
+type ExpectedAction = jasmine.ObjectContaining<Action>;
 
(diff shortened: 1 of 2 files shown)


alexeagle:
Hmm, the fix is not so simple, @qiyigg will have to take this over.

packages/angular_devkit/schematics/src/tree/action_spec.ts
```diff
@@ -87,7 +87,7 @@ describe('Action', () => {
       actions.optimize();
       expect(actions.length).toBe(1);
       expect(actions.get(0)).toEqual(
-        jasmine.objectContaining<Action>({ kind: 'c', path: normalize('/test2') }),
+        jasmine.objectContaining({ kind: 'c', path: normalize('/test2') }),
```
clydin: casting the object literal should fix the errors

alexeagle: :+1:

Issue #11255, opened 2018-06-15
Module not found: Error: Can't resolve 'commonjs-proxy:.


AlexanderKozhevin:
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Area
```
- [ ] devkit
- [ ] schematics
```

node v8.11.1
npm 6.1.0

macOS Sierra

### Repro steps
I just followed example from [official wiki](https://github.com/angular/angular-cli/wiki/stories-create-library) to create new library. But when I tried to use new module or component, I got an error in the console.
I created project from scratch with angular-cli (Angular CLI: 6.1.0-rc.0).
Angular (6.1.0-beta.1)

```
ERROR in ./dist/tvmaze/fesm5/tvmaze.js
Module not found: Error: Can't resolve 'commonjs-proxy:./lib/tvmaze.component' in '/Users/alexanderkozhevin/Desktop/lib-demo/dist/tvmaze/fesm5'
ERROR in ./dist/tvmaze/fesm5/tvmaze.js
Module not found: Error: Can't resolve 'commonjs-proxy:./lib/tvmaze.module' in '/Users/alexanderkozhevin/Desktop/lib-demo/dist/tvmaze/fesm5'
ERROR in ./dist/tvmaze/fesm5/tvmaze.js
Module not found: Error: Can't resolve 'commonjs-proxy:./lib/tvmaze.service' in '/Users/alexanderkozhevin/Desktop/lib-demo/dist/tvmaze/fesm5'
ERROR in ./dist/tvmaze/fesm5/tvmaze.js
Module not found: Error: Can't resolve 'commonjs-proxy:./public_api' in '/Users/alexanderkozhevin/Desktop/lib-demo/dist/tvmaze/fesm5'
ERROR in ./dist/tvmaze/fesm5/tvmaze.js
Module not found: Error: Can't resolve 'commonjs-proxy:./tvmaze.component' in '/Users/alexanderkozhevin/Desktop/lib-demo/dist/tvmaze/fesm5'
ERROR in ./dist/tvmaze/fesm5/tvmaze.js
Module not found: Error: Can't resolve 'commonjs-proxy:@angular/core' in '/Users/alexanderkozhevin/Desktop/lib-demo/dist/tvmaze/fesm5'
ERROR in ./dist/tvmaze/fesm5/tvmaze.js
[...]

PR #11257, opened 2018-06-15
fix(@schematics/angular): add `module` option in library tsconfig

alan-agius4:
When not specified the `commonjs` will be used which will cause erros when consuming the library `Module not found: Error: Can't resolve ' commonjs-proxy:../file`

https://github.com/dherges/ng-packagr/issues/944#issuecomment-397589885

Closes: #11255

//cc @filipesilva

### DIFF
--- a/packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
+++ b/packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "outDir": "<%= relativePathToWorkspaceRoot %>/out-tsc/lib",
     "target": "es2015",
+    "module": "es2015",
     "moduleResolution": "node",
     "declaration": true,
     "sourceMap": true,


devoto13:
`tsconfig.lib.json` extends `tsconfig.json`, where `module` [is set](https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/workspace/files/tsconfig.json) to `es2015`.

I think this is a regression from https://github.com/angular/angular-cli/pull/11196. It moved the `module` property from multiple places to only `tsconfig.json`. So it should work perfectly fine in newly generated CLI workspaces. The problem is that existing workspaces don't have `module: "es2015"` in the root `tsconfig.json` and therefore generating new lib project in the existing workspace leads to the above error.

alan-agius4:
Yes indeed it's a regression from that PR, not sure how to proceed.

alan-agius4:
Shall we add this in the migration similar to this?
https://github.com/angular/angular-cli/pull/11261

Issue #11258, opened 2018-06-15
Building library gives error TS5060: Option 'paths' cannot be used without specifying '--baseUrl' option.


StephenFluin:
### Bug Report or Feature Request (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
```

This bug has hit me a few times during live demos.

### Area
Unsure of source, problem exhibits in CLI.

### Versions
CLI starts on 1.x but moves to 6.x.

/home/stephen/.angular-config.json:

```
{
  "version": 1,
  "cli": {
    "packageManager": "yarn"
  }
}
```

### Repro steps
(use Yarn by default globally)
yarn global add @angular/cli@^1.0.0
ng new my-app
cd my-app
yarn add @angular/angular-cli
ng update @angular/angular-cli
ng update @angular/core
ng generate library my-shared-lib
ng build --project=my-shared-lib

### The log given by the failure
```
$ ng build --project=my-shared-lib
Building Angular Package
Building entry point 'my-shared-lib'
Rendering Stylesheets
Rendering Templates
Compiling TypeScript sources through ngc

BUILD ERROR
error TS5060: Option 'paths' cannot be used without specifying '--baseUrl' option.

Error: error TS5060: Option 'paths' cannot be used without specifying '--baseUrl' option.

    at Object.<anonymous> (/super/workspace/test/update-library-test/node_modules/ng-packagr/lib/ngc/compile-source-files.js:53:68)
    at Generator.next (<anonymous>)
    at /super/workspace/test/update-library-test/node_modules/ng-packagr/lib/ngc/compile-source-files.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/super/workspace/test/update-library-test/node_modules/ng-packagr/lib/ngc/compile-source-files.js:3:12)
    at Object.compileSourceFiles (/super/workspace/test/update-library-test/node_modules/ng-packagr/lib/ngc/compile-source-files.js:19:12)
    at Object.<anonymous> (/super/workspace/test/update-library-test/node_modules/ng-packagr/lib/ng-v5/entry-point/ts/compile-ngc.transform.js:26:32)
    at Generator.next (<anonymous>)
    at /super/workspace/test/update-library-test/node_modules/ng-packagr/lib/ng-v5/entry-point/ts/compile-ngc.transform.js:7:71
    at new Promise (<anonymous>)

error TS5060: Option 'paths' cannot be used without specifying '--baseUrl' option.

[...]

PR #11261, opened 2018-06-15
fix(@schematics/angular): add `baseUrl` in root tsconfig when migrating

alan-agius4:
Closes: #11258 

The problem is that in Angular CLI 1.X.X the root `tsconfig` didn't have `baseUrl`, and when upgrading to v6 the `baseUrl` option is not added. `baseUrl` is  now mandatory especially when creating a library.

//cc @StephenFluin

### DIFF
--- a/packages/schematics/angular/migrations/update-6/index.ts
+++ b/packages/schematics/angular/migrations/update-6/index.ts
@@ -32,7 +32,6 @@ import {
   addPackageJsonDependency,
 } from '../../utility/dependencies';
 import {
-  appendPropertyInAstObject,
   appendValueInAstArray,
   findPropertyInAstObject,
 } from '../../utility/json-utils';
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -724,6 +724,43 @@ function updateTsLintConfig(): Rule {
...
+      findPropertyInAstObject(compilerOptionsAstNode, 'module')
+    ) {
+      return host;
+    }
+
+    const compilerOptions = compilerOptionsAstNode.value;
+    compilerOptions.module = compilerOptions.module || 'es2015';
+    compilerOptions.baseUrl = compilerOptions.baseUrl || './';
+
+    host.overwrite(tsConfigPath, JSON.stringify(tsCfgAst.value, null, 2));
```
alan-agius4: I tired to use the JSON AST for this, but updating multiple properties within the same JSON Object is kinda tricky with `appendPropertyInAstObject` as it kept appending `,` after the last existing property. I also tried to use the `.remove` and `insertLeft`, but it kinda get too much complicated for this trivial thing.

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -724,6 +724,43 @@ function updateTsLintConfig(): Rule {
...
+  return (host: Tree, context: SchematicContext) => {
+    const tsConfigPath = '/tsconfig.json';
+    const buffer = host.read(tsConfigPath);
+    if (!buffer) {
+      return;
+    }
+
+    const tsCfgAst = parseJsonAst(buffer.toString(), JsonParseMode.Loose);
+    if (tsCfgAst.kind != 'object') {
+      throw new SchematicsException('Invalid tsconfig. Was expecting an object');
```
clydin: The message should indicate the root tsconfig is the problematic one.  Also, I'm not sure that this would be fatal.  Maybe log an error/warning and return?  Same for the check below.

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -724,6 +724,43 @@ function updateTsLintConfig(): Rule {
...
+    if (
+      findPropertyInAstObject(compilerOptionsAstNode, 'baseUrl') &&
+      findPropertyInAstObject(compilerOptionsAstNode, 'module')
+    ) {
+      return host;
+    }
+
+    const compilerOptions = compilerOptionsAstNode.value;
+    compilerOptions.module = compilerOptions.module || 'es2015';
+    compilerOptions.baseUrl = compilerOptions.baseUrl || './';
```
clydin: If this is already set and not a variation of the current directory (e.g., "", ".", or "./") then a warning should probably be shown indicating that generated library path mapping will not work.

packages/schematics/angular/migrations/update-6/index.ts
alan-agius4: I think it is fatal since other schematics  such as generate library always extend the root ts config.

What do you think?

clydin: True.  Just the verbiage adjustment then.

alan-agius4:
@clydin PR updated

alan-agius4:
With AppVeyor, I am getting this error quite a lot even in my other PRs

Error: ENOTEMPTY: directory not empty, rmdir
'C:\projects\angular-cli\tests\@angular_devkit\build_angular\test-project-host-hello-world-app-ao7kthpblkg\src'
177
<https://ci.appveyor.com/project/AngularCLI/angular-cli/build/1.0.11959#L177>
at Object.fs.rmdirSync (fs.js:821:3)

packages/schematics/angular/migrations/update-6/index.ts
```diff
@@ -724,6 +725,61 @@ function updateTsLintConfig(): Rule {
...
+    ) {
+      return host;
+    }
+
+    const compilerOptions = compilerOptionsAstNode.value;
+    const { baseUrl = './', module = 'es2015'} = compilerOptions;
+
+    const validBaseUrl = ['./', '', '.'];
+    if (!validBaseUrl.includes(baseUrl as string)) {
+      console.log(terminal.yellow(
```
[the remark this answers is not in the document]
alan-agius4: Apologies about that. PR updated.

PR #11271, opened 2018-06-16
feat(@angular-devkit/build-ng-packagr): move `test.ts` outside of `src`

alan-agius4:
`test.ts` can ube used by the primary and secondary entrypoints, thus it's best to move it outside of the `src` folder. as With secondary entrypoints one would have the following strcuture;

```
- library
-- src
--- lib (feature)
-- secondary (2nd Entry Point)
--- src
```

See: dherges/ng-packagr#900 (comment)

//cc @filipesilva


filipesilva:
I'm unsure of how I feel about this change. It's true that secondary entry points should also use the same `test.ts` file, but that file is still part of the source files and that's why it's within `src/`.

I suppose it's a matter of semantics. One can consider `test.ts` to be part of the tooling, instead of the source files.

Changing the default for libraries begs the question of why the default for apps is not changed too though. Our tooling should support it either way so that shouldn't matter.

@hansl @robwormald thoughts?

alan-agius4:
Yeah, it might makes sense to propagate the change across the board.

I’ll wait for feedback from Rob and Hans, and will update accordingly.

alan-agius4:
Bump!

sumigoma:
@alan-agius4 @filipesilva 
Please forgive me if this is a dumb question, but can either of you (or someone else) explain why the schematic for libraries introduces a `lib` folder under `src` by default? The default structure and configuration in `angular-cli` results in publishable packages with a `lib` folder at the root, rather than a `src` folder like most of the core `@angular` packages (e.g. `@angular/common`). Since this is a library, I find the additional `lib` designation superfluous, whereas having `src` in the resulting bundle with the `.d.ts` files makes sense.

Moreover, shouldn't `ng-package.json` be placed within `src` by default with an entryFile of `index.ts`? This would make it parallel to how secondary entrypoints are currently supported in `ng-packagr` - putting an `ng-package.json` file in each secondary entrypoint folder at the root of `src` seems to produce the desired structure mentioned in https://github.com/dherges/ng-packagr/issues/900 (resembling `@angular/material`).

alan-agius4:
@sumigoma, the `lib` folder is a feature folder to my knowledge. Having no `src`  dir in the releasable package is not a problem at all.  If you want to have something similar to `@angular/material` you'd need to create secondary entrypoints. https://github.com/dherges/ng-packagr/blob/master/docs/secondary-entrypoints.md

sumigoma:
@alan-agius4 yep, I understand that having no `src` folder is not a problem, but to me having a `lib` folder within a library project seems unnecessary. Maybe it's there to enable a structure like 
```
- library
-- src
--- lib (feature)
--- test.ts
--- etc.
```
but based on the conclusion in this thread, if `test.ts` were to be moved out of `src` we wouldn't need the `lib` folder.
```
- library
-- src (feature)
-- test.ts
-- etc.
```
The source for the entire library is in `src`, so do we need the `lib` folder? For secondary endpoints we can structure it like
```
- library
-- src (feature)
--- secondary (2nd entrypoint)
-- test.ts
-- etc.
```
where you don't need another `src` folder since it's all still technically part of the source of this project and under the top `src` folder. I've confirmed that putting an `ng-package.json` in each of these secondary entrypoint folders produces the `@angular/material` structure.

alan-agius4:
@sumigoma, there are many options and architecture on how to structure a library, you can go with the `material`, `angular` or another approach. It's all about what you want to achieve and how your library should be consumed.  

I you want to have a standard library without entrypoints ie feature based, the CLI schematic is awesome to get you started as it offer the feature based folders example;

```
- library
-- src
--- lib  (feature)
--- lib_1 (feature)
--- test.ts
--- etc.
``

If you don't like the `lib` folder you can deleted, but it's there due to the libraries usually have multiple features which are not always secondary entrypoints, similar to `Angular` repo, ex: 

```
- library
-- testing (secondary entrypoint)
-- src
--- lib  (feature)
--- lib_1 (feature)
--- test.ts
--- etc.
```

Also, the output structure is effected by where the `public_api.ts` file is placed, as if it is placed outside of the `src`, this last folder will be emitted in the `dist`.

But I guess the placement of the `public_api` is another thing, if it should be under `src` or outside.

**Comment by sumigoma** [comment]
Thanks for the clarification @alan-agius4. I will certainly change `angular.json` to suit my personal needs, but I am just questioning the default structure that `angular-cli` produces. 

https://github.com/angular/angular/tree/master/packages/common has multiple feature folders under `src` as you mention, but nowhere have I ever seen a folder called `lib`, ever, in any of the angular libraries. Instead, I've seen structures like what you describe:
e.g.
```
- library
-- src
--- feat1 (the name of the feature is here, not "lib")
--- feat2
```
Maybe the thinking behind putting a `lib` folder under `src` by default was to *urge people who want multiple features to have to rename it*, but in that case I think it'd make more sense for it to create a `feature1` folder, to encourage renaming and to guide its usage. This isn't a roadblock since the cli is flexible to support any organizational structure as you say (through placement of `public-api.ts` and changing `angular.json`) - I'm just wondering if this is the optimal structure for most library developers.

hansl:
I'm going to close this as it is too close to RC date and we probably don't want to start a debate about where a file should be. There is a task in Jira to rethink the file structure of default CLI projects and it should be part of those discussions. Thanks!

PR #11274, opened 2018-06-16
Schematic Tree fixes

clydin:
Based on #11178

### DIFF
--- a/packages/angular_devkit/core/src/virtual-fs/host/memory.ts
+++ b/packages/angular_devkit/core/src/virtual-fs/host/memory.ts
@@ -33,6 +33,7 @@ import {
   Stats,
 } from './interface';
 
+
 export interface SimpleMemoryHostStats {
   readonly content: FileBuffer | null;
 }
@@ -210,6 +211,23 @@ export class SimpleMemoryHost implements Host<{}> {
     } else {
       const content = this._cache.get(from);
       if (content) {
+        const fragments = split(to);
+        const newDirectories = [];
+        let curr: Path = normalize('/');
+        for (const fr of fragments) {
+          curr = join(curr, fr);
+          const maybeStats = this._cache.get(fr);
+          if (maybeStats) {
+            if (maybeStats.isFile()) {
+              throw new PathIsFileException(curr);
+            }
+          } else {
+            newDirectories.push(curr);
+          }
+        }
+        for (const newDirectory of newDirectories) {
+          this._cache.set(newDirectory, this._newDirStats());
+        }
         this._cache.delete(from);
         this._cache.set(to, content);
       }
--- a/packages/angular_devkit/schematics/src/tree/filtered.ts
+++ b/packages/angular_devkit/schematics/src/tree/filtered.ts
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import { DelegateTree } from './delegate';
 import { FilePredicate, Tree } from './interface';
 import { VirtualTree } from './virtual';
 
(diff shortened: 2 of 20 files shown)


packages/angular_devkit/core/src/virtual-fs/host/memory.ts
```diff
@@ -211,6 +212,20 @@ export class SimpleMemoryHost implements Host<{}> {
       const content = this._cache.get(from);
       if (content) {
         this._cache.delete(from);
```
hansl: This side effect should be moved to the end of the function (next to the set) to avoid inconsistent state if an exception is thrown.

packages/angular_devkit/schematics/src/tree/filtered.ts
```diff
@@ -38,3 +39,15 @@ export class FilteredTree extends VirtualTree {
...
   }
 }
+
+export class FilterTree extends DelegateTree {
+  constructor(tree: Tree, filter: FilePredicate<boolean> = () => true) {
+    super(tree.branch());
+
+    tree.visit(path => {
+      if (!filter(path)) {
+        this.delete(path);
```
hansl: This will delete files (as actions), no? A FilterTree should contains only actions that are pertained to files that match the pattern, unless you intend to do something else here. But if I have a tree that contains file1 and file2 and return FilterTree(x => x != file2), it would actually delete file2 which is probably not what we mean (and not what FilteredTree would have done).

I don’t see a test on actions of filtertree which would have caught this.

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -98,6 +96,7 @@ export class HostTree implements Tree {
   private _id = _uniqueId++;
```
hansl: I don’t see this part being fixed from my previous PR; IDs need to be negative to avoid ID clash with VirtualTree.

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -426,3 +404,17 @@ export class HostTree implements Tree {
     );
   }
 }
+
+export class HostCreateTree extends HostTree {
+  constructor(host: virtualFs.Host) {
```
hansl: Should be the same host type as it’s superclass (ReadonlyHost<{}>)

packages/angular_devkit/schematics/src/tree/static.ts
```diff
@@ -16,34 +16,41 @@ export function empty() {
 }
 
 export function branch(tree: Tree) {
```
hansl: Please add TODOs to these functions to remove those VirtualTree hacks for 7.0

clydin: Done.

packages/angular_devkit/schematics/src/tree/host-tree.ts
clydin: Done.

packages/angular_devkit/schematics/src/tree/host-tree.ts
clydin: Done.

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
clydin: Done.

packages/angular_devkit/schematics/src/tree/filtered.ts
clydin: New host tree specific filter added that does not use delete.

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -426,3 +404,68 @@ export class HostTree implements Tree {
...
+    });
+  }
+}
+
+export class FilterHostTree extends HostTree {
+  constructor(tree: HostTree, filter: FilePredicate<boolean> = () => true) {
+    const newBackend = new virtualFs.SimpleMemoryHost();
+    // cast to allow access
+    const originalBackend = (tree as FilterHostTree)._backend;
+    originalBackend.list(normalize('/'))
```
hansl: Are you just filtering the root?

clydin: Indeed.  Fixed with additional tests.

Issue #11301, opened 2018-06-19
`npm start` does not work when there is a library and `*` path mapping to resolve packages from root `node_modules`


peterbakonyi05:
### Bug Report or Feature Request (mark with an `x`)
```
- [x ] bug report -> please search issues before submitting
- [ ] feature request
```

### Area
```
- [x ] devkit
- [ ] schematics
```

### Versions
node v8.11.1
npm 5.6.0
Windows 10

### Repro steps
1. create a new app with angular cli
2. add a library
3. add extra path mapping to `tsconfig.json`:
```
"*": [
     "node_modules/*"
]
```
4. run `npm start` and notice the error

### The log given by the failure
```
ERROR in ./src/app/app.module.ts
Module not found: Error: Can't resolve 'ngx-test' in 'C:\Github\ngx-library-npm-link\src\app'
ℹ ｢wdm｣: Failed to compile.
```

### Desired functionality
`npm start` works. Also note that building directly with `ngc` it works correctly.

There can be other angular libraries that need to be npm linked during development because they are not in the same repo. In that case, the common dependencies (@angular, @ngrx etc...) need to be resolved from the root node_modules folder to avoid bundling and resolving the dependencies from different locations.

### Mention any other details that might be useful
Library with minimal reproduction:
https://github.com/peterbakonyi05/ngx-library-npm-link

alan-agius4:
@Brocco this issue is not related to `ng-packagr` but rather `ngtools/webpack` check #11304

PR #11304, opened 2018-06-19
fix(@ngtools/webpack): fixes module resolution when using `*` token

alan-agius4:
At the moment module resolution is done via typescript when there is a single `*` token. For maintainability purposes it's best to resolve paths via typescript module resolution strategy. 

Module resolution now is solely done via typescript when there are paths in tsconfig .

Closes: #11301

//cc @filipesilva, I know you usually working in this, can you help me out please, where do you usually add the tests, I cannot find any :(

### DIFF
--- a/packages/ngtools/webpack/src/paths-plugin.ts
+++ b/packages/ngtools/webpack/src/paths-plugin.ts
@@ -12,15 +12,15 @@ import {
   NormalModuleFactoryRequest,
 } from './webpack';
 
-
 export function resolveWithPaths(
   request: NormalModuleFactoryRequest,
   callback: Callback<NormalModuleFactoryRequest>,
   compilerOptions: ts.CompilerOptions,
   host: ts.CompilerHost,
   cache?: ts.ModuleResolutionCache,
+  dtsModuleResolutionCache?: Map<string, string>,
 ) {
-  if (!request || !request.request || !compilerOptions.paths) {
+  if (!request || !request.request) {
     callback(null, request);
 
     return;
@@ -49,69 +49,9 @@ export function resolveWithPaths(
     return;
   }
 
-  // check if any path mapping rules are relevant
-  const pathMapOptions = [];
-  for (const pattern in compilerOptions.paths) {
-      // can only contain zero or one
-      const starIndex = pattern.indexOf('*');
-      if (starIndex === -1) {
-        if (pattern === originalRequest) {
-          pathMapOptions.push({
-            partial: '',
-            potentials: compilerOptions.paths[pattern],
-          });
-        }
-      } else if (starIndex === 0 && pattern.length === 1) {
-        pathMapOptions.push({
-          partial: originalRequest,
-          potentials: compilerOptions.paths[pattern],
-        });
-      } else if (starIndex === pattern.length - 1) {
-        if (originalRequest.startsWith(pattern.slice(0, -1))) {
-          pathMapOptions.push({
-            partial: originalRequest.slice(pattern.length - 1),
-            potentials: compilerOptions.paths[pattern],
-          });
-        }
-      } else {
-        const [prefix, suffix] = pattern.split('*');
-        if (originalRequest.startsWith(prefix) && originalRequest.endsWith(suffix)) {
-          pathMapOptions.push({
-            partial: originalRequest.slice(prefix.length).slice(0, -suffix.length),
-            potentials: compilerOptions.paths[pattern],
-          });
-        }
-      }
-  }
-
-  if (pathMapOptions.length === 0) {
-    callback(null, request);
-
-    return;
-  }
-
-  if (pathMapOptions.length === 1 && pathMapOptions[0].potentials.length === 1) {
-    const onlyPotential = pathMapOptions[0].potentials[0];
-    let replacement;
-    const starIndex = onlyPotential.indexOf('*');
-    if (starIndex === -1) {
-      replacement = onlyPotential;
-    } else if (starIndex === onlyPotential.length - 1) {
-      replacement = onlyPotential.slice(0, -1) + pathMapOptions[0].partial;
-    } else {
-      const [prefix, suffix] = onlyPotential.split('*');
-      replacement = prefix + pathMapOptions[0].partial + suffix;
-    }
-
-    request.request = path.resolve(compilerOptions.baseUrl || '', replacement);
-    callback(null, request);
-
-    return;
-  }
-
   const moduleResolver = ts.resolveModuleName(
     originalRequest,
-    request.contextInfo.issuer,
+    request.contextInfo.issuer.replace(/\\/g, '/'),
     compilerOptions,
     host,
     cache,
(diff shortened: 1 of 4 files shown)


alan-agius4:
Need to look why karma still fails when using 
```json
      "*": [
        "node_modules/*"
      ]
```

alan-agius4:
I think now the e2e is failing and rightly so!

```bash
Module build failed: Error: /tmp/angular-cli-e2e-118519-141-12iaj2q.ncmw/test-project/node_modules/@firebase/polyfill/index.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
The missing file seems to be part of a third party library. TS files in published libraries are often a sign of a badly packaged library. Please open an issue in the library repository to alert its author and ask them to package the library using the Angular Package Format (https://goo.gl/jB3GVv).
    at AngularCompilerPlugin.getCompiledFile (/tmp/angular-cli-e2e-118519-141-12iaj2q.ncmw/test-project/node_modules/@ngtools/webpack/src/angular_compiler_plugin.js:730:23)
    at plugin.done.then (/tmp/angular-cli-e2e-118519-141-12iaj2q.ncmw/test-project/node_modules/@ngtools/webpack/src/loader.js:41:31)
```

tests/legacy-cli/e2e/tests/misc/module-resolution.ts
```diff
@@ -12,10 +12,19 @@ export default async function () {
   });
   await ng('build');
 
+  await updateJsonFile('tsconfig.json', tsconfig => {
+    tsconfig.compilerOptions.paths = {
+      '@app/*': ['*'],
```
alan-agius4: Are these actually being create and consumed? I cannot find any reference to them anywhere.

I copied the paths from a test within the same file, but I have doubts they are actually correct.

tests/legacy-cli/e2e/tests/misc/module-resolution.ts
```diff
@@ -38,7 +46,7 @@ export default async function () {
   await ng('build', '--aot');
   await ng('test', '--watch=false');
 
-  await silentNpm('install', 'firebase@4.9.0');
+  await silentNpm('install', 'firebase@4.13.1');
```
alan-agius4: In `4.9.0` firebase used to publish typescript files in their dist. Which causes the below error;
```
Module build failed: Error: /tmp/angular-cli-e2e-118519-141-12iaj2q.ncmw/test-project/node_modules/@firebase/polyfill/index.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
The missing file seems to be part of a third party library. TS files in published libraries are often a sign of a badly packaged library. Please open an issue in the library repository to alert its author and ask them to package the library using the Angular Package Format (https://goo.gl/jB3GVv).
    at AngularCompilerPlugin.getCompiledFile (/tmp/angular-cli-e2e-118519-141-12iaj2q.ncmw/test-project/node_modules/@ngtools/webpack/src/angular_compiler_plugin.js:730:23)
    at plugin.done.then (/tmp/angular-cli-e2e-118519-141-12iaj2q.ncmw/test-project/node_modules/@ngtools/webpack/src/loader.js:41:31)
```

https://unpkg.com/@firebase/polyfill@0.1.6/index.ts

This was fixed with this PR: https://github.com/firebase/firebase-js-sdk/pull/695

tests/legacy-cli/e2e/tests/misc/module-resolution.ts
```diff
@@ -60,12 +68,12 @@ export default async function () {
       '@firebase/polyfill': ['@firebase/polyfill/index.ts'],
     };
   });
-  await expectToFail(() => ng('build'));
+  await ng('build');
```
alan-agius4: these shouldn't fail as typescript is able to resolve these even with wrong paths. Since they are under `node_modules`

alan-agius4:
CI green apart from the usual AppVeyor recurring error.

alan-agius4:
@clydin, do you mind if you re-trigger the CI? Thanks

packages/ngtools/webpack/src/paths-plugin.ts
```diff
@@ -12,15 +12,15 @@ import {
...
 export function resolveWithPaths(
   request: NormalModuleFactoryRequest,
   callback: Callback<NormalModuleFactoryRequest>,
   compilerOptions: ts.CompilerOptions,
   host: ts.CompilerHost,
   cache?: ts.ModuleResolutionCache,
+  dtsModuleResolutionCache?: Map<string, string>,
 ) {
-  if (!request || !request.request || !compilerOptions.paths) {
+  if (!request || !request.request) {
```
alan-agius4: Note to reviewer, why should module resolution for typescript files be different from the standard tsc compiler resolution strategy.

packages/ngtools/webpack/src/paths-plugin.ts
```diff
@@ -49,69 +49,9 @@ export function resolveWithPaths(
...
-    request.request = path.resolve(compilerOptions.baseUrl || '', replacement);
-    callback(null, request);
-
-    return;
-  }
-
   const moduleResolver = ts.resolveModuleName(
     originalRequest,
-    request.contextInfo.issuer,
+    request.contextInfo.issuer.replace(/\\/g, '/'),
```
alan-agius4: The `replace` is required for Windows

tests/legacy-cli/e2e/tests/misc/module-resolution.ts
clydin: That version is intentionally chosen for the reason described.  The bundler's package resolution to determine the actual javascript file to use within the application should still function in this case.

packages/ngtools/webpack/src/paths-plugin.ts
clydin: Typescript resolves packages for either TypeScript source or type definitions.  When bundling, the actual executable javascript for the given package is required.

tests/legacy-cli/e2e/tests/misc/module-resolution.ts
alan-agius4: So, ngtools/webpack should have a different resolution than ngc to cater for such instances?

clydin: It's also important that the appropriate bundle format is chosen from the `package.json` (FESM2015, ESM2015, etc.) which is something that TypeScript has no knowledge when resolving.

alan-agius4:
I got you! Thanks @clydin  for your time into reviewing this and the feedback.

Though, I am still a bit puzzled, what is the reason behing why a resolution of a
package that fails in ngc should not fail in webpack and vice-versa.

PS: I think there is a bug here though that I noticed while doing this; https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/paths-plugin.ts

If the path of the ‘package.json’ exists shouldn’t the dirname by passed back to webpack?

tests/legacy-cli/e2e/tests/misc/module-resolution.ts
alan-agius4: The resolution itself of the module formats still happends via webpack, as to webpack I am passing the directory that contains the package.json and only fallback to the js file if it doesn’t exists. (Same as current basicilly)

alan-agius4:
Overall I do see some your points @clydin and I agree with you, main reason is that in some cases the declarations files might not be in the same dir of the package.json like for the APF. So finding the package.json directory and pass it to webpack so that the later is able to load the proper module format such as fesm, will be a huge issue.

clydin:
Yes.  that last point is the most critical.  As to `ngc`, it is essentially `tsc` will additional transformers attached to create the necessary factory files.  But otherwise it is similar in respects to `tsc` in that it doesn't need to resolve the actual executable javascript and bundle the final product.

I put together a PR (#11351) to improve the algorithm to more closely match TypeScripts.  The next step would be to create a full resolver plugin to allow for multiple resolve attempts using the webpack resolution system.

Issue #11343, opened 2018-06-22
ng g library command should able to skip install dependencies


cashastri2016:
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Area
```
- [ ] devkit
- [x] schematics
```

### Versions

Angular CLI: 6.0.8
Node: 10.1.0
OS: darwin x64
Angular: 
... 

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.6.8
@angular-devkit/core         0.6.8
@angular-devkit/schematics   0.6.8
@schematics/angular          0.6.8
@schematics/update           0.6.8
rxjs                         6.2.0
typescript                   2.7.2

### Repro steps
ng g library --skip-install

### Desired functionality
CLI should not install dependencies

PR #11345, opened 2018-06-22
feat(@schematics/angular): add `skipInstall` option to `library` sche…

alan-agius4:
add `skipInstall` option to `library` schematic

Closes: #11343


filipesilva:
Thanks for this update 👍

Issue #11067, opened 2018-05-30
ng lint --format junit is missing


DaveXCS:
### Versions
```
Angular CLI: 6.0.5
Node: 10.2.1
OS: win32 x64
Angular: 6.0.3
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.6.5
@angular-devkit/build-angular     0.6.5
@angular-devkit/build-optimizer   0.6.5
@angular-devkit/core              0.6.5
@angular-devkit/schematics        0.6.5
@angular/cli                      6.0.5
@ngtools/webpack                  6.0.5
@schematics/angular               0.6.5
@schematics/update                0.6.5
rxjs                              6.2.0
typescript                        2.7.2
webpack                           4.8.3
```

### Repro steps
* ng lint myproject --format junit

### Observed behavior
```
Schema validation failed with the following errors:
  Data path ".format" should be equal to one of the allowed values.
```

### Desired behavior
junit is one of the official supported formatters and needed as output for some ci server:
https://palantir.github.io/tslint/formatters/

greghopkins:
Having same issue. 👍

beqa7137:
What should we change in our project to make `ng lint` work? (6.0.8)

PR #11349, opened 2018-06-22
Tslint output formatter fixes

clydin:
Also alphabetizes the list.

Fixes: #11067
Fixes: #11361

### DIFF
--- a/packages/angular_devkit/build_angular/src/tslint/schema.json
+++ b/packages/angular_devkit/build_angular/src/tslint/schema.json
@@ -1,4 +1,5 @@
 {
+  "$schema": "http://json-schema.org/draft-07/schema",
   "title": "TSlint Target",
   "description": "TSlint target options for Build Facade.",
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/tslint/schema.json
```diff
@@ -43,16 +44,24 @@
...
-        "prose",
-        "json",
-        "stylish",
-        "verbose",
-        "pmd",
-        "msbuild",
-        "checkstyle",
-        "vso",
-        "fileslist"
+      "anyOf": [
```
filipesilva: What's the difference between just having the `enum` and having `anyOf` with the `enum` and `minLength`?

clydin: The second allows specifying custom formatters.  TSLint will try to load a formatter if it is not one of the builtin options.

PR #11350, opened 2018-06-22
docs(@angular/cli): add commented `Object.setPrototypeOf` polyfills

everedifice:
since rxjs v6.0.0-alpha.1 IE10 and below need `Object.setPrototypeOf` polyfill.
change polyfill.ts for that.

related issue: https://github.com/angular/angular/issues/24488

### DIFF
--- a/packages/schematics/angular/application/files/src/polyfills.ts
+++ b/packages/schematics/angular/application/files/src/polyfills.ts
@@ -40,6 +40,25 @@
 /** IE10 and IE11 requires the following for the Reflect API. */
 // import 'core-js/es6/reflect';
 
+/** IE9 and IE10 requires the following polyfill for using rxjs 6.0 */
+// (function() {
+//   Object.setPrototypeOf = Object.setPrototypeOf || ({__proto__: []} instanceof Array ? setProtoOf : mixinProperties);
+//
+//   function setProtoOf(obj, proto) {
+//     obj.__proto__ = proto;
+//     return obj;
+//   }
+//
+//   function mixinProperties(obj, proto) {
+//     for (const prop in proto) {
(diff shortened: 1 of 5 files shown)


packages/schematics/angular/application/files/src/polyfills.ts
```diff
@@ -40,6 +40,25 @@
 /** IE10 and IE11 requires the following for the Reflect API. */
 // import 'core-js/es6/reflect';
 
+/** IE9 and IE10 requires the following polyfill for using rxjs 6.0 */
```
alan-agius4: This polyfill is already included when you include the `IE9`, `IE10` and `IE11` polyfills above.

https://github.com/everedifice/angular-cli/blob/master/packages/schematics/angular/application/files/src/polyfills.ts#L23

everedifice:
Background:

Since rxjs@6.0.0 IE10 need `Object.setPrototypeOf` polyfill. You can see that change in [here](https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#breaking-changes-7)

rxjs use `Object.setPrototypeOf` at

- https://github.com/ReactiveX/rxjs/blob/d0ea152f3c5575455a73cc43b8fad6a750d24208/src/internal/util/ArgumentOutOfRangeError.ts#L17
- https://github.com/ReactiveX/rxjs/blob/d0ea152f3c5575455a73cc43b8fad6a750d24208/src/internal/util/EmptyError.ts#L17

Angular depends on core-js polyfill. but the polyfill need IE11+. you can check that [here](https://github.com/zloirock/core-js#ecmascript-object)

So if Angular want to support IE10. need other polyfill.

PR #11351, opened 2018-06-22
Improve path mapping support

clydin:
Fixes: #11301

### DIFF
--- a/packages/ngtools/webpack/src/paths-plugin.ts
+++ b/packages/ngtools/webpack/src/paths-plugin.ts
@@ -52,36 +52,47 @@ export function resolveWithPaths(
   // check if any path mapping rules are relevant
   const pathMapOptions = [];
   for (const pattern in compilerOptions.paths) {
-      // can only contain zero or one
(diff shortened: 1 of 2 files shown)


packages/ngtools/webpack/src/paths-plugin.ts
```diff
@@ -109,6 +131,9 @@ export function resolveWithPaths(
     return;
   }
 
+  // TODO: The following is used when there is more than one potential and will not be
+  //       needed once this is turned into a full webpack resolver plugin
+
```
alan-agius4: Shouldn’t we return the dirname when package.jaon exists?

alan-agius4:
Veru nice 😀

packages/ngtools/webpack/src/paths-plugin.ts
clydin: Would you mind opening a followup PR with your suggested changes once this is merged?

alan-agius4: Sure thing, I can do that 😊

Issue #11100, opened 2018-06-04
Build with --watch flag does not work with a lib project


zbarbuto:
### Versions
```
Angular CLI: 6.0.7
Node: 9.6.1
OS: darwin x64
Angular: 6.0.3
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.6.5
@angular-devkit/build-angular      0.6.5
@angular-devkit/build-ng-packagr   0.6.5
@angular-devkit/build-optimizer    0.6.5
@angular-devkit/core               0.6.5
@angular-devkit/schematics         0.6.5
@angular/cdk                       6.1.0
@angular/cli                       6.0.7
@angular/material                  6.1.0
@ngtools/json-schema               1.1.0
@ngtools/webpack                   6.0.5
@schematics/angular                0.6.5
@schematics/update                 0.6.7
ng-packagr                         3.0.0
rxjs                               6.2.0
typescript                         2.7.2
webpack                            4.8.3
```

### Observed behavior

Running `ng build --watch` works perfectly fine on the default app in a monorepo with the default app. 

However, running `ng build mylib --watch` or `ng build --watch mylib` after generating a lib with the CLI gives `Unknown option: '--watch'`

### Desired behavior

Can build any of the apps or libs in an Angular cli monorepo in --watch mode.

### Mention any other details that might be useful (optional)

Have tried various combinations of ordering the arguments as well as `--watch=true` and `-w`. None seem to work.

elclanrs:
This is not supported yet. I'm eagerly awaiting for [this PR](https://github.com/dherges/ng-packagr/pull/884) to be merged in ng-packagr.

gabrielalan:
So they launch a new CLI breaking one of the commands? 😂

jrista:
Any idea on when this might be available? We are doing a lot of work with libs at the moment, no ability to watch is really rough...

alan-agius4:
This should become available with Angular-CLI v6.2

bezysoftware:
I installed the latest 6.2.0 beta.0, but am still unable to do `ng build --watch` for a library ("unknown switch"). Here's outcome of `ng --version`:

```
Angular CLI: 6.2.0-beta.0
Node: 8.11.3
OS: win32 x64
Angular: 6.1.0
... common, compiler, compiler-cli, core, platform-browser
[...]

PR #11358, opened 2018-06-24
feat: add support for `watch` when building a library

alan-agius4:
add support for `watch` when building a library

`ng-packagr` version `4.0.0-rc.2`, lands the incremental builds feature.

More info:  https://github.com/dherges/ng-packagr/blob/master/CHANGELOG.md#400-rc2-2018-06-23

BREAKING CHANGE:
- `ng-packagr` version `4.0.0-rc.2` or greater is required.
- `enableResourceInlining` needs to be enabled for libraries that contain components

Ps: most likely this should go in V7, but still felt like doing it, so to be ready.

Closes: #11100

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7466,12 +7466,50 @@
         "chalk": {
           "version": "2.4.1",
           "resolved": "https://registry.npmjs.org/chalk/-/chalk-2.4.1.tgz",
-          "integrity": "sha512-ObN6h1v2fTJSmUXoS3nMQ92LbDK9be4TV+6G+omQlGJFdcUX5heKi1LZ1YnRMIgwTLEj3E24bT6tYni50rlCfQ==",
+          "integrity": "sha1-GMSasWoDe26wFSzIPjRxM4IVtm4=",
           "requires": {
             "ansi-styles": "^3.2.1",
             "escape-string-regexp": "^1.0.5",
             "supports-color": "^5.3.0"
           }
+        },
+        "find-up": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/find-up/-/find-up-3.0.0.tgz",
+          "integrity": "sha1-SRafHXmTQwZG2mHsxa41XCHJe3M=",
+          "requires": {
+            "locate-path": "^3.0.0"
+          }
+        },
+        "locate-path": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/locate-path/-/locate-path-3.0.0.tgz",
+          "integrity": "sha1-2+w7OrdZdYBxtY/ln8QYca8hQA4=",
+          "requires": {
+            "p-locate": "^3.0.0",
+            "path-exists": "^3.0.0"
+          }
+        },
+        "p-limit": {
+          "version": "2.0.0",
+          "resolved": "https://registry.npmjs.org/p-limit/-/p-limit-2.0.0.tgz",
+          "integrity": "sha1-5iTtVO6MRgp3izyfNnBJb/ileuw=",
+          "requires": {
+            "p-try": "^2.0.0"
+          }
+        },
+        "p-locate": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/p-locate/-/p-locate-3.0.0.tgz",
+          "integrity": "sha1-Mi1poFwCZLJZl9n0DNiokasAZKQ=",
+          "requires": {
+            "p-limit": "^2.0.0"
+          }
+        },
+        "p-try": {
+          "version": "2.0.0",
+          "resolved": "https://registry.npmjs.org/p-try/-/p-try-2.0.0.tgz",
+          "integrity": "sha1-hQgLuHxkaI+keZb+j3376CEXYLE="
         },
         "read-pkg": {
           "version": "3.0.0",
@@ -7488,14 +7526,14 @@
           "resolved": "https://registry.npmjs.org/read-pkg-up/-/read-pkg-up-4.0.0.tgz",
           "integrity": "sha512-6etQSH7nJGsK0RbG/2TeDzZFa8shjQ1um+SwQQ5cwKy0dhSXdOncEhb1CPpvQG4h7FyOV6EB6YlV0yJvZQNAkA==",
           "requires": {
-            "find-up": "^2.0.0",
+            "find-up": "^3.0.0",
             "read-pkg": "^3.0.0"
           }
         },
         "source-map": {
           "version": "0.6.1",
           "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
-          "integrity": "sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g=="
+          "integrity": "sha1-dHIq8y6WFOnCh6jQu95IteLxomM="
         },
         "uglify-js": {
           "version": "3.4.5",
@@ -8531,15 +8569,6 @@
           "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
           "integrity": "sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g=="
         }
-      }
-    },
-    "postcss-clean": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/postcss-clean/-/postcss-clean-1.1.0.tgz",
-      "integrity": "sha512-83g3GqMbCM5NL6MlbbPLJ/m2NrUepBF44MoDk4Gt04QGXeXKh9+ilQa0DzLnYnvqYHQCw83nckuEzBFr2muwbg==",
-      "requires": {
-        "clean-css": "^4.x",
-        "postcss": "^6.x"
       }
     },
     "postcss-import": {
--- a/packages/angular_devkit/build_ng_packagr/package.json
+++ b/packages/angular_devkit/build_ng_packagr/package.json
@@ -11,7 +11,8 @@
   "dependencies": {
     "@angular-devkit/architect": "0.0.0",
     "@angular-devkit/core": "0.0.0",
-    "rxjs": "^6.0.0"
+    "rxjs": "^6.0.0",
+    "semver": "^5.3.0"
   },
   "peerDependencies": {
     "ng-packagr": "^2.2.0 || ^3.0.0 || ^4.0.0"
--- a/packages/angular_devkit/build_ng_packagr/src/build/index.ts
+++ b/packages/angular_devkit/build_ng_packagr/src/build/index.ts
@@ -12,9 +12,14 @@ import {
   BuilderConfiguration,
   BuilderContext,
 } from '@angular-devkit/architect';
-import { getSystemPath, normalize, resolve } from '@angular-devkit/core';
+import { getSystemPath, normalize, resolve, tags } from '@angular-devkit/core';
+import * as fs from 'fs';
 import * as ngPackagr from 'ng-packagr';
-import { Observable } from 'rxjs';
+import { EMPTY, Observable } from 'rxjs';
+import { catchError, tap } from 'rxjs/operators';
+import * as semver from 'semver';
+
+const NEW_NG_PACKAGR_VERSION = '4.0.0-rc.3';
 
 // TODO move this function to architect or somewhere else where it can be imported from.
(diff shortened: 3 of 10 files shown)


packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
```diff
@@ -24,7 +24,8 @@
     "fullTemplateTypeCheck": true,
     "strictInjectionParameters": true,
     "flatModuleId": "AUTOGENERATED",
-    "flatModuleOutFile": "AUTOGENERATED"
+    "flatModuleOutFile": "AUTOGENERATED",
+    "enableResourceInlining": true
```
filipesilva: Is this required for the watch mode? If so we might need to warn users of older versions about it.

packages/schematics/angular/library/files/__projectRoot__/ng-package.json
```diff
@@ -1,7 +1,6 @@
 {
   "$schema": "<%= relativePathToWorkspaceRoot %>/node_modules/ng-packagr/ng-package.schema.json",
   "dest": "<%= relativePathToWorkspaceRoot %>/<%= distRoot %>",
-  "deleteDestPath": false,
```
filipesilva: I'd like to understand the motivation to remove this preset better. Does this option not play well with watch mode?

I originally added it to enable users to re-run build in dev mode without breaking watched files. But now with watch mode existing, this usage might be different.

I guess ideally we'd want a single `ng-package.json` file at all time, with the destination path being automatically deleted in single-run mode and not in watch mode.

But that does beg the question: if `deleteDestPath` does not interact well with watch mode, maybe it should be ignored in `ng-packagr` directly.

filipesilva: Regardless of the final semantics of watch mode, I do agree on doing away with the dev/prod configs if their only difference is watch mode support.

packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
alan-agius4: Yes this is required for V4. And we removed a lot of internal logic to inline resources.

packages/angular_devkit/build_ng_packagr/package.json
```diff
@@ -14,6 +14,6 @@
     "rxjs": "^6.0.0"
   },
   "peerDependencies": {
-    "ng-packagr": "^2.2.0 || ^3.0.0 || ^4.0.0-rc.0"
```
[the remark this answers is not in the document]
alan-agius4: I like the whole idea!

alan-agius4:
@filipesilva I updated the PR as suggested.

Do you have any sample on how should I test the `watch`?

alan-agius4:
AppVeyor is failing due to the recurring issue `Error: ENOTEMPTY: directory not empty, rmdir`

packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
filipesilva: Ok, then I think we'll need to check this as well when asserting compatibility.

packages/schematics/angular/library/files/__projectRoot__/ng-package.json
filipesilva: It's also important to remove the dist dir overall though. If you don't, artifacts that aren't there in subsequent builds will still be present because of previous builds.

With apps this is especially relevant because of assets, but less relevant for libraries. Still, the correct thing to do on a fresh, non-watch build is to clean (delete) the outpath before re-generating artifacts.

That was why we had it on on the production builds, and not the development ones. The production builds were the 'correct' ones, while the dev builds made that concession for a makeshift watch mode.

packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
alan-agius4: For v3, this flag might cause some unwanted behaviour I think.

packages/schematics/angular/library/files/__projectRoot__/ng-package.json
alan-agius4: By default `ng-packagr` deletes the `dist` directory when when triggering the build/watch

alan-agius4:
@filipesilva do you mind re-triggering the CI? seems like the errors are unrelated

filipesilva:
A good way to test the watch mode is to both check the build events are being emitted properly, and that the build artifacts contain what you expect. 

A good example can be found here https://github.com/angular/angular-cli/blob/6449a753641340d8fc19a752e1a1ced75f974efa/packages/angular_devkit/build_angular/test/browser/rebuild_spec_large.ts#L23-L112

The appveyor error (below) you are seeing is unrelated to your PR though. It's something I have been unable to catch properly on appveyor builds. Will look at in again today.

```
  19 Browser Builder rebuilds
    √ rebuilds on TS file changes (19 secs)
Error: ENOTEMPTY: directory not empty, rmdir 'C:\projects\angular-cli\tests\@angular_devkit\build_angular\test-project-host-hello-world-app-5t0wtykex0e\src'
    at Object.fs.rmdirSync (fs.js:821:3)
    at MergeMapSubscriber.isDirectory.pipe.operators_1.concatMap.isDir [as project] (C:\projects\angular-cli\packages\angular_devkit\core\node\file:\C:\projects\angular-cli\packages\angular_devkit\core\node\host.ts:273:16)
    at MergeMapSubscriber._tryNext (C:\pr
```

BTW regarding timing, this PR cannot go in for 6.1 because we are already in RC. But it can go in for 6.2.

packages/schematics/angular/library/files/__projectRoot__/ng-package.json
filipesilva: Ah I wasn't aware of that! Then removing the option makes perfect sense.

alan-agius4:
Thanks @filipesilva, that was really helpful.

I have added the unit tests.

Ps: I love how the Virtual FileSystem thingy works in the CLI. Big kudos guys :)

packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
filipesilva: Yeah makes sense. 

Maybe we can fully remove these values and let ng-packagr take care of them? I think we ended up adding them manually to get around a tsconfig merging bug in ng-packagr. It might not be present anymore.

If they are still needed then we need to do the health check both ways then: that they are not present when using ng-packagr v3, and that they are when using v4.

alan-agius4: Actually I gave this a bit more thought. And I think. IN V3, this might not have any `sideEffects` as there is a previous step to transforms the components.

I have updated the v3 branch to include this flag, and from the Unit Tests all seem fine
https://github.com/dherges/ng-packagr/pull/989

package-lock.json
```diff
@@ -180,7 +180,7 @@
     "@types/estree": {
```
alan-agius4: Not sure why so many changes, I am using `npm --version 6.1.0`

alan-agius4: Looks like it was due to a private npm repo

alan-agius4:
@filipesilva I am now having the rmdir error on my newly added tests aswell. :/

filipesilva:
Yeah we are trying to do a FS abstraction with the Hosts but still have a couple of parts missing for true cross platform support. Glad you like it!

I think the current rmdir error you have is legit. To make watch mode work with tooling, it is necessary that the watcher can exit gracefully.

For webpack, that happens in https://github.com/angular/angular-cli/blob/47f937e28f0a9d82997bd29630a8b3b92f44c0f5/packages/angular_devkit/build_webpack/src/webpack/index.ts#L71-L72

This is actually not as big a problem in linux/osx as it is on windows. 

In linux/osx, if you don't exit the watchers gracefully, you get hanging process. In our test runner that isn't a problem because we do process.exit at the end. We do this because we have found some situations when processes still hang in webpack and in karma.

In windows however, you get errors when removing files because the watcher is keeping a file lock.

So I think you need a API to tell ng-packagr to exit gracefully in order to use it with tooling (such as the ng-packagr builder).

alan-agius4:
Ah year, the teardown logic makes perfect sense as I am doing the subscription internally and I was never unsubscribing.

It was odd though because I didn't have the same errors when running the tests locally on a Windows machine.

Ps: teardown logic added

packages/schematics/angular/library/files/__projectRoot__/tsconfig.lib.json
filipesilva: Awesome!

packages/angular_devkit/build_ng_packagr/src/build/index.ts
```diff
@@ -12,20 +12,55 @@ import {
...
+  watch?: boolean;
+}
+
+function checkNgPackagrVersion(projectRoot: string): boolean {
+  let ngPackagrJsonPath;
+
+  try {
+    ngPackagrJsonPath = resolveProjectModule(projectRoot, 'ng-packagr/package.json');
+  } catch {
+    // @angular/service-worker is not installed
```
filipesilva: Nit: comment is lying. 

Not really important to get this merged but should be changed if a rebase is needed.

alan-agius4: I will rebase and do it :)

alan-agius4:
i'll also update the docs in this PR.

alan-agius4:
Docs have been updated.

alan-agius4:
@filipesilva docs have been updated, would you mind review it again please?

Thanks

filipesilva:
Doc changes look good too, great work!

We'll be looking at getting this merged after 6.1 final, which is when we will release 6.2.0-beta.0.

pterblgh:
Correct me if I'm wrong, but this feature will make available for monorepos with multiple libraries to live-reload the main application which uses the libraries?

Currently I have a working setup with multiple libraries and a consuming app with lerna, but the developer experience is not quite good because every time someone makes a change in a library it's needed to be rebuilt. Is this feature going to solve this problem?

alan-agius4:
Hi, while I don’t have full picture of your architecture with lerna. However this should work, by spawning 2 process one that watches for the changes in the library and the other one that is watching the app. So both library and app will do a partial re-build.

pterblgh:
Thanks for the quick answer @alan-agius4, I'm using Angular CLI for both building the libraries and the main application. I guess I will setup a lerna command to execute `ng build --watch` in the libraries and to start the main application.

If everything goes well I guess it should work correctly, because right now if I have a running main app in a process and I trigger a manual library build, the main app partially re-builds as expected.

smoke:
@pterblgh You don't need lerna, I have setup the following scripts https://github.com/ngx-api-utils/ngx-api-utils/blob/master/package.json#L13 and it works quite well. For library / package rebuild on change I am using https://github.com/ngx-api-utils/ngx-api-utils/blob/master/package.json#L24 so you can also have `nodemon` doing it. 
Of course when we have `watch` when building a library, speed will be increased because of the incremental builds. 
So I am quite eager seeing this in @angular/cli!

packages/angular_devkit/build_ng_packagr/src/build/index.ts
```diff
@@ -53,12 +88,30 @@ export class NgPackagrBuilder implements Builder<NgPackagrBuilderOptions> {
         ngPkgProject.withTsConfig(tsConfigPath);
       }
 
-      ngPkgProject.build()
-        .then(() => {
-          obs.next({ success: true });
-          obs.complete();
-        })
-        .catch((e) => obs.error(e));
+      if (options.watch) {
```
gatimus: Since ng-packagr `build()` uses `buildAsObservable()` internally why not remove the `if` and the `new Observable` statement and do `return ngPkgProject..withOptions({ watch: options.watch })buildAsObservable().pipe(map(() => {return { success: true };}))`

alan-agius4: The reason being is that angular cli still needs to support ng-package v2 and v3. And this method is only available in ng-packagr v4

gatimus: Never mind, I see, for backwards compatibility with ng-packagr 3.x when not using watch, right?

alan-agius4: Yes

alan-agius4:
Will eventually rebase when it's ready to get merged. Also, I created another PR to bump ng-packagr to v4 final

https://github.com/angular/angular-cli/pull/11622

filipesilva:
@alan-agius4 should be ok to rebase now, https://github.com/angular/angular-cli/pull/11622 is in as well.

alan-agius4:
thanks @filipesilva, I just rebased.

PR #11365, opened 2018-06-25
refactor: remove redundant `await`'s

alan-agius4:
An async function always wraps the return value in a Promise. Using return await just adds extra time before the overreaching promise is resolved without changing the semantics.


clydin:
The `await` within the try/catch blocks is necessary to allow the catch blocks to function as intended.

alan-agius4:
re-added those on try/catch

PR #11388, opened 2018-06-27
Add graphql-tag

tianhuil:
graphql-tag isn't added as a feature and this is an easy solution that would sidestep many of the issues brought up by https://github.com/angular/angular-cli/issues/1656

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "file-loader": "^1.1.11",
     "font-awesome": "^4.7.0",
     "glob": "^7.0.3",
+    "graphql-tag": "^2.9.2",
     "html-webpack-plugin": "^3.0.6",
     "husky": "^0.14.3",
     "istanbul": "^0.4.5",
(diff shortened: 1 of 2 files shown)


tianhuil:
I signed it.

thefliik:
@tianhuil I think you might need to reply `I signed it!` exactly as is (i.e. with the `!` ) for the google bot to pick up on it.

tianhuil:
I signed it!

tianhuil:
Thanks @thefliik.

hansl:
Closing this as this is not something we want to support. `ngx-build-plus` is the appropriate way of handling non-standard webpack configurations now. We are in the process of opening the Architect API so these can be part of community provided packages.

thefliik:
Makes sense. I found https://github.com/meltedspark/angular-builders which is a repo attempting to catalog 3rd party builders.

PR #11391, opened 2018-06-27
fix(@schematics/angular): remove ngoninit from default component

StephenFluin:
With the challenges and misconceptions around ngOnInit, I feel like it would be better to remove it from components by default.

1. initializations provided by ngOnInit are not seen by the type system and may cause the user to fail `strictPropertyInitialization` checks from typescript
1. The best practice is for data flows (such as observables) to be defined in the constructor
1. Users seem to get think they should access Inputs in the ngOnInit, but instead they should be using ngOnChanges

Thoughts? @hansl @robwormald


alan-agius4:
@StephenFluin I am really happy about this, as there is a big misconception around the 3rd point especially and I see this on a daily basis almost. 

Not sure if it’s mentioned anywhere on AIO.

hansl:
@StephenFluin Is this PR still desirable?

StephenFluin:
I think it's desirable, but it needs more conversation about best practices.

PR #11401, opened 2018-06-28
Schematic fixes

clydin:
Removes some unneeded casting; simplifies some of the logic; and fixes an issue with chaining rules that return rules.

### DIFF
--- a/packages/angular_devkit/schematics/src/rules/call.ts
+++ b/packages/angular_devkit/schematics/src/rules/call.ts
@@ -7,7 +7,7 @@
  */
 import { BaseException, isObservable } from '@angular-devkit/core';
 import { Observable, of as observableOf, throwError } from 'rxjs';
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/schematics/src/rules/call.ts
```diff
@@ -51,55 +51,53 @@ export class InvalidSourceResultException extends BaseException {
...
-  if (result === undefined) {
-    return throwError(new InvalidSourceResultException(result));
-  } else if (TreeSymbol in result) {
-    return observableOf(result as Tree);
-  } else if (isObservable(result)) {
+  if (isObservable(result)) {
     // Only return the last Tree, and make sure it's a Tree.
-    return (result as Observable<Tree>).pipe(
+    return result.pipe(
+      defaultIfEmpty(),
```
filipesilva: 👍

PR #11425, opened 2018-07-02
CI fixes for large specs

filipesilva:
Fixes in this PR should address the following `npm run test-large` CI flakes:
- Uncaught ENOENT errors when restoring `TestProjectHost`
- Tests failing when the previous test timed out
- Hanging servers on test time out

### DIFF
--- a/packages/angular_devkit/core/node/host.ts
+++ b/packages/angular_devkit/core/node/host.ts
@@ -264,25 +264,32 @@ export class NodeJsSyncHost implements virtualFs.Host<fs.Stats> {
   delete(path: Path): Observable<void> {
     return this.isDirectory(path).pipe(
       concatMap(isDir => {
+        // TODO: remove this try+catch when issue https://github.com/ReactiveX/rxjs/issues/3740 is
+        // fixed.
         if (isDir) {
-          // Since this is synchronous, we can recurse and safely ignore the result.
-          for (const name of fs.readdirSync(getSystemPath(path))) {
-            this.delete(join(path, name)).subscribe();
-          }
-          try {
-            fs.rmdirSync(getSystemPath(path));
-          } catch (error) {
-            return throwError(error);
-          }
+          const dirPaths = fs.readdirSync(getSystemPath(path));
+          const rmDirComplete = new Observable((obs) => {
+            try {
+              fs.rmdirSync(getSystemPath(path));
+              obs.complete();
+            } catch (e) {
+              obs.error(e);
+            }
+          });
+
+          return concat(
+            ...dirPaths.map(name => this.delete(join(path, name))),
+            rmDirComplete,
+          );
         } else {
           try {
             fs.unlinkSync(getSystemPath(path));
-          } catch (error) {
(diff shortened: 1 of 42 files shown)


scripts/test.ts
```diff
@@ -171,6 +171,11 @@ export default function (args: ParsedArgs, logger: logging.Logger) {
     runner.env.addReporter(new IstanbulReporter());
   }
 
+  if (args.large) {
+    // Default timeout for large specs is 5 minutes.
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;
```
clydin: 5 minutes seems fairly long.  Are there tests that will take that long?

packages/angular_devkit/core/node/host.ts
```diff
@@ -265,24 +265,21 @@ export class NodeJsSyncHost implements virtualFs.Host<fs.Stats> {
...
       concatMap(isDir => {
         if (isDir) {
-          // Since this is synchronous, we can recurse and safely ignore the result.
-          for (const name of fs.readdirSync(getSystemPath(path))) {
-            this.delete(join(path, name)).subscribe();
-          }
-          try {
+          const dirPaths = fs.readdirSync(getSystemPath(path));
+          const rmDirComplete = new Observable((obs) => {
             fs.rmdirSync(getSystemPath(path));
```
clydin: Unless the minimum version of rxjs is increased everywhere, a try/catch block is needed to work around a rxjs defect (https://github.com/ReactiveX/rxjs/issues/3740).

scripts/test.ts
filipesilva: The biggest timeout before was 120s, and I think some of the timeouts I left are actually 180s now. 

Neither goes up to 5mins though, my comment is just wrong.

packages/angular_devkit/core/node/host.ts
filipesilva: Will re-add with the comment that the rest in https://github.com/angular/angular-cli/commit/cfa03826facc37bb8911eb584eafad856c13ce8a have.

packages/angular_devkit/core/node/host.ts
```diff
@@ -264,25 +264,28 @@ export class NodeJsSyncHost implements virtualFs.Host<fs.Stats> {
...
-          }
-        } else {
-          try {
+        // TODO: remove this try+catch when issue https://github.com/ReactiveX/rxjs/issues/3740 is
+        // fixed.
+        try {
+          if (isDir) {
+            const dirPaths = fs.readdirSync(getSystemPath(path));
+            const rmDirComplete = new Observable((obs) => {
+              fs.rmdirSync(getSystemPath(path));
```
clydin: The try/catch needs to be around this call inside the observable constructor.
```
try {
  fs.rmdirSync(getSystemPath(path));
  obs.complete();
} catch (e) {
  obs.error(e);
}
```

filipesilva: Done

PR #11432, opened 2018-07-02
Refactors for tslint rules

hansl:
Read the commit messages for more information about some of those changes.

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -902,12 +902,10 @@ export class AngularCompilerPlugin {
   }
 
   writeI18nOutFile() {
-    function _recursiveMkDir(p: string): Promise<void> {
-      if (fs.existsSync(p)) {
-        return Promise.resolve();
-      } else {
-        return _recursiveMkDir(path.dirname(p))
-          .then(() => fs.mkdirSync(p));
+    function _recursiveMkDir(p: string) {
+      if (!fs.existsSync(p)) {
+        _recursiveMkDir(path.dirname(p));
+        fs.mkdirSync(p);
       }
     }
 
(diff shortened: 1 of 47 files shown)


packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -204,7 +204,7 @@ export class AngularCompilerPlugin {
 
     this._rootNames = config.rootNames.concat(...this._singleFileIncludes);
     this._compilerOptions = { ...config.options, ...options.compilerOptions };
-    this._basePath = config.options.basePath || '';
+    this._basePath = config.options.basePath || basePath || '';
```
filipesilva: This might change how the plugin works. What was the motivation? I get the feeling it should be either `config.options.basePath || ''` or `basePath || ''` but not the current one.

hansl: So looking at the code `basePath` itself wasn't used, and there was a lot of logic around it. I also thought it might change behaviour, but it seems that all tests are running fine. I can revert and remove the `basePath` logic entirely.

filipesilva: I'm ok with leaving it as is now.

tslint.json
```diff
@@ -1,16 +1,53 @@
 {
-  "extends": "tslint-no-circular-imports",
+  "//": "Consider using no-identical-functions and disabling in faulty tests",
```
chigix: Why not directly use json comment in this json file instead of the bunch of tricky json keys? The current tslint is actually breaks some [RFCs](https://tools.ietf.org/html/rfc7159#section-4) and trigger unnecessary warning about duplicate keys problem in some IDEs.

hansl: JSON does not support comments. TSLint would not be able to read this file if there was comments in it.

alan-agius4: @hansl it is true that json doesn’t support comments, but tslint does.

https://github.com/palantir/tslint/blob/master/test/configurationTests.ts

alan-agius4: Updated link above, as I copied the wrong one.

Issue #7271, opened 2017-08-04
Updating autoprefixer to v7.1.x and postcss v6.0.x


JulienBourgain:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Versions.
@angular/cli@1.3.0-rc.5

### Desired functionality.

In my project, i use css-grids and the default option in autoprefixer 6 is grid: true 
In autoprefixer 7, it's grid: false

i think there is some other improvements

clydin:
This is waiting on cssnano 4.0 to be released to have a full postcss 6 pipeline.

mgol:
Note that Autoprefixer 7 disables CSS Grid prefixes for IE by default so unless CLI passes the option explicitly (and unless there are more breaking changes there) the update to Autoprefixer 7 will be breaking and will require a major version bump of Angular CLI.

kylecordes:
@mgol  I wonder if would be sufficient to simply set grid: true, if that will be enough to make it possible to move up to v7, without having to simultaneously solve the problem of making autoprefixer configurable.

kylecordes:
@mgol  I have a workaround in progress that does this and appears to work fine. Writeup coming soon.

alan-agius4:
These dependencies have been update

Issue #5964, opened 2017-04-14
CSS Grid Layout generates compile warnings from autoprefixer


Stexxen:
### Bug Report or Feature Request (mark with an `x`)
```
- [ X] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.
@angular/cli: 1.0.0
node: 7.5.0
os: darwin x64
@angular/animations: 4.0.2
@angular/common: 4.0.2
@angular/compiler: 4.0.2
@angular/compiler-cli: 4.0.2
@angular/core: 4.0.2
@angular/flex-layout: 2.0.0-beta.7
@angular/forms: 4.0.2
@angular/http: 4.0.2
@angular/material: 2.0.0-beta.3
@angular/platform-browser: 4.0.2
@angular/platform-browser-dynamic: 4.0.2
@angular/platform-server: 4.0.2
@angular/router: 4.0.2
@angular/cli: 1.0.0
@angular/language-service: 2.4.10

### Repro steps.

Use any current version [CSS Grid Layout](http://caniuse.com/#feat=css-grid) statement that does not have an equivalent in IE or Edge, and which autoprefixer cannot convert
They use an outdated version of the spec, see point 2 - http://caniuse.com/#feat=css-grid

ie.
```css
.header {
  grid-column-start: 1;
  grid-column-end: 2;
  grid-row-start: 1;
  grid-row-end: 2;
}
```

### The log given by the failure.

```
WARNING in ./src/app/dashboard/dashboard.component.css
autoprefixer: /opt/dev/angular2/project6/src/app/dashboard/dashboard.component.css:26:3: IE 
supports only grid-row-end with span. You should add grid: false option to Autoprefixer and use
some JS grid polyfill for full spec support
 @ ./src/app/dashboard/dashboard.component.ts 18:17-53
 @ ./src/app/app.module.ts
 @ ./src/main.ts
 @ multi webpack-dev-server/client?http://localhost:5200 ./src/main.ts

```

### Desired functionality.
The error from Autoprefixer mentions setting `grid: false` in config, but I'm aware through other issues against this repository, that changing any webpack related config is frowned upon and made intentionally difficult. The reasons for this I can fully understand.

But in this case the warnings are seemingly unavoidable as am using CR Level 1 CSS Grid Layout and I have no intention of supporting IE or Edge in their current form. Edge is [working towards CR Level 1 
](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/gridupdate/)
[...]

PR #11453, opened 2018-07-04
Disable Autoprefixer Grid

ai:
@hansl thanks for being so brave to enable Grid feature in Autoprefixer.

Unfortunately, it may not be the best idea to enable it for everyone. This feature has a lot of [limits](https://css-tricks.com/css-grid-in-ie-css-grid-and-the-new-autoprefixer/). And must be explicitly controlled by a developer.

For instance, users complain about Grid warnings: https://github.com/postcss/autoprefixer/issues/1062#issuecomment-402202643

Maybe we can add an extra option to Angular CLI to enable Grid feature?

/cc @clydin

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
@@ -152,7 +152,7 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
         loader,
         filename: `[name]${hashFormat.file}.[ext]`,
       }),
-      autoprefixer({ grid: true }),
+      autoprefixer(),
     ];
   };
 


ai:
I signed it!

tfalvo:
Sorry but I don't understand, why we should disable completely grid feature in autoprefixer.
For example, in my case, I use grid layout and it was working perfectly with Angular.
But what about update you made in project mentioned at the beginning of [this discussion](https://github.com/postcss/autoprefixer/issues/1062#issuecomment-402202643).
Please @yepninja could you explain what is this change "IE doesn't support auto-fill and auto-fit. Let's don't add a prefixed prop and create warnings for these values.". And so why theses warnings appear even if you add browser rules omission on autoprefixer.

Is it possible to "unscope" IE but not all the grid feature ?

Sorry in advance if my comprehension is not good. but i'm a little afraid we'll shut down the tree while we can probably just cut a branch... ;-)

ai:
We don't need to disable it completely. Current idea is to disable it by default and enable only if you understand it's limits.

But I don't know Angular CLI enough to suggest a option to enable/disable Grid on the flight.

clydin:
Unfortunately, disabling (permanently or by default) would be a breaking change.  Also, the default `browserslist` file for a new project ([found here](https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/application/files/root/browserslist)) does not include IE9-11.  Are there issues beyond IE supportthat need to be considered?

ai:
@clydin maybe we should give API to at least disable it?

What is user support IE 11 but use different polyfill for Grid?

clydin:
Making it configurable is a potential option.  I'll bring it up for discussion.

Also,  I only found two alternate polyfills for IE.  Both appear to be based on outdated specs and are runtime based.  The second of which would make them impractical for use with Angular AOT as the component stylesheets are injected/removed at runtime.  It appears that the autoprefixer approach is the most practical for Angular.

tomschreck:
I found this workaround:  https://github.com/angular/angular-cli/issues/5964#issuecomment-294161554.  The workaround has the comments above each item in CSS block.  I found I only had to insert the comment once in each block.

hansl:
I'm tempted to merge this PR, unless @clydin has some reservations. I'll merge this tomorrow morning if there's no objection.

mjgoethe:
@hansl Could this be made configurable? It's definitely an important feature for some. Alternatively, one possible workaround would be the grid control comment support that was added in autoprefixer 9.2, but Angular CLI is currently on 9.1.5.

garethrpratt:
@hansl as I understand it, the "grid: true" was originally introduced so that upgrading autoprefixer within angular-build wasn't a breaking change (as it used to default to true):
https://github.com/angular/angular-cli/issues/7271#issuecomment-353068912

We were also making use of this functionality, it would be useful to have a mechanism to turn it back on if possible.

cedrickring:
Would love to get this option back, as I don't want to prefix all my grid styles explicitly. (Took me ~3 hours to find out that it got disabled several weeks ago... and it's not even documented anywhere).

garethrpratt:
Looks like the fix has now been merged to master:
https://github.com/angular/angular-cli/pull/12785

PR #11463, opened 2018-07-05
Remove Build Optimizer integration tests

filipesilva:


.circleci/config.yml
```diff
@@ -58,35 +58,12 @@ jobs:
     environment:
       BASH_ENV: ~/.profile
     resource_class: xlarge
-    parallelism: 4
```
clydin: typo?

filipesilva: Yes, good catch.

PR #11464, opened 2018-07-05
test: shard test-large

filipesilva:


scripts/test.ts
```diff
@@ -204,6 +204,13 @@ export default function (args: ParsedArgs, logger: logging.Logger) {
     logger.info(`Found ${tests.length} spec files, out of ${allTests.length}.`);
   }
 
+  if (args.shard) {
```
clydin: the first shard is zero which will fail the check here.

filipesilva: Fixed

PR #11489, opened 2018-07-07
Update jasmine

alan-agius4:
feat(@schematics/schematics): update jasmine to 3.3.1+

feat(@schematics/angular): update workspace jasmine to 3.3.1+
 
build: update jasmine to 3.3.1+

test: fix chunk in entry test

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "git-raw-commits": "^2.0.0",
     "husky": "^1.3.1",
     "istanbul": "^0.4.5",
-    "jasmine": "^2.6.0",
+    "jasmine": "^3.3.1",
     "jasmine-spec-reporter": "^4.2.1",
     "karma": "~4.0.1",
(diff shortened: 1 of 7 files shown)


hansl:
We don't use Jasmine 3 internally yet, so this PR should be blocked until it's supported inside Google.

alan-agius4:
Odd as Angular/Angular are using Jasmine 3
https://github.com/angular/angular/blob/master/package.json

hansl:
It seems that's the case, but internally we still have Jasmine 2.9.

hansl:
Closing this. Once we support Jasmine 3 internally we can reopen. Until then, there's no advantage from moving the deps.

alexeagle:
Igor told me we need to update externally first so that John Barton can finish the internal update. I think this can land first?

alan-agius4:
@alexeagle, I just rebased, if this needs to go in. It should be all good once it's green.

PR #11491, opened 2018-07-07
fix(@angular-devkit/build-angular): don't add hash to lazy styles

### LINKED ISSUE #11235: Once I have created a lazy loaded css file, how do I lazy load it?
Among the many awesome things in the Angular CLI is the ability to lazy load a CSS file. My problem is that the documentation doesn't specify how to load it once you create it. When you run the build, the file that gets created for your css has a hash in the name, like `foo.asdf908a7sd0as7g.bundle.css`. Because the name of the production file isn't a static name, I can't figure out how to lazy load it. If there is a known way to do this, please show me, and I will help out and PR the docs. 

```
- [X] feature request
```

### Area
```
- [ maybe ] devkit
- [ ] schematics
- [ X ] documentation
```

### Versions

Angular CLI: 1.7.4
Node: 9.5.0
OS: darwin x64
Angular: 5.2.11

@angular/cdk: 5.2.5
@angular/cli: 1.7.4
@angular/material: 2.0.0-beta.12
@angular-devkit/build-optimizer: 0.3.2
@angular-devkit/core: 0.4.9
@angular-devkit/schematics: 0.3.2
@ngtools/json-schema: 1.2.0
@ngtools/webpack: 1.10.2
@schematics/angular: 0.3.2
@schematics/package-update: 0.3.2
typescript: 2.6.2
webpack: 3.11.0

### Repro steps

1. Create a css file called `foo.css`.
2. Edit the '.angular-cli.json` (I know in v6 it is called something else) to include the following in styles:
```
"styles":[
   ...
   {"input": "foo.css", "lazy": true}
]
```

There is no log, cause there is no failure. 

### Desired functionality

I would love to see a way to lazily load these css files. 

#5880 was a similar question that never got answered and was closed. A few people suggested some work arounds that don't work for production mode when the css file gets the hash included in it's name. The suggestions only work in dev mode. 

In a [separate issue ](https://github.com/angular/angular-cli/issues/4685#issuecomment-280008219) @filipesilva acknowledges that while a way to create lazily loaded css files exists, there is no provided way to load these created files. We need to get a way to allow people to load these created css files. In that issue, the participants are using native webpack functionality to load these css files. I don't want to use Webpack directly, because that will make it harder for me to upgrade my CLI one day when it switches to Bazel.

filipesilva:
Fix #11235

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
@@ -17,6 +17,7 @@ import { findUp } from '../../utilities/find-up';
 import { RawCssLoader } from '../../plugins/webpack';
 import { ExtraEntryPoint } from '../../../browser/schema';
 import { normalizeExtraEntryPoints } from './utils';
+import { RemoveHashPlugin } from '../../plugins/remove-hash-plugin';
 
 const postcssUrl = require('postcss-url');
 const autoprefixer = require('autoprefixer');
@@ -158,7 +159,7 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
 
   // use includePaths from appConfig
   const includePaths: string[] = [];
-  let lessPathOptions: { paths?: string[] } = { };
+  let lessPathOptions: { paths?: string[] } = {};
 
   if (buildOptions.stylePreprocessorOptions
     && buildOptions.stylePreprocessorOptions.includePaths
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/plugins/rename-output-plugin.ts
```diff
@@ -0,0 +1,43 @@
...
+import { Compiler, compilation } from 'webpack';
+import { HashFormat } from '../models/webpack-configs/utils';
+
+
+export interface RemoveHashPluginOptions {
+  chunkIds: string[];
+  hashFormat: HashFormat;
+}
+
+export class RemoveHashPlugin {
```
clydin: File name doesn't match class.

filipesilva: Fixed

PR #11493, opened 2018-07-07
docs(@angular/cli): resolve HTTP 301 redirects

jsshandle:


jsshandle:
Hello @filipesilva, right now the existing links are working. However, they are responding with a HTTP 301 (moved permanently) redirect. Which means the current URL might stop working in the future and may be considered depricated. For example the URL https://angular.github.io/protractor/ is redirected to http://www.protractortest.org/, which seems to be the new project page of Protractor.

PR #11497, opened 2018-07-08
 feat(@angular-devkit/build-angular): add profile option to browser builder

filipesilva:
This should help users send us profile logs for builds that take too long.

Blocked on https://github.com/angular/angular-cli/pull/11488 due to https://github.com/webpack/webpack/issues/7450.


package.json
```diff
@@ -125,6 +125,7 @@
     "source-map": "^0.5.6",
     "source-map-loader": "^0.2.3",
     "source-map-support": "^0.5.0",
+    "speed-measure-webpack-plugin": "^1.2.2",
```
hansl: This should use the new `webpack/lib/debug/ProfilingPlugin` instead of installing a new package.

filipesilva: It does both actually.

hansl: What does speed-measure-webpack-plugin does that ProfilingPlugin doesn't? It's an additional dependency.

filipesilva: ProfilingPlugin gives you a chrome perf profile, while SpeedMeasurePlugin gives you total aggregated time per plugin and loader. 

It's hard to get that overview with just ProfilingPlugin because the events not necessarily together. 

Loaders, for instance, can have many tiny events all over the place, that can add up quickly.

The API to consume SpeedMeasurePlugin is also easier to parse so we can use it to automatocally track individual plugin and loader performance over time instead of just build time.

I think those are the main differences.

PR #11503, opened 2018-07-09
refactor: types for content should be a pure template

hansl:
Limitting it to a subset of JSON values is limiting. Templates can call functions,
list arrays, access classes, etc.


hansl:
Added a new feature that can wait until 6.2 but will be necessary to be synced into Google3 when 6.1 is released.

packages/angular_devkit/schematics/src/rules/template_spec.ts
```diff
@@ -31,7 +31,7 @@ function _entry(path?: string, content?: string): FileEntry {
 }
 
 
-describe('applyPathTemplate', () => {
+fdescribe('applyPathTemplate', () => {
```
alfaproject: Oops

hansl: heh. fixed.

Issue #10624, opened 2018-05-04
"ng update" doesn't support private repos such as FontAwesome 5 Pro


bjornharvold:
### Versions
```
Angular CLI: 6.0.0
Node: 9.11.1
OS: darwin x64
Angular: 6.0.0
... animations, cli, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, platform-server, router
... service-worker

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.6.0
@angular-devkit/build-angular     0.6.0
@angular-devkit/build-optimizer   0.6.0
@angular-devkit/core              0.6.0
@angular-devkit/schematics        0.6.0
@ngtools/webpack                  6.0.0
@schematics/angular               0.6.0
@schematics/update                0.6.0
rxjs                              6.1.0
typescript                        2.7.2
webpack                           4.6.0
```

### Repro steps
* Add dependency on a private repo such as FontAwesome 5 Pro
* Run ng update @angular/core

### Observed behavior
```
Not found : @fortawesome/pro-regular-svg-icons
```

### Mention any other details that might be useful (optional)
This works fine using npm directly because the private repo has already been registered with npm

Ben305:
This should be fixed sooner than later

Timebutt:
There is definitely a problem when using a private npm repository. I could only run `ng update` correctly once I removed my `.npmrc` file, and removed every private package from `package.json`. After running `ng update @angular/cli'` it finally correctly generated the `angular.json` file and I readded all the references I removed before.

clydin:
Can you provide the list of option names used within `.npmrc`?

Timebutt:
Sure, this is what my `.npmrc` looks like:

```
registry=https://npm.showpad.io:8080
save-exact=true
//npm.showpad.io:8080/:_authToken=${NPM_TOKEN}
```

DaSchTour:
While looking at the growing list of bugs relating `.npmrc` #10704 #10571 #10660 I think this should be fixed asap

hansl:
Hi @bjornharvold, @DaSchTour,

Could you verify this is still an issue with 6.0.1?  There were a few fixes regarding the registry flag (we have 1 fix in queue for the strict-ssl flag). Thanks!

dimkanewtown:
```
    _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
[...]

PR #11505, opened 2018-07-10
fix(@schematics/update): fix strict ssl always as true bug

tengis:
The last release fixes an auth token issue on #10624 However, it doesn't get the correct **strict-ssl** option, since it uses package **rc** - boolean options comes as either boolean or undefined.

### DIFF
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -63,18 +63,18 @@ function getNpmConfigOption(
 
 function getNpmClientSslOptions(
   strictSsl?: string | boolean | undefined,
-  cafile?: string | boolean | undefined
+  cafile?: string | boolean | undefined,
 ) {
   const sslOptions: { strict?: boolean; ca?: Buffer } = {};
 
   if (strictSsl === undefined) {
     sslOptions.strict = true;
   } else {
-    sslOptions.strict = <boolean>strictSsl;
+    sslOptions.strict = <boolean> strictSsl;
   }
 
   if (cafile) {
-    sslOptions.ca = readFileSync(<string>cafile);
+    sslOptions.ca = readFileSync(<string> cafile);
   }
 
   return sslOptions;
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -62,15 +62,15 @@ function getNpmConfigOption(
 }
 
 function getNpmClientSslOptions(
-  strictSsl?: string | boolean | undefined,
-  cafile?: string | boolean | undefined,
+  strictSsl?: boolean,
+  cafile?: string,
 ) {
   const sslOptions: { strict?: boolean; ca?: Buffer } = {};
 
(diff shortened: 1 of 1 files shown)


packages/schematics/update/update/npm.ts
```diff
@@ -61,17 +61,20 @@ function getNpmConfigOption(
...
+) {
+  const sslOptions: { strict?: boolean; ca?: Buffer } = {};
 
-  if (strictSsl === 'false') {
-    sslOptions.strict = false;
-  } else if (strictSsl === 'true') {
+  if (strictSsl === undefined) {
     sslOptions.strict = true;
+  } else {
+    sslOptions.strict = <boolean> strictSsl;
```
clydin: if the `strictSsl` value is either a boolean or undefined then a direct assignment to `sslOptions.strict` should be the only change need (excepting the parameter type change for the function).  The entire conditional structure can be removed.  The underlying library defaults to `true` if the value is not set or undefined.

tengis: Done :+1:

packages/schematics/update/update/npm.ts
```diff
@@ -61,17 +61,16 @@ function getNpmConfigOption(
   return value;
 }
 
-function getNpmClientSslOptions(strictSsl?: string, cafile?: string) {
-  const sslOptions: { strict?: boolean, ca?: Buffer } = {};
+function getNpmClientSslOptions(
+  strictSsl?: string | boolean | undefined,
```
clydin: `strictSsl?: boolean`

packages/schematics/update/update/npm.ts
```diff
@@ -61,17 +61,16 @@ function getNpmConfigOption(
...
+  cafile?: string | boolean | undefined,
```
clydin: `cafile?: string`

packages/schematics/update/update/npm.ts
```diff
@@ -61,17 +61,16 @@ function getNpmConfigOption(
...
+) {
+  const sslOptions: { strict?: boolean; ca?: Buffer } = {};
 
-  if (strictSsl === 'false') {
-    sslOptions.strict = false;
-  } else if (strictSsl === 'true') {
-    sslOptions.strict = true;
-  }
+  sslOptions.strict = <boolean> strictSsl;
```
clydin: remove cast please

packages/schematics/update/update/npm.ts
```diff
@@ -61,17 +61,16 @@ function getNpmConfigOption(
...
 
   if (cafile) {
-    sslOptions.ca = readFileSync(cafile);
+    sslOptions.ca = readFileSync(<string> cafile);
```
clydin: remove cast please

tengis: Cools! I couldn't believe missed them. Thanks

clydin:
Superseded by #11680

Issue #7305, opened 2017-08-08
ng test --environment works only once, but not on watch auto re-run


lukemadera:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.

node --version
v6.10.1
npm --version
3.10.10
macOS (Sierra)

### Repro steps.

1. `npm run test` (which runs `ng test --environment=local`)
This works fine (tests run successfully, using the specified environment).
2. change a .spec test file
This fails - it tries to re-run tests due to the change, but this time the environment is NOT used, instead the defaults `environment.ts` is used (which in our case leads to an error because it does not exist. Even if it did exist, it still is a bug as the proper environment is not being used).

### The log given by the failure.

```
npm test

> pl-components-ng2@2.0.14 test /Users/lukemadera/Documents/code/presencelearning/pl-components-ng2
> ng test --environment=local

 10% building modules 1/1 modules 0 active08 08 2017 11:06:31.697:WARN [karma]: No captured browser, open http://localhost:9876/
08 08 2017 11:06:35.008:INFO [karma]: Karma v1.7.0 server started at http://0.0.0.0:9876/
08 08 2017 11:06:35.010:INFO [launcher]: Launching browser Chrome with unlimited concurrency
08 08 2017 11:06:35.015:INFO [launcher]: Starting browser Chrome
[../../../../../src recursive \.spec\.ts$] ./src \.spec\.ts$ 1.32 kB {main} [built]      
[../../../../../src/lib/pipes/commify.pipe.spec.ts] ./src/lib/pipes/commify.pipe.spec.ts 1.2 kB {main} [optional] [built]
[../../../../../src/lib/pl-button/index.spec.ts] ./src/lib/pl-button/index.spec.ts 135 bytes {main} [optional] [built]
[../../../../../src/lib/utilities/object.spec.ts] ./src/lib/utilities/object.spec.ts 1.74 kB {main} [optional] [built]
[../../../../../src/polyfills.ts] ./src/polyfills.ts 2.53 kB {polyfills} [built]
[../../../../../src/test.ts] ./src/test.ts 1.02 kB {main} [built]
[../../../../zone.js/dist/async-test.js] ./node_modules/zone.js/dist/async-test.js 3.23 kB {vendor} [built]
[../../../../zone.js/dist/fake-async-test.js] ./node_modules/zone.js/dist/fake-async-test.js 16.3 kB {vendor} [built]
[...]

PR #11509, opened 2018-07-10
Fix file watching with file replacements

### LINKED ISSUE #11339: Watching doesn't work for file paths remapped by the provided FS host
```
- [ ] feature request
```

### Area
```
- [X] devkit - @ngtools/webpack
- [ ] schematics
```

### Versions
node v9.5.0
npm 5.7.1
Linux Fedora 25

### Repro steps
Create a new ng cli project. Build it in production mode. As specified by the `angular.json` file, the requests to the `environment.ts` file are remapped to `environment.prod.ts`. However, if you change the `environment.prod.ts` file, that webpack won't rebuild the project. Also, if you change the originally requested file - `environment.ts`, webpack will detect the change and trigger a build.

### Desired functionality
The app should be rebuild when a remapped file is changed.

It looks like the original paths (and not the remapped ones) are added as dependencies for the webpack compilation in the `@ngtools/webpack` loader - [here](https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/loader.ts#L77-L87). The same thing happens for the original component templates and styles (html, css, scss files).

clydin:
Fixes #11339


packages/angular_devkit/build_angular/src/utils/add-file-replacements.ts
```diff
@@ -22,53 +29,64 @@ export class MissingFileReplacementException extends BaseException {
...
-  root: Path,
-  host: virtualFs.AliasHost,
-  fileReplacements: FileReplacement[],
-): Observable<null> {
+export interface NormalizedFileReplacement {
+  replace: Path;
+  with: Path;
+}
 
+export function normalizeFileReplacements(
```
filipesilva: Nit: file name does not match export name.

tests/legacy-cli/e2e/tests/build/rebuild-replacements.ts
```diff
@@ -0,0 +1,31 @@
+import { getGlobalVariable } from '../../utils/env';
```
filipesilva: WDYT of replacing this with a new spec large in `replacements_spec_large`?

This is the one I had on my branch:

```
it('file replacements work with watch mode', (done) => {
    const overrides = {
      fileReplacements: [
        {
          replace: normalize('/src/meaning.ts'),
          with: normalize('/src/meaning-too.ts'),
        },
      ],
      watch: true,
    };

    let buildNumber = 0;

    runTargetSpec(host, browserTargetSpec, overrides, 45000, createConsoleLogger()).pipe(
      debounceTime(1000),
      tap((buildEvent) => expect(buildEvent.success).toBe(true)),
      tap(() => {
        const fileName = join(outputPath, 'main.js');
        buildNumber += 1;

        switch (buildNumber) {
          case 1:
            expect(virtualFs.fileBufferToString(host.scopedSync().read(fileName)))
              .toMatch(/meaning\s*=\s*42/);
            expect(virtualFs.fileBufferToString(host.scopedSync().read(fileName)))
              .not.toMatch(/meaning\s*=\s*10/);
            host.writeMultipleFiles({
              'src/meaning-too.ts': 'export var meaning = 84;',
            });
            break;

          case 2:
            expect(virtualFs.fileBufferToString(host.scopedSync().read(fileName)))
              .toMatch(/meaning\s*=\s*84/);
            expect(virtualFs.fileBufferToString(host.scopedSync().read(fileName)))
              .not.toMatch(/meaning\s*=\s*42/);
            break;
        }
      }),
      take(2),
  });
```

It should be faster than the e2e and also test the content is correct.

filipesilva:
Also I think this fixes https://github.com/angular/angular-cli/issues/7305

packages/angular_devkit/build_angular/src/utils/add-file-replacements.ts
clydin: Done.

tests/legacy-cli/e2e/tests/build/rebuild-replacements.ts
clydin: Added.

filipesilva:
LGTM, great work!

hansl:
I fear this doesn't cover the NativeScript case with pattern matching (matching all `.html.tns` files to `.tns`). They'd have to list all those files, and even then it doesn't work when creating new files...

clydin:
Based on the custom host found [here](https://github.com/NativeScript/nativescript-dev-webpack/blob/master/host/platform.ts), code similar to the following should work:
```
            new AngularCompilerPlugin({
                entryModule: resolve(appPath, "app.module#AppModule"),
                tsConfigPath: join(__dirname, "tsconfig.esm.json"),
                skipCodeGeneration: !aot,
                hostReplacementPaths: path => {
                    const { dir, name, ext } = parse(path);

                    for (const platform of extensions) {
                        const platformFileName = `${name}.${platform}${ext}`;
                        const platformPath = toSystemPath(join(dir, platformFileName));
            
                        try {
                            const stat = statSync(platformPath);
                            if (stat && stat.isFile()) {
                                return platformPath;
                            }
                        } catch(_e) {
                            // continue checking the other platforms
                        }
                    }
            
                    return path;
                },
            }),
```

Another option with this PR is to try to use the [PlatformFSPlugin](https://github.com/NativeScript/nativescript-dev-webpack/blob/master/plugins/PlatformFSPlugin.ts) which appears to be used within the non-Angular demos found in the repository.

PR #11510, opened 2018-07-10
fix(@angular/cli): change serve fallback config to support Accept header wildcard

davidsmith-accesso:
If a resource request is not present on the server and the Accept header only contains the wildcard / to accept all content, the server responds with a 404 instead of the index.html content.

Testing:
Execute a GET request against any resource not available on the local dev server with the following Accept header:
Accept: */*
The server will return a 404 instead of a 200 with the index.html content

### DIFF
--- a/packages/@angular/cli/tasks/serve.ts
+++ b/packages/@angular/cli/tasks/serve.ts
@@ -220,7 +220,7 @@ export default Task.extend({
       historyApiFallback: {
         index: `${servePath}/${appConfig.index}`,
         disableDotRule: true,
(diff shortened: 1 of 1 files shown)


hansl:
Hi @davidsmith-accesso ,

I like this PR, but we don't accept new features or non-urgent fixes on 1.7. I'm going to close this and see if we can make it for 7.0.

Thanks!

hansl:
Actually re-reading the code this is probably not something we want. Images and CSS should not return a 200 code with the index.html content. Only if asking for HTML. Hope this clears things up. Thanks!

PR #11511, opened 2018-07-10
refactor: fix types to remove any

hansl:
require() returns any so we were missing some typing errors.

### DIFF
--- a/packages/angular/cli/lib/cli/index.ts
+++ b/packages/angular/cli/lib/cli/index.ts
@@ -8,47 +8,65 @@
 
 import { logging, terminal } from '@angular-devkit/core';
 import { filter } from 'rxjs/operators';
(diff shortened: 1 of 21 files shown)


packages/angular/cli/lib/cli/index.ts
```diff
@@ -8,47 +8,58 @@
...
 
     // deprecated
-    'get': require('../../commands/getset').default,
-    'set': require('../../commands/getset').default,
-  };
+    import('../../commands/getset'),
+    import('../../commands/getset'),
+  ];
+
+  return Promise.all(allCommands).then((allCommands: { default: CommandConstructor }[]) => {
```
alexeagle: Nit: could replace block and return statement with fat arrow

Issue #11533, opened 2018-07-12
ng e2e documentation, clarify [project] parameter requires -e2e suffix


tongfa:
### Bug Report or Feature Request (mark with an `x`)
```
- [ x] bug report -> please search issues before submitting
- [ ] feature request
```

### Area
```
- [ x] devkit
- [ ] schematics
```

### Versions
6.0.8

Will attach a PR shortly.

tongfa:
https://github.com/angular/angular-cli/pull/11534

hansl:
The ng e2e command accept any project, but by default we use the same project name and add -e2e to it. There's nothing preventing you from renaming that project (or having a new one) with a different name.

Going to close this but I'll keep a note with the documentation team to see if it's something we want to explain.

PR #11534, opened 2018-07-12
docs: clarify ng e2e project parameter format

tongfa:
The `ng e2e [project]` command requires an `-e2e` extension after the project name.
Let us document this.

https://github.com/angular/angular-cli/issues/11533

### DIFF
--- a/docs/documentation/e2e.md
+++ b/docs/documentation/e2e.md
@@ -15,6 +15,12 @@ ng e2e [project]
 ng e2e
 ```
 
+Add '-e2e' extension to project names.
+
+```bash
+ng e2e myproject-e2e
+```
+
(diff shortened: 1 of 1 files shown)


hansl:
Hi @tongfa,

This is wrong; the `ng e2e` command accept any project, but by default we use the same project name and add `-e2e` to it. There's nothing preventing you from renaming that project (or having a new one) with a different name.

Going to close this as it would be misleading.

hansl:
Hi @tongfa,

The ng e2e command accept any project, but by default we use the same project name and add -e2e to it. There's nothing preventing you from renaming that project (or having a new one) with a different name.

Going to close this as it would be misleading.

Issue #10940, opened 2018-05-21
Angular Cli: 6.0.1 code coverage thresholds not failing ng test task


banjankri:
After upgrade to Angular Cli 6.x the code coverage thresholds enforced by the _karma-coverage-istanbul-reporter_ do not make the testing command fail with _emitWarning_ set to false as they used to when using Angular Cli 6.0.1.

### Versions
```
Angular CLI: 6.0.1
Node: 8.9.1
OS: win32 x64
Angular: 6.0.1
... animations, cli, common, compiler, compiler-cli, core, forms
... http, platform-browser, platform-browser-dynamic
... platform-server, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.6.1
@angular-devkit/build-angular     0.6.1
@angular-devkit/build-optimizer   0.6.1
@angular-devkit/core              0.6.1
@angular-devkit/schematics        0.6.1
@ngtools/webpack                  6.0.1
@schematics/angular               0.6.1
@schematics/update                0.6.1
rxjs                              6.1.0
typescript                        2.7.2
webpack                           4.6.0
```

### Repro steps
1) Create a repo using Angular Cli 1.7.x
2) Enforce thresholds
3) See the test command failing
4) Upgrade Angular Cli and migrate

### Observed behavior
```
The error is emitted by the reporter, however seems ignored by the angular client.
```

### Desired behavior
The test command should be failing when thresholds not met as it used to.

### Mention any other details that might be useful (optional)
Tried with both karma@1.7.x and 2.x, reporter version 0.x and 1.x. Seems that stops working as soon as the upgrade of Angular Client is done. Same repo with Angular Client 1.7 behaves correctly.

mboughaba:
@filipesilva can you please let us know if you think this is an issue with cli? if so is there a workaround?
any idea where to start the investigation?

DanKotowski:
Our team is also experiencing the same issue.

samneub:
My team also has this issue and it's fairly integral to our workflow. Hope this can get prioritized, thanks!

OClyde:
Same issue here, currently preventing our regular CI workflow to work optimally

b-johnse:
Same for code-coverage thresholds not exiting correctly.
I'm also seeing that if the browser crashes during tests (such as phantomJS) but the partial execution all passed it counts the whole step as successful. So perhaps the disconnect has to do with karma?

banjankri:
[...]

PR #11539, opened 2018-07-13
fix(@angular-devkit/build-angular) changing order of event and coverage reporters

banjankri:
Fixing an issue when event reporter is executed before coverage reporter, which was causing the test command to pass even though unit testing coverage thresholds were not met. Closes https://github.com/angular/angular-cli/issues/10940


alexw10:
@hansl has this been released? in version 6.1.0 I am still seeing this issue occur.

PR #11547, opened 2018-07-13
feat(@angular-devkit/core): better support for JSON5

hansl:
Supports for new lines and hexa was useful for doc and descriptions (including
schematics and builder declarations). Adding the other numerical constants was
easy enough. This fully completes JSON5 support.

### DIFF
--- a/packages/angular_devkit/core/src/json/parser.ts
+++ b/packages/angular_devkit/core/src/json/parser.ts
@@ -158,6 +158,38 @@ function _readExpNumber(context: JsonParserContext,
 
 
 /**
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/core/src/json/parser.ts
```diff
@@ -242,6 +298,7 @@ function _readString(context: JsonParserContext, comments = _readBlanks(context)
       };
     } else if (char == '\\') {
       char = _token(context);
+      debugger;
```
filipesilva: Leftover debugger statement.

hansl: Fixed.

PR #11559, opened 2018-07-16
fix(@angular-devkit/build_angular): use webpack 4 hooks for single run in karma plugin

Teamop:
Right now, `ng test --watch=false` command will get the warning `(node:15240) DeprecationWarning: Tapable.plugin is deprecated. Use new API on '.hooks' instead`
And as angular-cli uses the webpack 4 right now, I just refactor the code to use the new `hooks` api.


filipesilva:
LGTM, thanks!

PR #11574, opened 2018-07-17
fix(@angular-devkit/core): fix cannot delete directory in memory of virtual-fs

Teamop:
Seems typo in the code, so just fix the code and update the tests.

### DIFF
--- a/packages/angular_devkit/core/src/virtual-fs/host/memory.ts
+++ b/packages/angular_devkit/core/src/virtual-fs/host/memory.ts
@@ -180,7 +180,7 @@ export class SimpleMemoryHost implements Host<{}> {
     path = this._toAbsolute(path);
     if (this._isDirectory(path)) {
       for (const [cachePath] of this._cache.entries()) {
-        if (path.startsWith(cachePath + NormalizedSep)) {
+        if (cachePath.startsWith(path + NormalizedSep) || cachePath === path) {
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/core/src/virtual-fs/host/memory.ts
```diff
@@ -180,7 +180,7 @@ export class SimpleMemoryHost implements Host<{}> {
     path = this._toAbsolute(path);
     if (this._isDirectory(path)) {
       for (const [cachePath] of this._cache.entries()) {
-        if (path.startsWith(cachePath + NormalizedSep)) {
+        if (cachePath.startsWith(path)) {
```
hansl: If you have a file that starts with the path string but isn’t a folder (in your test create a file named `/subfile`) it will be deleted.

Teamop: I don't think a file will have the same name with a directory name under the same parent path, do you think right?

Teamop: oh, if the file with the same prefix with the path, but with different suffix, it will delete, I see

Teamop: done

Teamop:
@hansl could you review again? thanks

PR #11575, opened 2018-07-18
refactor(@ngtools/webpack): use webpack resolver plugin for path mapping

### LINKED ISSUE #10376: Module resolution is broken when wildcard path exists
### Versions
```
               |___/
    
Angular CLI: 1.7.4
Node: 9.9.0
OS: darwin x64
Angular: 5.2.10

```

### Repro steps

Add the following to tsconfig.json:

1. Create a new CLI project:

`ng new angular-cli-is-broken`

2. Add a wildcard path mapping to the tsconfig:

```js
"paths": {
  "*": ["./src/app/*"]
}
```

3. Build the project:

```
npm run build
```

### Observed behavior

Resolution of modules does not work:

```
ERROR in ./src/app/app.module.ngfactory.js
Module not found: Error: Can't resolve '/Users/jon/GitHub/angular-cli-is-broken/src/src/app/@angular/common' in '/Users/jon/GitHub/angular-cli-is-broken/src/app'
```

Etc.

### Desired behavior

It should resolve modules correctly, as it did in 1.6.

This worked perfectly in 1.6. You can confirm this by regressing the CLI version in package.json to 1.6, and confirming that `ng build` now works.

clydin:
Fixes NativeScript/nativescript-dev-webpack#607
Fixes #10376

Closes/Supersedes angular/devkit#753

### DIFF
--- a/packages/ngtools/webpack/src/paths-plugin.ts
+++ b/packages/ngtools/webpack/src/paths-plugin.ts
@@ -6,54 +6,97 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import * as path from 'path';
-import * as ts from 'typescript';
-import {
-  Callback,
-  NormalModuleFactoryRequest,
-} from './webpack';
+import { CompilerOptions, MapLike } from 'typescript';
+import { NormalModuleFactoryRequest } from './webpack';
 
+const getInnerRequest = require('enhanced-resolve/lib/getInnerRequest');
 
-export function resolveWithPaths(
-  request: NormalModuleFactoryRequest,
-  callback: Callback<NormalModuleFactoryRequest>,
-  compilerOptions: ts.CompilerOptions,
-  host: ts.CompilerHost,
-  cache?: ts.ModuleResolutionCache,
-) {
-  if (!request || !request.request || !compilerOptions.paths) {
-    callback(null, request);
+export interface TypeScriptPathsPluginOptions extends Pick<CompilerOptions, 'paths' | 'baseUrl'> {
 
-    return;
+}
+
+export class TypeScriptPathsPlugin {
+  constructor(private _options: TypeScriptPathsPluginOptions) { }
+
+  // tslint:disable-next-line:no-any
+  apply(resolver: any) {
+    if (!this._options.paths || Object.keys(this._options.paths).length === 0) {
+      return;
+    }
+
+    const target = resolver.ensureHook('resolve');
+    const resolveAsync = (request: NormalModuleFactoryRequest, requestContext: {}) => {
+      return new Promise<NormalModuleFactoryRequest | undefined>((resolve, reject) => {
+        resolver.doResolve(
+          target,
+          request,
+          '',
+          requestContext,
+          (error: Error | null, result: NormalModuleFactoryRequest | undefined) => {
+            if (error) {
+              reject(error);
+            } else {
+              resolve(result);
+            }
+          },
+        );
+      });
+    };
+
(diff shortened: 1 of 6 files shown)


tests/legacy-cli/e2e/tests/misc/module-resolution.ts
```diff
@@ -73,14 +73,14 @@ export default async function () {
 
   await updateJsonFile('tsconfig.json', tsconfig => {
     tsconfig.compilerOptions.paths = {
-      '@firebase/polyfill': ['@firebase/polyfill/index.ts'],
```
hansl: What's going on with these paths? It looks like breaking backward compatibility.

clydin: Bad test.  They were failing due to a bug (which this PR fixes) rather then the intended reason.  it should be failing due to non-project typescript files being pulled into the compilation which it now does.  Before it was failing due to missing files.

clydin:
Can hold off for 6.2 or even 7.0, if prefered.  This also cleans up and fixes several other behavioral discrepancies in the process.

packages/ngtools/webpack/src/paths-plugin.ts
```diff
@@ -112,63 +153,23 @@ export function resolveWithPaths(
...
-    request.request = path.resolve(compilerOptions.baseUrl || '', replacement);
-    callback(null, request);
-
-    return;
-  }
-
-  // TODO: The following is used when there is more than one potential and will not be
-  //       needed once this is turned into a full webpack resolver plugin
-
-  const moduleResolver = ts.resolveModuleName(
```
hansl: So we give up on TypeScript resolution, which fixed a load of issues and supports tsconfig's path mapping. See https://github.com/angular/angular-cli/pull/7473

hansl: I'm mostly curious of the two following assumptions in this PR:

1. we implement TypeScript's resolution ourselves, without any bugs, and
1. we rely on TypeScript never updating their resolution algorithm, otherwise this will likely break.

hansl: Another example is how we lose support for `rootDirs`, which I think we don't test for.

clydin: `rootDirs` never really worked properly (if ever).  It is also a fairly esoteric option and far more useful outside of bundling scenarios.  However, we could also add support for it.  The logic is very similar to the paths option but for relative imports.

#7473 actually demonstrates some of the issues with using `resolveModuleName`.  A large and ever growing list of special processing is needed to attempt to determine and further locate the correct file to be bundled.  As opposed to the correct file to provide type information for TypeScript transpilation.

Of note also is that the other two typescript loaders (`ts-loader`/`awesome-typescript-loader`) both take the same general approach of re-implementing the algorithm.
In regards to breakage, if TypeScript changed the algorithm or configuration options, it would result in a massive breaking change in regards to TypeScript projects in general.  Failure of the TypeScript team to notify the user base of such a pending change in advance would result in a large percentage of existing TypeScript projects to fail.  Advanced notification would mitigate the danger inherent in such a scenario.

Overall, the more ideal scenario would be if the underlying intrinsics for `resolveModuleName` within TypeScript were made public (someone alterations may also be necessary to allow for full re-usability;  this would need further analysis).  This could potentially allow this plugin to leverage more of the existing logic within TypeScript and limit the need to re-implement.  However, the full Webpack resolver plugin approach would still be the appropriate implementation option as it allows multiple resolution attempts and

Issue #8794, opened 2017-12-07
[app-shell] Doesn't work together with @angular/service-worker: "Error: Hash mismatch (cacheBustedFetchFromNetwork)" (+ explanation, + solution)


webmaxru:
cc: @alxhub @StephenFluin 

### Versions
```
Angular CLI: 1.6.0
Node: 8.2.1
OS: darwin x64
Angular: 5.1.0
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, platform-server, router

@angular/cli: 1.6.0
@angular-devkit/build-optimizer: 0.0.35
@angular-devkit/core: 0.0.22
@angular-devkit/schematics: 0.0.41
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.0
@schematics/angular: 0.1.10
@schematics/schematics: 0.0.10
typescript: 2.4.2
webpack: 3.10.0
```

### Repro steps
Generate a new app with --service-worker, then generate Universal app, then generate app-shell (if it fails, use workaround from https://github.com/angular/angular-cli/issues/8793) and run in the browser

### Observed behavior

At /ngsw/state:
```
NGSW Debug Info:

Driver state: EXISTING_CLIENTS_ONLY (Degraded due to failed initialization: Hash mismatch (cacheBustedFetchFromNetwork): http://localhost:8080/index.html: expected e809040220500c7210a197856aa3ecdf8ddd3618, got 100d6731232af53d7a592198666378360e223ba9 (after cache busting)
Error: Hash mismatch (cacheBustedFetchFromNetwork): http://localhost:8080/index.html: expected e809040220500c7210a197856aa3ecdf8ddd3618, got 100d6731232af53d7a592198666378360e223ba9 (after cache busting)
    at PrefetchAssetGroup.cacheBustedFetchFromNetwork (http://localhost:8080/ngsw-worker.js:657:27)
    at <anonymous>)
Latest manifest hash: 0f59e15ff1beb35d4ff048bb6cc056e34917081f
Last update check: 6m4s739u
```

### Desired behavior
Same run/update SW behavior as with non-app-shell node

### Mention any other details that might be useful (optional)
It happens because app-shell version of `index.html` generated after the generation of `ngsw.json`, where the hash of `index.html` located. So `ngsw.json` contains an incorrect hash, this is why @angular/service-worker throws this error after checking the integrity. Solution: make `ngsw.json` generation a last step in the flow.

[...]

PR #11583, opened 2018-07-18
Re-run ngsw after modifications in app-shell

filipesilva:
Fix https://github.com/angular/angular-cli/issues/8794

### DIFF
--- a/packages/angular_devkit/build_angular/src/app-shell/index.ts
+++ b/packages/angular_devkit/build_angular/src/app-shell/index.ts
@@ -11,10 +11,12 @@ import {
   BuilderConfiguration,
   BuilderContext,
 } from '@angular-devkit/architect';
-import { Path, getSystemPath, join, normalize, virtualFs } from '@angular-devkit/core';
+import { Path, getSystemPath, join, normalize, resolve, virtualFs } from '@angular-devkit/core';
 import { Observable, forkJoin, from, merge, of, throwError } from 'rxjs';
 import { concatMap, map, switchMap } from 'rxjs/operators';
 import { requireProjectModule } from '../angular-cli-files/utilities/require-project-module';
+import { augmentAppWithServiceWorker } from '../angular-cli-files/utilities/service-worker';
+import { BrowserBuilderSchema } from '../browser/schema';
 import { BuildWebpackServerSchema } from '../server/schema';
 import { BuildWebpackAppShellSchema } from './schema';
 
@@ -30,7 +32,9 @@ export class AppShellBuilder implements Builder<BuildWebpackAppShellSchema> {
       let success = true;
       const subscription = merge(
         this.build(options.serverTarget, {}),
-        this.build(options.browserTarget, { watch: false }),
+        // Never run the browser target in watch mode.
+        // If service worker is needed, it will be added in this.renderUniversal();
+        this.build(options.browserTarget, { watch: false, serviceWorker: false }),
       ).subscribe((event: BuildEvent) => {
         // TODO: once we support a better build event, add support for merging two event streams
         // together.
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/utilities/service-worker/index.ts
```diff
@@ -14,14 +14,20 @@ import * as fs from 'fs';
...
 export const NEW_SW_VERSION = '5.0.0-rc.0';
 
 
 class CliFilesystem implements Filesystem {
+  private _augmentedSwPaths = [
+    'ngsw.json',
+    'ngsw-worker.js',
+    'worker-basic.min.js',
+    'safety-worker.js',
+  ];
```
gkalpak: This hard-coded list feels very fragile :grin:

packages/angular_devkit/build_angular/src/app-shell/index.ts
```diff
@@ -109,25 +111,31 @@ export class AppShellBuilder implements Builder<BuildWebpackAppShellSchema> {
...
+    const builderConfig = architect.getBuilderConfiguration<BrowserBuilderSchema>({
       project,
       target,
       configuration,
     });
 
     return architect.getBuilderDescription(builderConfig).pipe(
       concatMap(description => architect.validateBuilderOptions(builderConfig, description)),
-      map(config => join(normalize(config.options.outputPath), 'index.html')),
+      // map(config => join(normalize(config.options.outputPath), 'index.html')),
```
gkalpak: Shouldn't this be removed?

packages/angular_devkit/build_angular/src/app-shell/index.ts
```diff
@@ -161,9 +169,21 @@ export class AppShellBuilder implements Builder<BuildWebpackAppShellSchema> {
           })
           .then((html: string) => {
             return this.context.host
-              .write(outputPath, virtualFs.stringToFileBuffer(html))
+              .write(outputIndexPath, virtualFs.stringToFileBuffer(html))
               .toPromise();
           })
+          .then(() => {
+            if (browserOptions.serviceWorker) {
+              return augmentAppWithServiceWorker(
```
gkalpak: I wonder if re-running this would introduce a noticeable delay for large projects (with lots of files to hash).

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/service-worker/index.ts
filipesilva: Those are the files we actually write back to disk further down on this file, so I think the knowledge of which files these are needs to be on the file anyway. Someone changing this file would have to change paths in both places though.

packages/angular_devkit/build_angular/src/app-shell/index.ts
filipesilva: Oops!

packages/angular_devkit/build_angular/src/app-shell/index.ts
filipesilva: That's a concern, yeah. I don't think `ngsw` has a way to just update the hashes for some files though. I don't want to duplicate the hashing logic it uses inside the CLI.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/service-worker/index.ts
```diff
@@ -14,14 +14,20 @@ import * as fs from 'fs';
...
-import { map, reduce, switchMap, concatMap, mergeMap, toArray, tap } from "rxjs/operators";
+import { map, reduce, switchMap, concatMap, mergeMap, toArray, tap, filter } from "rxjs/operators";
 import { Observable, merge, of, from } from "rxjs";
 
 
 export const NEW_SW_VERSION = '5.0.0-rc.0';
 
 
 class CliFilesystem implements Filesystem {
+  private _augmentedSwPaths = [
```
alxhub: Maybe extract this to a `const` instead of a private field, to further indicate these are hardcoded paths.

packages/angular_devkit/build_angular/src/app-shell/index.ts
alxhub: I'm not sure this is worth optimizing just yet.

Honestly, the "correct" solution to me is to not generate `ngsw.json` until the entire build is done, including the App Shell. Running it twice seems like a hack (but a hack that will work).

filipesilva:
@gkalpak @alxhub took a slightly different approach: when running the browser target always override `serviceWorker` to false, and then after rendering universal augment with service worker if the browser target had service worker on.

This way it only runs once and there's no need to change the service worker code. PTAL when you have time.

PR #11595, opened 2018-07-19
feat(@angular-devkit/core): support both 'targets' and 'architect' keys in workspace

filipesilva:


hansl:
Instead of doing `if`/`else` everywhere, is it simpler to rename the property when we load the file? Am I missing something?

filipesilva:
@hansl you mean in the schematics? That way you lose the original keys when the schematic updates the `angular.json` file.

We can do that, but that sounds a lot like sneaking a migration in while running a generator.

PR #11622, opened 2018-07-22
update ng-packagr to version 4

alan-agius4:
update `ng-packagr` to version 4

ng-packagr v4 requires that enableResourceInlining needs to be enabled for libraries that contain components

Relates also to https://github.com/angular/angular-cli/pull/11358


tests/angular_devkit/build_ng_packagr/ng-packaged/projects/lib/tsconfig.lib.json
```diff
@@ -21,7 +21,8 @@
     "fullTemplateTypeCheck": true,
     "strictInjectionParameters": true,
     "flatModuleId": "AUTOGENERATED",
-    "flatModuleOutFile": "AUTOGENERATED"
+    "flatModuleOutFile": "AUTOGENERATED",
+    "enableResourceInlining": true
```
alan-agius4: We need to properly communicate that if consumers are using `ng-packagr v3` when they update to `v4` they need to add this flag in their existing tsconfig.

filipesilva: WDYT of checking that this field is missing and adding a warning/error at runtime?

alan-agius4: I was actually thinking of switching this option to true by default in ng-packagr

filipesilva: That works too, and seems like you have already 👍 https://github.com/dherges/ng-packagr/pull/1021

alan-agius4: yeah :)

PR #7931, opened 2017-10-04
fix(@angular/cli): Disable comparisons feature in uglify compression in prod

Wykks:
See #5804

PR #11623, opened 2018-07-23
Patching 1.3 branch with ng build breaking on uglify per 1.5 branch

jellis:
Whilst I'm aware we probably should be upgrading, we've got a lot of applications that are only being built with 1.3.x - until we're able to get through and test all the breakages, it'd be great if we could get this merged in (as it has been for some other branches/versions).

### DIFF
--- a/packages/@angular/cli/models/webpack-configs/production.ts
+++ b/packages/@angular/cli/models/webpack-configs/production.ts
@@ -92,7 +92,7 @@ export const getProdConfig = function (wco: WebpackConfigOptions) {
     }));
   }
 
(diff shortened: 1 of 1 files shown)


jellis:
See https://github.com/angular/angular-cli/pull/7931 for reference point.

filipesilva:
Hi @jellis, I'm sorry to say but version `1.3` is outside of our LTS and is not kept up to date anymore. The best I can say is that you can use something like https://github.com/ds300/patch-package to change the package in place.

jellis:
Such is life.

PR #11628, opened 2018-07-23
fix(@angular-devkit/build-angular): should not log duplicate messages

Teamop:
Should not log duplicate messages for `specFailure` and `onRunComplete`, as they all have been handled by the `baseReporter`

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
@@ -240,6 +240,9 @@ const eventReporter: any = function (this: any, baseReporterDecorator: any) {
       failureCb && failureCb();
     }
   }
+
+  // avoid duplicate failure message
+  this.specFailure = () => {};
 };
 
 eventReporter.$inject = ['baseReporterDecorator'];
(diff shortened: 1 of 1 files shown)


filipesilva:
Hi, can you give me some context about the problem you are fixing? A way for me to see it happening would be important as well.

Teamop:
@filipesilva , yes it for `ng test`, right now, all these reporters will log the same message twice, so here is just avoid the duplicate messages from these reporters, just leave to `baseReporter` to deal with them

filipesilva:
I tried adding an error to an existing project, then introducing your changes, and I could still see duplicated error messages. Can you show me what do you see before and after your changes?

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
```diff
@@ -240,6 +240,9 @@ const eventReporter: any = function (this: any, baseReporterDecorator: any) {
       failureCb && failureCb();
     }
   }
+
+  // avoid duplicate failure message
+  this.specFailure = (_browser: any, result: any) => {};
```
filipesilva: This can be just `this.specFailure = () => {};`, and the same applies to the other ones.

Teamop:
hi @filipesilva, right now, with these changes, you will still see some duplicate, as these is also another reporter `progress` which comes from karma, and I have also submitted a PR karma-runner/karma＃3080

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
Teamop: done, thanks

Teamop:
@filipesilva could you restart all the CI tests? seems all of them aren't triggered

Teamop:
I will close and reopen to see if the CI can be restarted.

filipesilva:
LGTM, thanks for this fix!

PR #11630, opened 2018-07-23
JSON schema smart default schema and undefined transform feature improvements

clydin:


packages/angular_devkit/core/src/json/schema/visitor.ts
```diff
@@ -153,7 +150,14 @@ export function visitJson<ContextT>(
...
 ): Observable<JsonValue> {
-  return _visitJsonRecursive(json, visitor, buildJsonPointer([]), schema, refResolver, context);
+  return _visitJsonRecursive(
+    json,
+    visitor,
+    buildJsonPointer([]),
+    schema,
+    refResolver,
+    context,
+  ) as Observable<JsonValue>;
```
hansl: I'm not sure I follow this in context of `Observable<JsonValue | undefined>`. Why do you need both `| undefined` and the casting? Seems to me that the Observable _can_ contain undefined (since you don't filter them out).

clydin: that was an attempt to not alter the public signature of the function.

hansl: Let's filter out undefined just in case :)

clydin: Removed the changes to the file.  They were not material to the rest of the commit and the public API should probably receive a more thorough review for 7.0 anyway.

PR #11648, opened 2018-07-24
Fix 11034

### LINKED ISSUE #11034: LintFix config option disappeared in v6?! (ng generate produces invalid code!)
**Using Angular CLI v6.0.5. New Project**

In pre-v6 projects, in `".angular-cli.json"` under `"defaults"` section we had `"lintFix": true` setting, which ensured that after generating new component/service/whatever the linting fixes where performed automatically. Now in new `"angular.json"` (which replaces `".angular-cli.json"`) I can't find similar option. Which is a HUGE problem!

Suppose I have changed linter defaults to prefer "double quotes" for string (whoever said 'single quotes' should be default for TypeScript needs to be fired IMHO). Now every time I generate new component with `ng g component MyNewComponent`, it is created with all 'single quote' strings inside, instead of "double quotes"! And I have to manually fix them every time which is a nightmare! `lintFix` setting solved that in previous CLI versions but now, what is an alternative?!

hansl:
See commits for more information.

Fixes #11034


packages/angular/cli/models/architect-command.ts
```diff
@@ -88,8 +88,30 @@ export abstract class ArchitectCommand extends Command<ArchitectCommandOptions>
...
+          if (builder === null) {
+            builder = targetDesc.builder;
+          }
+          if (targetDesc.builder !== builder) {
+            isSameBuilder = false;
+          }
+        }
+
+        if (!isSameBuilder) {
+          throw new Error('Architect commands with multiple targets cannot specify overrides.'
```
filipesilva: Can you change this error message to indicate that the problem is not the multiple targets, but rather that they have different builders?

tests/legacy-cli/e2e/tests/generate/lint-fix.ts
```diff
@@ -14,13 +14,10 @@ export default function () {
...
   }`;
 
-  // TODO(architect): reenable after figuring out what happens to --lint-fix.
-  return;
-
   return Promise.resolve()
 
     // setup a double-quote tslint config
-    .then(() => writeFile('src/app/tslint.json', nestedConfigContent))
+    .then(() => writeFile('tslint.json', nestedConfigContent))
```
filipesilva: Nit: being at toplevel means it's not a tested config anymore.

packages/angular/cli/models/architect-command.ts
hansl: Done.

tests/legacy-cli/e2e/tests/generate/lint-fix.ts
hansl: Done.

PR #11649, opened 2018-07-25
fix: suggest windows-compatible typescript install command

alirezamirian:
Specifying version range in single quote doesn't work in windows CMD and it returns:
> The system cannot find the file specified.


alirezamirian:
Done. Sorry for the delay.

PR #11654, opened 2018-07-25
fix(@schematics/angular): update Angular to 6.1 for new projects

clydin:
Blocked on an Angular 6.1 final release


packages/schematics/angular/utility/latest-versions.ts
```diff
@@ -8,10 +8,10 @@
 
 export const latestVersions = {
   // These versions should be kept up to date with latest Angular peer dependencies.
-  Angular: '^6.0.3',
+  Angular: '^6.1.0',
```
filipesilva: This change will have to be a `feat()` since it changes the generated project.

clydin: done.

PR #11670, opened 2018-07-26
fix(@schematics/angular): increase version range for TypeScript

sarunint:


clydin:
Note that TS is used internally by the Angular schematics and is not required to be in lockstep with the version used to build the application.  (It would be preferable however).

@Brocco can you check to make sure that TS 2.9.2 will work with the transforms used in the Angular schematics?

hansl:
If the e2e tests pass this is good enough of a test for me.

PR #11680, opened 2018-07-27
Fix 11658

### LINKED ISSUE #11658: clarification: is ng update on 6.1.0 now connecting to cloudflare?
Hi, I use angular behind a corp firewall and after upgrading to 6.1.0 cli `ng update` I'm seing that the cli is trying to connect to a cloudflare ip:
```sh
cliente ➜ ng update
connect ECONNREFUSED 104.16.18.35:443
cliente ➜ 
```

this was not happening prior to 6.1.0. What is the cli trying to do with cloudflare? is this really necessary?

hansl:
fix(@schematics/update): use npm get for everything except _authToken 

For _authToken we read the npmrc in the global. We might need a better global solution
for other flags, but this will have to do for now.

Fixes #11658


filipesilva:
Somewhat related PR https://github.com/angular/angular-cli/pull/11659

clydin:
It's also closed now :)

packages/schematics/update/update/npm.ts
```diff
@@ -44,18 +113,10 @@ function getNpmConfigOption(
...
-
-  const optionValue = npmConfig && npmConfig[fullOption];
-  if (optionValue == undefined || optionValue == null) {
-    subject.next();
-  } else {
-    subject.next(optionValue);
-  }
-
-  subject.complete();
+  value = fullOption == '_authToken'
```
clydin: this should probably be `option.startsWith('_')`

hansl: Done. Investigating CI failures then I'll repush.

clydin: That should fix at least some of them.

Issue #11687, opened 2018-07-27
Version 6.1.0 not downward compatible 


NCC1701M:
### Bug Report or Feature Request (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
```

### Command (mark with an `x`)
At least!!!
```
- [ ] new
- [X] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [x] version
- [ ] doc
```

### Versions
Node: v8.11.2
npm: 5.6.0
ng: 6.1.1 (globale) 1.6.0 (locale)

#### Information
ng --version fails, too.

### Repro steps
- Use a Angular 5 project with cli ^1.6.0
- Install cli 6.1.1 globale `ng i -g @angular/cli@latest`
- Try to build A5 project

### The log given by the failure (ng build)
<pre>
C:\Temp\MyProject\node_modules\@angular\cli\ember-cli\lib\ui\index.js:56
    this.outputStream.write(data);
                      ^

TypeError: Cannot read property 'write' of undefined
    at UI.write (C:\Temp\MyProject\node_modules\@angular\cli\ember-cli\lib\ui\index.js:56:23)
    at UI.writeLine (C:\Temp\MyProject\node_modules\@angular\cli\ember-cli\lib\ui\index.js:69:8)
    at callback (C:\Temp\MyProject\node_modules\@angular\cli\tasks\build.js:44:29)
    at emitRecords.err (C:\Temp\MyProject\node_modules\webpack\lib\Compiler.js:265:13)
    at Compiler.emitRecords (C:\Temp\MyProject\node_modules\webpack\lib\Compiler.js:371:38)
    at emitAssets.err (C:\Temp\MyProject\node_modules\webpack\lib\Compiler.js:258:10)
    at applyPluginsAsyncSeries1.err (C:\Temp\MyProject\node_modules\webpack\lib\Compiler.js:364:12)
    at next (C:\Temp\MyProject\node_modules\tapable\lib\Tapable.js:218:11)
    at Compiler.compiler.plugin (C:\Temp\MyProject\node_modules\webpack\lib\performance\SizeLimitsPlugin.js:99:4)
    at next (C:\Temp\MyProject\node_modules\tapable\lib\Tapable.js:220:14)
[...]

PR #11690, opened 2018-07-27
fix(@angular/cli): properly bootstrap 1.x from global 6.x

clydin:
Fixes: #11687


hansl:
@alexeagle Some context; this code is (mostly) still from Ember CLI, and we just been fixing it instead of reworking it.

PR #11692, opened 2018-07-27
Move microsite to cli repo as static files

hansl:
And fixes.


etc/cli.angular.io/license.html
```diff
@@ -0,0 +1,23 @@
+<pre style="word-wrap: break-word; white-space: pre-wrap;">The MIT License
+
+Copyright (c) 2014-2016 Google, Inc.
```
alexeagle: is this the same way it's licensed in the original?

hansl: Yes, I only copied the file over. I'll add a commit to remove the years.

hansl: Note that the license file isn't pointed at anywhere. It was just part of the original website so I published it verbatim. Seems like all microsites have one (also with dates): https://universal.angular.io/license.html

PR #11697, opened 2018-07-28
fix(@angular-devkit/schematics): Fix issues for file-syetem-engine running in google3

qiyigg:
Pass 'name' as part of FileSystemCollectionDesc when creating
collection description.

### DIFF
--- a/packages/angular_devkit/schematics/tools/file-system-engine-host.ts
+++ b/packages/angular_devkit/schematics/tools/file-system-engine-host.ts
@@ -57,7 +57,10 @@ export class FileSystemEngineHost extends FileSystemEngineHostBase {
       throw new CollectionMissingSchematicsMapException(name);
     }
 
-    return desc as FileSystemCollectionDesc;
+    return {
+      ...desc,
+      name,
+    } as FileSystemCollectionDesc;
   }
 
   protected _transformSchematicDescription(


packages/angular_devkit/schematics/tools/file-system-engine-host.ts
```diff
@@ -36,6 +36,17 @@ export class FileSystemEngineHost extends FileSystemEngineHostBase {
       return join(this._root, name, 'collection.json');
     }
 
+    // Allow `${_root}/${name_without_@schematics_prefix}/collection.json.
+    const SCHEMATICS_PREFIX = '@schematics/';
```
hansl: I don't follow the logic in this. If it starts with `@schematics/` try without the prefix? Why not change the default collection instead?

qiyigg:
This is to minify the difference between internal and external CLI and how
they define and invoke schematics through short name.
e.g internally we could still use @schematics/angular or
@schematics/schematics to refer to a schematics

packages/angular_devkit/schematics/tools/file-system-engine-host.ts
qiyigg: If you think this is too much hack here, we could change the default collection instead. This is just in case we might use external schematics internally some day and external schematics can still resolve its dependent collection without local modification.

qiyigg:
@hansl removed the @schematics prefix support, we can add it later when we really need do so.

PR #11711, opened 2018-07-30
docs(@angular/cli): update Development Hints

noelmace:
Due to the changes the devkit reintegration to this repository have bring,
this old documentation didn't worked anymore, as the @angular/cli dependencies
versions now are dynamics. So, we now need to build it before linking it with npm.

### DIFF
--- a/packages/angular/cli/README.md
+++ b/packages/angular/cli/README.md
@@ -181,7 +181,7 @@ You can also use `ng new foo --link-cli` to automatically link the `@angular/cli
 Please read the official [npm-link documentation](https://docs.npmjs.com/cli/link)
 and the [npm-link cheatsheet](http://browsenpm.org/help#linkinganynpmpackagelocally) for more information.
 
-To run the Angular CLI test suite use the `node ./tests/legacy-cli/run_e2e` command.
+To run the Angular CLI E2E test suite, use the `node ./tests/legacy-cli/run_e2e` command.
 It can also receive a filename to only run that test (e.g. `node ./tests/legacy-cli/run_e2e tests/legacy-cli/e2e/tests/build/dev-build.ts`).
 
 As part of the test procedure, all packages will be built and linked.
--- a/packages/angular/cli/README.md
+++ b/packages/angular/cli/README.md
@@ -153,7 +153,7 @@ npm link
 ```
 
 `npm link` is very similar to `npm install -g` except that instead of downloading the package
-from the repo, the just builded `dist/@angular/cli/` folder becomes the global package.
+from the repo, the just built `dist/@angular/cli/` folder becomes the global package.
 Additionally, this repository publishes several packages and we use special logic to load all of them
 on development setups.
 
--- a/packages/angular/cli/README.md
+++ b/packages/angular/cli/README.md
@@ -146,17 +146,20 @@ You can find more details about changes between versions in [the Releases tab on
 
 ```bash
 git clone https://github.com/angular/angular-cli.git
-cd angular-cli
+npm install
+npm run build
+cd dist/@angular/cli
 npm link
 ```
 
 `npm link` is very similar to `npm install -g` except that instead of downloading the package
-from the repo, the just cloned `angular-cli/` folder becomes the global package.
+from the repo, the just builded `dist/@angular/cli/` folder becomes the global package.
 Additionally, this repository publishes several packages and we use special logic to load all of them
 on development setups.
 
 Any changes to the files in the `angular-cli/` folder will immediately affect the global `@angular/cli` package,
-allowing you to quickly test any changes you make to the cli project.
+meaning that, in order to quickly test any changes you make to the cli project, you should simply just run `npm run build`
+again.
 
 Now you can use `@angular/cli` via the command line:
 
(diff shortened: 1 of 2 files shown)


noelmace:
Has a matter of fact, I still have issues with this, as, after a new `npm run build`, I get : 
```
$ ng help

zsh: permission non accordée: ng
``` 
because `dist/@angular/cli/bin/ng` doesn't have execution rights.

But even after a chmod +x ng, I get : 
```
$ ng help

internal/modules/cjs/loader.js:596
    throw err;
    ^

Error: Cannot find module '@angular-devkit/core'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:594:15)
    at Function.Module._load (internal/modules/cjs/loader.js:520:25)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/home/noel/git/hub/angular/angular-cli/dist/@angular/cli/lib/init.js:13:16)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:713:10)
    at Module.load (internal/modules/cjs/loader.js:612:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:551:12)
    at Function.Module._load (internal/modules/cjs/loader.js:543:3)
```
The only workaround i could get for now is to re-run `npm i -g @angular/cli@next̀` and then `npm link`.

noelmace:
I also added some informations in order to help contributors to debug using VS Code like I do.

packages/angular/cli/README.md
```diff
@@ -146,17 +146,20 @@ You can find more details about changes between versions in [the Releases tab on
...
-cd angular-cli
+npm install
+npm run build
+cd dist/@angular/cli
 npm link
 ```
 
 `npm link` is very similar to `npm install -g` except that instead of downloading the package
-from the repo, the just cloned `angular-cli/` folder becomes the global package.
+from the repo, the just builded `dist/@angular/cli/` folder becomes the global package.
```
filipesilva: `builded`->`built`

packages/angular/cli/README.md
```diff
@@ -178,12 +181,34 @@ You can also use `ng new foo --link-cli` to automatically link the `@angular/cli
 Please read the official [npm-link documentation](https://docs.npmjs.com/cli/link)
 and the [npm-link cheatsheet](http://browsenpm.org/help#linkinganynpmpackagelocally) for more information.
 
-To run the Angular CLI test suite use the `node tests/run_e2e.js` command.
-It can also receive a filename to only run that test (e.g. `node tests/run_e2e.js tests/e2e/tests/build/dev-build.ts`).
+To run the Angular CLI test suite use the `node ./tests/legacy-cli/run_e2e` command.
```
filipesilva: Can you specify here that these instructions are for the E2E suite only please?

If you have time, instructions for the unit test and large unit test suite would be nice as well (`npm test` and `npm test -- --large`).

packages/angular/cli/README.md
```diff
@@ -178,12 +181,34 @@ You can also use `ng new foo --link-cli` to automatically link the `@angular/cli
...
+It can also receive a filename to only run that test (e.g. `node ./tests/legacy-cli/run_e2e tests/legacy-cli/e2e/tests/build/dev-build.ts`).
 
 As part of the test procedure, all packages will be built and linked.
 You will need to re-run `npm link` to re-link the development Angular CLI environment after tests finish.
 
+### Debugging with VS Code
```
filipesilva: This configuration must be launched from the CLI repo, right? I'm not too sure what situation is it useful in. One doesn't usually run any `ng` commands inside the repo proper.

packages/angular/cli/README.md
noelmace: No problem, I'll add the missing informations soon.

packages/angular/cli/README.md
noelmace: I used to do this in angular-cli before the devkit merge, in order to debug issues repo steps and get to the point more quickly. As you can see, the configuration use another project directory as "cwd". I didn't found how to do this well with npm link. I think we may need to add some script in order to link every built project and add a watch mode, but this would make this PR much more larger that it was intended to be. This section could go to another PR with this kind of script.

filipesilva: Ah, I had not noticed the `cwd` to another project!

filipesilva:
@noelmace I think the current CI failures are due to something already fixed in master, can you rebase please?

noelmace:
Rebase done.

filipesilva:
Thanks for updating our README 👍

noelmace:
You're welcome. BTW, if you could give some informations about how you proceed to debug issues, I would really appreciate that.

filipesilva:
To be honest I usually fiddle directly on node_modules and use the node inspector directly. The hardest things to debug end up being things that rely on resolution and that's very dependent on location and working off of built sources. So I usually don't rely as much on VSCode's debugger and just go and dig in. Stuff that's unit-testable I feel is good to use the built-in debugger though.

noelmace:
Ok thank you, that's what I'm also doing often, it's good to know that it isn't bad 😆. I'm trying to find a solution which would permit to directly test in progress development locally in every Angular CLI based projects, but yeah, I agree that testing is the way to go as much as possible. I'll push some PR when / if I'll find a satisfying enough solution about this other case then, right after the test docs you asked for.

PR #11717, opened 2018-07-31
docs: add reference to `vendor-source-map` flag

alan-agius4:


alan-agius4:
Superseeded by https://github.com/angular/angular-cli/pull/11732

PR #11729, opened 2018-07-31
fix(@angular-devkit/build-angular): load component source maps inline…

### LINKED ISSUE #9099: Source map not working for SCSS files
Recently I updated cli version 1.3.2 to 1.6.3 and after that I can't see scss source mapped file in developer tools in Chrome of Firefox .

### Versions : 1.6.3
```
Angular CLI: 1.6.3
Node: 8.9.3
OS: win32 x64
Angular: 5.0.0
... http, platform-browser, platform-browser-dynamic
... platform-server, router

@angular/cli: 1.6.3
@angular/language-service: 4.4.6
@angular-devkit/build-optimizer: 0.0.36
@angular-devkit/core: 0.0.22
@angular-devkit/schematics: 0.0.42
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.3
@schematics/angular: 0.1.11
@schematics/schematics: 0.0.11
typescript: 2.4.2
webpack: 3.10.0
```

### Repro steps
* Create new angular app using `ng new sample-app`
* Add scss file in angular-cli.json  `"styles": ["styles/styles.scss"]`
* Launch app using  `ng serve -sm -ec`
* Inspect styles using chrome dev-tools.

### Observed behavior
No actual scss file, only style.bundle.css where all the styles compiled. 
image

### Desired behavior
Actual scss file can be seen the developer tools with the line number
image

ajspera:
… so they work

without this, the way styles are into the DOM breaks the default sourcemap option
postcss-loader option docs
https://github.com/postcss/postcss-loader#inline
fixes #9099


ajspera:
I signed it!

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
```diff
@@ -267,7 +267,7 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
             options: {
               ident: buildOptions.extractCss ? 'extracted' : 'embedded',
               plugins: postcssPluginCreator,
-              sourceMap: cssSourceMap
+              sourceMap: cssSourceMap ? 'inline' : false
```
clydin: I'm assuming based on the linked issue that you meant to change the source map configuration for component styles located [here](https://github.com/ajspera/angular-cli/blob/4e569e575c0affdc8b015559fecf8fa9748bde6e/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts#L247) and not the global styles config.

ajspera: Actually both are affected by the problem, but I had missed that there were 2 instances of postcss-loader. With each using 'inline' instead of a boolean, both the global maps and component maps are now working. I updated the commit and the commit message to reflect that.

clydin: If that is the case then the comment in this section is very relevant.  `style-loader` 1.0 (unreleased) automatically handles the source maps in this regard.  As a stopgap this appears to be an acceptable solution.  However, if the global stylesheets are extracted, they should not contain inline source maps.  The conditional in this case should set the value to `true` if source maps and extraction is enabled; `inline` if only source maps are enabled; and `false` otherwise.

ajspera: I had thought about that, but when I tried `ng build --source-map --extract-css` without the inline sourcemaps, they did not work. The chrome inspector just referenced the generated css files. Looking a bit further, it does generate a map file for the styles, but there are no mappings to the scss files in it.
#### dist/styles.css.map
```
{"version":3,"sources":[],"names":[],"mappings":"","file":"styles.css","sourceRoot":""}
```
There may be another bug happening on top of this, although the always inline will resolve it as is until it's found. I suppose that's up to you whether you want extracted sourcemaps broken until it's found or supplement it as is (always inline if true) until then.

clydin: Extracted stylesheets should have working source maps.   If they are not working that would be a separate issue.  Altering the behavior for extracted souremaps in this way would be require a major version bump as end users may not realize the source maps are embedded resulting in the source being pushed unknowingly in production builds.

ajspera: It will still only push source maps if --source-map is explicitly set, but I suppose your point may be relevant if people are building source maps and then manually pulling the map files from their build as part of a deployment process which I would consider unorthodox. I do agree that is a separate issue, but its also related to the referenced issue since it isn't completely resolved until the extracted scss source maps also work.

clydin: If you are willing to make the suggested change, we should be able to roll this into the next minor.

ajspera: Cool, I should be able to get that in today. Just to confirm, the component styles are always injected inline via js even with extract-css, so the change just goes on the global styles, right?

clydin: yes.

ajspera: The change is up.

ajspera:
Sorry about the close and re-open. I had my gitlab account in my config when I rebased and reverting closed this. Everything should be good to go now, though, as long as the rebase fixed those failed tests.

ajspera:
@clydin A couple of rebases later and the appveyor CI finally passed. All automated checks are good, just needing approval.

PR #11735, opened 2018-08-01
refactor(@angular-devkit/schematics): add a BaseWorkflow which implem…

hansl:
…ents logic

And receives in its constructor the enginehost and registry. This simplifies the creation
of the NodeWorkflow, or later on the Google3Workflow etc, since all the duplicate logic
is now in a single base class.

This is yak shaving for internal stuff.


packages/angular_devkit/schematics/src/workflow/base.ts
```diff
@@ -0,0 +1,167 @@
...
+  host: virtualFs.Host;
+  engineHost: EngineHost<{}, {}>;
+  registry?: schema.CoreSchemaRegistry;
+
+  force?: boolean;
+  dryRun?: boolean;
+}
+
+
+export class BaseWorkflow implements Workflow {
```
alexeagle: should it be an abstract class? does it make sense to construct instances of this?

hansl: My thinking was that the code of NodeWorkflow was so simple that someone could create an instance of this directly and skip the middleman.

clydin: If it can be used directly `Base` seems somewhat inappropriate at least semantically.  `Core`? `Standard`?

packages/angular_devkit/schematics/src/workflow/base.ts
```diff
@@ -0,0 +1,167 @@
...
+import { Observable, Subject, concat, of, throwError } from 'rxjs';
+import { concatMap, defaultIfEmpty, ignoreElements, last, map, tap } from 'rxjs/operators';
+import { EngineHost, SchematicEngine } from '../engine';
+import { UnsuccessfulWorkflowExecution } from '../exception/exception';
+import { standardFormats } from '../formats';
+import { DryRunEvent, DryRunSink } from '../sink/dryrun';
+import { HostSink } from '../sink/host';
+import { HostTree } from '../tree/host-tree';
+import { Tree } from '../tree/interface';
+import { optimize } from '../tree/static';
```
clydin: Not something that should be addressed in this PR.  But is there something we can do to limit the volume of independent imports?  This seems like it could lead to a refactoring nightmare in the future.

hansl: Without having circular dependencies, not sure.

packages/angular_devkit/schematics/src/workflow/base.ts
hansl: After thinking a bit more, I'll keep the name and make this an abstract class. There are functionality missing from this that needs to be addressed at some points.

clydin: If there are no abstract methods then making the class abstract creates more of a handicap then anything else.

hansl: Making this right would be more work than just this refactor. This is to facilitate internal work.

What would be missing from this to be a full class;
1. a set of default transforms for the registry. you shouldn't pass the registry into the constructor.
1. a set of task executors and a way to pass in executors without being implementation dependent (ie. we need to be in a NodeJS 

The proper way to fix those (IMHO):
1. refactor the jobs API and use jobs instead of task executors (work in progress for 7.0)
1. define what a workflow is supposed to look like in _ALL_ environments. We don't have a clear view of what a node-free environment looks like right now because node is spread through the devkit packages.
1. having a clear interface separation between creating job types and their implementations, define basic types for the base workflow. This set might be empty, it's hard to tell for now.
1. Same for registry extensions. Have a default set of them and separate implementation from interface so base declares interfaces without implementations.
1. move the whole `tools/` directory out and into a separate package, into a `tools-node` and a `tools-base` or something.
1. Build all implementations for above base set of functionalities for Node.
1. make the workflow the default experience, and have node modules resolution be supported by the enginehost (NodeModulesEngineHost already does that) exclusively.
1. Remove any current subclass of Workflow because at this point we don't need them.
1. Make new subclasses for specialization of interfaces.
1. Create factories for node workflow that registers implementations of the default set of executors we want for each specialization of interfaces.

Does that make sense?

clydin: Ok. So this is more or less interim to unblock other work with further design work on the table for 7.  I'm good 👍

PR #11742, opened 2018-08-02
feat(@angular/cli): Update commands to extract meta data into JSON

Brocco:


packages/angular/cli/commands/build.json
```diff
@@ -0,0 +1,39 @@
...
+  "description": "Builds your app and places it into the output path (dist/ by default).",
+  "$longDescription": "",
+
+  "$aliases": [ "b" ],
+  "$scope": "in",
+  "$type": "architect",
+  "$impl": "./build#BuildCommand",
+
+  "type": "object",
+  "allOf": [
```
hansl: Why not sure `properties`?

packages/angular/cli/commands/add.json
```diff
@@ -0,0 +1,28 @@
...
+  "$schema": "http://json-schema.org/schema",
+  "id": "AddCommandOptions",
+  "description": "Add support for a library to your project.",
+  "$longDescription": "",
+
+  "$scope": "in",
+  "$impl": "./add#AddCommand",
+
+  "type": "object",
+  "allOf": [
```
hansl: Why not just `properties`?

packages/angular/cli/commands/config.json
```diff
@@ -0,0 +1,43 @@
...
+  "id": "ConfigCommandOptions",
+  "description": "Get/set configuration values.",
+  "$longDescription": "",
+
+  "$aliases": [],
+  "$scope": "in",
+  "$impl": "./config#ConfigCommand",
+
+  "type": "object",
+  "allOf": [
```
hansl: Same as above.

packages/angular/cli/custom-typings.d.ts
```diff
@@ -12,3 +12,26 @@ declare module 'yargs-parser' {
   const yargsParser: <T = any>(args: string | string[], options?: typeof parseOptions) => T;
   export = yargsParser;
 }
+
+declare module 'json-schema-traverse' {
+  export interface TraverseOptions {
+    allKeys?: boolean;
+  }
+  export type TraverseCallback = (
+    schema: any,
```
hansl: `JsonObject`?

packages/angular/cli/models/json-schema.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function getOptions(schemaText: string, onlyRootProperties = true): Promise<Option[]> {
+  // TODO: refactor promise to an observable then use `.toPromise()`
+  return new Promise((resolve, reject) => {
+    const fullSchema = parseJson(schemaText);
+    const traverseOptions = {};
+    const options: Option[] = [];
+    // tslint:disable-next-line:no-any
+    function postCallback(schema: any,
```
hansl: `JsonObject`.

packages/angular/cli/models/json-schema.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const callbacks = { post: postCallback };
+
+    jsonSchemaTravese(<object> fullSchema, traverseOptions, callbacks);
+  });
+}
+
+function isPropertyNested(jsonPath: string): boolean {
+  return jsonPath.split('/')
+    .filter(part => part === 'properties')
```
hansl: `|| part == 'items'` for arrays

packages/angular/cli/commands/build.json
clydin: is there a reason this is even a schema?  A majority of the actual fields are not even schema related.  An actual JSON file (with an associated schema) would seem more appropriate here.   The JSON schema elements can still be leveraged to define the option constraints and it would make the file far more intuitive.  There's also the advantage that the command schema can be used to validate all the command JSON files at build time (or in a CI check) to ensure proper formatting, required fields, etc.

packages/angular/cli/models/json-schema.ts
```diff
@@ -0,0 +1,80 @@
...
+        if (onlyRootProperties && isPropertyNested(jsonPointer)) {
+          includeOption = false;
+        }
+
+        const option: Option = {
+          name: property,
+          // ...schema,
+
+          description: schema.description,
+          type: schema.type,
```
clydin: This will fail for more complex constraints (`anyOf`/`allOf`/`not`/etc.).  type can potentially be an array as well and is also optional in which case anything is allowed.

packages/angular/cli/models/json-schema.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { JsonObject, JsonValue, parseJson } from '@angular-devkit/core';
+import * as jsonSchemaTravese from 'json-schema-traverse';
+import { Option, OptionSmartDefault } from './command';
+
+export async function convertSchemaToOptions(schema: string): Promise<Option[]> {
```
hansl: Is this called anywhere?

packages/angular/cli/models/json-schema.ts
```diff
@@ -0,0 +1,106 @@
...
+  const options = await getOptions(schema);
+
+  return options;
+}
+
+function getOptions(schemaText: string, onlyRootProperties = true): Promise<Option[]> {
```
hansl: Same with this.

Brocco: This is only an local function used by `convertSchemaToOptions`

packages/angular/cli/models/json-schema.ts
Brocco: This is only called by the command runner.

qiyigg:
@Brocco @hansl With this PR, the schematic-command.ts is still using fixed workflow (NodeJsWorkFlow), we still cannot specify which workflow to use. Even if we could specify workflow in the xx.json, we still don't want to maintain all the xx.json separately.

packages/angular/cli/models/command-runner.ts
```diff
@@ -70,18 +101,27 @@ export async function runCommand(commandMap: CommandMap,
     ? CommandScope.inProject
     : CommandScope.outsideProject;
 
-  let Cmd: CommandConstructor | null;
-  Cmd = commandName ? findCommand(commandMap, commandName) : null;
+  const commandMapPath = findUp('commands.json', __dirname);
```
qiyigg: This means if we internally want to have our own set of commands, we have to maintain a commands.json file under external CLI repo. Any way to avoid this?

packages/angular/cli/models/json-schema.ts
```diff
@@ -0,0 +1,106 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { JsonObject, JsonValue, parseJson } from '@angular-devkit/core';
+import * as jsonSchemaTravese from 'json-schema-traverse';
```
petebacondarwin: typo: Travese => Traverse

hansl:
Superseded by #11793

PR #11763, opened 2018-08-03
fix(@angular-devkit/build-angular): remove inlining of font files

alan-agius4:
Inlining font files which are less than 10kb in CSS causes some issues;
1. Fonts get downloaded with the stylesheet rather than one first usage. Since CSS resources get downloaded lazily.
2. Differnt font files are used by different browsers. For instance `woff2` cannot be used by older `IOS`, which means extra data over the wire that cannot be used.
3. When using a lot of font variants and there are under 10Kb, it quickly adds up to the bundle size.
4. Considering also `HTTP2` not inlining recouces might be also be more befenifcial in some cases.

//cc @filipesilva & @clydin


alan-agius4:
Superseeded by https://github.com/angular/angular-cli/pull/11826

PR #11766, opened 2018-08-03
Yarn Workspaces

clydin:
Advantages:
* Removes duplication of package dependencies in root
* Allows packages to have differing versions of dependencies then other packages
* Allows packages to have differing versions of dependencies then root development dependencies
* **Allows unit tests to run against the actual package dependency versions**
* Unit test only dependencies can now be defined within the using package

Notes:
* ~~build optimizer unit tests fail due to the actual dependency version of the package being used (TS 2.9).~~ The TS version used to compile the repo was recently updated to 2.9 and the tests were updated as a result.
* ~~Bazel appears to incorrectly resolve the internal TS Javascript type definitions when multiple TS versions are present (build optimizer exhibits this).~~ The TS version has been bumped to 2.9 and is now the same as build optimizer's version.  This still appears to be an outstanding typescript_rules bazel issue.
* A minimal set of yarn package resolutions is currently being used to mimic the versions previously present in the npm package lock.  Multiple test and code changes will be necessary to update the dependency versions to latest.
* The root package.json still contains several `dependencies` (vs. `devDependencies`) due to the implicit import lint rule that also affects the development/build script code.  The lint setup does not support nested tslint configs which currently limits the ability to have different sets of rules for the packages and the development code.


alan-agius4:
@clydin, you need to rebase, as you have some conflicts.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/service-worker/index.ts
```diff
@@ -21,7 +20,7 @@ import { Observable, merge, of, from } from "rxjs";
 export const NEW_SW_VERSION = '5.0.0-rc.0';
 
 
-class CliFilesystem implements Filesystem {
```
filipesilva: This doesn't seem to be a good change, since it will cause future changes to not respect the `Filesystem` constraints. 

Can you give some context about why this is needed?

.circleci/config.yml
```diff
@@ -18,39 +18,43 @@ jobs:
       - checkout: *post_checkout
```
filipesilva: If `post_checkout` is no longer used in any other job, we should move it from an anchor to just inline here.

.circleci/config.yml
```diff
@@ -18,39 +18,43 @@ jobs:
...
       - restore_cache: *_root_package_lock_key
       - run: yarn install --frozen-lockfile
+      - persist_to_workspace:
```
filipesilva: So the idea here is that we persist the post-install filesystem to the workflow-only FS layer, from which all other jobs inherit, then instead of caching `node_modules` proper we cache the yarn cache?

.circleci/config.yml
```diff
@@ -18,39 +18,43 @@ jobs:
       - checkout: *post_checkout
       - restore_cache: *_root_package_lock_key
       - run: yarn install --frozen-lockfile
```
filipesilva: This install has a couple of warnings on CI:
```
warning "workspace-aggregator-15c1ff0e-53e5-4313-986b-128baf16c8df > @angular-devkit/build-ng-packagr@0.0.0" has unmet peer dependency "ng-packagr@^2.2.0 || ^3.0.0 || ^4.0.0".
warning "workspace-aggregator-15c1ff0e-53e5-4313-986b-128baf16c8df > @angular-devkit/build-webpack@0.0.0" has unmet peer dependency "webpack@^4.6.0".
warning "workspace-aggregator-15c1ff0e-53e5-4313-986b-128baf16c8df > @angular-devkit/build-webpack@0.0.0" has unmet peer dependency "webpack-dev-server@^3.1.4".
warning "workspace-aggregator-15c1ff0e-53e5-4313-986b-128baf16c8df > @ngtools/webpack@0.0.0" has unmet peer dependency "webpack@^4.0.0".
warning "workspace-aggregator-15c1ff0e-53e5-4313-986b-128baf16c8df > @angular-devkit/build-ng-packagr > ng-packagr@4.0.1" has unmet peer dependency "@angular/compiler@^6.0.0".
warning "workspace-aggregator-15c1ff0e-53e5-4313-986b-128baf16c8df > @angular-devkit/build-ng-packagr > ng-packagr@4.0.1" has unmet peer dependency "@angular/compiler-cli@^6.0.0".
warning "workspace-aggregator-15c1ff0e-53e5-4313-986b-128baf16c8df > @angular-devkit/build-ng-packagr > ng-packagr@4.0.1" has unmet peer dependency "tslib@^1.9.0".
```

.circleci/config.yml
filipesilva: I compared two CI runs to see how the install speed changed and it looks like the initial install took a while longer (~2m vs ~1m) but the other jobs didn't really change. 

I'm not too worried about the 1m difference and tbh it might just be because this particular run did not hit the yarn cache.

.circleci/config.yml
```diff
@@ -18,39 +18,43 @@ jobs:
...
           paths:
-            - "node_modules"
+            - ~/.cache/yarn
   
   lint:
     <<: *defaults
     steps:
-      - checkout: *post_checkout
-      - restore_cache: *_root_package_lock_key
+      - attach_workspace:
```
filipesilva: Perhaps `attach_workspace` should be an anchor? The copy paste amount would be similar but if in the future we change the attach root (or any other `attach_workspace` options) we only have to change it in one place.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/service-worker/index.ts
clydin: `@angular/service-worker` was an implicit dependency but due to the previous root package structure it was not caught.  This was already problematic as the version of the library directly imported can potentially differ from the version actual used and executed further in the file.  That version is required from the project itself whereas this one relies solely on node resolution.

.circleci/config.yml
clydin: I left it to maintain commonality with core's usage.  If it is ever updated there it could then be easily update here as well.  Can change it if you prefer.

.circleci/config.yml
clydin: The advantage is that a setup is generated once in the first install job and then the reused pristinely within each of the workflow jobs.  It essentially guarantees each job within a particular workflow execution is run against an identical setup.  There's also the advantage that external network access (npm registry, github, etc.) is limited to the first step (excepting individual test needs within a job).

.circleci/config.yml
clydin: Yes.  This is a yarn issue.  It is not accounting for hoisted dev dependencies in workspaces.
ref: https://github.com/yarnpkg/yarn/issues/5810

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/service-worker/index.ts
filipesilva: We can still import the type, and use `// tslint:disable-line:no-implicit-dependencies`. This way we don't lose the type safety while building.

Do do this for TSLint in `build-angular` for instance.

.circleci/config.yml
filipesilva: Yeah having the exact same copy is a really good guarantee.

.circleci/config.yml
clydin: Done.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/service-worker/index.ts
clydin: True.  Also cleaned up some of the lint issues and used the types throughout.

.circleci/config.yml
filipesilva: I think that's a fair reason really, no need to change it IMHO.

PR #11791, opened 2018-08-06
feat(@angular-devkit/build-angular): support dart-sass for SASS/SCSS

clydin:
This provides the ability for a developer to use the pure javascript version of sass, if desired.  It will try to use dart-sass (i.e., npm package `sass`) if the package is manually installed and fallback to `node-sass` if not found.  In this case it is found, it will also try to use `fibers` if available as well.

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -37,7 +37,7 @@
     "postcss-url": "^7.3.2",
     "raw-loader": "^0.5.1",
     "rxjs": "^6.0.0",
-    "sass-loader": "~6.0.7",
+    "sass-loader": "^7.1.0",
     "semver": "^5.5.0",
     "source-map-support": "^0.5.0",
     "source-map-loader": "^0.2.3",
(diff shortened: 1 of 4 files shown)


hansl:
Please talk to @jelbourn before continuing this.

jelbourn:
Some points to consider: 

* dart-sass _is_ now the canonical implementation, meaning that it will always be the first to have new features and fixes
* dart-sass is compiled to JavaScript for the `sass` npm package, meaning that it won't run into the same platform/compilation issues that `node-sass` users may encounter
* The javascript-from-dart version of the compiler is still noticeably slower than node-sass. This will hopefully get better over time 

You probably want to get some data on actual performance differences before fully committing to a decision.

clydin:
Definitely agree on the performance concerns.  The main goal of this PR was to determine minimum viability (i.e., CI passes).

Further benchmarking would be need on several SCSS-heavy projects.

Also, one potential option here is to automatically use the `fibers` package if present within the project.  This would have the advantage of allowing SCSS-heavy projects to at least mitigate the performance penalty, if desired, while also not requiring additional native packages (the purpose of moving away from node-sass).

alan-agius4:
@clydin if you need any help to test this against a large scss project, let me know.

 I have a private project which has loads of scss file which unfortunately I cannot share.

mattdsteele:
I would love for this to be available as an option.

For context, our organization has firewall rules that block downloading `node-sass` binaries from GitHub, so we have some complex tooling and NPM config to publish and consume the binaries in-house. Moving to dart-sass would simplify our environment quite a bit.

PR #11793, opened 2018-08-06
Commands json schema

hansl:


hansl:
This is a follow up and replaces #11742

hansl:
Will ignore CLA Bot as Mike signed it at the time of his PR, and I obviously signed it too :)

petebacondarwin:
Why are none of the schema properties "required"?

hansl:
Smart defaults (the `$default` property) don't work well right now with required :/

packages/angular/cli/commands/generate.ts
```diff
@@ -60,7 +48,8 @@ export class GenerateCommand extends SchematicCommand {
     const [collectionName, schematicName] = this.parseSchematicInfo(options);
 
     // remove the schematic name from the options
-    options._ = options._.slice(1);
+    // options._ = options._.slice(1);
```
filipesilva: Leftover commented code.

packages/angular/cli/models/command.ts
```diff
@@ -51,47 +55,76 @@ export abstract class Command<T = any> {
...
+      this.logger.info(description);
+    }
+    this.printHelpUsage(commandName, this.options);
     this.printHelpOptions(this.options);
   }
 
-  protected printHelpUsage(name: string, args: string[], options: Option[]) {
+  private getArgIndex(def: OptionSmartDefault | undefined): number {
+    if (def === undefined || def.index === undefined || typeof def.index !== 'number') {
+      return 99999;
```
filipesilva: I don't think we'll get to 99999 args but this looks like it should be a non-number type, probably `null`.

packages/angular_devkit/schematics/tools/index.ts
```diff
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 export * from './description';
+export * from './export-ref';
```
filipesilva: This change to the public API should be a feat in it's own commit.

packages/angular/cli/models/command-runner.ts
```diff
@@ -70,18 +101,27 @@ export async function runCommand(commandMap: CommandMap,
     ? CommandScope.inProject
     : CommandScope.outsideProject;
 
-  let Cmd: CommandConstructor | null;
-  Cmd = commandName ? findCommand(commandMap, commandName) : null;
+  const commandMapPath = findUp('commands.json', __dirname);
```
qiyigg: Possible to make the config file path as a parameter?

packages/angular_devkit/schematics/tools/index.ts
hansl: `ExportRef` should be part of core anyway.

packages/angular/cli/commands/generate.ts
hansl: Done.

packages/angular_devkit/schematics/tools/index.ts
hansl: It's in a feature. This was from Mike's PR so I don't want to refactor it. I'll refactor it later (would like to move it out of schematics).

packages/angular/cli/models/command-runner.ts
hansl: Done.

packages/angular/cli/commands/add.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "id": "AddCommandOptions",
+  "description": "Add support for a library to your project.",
+  "$longDescription": "",
+
+  "$scope": "in",
+  "$impl": "./add-impl#AddCommand",
```
qiyigg: It seems we still have to maintain a bunch of ${command}.json files internally. I don't know how often will these json files change? Will it make sense for us to maintain a different set of json files?

hansl: I want to unify schematics, architect and cli json schema files. There is already one code path but I want to make it more streamlined. This could also open up the way to custom commands down the road when we want to.

hansl:
@googlebot Please rerun.

PR #11794, opened 2018-08-06
feat(@angular/cli): allow specifying workflow and engineHost for schematic commands

qiyigg:


packages/angular/cli/models/schematic-command.ts
```diff
@@ -7,17 +7,36 @@
  */
 
 // tslint:disable:no-global-tslint-disable no-any
-import { JsonObject, experimental } from '@angular-devkit/core';
-import { normalize, strings, tags, terminal, virtualFs } from '@angular-devkit/core';
+import { JsonObject,
```
hansl: Please reformat your file not using clang format. I think @filipesilva uses prettier which works with the correct settings.

qiyigg: Actually, I formatted it manually... @filipesilva @hansl  Any commands can I use? I tried 'npm run admin -- lint --fix',  it doesn't fix the 'import'.

filipesilva: I don't use `prettier` but rather VS Code auto format which is not enforceable since it's editor specific.

In general for multiline imports everywhere else in the CLI we do it as below, so that's my preference.
```
import {
  import1,
  import2,
  // etc
} from 'package';
```

qiyigg: Update import accordingly.

packages/angular/cli/models/schematic-command.ts
```diff
@@ -44,15 +71,36 @@ export interface GetOptionsResult {
...
+  private _engineHost: FileSystemEngineHostBase;
+  private _engine: Engine<FileSystemCollectionDesc, FileSystemSchematicDesc>;
+  private _workFlow: workflow.BaseWorkflow;
   argStrategy = ArgumentStrategy.Nothing;
 
+  constructor(
+      context: CommandContext, logger: logging.Logger,
+      engineHost?: FileSystemEngineHostBase) {
+    super(context, logger);
+    this._engineHost =
```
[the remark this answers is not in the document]
qiyigg: done

PR #11815, opened 2018-08-08
Fix(@schematics/angular): support `Fetch as Google`（Google bot)

rdlabo:
Fetch as Google (Google's bot checker) use ES5. So ES5 require polyfills (`Fetch as Google` need `core-js/es6/array` only ).
I think that Angular needs a response to Google in order to become a PWA. 

thanks.

### DIFF
--- a/packages/schematics/angular/application/files/src/polyfills.ts
+++ b/packages/schematics/angular/application/files/src/polyfills.ts
@@ -28,11 +28,18 @@
 // import 'core-js/es6/math';
 // import 'core-js/es6/string';
 // import 'core-js/es6/date';
-// import 'core-js/es6/array';
 // import 'core-js/es6/regexp';
 // import 'core-js/es6/map';
 // import 'core-js/es6/weak-map';
 // import 'core-js/es6/set';
+
+/**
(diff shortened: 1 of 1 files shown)


filipesilva:
@alexeagle are you familiar with this situation?

lacolaco:
FYI: Googlebot uses a renderer following the similar spec to Chrome 41.
https://developers.google.com/search/docs/guides/rendering

ES6 support is almost dropped.
https://caniuse.com/#compare=chrome+41

I think that current polyfills set is no bad, but it should be mentioned that crawlers possibly requires ES5 support. 
I hope comments for the crawlers support will be placed in polyfill.ts as well as IE support.

hansl:
Hi @rdlabo ,

This PR looks good, and I'd like to put it in. I agree that support for google bot is important. However, I feel this should be left to the user and not be on by default. Could you keep your comment but comment out the import statement?

Also, please rebase on top of master. Things have changed quite a bit since this PR was made.

Thanks!

alexeagle:
@clydin will make a minor edit and push to this branch to fix it up

rdlabo:
@hansl Thanks for review. I made and rebased!

PR #11836, opened 2018-08-09
Schematic Prompts

clydin:


packages/angular_devkit/core/src/json/schema/interface.ts
```diff
@@ -74,7 +74,7 @@ export interface SchemaFormat {
 }
 
 export interface SmartDefaultProvider<T> {
-  (schema: JsonObject): T | Observable<T>;
+  (schema: JsonObject): T | Observable<T> | Promise<T>;
```
hansl: This makes smart default providers async. I'd prefer we stick to observables.

clydin: The return value is coming from third party code.  The observable can't be assumed to be synchronously completed.  Not having the promise as an option forces consumers to directly depend on rxjs and wrap all promise generating code in a `from`.  It also prevents the direct use of an async function which is the use case here.

The schema validation itself also needs to be assumed async since ajv can potentially complete either synchronously or asynchronously.

packages/angular/cli/commands/new.json
```diff
@@ -33,6 +33,11 @@
       "default": false,
       "aliases": ["v"],
       "description": "Adds more details to output logging."
+    },
+    "non-interactive": {
```
hansl: No negative flags please. Make the default to `true`.

clydin: Done.

PR #11841, opened 2018-08-10
fix(@angular-devkit/schematics): throw `InvalidCollectionJsonExceptio…

### LINKED ISSUE #11818: dev-kit/schematics: collection.json incorrectly formatted error swallowed
```
- [] bug report -> please search issues before submitting
- [X] feature request
```
If the `collection.json` is not formatted in proper JSON schematics pukes and the error message is misleading. I understand the root cause was user error. However, the error message returned lead me down a deep path of questioning how the schematic was being resolved. Only when I started debugging the dev-kit/schematic code, was I able to determine the root of the issue. 

### Repro steps
When executing a schematic if the `collection.json` is not json formatted (in my case I was missing a brace) and error occurs.

```
Collection "@company/package" cannot be resolved.
Error: Collection "@company/package" cannot be resolved.
    at NodeModulesEngineHost._resolveCollectionPath (C:\test-schematic\node_modules\@angular-devkit\schematics\tools\node-module-engine-host.js:88:15)
    at NodeModulesEngineHost.createCollectionDescription (C:\test-schematic\schem\node_modules\@angular-devkit\schematics\tools\file-system-engine-host-base.js:106:27)
....
```

### Desired functionality
Add a better error description would be useful, or do not swallow error.

Code [located at ln 104 in `node-module-engine-host.ts`](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/schematics/tools/node-module-engine-host.ts#L104)

alan-agius4:
throw `InvalidCollectionJsonException` when collection file is invalid

Closes #11818

Not quite sure, how can I test this, since I cannot use the `node-modules-test-engine-host`

### DIFF
--- a/packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
+++ b/packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
@@ -7,7 +7,9 @@
  */
 import {
   BaseException,
+  InvalidJsonCharacterException,
   JsonObject,
+  UnexpectedEndOfInputException,
   isObservable,
   normalize,
   virtualFs,
@@ -52,8 +54,18 @@ export class CollectionCannotBeResolvedException extends BaseException {
   }
 }
 export class InvalidCollectionJsonException extends BaseException {
-  constructor(_name: string, path: string) {
-    super(`Collection JSON at path ${JSON.stringify(path)} is invalid.`);
+  constructor(
+    _name: string,
+    path: string,
+    jsonException?: UnexpectedEndOfInputException | InvalidJsonCharacterException,
+  ) {
+    let msg = `Collection JSON at path ${JSON.stringify(path)} is invalid.`;
+
+    if (jsonException) {
+      msg = `${msg} ${jsonException.message}`;
+    }
+
+    super(msg);
   }
 }
 export class SchematicMissingFactoryException extends BaseException {
--- a/packages/angular_devkit/schematics/tools/node-module-engine-host.ts
+++ b/packages/angular_devkit/schematics/tools/node-module-engine-host.ts
@@ -5,7 +5,11 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { BaseException } from '@angular-devkit/core';
+import {
+  BaseException,
+  InvalidJsonCharacterException,
+  UnexpectedEndOfInputException,
+} from '@angular-devkit/core';
 import * as core from '@angular-devkit/core/node';
 import { dirname, join, resolve as resolvePath } from 'path';
 import { RuleFactory } from '../src';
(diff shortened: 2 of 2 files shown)


packages/angular_devkit/schematics/tools/node-module-engine-host.ts
```diff
@@ -102,7 +102,11 @@ export class NodeModulesEngineHost extends FileSystemEngineHostBase {
         return collectionPath;
       }
     } catch (e) {
+      if (!(e instanceof FileDoesNotExistException)) {
+        throw new InvalidCollectionJsonException(name, collectionPath);
```
hansl: Could you just rethrow the error instead? The sub error can be a JSON error which contains more information (line and character), or another error.

alan-agius4: I initially was gonna do did that, the only drawback was the in the error there won’t be information about which files cause the JSON error. 

An alternative would be that `InvalidCollectionJsonException` would accept another parameter, and eventually if error.message is present it will be concatinated with original message that `InvalidCollectionJsonException` is showing to the user.

What do yo think?

alan-agius4: Something similar to this error https://github.com/angular/angular-cli/issues/11861 with no context of which file is the problem.

hansl: I think either solution would be better:

1. An alternative would be that InvalidCollectionJsonException would accept another parameter of type `InvalidJsonException` and transfer over the info, or
2. A separate new `InvalidJsonFileException` which takes a path and the json exception.

I'm fine with either. I have a slight preference to 1.

alan-agius4: I also prefer for the first option. I have pushed an updated. When you said `transfer over the info` I was not sure if you were referring the error message or stack or both. For the time being I did the message.

alan-agius4:
@hansl, can you re-review this please? Thanks.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
```diff
@@ -91,16 +103,16 @@ export class SchematicNameCollisionException extends BaseException {
  * all other EngineHost provided by the tooling part of the Schematics library.
  */
 export abstract class FileSystemEngineHostBase implements
-    EngineHost<FileSystemCollectionDescription, FileSystemSchematicDescription> {
+  EngineHost<FileSystemCollectionDescription, FileSystemSchematicDescription> {
```
hansl: Why all the style changes? Your PR should only include code you changed.

alan-agius4: Yeah agreeded  have done Autoformat when fixing the rebase conflict. 

Will fix this later

Issue #11738, opened 2018-08-01
Question/Documentation: How to use angular-cli when not globally installed?


danielhelfand:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Command (mark with an `x`)
```
All commands
```

### Versions
`angular-cli 6.1.1`
`npm 5.6.0`

### Repro steps
Using CentOS 7 OS, install `npm`:
```
curl -sL https://rpm.nodesource.com/setup_8.x | bash -
yum install -y nodejs
```

Install angular-cli non globally:
```
npm install @angular/cli@6.1.1
```

### The log given by the failure
`bash: ng: command not found`

### Desired functionality
Ability to install and use multiple versions of the angular-cli on a CentOS 7 OS. I would like to know if this is possible, and I would like for there to be documentation on how to do this.

ctrl-brk:
You should be able to access it from node_modules/.bin/ng

alan-agius4:
You can you it like such
```
 npm run ng <command> 
```

there are also predefined scripts in your package.json like `npm start` and  `npm test` etc...

Alternatively you could also use `npx` https://www.npmjs.com/package/npx

hansl:
Closing this as its been answered.

danielhelfand:
@hansl Is there documentation anywhere for this? If not, I would like to keep this issue open. I can open a separate issue, but I think it would be better to keep it here.

SangKa:
Hi @alan-agius4 , why `npm run ng <command>`  works when not globally cli installed? Does npm can figure out the `ng` means `node_modules/.bin/ng` ?
Look forward to your reply. Thanks.

alan-agius4:
Yes, this is based on how node resolution work.

As when using a node package inside an `npm script` it will try locate the module locally first and then it will look for them globally.

SangKa:
Thanks for your inspiring, I found the reason in the [npm document](https://docs.npmjs.com/cli/run-script).

In addition to the shell's pre-existing PATH, npm run adds `node_modules/.bin` to the PATH provided to scripts. Any binaries provided by locally-installed dependencies can be used without the `node_modules/.bin` prefix. For example, if there is a devDependency on tap in your package, you should write:

```
"scripts": {"test": "tap test/\*.js"}
```

instead of

```
"scripts": {"test": "node_modules/.bin/tap test/\*.js"}  
```

to run your tests.

PR #11864, opened 2018-08-11
docs: add local installation and specific version installation instructions for angular-cli

danielhelfand:
This pull request adds documentation on how to locally install and use the angular-cli as well as installation instructions on installing specific versions of the angular-cli.

This addresses issues raised in #11738 and includes advice that came as a result of the issue.


danielhelfand:
@googlebot I signed it!

PR #11874, opened 2018-08-13
feat(@schematics/angular): remove useless import for Ivy

cexbrayat:
Using cli `6.2.0-beta.2` with the new `experimentalIvy` flag leads to:

    ERROR in src/app/app.module.ts(1,1): error TS6133: 'BrowserModule' is declared but its value is never read.

### DIFF
--- a/packages/schematics/angular/application/other-files/app.module.ts
+++ b/packages/schematics/angular/application/other-files/app.module.ts
@@ -1,6 +1,6 @@
-import { BrowserModule } from '@angular/platform-browser';
+<% if (!experimentalIvy) { %>import { BrowserModule } from '@angular/platform-browser';<% } %>
 import { NgModule } from '@angular/core';
-<% if (routing) { %>
+<% if (routing && !experimentalIvy) { %>
 import { AppRoutingModule } from './app-routing.module';<% } %>
 import { AppComponent } from './app.component';
 


packages/schematics/angular/application/other-files/app.module.ts
```diff
@@ -1,4 +1,4 @@
-import { BrowserModule } from '@angular/platform-browser';
+<% if (!experimentalIvy) { %>import { BrowserModule } from '@angular/platform-browser';<% } %>
```
alan-agius4: I think there is the same problem for `AppRoutingModule`, as at the moment it looks like this is only being used if it's not `experimentalIvy`

https://github.com/angular/angular-cli/blob/151a0b2cec417ad8dc6637dc88f4684cafff3642/packages/schematics/angular/application/other-files/app.module.ts#L11-L13

I think that `AppRoutingModule` shouldn't be effect by the fact if  `experimentalIvy` is enabled or disabled. 

That said, maybe @hansl can confirm about this.

cexbrayat: You're right. I generated it without the `routing` flag so I didn't see it, but I can add it if @hansl confirms.

hansl: Yes. I don't even know if we support router in Ivy.

cexbrayat: I amended the commit to remove the `AppRoutingModule` if the `experimentalIvy` flag is used.

cexbrayat:
@alexeagle no worries, it's not in the default tsconfig. Sorry if my description was misleading: I just wanted to point out that the useless imports should be removed.

vikerman:
Rerunning failed tests...

vikerman:
Please assign appropriate merge target

PR #11889, opened 2018-08-14
feat: add typescript 3 as peer dependency and repo devDep

hansl:
The default being created is still 2.9.2.


packages/ngtools/webpack/package.json
```diff
@@ -31,11 +31,10 @@
     "webpack-sources": "^1.1.0"
   },
   "peerDependencies": {
-    "typescript": "~2.4.0 || ~2.5.0 || ~2.6.0 || ~2.7.0 || ~2.8.0 || ~2.9.0",
+    "typescript": "~2.4.0 || ~2.5.0 || ~2.6.0 || ~2.7.0 || ~2.8.0 || ~2.9.0 || ~3.0.0",
```
clydin: can shorten this to `2.4 - 3.0`

alan-agius4:
Blocking issue in `angular/angular` https://github.com/angular/angular/pull/25275

Side note, it looks like Angular 7 will support only TS 3+. Looking at the change @IgorMinar did https://github.com/angular/angular/pull/25275/commits/f3eb7cf9b739be2d69bc40f810eca446570d6be2

hansl:
Yeah we will update `latestVersions` when we need to.

PR #11925, opened 2018-08-17
fix(@schematics/angular): generate application failing when projectRoot is provided

vovikdrg:
Right now generate application will be written to projects folder same as  libraries, but its possible to provide location by projectRoot. 

Problem is that e2e tests path is wrong and tasks fails.

This PR is to fix it and allow to use projectRoot parameter as it was designed

### DIFF
--- a/docs/documentation/generate/application.md
+++ b/docs/documentation/generate/application.md
@@ -96,3 +96,12 @@ Create an Angular application.
     Do not add dependencies to package.json.
   </p>
 </details>
+<details>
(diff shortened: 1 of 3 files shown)


vovikdrg:
I signed it!

vovikdrg:
I would appreciate if somebody guide me how to cover this by tests

hansl:
That Appveyor failure is a flake. Merging this in without the doc changes.

PR #11964, opened 2018-08-21
feat(@schematics/angular): let Universal schematic use configurations from build architect

beeman:
When using Angular Universal you generally want to use the same `configurations` for the browser and the server app, mainly copying over the correct `environment` file on a production build.

This is a practice also seen in the [universal-starter](https://github.com/angular/universal-starter/blob/master/angular.json#L115-L124) project.

This PR copies over the configuration from the `build` architect, saving the user from making this change manually.

The nice thing about this approach is that it allows both a `development` and `production` build to be created:

```
# Build using the development environment
ng run my-app:server
# Build using the production environment
ng run my-app:server:production
```


packages/schematics/angular/package.json
```diff
@@ -9,6 +9,7 @@
   ],
   "schematics": "./collection.json",
   "dependencies": {
+    "@angular-devkit/architect": "0.0.0",
```
filipesilva: While it does make sense to add Architect here, we've tried not to do it because the schematics shouldn't strictly need it. It's a bit murky really... but it would be better to not add the dependency at the moment and wait until we figure out the intended deps.

beeman: Makes sense, removed the dep!

alexeagle:
@filipesilva to add a JIRA task for us to resolve the newly introduced `TODO`

PR #11979, opened 2018-08-23
feat(@schematics/angular): add several prompts to ng-new

clydin:
Initial prompt usage when executing `ng new` for 7.0.

When executing `ng new` (no options are required now), the user will be prompted for the following:
* name of the project
* selector prefix for components
* whether they will use Angular routing
* stylesheet format (css, scss, etc.)

If the user provides the option on the command line (e.g., `--style scss`), the prompt for that option will not be displayed.

### DIFF
--- a/packages/schematics/angular/ng-new/schema.json
+++ b/packages/schematics/angular/ng-new/schema.json
@@ -15,7 +15,8 @@
       "$default": {
         "$source": "argv",
         "index": 0
-      }
+      },
+      "x-prompt": "What name would you like to use for the project?"
     },
     "experimentalIvy": {
       "description": "EXPERIMENTAL: Specifies whether to create a new application which uses the Ivy rendering engine.",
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/ng-new/schema.json
```diff
@@ -95,19 +96,33 @@
     "routing": {
       "type": "boolean",
       "description": "Generates a routing module.",
-      "default": false
+      "default": false,
+      "x-prompt": "Will the project use Angular routing?"
```
StephenFluin: What tense & tone should be used for prompts?

It feels like this question should be more about what WE will do, like 'Would you like to generate a routing module?' or 'Should we add routing to the project?'

packages/schematics/angular/ng-new/schema.json
```diff
@@ -95,19 +96,33 @@
...
     "style": {
       "description": "The file extension to be used for style files.",
       "type": "string",
-      "default": "css"
+      "default": "css",
+      "x-prompt": {
+        "message": "Select preferred style format for the project",
+        "type": "list",
+        "items": [
+          { "value": "css", "label": "CSS" },
```
StephenFluin: I think the Style options should be ordered by popularity, so:

1. css
2. scss
3. sass
4. less
5. stylus

packages/schematics/angular/ng-new/schema.json
```diff
@@ -95,19 +96,33 @@
...
     },
     "prefix": {
       "type": "string",
       "format": "html-selector",
       "description": "The prefix to apply to generated selectors.",
+      "minLength": 1,
       "default": "app",
-      "alias": "p"
+      "alias": "p",
+      "x-prompt": "Choose an HTML selector prefix for the project:"
```
StephenFluin: I don't think we should prompt for a selector. I think `app-` is fine for most folks. Super open to other ideas on this one.

packages/schematics/angular/ng-new/schema.json
clydin: Updated.

packages/schematics/angular/ng-new/schema.json
clydin: Done.

packages/schematics/angular/ng-new/schema.json
clydin: Removed.  Keeping it simple for now makes sense.  We can also add more later.

packages/schematics/angular/ng-new/schema.json
```diff
@@ -95,19 +96,32 @@
     "routing": {
       "type": "boolean",
       "description": "Generates a routing module.",
-      "default": false
+      "default": false,
+      "x-prompt": "Would you like to generate a routing module?"
```
alexeagle: what happens if the user doesn't know what a "routing module" is?

PR #11996, opened 2018-08-24
feat(@angular-devkit/build-angular): use terser instead of uglify-es

### LINKED ISSUE #9340: --prod does not work in Firefox ESR 45.9.0
Hi,

running ´ng serve --prod´ and opening the application in Firefox ESR 45.9.0 throws an exception.

> TypeError: this is undefined

### Versions
```
ng: 1.6.5
npm: 5.6.0
Node: 8.9.4
```

### Repro steps

* ng new foobar
* cd foobar
* ng serve --prod
* Open in Firefox ESR 45.9.0

### Observed behavior

Blank page, error in console

### Desired behavior

Everything works as expected

BR
Jakob

filipesilva:
Should help bring down prod build times, @SanderElias reported a ~60% reduction on his project.

Should fix #9340.

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -46,7 +46,7 @@
     "stylus": "^0.54.5",
     "stylus-loader": "^3.0.2",
     "tree-kill": "^1.2.0",
-    "uglifyjs-webpack-plugin": "^1.2.5",
+    "terser-webpack-plugin": "^1.0.2",
     "url-loader": "^1.0.1",
     "webpack": "^4.15.1",
     "webpack-dev-middleware": "^3.1.3",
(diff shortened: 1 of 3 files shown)


filipesilva:
Thanks for the heads up @kzc, didn't remember that. Fix up at https://github.com/angular/angular-cli/pull/12033

filipesilva:
@kubbafox this is now released in `@angular-devkit/build-angular@0.9.0-beta.2`.

kqadem:
@filipesilva Is this an ng 7 only feature or can we expect to make use of it while remaining on ng 6.x? For. e. g. by just upgrading cli and/or the devkit packages?

kqadem:
@filipesilva Wow, thx for the quick reply! Alright, let's see how it comes. Having a quite big enterprise product here going to be launched soon prohitibing us major version changes.

Since the change only involves switching a package, we might even consider kind of dirty mapping from uglify to terser... I mean, **~60% build time reduction** :scream:

filipesilva:
Well that was for a specific project. I can't say that'll be the case for every project. Have you tried the beta on your app?

kqadem:
I think I'm going to have a look at it by the end of the week.

> Well that was for a specific project. I can't say that'll be the case for every project

yep, could be even more 😜

cquixada:
We had the same problem involving uglify-es. Changing the version from "angular-devkit/build-angular" to "0.9.0-rc.3", even with Angular CLI 6.2.5, our application is running again with older versions of FF without needing "build-optimizer=false". Thanks.

kqadem:
@cquixada did you notice any improvements regarding build time?

PR #12004, opened 2018-08-24
fix(@angular-devkit/schematics): export task executor

qiyigg:


clydin:
Not being imported is intentional.  Those are only relevant to node environments.

qiyigg:
@clydin My understanding is: Those are executors, which are necessary for external workflow. Especially, BuiltinTaskExecutor.RunSchematic should be useful for all workflows. If you'd prefer access the executors through tasks/node/index.ts, I'd suggest reverting the module_root change here(https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/schematics/BUILD#L74)

qiyigg:
This was done by #12008
Close this PR right now.

PR #12014, opened 2018-08-27
feat(@schematics/angular): add default budget of 2mb/5mb

StephenFluin:


packages/schematics/angular/application/index.ts
```diff
@@ -169,6 +169,11 @@ function addAppToWorkspaceFile(options: ApplicationOptions, workspace: Workspace
             extractLicenses: true,
             vendorChunk: false,
             buildOptimizer: true,
+            budgets: [{
+              type: 'initial',
+              maximumWarning: '2mb',
+              maximumError: '5mb',
```
IgorMinar: Can you please document inline what are these numbers based on and point to some reference material so that the developers can make good decisions when adjusting these numbers?

StephenFluin: @IgorMinar Because the generated `angular.json` is valid JSON, it doesn't support comments, so I am not aware of any way to explain the meaning of these values inline. What were you imagining?

hansl: It's JSON5, if you really want. Not all editors support JSON5 though.

jenniferfell: Captured to-do item here: https://github.com/angular/angular/issues/26106

hansl:
Approved by Igor and I, e2e is a flake and author is a google employee. I'm going to merge this in.

PR #12015, opened 2018-08-28
refactor(@angular/cli): Fix EVERYTHING.

hansl:
WIP.


packages/angular_devkit/core/src/json/parser.ts
```diff
@@ -860,3 +879,23 @@ export function parseJson(input: string, mode = JsonParseMode.Default): JsonValu
...
+
+/**
+ * Parse a JSON string into its value.  This discards the AST and only returns the value itself.
+ * It also absorbs JSON parsing errors and return a new error with the path in it. Useful for
+ * showing errors when parsing from a file.
+ * @param input The string to parse.
+ * @param mode The mode to parse the input with. {@see JsonParseMode}.
+ * @returns {JsonValue} The value represented by the JSON string.
+ */
+export function parseJsonFile(input: string, mode = JsonParseMode.Default, path: string) {
```
clydin: based on the name of the function, I'd expect `input` to be a filename and the function to load and parse an actual file.  Can `parseJson` include a `path`/`sourceFile` option (bike shedding the name, sorry)?

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -223,29 +342,33 @@ export class CoreSchemaRegistry implements SchemaRegistry {
 
     return validator
       .pipe(
-        map<ajv.ValidateFunction, SchemaValidator>(validate => (data, options) => {
-          const validationOptions: SchemaValidatorOptions = {
-            withPrompts: true,
-            ...options,
-          };
```
clydin: This was used to make defaults more obvious and eliminated the need to check for undefined everywhere the option was used if the default was true.  It was the reason Partial<> was used in the function declaration as well.

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -175,39 +195,138 @@ export class CoreSchemaRegistry implements SchemaRegistry {
...
-    if (existingValidator) {
-      return observableOf(existingValidator);
+  /**
+   * Compile and return a validation function for the Schema.
+   *
+   * @param {JsonObject | string} schema The schema to validate. If a string, will fetch the schema
+   *     before compiling it (using schema as a URI).
+   * @returns {Observable<SchemaValidator>} An Observable of the Validation function.
+   */
+  compile(schema: JsonObject | string): Observable<SchemaValidator> {
```
clydin: Is the string variant used anywhere?  Also, I think that by looking at the signature, I would assume that the string would be the actual JSON content rather than a URI.

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -175,39 +195,138 @@ export class CoreSchemaRegistry implements SchemaRegistry {
...
     }
 
-    // tslint:disable-next-line:no-any
-    const id = (validate.schema as any).$id || (validate.schema as any).id;
-    let fullReference = (ref[0] === '#' && id) ? id + ref : ref;
+    let refMap = validate as AjvRefMap;
+    const rootRefMap = validate.root as AjvRefMap;
+
+    // Resolve from the root if it's different.
+    if (validate.root && validate.schema !== rootRefMap.schema) {
```
clydin: Why always use the root?  Would that break relative refs within subschemas?

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -103,30 +112,40 @@ export class CoreSchemaRegistry implements SchemaRegistry {
...
     });
 
     this._ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-04.json'));
     this._ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-06.json'));
-
-    this.addPostTransform(addUndefinedDefaults);
   }
 
-  private _fetch(uri: string): Promise<JsonObject> {
+  private _fetch(uri: string): Observable<JsonObject> {
```
clydin: using a promise seems to make more sense here since the observable is being created to then be turned immediately into a promise.

Actually and more generally, as an experiment, I converted the compile/validate method to use async/await instead of observables.  It's shorter, easier to read, and actually debuggable.  With the public API remaining observable-based and consuming code needing to assume that the validating function is asynchronous anyway, I'm not sure there is a compelling reason to use observables internally here.

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -339,14 +469,16 @@ export class CoreSchemaRegistry implements SchemaRegistry {
         valid: true,
         compile: (schema, _parentSchema, it) => {
           const compilationSchemInfo = this._currentCompilationSchemaInfo;
-          if (!compilationSchemInfo) {
+          if (compilationSchemInfo === undefined) {
             throw new Error('Invalid JSON schema compilation state');
+          } else if (compilationSchemInfo === null) {
+            return () => true;
```
clydin: You can collapse these into the second conditional block since not having the info object is now valid in a non-compilation scenario.

packages/angular_devkit/core/src/json/parser.ts
hansl: I was thinking about that. I'll add it, it's no big deal.

packages/angular_devkit/core/src/json/schema/registry.ts
hansl: Removing that code did not change anything else. You're checking that `withPrompt === false` to shortcut the code.

packages/angular_devkit/core/src/json/schema/registry.ts
hansl: No, I'll remove.

clydin: I can see the concept being useful, however.  Maybe a separate method?

packages/angular_devkit/core/src/json/schema/registry.ts
hansl: Doesn't look like it. We only use the root if it's available. The root AFAIK will _always_ have the new refs as they're resolved (in async) before calling this method, and added to the root.

packages/angular_devkit/core/src/json/schema/registry.ts
clydin: Reason I mentioned it is that there are two code blocks similar to the following farther down:
```
if (options === undefined
	      || options.applyPreTransforms === undefined
              || options.applyPreTransforms) {
```

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -457,10 +457,8 @@ export class CoreSchemaRegistry implements SchemaRegistry {
         valid: true,
         compile: (schema, _parentSchema, it) => {
           const compilationSchemInfo = this._currentCompilationSchemaInfo;
-          if (compilationSchemInfo === undefined) {
+          if (compilationSchemInfo === undefined || compilationSchemInfo === null) {
             throw new Error('Invalid JSON schema compilation state');
```
clydin: `return () => true;` here instead of the exception and the null potential can be removed completely.

packages/angular_devkit/core/src/json/schema/registry.ts
clydin: Works for me then.

packages/angular_devkit/core/src/json/schema/registry.ts
hansl: Doesn't work (see e2e errors). I thought I looked at that code path already.

packages/angular_devkit/core/src/json/schema/registry.ts
hansl: Done.

packages/angular_devkit/core/src/json/schema/registry.ts
hansl: It is, but Observables are cancellable and have other advantages. I say that even for our private stuff we should use Promises carefully.

I changed back `_fetch()`; I was using it in flatten at some point but now I just use ajv's logic.

packages/angular_devkit/core/src/json/schema/registry.ts
hansl: Out of scope of this PR, but yes.

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -331,15 +331,18 @@ export class CoreSchemaRegistry implements SchemaRegistry {
     return validator
       .pipe(
         map(validate => (data: JsonValue, options?: SchemaValidatorOptions) => {
-          const validationOptions: SchemaValidatorOptions = options || {};
+          const validationOptions: SchemaValidatorOptions = {
+            withPrompts: true,
+            applyPostTransforms: true,
+            applyPreTransforms: true,
+            ...(options || {}),
```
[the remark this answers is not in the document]
hansl: Done.

hansl:
Rebased and repushed. Let's see CI on this run.

alexeagle:
In the future could you make your commit descriptions include the motivation? In particular when I see
```
feat(@angular/cli): remove yargs-parser and implement parsing

This is fully backward compatible.
```
I am wondering why? What's the context and why is it required for adding the `--help-json` feature?

PR #12022, opened 2018-08-28
Add benchmark package and AIO benchmark

filipesilva:


filipesilva:
Somewhat softblocked on https://github.com/soyuka/pidusage/issues/80, since I can't get the correct elapsed times on docker. But I can see them correct locally.

.monorepo.json
```diff
@@ -87,6 +87,11 @@
       "hash": "9d2161b7ca9044c2286c7d66e10ead12",
       "snapshotRepo": "angular/angular-devkit-architect-cli-builds"
     },
+    "@angular-devkit/benchmark": {
```
alexeagle: Is it a necessary part of this PR to introduce (and publish) a new npm package?

benchmark/aio/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "aio-benchmark",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "initialize": "yarn clone && yarn setup && yarn update", 
+    "clone": "(git clone https://github.com/angular/angular || true) && cd angular && git fetch origin && git reset --hard dd2a650c3455f3bc0a88f8181758a84aacb25fea",
+    "setup": "cd angular && yarn && cd aio && yarn && yarn setup",
+    "//": "Shouldn't need to install the package twice, but the first install seems to leave two @ngtools/webpack installs around.",
```
alexeagle: ouch... can (should) we fix it now?

benchmark/aio/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "update": "cd angular/aio && yarn add ../../../../dist/_angular-devkit_build-angular.tgz --dev && yarn add ../../../../dist/_angular-devkit_build-angular.tgz --dev",
+    "benchmark": "cd angular/aio && benchmark --verbose -- yarn ~~build --configuration=stable"
```
alexeagle: what is this `~~` thing?

packages/angular_devkit/benchmark/BUILD
```diff
@@ -0,0 +1,57 @@
...
+            "src/**/*_benchmark.ts",
+        ],
+    ),
+    module_name = "@angular-devkit/benchmark",
+    module_root = "src/index.d.ts",
+    deps = [
+        "//packages/angular_devkit/core",
+        "@rxjs",
+        "@rxjs//operators",
+        # @typings: node
```
alexeagle: @gregmagolan ideally with the changes you're working on, this should become a real load-bearing dependency so it matches google3

benchmark/aio/package.json
filipesilva: Spent a lot of time trying to work around yarn behaviour when installing local packages and it seems broken in a lot of ways. I've seen double packages being left around, mysterious 'failure to link' and others. Can't say I'm very inclined to continue mucking about in it right now.

.monorepo.json
filipesilva: It was part of the original design, so that the benchmark could be ran over other projects.

benchmark/aio/package.json
filipesilva: It's a script in AIO:
```
"~~build": "ng build",
```
It's one of the only ones that doesn't have pre and post scripts that would add noise to the benchmark.

packages/angular_devkit/benchmark/bin/benchmark.ts
```diff
@@ -0,0 +1,140 @@
...
+        --cwd                     Current working directory to run the process in.
+        --output-file             File to output benchmark log to.
+        --overwrite-output-file   If the output file should be overwritten rather than appended to.
+        --prefix                  Logging prefix.
+
+    Example:
+        benchmark --iterations=3 -- node my-script.js
+  `);
+
+  process.exit(exitCode);
```
alexeagle: I want tests and users to be able to call the main. That lets me call the program the same way it is from the command-line, not missing any bits like arg handling when I call it via API call.

Could you end the script with 
```
if (require.main === module) {
  process.exitCode = main(process.argv.slice(2));
}
```
and remove all the process.exit() calls? Those make it impossible to re-use this in another context

Also this demonstrates that there is no integration test that calls the top-level of the program

packages/angular_devkit/benchmark/bin/benchmark.ts
```diff
@@ -0,0 +1,140 @@
...
+  // The node typing sometimes don't have a never type for process.exit().
```
alexeagle: did you investigate why? can we just fix our `@types/node` version?

packages/angular_devkit/benchmark/src/command.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+export class Command {
```
alexeagle: strikes me as overengineered but I'm pretty novice at this codebase and our design patterns...

packages/angular_devkit/benchmark/src/default-reporter.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { logging, tags } from '@angular-devkit/core';
+import { AggregatedMetric, BenchmarkReporter, Metric } from './interfaces';
+
+export const defaultReporter = (logger: logging.Logger): BenchmarkReporter => (process, groups) => {
```
alexeagle: google3 style guide always prefers
export function f() {}
over
export const f = () => {};

maybe we should do the same?

packages/angular_devkit/benchmark/src/default-stats-capture.ts
```diff
@@ -0,0 +1,60 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Observable } from 'rxjs';
+import { map, reduce } from 'rxjs/operators';
+import { AggregatedProcessStats, Capture, MetricGroup, MonitoredProcess } from './interfaces';
+import { comulativeMovingAverage, max } from './utils';
+
+
+export const defaultStatsCapture: Capture = (
```
alexeagle: ditto, why not the `function` keyword

packages/angular_devkit/benchmark/src/default-stats-capture.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Observable } from 'rxjs';
+import { map, reduce } from 'rxjs/operators';
+import { AggregatedProcessStats, Capture, MetricGroup, MonitoredProcess } from './interfaces';
+import { comulativeMovingAverage, max } from './utils';
```
alexeagle: spelling: cumulative

.monorepo.json
clydin: I'm not sure we should publish it yet though.  There's already a large amount of packages under management.

packages/angular_devkit/benchmark/src/interfaces.ts
```diff
@@ -0,0 +1,49 @@
+/**
```
alexeagle: nit: should this be a `.d.ts` file instead to enforce that it has no values?

packages/angular_devkit/benchmark/src/monitored-process.ts
```diff
@@ -0,0 +1,106 @@
...
+      const { cmd, cwd, args } = this.command;
+      const spawnOptions: SpawnOptions = { cwd: cwd, shell: true };
+
+      // Spawn the process.
+      const childProcess = spawn(cmd, args, spawnOptions);
+
+      // Emit output and stats.
+      childProcess.stdout.on('data', (data: Buffer) => this.stdout.next(data));
+      childProcess.stderr.on('data', (data: Buffer) => this.stderr.next(data));
+      const statsSubs = timer(0, this.pollingRate).pipe(
```
alexeagle: wow all the rxjs usage here feels un-idiomatic for a command-line tool but I'm learning :)

packages/angular_devkit/benchmark/src/monitored-process.ts
```diff
@@ -0,0 +1,106 @@
...
+          for (const pid of pids) {
+            cpu += statsByProcess[pid].cpu;
+            memory += statsByProcess[pid].memory;
+          }
+
+          return {
+            processes, cpu, memory, pid, ppid, ctime, elapsed, timestamp,
+          } as AggregatedProcessStats;
+        }),
+        tap(stats => this.stats.next(stats)),
```
alexeagle: makes me think: any idea how to edit a file in between runs so that we can measure the incremental dev RTT like the ibazel one does?

packages/angular_devkit/benchmark/src/monitored-process_spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it('captures stats', (done) => {
+    const process = new LocalMonitoredProcess(cmd);
+    const statsOutput: string[] = [];
+    process.stderr$.subscribe(data => statsOutput.push(data.toString()));
+
+    process.run().pipe(
+      tap(() => expect(statsOutput.length).toBeGreaterThan(0)),
+    ).toPromise().then(done, done.fail);
```
alexeagle: curious, would it be more readable with `async`/`await` ?

.monorepo.json
filipesilva: How about we build it, but exclude from the publishing process?

clydin: sounds good. is there a facility to do that?

filipesilva: The publish script honors the `"private": true,` field in package.json and does not attempt to publish those packages. I added that to the benchmark package.

.circleci/config.yml
```diff
@@ -93,6 +96,21 @@ jobs:
       - attach_workspace: *attach_options
       - run: npm run admin -- build
 
+  benchmark:
```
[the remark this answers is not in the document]
filipesilva: I actually thought this was part of the TOOL-100 (import an example app into our repo) subtask but looking at it again that's not the case. The task only specifies importing, not running the benchmark over the app.

The idea was to be able to compare the results vs master when doing PRs that should affect performance. 

If the CI job remains it's probably worth it to add the benchmark log as an CircleCI artifact to be easier to consult like `angular-bazel-example` does with the bundles (https://circleci.com/gh/gregmagolan/angular-bazel-example/653#artifacts/containers/0).

benchmark/aio/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "aio-benchmark",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "initialize": "yarn clone && yarn setup && yarn update", 
+    "clone": "(git clone https://github.com/angular/angular || true) && cd angular && git fetch origin && git reset --hard dd2a650c3455f3bc0a88f8181758a84aacb25fea",
```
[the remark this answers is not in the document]
filipesilva: Replacing with 
```
    "clone": "(git clone https://github.com/angular/angular --depth 1 || true) && cd angular && git fetch origin dd2a650c3455f3bc0a88f8181758a84aacb25fea && git reset --hard dd2a650c3455f3bc0a88f8181758a84aacb25fea",
```

.circleci/config.yml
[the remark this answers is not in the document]
filipesilva: Discussed offline and for now we will remove the CI job and run it manually.

.monorepo.json
```diff
@@ -87,6 +87,11 @@
...
+      "name": "Benchmark",
+      "section": "Tooling",
+      "version": "0.0.0"
```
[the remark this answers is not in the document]
filipesilva: That is correct, starting at 0.0.x iirc.

packages/angular_devkit/benchmark/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+NB: the `fibonacci.js` in the example is a naive implementation of a fibonacci number calculator:
+```
+// fibonacci.js
+const fib = (n) => n > 1 ? fib(n - 1) + fib(n - 2) : n;
+console.log(fib(parseInt(process.argv[2])));
+```
+
+## API Usage
```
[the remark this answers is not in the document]
filipesilva: I added this section after @hansl asked if it was possible to output in JSON. Using it like a library enables this case (and others).

I think what'll happens is that users that want to upload the results will use it as a library instead of calling the binary and parsing the file/stdout.

I think it's worthwhile to leave the usage listed in the readme as documentation for our own team. Then adding a warning/disclaimer below about stability and and keeping the package private should be enough for making our commitment explicit.

packages/angular_devkit/benchmark/bin/benchmark.ts
filipesilva: I did not investigate, this part was copy pasted from our other CLI tools. 

ATM I've removed this section on account of removing all the process.exit.

packages/angular_devkit/benchmark/bin/benchmark.ts
filipesilva: Wrapped binary code into a exported main function, calling it only if this is the main for the node program.

packages/angular_devkit/benchmark/src/command.ts
filipesilva: Simplified by removing the getters. They were there so the command did not change midway but realistically that's not going to happen.

packages/angular_devkit/benchmark/src/interfaces.ts
filipesilva: Done

packages/angular_devkit/benchmark/src/monitored-process.ts
filipesilva: I had some ideas, like a triggering capture. It's in the benchmark doc, but cut out of this version. It's the next thing that should be worked on imho.

packages/angular_devkit/benchmark/src/monitored-process_spec.ts
filipesilva: It goes from
```
  it('works', (done) => {
    const process = new LocalMonitoredProcess(cmd);
    process.run().pipe(
      tap(code => expect(code).toBe(0)),
  });
```

to

```
  it('works', async () => {
    const process = new LocalMonitoredProcess(cmd);
    const res = await process.run().pipe(toArray()).toPromise();
    expect(res).toEqual([0]);
  });
```

I think it looks nicer and also ensures the number of emissions.

packages/angular_devkit/benchmark/src/monitored-process.ts
filipesilva: Yeah the three output observables here (`stats$`, `stdout$` and `stderr$`) are not very idiomatic. I didn't quite know how else I could accomplish what I wanted though.

The idea was to allow captures to be setup before the process ran to ensure no output was missed. So the observables had exist when instantiating a `LocalMonitoredProcess`, and output was sent there as it came from the different sources.

I couldn't find a better way to do it. Do you have any suggestions?

packages/angular_devkit/benchmark/src/default-reporter.ts
filipesilva: I'm happy doing that given we actually enforce it, e.g. via linting. 

Personally I prefer `export const f = () => {};` myself, for instance. Others might too. Without the style being enforced, it's up to the reviewer whether to request it to change.

I'd be reluctant to ask request that change in a PR otherwise ready to go, and eventually a couple of these uses would go in.

packages/angular_devkit/benchmark/README.md
```diff
@@ -0,0 +1,113 @@
...
+It currently shows only time, process, cpu, and memory used but might be extended in the future.
+
+This tool was created to provide an objective and reproducible way of benchmarking process
+performance.
+
+Given a process (or its source code), process inputs and environment, keeping two of these elements
+constant while varying the third should allow for meaningful benchmarks over time.
+
+In the context of the DevKit, we publish many CLI tools and have access to their source code.
+By tracking tool resource usage we can catch performance regressions or improvements on our CI.
```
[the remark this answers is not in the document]
filipesilva: Done

packages/angular_devkit/benchmark/README.md
```diff
@@ -0,0 +1,113 @@
...
+    --output-file             File to output benchmark log to.
+    --overwrite-output-file   If the output file should be overwritten rather than appended to.
+    --prefix                  Logging prefix.
+
+Example:
+    benchmark --iterations=3 -- node my-script.js
+```
+
+
+NB: the `fibonacci.js` in the example is a naive implementation of a fibonacci number calculator:
```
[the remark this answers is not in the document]
filipesilva: Done

packages/angular_devkit/benchmark/README.md
```diff
@@ -0,0 +1,113 @@
+# Angular Devkit Benchmark
+
+This tools provides benchmark information for processes.
```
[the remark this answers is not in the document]
filipesilva: Done

packages/angular_devkit/benchmark/src/default-stats-capture.ts
filipesilva: Done

benchmark/aio/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "name": "aio-benchmark",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "initialize": "yarn clone && yarn setup && yarn update", 
+    "clone": "(git clone https://github.com/angular/angular --depth 1 || true) && cd angular && git fetch origin dd2a650c3455f3bc0a88f8181758a84aacb25fea && git checkout -f FETCH_HEAD",
+    "setup": "cd angular && yarn && cd aio && yarn && yarn setup",
+    "//": "Shouldn't need to install the package twice, but the first install seems to leave two @ngtools/webpack installs around.",
+    "update": "cd angular/aio && yarn add ../../../../dist/_angular-devkit_build-angular.tgz --dev && yarn add ../../../../dist/_angular-devkit_build-angular.tgz --dev",
```
alexeagle: nit: could move the workaround into a `postupdate` script

package.json
```diff
@@ -42,7 +43,6 @@
     "url": "https://github.com/angular/angular-cli.git"
   },
   "engines": {
-    "node": ">=10.3.0 <11.0.0",
```
alexeagle: why removed? bad merge?

packages/angular_devkit/benchmark/src/default-reporter.ts
alexeagle: I agree we should configure the linter rather than make these decisions one-off

packages/angular_devkit/benchmark/src/main.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  // Print help.
+  if (argv['help']) {
+    usage(logger);
+
+    return 0;
+  }
+
+  // Exit early if we can't find the command to benchmark.
+  if (!Array.isArray(argv['--']) || (argv['--'] as Array<string>).length < 1) {
```
alexeagle: nit: I'm surprised you need to cast `['--']` as a `string[]`

packages/angular_devkit/benchmark/src/main_spec.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { existsSync, readFileSync, unlinkSync, writeFileSync } from 'fs';
+import { basename, dirname, join } from 'path';
+import { main } from './main';
+// tslint:disable-next-line:no-implicit-dependencies
```
alexeagle: strange I thought this was `no-require-imports`

packages/angular_devkit/benchmark/src/monitored-process.ts
alexeagle: I was thinking at a higher level, that since our code doesn't need to respond to externally-generated events, we don't really need to be "reactive" in managing state in the app, and could use a more procedural style of programming without any rxjs. It's a broader style discussion for the whole repo.

package.json
filipesilva: Not sure how it got there. Glad you caught it!

packages/angular_devkit/benchmark/src/main.ts
filipesilva: It is not `['--']` that is being cast but rather `argv['--']`. Depending on the content of `process.argv`, that array might not exist.

Looking back on this logic I think I'm casting it way more than needed. Will make an intermediate var that's properly typed.

packages/angular_devkit/benchmark/src/main_spec.ts
filipesilva: Disabling TSLint here was because the `temp` dependency is not declared in the dependencies or devdependencies of this package. Rather, it's a top level dependency of the monorepo that we use for tests, and in other tests where it is used (`/core/node/host_spec.ts`) this is how we import it. Maybe we should make these explicit in the devDependencies?

benchmark/aio/package.json
filipesilva: Yeah that makes the workaround nature more obvious.

benchmark/aio/package.json
```diff
@@ -0,0 +1,18 @@
...
+    "clone": "(git clone https://github.com/angular/angular --depth 1 || true) && cd angular && git fetch origin dd2a650c3455f3bc0a88f8181758a84aacb25fea && git checkout -f FETCH_HEAD",
+    "setup": "cd angular && yarn && cd aio && yarn && yarn setup",
+    "update": "cd angular/aio && yarn add ../../../../dist/_angular-devkit_build-angular.tgz --dev",
+    "//": "Shouldn't need to install the package twice, but the first install seems to leave two @ngtools/webpack installs around.",
+    "postupdate": "cd angular/aio && yarn add ../../../../dist/_angular-devkit_build-angular.tgz --dev",
+    "benchmark": "cd angular/aio && benchmark --verbose -- yarn ~~build --configuration=stable"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC"
```
clydin: Is this right?  probably should just remove as it is not an actual package.

PR #12023, opened 2018-08-28
style(@schematics/angular): fix typo and English in environment.ts

bedag-moo:


bedag-moo:
commit amended to remove the offending space character

PR #12027, opened 2018-08-28
feat(@angular-devkit/build-angular): remove inlining of assets in css

alan-agius4:
BREAKING CHANGE: Assets under 10Kib are not longer inlined in css

superseeds https://github.com/angular/angular-cli/pull/11826


alan-agius4:
@filipesilva, I think there is but I was gonna do a separate PR to keep the scope of the PR small, What do you reckon?

clydin:
postcss-url can be removed completely with some additional refactoring but I agree a separate PR would be appropriate for that.

PR #12037, opened 2018-08-29
refactor(@angular-devkit/build-angular): clean up style webpack config

alan-agius4:
Clean up several parts of the styles config and also removed the dependency on `postcss-url`

//cc @filipesilva & @clydin

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/postcss-cli-resources.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/postcss-cli-resources.ts
@@ -1,7 +1,3 @@
-// tslint:disable
-// TODO: cleanup this file, it's copied as is from Angular CLI.
-
-
 /**
  * @license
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/plugins/postcss-cli-resources.ts
```diff
@@ -64,11 +72,42 @@ export default postcss.plugin('postcss-cli-resources', (options: PostcssCliResou
       return cachedUrl;
     }
 
+    if (inputUrl.startsWith('~')) {
```
clydin: This block should go.  It was meant to bridge the gap for 6.x by providing rudimentary support.  It doesn't actual work properly either.  Instead just remove the tilde and continue processing.

alan-agius4: Updated.

PR #12038, opened 2018-08-29
feat(@angular-devkit/build-angular): update webpack dependencies

clydin:


yarn.lock
```diff
@@ -7778,13 +7878,20 @@ webpack-sources@^1.0.1, webpack-sources@^1.1.0:
     source-list-map "^2.0.0"
     source-map "~0.6.1"
 
+webpack-sources@^1.2.0:
```
alan-agius4: Nice, this should improve the CI build as well. As people are claiming good improvements on webpack when having sourcemaps enabled.

filipesilva: Would be cool to benchmark it the before and after with #12022. Only after the CI error is addressed though.

ralshammrei2:
#12038

clydin:
Covered by #12185

PR #12045, opened 2018-08-30
docs: fix proxy config paths

NiklasTeich:
Fixed serve proxy.conf.json path. The documentation wants us to create the config file next to the package.json (project root), not the src subdirectory. So, the proxyConfig needs that path.


alan-agius4:
Can you please rebase and fix your commit message? It should be something along 

`docs: fix proxy config paths`

alan-agius4:
Superseded  by 
https://github.com/angular/angular-cli/pull/12071

PR #12046, opened 2018-08-30
build: update repo to use `typescript` `3.0.x`

alan-agius4:
Schematics still create the default project with typescript 2.9 and angular 6

superseeds https://github.com/angular/angular-cli/pull/11889


packages/angular_devkit/build_angular/test/browser/build-optimizer_spec_large.ts
```diff
@@ -21,7 +21,7 @@ describe('Browser Builder build optimizer', () => {
 
   it('works', (done) => {
     const overrides = { aot: true, buildOptimizer: true };
-    runTargetSpec(host, browserTargetSpec, overrides).pipe(
+    runTargetSpec(host, browserTargetSpec, overrides, DefaultTimeout * 2).pipe(
```
alan-agius4: With TypeScript 3, build optimiser is taking more time to complete. For this spec takes `28 secs` in TS2.9 while `36 secs` in TS3 (locally)

packages/ngtools/webpack/package.json
```diff
@@ -31,11 +31,10 @@
     "webpack-sources": "^1.1.0"
   },
   "peerDependencies": {
-    "typescript": "~2.4.0 || ~2.5.0 || ~2.6.0 || ~2.7.0 || ~2.8.0 || ~2.9.0",
+    "typescript": ">=2.4.0 < 3.1",
     "webpack": "^4.0.0"
   },
   "devDependencies": {
-    "typescript": "~2.9.2",
```
filipesilva: It's still a dev dependency of this package I think. Removing it works because we still have the toplevel one, but keep with workspaces we should have all the dev deps here. That becomes relevant if we have stricter tslint import checks for example.

alan-agius4: updated

PR #12048, opened 2018-08-30
fix: limit rxjs to 6.2.x due to a defect in 6.3.0

clydin:


package.json
```diff
@@ -107,6 +107,6 @@
   "resolutions": {
     "@types/webpack": "4.4.0",
     "@types/webpack-dev-server": "2.9.4",
-    "rxjs": "6.0.0"
+    "rxjs": "~6.2.0"
```
hansl: If that's the same as the main one, then you can remove this line IIRC.

PR #12060, opened 2018-08-31
Revert "build: limit rxjs to 6.2.x due to a defect in 6.3.0 (#12048)"

alan-agius4:
This reverts commit 7e63dd791b67ad49de4b0b2f55ecc386e4cad20e.

Verifying that `rxjs 6.1.3` doesn't throw an error

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "license-checker": "^20.1.0",
     "minimatch": "^3.0.4",
     "minimist": "^1.2.0",
-    "rxjs": "~6.2.0",
+    "rxjs": "^6.0.0",
     "semver": "^5.3.0",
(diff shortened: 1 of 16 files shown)


alan-agius4:
Blocked by: https://github.com/ReactiveX/rxjs/pull/4078

skonx:
Hi. I've just updated rxjs from 6.2.2 to 6.3.1 and got the error `TypeError: sink._addParentTeardownLogic is not a function` running a `ng serve`. 

Is this linked with this ticket?

clydin:
We should also set the minimum version to at least 6.2 as well.  This brings in the exception handling fix that forced workaround code throughout the codebase.

PR #12068, opened 2018-09-03
docs: convert console image in microsite to css

Ansimorph:
As a user I'd like to copy & paste the commands to install the ng cli directly from cli.angular.io. Because the console displayed on the page is an image this is not possible. I witnessed the frustration this generates first hand several times while guiding new users through the install process.

To fix that I recreated the console image in pure CSS. The commands can now be copied and the console responsively scales for very small screens. Also this approach is more performant, since the original image is bigger than the extra css + the monospace font.

I tried to match the browser support of Material Design Lite in my CSS, which is why I avoided flexbox. All CSS is minified with CSSO.


hansl:
I published this manually.

PR #12071, opened 2018-09-03
docs(stories): update proxy.md with correct path

jkuri:
Since it is stated;
> We create a file next to our project's package.json called proxy.conf.json with the content

Then paths to `proxy.conf.json` in `angular.json` are not prefixed with `src/`.


alan-agius4:
Can you please fix the commit message, It should be `docs: update proxy.md with correct path`?

jkuri:
@alan-agius4 done.

PR #12172, opened 2018-09-05
fix(main): match styleguide catch callback

sgwatgit:
Referencing the original PR here https://github.com/angular/devkit/pull/422

https://angular.io/guide/styleguide#style-02-05

In the styleguide rule above the catch callback uses console.error when the app boostrapping fails. I am updating the template to match.

I wasn't sure if the success console.log should also be included but that seems a bit too opinionated.


alan-agius4:
Can you please fix your commit message to `style: match styleguide catch callback`

sgwatgit:
All fixed :)

PR #12175, opened 2018-09-05
Update Build Optimizer for Angular 7

filipesilva:


packages/angular_devkit/build_optimizer/src/transforms/wrap-enums_spec.ts
```diff
@@ -79,6 +80,61 @@ describe('wrap-enums', () => {
     expect(tags.oneLine`${transform(input)}`).toEqual(tags.oneLine`${output}`);
   });
 
+  it('wraps ts 2.3 - 2.6 enums in IIFE, even if they were renamed due to scope hoisting', () => {
```
IgorMinar: isn't this "2.3 - 3.0" ?

packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
```diff
@@ -427,7 +428,21 @@ function pickDecorateNodesToRemove(
...
+    const paramCalls = elements.filter((el) => {
+      if (!isTslibHelper(el, '__param', tslibImports, checker)) {
+        return false;
+      }
+      if (el.arguments.length != 2) {
+        return false;
+      }
+      if (el.arguments[0].kind !== ts.SyntaxKind.NumericLiteral) {
+        return false;
+      }
```
IgorMinar: why do you think it's important to have the length and arguments[0] check? do you think we could misidentify the __params helper somehow?

filipesilva: Throughout Build Optimizer we try to keep these checks tight. Sometimes all we can say is that something has the right identifier text, and there are some situations where it can go wrong. 

Since Build Optimizer will end up processing a lot of code we try to err on the side of caution. Here the best we can say is "it's a call expression called __param with two arguments, where the first one one is a number". 

I think it's very unlikely to reach this section of the code if we're not in the happy path, but if we remove something that we shouldn't then it's very likely the app breaks.

packages/angular_devkit/build_optimizer/src/transforms/wrap-enums_spec.ts
filipesilva: Yes, it is now. I'll edit it.

packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
IgorMinar: sgtm

packages/angular_devkit/build_optimizer/src/transforms/wrap-enums_spec.ts
```diff
@@ -14,6 +14,7 @@ import { getWrapEnumsTransformer } from './wrap-enums';
 const transform = (content: string) => transformJavascript(
   { content, getTransforms: [getWrapEnumsTransformer] }).content;
 
+// tslint:disable-next-line:no-big-function
```
alexeagle: note, we have some effort from google3 to teach tslint how to have relaxed constraints in tests, would be worth pushing that forward and using it...

PR #12176, opened 2018-09-05
ci: update CI to node 10

filipesilva:
Blocked on https://github.com/angular/angular/pull/25812
Related to https://github.com/angular/angular/pull/25822

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,9 @@
 ## IMPORTANT
 # If you change the `docker_image` version, also change the `cache_key` suffix and the version of
 # `com_github_bazelbuild_buildtools` in the `/WORKSPACE` file.
-var_1: &docker_image angular/ngcontainer:0.3.3
-var_2: &cache_key angular_devkit-{{ checksum "yarn.lock" }}-0.3.3-2
+var_1: &docker_image angular/ngcontainer:0.5.0
+var_2: &cache_key angular_devkit-{{ checksum "yarn.lock" }}-0.5.0
+var_3: &node_8_docker_image angular/ngcontainer:0.3.3
 
 # Settings common to each job
 anchor_1: &defaults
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -13,7 +13,9 @@
 # If you change the `docker_image` version, also change the `cache_key` suffix and the version of
 # `com_github_bazelbuild_buildtools` in the `/WORKSPACE` file.
 var_1: &docker_image angular/ngcontainer:0.3.3
-var_2: &cache_key angular_devkit-{{ checksum "yarn.lock" }}-0.3.3-2
+var_2: &node_8_docker_image angular/ngcontainer:0.3.3
+# TODO: CHANGE THIS CACHE KEY
```
filipesilva: Must be changed after ngcontainer is updated.

package.json
```diff
@@ -35,15 +35,15 @@
...
+    "webdriver-update-circleci": "webdriver-manager update --standalone false --gecko false --versions.chrome $CHROMEDRIVER_VERSION_ARG "
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/angular/angular-cli.git"
   },
   "engines": {
-    "node": ">= 8.9.0",
-    "npm": ">= 5.5.1"
+    "node": ">=10.3.0 <11.0.0",
```
clydin: This may confuse people looking at the repo.  Do we require 10 to build/develop?

filipesilva: With this PR, we run all tests on Node 10 and run only test running the CLI on Node 8. So developing on node 8 is a bit a at-your-own-risk kind of thing.

I think yarn will yell at you because it actually bothers to check the engines properly so I'm not very opposed to leaving it at `"node": ">= 8.9.0",` to stop that. But it is misleading.

hansl: This line is propagated to all `package.json` from the build script.

filipesilva: Followup PR on https://github.com/angular/angular-cli/pull/12202.

PR #12178, opened 2018-09-05
fix(@schematics/angular): use TS 3.0+ for experimentalAngularNext

cexbrayat:
Sets the TypeScript version to `~3.0.3` if the option `experimentalAngularNext` (or `experimentalIvy`) is set.

Currently the generated app uses TS ~2.9.2 and throws if the `next` version of Angular is used:

    Error: The Angular Compiler requires TypeScript >=3.0.1 and <3.1.0 but 2.9.2 was found instead.

### DIFF
--- a/packages/schematics/angular/workspace/files/package.json
+++ b/packages/schematics/angular/workspace/files/package.json
@@ -42,6 +42,6 @@
     "protractor": "~5.4.0",
     "ts-node": "~7.0.0",
     "tslint": "~5.11.0",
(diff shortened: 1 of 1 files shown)


alan-agius4:
Note: on master this is superseded by https://github.com/angular/angular-cli/commit/7421d63d4ae51bd3862df55c830d4c00874671ad#diff-e86faf3f22e7942a51391830c102eee9

hansl:
Closing this as it's already in master for everyone (not just experimental).

PR #12185, opened 2018-09-06
feat: pin down all direct dependencies

alan-agius4:
This is so that CLI users get the same direct dependency versions that were tested on the CI.


packages/angular_devkit/build_angular/package.json
```diff
@@ -11,52 +11,52 @@
...
+    "less-loader": "4.1.0",
+    "license-webpack-plugin": "1.4.0",
+    "loader-utils": "1.1.0",
+    "mini-css-extract-plugin": "0.4.2",
+    "minimatch": "3.0.4",
+    "parse5": "4.0.0",
+    "opn": "5.3.0",
+    "portfinder": "1.0.17",
+    "postcss": "7.0.2",
+    "postcss-import": "12.0.0",
```
clydin: Avoid major version bumps for now.  We should analyze them first to see if there are any relevant breaking changes and update them in a separate PR.

packages/schematics/angular/workspace/files/package.json
```diff
@@ -20,28 +20,28 @@
...
+    "core-js": "2.5.7",
     "rxjs": "<%= latestVersions.RxJs %>",
     "zone.js": "<%= latestVersions.ZoneJs %>"
   },
   "devDependencies": {
-    "@angular/cli": "<%= experimentalAngularNext ? 'next' : '~' + version %>",
+    "@angular/cli": "<%= experimentalAngularNext ? 'next' : version %>",
     "@angular/compiler-cli": "<%= experimentalAngularNext ? 'next' : latestVersions.Angular %>",
     "@angular/language-service": "<%= experimentalAngularNext ? 'next' : latestVersions.Angular %>",
-    "@types/jasmine": "~2.8.8",
```
clydin: I think we should leave the schematic templates out of this first round.  We need an update strategy for these as well as possible coordination with core.

packages/angular_devkit/build_angular/package.json
alan-agius4: Updated

PR #12194, opened 2018-09-06
angular/devkit#603 - Fix deprecation warning with Purify Webpack plugin

mtraynham:
@alexeagle, I've ported this from angular/devkit#603

> Update Purify plugin to fix deprecation warning with Webpack 4


clydin:
Superseded by #12171

It was deprecated for 6.x and removed for 7.x.  It is no longer needed with webpack 4 and its built-in module concatenation plugin.

PR #12204, opened 2018-09-07
fix(@angular/cli): fix schema error Property `targets` is not allowed

### LINKED ISSUE #12192: Schema doesn't support 'targets'
```
- [ ] feature request
```

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

6.2.0

### Repro steps

Install angu;ar cli 6.2.0

The schema still only support 'architect' and not the new 'targets' key

### Desired functionality

Support 'targets' so my editor won't complain.

alan-agius4:
Closes #12192

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -327,11 +327,39 @@
           "additionalProperties": {
             "$ref": "#/definitions/project/definitions/target"
           }
+        },
+        "targets": {
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/project/definitions/target"
+          }
         }
       },
       "required": [
         "root",
         "projectType"
+      ],
+      "anyOf": [
+        {
+          "required": ["architect"],
+          "not": {
+            "required": ["targets"]
+          }
+        },
+        {
+          "required": ["targets"],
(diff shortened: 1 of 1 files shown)


packages/angular/cli/lib/config/schema.json
```diff
@@ -322,7 +322,7 @@
             "library"
           ]
         },
-        "architect": {
+        "targets": {
```
clydin: either one (but not both at the same time) should be allowed for 7.x.  Unfortunately, this makes the schema far more complicated.

packages/angular/cli/lib/config/schema.json
```diff
@@ -327,12 +327,40 @@
...
+            "required": ["targets"]
+          }
+        },
+        {
+          "required": ["targets"],
+          "not": {
+            "required": ["architect"]
+          }
+        },
+        {
```
alan-agius4: This is because `targets` and  `architect` are not really required.

PR #12205, opened 2018-09-07
feat: docker init

bbogdanov:
Introducing only the **init** action. 

Tested with a brand new application generated with the latest cli.

Also tested with angular-cli@1.6.5 generated app with the following required actions:
- rm -rf node_modules
- npm link @angular/cli
- ng update --all
- yarn

### DIFF
--- a/docs/documentation/docker.md
+++ b/docs/documentation/docker.md
@@ -1,0 +1,29 @@
+<!-- Links in /docs/documentation should NOT have `.md` at the end, because they end up in our wiki at release. -->
+
+# ng docker
+
+## Overview
+`ng docker <action> [options]` Creates/modifes docker files.
+
+## Available Commands:
+ - [init](docker/init)
+
+## Options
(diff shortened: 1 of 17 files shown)


bbogdanov:
Bot run

bbogdanov:
Any thoughts on this one?

hansl:
Hi @bbogdanov ,

As much as an `ng docker` command would be interesting, I don't think we're ready to maintain it and the code that comes with this PR. I think the Angular project and the Docker container should be separate commands.

May I suggest that you make a schematic to create new projects with a docker already configured, instead? This is definitely something we support.

Thanks. Unfortunately, I'm going to close this as it is not a feature we're interested in.

bbogdanov:
@hansl Based on your answer maybe you also would like to remove https://github.com/angular/angular-cli/blob/master/docs/design/docker-deploy.md

PR #12223, opened 2018-09-10
feat(@schematics/angular): Add the option to create an empty workspace

### LINKED ISSUE #12216: Allow creating a workspace without any projects
This is a feature request

### Desired functionality
When creating a new workspace an application is created automatically, I would like to be able to create a new workspace with no applications inside.

This would be helpful for just creating a library without an application. Also it would permit creating **all** projects inside of the `projects` directory.

### Proposal
Add a new option to the `ng-new` schematic named `empty` which would default to false, which can be used to create the workspace without any projects.

```
ng new my-workspace --empty
# or with an alias
ng new my-workspace -e
```

### Consideration
This logic could be implemented as a custom schematic, but I feel that there is value in providing this in the default schematic collection to reach a broader audience, and at the same time making the default `ng-new` more extensible for other custom schematic collections that are based upon these default schematics.

Brocco:
Fixes #12216


packages/schematics/angular/ng-new/index_spec.ts
```diff
@@ -56,4 +56,24 @@ describe('Ng New Schematic', () => {
...
+  describe('empty option', () => {
+    it('should create an empty workspace', () => {
+      const options = { ...defaultOptions, empty: true };
+
+      const tree = schematicRunner.runSchematic('ng-new', options);
+      const files = tree.files;
+      expect(files.indexOf('/bar/angular.json')).toBeGreaterThanOrEqual(0);
+    });
+
+    it('should not create a project', () => {
```
alexeagle: nit: I think I'd put both assertions in the same test case

packages/schematics/angular/ng-new/schema.d.ts
```diff
@@ -72,4 +72,8 @@ export interface Schema {
    * Skip creating spec files.
    */
   skipTests?: boolean;
+  /**
+   * Create an empty workspace with no projects.
+   */
+  empty?: boolean;
```
alexeagle: just FYI we're working to make this file generated from the `.json` version

packages/schematics/angular/ng-new/schema.json
```diff
@@ -131,6 +131,12 @@
       "type": "boolean",
       "default": false,
       "alias": "S"
+    },
+    "empty": {
```
alexeagle: hmm, what do you think about reversing the logic, so that the options all represent positives? Otherwise it feels like it's added retroactively

- new option would be `createProject` or similar
- it defaults to true
- if you want an empty project you do `ng new --createProject=false` (or do we support `ng new --noCreateProject`?)

packages/schematics/angular/ng-new/index_spec.ts
Brocco: Agreed, updating.

packages/schematics/angular/ng-new/schema.d.ts
Brocco: I think that's a good idea, it was present (although manual) before, but without this the TS compilation will fail.

packages/schematics/angular/ng-new/schema.json
Brocco: I actually disagree with this. I think the default should be negative and the options "turn them on"
As a user the cognitive load of turning on "empty" via `--empty` is easier that the idea of "create project false" via `--noCreateProject".

This is the same approach for most CLI tools I've used.

deleting a directory: `rm -rf` where you're setting recursive and force to true
pushing a branch `git push origin some-branch -f` sets force to true

Taking this to the next level, I think a standardization of the provided boolean options for schematics and architect should be done at some point. It will provide consistency to users about how to use any of the available boolean options.

alexeagle: What I'm saying is, imagine you didn't know what is the current behavior of Angular CLI.

There is a command that creates a new workspace, it can optionally also create a project.

Empty is the natural default state of the new workspace. The option `--createProject` also does a second step of populating the new workspace.

When I say negative, I mean that `--empty` is saying that there's something that should *not* be created.

Now, it so happens that for backwards compat, our users expect that `--createProject` should be the default since that matches what we do today.

Brocco: I certainly concede to that idea, but one of the principles on which the CLI was started (at least in my eyes) was to get started quickly by reducing the barrier to entry. With that being said, I think the idea of changing the default behavior of `ng new` to not create a project would be a mistake.

Also, I think this conversation could/should be moved over to the actual issue for historical purposes.

alexeagle:
I added a comment on the attached issue, good point about keeping discussion there

alexeagle:
Discussed in team meeting, we'd like the new option to be `--createApplication` and it defaults to true.

We decided having positive flags outweighs the desire to have defaults be false.

Brocco:
Thanks @alexeagle I will update the PR to reflect that decision.

Brocco:
Possible 3rd option...

instead of having a boolean create an enumeration option named `project` that can have 3 potential values...
`application` (default)
`library` (would create just a library)
`none` (which would create an empty workspace)

alexeagle:
fine with me to have `--createProject=[application,library,none]`
so long as there isn't much extra code to handle the `--createProject=library` case

hansl:
Leaving this opened, but it seems it will miss the 7.0 RC deadline. I'll keep it on our backlog so we can look at it again around 7.1. Cheers!

PR #12229, opened 2018-09-10
refactor(@schematics/angular): make interaction with architect targets type safe

alan-agius4:


packages/schematics/angular/utility/project-targets.ts
```diff
@@ -6,15 +6,31 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { experimental } from '@angular-devkit/core';
+import { SchematicsException, Tree } from '@angular-devkit/schematics';
+import { getProject, isWorkspaceProject } from './project';
+import { WorkspaceProject, WorkspaceSchema, WorkspaceTarget } from './workspace-models';
 
+export function getProjectTargets(project: WorkspaceProject): WorkspaceTarget;
```
filipesilva: Could we do without these two first signatures? The third one seems to be a superset of them which makes me think it'd be the only one needed.

packages/schematics/angular/utility/workspace-models.ts
```diff
@@ -0,0 +1,150 @@
...
+export type E2EBuilderTarget = BuilderTarget<Builders.Protractor, E2EOptions>;
+
+export interface WorkspaceSchema extends experimental.workspace.WorkspaceSchema {
+    projects: {
+        [key: string]: WorkspaceProject<ProjectType.Application | ProjectType.Library>;
+    };
+}
+
+// tslint:disable-next-line:no-any
+export interface WorkspaceProject<TProjectType extends ProjectType = any>
```
filipesilva: Could we do without the `any` here? The main usecase seems to be `ProjectType.Application` so that seems a sensible default.

packages/schematics/angular/utility/workspace-models.ts
```diff
@@ -0,0 +1,150 @@
...
+     * Tool options.
+     */
+    architect?: WorkspaceTarget<TProjectType>;
+    /**
+     * Tool options.
+     */
+    targets?: WorkspaceTarget<TProjectType>;
+}
+
+export interface WorkspaceTarget<TProjectType extends ProjectType = ProjectType.Application> {
```
filipesilva: This should be `WorkspaceTargets` (plural) where any of the keys are a target proper.

packages/schematics/angular/utility/workspace-models.ts
alan-agius4: The thing is that having it as a default would break a lot of usages, and would require a bit of boilerplate for usages. Example;
```ts
const project = getProject<ProjectType.Application>("foo");
```
Because it will have a cascading effect with such errors;
```
'WorkspaceProject<ProjectType>' is not assignable to type 'WorkspaceProject<ProjectType.Application>'.
```

Alternativly if you prefer we can do two methods 
```ts
const project1 = getLibraryProject("foo");
const project2 = getApplicationProject("foo");
```

filipesilva: With the diff below I think the changes are complete. 
```
$ git diff
diff --git a/packages/schematics/angular/utility/config.ts b/packages/schematics/angular/utility/config.ts
index 5033f77c5..a1e0018b1 100644
--- a/packages/schematics/angular/utility/config.ts
+++ b/packages/schematics/angular/utility/config.ts
@@ -7,7 +7,7 @@
  */
 import { JsonParseMode, parseJson } from '@angular-devkit/core';
 import { Rule, SchematicContext, SchematicsException, Tree } from '@angular-devkit/schematics';
-import { WorkspaceProject, WorkspaceSchema } from './workspace-models';
+import { ProjectType, WorkspaceProject, WorkspaceSchema } from './workspace-models';


 // The interfaces below are generated from the Angular CLI configuration schema
@@ -497,10 +497,10 @@ export function getWorkspace(host: Tree): WorkspaceSchema {
   return parseJson(content, JsonParseMode.Loose) as {} as WorkspaceSchema;
 }

-export function addProjectToWorkspace(
+export function addProjectToWorkspace<TProjectType extends ProjectType = ProjectType.Application>(
   workspace: WorkspaceSchema,
   name: string,
-  project: WorkspaceProject,
+  project: WorkspaceProject<TProjectType>,
 ): Rule {
   return (host: Tree, context: SchematicContext) => {

diff --git a/packages/schematics/angular/utility/project.ts b/packages/schematics/angular/utility/project.ts
index caf3c1c76..13df722f0 100644
--- a/packages/schematics/angular/utility/project.ts
+++ b/packages/schematics/angular/utility/project.ts
@@ -24,15 +24,15 @@ export function buildDefaultPath(project: WorkspaceProject): string {
   return `${root}${projectDirName}`;
 }

-export function getProject(
+export function getProject<TProjectType extends ProjectType = ProjectType.Application>(
   workspaceOrHost: WorkspaceSchema | Tree,
   projectName: string,
-): WorkspaceProject {
+): WorkspaceProject<TProjectType> {
   const workspace = isWorkspaceSchema(workspaceOrHost)
     ? workspaceOrHost
     : getWorkspace(workspaceOrHost);

-  return workspace.projects[projectName];
+  return workspace.projects[projectName] as WorkspaceProject<TProjectType>;
 }

 export function isWorkspaceSchema(workspace: unknown): workspace is WorkspaceSchema {
diff --git a/packages/schematics/angular/utility/workspace-models.ts b/packages/schematics/angular/utility/workspace-models.ts
index 6eb5c9a4f..b262e26d1 100644
--- a/packages/schematics/angular/utility/workspace-models.ts
+++ b/packages/schematics/angular/utility/workspace-models.ts
@@ -119,8 +119,7 @@ export interface WorkspaceSchema extends experimental.workspace.WorkspaceSchema
     };
 }

-// tslint:disable-next-line:no-any
-export interface WorkspaceProject<TProjectType extends ProjectType = any>
+export interface WorkspaceProject<TProjectType extends ProjectType = ProjectType.Application>
     extends experimental.workspace.WorkspaceProject {
     /**
     * Project type.
```

In `packages/schematics/angular/library/index.ts`, line 172, the type is then inferred as `ProjectType.Library` with no need to specify it:
image

alan-agius4: @filipesilva thanks so much for this. Feeling silly, that I didn't think of it before.

filipesilva: It's a bit more work but in the past it has paid dividends. Usually every `any` we manage to get rid of prevents regressions down the line.

kyliau:
There's a conflict merging this commit to the patch 6.2.x branch. Could you please take a look?

alan-agius4:
@kyliau updated to target master

PR #12233, opened 2018-09-11
Initial Cleanup of Webpack Plugin

clydin:


.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ jobs:
     <<: *defaults
     steps:
       - attach_workspace: *attach_options
-      - run: npm run test -- --code-coverage --full
```
filipesilva: Maybe we can move the code coverage to Bazel tests. But I don't know if it will support async/await there either.

What breaks btw? If we don't use the code coverage facility we might as well remove the code for it, since it will rot over time unused.

Also cc @hansl since he was that one that added code coverage in the first place.

packages/ngtools/webpack/package.json
```diff
@@ -31,10 +31,13 @@
     "webpack-sources": "1.2.0"
   },
   "peerDependencies": {
+    "@angular/compiler-cli": ">= 5.0.0 <8.0.0",
```
filipesilva: I think this will cause unmet peerdep errors in CLI projects, because devkit/build-angular (who imports ngtools/webpack) does not install angular/compiler-cli. Can you check this with a local build please?

.circleci/config.yml
clydin: The code coverage reports are neither displayed or saved so running with it only served to increase the run time.

The version of Istanbul used in the repository is outdated and doesn't support newer ES features.  Upgrading code coverage would probably be a good followup.

packages/ngtools/webpack/package.json
clydin: It doesn't like beta versions at the moment (fixing that), but otherwise no peer dependency warnings.

PR #12255, opened 2018-09-12
fix(@angular/cli): remove eject command stub

clydin:


hansl:
What's the issue that this is fixing?

clydin:
https://github.com/angular/angular-cli/pull/10946#issuecomment-415496387

hansl:
> @IgorMinar we discussed again, we don't plan to add back the eject command, instead we will follow the proposal to provide some docs on our wiki and a webpack config or seed repo that shows how to do it.

Just talked to @alexeagle and @StephenFluin; the consensus is that we should have an eject command but show documentation and a link to the wiki documentation explaining why we don't support it and to use ngx-build-plus.

Could you talk to Stephen to get the proper messaging?

PR #12277, opened 2018-09-14
refactor(@schematics/angular): Parameterize findModule

kyliau:
This commit modifies `findModule` to accept custom file extensions
so that a different filename convention for Angular modules is allowed.

### DIFF
--- a/packages/schematics/angular/utility/find-module_spec.ts
+++ b/packages/schematics/angular/utility/find-module_spec.ts
@@ -74,6 +74,25 @@ describe('find-module', () => {
         expect(err.message).toMatch(/Could not find an NgModule/);
       }
     });
+
+    it('should accept custom ext for module', () => {
+      const host = new EmptyTree();
+      const modulePath = '/foo/src/app/app_module.ts';
+      host.create(modulePath, 'app module');
+      // Should find module if given a custom ext
+      const foundModule = findModule(host, 'foo/src/app/bar', '_module.ts');
+      expect(foundModule).toBe(modulePath);
+      // Should not find module if using default ext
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/utility/find-module_spec.ts
```diff
@@ -74,6 +74,14 @@ describe('find-module', () => {
         expect(err.message).toMatch(/Could not find an NgModule/);
       }
     });
+
+    it('should accept custom ext for module', () => {
```
hansl: Could you add a falsifying test that checks that it doesn't match previous files when using the new arguments?

packages/schematics/angular/utility/find-module_spec.ts
```diff
@@ -100,5 +108,14 @@ describe('find-module', () => {
       const modPath = findModuleFromOptions(tree, options);
       expect(modPath).toEqual('/projects/my-proj/src/admin/foo.module.ts' as Path);
     });
+
+    it('should find a module using custom ext', () => {
```
hansl: Same comment.

kyliau: I changed the behavior such that if `moduleExt` is specified, the filename must conform to the ext pattern. So even if a file exists but it does not conform to the pattern, an error will be thrown. Please see the test case below.

kyliau: On second thought, maybe that's bad idea. Instead, it should ignore `moduleExt` if `module` or `module + .ts` exists.

hansl:
Moving this to master only as its specific to google3.

PR #12284, opened 2018-09-15
Set of fixes for better npmrc parameters support

smnbbrv:
Heavily related to a hot topic #10624. This issue is really annoying for lots of people.

Targets two points:

1. Support for per-project `.npmrc` file. This is done for the projects that include bundled `.npmrc` with readonly permissions, see https://docs.npmjs.com/files/npmrc
2. Support for private repositories like Nexus and Artifactory. I did not test with Artifactory because currently I have none, however Nexus works perfectly with the applied changes. Artifactory has a similar way of authentication, so it should also be covered.

Changes type:
- `alwaysAuth` is changed to `always-auth`, which is a proper name, see https://docs.npmjs.com/misc/config#always-auth
- added email parameter
- `_auth` is normally a base64 encoded `token:password` string => added extracting those credentials from the token and left a fallback token assignment in case something else is stored in `_auth`

Thank you in advance.

### DIFF
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -56,7 +56,7 @@ function _readNpmRc(): Observable<{ [key: string]: string }> {
 
     allOptionsArr.forEach(x => {
       const [key, ...value] = x.split('=');
-      allOptions[key] = value.join('=');
+      allOptions[key.trim()] = value.join('=').trim();
     });
 
     subject.next(allOptions);
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -33,15 +33,21 @@ function _readNpmRc(): Observable<{ [key: string]: string }> {
     // TODO: have a way to read options without using fs directly.
     const path = require('path');
     const fs = require('fs');
+    const perProjectNpmrc = path.resolve('.npmrc');
 
     let npmrc = '';
-    if (process.platform === 'win32') {
-      if (process.env.LOCALAPPDATA) {
-        npmrc = fs.readFileSync(path.join(process.env.LOCALAPPDATA, '.npmrc')).toString('utf-8');
-      }
+
+    if (fs.existsSync(perProjectNpmrc)) {
+      npmrc = fs.readFileSync(perProjectNpmrc).toString('utf-8');
     } else {
-      if (process.env.HOME) {
-        npmrc = fs.readFileSync(path.join(process.env.HOME, '.npmrc')).toString('utf-8');
+      if (process.platform === 'win32') {
+        if (process.env.LOCALAPPDATA) {
+          npmrc = fs.readFileSync(path.join(process.env.LOCALAPPDATA, '.npmrc')).toString('utf-8');
+        }
+      } else {
+        if (process.env.HOME) {
+          npmrc = fs.readFileSync(path.join(process.env.HOME, '.npmrc')).toString('utf-8');
+        }
       }
     }
 
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -198,7 +198,8 @@ export function getNpmPackageJson(
         getNpmConfigOption('_authToken', registryKey),
         getNpmConfigOption('username', registryKey, true),
         getNpmConfigOption('password', registryKey, true),
-        getNpmConfigOption('alwaysAuth', registryKey, true),
+        getNpmConfigOption('email', registryKey, true),
+        getNpmConfigOption('always-auth', registryKey, true),
       ).pipe(
         toArray(),
         concatMap(options => {
@@ -211,6 +212,7 @@ export function getNpmPackageJson(
             authToken,
             username,
             password,
+            email,
             alwaysAuth,
           ] = options;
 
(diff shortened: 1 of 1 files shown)


packages/schematics/update/update/npm.ts
```diff
@@ -216,17 +224,36 @@ export function getNpmPackageJson(
...
+
           if (authToken) {
             auth.token = authToken;
           } else if (token) {
-            auth.token = token;
+            try {
+              // attempt to parse username and password from base64 token
+              // to enable Artifactory / Nexus-like repositories support
+              const parsedToken = Buffer.from(token, 'base64').toString('ascii');
+              const [extractedUsername, extractedPassword] = parsedToken.split(':');
```
hansl: Looks like it cannot have `:` in the password. Is this expected? You might want to use this instead:

```typescript
const delimiter = ':';
const [extractedUsername, ...passwordArr] = parsedToken.split(delimiter);
const extractedPassword = passwordArr.join(delimiter);
```

Just slightly more ugly, but there's no native function that does it.

smnbbrv: true, good point. Thanks for reviewing! I will apply the changes soon

smnbbrv:
@hansl updated the PR according to your comments and going to sleep :) If you have more comments / there is something more to fix, I'm glad to help

mrmaxsteel:
FYI tested with Artifactory, did not work, see: https://github.com/angular/angular-cli/issues/10624#issuecomment-421993521

mrmaxsteel:
@smnbbrv I found the issue, your code to parse the username:password from the base64 `_auth` token doesn't work when there is whitespace between key and value in .npmrc.

For example, it works when .npmrc looks like:
```
_auth=<base64 user:pass>
```
But not when:
```
_auth = <base64 user:base>
```

smnbbrv:
@hansl I added another fix that adds support of whitespaces around '='. Could you, please, revalidate?

@mrmaxsteel it's not really my code, I'm just a random stranger who wants the issue to be fixed :D Added fix to your findings. That's really cool that Artifactory works as well 👍

mrmaxsteel:
Validated latest change, all works 👍

smnbbrv:
@hansl I can't really understand the ci failures... Does not seem to me like it is the issue connected to the changes I made

packages/schematics/update/update/npm.ts
```diff
@@ -216,17 +224,38 @@ export function getNpmPackageJson(
...
+              const delimiter = ':';
+              const parsedToken = Buffer.from(token, 'base64').toString('ascii');
+              const [extractedUsername, ...passwordArr] = parsedToken.split(delimiter);
+              const extractedPassword = passwordArr.join(delimiter);
+
+              if (extractedUsername && extractedPassword) {
+                auth.username = extractedUsername;
+                auth.password = extractedPassword;
+              } else {
+                throw new Error('Unable to extract username and password from _auth token');
```
aguacongas: Why do you try catch and throw an error instead to set auth.token instead of just setting the token in the else block ?

packages/schematics/update/update/npm.ts
```diff
@@ -33,15 +33,21 @@ function _readNpmRc(): Observable<{ [key: string]: string }> {
...
 
     let npmrc = '';
-    if (process.platform === 'win32') {
-      if (process.env.LOCALAPPDATA) {
-        npmrc = fs.readFileSync(path.join(process.env.LOCALAPPDATA, '.npmrc')).toString('utf-8');
-      }
+
+    if (fs.existsSync(perProjectNpmrc)) {
+      npmrc = fs.readFileSync(perProjectNpmrc).toString('utf-8');
     } else {
```
aguacongas: You should parse per project .npmrc and user .nmprc because the repository url can be in the per project file and credentials in the user file

packages/schematics/update/update/npm.ts
smnbbrv: because the error message gives more information for the code reader + it is easier later (in case somebody wants it) to throw an error which is way more user-friendly than swallowing it and assigning a token

smnbbrv: however this is just a matter of taste

packages/schematics/update/update/npm.ts
smnbbrv: True.

However if we are too picky here we enter the pain zone. First, then we need to support the global npmrc + userconfig flag / env variable as well. Second, we need to properly resolve the collisions between _authToken and _auth (because both could be present and we need to pick). Finally, the current approach is reading npmrc every time the package info is getting resolved which is not a standard npm behavior. Etc.

I'm not yet talking about having two different ways of getting NPM parameters.

The whole approach is not perfect and this PR does not pretend to be a fully working solution; at least it would be way more helpful than the current approach. Probably one needs to deeply rethink the concept.

packages/schematics/update/update/npm.ts
aguacongas: I get your point, regarding performance it's not really good but we don't realy need performance it that case

packages/schematics/update/update/npm.ts
smnbbrv: The private repos are nearly never used open source, hence no auth is required. Also if you change the order of resolving those files then you go in clear conflict with Npm itself

aguacongas: #10624 is regarding an issue with fontawesome pro, so I guess it's used in open source. and npm don't need authentication nor .npmrc.

smnbbrv: https://fontawesome.com/pro - 60$ per license for 5 seats. What open source projects are you talking about?

smnbbrv: OK, I have more arguments, but the discussion went away from the issue. What we discuss is only the cases when there is a project-specific .npmrc which is not containing `_auth` or `_authToken`, but there is one in some of the .npmrc above right?

aguacongas: Exactly, in that case, your PR doesn't fix the issue I guess

smnbbrv:
closing in favour of https://github.com/angular/angular-cli/pull/12350

hansl:
I'm going to let this in, but we should fix the local/global issue in RC.

PR #12286, opened 2018-09-17
feat(@ngtools/webpack): add support for `traceResolution`

### LINKED ISSUE #8676: how to turn on module resolution stats provided by typescript compiler for an  angular cli app?
I would like to trace moduleResolution stats while compiling app code. the compiler is not printing details on how modules resolved.  Is this expected?

my tsconfig.app.json is like this
{
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "traceResolution": true,
    "baseUrl": "./",
    "module": "es2015",
    "types": []
  }
}

tsconfig

### Versions
Angular CLI: 1.5.4
Node: 8.9.1
OS: win32 x64
Angular: 5.0.3

@angular/cli: 1.5.4
@angular-devkit/core: 0.0.21
@angular-devkit/schematics: 0.0.37
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.4
@schematics/angular: 0.1.7
typescript: 2.4.2
webpack: 3.8.1
```



### Observed behavior
$ ng build --app 1
Date: 2017-11-29T12:28:09.911Z
Hash: 289e296055d0854f9bc4
Time: 6474ms
chunk {inline} inline.bundle.js, inline.bundle.js.map (inline) 5.83 kB [entry] [rendered]
chunk {main} main.bundle.js, main.bundle.js.map (main) 7.81 kB [initial] [rendered]
chunk {polyfills} polyfills.bundle.js, polyfills.bundle.js.map (polyfills) 200 kB [initial] [rendered]
chunk {styles} styles.bundle.js, styles.bundle.js.map (styles) 11.4 kB [initial] [rendered]
chunk {vendor} vendor.bundle.js, vendor.bundle.js.map (vendor) 2.29 MB [initial] [rendered]

I am trying to merge two apps at runtime and generate webpack assets. traceResolution stats would be helpful in debugging issue with my tsconfig files.

### PR DESCRIPTION
**Pull request by alan-agius4**
Closes #8676

### DIFF
--- a/packages/ngtools/webpack/src/compiler_host.ts
+++ b/packages/ngtools/webpack/src/compiler_host.ts
@@ -435,6 +435,10 @@ export class WebpackCompilerHost implements ts.CompilerHost {
       return this.readFile(fileName);
     }
   }
+
+  trace(message: string) {
+    console.log(message);
+  }
 }
 
 
--- a/tests/legacy-cli/e2e/tests/misc/trace-resolution.ts
+++ b/tests/legacy-cli/e2e/tests/misc/trace-resolution.ts
@@ -1,0 +1,22 @@
+import { ng } from '../../utils/process';
+import { updateJsonFile } from '../../utils/project';
+
+export default async function () {
+  await updateJsonFile('tsconfig.json', tsconfig => {
+    tsconfig.compilerOptions.traceResolution = true;
+  });
+
+  const { stdout: stdoutTraced } = await ng('build');
+  if (!/Resolving module/.test(stdoutTraced)) {
+    throw new Error(`Modules resolutions must be printed when 'traceResolution' is enabled.`);
+  }
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: packages/ngtools/webpack/src/compiler_host.ts [hunk 1]
```diff
@@ -435,6 +435,10 @@ export class WebpackCompilerHost implements ts.CompilerHost {
       return this.readFile(fileName);
     }
   }
+
+  trace(message: string) {
+    console.log(message);
```
filipesilva: I don't think we should use `console.log` for this, but not sure what else we should use. The webpack compilation only allows for warnings and errors, and they show only at the end of the compilation.

@clydin WDYT?

alan-agius4:
@filipesilva updated scope as requested. Re the `console.log` I am not sure what else we can do instead.

PR #12297, opened 2018-09-18
fix(@angular/cli): support `--version` option

clydin:
It was supported in 6.x and appears to have been lost in the 7.0 betas.

### DIFF
--- a/packages/angular/cli/models/command-runner.ts
+++ b/packages/angular/cli/models/command-runner.ts
@@ -115,7 +115,11 @@ export async function runCommand(
 
   // if no commands were found, use `help`.
   if (commandName === undefined) {
-    commandName = 'help';
+    if (args.length === 1 && args[0] === '--version') {
+      commandName = 'version';
+    } else {
+      commandName = 'help';
+    }
   }
 
   let description: CommandDescription | null = null;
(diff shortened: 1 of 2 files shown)


tests/legacy-cli/e2e/tests/misc/version.ts
```diff
@@ -1,13 +1,19 @@
...
-    .then(() => ng('version'))
-    // Doesn't fail outside a project.
-    .then(() => process.chdir('/'))
-    .then(() => ng('version'));
+import { deleteFile } from '../../utils/fs';
+import { ng } from '../../utils/process';
+
+
+export default async function() {
+  await ng('version');
```
hansl: Could you take the stdout from this and verify it's the same as with `--version`.

packages/angular/cli/models/command-runner.ts
```diff
@@ -115,7 +115,11 @@ export async function runCommand(
 
   // if no commands were found, use `help`.
   if (commandName === undefined) {
-    commandName = 'help';
+    if (args.length === 1 && args[0] === '--version') {
```
hansl: Do you think it would make sense to make this an argument to the `help` command?

tests/legacy-cli/e2e/tests/misc/version.ts
clydin: Done.

packages/angular/cli/models/command-runner.ts
clydin: I think I'd keep them separate.  Their underlying purpose and implementation is sufficiently different.

PR #12300, opened 2018-09-18
Various Help and Parsing features and fixes

hansl:
Added a new `usageNotes` field to command descriptions, moved suboptions to subcommands which are tied to options (removing the awkward `suboption` type), removed unused field `type` from the interface, added enums and multiple types support to the argument parser, and moved `--help-json` to `--help=json`.


packages/angular/cli/utilities/json-schema.ts
```diff
@@ -73,13 +73,29 @@ export async function parseJsonSchemaToCommandDescription(
     const ldPath = resolve(dirname(jsonPath), schema.$longDescription);
     longDescription = readFileSync(ldPath, 'utf-8');
   }
+  let usageNotes = '';
+  if (typeof schema.$usageNotes == 'string' && schema.$usageNotes) {
+    const unPath = resolve(dirname(jsonPath), schema.$usageNotes);
+    usageNotes = readFileSync(unPath, 'utf-8');
```
clydin: error handling if the file does not exist?

packages/angular/cli/utilities/json-schema.ts
```diff
@@ -73,13 +73,29 @@ export async function parseJsonSchemaToCommandDescription(
...
   const type = _getEnumFromValue(schema.$type, CommandType, CommandType.Default);
   const description = '' + (schema.description === undefined ? '' : schema.description);
   const hidden = !!schema.$hidden;
 
-  return { name, description, longDescription, hidden, type, options, aliases, scope, impl };
+  return {
+    name,
+    description,
+    ...(longDescription ? { longDescription } : {}),
+    ...(usageNotes ? { usageNotes } : {}),
```
clydin: are the spread operators needed here?

packages/angular_devkit/core/src/json/schema/utility.ts
```diff
@@ -23,6 +24,30 @@ export function getTypesOfSchema(schema: JsonObject | true): Set<string> {
     potentials = new Set([schema.type]);
   } else if (Array.isArray(schema.type)) {
     potentials = new Set(schema.type as string[]);
+  } else if (isJsonArray(schema.enum)) {
+    // Gather the type of each enum values, and use that as a starter for potential types.
+    const filteredTypes = [...schema.enum].map(v => {
```
clydin: for/of would probably be cleaner here and reduce the iterations from four to one.

packages/angular/cli/commands/definitions.json
```diff
@@ -30,12 +30,9 @@
...
         "help": {
-          "type": "boolean",
-          "description": "Shows a help message."
-        },
-        "helpJson": {
-          "type": "boolean",
-          "description": "Shows the metadata associated with each flags, in JSON format."
+          "type": ["boolean", "string"],
+          "description": "Shows a help message. Use '--help=json' as a value to output the help in JSON format.",
+          "default": false
```
clydin: `"enum": [true, "json"]`?

or more maybe:
```
  "anyOf": [
    {
      "enum": [ true, "json" ]
    },
    {
      "pattern": "^[Jj][Ss][Oo][Nn]$"
    }
  ]
```

hansl: The first one would not handle the `—no-help` case. The second one wouldn’t list the enumerations properly.

packages/angular/cli/utilities/json-schema.ts
hansl: I don’t want the keys to be in the object, even if the value is undefined.

packages/angular/cli/commands/definitions.json
clydin: the tsconfig schema uses that structure to handle casing and vscode displays the enum values (not sure what its code is doing).

clydin: I'm ok with it either way though.

packages/angular/cli/utilities/json-schema.ts
hansl: Done.

packages/angular_devkit/core/src/json/schema/utility.ts
hansl: Done. It does look cleaner.

packages/angular/cli/commands/definitions.json
hansl: I would think vscode is showing it from the type itself? The `.d.ts` is cleaner.

I also changed the code of the parser to reject values that are not part of the enum (and adding it to leftovers). I wasn't sure what sort of behaviour would be correct here, TBH. If we set it and it's not part of the enum, then I would consider that a breach of contract. Ignoring the flag might be slightly better but lead to user errors.

I could use some whiteboarding on this actually.

PR #12304, opened 2018-09-19
Correct Several Cases of Implicit Returns

clydin:


clydin:
TS has the `noImplicitReturns` option which could eventually be enabled.  For this PR, I just hit the most obvious cases.

PR #12314, opened 2018-09-20
Move reflect metadata from the polyfills to the dev environment file.

hansl:


packages/schematics/angular/application/files/src/environments/environment.ts
```diff
@@ -2,6 +2,11 @@
 // `ng build --prod` replaces `environment.ts` with `environment.prod.ts`.
 // The list of file replacements can be found in `angular.json`.
 
+// Used for reflect-metadata in JIT. For size and performance reasons, this should not be imported
+// in AOT unless absolutely needed. All Angular metadata is removed in AOT so this is only needed,
+// by default, in development.
+import 'core-js/es7/reflect';
```
filipesilva: I don't think the environment file is the right place to add polyfills. 

Environments here are for runtime variables, and we encourage users to add API keys and such here. It's common to test your app over the production runtime environment on a development build.

We now have general purpose file replacements as well. Why not have `polyfills.ts` and `polyfills.aot.ts` and effect the replacement on AOT configurations instead? The default production configuration would replace both `environment.ts` and `polyfills.ts`.

hansl: That's not a bad idea, but it will make the update step more complicated.

hansl: actually since the polyfill file is an option we should just override _that_.

clydin: I'm not sure we can automatically move the polyfill without breaking existing applications.  Some may rely on the reflect APIs which will have mysteriously disappeared after updating.

hansl: Update would: copy the file, remove the reflect.

New project gets a dev polyfills that imports the AOT one.

clydin: doesn't have to be a copy.  The JIT file would just contain reflect and then import the original polyfills file.

But that brings up another problem: `ng build --prod --no-aot`

clydin: The ideal scenario here would be for the CLI to analyze the code and only include the polyfill if actually used.

IgorMinar:
after further discussion it's likely that we'll go with #12346 instead.

hansl:
I'll reuse this PR for the migration

PR #12332, opened 2018-09-20
docs: example command to generate library

codermarcos:
Minor change in the overview part i included the command example to generate a library.


codermarcos:
I signed it!

PR #12335, opened 2018-09-21
Conditional IE Polyfill Concept

clydin:
* No-Cost IE9-11 support
  - no need to manually import and manage individual IE9-11 polyfills
  - Required ES2015 polyfills are only loaded by browsers that require them
  - Controllable via a new `supportIE` option; no behavior change if option is not enabled
  - `supportIE` is enabled in newly generated projects
  - allows for custom polyfill solutions by not using the new option.
  - works in conjunction with conditional JIT polyfills which will load IE specific JIT polyfills

### DIFF
--- a/tests/legacy-cli/e2e/tests/build/polyfills.ts
+++ b/tests/legacy-cli/e2e/tests/build/polyfills.ts
@@ -1,19 +1,34 @@
-import { expectFileToMatch } from '../../utils/fs';
-import { ng } from '../../utils/process';
 import { oneLineTrim } from 'common-tags';
+import {
+  expectFileSizeToBeUnder,
+  expectFileToExist,
+  expectFileToMatch,
+  getFileSize,
+} from '../../utils/fs';
+import { ng } from '../../utils/process';
+import { expectToFail } from '../../utils/utils';
 
-export default function () {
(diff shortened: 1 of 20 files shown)


tests/legacy-cli/e2e/tests/build/polyfills.ts
```diff
@@ -1,18 +1,26 @@
...
 
-  return Promise.resolve()
-    .then(() => ng('build'))
+    await ng('build', '--aot');
     // files were created successfully
-    .then(() => expectFileToMatch('dist/test-project/polyfills.js', 'core-js'))
-    .then(() => expectFileToMatch('dist/test-project/polyfills.js', 'zone.js'))
-    // index.html lists the right bundles
-    .then(() => expectFileToMatch('dist/test-project/index.html', oneLineTrim`
+    await expectToFail(() => expectFileToMatch('dist/test-project/polyfills.js', 'core-js/es7/reflect'));
```
IgorMinar: @clydin can you please make this assertion more explicit? we don't want to pass through this if expectFileToMatch fails for other reasons (like file not found).

clydin: Updated.
The polyfills file now needs to also exist and the AOT file size needs to be less than the JIT file size.

PR #12346, opened 2018-09-21
JIT Conditional Polyfills

clydin:


packages/schematics/angular/application/files/src/polyfills.ts
```diff
@@ -40,12 +40,6 @@
 /** IE10 and IE11 requires the following for the Reflect API. */
 // import 'core-js/es6/reflect';
 
-
-/** Evergreen browsers require these. **/
-// Used for reflect-metadata in JIT. If you use AOT (and only Angular decorators), you can remove.
-import 'core-js/es7/reflect';
```
hansl: Are we not removing es6 reflect for IE10/11 ?  (see above in the code)

hansl: My question basically is; is leaving the commented out polyfill the line above intended or not?

hansl: Confirmed offline with Charles that it was intentional. Approving.

PR #12350, opened 2018-09-22
feat(@schematics/update): add full support of npmrc / npm config

### LINKED ISSUE #10624: "ng update" doesn't support private repos such as FontAwesome 5 Pro

smnbbrv:
A logical consequence of https://github.com/angular/angular-cli/pull/12284.

Fully fixes #10624 and related issues.

The current npm config support is not full and most likely cannot ever become one. There are lots of (legacy) authentication methods, hierarchical .npmrc, all that is running in various OS, etc.

This is a brand new approach: npm is used as a service. It appears that internally npm not only allows reading all the configuration (including _auth, _authToken etc), but also already provides a resolver for various authentication models.

What is changed:
- reduced file size by a factor of 2
- improved performance
- less async, more sync => way more readable, and, btw, more performant

and the most important:
- so much less dirty code
- there is no need to deal with NPM config and authentication anymore: NPM does it as NPM would do it

How to try on existing @angular/cli project:
- npm i -D npm
- paste the code below into `your_project/node_modules/@schematics/update/update/npm.js`

```js
"use strict";
const fs_1 = require("fs");
const npm = require("npm");
const rxjs_1 = require("rxjs");
const operators_1 = require("rxjs/operators");
const url = require("url");
const RegistryClient = require('npm-registry-client');
const npmPackageJsonCache = new Map();
const ensuredNpm = new rxjs_1.Observable(subject => {
    npm.load(() => {
        subject.next();
        subject.complete();
    });
}).pipe(operators_1.shareReplay());
function getNpmConfigOption(option, scope) {
    if (scope) {
        return npm.config.get(`${scope}:${option}`) || getNpmConfigOption(option);
    }
    return npm.config.get(option);
}
function getNpmClientSslOptions(strictSsl, cafile) {
    const sslOptions = {};
    if (strictSsl === 'false') {
        sslOptions.strict = false;
    }
    else if (strictSsl === 'true') {
        sslOptions.strict = true;
    }
    if (cafile) {
        sslOptions.ca = fs_1.readFileSync(cafile);
    }
    return sslOptions;
}
/**
 * Get the NPM repository's package.json for a package. This is p
 * @param {string} packageName The package name to fetch.
 * @param {string} registryUrl The NPM Registry URL to use.
 * @param {LoggerApi} logger A logger instance to log debug information.
 * @returns An observable that will put the package.json content.
 * @private
 */
function getNpmPackageJson(packageName, registryUrl, logger) {
    const scope = packageName.startsWith('@') ? packageName.split('/')[0] : undefined;
    return ensuredNpm.pipe(operators_1.map(() => {
        const partialUrl = registryUrl
            || getNpmConfigOption('registry', scope)
            || 'https://registry.npmjs.org/';
        const partial = url.parse(partialUrl);
        let fullUrl = new url.URL(`http://${partial.host}/${packageName.replace(/\//g, '%2F')}`);
        try {
            const registry = new url.URL(partialUrl);
            registry.pathname = (registry.pathname || '')
                .replace(/\/?$/, '/' + packageName.replace(/\//g, '%2F'));
            fullUrl = new url.URL(url.format(registry));
        }
        catch (_a) { }
        logger.debug(`Getting package.json from '${packageName}' (url: ${JSON.stringify(fullUrl)})...`);
        return fullUrl;
    }), operators_1.switchMap(fullUrl => {
        let maybeRequest = npmPackageJsonCache.get(fullUrl.toString());
        if (maybeRequest) {
            return maybeRequest;
        }
        const subject = new rxjs_1.ReplaySubject(1);
        const auth = npm.config.getCredentialsByURI(fullUrl.toString());
        const client = new RegistryClient({
            proxy: {
                http: getNpmConfigOption('proxy'),
                https: getNpmConfigOption('https-proxy'),
            },
            ssl: getNpmClientSslOptions(getNpmConfigOption('strict-ssl'), getNpmConfigOption('cafile')),
        });
        client.log.level = 'silent';
        client.get(fullUrl.toString(), { auth, timeout: 30000 }, (error, data) => {
            if (error) {
                subject.error(error);
            }
            subject.next(data);
            subject.complete();
        });
        maybeRequest = subject.asObservable();
        npmPackageJsonCache.set(fullUrl.toString(), maybeRequest);
        return maybeRequest;
    }));
}
exports.getNpmPackageJson = getNpmPackageJson;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "@types/loader-utils": "^1.1.3",
     "@types/minimist": "^1.2.0",
     "@types/node": "8.10.10",
+    "@types/npm": "2.0.29",
     "@types/request": "^2.47.1",
     "@types/semver": "^5.5.0",
     "@types/source-map": "0.5.2",
--- a/packages/schematics/update/package.json
+++ b/packages/schematics/update/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@angular-devkit/core": "0.0.0",
     "@angular-devkit/schematics": "0.0.0",
+    "npm": "6.4.1",
     "npm-registry-client": "8.6.0",
     "semver": "5.5.1",
     "semver-intersect": "1.4.0",
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -6,120 +6,30 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import { logging } from '@angular-devkit/core';
-import { exec } from 'child_process';
 import { readFileSync } from 'fs';
-import { Observable, ReplaySubject, concat, of } from 'rxjs';
-import {
-  catchError,
-  concatMap,
-  defaultIfEmpty,
-  filter,
-  first,
-  map,
-  shareReplay,
-  toArray,
-} from 'rxjs/operators';
+import * as npm from 'npm';
+import { Observable, ReplaySubject } from 'rxjs';
+import { map, shareReplay, switchMap } from 'rxjs/operators';
 import * as url from 'url';
 import { NpmRepositoryPackageJson } from './npm-package-json';
 
 const RegistryClient = require('npm-registry-client');
 
 const npmPackageJsonCache = new Map<string, Observable<NpmRepositoryPackageJson>>();
-const npmConfigOptionCache = new Map<string, Observable<string | undefined>>();
 
-
-function _readNpmRc(): Observable<{ [key: string]: string }> {
-  return new Observable<{ [key: string]: string }>(subject => {
-    // TODO: have a way to read options without using fs directly.
-    const path = require('path');
-    const fs = require('fs');
-
-    let npmrc = '';
-    if (process.platform === 'win32') {
-      if (process.env.LOCALAPPDATA) {
-        npmrc = fs.readFileSync(path.join(process.env.LOCALAPPDATA, '.npmrc')).toString('utf-8');
-      }
-    } else {
-      if (process.env.HOME) {
-        npmrc = fs.readFileSync(path.join(process.env.HOME, '.npmrc')).toString('utf-8');
-      }
-    }
-
-    const allOptionsArr = npmrc.split(/\r?\n/).map(x => x.trim());
-    const allOptions: { [key: string]: string } = {};
-
-    allOptionsArr.forEach(x => {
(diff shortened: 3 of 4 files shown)


clydin:
Thank you for the effort on this contribution.  However, depending directly on `npm` is not a viable solution.  Beyond the license validation issues, the dependency tree is massive.  But most importantly, there does not appear to be an official public API which means the functions being used cannot be relied upon to continue to exist or function as currently assumed.  This presents a serious ongoing maintenance concern.
Yarn support is also being considered as an addition to `ng update` and it would definitely not be viable to depend directly on `yarn` as well.

smnbbrv:
hi @clydin 

thank you for quick review!

Let's have a look at those problems:

- license validation: I don't really see any problem with their license; of course I'm not a lawyer but they allow everything you want
- the dependency tree: well, for the tool such as angular/cli it is not a big problem, is it? The dependencies don't go to the browser
- no official API: the version of npm is fixed to 6.4.1. If there is a newer version that adds breaking changes you could still rely on 6.4.1 because it would be available locally in @schematics/update regardless of what version other packages might use

Let's have a look at what we currently have: `ng update` pretends to behave like NPM. So either one needs to use NPM to have functionality "as is" or massively replicate all NPM's features regarding auth and configuration.

Or, using NPM as a solution until they break the API **and** @schematics/update would need a newer version of NPM.

This issue https://github.com/angular/angular-cli/issues/10624 is open since `ng update` was introduced and it is a real pain. Many packages starting to say: hey, run `ng update` to update to the actual version, but what if people cannot?

PR #12356, opened 2018-09-24
fix(@angular/cli): allow global config command outside project

clydin:


hansl:
Cannot be merged in Patch, after he big refactor (which was master only).

PR #12357, opened 2018-09-24
chore(rxjs): update rxjs to 6.3.2

benlesh:
Updates CLI to 6.3.2

NOTE: There were two typings that needed to be updated here. One was an honest fix due to improved typings and the limitations of TypeScript, the other is a bit more curious (See the changes to base.ts). There's apparently a problem with spreading an array of operators into `pipe`, but only in this case. I'm working on replicating the issue outside of this project, but I haven't been able to do so.


packages/angular_devkit/core/src/json/schema/visitor.ts
```diff
@@ -69,7 +69,7 @@ function _visitJsonRecursive<ContextT>(
       ? value as Observable<JsonValue>
       : observableOf(value as JsonValue)
   ).pipe(
-    concatMap((value: JsonValue) => {
+    concatMap((value): Observable<JsonValue> => {
```
hansl: That’s not valid TypeScript is it?

hansl: It's missing the input type, or doesn't it need it?

benlesh: It's what Rado recommended in these situations. Unfortunately, TS can't infer the return in these cases where there's branching code.

benlesh: The input type is inferred.

alan-agius4:
@benlesh Thanks for this. However the E2E's will fail due to an issue that we had a chat about some time ago

https://github.com/ReactiveX/rxjs/issues/4135

packages/schematics/schematics/package.json
```diff
@@ -11,6 +11,7 @@
   "schematics": "./collection.json",
   "dependencies": {
     "@angular-devkit/core": "0.0.0",
-    "@angular-devkit/schematics": "0.0.0"
+    "@angular-devkit/schematics": "0.0.0",
+    "rxjs": "6.3.2"
```
alan-agius4: Why does this need to be added here? AFAIK schematics/schematics do not directly depend on `rxjs`

packages/angular_devkit/schematics/package.json
```diff
@@ -14,6 +14,6 @@
   ],
   "dependencies": {
     "@angular-devkit/core": "0.0.0",
-    "rxjs": "6.2.2"
+    "rxjs": "6.3"
```
alan-agius4: Shouldn't this be `6.3.2` for consistence? As this will be resolved to `6.3.x`

benlesh:
Closing in favor of https://github.com/angular/angular-cli/pull/12362

PR #12362, opened 2018-09-25
Rxjs 6.3

clydin:


benlesh:
Since RxJS 6.3.3 has been released, this should be unblocked now.

benlesh:
cc @alexeagle to confirm?

alexeagle:
Yeah, needs update to the non-breaking rxjs release. Then CLI users during RC will confirm that there isn't anything else broken when our packages use 6.3 but some schematics use 6.2

PR #12370, opened 2018-09-26
refactor(@angular-devkit/build-angular): clean up overriding of options

alan-agius4:


packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -391,23 +392,33 @@ export class DevServerBuilder implements Builder<DevServerBuilderOptions> {
...
+      'baseHref',
+      'progress',
+      'poll',
+    ];
+
+    // remove options that are undefined or not to be overrriden
+    const overrides = Object.keys(options)
+      .reduce<Partial<BrowserBuilderSchema>>(
+        (previous, key: DevServerBuilderOptionsKeys) => (
+          options[key] !== undefined && overridesOptions.includes(key)
```
hansl: It's simpler to just overload _if_ it's defined, no? The `...(x ? { x } : {})` trick was used because it aligns properly vertically and makes it easier to read, but in this case it makes it harder.

How about this?

```typescript
const overrides = Object.keys(options)
  .filter(key => options[key] !== undefined && overridesOptions.includes(key))
  .reduce((previous, key) => {
    previous[key] = options[key];

    return previous;
  });
```

alan-agius4: I updated the implementation however, had to change a bit the `reduce`

PR #12372, opened 2018-09-26
fix(@angular/cli): do not check if node_modules are inside sources dir

### LINKED ISSUE #11685: Regression: project is not working anymore from 6.0.8 -> 6.1.0 (in lerna monorepos)
```
- [ ] feature request
```

```
- [x] all
```

### Repro steps
Our angular/cli project ist located as one package of a larger lerna monorepo. We also make use of yarn workspaces, so all (or most) dependencies are hoisted within one single root node_modules folder. The resulting structure is something like
```
- node_modules <-- angular dependencies installed here
  - @angular
    - cli
    - core
    - ...
- packages
  - angular-app
    - package.json <-- angular dependencies defined here
    - src
    - ....
- package.json <-- no angular dependencies defined here
```

[this line](https://github.com/angular/angular-cli/blob/master/packages/angular/cli/upgrade/version.ts#L48) now (as of 6.1.0) prevents the project from working (at all) because `relativePotential.startsWith('..') === true` because in our structure `relativePotential = ../../node_modules/@angular/core/package.json`

**everything was working fine with 6.0.8**

You seem to not be depending on "@angular/core" and/or "rxjs". This is an error.

### Desired functionality
I can work on my angular project with latest versions of everything

nicolas-geniteau:
It should not verify that angular and rxjs packages are inside project
directory. To correctly resolve packages is enough.

Close #11685

Signed-off-by: Nicolas Geniteau <nicolas.geniteau@gmail.com>

### DIFF
--- a/packages/angular/cli/upgrade/version.ts
+++ b/packages/angular/cli/upgrade/version.ts
@@ -41,17 +41,6 @@ export class Version {
     let angularPkgJson;
     let rxjsPkgJson;
 
-    const isInside = (base: string, potential: string): boolean => {
-      const absoluteBase = path.resolve(base);
-      const absolutePotential = path.resolve(potential);
-      const relativePotential = path.relative(absoluteBase, absolutePotential);
-      if (!relativePotential.startsWith('..') && !path.isAbsolute(relativePotential)) {
-        return true;
-      }
-
-      return false;
-    };
-
     try {
       const resolveOptions = {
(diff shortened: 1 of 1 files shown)


hansl:
Hi @nicolas-geniteau ,

This change has been already covered by https://github.com/angular/angular-cli/pull/12325.

Thanks.

PR #12373, opened 2018-09-26
build: fix build

alan-agius4:


packages/schematics/angular/e2e/index.ts
```diff
@@ -125,7 +125,7 @@ export default function (options: E2eOptions): Rule {
 
     if (options.projectRoot !== undefined) {
       newProjectRoot = options.projectRoot;
-      appDir = newProjectRoot;
+      appDir = options.projectRoot;
```
hansl: Undo this change?

alan-agius4: TypeScript is complaining maybe I do `!` and disable it with tslint

PR #12379, opened 2018-09-27
style: improve formatting of spec file

### LINKED ISSUE #12161: __name@dasherize__.component.spec.ts don't use async in spec case
_From @trimeyko on September 29, 2017 13:18_

Component spec file scheme is old and not consistent with spec in new app since it's not using async() in spec case, should be used by default.

component:
https://github.com/angular/devkit/blob/master/packages/schematics/angular/component/files/__path__/__name%40dasherize%40if-flat__/__name%40dasherize__.component.spec.ts

spec in new app:
https://github.com/angular/devkit/blob/master/packages/schematics/angular/application/other-files/app.component.spec.ts

_Copied from original issue: angular/devkit#182_

alan-agius4:
style: improve formatting of spec file

  
style: remove redundant `async` in tests
In component specs only the `compileComponents` is async thus the `async` on `it` are redudnant

Closes #12161

### DIFF
--- a/packages/schematics/angular/application/other-files/app.component.spec.ts
+++ b/packages/schematics/angular/application/other-files/app.component.spec.ts
@@ -1,7 +1,7 @@
 import { TestBed, async } from '@angular/core/testing';<% if (routing) { %>
 import { RouterTestingModule } from '@angular/router/testing';<% } %>
 import { AppComponent } from './app.component';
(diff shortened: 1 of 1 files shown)


alan-agius4:
Superseded by https://github.com/angular/angular-cli/pull/12380

Issue #11111, opened 2018-06-05
bug: not generate files to the defaultProject


sandangel:
### Versions

```bash
➜ ng -v

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/


Angular CLI: 6.0.7
Node: 10.3.0
OS: linux x64
Angular: 6.0.3
... animations, common, compiler, compiler-cli, core, elements
... forms, http, language-service, platform-browser
... platform-browser-dynamic, platform-server, router
... service-worker

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.6.7
@angular-devkit/build-angular      0.6.7
@angular-devkit/build-ng-packagr   0.6.7
@angular-devkit/build-optimizer    0.6.7
@angular-devkit/core               0.6.7
@angular-devkit/schematics         0.6.7
@angular/cdk                       6.2.1
@angular/cli                       6.0.7
@angular/flex-layout               6.0.0-beta.15
@angular/material                  6.2.1
@angular/pwa                       0.7.0-beta.1
@ngtools/json-schema               1.1.0
@ngtools/webpack                   6.0.7
@schematics/angular                0.6.7
@schematics/update                 0.6.7
ng-packagr                         3.0.1
rxjs                               6.2.0
typescript                         2.7.2
webpack                            4.8.3
```

### Repro steps

```bash
ng new test-proj
ng g application test-app
// edit angular.json defaultProject -> test-app

ng g c test-comp         
CREATE src/app/test-comp/test-comp.component.css (0 bytes)
CREATE src/app/test-comp/test-comp.component.html (28 bytes)
CREATE src/app/test-comp/test-comp.component.spec.ts (643 bytes)
CREATE src/app/test-comp/test-comp.component.ts (280 bytes)
UPDATE src/app/app.module.ts (406 bytes)
```

### Observed behavior
```
<!-- Normally this includes a stack trace and some more information. -->
```
ref: https://github.com/angular/angular-cli/issues/10352 , #10327
cli always generate schematics (pwa,component...) to the project under src/ folder

### Desired behavior

[...]

PR #12394, opened 2018-09-28
Deprecate defaultProject

filipesilva:
More context in https://github.com/angular/angular-cli/issues/11111#issuecomment-425176217

Fix https://github.com/angular/angular-cli/issues/11111

### DIFF
--- a/docs/documentation/angular-workspace.md
+++ b/docs/documentation/angular-workspace.md
@@ -8,7 +8,7 @@
 
 - **newProjectRoot** (`string`): Path where new projects will be created.
 
-- **defaultProject** (`string`): Default project name used in commands.
+- **defaultProject** (`string`): (**DEPRECATED**) Default project name used in commands.
 
 - **cli**: Workspace configuration options for Angular CLI.
   - *defaultCollection* (`string`): The default schematics collection to use.
--- a/packages/angular_devkit/core/src/workspace/workspace-schema.ts
+++ b/packages/angular_devkit/core/src/workspace/workspace-schema.ts
@@ -20,6 +20,7 @@ export interface WorkspaceSchema {
    */
   newProjectRoot?: string;
   /**
+   * @deprecated From 7.0.0
    * The default project.
    */
   defaultProject?: string;
--- a/packages/angular_devkit/core/src/workspace/workspace.ts
+++ b/packages/angular_devkit/core/src/workspace/workspace.ts
@@ -140,6 +140,9 @@ export class Workspace {
     return workspaceProjectClone;
   }
 
+  /**
+   * @deprecated From 7.0.0
+   */
   getDefaultProjectName(): string | null {
     this._assertLoaded();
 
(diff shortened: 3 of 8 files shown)


packages/angular_devkit/core/src/workspace/workspace.ts
```diff
@@ -140,6 +140,7 @@ export class Workspace {
     return workspaceProjectClone;
   }
 
+  // `defaultProject` is deprecated and should be removed in CLI v8.
   getDefaultProjectName(): string | null {
```
alan-agius4: Can you use `@deprecated`? As when using it tslint is able to warn users. Same thing in the IDE

https://palantir.github.io//tslint/rules/deprecation/

packages/angular_devkit/core/src/workspace/workspace-schema.ts
```diff
@@ -20,7 +20,7 @@ export interface WorkspaceSchema {
    */
   newProjectRoot?: string;
   /**
-   * The default project.
+   * (DEPRECATED) The default project.
```
alan-agius4: Same

docs/documentation/angular-workspace.md
```diff
@@ -6,7 +6,7 @@
 
 - **version** (`integer`): File format version. This is currently `"1"`.
 
-- **newProjectRoot** (`string`): Path where new projects will be created.
+- **newProjectRoot** (`string`): (**DEPRECATED**) Path where new projects will be created.
```
alexeagle: this looks wrong, shouldn't it be on the following option?

filipesilva: Jeez, that's what I get for amending commits in a rush.

packages/angular_devkit/core/src/workspace/workspace-schema.ts
```diff
@@ -20,7 +20,8 @@ export interface WorkspaceSchema {
    */
   newProjectRoot?: string;
   /**
-   * The default project.
+   * @deprecated From 7.0.0
+   * The default project.*
```
alan-agius4: Nit: you have an extra `*` at the end

filipesilva: Thanks, fixed

alexeagle:
Blocked because we will reopen a larger discussion of project layout for ng8

filipesilva:
We need to actually design what non-project scoped commands should do. Currently default project almost never applies, and when it does it might not be the right thing anyway.

SchnWalter:
From what I can tell from https://github.com/angular/angular-cli/issues/11111#issuecomment-470635587, this PR should be closed. Although there are still inconsistencies between various commands where the project name is sometimes passed using `--projects` and other times it's `--name`. I'll look into this in the upcoming weeks and if it's clear what the default flag should be, I'll create some PRs to streamline the options.

dgp1130:
Going through old PRs. After some discussion, we still think we should deprecate `defaultProject`, but this PR is fairly old and not usable in its current state anyways. Filed https://github.com/angular/angular-cli/issues/20661 to follow up with a deprecation.

Issue #8112, opened 2017-10-19
Custom root / --source-dir path causes components to be generated in wrong folders


dragonflypl:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.

```
Angular CLI: 1.5.0-rc.1
Node: 6.9.1
OS: win32 x64
Angular: 4.4.6
... common, compiler, core, forms, platform-browser
... platform-browser-dynamic, router

@angular/animations: 4.4.5
@angular/cli: 1.5.0-rc.1
@angular/compiler-cli: 4.4.5
@angular/http: 4.4.5
@angular/language-service: 4.4.5
@angular/tsc-wrapped: 4.4.5
@angular-devkit/build-optimizer: 0.0.28
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.34
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.0-rc.1
@schematics/angular: 0.0.46
typescript: 2.3.4
webpack: 3.7.1
```

### Repro steps.

Step 1:

> ng new fooGui --prefix foo --style scss --routing true --skip-git true --source-dir src/main/client

Step 2:

> cd fooGui

Step 3:

> ng g c helloWorld --skip-import

### The log given by the failure.

Output is:

```
  create src/main/client/src/main/client/app/hello-world/hello-world.component.html (30 bytes)
  create src/main/client/src/main/client/app/hello-world/hello-world.component.spec.ts (657 bytes)
  create src/main/client/src/main/client/app/hello-world/hello-world.component.ts (289 bytes)
  create src/main/client/src/main/client/app/hello-world/hello-world.component.scss (0 bytes)
```

### Desired functionality.

IMHO, component files should be created inside: 

```
  create src/main/client/app/hello-world/hello-world.component.html (30 bytes)
  create src/main/client/app/hello-world/hello-world.component.spec.ts (657 bytes)
  create src/main/client/app/hello-world/hello-world.component.ts (289 bytes)
  create src/main/client/app/hello-world/hello-world.component.scss (0 bytes)
```

Brocco:
I was not able to reproduce this with 1.5.0-rc.5

I created the project as you described and ran:

`ng g c helloWorld --skip-import --dry-run`

and saw the following output:
```
  create src/main/client/app/hello-world/hello-world.component.scss (0 bytes)
[...]

PR #12398, opened 2018-09-28
feat(@schematics/angular): add new projectRoot flag for initial project

hansl:
We used to support it. See #8112


alexeagle:
@IgorMinar should this block on us figuring out a different project structure for 8?

PR #12400, opened 2018-09-28
Cleanup Webpack HTML plugin configuration

clydin:
Also, restores previous behavior wherein not specifying the `index` option would disable index HTML generation.


clydin:
yarn 1.10.1 adds an integrity field for each package

alan-agius4:
AppVeyor failure looks like a flake

PR #12402, opened 2018-09-28
fix(@angular-devkit/build-angular): don't rerun tests on unchanged co…

### LINKED ISSUE #11880: Tests Run When Swap Files (*.swp, *.swo, *.swn, ...) Change
```
- [ ] feature request
```

```
- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

node: v8.11.3
npm: 5.6.0
Angular CLI: 6.0.8
Ubuntu Linux: 16.04.1

### Repro steps

Run tests: ng test
Open any file with vim and the tests run: vi src/app/heroes/heroes.component.ts
Change the file without saving and the tests run.  Anything that causes a change in the vim-generated swap file triggers a test re-run.

### Desired functionality

The tests should only run once.

Seems like a similar issue has already been reported here: https://github.com/angular/angular-cli/issues/4593  Being a vim user I really appreciate the fix; however, the issue is also present when running unit tests.

filipesilva:
…mpilation

Fix #11880

### DIFF
--- a/packages/angular_devkit/build_angular/test/karma/rebuilds_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/karma/rebuilds_spec_large.ts
@@ -13,7 +13,8 @@ import { host, karmaTargetSpec } from '../utils';
 
 // Karma watch mode is currently bugged:
 // - errors print a huge stack trace
-// - karma does not have a way to close the server gracefully.
+// - karma does not have a way to close the server
+//   gracefully (https://github.com/karma-runner/karma/issues/3149)
 // TODO: fix these before 6.0 final.
 xdescribe('Karma Builder watch mode', () => {
   beforeEach(done => host.initialize().toPromise().then(done, done.fail));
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/test/karma/rebuilds_spec_large.ts
```diff
@@ -64,4 +65,9 @@ xdescribe('Karma Builder watch mode', () => {
       take(3),
     ).toPromise().then(done, done.fail);
   }, 30000);
+
+  it('does not rebuild when nothing changed', (done) => {
+    // Start the server in watch mode, wait for the first build to finish, touch
+    // test.js without changing it, wait 5s then exit unsuscribe, verify only one event was emitted.
+  }, 30000);
```
clydin: There's no code in here.  is that intentional?

filipesilva: Yeap, see the comment above. We can't actually test karma in watch mode because there's no way for the karma server to exit gracefully without killing the processes.

alexeagle: This at least needs a comment, anyone looking at this code later will be very confused why it claims to test something and then doesn't.
Did you think about another way to test this?

PR #12410, opened 2018-09-30
 build: update to TypeScript 3.1 & Angular 7 RC 0

alan-agius4:
build: update to TypeScript 3.1
feat(@ngtools/webpack): add support for `TypeScript` 3.1

Closes: https://github.com/angular/angular-cli/issues/12413


packages/angular/cli/utilities/json-schema.ts
```diff
@@ -19,16 +19,20 @@ import {
   Value,
 } from '../models/interface';
 
-function _getEnumFromValue<E, T extends string>(v: json.JsonValue, e: E, d: T): T {
-  if (typeof v !== 'string') {
-    return d;
+function _getEnumFromValue<E, T extends E[keyof E]>(
```
alan-agius4: The types here had to be changed due to TS 3.1, I updated the param names to make the more readable.

alan-agius4: The types here had to be changed due to TS 3.1, I updated the param names to make the more readable.

packages/angular/cli/utilities/json-schema.ts
```diff
@@ -19,16 +19,20 @@ import {
   Value,
 } from '../models/interface';
 
-function _getEnumFromValue<E, T extends string>(v: json.JsonValue, e: E, d: T): T {
-  if (typeof v !== 'string') {
-    return d;
+function _getEnumFromValue<E, T = E[keyof E]>(
```
alan-agius4: The types here had to be changed due to TS 3.1, I updated the param names to make the more readable.

noelmace:
Oh snap I didn't saw this :sweat: Guess #12415 and #12416 can be closed then.

packages/angular/cli/utilities/json-schema.ts
```diff
@@ -19,16 +19,20 @@ import {
...
+  defaultValue: T,
+): T {
+  if (typeof value !== 'string') {
+    return defaultValue;
   }
 
-  if (Object.values(e).indexOf(v) !== -1) {
-    return v as T;
+  if (Object.values(enumeration).indexOf(value) !== -1) {
+    return value as unknown as T;
```
alexeagle: is this really the long-term fix? if not needs a TODO

tests/legacy-cli/e2e/tests/misc/typescript-warning.ts
```diff
@@ -3,9 +3,13 @@ import { updateJsonFile } from '../../utils/project';
 
 export default async function () {
   // typescript@2.8.0-dev.20180320 is not part of the officially supported range in latest stable.
-  let unsupportedTsVersion = '2.8.0-dev.20180320';
+  const unsupportedTsVersion = '2.8.0-dev.20180320';
 
   await updateJsonFile('src/tsconfig.app.json', configJson => {
+    configJson.compilerOptions = {
+      ...configJson.compilerOptions,
+      skipLibCheck: true,
```
alexeagle: does this have any bearing on whether users would also need to skip type-checks of `.d.ts` files?

packages/angular/cli/utilities/json-schema.ts
alan-agius4: This is pretty permanent as type `T` doesn't provide enough type sufficiently overlaps for type `string`

tests/legacy-cli/e2e/tests/misc/typescript-warning.ts
alan-agius4: They will still get the `ERROR in The Angular Compiler requires TypeScript >=3.1.1 and <3.2.0 but 3.0.3 was found instead.` but also a error from `tsc` because of some typings in Angular Core are not compatible with this version of TypeScript.

It shouldn't be turned on for users, as in reality they are being informed that they are using an unsupported version.

PR #12411, opened 2018-09-30
Fix cpu profiling

noelmace:
As v8-profiler doesn't support node 8 and 10, it was impossible to capture a CPU profile using the NG_CLI_PROFILING environment variable.

Replace this dependency by v8-profiler-node8 and add related documentation.

FYI: as some cli e2e tests where already broken on master, I couldn't run them all.

### DIFF
--- a/packages/angular/cli/lib/init.ts
+++ b/packages/angular/cli/lib/init.ts
@@ -16,6 +16,8 @@ import { SemVer } from 'semver';
 import { Duplex } from 'stream';
 import { isWarningEnabled } from '../utilities/config';
 
+import { Profiler } from 'v8-profiler';
+
 const packageJson = require('../package.json');
 
 function _fromPackageJson(cwd?: string) {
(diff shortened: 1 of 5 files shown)


packages/angular/cli/lib/init.ts
```diff
@@ -16,6 +16,9 @@ import { SemVer } from 'semver';
 import { Duplex } from 'stream';
 import { isWarningEnabled } from '../utilities/config';
 
+// from @types/v8-profiler (no dependency to v8-profiler as we use v8-profiler-node8)
+import { Profiler } from 'v8-profiler'; // tslint:disable-line:no-implicit-dependencies
```
hansl: Sorry, this should not be there. We don't want to depend v8-profiler, just use it if it's asked for. Maybe a dynamic import would be better?

noelmace: I think this only depends on @types/v8-profiler, but I may be wrong. I struggled a little with how to respect both no-any and no-implicit-any here. Should I simply add a disable on no-any or define a simple interface for that ?

noelmace: Replaced by a minimal Profiler type inspired by @types/v8-profiler.

PR #12415, opened 2018-09-30
feat(@schematics/angular): update Angular and TS versions

noelmace:
The CLI legacy e2e tests where failing because Angular 7.0.0-rc.0 embed
https://github.com/angular/angular/pull/26151/, and TS version 3.0 wasn't
right for this. This commit fix this error as reported in #12423, but isn't enought
to support TS 3.1 (see #12416 for more commits on this).

Close #12413

PR #12416, opened 2018-09-30
Upgrade ts 3.1

noelmace:
Use TS 3.1 as Angular now requires it (see https://github.com/angular/angular/pull/26151).

You can find a more simple solution to this in #12415. :warning: this PR embed the commit from this other PR.


alexeagle:
Thanks for chipping in - as you saw we already started the update (we hoped to release on Friday along with the framework but ran out of time.)

PR #12419, opened 2018-10-01
Type checker replacements

filipesilva:


packages/angular_devkit/build_angular/test/browser/replacements_spec_large.ts
```diff
@@ -154,4 +155,53 @@ describe('Browser Builder file replacements', () => {
...
+      }],
+      watch: true,
+    };
+
+    const unexpectedError = `Property 'two' does not exist on type '{ one: number; }'`;
+    const expectedError = `Property 'prop' does not exist on type '{}'`;
+    const logger = new TestLogger('rebuild-type-errors');
+
+    // Race between a timeout and the expected log entry.
+    const stop$ = race<null | string>(
```
alan-agius4: Wouldn't the below result in the same test?

```ts
    let errorAdded = false;
    runTargetSpec(host, browserTargetSpec, overrides, DefaultTimeout, logger)
      .pipe(
        tap((buildEvent) => expect(buildEvent.success).toBe(true, 'build should succeed')),
        tap(() => {
          // Introduce a known type error to detect in the logger filter.
          if (!errorAdded) {
            host.appendToFile('src/main.ts', 'console.log({}.prop);');
            errorAdded = true;
          } else {
            expect(logger).toContain(expectedError);
          }
        }),
        tap(() => {
          expect(logger).not.toContain(unexpectedError);
          logger.clear();
        }),
        take(2),
      ).toPromise();
```

filipesilva: No, because the type checker logs are not synchronized with builds. Since the type checker runs separately, it will finish rebuilds separately from the main webpack process. 

In this specific case, the type checker rebuild will take longer. We need to start a build and and rebuild on the main compilation process, then wait for the forked process to send logs messages.

alan-agius4: Ah cause it runs in a forked process. 👍

alan-agius4:
```
Converting circular structure to JSON
    at JSON.stringify (<anonymous>)
    at new Server (/tmp/angular-cli-e2e-11892-90-1lxtz35.zgki/test-project/node_modules/karma/lib/server.js:65:41)
    at Observable.rxjs_1.Observable.obs [as _subscribe] (/tmp/angular-cli-e2e-11892-90-1lxtz35.zgki/test-project/node_modules/@angular-devkit/build-angular/src/karma/packages/angular_devkit/build_angular/src/karma/index.ts:102:29)
 
```

Looks like you shouldn't pass the `logger` to `karma`.

filipesilva:
@alan-agius4 I think it's not the logger since karma doesn't import it directly, but rather the circular dependency between the type checker and compiler plugin introduced by the new message. I'll refactor the messages into a separate file.

filipesilva:
@alan-agius4 ran some debugging on the circular object reported by Karma and indeed it's the logger property inside the `AngularCompilerPlugin` instance.

@clydin that's probably the best approach, yeah. Unsure of the limitation on Node's child process IPC channel used here for logging (https://nodejs.org/api/child_process.html). I think it's very likely that the main thread is blocked transferring large files. Doing that properly is essentially a virtual FS server, something that I discussed with @hansl in the past. Not transferring unchanged files is very important in that scenario too.

PR #12457, opened 2018-10-03
 feat(jsonpFunction): allow overriding jsonpFunction for multiple webpack builds on the same pag

kylecannon:
This PR will allow for many code split builds (ex: many angular elements builds) to live on the same page and work correctly without clashing the default webpackJsonp global.


mgechev:
Let us not leak the webpack API further. I'd recommend you to use `ngx-build-plus` for your use case, so you can extend the build.

Before opening a feature PR, I'd recommend you to first start a discussion in an issue so that we can decide if the feature is in the scope of the CLI or not. Thank you again for the contribution :)

PR #12473, opened 2018-10-04
ci: exit with `0` error code when running templates

alan-agius4:
@hansl not sure why this is run in CI as changing the templates during CI won't commit it. I Think ideally it should validate if the contents is the same if if they are not the same it should exit with 1 other wise 0

### DIFF
--- a/scripts/templates.ts
+++ b/scripts/templates.ts
@@ -31,5 +31,5 @@ export default async function(_options: {}, logger: logging.Logger): Promise<num
   _runTemplate('./templates/readme', '../README.md', logger);
   _runTemplate('./templates/contributing', '../CONTRIBUTING.md', logger);
 
-  return 1;
+  return 0;
 }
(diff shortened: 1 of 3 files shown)


scripts/templates.ts
```diff
@@ -24,12 +24,22 @@ function _runTemplate(inputPath: string, outputPath: string, logger: logging.Log
     encode: (x: string) => global.encodeURIComponent(x),
     require: (x: string) => require(path.resolve(path.dirname(inputPath), x)),
   });
+
+  const fileExists = fs.existsSync(outputPath);
```
hansl: This is done in CI using `git` to do the check (check `validate-templates`). The user should not have to delete files before overwriting it.

This change is really inconvenient for the user. This script is not a validation, it's just generating the templates.

hansl: Actually the validation is in `scripts/validate.ts`.

PR #12498, opened 2018-10-05
Add `--minimal` flag to enable barebones project.

### LINKED ISSUE #11904: `--minimal` flag does nothing
We should get styleguide guidance for what the `--minimal` flag should act on 6.x / 7.x. Or just implement it the way it was before.

rsarky:
Fixes #11904

### DIFF
--- a/packages/schematics/angular/application/schema.json
+++ b/packages/schematics/angular/application/schema.json
@@ -66,6 +66,11 @@
       "type": "boolean",
       "default": false,
       "description": "Do not add dependencies to package.json."
+    },
+    "minimal": {
+      "description": "Create a barebones project without any testing frameworks",
+      "type": "boolean",
+      "default": false
     }
   },
   "required": [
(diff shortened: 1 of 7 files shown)


packages/schematics/angular/application/schema.json
```diff
@@ -66,6 +66,11 @@
       "type": "boolean",
       "default": false,
       "description": "Do not add dependencies to package.json."
+    },
+    "minimal": {
+      "description": "Create a barebones project without any testing frameworks",
```
alan-agius4: Barebones > barebone

packages/schematics/angular/workspace/schema.json
```diff
@@ -73,6 +73,11 @@
       "$default": {
         "$source": "ng-cli-version"
       }
+    },
+    "minimal": {
+      "description": "Create a barebones project without any testing frameworks",
```
alan-agius4: Barebones > barebone

PR #12499, opened 2018-10-06
Add documented public api

### LINKED ISSUE #8744: Support mutation testing with angular cli
We're investigating the possibility of supporting angular in [Stryker, a mutation testing framework](https://stryker-mutator.github.io). I was wondering if it was possible to run test with `ng test --watch=false --single-run=false` and keep open the karma server so we can fire `karma run` commands.

The idea would be that we make small code changes and rerun the tests with `karma run`. Ideally we would be able to distinguish between a failed test and a compile error.

nicojs:
Fixes #8744 

@filipesilva here it is. I tried to follow the guidance in your latest remark in #8744. Could you take a look.

Is type information shipped with the angular cli? Would that be possible? Having type information for the programmatic use case would be nice.

### DIFF
--- a/packages/angular/cli/README.md
+++ b/packages/angular/cli/README.md
@@ -37,6 +37,7 @@ with NPM 5.5.1 or higher.
 * [Updating Angular CLI](#updating-angular-cli)
 * [Development Hints for working on Angular CLI](#development-hints-for-working-on-angular-cli)
 * [Documentation](#documentation)
+* [Programmatic usage](#programmatic-usage)
 * [License](#license)
 
 ## Installation
@@ -251,6 +252,18 @@ In addition to this one, another, more elaborated way to capture a CPU profile u
 ## Documentation
 
 The documentation for the Angular CLI is located in this repo's [wiki](https://github.com/angular/angular-cli/wiki).
+
+## Programmatic usage
+
+You can also use angular cli programmatically. Here is an example on how to run `ng test --watch false` from inside your nodejs application:
+
+```js
+import cli from '@angular/cli';
+
+const returnCode = await cli({ args: ['test', '--watch', 'false'] });
+```
+
+Types for typescript are included.
 
 ## License
 
--- a/packages/angular/cli/lib/cli/main.ts
+++ b/packages/angular/cli/lib/cli/main.ts
@@ -12,8 +12,24 @@ import { runCommand } from '../../models/command-runner';
 import { getWorkspaceRaw } from '../../utilities/config';
 import { getWorkspaceDetails } from '../../utilities/project';
 
+export interface CliOptions {
+  testing?: boolean;
+  /**
+   * The cli arguments to provide to the command (without the `ng`)
+   */
+  args: string[];
+}
 
-export default async function(options: { testing?: boolean, cliArgs: string[] }) {
+/**
+ * Run an arbitrary `ng` command.
+ *
+ * @example
+ * const returnCode = await cli({ args: ['test', '--watch', 'false' ] } );
+ *
+ * @param options The cli options, including the arguments use to run
+ * @returns {Promise<number>} A promise that resolves to the exit code of the ng command.
+ */
+export default async function(options: CliOptions): Promise<number> {
   const logger = new logging.IndentLogger('cling');
   let loggingSubscription;
   if (!options.testing) {
@@ -34,7 +50,7 @@ export default async function(options: { testing?: boolean, cliArgs: string[] })
   }
 
   try {
-    const maybeExitCode = await runCommand(options.cliArgs, logger, projectDetails);
+    const maybeExitCode = await runCommand(options.args, logger, projectDetails);
     if (typeof maybeExitCode === 'number') {
       console.assert(Number.isInteger(maybeExitCode));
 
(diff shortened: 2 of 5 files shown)


packages/angular/cli/lib/main.ts
```diff
@@ -0,0 +1,2 @@
+
```
filipesilva: `cli/lib/index.ts` can be renamed to `main.ts` instead, no need for an extra file.

packages/angular/cli/lib/cli/index.ts
```diff
@@ -12,8 +12,24 @@ import { runCommand } from '../../models/command-runner';
 import { getWorkspaceRaw } from '../../utilities/config';
 import { getWorkspaceDetails } from '../../utilities/project';
 
+export interface CliOptions {
+  testing?: boolean;
+  /**
+   * The cli arguments to provide to the command (without the `ng`)
+   */
+  cliArgs: string[];
```
filipesilva: Can you change `cliArgs` to just `args`? Being a public api the name is now more important.

packages/angular/cli/lib/main.ts
nicojs: Ok, but than it will be exported with a default export. That is OK?

nicojs:
@filipesilva 

I've implemented your review comments. The api change is back (default export), should be green now.

packages/angular/cli/lib/main.ts
filipesilva: I think the default export is ok, yes.

filipesilva:
I'm seeing red CI, I think it's because of this error:
```
TS2307: Cannot find module '@angular/cli/lib/cli'.

10 import cli from '@angular/cli/lib/cli';
````

Likely due to the filename change.

We also validade commit messages (https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#-commit-message-guidelines). A think a good single commit name would be `feat(@angular/cli): add public programmatic API`. 

In regards to when you can expect it, it would be 7.1 at the earliest. We are in a RC period already which means no no features are merged. But 7.1 would be shortly after 7.0.

**Comment by nicojs** [comment]
Both issues should be fixed now.

> In regards to when you can expect it, it would be 7.1 at the earliest. We are in a RC period already which means no no features are merged. But 7.1 would be shortly after 7.0.

That's fine. The API was already there in hindsight. I didn't realize that. This PR is really only documenting the hidden feature.

**Comment by nicojs** [comment]
@filipesilva any idea what's going wrong with the build? Sorry for the mess. 🗑

**Comment by filipesilva** [comment]
I think one of our build error reporting functions isn't reporting the error correctly. 

But it is in https://circleci.com/gh/angular/angular-cli/21583:
```
TS2345: Argument of type '{ cliArgs: string[]; }' is not assignable to parameter of type 'CliOptions'.
    Object literal may only specify known properties, and 'cliArgs' does not exist in type 'CliOptions'.
```

So somewhere the `cliArgs` is still being passed. You should be able to see this locally by running `npm run build`. It will not exit with an error code but should show it.

nicojs:
I seem to have a problem with installing the dependencies: 
image

I'll try negotiating with the build server 🙈

nicojs:
This seems to be better.

nicojs:
Great, thanks. This makes for a stable Stryker integration :+1:

filipesilva:
Thinking about the whole `args` vs `cliArgs` issue makes me think we should keep `cliArgs`... As you said, the API had exposed for a while now, it just wasn't documented. That means some people might be relying on it.

WDYT of changing it back to `cliArgs`, keep the internal rename changes, and renaming the commit to `docs(@angular/cli): document public programmatic API`? 

That better reflects the work that ended up being done in this PR, and doesn't require a minor. We could get it in in 7.0.

nicojs:
Hmm good point. Didn't think of that, this will break Stryker as well...

However I do agree that `args` might be better. We could also deprecate `cliArgs`, but make sure it still works. If both `args` and `cliArgs` are used, `cliArgs` are ignored.

However, changing it back to `cliArgs` is fine by me to.

filipesilva:
That's a ways to go about it. I'll bring it up with the team and see what they prefer as well.

nicojs:
If you're worried about backward compatibility, we could also leave the name `index.ts` instead of `main.ts`, or keep an index.ts with a deprecated message and proxy to main.ts

filipesilva:
That part doesn't bother me much, as it's not part of the public API per se. Importing `@angular/cli` exposes the API, but importing `@angular/cli/lib/index` is a deep import and not part of the API.

filipesilva:
@nicojs we discussed this within the team and decided to not document this API right now. Documenting it implies stability, and we can't commit to this API being stable yet. One of the changes we're planning is to separate local and global CLI and that will most likely affect his API.

I think not having this PR merged won't affect your usecase because you are already using the existing API and it hasn't changed. But let me know if it does affect you.

nicojs:
> Documenting it implies stability, and we can't commit to this API being stable yet. 

Well.. strictly speaking it would force you to not change the API with a patch/minor release, which helps Stryker and other tools (intellij plugin, etc).

We are indeed not effected right now. However, could you try to do the following:

* Don't break this undocumented API with a patch/minor release
* When you break it with a major release, add it to the changelog (even though it is undocumented). Preferably with an example of the new API

That would be great.

filipesilva:
If/when we change it, I expect it to be well documented. I can't 100% guarantee it will be in a major, but can say that we avoid changing that sort of things in patch/minor especially now that we're all aware there are at least a couple of tools that are using it.

PR #12500, opened 2018-10-06
refactor(@angular/schematics): Add an extra option to not use dasheri…

vikerman:
…ze in findModuleFromOptions

In Google we prefer underscore instead of dashes. So this option is useful there.

### DIFF
--- a/packages/schematics/angular/utility/find-module.ts
+++ b/packages/schematics/angular/utility/find-module.ts
@@ -25,6 +25,7 @@ export interface ModuleOptions {
   skipImport?: boolean;
   moduleExt?: string;
   routingModuleExt?: string;
+  nameFormatter?: (str: string) => string;
 }
 
 const MODULE_EXT = '.module.ts';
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/utility/find-module.ts
```diff
@@ -42,8 +43,9 @@ export function findModuleFromOptions(host: Tree, options: ModuleOptions): Path
   const routingModuleExt = options.routingModuleExt || ROUTING_MODULE_EXT;
 
   if (!options.module) {
+    options.normalizeName = options.normalizeName || strings.dasherize;
```
hansl: Instead of allowing for overwriting the name, why don't you let the parent pass in the formatting function (with dasherize being the default)? We have [`strings.underscore()`](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/core/src/utils/strings.ts#L106) which would allow you to format the name properly in google, whichever the input. What do you think?

vikerman: Ugh Sorry. I think the variable is badly named. This is exactly what I'm trying to do.

  normalizeName?: (str: string) => string;

Renamed it to nameFormatter.

PR #12502, opened 2018-10-06
fix(@angular/cli): Filter duplicate help statements

### LINKED ISSUE #12492: `--help` Outputs duplicate flags.
```
- [ ] feature request
```

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

- [x] --help
```

### Versions

`node v10.11.0`
`npm v6.4.1`
`ng v6.2.3`
`Linux x64 with Ubuntu`

### Repro steps

`ng new --help`
or
`ng update --help`

Example:
```
~❯ ng new --help
Creates a new directory and a new Angular app.
usage: ng new <name> [options]
arguments:
  name
    The name of the workspace.
options:
  --collection (-c)
    Schematics collection to use.
  --collection (-c)
    Schematics collection to use.
  --directory 
    The directory name to create the workspace in.
  --dryRun (-d)
    Run through without making any changes.
  --dryRun (-d)
    Run through without making any changes.
  --experimental-ivy 
    EXPERIMENTAL: Specifies whether to create a new application which uses the Ivy rendering engine.
  --force (-f)
    Forces overwriting of files.
  --force (-f)
    Forces overwriting of files.
  --inline-style (-s)
    Specifies if the style will be in the ts file.
  --inline-template (-t)
    Specifies if the template will be in the ts file.
  --new-project-root 
    The path where new projects will be created.
  --prefix (-p)
    The prefix to apply to generated selectors.
  --routing 
    Generates a routing module.
  --skip-git (-g)
    Skip initializing a git repository.
  --skip-install 
    Skip installing dependency packages.
  --skip-tests (-S)
    Skip creating spec files.
  --style 
    The file extension to be used for style files.
  --verbose (-v)
    Adds more details to output logging.
  --verbose (-v)
    Adds more details to output logging.
  --view-encapsulation 
    Specifies the view encapsulation strategy.
```
### Desired functionality

Remove the duplicate help statements.

rsarky:
Fixes #12492

### DIFF
--- a/packages/angular/cli/utilities/json-schema.ts
+++ b/packages/angular/cli/utilities/json-schema.ts
@@ -155,6 +155,12 @@ export async function parseJsonSchemaToOptions(
     const ptr = json.schema.parseJsonPointer(pointer);
     const name = ptr[ptr.length - 1];
 
+    if (options.some((e) => {
+      return e.name === name;
+    })) {
+      // Skip duplicates.
+      return;
+    }
     if (ptr[ptr.length - 2] != 'properties') {
       // Skip any non-property items.
       return;


packages/angular/cli/models/command.ts
```diff
@@ -101,6 +101,9 @@ export abstract class Command<T extends BaseCommandOptions = BaseCommandOptions>
       this.logger.info(`options:`);
       opts
         .filter(o => !o.hidden)
+        .filter((val, index, self) => {
```
alan-agius4: Ideally we fix the schemas rather than filtering the out what’s duplicate.

rsarky: Hey. Thanks a lot for reviewing. 
I couldn't figure out what in the schema is causing the duplicate output

hansl: I actually think the proper fix might be in the [parseJsonSchemaToOptions](https://github.com/angular/angular-cli/blob/master/packages/angular/cli/utilities/json-schema.ts#L158) function. Having multiple times the same property isn't an error in JSON Schema world (as long as those properties match each others), but for the options you should only have one. So we should skips options we already found.

At that line, add a check for `if (options.some(o => o.name === name) { return; }` and you should be good. That way you fix the source of the problem and not the symptoms :)

rsarky: Understood.
I have updated the PR.

rsarky:
Closing this as it is already handled in the latest release

PR #12520, opened 2018-10-08
Update doc strings for command reference

jbogarthyde:
This is an initial draft transferring content from google doc draft to JSON sources. 
Basic command descriptions and generation schematic options included.
This is to meant to test the git and cli-doc build process as well.

### DIFF
--- a/packages/angular/cli/commands/doc.json
+++ b/packages/angular/cli/commands/doc.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/schema",
   "$id": "ng-cli://commands/doc.json",
-  "description": "Opens the official Angular API documentation for a given keyword.",
+  "description": "Opens the official Angular documentation (angular.io) in a browser, and searches for a given keyword.",
   "$longDescription": "",
 
   "$aliases": [ "d" ],
@@ -14,7 +14,7 @@
       "properties": {
         "keyword": {
           "type": "string",
-          "description": "The query to search upon.",
+          "description": "The keyword to search for, as provided in the search bar in angular.io.",
           "$default": {
             "$source": "argv",
             "index": 0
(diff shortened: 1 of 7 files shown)


packages/angular/cli/commands/doc.json
```diff
@@ -1,8 +1,8 @@
 {
   "$schema": "http://json-schema.org/schema",
   "$id": "ng-cli://commands/doc.json",
-  "description": "Opens the official Angular API documentation for a given keyword.",
-  "$longDescription": "",
+  "description": "Opens the Angular documentation in a browser, with a search for a given term.",
+  "$longDescription": "./doc.md,
```
clydin: missing end quote

packages/angular/cli/commands/generate.json
```diff
@@ -1,8 +1,8 @@
 {
   "$schema": "http://json-schema.org/schema",
   "$id": "ng-cli://commands/generate.json",
-  "description": "Generates and/or modifies files based on a schematic.",
-  "$longDescription": "",
+  "description": "Creates, initializes, and modifies files to create a generic version of an app or app element in the given or default project, according to the given schematic.",
+  "$longDescription": "Takes the name of the project, as specified in the \"projects\" section of the angular.json workspace configuration file. \n When a project name is not supplied, the configured default project of the workspace is used.",
```
clydin: The generate command actually doesn't take the name of a project in the same way as the other commands.  It is currently an outlier and we've been discussing the best path forward to potentially remedy the situation.

clydin: `@longDescription` also needs to point to a separate file.

packages/angular/cli/commands/doc.json
jbogarthyde: moved long desc back into short desc

packages/angular/cli/commands/generate.json
jbogarthyde: No long description now. 
I'm also trying to formulate better boilerplate for the ones that do take an optional project. Maybe we can add it to definitions.json

PR #12526, opened 2018-10-09
refactor(@schematics/update): cleanup/optimize npm fetch

clydin:
massive code reduction as well as built-in caching
broader support for `.npmrc` options


alexeagle:
@hansl do you know whether this validation check is intended just for google3 import? This licensing violation is for a package we don't use in google3.

Airblader:
Upstream has been resolved. Just wanted to notify since removing the blocked label alone probably didn't do that.

PR #12527, opened 2018-10-09
Enable the ability to configure WebpackDevServer.Configuration.headers

vadim-shb:
Enable the ability to configure WebpackDevServer.Configuration.headers from angular.json.

I, personally, need it for easier CORS configuring of my angular-ionic application with third-party API.
I believe there are another configuration possibilities with this configuration available.


clydin:
Closing due to inactivity.

PR #12547, opened 2018-10-10
fix(@angulat-devkit/build-angular): dont serve files on cwd

filipesilva:


filipesilva:
Yeah I thought we had removed this a while ago. I noticed it was still happening while debugging another issue.

PR #12554, opened 2018-10-11
fix(@angular/cli): error out when command json is invalid

filipesilva:
Related to https://github.com/angular/angular/pull/26391


packages/angular/cli/commands/e2e.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "http://json-schema.org/schema",
   "$id": "ng-cli://commands/e2e.json",
   "description": "Builds and serves an Angular app, and runs end-to-end tests using Protractor.",
-  "$longDescription": "Must be executed from within a workspace directory.\n When a project name is not supplied, the configured default e2e project of the workspace is used.",
+  "$longDescription": "./e2e-long.md",
```
petebacondarwin: So was this the error that caused the output to be invalid?

filipesilva: Yes, that was the one. That should be a file path. We output a warning instead of failing before. 

The warning would actually happen on every `ng` command, not just the help ones.

scripts/snapshots.ts
```diff
@@ -117,9 +117,9 @@ export default async function(opts: SnapshotsOptions, logger: logging.Logger) {
     const options = { cwd: newProjectRoot };
     const childLogger = logger.createChild(commandName);
     const stdout = _exec(ngPath, [commandName, '--help=json'], options, childLogger);
-    if (stdout.trim()) {
-      fs.writeFileSync(path.join(helpOutputRoot, commandName + '.json'), stdout);
-    }
+    // Make sure the output is JSON before printing it, and format it as well.
+    const jsonOutput = JSON.stringify(JSON.parse(stdout.trim()), undefined, 2);
```
petebacondarwin: 😻

scripts/snapshots.ts
```diff
@@ -117,9 +117,9 @@ export default async function(opts: SnapshotsOptions, logger: logging.Logger) {
...
+    fs.writeFileSync(path.join(helpOutputRoot, commandName + '.json'), jsonOutput);
```
filipesilva: This looks like the kind of thing that should be part of build, instead of snapshot. 

@hansl @clydin is there a reason it's not there?

hansl: Build does not generate the json files. We don’t publish those files to npm.

PR #12558, opened 2018-10-11
fix(@angular/cli): use stderr for all diagnostic related log messages

clydin:


clydin:
7.0 would make sense then.  But I highly doubt anyone is relying on warnings being present on stdout and completely ignoring stderr (where log.error and log.fatal are present).  Generally if you want everything you would already be redirected stderr into stdout (`2>&1`).

PR #12559, opened 2018-10-11
fix(@ngtools/webpack): report a warning when lazy route discovery is disabled

### LINKED ISSUE #12025: lazy loading chunks not built if bootstrap is not in main.ts file
Currently if people want to use lazy loading within Angular CLI projects and their `bootstrap` code is not placed inside of the `$PROJECT_ROOT/main.ts` file, the chunks for the lazy-loading module are not being generated.

This leads to JIT runtime errors that indicate that the lazy module couldn't be found.

```ts
core.js:1673 ERROR Error: Uncaught (in promise): Error: Cannot find module 'lazy/slider.module'
Error: Cannot find module 'lazy/slider.module'
    at $_lazy_route_resource lazy namespace object:5
```

Moving the `boostrap` call to the `main.ts` file fixes this issue (credits to @filipesilva for this finding)

--

It would be good to have some warning or note that implicates that lazy modules won't be detected if the `bootstrap` call is placed somewhere else.

Issue demonstration: https://github.com/DevVersion/lazy-load-test-cli

cc. @filipesilva

### LINKED ISSUE #12238: Angular compiler plugin not reporting some cases of not finding an entry module
The three lines starting at

https://github.com/angular/angular-cli/blob/b403aa69c107e2d0a98c775e7b2efd946c30317a/packages/ngtools/webpack/src/entry_resolver.ts#L155

are bailing out without throwing the error at the end of `resolveEntryModuleFromMain`, which causes the build to proceed silently but generate bundles that actually fail at runtime. This should probably throw the same error.

alan-agius4:
At the moment this will cause a runtime error instead. With this change the error is thrown during the build.

Closes #12238 and Closes #12025

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/no-entry-module_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/no-entry-module_spec_large.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { runTargetSpec } from '@angular-devkit/architect/testing';
+import { TestLogger, runTargetSpec } from '@angular-devkit/architect/testing';
 import { tap } from 'rxjs/operators';
 import { browserTargetSpec, host } from '../utils';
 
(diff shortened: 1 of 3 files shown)


alan-agius4:
Umm seems like this breaks some current stuff, and it looks that we don't always have a bootstrap code.

PR #12561, opened 2018-10-11
docs(@angular/cli): correct e2e long description field

clydin:


filipesilva:
Related to https://github.com/angular/angular-cli/pull/12554.

clydin:
There's actually two of each format in there. so I picked the short form.

filipesilva:
Oh you're right, I didn't notice the `add.md`. Nevermind my request then.

clydin:
Too late :).  It makes more sense anyway since there could also be `e2e-usage.md` for usage notes.

PR #12564, opened 2018-10-11
feat(@schematics/update) import user-agent from NPM config

conspicuousClockwork:
Configuration of the user-agent header in requests is a commonly available feature; previous versions of angular-cli do not pull this value from NPM.


conspicuousClockwork:
@clydin any thoughts? A few other projects are particular about how and if their user-agent headers can be customized, wanted to know if there's already been a precedent or discussion about this.

PR #12571, opened 2018-10-12
fix(@ngtools/webpack): remove usage of private compiler API

clydin:


packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -890,31 +859,27 @@ export class AngularCompilerPlugin {
...
-        this._processLazyRoutes(this._getLazyRoutesFromNgtools());
-      } else {
-        const changedTsFiles = this._getChangedTsFiles();
-        if (changedTsFiles.length > 0) {
-          this._processLazyRoutes(this._findLazyRoutesInAst(changedTsFiles));
-        }
-      }
-      if (this._options.additionalLazyModules) {
-        this._processLazyRoutes(this._options.additionalLazyModules);
+    // Try to find lazy routes if we have an entry module.
```
alan-agius4: The `entryModule` check is still required as without it a number of lazy loading tests fails. Mainly the once that are using import or `System.import` as it will fail with an error

Cannot read property 'isComponent' of undefined

clydin: It's only required with JIT mode actually.  The application metadata is only fully analyzed with AOT.  The `listLazyRoutes` method needs a starting point in JIT as a result since it doesn't have any other information to use.

PR #12575, opened 2018-10-12
feat(@ngtools/webpack): Automatically bundle module workers

developit:
This pull request adds support for automatic bundling of module Workers using [worker-plugin](https://github.com/googlechromelabs/worker-plugin).

In essence, it teaches Angular CLI to compile this standard browser syntax:

```js
const worker = new Worker('./some/path', { type: 'module' });
```

... into this optimized output, bundling the worker code as if it were a split point:

```js
const worker = new Worker('0.worker.12345.js');
```

/cc @robwormald


IgorMinar:
I've looked into this briefly, and in order to get this in, we'll need at least the following:

- [ ] support for IE11 (no `type: module`)
- [ ] ensure that I can use typescript in the worker modules and that these modules are built as a separate compilation unit with separate tsconfig-worker.json 
- [ ] an e2e tests in the CLI repo (with worker code written in typescript, one version with `type: "module"`, and one without)
- [ ] brief documentation on angular.io (location tbd)
- [ ] bike-shed `autoBundleWorkerModules` name, it doesn't seem to describe the feature quite accurately. how about `buildWorkerEntryPoints`?
- [ ] confirm that we want enable this by default (I think so, as long as it's possible to turn it off)

developit:
Hi @IgorMinar - just clarifying:

- The output does not use `{ type:module }`. It is compiled to a "classic" Worker. The module type is simply used as a compilation hint and to align the semantics of modules being availble within the worker context due to Webpack being run there.
- I've used this in TypeScript projects and it works - configuration is inherited from the parent compilation.  I'm not sure about typescript-worker.json support.
- I'm ambivalent on the name, the current one came from @robwormald

IgorMinar:
additional input into the design discussion: https://www.npmjs.com/package/worker-loader

that loader seems too magical and webpack specific, I prefer @developit's approach more.

alexeagle:
@developit I don't want this to get lost - but Igor has a checklist above of things we need to be able to merge this. Do you have time to do some or all of those things? Otherwise we need to find someone else to get it over the finish line.

LanceEa:
@developit @alexeagle - this would be a cool feature and would love to see it land. Any plan to move forward with this or any help needed? I can commit some time to help out wherever needed to get it over the finish line.

mgechev:
@LanceEa, we're working on landing this.

daporro:
why was this closed?

alexzuza:
@daporro I think it was closed in favor of https://github.com/angular/angular-cli/pull/13700

PR #12580, opened 2018-10-13
Update Autoprefixer and PostCSS

ai:
Hi from PostCSS community. I updated Autoprefixer, PostCSS and PostCSS loader/imports.

Autoprefixer 9 and PostCSS 7 didn’t have any breaking changes. Just drop Node.js 4 support.

@vikerman this update will bring to Angular users many fixes and reduce meanless issue reports in Autoprefixer repo (like [this](https://github.com/postcss/autoprefixer/issues/1137))


ai:
@vikerman do you know the reason for this issue in CI?

```
The package "@ngtools/webpack" does not have a BUILD file associated to it. You must either set an exception or make sure it can be built using Bazel.
```

ai:
@alan-agius4 thanks. Fixed.

alan-agius4:
@ai, the commit is still invalid.

Please use: 
`feat(@angular-devkit/build-angular): update Autoprefixer and PostCSS`

Thanks

ai:
Could it be 7.0.1? There is no API changes, only bugfixes.

ai:
Also, we need a way to prevent it in the feature. Autoprefixer 9 was released from month ago. Maybe we should add checklist “what to do before release” and add “Update dependencies” in that list? (I do it before releasing any of my project)

Issue #9202, opened 2018-01-14
Adding app-shell breaks down lazy loading on server-side


krivochenko:
### Versions
```
Angular CLI: 1.6.4
Node: 8.9.1
OS: win32 x64
Angular: 5.2.0
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, platform-server, router
... service-worker

@angular/cli: 1.6.4
@angular-devkit/build-optimizer: 0.0.38
@angular-devkit/core: 0.0.25
@angular-devkit/schematics: 0.0.48
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.4
@schematics/angular: 0.1.13
@schematics/schematics: 0.0.13
typescript: 2.5.3
webpack: 3.10.0
```

### Repro steps
* git clone git@github.com:krivochenko/lazy.git
* npm install
* build:ssr
* serve:ssr - on this step everything works fine. Including server-side rendering with lazy loading. There isn't errors  in console of server.
* ng generate app-shell --universal-app universal
* build:ssr
* serve:ssr

### Observed behavior
Server-side rendering doesn't work (in source code of page I see only router-outlet). In console of server I get error: 
```
Cannot find module 'app/pages/contacts/contacts.module.ngfactory'
```

### Desired behavior
I expect server-side rendering work without errors.

### Mention any other details that might be useful (optional)
I created simple repo for reproduction of issue by following steps:
* ng new lazy --service-worker --routing
* adding HomeComponent and ContactsComponent for checking lazy loading
* adding universal followed with [guide](https://github.com/angular/angular-cli/wiki/stories-universal-rendering)

Thanks for help

krivochenko:
As I understand, using App Shell have a meaning only if I don't use server-side rendering. If I use Universal-app for server-side rendering, my web-server returns not blank page, so using App Shell is redundant.

But I need some kind of App Shell for PWA, so I have to provide some URL with App Shell by myself without using `ng generate app-shell ...`

Please correct me, if I missed something.

FrozenPandaz:
Hi,

I have investigated this issue to a degree. I also cannot find any way of using app-shells with Universal...

Sorry i wrote a novel, feel free to skip down to the solutions

Conflict:
* Prerendering the app shell is done with the app server module
* The ModuleMapLoaderModule requires a module map, without it the app shell wont render
* It seems extraProviders during bootstrap are the bottom layer of providers
[...]

PR #12587, opened 2018-10-15
docs: update universal rendering story

alan-agius4:
Closes: #9202

And also updated the docs to match https://github.com/angular/universal-starter


alexeagle:
@alan-agius4 why does this need to go on the patch branch?

alan-agius4:
@alexeagle to be both branches updated since it's relevant to both versions. 

That said I think docs get only published from `master`, thus feel free to change the target to master only if you deem that it's better.

PR #12590, opened 2018-10-15
feat(@angular-devkit/build-optimizer): replace more ts helpers

alan-agius4:


packages/angular_devkit/build_optimizer/src/transforms/import-tslib.ts
```diff
@@ -7,13 +7,35 @@
  */
 import * as ts from 'typescript';
 
+const tsHelpers = [
```
filipesilva: Are you sure all the helpers follow the same regex format?

In some code paths we don't actually call `testImportTslib` so I worry about replacing a bunch of new variable names with helpers. 

For instance, I wasn't be surprised if code in the wild used a variable named `__values` or `__rest`. I think we need some more checks to make sure we're not replacing things we shouldn't.

alan-agius4: The new helpers so match the same RegExp format.

Any ideas on `some more checks`?

filipesilva: No good concrete idea, no. On a high level we want to make sure we only replace data structures that really are helpers. 

At the moment we use the regex to say something like "this file contains at least one var with one of these names, the assignment starts with a parenthesis, there's a couple of indented newlines, and it ends with a curly brace and a semicolon". 

Then when replacing  vars, we just replace all that have a specific name. This isn't a very strict definition really. If there's is a real `__extends` helper in a file, something as simple as `var __values = [1,2,3];` will also be replaced with `import { __values } from "tslib";`. This is of course bad.

Ideally we want to only replace the exact helper structure. But we also don't want these checks to be super slow, or hard to maintain.

One idea is to use TS to parse the exact structure we want into a `ts.Node`, then have some method that is able to compare nodes for the same structure. If we store that structure on the module top level, we don't need to keep recomputing it. 

I don't think this is a trivial change though. But I think we need such safety if we want to increase the number of helpers affected.

alan-agius4:
Dropped in favor of #12598

Issue #12260, opened 2018-09-13
AOT builds with webpack watch result in EOENT of ngFactory file


kspearrin:
Trying to perform Angular AOT builds while using `webpack --watch` results in errors. The first build works fine, however, subsequent builds from changes triggered by watch fail with:

```
Error: ENOENT: no such file or directory, stat 'C:\myapp\src\popup\app-routing.module.ngfactory.js'
```

This problem did not start until I upgraded angular CLI to v6 as well as webpack 4. See upgrade commit here: https://github.com/bitwarden/browser/commit/aaf774308e50823983a0823361271f15367927da

I am using webpack directly rather than through the Angular CLI (`ng serve`).

This seems like the same issue here (user is using the CLI here): https://github.com/angular/angular-cli/issues/11380 . It was closed by the author since he found a workaround, however, the workaround does not apply to my project.

Other related issues I found on the web:

- https://github.com/NativeScript/nativescript-dev-webpack/pull/574
- https://github.com/NativeScript/nativescript-dev-webpack/issues/566

### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Command (mark with an `x`)
```
- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [x] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

```
node --version
v10.8.0
```
    "@angular/compiler-cli": "^6.1.7",
    "@ngtools/webpack": "^6.2.1",
    "webpack": "^4.18.0",
    "webpack-cli": "^3.1.0"

### Repro steps

I have two OSS projects where this issue is occurring. You can use them as examples to reproduce this error from if you like.

https://github.com/bitwarden/browser
https://github.com/bitwarden/browser/blob/master/webpack.config.js

https://github.com/bitwarden/web
https://github.com/bitwarden/web/blob/master/webpack.config.js

On either, just run:

```
npm install
npm run build:prod:watch
```

Then change a file in `./src/popup/` or `./src/app` to trigger the watch.

### The log given by the failure

```
Error: ENOENT: no such file or directory, stat 'C:\myapp\src\popup\app-routing.module.ngfactory.js'
    at Object.statSync (fs.js:815:3)
[...]

Issue #9669, opened 2018-02-17
ng serve (dev mode) doesn't serve updated file from assets directory


kdcro101:
ng serve (dev mode) doesn't serve updated file from assets directory, it serves "version" that was first transmitted (caching problem?).

When file in assets directory is changed via external shell script (child tsconfig to generate javascript for web workers [tsc + browserify + uglify]), ng serve serves version of file that was transmitted on first request.

Server needs to be restarted for changes to pick-up. 

### Versions
```
Angular CLI: 1.6.8
Node: 8.9.3
OS: linux x64
Angular: 5.2.5
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cdk: 5.2.1
@angular/cli: 1.6.8
@angular/flex-layout: 2.0.0-beta.12
@angular/material: 5.2.1
@angular-devkit/build-optimizer: 0.0.42
@angular-devkit/core: 0.0.29
@angular-devkit/schematics: 0.0.52
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.8
@schematics/angular: 0.1.17
typescript: 2.5.3
webpack-bundle-analyzer: 2.10.0
webpack: 3.10.0
```

### Repro steps

run `ng serve`, change some javascript file inside assets directory
`assets/workers/some.js` for example, using above mentioned tools

### Observed behavior
ng serve - serves outdated js file which doesn't exist on storage/drive (caching?)

### Desired behavior
To serve updated file from storage.

kdcro101:
Also, when files are copied into project (imported type defs) it doesn't trigger change mechanism.
If I am having error  which is reported via console output of `ng serve` after copying  into project , corrected .ts files, using linux cp  command it doesn't trigger ng serve rebuild processs, so `ng serve` must be restarted

clydin:
This appears to be a bug but we will need to look at a reproduction to find and fix the problem. Can you setup a minimal reproduction please? 

You can read [here](https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#-submitting-an-issue) why this is needed. A good way to make a minimal reproduction is to create a new app via `ng new repro-app` and adding the minimum possible code to show the problem. Then you can push this repository to github and link it here.

MickL:
I can reproduce this on latest version `@angular/cli ~6.0.8`

virgil-av:
[...]

PR #12591, opened 2018-10-15
Fix virtual file invalidation

filipesilva:
Followup to https://github.com/angular/angular-cli/pull/12588, should fix https://github.com/angular/angular-cli/issues/12260 going forward.

Fix https://github.com/angular/angular-cli/issues/9669
Fix https://github.com/angular/angular-cli/issues/12657

### DIFF
--- a/packages/ngtools/webpack/src/compiler_host.ts
+++ b/packages/ngtools/webpack/src/compiler_host.ts
@@ -33,6 +33,13 @@ export class WebpackCompilerHost implements ts.CompilerHost {
   private _basePath: Path;
   private _resourceLoader?: WebpackResourceLoader;
   private _sourceFileCache = new Map<string, ts.SourceFile>();
+  private _virtualFileExtensions = [
+    '.js', '.js.map',
+    '.ngfactory.js', '.ngfactory.js.map',
+    '.ngstyle.js', '.ngstyle.js.map',
+    '.ngsummary.json',
+  ];
+
 
   constructor(
     private _options: ts.CompilerOptions,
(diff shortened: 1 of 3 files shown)


packages/ngtools/webpack/src/compiler_host.ts
```diff
@@ -79,19 +86,24 @@ export class WebpackCompilerHost implements ts.CompilerHost {
...
-
     this._sourceFileCache.delete(fullPath);
 
     try {
       const exists = this._syncHost.isFile(fullPath);
       if (exists) {
         this._changedFiles.add(fullPath);
       }
-    } catch { }
+    } catch {
```
clydin: `isFile` may or may not throw if the file exists.  Some host do and some do not.
Maybe something similar to the following:
```
let exists = false;
try {
  exists = this._syncHost.isFile(fullPath);
  if (exists) {
    this._changedFiles.add(fullPath);
  }
} catch { }

// File doesn't exist anymore, so we should delete the related virtual files.
if (!exists && fullPath.endsWith('.ts')) {
  this._virtualFileExtensions.forEach(ext => {
    const virtualFile = (fullPath.slice(0, -3) + ext) as Path;
    if (this._memoryHost.exists(virtualFile)) {
      this._memoryHost.delete(virtualFile);
    }
  });
}
```

alan-agius4:
@filipesilva I think this also solves https://github.com/angular/angular-cli/issues/12620?

filipesilva:
I'm not sure it's related. Have you tested it?

PR #12593, opened 2018-10-15
feat(@ngtools/webpack): allow custom lazy module resource

filipesilva:
Fix #12591

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -106,6 +106,7 @@ export interface AngularCompilerPluginOptions {
 
   // added to the list of lazy routes
   additionalLazyModules?: { [module: string]: string };
+  additionalLazyModuleResources?: string[];
 
   // The ContextElementDependency of correct Webpack compilation.
   // This is needed when there are multiple Webpack installs.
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -102,6 +102,7 @@ export interface AngularCompilerPluginOptions {
   missingTranslation?: string;
   platform?: PLATFORM;
   nameLazyFiles?: boolean;
+  lazyModulesResource?: string;
```
hansl: `string | string[]`?

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -123,6 +124,15 @@ export enum PLATFORM {
...
 }
 
+// APFv6 does not have single FESM anymore. Instead of verifying if we're pointing to
+// FESMs, we resolve the `@angular/core` path and verify that the path for the
+// module starts with it.
+// This may be slower but it will be compatible with both APF5, 6 and potential future
+// versions (until the dynamic import appears outside of core I suppose).
+// We resolve any symbolic links in order to get the real path that would be used in webpack.
+const angularCorePackagePath = require.resolve('@angular/core/package.json');
+const angularCoreResourceRoot = fs.realpathSync(path.dirname(angularCorePackagePath));
```
hansl: `realpath` might or might not play well with bazel and `preserve-symlink`. Just want to make sure it's taken into account.

filipesilva: I think it should be ok since it's the realpath to a file that was already resolved via require.resolve (thus using Bazel resolution). It's also the same logic as before though, so this change isn't breaking that.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
filipesilva: I'm unsure what the final semantics here should be. Should we allow several extra lazy module resources? Each would be a file containing a loader and it will have all the lazy modules as context.

filipesilva: Changed it always be an array.

filipesilva:
We should consider exposing this via an CLI build option in the future. It can help in cases where you have your own resource loader, like custom routers.

PR #12598, opened 2018-10-15
feat(@schematics/angular): enable typescript helpers in workspace

alan-agius4:


packages/schematics/angular/migrations/migration-collection.json
```diff
@@ -14,6 +14,11 @@
       "version": "7.0.0-rc.0",
       "factory": "./update-7/index#polyfillMetadataRule",
       "description": "Update an Angular CLI project to version 7."
+    },
+    "migration-04": {
+      "version": "7.1.0-beta.0",
```
alan-agius4: Since it's feature this will be out with the next betas

packages/schematics/angular/utility/latest-versions.ts
```diff
@@ -12,6 +12,7 @@ export const latestVersions = {
   RxJs: '~6.3.3',
   ZoneJs: '~0.8.26',
   TypeScript: '~3.1.1',
+  TsLib: '^1.9.0',
   // The versions below must be manually updated when making a new devkit release.
   DevkitBuildAngular: '~0.9.0-rc.1',
   DevkitBuildNgPackagr: '~0.9.0-rc.1',
```
Splaktar: It seems like Angular and DevkitBuild* here need to be updated as well? Perhaps in a separate PR, but they seem quite out of date and are this way on `master` too.

PR #12643, opened 2018-10-19
fix(@schematics/angular): update `@angular` `peerDependencies` when g…

alan-agius4:
update `@angular` `peerDependencies` of new libraries

### DIFF
--- a/packages/schematics/angular/library/files/__projectRoot__/package.json
+++ b/packages/schematics/angular/library/files/__projectRoot__/package.json
@@ -2,7 +2,7 @@
   "name": "<%= dasherize(packageName) %>",
   "version": "0.0.1",
   "peerDependencies": {
-    "@angular/common": "^6.0.0-rc.0 || ^6.0.0",
-    "@angular/core": "^6.0.0-rc.0 || ^6.0.0"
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/library/files/__projectRoot__/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "<%= dasherize(packageName) %>",
   "version": "0.0.1",
   "peerDependencies": {
-    "@angular/common": "^6.0.0-rc.0 || ^6.0.0",
-    "@angular/core": "^6.0.0-rc.0 || ^6.0.0"
+    "@angular/common": "^7.0.0",
```
filipesilva: Can you change this to use `angular-cli/packages/schematics/angular/utility/latest-versions.ts` instead?

alan-agius4: I was gonna do it, though I noticed that we are targeting that path versions in the `latest-versions.ts` should I replace the `~` with a `^` here?

alan-agius4: Updated

PR #12655, opened 2018-10-19
Various QoL with TSLint and Docs

hansl:
An example of a TSLint run that should have failed but TSLint didn't fail; https://circleci.com/gh/angular/angular-cli/23346


filipesilva:
Really like the nice readmes!

Issue #11005, opened 2018-05-24
ng lint --format stylish by default should apply to main app


johnpapa:
### Versions
```

Angular CLI: 6.0.3
Node: 8.11.1
OS: darwin x64
Angular: 6.0.3
... animations, cli, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.6.3
@angular-devkit/build-angular     0.6.3
@angular-devkit/build-optimizer   0.6.3
@angular-devkit/core              0.6.3
@angular-devkit/schematics        0.6.3
@ngtools/webpack                  6.0.3
@schematics/angular               0.6.3
@schematics/update                0.6.3
rxjs                              6.2.0
typescript                        2.7.2
webpack                           4.8.3
    
```

### Repro steps

```
ng new my-dream-app
cd my-dream-app
ng lint --format stylish
```

### Observed behavior
```
▶ng lint --format stylish
Architect commands with multiple targets cannot specify overrides.'lint' would be run on the following projects: my-app,my-app-e2e
Error: Architect commands with multiple targets cannot specify overrides.'lint' would be run on the following projects: my-app,my-app-e2e
    at LintCommand.validate (/Users/papa/_play/my-app/node_modules/@angular/cli/models/architect-command.js:75:23)
    at /Users/papa/_play/my-app/node_modules/@angular/cli/models/command-runner.js:274:39
    at Generator.next (<anonymous>)
    at /Users/papa/_play/my-app/node_modules/@angular/cli/models/command-runner.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/papa/_play/my-app/node_modules/@angular/cli/models/command-runner.js:3:12)
    at validateAndRunCommand (/Users/papa/_play/my-app/node_modules/@angular/cli/models/command-runner.js:273:12)
    at Object.<anonymous> (/Users/papa/_play/my-app/node_modules/@angular/cli/models/command-runner.js:100:26)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/papa/_play/my-app/node_modules/@angular/cli/models/command-runner.js:4:58)
```

### Desired behavior

By default if no app is specified, it should cover the main app (not the e2e)

mikesimmonds:
For any less-experienced developers like myself coming across this issue:
When passing flags to `ng lint` (eg --fix) you must include the name of the project (found in package.json).
[...]

Issue #10656, opened 2018-05-04
Lost ability to auto fix linting errors


samherrmann:
### Versions
```
Angular CLI: 6.0.0
Node: 8.11.1
OS: linux x64
Angular: 6.0.0
... animations, cli, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.6.0
@angular-devkit/build-angular     0.6.0
@angular-devkit/build-optimizer   0.6.0
@angular-devkit/core              0.6.0
@angular-devkit/schematics        0.6.0
@ngtools/webpack                  6.0.0
@schematics/angular               0.6.0
@schematics/update                0.6.0
rxjs                              6.1.0
typescript                        2.7.2
webpack                           4.6.0
```

### Repro steps
Run the following command in a Angular v6 project:
```
ng lint --fix
```

### Observed behavior
Terminal output:
```
Architect commands with multiple targets cannot specify overrides.
```

### Desired behavior
As a developer I would like to use the `fix` flag to lint my code on the development machine. I am aware that I am able to add the `fix` option in `angular.json` per [Angular workspace documentation](https://github.com/angular/angular-cli/wiki/angular-workspace) and [@angular-devkit/build-angular:tslint](https://github.com/angular/angular-cli/blob/v6.0.0-rc.8/packages/%40angular/cli/lib/config/schema.json#L1519-L1594). Doing so will cause the CI machine to also "auto fix" linting errors. The desired functionality is to allow developers to easily auto fix linting errors while the CI machine should simply fail upon encountering linting errors.

isaackwan:
I think per [this SO solution](https://stackoverflow.com/a/50189944/1348400) we need to specify the 'library' first, before any 'overrides', i.e. `--fix`. So for my project it's either `web2` or `web2-e2e`. I was able to fix my source file with:

```
$ ng lint web2 --fix
Fixed 7 error(s) in /home/isaac/Development/csci4140/web2/src/app/components/bill-new/bill-new.component.ts
<...>

All files pass linting.
```

samherrmann:
Great! I guess the documentation needs an update then. The following is what I have in my scripts now:
```
"lint": "ng lint project-foo && ng lint project-foo-e2e",
"lint-fix":  "ng lint project-foo --fix && ng lint project-foo-e2e --fix",
```

kvincent3:
This works for me : 
[...]

PR #12673, opened 2018-10-21
fix(@angular/cli): pass arguments to all targets

### LINKED ISSUE #10657: ng lint with format flag errors
### Versions
```
Angular CLI: 6.0.0
Node: 9.11.1
OS: win32 x64
Angular: 6.0.0

@angular-devkit/architect         0.6.0
@angular-devkit/build-angular     0.6.0
@angular-devkit/build-optimizer   0.6.0
@angular-devkit/core              0.6.0
@angular-devkit/schematics        0.6.0
@ngtools/webpack                  6.0.0
@schematics/angular               0.6.0
@schematics/update                0.6.0
rxjs                              6.1.0
typescript                        2.7.2
webpack                           4.6.0
```

### Repro steps
1. Create new application via angular-cli.
1. Run `ng lint` with format flag.
```bash
ng new my-test
cd my-test
ng lint --format checkstyle
```

I can edit _angular.json_ to add `"format":"checkstyle"` option to "my-test" and "my-test-e2e" projects. Then `ng lint` will run successfully and console will contain checkstyle output for each project.

### Observed behavior
```
Architect commands with multiple targets cannot specify overrides.'lint' would be run on the following projects: my-test,my-test-e2e
Error: Architect commands with multiple targets cannot specify overrides.'lint' would be run on the following projects: my-test,my-test-e2e
    at LintCommand.validate (C:\repos\my-test\node_modules\@angular\cli\models\architect-command.js:91:23)
    at C:\repos\my-test\node_modules\@angular\cli\models\command-runner.js:274:39
    at Generator.next (<anonymous>)
    at C:\repos\my-test\node_modules\@angular\cli\models\command-runner.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (C:\repos\my-test\node_modules\@angular\cli\models\command-runner.js:3:12)
    at validateAndRunCommand (C:\repos\my-test\node_modules\@angular\cli\models\command-runner.js:273:12)
    at Object.<anonymous> (C:\repos\my-test\node_modules\@angular\cli\models\command-runner.js:100:26)
    at Generator.next (<anonymous>)
    at fulfilled (C:\repos\my-test\node_modules\@angular\cli\models\command-runner.js:4:58)
```

When using format option in _angular.json_, I get 2 checkstyle outputs (one for each project).
```
ng lint
<?xml version="1.0" encoding="utf-8"?><checkstyle version="4.3"></checkstyle>
<?xml version="1.0" encoding="utf-8"?><checkstyle version="4.3"></checkstyle>
```

### Desired behavior
* I would like to be able to pass in `format` param via command line so that it can be used via CI and local builds. 
* Default output is prose in console.
* Aggregate all linting errors preferred but can work around. (This was previous behavior).
   * Aggregate lint output for all projects so that I can pipe to file
   * For checkstyle it has all errors in one xml output instead of 2

We use Jenkins CI and use checkstyle reporter to file for CI builds only, but would like to retain prose console output for local builds. We cannot do that when format is set in _angular.json_.

This is similar to #10422 but reporter was using multiple libraries. This issue is visible in a basic application.

mnahkies:
When running a command with args against multiple targets, all targets should be given the args. As parseArguments was mutating the passed args array this wasn't the case. 

Fix by making a shallow clone of the array. This was especially noticeable when using the `ng lint --fix` command on a newly generated project, as files in the app target would be fixed, but e2e target would be only be linted (with no fix).

Possibly closes #10657, #10656, #11005

### DIFF
--- a/packages/angular/cli/models/parser.ts
+++ b/packages/angular/cli/models/parser.ts
@@ -122,7 +122,7 @@ function _removeLeadingDashes(key: string): string {
 
 function _assignOption(
   arg: string,
-  args: string[],
+  nextArg: string | undefined,
   { options, parsedOptions, leftovers, ignored, errors, warnings }: {
     options: Option[],
     parsedOptions: Arguments,
@@ -134,9 +134,10 @@ function _assignOption(
   },
 ) {
   const from = arg.startsWith('--') ? 2 : 1;
+  let consumedNextArg = false;
   let key = arg.substr(from);
   let option: Option | null = null;
-  let value = '';
+  let value: string | undefined = '';
   const i = arg.indexOf('=');
 
   // If flag is --no-abc AND there's no equal sign.
@@ -155,7 +156,7 @@ function _assignOption(
       // Set it to true if it's a boolean and the next argument doesn't match true/false.
       const maybeOption = _getOptionFromName(key, options);
       if (maybeOption) {
-        value = args[0];
+        value = nextArg;
         let shouldShift = true;
 
         if (value && value.startsWith('-')) {
@@ -167,7 +168,7 @@ function _assignOption(
 
         // Only absorb it if it leads to a better value.
         if (shouldShift && _coerce(value, maybeOption) !== undefined) {
-          args.shift();
+          consumedNextArg = true;
         } else {
           value = '';
         }
@@ -183,9 +184,9 @@ function _assignOption(
   }
 
   if (option === null) {
-    if (args[0] && !args[0].startsWith('-')) {
-      leftovers.push(arg, args[0]);
-      args.shift();
+    if (nextArg && !nextArg.startsWith('-')) {
+      leftovers.push(arg, nextArg);
+      consumedNextArg = true;
     } else {
       leftovers.push(arg);
     }
@@ -220,6 +221,8 @@ function _assignOption(
       ignored.push(arg);
     }
   }
+
+  return consumedNextArg;
 }
 
 
@@ -303,15 +306,18 @@ export function parseArguments(
 
   const state = { options, parsedOptions, positionals, leftovers, ignored, errors, warnings };
 
-  for (let arg = args.shift(); arg !== undefined; arg = args.shift()) {
+  for (let argIndex = 0; argIndex < args.length; argIndex++) {
+    const arg = args[argIndex];
+    let consumedNextArg = false;
+
     if (arg == '--') {
       // If we find a --, we're done.
-      leftovers.push(...args);
+      leftovers.push(...args.slice(argIndex + 1));
       break;
     }
 
     if (arg.startsWith('--')) {
-      _assignOption(arg, args, state);
+      consumedNextArg = _assignOption(arg, args[argIndex + 1], state);
     } else if (arg.startsWith('-')) {
       // Argument is of form -abcdef.  Starts at 1 because we skip the `-`.
       for (let i = 1; i < arg.length; i++) {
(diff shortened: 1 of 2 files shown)


imcotton:
ping @clydin

mnahkies:
@clydin this is a fair comment - I'll have a look at this tonight

packages/angular/cli/models/parser.ts
```diff
@@ -259,6 +259,8 @@ export function parseFreeFormArguments(args: string[]): Arguments {
  * @returns An object that contains a property per option.
  */
 export function parseArguments(args: string[], options: Option[] | null): Arguments {
+  args = args.slice();
```
imcotton:
```suggestion
export function parseArguments(originArgs: string[], options: Option[] | null): Arguments {
  const args = [ ...originArgs ];
```
Something like this?

imcotton: I'm a little bit of confused, wouldn't the below tweaks guarantees the `args` been local from the beginning? (already renamed the param form `args` to `originArgs`)

```diff
- export function parseArguments(args: string[], options: Option[] | null): Arguments {
+ export function parseArguments(originArgs: string[], options: Option[] | null): Arguments {
+  const args = [ ...originArgs ];
```

Or you meant the local variable should being `ReadonlyArray` type?

clydin: It would still be mutating and creating numerous copies that don't need to be created.  Don't mutate the input argument and there's no need to make copies at all.

mnahkies:
@clydin and now I remember why I didn't do this in the first place, not quite the simple fix I had hoped! The file is well covered by unit tests though, so I'm pretty confident that the behaviour has not been changed (aside from stopping the mutation)

mnahkies:
Hi @clydin any chance you could review this soon? Cheers

packages/angular/cli/models/parser.ts
```diff
@@ -270,30 +284,73 @@ export function parseArguments(args: string[], options: Option[] | null): Argume
   const ignored: string[] = [];
   const errors: string[] = [];
 
-  for (let arg = args.shift(); arg !== undefined; arg = args.shift()) {
+  for (let i = 0; i < args.length; i++) {
```
clydin: Instead of using 'i' which is already used below, use something similar to `argIndex`.   Eliminates the need to change the other loop to use `j`.

packages/angular/cli/models/parser.ts
```diff
@@ -109,19 +109,29 @@ function _getOptionFromName(name: string, options: Option[]): Option | undefined
...
+function _assignOption({
+  arg,
+  nextArg,
+  options,
+  parsedOptions,
+  leftovers,
+  ignored,
+  errors,
+}: {
   arg: string,
```
clydin: change the first argument from being the actual argument to the index of the argument.  You can then access the current argument and next argument by using the existing `args` array.

Also, have it return the new index.  This would eliminate the extra conditionals present in the calling function.

mnahkies: I don't think this is possible, as the value for "arg" is computed rather than from the array for one of the function calls (I'll leave a comment pointing out the call in question)

packages/angular/cli/models/parser.ts
```diff
@@ -270,30 +284,73 @@ export function parseArguments(args: string[], options: Option[] | null): Argume
...
+        const flag = arg[j];
         // If the next character is an '=', treat it as a long flag.
-        if (arg[i + 1] == '=') {
-          const f = '--' + flag + arg.slice(i + 1);
-          _assignOption(f, args, options, parsedOptions, positionals, leftovers, ignored, errors);
+        if (arg[j + 1] == '=') {
+          const f = '--' + flag + arg.slice(j + 1);
+
+          const consumedNextArg = _assignOption({
+            arg: f,
```
mnahkies: For this usage of _assignOption we pass in a computed value rather than a value from the args array

mnahkies:
Rebased on current master

alan-agius4:
This would also fix https://github.com/angular/angular-cli/issues/13111

PR #12703, opened 2018-10-22
refactor(@angular-devkit/build-angular): cleanup/fix linting of webpack configs

clydin:


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -5,54 +5,43 @@
...
+import { WebpackConfigOptions } from '../build-options';
+import { getOutputHashFormat, normalizeExtraEntryPoints } from './utils';
 
 const ProgressPlugin = require('webpack/lib/ProgressPlugin');
 const CircularDependencyPlugin = require('circular-dependency-plugin');
 const TerserPlugin = require('terser-webpack-plugin');
 const StatsPlugin = require('stats-webpack-plugin');
 
-/**
- * Enumerate loaders and their dependencies from this file to let the dependency validator
```
filipesilva: Don't these work anymore? I think we want them to work.

clydin: we don't have the duplicate package.json structure anymore plus the `no-implicit-dependencies` tslint rule will fail if a dependency is imported that is not listed in the package's dependencies.

Issue #12667, opened 2018-10-20
experimental ivy creates angular cli project v7.0.0-rc3 while cli is version 7.0.2


kfirufk:
```

- [x ] bug report 
```

### Command (mark with an `x`)
```
- [x ] new
```

### Versions

MacOS Mojave 10.4
NodeJS v10.12.0
Angular CLI 7.0.2

### Repro steps

`ng new a --experimental-ivy
`

### Desired functionality
it should create a project with angular 7 final

### Mention any other details that might be useful
it creates a project with angular cli v7.0.0-rc3, it doesn't happen if I don't enable experimental ivy.

alan-agius4:
When selecting to create a project with `experminalIvy` it will create the project with the `next` versions of the package instead of the stable one.

PR #12709, opened 2018-10-23
fix(@schematics/angular): experimentalIvy creates project using `next` packages

alan-agius4:
Close: #12667

### DIFF
--- a/packages/schematics/angular/ng-new/index.ts
+++ b/packages/schematics/angular/ng-new/index.ts
@@ -40,7 +40,6 @@ export default function (options: NgNewOptions): Rule {
   const workspaceOptions: WorkspaceOptions = {
     name: options.name,
     version: options.version,
-    experimentalAngularNext: options.experimentalIvy,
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/ng-new/index.ts
```diff
@@ -40,7 +40,6 @@ export default function (options: NgNewOptions): Rule {
   const workspaceOptions: WorkspaceOptions = {
     name: options.name,
     version: options.version,
-    experimentalAngularNext: options.experimentalIvy,
```
clydin: This line is the only part that can be safely removed without causing a breaking change.  Even though the `experimentalAngularNext` is hidden it can still be used.

alan-agius4: Should we deprecate it? Or leave it as is?

clydin: @hansl thoughts on the `experimentalAngularNext` option?

PR #12721, opened 2018-10-23
build: Use fine-grained node_module deps

kyliau:
This commit updates the BUILD files to specify fine-grained node_module deps
by replacing "@typings" comments with actual @npm node module.

Moved tools/bazel.rc -> .bazelrc

Removed "jasmine" typings from base tsconfig.json

Added @bazel/karma to devDependencies, needed for `ts_web_test`


.bazelignore
```diff
@@ -0,0 +1 @@
+node_modules
```
alexeagle: probably also want a `dist` directory here

packages/angular_devkit/build_optimizer/package.json
```diff
@@ -8,6 +8,7 @@
     "build-optimizer": "./src/build-optimizer/cli.js"
   },
   "dependencies": {
+    "@angular-devkit/core": "^7.0.2",
```
alexeagle: I don't think this should be fixed at 7.0.2 - do we really need this line added?

packages/angular_devkit/schematics/tools/export-ref_spec.ts
```diff
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+/// <reference types="@types/jasmine"/>
```
alexeagle: instead of adding these, we should probably do the same as angular/angular where the tsconfig-test.json gives the jasmine types for free

/cc @gregmagolan

tools/BUILD
```diff
@@ -10,10 +10,10 @@ load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
 nodejs_binary(
     name = 'quicktype_runner',
     data = [
-        ":quicktype_runner.js",
+        "quicktype_runner.js",
     ],
-    node_modules = "//:quicktype_node_modules",
-    entry_point = "angular_devkit/tools/quicktype_runner.js",
+    node_modules = "@npm//quicktype-core",
```
alexeagle: there should be a corresponding delete of the code that declared the quicktype_node_modules

tsconfig.json
```diff
@@ -36,7 +36,6 @@
       "./node_modules/@types"
     ],
     "types": [
-      "jasmine",
       "node"
```
alexeagle: yeah that jasmine removal should be moved to a tsconfig-test.json that we use for any `testonly=1` compilation

.bazelignore
kyliau: done

packages/angular_devkit/build_optimizer/package.json
kyliau: yeah, it's not needed. deleted this line.

packages/angular_devkit/schematics/tools/export-ref_spec.ts
kyliau: removed reference and added tsconfig-test.json

tsconfig.json
kyliau: ack

tools/BUILD
kyliau: done

BUILD
```diff
@@ -28,7 +29,7 @@ NODE_MODULES_EXCLUDE = [
 filegroup(
     name = "node_modules",
```
filipesilva: Is this glob still needed? I thought it wasn't with the fine grained deps.

package.json
```diff
@@ -101,16 +103,21 @@
     "istanbul": "^0.4.5",
     "jasmine": "^2.6.0",
     "jasmine-spec-reporter": "^3.2.0",
+    "karma-jasmine-html-reporter": "^0.2.2",
     "license-checker": "^20.1.0",
     "minimatch": "^3.0.4",
     "minimist": "^1.2.0",
+    "npm-registry-client": "^8.6.0",
+    "pidtree": "^0.3.0",
```
filipesilva: Shouldn't these dependencies be spread out over the packages that use them instead? 

Having them here is odd. Why are these here but not others? The choice of packages to include here seems like a side effect of how yarn is hoisting packages.

packages/angular_devkit/core/node/resolve.ts
```diff
@@ -259,7 +259,7 @@ export function resolve(x: string, options: ResolveOptions): string {
   }
 
   function nodeModulesPaths(start: string, opts: ResolveOptions) {
-    const modules = ['node_modules'];
+    const modules = ['node_modules', path.join('npm', 'node_modules')];
```
filipesilva: Where does the magic string `npm` come from here? This looks like it should be documented and maybe configurable, since `@angular-devkit/core` is used outside of this repository.

package.json
clydin: Only dependencies used by the development scripts and build system should be included here

packages/angular_devkit/core/node/resolve.ts
clydin: This is used at runtime during an ng build or serve.
Bazel should have no effect here.

BUILD
alexeagle: good catch, it should be removed

package.json
clydin: But these seem arbitrary compared to the variety of other dependencies used by the packages that are not added here.

A single directory per Bazel package I think would be preferable over a single directory per Bazel workspace.  Each package may have varying and conflicting npm dependencies.

> yarn workspaces requires the modules to be layed out in many directories.

it's not really yarn workspaces but rather the packages themselves and their dependencies that require it.  Otherwise the packages are not receiving the dependencies they specify.

packages/angular_devkit/core/node/resolve.ts
kyliau: I reverted this change, and instead added a "fake" node_modules directory so that the resolve() test works.

package.json
kyliau: This will be addressed by the upcoming work in https://angular-team.atlassian.net/browse/TOOL-463

PR #12735, opened 2018-10-24
Enable Webpack E2Es

### LINKED ISSUE #12686: ERROR in Patterns must be a string or an array of strings
```
- [x] bug report
```

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

node v8.11.2
npm v5.6.0
OS: (irrelevant)

### Repro steps

I am using @ngtools/webpack and [webpack-contrib/copy-webpack-plugin](https://github.com/webpack-contrib/copy-webpack-plugin) and in that package [this](https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/utils/promisify.js#L1-L10) function uses `virtual_file_system_decorator`s stat function and if I change that to

    stat(path, callback) {
        let result = this._webpackCompilerHost.stat(path);
        if (result) {
            callback(null, this._webpackCompilerHost.stat(path));
        } else {
            callback(path, undefined);
        }
        // commented out what came after https://github.com/angular/angular-cli/commit/da7daef34bbdd0dbfbe83f21f71d45996eeb8dab
        // try {
        //     // tslint:disable-next-line:no-any
        //     callback(null, this._webpackCompilerHost.stat(path));
        // }
        // catch (e) {
        //     // tslint:disable-next-line:no-any
        //     callback(e, undefined);
        // }
    }

the error doesn't exist any more and everything compiles.

This happens when I upgraded from angular 6 to angular 7 which involved updating this package in the process.

> ERROR in Patterns must be a string or an array of strings

alan-agius4:


packages/ngtools/webpack/src/virtual_file_system_decorator.ts
```diff
@@ -26,12 +26,14 @@ export class VirtualFileSystemDecorator implements InputFileSystem {
   }
 
   stat(path: string, callback: (err: Error, stats: Stats) => void): void {
-    try {
+    const result = this._webpackCompilerHost.stat(path);
```
clydin: `stat` can throw which would stop the callback from running.

Also, the webpack input file system directly uses the node callback which provides the error directly in the callback.  This error could be for any reason and not just that the file doesn't exist.

alan-agius4: I haven't seen a throw myself, when the fail doesn't exists stats returns `null`, Seeing the code https://github.com/angular/angular-cli/blob/a53cf44e1db15f7e0680539fa9857506072e5104/packages/ngtools/webpack/src/compiler_host.ts#L190-L20. I looks like it is handled
 
The error itself origins from `globby`: https://github.com/sindresorhus/globby/blob/master/index.js#L14

clydin: So the real issue appears to be here then: https://github.com/webpack-contrib/copy-webpack-plugin/blob/1f9272945957077a623683e1c4d959d477e1555c/src/preProcessPattern.js#L76

That conditional should have the same checks as the catch block.

alan-agius4:
As discussed offline with @clydin, we'll create a PR to copy-webpack-plugin so that they handle undefined values in the `.then` block

PR #12741, opened 2018-10-24
Edit and expand command docs

jbogarthyde:
This one is just the top-level commands. Will do separate PR for subcommands (generation schematics).

### DIFF
--- a/packages/angular/cli/commands/definitions.json
+++ b/packages/angular/cli/commands/definitions.json
@@ -14,14 +14,14 @@
           }
         },
         "configuration": {
-          "description": "A named configuration environment, as specified in the \"configurations\" section of angular.json.",
+          "description": "A named build target, as specified in the \"configurations\" section of angular.json.\nEach named target is accompanied by a configuration of option defaults for that target.",
           "type": "string",
           "aliases": [
             "c"
           ]
         },
         "prod": {
-          "description": "When true, sets the build configuration to the production environment.\nAll builds make use of bundling and limited tree-shaking, A production build also runs limited dead code elimination using UglifyJS.",
+          "description": "When true, sets the build configuration to the production target.\nAll builds make use of bundling and limited tree-shaking, A production build also runs limited dead code elimination using UglifyJS.",
           "type": "boolean"
         }
       }
@@ -42,13 +42,13 @@
           "type": "boolean",
           "default": false,
           "aliases": [ "d" ],
-          "description": "When true, run through and report activity without writing out results."
+          "description": "When true, runs through and reports activity without writing out results."
         },
         "force": {
           "type": "boolean",
           "default": false,
           "aliases": [ "f" ],
-          "description": "When true, force overwriting of existing files."
+          "description": "When true, forces overwriting of existing files."
         },
         "interactive": {
           "type": "boolean",
--- a/packages/angular/cli/commands/definitions.json
+++ b/packages/angular/cli/commands/definitions.json
@@ -21,7 +21,7 @@
           ]
         },
         "prod": {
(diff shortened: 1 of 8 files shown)


packages/angular/cli/commands/run.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/schema",
   "$id": "ng-cli://commands/run.json",
-  "description": "Runs a custom target defined in your project.",
+  "description": "Runs an architect target with a custom build configuration defined in your project.",
```
alan-agius4: Note sure about the terminology here, if it should be `build` or `builder`

```
custom build configuration -> custom builder configuration
```

packages/angular/cli/commands/run.json
```diff
@@ -15,14 +15,14 @@
...
-          "description": "The target to run.",
+          "description": "The command to run.",
           "$default": {
             "$source": "argv",
             "index": 0
           }
         },
         "configuration": {
-          "description": "Specify the configuration to use.",
+          "description": "A named build configuration, defined in the \"configurations\" section of angular.json.\nThe build uses the configuration associated with the given target.",
```
alan-agius4: same
```
A named build configuration - > A named builder configuration
The build uses -> The builder uses
```

packages/angular/cli/commands/definitions.json
```diff
@@ -14,14 +14,14 @@
...
-          "description": "A named configuration environment, as specified in the \"configurations\" section of angular.json.",
+          "description": "A named build target, as specified in the \"configurations\" section of angular.json.\nEach named target is accompanied by a configuration of option defaults for that target.",
           "type": "string",
           "aliases": [
             "c"
           ]
         },
         "prod": {
-          "description": "When true, sets the build configuration to the production environment.\nAll builds make use of bundling and limited tree-shaking, A production build also runs limited dead code elimination using UglifyJS.",
+          "description": "When true, sets the build configuration to the production target.\nAll builds make use of bundling and limited tree-shaking, A production build also runs limited dead code elimination using UglifyJS.",
```
hansl: `production configuration`. The target is chosen by the command (`ng build` uses the `build` target), etc.

packages/angular/cli/commands/run.json
hansl: I would use `Builder` when it's not a build command, yes. So in this case yes, that's a good idea.

jbogarthyde: Is `Builder` another tool or technical term that we need to introduce?

jbogarthyde:
Please provide an example using the `run` command with the `universal` target.

jbogarthyde:
@hansl I have addressed your requested changes - please check and remove blocker.
@alan-agius4 
Do either of you know why it's failing continuous-integration?

packages/angular/cli/commands/definitions.json
alan-agius4: Should we change this to Terser instead of UglifyJS?

jbogarthyde: Removing tool reference.

packages/angular/cli/commands/run-long.md
```diff
@@ -0,0 +1,16 @@
...
+For example, the "server" target for a newly generated app has a predefined
+alternate configuration named "production".
+
+You can define new targets and their configuration options in the "architect" section
+of the `angular.json` file.
+If you do so, you can run them from the command line using the `ng run` command.
+Execute the command using the following format.
+
+```
+ng run project:target[.configuration]
```
alan-agius4: Should be `ng run project:target[:configuration]`

jbogarthyde:
Moved these commits to docs-preview branch - PR Docs preview #12837

PR #12743, opened 2018-10-24
Add and edit docs for generate subcommands 

jbogarthyde:
WIP - transfer edited doc content for generation schematics from draft to source files

### DIFF
--- a/packages/schematics/angular/application/schema.json
+++ b/packages/schematics/angular/application/schema.json
@@ -3,14 +3,15 @@
   "id": "SchematicsAngularApp",
   "title": "Angular Application Options Schema",
   "type": "object",
+  "description": "Generates a new basic app definition in the \"projects\" subfolder of the workspace.",
   "properties": {
     "projectRoot": {
-      "description": "The root directory of the new application.",
+      "description": "The root folder of the new app.",
       "type": "string",
       "visible": false
     },
     "name": {
-      "description": "The name of the application.",
+      "description": "The name of the new app.",
       "type": "string",
       "$default": {
         "$source": "argv",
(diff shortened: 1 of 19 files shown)


packages/schematics/angular/guard/schema.json
```diff
@@ -15,18 +16,18 @@
     },
     "spec": {
       "type": "boolean",
-      "description": "Specifies if a spec file is generated.",
+      "description": "When true (the default), generates a  \"spec.ts\" test file for the new guard.",
```
alan-agius4: Nit: would be nice to have the same descriptions for the same flag. 

For instance sometimes, it is `"When true, generates...`` and sometimes it is `When true (the default), generates ..`

alan-agius4:
@jbogarthyde, Can you please rebase?

packages/schematics/angular/application/schema.json
```diff
@@ -3,14 +3,15 @@
   "id": "SchematicsAngularApp",
   "title": "Angular Application Options Schema",
   "type": "object",
+  "description": "Generates a new basic app definition in the \"projects\" subfolder of the workspace.",
   "properties": {
     "projectRoot": {
-      "description": "The root directory of the new application.",
+      "description": "The root folder of the new app.",
```
alan-agius4: just a small nit, but in general we use `directory` instead of `folder` as terminology

jbogarthyde: Jury is out on this one. I tend to prefer folder. 
@jenniferfell 
The Author's Style Guide does not address the issue -- and in fact, that page uses "folder" 19 times and "directory" 12 times - sometimes in the same sentence.

jbogarthyde:
Moved these commits to docs-preview branch - PR Docs preview #12837

PR #12753, opened 2018-10-25
feat(@angular/cli): autodetect default package manager

clydin:


packages/angular/cli/utilities/check-package-manager.ts
```diff
@@ -5,49 +5,54 @@
...
-          }
-        }
-      });
 }
 
-function checkYarn() {
-  return execPromise('yarn --version')
-    .then(() => true, () => false);
-}
+export function getPackageManager(root: string): string {
```
filipesilva: Filename suggests the main export is `checkPackageManager` but instead it is `getPackageManager`. One of these should probably change. From the code, it looks like `getPackageManager` is the right name.

packages/angular/cli/utilities/config.ts
```diff
@@ -151,7 +151,7 @@ export function getProjectByCwd(workspace: experimental.workspace.Workspace): st
   }
 }
 
-export function getPackageManager(): string {
+export function getPackageManager(): string | null {
```
filipesilva: Maybe this should change to `getConfiguredPackageManager ` to avoid confusion?

clydin: Done.

packages/angular/cli/utilities/check-package-manager.ts
clydin: Done.

PR #12762, opened 2018-10-26
Updated Readme with correct node and npm versions

ausshadu:


filipesilva:
Heya, it's true that these versions are wrong but you are updating an old branch (6.0.x). The readme that users see on this repository is newer and doesn't show these anymore, so I don't think we need to change it. Thanks for trying to help us though!

PR #12768, opened 2018-10-26
fix(@schematics/angular): update `@angular-devkit/build-ng-packagr` w…

### LINKED ISSUE #12642: ng-packagr was not updated
```
- [ ] feature request
```

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Desired functionality
I just updated to 7.0.0 using command `ng update @angular/cli @angular/core`.
But `ng-packagr` was not updated and also there was problem with `tsickle` versioning.
Now since `ng-packagr` is used for generating library it should be updated as well.

alan-agius4:
…hen migrating

Closes #12642

### DIFF
--- a/packages/schematics/angular/migrations/migration-collection.json
+++ b/packages/schematics/angular/migrations/migration-collection.json
@@ -24,6 +24,11 @@
       "version": "7.0.4",
       "factory": "./update-7",
       "description": "Update an Angular CLI project to version 7."
+    },
+    "migration-06": {
+      "version": "7.0.3",
+      "factory": "./update-7/index#updateDevkitBuildNgPackagr",
+      "description": "Update an Angular CLI project to version 7."
     }
   }
 }
--- a/packages/schematics/angular/migrations/update-7/devkit-ng-packagr.ts
+++ b/packages/schematics/angular/migrations/update-7/devkit-ng-packagr.ts
@@ -1,0 +1,33 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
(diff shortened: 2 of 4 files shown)


packages/schematics/angular/migrations/migration-collection.json
```diff
@@ -19,6 +19,11 @@
       "version": "7.1.0-beta.0",
       "factory": "./update-7/index#typeScriptHelpersRule",
       "description": "Update an Angular CLI project to version 7."
+    },
+    "migration-05": {
+      "version": "7.0.3",
```
filipesilva: Yesterday we were talking about not adding migrations on minors/patch versions, so I'm not sure if we should add this one.

alan-agius4: In that case maybe I should add it as `7.1.0-beta.0`, what do you reckon?

clydin: I think this is really fixing a deficiency in the 7.0 update process.  So it would make sense to have it in a 7.0 patch release so that anyone updating from earlier majors will immediately benefit.

filipesilva: So the logic is that newly updated projects (6->7) will include this migration, but that we don't expect users to run update from 7->7 and to get it, right? I'm ok with that.

PR #12783, opened 2018-10-29
fix(@angular/cli): support all single dashes prefixed arguments

### LINKED ISSUE #12771: 'ng serve' fails with -c option
```
- [ ] feature request
```

```
- [ ] new
- [ ] build
- [x ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

* node version: 9.5.0
* npm version: 5.6.0
* angular version: 7.0.1 (with cli 7.0.3)

### Repro steps

I just had, when on v6.4.x (and below) the following:
`
ng serve--aot -c=uat --port=4201
`
and it worked fine.

When i upgraded, the CLI outputed this:

```
ng serve --aot -c=uat --port 4201
Unknown option: '--c'
```

Note that it had 2 '-' in the output of the CLI.

A workaround is to use `--configuration` option, but the shortcut is broken.

### LINKED ISSUE #12861: unable to install angular cli
kindely get solution to this error

i installed node and then installing angular

npm install -g @angular/cli
after that if it is installed correclty or not checking purpose i m using ng -v command it is unable to display the angular cli picture
screenshot 1

alan-agius4:
Fixes #12771
Fixes #12861

Also fixes issue that when you execute a pass an alias with a single `-` and doesn't exist. it will print out an alias with two `--`

```
ng build -x
Unknown option: '--x'
```


packages/angular/cli/models/parser_spec.ts
```diff
@@ -128,6 +128,7 @@ describe('parseArguments', () => {
     '--e3 true': { e3: true },
     '--e3=true': { e3: true },
     'a b c 1': { p1: 'a', p2: 'b', '--': ['c', '1'] },
+    '-p=1 -c=prod': {'--': ['-p=1', '-c=prod'] },
```
hansl: You're only adding tests for the last change, not the whole thing. Could you add the following test cases:

- `--p --c`
- `--p 123`
- `--p -c`
- `-p --c`
- `-p --c 123`
- `--c 123 -p`

? See if this works.

alan-agius4: @hansl, sure thing. I added the requested tests.

PR #12787, opened 2018-10-29
refactor: use `createConsoleLogger` and remove duplicate code

alan-agius4:


packages/angular_devkit/core/node/cli-logger.ts
```diff
@@ -32,7 +32,8 @@ export function createConsoleLogger(
           color = terminal.white;
           break;
         case 'warn':
-          color = (x: string) => terminal.bold(terminal.yellow(x));
+          color = terminal.yellow;
```
clydin: lost the bold here.

alan-agius4: updated

PR #12789, opened 2018-10-29
style: add comments in tslint.json

alan-agius4:
tslint does support comments in json


filipesilva:
@hansl I think there was a reason why we didn't use these comments overall. Do you remember anything?

alan-agius4:
@filipesilva, I think the reason was that along time ago tslint didn't support comments in json.

hansl:
Yeah and some people were complaining their editor was not compatible. I'm okay with this change though.

PR #12793, opened 2018-10-29
fix(@angular/cli): handle case senstive aliases

### LINKED ISSUE #12780: cli 7.0.3 '-S' flag creates inline-style project (--inline-style) instead of (--skip-tests)
```
- [ ] feature request
```

```
- [X] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions
macOS Mojave (10.14)

### Repro steps
**ng new project -S**

The project will be created with (--inline-style) and it will skip the CSS file, instead of the spec.ts file (--skip-tests)

If a new project is created without the '-S' flag then the CSS file is created. 

### Desired functionality
the '-S' flag should skip the spec.ts file from being installed

screen shot 2018-10-28 at 5 24 01 am

alan-agius4:
Closes #12780

Not sure for we can handle something like `noS` as at the moment this will eventually converted to `s`

### DIFF
--- a/packages/angular/cli/models/parser.ts
+++ b/packages/angular/cli/models/parser.ts
@@ -29,10 +29,10 @@ function _coerceType(str: string | undefined, type: OptionType, v?: Value): Valu
       }
 
       return _coerceType(str, OptionType.Boolean, v) !== undefined
-           ? _coerceType(str, OptionType.Boolean, v)
-           : _coerceType(str, OptionType.Number, v) !== undefined
-           ? _coerceType(str, OptionType.Number, v)
-           : _coerceType(str, OptionType.String, v);
+        ? _coerceType(str, OptionType.Boolean, v)
+        : _coerceType(str, OptionType.Number, v) !== undefined
+          ? _coerceType(str, OptionType.Number, v)
+          : _coerceType(str, OptionType.String, v);
 
     case OptionType.String:
       return str || '';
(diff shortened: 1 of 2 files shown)


packages/angular/cli/models/parser.ts
```diff
@@ -93,21 +93,28 @@ function _coerce(str: string | undefined, o: Option | null, v?: Value): Value |
 
 
 function _getOptionFromName(name: string, options: Option[]): Option | undefined {
-  const cName = strings.camelize(name);
+  const camelName = strings.STRING_CAMELIZE_REGEXP.test(name)
```
alan-agius4: Not sure if I just check for `-`.

packages/angular/cli/models/parser.ts
```diff
@@ -93,21 +93,28 @@ function _coerce(str: string | undefined, o: Option | null, v?: Value): Value |
...
-    if (option.aliases.some(x => x == name || x == cName)) {
+    if (option.aliases.some(x => x === name || x === camelName)) {
       return option;
     }
   }
 
   return undefined;
 }
 
+function removeLeadingDashes(key: string): string {
```
hansl: non-exported functions should start with a `_`.

alan-agius4: Updated.

PR #12806, opened 2018-10-30
fix(angularMajorCompatGuarantee): add check for range === null fixes #12644

### LINKED ISSUE #12644: Updating Angular CLI & Core: "Cannot read property 'split' of null"
```
- [ ] feature request
```

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [X] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

- Node version: 8.11.4
- npm version: 6.4.1
- Angular version: 6.1.8
- Angular CLI version: 6.2.3
- OS: Windows 10 x64
- (rxjs: 6.2.2)
- (typescript: 2.9.2)

### Repro steps
Using Angular with Angular Fire, Angular Flex, ngxs, firebase, rxjs.
Ran command:
`ng update @angular/cli @angular/core`

Only got this one line, nothing else:
`Cannot read property 'split' of null`

### Desired functionality
A working update would be the best, at least some more details on why Angular couldn't be updated would be nice.

tkrugg:
- Is there any better way to handle this check with typescript?
- I didn't find any tests targeting specifically peer dependencies. Can you point me at it?
- is there any linter formatter I should ran?


tkrugg:
I've signed it

packages/schematics/update/update/index.ts
```diff
@@ -33,6 +33,9 @@ type PeerVersionTransform = string | ((range: string) => string);
 // We export it to allow for testing.
 export function angularMajorCompatGuarantee(range: string) {
   range = semver.validRange(range);
```
alan-agius4: Can we understand why `range` is invalid in the first place? what the the inputted value?

tkrugg: when running `ng update`, the range variable takes the values of all the ranges found in the package files of your dependencies. They may and may not be valid ranges for semver (the very reason .validRange is called here I suppose).

If we put that aside.  Strictly speaking, the code fail prone written as is:
```js
const output = semver.validRange(range) // output === null
...
semver.validRange(output) // fails
```

alan-agius4:
Can you please rebase?

tkrugg:
rebased

mgechev:
@tkrugg the build is still failing. Would you fix the typing errors?

packages/schematics/update/update/index.ts
```diff
@@ -33,6 +33,9 @@ type PeerVersionTransform = string | ((range: string) => string);
...
+  if (range === null) {
```
alan-agius4: I think in here, we should return the non mutated `range` as like this you are returning `null` which is also bypassing TypeScript type inference as this method shouldn't return `null`.

The problem here is the `range` is being mutated. I think the actual code should be something along;

```ts
export function angularMajorCompatGuarantee(range: string): string {
  let newRange = semver.validRange(range);
  if (!newRange) {
    return range;
  }

  let major = 1;
  while (!semver.gtr(major + '.0.0', newRange)) {
    major++;
    if (major >= 99) {
      // Use original range if it supports a major this high
      // Range is most likely unbounded (e.g., >=5.0.0)
      return newRange;
    }
  }

  // Add the major version as compatible with the angular compatible, with all minors. This is
  // already one major above the greatest supported, because we increment `major` before checking.
  // We add minors like this because a minor beta is still compatible with a minor non-beta.
  for (let minor = 0; minor < 20; minor++) {
    newRange += ` || ^${major}.${minor}.0-alpha.0 `;
  }

  return semver.validRange(newRange) || range;
}
```

alan-agius4:
@tkrugg can you check the above comment?

Thank you.

packages/schematics/update/update/index.ts
tkrugg: agreed, changed it 
Sorry for missing your reply.

PR #12808, opened 2018-10-30
docs: mention Terser instead of UglifyJS

cexbrayat:
Now that https://github.com/angular/angular-cli/pull/11996 landed, the CLI uses Terser instead of UglifyJS.

### DIFF
--- a/docs/documentation/build.md
+++ b/docs/documentation/build.md
@@ -35,7 +35,7 @@ ng build --base-href /myUrl/
 ### Bundling & Tree-Shaking
 
 All builds make use of bundling and limited tree-shaking, while `--prod` builds also run limited
-dead code elimination via UglifyJS.
+dead code elimination via Terser.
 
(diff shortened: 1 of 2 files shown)


alan-agius4:
Thanks for the PR. However, the documentation here is correct as this is strictly for version 6. Version 7 of the document has been moved to https://angular.io/guide/build

There is currently an open PR to improve the overall description https://github.com/angular/angular-cli/pull/12741

cexbrayat:
@alan-agius4 Ha right, sorry.

Maybe this is still interesting for the README of build-optimizer which still mention UglifyJS though?

alan-agius4:
If you want to go ahead a remove it that it’s fine by me. Though I suggest not to mention the brand or tool that can be used as strictly speaking it can work with any js optimization tool.

For instance instead of `UglifyJS` you can use `JS optimization tools`

PR #12813, opened 2018-10-30
docs: production build configuration for app-shell

mhadaily:
ng run <Project>:app-shell will not trigger production build, to trigger production build, we should explicitly pass it as an argument.


mhadaily:
@filipesilva done

PR #12814, opened 2018-10-30
feat(@angular-devkit/schematics-cli): support console prompt

### LINKED ISSUE #12688: x-prompt doesn't work if use schematics-cli
```
- [x] feature request
```

### Repro steps

For an example:
schematics @schematics/angular:component

Error: Schematic input does not validate against the Schema: {}
Errors:

But it works when I use ng g сommand
ng g @schematics/angular:component

  Data path "" should have required property 'name'.

noelmace:
Support schematics prompt (x-prompt) using inquirer.
Reproduce the same behavior introduced in @angular/cli by 781f9b404c263.

close #12688


noelmace:
WIP: dryRun isn't true by default when using a local schematic. As this is a regression, I'm investigating this.

noelmace:
I think this would be a good idea to define this default promptProvider in @angular-devkit/core, as we do with the default indent logger. Here the problem here would be to add inquirer as a dependency of core, but I think we could delete it (even I wouldn't have the time right now to do it in this PR, I could do it later). What do you think ?

packages/angular_devkit/schematics_cli/bin/schematics.ts
```diff
@@ -109,6 +152,7 @@ export async function main({
   const dryRun: boolean = argv['dry-run'] === null ? debug : argv['dry-run'];
   const force = argv['force'];
   const allowPrivate = argv['allow-private'];
+  const interactive: boolean = argv['interactive'] !== null || argv['interactive'];
```
alan-agius4: Would it better to add the `interactive` in https://github.com/angular/angular-cli/blob/d8764e103de8292845a4acc366d9e036aaa65484/packages/angular_devkit/schematics_cli/bin/schematics.ts#L336-L363

Add a a default to `true` and the usage here would be ` const interactive: boolean = argv['interactive'];

noelmace: Done. Thx for the feedback.

PR #12824, opened 2018-10-31
Stop karma gracefully, test watch mode, and check syntactic diagnostics on rebuilds

filipesilva:
Followup to karma-runner/karma#3153

### DIFF
--- a/packages/angular_devkit/build_angular/src/karma/index.ts
+++ b/packages/angular_devkit/build_angular/src/karma/index.ts
@@ -97,6 +97,7 @@ export class KarmaBuilder implements Builder<KarmaBuilderSchema> {
           // When this workaround is removed, user projects need to be updated to use a Karma
           // version that has a fix for this issue.
           toJSON: () => { },
+          logger: this.context.logger,
         };
 
         // TODO: inside the configs, always use the project root and not the workspace root.
--- a/packages/angular_devkit/build_angular/src/karma/index.ts
+++ b/packages/angular_devkit/build_angular/src/karma/index.ts
@@ -108,15 +108,14 @@ export class KarmaBuilder implements Builder<KarmaBuilderSchema> {
 
         // Complete the observable once the Karma server returns.
         const karmaServer = new karma.Server(karmaOptions, () => obs.complete());
-        karmaServer.start();
+        const karmaStartPromise = karmaServer.start();
 
         // Cleanup, signal Karma to exit.
         return () => {
-          // Karma does not seem to have a way to exit the server gracefully.
-          // See https://github.com/karma-runner/karma/issues/2867#issuecomment-369912167
-          // TODO: make a PR for karma to add `karmaServer.close(code)`, that
-          // calls `disconnectBrowsers(code);`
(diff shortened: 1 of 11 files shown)


packages/angular_devkit/build_angular/src/karma/index.ts
```diff
@@ -108,15 +109,14 @@ export class KarmaBuilder implements Builder<KarmaBuilderSchema> {
...
 
         // Cleanup, signal Karma to exit.
         return () => {
-          // Karma does not seem to have a way to exit the server gracefully.
-          // See https://github.com/karma-runner/karma/issues/2867#issuecomment-369912167
-          // TODO: make a PR for karma to add `karmaServer.close(code)`, that
-          // calls `disconnectBrowsers(code);`
-          // karmaServer.close();
+          // Karma only has the `stop` method start with 3.1.1, so we must defensively check.
+          if (karmaServer.stop && typeof karmaServer.stop === 'function') {
```
alan-agius4: But isn’t karma 3.1.1 are direct dependency of build-angular? So what case should karmaServer.stop doesn’t exist? Apart from if someone has yarn resolutions.

packages/ngtools/webpack/src/gather_diagnostics.ts
```diff
@@ -9,6 +9,11 @@ import { Diagnostic, Diagnostics, Program } from '@angular/compiler-cli';
 import * as ts from 'typescript';
 import { time, timeEnd } from './benchmark';
 
+export const enum DiagnosticMode {
+  All,
```
hansl: Is const needed? Also, you could use flags so it's possible to make combinations. Unless this is meant to be used as an internal only API, which it doens't look like.

Something like:

```
export enum DiagnosticMode {
  Syntactic = 1 << 0,
  Semantic = 1 << 1,

  All = Syntactic | Semantic,
  Default = All,
}
```

What do you think?

packages/ngtools/webpack/src/gather_diagnostics.ts
```diff
@@ -52,34 +58,44 @@ export function gatherDiagnostics(
     }
   }
 
+  const gatherSyntacticDiagnostics = mode == DiagnosticMode.All || mode == DiagnosticMode.Syntactic;
```
hansl: With the above you can just use `mode & DiagnosticMode.Syntactic != 0`

filipesilva: Updated to `(mode & DiagnosticMode.Syntactic) != 0` because `!=` seems to have precedence over `&`.

packages/ngtools/webpack/src/gather_diagnostics.ts
filipesilva: I think that's much better really. Updated, thank you!

packages/ngtools/webpack/src/gather_diagnostics.ts
hansl: JavaScript!

PR #12843, opened 2018-11-01
feat(@angular/cli): allow flags to have deprecation

hansl:
The feature comes from the "x-deprecated" field in schemas (any schema that is used
to parse arguments), and can be a boolean or a string.

The parser now takes a logger and will warn users when encountering a deprecated
option. These options will also appear in JSON help.


hansl:
@alan-agius4 It is exposed to JSON help; the `Option` interface is what's used to show the help.

vikerman:
@hansl - Please rebase

hansl:
@kyliau @vikerman done.

PR #12847, opened 2018-11-02
fix(@angular-devkit/build-optimizer): update typescript to 3.1.6

mattlewis92:
The build optimiser is still stuck on ts 3.1.5 so all our builds still take way longer on angular 7, hopefully this should be the last thing to solve the ng 7 perf issues 🎉


alan-agius4:
Superseeded by https://github.com/angular/angular-cli/pull/12849

PR #12854, opened 2018-11-02
ci: fast track docs-preview snapshot

hansl:
Remove all jobs from the docs-preview branch, and only do `publish_snapshot`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -168,6 +168,10 @@ workflows:
       - build:
           requires:
             - install
+          filters:
+            branches:
+              ignore:
+              - /docs-preview/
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -168,6 +168,10 @@ workflows:
       - build:
           requires:
             - install
+          filters:
+            branches:
+              ignore:
+              - /docs-preview/
```
filipesilva: Nit: can you change this to just `- docs-preview`? In CircleCI these regexes must match the entire string so it shouldn't be a problem, but it doesn't really need to be a regex in the first place.

PR #12857, opened 2018-11-02
enhancement: improve performance of move()

FrozenPandaz:
## Current Behavior

`move` goes through every single file in the workspace which includes `node_modules` which could possible be nested via tools like Lerna. Also for very large workspaces the same issues exist for source code.

## Expected Behavior

Credits to @JamesHenry:

`move` should operate on just the sub tree so it will be fast even for huge workspaces.


vikerman:
@hansl - Please add merge target

JamesHenry:
🙌

clydin:
Just a comment that mutating within a visitor is probably not something that should be encouraged and probably should break.  I know it is already there but walking the entire tree before actually visiting has performance implications in light of the ability to short circuit as well as logic issues depending on the mutation of the original tree.

We should also consider deprecating `exists` as it is ambiguous regarding directories (`fileExists`/`directoryExists`?).

PR #12871, opened 2018-11-05
fix(@schematics/update): improve npmrc discovery

clydin:
Supports the following:
* global npmrc
* user npmrc
* project npmrc

yarn variants of the above are also supported if yarn is the CLI's package manager.


packages/angular/cli/models/schematic-command.ts
```diff
@@ -61,7 +61,7 @@ export interface BaseSchematicSchema {
 export interface RunSchematicOptions extends BaseSchematicSchema {
   collectionName: string;
   schematicName: string;
-
+  additionalOptions?: { [key: string]: {} };
```
alan-agius4: Nit: Why not `Record<string, {}>`, It's not the first time I've seen this.

clydin: Habit more than anything else; `Record` is fairly new.  I also tend to group `Record` in with `Pick`/`Extract` as more of type _reconstruction_ operators than anything else.  `Record` is very interesting in combination with string enums or string literal types.

In this case it probably is cleaner to use it though.

PR #12873, opened 2018-11-06
fix(@angular-devkit/core): fix 'this' check

vikerman:
This causes TypeError in Google when used with prompts. Just fixing the check.


clydin:
Running in strict mode should also fix it as well.

vikerman:
@clydin - Ah this explains why this only happens inside Google. We need this in Google where we don't seem to be running strict mode.

PR #12908, opened 2018-11-09
fix(@angular/cli) 1st project consuming arguments to a multiTarget command

terencehonles:
- Angular CLI version latest (tested on 7.1.0-beta.1 & master 4059fe83b485edcf42b8a795ea74289f1531688e)
- use case angular.json w/ 2 projects and command (i.e. karma w/ `{singleRun: false}` and command line `ng test --no-progress --no-watch` which will only send the options to the first project and the second one will hang (because of `--no-watch` not be given  to the second project). Visibly you can notice that the first will not show any build progress (as expected from `--no-progress`), but the second one will.


alan-agius4:
This should be addressed in https://github.com/angular/angular-cli/pull/12673

As the original issue is that the options are mutated within the parser logic which is not correct.

PR #12920, opened 2018-11-10
ci: run minimal tests on PRs and original set on release branches

clydin:


clydin:
Couldn't find a way around that.  adding master to the `only` block caused all PRs to run the full set.

filipesilva:
If there's no way to do this with appveyor we might need to rethink how we the caretaker merges PRs... otherwise our coverage will suffer.

hansl:
@clydin Could you take a minute to see if we could look at some environment variable to know which branch we're on (vs PR)?

Issue #12163, opened 2018-09-05
Interactive input dialog


alexeagle:
_From @dansanderson on June 14, 2017 22:40_

The command line tool should have two running modes: an interactive mode and a batch mode. In batch mode, all inputs for the schematic are specified on the command line, and input validation errors terminate the tool with a message. In interactive mode, a schematic can be invoked with or without arguments on the command line, and the tool begins an interactive dialog describing the behavior of the schematic, and presents and prompts for values for some or all inputs. Inputs with defaults or whose values were provided by another source are presented for confirmation ("press enter to accept default"-style). After all prompts have been presented, show a confirmation screen with all inputs and accepted values, and allow selective adjustment of values.

Schematic properties will need short and long descriptions and a value type (string, boolean, maybe number).

The confirmation screen can come after all rules have been evaluated to report conflicts and errors and allow the user to adjust inputs interactively to resolve the conflicts. After confirmation, the schematic can execute side effects (FSSink, task execution).

(Ask me for a more detailed design or demo. I have a full implementation in a related tool.)

_Copied from original issue: angular/devkit#11_

PonnarV:
### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### ### 

> 

> [``]([url]([url]([url]([url]([url]([url](url)))))))

> 

- 

- 

- 

- 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] 

- [ ] ### ### ### ### ### ### ### ### 

> 

> 

> 

> 
        #########
- [ ] 

- [ ] 

1. 

> 

> [****](url)

> 

> [`****`](

- url

)

cyrilletuzi:
@alexeagle How do you choose the batch mode in CLI v7?

Interactivity can be a problem for tools like Angular Schematics extension for VS Code (see cyrilletuzi/vscode-angular-schematics#14) (and I suppose the Angular Console will be concerned as well).

clydin:
There's no current batch mode.  But you can turn off prompts with `--no-interactive`/ `--interactive false`.

cyrilletuzi:
@clydin Thanks for the answer. I have a few more questions to be sure about how to handle that:

1. Do you confirm that `no-interactive` can be used with all commands? (this option doesn't seem to appear in the commands schemas)

2. Is `no-interactive` backward-compatible (no error if still in CLI v6 for example)?

[...]

PR #12927, opened 2018-11-11
prompt support for value type number

noelmace:
Permit to support x-prompt for a schematic parameter definition if its type is set to number.

Previously, if used with type 'number', the prompt was trap in a loop as input never was considered correct.

The part on the cli will also need to be reproduced on schematics-cli after #12814.

Related to #12163.

### angular-devkit/core & ajv
Type number wasn't set, and the value wasn't converted to a number if needed in the prompt validator.

### angular-devkit/cli & inquirer
Inquirer doesn't convert it result to a number even if definition type is set to 'number' following https://github.com/SBoudrias/Inquirer.js/pull/663

## Reproduction steps

You can reproduce the previous behavior by using the @angular-buddies/prettier@1.0.0-alpha.2:prettier-config schematic.

https://github.com/angular-buddies/angular-buddies/blob/v1.0.0-alpha.2/packages/prettier/src/config/schema.json#L14-L19

### DIFF
--- a/packages/angular/cli/models/schematic-command.ts
+++ b/packages/angular/cli/models/schematic-command.ts
@@ -292,6 +292,7 @@ export abstract class SchematicCommand<
 
     if (options.interactive !== false && process.stdout.isTTY) {
       workflow.registry.usePromptProvider((definitions: Array<schema.PromptDefinition>) => {
+        const numberQuestions: string[] = [];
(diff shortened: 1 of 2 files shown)


clydin:
Closing as the CLI currently supports the use of a schema property's type when validating and returning the value from a prompt.  This includes both the `number` and `integer` types.
However, thank you for the contribution.

PR #12954, opened 2018-11-13
feat(@angular/cli): initial minimal BEP support

clydin:


alan-agius4:
I think this new option should be also added https://github.com/angular/angular-cli/blob/master/packages/angular/cli/lib/config/schema.json

packages/angular_devkit/build_angular/src/browser/schema.d.ts
```diff
@@ -227,6 +227,8 @@ export interface BrowserBuilderSchema {
    * Output profile events for Chrome profiler.
    */
   profile: boolean;
+
+  bep?: string;
```
alan-agius4: Maybe a jsdoc comment here?

packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts
```diff
@@ -57,11 +59,44 @@ export class WebpackDevServerBuilder implements Builder<WebpackDevServerBuilderS
...
+        this.context.logger.warn('BEP support is experimental and subject to change.');
+      }
+
+      webpackCompiler.hooks.beforeCompile.tapPromise('build-webpack', async () => {
+        if (this.bepPath) {
+          await appendEvent(
+            this.context.host,
+            resolve(normalize(process.cwd()), normalize(this.bepPath)),
+            {
+              id: BuildEventId.BuildStarted,
```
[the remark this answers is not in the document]
clydin: The CLI itself has no knowledge of whether a _builder_ is actually going to actually build or just run a task (e.g., lint).  The CLI is also not currently designed to accept more than one (the last) event from a builder.  This limitation would prevent watch/serve modes of operation from emitting build events per underlying build.

packages/angular_devkit/build_webpack/src/bep.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Path, virtualFs } from '@angular-devkit/core';
+
+
+export type BuildEvent = BuildStarted | BuildFinished;
```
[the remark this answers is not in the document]
clydin: I was hesitant to use it as it generates about ~20K lines (~1MB) of code.

packages/angular/cli/models/architect-command.ts
```diff
@@ -168,14 +170,41 @@ export abstract class ArchitectCommand<
...
+      BepGenerator.createBuildStarted(command),
+    );
+
+    let last = 1;
+    await this._architect.run(configuration, { logger: this.logger }).forEach(event => {
+      last = event.success ? 0 : 1;
+
+      appendEvent(
+        buildEventLog,
+        BepGenerator.createBuildFinished(last),
```
alexeagle: I'm a bit confused, if this happens in a loop, does it mean we produce more than one buildfinished event? I'd expect there is always a single one

packages/angular/cli/utilities/bep.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function appendEvent(
+  path: string,
+  event: BuildEventMessage,
+): void {
+  const exists = fs.existsSync(path);
+
+  let content = '';
+  if (exists) {
+    content = fs.readFileSync(path, 'utf8');
```
alexeagle: this seems wasteful, can't we just open the file in append mode rather than read its contents?
Even better, could we hold the filehandle open and just write/flush to the buffer?

packages/angular/cli/models/architect-command.ts
clydin: In watch/dev-server mode there are potentially multiple builds per command invocation.

packages/angular/cli/utilities/bep.ts
clydin: yes. it is.  However, currently the frequency of writes is quite minimal.  I was going to eventually create a `BepJsonWriter` class that encapsulated the logic as you described along with the message generation helpers.

clydin: `BepJsonWriter` introduced to better encapsulate the message writing.

tests/legacy-cli/e2e/tests/experimental/bep.ts
```diff
@@ -0,0 +1,10 @@
+import { ng } from '../../utils/process';
+import { expectFileToMatch } from '../../utils/fs';
+
+
+export default async function() {
+  await ng('build', '--build-event-log', 'bep-log.txt');
+
+  await expectFileToMatch('bep-log.txt', '{"id":{"started":{}},"started":{"command":"build",');
+  await expectFileToMatch('bep-log.txt', '{"id":{"finished":{}},"finished":{"finish_time_millis":');
```
alexeagle: it would be nice to have a golden file test too, so there's a specimen of the BEP file that we create which is guaranteed up-to-date

PR #12955, opened 2018-11-13
feat(@angular/cli): option to build and run only specified spec files

### LINKED ISSUE #3603: Run A Single Unit Test
### OS?
Windows 8, Mac OSX El Capitan

### Versions.
angular-cli: 1.0.0-beta.19-3
node: 6.3.0
os: win32 x64

This is a feature request. The ability to run a single `.spec.ts` file would be excellent, when you guys get the chance. I recently switched from using Angular1 in Visual Studio where one could use the Test Explorer or the Chutzpah extension to run a test on a single component. Working on a large enterprise application, it's a huge pain to run the entire unit test suite (which takes a few minutes) every time I refactor a single component (and thus only need to run that component's tests).

FDIM:
A new `--spec` option is added. It is a glob/pattern that specifies which specs should be included in the build.
e.g. `ng test --spec=app/services/auth*` (`.spec.ts` is added automatically if missing)

When karma is already running, selected spec can be updated by running `ng test --spec=app/login* --spec-update`

With such option it is also trivial to add custom vscode task to be able to run currently open spec file.

Behind the scenes a new `test.generated.ts` (could be temp?) is saved next to original `test.ts` which instead of using `require.context` imports all matching files. This gives quite a big boost in the project at work (at least 10s on each iterative build!)

Fixes #3603

Motivation: It is a pain in a big application to wait until entire app is rebuilt just to run single `fit` block. Takes almost a minute with 4000+ tests and barely works when code coverage is enabled. 

Missing: docs(?) and tests 

Disclaimer: this is quick and dirty solution that can be improved a lot but some feedback before making it nice would be great!

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
@@ -7,7 +7,7 @@
  */
 // tslint:disable
 // TODO: cleanup this file, it's copied as is from Angular CLI.
-
+import * as fs from 'fs';
 import * as path from 'path';
 import * as glob from 'glob';
 import * as webpack from 'webpack';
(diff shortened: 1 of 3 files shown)


mgechev:
Closing it due to inactivity. Please reopen if you come back to it.

FDIM:
@mgechev I've updated the branch - refactored it a bit, added tests and I think it's ready for review

PR #12956, opened 2018-11-13
docs: correct target in run command docs

cexbrayat:
@jbogarthyde I think this commit https://github.com/angular/angular-cli/commit/1c39f54a24eb0893735d3cf87e8f3d6403bf8ae2 recently merged references a target "server" that doesn't exist instead of the "serve" target.


alan-agius4:
Thanks for this though this is superseeded by 
https://github.com/angular/angular-cli/pull/12958 as there are two typos.

Issue #12731, opened 2018-10-24
Update from 6 to 7, the 'dist' root directory got bloated after 'ng build --prod'


Teamop:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Command (mark with an `x`)
```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

angular-cli 7.0.2

### Repro steps

I'm using `ag-grid` and `ag-grid-angular` with `angular-cli`, and I also need to do some customization for the theme so I need to do that with scss file provided by the `ag-grid`.
With CLI v6, the dist folder like this:
![image](https://user-images.githubusercontent.com/8519685/47423833-a6f79d80-d7b8-11e8-86cc-9ea15a083a46.png)
After I updated to v7, like this(just part):
![image](https://user-images.githubusercontent.com/8519685/47423878-ca224d00-d7b8-11e8-8106-9abe4a261119.png)

Here, we can see the SVG files are all in the root directory. 
Then I did some research, there SVG files are required in the ag-grid style, but before in the v6, they are all got inlined. But with v7, they are all just individual files. I believe this change is #12267. And based on this change, **these files now are in the right place**. But compared with the 6 version, seems a little messy.
 

### Desired functionality

These SVG files should be in a sub folder, like 'dist/images/xx'

### Mention any other details that might be useful

Maybe related to #6315

alan-agius4:
Indeed this is by design that SVG are no longer inlined.

I am going to close this in favor of #6315 to move assets into sub-directories.

bampakoa:
I have experienced the same issue but with PNG files that are referenced inside vendor CSS files such as 

`background-image: url("../img/remove-icon-small.png")`

In my case, I do not use any of these so the **dist** directory is bloated with unnecessary files. I understand that this is by design but are you planning to provide a workaround or should I stick to remove them using a custom automation?

Issue #6315, opened 2017-05-15
How to sort output chunks, font and other files in appropiate directories


benpoe:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [ *] feature request
```

### Versions.
@angular/cli: 1.0.1
node: 7.7.2
os: win32 x64
@angular/animations: 4.1.0
@angular/common: 4.1.0
@angular/compiler: 4.1.0
@angular/core: 4.1.0
@angular/forms: 4.1.0
@angular/http: 4.1.0
@angular/platform-browser: 4.1.0
@angular/platform-browser-dynamic: 4.1.0
@angular/platform-server: 4.1.0
@angular/router: 4.1.0
@angular/cli: 1.0.1
@angular/compiler-cli: 4.1.0

### Repro steps.
npm run build:prod:aot

### The log given by the failure.
none

### Desired functionality.
When doing a "npm run build" all js-chunks, bundles, fonts are thrown into the "dist"-directory.
Using webpack I'd have sorted them into appropiate directories.  js in "js", fonts in "fonts" etc.

```

              {
                  test: /\.woff(2)?(\?v=.+)?$/,
                  //    use: 'url-loader?limit=10000&mimetype=application/font-woff'
                  use: 'url-loader?limit=10000&mimetype=application/font-woff&name=**assets/fonts/**[name].[ext]'
              },

              {
                  test: /\.(ttf|eot|svg)(\?v=.+)?$/,
                  //use: 'file-loader'
                  loader: 'file-loader?name=**assets/fonts/**[name].[ext]'
              },
```

```
        output: {
            path: helpers.root('dist'),
            filename: '**js/**[name].[chunkhash].bundle.js',
            sourceMapFilename: '**js/**[name].[chunkhash].bundle.map',
            chunkFilename: '**js/**[id].[chunkhash].chunk.js'
        },
```

It'd be great if angular-cli would have an option to configure the output too.

benpoe:
It'd be handy if .angular-cli.json has the option to sort filechunks.

For example the angular-cli.json could look like this:
 "scripts": [
        "../node_modules/jquery/dist/jquery.js",
        "../node_modules/easy-pie-chart/dist/jquery.easypiechart.js",
        "../node_modules/jquery-slimscroll/jquery.slimscroll.js",
        "../node_modules/tether/dist/js/tether.js",
        "../node_modules/bootstrap/dist/js/bootstrap.js",
[...]

PR #12967, opened 2018-11-15
 feat(@angular-devkit/build-angular): add option to allow outputting css assets to a different folder

alan-agius4:
@clydin I was unable to find the issue in our issue tracker. If you have a link that would be great.

Also, I was unsure if we should add this to absolute paths. I reckoned not.

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -623,6 +623,10 @@
               "type": "string",
               "description": "Path where output will be placed."
             },
+            "resourcesOutputPath": {
+              "type": "string",
+              "description": "The path where style resources will be placed, relative to outputPath."
+            },
             "aot": {
               "type": "boolean",
               "description": "Build using Ahead of Time compilation.",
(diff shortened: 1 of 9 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/plugins/postcss-cli-resources.ts
```diff
@@ -117,7 +119,7 @@ export default postcss.plugin('postcss-cli-resources', (options: PostcssCliResou
 
         const outputPath = interpolateName(
           { resourcePath: result } as webpack.loader.LoaderContext,
-          filename,
+          path.join(assetsOutputPath, filename),
```
clydin: do the join after the interpolate and use `path.posix.join`.  The option's value shouldn't be altered by the interpolation.

packages/angular/cli/lib/config/schema.json
```diff
@@ -623,6 +623,10 @@
               "type": "string",
               "description": "Path where output will be placed."
             },
+            "assetsOutputPath": {
```
clydin: `resources` instead of `assets`.  Assets already have a distinct meaning and behavior.

Teamop:
Maybe related to #6315， #12731

PR #12969, opened 2018-11-15
Direct component template loading in AOT

clydin:
Currently within the CLI, a full webpack child compilation is created and executed to load a template via the `raw-loader` in AOT.  The end result is equivalent to reading the file directly.  This PR eliminates the overhead of the extra webpack compilation for the CLI by adding an option to the `@ngtools/webpack` plugin (defaulting to false to retain existing behavior) that reads the HTML component templates directly from the compiler host if enabled.  This option is then enabled for the CLI.


clydin:
Updated the readme with the option and information about when not to enable it.

liujingbreak:
It seems to be hard for a customzied/extend Angular cli builder to turn this option off.

PR #12970, opened 2018-11-15
Jobs

hansl:
There is a README added which contains how to use the API.


packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,207 @@
...
+  },
+  output: { type: 'boolean' },
+});
+
+// Extending makes sure the input and output matches the dispatcher.
+const npmInstall = jobs.createJob(/* ... */, { name: 'npm-install', extends: dispatcher });
+const yarnInstall = jobs.createJob(/* ... */, { name: 'yarn-install', extends: dispatcher });
+const pnpmInstall = jobs.createJob(/* ... */, { name: 'pnpm-install', extends: dispatcher });
+
+declare const registry: jobs.SimpleJobRegistry;
```
alan-agius4: `declare` here is not correct, since you are not writing a declaration file.

hansl: This is an example in the README. It's meant to provide context on the type, not be copy-pasted as is.

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,207 @@
+# Jobs
+
+Jobs is the Angular DevKit subsystem for scheduling and running generic functions with clearly 
+typed inputs and outputs. A Job is
```
Brocco: "A Job is" is not a sentence, would ❤️ to see a longer definition here.

hansl: Clarifying this. I'm reworking the README a bit and by EOD I'll have a proper description to replace that non-sentence.

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+scheduler.schedule('add', [1, 2, 3, 4])
+    .output.subscribe(x => console.log(x));  // Will output 10.
+```
+
+This is a lot of boilerplate, so we made some helpers to improve readability and manage input 
+and output automatically;
+
+```typescript
+// Add is a JobHandler function, like the above.
+export const add = jobs.createJob<number[], number>(
```
filipesilva: Does `createJob` create the input and output schemas from the generics?

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+      },
+  },
+);
+
+// Get a hold of a scheduler that refers to the job above.
+declare const scheduler: jobs.Scheduler;
+
+const job = scheduler.schedule('count');
+job.channels['side'] && job.channels['side'].subscribe(x => console.log(x));
+const c = job.getChannel<string>('side', { type: 'string' });
```
filipesilva: Are these two different ways to subscribe to the same channel?

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+if (c) {
+  c.subscribe((x: string) => console.log(x));
+}
+```
+
+## Job Extension
+A Job can extend other jobs, meaning their inputs and outputs have to be a superset of the job 
+extended. This is useful for making compatible jobs that can be replaced by others.
```
filipesilva: I think this section really needs some examples and more details. Things like a job extending each other, what it means to replace a job, and a replacement example.

Another thing that I don't understand is the usage pattern for jobs that have a superset of the input and outputs that is not. 

Consider I have a JobFoo, with known inputs and outputs. JobBar extends JobFoo, and has additional inputs and outputs.

If I want to transparently replace JobFoo with JobBar, I can only rely on using the known input/output for JobFoo. Thus the additional input/output that JobBar has will always use defaults.

I'd like to understand how this impacts the use cases that Schematics and Architect have for Jobs.

For instance, having a generic `node_install` job that is ultimately replaced by `npm_install` or `yarn_install`, or some other. 

Should the base `node_install` have only the intersection of the input/output of the (known) possible replacements? Or should it have some input/output that we expect to be ignored by some of the replacements?

Replacements are one of the key requirements of Jobs so it's important that this be clear.

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Job Groups
+Groups are an additional way to synchronize and execute logic on scheduling the job, from the 
+side of the scheduler. This logic will always apply in the thread the job is scheduled, and not
+where the job is created/registered.
+
+Job groups are a job helper that redirect to different jobs. To create a job group, use the
```
filipesilva: `Job groups are a job helper that redirect to different jobs.` this is a really confusing way of explaining what job groups are. I guess it's like a dispatcher?

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+  input: {
+    properties: {
+      moduleName: { type: 'string' },
+    },
+  },
+  output: { type: 'boolean' },
+});
+
+// Extending makes sure the input and output matches the group.
+const npmInstall = jobs.createJob(/* ... */, { name: 'npm-install', extends: group });
```
filipesilva: Can jobs extend both other jobs, and groups?

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+// Default to npm.
+group.setDefaultDelegate(npmInstall.name);
+// If the user is asking for yarn over npm, uses it.
+group.addConditionalDelegate(() => userWantsYarn, yarnInstall.name);
+```
+
+## Execution Strategy
+
+Jobs are always run in parallel and will always start, but many helper functions are provided
```
filipesilva: What does `will always start` mean? Is that that they always start immediately?

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+      channels: {
+        'side': { type: 'string', const: 'more' },
+      },
+  },
+);
+
+// Get a hold of a scheduler that refers to the job above.
+declare const scheduler: jobs.Scheduler;
+
+const job = scheduler.schedule('count');
```
filipesilva: I think this job is missing the input.

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+group.addConditionalDelegate(() => userWantsYarn, yarnInstall.name);
+```
+
+## Execution Strategy
+
+Jobs are always run in parallel and will always start, but many helper functions are provided 
+when creating a job to help you control the execution strategy;
+
+1. `serialize()`. Multiple runs of this job will be queued with each others.
+1. `reuse(replayEvents = false)`. Jobs with this strategy will reuse an already running job. If
```
filipesilva: I don't understand the reuse strategy. A job has a single input. How can a running job receive more inputs?

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+  ),
+);
+```
+
+Strategies can be reused to synchronize between jobs. For example, given jobs `jobA` and `jobB`, 
+you can reuse the strategy to serialize both jobs together;
+
+```typescript
+const strategy = jobs.strategy.serialize();
+const jobA = strategy(jobs.createJob(...));
```
filipesilva: Can I remove a job from a strategy?

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+const strategy1 = jobs.strategy.serialize();
+const strategy2 = jobs.strategy.serialize();
+const jobA = strategy1(jobs.createJob(...));
+const jobB = strategy1(strategy2(jobs.createJob(...)));
+const jobC = strategy2(jobs.createJob(...));
+```
+
+# Scheduling Jobs
+Jobs can be scheduled using a `Scheduler` interface, which contains a `schedule()` method. The
+method takes a name, an input and a list of options.
```
filipesilva: What are these options?

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Synchronizing and Dependencies
```
filipesilva: Isn't synchronization managed by the execution strategy? 

Are these different things, and if so, what are the different use cases?

hansl:
Sorry guys I've pushed instead of pulled my branch this morning. This is going to be one of those days... 😩

packages/angular_devkit/core/node/scheduler.ts
```diff
@@ -0,0 +1,73 @@
...
+    const handler = pkg[exportName || 'default'];
+    if (!handler) {
+      return null;
+    }
+
+    const input = ['boolean', 'object'].includes(typeof pkg.input)
+      ? pkg.input : handler.input !== undefined ? handler.input : true;
+    const output = ['boolean', 'object'].includes(typeof pkg.output)
+      ? pkg.output : handler.output !== undefined ? handler.output : true;
+    const channels = ['boolean', 'object'].includes(typeof pkg.channels)
```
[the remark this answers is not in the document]
hansl: Done. And added a `json.schema.isJsonSchema()` to replace that weird check.

packages/angular_devkit/core/src/jobs/api.ts
```diff
@@ -0,0 +1,421 @@
...
+  /**
+   * Outputs of this Job. This is sugar for
+   * `this.pipe(filter(x => x.kind == JobEventKind.Output), map(x => x.output))`.
+   */
+  readonly output: Observable<O>;
+
+  /**
+   * Last output for this Job. This is sugar for `output.pipe(last()).toPromise()`.
+   */
+  readonly promise: Promise<O | undefined>;
```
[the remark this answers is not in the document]
hansl: There might not be a last value, if you don't output anything.

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,207 @@
...
+  input: {
+    properties: {
+      moduleName: { type: 'string' },
+    },
+  },
+  output: { type: 'boolean' },
+});
+
+// Extending makes sure the input and output matches the dispatcher.
+const npmInstall = jobs.createJob(/* ... */, { name: 'npm-install', extends: dispatcher });
```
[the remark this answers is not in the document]
hansl: Yeah I'm starting to think that `extends` as a concept is specific to the registry and should not be part of the job declaration. It was deeper integrated with the API and then I moved it slowly out of it. I think I should remove it entirely (and keep it as a field in the `register()` functions).

packages/angular_devkit/core/src/jobs/api.ts
```diff
@@ -0,0 +1,421 @@
...
+/**
+ * Job Name is just a string (needs to be serializable).
+ */
+export type JobName = string;
+
+/**
+ * Metadata associated with a job.
+ */
+export interface JobDescription extends JsonObject {
+  readonly name: JobName;
```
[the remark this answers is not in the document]
hansl: I'm naming this as a hint that everywhere that type is used should be the job name itself and not just a random string (like a message). This is used as a hint ot users, not actually typing.

packages/angular_devkit/core/src/jobs/create-job.ts
```diff
@@ -0,0 +1,204 @@
...
+            }),
+            () => subject.next({
+              kind: JobEventKind.ChannelComplete, description, name,
+            }),
+          );
+
+          return s;
+        }),
+        inputChannel: inputChannel.asObservable(),
+        logger,
```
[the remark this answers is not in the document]
hansl: Done.

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+typed inputs and outputs. A Job is code (ie. a function) that can be scheduled, synchronized and
+schedule other jobs. Those Jobs can be run in parallel or follow an execution strategy. They can
+also be extended to allow for transformations, combination and refactoring.
+
+The whole API is serializable, allowing someone to use a Node Stream or message channel to
+communicate between the job and the job scheduler.
+
+## Input, Output and Channels
+Jobs can emit multiple outputs, and receives an input when first started. In addition, jobs can
+listen to an input channel, and multiple output channels. Those channels are asynchronous
```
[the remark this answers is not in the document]
hansl: Output Channels are not listened to. I'll fix this.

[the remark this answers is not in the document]
hansl: From the Job perspective, they're output, so they're really `Observer<T>`.

packages/angular_devkit/core/src/jobs/README.md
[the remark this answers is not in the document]
hansl: Going to go with `implements`, as it's more an interface related concept.

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+change. A user can listen to these events, or use the `state` member which will correspond to
+one of those.
+
+## `Job<OutputType>` Object
+The `Job` object is returned when scheduling a job. It contains multiple members and utilities to 
+ease the usage of jobs.
+
+1. `id`. A unique symbol that can be used as a Map key.
+1. `description`. The description of the job from the scheduler. See `JobDescription` object.
+1. `input`. An `Observer` that can be used to send validated inputs to the Job itself.
```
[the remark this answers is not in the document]
hansl: I renamed `input` to `argument` to differentiate inputs from the input channel and argument which are passed in when scheduling the job.

packages/angular_devkit/core/src/jobs/README.md
hansl: It was removed. It doesn't make sense.

hansl: Essentially it was a "Never rerun the job, just reuse the first run".

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,319 @@
...
+
+If a Job was an executable, the input would be the arguments on the command line, the input 
+channel would be STDIN, the output channel would be STDOUT and STDERR (through Observable errors),
+and the channels would be additional pipes out.
+
+## LifeCycle
+A Job goes through multiple LifeCycle events before its completion;
+1. `JobState.Queued`. The job was queued and is waiting to start.
+1. `JobState.Created`. The job's dependencies are done running, and the job has been started.
+1. `JobState.Started`. The job has been started. This should be triggered by a JobEventStart event,
```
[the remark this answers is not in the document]
hansl: Clarified.

packages/angular_devkit/core/src/jobs/README.md
hansl: Clarified.

packages/angular_devkit/core/src/jobs/README.md
hansl: These are different things. The difference is that strategy is set by the actual Job code, while this synchronization is done by the person scheudling the job. Clarfiied.

hansl:
Moved to https://github.com/angular/angular-cli/pull/13029.

PR #12972, opened 2018-11-15
feat(@angular-devkit/schematics-cli): add prompt support using Inquirer

hansl:


skatemaza:
With the @angular-devkit/schematics-cli@0.11.0 installed globally it still does not work for me when calling `schematics @schematics/angular:component`

It results into:
```
Error: Schematic input does not validate against the Schema: {}
Errors: Data path "" should have required property 'name'.
```

mmoats:
> With the @angular-devkit/schematics-cli@0.11.0 installed globally it still does not work for me when calling `schematics @schematics/angular:component`
> 
> It results into:
> 
> ```
> Error: Schematic input does not validate against the Schema: {}
> Errors: Data path "" should have required property 'name'.
> ```
> I work with a Win10 machine.

Pretty sure this didn't make it to the release, didn't pass all 13 checks.  See https://github.com/angular/angular-cli/compare/v7.1.0...master

PR #12983, opened 2018-11-17
Add prompt checkbox support

okeefem2:
Hi there! 

This is my first contribution, and actually my first contribution to an open source project of this size. Please let me know if there is anything I may have missed in terms of adding tests or could have done better etc.

Very excited to submit this and hope the community will find it useful!

Thanks,

Michael

----

PR description:

* Added general checkbox support using inquirer for schematics prompts
* Added checkbox 'implements' option to guard schematics to allow user to choose implementations (CanActivate, CanActivateChild or CanLoad)

Notes:

I Purposefully left out CanDeactivate because of the added complexity of needed a Component to implement. I was definitely considering that as a follow up to this PR, but since that would take a little more time on my part I wanted to get this out there first.


okeefem2:
I will check out those failed test/checks tonight when I get home!

PR #13003, opened 2018-11-20
WIP: feat(@angular-devkit/schematics): add support for jobs as tasks 

hansl:
The FileSystemEngineHostBase now has an adapter that understands Jobs as
if they were tasks. This keep the current API but allow tools to register
jobs as tasks. In 8.0 we will have to take the breaking change and clean
up tasks.
28cd458

### DIFF
--- a/packages/angular_devkit/core/src/jobs/README.md
+++ b/packages/angular_devkit/core/src/jobs/README.md
@@ -1,0 +1,205 @@
+# Jobs
+
+Jobs is the Angular DevKit subsystem for scheduling and running generic functions with clearly 
+typed inputs and outputs. A Job is 
+
+The whole API is serializable, allowing someone to use a Node Stream or message channel to 
+communicate between the job and the job scheduler.
+
+## Input, Output and Channels
+Jobs can emit multiple outputs, and receives an input when first started. In addition, jobs can
+listen to an input channel, and multiple output channels. Those channels are asynchronous
+JSON messages that can be typed.
+
+If a Job was an executable, the input would be the arguments on the command line, the input 
+channel would be STDIN, the output channel would be STDOUT and STDERR (through Observable errors),
+and the channels would be additional pipes out.
+
+# Creating Jobs
+
+A job is at its core a function with a Schema associated to it;
+
+```typescript
+import { Observable } from 'rxjs';
+import { jobs } from '@angular-devkit/core';
+
+const input = {
+  type: 'array', items: { type: 'number' },
+};
+const output = {
+  type: 'number',
+};
+
+export function add(input: number[]): Observable<jobs.JobEvent<number>> {
+  return new Observable(o => {
+    o.next({ kind: jobs.JobEventKind.Start });
+    o.next({
+      kind: jobs.JobEventKind.Output,
+      output: input.reduce((total, curr) => total + curr, 0),
+    });
+    o.next({ kind: jobs.JobEventKind.End });
+  });
+}
+
+// Register the job in a SimpleJobRegistry. Different registries have different API.
+declare const registry: SimpleJobRegistry;
+registry.register('add', add, {
+  input: input,
+  output: output,
+});
+
+// Calling the job, and logging its output.
+declare const scheduler: jobs.Scheduler;
+scheduler.schedule('add', [1, 2, 3, 4])
+    .output.subscribe(x => console.log(x));  // Will output 10.
+```
+
+This is a lot of boilerplate, so we made some helpers to improve readability and manage input 
+and output automatically;
+
+```typescript
+// Add is a JobHandler function, like the above.
+export const add = jobs.createJob<number[], number>(
+  input => input.reduce((total, curr) => total + curr, 0),
+);
+
+// Schedule like above.
+```
+
+You can also return a Promise or an Observable, as jobs are asynchronous. This helper will set 
+progress to 0, start and end events appropriately, and will pass in a logger. It will also manage
+channels automatically (see below).
+
+A more complex job can be declared like this:
+
+```typescript
+import { Observable } from 'rxjs';
+import { JobRegistry } from './api'
+
+// Show progress with each count. Output "more" in a channel.
+export const count = jobs.createJob<number, number>(
+  // Receive a context that contains additional methods, like progress, logger and channels.
+  (input: number, { progress, channels }) => new Observable<number>(o => {
+    let i = 0;
+    function doCount() {
+      o.next(i++);
+      progress(i / input);
+      channels['side'].next('more');
+  
+      if (i < input) {
+        setTimeout(doCount, 100);
+      } else {
+        o.complete();
+      }
+    }
+    setTimeout(doCount, 100);
+  }),
+  {
+      input: { type: 'number' },
+      output: { type: 'number' },
+      channels: {
+        'side': { type: 'string', const: 'more' },
+      },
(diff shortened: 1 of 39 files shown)


packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,205 @@
...
+  },
+  output: { type: 'boolean' },
+});
+
+// Extending makes sure the input and output matches the group.
+const npmInstall = jobs.createJob(/* ... */, { name: 'npm-install', extends: group });
+const yarnInstall = jobs.createJob(/* ... */, { name: 'yarn-install', extends: group });
+const pnpmInstall = jobs.createJob(/* ... */, { name: 'pnpm-install', extends: group });
+
+declare const registry: jobs.SimpleJobRegistry;
```
alan-agius4: `declare` here is not correct, since you are not writing a declaration file.

packages/angular_devkit/core/src/jobs/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+## SimpleJobRegistry
+A registry that accept job registration, and can also schedule jobs.
+
+## NodeModuleJobScheduler
+A scheduler that loads jobs using their node package names. These jobs need to use the
+`createJob()` helper and report their input/output schemas that way.
+
+```javascript
+declare const scheduler: NodeModuleJobScheduler;
```
alan-agius4: Same., there shouldn't be the `declare` keyword here.

packages/angular_devkit/core/src/jobs/create-job.ts
```diff
@@ -72,6 +72,28 @@ export interface JobHandlerWithExtra<
 }
 
 
+export function renameJob<I extends JsonValue, O extends JsonValue>(
```
clydin: This doesn't seem used anywhere.

It also does far more than just rename a job.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
```diff
@@ -321,10 +339,27 @@ export abstract class FileSystemEngineHostBase implements
       return factory();
     }
 
+    if (this._scheduler.has(name)) {
+      return observableOf(<T>(options: T | undefined, context: SchematicContext) => {
+        if (options !== undefined && !json.isJsonObject(options)) {
+          throw new UnregisteredTaskException(name);
```
clydin: Wrong exception?

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
```diff
@@ -321,10 +339,27 @@ export abstract class FileSystemEngineHostBase implements
...
+        }
+        const job = this._scheduler.schedule(
+          name,
+          options ? { ...JSON.parse(JSON.stringify(options)) } : context,
```
clydin: Why the `stringify`/`parse`?

The Task also expects the context (second parameter) to be the `SchematicContext` for the scheduling schematic.  That appears to be lost upon scheduling.  Is that an intentional API change?

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
```diff
@@ -321,10 +339,27 @@ export abstract class FileSystemEngineHostBase implements
...
+          {
+            logger: context.logger,
+          },
+        );
+
+        return job.promise.then(() => {});
```
clydin: `job.output.pipe(ignoreElements());`?

packages/angular_devkit/schematics/tasks/node/index.ts
```diff
@@ -28,8 +29,11 @@ export class BuiltinTaskExecutor {
     name: RunSchematicName,
     create: () => import('../run-schematic/executor').then(mod => mod.default()),
   };
-  static readonly TslintFix: TaskExecutorFactory<{}> = {
-    name: TslintFixName,
-    create: () => import('../tslint-fix/executor').then(mod => mod.default()),
-  };
+  static readonly TslintFix = jobs.lazyLoadJob(() => {
```
clydin: These are still tasks even though they are internally implemented as jobs.
Why not keep the TaskExecutorFactory and add the `input` schema option.  And maybe a temporary flag to indicate this is a job-based task.
Tasks don't return any values so there is no need for an output schema.

The registration method can then wrap the job-based tasks as appropriate.

packages/angular_devkit/core/src/jobs/api.ts
```diff
@@ -419,3 +419,15 @@ export interface Scheduler<
     options?: ScheduleJobOptions,
   ): Job<O>;
 }
+
+
+/**
+ * Returns true if the passed in value is a JobHandler<>.
+ */
+export function isJobHandler<InputT extends JsonValue, OutputT extends JsonValue>(
```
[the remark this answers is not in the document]
hansl: I would like to have this as a comprehensive check though. Maybe I should restrict the `JobHandler<>` API so people don't create JobHandlers anymore, and tell people to use `createJob()`.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
```diff
@@ -116,9 +118,13 @@ export abstract class FileSystemEngineHostBase implements
 
   private _transforms: OptionTransform<{}, {}>[] = [];
   private _taskFactories = new Map<string, () => Observable<TaskExecutor>>();
+  private _scheduler = new jobs.SimpleJobRegistry();
+
+  get scheduler() { return this._scheduler; }
```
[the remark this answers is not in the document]
hansl: Removed. Let's wait until it's used before exposing it.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
hansl: That doesn't subscribe. Promise does.

packages/angular_devkit/core/src/jobs/create-job.ts
hansl: It's used in the Tasks API. I'll rename it.

hansl: _And_ move it to the original Jobs PR (not this one).

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
hansl: We also return the new `Promise<void>` which is necessary.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
hansl: Want to make a pure JSON copy of it.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
clydin: It shouldn’t subscribe.  It should return a promise or observable.

packages/angular_devkit/schematics/tools/file-system-engine-host-base.ts
clydin: The options are already being checked if they are a JsonObject above.  Turning a potentially large object into a string and then back again is not very efficient.

PR #13004, opened 2018-11-20
feat(@angular/cli): adding checkbox prompt support

okeefem2:
Hi there!

This is my first contribution, and actually my first contribution to an open source project of this size. Please let me know if there is anything I may have missed in terms of adding tests or could have done better etc.

Very excited to submit this and hope the community will find it useful!

Thanks,

Michael

PR description:

Added general checkbox support using inquirer for schematics prompts
Added checkbox 'implements' option to guard schematics to allow user to choose implementations (CanActivate, CanActivateChild or CanLoad)
Notes:

I Purposefully left out CanDeactivate because of the added complexity of needed a Component to implement. I was definitely considering that as a follow up to this PR, but since that would take a little more time on my part I wanted to get this out there first.

### DIFF
--- a/packages/angular/cli/models/schematic-command.ts
+++ b/packages/angular/cli/models/schematic-command.ts
@@ -49,7 +49,6 @@ import { BaseCommandOptions, Command } from './command';
 import { Arguments, CommandContext, CommandDescription, Option } from './interface';
 import { parseArguments, parseFreeFormArguments } from './parser';
 
-
 export interface BaseSchematicSchema {
   debug?: boolean;
   dryRun?: boolean;
@@ -309,7 +308,8 @@ export abstract class SchematicCommand<
               question.type = 'confirm';
               break;
             case 'list':
-              question.type = 'list';
+            case 'checkbox':
+              question.type = definition.type;
               question.choices = definition.items && definition.items.map(item => {
                 if (typeof item == 'string') {
                   return item;
--- a/packages/angular_devkit/core/src/json/schema/interface.ts
+++ b/packages/angular_devkit/core/src/json/schema/interface.ts
@@ -99,7 +99,7 @@ export interface PromptDefinition {
   id: string;
   type: string;
   message: string;
-  default?: string | number | boolean | null;
+  default?: string | string[] | number | boolean | null;
   priority: number;
   validator?: (value: string) => boolean | string | Promise<boolean | string>;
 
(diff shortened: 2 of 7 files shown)


okeefem2:
Created new PR because I messed up the commit subject format on the first one. This way too the history is cleaner.

packages/angular/cli/models/schematic-command.ts
```diff
@@ -309,7 +308,8 @@ export abstract class SchematicCommand<
               question.type = 'confirm';
               break;
             case 'list':
-              question.type = 'list';
+            case 'checkbox':
+              question.type = definition.type;
```
clydin: The initial plan for this concept was to model this as a `multiselect` question option on the list type itself. Could you make that change?

packages/angular_devkit/core/src/json/schema/prompt_spec.ts
```diff
@@ -262,6 +262,45 @@ describe('Prompt Provider', () => {
...
+      registry
+        .compile({
+          properties: {
+            test: {
+              type: 'array',
+              enum: [
+                'one',
+                'two',
+                'three',
+              ],
```
clydin: This is an invalid schema, something cannot be an array and also be equal to `one`/`two`/`three`. For array definition, the items field should be used. For additional information, please see here: https://json-schema.org/understanding-json-schema/reference/array.html

packages/angular_devkit/core/src/json/schema/interface.ts
```diff
@@ -109,7 +109,7 @@ export interface PromptDefinition {
 }
 
 export type PromptProvider = (definitions: Array<PromptDefinition>)
-  => SubscribableOrPromise<{ [id: string]: JsonValue }>;
+  => SubscribableOrPromise<{ [id: string]: JsonValue } | { [id: string]: JsonValue }[]>;
```
clydin: This shouldn't need to be changed.  Each `id` (i.e., field) should have one value (a `JsonValue` can be an array).

okeefem2:
@clydin I would be happy to get those done ASAP! Thank you for the review/information.

packages/angular_devkit/core/src/json/schema/interface.ts
okeefem2: That makes sense. This was a bit of an artifact from some typing errors I had caused while debugging. Good catch!

okeefem2:
Closing to split features into two PRs. Please refer to https://github.com/angular/angular-cli/pull/13031 fpr checkbox/multiselect support. Guard schematic changes to come.

PR #13031, opened 2018-11-22
Feature: Added support for multiselect list prompt

okeefem2:
@clydin Here is the PR for just the core changes to @angular/cli and @angular-devkit-core. Including the changes you requested on https://github.com/angular/angular-cli/pull/13004. Please let me know if any further changes are necessary! 

I will open a second PR with the guard schematic changes if/when this one is merged in.

Thank you, 

Michael

### DIFF
--- a/packages/angular/cli/models/schematic-command.ts
+++ b/packages/angular/cli/models/schematic-command.ts
@@ -309,7 +309,7 @@ export abstract class SchematicCommand<
               question.type = 'confirm';
               break;
             case 'list':
-              question.type = 'list';
+              question.type = !!definition.multiselect ? 'checkbox' : 'list';
               question.choices = definition.items && definition.items.map(item => {
                 if (typeof item == 'string') {
                   return item;
--- a/packages/angular_devkit/core/src/json/schema/interface.ts
+++ b/packages/angular_devkit/core/src/json/schema/interface.ts
@@ -99,13 +99,14 @@ export interface PromptDefinition {
   id: string;
   type: string;
   message: string;
-  default?: string | number | boolean | null;
+  default?: string | string[] | number | boolean | null;
   priority: number;
   validator?: (value: string) => boolean | string | Promise<boolean | string>;
 
   items?: Array<string | { value: JsonValue, label: string }>;
 
   raw?: string | JsonObject;
+  multiselect?: boolean;
 }
 
 export type PromptProvider = (definitions: Array<PromptDefinition>)
--- a/packages/angular_devkit/core/src/json/schema/registry.ts
+++ b/packages/angular_devkit/core/src/json/schema/registry.ts
@@ -568,6 +568,7 @@ export class CoreSchemaRegistry implements SchemaRegistry {
           priority: 0,
           raw: schema,
           items,
+          multiselect: type === 'list' ? schema.multiselect : false,
           default: typeof parentSchema.default == 'object' ? undefined : parentSchema.default,
           async validator(data: string) {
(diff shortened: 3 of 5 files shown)


packages/angular_devkit/core/src/json/schema/interface.ts
```diff
@@ -99,13 +99,14 @@ export interface PromptDefinition {
   id: string;
   type: string;
   message: string;
-  default?: string | number | boolean | null;
+  default?: string | string [] | number | boolean | null;
```
clydin: extra space on `string[]`

packages/angular/cli/models/schematic-command.ts
```diff
@@ -309,7 +310,8 @@ export abstract class SchematicCommand<
               question.type = 'confirm';
               break;
             case 'list':
-              question.type = 'list';
+              const multiselect = !!definition.raw && (definition.raw as JsonObject)['multiselect'];
+              question.type = !!multiselect ? 'checkbox' : 'list';
```
clydin: should be able to use `definition.multiselect` directly here.

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -624,23 +625,44 @@ export class CoreSchemaRegistry implements SchemaRegistry {
...
       map(answers => {
-        for (const path in answers) {
-          const pathFragments = path.split('/').map(pf => {
-            if (/^\d+$/.test(pf)) {
-              return pf;
-            } else {
-              return '\'' + pf + '\'';
-            }
-          });
+        const pathFragmenter = (pf: string) => {
```
clydin: I don't think these changes are needed.  `answers` should always be an object.

packages/angular/cli/models/schematic-command.ts
okeefem2: I was having trouble with this. At this point in the process, definition.multiselect is not present. It is only present in definition.raw. I am guessing there is some step where that is either not being set to the definition or is being lost. Would you have any insight as to where that might be?

packages/angular_devkit/core/src/json/schema/registry.ts
okeefem2: You are right! I guess when I was first working on this the answers were coming through as an array of the selected values. I realize now that was an artifact of my erroneous test case. I will make the change right away! For making changes, what is the standard for the commits? @clydin

clydin: The process is generally to make additional commits locally; then squash (`git rebase -i master`) the fix-up commits into the relevant previously existing commit.

packages/angular/cli/models/schematic-command.ts
okeefem2: I made the requested changes minus this one... When testing this I am still having the issue where multiselect is not present at this point. I can see that it is present at some point before this, but  it is being stripped out at some point but I am having trouble identifying where.

okeefem2: I think I have possibly figured out what is wrong. The built version of @angular/cli has a dependency on @angular-devkit/core, looking into the registry.js in the node_modules of @angular/cli I do not see the changes I have made, but looking in dist folder I do see the changes. So if I link the built @angular-devkit/core to the built @angular/cli this should show the changes I expect. Is there a better way to build the packages locally to test changes across multiple packages @clydin?

okeefem2: Either way, this did work. So all changes have been addressed!

Issue #8170, opened 2017-10-24
Generate validator


zlepper:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Desired functionality.
I would like to be able to generate a new custom validator like i'm generating components or services. 
This includes all the necessary selectors and providers. The class that is generated should implement the `Validator` interface.

The use case for this is that i'm only doing this semi-frequently, so I can't remember all the "boilerplate" like the providers. I have existing examples i can look at, but doing `$ ng g validator Foo` to generate a `FooValidator` would ease that workflow a lot.

ciesielskico:
With the next release (it is already possible with the RC of 1.5.0) you will be able to implement your own schematics, something like your own blueprints for  the generator.

zlepper:
Ohh, that sounds neat. Though an official template from the Angular team would still be nice, especially for new project/computers, depending on how it works.

clydin:
Just wanted to add a quick note that there are some examples of what the files could look like in this issue: https://github.com/angular/angular-cli/issues/8709

ShinDarth:
It would be nice if we had something like:

`ng generate validator path/to/my-custom-validator`

which automatically generates skeletons for a custom form validator and its unit tests.

For example:

`my-custom.validator.ts`

```
import { AbstractControl, ValidationErrors, ValidatorFn } from '@angular/forms';

export function myCustomValidator(arg1: string, arg2: string): ValidatorFn {
  return (form: AbstractControl): ValidationErrors | null => {
    const error: ValidationErrors = { myCustomErrorName: true };

    // ...

    form.setErrors(null);
    return null;
  };
}
```


`my-custom.validator.spec.ts`

```
import { AbstractControl, FormControl, FormGroup, ValidationErrors, ValidatorFn } from '@angular/forms';
import Spy = jasmine.Spy;

import { myCustomValidator } from './my-custom-validator.ts';

describe('MyCustomValidator', () => {
  let form: FormGroup;
  let myCustomValidatorFn: ValidatorFn;

  const error: ValidationErrors = { myCustomErrorName: true };

  beforeEach(() => {
    form = new FormGroup({
      // ...
    });

    myCustomValidatorFn = myCustomValidator(...);

    spySetErrors = spyOn(form, 'setErrors').and.callThrough();
  });

  it('should correctly ...', () => {
    // ...

    expect(myCustomValidatorFn(form)).toEqual(...);

[...]

PR #13047, opened 2018-11-25
Add Validator generation

FrancescoBorzi:
I'm trying to address: https://github.com/angular/angular-cli/issues/8170

I need help to figure out how to complete this PR. First of all, how can I execute the code that I added to check if it's really working?

### DIFF
--- a/packages/schematics/angular/collection.json
+++ b/packages/schematics/angular/collection.json
@@ -89,6 +89,11 @@
       "description": "Create an Angular universal app.",
       "schema": "./universal/schema.json"
     },
+    "validator": {
+      "factory": "./validator",
+      "description": "Create an Angular form validator.",
+      "schema": "./validator/schema.json"
+     },
     "appShell": {
       "aliases": [ "app-shell" ],
       "factory": "./app-shell",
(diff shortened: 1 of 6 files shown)


FrancescoBorzi:
any help with this?

okeefem2:
@FrancescoBorzi One thing you can do (someone from the team may have better advice) is to build the project `npm run build`, then run npm link in the package that you made changes in, for you this would be `dist/schematics/angular` this links this built package to your global npm. Then in any angular project on your machine, run `npm link @schematics/angular` to create a symlink between your global and local (project) packages. Then you can use the code you wrote. Another thing I messed up at first is that you need to adhere to strict commit message guidelines, please [refer to the documentation](https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#commit).

FrancescoBorzi:
@okeefem2 thanks a lot for your suggestion, that helped. About the commit message I guess we can deal with it later since the commits can be squashed and renamed when merging, right?

okeefem2:
@FrancescoBorzi Happy to help! I look forward to being able to generate Validators!

alexeagle:
Hi there,

Thanks for your proposal. We thought about this a fair amount.

One thing to observe is that angular-cli contains schematics for angular packages. This makes our ownership model more complex within our team. If someone who owns the `@angular/forms` package understands validators the best, then it seems like the `@angular/forms` schematic ought to be included in that package and owned by the same person.

Another observation is that creating a validator function is not difficult for users. But maintaining this code is an extra burden for us. So it doesn't meet our (current) bar for how much functionality we want to own.

You're free to publish a third-party schematic package that provides this, of course. We're happy to have your work be part of the angular community, we just don't want to merge it to this location.

FrancescoBorzi:
@alexeagle thanks for your answer, I understand. Any suggestions about how could that be included into a third-party schematic package that can be easily used by users?

PR #13048, opened 2018-11-26
build: add support for `x-deprecated` when generating schema

alan-agius4:
When using the CLI and a property is deprecated we emit a warning by using the `x-deprecated` field. However API consumers have no way to know that a property is deprecated at the moment.

This converts `x-deprecated` to  `@deprecated` comments


alan-agius4:
Yeah, but with producers, you cannot amend the description, only add addition items to schema.

Unless there is another part of the API, that I don't know of.

PR #13049, opened 2018-11-26
feat(@schematics/angular): consistent naming of options and arguments that do the same thing

### LINKED ISSUE #12784: Consistent naming of options and arguments that do the same thing
At the moment, certain options and arguments that do the same thing should are not named the same across all schematics.

In `ng new` to skip the creation of spec files, one needs to use `--skip-test`, with an alias of `-S`, while in other schematics it is `--spec` and no alias.

In the component schematics `ng generate component`, to specify the style preprocessor you need to use  `--styleext`, while in `ng new` this is named `--style`

We should discuss this, choose the preferred namings and align all the same options to make it easier for the users.

alan-agius4:
feat(@schematics/angular): consistent naming of options and arguments that do the same thing

This aligns options that do the same thing:
1) `skipSpecs` and `spec` has been deprecated in favor of `skipTests`.
2) `styleext` has been deprecated in favor of `style` since the latest is two words.

Fixes #12784

### DIFF
--- a/tools/quicktype_runner.js
+++ b/tools/quicktype_runner.js
@@ -131,6 +131,7 @@ async function generate(inPath) {
     rendererOptions: {
       'just-types': 'true',
       'explicit-unions': 'true',
+      'acronym-style': 'camel',
     },
   });
 
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
   "dependencies": {
     "glob": "^7.0.3",
     "node-fetch": "^2.2.0",
-    "quicktype-core": "^5.0.41",
+    "quicktype-core": "^6.0.15",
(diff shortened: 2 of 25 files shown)


tools/quicktype_runner.js
```diff
@@ -12,11 +12,31 @@ const {
   JSONSchema,
   JSONSchemaInput,
   JSONSchemaStore,
-  TypeScriptTargetLanguage,
   parseJSON,
   quicktype,
+  splitIntoWords,
```
alan-agius4: This is obsolete when https://github.com/quicktype/quicktype/pull/1119#event-1987297419 is released

alan-agius4:
@hansl updated, totally forgot about that.

PR #13051, opened 2018-11-26
docs(@angular/cli): change default test watch flag

Mathou54:


Mathou54:
hi @alan-agius4, sorry for the mistake, I hope it is better like that.

alan-agius4:
Hi, you need to squash your commits into a single one, as like this the original commit is still present.

mbward:
Looks like this commit forgot to update the docs too.  Filed https://github.com/angular/angular/issues/28140.

PR #13062, opened 2018-11-27
feat(@angular-devkit/build-angular): fine grain settings for sourceMaps

### LINKED ISSUE #7527: Add option for hidden-source-map to use with ng build --prod
```
- [x] feature request
```
### Versions.
1.4.0-rc2

### Repro steps.
Feature

Feature

### Desired functionality.
Add an option to generate hidden source maps for production builds. Hidden source maps are helpful because they are full source maps, but not referenced in the minified files. This allows a developer to use the source maps for debugging purposes without letting the browser know they exist. Error tracking sites such as Atatus allow users to upload source maps manually so the stack traces are more useful and the maps never need to be made public. 

In my case, I have modified my fork of the cli to generate hidden source maps when using the production build and source-maps=true. This has worked for me, but I didn't expect it would be the best fit for the whole project. 
```js
devtool: buildOptions.sourcemaps ? 'hidden-source-map' : false,
```

https://github.com/mhamel06/angular-cli/blob/2f551150bccdeb6121c79f0b1fd65bfdd68c33ce/packages/%40angular/cli/models/webpack-configs/production.ts#L108-L110

https://webpack.js.org/configuration/devtool/

alan-agius4:
This PR add more control over which sourceMaps you want, Now you can enable sourceMaps for scripts only, styles only or both. Also we added another functionality which are hidden sourcemaps. These are normally used for error reporting tools.

Fixes #7527

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -633,9 +633,39 @@
               "default": false
             },
             "sourceMap": {
-              "type": "boolean",
               "description": "Output sourcemaps.",
-              "default": true
+              "default": true,
+              "oneOf": [
+                {
+                  "type": "object",
+                  "properties": {
+                    "scripts": {
+                      "type": "boolean",
+                      "description": "Output sourcemaps for all scripts.",
+                      "default": true
+                    },
+                    "styles": {
+                      "type": "boolean",
+                      "description": "Output sourcemaps for all styles.",
+                      "default": true
+                    },
+                    "hidden": {
+                      "type": "boolean",
+                      "description": "Output sourcemaps used for error reporting tools.",
+                      "default": false
+                    },
+                    "vendor": {
+                      "type": "boolean",
+                      "description": "Resolve vendor packages sourcemaps.",
+                      "default": false
+                    }
+                  },
+                  "additionalProperties": false
+                },
+                {
+                  "type": "boolean"
+                }
+              ]
             },
             "vendorSourceMap": {
               "type": "boolean",
@@ -1045,8 +1075,34 @@
               "description": "Build using Ahead of Time compilation."
             },
             "sourceMap": {
-              "type": "boolean",
-              "description": "Output sourcemaps."
+              "description": "Output sourcemaps.",
+              "default": true,
+              "oneOf": [
+                {
+                  "type": "object",
+                  "properties": {
+                    "scripts": {
+                      "type": "boolean",
+                      "description": "Output sourcemaps for all scripts.",
+                      "default": true
+                    },
+                    "styles": {
+                      "type": "boolean",
+                      "description": "Output sourcemaps for all styles.",
+                      "default": true
+                    },
+                    "vendor": {
+                      "type": "boolean",
+                      "description": "Resolve vendor packages sourcemaps.",
+                      "default": false
+                    }
+                  },
+                  "additionalProperties": false
+                },
+                {
+                  "type": "boolean"
+                }
+              ]
             },
             "vendorSourceMap": {
               "type": "boolean",
@@ -1189,9 +1245,34 @@
               "description": "Defines the build environment."
             },
             "sourceMap": {
-              "type": "boolean",
               "description": "Output sourcemaps.",
-              "default": true
+              "default": true,
+              "oneOf": [
+                {
+                  "type": "object",
+                  "properties": {
+                    "scripts": {
+                      "type": "boolean",
+                      "description": "Output sourcemaps for all scripts.",
+                      "default": true
+                    },
+                    "styles": {
+                      "type": "boolean",
+                      "description": "Output sourcemaps for all styles.",
+                      "default": true
+                    },
+                    "vendor": {
+                      "type": "boolean",
+                      "description": "Resolve vendor packages sourcemaps.",
+                      "default": false
+                    }
+                  },
+                  "additionalProperties": false
+                },
+                {
+                  "type": "boolean"
+                }
+              ]
             },
             "progress": {
               "type": "boolean",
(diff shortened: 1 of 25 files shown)


alan-agius4:
Still need to add a couple of tests but this is the idea. @clydin, @hansl, @filipesilva

packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
```diff
@@ -23,7 +24,8 @@ export interface BuildOptions {
   outputPath: string;
   resourcesOutputPath?: string;
   aot?: boolean;
-  sourceMap?: boolean;
+  sourceMap?: SourceMap;
```
clydin: The build options at this point should be fully normalized which should be handled at least one level up from here.

alan-agius4: The thing is that in a number of cases, I need to know if the sourceMaps is for scripts or styles. 

Example: 
https://github.com/angular/angular-cli/pull/13062/files#diff-fdb0776b9f2db468444818966705634cR50
https://github.com/angular/angular-cli/pull/13062/files#diff-ce7fb1412b327df10292e336d2e96e16R20

You propose to create two options? `scriptsSourceMap` and `styleSourceMap` to normalize it and ?

clydin: There's really two different sets of options here.  The "boolean or object" options are a user interface level concern.  But at the level here the options object should be fully actionable and unambiguous.

packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
```diff
@@ -15,6 +15,7 @@ import {
   Budget,
   CurrentFileReplacement,
   ExtraEntryPoint,
+  SourceMap,
```
hansl: Name this `SourceMapOption` or something. `SourceMap` is also a node module we're using.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
```diff
@@ -66,3 +67,25 @@ export function normalizeExtraEntryPoints(
...
+  hiddenSourceMap = false,
+  inlineSourceMap = false,
+): SourceMapDevToolPlugin {
+  const include = [];
+  if (scriptsSourceMap) {
+    include.push(/js$/);
+  }
+
+  if (stylesSourceMap) {
+    include.push(/css$/);
```
filipesilva: What about sass/others?

packages/angular/cli/lib/config/schema.json
```diff
@@ -633,9 +633,48 @@
               "default": false
             },
             "sourceMap": {
-              "type": "boolean",
+              "type": [
```
filipesilva: Schema bookkeeping is getting pretty annoying. We should find a way around this sometime.

alan-agius4: Yeah, it’s pretty hard even just to add a new property. You have to add t in multiple places

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
alan-agius4: This is actually for the emitted modules. So it can never be sass etc...

filipesilva: Got it, nvm then.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
```diff
@@ -66,3 +67,25 @@ export function normalizeExtraEntryPoints(
...
+  }
+
+  return new SourceMapDevToolPlugin({
+    filename: inlineSourceMap ? undefined : '[file].map',
```
clydin: does the file format account for the hashing in production?

packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -82,13 +82,49 @@
       "default": false
     },
     "sourceMap": {
-      "type": "boolean",
+      "type": ["boolean", "object", "string"],
```
clydin: to reduce the surface area somewhat, what do you think about removing the string option?  If someone wants hidden then they need to use the object form.  It is an advanced option.

packages/angular_devkit/build_angular/src/karma/schema.json
```diff
@@ -63,13 +63,43 @@
...
+            },
+            "vendor": {
+              "type": "boolean",
+              "description": "Resolve vendor packages sourcemaps.",
+              "default": false
+            }
+          },
+          "additionalProperties": false
+        },
+        {
```
clydin: this can be simplified to `{ type: "boolean" }`

packages/angular_devkit/build_angular/src/karma/schema.json
```diff
@@ -63,13 +63,43 @@
       "description": "Defines the build environment."
     },
     "sourceMap": {
-      "type": "boolean",
+      "type": ["boolean", "object"],
```
clydin: This is redundant with the `oneOf`.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
alan-agius4: How about ill add a test for that? :)

packages/angular_devkit/build_angular/src/browser/schema.json
alan-agius4: I agree.

packages/angular_devkit/build_angular/src/karma/schema.json
alan-agius4: done

PR #13074, opened 2018-11-28
 feat(@schematics/angular): account for root level assets and `resourcesOutputPath`

### LINKED ISSUE #13067: Moving/Renaming CSS resources breaks ServiceWorker caching
```
- [x] feature request
```

```
- [x] build
```

### Versions
```
> node --version
v10.13.0
```
```
> npm --version
6.4.1
```
```
> ng --version
Angular CLI: 7.1.0
Node: 10.13.0
OS: win32 x64
Angular: 7.1.0
... router, service-worker

@angular-devkit/architect         0.11.0
@angular-devkit/build-angular     0.11.0
@angular-devkit/build-optimizer   0.11.0
@angular-devkit/build-webpack     0.11.0
@angular-devkit/core              7.1.0
@angular-devkit/schematics        7.1.0
@angular/pwa                      0.11.0
@ngtools/webpack                  7.1.0
@schematics/angular               7.1.0
@schematics/update                0.11.0
typescript                        3.1.6
```

Windows 10

### Repro steps

1. Create a PWA and use an image in CSS (e.g. as background):
```
ng new test-app
cd test-app
ng add @angular/pwa
npm install http-server --save-dev
node --print "'body { background: url(assets/icons/icon-72x72.png); }'" >> src/styles.css
```

2. Serve with ServiceWorker:
```
ng build --prod
node_modules/.bin/http-server dist/test-app -p 4200 -o
```

3. Go offline (e.g. open `DevTools > Network` and check "Offline"):

4. Reload page.

Result: There is no background :sob:

### Desired functionality
There should be background 🌞
(See below.)

Basically, the issue is that we configure the SW (in `ngsw-config.json`) to cache specific resources and make them available offline. In this case, the `/assets/**` files are configured to be cached by default. Since the CSS references an image inside `assets/`, one would expect it to be available offline without any further configuration.

Instead, the image is moved to the root output directory during the build (and renamed to include the hash) and the CSS is rewritten to look for that top-level image. This works fine as long as there is network connectivity, but the SW is not configured to cache this top-level image. As soon as the user goes offline, the image is not available any more (since the SW cannot provide it) and the nice background is gone 😢

It is easy to fix the issue once you knwo what it is (e.g. by configuring the SW to cache the resulting file or using absolute path in CSS, which are not affected). But the problem is hard to find (e.g. usually there are no tests for how the app behaves in offline mode) and not straight forward to diagnose (i.e. you have to know about the rewriting).

@filipesilva suggested that the SW config should modified to account for the moved/renamed files.

alan-agius4:
By default we are only account for assets inside the assets folder. Which breaks the offline experience.

Fixes #13067

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/service-worker_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/service-worker_spec_large.ts
@@ -20,8 +20,9 @@ describe('Browser Builder service worker', () => {
         name: 'app',
         installMode: 'prefetch',
         resources: {
-          files: ['/favicon.ico', '/index.html'],
-          versionedFiles: [
+          files: [
+            '/favicon.ico',
+            '/index.html',
             '/*.bundle.css',
             '/*.bundle.js',
             '/*.chunk.js',
@@ -33,7 +34,10 @@ describe('Browser Builder service worker', () => {
         installMode: 'lazy',
         updateMode: 'prefetch',
         resources: {
-          files: ['/assets/**'],
+          files: [
+            '/assets/**',
+            '/*.(eot|svg|cur|jpg|png|webp|gif|otf|ttf|woff|woff2|ani)',
+          ],
         },
       },
     ],
@@ -55,6 +59,7 @@ describe('Browser Builder service worker', () => {
     host.writeMultipleFiles({
       'src/ngsw-config.json': JSON.stringify(manifest),
       'src/assets/folder-asset.txt': 'folder-asset.txt',
+      'src/styles.css': `body { background: url(./spectrum.png); }`,
     });
 
     const overrides = { serviceWorker: true };
@@ -91,6 +96,7 @@ describe('Browser Builder service worker', () => {
               updateMode: 'prefetch',
               urls: [
                 '/assets/folder-asset.txt',
+                '/spectrum.png',
               ],
               patterns: [],
             },
@@ -100,6 +106,7 @@ describe('Browser Builder service worker', () => {
             '/favicon.ico': '84161b857f5c547e3699ddfbffc6d8d737542e01',
             '/assets/folder-asset.txt': '617f202968a6a81050aa617c2e28e1dca11ce8d4',
             '/index.html': '843c96f0aeadc8f093b1b2203c08891ecd8f7425',
+            '/spectrum.png': '8d048ece46c0f3af4b598a95fd8e4709b631c3c0',
           },
         });
       }),
--- a/packages/schematics/angular/service-worker/files/ngsw-config.json
+++ b/packages/schematics/angular/service-worker/files/ngsw-config.json
@@ -18,7 +18,8 @@
       "updateMode": "prefetch",
       "resources": {
         "files": [
-          "/assets/**"
+          "/assets/**",
+          "<%= resourcesOutputPath %>/*.(eot|svg|cur|jpg|png|webp|gif|otf|ttf|woff|woff2|ani)"
         ]
       }
     }
--- a/packages/schematics/angular/service-worker/index.ts
+++ b/packages/schematics/angular/service-worker/index.ts
@@ -26,30 +26,41 @@ import { getWorkspace, updateWorkspace } from '../utility/config';
 import { addPackageJsonDependency, getPackageJsonDependency } from '../utility/dependencies';
 import { getAppModulePath } from '../utility/ng-ast-utils';
 import { getProjectTargets, targetBuildNotFoundError } from '../utility/project-targets';
-import { BrowserBuilderOptions, BrowserBuilderTarget } from '../utility/workspace-models';
+import {
+  BrowserBuilderOptions,
+  BrowserBuilderTarget,
+  WorkspaceSchema,
+} from '../utility/workspace-models';
 import { Schema as ServiceWorkerOptions } from './schema';
+
+function getProjectConfiguration(
+  workspace: WorkspaceSchema,
+  options: ServiceWorkerOptions,
+): BrowserBuilderOptions {
+  const projectTargets = getProjectTargets(workspace, options.project);
+  if (!projectTargets[options.target]) {
+    throw new Error(`Target is not defined for this project.`);
+  }
+
+  const target = projectTargets[options.target] as BrowserBuilderTarget;
+  let applyTo = target.options;
+
+  if (options.configuration &&
+    target.configurations &&
+    target.configurations[options.configuration]) {
+    applyTo = target.configurations[options.configuration] as BrowserBuilderOptions;
+  }
+
+  return applyTo;
+}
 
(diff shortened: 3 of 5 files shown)


packages/schematics/angular/service-worker/files/ngsw-config.json
```diff
@@ -18,7 +18,8 @@
       "updateMode": "prefetch",
       "resources": {
         "files": [
-          "/assets/**"
+          "/assets/**",
+          "<%= resourcesOutputPath %>/*.(eot|svg|cur|jpg|png|webp|gif|otf|ttf|woff|woff2|ani)"
```
alan-agius4: @gkalpak this is the line that interests you.

packages/angular_devkit/build_angular/test/browser/service-worker_spec_large.ts
```diff
@@ -20,8 +20,9 @@ describe('Browser Builder service worker', () => {
         name: 'app',
         installMode: 'prefetch',
         resources: {
-          files: ['/favicon.ico', '/index.html'],
-          versionedFiles: [
```
clydin: Nice.  I've had this on my list to fix for awhile.

packages/schematics/angular/service-worker/files/ngsw-config.json
clydin: Note that while this covers the vast majority of cases, it is not technically complete as any file type can be used within a CSS `url()`.

The config would need to be modified at build time to account for all the resources.

clydin: This would also break if the `resourcesOutputPath` was changed after service worker addition.  Some documentation should cover that for now, however.

alan-agius4: That is true, there is an alternative, but I felt it might be to inclusive. 
It's to use: `<%= resourcesOutputPath %>/*`

alan-agius4: @clydin, indeed that is correct `resourcesOutputPath`. I'll open a PR for AIO to document this.

That said, there is an alternative. instead of that glob, use `"<%= resourcesOutputPath %>/*` but it feels a bit to inclusive.

alan-agius4:
Hi, If no one has anything in contrary I will unblock this.

Regarding documentation, I think the best place is to add something in https://angular.io/guide/service-worker-getting-started.

I'll open a separate PR for this.

packages/schematics/angular/service-worker/files/ngsw-config.json
gkalpak: > Feels a bit of a magic, and might not be clear to the users, why certain files are being included even if not whitelisted in the config.

Well, that's not much more magic than renaming and moving files around :grin:
The ideal fix, imo, would be to update all asset group configs to match the renamed/moved files that they would have matched if the files were not renamed/moved. This would essentially ensure that the cli makes up for the renaming/moving magic.

But this will be a little more complicated, so not sure if it is worth the extra effort (and costs that come with the more complicated processing).

alan-agius4: Thinking about it, there is no need to modify the config at build time, as this can be done in a glob like `**/*` which will catch anything which is not caught in the app group. Alternatively not to be so inclusive `**/*.(eot|svg|cur|jpg|png|webp|gif|otf|ttf|woff|woff2|ani)"` which should solve the issue of the resourceOutPath.

Also, updating the asset group configs during build time to match the actual files processed by the CLI, is pretty straight forward.

I am more inclined towards having it hardcoded in the config file for 2 main reason, experienced users have more control over the assets groups, as you cannot use negations.

1) Option one is to leave it as is in the current PR.
2) Use wildcards in the config `**/*`
2) Use wildcards with extensions  the config `**/*.(eot|svg|cur|jpg|png|webp|gif|otf|ttf|woff|woff2|ani`
3) Build assets groups during build time based on the files processed by webpack.
4) Do step 3, only if assets group `files` under assets group is empty or undefined.

@gkalpak & @clydin, what do you think?

gkalpak: > Thinking about it, there is no need to modify the config at build time, as this can be done in a glob like **/* which will catch anything which is not caught in the app group.

This is not necessarily something you want to do (cache every single file in your dist directory).

> Also, updating the asset group configs during build time to match the actual files processed by the CLI, is pretty straight forward.

I am not saying it is impossible (or even too hard), but it is not as straight-forward as the current solution.
But if you feel otherwise, let's do it :smiley:

> I am more inclined towards having it hardcoded in the config file for 2 main reason, experienced users have more control over the assets groups, as you cannot use negations.

I don't understand what you are saying.
How do experienced users get more control? Why can't you use negations?

> Here are some options:
>
> Option one is to leave it as is in the current PR.

Definitely not ideal, but I can probably live with that option.

> Use wildcards in the config **/*

That doesn't sound like a good idea.

> Use wildcards with extensions the config **/*.(eot|svg|cur|jpg|png|webp|gif|otf|ttf|woff|woff2|ani

Definitely not ideal, but I can probably live with that too.

> Build assets groups during build time based on the files processed by webpack.

That would be the most accurate fix imo, but will add extra complexity/maintainance cost (for us, not for the users).
(Not convinced it is worth the extra effort, but if you feel like going that way I won't stop you :grin:)

alan-agius4: Negations won’t work as they are not supported by the service worker generator see: https://angular.io/guide/service-worker-config#resources

So basicilly a user will not be able to exclude a file which the CLI added via the config during build time.

I am more for the wildcard with the extension approach, it a good default and with a documentation users can configure it as they wish. 

Re the adding config during build time, it’s pretty straight forward, the only problem comes which files you want want to cache? Basically it will be the same as adding `**/*.(gif|jpg)...` in the config, since as you mentioned it doesn’t make sense to cache everything in the dist.

gkalpak: How is that part of the documentation implying that you can't use negations? I am pretty sure you can :grin:
It is also mentioned near the top of the same document:

> The ! prefix marks the pattern as being negative, meaning that only files that don't match the pattern will be included.

##
Re the adding config during build time, you would need to iterate over each asset group in the config and do the following (off the top of my head):
1. Determine which files it matches (based on positive and negative glob patterns).
2. Check if any of these files have been renamed/moved.
3. Update the asset group config to also match the renamed/moved file.

But that is the super naive approach that can break with certain patterns (e.g. a negative pattern that does not match the original file but does match the renamed/moved file).

Actually, it is not possible to implement this in an 100% safe way, so probably better to go with the current approach + docs.

alan-agius4: I am too tired, I miss interpreted the docs 😂! Thanks George.

PR #13086, opened 2018-11-29
style: add .gitattributes file

alan-agius4:
On windows unless otherwise specified when cloning the repo the line ending will be CRLF which will cause some tests large specs to fail

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+# Auto detect text files and perform LF normalization
+* text=auto
+
+# JS and TS files must always use LF for tools to work
+*.js eol=lf
+*.ts eol=lf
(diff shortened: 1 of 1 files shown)


filipesilva:
@alan-agius4 does this work on the initial clone or is it just for new files?

I'd also like to make sure this repo and angular/angular are in sync regarding git checkouts on Windows. Is this the same that is done there?

alan-agius4:
@filipesilva this works on both the initial clone and newly added files.

Angular/Angular have something similar https://github.com/angular/angular/blob/master/.gitattributes but I have added some addition extensions for ours.

filipesilva:
Awesoem then!

PR #13106, opened 2018-12-01
fix(@schematics/update): only set registry option when used

clydin:


alan-agius4:
Is there an open issue for this?

PR #13108, opened 2018-12-01
docs: change link for CLI readme to be relative

alan-agius4:
//cc @gkamperis

Caretaker note: No need to release for this. This should be back-ported up till 6.2.x


gkamperis:
@alan-agius4 again... don't know if I am replying to early in the process but

https://github.com/angular/angular-cli/tree/6.2.x

still shows the wrong package versions.

PR #13109, opened 2018-12-01
feat(@schematics/angular): Added multiselect guard implements option

### LINKED ISSUE #13400: Add option to generate a guard for CanDeactivate, CanActivateChild, CanLoad
```
- [x] feature request
```

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Desired functionality

Actually (7.2.0), when a guard is generated, it implements `CanActivate`.
I would like to see an option to generate a guard with `CanDeactivate`, `CanActivateChild` or `CanLoad`.

okeefem2:
Follow up to https://github.com/angular/angular-cli/pull/13031, using the multiselect option for the guard schematic to ask the user which methods they would like to implement. I purposefully left out CanDeactivate because of the need to tie in a component. This would be a good candidate for a future PR though!

Fixes #13400

Thanks,

Michael

### DIFF
--- a/packages/schematics/angular/guard/files/__name@dasherize__.guard.ts
+++ b/packages/schematics/angular/guard/files/__name@dasherize__.guard.ts
@@ -1,14 +1,24 @@
 import { Injectable } from '@angular/core';
-import { CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot, UrlTree } from '@angular/router';
+import { <%= implementationImports %>ActivatedRouteSnapshot, RouterStateSnapshot, UrlTree } from '@angular/router';
 import { Observable } from 'rxjs';
 
 @Injectable({
   providedIn: 'root'
 })
-export class <%= classify(name) %>Guard implements CanActivate {
-  canActivate(
+export class <%= classify(name) %>Guard implements <%= implementations %> {
+  <% if (implements.includes('CanActivate')) { %>canActivate(
     next: ActivatedRouteSnapshot,
     state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
     return true;
   }
+  <% } %><% if (implements.includes('CanActivateChild')) { %>canActivateChild(
+    next: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    return true;
+  }
+  <% } %><% if (implements.includes('CanLoad')) { %>canLoad(
+    route: Route,
+    segments: UrlSegment[]): Observable<boolean> | Promise<boolean> | boolean {
+    return true;
+  }<% } %>
 }
--- a/packages/schematics/angular/guard/index.ts
+++ b/packages/schematics/angular/guard/index.ts
@@ -36,6 +36,20 @@ export default function (options: GuardOptions): Rule {
     if (options.path === undefined) {
       options.path = buildDefaultPath(project);
     }
+    if (options.implements === undefined) {
+      options.implements = [];
+    }
+
+    let implementations = '';
+    let implementationImports = '';
+    if (options.implements.length > 0) {
+      implementations = options.implements.join(', ');
+      implementationImports = `${implementations}, `;
+      // As long as we aren't in IE... ;)
+      if (options.implements.includes('CanLoad')) {
+        implementationImports = `${implementationImports}Route, UrlSegment, `;
+      }
+    }
 
     const parsedPath = parseName(options.path, options.name);
     options.name = parsedPath.name;
@@ -47,6 +61,8 @@ export default function (options: GuardOptions): Rule {
     const templateSource = apply(url('./files'), [
       options.skipTests ? filter(path => !path.endsWith('.spec.ts')) : noop(),
       template({
+        implementations,
+        implementationImports,
         ...strings,
         ...options,
       }),
--- a/packages/schematics/angular/guard/schema.json
+++ b/packages/schematics/angular/guard/schema.json
@@ -47,6 +47,24 @@
       "type": "boolean",
       "default": false,
       "description": "When true, applies lint fixes after generating the guard."
+    },
+    "implements": {
+      "type": "array",
+      "description": "Specifies which interfaces to implement.",
+      "uniqueItems": true,
+      "items": {
+        "type": "string"
+      },
+      "x-prompt": {
+        "message": "Which interfaces would you like to implement?",
+        "type": "list",
+        "multiselect": true,
+        "items": [
+          "CanActivate",
+          "CanActivateChild",
+          "CanLoad"
+        ]
+      }
     }
   },
   "required": [
--- a/packages/schematics/angular/guard/index_spec.ts
+++ b/packages/schematics/angular/guard/index_spec.ts
@@ -10,7 +10,6 @@ import { Schema as ApplicationOptions } from '../application/schema';
 import { Schema as WorkspaceOptions } from '../workspace/schema';
 import { Schema as GuardOptions } from './schema';
 
-
 describe('Guard Schematic', () => {
   const schematicRunner = new SchematicTestRunner(
     '@schematics/angular',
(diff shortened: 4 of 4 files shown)


packages/schematics/angular/guard/schema.json
```diff
@@ -41,6 +41,20 @@
...
+    },
+    "implements": {
+      "type": "array",
+      "description": "Specifies which interfaces to implement.",
+      "x-prompt": {
+        "message": "Which interfaces would you like to implement?",
+        "type": "list",
+        "multiselect": true,
+        "items": [
+          { "value": "CanActivate", "label": "CanActivate" },
```
clydin: You can remove the label.  The label is optional and will use the value if not present.

packages/schematics/angular/guard/schema.json
```diff
@@ -41,6 +41,20 @@
       "type": "boolean",
       "default": false,
       "description": "When true, applies lint fixes after generating the guard."
+    },
+    "implements": {
+      "type": "array",
```
clydin: Add `"uniqueItems": true` as well as:
```
"items": {
  "type": "string"
}
```

packages/schematics/angular/guard/index.ts
```diff
@@ -37,13 +37,26 @@ export default function (options: GuardOptions): Rule {
       options.path = buildDefaultPath(project);
     }
 
+    let implementations = '';
+    let implementationImports = '';
+    if (options.implements && options.implements.length > 0) {
```
clydin: set `options.implements` to an empty array if undefined first to avoid the repeat checks

okeefem2:
@clydin Thanks for the review! I have pushed changes, let me know if any more are necessary.

packages/schematics/angular/guard/index.ts
```diff
@@ -47,6 +61,8 @@ export default function (options: GuardOptions): Rule {
     const templateSource = apply(url('./files'), [
       options.skipTests ? filter(path => !path.endsWith('.spec.ts')) : noop(),
       template({
+        implementations: implementations,
+        implementationImports: implementationImports,
```
clydin: nit: you can shorten both of these to just:
```
implementations,
implementationImports,
```

packages/schematics/angular/guard/files/__name@dasherize__.guard.ts
```diff
@@ -1,14 +1,24 @@
...
     next: ActivatedRouteSnapshot,
     state: RouterStateSnapshot): Observable<boolean> | Promise<boolean> | boolean {
     return true;
   }
+  <% } %><% if (implements && implements.includes('CanActivateChild')) { %>canActivateChild(
+    next: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean> | Promise<boolean> | boolean {
+    return true;
+  }
+  <% } %><% if (implements && implements.includes('CanLoad')) { %>canLoad(
```
clydin: nit: can now remove the `implements` truthy check here and above.

okeefem2: Okay pushed those suggestions!

alan-agius4:
Can you add fixes #13400?

Thanks

alan-agius4:
Can you add fixes #13400?

Thanks

packages/schematics/angular/guard/schema.json
```diff
@@ -47,6 +47,24 @@
...
+      "description": "Specifies which interfaces to implement.",
+      "uniqueItems": true,
+      "items": {
+        "type": "string"
```
hansl: This looks like an `enum` instead of general string.

kyliau:
@okeefem2 Could you please fix the merge conflict?

packages/schematics/angular/guard/schema.json
okeefem2: If I change this to `enum`, I get the following build error:

`Error: Invalid type enum in JSON Schema at Schema#/properties/implements/items`

I realized that the way this is being used the items could be updated to just a list of strings, thank you for pointing that out. Let me know what you think!

alan-agius4: Should be something like;

```js
    "implements": {
      "type": "array",
      "default": "none",
      "enum": [
           CanActivate",
          "CanActivateChild",
          "CanLoad"
      ]
    },
```

packages/schematics/angular/guard/files/__name@dasherize__.guard.ts
```diff
@@ -1,14 +1,24 @@
...
 })
-export class <%= classify(name) %>Guard implements CanActivate {
-  canActivate(
+export class <%= classify(name) %>Guard implements <%= implementations %> {
+  <% if (implements.includes('CanActivate')) { %>canActivate(
     next: ActivatedRouteSnapshot,
     state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
     return true;
   }
+  <% } %><% if (implements.includes('CanActivateChild')) { %>canActivateChild(
```
alan-agius4: you can eventually use the enum type here.

packages/schematics/angular/guard/schema.json
okeefem2: @alan-agius4 I'm sorry I could be totally missing something (my first time working with json schema 😬), but the x prompt logic requires this array to be names as `items: []` as seen in https://github.com/angular/angular-cli/blob/master/packages/angular/cli/models/schematic-command.ts#L311

Are you saying in addition to the x-prompt defintion, also include the `enum` definition you describe here?

Thanks for your help btw! Just trying to learn and make sure I understand.

alan-agius4: Not really, I think it should be something like;

```
 "implements": {
      "type": "array",
      "default": "CanActivate",
      "enum": [
           "CanActivate",
           "CanActivateChild",
           "CanLoad"
      ],
      "x-prompt": {
        "message": "Which interfaces would you like to implement?",
        "type": "list",
        "multiselect": true,
        "items": [
          "CanActivate",
          "CanActivateChild",
          "CanLoad"
        ]
      }
}
...

Comment (okeefem2) [reply packages/schematics/angular/guard/schema.json]: Okay that makes sense thank you!

File: packages/schematics/angular/guard/files/__name@dasherize__.guard.ts (hunk 7 above)
Comment (okeefem2) [reply packages/schematics/angular/guard/files/__name@dasherize__.guard.ts]: @alan-agius4 Are you saying to create an enum for the values instead of comparing hard coded strings? If yes would this live in the index.ts file or should it have it’s own file?

Comment (alan-agius4) [reply packages/schematics/angular/guard/files/__name@dasherize__.guard.ts]: Hi, he enum is created from the json schema, you don't need to redefine it. 

You can run `yarn build` it will generate an new schema with the enum you created in the json schema.

So you can import it like such 
```ts
import { Schema as GuardOptions, Implement } from './schema';
```

packages/schematics/angular/guard/schema.json
alan-agius4: Note: updated the above is I spotted a mistake

clydin: Note also that I have a followup planned to add auto-configuration of multi-select based on the schema.  It will eliminate the need to manually specify most of the prompt options and is the reason I originally suggested just using a string constraint for now (this schema would be adjusted in the process).

packages/schematics/angular/guard/files/__name@dasherize__.guard.ts
okeefem2: Oh very cool. I will make these changes later today when I get home. Thanks!

packages/schematics/angular/guard/schema.json
okeefem2: @clydin would I be able to help out with the implementation of the multi select auto configuration?? I’d love to keep learning and contributing and this could transition well from the work I’ve done so far. Let me know what you think and thanks everyone for the help so far!

cyrilletuzi: @clydin @okeefem2 @alan-agius4 For info, the decision made here causes issues like this one: cyrilletuzi/vscode-angular-schematics#32. The x-prompts is a specific use of the schema, the classic schema properties should be respected for other tools to work (ie. this is an enum). For the same reason, `implements` should be flagged as `required`.

PR #13116, opened 2018-12-03
Fine grain optimization

alan-agius4:
We can now enabled/disabled optimisations for scripts and styles.

Ex: 
```
 optimization: {
   scripts: true,
   styles: false,
 },
```

### DIFF
--- a/packages/angular_devkit/build_angular/src/dev-server/index.ts
+++ b/packages/angular_devkit/build_angular/src/dev-server/index.ts
@@ -24,7 +24,12 @@ import * as WebpackDevServer from 'webpack-dev-server';
 import { checkPort } from '../angular-cli-files/utilities/check-port';
 import { BrowserBuilder, NormalizedBrowserBuilderSchema, getBrowserLoggingCb } from '../browser/';
 import { BrowserBuilderSchema } from '../browser/schema';
-import { normalizeAssetPatterns, normalizeFileReplacements, normalizeSourceMaps } from '../utils';
+import {
+  normalizeAssetPatterns,
+  normalizeFileReplacements,
+  normalizeOptimization,
+  normalizeSourceMaps,
+} from '../utils';
 const opn = require('opn');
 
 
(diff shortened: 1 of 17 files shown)


packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -214,13 +231,13 @@ export class DevServerBuilder implements Builder<DevServerBuilderOptions> {
         htmlAcceptHeaders: ['text/html', 'application/xhtml+xml'],
       } as WebpackDevServer.HistoryApiFallbackConfig,
       stats: false,
-      compress: browserOptions.optimization,
+      compress: styles || scripts,
```
alan-agius4: Not sure, why this is not always `true`

packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -214,13 +231,13 @@ export class DevServerBuilder implements Builder<DevServerBuilderOptions> {
...
       watchOptions: {
         poll: browserOptions.poll,
       },
       https: options.ssl,
       overlay: {
-        errors: !browserOptions.optimization,
+        errors: !(styles || scripts),
```
alan-agius4: Not sure, why is based on optimisation flag

packages/angular_devkit/build_angular/src/dev-server/index.ts
filipesilva: I think it was to make the dev server faster, by not running compression on dev builds.

packages/angular_devkit/build_angular/src/dev-server/index.ts
filipesilva: I guess to not show errors on prod builds? But we already heavily discourage the server from being used in prod, since it's not a dev server. So I think it's ok to always be true.

alan-agius4:
Super seeded by https://github.com/angular/angular-cli/pull/13120

Issue #13095, opened 2018-11-29
Package "@angular/compiler-cli" has an incompatible peer dependency to "typescript" (requires ">=3.1.1 <3.2", would install "3.2.1")


billfranklin:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Command (mark with an `x`)
```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

                  Node Version: v8.11.4
                  npm Version: 6.4.1
                  ng Version: 
                  Angular CLI: 7.0.4
                  Node: 8.11.4
                  OS: win32 x64
                  Angular: 7.0.2
                  ... animations, common, compiler, compiler-cli, core, forms
                  ... http, language-service, platform-browser
                  ... platform-browser-dynamic, router

                  Package                           Version
                  -----------------------------------------------------------
                  @angular-devkit/architect         0.10.4
                  @angular-devkit/build-angular     0.10.4
                  @angular-devkit/build-optimizer   0.10.4
                  @angular-devkit/build-webpack     0.10.4
                  @angular-devkit/core              7.0.4
                  @angular-devkit/schematics        7.0.4
                  @angular/cli                      7.0.4
                  @ngtools/webpack                  7.0.4
                  @schematics/angular               7.0.4
                  @schematics/update                0.10.4
                  rxjs                              6.3.3
                  typescript                        3.1.6
                  webpack                           4.19.1

### Repro steps
                  ng update --all

### The log given by the failure

                  Package "@angular/compiler-cli" has an incompatible peer dependency to "typescript" (requires ">=3.1.1 <3.2", would install "3.2.1")
                  Package "@angular/compiler-cli" has an incompatible peer dependency to "typescript" (requires ">=3.1.1 <3.2", would install "3.2.1").
Incompatible peer dependencies found. See above.

KrzysztofKarol:
Hi @billfranklin 
It's normal behavior.
Latest `@angular/angular-cli:7.1.0` is shipped with `typescript:3.1.6` and supports only patches: `typescript < 3.2`.
[...]

PR #13123, opened 2018-12-03
#13095 Update package.json

rakishii13:


rakishii13:
I signed it!

rakishii13:
Okay, so should the compiler get that update first then?

Issue #6286, opened 2017-05-12
Add Coverage to e2e tests


BenjaminSchubert:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
```

### Versions.

N.A.

### Repro steps.

N.A.

### The log given by the failure.

N.A.

### Desired functionality.

I'd like to see end to end tests being able to also have test coverage as for normal tests.

This is useful when doing a project to make sure that nothing is out of the spec.

I'd be glad to help if this feature is accepted, but would need a bit of guidance on getting started. 

Thanks a lot !

sumitarora:
@BenjaminSchubert AngularCLI uses protractor to run `e2e` tests, which doesn't support code coverage. https://github.com/angular/protractor/issues/370 Also having `e2e` do the code coverage won't be that much feasible as you would have to write `e2e` test for each and every case, branch etc.  which can easily be tested using Unit Tests. E2E tests takes more time to run than unit tests so if you have lot of e2e tests covering each and every case it will take lot of time.

BenjaminSchubert:
@sumitarora Well, there is at least a grunt plugin working for protractor and coverage, which means it is feasible.

I also know that e2e tests are taking much longer to run than unittests. But the aim would be to be able to know roughly if you ran through all your specification or if you have dead code. Moreover, simple tests that do not require unit tests could be covered easily by these. I therefore think that having both coverage is helpful

cmelion:
@sumitarora, FWIW we routinely get 100% coverage on our E2E tests for small to medium sized web-apps.

```
    plugins : [{
        path: 'node_modules/protractor-istanbul-plugin',
        outputPath: 'reports/feature/coverage'
    }]
```

Stalyon:
Hi @cmelion,
I am interested in this coverage report.
Where did you add these lines? In protractor.conf.js?

cmelion:
@Stalyon yep (protractor.feature.coverage.conf.js), we have separate conf files for coverage, mostly to avoid dealing with instrumentation obfuscating the code when debugging during normal test runs.

nrazvan87:
@cmelion I've added this plugin and the settings in `protractor.conf.js`.
After that, I've ran: `ng e2e`. No reports are generated, do you have any clues where I should investigate? No extra errors/logs are displayed.

cmelion:
[...]

PR #13131, opened 2018-12-04
feat(@angular-devkit/build-angular): code coverage for ng build, ng serve and ng e2e

divdavem:
This PR is related to #6286 and allows `--code-coverage` and `--code-coverage-exclude` options to be used with `ng build`, `ng serve` and `ng e2e`, so that it is easy to instrument the code for coverage.

When used with these commands, the `--code-coverage` switch does not automatically create a file with the report but allows the developers to access the `__coverage__` variable from Istanbul.js inside the browser to get coverage data.
That object can then be used to generate a report as described in https://istanbul.js.org/docs/advanced/coverage-object-report/

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
@@ -69,12 +69,12 @@ export interface BuildOptions {
   lazyModules: string[];
   platform?: 'browser' | 'server';
   fileReplacements: CurrentFileReplacement[];
-}
 
-export interface WebpackTestOptions extends BuildOptions {
   codeCoverage?: boolean;
   codeCoverageExclude?: string[];
 }
+
+export interface WebpackTestOptions extends BuildOptions {}
 
 export interface WebpackConfigOptions<T = BuildOptions> {
   root: string;
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -6,8 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import * as CopyWebpackPlugin from 'copy-webpack-plugin';
+import * as glob from 'glob';
 import * as path from 'path';
-import { HashedModuleIdsPlugin, debug } from 'webpack';
+import { HashedModuleIdsPlugin, Rule, debug } from 'webpack';
 import { AssetPatternObject } from '../../../browser/schema';
 import { BundleBudgetPlugin } from '../../plugins/bundle-budget';
 import { CleanCssWebpackPlugin } from '../../plugins/cleancss-webpack-plugin';
@@ -32,7 +33,7 @@ export const buildOptimizerLoader: string = g['_DevKitIsLocal']
 
 // tslint:disable-next-line:no-big-function
 export function getCommonConfig(wco: WebpackConfigOptions) {
-  const { root, projectRoot, buildOptions } = wco;
+  const { root, projectRoot, buildOptions, sourceRoot: include } = wco;
   const { styles: stylesOptimization, scripts: scriptsOptimization } = buildOptions.optimization;
   const {
     styles: stylesSouceMap,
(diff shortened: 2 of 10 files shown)


divdavem:
@filipesilva Thank you for your answer. I am glad that a human (and not only bots) looked at my PR!

> It is unclear to me how it is meant to be used. 

I explained in [this comment of #6286](https://github.com/angular/angular-cli/issues/6286#issuecomment-444902960) how this new feature can be used with e2e tests: 

> For info, I have opened PR #13131 to simplify the instrumentation of files, so that it is possible to use `--code-coverage` and `--code-coverage-exclude` with `ng build` `ng serve` or `ng e2e`.
> This only instruments files as part of the build done by `ng`.
> 
> To get a coverage report, it is then still necessary to record the coverage, for example with an `onComplete` function in the protractor configuration:
> 
> ```js
> async onComplete() {
>   console.log('Retrieving coverage...');
>   const coverage = await browser.executeScript('return JSON.stringify(window.__coverage__);');
>   const coverageFile = path.join(__dirname, ".nyc_output", "out.json");
>   console.log(`Coverage retrieved (${coverage.length} bytes), saving it to ${coverageFile}`);
>   fs.writeFileSync(coverageFile, coverage);
>   console.log("Coverage saved successfully!");
> };
> ```
> 
> And then it is possible to produce both the lcov and html reports with
> 
> ```
> nyc report --reporter=lcov --report-dir=coverage-e2e
> ```

The simple change that I did in this PR simplifies a lot the current steps (detailed in https://github.com/edvlucas/angular-e2e-coverage) that have to be done to instrument the code to get a coverage report for e2e tests.

More generally, I think that being able to produce instrumented code that can be used just like non-instrumented code but also fills a `__coverage__` variable with coverage information when it is executed is interesting.

> There are also no tests, which are required for new features.

I will add some tests.

> using that PR and running ng e2e --code-coverage does not give me any coverage over end-to-end tests

If you think it is required to include, as part of this PR, the production of the coverage report for e2e tests (instead of only the instrumentation), we indeed, as you said, have to discuss how to best implement the collection of coverage results.
A basic idea would be to wrap the configuration of protractor to add something similar to the above `onComplete` function. How should we do that? We could call protractor with a different file than the one configured, and that file would call the original configuration file, and return a modified configuration... but then we would need to also update the paths that are interpreted as relative to the configuration file, and this does not seem very clean.
Also, in order to support all use cases, it is probably not sufficient to only collect coverage at the end from the current window, because e2e tests can possibly cause multiple pages to be opened or navigated to in the browser. In order to get the complete coverage, it is needed to collect the coverage from all browser windows that run instrumented code.
To solve this issue, another idea would be to let application developers call a provided function (with an optional reference to a browser window as an argument) from their protractor configuration or from their tests to collect coverage from that browser window (or the default browser window) at that time...

Please tell me what you think.

divdavem:
Thank you for your answer.

We could also add something like the following `onComplete` function as part of the default `protractor.conf.js` file:

```js
  async onComplete() {
    const coverageData = await browser.executeScript('return JSON.stringify(window.__coverage__);');
    if (coverageData) {
      const { createReporter: createCoverageReporter } = require('istanbul-api');
      const { createCoverageMap } = require('istanbul-lib-coverage');
      const coverageMap = createCoverageMap(JSON.parse(coverageData));
      const reporter = createCoverageReporter();
      reporter.dir = require('path').join(__dirname, 'coverage-e2e');
      reporter.addAll(['json', 'lcov', 'text'])
      reporter.write(coverageMap);
    }
  }
```

This would automatically generate the report in the `coverage-e2e` folder for simple cases if `--code-coverage` is used, and still let users change the logic if needed (if they reload a page or open multiple pages in their tests).

mgechev:
Thank you for the contribution! We discussed this feature today. Although we think it's out of the scope of the CLI at this point, we believe it could be beneficial for our users. I'd recommend you to look at [ngx-build-plus](https://github.com/manfredsteyer/ngx-build-plus) and consider letting folks take advantage of e2e test coverage through a plugin.

beckmx:
@divdavem I need this feature, my company is using cypress and I need a way to extract the coverage, is your feature implemented in a package or something?

divdavem:
> @divdavem I need this feature, my company is using cypress and I need a way to extract the coverage, is your feature implemented in a package or something?

@beckmx I did not implement the feature in a package. You can have a look at [this PR I did on ng-bootstrap](https://github.com/ng-bootstrap/ng-bootstrap/pull/2935) for an example on how to implement it on an application.

PR #13132, opened 2018-12-04
fix(@schematics/angular): fix #12886, import zone related flags in other file.

### LINKED ISSUE #12886: Configuring zone.js in `src/polyfills.ts` won't work, comments are misleading
```
- [ ] feature request
```

```
- [X] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

Node v8.11.1
NPM 6.3.0
Angular CLI: 7.0.4
### Repro steps

Project skeleton file `src/polyfills.ts` contains incorrect comments about how to configure zone.js.

polyfills.ts contains following comments, but if developer open any of comments like 
`(window as any).__Zone_disable_requestAnimationFrame = true;`, it won't work, since after Webpack compilation, it actually moves `import "zone.js/dist/zone";` to the top of this file together with other `import` statements.

Therefore Browser executes `import "zone.js/dist/zone";` earlier than any of lines like `(window as any).__Zone_disable_requestAnimationFrame = true;`, zone.js will not have chance to read those configuration variables at all.

```ts
/**
 * By default, zone.js will patch all possible macroTask and DomEvents
 * user can disable parts of macroTask/DomEvents patch by setting following flags
 */

 // (window as any).__Zone_disable_requestAnimationFrame = true; // disable patch requestAnimationFrame
 // (window as any).__Zone_disable_on_property = true; // disable patch onProperty such as onclick
 // (window as any).__zone_symbol__BLACK_LISTED_EVENTS = ['scroll', 'mousemove']; // disable patch specified eventNames

 /*
 * in IE/Edge developer tools, the addEventListener will also be wrapped by zone.js
 * with the following flag, it will bypass `zone.js` patch for IE/Edge
 */
// (window as any).__Zone_enable_cross_context_check = true;

/***************************************************************************************************
 * Zone JS is required by default for Angular itself.
 */
import 'zone.js/dist/zone';  // Included with Angular CLI.
```

N/A

### Desired functionality

Developers should be notified that they need to create a separate file maybe call `zone-setting.ts` and put actual `(window as any).__Zone_disable_requestAnimationFrame = true;` in that file. Then `import "./zone-setting.ts";` before `import 'zone.js/dist/zone';` to make it work.

N/A

JiaLiPassion:
fix #12886.

all zone related `flags` now in `polyfills.ts`, which will not work properly, because `webpack` will put `import 'zone.js/dist/zone'` before `setting those flags`, which will make those `flags useless`.

So in this PR, I created a standalone file `zone-flags.ts` to keep the `import` in correct order.

### DIFF
--- a/packages/schematics/angular/application/files/src/polyfills.ts
+++ b/packages/schematics/angular/application/files/src/polyfills.ts
@@ -57,17 +57,26 @@
 /**
  * By default, zone.js will patch all possible macroTask and DomEvents
  * user can disable parts of macroTask/DomEvents patch by setting following flags
+ * because those flags need to be set before `zone.js` being loaded, and webpack
+ * will put import in the top of bundle, so user need to create a separate file
+ * in this directory (for example: zone-flags.ts), and put the following flags
+ * into that file, and then add the following code before importing zone.js.
+ * import './zone-flags.ts';
+ *
+ * The flags allowed in zone-flags.ts are listed here.
+ *
+ * The following flags will work for all browsers.
+ *
+ * (window as any).__Zone_disable_requestAnimationFrame = true; // disable patch requestAnimationFrame
+ * (window as any).__Zone_disable_on_property = true; // disable patch onProperty such as onclick
+ * (window as any).__zone_symbol__BLACK_LISTED_EVENTS = ['scroll', 'mousemove']; // disable patch specified eventNames
+ *
+ *  in IE/Edge developer tools, the addEventListener will also be wrapped by zone.js
+ *  with the following flag, it will bypass `zone.js` patch for IE/Edge
+ *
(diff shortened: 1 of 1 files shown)


JiaLiPassion:
@alan-agius4, thanks for the review, commit message updated.

clydin:
As these are not commonly set options, can you adjust the PR to add additional instructions to the existing comment regarding the need to create an additional file rather than adding an additional file to all projects?

JiaLiPassion:
@clydin, sure, thanks for the review, I updated the comment and removed the new file.

PR #13147, opened 2018-12-07
feat(@schematics/angular): set use-host-property-decorator to false in new apps

beeman:
Inspired by [this twitter thread](https://twitter.com/aaronfrost/status/1070389114959097856), this PR changes the default setting for the `use-host-property-decorator` codelyzer rule. 

The rule gives the developer the impression that the `host` property in the `Component` decorator is deprecated, which is not the case. 

In fact, the [angular/material2](https://github.com/angular/material2) project maintained by the Angular team uses it more than 140 times throughout the project.

Personally I've spend a few hours over the years to work around this warning, refactoring `host` properties to `@HostBinding` as I thought it was deprecated. The Twitter thread above thought me it's not, which is why I think it should be disabled.

Here we can see other people [agree on this](https://twitter.com/aaronfrost/status/1070389114959097856), and have been [for a while](https://github.com/angular/angular.io/issues/1763#issuecomment-264036519).

### DIFF
--- a/packages/schematics/angular/workspace/files/tslint.json
+++ b/packages/schematics/angular/workspace/files/tslint.json
@@ -120,7 +120,7 @@
     "no-output-on-prefix": true,
     "use-input-property-decorator": true,
     "use-output-property-decorator": true,
-    "use-host-property-decorator": true,
+    "use-host-property-decorator": false,
     "no-input-rename": true,
     "no-output-rename": true,
     "use-life-cycle-interface": true,


aaronfrost:
Likewise. I legit thought it was deprecated cause of the warning that is thrown. That is my bad, though. I should have read it closer. 

I do think that the need to add a class to the host element of a component is common enough, and that the syntax to do that with the `@HostBinding` is so much more verbose (and visually messy when Prettier puts the annotation on one line and the hostbinding variable on another line), that it makes sense to turn this rule off. If a company wants to turn it back on, great. More power to them. 

An example of a class I put on host elements a lot is `flex-col`. We all know what this class does based on it's name. That looks like this in the component. 
```
// Host property
host : {
    '[class.flex-col]': 'true'
}

// HostBinding
@HostBinding('class.flex-col')
flexCol = true;
```

So while adding one class is less LOC, it is visually more messy. But what about when I want to add several classes, like I very very often times do. 

```
// Host property
host : {
    '[class.flex-col]': 'true',
    '[class.flex-justify-center]': 'true',
    '[class.flex-align-center]': 'true',
}

// HostBinding
@HostBinding('class.flex-col')
flexCol = true;
@HostBinding('class.flex-justify-center')
flexJustifyCenter = true;
@HostBinding('class.flex-align-center')
flexAlignCenter = true;
```

It becomes quickly apparent that the host property is easier to read and should be prescribed and preferred to the latter. 

My two cents. I love Angular. I love the Angular team. I think that new projects would make more sense if this rule was off by default.

clydin:
Note that this rule exists and is enabled by default due to the style guide's recommendation: https://angular.io/guide/styleguide#style-06-03.

The rule's implementation (or rather lack of options) is more of the issue.  The examples from the style guide demonstrate why the decorators are preferred **if** data binding is used.
However, in the cases where the value is a constant, the host property is indeed less verbose and does not suffer from the concerns of the host property data binding scenario.

Instead of disabling the rule, a more appropriate action may be to open an issue/PR with codelyzer to either adjust, expand, or create a new rule that analyzes how the host property is used rather than only that it is used.

aaronfrost:
I think that is a fair assessment of what is going on. I think that it may be the best way forward as well.

beeman:
Good point, I agree with @clydin's comment. Closing the PR.

PR #13156, opened 2018-12-09
docs: remove space for webpack-bundle-analyzer url

PascalHonegger:
The url to the webpack-bundle-analyzer website had a space in it and was therefore not a valid url. Removing the space makes the url valid (and easily clickable).


PascalHonegger:
Done 👍

PR #13159, opened 2018-12-10
refactor: add time logs in `_listLazyRoutesFromProgram`

alan-agius4:
refactor: add time logs in `_listLazyRoutesFromProgram`

Add these time logs as at the moment we are not logging anything for the discovery lazy loaded modules

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -21,7 +21,6 @@ import {
   DEFAULT_ERROR_CODE,
   Diagnostic,
   EmitFlags,
-  LazyRoute,
   Program,
   SOURCE,
   UNKNOWN_ERROR_CODE,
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -423,25 +422,31 @@ export class AngularCompilerPlugin {
...
-        this.entryModule.path + '#' + this.entryModule.className,
-      );
+      entryRoute = this.entryModule.path + '#' + this.entryModule.className;
     } else {
-      lazyRoutes = (this._program as Program).listLazyRoutes();
+      ngProgram = this._program as Program;
     }
 
+    time('AngularCompilerPlugin._listLazyRoutesFromProgram.listLazyRoutes');
+    const lazyRoutes = ngProgram.listLazyRoutes(entryRoute);
```
clydin: Can you add a comment here that in AOT,  the entry point isn’t used and all routes within the program are returned regardless of entry point?

alan-agius4: done.

Issue #10614, opened 2018-05-03
ng update gives error when dependency is local only


hodlbank:
### Versions
```
Angular CLI: 6.0.0-rc.10
Node: 8.11.1
OS: linux x64
Angular: 6.0.0-rc.6
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.5.13
@angular-devkit/core         0.5.13
@angular-devkit/schematics   0.5.13
@angular/cdk                 6.0.0-rc.14
@angular/cli                 6.0.0-rc.10
@angular/material            6.0.0-rc.14
@schematics/angular          0.5.13
@schematics/update           0.5.13
rxjs                         6.0.0
typescript                   2.8.3
```

### Repro steps

* have Angular 5.x project with local-only dependencies, i.e. dependency added into package.json but does not exist in the npmjs.org repository .
* run `ng update @angular/cli` in order to convert .angular-cli.json into proper .angular.json

### Observed behavior
```
> ng update @angular/cli --migrate-only --from=1.7.4
Registry returned 404 for GET on https://registry.npmjs.org/some-private-package
```

### Desired behavior
I expect that ng update will just convert my .angular-cli.json into proper .angular.json without digging into my dependencies.

james-tonicanalytics:
I am also experiencing this issue.

james-tonicanalytics:
The temporary solution that stopped the error occurring was to delete the local dependency from the package.json file and then run ng update @angular/cli. Once the .angular-cli.json was migrated to angular.json

jycouet:
Hello, 
I have the same issue!

What is really strange is that I have 2 local packages... And when I removed one, it's working!

scottleedavis:
I have this problem too.  Also I have the same problem with a github/gitlab reference, see #10642

thorsten:
Same issue here.

Al-Sharif:
Same here

daniloff200:
hey @hansl, @mgechev 
I think, this issue can be closed, cause #13167 was merged

mgechev:
Please, open a new issue if this is still reproducible with the latest CLI, following the template and providing a minimal demo.

PR #13167, opened 2018-12-11
fix(@schematics/update): ignore npm 404 error

tiaguinho:
This PR will fix the issue https://github.com/angular/angular-cli/issues/10614

### DIFF
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -9,8 +9,8 @@ import { logging } from '@angular-devkit/core';
 import { existsSync, readFileSync } from 'fs';
 import { homedir } from 'os';
 import * as path from 'path';
-import { Observable, from } from 'rxjs';
-import { shareReplay } from 'rxjs/operators';
+import { EMPTY, Observable, from } from 'rxjs';
+import { catchError, shareReplay } from 'rxjs/operators';
 import { NpmRepositoryPackageJson } from './npm-package-json';
 
 const ini = require('ini');
@@ -133,7 +133,15 @@ export function getNpmPackageJson(
     },
   );
 
-  const response = from<NpmRepositoryPackageJson>(resultPromise).pipe(shareReplay());
+  // TODO: find some way to test this
+  const response = from<NpmRepositoryPackageJson>(resultPromise).pipe(
+    shareReplay(),
+    catchError(err => {
+      logger.warn(err.message || err);
+
+      return EMPTY;
+    }),
+  );
   npmPackageJsonCache.set(packageName, response);
 
   return response;


tiaguinho:
What do you guys think?
cc: @hansl @clydin

tiaguinho:
@alan-agius4 just did it!

tiaguinho:
@clydin we can use the _workspace_ configuration to search local packages and filter ```_getAllDependencies```.
This way, only *npm* packages will be pass to ```getNpmPackageJson```.

tiaguinho:
Changed!

cc: @alan-agius4 @clydin

packages/schematics/update/update/npm.ts
```diff
@@ -133,7 +133,14 @@ export function getNpmPackageJson(
     },
   );
 
-  const response = from<NpmRepositoryPackageJson>(resultPromise).pipe(shareReplay());
+  const response = from<NpmRepositoryPackageJson>(resultPromise).pipe(
+    shareReplay(),
+    catchError(err => {
+      logger.warn(err);
+
+      return EMPTY;
```
clydin: use `empty()` instead of `EMPTY`.

packages/schematics/update/update/npm.ts
```diff
@@ -133,7 +133,14 @@ export function getNpmPackageJson(
     },
   );
 
-  const response = from<NpmRepositoryPackageJson>(resultPromise).pipe(shareReplay());
+  const response = from<NpmRepositoryPackageJson>(resultPromise).pipe(
+    shareReplay(),
+    catchError(err => {
+      logger.warn(err);
```
clydin: Can you change this to `logger.warn(err.message || err)`?  The user doesn't need the full stack trace.

tiaguinho:
Hi @clydin just made the changes you asked.
If I miss something, please let me know.

Thanks!

packages/schematics/update/update/npm.ts
hansl: Disregard, `EMPTY` is fine (see https://rxjs.dev/api/index/const/EMPTY)

packages/schematics/update/update/npm.ts
```diff
@@ -133,7 +133,14 @@ export function getNpmPackageJson(
     },
   );
 
-  const response = from<NpmRepositoryPackageJson>(resultPromise).pipe(shareReplay());
+  const response = from<NpmRepositoryPackageJson>(resultPromise).pipe(
+    shareReplay(),
+    catchError(err => {
+      logger.warn(err.message || err);
+
+      return empty();
```
hansl: Sorry, please use `EMPTY` (see https://rxjs.dev/api/index/const/EMPTY).

hansl:
Additionally, could you add a `// TODO: find some way to test this`? Testing npm requests is hard, but we should have proper integration tests with this (in some future).

packages/schematics/update/update/npm.ts
tiaguinho: hahaha, no problem.

PR #13176, opened 2018-12-12
update build optimizer to use async source-map api

alan-agius4:
build: update source-map to `0.7.3`

Remove `@types/source-map` as this is a stub types definition for source-map (https://github.com/mozilla/source-map). source-map provides its own type definitions.

Also, update source-map to ``0.7.3` as currently certain packages like devkit core are already on this version


clydin:
This doesn’t work without changes to the code coverage handling as well as changing all direct usage of the package to the new async API.

clydin:
I actually had a PR for the build optimizer part but dropped it until the dev code coverage issues could be addressed.

PR #13178, opened 2018-12-12
ci: add renovate configuration

alan-agius4:
Since we have a lot of pin dependencies renovate will help us automate dependency updates.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,47 @@
+{
+  "pinVersions": false,
+  "semanticCommits": true,
+  "semanticPrefix": "build",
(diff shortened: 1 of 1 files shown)


clydin:
We should probably exclude all package.json templates inside schematic packages

alan-agius4:
@alexeagle I add a rate limiting of 2 PR / hour. and scheduled it run between 10pm - 4am every weekday (Pacific TimeZone) and run freely during the weekends.

alexeagle:
I don't think this scheduling actually improves it, we can still get 12 PR's per day. I'd rather get one PR a week or so. But let's try this and see how it works.

PR #13197, opened 2018-12-14
fix(@angular/cli): remove redundant period in deprecation warning

cexbrayat:
As `parser.ts` currently always adds a period at the end of the warning message and the `x-deprecated` messages also have one, this removes the always added one to avoid a double period in the warning message.

Fixes:

    Option "styleext" is deprecated: Use "style" instead..

to:

    Option "styleext" is deprecated: Use "style" instead.

### DIFF
--- a/packages/angular/cli/models/parser.ts
+++ b/packages/angular/cli/models/parser.ts
@@ -206,7 +206,7 @@ function _assignOption(
 
         if (option.deprecated !== undefined && option.deprecated !== false) {
           warnings.push(`Option ${JSON.stringify(option.name)} is deprecated${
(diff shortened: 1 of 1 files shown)


cexbrayat:
@alan-agius4 OK, I hesitated :) 
I'll update and ping you when done.

cexbrayat:
@alan-agius4 Updated, let me know if that's better.

alan-agius4:
Bdw you’d need to add the period only when there is no description.

alan-agius4:
Small nit: would you mind fixing the commit message scope?

cexbrayat:
@alan-agius4 Sorry, I forgot about it. Done.

PR #13213, opened 2018-12-18
feat(@schematics/angular): extend tslint:recommended

### LINKED ISSUE #6179: [lint] tslint.json could extend tslint:recommended
```
- [X] feature request
```

### Versions.
@angular/cli: 1.1.0-beta.0
node: 7.4.0

### Desired functionality.
TSLint has introduced `tslint:recommended` (and `tslint:latest`) for a while, allowing a config to extend the recommended default, and just override a few rules.

It means we could simplify the configuration, and just keep a few overrides that are common in the Angular community.

See the recommended set of rules: https://github.com/palantir/tslint/blob/master/src/configs/recommended.ts

I gave it a naive first try, this is how the tslint.json would look like (to keep the configuration really close to the one we have right now):

```json
{
  "rulesDirectory": [
    "node_modules/codelyzer"
  ],
  "extends": "tslint:recommended",
  "rules": {
    "array-type": false,
    "arrow-parens": false,
    "interface-name": false,
    "import-blacklist": [true, "rxjs"],
    "max-line-length": [
      true,
      140
    ],
    "member-access": false,
    "member-ordering": [
      true,
      "static-before-instance",
      "variables-before-functions"
    ],
    "no-console": [
      true,
      "debug",
      "info",
      "time",
      "timeEnd",
      "trace"
    ],
    "no-empty": false,
    "no-inferrable-types": [true, "ignore-params"],
    "no-non-null-assertion": true,
    "no-string-literal": false,
    "no-switch-case-fall-through": true,
    "object-literal-sort-keys": false,
    "ordered-imports": false,
    "one-line": [
      true,
      "check-open-brace",
      "check-catch",
      "check-else",
      "check-whitespace"
    ],
    "quotemark": [
      true,
      "single"
    ],
    "trailing-comma": false,
    "variable-name": false,

    "directive-selector": [true, "attribute", "pr", "camelCase"],
    "component-selector": [true, "element", "pr", "kebab-case"],
    "use-input-property-decorator": true,
    "use-output-property-decorator": true,
    "use-host-property-decorator": true,
    "no-input-rename": true,
    "no-output-rename": true,
    "use-life-cycle-interface": true,
    "use-pipe-transform-interface": true,
    "component-class-suffix": true,
    "directive-class-suffix": true,
    "no-access-missing-member": true,
    "templates-use-public": true,
    "invoke-injectable": true
  }
}
```
If you think it is interesting, I can work on a PR.

cexbrayat:
Updates the `tslint.json` file to extend to the `tslint:recommended` set.
Removes the rules that have the same value as the recommended one and explicitely deactivates some rules that are activated in the recommended set, but were not in ours.

Fixes #6179 

cc @mgechev this is to start the discussion, feedback welcome

### DIFF
--- a/packages/schematics/angular/workspace/files/tslint.json
+++ b/packages/schematics/angular/workspace/files/tslint.json
@@ -1,32 +1,25 @@
 {
+  "extends": "tslint:recommended",
   "rulesDirectory": [
     "codelyzer"
   ],
   "rules": {
-    "arrow-return-shorthand": true,
-    "callable-types": true,
-    "class-name": true,
-    "comment-format": [
-      true,
-      "check-space"
-    ],
-    "curly": true,
+    "adjacent-overload-signatures": false,
+    "align": false,
+    "array-type": false,
+    "arrow-parens": false,
+    "ban-types": false,
     "deprecation": {
       "severity": "warn"
     },
-    "eofline": true,
-    "forin": true,
+    "function-constructor": false,
     "import-blacklist": [
       true,
       "rxjs/Rx"
     ],
-    "import-spacing": true,
-    "indent": [
-      true,
-      "spaces"
-    ],
-    "interface-over-type-literal": true,
-    "label-position": true,
+    "interface-name": false,
+    "jsdoc-format": false,
+    "max-classes-per-file": false,
     "max-line-length": [
       true,
       140
@@ -43,8 +36,10 @@
         ]
       }
     ],
-    "no-arg": true,
-    "no-bitwise": true,
+    "new-parens": false,
+    "no-angle-bracket-type-assertion": false,
+    "no-conditional-assignment": false,
+    "no-consecutive-blank-lines": false,
     "no-console": [
       true,
       "debug",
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/workspace/files/tslint.json
```diff
@@ -1,32 +1,25 @@
...
   "rules": {
-    "arrow-return-shorthand": true,
-    "callable-types": true,
-    "class-name": true,
-    "comment-format": [
-      true,
-      "check-space"
-    ],
-    "curly": true,
+    "adjacent-overload-signatures": false,
```
mgechev: Looks like in the current CLI and tslint this has value `true`, lmk if I'm wrong.

cexbrayat: Hmm, I don't think it is present in the current tslint.json, so I guess it's not activated (unless I'm missing something?).

mgechev:
`object-literal-key-quotes` should be also disabled.

cexbrayat:
@mgechev good catch: I added `object-literal-key-quotes: false` 👍

PR #13214, opened 2018-12-18
style(@schematics/angular): format jsdoc in polyfills.ts

cexbrayat:
This line makes the `jsdoc-format` tslint rule fails if activated as it is not a correct JSDoc formating.


alan-agius4:
Can you kindly use the `style` commit message type?

cexbrayat:
@alan-agius4 Done

PR #13224, opened 2018-12-19
fix(@angular-devkit/build-angular): script chunk in angular.json return false then calling isInitial

### LINKED ISSUE #13223: Webpack extra script chunk return canBeInitial as false.
```
```

```
- [x] build
```

### Versions

node --version
v10.12.0
npm --version
6.4.1
Angular CLI: 7.1.1
Node: 10.12.0
OS: darwin x64
Angular: 7.1.1
... router

@angular-devkit/architect         0.11.1
@angular-devkit/build-angular     0.11.1
@angular-devkit/build-optimizer   0.11.1
@angular-devkit/build-webpack     0.11.1
@angular-devkit/core              7.1.1
@angular-devkit/schematics        7.1.1
@angular/elements                 7.1.3
@ngtools/webpack                  7.1.1
@schematics/angular               7.1.1
@schematics/update                0.11.1
typescript                        3.1.6

### Repro steps
When using ngx-build-plus with html-webpack-plugin
The Global scripts chunk is not isInitial.
Since the extra chunk in ScriptsWebpackPlugin has not add with entrypoint,
therefore the html-webpack-plugin will nerver add the script to the html (3.2.0).

### Desired functionality
The lazy options should pass to ScriptsWebpackPlugin.
It the script is not lazy than 
`chunk.addGroup(entrypoint);`

ps: html-webpack-plugin have remove this feather in master but it should leave the option to plugin developer.

Root cause
https://github.com/jantimon/html-webpack-plugin/blob/v3.2.0/index.js#L111
https://github.com/jantimon/html-webpack-plugin/blob/v3.2.0/index.js#L384

Tsuki:
Fixes #13223

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/scripts-webpack-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/scripts-webpack-plugin.ts
@@ -83,7 +83,7 @@ export class ScriptsWebpackPlugin {
 
     const entrypoint = new EntryPoint(this.options.name);
     entrypoint.pushChunk(chunk);
-
+    chunk.addGroup(entrypoint);
     compilation.entrypoints.set(this.options.name, entrypoint);
     compilation.chunks.push(chunk);
     compilation.assets[filename] = source;
--- a/packages/angular_devkit/build_angular/test/browser/scripts-array_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/scripts-array_spec_large.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { DefaultTimeout, runTargetSpec } from '@angular-devkit/architect/testing';
+import { DefaultTimeout, TestLogger, runTargetSpec } from '@angular-devkit/architect/testing';
 import { PathFragment, join, normalize, virtualFs } from '@angular-devkit/core';
 import { tap } from 'rxjs/operators';
 import { browserTargetSpec, host } from '../utils';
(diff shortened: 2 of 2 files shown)


mgechev:
@Tsuki thanks for the PR!

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/scripts-webpack-plugin.ts
```diff
@@ -83,7 +84,9 @@ export class ScriptsWebpackPlugin {
 
     const entrypoint = new EntryPoint(this.options.name);
     entrypoint.pushChunk(chunk);
-
+    if (!this.options.lazy){
+      chunk.addGroup(entrypoint);
```
clydin: Lazy shouldn’t be a guard for this.   The chunk is already part of the group from the pushChunk call.  The issue here is that it was assumed that the pushChunk call performs the reciprocal action.  The correct behavior is to always execute both actions.

Deciding which entrypoints should be present and how they are represented in an HTML fileis a function of the html processing plugin.

Tsuki: In current situation, the chunk [canBeInitial](https://github.com/webpack/webpack/blob/master/lib/Chunk.js#L199) always return false when not put the ChunkGoup in to the chunk. 
The chunk will [filter out](https://github.com/jantimon/html-webpack-plugin/blob/v3.2.0/index.js#L384) even you stat in the filter option.

clydin: Yes.  The only change that is need here is the addition of `chunk.addGroup(entrypoint);`.  The lazy logic is only relevant to the HTML plugin.

Tsuki: ok changed

PR #13226, opened 2018-12-19
Add support for Typescript 3.2

### LINKED ISSUE #13212: `import as` results in the alias being undefined with Typescript 3.2
```
```

```
- [x] serve
- [x] e2e
```

### Versions

```
Angular CLI: 7.1.3
Node: 10.11.0
OS: darwin x64
Angular: 7.2.0-beta.2+55.sha-a6e6763
... animations, common, compiler, compiler-cli, core, elements

@angular-devkit/architect         0.10.7
@angular-devkit/build-angular     0.10.7
@angular-devkit/build-optimizer   0.10.7
@angular-devkit/build-webpack     0.10.7
@angular-devkit/core              7.1.3
@angular-devkit/schematics        7.1.3
@angular/cdk                      7.0.0
@angular/cli                      7.1.3
@angular/material                 7.0.0
@ngtools/webpack                  7.0.7
@schematics/angular               7.1.3
@schematics/update                0.11.3
typescript                        3.2.2
webpack                           4.19.1
```

### Repro steps

After typescript 3.2 upgrade, I've started seeing e2e test failures due to certain symbols not being defined - in all cases these symbols came from rxjs and were aliased to a different name.

The problem occurs during dev builds without optimizations turned on and causes both `ng serve` and `ng e2e` to fail with runtime errors in the browser

For example: 
```
import { from as fromPromise, Observable } from 'rxjs';
``` 
Would end up with `fromPromise` being undefined at runtime.

### Workarounds

- remove the `as` alias and refactor the code to use the original symbol name.
- revert back to typescript 3.1

### More info

I'll try to create a minimal reproduction outside of aio but for now you can review the "workaround" commit in https://github.com/angular/angular/pull/27536/commits

In the meantime I've tested the `as` imports in vanilla tsc and it seems to be working just fine there, so I suspect that it's some kind of interaction with cli or webpack that is causing this.

alan-agius4:
- Update to TypeScript 3.2

- fix(@ngtools/webpack): `import as` results in the alias being undefined with Typescript 3.2

When using the `specifier.propertyName` with `typeChecker.getSymbolAtLocation` it will return a more detailed symbol then we originally have in the `usedSymbols` set.

We should probably use `symbol.id` to actually check if the symbols are the same, however the `id` is not exposed in the Symbol interface.

Using `node.name` will return the same symbol that we have stored in the set.

Fixes #13212


alan-agius4:
The 1Kb size change is due to the Angular 7.2 upgrade.

etc/api/angular_devkit/schematics/tools/index.d.ts
```diff
@@ -147,4 +147,4 @@ export declare class SchematicNameCollisionException extends BaseException {
     constructor(name: string);
 }
 
-export declare function validateOptionsWithSchema(registry: schema.SchemaRegistry): <T extends {}>(schematic: FileSystemSchematicDescription, options: T, context?: import("@angular-devkit/schematics/src/engine/interface").TypedSchematicContext<import("@angular-devkit/schematics/tools/tools/description").FileSystemCollectionDescription, FileSystemSchematicDescription> | undefined) => Observable<T>;
+export declare function validateOptionsWithSchema(registry: schema.SchemaRegistry): <T extends {}>(schematic: FileSystemSchematicDescription, options: T, context?: import("@angular-devkit/schematics").TypedSchematicContext<import("@angular-devkit/schematics/tools/tools/description").FileSystemCollectionDescription, FileSystemSchematicDescription> | undefined) => Observable<T>;
```
filipesilva: Why did this API file have to change?

clydin:
Actually we could probably clean that up some and grab the aliased symbol if present when looping through the file.

alan-agius4:
@clydin I have added some more tests, can you check if that's the tests you had in mind?

PR #13234, opened 2018-12-19
fix(schematics/angular): assert no error logs in e2e

kyliau:
This commit adds an assertion to the e2e test to make sure there are
no error logs emitted by the browser.

### DIFF
--- a/packages/schematics/angular/e2e/files/src/app.e2e-spec.ts
+++ b/packages/schematics/angular/e2e/files/src/app.e2e-spec.ts
@@ -1,4 +1,5 @@
 import { AppPage } from './app.po';
+import { browser, logging } from 'protractor';
 
 describe('workspace-project App', () => {
   let page: AppPage;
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/e2e/files/src/app.e2e-spec.ts
```diff
@@ -11,4 +12,13 @@ describe('workspace-project App', () => {
     page.navigateTo();
     expect(page.getTitleText()).toEqual('Welcome to <%= relatedAppName %>!');
   });
+
+  afterEach(async () => {
+    // Assert that there are no errors emitted from the browser
+    const logs = await browser.manage().logs().get('browser');
+    const errors = logs.filter((entry: logging.Entry) => {
```
hansl: Is using `some()` easier to read here?

```typescript
expect(logs.some((entry: logging.entry) => entry.level === logging.level.SEVERE)).toBe(false);
```

hansl: Or go fancy (better reading but more obscure API parts):

```typescript
expect(logs).not.toContain(jasmine.objectContaining({ level: logging.level.SEVERE }));
```

kyliau: I prefer the second approach because then jasmine would print the actual logs when the assertion fails.
The first approach would just print `expect  true to be false`, which is not helpful.

PR #13249, opened 2018-12-20
fix(@schematics/angular): generate coverage for libs in their own folder

ahasall:
BEFORE:
`ng test foo --code-coverage` would generate the code coverage in the `coverage` folder and would overwrite the code coverage folder of the main app.

AFTER:
Now the code coverage resides in its own folder (`coverage/foo`).

This mimic the behaviour of the `ng build` command and adds more consistency.

### DIFF
--- a/packages/schematics/angular/application/files/root/karma.conf.js
+++ b/packages/schematics/angular/application/files/root/karma.conf.js
@@ -16,7 +16,7 @@ module.exports = function (config) {
       clearContext: false // leave Jasmine Spec Runner output visible in browser
     },
     coverageIstanbulReporter: {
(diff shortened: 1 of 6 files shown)


ahasall:
Thank you @alan-agius4, for the review. I think also that we should to the same for the main app.

ahasall:
I update the PR to do the same for the 'application' schematic

mgechev:
@ahasall any updates?

ahasall:
Hi @mgechev, Did you mean to tag @hansl, I thought that we were waiting for his thoughts.

hansl:
This is a new "feature" at least, should probably be in 7.3 but not 7.2

ahasall:
Therefore do you want me to change the commit message?

PR #13253, opened 2018-12-20
docs: typo 'they changes' -> 'the changes'

bartduisters:


hansl:
Sorry this goes to master only since it's about the wiki. We don't update the wiki anymore so I have to check internally.

PR #13273, opened 2018-12-21
build: update several dependencies

alan-agius4:
This is so that renovate won't trigger for these dependencies


package.json
```diff
@@ -70,7 +70,7 @@
     "temp": "^0.9.0",
     "tslint": "^5.11.0",
     "typescript": "~3.2.2",
-    "yarn": "^1.10.1"
+    "yarn": "^1.12.3"
```
clydin: why is this here at all?

PR #13284, opened 2018-12-22
feat(@ngtools/webpack): allow .svg files as templates

### LINKED ISSUE #10567: Using `templateUrl` with a `.svg` file
Hey friendly @angular/cli team!

Would there be any appetite to enable using a `.svg` file as a component template?

As far as I can tell it would be a small change in `webpack-config/common.js` from:

```javascript
{ test: /\.html$/, loader: 'raw-loader' },
{ test: /\.(eot|svg|cur)$/, // ...
```

to 

```javascript
{ test: /\.(html|svg)$/, loader: 'raw-loader' },
{ test: /\.(eot|cur)$/, // ...
```

Are there any downsides to this? Are there any other places that the change would need to be handled? It seems to work when I change it in my node_modules.

Let me know if this is something that would be useful/wanted, and I'll make a PR.

Thanks!

oocx:
With directTemplateLoading enabled, components 
can now use .svg files as templates. For AOT builds, 
the Angular compiler host now reads .svg files 
directly when reading component templates. 
For JIT builds, replaceResources creates a require call
that directly uses raw-loader instead of using the 
loader provided by the current webpack configuration.

Closes #10567

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -307,7 +307,6 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
     },
     module: {
       rules: [
-        { test: /\.html$/, loader: 'raw-loader' },
         {
           test: /\.(eot|svg|cur|jpg|png|webp|gif|otf|ttf|woff|woff2|ani)$/,
           loader: 'file-loader',
--- a/packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
@@ -1,0 +1,59 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { runTargetSpec } from '@angular-devkit/architect/testing';
+import { join, normalize, virtualFs } from '@angular-devkit/core';
+import { tap } from 'rxjs/operators';
+import { browserTargetSpec, host, outputPath } from '../utils';
+
+describe('Browser Builder allow svg', () => {
+
+  beforeEach(done => host.initialize().toPromise().then(done, done.fail));
+  afterEach(done => host.restore().toPromise().then(done, done.fail));
+
+  it('works with aot',
+    (done) => {
+
+    const svg = `
+      <svg xmlns="http://www.w3.org/2000/svg">
+        <text x="20" y="20" font-size="20" fill="red">Hello World</text>
+      </svg>`;
+
+    host.writeMultipleFiles({
+      './src/app/app.component.svg': svg,
+      './src/app/app.component.ts': `
+        import { Component } from '@angular/core';
+
+        @Component({
+          selector: 'app-root',
+          templateUrl: './app.component.svg',
+          styleUrls: []
+        })
+        export class AppComponent {
+          title = 'app';
+        }
+      `,
+    });
+
+    const overrides = { aot: true };
+
+    runTargetSpec(host, browserTargetSpec, overrides).pipe(
+      tap((buildEvent) => expect(buildEvent.success).toBe(true)),
+      tap(() => {
+        const content = virtualFs.fileBufferToString(
+          host.scopedSync().read(join(outputPath, 'main.js')),
+        );
+
+        expect(content).toContain('":svg:svg"');
+        expect(host.scopedSync().exists(normalize('dist/app.component.svg')))
+          .toBe(false, 'should not copy app.component.svg to dist');
+      }),
+    ).toPromise().then(done, done.fail);
+  });
+
+});
--- a/packages/ngtools/webpack/src/transformers/replace_resources.ts
+++ b/packages/ngtools/webpack/src/transformers/replace_resources.ts
@@ -10,6 +10,7 @@ import * as ts from 'typescript';
 export function replaceResources(
   shouldTransform: (fileName: string) => boolean,
   getTypeChecker: () => ts.TypeChecker,
+  directTemplateLoading = false,
 ): ts.TransformerFactory<ts.SourceFile> {
 
   return (context: ts.TransformationContext) => {
@@ -19,7 +20,7 @@ export function replaceResources(
       if (ts.isClassDeclaration(node)) {
         node.decorators = ts.visitNodes(
           node.decorators,
-          (node: ts.Decorator) => visitDecorator(node, typeChecker),
+          (node: ts.Decorator) => visitDecorator(node, typeChecker, directTemplateLoading),
         );
       }
 
@@ -34,7 +35,10 @@ export function replaceResources(
   };
 }
 
-function visitDecorator(node: ts.Decorator, typeChecker: ts.TypeChecker): ts.Decorator {
+function visitDecorator(
+  node: ts.Decorator,
+  typeChecker: ts.TypeChecker,
+  directTemplateLoading: boolean): ts.Decorator {
   if (!isComponentDecorator(node, typeChecker)) {
     return node;
   }
@@ -56,7 +60,8 @@ function visitDecorator(node: ts.Decorator, typeChecker: ts.TypeChecker): ts.Dec
   // visit all properties
   let properties = ts.visitNodes(
     objectExpression.properties,
-    (node: ts.ObjectLiteralElementLike) => visitComponentMetadata(node, styleReplacements),
+    (node: ts.ObjectLiteralElementLike) =>
+      visitComponentMetadata(node, styleReplacements, directTemplateLoading),
   );
 
   // replace properties with updated properties
@@ -83,6 +88,7 @@ function visitDecorator(node: ts.Decorator, typeChecker: ts.TypeChecker): ts.Dec
 function visitComponentMetadata(
   node: ts.ObjectLiteralElementLike,
   styleReplacements: ts.Expression[],
+  directTemplateLoading: boolean,
 ): ts.ObjectLiteralElementLike | undefined {
   if (!ts.isPropertyAssignment(node) || ts.isComputedPropertyName(node.name)) {
     return node;
@@ -98,7 +104,7 @@ function visitComponentMetadata(
       return ts.updatePropertyAssignment(
         node,
         ts.createIdentifier('template'),
-        createRequireExpression(node.initializer),
+        createRequireExpression(node.initializer, directTemplateLoading ? '!raw-loader!' : ''),
       );
 
     case 'styles':
@@ -133,13 +139,13 @@ function visitComponentMetadata(
   }
 }
 
-export function getResourceUrl(node: ts.Expression): string | null {
+export function getResourceUrl(node: ts.Expression, loader = ''): string | null {
   // only analyze strings
   if (!ts.isStringLiteral(node) && !ts.isNoSubstitutionTemplateLiteral(node)) {
     return null;
   }
 
-  return `${/^\.?\.\//.test(node.text) ? '' : './'}${node.text}`;
+  return `${loader}${/^\.?\.\//.test(node.text) ? '' : './'}${node.text}`;
 }
 
 function isComponentDecorator(node: ts.Node, typeChecker: ts.TypeChecker): node is ts.Decorator {
@@ -155,8 +161,8 @@ function isComponentDecorator(node: ts.Node, typeChecker: ts.TypeChecker): node 
   return false;
 }
 
-function createRequireExpression(node: ts.Expression): ts.Expression {
-  const url = getResourceUrl(node);
+function createRequireExpression(node: ts.Expression, loader = ''): ts.Expression {
+  const url = getResourceUrl(node, loader);
   if (!url) {
     return node;
   }
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -863,7 +863,8 @@ export class AngularCompilerPlugin {
 
     if (this._JitMode) {
       // Replace resources in JIT.
-      this._transformers.push(replaceResources(isAppPath, getTypeChecker));
+      this._transformers.push(
+        replaceResources(isAppPath, getTypeChecker, this._options.directTemplateLoading));
     } else {
       // Remove unneeded angular decorators.
       this._transformers.push(removeDecorators(isAppPath, getTypeChecker));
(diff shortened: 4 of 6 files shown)


alan-agius4:
This might be a breaking change, if for a reason at the moment you have a file named `.component.svg` It’s behaviour will change.

packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
```diff
@@ -0,0 +1,82 @@
...
+import { join, normalize, virtualFs } from '@angular-devkit/core';
+import { tap } from 'rxjs/operators';
+import { browserTargetSpec, host, outputPath } from '../utils';
+
+describe('Browser Builder allow svg', () => {
+
+  beforeEach(done => host.initialize().toPromise().then(done, done.fail));
+  afterEach(done => host.restore().toPromise().then(done, done.fail));
+
+  it('can use .component.svg as templates (uses raw-loader and does not copy .svg to dist)',
```
alan-agius4: We’d need to add addition tests, such to test that with AOT, resources get inlined as expected.

packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
```diff
@@ -0,0 +1,82 @@
...
+      './src/other.svg': `<svg></svg>`,
+      './src/main.ts': `declare var require: Function;
+      const x = require('./other.svg');
+      console.log(x);`,
+    });
+
+    host.replaceInFile(
+      'tsconfig.json',
+      '"target": "es5"',
+      '"target": "es5", "allowJs": true',
```
alan-agius4: Why allowJs is needed?

oocx: That's just copied from the test that @filipesilva mentioned could be used as  example: https://github.com/angular/angular-cli/issues/10567#issuecomment-435126150

It's probably not necessary and can be removed.

packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
oocx: I can give it a try. Making the change to the webpack.config took me about 10 minutes, figuring out how tests work in angular-cli and writing those tests took about  5 hours, so I might need some help to do that.

alan-agius4: I am here to answer any questions.

oocx:
Should I just add a breaking change note to the commit message? I can't think of a way to implement this that's not a potentially breaking change.

alan-agius4:
I think we should try to use the `issuer`. See: 
https://github.com/angular/angular-cli/issues/10567#issuecomment-435364491

oocx:
This could also be a breaking change if you already import .svg from .js files.

packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
oocx: For the test with a ```@Component``` and aot, the test in styles_spec_large.ts only tests that the build is successful, but does not test if the aot compiler output is correct. How would I do that here? 

I could try to figure out what the correct compiler output would look like and do a contains-test for that in main.js. Would that be the correct way to test this? However, such a test would probably break when ivy is enabled, as the compiler output would be different.

filipesilva:
I'm not too keen on the `.component.svg` convention. We've tried to stay away from having to name your files in a certain way to use them because it's a somewhat brittle convention, and it's hard to inform users of it as well. I will bring it up with the team to see what everyone thinks.

oocx:
I've just spent several hours debugging webpack. For some reason I don't understand yet, webpack does not always provide an issuer. 

When you look at webpack\lib\RuleSet.js, there is a method _run that evaluates rules. The first parameter to run is an object "data", which has the object against which the rule should be evaluated. Here, data.issuer is not always set, in many cases (especially in my experiments with using a .svg as template), issuer is set to "".

I currently don't see a way to use issuer here without hours of reverse engineering webpack. 

I'll ask for help in the webpack gitter. However, I already spent about 8 hours today for a simple change that is not even 5 lines of configuration (maybe 15 minutes for the actual change, a few hours to figure out how to unit test it, then again several hours to work on feedback). If I don't find a solution soon, the problem is just not worth investing more time.

oocx:
@alan-agius4 as requested, I've added a test that uses a component and aot. Is this what you had in mind, or am I still missing something?

I'm still switching loaders based on the .component.svg suffix, as the idea to use the issuer does not seem to work.

packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
```diff
@@ -0,0 +1,106 @@
...
+  beforeEach(done => host.initialize().toPromise().then(done, done.fail));
+  afterEach(done => host.restore().toPromise().then(done, done.fail));
+
+  it('can use .component.svg as templates (uses raw-loader and does not copy .svg to dist)',
+    (done) => {
+
+    host.writeMultipleFiles({
+      './src/app.component.svg': `<svg></svg>`,
+      './src/main.ts': `declare var require: Function;
+      const x = require('./app.component.svg');
```
clydin: This syntax is non-standard and not officially supported and shouldn't be part of a test.  This applies to the second test below as well.  The feature being added is SVG support within a Component template;  that use case is what should be tested.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -294,9 +294,10 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
     },
     module: {
       rules: [
-        { test: /\.html$/, loader: 'raw-loader' },
+        { test: /\.(html|component.svg)$/, loader: 'raw-loader' },
```
clydin: Special naming rules should be avoided.

By changing the replace resources transformer [here](https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/transformers/replace_resources.ts#L142) to inline the raw loader usage, the need for the naming convention can be avoided.  The raw-loader rule could actually be removed completely with such a change.

This is untested, but something similar to the following should work:
```
return `!raw-loader!${/^\.?\.\//.test(node.text) ? '' : './'}${node.text}`;
```

oocx: I tested your idea, it works. However, the same code is also used for loading html templates and styles. Loaders for css preprocessors would not get invoked any more. It would also make it impossible to customize loaders for .html templates.

If we only use raw-loader here when the template is an .svg file, then we would not break existing code. Something like:
```
if (/\.svg$/.test(node.text)) {
    return `!raw-loader!${/^\.?\.\//.test(node.text) ? '' : './'}${node.text}`;
} else {
    return `${/^\.?\.\//.test(node.text) ? '' : './'}${node.text}`;
}
```
What do you think? If everyone agrees, I'll update the PR with that change and replace the unit tests I added for the previous solution. I would add tests in replace_resources_spec.ts. For the change in compiler_host.ts, I'm not sure yet what the best place to test this would be, as there is no compiler_host_spec.ts.

clydin: The suggested line above should only be applied to `templateUrl` augmentations.  `styleUrls` behavior should remain unchanged.  The plugin has an option named `directTemplateLoading` which is enabled by the CLI and would be the guard for the inline 'raw-loader` usage.

oocx: `getResourceUrl` is called by `getRequireExpresion` which is also used when processing the `styleUrls` property in `visitComponentMetadata`. That's why I suggested that `if`. 

I could also add an argument to `getRequireExpression` and `getResourceUrl` that determines if raw-loader should be used.

It could look something like this:
```
// in visitComponentMetadata:
case 'templateUrl':
            return ts.updatePropertyAssignment(node, ts.createIdentifier('template'), createRequireExpression(node.initializer, '!raw-loader!'));

// createRequireExpression
function createRequireExpression(node: ts.Expression, loader = ''): ts.Expression {
  const url = getResourceUrl(node, loader);
// ...

export function getResourceUrl(node: ts.Expression, loader = ''): string | null {
  // only analyze strings
  if (!ts.isStringLiteral(node) && !ts.isNoSubstitutionTemplateLiteral(node)) {
    return null;
  }

  return `${loader}${/^\.?\.\//.test(node.text) ? '' : './'}${node.text}`;
}
```

packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
alan-agius4: You can try to see if the actual template contain is in the bundle such as a string. That should be enough to determine if it worked or not.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
clydin: After discussing with the team, this solution appears viable.  Can you make the updates and then we can proceed with another round of reviews?

oocx: Done. I hope I did not miss any of the comments.

alan-agius4:
Can you kindly fix the commit message as the feature is bound to `@ngtools/webpack` ?

packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -98,7 +98,7 @@ function visitComponentMetadata(
       return ts.updatePropertyAssignment(
         node,
         ts.createIdentifier('template'),
-        createRequireExpression(node.initializer),
+        createRequireExpression(node.initializer, '!raw-loader!'),
```
alan-agius4: Shouldn't this be done when `directTemplateLoading` ?

oocx: I was assuming this would still be needed for the case 'directTemplateLoading=false'. Or do we only want to support this with directTemplateLoading enabled? This would make the PR much simpler, I would then just remove the changes I made to replace_resources.

alan-agius4: I think it should be when it's `true`, let's confirm with @clydin.

clydin: The check should actually be in both places (and should have been there for the initial direct template loading implementation).
So yes; please add a guard here based on the value of `directTemplateLoading`.
This is to ensure that custom Webpack configurations that use the plugin directly can use their own custom loader configurations to handle templates, if desired.

alan-agius4:
@oocx, are you willing to continue on this and address the above comments? 

Thanks.

oocx:
@alan-agius4 yes, I can probably continue working on it next weekend

oocx:
I hope I understood and implemented all the requested changes correctly this time.

alan-agius4:
@oocx, ci is red.

oocx:
@alan-agius4 Thanks. I'm sorry, I don't know how I could miss that. I'll fix that next weekend.

oocx:
"ci/angular: size  — cli/new-production/test-project/main.js increased by 24.00KB." 

How can I troubleshoot that? Is there a way to get details for that check? (for exmple, get a diff of the increased output vs. thee baseline)? I don't see a size increase with my local build (for "ng build --prod", but I don't even know which build settings this check tests).

clydin:
For the size check, try rebasing against the latest upstream master.

oocx:
@clydin thanks, rebasing helped.

packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
```diff
@@ -0,0 +1,59 @@
...
+    const overrides = { aot: true };
+
+    runTargetSpec(host, browserTargetSpec, overrides).pipe(
+      tap((buildEvent) => expect(buildEvent.success).toBe(true)),
+      tap(() => {
+        const content = virtualFs.fileBufferToString(
+          host.scopedSync().read(join(outputPath, 'main.js')),
+        );
+
+        expect(content).toContain('":svg:svg"');
```
filipesilva: Where does `":svg:svg"` come from btw? I don't see it in the svg content.

oocx: That's part of the code that is generated by the Angular for the .svg template.

For example, when I have the following .svg file
```
<svg xmlns="http://www.w3.org/2000/svg">
  <text x="20" y="20" font-size="20" fill="red">Hello World</text>
</svg>
```
then the Angular template compiler generates this code:
```
function View_SvgComponentComponent_0(_l) { return _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵvid"](0, [(_l()(), _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵeld"](0, 0, null, null, 2, ":svg:svg", [["xmlns", "http://www.w3.org/2000/svg"]], null, null, null, null, null)), (_l()(), _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵeld"](1, 0, null, null, 1, ":svg:text", [["fill", "red"], ["font-size", "20"], ["x", "20"], ["y", "20"]], null, null, null, null, null)), (_l()(), _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵted"](-1, null, ["Hello World"]))], null, null); }
```
I don't know exactly how the template compiler and view engine work. For my test, I just looked at the generated code when I manually tested svg support, and thought that checking for "svg:svg" would be a good way to test if Angular actually used the .svg file as template.

Without the changes from this PR, the same .svg used as template would generate the following code:
```
function View_SvgComponentComponent_0(_l) { return _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵvid"](0, [(_l()(), _angular_core__WEBPACK_IMPORTED_MODULE_1__["ɵted"](-1, null, ["svg-component.component.svg"]))], null, null); }
```
It just outputs the file name as string.

I hope that answers your question :)

filipesilva:
@oocx this is now merged and should be part of 8.0. Thanks for all the work you've done on it!

packages/angular_devkit/build_angular/test/browser/svg_spec_large.ts
filipesilva: It does answer my question, yes. I think we'll need to alter this test for when we move to the Ivy compiler. But as of now it doesn't need to change.

oocx:
Thanks everyone for your feedback! You have been very helpful and provided valuable feedback, so this was a great experience for me as new contributor.

alan-agius4:
@oocx, thanks for sharing your experience. Looking forward to have more PRs from you in the near future :)

Thanks again for your contribution.

Issue #8595, opened 2017-11-22
Port/address binding issue using ng-serve


rmedaer:
### Versions
```
Angular CLI: 1.5.0
Node: 9.2.0
OS: linux x64
Angular: 5.0.0
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cdk: 5.0.0-rc0
@angular/cli: 1.5.0
@angular/material: 5.0.0-rc0
@angular-devkit/build-optimizer: 0.0.32
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.35
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.0
@schematics/angular: 0.1.1
typescript: 2.4.2
webpack: 3.8.1
```

### Repro steps

> It might be useful to read the last paragraph to understand the issue first.

* Bind on port 80 + address 127.0.0.3 with 3rd tool. Example:
```
authbind nc -l -p 80 -s 127.0.0.3
```
* Make sure that you're binding using `netstat`:
```
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 127.0.0.3:80            0.0.0.0:*               LISTEN      26166/nc            
```
* Start Angular CLI as following:
```
authbind ng serve --port 80 --host 127.0.0.2
```

### Observed behavior
Angular CLI answer that the port is already in use ...

```
Port 80 is already in use. Use '--port' to specify a different port.
```

This answer is coming from `portfinder#getPort`. 
I listen to you:
 
> It's an issue in `portfinder` ! 

Well, I don't think so. Even if the documentation of portfinder is not explicit, by definition, portfinder try to find a port .. So, as I understand, we should change the testing procedure in Angular CLI  (`utilities/check-port.js`)  to only test the given port with given address otherwise default value `0.0.0.0`.

Furthermore, portfinder is testing all the ports and binding addresses ! WTF ?! I might be (for sure) a big improvement to test only what we have to use.
 

### Desired behavior
Do not test all the ports and binding addresses ! 

### Mention any other details that might be useful (optional)
Let me introduce the way I work to understand the issue. When I develop an Angular app, I have so many times constraints linked to CORS. I try to be as close as possible then a production environment using domain name resolution and address binding on localhost. In my setup I have a backend API and its front-end using Angular CLI (only for development purpose). Here is my development network setup:

[...]

PR #13292, opened 2018-12-24
refactor(@angular-devkit/build-angular): improve check port

tiaguinho:
@hansl I have checked the usage of ```checkPort``` function, and the param ```basePort``` was never passed.
I also read the code of portfinder and saw that, when we set the port option, basePort will not be used.
Instead of that, I think set the highestPort to the same as the port we are looking for can improve the time to check if the port is opened.

What do you think?

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -29,7 +29,6 @@
     "minimatch": "3.0.4",
     "parse5": "4.0.0",
     "opn": "5.4.0",
-    "portfinder": "1.0.20",
     "postcss": "7.0.7",
     "postcss-import": "12.0.1",
     "postcss-loader": "3.0.0",
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
@@ -5,26 +5,36 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import * as net from 'net';
 
-import { Observable } from 'rxjs';
-const portfinder = require('portfinder');
-
-
-export function checkPort(port: number, host: string, basePort = 49152): Observable<number> {
-  return new Observable(obs => {
-    portfinder.basePort = basePort;
-    // tslint:disable:no-any
-    portfinder.getPort({ port, host }, (err: any, foundPort: number) => {
-      if (err) {
-        obs.error(err);
-      } else if (port !== foundPort && port !== 0) {
-        // If the port isn't available and we weren't looking for any port, throw error.
-        obs.error(`Port ${port} is already in use. Use '--port' to specify a different port.`);
-      } else {
-        // Otherwise, our found port is good.
-        obs.next(foundPort);
-        obs.complete();
+export function checkPort(port: number, host: string, basePort = 49152): Promise<number> {
+  return new Promise<number>((resolve, reject) => {
+    function _getPort(portNumber: number) {
+      if (portNumber > 65535) {
+        reject(new Error(`There is no port available.`));
       }
-    });
+
+      const server = net.createServer();
+
+      server.once('error', (err: Error & {code: string}) => {
+        if (err.code !== 'EADDRINUSE') {
+          reject(err);
+        } else if (port === 0) {
+          _getPort(portNumber + 1);
+        } else {
+          // If the port isn't available and we weren't looking for any port, throw error.
+          reject(
+            new Error(`Port ${port} is already in use. Use '--port' to specify a different port.`),
+          );
+        }
+      })
+      .once('listening', () => {
+        server.close();
+        resolve(portNumber);
+      })
+      .listen(portNumber, host);
+    }
+
+    _getPort(port || basePort);
   });
 }
--- a/packages/angular_devkit/build_angular/src/dev-server/index.ts
+++ b/packages/angular_devkit/build_angular/src/dev-server/index.ts
@@ -16,7 +16,7 @@ import { WebpackDevServerBuilder } from '@angular-devkit/build-webpack';
 import { Path, getSystemPath, resolve, tags, virtualFs } from '@angular-devkit/core';
 import { Stats, existsSync, readFileSync } from 'fs';
 import * as path from 'path';
-import { Observable, throwError } from 'rxjs';
+import { Observable, from, throwError } from 'rxjs';
 import { concatMap, map, tap } from 'rxjs/operators';
 import * as url from 'url';
 import * as webpack from 'webpack';
@@ -67,7 +67,7 @@ export class DevServerBuilder implements Builder<DevServerBuilderOptions> {
     let first = true;
     let opnAddress: string;
 
-    return checkPort(options.port, options.host).pipe(
+    return from(checkPort(options.port, options.host)).pipe(
       tap((port) => options.port = port),
       concatMap(() => this._getBrowserOptions(options)),
       tap(opts => browserOptions = normalizeBuilderSchema(
--- a/yarn.lock
+++ b/yarn.lock
@@ -7464,15 +7464,6 @@ popper.js@^1.14.1:
   resolved "https://registry.yarnpkg.com/popper.js/-/popper.js-1.14.4.tgz#8eec1d8ff02a5a3a152dd43414a15c7b79fd69b6"
   integrity sha1-juwdj/AqWjoVLdQ0FKFce3n9abY=
 
-portfinder@1.0.20:
-  version "1.0.20"
-  resolved "https://registry.yarnpkg.com/portfinder/-/portfinder-1.0.20.tgz#bea68632e54b2e13ab7b0c4775e9b41bf270e44a"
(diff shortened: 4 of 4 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
```diff
@@ -10,9 +10,9 @@ import { Observable } from 'rxjs';
 const portfinder = require('portfinder');
 
 
-export function checkPort(port: number, host: string, basePort = 49152): Observable<number> {
+export function checkPort(port: number, host: string): Observable<number> {
   return new Observable(obs => {
-    portfinder.basePort = basePort;
+    portfinder.highestPort = port;
```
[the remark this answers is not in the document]
tiaguinho: I think will not change the behavior, because basePort is never used when we set a ```port``` as option (https://github.com/indexzero/node-portfinder/blob/master/lib/portfinder.js#L108).

About the ```stopPort```, I just set the ```highestPort``` to maintain the code style. But, if you think will be more understandable set the ```stopPort``` instead, I can change the code.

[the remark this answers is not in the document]
alan-agius4: Indeed when the basePort will not be used and was not used. However, now introducing `stopPort` might have an effect because both start and stop ports will be 0.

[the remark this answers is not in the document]
tiaguinho: I have read again the code and you are right. Setting the ```stopPort``` the same as ```port``` will throw an error saying that there is no port available, which is not true.
Maybe set the ```stopPort``` as 49152, which was the previous value of ```basePort``` can be a better idea.

What do you think?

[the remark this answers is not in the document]
alan-agius4: Seeing the code, I don't see any benefits in using `stopPort/highestPort`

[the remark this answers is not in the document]
tiaguinho: Agree.

alan-agius4:
good catch @clydin.

alan-agius4:
Port 0 is a wildcard port, which tells the system to find a suitable available port.

This can be useful in CI’s

clydin:
The package really isn’t needed at all.  If you are interested in a refactor, you could completely remove its usage.

If the supplied port is not zero, try to bind to the port and host.  If it fails, throw the error. If it works, pass on the port.

alan-agius4:
@tiaguinho, yeah this should solve #8595

packages/angular_devkit/build_angular/package.json
```diff
@@ -29,7 +29,6 @@
     "minimatch": "3.0.4",
     "parse5": "4.0.0",
     "opn": "5.4.0",
```
alan-agius4: You should also update the yarn lock file

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
```diff
@@ -7,14 +7,27 @@
  */
 
 import { Observable } from 'rxjs';
-const portfinder = require('portfinder');
+const net = require('net');
```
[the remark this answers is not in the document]
alan-agius4: Still pending.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
```diff
@@ -5,26 +5,34 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+const net = require('net');
 
-import { Observable } from 'rxjs';
-const portfinder = require('portfinder');
+export function checkPort(port: number, host: string, basePort = 49152): Promise<number> {
+  return new Promise<number>((resolve, reject) => {
```
alan-agius4: No need to create a promise, use async/await

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
```diff
@@ -5,26 +5,34 @@
...
-        // Otherwise, our found port is good.
-        obs.next(foundPort);
-        obs.complete();
-      }
-    });
+      server.once('error', err => {
+        if (port !== 0 && err.code === 'EADDRINUSE') {
+          // If the port isn't available and we weren't looking for any port, throw error.
+          reject(`Port ${port} is already in use. Use '--port' to specify a different port.`);
+        } else if (port === 0) {
```
alan-agius4: this is incorrect, the error should be displayed when port is not `0`

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
```diff
@@ -5,26 +5,34 @@
...
-      } else {
-        // Otherwise, our found port is good.
-        obs.next(foundPort);
-        obs.complete();
-      }
-    });
+      server.once('error', err => {
+        if (port !== 0 && err.code === 'EADDRINUSE') {
+          // If the port isn't available and we weren't looking for any port, throw error.
+          reject(`Port ${port} is already in use. Use '--port' to specify a different port.`);
```
alan-agius4: Shouldn't this be `new Error`?

tiaguinho: Exceeds the maximum line length. Any idea how can I do this?

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
```diff
@@ -5,26 +5,32 @@
...
-      } else {
-        // Otherwise, our found port is good.
-        obs.next(foundPort);
-        obs.complete();
-      }
-    });
+      server.once('error', err => {
+        if (port !== 0 && err.code === 'EADDRINUSE') {
+          // If the port isn't available and we weren't looking for any port, throw error.
+          reject(new Error(`Port ${port} is already in use. Use '--port' to specify a different port.`));
```
hansl: You don't want to fallback to calling `_getPort()` here. You should return (or wrap the `_getPort()` call in an `else` clause.

clydin: What about something similar to the following? It only checks the error code once and is more explicit that port === 0 has special handling.
```  if (err.code !== 'EADDRINUSE') {
    reject(err);
  } else if (port === 0) {
    _getPort(portNumber + 1);
  } else {
    reject(new Error(`Port ${port} is already in use. Use '--port' to specify a different port.`));
  }
```

The zero handling logic can also be removed in an upcoming major release.  The OS knows how to handle the value.  The check can be completely skipped in that situation as well.

tiaguinho: Beautiful!!
Squash the commits too?

clydin: Yes please.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
```diff
@@ -5,26 +5,36 @@
...
-      } else if (port !== foundPort && port !== 0) {
-        // If the port isn't available and we weren't looking for any port, throw error.
-        obs.error(`Port ${port} is already in use. Use '--port' to specify a different port.`);
-      } else {
-        // Otherwise, our found port is good.
-        obs.next(foundPort);
-        obs.complete();
-      }
-    });
+      server.once('error', err => {
```
clydin: Change this to `server.once('error', (err: Error & { code: string }) => {`

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
alan-agius4:
```
reject(
    new Error`Port ${port} is already in use. Use '--port' to specify a different port.`
);
```

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/check-port.ts
```diff
@@ -5,26 +5,36 @@
...
+        if (err.code !== 'EADDRINUSE') {
+          reject(err);
+        } else if (port === 0) {
+          _getPort(portNumber + 1);
+        } else {
+          // If the port isn't available and we weren't looking for any port, throw error.
+          reject(new Error(`
```
alan-agius4: this will print addition break lines in the Error and potentially in the console. Use `tags.stripIndent`

Ex
```
reject(
    new Error(`Port ${port} is already in use. Use '--port' to specify a different port.`)
);
```

or

tags.stripIndent
```
reject(new Error(tags.stripIndent`
     Port ${port} is already in use. Use '--port' to specify a different port.
`));
```

tiaguinho: This fails on lint check.

alan-agius4: Add comma add the end.

tiaguinho: Done!

PR #13306, opened 2018-12-27
fix shadow variable name, use `find` method instead of `filter`

thinhbuzz:


thinhbuzz:
I signed it!

Vào 11:19, 27 Th12, 2018, lúc 11:19, 11:19 đã viết:

PR #13309, opened 2018-12-27
revert: fix(@ngtools/webpack): emit lazy routes errors on rebuilds

### LINKED ISSUE #13102: Slower compilation when using @ngtools/webpack@7.1.0
Initially reported in https://github.com/angular/angular/issues/26674#issuecomment-443263655 by @sod

---

hm indeed, just updating ngtools gives me the slowdown. Also using angular 7.1 with old ngtools 7.0.7 is significantly faster.

versions | 1st run | 2nd run | 3rd run | 4th run | 5th run
-|--|--|--|--|-
`@angular 7.0.4, @ngtools 7.0.7` | 31s | 15s | 7s | 5s | 5s
`@angular 7.0.4, @ngtools 7.1.0` | 29s | 14.5s | 15.5s | 12s | 12s
`@angular 7.1.1, @ngtools 7.1.0` | 29s | 13.5s | 13.5s | 15s | 16s
`@angular 7.1.1, @ngtools 7.0.7` | 28s | 12.8s | 7s | 5s | 4.2s

alan-agius4:
This reverts commit edb84b340ff996df2ca6a2aaf765304cc64fef3e

The team has decided to bring back the faster but potentially less accurate method of detecting lazy routes upon JIT rebuilds (first builds will always use the more complete Angular compiler method). Applications that do not have lazy routes within libraries and that only use direct string literals with loadChildren should not be affected by the potential of less accurate detection. Note that the function overload of loadChildren also does not apply to this situation.

For those projects where correctness of lazy route detection outweighs rebuild speed, please consider using AOT mode for development. AOT mode will also provide a full set of template errors as well which JIT mode is not capable of doing.

Fixes #13102

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/lazy-module_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/lazy-module_spec_large.ts
@@ -6,9 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { DefaultTimeout, TestLogger, runTargetSpec } from '@angular-devkit/architect/testing';
+import { DefaultTimeout, runTargetSpec } from '@angular-devkit/architect/testing';
 import { join, normalize } from '@angular-devkit/core';
-import { take, tap } from 'rxjs/operators';
+import { tap } from 'rxjs/operators';
 import { BrowserBuilderSchema } from '../../src';
 import { browserTargetSpec, host } from '../utils';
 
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/test/browser/lazy-module_spec_large.ts
```diff
@@ -90,50 +89,6 @@ describe('Browser Builder lazy modules', () => {
     ).toPromise().then(done, done.fail);
   });
 
-  it('should show error when lazy route is invalid on watch mode AOT', (done) => {
```
filipesilva: Can we keep this test?

alan-agius4: Re-added.

Issue #8981, opened 2017-12-23
file hash different when build with and without sourcemap


mix5003:
### Versions
```
Angular CLI: 1.6.1
Node: 8.9.3
OS: linux x64
Angular: 5.1.1
... animations, common, compiler, compiler-cli, core, forms
... http, platform-browser, platform-browser-dynamic
... platform-server, router

@angular/cli: 1.6.1
@angular-devkit/build-optimizer: 0.0.36
@angular-devkit/core: 0.0.22
@angular-devkit/schematics: 0.0.42
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.9.1
@schematics/angular: 0.1.11
@schematics/schematics: 0.0.11
typescript: 2.4.2
webpack: 3.10.0
```

### Repro steps
1. `ng new test-ng`
2. `cd test-ng`
3. `npm run build -- -sm` 
```chunk {0} polyfills.f6719bbdf7981a979ac6.bundle.js, polyfills.f6719bbdf7981a979ac6.bundle.js.map (polyfills) 60.9 kB [initial] [rendered]
chunk {1} main.19a2774fa8a655b05e6f.bundle.js, main.19a2774fa8a655b05e6f.bundle.js.map (main) 152 kB [initial] [rendered]
chunk {2} styles.d41d8cd98f00b204e980.bundle.css, styles.d41d8cd98f00b204e980.bundle.css.map (styles) 66 bytes [initial] [rendered]
chunk {3} inline.d92eba321460843f805b.bundle.js, inline.d92eba321460843f805b.bundle.js.map (inline) 1.51 kB [entry] [rendered]
```
4. `npm run build`
```chunk {0} polyfills.f039bbc7aaddeebcb9aa.bundle.js (polyfills) 60.9 kB [initial] [rendered]
chunk {1} main.7eb03a969263f31e13db.bundle.js (main) 152 kB [initial] [rendered]
chunk {2} styles.d41d8cd98f00b204e980.bundle.css (styles) 0 bytes [initial] [rendered]
chunk {3} inline.1bcc593280127905b393.bundle.js (inline) 1.45 kB [entry] [rendered]
```
as you can see when it build with and without sourcemap. file that it build has different hash.
i use diff tool and it said just 1 line add when enable sourcemap it something like
```//# sourceMappingURL=main.b0acc5628cb9dda9f019.bundle.js.map```

after i try many time i found `--build-optimizer` is the cause to make this happen
[...]

PR #13316, opened 2018-12-27
fix(@angular-devkit/build-angular): keep hash equal

Jefiozie:
Keep hashes equal with/without sourcemaps. 

Closes: #8981

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/source-map_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/source-map_spec_large.ts
@@ -7,8 +7,9 @@
  */
 
 import { runTargetSpec } from '@angular-devkit/architect/testing';
-import { join, normalize, virtualFs } from '@angular-devkit/core';
-import { tap } from 'rxjs/operators';
+import { PathFragment, join, normalize, virtualFs } from '@angular-devkit/core';
+import { forkJoin } from 'rxjs';
+import { mergeMap, switchMap, tap } from 'rxjs/operators';
 import { BrowserBuilderSchema, OutputHashing } from '../../src';
 import { browserTargetSpec, host } from '../utils';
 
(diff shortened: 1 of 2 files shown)


Jefiozie:
Will fix the failing test later upcoming days.

packages/angular_devkit/build_angular/test/browser/source-map_spec_large.ts
```diff
@@ -220,4 +221,44 @@ describe('Browser Builder source map', () => {
...
+    };
+    let mainFileOutput: PathFragment | undefined;
+    let mainFileWithSourceMapOutput: PathFragment | undefined;
+
+    const runWithoutSourceMaps = runTargetSpec(host, browserTargetSpec, overridesWithoutSourcemaps)
+    .pipe(
+      tap((buildEvent) => expect(buildEvent.success).toBe(true)),
+      tap(() => {
+        mainFileWithSourceMapOutput = host.fileMatchExists(outputPath, /main\.[0-9a-f]{20}\.js/);
+      })).toPromise();
```
alan-agius4: No need to convert this to promise.

packages/angular_devkit/build_angular/test/browser/source-map_spec_large.ts
```diff
@@ -220,4 +221,44 @@ describe('Browser Builder source map', () => {
...
+
+
+    const runWithSourceMaps = runTargetSpec(host, browserTargetSpec, overridesWithSourceMaps).pipe(
+      tap((buildEvent) => expect(buildEvent.success).toBe(true)),
+      tap(() => {
+        mainFileOutput = host.fileMatchExists(outputPath, /main\.[0-9a-f]{20}\.js/);
+      }),
+    ).toPromise();
```
alan-agius4: No need to convert this to promise.

packages/angular_devkit/build_angular/test/browser/source-map_spec_large.ts
```diff
@@ -220,4 +221,44 @@ describe('Browser Builder source map', () => {
...
+
+    forkJoin(runWithoutSourceMaps, runWithSourceMaps).toPromise()
+    .then(() => {
+      expect(mainFileWithSourceMapOutput).toEqual(mainFileOutput);
```
alan-agius4: This test, is actually not checking that the file names are the same, but rather that both filenames contains hashes.
You should extract the hash and match it.

clydin: Note that this is a bad test.  The file names should not be the same.  The content is different when changing the source map option which means the hash in file name should also be different.  If they were the same then this would have negative consequences for long term caching; especially when SRI is enabled (in which case the site will fail to load).

alan-agius4: Yes they should be different, not sure why it was marked as a bug.

clydin: Currently it appears that they may be different for the wrong reason.  Essentially one bug partially cancelling out another.  Some further investigation may be prudent.

mgechev:
@Jefiozie any updates?

Jefiozie:
Sorry I forgot about this.. I will have a look upcoming week. Thanks for the ping.

Jefiozie:
I'm closing this for now as I believe we need some disucssion on the issue.

PR #13325, opened 2018-12-28
feat(@schematics/angular): JSON schema for service worker config

### LINKED ISSUE #13324: ngsw-config.json has no schema, so no validation and no assistance
### Feature Request
```
- [x] feature request
```

### Commands
```
- [x] generate service-worker
- [x] add @angular/pwa
```

### Versions
All CLI versions.

### Repro steps
```ng g service-worker```
or
```ng add @angular/pwa```

There is no JSON schema for the `ngsw-config.json`, so no validation and no assistance.

This is especially inconvenient as a wrong config will often build with no errors, but then the service worker won't work.

### Desired functionality
Validation and assistance for `ngsw-config.json`.

PR done in #13325

### PR DESCRIPTION
**Pull request by cyrilletuzi**
Fixes #13324 

Depends on angular/angular#27859

### DIFF
--- a/packages/schematics/angular/service-worker/index.ts
+++ b/packages/schematics/angular/service-worker/index.ts
@@ -169,13 +169,16 @@ export default function (options: ServiceWorkerOptions): Rule {
       throw new SchematicsException(`Service worker requires a project type of "application".`);
     }
 
+    const relativePathToWorkspaceRoot = project.root ?
+      project.root.split('/').filter(x => x !== '').map(x => '..').join('/') : '.';
+
     let { resourcesOutputPath = '' } = getProjectConfiguration(workspace, options);
     if (resourcesOutputPath) {
       resourcesOutputPath = '/' + resourcesOutputPath.split('/').filter(x => !!x).join('/');
     }
 
     const templateSource = apply(url('./files'), [
-      applyTemplates({ ...options, resourcesOutputPath }),
+      applyTemplates({ ...options, resourcesOutputPath, relativePathToWorkspaceRoot }),
       move(project.root),
     ]);
 
--- a/packages/schematics/angular/service-worker/index_spec.ts
+++ b/packages/schematics/angular/service-worker/index_spec.ts
@@ -101,6 +101,31 @@ describe('Service Worker Schematic', () => {
       .toBe('projects/bar/ngsw-config.json');
   });
 
+  it('should add $schema in ngsw-config.json with correct relative path', () => {
+    const pathToNgswConfigSchema = 'node_modules/@angular/service-worker/config/schema.json';
+
+    const name = 'foo';
+    const rootAppOptions: ApplicationOptions = {
+      ...appOptions,
+      name,
+      projectRoot: '',
+    };
+    const rootSWOptions: ServiceWorkerOptions = {
+      ...defaultOptions,
+      project: name,
+    };
+    const rootAppTree = schematicRunner.runSchematic('application', rootAppOptions, appTree);
+    const treeInRoot = schematicRunner.runSchematic('service-worker', rootSWOptions, rootAppTree);
+    const pkgTextInRoot = treeInRoot.readContent('/ngsw-config.json');
+    const configInRoot = JSON.parse(pkgTextInRoot);
+    expect(configInRoot.$schema).toBe(`./${pathToNgswConfigSchema}`);
+
+    const treeNotInRoot = schematicRunner.runSchematic('service-worker', defaultOptions, appTree);
+    const pkgTextNotInRoot = treeNotInRoot.readContent('/projects/bar/ngsw-config.json');
+    const configNotInRoot = JSON.parse(pkgTextNotInRoot);
+    expect(configNotInRoot.$schema).toBe(`../../${pathToNgswConfigSchema}`);
+  });
+
   it('should add root assets RegExp', () => {
     const tree = schematicRunner.runSchematic('service-worker', defaultOptions, appTree);
     const pkgText = tree.readContent('/projects/bar/ngsw-config.json');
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: packages/schematics/angular/service-worker/files/ngsw-config.json [hunk 1]
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "./node_modules/@schematics/angular/service-worker/config/schema.json",
```
alan-agius4: When having multiple apps inside the workspace the `ngsw-config.json` is not generated in the root of the workspace, but in a sub directory example: `/projects/myapp/src/ngsw-config.json`, this will cause the schema not to be found.

You'd need to do something like: https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/application/files/root/tsconfig.app.json#L2

cyrilletuzi: Is the relativePathToWorkspaceRoot variable available in this context too?

packages/schematics/angular/service-worker/config/schema.json
```diff
@@ -0,0 +1,134 @@
+{
```
alan-agius4: I am not sure if the schema should be here in the first place. I think the right place for this would be in `@angular/service-worker`. As really the schema can be used by anyone who uses that package, and doesn't necessary have a dependency on `@schematics/angular`. Also, it will ease maintenance, as it's harder to keep models in sync within different repos. 

@gkalpak, what do you think?

cyrilletuzi: I was also not sure of the place of the schema. Just waiting a green light, and I'll do the PR in Angular repo.

packages/schematics/angular/service-worker/files/ngsw-config.json
alan-agius4: No, though you can add it. See: https://github.com/angular/angular-cli/blob/2326850374cd393a825e339beb5846afcff87f8a/packages/schematics/angular/application/index.ts#L284-L295

cyrilletuzi: OK. I won't be confident to do this part without testing, but as it happens at generation, I don't know how to test. How can I create an Angular project from the Angular CLI from my fork repo?

alan-agius4: https://github.com/angular/angular-cli/blob/master/packages/angular/cli/README.md#development-hints-for-working-on-angular-cli

Though in your case you'd need to link the `@schematics/angular` package.

Or even better create a Unit Test.

packages/schematics/angular/service-worker/config/schema.json
cyrilletuzi: Done in angular/angular#27859

packages/schematics/angular/service-worker/files/ngsw-config.json
cyrilletuzi: I'll manage this part when angular/angular#27859 is merged.

cyrilletuzi:
angular/angular#27859 is ready and will be merged soon, so I've updated this PR (to manage the relative path + a test), so it can make it to 7.2.0 (as it would be very helpful during the Angular courses I teach, students were lost without any assistance on this).

packages/schematics/angular/service-worker/index.ts
```diff
@@ -171,13 +171,17 @@ export default function (options: ServiceWorkerOptions): Rule {
       throw new SchematicsException(`Service worker requires a project type of "application".`);
     }
 
+    const projectRoot = ('root' in project) ? project.root : '';
```
alan-agius4: This can be simplified to 
```
    const relativePathToWorkspaceRoot = project.root
      ? project.root.split('/').filter(x => x !== '').map(x => '..').join('/')
      : '';
```

cyrilletuzi:
Change done.

mgechev:
@cyrilletuzi would you rebase? We should be able to merge.

packages/schematics/angular/service-worker/index_spec.ts
```diff
@@ -98,6 +98,31 @@ describe('Service Worker Schematic', () => {
...
+      name,
+      projectRoot: '',
+    };
+    const rootSWOptions: ServiceWorkerOptions = {
+      ...defaultOptions,
+      project: name,
+    };
+    const rootAppTree = schematicRunner.runSchematic('application', rootAppOptions, appTree);
+    const treeInRoot = schematicRunner.runSchematic('service-worker', rootSWOptions, rootAppTree);
+    const pkgTextInRoot = treeInRoot.readContent('/ngsw-config.json');
```
gkalpak: `pkgText...` --> `configText...`

gkalpak:
> @cyrilletuzi would you rebase? We should be able to merge.

@mgechev, this is blocked on angular/angukar#27859.

packages/schematics/angular/service-worker/index_spec.ts
cyrilletuzi: I've kept the same names as in the other tests.

cyrilletuzi:
I'll rebase when it's time.

packages/schematics/angular/service-worker/index_spec.ts
gkalpak: It doesn't change the fact that it is a bad, misleading name 😛
I would change it in other tests too 😇

cyrilletuzi:
Closing, as the PR in Angular repo was not merged.

For people interested, it is available as an external contribution in [`@ngx-pwa/ngsw-schema` lib](https://github.com/cyrilletuzi/ngsw-schema).

alan-agius4:
Hi @cyrilletuzi, would you mind rebasing this since the angular/angular PR is merged?

alan-agius4:
@gkalpak, that can be done via an E2E test. 

But that's no as trivial as a unit test to do. I think for the time being, it would best to leave it out.

cyrilletuzi:
@alan-agius4 Rebased.

PR #13348, opened 2019-01-02
[WIP] Use Webpack 5 and add cache option

filipesilva:
Extremely experimental. Just opening a PR to run the Webpack 5 alphas through the test suite and see what breaks.

We will only update to Webpack 5 once it is stable and most of the plugins we use also support it.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -299,6 +299,16 @@
       "type": "boolean",
       "description": "Output profile events for Chrome profiler.",
       "default": false
+    },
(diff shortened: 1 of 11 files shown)


packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -299,6 +299,16 @@
       "type": "boolean",
       "description": "Output profile events for Chrome profiler.",
       "default": false
+    },
```
alan-agius4: Note: let’s remember to add these in the CLI schema

alan-agius4:
Closing since this was a POC

PR #13349, opened 2019-01-02
ci: add windows buildkite

filipesilva:
Related to https://github.com/angular/angular/pull/27205.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,7 @@
     "validate-commits": "./bin/devkit-admin validate-commits",
     "prepush": "node ./bin/devkit-admin hooks/pre-push",
     "preinstall": "node ./tools/yarn/check-yarn.js",
-    "webdriver-update-appveyor": "webdriver-manager update --standalone false --gecko false --versions.chrome 2.37",
-    "webdriver-update-circleci": "webdriver-manager update --standalone false --gecko false --versions.chrome $CHROMEDRIVER_VERSION_ARG "
+    "webdriver-update": "webdriver-manager update --standalone false --gecko false --versions.chrome 2.45"
   },
   "repository": {
     "type": "git",
@@ -66,6 +65,7 @@
   "dependencies": {
     "glob": "^7.0.3",
     "node-fetch": "^2.2.0",
+    "puppeteer": "1.11.0",
     "quicktype-core": "^6.0.15",
     "temp": "^0.9.0",
     "tslint": "^5.11.0",
--- a/tests/legacy-cli/e2e/utils/project.ts
+++ b/tests/legacy-cli/e2e/utils/project.ts
@@ -1,4 +1,5 @@
-import { readFile, writeFile, replaceInFile } from './fs';
+import * as fs from 'fs-extra';
+import { readFile, writeFile, replaceInFile, prependToFile } from './fs';
 import { execAndWaitForOutputToMatch, npm, silentNpm, ng } from './process';
 import { getGlobalVariable } from './env';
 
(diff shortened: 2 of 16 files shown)


tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -186,58 +187,62 @@ export function useCIDefaults(projectName = 'test-project') {
...
+      json['devDependencies']['karma-chrome-launcher'] = '~2.2.0'; // Minimum for ChromeHeadless.
+    }))
+    // Use Pupeteer in protractor if a config is found on the project.
+    .then(() =>  {
+      if (fs.existsSync(protractorConf)) {
+        return replaceInFile(protractorConf,
+          `'browserName': 'chrome'`,
+          `'browserName': 'chrome',
+          chromeOptions: {
+            args: ['--headless', '--no-sandbox', '--disable-gpu'],
```
clydin: is `--disable-gpu` still needed? This was originally due to a bug with windows support.

package.json
```diff
@@ -35,8 +35,7 @@
     "validate-commits": "./bin/devkit-admin validate-commits",
     "prepush": "node ./bin/devkit-admin hooks/pre-push",
     "preinstall": "node ./tools/yarn/check-yarn.js",
-    "webdriver-update-appveyor": "webdriver-manager update --standalone false --gecko false --versions.chrome 2.37",
-    "webdriver-update-circleci": "webdriver-manager update --standalone false --gecko false --versions.chrome $CHROMEDRIVER_VERSION_ARG "
+    "webdriver-update": "webdriver-manager update --standalone false --gecko false --versions.chrome 2.45"
```
clydin: Hard coding the version in multiple places should be avoided if possible.  Would it be possible to centralize the chrome and driver version information?

filipesilva: Hmm no ergonomic solution comes to mind. Storing it on a env var would need the var to be set, and a cross platform solution.

Storing it on a node module would require a roundabout way of calling `webdriver-update` from the node bin.

Right now it's in 2 places, which is better than before, but still would be better to be in a single place. Although to be fair it's on two places because each of them happen to install a certain version of puppeteer, and these two things happen to match (e.g. one can imagine one of these places using a different puppeteer/webdriver combo).

Did you have an approach in mind?

tests/legacy-cli/e2e/utils/project.ts
filipesilva: Might not be, no. Will remove and see if it's ok.

filipesilva:
@clydin performance wise, caching and sharding would both help right now. Caching would speed up the initial setup. Sharding would make each individual shard take less time to complete.

I have explored caching in https://github.com/angular/angular/pull/27990, but haven't explored sharding. I expect we'd do something similar as today in CircleCI.

Long term further Bazel integration should make sharding moot though. Under an ideal Bazel setup the remote cache/execution would manage cache. Initial cache for node_modules and Bazel artifacts would still be of help.

clydin:
Something to note though is that circleCI (on Linux) takes a total of ~48 minutes to run the full suite of 187 tests and that is with a full rebuild and setup of the E2E infrastructure four times.  Whereas this and appveyor are only running 87 tests.
Is this just a windows file system access performance issue (or other windows specific concern)?

filipesilva:
I'm not aware of the specific cause of windows taking longer overall for our e2e suite. It's worth looking into further. FS is much slower on windows, yes, but that can't be the full story.

filipesilva:
We'll also need to ensure there's a hard timeout in the future. In https://buildkite.com/angular/cli/builds/26#d9a71d0c-0863-4bb5-8ab2-ffb07d89a1a6 `npm install` was seen taken almost two days before I manually cancelled it.

kyliau:
>  Filipe on Slack: The snapshot_publish circleci check shouldn't be running, but I think it runs because I made the branch directly on angular-cli instead of on a fork.

Merging this despite `ci/circleci: snapshot_publish` failing.

Issue #10612, opened 2018-05-03
Mixing configurations


ocombe:
### Versions
v6

### Observed behavior
You cannot use `--prod` with a custom `configuration` since `--prod` is an alias for `--configuration=production`. This means that if I make a custom configuration for my project, I need to duplicate it for the prod build to include everything from the `production` configuration.

### Desired behavior
I'd like to be able to mix configurations, where each one overwrites the precedent. For example with `-c=prod,xxx`, where `xxx` overwrites `production` values if they both define the same property

For example if I'm doing i18n, I would like to have something like that:
```
          "configurations": {
            "production": {
              "fileReplacements": [
                {
                  "replace": "src/environments/environment.ts",
                  "with": "src/environments/environment.prod.ts"
                }
              ],
              "optimization": true,
              "outputHashing": "all",
              "sourceMap": false,
              "extractCss": true,
              "namedChunks": false,
              "aot": true,
              "extractLicenses": true,
              "vendorChunk": false,
              "buildOptimizer": true
            },
            "fr": {
              "aot": true,
              "outputPath": "dist/my-project-fr/",
              "i18nFile": "src/locale/messages.fr.xlf",
              "i18nFormat": "xlf",
              "i18nLocale": "fr",
              "i18nMissingTranslation": "error"
            }
          }
        },
```

And use `-c=production,fr`, or `--prod -c=fr`

filipesilva:
@buu700 was also interested in something similar in https://github.com/angular/angular-cli/issues/10559

phl3x0r:
That would be a great addition.
At the moment (in the 1.x.x config) I have a setup with many apps sharing a lot of the same (duplicate) configuration. I was hoping that 6.0 would provide a way to share common configuration between app flavors, but at this point, the config will just be even more bloated than before.
In my case it's not the i18n files (although potentially it will be in the future), but assets, styles and environment files that changes between customers (apps).

MickL:
Alternative to using multiple configurations they could have a extend feature as mentioned here: https://github.com/angular/angular-cli/issues/11206

MickL:
@hansl Personally i would set this issue to priority 2 (required):

serve (builder `@angular-devkit/build-angular:dev-server`) does not allow to set properties like `assets` or `fileReplacements`. So instead of:

`ng build -c=prod,theme1,en` and `ng serve -c=theme1,en`

we have to make TONS configurations

```
"build": {
   ...
    "configurations": {
        "theme1": {
[...]

PR #13370, opened 2019-01-05
Support mixing configurations

tomblachut:
Inspired by #10612 I decided to tackle configuration bloat myself. Noticed that current Architect is being deprecated, but I believe this feature could land before 8.x.

Looking forward to getting feedback, never contributed to OSS before.

### DIFF
--- a/packages/angular_devkit/architect/src/architect.ts
+++ b/packages/angular_devkit/architect/src/architect.ts
@@ -197,24 +197,32 @@ export class Architect {
     const {
       project: projectName,
       target: targetName,
-      configuration: configurationName,
+      configuration: configurationSpecifier,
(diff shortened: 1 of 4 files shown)


hansl:
This is a great feature, but unfortunately we are re-working the architect API, so we won't let this in before the rework is done. The good news is that this feature can also be implemented after the rework and we should consider it for 8.0 (or 8.1). We'll let you know when this is ready. Cheers!

Please refer to https://github.com/angular/angular-cli/pull/13463 for the PR of the new architect API.

mgechev:
@tomblachut closing this PR since #13463 will change the architect API completely. Let's revisit once #13463 lands!

Issue #13024, opened 2018-11-21
[ivy]: experimentalIvy flag generates an app with no change detection


cexbrayat:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Command (mark with an `x`)
```
- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

Angular CLI: 7.1.0-rc.0
Node: 8.11.3
OS: darwin x64
Angular: 7.1.0-rc.0

### Repro steps

Generate a new app with `experimentalIvy`

```
ng new ivy-cd --experimental-ivy --defaults
cd ivy-cd
$(npm bin)/ivy-ngcc
ng serve --aot
```

add a simple `setTimeout` in `app.component.ts` like:

```typescript
constructor() {
  setTimeout(() => {
    console.log('set timeout');
    this.title = 'hello';
  }, 2000);
}
```

will result in a trace in the browser console, but no update in the DOM. It looks like change detection is not run. After discussing with @jasonaden on Slack, he confirms that if the app is bootstrapped using `renderComponent` (as it is currently when using the `experimentalIvy` flag) then CD is not running. But he pointed out that using `platformBrowser` fixes it, and indeed it does.

### Desired functionality

The `experimentalIvy` flag could be simplified to probably just add the Angular compiler option in `tsconfig.app.json` but the rest could be removed and use the default bootstrap mechanism.

### Mention any other details that might be useful

I can submit a PR if the CLI team agrees.

@jasonaden asked me to also cc @alxhub for his input.

dirkluijk:
@cexbrayat Same issue here. How should I use platformBrowser (the workaround you mentioned) exactly?

klimentru1986:
@cexbrayat try to use [markDirty](https://github.com/angular/angular/blob/f8096d499324cf0961f092944bbaedd05364eea1/packages/core/src/render3/instructions.ts#L2557) function. [Example here](https://github.com/klimentru1986/angular-ivy/blob/master/src/app/app.component.ts).

Or you can do it in bowser console
```
// Get component
let comp = ng.getComponent(document.querySelector('app-root'));
// MarkDirty
ng.markDirty(comp);

[...]

PR #13382, opened 2019-01-08
Ivy backwards compatible bootstrap

alan-agius4:
feat(@schematics/angular): `experimentalIvy` should generates backwards compatible solution

Usage:
```
ng new foo --experimentalIvy
```

Users will need to use `ngcc` so to process code coming from NPM and produce the equivalent Ivy version. This needs to be done once after installing the node packages.

```
node_modules/.bin/ivy-ngcc
```

`ngtsc` compilter will then be used when building using the `--aot` or `--prod` flag.

feat(@ngtools/webpack): add support for Ivy compatible solution
Ivy compatible solution still uses the 'standard' bootstrapping code, however the difference is that there are no factories.

Using a similar approach for `ngc` but without factories and instead of `bootstrapModuleFactory` it uses `bootstrapModule`.

More context: https://github.com/angular/angular-cli/issues/13024#issuecomment-449146917

TODO check with Framework: 

- [x] Confirm that using `bootstrapModule` is the right approach
- [x] How to retrieve lazy loaded modules as it seems that `listLazyRoutes` is not returning anything, - https://github.com/angular/angular/pull/27697
- [x] TBD - Should we add `ivy-ngcc` as a post install script
- [x] TBD - Should we turn on factory shimming by default in tsconfig via `allowEmptyCodegenFiles`

Fixes: #13024

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -406,12 +406,12 @@ export class AngularCompilerPlugin {
     }
 
     // If there's still no entryModule try to resolve from mainPath.
-    if (!this._entryModule && this._mainPath && !this._compilerOptions.enableIvy) {
+    if (!this._entryModule && this._mainPath) {
       time('AngularCompilerPlugin._make.resolveEntryModuleFromMain');
       this._entryModule = resolveEntryModuleFromMain(
         this._mainPath, this._compilerHost, this._getTsProgram() as ts.Program);
 
-      if (!this.entryModule) {
+      if (!this.entryModule && !this._compilerOptions.enableIvy) {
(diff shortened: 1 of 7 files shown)


alan-agius4:
Still need https://github.com/angular/angular/pull/27848 to get prod build working correctly.

alan-agius4:
@IgorMinar, 2 questions please:

Should we add `ivy-ngcc` as a post install script?
Should we turn on factory shimming by default in tsconfig via `allowEmptyCodegenFiles`?

IgorMinar:
let's try not to turn on factoryShimming and see if we can get away with that without causing too much disruption. It would be better if we avoided factories and left it as a backup option for people that absolutely need them.

ivy-ngcc as a postinstall script sounds good. let's try that.

PR #13389, opened 2019-01-09
fix(@schematics/update): check if replace is a function

joeson1:
When substituting environment variables in npmrc options, string replace was also applied on boolean values. This caused the ng update process to fail #10624

### DIFF
--- a/packages/schematics/update/update/npm.ts
+++ b/packages/schematics/update/update/npm.ts
@@ -84,7 +84,9 @@ function readOptions(
 
   // Substitute any environment variable references
   for (const key in options) {
-    options[key] = options[key].replace(/\$\{([^\}]+)\}/, (_, name) => process.env[name] || '');
+    if (typeof options[key] === 'string') {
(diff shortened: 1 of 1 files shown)


packages/schematics/update/update/npm.ts
```diff
@@ -84,7 +84,9 @@ function readOptions(
 
   // Substitute any environment variable references
   for (const key in options) {
-    options[key] = options[key].replace(/\$\{([^\}]+)\}/, (_, name) => process.env[name] || '');
+    if (options[key].replace !== undefined && typeof options[key].replace === 'function') {
```
clydin: maybe `options[key] && typeof options[key].replace === 'function')` just incase it is undefined/null for some reason?

clydin: actually we only care about strings here so `typeof options[key] === 'string'` should be sufficient.

joeson1:
@alan-agius4  I squashed the commits

PR #13398, opened 2019-01-10
docs(@angular/cli): allow spaces in project name

L2jLiga:


clydin:
Thank you for the contribution.  However, projects intentionally cannot contain spaces as they are designed to conform to the requirements of an npm package name.  This is also not a documentation change as it has systemic consequences.

PR #13403, opened 2019-01-10
Conditional ES5 Browser Polyfill Loading

clydin:
- no need to manually import and manage individual ES2015 polyfills required by Angular
  - ES2015 polyfills are only loaded by browsers that require them
  - Controllable via a new `es5BrowserSupport` option; no behavior change if option is not enabled
  - `es5BrowserSupport` is enabled in newly generated projects
  - allows for custom polyfill solutions by not using the new option
  - works in conjunction with conditional JIT polyfills and will load ES2015 specific JIT polyfills as needed
  - Saves ~56KB on native ES2015 browsers

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/es2015-polyfills.js
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/es2015-polyfills.js
@@ -1,0 +1,21 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import 'core-js/es6/symbol';
+import 'core-js/es6/object';
+import 'core-js/es6/function';
+import 'core-js/es6/parse-int';
+import 'core-js/es6/parse-float';
+import 'core-js/es6/number';
+import 'core-js/es6/math';
+import 'core-js/es6/string';
+import 'core-js/es6/date';
+import 'core-js/es6/array';
+import 'core-js/es6/regexp';
+import 'core-js/es6/map';
+import 'core-js/es6/weak-map';
+import 'core-js/es6/set';
--- a/packages/schematics/angular/application/files/src/polyfills.ts
+++ b/packages/schematics/angular/application/files/src/polyfills.ts
@@ -18,30 +18,8 @@
  * BROWSER POLYFILLS
  */
 
-/** IE9, IE10, IE11, and Chrome <55 requires all of the following polyfills.
- *  This also includes Android Emulators with older versions of Chrome and Google Search/Googlebot
- */
-
-// import 'core-js/es6/symbol';
-// import 'core-js/es6/object';
-// import 'core-js/es6/function';
-// import 'core-js/es6/parse-int';
-// import 'core-js/es6/parse-float';
-// import 'core-js/es6/number';
-// import 'core-js/es6/math';
-// import 'core-js/es6/string';
-// import 'core-js/es6/date';
-// import 'core-js/es6/array';
-// import 'core-js/es6/regexp';
-// import 'core-js/es6/map';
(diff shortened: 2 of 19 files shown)


tests/legacy-cli/e2e/tests/basic/scripts-array.ts
```diff
@@ -54,6 +54,7 @@ export default function () {
     // index.html lists the right bundles
     .then(() => expectFileToMatch('dist/test-project/index.html', oneLineTrim`
       <script type="text/javascript" src="runtime.js"></script>
+      <script type="text/javascript" src="es2015-polyfills.js" nomodule></script>
```
mathiasbynens: Drive-by comment: can we drop the `type="text/javascript"` please? It’s redundant: https://mathiasbynens.be/notes/html5-levels#type-attributes

clydin: Yes.  A refactor and cleanup of the index generation is planned that will also address this.

packages/schematics/angular/application/files/src/polyfills.ts
```diff
@@ -18,31 +18,9 @@
...
-// import 'core-js/es6/math';
-// import 'core-js/es6/string';
-// import 'core-js/es6/date';
-// import 'core-js/es6/array';
-// import 'core-js/es6/regexp';
-// import 'core-js/es6/map';
-// import 'core-js/es6/weak-map';
-// import 'core-js/es6/set';
-
 /** IE10 and IE11 requires the following for NgClass support on SVG elements */
```
alan-agius4: Any reason why this was not moved as well?

clydin: It requires an additional package install. It is also optional and DOM related rather than ES2015 related.

tomgruszowski:
Are there any docs on how to enable this?

belf2601:
How can we enable on current projects ?

diicar:
It's enabled by defauly, just need to Update cli to v7.3.x I believe.

neilhem:
Add `es5BrowserSupport: true` to angular.json options

maxailloud:
How can I fix the ```Schema validation failed with the following errors:
  Data path "" should NOT have additional properties(es5BrowserSupport).``` error?
I just added the option to my build option.

clydin:
@maxailloud The project most likely has an older version of `@angular-devkit/build-angular` (for this feature 0.13.0+ is needed).

Also, the team is more than happy to help when possible, but please avoid using a PR for these type of discussions as they are not generally monitored.

maxailloud:
Oh my bad, thought it was an issue.
And btw don't know why but i still had the 0.12.4 version of it, even after executing ng update @angular/cli.

Thank you and sorry again.

paulogr:
And on your previously created projects, don't forget to remove those es6 polyfills imports from polyfills.ts file for this feature be real effective!

packages/angular_devkit/build_angular/src/angular-cli-files/models/es2015-polyfills.js
```diff
@@ -0,0 +1,21 @@
...
+import 'core-js/es6/parse-float';
+import 'core-js/es6/number';
+import 'core-js/es6/math';
+import 'core-js/es6/string';
+import 'core-js/es6/date';
+import 'core-js/es6/array';
+import 'core-js/es6/regexp';
+import 'core-js/es6/map';
+import 'core-js/es6/weak-map';
+import 'core-js/es6/set';
```
DmitryEfimenko: I might be mistaken, but it's my understanding that polyfills are moved into an internal file. I often find that I also need to import some of the es7 modules (for example to support `Object.entries`) into polifyls. Having them in an internal file won't allow me adding additional polifylls only required for legacy browsers. It'd be my preference if there was a new file generated by CLI and included in the app: `polifyls.legacy.ts` or something like that.

PR #13410, opened 2019-01-11
refactor: use `.template` suffix for all schematic files

### LINKED ISSUE #13378: Use ".template" suffix for schematic template files
Currently in the `cli-hello-world-ivy` integration test for `angular/angular`:

```
Failed to read entry point info from /home/circleci/ng/integration/cli-hello-world-ivy/node_modules/@schematics/angular/workspace/files/package.json with error SyntaxError: Unexpected token < in JSON at position 1121.
```

It looks like this happens because ngcc just tries to read all `package.json` files and **`@schematics/angular`** ships their templated `package.json` file without using the `.template` suffix.

If there is nothing opposed to switching the `package.json` to use `package.json.template`, I think that it would make sense in order to avoid such warnings/errors in general.

Note that this initially caused errors for NGCC, but has been switched to a warning in https://github.com/angular/angular/pull/26539.

alan-agius4:
Currently when using `ivy-ngcc` it will print out a warning

```
Failed to read entry point info from //node_modules/@schematics/angular/workspace/files/package.json with error SyntaxError: Unexpected token < in JSON at position 1121.
```

Fixes #13378


kyliau:
@alan-agius4 Could you please fix the conflicting files?

alan-agius4:
@kyliau rebase done.

kyliau:
@alan-agius4 Looks like there's still conflicts :(

alan-agius4:
@kyliau rebased again.

PR #13423, opened 2019-01-12
feat(@angular-devkit/build-angular): option to build and test only specified spec files

### LINKED ISSUE #3603: Run A Single Unit Test

FDIM:
A new `--spec` option is added. It is a glob/pattern that specifies which specs should be included in the build. 

Command:
 `ng test --spec=app/services/auth*`

When karma is already running, selected spec can be updated by running following command: 
`ng test --spec=app/login* --spec-update`

Few rules are applied to the pattern to make certain things possible:
* `.spec.ts` is added automatically if missing (less typing)
* if it ends with `*.ts` and doesn't have `.spec.ts`, extension will be replaced with `.spec.ts`
* if path starts at root, source root directory will be trimmed. So that you could use relative file path from editor to make custom tasks to run the command

Behind the scenes a new `test.generated.ts` is saved next to original `test.ts` which instead of using `require.context` imports all matching files. This gives quite a big boost in the project at work (at least 10s on each iterative build!)

Fixes #3603

Motivation: It is a pain in a big application to wait until entire app is rebuilt just to run single `fit` block. Takes almost a minute with 4000+ tests and barely works when code coverage is enabled. 

Missing: docs(?)

Example tasks for vscode:
````
{
  // See https://go.microsoft.com/fwlink/?LinkId=733558
  // for the documentation about the tasks.json format
  "version": "2.0.0",
  "tasks": [
    {
      "type": "shell",
      "label": "test selected spec",
      "command": "node_modules/.bin/ng",
      "args": [
        "test",
        "--spec",
        "${relativeFile}"
      ],
      "problemMatcher": []
    },
    {
      "type": "shell",
      "label": "switch selected spec",
      "command": "node_modules/.bin/ng",
      "args": [
        "test",
        "--spec",
        "${relativeFile}",
        "--spec-update"
      ],
      "problemMatcher": [],
      "isBackground": true
    }
  ]
}
````

PS. The setup for writing tests is quite nice!

EDIT: `spec-update` option is dropped and not part of the PR anymore

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
@@ -1,0 +1,58 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { logging } from '@angular-devkit/core';
+import { getOptions } from 'loader-utils';
+import { extname, join } from 'path';
+import { loader } from 'webpack';
+
+export interface SingleTestTransformLoaderOptions {
+  files: string[]; // list of paths relative to main
+  logger: logging.Logger;
+}
+
+export const SingleTestTransformLoader = require.resolve(join(__dirname, 'single-test-transform'));
+
+/**
+ * This loader transforms the default test file to only run tests
+ * for some specs instead of all specs.
+ * It works by replacing the known content of the auto-generated test file:
+ *   const context = require.context('./', true, /\.spec\.ts$/);
+ *   context.keys().map(context);
+ * with:
+ *   const context = { keys: () => ({ map: (_a) => { } }) };
+ *   context.keys().map(context);
+ * So that it does nothing.
+ * Then it adds import statements for each file in the files options
+ * array to import them directly, and thus run the tests there.
+ */
+export default function loader(this: loader.LoaderContext, source: string) {
+  const options = getOptions(this) as SingleTestTransformLoaderOptions;
+  const lineSeparator = process.platform === 'win32' ? '\r\n' : '\n';
+
+  const targettedImports = options.files
+    .map(path => `require('./${path.replace('.' + extname(path), '')}');`)
+    .join(lineSeparator);
+
+  // TODO: maybe a documented 'marker/comment' inside test.ts would be nicer?
+  const regex = /require\.context\(.*/;
+
+  // signal the user that expected content is not present
+  if (!regex.test(source)) {
+    const message = [
+      `The 'include' option requires that the 'main' file for tests include the line below:`,
+      `const context = require.context('./', true, /\.spec\.ts$/);`,
+      `Arguments passed to require.context are not strict and can be changed`,
+    ];
+    options.logger.error(message.join(lineSeparator));
+  }
+
+  const mockedRequireContext = '{ keys: () => ({ map: (_a) => { } }) };' + lineSeparator;
+  source = source.replace(regex, mockedRequireContext + targettedImports);
+
+  return source;
+}
--- a/packages/angular_devkit/build_angular/src/karma/index.ts
+++ b/packages/angular_devkit/build_angular/src/karma/index.ts
@@ -6,7 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
-import { resolve } from 'path';
+import { experimental, getSystemPath, join } from '@angular-devkit/core';
+import { dirname, resolve } from 'path';
 import { Observable, from } from 'rxjs';
 import { defaultIfEmpty, switchMap } from 'rxjs/operators';
 import * as webpack from 'webpack';
@@ -17,6 +18,11 @@ import {
   getTestConfig,
   getWorkerConfig,
 } from '../angular-cli-files/models/webpack-configs';
+import {
+  SingleTestTransformLoader,
+  SingleTestTransformLoaderOptions,
+} from '../angular-cli-files/plugins/single-test-transform';
+import { findTests } from '../angular-cli-files/utilities/find-tests';
 import { Schema as BrowserBuilderOptions } from '../browser/schema';
 import { ExecutionTransformer } from '../transforms';
 import { assertCompatibleAngularVersion } from '../utils/version';
@@ -24,7 +30,7 @@ import { generateBrowserWebpackConfigFromContext } from '../utils/webpack-browse
 import { Schema as KarmaBuilderOptions } from './schema';
 
 // tslint:disable-next-line:no-implicit-dependencies
-export type KarmaConfigOptions = import ('karma').ConfigOptions & {
+export type KarmaConfigOptions = import('karma').ConfigOptions & {
   buildWebpack?: unknown;
   configFile?: string;
 };
@@ -34,12 +40,12 @@ async function initialize(
   context: BuilderContext,
   webpackConfigurationTransformer?: ExecutionTransformer<webpack.Configuration>,
   // tslint:disable-next-line:no-implicit-dependencies
-): Promise<[typeof import ('karma'), webpack.Configuration]> {
-  const { config } = await generateBrowserWebpackConfigFromContext(
+): Promise<[experimental.workspace.Workspace, typeof import('karma'), webpack.Configuration]> {
+  const { config, workspace } = await generateBrowserWebpackConfigFromContext(
     // only two properties are missing:
     // * `outputPath` which is fixed for tests
     // * `budgets` which might be incorrect due to extra dev libs
-    { ...options as unknown as BrowserBuilderOptions, outputPath: '', budgets: undefined },
+    { ...((options as unknown) as BrowserBuilderOptions), outputPath: '', budgets: undefined },
     context,
     wco => [
       getCommonConfig(wco),
@@ -54,6 +60,7 @@ async function initialize(
   const karma = await import('karma');
 
   return [
+    workspace,
     karma,
     webpackConfigurationTransformer ? await webpackConfigurationTransformer(config[0]) : config[0],
   ];
@@ -63,71 +70,110 @@ export function execute(
   options: KarmaBuilderOptions,
   context: BuilderContext,
   transforms: {
-    webpackConfiguration?: ExecutionTransformer<webpack.Configuration>,
+    webpackConfiguration?: ExecutionTransformer<webpack.Configuration>;
     // The karma options transform cannot be async without a refactor of the builder implementation
-    karmaOptions?: (options: KarmaConfigOptions) => KarmaConfigOptions,
+    karmaOptions?: (options: KarmaConfigOptions) => KarmaConfigOptions;
   } = {},
 ): Observable<BuilderOutput> {
   // Check Angular version.
   assertCompatibleAngularVersion(context.workspaceRoot, context.logger);
 
   return from(initialize(options, context, transforms.webpackConfiguration)).pipe(
-    switchMap(([karma, webpackConfig]) => new Observable<BuilderOutput>(subscriber => {
-      const karmaOptions: KarmaConfigOptions = {};
+    switchMap(
+      ([workspace, karma, webpackConfig]) =>
+        new Observable<BuilderOutput>(subscriber => {
+          const karmaOptions: KarmaConfigOptions = {};
 
-      if (options.watch !== undefined) {
-        karmaOptions.singleRun = !options.watch;
-      }
+          if (options.watch !== undefined) {
+            karmaOptions.singleRun = !options.watch;
+          }
 
-      // Convert browsers from a string to an array
-      if (options.browsers) {
-        karmaOptions.browsers = options.browsers.split(',');
-      }
+          // Convert browsers from a string to an array
+          if (options.browsers) {
+            karmaOptions.browsers = options.browsers.split(',');
+          }
 
-      if (options.reporters) {
-        // Split along commas to make it more natural, and remove empty strings.
-        const reporters = options.reporters
-          .reduce<string[]>((acc, curr) => acc.concat(curr.split(',')), [])
-          .filter(x => !!x);
+          if (options.reporters) {
+            // Split along commas to make it more natural, and remove empty strings.
+            const reporters = options.reporters
+              .reduce<string[]>((acc, curr) => acc.concat(curr.split(',')), [])
+              .filter(x => !!x);
 
-        if (reporters.length > 0) {
-          karmaOptions.reporters = reporters;
-        }
-      }
+            if (reporters.length > 0) {
+              karmaOptions.reporters = reporters;
+            }
+          }
 
-      // Assign additional karmaConfig options to the local ngapp config
-      karmaOptions.configFile = resolve(context.workspaceRoot, options.karmaConfig);
+          // prepend special webpack loader that will transform test.ts
+          if (
+            webpackConfig &&
+            webpackConfig.module &&
+            options.include &&
+            options.include.length > 0
+          ) {
+            const mainFilePath = getSystemPath(join(workspace.root, options.main));
+            const files = findTests(
+              options.include,
+              dirname(mainFilePath),
+              getSystemPath(workspace.root),
+            );
+            // early exit, no reason to start karma
+            if (!files.length) {
+              subscriber.error(
+                `Specified patterns: "${options.include.join(', ')}" did not match any spec files`,
+              );
 
-      karmaOptions.buildWebpack = {
-        options,
-        webpackConfig,
-        // Pass onto Karma to emit BuildEvents.
-        successCb: () => subscriber.next({ success: true }),
-        failureCb: () => subscriber.next({ success: false }),
-        // Workaround for https://github.com/karma-runner/karma/issues/3154
-        // When this workaround is removed, user projects need to be updated to use a Karma
-        // version that has a fix for this issue.
-        toJSON: () => { },
-        logger: context.logger,
-      };
+              return;
+            }
 
-      // Complete the observable once the Karma server returns.
-      const karmaServer = new karma.Server(
-        transforms.karmaOptions ? transforms.karmaOptions(karmaOptions) : karmaOptions,
-        () => subscriber.complete());
-      // karma typings incorrectly define start's return value as void
-      // tslint:disable-next-line:no-use-of-empty-return-value
-      const karmaStart = karmaServer.start() as unknown as Promise<void>;
+            webpackConfig.module.rules.unshift({
+              test: path => path === mainFilePath,
+              use: {
+                // cannot be a simple path as it differs between environments
+                loader: SingleTestTransformLoader,
+                options: {
+                  files,
+                  logger: context.logger,
+                } as SingleTestTransformLoaderOptions,
+              },
+            });
+          }
 
-      // Cleanup, signal Karma to exit.
-      return () => {
-        // Karma only has the `stop` method start with 3.1.1, so we must defensively check.
-        const karmaServerWithStop = karmaServer as unknown as { stop: () => Promise<void> };
-        if (typeof karmaServerWithStop.stop === 'function') {
-          return karmaStart.then(() => karmaServerWithStop.stop());
-        }
-      };
-    })),
+          // Assign additional karmaConfig options to the local ngapp config
+          karmaOptions.configFile = resolve(context.workspaceRoot, options.karmaConfig);
+
+          karmaOptions.buildWebpack = {
+            options,
+            webpackConfig,
+            // Pass onto Karma to emit BuildEvents.
+            successCb: () => subscriber.next({ success: true }),
+            failureCb: () => subscriber.next({ success: false }),
+            // Workaround for https://github.com/karma-runner/karma/issues/3154
+            // When this workaround is removed, user projects need to be updated to use a Karma
+            // version that has a fix for this issue.
+            toJSON: () => {},
+            logger: context.logger,
+          };
+
+          // Complete the observable once the Karma server returns.
+          const karmaServer = new karma.Server(
+            transforms.karmaOptions ? transforms.karmaOptions(karmaOptions) : karmaOptions,
+            () => subscriber.complete(),
+          );
+          // karma typings incorrectly define start's return value as void
+          // tslint:disable-next-line:no-use-of-empty-return-value
+          const karmaStart = (karmaServer.start() as unknown) as Promise<void>;
+
+          // Cleanup, signal Karma to exit.
+          return () => {
+            // Karma only has the `stop` method start with 3.1.1, so we must defensively check.
+            const karmaServerWithStop = (karmaServer as unknown) as { stop: () => Promise<void> };
+            if (typeof karmaServerWithStop.stop === 'function') {
+              return karmaStart.then(() => karmaServerWithStop.stop());
+            }
+          };
+        }),
+    ),
     defaultIfEmpty({ success: false }),
   );
 }
--- a/packages/angular_devkit/build_angular/src/karma/schema.json
+++ b/packages/angular_devkit/build_angular/src/karma/schema.json
@@ -63,6 +63,13 @@
       "type": "string",
       "description": "Defines the build environment."
     },
+    "include": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "Globs of files to include, relative to workspace or project root. \nThere are 2 special cases:\n - when a path to directory is provided, all spec files ending \".spec.@(ts|tsx)\" will be included\n - when a path to a file is provided, and a matching spec file exists it will be included instead"
+    },
     "sourceMap": {
       "description": "Output sourcemaps.",
       "default": true,
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-tests.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-tests.ts
@@ -1,0 +1,62 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { existsSync } from 'fs';
+import * as glob from 'glob';
+import { basename, dirname, extname, join } from 'path';
+import { isDirectory } from './is-directory';
+
+// go through all patterns and find unique list of files
+export function findTests(patterns: string[], cwd: string, workspaceRoot: string): string[] {
+  return patterns.reduce(
+    (files, pattern) => {
+      const relativePathToMain = cwd.replace(workspaceRoot, '').substr(1); // remove leading slash
+      const tests = findMatchingTests(pattern, cwd, relativePathToMain);
+      tests.forEach(file => {
+        if (!files.includes(file)) {
+          files.push(file);
+        }
+      });
+
+      return files;
+    },
+    [] as string[],
+  );
+}
+
+function findMatchingTests(pattern: string, cwd: string, relativePathToMain: string): string[] {
+  // normalize pattern, glob lib only accepts forward slashes
+  pattern = pattern.replace(/\\/g, '/');
+  relativePathToMain = relativePathToMain.replace(/\\/g, '/');
+
+  // remove relativePathToMain to support relative paths from root
+  // such paths are easy to get when running scripts via IDEs
+  if (pattern.startsWith(relativePathToMain + '/')) {
+    pattern = pattern.substr(relativePathToMain.length + 1); // +1 to include slash
+  }
+
+  // special logic when pattern does not look like a glob
+  if (!glob.hasMagic(pattern)) {
+    if (isDirectory(join(cwd, pattern))) {
+      pattern = `${pattern}/**/*.spec.@(ts|tsx)`;
+    } else {
+      // see if matching spec file exists
+      const extension = extname(pattern);
+      const matchingSpec = `${basename(pattern, extension)}.spec${extension}`;
+
+      if (existsSync(join(cwd, dirname(pattern), matchingSpec))) {
+        pattern = join(dirname(pattern), matchingSpec).replace(/\\/g, '/');
+      }
+    }
+  }
+
+  const files = glob.sync(pattern, {
+    cwd,
+  });
+
+  return files;
+}
--- a/packages/angular_devkit/build_angular/test/karma/selected_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/karma/selected_spec_large.ts
@@ -1,0 +1,143 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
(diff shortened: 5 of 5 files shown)

### CODE REVIEW

**Comment by FDIM** [comment]
Original #12955 was closed because of inactivity - 1 day too soon. This one is as ready as I can make it

**Comment by FDIM** [comment]
@filipesilva let me know if you need any help with pushing this futher/making solution better

**Comment by filipesilva** [comment]
Heya @FDIM, thank you for tackling this feature! I want to give you an update on the status of this PR. 

We're looking at the feature as you have designed it so far and thinking about the pros and cons. We're also considering what extra work needs to be done to update projects to use it (e.g. ignoring the generated file). I should have some news soon as I drive it forward within the team!

**Comment by filipesilva** [comment]
@FDIM heya, we got together and discussed this PR yesterday. This PR adds two new pieces of functionality: running targeted specs, and changing the targeted spec of a running process. We feel there's a lot of value in the former, but comparatively less so in the latter since it's a subset of the usescases of the first.

The main gripe the team has with it is the temporary file. In this PR the temporary file is used to transform the original test code via regexes running it, and communicate between processes when updating the targeted spec

Creating temporary files on disk has a few problems though:
- requires a migration of existing projects (via `ng update`) to ignore it on git
- it's surprising for users that files are created in their sources
- it can remain on disk despite best efforts to remove it (e.g. process crash)
- multiple targeted spec runs will interfere with each other because they use the same temp file

So for now we'd like to not add the temporary file at all. There are some ways of limiting the specs without it so we can still have targeted specs. But changing targeted specs mid-run without it would not be easy. So for now we propose to have targeted specs but not changing them on the fly, and later we could add a way to do that as well. 

If we ignore the cross process communication we could easily avoid the temporary file by either
- editing the tsconfig `include` for that run, causing only some specs to be compiled, or
- creating a custom context for the require in `test.ts`, similar to what we do for lazy routes

I favor the first option myself because it also makes for a smaller, TS compilation, which should make both initial build and rebuilds faster.

I understand it's disappointing to be asked to reduce the scope of something you've designed and that covers your usecase very well. But for us as maintainers these questions of migration and usability are always key concerns. 

For instance, anything that requires a migration can only be on a major since our policy is to only require migrations for majors. On the other hand something that requires no interaction with the user source is much easier to land. Things that have any chance of leaving artifacts in the user sources are also scrutinized because they are hard to clean after.

Please let me know what you think.

**Comment by FDIM** [comment]
@filipesilva I agree 100% that generated/temp file is not ideal, but I do not know how else scope could be limited effectively without a way to feed webpack/typescript a stream/string instead of an entry point file.

I'll test if your theory holds by changing tsconfig `include` option and get back to you. Do you have any pointers for custom context? I'd like to see how that works as well.

On a flip side, I do not really agree with other team concerns:

> * requires a migration of existing projects (via `ng update`) to ignore it on git
> * it's surprising for users that files are created in their sources
> * it can remain on disk despite best efforts to remove it (e.g. process crash)

These steps can be fairly easily handled manually and documented as a quirk of `-spec` option for performance reasons - this is my number 1 problem for spending time on this! Not sure if you realized, but if you do not use the option - file will not be generated. I'd rather spend 5 minutes configuring my project manually and having fastest rebuild time than making this automatic process, but potentially slower (currently I have no idea if tsconfig include option would slower/faster and if it would be a working solution) and having to restart the process to switch to a different test. 
In other words, I am in favor of development experience/quick iteration when doing unit tests over configuration. Some things do not have simple solutions.

> * multiple targeted spec runs will interfere with each other because they use the same temp file

I'd say this is more like a limitation that you can have only single targeted test run at the same time per project. Isn't it safe to assume that one would use this option when working on a specific feature/task?

Thanks for looking into this!

**Comment by filipesilva** [comment]
This is where we manipulate the custom context for lazy routes: https://github.com/angular/angular-cli/blob/6fc6e84b0a0d84f04be5b21cf7c0a6411829c554/packages/ngtools/webpack/src/angular_compiler_plugin.ts#L697-L753

> These steps can be fairly easily handled manually and documented as a quirk of `--spec` option for performance reasons

Generally speaking, anything that requires a user to go figure out the quirks of the tool isn't great. And anything that requires a user to manually change their project isn't going to happen for the vast majority of projects.

Yes, the performance would be better. And I also understand that it would be much preferable for your usecase to have the "hot switch" feature. But I think that for most users, the base "targetted spec" feature gets them most of the way to faster test turnaround time without any quirks, pitfalls, needed upgrades, etc. Then the feature can be enhanced over time.

One thing I ask you to keep in mind is that we have a two release cycle deprecation policy. Since we release roughly every six months, that means that a new feature will need to be around and maintained for at least one year. I hope this puts into perspective why we favor less user-facing complexity and things that don't require upgrade paths.

**Comment by FDIM** [comment]
You have very valid points from maintenance point of view and I agree, but this does not help in landing such feature to the 'core'. I am also a bit puzzled why there is no demand for such option (maybe not many enterprise apps have tests... I hope not!) or what people do after they pass certain number of tests (gonna have to check jest).

Anyway, I've tried tinkering and customizing tsconfig include option and from my understanding solution would not be simpler. Basically I'd have to recursively find all related files to targeted specs and pass them via `files` option or `include` option. I don't think there is a way to pass a mask to typescript and let it handle this automatically. 

To make matters worse, consider this tsconfig:
```
{
    "extends": "./tsconfig.json",
    "exclude": [
        "./node_modules"
    ],
    "include": [
        "source/polyfills.ts",
        "source/test.ts",
        "source/common/components/input/**/*.ts"
    ]
}
```
One could expect that only files related to the listed ones would be included, but no - typescript also tries to compile files that are not referenced directly or indirectly (such as `source/common/services/*` or `source/mw/*`). Only changing `require.context('./', true, /\.spec\..s$/);` to `require.context('./', true, /common\/components\/input\/.*\.spec\..s$/);` in `test.ts` fixes the issue.


Could you share the solution that worked for you when you discussed include option as a potential alternative? 

In the meantime I'll investigate what custom context does and how it works - thanks for that!

**Comment by filipesilva** [comment]
The tsconfig you showed is quite different from the ones we generate for specs, mind you. We explicitely set both `files` and `include`. If you don't set `files`, everything is included by default.

Tried my hand at manipulating the tsconfig manually:
- made a new app with `ng new targeted-specs-tsconfig`
- added a failing spec `src/app/failing.spec.ts`: `it('should fail', () => expect(false).toBeTruthy());`
- `npm run test -- --watch=false --browsers=ChromeHeadless` fails
- edit `tsconfig` to have this include property:
```
  "include": [
    "**/app.component.spec.ts",
    "**/*.d.ts"
  ]
```
- `npm run test -- --watch=false --browsers=ChromeHeadless` fails with a badly formatted message that says
```
Uncaught Error: Module build failed (from ./node_modules/@ngtools/webpack/src/index.js):
Error: D:\\sandbox\\targeted-specs-tsconfig\\src\\app\\failing.spec.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
```

I expected that to work. But it doesn't because our webpack plugin errors out when requested TS files that aren't in the compilation.

We added this error specifically to avoid implicitly importing TS files. Maybe this is a restriction that we can relax on this particular case, and serve an empty file instead. But either way this approach got more complicated.

**Comment by filipesilva** [comment]
If you want to check how ignoring that error looks like, open `node_modules/@ngtools/webpack/src/angular_compiler_plugin.js` and comment out this block:
https://github.com/angular/angular-cli/blob/6fc6e84b0a0d84f04be5b21cf7c0a6411829c554/packages/ngtools/webpack/src/angular_compiler_plugin.ts#L1016-L1030

Doing that will work with my previous example and only run `src/app/app.component.spec.ts`.

I'm unsure of what the best API within the plugin for this would be, but probably an option that alters the include array, and ignore the error above if the file matches the alteration.

**Comment by FDIM** [comment]
@filipesilva I think I found a perfect solution!

I'll make a webpack loader (see last commit) that will be registered in karma builder for test.ts files which will transform it in the same way as in original suggestion.

I tested the idea in angular 8 project by modifying karma js file manually in node_modules and it seemed to work.

Now I have to figure out again how to work on cli locally

**Comment by FDIM** [comment]
I think I have it working! Only missing 2 things now: 
- support directory when one is passed as spec
- figure out sourceRoot from target (so it would also work in multi-project setup)

**Comment by FDIM** [comment]
PR is ready for review. @filipesilva would you mind having a look at current implementation and provide some feedback? Thanks!

Aside of that some help regarding failing circleci test would be great, as the same tests are passing locally and the feature is working as intended when `build_angular` package is installed in another project.

File: packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+
+export default function loader(this: loader.LoaderContext, source: string) {
+  const options = getOptions(this) as SingleTestTransformLoaderOptions;
+  const start = 'import \'';
+  const end = '\';';
+  const testCode = start + options.files
+    .map(path => `./${path.replace('.ts', '')}`)
+    .join(`${end}\n${start}`) + end;
+
+  // TODO: maybe a documented 'marker/comment' inside test.ts would be nicer?
```
FDIM: For our use case at work it would be also nice to be able to remove some imports / matching regexes when running specific test. We need to exclude an import that includes angularjs and basically entire project for old unit tests to pass. Should I cover such use case here? 

I am fine with monkey patching this script in post install until we finish upgrading, so personally it would not block me/us from using this feature.

packages/angular_devkit/build_angular/src/karma/index.ts
```diff
@@ -95,6 +111,33 @@ export function execute(
         }
       }
 
+      // prepend special webpack loader that will transform test.ts
+      if (webpackConfig && webpackConfig.module && options.spec) {
+        const sourceRoot = workspace.getProject(projectName).sourceRoot || '';
+        const files = findSpecs(options.spec, workspace.root, sourceRoot);
```
FDIM: I'll see if I can add array support - could be beneficial for CI environment to run only diff

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
```diff
@@ -0,0 +1,31 @@
...
+import { logging } from '@angular-devkit/core';
+import { getOptions } from 'loader-utils';
+import { loader } from 'webpack';
+
+export interface SingleTestTransformLoaderOptions {
+  files: string[];
+  logger: logging.Logger;
+}
+
+export default function loader(this: loader.LoaderContext, source: string) {
```
filipesilva: Can you add an explanation of what this loader does in a comment?

Something like:
```
/**
 * This loader transforms the default test file to only run tests
 * for some specs instead of all specs.
 * It works by replacing the known content of the auto-generated test file:
 *   const context = require.context('./', true, /\.spec\.ts$/);
 *   context.keys().map(context);
 * with:
 *   const context = { keys: () => ({ map: (_a) => { } }) };
 *   context.keys().map(context);
 * So that it does nothing.
 * Then it also adds import statements for each file in the files options
 * array to import them directly, and thus run the tests there.
 */
```

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
```diff
@@ -0,0 +1,31 @@
...
+  const start = 'import \'';
+  const end = '\';';
+  const testCode = start + options.files
+    .map(path => `./${path.replace('.ts', '')}`)
+    .join(`${end}\n${start}`) + end;
+
+  // TODO: maybe a documented 'marker/comment' inside test.ts would be nicer?
+  let mockedRequireContext = '{ keys: () => ({ map: (_a) => { } }) };';
+  mockedRequireContext += process.platform === 'win32' ? '\r\n' : '\n';
+  source = source.replace(/require\.context\(.*/, mockedRequireContext + testCode);
```
filipesilva: If we can't replace the file, we should bail out and offer an informative message instead. This way the user isn't left wondering why the command isn't working.

A good error message would be something along these lines:
```
Tried to transform main file for tests but could not find expected code.

The 'spec' option requires that the 'main' file for tests include the line below:
  const context = require.context('./', true, /\.spec\.ts$/);

Please ensure this line is present to use the `spec` option.  
```

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
```diff
@@ -0,0 +1,31 @@
...
+    input = input.substr(sourceRoot.length + 1); // +1 to include slash
+  }
+
+  if (input.endsWith('.ts') && !input.endsWith('.spec.ts')) {
+    input = input.substr(0, input.length - 2) + 'spec.ts';
+  } else if (!input.endsWith('.spec.ts') && input.indexOf('*') === -1) {
+    input += '**/*.spec.ts';
+  }
+  const files = glob.sync(input, {
+    cwd: getSystemPath(join(workspaceRoot, sourceRoot)),
```
filipesilva: WDYT of resolving the main test file instead, and use it's dir as `cwd`? That way you don't need to assume it's at the sourceRoot.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  input = input.replace(/\\/g, '/'); // normalize input
+
+  // remove source root to support relative paths from root
+  // such paths are easy to get when running script via IDEs
+  if (sourceRoot && input.startsWith(sourceRoot + '/')) {
+    input = input.substr(sourceRoot.length + 1); // +1 to include slash
+  }
+
+  if (input.endsWith('.ts') && !input.endsWith('.spec.ts')) {
```
filipesilva: I'm not sure I agree with this bit. Some projects might have their own spec suffixes that we shouldn't assume.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
```diff
@@ -0,0 +1,31 @@
...
+    input = input.substr(0, input.length - 2) + 'spec.ts';
+  } else if (!input.endsWith('.spec.ts') && input.indexOf('*') === -1) {
+    input += '**/*.spec.ts';
```
filipesilva: What is this case meant to address? Is it when the path is for a directory? We can test for that by checking the path exists and is a directly instead. I feel that's a bit safer because it assumes less. For instance, we also support `.tsx` and `.js`.

packages/angular_devkit/build_angular/src/karma/index.ts
```diff
@@ -95,6 +111,33 @@ export function execute(
...
+        }
+        // local unit test requires .ts extension
+        // without one specs will fail with unclear error
+        // release build does not need any extension
+        const ext = __filename.endsWith('.ts') ? '.ts' : '';
+
+        webpackConfig.module.rules.unshift({
+          test: /test\.ts$/,
+          use: {
+            loader: resolve(__dirname, '../angular-cli-files/plugins/single-test-transform' + ext),
```
filipesilva: You can do the same thing as below to get around this problem:
https://github.com/angular/angular-cli/blob/5df02a3de5257162c7b8a5e2c65d388debcd51f0/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/webpack.ts#L22

It's also a bit friendlier as it exposes it (even though it's not public API).

packages/angular_devkit/build_angular/src/karma/index.ts
```diff
@@ -95,6 +111,33 @@ export function execute(
         }
       }
 
+      // prepend special webpack loader that will transform test.ts
+      if (webpackConfig && webpackConfig.module && options.spec) {
+        const sourceRoot = workspace.getProject(projectName).sourceRoot || '';
+        const files = findSpecs(options.spec, workspace.root, sourceRoot);
+        // early exit, no reason to start karma
+        if (!files.length) {
+          subscriber.error(`Specified pattern: "${options.spec}" did not match any spec files`);
```
filipesilva: This may look like a small detail but it's really helpful!

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
filipesilva: I think that would be a bit too much right now... but an easy alternative is to have an alternate test configuration in `angular.json` that uses a different `main` for tests.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface SingleTestTransformLoaderOptions {
+  files: string[];
+  logger: logging.Logger;
+}
+
+export default function loader(this: loader.LoaderContext, source: string) {
+  const options = getOptions(this) as SingleTestTransformLoaderOptions;
+  const start = 'import \'';
+  const end = '\';';
+  const testCode = start + options.files
```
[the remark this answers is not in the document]
filipesilva: Actually, I think these can't be static `import` statements, because `import` statements will load their path before loading the current module.

Instead these need to be `require()` calls so that they are dynamic and happen at this particular point, after the setup in the beginning of the test file.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
FDIM: True, I can go with main file directory. I still need to know sourceRoot path though to support input glob relative to workspace root.

Will update

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
FDIM: I think I'll remove the option and require user to pass explicitly the file they want to run. 

To me this is useful sometimes - I can modify ts file, run custom vscode command to run selected test (while ts file is selected) and matching unit test ends up running.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
FDIM: Well, that only solves part of the problem - I would still need to know what files to look for.

Thinking more about it, it may be better to leave this for custom command in editor. 

Currently colleagues with webstorm are benefiting from this - they are able to right click a directory in files tree and run all unit test inside. It would be trivial to add `**/*.my-spec-files.ext` there.

packages/angular_devkit/build_angular/src/karma/index.ts
FDIM: Will test and see if this works in CI and local environment

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
filipesilva: You know, I was thinking about this a bit more and I think it's ok to assume `.spec.ts`, as long as that does not cause valid filenames to be ignored. We don't really give users options to change the spec suffix when generating files, but we don't stop them from changing it afterwards (and it will still work, given they change `test.ts` as well).

So for instance, if the file is `something.custom.ts`, and there does indeed exist a file on disk called `something.custom.spec.ts`, it's ok to use that one instead. But if `something.custom.spec.ts` doesn't exist, the original file should still be used.

WDYT? Does this make sense with the usecases you already have at your company?

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
filipesilva: Can you tell me a bit more about the right click case? I don't really use webstorm myself. Is it just right click, or something like right click, copy path?

Maybe we could go with the assumption that the dir case only works for `.spec.ts` as a convenience. It doesn't seem to cause any bad cases.

FDIM: I am not a user either, but from I've seen it's:
- right click
- find external tools option
- run predefined / imported 'tool'

It basically runs a specified command and can substitute certain variables, e.g. relative path from root of selected file / directory. This is much like tasks.json for vscode, but imported from xml or manually configured somewhere in settings.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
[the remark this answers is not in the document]
FDIM: I don't think the 2nd comment is a problem, this is the transformed `test.ts`:
image

Aside of that, I've replaced `${path.replace('.ts', '')`, with `${path.replace('.' + extname(path), '')` to be on the safe side (extname is imported from 'path')

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
FDIM: I am perfectly OK without assuming too many things. Such cases/conventions would need to be documented as well, otherwise it might be just me using them :)

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
FDIM: Now I remember why I did not use the value of `webpackConfig.entry`, it's type is `string | string[] | Entry | EntryFunc` - I don't know if its safe to assume that I'll get a string. Not sure if you know, but the last one in the list may also return a promise.

Somehow this line that you commented on seems simpler than supporting all variants of entry.

packages/angular_devkit/build_angular/src/karma/index.ts
FDIM: Seems to work during development and when it is build and installed

packages/angular_devkit/build_angular/src/karma/index.ts
```diff
@@ -95,6 +111,33 @@ export function execute(
...
+
+          return;
+        }
+        // local unit test requires .ts extension
+        // without one specs will fail with unclear error
+        // release build does not need any extension
+        const ext = __filename.endsWith('.ts') ? '.ts' : '';
+
+        webpackConfig.module.rules.unshift({
+          test: /test\.ts$/,
```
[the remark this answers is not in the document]
FDIM: I've changed this to `test: (path) => path === getSystemPath(join(workspace.root, options.main))` and this seem to work

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
FDIM: Done, now using `dirname(mainFilePath)`

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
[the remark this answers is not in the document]
filipesilva: I really thing the dynamic `require` are necessary. This is what I mean: https://stackoverflow.com/a/29334761/2116927

So even though the `import` is after `getTestBed()` (the important bit of initialization), it will run before it.

We don't have to assume this though. We can test it. 

- `npm i -g http-server`
- create these files
```
// index.html
<script src="./main.js" type="module"></script>

// main.js
console.log('before importing another-module');
import './another-module.js';
console.log('after importing another-module');

// another-module.js
console.log('inside another-module');
```
- run `http-server`
- open http://localhost:8080/, see console
```
another-module.js:1 inside another-module
main.js:1 before importing another-module
main.js:3 after importing another-module
```

You can see that the content of `another-module.js` was evaluated before the statement preceding the import statement.

In contrast, `require` statements are dynamic and sync, and will occur in that particular place. This of course is dependent on how Webpack implements the semantics, but as long as we stick to the intended semantics then it should be fine (and if it isn't, then it's a bug somewhere else).

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
filipesilva: Ok, then my request for these conditionals is to change it to this:
```
import { isDirectory } from './is-directory';
import { extname, basename } from 'path';

// ...

const matchingSpec = `${basename(input, extname(input))}.spec.ts`;

if (existsSync(matchingSpec)) {
  // Assume the user actually wants to run the spec, if there is one.
  input = matchingSpec;
} else if (isDirectory(input)) {
  // If we know it is a directory, instead assume the user wants the default spec glob inside.
  input += '**/*.spec.ts';
}
```

This way we don't exclude `.tsx` files and we verify the directory instead of something unexpected. Does this look good to you?

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/single-test-transform.ts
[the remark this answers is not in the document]
FDIM: hm, I did not expect behavior like this - Looks like I've learnt something new today. Thanks!

Aside of that, I've changed to use require instead of import when transforming test.ts. Still works :)

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/find-specs.ts
FDIM: It ended up a bit more complicated, but shares the same idea. Hopefully all will be good now

FDIM:
PR should be ready now. I've rebased from master and updated karma/schema.json with more details about `spec` option.

FDIM:
I just had to adjust the text in `schema.json`, apparently having `**/*` in description results in broken `schema.ts` file. As it ends multi-line comment.

packages/angular_devkit/build_angular/src/karma/index.ts
```diff
@@ -95,6 +112,31 @@ export function execute(
         }
       }
 
+      // prepend special webpack loader that will transform test.ts
+      if (webpackConfig && webpackConfig.module && options.spec) {
+        const sourceRoot = workspace.getProject(projectName).sourceRoot || '';
```
clydin: Why is `sourceRoot` used here?
The current behavior with the `require.context` call is that the files are relative to the `test.ts` (`mainFilePath` in this case).
However, I think it would be more intuitive if the option's base directory was the workspace root as this would be consist with all other options.  A project might also not have the concept of a source root (its currently really an extension option for the browser builder).
This would also eliminate the need for all the extra code to find the source root.

packages/angular_devkit/build_angular/src/karma/schema.json
```diff
@@ -63,6 +63,10 @@
       "type": "string",
       "description": "Defines the build environment."
     },
+    "spec": {
```
clydin: `spec` seems quite specific to jasmine.  I think either `tests` (or maybe `files`?) would be more descriptive of the purpose of the option.
Also should this be an array of strings for additional flexibility?
Or potentially, what about an `include`/`exclude` option set for maximum configurability and similarity to other configuration files (e.g., `tsconfig`).  I would tend to favor this last option.

Thoughts @filipesilva?

filipesilva: Similar pattern to TS sounds good. I'm not too keen on adding all `files/include/exclude` but maybe just `files` would be enough for now, and later `include/exclude` if needed.

FDIM: > spec seems quite specific to jasmine. I think either tests (or maybe files?) would be more descriptive of the purpose of the option.

I do not think that files is a good choice, it is not a path - it is a glob. To me `include` represents more closely to what it does and aligns to tsconfig. 

> Also should this be an array of strings for additional flexibility?

Maybe, but single value would be a good start. It is a bit more problematic as underlying lib does [not support multiple patterns](https://github.com/isaacs/node-glob/issues/217). There is a workaround with `{,,,}`

packages/angular_devkit/build_angular/src/karma/index.ts
FDIM: Well, option's base directory is workspace root - path relative to source root is an alternative that is exposed, but mainly needed internally. `sourceRoot` is simply subtracted from what is passed as `spec`, so that typescript would be able to find these files in `test.ts` (main) as path relative to it is needed (see [findSpecs](https://github.com/angular/angular-cli/pull/13423/files#diff-dccf1f76f21004c8dc59dd26881c6816R13)).

Any suggestions on how to refactor this to not rely on sourceRoot?

FDIM:
@filipesilva I see you've added 8.1 milestone, which is nice!
I am just a bit confused on what should I change to finalize this? Would it be enough to rename option from `spec` to `include` (or `files`)?

packages/angular_devkit/build_angular/src/karma/schema.json
clydin: I think `include` makes sense from the glob usage perspective.

packages/angular_devkit/build_angular/src/karma/index.ts
clydin: The `glob.sync` call's `cwd` option can be the workspace root and the `absolute` option would also be needed.  From there, get the absolute path of the directory name of the main file; and then loop and convert the test file list from the glob call to relative paths from the main file.

packages/angular_devkit/build_angular/src/karma/schema.json
clydin: For list support, the glob call can be called multiple times with all the results added to a Set to de-duplicate.

FDIM: Done, renamed option to `include` and changed it's type to array. Was easier than expected, so thanks for the push!

packages/angular_devkit/build_angular/src/karma/index.ts
FDIM: I've ended up with a bit different solution - sourceRoot is no longer used and I've reverted the changes. I hope it's ok now.

FDIM:
I've rebased from master and I think this is ready to go. `include` option can now be passed multiple times and appropriate specs are executed in test project.
It's just a bit strange that I see a warning that the option is passed multiple times. Can one of point out how one can pass an array to cli? e.g. `codeCoverageExclude`, there is no example in docs :(

filipesilva:
@FDIM I updated your original PR message to list the new `--include` name, and remove references to `--spec-update`. I was pointing people here and they were confused that the spec option wasn't listed in docs.

krokofant:
```
Angular CLI: 8.1.2
Node: 10.16.0
OS: win32 x64
Angular: 8.1.2
... router, upgrade

@angular-devkit/architect         0.800.6
@angular-devkit/build-angular     0.800.6
@angular-devkit/build-optimizer   0.800.6
@angular-devkit/build-webpack     0.800.6
@angular-devkit/core              7.3.4
@angular-devkit/schematics        8.1.2
@angular/cdk                      8.1.1
@ngtools/webpack                  8.0.6
@schematics/angular               8.1.2
@schematics/update                0.801.2
rxjs                              6.5.2
typescript                        3.4.5
webpack                           4.30.0
```

```
npx ng test --include something
Unknown option: '--include'
Unknown option: 'something'
```
Shouldn't this work?

filipesilva:
@krokofant yeah you need to update `@angular-devkit/build-angular` to include these changes.

PR #13427, opened 2019-01-13
feat(@schematics/angular): activate recommended tslint rules

cexbrayat:
PR https://github.com/angular/angular-cli/pull/13213 made `tslint.json` extend `tslint:recommended`.
As a first step, we deactivated explicitely all rules that were in the recommended set but were not in the old `tslint.json` file to keep the exact same configuration.

This PR activates a few of the recommended rules that were explicitely deactiviated, making te confoguration closer to the recommended one, and shorter. I tried to keep the "Angular style guide" and chose to activate the ones that I think are not controversial. Let me know what you think.

cc @mgechev

### DIFF
--- a/packages/schematics/angular/workspace/files/tslint.json
+++ b/packages/schematics/angular/workspace/files/tslint.json
@@ -4,21 +4,16 @@
     "codelyzer"
   ],
   "rules": {
-    "adjacent-overload-signatures": false,
-    "align": false,
     "array-type": false,
     "arrow-parens": false,
-    "ban-types": false,
     "deprecation": {
       "severity": "warn"
     },
-    "function-constructor": false,
     "import-blacklist": [
       true,
       "rxjs/Rx"
     ],
     "interface-name": false,
-    "jsdoc-format": false,
     "max-classes-per-file": false,
     "max-line-length": [
       true,
@@ -36,9 +31,6 @@
         ]
       }
     ],
-    "new-parens": false,
-    "no-angle-bracket-type-assertion": false,
-    "no-conditional-assignment": false,
     "no-consecutive-blank-lines": false,
     "no-console": [
       true,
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/workspace/files/tslint.json
```diff
@@ -4,22 +4,16 @@
...
       "severity": "warn"
     },
-    "function-constructor": false,
     "import-blacklist": [
       true,
       "rxjs/Rx"
     ],
     "interface-name": false,
-    "jsdoc-format": false,
-    "max-classes-per-file": false,
```
clydin: This should probably left as is since it is highly opinionated especially considering the recommended value is one class per file.

packages/schematics/angular/workspace/files/tslint.json
```diff
@@ -53,56 +44,21 @@
...
-    "no-namespace": false,
     "no-non-null-assertion": true,
     "no-redundant-jsdoc": true,
-    "no-reference": false,
-    "no-reference-import": false,
-    "no-string-literal": false,
     "no-switch-case-fall-through": true,
-    "no-unsafe-finally": false,
     "no-use-before-declare": true,
-    "no-var-requires": false,
```
clydin: Using require in this form is quite common.  I’m not sure if this is a good idea at this point in time.

cexbrayat: @clydin I don't mind removing it, but my understanding of the rule was that it doesn't forbid to use require, but only allows it in imports. And as it pushes to use the recommended ES2015 import syntax, I thought it was a good to include it (and I don't recall having to use `var lib = require('lib')` in an CLI project, but maybe you have an example in mind?).

cexbrayat:
@clydin Thanks for the review. For now, I deactivated `max-classes-per-file` per your feedback and amended the the PR, and I'll remove `no-var-require` if necessary after my answer on your comment.

packages/schematics/angular/workspace/files/tslint.json
clydin: I agree with the sentiment but it may be more problematic than useful.
Unfortunately it is also a TypeScript specific extension to the import syntax.  The [standard](https://www.ecma-international.org/ecma-262/9.0/#sec-imports) does not include it.  Users will also receive type errors if the module does not have type definitions available.

cexbrayat: All right I removed the change.

PR #13432, opened 2019-01-14
build: update `webpack` and `postcss`

alan-agius4:
Update `webpack` to `4.28.4` and `postcss` to `7.0.11`

PR #13436, opened 2019-01-14
build: update `webpack` and `postcss`

alan-agius4:
Backport https://github.com/angular/angular-cli/pull/13432 to patch


alan-agius4:
Back ported in 
https://github.com/angular/angular-cli/pull/13436

PR #13442, opened 2019-01-15
build: add Alan to the mailmap file

hansl:
He uses many addresses and some of his computers are not configured
properly :P


kyliau:
@alan-agius4 Is this change needed in the patch branch?

alan-agius4:
@kyliau updated label

PR #13444, opened 2019-01-15
fix(@schematics/angular): stop offering SASS option for ng new

alexeagle:
The old Sass language referred to as "SASS" here is no longer relevant. This was denoted with the file extension ".sass"
Any new projects should use the modern Sass language which is expressed in ".scss" files.
This change does not remove any support for projects which were already created with ".sass" files, we simply stop offering
this option when creating new projects.

Also correct the capitalization of Less based on how they spell it on their website.

### DIFF
--- a/packages/ngtools/webpack/README.md
+++ b/packages/ngtools/webpack/README.md
@@ -48,7 +48,7 @@ The loader works with webpack plugin to compile your TypeScript. It's important 
 ## Features
 The benefits and ability of using [`@ngtools/webpack`](https://www.npmjs.com/~ngtools) standalone from the Angular CLI as presented in [Stephen Fluin's Angular CLI talk](https://youtu.be/uBRK6cTr4Vk?t=6m45s) at Angular Connect 2016:
 
-* Compiles SCSS/LESS
+* Compiles Sass/Less into CSS
 * TypeScript transpilation
 * Bundles JavaScript, CSS
 * Asset optimization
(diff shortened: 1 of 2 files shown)


andy23512:
Hi all,
With reading some related issues, I still don't understand the reason for dropping support to `.sass` files. This causes some trouble to ones who want to use the old [indented-syntax](http://sass-lang.com/documentation/file.INDENTED_SYNTAX.html) (at least in newly generated projects). Is it dropped just because it is an **old syntax**? Could someone explain the reason for me, please?

ericcoulthard:
I would also like to know the reason for dropping support to `.sass` files. Weren't .sass and .scss released at the same time? So it doesn't make sense to me to call .sass the old language.

DGmip:
I thought there was a **bug** (ng g c component-name --style=sass no longer works, either, it's producing scss files ) 
I can't believe they merged this...
What about those of us who prefer using the sass indented syntax?
This is really a problem for my workflow as I am using 
`let g:angular_cli_stylesheet_format = 'sass'`
in my .vimrc file.
This really needs to be reverted.

yangshuairocks:
I'm about to submit a bug and I found this. Could someone explain the reason why sass support was removed? You can remove a feature in Angular but please at least describe a little bit about the reason.

DGmip:
> I'm about to submit a bug and I found this. Could someone explain the reason why sass support was removed? You can remove a feature in Angular but please at least describe a little bit about the reason.

This is already added back in about 10 days ago; see the reference to the previous comment.

PR #13445, opened 2019-01-15
fix(@schematics/angular): remove SCSS from ng-new prompt

kyliau:
The stylesheet extension `.sass` should be deprecated in favor of `.scss`.
See SASS documentation [here](http://sass-lang.com/guide).

In `ng new`, the **value** `scss` generates stylesheets with `.scss`
extension whereas the **value** `sass` generates stylesheets with
`.sass` extension. 
We want `.scss` extension. However, the **label** for `.scss` extension should
be SASS.

This is a non-breaking change since it only affect new CLI projects.

### DIFF
--- a/packages/schematics/angular/ng-new/schema.json
+++ b/packages/schematics/angular/ng-new/schema.json
@@ -119,8 +119,7 @@
         "type": "list",
         "items": [
           { "value": "css", "label": "CSS" },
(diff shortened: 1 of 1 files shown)


kyliau:
Alex already merged to master: https://github.com/angular/angular-cli/commit/4f358f48f850a5a5c80d51a597b5dec809cd208e

PR #13448, opened 2019-01-16
Highlight creating workspace without initial app in doc

jbogarthyde:
I just learned that you can create a workspace without the anomalous initial app in /src, and then populate it with apps using `ng g a`, so that multiple apps are all in /projects (with libs).  This is much more desirable for a workspace with multiple apps.

I think `--createApplication==false` should be the default in future to avoid confusion and awkwardness for new users -- but for now, I'm proposing that we point out the benefits in the API doc for `ng new`.

### DIFF
--- a/packages/schematics/angular/ng-new/schema.json
+++ b/packages/schematics/angular/ng-new/schema.json
@@ -132,12 +132,12 @@
       "alias": "S"
     },
     "createApplication": {
-      "description": "When true (the default), creates a new initial app project in the new workspace.",
+      "description": "When true (the default), creates a new initial app project in the src folder of the new workspace. When false, creates an empty workspace with no initial app. You can then use the generate application command so that all apps are created in the projects folder.",
       "type": "boolean",
       "default": true
     },
     "minimal": {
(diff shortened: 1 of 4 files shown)


packages/schematics/angular/ng-new/schema.json
```diff
@@ -132,12 +132,12 @@
...
     },
     "createApplication": {
-      "description": "When true (the default), creates a new initial app project in the new workspace.",
+      "description": "When true (the default), creates a new initial app project in the src folder of the new workspace. When false, creates an empty workspace with no initial app. You can then use the generate application command so that all apps are created in the projects folder.",
       "type": "boolean",
       "default": true
     },
     "minimal": {
-      "description": "When true, creates a project without any testing frameworks and should be used for learning purposes only.",
+      "description": "When true, creates a project without any testing frameworks. (Use for learning purposes only).",
```
alan-agius4: Can we also update the minimal description in the other schemes. So that they are all aligned?

See https://github.com/angular/angular-cli/pull/13408/files for reference.

PR #13463, opened 2019-01-17
New Architect API

hansl:
For now just a proof of concept, but this branch will be reused for the final implementation.


packages/angular_devkit/architect/src/api.ts
```diff
@@ -0,0 +1,230 @@
...
+   * system path and will not be normalized; ie. on Windows it will starts with `C:\\` (or
+   * whatever drive).
+   */
+  currentDirectory: string;
+
+  /**
+   * The target that was used to run this builder.
+   * Target is optional if a builder was ran using `scheduleBuilder()`.
+   */
+  target?: Target;
```
clydin: What do you think about moving this to `input` (or rather combining them)?
This would allow for future _mutli-target_ builders that are aware of all their targets. Such a builder would receive a list of "target/inputs" and then act upon them as efficiently as possible.  The tslint builder for instance could be applicable here as this would allow the builder to generate one comprehensive output (especially important for certain output reporters such as XML; which is currently broken in this scenario).

packages/angular_devkit/architect/src/api.ts
```diff
@@ -0,0 +1,230 @@
...
+  scheduleBuilder(
+    builderName: string,
+    options?: json.JsonObject,
+  ): Promise<BuilderRun>;
+
+  /**
+   * Set the builder to running. This should be used if an external event triggered a re-run,
+   * e.g. a file watched was changed.
+   */
+  running(): void;
```
clydin: This should have a verb.  Currently from the name I'd expect it to return a boolean signifying whether it is running.

`setRunning`? `reportRunning`?  -- depending on what it actually does.

Same for status and progress.  I believe this would help to make the API more intuitive.
There is already `schedule` as a verb/action above.  I think having `report` as well would nicely define the set of actions that can be performed via the context.  A builder can either schedule additional executions or report on its behavior.

packages/angular_devkit/architect/src/api.ts
hansl: The `input` is not accessible to builders, only `context` (this interface) and `options`. Also, since a builder is always instantiated once per target and/or per builder name itself, I'm not sure that would make sense. And unless the builder is watching, one input => one context => one builder call, so theoretically that `target` member could change between two calls.

For a multi-target builder, I think we should extend this interface and (potentially) make it a separate class, passing in `targets[]` or something.

packages/angular_devkit/architect/src/api.ts
hansl: Yes, I agree, and I scratched my head a bit on that one. `report` is good.

packages/angular_devkit/architect/src/api.ts
clydin: Sorry I meant `options`.  As long as there would be a way to match the options with the targets in that scenario, I'm ok with that.  Also, I wouldn't make all builders accept an array just ones that would be created/marked as multi-target (`createMultitargetBuilder` maybe?).  Just didn't want to box ourselves into a corner moving forward.

hansl: Options should not have stuff that cannot be validated by the builder. It’s whats in the options object in the workspace. We could do smart defaults like in schematics but considering smart defaults never really went anywhere I’d like to avoid it for now. 

I think a separate builder class would be the way to go.

packages/angular_devkit/core/src/terminal/index.ts
```diff
@@ -6,4 +6,5 @@
  * found in the LICENSE file at https://angular.io/license
  */
 export * from './text';
+export * from './caps';
```
clydin: This doesn't appear to be used anywhere?

packages/angular_devkit/architect/node/node-modules-architect-host.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  async getOptionsForTarget(target: Target): Promise<json.JsonObject> {
+    const targetSpec = this._workspace.getProjectTargets(target.project)[target.target];
+    if (target.configuration && !targetSpec['configurations']) {
+      throw new Error('Configuration not set in the workspace.');
+    }
+
+    return {
+      ...targetSpec['options'],
+      ...(target.configuration ? targetSpec['configurations'][target.configuration] : 0),
```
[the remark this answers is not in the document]
hansl: As per offline discussion, `{ ...0 }` is valid javascript.

packages/angular_devkit/architect/src/schedule-by-name.ts
```diff
@@ -0,0 +1,127 @@
...
+    stateSubscription = job.outboundBus.subscribe(event => {
+      if (event.kind === experimental.jobs.JobOutboundMessageKind.Start) {
+        job.input.next({
+          id,
+          currentDirectory: workspaceRoot,
+          workspaceRoot: currentDirectory,
+          info: info,
+          options: buildOptions,
+          ...(options.target ? { target: options.target } : {}),
+        });
```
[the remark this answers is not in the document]
hansl: Done.

packages/angular_devkit/core/src/terminal/index.ts
hansl: It is in the architect CLI.

hansl: (see the progress bar stuff).

hansl:
Just force pushed the `fixup`s into their commits. This is ready to merge pending CI.

PR #13473, opened 2019-01-18
docs(polyfills.ts): Clarify about reflect API and when it is needed

oBusk:
I found it very confusing if we needed `es6/reflect` or not when looking at `polyfills.ts` and https://angular.io/guide/browser-support#mandatory-polyfills 

I Think that this might clarify it somewhat. I also created a PR to try to clarify the docs (https://github.com/angular/angular/pull/28238) 

I don't know if the wording was the greatest, please poke me about any improvements I can do

### DIFF
--- a/tests/angular_devkit/build_webpack/angular-app/src/polyfills.ts
+++ b/tests/angular_devkit/build_webpack/angular-app/src/polyfills.ts
@@ -44,7 +44,11 @@
 /** IE10 and IE11 requires the following for NgClass support on SVG elements */
 // import 'classlist.js';  // Run `npm install --save classlist.js`.
 
-/** IE10 and IE11 requires the following for the Reflect API. */
+/**
+ * IE10 and IE11 requires the following for the Reflect API.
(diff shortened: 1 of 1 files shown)


clydin:
With the latest CLI 7.3 beta, neither polyfill is required to be specified directly.  If running in JIT mode, the CLI will automatically include the needed JIT specific polyfills.

oBusk:
> With the latest CLI 7.3 beta, neither polyfill is required to be specified directly. If running in JIT mode, the CLI will automatically include the needed JIT specific polyfills.

But the clarification I was trying to do was to explain why `es6/reflect` is included in `polyfills.ts` at all? If neither is needed, should it be omitted from `polyfills.ts` completely?

clydin:
They’re actually not anymore.  The file you are changing is a test project used for CI.  The new project template for the file can be found here: https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/application/files/src/polyfills.ts.template

However, thank you for taking the time to help and contribute to the project.

PR #13474, opened 2019-01-18
Build-optimizer fixes for FESM/ESM bundles

filipesilva:
fix(@angular-devkit/build-optimizer): prefix renamed classes
Module concatenation may rename classes with the same name. The renaming logic is specific to the bundler so we can't really foresee it.

But the fact remains that the inner function declaration doesn't need to have the same name as the outer one.

---

fix(@angular-devkit/build-optimizer): identify relative imports in angular core

Build optimizer was broken for non-FESM files inside @angular/core because it couldn't identify relative imports were still inside core.

This change adds a known list of angular core files as a default, and also allows passing in a override.

### DIFF
--- a/etc/api/angular_devkit/build_optimizer/src/_golden-api.d.ts
+++ b/etc/api/angular_devkit/build_optimizer/src/_golden-api.d.ts
@@ -12,6 +12,8 @@ export declare function getPrefixFunctionsTransformer(): ts.TransformerFactory<t
 
 export declare function getScrubFileTransformer(program: ts.Program): ts.TransformerFactory<ts.SourceFile>;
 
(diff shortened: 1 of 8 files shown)


clydin:
It's used to detect Angular specific decorators and remove them if found.  Only angular decorators are safe to remove since they are known to not contain any side effects.

filipesilva:
@alexeagle `@angular/core` is a special case because it contains relative imports.  

We have a list of specifiers that should be scrubbed out:

https://github.com/angular/angular-cli/blob/c5de8e3b0e5c1499a0a17da0208297cd82b36095/packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts#L33-L50

At the moment we detect imports to `@angular/core`, and known decorator imports inside the angular core FESMs. But not relative imports for the ESM bundles because we didn't know they were still part of angular core.

Issue #13487, opened 2019-01-21
Optimize ES2015 static fields


filipesilva:
# 🚀 Feature request

### Description
Using the ES2015 target, TypeScript transpiles classes with static fields as a class plus the assignment of the static fields over the class itself.

As an example, the `TemplateRef` class in https://github.com/angular/angular/blob/73dcd72afbd782161c1674f1e647a1b357a78a67/packages/core/src/linker/template_ref.ts#L34-L62:
```
export abstract class TemplateRef<C> {
  abstract get elementRef(): ElementRef;

  abstract createEmbeddedView(context: C): EmbeddedViewRef<C>;

  static __NG_ELEMENT_ID__:
      () => TemplateRef<any>| null = () => SWITCH_TEMPLATE_REF_FACTORY(TemplateRef, ElementRef)
}
```

Is transpiled to ES2015 as:
```
class TemplateRef {
}
TemplateRef.__NG_ELEMENT_ID__ = () => SWITCH_TEMPLATE_REF_FACTORY(TemplateRef, ElementRef);
```

The property assignment in the transpiled code prevents the class from being removed by minifiers such as terser.

### Describe the solution you'd like
Build Optimizer converts ES2015 static fields in such a way that they are not retained after minification.

PR #13496, opened 2019-01-22
feat(@angular-devkit/build-optimizer): also fold ES2015 classes

filipesilva:
Although ES5 classes had their static properties folded in, ES2015 ones did not.

This PR adds that new functionality.

It should also make this particular transform a bit faster since it will stop early.

Fix https://github.com/angular/angular-cli/issues/13487

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/transforms/class-fold.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/class-fold.ts
@@ -9,8 +9,8 @@ import * as ts from 'typescript';
 
 interface ClassData {
   name: string;
-  class: ts.VariableDeclaration;
-  classFunction: ts.FunctionExpression;
+  declaration: ts.VariableDeclaration | ts.ClassDeclaration;
+  function?: ts.FunctionExpression;
   statements: StatementData[];
 }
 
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_optimizer/src/transforms/class-fold.ts
```diff
@@ -61,8 +67,49 @@ export function getFoldFileTransformer(program: ts.Program): ts.TransformerFacto
...
+            '@__PURE__',
+            false,
+          );
+
+          // Move the original class modifiers to the var statement.
+          const newNode = ts.createVariableStatement(
+            clazz.declaration.modifiers,
+            ts.createVariableDeclarationList([
+              ts.createVariableDeclaration(clazz.name, undefined, pureIife),
+            ]),
```
alan-agius4: small nit: Maybe add `ts.NodeFlags.Const` as last arg to `ts.createVariableDeclarationList`

packages/angular_devkit/build_optimizer/src/transforms/class-fold.ts
```diff
@@ -61,8 +67,49 @@ export function getFoldFileTransformer(program: ts.Program): ts.TransformerFacto
...
-          return ts.visitEachChild(newNode, visitor, context);
+          return newNode;
+        }
+
+        // Check if node is a ES2015 class to replace with a pure IIFE.
+        clazz = classes.find((cl) => !cl.function && cl.declaration === node);
+        if (clazz) {
+          // There doesn't seem to be a way to make a statement from a class declaration.
+          // But it seems to work if you just jam it in there.
+          const classStatement = clazz.declaration as {} as ts.ExpressionStatement;
```
clydin: A `ClassDeclaration` is already a statement.  The casting shouldn't be necessary.

packages/angular_devkit/build_optimizer/src/transforms/class-fold.ts
```diff
@@ -80,6 +127,19 @@ function findClassDeclarations(node: ts.Node): ClassData[] {
   const classes: ClassData[] = [];
   // Find all class declarations, build a ClassData for each.
   ts.forEachChild(node, (child) => {
+    // Check if it is a named class declaration first.
+    // Technically it doesn't need a name in TS if it's the default export, but when downleveled
+    // it will be have a name (e.g. `default_1`).
+    if (ts.isClassDeclaration(child) && child.name && ts.isIdentifier(child.name)) {
```
clydin: `ts.isIdentifier` is unneeded here.  `name` is either undefined or an `Identifier`.

filipesilva: Fixed

packages/angular_devkit/build_optimizer/src/transforms/class-fold.ts
filipesilva: Fixed

packages/angular_devkit/build_optimizer/src/transforms/class-fold.ts
filipesilva: Fixed (the right cast was to `ts.ClassDeclaration`, since it could be either that or `ts.VariableDeclaration`).

PR #13515, opened 2019-01-24
fix(@schematics/angular): update latest version of devkit

cexbrayat:
Same thing that happened in the `beta.0` relase: devkit has not been updated when releasing `rc.0`.

Installing the CLI `7.3.0-rc.0` brings devkit in version `0.13.0-beta.0`


alan-agius4:
@cexbrayat, thanks for this however, `~0.13.0-beta.0` installs the RC version.

We'll be updated the latest versions when we cut the stable release.

Issue #10128, opened 2018-04-02
[@ngtools/webpack] Webpack cannot distinguish on context and would fail to load the proper one.


evgeniyefimov:
### Versions
```
node: v8.11.0
npm: 5.7.1
Windows 10
```
```
"@angular/animations": "5.2.9",
"@angular/cdk": "5.2.4",
"@angular/common": "5.2.9",
"@angular/compiler": "5.2.9",
"@angular/core": "5.2.9",
"@angular/forms": "5.2.9",
"@angular/http": "5.2.9",
"@angular/platform-browser": "5.2.9",
"@angular/platform-browser-dynamic": "5.2.9",
"@angular/router": "5.2.9",
"@angular/compiler-cli": "5.2.9",
"@ngtools/webpack": "1.10.2",
"typescript": "2.7.2",
"webpack": "3.11.0",
```

### Repro steps
**app-routing.module.ts**
```
export const ROUTES: Routes = [
    {
        loadChildren: "./student/student.module#StudentModule",
        path: "student",
    },
];
```

**student-routing.module.ts**
```
export const ROUTES: Routes = [
     {
         loadChildren: "./question/question.module#QuestionModule",
         path: "question",
     },
     {
         loadChildren: "./settings/settings.module#SettingsModule",
         path: "settings",
     },
];
```

**question-routing.module.ts**
```
export const ROUTES: Routes = [
     {
         loadChildren: "./settings/settings.module#SettingsModule",
         path: "settings",
     },
];
```

### Observed behavior
```
ERROR in NaNbut they point to different modules "(C:/Projects/Web/App/src/student/settings/settings.module.ts and "C:/Projects/Web/App/src/student/questions/settings/settings.module.ts"). Webpack cannot distinguish on context and would fail to load the proper one.
```

### Desired behavior
We have a lot of modules with the same names (settings, instructions, etc.), but they are belong to different parents with different urls, they are not exported outside of the parent's scope. It works fine with awesome-typescript-loader, angular-router-loader and angular2-template-loader. We expected the same behavior with `@ngtools/webpack`.

### Mention any other details that might be useful (optional)

I removed all other lazy modules from the project to check and this issue was still there.
Tried to use absolute paths as described here https://github.com/angular/angular-cli/issues/8722#issuecomment-366002955, but had another issue
`ERROR in Could not resolve module App/src/student/student.module relative to C:/Projects/Web/App/src/app-routing.module.ts`

**`tsconfig.json`**
```
{
[...]

PR #13524, opened 2019-01-25
fix(@schematics/angular): enable lazy loading with ivy

cexbrayat:
Angular `8.0.0-beta.1` supports lazy loading with Ivy (at least a partial support), but currently the CLI needs `"allowEmptyCodegenFiles": true` to avoid errors like:

```
ERROR in ./src/$$_lazy_route_resource lazy namespace object
Module not found: Error: Can't resolve '/src/app/races/races.module.ngfactory.js' in '/src/$$_lazy_route_resource'
```

As discussed with @gkalpak this might be a temporary workaround, but it might help users giving a try to Ivy (as the workaround is not really trivial ^^) .

### DIFF
--- a/packages/schematics/angular/application/files/root/tsconfig.app.json.template
+++ b/packages/schematics/angular/application/files/root/tsconfig.app.json.template
@@ -9,6 +9,7 @@
     "**/*.spec.ts"
   ]<% if (experimentalIvy) { %>,
   "angularCompilerOptions": {
-    "enableIvy": "ngtsc"
+    "enableIvy": "ngtsc",
+    "allowEmptyCodegenFiles": true
   }<% } %>
 }


gkalpak:
Related Jira issue: [FW-737]

[FW-737]: https://angular-team.atlassian.net/browse/FW-737

alan-agius4:
The last time I talked with @igor, he mentioned that we should enable this by default. So I am going to wait for his feedback.

alan-agius4:
The last time I talked with @igor, he mentioned that we should NOT enable this by default. So I am going to wait for his feedback.

alan-agius4:
Hi, Igor,

I didn’t look into this recently, but in the webpack pipeline, we do change
the modules and file paths to factories when not having a jit compilation.

We should definitely change this and do so only for AOT and when enabling
factories shimming.

However, I doubt that lazy loading will work in ivy without factories
because of the linker in core, as it always expects to have factories
https://github.com/angular/angular/blob/master/packages/core/src/linker/system_js_ng_module_factory_loader.ts

clydin:
Can we have the Ivy AOT compiler normalize the string form to the `System.import`/`import` form?  While solving this problem, it would also provide several nice value added benefits:
* In AOT mode, remove the need for the CLI to perform lazy route detection and special context webpack module mapping (a large/complex unit of code)
* Remove a class of hard to fix lazy route naming related defects such as this: https://github.com/angular/angular-cli/issues/10128
* Allow for more optimized bundling and potential reduced overall application bundle sizes and loading times.  The lazy route hierarchy would no longer be flattened which would allow for the option of common modules to be directly included in the lowest common ancestor instead of always in separately loaded lazy bundles.

alan-agius4:
Superseded by https://github.com/angular/angular-cli/pull/13667

PR #13525, opened 2019-01-25
docs(@angular-devkit/schematics): fix example

peterbsmith2:


packages/angular_devkit/schematics/README.md
```diff
@@ -173,7 +173,7 @@ export default function (options: ClassOptions): Rule {
       ]
     );
 
-    return branchAndMerge(mergeWith(templateSource));
+    return branchAndMerge(mergeWith(templateSource))(tree, context);
```
alan-agius4: Hi, thanks for this, however the current documentation is correct.

Users don't need to invoke this method.

alan-agius4: The call itself, should be redundant.

Not sure why in your case it fails without it. We have a couple of schematics similar to this example which you can take a look at;
https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/enum/index.ts

peterbsmith2: After looking into it more I understand that the `(tree, context)` call is used for making composable calls...as in, this call allows the schematic output to be passed to another. Is that right?

PR #13571, opened 2019-02-01
fix(@angular-devkit/build-angular): throw error when multiple bundles…

### LINKED ISSUE #13568: Wrong file order in index.html when bundleName is used
# 🐞 Bug report

```
- [x] build
- [x] serve
```

### Is this a regression?
No

### Description
When i use a `bundleName` like the following inside the `angular.json` for the `styles` array. The order of scripts generated in the index.html are in the wrong order.

`angular.json`
```
"build": {
          "options": {
            "outputPath": "dist",
            "index": "src/index.html",
            "main": "src/ng2/main.ts",
            "polyfills": "src/ng2/polyfills.ts",
            "tsConfig": "src/tsconfig.app.json",
            "assets": [
              "src/ng2/favicon.ico",
              "src/ng2/assets"
            ],
            "styles": [
              { "input": "src/ng2/main.scss", "bundleName": "main" }
            ],
            "scripts": [
              "build/main-ng1.js"
            ]
          },
```

Order of scripts in `index.html` after running `ng build`
```
<script type="text/javascript" src="runtime.js"></script>
<script type="text/javascript" src="polyfills.js"></script>
<script type="text/javascript" src="main.js"></script>
<script type="text/javascript" src="scripts.js"></script>
```

The Problem is, that the `main.js` get inserted before the `scripts.js` and this will fail if you have some external dependency added via `scripts.js`

I have already digged into the implementation in `@angular-devkit/build-angular/src/angular-cli-files/plugins/index-html-webpack-plugin.js` and found out that the problem is the bundle name `main` which is added into the `this._options.entrypoints` array before the `scripts` entrypoint is added.

## 🔬 Minimal Reproduction
See example above

## 🌍 Your Environment
<pre><code>
Angular CLI: 7.3.0
Node: 8.11.1
OS: darwin x64
Angular: 7.2.0
... router, upgrade

@angular-devkit/architect         0.12.1
@angular-devkit/build-angular     0.12.1
@angular-devkit/build-optimizer   0.12.1
@angular-devkit/build-webpack     0.12.1
@angular-devkit/core              7.2.1
@angular-devkit/schematics        7.3.0
@angular/cli                      7.3.0
@ngtools/webpack                  7.2.1
@schematics/angular               7.3.0
@schematics/update                0.13.0
typescript                        3.2.2
</code></pre>

alan-agius4:
… have been name the same

Naming more than 1 bundle with the same name might cause undefined behavior.

Fixes #13568

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/package-chunk-sort.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/package-chunk-sort.ts
@@ -11,23 +11,36 @@ import { normalizeExtraEntryPoints } from '../models/webpack-configs/utils';
 export function generateEntryPoints(
   appConfig: { styles: ExtraEntryPoint[], scripts: ExtraEntryPoint[] },
 ) {
-  const entryPoints = ['es2015-polyfills', 'polyfills', 'sw-register'];
 
   // Add all styles/scripts, except lazy-loaded ones.
-  [
-    ...normalizeExtraEntryPoints(appConfig.styles, 'styles')
+  const extraEntryPoints = (
+    extraEntryPoints: ExtraEntryPoint[],
+    defaultBundleName: string,
+  ): string[] => {
+    const entryPoints = normalizeExtraEntryPoints(extraEntryPoints, defaultBundleName)
       .filter(entry => !entry.lazy)
-      .map(entry => entry.bundleName),
-    ...normalizeExtraEntryPoints(appConfig.scripts, 'scripts')
-      .filter(entry => !entry.lazy)
-      .map(entry => entry.bundleName),
-  ].forEach(bundleName => {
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/utilities/package-chunk-sort.ts
```diff
@@ -29,5 +29,13 @@ export function generateEntryPoints(
 
   entryPoints.push('main');
 
+  const duplicateBundles = entryPoints.find(x => (
```
clydin: This would only catch `main` since the loop above prevents double inclusion of any other name.

alan-agius4: updated.

PR #13552, opened 2019-01-31
fix(@schematics/angular): set config when using sass as preprocessor

alan-agius4:
Fixes #13550

PR #13591, opened 2019-02-04
fix(@schematics/angular): use `scss` file extension instead of `sass` for `styles.scss`

yjaaidi:
When using `ng new --style=sass`, `ng-new` schematic generates a `styles.scss` file but in `build` and `test` architects in `angular.json` references `src/styles.sass`.

This fix solves the mismatch issue for users generating an application with `style` option set to `sass`.


clydin:
Thank you for the contribution.  However, this is a duplicate of https://github.com/angular/angular-cli/pull/13552

yjaaidi:
D'OH! 😱
I don't understand how I missed the issue. I've looked for it though... but I don't remember what keywords I used 😅

PR #13596, opened 2019-02-05
docs: link to angular.io/cli instead of old wiki

cexbrayat:
The page on cli.angular.io still link to the old wiki instead of the new docs.


cexbrayat:
@alan-agius4 Done 👌

PR #13598, opened 2019-02-05
Workspaces -- Phase 1

clydin:
NOTE: No changes have yet been made to the public API of the package.

This introduces the core and eventual public API for the stable workspace API.  This is not intended to be a feature complete PR but rather represents the initial base infrastructure necessary for pending feature additions via future PRs.

Development Plan:
* Phase 1 (this PR) - Public API (not exposed until phase 3) and format agnostic reading/writing support
* Phase 2 - JSON format reader/writer for existing format (includes change tracking support thereof)
* Phase 3 - Integration into `@schematics/angular` package and exposure of public API
* Phase 4 - Query API additions (e.g., finding specific project target names/builders)


packages/angular_devkit/core/src/workspace/core_spec.ts
```diff
@@ -0,0 +1,370 @@
...
+
+        return false;
+      },
+    };
+
+    await readWorkspace(requestedPath, host, WorkspaceFormat.JSON);
+
+    isFileChecks.sort();
+    expect(isFileChecks).toEqual([
+      getSystemPath(join(normalize(requestedPath), 'angular.json')),
```
filipesilva: How do you verify other formats are not attempted? I think you need to not return file contents and verify all attempted file checks.

packages/angular_devkit/core/src/workspace/definitions.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+export interface TargetDefinition {
+  options?: Record<string, JsonValue | undefined>;
+  configurations?: Record<string, Record<string, JsonValue | undefined> | undefined>;
+
+  builder: string;
+}
+
+export type DefinitionCollectionListener<V> = (
```
filipesilva: What are our usecases for the listeners?

packages/angular_devkit/core/src/workspace/definitions.ts
```diff
@@ -0,0 +1,255 @@
+/**
```
filipesilva: There are no tests for this file.

packages/angular_devkit/core/src/workspace/definitions.ts
clydin: Formats may require change tracking of the collections to facilitate fine-grained updates of the underlying storage medium.  The JSON format, for instance, allows for this.

packages/angular_devkit/core/src/_golden-api.ts
```diff
@@ -9,7 +9,7 @@
 export * from './exception/exception';
 
  // Start experimental namespace
-export * from './workspace/index';
+export * from './experimental/workspace/index';
```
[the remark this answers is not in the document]
clydin: The existing workspace elements were already experimental from the public API surface.  The file structure was misaligned in that aspect.  The first commit in the PR corrects this and moves the actual files into a more appropriate location.  These existing elements are still available as the new subsystem is not yet exposed publicly as this PR represents the core infrastructure for additional work which will be completed in subsequent PRs.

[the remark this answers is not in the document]
filipesilva: I agree that the old one was in the wrong folder and that `refactor(@angular-devkit/core): move existing workspace API under experimental` corrects that.

But I don't agree with keeping the old one and new one side by side in master. It feels to me like it would cause unnecessary extra work to develop it master. Mostly extra reviews, API updates, and rot prevention. 

The alternative would be to develop this replacement in a feature branch and wholly replace the old workspace when it is ready. This seems like a better alternative to me.

Are there reasons why the feature branch option is inadvisable here?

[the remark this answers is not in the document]
filipesilva: Discussed offline, this phased development model had already been discussed but I was not aware.

packages/angular_devkit/core/src/workspace/definitions.ts
filipesilva: Discussed offline, tests for this one will be in a future phase.

packages/angular_devkit/core/src/workspace/definitions.ts
filipesilva: SGTM

packages/angular_devkit/core/src/workspace/core_spec.ts
clydin: Updated with a test only hook for the supported files.

clydin:
Fixed that and now the test cleans itself up.

PR #13604, opened 2019-02-06
fix(@angular/cli): Add an error message when we call `ng doc` without…

wKoza:
… keyword. Currently, we open a page with the keyword `undefined`.

### DIFF
--- a/packages/angular/cli/commands/doc-impl.ts
+++ b/packages/angular/cli/commands/doc-impl.ts
@@ -14,6 +14,11 @@ const opn = require('opn');
 
 export class DocCommand extends Command<DocCommandSchema> {
   public async run(options: DocCommandSchema & Arguments) {
+    if (!options.keyword) {
+      this.logger.error('You should specify a keyword, for instance, `ng doc ActivatedRoute`.');
+
+      return 0;
+    }
     let searchUrl = `https://angular.io/api?query=${options.keyword}`;
     if (options.search) {
       searchUrl = `https://www.google.com/search?q=site%3Aangular.io+${options.keyword}`;


packages/angular/cli/commands/doc-impl.ts
```diff
@@ -14,13 +14,18 @@ const opn = require('opn');
...
-      searchUrl = `https://www.google.com/search?q=site%3Aangular.io+${options.keyword}`;
-    }
+    if (options.keyword) {
+      let searchUrl = `https://angular.io/api?query=${options.keyword}`;
+      if (options.search) {
+        searchUrl = `https://www.google.com/search?q=site%3Aangular.io+${options.keyword}`;
+      }
 
-    return opn(searchUrl, {
+      return opn(searchUrl, {
```
mgechev: The indentation doesn't look right. `wait: false` needs two more tabs?

wKoza: 😱

packages/angular/cli/commands/doc-impl.ts
```diff
@@ -14,13 +14,18 @@ const opn = require('opn');
 
 export class DocCommand extends Command<DocCommandSchema> {
   public async run(options: DocCommandSchema & Arguments) {
-    let searchUrl = `https://angular.io/api?query=${options.keyword}`;
-    if (options.search) {
-      searchUrl = `https://www.google.com/search?q=site%3Aangular.io+${options.keyword}`;
-    }
+    if (options.keyword) {
```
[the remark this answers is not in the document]
wKoza: done

mgechev:
Validate seem unhappy because of the commit message

wKoza:
Yes, I've seen but I'm not well equipped `Windows` :wink:...

wKoza:
validate doesn't like `--fixup` ? OK OK ...

PR #13605, opened 2019-02-06
feat(@angular/cli): Add VsCode extension recommendations

mrmeku:


mgechev:
I'd say we should keep this outside of the CLI. Not all users of Angular CLI are using VSCode.

mrmeku:
@mgechev Isn't VSCode the recommended editor for Angular? 

Its referenced in the style guide https://angular.io/guide/styleguide#file-templates-and-snippets
And also the setup guide https://angular.io/guide/language-service#visual-studio-code

The goal of the PR was only to get new comers to angular started quickly. Given that the setup guide calls out for them to start with VS Code doesn't it make sense to guide those user's toward having the core extensions for that editor as well?

packages/schematics/angular/migrations/migration-collection.json
```diff
@@ -29,6 +29,11 @@
       "version": "7.0.3",
       "factory": "./update-7/index#updateDevkitBuildNgPackagr",
       "description": "Update an Angular CLI project to version 7."
+    },
+    "migration-07": {
+      "version": "7.4.0",
```
alexeagle: we aren't going to ship a 7.4
can you change this to 8?

packages/schematics/angular/migrations/update-vscode-recommendations/update-7.4.0.ts
```diff
@@ -0,0 +1,29 @@
...
+import { updateJsonInTree } from './ast-utils';
+
+export default function(): Rule {
+  return chain([addExtensionRecommendations]);
+}
+
+const addExtensionRecommendations = updateJsonInTree(
+  '.vscode/extensions.json',
+  (json: { recommendations?: string[] }) => {
+    ['angular.ng-template', 'eg2.tslint', 'nrwl.angular-console'].forEach(
```
alexeagle: Let's not introduce angular-console in the same PR, that's going to require a mit more discussion with DevRel. Open a follow-up PR adding that.

@mgechev points out that the language service doesn't work with Ivy yet but I think that's okay for v8 that we recommend an extension for most users who don't opt-in

mgechev: There's some additional work going on around the language service // cc @kyliau and @IgorMinar

mgechev:
What about the prompt?

packages/schematics/angular/migrations/migration-collection.json
mrmeku: I chose 7.4.0 under the assumption that there will be a version of the cli released with that version number. Is that accurate? Otherwise I can make it 8.0.0

packages/schematics/angular/migrations/update-vscode-recommendations/update-7.4.0.ts
mrmeku: Cool, I'll remove console for this one. I also removed tslint. Would you like me to add tslint back in?

packages/schematics/angular/migrations/migration-collection.json
mrmeku: Definitely

mrmeku:
> What about adding a prompt?

Can you elaborate on what you mean? This feature isn't flag controlled so I'm not sure what a prompt would do exactly

packages/schematics/angular/migrations/update-vscode-recommendations/update-7.4.0.ts
alexeagle: yeah tslint belongs because the schematic creates tslint.json

alexeagle:
I considered including a prompt but I don't think the extra overhead of asking everyone this question and making `ng new` take longer is worth the marginal benefit of some people not getting the file created.

packages/schematics/angular/migrations/update-vscode-recommendations/ast-utils.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Rule, Tree } from '@angular-devkit/schematics';
```
alexeagle: seems kinda crazy to me that schematics doesn't have some reusable helper for the common case of mutating some json.

Aside, have you seen https://stedolan.github.io/jq/ ? it's an amazing command-line based tool that gives you unix sed-style power over JSON

packages/schematics/angular/workspace/files/__dot__vscode/extensions.json.template
```diff
@@ -0,0 +1,5 @@
+{
+  "recommendations": [
+    "angular.ng-template"
```
alexeagle: let's add back tslint

packages/schematics/angular/migrations/update-vscode-recommendations/update-7.4.0.ts
mrmeku: Done

packages/schematics/angular/migrations/update-vscode-recommendations/ast-utils.ts
mrmeku: We have a lot of these kinds of useful helpers in our nx repo to help with making cli schematics. Maybe someday we can move some of the better utilities upstream if you think they would be handy

packages/schematics/angular/workspace/files/__dot__vscode/extensions.json.template
mrmeku: Done

packages/schematics/angular/migrations/update-vscode-recommendations/ast-utils.ts
mrmeku: jq looks really cool I think I'm going to start looking into using that more in the future. We've been trying to get values from JSONs on disk without parsing them and that might do the trick

mrmeku:
This is ready for another round of comments or to be merged

packages/schematics/angular/migrations/update-vscode-recommendations/ast-utils.ts
alexeagle: I started hooking jq into the bazel pipeline at one point, works nicely as an independant action. maybe should do that again and land it...

clydin: The upcoming workspace writer feature requires non-destructive (retention of comments/formatting/etc) JSON5 mutation support.  Once that lands, the underlying components can then be generalized and moved to an additional Public API for scenarios such as this.

PR #13606, opened 2019-02-07
fix(@angular-devkit/build-angular): correct the default value of es5B…

mgechev:
…rowserSupport

Based on [this value](https://github.com/angular/angular-cli/blob/86356a747dccad75e4721a7d57e68c856a431c0b/packages/schematics/angular/application/index.ts#L213).

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -303,7 +303,7 @@
     "es5BrowserSupport": {
       "description": "Enables conditionally loaded ES2015 polyfills.",
       "type": "boolean",
-      "default": false
+      "default": true
     }
   },
   "additionalProperties": false,


clydin:
For 8.0, this should be enabled by default as it will bring in the differential building as well.  The default tsconfig target should also be moved to ES2015 as well once that lands.

mgechev:
Just a recap of the discussion from Slack: the schema value is used as a fallback (default) value for the `es5BrowserSupport` flag. If folks update Angular CLI to 7.3, they will get the polyfills from polyfills.js (which include es2015) and es2015-polyfills.js, which includes polyfills.js + es2015 polyfills. We do not want this, because the size of both bundles will grow.

Alternatively, we can keep the es2015 bundles in polyfills.js and use only a subset of the bundles in the internal polyfill bundle (currently called es2015-polyfills.js), but this would introduce additional complexity.

PR #13614, opened 2019-02-07
fix(@angular-devkit/build-angular): error when both `baseUrl` and `devServerTarget` are used in protractor

### LINKED ISSUE #13611: --base-url has no effect
# 🐞 Bug report

The option `--base-url` has no effect. 

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [x] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 6.x.x

### Description

 
The option `--base-url` has no effect on the `e2e`-command. Also the `baseUrl`-entry in the `protractor.conf.js` has no effect.

## 🔬 Minimal Reproduction

Create new project via `ng new demo` and there modify the `e2e`-script or the `protractor.conf.js` as described. It has no affect. The app is always started at `http://localhost:4200` and the browser also opens this url. 
I have to run the tests in a no-serve scenario on a different url.

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍 Your Environment
<pre><code>

macos
@angular/cli@7.3.0 
</code></pre>

**Anything else relevant?**

alan-agius4:
fix(@angular-devkit/build-angular): error when both `baseUrl` and `devServerTarget` are used in protractor

Users cannot use both baseUrl and devServerTarget, and the later superseeds the baseUrl.

Fixes #13611

### DIFF
--- a/packages/angular_devkit/build_angular/src/protractor/index.ts
+++ b/packages/angular_devkit/build_angular/src/protractor/index.ts
@@ -44,7 +44,14 @@ export class ProtractorBuilder implements Builder<ProtractorBuilderOptions> {
     const options = builderConfig.options;
     const root = this.context.workspace.root;
     const projectRoot = resolve(root, builderConfig.root);
-    // const projectSystemRoot = getSystemPath(projectRoot);
+
+    // ensure that either one of this option is used
+    if (options.devServerTarget && options.baseUrl) {
+      throw new Error(tags.stripIndents`
+      The 'baseUrl' option cannot be used with 'devServerTarget'.
+      When present, 'devServerTarget' will be used to automatically setup 'baseUrl' for Protractor.
+      `);
+    }
 
     // TODO: verify using of(null) to kickstart things is a pattern.
     return of(null).pipe(
@@ -81,7 +88,7 @@ export class ProtractorBuilder implements Builder<ProtractorBuilderOptions> {
         }
 
         // Compute baseUrl from devServerOptions.
-        if (options.devServerTarget && builderConfig.options.publicHost) {
+        if (builderConfig.options.publicHost) {
           let publicHost = builderConfig.options.publicHost;
           if (!/^\w+:\/\//.test(publicHost)) {
             publicHost = `${builderConfig.options.ssl
@@ -90,7 +97,7 @@ export class ProtractorBuilder implements Builder<ProtractorBuilderOptions> {
           }
           const clientUrl = url.parse(publicHost);
           baseUrl = url.format(clientUrl);
-        } else if (options.devServerTarget) {
+        } else {
           const result: DevServerResult | undefined = buildEvent.result;
 
           baseUrl = url.format({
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/protractor/index.ts
```diff
@@ -81,7 +85,7 @@ export class ProtractorBuilder implements Builder<ProtractorBuilderOptions> {
         }
 
         // Compute baseUrl from devServerOptions.
-        if (options.devServerTarget && builderConfig.options.publicHost) {
+        if (builderConfig.options.publicHost) {
```
alan-agius4: This method is never called when `options.devServerTarget` is falsy

https://github.com/angular/angular-cli/blob/8c441d3f76d61d45076cfc2ef3fbcbf818422a14/packages/angular_devkit/build_angular/src/protractor/index.ts#L55

packages/angular_devkit/build_angular/src/protractor/index.ts
```diff
@@ -44,7 +44,11 @@ export class ProtractorBuilder implements Builder<ProtractorBuilderOptions> {
     const options = builderConfig.options;
     const root = this.context.workspace.root;
     const projectRoot = resolve(root, builderConfig.root);
-    // const projectSystemRoot = getSystemPath(projectRoot);
+
+    // ensure that either one of this option is used
+    if (options.devServerTarget && options.baseUrl) {
+      throw new Error(`The 'baseUrl' option cannot be used with 'devServerTarget'.`);
```
filipesilva: I think it makes sense to error out since we use the devServerTarget to compute baseUrl. But we should be explicit here. Something like:

```
The 'baseUrl' option cannot be used with 'devServerTarget'.
When present, 'devServerTarget' will be used to automatically setup 'baseUrl' for Protractor.
```

clydin: I would actually expect it to override what's in the dev server target.  Otherwise, the baseUrl setting seems unnecessary as the overwhelming usage would be with a dev server target.

alan-agius4: What if host or port are also provided?

clydin: The same.  Who knows how the developer has the environment setup. Or how they want protractor to connect to the application.

alan-agius4: I mean, if host and or port are provided with baseUrl which one should override which?

clydin: Well, that's really a different question.  But I suppose it raises the additional question of why there are two ways to do the same thing?

alan-agius4: I’ll put this up for discussion.

alan-agius4:
@filipesilva, PR updated following the team meeting

I left the error message, as it's best to error out rather than have undefined behaviour, for the time being until this option is removed.

PR #13615, opened 2019-02-07
fix(@angular/cli): change strategy for opening the browser when we ca…

### LINKED ISSUE #13575: ng doc does not open the documentation in the browser.
# 🐞 Bug report

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [x ] doc
```

### Is this a regression?

### Description

Typing `ng doc xxx` does not do anything. It does not generate an error and does not open the browser to the documentation.

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍 Your Environment
<pre><code>

Angular CLI: 7.2.3
Node: 10.15.0
OS: win32 x64
Angular:
...

@angular-devkit/architect    0.12.3
@angular-devkit/core         7.2.3
@angular-devkit/schematics   7.2.3
@schematics/angular          7.2.3
@schematics/update           0.12.3
rxjs                         6.3.3
typescript                   3.2.2

</code></pre>

**Anything else relevant?**

wKoza:
…ll `ng doc` with Windows.

Closes #13575

### DIFF
--- a/packages/angular/cli/commands/doc-impl.ts
+++ b/packages/angular/cli/commands/doc-impl.ts
@@ -24,8 +24,11 @@ export class DocCommand extends Command<DocCommandSchema> {
       searchUrl = `https://www.google.com/search?q=site%3Aangular.io+${options.keyword}`;
     }
 
-    return opn(searchUrl, {
-      wait: false,
+    // We should wrap `opn` in a new Promise because `opn` is already resolved
+    await new Promise(() => {
+      opn(searchUrl, {
+        wait: false,
+      });
     });
   }
 }


packages/angular/cli/commands/doc-impl.ts
```diff
@@ -19,8 +19,14 @@ export class DocCommand extends Command<DocCommandSchema> {
       searchUrl = `https://www.google.com/search?q=site%3Aangular.io+${options.keyword}`;
     }
 
-    return opn(searchUrl, {
-      wait: false,
-    });
+    let opt;
+    if (process.platform === 'win32') {
+      // window doesn't support to be called without `wait: true`
+      opt = { app: 'chrome', wait: true };
```
alan-agius4: Is providing an app necessary?

Does the following work on windows?
https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/dev-server/index.ts

wKoza: Yes, it's the problem (look at [this](https://github.com/sindresorhus/opn#wait)).
I'm not sure to understand. `WebpackDevServer` ? of course, we are limited, here, by  `opn`.

alan-agius4: we use `opn` in other places, and we don't use `wait: true`, nor `wait: false`. 

We use the following;
```
opn(opnAddress);
```


Does this mean that `ng serve --open` is also broken on windows?

wKoza: OK. `ng serve --open` works well with Windows 10. I'll test on Windows 7 tomorrow.
`ng doc`doesn't work neither on 10 nor seven.

`wait: true` is the default value, so we use `wait:true` when we call `ng serve --open`. Finally,  
> On Windows you have to explicitly specify an app for it to be able to wait

seems not to be right...

I did not pay attention but the behavior is not the same whether we have already the browser open or not. If the browser is already open, `wait:true` doesn't block the prompt but it do if the browser is closed. So, it's not that is expected.

wKoza: A workaround could be this:

```
return new Promise((res,err) => {
			opn(searchUrl, {
                           wait: false,
                         })
                          .then((r) => setTimeout(()=> res(r),500))
			 .catch((e) => err(e));
		});

```

alan-agius4: @wKoza, I have looked into this a bit and it seems that we need to resolve the promise once the process fired the `open` event.

I think it should be something like:

```ts
        const openProcess = await opn(searchUrl, { wait: false })

        return new Promise((resolve, reject) => {
            openProcess
                .once('error', reject)
                .once('open', () => resolve());
        });
```

Ps: its untested but I think it should work. If you manage to get this working on a Windows machine. I'll try to test it out on a Mac if you don't have one. Just to be double sure if it works cross-platform.

wKoza: Bad news, on a Windows 7, `ng serve --open` doesn't work:

notfound

If I change the code of `dev-server/index.ts` like this:

```
if (first && options.open) {
                first = false;
                opn(opnAddress, {
                  app: 'chrome',
                  wait: true,
                });
            }
            return buildEvent;
```

I suppose I'm on [this case](https://webpack.js.org/configuration/dev-server/#devserver-open):
> If no browser is provided (as shown above), your default browser will be used. To specify a different browser, just pass its name instead of boolean:

I think we could add this option to angular-cli.

But, returning to the specifics of what we've been talking about, your code works but this codes works also:

```
opn(searchUrl, {
          app: chromeName,
          wait: false,
        });
new Promise((resolve, reject) => {});
```

or 

````
return new Promise((res,err) => {
     opn(searchUrl, {
       app: chromeName,
       wait: false,
     })
   });
````

Comment (alan-agius4) [reply packages/angular/cli/commands/doc-impl.ts]: Strange, also more surprising 

doing the below seems to work

```
    return from(opn(searchUrl, { wait: false }))
      .pipe(
        map(() => undefined),
      )
      .toPromise();
```

Comment (alan-agius4) [reply packages/angular/cli/commands/doc-impl.ts]: Maybe the cleanest one I can think of would be, what do you think?;

```
await Promise.all([opn(searchUrl, { wait: false })]);

return;
```

**Comment by alan-agius4** [comment]
@wKoza, just a side note, we should not add an addition open for the browser to open. We should always rely on the default browser.

File: packages/angular/cli/commands/doc-impl.ts (hunk 1 above)
Comment (wKoza) [reply packages/angular/cli/commands/doc-impl.ts]: I've tested but it doesn't work. `maybeExitCode` is undefined [here](https://github.com/angular/angular-cli/blob/master/packages/angular/cli/lib/cli/index.ts#L32).

The only code where we can grap the 'resolve' [here](https://github.com/angular/angular-cli/blob/master/packages/angular/cli/lib/cli/index.ts#L32)' is :

```js
return new Promise((resolve, reject) => {
      opn(searchUrl, {
        app: chromeName,
        wait: false
      }).then((r) => setTimeout(()=> resolve(r), 1000))
    });
```

Otherwise, this code "works", 

```
return new Promise((res,err) => {
     opn(searchUrl, {
       app: chromeName,
       wait: false,
     })
   });
```

But we cannot grap the `resolve`, it's too bad.

And I understand about the option.

wKoza: I've tested but it doesn't work, the browser doesn't  start :confused:

wKoza:
@alan-agius4 , I understand concerning the option property. I hadn't default browser (It's not my PC ...); But now It's ok; I no longer need of the `app` property

alan-agius4:
@wKoza small nit, would you mind fixing the commit msg? As the description is not correct.


Thanks

PR #13620, opened 2019-02-08
fix(@angular-devkit/build-angular): remove unneeded script element type

clydin:
The `type` attribute for `script` elements is not required for HTML5.


mgechev:
Two failures because of incorrect hash of a file. LGTM otherwise!

PR #13626, opened 2019-02-08
fix(@schematics/angular): use enableIvy as true

filipesilva:
Blocked on https://github.com/angular/angular/pull/28616


filipesilva:
The test-large isn't but the e2e is. I think this PR needs to wait for the next release of angular to use that in the e2e test.

Issue #13477, opened 2019-01-18
No warning given when using ES2017 


filipesilva:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

No.

### Description

Zone.js does not support native async/await (https://github.com/angular/zone.js/pull/1140) as present in ES2017. This means code that happens inside async/await blocks is invisible to zone.js thus not triggering change detection.

Since this is a footgun for developer, Angular CLI should show a warning about this when ES2017 is set as the typescript target.

## 🔬 Minimal Reproduction
https://github.com/angular/zone.js/pull/1140

## 🔥 Exception or Error
No error, just no change detection.

## 🌍 Your Environment
<pre><code>
Angular CLI 7.x
</code></pre>

**Anything else relevant?**

PR #13643, opened 2019-02-09
fix(@angular-devkit/build-angular): added warning for ES2017

hrocha16:
Fixes https://github.com/angular/angular-cli/issues/13477

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import { tags } from '@angular-devkit/core';
 import * as CopyWebpackPlugin from 'copy-webpack-plugin';
 import * as path from 'path';
 import { HashedModuleIdsPlugin, debug } from 'webpack';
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -272,6 +272,12 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
     );
   }
 
+  if (wco.tsConfig.options.target === 4) {
+    wco.logger.warn(`\nWARNING: Zone.js does not support native async/await
```
alan-agius4: Hi, you need to use `stripIndent` in order to avoid having extra spacing in the terminal.

```
        wco.logger.warn(tags.stripIndent`
          WARNING: Zone.js does not support native async/await in ES2017.  
          These blocks are not intercepted by zone.js and will not triggering change detection.
          See: https://github.com/angular/zone.js/pull/1140 for more information.
        `);
      
```

alan-agius4:
Kindly squash your commits into a single one.

Thanks.

hrocha16:
Thanks!

alan-agius4:
@filipesilva PTAL

filipesilva:
LGTM, thanks for taking this on @hrocha16!

liorwohl:
Hi, is there a way to hide the warning? I dont even use zone.js (its an Angular Element) but I get this warning anyway.
Thanks

PR #13648, opened 2019-02-11
feat(@angular/cli): add new force flag for e2e

NisStrom:


alan-agius4:
Thank you for the contribution. However, ideally before opening a PR file file an issue, so that we can discussion the use-case and see if it falls in the CLI scope.

Thanks,

PR #13667, opened 2019-02-13
fix(@ngtools/webpack): don't load ngfactories for lazy routes with Ivy

filipesilva:
Supersedes https://github.com/angular/angular-cli/pull/13524


filipesilva:
Related and probably blocked on https://github.com/angular/angular/pull/28685

PR #13684, opened 2019-02-15
fix(@angular-devkit/build-angular): disable by default stylesheet root relative URL rebasing

clydin:
BREAKING CHANGE:
Root relative URLs are a standardized method to reference a resource path from the root of a host.  The previous behavior of the Angular CLI prevented this from occuring and resulted in an inability to reference stylesheet assets in this manner.  The initial reason for this behavior is no longer present in the internal implementation of the Angular CLI.  Therefore, this now unnecessary and non-standard behavior is being phased out.  If an application currently relies on this behavior, a compatibility option `rebaseRootRelativeCssUrls` has been provided for the 8.x release cycle to facilitate transition away from this non-standard and limiting behavior.  The recommended method to transition is to use relative paths within the source stylesheet.  This allows the build system to process and generate a full URL for the asset.

This change also aligns the CLI behavior with other commonly used tools including Webpack's css-loader. The compatibility option will only be needed for projects that used root relative URLs within stylesheets and also used either a custom HTML base HREF and/or deploy URL option.
This option is a new option for the `build` target of an application (using the default `@angular-devkit/build-angular:browser` builder).  For example:
```
"build": {
    "options": {
        ...
        "rebaseRootRelativeCssUrls": true,
        ...
    },
    ...
}
```

To further demonstrate the behavioral differences, assuming that a stylesheet has a `url()` containing `/image.jpg`.  The URL would be transformed as follows.  Note that in the behavior prior to 8.0, referencing an asset from the root of the origin is not present when providing the HTML base HREF or deploy URL options.

|  | Prior to 8.0 | 8.0 and later |
|:-:|:-:|:-:|
| Base HREF of `/base/` |`/base/image.jpg`|`/image.jpg`|
| Deploy URL of `deployed/` |`deployed/image.jpg`|`/image.jpg`|
| Both of the above | `/base/deployed/image.jpg` | `/image.jpg` |
| **Deploy URL of `/deployed/` |`/deployed/image.jpg`|`/image.jpg`|

** -- For prior behavior, the base HREF will be ignored when the deploy URL is root relative.

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -875,6 +875,12 @@
               "description": "Enables conditionally loaded ES2015 polyfills.",
               "type": "boolean",
               "default": false
+            },
+            "rebaseRootRelativeCssUrls": {
+              "description": "Change root relative URLs in stylesheets to include base HREF and deploy URL. Use only for compatibility and transition. The behavior of this option is non-standard and will be removed in the next major release.",
+              "type": "boolean",
+              "default": false,
+              "x-deprecated": true
             }
           },
           "additionalProperties": false,
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
@@ -71,6 +71,8 @@ export interface BuildOptions {
   lazyModules: string[];
   platform?: 'browser' | 'server';
   fileReplacements: CurrentFileReplacement[];
+    /** @deprecated use only for compatibility in 8.x; will be removed in 9.0 */
+  rebaseRootRelativeCssUrls?: boolean;
 }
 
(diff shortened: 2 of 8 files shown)


packages/angular/cli/lib/config/schema.json
```diff
@@ -875,6 +875,12 @@
               "description": "Enables conditionally loaded ES2015 polyfills.",
               "type": "boolean",
               "default": false
+            },
+            "rebaseRootRelativeCssUrls": {
+              "description": "Change root relative URLs in stylesheets to include base HREF and deploy URL. Use only for compatibility and transition.  This behavior of this option is non-standard and will be removed in the next major release.",
```
alan-agius4: Nit This behavior of this option  -> The behavior of this option

Also note you have a double space before ‘this’

clydin: Done.

Issue #13680, opened 2019-02-14
support of pnpm when installing a new app


kkoo95:
# 🚀 Feature request

### Command

```
- [x] new
- [x] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
a new app or library generated with CLI 7.3.1 but installed with pnpm doesn't compile.
```
$ ng new test-pnpm --skip-install
$ cd test-pnpm
$ pnpm i
$ ng serve
```
result: 
```
** Angular Live Development Server is listening on localhost:4200, open your browser on http://localhost:4200/ **

Date: 2019-02-14T21:12:54.636Z
Hash: e8745ed29e7097ee7e23
Time: 15155ms
chunk {es2015-polyfills} es2015-polyfills.js, es2015-polyfills.js.map (es2015-polyfills) 368 kB [initial] [rendered]
chunk {main} main.js, main.js.map (main) 9.35 kB [initial] [rendered]
chunk {polyfills} polyfills.js, polyfills.js.map (polyfills) 262 kB [initial] [rendered]
chunk {runtime} runtime.js, runtime.js.map (runtime) 6.08 kB [entry] [rendered]
chunk {styles} styles.js, styles.js.map (styles) 514 bytes [initial] [rendered]
chunk {vendor} vendor.js, vendor.js.map (vendor) 3.56 MB [initial] [rendered]

ERROR in (webpack)/hot/emitter.js
Module not found: Error: Can't resolve 'events' in 'D:\dev\clones\test-pnpm\node_modules\.registry.npmjs.org\webpack\4.29.0\webpack@                                                  4.29.0\node_modules\webpack\hot'
ERROR in ./src/app/app.component.ts
Module not found: Error: Can't resolve 'postcss-loader' in 'D:\dev\clones\test-pnpm'
ERROR in ./src/app/app.component.ts
Module not found: Error: Can't resolve 'raw-loader' in 'D:\dev\clones\test-pnpm'
ERROR in multi ./src/styles.css
Module not found: Error: Can't resolve 'style-loader' in 'D:\dev\clones\test-pnpm'
i ｢wdm｣: Failed to compile.
```
From what i read, it would have to do with how angular is declaring its dependencies in the generated package.json. 

### Describe alternatives you've considered
[...]

PR #13687, opened 2019-02-16
fix(@angular-devkit/build-angular): improve webpack loader resolution

clydin:
Previously, all loaders either needed to be in the workspace's node modules directory or a node modules directory directly within the build angular package.  A package manager can potentially hoist a loader to a node modules location inbetween the two and cause loader resolution to fail.  This change causes webpack to check all intermediate node modules directories in addition to the initial two locations.

Partial fix for #13680


clydin:
Size check was unrelated to this PR.

PR #13691, opened 2019-02-17
feat(@schematics/angular): Update rxjs to 6.4.0

kyliau:
6.4.0 is needed so that Bazel does not have to build rxjs from source

See https://github.com/angular/angular/pull/28720


alan-agius4:
Can we make this as a `feature` and target only `master`, as it seems to me that it's a feature to scafold new projects with rxjs 6.4 instead of 6.3?

kyliau:
> Can we make this as a `feature` and target only `master`, as it seems to me that it's a feature to scafold new projects with rxjs 6.4 instead of 6.3?

Yes, that seems like a better approach. Thanks!

alan-agius4:
Commit message type should be `feat`

Issue #8138, opened 2017-10-21
tsconfig paths and module resolution errors in VS Code


SpiraMira:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Versions.

> @angular/cli: 1.4.9
node: 7.9.0
os: darwin x64
@angular/animations: 4.4.5
@angular/common: 4.4.5
@angular/compiler: 4.4.5
@angular/core: 4.4.5
@angular/forms: 4.4.5
@angular/http: 4.4.5
@angular/platform-browser: 4.4.5
@angular/platform-browser-dynamic: 4.4.5
@angular/router: 4.4.5
@angular/cli: 1.4.9
@angular/compiler-cli: 4.4.5
typescript: 2.3.4

macOS Sierra 10.12.6
VS Code 1.17.2
### Repro steps.

My tsconfig path property seems broken.
When I replace my path mapped imports (@shared, @assets etc.) with the equivalent coded relative paths, typescript module resolution seems to work (vs code intellisense is happy)
I have a large angular 2 project and am used to aliasing my asset and shared directories this way.
NOTE: project builds and (seems) to run just fine
This has been reported often, but I can't make sense of the current state of this issue.
This more than annoying as VS Code reports Angular spurious errors on all of my components.
I also suspect this is why I get the infamous Can't read property isSkipSelf of null when I open my templates (html) in VS Code.
I was running 1.4.2 and just tried ~1.4.2, which resolves to 1.4.9 in my package.json, to no avail.
Will 1.5.x fix this?
How can I help debug this without providing a copy of my app?

Thanks in advance. Regards.

filipesilva:
Can you setup a minimal repro please? You can read [here](https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#-submitting-an-issue) why this is needed. A good way to make a minimal repro is to create a new app via `ng new repro-app` and adding the minimum possible code to show the problem. Then you can push this repository to github and link it here.

Veabers:
I am not the OP, but here is a sample repo showing a component imported via the path alias displaying the "Cannot find module" error in vscode, but successfully building with npm run build --prod.

https://github.com/Veabers/tsconfig-paths-repro-app

sandangel:
vscode seem to only look into nearest tsconfig.json file, while angular read your tsconfig.app.json You should add relative path config to both tsconfig.json and tsconfig.app.json.

[...]

PR #13700, opened 2019-02-18
Web Worker support

filipesilva:
Followup to https://github.com/angular/angular-cli/pull/12575

Todo:
- [x] base implementation
- [x] tests for base implementation
- [x] typescript support
  - [x] make POC
  - [x] test rebuilds work
  - [x] figure out tsconfig setup
  - [x] keep the worker AngularCompilerPlugin from looking for lazy routes and other Angular-specific functionality
- [x] decide on flag names
- [x] documentation (to be added to AIO via https://github.com/angular/angular/pull/29633)
- [x] fix/workaround warning in https://github.com/GoogleChromeLabs/worker-plugin/issues/20
- [x] sketch out future interaction with `@angular-devkit/platform-worker`

Close https://github.com/angular/angular-cli/issues/5885


packages/angular/cli/lib/config/schema.json
```diff
@@ -835,6 +835,11 @@
             "ngswConfigPath": {
               "type": "string",
               "description": "Path to ngsw-config.json."
+            },  
+            "autoBundleWorkerModules": {
```
alan-agius4: Maybe we can change the name? Not very keen on this one.

filipesilva: Igor also wasn't keen on the name, so I image we will change it.

We also need an option for the worker tsconfig.

Serginho:
@filipesilva I'm following this feature closely and I can see worker plugin won't support lazy loading.

Why don't you bundle the workers using webpack multiple targets?
`export.module = [{target: 'web', ....}, {target: 'webworker' ...}]`

Or maybe you are not doing this thinking in platform-webworker.

filipesilva:
@Serginho at the moment I'm just following the base functionality of https://github.com/googlechromelabs/worker-plugin. It only supports `{ type:'module' }`. The functionality as is right now isn't meant to support `@angular/platform-webworker`, but rather just to bundle web workers.

packages/angular/cli/lib/config/schema.json
developit: Suggestion: `bundleModuleWorkers`

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -128,6 +129,11 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
     });
   }
 
+  if (buildOptions.autoBundleWorkerModules) {
+    const workerPluginInstance = new WorkerPlugin();
```
developit: As per my reply to googlechromelabs/worker-plugin#20, you can now pass an option to suppress the `globalObject` warning:

```suggestion
    const workerPluginInstance = new WorkerPlugin({
      globalObject: false
    });
```

filipesilva: Following our conversation in https://github.com/GoogleChromeLabs/worker-plugin/issues/20, I think `false` really is the best option for our case since we don't emphasize HMR in Angular CLI. We'll probably revisit this setting in the future.

developit: Yup! I figured that'd be the case. Cheers.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/worker.ts
```diff
@@ -0,0 +1,29 @@
...
+
+
+export function getWorkerConfig(wco: WebpackConfigOptions): Configuration {
+  const { buildOptions } = wco;
+  const workerTsConfigPath = buildOptions.workerTsConfig
+    ? resolve(wco.root, buildOptions.workerTsConfig)
+    : undefined;
+
+  return {
+    plugins: [new WorkerPlugin({
```
filipesilva: The child compilation where this plugin runs doesn't seem to run the environment hook.

This messes up virtual file system creation for AngularCompilerPlugin and breaks compilation.

filipesilva: Fixed by making the webpack plugin in child compilations pick up the parent WebpackCompilerHost from the VirtualFileSystemDecorator.

packages/angular/cli/lib/config/schema.json
```diff
@@ -836,6 +836,11 @@
...
```
hansl: This doesn't seem to be used.

filipesilva: Good catch 👍

packages/schematics/angular/web-worker/index.ts
```diff
@@ -0,0 +1,175 @@
...
+        if (filesAstNode.value.indexOf(workerGlob) == -1) {
+          appendValueInAstArray(recorder, filesAstNode, workerGlob);
+        }
+      }
+
+      host.commitUpdate(recorder);
+    }
+
+    // Add webworker lib to root tsconfig for editor support
+    // TODO
```
filipesilva: This needs to be done before the PR is finished.

packages/schematics/angular/web-worker/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    // The dom and webworker libs are mutually exclusive.
+    // Overwrite the lib option for all known tsconfigs to prevent errors.
+    // TODO
```
filipesilva: This needs to be done before the PR is finished.

packages/schematics/angular/web-worker/index.ts
```diff
@@ -0,0 +1,172 @@
...
+      return;
+    }
+
+    config.experimentalWebWorkerTsConfig =
+      `${root.endsWith('/') ? root : root + '/'}tsconfig.worker.json`;
+
+    // Add project tsconfig.json and tsconfig.worker.json.
+    // The project level tsconfig.json with webworker lib is for editor support since
+    // the dom and webworker libs are mutually exclusive.
+    // Note: this schematic does not change other tsconfigs to use the project-level tsconfig.
```
filipesilva: This schematic makes a two new tsconfig files:
- `path/to/project/tsconfig.worker.json` for the worker TS compilation itself
- `path/to/project/tsconfig.json` for the editor to support both DOM and WebWorker typings inside the project

As is I did not make the remaining tsconfigs in `path/to/project/` extend `path/to/project/tsconfig.json` because it's hard to ensure they are the default tsconfigs we generate on a new project.

But I believe we should eventually transition into having a `path/to/project/tsconfig.json` anyway in all projects because of other editor support issues like https://github.com/angular/angular-cli/issues/8138#issuecomment-458189246.

packages/angular_devkit/build_angular/test/browser/web-worker_spec_large.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    runTargetSpec(host, browserTargetSpec, overrides, DefaultTimeout * 3).pipe(
+      // Note(filipesilva): Wait for files to be written to disk... and something else?
+      // 1s should be enough for files to be written to disk.
+      // However, with a 1s to 3s delay, I sometimes (roughly 1 in 3 tests) saw TS compilation
+      // succeeding and emitting updated content, but the TS loader never called for
+      // 'src/worker/dep.ts'.
+      // But increasing this delay to 5s lead to no failed test in over 40 runs locally.
+      // In CI I still saw a rebuild fail with 5s, so I increased it to 10s.
+      // I think there might be a race condition related to child compilers somewhere in webpack.
```
filipesilva: Still saw a failure with the 10s debounce time in https://circleci.com/gh/angular/angular-cli/43023#tests/containers/1. Need to debug further.

owenhaynes:
Will this also enable sheared workers as well?

filipesilva:
@owenhaynes I'm not sure what you mean, can you give an example please?

owenhaynes:
https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker.

I think its possible using this branch, but am struggling to get the complied worker code to be generated when building.

So I assume i have yarn linked the repo wrong, just followed the instructions on the read me.

Also have encountered the following problems, but have temporary fixed them. 
Typescript does not have shared worker in the standard library. The typing @types/sharedworker is out of data and does not support the options parameter. This is easy to fix, have a local fixed copy which add the options to make the compiler happy.

filipesilva:
If https://github.com/GoogleChromeLabs/worker-plugin supports `SharedWorker`, then it should work here. But I don't see anything in the README or issue list about it. You could open a issue there for it, and when it supports it we can pull it in.

owenhaynes:
Will create a fork of it tomorrow and see if I can fix it to support shared workers, will report back.

owenhaynes:
Yes works with sharedworker. Just need get my JS skills up to scratch to give a PR back to https://github.com/GoogleChromeLabs/worker-plugin

filipesilva:
Awesome!

packages/angular_devkit/build_angular/test/browser/web-worker_spec_large.ts
developit: @filipesilva Is it possible the child compiler needs additional plugins? If it does and their configuration is the same, adding a string match pattern to the [`plugins[]` option](https://github.com/GoogleChromeLabs/worker-plugin#plugins) in worker-plugin would copy them over.

I'm not sure where to look in this repo to peek at how file watching is being done, any tips appreciated!

filipesilva: I'm looking at this currently. I'm not 100% sure of what the problem is but I've managed to reproduce it locally. I think it's related to the shared bits of our TS compilation pipeline.

packages/angular_devkit/build_angular/test/browser/web-worker_spec_large.ts
```diff
@@ -0,0 +1,179 @@
...
+            // Original worker content should be there.
+            workerContent = virtualFs.fileBufferToString(host.scopedSync().read(workerPath));
+            expect(workerContent).toContain('bar');
+            // Change content of worker dependency.
+            host.writeMultipleFiles({ 'src/app/dep.ts': `export const foo = 'baz';` });
+            phase = 2;
+            break;
+
+          case 2:
+            // TODO(filipesilva): Should this change? Check with Jason Miller.
```
filipesilva: @developit figured out the rebuild issue. It's because the worker changes name on rebuilds. Is this intended?

packages/angular_devkit/build_angular/test/browser/web-worker_spec_large.ts
filipesilva: I think I figured it out, let's continue the conversation in the other comment (https://github.com/angular/angular-cli/pull/13700#discussion_r268188296).

packages/angular_devkit/build_angular/test/browser/web-worker_spec_large.ts
```diff
@@ -0,0 +1,179 @@
...
+ */
+
+import { DefaultTimeout, TestLogger, runTargetSpec } from '@angular-devkit/architect/testing';
+import { join, virtualFs } from '@angular-devkit/core';
+import { debounceTime, takeWhile, tap } from 'rxjs/operators';
+import { browserTargetSpec, host, outputPath } from '../utils';
+
+
+describe('Browser Builder Web Worker support', () => {
+  beforeEach(done => host.initialize().toPromise().then(done, done.fail));
```
filipesilva: This spec large should be using the new Architect API instead.

packages/angular_devkit/build_angular/test/browser/web-worker_spec_large.ts
filipesilva: I'm also seeing a weird behaviour where sometimes the child compilation picks up on changed files, and sometimes it doesn't.

In this test I have seen all these outputs just by running the rebuild test repeatedly with no code changes:
- webpack loader is called for `src/app/dep.ts` twice and registers the change from `bar` to `baz`, `baz` ends up in `0.worker.js`
- same as above but `baz` ends up in `1.worker.js`
- loader is only called once (initially), neither `0.worker.js` nor `1.worker.js` contain `baz`

Do you have any idea of what's happening here?

developit: I think the worker changing names may be somewhat unavoidable. There's a `filename` option that would make it possible to use a static name, but that doesn't seem like the right solution here either.

There's [no caching enabled](https://github.com/GoogleChromeLabs/worker-plugin/blob/master/src/loader.js#L27) in worker-plugin, so it's likely something deeper in Webpack.

I'm going to add some rebuild tests to the plugin to see if I can trigger this in isolation. My hunch is that the [parser hook's instance counter](https://github.com/GoogleChromeLabs/worker-plugin/blob/7ea61fb83a02170d0235e4590c4deb617874e70a/src/index.js#L35) isn't being reset because the parser is reused across webpack compilations. It would certainly explain the `0` --> `1` bug there.

filipesilva: Been trying to debug further and have some findings. 

In our test suite we run webpack several times in the same thread. I think the test order affects the result of rebuilds with `worker-plugin`.

To test this I added a lot of debug logs and tried running the basic test and the rebuild test in different orders. Then I collected the logs just from the rebuild test.

One of the things I logged was the entry for the child compilation: https://github.com/GoogleChromeLabs/worker-plugin/blob/7ea61fb83a02170d0235e4590c4deb617874e70a/src/loader.js#L77. The others were build count, changed file being loaded, and which bundle contains the changed content.

If I run the rebuild test first and the basic test second I see:
- `dep.ts` is loaded with `bar`
- child compilation finishes for entry `0.worker.js`
- first build finishes
- `dep.ts` is loaded with `baz`
- child compilation finishes for entry `1.worker.js`
- second build finishes
- `0.worker.js` does not contains `baz`
- `1.worker.js` contains `baz`

If I run the rebuild test first and the basic test second I see:
- `dep.ts` is loaded with `bar`
- child compilation finishes for entry `0.worker.js`
- child compilation finishes for entry `1.worker.js`
- first build finishes
- `dep.ts` is loaded with `baz`
- child compilation finishes for entry `0.worker.js`
- second build finishes
- `0.worker.js` contains `baz`
- `1.worker.js` does not contains `baz`

So the end result is that the order of the tests seems to affect which bundle ends up containing the changed code.

But more interestingly, when the rebuild test is second, before the first build finishes there are a total of two child compilations that run. That part I find rather odd. I don't understand why that would happen. Even weirder is that the next child compilation is again the `0` one.

packages/angular_devkit/build_angular/test/browser/web-worker_spec_large.ts
filipesilva: Done

filipesilva:
@clydin @alan-agius4 PTAL. Aside from the rebuild issue that @developit is looking at, I think everything else is done. Documentation should be added to AIO after this PR.

packages/schematics/angular/web-worker/index.ts
```diff
@@ -0,0 +1,205 @@
...
+        if (optsAstNode && optsAstNode.kind != 'object') {
+          throw new SchematicsException(
+            'Invalid tsconfig "compilerOptions" property; Was expecting an object.');
+        }
+        const libAstNode = findPropertyInAstObject(tsCfgAst, 'lib');
+        if (libAstNode && libAstNode.kind != 'array') {
+          throw new SchematicsException('Invalid tsconfig "lib" property; expected an array.');
+        }
+        const newLibProp = 'webworker';
+        if (libAstNode && libAstNode.value.indexOf(newLibProp) == -1) {
```
alan-agius4: Nit: `libAstNode.value.indexOf(newLibProp) == -1` -> `!libAstNode.value.includes(newLibProp)`

packages/schematics/angular/web-worker/index.ts
```diff
@@ -0,0 +1,205 @@
...
+    options.name = parsedPath.name;
+    options.path = parsedPath.path;
+    const root = project.root || project.sourceRoot || '';
+
+    const templateSource = apply(url('./files/worker'), [
+      applyTemplates({ ...options, ...strings }),
+      move(parsedPath.path),
+    ]);
+
+    context.addTask(new NodePackageInstallTask());
```
alan-agius4: I don't see any reason why we need this, we are not adding new dependencies.

packages/schematics/angular/web-worker/index.ts
```diff
@@ -0,0 +1,205 @@
...
+  return target.options;
+}
+
+function addConfig(options: WebWorkerOptions, root: string): Rule {
+  return (host: Tree, context: SchematicContext) => {
+    context.logger.debug('updating project configuration.');
+    const workspace = getWorkspace(host);
+    const config = getProjectConfiguration(workspace, options);
+
+    if (!!config.webWorkerTsConfig) {
```
alan-agius4: `!!` is not really needed

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/worker.ts
```diff
@@ -0,0 +1,30 @@
...
+import { resolve } from 'path';
+import { Configuration } from 'webpack';
+import { WebpackConfigOptions } from '../build-options';
+import { getTypescriptWorkerPlugin } from './typescript';
+
+const WorkerPlugin = require('worker-plugin');
+
+
+export function getWorkerConfig(wco: WebpackConfigOptions): Configuration {
+  const { buildOptions } = wco;
```
alan-agius4: Maybe put the `const WorkerPlugin = require('worker-plugin');` in this method to require it only when needed.

packages/schematics/angular/web-worker/schema.d.ts
```diff
@@ -0,0 +1,30 @@
+/**
```
alan-agius4: This file should not be needed.

`schema.d.ts` are generated from `schema.json` during build time using quicktype.

packages/schematics/angular/web-worker/schema.json
```diff
@@ -0,0 +1,44 @@
...
+      "visible": false
+    },    
+    "project": {
+      "type": "string",
+      "description": "The name of the project.",
+      "$default": {
+        "$source": "projectName"
+      }
+    },
+    "target": {
```
alan-agius4: Is this needed? I don't see other supported targets

filipesilva: No, it is not now. At some point it was but not anymore. Removed.

packages/schematics/angular/web-worker/schema.d.ts
filipesilva: Removed

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/worker.ts
filipesilva: I agree with the perf argument but it goes against the general pattern we use for imports. I'd prefer to not go against the grain without a good reason, and this plugin is very small so I don't think there is much to be gained.

packages/schematics/angular/web-worker/index.ts
filipesilva: Fixed

packages/schematics/angular/web-worker/index.ts
filipesilva: Indeed it is not needed. Left it in as a bad copy paste. Thanks for finding this.

packages/schematics/angular/web-worker/index.ts
filipesilva: Fixed

packages/angular_devkit/build_angular/package.json
```diff
@@ -51,7 +51,8 @@
     "webpack-dev-server": "3.2.1",
     "webpack-merge": "4.2.1",
     "webpack-sources": "1.3.0",
-    "webpack-subresource-integrity": "1.1.0-rc.6"
+    "webpack-subresource-integrity": "1.1.0-rc.6",
+    "worker-plugin": "^3.1.0"
```
clydin: Pin version

filipesilva: Done

PR #13710, opened 2019-02-19
fix(@schematics/angular): add file extensions to style prompt

alan-agius4:
This is causing major confusion as users are not aware that choosing `sass` will generate files with `scss` extension.

Another reason for this confusion is that in other schematics, `style` accepts either a preprocessor or file extension while in this prompt only a preprocessor is provided.


alan-agius4:
@filipesilva & @hansl, all agree with this as well?

PR #13712, opened 2019-02-19
fix(@angular/cli): remove redundant typescript/compiler-cli compatibility check

clydin:
Angular 5.0+ has a full peer dependencies setup (with 6.0+ also having a configurable runtime error check) to ensure that an appropriate version of typescript is available for compilation.  Angular CLI 8.0+ does not support Angular versions prior to these and therefore the warning is redundant.  For the case where the developer wishes to use an unsupported TypeScript version, the developer would currently need to adjust two similar but differently named settings in two different configuration files.  With this change only one is needed.


packages/angular/cli/upgrade/version.ts
```diff
@@ -110,67 +109,4 @@ export class Version {
     }
   }
 
-  static assertTypescriptVersion(projectRoot: string) {
-    if (!isWarningEnabled('typescriptMismatch')) {
```
filipesilva: We should also deprecate, or wholly remove, this option. There are many files that reference it, including tests.

clydin: Cleaned up the option as well.

PR #13715, opened 2019-02-20
Architect browser schema

hansl:
refactor(@angular-devkit/build-angular): clean up some interfaces and schemas

Remove the manually maintained schema.d.ts from the browser builder, and use
the Schema JSON file to generate it. This had a lot of repercussions around
the whole build-angular code base and the different interfaces that were
manually kept.


packages/angular_devkit/build_angular/test/browser/source-map_spec_large.ts
```diff
@@ -120,8 +121,10 @@ describe('Browser Builder source map', () => {
   it('supports styles only sourcemaps', (done) => {
     const overrides: Partial<BrowserBuilderSchema> = {
       sourceMap: {
```
alan-agius4: why are these options required now?

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/browser.ts
```diff
@@ -67,9 +67,9 @@ export function getBrowserConfig(wco: WebpackConfigOptions) {
 
   if (!isEval && (scriptsSourceMap || stylesSourceMap)) {
     extraPlugins.push(getSourceMapDevTool(
-      scriptsSourceMap,
-      stylesSourceMap,
-      hiddenSourceMap,
+      scriptsSourceMap === undefined ? true : scriptsSourceMap,
```
alan-agius4: Can we use the normalizeSourceMapOptions method for this cases, setting these default in multiple places seems risky.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/server.ts
```diff
@@ -21,9 +21,9 @@ export function getServerConfig(wco: WebpackConfigOptions) {
     const { scripts, styles, hidden } = wco.buildOptions.sourceMap;
 
     extraPlugins.push(getSourceMapDevTool(
-      scripts,
-      styles,
-      hidden,
+      scripts === undefined ? true : scripts,
+      styles === undefined ? true : styles,
+      hidden === undefined ? true : hidden,
```
alan-agius4: When undefined` it's `true`? this is changing the behaviour

packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -105,7 +86,7 @@ export class DevServerBuilder implements Builder<DevServerBuilderOptions> {
         const serverAddress = url.format({
           protocol: options.ssl ? 'https' : 'http',
           hostname: options.host === '0.0.0.0' ? 'localhost' : options.host,
-          port: options.port.toString(),
+          port: (options.port || 4200).toString(),
```
alan-agius4: Why not use the defaults in the schema?

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/browser.ts
hansl: We should, or at least add MORE normalizing methods.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/server.ts
hansl: Current behavior is to default to true, no? The schema says it's true by default, the type says it's optional.

packages/angular_devkit/build_angular/src/dev-server/index.ts
hansl: The type is optional, so typescript complains here. I need to figure out how to better do this.

packages/angular_devkit/build_angular/test/browser/source-map_spec_large.ts
hansl: They're not, I just went back and forth between required and not required.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/server.ts
alan-agius4: It’s true for souceMap but not when it’s a complex object, when it is a complex object at the moment, hidden is also false.

PR #13716, opened 2019-02-20
fix(@schematics/angular): fix missing semicolons in extensions.json.template

cipchk:
Fixes #13716


cipchk:
I signed it!

kyliau:
@alan-agius4 this `.vscode` directory is not available in the patch branch. Should this PR target master only?

alan-agius4:
@kyliau, yes please

PR #13720, opened 2019-02-20
fix(@schematics/angular): `findModuleFromOptions` not handling proper…

### LINKED ISSUE #13714: Newly created component not being added to submodule
# 🐞 Bug report

```
- [X] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

 Not sure

#### Expected behaviour:
When a new component is generated using `ng g c someModule/newComponent`, it should be added to sub module.

#### Actual behaviour:
Newly generated component is added to root module instead of in sub module.

#### Scenario:
If i have a submodule called DeleteModule
app ===> delete ===> delete.module.ts

and if generate a component using` ng g c delete/newComponent`. the newly generated components is added to DeleteModule. However if my DeleteModule's structure is
app => FeatureModule ===> delete ===> delete.module.ts

and i try to generate new component using `ng g c FeatureModule/delete/someComponent`, thee newly generated component is added to root module instead of in sub module.

## 🔬 Minimal Reproduction
1. Create a new angular project using ng new app.
2. Create a sub module using `ng g m FeatureModule/delete`.
3. Create a component using `ng g c FeatureModule/delete/newComponent`.

## 🔥 Exception or Error
None
</code></pre>

## 🌍 Your Environment
<pre><code>
Defaults
</code></pre>

**Anything else relevant?**
None

alan-agius4:
…ly different casing in name

At the moment users can have various casing and separators in paths, we should not always dasherize the name when resolving modules.

As for example when providing something like:

```
/module/SubModule/feature
```

It won't be able to resolve the modules properly as `sub-module` does't exist.

This PR also updates the test for underscore as previously it was not properly testing this usercase, since the formatter was used on name and not th path.

Fixes #13714

### DIFF
--- a/packages/schematics/angular/utility/find-module.ts
+++ b/packages/schematics/angular/utility/find-module.ts
@@ -12,7 +12,6 @@ import {
   join,
   normalize,
   relative,
-  strings,
 } from '@angular-devkit/core';
 import { DirEntry, Tree } from '@angular-devkit/schematics';
 
@@ -25,7 +24,6 @@ export interface ModuleOptions {
   skipImport?: boolean;
   moduleExt?: string;
   routingModuleExt?: string;
-  nameFormatter?: (str: string) => string;
 }
 
 const MODULE_EXT = '.module.ts';
@@ -43,8 +41,7 @@ export function findModuleFromOptions(host: Tree, options: ModuleOptions): Path 
   const routingModuleExt = options.routingModuleExt || ROUTING_MODULE_EXT;
 
   if (!options.module) {
-    options.nameFormatter = options.nameFormatter || strings.dasherize;
-    const pathToCheck = (options.path || '') + '/' + options.nameFormatter(options.name);
+    const pathToCheck = (options.path || '') + '/' + options.name;
 
     return normalize(findModule(host, pathToCheck, moduleExt, routingModuleExt));
   } else {
--- a/packages/schematics/angular/utility/find-module_spec.ts
+++ b/packages/schematics/angular/utility/find-module_spec.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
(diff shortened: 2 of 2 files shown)


packages/schematics/angular/utility/find-module_spec.ts
```diff
@@ -111,12 +111,20 @@ describe('find-module', () => {
       expect(modPath).toEqual('/projects/my-proj/src/app.module.ts' as Path);
     });
 
-    it('should find a module if nameFormatter is provided', () => {
```
alan-agius4: this test was actually not testing anything, as `nameFormatter ` was only used when `name` is provided

packages/schematics/angular/utility/find-module.ts
```diff
@@ -25,6 +24,7 @@ export interface ModuleOptions {
   skipImport?: boolean;
   moduleExt?: string;
   routingModuleExt?: string;
+  /** @deprecated This option is no longer required and will be removed in version 9 */
```
clydin: These are all internal.  The deprecation isn’t necessary here.

alan-agius4: The thing is that this is used within google so while it’s used internally if we remove it it will break inside google.

clydin: That would actually be more of a reason to completely remove it since you would then get a compile time error instead of differing runtime behavior; which may not be initially noticed.

alexeagle: Removing it is okay too, but @kyliau should probably try a g3sync to see how difficult it will be to keep the sync going.
We haven't synced it in a while IIRC so it might already be hard to do...

alan-agius4: Removed

PR #13726, opened 2019-02-20
update default packages

hpigula:


alan-agius4:
Hi, the dependencies added here are not used internally within the CLI and hence should not be added.

Thanks

PR #13727, opened 2019-02-21
build-webpack New Builders

hansl:
Moved the build-webpack builders to the new API. Also some QoL changes around.


packages/angular_devkit/build_webpack/builders.json
```diff
@@ -2,11 +2,19 @@
   "$schema": "../architect/src/builders-schema.json",
   "builders": {
     "webpack": {
+      "implementation": "./src/webpack/index2",
       "class": "./src/webpack",
       "schema": "./src/webpack/schema.json",
       "description": "Build a webpack app."
     },
+    "dev-server": {
```
alan-agius4: why 2 builders that do the same thing?

packages/angular_devkit/build_webpack/src/plugins/architect.ts
```diff
@@ -0,0 +1,157 @@
...
+  'optimizeExtractedChunksBasic',
+  'optimizeExtractedChunks',
+  'optimizeExtractedChunksAdvanced',
+  'afterOptimizeExtractedChunks',
+];
+
+/**
+ * A webpack plugin that reports status and progress to Architect.
+ */
+export class ArchitectPlugin {
```
alan-agius4: <3

packages/angular_devkit/build_webpack/builders.json
hansl: I was thinking of removing the other one. `build-webpack:webpack-dev-server` is redundant. Same for `build-webpack:webpack`, was thinking of renaming it `build`.

packages/angular_devkit/build_webpack/src/plugins/architect.ts
hansl: Yup. It's used all around, makes code way cleaner.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2.ts
```diff
@@ -0,0 +1,101 @@
...
+        log(stats, config);
+
+        obs.next({
+          success: !stats.hasErrors(),
+          ...result,
+        } as DevServerBuildResult);
+      });
+
+      server.listen(
+        devServerConfig.port === undefined ? 8080 : devServerConfig.port,
```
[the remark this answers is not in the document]
hansl: Not for this builder: https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_webpack/src/webpack-dev-server/index.ts#L57

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2.ts
```diff
@@ -0,0 +1,101 @@
...
+        devServerConfig.host === undefined ? 'localhost' : devServerConfig.host,
+        function (this: net.Server, err) {
+          if (err) {
+            obs.error(err);
+          } else {
+            result = {
```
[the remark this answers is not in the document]
hansl: This was the logic of the previous builder.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2.ts
```diff
@@ -0,0 +1,101 @@
...
+      const server = new WebpackDevServer(webpackCompiler, devServerConfig);
+      let result: DevServerBuildResult;
+
+      webpackCompiler.hooks.done.tap('build-webpack', (stats) => {
+        // Log stats.
+        log(stats, config);
+
+        obs.next({
+          success: !stats.hasErrors(),
+          ...result,
```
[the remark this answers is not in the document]
hansl: Good point. Fixed.

packages/angular_devkit/build_webpack/src/webpack/index2.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export interface WebpackLoggingCallback {
+  (stats: webpack.Stats, config: webpack.Configuration): void;
+}
+export interface WebpackFactory {
+  (config: webpack.Configuration): Observable<webpack.Compiler>;
+}
+
+
+export function runWebpack(
```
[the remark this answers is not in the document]
hansl: The index2_spec_large.ts are the tests moved from the previous builder. This is as much of an entry point as any public/protected class members from the previous builder.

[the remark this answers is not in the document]
hansl: Just as a FYI, I'm planning to do a coverage at one point before release, and add tests for missing pieces where it makes sense.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2.ts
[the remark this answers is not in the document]
hansl: It seems that this callback is called before the compilation starts, so that would be the case (the done hook is called after the server is listening).

packages/angular_devkit/build_webpack/builders.json
alan-agius4: I think it makes sense

packages/angular_devkit/build_webpack/builders.json
```diff
@@ -1,12 +1,26 @@
 {
   "$schema": "../architect/src/builders-schema.json",
   "builders": {
+    "build": {
+      "implementation": "./src/webpack/index2",
```
filipesilva: Is the idea here that transitioning to the new Architect API means adding the `implementation` field to existing builders, keeping both current and future version at the same time, then remove `class`/`schema` when ready to changeover?

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2.ts
```diff
@@ -0,0 +1,101 @@
...
+    webpackFactory?: WebpackFactory,
+  } = {},
+): Observable<BuilderOutput> {
+  const createWebpack = options.webpackFactory || (config => of(webpack(config)));
+  const log: WebpackLoggingCallback = options.logging
+    || ((stats, config) => context.logger.info(stats.toString(config.stats)));
+
+  config = webpackMerge(config, {
+    plugins: [
+      new ArchitectPlugin(context),
```
filipesilva: Should this be added if progress reporting is not needed?

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2_spec_large.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  it('works', async () => {
+    const run = await architect.scheduleTarget(webpackTargetSpec);
+    const output = await run.result as DevServerBuildResult;
+    expect(output.success).toBe(true);
+
+    const response = await fetch(`http://${output.address}:${output.port}/bundle.js`);
+    expect(await response.text()).toContain(`console.log('hello world')`);
+
+    await run.stop();
```
filipesilva: Would be nice to verify that it has stopped too, by trying to fetch and expecting it to fail.

packages/angular_devkit/build_webpack/src/webpack/index2.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+      };
+
+      try {
+        if (config.watch) {
+          const watchOptions = config.watchOptions || {};
+          const watching = webpackCompiler.watch(watchOptions, callback);
+
+          // Teardown logic. Close the watcher when unsubscribed from.
+          return () => watching.close(() => { });
```
filipesilva: Is this working? Would be nice to test it.

packages/angular_devkit/build_webpack/src/webpack/index2_spec_large.ts
```diff
@@ -0,0 +1,76 @@
...
+      await createArchitect(workspaceRoot);
+    });
+
+    it('works', async () => {
+      const run = await architect.scheduleTarget({ project: 'app', target: 'build' });
+      const output = await run.result;
+
+      expect(output.success).toBe(true);
+      expect(await vfHost.exists(join(outputPath, 'bundle.js')).toPromise()).toBe(true);
+      await run.stop();
```
filipesilva: It is always necessary to stop manually?

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2.ts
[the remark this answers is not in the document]
clydin: Seems fragile but as long as webpack maintains the ordering, all good.

packages/angular_devkit/build_webpack/src/webpack/index2.ts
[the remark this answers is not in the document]
clydin: Since builders can no longer be directly augmented or have their behavior otherwise adjusted, utility functions such as this and the ones in the second PR are going to be used extensively in third-party builders.  I think extensive test coverage would be prudent in this case to not only ensure functionality but to also prevent future unintentional breaking changes.

packages/angular_devkit/build_webpack/builders.json
hansl: Yes, that's the general idea. Since we will remove the old API though, this will not show up past the beta period. The old API will be completely deleted before we move to RC.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2.ts
hansl: Progress reporting should always be needed ;)

And yes, we might want to add more info to it later on.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index2_spec_large.ts
hansl: The `stop()` returned Promise resolves when it's actually done (ie. the observable is completed).

packages/angular_devkit/build_webpack/src/webpack/index2.ts
hansl: I'll add some missing test coverage later on.

packages/angular_devkit/build_webpack/src/webpack/index2_spec_large.ts
hansl: Yes, as otherwise we might leak memory.

PR #13738, opened 2019-02-21
fix(@angular-devkit/build-angular): add promise polyfill to --es5BrowserSupport provided polyfills

### LINKED ISSUE #13735: es5BrowserSupport add Promise polyfill
# 🚀 Feature request

```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description

When using the new `--es5BrowserSupport` flag I noticed that **Promise** is not being included.

And I can see it's not part of the list here:
https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/angular-cli-files/models/es2015-polyfills.js

Hence, when I open my app in IE11 I get "Promise is undefined" error. I can fix this by adding `import 'core-js/es6/promise` in my `polyfills.ts` file, but I am thinking it should be part of the defaults of the new flag?

I can submit a PR - but could somebody confirm please this is a real issue and I'm not just missing something?

Add `import 'core-js/es6/promise;` to

https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/angular-cli-files/models/es2015-polyfills.js

Add `import 'core-js/es6/promise` in my `polyfills.ts` file.

rarmatei:
Since Angular uses Promises, we need to provide a polyfill for it to support ES5 only browsers.

Closes #13735

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/es2015-polyfills.js
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/es2015-polyfills.js
@@ -19,3 +19,4 @@ import 'core-js/es6/regexp';
 import 'core-js/es6/map';
 import 'core-js/es6/weak-map';
 import 'core-js/es6/set';
+import 'core-js/es6/promise';


clydin:
Can you update the commit message to start with “fix(@angular-devkit/build-angular):” instead of “build”?

rarmatei:
done! 👍

mattlewis92:
I don't think this should actually be needed as [zone.js provides its own Promise polyfill](https://github.com/angular/zone.js/blob/master/lib/common/promise.ts#L275-L447). I only know this because including both the polyfill from core-js and the one from zone.js broke our app in really weird ways: https://github.com/PatrickJS/angular-starter/issues/709 (That was a year ago though, so maybe that specific issue has been fixed now in either core-js or zone.js). Our app is using the new es2015 conditional polyfill loading and it loads fine in IE11 with no error caused by a missing promise polyfill, plus it was never included in the [original polyfills schematic](https://github.com/angular/angular-cli/blob/7.0.x/packages/schematics/angular/application/files/src/polyfills.ts#L21-L35).

PR #13690, opened 2019-02-16
fix(@schematics/angular): fixes issue that `viewEncapsulation` is not…

alan-agius4:
… set in `app.component.ts` and in `angular.json`

Fixes #13689

PR #13740, opened 2019-02-22
fix(@schematics/angular): remove extra comma in component schematic

cexbrayat:
This was introduced by #13690

Previously to this commit, a trailing comma was always added after styles/styleUrls
The `viewEncapsulation` also adds one, so there is an extra one.

Fixes: #13742


alan-agius4:
Thinking about this, we should probably remove ViewEncapsulation from this file. Inc the import.

Since now we set the defaultViewEncapsulation in main.ts

cexbrayat:
@alan-agius4 Sure. I updated the PR

PR #13748, opened 2019-02-23
Architect build browser

hansl:


hansl:
This should be rebased on top of master once https://github.com/angular/angular-cli/pull/13727 is in.

etc/api/angular_devkit/core/src/_golden-api.d.ts
```diff
@@ -1018,6 +1018,13 @@ export declare class Workspace {
     loadWorkspaceFromHost(workspacePath: Path): Observable<this>;
     loadWorkspaceFromJson(json: {}): Observable<this>;
     validateAgainstSchema<T = {}>(contentJson: {}, schemaJson: JsonObject): Observable<T>;
+    protected static _workspaceFile: string[];
```
filipesilva: `_workspaceFile` looks like it should be `_workspaceFiles` since it is an array.

packages/angular_devkit/core/src/workspace/workspace.ts
```diff
@@ -55,15 +63,69 @@ export class AmbiguousProjectPathException extends BaseException {
...
+  constructor(
+    private _root: Path,
+    private _host: virtualFs.Host<{}>,
+    registry?: schema.CoreSchemaRegistry,
+  ) {
+    if (registry) {
+      this._registry = registry;
+    } else {
+      this._registry = new schema.CoreSchemaRegistry();
+      this._registry.addPostTransform(schema.transforms.addUndefinedDefaults);
```
filipesilva: If `registry` is provided, does it need to already have this transform? 

If so, we should check and error out.

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -0,0 +1,249 @@
+/**
```
filipesilva: This builder really should be thoroughly tested as it's the most important one. 

Since the exposed functions are meant to be used by third party builders and thus constitute public API, they also need to be tested.

etc/api/angular_devkit/core/src/_golden-api.d.ts
hansl: Renamed to `_workspaceFileNames`.

packages/angular_devkit/core/src/workspace/workspace.ts
hansl: No. Honestly `Registry` should always be required. All the places we  build it and yet it's just best to always use the same instance for everything.

packages/angular_devkit/build_angular/src/browser/index2.ts
hansl: Yes I should move all the tests for the browser builder. I'll do that asap.

packages/angular_devkit/core/src/workspace/workspace.ts
hansl: The problem is that this is part of a public API so we can't break it.

clydin: It's marked as experimental.

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -0,0 +1,249 @@
...
+    workspace,
+    host,
+    context.logger,
+  );
+
+  return [projectRoot, config];
+}
+
+
+export function buildWebpackBrowser(options: BrowserBuilderSchema, context: BuilderContext) {
```
clydin: This should probably be a factory which accepts a function to modify the final webpack config.  Currently there is no way for a custom builder to use the browser builder logic without copying this function verbatim.

packages/angular_devkit/build_angular/src/browser/index2.ts
hansl: Although this is going well, there is a LOT of tests for this builder so it's taking me more time. The good news is that this is mostly manual work and there was only 1 bug found in architect so far. Should be done really soon.

packages/angular_devkit/build_angular/src/browser/index2.ts
hansl: Done. Please take another look.

PR #13754, opened 2019-02-25
fix(@schematics/angular): default newProjectRoot to current directory

clydin:
Fixes #13704

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -151,7 +151,7 @@ function addAppToWorkspaceFile(options: ApplicationOptions, workspace: Workspace
   // }
   let projectRoot = options.projectRoot !== undefined
     ? options.projectRoot
-    : `${workspace.newProjectRoot}/${options.name}`;
+    : `${workspace.newProjectRoot || ''}/${options.name}`;
   if (projectRoot !== '' && !projectRoot.endsWith('/')) {
     projectRoot += '/';
   }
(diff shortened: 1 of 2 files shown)


FrozenPandaz:
@clydin FYI, this does not fix the breaking change which we discussed. There is still logic which expects `newProjectRoot` to be (or not to be) `undefined`.

clydin:
`newProjectRoot` is only in a conditional which checks if it is truthy in the application schematic:
https://github.com/angular/angular-cli/blob/dad48c3af6bb2e3e94d7a29f0c70cf85400c1563/packages/schematics/angular/application/index.ts#L351

I updated the PR to maintain the existing behavior for that case.  Even though there is an argument to be made that it is actually a defect.  As is, the 'e2e' directory would overlap an existing 'e2e' folder at the workspace root and the schematic would fail in the case of the workspace's newProjectRoot being falsy.

FrozenPandaz:
Awesome! Those changes fix the breaking change.

Thank you so much!

packages/schematics/angular/application/index.ts
```diff
@@ -326,7 +326,7 @@ export default function (options: ApplicationOptions): Rule {
       };
 
     const workspace = getWorkspace(host);
-    let newProjectRoot = workspace.newProjectRoot || 'projects';
+    let newProjectRoot = workspace.newProjectRoot || '';
```
hansl: Isn't that a breaking change? This seems to be 2 changes; a fix and a feature...

PR #13775, opened 2019-02-26
Integrate prettier code formatter

clydin:


hansl:
Considering none of my code got reformatted, I consider myself my own prettier 🤪

I personally just don't want to end up worse than without a formatter. Thinking about some of the past choices made by formatters ending up unreadable... (trying to find an example in the archives)

clydin:
It's now setup to be fully incremental.  Only modified lines in a commit will be formatted.  Existing code will stay as-is until changed by a commit.

mgechev:
@hansl I was using prettier for the past over a year. I haven't had any complains, it should be safe to introduce it in the CLI.

PR #13776, opened 2019-02-26
feat(@angular/cli): recommend plugins that John Papa recommends

alexeagle:
This list comes from https://marketplace.visualstudio.com/items?itemName=johnpapa.angular-essentials
But we don't write a .prettier.json config file so we don't include the Prettier plugin yet

Suggested by @StephenFluin

### DIFF
--- a/packages/schematics/angular/migrations/update-vscode-recommendations/update-8.0.0.ts
+++ b/packages/schematics/angular/migrations/update-vscode-recommendations/update-8.0.0.ts
@@ -18,7 +18,13 @@ const addExtensionRecommendations = updateJsonInTree(
     [
       'angular.ng-template',
       'nrwl.angular-console',
-      'ms-vscode.vscode-typescript-tslint-plugin'].forEach(extension => {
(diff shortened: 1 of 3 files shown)


mgechev:
[This](https://marketplace.visualstudio.com/items?itemName=Mikael.Angular-BeastCode) set of snippets has over 8m downloads.

We should probably evaluate both options.

alexeagle:
Switched to `Mikael.Angular-BeastCode` snippets, sorry @johnpapa :)

alexeagle:
@mgechev can we merge this, or would you like to discuss more?

johnpapa:
that's fine ... but check the *installs* not download. VS Code shows both. Downloads count for every update. Installs are current installations.

alexeagle:
Sorry @johnpapa we decided it was premature to merge this without nailing down what is our criteria and how we measure it. We'll revert and try again :)

johnpapa:
No worries - I'm cool with whatever you choose. I just wanted you to know about the metrics since it sounded like that was a factor.

If you dont choose mine, we're still friends :)

PR #13777, opened 2019-02-27
fix(@schematics/angular): reintroduce `.sass` as a supported file ext…

### LINKED ISSUE #13739: Revert 'fix(@schematics/angular): stop offering SASS option for ng new #13444'
I initially thought this was a **bug**!
`ng g c component-name --style=sass` generates  .scss files
`ng new my-project --style=sass` generates .scss files
What is the reason behind this change? 
What about those of us who prefer using the sass indented syntax? It is still supported and they are not dropping it so I don't see any reason to remove it from ng generate and ng new simply because most developers prefer the scss syntax.
This is really causing a problem for my workflow as I am using `let g:angular_cli_stylesheet_format = 'sass'` in my .vimrc so I now have to convert everything to scss and start working with syntax I don't like to use.

 https://github.com/angular/angular-cli/pull/13444

@andyg3

alan-agius4:
…ention

Sass Indented (.sass) is fully supported by the Sass team and we should still offer and support it.

Fixes #13739


packages/schematics/angular/ng-new/schema.json
```diff
@@ -125,10 +125,11 @@
         "message": "Which stylesheet format would you like to use?",
         "type": "list",
         "items": [
-          { "value": "css",  "label": "CSS    (.css )" },
-          { "value": "sass", "label": "Sass   (.scss) [ http://sass-lang.com   ]" },
-          { "value": "less", "label": "Less   (.less) [ http://lesscss.org     ]" },
-          { "value": "styl", "label": "Stylus (.styl) [ http://stylus-lang.com ]" }
+          { "value": "css",  "label": "CSS" },
+          { "value": "scss", "label": "SCSS   [ http://sass-lang.com   ]" },
```
alan-agius4: Note: Unlike Sass, SCSS is always written in all caps

clydin: `Sass SCSS`/`Sass Indented`?

filipesilva: Is the name `Sass Indented` or `Indented Sass`?

clydin: `Sass Indented`
https://sass-lang.com/documentation/file.INDENTED_SYNTAX.html

clydin: I suppose we could ask for clarification on full official naming.

alan-agius4: Sass's indented syntax (also known simply as "Sass”)

clydin: I'm fine with it either way.  But I think we just need a general consensus so this isn't revisited in another 2 weeks.

alan-agius4: In their docs to be clear they mostly use the syntax kind as references ie Sass and SCSS

alexeagle: I think the links here should go all the way to the docpage https://sass-lang.com/documentation/file.INDENTED_SYNTAX.html

Somehow users need to know more to pick the right thing from the list, most of them are probably not aware there are multiple syntax nor how to pick between them

AS much as we can disambiguate in the label, that's also good. Maybe

`SCSS (CSS syntax)  [http://sass-lang.com/documentation/file.SASS_REFERENCE.html#syntax]`
`Sass (intented syntax) [https://sass-lang.com/documentation/file.INDENTED_SYNTAX.html]`


either way I don't care to revisit this :)

alan-agius4: @alexeagle I think you meant `Sass (SCSS syntax)`?

I think in general and this was also how it was before, SCSS refers to the SCSS syntax and Sass refers to the indented syntax.

My point is that Sass & SCSS are the most commonly used terms, and for novice users putting other terms might cause more confusion then good.

In most cases, for example, in Sass documentation, when they refer to Sass, they refer to the indented syntax and not the pre-processor.

https://sass-lang.com/documentation/file.SASS_REFERENCE.html#syntax_selection
> SCSS syntax works straightforwardly in Sass by using newlines instead of semicolons and indentation instead of braces.

packages/schematics/angular/ng-new/schema.json
```diff
@@ -125,10 +125,11 @@
         "message": "Which stylesheet format would you like to use?",
         "type": "list",
         "items": [
-          { "value": "css",  "label": "CSS    (.css )" },
-          { "value": "sass", "label": "Sass   (.scss) [ http://sass-lang.com   ]" },
-          { "value": "less", "label": "Less   (.less) [ http://lesscss.org     ]" },
-          { "value": "styl", "label": "Stylus (.styl) [ http://stylus-lang.com ]" }
+          { "value": "css",  "label": "CSS" },
+          { "value": "scss", "label": "SCSS   [ http://sass-lang.com/documentation/file.SASS_REFERENCE.html#syntax ]" },
```
alexeagle: opt suggestion: add parenthesis saying `(css-style syntax)` `(indented syntax)`

alan-agius4: I am still not sure about the added value of the above tbh. And I fear more confusion with introduction of new terms.

PR #13780, opened 2019-02-27
feat(@schematics/angular): change layout of e2e files

### LINKED ISSUE #12326: Allow generation of apps with nested e2e projects instead of sibling e2e projects.
This is a feature request

### Desired functionality

When generating a new app inside of a workspace, I would like to be able to nest the respective e2e project inside the new app. 

Current structure: sibling e2e `(ng generate application app2)`
```
...
projects
    -app2
        -src
    -app2-e2e
```
Desired structure: nested e2e
```
...
projects
    -app2
        -src
        -e2e
```

Initially raised this as a feature request / PR in Nrwl's NX repo here: https://github.com/nrwl/nx/issues/743
But it was determined that this is the appropriate place to raise this feature request.

Related question on SO here: https://stackoverflow.com/questions/51763075/why-does-nrwl-nx-create-a-separate-e2e-project-for-each-app/

### Proposal
Add a new option to the `ng generate application` schematic named `nestE2e` (although naming things isn't exactly my specialty), which can be used to create a nested e2e project (as opposed to a sibling e2e project) for the newly generated app. 
```
ng generate application my-app --nestE2e
```

### Consideration
This would be helpful for working in a monorepo, or for working in an empty workspace (#12216). As more apps get generated, you wouldn't be left with 2x the number of folders cluttering (imo) the projects directory. 

The desired structure above is also more familiar to people who work on standalone apps generated with ng new, in that e2e code is kept within the app directory.

alan-agius4:
With this change E2E files will be relocated inside an existing application instead of creating a seperate E2E project. This will also remove a lot of extra boilerplating inside the workspace configuration file.

File layout:
```
│   browserslist
│   karma.conf.js
│   tsconfig.app.json
│   tsconfig.spec.json
│   tslint.json
│
├───e2e
│   │   protractor.conf.js
│   │   tsconfig.e2e.json
│   │
│   └───src
│           app.e2e-spec.ts
│           app.po.ts
│
└───src
    │   favicon.ico
    │   index.html
    │   main.po.ts
    │   main.ts
    │   polyfills.ts
    │   styles.css
    │   test.ts
    │
    ├───app
    │       app.component.css
    │       app.component.html
    │       app.component.spec.ts
    │       app.component.ts
    │       app.module.ts
    │
    ├───assets
    │       .gitkeep
    │
    └───environments
            environment.prod.ts
            environment.ts
```

Ref: TOOL-699

Closes #12326


clydin:
I think we should keep the E2E folder just move it into the `src`.  They are really run/compiled completely separate from the actual application unlike the unit tests which are compiled into the application.

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -48,7 +48,7 @@ export async function createProject(name: string, ...args: string[]) {
 }
 
 export async function prepareProjectForE2e(name) {
```
filipesilva: The `useCIChrome` function also needs to be altered to find the protractor config.  I think it works right now because it's using the system Chrome instead.

alan-agius4: Shouldn't this already be covered by:
https://github.com/angular/angular-cli/blob/9f985eefac84fffa6d942efa1b1cf8c64c1a4790/tests/legacy-cli/e2e/utils/project.ts#L57

and 

https://github.com/angular/angular-cli/blob/9f985eefac84fffa6d942efa1b1cf8c64c1a4790/tests/legacy-cli/e2e/utils/project.ts#L226

filipesilva: According to the PR message, the config will be at `src/e2e/protractor.conf.js`. But the projectDir itself is at `src/` because there is no separate e2e project. So `src/protractor.conf.js` does not exist. Am I missing something here?

alan-agius4: @filipesilva according to the PR message the `protractor.conf.js` is located in `e2e/protractor.conf.js`

```
├───e2e
│   │   protractor.conf.js
│   │   tsconfig.e2e.json
│   │
│   └───src
│           app.e2e-spec.ts
│           app.po.ts
│
└───src
    │   favicon.ico
    │   index.html
    │   main.po.ts
    │   main.ts
    │   polyfills.ts
    │   styles.css
    │   test.ts
```

filipesilva: Discussed offline, I was misunderstanding the changes.

PR #13787, opened 2019-02-28
fix(@angular-devkit/architect-cli): fix paths to support Windows

alan-agius4:


packages/angular_devkit/architect_cli/bin/architect.ts
```diff
@@ -207,15 +207,15 @@ async function main(args: string[]): Promise<number> {
     return 3;
   }
 
-  const root = dirname(normalize(configFilePath));
+  const root = path.dirname(configFilePath);
   const configContent = readFileSync(configFilePath, 'utf-8');
```
alan-agius4: `readFileSync` cannot use the devkit paths

hansl:
This cannot be applied to patch...

PR #13788, opened 2019-02-28
Add support for tslint 5.13+, Update tslint and karma dependencies

alan-agius4:
- feat(@angular-devkit/build-angular): add support for tslint 5.13+ 
Tslint 5.13+ introduced an addition parameter for the format method, mainly used for check-style formatter. During runtime users using an older version of tslint that doesn't expose this it is not a problem as this parameter will be ignored.

See: https://github.com/palantir/tslint/commit/9000479b69c1f4808acf1dee284b47fbbbfbef13

- feat(@schematics/angular): update minimum karma and tslint versions
Karma 4.0.1 removes usages of some vulnerable dependencies
See: https://github.com/karma-runner/karma/releases/tag/v4.0.1

Also, this updates tslint to ~5.13.1 which comes with several bugfixes https://github.com/palantir/tslint/blob/master/CHANGELOG.md#change-log

Supersedes https://github.com/angular/angular-cli/pull/13813

### DIFF
--- a/packages/angular_devkit/build_angular/src/tslint/index.ts
+++ b/packages/angular_devkit/build_angular/src/tslint/index.ts
@@ -35,6 +35,10 @@ export interface TslintBuilderOptions {
   files: string[];
 }
 
+interface LintResult extends tslint.LintResult {
+  fileNames: string[];
+}
+
 export default class TslintBuilder implements Builder<TslintBuilderOptions> {
 
   constructor(public context: BuilderContext) { }
@@ -82,7 +86,7 @@ export default class TslintBuilder implements Builder<TslintBuilderOptions> {
           : null;
         const Linter = projectTslint.Linter;
 
-        let result: undefined | tslint.LintResult;
+        let result: undefined | LintResult;
         if (options.tsConfig) {
           const tsConfigs = Array.isArray(options.tsConfig) ? options.tsConfig : [options.tsConfig];
           const allPrograms =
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_angular/test/tslint/works2_spec_large.ts
```diff
@@ -9,13 +9,19 @@
 import { WorkspaceNodeModulesArchitectHost } from '@angular-devkit/architect/node';
```
alan-agius4: @hansl kindly note that I had to do these changes to make the tests run on windows.

filipesilva:
For the record, this was a TSLint breaking change on a minor: https://github.com/palantir/tslint/issues/4555. We should keep an eye on that issue to see what the resolution is.

Teamop:
seems fixed in v5.13.1

packages/angular_devkit/build_angular/src/tslint/index.ts
```diff
@@ -49,7 +53,7 @@ export default class TslintBuilder implements Builder<TslintBuilderOptions> {
 
     const version = tslint.Linter.VERSION && tslint.Linter.VERSION.split('.');
     if (!version || version.length < 2 || Number(version[0]) < 5 || Number(version[1]) < 5) {
-      throw new Error('TSLint must be version 5.5 or higher.');
+      throw new Error('TSLint must be version 5.13 or higher.');
```
filipesilva: If we need a newer TSLint we need a migration for it. We shouldn't just show an error after update.

Even better than that would be to not need an update at all though... TSLint is going away in the future for ESLint, and on their side they have reverted the breaking change anyway. 

Maybe it's better to not use the optional argument at all and only require an update when TSLint becomes deprecated.

alan-agius4: Following our offline discussion, we don't to do this check nor a migration. 
More details in the PR description.

alan-agius4:
The PR build will fail as it depends on https://github.com/angular/angular-cli/pull/13813

packages/angular_devkit/build_angular/src/tslint/index.ts
filipesilva: This check isn't needed because the extra parameter is just ignored in versions 5.5 to 5.12.

PR #13801, opened 2019-03-01
feat(@schematics/angular): tslint migration for 8

mgechev:
Migration of the `tslint.json` file required by the refactoring of
codelyzer. For more information check this PR
https://github.com/mgechev/codelyzer/pull/754.

// cc @rafaelss95, would you check if I've missed anything?

### DIFF
--- a/packages/schematics/angular/migrations/update-8/index.ts
+++ b/packages/schematics/angular/migrations/update-8/index.ts
@@ -1,0 +1,103 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { JsonParseMode, parseJsonAst } from '@angular-devkit/core';
+import {
+  Rule,
+  SchematicContext,
+  Tree,
+  chain,
+} from '@angular-devkit/schematics';
+import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
+import {
+  NodeDependency,
+  NodeDependencyType,
+  addPackageJsonDependency,
+} from '../../utility/dependencies';
+import { findPropertyInAstObject } from '../../utility/json-utils';
+
+const ruleMapping: {[key: string]: string} = {
+  'contextual-life-cycle': 'contextual-lifecycle',
+  'no-conflicting-life-cycle-hooks': 'no-conflicting-lifecycle',
+  'no-life-cycle-call': 'no-lifecycle-call',
+  'use-life-cycle-interface': 'use-lifecycle-interface',
+  'decorator-not-allowed': 'contextual-decorator',
+  'enforce-component-selector': 'use-component-selector',
+  'no-output-named-after-standard-event': 'no-output-native',
+  'use-host-property-decorator': 'no-host-metadata-property',
+  'use-input-property-decorator': 'no-inputs-metadata-property',
+  'use-output-property-decorator': 'no-outputs-metadata-property',
+  'no-queries-parameter': 'no-queries-metadata-property',
+  'pipe-impure': 'no-pipe-impure',
+  'use-view-encapsulation': 'use-component-view-encapsulation',
+  i18n: 'template-i18n',
+  'banana-in-box': 'template-banana-in-box',
+  'no-template-call-expression': 'template-no-call-expression',
+  'templates-no-negated-async': 'template-no-negated-async',
+  'trackBy-function': 'template-use-track-by-function',
+  'no-attribute-parameter-decorator': 'no-attribute-decorator',
+  'max-inline-declarations': 'component-max-inline-declarations',
+};
+
+function updateTsLintConfig(): Rule {
+  return (host: Tree) => {
+    const tsLintPath = '/tslint.json';
+    const buffer = host.read(tsLintPath);
+    if (!buffer) {
+      return host;
+    }
+    const tsCfgAst = parseJsonAst(buffer.toString(), JsonParseMode.Loose);
+
+    if (tsCfgAst.kind != 'object') {
+      return host;
+    }
+
+    const rulesNode = findPropertyInAstObject(tsCfgAst, 'rules');
+    if (!rulesNode || rulesNode.kind != 'object') {
+      return host;
+    }
+
+    const recorder = host.beginUpdate(tsLintPath);
+
+    rulesNode.properties.forEach(prop => {
+      const mapping = ruleMapping[prop.key.value];
+      if (mapping) {
+        recorder.remove(prop.key.start.offset + 1, prop.key.value.length);
+        recorder.insertLeft(prop.key.start.offset + 1, mapping);
+      }
+    });
+
+    host.commitUpdate(recorder);
+
+    return host;
+  };
+}
+
+function updatePackageJson() {
+  return (host: Tree, context: SchematicContext) => {
+    const dependency: NodeDependency = {
+      type: NodeDependencyType.Dev,
+      name: 'codelyzer',
+      version: '^5.0.0',
+      overwrite: true,
+    };
+
+    addPackageJsonDependency(host, dependency);
+    context.addTask(new NodePackageInstallTask());
+
+    return host;
+  };
+}
+
+export default function(): Rule {
+  return () => {
+    return chain([
+      updateTsLintConfig(),
+      updatePackageJson(),
+    ]);
+  };
+}
--- a/packages/schematics/angular/migrations/update-8/index_spec.ts
+++ b/packages/schematics/angular/migrations/update-8/index_spec.ts
@@ -1,0 +1,85 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+// tslint:disable:no-big-function
+import { EmptyTree } from '@angular-devkit/schematics';
+import { SchematicTestRunner, UnitTestTree } from '@angular-devkit/schematics/testing';
+
+const renames = [
+  'use-lifecycle-interface',
+  'no-host-metadata-property',
+  'no-outputs-metadata-property',
+  'no-inputs-metadata-property',
+];
+
+describe('Migration to version 8', () => {
+  const schematicRunner = new SchematicTestRunner(
+    'migrations',
+    require.resolve('../migration-collection.json'),
+  );
+
+  let tree: UnitTestTree;
+  const tslintPath = '/tslint.json';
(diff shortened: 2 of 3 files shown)


packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    return host;
+  };
+}
+
+export default function(): Rule {
+  return () => {
+    return chain([
+      updateTsLintConfig(),
+      (host: Tree, context: SchematicContext) => {
```
alan-agius4: I don’t think this is relevant for this migration, lint rules has no potential side effects on npm scripts

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      return host;
+    }
+
+    const recorder = host.beginUpdate(tsLintPath);
+
+    rulesNode.properties.forEach(prop => {
+      const mapping = ruleMapping[prop.key.value];
+      if (mapping) {
+        recorder.remove(prop.key.start.offset + 1, prop.key.value.length);
+        recorder.insertLeft(prop.key.start.offset + 1, mapping);
```
alan-agius4: Shouldn’t we append, if this rules is not found in the users tslint?

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,85 @@
+/**
```
alan-agius4: Probably it makes sense that in this rule we update the codelyzer package aswell.

rafaelss95:
LGTM, @mgechev :)

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -33,6 +33,9 @@ const ruleMapping: {[key: string]: string} = {
   'no-template-call-expression': 'template-no-call-expression',
   'templates-no-negated-async': 'template-no-negated-async',
   'trackBy-function': 'template-use-track-by-function',
+  'no-attribute-parameter-decorator': 'no-attribute-decorator',
+  'component-change-detection': 'prefer-on-push-component-change-detection',
```
rafaelss95: This rule is not released yet so I think you can get rid of this.

mgechev: Nothing is released yet. I'd add a label "blocked" so we don't merge.

packages/schematics/angular/migrations/update-8/index.ts
mgechev: 👍 done

packages/schematics/angular/migrations/update-8/index.ts
rafaelss95: I mean that the rule `component-change-detection` doesn't need to be updated because it'll probably only be released on the next major of Codelyzer.

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,128 @@
...
+  if (tree.exists(possiblePath)) {
+    return possiblePath;
+  }
+
+  throw new SchematicsException('Could not find configuration file');
+}
+
+export default function(): Rule {
+  return (host: Tree) => {
+    const configPath = getConfigPath(host);
```
alan-agius4: No need for this logic, you can use the `getPackageJsonDependency`  https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/utility/dependencies.ts#L65 util

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,128 @@
...
+    const dependency: NodeDependency = {
+      type: NodeDependencyType.Dev,
+      name: 'codelyzer',
+      version: '^5.0.0',
+      overwrite: true,
+    };
+
+    addPackageJsonDependency(host, dependency);
+
+    context.addTask(new NodePackageInstallTask({
```
alan-agius4: I am not sure if this is still needed, I haven't seen this around since version 6.

//cc @clydin

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,128 @@
...
+  return (host: Tree) => {
+    const configPath = getConfigPath(host);
+    const configBuffer = host.read(normalize(configPath));
+    if (configBuffer == null) {
+      throw new SchematicsException(`Could not find configuration file (${configPath})`);
+    }
+
+    const config = parseJson(configBuffer.toString(), JsonParseMode.Loose);
+    if (typeof config != 'object' || Array.isArray(config) || config === null) {
+      throw new SchematicsException('Invalid angular-cli.json configuration; expected an object.');
```
alan-agius4: If passing the package manager is still needed, the config file here should be `angular.json` and not `angular-cli.json`

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,122 @@
...
+      }
+    });
+
+    host.commitUpdate(recorder);
+
+    return host;
+  };
+}
+
+function updatePackageJson(config: CliConfig) {
```
alan-agius4: redundant argument

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (tree.exists(possiblePath)) {
+    return possiblePath;
+  }
+
+  throw new SchematicsException('Could not find configuration file');
+}
+
+export default function(): Rule {
+  return (host: Tree) => {
+    const configPath = getConfigPath(host);
```
alan-agius4: The code here, L109-L118 is now redundant, as `updatePackageJson` no longer require the config

packages/schematics/angular/migrations/update-8/index_spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      'directive-class-suffix': true,
+    },
+  };
+  const packageJson = {
+    devDependencies: {
+      codelyzer: '^4.5.0',
+    },
+  };
+
+  describe('tslint.json', () => {
```
alan-agius4: Optional nit: change the description of this test, and you are also testing that the `package.json`  is updated.

scripts/test.ts
```diff
@@ -213,6 +213,8 @@ export default function (args: ParsedArgs, logger: logging.Logger) {
     logger.info(`Found ${tests.length} spec files, out of ${allTests.length}.`);
   }
 
+  tests = tests.filter(e => e.endsWith('update-8/index_spec.ts'));
```
alan-agius4: this shouldn't be committed.

Note: you can use `fit` to run a single test

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,110 @@
...
+  return (host: Tree) => {
+    const dependency: NodeDependency = {
+      type: NodeDependencyType.Dev,
+      name: 'codelyzer',
+      version: '^5.0.0',
+      overwrite: true,
+    };
+
+    addPackageJsonDependency(host, dependency);
+
```
alan-agius4: You need to add `context.addTask(new NodePackageInstallTask());` here so to install the dependency update.

This commit was already correct: https://github.com/angular/angular-cli/commit/544927e9a73a7682c52ac30d20b733aee71939d9#diff-3b943998563d6e245cac63da45ce53f2R92

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,110 @@
...
+function updatePackageJson() {
+  return (host: Tree) => {
+    const dependency: NodeDependency = {
+      type: NodeDependencyType.Dev,
+      name: 'codelyzer',
+      version: '^5.0.0',
+      overwrite: true,
+    };
+
+    addPackageJsonDependency(host, dependency);
```
alan-agius4: You need to add `context.addTask(new NodePackageInstallTask());` here so to install the dependency update.

This commit was already correct: https://github.com/angular/angular-cli/commit/544927e9a73a7682c52ac30d20b733aee71939d9#diff-3b943998563d6e245cac63da45ce53f2R92

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  'no-queries-parameter': 'no-queries-metadata-property',
+  'pipe-impure': 'no-pipe-impure',
+  'use-view-encapsulation': 'use-component-view-encapsulation',
+  i18n: 'template-i18n',
+  'banana-in-box': 'template-banana-in-box',
+  'no-template-call-expression': 'template-no-call-expression',
+  'templates-no-negated-async': 'template-no-negated-async',
+  'trackBy-function': 'template-use-track-by-function',
+  'no-attribute-parameter-decorator': 'no-attribute-decorator',
+  'max-inline-declarations': 'component-max-inline-declarations',
```
wKoza: The rule `'component-change-detection': 'prefer-on-push-component-change-detection'` should be added now.

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    return host;
+  };
+}
+
+function updatePackageJson() {
+  return (host: Tree, context: SchematicContext) => {
+    const dependency: NodeDependency = {
+      type: NodeDependencyType.Dev,
+      name: 'codelyzer',
+      version: '^5.0.0',
```
wKoza: Currently, we use the tilde `~`. this change is it wanted ?

packages/schematics/angular/migrations/update-8/index.ts
rafaelss95: @wKoza this rule did not exist before. It's part of v5 beta, already with the "new" name.

wKoza: Okay, I didn't understand your previous comment and I see this rule since a lot of time.

packages/schematics/angular/migrations/update-8/index.ts
mgechev: We've been careful with not introducing breaking changes between minor releases. I think we can have a `^`.

packages/schematics/angular/migrations/update-8/index.ts
```diff
@@ -0,0 +1,103 @@
+/**
```
alexeagle: why is this the `index` file? aren't there a bunch of migrations for 8? I'd expect this to be update-codelyzer or something

mgechev: I have a follow up PR for polyfills migration. I'll update the script there.

PR #13802, opened 2019-03-02
fix(@angular-devkit/build-angular): remove index option as required

### LINKED ISSUE #11816: Why angular inject scripts in my code because I want to inject manually!
Angular 6 inject scripts in my code, and I don't want that!

capture
first

This happens after I run my .net core app which uses static files generated by angular cli with ng build command

clydin:
It is not a required option and is allowed to be undefined ([relevant code](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/browser.ts#L39)).  This supports use cases where an index HTML file may not be used directly or is autogenerated in some other manner.

Closes #11816

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
@@ -62,7 +62,7 @@ export interface BuildOptions {
   es5BrowserSupport?: boolean;
 
   main: string;
-  index: string;
+  index?: string;
   polyfills?: string;
   budgets: Budget[];
   assets: AssetPatternClass[];
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -324,7 +324,6 @@
   "additionalProperties": false,
   "required": [
     "outputPath",
-    "index",
     "main",
     "tsConfig"
   ],


packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -197,7 +197,7 @@ export class DevServerBuilder implements Builder<DevServerBuilderSchema> {
       port: options.port,
       headers: { 'Access-Control-Allow-Origin': '*' },
       historyApiFallback: {
```
alan-agius4: If no index is provided shouldn't this be disabled?

clydin: The default value for it is `index.html` so that seemed appropriate.  This also enables a use case wherein a developer copies a pre-generated index file via an asset configuration.

clydin:
Actually, the reason I added the defined check was to support use cases that don't need the index.  I just neglected to update the schema to allow it to happen.  If a developer is integrating the build files into a larger project (ASP.net or other serve side framework), the index file and its processing is unnecessary and adds complexity to the setup since it needs to be removed (it could cause odd errors if accidentally left in a location that will eventually be publicly accessible).  There's also the use case where a completely custom templating based setup (or even a static site generator) could be used to generate a final index.html file from the ng output directory.  The order of the scripts within the document is also well prescribed and cannot change between majors as it would be a breaking change.  The webpack stats output also provides information related to the scripts which can be leveraged to generate if needed.  This also brings up an additional feature request mentioned previously of generating a basic manifest of all output files and their hashes.  This also would aid in advanced CSP rule generation for instance (something I think we should be making as easy as possible).

There's also many potential developer setups that don't rely on `ng serve` at all or only use it to serve the actual bundles as part of a larger setup (e.g., nginx at the front with an asp.net server and ng serve behind it; or just an asp.net server).

I'm good with moving this to a feature and adding a warning on serve and some additional testing.  I think we should be supporting the build only scenarios in addition to the full development scenarios.

PR #13810, opened 2019-03-03
fix(@schematics/angular): use peerDependency on typescript

alexeagle:
This saves 7MB in the initial install of @angular/cli which depends on this package


alexeagle:
@clydin points out this is a breaking change to the install instructions

PR #13821, opened 2019-03-05
Remove dep @schematics/angular -> typescript

alexeagle:


packages/schematics/angular/BUILD
```diff
@@ -49,13 +49,20 @@ ts_library(
...
+        "//packages/schematics/angular/third_party/github.com/Microsoft/TypeScript",
         "@npm//@types/node",
-        "@npm//typescript",
+        "@npm//rxjs",
     ],
 )
 
+jasmine_node_test(
+    name = "no_typescript_runtime_dep_test",
+    srcs = ["no_typescript_runtime_dep_test.js"],
```
hansl: Why not `_spec`? :zoidberg:

PR #13811, opened 2019-03-04
 feat(@schematics/angular): change layout for root applications

alan-agius4:
This change aligns the file layout of applications generated with `ng new` and `ng generate`

PR #13830, opened 2019-03-06
feat(@schematics/angular): relocate `ng g service-worker` files

alan-agius4:
This is a follow up on https://github.com/angular/angular-cli/pull/13811, with this now all application have the same structure and hence `ng g service-worker` should follow suit and have the same folder structure if generated in a root app, or an app inside the projects.

### DIFF
--- a/packages/schematics/angular/service-worker/index.ts
+++ b/packages/schematics/angular/service-worker/index.ts
@@ -60,7 +60,7 @@ function updateConfigFile(options: ServiceWorkerOptions, root: string): Rule {
 
     const config = getProjectConfiguration(workspace, options);
     config.serviceWorker = true;
-    config.ngswConfigPath = `${root.endsWith('/') ? root : root + '/'}ngsw-config.json`;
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/service-worker/index.ts
```diff
@@ -60,7 +60,7 @@ function updateConfigFile(options: ServiceWorkerOptions, root: string): Rule {
 
     const config = getProjectConfiguration(workspace, options);
     config.serviceWorker = true;
-    config.ngswConfigPath = `${root.endsWith('/') ? root : root + '/'}ngsw-config.json`;
+    config.ngswConfigPath = `${root && !root.endsWith('/') ? root + '/' : root}ngsw-config.json`;
```
clydin: If root happens to be `undefined` the string will be `undefinedngsw-config.json`

alan-agius4: If `project.root` is ever undefined. It will break half of our workflow :p

PR #13835, opened 2019-03-06
feat(@angular-devkit/core): analytics interfaces and basic implementations

hansl:
This is a preliminary analytics interface. It is meant to be used by every
other subsystem where useful, but can be replaced by implementations.

### DIFF
--- a/packages/angular_devkit/core/src/analytics/api.ts
+++ b/packages/angular_devkit/core/src/analytics/api.ts
@@ -1,0 +1,52 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+export interface CustomDimensionsAndMetricsOptions {
+  dimensions?: (boolean | number | string)[];
+  metrics?: (boolean | number | string)[];
+}
+
+export interface EventOptions extends CustomDimensionsAndMetricsOptions {
+  label?: string;
+  value?: string;
+}
+
+export interface ScreenviewOptions extends CustomDimensionsAndMetricsOptions {
+  appVersion?: string;
+  appId?: string;
+  appInstallerId?: string;
+}
+
+export interface PageviewOptions extends CustomDimensionsAndMetricsOptions {
+  hostname?: string;
+  title?: string;
+}
+
+export interface TimingOptions extends CustomDimensionsAndMetricsOptions {
+  label?: string;
+}
+
+/**
+ * Interface for managing analytics. This is highly platform dependent, and mostly matches
+ * Google Analytics. The reason the interface is here is to remove the dependency to an
+ * implementation from most other places.
+ *
+ * The methods exported from this interface more or less match those needed by us in the
+ * universal analytics package, see https://unpkg.com/@types/universal-analytics@0.4.2/index.d.ts
+ * for typings. We mostly named arguments to make it easier to follow, but didn't change or
+ * add any semantics to those methods. They're mapping GA and u-a one for one.
+ *
+ * The Angular CLI (or any other kind of backend) should forward it to some compatible backend.
+ */
+export interface Analytics {
+  event(category: string, action: string, options?: EventOptions): void;
+  screenview(screenName: string, appName: string, options?: ScreenviewOptions): void;
+  pageview(path: string, options?: PageviewOptions): void;
+  timing(category: string, variable: string, time: string | number, options?: TimingOptions): void;
+
+  flush(): Promise<void>;
+}
--- a/packages/angular_devkit/core/src/analytics/forwarder.ts
+++ b/packages/angular_devkit/core/src/analytics/forwarder.ts
@@ -1,0 +1,131 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { JsonObject } from '../json';
+import { Analytics, EventOptions, PageviewOptions, ScreenviewOptions, TimingOptions } from './api';
+
+
+export enum AnalyticsReportKind {
+  Event = 'event',
+  Screenview = 'screenview',
+  Pageview = 'pageview',
+  Timing = 'timing',
+}
+
+export interface AnalyticsReportBase extends JsonObject {
+  kind: AnalyticsReportKind;
+}
+
+export interface AnalyticsReportEvent extends AnalyticsReportBase {
+  kind: AnalyticsReportKind.Event;
+  options: JsonObject & EventOptions;
+  category: string;
+  action: string;
+}
+export interface AnalyticsReportScreenview extends AnalyticsReportBase {
+  kind: AnalyticsReportKind.Screenview;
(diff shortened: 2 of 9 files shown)


packages/angular_devkit/core/src/analytics/api.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export interface CustomDimensionsAndMetricsOptions {
+  dimensions?: (boolean | number | string)[];
+  metrics?: (boolean | number | string)[];
+}
+
+export interface EventOptions extends CustomDimensionsAndMetricsOptions {
+  label?: string;
+  value?: string;
+  dimensions?: (boolean | number | string)[];
```
filipesilva: I think you left in the inherited properties by accident, or maybe you meant to change them.

packages/angular_devkit/core/src/analytics/forwarder.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+export interface AnalyticsReportBase extends JsonObject {
+  kind: AnalyticsReportKind;
+}
+
+export interface AnalyticsReportEvent extends AnalyticsReportBase {
+  kind: AnalyticsReportKind.Event;
+  options: JsonObject & EventOptions;
+  c: string;
```
filipesilva: Is there a good reason to leave these properties as single letter instead of a real name? If so, it should be mentioned somewhere.

packages/angular_devkit/core/src/analytics/forwarder.ts
```diff
@@ -0,0 +1,131 @@
...
+  ;
+
+/**
+ * A function that can forward analytics along some stream. AnalyticsReport is already a
+ * JsonObject descendant, but we force it here so the user knows it's safe to serialize.
+ */
+export type AnalyticsForwarderFn = (report: JsonObject & AnalyticsReport) => void;
+
+/**
+ * A class that follows the Analytics interface and forward analytic reports (JavaScript objects).
```
filipesilva: Nit: `and forward` -> `and forwards`

packages/angular_devkit/core/src/analytics/logging.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Logger } from '../logger';
+import { Analytics, EventOptions, PageviewOptions, ScreenviewOptions, TimingOptions } from './api';
+
+/**
+ * Analytics implementation that reports to multiple analytics backend.
```
filipesilva: This comment should mention logging instead.

packages/angular_devkit/core/src/analytics/forwarder.ts
```diff
@@ -0,0 +1,131 @@
...
+      c: category,
+      v: variable,
+      t: time,
+      options: { ...options } as JsonObject,
+    });
+  }
+
+  // We do not support flushing.
+  flush(): never {
+    throw new Error('Flush is not supported.');
```
filipesilva: I find this pattern a bit confusing:
- `AnalyticsForwarder` doesn't support flushing and throws
- `LoggingAnalytics` and `NoopAnalytics` just resolve a promise without doing anything
- `MultiAnalytics` flushes all of its backends

There are 3 other `Analytics` implementations for `MultiAnalytics` to use. Two of them do nothing, and one of them throws. 

I think this leaves flushing with `MultiAnalytics` analytics in an odd place since it will always fail if there's at least one `AnalyticsForwarder`, and you can't really know which one failed, or if multiple failed, or how to proceed.

Maybe flushing should instead do nothing when it's not supported?

packages/angular_devkit/core/src/analytics/forwarder.ts
```diff
@@ -0,0 +1,131 @@
...
+ * JsonObject descendant, but we force it here so the user knows it's safe to serialize.
+ */
+export type AnalyticsForwarderFn = (report: JsonObject & AnalyticsReport) => void;
+
+/**
+ * A class that follows the Analytics interface and forward analytic reports (JavaScript objects).
+ * AnalyticsReporter is the counterpart which takes analytic reports and report them to another
+ * Analytics interface.
+ */
+export class AnalyticsForwarder implements Analytics {
```
filipesilva: The other two implementations follow the `SomethingAnalytics` naming scheme but this one doesn't. Maybe this should be called `ForwardingAnalytics`?

scripts/validate.ts
```diff
@@ -43,6 +44,11 @@ export default async function (options: { verbose: boolean }, logger: logging.Lo
   error = await validateCommits({}, logger.createChild('validate-commits')) != 0
        || error;
 
+  logger.info('');
+  logger.info('Running DO_NOT' + '_SUBMIT validation...');
```
filipesilva: Nit: here `DO_NOT' + '_SUBMIT` is used to (I assume) get around the check, but `DO_NOT${''}_SUBMIT` is used in `scripts/validate-do-not-submit.ts` instead. Would be helpful to have a comment explaining this and to make the workaround consistent.

scripts/validate-do-not-submit.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export default async function (argv: ValidateCommitsOptions, logger: logging.Logger) {
+  logger.info('Getting merge base...');
+
+  const prNumber = process.env['CIRCLE_PR_NUMBER'] || '';
+  let baseSha = '';
+  let sha = '';
+
+  if (prNumber) {
+    const url = `https://api.github.com/repos/angular/angular-cli/pulls/${prNumber}`;
```
filipesilva: Nice way to determine the base 👍 

We should do something similar in our circleci post checkout step to properly rebase instead of fiddling with the merge commits: https://github.com/angular/angular-cli/blob/e96afb874d4a03a8bf0a9cac9d79dd7dc7db31de/.circleci/config.yml#L31-L48

packages/angular_devkit/core/src/analytics/api.ts
hansl: That was a mistake, good catch!

packages/angular_devkit/core/src/analytics/forwarder.ts
hansl: No really. They're are stringified and parsed to/from JSON, but I don't think it makes enough of a difference. Fixed.

packages/angular_devkit/core/src/analytics/forwarder.ts
hansl: You're right, it should do nothing or flush. Ideally it should actually flush and wait for the other side to be done, but that requires two-way communication and a lot more plumbing than I'm comfortable with. So it will be a do nothing and hope for the best.

packages/angular_devkit/core/src/analytics/forwarder.ts
hansl: Good name, done.

scripts/validate.ts
hansl: Done.

scripts/validate-do-not-submit.ts
hansl: Yeah I'm not sure. We should re-evaluate at some point

PR #13837, opened 2019-03-06
feat(@angular-devkit/schematics): support unicode character HTML element names

clydin:
The HTML specification allows for a wide variety of characters to be present within a custom element name.  The previous behavior limited the names to mostly alphanumeric characters.  This change opens up the names to include the characters specified within the specification for custom element names.

For example: `app--2rootßå` and `app-root😀` are valid names which work (serve/test) and with AOT.

### DIFF
--- a/packages/angular_devkit/schematics/src/formats/html-selector.ts
+++ b/packages/angular_devkit/schematics/src/formats/html-selector.ts
@@ -5,18 +5,47 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-
 import { schema } from '@angular-devkit/core';
 
+// As per https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name
+// * Without mandatory `-` as the application prefix will generally cover its inclusion
+// * And an allowance for upper alpha characters
 
-// Must start with a letter, and must contain only alphanumeric characters or dashes.
-// When adding a dash the segment after the dash must also start with a letter.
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/schematics/src/formats/html-selector.ts
```diff
@@ -5,18 +5,47 @@
...
+  [0x203F, 0x2040],
+  [0x2070, 0x218F],
+  [0x2C00, 0x2FEF],
+  [0x3001, 0xD7FF],
+  [0xF900, 0xFDCF],
+  [0xFDF0, 0xFFFD],
+  [0x10000, 0xEFFFF],
+];
+
+function isValidElementName(name: string) {
```
alan-agius4: Seeing this https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name

Valid elements should not contain upper cases.

Nit: add return type.

clydin: I made an exception for that since this is currently used for directives as well and many/most of them are camel case. The comment block above has a note about the inclusion of upper alpha characters.
I also didn't want to at least initially cause any further restrictions on the naming.  There's also a list of reserved element names that is also not included but since this is also used for partial (no prefix) names, checking for them doesn't work well currently.
But I agree that we should eventually be more restrictive for the actual element names.  This would require changes to the schematics as well as additional rules for custom attribute names.

alan-agius4: In that case LGTM.

PR #13838, opened 2019-03-06
fix(@angular-devkit/build-angular): ensure correct dev server path with public host option

### LINKED ISSUE #13627: Wrong SockJS Host validation with Reverse Proxy
# 🐞 Bug report

Angular CLI version : 7.2.3

```
- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

I don't know

### Description

When using "ng serve" behind a reverse proxy on a sub-path, the SockJS Server throws 400 errors because it fails to validate the Host/Origin header.

## 🔬 Minimal Reproduction

I want to have my application served on my workstation at `http://dev.local/common-demo`

1. Create a new Angular CLI project
2. Configure the `serve` task in `angular.json` with
```json
            "host": "localhost",
            "port": 9038,
            "publicHost": "dev.local/common-demo/sockjs-node",
            "baseHref": "/common-demo/",
            "servePath": "/"
```
3. Configure you workstation reverse proxy with (here Apache is used)
```
ProxyPass /common-demo http://localhost:9038
ProxyPassReverse /common-demo http://localhost:9038
```

## 🔥 Exception or Error

Chrome console shows :

```
WebSocket connection to 'ws://dev.local/common-demo/sockjs-node/287/kbmqffv3/websocket' failed: Error during WebSocket handshake: Unexpected response code: 400

Invalid Host/Origin header

[WDS] Disconnected!
```

**Notes**

If I set `disableHostCheck` to `true`, it works, but I don't want to.

If I set `publicHost` to `dev.local` the app tries to connect to `ws://dev.local/sockjs-node` which is not valid as it ignores the `baseHref`.

The `servePath` needs to be `/` as the app is served at the root of `localhost:9038` (by default `servePath` inherits from `baseHref` but the sub-path is handled by the Reverse Proxy in my setup).

clydin:
If a custom path is provided the webpack dev server client drops the sockjs-node segment. This adds it back so that behavior is consistent when using a custom URL path.

Fixes #13627

### DIFF
--- a/packages/angular_devkit/build_angular/src/dev-server/index.ts
+++ b/packages/angular_devkit/build_angular/src/dev-server/index.ts
@@ -144,15 +144,14 @@ export function serveWebpackBrowser(
       options.port = port;
 
       // Resolve public host and client address.
-      let clientAddress = `${options.ssl ? 'https' : 'http'}://0.0.0.0:0`;
+      let clientAddress = url.parse(`${options.ssl ? 'https' : 'http'}://0.0.0.0:0`);
       if (options.publicHost) {
         let publicHost = options.publicHost;
         if (!/^\w+:\/\//.test(publicHost)) {
           publicHost = `${options.ssl ? 'https' : 'http'}://${publicHost}`;
         }
-        const clientUrl = url.parse(publicHost);
-        options.publicHost = clientUrl.host;
-        clientAddress = url.format(clientUrl);
+        clientAddress = url.parse(publicHost);
+        options.publicHost = clientAddress.host;
       }
 
       // Resolve serve address.
@@ -338,7 +337,7 @@ function _addLiveReload(
   options: DevServerBuilderSchema,
   browserOptions: BrowserBuilderSchema,
   webpackConfig: webpack.Configuration,
-  clientAddress: string,
+  clientAddress: url.UrlWithStringQuery,
   logger: logging.LoggerApi,
 ) {
   if (webpackConfig.plugins === undefined) {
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -243,7 +242,14 @@ export class DevServerBuilder implements Builder<DevServerBuilderSchema> {
...
     }
-    const entryPoints = [`${webpackDevServerPath}?${clientAddress}`];
+
+    // If a custom path is provided the webpack dev server client drops the sockjs-node segment.
+    // This adds it back so that behavior is consistent when using a custom URL path
+    if (clientAddress.pathname) {
+      clientAddress.pathname = path.posix.join(clientAddress.pathname, 'sockjs-node');
+    }
+
+    const entryPoints = [`${webpackDevServerPath}?${url.format(clientAddress)}`];
```
alan-agius4: Any reason why we don't use `addDevServerEntrypoints`?
https://github.com/webpack/webpack-dev-server/blob/21687c3bfc95752957d16470b97d281001dc6a07/lib/Server.js#L971
https://github.com/webpack/webpack-dev-server/blob/21687c3bfc95752957d16470b97d281001dc6a07/lib/utils/addEntries.js#L6

clydin: Historical mostly.

I was intending this to be minimal so that it could go on patch as well.
Once the update builders are in place, I'd like to go through each one and cleanup/refactor where possible.

PR #13840, opened 2019-03-06
Various QoL for caretaker

hansl:
Just got a mistake (caught in time) so decided to do a long overdue refactoring for release script(s).

### DIFF
--- a/.monorepo.json
+++ b/.monorepo.json
@@ -42,16 +42,10 @@
   },
   "packages": {
     "@_/benchmark": {
-      "version": "0.14.0-beta.5",
-      "hash": "5225044ecf49d630564a68b710f6b82f"
     },
     "@_/builders": {
-      "version": "0.14.0-beta.4",
-      "hash": "569aef8bc23693c344e71b9755e62c50"
     },
     "devkit": {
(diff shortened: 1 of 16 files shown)


.monorepo.json
```diff
@@ -1,4 +1,8 @@
 {
+  "versions": {
+    "stable": "8.0.0-beta.5",
```
alexeagle: FYI on Angular we determine the version based on git tags; see https://github.com/angular/angular/blob/master/tools/bazel_stamp_vars.js

If we also had a scheme for making the 0.* versions based on the stable ones, then we'd only need one bit to describe the versioning.

That would help us move to bazel for build/publish someday

lib/packages.ts
```diff
@@ -28,53 +26,19 @@ export interface PackageInfo {
   build: string;
   tar: string;
   private: boolean;
+  experimental: boolean;
```
alexeagle: does this change mean we can update release.md to stop using --force?

hansl: Yes, done.

.monorepo.json
hansl: All done.

PR #13848, opened 2019-03-07
fix(@ngtools/webpack): `getCurrentDirectory` should not return a devkit path

alan-agius4:
At the moment we are leaking devkit paths as `getCurrentDirectory` is used by the ngtsc compiler if no `rootDir` directories is provided.

`getCurrentDirectory` is also used to construct the `rootDirs` option which is causing this to be incorrect as under Windows as `rootDirs` will be have an invalid value example: `/c/foo/bar`

This PR complements https://github.com/angular/angular/pull/29151

**Note** that this currently works because we are normalizing paths with posix seperators and is mimicking the behaviour of the logical file system used by the ngtsc.

Fixes: https://github.com/angular/angular-cli/issues/13849


appveyor-e2e.bat
```diff
@@ -2,11 +2,11 @@
 IF not defined APPVEYOR_PULL_REQUEST_NUMBER (
   IF not defined BUILDKITE_PULL_REQUEST (
     REM Run full test suite if not on a Appveyor/Buildkite PR.
-    node tests\legacy-cli\run_e2e.js --appveyor "--glob=tests/{basic,commands,generate,build/styles}/**"
+    node tests\legacy-cli\run_e2e.js --appveyor "--glob=tests/{basic,commands,generate,build/styles/ivy}/**"
```
filipesilva: Shouldn't this be `,ivy` instead of `/ivy`?

tests/legacy-cli/e2e/tests/ivy/ivy-lazy-load.ts
```diff
@@ -0,0 +1,88 @@
+/**
```
filipesilva: The tests are removed in the previous commit, but only re-added here. They should be moved in this commit instead.

appveyor-e2e.bat
alan-agius4: Yes, totally fixed.

PR #13852, opened 2019-03-07
Extensions

alexeagle:
Rolling this forward with a README explaining our agreed-on policy.

@cyrilletuzi we want to find a fair way to include extensions from the community, including yours. We are stuck on how to do this without confusing users with a set of overlapping extensions - we assume that many users will install all of our recommendations. That forces us to pick winners and that's not a good position for us to be in.
The governance constraint is the primary one we'll use for guidance in version 8. So the easiest path to get your code in front of users is to make it first-party.
Happy to discuss further.

/fyi @mrmeku


alexeagle:
Not ready to merge, needs more input

mgechev:
Closing since we don't have progress here for the past few months.

PR #13855, opened 2019-03-08
Architect qol 2

hansl:
These came out when discussing with Nrwl. They should be in before the next release.

### DIFF
--- a/packages/angular_devkit/architect/src/api_spec.ts
+++ b/packages/angular_devkit/architect/src/api_spec.ts
@@ -1,0 +1,49 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Target, targetFromTargetString, targetStringFromTarget } from './api';
+
+describe('Architect API', () => {
+  const goldens: { [t: string]: Target } = {
+    'a:b:c': { project: 'a', target: 'b', configuration: 'c' },
+    'a:b': { project: 'a', target: 'b' },
+    ':b': { project: '', target: 'b' },
+    ':b:c': { project: '', target: 'b', configuration: 'c' },
+    ':': { project: '', target: '' },
+    '::': { project: '', target: '', configuration: '' },
+    'a:b:': { project: 'a', target: 'b', configuration: '' },
+    'a::c': { project: 'a', target: '', configuration: 'c' },
+    'a:b:c:': { project: 'a', target: 'b', configuration: 'c' },
+    'a:b:c:d': { project: 'a', target: 'b', configuration: 'c' },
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_angular/src/index.ts
```diff
@@ -15,9 +15,26 @@
...
+  FileReplacement,
+  OptimizationClass as OptimizationObject,
+  OptimizationUnion,
+  OutputHashing,
+  Schema as BrowserBuilderSchema,
+  SourceMapClass as SourceMapObject,
+  SourceMapUnion,
+  StylePreprocessorOptions,
+  Type,
+} from './browser/schema';
```
FrozenPandaz: It seems this used to be
```
export * from './browser/schema';
```

Any reason why that is not the case anymore? If it is because of conflicting names, can they be exported differently?

packages/angular_devkit/architect/src/api_spec.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Target, targetFromTargetString, targetStringFromTarget } from './api';
+
+describe('Architect API', () => {
+  const goldens: { [t: string]: Target } = {
+    'a:b:c': { project: 'a', target: 'b', configuration: 'c' },
+    'a:b': { project: 'a', target: 'b' },
+    ':b': { project: '', target: 'b' },
```
FrozenPandaz: Is having an empty string for a `project` actually valid?

packages/angular_devkit/architect/src/api_spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    ':b:c': { project: '', target: 'b', configuration: 'c' },
+    ':': { project: '', target: '' },
+    '::': { project: '', target: '', configuration: '' },
+    'a:b:': { project: 'a', target: 'b', configuration: '' },
+    'a::c': { project: 'a', target: '', configuration: 'c' },
```
FrozenPandaz: Is having an empty string for a `target` actually valid?

FrozenPandaz:
LGTM other than a few questions 👍

packages/angular_devkit/architect/src/api_spec.ts
hansl: Technically yes.

packages/angular_devkit/build_angular/src/index.ts
hansl: QuickType creates the main interface as `Schema`, and suffix new schemas with `Schema` while this was `Object` before. For now I'm aiming at 0 breaking change, but will probably only rename the root `Schema` later (so look out for that breaking change).

packages/angular_devkit/architect/src/api_spec.ts
hansl: Technically yes. This would be the only place we check for that.

Issue #13864, opened 2019-03-10
ng new generates wrong file type


yangshuairocks:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

I don't know, as I'm a new Angular user.

### Description

 A clear and concise description of the problem...
```
C:\my_projects>ng new --help
arguments:
  name
    The name of the new workspace and initial project.

options:
...
  --style
    The file extension or preprocessor to use for style files.
...
```
As you can see, `--style` specifies the file extension and preprocessor. However, from the output of below command, you can see `.scss` files are generated even I specified `--style sass`.

```

C:\my_projects>ng new my-app --style sass
? Would you like to add Angular routing? No
...
CREATE my-app/src/styles.scss (80 bytes)
...
CREATE my-app/src/app/app.component.scss (0 bytes)
...
C:\my_projects>
```

I tried both `--style sass` and `--style scss` but in either cases only `.scss` files are generated. What I expected `.sass` files generated based on the help of `ng new` command.

## 🔥 Exception or Error
<pre><code>

None.
</code></pre>

## 🌍 Your Environment
<pre><code>

C:\my_projects>ng version

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/

Angular CLI: 7.3.5
Node: 10.13.0
OS: win32 x64
Angular:
...

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.13.5
@angular-devkit/core         7.3.5
@angular-devkit/schematics   7.3.5
@schematics/angular          7.3.5
@schematics/update           0.13.5
rxjs                         6.3.3
typescript                   3.2.4

C:\my_projects>
</code></pre>

**Anything else relevant?**

alan-agius4:
[...]

PR #13867, opened 2019-03-11
fix(@schematics/angular): reintroduce `.sass` as a supported file ext…

alan-agius4:
One of the commits in https://github.com/angular/angular-cli/pull/13777 didn't make it to the patch branch.

closes: #13864


vikerman:
Merged as 0fca5f7389b4c86bde1c9f79e21f51e19921d3ec into 7.3.x branch.

PR #13872, opened 2019-03-11
ng config bug

EKashpersky:
Fixed resolving values from angular.json using ng config <path>


alan-agius4:
Hi, thanks for your contribution, however `parseJsonPath` will always return an array of string. Since the usage of string methods such as `split`.

Maybe it would be better to fix just the comment `* ["a", 3, "foo", "bar", 2].` in a separate PR.

Cheers.

PR #13876, opened 2019-03-11
feat(@angular-devkit/build-angular): implement stable architect API for protractor

clydin:
This is a continuation of the Angular builder refactoring following the introduction of the new stable architect builder API.

This is currently blocked on the introduction of the dev server builder refactoring.


packages/angular_devkit/build_angular/src/protractor/index2.ts
```diff
@@ -0,0 +1,148 @@
...
+      context.reportStatus('Error: ' + error);
+
+      return { success: false };
+    }
+  }
+
+  let baseUrl;
+  let server;
+  if (options.devServerTarget) {
+    const parts = (options.devServerTarget as string).split(':');
```
hansl: Please use the new `targetFromTargetString()` function. It will throw if it's not in the "proper format".

Also, is `as string` necessary here? If TypeScript is confused by the type (it happened to me) you might be able to fix it with `options.devServerTarget !== undefined` in the if above).

packages/angular_devkit/build_angular/src/protractor/index2.ts
```diff
@@ -0,0 +1,148 @@
...
+  } catch {
+    return { success: false };
+  } finally {
+    if (server) {
+      await server.stop();
+    }
+  }
+}
+
+export default createBuilder<Record<string, string> & ProtractorBuilderOptions>(execute);
```
hansl: What's the rationale of using `Record<string, string>` over `JsonObject`?

packages/angular_devkit/build_angular/src/protractor/index2.ts
```diff
@@ -0,0 +1,148 @@
...
+      baseUrl = url.format({
+        protocol: serverOptions.ssl ? 'https' : 'http',
+        hostname: options.host,
+        port: result.port.toString(),
+      });
+    }
+  }
+
+  try {
+    return await runProtractor(context.workspaceRoot, { ...options, baseUrl });
```
hansl: Funny I was always thinking that the protractor supported a watch mode, but seems not. At least we got rid of `of(null)` and that TODO. Good work!

packages/angular_devkit/build_angular/test/protractor/works_spec_large.ts
```diff
@@ -5,47 +5,58 @@
...
  */
+import { Architect } from '@angular-devkit/architect/src/index2';
+import { JsonObject, normalize } from '@angular-devkit/core';
+import { createArchitect, host, protractorTargetSpec } from '../utils';
 
-import { runTargetSpec } from '@angular-devkit/architect/testing';
-import { normalize } from '@angular-devkit/core';
-import { retry } from 'rxjs/operators';
-import { host, protractorTargetSpec } from '../utils';
+describe('Protractor Builder', () => {
```
hansl: Any reason to remove the `retry()` on some of them?  Is it more stable now?

packages/angular_devkit/build_angular/src/protractor/index2.ts
clydin: Done.

The cast was left over from a copy and paste.  It's not needed.

packages/angular_devkit/build_angular/src/protractor/index2.ts
clydin: It made the compiler stop complaining and didn't require an extra import.  `JsonObject` is functionally a `Record<string, string | boolean | ...>` and the compiler just wanted something with an index signature.

Using either option breaks the type constraint, the following will happily compile for instance:
```
export default createBuilder<JsonObject & ProtractorBuilderOptions & { f: Function }>(execute);
```

It may be better to just drop the constraint.  A more comprehensive runtime check could be added if needed.

packages/angular_devkit/build_angular/test/protractor/works_spec_large.ts
clydin: Appears more stable.  I've been running them repeatedly over the last week without issue.  I think it also makes the tests more representative as well.  Ideally, we shouldn't have to repeat tests to make them pass.

PR #13877, opened 2019-03-11
feat(@angular-devkit/build-ng-packagr): implement stable architect API

clydin:
This is a continuation of the Angular builder refactoring following the introduction of the new stable architect builder API.

### DIFF
--- a/packages/angular_devkit/build_ng_packagr/src/build/index2.ts
+++ b/packages/angular_devkit/build_ng_packagr/src/build/index2.ts
@@ -1,0 +1,44 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect/src/index2';
+import { resolve } from 'path';
+import { Observable, from } from 'rxjs';
+import { catchError, mapTo, switchMap } from 'rxjs/operators';
+import { Schema as NgPackagrBuilderOptions } from './schema';
+
+async function initialize(
+  options: NgPackagrBuilderOptions,
+  root: string,
+): Promise<import ('ng-packagr').NgPackagr> {
+  const packager = (await import('ng-packagr')).ngPackagr();
+
+  packager.forProject(resolve(root, options.project));
+
+  if (options.tsConfig) {
+    packager.withTsConfig(resolve(root, options.tsConfig));
+  }
+
+  return packager;
+}
+
+export function buildLibrary(
+  options: NgPackagrBuilderOptions,
+  context: BuilderContext,
+): Observable<BuilderOutput> {
+  return from(initialize(options, context.workspaceRoot)).pipe(
+    switchMap(packager => options.watch ? packager.watch() : packager.build()),
+    mapTo({ success: true }),
+    catchError(error => {
+      context.reportStatus('Error: ' + error);
+
+      return [{ success: false }];
+    }),
+  );
+}
+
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_ng_packagr/src/build/schema.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "http://json-schema.org/draft-07/schema",
```
alan-agius4: We we also update 
```
  "project": {
      "type": "string",
      "description": "The file path of the package.json for distribution via npm."
    },
```

to 
```
  "project": {
      "type": "string",
      "description": "The full path for the ng-packagr configuration file, relative to the current workspace."
    },
```

packages/angular_devkit/build_ng_packagr/src/build/index2.ts
```diff
@@ -0,0 +1,44 @@
...
+  root: string,
+): Promise<import ('ng-packagr').NgPackagr> {
+  const packager = (await import('ng-packagr')).ngPackagr();
+
+  packager.forProject(resolve(root, options.project));
+
+  if (options.tsConfig) {
+    packager.withTsConfig(resolve(root, options.tsConfig));
+  }
+
```
alan-agius4: You can do something
```
packager.withOptions({
		watch: options.watch
})
```

packages/angular_devkit/build_ng_packagr/src/build/index2.ts
```diff
@@ -0,0 +1,44 @@
...
+  return packager;
+}
+
+export function buildLibrary(
+  options: NgPackagrBuilderOptions,
+  context: BuilderContext,
+): Observable<BuilderOutput> {
+  return from(initialize(options, context.workspaceRoot)).pipe(
+    switchMap(packager => options.watch ? packager.watch() : packager.build()),
```
alan-agius4: Nit
```ts
switchMap(packager => packagr.buildAsObservable())
```

packages/angular_devkit/build_ng_packagr/src/build/schema.json
hansl: Considering the draft won't be changing much, we should start using a non-draft version of the schema spec; `http://json-schema.org/schema`.

clydin: `http://json-schema.org/schema` points to latest version of the schema and will change over time.  It currently points to the `draft-07` version which allows everything to work for now.  However, there is no guarantee that the latest version will be supported by the CLI nor that the file will conform to the latest version.  By specifying a specific version the `$schema` identifier can be used as both a version identifier for the format of the file and as the location of the schema.  This allows processors to use the specified version/format of the specification when parsing and processing the file.  (ref: http://json-schema.org/draft-07/json-schema-core.html#rfc.section.7)

packages/angular_devkit/build_ng_packagr/src/build/index2.ts
clydin: Will `watch` and `build` be available longterm?  I think I actually prefer the separate methods as they are really different actions; and `build` returning a promise helps simplify and demonstrate that a build is a one time event.

alan-agius4: yeah should be.

packages/angular_devkit/build_ng_packagr/src/build/schema.json
hansl: (we use that value too when running quicktype)

This is non blocking, but we use the "latest" URI in all but 2 of our schemas, so I'd say it's better to use the same everywhere than just keep splitting the code base.

PR #13878, opened 2019-03-11
refactor(@angular-devkit/build-angular): auto-generate karma builder option schema

clydin:
Preparation for refactoring the karma builder following the introduction of the new stable architect builder API.


packages/angular_devkit/build_angular/src/karma/schema.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "http://json-schema.org/draft-07/schema",
```
hansl: Could you use the non-draft Schema URL?

PR #13903, opened 2019-03-13
fix(@angular-devkit/build-angular): resolve fonts with space in filename

### LINKED ISSUE #9648: 1.7.0 fonts are corrupted after prod build
### Versions
Angular CLI: 1.7.0
Node: 8.9.1
OS: win32 x64
Angular: 5.2.5

@angular/cli: 1.7.0
@angular-devkit/build-optimizer: 0.3.1
@angular-devkit/core: 0.3.1
@angular-devkit/schematics: 0.3.1
@ngtools/json-schema: 1.2.0
@ngtools/webpack: 1.10.0
@schematics/angular: 0.3.1
@schematics/package-update: 0.3.1
typescript: 2.7.1
webpack: 3.11.0

### Repro steps
- Install @angular/cli: 1.7.0
- import fontawesome in .angular-cli.json or in styles.scss
- run ng build --prod

### Observed behavior
Failed to decode downloaded font: http://***/fontawesome-webfont.af7ae505a9eed503f8b8.woff2?v=4.7.0
OTS parsing error: invalid version tag

### Desired behavior
FontAwesome icons works fine

With cli 1.6.8 everything is OK

alan-agius4:
At the moment the uri of the font instead of spaced it will be `%20`, hence we need to decode it first before trying to resolve it.

Fixes #9648

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/styles2_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/styles2_spec_large.ts
@@ -8,7 +8,7 @@
 // tslint:disable:no-big-function
 
 import { Architect } from '@angular-devkit/architect/src/index2';
-import { normalize, tags } from '@angular-devkit/core';
+import { logging, normalize, tags } from '@angular-devkit/core';
 import { browserBuild, createArchitect, host } from '../utils';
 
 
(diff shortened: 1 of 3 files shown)


alan-agius4:
Targeting master only as it's won't apply cleanly on the patch branch due to the new specs.

packages/angular_devkit/build_angular/test/browser/styles2_spec_large.ts
```diff
@@ -575,4 +575,24 @@ describe('Browser Builder styles', () => {
...
+        body {
+          font-family: "Font Awesome";
+        }
+      `,
+      'src/assets/fa solid-900.woff2': '',
+    });
+
+    const overrides = { extractCss: true };
+    const { output } = await browserBuild(architect, host, target, overrides);
+    expect(output.success).toBe(true);
```
filipesilva: Is this expectation correct? From reading the issue it doesn't seem like this bug would cause the build to error out.

alan-agius4: At the moment it does cause a build error when there is a file with a space.

PR #13905, opened 2019-03-13
feat(@angular-devkit/build-angular): remove Closure compiler i18n code for ivy

ocombe:
For runtime i18n with ivy we generate code for both Closure compiler (used at Google) and for external people that don't use Closure compiler.
We added a new global flag named `ngI18nClosureMode` that allows uglify (and Closure compiler) to effectively tree-shake all of the code that isn't used in the current setup.
By default we remove all of the Closure compiler code because we assume that it won't be used.

Required for https://github.com/angular/angular/pull/28689.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -237,10 +237,11 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
       },
 
       // On server, we don't want to compress anything. We still set the ngDevMode = false for it
-      // to remove dev code.
+      // to remove dev code, and ngI18nClosureMode to remove Closure compiler i18n code
       compress: (buildOptions.platform == 'server' ? {
         global_defs: {
           ngDevMode: false,
+          ngI18nClosureMode: false,
         },
       } : {
           pure_getters: buildOptions.buildOptimizer,
@@ -249,6 +250,7 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
           passes: buildOptions.buildOptimizer ? 3 : 1,
           global_defs: {
             ngDevMode: false,
+            ngI18nClosureMode: false,
           },
         }),
       // We also want to avoid mangling on server.


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -237,10 +237,11 @@ export function getCommonConfig(wco: WebpackConfigOptions) {
       },
 
       // On server, we don't want to compress anything. We still set the ngDevMode = false for it
-      // to remove dev code.
+      // to remove dev code, and ngI18nClosureMode to remove Closure compiler i18n code
       compress: (buildOptions.platform == 'server' ? {
         global_defs: {
           ngDevMode: false,
+          ngI18nClosureMode: false,
```
filipesilva: What happens if this isn't added? Are bundles bigger or are they broken?

ocombe: just bigger

filipesilva: For how long is this flag expected to be around? Is it part of the necessary tooling for Ivy in general, or is it expected to not be necessary after a certain version?

mhevery: @filipesilva `ngI18nClosureMode` will be around for as long as we want to support `goog.getMsg`. The reason we need this is because we generate 

````
let greeting;
if (ngI18nClosureMode) {
  const MSG_greeting = goog.getMsg('Hello');
} else {
  greeting = localize`Hello`;
}
````

We need to generate the above because we need to ship the code in the above format to NPM so that developers which download a 3P library can choose to compile it with closure or without.

Come to think of it both `ngDevMode` and `ngI18nClosureMode` need to be defined only for the final application but NOT for library compilations. We want to make sure that when we ship things into NPM `ngDevMode` and `ngI18nClosureMode` are preserved in the source code, because it is only the application developer who can make a choice if they are building a debug build or production and if they are using `goog.getMsg` or something else. 

What is the current best practice for building an angular library? Would the above settings apply? If so can we have a different pass for libraries which do not set these values?

filipesilva: Libraries use a separate pipeline. This one is just for apps.

PR #13909, opened 2019-03-13
feat(@schematics/angular): allow ember, ember-cli, vendor, app and test names

hansl:
We dont care about the app name anymore.


hansl:
The e2e might fail, I couldn't find if we tested that creating apps with those names failed. If that's the case I'll just change those tests to pass (to avoid regressions).

sungdzda:
🎉

PR #13778, opened 2019-02-27
Dependency updates for Angular 8 and TS 3.3.3333

filipesilva:
Fix #13842

PR #13910, opened 2019-03-13
feat(@angular/cli): update to TypeScript 3.3

alexeagle:
This is the minimal version for Angular 8

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "quicktype-core": "^6.0.15",
     "temp": "^0.9.0",
     "tslint": "^5.13.1",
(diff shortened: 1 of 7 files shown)


alexeagle:
Oh looks like this was already in progress in #13778


@filipesilva should I close this one? Angular is on 3.3.3333 now so we are unblocked

filipesilva:
@alexeagle we'll also need a release from Angular and to update the versions here. Between this one and mine, however you prefer to do it really.

alan-agius4:
I'd prefer if we continue on the other PR, as there are some comments already.

PR #13914, opened 2019-03-14
test: pin Angular snapshot versions

alexeagle:
Next step is to set up Renovate to send PRs against this package.json file


alan-agius4:
Renovate should pick this automatically

PR #13919, opened 2019-03-14
fix(@angular-devkit/build-angular): passing port 0 when using serve w…

alan-agius4:
…ill not find a suitable port

At the moment, when passing port `0` it will default to `4200` which is incorrect. `4200` should be used a baseport.

This is the same as https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/dev-server/index2.ts#L127


clydin:
That entire function should technically be bypassed if the port is zero.  The OS knows how to properly handle allocation of the port and can do it far more efficiently than the manual method employed in the check port function.

alan-agius4:
@clydin, totally agree and I think it should be changed through the number of builders that are using it.
I'll try to fit that refactoring and removing the check port method in the coming days, before the RC.

PR #13926, opened 2019-03-15
feat(@angular-devkit/build-angular): deprecate `profile` option in build

alan-agius4:
This flag has been deprecated in favor of the `NG_BUILD_PROFILING`  environment variable. This is mainly due not to expose debugging flag in our API.


filipesilva:
LGTM but want to double check the flag name with @hansl

PR #13928, opened 2019-03-15
feat(@angular-devkit/build-webpack): report emitted files

alan-agius4:
With this change the builder will report the emitted chunks and assets after the compilation, this is needed for deferential loading so that we can build an index from the outputs of multiple builds

### DIFF
--- a/packages/angular_devkit/build_webpack/src/utils.ts
+++ b/packages/angular_devkit/build_webpack/src/utils.ts
@@ -1,0 +1,47 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as webpack from 'webpack';
+
+export interface EmittedFiles {
+  name?: string;
+  file: string;
+  initial: boolean;
+  extension: string;
+}
+
+export function getEmittedFiles(compilation: webpack.compilation.Compilation): EmittedFiles[] {
+  const getExtension = (file: string) => file.split('.').reverse()[0];
+  const files: EmittedFiles[] = [];
+
+  for (const chunk of Object.values(compilation.chunks)) {
+    const entry: Partial<EmittedFiles> = {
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_webpack/src/webpack/index2.ts
```diff
@@ -53,7 +57,10 @@ export function runWebpack(
         // Log stats.
         log(stats, config);
 
-        obs.next({ success: !stats.hasErrors() });
+        obs.next({
+          success: !stats.hasErrors(),
+          emittedFiles: getEmittedFiles(stats.compilation),
```
alan-agius4: I am not sure if this is what we want, in general I would prefer, if the architect has a separate channel similar to progress, that we can send events too.

Something like `reportStatus`, but something more generic events that builders could report, this approach would also be useful for starting a server/listening on a port in devserver.
Something along
````
reportEvent('emittedFiles', {data});
reportEvent('serverStarted', {address, port, host});
````

File: packages/angular_devkit/build_webpack/src/utils.ts [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as webpack from 'webpack';
+
+export interface EmittedFiles {
+  name?: string;
+  file: string;
+  isInitial: boolean;
```
clydin: Nit:  maybe just “initial” here?  Less webpack specific naming.

packages/angular_devkit/build_webpack/src/webpack/index2.ts
clydin: I agree.  Something similar would be beneficial.  We can move forward with this PR though and eventuallly deprecate as needed in the future when more extensive functionality is in place.

packages/angular_devkit/build_webpack/src/utils.ts
alan-agius4: Agree, done.

PR #13932, opened 2019-03-16
New Architect API server builder

hansl:


packages/angular_devkit/build_angular/src/server/index2.ts
```diff
@@ -0,0 +1,167 @@
...
+      normalize(root),
+      projectRoot,
+      sourceRoot,
+      options,
+    );
+
+    return { normalizedOptions, projectRoot };
+  }
+
+  // TODO: verify using of(null) to kickstart things is a pattern.
```
clydin: looks like an outdated comment

hansl: Removed.

PR #13942, opened 2019-03-19
fix(@angular/cli): add packageGroup for @angular/cli

### LINKED ISSUE #13581: ng update does not check or update @angular-devkit packages
I have angular CLI 7.3.0.  According to the release description [here](https://github.com/angular/angular-cli/releases/tag/v7.3.0), This release also involved releasing `@angular-devkit/build-angular 0.13.0`.  

As you can see in my screenshot below, I have @angular-devkit/build-angular, @angular-devkit/architect, @angular-devkit/build-optimizer, @angular-devkit/build-webpack all at version 0.10.2.

I then do an `ng update`, and the CLI tells me that everything is in order.  Why?  Shouldn't it tell me that I need to update all of the previously mentioned packages to 0.13.0?  

Or better yet, shouldn't those packages automatically update when I do an `ng update @angular/cli`?  That is the command I did to update to CLI 7.3.0, and the @angular-devkit packages did not update

screenshot

hansl:
This should add ng-update support to updating ALL the packages that are published from this repo.

Fixes #13581

### DIFF
--- a/packages/angular/cli/package.json
+++ b/packages/angular/cli/package.json
@@ -41,6 +41,19 @@
     "symbol-observable": "1.2.0"
   },
   "ng-update": {
-    "migrations": "@schematics/angular/migrations/migration-collection.json"
+    "migrations": "@schematics/angular/migrations/migration-collection.json",
+    "packageGroupName": "@angular/cli",
+    "packageGroup": {
+      "@angular-devkit/architect": "0.0.0",
+      "@angular-devkit/build-angular":  "0.0.0",
+      "@angular-devkit/build-ng-packagr": "0.0.0",
+      "@angular-devkit/build-optimizer":  "0.0.0",
+      "@angular-devkit/build-webpack": "0.0.0",
+      "@angular-devkit/core":  "0.0.0",
+      "@angular-devkit/schematics": "0.0.0",
+      "@ngtools/webpack": "0.0.0",
+      "@schematics/angular":  "0.0.0",
+      "@schematics/update":  "0.0.0"
+    }
   }
 }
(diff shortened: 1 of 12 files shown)


packages/angular/cli/package.json
```diff
@@ -41,6 +41,9 @@
     "symbol-observable": "1.2.0"
   },
   "ng-update": {
-    "migrations": "@schematics/angular/migrations/migration-collection.json"
+    "migrations": "@schematics/angular/migrations/migration-collection.json",
+    "packageGroup": {
+      "@angular/cli": "0.0.0"
```
alan-agius4: Shouldn’t this package group contain the packages to updates? Aka for the cli shouldn’t it be @angular-devkit/build-angular and @angular-devkit/ng-packagr?

hansl: It's all done in the `build.ts` actually.

hansl: I'll remove this line. it's basically useless.

scripts/build.ts
```diff
@@ -350,48 +423,10 @@ export default async function(
...
-            obj[depName] = (obj[depName] as string).replace(/\b0\.0\.0\b/, v);
-          }
-        }
-      }
-    }
-
-    fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2) + '\n');
-  }
+  logger.info('Setting packageGroups...');
+  await _setPackageGroups(sortedPackages, logger.createChild('packageGroups'));
```
clydin: Won't this add package groups to all the dev kit packages each with all the other dev kit packages included?

hansl: I'll change it a bit.

hansl: Done, I added a check instead that build will add package group packages if they're in the package.json already, and added all packages that I thought were okay.

hansl:
@clydin @alan-agius4 please take another look.

alexeagle:
@clydin is this obsolete now that you made a smaller-scope change?

alexeagle:
Charles took it over

PR #13948, opened 2019-03-19
Support `import()` syntax for loadChildren in ViewEngine

filipesilva:
This feature ONLY matches the format below:
```
loadChildren: () => import('IMPORT_STRING').then(m => m.EXPORT_NAME)
```

It will not match nor alter variations, for instance:
- not using arrow functions
- not using `m` as the module argument
- using `await` instead of `then`
- using a default export (https://github.com/angular/angular/issues/11402)

The only parts that can change are the ones in caps: IMPORT_STRING and EXPORT_NAME.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,6 +106,16 @@ jobs:
           path: /tmp/dist
           destination: cli/new-production
 
+  e2e-cli-ivy:
+    <<: *defaults
+    environment:
+      BASH_ENV: ~/.profile
+    resource_class: xlarge
+    parallelism: 4
+    steps:
+      - attach_workspace: *attach_options
+      - run: xvfb-run -a node ./tests/legacy-cli/run_e2e --nb-shards=${CIRCLE_NODE_TOTAL} --shard=${CIRCLE_NODE_INDEX} --ivy          
+
   e2e-cli-ng-snapshots:
     <<: *defaults
     environment:
@@ -122,23 +132,6 @@ jobs:
             fi
       - attach_workspace: *attach_options
       - run: xvfb-run -a node ./tests/legacy-cli/run_e2e --nb-shards=${CIRCLE_NODE_TOTAL} --shard=${CIRCLE_NODE_INDEX} --ng-snapshots
-
-  e2e-cli-ivy-snapshots:
-    <<: *defaults
-    environment:
-      BASH_ENV: ~/.profile
-    resource_class: xlarge
-    parallelism: 4
-    steps:
-      - run:
-          name: Don't run expensive e2e tests for forks other than renovate-bot and angular
-          command: >
-            if [[ "$CIRCLE_PR_USERNAME" != "renovate-bot" ]] &&
-               [[ "$CIRCLE_PROJECT_USERNAME" != "angular" || $CIRCLE_BRANCH != "master" ]]; then
-              circleci step halt
-            fi
-      - attach_workspace: *attach_options
-      - run: xvfb-run -a node ./tests/legacy-cli/run_e2e --nb-shards=${CIRCLE_NODE_TOTAL} --shard=${CIRCLE_NODE_INDEX} --ng-snapshots --ivy
 
   build:
     <<: *defaults
@@ -210,6 +203,9 @@ workflows:
       - e2e-cli:
           requires:
             - build
+      - e2e-cli-ivy:
+          requires:
+            - build            
       - snapshot_publish_docs:
           requires:
           - install
@@ -218,9 +214,6 @@ workflows:
               only:
               - /docs-preview/
       - e2e-cli-ng-snapshots:
-          requires:
-            - build
-      - e2e-cli-ivy-snapshots:
           requires:
             - build
       - snapshot_publish:
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -760,39 +760,36 @@ export class AngularCompilerPlugin {
             return result;
           }
 
-          return this.done.then(
-            () => {
-              // This folder does not exist, but we need to give webpack a resource.
-              // TODO: check if we can't just leave it as is (angularCoreModuleDir).
-              result.resource = path.join(this._basePath, '$$_lazy_route_resource');
-              // tslint:disable-next-line:no-any
-              result.dependencies.forEach((d: any) => d.critical = false);
-              // tslint:disable-next-line:no-any
-              result.resolveDependencies = (_fs: any, options: any, callback: Callback) => {
-                const dependencies = Object.keys(this._lazyRoutes)
-                  .map((key) => {
-                    const modulePath = this._lazyRoutes[key];
-                    if (modulePath !== null) {
-                      const name = key.split('#')[0];
+          await this.done;
 
-                      return new this._contextElementDependencyConstructor(modulePath, name);
-                    } else {
-                      return null;
-                    }
-                  })
-                  .filter(x => !!x);
+          // This folder does not exist, but we need to give webpack a resource.
+          // TODO: check if we can't just leave it as is (angularCoreModuleDir).
+          result.resource = path.join(this._basePath, '$$_lazy_route_resource');
+          // tslint:disable-next-line:no-any
+          result.dependencies.forEach((d: any) => d.critical = false);
+          // tslint:disable-next-line:no-any
+          result.resolveDependencies = (_fs: any, options: any, callback: Callback) => {
+            const dependencies = Object.keys(this._lazyRoutes)
+              .map((key) => {
+                const modulePath = this._lazyRoutes[key];
+                if (modulePath !== null) {
+                  const name = key.split('#')[0];
 
-                if (this._options.nameLazyFiles) {
-                  options.chunkName = '[request]';
+                  return new this._contextElementDependencyConstructor(modulePath, name);
+                } else {
+                  return null;
                 }
+              })
+              .filter(x => !!x);
 
-                callback(null, dependencies);
-              };
+            if (this._options.nameLazyFiles) {
+              options.chunkName = '[request]';
+            }
 
-              return result;
-            },
-            () => undefined,
-          );
+            callback(null, dependencies);
+          };
+
+          return result;
         });
       });
     }
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -46,6 +46,7 @@ import {
   exportLazyModuleMap,
   exportNgFactory,
   findResources,
+  importFactory,
   registerLocaleData,
   removeDecorators,
   replaceBootstrap,
@@ -108,6 +109,7 @@ export interface AngularCompilerPluginOptions {
   // When using Ivy, the string syntax is not supported at all. Thus we shouldn't attempt that.
   // This option is also used for when the compilation doesn't need this sort of processing at all.
   discoverLazyRoutes?: boolean;
+  importFactories?: boolean;
 
   // added to the list of lazy routes
   additionalLazyModules?: { [module: string]: string };
@@ -140,6 +142,7 @@ export class AngularCompilerPlugin {
   private _moduleResolutionCache: ts.ModuleResolutionCache;
   private _resourceLoader?: WebpackResourceLoader;
   private _discoverLazyRoutes = true;
+  private _importFactories = false;
   // Contains `moduleImportPath#exportName` => `fullModulePath`.
   private _lazyRoutes: LazyRouteMap = {};
   private _tsConfigPath: string;
@@ -298,7 +301,12 @@ export class AngularCompilerPlugin {
       this._platformTransformers = options.platformTransformers;
     }
 
-    if (options.discoverLazyRoutes !== undefined) {
+    // Determine if lazy route discovery via Compiler CLI private API should be attempted.
+    if (this._compilerOptions.enableIvy) {
+      // Never try to discover lazy routes with Ivy.
+      this._discoverLazyRoutes = false;
+    } else if (options.discoverLazyRoutes !== undefined) {
+      // The default is to discover routes, but it can be overriden.
       this._discoverLazyRoutes = options.discoverLazyRoutes;
     }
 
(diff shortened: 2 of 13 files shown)


filipesilva:
Blocked on https://github.com/angular/angular/pull/29392

packages/ngtools/webpack/src/transformers/import_factory.ts
```diff
@@ -0,0 +1,120 @@
...
+  // loadChildren: something
+  const name = node.name;
+  if (!(
+    ts.isIdentifier(name)
+    && name.text == 'loadChildren'
+  )) {
+    return node;
+  }
+
+  // loadChildren: () => something
```
filipesilva: Maybe we add a warning if we bail past this point. Something like
```
Found `loadChildren` with a non-string syntax in FILENAME_GOES_HERE but could not transform it. 

Make sure it matches the format below:

loadChildren: () => import('IMPORT_STRING').then(m => m.EXPORT_NAME)

Please note that only IMPORT_STRING and EXPORT_NAME can be replaced in this format.
```

filipesilva: Added warning.

tests/legacy-cli/e2e/tests/build/lazy-load-syntax.ts
```diff
@@ -69,22 +90,42 @@ export default async function () {
...
+  // Test `import()` style lazy load.
+  await updateJsonFile('angular.json', json => {
+    // Add the experimental flag to import factories in View Engine.
+    const buildTarget = json['projects'][projectName]['architect']['build'];
+    buildTarget['options']['experimentalImportFactories'] = true;
+  });
+  // Both Ivy and View Engine should support it.
+  await replaceLoadChildren(`() => import('./lazy/lazy.module').then(m => m.LazyModule)`);
+  await ng('e2e');
+  await ng('e2e', '--prod');
```
filipesilva: This test will fail because this PR isn't using a Angular release that contains https://github.com/angular/angular/pull/29392. Once that is released this PR should be updated to use it.

packages/ngtools/webpack/src/transformers/import_factory.ts
```diff
@@ -0,0 +1,201 @@
...
+
+Please note that only IMPORT_STRING and EXPORT_NAME can be replaced in this format.`;
+        const emitWarning = () => warningCb(warning);
+        const visitNode: ts.Visitor = (node: ts.Node) => {
+          // Only try to transform 'loadChildren' property assignments.
+          if (ts.isVariableDeclaration(node)) {
+            return replaceImport(node, context, emitWarning);
+          }
+
+          return ts.visitEachChild(node, visitNode, context);
```
clydin: If ngc always creates the variable declaration at the root level then only the source file's children should need to be analyzed.  Walking the entire tree can be avoided.  The regex test above can be avoided as well since top level nodes are typically fairly minimal and testing a large block of text will be much less efficient.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -103,6 +104,8 @@ export interface AngularCompilerPluginOptions {
   nameLazyFiles?: boolean;
   logger?: logging.Logger;
   directTemplateLoading?: boolean;
+  discoverLazyRoutes?: boolean;
```
alan-agius4: Can you kindly explain the reasoning behind this option?

filipesilva: When using the `loadChildren` string syntax, `@ngtools/webpack` must query `@angular/compiler-cli` via a private API to know which lazy routes exist. This increases build and rebuild time.

When using Ivy, the string syntax is not supported at all. Thus we shouldn't attempt that. 

This commit also exists in https://github.com/angular/angular-cli/pull/13700 for slightly different reasons. There, it is meant to disable this sort of processing for compilations that do not need it.

alan-agius4: Thanks for the clarification, it makes more sense now :)

clydin: Should we warn (or probably error) in the transformer (or somewhere else) if the string form is found when this option is off?

Users may not know why their applications are not working.

packages/ngtools/webpack/src/transformers/import_factory.ts
```diff
@@ -0,0 +1,201 @@
...
+Found 'loadChildren' with a non-string syntax in ${sourceFile.fileName} but could not transform it.
+Make sure it matches the format below:
+
+loadChildren: () => import('IMPORT_STRING').then(m => m.EXPORT_NAME)
+
+Please note that only IMPORT_STRING and EXPORT_NAME can be replaced in this format.`;
+        const emitWarning = () => warningCb(warning);
+        const visitNode: ts.Visitor = (node: ts.Node) => {
+          // Only try to transform 'loadChildren' property assignments.
+          if (ts.isVariableDeclaration(node)) {
```
clydin: Does ngc generate these for anything else?  Should we verify that the declaration is assigned to `loadChildren`?

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -922,27 +947,29 @@ export class AngularCompilerPlugin {
...
 
-    // Find lazy routes
-    lazyRouteMap = {
-      ...lazyRouteMap,
-      ...this._options.additionalLazyModules,
-    };
+      // Find lazy routes
+      lazyRouteMap = {
+        ...lazyRouteMap,
+        ...this._options.additionalLazyModules,
```
clydin: Should it still support the additional lazy modules?

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -697,64 +716,66 @@ export class AngularCompilerPlugin {
...
+              this.options.additionalLazyModuleResources.includes(resource));
+
+          if (!result || !this.done || !isLazyModuleResource(result.resource)) {
             return result;
-          },
-          () => undefined,
-        );
+          }
+
+          return this.done.then(
```
alan-agius4: Nit: we can use await here to reduce a bit the indentations and nesting.

packages/ngtools/webpack/src/transformers/import_factory.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  const initializer = node.initializer;
+  if (initializer === undefined) {
+    return node;
+  }
+
+  // ɵ0 = () => something
+  if (!(
+    ts.isArrowFunction(initializer)
```
alan-agius4: What about if target is es5?

packages/ngtools/webpack/src/transformers/import_factory_spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      const transformer = importFactory(() => warningCalled = true);
+      const result = transformTypescript(input, [transformer]);
+
+      expect(tags.oneLine`${result}`).toEqual(tags.oneLine`${output}`);
+      expect(warningCalled).toBeFalsy();
+    });
+
+    it('should not transform if the format is different than expected', () => {
+      const input = tags.stripIndent`
+        const ɵ0 = function () { return import("./lazy/lazy.module").then(m => m.LazyModule); };
```
alan-agius4: Any reason why this is not supported, ie when targeting es5?

packages/ngtools/webpack/src/transformers/import_factory.ts
filipesilva: The target does not make a difference. This transform operates over the result from NGC and before any downleveling done by TS proper.

packages/ngtools/webpack/src/transformers/import_factory_spec.ts
filipesilva: Supporting this format would introduce extra complexity without much benefit. This functionality is meant to exclusively support the usecase of switching back and forth between Ivy and View Engine. It is not meant to generally support this syntax.

In regards to the target, the answer is the same as the question above.

clydin: For reference: https://github.com/angular/angular-cli/blob/2d7022fee5418594c8d7d2c86e830f1984bfed19/packages/ngtools/webpack/src/angular_compiler_plugin.ts#L1213

packages/ngtools/webpack/src/transformers/import_factory.ts
filipesilva: AFAIK ngc generates these `ɵ0 = something` nodes when removing functions from decorators so they can be exported to factories. Which means the current warning would be emitted for other things that we don't want. 

To help prevent this I moved the warning to only show when `ɵ0 = () => import('IMPORT_STRING').then(something)` is found onwards.

Before I figured out this transform only got the NGC transformed code I actually verified that it was on a `loadChildren`. But then came the indirection of `const ɵ0` which made that harder.

I tried to get from the variable declaration to its references with the type checker and couldn't find a good way. I could save the reference and then further visit all other nodes (like we do in some other transforms) but that would not be very performant.

 Since we were already aiming for support on a very specific use case (switching back between Ivy and VE and with a experimental flag) I thought it wouldn't be worth the work to make it much more robust.

I think that warning later in the structure reduces false negatives (warning about something that isn't a loadChildren import). But I do agree false positives (altering something that isn't a loadChildren import) are a problem. Do you have any suggestion on how to verify this in an efficient manner?

packages/ngtools/webpack/src/transformers/import_factory.ts
filipesilva: That is a better approach, yes. I modified the code to only recurse into `VariableStatement` to find `VariableDeclaration` within, since the structure is `VariableStatement -> VariableDeclarationList -> VariableDeclaration`. PTAL

packages/ngtools/webpack/src/angular_compiler_plugin.ts
filipesilva: Refactored to use await.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
filipesilva: I don't think so. We do currently support this for VE. But it hinges on `SystemJsNgModuleLoaderConfig`, which will be deprecated for Ivy (https://github.com/angular/angular/pull/28685#issuecomment-473387576). So I think that for Ivy projects this functionality shouldn't work at all.

This in turn means we should warn that it's not working. Added a warning for that.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
filipesilva: I think not because that means we need to always check, which incurs a performance penalty on every AOT build.

The "right way" of doing that probably a Ivy migration when it's stable.

packages/ngtools/webpack/src/transformers/import_factory.ts
```diff
@@ -0,0 +1,198 @@
...
+ * constructs, like `class` and arrow functions, still remain.
+ *
+ */
+
+export function importFactory(
+  warningCb: (warning: string) => void,
+): ts.TransformerFactory<ts.SourceFile> {
+  return (context: ts.TransformationContext) => {
+    return (sourceFile: ts.SourceFile) => {
+      const warning = `
```
alan-agius4: Should this be a warning or an error, since it will cause a runtime error?

tests/legacy-cli/e2e/tests/build/lazy-load-syntax.ts
```diff
@@ -5,17 +5,36 @@
...
+import { appendToFile, prependToFile, readFile, replaceInFile, writeFile } from '../../utils/fs';
 import { ng } from '../../utils/process';
-import { createProject, updateJsonFile } from '../../utils/project';
+import { updateJsonFile } from '../../utils/project';
 import { expectToFail } from '../../utils/utils';
 
 export default async function () {
-  const projectName = 'ivy-lazy-loading';
+  const argv = getGlobalVariable('argv');
+  const ivyProject = argv['ivy'];
```
alan-agius4: How are these run during the e2e-cli tests?

Isn't the `ivy` arg used only in the cli snapshots?

filipesilva: Yes, Ivy tests are only run for snapshots now. I agree with you that we would love coverage. I think moving the ivy job to not use the snapshots, and to always run, is a good approach. We're starting to have more cde that interacts with Ivy so we should have confidence in PRs for it.

packages/ngtools/webpack/src/transformers/import_factory.ts
filipesilva: It won't necessarily cause a runtime error. VE can still use both types of imports.

.circleci/config.yml
```diff
@@ -106,24 +106,17 @@ jobs:
           path: /tmp/dist
           destination: cli/new-production
 
-  e2e-cli-ng-snapshots:
+  e2e-cli-ivy:
```
alexeagle: nit: why switch the order of these? makes the delta harder to review

packages/ngtools/webpack/src/angular_compiler_plugin.ts
alexeagle: let's try to generally answer questions like this in the form of a comment in the code - no one later will find this thread

packages/ngtools/webpack/src/transformers/import_factory.ts
```diff
@@ -0,0 +1,198 @@
...
+Found 'loadChildren' with a non-string syntax in ${sourceFile.fileName} but could not transform it.
+Make sure it matches the format below:
+
+loadChildren: () => import('IMPORT_STRING').then(m => m.EXPORT_NAME)
+
+Please note that only IMPORT_STRING and EXPORT_NAME can be replaced in this format.`;
```
alexeagle: link to the guide again, if ppl get confused we'll be able to help them there

.circleci/config.yml
filipesilva: To keep similar e2e jobs (`e2e-cli` and `e2e-cli-ivy`) together.  

The old `e2e-cli-ng-snapshots` was after `e2e-cli-ivy-snapshots` because they shared the snapshot flag and only ran on some builds. Now `e2e-cli-ivy-snapshots` is instead `e2e-cli-ivy` and should instead be next to `e2e-cli`.

filipesilva: Added.

packages/ngtools/webpack/src/transformers/import_factory.ts
filipesilva: Done

packages/ngtools/webpack/src/angular_compiler_plugin.ts
filipesilva: Done.

PR #13949, opened 2019-03-19
feat(@ngtools/webpack): update `require` statements to work with webp…

alan-agius4:
…ack raw-loader v2.

webpack raw-loader released a breaking change https://github.com/webpack-contrib/raw-loader/releases/tag/v2.0.0 and now they use ES Module export instead of CommonJS.

BREAKING CHANGE: we now support only raw-loader version 2+ as the emitted code has changed to support version 2 of the mentioned loader.

Before
```
Component({
    selector: 'app-root',
    template: require("!raw-loader!./app.component.html"),
    styles: [require("./app.component.css")]
})
```

Now
```
Component({
    selector: 'app-root',
    template: require("!raw-loader!./app.component.html").default,
    styles: [require("./app.component.css").default]
})
```

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -32,7 +32,7 @@
     "postcss": "7.0.14",
     "postcss-import": "12.0.1",
     "postcss-loader": "3.0.0",
-    "raw-loader": "1.0.0",
+    "raw-loader": "2.0.0",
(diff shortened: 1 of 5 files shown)


JounQin:
@alan-agius4 Any update on this?

alan-agius4:
superseded by https://github.com/angular/angular-cli/pull/15311

PR #13950, opened 2019-03-19
feat(@angular-devkit/build-angular): use `sass` instead of `node-sass`

alan-agius4:
BREAKING CHANGE: we no longer use `node-sass` by default, instead we use `sass` (dart-sass).

While in most cases this should not have an impact the generated code, dart sass has some differences from Ruby Sass. See https://github.com/sass/dart-sass#behavioral-differences-from-ruby-sass for more details

If you wish to use `node-sass` instead of `sass` you need to install node-sass as a dev dependency.

npm users
```
npm install node-sass --save-dev
```

yarn users
```
yarn add node-sass --dev
```

**Note**: for larger Sass projects to avoid this performance hit you might want to install `fibers`. See: https://www.npmjs.com/package/fibers

Ref: TOOL-748

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -34,6 +34,7 @@
     "postcss-loader": "3.0.0",
     "raw-loader": "1.0.0",
     "rxjs": "6.4.0",
+    "sass": "1.17.3",
     "sass-loader": "7.1.0",
     "semver": "5.6.0",
     "source-map-support": "0.5.11",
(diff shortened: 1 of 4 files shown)


alan-agius4:
I have some doubts re the console msg, as we'll have too many console infos when starting the dev server and 

1) The more we add the more people will start ignored them
2) They get invasive as there is no way to turn these off at the moment.

alan-agius4:
@alexeagle what are your thoughts about the above? So we still do the console msg?

PR #13953, opened 2019-03-19
Revert "refactor: deprecate protractor baseUrl option from builder"

### LINKED ISSUE #13952: `baseUrl` flag in protractors schema is deprecated
# 🚀 Feature request

Please do not deprecate and remove the `baseUrl` flag in the protractor schema.

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ x ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
In our CI/CD system, we run our e2e tests as a part of our pre-merge validations, which include deploying the SPA configured to run against our staging API. In order to do this cleanly, and without namespace conflicts/races, we give each of these applications their own base-href, as follows:

```
ng build
    --configuration=staging
    --output-path=./dist-e2e
    --base-href=/ci-cd-draft-build-${JOB_NUMBER}/
```

This is then uploaded to S3, and tested as follows:
```
ng e2e --baseUrl=https://<s3-bucket>/ci-cd-draft-build-${JOB_NUMBER}/
```
We also use these draft builds to perform Beta feature usability tests, product demos, and a variety of other activities that are greatly enhanced by having a fully functioning UI. Similarly, we publish whitelabel UI's for our customers, which we also run protractor-driven acceptance tests against, and whose base-url's vary greatly.

Just now, I noticed this message:
```
Option "baseUrl" is deprecated: Use "baseUrl" in the Protractor config file instead.
```
Which I traced to the following entry in the protractor schematic.
```
    "baseUrl": {
      "type": "string",
      "description": "Base URL for protractor to connect to.",
      "x-deprecated": "Use \"baseUrl\" in the Protractor config file instead."
    }
```
As our `buildUrl` value is not known ahead of time, we cannot easily use the protractor config file; while we could attempt to inject it via some form of environment parameter, modifying environment parameters is more cumbersome than adding a commandline flag.

We do not want this flag to be removed. It's not harming anyone, why is it being removed?

Writing a configuration file for CI/CD, for every customer, environment, and build target we have. This will litter our project with config files, and we'd really like to avoid that.

krotscheck:
This reverts commit 5415bcc6f872b181b6b05b2a5c718a9146ea3113.

Right now, this is flagged as "Deprecated" in the CLI, which isn't actually the case. As per the new protractor implementation in https://github.com/angular/angular-cli/commit/a49d278baa8561a39566dffd5c947bcc79b485c0 , `baseUrl` is supported as long as `devServerTarget` is not included.

This patch reverts the deprecation note.

Resolves #13952

### DIFF
--- a/packages/angular_devkit/build_angular/src/protractor/schema.json
+++ b/packages/angular_devkit/build_angular/src/protractor/schema.json
@@ -46,8 +46,7 @@
     },
     "baseUrl": {
       "type": "string",
-      "description": "Base URL for protractor to connect to.",
-      "x-deprecated": "Use \"baseUrl\" in the Protractor config file instead."
+      "description": "Base URL for protractor to connect to."
     }
   },
   "additionalProperties": false,


alan-agius4:
Thanks for this, however kindly see https://github.com/angular/angular-cli/issues/13952#issuecomment-474713484

PR #13968, opened 2019-03-22
feat(@ngtools/webpack): add NGCC as part of the workflow

alan-agius4:
- feat(@schematics/angular): remove ivy-ngcc postinstall script from new application
- refactor(@ngtools/webpack): move interfaces to separate file
This is to avoid circular imports

- feat(@ngtools/webpack): add NGCC as part of the workflow  …
When add module is resolved, it will try to convert the module using the `NGCC` API.

- test: add e2e test for ivy opt-out

- refactor(@angular-devkit/build-angular): use `@ngtools/webpack` getMainFields helper


alan-agius4:
Blocked until the framework release the API in the beta version.

packages/ngtools/webpack/src/utils_spec.ts
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { CompilerOptions } from '@angular/compiler-cli';
+import * as ts from 'typescript';
+import { PLATFORM } from './interfaces';
+import { getMainFields } from './utils';
+
+describe('@ngtools/webpack utils', () => {
```
filipesilva: Loving the comprehensive tests 👍

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/server.ts
```diff
@@ -24,7 +25,7 @@ export function getServerConfig(wco: WebpackConfigOptions) {
 
   const config: Configuration = {
     resolve: {
-      mainFields: [...(wco.supportES2015 ? ['es2015'] : []), 'main', 'module'],
+      mainFields: getMainFields(wco.tsConfigPath, PLATFORM.Browser),
```
filipesilva: This looks like it should be `PLATFORM.Server`

tests/legacy-cli/e2e/tests/ivy/ivy-opt-out.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { replaceInFile } from '../../utils/fs';
+import { request } from '../../utils/http';
+import { killAllProcesses, ng } from '../../utils/process';
+import { createProject } from '../../utils/project';
+
+export default async function() {
```
filipesilva: Very nice test to have. Will need to enhance it later with any further opt-out actions.

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  constructor(
+    private readonly propertiesToConsider: string[],
+    private readonly host: WebpackCompilerHost,
+  ) {
+    // this is done for the sole reason that @ngtools/webpack
+    // support versions of Angular that don't have NGCC API
+    try {
+      this._ngcc = require('@angular/compiler-cli/ngcc');
+    } catch {
```
filipesilva: Nit: I find this usage pattern a bit odd. So if the require fails, there's a `NgccProcessor` whose `processModule` is a noop. I'd expect there to not be a `NgccProcessor` at all.

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  processModule(
```
filipesilva: @petebacondarwin PTAL at our API usage of ngcc.

packages/ngtools/webpack/src/compiler_host.ts
```diff
@@ -31,6 +34,7 @@ export class WebpackCompilerHost implements ts.CompilerHost {
   private _memoryHost: virtualFs.SyncDelegateHost;
   private _changedFiles = new Set<string>();
   private _basePath: Path;
+  private _ngccProcessor: NgccProcessor;
```
filipesilva: Can you add some context here? 

Our previous approach was to make a standalone webpack plugin for `ngcc`. But we had to change that plan because the TS compilation itself (which happens before webpack bundling) could fail when looking up module typings, since the `.d.ts` files are also modified by `ngcc`. 

So now we hook into the TS module resolution and process them there.

Now that I think about this a bit more it begs the question: should `@angular/compiler-cli` do this directly? Is there some fundamental flaw with this approach we are missing, is it a matter of performance (since we don't do all entry points here), or is it just that we hadn't considered it before?

cc @petebacondarwin

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const format = this.propertiesToConsider.find(mf => typeof packageJson[mf] === 'string');
+    if (!format || this._ngcc.hasBeenProcessed(packageJson, format)) {
+      // Unknown format or module has already been processed by NGCC.
+      this._processedModules.add(moduleName);
+
+      return;
+    }
+
+    const timeLabel = `NgccProcessor.processModule.ngcc.process+${moduleName}`;
```
filipesilva: 👍 we can now use this to diagnose perf.

packages/ngtools/webpack/src/compiler_host.ts
clydin: Are the d.ts files even relevant here though?  The only TS files should be application code and that shouldn’t be ivy specific.

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  processModule(
+    moduleName: string,
+    resolvedModule: ts.ResolvedModule | ts.ResolvedTypeReferenceDirective,
+  ): void {
+    const resolvedFileName = resolvedModule.resolvedFileName;
+    if (
+      !this._ngcc
+      || !resolvedFileName
+      || moduleName.startsWith('.')
```
petebacondarwin: what about `startsWith('/')`?

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -0,0 +1,103 @@
...
+      || this._processedModules.has(moduleName)) {
+      // Skip when module is unknown, relative or NGCC compiler is not found or already processed.
+      return;
+    }
+
+    const packageJsonPath = this.tryResolvePackage(moduleName, resolvedFileName);
+    if (!packageJsonPath) {
+      // add it to processed to the second time round we skip this
```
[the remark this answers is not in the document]
petebacondarwin: Oh! `to` -> `so` :-)

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -0,0 +1,103 @@
...
+      throw new Error(`Cannot parse '${packageJsonPath}' as JSON.`);
+    }
+
+    const format = this.propertiesToConsider.find(mf => typeof packageJson[mf] === 'string');
+    if (!format || this._ngcc.hasBeenProcessed(packageJson, format)) {
+      // Unknown format or module has already been processed by NGCC.
+      this._processedModules.add(moduleName);
+
+      return;
+    }
```
petebacondarwin: If you are going to have to load the packageJson anyway then there is not much performance benefit in not just calling ngcc everytime. Ngcc will still check the build markers and also noop if the format is not recognized.

So you could simplify this whole block to just call ngcc with the `packageJsonPath`.

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const timeLabel = `NgccProcessor.processModule.ngcc.process+${moduleName}`;
+    time(timeLabel);
+    this._ngcc.process({
+      basePath: packageJsonPath.substring(0, packageJsonPath.indexOf(moduleName)),
```
petebacondarwin: Nice

packages/ngtools/webpack/src/compiler_host.ts
alan-agius4: Dts files from external modules needs to be transformed for a ngtsc (Ivy) compilation.

Ngcc should only be run on external modules to transform dts and js files.

clydin: Yes. but what TS code is being type checked that actually needs those d.ts files changed?

clydin: Further, if the ivy compiler itself needs additional type information, can it not just do the analysis as needed itself?

alan-agius4: @petebacondarwin, can you chime in?

packages/ngtools/webpack/src/ngcc_processor.ts
alan-agius4: That's actually a good point, we should't process absolute modules, that said, `/` alone is not enough for Windows users.

packages/ngtools/webpack/src/ngcc_processor.ts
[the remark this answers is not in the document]
alan-agius4: Hahaha yeah thanks for pointing this out :)

packages/ngtools/webpack/src/ngcc_processor.ts
alan-agius4: Updating with my next push.

packages/ngtools/webpack/src/compiler_host.ts
alan-agius4: Following the NGCC sync meeting, This current implementation is the preferred solution and NGTSC should not be aware of the existence of NGCC and these two compilation should be kept separate.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -660,13 +618,34 @@ export class AngularCompilerPlugin {
...
+        // this is done for the sole reason that @ngtools/webpack
+        // support versions of Angular that don't have NGCC API
+        if (this._compilerOptions.enableIvy) {
+          ngcc = require('@angular/compiler-cli/ngcc');
+        }
+      } catch {
+      }
+
+      if (ngcc) {
+        const mainFields = getMainFields(this._compilerOptions, this._platform, false);
```
clydin: This whole block can be moved into the `if (this._compilerOptions.enableIvy)` block above and eliminate the extra conditional as well as one less local variable in the function's scope.

packages/ngtools/webpack/src/compiler_host.ts
```diff
@@ -354,13 +357,46 @@ export class WebpackCompilerHost implements ts.CompilerHost {
...
     console.log(message);
   }
-}
 
+  resolveModuleNames(
+    moduleNames: string[],
+    containingFile: string,
+  ): (ts.ResolvedModule | undefined)[] {
+    return moduleNames.map(moduleName => {
+      const { resolvedModule } = ts.resolveModuleName(
```
clydin: Thinking about this more.  We are essentially resolving modules using two completely different code paths in a single compilation.   Transpilation is using the TS built-in resolution and bundling is using webpack's.  This seems particularly error prone and fragile.  While they ideally should resolve to the same place, there is no guarantee; especially when developers are including this plugin in a custom webpack configuration (a use case which should be fully supported).

Should this be deferring to webpack's module resolution for consistency?

alexeagle:
@alan-agius4 could you include the reasoning for not having ngtsc be in change of running ngcc, that you mention from the sync meeting?

Discussed in standup this morning and @kara and @mhevery think it's an interesting idea to run it this way

alan-agius4:
@alexeagle the main reasoning that I got for not doing this was not to make ngtsc dependent on ngcc.

Though maybe there is more to it @petebacondarwin & @alxhub would know more. 

Also this won’t be transparent for us, as ngtsc will need to return the packages that have been transformed by ngcc so that we can purge them from cache.

petebacondarwin:
This PR *is* effectively getting ngtsc to trigger ngcc.

Already the CLI is providing a specially configured TypeScript `CompilerHost` to ngtsc, which it then uses to do its processing.

Here we are adding additional customisation to this `CompilerHost` that also triggers ngcc when module resolution is taking place. This is the right place to do this processing, and is also the best API for customising the ngtsc compilation behaviour.

The alternative of adding some kind of flag to the `CompilerOptions`, which then requires ngtsc to add this ngcc processing into the CompilerHost that it is handed is not ideal:

1) It makes ngtsc aware of ngcc, which is not desirable and makes ngtsc harder to maintain (and test).
2) It would require ngtsc to have significant additional code that would have to override the `CompilerHost` that is passed to it, which could create conflicts of interest.
3) It adds bloat to ngtsc for a use case which is specific to CLI (and would need to be removed if/when ngcc is no longer needed.

filipesilva:
Not to disagree with the other points, but running ngcc transparently and only when needed isn't a use case that is specific to the CLI. It would in fact be the better way of running ngcc overall.

filipesilva:
Thinking about it some more I also do not understand the maintenance argument. `ngcc` is meant to be used with `ngtsc` as part of `@angular/compiler-cli` exposed in the public API, thus subject to current deprecation policies. It will be around for a while.

As it stands the maintenance cost for `ngcc` is distributed between `@angular/compiler-cli` and projects/tooling that consume it, either via the post install hook or this Angular CLI hook. You might regard the postinstall hook as not requiring maintenance, but it does: it must be added and removed manually, and will not work outside certain version ranges of `@angular/compiler-cli`.

But any usage is completely bound to the API exposed by `@angular/compiler-cli`. So if the API in `@angular/compiler-cli` changes then consumers must support both new and old API, as projects can be using either.

This is to say that confining maintenance cost to `@angular/compiler-cli` is preferable as it reduces total maintenance and requires no updates in other projects/tooling efforts.

petebacondarwin:
Thanks for your input @filipesilva. What you say makes a lot of sense.

There is definitely value in avoiding significant external configuration of ngtsc for build tools like webpack. We designed ngcc as a standalone tool in the first place to avoid having to publish multiple integrations with build tools (of which webpack is one).

I think the concern is that each integration might need to configure things (like the `CompilerHost`) slightly differently, and so it would be better to let them do so directly rather than trying to apply a one-size-fits-all feature in ngtsc.

My understanding is that the life of ngcc is only until we have moved to ivy as the default, at which point we would expect libraries to start publishing their distributables compiled for ivy directly.

Despite all of this, I wonder if it might be worth publishing a wrapper around ngc/ngtsc, which basically does the customisation of the `CompilerHost` similar to what we are doing here. Tools could then choose to use this wrapper rather than the "standard" `ngc` that we have right now.

alxhub:
I think combining the two tools is the wrong approach, because their missions and requirements are so different.

`ngtsc` is a thin wrapper around the TypeScript compiler. It has one goal: to turn your `.ts` files into `.js` files. It tries to do this as efficiently as possible, independently of the rest of the build system, and it has this same job each and every time it's invoked. Right now like any compiler, it assumes that any dependencies of the compilation have been installed and are ready to go.

`ngcc`, on the other hand, is intimately concerned with the build system. It has to deal with multiple package formats and dependencies between libraries. It has a vastly different lifecycle, and must understand when it has previously compiled packages and avoid recompiling them to save time. It also must understand the requirements of the overall bundling operation to know which package formats must be compiled and which are safe to ignore in the interest of speed.

Combining the two would require `ngtsc`, via its API, to become aware of much more detail about the current build process in order to run `ngcc` correctly. This would effectively transform `ngtsc` from a compiler into a build tool. I think that's a job that's better left to the real build toolchain, whether that's the CLI or Blaze or `npm` scripts.

I also disagree that running ngcc-on-demand via a hook in the `ts.CompilerHost` is the right approach for all users. This effectively serializes its operation, and I strongly believe there are major performance benefits to running `ngcc` in a multi-process mode, in parallel across the whole of `node_modules`. That's something we've yet to explore, but I think this would make `ngcc` as a post-install hook reasonably performant.

packages/ngtools/webpack/src/compiler_host.ts
alan-agius4: Maybe indeed we should change this. Can we have a separate chat and maybe address this in a separate PR. Pinged you already on slack.

petebacondarwin:
Actually the strongest argument above from @alxhub is that we don't want ngtsc to need to know about package formats and bundles, etc. That is definitely a concern of the build tool not the TS compiler.

alxhub:
Also, there is a non-trivial build problem: `ngtsc` and `ngcc` have different deps themselves (`ngcc` is more liberal and uses `canonical-path`, `yargs`, `dependency-graph`, etc). These are part of our `package.json` but aren't available in google3. This works because `ngcc` code is excluded from the sync.

If we were to start building `ngcc` into `ngtsc` more directly, we'd quickly get into a thorny situation where the internal `ngtsc` would have to not even _not_ even import `ngcc`, but the external `ngtsc` would. This would require local modifications to be applied in our sync script, etc, and would generally be a messy situation.

clydin:
This actually leads back to one of my original questions.  If `ngcc` is intended to support bundlers and not `ngtsc`, should it be modifying `.d.ts` files when bundlers don't use or need them?

petebacondarwin:
@clydin - the purpose of ngcc is two fold:

* enable libraries to run under the ivy rendering engine
* support ngtsc being able to build code (for ivy) that depends upon code that was originally built for the view-engine renderer.

In the former case, Ivy calls static methods on classes to be able to get the definitions of components, etc. So ngcc needs to add these methods.
In the latter case, ngtsc needs information about the components in the library, which it gets from the typings. So ngcc needs to update these too.

clydin:
Since that is the case, wouldn't that make `ngcc` more intimately linked to `ngtsc` than any bundler?  `ngtsc` will not currently work without it and a bundler is irrelevant if there is nothing to bundle.

alexeagle:
Need to update latest-versions.ts to beta.10 also?

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -660,13 +630,33 @@ export class AngularCompilerPlugin {
...
+        } catch {
+        }
+
+        if (ngcc) {
+          ngccProcessor = new NgccProcessor(
+            ngcc,
+            this._mainFields,
+            compilerWithFileSystems.inputFileSystem,
+          );
+        }
```
petebacondarwin: Why have this block outside the try-catch? It could be much simpler:

```ts
let ngccProcessor: NgccProcessor | undefined;
if (this._compilerOptions.enableIvy) {
  try {
    // this is done for the sole reason that @ngtools/webpack
    // support versions of Angular that don't have NGCC API
    const ngcc: typeof import('@angular/compiler-cli/ngcc') | undefined = require('@angular/compiler-cli/ngcc');
    ngccProcessor = new NgccProcessor(
      ngcc,
      this._mainFields,
      compilerWithFileSystems.inputFileSystem,
    );
  }
} catch {
}
```

alan-agius4: As like this if NgccProcessor throws an error during it's initialisation it will be swallowed.

petebacondarwin: OK OK

alan-agius4:
@alexeagle It should not be needed as ~8.0.0-beta.9 does pick up beta.10

PR #13981, opened 2019-03-25
Update build system to use core-js 3

clydin:
Also, required ES polyfills are now managed directly by the build system.  If additional polyfills are required, packages can be manually added as needed.  This reduces the number of direct dependencies for a new project.


alan-agius4:
At the moment core-js is not a directly dependency of build-angular. But rather it’s defined in the workspace package.json

https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/workspace/files/package.json.template

alan-agius4:
@clydin, can you kindly reference https://github.com/angular/angular-cli/issues/13954https://github.com/angular/angular-cli/issues/13954?

packages/schematics/angular/library/files/src/test.ts.template
```diff
@@ -1,7 +1,5 @@
 // This file is required by karma.conf.js and loads recursively all the .spec and framework files
```
alan-agius4: Is this intentional?

clydin: An experiment.

Cleaned up both.

packages/schematics/angular/workspace/files/package.json.template
```diff
@@ -19,7 +19,6 @@
     "@angular/platform-browser": "<%= latestVersions.Angular %>",
     "@angular/platform-browser-dynamic": "<%= latestVersions.Angular %>",
     "@angular/router": "<%= latestVersions.Angular %>",
-    "core-js": "^2.5.4",
```
alexeagle: how does the user get core-js on their machine if it's no longer in this package.json?

clydin: It's a dependency of `build-angular` now; it was a hidden dependency before.  Moving it allows full control of the version preventing any potential mismatches; as well as allowing alternate polyfilling strategies moving forward without changing/breaking user's project dependencies.

packages/angular_devkit/build_angular/src/angular-cli-files/models/es5-polyfills.js
```diff
@@ -0,0 +1,84 @@
...
+import 'core-js/es/math';
+import 'core-js/es/date';
+import 'core-js/es/regexp';
+import 'core-js/es/map';
+import 'core-js/es/weak-map';
+import 'core-js/es/set';
+import 'core-js/es/promise';
+
+// ES2015 object capabilities
+import 'core-js/es/object/create';
```
nseni: Why is it necessary to import each function individually?
Why not import everything at once by `import 'core-js/es/object';`?

clydin: Angular only requires the ES2015 polyfills.  In `core-js` 3.0, `core-js/es/object` includes all current ES polyfills (up to ES2019).  This was an attempt to reduce the bundle size increase.  Using the single imports resulted in an increase of ~15kb to the polyfills file.  Further work is currently being done to attempt to close the gap even further.

PR #13982, opened 2019-03-25
Disable inline sourcemaps for builds & increase target to ES2018 (Node.js 10.3+)

clydin:
This also includes separate commits for increasing the target version to ES2018 which is supported by Node.js 10.3.0+ ([reference](https://node.green/#ES2018)) and some minor cleanup to the tsconfig file.

### DIFF
--- a/tsconfig-test.json
+++ b/tsconfig-test.json
@@ -1,6 +1,12 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "inlineSourceMap": true,
+    "sourceRoot": ".",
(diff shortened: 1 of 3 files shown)


tsconfig-test.json
```diff
@@ -1,6 +1,12 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "inlineSourceMap": true,
+    "sourceRoot": ".",
+    // Inline sources are necessary for our tests to show the proper sources, since we are using
+    // Istanbul (not Constantinople) as well, and applying both source maps to get the original
```
hansl: <3

PR #13983, opened 2019-03-25
Telemetry various fix

hansl:
The prompt was never shown to the user because of a combination of issues;

1. the "scripts" key in package.json is not kept during building.
1. when a user lack a global configuration file, we return as if the user answered false (because it's `null`, not `undefined`).
1. prompting on non-tty terminals also meant that CI was broken.

This fixes those issues.


cyrilletuzi:
Hi @hansl, is this supposed to be OK in beta.10? Because my Circle-CI is failing because of the new analytics prompt.

PR #13986, opened 2019-03-26
Analytics for Builds, and other fixes

hansl:
Multiple changes:

1. Builder stats for every build. This was tested using logging and Architect CLI.  
    The only missing piece is for the Angular CLI to pass an instance of `Analytics` interface.
1. Validation script for the user analytics, to make sure the document is kept up to date.
1. Rework some metrics into dimensions, which lets us make better reports.
1. adding duration tracking for all commands.

### DIFF
--- a/packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
+++ b/packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
@@ -42,14 +42,23 @@ export function countOccurrences(source: string, match: string, wordBreak = fals
   // We condition here so branch prediction happens out of the loop, not in it.
   if (wordBreak) {
     const re = /\w/;
-    for (let pos = source.lastIndexOf(match); pos >= 0; pos = source.lastIndexOf(match, pos - 1)) {
-      if (!(re.test(source[pos - 1]) || re.test(source[pos + match.length]))) {
+    for (let pos = source.lastIndexOf(match); pos >= 0; pos = source.lastIndexOf(match, pos)) {
+      if (!(re.test(source[pos - 1] || '') || re.test(source[pos + match.length] || ''))) {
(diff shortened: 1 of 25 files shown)


packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
```diff
@@ -0,0 +1,256 @@
...
+
+/**
+ * Faster than using a RegExp, so we use this to count occurences in source code.
+ * @param source The source to look into.
+ * @param match The match string to look for.
+ * @param wordBreak Whether to check for word break before and after a match was found.
+ * @return The number of matches found.
+ * @private
+ */
+export function countOccurrences(source: string, match: string, wordBreak = false): number {
```
alexeagle: would be nice to unit test this - in case we find some weird logging edge case

packages/angular_devkit/build_angular/src/utils/analytics.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+export const build = {
```
alexeagle: check if its used

packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
hansl: Done (and fixed a bug)

packages/angular_devkit/build_angular/src/utils/analytics.ts
hansl: Whole file removed.

PR #13992, opened 2019-03-26
fix(@ngtools/webpack): don't invalidate cache after first run

alan-agius4:
- fix(@ngtools/webpack): don't invalidate cache after first run
At the moment, since there are no old files in the compilation it will cause all source files to be invalidate after the first run. This shouldn't be done as it will slow down the 2nd recompilation.

- fix(@ngtools/webpack): replace resources should not return class node when modified
This is also the root cause of spec large fail occasionally as we keep checking the child class nodes

### DIFF
--- a/packages/ngtools/webpack/src/transformers/replace_resources.ts
+++ b/packages/ngtools/webpack/src/transformers/replace_resources.ts
@@ -18,9 +18,21 @@ export function replaceResources(
 
     const visitNode: ts.Visitor = (node: ts.Decorator) => {
       if (ts.isClassDeclaration(node)) {
-        node.decorators = ts.visitNodes(
+        const decorators = ts.visitNodes(
           node.decorators,
           (node: ts.Decorator) => visitDecorator(node, typeChecker, directTemplateLoading),
+        );
+
+        // todo: we need to investigate and confirm that using
+        //  `updateClassDeclaration` has no regressions
+        return ts.updateClassDeclaration(
+          node,
+          decorators,
+          node.modifiers,
+          node.name,
+          node.typeParameters,
+          node.heritageClauses,
+          node.members,
         );
       }
 
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -361,12 +361,7 @@ export class AngularCompilerPlugin {
 
     // Use an identity function as all our paths are absolute already.
     this._moduleResolutionCache = ts.createModuleResolutionCache(this._basePath, x => x);
-
-    const tsProgram = this._getTsProgram();
-    const oldFiles = new Set(tsProgram ?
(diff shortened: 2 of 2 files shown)


packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -18,10 +18,20 @@ export function replaceResources(
 
     const visitNode: ts.Visitor = (node: ts.Decorator) => {
       if (ts.isClassDeclaration(node)) {
-        node.decorators = ts.visitNodes(
+        const decorators = ts.visitNodes(
```
clydin: This was originally intentional as symbol information would be lost on the TS side and cause runtime errors due to non-Angular decorators becoming disconnected from type information.

alan-agius4: You mean the missing return, or not using `updateClassDeclaration`?

alan-agius4:
Actually it was by pure coincidence, After I did the cache invalidation thing, the spec failed constantly with the error `Cannot read property 'kind' of undefined` in that test that is usefully flaky.

The flakiness seems to be related to the same error, as I got that error locally when I tried a couple of runs of that spec.

packages/ngtools/webpack/src/transformers/replace_resources.ts
clydin: the `updateClassDeclaration`.   Mutating the existing node was the intentional part.

alan-agius4: Okay, Ill revert that part. As long as there is a `return node` to stop the recursion it should work.

clydin: This also may have been corrected in newer version of TS.  Can you add a comment that this should be reviewed and a test case added?

alan-agius4: At the moment, without this change, there is a spec which fails `rebuilds on TS file changes`, which is causing `Cannot read property 'kind' of undefined`

clydin:
A fully virtual filesystem would be ideal for these tests since it would isolate the underlying OS's behavior from the test scenario.  Unfortunately, chokidar makes that pretty much impossible.

packages/ngtools/webpack/src/transformers/replace_resources.ts
alan-agius4: As discussed on slack, we still need the `updateClassDeclaration` and without this it causes some `Karma Builder watch mode` specs to fail. 

`updateClassDeclaration` is also used by the FW when replacing resources (https://github.com/angular/angular/blob/dc10355d61a98092b5f8cabc0993d737ed2e721f/packages/compiler-cli/src/transformers/inline_resources.ts#L119) so it should be safe to use. I left a comment none-the-less.

PR #13993, opened 2019-03-26
feat(@angular-devkit/build-angular): differential loading check

manfredsteyer:


packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
```diff
@@ -0,0 +1,138 @@
...
+  const target = getTargetVersion(tsConfigFileName);
+
+  return es5SupportNeeded && target !== ScriptTarget.ES5;
+}
+
+export function getTargetVersion(tsConfigFileName: string): ScriptTarget {
+  return _getTargetVersion(tsConfigFileName, new Set<string>());
+}
+
+function _getTargetVersion(tsConfigFileName: string, visited: Set<string>): ScriptTarget {
```
alan-agius4: We shouldn't be re-creating the ts parsing parsing logic of typescript. You can simply use `readTsconfig` method.

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
```diff
@@ -0,0 +1,138 @@
+/**
```
alan-agius4: Can you please add return types to methods?

alan-agius4: Maybe this file is better names as `utils.ts`, what do you think?

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    return _getTargetVersion(nextTsConfigFileName, visited);
+  } else {
+    // Fallback to ES5
+    return ScriptTarget.ES5;
+  }
+
+}
+
+export function isEs5BrowserSupportNeeded(
```
alan-agius4: Some docs would help here. Example what is a `query`?

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
```diff
@@ -0,0 +1,138 @@
...
+
+function convertSupportTableToBrowserVersionTuples(supportTable: object): Map<string, string> {
+  // The supportTable bases upon data from the caniuse api.
+  // It uses the y property to indicate the version beginning
+  // with the feature in question is supported.
+  const browsers = Object.getOwnPropertyNames(supportTable) as string[];
+  const supportMap = new Map<string, string>();
+
+  browsers
+     // @ts-ignore
```
alan-agius4: Why is this needed?

packages/angular_devkit/build_angular/test/differential-loading/is-differential-loading-needed_spec_large.ts
```diff
@@ -0,0 +1,68 @@
+/**
```
alan-agius4: This seems to be a `spec` file which can be combined with the others.

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
manfredsteyer: We don't have typings for the data structures provided by caniuse. Hence, here it's just an object (supportTable: object). However, `supportTable[b].y` leads to an implicit any issue.

packages/angular_devkit/build_angular/test/differential-loading/is-differential-loading-needed_spec_large.ts
manfredsteyer: Which others do you mean?

tests/angular_devkit/build_angular/test-differential-loading/tsconfig.app.json
```diff
@@ -0,0 +1,8 @@
+{
```
alan-agius4: I am not quite sure why we need to add all these tsconfig.json files.

The large spec, should cover an existing application that we have a large spec.

packages/angular_devkit/build_angular/test/differential-loading/is-differential-loading-needed_spec_large.ts
alan-agius4: This one https://github.com/angular/angular-cli/pull/13993/files#diff-7379beedd1dbd6738b8da9124058d8b2

bin/gen-browser-infos.js
```diff
@@ -0,0 +1,13 @@
+const caniuse = require('caniuse-api');
```
alan-agius4: What is the usage of this file?

alan-agius4: how is this hooked in the build system?

bin/gen-browser-infos.js
```diff
@@ -0,0 +1,13 @@
...
+const fs = require('fs');
+
+const path =
+  '../packages/angular_devkit/build_angular/src/differential-loading/es6-module-support.json';
+const contents = JSON.stringify(caniuse.getSupport('es6-module'));
+
+fs.writeFileSync(path, contents, { encoding: 'UTF-8' });
+
```
alan-agius4: Nit: remove extra breaklines

packages/angular_devkit/build_angular/package.json
```diff
@@ -13,6 +13,7 @@
     "@ngtools/webpack": "0.0.0",
     "ajv": "6.10.0",
     "autoprefixer": "9.4.9",
+    "browserslist": "^4.4.2",
```
alan-agius4: Direct dependencies should be pinned.

packages/angular_devkit/build_angular/package.json
```diff
@@ -26,24 +27,25 @@
...
+    "source-map-support": "0.5.11",
+    "speed-measure-webpack-plugin": "1.3.1",
     "stats-webpack-plugin": "0.7.0",
     "style-loader": "0.23.1",
     "stylus": "0.54.5",
     "stylus-loader": "3.0.2",
-    "tree-kill": "1.2.1",
     "terser-webpack-plugin": "1.2.3",
+    "tree-kill": "1.2.1",
+    "typescript": "^3.3.4000",
```
alan-agius4: TypeScript should be a peerDependency and we need to also support older versions of TS
```
    "typescript": ">=2.7 < 3.4",
```

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
alan-agius4: What be good to do some internal interfaces, for type checking.

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
```diff
@@ -0,0 +1,138 @@
...
+  } else if (version1Tupel[1] < version2Tupel[1]) {
+    return -1;
+  } else if (version1Tupel[1] > version2Tupel[1]) {
+    return 1;
+  } else {
+    return 0;
+  }
+}
+
+function toVersionTupel(version: string): [number, number] {
```
alan-agius4: Kindly confused what this tuples means. What is version `[0]` and what is `[1]`?

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
manfredsteyer: +1

bin/gen-browser-infos.js
mgechev: 👍

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
mgechev: Coincidental cohesion is not ideal. I like the current name. Maybe `_` instead of `-` so we can be consistent?

packages/angular_devkit/build_angular/test/differential-loading/is-differential-loading-needed_spec_large.ts
manfredsteyer: Ah, I see. It's about merging the `spec` with the `large_spec` file, right? 

Currently, `spec` does everything we can do in memory. If I need to access the hard disk, I'm using `large_spec`. Isn't this the intended way?

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
```diff
@@ -0,0 +1,138 @@
...
+export function isEs5BrowserSupportNeeded(
+  tsConfigFileName: string | null, queries: string[] | undefined = undefined) {
+
+  const options: browserslist.Options = {};
+  if (tsConfigFileName) {
+    options.path = path.dirname(tsConfigFileName);
+  }
+
+  const browsersList: string[] = browserslist(queries, options);
+
```
mgechev: Extra space

tests/angular_devkit/build_angular/test-differential-loading/tsconfig.app.json
mgechev: Below we have an example of a tsconfig cycle, this is probably not covered by our existing app?

Maybe `readTsconfig` covers this?

packages/angular_devkit/build_angular/test/differential-loading/is-differential-loading-needed_spec_large.ts
alan-agius4: Typically test large are used to test the full architect builder and not a unit.

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
alan-agius4: That is true, though I am not sure about the file name per see, since the exported methods here, are really not related to 'differential loading enabled' per see.

Such as `getScriptTarget`, also, I won’t name a file with an `is` prefix in that case. Unless this file has a single responsibility. Though still naming a module with an `is` prefix doesn’t seems right to me. imho.

manfredsteyer: For me, both is ok. The only function we need to export is `isDifferentialLoadingNeeded`. The others are only exported for testability.

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
manfredsteyer: +1

packages/angular_devkit/build_angular/test/differential-loading/is-differential-loading-needed_spec_large.ts
manfredsteyer: Ok, I see. In this case, merging makes sense.

tests/angular_devkit/build_angular/test-differential-loading/tsconfig.app.json
manfredsteyer: Yes, this was the idea. As I was not aware of `readTsconfig`, I wrote this by my self. And hence, I wanted to test the logic with a tsconfig pointing to ES5, a tsconfig pointing to ES2015, a tsconfig w/o a target pointing to another tsconfig with a extends cycle. 

Now, as I'm using `readTsconfig` this does not make sense anymore as we can assume that this function which is part of typescript does its job.

bin/gen-browser-infos.js
manfredsteyer: We can do this. I was not sure if we should add this 1.x MB lib even though we just need this simple piece of data.

bin/gen-browser-infos.js
manfredsteyer: +1

packages/angular_devkit/build_angular/package.json
manfredsteyer: +1

packages/angular_devkit/build_angular/package.json
manfredsteyer: +1

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
manfredsteyer: It's about the major and the minor part of the version. And about comparing two versions with a major and a minor. I can replace the tuple with a class. 

Or do we already have such a function somewhere? 

In this case, reusing it would be the way to go, I guess ...

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
manfredsteyer: What do you mean?

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
manfredsteyer: I'll rename the variable to `browserslistQuery`.

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
manfredsteyer: it's now called util

packages/angular_devkit/build_angular/src/differential-loading/is-differential-loading-needed.ts
clydin: The `semver` package can be used and is already a dependency.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+function toVersionTupel(version: string): [number, number] {
+  const temp = version.split('.');
+  if (temp.length === 1) {
+    temp[1] = '0';
+  }
+
+  return [parseInt(temp[0]), parseInt(temp[1])];
+}
+
+function parseBrowserVersionTuple(raw: string): [string, string] {
```
clydin: Doesn't appear to be used?

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as browserslist from 'browserslist';
+import * as caniuse from 'caniuse-api';
+import * as path from 'path';
+import { ScriptTarget } from 'typescript';
+import { readTsconfig } from '../angular-cli-files/utilities/read-tsconfig';
```
clydin: This will most likely be removed in an upcoming refactoring/cleanup.  You can use `ts.getParsedCommandLineOfConfigFile` directly.

tests/angular_devkit/build_angular/test-differential-loading/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html><html lang="en"><head>
```
clydin: These HTML files don't appear to be used.

alan-agius4:
Can you also rebase?

packages/angular_devkit/build_angular/package.json
```diff
@@ -26,55 +29,58 @@
     "loader-utils": "1.2.3",
     "mini-css-extract-plugin": "0.5.0",
     "minimatch": "3.0.4",
-    "parse5": "4.0.0",
     "opn": "5.4.0",
+    "parse5": "4.0.0",
```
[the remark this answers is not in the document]
manfredsteyer: `parse5` is used in `index-html-webpack-plugin.ts`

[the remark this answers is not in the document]
clydin: yes.  however, it is being moved around; along with several other entries (some with version number changes as well)

packages/angular_devkit/build_angular/package.json
```diff
@@ -14,6 +14,8 @@
     "@ngtools/webpack": "0.0.0",
     "ajv": "6.10.0",
     "autoprefixer": "9.5.0",
+    "browserslist": "^4.5.2",
```
alan-agius4: pin direct dependencies please

alexeagle: alternatively, consider vendoring them into our repo so we don't add more transitive dependencies

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  tsConfigFileName: string | null, browserslistQuery: string[] | undefined = undefined): boolean {
+
+  const neededFeature = 'es6-module';
+
+  const options: browserslist.Options = {};
+  if (tsConfigFileName) {
+    options.path = path.dirname(tsConfigFileName);
+  }
+
+  const browsersList: string[] = browserslist(browserslistQuery, options);
```
alan-agius4: We don't need to use `browserslist` ourselves `caniuse.isSupported`, supports `browserslistQuery`.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export function isDifferentialLoadingNeeded(
+  tsConfigFileName: string, browserslistQuery: string[] | undefined = undefined): boolean {
+  const es5SupportNeeded = isEs5BrowserSupportNeeded(tsConfigFileName, browserslistQuery);
+  const target = getTargetVersion(tsConfigFileName);
+
+  return es5SupportNeeded && target !== ts.ScriptTarget.ES5;
+}
+
+export function getTargetVersion(tsConfigFileName: string): ts.ScriptTarget {
```
alexeagle: is it possible to reference a `ts.Program` already at this point, so we don't have to worry about things like whether we found the right `tsconfig` file?

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    // fall back to ES5
+    return ts.ScriptTarget.ES5;
+  }
+
+  return target;
+}
+
+export function isEs5BrowserSupportNeeded(
+  tsConfigFileName: string | null, browserslistQuery: string[] | undefined = undefined): boolean {
```
alan-agius4: Wouldn't these browser list be coming from `browserslist` file which is being read below?

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
mgechev: We're using this to get the `target` output. We determine the maximum version (i.e. the `script[module]` one) from this property.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
manfredsteyer: Absolutely. I'm currently using the option to pass in a `browserslistQuery` instead of reading the `browserslist` file for testing.

packages/angular_devkit/build_angular/package.json
manfredsteyer: I've pinned it now

packages/angular_devkit/build_angular/package.json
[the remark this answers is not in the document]
manfredsteyer: It's now updated and it has a minimal amount of changes.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
alexeagle: sure, but how does the user of this API know how to locate the correct tsconfig file if there are several in the project?

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
alan-agius4: Ummm, probably we need to remove these before merging this PR.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
manfredsteyer: The user of this API will be the browser builder which finds the tsconfig.json to use in the `angular.json`. It can also get overruled by a command line parameter.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
manfredsteyer: I've changed it now. The unit tests read the browserlistQuerys now from the hard disk. Hence, there is no need for this `browserslistQuery` parameter anymore.

packages/angular_devkit/build_angular/test/differential-loading/differential_loading_spec.ts
```diff
@@ -0,0 +1,75 @@
+/**
```
alan-agius4: I think there is a bit of misunderstanding here, specs are usefully next to the feature file. These specs here are usefully used to test the architect and named `large`.

These are more of an integration test rather than a unit. Unit tests should reside next to the feature.

Also, we should not interact with the File System directly, we should use the Abstraction layer as this would reduce a couple of things.

1) Makes mocking easier.
2) No need to create a physical file for each test. 
3) Have the mocks as part of the spec file hence makes the test more readable.

```
const workspaceRoot = path.join(
  devkitRoot,
  'tests/angular_devkit/build_angular/test-differential-loading/');
const host = new TestProjectHost(normalize(workspaceRoot));

describe('differential loading', () => {
  it('detects the need for ES5 if IE 9-11 has to be supported', () => {
    const path = normalize('browerlist');
    host.writeMultipleFiles({
        path: 'last 1 chrome version'
    });
    const es5SupportNeeded = isEs5BrowserSupportNeeded(workspaceRoot, path);
    expect(es5SupportNeeded).toBe(true);
  });
});
```

The `host` itself has a lot for method, that can help you with transforming a file, overriding it's content, delete it etc..., the important one is that it will never override the file on the hard drive.

alan-agius4: Umm actually `TestProjectHost` seems to be deprecated, @clydin what's the new preferred way to approach this?

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
alan-agius4: If the user of this is browser builder, we already know the script target and if es5Support is needed or not. As there is where we modify the mainFields to include es2015 or not. So why do we need to get this again?

When you say that it can be overrules by a command line parameter, can you please elaborate? Wouldn't such flag be in the browser builder?

I think in general there is a bit of a lack of understanding from me on how everything will be glued together.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    // fall back to ES5
+    return ts.ScriptTarget.ES5;
+  }
+
+  return target;
+}
+
+export function isEs5BrowserSupportNeeded(
+  browserslistPath: string, browserslistConfigFile: string | undefined = undefined): boolean {
```
alan-agius4: How would we know the `browserslistConfigFile` from the CLI? this can be anything which browserlist supports. While at the moment in the CLI we generate an emotionless `browserlist`, the users are free to change this as long as browserlist can read it? 

Hence, why not simply provide the workspacePath and let `browserslist` resolve it''s configuration file? Similar to how it's currently done by autoprefixer in the style config.

`browserslist` configuration can be in various files such as;

1. package.json
2. .browserslistrc
3. browserslist

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
mgechev: @alan-agius4 what you're saying sounds right and aligns with the design. On `browserlist` depends if we should generate ES5 bundles or not.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
```diff
@@ -0,0 +1,64 @@
+/**
```
alan-agius4: To determine whether deferential loading is needed it should be simple;

```ts
// In browser builder we already know if the it supports `supportES2015` as we use this to populate the correct `mainFields`
export function isDifferentialLoadingNeeded(projectRoot: string, supportES2015: boolean): boolean {
  const browsersList: string[] = browserslist(undefined, {path: projectRoot});
  return supportES2015 && !caniuse.isSupported('es6-module', browsersList.join(', '));
}
```

or
```ts
// In browser builder we already know the target
export function isDifferentialLoadingNeeded(projectRoot: string, target: ts.ScriptTarget): boolean {
  const supportES2015 = target !== projectTs.ScriptTarget.ES3 && target !== projectTs.ScriptTarget.ES5;
  const browsersList: string[] = browserslist(undefined, {path: projectRoot});
  return supportES2015 && !caniuse.isSupported('es6-module', browsersList.join(', '));
}
```

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
manfredsteyer: This should be the default, I agree. The current design doc also specifies that we should be able to define one using a parameter (command line parameter when calling ng build or a parameter within angular.json). I think, this would come in-handy b/c so one could build versions for different targets.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
manfredsteyer: I see your point. Thank you!

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
manfredsteyer: Ok, I see what you mean. THX!

packages/angular_devkit/build_angular/test/differential-loading/differential_loading_spec.ts
alan-agius4: Still pending.

packages/angular_devkit/build_angular/src/differential-loading/utils.ts
alan-agius4: At the moment, having different multiple browserlist configs is not supported by autoprefixer. it only supports either resolving the browserlist file as per it's internal resolution strategy or an array of browsers to support.

My point is that if we are going to provide a way for users to change the browserlist with a configuration parameter.  We need to change the logic in a couple of builders as this would also determine which prefixes would need to be used during server builder.

IMHO, if such approach will need to happen it shouldn't be the responsibility of the differential loader utils and it should be somewhere more centralised:

In general I'd like to avoid having to parse the same file multiple times.
-> parse browser list config as required
--> pass this down the differential loader
--> pass this down to the webpack styles configuration.

That said, I am not seeing what would be the users case of building the application several times using different browser versions. (I will check the design doc)

manfredsteyer: Ok, so we leave this away for now, right?

manfredsteyer: I removed this param now

packages/angular_devkit/build_angular/test/differential-loading/differential_loading_spec.ts
manfredsteyer: I've tried this. While this totally makes sense, it seems that in this very case, using the virtual filesystem does not work. The reason is that browserslist has its own machanism for finding and loading the `browswerslist` config. This one isn't using the virtual filesystem.

alan-agius4: This should work, as the TestProjectHost writes files to disks.

See below;
Screenshot 2019-04-02 at 17 45 13



If you need a hand let me know.

manfredsteyer: It seems to be a Windows-related issue: `host.root()` returns sth like `/c/my-dir` instead of `c:/my-dir`. The forward slashes are no issue here.

alan-agius4: Yeah, that is expected if you pass a devkit path directly to the method yourself that is using FS.

Devkit paths are always posix normalised, which means on windows `c:\\` will be `/c/` which means that any interaction with FS will fail on windows. Devkit paths/methods cannot be used with direct interaction with fs. 

The projectRoot, is normalised is passed from `wco`. See https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/browser/index.ts#L127

The safest and cleaner approach would be changing the `projectRoot` to Path and using `getSystemPath`

```
export function isDifferentialLoadingNeeded(
  projectRoot: Path,
  target: ts.ScriptTarget): boolean {

   const supportES2015 = target !== ts.ScriptTarget.ES3 && target !== ts.ScriptTarget.ES5;
  const browsersList: string[] = browserslist(
    undefined, {
      path: getSystemPath(projectRoot),
    });

   return supportES2015 && !caniuse.isSupported('es6-module', browsersList.join(', '));
}
```

alan-agius4:
@manfredsteyer is this absolute in favor of the other PR or not?

packages/angular_devkit/build_angular/test/differential-loading/differential_loading_spec.ts
manfredsteyer: Thanks!

manfredsteyer:
> in favor of the



> @manfredsteyer is this absolute in favor of the other PR or not?

Yes, it is @alan-agius4

PR #13994, opened 2019-03-26
feat(@angular/cli): support stable architect API

clydin:
Conversion of the CLI to use the new stable architect API.
Additional refactoring and cleanup is possible.  The goal of this PR is to get the capabilities introduced.

### DIFF
--- a/packages/angular/cli/models/architect-command.ts
+++ b/packages/angular/cli/models/architect-command.ts
@@ -5,12 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {
-  Architect,
-  BuilderConfiguration,
-  BuilderContext,
-  TargetSpecifier,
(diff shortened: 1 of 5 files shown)


packages/angular/cli/models/architect-command.ts
```diff
@@ -5,12 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {
-  Architect,
-  BuilderConfiguration,
-  BuilderContext,
-  TargetSpecifier,
-} from '@angular-devkit/architect';
+import { index2 } from '@angular-devkit/architect';
```
alexeagle: nit: maybe `import { index2 as architect }` so that later when we remove the old architect API we don't have so many delta lines where this is referenced below?

clydin: The imports will be direct once we switch over completely so we’ll end up with the same delta.

PR #14004, opened 2019-03-27
feat(@schematics/angular): drop es6 from modern polyfills

mgechev:
1. Remove imports of es6 polyfills introduced by the CLI.
2. Refactor the migrations for version 8 by moving the codelyzer and
polyfill transforms into different files.

The PR drops all `core-js/es6` polyfills that we've introduced with the
CLI, except the commented ones. We do not remove commented imports,
since they are not part of the internal es6 polyfills.

The migration automatically drops the
associated comments with the removed imports since they are part of the node - under its
`jsDoc` property.


packages/schematics/angular/migrations/update-8/drop-es6-polyfills.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const sourceFile = ts.createSourceFile(path, source.toString(), ts.ScriptTarget.Latest, true);
+    const imports = sourceFile.statements
+        .filter(s => s.kind === ts.SyntaxKind.ImportDeclaration) as ts.ImportDeclaration[];
+
+    if (imports.length === 0) { return; }
+
+    for (const i of imports) {
+      const module = ts.isStringLiteral(i.moduleSpecifier) && i.moduleSpecifier.text;
+      if (module && (module.startsWith('core-js/es6/') || module.startsWith('zone.js/dist/zone'))) {
```
clydin: we should only remove the polyfills that the CLI placed in the file.  The developer may have added addition ones.

PR #14006, opened 2019-03-28
refactor(@angular-devkit/build-ng-packagr): cleanup experimental architect API

clydin:
Removes alternate `index2` and any remaining elements of the experimental architect API.


packages/angular_devkit/build_ng_packagr/src/build/index.ts
```diff
@@ -5,80 +5,41 @@
...
+import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect/src/index2';
+import { resolve } from 'path';
+import { Observable, from } from 'rxjs';
+import { catchError, mapTo, switchMap } from 'rxjs/operators';
+import { Schema as NgPackagrBuilderOptions } from './schema';
+
+async function initialize(
+  options: NgPackagrBuilderOptions,
+  root: string,
+): Promise<import ('ng-packagr').NgPackagr> {
```
alan-agius4: Since `ng-packagr` is a peerDep why not import it directly?

clydin: Limits the volume of modules/packages that need to be loaded until the builder is actually executed.

PR #14005, opened 2019-03-27
feat(@schematics/angular): update to codelyzer 5

mgechev:
This PR also updates the hello world sample app and migrates its `tslint.json` file.

PR #14007, opened 2019-03-28
feat(@schematics/angular): update codelyzer to version 5

alan-agius4:
More about rules renaming in this commit: https://github.com/angular/angular-cli/commit/4b89fe21f4cde2a9a605e0cb571a3e223cf19a6f#diff-3b943998563d6e245cac63da45ce53f2R23


alan-agius4:
Seems like @mgechev aready is addressing this https://github.com/angular/angular-cli/pull/14005

PR #14015, opened 2019-03-28
updating webpack config for differential loading

manfredsteyer:


packages/angular_devkit/build_angular/src/differential-loading/update-webpack-config.ts
```diff
@@ -0,0 +1,132 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+type WebpackConfig = any; // tslint:disable-line
```
alexeagle: why? either give a local type that matches the shape, or import a correct type, or explain in the //tslint comment why you can't

packages/angular_devkit/build_angular/src/differential-loading/update-webpack-config.ts
```diff
@@ -0,0 +1,132 @@
...
+type WebpackPlugin = any; // tslint:disable-line
+
+export function updateWebpackConfig(config: WebpackConfig): WebpackConfig {
```
alexeagle: "update" is not descriptive. could you find a name that says what is the update?

createSubConfigs or something? (what does webpack call it when configs are nested in this way?)

packages/angular_devkit/build_angular/src/differential-loading/update-webpack-config.ts
```diff
@@ -0,0 +1,132 @@
...
+  const entry = config.entry;
+  const polyfills: string[] = entry.polyfills;
+
+  if (!polyfills) {
+    return { ...entry };
+  }
+
+  const tweakPolyfills = (file: string) => {
+
+    // TODO: Do we need an own polyfill file for modern browsers? If yes, where is it created?
```
alexeagle: when do we plan to fix the TODO? can we figure it out in this PR?

packages/angular_devkit/build_angular/src/differential-loading/update-webpack-config.ts
```diff
@@ -0,0 +1,132 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
```
alexeagle: needs some file-level commenting explaining what's going on, in case the reader hasn't seen the design doc. linking to the design might also be nice

packages/angular_devkit/build_angular/src/differential-loading/update-webpack-config.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+function buildModernStatsPlugin(statsPlugin: WebpackPlugin): object {
+  const modernStatsPlugin = new statsPlugin.constructor();
+  modernStatsPlugin.output = 'stats.modern.json';
+
+  return modernStatsPlugin;
+}
+
+function buildLegacyAcp(acp: WebpackPlugin): object {
```
alexeagle: `Acp` is that intentional spelling?

packages/angular_devkit/build_angular/src/differential-loading/update-webpack-config.ts
```diff
@@ -0,0 +1,132 @@
...
+  const options = acp._options;
+  const modernOptions = { ...options };
+
+  // TODO: Where shall we create this tsconfig with target=es5?
```
alexeagle: will we address the TODO in this PR?

packages/angular_devkit/build_angular/src/differential-loading/update-webpack-config.ts
manfredsteyer: Question for reviewers:  Do we need an own polyfill file for modern browsers? If yes, where is it created?

packages/angular_devkit/build_angular/src/differential-loading/update-webpack-config.ts
manfredsteyer: Question for reviewers: Where shall we create this tsconfig with target=es5? Schematics? On the fly if it does not exist?

packages/angular_devkit/build_angular/src/differential-loading/create-webpack-multi-config.ts
```diff
@@ -0,0 +1,185 @@
+/**
```
alan-agius4: inline comments near the implementations would definitely help the reviewers and future engineers that go through the code, without the need of going back and forth to the design doc.

packages/angular_devkit/build_angular/src/differential-loading/create-webpack-multi-config.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+  };
+
+  const modernPolyfills = polyfills.map(tweakPolyfills);
+  const modernEntry = { ...entry as webpack.Entry, polyfills: modernPolyfills };
+
+  return modernEntry;
+}
+
+function buildLegacyPlugins(config: webpack.Configuration): webpack.Plugin[] | undefined {
```
alan-agius4: Why not generate a webpack config which is already property configured for an es2015 or es5 build?

packages/angular_devkit/build_angular/src/differential-loading/create-webpack-multi-config.ts
```diff
@@ -0,0 +1,185 @@
...
+export function createWebpackMultiConfig(config: webpack.Configuration): webpack.Configuration[] {
+
+  const legacyConfig: webpack.Configuration = {
+    ...config,
+    output: buildLegacyOutput(config),
+    plugins: buildLegacyPlugins(config),
+    resolve: buildLegacyResolve(config),
+  };
+
+  const modernConfig: webpack.Configuration = {
```
alan-agius4: With this approach you are missing some configuration for es2015 https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts#L236 and https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts#L216

Hence, I strongly recommend that we avoid doing such transformations and let the webpack config builder return the appropriate configs

packages/angular_devkit/build_angular/src/differential-loading/create-webpack-multi-config.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  let tsConfigPath = modernOptions.tsConfigPath;
+  if (tsConfigPath.endsWith('.json')) {
+    tsConfigPath = tsConfigPath.substr(0, tsConfigPath.length - 5);
+  }
+  tsConfigPath += '.legacy.json';
+
+  modernOptions.tsConfigPath = tsConfigPath;
+
+  const legacyAcp = new AngularCompilerPlugin(modernOptions);
```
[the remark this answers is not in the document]
manfredsteyer: Yes. It works if we adopt some methods a bit.

packages/angular_devkit/build_angular/src/differential-loading/create-webpack-multi-config.ts
```diff
@@ -0,0 +1,185 @@
...
+}
+
+function buildLegacyAngularCompilerPlugin(acp: AngularCompilerPlugin): AngularCompilerPlugin {
+  const options = acp.options;
+  const modernOptions = { ...options };
+
+  // TODO: Where shall we create this tsconfig with target=es5?
+  //  Schematics? On the fly if it does not exist?
+
+  let tsConfigPath = modernOptions.tsConfigPath;
```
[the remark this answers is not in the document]
manfredsteyer: TBH, I don't like it either. However, I see no alternative for this. Or is it possible to pass an object literal representing the contents of a tsconfig.json to the AngularCompilerPlugin? I guess, in this case the AngularCompilerPlugin also needs to write it to hard disk.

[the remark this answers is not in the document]
alan-agius4: You can pass the compiler options to the AngularCompilerPlugin. https://github.com/angular/angular-cli/blob/42ef58295e2c1ce43be98d431c8b4055233823be/packages/ngtools/webpack/src/interfaces.ts#L27

packages/angular_devkit/build_angular/src/differential-loading/create-webpack-multi-config.ts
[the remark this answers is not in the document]
alan-agius4: Just to confirm are going to go for webpack multicompiler? At the moment also @angular-devkit/webpack, doesn’t have any implementation for webpack multi-compiler. 

Also can you elaborate what you mean by adopt some methods a bit?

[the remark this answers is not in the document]
alan-agius4: At the moment @angular-devkit/build-webpack doesn’t have implementations for multi-compiler.

packages/angular_devkit/build_angular/src/differential-loading/create-webpack-multi-config.ts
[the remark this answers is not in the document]
manfredsteyer: Ah, that means I could read the tsconfig.json into an object, modify it's target in memory and assign it to the ``compilerOptions`` property, right? In this case we don't need a seperate tsconfig.json for this target on disk. Do I see this right?

PR #12012, opened 2018-08-27
fix(@angular-devkit/build-angular): exclude `.map` files from budget …

alan-agius4:
exclude `.map` files from budget calculations

Closes #11999

Issue #11999, opened 2018-08-24
source maps should not be included in initial budgets


StephenFluin:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Command (mark with an `x`)
```
- [x] build
```

### Versions
Angular CLI: 6.2.0-rc.0
Node: 9.2.0
OS: linux x64
Angular: 6.1.4
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.8.0-rc.0
@angular-devkit/build-angular     0.8.0-rc.0
@angular-devkit/build-optimizer   0.8.0-rc.0
@angular-devkit/build-webpack     0.8.0-rc.0
@angular-devkit/core              0.8.0-rc.0
@angular-devkit/schematics        0.8.0-rc.0
@angular/cdk                      6.4.6
@angular/cli                      6.2.0-rc.0
@angular/material                 6.4.6
@ngtools/webpack                  6.2.0-rc.0
@schematics/angular               0.8.0-rc.0
@schematics/update                0.8.0-rc.0
rxjs                              6.2.2
typescript                        2.7.2
webpack                           4.17.1

### Repro steps
ng new my-app
yarn add @angular/cli@next @angular-devkit/build-angular@next
add budget to angular.json:
```
               {
                  "type": "initial",
                  "maximumWarning": "200kb",
                  "maximumError": "500kb"
                }

```
increase bundle size to warning level 400kb by adding material and polyfills
ng build --prod --source-map=true

### The log given by the failure
```
$ ng build --prod --source-map=true
[...]

PR #14017, opened 2019-03-29
fix(@angular-devkit/build-angular): exclude map files from 'bundle' budgets

NickGeek:
#11999 raised the issue of bundle budgets for the `initial` type including source maps. That issue was fixed in #12012. However, as was noted by @Tonio31, source maps should also be excluded from `bundle`-type budgets.


NickGeek:
Ah, sorry about that! I'll go and write a proper test :-)

NickGeek:
Thanks for getting that sorted and merged in! Sorry I couldn't get that fixed up earlier, uni got a bit hectic.

PR #14022, opened 2019-03-29
fix(@angular-devkit/architect): remove experimental API

clydin:
This leaves several `index2` alternate entry points for the stable API to support the full transition.  These will be removed in a final cleanup PR.


packages/angular_devkit/build_angular/test/browser/aot_spec_large.ts
```diff
@@ -7,11 +7,9 @@
  */
 
 import { Architect } from '@angular-devkit/architect/src/index2';
```
filipesilva: Shouldn't this also change to `@angular-devkit/architect`?

clydin: Yes.  There are a large number of those in the `build-angular` package (60+ files).  I have a separate PR pending that will cleanup the whole package.

filipesilva: SGTM

PR #14023, opened 2019-03-29
build: integrate prettier code formatting

clydin:
Integrates prettier code formatting for changed files.  Also adds a CI check to ensure changed files are formatted in the event that the pre-commit git hook is bypassed.  Conflicting TSLint style rules are also removed.

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,0 +1,3 @@
+{
+  "*.{js,ts}": ["prettier --write", "git add"]
+}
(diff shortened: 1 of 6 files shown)


.lintstagedrc
```diff
@@ -0,0 +1,3 @@
+{
+  "*.{js,ts}": ["prettier --write", "git add"]
```
alan-agius4: Wouldn’t it make sense to include json so that schemas are formatted as well?

clydin: Wanted to start with the code first.  There's also tests that rely on specific JSON file contents so there's some effort there to get the exclusions correct.

PR #14024, opened 2019-03-29
fix(@schematics/angular): buildRelativePath handles files in root

ukrukarg:
Before, if one of the arguments was a file in root (eg "/module")
code would fail with: "" must be an absolute path.


ukrukarg:
More context: 
- run into this writing my own schematics for a project that doesn't follow standard angular directory structure (meaning, there isn't always app/ directory and we do have ts files in the same directory as angular.json file.

mgechev:
Ups, sorry, I forced pushed the wrong revision. @ukrukarg would you open the PR again?

PR #14029, opened 2019-03-30
refactor(@angular/cli): use standard node resolution methods where possible

clydin:


packages/angular/cli/lib/init.ts
```diff
@@ -77,14 +76,7 @@ if (process.env['NG_CLI_PROFILING']) {
...
 try {
-  const projectLocalCli = resolve(
-    '@angular/cli',
-    {
-      checkGlobal: false,
-      basedir: process.cwd(),
-      preserveSymlinks: true,
-    },
-  );
+  const projectLocalCli = require.resolve('@angular/cli', { paths: [process.cwd()] });
```
alan-agius4: Umm... I am not sure this retains the same behaviour. 

When providing `paths`, it will retain the global paths. Hence, if it is not resolved locally it will still fallback to the global paths.

clydin: It won't use the actual node modules global paths.  A global module is only able to require another global module and a local module is only able to require another local module.  When the paths option is used Node creates a fake parent module that is located at the provided path for each entry in the option; bypassing the actual parent module.

The global paths mentioned in the documentation (https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders) are actually historical artifacts and only relevant for ancient versions of node.  These were used when the local node modules concept was not yet formalized and don't exist on modern node systems.  The actual global node modules location is not located at any of those locations and is actually a construct of the package manager itself (yarn is different than npm, for instance).

alan-agius4: Thanks for the detailed explanation

PR #14038, opened 2019-04-01
feat(@schematics/angular): drop zones from newly generated projects

mgechev:
For consistency between existing and newly generated projects (see
https://github.com/angular/angular-cli/pull/14004).

### DIFF
--- a/packages/schematics/angular/application/files/src/polyfills.ts.template
+++ b/packages/schematics/angular/application/files/src/polyfills.ts.template
@@ -28,36 +28,6 @@
  */
 // import 'web-animations-js';  // Run `npm install --save web-animations-js`.
 
-/**
- * By default, zone.js will patch all possible macroTask and DomEvents
- * user can disable parts of macroTask/DomEvents patch by setting following flags
- * because those flags need to be set before `zone.js` being loaded, and webpack
- * will put import in the top of bundle, so user need to create a separate file
- * in this directory (for example: zone-flags.ts), and put the following flags
- * into that file, and then add the following code before importing zone.js.
- * import './zone-flags.ts';
- *
- * The flags allowed in zone-flags.ts are listed here.
- *
- * The following flags will work for all browsers.
- *
- * (window as any).__Zone_disable_requestAnimationFrame = true; // disable patch requestAnimationFrame
- * (window as any).__Zone_disable_on_property = true; // disable patch onProperty such as onclick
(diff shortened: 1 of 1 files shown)


mgechev:
Another inconsistency:

```
// import 'core-js/es6/weak-map';

/** IE10 and IE11 requires the following for NgClass support on SVG elements */
// import 'classlist.js';  // Run "npm install --save classlist.js".

/** Evergreen browsers require these. **/
// Used for reflect-metadata in JIT. If you use AOT (and only Angular decorators), you can remove.
import 'core-js/es7/reflect';
```

We have the above snippet in existing projects but we've dropped it from `polyfills.ts.template`. @clydin, we should probably drop these if we include them in the `es2015` polyfills bundle.

alan-agius4:
How would users be able to enable the following?

```
 * (window as any).__Zone_disable_requestAnimationFrame = true; // disable patch requestAnimationFrame	
 * (window as any).__Zone_disable_on_property = true; // disable patch onProperty such as onclick	
 * (window as any).__zone_symbol__UNPATCHED_EVENTS = ['scroll', 'mousemove']; // disable patch specified eventNames
```

I am still not fully understanding how the zone polyfill will be added for non es2015 builds.~

Will `zone.js` be added by the build system at the top of polyfills files?

mgechev:
>```
> * (window as any).__Zone_disable_requestAnimationFrame = true; // disable patch >requestAnimationFrame	
> * (window as any).__Zone_disable_on_property = true; // disable patch onProperty such as onclick	
> * (window as any).__zone_symbol__UNPATCHED_EVENTS = ['scroll', 'mousemove']; // disable patch >specified eventNames
>```

This part probably we should keep?

> I am still not fully understanding how the zone polyfill will be added for non es2015 builds.~

cc @clydin

clydin:
I'm not sure that the current set of zone.js configuration comments are overly useful in their current incarnation.  To use any of those settings a separate file needs to be created and then imported before the zone.js import.  If creating a separate file, it would be just as easy to either add the zone.js options to an inline script block in the `index.html` or as an asset referenced via a script element in the HTML.  These are also advanced options that, I think, would be better served with a documentation guide detailing when/if they should even be used.
Also of consideration is if zone.js itself should offer a side-effect free based import that allows a configuration object to be passed to an _initialization_ function.  This would allow customization of zone.js behavior without the current magic global properties and separate file/import complications.

PR #14048, opened 2019-04-02
feat(@ngtools/webpack): redirect ngcc errors and warnings to webpack

alan-agius4:
With this change NGCC will no longer emit logs to console. 

Also, errors and warnings will be redirected to webpack


alan-agius4:
Currently blocked, as I am using the snapshot

PR #14054, opened 2019-04-02
feat(@angular/cli): verify Angular version is supported

filipesilva:


packages/angular/cli/upgrade/version.ts
```diff
@@ -61,6 +65,27 @@ export class Version {
       process.exit(2);
     }
 
+    try {
+      const angularCliPkgPath = require.resolve('@angular/cli/package.json', resolveOptions);
```
clydin: Since we are already in the cli package, should there be an easier (and more efficient) method to retrieve the version?  (e.g., static version field that is set at build time)

packages/angular/cli/upgrade/version.ts
```diff
@@ -61,6 +65,27 @@ export class Version {
...
+      `)));
+      process.exit(2);
+    }
+
+    const angularCliVersion = new Version(angularCliPkgJson['version']);
+    const upperBound = angularCliVersion.major;
+    const lowerBound = upperBound - 1;
+    // e.g. CLI 8.0 supports '>=7.0.0 <9.0.0 || ^7.0.0-beta.0 || ^8.0.0-beta.0'.
+    const supportedAngularSemver = `>=${lowerBound}.0.0 <${upperBound + 1}.0.0` +
+      ` || ^${lowerBound}.0.0-beta.0 || ^${upperBound}.0.0-beta.0`;
```
clydin: The `semver` package supports the `includePrerelease ` option now:  https://github.com/npm/node-semver#functions

packages/angular/cli/upgrade/version.ts
```diff
@@ -70,9 +95,9 @@ export class Version {
       return;
     }
 
-    if (!angularVersion.isGreaterThanOrEqualTo(new SemVer('5.0.0'))) {
+    if (!angularVersion.satisfies(supportedAngularSemver)) {
       console.error(terminal.bold(terminal.red(tags.stripIndents`
-          This version of CLI is only compatible with Angular version 5.0.0 or higher.
+          This version of CLI is only compatible with Angular versions ${supportedAngularSemver}.
```
clydin: I think this should also tell the user what versions they currently have.  Thoughts?

packages/angular/cli/upgrade/version.ts
filipesilva: Unsure. I mean, it's nice and all, but worth adding? I don't think we access this version in many places. Might also complicate things with some global/local setups. 

If we decide to do this, we'd need to add tooling to automatically update it. The bazel npm package toolchain does that I think, so maybe waiting for that would a good approach.

packages/angular/cli/upgrade/version.ts
filipesilva: SGTM, changed.

packages/angular/cli/upgrade/version.ts
filipesilva: I think that's a good idea, added.

packages/angular/cli/upgrade/version.ts
clydin: sounds like a plan.

packages/angular/cli/upgrade/version.ts
```diff
@@ -70,9 +94,10 @@ export class Version {
       return;
     }
 
-    if (!angularVersion.isGreaterThanOrEqualTo(new SemVer('5.0.0'))) {
+    if (!angularVersion.satisfies(supportedAngularSemver)) {
```
clydin: looks like the pre-release option is not used here?

filipesilva: `angularVersion` is an instance of `Version`. I added it directly in the `satisfies` method:
```
return this._semver !== null && satisfies(this._semver, other, { includePrerelease: true });
```

packages/angular/cli/upgrade/version.ts
```diff
@@ -61,6 +68,26 @@ export class Version {
...
+        This likely means your local installation is broken. Please reinstall your packages.
+      `)));
+      process.exit(2);
+    }
+
+    const angularCliVersion = new Version(angularCliPkgJson['version']);
+    const upperBound = angularCliVersion.major;
+    const lowerBound = upperBound - 1;
+    // e.g. CLI 8.0 supports '>=7.0.0 <9.0.0', including pre-releases (betas, rcs, snapshots).
+    const supportedAngularSemver = `>=${lowerBound}.0.0 <${upperBound + 1}.0.0`;
```
filipesilva: @alexeagle can you verify if the N and N-1 versions are the right supported range?

alexeagle: It seems right to me. let me check with Stephen to validate...

tests/legacy-cli/e2e/tests/misc/supported-angular.ts
```diff
@@ -0,0 +1,54 @@
+import { SemVer } from 'semver';
```
alexeagle: I am surprised that so much code is added in this PR. Didn't we do something similar in earlier versions?

clydin: It was hard coded to check for angular 5.0+.
This file is on my list to refactor after the RC.  About half of it is either unused or unneeded.

filipesilva:
Blocked on https://github.com/angular/angular/pull/29705

alexeagle:
@filipesilva is it unblocked?

filipesilva:
@alexeagle need a release from Angular, or to update our test setup to support Angular snapshots for all CI jobs.

tests/legacy-cli/e2e/assets/1.0-project/src/app/app.module.ts
```diff
@@ -1,7 +1,6 @@
 import { BrowserModule } from '@angular/platform-browser';
 import { NgModule } from '@angular/core';
 import { FormsModule } from '@angular/forms';
-import { HttpModule } from '@angular/http';
```
filipesilva: `@angular/http` wasn't actually used in this app and is deprecated, causing problems after updating in the test.

PR #14061, opened 2019-04-02
feat(@angular-devkit/build-angular): differential loading

manfredsteyer:
This PR adds differential loading to the browser builder. First, it checks if differential loading is needed. This is the case if the compilation target is ES2015 while the browserslist points to ES5 browsers.

For providing differential loading, it calls the methods for creating the webpack config for each compilation target (e. g. ES5 and ES2015). The needed differences between those configurations are defined using parameters also added by this PR. Then it calls webpack for each of them, merges the results and writes the index.html.

This feature is currently hidden behind a flag. To activate it, set the ENABLE_DIFFERENTIAL_LOADING flag in packages/angular_devkit/build_angular/src/browser/index.ts.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -8,6 +8,7 @@
 import { tags } from '@angular-devkit/core';
 import * as CopyWebpackPlugin from 'copy-webpack-plugin';
 import * as path from 'path';
+import * as ts from 'typescript';
 import {
   Configuration,
   ContextReplacementPlugin,
@@ -21,8 +22,8 @@ import { CleanCssWebpackPlugin } from '../../plugins/cleancss-webpack-plugin';
 import { ScriptsWebpackPlugin } from '../../plugins/scripts-webpack-plugin';
 import { findAllNodeModules, findUp } from '../../utilities/find-up';
 import { requireProjectModule } from '../../utilities/require-project-module';
-import { BuildOptions, WebpackConfigOptions } from '../build-options';
-import { getOutputHashFormat, normalizeExtraEntryPoints } from './utils';
+import { WebpackConfigOptions } from '../build-options';
+import { getEsVersionForFileName, getOutputHashFormat, normalizeExtraEntryPoints } from './utils';
 
 const ProgressPlugin = require('webpack/lib/ProgressPlugin');
 const CircularDependencyPlugin = require('circular-dependency-plugin');
@@ -55,16 +56,36 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
   const extraPlugins: any[] = [];
   const entryPoints: { [key: string]: string[] } = {};
 
+  const targetInFileName = getEsVersionForFileName(
+    buildOptions.scriptTargetOverride,
+    buildOptions.esVersionInFileName,
+  );
+
   if (buildOptions.main) {
     entryPoints['main'] = [path.resolve(root, buildOptions.main)];
   }
 
+  const es5Polyfills = path.join(__dirname, '..', 'es5-polyfills.js');
   if (buildOptions.es5BrowserSupport) {
-    entryPoints['polyfills.es5'] = [path.join(__dirname, '..', 'es5-polyfills.js')];
+    entryPoints['polyfills.es5'] = [es5Polyfills];
+    if (!buildOptions.aot) {
+      entryPoints['polyfills.es5'].push(path.join(__dirname, '..', 'es5-jit-polyfills.js'));
+    }
+  }
+
+  if (buildOptions.es5BrowserSupport === undefined
+    && buildOptions.scriptTargetOverride === ts.ScriptTarget.ES5) {
+    entryPoints['polyfills'] = [es5Polyfills];
+    if (!buildOptions.aot) {
+      entryPoints['polyfills'].push(path.join(__dirname, '..', 'es5-jit-polyfills.js'));
+    }
   }
 
   if (buildOptions.polyfills) {
-    entryPoints['polyfills'] = [path.resolve(root, buildOptions.polyfills)];
+    entryPoints['polyfills'] = [
+      ...(entryPoints['polyfills'] || []),
+      path.resolve(root, buildOptions.polyfills),
+    ];
   }
 
   if (!buildOptions.aot) {
@@ -72,18 +93,11 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
       ...(entryPoints['polyfills'] || []),
       path.join(__dirname, '..', 'jit-polyfills.js'),
     ];
-
-    if (buildOptions.es5BrowserSupport) {
-      entryPoints['polyfills.es5'] = [
-        ...entryPoints['polyfills.es5'],
-        path.join(__dirname, '..', 'es5-jit-polyfills.js'),
-      ];
-    }
   }
 
   if (buildOptions.profile || process.env['NG_BUILD_PROFILING']) {
     extraPlugins.push(new debug.ProfilingPlugin({
-      outputPath: path.resolve(root, 'chrome-profiler-events.json'),
+      outputPath: path.resolve(root, `chrome-profiler-events${targetInFileName}.json`),
     }));
   }
 
@@ -104,7 +118,6 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
           }
 
           existingEntry.paths.push(resolvedPath);
-
         } else {
           prev.push({
             bundleName,
@@ -177,7 +190,7 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
   }
 
   if (buildOptions.statsJson) {
-    extraPlugins.push(new StatsPlugin('stats.json', 'verbose'));
+    extraPlugins.push(new StatsPlugin(`stats${targetInFileName}.json`, 'verbose'));
   }
 
   let sourceMapUseRule;
@@ -305,7 +318,7 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
       futureEmitAssets: true,
       path: path.resolve(root, buildOptions.outputPath as string),
       publicPath: buildOptions.deployUrl,
-      filename: `[name]${hashFormat.chunk}.js`,
+      filename: `[name]${targetInFileName}${hashFormat.chunk}.js`,
       // cast required until typings include `futureEmitAssets` property
     } as Output,
     watch: buildOptions.watch,
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
@@ -12,6 +12,7 @@ import * as path from 'path';
 import { basename, normalize } from '@angular-devkit/core';
 import { ExtraEntryPoint, ExtraEntryPointClass } from '../../../browser/schema';
 import { SourceMapDevToolPlugin } from 'webpack';
+import * as ts from 'typescript';
 
 export const ngAppResolve = (resolvePath: string): string => {
   return path.resolve(process.cwd(), resolvePath);
@@ -88,4 +89,15 @@ export function getSourceMapDevTool(
     include,
     append: hiddenSourceMap ? false : undefined,
   });
+}
+
+/**
+ * Returns an ES version file suffix to differentiate between various builds.
+ */
+export function getEsVersionForFileName(
+  scriptTargetOverride: ts.ScriptTarget | undefined,
+  esVersionInFileName = false,
+): string {
+  return scriptTargetOverride && esVersionInFileName ?
+    '-' + ts.ScriptTarget[scriptTargetOverride].toLowerCase() : '';
 }
--- a/packages/angular_devkit/build_angular/src/browser/index.ts
+++ b/packages/angular_devkit/build_angular/src/browser/index.ts
@@ -5,16 +5,14 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+
 import {
   BuilderContext,
   BuilderInfo,
   BuilderOutput,
   createBuilder,
 } from '@angular-devkit/architect';
-import {
-  WebpackLoggingCallback,
-  runWebpack,
-} from '@angular-devkit/build-webpack';
+import { WebpackLoggingCallback, runWebpack } from '@angular-devkit/build-webpack';
 import {
   Path,
   analytics,
@@ -25,27 +23,27 @@ import {
   logging,
   normalize,
   resolve,
-  schema,
-  virtualFs,
+  schema, virtualFs,
 } from '@angular-devkit/core';
 import { NodeJsSyncHost } from '@angular-devkit/core/node';
 import * as fs from 'fs';
-import { Observable, from, of } from 'rxjs';
+import { Observable, combineLatest, from, of } from 'rxjs';
 import { concatMap, map, switchMap } from 'rxjs/operators';
-import * as ts from 'typescript'; // tslint:disable-line:no-implicit-dependencies
+import * as ts from 'typescript';
+import * as webpack from 'webpack';
 import { NgBuildAnalyticsPlugin } from '../../plugins/webpack/analytics';
 import { WebpackConfigOptions } from '../angular-cli-files/models/build-options';
 import {
   getAotConfig,
   getBrowserConfig,
   getCommonConfig,
+  getEsVersionForFileName,
   getNonAotConfig,
   getStatsConfig,
   getStylesConfig,
   getWorkerConfig,
 } from '../angular-cli-files/models/webpack-configs';
 import { readTsconfig } from '../angular-cli-files/utilities/read-tsconfig';
-import { requireProjectModule } from '../angular-cli-files/utilities/require-project-module';
 import { augmentAppWithServiceWorker } from '../angular-cli-files/utilities/service-worker';
 import {
   statsErrorsToString,
@@ -60,7 +58,7 @@ import {
 } from '../utils';
 import { Schema as BrowserBuilderSchema } from './schema';
 
-import webpack = require('webpack');
+
 const SpeedMeasurePlugin = require('speed-measure-webpack-plugin');
 const webpackMerge = require('webpack-merge');
 
@@ -68,7 +66,6 @@ const webpackMerge = require('webpack-merge');
 export type BrowserBuilderOutput = json.JsonObject & BuilderOutput & {
   outputPath: string;
 };
-
 
 export function createBrowserLoggingCallback(
   verbose: boolean,
@@ -101,7 +98,7 @@ export function buildWebpackConfig(
     analytics?: analytics.Analytics,
     builderInfo?: BuilderInfo,
   } = {},
-): webpack.Configuration {
+): webpack.Configuration[] {
   // Ensure Build Optimizer is only used with AOT.
   if (options.buildOptimizer && !options.aot) {
     throw new Error(`The 'buildOptimizer' option cannot be used without 'aot'.`);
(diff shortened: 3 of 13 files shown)


packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  // Temporarily commented out
+  // afterEach(async () => host.restore().toPromise());
+
+  it('works', async () => {
+
+    const tsConfigJsonPath = join(host.root(), normalize('tsconfig.json'));
+
+    // HACK: Somehow there is currently an issue under Windows
+    const tsConfigJsonPathCorrected = tsConfigJsonPath.replace('/C/', 'C:/');
```
manfredsteyer: @alan-agius4 This is what i meant

alan-agius4: Let’s take it in slack, and I’ll explain what you need to do.

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Architect } from '@angular-devkit/architect/src/index2';
+import { join, normalize } from '@angular-devkit/core';
+import { updateJsonFile } from '../../../../../tests/legacy-cli/e2e/utils/project';
```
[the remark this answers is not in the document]
manfredsteyer: I see. What is the best way to update that file? Just use the `read` and `write` methods the host provides or are there some helpers?

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -260,7 +302,27 @@ export function buildWebpackBrowser(
         normalize(workspace.getProject(projectName).root),
       );
 
-      return runWebpack(config, context, { logging: loggingFn }).pipe(
+      return combineLatest(
+        configs.map( (config, idx) => runWebpack(
```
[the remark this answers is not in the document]
manfredsteyer: You mean, b/c this is executed twice?

Yes, this is also listed as a drawback in the design doc. 

My proposal is to provide optimizations for this in further minor versions (e. g. 8.1). We can try to parallelize this and we can avoid copying assets twice.

packages/angular_devkit/build_webpack/src/utils.ts
```diff
@@ -13,16 +13,24 @@ export interface EmittedFiles {
...
   initial: boolean;
   extension: string;
+
+  // if several builds are created at once for differential serving
+  // this points to the index of the configuration that was used
+  configIndex?: number;
 }
 
-export function getEmittedFiles(compilation: webpack.compilation.Compilation): EmittedFiles[] {
+export function getEmittedFiles(
```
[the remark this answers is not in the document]
manfredsteyer: We need to know later, which configuration -- the es5 or the es2015 config -- was used to produce this file. Ofc, we could also determine this somewhere else, but what I want to prevent is having two data structures like ``EmittedFiles`` and ``EmittedFilesWithConfigIndex``.

What do you think?

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -102,47 +103,70 @@ export function buildWebpackConfig(
...
-    const smp = new SpeedMeasurePlugin({
-      outputFormat: 'json',
-      outputTarget: getSystemPath(join(root, 'speed-measure-plugin.json')),
-    });
+    const webpackConfig = webpackMerge(webpackConfigs);
+
+    if (options.profile || process.env['NG_BUILD_PROFILING']) {
+        const smp = new SpeedMeasurePlugin({
+          outputFormat: 'json',
+          outputTarget: getSystemPath(join(root, 'speed-measure-plugin.json')),
```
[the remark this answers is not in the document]
manfredsteyer: good catch. thx

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -6,9 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
```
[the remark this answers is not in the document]
manfredsteyer: I've run ``yarn run fix``. Is there any other script for this?

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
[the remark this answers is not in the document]
alan-agius4: I suggest to take a look at https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/test/browser/allow-js_spec_large.ts#L25 for an example.

packages/angular_devkit/build_angular/src/browser/index2.ts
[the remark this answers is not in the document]
alan-agius4: Um at the moment formatting is not done via a task. `fix` task only fixes tslint issues.

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
```diff
@@ -0,0 +1,55 @@
...
+    const tsConfigJsonPathCorrected = tsConfigJsonPath.replace('/C/', 'C:/');
+
+    await updateJsonFile(tsConfigJsonPathCorrected, json => {
+      json.compilerOptions.target = 'ES2015';
+    });
+
+    await browserBuild(architect, host, target);
+
+    // Default files should be in outputPath.
+    expect(await host.scopedSync().exists(normalize('dist/runtime-es5.js'))).toBe(true);
```
alan-agius4: Maybe you can simplify this test by using the jasmine helps.

```ts
    expect(host.scopedSync().list(normalize('dist')))
      .toEqual(jasmine.arrayContaining([
        // list of files
      ]);

```

packages/angular_devkit/build_webpack/src/utils.ts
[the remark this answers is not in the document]
alan-agius4: I don't think this should the responsibility of the webpack runner, it should be handled by the browser builder.

You are using `combineLatest` with always respects the order, hence you know what emits what.

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
```
alan-agius4: I'd like to add a couple of more test here.

1) Watch mode works as expected
3) Can recover from errors
4) Would also like to verify the contents on the es5 and es2015, Example having a `function` and `class` respectively,
5) Verify that web workers also get differential loaded.

packages/angular_devkit/build_angular/src/browser/index2.ts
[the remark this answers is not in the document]
alan-agius4: I think the main issue here, would be that the most heavy parts such as typescript compilation, running terser and buildOptimizer would run twice on the same process.

Ps `combineLatest` is already parallelizing this.

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
manfredsteyer: Thx! This helped!

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -102,47 +103,70 @@ export function buildWebpackConfig(
...
-    logger: logger.createChild('webpackConfigOptions'),
-    projectRoot: getSystemPath(projectRoot),
-    buildOptions: options,
-    tsConfig,
-    tsConfigPath,
-    supportES2015,
-  };
+  const target = tsConfig.options.target;
+
+  const supportES2015 = target !== projectTs.ScriptTarget.ES3
```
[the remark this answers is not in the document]
manfredsteyer: This duplication is also used in the server builder. Perhaps we should do a seperate refactoring PR. What do you think?

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -102,47 +103,70 @@ export function buildWebpackConfig(
...
+    wco.buildOptions.progress = defaultProgress(wco.buildOptions.progress);
+
+    const webpackConfigs: {}[] = [
+        getCommonConfig(wco),
+        getBrowserConfig(wco),
+        getStylesConfig(wco),
+        getStatsConfig(wco),
+      ];
+
+    if (wco.buildOptions.main || wco.buildOptions.polyfills) {
```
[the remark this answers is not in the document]
manfredsteyer: It's in `wco.buildOptions.scriptTargetOverride`

packages/angular_devkit/build_angular/src/browser/index2.ts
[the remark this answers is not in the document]
manfredsteyer: I've autoformatted it with web storm now.

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
manfredsteyer: Note to myself: Also check for serviceworker

manfredsteyer: I might need help with the web worker test.

packages/angular_devkit/build_angular/src/browser/index2.ts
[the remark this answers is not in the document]
clydin: Yes.  It’s on my list for refactoring prior to the final release.

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
manfredsteyer: I'll look into this tomorrow.

packages/angular_devkit/build_angular/src/browser/index2.ts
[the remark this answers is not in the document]
alan-agius4: You are only modifying one instance of the AOT plugin.

The one I highlighted above is untouched hence it will always be build with the same target.

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
alan-agius4: Just ping with with any question.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -49,6 +50,9 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
   const extraPlugins: any[] = [];
   const entryPoints: { [key: string]: string[] } = {};
 
+  const targetInFileName = getEsVersionForFileName(wco);
```
alan-agius4: Optional Nit: no need to pass in the entire webpack config options.

packages/angular_devkit/build_angular/src/browser/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  const supportES2015 = target !== ts.ScriptTarget.ES3 && target !== ts.ScriptTarget.ES5;
+
+  return supportES2015 && isEs5SupportNeeded(projectRoot);
+}
+
+let _isEs5SupportNeeded: boolean | undefined = undefined;
+
+export function isEs5SupportNeeded(projectRoot: string): boolean {
+
+  if (typeof _isEs5SupportNeeded !== 'undefined') {
```
alan-agius4: This doesn't seem correct. The function will not always return the same value for the same file, because a value might be already set.

What is the reason behind this? Have you a performance hit?

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -206,6 +249,72 @@ export type BrowserConfigTransformFn = (
...
+      const path = join(root, normalize(fileName), normalize(fileName));
+      const buffer = await host.read(path).toPromise();
+
+      return virtualFs.fileBufferToString(buffer);
+    },
+    unfilteredSortedFiles: [
+      ...files1,
+      ...files2,
+    ],
+    noModuleFiles: new Set<string>(['polyfills.es5-es5.js']),
```
alan-agius4: As per slack convo `polyfills.es5.js `

packages/angular_devkit/build_angular/src/browser/index2.ts
```diff
@@ -260,27 +369,46 @@ export function buildWebpackBrowser(
...
+              // if we are doing differential loading, we need to write the index.html
+              // console.debug('writing index.html', buildEvent);
+              return from(writeIndexHtml(buildEvents, options, root, host)).pipe(
+                map(_ => buildEvents),
+              );
+            }
+
+            return of(buildEvents);
+          }),
+          map(buildEvents => mergeBuildResult(buildEvents)),
```
alan-agius4: The webpack builder only returns `success` and `emittedFiles`.

emittedFiles is not used beyond here as also, if flattened it's harder to use.

Hence most likely we can do;
```ts
map(buildEvents => {success: buildEvents.every(x => x.success)}
```

packages/angular_devkit/build_angular/src/browser/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  target: ts.ScriptTarget): boolean {
+
+  const supportES2015 = target !== ts.ScriptTarget.ES3 && target !== ts.ScriptTarget.ES5;
+
+  return supportES2015 && isEs5SupportNeeded(projectRoot);
+}
+
+let _isEs5SupportNeeded: boolean | undefined = undefined;
+
+export function isEs5SupportNeeded(projectRoot: string): boolean {
```
alan-agius4: some js docs would help :)

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/generate-index-html_spec.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { generateIndexHtml } from './generate-index-html';
+
+describe('index-html-webpack-plugin', () => {
```
alan-agius4: the spec here seems to be testing something else and not the webpack plugin

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
```diff
@@ -89,176 +88,26 @@ export class IndexHtmlWebpackPlugin {
...
-
-      indexSource.insert(
-        styleInsertionPoint,
-        parse5.serialize(styleElements, { treeAdapter }),
-      );
+      // If this plugin calls generateIndexHtml it always uses type: 'none' to align with
+      // its original behavior.
+      const compiledFiles: CompiledFileInfo[] = unfilteredSortedFiles.map(f => ({
+        file: f,
+        type: 'none' as CompiledFileType,
```
alan-agius4: Nit: this should should be auto inferred.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/generate-index-html.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  const noModuleFilesArray = params.unfilteredSortedFiles
+    .filter(f => f.type === 'nomodule')
+    .map(f => f.file);
+
+  noModuleFilesArray.push(...params.noModuleFiles);
+
+  const noModuleFiles = new Set<string>(noModuleFilesArray);
+
+  for (const file of fileNames) {
```
alan-agius4: You can now use the extension.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
```diff
@@ -67,6 +65,7 @@ export class IndexHtmlWebpackPlugin {
       (compilation as compilation.Compilation & { fileDependencies: Set<string> })
```
alan-agius4: I think the sorting logic needs to move downwards, in the index generator, as otherwise, we'll have two place that have sorting logic.

As mentioned, the sorting logic, should not happened at `@angular/build-webpack`, as that can be used with any webpack config and if not aware of any logic with `build-angular`, and not aware that `polyfills` entrypoint should take priority over `main`, this logic should really happen when generating the index, which is the sole place that will be used by all builders to generate an index.

packages/angular_devkit/build_angular/src/browser/index2.ts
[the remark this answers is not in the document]
manfredsteyer: Cool. Thx.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
manfredsteyer: Good point. That means we move it to the `generate-index-html.ts`, right?

alan-agius4: yes

packages/angular_devkit/build_angular/src/browser/index2.ts
alan-agius4: Seeing the usages we might only need if it was success or not hence. Most likely we can get away with using
```
     map(buildEvents => {success: buildEvents.every(x => x.success)}),
```

alexeagle:
@mgechev we are chatting about diff-loading in the triage meetings. Some ideas to make this more grop effort:

1) Could we reduce the scope for this PR to just produce the two different bundles into the dist/ directory?
2) @clydin can help with writing the index next, and figure out the right script ordering
3) Let's land it with a private flag that disables the feature to start with so we can get this in without breaking the next release

packages/angular_devkit/build_webpack/src/utils.ts
```diff
@@ -19,14 +19,17 @@ export function getEmittedFiles(compilation: webpack.compilation.Compilation): E
...
 
-  for (const chunk of Object.values(compilation.chunks)) {
-    const entry: Partial<EmittedFiles> = {
-      name: chunk.name,
-      initial: chunk.isOnlyInitial(),
-    };
-
-    for (const file of chunk.files) {
-      files.push({ ...entry, file, extension: getExtension(file) } as EmittedFiles);
+  for (const key of compilation.entrypoints.keys()) {
```
alan-agius4: Can you explain a bit why this is needed? This won't return all the files like `compilation.chunks`.

Also, some tests to verify that it generates what you need would be great.

packages/angular_devkit/build_webpack/src/utils.ts
```diff
@@ -19,14 +19,17 @@ export function getEmittedFiles(compilation: webpack.compilation.Compilation): E
...
+    const entrypoint = compilation.entrypoints.get(key);
+    if (entrypoint && entrypoint.getFiles) {
+      for (const file of entrypoint.getFiles()) {
+        files.push({
+          name: key as string,
+          file,
+          extension: getExtension(file),
+          initial: false,
```
alan-agius4: This doesn't seem right.

packages/angular_devkit/build_angular/src/browser/index2.ts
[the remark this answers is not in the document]
manfredsteyer: Ah, I see. This seems to be related to the new worker features. I think, according to @alexeagle [feedback below](https://github.com/angular/angular-cli/pull/14061#issuecomment-479990716), we should do this in a seperate PR to avoid this one becomes too large.

Do I see this right?

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
manfredsteyer: I'd like to keep this for the time being, as the pattern used here is also used in an existing test. Is this ok?

packages/angular_devkit/build_angular/src/browser/index2.ts
manfredsteyer: I think it's best, to do this with another PR to respect @alexeagle's feedback here: https://github.com/angular/angular-cli/pull/14061#issuecomment-479990716)

manfredsteyer:
> Some more comments above, also since we have a redundant css build. IE, we are building css twice. I am not seeing any logic to remove the CSS files from one of the builds.

Let's do this in another PR regarding optimizations. Is this ok?

packages/angular_devkit/build_webpack/src/utils.ts
manfredsteyer: It's about determining the right order of the script tags: We just have the right order of the entrypoints. However, one entry point can lead to several files, like `runtime.js` which is not an entrypoint itself. This was not reflected by the original result of this method.

packages/angular_devkit/build_webpack/src/utils.ts
manfredsteyer: As agreed on slack, I've removed the `initial` property now b/c it's not needed for the time being.

packages/angular_devkit/build_webpack/src/utils.ts
manfredsteyer: Btw: This new logic was original part of the `IndexHtmlPlugin`

alan-agius4:
@manfredsteyer, my point was more because in the html you have duplicate css tags. But yeah let’s do the index generation in a separate PR.

For the time being let’s focus on:
1. Emitting all the needed files and with the correct content.
2. Add a couple of test for the above.
2. Make this green.
3. Disable differential loading feature with a flag in code

manfredsteyer:
Ah, now I see. The ``generate-index-html`` function uses the following code that was originally part of the ``IndexHtmlPlugin``:

```
for (const file of fileNames) {

    if (existingFiles.has(file)) {
      continue;
    }

    [...]
}
```

This filters out duplicate files before css and script references are written.

packages/angular_devkit/build_webpack/src/index.ts
```diff
@@ -11,4 +11,4 @@ export * from './webpack-dev-server';
 
 export * from './plugins/architect';
 
-export { EmittedFiles } from './utils';
+export { getEmittedFiles, EmittedFiles } from './utils';
```
alan-agius4: This is intended to be private. You need to get the emitted files, from the compilation result.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
```diff
@@ -67,41 +68,38 @@ export class IndexHtmlWebpackPlugin {
 
       const loadOutputFile = (name: string) => compilation.assets[name].source();
 
-      // Get all files for selected entrypoints
-      const unfilteredSortedFiles: string[] = [];
+      // Get all files for script references
+      const files = getEmittedFiles(compilation);
```
alan-agius4: This is not meant to be used like this. the `getEmittedFiles` is not intended to be part of the public api.

packages/angular_devkit/core/src/index.ts
```diff
@@ -23,3 +23,5 @@ export {
   logging,
   terminal,
 };
+
+export const ENABLE_DIFFERENTIAL_LOADING = false;
```
alan-agius4: This is wrong. core has nothing to do with core package. 

I suggest override in the builder.

Ie, change;
```
const differentialLoading = isDifferentialLoadingNeeded(
    root: getSystemPath(root),projectRoot, target || projectTs.ScriptTarget.ES5)
```

to 
```
// const differentialLoading = isDifferentialLoadingNeeded(
    root: getSystemPath(root),projectRoot, target || projectTs.ScriptTarget.ES5)
const differentialLoading = false;
```

And xdescribe the builder specs for the time being.

tests/angular_devkit/build_angular/test-differential-loading/browserslist
```diff
@@ -0,0 +1,4 @@
+last 2 versions
```
alan-agius4: Delete unused files

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
```diff
@@ -12,6 +12,9 @@ import * as path from 'path';
 import { basename, normalize } from '@angular-devkit/core';
 import { ExtraEntryPoint, ExtraEntryPointClass } from '../../../browser/schema';
 import { SourceMapDevToolPlugin } from 'webpack';
+import { WebpackConfigOptions } from '../build-options';
+import * as ts from './common';
+import * as typescript from 'typescript';
```
alan-agius4: small not `* as ts`

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/generate-index-html.ts
```diff
@@ -15,8 +15,11 @@ import {
 
 const parse5 = require('parse5');
 
+export type LoadOutputFileFunctionType = (file: string) => Promise<string>;
 
-export type LoadOutputFileFunctionType = (file: string) => string;
+export interface EntryInfo {
```
alan-agius4: Seems unused.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
```diff
@@ -89,3 +92,13 @@ export function getSourceMapDevTool(
     append: hiddenSourceMap ? false : undefined,
   });
 }
+
+export function getEsVersionForFileName(
```
alan-agius4: Nit: We don't typically use this pattern

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/generate-index-html.ts
```diff
@@ -27,7 +30,9 @@ export interface GenerateIndexHtmlParams {
   deployUrl?: string;
   sri: boolean;
   // the files emitted by the build
-  unfilteredSortedFiles: CompiledFileInfo[];
+  unfilteredUnsortedFiles: CompiledFileInfo[];
+  // list with entry points in the order we need to reference their bundles
+  entries: string[];
```
alan-agius4: Nit: for consistency I'd name it `entrypoints`

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
```diff
@@ -67,41 +68,38 @@ export class IndexHtmlWebpackPlugin {
...
-          unfilteredSortedFiles.push(...files);
-
-          if (this._options.noModuleEntrypoints.includes(entryName)) {
-            files.forEach(file => noModuleFiles.add(file));
-          } else {
-            files.forEach(file => otherFiles.add(file));
-          }
+
+      // Find all files which have been created out of configured noModule entry points
+      compiledFiles.forEach(f => {
```
alan-agius4: I strongly suggest that we don't do any index modifications as part of this PR. We should keep that to another PR.

The logic here is different then the previous one as `compiledFiles` will return all files that have been emitted by webpack and not just the entrypoints code.

This includes, images, lazy loaded js, lazy loaded routes, lazy loaded css. which in this case need to be filtered out and not written by the index generator.

We need tests on for this, as this currently breaks the current workflow and hence, we cannot even merge this PR if we turn off the feature.

How are you filling out lazy loaded js files?

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
alan-agius4: If you have access to a webpack compilation, you have the chunk of data that you need, the emittedFiles, will return way more files then you need.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts
manfredsteyer: Which pattern do you mean?

alan-agius4:
```
  params: {
    scriptTargetOverride: typescript.ScriptTarget | undefined,
    esVersionInFileName: boolean,
  }): string {
```

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -5,47 +5,55 @@
...
+  schema, virtualFs,
 } from '@angular-devkit/core';
 import { NodeJsSyncHost } from '@angular-devkit/core/node';
 import * as fs from 'fs';
-import { Observable, from, of } from 'rxjs';
+import { Observable, combineLatest, from, of } from 'rxjs';
 import { concatMap, map, switchMap } from 'rxjs/operators';
-import * as ts from 'typescript'; // tslint:disable-line:no-implicit-dependencies
+import * as ts from 'typescript';
+import webpack = require('webpack');
```
alan-agius4: from webpack

PR #14090, opened 2019-04-05
feat(@angular/cli): automatically update angular builders when CLI is updated

clydin:


packages/angular/cli/package.json
```diff
@@ -43,6 +43,12 @@
     "uuid": "^3.3.2"
   },
   "ng-update": {
-    "migrations": "@schematics/angular/migrations/migration-collection.json"
+    "migrations": "@schematics/angular/migrations/migration-collection.json",
+    "packageGroup": {
+      "@angular/cli": "0.0.0",
+      "@angular-devkit/build-angular":  "0.0.0",
+      "@angular-devkit/build-ng-packagr": "0.0.0",
+      "@angular-devkit/build-webpack": "0.0.0"
```
alan-agius4: Is this needed?

Isn’t this a transitive dependency of build-angular?

alan-agius4:
Can we also reference this issue?

https://github.com/angular/angular-cli/issues/13581

packages/angular/cli/package.json
clydin: Could be used directly as well.  If not present here the 7.x version would still be left after an upgrade (and non-functional) in that scenario.

Splaktar:
As asked in https://github.com/angular/angular-cli/pull/13942#issuecomment-482682871, should this change have also closed https://github.com/angular/angular-cli/pull/13942?

PR #14091, opened 2019-04-05
feat(@angular-devkit/schematics): Source files via HTTP

Brocco:


vikerman:
Arbitrary remote file fetch seems dangerous.

PR #14097, opened 2019-04-07
Refactor/Cleanup browser builder

clydin:
Switches the browser builder to use the existing Webpack configuration generation utility functions.  Also, provides improved and centralized parsing of the tsconfig file.


clydin:
TypeScript is not being updated in the _update_ tests; a minimum of 2.7 is required for that function.

filipesilva:
I think that'll be addressed in https://github.com/angular/angular-cli/pull/14054, where I change the update tests to also update Angular (and thus the TS peerdep)

PR #14106, opened 2019-04-08
Improve handling of nomodules and modules in index generation

alan-agius4:
**fix(@angular-devkit/build-webpack): add dot to file extension in emitted files**

**feat(@angular-devkit/build-webpack): include entrypoints in emitted files**
Entrypoints might have other files associate with them such as runtime.js, it is is paramount to keep the relation between them especially when this result is needed to generate an index file

 **feat(@angular-devkit/build-angular): improve handling of nomodules and modules in index generation**

Since when having differential loading we already know which files originated from which build. We shouldn't need to merge and transform this data.

With this change, the index generator accepts a couple of new inputs.
1. `files` - used for Js and CSS files which require nomodule nor module attributes
2. `moduleFiles` - Js files that need to have a `module` attribute
3. `noModuleFiles`  - Js files that need to have a `nomodule` attribute
4. `entrypoints` - used to sort the insertion of files in the HTML file

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/generate-index-html.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/generate-index-html.ts
@@ -16,37 +16,39 @@ import {
 const parse5 = require('parse5');
 
 
-export type LoadOutputFileFunctionType = (file: string) => string;
+export type LoadOutputFileFunctionType = (file: string) => Promise<string>;
 
-export interface GenerateIndexHtmlParams {
-  // input file name (e. g. index.html)
+export interface GenerateIndexHtmlOptions {
+  /* Input file name (e. g. index.html) */
   input: string;
-  // contents of input
+  /* Input contents */
   inputContent: string;
   baseHref?: string;
   deployUrl?: string;
   sri: boolean;
-  // the files emitted by the build
-  unfilteredSortedFiles: CompiledFileInfo[];
-  // additional files that should be added using nomodule
-  noModuleFiles: Set<string>;
-  // function that loads a file
-  // This allows us to use different routines within the IndexHtmlWebpackPlugin and
-  // when used without this plugin.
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/plugins/generate-index-html.ts
```diff
@@ -55,41 +50,30 @@ export interface CompiledFileInfo {
...
+  const mergedFiles = [...noModuleFiles, ...moduleFiles, ...unfilteredSortedFiles];
+  for (const entrypoint of entrypoints) {
+    const files = mergedFiles.filter(x => x.name === entrypoint);
+    for (const { extension, fileName } of files) {
+      switch (extension) {
+        case 'js':
+          scripts.push(fileName);
+          break;
+        case 'css':
+          stylesheets.push(fileName);
```
alan-agius4: Self note: use a set and .add insert to avoid duplicates for the start.

alan-agius4:
//cc @manfredsteyer

manfredsteyer:
Looked for some edge cases and it LGTM!

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
```diff
@@ -65,45 +66,37 @@ export class IndexHtmlWebpackPlugin {
...
-          unfilteredSortedFiles.push(...files);
-
-          if (this._options.noModuleEntrypoints.includes(entryName)) {
-            files.forEach(file => noModuleFiles.add(file));
-          } else {
-            files.forEach(file => otherFiles.add(file));
-          }
+      const files: FileInfo[] = [];
+      const noModuleFiles: FileInfo[] = [];
+
```
alan-agius4: Note to self: remove this line

PR #14114, opened 2019-04-09
feat(@ngtools/webpack): support loadchildren string syntax in Ivy

filipesilva:


filipesilva:
Superseded (and necessary for) https://github.com/angular/angular-cli/pull/14147.

PR #14115, opened 2019-04-09
feat(@schematics/angular): add lazy module path fixer

phenomnomnominal:
Hey @alexeagle, here's the start of this. I'm not 100% sure on the approach - my understanding is that if I do it this way I may need to check the compiled lint rule into the repo? 

Keen for comments before i go any further.


filipesilva:
@clydin can you have a look at this PR? @phenomnomnominal says he can't get the fixer to actually run in the tests.

packages/schematics/angular/migrations/update-8/update-lazy-module-paths_spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  describe('Migration to import() style lazy routes', () => {
+    beforeEach(() => {
+      tree = new UnitTestTree(new EmptyTree());
+      tree.create(packageJsonPath, JSON.stringify(packageJson, null, 2));
+    });
+
+    it('should replace the module path string', () => {
+      tree.create(lazyRoutePath, lazyRoute);
+      tree = schematicRunner.runSchematic('migration-07', defaultOptions, tree);
```
alexeagle: I think your test problem may be here - you should be running `migration-08` since that's where the fixer is registered

phenomnomnominal: The other test files in `update-8` also use `migrattion-07` 😞 

```
"migration-07": {
    "version": "8.0.0",
    "factory": "./update-8",
    "description": "Update an Angular CLI project to version 8."
}
```

I've verified that the actual tasks are running and it gets to `context.addTask(new TslintFixTask({`, but then the fix itself isn't run.

I'll keep poking around :)

PR #14120, opened 2019-04-10
feat(@angular-devkit/build-angular): wire up index generation for differential loading

alan-agius4:
feat(@angular-devkit/build-angular): wire up index generation for differential loading

This change wires up the index generation for multiple bundles, this however doesn't enable differential loading for users just yet. As this will be done in a separate PR to change over specs to conform with the new files

Most of the testing will be covered when within this on by the current e2e and spec large. (Currently working on in a separate PR)


packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -113,11 +115,16 @@ export function buildWebpackConfig(
...
     : new logging.NullLogger();
 
+
+  // At the moment, only the browser builder supports differential loading
+  // However this config generation is used by multiple builders such as dev-server
+  // const builderInfo = additionalOptions.builderInfo;
   const scriptTarget = tsConfig.options.target;
-  // todo enabe when differential loading is complete
-  // const differentialLoading = isDifferentialLoadingNeeded(projectRoot, scriptTarget);
+  // const differentialLoading =
```
alan-agius4: Not sure, if there is a better way apart from providing an addition parameter to the upper function and pass it down. 

But I think for the time being, this is enough

PR #14122, opened 2019-04-10
feat(@angular-devkit/build-angular): support web workers in karma bui…

filipesilva:
…lder

Followup to https://github.com/angular/angular-cli/pull/13700, the karma builder did not get the same support but it should.

### DIFF
--- a/packages/angular_devkit/build_angular/src/karma/schema.json
+++ b/packages/angular_devkit/build_angular/src/karma/schema.json
@@ -185,6 +185,10 @@
       "items": {
         "type": "string"
       }
+    },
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/karma/schema.json
```diff
@@ -185,6 +185,10 @@
       "items": {
```
alan-agius4: Nit: can you add this also to https://github.com/angular/angular-cli/blob/master/packages/angular/cli/lib/config/schema.json?

filipesilva:
@alan-agius4 right you are, added.

PR #14126, opened 2019-04-10
Partial update of Angular schematics with new workspace rules

clydin:
This adds initial infrastructure, fixes several bugs, and updates several of the Angular schematics.  Each schematic update is in a separate commit.  Tests needed to be updated to use the async version of the test runner but are otherwise unmodified.

### DIFF
--- a/packages/schematics/angular/class/schema.json
+++ b/packages/schematics/angular/class/schema.json
@@ -52,7 +52,6 @@
     }
   },
   "required": [
-    "name",
-    "project"
+    "name"
   ]
 }
--- a/packages/schematics/angular/utility/workspace.ts
+++ b/packages/schematics/angular/utility/workspace.ts
@@ -1,0 +1,90 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { virtualFs, workspaces } from '@angular-devkit/core';
+import { Rule, Tree } from '@angular-devkit/schematics';
+
+function createHost(tree: Tree): workspaces.WorkspaceHost {
+  return {
+    async readFile(path: string): Promise<string> {
+      const data = tree.read(path);
+      if (!data) {
+        throw new Error('File not found.');
+      }
+
+      return virtualFs.fileBufferToString(data);
+    },
+    async writeFile(path: string, data: string): Promise<void> {
+      return tree.overwrite(path, data);
+    },
+    async isDirectory(path: string): Promise<boolean> {
+      // approximate a directory check
+      return !tree.exists(path) && tree.getDir(path).subfiles.length > 0;
+    },
+    async isFile(path: string): Promise<boolean> {
+      return tree.exists(path);
+    },
+  };
+}
+
+export function updateWorkspace(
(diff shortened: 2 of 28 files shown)


packages/schematics/angular/utility/workspace.ts
```diff
@@ -0,0 +1,90 @@
...
+export function updateWorkspace(
+  updater: (workspace: workspaces.WorkspaceDefinition) => void | PromiseLike<void>,
+): Rule;
+export function updateWorkspace(
+  workspace: workspaces.WorkspaceDefinition,
+): Rule;
+export function updateWorkspace(
+  updaterOrWorkspace: workspaces.WorkspaceDefinition
+    | ((workspace: workspaces.WorkspaceDefinition) => void | PromiseLike<void>),
+): Rule {
```
filipesilva: Nit: why do we need the extra signatures above? The arg typing here looks ok.

packages/schematics/angular/utility/workspace.ts
```diff
@@ -0,0 +1,90 @@
...
+ * @param project The project which will have its default path generated.
+ */
+export function buildDefaultPath(project: workspaces.ProjectDefinition): string {
+  const root = project.sourceRoot ? `/${project.sourceRoot}/` : `/${project.root}/src/`;
+  const projectDirName = project.extensions['projectType'] === 'application' ? 'app' : 'lib';
+
+  return `${root}${projectDirName}`;
+}
+
+export async function createDefaultPath(tree: Tree, projectName: string): Promise<string> {
```
filipesilva: Nit: `buildDefaultPath` and `createDefaultPath` mostly mean the same. Either one should be renamed, or there should be a single one.

filipesilva: Nit: `buildDefaultPath` and `createDefaultPath` mostly mean the same. Either one should be renamed, or there should be a single one.

packages/schematics/angular/utility/workspace.ts
filipesilva: Nit: why do we need the extra signatures above? The arg typing here looks ok.

packages/schematics/angular/class/schema.json
```diff
@@ -52,7 +52,6 @@
     }
   },
   "required": [
-    "name",
-    "project"
```
filipesilva: Shouldn't project here still be required? It will error out inside `createDefaultPath` if it's not there.

clydin: The smart default will provide a value if not specified so it’s not actually required to be provided by the user.  The schema type generator doesn’t know about smart defaults though so it gives the property an optional type.  Marking it as required “fixes” the schema type generator but is technically incorrect.  This results in issues when other consumers attempt to use the schemas (UI based ng generate utilities for instance) since they will require the user to specify the property even though a value will automatically be provided.

packages/schematics/angular/utility/workspace.ts
clydin: Agree.  The former will be phased out and removed once everything is transferred over

packages/schematics/angular/utility/workspace.ts
clydin: cleaner in an IDE and allows for eventual separate jsdoc comments tailored to each overload.  The second form may also gain a read options parameter at a later time.

PR #14139, opened 2019-04-12
feat(@angular-devkit/core): make new workspace API public

clydin:


packages/angular_devkit/core/src/_golden-api.ts
```diff
@@ -51,3 +51,7 @@ export * from './utils/lang';
 export * from './virtual-fs/path';
 export * from './virtual-fs/host/index';
 // End virtualFs namespace
+
+// Start workspace namespace
+export * from './workspace';
```
alan-agius4: This should be `workspace/index` are api guardian uses classic module resolution

clydin: It passes though?

alan-agius4: Yeah, cause API guardian completely ignores such exports/imports.

clydin: Updated.

PR #14143, opened 2019-04-12
feat(	@schematics/angular): add more information and browsers in browserlist

alan-agius4:
//cc @brandonroberts since you mentioned this a couple of days ago.

### DIFF
--- a/packages/schematics/angular/application/files/browserslist.template
+++ b/packages/schematics/angular/application/files/browserslist.template
@@ -1,11 +1,13 @@
-# This file is currently used by autoprefixer to adjust CSS to support the below specified browsers
+# This file is used by the build system to adjust CSS and JS output to support the specified browsers below.
 # For additional information regarding the format and rule options, please see:
 # https://github.com/browserslist/browserslist#queries
-#
-# For IE 9-11 support, please remove 'not' from the last line of the file and adjust as needed
+
+# Googlebot uses an older version of Chrome
+# For additional information see: https://developers.google.com/search/docs/guides/rendering
 
 > 0.5%
 last 2 versions
 Firefox ESR
 not dead
-not IE 9-11
+not IE 9-11 # For IE 9-11 support, remove 'not'.
+not Chrome 41 # For Googlebot support, remove 'not'.


packages/schematics/angular/application/files/browserslist.template
```diff
@@ -1,11 +1,16 @@
...
+# For IE 9-11 support, remove '#' next to 'IE 9-11'.
 
 > 0.5%
 last 2 versions
 Firefox ESR
 not dead
-not IE 9-11
\ No newline at end of file
+# IE 9-11
+# Chrome 41
```
alan-agius4: @mgechev / @alexeagle should we enable this by default? GoogleBot always requires ES5 bundles.

packages/schematics/angular/application/files/browserslist.template
```diff
@@ -1,11 +1,16 @@
-# This file is currently used by autoprefixer to adjust CSS to support the below specified browsers
+# This file is used by the build system to adjust CSS and JS output to support the below specified browsers
```
mgechev: `the below specified browsers` -> `the specified browsers below`

packages/schematics/angular/application/files/browserslist.template
mgechev: IMO, we should enable support for Chrome 41.

brandonroberts:
LGTM after commented issues are addressed

packages/schematics/angular/application/files/browserslist.template
```diff
@@ -1,11 +1,15 @@
...
+# For IE 9-11 support, remove '#' next to 'IE 9-11'.
 
 > 0.5%
 last 2 versions
 Firefox ESR
 not dead
-not IE 9-11
\ No newline at end of file
+# IE 9-11
+Chrome 41 # Googlebot
```
mgechev: @clydin had some good arguments for not including Chrome 41. // cc @alexeagle

alan-agius4: Disabling again by default.

packages/schematics/angular/application/files/browserslist.template
```diff
@@ -1,11 +1,16 @@
...
+# This file is used by the build system to adjust CSS and JS output to support the specified browsers below.
 # For additional information regarding the format and rule options, please see:
 # https://github.com/browserslist/browserslist#queries
-#
-# For IE 9-11 support, please remove 'not' from the last line of the file and adjust as needed
+
+# Googlebot uses an older version of Chrome
+# For additional information see: https://developers.google.com/search/docs/guides/rendering
+
+# For Googlebot support, remove '#' next to 'Chrome 41'.
```
alexeagle: why aren't these comments adjacent to the line you're supposed to change? because if the file gets long then not all instructions are on the first screen?

alan-agius4: Good point

PR #14145, opened 2019-04-12
feat(@schematics/angular): migrate existing ES5 projects to support differential loading

alan-agius4:
feat(@schematics/angular): migrate existing ES5 projects to support differential loading

With this change, differential loading will be enabled as we add an non ever green browser and change the tsconfig script target to es2015.

### DIFF
--- a/packages/schematics/angular/migrations/update-8/differential-loading.ts
+++ b/packages/schematics/angular/migrations/update-8/differential-loading.ts
@@ -1,0 +1,113 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {
+  JsonParseMode,
+  isJsonObject,
+  join,
+  normalize,
+  parseJson,
+  parseJsonAst,
+} from '@angular-devkit/core';
+import { Rule, Tree } from '@angular-devkit/schematics';
+import {
+  findPropertyInAstObject,
+  insertPropertyInAstObjectInOrder,
+} from '../../utility/json-utils';
+
+// tslint:disable-next-line:max-line-length
+const browserslistContent = `# This file is used by the build system to adjust CSS and JS output to support the specified browsers below.
+# For additional information regarding the format and rule options, please see:
+# https://github.com/browserslist/browserslist#queries
+
+# Googlebot uses an older version of Chrome
+# For additional information see: https://developers.google.com/search/docs/guides/rendering
+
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/migrations/update-8/ts-script-target.ts
```diff
@@ -0,0 +1,51 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { JsonParseMode, parseJsonAst } from '@angular-devkit/core';
+import { Rule, Tree } from '@angular-devkit/schematics';
+import {
+  findPropertyInAstObject,
+  insertPropertyInAstObjectInOrder,
+} from '../../utility/json-utils';
+
+export function updateTsConfigScriptTarget(): Rule {
```
mgechev: We do not have a guarantee that this would enable differential loading.

We'd have to parse browserlist to make sure es5 browser support is required.

alan-agius4: Yeah true though this is solely to update tsconfig target to es2015 which is one of the criteria’s to enable differential loading.

At the point we don’t really need any knowledge of differential loading is enabled or not.

Anyways, this is obsolete, since we will not update the current projects to es2015.

packages/schematics/angular/migrations/update-8/differential-loading.ts
```diff
@@ -0,0 +1,98 @@
...
+      if (!isJsonObject(project)) {
+        continue;
+      }
+      if (typeof project.root != 'string' || project.projectType !== 'application') {
+        continue;
+      }
+
+      const browserslistPath = join(normalize(project.root), '/browserslist');
+      const source = tree.read(browserslistPath);
+      if (!source) {
```
mgechev: What does it mean to not have the source? Should we create the browserlist in this case?

alan-agius4: It means, that it will fallback to the default browserlist configuration or it can mean that a users specified the browserlist configuration in another way. 

I think creating this file might be a good approach

mgechev: Sounds good. If we create the file, are we going to overwrite the user's configuration if they are not using the defaults? This might be a bit frustrating for them, I guess?

alan-agius4: Only if they are using `.browserslistrc`

alan-agius4: Only if they are using `.browserslistrc`

PR #14147, opened 2019-04-12
Update to Angular beta.12

filipesilva:
Blocked on an Angular version that supports TS 3.4.x, should be the upcoming 8.0.0-beta.12.


packages/angular/cli/models/schematic-command.ts
```diff
@@ -492,7 +492,7 @@ export abstract class SchematicCommand<
     }
 
     const pathOptions = o ? this.setPathOptions(o, workingDir) : {};
-    let input = Object.assign(pathOptions, args);
+    let input = { ...pathOptions, ...args };
```
filipesilva: Without this change, I got the following TS error:
```
packages/angular/cli/models/schematic-command.ts:500:5 - error TS2322: Type '{ '--'?: string[] | undefined; } | { '--'?: string[] | undefined; }' is not assignable to type '{ [name: string]: string
; } & Arguments'.
  Type '{ '--'?: string[] | undefined; }' is not assignable to type '{ [name: string]: string; } & Arguments'.
    Type '{ '--'?: string[] | undefined; }' is not assignable to type '{ [name: string]: string; }'.
      Property ''--'' is incompatible with index signature.
        Type 'string[] | undefined' is not assignable to type 'string'.
          Type 'undefined' is not assignable to type 'string'.

500     input = {
```

As far as I can tell, that's because `Object.assign` makes an intersection type (`{ [name: string]: string; } & Arguments`, which doesn't let you have the undefined) while `{...something, ...somethingElse}` does a sort of union type instead.

tests/legacy-cli/e2e/tests/build/lazy-load-syntax.ts
```diff
@@ -125,7 +125,8 @@ export default async function () {
     buildTarget['options']['experimentalImportFactories'] = true;
   });
   // Both Ivy and View Engine should support it.
-  await replaceLoadChildren(`() => import('./lazy/lazy.module').then(m => m.LazyModule)`);
+  // TODO: remove `as any` after https://github.com/angular/angular/pull/29832 is released.
```
filipesilva: Related to https://github.com/angular/angular/pull/29832

packages/angular/cli/models/schematic-command.ts
alexeagle: thanks for explaining

PR #14149, opened 2019-04-12
Update golden files for core and architect

alan-agius4:
Add analytics to core golden files
Add architect api to golden files

Remove `_golden-api.ts` from architect as we are no longer exporting namespaces


alan-agius4:
Because the new API have never been added to the `golden_api.d.ts`

PR #14160, opened 2019-04-14
Builder Transforms Standardization

clydin:
This adds Webpack configuration transform support to the server builder and also standardizes the transform usage across the currently support builders.

### DIFF
--- a/packages/angular_devkit/build_angular/src/server/index.ts
+++ b/packages/angular_devkit/build_angular/src/server/index.ts
@@ -5,23 +5,14 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { BuilderOutput, createBuilder } from '@angular-devkit/architect';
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_angular/src/server/index.ts
```diff
@@ -122,65 +65,41 @@ export default createBuilder<
...
-  const tsConfigPath = getSystemPath(normalize(resolve(root, normalize(options.tsConfig))));
-  const tsConfig = readTsconfig(tsConfigPath);
+}
 
-  const projectTs = requireProjectModule(getSystemPath(projectRoot), 'typescript') as typeof ts;
+export default createBuilder<json.JsonObject & BuildWebpackServerOptions, ServerBuilderOutput>(
+  execute,
+);
 
-  const supportES2015 = tsConfig.options.target !== projectTs.ScriptTarget.ES3
```
alan-agius4: Good catch!

alan-agius4: good catch

PR #14166, opened 2019-04-15
feat(@angular-devkit/build-angular): deprecate `es5BrowserSupport` op…

alan-agius4:
…tion in browser builder

In future, this will be determined from the list of supported browsers specified in the 'browserslist' file.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -308,7 +308,7 @@
     "es5BrowserSupport": {
       "description": "Enables conditionally loaded ES2015 polyfills.",
       "type": "boolean",
-      "default": false
+      "x-deprecated": "This will be determined from the list of supported browsers specified in the 'browserslist' file."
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -308,7 +308,8 @@
     "es5BrowserSupport": {
       "description": "Enables conditionally loaded ES2015 polyfills.",
       "type": "boolean",
-      "default": false
+      "default": false,
```
clydin: I think we should remove the default which would allow false to actually mean disabled and undefined (not explicitly specified) means auto-detect.

HitomiTenshi:
@alan-agius4 Just out of curiosity, will we be able to somehow either disable loading polyfills entirely, or load different `browserslist` files now that `es5BrowserSupport` is deprecated?

We are targeting multiple environments with multiple configurations in `angular.json` and we don't need polyfills for our `electron` target.

PR #14170, opened 2019-04-15
build: use Bazel from npm

alexeagle:
Also hook up bazel format/lint


package.json
```diff
@@ -20,6 +20,10 @@
   ],
   "scripts": {
     "admin": "node ./bin/devkit-admin",
```
alan-agius4: Nit: would be nice to have `"bazel": "bazel",`

clydin: `npx bazel`?

filipesilva: Having the script proper is better I feel. It makes it pretty obvious that it's the right way of using `bazel`, whether via `yarn` or `npm`, and it's more ergonomic for scripts too. I don't know if all the docker images we use have `npx` either, but I guess they would.

clydin: It comes with npm.  If you’re only using yarn you should be able to just do ‘yarn bazel’

alexeagle: yeah you can `yarn bazel` if you want to pick your targets.

.circleci/config.yml
```diff
@@ -105,7 +105,7 @@ jobs:
     parallelism: 4
     steps:
       - attach_workspace: *attach_options
-      - run: xvfb-run -a node ./tests/legacy-cli/run_e2e --nb-shards=${CIRCLE_NODE_TOTAL} --shard=${CIRCLE_NODE_INDEX}
+      - run: PATH=~/.npm-global/bin:$PATH xvfb-run -a node ./tests/legacy-cli/run_e2e --nb-shards=${CIRCLE_NODE_TOTAL} --shard=${CIRCLE_NODE_INDEX}
```
filipesilva: The `e2e-cli-ng-snapshots` job also needs this change. It only runs on master.

scripts/validate-licenses.ts
```diff
@@ -73,6 +73,14 @@ const ignoredPackages = [
   'true-case-path@1.0.2', // Apache-2.0 but broken license in package.json
   'uws@9.14.0', // zlib license
 
+  // * Need to publish these with LICENSE files
+  // see https://github.com/bazelbuild/rules_nodejs/pull/696
```
filipesilva: https://github.com/bazelbuild/rules_nodejs/pull/696 is now merged, can we get a release that removes this TODO?

.circleci/config.yml
```diff
@@ -13,14 +13,15 @@
 # If you change the cache key prefix, also sync the restore_cache fallback to match.
 # Keep the static part of the cache key as prefix to enable correct fallbacks.
 # See https://circleci.com/docs/2.0/caching/#restoring-cache for how prefixes work in CircleCI.
-var_1: &docker_image circleci/node:10.12
-var_2: &cache_key angular_devkit-0.10.0-{{ checksum "yarn.lock" }}
+var_1: &default_docker_image circleci/node:10.12
+var_2: &browsers_docker_image circleci/node:10.12-browsers
```
filipesilva: Is this needed? We use `puppeteer` for our e2e tests so I didn't expect it to be necessary.

alexeagle: sadly headless chrome still requires dynamic linked dependency on libx11 which is pretty lame. We should raise with the Chrome team.

.circleci/config.yml
alexeagle: thanks

scripts/validate-licenses.ts
alexeagle: well, the npm version numbers for the bazel packages are identical to the upstream version numbers. If I publish `@bazel/bazel@0.24.2` it will be confusing since the latest is https://github.com/bazelbuild/bazel/releases/tag/0.24.1

PR #14172, opened 2019-04-15
Edit and add doc for CLI usage analytics options

jbogarthyde:
I don't know if the target should be master or master-and-patch. Is `ng analytics` going into version 8?

### DIFF
--- a/packages/angular/cli/commands/analytics.json
+++ b/packages/angular/cli/commands/analytics.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/schema",
   "$id": "ng-cli://commands/analytics.json",
-  "description": "Configures usage metric gathering for the Angular CLI. See http://angular.io/MORE_INFO_HERE",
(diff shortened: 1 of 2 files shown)


packages/angular/cli/commands/analytics.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/schema",
   "$id": "ng-cli://commands/analytics.json",
-  "description": "Configures usage metric gathering for the Angular CLI. See http://angular.io/MORE_INFO_HERE",
+  "description": "Configures the gathering of Angular CLI usage metrics. See https://next.angular.io/cli/usage-analytics-gathering.",
```
alexeagle: hmm, as soon as we ship 8.0 final this should no longer be `next` - we ought to have a way for CLI to know if you're currently running from the `next` npm tag and redirect all the AIO links correctly. But outside the scope of this change...

PR #14173, opened 2019-04-16
Complete Angular Schematic updates to new workspace rules

clydin:
This fixes several bugs and updates the remainder of the Angular schematics. Each schematic update is in a separate commit. Tests needed to be updated to use the async version of the test runner but are otherwise unmodified.


packages/angular_devkit/core/src/workspace/json/writer.ts
```diff
@@ -175,7 +175,12 @@ function stringify(
 
     return content.replace(/\n/g, spacing);
   } else {
-    return JSON.stringify(value);
+    const content = JSON.stringify(value);
+    if (content.length > 60) {
+      return stringify(value, true, depth, indent);
```
filipesilva: Can you add a comment explaining why this is necessary, and what the 60 is?

clydin: removed.  It's not needed.

PR #14180, opened 2019-04-16
fix(@ngtools/webpack): replace bootstrap code under Ivy unless running in JIT mode

alan-agius4:
fix(@ngtools/webpack): replace bootstrap code under Ivy unless running in JIT mode

The bootstrap code always needs to be replaced if not running in JIT mode as other wise the entire compiler will be pulled in because we will not replace the bootstrapping code to use `platform-browser` instead of `platform-browser-dynamic`.

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -948,13 +948,13 @@ export class AngularCompilerPlugin {
             this._normalizedLocale));
         }
 
-        if (this._useFactories) {
-          // Replace bootstrap in browser AOT.
+        if (!this._JitMode) {
+          // Replace bootstrap in browser non JIT Mode.
           this._transformers.push(replaceBootstrap(
             isAppPath,
             getEntryModule,
             getTypeChecker,
-            !!this._compilerOptions.enableIvy,
+            this._useFactories,
           ));
         }
       } else if (this._platform === PLATFORM.Server) {
--- a/tests/legacy-cli/e2e/tests/build/prod-build.ts
+++ b/tests/legacy-cli/e2e/tests/build/prod-build.ts
@@ -1,30 +1,37 @@
-import {join} from 'path';
-import {readdirSync} from 'fs';
-import {expectFileToExist, expectFileToMatch} from '../../utils/fs';
-import {ng} from '../../utils/process';
-import {expectGitToBeClean} from '../../utils/git';
+import { readdirSync } from 'fs';
+import { join } from 'path';
+import { getGlobalVariable } from '../../utils/env';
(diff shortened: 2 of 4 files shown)


packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -278,9 +278,10 @@ export class AngularCompilerPlugin {
       );
     }
 
-    if (!this._JitMode && !this._compilerOptions.enableIvy) {
-      // Only attempt to use factories when AOT and not Ivy.
-      this._useFactories = true;
+    if (!this._JitMode) {
+      // Only attempt to use factories when AOT or Ivy with factory shims.
+      this._useFactories = !this._compilerOptions.enableIvy
```
alan-agius4: @filipesilva was this not included on purpose of there it was just an oversight?

alan-agius4: @filipesilva was this not included on purpose of there it was just an oversight?

tests/legacy-cli/e2e/tests/build/prod-build.ts
```diff
@@ -1,30 +1,37 @@
...
-      const mainES5 = readdirSync('./dist/test-project').find(name => !!name.match(/main-es5.[a-z0-9]+\.js/));
-      expectFileToMatch(`dist/test-project/${mainES5}`, /bootstrapModuleFactory\(/);
+  const argv = getGlobalVariable('argv');
+  const ivyProject = argv['ivy'];
+  const bootstrapRegExp = ivyProject
+    ? /bootstrapModule\([a-z]+\)\./
+    : /bootstrapModuleFactory\([a-z]+\)\./;
 
-      const mainES2015 = readdirSync('./dist/test-project').find(name => !!name.match(/main-es2015.[a-z0-9]+\.js/));
-      expectFileToMatch(`dist/test-project/${mainES5}`, /bootstrapModuleFactory\(/);
```
alan-agius4: This regexp was actually matching the entire contents of the bundle, and in fact this test was not failing for Ivy

tests/legacy-cli/e2e/tests/build/prod-build.ts
```diff
@@ -1,30 +1,37 @@
...
+import { readdirSync } from 'fs';
+import { join } from 'path';
+import { getGlobalVariable } from '../../utils/env';
+import { expectFileToExist, expectFileToMatch } from '../../utils/fs';
+import { expectGitToBeClean } from '../../utils/git';
+import { ng } from '../../utils/process';
 
 
-export default function() {
+export default async function () {
```
filipesilva: Can you also add a basic size check here please? Allowing something like the current size plus or minus 10%.

PR #14181, opened 2019-04-16
feat(@schematics/angular): add lazy module path fixer

phenomnomnominal:
Apologies, I messed up my original PR trying to rebase stuff :/

@alexeagle could you please ping the appropriate people for help with the tests?

Sorry again!

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     ]
   },
   "dependencies": {
+    "@phenomnomnominal/tsquery": "3.0.0",
     "@types/debug": "^4.1.2",
     "@types/node-fetch": "^2.1.6",
     "@types/progress": "^2.0.3",
--- a/packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts
+++ b/packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts
@@ -1,0 +1,30 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {
+  Rule,
+  SchematicContext,
+  Tree,
+} from '@angular-devkit/schematics';
+import {
+  TslintFixTask,
+} from '@angular-devkit/schematics/tasks';
+import * as path from 'path';
+
+export const updateLazyModulePaths = (): Rule => {
+  return (_: Tree, context: SchematicContext) => {
+    context.addTask(new TslintFixTask({
+      rulesDirectory: path.join(__dirname, 'rules'),
+      rules: {
+        'no-lazy-module-paths': [true],
+      },
+    }, {
+      includes: '**/*.ts',
+      silent: false,
+    }));
+  };
+};
--- a/packages/schematics/angular/migrations/update-8/update-lazy-module-paths_spec.ts
+++ b/packages/schematics/angular/migrations/update-8/update-lazy-module-paths_spec.ts
@@ -1,0 +1,85 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { getSystemPath, normalize, virtualFs } from '@angular-devkit/core';
+import { TempScopedNodeJsSyncHost } from '@angular-devkit/core/node/testing';
+import { HostTree } from '@angular-devkit/schematics';
+import { SchematicTestRunner, UnitTestTree } from '@angular-devkit/schematics/testing';
+
+describe('Migration to version 8', () => {
+  const schematicRunner = new SchematicTestRunner(
+    'migrations',
+    require.resolve('../migration-collection.json'),
+  );
+
+  let tree: UnitTestTree;
+  let host: TempScopedNodeJsSyncHost;
+
+  const lazyRoutePath = normalize('src/lazy-route.ts');
+  const lazyRoute = virtualFs.stringToFileBuffer(`
+    import { Route } from '@angular/router';
+    const routes: Array<Route> = [
+      {
+        path: '',
+        loadChildren: './lazy/lazy.module#LazyModule'
+      }
+    ];
+  `);
+
+  const lazyChildRoute = virtualFs.stringToFileBuffer(`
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -62,6 +62,7 @@
     ]
   },
   "dependencies": {
+    "@phenomnomnominal/tsquery": "3.0.0",
```
alan-agius4: This should also be added in `packages/schematics/angular/package.json`

packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const updateLazyModulePaths = (): Rule => {
+  return (_: Tree, context: SchematicContext) => {
+    context.addTask(new TslintFixTask({
+      rulesDirectory: path.join(__dirname, 'rules'),
+      rules: {
+        'no-lazy-module-paths': [true],
+      },
+    }, {
+      includes: '*.ts',
```
alan-agius4: Shouldn't this be `**/*.ts`?

packages/schematics/angular/migrations/update-8/update-lazy-module-paths_spec.ts
```diff
@@ -0,0 +1,78 @@
...
+      {
+        path: '',
+        children: [{
+          path: 'child',
+          loadChildren: './lazy/lazy.module#LazyModule'
+        }]
+      }
+    ];
+  `;
+  const packageJson = {
```
alan-agius4: you can remove the below if you use 
```
    beforeEach(async () => {
      tree = new UnitTestTree(new EmptyTree());
      tree = await schematicRunner.runExternalSchematicAsync(
        require.resolve('../../collection.json'), 'ng-new',
        {
          name: 'migration-test',
          version: '1.2.3',
          directory: '.',
        },
        tree,
      ).toPromise();
    });
```

Instead of 
```
   beforeEach(() => {
      tree = new UnitTestTree(new EmptyTree());
      tree.create(packageJsonPath, JSON.stringify(packageJson, null, 2));
    });
```

packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts
phenomnomnominal: Damn, I hoped this was going to get the tests passing 😞

alan-agius4:
I will try to look into the test issue that you are having.

alan-agius4:
I will try to look into the test issue that you are having. 

I am thinking it might be because tslint uses the actual file system. Also I am not entirely use that you can use a TypeScript tslint rule for testing. 

Maybe this test can help you https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/schematics/tasks/tslint-fix/executor_spec.ts

alan-agius4:
@phenomnomnominal. I did a couple of changes locally to try to get this running but it seems that the actual rules for some reasons is throwing an error

```
[ '/tmp/devkit-host-1555415483626-20046/lazy-route.ts' ]
The 'no-lazy-module-paths' rule threw an error in '/tmp/devkit-host-1555415483626-20046/lazy-route.ts':
SyntaxError: Expected " ", "#", "*", "+", ",", ".", ":", ":first-child", ":has(", ":last-child", ":matches(", ":not(", ":nth-child(", ":nth-last-child(", ">", "[", "~" or [^ [\],():#!=><~+.] but ")" found.
```

Would you mind if I push the changes to your branch to at least make the lint rule running in the specs?

phenomnomnominal:
Not at all please do! I think I've spotted the bug in the rule - seems like I have an extra `)` from when I split it to multiple lines to pass lint 😅

phenomnomnominal:
Okay, I think I'm, actually getting closer! I have the rule running in a test I think!

alan-agius4:
@phenomnomnominal, just pushed. 

Note for later, kindly squash the commits :)

packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {
```
alan-agius4: Nit: is this tslint disable required here?

This package has a direct dependency on `@angular-devkit/schematics` which should make this redundant.

filipesilva:
This migration also needs to add `"experimentalImportFactories": true` to `angular.json` to make the import syntax supported in VE.

phenomnomnominal:
@filipesilva sure! Excuse my ignorance, but I guess that means a separate PR? Does the migration also need to modify the `tsconfig.json` file?

filipesilva:
Heya @phenomnomnominal! Yes that would need a separate PR. Now that I think about it, you raise a good point: the tsconfig also needs to use `"module": "esnext"` actually, otherwise the import won't work at all when compiling.

I think what we really need is a full e2e test to ensure all these things are happening on update.

phenomnomnominal:
Great, I will have a go at a PR 👍

filipesilva:
Let me know if you need help with the e2e. It will probably be a very similar to [update-1.7.ts](https://github.com/angular/angular-cli/blob/master/tests/legacy-cli/e2e/tests/basic/update-1.7.ts), that tests updating from [this project](https://github.com/angular/angular-cli/tree/master/tests/legacy-cli/e2e/assets/1.7-project).

The project that you want to update should have a lazy module with the string syntax and an e2e, much like [this](https://github.com/angular/angular-cli/blob/b95933a57d217a82b29fe04fbe333410fe5b8f4f/tests/legacy-cli/e2e/tests/build/lazy-load-syntax.ts).

Then you update the project, check that `import() etc etc` is there, run `ng e2e` and `ng e2e --prod`, and that makes us pretty sure the update went well.

filipesilva:
@phenomnomnominal I had some time today and ended up doing the e2e (https://github.com/angular/angular-cli/pull/14228) and removing the option overall (https://github.com/angular/angular-cli/pull/14231).

phenomnomnominal:
@filipesilva oh thank you so much! I had a go with the e2e test and got stuff running but couldn't figure out how to run it against the local version of the CLI... but either way, thanks!

And removing the experimental flag seems smart since we're forcing people to use it!

PR #14182, opened 2019-04-16
fix(@angular-devkit/architect): architect should use provided registry

alan-agius4:


packages/angular_devkit/architect/src/architect.ts
```diff
@@ -318,16 +318,16 @@ export class Architect {
     // Create private jobs.
     privateArchitectJobRegistry.register(_getTargetOptionsFactory(_host));
     privateArchitectJobRegistry.register(_getBuilderNameForTargetFactory(_host));
-    privateArchitectJobRegistry.register(_validateOptionsFactory(_host, _registry));
+    privateArchitectJobRegistry.register(_validateOptionsFactory(_host, this._registry));
```
clydin: `_registry` is a parameter of the constructor (this code block).  Using `this` is not required since it is also local.

alan-agius4: Umm... TypeScript is warning that registry is unused since `_registry` is marked with a `private` https://github.com/angular/angular-cli/blob/24d2776ec2067c42474f82614d29a64f91903ceb/packages/angular_devkit/architect/src/architect.ts#L314

clydin: In that case, the `private` should be removed since it is not actually used as a class instance member.

packages/angular_devkit/architect/src/architect.ts
```diff
@@ -311,7 +311,7 @@ export class Architect {
 
   constructor(
     private _host: ArchitectHost,
-    private _registry: json.schema.SchemaRegistry = new json.schema.CoreSchemaRegistry(),
+    _registry: json.schema.SchemaRegistry = new json.schema.CoreSchemaRegistry(),
```
clydin: nit: remove underscore as well

PR #14183, opened 2019-04-16
feat(@angular/cli): add deploy command

filipesilva:
Does not do anything on a new project. Will only work after a `deploy` target is added to a project.

We should put up documentation on https://angular.io/guide/workspace-config about this.

### DIFF
--- a/packages/angular/cli/BUILD
+++ b/packages/angular/cli/BUILD
@@ -50,6 +50,7 @@ ts_library(
         ":analytics_schema",
         ":build_schema",
         ":config_schema",
+        ":deploy_schema",
         ":deprecated_schema",
         ":doc_schema",
         ":e2e_schema",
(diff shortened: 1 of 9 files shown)


alexeagle:
wait, what's the design/roadmap for this? In my recent discussions my understanding is that we are shipping with only `ng run` for v8

filipesilva:
We need to flesh out the story for this a bit further before adding it.

alfaproject:
I'm using the [serverless](https://github.com/serverless/serverless) framework and added a custom schematic in my workspace to run any serverless command in any target like so: `ng run backoffice:serverless-deploy`

This was kind of a test to see how feasible it would be to use Angular CLI instead of the native `serverless` CLI but that syntax is not very ergonomic, so I was wondering if there are plans to let people extend Angular CLI commands with their own set, so that I would enable the option of doing:
* `ng run sls-deploy <target>`
* `ng run sls-remove <target>`
* `ng run sls-info <target>`
* etc

PR #14188, opened 2019-04-16
fix(@schematics/angular): generate `tsconfig.worker.json` outside of …

alan-agius4:
…the `src` folder

This is to align with the folder structure of version 8, were tsconfigs are outside of the `src` folder

Also, this change remove the dud `tsconfig.json` in the `src` folder and instead we add the triple slash lib reference `/// <reference lib="webworker" />` for IDE support.


alan-agius4:
//cc @gkalpak

packages/schematics/angular/web-worker/index.ts
```diff
@@ -169,11 +174,12 @@ export default function (options: WebWorkerOptions): Rule {
     const parsedPath = parseName(options.path, options.name);
     options.name = parsedPath.name;
     options.path = parsedPath.path;
-    const root = project.root || project.sourceRoot || '';
+    const root = project.root || '';
```
gkalpak: AFAICT, this raises the following concerns (maybe they are not problems - I am not sure):
1. The [`exclude` path in `tsconfig.json`] should also be updated to include `src/` (`**/*.worker.ts` --> `src/**/*.worker.ts`).
2. By updating the main `tsconfig.json` with `lib: [..., 'webworker']` (i.e. [here][2] or [there][3]), this lib is now applied to all configs which inherit from the main onet (e.g. `tsconfig.app.json`, `tsconfig.spec.json`, `e2e/tsconfig.json`).
    Is this intentional (or should other configs be updated to overwrite `lib` and not include `webworker`)?

[1]: https://github.com/angular/angular-cli/blob/c6315fee4/packages/schematics/angular/web-worker/index.ts#L76
[2]: https://github.com/angular/angular-cli/blob/c6315fee4121ed29c7e25496e692cf2cc66ff27e/packages/schematics/angular/web-worker/index.ts#L60
[3]: https://github.com/angular/angular-cli/blob/c6315fee4121ed29c7e25496e692cf2cc66ff27e/packages/schematics/angular/web-worker/files/project-tsconfig/tsconfig.json.template#L7
[4]: https://github.com/angular/angular-cli/blob/c6315fee4121ed29c7e25496e692cf2cc66ff27e/packages/schematics/angular/web-worker/files/worker-tsconfig/tsconfig.worker.json.template#L7

alan-agius4: While the first point itself is not a problem per see, the second one might be.

Also this might be confusing for users to have multiple tsconfigs.
````
-- tsconfig.json
-- tsconfig.worker.json
-- tsconfig.app.json
-- tsconfig.spec.json
-- src
---- tsconfig.json
```

@filipesilva, what are your toughts about this? I don't think there is much we can be apart from creating the new `tsconfig.json` under `src`, and make a note in it stating that that config is soley used for IDEs

alan-agius4: While the first point itself is not a problem per see, the second one might be.

Also this might be confusing for users to have multiple tsconfigs though I cannot think of an alternative and place the second tsconfig in the src folder. 

Ie:
```
-- tsconfig.json
-- tsconfig.worker.json
-- tsconfig.app.json
-- tsconfig.spec.json
-- src
---- tsconfig.json
```

@filipesilva, what are your thoughts about this? I don't think there is much we can be apart from creating the new `tsconfig.json` under `src`, and make a note in it stating that that config is soley used for IDEs

alan-agius4: Also it seems that this tsconfig https://github.com/alan-agius4/angular-cli/blob/c6315fee4121ed29c7e25496e692cf2cc66ff27e/packages/schematics/angular/web-worker/files/project-tsconfig/tsconfig.json.template has both DOM and webworker. I was under the impression that they are mutually exclusive.

Also this config has no include/exclude pattern which means any ts file within the project will be included.

alan-agius4: Also possibly the worker tsconfig should inherit from the workspace tsconfig not the ide tsconfig.

filipesilva: The general idea is that there is at least one tsconfig that is called just `tsconfig.json` and includes typings for both `dom` and `webworker`. Such configs are meant to be picked up by the IDE to provide accurate typings.

There are also *at least* three different known locations for app tsconfigs in the wild:
- `somefolder/projectname/tsconfig.app.json`, created on multi-project workspaces
- `src/tsconfig.app.json`, created by CLI 1,6,7
- `tsconfig.app.json`, created by CLI 8

Also of note is that in CLI 8 the multi-project workspace story is "don't use your current one, but make a new one with `ng new --create-application` instead".

The new `tsconfig.worker.json` suffers from a similar problem as `tsconfig.spec.json`, in that files for its compilation are intersped with application files. This means we can't rely on a clear separation of TS compilation contexts by folder and must show typings for both in many files.

Compounding on this problem is that `dom` and `webworker` typings are incompatible (https://github.com/Microsoft/TypeScript/issues/20595) for compilation purposes. So any compilation unit that has them both will fail. This doesn't seem to be a problem for IDE support though. 

So, back to the original questions:
1. what base `tsconfig.json` should `tsconfig.worker.json` extend from
2. what IDE `tsconfig.json` should provide the `dom`+`webworker` typings

I think the super correct answer to both 1 and 2 is the root `/tsconfig.json`. But this means altering that one to have both `dom`+`webworker`, and then altering all tsconfigs that inherit from it to only have either `dom` or `webworker`. 

This is non trivial and high risk. We don't know what configs inherit from the base one and missing any of them will cause that compilation to fail. Also when making new projects after this change, the new project would also need to know that the base tsconfig has incompatible typings and perform the same change.

So the less risky approach is for 1 to be the root `/tsconfig.json`, and 2 to be a new `src/tsconfig.json` that we create just for IDE support, and from which nothing inherits. I like this option better because it doesn't break anyone, and is more future proof. Users can just drop the IDE one if the incompatible typings situation changes in the future, or if IDEs start supporting the differently names tsconfigs. But it adds another file and we want to keep the count low.

Another option is for 1 to be the root `/tsconfig.json`, and instead of doing 2 we have some hard cast in the `.worker.ts` file similar to the approach described in https://github.com/Microsoft/TypeScript/issues/20595#issuecomment-351030256. I tried a bunch of these and couldn't get a good one. But given that the other approaches have significant downsides maybe it's worth looking further into.

packages/angular_devkit/build_angular/test/browser/web-worker_spec_large.ts
```diff
@@ -26,6 +26,8 @@ describe('Browser Builder Web Worker support', () => {
   const workerFiles: { [k: string]: string } = {
     'src/app/dep.ts': `export const foo = 'bar';`,
     'src/app/app.worker.ts': `
+      /// <reference lib="webworker" />
```
alan-agius4: Note: this is solely used for IDE purposes.

packages/schematics/angular/web-worker/index.ts
alan-agius4: While for the compilation we do need a seperate tsconfig as otherwise we'll get compilation errors like 
```
node_modules/typescript/lib/lib.webworker.d.ts:4344:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'navigator' must be of type 'Navigator', but here has type 'WorkerNavigator'.

4344 declare var navigator: WorkerNavigator;
```

For IDE support, we can use the triple slash lib reference `/// <reference lib="webworker" />` in the web worker file, though this doesn't doesn't fix the issue were the same symbols are defined in both DOM and WebWorker libs (Which is already present with the dud tsconfig, since the later has both dom and webworker as libs).

Such as an example `self`

Writing something like:
```ts
const _self = self;
```

Would result in `_self` being of type `Window` and not `WorkerGlobalScope`.

This is because of what @filipesilva explained above, and the only way I can think to solve this, is that workers, are not created inside the features folders, but rather in a separate folder similar to `e2e`.

Ie: 
```
-- tsconfig.json
-- tsconfig.app.json
-- tsconfig.spec.json
-- src
---- workers
------ worker-1.ts
------ tsconfig.json
---- feature-foo
------ feature-foo.component.ts
```

packages/schematics/angular/web-worker/index.ts
```diff
@@ -169,11 +133,11 @@ export default function (options: WebWorkerOptions): Rule {
...
 
     const needWebWorkerConfig = !projectTargetOptions.webWorkerTsConfig;
     if (needWebWorkerConfig) {
-      const workerConfigPath = `${root.endsWith('/') ? root : root + '/'}tsconfig.worker.json`;
+      const workerConfigPath = join(normalize(root), 'tsconfig.worker.json');
```
gkalpak: Won't this result in a normalized absolute path? If so, is that intended?
(Or is `project.root` relative?)

alan-agius4: Project root is relative 😊

PR #14191, opened 2019-04-16
fix(@schematics/angular): add `tsconfig.worker.json` to list of tscon…

alan-agius4:
…figs to lint

//cc @gkalpak

### DIFF
--- a/packages/schematics/angular/web-worker/index.ts
+++ b/packages/schematics/angular/web-worker/index.ts
@@ -13,7 +13,7 @@ import {
 import { appendValueInAstArray, findPropertyInAstObject } from '../utility/json-utils';
 import { parseName } from '../utility/parse-name';
 import { buildDefaultPath, getWorkspace, updateWorkspace } from '../utility/workspace';
-import { BrowserBuilderOptions } from '../utility/workspace-models';
+import { BrowserBuilderOptions, LintBuilderOptions } from '../utility/workspace-models';
 import { Schema as WebWorkerOptions } from './schema';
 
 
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/web-worker/index.ts
```diff
@@ -173,8 +173,15 @@ export default function (options: WebWorkerOptions): Rule {
 
     const needWebWorkerConfig = !projectTargetOptions.webWorkerTsConfig;
     if (needWebWorkerConfig) {
-      projectTargetOptions.webWorkerTsConfig =
-        `${root.endsWith('/') ? root : root + '/'}tsconfig.worker.json`;
+      const workerConfigPath = `${root.endsWith('/') ? root : root + '/'}tsconfig.worker.json`;
```
alexeagle: are these paths always posix? how do we know to use forward slash?

filipesilva: The convention is that they are always posix and relative to the `angular.json`. But we don't enforce it or error out if they are not posix. We do interpret them through the path abstraction in the devkit to convert them to system paths.

alan-agius4: I think after this PR, I’ll do another PR to refactor some logic in these schematics as there is some repetition among various schematics to generate a relative path.

alan-agius4: Most of our schematics rely on the fact that paths are posix.

Probably we can do some refactoring to support this. But i’d prefer if it was done in a separate PR.

Also, I’d like to extract some common methods which are repeated into various schematics so that they can be re-used.

PR #14202, opened 2019-04-17
docs(@schematics/angular): fix typo in web worker schematics ("service worker" --> "web worker")

gkalpak:
☝️

### DIFF
--- a/packages/schematics/angular/web-worker/schema.json
+++ b/packages/schematics/angular/web-worker/schema.json
@@ -3,14 +3,14 @@
   "id": "SchematicsAngularWebWorker",
   "title": "Angular Web Worker Options Schema",
   "type": "object",
-  "description": "Pass this schematic to the \"run\" command to create a Web Worker",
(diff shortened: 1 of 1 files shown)


gkalpak:
BTW, is this correct:

https://github.com/angular/angular-cli/blob/ab3dcdd04fa23b969bc0cef886cd17283bd5ce9f/packages/schematics/angular/web-worker/schema.json#L6

It feels wrong (at least I don't understand what it means), but I am not sure what would be the correct description.
Maybe replace `run` with `generate`?

alan-agius4:
@gkalpak, I don’t think that is correct it should be something along `Creates a new generic web worker definition in the given or default project."`

alan-agius4:
Feel free to fix that in this PR or a separate one 😊

gkalpak:
Fixed.

alan-agius4:
@gkalpak, can you please rebase to make this green? thanks

Issue #14190, opened 2019-04-16
Telemetry: don't track implicitly run schematics


StephenFluin:
* We should only track user entered commands (like schematics), not implied schematics. IE when a user runs `ng new` we should track that, but not the fact that they implicitly invoked `ng generate workspace` or `ng generate project`
* Aliases should be combined (ng generate c and ng generate component should be combined into a single type of GA telemetry report)
* Several of our custom metrics (at least Bundle Sizes / build times / ngOnInit) aren't being reported / I can't figure out how to see them :)

PR #14205, opened 2019-04-17
build: Resolve ng generate aliases before logging analytics

alexeagle:


alexeagle:
partial fix for #14190

PR #14206, opened 2019-04-18
feat(@angular-devkit/build-angular): warn when target greater es2015

mgechev:
When using differential loading with targets ES5 and ES2016+, the
browser with ESM support will pick `script[type="module"]` scripts even
without supporting ES2016+ syntax.

We want to warn users in this case.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/index.ts
+++ b/packages/angular_devkit/build_angular/src/browser/index.ts
@@ -18,13 +18,15 @@ import {
   logging,
   normalize,
   resolve,
+  tags,
   virtualFs,
 } from '@angular-devkit/core';
 import { NodeJsSyncHost } from '@angular-devkit/core/node';
 import * as fs from 'fs';
 import * as path from 'path';
-import { concat, from, of, zip } from 'rxjs';
+import { from, of } from 'rxjs';
 import { bufferCount, catchError, concatMap, map, mergeScan, switchMap } from 'rxjs/operators';
+import * as ts from 'typescript';
 import * as webpack from 'webpack';
 import { NgBuildAnalyticsPlugin } from '../../plugins/webpack/analytics';
 import { WebpackConfigOptions } from '../angular-cli-files/models/build-options';
@@ -38,6 +40,7 @@ import {
   getWorkerConfig,
 } from '../angular-cli-files/models/webpack-configs';
 import { writeIndexHtml } from '../angular-cli-files/utilities/index-file/write-index-html';
+import { readTsconfig } from '../angular-cli-files/utilities/read-tsconfig';
 import { augmentAppWithServiceWorker } from '../angular-cli-files/utilities/service-worker';
 import {
   statsErrorsToString,
@@ -45,7 +48,7 @@ import {
   statsWarningsToString,
 } from '../angular-cli-files/utilities/stats';
 import { ExecutionTransformer } from '../transforms';
-import { deleteOutputDir } from '../utils';
+import { deleteOutputDir, isEs5SupportNeeded } from '../utils';
 import { generateBrowserWebpackConfigFromContext } from '../utils/webpack-browser-config';
 import { Schema as BrowserBuilderSchema } from './schema';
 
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -303,6 +303,16 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
     `);
   }
 
+  if (isEs5SupportNeeded(projectRoot) &&
```
alan-agius4: This warning will trigger everywhere not only in `build`, but also `serve`, `test` etc..

Also I don't think this warning should be tight with differential loading because a user can opt not to have ES5 browser support and still encounter this issue.

For example when targeting `ES2017`, and non of the browsers in browserlist require ES5 but still they don't support ES2016+ features.

Note: we already display a message for when users target es2017 maybe we should improve that message?
https://github.com/angular/angular-cli/blob/5fc1f2499cbe57f9a95e4b0dfced130eb3a8046d/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts#L298

alan-agius4: We should improve the below check and use the enums
```ts
 if (wco.tsConfig.options.target === 4) { 
```

clydin:
This is good for 8.0.  But I think that long-term, it should leverage the Browserslist to only show a warning if the required browsers do not support the ES version.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
mgechev: >Also I don't think this warning should be tight with differential loading because a user can opt not to have ES5 browser support and still encounter this issue.

Not sure if this is a concern in such scenario. IMO, it's dangerous to think that if you have ES5 and ES2016 you cover the entire interval of ES5-ES2016 browsers continuously. If you have only ES2016 bundle, it's more obvious that you support only ES2016+ browsers.

mgechev:
@clydin currently the check verifies if there are browsers which require ES5 support and warns only in such case.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
mgechev: @alan-agius4 https://github.com/angular/angular-cli/pull/14213

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -38,14 +40,15 @@ import {
   getWorkerConfig,
 } from '../angular-cli-files/models/webpack-configs';
 import { writeIndexHtml } from '../angular-cli-files/utilities/index-file/write-index-html';
+import { readTsconfig } from '../angular-cli-files/utilities/read-tsconfig';
```
alan-agius4: Redundant import

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -18,13 +18,15 @@ import {
...
   resolve,
+  tags,
   virtualFs,
 } from '@angular-devkit/core';
 import { NodeJsSyncHost } from '@angular-devkit/core/node';
 import * as fs from 'fs';
 import * as path from 'path';
 import { from, of, zip } from 'rxjs';
 import { catchError, concatMap, map, switchMap } from 'rxjs/operators';
+import { ScriptTarget } from 'typescript';
```
alan-agius4: Typically for TypeScript we use lazy import

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -182,8 +185,20 @@ export function buildWebpackBrowser(
         normalize(workspace.getProject(projectName).root),
       );
 
+      const tsConfig = readTsconfig(projectRoot);
```
alan-agius4: The tsconfig path should come from the browserOptions.

PR #14210, opened 2019-04-18
fix(@angular-devkit/build-angular): run build steps for differential …

manfredsteyer:
…loading in sequence to avoid confusing progress information

Before, the build tasks run in parallel and so the different webpack
instanced competed over the same lines on the console.

This causes an issue with watch mode. We'll address this with an further
PR.

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
@@ -63,6 +63,36 @@ describe('Browser Builder with differential loading', () => {
       .toEqual(jasmine.arrayWithExactContents(expectedOutputs));
   });
 
+  it('deactivates differential loading for watch mode', async () => {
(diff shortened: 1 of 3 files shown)


alexeagle:
could we consider disabling diff loading in watch mode in the same PR, so we don't have to be careful to avoid releasing a broken RC if this lands and that fix doesn't?

packages/angular_devkit/build_angular/test/browser/differential_loading_spec_large.ts
```diff
@@ -63,6 +63,36 @@ describe('Browser Builder with differential loading', () => {
       .toEqual(jasmine.arrayWithExactContents(expectedOutputs));
   });
 
+  it('deactivates differential loading for watch mode', async () => {
+    const { files } = await browserBuild(architect, host, target, { watch: true });
+
+    const expectedOutputs = [
```
alan-agius4: Small nit: we should emit es2015 builds and not es5 to be consistent with server, e2e, test and all the other builders

PR #14212, opened 2019-04-18
feat(@schematics/angular): enable support for Chrome 41

mgechev:
We want apps to be SEO friendly by default. // cc @alexeagle


mohammedzamakhan:
@mgechev Chrome 41 support can be removed from the browserlist now, as googlebot started using Evergreen chrome https://twitter.com/mohamedzamakhan/status/1125997359459176448

clydin:
PR is already pending and approved: #14370

mohammedzamakhan:
@clydin Thanks for the update 👍

PR #14231, opened 2019-04-22
Remove experimentalImportFactories

### LINKED ISSUE #14218: Add experimentalImportFactories when migrating to loadChildren import syntax
The https://github.com/angular/angular-cli/pull/14181 migration also needs to add `"experimentalImportFactories": true` to `angular.json` (https://github.com/angular/angular-cli/pull/13948) to make the import syntax supported in VE. 

The tsconfig file should also be updated to use `"module": "esnext"`, and a e2e test should be added to test the update works and an app with lazy routes will be updated and runs properly.

filipesilva:
It will just be always be used when applicable.

Fix #14218

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -886,11 +886,6 @@
             "webWorkerTsConfig": {
               "type": "string",
               "description": "TypeScript configuration for Web Worker modules."
-            },
-            "experimentalImportFactories": {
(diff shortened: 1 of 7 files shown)


alexeagle:
Discussed with Igor - we agree with removing the flag for RC0 and always enable. Then if we decide that fix-lazy-routes can't stay in RC we'll have to re-introduce this.

PR #14241, opened 2019-04-23
fix(@schematics/angular): move and update existing `src/browserslist`

### LINKED ISSUE #14232: Existing `src/browserlist` should be updated or moved
Projects created using CLI 7 have `src/browserslist`, but the migration expects it to be at root instead.

We should either update `src/browserlist` in place, or move it to root.

alan-agius4:
Older projects have their `browserslist` file under the `src` folder.

Fixes #14232

### DIFF
--- a/packages/schematics/angular/migrations/update-8/differential-loading_spec.ts
+++ b/packages/schematics/angular/migrations/update-8/differential-loading_spec.ts
@@ -74,5 +74,15 @@ describe('Migration to version 8', () => {
       expect(tree2.readContent('/browserslist'))
         .toContain('Googlebot support');
     });
+
+    it('should move browserslist file if it exists in the sourceRoot', () => {
+      tree.create('/src/browserslist', 'last 2 Chrome versions');
+      tree.delete('/browserslist');
+      const tree2 = schematicRunner.runSchematic('migration-07', {}, tree.branch());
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/update-8/differential-loading_spec.ts
```diff
@@ -74,5 +74,15 @@ describe('Migration to version 8', () => {
       expect(tree2.readContent('/browserslist'))
         .toContain('Googlebot support');
     });
+
+    it('should move browserslist file if it exists in the sourceRoot', () => {
+      tree.create('/src/browserslist', 'last 2 Chrome versions');
+      tree.delete('/browserslist');
+      const tree2 = schematicRunner.runSchematic('migration-07', {}, tree.branch());
```
alexeagle: I am confused why these all say `migration-07`

alan-agius4: It's the migration name. For example for Angular CLI version 8, we have multiple migrations.

alan-agius4: https://github.com/alan-agius4/angular-cli/blob/42cfb32469dc51e651e6e07763878bf935a92cf3/packages/schematics/angular/migrations/migration-collection.json#L33-L41

PR #14244, opened 2019-04-23
Fix e2e snapshots

alan-agius4:


alan-agius4:
Successful build https://circleci.com/workflow-run/9cf48141-cd62-4d12-a1dd-77a358465f7e

PR #14249, opened 2019-04-23
update rxjs to version 6.5.1

alan-agius4:


Brocco:
Thank you for this!
I am going to rebase #14091 and use fromFetch to make the remote calls, thus eliminating the need for node-fetch (and corresponding types)

alexeagle:
I reverted this because @IgorMinar says that we don't want rxjs 6.5 until it's released in g3 (which is still on 6.3)
We can land this again as soon as the 8.0.x branch is cut and master targets 8.1

filipesilva:
FYI updated user projects will still use RxJS 6.5.1. The update will use the latest valid peerdep and that's valid.

PR #14254, opened 2019-04-23
fix(@schematics/angular): avoid double header/comments in es6 polyfill migration

clydin:
If a polyfills file is original, then replace completely with latest version of the file.  This removes the need to deeply analyze the file as well as ensuring all comments are up to date.  Also skips analysis if there is no mention of `core-js` of any form in the file.
The header re-addition check is also reduced in scope to account for varying versions of the header content.
The WeakMap polyfill was also removed as it is automatically included at build time.

### DIFF
--- a/packages/schematics/angular/migrations/update-8/drop-es6-polyfills.ts
+++ b/packages/schematics/angular/migrations/update-8/drop-es6-polyfills.ts
@@ -5,9 +5,19 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-
-import { JsonParseMode, isJsonObject, join, normalize, parseJson } from '@angular-devkit/core';
-import { Rule, Tree, chain, noop } from '@angular-devkit/schematics';
+import { JsonParseMode, isJsonObject, parseJson } from '@angular-devkit/core';
+import {
+  MergeStrategy,
+  Rule,
+  Tree,
+  apply,
+  chain,
+  filter,
+  mergeWith,
+  move,
+  url,
+} from '@angular-devkit/schematics';
+import { createHash } from 'crypto';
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/update-8/drop-es6-polyfills.ts
```diff
@@ -59,33 +78,64 @@ function dropES2015PolyfillsFromFile(polyfillPath: string): Rule {
       return;
     }
 
-    // Start the update of the file.
-    const recorder = tree.beginUpdate(polyfillPath);
+    // normalize line endings to increase hash match chances
+    const content = source.toString().replace(/\r\n|\r/g, '\n');
```
filipesilva: 👍

packages/schematics/angular/migrations/update-8/drop-es6-polyfills.ts
```diff
@@ -45,12 +56,20 @@ const header = `/**
...
+const browserPolyfillsHeader = 'BROWSER POLYFILLS';
+
+const knownPolyfillHashes = [
+  '3dba718d7afe009e112e10d69073d2a2', // 6.0 - unmodified
+  'fccdb76b06ea636933f8b99b1c8d9725', // 6.0 - all core-js uncommented
+  '97e16639be1de06153695f5fefde745d', // 7.0 - unmodified
+  'd6c13d6dcf94ff3749283f33dd0d4864', // 7.0 - all core-js uncommented
+  '79bf0fd46c215e5f4145e15641c325f3', // 7.2 - unmodified
+  '6fe8080c7e38ee0ce677fdbc3884377a', // 7.2 - all core-js uncommented
+  '8e7f6abb3d2dca03b4dbb300e400a880', // 7.3 - unmodified
```
alexeagle: will we have to maintain this in the future? please add something to the right doc so we remember to keep it updated

Issue #14011, opened 2019-03-28
Analytics prompt even after disabling


cexbrayat:
# 🐞 Bug report

### Command (mark with an `x`)
```
- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

No

### Description

After installing the latest CLI beta ( `8.0.0-beta.10`) , you get the prompt about analytics.
If answering no, the assumption would be to not get asked when creating a new project later.
At least not when running `ng new ... --defaults` or `ng new ... --no-interactive`.

## 🔬 Minimal Reproduction
```
ng analytics off
ng new cli-analytics --default --no-interactive
# still get the analytics prompt
```

duluca:
I'd like to second this. The prompt also causes an issue when trying to run `npm ci` on a CI server, it halts the installation process with a prompt.

cyrilletuzi:
I think it's the same issue, but Circle CI is still broken with Angular CLI beta.16, due to the analytics prompt (which should not happen in non-interactive mode). Install is done with a classic `npm install`.

gnomeontherun:
This breaks any CI install by asking for analytics permissions on CI and blocking completion of npm install. It is before the CLI is even ready to run locally. I believe the prompt should not run on NPM install, but rather on first use of an ng command. with the exception of `ng analytics` so a CI could disable it.

hippee-lee:
As a travis-ci workaround I added the environment variable: `NG_CLI_ANALYTICS=ci`
as described here: https://github.com/angular/angular-cli/blob/f8fe4b0ee82d880396119a2594f3cee8dadb2947/docs/design/analytics.md#ci

One other note to the cli team: in the cli output with the analytics prompt, the link for more info provided is https://angular.io/analytics but that 404's.
Right now, prolly b/c its still a beta, it is on: https://next.angular.io/analytics

clydin:
Original issue was resolved in the referenced PR.  CI related issues with project level analytics options are being tracked in #14265.

PR #14256, opened 2019-04-24
fix(@angular/cli): only show postinstall prompt when global analytics not configured

clydin:
Partially addresses #14011


alan-agius4:
Can you reference https://github.com/angular/angular-cli/issues/14011?

PR #14260, opened 2019-04-24
test: disable rebuild replacements when running ivy e2e

alan-agius4:
Rebuilds under ivy are broken at the moment

See: https://github.com/angular/angular/issues/30079


alan-agius4:
Obsolete as framework released a fix.

PR #14264, opened 2019-04-24
fix(@angular/cli): check/prompt for workspace level analytics configuration

clydin:


alan-agius4:
@clydin can you please rebase?

packages/angular/cli/models/analytics.ts
```diff
@@ -520,6 +521,66 @@ export function getGlobalAnalytics(): UniversalAnalytics | undefined {
   }
 }
 
+export function hasWorkspaceAnalyticsConfiguration(): boolean {
+  try {
```
alan-agius4: Nit: since the retrieval of config is used in multiple places. Maybe we can extract it into a common function to ease maintenance.

Ex: 

```ts
export type AnalyticsConfiguration =
  | string
  | undefined
  | null
  | { uid?: string };

function getWorkspaceAnalyticsConfiguration(): AnalyticsConfiguration {
  analyticsDebug("getWorkspaceAnalyticsConfiguration");
  try {
    const globalWorkspace = getWorkspace("local");
    const analyticsConfig: AnalyticsConfiguration =
      globalWorkspace &&
      globalWorkspace.getCli() &&
      globalWorkspace.getCli()["analytics"];

    if (analyticsConfig !== undefined) {
      return true;
    }
  } catch (err) {
    analyticsDebug(
      "Error happened during reading of analytics config: %s",
      err.message
    );
  }

  return undefined;
}

export function hasWorkspaceAnalyticsConfiguration(): boolean {
  return getWorkspaceAnalyticsConfiguration() !== undefined;
}

/**
 * Get the workspace analytics object for the user. This returns an instance of UniversalAnalytics,
 * or undefined if analytics are disabled.
 *
 * If any problem happens, it is considered the user has been opting out of analytics.
 */
export function getWorkspaceAnalytics(): UniversalAnalytics | undefined {
  analyticsDebug("getWorkspaceAnalytics");

  const analyticsConfig = getWorkspaceAnalyticsConfiguration();
  if (analyticsConfig === false) {
    analyticsDebug("Analytics disabled. Ignoring all analytics.");

    return undefined;
  } else if (analyticsConfig === undefined || analyticsConfig === null) {
    analyticsDebug("Analytics settings not found. Ignoring all analytics.");

    return undefined;
  } else {
    let uid: string | undefined = undefined;
    if (typeof analyticsConfig == "string") {
      uid = analyticsConfig;
    } else if (
      typeof analyticsConfig == "object" &&
      typeof analyticsConfig["uid"] == "string"
    ) {
      uid = analyticsConfig["uid"];
    }

    analyticsDebug("client id: %j", uid);
    if (uid == undefined) {
      return undefined;
    }

    return new UniversalAnalytics(angularTrackingId, uid);
  }
}

```

PR #14276, opened 2019-04-26
fix(@schematics/angular): normalize differing TS AST versions in lazy module migration

clydin:
The version of TypeScript used by TSLint and the vendored version may not match.  This is especially likely during an update.  Since the TypeScript `SyntaxKind` enum numeric values change between versions, a version mismatch can cause invalid analysis of the AST.  This change normalizes the AST to the vendored version.

Since the update tests pull versions from the npm registry, CI will fail until this change has been published.  This fix was tested manually by installing the locally built packages in a 7.0.7 project and running the migrations.

### DIFF
--- a/packages/schematics/angular/migrations/update-8/rules/noLazyModulePathsRule.ts
+++ b/packages/schematics/angular/migrations/update-8/rules/noLazyModulePathsRule.ts
@@ -21,7 +21,16 @@ export class Rule extends Rules.AbstractRule {
     const ruleName = this.ruleName;
     const changes: RuleFailure[] = [];
 
-    ts.forEachChild(ast, function analyze(node) {
+    // NOTE: This should ideally be excluded at a higher level to avoid parsing
+    if (ast.isDeclarationFile || /[\\\/]node_modules[\\\/]/.test(ast.fileName)) {
+      return [];
+    }
+
+    // Workaround mismatched tslint TS version and vendored TS version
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/migrations/update-8/rules/noLazyModulePathsRule.ts
```diff
@@ -21,7 +21,16 @@ export class Rule extends Rules.AbstractRule {
     const ruleName = this.ruleName;
     const changes: RuleFailure[] = [];
 
-    ts.forEachChild(ast, function analyze(node) {
+    // NOTE: This should ideally be excluded at a higher level to avoid parsing
```
filipesilva: What's stopping this from happening? Parsing all of node modules initially sounds pretty bad.

clydin: Feature additions/fixes would be needed for the tslint fix task used here: https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts#L20

filipesilva: An `exclude` option in https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/schematics/tasks/tslint-fix/options.ts#L12 would be good I suppose. We're already using the `include` but that'll match `node_modules` too.

PR #14287, opened 2019-04-26
Revert "fix(@angular-devkit/build-angular): remove pure_getters"

kara:
This reverts commit 9a9939fc4eb92f50241bc1e46b6623505a587bc5.

We are temporarily reverting the pure_getters removal because it caused a 2x size
regression in Ivy (non-compatibility mode). We need time to investigate why this
is happening and re-assess how we want to move forward.

See https://github.com/angular/angular/commit/fcacb2a4a2e1ad2197b97d2a194d544556f2dcc5 for size regression.


filipesilva:
Related to https://github.com/angular/angular-cli/pull/14287

PR #14303, opened 2019-04-29
remove unused ascii-progress library, resolve toplevel browserslist to 4.5.5 

filipesilva:
build: resolve toplevel browserslist to 4.5.5

    Prevents `BrowserslistError: Unknown browser kaios` error on test-large due to browserslist version skew between hoisted packages.

---

    build: remove unused ascii-progress library

    Ran into https://github.com/bubkoo/ascii-progress/issues/8 on Windows, but we don't even use this dependency anyway.

### DIFF
--- a/packages/angular_devkit/architect_cli/package.json
+++ b/packages/angular_devkit/architect_cli/package.json
@@ -16,7 +16,6 @@
     "@angular-devkit/architect": "0.0.0",
     "@angular-devkit/core": "0.0.0",
     "@types/progress": "^2.0.3",
-    "ascii-progress": "^1.0.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -143,5 +143,8 @@
     "hooks": {
       "pre-push": "node ./bin/devkit-admin hooks/pre-push"
     }
+  },
+  "resolutions": {
+    "browserslist": "4.5.5"
```
clydin: Now that this has _fixed_ yarn's lock file with the newer version, this is not actually needed.

filipesilva: Removed

PR #14312, opened 2019-04-30
refactor(@angular/cli): avoid aggressive eager command loading

clydin:
Currently, upon execution `ng` will load all description files and code for all available commands.  This requires a large amount of unnecessary file access and processing since only at most one command will be executed.  This change limits the loading to only the command being executed in the common case and a subset of commands in the event an alias is used.  The help command now loads all commands during its execution which is needed to gather command description information.  Further improvements are possible by only loading the necessary metadata instead of the execution code (and its dependencies) as well.
This change allows for savings of ~250ms per execution and reduces memory consumption by 10-20 MBs depending on the command.

Examples:
Before -- `./node_modules/.bin/ng version  0.99s user 0.17s system 113% cpu 1.020 total`
After -- `./node_modules/.bin/ng version  0.70s user 0.13s system 110% cpu 0.749 total`

Before -- `./node_modules/.bin/ng g c a  1.91s user 0.30s system 111% cpu 1.996 total`
After -- `./node_modules/.bin/ng g c a  1.62s user 0.27s system 110% cpu 1.715 total`

### DIFF
--- a/packages/angular/cli/models/command-runner.ts
+++ b/packages/angular/cli/models/command-runner.ts
@@ -17,25 +17,41 @@ import {
 } from '@angular-devkit/core';
 import * as debug from 'debug';
 import { readFileSync } from 'fs';
-import { dirname, join, resolve } from 'path';
-import { findUp } from '../utilities/find-up';
+import { join, resolve } from 'path';
 import { parseJsonSchemaToCommandDescription } from '../utilities/json-schema';
 import { UniversalAnalytics, getGlobalAnalytics, getSharedAnalytics } from './analytics';
 import { Command } from './command';
-import {
-  CommandDescription,
-  CommandDescriptionMap,
-  CommandWorkspace,
-} from './interface';
+import { CommandDescription, CommandWorkspace } from './interface';
 import * as parser from './parser';
 
 const analyticsDebug = debug('ng:analytics:commands');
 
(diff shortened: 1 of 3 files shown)


packages/angular/cli/models/command-runner.ts
```diff
@@ -17,26 +17,41 @@ import {
...
-  CommandDescription,
-  CommandDescriptionMap,
-  CommandWorkspace,
-} from './interface';
+import { CommandDescription, CommandWorkspace } from './interface';
 import * as parser from './parser';
 
 const analyticsDebug = debug('ng:analytics:commands');
 
+const standardCommands = {
```
alan-agius4: You not simply read this from the `commands.json`, rather than having it duplicated here.

clydin: The data is constant so there’s really no reason to find, read, and parse a json file on every execution.

clydin: The file itself can not be removed yet as it is unfortunately used by one of the validation tests due to the top level help command not provided a json help option (unlike the command specific help).  Once this is remedied, the file can be removed as well.

filipesilva: Having this duplicated means at some point we'll change one and forget to change the other. If we can't remove it now then we should have a warning here.

clydin: Done.

PR #14314, opened 2019-04-30
refactor(@angular-devkit/architect): support stable workspaces API

clydin:


packages/angular_devkit/architect/node/node-modules-architect-host.ts
```diff
@@ -109,4 +122,17 @@ export class WorkspaceNodeModulesArchitectHost implements ArchitectHost<NodeModu
 
     throw new Error('Builder is not a builder');
   }
+
+  private findProjectTarget(target: Target) {
+    if ('getProjectTargets' in this._workspace) {
```
alexeagle: could you add a comment here? I'm guessing this conditional is for backwards compat with the old workspaces API?

PR #14323, opened 2019-05-01
fix(@schematics/angular): avoid tslint overhead for lazy module migration

clydin:
By using a pure schematic based implementation the overhead of running tslint can be removed.  This also has the benefit of allowing direct control over the files loaded and to modify files only within the schematic tree context.
For a hello world application, there was a ~20% performance improvement for the CLI migration from 7.0 to 8.0 as well as a ~10% reduction in memory usage.


packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts
```diff
@@ -5,26 +5,76 @@
...
+          const valueNode = node.initializer;
+          const parts = valueNode.text.split('#');
+          const path = parts[0];
+          const moduleName = parts[1] || 'default';
+
+          let fix = `() => import('${path}').then(m => m.${moduleName})`;
+
+          // Try to fix indentation in replacement:
+          const { character } = sourceFile.getLineAndCharacterOfPosition(node.getStart());
+          fix = fix.replace(/\n/g, `\n${' '.repeat(character)}`);
```
filipesilva: `fix` has no `\n`, I don't think this'll ever replace anything.

packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts
```diff
@@ -5,26 +5,76 @@
...
-        'no-lazy-module-paths': [true],
-      },
-    }, {
-      includes: '**/*.ts',
-      silent: false,
-    }));
+import { DirEntry, Rule, UpdateRecorder } from '@angular-devkit/schematics';
+import * as ts from '../../third_party/github.com/Microsoft/TypeScript/lib/typescript';
+
+function* visit(directory: DirEntry): IterableIterator<ts.SourceFile> {
```
filipesilva: Really nice use of a generator 👍 

This should reduce TS source files in memory to a constant 1.

packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts
clydin: Removed.

PR #14324, opened 2019-05-01
fix(@schematics/angular): reduce package installs for 8.0 migrations

clydin:
At most one package manager install will now be executed for 8.0 migrations.  This provides a ~30% performance improvement for a hello world project.


packages/schematics/angular/migrations/update-8/update-builders.ts
```diff
@@ -55,9 +49,5 @@ export function updateBuilders() {
         },
       );
     }
-
-    if (updates) {
-      context.addTask(new NodePackageInstallTask());
```
filipesilva: Doesn't this removal mean that this migration install the added packages?

PR #14329, opened 2019-05-02
fix(@schematics/angular): insertAfterLastOccurrence AST util

hawkgs:
The introduced change fixes a bug in the `insertAfterLastOccurrence` AST utility function which is part of the `@schematics/angular` package. By design, the function should be able to insert a string in an empty node list, hence the `fallbackPos` argument. However, the check performed in the function implementation itself blocks the user to do so; an error is thrown instead.

### DIFF
--- a/packages/schematics/angular/utility/ast-utils.ts
+++ b/packages/schematics/angular/utility/ast-utils.ts
@@ -191,10 +191,7 @@ export function insertAfterLastOccurrence(nodes: ts.Node[],
                                           syntaxKind?: ts.SyntaxKind): Change {
   // sort() has a side effect, so make a copy so that we won't overwrite the parent's object.
   let lastItem = [...nodes].sort(nodesByPosition).pop();
-  if (!lastItem) {
-    throw new Error();
-  }
-  if (syntaxKind) {
+  if (syntaxKind && lastItem) {
     lastItem = findNodes(lastItem, syntaxKind).sort(nodesByPosition).pop();
   }
   if (!lastItem && fallbackPos == undefined) {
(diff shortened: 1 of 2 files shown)


mgechev:
I reran the build-bazel workflow. Looks like a flake.

packages/schematics/angular/utility/ast-utils.ts
```diff
@@ -191,10 +191,7 @@ export function insertAfterLastOccurrence(nodes: ts.Node[],
                                           syntaxKind?: ts.SyntaxKind): Change {
   // sort() has a side effect, so make a copy so that we won't overwrite the parent's object.
   let lastItem = [...nodes].sort(nodesByPosition).pop();
-  if (!lastItem) {
-    throw new Error();
-  }
-  if (syntaxKind) {
+  if (syntaxKind && lastItem) {
```
clydin: This still appears quite inefficient.  What about something like the following:
```
  let lastItem: ts.Node | undefined;
  for (const node of nodes) {
    if (!lastItem || lastItem.getStart() < node.getStart()) {
      lastItem = node;
    }
  }
  if (syntaxKind && lastItem) {
```
Efficiently handles the length 0 & 1 cases and avoids the extra array creation, sort mutations, and pop mutation.

hawkgs: Maybe we should add this in a separate PR since the change emphasizes on/covers the insertion bug rather than improving the efficiency? What do you think?

Anyway, I agree with you that what you proposed is better performance-wise.

clydin: Ok

PR #14345, opened 2019-05-05
feat(@schematics/angular): route generator

hawkgs:
Introduces the lazy routing module generator as part of the existing module generator.

**Feature document:** [Route generator in Angular CLI](https://docs.google.com/document/d/1NAJEWTG54cu4kW7XBp0gVadmz-gqJouW2XDs4c9PPJM/edit?usp=sharing)

### DIFF
--- a/packages/schematics/angular/module/files/__name@dasherize@if-flat__/__name@dasherize__.module.ts.template
+++ b/packages/schematics/angular/module/files/__name@dasherize@if-flat__/__name@dasherize__.module.ts.template
@@ -1,13 +1,19 @@
 import { NgModule } from '@angular/core';<% if (commonModule) { %>
-import { CommonModule } from '@angular/common';<% } %><% if (routing) { %>
-
+import { CommonModule } from '@angular/common';<% } %><% if (lazyRouteWithoutRouteModule) { %>
+import { Routes, RouterModule } from '@angular/router';<% } %>
+<% if (routing || lazyRouteWithRouteModule) { %>
 import { <%= classify(name) %>RoutingModule } from './<%= dasherize(name) %>-routing.module';<% } %>
+<% if (lazyRouteWithoutRouteModule) { %>
+const routes: Routes = [
+  { path: '', component: <%= classify(name) %>Component }
+];<% } %>
 
 @NgModule({
   declarations: [],
   imports: [<% if (commonModule) { %>
-    CommonModule<%= routing ? ',' : '' %><% } %><% if (routing) { %>
-    <%= classify(name) %>RoutingModule<% } %>
+    CommonModule<%= routing || lazyRouteWithRouteModule ? ',' : '' %><% } %><% if (routing || lazyRouteWithRouteModule) { %>
+    <%= classify(name) %>RoutingModule<% } %><%= lazyRouteWithoutRouteModule ? ',' : '' %><% if (lazyRouteWithoutRouteModule) { %>
+    RouterModule.forChild(routes)<% } %>
   ]
 })
 export class <%= classify(name) %>Module { }
--- a/packages/schematics/angular/module/index.ts
+++ b/packages/schematics/angular/module/index.ts
@@ -5,7 +5,7 @@
 * Use of this source code is governed by an MIT-style license that can be
 * found in the LICENSE file at https://angular.io/license
 */
-import { normalize, strings } from '@angular-devkit/core';
+import { Path, normalize, strings } from '@angular-devkit/core';
 import {
   Rule,
   SchematicsException,
@@ -17,17 +17,28 @@ import {
   mergeWith,
   move,
   noop,
+  schematic,
   url,
 } from '@angular-devkit/schematics';
 import * as ts from '../third_party/github.com/Microsoft/TypeScript/lib/typescript';
-import { addImportToModule } from '../utility/ast-utils';
+import { addImportToModule, addRouteDeclarationToModule } from '../utility/ast-utils';
 import { InsertChange } from '../utility/change';
 import { buildRelativePath, findModuleFromOptions } from '../utility/find-module';
 import { applyLintFix } from '../utility/lint-fix';
 import { parseName } from '../utility/parse-name';
 import { createDefaultPath } from '../utility/workspace';
-import { Schema as ModuleOptions } from './schema';
+import { RoutingScope, Schema as ModuleOptions } from './schema';
 
+function buildRelativeModulePath(options: ModuleOptions, modulePath: string): string {
+  const importModulePath = normalize(
+    `/${options.path}/`
+    + (options.flat ? '' : strings.dasherize(options.name) + '/')
+    + strings.dasherize(options.name)
+    + '.module',
+  );
+
+  return buildRelativePath(modulePath, importModulePath);
+}
 
 function addDeclarationToNgModule(options: ModuleOptions): Rule {
   return (host: Tree) => {
@@ -41,16 +52,10 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
     if (text === null) {
       throw new SchematicsException(`File ${modulePath} does not exist.`);
     }
-    const sourceText = text.toString('utf-8');
+    const sourceText = text.toString();
     const source = ts.createSourceFile(modulePath, sourceText, ts.ScriptTarget.Latest, true);
 
-    const importModulePath = normalize(
-      `/${options.path}/`
-      + (options.flat ? '' : strings.dasherize(options.name) + '/')
-      + strings.dasherize(options.name)
-      + '.module',
-    );
-    const relativePath = buildRelativePath(modulePath, importModulePath);
+    const relativePath = buildRelativeModulePath(options, modulePath);
     const changes = addImportToModule(source,
                                       modulePath,
                                       strings.classify(`${options.name}Module`),
@@ -68,6 +73,66 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
   };
 }
 
+function addRouteDeclarationToNgModule(
+  options: ModuleOptions,
+  routingModulePath: Path | undefined,
+): Rule {
+  return (host: Tree) => {
+    if (!options.route) {
+      return host;
+    }
+    if (!options.module) {
+      throw new Error('Module option required when creating a lazy loaded routing module.');
+    }
+
+    let path: string;
+    if (routingModulePath) {
+      path = routingModulePath;
+    } else {
+      path = options.module;
+    }
+
+    const text = host.read(path);
+    if (!text) {
+      throw new Error(`Couldn't find the module nor its routing module.`);
+    }
+
+    const sourceText = text.toString();
+    const addDeclaration = addRouteDeclarationToModule(
+      ts.createSourceFile(path, sourceText, ts.ScriptTarget.Latest, true),
+      path,
+      buildRoute(options, options.module),
+    ) as InsertChange;
+
+    const recorder = host.beginUpdate(path);
+    recorder.insertLeft(addDeclaration.pos, addDeclaration.toAdd);
+    host.commitUpdate(recorder);
+
+    return host;
+  };
+}
+
+function getRoutingModulePath(host: Tree, options: ModuleOptions): Path | undefined {
+  let path: Path | undefined;
+  const modulePath = options.module as string;
+  const routingModuleName = modulePath.split('.')[0] + '-routing';
+  const { module, ...rest } = options;
+
+  try {
+    path = findModuleFromOptions(host, { module: routingModuleName, ...rest });
+  } catch {}
+
+  return path;
+}
+
+function buildRoute(options: ModuleOptions, modulePath: string) {
+  const relativeModulePath = buildRelativeModulePath(options, modulePath);
+  const moduleName = `${strings.classify(options.name)}Module`;
+  const loadChildren = `() => import('${relativeModulePath}').then(m => m.${moduleName})`;
+
+  return `{ path: '${options.route}', loadChildren: ${loadChildren} }`;
+}
+
 export default function (options: ModuleOptions): Rule {
   return async (host: Tree) => {
     if (options.path === undefined) {
@@ -82,19 +147,41 @@ export default function (options: ModuleOptions): Rule {
     options.name = parsedPath.name;
     options.path = parsedPath.path;
 
+    let routingModulePath: Path | undefined;
+    const isLazyLoadedModuleGen = options.route && options.module;
+    if (isLazyLoadedModuleGen) {
+      options.routingScope = RoutingScope.Child;
+      routingModulePath = getRoutingModulePath(host, options);
+    }
+
     const templateSource = apply(url('./files'), [
-      options.routing ? noop() : filter(path => !path.endsWith('-routing.module.ts.template')),
+      options.routing || isLazyLoadedModuleGen && !!routingModulePath
+        ? noop()
+        : filter(path => !path.endsWith('-routing.module.ts.template')),
       applyTemplates({
         ...strings,
         'if-flat': (s: string) => options.flat ? '' : s,
+        lazyRoute: isLazyLoadedModuleGen,
+        lazyRouteWithoutRouteModule: isLazyLoadedModuleGen && !routingModulePath,
+        lazyRouteWithRouteModule: isLazyLoadedModuleGen && routingModulePath,
         ...options,
       }),
       move(parsedPath.path),
     ]);
+    const moduleDasherized = strings.dasherize(options.name);
+    const modulePath =
+      `${!options.flat ? moduleDasherized + '/' : ''}${moduleDasherized}.module.ts`;
 
     return chain([
-      addDeclarationToNgModule(options),
+      !isLazyLoadedModuleGen ? addDeclarationToNgModule(options) : noop(),
+      addRouteDeclarationToNgModule(options, routingModulePath),
       mergeWith(templateSource),
+      isLazyLoadedModuleGen
+        ? schematic('component', {
+            ...options,
+            module: modulePath,
+          })
+        : noop(),
       options.lintFix ? applyLintFix(options.path) : noop(),
     ]);
   };
--- a/packages/schematics/angular/module/index_spec.ts
+++ b/packages/schematics/angular/module/index_spec.ts
@@ -33,7 +33,7 @@ describe('Module Schematic', () => {
     name: 'bar',
     inlineStyle: false,
     inlineTemplate: false,
-    routing: false,
+    routing: true,
     skipTests: false,
     skipPackageJson: false,
   };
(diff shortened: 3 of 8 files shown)


packages/schematics/angular/module/files/__name@dasherize@if-flat__/__name@dasherize__.module.ts.template
```diff
@@ -1,13 +1,20 @@
 import { NgModule } from '@angular/core';<% if (commonModule) { %>
-import { CommonModule } from '@angular/common';<% } %><% if (routing) { %>
-
-import { <%= classify(name) %>RoutingModule } from './<%= dasherize(name) %>-routing.module';<% } %>
+import { CommonModule } from '@angular/common';<% } %><% if (lazyRoute && routeDeclarationInlined) { %>
+import { Routes, RouterModule } from '@angular/router';<% } %>
+<% if (routing || lazyRoute && !routeDeclarationInlined) { %>
```
alan-agius4: I won't extract these two conditions into template variables that are provided from the schematic itself, as they are bloating the template and they are making it hard to comprehend and view the important bit.

`routing || lazyRoute && !routeDeclarationInlined`
`lazyRoute && routeDeclarationInlined`

packages/schematics/angular/utility/project.ts
```diff
@@ -46,3 +71,19 @@ export function isWorkspaceSchema(workspace: any): workspace is WorkspaceSchema
 export function isWorkspaceProject(project: any): project is WorkspaceProject {
   return !!(project && (project as WorkspaceProject).projectType);
 }
+
+export function isProjectUsingIvy(host: Tree, project: WorkspaceProject): boolean {
+  const projectRoot = getProjectRoot(project);
+  const config = getProjectConfig(host, projectRoot + 'tsconfig.app.json');
```
alan-agius4: We need to read this from the workspace, we don't have any guarantee that this wan't changed.

packages/schematics/angular/utility/project.ts
```diff
@@ -35,6 +43,23 @@ export function getProject<TProjectType extends ProjectType = ProjectType.Applic
   return workspace.projects[projectName] as WorkspaceProject<TProjectType>;
 }
 
+export function getProjectConfig(host: Tree, path: string) {
+  const configName = path.split('/').pop();
```
alan-agius4: you can use `path.basename(path)` for this.

packages/schematics/angular/utility/project.ts
```diff
@@ -35,6 +43,23 @@ export function getProject<TProjectType extends ProjectType = ProjectType.Applic
...
+  const configFile = host.read(path);
+  if (!configFile) {
+    throw new Error(`Couldn't find project config (${configName}).`);
```
alan-agius4: This seem to be related to tsconfig. I would make it clear what config it is that it cannot be found.

packages/schematics/angular/utility/project.ts
```diff
@@ -35,6 +43,23 @@ export function getProject<TProjectType extends ProjectType = ProjectType.Applic
   return workspace.projects[projectName] as WorkspaceProject<TProjectType>;
 }
 
+export function getProjectConfig(host: Tree, path: string) {
```
alan-agius4: getProjectTsConfig Maybe? As config is more broad and might imply workspace project config,

packages/schematics/angular/module/index.ts
```diff
@@ -68,6 +70,74 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
...
+
+  return path;
+}
+
+function buildRoute(options: ModuleOptions, ivyEnabled: boolean) {
+  let loadChildren: string;
+  const modulePath = `./${options.name}/${options.name}.module`;
+  const moduleName = `${strings.classify(options.name)}Module`;
+
+  if (ivyEnabled) {
```
alan-agius4: The import syntax works for VE, hence we should always use this syntax

packages/schematics/angular/utility/project.ts
```diff
@@ -46,3 +71,19 @@ export function isWorkspaceSchema(workspace: any): workspace is WorkspaceSchema
 export function isWorkspaceProject(project: any): project is WorkspaceProject {
   return !!(project && (project as WorkspaceProject).projectType);
 }
+
+export function isProjectUsingIvy(host: Tree, project: WorkspaceProject): boolean {
```
alan-agius4: Should be unneeded

packages/schematics/angular/utility/project.ts
```diff
@@ -35,6 +43,23 @@ export function getProject<TProjectType extends ProjectType = ProjectType.Applic
...
+export function getProjectConfig(host: Tree, path: string) {
+  const configName = path.split('/').pop();
+  const configFile = host.read(path);
+  if (!configFile) {
+    throw new Error(`Couldn't find project config (${configName}).`);
+  }
+  const configText = configFile.toString('utf-8');
+  let config: ProjectConfig;
+  try {
+    config = JSON.parse(configText || '{}');
```
alan-agius4: TS config can contain comments,
 
You should use the JSON parser.

packages/schematics/angular/module/index.ts
```diff
@@ -68,6 +70,74 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
...
+
+    const recorder = host.beginUpdate(path);
+    recorder.insertLeft(addDeclaration.pos, addDeclaration.toAdd);
+    host.commitUpdate(recorder);
+
+    return host;
+  };
+}
+
+function getRoutingModulePath(host: Tree, options: ModuleOptions): Path | undefined {
```
alan-agius4: This doesn't seem right. This method will always fall in the `catch` because you cannot locate a routing module with `findModuleFromOptions`

https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/utility/find-module.ts#L110

packages/schematics/angular/module/index.ts
```diff
@@ -68,6 +70,74 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
...
+  try {
+    path = findModuleFromOptions(host, { module: routingModuleName, ...rest });
+  } catch {}
+
+  return path;
+}
+
+function buildRoute(options: ModuleOptions, ivyEnabled: boolean) {
+  let loadChildren: string;
+  const modulePath = `./${options.name}/${options.name}.module`;
```
alan-agius4: Paths should be dasharized.

packages/schematics/angular/module/index.ts
```diff
@@ -82,18 +152,37 @@ export default function (options: ModuleOptions): Rule {
...
     const templateSource = apply(url('./files'), [
-      options.routing ? noop() : filter(path => !path.endsWith('-routing.module.ts.template')),
+      options.routing || isLazyLoadedModuleGen && !!routingModulePath
+        ? noop()
+        : filter(path => !path.endsWith('-routing.module.ts.template')),
       applyTemplates({
         ...strings,
         'if-flat': (s: string) => options.flat ? '' : s,
+        lazyRoute: isLazyLoadedModuleGen,
+        routeDeclarationInlined: !routingModulePath,
```
alan-agius4: Why would we want this? If no routing module is found, we should generate a routing module.

packages/schematics/angular/module/index.ts
```diff
@@ -82,18 +152,37 @@ export default function (options: ModuleOptions): Rule {
...
     ]);
 
     return chain([
-      addDeclarationToNgModule(options),
+      !isLazyLoadedModuleGen ? addDeclarationToNgModule(options) : noop(),
+      addRouteDeclarationToNgModule(options, project, routingModulePath),
+      isLazyLoadedModuleGen
+        ? schematic('component', {
+            ...options,
+            skipImport: true,
```
alan-agius4: The lazy loaded component needs to be imported in the lazy loaded module.

packages/schematics/angular/module/index.ts
```diff
@@ -68,6 +70,74 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
...
+function addRouteDeclarationToNgModule(
+  options: ModuleOptions,
+  project: WorkspaceProject,
+  routingModulePath: Path | undefined,
+): Rule {
+  return (host: Tree) => {
+    if (!options.route) {
+      return host;
+    }
+    if (options.route && !options.module) {
```
alan-agius4: `options.route && !options.module` -> `!options.module`

packages/schematics/angular/module/index.ts
```diff
@@ -68,6 +70,74 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
...
+      throw new Error('Module option required when creating a lazy loaded routing module.');
+    }
+
+    let path: string;
+    if (routingModulePath) {
```
alan-agius4: The castings here should be redundant otherwise the checks are not correct.

packages/schematics/angular/module/index.ts
```diff
@@ -68,6 +70,74 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
...
+      path = options.module as string;
+    }
+
+    const text = host.read(path);
+    if (!text) {
+      throw new Error(`Couldn't find the module nor its routing module.`);
+    }
+
+    const ivyEnabled = isProjectUsingIvy(host, project);
+    const sourceText = text.toString('utf-8');
```
alan-agius4: `toString()` should be enough by default it is utf8

packages/schematics/angular/module/files/__name@dasherize@if-flat__/__name@dasherize__.module.ts.template
hawkgs: Do you mean that it is better to stick with using the generator `options` and compose the checks inside the template rather than creating new template variables that incorporate them? Sorry, if I didn't understand you correct.

packages/schematics/angular/utility/project.ts
hawkgs: Thanks for noting :+1:

packages/schematics/angular/utility/project.ts
hawkgs: `configName` displays the file name. Am I missing something?

packages/schematics/angular/utility/project.ts
hawkgs: Yes, totally makes sense. :+1:

packages/schematics/angular/utility/project.ts
hawkgs: Thanks for the catch!

packages/schematics/angular/module/index.ts
hawkgs: Actually, it should work. Frankly, I haven't looked into the function itself (`findModuleFromOptions`) but if it always used to hit the `catch`, [this test](https://github.com/angular/angular-cli/blob/389f9c7af55063f51d7a78e16af7ab163315346d/packages/schematics/angular/module/index_spec.ts#L148) would always fail.

In other words, we won't be able to access `app-routing.module.ts` and insert the declaration in that case.

packages/schematics/angular/module/index.ts
hawkgs: Thanks for the catch :+1:

packages/schematics/angular/module/index.ts
hawkgs: Because we would like to stick to the preferred by the user strategy, i.e.: route declarations reside in parent module or in parent routing module. If, for example, the parent module has the declarations inlined, we won't generate a child routing module but rather put the routes in the child module. This is part of the document:

> Depending on whether we’ve inserted the route in the parent routing module or in the parent module [i.e. where is the `RouterModule` declared] we’ll determine if we want to generate a child routing module.

packages/schematics/angular/module/index.ts
hawkgs: Ah, I am manually doing this in the template rather than making use of the component schematics functionality. Makes sense.

packages/schematics/angular/module/index.ts
hawkgs: You are right. They are redundant.

packages/schematics/angular/module/index.ts
hawkgs: That's my fault. Fortunately, this will decrease the size of the change this PR introduces.

packages/schematics/angular/module/files/__name@dasherize@if-flat__/__name@dasherize__.module.ts.template
alan-agius4: I meant template options, example
```ts
      applyTemplates({
        ...strings,
        lazyRoute: routing || lazyRoute && !routeDeclarationInlined,
        ...,
      }),
```

packages/schematics/angular/module/index.ts
alan-agius4: Makes sense, thanks

packages/schematics/angular/module/index.ts
alan-agius4: I think this should be done via component schematic.

packages/schematics/angular/utility/project.ts
alan-agius4: Most likely actually this method will become redundant.

hawkgs: Indeed

packages/schematics/angular/module/index.ts
hawkgs: I just pointed that I am doing it the wrong way, apparently. I'll have to keep the import in the routing module though, since we don't need the component added to the declarations array there.

packages/schematics/angular/module/files/__name@dasherize@if-flat__/__name@dasherize__.module.ts.template
hawkgs: I see, but using the `routing` option doesn't necessarily mean that we have a lazy route. Besides, we have both `routing || lazyRoute && !routeDeclarationInlined` && `lazyRoute && routeDeclarationInlined`. 

What about combining just the current `lazyRoute` and `routeDeclarationInlined`? Something like `lazyRouteWithoutRouteMod` and `lazyRouteWithRouteMod`? Still long names though. Do you think that's a good idea?

alan-agius4: This is good, thought I’d go for the full name, ie: `Module` instead of `Mod` as it’s easier to comprehend.

hawkgs: Yeah, it's not worth saving 3 letters in the sake of less comprehensive naming. Will do it :+1:

packages/schematics/angular/utility/ast-utils.ts
```diff
@@ -568,3 +579,89 @@ export function isImported(source: ts.SourceFile,
...
+  } else {
+    const routesVarName = routesArg.getText();
+    const routesVar = source.statements
+      .filter((s: ts.Statement) => s.kind === ts.SyntaxKind.VariableStatement)
+      .find((v: ts.VariableStatement) => {
+        return v.declarationList.declarations[0].name.getText() === routesVarName;
+      }) as ts.VariableStatement | undefined;
+
+    if (!routesVar) {
+      throw new Error(`The route declaration variable "${routesVarName}" is used but not defined.`);
```
clydin: Can the error provide file and position information as well?  Same for others where applicable.

hawkgs: I'll see what I can do about that. :+1:

hawkgs: I assume the line of the error and the file name will be enough? Intentionally omitted the column/character position since I think that it may just add an additional noise.

alan-agius4:
Can you please rebase so we make this green?

packages/schematics/angular/module/index.ts
```diff
@@ -68,6 +68,67 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
...
+      path = options.module;
+    }
+
+    const text = host.read(path);
+    if (!text) {
+      throw new Error(`Couldn't find the module nor its routing module.`);
+    }
+
+    const sourceText = text.toString();
+    const addDeclaration = addRouteDeclarationToModule(
```
alan-agius4: Nit: 
```ts
    const sourceFile = ts.createSourceFile(path, sourceText, ts.ScriptTarget.Latest, true);
    const changes = addRouteDeclarationToModule(source, path, buildRoute(options));
    const output = applyChanges(path, text , changes);
```

hawkgs:
Done.

packages/schematics/angular/module/index_spec.ts
```diff
@@ -118,4 +118,75 @@ describe('Module Schematic', () => {
       .toPromise();
     expect(appTree.files).toContain('/projects/bar/custom/app/foo/foo.module.ts');
   });
+
+  describe('lazy route generator', () => {
```
alan-agius4: Can we add a couple of tests when the module is flat?

Ie when using the `flat` option?

From the code, I don't see that this is being handled at the moment.

packages/schematics/angular/utility/ast-utils.ts
```diff
@@ -344,6 +344,28 @@ export function getFirstNgModuleName(source: ts.SourceFile): string|undefined {
   return moduleClass.name.text;
 }
 
+export function getMetadataField(
```
alan-agius4: I believe this can be simplified a bit

```ts
export function getMetadataField(
  node: ts.ObjectLiteralExpression,
  metadataField: string,
): ts.ObjectLiteralElement[] {
  return node.properties
    .filter(prop => ts.isPropertyAssignment(prop))
    // Filter out every fields that's not "metadataField". Also handles string literals
    // (but not expressions).
    .filter(({ name }) => {
      return (ts.isIdentifier(name) || ts.isStringLiteral(name))
        && name.text === metadataField;
    });
}


```

hawkgs: Sounds good. By the way, this is more like a question -- is there any valid reason for using `text` property over `getText()`? I noticed that in one of the functions that I introduced, using `text` breaks the tests because it's `undefined`, while `getText` returns `''` (ref: `typescript.js:96792`). Maybe using `getText` is safer since it always returns a string (not that it matters in the current change)?

packages/schematics/angular/module/index_spec.ts
hawkgs: Thanks for the catch! Indeed, I haven't taken care of this. I added a single test since there aren't edge cases that I could come up with.

packages/schematics/angular/utility/ast-utils.ts
alan-agius4: When using to compare, returning `undefined` is perfectly fine. using `text` directly. Is slightly more performant.

packages/schematics/angular/module/index.ts
```diff
@@ -68,6 +68,67 @@ function addDeclarationToNgModule(options: ModuleOptions): Rule {
...
+  try {
+    path = findModuleFromOptions(host, { module: routingModuleName, ...rest });
+  } catch {}
+
+  return path;
+}
+
+function buildRoute(options: ModuleOptions) {
+  const dasherizedName = strings.dasherize(options.name);
+  const modulePath = `./${!options.flat ? dasherizedName + '/' : ''}${dasherizedName}.module`;
```
alan-agius4: This still doesn't seem right, As a module can be nested when when is flat, while it's flat it is not not in the same folder of the `app.module`

```
ng g m --flat /foo/bar/baz
```

You should re-use the `relativePath` to have the correct module path https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/module/index.ts#L53 instead

Also, you should always take into consideration the `options.path` option.

packages/schematics/angular/module/index.ts
```diff
@@ -82,19 +143,42 @@ export default function (options: ModuleOptions): Rule {
...
+      `${!options.flat ? moduleDasherized + '/' : ''}${moduleDasherized}.module.ts`;
 
     return chain([
-      addDeclarationToNgModule(options),
+      !isLazyLoadedModuleGen ? addDeclarationToNgModule(options) : noop(),
+      addRouteDeclarationToNgModule(options, routingModulePath),
       mergeWith(templateSource),
+      isLazyLoadedModuleGen
+        ? schematic('component', {
+            ...modulelessOptions,
```
alan-agius4: Nit: 
```
{
...options,
module: modulePath,
}
```

packages/schematics/angular/module/index.ts
hawkgs: Ah, got it.

packages/schematics/angular/module/index.ts
hawkgs: I think it's safer to keep it with `modulelessOptions` because, if we use spread on `options`, we are overriding `module`. That way, the validity of this statement is dependent on the order which might be implicit to the readers (i.e. if someone switches the positions of these two, for some reason, we will end up with a bug). Anyway, that's my personal perception.

alan-agius4:
Thanks for your contribution and the awesome efforts you put in this PR. Keep up the good work, and I hope that we can have you back with another PR soon :)

PR #14348, opened 2019-05-07
A little space

ommehaz:
just a space too much


alan-agius4:
Thanks for this however it seems that the CLA was not signed.

PR #14349, opened 2019-05-07
ci: test on node 12

filipesilva:


filipesilva:
The `tests/basic/update-7.0` e2e test fails with the error below:

```
ENOENT: no such file or directory, stat '/tmp/angular-cli-e2e-201947-111-1sw3b9b.v5l8/assets/7.0-project-1557230497707/node_modules/.bin/in-install'
```

This seems to be related to node-sass rollback when using `--force` on update. https://github.com/angular/angular-cli/pull/14331 removes the need to force on update so that should clear this too.

.circleci/config.yml
```diff
@@ -115,6 +116,22 @@ jobs:
...
+      - image: *browsers_docker_image_node_12
+    environment:
+      BASH_ENV: ~/.profile
+      NPM_CONFIG_PREFIX: ~/.npm-global
+    resource_class: xlarge
+    parallelism: 4
+    steps:
+      - attach_workspace: *attach_options
+      - run: PATH=~/.npm-global/bin:$PATH xvfb-run -a node ./tests/legacy-cli/run_e2e --nb-shards=${CIRCLE_NODE_TOTAL} --shard=${CIRCLE_NODE_INDEX}
+      - store_artifacts:
```
clydin: Remove this part so it doesn't overwrite the artifacts from the other job.

filipesilva: Right you are, done.

PR #14350, opened 2019-05-07
Fixes to serve path

alan-agius4:
fix(@angular-devkit/build-angular): using `servePath` results in resources not being found

At the moment the default baseHref is `/`, which means resources won't be found when using the `servePath`. Because the later only effect the serving of the file and doesn't change the paths in the script tag (rightly so). However having the baseHref as `/` will cause the resources to be loaded from root.

fix(@angular-devkit/build-angular): `servePath` is not respected in protractor

### DIFF
--- a/packages/angular_devkit/build_angular/src/dev-server/index.ts
+++ b/packages/angular_devkit/build_angular/src/dev-server/index.ts
@@ -107,6 +107,12 @@ export function serveWebpackBrowser(
     // In dev server we should not have budgets because of extra libs such as socks-js
     overrides.budgets = undefined;
 
+    // When having a serve path we need to change the
+    // baseHref from '/' as otherwise resources won't be found.
+    if (options.servePath && !(rawBrowserOptions.baseHref || options.baseHref)) {
+      overrides.baseHref = '';
+    }
+
     const browserName = await context.getBuilderNameForTarget(browserTarget);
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -107,6 +107,12 @@ export function serveWebpackBrowser(
     // In dev server we should not have budgets because of extra libs such as socks-js
     overrides.budgets = undefined;
 
+    // When having a serve path we need to change the
+    // baseHref from '/' as otherwise resources won't be found.
+    if (options.servePath && !(rawBrowserOptions.baseHref || options.baseHref)) {
+      overrides.baseHref = '';
```
clydin: Won't this break routing?

The serve path option is intended as an advanced option used to override the default serve path which is calculated from the deploy URL and base HREF (useful in reverse proxy scenarios, for instance).  Using the serve path option on its own, will most likely break the application in some form.

alan-agius4: I see, I will test it with routing, but if the servePath is indented to be used with baseHref I will remove this part and add the baseHref in the test.

alan-agius4: @clydin, the router seems to be working fine. Not sure if you prefer to keep the new logic for the baseHref or remove this and rely on he fact that the baseHref will be provided by the users

PR #14394, opened 2019-05-10
Update command option checking/error reporting improvements

clydin:
This adds extensive option checking and error handling with the goal to catch user errors as early in the process as possible.  Bad option combinations and/or values will result in actionable information and stop the process before network access and processing occurs.  Executing migration packages are now also displayed to the user as well.

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -5,62 +5,265 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { normalize } from '@angular-devkit/core';
+import * as path from 'path';
+import * as semver from 'semver';
 import { Arguments, Option } from '../models/interface';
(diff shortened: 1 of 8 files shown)


packages/angular/cli/commands/update-impl.ts
```diff
@@ -5,62 +5,268 @@
...
+      return 1;
+    }
 
-  async run(options: UpdateCommandSchema & Arguments) {
     const packageManager = getPackageManager(this.workspace.root);
+    this.logger.info(`Using package manager: '${packageManager}'`);
+
+    if (options.all || packages.length === 0) {
+      // Either update all packages or show status
+      return this.runSchematic({
```
alan-agius4: Why do we have two `force` options?

clydin: The first is the schematic runtime force option which allows overwriting of already existing files on create actions (https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/schematics/src/sink/host.ts#L31)

The second is the update schematic's option.

PR #14406, opened 2019-05-12
fix(@angular-devkit/build-optimizer): remove deprecated test functions from public API

clydin:


clydin:
They're still used internally (for now).  This allows them to be removed/altered without breaking anyone moving forward.

PR #14412, opened 2019-05-13
fix(@angular-devkit/schematics): deprecate synchronous testing methods

clydin:
All schematics can potentially be async.  The synchronous test method variants (`runSchematic`/`runExternalSchematic`) will fail if the schematic, any of its rules, or any schematic it calls are async.

### DIFF
--- a/packages/angular_devkit/schematics/testing/schematic-test-runner.ts
+++ b/packages/angular_devkit/schematics/testing/schematic-test-runner.ts
@@ -85,6 +85,12 @@ export class SchematicTestRunner {
       .pipe(map(tree => new UnitTestTree(tree)));
   }
 
+  /**
+   * @deprecated Since v8.0.0 - Use {@link SchematicTestRunner.runSchematicAsync} instead.
+   * All schematics can potentially be async.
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/schematics/testing/schematic-test-runner.ts
```diff
@@ -85,6 +85,11 @@ export class SchematicTestRunner {
       .pipe(map(tree => new UnitTestTree(tree)));
   }
 
+  /**
+   * @deprecated Use `runSchematicAsync`.  All schematics can potentially be async. This
+   * synchronous variant will fail if the schematic, any of its rules, or any schematics
+   * it calls are async.
+   */
```
alan-agius4: Maybe we should point out in which version these will be removed? 

It will also be useful for us when cutting a major release to know if we should remove it or not since these I guess will be removed in version 10 correct?

clydin: Added a "deprecated since" version.  This will land in 8.0.0 so they can probably go 9.0.

PR #14414, opened 2019-05-13
Support dynamic import of lazy route through reexports

### LINKED ISSUE #14399: Dynamic Imports broken when pointing at a barrel file
# 🐞 Bug report

[x] build --aot
[x] serve --aot

### Is this a regression?

Sort of, using a loadChildren string works. But Dynamic imports which are supposed to be a drop in replacement, do not work.

### Description

When a loadChildren callback uses a dynamic import to bring in a lazy loaded module from a barrel file, the path gets changed to `./path-to-barrel.ngfactory` in AoT. Which does not work.

## 🔬 Minimal Reproduction

https://github.com/FrozenPandaz/angular-bugs/tree/dyna-imp-bug

## 🔥 Exception or Error
Module not found: Error: Can't resolve './shared.ngfactory' in '/home/jason/projects/temp/test-proj/src/app'

The src is 
```ts
loadChildren: import('./shared').then(mod => mod.LazyModule)
```

## 🌍 Your Environment
Latest rc

**Anything else relevant?**
This is important for Nx, which has public apis for its libs.

filipesilva:
Fix #14399


packages/ngtools/webpack/src/transformers/import_factory.ts
```diff
@@ -176,20 +183,51 @@ function replaceImport(
...
+  // Try to resolve the import. It might be a reexport from somewhere and the ngfactory will only
+  // be present next to the original module.
+  const exportedSymbol = typeChecker.getSymbolAtLocation(exportNameId);
+  if (!exportedSymbol) {
+    return warnAndBail();
+  }
+
+  // Symbols have parents. But it's an internal thing. We shouldn't use it.
+  // TODO: find a better way of discovering the file where this symbol is declared.
+  type SymbolWithParent = ts.Symbol & { parent?: SymbolWithParent };
```
filipesilva: There must be a better way to do this. Still looking for it.

packages/ngtools/webpack/src/transformers/import_factory.ts
```diff
@@ -176,20 +183,51 @@ function replaceImport(
...
+    rootSymbol = rootSymbol.parent;
+  }
+
+  if (!ts.isSourceFile(rootSymbol.valueDeclaration)) {
+    return warnAndBail();
+  }
+
+  // SourceFile have paths. But it's an internal thing. We shouldn't use it.
+  // TODO: find a better way of discovering the source file path.
+  type SourceFileWithPath = ts.SourceFile & { path: string };
```
filipesilva: Same as above.

Issue #14355, opened 2019-05-07
Telemetry: bundle size, build times, ngOnInit metrics aren't reported


alexeagle:
Split out from #14190

PR #14417, opened 2019-05-13
fix(@angular/cli): prevent nested schematic execution from generating analytics

### LINKED ISSUE #14190: Telemetry: don't track implicitly run schematics

clydin:
Fixes #14190

### DIFF
--- a/packages/angular/cli/models/schematic-command.ts
+++ b/packages/angular/cli/models/schematic-command.ts
@@ -265,47 +265,11 @@ export abstract class SchematicCommand<
           analytics: this.analytics,
         };
       } else {
-        return {
-          ...context,
-          analytics: new analytics.NoopAnalytics(),
-        };
+        return context;
       }
     });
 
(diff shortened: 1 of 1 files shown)


filipesilva:
LGTM but I think this only addresses one out of the three points in https://github.com/angular/angular-cli/issues/14190. It shouldn't be closed just by this PR.

clydin:
2nd point was addressed by #14205
3rd point was addressed by #14395 and broken out into separate issue #14355

filipesilva:
Makes sense now, thanks for explaining it to me 👍

PR #14426, opened 2019-05-14
fix(@angular/cli): @bazel/* are not shown in `ng version`

kyliau:
Update `ng version` to show versions for @bazel/* packages to help users
report meaningful errors when they opt in to Bazel.


alan-agius4:
@kyliau, LGTM. Kindly check slack convo re the target label.

kyliau:
Updated commit message, fix should target both master & patch.
Thank you, @alan-agius4 !

PR #14441, opened 2019-05-16
fix(@schematics/angular): update or add module and target to non extend…

### LINKED ISSUE #14436: Older CLI projects don't match the tsconfig layout of newer projects
# 🐞 Bug report

```
- [x] update
```

### Is this a regression?
No

### Description
Older projects (such as https://github.com/GDGTC/devfest/commits/master/src/tsconfig.json) have a single tsconfig file. This was never migrated by our tooling to `tsconfig.app.json` which means that some of our assumptions about these files are incorrect. This breaks things such as the schematics in 8 to take advantage of differential loading.

We should consider detecting these configs and migrating them.

alan-agius4:
…ed tsconfig

Fixes #14436

### DIFF
--- a/packages/schematics/angular/migrations/update-8/differential-loading.ts
+++ b/packages/schematics/angular/migrations/update-8/differential-loading.ts
@@ -6,7 +6,6 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {
-  JsonAstObject,
   JsonParseMode,
   isJsonObject,
   join,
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/update-8/differential-loading.ts
```diff
@@ -171,10 +128,37 @@ function getCompilerOptionsAstObject(host: Tree, tsConfigPath: string): JsonAstO
     return;
   }
 
+  const configExtends = findPropertyInAstObject(tsCfgAst, 'extends');
+  const isExtendedConfig = configExtends && configExtends.kind === 'string';
```
filipesilva: @clydin asked this check to be stricter:

> Can we add some safety checks to the extends checks for the tsconfig?  If it extends from more than one or if the depth of extension is more than one, it should probably bail.
> a config that’s heavily customized will probably break if the migration starts modifying options

alan-agius4: Let’s take this on slack, bailing out increases the changes of leaving broken projects as this migration is mandatory for other migrations such as for the lazy loading import syntax change.

alexeagle:
@alan-agius4 this is too risky for 8 but @StephenFluin asks, could we find the tsconfig in the old location and update the module/target setting there?

alan-agius4:
@alexeagle thats exactly what this PR does.

packages/schematics/angular/migrations/update-8/differential-loading.ts
alexeagle: sounds right to me - users can always look at the diff to understand what we did

PR #14468, opened 2019-05-19
fix for #14467 - fix(@schematics/angular): webworker snippet should not generate to module

### LINKED ISSUE #14467: Code generated for webWorker in incorrect file
# 🐞 Bug report

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

Not Sure

### Description

Create a new app with routing as Yes
run "ng g webWorker app"

webWorker code is added to app-routing.module.ts file

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍 Your Environment
<pre><code>

  "@angular/cli": "~8.0.0-rc.3",
"@angular/common": "~8.0.0-rc.4",
</code></pre>

**Anything else relevant?**

santoshyadav198613:
Fixes #14467

### DIFF
--- a/packages/schematics/angular/web-worker/index.ts
+++ b/packages/schematics/angular/web-worker/index.ts
@@ -63,8 +63,10 @@ function addSnippet(options: WebWorkerOptions): Rule {
     }
 
     const siblingModules = host.getDir(options.path).subfiles
-      // Find all files that start with the same name, are ts files, and aren't spec files.
-      .filter(f => f.startsWith(options.name) && f.endsWith('.ts') && !f.endsWith('spec.ts'))
+      // Find all files that start with the same name, are ts files,
+      // and aren't spec or module files.
+      .filter(f => f.startsWith(options.name) && f.endsWith('.ts')
+      && !f.endsWith('spec.ts') && !f.endsWith('-.module.ts'))
       // Sort alphabetically for consistency.
       .sort();
 
--- a/packages/schematics/angular/web-worker/index.ts
+++ b/packages/schematics/angular/web-worker/index.ts
@@ -62,11 +62,11 @@ function addSnippet(options: WebWorkerOptions): Rule {
       return;
     }
 
+    const fileRegExp = new RegExp(`^${options.name}.*\.ts`);
     const siblingModules = host.getDir(options.path).subfiles
       // Find all files that start with the same name, are ts files,
       // and aren't spec or module files.
-      .filter(f => f.startsWith(options.name) && f.endsWith('.ts')
-      && !f.endsWith('spec.ts') && !f.endsWith('-.module.ts'))
+      .filter(f => fileRegExp.test(f) && !/(module|spec)\.ts$/.test(f))
       // Sort alphabetically for consistency.
       .sort();
 
--- a/packages/schematics/angular/web-worker/index_spec.ts
+++ b/packages/schematics/angular/web-worker/index_spec.ts
@@ -11,7 +11,7 @@ import { Schema as WorkspaceOptions } from '../workspace/schema';
 import { Schema as WebWorkerOptions } from './schema';
 
 
-describe('Service Worker Schematic', () => {
+describe('Web Worker Schematic', () => {
   const schematicRunner = new SchematicTestRunner(
     '@schematics/angular',
     require.resolve('../collection.json'),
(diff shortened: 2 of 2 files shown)


alan-agius4:
@santoshyadav198613, thanks for this. Can you kindly fix the `fixes` in the commit message as well?
Thanks.

santoshyadav198613:
Sure @alan-agius4 , will do thanks.

packages/schematics/angular/web-worker/index.ts
```diff
@@ -140,7 +140,8 @@ function addSnippet(options: WebWorkerOptions): Rule {
 
     const siblingModules = host.getDir(options.path).subfiles
       // Find all files that start with the same name, are ts files, and aren't spec files.
-      .filter(f => f.startsWith(options.name) && f.endsWith('.ts') && !f.endsWith('spec.ts'))
+      .filter(f => f.startsWith(options.name)
```
alan-agius4: @filipesilva should web-workers should be inserted only to components?

@santoshyadav198613 - the statements here can be simplified slightly, if the above is the case.
```ts
.filter(f => f.startsWith(options.name) && f.endsWith('component.ts'))
```

filipesilva: Now that you mention it, it's very common to insert them into services as well I believe. I think we should specialcase spec files and routing files actually...

So something like this:
```
.filter(f => f.startsWith(options.name) && f.endsWith('.ts') 
  && !f.endsWith('spec.ts') && !f.endsWith('-routing.module.ts'))
```

These two are indeed special because our generators make them automatically.

santoshyadav198613: But in that case, we have interfaces, directives as well.

filipesilva: Yes, it should be ok to put a webworker in those. Specs and routing files are different because are auxiliary to the other, more important, file. And in this specific code that is problematic because we sort by filename.

santoshyadav198613: thanks a lot, will make the changes. You guys are super helpful.

santoshyadav198613: Also i think we should not consider moduel at all so changing it to
 .filter(f => f.startsWith(options.name) && f.endsWith('.ts') 
  && !f.endsWith('spec.ts') && !f.endsWith('-.module.ts')) 
would be better.

filipesilva: I think that's a fair assumption, yes. That looks good to me.

santoshyadav198613: Done with the changes.

packages/schematics/angular/web-worker/index.ts
```diff
@@ -139,8 +139,10 @@ function addSnippet(options: WebWorkerOptions): Rule {
     }
 
     const siblingModules = host.getDir(options.path).subfiles
-      // Find all files that start with the same name, are ts files, and aren't spec files.
-      .filter(f => f.startsWith(options.name) && f.endsWith('.ts') && !f.endsWith('spec.ts'))
+      // Find all files that start with the same name, are ts files,
+      // and aren't spec or module files.
+      .filter(f => f.startsWith(options.name) && f.endsWith('.ts')
+      && !f.endsWith('spec.ts') && !f.endsWith('-.module.ts'))
```
alan-agius4: This doesn't look right the `-.module.ts` will not match any file that is generated via the CLI.

I think you meant something like
```ts
.filter(f => f.startsWith(options.name) && f.endsWith('.ts')
      && !/\.(module|spec)\.ts$/.test(f))

```

packages/schematics/angular/web-worker/index_spec.ts
```diff
@@ -35,7 +35,7 @@ describe('Service Worker Schematic', () => {
     name: 'bar',
     inlineStyle: false,
     inlineTemplate: false,
-    routing: false,
+    routing: true,
```
alan-agius4: Changing this value, will ensure that the new logic is tested.

santoshyadav198613:
> Hi @santoshyadav198613, sorry to barge in your PR and pushing directly. But we need to get this merged ASAP for the version 8 final.
> 
> Thanks a lot for your contribution.
> 
> //cc @alexeagle, this is a go!

thanks @alan-agius4 , sorry was not feeling well, so was not able to make changes last night. But happy that I was able to find the issue.

Issue #14424, opened 2019-05-14
Fix forwardRef in ES2015 via a pass


filipesilva:
We should address https://github.com/angular/angular/issues/30106 by adding an extra transformer that transforms the `__metadata` to a mini-forwardRef to avoid the error.

PR #14473, opened 2019-05-19
fix(@ngtools/webpack): downlevel constructor parameter type information

clydin:
The Angular JIT compiler uses the output of TypeScript's `emitDecoratorMetadata` to support dependency injection at runtime.  This option stores type and DI token information so that the runtime JIT compiler can inject the proper provider.  This option is not needed for AOT compilation and the metadata is indirectly stripped from the output due to all Angular decorators being removed during the compilation process.

Unfortunately, the `emitDecoratorMetadata` option does not support forward references (or more generally circular dependencies) within ES2015+ output due to the TypeScript compiler transforming the metadata into a direct reference of the yet to be defined class.  This results in a runtime TDZ error.  This is a known limitation of the `emitDecoratorMetadata` option ([reference](https://github.com/microsoft/TypeScript/issues/27519)).  This situation is not unique to Angular and can occur in any TypeScript code.  As it is marked as a design limitation, it is considered that it cannot be fixed within the current architecture of TypeScript.

As a result, the change in this PR removes the need for the use of the option by storing the DI information in a static property function via the use of the tsickle `ctorParameters` transformation.  By leveraging the existing functionality, no changes to the framework code is necessary.  Also, minimal new code is required within the CLI, as the relevant tsickle code can be extracted and used with several local modifications.

For forward references to function in ES2015+, the `emitDecoratorMetadata` option must be disabled based on the current implementation.  To support both the option and forward references, the metadata generated by TypeScript would need to be altered to remove the type information that would trigger a TDZ error.  However, by doing so the metadata information is no longer useful to Angular and technically incorrect as it will not contain the information from the source code as described by the option.

Closes: #14424

### DIFF
--- a/packages/schematics/angular/library/index.ts
+++ b/packages/schematics/angular/library/index.ts
@@ -104,6 +104,11 @@ function addDependenciesToPackageJson() {
         version: '^5.1.0',
       },
       {
+        type: NodeDependencyType.Dev,
+        name: 'tsickle',
+        version: '^0.35.0',
+      },
+      {
         type: NodeDependencyType.Default,
         name: 'tslib',
         version: latestVersions.TsLib,
--- a/packages/schematics/angular/workspace/files/tsconfig.json.template
+++ b/packages/schematics/angular/workspace/files/tsconfig.json.template
@@ -7,7 +7,6 @@
     "declaration": false,
     "module": "esnext",
     "moduleResolution": "node",
-    "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "importHelpers": true,
     "target": "es2015",
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -60,6 +60,7 @@ import {
   replaceServerBootstrap,
 } from './transformers';
 import { collectDeepNodes } from './transformers/ast_helpers';
+import { downlevelConstructorParameters } from './transformers/ctor-parameters';
 import {
   AUTO_START_ARG,
 } from './type_checker';
@@ -922,6 +923,9 @@ export class AngularCompilerPlugin {
       // Replace resources in JIT.
       this._transformers.push(
         replaceResources(isAppPath, getTypeChecker, this._options.directTemplateLoading));
+      // Downlevel constructor parameters for DI support
+      // This is required to support forwardRef in ES2015 due to TDZ issues
+      this._transformers.push(downlevelConstructorParameters(getTypeChecker));
     } else {
       // Remove unneeded angular decorators.
       this._transformers.push(removeDecorators(isAppPath, getTypeChecker));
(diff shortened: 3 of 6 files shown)


packages/schematics/angular/workspace/files/tsconfig.json.template
```diff
@@ -7,7 +7,6 @@
     "declaration": false,
     "module": "esnext",
     "moduleResolution": "node",
-    "emitDecoratorMetadata": true,
```
alan-agius4: `emitDecoratorMetadata` is required for libraries. I think when removing this we should add `emitDecoratorMetadata` in the library tsconfig.

clydin: I was hoping to add the transformer to library compilation as well but was unable to find an extension point within ng-packgr.

Added the option to the library schematic pending TS transformer support for ng-packgr.

packages/schematics/angular/library/index.ts
```diff
@@ -103,6 +103,11 @@ function addDependenciesToPackageJson() {
         name: 'ng-packagr',
         version: '^5.1.0',
       },
```
alan-agius4: //cc @alexeagle kinda note this change!

Splaktar:
This seems to be at least partially related to https://github.com/angular/angular/issues/30586.

trotyl:
@IgorMinar I believe https://github.com/angular/angular/issues/25039 would be a better approach, which eliminate the need for `emitDecoratorMetadata` at semantic level, and will not result any inconsistency with or without using Angular CLI.

IgorMinar:
I agree that angular/angular#25039 is better but it's not a change we can implement in 8.0. It's too late for that, we are in an RC and in risk-averse mode. We can however implement it in 8.1.

Keep in mind that while that solution is better, this PR is sufficient to prevent regressions in 8.0 - to ensure that things that worked prior to 8.0 still work with 8.0 and all the new defaults (e.g. differential loading). The proposal described by Paul enables something that never worked before - bazel+ssr. We should do this but again, not in 8.0.

So the least risky solution is to
- go ahead with this PR as it's low risk and unblocks the release
- improve upon this via a solution described in angular/angular#25039 in 8.1

IgorMinar:
I agree that angular/angular#25039 is better but it's not a change we can implement in 8.0. It's too late for that, we are in an RC and in risk-averse mode. We can however implement it in 8.1.

Keep in mind that while that solution is better, this PR is sufficient to prevent regressions in 8.0 - to ensure that things that worked prior to 8.0 still work with 8.0 and all the new defaults (e.g. differential loading). The proposal described by Paul enables something that never worked before - bazel+ssr. We should do this but again, not in 8.0.

So the least risky solution is to
- go ahead with this PR as it's low risk and unblocks the release
- improve upon this via a solution described in angular/angular#25039 in 8.1

IgorMinar:
With regards to angular/angular#25039 - it's solving a very different problem and while it could potentially also solve this one, it would only work for new code and not for any of the existing application/library code which would need to be migrated.

evmar:
Does this mean we can remove the decorator downlevelling code from tsickle?  When?  (I think it is great that Angular owns this particular code because I never understood it well.)

alexeagle:
@evmar I think that could happen already, tsickle is already an optional dependency from Angular ecosystem. It's really hard for us to be sure about the consequences without trying it though - there might be users like Lucidchart who still depend on that behavior

PR #14492, opened 2019-05-22
fix(@angular/cli): ensure git repository is clean before updates

clydin:
If not asking for update status then this change checks for a clean git repository before proceeding.  This allows the user to easily reset any changes from the update and avoid any developer code changes from being mixed with update changes.


Splaktar:
Is there a setting where we can disable this in `angular.json` or something? It makes updating in real world apps (where you need to `ng update` 3-5+ times) quite onerous and results in being encouraged to commit changes that don't even build.

clydin:
The `--allow-dirty` option is available from 8.0.1 onward.

PR #14504, opened 2019-05-23
fix(@ngtools/webpack): update supported compiler-cli supported version

alan-agius4:
`@ngtools/webpack` only supports compiler version 8 as per the peerDependency.

The fixes the `isSupported` version static and also removes code that was to support prior versions of compiler-cli


tests/legacy-cli/e2e/tests/packages/webpack/test-app.ts
```diff
@@ -6,9 +6,9 @@ import {expectFileSizeToBeUnder, replaceInFile, expectFileToMatch} from '../../.
 
 export default function(skipCleaning: () => void) {
   return Promise.resolve()
-    .then(() => createProjectFromAsset('webpack/test-app-ng5'))
+    .then(() => createProjectFromAsset('webpack/test-app'))
     .then(() => exec(normalize('node_modules/.bin/webpack-cli')))
-    .then(() => expectFileSizeToBeUnder('dist/app.main.js', 375 * 1024))
+    .then(() => expectFileSizeToBeUnder('dist/app.main.js', 470 * 1024))
```
clydin: Why did this get larger?

alan-agius4: The framework version was change from 5 to 8.

PR #14508, opened 2019-05-23
Differentiate lib build from app build in reference doc

jbogarthyde:
Explain that `ng build` command invokes different builders for apps and libs, and that only two options apply to libs.
Issue https://github.com/angular/angular/issues/26569

This is a stop-gap doc fix - it points out that `ng build app-project` uses a different builder than `ng build lib-project`, and that most options doc'd here apply only to apps. This can keep readers from trying to apply options to libs, but we really need a separate doc page for the lib builder.

### DIFF
--- a/packages/angular/cli/commands/build-long.md
+++ b/packages/angular/cli/commands/build-long.md
@@ -1,14 +1,18 @@
-Uses the [webpack](https://webpack.js.org/) build tool, with default configuration options specified in the workspace configuration file (`angular.json`) or with a named alternative configuration. 
-A "production" configuration is created by default when you use the CLI to create the project, and you can use that configuration by specifying the `--prod` option.
+The command can be used to build a project of type "application" or "library".
+When used to build a library, a different builder is invoked, and only the `ts-config`, `configuration`, and `watch` options are applied.
+All other options apply only to building applications.
+
+The application builder uses the [webpack](https://webpack.js.org/) build tool, with default configuration options specified in the workspace configuration file (`angular.json`) or with a named alternative configuration.
+A "production" configuration is created by default when you use the CLI to create the project, and you can use that configuration by specifying the `--configuration="production"` or the `--prod="true"` option.
 
 The configuration options generally correspond to the command options.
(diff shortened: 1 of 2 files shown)


packages/angular/cli/commands/build-long.md
```diff
@@ -1,14 +1,18 @@
-Uses the [webpack](https://webpack.js.org/) build tool, with default configuration options specified in the workspace configuration file (`angular.json`) or with a named alternative configuration. 
-A "production" configuration is created by default when you use the CLI to create the project, and you can use that configuration by specifying the `--prod` option.
+The command can be used to build a project of type "application" or "library".
+When used to build a library, a different builder is invoked, and only the `tsconfig` and `watch` options are applied.
```
alan-agius4: `tsconfig` should be `ts-config`.

Users can also use `configuration`

packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -23,7 +23,7 @@
     },
     "tsConfig": {
       "type": "string",
-      "description": "The full path for the TypeScript configuration file, relative to the current workspace."
+      "description": "The full path for the TypeScript configuration file, relative to the current workspace. Applies to both libraries and applications."
```
alan-agius4: As discussed on slack, these descriptions here are for the application specific builder. 

This builder is not able to build a library.

jbogarthyde:
@alan-agius4 Please approve if you're ok with this.

jbogarthyde:
@alan-agius4 can you approve?

packages/angular_devkit/build_ng_packagr/src/build/schema.json
```diff
@@ -10,11 +10,11 @@
     },
     "tsConfig": {
       "type": "string",
```
alan-agius4: The changes in this files are still here. This builder is library specific.

jbogarthyde:
@alan-agius4 Requires approval

PR #14530, opened 2019-05-28
fix(@angular/cli): ng generate --help shows the wrong collection

### LINKED ISSUE #14519: ng g --list shows the wrong collection
# 🐞 Bug report

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

I don't know but the code makes it seem like it worked once upon a time

### Description

Running `ng g --list` in a workspace which has a different default schematic than `@schematics/angular` prints out:

```
[--help stuff]

Available Schematics:
  Collection "@schematics/angular":
    library
    workspace-schematic
```

The schematics listed are indeed the schematics available in the default collection but the default collection is actually `@nrwl/workspace`.

It looks like the source code used to or intends to print out available schematics for all collections. I think this would be the preferred behavior.

## 🔬 Minimal Reproduction
```
npx create-nx-workspace@next ng-bug
cd ng-bug
ng g --list
```

## 🌍 Your Environment
CLI `8.0.0-rc.4` 

**Anything else relevant?**
This affects all Nx users + more?

alan-agius4:
At the moment, collectionName and schematicCollections are not set in various schematics command which result in fallbacking to the hardcoded default collectionName https://github.com/angular/angular-cli/blob/master/packages/angular/cli/models/schematic-command.ts#L79

Hence, this will result in incorrect information being present when using the `--help`, `--list`.

Fixes #14519

### DIFF
--- a/packages/angular/cli/commands/generate-impl.ts
+++ b/packages/angular/cli/commands/generate-impl.ts
@@ -6,7 +6,6 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-// tslint:disable:no-global-tslint-disable no-any
 import { terminal } from '@angular-devkit/core';
 import { Arguments, SubCommandDescription } from '../models/interface';
 import { SchematicCommand } from '../models/schematic-command';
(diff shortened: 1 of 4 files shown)


packages/angular/cli/commands/update-impl.ts
```diff
@@ -29,6 +29,7 @@ const oldConfigFileNames = [
 
 export class UpdateCommand extends SchematicCommand<UpdateCommandSchema> {
   public readonly allowMissingWorkspace = true;
+  protected collectionName = '@schematics/update';
```
clydin: This is intentionally not present.  The command itself handles the options.  It's only a schematic command because there's no other mechanism currently to easily execute a schematic.

alan-agius4: Removed, PTAL.

PR #14537, opened 2019-05-28
feat(@schematics/angular): default Ivy apps to AOT true

filipesilva:
Testing on AIO with Angular master as of 28/05/2019 I got these results:
JIT ~414ms (369, 378, 408, 323, 593)
AOT using VE ~1383ms (1365, 1185, 1767, 1135, 1467)
AOT using Ivy ~517ms (600, 391, 444, 756, 394)

This is largely due to https://github.com/angular/angular/pull/29380 and https://github.com/angular/angular/pull/30238.

The second PR above was not merged to master, and thus will not be in 8.0.0. This PR should be merged to match it.

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -194,6 +194,7 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
           main: `${sourceRoot}/main.ts`,
           polyfills: `${sourceRoot}/polyfills.ts`,
           tsConfig: `${projectRoot}tsconfig.app.json`,
+          aot: !!options.enableIvy,
           assets: [
             `${sourceRoot}/favicon.ico`,
             `${sourceRoot}/assets`,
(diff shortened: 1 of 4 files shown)


alexeagle:
Your comment "This PR should be merged to match it." still makes me hesitate about including this change in CLI 8.0.1 when used with Angular 8.0.0.

Going to take the low-risk option of not including it in today's release...

filipesilva:
@alexeagle I think it should go in the CLI 8.0.x timeframe because of the currently live documentation in https://angular.io/guide/ivy.

It directs users making a new CLI project that they do not need to do anything else but use `--enable-ivy`. New projects do not have AOT set to true without this PR.

But for users that want to migrate an app, it directs them to set AOT:

> AOT compilation with Ivy is faster and should be used by default. In the angular.json workspace configuration file, set the default build options for your project to always use AOT compilation.

To be fair the way it is all worded sets the expectation of `ng new shiny-ivy-app --enable-ivy` creating a project that already uses at least `@angular/core@8.1.x`. Maybe the correct thing is to also change that. WDYT?

alexeagle:
merging since angular 8.1.0-next.1 is out

Issue #8986, opened 2017-12-24
ng serve documentation (specifically --public-host/live-reload-client)


dtabaka:
### Versions
```
Node 6.11.2, Npm 5.6.0, Angular CLI 1.5.0, Angular 5, Asp.NET Core 2.0, Visual Studio 2017
```
### Repro steps

I'm integrating the Angular CLI with Visual Studio 2017 & an ASP.NET Core 2.0 application where I successfully have the live reloading working when I make a change to my Angular components.  I have the following script tags in my _layout.cshtml during development.

  <script type="text/javascript" src="http://localhost:4200/inline.bundle.js"></script>
etc.

### Desired behavior
However, for this to work I stumbled upon a post that mentioned I had to configure my package.json "start" script as follows:

 "scripts": {
    "start": "ng serve  --live-reload-client http://localhost:4200/",

I would like:
1.) To better understand this and what it is doing behind the scenes.  From what I understand **ng serve** uses webpack dev server under the covers so I'm wondering how this configures it or what the equivalent setting in the webpack.config.json file would be.

2.) ng serve docs have this listed as **public-host** (aka **live-client-reload**) but not much more than that.  If there was more information it might help clear up 1.)

Splaktar:
There are docs now in https://angular.io/cli/serve, but it's still not clear what `--public-host` really does and there are no examples of how/when this would be useful.

All it says is "Specify the URL that the browser client will use."

I came to the docs trying to fix an issue with live reload on Chrome OS. Not mentioning live reload and public host are related cost me a good bit of time and frustration.

@jbogarthyde @jenniferfell this docs issue is closer to being addressed now. I think a little bit of refinement to the existing `ng serve` docs should solve this.

Splaktar:
This appears to be resolved in `7.3.0` via https://github.com/angular/angular-cli/commit/4729397a680010ceb79b8cdfea96295a74e16715.

Splaktar:
But it appears that previous PR only changed one entry for this, but there are multiple. https://github.com/angular/angular-cli/pull/14538 is working to address this for the dev-server builder.

The [documentation for this on AIO](https://next.angular.io/cli/serve) has not been updated or improved, so this issue is not yet resolved.

Splaktar:
[...]

PR #14538, opened 2019-05-28
Update doc to make connection of live-reload and public-host explicit

jbogarthyde:
Document the relationship of `ng serve` options `--live-reload` and `--public-host`.
Address Issue https://github.com/angular/angular-cli/issues/8986

### DIFF
--- a/packages/angular_devkit/build_angular/src/dev-server/schema.json
+++ b/packages/angular_devkit/build_angular/src/dev-server/schema.json
@@ -49,12 +49,12 @@
     },
     "liveReload": {
       "type": "boolean",
-      "description": "Whether to reload the page on change, using live-reload.",
+      "description": "Whether to reload the page on change, using live-reload with the public host.",
       "default": true
     },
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/dev-server/schema.json
```diff
@@ -49,12 +49,12 @@
     },
     "liveReload": {
       "type": "boolean",
-      "description": "Whether to reload the page on change, using live-reload.",
+      "description": "Whether to reload the page on change, using live-reload with the public host.",
       "default": true
     },
     "publicHost": {
       "type": "string",
-      "description": "Specify the URL that the browser client will use."
```
Splaktar: It seems like this should be the same as https://github.com/jbogarthyde/angular-cli/blob/master/packages/angular/cli/lib/config/schema.json#L1098-L1101

alan-agius4:
Caretaker: kindly squash commits.

PR #14539, opened 2019-05-28
fix(@schematics/angular): add migration to update `web-animations-js`…

alan-agius4:
… polyfill

Previous versions rely on `this` which breaks differential loading since it's `undefined` when using script tags with type module.


packages/schematics/angular/migrations/update-8/update-dependencies.ts
```diff
@@ -49,5 +49,19 @@ export function updateBuilders() {
         },
       );
     }
+
+    const webAnimationsJsVersion = '^2.3.2';
```
alan-agius4: @mgechev, if the release that they cut is not this version, we need to update this value.

alan-agius4:
Blocked on https://github.com/web-animations/web-animations-js/pull/208

alexeagle:
unblocked this by publishing a release with Minko's fix to the angular fork of the project.

PR #14560, opened 2019-05-29
ci: add Apache to the licenses list

alan-agius4:
Related failure https://circleci.com/gh/angular/angular-cli/61130?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

### DIFF
--- a/scripts/validate-licenses.ts
+++ b/scripts/validate-licenses.ts
@@ -29,6 +29,7 @@ const licensesWhitelist = [
   // Regular valid open source licenses supported by Google.
   'MIT',
   'ISC',
+  'Apache',
   'Apache-2.0',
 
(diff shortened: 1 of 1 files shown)


scripts/validate-licenses.ts
```diff
@@ -29,6 +29,7 @@ const licensesWhitelist = [
   // Regular valid open source licenses supported by Google.
   'MIT',
   'ISC',
+  'Apache',
```
clydin: That’s not a valid SPDX identifier.  There are three versions of the Apache license.

Can we fix the package instead?

alan-agius4: I'll try to do that, Probably they are all in the same mono repo

PR #14596, opened 2019-05-30
fix(@angular/cli): provide an option to update dirty repositories

clydin:
Automated use of the update command may necessitate usage in a dirty repository.  This enables that use case.

//cc @FrozenPandaz

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -83,12 +83,19 @@ export class UpdateCommand extends SchematicCommand<UpdateCommandSchema> {
 
     // If not asking for status then check for a clean git repository.
     // This allows the user to easily reset any changes from the update.
-    if ((packages.length !== 0 || options.all) && !this.checkCleanGit()) {
-      this.logger.error(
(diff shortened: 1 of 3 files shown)


packages/angular/cli/commands/update-impl.ts
```diff
@@ -83,12 +83,19 @@ export class UpdateCommand extends SchematicCommand<UpdateCommandSchema> {
...
-      );
+    const statusCheck = packages.length === 0 && !options.all;
+    if (!statusCheck && !this.checkCleanGit()) {
+      if (options.allowDirty) {
+        this.logger.warn(
+          'Repository is not clean.  Update changes will be mixed with pre-existing changes.',
+        );
+      } else {
+        this.logger.error(
+          'Repository is not clean.  Please commit or stash any changes before updating.',
```
alexeagle: shouldn't this message also suggest
`or run with --allowDirty to make edits to the current state`

clydin: They should ideally commit or stash rather than ignore the dirty state

PR #14605, opened 2019-05-31
(fix) error when project name is dir

Eusen:
sometimes, I want named a project or library like "@superme/angular", it work, but angular.json report an schema error: "@superme/angular is not allowed". So I suggest changing pattern properties to \\w* to match any words.


Eusen:
image
These errors do not seem to be directly related to what I submitted.

vikerman:
Thanks for the PR!

We should change the regex but not to \\w which is too lenient. Maybe we should use the regex used by npm to validate scoped packages.

PR #14606, opened 2019-05-31
fix(@schematics/angular): update `ng-packagr` and `tsickle` when migr…

alan-agius4:
…ating to version 8

### DIFF
--- a/packages/schematics/angular/migrations/update-8/update-dependencies.ts
+++ b/packages/schematics/angular/migrations/update-8/update-dependencies.ts
@@ -11,55 +11,29 @@ import { latestVersions } from '../../utility/latest-versions';
 
 export function updateDependencies() {
   return (host: Tree) => {
-    let current = getPackageJsonDependency(host, '@angular-devkit/build-angular');
-    if (current && current.version !== latestVersions.DevkitBuildAngular) {
-      addPackageJsonDependency(
-        host,
-        {
-          type: current.type,
-          name: '@angular-devkit/build-angular',
-          version: latestVersions.DevkitBuildAngular,
-          overwrite: true,
-        },
-      );
-    }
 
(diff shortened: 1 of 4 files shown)


packages/schematics/angular/migrations/update-8/update-dependencies.ts
```diff
@@ -11,55 +11,29 @@ import { latestVersions } from '../../utility/latest-versions';
...
-          type: current.type,
-          name: '@angular-devkit/build-ng-packagr',
-          version: latestVersions.DevkitBuildNgPackagr,
-          overwrite: true,
-        },
-      );
-    }
+    const dependenciesToUpdate: Record<string, string> = {
+      '@angular-devkit/build-angular': latestVersions.DevkitBuildAngular,
+      '@angular-devkit/build-ng-packagr': latestVersions.DevkitBuildNgPackagr,
```
clydin: can you also add `'@angular-devkit/build-webpack'` as well?

alan-agius4: Added

packages/schematics/angular/utility/latest-versions.ts
```diff
@@ -16,4 +16,8 @@ export const latestVersions = {
   // The versions below must be manually updated when making a new devkit release.
   DevkitBuildAngular: '~0.800.0-beta.13',
   DevkitBuildNgPackagr: '~0.800.0-beta.13',
+  DevkitBuildWebpack: '~0.800.0-beta.13',
```
clydin: The beta specifier shouldn’t be needed anymore.

alan-agius4: Oops, PTAL

PR #14607, opened 2019-05-31
refactor: deprecate `isSupported` static in `AngularCompilerPlugin`

alan-agius4:


mgechev:
Looking at this right now.

PR #14611, opened 2019-05-31
refactor(@schematics/angular): remove non-useful or operator

ahasall:
There is already the following code that sets `options.directory` to `options.name` if no directory is passed as an option.
```typescript
if (!options.directory) {
    options.directory = options.name;
}
```

So the `options.directory || options.name` is unnecessary


alan-agius4:
@ahasall, can you please rebase to make this green?

ahasall:
@alan-agius4 the branch is update

ahasall:
Or do you want me to rebase just so that the build can trigger again ?

PR #14622, opened 2019-06-01
fix(@angular/cli): resolve update migrations from referenced package root

clydin:
This ensures that migration fields that reference other packages use the package version specified in the migrating package's dependencies and not the version that a package manager happens to hoist to the root of the workspace.

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import { execSync } from 'child_process';
+import * as fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
 import { Arguments, Option } from '../models/interface';
@@ -22,10 +23,7 @@ import { Schema as UpdateCommandSchema } from './update';
 
 const npa = require('npm-package-arg');
 
-const oldConfigFileNames = [
-  '.angular-cli.json',
-  'angular-cli.json',
-];
+const oldConfigFileNames = ['.angular-cli.json', 'angular-cli.json'];
 
 export class UpdateCommand extends SchematicCommand<UpdateCommandSchema> {
   public readonly allowMissingWorkspace = true;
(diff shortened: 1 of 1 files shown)


packages/angular/cli/commands/update-impl.ts
```diff
@@ -177,12 +175,45 @@ export class UpdateCommand extends SchematicCommand<UpdateCommandSchema> {
       } else if (typeof migrations !== 'string') {
         this.logger.error('Package contains a malformed migrations field.');
 
+        return 1;
+      } else if (path.posix.isAbsolute(migrations) || path.win32.isAbsolute(migrations)) {
```
alan-agius4: Why not `path.isAbsolute`? in windows `/foo/abc` is still considered as absolute path.

packages/angular/cli/commands/update-impl.ts
```diff
@@ -177,12 +175,45 @@ export class UpdateCommand extends SchematicCommand<UpdateCommandSchema> {
...
+      if (fs.existsSync(localMigrations)) {
+        migrations = localMigrations;
+      } else {
+        // Try to resolve from package location.
+        // This avoids issues with package hoisting.
+        try {
+          migrations = require.resolve(migrations, { paths: [packageNode.path] });
+        } catch (e) {
+          if (e.code === 'MODULE_NOT_FOUND') {
+            this.logger.error('Migrations for package were not found.');
```
alan-agius4: Maybe we should highlight the package name, especially considering that `ng update` can accept multiple packages.

clydin: You can only pass one package when using the migrate only option.

packages/angular/cli/commands/update-impl.ts
clydin: I wanted to ensure that it errors on all cases of an absolute path on all systems.  Windows has `C:`,`\\`, etc.

alan-agius4: image

Issue #13383, opened 2019-01-08
Unnecessary prompts for ng new when using create application false (i.e. Conditional Prompts) 


DenysVuika:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Command (mark with an `x`)
```
- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

CLI: 7.1.4

### Repro steps

Generating workspace without application still causes questions about the app:

```sh
ng new myapp --create-application=false
? Would you like to add Angular routing? No
? Which stylesheet format would you like to use? CSS

CREATE myapp/README.md (1027 bytes)
CREATE myapp/angular.json (135 bytes)
CREATE myapp/package.json (1262 bytes)
CREATE myapp/tsconfig.json (435 bytes)
CREATE myapp/tslint.json (2824 bytes)
CREATE myapp/.editorconfig (246 bytes)
CREATE myapp/.gitignore (576 bytes)
```

### Desired functionality

The CLI should not ask questions about application it is not going to generate anyway.

alan-agius4:
Hi, thanks for reporting this. 

`create-application` is not a known option, and actually this action should result in an error (
https://github.com/angular/angular-cli/issues/12549)

See https://angular.io/cli/generate#application-command for a full list of available options.

DenysVuika:
@alan-agius4 you can get the info via

```sh
ng new --help

--create-application 
    When true (the default), creates a new initial app project in the new workspace.
```

alan-agius4:
@DenysVuika, apologies.

Splaktar:
There are actually a range of different issues related to `--create-application=false`. It needs a good deal of cleanup.

For example, it's not expected that `ng new --create-application=false --styles=scss --routing=false` would accept the `styles` and `routing` options and then completely ignore them. They don't get added to the `angular.json`, so a subsequent `ng g app myapp` gets generated with `.css` files. 

If these options aren't compatible with `--create-application=false` (which I believe is the case since there is no spot to put them in `angular.json` for an empty workspace), then this should be documented and there should be a warning or error message when trying to use them together.

Here's an empty workspace `angular.json` for reference:
```js
{
[...]

PR #14638, opened 2019-06-02
fix(@schematics/angular): application schematic should prompt for --ro…

ahasall:
…uting and --style

refs #13383


ahasall:
For now, the tests are failing...because I still need to fix them :)

PR #14677, opened 2019-06-05
fix(@ngtools/webpack): always strip BOM when reading files

alan-agius4:
This is to align with the tsc readFile logic https://github.com/microsoft/TypeScript/blob/9da05243ff7ba9a2776d5a52118d680348b1454d/src/compiler/sys.ts#L1067

### DIFF
--- a/packages/ngtools/webpack/src/compiler_host.ts
+++ b/packages/ngtools/webpack/src/compiler_host.ts
@@ -147,13 +147,15 @@ export class WebpackCompilerHost implements ts.CompilerHost {
     const filePath = this.resolve(fileName);
 
     try {
+      let content: ArrayBuffer;
       if (this._memoryHost.isFile(filePath)) {
-        return virtualFs.fileBufferToString(this._memoryHost.read(filePath));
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/compiler_host.ts
```diff
@@ -147,28 +147,22 @@ export class WebpackCompilerHost implements ts.CompilerHost {
...
-    const filePath = this.resolve(fileName);
-
-    if (this._memoryHost.isFile(filePath)) {
-      return Buffer.from(this._memoryHost.read(filePath));
-    } else {
-      const content = this._syncHost.read(filePath);
-
-      return Buffer.from(content);
-    }
+    return Buffer.from(this.readFile(fileName) || '');
```
clydin: This is backwards.  The Buffer should be the raw data from the file.  The text content is based on this and may contain a _header_ (i.e., the BOM).  Nothing should be altered when getting the raw data.  It's completely valid for a binary file to contain the BOM values.

PR #14716, opened 2019-06-10
DL fixes for safari and false negatives for ES6 module support

### LINKED ISSUE #14580: Differential loading: Edge not considered ES2015 browser
# 🐞 Bug report

```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

Kind of, due to the introduction of differential loading.

### Description

Edge 15-18 are considered non-ES2015 browsers when compiling the app. If you target only Edge 18 in `browserslist`, I'd expect to only generate the ES2015 code (similar to what happens when only targeting the latest Firefox/Chrome release); but ES5 code is also generated. This is specially worrying because, due to a Edge bug, when running the app, the ES5 bundle is downloaded (but not executed), and the ES2015 bundle is downloaded twice (reference https://github.com/philipwalton/webpack-esnext-boilerplate/issues/1#issuecomment-445280691 ).

The result is that, while in Angular 7 I already compiled for ES2015 giving Edge users a optimized bundle, in Angular8 they are going to download more than 3 times the previous size to end up running the same ES2015 bundle.

## 🔬 Minimal Reproduction
Edit browserslist file with only this content:
```
Edge 18
```
And run `ng build --prod`

## 🔥 Exception or Error
Both ES5 and ES2015 are generated.

## 🌍 Your Environment
<pre>
Angular CLI: 8.0.0
Node: 12.2.0
OS: darwin x64
Angular: 8.0.0
... forms, language-service, material, platform-browser

@angular-devkit/architect         0.800.0
@angular-devkit/build-angular     0.800.0
@angular-devkit/build-optimizer   0.800.0
@angular-devkit/build-webpack     0.800.0
@angular-devkit/core              8.0.0
@angular-devkit/schematics        8.0.0
@angular/flex-layout              8.0.0-beta.26
@angular/pwa                      0.800.0
@ngtools/webpack                  8.0.0
@schematics/angular               8.0.0
@schematics/update                0.800.0
rxjs                              6.5.2
typescript                        3.4.5
webpack                           4.30.0
</pre>

**Anything else relevant?**

### LINKED ISSUE #14680: Attribute noModule is ignored by Safari 10
# 🐞 bug report

### Affected Package

 The issue is caused by the use of the `nomodule` attribute in the `<script>` tag.

### Is this a regression?

 Yes, it was working in Angular 7

### Description

 
The build generated by Angular v8 with both `es5` and `es2015` flavor cannot be loaded on Safari 10. It's caused by the fact that Safari doesn't seems to acknownledge the `nomodule` attribute and so, load both versions which leads to an error:

```
Zone.js has detected that ZoneAwarePromise `(window|global).Promise` has been overwritten. Most likely cause is that a Promise polyfill has been loaded after Zone.js (Polyfilling Promise api is not necessary when zone.js is loaded. If you must load one, do so before loading zone.js.)
```

## 🔬 Minimal Reproduction

You can see a simple (no config) repo [here](https://github.com/Embraser01/ng-safari-bug) and [here](https://quizzical-joliot-50ad33.netlify.com/), the same site deployed.

## 🔥 Exception or Error

You can see in the console 
<pre><code>
Zone.js has detected that ZoneAwarePromise `(window|global).Promise` has been overwritten. Most likely cause is that a Promise polyfill has been loaded after Zone.js (Polyfilling Promise api is not necessary when zone.js is loaded. If you must load one, do so before loading zone.js.)
</code></pre>

## 🌍  Your Environment

**Angular Version:** v8.0.0
<pre><code>

                |___/
    

Angular CLI: 8.0.1
Node: 10.15.0
OS: win32 x64
Angular: 8.0.0
... router

@angular-devkit/architect         0.800.1
@angular-devkit/build-angular     0.800.1
@angular-devkit/build-optimizer   0.800.1
@angular-devkit/build-webpack     0.800.1
@angular-devkit/core              8.0.1
@angular-devkit/schematics        8.0.1
@angular/cli                      8.0.1
@ngtools/webpack                  8.0.1
@schematics/angular               8.0.1
@schematics/update                0.800.1
rxjs                              6.4.0
typescript                        3.4.5
webpack                           4.30.0
</code></pre>

**Anything else relevant?**

It seems to be related to Safari ignoring `nomodule` attribute.
It affect at least Safari 10.0.0 (on iOS 10.3).
I'll test on more devices but it takes some time (I test on simulator).

I also found [this](https://gist.github.com/samthor/64b114e4a4f539915a95b91ffd340acc/)

alan-agius4:
**fix(@angular-devkit/build-angular): nomodule polyfill for Safari 10.1 and iOS Safari 10.3**

The nomodule polyfill needs to be inject prior to any script and be
outside of webpack compilation because otherwise webpack will cause the
script to be wrapped in `window["webpackJsonp"]` which causes this to
fail.

Also, this ensures that it's loaded and executed prior to any other
nomodule script.

This polyfill will only be injected when the either Safari 10.1 or iOS
Safari 10.3 support is required, which is based on the browsers  defined in browserslist file.

Fixes #14680

**fix(@angular-devkit/build-angular): browsers that partially support es6 modules are being marked as not supported**

`caniuse-api`  only returns `true` for a feature when it's fully supported. This might causes redundant ES5 builds if users opt to support only browsers which do support ES6 modules but either require polyfills such as Safari 10.1 or when both bundles will be downloaded such as `Edge 18`

See: https://github.com/Nyalab/caniuse-api/blob/1b74c10257e36877709da4f6630513db10203a78/src/index.js#L49 and Nyalab/caniuse-api#82

Fixes #14580


packages/angular_devkit/build_angular/src/angular-cli-files/utilities/index-file/augment-index-html.ts
```diff
@@ -246,3 +265,38 @@ function _generateSriAttributes(content: string) {
...
+ * Again: this will **not** prevent inline script, e.g.:
+ * <script nomodule>alert('no modules');</script>.
+ *
+ * This workaround is possible because Safari supports the non-standard 'beforeload' event.
+ * This allows us to trap the module and nomodule load.
+ *
+ * Note also that `nomodule` is supported in later versions of Safari - it's just 10.1 that
+ * omits this attribute.
+ */
+const _noModulePolyfill = tags.oneLine`
```
clydin: Inline scripts are not a recommended approach from a security perspective.  They also increase the complexity of CSP configurations.  Can you switch this to a file based script?

alan-agius4: As discussed offline, there might be a way, by leveraging the ScriptsPlugins. Will give it a shot tomorrow.

alan-agius4:
Patch version of this PR is available here: https://github.com/angular/angular-cli/pull/14730

PR #14728, opened 2019-06-11
test: fix flakes of `rebuilds TS worker`

alan-agius4:
Not quite sure why such a long 'debounceTime' is needed anything under `2500` is a constant failure locally and this is also rather flaky on CI when it doesn't run as the first test.

It seems that the outputted files contents don't get updated in time.


filipesilva:
It also doesn't make sense to me that the debounce time is needed. IIRC the webworker plugin works with child compilations for webpack. That means that everything should be written to disk at the same time.

PR #14730, opened 2019-06-11
DL fixes for safari and false negatives for ES6 module support (PATCH)

alan-agius4:
Patch version of https://github.com/angular/angular-cli/pull/14716 as it doesn't apply cleanly


packages/angular_devkit/build_angular/src/utils/build-browser-features.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * True, when a browser feature is supported partially or fully.
+   */
+  isFeatureSupported(featureId: string): boolean {
+    // y: feature is fully available
+    // n: feature is unavailable
+    // a: feature is partially supported
+    // x: feature is prefixed
+    const criterias = [
```
clydin: Nit: criteria is already plural

alan-agius4: updated in both branches

PR #14731, opened 2019-06-11
 feat(@schematics/angular): next option for ng new

atiris:
As with `ng update`, a `ng new` should support the creation of a workspace with the next version of Angular. This is useful, for example, for testing, and now, when creating a new IVY project (using next packages of Angular is recommended with IVY)

_This is a small feature._


alan-agius4:
Note: this doesn't mean that it will always work, as `next` might have different requirements in terms of other dependencies such `rxjs` and `typescript`. 

Also the project scaffolding might not be fully up-to-date as the schematic used to generate the `next` project will be based on an older version of the `@schematics/angular`.

atiris:
Ok, thanks for the check. :+1: I only assumed a case if someone wants to use the latest versions of Angular (like now with IVY) so I assumed it would be a small change, of course if you want a more robust solution (or this option is not so beneficial to have separated switch in CLI) subsequent discussion will be a good continuation :wink:

alan-agius4:
Note there is also this an alternative to the above, if you don’t want to update the global CLI but use a next version.

```
npx @angular/cli@next new —-enable-ivy

**Comment by atiris** [comment]
Well, there is always an alternative. The question is whether it is more beneficial for programmers (me :stuck_out_tongue:). Anyway I tested your suggestion:

```bash
# using npx as alternative
npx @angular/cli@next new angular-npx --defaults --skip-git --enable-ivy
cd angular-npx
ng --version

Angular CLI: 8.0.2
Node: 11.15.0
OS: linux x64
Angular: 8.0.0
...

# package.json pick
"dependencies": {
    "@angular/core": "~8.0.0", ...
```

```bash
# using option in ng new schematics
bin/ng new angular-ng --defaults --next --skip-git --enable-ivy
cd angular-ng
ng --version

Angular CLI: 8.0.0
Node: 11.15.0
OS: linux x64
Angular: 8.1.0-next.1
...

# package.json pick
"dependencies": {
    "@angular/core": "next", ...
```

Only the second case is what I want to achieve. What am I doing wrong?

alan-agius4:
You are actually not doing anything wrong, it's just that at the moment the `next` version of the CLI is not yet published.

atiris:
:smiley: ok, in that case, now I'm more inclined to use `--next` option in `ng new` :wink:

clydin:
However, to properly implement this, the `ng new` process would need to be completely reimagined to pull the `next` version (or `latest` depending on the inclusion of this option) of the Angular schematics and use those to generate a new project.  Without this capability, the generated project may not (and most likely will not) be setup to property use of the next version of the packages.  While work is planned in this area, the design has not yet reached a juncture wherein an option such as this would be viable.  Once that point is reached this is a definitely a feature which could be supported.

PR #14733, opened 2019-06-11
refactor: ignore vscode history folder in git

atiris:
**Ignore .history/ directory in git.**

After using Visual Studio Code to work with this repository, many files remain in the .history directory that git show as "new" but all should be ignored. Based on this, a .gitignore file is updated.

_this is a small change_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ dist-schema/
 .idea/
 jsconfig.json
 .vscode/
+.history/
 
 # Typings file.
 typings/


clydin:
If using the [local-history](https://marketplace.visualstudio.com/items?itemName=xyz.local-history) plugin, this issue may be of interest: https://github.com/zabel-xyz/local-history/issues/26

There are instructions to put the history data in the `.vscode` directory to avoid adding to every `.gitignore` file.

atiris:
Yes, you're right, now it looks more like I not set my local vscode environment optimally. :roll_eyes: (fixed in my workspace)
  
I don't know how many users know about this option in local history extension (unfortunately it's not the default behavior)... so if you think this is not a beneficial change, reject this PR peacefully :wink:  

Thanks for the reply and investigation :+1:

PR #14753, opened 2019-06-12
fix(@ngtools/webpack): implement getModifiedResourceFiles for Ivy AOT

clydin:
This method enables the Ivy incremental compiler to track changes to resource files and their requisites.

Related https://github.com/angular/angular/pull/30954


filipesilva:
To test this we should run our `yarn test-large` on Ivy, which we don't yet.

PR #14780, opened 2019-06-14
refactor(@angular/cli): standardize color handling and support checking

clydin:
Node.js 10+ provides built-in functionality to test for color support based on chalk's `supports-color` package as well as several others. This alleviates the need for custom code or third-party packages to determine color support.  In addition for this PR, the `ansi-colors` package is added to the CLI which provides color, cross-platform symbol, and style/color removal support.  The package is light-weight (~150 SLOC), contains typings, and has no dependencies. The removal support is leveraged to remove all styling from logger messages when color is not supported.  This removes a current defect in which color/styling is still displayed if generated manually or via methods that do not perform supportability checks.  This is very relevant for third-party code/packages.  Finally, the typically used console functions are overriden to leverage the logger to ensure that the color processing is applied to third-party code (e.g., webpack internally as well as some of its loaders and plugins) which output to the console directly.

The color removal capabilities will be leveraged to support improved TypeScript diagnostic reporting which will use TypeScript's built-in functionality that while providing improved messaging and context also includes color.

### DIFF
--- a/packages/angular/cli/lib/cli/index.ts
+++ b/packages/angular/cli/lib/cli/index.ts
@@ -5,24 +5,37 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { terminal } from '@angular-devkit/core';
 import { createConsoleLogger } from '@angular-devkit/core/node';
+import * as colors from 'ansi-colors';
+import { WriteStream } from 'tty';
+import { format } from 'util';
 import { runCommand } from '../../models/command-runner';
 import { getWorkspaceRaw } from '../../utilities/config';
 import { getWorkspaceDetails } from '../../utilities/project';
 
-
 export default async function(options: { testing?: boolean, cliArgs: string[] }) {
+  // Typings do not contain the function call (added in Node.js v9.9.0)
+  const supportsColor = process.stdout instanceof WriteStream &&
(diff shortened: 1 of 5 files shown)


packages/angular/cli/lib/cli/index.ts
```diff
@@ -5,25 +5,38 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { terminal } from '@angular-devkit/core';
 import { createConsoleLogger } from '@angular-devkit/core/node';
+import * as colors from 'ansi-colors';
```
alan-agius4: What is the reason for changing to use `ansi-color` instead of `@angular-devkit/core`?

clydin: The style removal capability is needed to support future work.  The cross-platform symbol capabilities will also be useful for future improvements to user messaging ([code link](https://github.com/doowb/ansi-colors/blob/master/symbols.js)).  Since it's already being imported there isn't a compelling reason to import another module to also provide color support if one happens to already be present.
It also doesn't have the bugs that are present currently.  Try the following for instance:
```
const { terminal } = require('@angular-devkit/core');
console.log(terminal.bold(`foo ${terminal.red(terminal.dim('bar'))} baz`));
```
Versus:
```
const color = require('ansi-colors');
console.log(color.bold(`foo ${color.red(color.dim('bar'))} baz`));
```
image

Issue #12365, opened 2018-09-25
As a CLI user, I want the doc command to guide me to the version of Angular I'm using


hansl:
Right now the `ng doc` command uses the latest version of the doc. It would be nice to (acceptance criteria):

1. provide a flag to specify which version to use, and
2. by default, use the version of angular core that's installed in the project.

This way people can use the doc command with the version of angular.io they care about.

cexbrayat:
This issue was partially fixed by #14788 as it's now possible to use:

```
ng doc --version 6 // opens v6.angular.io
ng doc --version next // opens next.angular.io
```

The second part of the issue requires to read the Angular version installed in the project.
@alan-agius4 @clydin What would you recommend to use to grab the Angular version in the `doc` command?

mgechev:
`require('@angular/core').VERSION` seems the most accurate way to me.

clydin:
I agree.  Wrap that in a try catch and use a split on the periods to extract the major version. Use the default domain if an exception occurs.

cexbrayat:
Thanks for the feedback!

@LakhyariMs Would you like to give it a try and open a new PR with this? Or do you prefer that someone else does it?

PR #14788, opened 2019-06-15
feat(@angular/cli): ng doc accepts a version flag

LakhyariMs:
With this commit, we can now specify a `version`
for the `ng doc` command

    ng doc --version 6
    ng doc -v 6

and this will open `v6.angular.io` instead of `angular.io`.
The default domain is still used
if no version is specified.

Refs #12365

### DIFF
--- a/packages/angular/cli/commands/doc-impl.ts
+++ b/packages/angular/cli/commands/doc-impl.ts
@@ -19,9 +19,21 @@ export class DocCommand extends Command<DocCommandSchema> {
 
       return 0;
     }
-    let searchUrl = `https://angular.io/api?query=${options.keyword}`;
+
+    let domain = 'angular.io';
+
+    if (options.version) {
+      if (options.version == 'next') {
+        domain = 'next.angular.io';
+      } else {
+        domain = `v${options.version}.angular.io`;
+      }
+    }
+
+    let searchUrl = `https://${domain}/api?query=${options.keyword}`;
+
     if (options.search) {
-      searchUrl = `https://www.google.com/search?q=site%3Aangular.io+${options.keyword}`;
+      searchUrl = `https://www.google.com/search?q=site%3A${domain}+${options.keyword}`;
     }
 
     // We should wrap `open` in a new Promise because `open` is already resolved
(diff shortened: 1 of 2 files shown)


cexbrayat:
Note to the team: this is a first time contributor, contributing to the CLI during the [HackCommitPush](http://hack-commit-pu.sh/) conference, which aims to help developers to contribute to open source projects :)

cexbrayat:
@clydin What do you think about:
```json
"version" : {
  "aliases": ["v"],
  "anyOf": [
    {
      "type": "integer",
      "minimum": 2
    },
    {
      "const": "next"
    }
  ],
  "description": "Contains the version of Angular to use for the documentation."
}
```
and then test for `3` in code (I don't think we can exclude 3 in the schema?)
Ideally, we should also check that the version is not greater than the latest released, but that might be trickier. So I guess the logic:

```typescript
if (options.version) {
  if (options.version === 'next') {
    domain =  'next.angular.io';
  } else if (+options.version !== 3) {
    domain = `v${options.version}.angular.io`;
  }
}
```
could be good enough for this first PR?

clydin:
Since 2 is the outlier, you could add it as a “const” in the anyOf and set the minimum in the integer section to 4.  Looks good though.

clydin:
Also, if interested in a follow up PR, another nice feature that builds on this would be to automatically use the version installed within the project.

cexbrayat:
@clydin Thanks for the feedback!
Yes we were thinking about it, but I was unsure about the best way to get the Angular version installed in the project. What would you recommend?

@LakhyariMs Do you think you can update the PR with the changes Charles is suggesting? 
Something like:
```
"version" : {
  "aliases": ["v"],
  "anyOf": [
    {
       "const": 2
    },
    {
      "type": "integer",
      "minimum": 4
    },
    {
      "const": "next"
    }
  ],
  "description": "Contains the version of Angular to use for the documentation."
}
```
and then something like:
```typescript
if (options.version) {
  if (options.version === 'next') {
    domain =  'next.angular.io';
  } else {
    domain = `v${options.version}.angular.io`;
  }
}
```

PR #14790, opened 2019-06-15
fix(@ngtools/webpack): ensure references for constructor parameter types

clydin:
By cloning the import specifier for each of the parameter types, TypeScript's import eliding algorithm will skip removal and prevent reference exceptions at runtime.

### DIFF
--- a/packages/ngtools/webpack/src/transformers/ctor-parameters.ts
+++ b/packages/ngtools/webpack/src/transformers/ctor-parameters.ts
@@ -202,8 +202,7 @@ export function decoratorDownlevelTransformer(
   diagnostics: ts.Diagnostic[],
 ): (context: ts.TransformationContext) => ts.Transformer<ts.SourceFile> {
   return (context: ts.TransformationContext) => {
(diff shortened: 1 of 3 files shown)


packages/ngtools/webpack/src/transformers/ctor-parameters.ts
```diff
@@ -202,49 +202,29 @@ export function decoratorDownlevelTransformer(
...
-        if (value) {
-          return value;
+        const typeSymbol = typeChecker.getSymbolAtLocation(name);
+        if (typeSymbol) {
+          parameterTypeSymbols.add(typeSymbol);
         }
 
-        // Otherwise this will be a locally declared name, just return that.
-        return name;
+        // Based on TS's strategy to allow the checker to reach this identifier
```
filipesilva: Do you think we might need to change this if TS's strategy changes? If so, maybe we should include some permalinks to the relevant TS source here.

clydin: Link added.

PR #14802, opened 2019-06-17
feat(@angular/cli): log unhandled exceptions

clydin:
Whenever an unhandled exception during command processing occurs, the message and stack trace will be saved to a debug log located in a temporary directory.  The user will be informed of the log location as well as the exception message and instructions to file a bug report at the GitHub repository.

### DIFF
--- a/packages/angular/cli/lib/cli/index.ts
+++ b/packages/angular/cli/lib/cli/index.ts
@@ -48,10 +48,30 @@ export default async function(options: { testing?: boolean, cliArgs: string[] })
     return 0;
   } catch (err) {
     if (err instanceof Error) {
-      logger.fatal(err.message);
-      if (err.stack) {
-        logger.fatal(err.stack);
+      try {
+        const fs = await import('fs');
+        const os = await import('os');
+        const tempDirectory = fs.mkdtempSync(fs.realpathSync(os.tmpdir()) + '/' + 'ng-');
+        const logPath = tempDirectory + '/angular-errors.log';
+        fs.appendFileSync(logPath, '[error] ' + (err.stack || err));
+
+        logger.fatal(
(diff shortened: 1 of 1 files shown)


packages/angular/cli/lib/cli/index.ts
```diff
@@ -48,10 +48,30 @@ export default async function(options: { testing?: boolean, cliArgs: string[] })
...
   } catch (err) {
     if (err instanceof Error) {
-      logger.fatal(err.message);
-      if (err.stack) {
-        logger.fatal(err.stack);
+      try {
+        const fs = await import('fs');
+        const os = await import('os');
+        const tempDirectory = fs.mkdtempSync(fs.realpathSync(os.tmpdir()) + '/' + 'ng-');
+        const logPath = tempDirectory + '/angular-debug.log';
```
alan-agius4: Any reason why the file name is named `-debug` Instead of `-errors`?

clydin: not particularly; changed it to `errors`.

packages/angular/cli/lib/cli/index.ts
```diff
@@ -48,10 +48,30 @@ export default async function(options: { testing?: boolean, cliArgs: string[] })
     return 0;
   } catch (err) {
     if (err instanceof Error) {
-      logger.fatal(err.message);
-      if (err.stack) {
-        logger.fatal(err.stack);
+      try {
+        const fs = await import('fs');
+        const os = await import('os');
+        const tempDirectory = fs.mkdtempSync(fs.realpathSync(os.tmpdir()) + '/' + 'ng-');
```
alan-agius4: You were too fast to push 😜 missed the slack convo.

Nit: Shouldn’t this be `ng-cli` or `ng` instead of `ng-`?

PR #14817, opened 2019-06-18
fix(@schematics/angular): remove indentation in component html 

### LINKED ISSUE #14816: Html file without indentation
# 🚀 Feature request

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
The HTML files for angular-cli are now also generated using an indentation of two spaces.
tslint can be used to fix this for the typescript files to your own style.

the HTML template is however excluded from this.

Most editors, like PHPStorm have the feature: "On edit, use indents of the openend file" so removing it and correcting it won't fix it unless you close the file so that it no longer knows it was indented incorrectly. Or press the 'convert' button.

Since the template only contains dummy test content within a <p> tag, i'd suggest to remove the whitespace and indentation, which will resolve this issue for most users, for the file then does not have an active indentation. And the contents will be cleared completely afterwords anyways.

Maybe my flow of toughts is incorrect. But I do really think this will fix some issues for a lot of people.

Update the templates (as in pull request) to not use any indentation

I can disable the as-is feature of the editor, but that is not wanted.
I can overwrite the component schematic to replace the HTML contents afterwords, but still.

SamanthaAdrichem:
Most editors remember the indentation, even if you remove all code. This way, there is no
indentation thus there is nothing for the editor to remember and there will be no need for
a HTML lint fix.

Fixes #14816

### DIFF
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.component.html.template
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.component.html.template
@@ -1,3 +1,1 @@
-<p>
-  <%= dasherize(name) %> works!
-</p>
+<p><%= dasherize(name) %> works!</p>


SamanthaAdrichem:
I don't understand why the unit test fails. The unit test does nothing with the HTML file. And it worked before i changed the commit message.

https://circleci.com/gh/angular/angular-cli/65211

I haven't changed anything else.

Any suggestions?

alan-agius4:
@SamanthaAdrichem, it's a flake.

I'll restart the CI step

alan-agius4:
Thanks for your contribution @SamanthaAdrichem :)

Issue #14621, opened 2019-05-31
emitWarnings coverage threshold breaks architect build


StephenFluin:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [x] test
```

### Is this a regression?
Yes

### Description
Adding `emitWarning` to istanbul reporter breaks the build

## 🔬 Minimal Reproduction
ng new my-app (works)
add a threshold to `karma.conf.js: (works)
```
thresholds: {
        statements: 80
      }
```

set `emitWarning` to false

```
thresholds: {
        emitWarning: false,
        statements: 80
      }
```
`ng test --code-coverage` // fails

## 🔥 Exception or Error
<pre><code>
31 05 2019 14:57:30.903:ERROR [karma-server]: TypeError: Cannot read property 'pct' of undefined
    at /usr/local/google/home/stephenfluin/workspace/test/coverage/node_modules/karma-coverage-istanbul-reporter/src/reporter.js:10:35
    at Array.forEach (<anonymous>)
    at checkThresholds (/usr/local/google/home/stephenfluin/workspace/test/coverage/node_modules/karma-coverage-istanbul-reporter/src/reporter.js:9:27)
    at createReport (/usr/local/google/home/stephenfluin/workspace/test/coverage/node_modules/karma-coverage-istanbul-reporter/src/reporter.js:157:31)
    at /usr/local/google/home/stephenfluin/workspace/test/coverage/node_modules/karma-coverage-istanbul-reporter/src/reporter.js:220:9
    at Array.forEach (<anonymous>)
    at BrowserCollection.forEach (/usr/local/google/home/stephenfluin/workspace/test/coverage/node_modules/karma/lib/browser_collection.js:84:26)
    at CoverageIstanbulReporter.onRunComplete (/usr/local/google/home/stephenfluin/workspace/test/coverage/node_modules/karma-coverage-istanbul-reporter/src/reporter.js:219:16)
    at Server.<anonymous> (/usr/local/google/home/stephenfluin/workspace/test/coverage/node_modules/karma/lib/events.js:40:26)
    at Server.emit (events.js:201:15)
    at Timeout._onTimeout (/usr/local/google/home/stephenfluin/workspace/test/coverage/node_modules/karma/lib/executor.js:53:22)
    at listOnTimeout (internal/timers.js:531:17)
    at processTimers (internal/timers.js:475:7)
Cannot read property 'success' of undefined
TypeError: Cannot read property 'success' of undefined
[...]

Issue #14118, opened 2019-04-10
ng test hangs on SIGTERM in architect-command.ts


rcollette:
# 🐞 Bug report

### Command (mark with an `x`)
```
- [ ] new
- [ ] build
- [ ] serve
- [x ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

Possibly as of 7.1 or 7.2 branch.  Was not seeing this behavior in CLI 1.2.6 but am seeing it in 7.3.7

### Description

I am running lint, test and build simultaneously using the npm concurrently package.  This has always worked in ngcli 1.2.6.   When one of the tasks fails, concurrently issues a SIGTERM to the other tasks (fail fast), and the other tasks should shut down gracefully.  In 7.3.7 this results in an error in `ng test`. 

What is making this issue bad for me, is that when running this in a docker build (node:10.15.3-alpine) on a gitlab runner, the docker build just hangs, possibly in an infinite loop.  The gitlab runner never releases.    

When I run the demo from macOS, I do get back to a command prompt.  I can't explain the discrepancy.

## 🔬 Minimal Reproduction
The basic reproduction steps are to use concurrently to run `ng test` and then `process.exit(1)` in another concurrently run task _after_ the tests have started to run.

Please run `npm run demo` in this reproduction.  
https://github.com/rcollette/ngcli-test-sigterm-demo

Keep in mind that the timeout in sigterm.js may need to be adjusted.   On my machine 10s was enough time to get the tests started but before they finish.

## 🔥 Exception or Error
Output from an actual application build.
<pre>
<code>
> concurrently --kill-others-on-fail "npm:lint" "npm:test:server" "npm:build:server" 

[test:server] 
[test:server] > embark3-ui@0.0.0 test:server /build
[test:server] > ng test --code-coverage --source-map --watch=false --progress=false
[test:server] 
[lint] 
[lint] > embark3-ui@0.0.0 lint /build
[lint] > ng lint --type-check
[lint] 
[build:server] 
[build:server] > embark3-ui@0.0.0 build:server /build
[build:server] > ng build --progress=false --prod --aot --source-map --es5-browser-support=false
[build:server] 
[lint] Linting "embark3-ui"...
[lint] 
[...]

PR #14828, opened 2019-06-19
fix(@angular-devkit/build-angular): karma cannot read property 'succe…

alan-agius4:
…ss' of undefined

When an error occures outside of the compilation, the observable gets completed without `success` being set. Which causes `Cannot read property 'success' of undefined`

Related to #14621 and might solve #14118

### DIFF
--- a/packages/angular_devkit/build_angular/src/karma/index.ts
+++ b/packages/angular_devkit/build_angular/src/karma/index.ts
@@ -8,7 +8,7 @@
 import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
 import { resolve } from 'path';
 import { Observable, from } from 'rxjs';
-import { switchMap } from 'rxjs/operators';
+import { defaultIfEmpty, switchMap } from 'rxjs/operators';
 import * as webpack from 'webpack';
 import {
   getCommonConfig,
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/karma/index.ts
```diff
@@ -127,7 +127,9 @@ export function execute(
           return karmaStart.then(() => karmaServerWithStop.stop());
         }
       };
-    })),
+    }).pipe(
```
clydin: This is already in a pipe call.  Should be able to put the default if empty in the same pipe.

alan-agius4: Good catch!

packages/angular_devkit/build_angular/src/karma/index.ts
```diff
@@ -8,7 +8,7 @@
 import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
 import { resolve } from 'path';
 import { Observable, from } from 'rxjs';
-import { switchMap } from 'rxjs/operators';
+import { switchMap, defaultIfEmpty } from 'rxjs/operators';
```
clydin: not alphabetical?

PR #14844, opened 2019-06-20
feat(@angular-devkit/build-angular): add crossorigin options

### LINKED ISSUE #14743: feat: Add support for crossorigin in differential loading
We are using a transparent authentication gateway from Azure (https://docs.microsoft.com/de-de/azure/app-service/overview-authentication-authorization) in front of our applications. It captures all requests and some 'new' tags like:

* <link rel="manifest" href="...">
* <script type="module" src="…">

need  the attribute `crossorigin="use-credentials"` to be working, as the 
It also seems, that Safari is a little bit stricter here, then chrome. Chrome just needs the attribute for the manifest, Safari for both.

If crossorigin is not provided, the browsers throw errors like:

```
https://login.microsoftonline.com/XXXXX/oauth2/authorize?response_type=code+id_token&redirect_uri=https%3A%2F%2Fapp.azurewebsites.net%2F.auth%2Flogin%2Faad%2Fcallback&client_id=XXXXX&scope=openid+profile+email&response_mode=form_post&resource=https%3A%2F%2Fgraph.windows.net&nonce=XXXXX&state=redir%3D%252Fmain-es2015.3959c9c93c3c451fefc1.js
denied by Cross-Origin Resource Sharing policy: Origin
https://app.azurewebsites.net is not allowed by Access-Control-Allow-Origin.
```

Some infos about the attribute:
<https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes>

We already did some tests:
* First we changed  "target": "es2015" to "target": "es5" in tsconfig.json
-> the application works

* Then we moved back to es2015 and added a npm build step to our pipeline (thanks to @manfredsteyer)

```
"build:prod": "ng build --prod && npm run build:replace",
"build:replace": "sed -i.bak 's/type=\"module\"/type=\"module\" crossorigin=\"use-credentials\"/g' dist/app/index.html",
```
-> the application works

Both workarounds are fine for us, but I think others would like to have an option to configure crossorigin for modules.

# 🚀 Feature request

```
- [X] build
```

### Description
Add a compiler option:
* p.E. moduleCrossorigin - off | '' | anonymous | use-credentials

Generated script tags in index.html should get an additional attribute if the new compilerOption is set:

If moduleCrossorigin is not set: `<script type="module" src="…">`
If moduleCrossorigin is '' or anonymous: `<script type="module" crossorigin="anonymous" src="…">`
If moduleCrossorigin is use-credentials: `<script type="module" crossorigin="use-credentials" src="…">`

Described manual replacements via npm scripts above.

alan-agius4:
This options allows to define the crossorigin attribute setting of elements that provide CORS support

By default we set the value to `anonymous`.

Closes #14743

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -890,6 +890,16 @@
             "webWorkerTsConfig": {
               "type": "string",
               "description": "TypeScript configuration for Web Worker modules."
+            },
+            "crossOrigin": {
+              "type": "string",
+              "description": "Define the crossorigin attribute setting of elements that provide CORS support.",
+              "default": "none",
+              "enum": [
+                "none",
+                "anonymous",
+                "use-credentials"
+              ]
             }
           },
           "additionalProperties": false,
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/index-html-webpack-plugin.ts
@@ -8,7 +8,11 @@
 import * as path from 'path';
 import { Compiler, compilation } from 'webpack';
 import { RawSource } from 'webpack-sources';
-import { FileInfo, augmentIndexHtml } from '../utilities/index-file/augment-index-html';
+import {
+  CrossOriginValue,
+  FileInfo,
+  augmentIndexHtml,
+} from '../utilities/index-file/augment-index-html';
 import { IndexHtmlTransform } from '../utilities/index-file/write-index-html';
 import { stripBom } from '../utilities/strip-bom';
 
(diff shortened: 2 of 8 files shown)


vikerman:
@alan-agius4  - Needs rebase

vikerman:
What is the documentation impact for this change?

Also are we adding crossorigin="anonymous" by default? Is that necessary?

alan-agius4:
Documentation in AIO will be auto generated when new options are added. ( from JSON Help)

With regards if we should always add it or not maybe we should discuss it further, and see if we should add a `none` option, as truth to be told it’s not really nesserary.

clydin:
Of note is that `crossorigin="anonymous"` on a script element only has an effect on error reporting within the browser. (ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script)

PR #14846, opened 2019-06-20
fix(@schematics/angular): conserve canonical doctype declaration case

oboukli:
Use the HTML5 doctype declaration as defined by W3C, as is, conserving its case.

https://dev.w3.org/html5/html-author/#doctype-declaration
https://developer.mozilla.org/en-US/docs/Glossary/Doctype

### DIFF
--- a/etc/cli.angular.io/index.html
+++ b/etc/cli.angular.io/index.html
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="">
 <head>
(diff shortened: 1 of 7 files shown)


mgechev:
Thanks for the contribution!

I believe we can keep the lowercase doctype since this string is [case-insensitive](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype).

I'll close the PR for now. Please, let me know if I'm missing anything.

PR #14847, opened 2019-06-20
fix(@angular-devkit/build-angular): polyfill URL to support $location

clydin:
The `$location` service relies on the URL constructor which is not present in IE9-11.  core-js provides a polyfill which has now been added to the ES5 polyfill set which will be loaded for IE9-11.

Size check failure is expected due to the addition of ES5 polyfills.
Gzipped size change is 38kb -> 44kb


alan-agius4:
Maybe we should add this polyfill only when IE9-11 support is needed?

filipesilva:
We're reluctant to add this polyfills for all ES5 users when it's just needed for hybrid apps. Instead we should document it and make it easier to add it.

PR #14850, opened 2019-06-21
refactor(@angular/cli): add missing defaults to update schema

Splaktar:
It's useful to indicate the defaults in the docs.
Currently `migrateOnly` and `allowDirty` don't list their defaults in the docs.
This addresses that.

Screen Shot 2019-06-20 at 21 14 16

### DIFF
--- a/packages/angular/cli/commands/update.json
+++ b/packages/angular/cli/commands/update.json
@@ -44,6 +44,7 @@
         },
         "migrateOnly": {
           "description": "Only perform a migration, does not update the installed version.",
+          "default": false,
           "type": "boolean"
         },
         "from": {
@@ -56,6 +57,7 @@
         },
         "allowDirty": {
           "description": "Whether to allow updating when the repository contains modified or untracked files.",
+          "default": false,
           "type": "boolean"
         }
       }


clydin:
The lack of a default actually has an underlying purpose.  It's the only way, currently, to ascertain whether the user provided the option on the command line.  If a default is present then the value will never be undefined.  This technically shouldn't matter in these cases but a review of the code should probably be done to ensure there are no regressions.

However, it's actually kind of strange that the `=true|false` part is in the first column.  They're really flags.  Their presence enables the prescribed behavior.  Personally, I think the `true|false` inclusion is more distracting and confusing than useful.  A manpage, for instance, just lists the flag options and I find it far easier to read ([example](https://www.freebsd.org/cgi/man.cgi?query=gzip&apropos=0&sektion=0&manpath=FreeBSD+12.0-RELEASE+and+Ports&arch=default&format=html)).

Splaktar:
It looks like the `build-bazel` step timed out (over 10 minutes). I don't think that there is anything in this PR that would cause or be related to that.

Splaktar:
@jenniferfell or @jbogarthyde can speak to the formatting of the docs. I know that this part of the docs formatting went through a number of rounds of tweaks and iterations based on feedback.

jbogarthyde:
@petebacondarwin 
 
These flags all default to false - are you saying that simply including, say, `--next` is the same as setting `--next="true"`?  That would be equivalent to your example man page, where all of the options seem to be flags, and none of them show possible values. 
That's not the case with the CLI -- some options are not flags, and the doc has to tell you what the allowed values are and how they are interpreted. Including `=true|false` for Booleans is intended to make syntax statements parallel by showing allowed value.

It's true that this notation doesn't cover the case of "no value" required.
It's possible to cover that case in the description - something like "When supplied or set to true", or "When not supplied or set to <default>...". 
To do that, I would have to know in each case what it does and add it to the description string. The "Default is x" line is generated from code, not part of the description string.

There's another complication with describing the Boolean options - for those that default to true, there is a "notX" version that can be included to make them false, equivalent to X=false. (Presumably the `notX` form doesn't need a value-- that's the same thing you're saying about these positive-action flags.) We decided not to make a point of that in the doc, since it seemed overly complicated.

alan-agius4:
I agree with @clydin.

Having the possible values in the first column is really distracting. Especially the format they are presented example `--from=from`. What is the actual expected value here, is it a number, string, boolean or both? 

In the first column, I'd expect the option name and aliases, while in a separate column I'd expect either the possible value type such as `string` or `boolean` or if it is an enum, the allowed values.

alan-agius4:
Also having `When true` in the description is a bit misleading to command argument, since a use doesn't need to supply `true` or `false` for an option to be truthy or falsely.

```
--my-option         // = true
--no-my-option    // =false
```

alan-agius4:
Closing as per the above discussion.

Splaktar:
@aikidave We still need to have the default listed in [the documentation](https://angular.io/cli/update) for `ng update`'s `--allow-dirty` and `--migrate-only` flags.

alan-agius4:
In all fairness I think for `allowDirty` we can provide a default value, because it doesn’t seem to have any underlying reason for it to be not defined.

alan-agius4:
Also in the docs `migrateOnly` is only listed as accepting a Boolean, but this is not correct as strings of schematic names are also accepted.

Though this is more of a change how we generate the JSON help which is later used to generate the documentation.

alan-agius4:
That’s said, I am not a big fan of having `true/false` listed in the command line arguments man page. I think this is quite unique and should be revised as I cannot recall I even saw it such man page.

PR #14854, opened 2019-06-21
fix(@angular/cli): log file path normalization

alan-agius4:
In windows the logfile path in the console will be printed with both back and forward slashes if not normalized.


vikerman:
@alan-agius4 - needs rebase

PR #14858, opened 2019-06-21
fix(@angular/cli): validate version in doc command

cexbrayat:
As the JSON Schema validation is minimum in the CLI, we have to also validate in code.
This PR:
- updates the JSON Schema to use `number` and `enum` instead of `integer` and `const` that are not supported.
- adds validation in the doc command implementation
- adds error reporting if the version is not valid
- fixes a typo in an error message in the parser

### DIFF
--- a/packages/angular/cli/commands/doc-impl.ts
+++ b/packages/angular/cli/commands/doc-impl.ts
@@ -23,10 +23,16 @@ export class DocCommand extends Command<DocCommandSchema> {
     let domain = 'angular.io';
 
     if (options.version) {
+      // version can either be a string containing "next"
       if (options.version == 'next') {
         domain = 'next.angular.io';
-      } else {
+      // or a number where version must be a valid Angular version (i.e. not 0, 1 or 3)
+      } else if (!isNaN(+options.version) && ![0, 1, 3].includes(+options.version)) {
         domain = `v${options.version}.angular.io`;
+      } else {
+        this.logger.error('Version should either be a number (2, 4, 5, 6...) or "next"');
+
+        return 0;
       }
     }
 
--- a/packages/angular/cli/commands/doc.json
+++ b/packages/angular/cli/commands/doc.json
@@ -27,16 +27,13 @@
           "description": "When true, searches all of angular.io. Otherwise, searches only API reference documentation."
         },
         "version" : {
-          "anyOf": [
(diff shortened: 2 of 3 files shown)


packages/angular/cli/commands/doc.json
```diff
@@ -29,14 +29,11 @@
...
-              "const" : 2
-            },
-            {
-              "type" : "integer",
+              "type": "number",
               "minimum": 4
             },
             {
-              "const": "next"
+              "enum": [2, "next"]
```
alan-agius4: I am not understanding this `2` here, what is the reason behind `2` being accepted?

packages/angular/cli/commands/doc-impl.ts
```diff
@@ -23,10 +23,16 @@ export class DocCommand extends Command<DocCommandSchema> {
     let domain = 'angular.io';
 
     if (options.version) {
+      // version can either be a string containing "next"
       if (options.version == 'next') {
         domain = 'next.angular.io';
-      } else {
+      // or a number where version must be a valid Angular version (i.e. not 0, 1 or 3)
+      } else if (!isNaN(+options.version) && ![0, 1, 3].includes(+options.version)) {
```
alan-agius4: Shouldn't need to revalidate.

packages/angular/cli/commands/doc.json
clydin: https://v2.angular.io

packages/angular/cli/commands/doc.json
```diff
@@ -29,14 +29,11 @@
         "version" : {
           "anyOf": [
```
alan-agius4: This should be `oneOf`

packages/angular/cli/commands/doc-impl.ts
cexbrayat: Sadly I think we need to, because the CLI support for JSON schema validation is really minimal. It was possible to do `ng doc --version whatever` or `ng doc --version 3` despite the schema. I talked about that with @clydin and we came to the conclusion that we needed to validate in code as well.

packages/angular/cli/commands/doc.json
cexbrayat: Amended with `oneOf`

PR #14864, opened 2019-06-22
fix: fix BUILD file filter for npm package to include both BUILD and BUILD.bazel file

gregmagolan:
This removes the file https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/third_party/github.com/Microsoft/TypeScript/BUILD.bazel which currently makes it into the npm dist at `@schematics/angular/third_party/github.com/Microsoft/TypeScript/BUILD.bazel`.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -264,7 +264,7 @@ export default async function(
         }
 
         // Remove Bazel files from NPM.
-        if (fileName.endsWith('BUILD')) {
+        if (fileName === 'BUILD' || fileName === 'BUILD.bazel') {
           return false;
         }
 


gregmagolan:
Creates some issues with validate job:
```
Running BUILD files validation...
    The package "@angular-devkit/build-angular" does not have a BUILD file associated to it. You must either set an exception or make sure it can be built using Bazel.
    The package "@angular-devkit/build-ng-packagr" does not have a BUILD file associated to it. You must either set an exception or make sure it can be built using Bazel.
    The package "@angular-devkit/build-webpack" does not have a BUILD file associated to it. You must either set an exception or make sure it can be built using Bazel.
    The package "@ngtools/webpack" does not have a BUILD file associated to it. You must either set an exception or make sure it can be built using Bazel.
    Found some BUILD files missing, which will be breaking your PR soon.
```
Any ideas @alan-agius4 @filipesilva?

gregmagolan:
This code looks kind of fishy 🐟  
```
  // TODO: enable this to break
  if (error) {
    // process.exit(1);
    logger.warn('Found some BUILD files missing, which will be breaking your PR soon.');
  }
```

alan-agius4:
@gregmagolan that should be a problem, the actual failure is due to the commit msg. Check slack #tools-bazel.

PR #14866, opened 2019-06-23
Point to angular.json when configuring proxy file

SerkanSipahi:


alan-agius4:
Hi, thanks for your contribution. However as stated I. The first line of this document, this is deprecated and we no longer accept PRs to improve and maintained it.

New documentation is available at http://angular.io

PR #14877, opened 2019-06-24
fix(@angular-devkit/build-angular): force event category to be something

hansl:
If it is empty (or undefined), the Universal Analytics package will not send events.


vikerman:
@hansl  - Tests are failing

hansl:
@vikerman are those flakes? They seem unrelated to this change.

PR #14890, opened 2019-06-25
Add `@angular/compiler-cli` version `8.1.0-next.0` as `peerDependency`

alan-agius4:
**fix(@angular-devkit/build-angular): add `@angular/compiler-cli` version `8.1.0-next.0` as `peerDependency`**
 
**fix(@ngtools/webpack): add `@angular/compiler-cli` version `8.1.0-next.0` as `peerDependency`**
 
**test: add `--force` to version 7 migration when updating prerelease versions**

We only use it for a few things but have a strict peerdep.

This strictness causes errors when updating the CLI from 7.x to 8.1 next projects:

Package "@angular-devkit/build-angular" has an incompatible peer dependency to "@angular/compiler-cli" (requires ">=8.0.0-beta.0 < 9.0.0" (extended), would install "8.1.0-next.3").
`build-angular` didn't have a peerdep on `@angular/compiler-cli` that supported `8.1.0-next`

This PR relaxes this requirement.

After the other commits get merged and released, we should to revert this commit.

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -86,7 +86,7 @@
     "zone.js": "^0.9.1"
   },
   "peerDependencies": {
-    "@angular/compiler-cli": ">=8.0.0-beta.0 < 9.0.0",
+    "@angular/compiler-cli": "^8.0.0-beta.0 || ^8.1.0-beta.0 || ^8.2.0-beta.0 || ^8.3.0-beta.0 || ^8.4.0-beta.0 || >=9.0.0-beta < 9",
     "typescript": ">=3.1 < 3.5"
   }
 }
--- a/packages/ngtools/webpack/package.json
+++ b/packages/ngtools/webpack/package.json
@@ -28,7 +28,7 @@
     "webpack-sources": "1.3.0"
   },
   "peerDependencies": {
-    "@angular/compiler-cli": ">=8.0.0-beta.0 < 9.0.0",
+    "@angular/compiler-cli": "^8.0.0-beta.0 || ^8.1.0-beta.0 || ^8.2.0-beta.0 || ^8.3.0-beta.0 || ^8.4.0-beta.0 || >=9.0.0-beta < 9",
     "typescript": ">=3.4 < 3.5",
     "webpack": "^4.0.0"
   },
--- a/tests/legacy-cli/e2e/tests/update/update-7.0.ts
+++ b/tests/legacy-cli/e2e/tests/update/update-7.0.ts
@@ -1,14 +1,11 @@
 import { createProjectFromAsset } from '../../utils/assets';
 import { expectFileMatchToExist, expectFileToExist, expectFileToMatch } from '../../utils/fs';
 import { ng, noSilentNg, silentNpm } from '../../utils/process';
-import {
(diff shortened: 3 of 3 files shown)


packages/angular_devkit/build_angular/package.json
```diff
@@ -86,7 +86,7 @@
     "zone.js": "^0.9.1"
   },
   "peerDependencies": {
-    "@angular/compiler-cli": ">=8.0.0-beta.0 < 9.0.0",
+    "@angular/compiler-cli": "^8.0.0-beta.0 || ^8.1.0-next.0",
```
filipesilva: Can you keep 9.0.0 as an exclusive upper bound? We really want to ensure that these packages are compatible with same version only.

In fact, this range also needs to allow for betas of 9. So all of these should be valid peerdeps:
```
8.0.0-beta.0
8.0.0-rc.0
8.0.0
8.0.1
8.1.0-beta.0
8.1.0-next.0
8.1.1
9.0.0-beta.0
9.0.0-rc.0
```

alan-agius4: There is a bit of misconception on what `< 9.0.0` really means. This range doesn't consider pre-releases when having a stable version `8`. This is because the stable version takes precedence.

`>=8.0.0-beta.0 < 9.0.0`  is equivalent for `^8.0.0-beta.0` . Semver will only match stable versions after `8.0.0`.

If you `>=8.0.0-beta.0 < 9.0.0` in semver calculator https://semver.npmjs.com/

You will get the below.
image

As you can see, `8.1.0-beta.0` is not matched because it's a prerelease.

alan-agius4: Another better is example  is this range is when for example having the below package list,

```
0.0.0-0
0.0.0-1
0.0.0-2
0.0.0-3
0.0.0-4
0.0.0-5
0.0.0-6
0.0.0-7
2.0.0-rc.0
2.0.0-rc.1
2.0.0-rc.2
2.0.0-rc.3
2.0.0-rc.4
2.0.0-rc.5
2.0.0-rc.6
2.0.0-rc.7
2.0.0
```

With the following range `>=0.0.0-0 <=2`. one would assume that `2.0.0-rc.0` is also matched, however this is not the same. As the above is interpreted as `^0.0.0-0 || ^2.0.0`, which would exclude any prerelease of `2.0.0` to be matched.

filipesilva: Yes, I agree that the current peerdep does not include the 9.0.0 beta/rc. What I mean is that it should, because it is the same idea as in here:
https://github.com/angular/angular-cli/blob/ce578ae9b3c92acc00fd768f2d551430113cedbf/packages/angular_devkit/build_angular/src/utils/version.ts#L61-L64

You are also right that I didn't understand that `>=8.0.0-beta.0 < 9.0.0` is equivalent to `^8.0.0-beta.0`. Thanks for explaining it.

So I guess the correct peerdep range is actually `^8.0.0-beta || >=9.0.0-beta <9.0.0` and the snippet above should match this as well (`^${cliMajor}.0.0-beta || ` + `>=${cliMajor + 1}.0.0-beta <${cliMajor + 1}.0.0`).

If I check `^7.0.0-beta || >=8.0.0-beta <8.0.0` it seems to give what I described:

image

PR #14905, opened 2019-06-26
feat(@schematics/angular): add `strict` compiler option to workspace tsconfig

### LINKED ISSUE #14694: Enable all strict type checking options.
# 🚀 Feature request

```
- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description

set compilerOptions `strict:true` in `tsconfig.json`

alan-agius4:
Add "stricter" compiler option to workspace tsconfig

Closes #14694

### DIFF
--- a/packages/schematics/angular/workspace/schema.json
+++ b/packages/schematics/angular/workspace/schema.json
@@ -31,6 +31,11 @@
       "type": "boolean",
       "default": false,
       "x-user-analytics": 14
+    },
+    "strict": {
+      "description": "Creates a workspace with stricter TypeScript compiler options.",
+      "type": "boolean",
+      "default": false
     }
   },
   "required": [
--- a/packages/schematics/angular/workspace/files/tsconfig.json.template
+++ b/packages/schematics/angular/workspace/files/tsconfig.json.template
@@ -2,7 +2,12 @@
   "compileOnSave": false,
   "compilerOptions": {
     "baseUrl": "./",
-    "outDir": "./dist/out-tsc",
+    "outDir": "./dist/out-tsc",<% if (strict) { %>
+    "noImplicitAny": true,
+    "noImplicitReturns": true,
+    "noImplicitThis": true,
+    "noFallthroughCasesInSwitch": true,
+    "strictNullChecks": true,<% } %>
     "sourceMap": true,
     "declaration": false,
     "downlevelIteration": true,
--- a/packages/schematics/angular/workspace/index_spec.ts
+++ b/packages/schematics/angular/workspace/index_spec.ts
@@ -71,4 +71,16 @@ describe('Workspace Schematic', () => {
     expect(files).not.toContain('/tslint.json');
     expect(files).not.toContain('/.editorconfig');
   });
+
+  it('should not add strict compiler options when false', async () => {
+    const tree = await schematicRunner.runSchematicAsync('workspace', { ...defaultOptions, strict: false }).toPromise();
+    const { compilerOptions } = JSON.parse(tree.readContent('/tsconfig.json'));
+    expect(compilerOptions.strictNullChecks).not.toBeDefined();
+  });
+
+  it('should not add strict compiler options when true', async () => {
+    const tree = await schematicRunner.runSchematicAsync('workspace', { ...defaultOptions, strict: true }).toPromise();
+    const { compilerOptions } = JSON.parse(tree.readContent('/tsconfig.json'));
+    expect(compilerOptions.strictNullChecks).toBe(true);
+  });
 });
--- a/tests/legacy-cli/e2e/tests/build/platform-server.ts
+++ b/tests/legacy-cli/e2e/tests/build/platform-server.ts
@@ -10,6 +10,8 @@ export default async function () {
   const veEnabled = argv['ve'];
 
   await ng('add', '@nguniversal/express-engine', '--client-project', 'test-project');
+  // todo remove once https://github.com/angular/universal/pull/1229 is landed
+  await silentNpm('install', '@types/express', '--save-dev');
 
   await updateJsonFile('package.json', packageJson => {
     const dependencies = packageJson['dependencies'];
--- a/tests/legacy-cli/e2e/tests/misc/browsers.ts
+++ b/tests/legacy-cli/e2e/tests/misc/browsers.ts
@@ -39,7 +39,7 @@ export default async function () {
   await replaceInFile(
     'e2e/src/app.e2e-spec.ts',
     'await browser.manage().logs().get(logging.Type.BROWSER)',
-    '[]',
+    '[] as any',
   );
 
   // Workaround defect in getText WebDriver implementation for Safari/Edge


vikerman:
This need more discussion. I want to understand the impact and also why the app specific tsconfig has no strictness flags set.

clydin:
From a review of similar tools, the following all set `strict` by default:
* [create-react-app](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/utils/verifyTypeScriptSetup.js#L107)
* [vue-cli](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-typescript/generator/template/tsconfig.json#L5)
* [firebase-tools](https://github.com/firebase/firebase-tools/blob/master/templates/init/functions/typescript/tsconfig.json#L8)

xiaoxiangmoe:
@vikerman the demo codes in [angular.io](https://angular.io/) need to be updated for strict type check.

filipesilva:
We need to check with our AIO examples if they even work with these options. Some of them are also problematic on their own, like `strictPropertyInitialization` that doesn't work well with `@Input` and such. So we should consider them individually.

xiaoxiangmoe:
`strictPropertyInitialization` works well with `definite assignment assertion modifiers`. see [Strict Class Initialization](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization)

```ts
class C {
    @Input() foo!: number;
    //          ^
    // Notice this '!' modifier.
    // This is the "definite assignment assertion"
}
```

If you need more help about strict type check, I am happy to help.

xiaoxiangmoe:
What do I need to do to get this PR merged?

vikerman:
@xiaoxiangmoe - On our side we need to turn one flag at a time in our documentation projects to make sure all our documentation works with each of these flags. And then enable it in the CLI by default.

For example adding a "!" is not acceptable without changing our documentation to reflect something like that(and I don't know whether it's a good idea to make everyone do that)

alan-agius4:
I just had a look at this, and it seems that AIO is already in `strict` mode  with the exception of `strictPropertyInitialization` which is false.

See: https://github.com/angular/angular/blob/1d3e22766a978bc8bd6e3ef29cec5cdb44e04278/aio/tsconfig.json#L22-L25

I have updated this PR to reflect the AIO settings.

xiaoxiangmoe:
@alan-agius4 can we enable strictPropertyInitialization in angular 9?

alan-agius4:
@xiaoxiangmoe, while I cannot say for certain. I highly doubt that we would be able to enable `strictPropertyInitialization` in version 9. Especially due to the `@input` issue explained above.

xiaoxiangmoe:
@alan-agius4 Maybe we can provide some instructions in the documentation, what do we need to do if we want to enable `strictPropertyInitialization`.

alan-agius4:
Hi @xiaoxiangmo, fyi there is an issue in angular/angular related to `strictPropertyInitialization` https://github.com/angular/angular/issues/24571. That said, at this point in time I would not recommend using `strictPropertyInitialization`.

vikerman:
The feedback from Angular team on latest proposal - 

1) Let's ship this for a major release(9.0) so that it gives us time to communicate to people who write tutorials
2) Have an option to turn off all strict flags in case someone wants to quickly prototype something (and something like noImplicitAny might come in the way)

imcotton:
This `@Input` does not guarantee its instance to get a `non-null` / `non-undefined` property (e.g. `[foo]="obs$ | async"`), therefore Component author should always anticipate its property as optional (e.g. `@Input() foo?: number`), which a) gains type safety b) complies with the `strict` restriction.

End users should fully embrace the `strict` checking from TypeScript, set `strictPropertyInitialization` to `false` here only because `@Input` by CLI by default is like to choosing a sad path for user before they were taking steps, it's not a defect, not necessary to compromise at the first place.

vikerman:
The current list which we think is basic for most samples/examples in the wild and also our ecosystem is
```
    "noImplicitAny": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "noFallthroughCasesInSwitch": true,
    "strictNullChecks": true
```

Splaktar:
I really like the idea of having a `strict` option and the ability to have it include these compiler flags in `tsconfig.json`. However, making it the default is a bit concerning. I like the fact that this is only being considered for version 9.

I would really prefer to make this `strict` option default to `false`, then use the Analytics to determine if the usage is sufficient to warrant making it the default. This would also give us time to see what kinds of issues people run into with `strict` mode before making it a default.

Perhaps strict mode that defaults to `false` is something that could go into a version 8 minor for a month or two? If no more version 8 minors are planned, then perhaps it can go into version 9 with default `false` and then be considered for enabling the default in version 10?

Splaktar:
I'd also like to have these settings tested against projects generated using all of the Angular Material Schematics and the [material.angular.io docs project](https://github.com/angular/material.angular.io/blob/master/src/tsconfig.json).

packages/schematics/angular/workspace/schema.json
```diff
@@ -31,6 +31,11 @@
       "type": "boolean",
       "default": false,
       "x-user-analytics": 14
+    },
+    "strict": {
+      "description": "Creates a workspace with stricter TypeScript compiler options.",
+      "type": "boolean",
+      "default": true
```
vikerman: We should probably set this to `false` and get the PR in. And move the decision on whether we want to turn in on for 9.0 based on data we collect for this, later.

vikerman:
> I'd also like to have these settings tested against projects generated using all of the Angular Material Schematics and the [material.angular.io docs project](https://github.com/angular/material.angular.io/blob/master/src/tsconfig.json) (and yes, I recognize that this change would only directly impact newly created projects).

The actual Angular material samples tsconfig is here and already using here  - https://github.com/angular/components/blob/cf70c4342e7ea277011cf289e721383b7784a18a/src/material-examples/tsconfig-build.json

We will look into what tests we have around testing Material schematics. Thanks.

packages/schematics/angular/workspace/schema.json
clydin: I think `"noImplicitAny": true,` should be there regardless of the setting.  Without it, a developer is essentially bypassing the overwhelming majority of the benefits of TypeScript.

Splaktar:
Glad to see the examples using that.

I enabled these checks in material.angular.io's [tsconfig](https://github.com/angular/material.angular.io/blob/ed06d61ed692caf06f877558af89d01863e78848/src/tsconfig.json) and I'm seeing about 30 errors. Half of them are due to `noImplicitAny`.

It's not the end of the world and I can try to put together a PR to fix this, but it's an indication that people will certainly hit this. I guess the hope is that it will encourage them to write cleaner code, but I think that there will certainly be issues with tutorials, guides, blog posts, and other material that "won't work" in the default Angular CLI after strict becomes enabled by default.

vikerman:
This is only for new projects. So we do fully expect existing projects to break when users manually switch the strict flags. We need to find a way to get all course creators to ensure that if they create a project created from scratch their code works with it.

packages/schematics/angular/workspace/schema.json
lacolaco: I guess just adding a new flag is not enough to get noticed by developers. It may be better to make a prompt like "Would you like to enable TypeScript strict type checking? (y/N)". And I agree on the `false` by default.

alan-agius4: I think for sure we should

1) Turn off by default, add prompt so that it's more discoverable by users 
2) We revisit this based on the feedback we get right before version 9 RC.

xiaoxiangmoe:
> If there's a prompt now, I think it would be clearer to just use the actual TS strict option. I think it will be confusing that stricter type checking is not actually strict as is the case for other frameworks.

Maybe we can consider providing `--strict=false` `--strict=stricter` , and then keep the default is `--strict=true`

clydin:
The current plan is to slowly move to a more strict configuration over the next set of versions.  This is to ensure that both first and third party documentation, training material, courseware, etc. are adequately prepared for the changes.

Splaktar:
Is there an associated GitHub or JIRA issue for verifying that all examples on AIO compile with the stricter flags? For verifying that CLI, PWA, and Universal Schematics generate code that compiles with the new flags?

I've opened https://github.com/angular/components/issues/17135 to track this for the Angular Material Schematics.

PR #14934, opened 2019-06-28
fix(@ngtools/webpack): NGCC should process secondary entry-points

petebacondarwin:
Previously secondary entry-points, e.g
`@angular/common/http`,  were being resolved to
the primary entry-point, e.g `@angular/common`.
This meant that these entry-points were not being
processed as part of a compilation leading to error messages.

This change inverts the `tryResolvePackage` logic
to attempt to use `require.resolve()` first and only
looking in the parent directory of the typings file
as a fallback if the main resolve fails.

Fixes https://github.com/angular/angular/issues/30972#issuecomment-506428267


AhsanAyaz:
Awesome @petebacondarwin. I was upgrading my library to use Angular CLI and stumbled upon the same issue with IVY.
Thanks much! 🎉

PR #14955, opened 2019-07-01
feat(@angular-devkit/build-angular): deprecate scripts and styles `la…

### LINKED ISSUE #14814: Deprecate styles and script `lazy` option in favor of a more self descriptive option
# 🚀 Feature request

### Description
The `lazy` option inside the `script` and `style` option is confusing as this option doesn't lazy load a bundle but rather it doesn't inject/reference the script in the HTML. While this option is an enabler for lazy loading, the users will still need to handle on how how this bundle will be lazy loaded. There are also potential use cases beyond lazy loading for the option.

Deprecate the `lazy` option in favour of a more self descriptive options such as `injected`.

alan-agius4:
…zy` option in favor of`inject`

The lazy option inside the script and style option is confusing as this option doesn't lazy load a bundle but rather it doesn't inject/reference the script in the HTML. While this option is an enabler for lazy loading, the users will still need to handle on how how this bundle will be lazy loaded. There are also potential use cases beyond lazy loading for the option.

Closes #14814

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -428,7 +428,13 @@
             "lazy": {
               "type": "boolean",
               "description": "If the bundle will be lazy loaded.",
-              "default": false
+              "default": false,
+              "x-deprecated": "Use 'inject' option with 'false' value instead."
+            },
+            "inject": {
+              "type": "boolean",
+              "description": "If the bundle will be referenced in the HTML file.",
+              "default": true
             }
           },
           "additionalProperties": false,
--- a/packages/angular_devkit/build_angular/test/browser/scripts-array_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/scripts-array_spec_large.ts
@@ -33,9 +33,9 @@ describe('Browser Builder scripts array', () => {
     'src/binput-script.js',
     'src/ainput-script.js',
     'src/cinput-script.js',
-    { input: 'src/lazy-script.js', bundleName: 'lazy-script', lazy: true },
+    { input: 'src/lazy-script.js', bundleName: 'lazy-script', inject: false },
     { input: 'src/pre-rename-script.js', bundleName: 'renamed-script' },
-    { input: 'src/pre-rename-lazy-script.js', bundleName: 'renamed-lazy-script', lazy: true },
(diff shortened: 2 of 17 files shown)


packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -418,7 +418,13 @@
             "lazy": {
               "type": "boolean",
               "description": "If the bundle will be lazy loaded.",
-              "default": false
+              "default": false,
+              "x-deprecated": "Use 'inject' instead."
```
filipesilva: Can you change this to `Use the 'inject' option with 'false' value instead.` please? Just changing `lazy` to `inject` won't be what users want.

packages/angular_devkit/build_angular/test/browser/scripts-array_spec_large.ts
```diff
@@ -11,19 +11,18 @@ import { TestLogger } from '@angular-devkit/architect/testing';
...
-    'src/zinput-script.js': 'console.log(\'zinput-script\');',
-    'src/finput-script.js': 'console.log(\'finput-script\');',
-    'src/uinput-script.js': 'console.log(\'uinput-script\');',
-    'src/binput-script.js': 'console.log(\'binput-script\');',
-    'src/ainput-script.js': 'console.log(\'ainput-script\');',
-    'src/cinput-script.js': 'console.log(\'cinput-script\');',
-    'src/lazy-script.js': 'console.log(\'lazy-script\');',
-    'src/pre-rename-script.js': 'console.log(\'pre-rename-script\');',
-    'src/pre-rename-lazy-script.js': 'console.log(\'pre-rename-lazy-script\');',
+    'src/input-script.js': "console.log('input-script'); var number = 1+1;",
```
filipesilva: Surprised the use of `"` does not break our linter.

alan-agius4: That is weird indeed.

alan-agius4:
What I had in mind is to add such things as a future migration. Ie, it will already be inside the code and unit tested but it won't run when doing `ng update`, as the target will be version `9.0.0-beta.0`. 

What do you reckon?

filipesilva:
I think that's a pretty good approach.

alan-agius4:
Will do the update schematic in a separate PR in the coming days

filipesilva:
Is it important that it's in a separate PR? Having everything in a single PR would generally be better.

Issue #14589, opened 2019-05-30
ng update did not change my import statements


markgoho:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [X] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

No, this behavior did not exist prior to 8.x

### Description

ran `ng update @angular/cli @angular/core`
`loadChildren` import statements were not updated to new syntax

## 🔬 Minimal Reproduction
I haven't dug into the migration code, but I have my routes in a `app-routing.module.ts` file:
``` typescript
import { NgModule } from '@angular/core';
import { Routes, RouterModule } from '@angular/router';

const routes: Routes = [
  {
    path: '',
    pathMatch: 'full',
    loadChildren: './home/home.module#HomeModule',
    data: { depth: 0 },
  },
];

@NgModule({
  imports: [RouterModule.forRoot(routes)],
  exports: [RouterModule],
})
export class AppRoutingModule {}
```

## 🔥 Exception or Error
No error, just old import pattern instead of new.

## 🌍 Your Environment
I was on Angular 7.2.15 and this was part of the update process to 8.0.0.

**Anything else relevant?**
I'm wondering if the migration code is targeting a certain file or something else that would have excluded this file from being updated to the new import pattern.

Here's the last bit from the `ng update` command:
``` bash
    ** Executing migrations for package '@angular/core' **
            ------ Static Query Migration ------
            With Angular version 8, developers need to
            explicitly specify the timing of ViewChild and
            ContentChild queries. Read more about this here:
            https://v8.angular.io/guide/static-query-migration
            ------------------------------------------------
UPDATE src/app/search/search.component.ts (2049 bytes)
```

alan-agius4:
Can you also confirm that you were not using an RC or beta version of Angular CLi?

markgoho:
@alan-agius4 not using an RC or beta.

When I try to manually update the import statements, I'm getting a Typescript error:
`Dynamic import is only supported when '--module' flag is 'commonjs' or 'esNext'.ts(1323)`
``` typescript
  {
    path: '',
    pathMatch: 'full',
    // loadChildren: './home/home.module#HomeModule',
    loadChildren: () => import('./home/home.module').then(m => m.HomeModule),
[...]

PR #14996, opened 2019-07-04
fix(@schematics/update): update `@angular/pwa` when migrating to 8

alan-agius4:
Related to https://github.com/angular/angular-cli/issues/14589#issuecomment-508026441


clydin:
There is currently no method to execute `ng add` schematics without installing the containing package first.  This is actually something I have considered supporting as there is the potential for a package that will only serve as a vehicle for the addition schematics and having it otherwise installed just adds noise to the project.

Issue #14993, opened 2019-07-04
ng ignore global npm config for add and update command


diyfr:
Ng command
```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [X] add
- [X] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```
### Description

ng `add` and  `update`  ignore npm global config

## 🔬 Minimal Reproduction
Define a private registry (it's also an proxy for publics packages)
```
npm config set registry https://myregistry/repository/myrepo/ --global
```
Set authentication
```
npm login --registry=https://myregistry/repository/myrepo/ 
```
Install angular/cli
```
npm install -g @angular/cli
```
And start a new project
```
ng new myproj
```
Ok project created and all dependencies installed (private registry checked in package-lock.json)

The problem starts here
I want to transform my project to pwa [see](https://angular.io/guide/service-worker-getting-started)
```
ng add @angular/pwa --project myproj
```
response  : `The add command requires to be run in an Angular project, but a project definition could not be found.`  
Ok change directory
```
cd myproj
ng add @angular/pwa
```
response  : `Unable to fetch package metadata: request to https://registry.npmjs.org/@angular%2fpwa failed, reason: connect ETIMEDOUT 104.16.26.35:443.`  (I'am also behind a enterprise proxy)

same result with `update` command, ng command want to access directly to registry.npmjs.org and not by private registry

## 🌍 Your Environment
<pre><code>
Angular CLI: 8.1.0
Node: 10.10.0
OS: win32 x64
Angular: 8.1.0
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.801.0
@angular-devkit/build-angular     0.801.0
@angular-devkit/build-optimizer   0.801.0
@angular-devkit/build-webpack     0.801.0
@angular-devkit/core              8.1.0
@angular-devkit/schematics        8.1.0
@ngtools/webpack                  8.1.0
@schematics/angular               8.1.0
@schematics/update                0.801.0
rxjs                              6.4.0
typescript                        3.4.5
webpack                           4.35.2
</code></pre>

Npm version  -> 6.9.2

alan-agius4:
Hi, I tried this out and global npmrc is being read.

```
[...]

PR #15000, opened 2019-07-05
fix(@angular/cli): add `verbose` option to `update` and `add`

alan-agius4:
At the moment there is no way to turn on the verbose logging for `ng update` and `ng add`. This is useful for use so that when users report issues such as npmrc is not read we can see the lookup locations.

This also removes some reduncant that were being provided in `executeSchematic`.

Related to https://github.com/angular/angular-cli/issues/14993

### DIFF
--- a/packages/angular/cli/commands/add.json
+++ b/packages/angular/cli/commands/add.json
@@ -30,6 +30,11 @@
               "format": "hostname"
             }
           ]
+        },
+        "verbose": {
+          "description": "Display additional details about internal operations during execution.",
+          "type": "boolean",
+          "default": false
         }
       },
       "required": [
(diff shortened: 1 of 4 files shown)


packages/angular/cli/commands/add.json
```diff
@@ -30,6 +30,11 @@
               "format": "hostname"
             }
           ]
+        },
+        "verbose": {
+          "description": "When true, display additional details during the add process.",
```
clydin: `When true,` doesn't really make sense here since it's a flag.
Can you also make the description more generic?  I'd like to eventually make it a global option for all commands.
Something maybe like: `Display additional details about internal operations during execution.`

alan-agius4: I do agree with you, through probably regarding the “When true”, we should probably remove it from all of the other descriptions as well. As it seems to be some the standard for Boolean options.

PR #15010, opened 2019-07-07
feat(@angular-devkit/build-angular): allow control of index output path

clydin:
This allows the output path of an application's index HTML file to be controlled independently of the input file.  The output path for the file will be considered relative to the application's configured output path.  This allows an application to contain multiple input index files for different configurations and allow the output file to remain constant.  This also enables the placement of the index file in a subdirectory within the output path or changing of the name of the output index file; neither of which was previously possible.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -270,8 +270,31 @@
       "x-deprecated": true
     },
     "index": {
-      "type": "string",
-      "description": "The name of the index HTML file."
+      "description": "Configures the generation of the application's HTML index.",
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "The path of a file to use for the application's HTML index. The filename of the specified path will be used for the generated file and will be created in the root of the application's configured output path."
+        },
+        {
+          "type": "object",
+          "description": "",
+          "properties": {
+            "input": {
+              "type": "string",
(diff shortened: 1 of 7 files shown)


packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -270,8 +270,31 @@
       "x-deprecated": true
     },
     "index": {
-      "type": "string",
-      "description": "The name of the index HTML file."
+      "description": "Configures the generation of the application's HTML index.",
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "The path of a file to use for the application's HTML index.  The filename of the specified path will be used for the generated file and will be created in the root of the application's configured output path."
```
alan-agius4: Nit: single space after the fullstop `index.  The`

packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -270,8 +270,31 @@
...
+            "input": {
+              "type": "string",
+              "minLength": 1,
+              "description": "The path of a file to use for the application's generated HTML index."
+            },
+            "output": {
+              "type": "string",
+              "minLength": 1,
+              "default": "index.html",
+              "description": "The output path of the application's generated HTML index file.  The full provided path will be used and will be considered relative to the application's configured output path."
```
alan-agius4: Same

PR #15021, opened 2019-07-09
build: update webpack to version 4.35.3

alan-agius4:
Supersedes https://github.com/angular/angular-cli/pull/15019

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -942,11 +942,6 @@ accepts@~1.3.7:
     mime-types "~2.1.24"
     negotiator "0.6.2"
 
-acorn-dynamic-import@^4.0.0:
-  version "4.0.0"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -11059,6 +11064,35 @@ webpack@4.35.2:
     watchpack "^1.5.0"
     webpack-sources "^1.3.0"
 
+webpack@4.35.3:
```
devoto13: Looks like it will install two versions of `webpack` package now (second record is above). Shall the version be updated in some other package as well?

alan-agius4: Good catch!

kyliau:
@alan-agius4 could you please resolve merge conflict? might need to create separate PR for patch. Thank you!

PR #15024, opened 2019-07-09
CircleCI Configuration Refactor

clydin:
Prep work (and some cleanup) for saucelabs-based differential loading testing.
The environment shell scripts have been transferred from FW.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,101 +7,129 @@
 # To validate changes, use an online parser, eg.
 # http://yaml-online-parser.appspot.com/
 
+version: 2.1
+
 # Variables
 
 ## IMPORTANT
 # If you change the cache key prefix, also sync the restore_cache fallback to match.
 # Keep the static part of the cache key as prefix to enable correct fallbacks.
 # See https://circleci.com/docs/2.0/caching/#restoring-cache for how prefixes work in CircleCI.
-var_1: &default_docker_image circleci/node:12.1
-var_2: &browsers_docker_image circleci/node:12.1-browsers
-var_3: &browsers_docker_image_node_10 circleci/node:10.12-browsers
-var_4: &cache_key angular_devkit-0.11.0-{{ checksum "yarn.lock" }}
-
-# Settings common to each job
-anchor_1: &defaults
-  working_directory: ~/ng
-  docker:
-    - image: *default_docker_image
-anchor_1_win: &defaults_win
-  working_directory: ~/ng
-  resource_class: windows.medium
-  shell: powershell.exe -ExecutionPolicy Bypass
-  machine:
-    image: windows-server-2019
-
-# After checkout, rebase on top of target branch.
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -93,15 +46,78 @@ anchor_7: &setup_bazel_remote_execution
       sudo bash -c "echo -e 'build --config=remote' >> .bazelrc.user";
       echo "Reading from remote cache for bazel remote jobs.";
 
+# Executor Definitions
+executors:
```
filipesilva: This and commands are pretty cool. I was wondering about multi-command abstractions a while back.

.circleci/config.yml
```diff
@@ -93,15 +46,78 @@ anchor_7: &setup_bazel_remote_execution
...
+      - image: circleci/node:12.1
+    working_directory: ~/ng
+
+  test-executor:
+    parameters:
+      nodeversion:
+        type: string
+        default: "12.1"
+    docker:
+      - image: circleci/node:<< parameters.nodeversion >>-browsers
```
filipesilva: Is it possible to use an anchor as a parameter default? Then we could have `var_5: 12.1` and use it in both executors. This way it would be easier for us to remember to change both together.

.circleci/config.yml
```diff
@@ -93,15 +46,78 @@ anchor_7: &setup_bazel_remote_execution
       sudo bash -c "echo -e 'build --config=remote' >> .bazelrc.user";
       echo "Reading from remote cache for bazel remote jobs.";
 
+# Executor Definitions
```
filipesilva: Might be a good idea to add links to docs on the first usage of these new features (executors, commands, parameters). At some point we're going to have to copy these into angular/angular and we might as well have the comments made already.

.circleci/config.yml
```diff
@@ -179,6 +178,9 @@ jobs:
     parallelism: 4
     steps:
       - attach_workspace: *attach_options
+      - run:
+          name: Initialize Environment
+          command: ./.circleci/env.sh
```
filipesilva: If we already have a env script for Linux, should we move the `~/.npm-global/bin` stuff there too?

tests/legacy-cli/e2e/setup/300-log-environment.ts
```diff
@@ -4,6 +4,9 @@ export default async function() {
   console.log('Environment:');
 
   Object.keys(process.env).forEach(envName => {
+    if (envName.toLowerCase().includes('key')) {
```
filipesilva: Can you add a comment explaining what this bailout is for?

.circleci/config.yml
clydin: I was actually considering doing the opposite long-term.  `environment` is cross-platform whereas the shell scripts are not.

filipesilva: Yes, that makes more sense.

PR #15030, opened 2019-07-10
fix(@schematics/angular): exclusively list the files in tsconfigs

alan-agius4:
With the omission of `includes` or `files` in tsconfig files tsc will pick up all the files under the rootDir and make them as part of the compilation. In certain cases, redundant files will be picked up which will cause a slower compilations.

Related to: TOOL-949

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/errors_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/errors_spec_large.ts
@@ -23,11 +23,13 @@ describe('Browser Builder errors', () => {
   it('shows error when files are not part of the compilation', async () => {
     host.replaceInFile(
       'src/tsconfig.app.json',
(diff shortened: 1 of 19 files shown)


filipesilva:
Indeed they can, but it is deprecated. Ivy supports the string syntax as well currently. We decoupled the two things.

Maybe we should only do this in v9 then? Or perhaps we need to document in https://angular.io/guide/deprecations#loadchildren-string-syntax that if you want to use the string syntax you need to add the file to the files array.

alan-agius4:
IMHO, I think an update to the docs should suffice.

alan-agius4:
Make this for Ivy for the time being. And make this the default version 9.

alan-agius4:
The VE, part might be needed if we go with https://github.com/angular/angular-cli/pull/15061#issuecomment-510871409

PR #15038, opened 2019-07-10
ci: use cache and workspace on windows

filipesilva:


filipesilva:
Note to caretaker: a repository admin will need to change the required `ci/circleci: install` check to `ci/circleci: setup`.

.circleci/config.yml
```diff
@@ -108,7 +101,7 @@ commands:
 
 # Job definitions
 jobs:
-  install:
+  setup:
```
filipesilva: Changed `install` to `setup` to better reflect what the job does (checkout, rebase, install) and to match the name used in the FW repository.

filipesilva:
@clydin I didn't track numbers at the time (which I should have) but iirc it wasn't a significant perf gain. The overhead of a new job was higher than the gain of using a cache. The benefit was mostly organizational, where a similar setup is now used for windows as is for linux. I think as soon as we add another windows job there will be a perf gain however.

PR #15039, opened 2019-07-10
Document possible values of analytics command

jbogarthyde:
The possible values are in the JSON (and an error message) but are not mentioned in the doc.

This is probably true of all enum values - in the standard API doc, value types are automatically extracted into the doc, but since these are always strings, we don't do that. We should have a way to automatically include enumerated value descriptions, that would go into the arg description. 

For now, I've put them into a long-description for the command.


jbogarthyde:
I need information about the "ci" value - how does it affect behavior?

Issue #14547, opened 2019-05-29
Enable Ivy in ng test


alan-agius4:
# 🚀 Feature request

### Command (mark with an `x`)

```
- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
At the moment https://v8.angular.io/guide/ivy documents how to opt-in into Ivy for builds, however this doesn't document how to opt-in into Ivy for tests. The same thing applies when generating a new Ivy application via the CLI. Ivy is only enabled for builds.

### Describe the solution you'd like
We should enable the Ivy compiler for tests as well, for both JIT and AOT mode, This is needed, because when enabling the `enableIvy` compiler option in JIT mode it will use process the packages with `ngcc` and will use the Ivy engine which also improves runtime speed in JIT mode.

There are 2 ways that I can think of to achieve this. Either by adding `enableIvy: true` to the specs tsconfig, or the specs tsconfig extends the app tsconfig.

filipesilva:
Closed via https://github.com/angular/angular-cli/pull/15044, given that tests only run in JIT at the moment.

PR #15044, opened 2019-07-11
feat(@schematic/angular): enable ivy in tests

cexbrayat:
If a project is generated with `enableIvy`, this commit adds the necessary configuration to `tsconfig.spec.json` to then run the tests with Ivy. Note that the CLI already does the correct work (runs `ngcc` and then runs the tests).

cc @filipesilva as we talked about it on Slack

### DIFF
--- a/packages/schematics/angular/application/files/tsconfig.spec.json.template
+++ b/packages/schematics/angular/application/files/tsconfig.spec.json.template
@@ -14,5 +14,8 @@
   "include": [
     "src/**/*.spec.ts",
     "src/**/*.d.ts"
-  ]
+  ]<% if (enableIvy) { %>,
+  "angularCompilerOptions": {
+    "enableIvy": true
+  }<% } %>
 }


alan-agius4:
This relates to https://github.com/angular/angular-cli/issues/14547.

With this option your are indeed optioning in to use ngcc but ts transpiration still happens using tsc and not  Is this what we want? Ie, still using only JIT builds during tests?

Should we have an AOT option to run tests in ivy and transpile using ngtsc, or always AOT, or always JIT?

filipesilva:
Yes, we still want unit tests to run with JIT. We just want to ensure that ngcc is ran if needed. For instance if you make a new project and run tests before build. Or if you have some angular libs only on test.

As to having the flag on the spec tsconfig or root tsconfig.... if we have them in individual tsconfigs it will be harder for the user to opt-in or opt back out. And in general we don't really expend a lot of testing in making sure mixed Ivy/VE environments are working properly. 

So maybe we should instead have it only on the workspace tsconfig and be done with it. That way it's a single config change. If you really need to have a non-Ivy project, you can opt-out in their leaf tsconfigs too.

At this point I'm kicking myself for not thinking about this earlier. https://angular.io/guide/ivy would also need to change to mention the workspace tsconfig.

alan-agius4:
I do agree with having the option in the workspace tsconfig. Another option would be to have the `specs` tsconfig extend the `app` tsconfig.

Though with the first approach, a user is not creating an ivy application but rather an ivy workspace. Ie: the `enableIvy` option in the application schematics will be made redundant.

Adding the option in the workspace config will also break `ng g library` by default unless we update the library tsconfig to disable ivy.

alan-agius4:
I do agree with having the option in the workspace tsconfig. Another option would be to have the `specs` tsconfig extend the `app` tsconfig.

Though with the first approach, a user is not creating an ivy application but rather an ivy workspace. Ie: the `enableIvy` option in the application schematics will be made redundant.

Adding the option in the workspace config will also break `ng g library` by default unless we update the library tsconfig to disable ivy.

cexbrayat:
@alan-agius4 👍 Done

PR #15052, opened 2019-07-11
test: use locally published version in E2E project setup

clydin:
This removes the need to perform a second _special_ local build of the packages during the E2E test process.

This also has the benefit of causing CI to fail if the package versions in `latest-versions.ts` are not kept synchronized with the CLI package versions.


tests/legacy-cli/verdaccio.yaml
```diff
@@ -18,7 +18,7 @@ packages:
     access: $all
     publish: $all
 
-  '@ngtools/*':
+  '@ngtools/webpack':
```
filipesilva: Was this change necessary or was it just cleanup? Just wondering, because `@angular-devkit/*` still has the star above.

clydin: “@ngtools/json-schema” is used by ng-packgr.  Since it no longer exists, it’s not published locally which causes the installation of ng-packgr to fail within the E2E tests.

PR #15061, opened 2019-07-12
feat(@ngtools/webpack): show warning when a TS compilations contains …

alan-agius4:
…unused files

When a tsconfig without `includes` or `files` all files under the `rootDir` will be included in the compilation. This results in redundant files to inserted as part of the ts compilation which in some cases might reduce the compilation drastically.

Related to: TOOL-949

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -593,6 +593,50 @@ export class AngularCompilerPlugin {
     }
   }
 
+  private _warnOnUnusedFiles(compilation: compilation.Compilation) {
+    // Only do the unused TS files checks when under Ivy
+    // since previously we did include unused files in the compilation
+    // See: https://github.com/angular/angular-cli/pull/15030
+    if (!this._compilerOptions.enableIvy) {
+      return;
+    }
+
+    const program = this._getTsProgram();
+    if (!program) {
+      return;
+    }
+
+    // Exclude the following files from unused checks
+    // - ngfactories & ngstyle might not have a correspondent
+    //   JS file example `@angular/core/core.ngfactory.ts`.
+    // - .d.ts files might not have a correspondent JS file due to bundling.
+    // - __ng_typecheck__.ts will never be requested.
+    const fileExcludeRegExp = /(\.(d|ngfactory|ngstyle)\.ts|ng_typecheck__\.ts)$/;
+
+    const usedFiles = new Set<string>();
+    for (const compilationModule of compilation.modules) {
+      if (!compilationModule.resource) {
+        continue;
+      }
+
+      usedFiles.add(forwardSlashPath(compilationModule.resource));
+
+      // We need the below for dependencies which 
+      // are not emitted such as type only TS files
+      for (const dependency of compilationModule.buildInfo.fileDependencies) {
+        usedFiles.add(forwardSlashPath(dependency));
+      }
+    }
+
+    const unusedFilesWarning = program.getSourceFiles()
+      .filter(({ fileName }) => !fileExcludeRegExp.test(fileName) && !usedFiles.has(fileName))
+      .map(({ fileName }) => `${fileName} is part of the TypeScript compilation but it's unused.`);
+
+    if (unusedFilesWarning.length) {
+      compilation.warnings.push(...unusedFilesWarning);
+    }
+  }
+
   // Registration hook for webpack plugin.
   // tslint:disable-next-line:no-big-function
   apply(compiler: Compiler & { watchMode?: boolean, parentCompilation?: compilation.Compilation }) {
@@ -609,6 +653,8 @@ export class AngularCompilerPlugin {
     // cleanup if not watching
     compiler.hooks.thisCompilation.tap('angular-compiler', compilation => {
       compilation.hooks.finishModules.tap('angular-compiler', () => {
+        this._warnOnUnusedFiles(compilation);
+
         let rootCompiler = compiler;
         while (rootCompiler.parentCompilation) {
           // tslint:disable-next-line:no-any
@@ -1174,7 +1220,7 @@ export class AngularCompilerPlugin {
           return null;
         }
       })
-      .filter(x => x);
+      .filter(x => x) as string[];
 
     const resourceImports = findResources(sourceFile)
       .map(resourcePath => resolve(dirname(resolvedFileName), normalize(resourcePath)));
@@ -1186,8 +1232,7 @@ export class AngularCompilerPlugin {
       ...this.getResourceDependencies(this._compilerHost.denormalizePath(resolvedFileName)),
     ].map((p) => p && this._compilerHost.denormalizePath(p)));
 
-    return [...uniqueDependencies]
-      .filter(x => !!x) as string[];
+    return [...uniqueDependencies];
   }
 
   getResourceDependencies(fileName: string): string[] {
--- a/packages/angular_devkit/build_angular/test/browser/unused-files-warning_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/unused-files-warning_spec_large.ts
@@ -1,0 +1,165 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Architect } from '@angular-devkit/architect';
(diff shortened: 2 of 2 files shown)


alan-agius4:
Note: tests to be added

Also without implementing this https://github.com/angular/angular-cli/pull/15030 for VE the warning will be shown all the time. 

This is because by default we do include redundant files in the compilations. There are;
- app.server.module.ts
- server.ts
- main.server.ts
- environment.prod.ts

Also, seeing this all current projects will have this warning.

clydin:
We may want to ignore anything involved in file replacements.

alan-agius4:
Only the last one, is in fileReplacement but, why should we ignore it? It is in the compilation because of the non strict includes that we currently have as otherwise they won’t be part of the compilation.

alan-agius4:
As discussed we should only enable this for Ivy.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -593,6 +596,35 @@ export class AngularCompilerPlugin {
...
+    //   JS file example `@angular/core/core.ngfactory.ts`.
+    // - .d.ts files might not have a corrispond JS file due to bundling.
+    // - __ng_typecheck__.ts will never be requested.
+    const fileExcludeRegExp = /(\.(d|ngfactory|ngstyle)\.ts|ng_typecheck__\.ts)$/;
+
+    const unusedFilesWarning = program.getSourceFiles()
+      .filter(({ fileName }) => !fileExcludeRegExp.test(fileName) && !this._usedSourceFiles.has(fileName))
+      .map(({ fileName }) => `${fileName} is part of the TypeScript compilation but it's unused.`);
+
+    if (unusedFilesWarning.length) {
```
alan-agius4: Note: there is also the option to add these warnings in the loader, using `this.emitWarning` though I found that those messages are more verbose.

```
WARNING in ./app/app.component.ts
Module Warning (from C:/git/angular-cli/packages/ngtools/webpack/src/index.ts):
Module is part of the TypeScript compilation but it's unused.
```

@filipesilva, what do you think?

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -1186,8 +1220,13 @@ export class AngularCompilerPlugin {
       ...this.getResourceDependencies(this._compilerHost.denormalizePath(resolvedFileName)),
     ].map((p) => p && this._compilerHost.denormalizePath(p)));
 
-    return [...uniqueDependencies]
-      .filter(x => !!x) as string[];
+    // Add used source files
+    this._usedSourceFiles.add(forwardSlashPath(fileName));
+    for (const f of esImports) {
+      this._usedSourceFiles.add(f);
```
filipesilva: Is it necessary to add imports as well here? I think `getDependencies` gets called for all TS files in the compilation anyway so sources should already be added on the previous call.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -1186,8 +1220,13 @@ export class AngularCompilerPlugin {
       ...this.getResourceDependencies(this._compilerHost.denormalizePath(resolvedFileName)),
     ].map((p) => p && this._compilerHost.denormalizePath(p)));
 
-    return [...uniqueDependencies]
-      .filter(x => !!x) as string[];
+    // Add used source files
+    this._usedSourceFiles.add(forwardSlashPath(fileName));
```
filipesilva: Somewhat reluctant on adding to the list here. I agree that it's a convenient place to hook into, but it adds somewhat unrelated things to the mental model for this call. 

Would this approach work as well if added to `getCompiledFile` instead (assuming adding imports is not needed either)?

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -609,6 +641,8 @@ export class AngularCompilerPlugin {
     // cleanup if not watching
     compiler.hooks.thisCompilation.tap('angular-compiler', compilation => {
       compilation.hooks.finishModules.tap('angular-compiler', () => {
+        this._warnOnUnusedFiles(compilation);
```
filipesilva: Should the list be cleared after the warning? 

I'm not sure how it would work with rebuilds though, because the loader might not request all files again. I'm not sure the unused files are removed from the compilation either.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -593,6 +596,35 @@ export class AngularCompilerPlugin {
     }
   }
 
+  private _warnOnUnusedFiles(compilation: compilation.Compilation) {
```
filipesilva: It looks like this warning will be shown on all builds, including rebuilds. Is this intentional? If so, we should also test on rebuilds to make sure it works correctly on adding and removing ts files.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -593,6 +596,35 @@ export class AngularCompilerPlugin {
...
+      return;
+    }
+
+    const program = this._getTsProgram();
+    if (!program) {
+      return;
+    }
+
+    // Exclude the following files from unused checks
+    // - ngfactories & ngstyle might not have a corrispond
```
filipesilva: Nit: `corrispond` -> `correspondent`

packages/ngtools/webpack/src/angular_compiler_plugin.ts
alan-agius4: Using `getCompiledFile` should do the trick

packages/ngtools/webpack/src/angular_compiler_plugin.ts
alan-agius4: Agreed

packages/ngtools/webpack/src/angular_compiler_plugin.ts
alan-agius4: Yeah, that's why I didn't clean it up after each build, because during rebuilds only the changed file will be requested.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
alan-agius4: That was the idea, to show on all builds. What do you think? 

If you agree, I will add some more tests to cover the mentioned test cases.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
alan-agius4: Just checked and `getCompiledFile` will not be called for all files, such as type only files.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
alan-agius4: agreeded

packages/ngtools/webpack/src/angular_compiler_plugin.ts
clydin: What about adding an additional Webpack compilation hook inside the Angular plugin that compares the set of TS source files with webpack modules?  Not sure of the appropriate hook off the top of my head though.

alan-agius4: That’s sound like a good idea I think the hook was called `finishedModules` or something like that. I’ll look into it tmr.

alan-agius4: Though thinking a bit I am not sure that this will work since I don't think type only files will be part of the webpack modules. Will double check tmr.

alan-agius4: @clydin Thinking a bit I am not sure that this will work since I don't think type only files will be part of the Webpack modules. Will double check tmr.

alan-agius4: @clydin I tried the above, but the problem is that type only files will not be part of the webpack modules.

alan-agius4: Managed to get the information needed through buildInfo.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
filipesilva: I think you marked this as resolved but ended up not changing it.

alan-agius4: Oops I’ll do it 😊

PR #15065, opened 2019-07-12
test: integrate browser testing for differential loading

clydin:
The following browsers are being tested against a production build of the CLI (includes differential loading, AOT, and optimization):

- Chrome: 41, 75
- Safari: 9.0, ~~10.1~~**, 11.1, 12.1
- Firefox: 48, 60, 68
- IE: 10, 11
- Edge: 14, 17, 18

** This browser is currently failing and disabled pending additional investigation.   Safari 10.1 has a javascript execution error and may be a legitimate failure.  See comments in code for some further details.

### DIFF
--- a/tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
+++ b/tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
@@ -1,0 +1,120 @@
+// @ts-check
+// Protractor configuration file, see link for more information
+// https://github.com/angular/protractor/blob/master/lib/config.ts
+
+const { SpecReporter } = require('jasmine-spec-reporter');
+
+const tunnelIdentifier = process.env['SAUCE_TUNNEL_IDENTIFIER'];
+
+/**
+ * @type { import("protractor").Config }
+ */
+exports.config = {
+  sauceUser: process.env['SAUCE_USERNAME'],
+  sauceKey: process.env['SAUCE_ACCESS_KEY'],
+
+  allScriptsTimeout: 11000,
+  specs: ['./src/**/*.e2e-spec.ts'],
+
+  multiCapabilities: [
+    {
+      browserName: 'chrome',
+      version: '41',
+      tunnelIdentifier,
+    },
+    {
+      browserName: 'chrome',
+      version: '75',
+      tunnelIdentifier,
+    },
+    {
+      browserName: 'safari',
+      platform: 'OS X 10.11',
+      version: '9.0',
+      tunnelIdentifier,
+    },
+    // TODO: Investigate. Failure:
+    // Failed: Error while running testForAngular: undefined is not an object (evaluating 'd.prototype[b].apply')
+    // {
+    //   browserName: 'safari',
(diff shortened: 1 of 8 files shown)


tests/legacy-cli/e2e/tests/misc/browsers.ts
```diff
@@ -0,0 +1,63 @@
...
+  await copyProjectAsset('protractor-saucelabs.conf.js', 'e2e/protractor-saucelabs.conf.js');
+
+  // Remove browser log checks as they are only supported with the chrome webdriver
+  await replaceInFile(
+    'e2e/src/app.e2e-spec.ts',
+    'await browser.manage().logs().get(logging.Type.BROWSER)',
+    '[]',
+  );
+
+  // Workaround defect in getText WebDriver implementation for Safari/Edge
```
alan-agius4: Why not update the E2E schematics directly? I'm guessing the bug is also present if someone tries to run the E2E in Safari/Edge directly right?

tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
```diff
@@ -0,0 +1,120 @@
...
+      tunnelIdentifier,
+    },
+    {
+      browserName: 'safari',
+      platform: 'OS X 10.11',
+      version: '9.0',
+      tunnelIdentifier,
+    },
+    // TODO: Investigate. Failure:
+    // Failed: Error while running testForAngular: undefined is not an object (evaluating 'd.prototype[b].apply')
```
filipesilva: Can you open an issue for this please?

tests/legacy-cli/e2e/tests/misc/browsers.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Leading and trailing space is not removed
+  await replaceInFile(
+    'e2e/src/app.e2e-spec.ts',
+    '\'should display welcome message\',',
+    '\'should display welcome message\', async',
```
filipesilva: I don't quite follow. How is adding `async`  related to trailing spaces?

tests/legacy-cli/e2e/tests/misc/browsers.ts
clydin: I felt that would be out of scope for this PR.

This is also a selenium/protractor deficiency which should really be fixed upstream.

The upcoming Protractor upgrade will also require a variety of changes for the E2E tests so I think it would be better to hold off on changes until then.

tests/legacy-cli/e2e/tests/misc/browsers.ts
clydin: The test is already pseudo-async.  The “getText” call returns a promise that Protractor _magically_ awaits.  I  normalized the test so it would be easier to debug.

tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
clydin: Done. #15084

PR #15081, opened 2019-07-15
feat(@angular/cli): known command version shortcut

stevermeister:
just to make it possible to run 

`$ ng -v` (with `-`)

that everybody got used to for all other tools


stevermeister:
Hm, one of the tests failed because:

```
error https://registry.yarnpkg.com/@bazel/bazel-darwin_x64/-/bazel-darwin_x64-0.24.1.tgz: Request failed "503 Service Unavailable"
```

mgechev:
The soonest we can release this would be version 9. I'll add it for further discussion.

vikerman:
`-v` is confusing because it is used to indicate verbose in some of our commands.

stevermeister:
I bet if you take stat it will be much more users who try to do `ng -v` than know what `verbose` option means

Issue #15053, opened 2019-07-11
scheduleBuilder should send a target when called with a context with a target


FrozenPandaz:
# 🚀 `scheduleBuilder` should send a `target` in the `context` when available

### Command (mark with an `x`)

```
- [ ] new
- [x] build
- [x] serve
- [x] test
- [x] e2e
- [ ] generate
- [ ] add
- [ ] update
- [x] lint
- [x] xi18n
- [x] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
Scheduling a child builder through a parent builder does not pass in the target of the parent builder

### Describe the solution you'd like
The child builder should receive the target of the parent builder when available.

.ie

`ng build app1` uses Builder: `@awesome/example:parent`.

`@awesome/example:parent` calls `scheduleBuilder('@awesome/example:child', {})`

`@awesome/example:child` should have `target: { project: 'app1', target: null }` in `context`.

> Note: `Target.target` is required at the moment. It could also be the target which the parent was called with but this is actually false. I can't think of a reason why someone would pass `context.target` into things like `scheduleTarget` or `getBuilderName` etc. but it is the same type. I guess those methods would need to handle the case where `Target.target` is falsy.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?
Perhaps we could add `Context.project` as that's all we really want?

PR #15082, opened 2019-07-15
Pass target to a builder called through `context.scheduleBuilder`

pbrezina:
A custom builder can call `context.scheduleBuilder` to call another builder.
However, the following call will fail with `Must either have a target from
the context or a default project.` exception because `scheduleBuilder` does
not pass `target` to the called builder.

```typescript
import { JsonObject } from '@angular-devkit/core'
import { BuilderContext, BuilderOutput, BuilderRun, createBuilder } from '@angular-devkit/architect';
import { Observable, from } from 'rxjs';
import { concatMap, map } from 'rxjs/operators';

export default createBuilder(_customBuilder);

function _customBuilder(options: JsonObject, context: BuilderContext): Observable<BuilderOutput> {
    const builder = '@angular-devkit/build-angular:browser';

    return from(context.scheduleBuilder(builder, options)).pipe(
        concatMap(run => run.result)
    );
}
```

This patch makes sure that parent target is passed to the called builder.

Resolves:
https://github.com/angular/angular-cli/issues/15053

### DIFF
--- a/packages/angular_devkit/architect/src/create-builder.ts
+++ b/packages/angular_devkit/architect/src/create-builder.ts
@@ -129,6 +129,7 @@ export function createBuilder<
           ) {
             const run = await scheduleByName(builderName, options, {
               scheduler,
+              target: scheduleOptions.target,
               logger: scheduleOptions.logger || logger.createChild(''),
               workspaceRoot: i.workspaceRoot,
               currentDirectory: i.currentDirectory,
--- a/packages/angular_devkit/architect/src/api.ts
+++ b/packages/angular_devkit/architect/src/api.ts
@@ -109,6 +109,11 @@ export interface ScheduleOptions {
    * to log a builder scheduled from your builder you should forward log events yourself.
    */
   logger?: logging.Logger;
+
+  /**
+   * Target to pass to the builder.
+   */
+  target?: Target;
 }
 
 /**


pbrezina:
I signed it!

alan-agius4:
Hi, can you please update the commit message to include the type and scope please?

```
fix(@angular-devkit/architect): ....
```

Also, you might need to rebase to make CI green.

Thanks.

pbrezina:
Done.

packages/angular_devkit/architect/src/create-builder.ts
```diff
@@ -129,6 +129,7 @@ export function createBuilder<
           ) {
             const run = await scheduleByName(builderName, options, {
               scheduler,
+              target: this.target,
```
clydin: This should not be default behavior.  In the example, the targets happen to align but this cannot be guaranteed.  `target` should be an option on the `ScheduleOptions` interface and default to undefined to maintain existing behavior.

pbrezina: I would argue that if you run, e.g. 'ng run project:my-target' then target should always be  `project:my-target`. If you run `ng build` target should always be `default-project:build`. So calling a builder from a custom build should in my opinion pass the target from the custom builder.

It can be made configurable, but in my opinion this should be the default behavior.

clydin: It's a breaking change to alter the default behavior.
Also a target references a specific configuration within a workspace file.  That configuration may have no relation to a builder that is manually scheduled within another.  Any number or types of builders may also be scheduled within one builder and a builder may act differently depending upon whether it is executed from a workspace target or not.

The exception that is attempting to be bypassed is also an implementation choice of the particular builder being scheduled (`@angular-devkit/build-angular:browser`) and not an issue with the system in a general case.

pbrezina: Ok, I put it to `ScheduleOptions` to make it configurable.

PR #15093, opened 2019-07-16
feat(@ngtools/webpack): show warning when large CSS are included in c…

alan-agius4:
…omponents

It’s very easy to inadvertently import toplevel css in component styles. Since component css is standalone and self-contained, it will never be shared between components and remains as a single large bundle for each component. This in turn adds a large amount of code that must be processed and increases bundle size.

Related to: TOOL-949

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -71,7 +71,7 @@ import {
   MESSAGE_KIND,
   UpdateMessage,
 } from './type_checker_messages';
-import { flattenArray, forwardSlashPath, workaroundResolve } from './utils';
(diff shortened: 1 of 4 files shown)


alan-agius4:
1) When the stylesUrl is an array should be sum of the total size of the styles or still check each file individually?
2) What should be the size limit?
3) Between dev and prod build the the size might varies quite a bit because of minifications and inline sourcemaps, how should be tackle this?

filipesilva:
I think the answers are:
1. individual css, because the problem is usually one inadvertently very large piece of css instead of several average sized ones
2. I don't know. Maybe 20kb? Perhaps we should roll this into the bundle budgets feature, and use a default there that users can change.
3. We should use a similar logic as bundle budgets so that the reasoning is familiar.

alan-agius4:
As suggested by yourself, let's try to use bundle budgets for this

PR #15105, opened 2019-07-17
feat(@angular/cli): implement deploy command

mgechev:
Implementation of the `ng deploy` command. Find the design document [here](https://docs.google.com/document/d/1qyVJUjoZEB3QHGBe82ogwUsnAKBv2lqWRbe8BafaGtQ/edit?usp=sharing).

// cc @vikerman @StephenFluin @clydin @alan-agius4 @filipesilva

### DIFF
--- a/packages/angular/cli/commands/deploy-impl.ts
+++ b/packages/angular/cli/commands/deploy-impl.ts
@@ -1,0 +1,49 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { ArchitectCommand, ArchitectCommandOptions } from '../models/architect-command';
+import { Arguments } from '../models/interface';
+import { Schema as DeployCommandSchema } from './deploy';
+
+const BuilderMissing = `
+
+Cannot find "deploy" target for the specified project.
+
+You should add a package that implements deployment capabilities for your
+favorite platform.
+
+For example:
+  ng add @angular/fire
+  ng add @azure/ng-deploy
+  ng add @zeit/ng-deploy
+
+Find more packages on npm https://www.npmjs.com/search?q=ng%20deploy
+`;
+
+export class DeployCommand extends ArchitectCommand<DeployCommandSchema> {
+  public readonly target = 'deploy';
+  public readonly missingTargetError = BuilderMissing;
+
+  public async run(options: ArchitectCommandOptions & Arguments) {
+    return this.runArchitectTarget(options);
+  }
+
+  public async initialize(options: DeployCommandSchema & Arguments): Promise<void> {
+    if (!options.help) {
+      return super.initialize(options);
+    }
+  }
+
+  async reportAnalytics(
+    paths: string[],
+    options: DeployCommandSchema & Arguments,
+    dimensions: (boolean | number | string)[] = [],
+    metrics: (boolean | number | string)[] = [],
+  ): Promise<void> {
+    return super.reportAnalytics(paths, options, dimensions, metrics);
+  }
+}
--- a/packages/angular/cli/commands/deploy.json
+++ b/packages/angular/cli/commands/deploy.json
@@ -1,0 +1,31 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "$id": "ng-cli://commands/deploy.json",
+  "description": "Invokes the deploy builder for a specified project. If no project is specified, the CLI will invoke the deploy builder for the default project in the workspace.",
+  "$longDescription": "./deploy-long.md",
+
+  "$aliases": [ "d" ],
+  "$scope": "in",
+  "$type": "architect",
+  "$impl": "./deploy-impl#DeployCommand",
+
+  "allOf": [
+    {
+      "properties": {
+        "project": {
+          "type": "string",
+          "description": "The name of the project to deploy.",
+          "$default": {
+            "$source": "argv",
+            "index": 0
+          }
+        }
+      },
+      "required": [
+      ]
+    },
+    {
+      "$ref": "./definitions.json#/definitions/base"
+    }
+  ]
+}
--- a/packages/angular/cli/models/architect-command.ts
+++ b/packages/angular/cli/models/architect-command.ts
@@ -35,6 +35,7 @@ export abstract class ArchitectCommand<
   protected multiTarget = false;
 
   target: string | undefined;
+  missingTargetError: string | undefined;
 
   public async initialize(options: T & Arguments): Promise<void> {
     await super.initialize(options);
@@ -75,11 +76,12 @@ export abstract class ArchitectCommand<
     }
 
     if (targetProjectNames.length === 0) {
-      throw new Error(`No projects support the '${this.target}' target.`);
+      throw new Error(this.missingTargetError || `No projects support the '${this.target}' target.`);
     }
 
     if (projectName && !targetProjectNames.includes(projectName)) {
-      throw new Error(`Project '${projectName}' does not support the '${this.target}' target.`);
+      throw new Error(this.missingTargetError ||
+        `Project '${projectName}' does not support the '${this.target}' target.`);
     }
 
     if (!projectName && commandLeftovers && commandLeftovers.length > 0) {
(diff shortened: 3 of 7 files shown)


packages/angular/cli/commands/deploy-impl.ts
```diff
@@ -0,0 +1,35 @@
...
+import { ArchitectCommand, ArchitectCommandOptions } from '../models/architect-command';
+import { Arguments } from '../models/interface';
+import { Schema as DeployCommandSchema } from './deploy';
+
+export class DeployCommand extends ArchitectCommand<DeployCommandSchema> {
+  public readonly target = 'deploy';
+  public readonly error = `Cannot find "deploy" target for the specified project.
+
+You can find a package that implements deployment capabilities for your
+favorite platform on npm: https://www.npmjs.com/search?q=ng%20deploy
```
StephenFluin: We should provide a short list of 2-3 examples. Maybe @angular/fire, azure, etc?

mgechev: What would be the recommendation criteria? We don't have first party deployment builders.

StephenFluin:
I think including the folks that we worked directly with for the initial
launch would be fair. Otherwise we could say "cloud providers of x
size/revenue/users".

scripts/validate-user-analytics.ts
```diff
@@ -26,6 +26,7 @@ const metricsTableRe = /<!--METRICS_TABLE_BEGIN-->([\s\S]*)<!--METRICS_TABLE_END
 function _exec(command: string, args: string[], opts: { cwd?: string }, logger: logging.Logger) {
   const { status, error, stdout } = spawnSync(command, args, {
     stdio: ['ignore', 'pipe', 'inherit'],
+    maxBuffer: 1024 * 1024,
```
filipesilva: This change should be in it's own commit.

packages/angular_devkit/build_angular/src/deploy/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { createBuilder } from "@angular-devkit/architect";
+
+const NotImplemented = `
+
+Cannot find "deploy" target for the specified project.
```
filipesilva: We could add this message directly on the CLI command instead of creating a dummy builder that is added to all new projects.

packages/angular_devkit/build_angular/builders.json
```diff
@@ -16,6 +16,11 @@
       "schema": "./src/dev-server/schema.json",
       "description": "Serve a browser app."
     },
+    "deploy": {
```
filipesilva: Adding a new builder to build-angular should be a separate feature commit.

packages/angular_devkit/build_angular/src/deploy/index.ts
mgechev: Without the architect builder the initialize phase of `ArchitectCommand` fails since it cannot find the target in `angular.json`.

filipesilva: Sure, but we can catch that error and show the message. It's ok to not have the default targets.

I don't quite agree with that future use of the dummy builder. The approach of using an existing builder to serve as an extension point to other builders subverts the current existing extension point which using builders themselves. With that proposed approach, a separately plugin approach would be needed just for deploy builders so that they could be used by the currently dummy builder.

packages/angular_devkit/build_angular/src/deploy/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+You should add a package that implements deployment capabilities for your
+favorite platform.
```
alan-agius4: Is it intentional that this is in another new line?

packages/angular_devkit/build_angular/src/deploy/index.ts
alan-agius4: Yeah, I agree with @filipesilva

packages/angular_devkit/build_angular/src/deploy/index.ts
mgechev: Yep

packages/angular_devkit/build_angular/src/deploy/index.ts
mgechev: Given the current implementation of the abstract `ArchitectCommand` it doesn't look like it's okay to not have default targets.

When someone invokes `ng deploy --help` the `initialize` method will fail because it'll not find a builder implementation.

Given the current constraints set by `ArchitectCommand.prototype.initialize`, it looks like that's the intended behavior. We can change it, sure.

Overriding the method in `DeployCommand`, catching the exception and handling it is an option but then we should reimplement part of the abstract `initialize`.

packages/angular_devkit/build_angular/builders.json
mgechev: Done

scripts/validate-user-analytics.ts
mgechev: Done

packages/angular_devkit/build_angular/src/deploy/index.ts
filipesilva: Sorry I wasn't very clear. What I meant was that it is ok for users to not have (our) default targets. For instance on `--minimal` projects the `test` and `e2e` targets aren't there and that should be handled on our side.

I haven't look at how hard it is to display a custom message for a missing target. But however hard it is I will be easier than introducing new public API (the dummy builder), requiring an update (adding the dummy builder to existing angular.json). So I think we should do it in a way that doesn't require those things, whether through error catching, refactoring the command initialization, or something else. @clydin might have ideas of how to implement this.

Maybe this just means that first party Architect commands that expect the presence of targets need to have better messages in general instead of telling users their angular config is bad.

packages/angular/cli/models/architect-command.ts
```diff
@@ -350,7 +352,8 @@ export abstract class ArchitectCommand<
         return allProjectsForTargetName;
       }
 
-      throw new Error(`Could not determine a single project for the '${targetName}' target.`);
+      throw new Error(this.missingTargetError ||
```
clydin: In this case there are multiple projects with the particular target and the target is a single instance target.  The missing error shouldn't apply here as it actually found several.

scripts/validate-user-analytics.ts
filipesilva: Yesterday I was looking at a similar problem in https://github.com/angular/angular-cli/pull/15176/commits/e4e04e24ac006c53f9644a4007133c080e7fd675 and now I know that `1024 * 1024` is actually the default. 

Is this change actually doing something? I guess only if `maxBuffer` was being passed as a smaller value in `opts` but I don't find `maxBuffer` referenced anywhere in scripts.

packages/angular/cli/commands/deploy.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "http://json-schema.org/schema",
+  "$id": "ng-cli://commands/deploy.json",
+  "description": "Invokes the deploy builder for a specified project. If no project is specified, the CLI will invoke the deploy builder for the default project in thw workspace.",
```
filipesilva: Nit: `thw` -> `the`

packages/angular/cli/models/architect-command.ts
```diff
@@ -74,12 +75,13 @@ export abstract class ArchitectCommand<
       }
     }
 
-    if (targetProjectNames.length === 0) {
-      throw new Error(`No projects support the '${this.target}' target.`);
+    if (!options.help && (targetProjectNames.length === 0)) {
```
filipesilva: Why does `!options.help` need to added to this condition and the one below it? If I call `ng deploy --help` this still looks like the error that shows up.

mgechev: Since we don't have a default dummy builder, the command is going to always throw when the user invokes `ng deploy --help`.

scripts/validate-user-analytics.ts
mgechev: Looks like a leftover.

mgechev:
As discussed today, there's some inconsistent behavior in the CLI in handling the `--help` option.

If we improve it, the implementation of `ng deploy` would be simpler and more consistent with the rest of the commands. Adding it for further discussion so we can sync on the work breakdown this Thursday :).

scripts/validate-user-analytics.ts
mgechev: Not a leftover. On node 10.x it's smaller so this change is required.

PR #15121, opened 2019-07-18
fix(@angular-devkit/build-angular): additional ES5 browser polyfills based on babel usage analysis

clydin:
Babel's usage based polyfill analyzer identified several additional functions which should be polyfilled for older browsers.


clydin:
Superseded by #15246

PR #15127, opened 2019-07-19
feat(@angular-devkit/build-angular): add bundle budget for component styles

alan-agius4:
It’s very easy to inadvertently import toplevel css in component styles. Since component css is standalone and self-contained, it will never be shared between components and remains as a single large bundle for each component. This in turn adds a large amount of code that must be processed and increases bundle size.

Note: this will only work for AOT compilations

Related to: TOOL-949

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/aot_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/aot_spec_large.ts
@@ -7,7 +7,7 @@
  */
 
 import { Architect } from '@angular-devkit/architect';
-import { join, normalize, virtualFs } from '@angular-devkit/core';
+import { join, logging, normalize, virtualFs } from '@angular-devkit/core';
 import { BrowserBuilderOutput } from '../../src/browser';
 import { createArchitect, host, ivyEnabled } from '../utils';
 
@@ -37,6 +37,30 @@ describe('Browser Builder AOT', () => {
       expect(content).toMatch(/platformBrowser.*bootstrapModuleFactory.*AppModuleNgFactory/);
     }
 
+    await run.stop();
+  });
+
+  it('shows warnings for component styles', async () => {
+    const overrides = {
+      aot: true,
+      optimization: true,
+    };
+
+    host.writeMultipleFiles({
+      'src/app/app.component.css': `
+        .foo { color: white; padding: 1px; };
+        .buz { color: white; padding: 2px; };
+      `,
+    });
+
+    const logger = new logging.Logger('');
+    const logs: string[] = [];
+    logger.subscribe(e => logs.push(e.message));
+
+    const run = await architect.scheduleTarget(targetSpec, overrides, { logger });
+    const output = await run.result;
+    expect(output.success).toBe(true);
+    expect(logs.join()).toContain('WARNING in Invalid selector');
     await run.stop();
   });
 });
--- a/packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
@@ -64,6 +64,68 @@ describe('Browser Builder bundle budgets', () => {
     await run.stop();
   });
 
+  it(`shows warnings for large component css when using 'anyComponentStyle' when AOT`, async () => {
+    const overrides = {
+      aot: true,
+      optimization: true,
+      budgets: [{ type: 'anyComponentStyle', maximumWarning: '1b' }],
+    };
+
+    const cssContent = `
+      .foo { color: white; padding: 1px; }
+      .buz { color: white; padding: 2px; }
+      .bar { color: white; padding: 3px; }
+    `;
+
+    host.writeMultipleFiles({
+      'src/app/app.component.css': cssContent,
+      'src/assets/foo.css': cssContent,
+      'src/styles.css': cssContent,
+    });
+
+    const logger = new logging.Logger('');
+    const logs: string[] = [];
+    logger.subscribe(e => logs.push(e.message));
+
+    const run = await architect.scheduleTarget(targetSpec, overrides, { logger });
+    const output = await run.result;
+    expect(output.success).toBe(true);
+    expect(logs.length).toBe(2);
+    expect(logs.join()).toMatch(/WARNING.+app\.component\.css/);
+    await run.stop();
+  });
+
+  it(`shows error for large component css when using 'anyComponentStyle' when AOT`, async () => {
+    const overrides = {
+      aot: true,
+      optimization: true,
+      budgets: [{ type: 'anyComponentStyle', maximumError: '1b' }],
+    };
+
+    const cssContent = `
+      .foo { color: white; padding: 1px; }
+      .buz { color: white; padding: 2px; }
+      .bar { color: white; padding: 3px; }
+    `;
+
+    host.writeMultipleFiles({
+      'src/app/app.component.css': cssContent,
+      'src/assets/foo.css': cssContent,
+      'src/styles.css': cssContent,
+    });
+
+    const logger = new logging.Logger('');
+    const logs: string[] = [];
+    logger.subscribe(e => logs.push(e.message));
+
+    const run = await architect.scheduleTarget(targetSpec, overrides, { logger });
+    const output = await run.result;
+    expect(output.success).toBe(false);
+    expect(logs.length).toBe(2);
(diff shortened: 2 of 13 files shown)


packages/angular_devkit/build_angular/test/browser/aot_spec_large.ts
```diff
@@ -39,4 +39,28 @@ describe('Browser Builder AOT', () => {
...
+    });
+
+    const logger = new logging.Logger('');
+    const logs: string[] = [];
+    logger.subscribe(e => logs.push(e.message));
+
+    const run = await architect.scheduleTarget(targetSpec, overrides, { logger });
+    const output = await run.result;
+    expect(output.success).toBe(true);
+    expect(logs.join()).toContain('WARNING in Invalid selector');
```
filipesilva: I don't understand what this warning means. Is it related to bundle budgets?

filipesilva: I don't understand what this warning means. Is it related to bundle budgets?

packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
```diff
@@ -64,6 +64,68 @@ describe('Browser Builder bundle budgets', () => {
     await run.stop();
   });
 
+  it(`shows warnings for large component css when using 'anyComponentStyle' when AOT`, async () => {
+    const overrides = {
+      aot: true,
+      optimization: true,
+      budgets: [{ type: 'anyComponentStyle', maximumWarning: '1b' }],
```
filipesilva: What's the plan for showing this warning overall? Will we update bundle budgets in 9 for all projects?

packages/angular_devkit/build_angular/test/browser/aot_spec_large.ts
alan-agius4: No it's related to the other fix which is required for the bundle budgets

https://github.com/angular/angular-cli/pull/15127/commits/39f1be02beb7391bfbb34cd4c342d97c94d8fb1c

packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
alan-agius4: I think the best option would be to do it for 9 for existing projects since users won't typically run migrations on minor and for new projects we can do it right away.

1) What should be the value `maximumWarning` and `maximumError` for new projects?
2) I am guessing that for the migration we won't want to add `maximumError` as this will break exiting projects.

packages/angular_devkit/build_angular/test/browser/aot_spec_large.ts
filipesilva: Discussed with Alan a bit more. In this example the semicolons are actually causing the `.buz` class to be dropped by the optimizer, but the user is not warned. When this goes in users with this sort of warning will start seeing them.

packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
filipesilva: Yeah I think it should be for 9 too. We should add the migration to this PR as well.

As for the default numbers, I'm not really sure. UTF-8 uses [1 to 4 bytes](https://stackoverflow.com/a/10229225/2116927) for each char, most commonly only 1 byte for US ASCII chars. So 1kb would be 1024 chars. Off the top of my head warning on 5kb and erroring on 20kb sounds within the realm of reasonable. @mgechev do you have a recommendation?

alan-agius4: Ill start working on the migrations, and than we just change the number when we have more feedback from @mgechev.

packages/schematics/angular/migrations/update-9/update-workspace-config.ts
```diff
@@ -121,3 +130,46 @@ function updateOption(property: 'scripts' | 'styles', recorder: UpdateRecorder,
...
+    let hasAnyComponentStyle = false;
+    for (const node of budgetOption.elements) {
+      if (!node || node.kind !== 'object') {
+        // skip non complex objects
+        continue;
+      }
+
+      const budget = findPropertyInAstObject(node, 'type');
+
+      // if lazy was not true, it is redundant hence, don't add it
```
filipesilva: Nit: I don't think this comment is related to the code below.

packages/schematics/angular/utility/json-utils.ts
```diff
@@ -191,3 +191,24 @@ export function findPropertyInAstObject(
 function _buildIndent(count: number): string {
   return count ? '\n' + ' '.repeat(count) : '';
 }
+
+function _strigifyContent(value: JsonValue, indentStr: string): string {
+  // TODO: Add snapshot tests
```
filipesilva: What do you mean by this?

alan-agius4: We should add golden files/snapshots tests for json ast manipulation to confirm that the indentation inserted is correct when modifying a file.

Currently it’s really hard to understand if the indentation is correct and also test the indentations itself.

Haven’t golden files/snapshot tests will be more straightforward but Ofc we need to set this up.

clydin: That function is spelled wrong btw.

We should really drop these when working with the workspace but we'll need to fix migrations first so that we can use the latest APIs inside a migration.

alan-agius4: Oops thanks for pointing that out.

packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
mgechev: A good source for setting up good defaults will be http://bit.ly/perf-budget-calculator

I'd recommend bumping the error limit up with few extra 100s of KBs.

alan-agius4: @mgechev the budgets will be per component.

filipesilva: As an example, in https://github.com/johnpapa/awesome-angular-workshop/ several components included 70kb of minified styles each. This happened because they included the whole material theme.

mgechev: Yes, missed the fact it's per component. Budget for component is definitely not a complete solution but sounds like a good start.

Here's a very rough estimation based on https://bit.ly/perf-budget-calculator

100KB from the calculator is very roughly 20 components (totally making this up)
20KB / 20 components = 1KB
Bumping this up with 5KB

~6KB default warning
~10KB default error

filipesilva: @alan-agius4 do we have a way forward with a global css bundle budget? Not asking to have it included in this PR, just asking if the changes in this PR would get in the way of that. I imagine a `anyGlobalCss` budget or something.

alan-agius4: Global styles meaning styles in general or that are set globally ie imported via style.css?

For the later we already have the `AnyCalculator` which can be used to set budgets on global css.

However, if we'd want to add a budget for css within the component + global css we can do it and this PR will provide a way forward for that.

My concerns would be that if want to sum up css from all the components and global stylesheet, you might end up with a misleading budget. Let's say the result of total CSS is 500Kb, when having lazy loading this will never be loaded/parsed at once.

mgechev: Dynamic budgets, such as LightWallet would be great in such scenario. In the same time, we might be able to estimate what would be loaded initially, right? In webpack we have this metadata in the compilation stats.

alan-agius4: But this is indeed is  the enabler for that :)

alan-agius4: We already have some budgets for initial chunks https://github.com/angular/angular-cli/blob/47c0db9eb394640501af5f0fd36495547466bc37/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts#L62

PR #15130, opened 2019-07-19
Update karma-debug.html with <base href="/"> tag, as in the karma-context.html

UliOnline:
same fix as in https://github.com/angular/angular-cli/pull/12889 , but for the karma debug context.
Issue described in the above PR happens again, if you switch to the karma debug runner. Karma will now use the karma-debug.html instead of the karma-context.html, rendering the initial fix useless.


UliOnline:
hmm i just realized, that this change breaks the karma debug runner. clicking on a component test now triggers all tests again instead of only this test, probably because the karma html page relies on base href to point to "/debug.html". So a more sophisticated solution is required here.

PR #15142, opened 2019-07-22
fix(@schematics/angular): change `findNodes` to stop recursive lookup…

### LINKED ISSUE #15117: Generate module command adds route path to wrong line
# 🐞 Bug report

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```
### Description
`ng generate module ranking --route ranking --module app.module` command is working very well. But if there is a data property to routes in `app-routing.module`. The command add route path to wrong line.

## 🔬 Minimal Reproduction
`ng new example-app --routing`
`ng generate module ranking --route ranking --module app.module`

Add data property to ranking route, then my app-routing.module.ts like below
```ts
import { Routes, RouterModule } from '@angular/router';

const routes: Routes = [
  {
    path: 'ranking',
    loadChildren: () => import('./ranking/ranking.module').then(m => m.RankingModule),
    data: { path: 'test', name: 'Test' },
  },
];

@NgModule({
  imports: [RouterModule.forRoot(routes)],
  exports: [RouterModule],
})
export class AppRoutingModule {}
```

After, run the following command
`ng generate module home --route home --module app.module`

It adds my home route path to the wrong line like below:
```ts
const routes: Routes = [
  {
    path: 'ranking',
    loadChildren: () => import('./ranking/ranking.module').then(m => m.RankingModule),
    data: { path: 'test', name: 'Test' },
    { path: 'home', loadChildren: () => import('./home/home.module').then(m => m.HomeModule) },
  },
];
```

alan-agius4:
… for child nodes of kind

Curtrently, when a node of kind is found, we recursivly continue to look up it's child nodes until the end of the AST. This ends up returing other nodes which we were not looking for as typically we are looking for the first level of children of the specified kind.

By default now, we stop recursivly looking for child nodes of kind when we encounter one.

Closes #15117

### DIFF
--- a/packages/schematics/angular/utility/ast-utils.ts
+++ b/packages/schematics/angular/utility/ast-utils.ts
@@ -93,9 +93,11 @@ export function insertImport(source: ts.SourceFile, fileToEdit: string, symbolNa
  * @param node
  * @param kind
  * @param max The maximum number of items to return.
+ * @param recursive Continue looking for nodes of kind recursive until end
+ * the last child even when node of kind has been found.
  * @return all nodes of kind, or [] if none is found
  */
-export function findNodes(node: ts.Node, kind: ts.SyntaxKind, max = Infinity): ts.Node[] {
+export function findNodes(node: ts.Node, kind: ts.SyntaxKind, max = Infinity, recursive = false): ts.Node[] {
   if (!node || max == 0) {
     return [];
   }
@@ -105,7 +107,7 @@ export function findNodes(node: ts.Node, kind: ts.SyntaxKind, max = Infinity): t
     arr.push(node);
     max--;
   }
-  if (max > 0) {
+  if (max > 0 && (recursive || node.kind !== kind)) {
     for (const child of node.getChildren()) {
       findNodes(child, kind, max).forEach(node => {
         if (max > 0) {
--- a/packages/schematics/angular/utility/ast-utils_spec.ts
+++ b/packages/schematics/angular/utility/ast-utils_spec.ts
@@ -479,6 +479,41 @@ describe('ast utils', () => {
       );
     });
 
+    it('should add a route to the routes to the correct array when having nested object literal', () => {
+      const moduleContent = `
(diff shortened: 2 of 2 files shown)


packages/schematics/angular/utility/ast-utils.ts
```diff
@@ -93,9 +93,11 @@ export function insertImport(source: ts.SourceFile, fileToEdit: string, symbolNa
  * @param node
  * @param kind
  * @param max The maximum number of items to return.
+ * @param recursive Continue looking for nodes of kind recursive until end
+ * the last child even when node of kind has been found.
  * @return all nodes of kind, or [] if none is found
  */
-export function findNodes(node: ts.Node, kind: ts.SyntaxKind, max = Infinity): ts.Node[] {
+export function findNodes(node: ts.Node, kind: ts.SyntaxKind, max = Infinity, recursive = false): ts.Node[] {
```
clydin: Should the default be the opposite so that existing calls retain the same behavior?  Or was that problematic for the other callers as well?

alan-agius4: Yeah, this is problematic in other calls like `insertAfterLastOccurrence`. Since when have nested nodes of the same kind. Example: `[[]], {a: {}}` the nodes returned will not be the correct once.

I don't thing we ever would want to get the recursive nested nodes of a kind (I might be wrong). When for example we are looking for `ObjectLiteralExpression` we would typically want the "direct" `ObjectLiteralExpression` and the the direct `ObjectLiteralExpression` and the nested ObjectLiteralExpression .

alan-agius4: I was actually tempted to remove the recursion for good.

clydin: Makes sense.  Just wanted to make sure.

PR #15154, opened 2019-07-24
redirect to index.html if root is called directly

CSchulz:
Followup for https://github.com/angular/angular-cli/commit/01247c9339d928cac1291f268ad17ee3a153c057#diff-6c0b312b8bd33db747e50549a66f7ab1

Restores the old behavior, if anyone calls the root directly it redirects to the index.html.

/cc @ajafff


clydin:
However, the current behavior is by design.  For further details, please see the discussion in issue https://github.com/angular/angular-cli/issues/16179.

PR #15164, opened 2019-07-25
fix(@angular-devkit/build-angular): rollback sass to version `1.22.7`

alan-agius4:
sass version `1.22.8` was rolled back and is no longer available on NPM

Supersedes https://github.com/angular/angular-cli/pull/15162


filipesilva:
For context, here is the issue in sass: https://github.com/sass/dart-sass/issues/780

PR #15165, opened 2019-07-25
fix(@angular-devkit/build-optimizer): don't wrap enum like nodes whic…

### LINKED ISSUE #15145: @angular-devkit/build-angular 0.802.0-next.1 throws ERROR NullInjectorError: StaticInjectorError
# 🐞 Bug report

### Command
- [x] build

### Is this a regression?
It was working fine in `"@angular-devkit/build-angular": "0.802.0-next.0",` and is broken in `"@angular-devkit/build-angular": "0.802.0-next.1",`.

 Yes, the previous version in which this bug was not present was: 
`"@angular-devkit/build-angular": "0.802.0-next.0"`

### Description
After upgrading my application from `"@angular-devkit/build-angular": "0.802.0-next.0",` to `"@angular-devkit/build-angular": "0.802.0-next.1",`. The application is throwing below error:

```
ERROR NullInjectorError: StaticInjectorError(a)[D -> Q]: 
  StaticInjectorError(Platform: core)[D -> Q]: 
    NullInjectorError: No provider for Q!
    at oe.get 

```

With sourcemaps enabled, it points to below code:

```
() => navigator.serviceWorker.register(script, { scope: options.scope })
            .catch((/**
         * @param {?} err
         * @return {?}
         */
        err => console.error('Service worker registration failed with:', err)))));
```

## 🔥 Exception or Error
<pre><code>
ERROR NullInjectorError: StaticInjectorError(a)[D -> Q]: 
  StaticInjectorError(Platform: core)[D -> Q]: 
    NullInjectorError: No provider for Q!
    at oe.get 
</code></pre>

## 🌍 Your Environment
<pre><code>
Package                                    Version
--------------------------------------------------------------------
@angular-devkit/architect                  0.802.0-next.1
@angular-devkit/build-angular              0.802.0-next.1
@angular-devkit/build-optimizer            0.802.0-next.1
@angular-devkit/build-webpack              0.802.0-next.1
@angular-devkit/core                       8.2.0-next.1
@angular-devkit/schematics                 8.2.0-next.1
@angular/cdk                               8.1.1
@angular/cli                               8.2.0-next.1
@angular/fire                              5.2.1
@angular/flex-layout                       8.0.0-beta.26
@angular/http                              8.0.0-beta.9
@angular/material                          8.1.1
@angular/material-moment-adapter           8.1.1
@angular/pwa                               0.802.0-next.1
@ngtools/webpack                           8.2.0-next.1
@nguniversal/module-map-ngfactory-loader   8.1.1
@schematics/angular                        8.2.0-next.1
@schematics/update                         0.802.0-next.1
rxjs                                       6.5.2
typescript                                 3.4.5
webpack                                    4.35.3
</code></pre>

alan-agius4:
…h are inside of methods.

With this change we stop recursive lookup when the current node is not a BlockLike.

This change should also improve the BO overall speed as it's reduces a lot of recursive lookups.

Fixes #15145

Note: Tested also on AIO to confirm size regression and didn't see any increase in file size in both ES5 and ES2015 builds

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/transforms/wrap-enums.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/wrap-enums.ts
@@ -56,7 +56,7 @@ function visitBlockStatements(
           return node;
       }
     } else {
-      return ts.visitEachChild(node, visitor, context);
+      return node;
     }
   };
 
@@ -112,7 +112,7 @@ function visitBlockStatements(
             oIndex++;
           }
 
-          const enumStatements = findStatements(name, statements, oIndex + 1);
+          const enumStatements = findStatements(name, statements, oIndex, 1);
           if (!enumStatements) {
             continue;
           }
@@ -316,6 +316,7 @@ function findStatements(
   name: string,
   statements: ts.NodeArray<ts.Statement>,
   statementIndex: number,
+  offset = 0,
 ): ts.Statement[] | undefined {
   let count = 1;
 
@@ -369,7 +370,7 @@ function findStatements(
   }
 
   if (count > 1) {
-    return statements.slice(statementIndex, statementIndex + count);
+    return statements.slice(statementIndex + offset, statementIndex + count);
   }
 
   return undefined;
--- a/packages/angular_devkit/build_optimizer/src/transforms/wrap-enums_spec.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/wrap-enums_spec.ts
@@ -88,6 +88,37 @@ describe('wrap enums and classes transformer', () => {
       expect(tags.oneLine`${transform(input)}`).toEqual(tags.oneLine`${output}`);
     });
 
+    it('should not wrap enum like which are inside of methods', () => {
+      const input = tags.stripIndent`
+        class LayoutDirective {
+          constructor(elRef) { }
+
+          applyStyleToElement(element, style, value) {
+            let styles = {};
+            if (typeof style === 'string') {
+                styles[style] = value;
+                style = styles;
+            }
+            styles = this.layoutConfig.disableVendorPrefixes ? style : applyCssPrefixes(style);
+            this._applyMultiValueStyleToElement(styles, element);
+          }
+        }
+        LayoutDirective.ctorParameters = () => [
+          { type: ElementRef }
+        ];
+      `;
+
+      const output = tags.stripIndent`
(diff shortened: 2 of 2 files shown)


packages/angular_devkit/build_optimizer/src/transforms/wrap-enums.ts
```diff
@@ -112,11 +112,16 @@ function visitBlockStatements(
...
-          const enumStatements = findStatements(name, statements, oIndex + 1);
+          const enumStatements = findStatements(name, statements, oIndex);
           if (!enumStatements) {
             continue;
           }
 
+          // first item is typically the const, var or export which is not needed
+          // as it will be created later on
+          // export var ChangeDetectionStrategy = {};
+          enumStatements.shift();
```
clydin: Why the shift instead of the plus one above?

alan-agius4: Because in `findStatements` there is already a `+1` https://github.com/angular/angular-cli/blob/66df0e77004d626b8aef5ade792373902b640b73/packages/angular_devkit/build_optimizer/src/transforms/wrap-enums.ts#L327 and adding a +1 to the above would skip a statement. example


Having these statements
```
VariableStatement <-- current
IfStatement <-- findStatements will start iterating from this. 
ExpressionStatement <-- (if we add +1 to oIndex) it will skip to this statement, which cause incorrect encapsulations as we are not validating that the above statement is valid and should be wrapped.
ExpressionStatement
```

clydin: In that case, why do we need the shift?  It should already be past the variable statement.

alan-agius4: Cause `findStatements` will iterate over the next statements to validate them. But will also return the statement at `oIndex`

Maybe can do something like `findStatements(name, statements, oIndex, offset);`?

clydin: That'll work.
This is the change that actually fixes the referenced issue, correct?
With the visit change will this actually be tested by the newly added test?

alan-agius4: > This is the change that actually fixes the referenced issue, correct?

Yes
> With the visit change will this actually be tested by the newly added test?

Added test

PR #15176, opened 2019-07-26
build: update to TS 3.5

filipesilva:


packages/schematics/angular/utility/latest-versions.ts
```diff
@@ -11,7 +11,7 @@ export const latestVersions = {
   Angular: '~8.2.0-next.2',
   RxJs: '~6.4.0',
   ZoneJs: '~0.9.1',
-  TypeScript: '~3.4.3',
+  TypeScript: '~3.5.3',
```
alan-agius4: Can you also bump tsickle pls?

PR #15186, opened 2019-07-29
Ivy libraries and workspace

alan-agius4:
- **feat(@schematics/angular): introduce Ivy workspace concept**

At the moment, to enable Ivy as the view rendering engine we are adding `enableIvy: true` in each of the applications tsconfigs. 
While this works this is not ideal when users want to opt-out, as users will need to modify multiple tsconfig. 

With the introduction of Ivy libraries for development this will be even harder and cumbersome.

We added the `enableIvy` option to the `workspace` schematic and remove the option for the `application` schematic.

`ng new <project-name> --enable-ivy` will pass down down the option the workspace schematic and an Ivy workspace will be created. 
Other schematics such as `application` and `library` will read the workspace (root) tsconfig to verify if a application or library with
 Ivy structure and configuration should be created.

- **feat(@schematics/angular): introduce Ivy libraries for development**

Since `NGCC` is non incremental and in library projects we have the original TS sources
we don't need to build a library using the `VE` and transform it using `NGCC`. Instead we can build the library using `NGTSC` (Ivy) directly
 as this enables faster incremental compilations and a better development experience.

 Libraries now have a `production` configuration, which enabled `VE` compilations. As it is not recommended to publish NGTSC (Ivy) 
 built libraries to NPM repositories, since Ivy libraries are not backwards compatible with the legacy View Engine.

- **feat(@schematics/angular): ivy library migration**

Add a migration to migrate existing libraries to the new library layout considering it will be the default in version 9.

### DIFF
--- a/packages/angular_devkit/build_ng_packagr/src/build/index_spec_large.ts
+++ b/packages/angular_devkit/build_ng_packagr/src/build/index_spec_large.ts
@@ -9,11 +9,12 @@ import { Architect } from '@angular-devkit/architect';
 import { WorkspaceNodeModulesArchitectHost } from '@angular-devkit/architect/node';
 import { TestProjectHost, TestingArchitectHost } from '@angular-devkit/architect/testing';
 import {
+  experimental,
+  getSystemPath,
   join,
   normalize,
   schema,
   virtualFs,
-  workspaces,
 } from '@angular-devkit/core'; // tslint:disable-line:no-implicit-dependencies
 import { map, take, tap } from 'rxjs/operators';
 
@@ -33,15 +34,14 @@ describe('NgPackagr Builder', () => {
     const registry = new schema.CoreSchemaRegistry();
     registry.addPostTransform(schema.transforms.addUndefinedDefaults);
 
-    const { workspace } = await workspaces.readWorkspace(
-      host.root(),
-      workspaces.createWorkspaceHost(host),
-    );
+    const workspaceSysPath = getSystemPath(host.root());
+    const workspace = await experimental.workspace.Workspace.fromPath(host, host.root(), registry);
     const architectHost = new TestingArchitectHost(
-      host.root(),
-      host.root(),
-      new WorkspaceNodeModulesArchitectHost(workspace, host.root()),
+      workspaceSysPath,
+      workspaceSysPath,
+      new WorkspaceNodeModulesArchitectHost(workspace, workspaceSysPath),
     );
+
     architect = new Architect(architectHost, registry);
   });
 
--- a/packages/angular_devkit/build_ng_packagr/src/build/index_spec_large.ts
+++ b/packages/angular_devkit/build_ng_packagr/src/build/index_spec_large.ts
@@ -18,10 +18,16 @@ import {
 } from '@angular-devkit/core'; // tslint:disable-line:no-implicit-dependencies
 import { map, take, tap } from 'rxjs/operators';
 
+export const ivyEnabled = process.argv.includes('--ivy');
+if (ivyEnabled) {
+  // tslint:disable-next-line:no-console
+  console.warn('********* IVY Enabled ***********');
+}
+
 const devkitRoot = (global as unknown as { _DevKitRoot: string})._DevKitRoot;
 const workspaceRoot = join(
   normalize(devkitRoot),
-  'tests/angular_devkit/build_ng_packagr/ng-packaged/',
+  `tests/angular_devkit/build_ng_packagr/ng-packaged${ivyEnabled ? '-ivy' : ''}/`,
 );
 
 describe('NgPackagr Builder', () => {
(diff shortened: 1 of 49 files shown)


packages/schematics/angular/application/schema.json
```diff
@@ -19,12 +19,6 @@
       },
       "x-prompt": "What name would you like to use for the application?"
     },
-    "enableIvy": {
```
alan-agius4: Since this is experimental, I don't think we need to deprecate.

docs/design/analytics.md
```diff
@@ -48,7 +48,6 @@ Note: There's a limit of 20 custom dimensions.
 | 5 | `Flag: --style` | `string` |
 | 6 | `--collection` | `string` |
 | 7 | `--buildEventLog` | `boolean` |
-| 8 | `Flag: --enableIvy` | `boolean` |
```
alan-agius4: Need to inform @StephenFluin about this, if we are going to remove it.

alan-agius4:
@filipesilva, we should have some documentation on AIO and make it clearer not to publish Ivy libraries to NPM.

packages/angular_devkit/build_ng_packagr/src/build/index_spec_large.ts
```diff
@@ -18,10 +18,16 @@ import {
 } from '@angular-devkit/core'; // tslint:disable-line:no-implicit-dependencies
 import { map, take, tap } from 'rxjs/operators';
 
+export const ivyEnabled = process.argv.includes('--ivy');
```
filipesilva: I don't think this ever runs. We use a glob for the ivy test large:
https://github.com/angular/angular-cli/blob/607e6f7cfd7b5f16231eae968fbf36e920c169e0/.circleci/config.yml#L340-L345

Can you alter the glob to include build_ng_packagr too?

packages/schematics/angular/application/schema.json
filipesilva: Yeah I think that's ok.

packages/angular_devkit/build_ng_packagr/src/build/index_spec_large.ts
alan-agius4: Good catch!

docs/design/analytics.md
alan-agius4: But the flag itself no longer exists on `ng generate application`, hence there is no data to collect. 

`enableIvy` exists on `ng new` and `ng generate workspace`

filipesilva: Yes, that's true. This file doesn't mention anything about what commands capture it though.

alan-agius4: yeah, I will add this to `ng-new` to that this matrix is captured there.

alan-agius4: https://github.com/angular/angular-cli/pull/15186/files#diff-e1a844356231f48df29c54e39fcb9951R25

packages/schematics/angular/application/schema.json
clydin: Don't we still need this option?  The migration won't run until 9 so the option won't be at the root tsconfig yet.

packages/schematics/angular/utility/workspace.ts
```diff
@@ -88,3 +89,28 @@ export async function createDefaultPath(tree: Tree, projectName: string): Promis
 
   return buildDefaultPath(project);
 }
+
+/**
+ * Returns `true` when the workspace root tsconfig is set to use Ivy
+ * as the Angular rendering engine.
+ */
+export function isIvyWorkspace(tree: Tree): boolean {
```
clydin: I'm not sure this is the correct file for this function since its not related to reading/writing the workspace file.

The function will also not handle extended configurations which would make it problematic to generate an application with ivy enabled in such a scenario.

alan-agius4:
@clydin 

> Don't we still need this option? The migration won't run until 9 so the option won't be at the root tsconfig yet.

No it's not needed any more, since now we have the concept of an ivy workspace instead of an Ivy application.

Though that is a good question about existing workspaces which are not Ivy but want to generate an Ivy application. Should this be supported?

//cc @filipesilva

packages/schematics/angular/utility/workspace.ts
alan-agius4: Is it common to have extended root level tsconfig? (Not application specific)?, pretty much in our schematics we only handle `/tsconfig.json` as a base for other tsconfigs.

Re the file, any thoughts were it so go?

clydin:
There's also the situation of existing ivy generated workspaces that are not configured as _ivy workspaces_.

filipesilva:
I think it's not a deal breaker if people have to manually set `enableIvy`. Realistically speaking you don't make apps/libs very often. And when you do, you mostly want to use the same compiler. I don't think many people would have to set it manually.

alan-agius4:
Superseded by https://github.com/angular/angular-cli/pull/15320

PR #15207, opened 2019-07-30
 fix(@schematics/angular): the `ng-new` schematic should not prompt for style nor for routing

### LINKED ISSUE #13383: Unnecessary prompts for ng new when using create application false (i.e. Conditional Prompts)

ahasall:
closes #13383

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -307,6 +307,10 @@ export default function (options: ApplicationOptions): Rule {
     }
     validateProjectName(options.name);
     options.prefix = options.prefix || 'app';
+    // This line and the one above shouldn't be needed, but at the moment they are.
+    // This is because the default value defined in the schema.json file is not
+    // correctly set when this schematic is run from another schematic.
+    options.style = options.style || Style.Css;
     const appRootSelector = `${options.prefix}-root`;
     const componentOptions: Partial<ComponentOptions> = !options.minimal ?
       {
--- a/packages/schematics/angular/ng-new/schema.json
+++ b/packages/schematics/angular/ng-new/schema.json
@@ -112,31 +112,7 @@
     "style": {
       "description": "The file extension or preprocessor to use for style files.",
       "type": "string",
-      "default": "css",
       "enum": ["css", "scss", "sass", "less", "styl"],
-      "x-prompt": {
-        "message": "Which stylesheet format would you like to use?",
-        "type": "list",
-        "items": [
-          { "value": "css", "label": "CSS" },
-          {
-            "value": "scss",
-            "label": "SCSS   [ https://sass-lang.com/documentation/syntax#scss                ]"
-          },
-          {
-            "value": "sass",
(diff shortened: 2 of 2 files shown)


packages/schematics/angular/application/index.ts
```diff
@@ -307,6 +307,7 @@ export default function (options: ApplicationOptions): Rule {
     }
     validateProjectName(options.name);
     options.prefix = options.prefix || 'app';
+    options.style = options.style || Style.Css;
```
alan-agius4: Is this need? Shouldn't it get the default which is set in the schema?

ahasall: I think it is needed.
If we call `ng generate application` directly it gets the default value. But it doesn't get it when called via `ng new`. Maybe there is a bug somewhere else.

I think this is why we also have `options.prefix = options.prefix || 'app';`

clydin: It technically shouldn't be needed but currently it might.
Outside of this PR, this should be investigated, fixed if it is indeed broken, and those two lines removed.

ahasall: Yes. Should this PR be on hold until the other issue is fixed?

clydin: I don't think so.  The line above it is already there.  Can you add a comment explaining the situation?

ahasall: Added a comment. Let me know if something else needs to be changed/added.

packages/schematics/angular/application/index.ts
```diff
@@ -307,6 +307,10 @@ export default function (options: ApplicationOptions): Rule {
     }
     validateProjectName(options.name);
     options.prefix = options.prefix || 'app';
+    // This line and the one above shouldn't be needed, but at the moment they are.
+    // This is because the default value defined in the schema.json file is not
+    // correctly set when this schematic is run from another schematic.
```
Splaktar: We should probably open an issue and add a link to it here w/ a TODO to remove this when that issue is resolved.

PR #15238, opened 2019-08-02
fix(@angular-devkit/build-angular): never use component css sourcemap…

filipesilva:
… when optimizations are on.

It will just increase bundle size without offering good debug experience.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
@@ -193,7 +193,13 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
         options: {
           ident: 'embedded',
           plugins: postcssPluginCreator,
-          sourceMap: cssSourceMap && !buildOptions.sourceMap.hidden ? 'inline' : false,
+          sourceMap: cssSourceMap
+            // Never use component css sourcemap when style optimizations are on.
+            // It will just increase bundle size without offering good debug experience.
+            && !buildOptions.optimization.styles
+            // Inline all sourcemap types except hidden ones, which are the same as no sourcemaps
+            // for component css.
+            && !buildOptions.sourceMap.hidden ? 'inline' : false,
         },
       },
       ...(use as webpack.Loader[]),


filipesilva:
According to @IgorMinar:

> If you are building the app with build build optimizations on, we should not inline any sourcemaps at all.
>
> The inline sourcemaps are useful for dev mode and debugging runtime issue, but don't help prod debugging easier in a way that you couldn't use devmode for and only add a lot of weight.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
```diff
@@ -193,7 +193,13 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
         options: {
           ident: 'embedded',
           plugins: postcssPluginCreator,
-          sourceMap: cssSourceMap && !buildOptions.sourceMap.hidden ? 'inline' : false,
+          sourceMap: cssSourceMap
+            // Never use component css sourcemap when optimizations are on.
+            // It will just increase bundle size without offering good debug experience.
+            && !buildOptions.optimization
```
alan-agius4: You should actually check it style optimization is enabled. As styles optimization can be disabled.

filipesilva: You're right, and that's also what was breaking a server builder test. Thanks!

clydin:
Actually, any reason we can't create file source maps for the component stylesheets?

filipesilva:
I'm not sure if the if the AOT compiler supports non-inline css sourcemaps actually. @IgorMinar @alxhub should we use them?

clydin:
Would it matter at the AOT level?  It's just a different source map comment url.

filipesilva:
I vaguely remember it mattered for AOT debugging, but not sure anymore. I think that's why we always inline them.

alan-agius4:
I think the problem was that when not inlining it didn’t work for a number of use cases.

- Components styles sourcemaps didn’t work on SSR unless inlined.
- I don’t think that client bundles can have multiple sourcemaps urls.

clydin:
The bundles wouldn't.  The injected style elements would each have one.  Instead of a data URL they would have a file URL.

PR #15243, opened 2019-08-03
test: fix failing prod build snapshots

alan-agius4:
The output of a prod build contains now:
```
bootstrapModuleFactory($l).catch(e=>console.error(e))}
```

Which is not being matched with the previously defined RegExp


alan-agius4:
Green snapshots build https://circleci.com/gh/angular/angular-cli/76721

PR #15246, opened 2019-08-05
Differential Loading via Bundle Downleveling

clydin:
This change removes the requirement to perform two full builds of the application to create a differential loaded application.  Instead only the ES2015 application is built and then the resulting application bundles are then downleveled directly.  This has the advantage of reducing the build time as well as a small size improvement in the ES5 bundles.  Lazy loaded bundles are also consistently named between the ES2015 and ES5 bundles.  Speed improvements range from 10% to 50% with larger applications benefiting the most.

The existing behavior is currently retained via the use of the `NG_BUILD_DIFFERENTIAL_FULL` environment variable.  The existing behavior is available at this point in time to ensure that any unknown problematic edge cases within the new build pipeline will not cause existing applications to fail to build.


package.json
```diff
@@ -131,7 +128,7 @@
     "rxjs": "~6.4.0",
     "sauce-connect": "https://saucelabs.com/downloads/sc-4.5.4-linux.tar.gz",
     "semver": "6.3.0",
-    "source-map": "^0.5.6",
+    "source-map": "^0.7.3",
```
alan-agius4: With this change we can close https://github.com/angular/angular-cli/issues/13228 and remove `source-map from the ignore list https://github.com/angular/angular-cli/blob/6f0e7bf437ae98d8d077c1f056f210fc407586f0/renovate.json#L21`

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/index-file/augment-index-html.ts
```diff
@@ -150,7 +150,10 @@ export async function augmentIndexHtml(params: AugmentIndexHtmlOptions): Promise
       const isModuleType = moduleFiles.some(scriptPredictor);
 
       if (isNoModuleType && !isModuleType) {
-        attrs.push({ name: 'nomodule', value: null }, { name: 'defer', value: null });
+        attrs.push({ name: 'nomodule', value: null });
+        if (!script.startsWith('polyfills-nomodule-es5')) {
```
alan-agius4: Good catch!

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -231,50 +239,183 @@ export function buildWebpackBrowser(
...
               }
+            } else if (isDifferentialLoadingNeeded && !fullDifferential) {
+              const { emittedFiles = [] } = firstBuild;
+              moduleFiles = [];
+              noModuleFiles = [];
+
+              // Common options for all bundle process actions
+              const actionOptions = {
+                optimize: normalizeOptimization(options.optimization).scripts,
+                sourceMaps:
```
alan-agius4: Nit: `normalizeSourceMaps(options.sourceMap).scripts`

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -90,16 +94,28 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
         entryPoints[polyfillsChunkName] = [path.join(__dirname, '..', 'es5-polyfills.js')];
+        if (!fullDifferential && differentialLoadingNeeded) {
+          // Add zone.js legacy support to the es5 polyfills
+          // This is a noop execution-wise if zone-evergreen is not used.
+          entryPoints[polyfillsChunkName].push('zone.js/dist/zone-legacy');
+        }
         if (!buildOptions.aot) {
+          // If not performing a full differential build the JIT polyfills need to be added to ES5
+          if (!fullDifferential && differentialLoadingNeeded) {
+            entryPoints[polyfillsChunkName].push(path.join(__dirname, '..', 'jit-polyfills.js'));
```
alan-agius4: `jit-polyfills.js` and `buildOptions.polyfills` are already added to the `polyfills` chunk.

 This would causes having polyfills to be ion both `polyfills-es5,js` and `polyfills.js`

https://github.com/angular/angular-cli/blob/0ea63bf3d2dae95ff9b0d35db42694c1a0ad4f4c/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts#L122-L136

packages/angular_devkit/build_angular/src/utils/webpack-browser-config.ts
```diff
@@ -89,6 +94,8 @@ export async function generateWebpackConfig(
         esVersionInFileName: true,
```
alan-agius4: I think `es5BrowserSupport: undefined` should be included

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -231,50 +239,183 @@ export function buildWebpackBrowser(
...
+                  (typeof options.sourceMap === 'object' && options.sourceMap.scripts),
+              };
+
+              const actions: {}[] = [];
+              const seen = new Set<string>();
+              for (const file of emittedFiles) {
+                // Scripts and non-javascript files are not processed
+                if (
+                  file.extension !== '.js' ||
+                  (file.name && scriptsEntryPointName.includes(file.name))
```
alan-agius4: This will cause an issue.

Scripts need to be processed or terser options in webpack need to be modified not to emit certain es6 optimizations.

One use case is when a variable in names the same as an object literal it will be emitted as such;

```js
var foo = true;
var bar = {foo:foo};
```

```js
var foo = true;
var bar = {foo};
```

Which will break in legacy browsers, can't quite remember the setting.

tests/legacy-cli/e2e/tests/misc/support-safari-10.1.ts
```diff
@@ -26,17 +26,34 @@ export default async function () {
...
+      <script src="main-es5.js" nomodule defer></script>
+    `);
+  } else {
+    await expectFileToMatch('dist/test-project/index.html', oneLineTrim`
+      <script src="polyfills-nomodule-es5.js" nomodule></script>
+      <script src="polyfills-es5.js" nomodule defer></script>
+      <script src="polyfills-es2015.js" type="module"></script>
+      <script src="styles-es2015.js" type="module"></script>
+      <script src="styles-es5.js" nomodule defer></script>
+      <script src="runtime-es2015.js" type="module"></script>
```
alan-agius4: Aren't polyfills dependent on runtime?

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -231,50 +239,183 @@ export function buildWebpackBrowser(
...
+
+                // Add the newly created ES5 bundles to the index as nomodule scripts
+                const newFilename = es5Polyfills
+                  ? file.file.replace('-es2015', '')
+                  : file.file.replace('es2015', 'es5');
+                noModuleFiles.push({ ...file, file: newFilename });
+              }
+
+              // Execute the bundle processing actions
+              await new Promise<void>((resolve, reject) => {
```
alan-agius4: Can we print a message that something is happening? I found it a bit confusing that after webpack emitted its stats the process continued to run without outputting any information and this was on a minimal repo. At first I though it got stuck.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -0,0 +1,179 @@
...
+    ],
+    minified: true,
+    sourceMaps: options.sourceMaps,
+  });
+
+  const newFilePath = options.filename.replace('es2015', 'es5');
+
+  // Adjust lazy loaded scripts to point to the proper variant
+  if (options.runtime) {
+    code = code.replace('"-es2015.', '   "-es5.');
```
alan-agius4: Shouldn't this be a global replace, and not replace only the first occurrence? 

Nit: remove the leading  spaces prior to `"-es5.`

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -0,0 +1,179 @@
...
+  if (map) {
+    code += `\n//# sourceMappingURL=${path.basename(newFilePath)}.map`;
+
+    fs.writeFileSync(newFilePath + '.map', map);
+  }
+
+  fs.writeFileSync(newFilePath, code);
+}
+
+async function mangleOriginal(options: ProcessBundleOptions): Promise<void> {
```
alan-agius4: Why is this an async function?

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -0,0 +1,179 @@
...
+    },
+    sourceMap: options.sourceMaps &&
+      options.map !== undefined && {
+        filename: path.basename(options.filename),
+        content: JSON.parse(options.map),
+      },
+  });
+
+  if (resultOriginal.error) {
+    return Promise.reject(resultOriginal.error);
```
alan-agius4: Shouldn't we just rethrow?

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -0,0 +1,179 @@
...
+        webkit: true,
+      },
+      sourceMap: options.sourceMaps && {
+        filename: path.basename(newFilePath),
+        content: map,
+      },
+    });
+
+    if (result.error) {
+      return Promise.reject(result.error);
```
alan-agius4: Why not `throw result.error`;

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as fs from 'fs';
+import * as path from 'path';
+import { SourceMapConsumer, SourceMapGenerator } from 'source-map';
+import { minify } from 'terser';
+
+const { transformAsync } = require('@babel/core');
```
alan-agius4: Any reason why not add `@types/babel-core`?

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -0,0 +1,179 @@
...
+    map = result.map;
+
+    // Mangle original code
+    await mangleOriginal(options);
+  } else if (map) {
+    map = JSON.stringify(map);
+  }
+
+  if (map) {
+    code += `\n//# sourceMappingURL=${path.basename(newFilePath)}.map`;
```
alan-agius4: This should not be appended if sourcemaps are hidden.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: It’s outdated unfortunately.  Appears to still be on 6.x (7.5 is latest).

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: The spaces keeps the line positions intact for source map purposes.

tests/legacy-cli/e2e/tests/misc/support-safari-10.1.ts
clydin: Webpack bundles can be in any order.  They won’t be executed until the runtime bundle is loaded and all the bundles it is dependent upon are executed.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: Tries that and it didn’t compile.  Didn’t investigate why though.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: Yes.  I’ll add it in.

packages/angular_devkit/build_angular/src/browser/index.ts
alan-agius4: The problem is that `undefined` is not assignable to 

```js
normalizeSourceMaps(sourceMap: SourceMapUnion)
```

Probably it should be changed to 
```js
export function normalizeSourceMaps(sourceMap: SourceMapUnion = false): SourceMapClass 
```

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
alan-agius4: Maybe an inline comment would help :)

package.json
clydin: Yes.  that would be a good followup to this.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: Done.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: Done.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: Done.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: Done.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: Done.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: Done.
The `ecma` option controls this.  It's now fixed to 5 for bundle differential builds.  The bundle processing will add the 6 optimizations to the ES2015 bundles.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: Done.

package.json
clydin: Actually, small change so just added it to that commit.

packages/angular_devkit/build_angular/src/utils/webpack-browser-config.ts
clydin: It is?

https://github.com/angular/angular-cli/pull/15246/files#diff-602ee11a21c4632098aa7846c328360cR98

JiaLiPassion:
@alan-agius4 , this one need to wait for https://github.com/angular/angular/pull/32016, because currently loading `zone-legacy.js` before `zone-evergreen.js` will throw error, this bug has been fixed in the above PR. thanks!

clydin:
@JiaLiPassion  All the browser tests are passing without that PR.  When is this error occurring?

JiaLiPassion:
@clydin, this error should occur when `loading zone-legacy.js` without loading `zone-evergreen.js` because `zone-legacy.js` will call `Zone.__symbol__` at the beginning, so in this PR, the `zone-legacy.js` will be loaded at the very beginning, is that correct?

clydin:
It's loaded before evergreen but they are bundled together.  Just tried it locally and there is no error.  Also added a console log statement inside the legacy register patch function and it was executed.

clydin:
Also, looking at the code, there shouldn't be an issue there.  `Zone` isn't accessed until the function is executed due to the use of ` var Zone = _global["Zone"];`
```
         (function(_global) {
          _global["__zone_symbol__legacyPatch"] = function() {
            var Zone = _global["Zone"];
            Zone.__load_patch("registerElement", function(global, Zone, api) {
              registerElementPatch(global, api);
              console.log('hello from zone legacy');
            });
            Zone.__load_patch("EventTargetLegacy", function(global, Zone, api) {
              eventTargetLegacyPatch(global, api);
              propertyDescriptorLegacyPatch(api, global);
            });
          };
        })(
          (typeof window !== "undefined" && window) ||
            (typeof self !== "undefined" && self) ||
            global
        );
```

JiaLiPassion:
@clydin, oh, got it, now the repo is still using old version of `zone.js` which doesn't have the problem, the new version of `zone.js (0.10.1)` have the error I described, sorry for the unclear information, this PR is ok with old version of `zone.js`. Thanks.

clydin:
Thanks for the info on the latest version though.

The CLI cannot use anything past 0.9.x until `@angular/core` updates its peer dependency.

Issue #6937, opened 2017-07-10
ng generate interceptor my-new-interceptor


uberspeck:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Desired functionality.
Given the new interceptor functionality it would be great to see a generator for them!

Khanji:
I vote yes!

josuevalrob:
Any solution??

Al-Sharif:
I vote yes

SummerFans:
I vote yes!!!

arturober:
Yes, please!

mgechev:
This is a perfect candidate for a third-party schematics which can be maintained by the community. Feel free to implement a generator and link to it here, with a comment. ❤️

mikelhamer:
@mgechev why has this been closed?

jaime-alex:
I vote yes.

SummerFans:
I vote yes! 👍

szabolcssimonyi:
I vote yes!

will093:
I vote yes please!

leandrofabianjr:
I vote yes!

IceTower:
I vote yes

hackdan:
I vote yes!

TomerBu:
I vote 👍 "yes, please!"

whyDontI:
+1

crutchcorn:
Please do not comment "+1" or "I vote yes". There is a react emoji that can be added to the initial issue to express such interest. Anything else only spams the thread (and anyone subscribed)

veliyat:
+1

PR #15260, opened 2019-08-06
feat(@schematics/angular): add generator for interceptor

FG-33:
It is now possible to generate interceptors using the angular-cli (see #6937):

`$ ng g interceptor(ic) test`

Result:
```
CREATE src/app/test.interceptor.spec.ts (417 bytes)
CREATE src/app/test.interceptor.ts (419 bytes)
```

Is master the correct one to merge into? The documentation says devkit:master but there's no such branch. devkit is too far behind I guess.

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -169,7 +169,7 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
   }
 
   if (options.skipTests || options.minimal) {
-    ['class', 'component', 'directive', 'guard', 'module', 'pipe', 'service'].forEach((type) => {
+    ['class', 'component', 'directive', 'guard', 'interceptor', 'module', 'pipe', 'service'].forEach((type) => {
       if (!(`@schematics/angular:${type}` in schematics)) {
         schematics[`@schematics/angular:${type}`] = {};
       }
--- a/packages/schematics/angular/collection.json
+++ b/packages/schematics/angular/collection.json
@@ -60,6 +60,11 @@
       "description": "Create a guard.",
       "schema": "./guard/schema.json"
     },
+    "interceptor": {
+      "factory": "./interceptor",
+      "description": "Create an interceptor.",
+      "schema": "./interceptor/schema.json"
+    },
     "interface": {
       "aliases": [ "i" ],
       "factory": "./interface",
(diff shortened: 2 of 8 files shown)


FG-33:
@googlebot I fixed it.

vikerman:
Thanks for the PR!

We think both the frequency of adding an interceptor and the complexity of the adding one are pretty low. So we are leaning towards not needing this schematic for now.

FG-33:
> Thanks for the PR!
> 
> We think both the frequency of adding an interceptor and the complexity of the adding one are pretty low. So we are leaning towards not needing this schematic for now.

If you say so. Thanks for the feedback though.

crutchcorn:
While in the end of any project, the maintainers are the final deciding voice and that within itself is something I respect greatly, it seems to me (from a user looking at the project) that many of the schematics are simply convinience tools. 

EG, the component generator. The component complexity seems relatively low, especially once you have an initial component to copy+paste off of. However, for an interceptor I don't often use them for various projects and then have to Google the correct usage of it (or find which of my many Angular projects has one to copy off of), so having this generator (even if the complexity is low overall) is a big convinience for me as I end up using it _enough_ to come looking for this issue every now and then 

The engagement with the linked issue (which has unfortunately fallen victim to the "+1" comment wave, but even with reacts alone) seems to show a further community interest 

Just giving my own thoughts, huge thank you to everyone on the Angular team for your persistent hard-work on the project regardless of where this issue lands

alan-agius4:
Can you also add `interceptor` in https://github.com/angular/angular-cli/blob/ace02f6c2a121a9120fd368617fe3f70992b47f6/packages/schematics/angular/application/index.ts#L172

FG-33:
@clydin I addressed all the mentioned changes. If anything's missing, let me know.

simonua:
I am glad this got merged in. Echoing @crutchcorn's appreciation for the Angular team as well as his reasoning for having a generator, we have several teams in my organization that are starting to develop with Angular 8 now and having a generator in place not only for the meat and potatoes of Angular but also these arguably more tangential use cases is helpful as I can simply mention the generator syntax in our guidelines and documentation and take more comfort in code being generated uniformly across teams. This increased conformance then cuts out nuisance PR comments that may be the result of copy/paste code propagation.

Long story short, thank you! It is appreciated.

PR #15283, opened 2019-08-09
fix(@angular-devkit/build-angular): move core-js out of dev-dependencies

Dervol03:
The polyfills in `build_angular/src/angular-cli-files/models` use `core-js` in production mode. Therefore, it needs to be in the regular dependencies, not just dev-dependencies.

This allows NPM/Yarn to install the proper `core-js` version when the main project uses a newer one.

The problem occurred when our project moved to `core-js@3.1.4`, still using Angular 7 and therefore using `@angular-devkit/build-angular@0.13.9`


clydin:

However, Angular CLI 7.x only supports core-js 2.x.  Please upgrade to 8.x if core-js 3.x is required.  Also note that core-js is no longer a project dependency for 8.0+ generated projects.

Dervol03:
@clydin 
Hi,
yes, I am aware that Angular 7.x only supports core-js 2.x.

The non-angular part of our project moved to core-js 3.1.4 which broke `build-angular`, because it does not define its core-js version in its production dependencies. Therefore, I moved the `core-js@^2.4.1` from `dev-dependencies` to `dependencies`, allowing NPM to detect that it needs to install an independent `core-js` version for `build-angular`, avoiding the need of migrating to Angular 8 right now.
I did NOT change the `core-js` version of `build-angular` on the 7.x branch, just made sure it correctly broadcasts it to NPM

clydin:
The project local version of core-js is used in 7.x.  This changed in 8.  This PR would be a breaking change as the version within  projects would no longer be used which could result in behavioral changes.

clydin:
You can use a variation of the following if core-js 3.x is required within the local project for a 7.x project:
```
{
  "compilerOptions": {
    "paths": {
      "core-js/es7/reflect": ["./node_modules/core-js/proposals/reflect-metadata"],
      "core-js/es6/*": ["./node_modules/core-js/es/*"]
    }
  }
}
```
The node modules path may need to be adjusted based on its location within the project.

Dervol03:
Okay, I'll setup the paths then. Thanks for your time and sorry for the inconvenience

clydin:
No problem at all.  And thank you for taking the time to help contribute to the project.

PR #15291, opened 2019-08-09
build: update bazel dependencies and configuration

clydin:
This leverages the configuration from FW as much as possible.
Bazel updated to 0.28.1
Rules for nodejs/typescript/karma/jasmine updated to 0.35.0

Ran into issues using the managed directories with yarn workspaces so that is disabled in the yarn install rule for now.


.bazelrc
```diff
@@ -2,11 +2,122 @@
...
+###############################
+
+# Create symlinks in the project:
+# - dist/bin for outputs
+# - dist/testlogs, dist/genfiles
+# - bazel-out
+# NB: bazel-out should be excluded from the editor configuration.
+# The checked-in /.vscode/settings.json does this for VSCode.
+# Other editors may require manual config to ignore this directory.
+# In the past, we say a problem where VSCode traversed a massive tree, opening file handles and
```
josephperrott: Should this be `we saw a problem` rather than `we say a problem`

.bazelrc
```diff
@@ -2,11 +2,122 @@
...
+# eventually a surprising failure with auto-discovery of the C++ toolchain in
+# MacOS High Sierra.
+# See https://github.com/bazelbuild/bazel/issues/4603
+build --symlink_prefix=dist/
+
+# Disable watchfs as it causes tests to be flaky on Windows
+# https://github.com/angular/angular/issues/29541
+build --nowatchfs
```
josephperrott: Perhaps we should have a platform specific set of rc files for this type of situation?  Since we would like `--watchfs` to be on for unix based systems, but need it disabled for windows.

.bazelrc
```diff
@@ -2,11 +2,122 @@
...
+# This allows us to avoid installing a second copy of node_modules
+common --experimental_allow_incremental_repository_updates
+
+# This option is changed to true in Bazel 0.27 and exposes a possible
+# regression in Bazel 0.27.0.
+# Error observed is in npm_package target `//packages/common/locales:package`:
+# ```
+# ERROR: /home/circleci/ng/packages/common/locales/BUILD.bazel:13:1: Assembling
+# npm package packages/common/locales/package failed: No usable spawn strategy found
+# for spawn with mnemonic SkylarkAction.  Your --spawn_strategyor --strategy flags
```
josephperrott: Should this be `--spawn_strategy or --strategy`?

WORKSPACE
```diff
@@ -1,57 +1,80 @@
...
 
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 http_archive(
     name = "build_bazel_rules_nodejs",
-    sha256 = "fb87ed5965cef93188af9a7287511639403f4b0da418961ce6defb9dcf658f51",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/0.27.7/rules_nodejs-0.27.7.tar.gz"],
+    # patch_args = ["-p1"],
```
josephperrott: Do we want to use the patch for angular-cli?

WORKSPACE
```diff
@@ -1,57 +1,80 @@
...
+    # Patch https://github.com/bazelbuild/rules_nodejs/pull/903
+    # patches = ["//tools:rollup_bundle_commonjs_ignoreGlobal.patch"],
+    sha256 = "6625259f9f77ef90d795d20df1d0385d9b3ce63b6619325f702b6358abb4ab33",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/0.35.0/rules_nodejs-0.35.0.tar.gz"],
 )
 
 # We use protocol buffers for the Build Event Protocol
```
josephperrott: Is this necessary still for BEP setup?  I know that we don't need to do this for `angular/angular` so I am not sure if we need to do this here.  Unless there is something additional I am not aware of.

WORKSPACE
clydin: Removed the commented out properties.  The patch is only needed to allow zone.js to be synced into g3 which doesn't apply to the CLI repo.

.bazelrc
clydin: Fixed up the grammatical issue.  Since this file is 99% a copy of FW's file now, they also exist in there as well.

WORKSPACE
clydin: The CLI has some basic BEP streaming support and there is a utility that supports debugging it which uses the `proto_library` rule that in turn requires this.

.bazelrc
clydin: If there is a way to make them be used automatically based on the platform outside of CI, that would be a good idea.

josephperrott: I don't know that there is a way to automatically choose the platform.

I wasn't thinking about how this wasn't CI specific, we can just leave it as `--nowatchfs`.  One day maybe we will be able to do a platform based configurations.

PR #15303, opened 2019-08-11
 fix(@angular-devkit/build-optimizer): don't add pure comments to tslib helpers

### LINKED ISSUE #15301: @angular-devkit/build-angular removes custom library decorators from another libraries on production build
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

I define  some decorators on one of my libraries  and use them a lot on other libraries.
When I build my application with --prod  Angular compiler removes using of them in generated code.

This problem occurs with **@angular-devkit/build-angular@0.802.0** and I have  no problem with **@angular-devkit/build-angular@0.801.3**

alan-agius4:
Closes #15301

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
+++ b/packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
@@ -5,9 +5,14 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import * as tslib from 'tslib';
 import * as ts from 'typescript';
 
 const pureFunctionComment = '@__PURE__';
+
+// We include only exports that start with '__' because tslib helpers
+// all start with a suffix of two underscores.
+const tslibHelpers = new Set<string>(Object.keys(tslib).filter(h => h.startsWith('__')));
 
 // Find all nodes from the AST in the subtree of node of SyntaxKind kind.
 export function collectDeepNodes<T extends ts.Node>(node: ts.Node, kind: ts.SyntaxKind): T[] {
@@ -40,4 +45,8 @@ export function hasPureComment(node: ts.Node): boolean {
   const leadingComment = ts.getSyntheticLeadingComments(node);
 
   return !!leadingComment && leadingComment.some(comment => comment.text === pureFunctionComment);
+}
+
+export function isHelperName(name: string): boolean {
+  return tslibHelpers.has(name);
 }
--- a/packages/angular_devkit/build_optimizer/src/transforms/prefix-functions.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/prefix-functions.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import * as ts from 'typescript';
-import { addPureComment, hasPureComment } from '../helpers/ast-utils';
+import { addPureComment, hasPureComment, isHelperName } from '../helpers/ast-utils';
 
 export function getPrefixFunctionsTransformer(): ts.TransformerFactory<ts.SourceFile> {
   return (context: ts.TransformationContext): ts.Transformer<ts.SourceFile> => {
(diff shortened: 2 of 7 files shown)


packages/angular_devkit/build_optimizer/src/transforms/prefix-functions.ts
```diff
@@ -78,9 +76,15 @@ export function findTopLevelFunctions(parentNode: ts.Node): Set<ts.Node> {
         topLevelFunctions.add(node);
       } else if (ts.isCallExpression(innerNode)) {
         let expression: ts.Expression = innerNode.expression;
+
+        if (isHelperName(expression.getText())) {
```
clydin: getText can be expensive.  Better to check for an identifier and then access it’s text property.

packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
```diff
@@ -41,3 +41,15 @@ export function hasPureComment(node: ts.Node): boolean {
 
   return !!leadingComment && leadingComment.some(comment => comment.text === pureFunctionComment);
 }
+
+export function isHelperName(name: string): boolean {
+  // TODO: there are more helpers than these, should we replace them all?
```
clydin: As a follow up, I think it would be useful to add a one time initialization that would import “tslib” and add all the exported helpers to a set and use that for checks here.  That way it wouldn’t constantly have to be synced.  Especially considering it’s new use in this PR.

alan-agius4: Thanks for the super fast review. I agree.

packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
```diff
@@ -5,9 +5,12 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+// tslint:disable-next-line: no-implicit-dependencies
+import * as tslib from 'tslib';
```
alan-agius4: Technically `tslib` should be marked as a peerDep as it is require.

Any reason why it's not at the moment?

clydin: It definitely should be now that it is used explicitly.

packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
```diff
@@ -5,9 +5,12 @@
...
 import * as ts from 'typescript';
 
 const pureFunctionComment = '@__PURE__';
+const tslibHelpers = new Set<string>(Object.keys(tslib));
```
clydin: Will this pull in anything else besides the helpers?

alan-agius4: There is nothing else defined apart from the helpers. 

if it makes you feel better, when we can do a filter `Object.keys(tslib).filter(x => x.startsWith('__')`

clydin: Good idea.  Just in case something changes in the future.

clydin: Could you add a short comment to explaining the significance of `__`?

packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
alan-agius4: are you for peer or direct dep?

packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
alan-agius4: done

packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
clydin: I think peer dep is more correct but we would need to propagate it up through `build-angular` since it is a project level dependency.

alan-agius4: Wouldn't we get something like `@angular-devkit/build-angular > @angular-devkit/build-optimizer" has incorrect peer dependency "tslib".`?

clydin: It would need to be a peer dep on `build-angular` as well.  That way `build-angular` satisfies `build-optimizer`.

But let's go with a direct dep for now. Since it won't cause peer dep warning changes which is problematic in a patch/minor release.

alan-agius4: Done with the direct dependecy

alan-agius4:
Didn't apply cleanly on patch. Patch version: https://github.com/angular/angular-cli/pull/15314

PR #15313, opened 2019-08-12
fix(@schematics/angular): default interface for guard

cexbrayat:
Currently, if the user hits `<Enter>` before selecting an interface to implement, the CLI will generate a broken guard that implements no interface. With this commit, the CLI generates a `CanActivate` guard by default.

### DIFF
--- a/packages/schematics/angular/guard/index.ts
+++ b/packages/schematics/angular/guard/index.ts
@@ -22,7 +22,7 @@ import {
 import { applyLintFix } from '../utility/lint-fix';
 import { parseName } from '../utility/parse-name';
 import { createDefaultPath } from '../utility/workspace';
-import { Schema as GuardOptions } from './schema';
+import { Implement as GuardInterface, Schema as GuardOptions } from './schema';
 
 
 export default function (options: GuardOptions): Rule {
@@ -31,19 +31,15 @@ export default function (options: GuardOptions): Rule {
       options.path = await createDefaultPath(host, options.project as string);
     }
 
-    if (options.implements === undefined) {
-      options.implements = [];
+    if (!options.implements) {
+      throw new SchematicsException('Option "implements" is required.');
     }
 
-    let implementations = '';
-    let implementationImports = '';
-    if (options.implements.length > 0) {
-      implementations = options.implements.join(', ');
-      implementationImports = `${implementations}, `;
-      // As long as we aren't in IE... ;)
-      if (options.implements.includes('CanLoad')) {
-        implementationImports = `${implementationImports}Route, UrlSegment, `;
-      }
+    const implementations = options.implements.join(', ');
+    let implementationImports = `${implementations}, `;
+    // As long as we aren't in IE... ;)
+    if (options.implements.includes(GuardInterface.CanLoad)) {
+      implementationImports = `${implementationImports}Route, UrlSegment, `;
     }
 
     const parsedPath = parseName(options.path, options.name);
--- a/packages/schematics/angular/guard/index_spec.ts
+++ b/packages/schematics/angular/guard/index_spec.ts
@@ -105,4 +105,17 @@ describe('Guard Schematic', () => {
       expect(fileString).toContain(functionName);
     });
   });
+
+  it('should use CanActivate if no implements value', async () => {
+    const options = { ...defaultOptions, implements: undefined };
+    const tree = await schematicRunner.runSchematicAsync('guard', options, appTree)
+      .toPromise();
+    const fileString = tree.readContent('/projects/bar/src/app/foo.guard.ts');
+    expect(fileString).toContain('CanActivate');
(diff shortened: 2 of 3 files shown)


packages/schematics/angular/guard/index.ts
```diff
@@ -37,13 +36,14 @@ export default function (options: GuardOptions): Rule {
...
     let implementations = '';
     let implementationImports = '';
-    if (options.implements.length > 0) {
-      implementations = options.implements.join(', ');
-      implementationImports = `${implementations}, `;
-      // As long as we aren't in IE... ;)
-      if (options.implements.includes('CanLoad')) {
-        implementationImports = `${implementationImports}Route, UrlSegment, `;
-      }
+    if (options.implements.length === 0) {
```
alan-agius4: I think the actual problem is within the schema.

I think the `implements` should be something along;
```
    "implements": {
      "type": "array",
      "uniqueItems": true,
      "enum": [
        "CanActivate",
        "CanActivateChild",
        "CanLoad"
      ],
      "default": "CanActivate",
      "x-prompt": {
        "message": "Which interfaces would you like to implement?",
        "type": "list",
        "multiselect": true,
        "items": [
          "CanActivate",
          "CanActivateChild",
          "CanLoad"
        ]
      }
    }
```

PS: the commit should probably be a `fix`

packages/schematics/angular/guard/index_spec.ts
```diff
@@ -105,4 +105,17 @@ describe('Guard Schematic', () => {
       expect(fileString).toContain(functionName);
     });
   });
+
+  it('should use CanActivate if no implements value', async () => {
+    const options = { ...defaultOptions, implements: []};
```
alan-agius4: default value should be undefined to be applied.

alan-agius4: If the user hits enter before selecting, is the value an empty array or undefined?

In that case, probable we'd need to handle empty array + undefined.

clydin:
What about adding `"minItems": 1,` to the schema for the option?  A guard without any of them doesn't really make sense.

alan-agius4:
I think the combination of all the suggested above. Will fix this schematic for both interactive and not. As at the moment when not interactive the valid implementation is too loose.

packages/schematics/angular/guard/index_spec.ts
cexbrayat: The value is an empty array, and the undefined case was already handled by assigning an empty array by default 👍

cexbrayat:
@alan-agius4 I amended the commit with your suggestions and @clydin 's 👍 That should handle the interactive and non interactive cases.

packages/schematics/angular/guard/index_spec.ts
alan-agius4: So the default empty array assignment is a problem. In general, it's preferred to use the schema defaults.

packages/schematics/angular/guard/index.ts
```diff
@@ -37,13 +36,14 @@ export default function (options: GuardOptions): Rule {
 
     let implementations = '';
```
alan-agius4: https://github.com/angular/angular-cli/blob/9b3d50804d6781f82658ac5f887259932ad29cd6/packages/schematics/angular/guard/index.ts#L33-L35

Can we removed as a default is always assigned when it's undefined.

packages/schematics/angular/guard/index_spec.ts
alan-agius4: The schema default gets applied prior of the schematic code execution. Meaning that if `option-x` has a default within the schematic it cannot be undefined.

packages/schematics/angular/guard/index.ts
alan-agius4: The logic here see

packages/schematics/angular/guard/index.ts
```diff
@@ -37,13 +36,14 @@ export default function (options: GuardOptions): Rule {
...
     let implementationImports = '';
-    if (options.implements.length > 0) {
-      implementations = options.implements.join(', ');
-      implementationImports = `${implementations}, `;
-      // As long as we aren't in IE... ;)
-      if (options.implements.includes('CanLoad')) {
-        implementationImports = `${implementationImports}Route, UrlSegment, `;
-      }
+    if (options.implements.length === 0) {
+      options.implements.push(GuardInterface.CanActivate);
```
alan-agius4: The logic here should now be redundant due to the defaults.

packages/schematics/angular/guard/index_spec.ts
alan-agius4: `implements` - cannot be an empty array, as it's invalid. it should be `undefined`. so that the defaults are applied.

PR #15320, opened 2019-08-13
Ivy workspaces

alan-agius4:
**feat(@schematics/angular): introduce Ivy libraries for development**  

Since `NGCC` is non incremental and in library projects we have the original TS sources
we don't need to build a library using the `VE` and transform it using `NGCC`. Instead we can build the library using `NGTSC` (Ivy) directly
 as this enables faster incremental compilations and a better development experience.

 Libraries now have a `production` configuration, which enabled `VE` compilations. As it is not recommended to publish NGTSC (Ivy)
 built libraries to NPM repositories, since Ivy libraries are not backwards compatible with the legacy View Engine.

**feat(@schematics/angular): remove enableIvy option** 
 With this change we remove the enableIvy option as now we only support generating Ivy application. Users who want to create a VE applications should follow the opt-out guide

### DIFF
--- a/packages/schematics/angular/library/files/tsconfig.lib.json.template
+++ b/packages/schematics/angular/library/files/tsconfig.lib.json.template
@@ -15,8 +15,6 @@
     "annotateForClosureCompiler": true,
     "skipTemplateCodegen": true,
     "strictMetadataEmit": true,
-    "fullTemplateTypeCheck": true,
-    "strictInjectionParameters": true,
     "enableResourceInlining": true
   },
   "exclude": [
--- a/packages/schematics/angular/library/index.ts
+++ b/packages/schematics/angular/library/index.ts
@@ -124,7 +124,7 @@ function addDependenciesToPackageJson() {
   };
 }
 
-function addAppToWorkspaceFile(
+function addLibToWorkspaceFile(
   options: LibraryOptions,
   projectRoot: string,
   projectName: string,
@@ -146,6 +146,11 @@ function addAppToWorkspaceFile(
           options: {
             tsConfig: `${projectRoot}/tsconfig.lib.json`,
             project: `${projectRoot}/ng-package.json`,
+          },
+          configurations: {
+            production: {
+              tsConfig: `${projectRoot}/tsconfig.lib.prod.json`,
+            },
           },
         },
         test: {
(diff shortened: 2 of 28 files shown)


alan-agius4:
When FW flips the `enableIvy` flag, I will revert the last commit. Also note that this is for Version 9.

packages/schematics/angular/library/index.ts
```diff
@@ -147,6 +147,11 @@ function addAppToWorkspaceFile(
             tsConfig: `${projectRoot}/tsconfig.lib.json`,
             project: `${projectRoot}/ng-package.json`,
           },
+          configurations: {
+            production: {
+              tsConfig: `${projectRoot}/tsconfig.lib.prod.json`,
```
filipesilva: We also need a migration for this, right?

tests/legacy-cli/e2e/setup/500-create-project.ts
```diff
@@ -18,13 +18,20 @@ export default async function() {
...
 
-    if (argv['ivy']) {
-      extraArgs.push('--enable-ivy');
-    }
-
     await ng('new', 'test-project', '--skip-install', ...extraArgs);
     await expectFileToExist(join(process.cwd(), 'test-project'));
     process.chdir('./test-project');
+
+    if (!argv['ivy']) {
```
filipesilva: At some point we should flip this flag, so that we have to opt-in VE tests instead of opt-in Ivy ones.

packages/schematics/angular/library/index.ts
alan-agius4: There is already a separate PR for that, I made it separate since it can go in now in master.
https://github.com/angular/angular-cli/pull/15272

I also, created some utils, which would be needed for another schematic

tests/legacy-cli/e2e/setup/500-create-project.ts
alan-agius4: Good idea. I will do it later today or tmr.

alan-agius4: As discussed offline, let’s tackle this in a separate PR were we enable the full test suit for Ivy and remove `—ivy` and add `—ve` so that all tests will run in ivy mode by default.

packages/schematics/angular/library/files/tsconfig.lib.json.template
```diff
@@ -15,8 +15,6 @@
     "annotateForClosureCompiler": true,
     "skipTemplateCodegen": true,
     "strictMetadataEmit": true,
```
alan-agius4: Note: the below options are not needed by ngtsc compiler, but we will leave them to make opt-out easier.
```
skipTemplateCodegen: true,
strictMetadataEmit: true,
enableResourceInlining: true,
```

PR #15321, opened 2019-08-13
fix(@ngtools/webpack): add more details to the unused files warnings

alan-agius4:


packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -637,7 +637,10 @@ export class AngularCompilerPlugin {
         continue;
       }
 
-      compilation.warnings.push(`${fileName} is part of the TypeScript compilation but it's unused.`);
+      compilation.warnings.push(
+        `${fileName} is part of the TypeScript compilation but it's unused.\n` +
+        `Only add the entrypoints to the 'files' or 'include' properties in your tsconfig.`,
```
filipesilva: Nit: `Only add the entrypoints` -> `Add only entry points`

alan-agius4: done

PR #15322, opened 2019-08-13
test: add E2E test for only ES2015 ngcc with differential loading

clydin:


clydin:
moved it under the ivy directory

PR #15324, opened 2019-08-13
fix(@schematics/angular): handle the inlineTemplate case for ng new

mgechev:
Update the template of `app.component.ts` when using `--inlineTemplate`.


packages/schematics/angular/application/other-files/app.component.ts.template
```diff
@@ -4,10 +4,11 @@ import { Component } from '@angular/core';
   selector: '<%= selector %>',<% if(inlineTemplate) { %>
   template: `
     <!--The content below is only a placeholder and can be replaced.-->
-    <div style="text-align:center">
+    <div style="text-align:center" class="content">
```
alan-agius4: Where is this class being declared? Cannot seem to find it outside of the external html file.

alan-agius4: Oh, it's required for the tests to pass.

PR #15345, opened 2019-08-15
refactor(@angular-devkit/core): convert json schema registry to use async/await

clydin:


packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -126,22 +126,27 @@ export class CoreSchemaRegistry implements SchemaRegistry {
     this._ajv.addMetaSchema(require('ajv/lib/refs/json-schema-draft-06.json'));
   }
 
-  private _fetch(uri: string): Promise<JsonObject> {
+  private async _fetch(uri: string): Promise<JsonObject> {
     const maybeSchema = this._uriCache.get(uri);
 
     if (maybeSchema) {
       return Promise.resolve(maybeSchema);
```
alan-agius4, line 133:
```suggestion
      return maybeSchema;
```

packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -126,22 +126,27 @@ export class CoreSchemaRegistry implements SchemaRegistry {
...
-        ).toPromise();
+    for (const handler of this._uriHandlers) {
+      let value = await handler(uri);
+      if (value === null || value === undefined) {
+        continue;
       }
+
+      if (isObservable(value)) {
+        value = await value.toPromise();
+      }
```
alan-agius4, line 145: `let value = await handler(uri);` will cause loose of type info on `value`.

I think something like the below is clearer and retains types.

```suggestion
      let value = handler(uri);
      if (!value) {
        continue;
      }

      if (isObservable(value)) {
        value = value.toPromise();
      }

      value = await value;
```

alan-agius4 (approved):
This is great as it removes a lot of unneeded RXJS complexity.

Just 2 comments, otherwise LGTM.

PR #15349, opened 2019-08-16
Remove/cleanup additional usage of experimental workspaces

clydin:


packages/angular/cli/commands/config-impl.ts
```diff
@@ -235,9 +231,12 @@ export class ConfigCommand extends Command<ConfigCommandSchema> {
...
       }
 
-      return this.get(config._workspace, options);
+      const workspace = ((config as {}) as { _workspace: experimental.workspace.WorkspaceSchema })
+        ._workspace;
+
+      return this.get(workspace, options);
     } else {
-      return this.set(options);
+      return await this.set(options);
```
alan-agius4: Nit: redundant `await`

packages/angular/cli/commands/new-impl.ts
```diff
@@ -43,7 +43,7 @@ export class NewCommand extends SchematicCommand<NewCommandSchema> {
     });
   }
 
-  private parseCollectionName(options: any): string {
-    return options.collection || this.getDefaultSchematicCollection();
+  private async parseCollectionName(options: any): Promise<string> {
+    return options.collection || await this.getDefaultSchematicCollection();
```
alan-agius4: Nit: redundant `await`

PR #15362, opened 2019-08-17
test: add replace_resources tests without helpers

alan-agius4:


alan-agius4:
Caretaker note: patch is 8.3 branch

PR #15364, opened 2019-08-17
refactor(@angular-devkit/build-angular): cache downlevel bundles

clydin:


packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -149,13 +156,21 @@ async function processWorker(options: ProcessBundleOptions): Promise<void> {
...
+    if (options.cacheKey && options.cachePath) {
+      const key = options.cacheKey + '-dl-map' + (options.vendorSourceMaps ? '-v' : '');
+      await cacache.put(options.cachePath, key, map);
+    }
+
     fs.writeFileSync(newFilePath + '.map', map);
   }
 
+  if (options.cacheKey && options.cachePath) {
+    await cacache.put(options.cachePath, options.cacheKey + '-dl', code);
```
alan-agius4: Why not provide the full `cacheKey` when invoking? Rather than just the hash to avoid building the cache key twice?

IE:

consumer
```js
const cacheKey = `${packageVersion}-${action.code.length}-${codeHash}-dl`;
processActions.push({
	...action,
	cacheKey,
	...
});
```
here:
```
await cacache.put(options.cachePath, options.cacheKey, code);
```

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -359,30 +370,117 @@ export function buildWebpackBrowser(
...
-                  if (error) {
-                    workerFarm.end(workers);
-                    reject(error);
-                  } else if (++completed === actions.length) {
-                    workerFarm.end(workers);
-                    resolve();
+
+              const processActions: typeof actions = [];
+              const cacheActions: { src: string; dest: string }[] = [];
+              for (const action of actions) {
```
alan-agius4: I am not fully grasping the need for parts of these logic statements. Maybe I am missing some use case or something important that I am not thinking of.

But given an input the output should always be the same optimization and sourcemap settings. Because the input files contents will differ based on the following build settings:
- sourceMaps: true, false, hidden or vendor
- script optimization: true or false
- optimizeOnly: true or false

Each of the above setting will create different content in the original file (.map or .js) and hence when generating the file hash it should be enough.

Hence, I would picture the logic to take into consideration both file contents which should also aid a bit readability.

Something along;
```ts
const cacheActions: { src: string; dest: string }[] = [];

for (const action of actions) {
  const codeHash = createHash('sha1')
    .update(action.code)
    .digest('hex');
  const cacheKey = `${packageVersion}-${action.code.length}-${codeHash}`;

  let mapCacheKey: undefined | string;
  if (action.map && action.sourceMaps) {
    const mapHash = createHash('sha1')
      .update(action.map)
      .digest('hex');

    mapCacheKey = `${packageVersion}-${action.map.length}-${mapHash}`;
  }

  const originalEntry = await cacache.get.info(cacheDownlevelPath, `${cacheKey}-orig`);
  const originalMapEntry = mapCacheKey && (await cacache.get.info(cacheDownlevelPath, `${mapCacheKey}-orig-map`));
  const isOriginalCached = originalEntry && (originalMapEntry || !action.sourceMaps);

  // if original file is not cached it's downleveled or the optimized bundle is not cached either
  if (isOriginalCached) {
    if (actions.optimizeOnly) {
      // bundle only needs to be optimized
      cacheActions.push({ src: originalEntry.path, dest: action.file });
      if (originalMapEntry) {
        cacheActions.push({ src: originalMapEntry.path, dest: `${action.file}.map` });
      }

      continue;
    } else {
      // Bundle needs to be downleveled
      const downlevelEntry = await cacache.get.info(cacheDownlevelPath, `${cacheKey}-dl`);
      const downlevelMapEntry = mapCacheKey && (await cacache.get.info(cacheDownlevelPath, `${mapCacheKey}-dl-map`));
      const isDownLeveledCached = downlevelEntry && (downlevelMapEntry || !action.sourceMaps);

      if (isDownLeveledCached) {
        const newFileName = action.filename.replace('-es2015.', '-es5.');
        cacheActions.push({ src: downlevelEntry.path, dest: newFileName });
        if (downlevelMapEntry) {
          cacheActions.push({ src: downlevelMapEntry.path, dest: `${newFileName}.map` });
        }

        continue;
      }
    }
  }

  processActions.push({
    ...action,
    cacheKey,
    mapCacheKey,
    cachePath: cacheDownlevelPath || undefined,
  });
}

for (const action of cacheActions) {
  fs.copyFileSync(action.src, action.dest, fs.constants.COPYFILE_FICLONE);
}
```

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -149,13 +156,21 @@ async function processWorker(options: ProcessBundleOptions): Promise<void> {
       code += `\n//# sourceMappingURL=${path.basename(newFilePath)}.map`;
     }
 
+    if (options.cacheKey && options.cachePath) {
+      const key = options.cacheKey + '-dl-map' + (options.vendorSourceMaps ? '-v' : '');
```
alan-agius4: We shouldn't be needed to generate different key prefixes for vendor sourcemaps. Check above.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: Completely reworked the logic to greatly simplify it as well as minimizing the number of cache access attempts to just the required entries.
I was avoiding hashing the source map files as they tend to be quite large.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: Centralized cache key calculation to the parent.

packages/angular_devkit/build_angular/src/browser/index.ts
alan-agius4: Much better :)

Issue #13656, opened 2019-02-12
Expose webpack-dev-server allowedHosts option


diminutivesloop:
# 🚀 Feature request

### Command (mark with an `x`)

```
- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
Due to constraints of our development environment my team usually accesses their local instance of our Angular application via their Windows machine name rather than an ip address or localhost. Since machine names are unique we can't set the `host` option in `angular.json` and as a result individual developers have to pass their machine name on the command line when running `ng serve`.

### Describe the solution you'd like
Since our machine names have a common parent domain it would simplify our development setup if the `allowedHosts` option of webpack-dev-server was exposed as an option for `ng serve`. This would enable us to configure a host wildcard that includes all of our machine names without additional user-local configuration.

### Describe alternatives you've considered
One alternative solution would be to enable the `disableHostCheck` option but this leaves our machines vulnerable to a dns rebinding attack.

PR #15366, opened 2019-08-18
feat(@angular-devkit/build-angular): expose webpack-dev-server's allowedHosts option

mrxz:
This PR adds `allowedHosts` as option and passes it along to the dev server. Issue #13656 describes a use-case for this. My use-case is slightly different, as I'm working on a multi-tenant application that derives the tenant from the host name. Having access to `allowedHosts` makes it possible to define multiple hosts pointing to 127.0.0.1 and using them to access the dev server (during development).

A test was added, but I'm not entirely happy with the end result. The `beforeEach` and `afterEach` are duplicate with `public-host_spec_large.ts` (although more spec files have the exact same code). It might also be better to combine it with the public-host spec file, as some test cases in there (no options specified, or with `disableHostCheck`) also apply to the `allowedHosts` option.

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -1119,6 +1119,14 @@
               "type": "string",
               "description": "The URL that the browser client (or live-reload client, if enabled) should use to connect to the development server. Use for a complex dev server setup, such as one with reverse proxies."
             },
+            "allowedHosts": {
+              "type": "array",
+              "description": "Whitelist of hosts that are allowed to access the dev server.",
+              "default": [],
+              "items": {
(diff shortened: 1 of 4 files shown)


mrxz:
@googlebot I signed it!

clydin:

The test is fine for now.  They could indeed use a refactor to combine similar items but that is outside the scope for this.

Can you rebase? Some changes to the CI checks were recently done.

Could you also add a `Closes #13656` to the message?

mrxz:
I have rebased and amended the commit message. Several tests in test-large fail, but I get the exact same results when running the test-large set on master, so it seems unrelated to the change.

PR #15371, opened 2019-08-19
test: use Ivy by default, opt-in to VE

filipesilva:


tests/angular_devkit/build_angular/hello-world-app/src/tsconfig.server.json
```diff
@@ -2,6 +2,7 @@
   "extends": "./tsconfig.app.json",
   "compilerOptions": {
     "outDir": "../dist-server",
+    "baseUrl": "./",
```
alan-agius4: This doesn’t seem right, the badeUrl should only be defined in the workspace config, otherwise paths for local workspace libraries won’t work.

filipesilva: I didn't really change that. Those diffs are a bit messed up but what I really did was `hello-world-app` -> `hello-world-app-ve` and then `hello-world-app-ivy` -> `hello-world-app`. So that was there before. It might be wrong still ofc.

filipesilva: I didn't really change that. Those diffs are a bit messed up but what I really did was `hello-world-app` -> `hello-world-app-ve` and then `hello-world-app-ivy` -> `hello-world-app`. So that was there before. It might be wrong still ofc.

filipesilva:
The saucelabs tests are failing on Internet Explorer 11 and 11. I wonder if it's related to https://github.com/angular/angular/issues/28449.

filipesilva:
Note: while debugging https://github.com/angular/angular/issues/32192, I also saw https://github.com/angular/angular/issues/30569 happening on dev builds.

packages/angular_devkit/build_angular/test/browser/rebuild_spec_large.ts
```diff
@@ -267,7 +267,7 @@ describe('Browser Builder rebuilds', () => {
 
   it('rebuilds after errors in AOT', async () => {
     // DISABLED_FOR_IVY - These should pass but are currently not supported
-    if (ivyEnabled) {
+    if (!veEnabled) {
```
alan-agius4: General note: in the near future we should investigate why this test is still broken.

alan-agius4:
@filipesilva I am not seeing any changes to `scripts/test.ts`

filipesilva:
Note to reviewers: you can ignore changes in `tests/devkit/build_angular`. I moved `hello-world-app` -> `hello-world-app-ve` and then `hello-world-app-ivy` -> `hello-world-app`. Git doesn't seem to deal very well with that in a single commit even with `git mv`.

filipesilva:
caretaker: someone with admin access will need to change these two required checks:
- `ci/circleci: e2e-cli-ivy` to `ci/circleci: e2e-cli-ve` 
- `ci/circleci: test-large-ivy` to `ci/circleci: test-large-ve`

tests/angular_devkit/build_angular/hello-world-app-ve/src/test.ts
```diff
@@ -11,13 +11,16 @@ import 'zone.js/dist/zone-testing';
 import { getTestBed } from '@angular/core/testing';
 import {
   BrowserDynamicTestingModule,
-  platformBrowserDynamicTesting,
+  platformBrowserDynamicTesting
```
SchnWalter: A TSLint rule would be useful here, in order to stick to stick to a partern and avoid removing and adding the trailing comma between MRs.

alan-agius4: These are tests assets which are not linted, since they may differ from the actual code patterns of the repo.

Issue #8046, opened 2017-10-13
Cli build should support loading modules with *.tsx extension.


majo44:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report
- [ ] feature request
```

### Versions.
```
Angular CLI: 1.5.0-rc.0
Node: 6.11.0
OS: win32 x64
Angular: 5.0.0-rc.2
... animations, common, compiler, compiler-cli, core, forms
... http, platform-browser, platform-browser-dynamic
... platform-server, router

@angular/cli: 1.5.0-rc.0
@angular-devkit/build-optimizer: 0.0.27
@angular-devkit/core: 0.0.20
@angular-devkit/schematics: 0.0.33
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 1.8.0-rc.0
@schematics/angular: 0.0.45
typescript: 2.5.3
webpack: 3.7.1
```

### Repro steps.
1. call `ng new testproject`
2. cd testproject
3. rename src\app\app.component.ts to src\app\app.component.tsx
4. add to tsconfig.json compilerOptions
    "jsx": "react",
    "jsxFactory": "h"
5. call `ng build`

### The log given by the failure.
```
ERROR in ./src/app/app.module.ts
Module not found: Error: Can't resolve './app.component' in 'c:\temp\testproject\src\app'
resolve './app.component' in 'c:\temp\testproject\src\app'
  using description file: c:\temp\testproject\package.json (relative path: ./src/app)
    Field 'browser' doesn't contain a valid alias configuration
  after using description file: c:\temp\testproject\package.json (relative path: ./src/app)
    using description file: c:\temp\testproject\package.json (relative path: ./src/app/app.component)
      no extension
        Field 'browser' doesn't contain a valid alias configuration
        c:\temp\testproject\src\app\app.component doesn't exist
      .ts
        Field 'browser' doesn't contain a valid alias configuration
        c:\temp\testproject\src\app\app.component.ts doesn't exist
      .js
        Field 'browser' doesn't contain a valid alias configuration
        c:\temp\testproject\src\app\app.component.js doesn't exist
      as directory
        c:\temp\testproject\src\app\app.component doesn't exist
[c:\temp\testproject\src\app\app.component]
[c:\temp\testproject\src\app\app.component.ts]
[c:\temp\testproject\src\app\app.component.js]
[c:\temp\testproject\src\app\app.component]
 @ ./src/app/app.module.ts 9:0-47
[...]

PR #15381, opened 2019-08-20
feat(@angular-devkit/build-angular): collect test coverage from JSX/TSX files

AdamVig:
When testing React with Karma in a monorepo managed with [Nx](https://nx.dev/) (built on top of the Angular CLI), I
found that TSX files were not showing up in test coverage reports. This change seems to fix that issue.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/test.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/test.ts
@@ -47,7 +47,7 @@ export function getTestConfig(
     }
 
     extraRules.push({
-      test: /\.(js|ts)$/,
+      test: /\.(jsx?|tsx?)$/,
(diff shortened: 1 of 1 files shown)


AdamVig:
@mgechev If you have the time, would you mind explaining what about this pull request needs to be discussed?

I am under the impression that TSX is supported by the project based on [this comment by filipesilva](https://github.com/angular/angular-cli/issues/8046#issuecomment-336949265) (from 2017, but I could not find anything more recent).

I need this change for a work project, so I am trying to determine if I should wait for it to get merged or if I should find a workaround.

mgechev:
Hey @AdamVig, since Angular does not use JSX supporting it extends our scope of work. During our triage meeting this/next Thursday we'll discuss how to proceed.

mgechev:
Since we already have partial tsx support we can merge this fix and discuss separately whether we want to keep maintaining the feature.

Thanks for the contribution!

PR #15400, opened 2019-08-21
fix(@angular-devkit/build-optimizer): replace multiple tslib helpers

alan-agius4:
Inlined tslib helpers can be suffixed with `$` and a number when having multiple helpers in the same file.

With this change we will replace all tslib inline helpers to imports from `tslib`

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/transforms/import-tslib.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/import-tslib.ts
@@ -12,7 +12,7 @@ import { isHelperName } from '../helpers/ast-utils';
  * @deprecated From 0.9.0
  */
 export function testImportTslib(content: string) {
-  const regex = /var (__extends|__decorate|__metadata|__param) = \(.*\r?\n\s+(.*\r?\n)*\s*\};/;
+  const regex = /var (__extends|__decorate|__metadata|__param)(\$\d+)? = \(.*\r?\n\s+(.*\r?\n)*\s*\};/;
 
   return regex.test(content);
 }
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_optimizer/src/helpers/ast-utils.ts
```diff
@@ -48,5 +48,5 @@ export function hasPureComment(node: ts.Node): boolean {
 }
 
 export function isHelperName(name: string): boolean {
-  return tslibHelpers.has(name);
+  return tslibHelpers.has(name.split('$')[0]);
```
clydin: This should probably remain as is for simplicity and to avoid any odd behavior.

packages/angular_devkit/build_optimizer/src/transforms/import-tslib.ts
```diff
@@ -34,11 +34,13 @@ export function getImportTslibTransformer(): ts.TransformerFactory<ts.SourceFile
           const declarations = node.declarationList.declarations;
 
           if (declarations.length === 1 && ts.isIdentifier(declarations[0].name)) {
-            const name = (declarations[0].name as ts.Identifier).text;
+            const name = declarations[0].name.text;
+            const helperName = name.split('$')[0];
```
clydin: I think some additional checks here would be good to prevent false positives.
If the split output length is greater than 2 or the second part is not a number, it's probably something else.

alan-agius4: I changed this to a RegExp

clydin: I tend to avoid regex in performance critical areas but this transformer in general is in need of a performance refactor so it's probably not a problem.

PR #15410, opened 2019-08-22
feat(@angular/cli): support for dev dependencies in add command

### LINKED ISSUE #15409: [ng add]: support for dev dependencies
# 🚀 Feature request

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [X] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
For our library, we want to provide schematics for the `ng add` command. However, this automatically installs our library as `dependency`. Shouldn't there be a way to install it to the `devDependencies`?

We do add our library to the `devDependencies` our selves (using schematics), but that results in the following warning:

```
Installed packages for tooling via npm.
    ✅️ Added "foo" into devDependencies
    🔍 Installing packages...
UPDATE package.json (980 bytes)
npm WARN The package foo is included as both a dev and production dependency.
```

* Support for an additional `dev` option: `ng add mylib --dev`.
* Update the `npmInstall()` task with an additional `dev: boolean` argument.

This could be done in a backwards compatible way.

### Alternative solution

* Let `ng add` perform an `npm install --no-save`
* Change the `package.json` ourselves

This might be breaking.

dirkluijk:
Closes #15409.

### DIFF
--- a/packages/angular/cli/commands/add-impl.ts
+++ b/packages/angular/cli/commands/add-impl.ts
@@ -135,7 +135,7 @@ export class AddCommand extends SchematicCommand<AddCommandSchema> {
       }
     }
 
-    await npmInstall(packageIdentifier.raw, this.logger, packageManager, this.workspace.root);
+    await npmInstall(packageIdentifier.raw, this.logger, packageManager, this.workspace.root, true, options.dev);
 
     return this.executeSchematic(collectionName, options['--']);
   }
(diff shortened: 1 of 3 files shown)


vikerman:
We agree we need a way to `ng add` package as a dev dependency - but not through a flag. This requires a proper infrastructure support in the CLI to do this properly.

PR #15412, opened 2019-08-22
fix(@schematics/angular): implements items type for guard schematic

cexbrayat:
Fixes the type of the items in the `schema.json` for `implements`, as `enum` is not picked up by the CLI.
Also migrates the existing e2e test to `async/await` and adds another one with `implements` to avoid regressions.

My previous PR #15313 fixed an issue in the guard schematic but introduced a regression (sorry!) as:

```json
"items": {
  "enum": [
    "CanActivate",
    "CanActivateChild",
    "CanLoad"
  ]
},
```
is not picked up by the CLI, so `--implements` ends up not being recognized when running:

    ng g guard foo --implements=CanActivate

The regression was released in CLI v8.3.0 stable (not in the next or rc releases).

If the issue is encountered, it can be easily worked around by using `--defaults` or manually picking the interface wanted in the prompt.

### DIFF
--- a/packages/schematics/angular/guard/schema.json
+++ b/packages/schematics/angular/guard/schema.json
@@ -59,7 +59,8 @@
           "CanActivate",
           "CanActivateChild",
           "CanLoad"
-        ]
+        ],
+        "type": "string"
       },
       "default": [
         "CanActivate"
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/guard/schema.json
```diff
@@ -55,11 +55,7 @@
       "description": "Specifies which interfaces to implement.",
       "uniqueItems": true,
       "items": {
-        "enum": [
-          "CanActivate",
-          "CanActivateChild",
-          "CanLoad"
-        ]
+        "type": "string"
```
alan-agius4: Adding this and leaving the enum should be enough. Can you please check?

PR #15414, opened 2019-08-22
 feat(@angular/cli): update with migrate only creates commit per migration

clydin:


packages/angular/cli/commands/update.json
```diff
@@ -62,6 +62,12 @@
           "description": "Display additional details about internal operations during execution.",
           "type": "boolean",
           "default": false
+        },
+        "skipCommits": {
+          "description": "Do not create source control commits for updates and migrations.",
+          "type": "boolean",
+          "default": false,
+          "aliases": ["C"]
```
alan-agius4: Nit: won't it be better to use `sc` as an aliases? So that it is not confused with `-c` `collection`

packages/angular/cli/commands/update-impl.ts
```diff
@@ -366,9 +440,46 @@ export class UpdateCommand extends SchematicCommand<UpdateCommandSchema> {
...
+
+      return result.trim();
+    } catch {
+      return null;
+    }
+  }
+}
+
+function coerceVersionNumber(version: string): string | null {
+  if (!version.match(/^\d{1,30}\.\d{1,30}\.\d{1,30}/)) {
```
alan-agius4: Why not use `semver.coerce`?

packages/angular/cli/commands/update-impl.ts
```diff
@@ -28,11 +33,96 @@ const npa = require('npm-package-arg');
...
+
+    const reporterSubscription = this.workflow.reporter.subscribe(event => {
+      // Strip leading slash to prevent confusion.
+      const eventPath = event.path.startsWith('/') ? event.path.substr(1) : event.path;
+
+      switch (event.kind) {
+        case 'error':
+          error = true;
+          const desc = event.description == 'alreadyExist' ? 'already exists' : 'does not exist.';
+          this.logger.warn(`ERROR! ${eventPath} ${desc}.`);
```
alan-agius4: Shouldn't an error use the `.error`?

alan-agius4: It should also be nice that instead of duplicating the method a function is created to share this logic.

clydin: It probably should be an error.

I would like to do a larger refactor of the schematic command at some point which would pull the schematic execution elements into a standalone class that could then be leveraged as needed.

packages/angular/cli/commands/update.json
clydin: Aliases should only be a single letter.  `ng update -vC` is a valid option set for verbose and no commits.  Also, `collection` is not an option for update.

PR #15416, opened 2019-08-23
fix(@ngtools/webpack): only add ctor params to decorated classes

clydin:
This mimics the behavior of TypeScript's `emitDecoratorMetadata` option.


clydin:
Was waiting on confirmation that the referenced issue was indeed caused by this.  I'll add it now.

PR #15424, opened 2019-08-23
feat(@schematics/angular): add migration to enable AOT by default

alan-agius4:
With this change we enable the AOT option for the browser builder when an application will use Ivy as rendering engine.


alan-agius4:
Should we enable `aot` in the schema by default?

PR #15454, opened 2019-08-27
build: 1.17.3 and ensure all https in yarn lock

josephperrott:


alan-agius4:
@josephperrott It is very unlikely that this PR will apply cleanly on the patch branch, since you amended the lock file.

alan-agius4:
Umm.. `fail("Unknown Yarn version {}".format...))`

PR #15455, opened 2019-08-27
fix(@angular/cli): handle promise rejection on analytics install prompt

clydin:
A rejection in this case should be ignored to prevent potential installation failure.


packages/angular/cli/bin/postinstall/analytics-prompt.js
```diff
@@ -8,9 +8,12 @@ if ('NG_CLI_ANALYTICS' in process.env) {
 try {
```
alan-agius4: The try/catch block seems to be redundant

alan-agius4: Well sure how come the require would fail since it imports a module within the same bundle. So if it fails it means we are not shipping it.  Unless the module itself run code when being required directly.

PR #15459, opened 2019-08-27
feat(@angular/cli): add support for custom npmrc path

AdamVig:
According to [the npm config docs](https://docs.npmjs.com/misc/config#npmrc-files), a user can specify the path to their
`npmrc` file using the environment variable `NPM_CONFIG_USERCONFIG`.

I ran into this issue on my system because I have my `npmrc` file in a custom location. I noticed it when running `ng
update @angular/cli` because there were 404 errors on a few internal packages, which indicated that my `npmrc` was not
being honored.

I'm not sure if this fix is sufficient, but wanted to at least start a conversation about the issue.

### DIFF
--- a/packages/angular/cli/utilities/package-metadata.ts
+++ b/packages/angular/cli/utilities/package-metadata.ts
@@ -91,8 +91,8 @@ function readOptions(
   }
 
   const defaultConfigLocations = [
-    path.join(globalPrefix, 'etc', baseFilename),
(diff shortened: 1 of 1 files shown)


alan-agius4:
@AdamVig, can you kindly rebase?

alan-agius4:
@vikerman this can be merged as it’s has some pending checks due to the recent CI changes.

PR #15470, opened 2019-08-28
fix(@angular-devkit/build-angular): ensure generator polyfills are present for ES5

clydin:


trotyl:
> The removed polyfills are still need for `ng serve` when targeting es5 for legacy browsers or build without DL.
> 
> Ex: `downlevelIteration` does require `symbol.iterator` polyfill.

@alan-agius4 It wasn't removed, [`/es/symbol`](https://github.com/zloirock/core-js/blob/master/packages/core-js/es/symbol/index.js) contains both `/modules/symbol` and specific predefined symbols.

alan-agius4:
Thanks @trotyl

PR #15471, opened 2019-08-28
feat(@angular-devkit/build-angular): add errorOnCircularDependencies build option

briancmpbll:
Add an errorOnCircularDependencies build option that will cause the build to treat circular dependencies as errors. This flag allows CI tools to easily treat detected circular dependencies as failed builds.


vikerman:
We think adding a generic flag to convert all warnings to errors would be useful and can solve this current issue.

briancmpbll:
@vikerman Is there an effort planned to add this generic flag?

clydin:
It's been placed on the backlog and has not been assigned yet.  You are more than welcome to attempt an implementation.

Please note that you will need to sign the CLA for a contribution to be accepted.

PR #15472, opened 2019-08-28
feat(@angular/cli): migrate e2e tests to async/await

dubzzz:
Protractor recently introduced a breaking change in version 6.0.0: no more custom control flow, everyone should use async/await instead.

Source: https://github.com/angular/protractor/blob/master/CHANGELOG.md
> Control flow is removed and you should use async await to run your tests.

This behaviour is already accessible in versions before 6.0.0 under the flag SELENIUM_PROMISE_MANAGER=false.

By setting this flag to false - which is the default in 6.0.0 - I hope to reduce the migration time of new users of Angular using the CLI and using Protractor for e2e.

I personnaly did the migration 2 years ago and I believe it is preferable to start directly in async/await mode. It makes the test easier to read and futurproof.

### DIFF
--- a/packages/angular_devkit/build_angular/test/app-shell/app-shell_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/app-shell/app-shell_spec_large.ts
@@ -210,15 +210,15 @@ import { createArchitect, host, veEnabled } from '../utils';
       'e2e/app.e2e-spec.ts': `
         import { browser, by, element } from 'protractor';
 
-        it('should have ngsw in normal state', () => {
-          browser.get('/');
+        it('should have ngsw in normal state', async () => {
+          await browser.get('/');
           // Wait for service worker to load.
-          browser.sleep(2000);
-          browser.waitForAngularEnabled(false);
-          browser.get('/ngsw/state');
+          await browser.sleep(2000);
+          await browser.waitForAngularEnabled(false);
+          await browser.get('/ngsw/state');
           // Should have updated, and be in normal state.
-          expect(element(by.css('pre')).getText()).not.toContain('Last update check: never');
-          expect(element(by.css('pre')).getText()).toContain('Driver state: NORMAL');
+          expect(await element(by.css('pre')).getText()).not.toContain('Last update check: never');
+          expect(await element(by.css('pre')).getText()).toContain('Driver state: NORMAL');
         });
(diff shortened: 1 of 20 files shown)


dubzzz:
@googlebot I signed it!

vikerman:
Thanks for the PR!

We discussed this in our team - This also needs a migration script to migrate all existing Protractor tests. We decided not to do it for now because of priorities.

Will revisit this as part of our next major release.

dubzzz:
@vikerman @alan-agius4 Please let me know as soon as you'll be ready to merge such PR. For the moment it is more like a proposal for a PR. I am not sure we need to wait for a migration script to do that. In a way, it would be better for new users to start directly with the async/await mode (easier to understand, faster).

**Protractor migration ROI**

In a sense a migration tool from Protractor with custom promise manager to async/await would be a nice thing but in reality the two approaches are not fully equivalent. Protractor in async/await mode is far faster than the old mode. The async/await mode can create port exhaustion quite easily on Windows on code like:

```js
element.all(by.repeater('todo in todoList.todos'))
      .filter(todo => todo.getText().then(label => label.indexOf('#10') !== -1))
      .each(todo => todo.element(by.css('input')).click())
```

The same code with async/await runs out of ports see https://github.com/angular/protractor/issues/4706 and https://github.com/dubzzz/protractor-move-async-await

In my current company we did the change but it was more than just putting async and await where it needs, we needed to tweak some of our page objects to avoid port exhaustion (to be able to run our tests on Windows).

alan-agius4:
Closing since this no longer the recommended path since protractor version 6 is deprecated.

dubzzz:
We can maybe bump towards the latest version of protractor instead?

alan-agius4:
Version 7 is essentially the same as version 5, but includes some breaking changes.

The CLI version 10 already uses version 7.

devoto13:
Is there an issue or publicly shared vision on how team plans to move forward with Protractor? I'm afraid that we will be stuck with undebuggable control flow same as we're stuck with zone.js without support for native async/await for several years and no clear way forward.

As for a migration path for existing apps. Maybe a fixer based on [`await-promise`](https://palantir.github.io/tslint/rules/await-promise/) + some iterations testing it against Google's codebase and improving fixer and Protractor types will solve most of it?

PR #15487, opened 2019-08-30
feat(@angular-devkit/build-angular): update sass-loader to 8.0.0

alan-agius4:
Full list of changes https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#800-2019-08-29

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
@@ -123,17 +123,11 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
   }
 
   let sassImplementation: {} | undefined;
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
```diff
@@ -123,17 +123,11 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
   }
 
   let sassImplementation: {} | undefined;
-  let fiber: {} | undefined;
```
vikerman: Why were the `fiber` parts removed?

clydin: Sass-loader handles it internally now.
Ref: https://github.com/webpack-contrib/sass-loader/blob/master/CHANGELOG.md#features

PR #15516, opened 2019-09-04
Small Feature - Adding pluralize to string utils

rmjoia:
when creating our own schemas, the function to pluralize is handy, as we're scaffolding new files.

stole the code from https://github.com/ngrx/platform/blob/master/modules/data/schematics-core/utility/strings.ts


rmjoia:
I suppose this is not a good solution, as only works for regular forms.

Issue #15532, opened 2019-09-05
Begin automatically locking inactive closed issues


josephperrott:
As part of a larger effort across the Angular organization, we will begin locking issues which are closed and have not been interacted with in the last 30 days.

You can [read more about our automatic conversation locking policy](https://github.com/angular/angular/blob/67d80f/docs/GITHUB_PROCESS.md#conversation-locking).

devoto13:
Is it possible to mute the bot? It was extremely annoying sending dozens of useless notifications per day.

clydin:
Unfortunately, not that I am aware.  You can, however, setup an email filtering rule to immediately delete the emails.

The good news is the backfilling is almost complete for this repository.

PR #15533, opened 2019-09-05
ci: set up angular-automatic-lock-bot for the repo

josephperrott:
Begins process for #15532
Completes, for cli, https://github.com/angular/dev-infra/issues/16


josephperrott:
Merged per @vikerman instruction offline

PR #15545, opened 2019-09-06
ci: update saucelabs protractor config to run e2e tests in series rather than parallel

josephperrott:
Our combined CI usage is often hitting the max number of concurrent sessions on saucelabs.  Currentl the saucelabs job takes 12 sessions at once because the tests run in parallel, instead we should run these in series to only use 1 session.  

It should not have a large impact on overall CI run time as each of the tests take ~10 seconds so we are moving from the Saucelabs steps taking ~1:45 to ~3:00 each.  Based on previous CI runs it looks like this change will still cause the saucelabs test runs to complete ~15 minutes before other jobs in the sequence.

### DIFF
--- a/tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
+++ b/tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
@@ -101,6 +101,9 @@ exports.config = {
     },
   ],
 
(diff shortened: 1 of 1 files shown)


alan-agius4:
@josephperrott any reason why we should backport this to patch branch?

josephperrott:
@alan-agius4 no, I suppose that we should be able to run it on any branch.  Will update label

PR #15570, opened 2019-09-10
docs: expand doc for new -route flag in generate module

jbogarthyde:
Add more detail to the API doc for `ng g m --route xx --module xx`

### DIFF
--- a/packages/schematics/angular/module/schema.json
+++ b/packages/schematics/angular/module/schema.json
@@ -41,7 +41,7 @@
     },
     "route": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/module/schema.json
```diff
@@ -41,7 +41,7 @@
     },
     "route": {
       "type": "string",
-      "description": "The route path for the lazy loaded module. Requires --module option."
+      "description": "The route path for a lazy-loaded module. When supplied, creates a component in the new module using the route name, and adds the route to the `routes` array in the declaring module's routing module. Requires `--module` option."
```
alan-agius4: Route name -> route path

Routes don’t have names

alan-agius4: Ohh I see what you mean now. I think you meant module name

PR #15597, opened 2019-09-13
ci: update schedule for locking inactive issues

### LINKED ISSUE #15532: Begin automatically locking inactive closed issues

josephperrott:
Closes #15532 

Now that we have worked through the backlog of inactive closed issues. We will move to automatically locking issues once per day.

### DIFF
--- a/.github/workflows/lock-closed.yml
+++ b/.github/workflows/lock-closed.yml
@@ -1,9 +1,9 @@
-name: Lock issues that are closed and inactive
+name: Lock Inactive Issues
 
 on:
   schedule:
-    # Run at the 40th minute of every hour
-    - cron: '40 * * * *'
+    # Run at 08:00 every day
+    - cron: '0 8 * * *'
 
 jobs:
   lock_closed:


.github/workflows/lock-closed.yml
```diff
@@ -1,9 +1,9 @@
-name: Lock issues that are closed and inactive
+name: Lock Inactive Issues
 
 on:
   schedule:
-    # Run at the 40th minute of every hour
-    - cron: '40 * * * *'
+    # Run at 00:00 every day
+    - cron: '0 0 * * *'
```
clydin: Should we keep them staggered just in case or should it not matter now?

josephperrott: I think it should be okay since we are talking about only a few issues a day now.  But we can stagger them to be safe.

PR #15603, opened 2019-09-15
feat(schematics/angular): remove dependency on tsickle

alan-agius4:
With this change we remove the requirement to add tsickle as a dependency when having a workspace library.

Since the CTOR downlevel transformer which was previously provided via tsickle  is now in ng-packagr version 5.5.1+ We migrate existing libraries to remove the need for tsickle.

//FYI @alexeagle


alexeagle:
cool 🌞

evmar:
Nice!

PR #15611, opened 2019-09-17
feat(@angular/cli): create commits per migration during update

clydin:


packages/angular/cli/commands/update-impl.ts
```diff
@@ -136,6 +136,7 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
       const description = schematic.description as typeof schematic.description & {
         version?: string;
       };
+      description.version = coerceVersionNumber(description.version) || undefined;
```
alan-agius4: Not: redundant `|| undefined`

PR #15631, opened 2019-09-19
Enable more strict options

xiaoxiangmoe:


vikerman:
We decided to add strict flags incrementally till we get everybody on board to the current set and chose the ones as the first few that we had tested out inside Google.

PR #15685, opened 2019-09-26
ci: update angular robot configuration to match current CI tasks

alan-agius4:


alan-agius4:
@clydin should https://github.com/angular/angular-cli/blob/4e1a7b537745f22f08b04024186fd16a95c017cc/.github/angular-robot.yml#L106 still be disabled?

clydin:
Yes. keep it disabled.  It caused comment spam unfortunately.

PR #15690, opened 2019-09-26
feat(@angular-devkit/build-angular): add experimentalRollupPass

filipesilva:
In applications that make heavy use of lazy routes and ES2015 libraries, this option can improve bundle sizes. It might also break your bundles in ways we don't understand fully, so please test and report any problems you find.

NOTE: the following are known problems with experimentalRollupPass
- vendorChunk, commonChunk, namedChunks: these won't work, because by the time webpack sees the chunks, the context of where they came from is lost.
- webWorkerTsConfig: workers must be imported via a root relative path (e.g.`app/search/search.worker`) instead of a relative path (`/search.worker`) because of the same reason as above.
- loadChildren string syntax: doesn't work because rollup cannot follow the imports.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/webpack-rollup-loader.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/webpack-rollup-loader.ts
@@ -1,0 +1,148 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * @license
+ * @author Erik Desjardins
+ * Forked as of SHA 10fb020f997a146725963b202d79290c8798a7a0 from https://github.com/erikdesjardins/webpack-rollup-loader.
+ * Licensed under a MIT license.
+ * See https://github.com/erikdesjardins/webpack-rollup-loader/blob/10fb020f997a146725963b202d79290c8798a7a0/LICENSE for full license.
+ */
+
+import { VirtualFileSystemDecorator } from '@ngtools/webpack/src/virtual_file_system_decorator';
+import { dirname, join } from 'path';
+import { OutputAsset, OutputChunk, rollup } from 'rollup';
+import { RawSourceMap } from 'source-map';
+import webpack = require('webpack');
+
+function splitRequest(request: string) {
+  const inx = request.lastIndexOf('!');
+  if (inx === -1) {
+    return {
+      loaders: '',
+      resource: request,
+    };
+  } else {
+    return {
+      loaders: request.slice(0, inx + 1),
+      resource: request.slice(inx + 1),
+    };
+  }
+}
+
+// Load resolve paths using Webpack.
+function webpackResolutionPlugin(
+  loaderContext: webpack.loader.LoaderContext,
+  entryId: string,
+  entryIdCodeAndMap: { code: string, map: RawSourceMap },
+) {
+  return {
+    name: 'webpack-resolution-plugin',
+    resolveId: (id: string, importerId: string) => {
+      if (id === entryId) {
+        return entryId;
+      } else {
+        return new Promise((resolve, reject) => {
+          // split apart resource paths because Webpack's this.resolve() can't handle `loader!`
+          // prefixes
+          const parts = splitRequest(id);
+          const importerParts = splitRequest(importerId);
+
+          // resolve the full path of the imported file with Webpack's module loader
+          // this will figure out node_modules imports, Webpack aliases, etc.
+          loaderContext.resolve(
+            dirname(importerParts.resource),
+            parts.resource,
+            (err, fullPath) => err ? reject(err) : resolve(parts.loaders + fullPath),
+          );
+        });
+      }
+    },
+    load: (id: string) => {
+      if (id === entryId) {
+        return entryIdCodeAndMap;
+      }
+
+      return new Promise((resolve, reject) => {
+        // load the module with Webpack
+        // this will apply all relevant loaders, etc.
+        loaderContext.loadModule(
+          id,
+          (err, source, map) => err ? reject(err) : resolve({ code: source, map: map }),
+        );
+      });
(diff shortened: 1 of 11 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/plugins/webpack-rollup-loader.ts
```diff
@@ -0,0 +1,142 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// Adapted from https://github.com/erikdesjardins/webpack-rollup-loader/blob/master/index.js
```
filipesilva: The module resolution logic and base plugin architecture were taken from https://github.com/erikdesjardins/webpack-rollup-loader.

We'll need to either include the original MIT license, or contribute a more general purpose setup back to the original project and use that instead of this modification. I've contacted the author in https://github.com/erikdesjardins/webpack-rollup-loader/issues/25.

filipesilva:
In a angular.io I saw a 164 kb total size reduction, of which 160 kb were in the main bundle. The app still seemed to work afterwards.

- without `experimentalRollupPass `
```
kamik@RED-X1C6 MINGW64 /d/work/angular/aio (master)
$ du -ach --max-depth=1 dist-current/*-es2015*.js
8.0K    dist-current/10-es2015.bb774fbfdbb48a910b56.js
8.0K    dist-current/11-es2015.ec8e5e9d098af69706e6.js
4.0K    dist-current/12-es2015.ff447765d4dcc3cdbf3d.js
8.0K    dist-current/13-es2015.dc65283eabf796cbf8ec.js
16K     dist-current/14-es2015.db8ad428e08bac925985.js
92K     dist-current/5-es2015.c78d9a47687395e877ff.js
56K     dist-current/6-es2015.6bba842a651b1f9027de.js
4.0K    dist-current/7-es2015.7fce2a8bfaf0d140d20c.js
12K     dist-current/8-es2015.3387f5b73e6e053ee2ca.js
8.0K    dist-current/9-es2015.c2dc15fbac12f79fc42b.js
572K    dist-current/main-es2015.f9bd029fbcd2b80424ca.js
52K     dist-current/polyfills-es2015.a8a8b190a3d22b96a76b.js
4.0K    dist-current/runtime-es2015.b843df94cdf1ca019c37.js
844K    total
```
- with `experimentalRollupPass `
```
kamik@RED-X1C6 MINGW64 /d/work/angular/aio (master)
$ du -ach --max-depth=1 dist-rollup/*-es2015*.js
8.0K    dist-rollup/10-es2015.3cc985d7c3319f055730.js
4.0K    dist-rollup/11-es2015.debfc94f984fa6a0fb50.js
8.0K    dist-rollup/12-es2015.9360d138e386d4702d90.js
8.0K    dist-rollup/13-es2015.525c075aff5aa96c53dd.js
8.0K    dist-rollup/14-es2015.36751c7863947ec28913.js
56K     dist-rollup/5-es2015.91b7fdaa30b328609bbe.js
88K     dist-rollup/6-es2015.014f97b73e13561afaf1.js
4.0K    dist-rollup/7-es2015.3ebcfad0d49d1f2c0a1e.js
12K     dist-rollup/8-es2015.19f1ea00491c6b553e39.js
16K     dist-rollup/9-es2015.0d1220debe4e53f393f4.js
412K    dist-rollup/main-es2015.b748da54ea794ed1d0a9.js
52K     dist-rollup/polyfills-es2015.a8a8b190a3d22b96a76b.js
4.0K    dist-rollup/runtime-es2015.5f093ee29eae729ea1b3.js
680K    total
```

filipesilva:
Blocking on reaching a resolution to the licensing issue in https://github.com/angular/angular-cli/pull/15690#discussion_r328841555.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/webpack-rollup-loader.ts
filipesilva: Resolved by adding an additional license header crediting the original author and license as of used SHA:
```
/**
 * @license
 * @author Erik Desjardins
 * Forked as of SHA 10fb020f997a146725963b202d79290c8798a7a0 from https://github.com/erikdesjardins/webpack-rollup-loader.
 * Licensed under a MIT license.
 * See https://github.com/erikdesjardins/webpack-rollup-loader/blob/10fb020f997a146725963b202d79290c8798a7a0/LICENSE for full license.
 */
```

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/webpack-rollup-loader.ts
```diff
@@ -0,0 +1,148 @@
...
+  // TODO: we should be able to do this in a more elegant way by again decorating webpacks
+  // input file system inside a custom WebpackRollupPlugin, modelled after AngularCompilerPlugin.
+  const vfs = this._compiler.inputFileSystem as VirtualFileSystemDecorator;
+  const virtualWrite = (path: string, data: string) =>
+    vfs.getWebpackCompilerHost().writeFile(path, data, false);
+
+  // Bundle with Rollup
+  const rollupOptions = {
+    ...options,
+    input: entryId,
```
alan-agius4: Question: should we disable treeshaking, such as pure functional calls removals if scripts optimisation is false?

filipesilva: I think we don't even want to use this if script optimization if false. But I haven't gotten around to further matching options into the rollup pass. I think it makes more sense to do that later after we start getting feedback from its shortcomings. We might have to add a lot of config then.

filipesilva: I think we don't even want to use this if script optimization if false. But I haven't gotten around to further matching options into the rollup pass. I think it makes more sense to do that later after we start getting feedback from its shortcomings. We might have to add a lot of config then.

PR #15693, opened 2019-09-27
 Improve performance of parallel bundle processing

clydin:


vikerman:
@clydin - Doesn't apply cleanly to patch branch. Can you create a separate PR for that?

PR #15717, opened 2019-09-30
feat(@schematics/angular): update browser output path when adding universal

alan-agius4:
Currently, in the CLI universal schematic we are setting the server output path to `dist/project-server` and not amending the build outputPath

Ex:
```
dist/project
dist/project-server
```

However, the above paths are being update when adding `nguniversal` to the below:

```
dist/project/browser
dist/project/server
```

With this change it is proposed to move that logic to upstream.

Related PR to clean up nguniversal schematics https://github.com/angular/universal/pull/1265


alan-agius4:
//cc @CaerusKaru

PR #15732, opened 2019-10-01
build: use import instead of require for fast-json-stable-stringify

kyliau:
`require` statements make g3 imports difficult because the dependency is not visible to the Bazel system. Moreover proper typings make the code more robust.


clydin:
Something happened to the lock file that doesn't appear correct.

kyliau:
> Something happened to the lock file that doesn't appear correct.

fixed! thanks for the heads up

Issue #15703, opened 2019-09-27
Build Optimizer  - Cannot read property 'kind' of undefined (isAngularDecoratorMetadataExpression)


alan-agius4:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?
Sort of, due to removal of tsickle & emit decorator metadata from libs

### Description
Build optimizer fails with Cannot read property 'kind' of undefined

## 🔬 Minimal Reproduction

Paste the below in one of the tests in **_scrub-file_spec.ts_**
```js
    import { __decorate } from 'tslib';
    import { Component, ElementRef, ContentChild, Injectable, NgModule } from '@angular/core';
    import { CommonModule } from '@angular/common';

      var FooBarComponent = /** @class */ (function () {
          function FooBarComponent(elementRef) {
              this.elementRef = elementRef;
              this.inlineButtons = [];
              this.menuButtons = [];
          }
          FooBarComponent.ctorParameters = function () { return [
              { type: ElementRef }
          ]; };
          __decorate([
              ContentChild('heading', { read: ElementRef, static: true })
          ], FooBarComponent.prototype, "buttons", void 0);
          FooBarComponent = __decorate([
              Component({
                  selector: 'custom-foo-bar',
                  template: "<h1 #heading>Foo-Bar!</h1>\n",
                  styles: ["h1{color:red;background-color:green;background-image:url(../styles/assets/test.png)}"]
              })
          ], FooBarComponent);
          return FooBarComponent;
      }());
```

## 🔥 Exception or Error
<pre><code>
TypeError: Cannot read property 'kind' of undefined
  Stack:
        at <Jasmine>
        at isAngularDecoratorMetadataExpression (/Users/alan/git/angular-cli/packages/angular_devkit/build_optimizer/src/transforms/packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts:267:33)
        at checkNodeForDecorators (/Users/alan/git/angular-cli/packages/angular_devkit/build_optimizer/src/transforms/packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts:59:13)
[...]

PR #15737, opened 2019-10-02
fix(@angular-devkit/build-optimizer): fix error when `__decorate` has…

alan-agius4:
… no `__metadata`

When a __decorator expression has no __metadata call, example:

```
__decorate([
    ContentChild('heading', { read: ElementRef, static: true })
], FooBarComponent.prototype, "buttons", void 0);
```

A Cannot read property 'kind' of undefined error will be thrown.

Closes: #15703


alan-agius4:
Patch version of #15705

PR #15749, opened 2019-10-03
feat(@angular-devkit/build-optimizer): don't use getImportTslibTransformer

filipesilva:
This transformer can cause size regressions when it introduces `tslib` imports across independent chunks.

It should be deprecated because tslib adoption has become more ubiquitous.

Should also speed up Build Optimizer processing time because there's one less thing to do.

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/transforms/import-tslib.ts
+++ b/packages/angular_devkit/build_optimizer/src/transforms/import-tslib.ts
@@ -9,15 +9,6 @@ import * as ts from 'typescript';
 import { getCleanHelperName } from '../helpers/ast-utils';
 
 /**
- * @deprecated From 0.9.0
- */
-export function testImportTslib(content: string) {
-  const regex = /var (__extends|__decorate|__metadata|__param)(\$\d+)? = \(.*\r?\n\s+(.*\r?\n)*\s*\};/;
-
-  return regex.test(content);
-}
-
-/**
  * @deprecated From 0.900.0
  */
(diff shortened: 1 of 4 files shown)


filipesilva:
I saw the mentioned size regression in https://github.com/angular/angular/pull/32957, debugged it and found the following:

We have the importTsLib transformer in BO that converts inlined ts declarations into tslib imports. We have a regex test that's meant to bypass the transformer to speed things up. When you don't have sourcemaps turned on, we don't use the regex test. `zone.js` inlines the `__assign` tslib function, that is not caught by the regex test.

When you make a build with sourcemaps, the regex fails to find the zone.js inlined tslib, and does nothing. When you make a build without sourcemaps, the regex isn't used, and the more sophisticated AST based test is used instead, which converts the zone.js tslib function into an import.

Adding an import to zone.js introduces a cross-chunk dependency between polyfills and main, because both use tslib. This prevents tslib from being concatenated with the rest of polyfills and of main, which in turn prevents further optimizations.

packages/angular_devkit/build_optimizer/src/transforms/import-tslib.ts
```diff
@@ -17,6 +17,9 @@ export function testImportTslib(content: string) {
   return regex.test(content);
 }
 
+/**
+ * @deprecated From 0.9.0
```
clydin:
```suggestion
 * @deprecated From 0.900.0
```

filipesilva: Done

PR #15760, opened 2019-10-04
New i18n schema

alan-agius4:
See individual commits.

### DIFF
--- a/etc/api/angular_devkit/core/src/_golden-api.d.ts
+++ b/etc/api/angular_devkit/core/src/_golden-api.d.ts
@@ -1251,12 +1251,18 @@ export declare class WorkspaceNotYetLoadedException extends BaseException {
 export interface WorkspaceProject {
     architect?: WorkspaceTool;
     cli?: WorkspaceTool;
+    i18n?: WorkspaceProjectI18n;
     prefix: string;
     projectType: "application" | "library";
     root: string;
     schematics?: WorkspaceTool;
     sourceRoot?: string;
     targets?: WorkspaceTool;
+}
+
+export interface WorkspaceProjectI18n {
+    locales: Record<string, string>;
+    sourceLocale?: string;
 }
 
 export interface WorkspaceSchema {
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -781,8 +781,8 @@
               "description": "Log progress to the console while building.",
               "default": true
             },
-            "translateLocales": {
-              "$ref": "#/definitions/buildersOptions/translateLocales"
+            "localize": {
+              "$ref": "#/definitions/buildersOptions/localize"
             },
             "i18nMissingTranslation": {
               "$ref": "#/definitions/buildersOptions/missingTranslation"
@@ -1806,8 +1806,8 @@
               "description": "Log progress to the console while building.",
               "default": true
             },
-            "translateLocales": {
-              "$ref": "#/definitions/buildersOptions/translateLocales"
+            "localize": {
+              "$ref": "#/definitions/buildersOptions/localize"
             },
             "i18nFile": {
               "type": "string",
@@ -2023,7 +2023,7 @@
         "enum": ["warning", "error", "ignore"],
         "default": "warning"
       },
-      "translateLocales": {
+      "localize": {
         "oneOf": [
           {
             "type": "boolean",
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -342,6 +342,9 @@
           "type": "string",
           "description": "Root of the project files."
         },
+        "i18n": {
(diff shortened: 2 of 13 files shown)


packages/angular/cli/lib/config/schema.json
```diff
@@ -398,6 +401,33 @@
...
+              "type": "string",
+              "description": "The locale id used in the source code.",
+              "default": "en-US"
+            },
+            "locales": {
+              "type": "object",
+              "additionalProperties": false,
+              "patternProperties": {
+                "^[a-z]{2}(-[a-zA-Z]{2,})?$": {
+                  "type": "string",
```
alan-agius4: Following a chat with @petebacondarwin, this might become an array to support multiple input files for the same locale

packages/angular/cli/lib/config/schema.json
```diff
@@ -342,6 +342,9 @@
           "type": "string",
           "description": "Root of the project files."
         },
+        "i18n": {
```
alan-agius4: Note: this option should only be application to "application" projects, though didn't manage to do a schema validation for this yet.

packages/angular/cli/lib/config/schema.json
```diff
@@ -398,6 +401,33 @@
...
+              "type": "string",
+              "description": "How to handle missing translations for i18n.",
+              "enum": ["warning", "error"]
+            },
+            "sourceLocale": {
+              "type": "string",
+              "description": "The locale id used in the source code.",
+              "default": "en-US"
+            },
+            "locales": {
```
alan-agius4: @clydin should we also add an `inlineLocales` option at this level?

packages/angular/cli/lib/config/schema.json
```diff
@@ -398,6 +401,33 @@
         "^[a-z]{1,3}-.*": {}
       },
       "definitions": {
+        "i18n": {
+          "description": "Project i18n options",
+          "type": "object",
+          "properties": {
+            "missingTranslation": {
```
alan-agius4: @clydin, should `missingTranslation` be at this level, builder level or both?

etc/api/angular_devkit/core/src/_golden-api.d.ts
```diff
@@ -1259,6 +1260,12 @@ export interface WorkspaceProject {
     targets?: WorkspaceTool;
 }
 
+export interface WorkspaceProjectI18N {
+    locales: Record<string, string>;
+    missingTranslation?: 'warning' | 'error' | 'ignore';
+    sourceLocale?: string;
```
petebacondarwin: What is the purpose of this property?

alan-agius4: //cc @clydin

alan-agius4: I am guessing that it is used to generate the output path for the default bundle.

clydin: It will have three uses:
* value for the output path, as mentioned
* value for lang html attribute in the index
* signifies $localize content should be used verbatim when inlining

A future fourth use will be automatic registration of application locale data.

packages/angular/cli/lib/config/schema.json
clydin: That should be an option in the build itself.  The goal here is to define the locales.  How they are used (inlined, processed/copied to the output for runtime i18n, etc.) is builder behavior.

packages/angular/cli/lib/config/schema.json
clydin: We can expand this as needed.  The string form will generally be the most common.  I'd also like to examine automatic discovery of library translations moving forward as well.

packages/angular/cli/lib/config/schema.json
```diff
@@ -1965,6 +2010,17 @@
           "additionalProperties": false
         }
       }
+    },
+    "buildersOptions": {
+      "inlineLocales": {
```
clydin: This can also be a boolean. `true` meaning all locales.

packages/angular/cli/lib/config/schema.json
clydin: This seems more of a build time behavior than a description of the translations for the project.

This option actually brings up a bigger issue in that the CLI currently does not have a mechanism to control warnings.  A mechanism to enable/disable warnings and treat warnings as errors would be useful in general.  This would then be one of those warnings.

alan-agius4: Do you prefer a new option for `missingTranslation ` or use the existing `i18nMissingTranslation`?

alan-agius4:
@clydin PTAL, Once confirmed I will rollout the schema changes to all the other builders.

packages/angular/cli/lib/config/schema.json
clydin: might as well use the existing since it simplifies any deprecation logic and migrations.

Issue #15779, opened 2019-10-08
Developer documentation for @angular-devkit/schematics is misleading and incomplete


djencks:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```
schematics
### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....
unknown

### Description

 A clear and concise description of the problem...
The angular-devkit CONTRIBUTING.md file, explaining how to build and locally install angular-devkit, points to the angular-cli README.md, https://github.com/angular/angular-cli/blob/master/packages/angular/cli/README.md#development-hints-for-working-on-angular-cli  
However, these instructions only include npm linking instructions for angular-cli, not any of angular-devkit.  These instructions do not result in a locally installed schematics command, for instance.

## 🔬 Minimal Reproduction

Following the instructions at https://github.com/angular/angular-cli/blob/master/packages/angular/cli/README.md#development-hints-for-working-on-angular-cli

git clone https://github.com/angular/angular-cli.git
yarn
npm run build
cd dist/@angular/cli
npm link

does not result in the schematics command calling the locally built angular-devkit/schematics-cli or use of the locally built angular-devkit/schematics

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍 Your Environment
<pre><code>

</code></pre>

**Anything else relevant?**

djencks:
The modified instructions in https://github.com/angular/angular-cli/pull/15781 work for me to get a local schematics change to work, but I hope there is a simpler way.

djencks:
I'd prefer that this issue, complete lack of usable instructions for how to build and test the schematics engine/framework, not be conflated with the idea of improving documentation of how to use the existing schematics engine.  For instance, most or all of the examples of schematics I found by googling use and explain chain, the solution for passing a Tree & Rules between schematics #15739, whereas I couldn't find any information at all about how to develop e.g. bug fixes.

alan-agius4:
Thanks for reporting this issue. This issue is now obsolete due to changes in the recent changes.

PR #15781, opened 2019-10-08
docs add npm link information for @angular-devkit/schematics and @ang…

djencks:
…ular-devkit/schematics-cli

The @angular-cli/README.md appears to be the build/link documentation
also for @angular-devkit/schematics and
@angular-devkit/schematics-cli, but following the instructions does
not result in usable local schematics. Add sufficient commands so
these projects can be used as local packages and the schematics
command works.

Fixes https://github.com/angular/angular-cli/issues/15779

### DIFF
--- a/packages/angular/cli/README.md
+++ b/packages/angular/cli/README.md
@@ -164,8 +164,15 @@ You can find more details about changes between versions in [the Releases tab on
 git clone https://github.com/angular/angular-cli.git
 yarn
 npm run build
-cd dist/@angular/cli
-npm link
+npm link dist/@angular/cli
+# if you wish to work with angular-devkit schematics, you also need
+npm link dist/@angular-devkit/schematics
+npm link dist/@angular-devkit/schematics-cli
+# these need to be run after each npm run build
+# in addition, npm link of schematics-cli installs the published copy of @angular-devkit/schematics into node_modules
+# To induce the schematics command to use the locally compiled version of @angular-devkit/schematics:
+(cd dist/@angular-devkit/schematics-cli; npm link @angular-devkit/schematics)
+# Similar commands may be needed to work with other angular-devkit packages.
 ```
 
 `npm link` is very similar to `npm install -g` except that instead of downloading the package


packages/angular/cli/README.md
```diff
@@ -164,8 +164,16 @@ You can find more details about changes between versions in [the Releases tab on
 git clone https://github.com/angular/angular-cli.git
 yarn
 npm run build
-cd dist/@angular/cli
-npm link
+(cd dist/@angular/cli;npm link)
```
alan-agius4: Thinking about this a bit more the npm links should be something like
```
npm link dist/@angular/cli
npm link dist/@angular-devkit/schematics
```

Also, if you are using `@angular/cli` you don't need to use `@angular-devkit/schematics-cli`

djencks:
I updated the npm link commands, and think I found a way to use npm link to get schematics-cli to use the local schematics build.  I'm not developing schematics for angular but using schematics to assemble scaffolding for another project, and per the existing documentation I was using schematics-cli during development.  As my local work has moved beyond this stage I don't any longer have a setup to actually test that these commands work but the results of running them look reasonable.

djencks:
I can't tell from the GitHub interface if I'm expected to do anything else at this point.

packages/angular/cli/README.md
```diff
@@ -164,8 +164,15 @@ You can find more details about changes between versions in [the Releases tab on
...
-cd dist/@angular/cli
-npm link
+npm link dist/@angular/cli
+# if you wish to work with angular-devkit schematics, you also need
+npm link dist/@angular-devkit/schematics
+npm link dist/@angular-devkit/schematics-cli
+# these need to be run after each npm run build
+# in addition, npm link of schematics-cli installs the published copy of @angular-devkit/schematics into node_modules
+# To induce the schematics command to use the locally compiled version of @angular-devkit/schematics:
+(cd dist/@angular-devkit/schematics-cli; npm link @angular-devkit/schematics)
```
alan-agius4: use `&&` instead of `;` as the later will not work on Windows.

alan-agius4:
Closing as this PR is stale. Feel free to open a new PR. Thanks.

Issue #15782, opened 2019-10-08
Extend the schematics node-package task to allow installing individual packages globally.


djencks:
# 🚀 Feature request

### Command (mark with an `x`)

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```
schematics

### Description
 A clear and concise description of the problem or missing capability...
While the original purpose of schematics is to provide scaffolding for node.js projects, they can also be used to set up scaffolding for non-javascript projects that happen to use node based tools.  In this case, installing packages locally doesn't make sense, whereas installing tools globally can be extremely useful.  In particular, I'm working on some schematics to set up [antora](https://gitlab.com/antora/antora) projects and installing antora globally is pretty much a necessary feature.  Rather than including a customized copy of the node-package task I think it makes sense to enhance the built-in task.

### Describe the solution you'd like
 If you have a solution in mind, please describe it.
Enhance the schematics node-package task with a global flag.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?
I can package this with my schematics, but this seems like a generally useful feature.

alan-agius4:
Installing npm packages globally poses security risks as it grants system-wide access to these packages. This broad accessibility means any project or script on your system can potentially utilize these packages, raising concerns about security vulnerabilities. 

Closing as allowing installing to install packages globally is not planned.

PR #15783, opened 2019-10-08
feat(@angular-devkit/schematics) implement global package installatio…

djencks:
…n in the schematics node-package task

Extend the schematics node-package task to allow installing individual packages globally.

This is for https://github.com/angular/angular-cli/issues/15782

I haven't found any existing tests of any kind for the node-package task, and don't know how to write one from scratch.  I've tested this change by hand in one of my projects locally with all 4 package managers.


alan-agius4:
Hi @djencks, 

Thanks for your contribution, however we'd like to avoid this as in general NPM installations of global packages should be avoided.

PR #15785, opened 2019-10-08
refactor(@schematics/angular): convert test to use async await

Splaktar:
Update to have newly generated `e2e` tests use an `async()` function and `await`s within that function.

Original discussion that lead to this PR: https://github.com/angular/material.angular.io/pull/641#discussion_r330674621

Relates to PR https://github.com/angular/angular-cli/pull/15472.


Splaktar:
It looks like this breaks e2e tests on IE11, Edge, Safari, and Firefox. It also breaks on Chrome 41 and 75.

PR #15810, opened 2019-10-11
feat(@angular-devkit/build-angular): initial support for i18n translation inlining

clydin:
Initial support for differential loading builds to enable integration testing.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/index.ts
+++ b/packages/angular_devkit/build_angular/src/browser/index.ts
@@ -11,20 +11,14 @@ import {
   WebpackLoggingCallback,
   runWebpack,
 } from '@angular-devkit/build-webpack';
-import {
-  join,
-  json,
-  logging,
-  normalize,
-  tags,
-  virtualFs,
-} from '@angular-devkit/core';
+import { join, json, logging, normalize, tags, virtualFs } from '@angular-devkit/core';
 import { NodeJsSyncHost } from '@angular-devkit/core/node';
 import * as findCacheDirectory from 'find-cache-dir';
 import * as fs from 'fs';
+import * as os from 'os';
 import * as path from 'path';
 import { Observable, from, of } from 'rxjs';
-import { catchError, concatMap, map, switchMap } from 'rxjs/operators';
+import { concatMap, map, switchMap } from 'rxjs/operators';
 import { ScriptTarget } from 'typescript';
 import * as webpack from 'webpack';
 import { NgBuildAnalyticsPlugin } from '../../plugins/webpack/analytics';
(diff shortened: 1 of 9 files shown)


packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -167,17 +162,54 @@ async function initialize(
...
+    if (usedFormats.size > 0) {
+      options.i18nFormat = [...usedFormats][0];
+    }
+  }
+
+  const originalOutputPath = options.outputPath;
+
+  // If inlining store the output in a temporary location to facilitate post-processing
+  if (i18n.shouldInline) {
+    options.outputPath = fs.mkdtempSync(path.join(fs.realpathSync(os.tmpdir()), 'angular-cli-'));
```
alan-agius4: As far as I recall, on Windows temp directories are not cleaned automatically, this means that for every build we will create a temporary directly. 

Why not use `find-cache-dir`

tests/legacy-cli/e2e/tests/i18n/ivy-localize.ts
```diff
@@ -0,0 +1,151 @@
...
+import { readNgVersion } from '../../utils/version';
+
+export default async function() {
+  if (getGlobalVariable('argv').ve) {
+    return;
+  }
+
+  let localizeVersion = '@angular/localize@' + readNgVersion();
+  if (getGlobalVariable('argv')['ng-snapshots']) {
+    localizeVersion = require('../../ng-snapshot/package.json').dependencies['@angular/localize'];
```
alan-agius4: `@angular/localize` needs to be added to `ng-snapshot/package.json` currently it is missing

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -11,20 +11,14 @@ import {
   WebpackLoggingCallback,
```
alan-agius4: From what I am seeing localization only happens if DL is enabled. But what if users don't have DL enabled?

tests/legacy-cli/e2e/tests/i18n/ivy-localize.ts
clydin: It was.  The rebase appears to have removed it.  Will fix.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: Current limitation of initial support.  Follow up PRs will expand support.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: Added code to delete the temporary directory on completion of the i18n processing.

I think a temporary directory is more appropriate since the output is transient. System temp locations are generally optimized by the operating system (automatically cleaned or potentially mounted as a memory filesystem).  This will also eventually be replaced with a build setup that will keep the files in memory throughout the pipeline and eliminate the need for this.

PR #15819, opened 2019-10-11
Support multiple configurations in commands and target strings

### LINKED ISSUE #10612: Mixing configurations

filipesilva:
It's now possible to use multiple configurations by separating them with a comma:
```
ng build --configuration=one,two,three
```

They will be applied from left to right. If `--prod` is also present, it will be considered to be the first configuration and thus able to be overriden.

You can also use it in target strings:
```
        "serve": {
          "options": {
            "browserTarget": "latest-project:build:one,two"
          },
          "configurations": {
            "production": {
              "browserTarget": "latest-project:build:production,one,two"
            }
          }
```

Fix #10612

### DIFF
--- a/packages/angular_devkit/architect/node/node-modules-architect-host.ts
+++ b/packages/angular_devkit/architect/node/node-modules-architect-host.ts
@@ -100,16 +100,26 @@ export class WorkspaceNodeModulesArchitectHost implements ArchitectHost<NodeModu
     if (targetSpec === undefined) {
       return null;
     }
-    if (
-      target.configuration
-      && !(targetSpec['configurations'] && targetSpec['configurations'][target.configuration])
-    ) {
-      throw new Error(`Configuration '${target.configuration}' is not set in the workspace.`);
+
+    let additionalOptions = {};
+
+    if (target.configuration) {
+      const configurations = target.configuration.split(',').map(c => c.trim());
+      for (const configuration of configurations) {
+        if (!(targetSpec['configurations'] && targetSpec['configurations'][configuration])) {
+          throw new Error(`Configuration '${configuration}' is not set in the workspace.`);
+        } else {
+          additionalOptions = {
+            ...additionalOptions,
+            ...targetSpec['configurations'][configuration],
+          };
+        }
+      }
     }
 
     return {
       ...targetSpec['options'],
-      ...(target.configuration ? targetSpec['configurations'][target.configuration] : 0),
+      ...additionalOptions,
     };
   }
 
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/architect/node/node-modules-architect-host.ts
```diff
@@ -100,16 +100,26 @@ export class WorkspaceNodeModulesArchitectHost implements ArchitectHost<NodeModu
...
-    if (
-      target.configuration
-      && !(targetSpec['configurations'] && targetSpec['configurations'][target.configuration])
-    ) {
-      throw new Error(`Configuration '${target.configuration}' is not set in the workspace.`);
+
+    let additionalOptions = {};
+
+    if (target.configuration) {
+      const configurations = target.configuration.split(',');
```
clydin: I think we should trim the elements here in case some does something like `-c "one, two, three"`

filipesilva: That's a good idea. Added.

ocombe:
Omg that is amazing! Thanks a lot!

robertjk:
Would be good to also update the documentation to mention that you can specify multiple configurations: https://angular.io/cli/build

filipesilva:
@robertjk PR for that up in https://github.com/angular/angular/pull/33229

PR #15826, opened 2019-10-12
test: fix i18n extraction e2e test (Make master green)

alan-agius4:
The default `source-language` is `en-US` and not `en`.


filipesilva:
Odd that this passed before...

PR #15833, opened 2019-10-14
Determine relative paths correctly

msbit:
As `git status --porcelain` always shows paths relative to the top
level, fetch the top level path in `checkCleanGit` and properly
determine whether any modified files are actually within the
Angular workspace root.

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -550,6 +550,7 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
 
   checkCleanGit() {
     try {
+      const topLevel = execSync('git rev-parse --show-toplevel', { encoding: 'utf8', stdio: 'pipe' });
       const result = execSync('git status --porcelain', { encoding: 'utf8', stdio: 'pipe' });
       if (result.trim().length === 0) {
         return true;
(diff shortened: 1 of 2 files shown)


packages/angular/cli/commands/update-impl.ts
```diff
@@ -559,7 +560,7 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
       for (const entry of result.split('\n')) {
         const relativeEntry = path.relative(
           path.resolve(this.workspace.root),
-          path.resolve(entry.slice(3).trim()),
+          path.resolve(`${topLevel.trim()}/${entry.slice(3).trim()}`),
```
clydin:
```suggestion
          path.resolve(topLevel.trim(), entry.slice(3).trim()),
```

tests/legacy-cli/e2e/tests/misc/update-git-clean-subdirectory.ts
```diff
@@ -0,0 +1,26 @@
...
+  await silentGit('init', '.');
+
+  await ng('new', 'subdirectory-test-project', '--skip-install');
+  process.chdir('./subdirectory-test-project');
+  await prepareProjectForE2e('subdirectory-test-project');
+
+  await writeFile('../added.ts', 'console.log(\'created\');\n');
+  await silentGit('add', '../added.ts');
+
+  const { message } = await expectToFail(() => ng('update', '--all'));
```
msbit: @clydin with this use of `expectToFail`, I could use some guidance.

I don't actually expect this to fail, but currently `ng update --all` errors out with warnings (`Incompatible peer dependencies found.`). What alternatives to I have to deal with both error and non-error conditions idiomatically?

msbit:
Thanks @clydin @vikerman

PR #15845, opened 2019-10-15
Update framework dependencies

alan-agius4:
See individual commits


alan-agius4:
Blocked a framework release. When released we need to trigger a `yarn.lock` update.

PR #15855, opened 2019-10-16
refactor(@angular-devkit/build-angular): use compiler parsed value of…

alan-agius4:
… enableIvy

At the moment we are relying on source content to determine if the compilation is under Ivy or VE. However, we do know what compilation we are in from the parsed tsconfig.

With this change we use the `enableIvy` to set the analytics metric

### DIFF
--- a/packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
+++ b/packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
@@ -69,7 +69,6 @@ export function countOccurrences(source: string, match: string, wordBreak = fals
  * Holder of statistics related to the build.
  */
 class AnalyticsBuildStats {
-  public isIvy = false;
   public errors: string[] = [];
   public numberOfNgOnInit = 0;
   public numberOfComponents = 0;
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
```diff
@@ -159,13 +160,7 @@ export class NgBuildAnalyticsPlugin {
...
       this._stats.numberOfComponents += countOccurrences(module._source.source(), ' Component({');
       // For Ivy we just count ɵcmp.
-      const numIvyComponents = countOccurrences(module._source.source(), 'ɵcmp', true);
-      this._stats.numberOfComponents += numIvyComponents;
-
-      // Check whether this is an Ivy app so that it can reported as part of analytics.
-      if (!this._stats.isIvy && numIvyComponents > 0) {
-        this._stats.isIvy = true;
-      }
+      this._stats.numberOfComponents += countOccurrences(module._source.source(), '.ɵcmp', true);
```
alan-agius4: Added the `.` to be more safe in matches

PR #15874, opened 2019-10-18
fix(@angular/cli): logic to determine if the installed CLI is out of …

alan-agius4:
With this change we now check if the current CLI version is the latest published version. If it is not, we install a temporary version to run the `ng update` with.


packages/angular/cli/commands/update-impl.ts
```diff
@@ -190,6 +194,21 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
 
   // tslint:disable-next-line:no-big-function
   async run(options: UpdateCommandSchema & Arguments) {
+    // Check if the current installed CLI version is older than the latest version.
+    if (await this.checkCLILatestVersion(options.verbose)) {
+      this.logger.warn(
+        'The installed Angular CLI version is older then the latest published version.\n' +
```
clydin:
```suggestion
        'The installed Angular CLI version is older than the latest published version.\n' +
```

packages/angular/cli/tasks/install-package.ts
```diff
@@ -0,0 +1,182 @@
...
+function getPackageManagerArguments(packageManager: PackageManager): PackageManagerOptions {
+  return packageManager === PackageManager.Yarn
+    ? {
+      silent: '--silent',
+      saveDev: '--dev',
+      install: 'add',
+      prefix: '--global-folder',
+    }
+    : {
+      silent: '--quite',
```
clydin:
```suggestion
      silent: '--quiet',
```

packages/angular/cli/tasks/install-package.ts
```diff
@@ -0,0 +1,182 @@
...
+    // Global installs on Unix systems go to {prefix}/lib/node_modules.
+    // Global installs on Windows go to {prefix}/node_modules (that is, no lib folder.)
+    tempNodeModules = join(tempPath, 'lib', 'node_modules');
+  } else {
+    tempNodeModules = join(tempPath, 'node_modules');
+  }
+
+  // Needed to resolve schematics from this location since we use a custom
+  // resolve strategy in '@angular/devkit-core/node'
+  process.env.NG_TEMP_MODULES_DIR = tempNodeModules;
```
clydin: The schematics resolution is the last place the custom resolver is being used.  We should move it to native methods when possible.  Can you add a note that this won't be needed when that refactor is done?

alan-agius4: will do, I actually have a following PR to remove that logic. Will try to land it in the coming days.

tests/legacy-cli/e2e/tests/i18n/ivy-localize-dl.ts
```diff
@@ -19,11 +19,12 @@ export default async function() {
...
   }
 
-  let localizeVersion = '@angular/localize@' + readNgVersion();
+  let localizeVersion = readNgVersion();
   if (getGlobalVariable('argv')['ng-snapshots']) {
     localizeVersion = require('../../ng-snapshot/package.json').dependencies['@angular/localize'];
   }
-  await npm('install', `${localizeVersion}`);
+  await npm('install', `@angular/localize@${localizeVersion}`);
+  await npm('run', 'webdriver-update');
```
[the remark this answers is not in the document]
alan-agius4: Yeah, I but without it seemed to fail a lot.

PR #15879, opened 2019-10-18
feat(@schematics/angular): update @types/node for version 9

filipesilva:
Angular will only support TS 3.6+ on version 9, and older versions of `@types/node` are incompatible with it and will cause all builds to fail.

Related to https://github.com/angular/angular/pull/33250


packages/schematics/angular/migrations/update-9/update-dependencies.ts
```diff
@@ -23,6 +23,7 @@ export function updateDependencies(): Rule {
       'ng-packagr': latestVersions.ngPackagr,
       'web-animations-js': '^2.3.2',
       'codelyzer': '^5.1.2',
+      '@types/node': '^12.11.1',
```
alan-agius4: We should change new projects to use this version there are a bunch of places still configured to use `~10.14.0`

alan-agius4:
superseded by https://github.com/angular/angular-cli/pull/15879

PR #15893, opened 2019-10-21
fix(@angular-devkit/build-angular): remove @angular/localize from application if localizing

clydin:


package.json
```diff
@@ -26,6 +26,8 @@
     "bazel:test": "bazel test //...",
     "build": "npm run admin -- build",
     "build-tsc": "tsc -p tsconfig.json",
+    "debug:test": "node --inspect-brk ./bin/devkit-admin test",
```
alan-agius4: 😍

PR #15895, opened 2019-10-21
fix(@angular/cli): logic to determine if the installed CLI is out of …

alan-agius4:
…date

With this change we now check if the current CLI version is the latest published version. If it is not, we install a temporary version to run the ng update with.

Patch version of https://github.com/angular/angular-cli/pull/15874


alan-agius4:
**Caretaker:** the required check is not available on patch branch

PR #15900, opened 2019-10-22
refactor(@angular-devkit/build-angular): allow control of cache location

### LINKED ISSUE #15865: ng build - option to specify cache directory
# 🚀 Feature request

```
- [ ] new
- [x ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
In our development environment setup, we have individual developer's node_modules link to project's node_modules which is controlled by architects who decides what packages should be used in the project.

However, when individual developers tried to build their own instance, we encounter this error:
`
[exec] Generating ES5 bundles for differential loading...

[exec] An unhandled exception occurred: EPERM: operation not permitted, lchown '....../node_modules/.cache/angular-build-dl/content-v2/sha512/d4/20'`

It appears during the build, ng will create some temporary files under node_modules/.cache and even perform chown operation on it. 

The build fails because individual developer does not have sufficient permission to operate on project's node_modules directory.

I wonder whether it's possible to have a build option that allow changing default cache directory. So we can set it to individual developer's own environment where they have sufficient permission.

clydin:
`NG_BUILD_CACHE` can specify an absolute path to be used as the cache location.  Caching can also be disabled by setting the variable to 0 or false.

Closes #15865


alan-agius4:
Closes https://github.com/angular/angular-cli/issues/15865?

PR #15907, opened 2019-10-22
Bazel packages update

alan-agius4:


gregmagolan:
@googlebot I consent

alan-agius4:
@gregmagolan thanks for the help!

PR #15913, opened 2019-10-22
test: enable existing aot i18n E2E test for ivy

clydin:
This is blocked on a FW xliff 1.2 parser fix for optional `target-language` attribute support.

### DIFF
--- a/tests/legacy-cli/e2e/tests/build/aot/aot-i18n.ts
+++ b/tests/legacy-cli/e2e/tests/build/aot/aot-i18n.ts
@@ -2,9 +2,10 @@ import * as express from 'express';
 import { resolve } from 'path';
 import { getGlobalVariable } from '../../../utils/env';
 import { appendToFile, copyFile, expectFileToExist, expectFileToMatch, replaceInFile, writeFile } from '../../../utils/fs';
-import { ng } from '../../../utils/process';
+import { ng, npm } from '../../../utils/process';
 import { updateJsonFile } from '../../../utils/project';
 import { expectToFail } from '../../../utils/utils';
+import { readNgVersion } from '../../../utils/version';
 
 export default async function () {
   const baseDir = 'dist/test-project';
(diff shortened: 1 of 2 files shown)


tests/legacy-cli/e2e/tests/build/aot/aot-i18n.ts
```diff
@@ -88,8 +99,16 @@ export default async function () {
   for (const { lang, translation } of langTranslations) {
     if (lang != 'en') {
       await copyFile('src/locale/messages.xlf', `src/locale/messages.${lang}.xlf`);
-      await replaceInFile(`src/locale/messages.${lang}.xlf`, '<source>Hello i18n!</source>',
-        `<source>Hello i18n!</source>\n<target>${translation}</target>`);
+      await replaceInFile(
+        `src/locale/messages.${lang}.xlf`,
+        'source-language="en-US"',
+        `source-language="en-US" target-language="${lang}"`,
```
filipesilva: `target-language` was not needed in VE but seems to be required on Ivy.

@clydin looked into this and it seems like the xliff 1.2 parser lists it as optional in http://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html.

But `@angular/localize` will throw if it's not there: https://github.com/angular/angular/blob/ec482dadb19eea5cd247eb7428a5d88640bd38f8/packages/localize/src/tools/src/translate/translation_files/translation_parsers/xliff1/xliff1_translation_parser.ts#L56

@petebacondarwin can you shed some light on this? Is this intended, and if so, is there documentation telling users they must do it?

filipesilva: Sounds like it's a bug, opened as https://github.com/angular/angular/issues/33323

petebacondarwin:
PR to unblock this: https://github.com/angular/angular/pull/33381

PR #15924, opened 2019-10-23
refactor(@angular-devkit/schematics): remove several deprecated APIs

alan-agius4:
refactor(@angular-devkit/schematics): remove several deprecated APIs

BREAKING CHANGE:

Removed several deprecated APIs:

- EngineHost `listSchematics` has been removed in favour of `listSchematicNames`.
- `FileSystemSink` - Use the new virtualFs.Host classes from `@angular-devkit/core`.
- `optimize` function has been remove as trees are automaticlly optimized.
- `DryRunSink` deprecated constructor that allowed to provide a directory path has been removed. Provide a `virtualFs.Host` instead.


vikerman:
Will need to check how this affects schematics already inside Google

PR #15939, opened 2019-10-25
fix(@schematics/angular): only ngcc in prepare

paullryan:
This fix is to make is so an npm/yarn install with --production mode will not fail as ngcc won't have been installed yet. This allows for workflows with docker and other production build mechanics.

### DIFF
--- a/packages/schematics/angular/workspace/files/package.json.template
+++ b/packages/schematics/angular/workspace/files/package.json.template
@@ -8,7 +8,7 @@
     "test": "ng test",
     "lint": "ng lint",
     "e2e": "ng e2e",
-    "postinstall": "ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points"
+    "prepare": "ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points"
   },
   "private": true,
   "dependencies": {


vikerman:
Hi - Can you first open an issue and explain what the use case is? We don't understand under what circumstances you are encountering this issue.

clydin:
The post-install script is no longer present so this change will not be necessary.

paullryan:
While I understand this is no longer needed, to @vikerman the use case is for a build that needs to run a --production npm install first then a full install, this allows for a docker cache to create a layer of just your production values speeding future builds. 

example
```
FROM node:10-alpine as builder
WORKDIR /app

RUN apk add --no-cache --virtual builds-deps build-base python cairo-dev freetype-dev pkgconfig git 

ARG NPM_TOKEN

COPY .npmrc yarn.lock package.json ./

COPY build ./build
COPY libs/scss ./libs/scss

RUN npm config set unsafe-perm true
RUN npm i -g @angular/cli

# first set aside prod dependencies so we can copy in to the prod image
# install all dependencies and add source code
RUN yarn install --pure-lockfile
COPY . .

RUN yarn set:version && yarn set:mdi
RUN yarn build:prod:lite

FROM nginx:alpine
WORKDIR /app

RUN apk add --no-cache sed

COPY ./server/nginx.conf /etc/nginx/conf.d/default.conf
COPY ./server/entrypoint.sh /entrypoint.sh
RUN chmod +x /entrypoint.sh

RUN rm -rf /usr/share/nginx/html/*
COPY --from=builder /app/dist/manager-client /usr/share/nginx/html

EXPOSE 80
CMD /entrypoint.sh && nginx -g 'daemon off;'
```

Just adding this in case you need to run down a similar path in the future.

PR #15951, opened 2019-10-28
fix(@angular/cli): improve statues and description logs during update 

alan-agius4:
With this change we improve the log messages of migration;

> the migration description
> the outcome of the migration
> we also remove the version of the migration which was misleading (Ex: 9.0.0-beta)

Example of new logs
image

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -163,10 +163,12 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
 
     migrations.sort((a, b) => semver.compare(a.version, b.version) || a.name.localeCompare(b.name));
 
+    this.logger.info(
+      colors.cyan(`** Executing migrations of package '${packageName}' **\n`),
+    );
+
     for (const migration of migrations) {
-      this.logger.info(
-        `** Executing migrations for version ${migration.version} of package '${packageName}' **`,
-      );
+      this.logger.info(`${colors.symbols.pointer}  ${migration.description.replace(/\. /g, '.\n   ')}`);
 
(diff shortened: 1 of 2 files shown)


alan-agius4:
Also needs https://github.com/angular/angular/pull/33440

packages/angular/cli/commands/update-impl.ts
```diff
@@ -177,6 +179,8 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
           }
         }
 
+        this.logger.error('⨉  Migration failed. See above for further details.\n');
```
clydin: You can use `colors.symbols.cross` to ensure a cross platform `x`

The same with the others.  `colors.symbols.pointer`/`colors.symbols.check`

packages/angular/cli/commands/update-impl.ts
```diff
@@ -163,10 +163,12 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
...
 
+    this.logger.info(
+      colors.cyan(`** Executing migrations of package '${packageName}' **\n`),
+    );
+
     for (const migration of migrations) {
-      this.logger.info(
-        `** Executing migrations for version ${migration.version} of package '${packageName}' **`,
-      );
+      this.logger.info(`${colors.symbols.pointer}  ${migration.description.replace(/\. /g, '.\n   ')}`);
```
clydin: As a future addition, I think it would be useful to have a `title` field in the migration definition.  So that the description doesn't need to be manipulated.

alan-agius4: The description is also manipulated so that the description itself is not printed in a single line.

vikerman:
Need FW release

PR #15956, opened 2019-10-28
fix(@schematics/update): handle peerDependenciesMeta

filipesilva:


packages/schematics/update/update/index_spec.ts
```diff
@@ -362,4 +362,34 @@ describe('@schematics/update', () => {
       }),
     ).toPromise().then(done, done.fail);
   }, 45000);
+
+  it('validates peer dependencies', done => {
+    // Add the basic migration package.
+    const content = virtualFs.fileBufferToString(host.sync.read(normalize('/package.json')));
+    const packageJson = JSON.parse(content);
+    const dependencies = packageJson['dependencies'];
+    dependencies['@angular-devkit/build-angular'] = '0.900.0-next.1';
```
filipesilva: This test will fail once `@angular-devkit/build-angular` doesn't have the same peer dependencies as today.

Most tests in this spec rely on a couple of test packages being published on npm, e.g. `@angular-devkit-tests/update-base`. If we want this test to be stable, we need to publish a new one of those to test optional peer dependencies.

packages/schematics/update/update/index.ts
```diff
@@ -210,7 +216,8 @@ function _validateUpdatePackages(
     logger.debug(`${name}...`);
 
     const peers = target.packageJson.peerDependencies || {};
```
alan-agius4: Nit can be simplified by
`const {peerDependencies = {}, peerDependenciesMeta = {}} = target.packageJson`

filipesilva: Fixed

PR #15965, opened 2019-10-29
feat(@angular-devkit/build-angular): export protractor builder

minijus:
When using custom builder for dev-server it is not possible to pass additional parameters through protractor builder to it (`ng e2e --additionalParam`) does not work.

Thus, exposing protractor builder (`executeProtractorBuilder`) would allow building custom protractor builders that internally use custom dev-server builders with additional parameters.


vikerman:
Missed 9.0 timeframe. Will revisit for 9.1.

PR #15968, opened 2019-10-29
 fix(@angular-devkit/build-angular): i18n app shell with Ivy

alan-agius4:


packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -699,8 +701,8 @@ export function buildWebpackBrowser(
           event =>
             ({
               ...event,
-              // If we use differential loading, both configs have the same outputs
-              outputPath: baseOutputPath,
+              baseOutputPath: baseOutputPath,
+              outputPaths: outputPaths || baseOutputPath,
```
clydin:
```suggestion
              outputPaths: outputPaths || [baseOutputPath],
```

clydin: I think adding back the `outputPath` option as deprecated and make it equal to the first value would be useful as well.  It would remove the need to change a large volume of tests as well as prevent a breaking change for anyone currently using the builder directly.

alan-agius4: PTAL

PR #15971, opened 2019-10-29
fix(@angular-devkit/build-angular): augment with service worker during localization

clydin:


tests/legacy-cli/e2e/tests/i18n/ivy-localize-serviceworker.ts
```diff
@@ -0,0 +1,189 @@
...
+  // Add service worker source configuration
+  const manifest = {
+    index: '/index.html',
+    assetGroups: [
+      {
+        name: 'app',
+        installMode: 'prefetch',
+        resources: {
+          files: [
+            '/favicon.ico',
```
alan-agius4: The files here don't seem right. 

They should be
```
          "/favicon.ico",
          "/index.html",
          "/manifest.webmanifest",
          "/*.css",
          "/*.js"
```

tests/legacy-cli/e2e/tests/i18n/ivy-localize-serviceworker.ts
```diff
@@ -0,0 +1,189 @@
...
+    await expectToFail(() => expectFileToMatch(`${baseDir}/${lang}/main-es5.js`, '$localize`'));
+    await expectToFail(() => expectFileToMatch(`${baseDir}/${lang}/main-es2015.js`, '$localize`'));
+    await expectFileToMatch(`${baseDir}/${lang}/main-es5.js`, lang);
+    await expectFileToMatch(`${baseDir}/${lang}/main-es2015.js`, lang);
+
+    // Expect service worker configuration to be present
+    await expectFileToExist(`${baseDir}/${lang}/ngsw.json`);
+
+    // Ivy i18n doesn't yet work with `ng serve` so we must use a separate server.
+    const app = express();
```
alan-agius4: From here downwards is redundant, since we are not testing that the service worker works.

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -672,29 +664,27 @@ export function buildWebpackBrowser(
...
+
+            if (!options.watch && options.serviceWorker) {
+              for (const outputPath of outputPaths) {
+                try {
+                  await augmentAppWithServiceWorker(
+                    host,
+                    root,
+                    normalize(projectRoot),
+                    normalize(outputPath),
+                    options.baseHref || '/',
```
alan-agius4: This is good, but I guess this won't work in a case with inlining of multiple locales and your url structure is something along these lines.

```
https://foo.com/en
https://foo.com/fr
https://foo.com/ca
```

As the contents of the SW would be absolute. Example:
```
      hashTable: {
        '/foo/favicon.ico': '84161b857f5c547e3699ddfbffc6d8d737542e01',
        '/foo/assets/folder-asset.txt': '617f202968a6a81050aa617c2e28e1dca11ce8d4',
        '/foo/index.html': 'a5359e8e1a516683b32bbb2f9e8bf402dae4738e',
      },
```

In all fairness this won't work as well, for non SW builds because of the absence of a base href in the index file.

tests/legacy-cli/e2e/tests/i18n/ivy-localize-serviceworker.ts
clydin: Copied it directly from here: https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/test/browser/service-worker_spec_large.ts#L13

tests/legacy-cli/e2e/tests/i18n/ivy-localize-serviceworker.ts
clydin: My goal there was to ensure that the service worker augmentation didn't break the application.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: Why is the base HREF set here at all actually?

clydin: I think the use of the base href here needs some more thought in general.  However, this at least allows the service worker to be created.  Worst case right now, they would then need to adjust the generated configuration files.  Not ideal but still usable.

alan-agius4: I take it back, there are absolute because we are setting them as such, with the baseHref of `/`.

PR #15973, opened 2019-10-29
fix(@angular/cli): updated CLI should not have analytics enabled

alan-agius4:
In the case users don't have analytics globally configured when the CLI will self update during ng update. It will prompt to configure this.

However, afterwards the update will fail with `Repository is not clean.  Please commit or stash any changes before updating.` as there would be uncommited local changed.

### DIFF
--- a/packages/angular/cli/tasks/install-package.ts
+++ b/packages/angular/cli/tasks/install-package.ts
@@ -155,6 +155,7 @@ export function runTempPackageBin(
     env: {
       ...process.env,
       NG_DISABLE_VERSION_CHECK: 'true',
+      NG_CLI_ANALYTICS: 'false',
     },
   });
 


packages/angular/cli/tasks/install-package.ts
```diff
@@ -155,6 +155,8 @@ export function runTempPackageBin(
     env: {
       ...process.env,
       NG_DISABLE_VERSION_CHECK: 'true',
+      NG_FORCE_TTY: 'false',
```
clydin: Thinking about it some more, we may want to add interactivity to the update process so just disabling analytics may be the better long-term option.

alan-agius4: Updated.

Issue #15961, opened 2019-10-29
[Request] Change `browserslist` file name into `.browserslistrc`


lacolaco:
# 🚀 Feature request

### Command (mark with an `x`)

```
- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
 A clear and concise description of the problem or missing capability...

Currently the CLI creates `browserslist` file by `ng new`. 
The file name `browserslist` is a kind of deprecation. Previously, the new file name `.browserslistrc` was introduced.

https://github.com/browserslist/browserslist/commit/7231ccd3a0e083871538fb6245f28afd6e7b92cd

They said `browserslist` is conflicting with the npx package name.

And in Angular.io, `.browserslistrc` is only mentioned. 

https://angular.io/guide/build#configuring-browser-compatibility

### Describe the solution you'd like
 If you have a solution in mind, please describe it.

Generate the same file as `.browserslistrc` on `ng new` or `ng generate application`.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

Browserslist still is supporing old config file name for backward-compatibility. So workaround is not needed.

PR #15977, opened 2019-10-29
feature(@schematics/angular): rename browserslist to .browserslistrc

bgotink:
Implements #15961 

I'm not sure whether this should also be renamed when using `ng update` to update to angular 9, so I didn't create a migration for it yet.


alan-agius4:
Hi, thanks for your contribution. 

Let's close this for the time being and reconsider this for version 10.

More context: https://github.com/angular/angular-cli/pull/16138#issuecomment-554014906

PR #15980, opened 2019-10-30
fix(@angular-devkit/build-angular): enable locale data transform for single localization

clydin:


packages/angular_devkit/build_angular/src/utils/i18n-options.ts
```diff
@@ -146,6 +146,12 @@ export async function configureI18nBuild<T extends BrowserBuilderSchema | Server
     if (usedFormats.size > 0) {
       buildOptions.i18nFormat = [...usedFormats][0];
     }
+
+    // If only one locale is specified set the deprecated option to enable the webpack plugin
+    // transform to register the locale directly in the output bundle.
+    if (i18n.inlineLocales.size === 1) {
+      buildOptions.i18nLocale = [...i18n.inlineLocales][0];
```
petebacondarwin: Why this complicated expression rather than `i18n.inlineLocales[0]`?

alan-agius4: Cause `inlineLocales` is a `Set` and not an `Array`

petebacondarwin: Is this better than `i18n.inlineLocales.values()[0]` then - apart from being 4 chars less?
Or is it just a matter of taste?

alan-agius4: I don't have a preference, I am fine with both.

clydin: The spread will work with any iterable object.  There's also `Array.from(i18n.inlineLocales)[0]` which works on both iterables and any array-like object (length field with numeric properties).  Performance-wise it gets quite complicated as arrays are heavily special cased within the VM.
But overall, I suppose, it's more personal preference than anything else.

PR #15985, opened 2019-10-30
fix(@angular-devkit/build-angular): adjust translation parser import paths

clydin:


petebacondarwin:
Note that the paths to these files have changed: https://github.com/angular/angular/tree/master/packages/localize/src/tools/src/translate/translation_files/translation_parsers

While you are changing this you can now include the XTB parser too: https://github.com/angular/angular/blob/master/packages/localize/src/tools/src/translate/translation_files/translation_parsers/xtb_translation_parser.ts

petebacondarwin:
@vikerman - shall I open a new issue to track the suggestion I made in this PR?

PR #15989, opened 2019-10-31
fix(@angular-devkit/build-angular): insert locale data when localizing

clydin:
Since the deprecated locale data registration was removed, existing tests that rely on locale data will fail if these changes do not properly inject the new form of the locale data.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/i18n-options.ts
+++ b/packages/angular_devkit/build_angular/src/utils/i18n-options.ts
@@ -19,7 +19,10 @@ import { createTranslationLoader } from './load-translations';
 export interface I18nOptions {
   inlineLocales: Set<string>;
   sourceLocale: string;
-  locales: Record<string, { file: string; format?: string; translation?: unknown }>;
+  locales: Record<
+    string,
+    { file: string; format?: string; translation?: unknown; dataPath?: string }
+  >;
   flatOutput?: boolean;
   readonly shouldInline: boolean;
 }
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -540,6 +541,13 @@ export async function inlineLocales(options: InlineOptions) {
       const setLocaleText = `var $localize=Object.assign(void 0===$localize?{}:$localize,{locale:"${locale}"});`;
       contentClone = content.clone();
       content.prepend(setLocaleText);
+
+      if (options.localeDataBasePath) {
+        const localeDataText = loadLocaleData(options.localeDataBasePath, locale, true);
+        if (localeDataText) {
+          content.prepend(localeDataText + ';');
+        }
```
IgorMinar: we should error if the localeData for a given locale was not found

IgorMinar: hmm. we do error above in i18n-options, so why this check?

clydin: This was mainly meant as a proof of concept to ensure that the injection didn’t break anything else unintentionally.

I’m in the process of reworking this now that the FW side has solidified.  I should have a more complete version up later today.

packages/angular_devkit/build_angular/src/utils/i18n-options.ts
```diff
@@ -127,6 +130,13 @@ export async function configureI18nBuild<T extends BrowserBuilderSchema | Server
   }
 
   if (i18n.inlineLocales.size > 0) {
+    const localeDataBasePath = findLocaleBaseDataPath();
```
petebacondarwin: Should one of `localeDataBasePath` or `findLocaleBaseDataPath` change to be consistent with each other?

clydin: yes. definitely.

packages/angular_devkit/build_angular/src/utils/i18n-options.ts
```diff
@@ -197,3 +207,18 @@ function mergeDeprecatedI18nOptions(
 
   return i18n;
 }
+
+function findLocaleDataPath(): string | null {
+  try {
+    const commonPath = path.dirname(require.resolve('@angular/common/package.json'));
```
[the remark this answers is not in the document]
clydin: @petebacondarwin Will there be a `package.json` inside the `locales/global` directory?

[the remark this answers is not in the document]
petebacondarwin: I didn't intend for there to be one since I didn't plan on these files being imported directly as modules.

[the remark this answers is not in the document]
clydin: Not an issue at all.  The current logic works in that case.

PR #15995, opened 2019-10-31
build: update @angular to version 9.0.0-next.15

alan-agius4:


packages/angular_devkit/build_angular/src/utils/load-translations.ts
```diff
@@ -27,36 +27,38 @@ export async function createTranslationLoader(): Promise<TranslationLoader> {
 
 async function importParsers() {
   try {
+    // In @angular/localize version 9.0.0-next.15 the parsers were located in the below locations.
```
alan-agius4: Change the order to go on the happy path first.

alan-agius4:
Related to the CI failure in https://github.com/angular/angular-cli/pull/15992

PR #16000, opened 2019-10-31
fix(@ngtools/webpack): skip NGCC when file system is read only

alan-agius4:
If the package.json of a module is read only we should skip calling NGCC. Under Bazel when running under sandbox the filesystem is read-only.

See also: https://github.com/angular/angular/pull/33366

//cc @alexeagle, @gkalpak & @petebacondarwin

### DIFF
--- a/packages/ngtools/webpack/src/ngcc_processor.ts
+++ b/packages/ngtools/webpack/src/ngcc_processor.ts
@@ -7,7 +7,7 @@
  */
 
 import { Logger, PathMappings, process as mainNgcc } from '@angular/compiler-cli/ngcc';
-import { existsSync } from 'fs';
+import { accessSync, constants, existsSync } from 'fs';
 import * as path from 'path';
 import * as ts from 'typescript';
 import { InputFileSystem } from 'webpack';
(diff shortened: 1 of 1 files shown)


petebacondarwin:
typo in commit : `ready` -> `read`

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -67,6 +67,17 @@ export class NgccProcessor {
       return;
     }
 
+    // If the package.json is read only we should skip calling NGCC.
+    // Under Bazel when running under sandbox the filesystem is read-only.
+    try {
```
alan-agius4: Thoughts... should we do the below for each module? Or if `node_modules` directory is read-only we don't invoke this processor from in the first place?

petebacondarwin: What about `dist` type folders?

alan-agius4: Good point!

petebacondarwin:
Yeah it doesn't seem quite right... but I think the situation was that the user had already run ngcc standalone on the node_modules folder before invoking Bazel, which will have built any Angular related packages. Because of the way the CLI-ngcc integration works, it will try to mark all (even non-Angular) packages as being processed to speed up the processing for future runs.

PR #16005, opened 2019-11-01
fix(@angular/cli) change analytics type to string

### LINKED ISSUE #15972: Incompatible types, actual string (when new lib created with analitics enabled)
Hi! My IDE highlights this line:
image

kaltepeter:
Closes #15972

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -67,7 +67,7 @@
           }
         },
         "analytics": {
-          "type": "boolean",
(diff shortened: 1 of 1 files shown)


packages/angular/cli/lib/config/schema.json
```diff
@@ -67,7 +67,7 @@
           }
         },
         "analytics": {
-          "type": "boolean",
+          "type": "string",
```
clydin:
```suggestion
          "type": ["boolean", "string"],
```
It can actually be either type.

PR #16008, opened 2019-11-01
fix(@schematics/angular): replace `'**/*.ts` file inclusion with `**/*.d.ts`

alan-agius4:
This fixes warnings such as the below;
```
WARNING in /test-update/src/environments/environment.prod.ts is part of the TypeScript compilation but it's unused.
Add only entry points to the 'files' or 'include' properties in your tsconfig.
```

When the previous generated add was in VE.


alan-agius4:
//cc @vikerman, this is the fix for the issue you experienced

alan-agius4:
Superseded by https://github.com/angular/angular-cli/pull/16050

Issue #16010, opened 2019-11-01
Global cli install keeps breaking because of Installing a temporary version to perform the update


dottodot:
Using the latest version of cli I keep getting this warning 

```
Your global Angular CLI version (8.3.17) is greater than your local
version (8.3.16). The local Angular CLI version is used.

To disable this warning use "ng config -g cli.warnings.versionMismatch false".
The installed Angular CLI version is older than the latest published version.
Installing a temporary version to perform the update.
```
and after this if I try using  `ng` I then get an error of 

```
bash: /usr/local/bin/ng: No such file or directory
```
If I then run 
```
npm i -g  @angular/cli
```
I get an error of 
```
npm ERR! Refusing to delete /usr/local/bin/ng: ../../../private/var/folders/h_/nr3rbk_917v1v7f01_tpt4_80000gn/T/.ng-temp-packages-n6S6NB/node_modules/.bin/ng symlink target is not controlled by npm /usr/local/lib/node_modules/@angular/cli
npm ERR! File exists: /usr/local/bin/ng
npm ERR! Move it away, and try again.
```
so my only way around is to remove `/usr/local/bin/ng` and then install again.

I have to do this every time I run `ng update` and it's been and issue since v8.3.13

alan-agius4:
@dottodot, can you please provide the output of `ng version`?

dottodot:
yes it's

```
Angular CLI: 8.3.17
Node: 10.17.0
OS: darwin x64
Angular: 8.2.13
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... platform-server, router, service-worker

Package                                    Version
--------------------------------------------------------------------
@angular-devkit/architect                  0.803.17
@angular-devkit/build-angular              0.803.17
@angular-devkit/build-optimizer            0.803.17
@angular-devkit/build-webpack              0.803.17
@angular-devkit/core                       8.3.17
@angular-devkit/schematics                 8.3.17
@angular/cdk                               8.2.3
@angular/cli                               8.3.17
@angular/flex-layout                       8.0.0-beta.27
@angular/material                          8.2.3
@angular/material-moment-adapter           8.2.3
@angular/pwa                               0.801.3
@ngtools/webpack                           8.3.17
@nguniversal/common                        8.1.1
@nguniversal/express-engine                8.1.1
@nguniversal/module-map-ngfactory-loader   8.1.1
@schematics/angular                        8.1.3
[...]

PR #16011, opened 2019-11-01
fix(@angular/cli): don't install using global command but rather install in a different folder

alan-agius4:
Fixes: #16010


alan-agius4:
Huge kudos to @cexbrayat that helped me verify that this fix works, as I was unable to replicate it locally.

alan-agius4:
Yeah I tried that myself

PR #16032, opened 2019-11-01
feat(@angular/cli): Support XDG Base Directory Specfication

muradm:
Explained here: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

Current change affects reading of global config from `.angular-config.json`. By default it will be created in user's home directory. If user moves it manually to `$HOME/.config/angular/.angular-config.json`, it will be subsequently read from new location.


muradm:
@googlebot I signed it!

PR #16050, opened 2019-11-03
Tsconfig migration fixes

alan-agius4:
See individual commits.

### DIFF
--- a/packages/schematics/angular/migrations/update-9/update-app-tsconfigs_spec.ts
+++ b/packages/schematics/angular/migrations/update-9/update-app-tsconfigs_spec.ts
@@ -64,7 +64,7 @@ describe('Migration to version 9', () => {
       expect(include).toEqual(['src/**/*.d.ts']);
     });
 
-    it('should not update apps tsConfig when tsconfig has include', async () => {
+    it('should update apps tsConfig when tsconfig has include', async () => {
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/update-9/update-app-tsconfigs_spec.ts
```diff
@@ -74,10 +74,41 @@ describe('Migration to version 9', () => {
 
       const tree2 = await schematicRunner.runSchematicAsync('migration-09', {}, tree.branch()).toPromise();
       const { files, include } = JSON.parse(tree2.readContent('tsconfig.app.json'));
-      expect(files).toEqual(undefined);
+      expect(files).toEqual(['src/main.ts', 'src/polyfills.ts']);
       expect(include).toEqual(['foo.ts']);
     });
 
+    it(`should update include '**/*.ts' in apps tsConfig to '**/*.d.ts'`, async () => {
```
filipesilva: This migration will cause projects that use the string format of `loadChildren` with broken compilations, because the lazy routes will no longer be part of the compilation.

That being said, we have already deprecated the string syntax and migrated these as part of the previous version 8 migration so it should be ok. If users have trouble with this we can provide instructions on how to add the lazy routes to the compilation for users that need to still use the string syntax.

Issue #16030, opened 2019-11-01
Updating to version 9 produces many commit messages with the same name


filipesilva:
This issue impacts 9.0.0-rc.0 and happened while updating https://github.com/SAP/cloud-commerce-spartacus-storefront.

The updating procedure produces (heh) several commit messages with the same name:

```
* 5b3de0f7d (HEAD -> update-9) migrate workspace for @angular/core@9.0.0-beta
* 4cb442ad1 migrate workspace for @angular/core@9.0.0-beta
* fde9f49c1 migrate workspace for @angular/core@9.0.0-beta
* 3691566e6 migrate workspace for @angular/core@9.0.0-beta
* bcde65f2f migrate workspace for @angular/cli@9.0.0-beta.0
* a7a17a1e2 Angular CLI update
```

This makes it hard to distinguish what they are doing.

Related to https://github.com/angular/angular-cli/issues/16013.

PR #16051, opened 2019-11-03
fix(@angular/cli): add migration name to commit header

alan-agius4:
Closes: #16030

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -189,7 +189,7 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
 
       // Commit migration
       if (commit) {
-        let message = `migrate workspace for ${packageName}@${migration.version}`;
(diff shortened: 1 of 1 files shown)


packages/angular/cli/commands/update-impl.ts
```diff
@@ -189,9 +189,9 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
 
       // Commit migration
       if (commit) {
-        let message = `migrate workspace for ${packageName}@${migration.version}`;
+        let message = `${packageName}@${migration.version} workspace migration - ${migration.name}.`;
```
clydin: I think we should remove the version.  They are typically set to values that are not very useful to the end user.
Maybe drop “workspace” as well for brevity.

alan-agius4: Yeah this is a perfect use case for the the `title` field.

PR #16053, opened 2019-11-04
fix(@angular-devkit/build-angular): allow localization with development server

clydin:


packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -119,14 +156,84 @@ export function serveWebpackBrowser(
...
+      // tslint:disable-next-line: no-non-null-assertion
+      const rules = webpackConfig.module!.rules;
+      const index = rules.findIndex(r => r.enforce === 'pre');
+      if (index === -1) {
+        throw new Error('Invalid internal webpack configuration');
+      }
+
+      const i18nRule: webpack.Rule = {
+        test: /\.(?:m?js|ts)$/,
+        exclude: /[\/\\](?:core-js|@babel|tslib)[\/\\]/,
```
filipesilva: Are these exclusions just a performance thing or does something break in those?

Also, I find the babel exclusion odd. I didn't expect any babel code to be bundled in the application.

packages/angular_devkit/build_angular/src/utils/i18n-inlining.ts
```diff
@@ -40,7 +40,7 @@ function emittedFilesToInlineOptions(
       es5,
       outputPath,
       missingTranslation,
-      setLocale: emittedFile.name === 'main',
+      setLocale: emittedFile.name === 'main' || emittedFile.name === 'vendor',
```
filipesilva: Is it possible for the locale to need to be set in lazy chunks?

petebacondarwin: From the point of view of Angular, we only need to set it in `main.js` so that we can pick it up in Angular services and pipes. And I guess, since it is global, that once it is set it will automatically be available to all lazy loaded chunks.

filipesilva: Then what's the situation where it needs to be set in vendor?

packages/angular_devkit/build_angular/src/dev-server/index.ts
clydin: Performance.  There is no possibility of those containing `$localize` so no point in parsing and processing each one.  The loader rule was moved over from my usage-based polyfills branch and the regex along with it.

tests/legacy-cli/e2e/tests/i18n/legacy.ts
```diff
@@ -190,42 +190,42 @@ export async function setupI18nConfig(useLocalize = true) {
...
-    } else {
-      const server = externalServer(outputPath);
-      try {
-        // Execute without a devserver.
-        await ng('e2e', `--configuration=${lang}`, '--devServerTarget=');
-      } finally {
-        server.close();
-      }
+    // Execute Application E2E tests with dev server
+    await ng('e2e', `--configuration=${lang}`, '--port=0');
```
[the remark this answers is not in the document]
clydin: Prevents odd behavior with port already in use scenarios.  This is specially relevant when running these locally.  I actually think the e2e command should be using zero as the default when used with the dev server.  It's easy for it to fail when you are actively developing the application.

packages/angular_devkit/build_angular/src/dev-server/index.ts
clydin: Removed the exclude.  Not really needed so no real reason to add to the complexity.

tests/legacy-cli/e2e/tests/i18n/legacy.ts
```diff
@@ -190,25 +190,27 @@ export async function setupI18nConfig(useLocalize = true) {
...
-    if (getGlobalVariable('argv')['ng-snapshots']) {
-      localizeVersion = require('../../ng-snapshot/package.json').dependencies['@angular/localize'];
-    }
-    await npm('install', `${localizeVersion}`);
+  // Install the localize package.
+  let localizeVersion = '@angular/localize@' + readNgVersion();
+  if (getGlobalVariable('argv')['ng-snapshots']) {
+    localizeVersion = require('../../ng-snapshot/package.json').dependencies['@angular/localize'];
   }
+  await npm('install', `${localizeVersion}`);
```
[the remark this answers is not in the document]
clydin: wrapped in an `--ve` conditional so that it is only added to ivy projects.  This test with the deprecated options is still run on ivy.

tests/legacy-cli/e2e/tests/i18n/legacy.ts
```diff
@@ -190,25 +190,27 @@ export async function setupI18nConfig(useLocalize = true) {
...
 }
 
 export default async function () {
   // Setup i18n tests and config.
   await setupI18nConfig(false);
 
+  // Legacy option usage with the en-US locale needs $localize
+  // Legacy usage did not need to process en-US and typically no i18nLocale options were present
+  await appendToFile('src/polyfills.ts', `import '@angular/localize/init';`);
```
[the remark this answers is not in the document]
clydin: wrapped in a --ve conditional so that it is only added to ivy projects. This test with the deprecated options is still run on ivy.

tests/legacy-cli/e2e/tests/i18n/legacy.ts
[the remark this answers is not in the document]
filipesilva: Yeah I've seen similar port problems in spec_large too, where a failing tests prevents others that are using the same port from working.

tests/legacy-cli/e2e/tests/i18n/legacy.ts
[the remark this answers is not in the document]
filipesilva: Ok, that makes sense to me. I hadn't thought that it still runs on the Ivy-only jobs.

PR #16083, opened 2019-11-06
fix(@angular/cli): handle null stderr during npm install

alan-agius4:


packages/angular/cli/tasks/install-package.ts
```diff
@@ -55,11 +55,11 @@ export function installPackage(
   });
 
   if (status !== 0) {
-    let errors = stderr.trim();
-    if (errors.length) {
-      errors += '\n';
+    let error = (stderr && stderr.trim()) || '';
```
clydin: If there was an error and stderr is non existent (null rather than just empty), this should probably fallback to showing stdout so that something is shown.

alan-agius4: Updated.

PR #16084, opened 2019-11-06
refactor: remove redundant `mergeWith` in `ng-new`

alan-agius4:


packages/schematics/angular/ng-new/index.ts
```diff
@@ -62,13 +59,9 @@ export default function(options: NgNewOptions): Rule {
...
-    mergeWith(
-      apply(empty(), [
-        schematic('workspace', workspaceOptions),
-        options.createApplication ? schematic('application', applicationOptions) : noop,
-        move(options.directory),
-      ]),
-    ),
+    schematic('workspace', workspaceOptions),
+    options.createApplication ? schematic('application', applicationOptions) : noop,
+    move(options.directory),
```
clydin: Could also potentially use applyToSubTree and avoid the move.

alan-agius4: Isn't be `path` required to exists to use the `applyToSubTree`?

alan-agius4:
Looks like it's needed after all 
```
ENOENT: no such file or directory, rename '/tmp/angular-cli-e2e-xa5Rjf/subdirectory/.git/HEAD' -> '/tmp/angular-cli-e2e-xa5Rjf/subdirectory/subdirectory-test-project/.git/HEAD'
```

Issue #16094, opened 2019-11-07
Add migration to change ts module to esnext


alan-agius4:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

In version 9 we are running the lazy-loading syntax migration to change the string based lazy-loading with `import()`. 

We however don't have a migration to ensure that that in the workspace tsconfig the module is set to `esnext` and this will failures at build time.

## 🔥 Exception or Error

```
ERROR in app/routing.module.ts:7:38 - error TS1323: Dynamic import is only supported when '--module' flag is 'commonjs' or 'esNext'.

7   {path: 'grid', loadChildren: () => import('./grid/grid.module').then(m => m.DocsGridModule)},
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
app/routing.module.ts:8:38 - error TS1323: Dynamic import is only supported when '--module' flag is 'commonjs' or 'esNext'.

8   {path: 'docs', loadChildren: () => import('./layout/layout.module').then(m => m.DocsLayoutModule)},
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
app/routing.module.ts:9:44 - error TS1323: Dynamic import is only supported when '--module' flag is 'commonjs' or 'esNext'.

9   {path: 'responsive', loadChildren: () => import('./responsive/responsive.module').then(m => m.DocsResponsiveModule)},
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
app/routing.module.ts:10:40 - error TS1323: Dynamic import is only supported when '--module' flag is 'commonjs' or 'esNext'.

10   {path: 'issues', loadChildren: () => import('./github-issues/github-issues.module').then(m => m.DocsGithubIssuesModule)},
                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
app/routing.module.ts:13:25 - error TS1323: Dynamic import is only supported when '--module' flag is 'commonjs' or 'esNext'.

13     loadChildren: () => import('./stack-overflow/stack-overflow.module').then(m => m.DocsStackOverflowModule)
```

[...]

PR #16095, opened 2019-11-07
docs: remove backticks from bug report in command section

alan-agius4:
With this change the author will be able to check the command in question after the issue creation.

Ex: https://github.com/angular/angular-cli/issues/16094


alan-agius4:
Caretaker: ci failures are unrelated. This is okay to merge.

PR #16105, opened 2019-11-07
fix(@angular/cli): disable update commit creation by default

### LINKED ISSUE #16013: Update to Angular v9 automatically checks in the result!
# 🐞 Bug report

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

### Description

 When running: `ng update @angular/cli @angular/core --next --allow-dirty`

It automatically checked in the result to my git repo. This is definitely NOT desired. In many use cases, this may be a "Trial" without a desire to check in. In other use cases, this may be a key dev environment that the developers would want to try out before committing.

In any case, this should *NOT* automatically check in without at least asking first.

UPDATE: In looking into this further, it does *THREE* commits during the update process.

## 🔬 Minimal Reproduction

1) Run: ng update @angular/cli @angular/core --next --allow-dirty

2) Check the Git change list, all changes are check in.

## 🔥 Exception or Error
<pre><code>

No error

</code></pre>

## 🌍 Your Environment
<pre><code>

Was 8.1.3 prior to the update.

Now: 
Angular CLI: 9.0.0-rc.0
Node: 10.16.0
OS: win32 x64
Angular: 9.0.0-rc.0
... router

@angular-devkit/architect         0.900.0-rc.0
@angular-devkit/build-angular     0.900.0-rc.0
@angular-devkit/build-optimizer   0.900.0-rc.0
@angular-devkit/build-webpack     0.900.0-rc.0
@angular-devkit/core              9.0.0-rc.0
@angular-devkit/schematics        9.0.0-rc.0
@ngtools/webpack                  9.0.0-rc.0
@schematics/angular               9.0.0-rc.0
@schematics/update                0.900.0-rc.0
rxjs                              6.5.3
typescript                        3.6.4
</code></pre>

**Anything else relevant?**

@vikram @StephenFluin

clydin:
Closes #16013

### DIFF
--- a/packages/angular/cli/commands/update.json
+++ b/packages/angular/cli/commands/update.json
@@ -63,8 +63,8 @@
           "type": "boolean",
           "default": false
         },
-        "skipCommits": {
-          "description": "Do not create source control commits for updates and migrations.",
+        "createCommits": {
+          "description": "Create source control commits for updates and migrations.",
           "type": "boolean",
           "default": false,
           "aliases": ["C"]
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -425,7 +425,7 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
         packageName,
         migrations,
         migrationRange,
-        !options.skipCommits,
+        options.createCommits,
       );
 
       return success ? 0 : 1;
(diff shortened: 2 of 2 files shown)


packages/angular/cli/commands/update.json
```diff
@@ -63,8 +63,8 @@
           "type": "boolean",
           "default": false
         },
-        "skipCommits": {
-          "description": "Do not create source control commits for updates and migrations.",
+        "createCommits": {
```
Splaktar: I would prefer `commitUpdates` or `commitChanges` as I think that is more descriptive since it includes a hint at what is in the commits.

I guess the current version does hint at the fact that it might create more than a single commit.

However, as a user, I feel like 'createCommits' is more vague and will confuse many developers.

PR #16108, opened 2019-11-07
Fix something that doesn't really make sense

OMGerCoder:
I think you should remove the 'app' because 'angular-playground app is running!' makes no sense.


OMGerCoder:
@googlebot I signed it!

PR #16110, opened 2019-11-07
fix(@angular/cli): Prints out when a commit is made in `ng update`.

dgp1130:
Any time a `git commit` is made, the CLI now prints out the hash and short message. For migrations, the message is simply the first line of the commit. For schematics, the commit message isn't all that helpful, so I used the list of packages instead.

I didn't see any good place to put tests for this code, so I didn't add any tests for this change. Let me know if there is a particular place I should include some.

Example run with the change:

```
$ ng update -C --next @angular/core --force
                  Package "@angular-devkit/build-angular" has an incompatible peer dependency to "typescript" (requires ">=3.1 < 3.6", would install "3.6.4").
    Updating package.json with dependency @angular/compiler-cli @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency @angular/compiler @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency @angular/common @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency @angular/core @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency @angular/animations @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency @angular/forms @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency @angular/language-service @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency @angular/platform-browser @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency zone.js @ "0.10.2" (was "0.9.1")...
    Updating package.json with dependency rxjs @ "6.5.3" (was "6.4.0")...
    Updating package.json with dependency @angular/platform-browser-dynamic @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency @angular/router @ "9.0.0-rc.1" (was "8.2.13")...
    Updating package.json with dependency typescript @ "3.6.4" (was "3.5.3")...
UPDATE package.json (1365 bytes)
✔ Packages installed successfully.
  Committed migration step (aa41b0513) for packages: @angular/core@next.
** Executing migrations of package '@angular/core' **

▸ Static flag migration.
  Removes the `static` flag from dynamic queries.
  As of Angular 9, the "static" flag defaults to false and is no longer required for your view and content queries.
  Read more about this here: https://v9.angular.io/guide/migration-dynamic-flag
✔ Migration succeeded.
  No changes to commit after migration.

▸ Missing @Injectable migration.
  In Angular 9, enforcement of @Injectable decorators for DI is a bit stricter.
  Read more about this here: https://v9.angular.io/guide/migration-injectable
✔ Migration succeeded.
  No changes to commit after migration.

▸ ModuleWithProviders migration.
  In Angular 9, the ModuleWithProviders type without a generic has been deprecated.
  This migration adds the generic where it is missing.
  Read more about this here: https://v9.angular.io/guide/migration-module-with-providers
✔ Migration succeeded.
  No changes to commit after migration.

▸ NGCC postinstall migration.
  Adds an ngcc invocation to npm/yarn's postinstall script.
  Read more about this here: https://v9.angular.io/guide/migration-ngcc
UPDATE package.json (1472 bytes)
✔ Packages installed successfully.
✔ Migration succeeded.
  Committed migration step (43bf7a265): @angular/core migration - migration-v9-postinstall-ngcc.

▸ Renderer to Renderer2 migration.
  As of Angular 9, the Renderer class is no longer available.
  Renderer2 should be used instead.
  Read more about this here: https://v9.angular.io/guide/migration-renderer
✔ Migration succeeded.
  No changes to commit after migration.

▸ Undecorated classes with decorated fields migration.
  As of Angular 9, it is no longer supported to have Angular field decorators on a class that does not have an Angular decorator.
  Read more about this here: https://v9.angular.io/guide/migration-undecorated-classes
✔ Migration succeeded.
  No changes to commit after migration.

▸ Undecorated classes with DI migration.
  As of Angular 9, it is no longer supported to use Angular DI on a class that does not have an Angular decorator.
  Read more about this here: https://v9.angular.io/guide/migration-undecorated-classes
✔ Migration succeeded.
  No changes to commit after migration.
```

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -139,6 +139,9 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
     }
   }
 
+  /**
+   * @return Whether or not the migrations were performed successfully.
+   */
   private async executeMigrations(
     packageName: string,
     collectionPath: string,
@@ -183,17 +186,22 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
         return false;
       }
 
+      this.logger.info(colors.green(`${colors.symbols.check} Migration succeeded.`));
+
       // Commit migration
       if (commit) {
-        let message = `${packageName} migration - ${migration.name}`;
-        if (migration.description) {
-          message += '\n' + migration.description;
+        const commitPrefix = `${packageName} migration - ${migration.name}`;
+        const commitMessage = migration.description
+          ? `${commitPrefix}\n${migration.description}`
+          : commitPrefix;
+        const committed = this.commit(commitMessage);
+        if (!committed) {
+          // Failed to commit, something went wrong. Abort the update.
+          return false;
         }
-        // TODO: Use result.files once package install tasks are accounted
-        this.createCommit(message, []);
       }
 
-      this.logger.info(colors.green(`${colors.symbols.check} Migration succeeded.\n`));
+      this.logger.info(''); // Extra trailing newline.
     }
 
     return true;
@@ -556,7 +564,11 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
     });
 
     if (success && options.createCommits) {
-      this.createCommit('Angular CLI update\n' + packagesToUpdate.join('\n'), []);
+      const committed = this.commit(
+          `Angular CLI update for packages - ${packagesToUpdate.join(', ')}`);
+      if (!committed) {
+        return 1;
+      }
     }
 
     // This is a temporary workaround to allow data to be passed back from the update schematic
@@ -590,6 +602,52 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
     return success ? 0 : 1;
   }
 
+  /**
+   * @return Whether or not the commit was successful.
+   */
+  private commit(message: string): boolean {
+    // Check if a commit is needed.
+    let commitNeeded: boolean;
+    try {
+      commitNeeded = hasChangesToCommit();
+    } catch (err) {
+      this.logger.error(`  Failed to read Git tree:\n${err.stderr}`);
+
+      return false;
+    }
+
+    if (!commitNeeded) {
+      this.logger.info('  No changes to commit after migration.');
+
+      return true;
+    }
+
+    // Commit changes and abort on error.
+    try {
+      createCommit(message);
+    } catch (err) {
+      this.logger.error(
+        `Failed to commit update (${message}):\n${err.stderr}`);
+
+      return false;
+    }
+
+    // Notify user of the commit.
+    const hash = findCurrentGitSha();
+    const shortMessage = message.split('\n')[0];
+    if (hash) {
+      this.logger.info(`  Committed migration step (${getShortHash(hash)}): ${
+          shortMessage}.`);
+    } else {
+      // Commit was successful, but reading the hash was not. Something weird happened,
+      // but nothing that would stop the update. Just log the weirdness and continue.
+      this.logger.info(`  Committed migration step: ${shortMessage}.`);
+      this.logger.warn('  Failed to look up hash of most recent commit, continuing anyways.');
+    }
+
+    return true;
+  }
+
   private checkCleanGit(): boolean {
     try {
       const topLevel = execSync('git rev-parse --show-toplevel', { encoding: 'utf8', stdio: 'pipe' });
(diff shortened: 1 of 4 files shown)


packages/angular/cli/commands/update-impl.ts
```diff
@@ -178,17 +178,20 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
...
-        // TODO: Use result.files once package install tasks are accounted
-        this.createCommit(message, []);
+      const commitPrefix = `${packageName} migration - ${migration.name}.`;
+      const commitMessage = migration.description
+          ? `${commitPrefix}\n${migration.description}`
+          : commitPrefix;
+      // TODO: Use result.files once package install tasks are accounted
+      const hash = commit ? this.createCommit(commitMessage, []) : null;
+      if (hash) {
+        this.logger.info(
```
clydin: I think it may be a good idea to log if no changes were made as well and remove any ambiguity as to whether a commit happened or not.

mgechev:
@dgp1130 would you rebase?

packages/angular/cli/commands/update-impl.ts
```diff
@@ -594,16 +603,10 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
...
-  private createCommit(message: string, files: string[]) {
+  private createCommit(message: string, files: string[]): string | null {
     try {
       execSync('git add -A ' + files.join(' '), { encoding: 'utf8', stdio: 'pipe' });
-
       execSync(`git commit --no-verify -m "${message}"`, { encoding: 'utf8', stdio: 'pipe' });
-    } catch (error) {}
-  }
-
-  private findCurrentGitSha(): string | null {
```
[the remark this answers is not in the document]
dgp1130: Done. It was actually useful to keep the functions separate for error handling.

packages/angular/cli/commands/update-impl.ts
```diff
@@ -178,17 +178,20 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
...
-          message += '\n' + migration.description;
-        }
-        // TODO: Use result.files once package install tasks are accounted
-        this.createCommit(message, []);
+      const commitPrefix = `${packageName} migration - ${migration.name}.`;
+      const commitMessage = migration.description
+          ? `${commitPrefix}\n${migration.description}`
+          : commitPrefix;
+      // TODO: Use result.files once package install tasks are accounted
+      const hash = commit ? this.createCommit(commitMessage, []) : null;
```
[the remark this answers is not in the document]
dgp1130: Done, I changed `createCommit()` to not return the hash, so this is obsolete anyways.

Originally I was trying to keep `hash` immutable, but `createCommit()` has a pretty significant side effect, so I agree that should probably be separate.

packages/angular/cli/commands/update-impl.ts
dgp1130: I added a log when a commit is not made, though this pulled the thread on a lot of complex error handling tasks. Hypothetically, any shell command can fail for arbitrary reasons. Previously the CLI just swallowed any `git` errors and silently continued. This could cause problems if one `git commit` fails, but the next succeeds, then the commit will contain extra files with the wrong message.

While working on this, I came across a few edge cases and did my best to handle/fix them:

1. If a commit fails, the CLI swallowed the error and silently continued.
    * Now the CLI logs the command's `stderr` and aborts.
1. A migration may not change any files, `git commit` would then fail because there is nothing to commit.
    * Now the CLI checks if there are changes to commit and just logs if there aren't any.
1. The command to look up the HEAD commit hash may fail, so we can't print it.
    * This is a non-fatal error, the CLI just prints a slightly less descriptive message and logs a warning that something weird happened, since this is unlikely to occur in practice.
1. I also noticed that one migration contained a commit message with a double quote character and was failing due to Bash expansion.
    * Changed the `git commit` command to read the commit message from `stdin`, where Bash escaping isn't a concern.

I did my best to minimize the nest if-statements, but that was difficult to do here. Let me know if you see any opportunities to simplify this.

I also chose not to log that a commit was skipped due to CLI option. I think that is becoming the default now, and I felt it was confusing to explicitly log that we are following default behavior for users who probably aren't even aware of the flag.

dgp1130:
Rebased onto `upstream/master`, should be mergeable now.

dgp1130:
I removed the `-C` option from a few tests which seemed to be used backwards now that the default has changed. This was causing tests to try to commit changes, but CircleCI does not run within a Git repository, so every command was failing. Previously we just ignored these errors, but now they are fatal to the `ng update` process. Easiest solution was to just disable commits for those tests.

packages/angular/cli/commands/update-impl.ts
```diff
@@ -178,17 +181,53 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
...
 
+      this.logger.info(colors.green(`${colors.symbols.check} Migration succeeded.`));
+
       // Commit migration
       if (commit) {
-        let message = `${packageName} migration - ${migration.name}`;
-        if (migration.description) {
-          message += '\n' + migration.description;
+        // Check if a commit is needed.
+        let commitNeeded: boolean;
```
clydin: Can you extract this logic into a common function?  Looks like just the message is unique between the two instances?

dgp1130: Done. I had to change some of the log text, but I think I was able to make it work reasonably well.

dgp1130:
Squashed everything to one commit which should stop CI from complaining about the `fixup!` commit messages. Should be ready to merge assuming all the status checks pass.

PR #16112, opened 2019-11-08
fix(@angular/cli): re-introduce install package using shell spawn

clydin:
This works around a Windows CI failure that related to the PATH environment variable and a failure to located the `npm` command.


alan-agius4:
Should also fix https://github.com/angular/angular-cli/issues/16115

SenthilKumaranC:
How to update angular cli with this patch? 

npm install --save-dev @angular/cli@xxxxxxx

PR #16125, opened 2019-11-08
new ascii art for v9

rgunczer:
Consider a new ASCII art for this new major release v9.0


rgunczer:
@googlebot I signed it!

PR #16126, opened 2019-11-09
fix(@angular-devkit/build-angular): prevent double sourcemap processing

clydin:
Fixes the CI failure for #16065 and provides a noticeable memory improvement.  The failing test from the aforementioned PR went from 2GB to process the main bundle to just over 550MB.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/browser.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/browser.ts
@@ -55,7 +55,7 @@ export function getBrowserConfig(wco: WebpackConfigOptions): webpack.Configurati
     extraPlugins.push(getSourceMapDevTool(
       !!scriptsSourceMap,
       !!stylesSourceMap,
-      hiddenSourceMap,
+      wco.differentialLoadingMode ? true : hiddenSourceMap,
     ));
   }
 


packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -253,6 +253,19 @@ export function buildWebpackBrowser(
       }
 
       const useBundleDownleveling = isDifferentialLoadingNeeded && !options.watch;
+
+      // disable appending sourcemap comments if using bundle downleveling
+      // not doing so will result in double sourcemap comments and a large increase in memory usage
+      // NOTE: Once the webpack config is refactored this should be done directly on the options
+      if (useBundleDownleveling && config.plugins) {
+        const plugin = config.plugins.find(p => p instanceof webpack.SourceMapDevToolPlugin);
```
alan-agius4: Why not set the sourcemap option to hidden, which will have the same result.

https://github.com/angular/angular-cli/blob/862b28f844cddcc406defac6a32d6e266221d499/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/utils.ts#L102

clydin: Updated.  I was hoping to avoid adding more differential loading special casing to the webpack configs themselves.  But they are going to be refactored so it should be ok for now.

PR #16138, opened 2019-11-10
refactor(@angular-devkit/schematics): rename browserslist to .browserslistrc

getabetterpic:
#15961 pointed out that the `browserslist` filename is deprecated. Need to rename it to `.browserslistrc`. Additionally the docs at https://angular.io/guide/build#configuring-browser-compatibility specifically mention `.browserslistrc` and not `browserslist`.

This PR renames the template file as well as updates the Angular 8 differential loading migrator. Updated the specs around the migrator as well.


filipesilva:
Discussed in meeting, we think we should move to the current default (`.browserslistrc`), but feel it's a risky proposition during the RC period.

Instead we'd like to:
- change new projects something during 9.x to use the new filename
- add a migration for 10.0 to update the filename

You should be able to just use the new filename, but we need to make sure to update users for 10.0 to also do it in order to keep everything consistent. We should not, however, change old migrations.

KingDarBoja:
There is another PR which refers to `browserlist` renaming issue, #15977. Just to help tracking related issues to avoid future duplicates.

Cheers!

alan-agius4:
Hi, thanks for your contribution. 

Let's close this for the time being and reconsider this for version 10.

More context: https://github.com/angular/angular-cli/pull/16138#issuecomment-554014906

PR #16152, opened 2019-11-11
docs(@angular/cli): Adds developer documentation and Intellij setup.

dgp1130:
I wrote down my understanding of the best ways to build/run/test/debug this repository. I also included some configs for Intellij to work a little bit better "out of the box".

I tried to get VSCode to work as well, however the debugger there was having trouble loading source maps, so I'll need to come back to that later.

Changes include:
* Updated README with instructions for new developers.
* Updated .gitignore for Intellij and VSCode.
* Added Intellij config files.
* Added `nodemon` as a dev dependency, which is needed for watch executions but wasn't listed.
* Removed an extraneous `debugger;` statement which I kept hitting.
* Added support for negated .gitignore expressions in the build tool, as my changes to that file were causing build errors.


README.md
```diff
@@ -44,16 +44,115 @@ analyze your code.
...
+
+* Unit tests
+  * Run: `yarn test --full`
+  * Debug: `yarn debug:test --full`
+  * Watch: `yarn test:watch --full`
+* Large tests
+  * Run: `yarn test-large --full`
+  * Debug: `yarn debug:test-large --full`
+* End to end tests
+  * Run: `yarn test-cli-e2e`
```
clydin: This won't work properly.  The package manager adjusts the path environment variable and can cause the wrong `ng` instance to run.

README.md
```diff
@@ -44,16 +44,115 @@ analyze your code.
...
+Do **not** `Import Project`, because that will overwrite the existing
+configuration.
+
+Once opened, the editor should automatically detect run configurations in the
+workspace. Use the drop down to choose which one to run and then click the `Run`
+button to start it. When executing a debug target, make sure to click the
+`Debug` icon to automatically attach the debugger (if you click `Run`, Node will
+wait forever for a debugger to attach).
+
+![Intellij IDEA run configurations](docs/images/run-configurations.png)
```
clydin: This doesn't appear to exist.

package.json
```diff
@@ -128,6 +128,7 @@
     "license-checker": "^20.1.0",
     "minimatch": "^3.0.4",
     "minimist": "^1.2.0",
+    "nodemon": "^1.19.4",
```
clydin: Unintentional?

README.md
dgp1130: What would be the correct way to run this then?

README.md
dgp1130: You're talking about the image? It should be included in this commit:

9f982fd?short_path=00879e8#diff-00879e81852f7d42256cd14fb6e799d1

package.json
dgp1130: Removed the dependency and the reference in the documentation. I also removed the script in `package.json`.

dgp1130:
I split up the commits and moved my README edits to `scripts/templates/readme.ejs`, which I didn't see before.

README.md
clydin: `node ./tests/legacy-cli/run_e2e.js`

dgp1130: Updated the docs and removed the `test-cli-e2e` Yarn script so no one else finds and uses it to shoot themselves in the foot.

dgp1130:
Do you think we need to discuss the IntelliJ configs with the team or should I just merge this?

clydin:
To be safe, added the discussion label for tomorrow's triage meeting.

dgp1130:
Concerns in the meeting were able not forcing personal configuration settings (such as themeing) directly in the repo. IntelliJ doesn't really put personal settings like that in these files. The settings will be picked up by default, but users should be able to change and modify them as they wish independently from the setting stored in source control.

PR #16154, opened 2019-11-12
LTS version of #16151

alan-agius4:
See: #16151


alan-agius4:
@googlebot I consent.

alan-agius4:
@mgechev this is good to merge, Ci statuses are pending because they are not available in the v8 branch.

PR #16169, opened 2019-11-12
fix(@angular-devkit/build-webpack): rewrite ES5 polyfills file name

dgp1130:
The asset name is incorrect because the file is renamed after generation by Webpack, so the stats file contains a broken link. This hacks together a quick rewrite to the asset list to use the new filename.


dgp1130:
Charles found a way to do this with a Webpack plugin in a much less hacky way. Closing this PR in favor of that one.

https://github.com/angular/angular-cli/pull/16181

PR #16178, opened 2019-11-13
fix(@angular-devkit/build-ng-packagr): add tsickle as optional peer dep

filipesilva:
This way it'll be automatically updated if present.

### DIFF
--- a/packages/angular_devkit/build_ng_packagr/package.json
+++ b/packages/angular_devkit/build_ng_packagr/package.json
@@ -11,7 +11,8 @@
     "rxjs": "6.5.3"
   },
   "peerDependencies": {
-    "ng-packagr": "^9.0.0-rc.0"
+    "ng-packagr": "^9.0.0-rc.0",
+    "tsickle": "~0.37.1"
   },
   "devDependencies": {
     "@angular/compiler": "9.0.0-rc.1",
@@ -19,5 +20,10 @@
     "@angular-devkit/core": "0.0.0",
     "ng-packagr": "~9.0.0-rc.0",
     "tslib": "^1.10.0"
+  },
+  "peerDependenciesMeta": {
+    "tsickle": {
+      "optional": true
+    }
   }
 }


IgorMinar:
this is cool @filipesilva! I didn't know that npm and yarn now supported optional peerDeps.

IgorMinar:
btw I came across some issues with tsickle 0.37.1 which we might need to be aware of: https://github.com/angular/angular/pull/33788

dgp1130:
@IgorMinar, is this safe to merge or are those issues blocking for this PR as well?

filipesilva:
@IgorMinar npm added support in https://github.com/npm/cli/pull/224 and yarn added in https://github.com/yarnpkg/yarn/issues/6487.

packages/angular_devkit/build_ng_packagr/package.json
```diff
@@ -11,13 +11,19 @@
     "rxjs": "6.5.3"
   },
   "peerDependencies": {
-    "ng-packagr": "^9.0.0-rc.0"
+    "ng-packagr": "^9.0.0-rc.0",
+    "tsickle": "~0.37.1"
```
alan-agius4: I was under the impression that the fix for this was the include the dependency in the packageGroup.

filipesilva: Yeah that was the original idea, but then Charles brough the optional peer dep up that both ensures it is updated, and ensures that if it is present on install (outside of update) it should be the right version. Do you think there are problems with this approach?

alan-agius4: Not at all, I was just curios to know why the change :)

clydin:
Also of note, `npm` had a defect that was fixed in 6.13.0 (currently the latest release) that would cause an incorrect warning about the peer dependency being missing but only on the first install.
It was originally thought to be a registry server only issue.  However, upon fixing the npm registry it was found that one of npm's dependencies also needed a fix as well.  But as of today yarn, npm, and pnpm all support optional peer dependencies.

filipesilva:
Discussed in a meeting, it's safe to do this because we will be removing it right after during the migration. This will just correct the peerdep failure if any exists.

dgp1130:

PR #16180, opened 2019-11-13
fix(@schematics/angular): uniformize guard spec with service spec

cexbrayat:
The current guard spec schematic uses `inject` whereas we dropped it for the service spec a while ago.
This commits updates the guard spec schematic to be similar to the current service spec.


dgp1130:

PR #16181, opened 2019-11-13
fix(@angular-devkit/build-angular): prevent webpack from adding suffixes to polyfill files

### LINKED ISSUE #15915: webpack-bundle-analyzer errors out on differential build - no such file
# 🐞 Bug report
As I wanted to analyze my bundles I added `webpack-bundle-analyzer` to my project.
Then I used the command
```bash
ng build --prod --statsJson && webpack-bundle-analyzer ./dist/app/stats-es2015.json
```
result:
image

This happened in mutliple of my projects. The common demoninator seems to be diffential loading. As soon as that is turned on, I get this error.

clydin:
The ES5 polyfills file was erroneously being suffixed with `es2015`.  The webpack configuration does not support conditional customization per chunk for the output filenames (`chunkFilename` option schema only supports string values).  This change adds an additional small webpack plugin that allows the chunk filenames to be adjusted based on the chunk name.  The plugin is only added when differential loading is enabled as this is the only time that a chunk currently requires its filename to be adjusted.

Closes #15915

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -15,10 +15,12 @@ import * as path from 'path';
 import { RollupOptions } from 'rollup';
 import { ScriptTarget } from 'typescript';
 import {
+  ChunkData,
   Compiler,
   Configuration,
   ContextReplacementPlugin,
   HashedModuleIdsPlugin,
+  Plugin,
   Rule,
   compilation,
   debug,
@@ -57,10 +59,12 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
     throw new Error('Cannot locate node_modules directory.');
   }
 
-  // tslint:disable-next-line:no-any
-  const extraPlugins: any[] = [];
+  const extraPlugins: Plugin[] = [];
   const extraRules: Rule[] = [];
   const entryPoints: { [key: string]: string[] } = {};
+
+  // determine hashing format
+  const hashFormat = getOutputHashFormat(buildOptions.outputHashing || 'none');
 
   const targetInFileName = getEsVersionForFileName(
     tsConfig.options.target,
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -146,6 +150,28 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
+                // Webpack typings do not contain MainTemplate assetPath hook
+                // The webpack.Compilation assetPath hook is a noop in 4.x so the template must be used
+                // tslint:disable-next-line: no-any
+                (compilation.mainTemplate.hooks as any).assetPath.tap(
+                  'build-angular',
+                  (filename: string, data: ChunkData) => {
+                    const isMap = filename && filename.endsWith('.map');
+
+                    return data.chunk && data.chunk.name === 'polyfills-es5'
+                      ? `polyfills-es5${hashFormat.chunk}.js${isMap ? '.map' : ''}`
```
filipesilva: I'm a bit confused here, this doesn't seem to change the name of `polyfills-es5-es2015.js`, instead it seems to target `polyfills-es5.js`.

clydin: The `assetPath` hook is used by webpack to create the output filename/path for each chunk.   The filename that is passed as an argument to the hook is actually the filename template from the webpack configuration (defined [here](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts#L477) in the CLI).  Chunk names are defined via the entry point configuration (`polyfills-es5` is defined [here](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts#L144)).
This plugin allows the template for the output filename to be altered on a per chunk basis.  In this case, the `polyfills-es5` chunk's filename template is changed to remove the `targetInFileName` segment (`-es2015` in the differential loading case).  This causes webpack to generate the correct name and removes the need to patch the stats file and clean up the filenames after the fact.

filipesilva: Ok I see now, thank you. The logic is a bit circuitous and I worry about someone else altering this code. Can you add it as a comment please?

dgp1130:

Issue #13059, opened 2018-11-27
ng update fails when using local library


elboukhari:
Bug Report or Feature Request (mark with an x)
- [X ] bug report -> please search issues before submitting
- [ ] feature request
Command (mark with an x)
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
Versions
node --version : v10.13.0
npm --version : 6.4.1
ng --version :

`Angular CLI: 7.1.0
Node: 10.13.0
OS: linux x64
Angular: 7.1.0
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, http, language-service, material
... material-moment-adapter, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.11.0
@angular-devkit/build-angular     0.11.0
@angular-devkit/build-optimizer   0.11.0
@angular-devkit/build-webpack     0.11.0
@angular-devkit/core              7.1.0
@angular-devkit/schematics        7.1.0
@ngtools/webpack                  7.1.0
@schematics/angular               7.1.0
@schematics/update                0.11.0
rxjs                              6.3.3
typescript                        3.1.6
webpack                           4.26.1`

Repro steps
just type :
ng update

The log given by the failure
Not found : XXXXXX ()

In my package.json this library is declared as :

I'm still having the issue even with :
"@angular/cli": "7.1.0"

the error is the same as in the previous comment:
404 Not Found - GET https://registry.npmjs.org/XXXXXXX - Not found

Desired functionality
just ignore local libraries and update angular ones

Mention any other details that might be useful

alan-agius4:
Hi, when you say local package? is it in a private npm repo? or using the `file://` protocol?

elboukhari:
it's ` file://` protocol, here is the corresponding line in `package.json` : 

`"MY-AWeSOME-LOCAL-LIB": "file:lib/MY-AWeSOME-LOCAL-LIB.tgz",`

meta72:
@alan-agius4 I have the same issue with a private npm repo. When I update any node module it says that it can't find my private module (which is on my private repo) on npmjs.org. This was working until cli version 7.0.7

denisoby:
[...]

PR #16185, opened 2019-11-13
fix(@angular-devkit/schematics): filters out non-NPM dependencies (#13059)

dgp1130:
Non-NPM dependencies in package.json are now filtered out of `ng update`. Dependencies which are linked as tarball or Git references will now be skipped during the update process.


packages/schematics/update/update/index.ts
```diff
@@ -850,11 +873,16 @@ export default function(options: UpdateSchema): Rule {
 
   return (tree: Tree, context: SchematicContext) => {
     const logger = context.logger;
-    const allDependencies = _getAllDependencies(tree);
-    const packages = _buildPackageList(options, allDependencies, logger);
+    const allDeps = _getAllDependencies(tree);
+
+    // Filter out non-NPM dependencies by checking if they link to a semver.
+    const npmDeps = new Map(Array.from(allDeps.entries())
+        .filter(([_, source]) => isSimpleSemVer(source)));
```
clydin: Consider using [npm-package-arg's](https://github.com/npm/npm-package-arg) `resolve` function to both validate and check for a registry specifier.  This is the logic that npm itself uses.

dgp1130: Thanks for the tip, all my Google searches were leading me to the `semver` package. Updated to use `npm-package-arg` and also added a test to verify the behavior.

packages/schematics/update/update/index.ts
```diff
@@ -826,6 +827,13 @@ function _formatVersion(version: string | undefined) {
   return version;
 }
 
+// Returns whether or not the given package spec (the value string in a
+// `package.json` dependency) is hosted in the NPM registry.
+function isPkgFromRegistry(name: string, spec: string): boolean {
+  const result = npa.resolve(name, spec);
```
clydin: This can throw if the name or spec is invalid.  In those cases, this should probably return false since update will fail when it attempts to analyze the package.

dgp1130: I was thinking that in the case of an invalid package name, we should just abort and let the update process crash. If this throws, then `package.json` must be malformed or corrupted and would not be usable by NPM either. Is that a case where we would want to the update process to suppress the error and skip that package?

dgp1130:
I had to install `npm-package-arg` which created the `package-lock.json`. Is that not desirable for this project? I figured dependencies just hadn't been added since `package-lock.json` became a thing. Is there a specific reason we're trying to avoid that here?

packages/schematics/update/update/index.ts
clydin: Barring catastrophic errors such as out of memory exceptions, the CLI itself should ideally never crash.
However, since the update logic is currently wrapped in a schematic, the only way to report an error is to throw an exception so that should work.  The exception messages should probably be augmented and rethrown with additional information/instructions regarding the failure so that they are actionable by the user though.

clydin:
`package-lock.json` is npm's lock file.  The repository is setup to use yarn workspaces (https://github.com/angular/angular-cli/blob/master/package.json#L56).  Once the dependency is added to the package.json for one of the repository packages, executing `yarn` within the repository will install in the appropriate locations for the package and update the root lock file.

This package also has a Bazel BUILD file which will need to be updated to include the new dependency as well (https://github.com/angular/angular-cli/blob/master/packages/schematics/update/BUILD#L24).

packages/schematics/update/update/index.ts
dgp1130: "crash" was the wrong word for me to use, but yes I agree. Updated to rethrow a more meaningful error message to the user. This does throw directly from the function rather than emit an error from a returned `Observable`. I think this is ok based on the fact that `_getAllDependencies()` also throws and I'm assuming there is a global error handler which will catch and print the message.

dgp1130:
Thanks for explaining the lock file, I didn't realize `yarn.lock` applied to every package in the workspace. Removed the extra `package-lock.json` and ran `yarn` to update `yarn.lock`.

Thanks for pointing out the `BUILD` file, I didn't think to look for one. I added what I think is right for my changes, but attempting to build it fails due to missing `@types/npm-package-arg`:

```
douglasparker@gryphon (~/Source/ng-cli)
[19-11-15 10:16:41]$ yarn bazel build //packages/schematics/update
yarn run v1.19.1
$ /usr/local/google/home/douglasparker/Source/ng-cli/node_modules/.bin/bazel build //packages/schematics/update
INFO: Writing tracer profile to '/usr/local/google/home/douglasparker/.cache/bazel/_bazel_douglasparker/fd23369152ca483d39c1138566f81234/command.profile.gz'
INFO: Build option --legacy_external_runfiles has changed, discarding analysis cache.
ERROR: /usr/local/google/home/douglasparker/Source/ng-cli/packages/schematics/update/BUILD:14:1: no such package '@npm//@types/npm-package-arg': BUILD file not found in directory '@types/npm-package-arg' of external reposi
tory @npm and referenced by '//packages/schematics/update:update'
ERROR: Analysis of target '//packages/schematics/update:update' failed; build aborted: no such package '@npm//@types/npm-package-arg': BUILD file not found in directory '@types/npm-package-arg' of external repository @npm
INFO: Elapsed time: 18.172s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (7 packages loaded, 6 targets configured)
    currently loading: @npm//rxjs ... (5 packages)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

My understanding of [Bazel's NodeJS toolchain](https://bazelbuild.github.io/rules_nodejs/install.html) is that it should install these dependencies from the `yarn.lock` file. Is there something else I need to do to include this dependency.

Follow up question, where is this Bazel rule used if not by `yarn build` or `yarn test`? Is this simply an artifact of an ongoing migration to Bazel, or is there some system which uses it instead of `yarn`?

packages/schematics/update/package.json
```diff
@@ -13,11 +13,13 @@
   "dependencies": {
     "@angular-devkit/core": "0.0.0",
     "@angular-devkit/schematics": "0.0.0",
+    "@types/npm-package-arg": "^6.1.0",
```
clydin: These type of dependencies are build time only and should be placed in the root package.json.

This should also fix the bazel build issue

dgp1130: I see, moved the dependency to the root `package.json`. I also made a dev dependency, since I don't think it would necessary at runtime (though the way `scripts/` are loaded with a require extension, I think they might require types at runtime, but I don't think schematics are loaded in that fashion). `yarn bazel build //packages/schematics/update` now succeeds.

Just out of curiosity, why are build-time dependencies required in the root `package.json`? I would think each package would include all the dependencies necessary to build them. Is this required for Bazel to generate the NPM build targets?

packages/schematics/update/update/index.ts
```diff
@@ -847,14 +858,23 @@ export default function(options: UpdateSchema): Rule {
...
-    const allDependencies = _getAllDependencies(tree);
-    const packages = _buildPackageList(options, allDependencies, logger);
-    const usingYarn = options.packageManager === 'yarn';
+    const npmDeps = new Map(_getAllDependencies(tree).filter(([name, spec]) => {
+      try {
+        return isPkgFromRegistry(name, spec);
+      } catch {
+        // Abort on failure because package.json is malformed.
+        throw new SchematicsException(
+          `Failed to parse dependency "${name}" with spec "${spec}" from`
```
clydin: nit: spell out specifier

dgp1130: Done.

dgp1130:
Merged in to 9.0.x patch branch. Fixed a minor conflict due to a version update to `pacote` which was not present in the patch branch.

PR #16189, opened 2019-11-14
fix(@angular-devkit/build-angular): fix numberOfComponents in JIT mode

alan-agius4:
Current we are always sending `0` as the number of components in JIT mode because ` Component({` doesn't match the component source

### DIFF
--- a/packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
+++ b/packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
@@ -158,7 +158,7 @@ export class NgBuildAnalyticsPlugin {
 
       // Count the number of `Component({` strings (case sensitive), which happens in __decorate().
       // This does not include View Engine AOT compilation, we use the ngfactory for it.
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/plugins/webpack/analytics.ts
```diff
@@ -158,7 +158,7 @@ export class NgBuildAnalyticsPlugin {
 
       // Count the number of `Component({` strings (case sensitive), which happens in __decorate().
       // This does not include View Engine AOT compilation, we use the ngfactory for it.
-      this._stats.numberOfComponents += countOccurrences(module._source.source(), ' Component({');
+      this._stats.numberOfComponents += countOccurrences(module._source.source(), 'Component({');
```
clydin: Just to clarify, the beginning space was the issue?

alan-agius4: Yes

dgp1130:

Issue #16191, opened 2019-11-14
ng update @angular/cli  always output list of updated file


elvisbegovic:
# 🐞 Bug report

### Command (mark with an `x`)

- [x] update

### Is this a regression?

I don't know

### Description

Run many times same `ng update` command migrate only on package `@angular/cli` from v8, CLI always output all files that are updated
![image](https://user-images.githubusercontent.com/2789203/68858526-86f98b80-06e5-11ea-910e-68ec6eeb2799.png)

## 🔬 Minimal Reproduction

One line repro:
`ng new clibug --minimal --defaults && cd clibug && npm install && ng update @angular/cli --migrateOnly --from=8 --allowDirty && ng update @angular/cli --migrateOnly --from=8 --allowDirty && ng update @angular/cli --migrateOnly --from=8 --allowDirty`

## 🌍 Your Environment
<details><summary>See Environment version</summary>
<pre><code>
Angular CLI: 9.0.0-rc.2
Node: 13.1.0
OS: win32 x64
Angular: 9.0.0-rc.2
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.900.0-rc.2
@angular-devkit/build-angular     0.900.0-rc.2
@angular-devkit/build-optimizer   0.900.0-rc.2
@angular-devkit/build-webpack     0.900.0-rc.2
@angular-devkit/core              9.0.0-rc.2
@angular-devkit/schematics        9.0.0-rc.2
@ngtools/webpack                  9.0.0-rc.2
@schematics/angular               9.0.0-rc.2
@schematics/update                0.900.0-rc.2
rxjs                              6.5.3
typescript                        3.6.4
webpack                           4.41.2
</code></pre>
</details>

**Anything else relevant?**
cc @alan-agius4

PR #16194, opened 2019-11-14
fix(@angular-devkit/schematics): only commit to disk when contents of file changes 

alan-agius4:
See each commit 
Fixes: #16191

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/host-tree.ts
+++ b/packages/angular_devkit/schematics/src/tree/host-tree.ts
@@ -341,7 +341,9 @@ export class HostTree implements Tree {
         throw new ContentHasMutatedException(path);
       } else {
         const newContent = record.apply(entry.content);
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/update-9/ivy-libraries.ts
```diff
@@ -96,7 +100,9 @@ export function updateLibraries(): Rule {
       }
     }
 
-    tree.commitUpdate(recorder);
+    if (hasUpdate) {
+      tree.commitUpdate(recorder);
```
clydin: This should really be the inherent behavior of `commitUpdate` but for now this works.

alan-agius4: Actually you gave me an idea :)

dgp1130:

PR #16196, opened 2019-11-14
Re-enable global locale data support for Ivy

clydin:
This also adds global local data support to the development server.

### DIFF
--- a/tests/legacy-cli/e2e/tests/i18n/ivy-localize-dl.ts
+++ b/tests/legacy-cli/e2e/tests/i18n/ivy-localize-dl.ts
@@ -33,8 +33,8 @@ export default async function() {
     await expectFileToMatch(`${outputPath}/main-es2015.js`, lang);
 
     // Verify the locale data is registered using the global files
-    // await expectFileToMatch(`${outputPath}/main-es5.js`, '.ng.common.locales');
-    // await expectFileToMatch(`${outputPath}/main-es2015.js`, '.ng.common.locales');
+    await expectFileToMatch(`${outputPath}/main-es5.js`, '.ng.common.locales');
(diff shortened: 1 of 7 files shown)


tests/legacy-cli/e2e/tests/i18n/ivy-localize-dl.ts
```diff
@@ -14,27 +14,21 @@ export default async function() {
...
+  for (const { lang, outputPath, translation } of langTranslations) {
     await expectFileToMatch(`${outputPath}/main-es5.js`, translation.helloPartial);
     await expectFileToMatch(`${outputPath}/main-es2015.js`, translation.helloPartial);
     await expectToFail(() => expectFileToMatch(`${outputPath}/main-es5.js`, '$localize`'));
     await expectToFail(() => expectFileToMatch(`${outputPath}/main-es2015.js`, '$localize`'));
 
     // Verify the locale ID is present
-    await expectFileToMatch(`${outputPath}/main-es5.js`, lang);
-    await expectFileToMatch(`${outputPath}/main-es2015.js`, lang);
+    await expectFileToMatch(`${outputPath}/vendor-es5.js`, lang);
```
petebacondarwin: How come the locale is written to `vendor` rather than `main`?

clydin: It's written to the same location to that of the Angular FW code.  The locale data is also placed there for builds since it is technically vendor code.
For an actual production build all of it would be in main since there is no vendor bundle.

dgp1130:
Merged to `9.0.x` patch branch: https://github.com/angular/angular-cli/commits/9.0.x.

PR #16203, opened 2019-11-16
fix(@schematics/angular): don't create e2e script when createApplication is false

### LINKED ISSUE #13412: e2e script is created in package.json when --createApplication=false
```
- [ ] feature request
```

```
- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions
Node: v10.13.0
NPM: v6.4.1
Angular CLI: v7.1.3

### Repro steps

1. ```ng new --createApplication=false```
2. e2e script is added to package.json

### Desired functionality

The e2e script should not be added to the package.json because no e2e directory or functionality is generated when createApplication flag is false.

timdeschryver:
Bugfix for the `ng new --createApplication=false` command.
Currently, it creates an `e2e` script in `package.json`. This change will only add the script when the application is created.

Closes #13412

### DIFF
--- a/packages/schematics/angular/e2e/index.ts
+++ b/packages/schematics/angular/e2e/index.ts
@@ -17,10 +17,22 @@ import {
   move,
   url,
 } from '@angular-devkit/schematics';
+import { JSONFile } from '../utility/json-file';
 import { relativePathToWorkspaceRoot } from '../utility/paths';
 import { getWorkspace, updateWorkspace } from '../utility/workspace';
 import { Builders } from '../utility/workspace-models';
 import { Schema as E2eOptions } from './schema';
+
+function addScriptsToPackageJson(): Rule {
+  return host => {
+    const pkgJson = new JSONFile(host, 'package.json');
+    const e2eScriptPath = ['scripts', 'e2e'];
+
+    if (!pkgJson.get(e2eScriptPath)) {
+      pkgJson.modify(e2eScriptPath, 'ng e2e', false);
+    }
+  };
+}
 
 export default function (options: E2eOptions): Rule {
   return async (host: Tree) => {
@@ -65,6 +77,7 @@ export default function (options: E2eOptions): Rule {
           }),
           move(root),
         ])),
+      addScriptsToPackageJson(),
     ]);
   };
 }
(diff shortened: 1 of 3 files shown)


clydin:
However, I don’t think this is the correct approach for this.
If the E2E npm script should only be present when an application is added to the application, then I think that the application schematic should be adding it.  This allows the workspace schematic to remain agnostic to the concept of an application.

alan-agius4:
I also, do agree with @clydin

timdeschryver:
I agree that it's a valid point raised, I did it because I thought to treat it the same as the minimal flag.

I will push a commit later with the changes.
So the workspace schematic won't have the e2e script, but it will be added via the application schematic, right? Or does the workspace schematic remain as is, and will the application schematic remove the e2e script if needed?

clydin:
I think the former (removing it from the workspace generation) would be the appropriate path.

packages/schematics/angular/utility/scripts.ts
```diff
@@ -0,0 +1,47 @@
...
+  if (!scriptsNode) {
+    // Haven't found the scripts key, add it to the root of the package.json.
+    appendPropertyInAstObject(recorder, packageJsonAst, 'scripts', {
+      [script.name]: script.script,
+    }, 2);
+  } else if (scriptsNode.kind === 'object') {
+    // check if script already added
+    const scriptNode = findPropertyInAstObject(scriptsNode, script.name);
+
+    if (!scriptNode) {
```
timdeschryver: This is taken from the dependencies utilities.
Except it doesn't overwrite the script value if it's already added.

timdeschryver:
Should this be part of the `e2e` schematic instead?

alan-agius4:
Hi, @timdeschryver, sorry for taking us long to get back to you about this.

> Should this be part of the `e2e` schematic instead?

Yes, I think that's the best approach.

timdeschryver:
Should be ready for another review @alan-agius4 
If there's something I can do, please let met know.

alan-agius4:
Hi @timdeschryver, sorry for the delay on this. I rebased the PR and did some changes to get this in!

Thanks and sorry once again.

timdeschryver:
@alan-agius4 no need to say sorry.
Thanks for the help!

PR #16206, opened 2019-11-16
fix(@angular-devkit/build-angular): allow new i18n options to work with VE

clydin:
This allows one set of options to be used for localization regardless of Ivy usage.  Also simplifies Ivy opt out.


dgp1130:

PR #16210, opened 2019-11-18
fix(@schematics/angular): migrate project to use new i18n options

alan-agius4:


packages/schematics/angular/migrations/update-9/update-workspace-config.ts
```diff
@@ -107,12 +114,33 @@ function addBuilderI18NOptions(recorder: UpdateRecorder, builderConfig: JsonAstO
...
+    if (localeId && localeId.kind === 'string') {
+      // add new localize option
+      insertPropertyInAstObjectInOrder(recorder, option, 'localize', [localeId.value], 12);
     }
 
-    // add new localize option
-    insertPropertyInAstObjectInOrder(recorder, option, 'localize', [localeId.value], 12);
+    // deprecated options
+    removePropertyInAstObject(recorder, option, 'i18nLocale');
+    removePropertyInAstObject(recorder, option, 'i18nFile');
```
clydin: If any of these are found and removed during the migration, two additional steps are needed:
1) install of the localize package
2) add localize option to the the main option set for the builder.  This can be sourced from the extraction options (if present), defaulted to “en-US”, or the user could be asked.

alan-agius4: PTAL

dgp1130:

PR #16214, opened 2019-11-18
refactor(@angular/cli): update Intellij/Webstorm settings

ahasall:
Intellij and Webstorm unfortunately don't the use the exact same
configuration system. For example, Intellij treats every project as
`JAVA_MODULE` but Webstorm doesn't recognize the `JAVA_MODULE`.
Therefore, I think that the modules.xml and iml should be left
out from version control.

refs #16152

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@ jsconfig.json
 
 # Intellij IDEA/WebStorm
 # https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+.idea/inspectionProfiles/
+.idea/**/compiler.xml
+.idea/**/encodings.xml
 .idea/**/workspace.xml
 .idea/**/tasks.xml
 .idea/**/usage.statistics.xml
--- a/.idea/angular-cli.iml
+++ b/.idea/angular-cli.iml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<module type="JAVA_MODULE" version="4">
+<module type="WEB_MODULE" version="4">
   <component name="NewModuleRootManager" inherit-compiler-output="true">
     <exclude-output />
     <content url="file://$MODULE_DIR$">


ahasall:
The following files should also be discussed:
* `.idea/compiler.xml`
* `.idea/encodings.xml`
* `.idea/inspectionProfiles`

cc @dgp1130

dgp1130:
Thanks for bringing up your experience with WebStorm. I wasn't aware projects were structured differently between the two platforms.

I tried checking out your commit and opening it in IntelliJ Ultimate, however, it doesn't load the project, likely because the `*.iml` file is missing. It seems that [JetBrains' recommendation](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) is to include `*.iml` files, so I would assume WebStorm would work with that. What actually is the behavior in WebStorm when the `*.iml` file is present?

Regarding the extra files you mentioned:
* `./idea/compiler.xml` - This only seems to affect Java projects as far as I can tell. Does this have any effect for Node projects? If not, I'm inclined to just `.gitignore` it.
* `./idea/encodings.xml` - I think this just specifies encodings for particular files. I don't think there are any files in the repo which require special encoding that IntelliJ couldn't figure out on its own. I figure we can either `.gitignore` it, or include one which just default to `UTF-8`.
* `./idea/inspectionProfiles/` - I don't think we have a highly specified style guide for this repo, so I don't think we can really make intelligent decisions on which inspections to enable/disable. There might be some inspections which obviously don't work and could be turned off here, but otherwise I think I'd rather leave this up to the user to configure at their leisure. I believe IntelliJ already supports custom profiles here external to the project, so I don't think we need to do much here.

ahasall:
I was having the following warning every time I open the project:

```
Unknown Module Type
Cannot determine module type ("JAVA_MODULE") for the following module:"angular-cli"
The module will be treated as a Unknown module.
```

WebStorm doesn't know the module type `JAVA_MODULE`. It uses the module type `WEB_MODULE` which seems to work well with Intellij.

I updated the PR, could you please try again?

dgp1130:
Tried again, and this configuration seems to work for IntelliJ Ultimate. Agreed that `Web Module` is more appropriate than `Java Module`.

LGTM, pending update to your commit message as we're no longer removing the `*.iml` file.

ahasall:
Yes! Just amended the commit message

dgp1130:
Thanks for catching this problem and sending the fix!

ahasall:
It's my pleasure!

Issue #16205, opened 2019-11-16
Angular CLI 8.3.19 ng update install temp cli


lneumeier:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 8.3.17 or 8.3.18

### Description
Running 
```
ng update
```
installs the latest angular-cli into a temp directory despite the fact, that the global installed angular-cli version is already installed. Why not use the global one?

logs:
```
ng update
Your global Angular CLI version (8.3.19) is greater than your local
version (8.3.18). The local Angular CLI version is used.

To disable this warning use "ng config -g cli.warnings.versionMismatch false".
The installed Angular CLI version is older than the latest published version.
Installing a temporary version to perform the update.
Installing packages for tooling via npm.

> @angular/cli@8.3.19 postinstall /private/var/folders/cy/4hs_g4mj4jlg6nyk_9vyxg9w0000gn/T/.ng-temp-packages-rFmoyk/node_modules/@angular/cli
> node ./bin/postinstall/script.js

npm WARN saveError ENOENT: no such file or directory, open '/private/var/folders/cy/4hs_g4mj4jlg6nyk_9vyxg9w0000gn/T/.ng-temp-packages-rFmoyk/package.json'
npm WARN enoent ENOENT: no such file or directory, open '/private/var/folders/cy/4hs_g4mj4jlg6nyk_9vyxg9w0000gn/T/.ng-temp-packages-rFmoyk/package.json'
npm WARN .ng-temp-packages-rFmoyk No description
npm WARN .ng-temp-packages-rFmoyk No repository field.
npm WARN .ng-temp-packages-rFmoyk No README data
npm WARN .ng-temp-packages-rFmoyk No license field.
```

After that running 
```
ng update @angular/cli @angular/core
```
another temp install happens.

ErVishalM:
I am also getting the same error.

simpers:
[...]

PR #16215, opened 2019-11-18
fix(@angular/cli): fix package manager warnings when running temporary install

alan-agius4:
Related to: #16205


dgp1130:

PR #16216, opened 2019-11-18
fix(@angular-devkit/build-angular): mocked context in single test transform matches webpack API

FDIM:
I've ran into an issue at work with the mocked context in single test transform - it does not match [webpack API](https://webpack.js.org/guides/dependency-management/#requirecontext).

Basically I have to filter out certain values from `require.context(...).keys()` before importing the actual file to fix another issue in `test.ts`. (Desired outcome is for code coverage to include untested ts files)

Currently with angular 8 (and master branch) it fails when running `ng test --include <file>`, as the value returned from keys does not have a `filter` method. This PR fixes that by returning an empty array instead of partial object.

@filipesilva since you approved my original PR, could you give a quick look here? It is a very minor change :)


filipesilva:
caretaker: please ignore the size check, this PR doesn't affect sizes.

PR #16228, opened 2019-11-19
Update ivy set calls

filipesilva:


filipesilva:
This PR is blocked by https://github.com/angular/angular/pull/33671 and https://github.com/angular/angular/pull/33337 being merged and released. Afterwards, this PR should be updated to use the new FW version.

packages/angular_devkit/build_optimizer/src/transforms/scrub-file.ts
```diff
@@ -304,7 +303,35 @@ function isAssignmentExpressionTo(exprStmt: ts.ExpressionStatement, name: string
 }
 
 function isIvyPrivateCallExpression(exprStmt: ts.ExpressionStatement) {
```
clydin: nit: The IIFE structure could technically be different depending on any intermediate processing steps.  The CLI's fixed configuration will prevent this but other usage may cause changes.  An unbounded number of parenthesized expressions are also technically possible.

Examples: `!function(){}()` or `(function(){}())` or `((function(){})())`

PR #16240, opened 2019-11-20
fix(@angular-devkit/build-angular): update link to update.angular.io

IgorMinar:
This link is rarely seen by developers but when they do see it, they should get the real one.


dgp1130:
Merged to 9.0.x patch branch.

PR #16268, opened 2019-11-22
fix(@angular-devkit/build-angular): fix budget checks

### LINKED ISSUE #15792: Initial bundle logs and budget warning are working wrong
# 🐞 Bug report

```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

Yes, the previous version in which this bug was not present was: 8.2.0

### Description
See the my production build logs:
```
chunk {4} polyfills-es2015.8f11f9b5202829aea96f.js (polyfills) 36.4 kB [initial] [rendered]
chunk {5} polyfills-es5.509fc4fe9eb4ca689d44.js (polyfills-es5) 123 kB [initial] [rendered]
chunk {10} 10-es2015.c32b61866de840ef129e.js () 493 kB  [rendered]
chunk {10} 10-es5.c32b61866de840ef129e.js () 493 kB  [rendered]
chunk {1} 1-es2015.067123cd46b49e92639e.js () 58.6 kB  [rendered]
chunk {1} 1-es5.067123cd46b49e92639e.js () 58.6 kB  [rendered]
chunk {3} main-es2015.9aaa072f8c4bafbb0741.js (main) 1.23 MB [initial] [rendered]
chunk {3} main-es5.9aaa072f8c4bafbb0741.js (main) 1.38 MB [initial] [rendered]
chunk {11} 11-es2015.0b8251f22b4aeeb5189d.js () 6.81 kB  [rendered]
chunk {11} 11-es5.0b8251f22b4aeeb5189d.js () 7.31 kB  [rendered]
chunk {12} 12-es2015.72933d71e5152be6b59d.js () 33.5 kB  [rendered]
chunk {12} 12-es5.72933d71e5152be6b59d.js () 34.6 kB  [rendered]
chunk {13} 13-es2015.1fc927ee306364cc00ce.js () 94.2 kB  [rendered]
chunk {13} 13-es5.1fc927ee306364cc00ce.js () 98.7 kB  [rendered]
chunk {14} 14-es2015.bb0b32700fe49a3f642f.js () 6.96 kB  [rendered]
chunk {14} 14-es5.bb0b32700fe49a3f642f.js () 7.55 kB  [rendered]
chunk {15} 15-es2015.810b7eb6e28f4c5a18a9.js () 47.4 kB  [rendered]
chunk {15} 15-es5.810b7eb6e28f4c5a18a9.js () 49.7 kB  [rendered]
chunk {16} 16-es2015.4654fbeb9ef11a461cf2.js () 1.27 kB  [rendered]
chunk {16} 16-es5.4654fbeb9ef11a461cf2.js () 1.27 kB  [rendered]
chunk {17} 17-es2015.e449da56215df3d9f458.js () 2.16 kB  [rendered]
chunk {17} 17-es5.e449da56215df3d9f458.js () 2.16 kB  [rendered]
chunk {18} 18-es2015.b5d8895043bdc2ccc136.js () 3.23 kB  [rendered]
chunk {18} 18-es5.b5d8895043bdc2ccc136.js () 3.23 kB  [rendered]
chunk {19} 19-es2015.e56ec75d19574d2ac0c7.js () 2.79 kB  [rendered]
chunk {19} 19-es5.e56ec75d19574d2ac0c7.js () 2.79 kB  [rendered]
chunk {20} 20-es2015.690e247eaeb6eb972646.js () 1.67 kB  [rendered]
chunk {20} 20-es5.690e247eaeb6eb972646.js () 1.67 kB  [rendered]
... Locale chunks...
chunk {243} 243-es2015.605d9123aa79c34ed23f.js () 1.58 kB  [rendered]
chunk {243} 243-es5.605d9123aa79c34ed23f.js () 1.58 kB  [rendered]
chunk {0} runtime-es2015.3958ea6a909818f107eb.js (runtime) 8.31 kB [entry] [rendered]
chunk {0} runtime-es5.3958ea6a909818f107eb.js (runtime) 8.31 kB [entry] [rendered]
chunk {2} font-awesome.min.css (font-awesome.min) 29.9 kB [initial] [rendered]
chunk {6} primeicons.css (primeicons) 7.47 kB [initial] [rendered]
chunk {7} primeng-nova-light-theme.css (primeng-nova-light-theme) 80.9 kB [initial] [rendered]
chunk {8} primeng.min.css (primeng.min) 76.7 kB [initial] [rendered]
chunk {9} styles.4953f15ce3da10a7ca92.css (styles) 138 kB [initial] [rendered]
Date: 2019-10-09T05:30:37.419Z - Hash: 5a3d03b7c603dc596e66 - Time: 78306ms

WARNING in budgets, maximum exceeded for initial. Budget 2 MB was exceeded by 462 kB.
```
Sum up the es2015 or es5 sizes. The result isn't more than 2 MB.

Also, these chunks are not initial. Because I marked the lazy.
```
chunk {2} font-awesome.min.css (font-awesome.min) 29.9 kB [initial] [rendered]
chunk {6} primeicons.css (primeicons) 7.47 kB [initial] [rendered]
chunk {7} primeng-nova-light-theme.css (primeng-nova-light-theme) 80.9 kB [initial] [rendered]
chunk {8} primeng.min.css (primeng.min) 76.7 kB [initial] [rendered]
```

See the my `angular.json`:
```json
 "build": {
  "options": {
    "outputPath": "dist/dev-app",
    "index": "apps/dev-app/src/index.html",
    "main": "apps/dev-app/src/main.ts",
    "polyfills": "apps/dev-app/src/polyfills.ts",
    "tsConfig": "apps/dev-app/tsconfig.app.json",
    "aot": false,
    "assets": ["apps/dev-app/src/favicon.ico", "apps/dev-app/src/assets"],
    "extractCss": true,
    "styles": [
      "apps/dev-app/src/styles.scss",
      "node_modules/bootstrap/dist/css/bootstrap.min.css",
      {
        "input": "node_modules/font-awesome/css/font-awesome.min.css",
        "lazy": true,
        "bundleName": "font-awesome.min"
      },
      {
        "input": "node_modules/primeng/resources/themes/nova-light/theme.css",
        "lazy": true,
        "bundleName": "primeng-nova-light-theme"
      },
      {
        "input": "node_modules/primeicons/primeicons.css",
        "lazy": true,
        "bundleName": "primeicons"
      },
      {
        "input": "node_modules/primeng/resources/primeng.min.css",
        "lazy": true,
        "bundleName": "primeng.min"
      }
    ],
    "scripts": []
}
```

## 🔬 Minimal Reproduction
`ng new angular-repro-app`

`cd angular-repro-app`

`npm install bootstrap`

And then open the `angular.json`, replace the `styles` array as shown below:
```json
"architect": {
   "build": {
         "options": {
           ...
          "styles": ["src/styles.css", 
              { "input": "node_modules/bootstrap/dist/css/bootstrap.css", "lazy": true }],
     }
```

`ng build --prod`

See the build logs:

```
chunk {0} runtime-es2015.e8a2810b3b08d6a1b6aa.js (runtime) 1.45 kB [entry] [rendered]
chunk {0} runtime-es5.e8a2810b3b08d6a1b6aa.js (runtime) 1.45 kB [entry] [rendered]
chunk {2} main-es2015.13bd2a83958218d67a87.js (main) 245 kB [initial] [rendered]
chunk {2} main-es5.13bd2a83958218d67a87.js (main) 294 kB [initial] [rendered]
chunk {3} polyfills-es2015.d1c6d91d83aaea5467fa.js (polyfills) 36.4 kB [initial] [rendered]
chunk {4} polyfills-es5.f6de9812ead48897ca6c.js (polyfills-es5) 123 kB [initial] [rendered]
chunk {1} bootstrap.css (bootstrap) 138 kB [initial] [rendered] // wrong
chunk {5} styles.3ff695c00d717f2d2a11.css (styles) 0 bytes [initial] [rendered]
```

## 🌍 My Environment
<pre><code>
Angular CLI: 8.3.8
Node: 10.16.0
OS: darwin x64
Angular: 8.2.9
... router

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.803.8
@angular-devkit/build-angular      0.803.8
@angular-devkit/build-ng-packagr   0.803.8
@angular-devkit/build-optimizer    0.803.8
@angular-devkit/build-webpack      0.803.8
@angular-devkit/core               8.3.8
@angular-devkit/schematics         8.3.8
@angular/cdk                       8.2.2
@angular/cli                       8.3.8
@ngtools/webpack                   8.3.8
@schematics/angular                8.3.8
@schematics/update                 0.803.8
ng-packagr                         5.5.1
rxjs                               6.4.0
typescript                         3.5.3
webpack                            4.39.2
</code></pre>

dgp1130:
Fixes #15792.

This turned out to be a little more complicated than expected due to the combination of a few different problems resulting in broken budget checks.

* File sizes were previously computed in a Webpack plugin, which is executed during certain lifecycle hooks. I needed to move this to run *after* the build when all the necessary file size information is known. The first commit refactors the size computations without any significant changes to the output.
* Files listed as `lazy: true` or `inject: false` in the app dev's `angular.json` file were included in the `initial` bundle, even though these resources must be loaded asynchronously. The second commit filters out files which were marked asynchronous from the `initial` bundle.
* Differential builds were included in the same size, so the `main` size was including both the ES2015 **and** ES5 sizes together, which isn't what the user would want. The third commit updates the relevant budgets to output two sizes, one for each differential build, rather than including them together.

I made separate commits performing each of these steps individually to make the logic a little easier to follow and understand.


packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
```diff
@@ -108,11 +266,11 @@ class AllCalculator extends Calculator {
  */
 class AnyComponentStyleCalculator extends Calculator {
   calculate() {
-    return Object.keys(this.compilation.assets)
-      .filter(key => key.endsWith('.css'))
-      .map(key => ({
-        size: this.compilation.assets[key].size(),
-        label: key,
+    return this.assets
```
alan-agius4: This line here doesn’t seem right, the css assets here would be global css file and not the component styles.

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -592,34 +595,55 @@ export function buildWebpackBrowser(
...
+              for (const budget of budgets) {
+                const sizes = calculateSizes(budget, webpackStats);
+                for (const threshold of calculateThresholds(budget)) {
+                  for (const {size, label} of sizes) {
+                    if (exceededThreshold(size, threshold)) {
+                      const colorize = colorizeSeverity(threshold.severity);
+                      const severity = stringifySeverity(threshold.severity);
+
+                      switch (threshold.type) {
+                        case ThresholdType.MAX: {
```
clydin: Moving the threshold checks into these cases would reduce the block depth by 1 and remove the need for the `exceededThreshold` function.
Maybe something like for max:
```
if (size <= threshold.limit) {
  continue;
}
```

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -592,34 +595,55 @@ export function buildWebpackBrowser(
...
-                  bundleInfoText +=
-                    '\n' + generateBundleStats({ ...chunk, size: asset && asset.size }, true);
+              const budgets = options.budgets || [];
+              for (const budget of budgets) {
+                const sizes = calculateSizes(budget, webpackStats);
+                for (const threshold of calculateThresholds(budget)) {
+                  for (const {size, label} of sizes) {
+                    if (exceededThreshold(size, threshold)) {
+                      const colorize = colorizeSeverity(threshold.severity);
+                      const severity = stringifySeverity(threshold.severity);
```
clydin: `ThresholdSeverity` is a string enum so `toUpperCase` can be called directly on `threshold.severity`.

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -72,6 +74,7 @@ import {
 import { Schema as BrowserBuilderSchema } from './schema';
 
 const cacheDownlevelPath = cachingDisabled ? undefined : findCachePath('angular-build-dl');
+const {bold, red, yellow} = terminal;
```
clydin: `terminal` is deprecated.  For now, probably drop the color.  Once structured diagnostic messaging is introduced, this can be handled on the consumer side (the CLI itself).

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
```diff
@@ -5,22 +5,94 @@
...
+  MAX = 'maximum',
+  MIN = 'minimum',
+}
+
+export enum ThresholdSeverity {
+  WARNING = 'warning',
+  ERROR = 'error',
+}
+
+export function* calculateThresholds(budget: Budget): IterableIterator<Threshold> {
```
clydin: Nice use of a generator.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
```diff
@@ -5,22 +5,109 @@
...
-export function calculateSizes(budget: Budget, compilation: Compilation): Size[] {
-  const calculatorMap = {
+export interface Threshold {
+  limit: number;
+  type: ThresholdType;
+  severity: ThresholdSeverity;
+}
+
+export enum ThresholdType {
+  MAX = 'maximum',
```
clydin: nit: Pascal case is typically used for enums.  All uppercase is generally reserved for module level constants.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
dgp1130: Ah, yes I think you're right. Is there a way to get only component styles from Webpack stats? I don't see component-specific styles in those statistics. Is there some way I can query or search all component styles specifically?

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
dgp1130: Thanks! Actually just threw in another one to check the budgets in order to share code a little better with the existing Webpack plugin.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
dgp1130: Done.

packages/angular_devkit/build_angular/src/browser/index.ts
dgp1130: Removed. This is a little weird since the older plugin (when not using differential builds) still uses those colors yet now the new differential build budgets won't.

packages/angular_devkit/build_angular/src/browser/index.ts
dgp1130: This function is more about decoupling than anything else. If the format of `ThresholdSeverity` changes (say to `'W'` and `'E'`), then we would not want this text to change. So this function is just adding a layer of indirection between the view and model.

packages/angular_devkit/build_angular/src/browser/index.ts
dgp1130: Good point. Done.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -553,8 +552,6 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
       noEmitOnErrors: true,
       minimizer: [
         new HashedModuleIdsPlugin(),
-        // TODO: check with Mike what this feature needs.
-        new BundleBudgetPlugin({ budgets: buildOptions.budgets }),
```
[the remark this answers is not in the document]
dgp1130: Good point. Revived the plugin and refactored it to call the updated bundle calculator module.

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: You can add the warnings/errors into the corresponding arrays within the webpack stats.  That would guarantee they would be displayed identically.

dgp1130: Good point. Updated to use `compilation.warnings` and `compilation.errors`.

packages/angular_devkit/build_angular/src/browser/index.ts
dgp1130: `stringifySeverity()` got deleted anyways since we're now using `compilation.warnings` and `compilation.errors` to render them to the user, so nothing to do here.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
clydin: Did this work previously?

This appears to be fairly complicated to do correctly since the process will differ greatly between JIT and AOT modes.  For JIT, this would need to hook into the ngtools JIT decorator transformer to determine the styles for each component and then calculate the sizes of any file based stylesheets.  For AOT, the resource loader runs a child webpack compilation which only outputs the resource (in this case the stylesheet) for the component.  As long as this plugin is provided to the child webpack compilation, this should work as is.  
However, in either case, this calculator should not be present in the main webpack compilation as none of the assets will be component stylesheets.

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -592,32 +602,20 @@ export function buildWebpackBrowser(
...
-                  bundleInfoText +=
-                    '\n' + generateBundleInfoStats(result.name, result.downlevel, chunk);
-                }
-              }
-
-              if (webpackStats && webpackStats.chunks) {
-                for (const chunk of webpackStats.chunks) {
-                  if (processedNames.has(chunk.id.toString())) {
-                    continue;
+              for (const chunk of webpackStats.chunks) {
```
clydin: Can you keep this part as-is?  There are potential processed files that are not Webpack chunks. Web worker files are one such case.

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -631,6 +629,24 @@ export function buildWebpackBrowser(
...
+              // Check for budget errors and display them to the user.
+              const budgets = options.budgets || [];
+              const budgetFailures = checkBudgets(budgets, webpackStats, processResults);
+              for (const {severity, message} of budgetFailures) {
+                switch (severity) {
+                  case ThresholdSeverity.Warning:
+                    webpackStats.warnings.push(message);
+                    break;
+                  case ThresholdSeverity.Error:
+                    webpackStats.errors.push(message);
```
clydin: To ensure the compilation is considered a failure when an error occurs here, the following will need to be added after line 654 (in the error logger if block): `return { success: false };`

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
clydin: If this aspect is working as it did before this PR, I think we should defer any behavior changes here to a separate PR.

packages/angular_devkit/build_angular/src/browser/index.ts
dgp1130: Good catch. Updated to return failure.

packages/angular_devkit/build_angular/src/browser/index.ts
dgp1130: Updated this to run `generateBundleStats()` on all chunks which don't have an associated `processResult`, which I think it was this was trying to do.

I made a simple project with a web worker, but I don't see that show up in `chunks` or `processResults` and it doesn't not display bundle size information even when built with the CLI from master. I'm not sure exactly what this is supposed to do in that case, so I can verify that my changes don't alter than behavior.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
dgp1130: I just tested this and component styles budget checks are no longer working. For differential builds, the plugin is disabled, so we must compute this after the build. Unfortunately we no longer have access the `compilation` property, so I don't think we can easily compute this in the old way.

Regarding JIT vs AOT, these metrics are only output if the user runs with `--prod` I believe. Is JIT still used in that context? Would we be able to get away with just getting AOT to work?

What would be the best way to get access to the underlying CSS file size data? Would we need to make another plugin that just handles CSS styles, or run the existing plugin when building differential builds, but only compute style data?

packages/angular_devkit/build_angular/src/browser/index.ts
clydin: They won't show until #16290 is merged.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
clydin: JIT in production is rare and it doesn't appear that it worked previously either.
For AOT, the plugin could be added to the child compilation directly ([code](https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/resource_loader.ts#L64)).  The options would need to be available at that location though.  That will probably need some thought as ideally the budget options should probably not be passed through the ngtools webpack plugin.

dgp1130: After some experimentation, it seems that component style budgets need to be checked in a Webpack plugin so their size information can be retrieved while we are still able to emit usable errors.

There is an existing plugin, but that is only used to check budgets for non-differential builds. I pulled out the `AnyComponentStyle` functionality to a separate plugin which is used for all builds.

PR #16280, opened 2019-11-25
fix(@angular-devkit/build-angular): add pattern validations for `brow…

alan-agius4:
…serTarget` and `serverTarget`

### DIFF
--- a/packages/angular_devkit/build_angular/src/app-shell/schema.json
+++ b/packages/angular_devkit/build_angular/src/app-shell/schema.json
@@ -6,11 +6,13 @@
   "properties": {
     "browserTarget": {
       "type": "string",
-      "description": "Target to build."
+      "description": "Target to build.",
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/app-shell/schema.json
```diff
@@ -6,11 +6,13 @@
...
       "type": "string",
-      "description": "Target to build."
+      "description": "Target to build.",
+      "pattern": ".+:.+(:.+)?"
     },
     "serverTarget": {
       "type": "string",
-      "description": "Server target to use for rendering the app shell."
+      "description": "Server target to use for rendering the app shell.",
+      "pattern": ".+:.+(:.+)?"
```
clydin: What about something like `"^[^:\s]+:[^:\s]+(:[^\s]+)?$"`?

alan-agius4: Seems more correct indeed.

PR #16290, opened 2019-11-26
fix(@angular-devkit/build-angular): support differential loading for web workers

### LINKED ISSUE #16275: Adding web worker via "ng generate webWorker" does not generate differential builds
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Unsure, I've only just encountered a scenario that would reveal this bug.

### Description
Differential builds are not generated for web workers, only es2015.

2019-11-24 11_58_26-NgWorkerEs5Test

## 🔬 Minimal Reproduction

- Step 1
`ng new NgWorkerDiffTest --routing=false --style=scss & cd NgWorkerDiffTest & ng generate webWorker MyWorker`
- Step 2
Modify `app.component.ts` to include:
<pre>
<code>
    if (typeof Worker !== "undefined") {
      const worker = new Worker("./my-worker.worker", { type: "module" });
    }
</code>
</pre>
- Step 3
`ng build` or `ng build --prod` - neither will produce anything but es2015 for the worker.

## 🌍 Your Environment
<pre><code>

Angular CLI: 9.0.0-rc.3
Node: 12.12.0
OS: win32 x64
Angular:
...

@angular-devkit/architect    0.900.0-rc.3
@angular-devkit/core         9.0.0-rc.3
@angular-devkit/schematics   9.0.0-rc.3
@schematics/angular          9.0.0-rc.3
@schematics/update           0.900.0-rc.3
rxjs                         6.5.3
</code></pre>

**Anything else relevant?**

In a way - IE11 will fail to load workers due to syntax errors.

clydin:
Fixes #16275

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/index.ts
+++ b/packages/angular_devkit/build_angular/src/browser/index.ts
@@ -331,11 +331,23 @@ export function buildWebpackBrowser(
 
               let mainChunkId;
               const actions: ProcessBundleOptions[] = [];
+              let workerReplacements: [string, string][] | undefined;
               const seen = new Set<string>();
               for (const file of emittedFiles) {
                 // Assets are not processed nor injected into the index
                 if (file.asset) {
-                  continue;
+                  // WorkerPlugin adds worker files to assets
+                  if (file.file.endsWith('.worker.js')) {
+                    if (!workerReplacements) {
+                      workerReplacements = [];
+                    }
+                    workerReplacements.push([
+                      file.file,
+                      file.file.replace(/\-es20\d{2}/, '-es5'),
+                    ]);
+                  } else {
+                    continue;
+                  }
                 }
 
                 // Scripts and non-javascript files are not processed
@@ -431,7 +443,7 @@ export function buildWebpackBrowser(
                 if (action.integrityAlgorithm && action.runtime) {
                   processRuntimeAction = action;
                 } else {
-                  processActions.push(action);
+                  processActions.push({ replacements: workerReplacements, ...action });
                 }
               }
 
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -334,7 +334,8 @@ export function buildWebpackBrowser(
               const seen = new Set<string>();
               for (const file of emittedFiles) {
                 // Assets are not processed nor injected into the index
-                if (file.asset) {
+                // WorkerPlugin adds worker files to assets
+                if (file.asset && !file.file.endsWith('.worker.js')) {
```
alan-agius4: I am not following how this will fix this issue.

How is the reference the file change? When doing `new WebWorker`

tests/legacy-cli/e2e/tests/build/worker.ts
```diff
@@ -23,6 +23,14 @@ export default async function () {
...
 
+  await ng('build');
+  await expectFileToExist('dist/test-project/0-es5.worker.js');
+  await expectFileToExist('dist/test-project/0-es2015.worker.js');
+
+  await ng('build', '--prod', '--output-hashing=none');
+  await expectFileToExist('dist/test-project/0-es5.worker.js');
+  await expectFileToExist('dist/test-project/0-es2015.worker.js');
+
   // console.warn has to be used because chrome only captures warnings and errors by default
```
alan-agius4: Should add a test that the reference to the web worker file is correct.

Example in the es2015 the es2015 worker is referenced and so on for the es5 version

packages/angular_devkit/build_angular/src/browser/index.ts
alan-agius4: Or that’s part of the runtime js substitutions which we already handle?

clydin: That makes it downlevel the file. Webpack doesn’t know that it exists outside of it being an asset.

clydin: Implemented the second half of this that will update the file references.

PR #16297, opened 2019-11-27
ci: run ngcc on flakey tests too

filipesilva:


scripts/test.ts
```diff
@@ -92,9 +92,10 @@ export default function(args: ParsedArgs, logger: logging.Logger) {
   if (args['ve']) {
     // tslint:disable-next-line:no-console
     console.warn('********* VE Enabled ***********');
-  } else if (args.shard !== undefined) {
+  } else {
     // CI is really flaky with NGCC
     // This is a working around test order and isolation issues.
+    console.warn('********* Running ngcc ***********');
     execSync('./node_modules/.bin/ngcc', { stdio: 'inherit' });
```
alan-agius4: We should probably change this to `./node_modules/.bin/ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points`


So that when running the tests locally you can easily switch between VE and Ivy runs as otherwise you'd need to remove all `node_modules`.

filipesilva: Yes, that would be better.

packages/angular_devkit/build_angular/test/browser/rebuild_spec_large.ts
```diff
@@ -315,13 +315,6 @@ describe('Browser Builder rebuilds', () => {
   });
 
   it('rebuilds after errors in AOT', async () => {
-    // DISABLED_FOR_IVY - These should pass but are currently not supported
-    if (!veEnabled) {
-      pending('Broken in Ivy: https://github.com/angular/angular/issues/32214');
```
filipesilva: This is fixed but not released yet: https://github.com/angular/angular/issues/32214

packages/angular_devkit/build_angular/test/browser/rebuild_spec_large.ts
```diff
@@ -314,14 +315,7 @@ describe('Browser Builder rebuilds', () => {
...
 
-  it('rebuilds after errors in AOT', async () => {
-    // DISABLED_FOR_IVY - These should pass but are currently not supported
-    if (!veEnabled) {
-      pending('Broken in Ivy: https://github.com/angular/angular/issues/32214');
-
-      return;
-    }
-
+  fit('rebuilds after errors in AOT', async () => {
```
alan-agius4: `fit` should be changed back to `it`

packages/angular_devkit/build_angular/test/browser/rebuild_spec_large.ts
```diff
@@ -330,7 +324,7 @@ describe('Browser Builder rebuilds', () => {
     host.replaceInFile('./src/app/app.component.ts', `'app-root'`, `(() => 'app-root')()`);
 
     const overrides = { watch: true, aot: true };
-    const logger = new TestLogger('rebuild-aot-errors');
+    const logger = new TestLogger('rebuild-aot-errors', createConsoleLogger());
```
alan-agius4: Can you explain why the `createConsoleLogger` logger is needed here?

filipesilva: It's what I use to see what's happening in a failing test that I don't expect to fail :D

Issue #12244, opened 2018-09-11
Components should create a `display:block` style by default


hansl:
_From @SanderElias on September 7, 2018 11:19_

## I'm submitting a...

<pre><code>
[ X ] Feature request
[ X ] Documentation issue or request
</code></pre>

## Current behaviour
When you create a new component, by default it gets no setting for css display.

## Expected behavior
it should have `display:block` by default. 

## What is the motivation/use case for changing the behavior?
I know this has been discussed before (#5960), and pops up in (#12340) too.
Still, I'm putitng this back on the roll.  For the following reasons:

1.  Current behaviour is confusing for new devs
2.  Devs need `display:block` most of the time (for me, its 99+%)
3. The web doesn't break if we put this in place. 
4. The expected impact of this "breaking" change is minimal
5. Saves a few bits from the app's payload
6. Less questions. no more, _why does my css..._.
7. One more thing less to remember when creating a component.

I want to elaborate on point 3 a bit. [The same discussion has been held in by standarts committee](https://github.com/w3c/webcomponents/issues/224), If you follow along that discussion you will see that the general consensus is that it is actually a good idea, but it can't be changed because that would break the web. Angular does not have that issue. Angular can change this, and the impact of the change would be minimal.  probably something that has less impact as for example rxjs 5 to 6. 
 
Others:
If we don't do this change, at least there should be more emphasis on this in the documentation.

_Copied from original issue: angular/angular#25856_

hansl:
_From @Meligy on September 8, 2018 1:3_

This issue root is in how Angular keeps the component custom elements with their custom tag names in the rendered DOM, unlike other frameworks that omit these tags from output like React.

The design decision might not be bad overall, and is probably here to stay, but a problematic effect of this design is that these component elements get the `display: inline;` treatment by default. That's what browsers do when they see tags with no style, including built in user agent styles.

It's an instance where lack of action (no `display` setting) is equal to an action (setting `display` to `inline`).

Since it's not even `inline-block`, this often causes weird results when you try to use `block` styles to child elements in the template+styles of the component. The margins etc don't work as expected, and the Angular user has to track it all the way to understanding the above behaviour in order to understand and fix the problem.

This is where a documentation addition might help, also maybe using the Angular CLI default component schematic to add `display: block;` and/or a comment describing the behaviour, to default component styles.

hansl:
_From @SanderElias on September 8, 2018 4:18_

[...]

PR #16299, opened 2019-11-27
feat(@schematics/angular): add opt-in option for 'display: block;'

rryter:
This is a proposal regarding the discussion about #12244,  components should create a `display:block` style by default (they really should). However the decision was made to **not change the default behaviour**.  So I was thinking maybe an opt-in kind of way to handle this is a feasible way to go:

## Setting the option using the CLI (each time)

```ng generate component dashboard --displayBlock=true```

## Setting a default value:
```
...
"projectType": "application",
"schematics": {
      "displayBlock": true
   }
 }
...
```

If you think this could be a way forward, I'll look into the following things:
- [x] Proper css formatting for each available option 
- [x] Check if the changes in "unrelated" spec files are really necessary.
- [x] Ensure quality with additional tests (especially e2e)

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -86,6 +86,12 @@
               "default": "Default",
               "alias": "c"
             },
+            "displayBlock": {
+              "description": "Specifies if the style will contain `:host { display: block; }`.",
+              "type": "boolean",
+              "default": false,
+              "alias": "b"
+            },
             "entryComponent": {
               "type": "boolean",
               "default": false,
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.__style__.template
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.__style__.template
@@ -1,0 +1,6 @@
+<% if(displayBlock){ if(style != 'sass') { %>:host {
+  display: block;
+}
+<% } else { %>\:host
+  display: block;
+<% }} %>
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
@@ -8,7 +8,13 @@ import { Component, OnInit<% if(!!viewEncapsulation) { %>, ViewEncapsulation<% }
     </p>
   `,<% } else { %>
   templateUrl: './<%= dasherize(name) %>.<%= dasherize(type) %>.html',<% } if(inlineStyle) { %>
-  styles: []<% } else { %>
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by vikerman** [comment]
We want to avoid too many options to the schematic. We need to discuss this(not here - but probably as part of the style guide) on whether this makes sense and whether this should be the default.

**Comment by dgp1130** [comment]
Apologies for the delay @rryter.

After some discussion, I think everyone is on board with the intent of this change. It's just a matter of deciding which of the many options discussed in [the issue](https://github.com/angular/angular-cli/issues/12244) is the best one going forward.

Personally, I'm in the camp of just adding:

```css
:host {
  display: block;
}
```

As it is the simplest and most straightforward option. No need for default stylesheets or CSS variables or global styles. Developers can clearly understand what this is doing and it will prompt some of them to look up `:host` as a side benefit. I'm on the fence about whether or not to include:

```css
:host([hidden]) {
  display: none;
}
```

I think it's good to remove a developer foot gun, but I also agree that we shouldn't encourage using the `hidden` attribute, when we have other, better options (such as `*ngIf`). Perhaps we could solve that separately by including a compiler warning if the developer includes `hidden` on an Angular component? That could point them to use `*ngIf` as an alternative. I'm not able to find any official documentation about our stance on `hidden`, but this might be a better option of subtly discouraging `hidden`.

rryter:
@dgp1130 thanks a lot for your guidance. I will do the necessary work as soon as I can and update the PR accordingly.

rryter:
I have handled those two cases in the template file directly. From what I understood when `inlineStyle` is `true` and the css is inlined directly in the `*.component.ts` file, then the syntax is always the same, plain `css`.

Here the two examples:

**sass**
```
\:host 
  display: block;
```

**css, less, scss, styl**
```
:host {
  display: block;
}
```
Regarding the `hidden` attribute I think the way with the compiler warning sounds reasonable. 
This way we do not 'advertise' it to everybody, should its use really be discouraged. I'm a bit lost on how to implement that tough.

dgp1130:
Charles, can you take a look at this when you get the chance? I don't know enough about CLI to be totally comfortable reviewing it. Semantically, I'm fine with this change.

If it's easy to add a warning about using `hidden` on an Angular component, then we should do that too, but I'm guessing that would require a compiler change and could probably be a separate effort.

rryter:
@dgp1130 Thanks a lot for your efforts on this matter.
@clydin Thanks a lot for the review, I think I removed the unnecessary changes.

rryter:
> Just two additional changes:
> 
> 1. The commit scope should be changed from `@angular/cli` to `@schematics/angular`.
> 2. The `packages/angular/cli/lib/config/schema.json` file should also contain the option in the component definition section.  This file is used by IDEs to validate and provide code completion.  There are plans to remove the need to put it in both places but for now it needs to be duplicated.

Thanks a lot for elaborating. I applied the requested changes. 👍

rryter:
Is there anything I can do to keep the ball rolling? @clydin @dgp1130

dgp1130:
Thanks for the contribution @rryter!

PR #16332, opened 2019-12-02
fix(@angular-devkit/build-angular): augment base HREF when localizing

### LINKED ISSUE #16301: Set baseHref for each locale
# 🚀 feature request

### Relevant Package

 This feature request is for @angular-devkit

### Description

With Angular 9-rc.3 the i18n/l10n workflow is greatly improved with:

```
      "i18n": {
        "locales": {
          "fr-FR": "locales/messages.fr-FR.xlf",
          . . .
        }
      },
```

Unfortunately, if we want each locale to have a localized `baseHref` -- this is not supported in the improved flow. 

Where in the previous "build each language" workflow we would have:
```
           "fr-FR": {
              "i18nLocale": "fr-FR",
              "i18nFile": "locales/messages.fr-FR.xlf",
              "outputPath": "dist/browser/fr-FR",
              "baseHref": "/fr-FR/"
            },
```

Ideally the baseHref could/should be treated as part of the localization task. The i18n object would need to change to support this:

```
      "i18n": {
        "locales": {
          "fr-FR": {
              "i18nFile": "locales/messages.fr-FR.xlf",
              "baseHref": "/fr-FR/"
          },
        . . .
        }
      },
```

More prescriptive alternative:

```
      "i18n": {
        "localizedBaseHref": true,
        "locales": {
          "fr-FR": "locales/messages.fr-FR.xlf",
          . . .
        }
      },
```

Where if localizedBaseHref is enabled, the baseHref is prefixed with a path segment for each language, ex: `/fr-FR/`. So if the application baseHref is set to `/foo-bar/`, it would become `/fr-FR/foo-bar/` for the French locale if `localizedBaseHref` is enabled.    

The most likely workaround would be to dynamically change the baseHref in the SSR express server. This approach has other obvious advantages.

clydin:
All locale i18n options now support an object form which allows a base HREF to be defined for the locale.  This allows each locale to optionally define a custom base HREF that will be combined with the base HREF defined for the build configuration.

By default if the shorthand form for the locale is used or the field is not present in the longhand form, the locale code will be used as the base HREF.  To disable automatic augmentation, a base HREF value of an empty string (`""`) can be used.  This will prevent anything from being added to the existing base HREF.

For common scenarios, the shorthand form will result in the preferred and recommended outcome of each built locale variant of the application having a defined base HREF containing the locale code.

Closes #16301

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -376,17 +376,57 @@
           "type": "object",
           "properties": {
             "sourceLocale": {
-              "type": "string",
-              "description": "Specifies the source language of the application.",
-              "default": "en-US"
+              "oneOf": [
+                {
+                  "type": "string",
+                  "description": "Specifies the source locale of the application.",
+                  "default": "en-US",
+                  "pattern": "^[a-z]{2}(-[a-zA-Z]{2,})?$"
+                },
+                {
+                  "type": "object",
+                  "description": "Localization options to use for the source locale",
+                  "properties": {
+                    "code": {
+                      "type": "string",
+                      "description": "Specifies the locale code of the source locale",
+                      "pattern": "^[a-z]{2}(-[a-zA-Z]{2,})?$"
+                    },
+                    "baseHref": {
+                      "type": "string",
+                      "description": "HTML base HREF to use for the locale (defaults to the locale code)"
+                    }
+                  },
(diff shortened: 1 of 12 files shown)


petebacondarwin:
I wonder if the default base HREF could be in some way computed from the `outputPath`?

dgp1130:
Merged all 3 commits to v9.0.x patch branch.

PR #16337, opened 2019-12-02
fix(@angular-devkit/build-angular): pass webpack factor to runWebpack

alan-agius4:
Patch version of https://github.com/angular/angular-cli/pull/16317


alan-agius4:
@mgechev  `build-bazel` is unrelated as the 9.0.x branch is currently red.

Issue #15952, opened 2019-10-28
Each build of an Angular library is missing random files in the output folder


JGH153:
# 🐞 Bug report

### Command (mark with an `x`)
```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?
First time creating library myself, but I have never seen this bug in other libraries.

### Description
Each run of my library (ng build hal-components) (my library) is missing random files in the dist/hal-components/lib folder. A few runs gives me all files, but most (ca 90%) is missing several files. The files missing changes on each run.

Restarting the computer temporarily solves the problem, but it returns after a few hours to one days of use without a restart

Still get the bug if I stop the ng serve command on the consuming project. I get the bug both with, and without the watch flag.

This bug is really annoying as I need to manually check if all files are present before publishing to NPM.

## 🔬 Minimal Reproduction
Library is still tiny so I can link to it: 
https://github.com/hafslundnett/hal-components/tree/build-bug
Only getting the bug on Windows (also tested on Mac, but did not get the bug).

Bug is periodic, but what I do is:
1. `yarn`
2. `ng build hal-components`
3. Check the `dist/hal-components/lib` for missing components from the library

## 🔥 Exception or Error
There is no error, just random missing files.

## 🌍 Your Environment
<pre><code>
Angular CLI: 8.3.12
Node: 10.16.0
OS: win32 x64
Angular: 8.1.3
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.801.3
@angular-devkit/build-angular      0.801.3
@angular-devkit/build-ng-packagr   0.801.3
@angular-devkit/build-optimizer    0.801.3
@angular-devkit/build-webpack      0.801.3
@angular-devkit/core               8.3.12
@angular-devkit/schematics         8.3.12
@angular/cdk                       8.2.3
@angular/cli                       8.3.12
@angular/material                  8.2.3
@ngtools/webpack                   8.1.3
@schematics/angular                8.3.12
@schematics/update                 0.803.12
ng-packagr                         5.7.0
rxjs                               6.4.0
typescript                         3.4.5
webpack                            4.35.2
</code></pre>

[...]

PR #16340, opened 2019-12-03
Library imports

alan-agius4:
- Currently the library schematic doesn't support adding a secondary entry-point and having deep imports is not recommanded.

It's best if paths are more stricter when having a secondary entry-point instead of a wildcard.

Instead of :
```
"lib/*": [
  "dist/lib/*"
]
```

Users should configure:
```
"lib/secondary": [
  "dist/lib/secondary"
]
```

This would allow a better DX experience when using auto imports in IDE's.

Closes: #15952


alan-agius4:
@clydin let's do that

PR #16343, opened 2019-12-04
Library imports patch

alan-agius4:
Patch version of https://github.com/angular/angular-cli/pull/16340/


alan-agius4:
CI failure is unrelated.

PR #16344, opened 2019-12-04
fix(@angular-devkit/build-angular): add notice when using bundleDepen…

alan-agius4:
…dencies: none with Ivy

Related to: https://github.com/angular/universal/issues/1354

### DIFF
--- a/packages/angular_devkit/build_angular/src/server/index.ts
+++ b/packages/angular_devkit/build_angular/src/server/index.ts
@@ -7,7 +7,7 @@
  */
 import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
 import { runWebpack } from '@angular-devkit/build-webpack';
-import { json } from '@angular-devkit/core';
+import { json, tags } from '@angular-devkit/core';
 import * as path from 'path';
 import { Observable, from } from 'rxjs';
 import { concatMap, map } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


alan-agius4:
Blocked on https://github.com/angular/angular/pull/34225

packages/angular_devkit/build_angular/src/server/index.ts
```diff
@@ -59,6 +59,14 @@ export function execute(
   const baseOutputPath = path.resolve(root, options.outputPath);
   let outputPaths: undefined | Map<string, string>;
 
+  if (options.bundleDependencies === 'none' && tsConfig.options.enableIvy) {
+    context.logger.warn(tags.stripIndent`
```
filipesilva: I think this falls within the category of "warnings you can't turn off are bad and people ignore them". But I also don't think it warrants another option in the config.

Can we maybe actually check instead? We know when we're using Ivy. And we can know if Ivy processed a package by checking something known, like `@angular/core`. Maybe the right thing to do is check if these match? But then we break the local lib development journey because it compiles directly to Ivy.

This does feel like an extremely rare case so maybe just always warning is ok.

alan-agius4: I think we can do it, as it's trivial.

filipesilva: I hadn't really considered we could just do it here in that way. Yeah that LGTM, now the only way to see the warning is if you use a precompiled angular/core for server builds. Which is very unlikely.

PR #16346, opened 2019-12-04
Fix patch ci

alan-agius4:
Some commits to the patch branch have been missed by the caretaker.


filipesilva:
@googlebot I consent.

dgp1130:
Merged all commits to 9.0.x patch branch, thanks for catching these misses.

Issue #16193, opened 2019-11-14
WARNING in Conflict: Multiple assets emit different content to the same filename 3rdpartylicenses.txt


filipesilva:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: .8.x

### Description

The following warning appears on production builds:
```
kamik@RED-X1C6 MINGW64 /d/sandbox/super-productivity (update-9)
$ yarn buildFrontend --output-path dist-ivy
yarn run v1.17.3
$ node --max_old_space_size=4096 ./node_modules/@angular/cli/bin/ng build --aot --prod --output-path dist-ivy
69% building 889/891 modules 2 active ....js??ref--18-0!D:\sandbox\super-productivity\node_modules\@ngtools\webpack\src\index.js!D:\sandbox\super-productivity\src\app\core\compression\lz.worker.ts 69% building 1154/1157 modules 3 active ...-18-0!D:\sandbox\super-productivity\node_modules\@ngtools\webpack\src\index.js!D:\sandbox\super-productivity\src\app\features\reminder\reminder.worker.ts 
chunk {0} runtime.0f9d2f90a0d2148a4e14.js, runtime.0f9d2f90a0d2148a4e14.js.map (runtime) 1.51 kB [entry] [rendered]
chunk {1} main.9009a33894da809485f9.js, main.9009a33894da809485f9.js.map (main) 3.33 MB [initial] [rendered]
chunk {2} polyfills.75f8db4dd1e518a7835b.js, polyfills.75f8db4dd1e518a7835b.js.map (polyfills) 121 kB [initial] [rendered]
chunk {3} polyfills-es5.aca657e0425e42bebda6.js, polyfills-es5.aca657e0425e42bebda6.js.map (polyfills-es5) 201 kB [initial] [rendered]
chunk {4} styles.c5a09efce74a2a2c2e94.css, styles.c5a09efce74a2a2c2e94.css.map (styles) 167 kB [initial] [rendered]
Date: 2019-11-14T13:28:02.212Z - Hash: c21370451102fb263dbb - Time: 154136ms

[...]

PR #16349, opened 2019-12-04
fix(@angular-devkit/build-angular): suppress duplicate 3rdpartylicenses.txt warning

dgp1130:
Refs #16193

This detects and filters out error messages about duplicate `3rdpartylicenses.txt`. This is a short-term fix intended to get us to 9.0.x release while a follow up effort will work to properly fix this bug. Left a `TODO` to remove this filter once a real fix is ready.


dgp1130:
Just merged to LTS (8.3.x) per https://github.com/angular/angular-cli/issues/16193#issuecomment-573947915.

PR #16352, opened 2019-12-04
fix(@ngtools/webpack): format Ivy diagnostics properly

vikerman:
By passing all diagnostics through formatDiagnostics from compiler-cli.

Without this we were printing negative TS error codes which was
compiler-cli way of marking Angular errors.

### DIFF
--- a/packages/ngtools/webpack/src/diagnostics.ts
+++ b/packages/ngtools/webpack/src/diagnostics.ts
@@ -141,12 +141,12 @@ export function reportDiagnostics(
   }
 
   if (tsErrors.length > 0) {
-    const message = ts.formatDiagnosticsWithColorAndContext(tsErrors, compilerHost);
+    const message = formatDiagnostics(tsErrors);
     reportError(message);
   }
 
   if (tsWarnings.length > 0) {
-    const message = ts.formatDiagnosticsWithColorAndContext(tsWarnings, compilerHost);
+    const message = formatDiagnostics(tsWarnings);
     reportWarning(message);
   }
 


vikerman:
cc @alxhub

vikerman:
> These should only be actual TS errors and warnings. They should be filtered by the code above.

My understanding is that in Ivy mode, the Angular compiler produces TS errors/warning with special negative numbers that need to be processed by compiler-cli formatDiagnostics. So this passes everything to the compiler-cli so that it can do that.

clydin:
The emitted diagnostics are bifurcated into TS origin diagnostics (`tsErrors`/`tsWarnings`) and Angular origin diagnostics (`ngErrors`/`ngWarnings`) via `isNgDiagnostic` from `@angular/compiler-cli`.  The Angular origin diagnostics are formatted using the Angular function and the TS origin diagnostics use the TS function.  The later provides a more comprehensive output that provides an equivalent user experience to running `tsc` directly.

If the Angular variant has been improved to provide a similar experience to `ts.formatDiagnosticsWithColorAndContext` then most of the enclosing function can be dropped.

Just tried with the latest RC and received the following for a TS origin error:
image
And the following for an Angular origin error:
image

vikerman:
Checked with AlexR. Apparently NgDiagnostics has completely different shape than TS diagnostics. And Ivy produces only TS diagnostics
So we shouldn't change isNgDiagnostics. Because the shape of the diagnostics is intentionally TS diagnostics

vikerman:
If you make an error in the decorator parameter you would currently get a TS error with a negative error number - Which this current PR fixes.

AlexR is separately fixing the compiler-cli to use `ts.formatDiagnosticsWithColorAndContext` instead of `ts.formatDiagnostics`

clydin:
I think adding an additional check for the origin/source during the filtering could be a good solution if `isNgDiagnostic` is intended to check for the shape of the object rather than the source.  The intention of the filtering was actually to separate them by their source.  It was just coincidental that the shapes were previously different.
```
if (isNgDiagnostic(diagnostic) || diagnostic.source === 'ngtsc') {
```

clydin:
Actually, it looks like the compiler cli already has the necessary helper function to fix this.  It just needs to be exported.
https://github.com/angular/angular/blob/f9e959e098f1c872070edb352241653d445ba8f0/packages/compiler-cli/src/ngtsc/diagnostics/src/util.ts#L13

Looks like it was written for the exact use case we have here.
https://github.com/angular/angular/blob/f9e959e098f1c872070edb352241653d445ba8f0/packages/compiler-cli/src/main.ts#L249

clydin:
Looks like https://github.com/angular/angular/pull/34234 is close to completion so I agree that the best path forward would be to move to its use directly for all cases allowing the logic here to be nicely simplified.

PR #16358, opened 2019-12-05
refactor: deprecated unused server builder options

alan-agius4:


packages/angular_devkit/build_angular/src/server/schema.json
```diff
@@ -212,7 +214,8 @@
     "namedChunks": {
       "type": "boolean",
       "description": "Use file name for lazy loaded chunks.",
-      "default": true
+      "default": true,
+      "x-deprecated": "Since version 9. This option has no effect on server platform."
```
clydin: By no effect, does that mean the chunks will always be named?

alan-agius4: You only get a single chunk with the server builder, even when you have lazy loading.

PR #16370, opened 2019-12-06
Fix bazel integration

alan-agius4:
//cc @alexeagle

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -489,6 +489,7 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
       alias,
     },
     resolveLoader: {
+      symlinks: !buildOptions.preserveSymlinks,
       modules: loaderNodeModules,
     },
     context: projectRoot,
--- a/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
@@ -126,9 +126,15 @@ export async function process(options: ProcessBundleOptions): Promise<ProcessBun
       filename: options.filename,
       inputSourceMap: manualSourceMaps ? undefined : sourceMap,
       babelrc: false,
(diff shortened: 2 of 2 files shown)


packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -128,7 +128,13 @@ export async function process(options: ProcessBundleOptions): Promise<ProcessBun
       babelrc: false,
       // modules aren't needed since the bundles use webpack's custom module loading
       // 'transform-typeof-symbol' generates slower code
-      presets: [['@babel/preset-env', { modules: false, exclude: ['transform-typeof-symbol'] }]],
+      presets: [[
+        require.resolve('@babel/preset-env'),
+        {
+          modules: false,
```
clydin: Can you move the relevant comments next to each option as well?

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -489,6 +489,7 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
       alias,
     },
     resolveLoader: {
+      symlinks: !buildOptions.preserveSymlinks,
```
clydin: Related: We should check to make sure all loaders are resolved  (`require.resolve`) before use to ensure the proper logic is used.  Especially relevant when peer dependencies are involved but also helpful for eventual pnp support.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
alan-agius4: Moved

alexeagle:
I think this PR should no longer be necessary once https://github.com/bazelbuild/rules_nodejs/pull/1430 lands
It will cause us to lay out the normal `node_modules` tree in the working directory and patch the symlink primitives in node, when you call architect via a load from index.bzl

mrmeku:
@alexeagle That might not necessarily be the case until we change the generated `nodejs_test` target to have its entry point live from within the normal `node_modules` directory. As it is the entry point will still be `external/npm` after that PR lands.

PR #16371, opened 2019-12-06
wip! webpack 5

alan-agius4:


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/browser.ts
```diff
@@ -93,15 +93,9 @@ export function getBrowserConfig(wco: WebpackConfigOptions): webpack.Configurati
           vendors: false,
           vendor: !!buildOptions.vendorChunk && {
             name: 'vendor',
-            chunks: 'initial',
```
clydin: Is initial chunks automatic now?  Only node module files from initial chunks should end up in the vendor chunk.  Otherwise the main bundle in production mode will contain all dependencies for the entire application including ones only used in lazy routes.

packages/angular_devkit/build_optimizer/package.json
```diff
@@ -13,6 +13,6 @@
     "source-map": "0.7.3",
     "tslib": "1.10.0",
     "typescript": "3.6.4",
-    "webpack-sources": "1.4.3"
+    "webpack-sources": "2.0.0-beta.8"
```
alan-agius4: Updating this seems to cause a bunch of tsc errors due to tapable

alan-agius4:
Closing since these were POC’s

Issue #16226, opened 2019-11-19
ng add @angular/pwa - Cannot read property 'length' of undefined


GeorgiyZhuravlev:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [x] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

I don't know

### Description

If your `src/app/app.module.ts` imports `environment` with the alias like this:
```
import {environment as env} from '../environments/environment';
```
`ng add @angular/pwa` fails with `Cannot read property 'length' of undefined`

## 🔬 Minimal Reproduction
1. create empty project with `angular-cli`, eg:
`ng new some-app --prefix=some`
2. add this line to `src/app/app.module.ts`
`import {environment as env} from '../environments/environment';`
3. try to add `@angular/pwa` to project
`ng add @angular/pwa --project some-app`

## 🔥 Exception or Error
<pre><code>
ng add @angular/pwa --project some-app

Installing packages for tooling via npm.
+ @angular/pwa@0.803.19
added 5 packages from 2 contributors and audited 18936 packages in 6.047s
found 0 vulnerabilities

Installed packages for tooling via npm.
Cannot read property 'length' of undefined
</code></pre>

## 🌍 Your Environment
<pre><code>
Angular CLI: 8.3.19
Node: 10.16.3
OS: darwin x64
Angular: 8.2.14
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.803.19
@angular-devkit/build-angular     0.803.19
@angular-devkit/build-optimizer   0.803.19
@angular-devkit/build-webpack     0.803.19
@angular-devkit/core              8.3.19
@angular-devkit/schematics        8.3.19
@angular/cli                      8.3.19
@angular/pwa                      0.803.19
@ngtools/webpack                  8.3.19
@schematics/angular               8.3.19
@schematics/update                0.803.19
rxjs                              6.4.0
typescript                        3.5.3
webpack                           4.39.2
</code></pre>

**Anything else relevant?**

PR #16377, opened 2019-12-06
fix(@schematics/angular): handle aliased or existing environment import

arturovt:
Closes: #16226

The exception `Cannot read property 'length' of undefined` wasn't really informative thus I had to find out where this exception was actually thrown. All ways led me to the `Chunk` class and the `insertLeft` method that used `buffer`'s variable `length` property. The `buffer` was `undefined`, that's why that exception was thrown. It was `undefined` because there is no way to check if `environment` is already imported.

This PR adds the ability to handle the already existing "environment" import or the aliased one.

### DIFF
--- a/packages/schematics/angular/utility/ast-utils.ts
+++ b/packages/schematics/angular/utility/ast-utils.ts
@@ -575,6 +575,50 @@ export function isImported(source: ts.SourceFile,
 }
 
 /**
+ * This function returns the name of the environment export
+ * whether this export is aliased or not. If the environment file
+ * is not imported, then it will return `null`.
+ */
+export function getEnvironmentExportName(source: ts.SourceFile): string | null {
+  // Initial value is `null` as we don't know yet if the user
+  // has imported `environment` into the root module or not.
+  let environmentExportName: string | null = null;
+
+  const allNodes = getSourceNodes(source);
+
+  allNodes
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/utility/ast-utils.ts
```diff
@@ -574,6 +574,59 @@ export function isImported(source: ts.SourceFile,
...
+    )
+    .find((declaration: ts.ImportDeclaration) => {
+      if (!declaration.importClause) {
+        return false;
+      }
+
+      // Find specifier that contains `environment` keyword in its text.
+      // Whether it's `environment` or `environment as env`.
+      const nodes = findNodes(declaration, ts.SyntaxKind.ImportSpecifier).filter(node =>
+        node.getText().includes('environment'),
```
clydin: `getText` can be an expensive operation.
Can you move the named bindings analysis up to this point and use it to check the values?

arturovt: I hope I understand you correctly. :slightly_smiling_face:

clydin:
Can you rebase on master?  This should fix the failing tests.

arturovt:
> Can you rebase on master? This should fix the failing tests.

Done. `setup-and-build-win` job seems to be happy:)

PR #16400, opened 2019-12-09
docs: add build system doc

filipesilva:
I think this still needs more detail but is a good starting point.

### DIFF
--- a/docs/design/build-system.md
+++ b/docs/design/build-system.md
@@ -1,0 +1,162 @@
+# Build System
+
+Angular CLI includes a first-party build system for Angular applications distributed as `@angular-devkit/build-angular`.
+This build system is responsible for creating a standalone application from user source files and third-party dependencies.
+
+`@angular-devkit/build-angular` itself integrates with the rest of Angular CLI by being an [Architect builder](https://angular.io/guide/cli-builder).
+This document describes a top level view of the functionality in `@angular-devkit/build-angular`, referred as just the "build system".
+Deprecated or soon to be removed features are not described here.
+
+In broad strokes the main areas are:
+- loading and processing sources
+- code splitting
+- production optimizations
+- post-processing steps
+
+Many tools are used in this process, and most of these steps happen within a [Webpack](https://webpack.js.org/) build.
+We maintain a number of Webpack-centric plugins in this repository, some of these are public but most are private since they are very specific to our setup.
+
+
+## Loading and processing sources
+
+Sources for Angular CLI browser apps are comprised of TypeScript files, style sheets, assets, scripts, and third party dependencies.
+A given build will load these sources from disk, process them, and bundle them together.
+
+
+### TypeScript
+
+Angular builds rely heavily on TypeScript-specific functionality for [Ahead-of-Time template compilation](https://angular.io/guide/aot-compiler) (AOT).
+Outside Angular CLI, this is performed by the Angular Compiler (`ngc`), provided by `@angular/compiler-cli`.
+To avail of Ahead-of-Time template compilation within a Webpack compilation we use and distribute the `@ngtools/webpack` Webpack plugin.
+
+Typescript sources are loaded from disk and compiled in-memory into JavaScript files that are stored in a virtual file system and made available to Webpack.
+During compilation we also perform a number of code transformations using TypeScript transformers that enable automatic usage of AOT, internationalization features, and server-side rendering.
+
+AOT compilation requires loading HTML and CSS resources, referenced on Angular Components, as standalone strings with no external dependencies.
+However, Webpack compilations operate on the basis of modules and references between them.
+To obtain the standalone string we compile resources using a separate Webpack child compilation then extract the results.
+These child compilations inherit configuration and access to the same files as the parent compilation, but have their own compilation life cycle and complete independently.
+
+The build system allows specifying replacements for specific files by replacing what path is loaded from the virtual file system.
+This is used for conditional loading of code at build time.
+
+### Stylesheets
+
+Two types of stylesheets are used in the build system: global stylesheets and component stylesheets.
+Global stylesheets are injected into the `index.html` file, while component stylesheets are loaded directly into compiled Angular components.
+
+The build system supports plain CSS stylesheets as well as the Sass, LESS and Stylus CSS pre-processors.
+Stylesheet processing functionality is provided by `sass-loader`, `less-loader`, `stylus-loader`, `postcss-loader`, `postcss-import`, augmented in the build system by custom webpack plugins.
+
+todo: style resources
+
+
+### Assets
+
+Assets in the build system refer specifically to a list of files or directories that are meant to be copied verbatim as build artifacts.
+These files are not processed and commonly include images, favicons, pdfs and other generic file types.
+They are loaded into the compilation using `copy-webpack-plugin`.
+
+
+### Scripts
+
+Scripts in the build system refer specifically to JavaScript files that are meant to be loaded directly on `index.html` without being processed.
+They are loaded into the compilation using a custom webpack plugin.
+
+
+### Third party dependencies
+
+Third party dependencies are mostly inside `node_modules` and are referenced via imports in source files.
+Stylesheet third party dependencies are treated mostly the same as sources.
+
+JavaScript third party dependencies suffer a more involved process.
+They are first resolved to a folder in `node_modules` via [Node Module Resolution](https://nodejs.org/api/modules.html#modules_modules).
+A given module might have several different entry points, for instance one for use in NodeJS and another one for using in the browser.
+Each entry point is listed in under a name pointing at a js file in that module's `package.json`.
+We use `es2015 > browser > module > main` a priority list, where the first key matched name determines which entry point to use.
+For instance, for a module that has both `browser` and `main` entry points, we pick `browser`.
+
+Once the actual JavaScript file is determined, it is loaded into the compilation together with it's source map.
+
+This resolution strategy supports the [Angular Package Format](https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/edit#heading=h.k0mh3o8u5hx).
+
+
+## Code splitting
+
+Code is automatically split into different files (or chunks, for js files) based on a few different triggers.
+
+The main TypeScript entry point and it's dependencies are bundled into the `main` chunk.
+Global styles and scripts get one file per entry, named after themselves.
+
+JavaScript code imported only via dynamic imports is automatically split into a separate chunk that is loaded asynchronously named after the file containing the dynamic import.
+If multiple asynchronous chunks contain a reference to the same module, it is placed in a new asynchronously loaded chunk named after the other chunks that use it.
+
+There is also a special chunk called `runtime` that contains the module loading logic and is loaded before the others.
+
+
+## Optimizations
+
+The build system contains optimizations aimed at improving the performance (for development builds) or the size of artifacts (for production builds).
+These are often mutually exclusive and thus we cannot just default to always using them.
+
+
+### Development optimizations
+
+Development optimizations focus on reducing rebuild time on watched builds.
+Although faster is always better, our threshold is to keep rebuilds even for large projects below 2 seconds.
+
+Computation needed to bundle code grows with its total size because of the cost of string concatenation and source map operations.
+Third party dependencies that are initially loaded are split into a synchronously loaded chunk called `vendor`.
+Splitting the infrequently changed vendor code from the frequently changed source code thus helps make rebuilds faster.
+
+When processing stylesheets, Webpack stores the intermediate modules as JavaScript code.
+The JavaScript wrapper code makes stylesheets larger and `mini-css-extract-plugin` must be used to obtain the actually stylesheet content into a CSS file.
+In development however, we skip the CSS extraction and leave it as JavaScript code for faster rebuild times.
+
+Watched builds split the processing load of TypeScript compilation between file emission on the main process and type checking on a forked process.
+Large projects can also opt-out of AOT compilation for faster rebuilds.
+
+
+### Production optimizations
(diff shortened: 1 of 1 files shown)


filipesilva:
@vikerman @dgp1130 does this look like an overview you could use while trying to understand the build system?

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+## Loading and processing sources
+
+Sources for Angular CLI browser apps are comprised of TypeScript files, style sheets, assets, scripts, and third party dependencies.
+A given build will load these sources from disk, process them, and bundle them together.
+
+
+### TypeScript
+
+Angular builds rely heavily on TypeScript-specific functionality for Ahead-of-Time template compilation (AOT).
+Outside Angular CLI, this is performed by the Angular Compiler (`ngc`), provided by `@angular/compiler-cli`.
```
dgp1130: What exactly is the difference between `ngc` and `ngtsc`? Are they synonymous or do they refer to different binaries?

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+To avail of Ahead-of-Time template compilation within a Webpack compilation we use and distribute the `@ngtools/webpack` Webpack plugin.
+
+Typescript sources are loaded from disk and compiled in-memory into Javascript files that are stored in a virtual file system and made available to Webpack.
+During compilation we also perform a number of code transformations using TypeScript transformers that enable automatic usage of AOT, internationalization features, and server-side rendering.
+
+AOT compilation requires loading HTML and CSS resources referenced on Angular Components. These are processed using a Webpack child compilation.
```
dgp1130: Is there any definition or explanation of what a "child compilation" is and why it is used for HTML and CSS assets? I encountered this earlier, but I couldn't find any real documentation and didn't fully grasp the purpose behind it.

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+
+The build system allows specifying replacements for specific files by replacing what path is loaded from the virtual file system.
+This is used for conditional loading of code at build time.
+
+### Stylesheets
+
+Two types of stylesheets are used in the build system: global stylesheets and component stylesheets.
+Global stylesheets are injected into the `index.html` file, while component stylesheets are loaded directly into compiled Angular components.
```
dgp1130: Are global stylesheets also injected into other entry points? Or are all Angular apps actual SPA's under the hood?

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+
+
+### Third party dependencies
+
+Third party dependencies are mostly inside `node_modules` and are referenced via imports in source files.
+Stylesheet third party dependencies are treated mostly the same as sources.
+
+JavaScript third party dependencies suffer a more involved process.
+They are first resolved to a folder in `node_modules` via [Node Module Resolution](https://nodejs.org/api/modules.html#modules_modules).
+Afterwards one of the entry points in that folder's `package.json` is used following the `es2015 > browser > module > main` priority list.
```
dgp1130: I don't understand what this `es2015 > browser > module > main` priority list means? Is this the loader trying to choose the best form of the dependency to import?

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+They are loaded into the compilation using `copy-webpack-plugin`.
+
+
+### Scripts
+
+Scripts in the build system refer specifically to Javascript files that are meant to be loaded directly on `index.html` without being processed.
+They are loaded into the compilation using a custom webpack plugin.
+
+
+### Third party dependencies
```
dgp1130: Are there any differences between Angular dependencies and standard NPM dependencies? I understand that Angular has a special format for distribution/consumption, so I'm curious if there is a separate pipeline there.

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+
+JavaScript code imported only via dynamic imports is automatically split into a separate chunk that is loaded asynchronously named after the file containing the dynamic import.
+If multiple asynchronous chunks contain a reference to the same module, it is placed in a new asynchronously loaded chunk named after the other chunks that use it.
+
+There is also a special chunk called `runtime` that contains the module loading logic and is loaded before the others.
+
+
+## Optimizations
+
+The build system contains optimizations aimed and improving the performance, for development builds, or the size of artifacts, for production builds.
```
dgp1130: Nit: The commas here are confusingly worded. I think you meant something like:

"The build system contains optimizations aimed **at** improving the performance **(for development builds)**, or the size of artifacts **(for production builds)**. These are often mutually exclusive and thus we cannot just default to always using them."

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+### Development optimizations
+
+Development optimizations focus on reducing rebuild time on watched builds.
+Although faster is always better, our threshold is to keep rebuilds even for large projects below 2 seconds.
+
+Computation needed to bundle code grows with its total size because of the cost of string concatenation and source map operations.
+Third party dependencies that are initially loaded are split into a synchronously loaded chunk called `vendor`.
+Splitting the infrequently changed vendor code from the frequently changed source code thus helps make rebuilds faster.
+
+When processing stylesheets Webpack stores the intermediate modules as javascript code.
```
dgp1130: Nit: When processing stylesheets, Webpack...

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+The javascript wrapper code makes stylesheets larger and `mini-css-extract-plugin` must be used to obtain the actually stylesheet content into a CSS file.
```
dgp1130: Nit: Should be consistent on the capitalization of "JavaScript" at least throughout this doc.

docs/design/build-system.md
```diff
@@ -0,0 +1,154 @@
...
+We maintain a number of Webpack-centric plugins in this repository, some of these are public but most are private since they are very specific to our setup.
+
+
+## Loading and processing sources
+
+Sources for Angular CLI browser apps are comprised of TypeScript files, style sheets, assets, scripts, and third party dependencies.
+A given build will load these sources from disk, process them, and bundle them together.
+
+
+### TypeScript
```
dgp1130: It might be good to explain what the AOT compilation actually does (and what is deferred when using JIT instead). My understanding is that it just compiles templates. In my also be good to discuss how the pipeline diverges/merges between the AOT and JIT forms.

docs/design/build-system.md
filipesilva: IIRC `ngc` has always been the angular compiler, and `ngtsc` was the name sometimes given to the reworked `ngc` with Ivy. But that named was dropped. I don't think there's such a thing as `ngtsc` at the moment.

There were plans to have `ngc` be just a typescript plugin, but I think that's paused at the moment.

docs/design/build-system.md
filipesilva: Does this change help? I couldn't find webpack docs for them.

> AOT compilation requires loading HTML and CSS resources, referenced on Angular Components, as standalone strings with no external dependencies.
However, Webpack compilations operate on the basis of modules and references between them.
To obtain the standalone string we compile resources using a separate Webpack child compilation then extract the results.
These child compilations inherit configuration and access to the same files as the parent compilation, but have their own compilation life cycle and complete independently.

docs/design/build-system.md
filipesilva: `index.html` is usually the only entry-point per se. I suppose the real answer is that the CLI build system only produces SPAs, yes. The server builder makes a universal app but at that point you are already in control of what's served.

docs/design/build-system.md
filipesilva: Not the loader, but rather webpack itself trying to resolve a given module to an entry point. Do this changes make it clearer?

> JavaScript third party dependencies suffer a more involved process.
They are first resolved to a folder in `node_modules` via [Node Module Resolution](https://nodejs.org/api/modules.html#modules_modules).
A given module might have several different entry points, for instance one for use in NodeJS and another one for using in the browser.
Each entry point is listed in under a name pointing at a js file in that module's `package.json`.
We use `es2015 > browser > module > main` a priority list, where the first key matched name determines which entry point to use.
For instance, for a module that has both `browser` and `main` entry points, we pick `browser`.

docs/design/build-system.md
filipesilva: No real differences. Angular deps are encouraged to follow APF (see below) but APF is geared for maximum compatibility anyway.

Added a mention to this section:

> This resolution strategy supports the [Angular Package Format](https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/edit#heading=h.k0mh3o8u5hx).

docs/design/build-system.md
filipesilva: Fixed

docs/design/build-system.md
filipesilva: Fixed

docs/design/build-system.md
filipesilva: Fixed

docs/design/build-system.md
filipesilva: Added a link to https://angular.io/guide/aot-compiler. I'd prefer to not try and define that here if possible, as the build system mostly consumes.

I touched slightly on JIT on the optimization sections by saying `Large projects can also opt-out of AOT compilation for faster rebuilds.`, but that was all. By default we try to focus on AOT, which is now the default as well. Ideally we'd just do AOT even.

What things do you think belong in a JIT vs AOT section?

docs/design/build-system.md
dgp1130: I see. Probably makes sense to leave that out of the docs then, including it would only lead to more confusion for readers.

docs/design/build-system.md
dgp1130: What exactly is the purpose of a child compilation here? What benefit to we get from using these child compilations as opposed to just doing the same thing in the parent compilation?

docs/design/build-system.md
dgp1130: Thanks for the clarification. It might be worth pointing out somewhere that the CLI only builds SPAs (doesn't necessarily need to be done in the stylesheets section).

docs/design/build-system.md
dgp1130: That helps a bit but I'm still a little confused. Is this a standard in `package.json`? This sounds similar to the `main` property, but I'm not familiar with a standard which allows listing multiple entry points for different environments. Is this an NPM thing or an Angular thing?

> Each entry point is listed in under a name pointing at a js file in that module's package.json.

What "name" is this referring to in the `package.json`? Is it looking for literally `es2015.js`, `browser.js`, etc.? Or is it looking for some `browser` property?

docs/design/build-system.md
dgp1130: Gotcha, thanks for the reference. Good to call that out here.

docs/design/build-system.md
dgp1130: I was mainly thinking about discussing how the build process changes between the two. If the only AOT step is compiling templates, then we could say we just skip that piece. The AOT doc you linked seems to discuss that pretty clearly, so I'm fine with a just including a link to that.

filipesilva: Hmm now that I think about it some more, there are a number of different code transformations we do that differ between JIT and AOT. I don't mention those anywhere and really should.

docs/design/build-system.md
```diff
@@ -0,0 +1,162 @@
...
```
vikerman: I think this section might be better titled as `Ahead-Of-Time Compilation`

docs/design/build-system.md
```diff
@@ -0,0 +1,162 @@
...
+- loading and processing sources
+- code splitting
+- production optimizations
+- post-processing steps
+
+Many tools are used in this process, and most of these steps happen within a [Webpack](https://webpack.js.org/) build.
+We maintain a number of Webpack-centric plugins in this repository, some of these are public but most are private since they are very specific to our setup.
+
+
+## Loading and processing sources
```
vikerman: It might be useful to have a diagram showing the different steps and how they relate to Webpack module loading/plugins etc.

docs/design/build-system.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Production optimizations
+
+Angular CLI focuses on enabling tree-shaking (removing unused modules) and dead code elimination (removing unused module code).
+These two categories have high potential for size reduction because of network effects: removing code can lead to more code being removed.
+
+The main tool we use to achieve this goal are the compression capabilities of [Terser](https://github.com/terser/terser).
+We also use Terser's mangling, by which names, but not properties, are renamed to shorter forms.
+The main characteristics of Terser to keep in mind is that it operates via static analysis and does not support the indirection introduced by module loading.
+Thus the rest of the pipeline is directed towards providing Terser with code that can be removed via static analysis in large single modules scopes.
```
vikerman: This section says we use Tereser's minification abilities. Does the include dead code elimination also? That's not clear here.

docs/design/build-system.md
```diff
@@ -0,0 +1,162 @@
...
+
+To this end we developed [@angular-devkit/build-optimizer](https://github.com/angular/angular-cli/tree/master/packages/angular_devkit/build_optimizer), a post-processing tool for TS code.
```
vikerman: Again a master diagram would be very useful.

docs/design/build-system.md
```diff
@@ -0,0 +1,162 @@
...
+Build Optimizer searches for code patterns produced by the TypeScript and Angular compiler that are known to inhibit dead code elimination, and converts them into equivalent structures that enable it instead.
```
vikerman: An example of such a transformation assuming the reader knows nothing would be very useful.

filipesilva:
Discussed with @vikerman and will address the remaining comments on followup PRs, meanwhile this should be merged.

PR #16413, opened 2019-12-10
doc: adds "PR action: merge" label

dgp1130:
This label should be placed by the author (or last reviewer if author is not a collaborator) when the PR is complete and ready to merged. This requires the author to explicitly acknolwedge that they are done with the PR and the caretaker is free to merge it. This label brings the CLI caretaking process into alignment with the frameworks and components repos.

A more automated solution is desired but difficult due to google3 merging challenges. That would be a separate problem, but this should at least get the Angular organization to a more consistent state and we can revisit that challenge when we get to it.

I chose to link to the existing PR target definitions in the framework docs, let me know if I should duplicate this for the CLI.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -18,7 +18,7 @@ TBD
 ## Merging PRs
 
 The list of PRs which are currently ready to merge (approved with passing status checks) can
-be found with [this search](https://github.com/angular/angular-cli/pulls?q=is%3Apr+is%3Aopen+review%3Aapproved+status%3Asuccess).
+be found with [this search](https://github.com/angular/angular-cli/pulls?q=is%3Apr+is%3Aopen+label%3A%22PR+action%3A+merge%22).
 This list should be checked daily and any ready PRs should be merged. For each
 PR, check the `PR target` label to understand where it should be merged to. If
 `master` is targetted, then click "Rebase and Merge". If the PR also targets a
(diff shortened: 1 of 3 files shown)


docs/process/release.md
```diff
@@ -18,7 +18,7 @@ TBD
 ## Merging PRs
 
 The list of PRs which are currently ready to merge (approved with passing status checks) can
-be found with [this search](https://github.com/angular/angular-cli/pulls?q=is%3Apr+is%3Aopen+review%3Aapproved+status%3Asuccess).
+be found with [this search](https://github.com/angular/angular-cli/pulls?q=is%3Apr+is%3Aopen+review%3Aapproved+status%3Asuccess+label%3A%22PR+action%3A+merge%22).
```
filipesilva: I wonder if `review:approved status:success` are necessary. There are cases where those might not be true and something is still ready for a merge, but with a caretaker comment. 

If we have a label that explicitly says "caretaker, please merge this", then that should be the source of truth. If there's a PR with the label that isn't ready to be merged, the caretaker should remove the label and say why it's being removed.

Having the meaning of the label be conditional on other checks seems to me is an opportunity for some PRs to be forgotten.

dgp1130: That's a good point, if the author makes a mistake it is good to show it to the caretaker who can make the intelligent decision on how to handle it.

The only other concern I see is that a caretaker might not manually verify that a PR is approved and passing CI and accidentally merge something which hasn't passed those checks. It seems that GitHub has the "Rebase and merge" button in bright red in this case, so that would be a tough mistake to make and probably not worth worrying about.

Removed `review:approved` and `status:success`.

PR #16424, opened 2019-12-11
fix(@ngtools/webpack): invalidate ngcc processor cache

### LINKED ISSUE #16397: Adding a dependency with yarn while serving the app breaks ngcc generated files import
# 🐞 Bug report

- [ ] new
- [ ] build
- [X] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the previous version in which this bug was not present was: 8.2.14

### Description

When running Angular 9 with `ng serve` then installing a dependency with `yarn`, `yarn` will overwrite angular dependencies' `package.json` files and remove `__ivy_ngcc__` files generated by `ngcc` when `ng serve` starts.
This produces an error on `ng serve` as it can't find IVy generated files like `./node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js`

Maybe `ng serve` could watch for these files and rebuild them if they disappear?

## 🔥 Exception or Error
<pre><code>
ERROR in ./node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js
Module build failed: Error: ENOENT: no such file or directory, open '/Users/r/Desktop/demo/node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js'
</code></pre>

filipesilva:
Fix #16397

### DIFF
--- a/packages/ngtools/webpack/src/ngcc_processor.ts
+++ b/packages/ngtools/webpack/src/ngcc_processor.ts
@@ -54,7 +54,8 @@ export class NgccProcessor {
     resolvedModule: ts.ResolvedModule | ts.ResolvedTypeReferenceDirective,
   ): void {
     const resolvedFileName = resolvedModule.resolvedFileName;
-    if (!resolvedFileName || moduleName.startsWith('.') || this._processedModules.has(moduleName)) {
+    if (!resolvedFileName || moduleName.startsWith('.')
+      || this._processedModules.has(resolvedFileName)) {
       // Skip when module is unknown, relative or NGCC compiler is not found or already processed.
       return;
     }
@@ -62,7 +63,7 @@ export class NgccProcessor {
     const packageJsonPath = this.tryResolvePackage(moduleName, resolvedFileName);
     if (!packageJsonPath) {
       // add it to processed so the second time round we skip this.
-      this._processedModules.add(moduleName);
+      this._processedModules.add(resolvedFileName);
 
       return;
     }
@@ -73,7 +74,7 @@ export class NgccProcessor {
       accessSync(packageJsonPath, constants.W_OK);
     } catch {
       // add it to processed so the second time round we skip this.
-      this._processedModules.add(moduleName);
+      this._processedModules.add(resolvedFileName);
 
       return;
     }
@@ -97,7 +98,11 @@ export class NgccProcessor {
     // tslint:disable-next-line:no-any
     (this.inputFileSystem as any).purge(packageJsonPath);
 
-    this._processedModules.add(moduleName);
+    this._processedModules.add(resolvedFileName);
+  }
+
+  invalidate(fileName: string) {
+    this._processedModules.delete(fileName);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


filipesilva:
@clydin highlighted that a risk in these cache-related cases is to end up having stale or incorrect modules being used, but that letting this fix sit in RC for a while will help and we can always pull it.

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -54,15 +54,16 @@ export class NgccProcessor {
     resolvedModule: ts.ResolvedModule | ts.ResolvedTypeReferenceDirective,
   ): void {
     const resolvedFileName = resolvedModule.resolvedFileName;
-    if (!resolvedFileName || moduleName.startsWith('.') || this._processedModules.has(moduleName)) {
+    if (!resolvedFileName || moduleName.startsWith('.')
```
alan-agius4: This will cause a module to be processed multiple times which might slow down the compilation a bit, unless the typings are not flattened, example RxJs.

This is because this method will be called for every dts file resolved.

filipesilva: Will ngcc actually process the module or skip it because it can see it's processed already? I think it skips it because the logs I saw only showed the ngcc logging messages once.

We could probably make cache take into account the relation between `moduleName` and `resolvedFileName` but I think it would introduce some extra risk actually. That's because the authority isn't this cache, but the ngcc cache. And it can actually happen that the whole module is already compiled but our cache doesn't know, since ngcc will compile a module dependencies anyway.

If we could instead invalidate whole modules directly, instead of filenames, I don't think there would be much risk. We can do that using `ts.resolveModuleName`, which is what the compiler host `resolveModuleNames` method that calls `processModule` does anyway. But this now means module resolution is invoked for file invalidations.

I'm not super keen on trying to make this cache much smarter in general because of the risks @clydin pointed out. The perf improvement seems a tradeoff too, unless we made it smarter. @alan-agius4 do you think it's worth it?

alan-agius4: @filipesilva it does skip it, however it will try to read the module and check that it is not already processed multiple times for each module causing quite a number of addition IO ops.

That said, I think we can leave it for now, and tackle it afterwards if we see a big slow down with a lot of dependencies.

clydin: I think it is also important to note that `resolveModuleName` resolves to `.d.ts` files.  Are all the relevant `.d.t.s` files watched in this situation?

filipesilva: I saw that `.d.ts` files for direct dependencies were being invalidated, but did not verify those other cases.

@petebacondarwin are there cases where the `.d.ts` resolved by `ts.resolveModuleName` is **not** the `.d.ts` file processed by `ngcc`?

Issue #13829, opened 2019-03-06
Incorrect behavior of isBinary util


nickbullock:
# 🐞 Bug report

Hi!

### Command (mark with an `x`)

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Is this a regression?

No.

### Description

I'm trying to generate file from custom schema.
I've found incorrect behavior caused by @angular-devkit/schematics/src/rules/utils/is-binary.js
This function takes small chunks (24) of file and trying to detect is file binary or not.
So if file starts with string contains for example cyrrilic characters, isBinary will return incorrect result (true) because some symbols will be broken after cutting of 24 bits.
@schematics/angular/component/index.js
Call stack is rootFn => applyTemplates => applyContentTemplate => isBinary
I think this lib implements this function more correctly https://github.com/gjtorikian/isBinaryFile

## 🔬 Minimal Reproduction

I've builded https://github.com/angular/angular-cli/tree/master/packages/schematics/angular/component by tsc and then changed template of component schema (schematics/component/files/__name@dasherize@if-flat__/__name@dasherize__.component.ts.template), added these lines to top (comment block)
```
/**
 * @overview     ЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯЯ
 * @copyright    Copyright (c) copyright, 2019
 */
```

and then run: ng g c foo

## 🔥 Exception or Error
Template file is not parsed because isBinary returned true, but it is not binary file

## 🌍 Your Environment

<pre><code>

windows 10
node 10
"@angular-devkit/core": "^7.3.4",
"@angular-devkit/schematics": "^7.3.4",
"@angular/cli": "^7.3.4",
"@types/jasmine": "^3.0.0",
"@types/node": "^8.0.31"
"jasmine": "^3.0.0",
"typescript": "3.2.4"
</code></pre>

**Anything else relevant?**

PR #16440, opened 2019-12-13
fix(@angular-devkit/schematics): fix the `isBinary` function

arturovt:
Closes: #13829

The problem with the current implementation is that the function reads only the first _calculated from Math.min_ bytes of the string. UTF-8 can have characters that consist of more than 2 bytes. JS character is a 2-byte size.

In this implementation, there is a high probability of reading only part of the character. For example in this case:
```ts
// `Hello world*Hello world*Hello world` in Japanese
const buffer = Buffer.from('こんにちは世界*こんにちは世界*こんにちは世界');
const chunkLength = 24;
const chunkBegin = 0;
const chunkEnd = Math.min(buffer.length, chunkBegin + chunkLength);
const contentChunkUTF8 = buffer.toString('utf-8', chunkBegin, chunkEnd);

console.log(buffer.toString('utf-8', chunkBegin, chunkEnd));
```
unknown characters (`�`) will be also printed to the console and we pass this string to the `isBinary` function then it will return `true` (but it's falsy).

The Node.js `util` module has the `TextDecoder` class out of the box that decodes strings using provided encoding or throws.


packages/angular_devkit/schematics/src/rules/utils/is-binary.ts
```diff
@@ -5,24 +5,16 @@
...
-    if ( charCode === 65533 || charCode <= 8 ) {
-      // 8 and below are control characters (e.g. backspace, null, eof, etc.).
-      // 65533 is the unknown character.
-      return true;
-    }
+export function isBinary(buffer: Buffer): boolean {
+  try {
+    // `TextDecoder.prototype.decode` will return a UTF-8 string
+    // if there are no decoding errors occured.
+    return !decoder.decode(buffer);
```
clydin: Attempting to decode the entire file is going to be highly inefficient.  The  buffer size could be unbounded (megabytes or even gigabytes).  A minimal amount of bytes should be analyzed when possible.  If 24 bytes is insufficient this could be increased probably up to even a kilobyte if needed.

packages/angular_devkit/schematics/src/rules/utils/is-binary_spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  it('should be false if provided value is a UTF-8 buffer', () => {
+    const fooBuffer = Buffer.from('foo');
+    const jpBuffer = Buffer.from('こんにちは世界！');
+
+    expect(isBinary(fooBuffer)).toBe(false);
+    expect(isBinary(jpBuffer)).toBe(false);
+  });
+
+  it('should be true if provided value is a buffer with unknown ASCII character', () => {
+    const buffer = Buffer.from([128]);
```
clydin: This is not correct.  128 is a valid extended ASCII character.  UTF-8 supports extended ASCII via fallback.

arturovt:
```js
const { TextDecoder } = require('util');

const decoder = new TextDecoder('utf-8', { fatal: true });

console.log(decoder.decode(Buffer.from([128])));
```

```
TypeError [ERR_ENCODING_INVALID_ENCODED_DATA]: The encoded data was not valid for encoding utf-8
```

packages/angular_devkit/schematics/src/rules/utils/is-binary.ts
arturovt: Agreed. I was actually thinking about some external package that provides such type of functionality. I don't know if it's a good idea to bring some package to this codebase.

The problem is that if we select a chunk based on the number of bytes then yes - it will not work correctly. But it will work if we select it based on number of characters.

https://github.com/parshap/truncate-utf8-bytes/blob/v1.0.2/lib/truncate.js

Seems like it's being used a lot regarding "used by" button. It doesn't work with buffers but we could copy-paste and change it a little bit, it's short.

Just thinking aloud...

clydin: Based on the singular usage of this function, I think we should consider removing it completely and instead adjust the logic within the function that uses this helper (https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/schematics/src/rules/template.ts#L52).  Direct usage of the `TextDecoder` to get the content combined with a try/catch block that handles the invalid data error should avoid the entire issue.

packages/angular_devkit/schematics/src/rules/utils/is-binary_spec.ts
clydin: That is because the algorithm used by `TextDecoder` does not implement fallback.

packages/angular_devkit/schematics/src/rules/utils/is-binary.ts
arturovt: Removed `is-binary.ts` file and moved all logic in the `template.ts` file.

packages/angular_devkit/schematics/src/rules/utils/is-binary_spec.ts
arturovt: Interesting...

Just been googling for some results about this and every answer on the StackOverflow or wherever else says like:
> JavaScript deals only in Unicode, not Extended ASCII.

clydin: It _kind of_ deals in unicode.  This article provides some background on the overall unicode issue: https://mathiasbynens.be/notes/javascript-unicode

clydin: The underlying issue here though is that `isBinary` shouldn't discriminate based on the text encoding method.  An additional reason why I think the new approach is more appropriate.

packages/angular_devkit/schematics/src/rules/template.ts
```diff
@@ -48,18 +48,22 @@ export interface PathTemplateOptions {
...
+    const { path, content } = entry;
+
+    try {
+      const decodedContent = decoder.decode(content);
+
+      return {
+        path,
+        content: Buffer.from(templateImpl(decodedContent, {})(options)),
+      };
+    } catch {
```
clydin: Can you add a check for the error code `ERR_ENCODING_INVALID_ENCODED_DATA` and re-throw if it is not that code?

arturovt: Done.

PR #16459, opened 2019-12-16
Revert "feat(@ngtools/webpack): add support for raw-loader 2 and 3 (#…

choco505:
99ba5ebaf7e457cebe48bd71d22635f074f49682.


alan-agius4:
Thanks for this, however I am closing the PR as I don't think we'd want to rollback to raw-loader version 1. Unless there is a very valid reason. 

If you are experiencing a problem, please open a new issue describing the current and expected behaviour.

Thanks.

PR #16524, opened 2020-01-02
fix(@angular-devkit/build-angular): add sourceMappingURL comment for …

### LINKED ISSUE #16522: Angular 8.3.21, ng build with target es 2015 doesn't show sourcemapped ts/css files in browser for debugging
Updated Angular cli to latest version (8.3.21) and from then my new projects are not showing ts/css files in the chrome/firefox browser dev tools.

Sourcemap files are available in the dist folder. 

Modifying the _target_ flag to _**es2015**_ in _tsconfig.json_ builds only the es2015 browser support files and then am able to find the ts files in the developer tools debugger section.

Thank you!

alan-agius4:
…ES2015 during differential loading

When having differential loading enabled we only add the `sourceMappingURL` comment when optimization is enabled, because we only process these bundles when we enabling optimization.

With this change we now process such bundles even when optimization is disabled and add `sourceMappingURL` when source maps are enabled and not hidden.

Closes #16522

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
@@ -39,7 +39,7 @@ export interface ProcessBundleOptions {
   optimize?: boolean;
   optimizeOnly?: boolean;
   ignoreOriginal?: boolean;
-  cacheKeys?: (string | null)[];
+  cacheKeys?: (string | undefined)[];
   integrityAlgorithm?: 'sha256' | 'sha384' | 'sha512';
   runtimeData?: ProcessBundleResult[];
   replacements?: [string, string][];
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -155,59 +155,21 @@ export async function process(options: ProcessBundleOptions): Promise<ProcessBun
...
-    await cachePut(
-      downlevelCode,
-      options.cacheKeys[CacheKey.DownlevelCode],
-      result.downlevel.integrity,
-    );
-    fs.writeFileSync(downlevelPath, downlevelCode);
+    result.downlevel = await processBundle({
+      ...options,
+      code: downlevelCode,
+      map: downlevelMap && JSON.stringify(downlevelMap),
```
clydin: Can `map` be a string or a `RawSourceMap` here with a check in the `processBundle` function to handle either?  Sourcemaps can easily be in the megabyte range.  Serializing and immediately deserializing can have performance and memory usage concerns.

clydin:
Also, this will most likely not apply cleanly to 8.3.x so a more minimal change set may be needed for that branch.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
alan-agius4: Yeah make sense! Done.

alan-agius4:
Blocking until the LTS version is green. As I am think it might be related to caching.

PR #16525, opened 2020-01-02
[LTS] fix(@angular-devkit/build-angular): add sourceMappingURL comment for …

alan-agius4:
Cherry-pick of https://github.com/angular/angular-cli/pull/16524 for LTS


alan-agius4:
Seems like there is a missing comment or something which is still causing patch to fail the tests. 

Will investigate.

PR #16539, opened 2020-01-03
fix(@angular-devkit/build-angular): change css optimizer from clean-css with cssnano

### LINKED ISSUE #13854: CleanCss breaks :host-context(selector selector)
# 🐞 Bug report

```
- [ ] new
- [x] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
CleanCss plugin breaks :host-context(selector selector) (space is important)

https://github.com/angular/angular/issues/23034#issue-309249510
#10454

## 🔬 Minimal Reproduction

Use :host-context(something .somethingElse) and run ng build --prod

My current fix is to remove CleanCss plugin with ngx-build-plus
```
cfg.optimization.minimizer = cfg.optimization.minimizer.reduce((acc, minimizer) => {
  return minimizer.constructor.name === 'CleanCssWebpackPlugin' ? acc : [...acc, minimizer];
}, []);
```
without this plugin everything works as expected

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍 Your Environment
<pre><code>

Angular CLI: 7.3.5
Node: 10.15.1
OS: darwin x64
Angular: 7.2.8
... platform-server, router

@angular-devkit/architect         0.13.5
@angular-devkit/build-angular     0.13.5
@angular-devkit/build-optimizer   0.13.5
@angular-devkit/build-webpack     0.13.5
@angular-devkit/core              7.3.5
@angular-devkit/schematics        7.3.5
@angular/cli                      7.3.5
@ngtools/webpack                  7.3.5
@schematics/angular               7.3.5
@schematics/update                0.13.5
rxjs                              6.4.0
typescript                        3.2.4
webpack                           4.29.0
</code></pre>

**Anything else relevant?**

### LINKED ISSUE #16123: CSS Variables are lost when running ng build --prod
When running ng build --prod in release candidate of Angular 9 the CSS variables are removed from some elements please see screenshots  these are originally in styles.css

Development Mode
Screenshot 2019-11-04 at 11 08 27

Production Mode
Screenshot 2019-11-04 at 11 08 55

alan-agius4:
Closes #16123 and closes #13854


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts
```diff
@@ -66,6 +67,14 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
       }),
       autoprefixer(),
     ];
+
+    if (wco.buildOptions.optimization.styles) {
+      postCssPlugins.push(
+        cssNano({ preset: 'default' }),
```
clydin: This will operate differently than the current setup.  Each input file will be optimized in isolation whereas with the previous approach the final output was optimized as a whole.  The later has advantages in terms of both performance and file size reduction.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/optimize-css-webpack-plugin.ts
```diff
@@ -44,21 +45,6 @@ export class CleanCssWebpackPlugin {
 
   apply(compiler: Compiler): void {
     hook(compiler, (compilation: compilation.Compilation, chunks: compilation.Chunk[]) => {
-      const cleancss = new CleanCSS({
-        compatibility: 'ie9',
-        level: {
-          2: {
-            skipProperties: [
-              'transition', // Fixes #12408
-              'font', // Fixes #9648
```
filipesilva: Does the new approach keep these issues fixed?

packages/angular_devkit/build_angular/test/browser/aot_spec_large.ts
```diff
@@ -40,30 +40,6 @@ describe('Browser Builder AOT', () => {
     await run.stop();
   });
 
-  it('shows warnings for component styles', async () => {
```
filipesilva: Can you provide some context on why is this test (and the next one) removed?

filipesilva:
I think this should target master only though, there's risk of changes in behaviour and there isn't a big need to couple that with 9.0.

packages/angular_devkit/build_angular/test/browser/aot_spec_large.ts
alan-agius4: CSSnano doesn't output such warning for these.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/optimize-css-webpack-plugin.ts
alan-agius4: Yes.

IgorMinar:
if we are deciding between making this change in 9.0 vs 9.1 then let's do it in 9.0. If we think the risk of breaking changes is high, then we should not make this change in 9.x at all and wait for v10.

The reasoning here is that this PR fixes features enabled by v9.0 framework changes so it would be good to get them to 9.0 as a fix, but if the risk of making this change is high then we should not make it in 9.x at all.

@alan-agius4 I'll let you assess which one is right given the guidelines above.

alan-agius4:
Unless @clydin or @filipesilva feel otherwise, I think we should go ahead and put it in version 9 as opposed to version 10. While there is a risk of breakages, I think it is manageable, worst case we can land a trivial fix to disable the rule that causes the problem.

alan-agius4:
Patch version: https://github.com/angular/angular-cli/pull/16647

PR #16573, opened 2020-01-07
docs: address remaining review items

filipesilva:
Followup from https://github.com/angular/angular-cli/pull/16400

### DIFF
--- a/docs/design/build-system.md
+++ b/docs/design/build-system.md
@@ -1,7 +1,7 @@
 # Build System
 
 Angular CLI includes a first-party build system for Angular applications distributed as `@angular-devkit/build-angular`.
-This build system is responsible for creating a standalone application from user source files and third-party dependencies.
+This build system is responsible for creating a standalone single-page application (SPA) from user source files and third-party dependencies.
 
 `@angular-devkit/build-angular` itself integrates with the rest of Angular CLI by being an [Architect builder](https://angular.io/guide/cli-builder).
 This document describes a top level view of the functionality in `@angular-devkit/build-angular`, referred as just the "build system".
@@ -17,13 +17,22 @@ Many tools are used in this process, and most of these steps happen within a [We
 We maintain a number of Webpack-centric plugins in this repository, some of these are public but most are private since they are very specific to our setup.
 
 
+## Overview diagram
+
+Below is a diagram of processing sources go through.
+It's not strictly accurate because some options remove certain processing stages while adding others.
+This diagram doesn't show these conditionals and only shows the possible processing steps.
+Relative paths, such as `./raw-css-loader.ts`, refer to internal plugins, while other names usually refer to public npm packages.
+
+![Overview diagram](https://g.gravizo.com/source/svg?https://raw.githubusercontent.com/angular/angular-cli/master/docs/design/build-system-overview.dot)
+
 ## Loading and processing sources
 
 Sources for Angular CLI browser apps are comprised of TypeScript files, style sheets, assets, scripts, and third party dependencies.
 A given build will load these sources from disk, process them, and bundle them together.
 
 
-### TypeScript
+### TypeScript and Ahead-Of-Time Compilation
 
 Angular builds rely heavily on TypeScript-specific functionality for [Ahead-of-Time template compilation](https://angular.io/guide/aot-compiler) (AOT).
 Outside Angular CLI, this is performed by the Angular Compiler (`ngc`), provided by `@angular/compiler-cli`.
@@ -34,7 +43,8 @@ During compilation we also perform a number of code transformations using TypeSc
 
 AOT compilation requires loading HTML and CSS resources, referenced on Angular Components, as standalone strings with no external dependencies.
 However, Webpack compilations operate on the basis of modules and references between them.
-To obtain the standalone string we compile resources using a separate Webpack child compilation then extract the results.
+It's not possible to get the full compilation result for a given resource before the end of the compilation in webpack.
+To obtain the standalone string for a HTML/CSS resource we compile it using a separate Webpack child compilation then extract the results.
 These child compilations inherit configuration and access to the same files as the parent compilation, but have their own compilation life cycle and complete independently.
 
 The build system allows specifying replacements for specific files by replacing what path is loaded from the virtual file system.
@@ -47,8 +57,6 @@ Global stylesheets are injected into the `index.html` file, while component styl
 
 The build system supports plain CSS stylesheets as well as the Sass, LESS and Stylus CSS pre-processors.
 Stylesheet processing functionality is provided by `sass-loader`, `less-loader`, `stylus-loader`, `postcss-loader`, `postcss-import`, augmented in the build system by custom webpack plugins.
-
-todo: style resources
 
 
 ### Assets
@@ -72,8 +80,9 @@ Stylesheet third party dependencies are treated mostly the same as sources.
 JavaScript third party dependencies suffer a more involved process.
 They are first resolved to a folder in `node_modules` via [Node Module Resolution](https://nodejs.org/api/modules.html#modules_modules).
 A given module might have several different entry points, for instance one for use in NodeJS and another one for using in the browser.
-Each entry point is listed in under a name pointing at a js file in that module's `package.json`.
-We use `es2015 > browser > module > main` a priority list, where the first key matched name determines which entry point to use.
+Although `package.json` only officially supports listing one entry point under the `main` key, it's common for npm packages to list [other entry points](https://2ality.com/2017/04/setting-up-multi-platform-packages.html).
+Each entry point is listed in under a key in that module's `package.json` whose value is a string containing a relative file path.
+We use `es2015 > browser > module > main` a priority list of keys, where the first key matched name determines which entry point to use.
 For instance, for a module that has both `browser` and `main` entry points, we pick `browser`.
 
 Once the actual JavaScript file is determined, it is loaded into the compilation together with it's source map.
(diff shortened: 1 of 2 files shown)


docs/design/build-system.md
```diff
@@ -1,7 +1,7 @@
 # Build System
 
 Angular CLI includes a first-party build system for Angular applications distributed as `@angular-devkit/build-angular`.
-This build system is responsible for creating a standalone application from user source files and third-party dependencies.
+This build system is responsible for creating a standalone SPA application from user source files and third-party dependencies.
```
filipesilva: Addresses https://github.com/angular/angular-cli/pull/16400#discussion_r356711074

docs/design/build-system.md
```diff
@@ -17,13 +17,46 @@ Many tools are used in this process, and most of these steps happen within a [We
 We maintain a number of Webpack-centric plugins in this repository, some of these are public but most are private since they are very specific to our setup.
 
 
+## Overview diagram
```
filipesilva: Addresses https://github.com/angular/angular-cli/pull/16400#discussion_r358402434 and https://github.com/angular/angular-cli/pull/16400#discussion_r358403869.

docs/design/build-system.md
```diff
@@ -17,13 +17,46 @@ Many tools are used in this process, and most of these steps happen within a [We
...
+```
+
 ## Loading and processing sources
 
 Sources for Angular CLI browser apps are comprised of TypeScript files, style sheets, assets, scripts, and third party dependencies.
 A given build will load these sources from disk, process them, and bundle them together.
 
 
-### TypeScript
+### TypeScript and Ahead-Of-Time Compilation
```
filipesilva: Addresses https://github.com/angular/angular-cli/pull/16400#discussion_r358401592

docs/design/build-system.md
```diff
@@ -34,7 +67,8 @@ During compilation we also perform a number of code transformations using TypeSc
 
 AOT compilation requires loading HTML and CSS resources, referenced on Angular Components, as standalone strings with no external dependencies.
 However, Webpack compilations operate on the basis of modules and references between them.
-To obtain the standalone string we compile resources using a separate Webpack child compilation then extract the results.
+It's not possible to get the full compilation result for a given resource before the end of the compilation in webpack.
+To obtain the standalone string for a HTML/CSS resource we compile it using a separate Webpack child compilation then extract the results.
```
filipesilva: Hopefully addresses https://github.com/angular/angular-cli/pull/16400#discussion_r356710447.

filipesilva: @dgp1130 does this answer the question? Let me know if it's still not very clear. I think this is a rather complicated topic.

docs/design/build-system.md
```diff
@@ -72,8 +104,9 @@ Stylesheet third party dependencies are treated mostly the same as sources.
 JavaScript third party dependencies suffer a more involved process.
 They are first resolved to a folder in `node_modules` via [Node Module Resolution](https://nodejs.org/api/modules.html#modules_modules).
 A given module might have several different entry points, for instance one for use in NodeJS and another one for using in the browser.
-Each entry point is listed in under a name pointing at a js file in that module's `package.json`.
-We use `es2015 > browser > module > main` a priority list, where the first key matched name determines which entry point to use.
+Although `package.json` only officially supports listing one entry point under the `main` key, it's common for npm packages to list [other entry points](https://2ality.com/2017/04/setting-up-multi-platform-packages.html).
```
filipesilva: Addresses https://github.com/angular/angular-cli/pull/16400#discussion_r356715022, cc @dgp1130.

docs/design/build-system.md
```diff
@@ -122,13 +155,13 @@ Large projects can also opt-out of AOT compilation for faster rebuilds.
 Angular CLI focuses on enabling tree-shaking (removing unused modules) and dead code elimination (removing unused module code).
 These two categories have high potential for size reduction because of network effects: removing code can lead to more code being removed.
 
-The main tool we use to achieve this goal are the compression capabilities of [Terser](https://github.com/terser/terser).
+The main tool we use to achieve this goal are the dead code elimination capabilities of [Terser](https://github.com/terser/terser).
```
filipesilva: Addresses https://github.com/angular/angular-cli/pull/16400#discussion_r358403721.

docs/design/build-system.md
```diff
@@ -122,13 +155,13 @@ Large projects can also opt-out of AOT compilation for faster rebuilds.
...
 We also use Terser's mangling, by which names, but not properties, are renamed to shorter forms.
 The main characteristics of Terser to keep in mind is that it operates via static analysis and does not support the indirection introduced by module loading.
 Thus the rest of the pipeline is directed towards providing Terser with code that can be removed via static analysis in large single modules scopes.
 
 To this end we developed [@angular-devkit/build-optimizer](https://github.com/angular/angular-cli/tree/master/packages/angular_devkit/build_optimizer), a post-processing tool for TS code.
-Build Optimizer searches for code patterns produced by the TypeScript and Angular compiler that are known to inhibit dead code elimination, and converts them into equivalent structures that enable it instead.
+Build Optimizer searches for code patterns produced by the TypeScript and Angular compiler that are known to inhibit dead code elimination, and converts them into equivalent structures that enable it instead (the link above contains some examples).
```
filipesilva: Addresses https://github.com/angular/angular-cli/pull/16400#discussion_r358404211. I didn't add an example because it would always be some 10 lines total. But the link above contains good examples right at start.

docs/design/build-system.md
filipesilva: @clydin @alan-agius4 PTAL as well.

docs/design/build-system.md
dgp1130: LGTM.

docs/design/build-system.md
dgp1130: I'd recommend using Graphviz to render this graph if possible. If you pull out a dot graph to a separate file, you can use http://www.gravizo.com/ to render it in the markdown.

docs/design/build-system.md
dgp1130: So child compilations are necessary to get all the information at the end of a build. From that, my next question is "Why do we need a full compilation result for intermediate resources"? What info does the result provide which is needed in the middle of the pipeline?

docs/design/build-system.md
```diff
@@ -72,8 +104,9 @@ Stylesheet third party dependencies are treated mostly the same as sources.
 JavaScript third party dependencies suffer a more involved process.
 They are first resolved to a folder in `node_modules` via [Node Module Resolution](https://nodejs.org/api/modules.html#modules_modules).
 A given module might have several different entry points, for instance one for use in NodeJS and another one for using in the browser.
-Each entry point is listed in under a name pointing at a js file in that module's `package.json`.
-We use `es2015 > browser > module > main` a priority list, where the first key matched name determines which entry point to use.
+Although `package.json` only officially supports listing one entry point under the `main` key, it's common for npm packages to list [other entry points](https://2ality.com/2017/04/setting-up-multi-platform-packages.html).
+Each entry point is listed in under a key in that module's `package.json` whose value is string containing a relative file path.
```
dgp1130: whose value is **a** string

docs/design/build-system.md
clydin:
```suggestion
This build system is responsible for creating a standalone single-page application (SPA) from user source files and third-party dependencies.
```

docs/design/build-system.md
clydin: The style sheet pipeline appears to be reversed.  The preprocessors are first with the raw/style loaders last.

docs/design/build-system.md
filipesilva: That's this bit:
> AOT compilation requires loading HTML and CSS resources, referenced on Angular Components, as standalone strings with no external dependencies.

docs/design/build-system.md
alan-agius4: Yes indeed, also the `scripts array` go through `terser-webpack-plugin` and not `terser`.

docs/design/build-system.md
dgp1130: So it's impossible to load a file as a standalone string with no dependencies without using child compilations? That feels like something that should be possible in the regular compilation, unless I'm missing something?

docs/design/build-system.md
dgp1130: Makes sense, thanks for the link.

docs/design/build-system.md
clydin: Not within a webpack plugin. Child compilations are kind of the standard method for these kind of operations.  Webpack also outputs JS and not a standalone string.  So the resource manager in the AOT plugin needs to execute the child compilation output in a separate VM context to get the actual string content for the resource ([code](https://github.com/angular/angular-cli/blob/0a83923f912c5514aee4789f0585f162f13495af/packages/ngtools/webpack/src/resource_loader.ts#L156)).

filipesilva: It's not possible in webpack because webpack has a phased compilation. AOT compilation needs the final string (with all optimizations, and no module references) for CSS/HTML resources while compiling TS. But at that point webpack is still in the loading all resources, and hasn't gotten to the optimization phase.

If we were to get the resource string at that point during the loading phase it would be both unoptimized, and referencing other resources as well (if using imports).

dgp1130: Ok, I think I see. So we need the fully optimized and bundled HTML/CSS resources, but those operations don't happen until the end of the compilation because they inherently require all the files to be processed. However, since we know they won't depend on other TS artifacts, we can run a child compilation which does not include TS files to let it compile the HTML/CSS files in their entirety, and use those during the TS compilations.

Does this sound like an accurate assessment?

filipesilva: Yes that is accurate. Except the part about not including TS files. Child compilations are still full compilations and can do (most) anything.

For instance the web worker support uses a child compilation for building the web worker separately. Not really because we designed it that way, but because https://github.com/GoogleChromeLabs/worker-plugin uses child compilations and we support TS workers.

docs/design/build-system.md
filipesilva: added

docs/design/build-system.md
filipesilva: fixed

docs/design/build-system.md
filipesilva: Ah yes the good old "reverse array order". How could I forget. Again.

filipesilva: This graphizo thing is nice, great suggestion @dgp1130. Updated to be a inline image in DOT.

docs/design/build-system.md
dgp1130: Right, I understand it can include TS files. I was saying that it was choosing not to in that particular case. This seems reasonable to me.

docs/design/build-system.md
dgp1130: Unfortunately the preview doesn't seem to be loading correctly. The one issue I've had with the service is getting it to work without binding to a specific `master` URL. Embedded the graph directly never seems to work for me. I've had better luck putting the DOT graph in a separate file and then linking to that ([example](https://github.com/dgp1130/chatter/#architecture)), though this doesn't work for previews.

filipesilva: Well that's unfortunate. It does show up correctly on my VSCode markdown preview:
image

dgp1130: Graph LGTM. Only minor nit is that I tend to prefer `node [shape=rectangle];` as I think that generally looks better for most graphs. Up to you if you want to do that.

filipesilva: Updated to a separated .dot file.

filipesilva: Updated to rectangle shapes, looks like this now:

image

filipesilva:
@alan-agius4 replaced now, thanks for the heads up.

filipesilva:
Note to caretaker: this PR cannot affect any of those checks and is good to merge.

PR #16599, opened 2020-01-08
build: use TypeScript 3.7

filipesilva:


packages/angular_devkit/build_optimizer/package.json
```diff
@@ -12,7 +12,7 @@
     "loader-utils": "1.2.3",
     "source-map": "0.7.3",
     "tslib": "1.10.0",
-    "typescript": "3.7.4",
+    "typescript": "3.6.4",
```
filipesilva: Build Optimizer has to stay on 3.6 for now because of https://github.com/microsoft/TypeScript/issues/36098. But that should be fine since its TS version is standalone.

PR #16601, opened 2020-01-09
Improve source map generation quality

clydin:
The `webpack-sources` package provides the `SourceMapSource` class which provides sourcemap merging functionality including identity mapping.  This provides for higher quality sourcemaps than the previous merging strategy.  The merging logic is also simplified as a result of the change.


filipesilva:
Awesome 🎉

PR #16632, opened 2020-01-10
build: update tree-kill to version ^1.2.2

sethjones:
Update tree-kill dependency to ^1.2.2 in response to https://npmjs.com/advisories/1432.  Advisory affects tree-kill dependency, not angular-cli.


clydin:
For validation and record keeping purposes, please add a note to the commit message that this advisory does not affect the CLI.

Issue #16629, opened 2020-01-10
Dependency Tree-kill 1.2.1 Vulnerability 


phoenix09208:
angular-devkit/build-angular@0.803.22 dependent on Tree-kill 1.2.1,

according to https://www.npmjs.com/advisories/1432, it is prone to command injection, please upgrade dependency to later version. Thanks

DibyodyutiMondal:
This affects angular 9 rc-8 too

cwilby:
A temporary workaround (that also works in situations where package maintainer has yet to release dependencies but y'know, security.)

First, run `npm audit` to determine which version needs to be fixed.

```                                                                                
                       === npm audit security report ===                        
                                                                                
┌──────────────────────────────────────────────────────────────────────────────┐
│                                Manual Review                                 │
│            Some vulnerabilities require your attention to resolve            │
│                                                                              │
│         Visit https://go.npm.me/audit-guide for additional guidance          │
└──────────────────────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Command Injection                                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ tree-kill                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=1.2.2                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ @angular-devkit/build-angular [dev]                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ @angular-devkit/build-angular > tree-kill                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1432                            │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

Ok, we need `tree-kill@1.2.2`.

Go to `https://registry.npmjs.org/${PACKAGE}`, in this case https://registry.npmjs.org/tree-kill and copy `versions.${VERSION}.dist.integrity`.

For `tree-kill@1.2.2`, it's `sha512-L0Orpi8qGpRG//Nd+H90vFB+3iHnue1zSSGmNOOCh1GLJ7rUKVwV2HvijphGQS2UmhUZewS9VgvxYIdgr+fG1A==`.

Finally, open `package-lock.json` and change the require version of `@angular-devkit/build-angular` and the version/integirty for `tree-kill`.

```
{
    ...
[...]

PR #16634, opened 2020-01-10
[8.3.x] fix(@angular-devkit/build-angular): update tree-kill dependency to 1.2.2

clydin:
While the CLI is not affected by the following advisory, this change will address the audit warning and prevent any potential future usage of the package within the CLI from being affected.
Advisory: https://www.npmjs.com/advisories/1432

Closes https://github.com/angular/angular-cli/issues/16629

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     "spdx-satisfies": "^4.0.0",
     "tar": "^4.4.4",
     "through2": "^2.0.3",
-    "tree-kill": "^1.2.0",
+    "tree-kill": "^1.2.2",
     "ts-api-guardian": "0.4.6",
     "ts-node": "^5.0.0",
     "tslint-no-circular-imports": "^0.7.0",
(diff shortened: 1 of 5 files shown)


jplatte:
Why is `@ngtools/webpack` using an exact version requirement in the first place?

alan-agius4:
We pin all transitive dependencies to avoid unexpected breakages when a 3rd party library gets a release.

So before updating a dependency we make sure that it doesn't break any of our tests.

ahmadabulaban:
When the new tag 8.3.23 (that contain this fix) will be available ?

Nikhilkapoor20:
Any update on this ?

kolibrizas:
@Nikhilkapoor20 @ahmadabulaban 
According to https://www.npmjs.com/package/@angular-devkit/build-angular 0.803.23 was released a few hours ago. It fixed tree-kill dependecy and now npm audit succeeds.

PR #16639, opened 2020-01-11
build: update tree-kill to version ^1.2.2

sethjones:
Update tree-kill dependency to ^1.2.2 in response to https://npmjs.com/advisories/1432.  Advisory affects tree-kill package, not angular-cli directly.


alan-agius4:
Thanks

PR #16646, opened 2020-01-11
wip! replace parser

alan-agius4:


petebacondarwin:
"yargs" FTW 🏋 
The title of the PR was a bit misleading initially :-) I thought you were replacing some code parser 🙀

alan-agius4:
Hahaha... someone is watching the CLI repo. 

Actually, I replaced it with the parser that `yargs` uses which is  `yargs-parser`. As the parser alone is enough as we don't need to full `yargs` functionality.

Ps: this is still a POC, hopefully we can land it :)

PR #16648, opened 2020-01-11
feat(@angular-devkit/build-angular): default to NodeJS value for pres…

Toxicable:
…erveSymlinks

Under bazel `preserveSymlinks` would have to be set in two different places, this makes it so it only has to be set once by using the value from NodeJS if it's set.

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -844,8 +844,7 @@
             },
             "preserveSymlinks": {
               "type": "boolean",
-              "description": "Do not use the real path when resolving modules.",
-              "default": false
+              "description": "Do not use the real path when resolving modules."
             },
             "extractLicenses": {
               "type": "boolean",
@@ -1478,8 +1477,7 @@
             },
             "preserveSymlinks": {
               "type": "boolean",
-              "description": "Do not use the real path when resolving modules.",
-              "default": false
+              "description": "Do not use the real path when resolving modules."
             },
             "browsers": {
               "type": "string",
@@ -1855,8 +1853,7 @@
             },
             "preserveSymlinks": {
               "type": "boolean",
-              "description": "Do not use the real path when resolving modules.",
-              "default": false
+              "description": "Do not use the real path when resolving modules."
             },
             "extractLicenses": {
               "type": "boolean",
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -248,8 +248,7 @@
     },
     "preserveSymlinks": {
       "type": "boolean",
-      "description": "Do not use the real path when resolving modules.",
-      "default": false
+      "description": "Do not use the real path when resolving modules. If unset then will default to `true` if NodeJS option --preserve-symlinks is set."
     },
     "extractLicenses": {
       "type": "boolean",
--- a/packages/angular_devkit/build_angular/src/karma/schema.json
+++ b/packages/angular_devkit/build_angular/src/karma/schema.json
@@ -126,8 +126,7 @@
     },
     "preserveSymlinks": {
       "type": "boolean",
-      "description": "Do not use the real path when resolving modules.",
-      "default": false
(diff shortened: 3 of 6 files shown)


Toxicable:
Not sure if this is the right place to put this config, let me know if there's a better place to set it.

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -141,6 +141,9 @@ export async function buildBrowserWebpackConfigFromContext(
     wco.buildOptions.webWorkerTsConfig ? getWorkerConfig(wco) : {},
   ];
 
+  // use the value for preserveSymlinks from NodeJS if it's not set in the config
+  options.preserveSymlinks = options.preserveSymlinks === undefined ? process.execArgv.includes('--preserve-symlinks') : options.preserveSymlinks;
```
alan-agius4: I think if we do this it should be somewhere else and effect all architect builders and not just this one. As this will result in inconsistencies between different builders.

Toxicable: Ah dam, I thought this function was used by all builders. Ideas on where I should be looking I stead?

alan-agius4: https://github.com/angular/angular-cli/blob/30df1470a0f18989db336d50b55a79021ab64c85/packages/angular_devkit/build_angular/src/utils/normalize-builder-schema.ts#L50-L72

Also, in the various schemas you'd need to remove the default, since there is a default of `false`. `options.preserveSymlinks` it will never be `undefined`.

That being said, let's ping @clydin and double check that he's onboard with this change.

clydin:
I’m ok with using it as a default.  However, it would be a breaking change since the default would be different for some people. This could manifest in either broken builds or builds containing different dependencies than anticipated or expected. As a feature, this would probably fall into the v10 timeframe.

Toxicable:
@clydin Sounds good to me, how would we make sure to come back and check this when there's a v10 branch?

docs/documentation/1-x/angular-cli.md
```diff
@@ -83,7 +83,7 @@
     - *progress* (`boolean`): Log progress to the console while building. Default is `true`.
     - *poll* (`number`): Enable and define the file watching poll time period (milliseconds).
     - *deleteOutputPath* (`boolean`): Delete output path before build. Default is `true`.
-    - *preserveSymlinks* (`boolean`): Do not use the real path when resolving modules. Default is `false`.
+    - *preserveSymlinks* (`boolean`): Do not use the real path when resolving modules. Default is set to `true` if NodeJS option --preserve-symlinks is set.
```
alan-agius4: This can be reverted as the docs here are for cli v1

packages/angular_devkit/core/node/resolve.ts
```diff
@@ -90,7 +90,7 @@ export interface ResolveOptions {
 
   /**
    * Whether or not to preserve symbolic links. If false, the actual paths pointed by
-   * the symbolic links will be used. This defaults to true.
+   * the symbolic links will be used. Defaults to `true` if NodeJS option --preserve-symlinks is set.
    */
   preserveSymlinks?: boolean;
```
alan-agius4: This is not effected by the change you did, also this is deprecated logic which is no longer used internally.

alan-agius4:
@Toxicable, I can create a v10 milestone and assign this Pr to it.

packages/angular_devkit/core/node/resolve.ts
Toxicable: Right, added to the scriptions of where I could find "Do not use the real" instead

packages/schematics/angular/utility/config.ts
```diff
@@ -388,7 +388,8 @@ export interface CliConfig {
            */
           deleteOutputPath?: boolean;
           /**
-           * Do not use the real path when resolving modules.
+         * Do not use the real path when resolving modules.
+           * set.
```
alan-agius4: Oops 😝

Issue #16654, opened 2020-01-13
Missing analytics-prompt when running yarn install


aldent95:
# 🐞 Bug report

### Is this a regression?
Yes, the previous version in which this bug was not present was: 7.2.14

### Description
When we try and run yarn install through our CI pipeline, we get an error saying that the module ./analytics-prompt cannot be found.

I've tried the following, which is all I can think of, to solve the issue:

- Disable the analytics for the project in the angular.json file
- Disable the analytics for the server using the environment variable
- Disable the analytics for the user by running ng analytics off for the CI user
- Removed the --no-interaction option from our yarn command

None of the above fixed the issue.
The weird thing is, the install works fine for my user account on the same server, same directory, exact same command. It also works fine on my local machine.

I've done all I can think of so hopefully someone here has some ideas.

## 🔬 Minimal Reproduction

Full install command that causes the issue:
yarn install --ignore-optional --prod --frozen-lockfile

Beyond that I can't really provide any other reproduction steps as I can't reproduce it outside of our pipeline.

## 🔥 Exception or Error
<pre><code>
C:\Temp\ProjectName\feature\Angular_Update\A@2>yarn install --ignore-optional --prod --frozen-lockfile 
yarn install v1.9.2
warning package.json: No license field
warning project-name@0.0.1: No license field
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.9: The platform "win32" is incompatible with this module.
info "fsevents@1.2.9" is an optional dependency and failed compatibility check. Excluding it from installation.
info fsevents@2.1.2: The platform "win32" is incompatible with this module.
info "fsevents@2.1.2" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning " > @angular/core@8.2.14" has unmet peer dependency "zone.js@~0.9.1".
warning " > @angular/flex-layout@7.0.0-beta.24" has incorrect peer dependency "@angular/cdk@^7.0.0-rc.0".
warning " > ng-packagr@5.7.1" has unmet peer dependency "tslib@^1.9.0".
warning " > codelyzer@5.2.1" has unmet peer dependency "tslint@^5.0.0".
[4/4] Building fresh packages...
error C:\Temp\ProjectName\feature\Angular_Update\A@2\node_modules\@angular\cli: Command failed.
Exit code: 1
Command: node ./bin/postinstall/script.js
Arguments: 
[...]

PR #16673, opened 2020-01-15
fix(@angular/cli): prevent postinstall script errors from blocking install

clydin:
Closes: #16654


dgp1130:
Merged to 9.0.x patch branch.

PR #16682, opened 2020-01-16
fix(@schematics/angular): add `skipLibCheck` to workspace tsconfig

alan-agius4:
Add skipLibCheck to speed up TypeScript analysis and prevent compilation errors caused by incompatible declaration files which were generated with older/new TS versions.

//cc @IgorMinar


alan-agius4:
@filipesilva, @IgorMinar suggested to do the migration in v10

dgp1130:
Merged to 9.0.x patch branch.

dgp1130:
Managed to forget to check for multiple commits while merging to 9.0.x. Just remembered now and merged the others. They might be a little out-of-order on that branch.

PR #16689, opened 2020-01-17
feat(@angular-devkit/benchmark): add capabilities to benchmark watch p…

alan-agius4:
…rocesses

### DIFF
--- a/packages/angular_devkit/benchmark/src/main_spec.ts
+++ b/packages/angular_devkit/benchmark/src/main_spec.ts
@@ -24,7 +24,7 @@ class MockWriteStream {
   }
 }
 
-fdescribe('benchmark binary', () => {
+describe('benchmark binary', () => {
   const benchmarkScript = require.resolve(join(__dirname, './test/fibonacci.js'));
   const exitCodeOneScript = require.resolve(join(__dirname, './test/exit-code-one.js'));
   const benchmarkWatchScript = require.resolve(join(__dirname, './test/watch-test-cmd.js'));
@@ -179,7 +179,7 @@ fdescribe('benchmark binary', () => {
   it('should error when watch-timeout is exceeded', async () => {
     const args = [
       '--watch-timeout',
-      '20',
+      '250',
       '--watch-matcher',
       'Wrong Match',
       '--watch-script',
@@ -191,5 +191,5 @@ fdescribe('benchmark binary', () => {
     const res = await main({ args, stdout, stderr });
     expect(stderr.lines).toContain('[benchmark] Timeout has occurred\n');
     expect(res).toEqual(1);
-  });
+  }, 30000);
 });
--- a/packages/angular_devkit/benchmark/src/main_spec.ts
+++ b/packages/angular_devkit/benchmark/src/main_spec.ts
@@ -8,6 +8,7 @@
 import { existsSync, readFileSync, unlinkSync, writeFileSync } from 'fs';
 import { basename, dirname, join } from 'path';
 import { main } from './main';
+
 // tslint:disable-next-line:no-implicit-dependencies
 const temp = require('temp');
 
@@ -23,9 +24,11 @@ class MockWriteStream {
   }
 }
 
-describe('benchmark binary', () => {
+fdescribe('benchmark binary', () => {
   const benchmarkScript = require.resolve(join(__dirname, './test/fibonacci.js'));
   const exitCodeOneScript = require.resolve(join(__dirname, './test/exit-code-one.js'));
+  const benchmarkWatchScript = require.resolve(join(__dirname, './test/watch-test-cmd.js'));
+  const watchTriggerScript = require.resolve(join(__dirname, './test/watch-test-script.js'));
   const outputFileRoot = temp.mkdirSync('benchmark-binary-spec-');
   const outputFile = join(outputFileRoot, 'output.log');
   let stdout: MockWriteStream, stderr: MockWriteStream;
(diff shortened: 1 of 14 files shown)


packages/angular_devkit/benchmark/src/main_spec.ts
```diff
@@ -142,4 +143,13 @@ describe('benchmark binary', () => {
     stdout.lines.forEach(line => expect(line).toMatch(/^\[abc\]/));
     expect(res).toEqual(0);
   });
+
+  it('uses watch-script and watch-matcher', async () => {
+    const benchmarkScript = require.resolve(join(__dirname, './test/watch-test-cmd.js'));
+    const watchScript = require.resolve(join(__dirname, './test/watch-test-script.js'));
+    const args = ['--watch-matcher', 'Complete', '--watch-script', watchScript, '--', 'node', benchmarkScript];
```
filipesilva: Shouldn't `watchScript` here also be prefixed with `node`? Or is the script always a node script? 

I think it's fine to limit it to a runnable script instead of a command with args though, mainly because handling multiple commands inside one command is hairy business.

packages/angular_devkit/benchmark/src/run-benchmark-watch.ts
```diff
@@ -0,0 +1,115 @@
...
+  const processFailed = new BaseException('Wrong exit code.');
+
+  const stats$ = monitoredProcess.stats$.pipe(
+    takeUntil(monitoredProcess.stdout$.pipe(
+      first(stdout => stdout.toString().includes(watchMatcher)),
+      timeout(watchTimeout),
+    )),
+  );
+
+  return monitoredProcess.run()
```
filipesilva: I feel a bit confused about the logic in this file, I guess because I haven't looked at this package in a while. 

Could you add some comments for readers here when you feel the PR isn't in draft anymore?

packages/angular_devkit/benchmark/src/run-benchmark-watch.ts
```diff
@@ -0,0 +1,115 @@
...
+      startWith(undefined),
+      concatMap(processExitCode => {
+        if (processExitCode !== undefined && processExitCode != command.expectedExitCode) {
+          logger.debug(`${debugPrefix()} exited with ${processExitCode} but `
+            + `${command.expectedExitCode} was expected`);
+
+          return throwError(processFailed);
+        }
+
+        return monitoredProcess.stdout$.pipe(
```
filipesilva: You might want to hook up the process stdout before you start to run it, otherwise you can miss stuff in the delay. I think? Maybe that's not true. 

If it's not true can you add a comment saying why it's ok to start first and then hook up?

packages/angular_devkit/benchmark/src/main_spec.ts
```diff
@@ -142,4 +143,13 @@ describe('benchmark binary', () => {
     stdout.lines.forEach(line => expect(line).toMatch(/^\[abc\]/));
     expect(res).toEqual(0);
   });
+
+  it('uses watch-script and watch-matcher', async () => {
```
filipesilva: Can you add successful and failing retry tests, and timeout tests as well?

Checking the verbose output helps to make sure the right thing is happening.

packages/angular_devkit/benchmark/src/run-benchmark-watch.ts
alan-agius4: You are right

packages/angular_devkit/benchmark/src/run-benchmark-watch.ts
alan-agius4: Done

packages/angular_devkit/benchmark/src/main_spec.ts
alan-agius4: Yes it is always a node script. Unless using the API where you can provide a `Command` object.

alan-agius4:
Need to look further why the timeout test is failing on ci in `test` step but works locally and the `build-bazel` step

packages/angular_devkit/benchmark/src/main_spec.ts
filipesilva: Can you mention on the help usage it's a node script?

packages/angular_devkit/benchmark/src/main_spec.ts
```diff
@@ -142,4 +145,51 @@ describe('benchmark binary', () => {
...
+      '--',
+      'node',
+      benchmarkWatchScript,
+    ];
+    const res = await main({ args, stdout, stderr });
+    expect(stdout.lines).toContain('[benchmark] Process Stats\n');
+    expect(res).toEqual(0);
+  }, 30000);
+
+  it('should not fail with exit code', async () => {
```
filipesilva: Maybe `should not fail with exit code` -> `should fail with exit code 1 when retries are exceeded`?

packages/angular_devkit/benchmark/src/main_spec.ts
```diff
@@ -142,4 +145,51 @@ describe('benchmark binary', () => {
...
+      '--',
+      'node',
+      exitCodeOneScript,
+    ];
+    const res = await main({ args, stdout, stderr });
+    expect(stderr.lines).toContain('[benchmark] Maximum number of retries (5) for command was exceeded.\n');
+    expect(res).toEqual(1);
+  });
+
+  it('should error when watch-timeout is exceeded', async () => {
```
filipesilva: Maybe we should treat timeouts as consuming an individual retry... I guess that depends on how flaky the thing is. On webpack we often have trouble with writes being detected and that hits the timeout, failing the whole thing.

We don't have to do that now though. We can see if this model is appropriate while benchmarking stuff and if it needs to be more robust, we tweak it.

packages/angular_devkit/benchmark/src/main_spec.ts
alan-agius4: Actually, the benchmarker will not return the exit code in the error message. That was actually

packages/angular_devkit/benchmark/src/main_spec.ts
alan-agius4: This is the current implementation. A failed iteration will consume a retry.

alan-agius4: This is the current implementation. A failed iteration will consume a retry.

filipesilva:
Awesome stuff, looking forward to using it for the benchmarks!

Issue #16146, opened 2019-11-11
Increased main.js bundle size but less lazy chunks after updating to angular 9.0.0-rc.1


wambowams:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description

After updating from Angular 8.2.3 to 9.0.0-rc.1 the bundle size was increased for main module and some of the feature modules:
Before it was:
![bundle_before9](https://user-images.githubusercontent.com/7945998/68595724-4d811000-049a-11ea-86db-9eb033f7cdf4.PNG)

After updating to Angular 9.0.0-rc.1 it is
![image](https://user-images.githubusercontent.com/7945998/68595789-730e1980-049a-11ea-90b9-f8f73ebb375d.png)

## 🔬 Minimal Reproduction
running `ng build --prod --namedChunks` with following configuration:
```
            "production": {
              "fileReplacements": [
                {
                  "replace": "src/environments/environment.ts",
                  "with": "src/environments/environment.prod.ts"
                }
              ],
              "optimization": true,
              "outputHashing": "all",
              "sourceMap": false,
              "extractCss": true,
              "namedChunks": false,
              "extractLicenses": true,
              "vendorChunk": false,
              "commonChunk": false,
              "buildOptimizer": true,
              "budgets": [
                {
                  "type": "initial",
                  "maximumWarning": "1mb",
                  "maximumError": "2mb"
                },
                {
                  "type": "anyComponentStyle",
                  "maximumWarning": "6kb"
                }
              ],
              "serviceWorker": true,
              "ngswConfigPath": "ngsw-config.json"
            }
          }
```

## 🌍 Your Environment
<pre><code>
Angular CLI: 9.0.0-rc.1
Node: 10.15.3
OS: win32 x64
Angular: 9.0.0-rc.1
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router, service-worker

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.900.0-rc.1
@angular-devkit/build-angular     0.900.0-rc.1
@angular-devkit/build-optimizer   0.900.0-rc.1
[...]

PR #16694, opened 2020-01-17
Control emitting Ivy class metadata and Angular module scope information

clydin:
This change allows the import eliding capabilities of the plugin to more completely remove the effects of the `setClassMetadata` and `setNgModuleScope` functions.  This provides equivalent behavior to the Decorator removal functionality for View Engine.  As a result, Angular components are no longer retained in chunks that do not use the component.

This corrects the factory code location problem discussed in https://github.com/angular/angular-cli/issues/16146#issuecomment-557559287.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/typescript.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/typescript.ts
@@ -122,10 +122,17 @@ export function getAotConfig(wco: WebpackConfigOptions, i18nExtract = false) {
   }
 
   const test = /(?:\.ngfactory\.js|\.ngstyle\.js|\.tsx?)$/;
+  const optimize = wco.buildOptions.optimization.scripts;
 
   return {
     module: { rules: [{ test, use: loaders }] },
-    plugins: [_createAotPlugin(wco, { tsConfigPath }, i18nExtract)]
+    plugins: [
+      _createAotPlugin(
+        wco,
+        { tsConfigPath, emitClassMetadata: !optimize, emitNgModuleScope: !optimize },
+        i18nExtract,
+      ),
+    ],
   };
 }
 
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -61,6 +61,7 @@ import {
 } from './transformers';
 import { collectDeepNodes } from './transformers/ast_helpers';
 import { downlevelConstructorParameters } from './transformers/ctor-parameters';
+import { removeIvyJitSupportCalls } from './transformers/remove-ivy-jit-support-calls';
 import {
   AUTO_START_ARG,
 } from './type_checker';
@@ -1002,6 +1003,15 @@ export class AngularCompilerPlugin {
         // Remove unneeded angular decorators in VE.
         // In Ivy they are removed in ngc directly.
         this._transformers.push(removeDecorators(isAppPath, getTypeChecker));
+      } else {
+        // Default for both options is to emit (undefined means true)
+        const removeClassMetadata = this._options.emitClassMetadata === false;
+        const removeNgModuleScope = this._options.emitNgModuleScope === false;
+        if (removeClassMetadata || removeNgModuleScope) {
+          this._transformers.push(
+            removeIvyJitSupportCalls(removeClassMetadata, removeNgModuleScope, getTypeChecker),
+          );
+        }
       }
       // Import ngfactory in loadChildren import syntax
       if (this._useFactories) {
--- a/packages/ngtools/webpack/src/interfaces.ts
+++ b/packages/ngtools/webpack/src/interfaces.ts
@@ -45,6 +45,11 @@ export interface AngularCompilerPluginOptions {
   logger?: logging.Logger;
   directTemplateLoading?: boolean;
 
+  /* @internal */
+  emitClassMetadata?: boolean;
+  /* @internal */
+  emitNgModuleScope?: boolean;
+
   /**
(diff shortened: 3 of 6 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/typescript.ts
```diff
@@ -122,10 +122,17 @@ export function getAotConfig(wco: WebpackConfigOptions, i18nExtract = false) {
   }
 
   const test = /(?:\.ngfactory\.js|\.ngstyle\.js|\.tsx?)$/;
+  const optimize = wco.buildOptions.optimization.scripts;
 
   return {
     module: { rules: [{ test, use: loaders }] },
-    plugins: [_createAotPlugin(wco, { tsConfigPath }, i18nExtract)]
+    plugins: [
+      _createAotPlugin(
```
IgorMinar: I assume that this really affects only aot+optimize setups, because we don't want to strip this metadata for jit+optimize builds.

packages/ngtools/webpack/src/transformers/remove-ivy-dev-calls.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as ts from 'typescript';
+import { collectDeepNodes } from './ast_helpers';
+import { RemoveNodeOperation, StandardTransform, TransformOperation } from './interfaces';
+import { makeTransform } from './make_transform';
+
+export function removeClassMetadata(): ts.TransformerFactory<ts.SourceFile> {
```
IgorMinar: there is a ton of duplication in removeClassMetadata and removeNgModuleScope functions. Not only that we are adding unnecessary code to our code base, but we are also making the transform twice as slow because we iterate over all the nodes twice.

I suggest you create a single transformer that will remove both types of nodes. You can still have two separate flags to control the behavior of the transformer, but you don't need two transformers.

IgorMinar:
@filipesilva can you please also review? thanks

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/typescript.ts
filipesilva: This transformer only removes the calls for files emitted by the TS compilation. If it's just a JIT compilation, user files won't have them at all. 

This behaviour begs the question though: are these calls present in js files inside node modules? Should they be removed from there too? What happens to workspace libraries built with Ivy (even though they shouldn't be built with Ivy for production)?

clydin:
Tests and optimizations are the remaining open items for this.  I was planning on finishing this up today.

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/typescript.ts
clydin: The shared application module technique is the prime benefactor of this change.  Since this is a common technique, the change here should provide a benefit for a large segment of users.  For application code, this is essentially identical to the current state of decorator removal in VE.

For libraries, these calls may be present depending on how the library is built. If present, they will still be stripped at the later stages when using the CLI via the existing methods.  Since libraries are generally bundled prior to packaging, this PR wouldn’t apply even if it did process them as there are no imports to remove.  Later stage DCE should be fairly effective in those cases though and is essentially equivalent to the current method of removing downlevel decorators from libraries in VE when using the CLI.

packages/ngtools/webpack/src/transformers/remove-ivy-dev-calls_spec.ts
```diff
@@ -0,0 +1,115 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { tags } from '@angular-devkit/core';  // tslint:disable-line:no-implicit-dependencies
+import { transformTypescript } from './ast_helpers';
+import { removeIvyDevCalls } from './remove-ivy-dev-calls';
+
+const input = tags.stripIndent`
+  export class AppModule {
+  }
```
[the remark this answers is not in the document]
clydin: I copied it from the output of `ngc` run against a newly generated project but can fix

packages/ngtools/webpack/src/transformers/remove-ivy-dev-calls_spec.ts
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { tags } from '@angular-devkit/core';  // tslint:disable-line:no-implicit-dependencies
+import { transformTypescript } from './ast_helpers';
+import { removeIvyDevCalls } from './remove-ivy-dev-calls';
+
+const input = tags.stripIndent`
+  export class AppModule {
```
filipesilva: Are the `i0` imports elided too? I think our code would remove them, but they aren't present in the input so we don't test that here.

I think webpack might remove them? But I'm not 100% sure, because by the time webpack sees it's just a static import. No other tooling sees the `i0` and knows it was being used but isn't needed anymore.

filipesilva: Now that I think about it more, I'm not sure this matters. I **think** other `i0` calls will still remain here.

packages/ngtools/webpack/src/transformers/remove-ivy-dev-calls_spec.ts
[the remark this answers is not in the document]
IgorMinar: interesting. keep it then as is. thanks

PR #16698, opened 2020-01-18
refactor(@angular-devkit/build-angular): remove nonexistent argument from rxjs path mapping call

clydin:
The rxjs path mapping function does not have any parameters and has not since version 5.5.0 which is not supported.


alan-agius4:
Shouldn't this also target patch?

dgp1130:
Merged to 9.0.x patch branch.

Issue #16708, opened 2020-01-19
Regression problem in tsconfig.json outDir that cause VSCode showing errors in all import from libs pointed with paths under /dist/...


dmorosinotto:
# 🐞 Bug report

### Command

- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the previous version in which this bug was not present was: 8.3.23

### Description

The root tsconfig.json used in the new v9.0.0-rc* template point `"outDir": "./dist"`
That caused annoying problem in VSCode when you rebuild libs referred by `"paths"` 
all the code  `imports {...} from lib` is marked as Errors by TS because they are build under `/dist/...`

## 🔬 Minimal Reproduction
Run those commands:
```
npx @angular/cli@next new --createApplication=false my-workspace

cd my-workspace
npm run ng -- generate library my-lib
npm run ng -- generate application my-app
```
The above commands are needed to setup the workspace for the repro with an application and a library pointed in the tsconfig.json ```"paths": { "my-lib": ["dist/my-lib"] }```

then add this line to app.module.ts to use the library inside the application:
```
// app.module.ts
import { MyLibModule } from 'my-lib';
```

and then run the compilation of the my-lib project in watch mode:
```
npm run ng -- build project=my-lib --watch
```

You will see in VSCode that all imports pointing to the my-lib will be signal as TS Errors 2307

## 🔥 Exception or Error
<pre><code>
Can't find module 'my-lib' ts(2307)
</code></pre>

## 🌍 Your Environment
<pre><code>
Angular CLI: 9.0.0-rc.9
Node: 12.14.1
OS: darwin x64

Angular: 9.0.0-rc.9
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router
Ivy Workspace: Yes

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.900.0-rc.9
@angular-devkit/build-angular      0.900.0-rc.9
@angular-devkit/build-ng-packagr   0.900.0-rc.9
@angular-devkit/build-optimizer    0.900.0-rc.9
@angular-devkit/build-webpack      0.900.0-rc.9
@angular-devkit/core               9.0.0-rc.9
[...]

PR #16709, opened 2020-01-19
fix(@schematics/angular): regression tsconfig.json #16708

dmorosinotto:
fix(@schematics/angular): regression tsconfig.json  …
Unverified
0e318ae
Regression in tsconfig.json set `"outDir": "./dist/out-tsc"` for problems in VSCode TS(2307) when building library referred in tsconfig "paths" 
 
Closes: #16708

### DIFF
--- a/packages/schematics/angular/library/index.ts
+++ b/packages/schematics/angular/library/index.ts
@@ -55,7 +55,7 @@ function updateJsonFile<T>(host: Tree, path: string, callback: UpdateJsonFn<T>):
   return host;
 }
 
-function updateTsConfig(packageName: string, distRoot: string) {
+function updateTsConfig(packageName: string, ...paths: string[]) {
 
   return (host: Tree) => {
     if (!host.exists('tsconfig.json')) { return host; }
@@ -67,7 +67,7 @@ function updateTsConfig(packageName: string, distRoot: string) {
       if (!tsconfig.compilerOptions.paths[packageName]) {
         tsconfig.compilerOptions.paths[packageName] = [];
       }
-      tsconfig.compilerOptions.paths[packageName].push(distRoot);
+      tsconfig.compilerOptions.paths[packageName].push(...paths);
     });
   };
 }
@@ -192,6 +192,7 @@ export default function (options: LibraryOptions): Rule {
     const folderName = `${scopeFolder}${strings.dasherize(options.name)}`;
     const projectRoot = join(normalize(newProjectRoot), folderName);
     const distRoot = `dist/${folderName}`;
+    const pathImportLib = `${distRoot}/${folderName.replace('/','-')}`;
     const sourceDir = `${projectRoot}/src/lib`;
 
     const templateSource = apply(url('./files'), [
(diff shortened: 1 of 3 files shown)


alan-agius4:
Thanks for this.

@clydin and myself looked at this and it seems that TS file watcher is not handling deletion well when excludes are involved.

While checking this I also saw some change in behaviour which involved path mappings which seems that auto imports were no longer being suggested.

Would you also mind to fix the following https://github.com/angular/angular-cli/blob/49de0a6f09dc0836ce989b703fbf049a39f8584b/packages/schematics/angular/library/index.ts#L70 so that itgenerates the below paths; 

```ts
    "paths": {
      "foo": [
        "dist/foo/foo",
        "dist/foo"
      ]
    }
```

instead of:
```ts
    "paths": {
      "foo": [
        "dist/foo"
      ]
    }
```

The first path will be used by TSC, while the second will be used by Webpack.

dmorosinotto:
> Would you also mind to fix the following

@alan-agius4 Ok I'll update the code to add the second ref in the paths.
Do you think that it'll be acceptable to pass another param (the folderName) to the updateTSConfig function to do this?

alan-agius4:
@dmorosinotto, no need as the `packageName` is the folder name.

dmorosinotto:
> @dmorosinotto, no need as the `packageName` is the folder name.

Well not exactly especially if you pass a scoped name (like `@myScope/myLib`) due to strings.dasharize + scopeName handling if you use the *packageName* to write something like 
```tsconfig.compilerOptions.paths[packageName].push(distRoot)+'/'+packageName;```

You'll have this output :
```
"paths": {
      "@my-scope/my-lib": [
        "dist/my-scope/my-lib",
        "dist/my-scope/my-lib/@my-scope/my-lib"
      ]
 } 
```
You can try to test it in this [blitz](https://stackblitz.com/edit/test-updatetsconfig)

that I suspect to be not exactly what you want, I think that this may be the correct one:
```
"paths": {
      "@my-scope/my-lib": [
        "dist/my-scope/my-lib",
        "dist/my-scope/my-lib/my-scope-my-lib"
      ]
} 
```

JFYI I tryed to create a library with `ng generate library @myScope/myLib` and then compile it with `ng build @myScoped/myLib` with the current latest CLI 9.0.0.rc-9 and this is the output in /dist I don't know exactly which path Webpack will need?
AFTER ng build @myScope:myLib


@alan-agius4  Can you clarify me the real expected behavior? so I'll code the need changes, thanks!

**Comment by alan-agius4** [comment]
You are right the paths should be without `@` and `-` and any other special symbol.

Also the order of the below is important, as changing the order will result in a different behaviour.
```js
paths": {
      "@my-scope/my-lib": [
        "dist/my-scope/my-lib/my-scope-my-lib", -> used by the IDE to suggest correct imports.
        "dist/my-scope/my-lib" -> used by the apps Webpack 
      ]
} 
```

dmorosinotto:
Sorry @alan-agius4 I've done the code change directly online in GitHub and I've made a lint error (for whitespace) just fixed, but the real problems are the tests broken, can you help me with it?

alan-agius4:
I’ll try to take a look in the coming  days.

dmorosinotto:
@alan-agius4 no problem, I've done some investigation looking and the details log and found the failing test code.
I changed it according to the new behaviour, hoping to have done it right, please [check](https://github.com/dmorosinotto/angular-cli/commit/fc68986c03a39a3c6e3dca9bf66a1df41ef83070) if I've done it right!

PR #16727, opened 2020-01-22
fix(@angular/cli): ensure built-in Angular schematics package can be found

clydin:
The `@angular/cli` package contains a built-in version of the `@schematics/angular` package.  The module resolution paths for the schematics workflow should not assume package hoisting behavior when resolving.  This change corrects that oversight.


dgp1130:
Merged to patch branch 9.0.x.

Issue #16627, opened 2020-01-10
function "statsToString" only prints the size of the first file


EmaGht:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?
No

### Description
I'm CLI-izing an old application and to do so, i've set up the build target with 
`@angular-builders/custom-webpack:browser`
My` webpack.config.js` emits the `main_client.js` together with an `initial.css`,
when this happens, the Angular-CLI function `statsToString `only prints the size of the first file it encounters like so:

```javascript
function statsToString(json, statsConfig) {
    const colors = statsConfig.colors;
    const rs = (x) => colors ? reset(x) : x;
    const w = (x) => colors ? bold(white(x)) : x;
    const changedChunksStats = json.chunks
        .filter((chunk) => chunk.rendered)
        .map((chunk) => {
//@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
//Notice "chunk.files[0]"
        const asset = json.assets.filter((x) => x.name == chunk.files[0])[0];
//@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
        return generateBundleStats({ ...chunk, size: asset && asset.size }, colors);
    });
```

which generates this not-so-good report where the sizes are 73KB for `initial.css` and around 8MB for `main-client.js `

```
.
.
.
chunk {main-client} initial.css, main-client.js (main-client) 73 kB [entry] [rendered]
```
Now, i know i'm not supposed to use my own webpack with the CLI etc.. etc.. etc.. but since the logger actually prints all initial file's names i would assume it would also sum their sizes?

Still, the build is somehow working and I'm already quite happy about that, but if you guys could fix up the size count of the initial chunk, that would be terrific! :)

Thanks!

filipesilva:
That's indeed not great :(

I'm surprised it took this long to come up too!

Would you be interested in taking a look?

EmaGht:
I'll try to take a look at it!
I'm pretty sure i will need a pointer or two when i get to the testing side of the matter but until then... wish me good luck :D

filipesilva:
[...]

PR #16734, opened 2020-01-22
Correctly prints the size of the chunk during the build statistic print

EmaGht:
This commits allows the cli to correctly print the sum of the sizes of the assets of a given chunk instead of the size of the first emitted asset

Attached:
-Before.PNG: What was happening before the fix in my local environment
-After.PNG: What's happening after the fix, notice that now the size of the "main-client" is the actual sum of the sizes of main-client.js and initial.css
-[webpack.config.js](https://pastebin.com/MreT3kuc): As requested, my current webpack configuration. This problem happens when i'm using the [MiniCssExtractPlugin ](https://github.com/webpack-contrib/mini-css-extract-plugin)to extract my initial.css together with the main client

Before.PNG
Before

After.PNG
After

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/stats.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/stats.ts
@@ -64,8 +64,9 @@ export function statsToString(json: any, statsConfig: any) {
   const changedChunksStats = json.chunks
     .filter((chunk: any) => chunk.rendered)
     .map((chunk: any) => {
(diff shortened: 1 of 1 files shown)


EmaGht:
@googlebot I signed it!

clydin:
Can you squash the commits?
Otherwise, I think this looks good.

alan-agius4:
Can you please also reference #16627 in the commit footer?

Ex:
```
Fixes: #16627 
```

EmaGht:
@clydin First time squashing, hopefully everything went according to plan 👍 
@alan-agius4 done!

Issue #16721, opened 2020-01-21
Angular rc9, stops escape unicode


glebmachine:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?
Yes, angular/cli 8.x.x work properly

 Yes, the previous version in which this bug was not present was: ....

### Description
Angular cli with rc9 stops escape cyrillic to unicode characters.
Previously all characters like `Э` becomes `\u042d`

As i understand it's about TerserPLugin `ascii_only` flag
 A clear and concise description of the problem...

## 🔬 Minimal Reproduction
1. create repo
2. add `alert('Э');` anywhere
3. buld
4. check for compile output, it should convert to `alert('\u042d');`

## 🔥 Exception or Error

<pre><code>

</code></pre>

## 🌍 Your Environment
<pre><code>

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 9.0.0-rc.8
Node: 12.1.0
OS: darwin x64

Angular: 9.0.0-rc.9
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router
Ivy Workspace: Yes

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.900.0-rc.5
@angular-devkit/build-angular      0.900.0-rc.9
@angular-devkit/build-ng-packagr   0.900.0-rc.9
@angular-devkit/build-optimizer    0.900.0-rc.9
@angular-devkit/build-webpack      0.900.0-rc.9
@angular-devkit/core               9.0.0-rc.5
@angular-devkit/schematics         8.3.14
@angular/cdk                       9.0.0-rc.7
@angular/cli                       9.0.0-rc.8
@ngtools/webpack                   9.0.0-rc.9
@schematics/angular                8.0.0
[...]

PR #16735, opened 2020-01-22
fix(@angular-devkit/build-angular): ensure output is ASCII only

clydin:
Fixes: #16721

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -404,6 +404,9 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
       safari10: true,
       output: {
         ecma: terserEcma,
+        // For differential loading, this is handled in the bundle processing.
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -404,6 +404,9 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
       safari10: true,
       output: {
         ecma: terserEcma,
+        // For differential loading, this is handled in the bundle processing.
+        // The experimental rollup support breaks without this check.
+        ascii_only: !differentialLoadingMode,
```
alan-agius4: @clydin I am not following how the experimental rollup support is effected by the `differentialLoadingMode` flag

clydin: The experimental rollup E2E test fails if this is always set to true.  Everything else passes without issue and it technically should work if just true.  But the functionality is covered either way.  I just wanted to mention it here in case someone decides to simplify this later.

PR #16738, opened 2020-01-23
test: add snapshot material e2e tests

alan-agius4:


tests/legacy-cli/e2e/tests/build/material.ts
```diff
@@ -0,0 +1,26 @@
+import { getGlobalVariable } from '../../utils/env';
+import { ng, silentNpm } from '../../utils/process';
+import { updateJsonFile } from '../../utils/project';
+
+const snapshots = require('../../ng-snapshot/package.json');
+
+export default async function () {
+    await ng('add', '@angular/material');
```
clydin: This could technically fail if there are breaking changes in the add schematic that affect the snapshot releases.  This should probably not be a problem as long as the add version and the snapshots are only one verson off but is something to watch.

alan-agius4:
I think that makes a lot of sense as we'd also wouldn't need to check if it's a snapshot build in the e2e specs to update and use the snapshot dependency version.

PR #16755, opened 2020-01-24
fix(@schematics/angular): add compliance with call-signature lint rule

### LINKED ISSUE #16754: Schematics compliance with call-signature lint rule
# 🚀 Feature request

### Command
- [x] lint

### Description

It's a very good practice in TypeScript to enable the `typedef: ["call-signature"]` TSLint rule, which ensures that functions return type is explicit.

Why?
- because otherwise TypeScript infers the return type, but based on the code, which assumes the code is correct (while it may not);
- it's the minimal documentation that should be expected in any code.

It would be nice if Angular CLI schematics could be compliant with that lint rule by default. It's currently not the case:
- in component schematic `ngOnInit()`
- in e2e schematics

The first one is especially annoying as then we need to add `: void` each time we create a new component.

That would also be more consistent with other schematics where the functions return type is indeed explicit (like guards, interceptors, etc.).

PR done in #16755. Note that it does not enable the lint rule, it just makes schematics compliant with it, so it has no consequences at all.

cyrilletuzi:
Fixes #16754 

Make Angular CLI schematics compliant with `typedef: ["call-signature"]` TSLint rule.

### DIFF
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
@@ -17,7 +17,7 @@ export class <%= classify(name) %><%= classify(type) %> implements OnInit {
 
   constructor() { }
 
-  ngOnInit() {
+  ngOnInit(): void {
   }
 
 }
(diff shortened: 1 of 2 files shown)


cyrilletuzi:
@clydin Done.

fmalcher:
I know come up with this very late but just stumbled upon this.
I don't really see the value of adding the `void` type to `ngOnInit`. It makes the code less readable and adds no value since the return type doesn't matter at all.

Has then just been added so that we can use this TSLint rule?!
Btw, adding types for the e2e test methods is fine for me since this is just one file.
For component files this changes the old code output for every newly generated component.

Issue #16767, opened 2020-01-27
Angular CLI component generations template for .spec.ts is missing an await


bjarketrux:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?
No

### Description
Angular CLI component generations template for `.spec.ts` is missing an `await`.
The code generated looks like this:

```JavaScript
  beforeEach(async(() => {
    TestBed.configureTestingModule({
      declarations: [ MyComponent ]
    })
    .compileComponents();
  }));
```

The anonymous function was made `async` in this issue: https://github.com/angular/angular-cli/issues/4774

But just making the function `async` will not make it wait for the asynchronous call to `compileComponents()`.

The code should be changed to this (added `await`):
```JavaScript
  beforeEach(async(() => {
    await TestBed.configureTestingModule({
      declarations: [ MyComponent ]
    })
    .compileComponents();
  }));
```

Source code: https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/component/files/__name%40dasherize%40if-flat__/__name%40dasherize__.__type%40dasherize__.spec.ts.template

## 🔬 Minimal Reproduction
Simply generate a new component with tests:
```
ng generate component my-component
```

bjarketrux:
I misread. `async` here is a function call not the `async` operator. Closing

PR #16768, opened 2020-01-27
Added await in front of asynchronious call.

bjarketrux:
See issue #16767

### DIFF
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.spec.ts.template
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.spec.ts.template
@@ -7,7 +7,7 @@ describe('<%= classify(name) %><%= classify(type) %>', () => {
   let fixture: ComponentFixture<<%= classify(name) %><%= classify(type) %>>;
 
   beforeEach(async(() => {
-    TestBed.configureTestingModule({
+    await TestBed.configureTestingModule({
       declarations: [ <%= classify(name) %><%= classify(type) %> ]
     })
     .compileComponents();


bjarketrux:
@googlebot I signed it!

packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.spec.ts.template
```diff
@@ -7,7 +7,7 @@ describe('<%= classify(name) %><%= classify(type) %>', () => {
   let fixture: ComponentFixture<<%= classify(name) %><%= classify(type) %>>;
 
   beforeEach(async(() => {
```
clydin: Can you turn this into an actual async function?  The await won’t work otherwise.  Please also remove the import of “async” as well if it is not used with the file anymore.

bjarketrux: Fark! I misread the async I thought it said:
```JavaScript
  beforeEach(async () => {
    TestBed.configureTestingModule({
      declarations: [ MyComponent ]
    })
    .compileComponents();
  });
```

I guess the async is a function call to some angular testing thingy :(

bjarketrux: Guess it's not a problem then... Will close.

clydin: The async function was used prior to native async/await support in the test framework.
If you are still interested, it would still be a useful change.

PR #16771, opened 2020-01-27
fix(@schematics/angular): add compliance with no-any lint rule

### LINKED ISSUE #16770: Schematics compliance with no-any lint rule
# 🚀 Feature request

### Command
- [x] lint

### Description

It's a very good practice in TypeScript to enable the `no-any` TSLint rule, which ensures that no `any` is used, as it's not type safe. Since [TypeScript 3.0](http://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type), when a value type is really not known, the right type is `unknown`, which is type safe.

It would be nice if Angular CLI schematics could be compliant with that lint rule by default. It's currently not the case:
- in `test.ts` configuration files (application and library)
- in pipe schematic
- in e2e boilerplate

That would also be more consistent with other schematics where unknown values are indeed typed as `unknown` (like in the [interceptor schematic](https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/interceptor/files/__name%40dasherize%40if-flat__/__name%40dasherize__.interceptor.ts.template)).

PR done in #16771. Note that it does not enable the lint rule, it just makes schematics compliant with it, so it has no consequences at all.

cyrilletuzi:
Fixes #16770

Make Angular CLI schematics compliant with `no-any` TSLint rule.

For `test.ts` files, as they are just configuration files for unit tests, not real code, the rule is just disabled.

### DIFF
--- a/packages/schematics/angular/application/files/src/test.ts.template
+++ b/packages/schematics/angular/application/files/src/test.ts.template
@@ -7,7 +7,12 @@ import {
   platformBrowserDynamicTesting
 } from '@angular/platform-browser-dynamic/testing';
 
-declare const require: any;
+declare const require: {
+  context(path: string, deep?: boolean, filter?: RegExp): {
+    keys(): string[];
+    <T>(id: string): T;
+  };
+};
 
 // First, initialize the Angular testing environment.
 getTestBed().initTestEnvironment(
--- a/packages/schematics/angular/e2e/files/src/app.po.ts.template
+++ b/packages/schematics/angular/e2e/files/src/app.po.ts.template
@@ -1,8 +1,8 @@
 import { browser, by, element } from 'protractor';
 
 export class AppPage {
-  navigateTo(): Promise<any> {
-    return browser.get(browser.baseUrl) as Promise<any>;
+  navigateTo(): Promise<unknown> {
+    return browser.get(browser.baseUrl) as Promise<unknown>;
   }
 
   getTitleText(): Promise<string> {
--- a/packages/schematics/angular/library/files/src/test.ts.template
+++ b/packages/schematics/angular/library/files/src/test.ts.template
@@ -8,7 +8,12 @@ import {
   platformBrowserDynamicTesting
 } from '@angular/platform-browser-dynamic/testing';
 
-declare const require: any;
+declare const require: {
+  context(path: string, deep?: boolean, filter?: RegExp): {
+    keys(): string[];
+    <T>(id: string): T;
+  };
+};
 
 // First, initialize the Angular testing environment.
 getTestBed().initTestEnvironment(
--- a/packages/schematics/angular/pipe/files/__name@dasherize@if-flat__/__name@dasherize__.pipe.ts.template
+++ b/packages/schematics/angular/pipe/files/__name@dasherize@if-flat__/__name@dasherize__.pipe.ts.template
@@ -5,7 +5,7 @@ import { Pipe, PipeTransform } from '@angular/core';
 })
 export class <%= classify(name) %>Pipe implements PipeTransform {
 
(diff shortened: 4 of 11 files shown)


cyrilletuzi:
Don't think the ci/circleci: test fail is related to the PR.

packages/schematics/angular/application/files/src/test.ts.template
```diff
@@ -7,6 +7,7 @@ import {
   platformBrowserDynamicTesting
 } from '@angular/platform-browser-dynamic/testing';
 
+// tslint:disable-next-line: no-any
 declare const require: any;
```
clydin: Can you define the type for this instead?  Tslint will be phased out moving forward and the less tslint comments that need to be migrated, the better.

In this case, the Webpack-only `require.context` function is all that is needed: https://webpack.js.org/api/module-methods/#requirecontext

cyrilletuzi: @clydin Done.

cyrilletuzi:
@filipesilva I agree that going back to `any` and disabling the lint rule for this line would be safer, that's why I did that in the first time.

The fact that it's TSLint-specific is not an issue: it's just 1 line in schematics and there will be the same equivalent line once the migration to eslint is done.

An alternative would be `test.ts` to not be part of the lint (it's just a configuration file, so it feels wrong that lint errors are reported for this file).

What do you think?

filipesilva:
@cyrilletuzi sorry I didn't mean to propose leaving it as `any`. I agree that now we have a better way of expressing those types. I'd just like our tests to reflect the user apps. 

I don't quite see the point in excluding `test.ts` from linting though. It's not very hard to make it lint compliant, while having exclusions is something we have to keep in mind. Moreso exclusions about types, that affect not only linting but compilation.

cyrilletuzi:
> Heya, can you change the committed files for the large specs as well please? They are in angular-cli/tests/angular_devkit/.

Done.

cyrilletuzi:
Done.

cyrilletuzi:
Unfortunately one test is missing because it adds `.filter()` before `.map()`.

I cant' find any correct typings:
- telling that `context()`returns `unknown[]` raises because `keys()` return type is `IteratorableIterator` (not `Array`)
- telling that `keys()` returns `unknown[]` raises because `map` doesn't expect the specific signature of context.

I've also tried to find the type in `@types/webpack`, without any success.

The only solution left I see is to manually add `filter()` to the type. Which feels wrong.

cyrilletuzi:
Found. It's in `@types/webpack-env` package. I'll see if I can find a solution with this.

filipesilva:
Ah yes, this was the one I was worried might fail. I'm glad we found this now, it would have been worse to find it later after some project were generated with the new typings!

cyrilletuzi:
Are we sure that `.map(context)` is the right code to do here?

Even when following official `@types/webpack-env` typings, it raises because the param passed to `.map()` doesn't match. And indeed, it feels very strange to pass as `.map()` param the value of the start of the expression (`context`).

filipesilva:
I'm pretty sure it was the right thing to do at the time, but if there's a better way I'm all for it.

cyrilletuzi:
Finally found the right type, from `@types/webpack-env`, but stripped to the minimum used, otherwise it would be huge.

It works, but it would certainly be better to use the full type from `@types/webpack-env`:
```typescript
declare const require: __WebpackModuleApi.RequireFunction;
```
but it would require to include the package somewhere in the CLI. Is that an option?

filipesilva:
Good job stripping the type down! Adding the webpack dep itself would be pretty problematic I think, as we'd have to add a direct dependency to the project to be correct... I think this is the least invasive approach.

PR #16786, opened 2020-01-29
revert skipLibCheck

### LINKED ISSUE #16696: Don't enable `skipLibCheck` in `--strict` mode
# 🚀 Feature request

- [x] config

### Description
I'm quite surprised by this move: #16682, which enables TypeScript `skipLibCheck` by default. From what I know, it's usually considered as a *temporary* workaround for issues that should be correctly resolved. While it may have performances and compatibility advantages, it degrades types strictness (ie. some real errors may not be reported anymore).

It goes in an opposite way than the new `--strict` option in v9 that finally enables strict mode.

I won't argue on your choice to make `skipLibCheck` the default (I suppose it comes from some issues I'm not aware of), but is it possible to *not* enable it when the `--strict` option has been provided, for people who wants to work with strict types?

alan-agius4:
Following a convo with @DanielRosenwasser we decided to revert this change.

More context: https://github.com/angular/angular-cli/issues/16696

Closes #16696


alan-agius4:
//cc @IgorMinar

PR #16822, opened 2020-02-03
fix(@ngtools/webpack): don't elide imports for  type references that …

### LINKED ISSUE #16808: Production build breaks in rc.11 & 12
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: rc.10

### Description

Production build breaks in rc11 and 12, if I disable optimization in `angular-cli.json` it goes away. It builds in rc10.

## 🔬 Minimal Reproduction
I have tried to reproduce it using `ng new` but the build doesn't emit the same metadata (`design:type`)

## 🔥 Exception or Error
image
image
<pre><code>
main-es2015.429909d094036241356a.js:6 Uncaught ReferenceError: Observable is not defined
    at main-es2015.429909d094036241356a.js:6
    at Object.6djU (main-es2015.429909d094036241356a.js:6)
    at i (runtime-es2015.4f2dc2db4ea44dd7357a.js:1)
    at Module.zUnb (main-es2015.429909d094036241356a.js:10)
    at i (runtime-es2015.4f2dc2db4ea44dd7357a.js:1)
    at Object.0 (main-es2015.429909d094036241356a.js:6)
    at i (runtime-es2015.4f2dc2db4ea44dd7357a.js:1)
    at t (runtime-es2015.4f2dc2db4ea44dd7357a.js:1)
    at Array.r [as push] (runtime-es2015.4f2dc2db4ea44dd7357a.js:1)
    at main-es2015.429909d094036241356a.js:1
</code></pre>

If I disable optimization it works, and Observable is imported from webpack imports instead of as a global variable.
image

## 🌍 Your Environment
<pre><code>
Angular CLI: 9.0.0-rc.12
Node: 13.7.0
OS: linux x64

Angular: 9.0.0-rc.11
Ivy Workspace: Yes

@angular-devkit/architect         0.900.0-rc.12
@angular-devkit/build-angular     0.900.0-rc.12
@angular-devkit/build-optimizer   0.900.0-rc.12
@angular-devkit/build-webpack     0.900.0-rc.12
@angular-devkit/core              8.3.24
@angular-devkit/schematics        8.3.24
@angular/cdk                      8.2.3
@angular/cli                      9.0.0-rc.12
@angular/language-service         9.0.0-rc.12
@ngtools/webpack                  9.0.0-rc.12
@schematics/angular               9.0.0-rc.12
@schematics/update                0.900.0-rc.12
rxjs                              6.5.4
typescript                        3.7.5
    
Done in 0.44s.

</code></pre>

**Anything else relevant?**

alan-agius4:
…are needed for decorator metadata

When `emitDecoratorMetadata` is set to true we don't elide type references imports that are used for decorated nodes.

Fixes #16808

Jira: TOOL-1308

### DIFF
--- a/packages/ngtools/webpack/src/transformers/ast_helpers.ts
+++ b/packages/ngtools/webpack/src/transformers/ast_helpers.ts
@@ -50,13 +50,14 @@ export function getLastNode(sourceFile: ts.SourceFile): ts.Node | null {
 // Test transform helpers.
 const basePath = '/project/src/';
 const fileName = basePath + 'test-file.ts';
+const typeScriptLibFiles = loadTypeScriptLibFiles();
 const tsLibFiles = loadTsLibFiles();
 
 export function createTypescriptContext(
   content: string,
   additionalFiles?: Record<string, string>,
   useLibs = false,
-  importHelpers = true,
+  extraCompilerOptions: ts.CompilerOptions = {},
 ) {
   // Set compiler options.
   const compilerOptions: ts.CompilerOptions = {
@@ -68,7 +69,9 @@ export function createTypescriptContext(
     target: ts.ScriptTarget.ESNext,
     skipLibCheck: true,
     sourceMap: false,
-    importHelpers,
+    importHelpers: true,
+    experimentalDecorators: true,
+    ...extraCompilerOptions,
   };
 
   // Create compiler host.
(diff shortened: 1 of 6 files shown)


packages/ngtools/webpack/src/transformers/ast_helpers.ts
```diff
@@ -50,25 +50,26 @@ export function getLastNode(sourceFile: ts.SourceFile): ts.Node | null {
...
   content: string,
   additionalFiles?: Record<string, string>,
   useLibs = false,
-  importHelpers = true,
+  extraCompilerOptions: ts.CompilerOptions = {},
 ) {
   // Set compiler options.
   const compilerOptions: ts.CompilerOptions = {
-    noEmitOnError: useLibs,
+    noEmitOnError: false,
```
filipesilva: Discussed offline, setting this to false means our transforms can be generating code that fails TS compilation, and we wouldn't know from running these tests. If there's some way we can not set this option at all it would be great.

alan-agius4: Should be fixed now. Note: we do have quite a number of tests that use `noEmitOnError: false` because of imports to `@angular/core` etc...

filipesilva: Thank you, we should eventually get to review that... but at least we're not introducing more now.

PR #16830, opened 2020-02-04
[LTS] fix(@angular-devkit/build-angular): provide supported browsers to babel

alan-agius4:
By default the browserslist `configPath` will default to `process.cwd()`. This might return incorrect data in a multi app workspace setup.

https://babeljs.io/docs/en/babel-preset-env#configpath

LTS version of https://github.com/angular/angular-cli/pull/16829


alan-agius4:
Ci failure seems not to be caused by the change.

Issue #16484, opened 2019-12-20
Schematics: rename to a non-existing dir fails on sync to disk


Brocco:
# 🐞 Bug report

### Command (mark with an `x`)
Schematics core

### Is this a regression?
No

### Description
When using `tree.rename` where the destination is a directory that does not exist succeeds using dry-run (and testing) but fails when actually syncing to disk

## 🔬 Minimal Reproduction
I created a repo with the reproduction [here](https://github.com/Brocco/schematic-bug)

## 🔥 Exception or Error
<pre><code>
no such file or directory
</code></pre>
This error is stating that the directory you're attempting to write to does not exist.

## 🌍 Your Environment
important version: @angular-devkit/schematics v8.3.20 

**Anything else relevant?**
The important piece is that it appears to work just fine in memory, the error occurs when trying to synchronize from the virtual file system to the actual file system.

PR #16835, opened 2020-02-04
Schematics: rename to a non-existing dir fails on sync to disk #16484

sacgrover:
Schematics: rename to a non-existing dir fails on sync to disk #16484


sacgrover:
@googlebot I signed it!

PR #16839, opened 2020-02-04
fix(@angular-devkit/core): Rename to a non-existing dir

sacgrover:
Schematics: rename to a non-existing dir fails on sync to disk #16484

### DIFF
--- a/packages/angular_devkit/core/node/host.ts
+++ b/packages/angular_devkit/core/node/host.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import * as fs from 'fs';
+import * as path from 'path';
 import { Observable, concat, from as observableFrom, of, throwError } from 'rxjs';
 import {
   concatMap,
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/core/node/host.ts
```diff
@@ -316,6 +317,9 @@ export class NodeJsSyncHost implements virtualFs.Host<fs.Stats> {
       // TODO: remove this try+catch when issue https://github.com/ReactiveX/rxjs/issues/3740 is
       // fixed.
       try {
+        if (!fs.existsSync(getSystemPath(path.parse(to).dir as Path))) {
```
alan-agius4: The castings here are not correct The `to` and `from` here are devkit paths are they are not interchangeable with FS paths, so you need to get the system path first.

So the logic here should look more like;
```ts
try {
  const toSystemPath = getSystemPath(to);
  if (!fs.existsSync(path.dirname(toSystemPath)) {
      fs.mkdirSync(path.dirname(toSystemPath), { recursive: true });
  }

  fs.renameSync(getSystemPath(from), toSystemPath);
  obs.next();
  obs.complete();
} catch (err) {
  obs.error(err);
}
```

sacgrover:
@alan-agius4 Asked changes done.

alan-agius4:
Can you please squash the commits? Otherwise looks good. Thanks.

sacgrover:
Squashing done. @alan-agius4

sacgrover:
@alan-agius4 when it will be merged? I am asking because I can see lot of new issues and I want to work on them but by mistake I committed in master branch now If I take a branch out of it then it will not be in synced with original and will have the pending merge changes in it.

dgp1130:
Merged to 9.0.x patch branch as well.

We have a caretaking process which merges ready PRs daily, so merging should always be pretty timely after approval and tests pass. [Full docs.](https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#-submitting-a-pull-request-pr)

Issue #16790, opened 2020-01-29
Make the version strategy of caniuse-lite incremental


studioTeaTwo:
# 🚀 Feature request

### Command (mark with an `x`)

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

I manage so that browserslist doesn't have  `isEs5SupportNeeded` to trun off differential loading. But the result is not stable.

`caniuse-lite` has its own data in json file in npm package.  Both `build-angular` and `browserslist` have `caniuse-lite`.
Since dependencies of `browserslist` is incremental like `caniuse-lite "^1.0.30001017"` and `caniuse-lite` of `build-angular` is fixed, the judgment result of `BuildBrowserFeatures` can change even if the source code of the repository is the same.
Modern browsers are updated frequently, so Angular-cli may not be updated in time.

For example:

build-angular@0.803.23 includes "caniuse-lite": "1.0.30001019".
This version don't have the information of android chrome 79.
`last 2 versions` on the `browserslist` can return 79-78.
BuildBrowserFeatures.isFeatureSupported('es6-module') returns false, because featureStatus for android chrome 79 is undefined.

### Describe the solution you'd like

The package.json of `@angular-devkit/build-angular` changes version dependencies of `caniuse-lite` incrementally like "^1.0.0".

### Describe alternatives you've considered

Make the `caniuse-lite` included in the browserlist exactly match the` caniuse-lite` of the build-angular.

1. Point the version of `caniuse-lite` in resolutions.
2. Remove `caniuse-lite` in package.json, use dependiecies of package.json in browserlist.

Or, change the logic of `BuildBrowserFeatures.isFeatureSupported`.
It may be better to find the lowest supported version and judge the version with an inequality sign. If using `caniuse-api`, it can be easily. But impact on polyfill creation, etc. cannot be fully determined.

dgp1130:
Should be reasonable to pin to `^1.0.0`, that way it can update without needing to update Angular.

dgp1130:
Updating this appears to cause some breakages, need to identify where they are coming from to verify that this can be properly unpinned.

In not, then we could update the dependency with each Angular minor release, so users would still get weekly updates. That's more of a fallback we aren't able to unpin this like we want to.

PR #16842, opened 2020-02-05
[8.3.x] Update browserslist & caniuse-lite

clydin:
Fixes red CI on 8.3.x branch.

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -18,7 +18,7 @@
     "autoprefixer": "9.6.1",
     "browserslist": "4.8.6",
     "cacache": "12.0.2",
-    "caniuse-lite": "1.0.30001019",
+    "caniuse-lite": "1.0.30001024",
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/package.json
```diff
@@ -16,9 +16,9 @@
     "@ngtools/webpack": "0.0.0",
     "ajv": "6.10.2",
     "autoprefixer": "9.6.1",
-    "browserslist": "4.8.3",
+    "browserslist": "4.8.6",
```
alan-agius4: Nit: probably we should unpin these as discussed here https://github.com/angular/angular-cli/issues/16790

clydin: I think we should probably only do that on master to prevent any accidental breakage.
Also, `caniuse-lite@1.0.30001025` appears to cause the same failure so it needs to be pinned until the issue is resolved upstream.

alan-agius4: Good point

PR #16846, opened 2020-02-06
fix(@angular/cli): change the url for doc search

### LINKED ISSUE #16840: Wrong request for ng doc -s
When i run ng doc component it works well but when i try to use -s for all searches option it gives wrong value 

image

santoshyadav198613:
Fixes #16840

### DIFF
--- a/packages/angular/cli/commands/doc-impl.ts
+++ b/packages/angular/cli/commands/doc-impl.ts
@@ -41,13 +41,13 @@ export class DocCommand extends Command<DocCommandSchema> {
         /* tslint:disable-next-line:no-implicit-dependencies */
         const currentNgVersion = require('@angular/core').VERSION.major;
         domain = `v${currentNgVersion}.angular.io`;
-      } catch (e) {}
+      } catch (e) { }
     }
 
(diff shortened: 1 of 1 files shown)


dgp1130:
Merged to patch branch 9.0.x as well.

packages/angular/cli/commands/doc-impl.ts
```diff
@@ -41,13 +41,13 @@ export class DocCommand extends Command<DocCommandSchema> {
...
         domain = `v${currentNgVersion}.angular.io`;
-      } catch (e) {}
+      } catch (e) { }
     }
 
     let searchUrl = `https://${domain}/api?query=${options.keyword}`;
 
     if (options.search) {
-      searchUrl = `https://www.google.com/search?q=site%3A${domain}+${options.keyword}`;
+      searchUrl = `https://${domain}/?search=${options.keyword}`;
```
gkalpak: I just realized that for archive version (i.e. older than the current stable), non-docs pages (including the home page) are redirected to `/docs` (losing the query params). So, this URL should be `https://${domain}/docs?search=${options.keyword}` 😅
(My bad for suggesting the wrong URL 😇)

I submitted #16856.

PR #16852, opened 2020-02-06
fix(@angular-devkit/build-angular): resolve webdriver-manager from protractor location

clydin:
The webdriver-manager package is a direct dependency of protractor.  The only guaranteed method to resolve the webdriver-manager package in this case is to resolve it from a base of the protractor location.


dgp1130:
Merged to patch branch 9.0.x as well.

PR #16855, opened 2020-02-06
ci: update browsers tests to test supported browsers only

alan-agius4:
https://next.angular.io/guide/browser-support


tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
```diff
@@ -55,18 +46,9 @@ exports.config = {
...
-      tunnelIdentifier,
-    },
-    {
-      browserName: 'firefox',
-      version: '60',
-      tunnelIdentifier,
-    },
-    {
-      browserName: 'firefox',
-      version: '68',
```
clydin: Latest Firefox ESR version is 68: https://www.mozilla.org/en-US/firefox/68.4.2/releasenotes/

tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
```diff
@@ -19,33 +19,24 @@ exports.config = {
...
-      platform: 'OS X 10.11',
-      version: '9.0',
+      platform: 'OS X 10.13',
+      version: '12.4',
       tunnelIdentifier,
     },
     {
       browserName: 'safari',
-      platform: 'OS X 10.12',
-      version: '10.1',
```
clydin: This was here to ensure that the nomodule polyfill works.

alan-agius4: I asked @IgorMinar if we should drop the polyfill in the first place since it’s an unsupported browser.

IgorMinar: Can we test the nomodules polyfill on ie?

IgorMinar:
We should he testing only against these browsers: https://angular.io/guide/browser-support#browser-support

tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
clydin: The polyfill is only used on Safari 10.1 (iOS Safari 10.3).  That specific version is in the unique situation of supporting module scripts but not the nomodule attribute.  This causes both sets of scripts (ES5 & ES2015) to be executed on page load.  Both the version prior (10.0) and the version after (11.0) do not have this issue.  Since 10.1 is technically out of the support range, the polyfill could be dropped and the polyfill script could be manually included by any application attempting to support 10.1.  (script link: https://gist.github.com/samthor/64b114e4a4f539915a95b91ffd340acc)

tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
alan-agius4: We don't seem to support FireFox ESR https://angular.io/guide/browser-support#browser-support

clydin: I’d consider it technically a latest version of Firefox; making Firefox essentially have two. The ESR latest version is used extensively in corporate environments (intranet applications, etc.) due to its support and stability guarantees.
There have been reported issues in the past concerning support of previous ESR versions.

@IgorMinar thoughts?

IgorMinar: I'm fine adding it, but then we should also document it and add it to the test suite no the framework repo.

tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
IgorMinar: sounds good to me. let's drop it.

tests/legacy-cli/e2e/assets/protractor-saucelabs.conf.js
```diff
@@ -19,33 +19,17 @@ exports.config = {
   multiCapabilities: [
     {
       browserName: 'chrome',
-      version: '41',
+      version: 'latest',
```
IgorMinar: Can we pin this and have a regular reminder to update it?

We should avoid unpinned depends so that ci doesn't go red unexpectedly

alan-agius4: Good point. Updated.

dgp1130:
Are we waiting for a review from @clydin, or is this ready to merge?

dgp1130:

PR #16856, opened 2020-02-06
fix(@angular/cli): fix the URL for doc search

gkalpak:
In #16846, the URL for docs search was updated to point to the corresponding vX.angular.io website (with an appropriate query for the search). However, on the archive versions of the website (i.e. versions older than the current stable), non-docs pages (including the home page) are redirected to `/docs` (losing the query params).

This commit fixes the URL to include `/docs`, so that it is not redirected (and preserve the query params that will trigger the search).


dgp1130:
Merged to 9.0.x patch branch as well.

PR #16894, opened 2020-02-09
feat(@angular-devkit/build-angular): pass "grep" and "invertGrep"

### LINKED ISSUE #13020: ng e2e --grep issue
```
- [x] feature request
```

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [x] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Desired functionality
I would like to tag Protractor tests and run them using tags. Currently `ng e2e` is not supporting --grep option:
`Unknown option: '--grep'`

This feature would solve a problem of running tagged tests and with different tags.

jdgarvey:
# Problem
Jasmine has a useful "tagging" feature that allows specific suites/specs to be targeted by pattern.
Protractor also exposes this via a `grep` option in both the Protractor CLI and Protractor config, as well as an `invertGrep` option that "inverts" whatever pattern was specified in the `grep` option. However, the Angular Protractor Builder currently does not recognize the `grep` or `invertGrep` options, so in order to take advantage of this useful feature, developers had to find a way around the builder.

# Solution
 I added these two options to the Protractor Builder JSON schema, as well as made sure they got passed to the Protractor runner correctly. I also needed to add these options to the top-level Angular CLI schema to satisfy `angular.json`.

# Notes
1. Because the Protractor config expects these options nested under a `jasmineNodeOpts` object, I had to use an additional interface to the one that is generated by `schema.json`. Open to better ways of doing this.
2. I tried to find a way to test my work in the associated test file, but I am not getting enough information from the `run.result` to make any tests of value. Also tried using a logger, but that didn't have any useful information for what I am working on. Open to suggestions here.
3. Fixes #13020

### DIFF
--- a/packages/angular_devkit/build_angular/src/protractor/index.ts
+++ b/packages/angular_devkit/build_angular/src/protractor/index.ts
@@ -17,12 +17,23 @@ import * as url from 'url';
 import { runModuleAsObservableFork } from '../utils';
 import { Schema as ProtractorBuilderOptions } from './schema';
 
+interface JasmineNodeOpts {
+  jasmineNodeOpts: {
+    grep?: string;
+    invertGrep?: boolean;
+  };
+}
+
 function runProtractor(root: string, options: ProtractorBuilderOptions): Promise<BuilderOutput> {
-  const additionalProtractorConfig: Partial<ProtractorBuilderOptions> = {
+  const additionalProtractorConfig: Partial<ProtractorBuilderOptions> & Partial<JasmineNodeOpts> = {
     elementExplorer: options.elementExplorer,
     baseUrl: options.baseUrl,
     specs: options.specs && options.specs.length ? options.specs : undefined,
     suite: options.suite,
+    jasmineNodeOpts: {
+      grep: options.grep,
+      invertGrep: options.invertGrep,
+    },
   };
 
   // TODO: Protractor manages process.exit itself, so this target will allways quit the
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -1646,6 +1646,15 @@
               "type": "string",
               "description": "Dev server target to run tests against."
             },
+            "grep": {
+              "type": "string",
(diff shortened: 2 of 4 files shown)


packages/angular_devkit/build_angular/src/protractor/index.ts
```diff
@@ -17,12 +17,23 @@ import * as url from 'url';
...
+interface JasmineNodeOpts {
+  jasmineNodeOpts: {
+    grep?: string;
+    invertGrep?: boolean;
+  };
+}
+
 function runProtractor(root: string, options: ProtractorBuilderOptions): Promise<BuilderOutput> {
-  const additionalProtractorConfig: Partial<ProtractorBuilderOptions> = {
+  const additionalProtractorConfig: Partial<ProtractorBuilderOptions> & Partial<JasmineNodeOpts> = {
```
filipesilva: I suppose the mistake here is that this was never `Partial<ProtractorBuilderOptions>` to begin with, and was actually a different type, one for protractor itself.

This isn't super pertinent to the PR content itself, it's just a comment.

jdgarvey:
Wow, that is a good idea.
Will give it a shot and let you know if I have any questions.
Thanks!

filipesilva:
Looks great! I think the only thing missing is to make it all a single commit instead of the three commits we have currently. Once that's done it's ready to merge.

filipesilva:
Thanks a bunch! This should be on master soon and then released on 9.1.0-beta. We don't add features to patch releases so it needs to be in the next minor instead.

packages/angular_devkit/build_angular/src/protractor/index.ts
alan-agius4: Nit: `Partial<ProtractorBuilderOptions> & Partial<JasmineNodeOpts>` -> `Partial<ProtractorBuilderOptions & JasmineNodeOpts>`

Issue #16857, opened 2020-02-06
CLI not respecting default schematics


gabrieltaets:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: Angular 8

### A clear and concise description of the problem...
 On Angular 9, even though I have my default schematics set to use .scss and not generate spec files, when using ng generate, the CLI ignores my preferences.
This used to work fine on Angular 8.

![image](https://user-images.githubusercontent.com/22807067/73983346-f9f55580-4914-11ea-8680-1afe5f37abef.png)

![image](https://user-images.githubusercontent.com/22807067/73983492-5789a200-4915-11ea-81e4-a8cb939963a8.png)

## 🔬 Minimal Reproduction
Set a schematics for styleext on angular.json and run ng generate.

## 🌍 Your Environment
<pre><code>
Angular CLI: 9.0.0-rc.10
Node: 10.16.3
OS: linux x64

Angular: 9.0.0-rc.10
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router, upgrade
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.900.0-rc.10
@angular-devkit/build-angular     0.900.0-rc.10
@angular-devkit/build-optimizer   0.900.0-rc.10
@angular-devkit/build-webpack     0.900.0-rc.10
@angular-devkit/core              9.0.0-rc.10
@angular-devkit/schematics        9.0.0-rc.10
@angular/cdk                      8.2.3
@angular/flex-layout              8.0.0-beta.27
@angular/material                 8.2.3
@ngtools/webpack                  9.0.0-rc.10
@schematics/angular               9.0.0-rc.10
@schematics/update                0.900.0-rc.10
rxjs                              6.5.3
typescript                        3.7.5
webpack                           4.41.2
</code></pre>

**Anything else relevant?**

gabrieltaets:
explicitly stating the collection does not seem to help:
[...]

PR #16924, opened 2020-02-11
fix(@schematics/angular): migrate spec and styleext schematic options

clydin:
Fixes: #16857


dgp1130:

Issue #16890, opened 2020-02-07
v9: ng add @angular/localize doesn't fix error


julmot:
# 🐞 bug report

### Affected Package

The issue is caused by package @angular/localize

### Is this a regression?

Yes, the previous version in which this bug was not present was: Angular v8

### Description

After running `ng update @angular/core @angular/cli` on an Angular v8 application the `ng serve` command ran without errors, but I've seen the following error in the browser console

```
Error: It looks like your application or one of its dependencies is using i18n.
Angular 9 introduced a global `$localize()` function that needs to be loaded.
Please run `ng add @angular/localize` from the Angular CLI.
(For non-CLI projects, add `import '@angular/localize/init';` to your polyfills.ts file)
```

as described in the [migration guide](https://angular.io/guide/migration-localize).

Therefore I've executed `ng add @angular/localize` like described. But:

1. This only adds the package @angular/localize to package.json, but doesn't add the import statement

```js
/******************************************************************
 * Load `$localize` - used if i18n tags appear in Angular templates.
 */
import '@angular/localize/init';
```

2. Even when manually adding the above mentioned import statement to the polyfills.ts file the error isn't fixed - even after restarting ng serve. I'm still seeing the issue and don't know to resolve it, since I've followed the migration guide.

## 🔥 Exception or Error

![000072](https://user-images.githubusercontent.com/8310677/74024353-fa87fd80-49a1-11ea-810a-10397db4f08e.png)

## 🌍  Your Environment

**Angular Version before updating:**
<pre><code>
Angular CLI: 8.3.23
Node: 12.13.1
OS: win32 x64
Angular: 8.2.14
... animations, common, compiler, compiler-cli, core, elements
... forms, language-service, platform-browser
... platform-browser-dynamic, router

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.803.23
@angular-devkit/build-angular     0.803.23
@angular-devkit/build-optimizer   0.803.23
@angular-devkit/build-webpack     0.803.23
@angular-devkit/core              8.3.23
@angular-devkit/schematics        8.3.23
@angular/cli                      8.3.23
@ngtools/webpack                  8.3.23
@schematics/angular               8.3.23
@schematics/update                0.803.23
rxjs                              6.5.4
typescript                        3.5.3
webpack                           4.39.2
</code></pre>

**Angular Version after updating:**
<pre><code>
Angular CLI: 9.0.1
Node: 12.13.1
OS: win32 x64

[...]

PR #16925, opened 2020-02-11
fix(@angular-devkit/build-angular): only remove localize polyfill if in AOT mode

clydin:
Fixes: #16890


dgp1130:

PR #16926, opened 2020-02-11
fix(@angular/cli): typo in error message

RiccardoBracci:
This change resolves the error in the console if the nodejs version is not supported showing the actual version instead of "undefined".

> Node.js version *undefined* detected.
> The Angular CLI requires a minimum Node.js version of either v10.13 or v12.0.


RiccardoBracci:
@googlebot I signed it!

dgp1130:

Issue #16935, opened 2020-02-11
Invalid workspace config when sourceLocale is es-419


javiermarinros:
When `sourceLocale` is set to "[es-419](https://github.com/angular/angular/blob/master/packages/common/locales/es-419.ts)", the following error is thrown:

An unhandled exception occurred: Workspace config file cannot be loaded: 
angular.json
Schema validation failed with the following errors:
  Data path ".projects[''].i18n.sourceLocale" should match pattern "^[a-z]{2}(-[a-zA-Z]{2,})?$".
  Data path ".projects[''].i18n.sourceLocale" should be object.
  Data path ".projects[''].i18n.sourceLocale" should match exactly one schema in oneOf.

PR #16936, opened 2020-02-11
fix(@angular/cli): expand locale pattern in workspace schema

clydin:
Fixes: #16935


dgp1130:

Issue #16906, opened 2020-02-10
[v9 regression] "ng build --prod" generates two same chunks on tageting `esnext`


JounQin:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: v8

### Description

 

We are using `babel-loader` to compile `.ts` files after `@ngtools/webpack`, so we're targeting `esnext` in `tsconfig.json`, that works just well.

```sh
Generating ES5 bundles for differential loading...
ES5 bundle generation complete.

chunk {2} runtime-latest.57a370de94fe47c97e9f.js (runtime) 2.65 kB [entry] [rendered]
chunk {2} runtime-latest.57a370de94fe47c97e9f.js (runtime) 2.65 kB [entry] [rendered]
chunk {15} 15-latest.0893519600f97ce410cd.js () 9.05 kB  [rendered]
chunk {15} 15-latest.0893519600f97ce410cd.js () 9.05 kB  [rendered]
chunk {5} 5-latest.f9279ccde3b65923fb53.js () 18.5 kB  [rendered]
chunk {5} 5-latest.f9279ccde3b65923fb53.js () 18.5 kB  [rendered]
chunk {19} 19-latest.f87125c1b6fb06aa05a3.js () 4.36 kB  [rendered]
chunk {19} 19-latest.f87125c1b6fb06aa05a3.js () 4.36 kB  [rendered]
chunk {0} common-latest.a00a44a79bbefc04d1dd.js (common) 19.8 kB  [rendered]
chunk {0} common-latest.a00a44a79bbefc04d1dd.js (common) 19.8 kB  [rendered]
chunk {3} 3-latest.6d1c406c0e8567854203.js () 26.2 kB  [rendered]
chunk {3} 3-latest.6d1c406c0e8567854203.js () 26.2 kB  [rendered]
chunk {21} 21-latest.0aac03da7c98fd91d405.js () 7.97 kB  [rendered]
chunk {21} 21-latest.0aac03da7c98fd91d405.js () 7.97 kB  [rendered]
chunk {8} main-latest.1d3d560137a98192c4b2.js (main) 68.4 kB [initial] [rendered]
[...]

PR #16937, opened 2020-02-11
fix(@angular-devkit/build-angular): generate correct filenames when targeting ESNext

clydin:
Fixes: #16906


dgp1130:

Issue #16934, opened 2020-02-11
Cannot build custom builder that extends the build-angular with v9


yowari:
Hello everyone!

This is my first Angular contribution, sorry in advance for any missing details 😄

# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the previous version in which this bug was not present was: 8.3.24

### Description

When trying to build a custom Angular builder, an error occurs telling there is an incorrect import `import("../../../build_webpack/src")` in `node_modules/@angular-devkit/build-angular/src/extract-i18n/index.d.ts:12:110`.

I believe this can be fixed if the `execute` function explicitly defines the return type (https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/extract-i18n/index.ts#L54), similar error happens to me sometimes when building my Angular library,

Of course, going to the `index.d.ts` file in the `node_modules` and correct import with `import("@angular-devkit/build-webpack")` works.

## 🔬 Minimal Reproduction

https://github.com/yowari/custom-builder-error

```
npm install
npm run build
```

## 🔥 Exception or Error
<pre><code>../../node_modules/@angular-devkit/build-angular/src/extract-i18n/index.d.ts:12:110 - error TS2307: Cannot find module '../../../build_webpack/src'.

12 export declare function execute(options: ExtractI18nBuilderOptions, context: BuilderContext): Promise&lt;import("../../../build_webpack/src").BuildResult&gt;;
</code></pre>

## 🌍 Your Environment
<pre><code>

</code></pre>

**Anything else relevant?**

PR #16938, opened 2020-02-11
fix(@angular-devkit/build-angular): provide explicit return type for extract-i18n builder

clydin:
Fixes: #16934


dgp1130:

PR #16939, opened 2020-02-11
fix(@schematics/angular): add include to type definitions in tsconfig

### LINKED ISSUE #16923: ng update should detect typings files and added it to the appropriate tsconfig file
# 🚀 Feature request

I have a typings.d.ts file that no longer is included in the compilation/build. I was getting errors about not being able to find my ambient types that were in that files. I noticed the tsconfig.app.json was updated to only compile the main.ts and the polyfills. Once I added the typings.d.ts explictly, it worked.

Yes, it's likely not a common scenario. But It might be good to include this scenario in the `ng update` process to v9. 

It would be nice if the ng update process noticed if there was a typings file and added it to the config.

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

alan-agius4:
Fixes #16923

### DIFF
--- a/packages/schematics/angular/migrations/update-9/update-app-tsconfigs.ts
+++ b/packages/schematics/angular/migrations/update-9/update-app-tsconfigs.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { JsonAstObject, logging } from '@angular-devkit/core';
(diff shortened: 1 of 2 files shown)


dgp1130:

PR #16942, opened 2020-02-11
fix(@schematics/angular) remove file extension from zone flags comment

alan-agius4:


alan-agius4:
//cc @mhartington

dgp1130:

PR #16951, opened 2020-02-12
feat(@angular/cli): update command should print error stack in verbose mode

devversion:
Currently, the Angular CLI only prints the erorr message if a migration
fails unexpectedly. This is not helpful because developers running `ng
update`
usually share the error upstream if they run into a bug. Sharing the error
upstream without a stack trace isn't helpful in most of the cases, so there
needs to be a way for developers to retrieve a more verbose error w/ stack trace.

Based on various ng-update issues, developers seem to pass the `--verbose`
flag with the assumption of getting a more _verbose_ error. This change
ensures that this works as expected.

Based on initial discussion over Slack, a better long-term solution
would be to always print the verbose stack trace to a debug/error file.
That way, the update experience is not too verbose by default, but the
intitial errors could be easily shared when an update issue is reported
upstream. This is out of scope for now though because the current goal
is to just provide a way to even retrieve the full error.

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -78,6 +78,7 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
   private async executeSchematic(
     collection: string,
     schematic: string,
+    verbose: boolean,
     options = {},
   ): Promise<{ success: boolean; files: Set<string> }> {
     let error = false;
@@ -144,6 +145,9 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
         this.logger.error('The update failed. See above.');
       } else {
         this.logger.fatal(e.message);
+        if (verbose) {
+          this.logger.fatal(e.stack || e);
+        }
       }
 
       return { success: false, files };
@@ -157,6 +161,7 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
     packageName: string,
     collectionPath: string,
     migrationName: string,
+    verbose: boolean,
     commit?: boolean,
   ): Promise<boolean> {
     const collection = this.workflow.engine.createCollection(collectionPath);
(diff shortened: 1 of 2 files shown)


devversion:
I think this could be also considered a fix. If we agree that this is reasonable for now, then this should be available as soon as possible IMO. Right now, reported issues for `ng update` migrations aren't helpful at all.. There is no good way to to ask folks for better errors (since there is no easy way 😄)

dgp1130:
Discussed in triage meeting, we don't want to obfuscate the `ng update` experience with stack traces which are mostly meaningless to the user. Instead, the better approach would be write these logs to a temporary file, and then when a migration fails point the user to that log file. It also works even if the user does not supply `--verbose`.

We are also planning on refactoring `ng update` in the near-future, so hopefully this is a simple change to make and doesn't require too much effort.

devversion:
@dgp1130 I agree with the general long-term solution. That was something that has already been discussed in Slack before this PR has been sent.

The goal of this change is to temporarily provide _a way_ to get a stack trace until the refactoring of `ng update` happens. I don't think that it obfuscates the `ng update` experience since it's an explicit opt-in through the `--verbose` flag. I'd expect the error to be printed _verbosely_ if I pass in `--verbose` explicitly. It's not a flag users will run by default.

I'm fine with this change being discarded in favor of the larger `ng update` refactor. As said, I 100% agree with the long-term solution, but for now, we *need* a way to get useful errors in `ng update`. What can people do to provide helpful errors when they report `ng update` issues upstream?

devversion:
To clarify: These are not errors reported by migrations to the user. It's about unexpected runtime exceptions that break the migration from running. For those (rare) situations I don't think that we _want_ to provide a non-verbose experience if `--verbose` is explicitly specified. Until the error is reported upstream to the schematic authors, there won't be any complete update experience.

The thing I care about is that `ng update` is a big successor for Angular releases. Without being able to tell where user-reported issues originate from, we can't improve things unless we spend significant time debugging/predicting where the _error message_ originates from.

dgp1130:
I definitely understand the immediate need of accessing the stack trace. To clarify a little bit, I don't think the proper fix of writing to a log file necessarily needs to wait until the upcoming `ng update` refactor. If we can make that happen now, then we would certainly merge that PR.

@alan-agius4 offered to help out and take a look a look at this when he gets some time. Hopefully it isn't too difficult to make this change. If this becomes impractical for whatever reason, then we could explore other temporary options until the ideal solution becomes viable.

devversion:
@dgp1130 Okay. That sounds reasonable. I will keep an eye on it. Thanks.

PR #16965, opened 2020-02-13
closed link tag in index.html

sacgrover:
Fixes angular#16964

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/index-file/augment-index-html.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/index-file/augment-index-html.ts
@@ -244,7 +244,7 @@ export async function augmentIndexHtml(params: AugmentIndexHtmlOptions): Promise
     treeAdapter.appendChild(styleElements, element);
   }
 
-  indexSource.insert(styleInsertionPoint, parse5.serialize(styleElements, { treeAdapter }));
+  indexSource.insert(styleInsertionPoint, parse5.serialize(styleElements, { treeAdapter }).replace('>', '/>'));
 
   // Adjust document locale if specified
   if (typeof params.lang == 'string') {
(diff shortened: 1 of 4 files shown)


sacgrover:
@dgp1130 Should I close this?

dgp1130:
Sorry @sacgrover, looks like this PR fell through the cracks a bit.

@clydin, @alan-agius4, can one of you take a look at this? I'm thinking the fix should probably happen with how the styles are generated rather than a regex replacement afterwards. I'm not quite familiar enough to know exactly where this fix should happen though.

IgorMinar (changes requested):
Hi @sacgrover, thanks for the pr. I think you are trying to fix a real issue but, can we please do this only as an option for xhtml deployments?

Doing this by default (for html) doesn't make sense, confuses people that don't know when the tags must be closed and when they don't have to be, and only bloats the payload very slightly but in the critical path.

I don't know how to implement the opt in, other cli folks should chime in on that, but it would not be too crazy to add --xhtml support to `ng new` and angular.json.

dgp1130:
This PR is blocked on supporting XHTML in FW https://github.com/angular/angular/issues/37779. We'll want to support it there more completely before we add support to the CLI as well, or else we'll just be whack-a-mole-ing bugs filed by users for all the edge cases missed because FW doesn't fully support it.

dgp1130:
Looking at old PRs, and this one seems to have been left in a weird state. It's awkwardly blocked on framework support and still has open questions as to how it would exactly work in tooling. There's also not much momentum or desire to make this happen on the Angular side right now, so leaving this PR hanging is just a disservice to the people who care about it. We're not quite at the right place to be able to accept PRs for XHTML right now.

We'll keep the discussion going at https://github.com/angular/angular-cli/issues/18044 and I'll bring this up in the relevant planning discussions to get a stronger direction and find the right priority for this work. For now, we think the best thing is to close this PR until we have a better idea of what a solution here will look like.

PR #16976, opened 2020-02-14
More bazel tests

filipesilva:


.circleci/config.yml
```diff
@@ -58,6 +58,18 @@ executors:
...
+    parameters:
+      nodeversion:
+        type: string
+        default: *default_nodeversion
+    docker:
+      - image: circleci/node:<< parameters.nodeversion >>-browsers
+    working_directory: ~/ng
+    environment:
+      NPM_CONFIG_PREFIX: ~/.npm-global
+    resource_class: 2xlarge+
```
filipesilva: This PR adds a new job  called `integration` that takes a total of ~57 minutes to run and replaces the current `test-large` and `test-large-ve` jobs. The last `test-large` on master took ~43 minutes total, and `test-large-ve` took 22m. These ran in parallel though, and on a less powerful machine, so it's not an obvious improvement.

I used a `2xlarge+` machine because on https://github.com/angular/angular/pull/33927/files#diff-1d37e48f9ceff6d8030570cd36286a61R297 I saw one being used as well, but I think there's room for exploring improvements.

Things that I should test:
- less powerful machine
- more parallelism at bazel level
- more parallelism at CI job level

filipesilva:
Blocked on https://github.com/angular/angular-cli/pull/17304, which should contain the first half of the work here but in a more manageable review chunk.

scripts/README.md
```diff
@@ -121,17 +121,13 @@ Flags:
 
 Compile and outputs the templates. For now, only README is being built and outputted.
 
-## test
+## bazel:test
```
filipesilva: Should remove `packages/ngtools/webpack/src/type_checker_bootstrap.js` when removing the old test setup as mentioned in https://github.com/angular/angular-cli/pull/17304#discussion_r398091282.

Also should remove the package.json test scripts for it.

.circleci/config.yml
filipesilva: Using `large` executor instead of `2xlarge+` increased the job time from 43 minutes to 63 minutes, ~50% increase in time. But according to https://circleci.com/pricing/, `large` is 20 credits per minute and `2xlarge+` is 100. A 50% increase in time sounds reasonable for a 80% decrease in price.

etc/api/BUILD
```diff
@@ -36,14 +36,16 @@ load("@npm_ts_api_guardian//:index.bzl", "ts_api_guardian_test")
...
-        "^BuilderProgressState$",
+        "^BuilderProgressState$$",
         # @angular-devkit/schematics
-        "^workflow$",
-        "^formats$",
+        "^workflow$$",
+        "^formats$$",
         # @angular-devkit/build-optimizer
-        "^buildOptimizerLoader$",
+        "^buildOptimizerLoader$$",
```
filipesilva, line 48: This replacement might need to be done in `ts_api_guardian` itself, as it already contains one of these replacements for shell compat in https://github.com/angular/angular/blob/08348fc2e8b48d58a30063b0b5b03d21158a315b/tools/ts-api-guardian/index.bzl#L111-L113.

The change seems to be related to https://github.com/bazelbuild/rules_nodejs/commit/5358d56.

filipesilva: Discussed with @gregmagolan, he has a better idea of where to fix this.

gregmagolan:
@googlebot I consent.

gregmagolan:
From slack.

Ok. Figured a few things out.
1. There is a linker bug when running outside of the sandbox on osx & linux so "local" breaks things but the "no-remote-exec" tag is better anyway as then it runs locally in the sandbox
2 The "workspaces" in the root package.json breaks the linker since it creates symlinks under the root node_modules to the source code (which is why I was seeing index.ts instead of index.js)

filipesilva:
Caretaker: the required CI jobs need to change when merging this PR. 

The following checks need to be added as required:
- `ci/circleci: integration`

The following checks need to be removed:
- `ci/circleci: build-bazel`
- `ci/circleci: test-large`
- `ci/circleci: test-large-ve`

These three were all folded into `ci/circleci: test` and run with Bazel.

alan-agius4:
@filipesilva can you also please update https://github.com/angular/angular-cli/blob/master/.github/angular-robot.yml#L43-L58 with the new CI tasks.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index_spec.ts
```diff
@@ -41,7 +39,9 @@ describe('Dev Server Builder', () => {
   }
 
   beforeEach(async () => {
-    await createArchitect(path.join(devkitRoot, 'tests/angular_devkit/build_webpack/basic-app/'));
+    const ngJsonPath = path.join(path.dirname(__filename), '../../test/basic-app/angular.json');
```
alan-agius4, line 42: Nit: `path.dirname(__filename)` -> `__dirname`

.bazelrc
```diff
@@ -125,6 +125,10 @@ build:remote --bes_results_url="https://source.cloud.google.com/results/invocati
  # Set remote caching settings
 build:remote --remote_accept_cached=true
 
+# Force remote exeuctions to consider the entire run as linux
```
gregmagolan, line 128: nit: this is copied from angular/angular and is required for OSX cross-platform RBE. the comment could use an explanation of why it is there: "required for OSX cross-platform RBE".

docs/process/bazel.md
```diff
@@ -42,6 +42,24 @@ Since the requirement is quite lax on Linux but quite strict on windows, what en
 happening is that lack of `require.resolve` calls go unnoticed until someone tries to run
 things on Windows, at which point it breaks.
 
+## Debugging jasmine_node_test
+
+On Linux, Bazel tests will run under a sandbox for isolation.
+You can turn off this sandbox by adding the [`local = True`](https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-tests) attribute to the rule.
```
gregmagolan, line 48: NB: for a few tests, sandbox is required as otherwise the rules_nodejs linker symlinks will conflict with the yarn workspace symlinks in node_modules

package.json
```diff
@@ -66,27 +66,47 @@
...
+    "@angular/platform-browser-dynamic": "9.1.2",
+    "@angular/platform-server": "9.1.2",
+    "@angular/router": "9.1.2",
+    "@angular/service-worker": "9.1.2",
+    "@babel/core": "7.9.0",
+    "@babel/generator": "7.9.4",
+    "@babel/preset-env": "7.9.0",
+    "@babel/template": "7.8.6",
+    "@jsdevtools/coverage-istanbul-loader": "3.0.3",
     "@bazel/bazel": "2.1.0",
```
gregmagolan, line 88: nit: switch the bazelisk?

gregmagolan: latest version has a `bazel` bin entry as an alias so `bazel` can still be used in package.json scripts

packages/angular_devkit/architect_cli/BUILD.bazel
```diff
@@ -13,6 +13,8 @@ ts_library(
     srcs = [
         "bin/architect.ts",
     ] + glob(["src/**/*.ts"]),
+    devmode_module = "commonjs",
```
gregmagolan, line 16: If `devmode_module` and `devmode_target` are set on all `ts_library` then it would be clearer to have a `/tools/defaults.bzl` which re-exports ts_library as a macro with these set. See angular/angular.

integration/angular_cli/browserslist
```diff
@@ -0,0 +1,12 @@
+# This file is used by the build system to adjust CSS and JS output to support the specified browsers below.
```
alan-agius4, line 1: Previously this contained only `last 2 Chrome versions` so that during "tests large" we don't run differential loading for each browser test.

packages/angular_devkit/build_webpack/src/webpack-dev-server/index_spec.ts
filipesilva: Fixed. Think I left that in because at some point I was resolving the data files on windows, but these tests don't run on windows right now so that's just noise.

.bazelrc
filipesilva: Added

integration/angular_cli/browserslist
filipesilva: `packages/angular_devkit/build_angular/test/hello-world-app/.browserslistrc` still contains `last 2 Chrome versions`, and that's the one that browser tests use.

docs/process/bazel.md
filipesilva: Added note.

package.json
filipesilva: Switched over.

packages/angular_devkit/architect_cli/BUILD.bazel
filipesilva: Added macro

.bazelrc
```diff
@@ -125,6 +125,11 @@ build:remote --bes_results_url="https://source.cloud.google.com/results/invocati
  # Set remote caching settings
 build:remote --remote_accept_cached=true
 
+# Force remote exeuctions to consider the entire run as linux.
```
clydin, line 128:
```suggestion
# Force remote executions to consider the entire run as linux.
```

tslint.json
```diff
@@ -59,7 +59,11 @@
     "no-string-throw": true,
     "encoding": true,
     "no-floating-promises": true,
-    "no-implicit-dependencies": true,
+    "no-implicit-dependencies": [
+      true,
+      // Specs can depend on the package they are testing.
+      ["@angular-devkit/build-angular"]
```
clydin, line 65: This applies to more than just specs.  Can you add a note that this is a workaround?  I think the ideal solution would be to have a separate tslint ruleset for spec files.  There are some other rules that are commonly disabled in spec files as well (big functions, etc.).

filipesilva: Replaced comment with:
```
      // Specs can depend on the package they are testing.
      // This is a workaround, we should instead have a dedicated
      // tslint config only for specs.
```

.bazelrc
filipesilva: Done

.circleci/config.yml
```diff
@@ -353,19 +328,10 @@ workflows:
             branches:
               ignore:
               - /docs-preview/
-      - build-bazel:
-          requires:
-            - build
       - test:
           requires:
             - build
```
gregmagolan, line 333: I believe `test` can depend only on `setup` since it doesn't use any of the legacy artifacts

.circleci/config.yml
```diff
@@ -353,19 +328,10 @@ workflows:
...
-      - test-large:
-          requires:
-            - build
-      - test-large:
-          name: test-large-ve
-          ve: true
-          glob: "packages/angular_devkit/@(build_angular|build_ng_packagr)/@(src|test)/@(build|browser)/*_spec_large.ts"
+      - integration:
           requires:
             - build
```
gregmagolan, line 336: Same for integration?

.circleci/config.yml
filipesilva: The legacy artifacts aren't used, but the build job is very quick to run (~35s) and will catch any build errors before proceeding to the more lengthy and resource intensive Bazel jobs. 

If they were to run in parallel and there was a build error, all three would just show the same build errors but take longer to fail and use up more CircleCI credits.

.circleci/config.yml
filipesilva: I'm following the same logic as above, since the integration test also builds sources.

etc/api/BUILD
gregmagolan: I believe I landed the fix for this in 1.6.0

.circleci/config.yml
gregmagolan: Makes sense to me. Perhaps leave a comment that this is intentional?

filipesilva: SGTM, added.

dgp1130:
Per https://github.com/angular/angular-cli/pull/16976#issuecomment-617075895 merged despite the pending statuses for removed checks.

Also updated status checks to drop `test-large`, `test-large-ve`, and `build-bazel` and added `integration` as suggested. I'll rebase existing PRs and verify they pick up the new settings.

gregmagolan:
🎉 Nice work @filipesilva

filipesilva:
@gregmagolan you know a lot of it was you :D

Issue #16320, opened 2019-11-29
Schematic with multi select x-promt errors when using command line parameter:  'path ".foo" should be array'


d-koppenhagen:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [x] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

No, it's not. Multi select prompt is just supported within angular cli > `9.0.0-rc.3`.

### Description

Creating an angular schematic that prompts the user for multiple options is possible since https://github.com/angular/angular-cli/issues/16104. The prompt works quite good now but using the command line for multi selections will cause an error:

```text
Error: Schematic input does not validate against the Schema: {"name":"paul"}
Errors:

  Data path ".name" should be array.
...
```

At first I opened an issue for the docs because I thought it's just not documented: https://github.com/angular/angular/issues/33851 But after checking the implementation of the angular guard schematic which has `--implements` parameter (which works) and creating the minimal reproduction repo (see underneath), I think it's a bug somewhere in `@angular-devkit/schematics-cli`. 

## 🔬 Minimal Reproduction
I created a very minimal repo for reproduction:
https://github.com/d-koppenhagen/ngx-multi-select-schematic

```bash
npm i -g @angular-devkit/schematics-cli@next
git clone git@github.com:d-koppenhagen/ngx-multi-select-schematic.git
cd ngx-multi-select-schematic
npm i
```

**✅ Using prompt works good:**
```bash
schematics .:hello --debug=false
? Which interfaces would you like to implement? (Press <space> to select, <a> to toggle all, <
i> to invert selection)
❯◉ hans
 ◉ peter
 ◉ paul
schematics .:hello --debug=false --name="paul, peter"
```

**❌Using command line parameter fails (event not with just a single name handed over:**
```text
schematics .:hello --debug=false --name="paul"
Error: Schematic input does not validate against the Schema: {"name":"paul"}
Errors:

  Data path ".name" should be array.
    at MapSubscriber.registry.compile.pipe.operators_1.map.result [as project] (/usr/local/lib/node_modules/@angular-devkit/schematics-cli/node_modules/@angular-devkit/schematics/tools/schema-option-transform.js:31:27)
[...]

PR #16985, opened 2020-02-15
fix(@angular/cli): correctly split array options

d-koppenhagen:
This PR will fix https://github.com/angular/angular-cli/issues/16320


splincode:
@alan-agius4 is it still active?

alan-agius4:
@splincode no.

@d-koppenhagen, thanks for your contribution.

We've been evaluating an alternative parsers for a number of months and currently we'd like to avoid changing the semantics of how the parser threats values as this can eventually lead to breaking changes depending on which parser we choose to go with.

PR #16988, opened 2020-02-16
fix(@angular/cli): set inlineTemplate and inlineStyle for minimal projects

chrisguttandin:
I noticed that when creating a project with the `minimal` flag like this `npx @angular/cli new my-demo --minimal true` it will create an `app.component.ts` with inline styles and inline templates. However when creating another component with `ng generate component another-component` it will have an external template and external styles.

This pull request tries to fix this inconsistency by setting the `inlineTemplate` and `inlineStyle` options when creating a `minimal` project to make sure all generated components will have inlined styles and templates.

Please let me know if there is anything that needs to be changed.

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -151,14 +151,15 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
 
   const schematics: JsonObject = {};
 
-  if (options.inlineTemplate === true
-    || options.inlineStyle === true
+  if (options.inlineTemplate
+    || options.inlineStyle
+    || options.minimal
     || options.style !== Style.Css) {
     const componentSchematicsOptions: JsonObject = {};
-    if (options.inlineTemplate === true) {
+    if (options.inlineTemplate || options.minimal) {
       componentSchematicsOptions.inlineTemplate = true;
     }
-    if (options.inlineStyle === true) {
+    if (options.inlineStyle || options.minimal) {
       componentSchematicsOptions.inlineStyle = true;
     }
     if (options.style && options.style !== Style.Css) {
--- a/packages/schematics/angular/application/index_spec.ts
+++ b/packages/schematics/angular/application/index_spec.ts
@@ -184,6 +184,19 @@ describe('Application Schematic', () => {
     expect(architect.e2e).not.toBeDefined();
   });
 
+  it('minimal=true should configure the schematics options for components', async () => {
+    const options = { ...defaultOptions, minimal: true };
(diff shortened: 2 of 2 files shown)


packages/schematics/angular/application/index.ts
```diff
@@ -155,10 +155,10 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
     || options.inlineStyle === true
```
alan-agius4: Shouldn't the minimal check be added here as well?

Can you add a Unit tests please?

packages/schematics/angular/application/index.ts
```diff
@@ -155,10 +155,10 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
...
     || options.style !== Style.Css) {
     const componentSchematicsOptions: JsonObject = {};
-    if (options.inlineTemplate === true) {
+    if (options.inlineTemplate === true || options.minimal) {
```
alan-agius4: Nit: ` === true ` can be removed from here.

alan-agius4: Nit: ` === true ` can be removed from here.

chrisguttandin:
Hi @alan-agius4, thanks for reviewing this. I addressed your feedback. Should I also add a test?

packages/schematics/angular/application/index.ts
chrisguttandin: Yes, sorry for overlooking this. Otherwise it doesn't make sense to check for `minimal` inside the nested if statements.

chrisguttandin:
Sorry, I @alan-agius4, I overlooked your first comment in which you asked me to add a test. Can you please point me at an example test which tests a similar thing?

alan-agius4:
@chrisguttandin, Yes please add a test. You can use this https://github.com/angular/angular-cli/blob/1dbcd8dccd05de377d02bd75ba2464434a2e7212/packages/schematics/angular/application/index_spec.ts#L176 as an example

packages/schematics/angular/application/index.ts
```diff
@@ -153,12 +153,13 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
 
   if (options.inlineTemplate === true
```
alan-agius4: Nit: can also remove `=== true` from here 😉

chrisguttandin: Okay, I changed that as well. I hesitated to do so since I didn't want to change the existing code.

chrisguttandin:
Thanks for the pointer. I added a test. Please let me know if there is anything else that needs to be changed.

PR #17003, opened 2020-02-17
fix(@schematics/angular): scroll issue on new app after adding Material

sacgrover:
Starter app is janky after adding Angular Material

Fixes angular#16966

### DIFF
--- a/packages/schematics/angular/application/other-files/app.component.html.template
+++ b/packages/schematics/angular/application/other-files/app.component.html.template
@@ -35,8 +35,11 @@
   }
 
   .toolbar {
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
     height: 60px;
-    margin: -8px;
     display: flex;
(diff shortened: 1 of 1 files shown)


alan-agius4:
@sacgrover, thanks for this! However, I don't think it's wise to use a deprecated API.  There are plain CSS alternatives to fix this issue, such as using an absolute position.

I think updating `toolbar` and `content` css classes to the below should suffice to address the linked issue.

```css
.toolbar {
    position: absolute;
    top: 0;
    left: 0;
    right: 0;
    height: 60px;
    display: flex;
    align-items: center;
    background-color: #1976d2;
    color: white;
    font-weight: 600;
}

.content {
    display: flex;
    margin: 82px auto;
    padding: 0 16px;
    max-width: 960px;
    flex-direction: column;
    align-items: center;
}
```

sacgrover:
@alan-agius4  changes done

PR #17027, opened 2020-02-20
fix(@angular-devkit/build-angular): make parallel a build option

Zwartpet:
As suggested in https://github.com/angular/angular-cli/issues/16860#issuecomment-588244896 I've started a change to make parallel an option to be able to configure Terser to run in containers and CircleCi.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -12,7 +12,6 @@ import {
 import { tags } from '@angular-devkit/core';
 import * as CopyWebpackPlugin from 'copy-webpack-plugin';
 import { existsSync } from 'fs';
-import { cpus } from 'os';
 import * as path from 'path';
 import { RollupOptions } from 'rollup';
 import { ScriptTarget } from 'typescript';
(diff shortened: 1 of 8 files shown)


Zwartpet:
@googlebot I signed it!

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -432,7 +432,7 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
     // Some environments, like CircleCI, report a large number of CPUs but trying to use them
     // Will cause `Error: Call retries were exceeded` errors.
     // https://github.com/webpack-contrib/terser-webpack-plugin/issues/143
-    const maxCpus = Math.min(cpus().length, 7);
+    const maxCpus = (buildOptions.parallel === undefined) ? Math.min(cpus().length, 7) : buildOptions.parallel;
```
beskit: Could you please use `Math.max(1, Math.min(cpus().length, 7) - 1)` to avoid freezing when using all CPU?

Before [commit](https://github.com/angular/angular-cli/commit/fd133138e2f6e1248c3ce5e523adb4f378861179 ) here was "true". According to [docs](https://webpack.js.org/plugins/terser-webpack-plugin/#parallel) default is `os.cpus().length - 1`. 

Thanks you!

Zwartpet: Updated

dgp1130:
After discussion in the tooling team, the main bug in #16860 is running out of memory. Adding a cap on worker count should address that particular problem. While we could add a separate option like you are suggesting, it doesn't have much need beyond preventing OoM errors. The cap was already implemented for Terser, but https://github.com/angular/angular-cli/pull/17028 should apply it to Jest as well. If you're still seeing memory issues after that is merged and released, then we can reevaluate.

Zwartpet:
I'll wait for the merge and release. Closing this PR as it would rather become an env variable then a build option.

Issue #17029, opened 2020-02-20
Using URL with protocol as baseHref during build removes one slash while using the --localize option


PierreDuc:
# 🐞 Bug report

### Command (mark with an `x`)
- [ ] new
- [x] build
- [ ] serve
- [ ] e2e
- [ ] test
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

The localize option was not available yet

### Description

Adding a full URL (https://path.to.assets.app.com) as baseHref to the build command in combination with the `--localize` removes one slash, resulting the base href in the `index.html` to be:

```
<base href="https:/path.to.app.com/{language_code}">
```
This will fail, because the assets will be loaded from a relative url, with this base href appended to it, which is not very nice

## 🔬 Minimal Reproduction

Create an app with localization configured, and run the build command:

```
ng build --baseHref=https://path.to.assets.app.com --prod --localize
```

## 🌍 Your Environment
<pre><code>
     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 9.0.2
Node: 12.13.0
OS: win32 x64

Angular: 9.0.1
... animations, common, compiler, compiler-cli, core, elements
... forms, language-service, platform-browser
... platform-browser-dynamic, router, service-worker
Ivy Workspace: Yes

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.900.2
@angular-devkit/build-angular      0.900.2
@angular-devkit/build-webpack      0.900.2
@angular-devkit/core               9.0.2
@angular-devkit/schematics         9.0.2
@angular/cdk                       9.0.0
@angular/cli                       9.0.2
@angular/fire                      6.0.0-rc.1
@angular/flex-layout               9.0.0-beta.29
@angular/localize                  9.0.2
@angular/material                  9.0.0
@angular/material-moment-adapter   9.0.0
@ngtools/webpack                   9.0.2
@schematics/angular                9.0.2
@schematics/update                 0.900.2
[...]

PR #17030, opened 2020-02-20
fix(@angular-devkit/build-angular): baseHref with protocol and localize option

alan-agius4:
`posix.join` will dedupe double forward slashes resulting in incorrect protocol.

Closes: #17029


dgp1130:

PR #16891, opened 2020-02-08
feat(@schematics/angular): Allow generation of suffix-free files

sacgrover:
Allow generation of suffix-free file and symbol names.
Fix #16811

Issue #16811, opened 2020-01-31
Allow generation of suffix-free file and symbol names


mgechev:
# 🚀 Feature request

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description
 A clear and concise description of the problem or missing capability...

During style guide meetings we discussed that although recommended for most projects, suffixes in names of components, directives, and services in particular could be redundant for some projects.

We'd not change the official style guide recommendation, but we'd want to enable folks with advanced use cases to drop the suffixes when needed.

### Describe the solution you'd like
 If you have a solution in mind, please describe it.

This proposal is for a global CLI flag per project which disables the current default behavior. When this flag is enabled it'll not suffix any of the generated files, nor the declared symbols inside of them.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

Drop suffixes completely, which is bad for most projects.

sacgrover:
@mgechev I have started working on it.

alan-agius4:
This has been addressed via https://github.com/angular/angular-cli/pull/17040

PR #17040, opened 2020-02-21
fix(@schematics/angular): Allow empty string in the type option

sacgrover:
Currently, Component and Class have the options to add custom type. In the case of class, It's already working fine with  an empty string in type but in the case  of component When setting the type to an empty string the file names generated will contain an extra period (.) which breaks the flow.

With this PR, It will generate the files without an extra period (.)

Reference #16811 and #16891

### DIFF
--- a/packages/schematics/angular/component/index.ts
+++ b/packages/schematics/angular/component/index.ts
@@ -7,6 +7,7 @@
  */
 import { strings } from '@angular-devkit/core';
 import {
+  FileOperator,
   Rule,
   SchematicsException,
   Tree,
@@ -14,6 +15,7 @@ import {
   applyTemplates,
   chain,
   filter,
+  forEach,
   mergeWith,
   move,
   noop,
@@ -49,7 +51,7 @@ function addDeclarationToNgModule(options: ComponentOptions): Rule {
       return host;
     }
 
-    options.type = !!options.type ? options.type : 'Component';
+    options.type = options.type != null ? options.type : 'Component';
 
     const modulePath = options.module;
     const source = readIntoSourceFile(host, modulePath);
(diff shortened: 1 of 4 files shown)


sacgrover:
@alan-agius4 / @clydin
Any advice on ci/circleci: build-bazel & ci/circleci: test-win FAIL?

packages/angular_devkit/schematics/src/index.ts
```diff
@@ -22,6 +22,7 @@ export * from './rules/random';
 export * from './rules/schematic';
 export * from './rules/template';
 export * from './rules/url';
+export * from './rules/rename';
```
clydin: This change adds an element to the public API surface.  This should be avoided if possible since each addition adds to the long term sustainment cost.

This is also why the  bazel build CI is failing.  It has a public API check.  New public API baselines need to be accepted before the check passes.  This helps ensure that new additions were not accidental.

packages/schematics/angular/component/index.ts
```diff
@@ -49,15 +50,15 @@ function addDeclarationToNgModule(options: ComponentOptions): Rule {
       return host;
     }
 
-    options.type = !!options.type ? options.type : 'Component';
+    options.type = options.type != null ? options.type : 'Component';
 
     const modulePath = options.module;
     const source = readIntoSourceFile(host, modulePath);
 
     const componentPath = `/${options.path}/`
```
clydin: Since this line is already calculating the file name base, I think we should consider splitting this into two pieces and using them on the templates.  “componentPath” can be the first two parts of the current.  And “componentFile” can be the remainder as the filename part.  This second part can then be used has the base for all file elements in the templates.
This would remove the need to fix the broken file names with a rename as well as optimize the name creation so that it doesn’t have to happen in multiple places as it does currently.

sacgrover: This componentPath we are using to add component in appModule. We are not using it for template currently.

packages/angular_devkit/schematics/src/index.ts
sacgrover: Instead of rename now I used forEach that was already available in public api

sacgrover:
Commit message changed.

dgp1130:

Fixed a minor merge conflict with the `styles` property due to the lack of https://github.com/angular/angular-cli/commit/d6fa2bde93c114e404eb5a38f9bf0c915a52fce6 in the release branch.

PR #17047, opened 2020-02-21
refactor(@angular/cli): Update error message when multiple NgModules to auto add a component to are found.

AlabasterAxe:
The prior message makes it sound like the only option is to skip the auto import. I've just updated it to let people know that there is another way to get the auto module import goodness.

Example of the new message in a dummy repository:
example_message


splincode:
@clydin is it still active?

alan-agius4:
Closing as this should be solved. Feel free to open a new PR if not.

Thanks.

PR #17050, opened 2020-02-21
build: exclude redundant cherry-picked commits when generating changelog

clydin:
If generating a changelog between tags that are on different branches (e.g., 9.0.3 on 9.0.x and 9.1.0-next.0 on master), commits that were cherry-picked and present in the previous version would also show in the newer version's changelog.  This update analyzes the commits and excludes any that fit this scenario.  Any commits that had conflicts will not be able to be matched authoritatively.  Manual review of the generated changelog may still be needed for attempted cherry-pick commits that had conflicts.

### DIFF
--- a/scripts/changelog.ts
+++ b/scripts/changelog.ts
@@ -8,6 +8,7 @@
 // tslint:disable:no-console
 // tslint:disable:no-implicit-dependencies
 import { JsonObject, logging } from '@angular-devkit/core';
+import { execSync } from 'child_process';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
@@ -29,7 +30,16 @@ export interface ChangelogOptions {
   stdout?: boolean;
 }
 
-export default function(args: ChangelogOptions, logger: logging.Logger) {
+function exec(command: string, input?: string): string {
+  return execSync(command, {
+    encoding: 'utf8',
+    stdio: 'pipe',
+    input,
+    maxBuffer: 10 * 1024 * 1024,
+  }).trim();
+}
+
+export default async function(args: ChangelogOptions, logger: logging.Logger) {
   const commits: JsonObject[] = [];
   let toSha: string | null = null;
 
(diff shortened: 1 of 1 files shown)


scripts/changelog.ts
```diff
@@ -39,6 +49,35 @@ export default function(args: ChangelogOptions, logger: logging.Logger) {
...
+    return;
+  }
+
+  // Collect patch identifiers for cherry-pick exclusion
+  const cherryPicked = new Set<string>();
+  const patchIds = new Map<string, string>();
+  const hashes = exec(`git rev-list ${from}...${to}`).split(/\s+/);
+  for (const hash of hashes) {
+    const [patchId] = exec('git patch-id', exec('git show ' + hash)).split(/\s+/);
+    const existing = patchIds.get(patchId);
```
ft-sagar: fdgzc

dgp1130:
Typically when I cherry-pick commits I include the `-x` flag, which leaves a reference in the commit to the original hash. Would it be possible to use this to identify cherry picks which were modified due to merge conflicts?

Assuming caretakers always add `-x` (which we could probably enforce somehow on release branches) and always cherry pick from master (which itself should include the "original" commit, not another cherry-pick), then reading that value would always get the authoritative commit hash. That hash should be stable, even if a merge conflict was fixed.

Could we compare these "authoritative" hashes to identify the new commits and generate a changelog from that set?

clydin:
The `-x` option only applies to cherry-picks without conflicts.  There's also no guarantee that the option will always be used so the patch-id method that compares the diffs of the commit would still need to be used.
The option should definitely be used when possible though since it provides an easy way for a user to relate the commits when reading through the history.

dgp1130:
> The -x option only applies to cherry-picks without conflicts.

Not true, I just had this happen while merging https://github.com/angular/angular-cli/pull/17040. I did a `git cherry-pick -x` and hit a merge conflict. After fixing it, I ran `git cherry-pick --confinue` which generated a commit message including the cherry pick line: https://github.com/angular/angular-cli/commit/35ad92c5a69752e8da025034fe3acef0a69c1812.

I would imagine there's a way to put a restriction on release branches that they must contain the `-x` info. If we did require this, could it be a more stable way of identifying commits? That way we don't need to manually check for merge conflicted commits (which I can't imagine anyone would take the time to do).

clydin:
The documentation for the option says it doesn't (https://git-scm.com/docs/git-cherry-pick) so I'm not sure we should rely on the behavior.  but we could augment the check to also try to find the text and analyze the referenced commit if no initial match is found.

This doesn't cover the case where separate PRs are created due to the conflicts though.  From my manual review of 9.1.0-next.0, this was the more common case.

dgp1130:
Interesting that `-x` does work with merge conflicts even though it is explicitly documented not to. Agreed that we probably shouldn't rely on this seemingly incorrect behavior. Also agreed that this doesn't address separate PRs to the release branch, so we'd still need to do the patch id trick there. Just wanted to see if utilizing `-x` could be better, but seems like it really wouldn't.

clydin:
Yes.  I agree, I was hoping it would be more useful especially since I noticed you started using it when cherry-picking.
I'll do some further experiments with trying to further reduce the extra commits.  While this PR isn't a perfect solution, it is a pretty good one (90+% of the commits were removed).

Issue #17053, opened 2020-02-22
Angular CLI can't tell when stdout isn't a TTY, always outputs ANSI


ocket8888:
# 🐞 Bug report

### Command (mark with an `x`)
- [x] new
- [x] build
- [x] serve
- [x] test
- [x] e2e
- [x] generate
- [x] add
- [x] update
- [x] lint
- [x] xi18n
- [x] run
- [x] config
- [x] help
- [x] version
- [x] doc

### Is this a regression?
No

### Description
The Angular CLI always outputs ANSI control codes to colorize output, regardless of whether or not the process's STDOUT is a TTY, and offers no option to forcibly override that behavior. This makes the output difficult/impossible (depending on the command) to parse without passing it through a filter to first strip out all of the ANSI characters.

I'm opening this as a bug - not a feature request - because #6478 has been closed without a "wontfix" label, indicating the intended behavior is that this problem has been solved.

## 🔬 Minimal Reproduction
Run any command in any environment with any passed option-arguments or "flags", observe colorized output.

## 🔥 Exception or Error
Here's some sample output when an `ng build` results in an error - thus producing a *lot* of colorized output:
<pre><code>
chunk {main} main.js, main.js.map (main) 628 bytes [initial] [rendered]
chunk {polyfills} polyfills.js, polyfills.js.map (polyfills) 122 kB [initial] [rendered]
chunk {runtime} runtime.js, runtime.js.map (runtime) 6.15 kB [entry] [rendered]
chunk {styles} styles.js, styles.js.map (styles) 11 kB [initial] [rendered]
Date: 2020-02-22T20:31:55.400Z - Hash: f4ca5d6cd60b894262e5 - Time: 5043ms

ERROR in [96msrc/app/app.component.ts[0m:[93m28[0m:[93m50[0m - [91merror[0m[90m TS2339: [0mProperty 'entities' does not exist on type 'EntitiesService'.

[7m28[0m  	const id = `${Object.keys(this.entitiesService.entities).length}-${(new Date()).getTime()}`;
[7m  [0m [91m 	                                               ~~~~~~~~[0m
[96msrc/app/app.component.ts[0m:[93m29[0m:[93m24[0m - [91merror[0m[90m TS2339: [0mProperty 'entities' does not exist on type 'EntitiesService'.

</code></pre>

## 🌍 Your Environment
<pre><code>

     _                      _                 ____ _     ___
[...]

PR #17055, opened 2020-02-23
fix(@angular/cli): directly remove ansi color codes when no color support

clydin:
Third party libraries can attempt to write color codes to the output even though the CLI has already determined that color should not be used.  The previously implemented color removal code is no longer functional since the update of ansi-colors to 4.1.0.  While this appears to be a defect in the aforementioned package, the new CLI removal method not only bypasses the defect but also unneeded execution logic that the CLI does not need in this case.

Fixes: #17053


dgp1130:

Issue #17063, opened 2020-02-11
[ivy/ngcc] Fails to inject optional dependency in Angular8 library


PapaNappa:
# 🐞 bug report

### Affected Package

 The issue is caused by package @angular/core or ngcc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 8

### Description

I have a library (still building with Angular8 for backwards compatibility) that has one main service `Lib2Service`:

```
@Injectable({ providedIn: 'root' })
export class Lib2Service {
  constructor(
      @Optional() private lib1: BaseLib1 | null,
  ) {}

  getValue() {
    if (this.lib1) return this.lib1.getValue();
    return -1;
  }
}
```

This `Lib2Service` takes an optional service `BaseLib1`.
The use case is to provide plugins to this service.

In this example, the 'plugin' `BaseLib1` has only a single method:

```
export class BaseLib1 {
  getValue() {
    return 0;
  }
}
```

The Angular8 library also provides a default implementation `Lib1Service`:

```
@Injectable()
export class Lib1Service extends BaseLib1 {
  getValue() {
    return 1;
  }
}
```

The intended use in any app is then to provide a certain implementation of `BaseLib1`, e.g. like so using the default implementation:

```
@NgModule({
  declarations: [AppComponent],
  imports: [BrowserModule],
  providers: [
    Lib1Service,
    { provide: BaseLib1, useExisting: Lib1Service },
  ],
  bootstrap: [AppComponent]
})
export class AppModule {
}
```

Then, any component can use `Lib2Service`, like so:

```
@Component({
  selector: 'app-root',
  template: `
      lib service: {{lib2Service.getValue()}}
  `,
  styles: []
})
export class AppComponent {
  constructor(
      public lib2Service: Lib2Service,
  ) {
  }
}
```

This works fine when the app is also using Angular 8, but with Angular 9, `Lib2Service` is constructed with its `lib1` parameter undefined.

If I add `@Inject(BaseLib1)` to the ctor parameter of `Lib2Service`, the dependency is injected correctly. However, this was not necessary with an Angular 8 app and breaks the existing version of the library.

## 🔬 Minimal Reproduction

I have reproduced the Angular 9 app at https://github.com/PapaNappa/Angular9-DI-Bug
It contains a .tgz of the Angular8 library, which is basically the three classes from the top of this post.

### Steps to reproduce:

1. `git clone https://github.com/PapaNappa/Angular9-DI-Bug`
[...]

PR #17077, opened 2020-02-25
fix(@ngtools/webpack): handle union type with a nullable argument

alan-agius4:
Currently constructor parameters with union types that contains nullable argument are not being converted properly and result in broken behaviour.

With this change we align the ctor-parameters downlevel transformer to be closer to the NGTSC reflector: https://github.com/angular/angular/blob/master/packages/compiler-cli/src/ngtsc/reflection/src/typescript.ts#L65-L66

This change should also be synced to ng-packagr.

Closes: #17063
Reference: FW-1883


dgp1130:

petebacondarwin:
🎉

Issue #17002, opened 2020-02-17
Templates are processed too early


giniedp:
# 🐞 Bug report

Having the `AngularCompilerPlugin` in a custom steup instantiated like this

```js
new AngularCompilerPlugin({
    platform: 0,
    mainPath: config.isTest
      ? config.frontend.entry.test
      : config.frontend.entry.main,
    tsConfigPath: config.isTest
      ? config.frontend.tsconfigTestFile
      : config.frontend.tsconfigAppFile,

    skipCodeGeneration: !config.isAot,
    sourceMap: true,
    forkTypeChecker: true,
    directTemplateLoading: false,

    compilerOptions: {
      fullTemplateTypeCheck: true,
      strictInjectionParameters: true,
      strictTemplates: false,
      enableIvy: config.isIvy
    }
  })
```

and a webpack rule for `.pug` templates 

```js
      {
        test: /\.(pug)$/,
        use: [{ loader: "apply-loader" }, { loader: "pug-loader" }]
      },
```

the templates are processed by the `ngtc` compiler without being transpiled to `.html`. The compiler fails when when template contains certain syntax with following error

```
  ERROR in Errors parsing template: Unexpected character "EOF" (Do you have an unescaped "{" in your template? Use "{{ '{' }}") to escape it.) ("h1 {{ title }}
  input([(ngModel)]="title", [ngModelOptions]="{ standalone: true }")
  router-outlet()
  [ERROR ->]"): /Users/alex/repositories/ginie/my-app/src/app/app.component.pug@3:0, Invalid ICU message. Missing '}'. ("h1 {{ title }}
  input([(ngModel)]="title", [ngModelOptions]="{ standalone: true }")
  router-outlet()
  [ERROR ->]"): /Users/alex/repositories/ginie/my-app/src/app/app.component.pug@3:0
```

- failure happens when compiled with `ivy` enabled and `aot` disabled
- failure does not happen with `ivy` and `aot` enabled
- fails only on certain syntax for example

this template would fail
```pug
h1 {{ title }}
input([(ngModel)]="title", [ngModelOptions]="{ standalone: true }")
```

but this template would not. Interestingly the application the renders correctly, as if the pug is transpiled at a later stage
```pug
h1 {{ title }}
input([(ngModel)]="title")
```

[...]

PR #17086, opened 2020-02-25
fix(@ngtools/webpack): always use VE for deprecated string route discovery

clydin:
When in JIT mode during lazy route discovery, the Ivy compiler willl attempt to parse templates which may fail when this plugin is used with webpack loaders that support custom template formats.  The string lazy routes must be discovered before the webpack build starts but the template loading/processing does not occur until the webpack build commences.

Fixes: #17002


dgp1130:

PR #17094, opened 2020-02-26
fix(@angular-devkit/build-angular): allow function in filename while …

sacgrover:
…changing the name of compiling chunks

Fixes angular#17087

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
@@ -149,12 +149,13 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
                 // tslint:disable-next-line: no-any
                 (compilation.mainTemplate.hooks as any).assetPath.tap(
                   'build-angular',
-                  (filename: string, data: ChunkData) => {
-                    const isMap = filename && filename.endsWith('.map');
+                (filename: string | ((data: ChunkData) => string), data: ChunkData) => {
+                    const assetName = typeof filename === 'function' ? filename(data) : filename;
+                    const isMap = assetName && assetName.endsWith('.map');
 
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -149,7 +149,9 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
                 // tslint:disable-next-line: no-any
                 (compilation.mainTemplate.hooks as any).assetPath.tap(
                   'build-angular',
-                  (filename: string, data: ChunkData) => {
+                // tslint:disable-next-line: no-any
+                  (filename: any, data: ChunkData) => {
```
clydin:
```suggestion
                  (filename: string | ((data: ChunkData) => string), data: ChunkData) => {
```

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -149,7 +149,9 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
+                    if (typeof filename === 'function') { filename = filename(data); }
```
clydin: Can you rework this so it does not reassign the argument?

sacgrover: Sure

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
sacgrover: Thanks for this.. I had no idea that we can declare function type this way too!! noted for future!

sacgrover:
@clydin is there any way to retry above fail checks without new commit?

dgp1130:

PR #17096, opened 2020-02-26
fix(@angular-devkit/build-angular): add missing styles extensions in anyComponentStyle budget

wojteko22:
Currently anyComponentStyle works only for css styles. Files with other extensions are ignored.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/any-component-style-budget-checker.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/any-component-style-budget-checker.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as path from 'path';
 import { Compiler, Plugin } from 'webpack';
 import { Budget, Type } from '../../../src/browser/schema';
 import { ThresholdSeverity, calculateThresholds, checkThresholds } from '../utilities/bundle-calculator';
@@ -32,8 +33,16 @@ export class AnyComponentStyleBudgetChecker implements Plugin {
           return;
         }
 
+        const cssExtensions = [
+          '.css',
+          '.scss',
+          '.less',
+          '.styl',
+          '.sass',
+        ];
+
         const componentStyles = Object.keys(compilation.assets)
-            .filter((name) => name.endsWith('.css'))
+            .filter((name) => cssExtensions.includes(path.extname(name)))
             .map((name) => ({
               size: compilation.assets[name].size(),
               label: name,
--- a/packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
@@ -11,6 +11,7 @@ import { createArchitect, host } from '../utils';
 
 
 describe('Browser Builder bundle budgets', () => {
+  const cssExtensions = ['css', 'scss', 'less', 'styl'];
   const targetSpec = { project: 'app', target: 'build' };
   let architect: Architect;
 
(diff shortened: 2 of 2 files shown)


wojteko22:
@googlebot I signed it!

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/any-component-style-budget-checker.ts
```diff
@@ -33,7 +33,7 @@ export class AnyComponentStyleBudgetChecker implements Plugin {
         }
 
         const componentStyles = Object.keys(compilation.assets)
-            .filter((name) => name.endsWith('.css'))
+            .filter((name) => ['.css', '.scss', '.less', '.styl'].some((styleExtension) => name.endsWith(styleExtension)))
```
alan-agius4: `.sass` is missing from the extension list, also this logic can be simplified slightly as there is no need to re-declare the list of styles extensions for every iteration.

```ts

const cssExtensions = [
  '.css',
  '.scss',
  '.less',
  '.styl',
  '.sass',
];

const componentStyles = Object.keys(compilation.assets)
    .filter(name => cssExtensions.includes(path.extname(name)))
    .map((name) => ({
      size: compilation.assets[name].size(),
      label: name,
    }));
```

alan-agius4:
@wojteko22, are you still willing to work on this?

wojteko22:
@alan-agius4 yep, I'll fix it this week

packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
```diff
@@ -64,66 +65,82 @@ describe('Browser Builder bundle budgets', () => {
...
-      .foo { color: white; padding: 1px; }
-      .buz { color: white; padding: 2px; }
-      .bar { color: white; padding: 3px; }
-    `;
+  cssExtensions.forEach(ext => {
+    it(`shows warnings for large component ${ext} when using 'anyComponentStyle' when AOT`, async () => {
+      const overrides = {
+        aot: true,
+        optimization: true,
+        budgets: [{ type: 'anyComponentStyle', maximumWarning: '1b' }],
```
alan-agius4: We can add `styles: [`src/styles.${ext}`]` here so that we are sure that the budget only applies to component styles?

packages/angular_devkit/build_angular/test/browser/bundle-budgets_spec_large.ts
```diff
@@ -64,66 +65,82 @@ describe('Browser Builder bundle budgets', () => {
...
-    const cssContent = `
-      .foo { color: white; padding: 1px; }
-      .buz { color: white; padding: 2px; }
-      .bar { color: white; padding: 3px; }
-    `;
+  cssExtensions.forEach(ext => {
+    it(`shows error for large component ${ext} when using 'anyComponentStyle' when AOT`, async () => {
+      const overrides = {
+        aot: true,
+        optimization: true,
```
alan-agius4: We can add `styles: [`src/styles.${ext}`]` here so that we are sure that the budget only applies to component styles?

eweap:
Hi @wojteko22 ,  thanks for your PR,

It seems that non-compiled output size (like scss variables or mixins imports) is counted to the total component styles, am I right ?

alan-agius4:
No, @eweap, that shouldn’t be the case.

eweap:
@alan-agius4 Okay thanks !

I am not comfortable enough with the angular-cli code to check myself but I see an increase of ~200kb of the component style after adding the theming functions `~@angular/material/theming`

And some stackoverflow topics started to pop out about that behavior .

alan-agius4:
@EvanCarroll, mind filling an issue please?, as that sounds like a bug which was never reported.I’ll try to take a look at this tmr.

eweap:
@alan-agius4 I think that I found an answer !

My `budgets` was set in `architect.build.options.budgets` instead of  `architect.build.configurations.production.budgets`, and so the `ng x18n` failed with

```
ERROR in Exceeded maximum budget for /path/to/my-component.scss. Budget 10 kB was not met by 197 kB with a total of 207 kB
```

So it's not a bug since it's documented at https://angular.io/guide/build#configuring-size-budgets :

> Define your size boundaries in the CLI configuration file, angular.json, in a budgets section for each configured environment.

Thanks and sorry to have wasted your time !

PR #17105, opened 2020-02-27
feat(@ngtools/webpack): integrate ngcc async

alan-agius4:
With this change we call NGCC async version prior to every build which will process the entire node modules tree.


packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -61,18 +107,9 @@ export class NgccProcessor {
     }
 
     const packageJsonPath = this.tryResolvePackage(moduleName, resolvedFileName);
-    if (!packageJsonPath) {
-      // add it to processed so the second time round we skip this.
-      this._processedModules.add(resolvedFileName);
-
-      return;
-    }
-
```
petebacondarwin: why remove this block?

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -49,6 +50,51 @@ export class NgccProcessor {
...
+
+    await mainNgcc({
+      basePath: this._nodeModulesDirectory,
+      propertiesToConsider: this.propertiesToConsider,
+      compileAllFormats: false,
+      createNewEntryPointFormats: true,
+      logger: this._logger,
+      async: true,
+      pathMappings: this._pathMappings,
+    });
```
petebacondarwin: I think we will also need to run ngcc again against `basePath`s that have been computed from the `pathMappings` property to catch local libraries too.

alan-agius4: How would that work, without providing a module name?


Currently for those will be handled via the sync pipeline.

packages/ngtools/webpack/src/ngcc_processor.ts
alan-agius4: It’s been moved https://github.com/angular/angular-cli/pull/17105/files#diff-456f73b84b7b60a60000b590b3301f7bR112

petebacondarwin: Oh! Then I think the comment before the line where it has been moved to needs to change. Since that only mentions readonly, I was thrown and didn't realise it also handled missing package.json files.

packages/ngtools/webpack/src/ngcc_processor.ts
petebacondarwin: Actually don't worry. It is handled inside ngcc already:  https://github.com/angular/angular/blob/a581773887eaf872f484e4a6469ad1a6068d8305/packages/compiler-cli/ngcc/src/entry_point_finder/directory_walker_entry_point_finder.ts#L32-L34

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -162,3 +199,13 @@ class NgccLogger implements Logger {
     this.compilationErrors.push(new Error(args.join(' ')));
   }
 }
+
+function isReadOnlyFile(fileName: string): boolean {
```
clydin: This appears to always return true?

alan-agius4: Oops, nice catch.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -821,6 +821,12 @@ export class AngularCompilerPlugin {
       );
     });
 
+    compiler.hooks.beforeCompile.tapPromise('angular-compiler', async () => {
```
alan-agius4: Self note: remove async.

PR #17107, opened 2020-02-27
Initial TypeScript 3.8 Support

clydin:
New projects are not yet created with TypeScript 3.8.  That change requires support from within the AOT compiler and can be done separately once supported.


dgp1130:
Looks like there are a couple merge conflicts to deal with here.

dgp1130:
Looks like more merge conflicts, sorry I didn't realize how large this was. I can try merging this before other PRs to avoid introducing another merge conflict.

Issue #11227, opened 2018-06-13
ng test --code-coverage gives wrong results and do not include all files


zielinskikamil:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report
- [ ] feature request
```

### Area
```
- [x] devkit
- [ ] schematics
```

### Versions
node v8.9.4
npm 5.6.0
macOS Sierra

### Repro steps
```
ng new CoverageTest
cd CoverageTest
ng g component componentA
ng g component componentB
// add dumy public method to componentb.ts
/*
 Additionally create some file and put there some simple class with one method. eg.
class ServiceA {
    public getHello(): string {
        return 'hello';
    }
}
*/

// now generate code coverage
ng test --code-coverage --watch false
// open coverage/index.html
```

### Description
Code coverage generates a report but its misleading:
1. It doesn't include `ServiceA` coverage
2. Despite there are just two methods in `componentb`, coverage is eqal to 80%. If I preview covered area it covered untested file as well!
3. After editing `test.ts` (replace  
`const context = require.context('./', true, /\.spec\.ts$/);`
with
`const context = require.context('./', true, /\/app\/.*\.ts$/);`
and reruning `ng test --code-coverage --watch false`
ServiceA is visible in report but has some unrealistic values (Statement Coverage at 80%, Branches at 100%, Lines 83%)

### Desired functionality
Ad 1. Coverage that does not include all files is misleading. How do I know which files are covered in unit tests?
Ad 2. How it happened that componentB is covered in 80% and newly created method is not covered in view? I suspect that mechanism that loads JS makes such faults positive scenario.
Ad 3. Why ServiceA shows as being covered despite there were no test for it?

All these 3 things look like a bug. Can you guys take a look at it?

jeanpaulattard:
I have also encountered this issue. 

I've starting writing tests for an application which has a considerable amount of files, and when i've opened the coverage report, to my surprise it only contained the files under test. (in previous projects working with boiler plate code, it always displayed all the files).

Is there a workaround for this? Do I have a missing config, or is it an issue in angular cli?

sumeet-singh04:
Any updates on this issue ? I am able to exclude files using the angular.json, but not sure how to add them.

LMFinney:
@sumeet-singh04 I just tried [this approach](https://stackoverflow.com/a/49259776), and it seemed to work without slowing my tests down appreciably. Does anyone know a downside to do this?

gabmorais:
[...]

Issue #1735, opened 2016-08-17
Option for Istanbul coverage to include all source files (as opposed to only files loaded during test)


westonpace:
> Please provide us with the following information:
> ---------------------------------------------------------------

1. OS?  Windows 7, 8 or 10. Linux (which distribution). Mac OSX (Yosemite? El Capitan?)
Any

2. Versions. Please run `ng --version`. If there's nothing outputted, please run
in a Terminal: `node --version` and paste the result here:

angular-cli: 1.0.0-beta.11-webpack.2
node: 6.2.2
os: linux x64

3. Repro steps. Was this an app that wasn't created using the CLI? What change did you
   do on your code? etc.

Create an application.  Test some, but not all of the source files.  Generated coverage will not include files that were not loaded by the tests.

4. The log given by the failure. Normally this include a stack trace and some
   more information.

N/A

5. Mention any other details that might be useful.

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

Similar to gotwarlost/istanbul#275 and karma-runner/karma-coverage#125

jtesser:
+1
karma-coverage has an option includeAllSources which could be used.

filipesilva:
If anyone has is willing to do a PR, or point me in the right direction to do it with webpack/karma-remap-istanbul, I'd be happy to look at it.

jtesser:
@filipesilva what exactly are you asking for I am confused.

westonpace:
Given that I understand webpack a little better now the includeAllSources option in karma would not work.  Under the hood that is including all files that were instrumented.  Webpack won't instrument code that isn't loaded so includeAllSources would not see it.

Instead I ended up using a simpler workaround.  I added a single file:

src/app.module.spec.ts

which is simply (e.g. no test suites or cases):

`import './app.module';`

This file will cause webpack to load the entire application.  The istanbul coverage will only be generated for local code (since node_modules is excluded).  As far as I can tell this is a pretty ideal solution with 2 caveats:

* It is still possible for code to be missed in coverage if that code is not included in the application at all.  One could potentially mitigate this with a webpack plugin that checked for unused code in ./src and reported it.
* It will cause an increase in test times depending on the size of the application.

westonpace:
[...]

PR #17136, opened 2020-03-03
ci(@angular-devkit/build-angular): add ability to include untested files to coverage result

kopach:
Fix for issue  #1735 and #11227. Add `karma-sabarivka-reporter` as dependency for any new Angular project. This would allow including (if needed) untested files to coverage result. By default, no extra files are included, `include` config is empty.

### DIFF
--- a/packages/angular/cli/README.md
+++ b/packages/angular/cli/README.md
@@ -162,6 +162,7 @@ You can find more details about changes between versions in [the Releases tab on
 
 ```bash
 git clone https://github.com/angular/angular-cli.git
+cd angular-cli
 yarn
 npm run build
 cd dist/@angular/cli
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -94,6 +94,7 @@
     "karma-coverage-istanbul-reporter": "~2.1.0",
     "karma-jasmine": "~3.1.0",
     "karma-jasmine-html-reporter": "^1.4.0",
+    "karma-sabarivka-reporter": "~3.0.2",
     "popper.js": "^1.14.1",
     "protractor": "~5.4.0",
     "tslib": "~1.11.0",
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
@@ -70,10 +70,15 @@ const init: any = (config: any, emitter: any, customFileHandlers: any) => {
   successCb = config.buildWebpack.successCb;
   failureCb = config.buildWebpack.failureCb;
 
-  // When using code-coverage, auto-add coverage-istanbul.
+  // When using code-coverage, auto-add coverage-istanbul & karma-sabarivka-reporter.
   config.reporters = config.reporters || [];
-  if (options.codeCoverage && config.reporters.indexOf('coverage-istanbul') === -1) {
-    config.reporters.push('coverage-istanbul');
+  if (options.codeCoverage) {
+    if (config.reporters.indexOf('coverage-istanbul') === -1) {
+      config.reporters.push('coverage-istanbul');
+    }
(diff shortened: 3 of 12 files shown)


kopach:
Hi @alan-agius4, sorry for reaching you out here. I see you're doing reviews on other PRs in this repo. Any chance, you could review also this one, please? Those are very simple changes. Also, those changes are based on people's need (see linked issues in description). Any comments/advices on this PR? I could apply to any chances, if needed. If you can't review this – give me, please, at least some hint, who can do this and/or when. Thanks.

alan-agius4:
Hi @kopach, thanks for the effort, I can do the review of this PR. 

However, let's me take this with the rest of the team to see if there is a consensus that we should add this in the first place.

kopach:
@alan-agius4 yes, of course. Thank you very much. Note also, that I'm angular-cli daily basis user as well as an author of `karma-sabarivka-reporter`. So, if there is anything I need to do over `karma-sabarivka-reporter` repo to make that functionality be available in `angular-cli` – just let me know.

alan-agius4:
@kopach thanks :)

alan-agius4:
Hi @kopach, firstly I'd like to thank you for the effort you put into this PR.

We have discussed this and we feel that the testing and code coverage story within the Angular CLI needs to be revised.

We'd also like to avoid adding additional dependencies unless absolutely necessary, which in this case might be unneeded, as there are other approaches to add untested files to the compilation. One of them being adding `test.ts` file add include more files.

Thanks once again.

kopach:
Hi @alan-agius4, thank you for your response. Your arguments are reasonable for me, I would also be afraid of adding some 3-rd party libs to highly used project unless there are good reasons for it. So, let me give you few of such. 

- even though there are indeed multiple ways to add untested files to coverage report, I believe `karma-sabarivka-reporter` serves slightly different purpose. For example, as project lead, I would like to make sure, there won't be any service/component not covered by tests. So, suppose, someone incidently added some service without test file or removed test file from some service. Currently, this service will be just excluded from coverage report. But with `karma-sabarivka-reporter` it still will be there, just with 0% coverage. `test.ts` - won't help here, I believe, as we can add there only `*spec.ts` which later will import source files. If we instead add there also all source files of the app (so, make `require.context('./', true, /\.ts$/)` instead of `require.context('./', true, /\.spec\.ts$/)`), to make sure, that all the files are under coverage control - tests won't just work.
- by default, on new project generation with `ng new my-app` -> `karma-sabarivka-reporter` will do nothing, config is empty, so no breaking change here, only feature :) 
- even if there are other solutions (which I would really like to know about as I've spent quite some time to find such until I've made that lib) – people want to have simple solution, the best would be to have it out of the box (see #1735 and #11227)
- angular was always encouraging people to write tests, with this tool, we can make it even more explicit and safe (no files will be incidently omitted from being tested)
- `karma-sabarivka-reporter` is 100% tested
- `karma-sabarivka-reporter` is updated regularly by automated tools to prevent from any new possible vulnerabilities in dependencies/sub-dependencies
- and last but not least, growing [download statistic testifies](https://www.npmjs.com/package/karma-sabarivka-reporter), that this tool is really wanted.

P.S.
If you still think, that this tool is too "young" to be merged to `angular-cli` – please give me some hints, how big popularity of this tool should be to make it reasonable for being used by `angular-cli`. Or what features should be there. Thanks.

alan-agius4:
Hi @kopach,

This is not a matter of using a relatively new library as our dependency, but rather we'd like to revisit the entire testing story within the CLI as we feel that there are some features we have some short comings or their behaviour is not clear, one of them being code coverage.

One of the questions that we have internally within the team is actually related to this feature. Should code coverage include untested all or not?

So before adding public facing dependencies and amending configurations we want to be absolutely clear on how testing and code coverage should be in the CLI. 

At this point also, no one is blocked from adding the `karma-sabarivka-reporter` in their karma configuration file.

Thanks once again!

kopach:
ok, clear. Thank you @alan-agius4.

PR #17149, opened 2020-03-04
refactor(@schematics/angular): remove redundant typeRoots

mgechev:
By default TypeScript will look up in this directory (see [docs](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types)).


dgp1130:

PR #17159, opened 2020-03-06
test: change webpack angular tests to use es2015 

alan-agius4:
This is to avoid having NGCC go through and transform ES5 entrypoints


tests/angular_devkit/build_webpack/angular-app/tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
-    "target": "es5",
+    "target": "es2015",
```
alan-agius4: This is to avoid having NGCC go through and transform ES5 entrypoints, currently NGCC is taking more time than expected to process `@angular/core`.

PR #17185, opened 2020-03-10
docs: fix typo in the schematics ng update

mprinc:
am -> an or probably even better the


alan-agius4:
@mprinc, actually I think as suggested `the` is more correct. Can you kindly update it? Thanks

mprinc:
OK, done

PR #17203, opened 2020-03-12
feat(@schematics/angular): generate new projects with tslint 6.1.0

alan-agius4:
tslint 6.1 adds support for TypeScript 3.8 syntax

See: https://github.com/palantir/tslint/blob/master/CHANGELOG.md#v610

Pr to update the package in the repo https://github.com/angular/angular-cli/pull/17202

See individual commits for more details.

### DIFF
--- a/packages/schematics/angular/migrations/update-10/update-tslint.ts
+++ b/packages/schematics/angular/migrations/update-10/update-tslint.ts
@@ -1,0 +1,156 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { JsonAstObject, JsonValue } from '@angular-devkit/core';
+import { Rule } from '@angular-devkit/schematics';
+import { addPackageJsonDependency, getPackageJsonDependency } from '../../utility/dependencies';
+import { findPropertyInAstObject, insertPropertyInAstObjectInOrder, removePropertyInAstObject } from '../../utility/json-utils';
+import { readJsonFileAsAstObject } from '../update-9/utils';
+
+export const TSLINT_VERSION = '~6.1.0';
+const TSLINT_CONFIG_PATH = '/tslint.json';
+
+const RULES_TO_DELETE: string[] = [
+  'no-use-before-declare',
+  'no-unused-variable',
+];
+
+const RULES_TO_ADD: Record<string, JsonValue> = {
+  align: {
+    options: ['parameters', 'statements'],
+  },
+  'arrow-return-shorthand': true,
+  curly: true,
+  eofline: true,
+  'import-spacing': true,
+  indent: {
+    options: ['spaces'],
+  },
+  'variable-name': {
+    options: ['ban-keywords', 'check-format', 'allow-pascal-case'],
+  },
+  semicolon: { options: ['always'] },
+  'space-before-function-paren': {
+    options: {
+      anonymous: 'never',
+      asyncArrow: 'always',
+      constructor: 'never',
+      method: 'never',
+      named: 'never',
+    },
+  },
+  'typedef-whitespace': {
+    options: [
+      {
+        'call-signature': 'nospace',
+        'index-signature': 'nospace',
+        parameter: 'nospace',
+        'property-declaration': 'nospace',
+        'variable-declaration': 'nospace',
+      },
+      {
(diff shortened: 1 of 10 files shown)


alan-agius4:
A number of recommended rules have been removed in version 6, Should we add them in our `tslint.json` or not?

https://github.com/palantir/tslint/pull/4871/files and https://github.com/palantir/tslint/pull/4312/files

dgp1130:
Discussed whether to preserve existing configs from previous version or accept new changes from TSLint.

New projects should include the new version and have a custom config with all the old rules from the previous TSLint version.

Existing projects in 9.x should have a migration allowing users to update to TSLint 6.1 if they choose to and include the old rules into their config.

In v10 we would enable the migration for everyone, need to consider how this works with Prettier as well.

This will prevent projects from backsliding on formatting. Once we get to ESLint migration, we can revisit how to best make that happen.

packages/schematics/angular/migrations/update-10/update-tslint.ts
```diff
@@ -0,0 +1,147 @@
...
+      }
+
+      const recorder = tree.beginUpdate(TSLINT_CONFIG_PATH);
+      removePropertyInAstObject(recorder, rulesAst, rule);
+      tree.commitUpdate(recorder);
+    }
+
+    // Add new rules only if the configuration doesn't extend 'tslint-config-prettier'.
+    // This is because some rules conflict with prettier.
+    const extendsAst = findPropertyInAstObject(tslintJsonAst, 'extends');
```
clydin: If an extends property within the extension hierarchy does not contain `tslint:recommended` then this part should be skipped.  Since this is attempting to augment upstream changes to the `tslint:recommended` ruleset to ensure a stable ruleset after update, without its presence there should be nothing to augment.

packages/schematics/angular/migrations/update-10/update-tslint.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (
+      (extendsAst?.kind === 'string' && extendsAst.value === 'tslint-config-prettier') ||
```
clydin: This should cover the `tslint-config-prettier` ruleset case but there may be other base rulesets in use that could also affect the final ruleset.  Order of the rulesets in the extends property can also cause changes.
I think parsing the configuration and analyzing the result will be needed here to ensure that any additions do not alter the final ruleset. Maybe a combination of [readConfigurationFile](https://github.com/palantir/tslint/blob/0ebda9b11a4fe7f27f6eb25aedc552960ca584c6/src/configuration.ts#L262) & [parseConfigFile](https://github.com/palantir/tslint/blob/0ebda9b11a4fe7f27f6eb25aedc552960ca584c6/src/configuration.ts#L533) from tslint's programmatic interface.

packages/schematics/angular/migrations/update-10/update-tslint.ts
```diff
@@ -0,0 +1,147 @@
...
+      addPackageJsonDependency(tree, {
+        type: current.type,
+        name: 'tslint',
+        version: TSLINT_VERSION,
+        overwrite: true,
+      });
+    }
+
+    // Update tslint config.
+    const tslintJsonAst = readJsonFileAsAstObject(tree, TSLINT_CONFIG_PATH);
```
clydin: tslint config files can also be JS or YAML in addition to JSON.  Maybe check, warn, and skip if one of those is found?

There can also be additional configs in subdirectories.  tslint has a discovery mechanism wherein the _closest_ tslint configuration per linted file is used for that file unless a configuration file is explicitly specified upon execution.

cexbrayat:
@alan-agius4 Thank you for considering my proposition!
I just reviewed the updated rules and it looks good to me, except one tiny nit: `variable-name` has been relaxed and now [allows names with a leading underscore](https://github.com/palantir/tslint/pull/4871/files#diff-6e3940e8ec3d59837c4435f32975ed2cR125), whereas it wasn't previously.
So to be stricly similar to what we had, I think the CLI tslint should include: 
```json
    "variable-name": {
        "options": ["ban-keywords", "check-format", "allow-pascal-case"],
    },
```

packages/schematics/angular/migrations/update-10/update-tslint.ts
alan-agius4: As discussed offline we'll only run the migration iif the config extends `tslint:recommended`

alan-agius4:
@clydin PTAL

PR #17218, opened 2020-03-13
fix(@ngtools/webpack): disable ngcc async under Bazel

alan-agius4:
Under Bazel some dependencies might be readonly we shouldn't run NGCC async version because we are unable to verify which modules are read-only and which not.

### DIFF
--- a/packages/ngtools/webpack/src/ngcc_processor.ts
+++ b/packages/ngtools/webpack/src/ngcc_processor.ts
@@ -52,18 +52,19 @@ export class NgccProcessor {
 
   /** Process the entire node modules tree. */
   process() {
-    const timeLabel = 'NgccProcessor.process';
-    time(timeLabel);
-
-    const corePackage = this.tryResolvePackage('@angular/core', this._nodeModulesDirectory);
-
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -52,19 +52,14 @@ export class NgccProcessor {
...
   /** Process the entire node modules tree. */
   process() {
-    const timeLabel = 'NgccProcessor.process';
-    time(timeLabel);
-
-    const corePackage = this.tryResolvePackage('@angular/core', this._nodeModulesDirectory);
-
-    // If the package.json is read only we should skip calling NGCC.
-    // With Bazel when running under sandbox the filesystem is read-only.
-    if (corePackage && isReadOnlyFile(corePackage)) {
```
clydin: I think it might be good to keep this as a check for other potential cases of read-only node modules directories.

alan-agius4: Good point, added back.

PR #17250, opened 2020-03-18
feat(@angular-devkit/build-angular): show warnings when depending on CommonJS

alan-agius4:
feat(@angular-devkit/build-angular): show warnings when depending on CommonJS.

Depending on CommonJS modules is know to cause optimization bailouts. With this change when running a browser build and scripts optimization is enabled we display a warning.

To suppress the warning for a particular package, users can use the `allowedCommonJsDepedencies` builder options.

Example:
```
"build": {
  "options": {
    ...
    "allowedCommonJsDepedencies": ["bootstrap"]
  },
}
```

Reference: TOOL-1328

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
@@ -87,8 +87,8 @@ export interface BuildOptions {
 
   /* Append script target version to filename. */
   esVersionInFileName?: boolean;
-
   experimentalRollupPass?: boolean;
+  allowedCommonJsDependencies?: string[];
 }
 
 export interface WebpackTestOptions extends BuildOptions {
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -380,6 +380,14 @@
       "type": "boolean",
       "description": "Concatenate modules with Rollup before bundling them with Webpack.",
       "default": false
+    },
+    "allowedCommonJsDependencies": {
+      "description": "A list of CommonJS packages that are allowed to be used without a built time warning.",
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "default": []
     }
   },
   "additionalProperties": false,
--- a/packages/angular_devkit/build_angular/test/browser/common-js-warning_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/common-js-warning_spec_large.ts
@@ -1,0 +1,57 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Architect } from '@angular-devkit/architect';
+import { logging } from '@angular-devkit/core';
+import { createArchitect, host } from '../utils';
+
+describe('Browser Builder commonjs warning', () => {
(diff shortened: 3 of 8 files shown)


alan-agius4:
Blocking until 9.1 branch is created.

alan-agius4:
To discuss:
- Should end users be able to disable these warnings? (If yes, discuss how)
- How to handle features for majors when there is no major branch yet.

alan-agius4:
Decided that we should go with an builder option that allows a list of packages to suppress the warning for.

packages/angular_devkit/build_angular/test/browser/common-js-warning_spec_large.ts
```diff
@@ -0,0 +1,56 @@
...
+    const run = await architect.scheduleTarget(targetSpec, undefined, { logger });
+    const output = await run.result;
+    expect(output.success).toBe(true);
+    expect(logs.join()).toMatch(/WARNING in.+app\.component\.ts depends on bootstrap\. CommonJS or AMD dependencies/);
+    await run.stop();
+  });
+
+  it('should not show warning when depending on a Common JS bundle which is allowed', async () => {
+    const overrides = {
+      allowedCommonJsDepedencies: [
```
alan-agius4: @filipesilva / @clydin should allowing a CommonJS dep allow also it's CommonJS transitive dependency automatically?

filipesilva: Hmm that's a tricky one. Transitive dependencies in general really. 

Realistically speaking many project, especially larger ones, will have dozens of transitive commonjs dependencies. So they get these changes and see a screenful of names they have to allow. That's not great. 

On the other hand the whole point of these warnings is to let users know where in their dep graph are there deoptimizations. So it's not any help to not show the transitives.

To make informed decisions the user will also need the full dependency path to the commonjs dependency.

Which then leads to your specific question: should we just show deps up to a commonjs dep, then ignore that subgraph? Yeah that seems a good way to cut down on the total shown.

packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -380,6 +380,14 @@
       "type": "boolean",
       "description": "Concatenate modules with Rollup before bundling them with Webpack.",
       "default": false
+    },
+    "allowedCommonJsDepedencies": {
+      "description": "A list of CommonJS packages or entry-points that are allowed to be used without a built time warning.",
```
filipesilva: Nit: I'd remove "or entry-points " here, because we mostly mean polyfills and main when talking about entry points.

packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
```diff
@@ -87,8 +87,8 @@ export interface BuildOptions {
 
   /* Append script target version to filename. */
   esVersionInFileName?: boolean;
-
   experimentalRollupPass?: boolean;
+  allowedCommonJsDepedencies?: string[];
```
filipesilva: `allowedCommonJsDepedencies` -> `allowedCommonJsDependencies` (missing `n` in `pen`)

packages/angular_devkit/build_angular/src/browser/schema.json
alan-agius4: Updated!

packages/angular_devkit/build_angular/src/angular-cli-files/models/build-options.ts
alan-agius4: Thanks!

PR #17261, opened 2020-03-20
docs: Update policy on LTS fixes

kyliau:
This commit adds guidance on when to merge commits into LTS branches and how PRs should be opened.

Please note that the third criteria is not in effect for now since we don't run CI on LTS branches. We'll discuss this in the upcoming tooling meeting.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -81,6 +81,20 @@ instance), then you will need to fetch the contributor's branch for your local
 Git instance to have knowledge of the commit being cherry picked onto the patch
 branch.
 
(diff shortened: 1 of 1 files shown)


kyliau:
> We don't run CI on LTS branches? I've seen several PRs specifically fixing CI on `8.3.x`. And if we don't today, should we be? I would expect that any releasable branch should have CI on it.
> 
> I assume we'll discuss the `LTS` label in the meeting.

As discussed in meeting today, we'll start running regular CI (not triggered by PRs) on LTS branches.

Issue #17262, opened 2020-03-20
Error | Could not find plugin "proposal-numeric-separator" | Generating ES5 bundles for differential loading


vegarda:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: a few hours ago

### Description

 A clear and concise description of the problem...
Build doesn't work

## 🔬 Minimal Reproduction

<pre><code>
npm install
node --max-old-space-size=4096 ./node_modules/@angular/cli/bin/ng build --prod
</code></pre>

## 🔥 Exception or Error
<pre><code>

Generating ES5 bundles for differential loading...
An unhandled exception occurred: [BABEL] C:\angular\dist\common-es2015.b1722ef05b43f664abc2.js: Could not find plugin "proposal-numeric-separator". Ensure there is an entry in ./available-plugins.js for it. (While processing: "C:\\angular\\node_modules\\@angular-devkit\\build-angular\\node_modules\\@babel\\preset-env\\lib\\index.js")
See "C:\Users\Vegard\AppData\Local\Temp\ng-J7YoSB\angular-errors.log" for further details.
</code></pre>

## 🌍 Your Environment
<pre><code>
Angular CLI: 8.3.25
Node: 12.14.1
OS: win32 x64
Angular: 8.2.14
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic
... platform-webworker, platform-webworker-dynamic, router
... service-worker

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.803.25
@angular-devkit/build-angular     0.803.25
@angular-devkit/build-optimizer   0.803.25
@angular-devkit/build-webpack     0.803.25
@angular-devkit/core              8.3.25
@angular-devkit/schematics        8.3.25
@angular/cdk                      8.2.3
@angular/cli                      8.3.25
@angular/material                 8.2.3
@angular/pwa                      0.803.25
@ngtools/webpack                  8.3.25
@schematics/angular               8.3.25
@schematics/update                0.803.25
rxjs                              6.5.4
typescript                        3.5.3
webpack                           4.39.2

</code></pre>

[...]

PR #17267, opened 2020-03-21
fix(@angular-devkit/build-angular): update babel to 7.8.7

clydin:
Babel versions 7.8.0-7.8.6 contain a defect that can cause failures due to babel plugin changes (even in patch versions).  Plugins are not pinned as they are deep transitive dependencies of the packages adjusted in this commit.  This failure will prevent a newly generated project from building and can only be worked around by disabling differential loading.

Fixes: #17262

NOTE TO CARETAKER: pending status of ci/angular: merge status is incorrect and can be ignored. The two missing status checks do not exist on this branch.


alan-agius4:
@googlebot I content.

alan-agius4:
@googlebot I consent

PR #17273, opened 2020-03-21
style(@schematics/angular): prevent adding redundant whitespace in ge…

wojteko22:
…nerated angular.json file

Previously "ng new" was generating angular.json file containing redundant whitespace after "version" property.
When file was edited some IDEs automatically removed this whitespace making additional change in diff.
If "packageManager" option is used, whitespace is still present.


dgp1130:
Also merged to patch branch 9.1.x.

PR #17286, opened 2020-03-23
fix(@angular-devkit/build-angular): reduce fast sourcemap threshold

clydin:
High fidelity sourcemap processing can be expensive for larger bundle sizes.  This reduces the threshold to the original 500KB value to improve performance.


dgp1130:

PR #17287, opened 2020-03-23
build: set up ts-circular-deps check to prevent new circular deps

josephperrott:
The repo current has zero circular typescript dependencies, this check will ensure no circular dependencies are added.


clydin:
There's a tslint rule in place that checks for circular imports: https://github.com/angular/angular-cli/blob/master/tslint.json#L6
https://www.npmjs.com/package/tslint-no-circular-imports

alan-agius4:
@josephperrott should this be closed? Since we are handling circular deps with tslint?

josephperrott:
@alan-agius4 and I spoke offline, and I believe its still an open question of if this should be added/included.

My main concern is that we prevent circular deps, which is already being done in this repository.  The argument for merging this in is to have the same method for circular deps discovery/enforcement used across our org.

alan-agius4:
Closing in favor of https://github.com/angular/angular-cli/pull/20655

PR #17295, opened 2020-03-24
fix(@ngtools/webpack): provide tsconfig path to ngcc

alan-agius4:


alan-agius4:
Blocked on https://github.com/angular/angular/pull/36180 and next FW release

alan-agius4:
//cc @petebacondarwin

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -28,26 +28,17 @@ export class NgccProcessor {
...
     this._logger = new NgccLogger(this.compilationWarnings, this.compilationErrors);
     this._nodeModulesDirectory = this.findNodeModulesDirectory(this.basePath);
-
-    const { baseUrl, paths } = this.compilerOptions;
-    if (baseUrl && paths) {
-      this._pathMappings = {
-        baseUrl,
-        paths,
-      };
-    }
```
petebacondarwin: This removal is only acceptable if my PR get into 9.1.0

packages/angular_devkit/build_angular/package.json
```diff
@@ -101,7 +101,7 @@
     "zone.js": "^0.10.0"
   },
   "peerDependencies": {
-    "@angular/compiler-cli": ">=9.0.0 < 10",
+    "@angular/compiler-cli": ">=9.1.0 < 10",
```
filipesilva: Does this mean that the 9.0.x CLI packages we have are incompatible with FW 9.1.x?

What happens when a user updates CLI but not FW, and vice versa?

alan-agius4: - CLI 9.0 is compatible with FW 9.1.
- FW 9.1 is compatible with CLI 9.0
- CLI 9.1 is NOT compatible with FW 9.0 (That's because this API/CLI options) are only available in 9.1

filipesilva: Can we then detect on the CLI side if we should use the new options or not? If we can't use the new options, then we should continue using the 9.0 ones.

The alternative would be to error out, but given our support policy I don't think that's a real option.

petebacondarwin: If you kept the `pathMappings` stuff in there, then it would be compatible I think, at least for the sync integration. For the async integration, ngcc might complain about an unknown `--tsconfig` option.

alan-agius4: Umm, the strict unknown option handling was added in 9.1, so we should be fine.

filipesilva: LGTM, thanks for these changes!

PR #17296, opened 2020-03-24
ci: fix version 7 LTS branch CI

alan-agius4:


alan-agius4:
Caretaker: the `ci/angular: merge status` check will not go green because of missing CI steps in LTS branch

PR #17316, opened 2020-03-26
feat: add @types/webpack-env

xiaoxiangmoe:


xiaoxiangmoe:
It looks like we can't delete `declare const require:` because of compatibility issues with older versions.

clydin:
Thank you for the contribution.  However, not adding additional project level dependencies was an intentional choice in this case ([PR](https://github.com/angular/angular-cli/pull/16771)).
This added package is also Webpack specific which is something that is generally avoided at the public interface level as Webpack is primarily considered an internal implementation detail.
Ideally, the usage of Webpack specific functionality in these files should eventually be removed but this may require larger changes in design and implementation to accomplish.

PR #17337, opened 2020-03-30
fix(@angular-devkit/build-angular): don't show commonjs usage for web…

alan-agius4:
…pack-dev-server utils


filipesilva:
Does this also cover the hot reload entry points?

https://github.com/angular/angular-cli/blob/1e258317b1f6ec1e957ee3559cc3b28ba602f3ba/packages/angular_devkit/build_angular/src/dev-server/index.ts#L605-L638

alan-agius4:
@filipesilva no, added!

PR #17341, opened 2020-03-30
Log stack trace on error

soyuka:
As it's done on lib/cli/index.js, very useful in case of errors as the message is almost never sufficient.

### DIFF
--- a/packages/angular/cli/commands/analytics-impl.ts
+++ b/packages/angular/cli/commands/analytics-impl.ts
@@ -87,6 +87,9 @@ export class AnalyticsCommand extends Command<AnalyticsCommandSchema> {
       }
     } catch (err) {
       this.logger.fatal(err.message);
+      if (err.stack) {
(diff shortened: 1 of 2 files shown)


soyuka:
@googlebot I signed it!

alan-agius4:
Hi @soyuka, thanks for your contribution however it is not desired to print stack traces in the console for a number of reasons.

Here are some;
* The real error message might get lost among the stacktrace
* Error messages alone should be clear and actionable.

In case you encounter an error message which is not actionable kindly file an issue.

Thanks.

soyuka:
I'm sorry but I disagree, at least give a `verbose` option that allows the stack trace to be printed. When working on specific schematics for example, an error can happen and it's impossible to debug without a proper stack trace. Stack traces are printed in the CLI: https://github.com/angular/angular-cli/blob/master/packages/angular/cli/lib/cli/index.ts#L96 and this behavior should be consistent.

PR #17361, opened 2020-04-01
feat(@schematics/angular): evergreen new applications

alan-agius4:
By default, we now generate new applications which support only evergreen browsers, as a result differential loading is now opt-in.

A new flag `--legacy-browsers`, was added to generate applications which support non evergreen browsers such as Internet Explorer 11.

**Note**: After an application is generated opting in and out of differential loading is still possible through the supported browsers configuration in the browserslist configuration file  defaulted to `.browserslistrc`.

**Upcoming others PRs**
- Docs will be updated in a separate PR in Angular/Angular TOOL-1370
- Change script target from ES2015 and ES2016 TOOL-1371
 
Reference TOOL-1364


packages/schematics/angular/application/files/.browserslistrc.template
```diff
@@ -5,8 +5,11 @@
 # You can see what browsers were selected by your queries by running:
 #   npx browserslist
 
-> 0.5%
-last 2 versions
-Firefox ESR
-not dead
-not IE 9-11 # For IE 9-11 support, remove 'not'.
\ No newline at end of file
+last 1 Chrome version
```
clydin, line 8: Can you add a link to the AIO browser support page to the file as well?

alan-agius4: Thanks a good Idea!

alan-agius4:
Blocking until we get more information from @atscott about dropping IE9-10 support in FW.

packages/angular_devkit/build_angular/src/utils/build-browser-features_spec.ts
```diff
@@ -29,9 +29,9 @@ describe('BuildBrowserFeatures', () => {
   afterEach(async () => host.restore().toPromise());
 
   describe('isDifferentialLoadingNeeded', () => {
-    it('should be true for for IE 9-11 and ES2015', () => {
+    it('should be true for for IE 11 and ES2015', () => {
```
clydin, line 32:
```suggestion
    it('should be true for IE 11 and ES2015', () => {
```

packages/schematics/angular/application/files/.browserslistrc.template
```diff
@@ -2,11 +2,17 @@
...
+last 1 Firefox version
+last 2 Edge major versions
+last 2 Safari major version
+last 2 iOS major versions
+Firefox ESR<% if (legacyBrowsers) { %>
+IE 11<% } else { %>
```
IgorMinar, line 17: I suggest that you change this to include IE9 and 10 for now to unblock this PR. And once we have a decision on IE9/10 and if the decision is positive then you can send a small PR to drop those two browsers from here.

alan-agius4: Okay, let's do that.

packages/angular_devkit/build_angular/src/utils/build-browser-features_spec.ts
alan-agius4: done

packages/angular_devkit/build_angular/src/utils/build-browser-features_spec.ts
```diff
@@ -64,36 +64,12 @@ describe('BuildBrowserFeatures', () => {
...
-
-      const buildBrowserFeatures = new BuildBrowserFeatures(
-        workspaceRootSysPath,
-        ScriptTarget.ES2015,
-      );
-      expect(buildBrowserFeatures.isFeatureSupported('es6-module')).toBe(true);
-    });
-
-    it('should be false for es6-module and IE9', () => {
+    it('should be false for es6-module and 11', () => {
```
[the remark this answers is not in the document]
alan-agius4: done

Issue #17320, opened 2020-03-26
bug(moment-adapter): "_rollupMoment__default is not defined"


kaito3desuyo:
(related with https://github.com/angular/angular/issues/36248)

#### Reproduction

https://github.com/kaito3desuyo/angular-test

Steps to reproduce:
1. clone this repository
2. ``ng build --prod``
3. ``npm i -g http-server``
4. move root directory to ``dist/angular-test`` Folder
5. ``http-server``
6. open ``http://localhost:8080``

This issue only occurs with production builds.
 
#### Expected Behavior

Screen is displayed correctly

#### Actual Behavior

Screen goes blank with console error

#### Environment

- Angular: 9.1.0
- CDK/Material: 9.2.0
- Browser(s): Chrome 80.0.3987.149
- Operating System (e.g. Windows, macOS, Ubuntu): Windows 10 x64 1909 18363.720

marsc:
Same problem here in production build  only with "@angular-devkit/build-angular": "~0.901.0". Downgraded for now.

ArtemeeSenin:
Have the same problem but with 
`main-es2015.8a5dc9b260c775d8aa4b.js:1 Uncaught ReferenceError: dragulaExpt__default is not defined`

TheParad0X:
Same problem here after an update from Angular 9.0.5 to the latest

mikejunt:
Yep, also seeing this issue with moment after upgrading

atlesj:
Same

fpmk:
any news?

pabloxn:
Same problem

deepakgonda:
Same problem with the production build.

acherniak:
Any good news?

jasoninrapid:
We're seeing the same problem on an Angular 8 to Angular 9 upgrade we did on March 26.  Details:

- ng2-dragular module got  transpiled with the same bad "dragulaExpt__default" referenced above.
- Downgrading angular devkit fix it for us.  Our (fixed) entry in the dev dependencies in package.json looks like: "@angular-devkit/build-angular": "0.900.7"

spock123:
I think most of us are stuck on 9.0.x for now. I can't risk workarounds or downgraded dependencies. Better to wait until this gets sorted out - at least for me.

demisx:
[...]

PR #17364, opened 2020-04-01
fix(@angular-devkit/build-optimizer): use TypeScript 3.6

clydin:
When used within the build optimizer, TypeScript 3.8 will consider certain imports to be unreferenced and errantly remove them even though they are used within the module.  This results in runtime errors during application load.  TypeScript 3.7 cannot be used due to it causing a build time error when used within the build optimizer.  The build optimizer uses an independent version of TypeScript which allows it to use TypeScript 3.6 while applications continue to use TypeScript 3.7 or 3.8 to build the actual application.

Fixes: #17320

### DIFF
--- a/packages/angular_devkit/build_optimizer/package.json
+++ b/packages/angular_devkit/build_optimizer/package.json
@@ -12,7 +12,7 @@
     "loader-utils": "2.0.0",
     "source-map": "0.7.3",
     "tslib": "1.11.1",
-    "typescript": "3.8.3",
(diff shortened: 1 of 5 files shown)


clydin:
TS 3.9 exhibits the same problem.
The issue has not yet been reported upstream.  Work is ongoing to produce a minimal reproduction.

alan-agius4:
Managed to fix the root cause and reported opened an issue upstream: https://github.com/microsoft/TypeScript/issues/38412

PR #17367, opened 2020-04-02
feat(@schematcs/angular): change `browserslist` file name into `.browserslistrc`

alan-agius4:
Closes: #15961


packages/schematics/angular/migrations/update-10/rename-browserslist-config.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { Path, join } from '@angular-devkit/core';
+import { DirEntry, Rule } from '@angular-devkit/schematics';
+
+function visit(directory: DirEntry): Path[] {
```
clydin, line 11: nit: good use case for a generator

alan-agius4: Indeed, it's a good use case. I updated the code to use a generator, can you please take a look?

If it's good to be merge please add the merge label. Thanks.

PR #17372, opened 2020-04-02
feat(@schematics/angular): enable stricter type checking and optimization effective coding rules 

alan-agius4:
With this change we enable stricter type checking and optimization effective coding rules when using the `--strict` option.

Changes in schematics
- `ng-new`: A prompt for the `strict` option was added. This option is a proxy and will be passed to the application and workspace schematics.
- `application`: A `package.json` was added in the `app` folder, to tell the bundlers whether the application is free from side-effect code. When `strict` is `true`. the `sideEffects` will be set `false`.
- `workspace` When `strict` is true, we add stricter TypeScript and Angular type-checking options.

Note: AIO is already using these strict TypeScript compiler settings. PR to enable `strictTemplates`  https://github.com/angular/angular/pull/36391

Reference: TOOL-1366

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -179,6 +179,34 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
   }
 
   const sourceRoot = join(normalize(projectRoot), 'src');
+  let budgets = [];
+  if (options.strict) {
+    budgets = [
+      {
+        type: 'initial',
+        maximumWarning: '500kb',
+        maximumError: '1mb',
+      },
+      {
+        type: 'anyComponentStyle',
+        maximumWarning: '2kb',
+        maximumError: '4kb',
+      },
+    ];
+  } else {
+    budgets = [
+      {
+        type: 'initial',
+        maximumWarning: '2mb',
+        maximumError: '5mb',
+      },
+      {
+        type: 'anyComponentStyle',
+        maximumWarning: '6kb',
+        maximumError: '10kb',
+      },
+    ];
+  }
 
   const project = {
     root: normalize(projectRoot),
@@ -219,17 +247,7 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
             extractLicenses: true,
             vendorChunk: false,
             buildOptimizer: true,
-            budgets: [
-            {
-              type: 'initial',
-              maximumWarning: '2mb',
-              maximumError: '5mb',
-            },
-            {
-              type: 'anyComponentStyle',
-              maximumWarning: '6kb',
-              maximumError: '10kb',
-            }],
+            budgets,
           },
         },
       },
--- a/packages/schematics/angular/application/other-files/package.json.template
+++ b/packages/schematics/angular/application/other-files/package.json.template
@@ -1,0 +1,6 @@
+{
+  "name": "<%= utils.dasherize(name) %>",
+  "private": true,
+  "description": "This is a special package.json file that is not used by package managers. It is however used to tell the tools and bundlers whether the code under this directory is free of code with non-local side-effect. Any code that does have non-local side-effects can't be well optimized (tree-shaken) and will result in unnecessary increased payload size. It should be safe to set this option to 'false' for new applications, but existing code bases could be broken when built with the production config if the application code does contain non-local side-effects that the application depends on.",
+  "sideEffects": <%= !strict %>
+}
--- a/packages/schematics/angular/ng-new/schema.json
+++ b/packages/schematics/angular/ng-new/schema.json
@@ -118,20 +118,21 @@
       "x-user-analytics": 12
     },
     "createApplication": {
-      "description": "When true (the default), creates a new initial app project in the src folder of the new workspace. When false, creates an empty workspace with no initial app. You can then use the generate application command so that all apps are created in the projects folder.",
+      "description": "When true (the default), creates a new initial application project in the src folder of the new workspace. When false, creates an empty workspace with no initial app. You can then use the generate application command so that all apps are created in the projects folder.",
       "type": "boolean",
       "default": true
     },
     "minimal": {
-      "description": "When true, creates a project without any testing frameworks. (Use for learning purposes only.)",
+      "description": "When true, creates a workspace without any testing frameworks. (Use for learning purposes only.)",
       "type": "boolean",
       "default": false,
       "x-user-analytics": 14
     },
     "strict": {
-      "description": "Creates a workspace with stricter TypeScript compiler options.",
+      "description": "Creates a workspace with stricter type checking and build optimization options.",
       "type": "boolean",
-      "default": false
+      "default": false,
+      "x-prompt": "Create a workspace with stricter type checking and more efficient production optimizations?"
     },
     "legacyBrowsers": {
       "type": "boolean",
--- a/packages/schematics/angular/workspace/files/tsconfig.json.template
+++ b/packages/schematics/angular/workspace/files/tsconfig.json.template
@@ -3,11 +3,10 @@
   "compilerOptions": {
     "baseUrl": "./",
     "outDir": "./dist/out-tsc",<% if (strict) { %>
-    "noImplicitAny": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
     "noImplicitReturns": true,
-    "noImplicitThis": true,
-    "noFallthroughCasesInSwitch": true,
-    "strictNullChecks": true,<% } %>
+    "noFallthroughCasesInSwitch": true,<% } %>
     "sourceMap": true,
     "declaration": false,
     "downlevelIteration": true,
@@ -20,9 +19,9 @@
       "es2018",
       "dom"
     ]
-  },
+  }<% if (strict) { %>,
   "angularCompilerOptions": {
-    "fullTemplateTypeCheck": true,
-    "strictInjectionParameters": true
-  }
+    "strictInjectionParameters": true,
+    "strictTemplates": true
+  }<% } %>
 }
--- a/packages/schematics/angular/workspace/schema.json
+++ b/packages/schematics/angular/workspace/schema.json
@@ -33,7 +33,7 @@
       "x-user-analytics": 14
     },
     "strict": {
-      "description": "Creates a workspace with stricter TypeScript compiler options.",
+      "description": "Creates a workspace with stricter type checking options.",
       "type": "boolean",
       "default": false
     },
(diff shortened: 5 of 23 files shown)


packages/schematics/angular/workspace/files/tsconfig.json.template
```diff
@@ -3,11 +3,8 @@
   "compilerOptions": {
     "baseUrl": "./",
     "outDir": "./dist/out-tsc",<% if (strict) { %>
-    "noImplicitAny": true,
-    "noImplicitReturns": true,
-    "noImplicitThis": true,
-    "noFallthroughCasesInSwitch": true,
```
clydin, line 9: `noFallthroughCasesInSwitch` should be kept.  It is not part of the `strict` option.

alan-agius4: Right!, fixed.

packages/schematics/angular/workspace/files/tsconfig.json.template
```diff
@@ -20,9 +19,10 @@
       "es2018",
       "dom"
     ]
-  },
+  }<% if (strict) { %>,
   "angularCompilerOptions": {
     "fullTemplateTypeCheck": true,
```
clydin, line 24: `fullTemplateTypeCheck` can be removed since `strictTemplates` supersedes it.

packages/schematics/angular/workspace/schema.json
```diff
@@ -33,9 +33,9 @@
       "x-user-analytics": 14
     },
     "strict": {
-      "description": "Creates a workspace with stricter TypeScript compiler options.",
+      "description": "Creates a workspace with stricter TypeScript and Angular compiler options.",
       "type": "boolean",
-      "default": false
+      "default": true
```
clydin, line 38: Should we consider deprecating this option for 10?

alan-agius4: I think that might be a good idea.

@IgorMinar, thoughts?

IgorMinar: let's write up a doc proposing that we enable the `strict` mode in CLI projects

The doc should contain at least these:

- the current default
- what is strict mode
- the impact of this change on
  - new projects
  - existing projects (what's the migration path?)
  - existing ecosystem
    - libraries
    - tutorial / blog posts/ documentation
    - workshop materials
- the proposed rollout plan

once we have all of this, we can review it, and then share it with GDEs, and the community for feedback to validate that our plan is solid before we actually change the default in the CLI.

packages/schematics/angular/application/other-files/package.json.template
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "<%= utils.dasherize(name) %>",
+  "private": true,
+  "description": "This is a fake package.json file which is not used by package managers. It is used to tell the bundlers whether the application is free from side-effect code.",
+  "sideEffects": <%= !strict %>
```
clydin, line 5: Technically only need the `sideEffects` field here.  I think the description is useful but I'm undecided about the others.

alan-agius4: I dont mind removing the others.

alan-agius4: @IgorMinar thoughts?

packages/schematics/angular/application/other-files/package.json.template
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "<%= utils.dasherize(name) %>",
+  "private": true,
+  "description": "This is a fake package.json file which is not used by package managers. It is used to tell the bundlers whether the application is free from side-effect code.",
```
IgorMinar, line 4:
```suggestion
  "description": "This is a special package.json file that is not used by package managers. It is however used to tell the tools and bundlers whether the code under this directory is free of code with non-local side-effect. Any code that does have non-local side-effects can't be well optimized (tree-shaken) and will result in unnecessary increased payload size. It should be safe to set this option to `false` for new applications, but existing code bases could be broken when built with the production config if the application code does contain non-local side-effects that the application depends on.",
```

packages/schematics/angular/application/other-files/package.json.template
IgorMinar: I think we should keep all otherwise nobody will understand what this package.json is here for and people will remove it because they'll think that it's junk.

packages/schematics/angular/ng-new/schema.json
```diff
@@ -118,20 +118,21 @@
...
       "default": false,
       "x-user-analytics": 14
     },
     "strict": {
-      "description": "Creates a workspace with stricter TypeScript compiler options.",
+      "description": "Creates a workspace with stricter type checking options.",
       "type": "boolean",
-      "default": false
+      "default": false,
+      "x-prompt": "Create a workspace with stricter type checking and optimization effective coding rules?"
```
IgorMinar, line 135:
```suggestion
      "x-prompt": "Create a workspace with stricter type checking and more efficient production optimizations?"
```

packages/schematics/angular/workspace/files/tsconfig.json.template
```diff
@@ -3,11 +3,11 @@
...
     "baseUrl": "./",
     "outDir": "./dist/out-tsc",<% if (strict) { %>
-    "noImplicitAny": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
     "noImplicitReturns": true,
-    "noImplicitThis": true,
     "noFallthroughCasesInSwitch": true,
-    "strictNullChecks": true,<% } %>
+    "strictPropertyInitialization": false,<% } %>
```
IgorMinar, line 10: why? Is this because of @Components with @Inputs?

I do think that developer's should provide a default value. That's what I do and it works quite well.

I recommend removing this line.

packages/schematics/angular/workspace/files/tsconfig.json.template
```diff
@@ -20,9 +20,9 @@
       "es2018",
```
IgorMinar, line 20: I thought we agreed that we'd update the `"module"` to `"es2020"` and `"target"` to "`es2016`" as well.

Did changing these settings break stuff?

IgorMinar: or is that a separate PR?

packages/schematics/angular/ng-new/schema.json
```diff
@@ -118,20 +118,21 @@
...
     "minimal": {
-      "description": "When true, creates a project without any testing frameworks. (Use for learning purposes only.)",
+      "description": "When true, creates a workspace without any testing frameworks. (Use for learning purposes only.)",
       "type": "boolean",
       "default": false,
       "x-user-analytics": 14
     },
     "strict": {
-      "description": "Creates a workspace with stricter TypeScript compiler options.",
+      "description": "Creates a workspace with stricter type checking options.",
```
IgorMinar, line 132:
```suggestion
      "description": "Creates a workspace with stricter type checking and build optimization options.",
```

packages/schematics/angular/workspace/schema.json
```diff
@@ -33,7 +33,7 @@
       "x-user-analytics": 14
     },
     "strict": {
-      "description": "Creates a workspace with stricter TypeScript compiler options.",
+      "description": "Creates a workspace with stricter type checking options.",
```
IgorMinar, line 36:
```suggestion
      "description": "Creates a workspace with stricter type checking and build optimization options.",

```

alan-agius4:
@IgorMinar, yes I skipped that intentionally as I wanted to follow up with a couple of other PRs to address the bundle budgets and module target.

packages/schematics/angular/workspace/files/tsconfig.json.template
alan-agius4: That will be in a separate PR, which I’ll will be working on in the coming days. Reason being is that changing the module and target is a big change as it changes a lot of files, and I prefer to keep PRs as small as possible, even for reviewers.

Also, the ever green apps changes conflicted with these changes since it touches almost the same files, so I preferred to land that first.

alan-agius4: Ever green apps PR is finally merged 🥳

packages/schematics/angular/workspace/files/tsconfig.json.template
alan-agius4: Yeah that’s the reason. 

I’ll also do a PR In AIO to remove that.

IgorMinar:
@alan-agius4 ok. Sgtm

packages/schematics/angular/workspace/schema.json
alan-agius4: I prefer to leave this as original, as workspace doesn't have any build settings.

- `strict` in application enables striker build optimisations rules 
- `strict` in a workspace enables stricter type checking for all projects

packages/schematics/angular/workspace/files/tsconfig.json.template
alan-agius4: So for AIO, we better do a followup task, as there a quite a lot of errors when I removed `"strictPropertyInitialization": false`

alan-agius4:
@IgorMinar, FYI I actually pushed the budget change in the PR.

packages/schematics/angular/application/index.ts
```diff
@@ -179,6 +179,34 @@ function addAppToWorkspaceFile(options: ApplicationOptions, appDir: string): Rul
   }
 
   const sourceRoot = join(normalize(projectRoot), 'src');
+  let budgets = [];
+  if (options.strict) {
+    budgets = [
+      {
+        type: 'initial',
+        maximumWarning: '400kb',
```
IgorMinar, line 187: Let's set this to 500kb in v10 just so that we are not too Draconian too quickly. We'll need to do a bit more testing before we know how low can we go given the state of the library ecosystem.

E.g. angular.io is at ~460kb today and that's a highly optimized app which is not the norm today.

In v11 and beyond we can drop the budget to 400kb or even lower once we have more data/experience of what that looks like in the real world.

packages/schematics/angular/workspace/schema.json
IgorMinar: I'm fine with keeping the original.

packages/schematics/angular/application/index.ts
alan-agius4: Agreed and updated.

dgp1130:
@alan-agius4, looks like this was the unlucky one again that hit a merge conflict today.

alan-agius4:
@dgp1130, conflict solved.

PR #17406, opened 2020-04-06
build: default to wombat proxy for publishing

josephperrott:
If no `--registry` argument is provided when calling to publish
use the Wombat proxy.  Additionally, updates the release process
documentation to instruct usage of the Wombat proxy.

### DIFF
--- a/scripts/publish.ts
+++ b/scripts/publish.ts
@@ -116,9 +116,9 @@ export default async function (args: PublishArgs, logger: logging.Logger) {
         if (args.tag) {
           publishArgs.push('--tag', args.tag);
         }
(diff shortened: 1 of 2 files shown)


scripts/publish.ts
```diff
@@ -118,6 +118,9 @@ export default async function (args: PublishArgs, logger: logging.Logger) {
         }
         if (args.registry) {
           publishArgs.push('--registry', args.registry);
```
alan-agius4, line 120: Nit: 

```ts
publishArgs.push('--registry', args.registry || 'https://wombat-dressing-room.appspot.com');
```

PR #17424, opened 2020-04-08
build: version from git tag which consists two or more digits

mgechev:
Currently, we can't publish v10 because the `--match` pattern in describe can't match the tag `v10.0.0`. The new glob has the following semantics:

```
v[0-9]*.[0-9]*.[0-9]*
# start with a "v" and a number
# followed by 0 or more characters
# followed by a "." and a number
# followed by 0 or more characters
# followed by a "." and a number
# followed by 0 or more characters
```

Couldn't make it stricter because didn't find an easy way to enable extglobs for git for everyone.

### DIFF
--- a/lib/packages.ts
+++ b/lib/packages.ts
@@ -164,7 +164,7 @@ function _getVersionFromGit(experimental: boolean): string {
   }
 
   const hasLocalChanges = _exec(`git status --porcelain`) != '';
-  const scmVersionTagRaw = _exec(`git describe --match v[0-9].[0-9].[0-9]* --abbrev=7 --tags`)
+  const scmVersionTagRaw = _exec(`git describe --match v[0-9]*.[0-9]*.[0-9]* --abbrev=7 --tags`)
     .slice(1);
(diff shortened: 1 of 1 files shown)


lib/packages.ts
```diff
@@ -164,7 +164,7 @@ function _getVersionFromGit(experimental: boolean): string {
   }
 
   const hasLocalChanges = _exec(`git status --porcelain`) != '';
-  const scmVersionTagRaw = _exec(`git describe --match v[0-9].[0-9].[0-9]* --abbrev=7 --tags`)
+  const scmVersionTagRaw = _exec(`git describe --match v[0-9]*.[0-9]*.[0-9]* --abbrev=7 --tags`)
```
kyliau, line 167: I think technically it should be `[0-9]+`, i.e. "one or more digits", but it's not really a big deal.

josephperrott: `git describe`s `--match` flag is a glob rather than a regex, so a `+` is actually not what we are looking for.  We ran into this on framework as well.

mgechev: Yes, `+` does not work in glob and `extglobs` doesn't seem like an option.

PR #17441, opened 2020-04-10
fix(@ngtools/webpack): give higher priority to ivy-specific entry-points

gkalpak:
When Ivy is enabled in an app, the app's dependencies are processed by ngcc (the Angular compatibility compiler), which will generate new Ivy-compatible entry-points in each package. To allow webpack find those entry-points, it will add references to them in `package.json`, named after the original property with the `_ivy_ngcc` suffix. So, for example, `es2015` and `module` will become `es2015_ivy_ngcc` and `module_ivy_ngcc`.

Previously, the `mainFields` passed to webpack would give higher priority to an Ivy entry-point compared to the corresponding non-ivy entry-point but not compared to other entry-points. For example, `es2015, module` would becode `es2015_ivy_ngcc, es2015, module_ivy_ngcc, module`. This could mean that `es2015` would be preferred over `module_ivy_ngcc`, even though the former is probably not Ivy-ready. Generally, if `es2015` exists and has a higher priority than `module`, then `es2015_ivy_ngcc` would be generated before (or instead of) `module_ivy_ngcc`, but this can be changed using an ngcc configuration file (`ngcc.config.js`).

This commit fixes this by ensuring that any `_ivy_ngcc` entry-point is considered before any of the original entry-points in the `mainFields` list.

NOTE:
While it is possible that a format is Ivy-compatible without bhaving an `_ivy_ngcc` suffix (e.g. if the user does a manual ngcc pass without `--create-ivy-entry-points`), it is quite unlikely to happen and even if it does, choosing a different format should be OK.

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -911,18 +911,18 @@ export class AngularCompilerPlugin {
         // When Ivy is enabled we need to add the fields added by NGCC
         // to take precedence over the provided mainFields.
         // NGCC adds fields in package.json suffixed with '_ivy_ngcc'
-        // Example: module -> module__ivy_ngcc
+        // Example: module -> module_ivy_ngcc
         // tslint:disable-next-line:no-any
         (compiler as any).resolverFactory.hooks.resolveOptions
           .for('normal')
           // tslint:disable-next-line:no-any
           .tap('WebpackOptionsApply', (resolveOptions: any) => {
-            const mainFields = (resolveOptions.mainFields as string[])
-              .map(f => [`${f}_ivy_ngcc`, f]);
+            const originalMainFields: string[] = resolveOptions.mainFields;
+            const ivyMainFields = originalMainFields.map(f => `${f}_ivy_ngcc`);
 
             return {
               ...resolveOptions,
-              mainFields: flattenArray(mainFields),
+              mainFields: [...ivyMainFields, ...originalMainFields],
             };
           });
       }


packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -911,18 +911,18 @@ export class AngularCompilerPlugin {
...
-        // Example: module -> module__ivy_ngcc
+        // Example: module -> module_ivy_ngcc
         // tslint:disable-next-line:no-any
         (compiler as any).resolverFactory.hooks.resolveOptions
           .for('normal')
           // tslint:disable-next-line:no-any
           .tap('WebpackOptionsApply', (resolveOptions: any) => {
-            const mainFields = (resolveOptions.mainFields as string[])
-              .map(f => [`${f}_ivy_ngcc`, f]);
+            const originalMainFields: string[] = resolveOptions.mainFiels;
```
clydin, line 920:
```suggestion
            const originalMainFields: string[] = resolveOptions.mainFields;
```

gkalpak: 🤦‍♂️

petebacondarwin: The dangers of untyped code :-)

packages/ngtools/webpack/src/angular_compiler_plugin.ts
```diff
@@ -911,18 +911,18 @@ export class AngularCompilerPlugin {
...
+        // Example: module -> module_ivy_ngcc
         // tslint:disable-next-line:no-any
         (compiler as any).resolverFactory.hooks.resolveOptions
           .for('normal')
           // tslint:disable-next-line:no-any
           .tap('WebpackOptionsApply', (resolveOptions: any) => {
-            const mainFields = (resolveOptions.mainFields as string[])
-              .map(f => [`${f}_ivy_ngcc`, f]);
+            const originalMainFields: string[] = resolveOptions.mainFields;
+            const ivyMainFields = originalMainFields.map(f => `${f}_ivy_ngcc`);
```
alan-agius4, line 921: I am not convinced with this change, this might have the side effect that when not an entrypoint is not processed by NGCC for a reason or another a wrong entrypoint will get added in the compilation.

Example, the UMD bundle was successfully converted with NGCC but the FESM5 wasn't so we'll end up with the 
following;
```
module_ivy_ngcc, main_ivy_ngcc, module, main
```

In this case since module was not converted, the UMD Ivy version of the bundle will be included instead which is not desired and expected and user's will only notice this if they due due to increase in bundle size.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
clydin: The webpack typings are lacking in certain areas unfortunately.

alan-agius4:
Why question would be why would `module` be converted with NGCC and `es2015` wouldn't?

> if es2015 exists and has a higher priority than module, then es2015_ivy_ngcc would be generated before (or instead of) module_ivy_ngcc

I am not following why `es2015` should be generated instead of `module`. They are 2 different formats and used differently in the CLI based on which compilation target the user has set and are not always interchangeable.

packages/ngtools/webpack/src/angular_compiler_plugin.ts
gkalpak: Why do you think it would be better to include the unprocessed FESM5 bundle?

alan-agius4: Cause it will error during run time. So at least the users will know that something went wrong.

gkalpak:
`es2015` and `module` were just an example. Typically, CLI uses `compileAllFormats: false`, which means that it will pass ngcc a list of format properties in order of preference and ngcc will process the first that is finds for each package.

So, if CLI passes `es2015, browser, module, main` ngcc will process the first that it finds. This means that if a package has all format properties in its `package.json`.

##
However, it is possible that an app developer or library author includes an `ngcc.config.js` file that instructs ngcc to ignore a format property. For example, say that a library's `es2015` format is in some way incompatible with ngcc and trying to process it will result in a build error. The developer can work-around that by instructing ngcc to ignore the `es2015` format property, so then ngcc will move to the second-best property (e.g. `browser`).

In that case, we want webpack to also pick up `browser_ivy_ngcc` (which is Ivy-ready) and not the (non-Ivy-ready) `es2015`.

##
BTW, this usecase is more likely to come up with the `main` property, where the formats can vary a lot. So, it happens that the `main` property of some (often old and unmaintained but still used) libraries cannot be processed by ngcc (see for example angular/angular#36558). This is problematic when building for SSR, since `main` is the first preference.

One way to fix this would be for the app developer to configure ngcc to ignore the `main` property and thus fal back to the second-best: `module`
In that case, we want webpack to pick `module_ivy_ngcc` and not the (unprocessed) `main` property.

Does that make sense?

packages/ngtools/webpack/src/angular_compiler_plugin.ts
gkalpak: But this will usually mean that the entry-point was explicitly ignored in ngcc, so it is expected that the entry-point was not converted. See https://github.com/angular/angular-cli/pull/17441#issuecomment-612098944.

alan-agius4:
Make sense.

> BTW, this usecase is more likely to come up with the main property, where the formats can vary a lot. So, it happens that the main property of some (often old and unmaintained but still used) libraries cannot be processed by ngcc (see for example angular/angular#36558). This is problematic when building for SSR, since main is the first preference.

In most cases at least since version 8, the SSR's preferred entrypoint is es2015 and not main.

gkalpak:
> In most cases at least since version 8, the SSR's preferred entrypoint is es2015 and not main.

Oh, I didn't know that. That's great!
<sub>(`main` can still be an issue with old, unmaintained libaries, since many of them do not have `es2015` bundles.)</sub>

dgp1130:

PR #17457, opened 2020-04-13
test: update ng-packagr tests tsconfig as per CLI default

alan-agius4:
This seems to be the root cause of the failures in the following PR:
https://github.com/angular/angular-cli/pull/17455


dgp1130:

PR #17463, opened 2020-04-13
fix(@angular/cli): ensure Node.js message is shown on old Node.js versions

clydin:
Removes errant comma within the message log call and allows the process to exit naturally to better ensure that it is written to the console.


dgp1130:

PR #17496, opened 2020-04-17
docs: add notes about various release foot-guns

dgp1130:
This is a few different edits to make caretakers more aware of problems that can occur during a release.
* Added a comment to reinforce that pushing tags needs to happen with the push of the release commit. Otherwise CI can fail `npm install` because the relevant tags are not set.
* Changed to check out the publish branch rather than the tag because checking out a tag fails the subsequent `publish` command.
* Added `yarn` before the `publish` command because dependencies may be out of date and cause errors.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -112,6 +112,9 @@ As commits are cherry-picked when PRs are merged, creating the release should be
 ```bash
 git commit -a -m 'release: vXX'
 git tag 'vXX'
+
+# Make sure to run these commands together, as missing tags can cause CI
+# failures.
 git push upstream && git push upstream --tags
 ```
 
@@ -138,29 +141,31 @@ After closing the tab, you have successfully logged in, it is time to publish.
 
 **It is a good idea to wait for CI to be green on the patch branch and tag before doing the release.**
 
-Check out the patch tag (e.g. `v6.7.8`), then run:
-```sh
-devkit-admin publish
+Check out the patch branch (e.g. `9.1.x`), then run:
+```bash
+yarn # Reload dependencies
+yarn admin publish
 ```
 
-Check out the minor tag (e.g. `v6.8.0-beta.0`), then run:
+If also publishing a prerelease, check out `master`, then run:
 ```bash
-devkit-admin publish --tag next
+yarn # Reload dependencies
+yarn admin publish --tag next
 ```
 
 ### Release Notes
 
-`devkit-admin changelog` takes `from` and `to` arguments which are any valid git
+`yarn run -s admin changelog` takes `from` and `to` arguments which are any valid git
 ref.
 
 For example, running the following command will output the release notes on
 stdout between v1.2.3 and 1.2.4:
 
 ```bash
-devkit-admin changelog --from=v1.2.3 --to=v1.2.4
+yarn run -s admin changelog --from=v1.2.3 --to=v1.2.4
 ```
 
-Copy the output (you can use `| pbcopy` on MacOS or `|xclip` on Linux) and
+Copy the output (you can use `| pbcopy` on MacOS or `| xclip` on Linux) and
 paste the release notes on [GitHub](https://github.com/angular/angular-cli/releases)
 for the tag just released.
 


docs/process/release.md
```diff
@@ -112,6 +112,9 @@ As commits are cherry-picked when PRs are merged, creating the release should be
 ```bash
 git commit -a -m 'release: vXX'
 git tag 'vXX'
+
+# Make sure to run these commands together, as missing tags can cause CI
+# failures.
 git push upstream && git push upstream --tags
```
alan-agius4, line 118: Maybe better `git push upstream --follow-tags`?

dgp1130: Interesting, I wasn't aware of that flag. However it seems to only push *annotated* tags, which doesn't fit our current convention of `git tag 'vXXX'`. (Unless we want to switch to using annotated tags.)

> `--follow-tags` only pushes annotated tags.
> 
> ```
> git tag -a -m "I'm an annotation" <tagname>
> ```
> 
> That would be pushed (as opposed to `git tag <tagname>`, a lightweight tag, which would not be pushed, &lt;snip&gt;)

https://stackoverflow.com/a/3745250

Another answer suggests:

```
git push upstream tag vXXX
```

But I think that is not well-understood behavior and is probably less intuitive than `git push upstream && git push upstream --tags`.

https://stackoverflow.com/a/55882829

dgp1130:
Also @alan-agius4, I think this should only be merged to `master` right? Since it is just a doc?

alan-agius4:
@dgp1130, it can also be merged to patch since it’s not a feature.

dgp1130:
@alan-agius4, is the philosophy that anything which can go to `patch`, should go to `patch`, in order to minimize the difference between `master` and `patch`? Previously I've only send PRs to `patch` if they are actually _needed_ to be there so I'm just curious how this decision is actually made?

alan-agius4:
In general, I’d say yes.

docs/process/release.md
```diff
@@ -138,19 +141,21 @@ After closing the tab, you have successfully logged in, it is time to publish.
...
+yarn # Reload dependencies
+./bin/devkit-admin publish
 ```
 
-Check out the minor tag (e.g. `v6.8.0-beta.0`), then run:
+If also publishing a prerelease, check out `master`, then run:
 ```bash
-devkit-admin publish --tag next
+yarn # Reload dependencies
+./bin/devkit-admin publish --tag next
```
alan-agius4, line 153: Same as above

docs/process/release.md
```diff
@@ -138,19 +141,21 @@ After closing the tab, you have successfully logged in, it is time to publish.
...
 ```
 
 ### Release Notes
 
-`devkit-admin changelog` takes `from` and `to` arguments which are any valid git
+`./bin/devkit-admin changelog` takes `from` and `to` arguments which are any valid git
```
alan-agius4, line 158: Same

docs/process/release.md
```diff
@@ -138,19 +141,21 @@ After closing the tab, you have successfully logged in, it is time to publish.
 
 **It is a good idea to wait for CI to be green on the patch branch and tag before doing the release.**
 
-Check out the patch tag (e.g. `v6.7.8`), then run:
-```sh
-devkit-admin publish
+Check out the patch branch (e.g. `9.1.x`), then run:
+```bash
+yarn # Reload dependencies
+./bin/devkit-admin publish
```
alan-agius4, line 147: The `bin` here is without the leading `.`. 
Also, maybe better use;
`yarn admin publish`

dgp1130: I switched it to `yarn admin`, except for changelog generation, because we need to copy the output which will include:

```
yarn run v1.22.0
$ node ./bin/devkit-admin
```

If we use `yarn admin` rather than calling `devkit-admin` directly.

Is `devkit-admin` supposed to be installed on the path? I always just run it locally within the repo. Is there a different way I'm supposed to use this?

docs/process/release.md
dgp1130: Changed to `yarn admin`.

docs/process/release.md
dgp1130: Changed to `yarn admin`.

docs/process/release.md
alan-agius4: I personally always use `yarn admin`. I don't have `devkit-admin` installed in my path and always try to keep my paths as clean as possible.

docs/process/release.md
```diff
@@ -138,26 +141,28 @@ After closing the tab, you have successfully logged in, it is time to publish.
...
 ```bash
-devkit-admin publish --tag next
+yarn # Reload dependencies
+yarn admin publish --tag next
 ```
 
 ### Release Notes
 
-`devkit-admin changelog` takes `from` and `to` arguments which are any valid git
+`./bin/devkit-admin changelog` takes `from` and `to` arguments which are any valid git
```
kyliau, line 158: On line 147 and 153 we use `yarn admin`, but here it's different. Could you please make them consistent?

docs/process/release.md
```diff
@@ -138,26 +141,28 @@ After closing the tab, you have successfully logged in, it is time to publish.
...
 ref.
 
 For example, running the following command will output the release notes on
 stdout between v1.2.3 and 1.2.4:
 
 ```bash
-devkit-admin changelog --from=v1.2.3 --to=v1.2.4
+./bin/devkit-admin changelog --from=v1.2.3 --to=v1.2.4
```
kyliau, line 165: ditto

docs/process/release.md
dgp1130: The changelog output is typically copied, but running `yarn admin` also prints:

```
yarn run v1.22.0
$ node ./bin/devkit-admin
```

Which we don't want in the changelog, so we shouldn't use `yarn admin` here. I could switch everything back to `./bin/devkit-admin` for consistency if that's better.

kyliau: Perhaps we could silent the output with `yarn run -s admin`?

dgp1130: Done. Wasn't aware of that flag, thanks for pointing it out!

docs/process/release.md
dgp1130: Updated to `yarn run -s admin`.

docs/process/release.md
dgp1130: Keen pointed out that we can use `yarn run -s admin` to silence the extra text. Updated to use that format.

PR #17507, opened 2020-04-20
refactor(@angular-devkit/schematics): remove deprecated FileSystemHost

alan-agius4:
BREAKING CHANGE
Deprecated `FileSystemHost` has been removed. Use `Host` directly instead.

Note: this change only effects direct `@angular-devkit/schematics` users and not application developers.


clydin:
Could you update the commit message with the additional text?  The changelog will only pick up the commit message text.

alan-agius4:
@clydin seems like I forgot to push 😱.

Will do so in a couple of mins.

PR #17508, opened 2020-04-20
refactor(@schematics/angular): remove deprecated guard `spec` option

alan-agius4:
BREAKING CHANGE
Deprecated guard schematic option `spec` has been removed. Please use `skipTests` instead.


clydin:
Should we trigger the migration again for 10?

alan-agius4:
Good idea, just in case someone added this option manually

PR #17533, opened 2020-04-23
 ci: remove circle ci cache fallback

alan-agius4:
Cache fallback can be useful in some cases, but in our case it is hindering the CI performance as the cache is growing everyday which results in slower CI times because of the restoring and saving phases.

A fresh cache size is about 450Mb, while currently with fallbacking enabled it has grown to 3.8Gb

Below are some timings taken were we see that having cache fallbacking is doesn't help CI times:

| Caching                   | Cache Size | Restore Cache | Yarn Install | Save Cache | Total Time Elapsed |
|---------------------------|------------|---------------|--------------|------------|--------------------|
| Cache w/ Fallback (Miss)  | 3.81Gb     | 2m 50s        | 30s          | 7m 20s     | 10m 40s            |
| Cache w/ Fallback (Hit)   | 3.81Gb      | 2m 50s        | 20s          | 0          | 3m 10s             |
| Cache wo/ Fallback (Miss) | 0          | 0             | 40s          | 30s        | 1m 10s              |
| Cache wo/ Fallback (Hit)  | 458mb      | 20s           | 20s          | 0          | 40s                |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,14 +12,10 @@ version: 2.1
 # Variables
 
 ## IMPORTANT
-# If you change the cache key prefix, also sync the fallback_cache_key fallback to match.
-# Keep the static part of the cache key as prefix to enable correct fallbacks.
 # Windows needs its own cache key because binaries in node_modules are different.
 # See https://circleci.com/docs/2.0/caching/#restoring-cache for how prefixes work in CircleCI.
-var_1: &cache_key angular_devkit-0.12.0-{{ checksum "yarn.lock" }}
-var_2: &cache_key_fallback angular_devkit-0.12.0
-var_1_win: &cache_key_win angular_devkit-win-0.12.0-{{ checksum "yarn.lock" }}
-var_2_win: &cache_key_fallback_win angular_devkit-win-0.12.0
+var_1: &cache_key angular_devkit-12.9-{{ checksum "yarn.lock" }}
+var_1_win: &cache_key_win angular_devkit-win-12.9-{{ checksum "yarn.lock" }}
 var_3: &default_nodeversion "12.9"
 # Workspace initially persisted by the `setup` job, and then enhanced by `setup-and-build-win`.
 # https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs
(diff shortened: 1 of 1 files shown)


filipesilva:
Is the assumption here is that, even if we using a smaller backup cache, it still would be slower than just ignoring it? The 3.81gb is surely bloated, but we can discard it by just changing key prefixes and thus go back to a similar cache size.

If the backup cache is also ~450mbs, then it should have have a similar total time of `Cache wo/ Fallback (Hit)` plus 30s of saving cache because we used the fallback and need to save for this key. That would bring total to same as `Cache wo/ Fallback (Miss)`.

alan-agius4:
@filipesilva,

> Is the assumption here is that, even if we using a smaller backup cache, it still would be slower than just ignoring it?

Yes, because there is the overhead of restoring the fallback cache and saving the new cache. 

> The 3.81gb is surely bloated, but we can discard it by just changing key prefixes and thus go back to a similar cache size.

We'd need to change the keys quite often, based on how many packages have been updated per week.

Earlier this week:
```
Found a cache from build 154706 at angular_devkit-0.12.0-ymJ325P4dGZIgoMbjUPVe05i_3Z27cHSu721Y7xbKv4=
Size: 3.1 GiB
Cached paths:
  * /home/circleci/.cache/yarn

Downloading cache archive...
Unarchiving cache...
```

Today:
```
No cache is found for key: angular_devkit-0.12.0-nE0g1dDpAmYnOaeX2jkZ0CVaUMKR+TIXe8H_ogCwYxM=
Found a cache from build 154694 at angular_devkit-0.12.0
Size: 3.8 GiB
Cached paths:
  * /home/circleci/.cache/yarn

Downloading cache archive...
Unarchiving cache...
```

> If the backup cache is also ~450mbs, then it should have have a similar total time of Cache wo/ Fallback (Hit) plus 30s of saving cache because we used the fallback and need to save for this key. That would bring total to same as Cache wo/ Fallback (Miss).

Indeed the timings would be very similar, but what would be the benefit seeing that without cache fallback is as performant, and we'd need to keep re-invalidated the cache key every couple of days manually, since we don't have any other means to re-invalid the fallback cache.

filipesilva:
Eish, wasn't expecting our package church to be that high. I agree it's better to just skip the fallback then.

alan-agius4:
I think you meant package cache 😛

PR #17534, opened 2020-04-23
Fix example on README

shmool:
The method `workspaces.readWorkspace()` returns an object containing the member `workspace`, and not the workspace itself directly.

Added destructuring to the assignment of `workspace`.


alan-agius4:
Addressed in https://github.com/angular/angular-cli/pull/18440

Issue #12797, opened 2018-10-29
Can't use relative paths in url() in scss files


yentheo:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
```

### Command (mark with an `x`)
```
- [x] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

npm: 6.4.1
node: v10.7.0
ng cli: 6.2.6
macOS High Sierra

### Repro steps

1. Create new project
> ng new ng-sass-test--style=scss
2. Add following files:
```
  - src
    - sprite
      - _sprite.scss
      - sprite.svg
```
With the following content:
```
# _sprite.scss
.sprite-common {
    background-image: url('sprite.svg');
}
```
Add the following line to `src/styles.scss`.
`@import '~src/sprite/sprite';`
Add the same line to `src/app/app.component.scss`.
### The log given by the failure

```
ERROR in ./src/styles.scss (./node_modules/raw-loader!./node_modules/postcss-loader/lib??embedded!./node_modules/sass-loader/lib/loader.js??ref--15-3!./src/styles.scss)
Module Error (from ./node_modules/postcss-loader/lib/index.js):
(Emitted value instead of an instance of Error) CssSyntaxError: /***/ng-sass-test/src/sprite/_sprite.scss:2:22: Can't resolve 'sprite.svg' in '/***/ng-sass-test/src'

  1 | .sprite-common {
> 2 |     background-image: url('sprite.svg');
    |                      ^
  3 | }

ERROR in ./src/app/app.component.scss
Module Error (from ./node_modules/postcss-loader/lib/index.js):
(Emitted value instead of an instance of Error) CssSyntaxError: /***/ng-sass-test/src/sprite/_sprite.scss:2:22: Can't resolve 'sprite.svg' in '/***/ng-sass-test/src/app'

  1 | .sprite-common {
> 2 |     background-image: url('sprite.svg');
    |                      ^
  3 | }
```

### Desired functionality

[...]

PR #17537, opened 2020-04-24
rebase relative stylesheet assets when using preprocessors

clydin:
Previously, when using a preprocessor, resources (e.g., `url(./my-image.jpg)`) referenced in a stylesheet that was imported into another stylesheet would retain the exact URL.  This would be problematic as the resource would not be at the relative location within the new combined stylesheet.  With this change the resource URLs will now be adjusted to reference the origin location of the resource.  This allows the resources to be found without any additional changes to the application or build process.
CSS and Less already functioned in this manner. This change brings Sass and Stylus to parity.

Fixes: #12797

### DIFF
--- a/packages/angular_devkit/build_angular/test/browser/styles_spec_large.ts
+++ b/packages/angular_devkit/build_angular/test/browser/styles_spec_large.ts
@@ -259,6 +259,17 @@ describe('Browser Builder styles', () => {
   it(`supports font-awesome imports`, async () => {
     host.writeMultipleFiles({
       'src/styles.scss': `
+        @import "font-awesome/scss/font-awesome";
+      `,
+    });
+
+    const overrides = { extractCss: true, styles: [`src/styles.scss`] };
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/test/browser/styles_spec_large.ts
```diff
@@ -639,4 +683,49 @@ describe('Browser Builder styles', () => {
...
+          `,
+        [`src/more-styles/styles-two.${ext}`]: tags.stripIndents`
+            .two {
+              background-image: url(images/global-img-relative.png);
+            }
+          `,
+      });
+
+      const overrides = {
+        extractCss: true,
```
alan-agius4, line 702: For the sake of it, shouldn't we run the test without sourcemaps since this behaviour relies on them?

clydin: Good idea.  Will do.

PR #17554, opened 2020-04-27
build: fix typescript 3.9 compilation

alan-agius4:


packages/ngtools/webpack/src/compiler_host.ts
```diff
@@ -361,7 +361,7 @@ export class WebpackCompilerHost implements ts.CompilerHost {
   getCanonicalFileName(fileName: string): string {
     const path = this.resolve(fileName);
 
-    return this.useCaseSensitiveFileNames ? path : path.toLowerCase();
+    return this.useCaseSensitiveFileNames() ? path : path.toLowerCase();
```
alan-agius4, line 364: oops!

PR #17573, opened 2020-04-29
refactor: use new @angular/localize source_file_utils path

alan-agius4:
In https://github.com/angular/angular/pull/36745 the `source_file_utils.ts` was moved.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
@@ -693,10 +693,28 @@ async function inlineLocalesDirect(ast: ParseResult, options: InlineOptions) {
   }
 
   const { default: generate } = await import('@babel/generator');
-  const utils = await import(
-    // tslint:disable-next-line: trailing-comma no-implicit-dependencies
-    '@angular/localize/src/tools/src/translate/source_files/source_file_utils'
-  );
+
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -693,10 +693,28 @@ async function inlineLocalesDirect(ast: ParseResult, options: InlineOptions) {
...
 
   const { default: generate } = await import('@babel/generator');
-  const utils = await import(
-    // tslint:disable-next-line: trailing-comma no-implicit-dependencies
-    '@angular/localize/src/tools/src/translate/source_files/source_file_utils'
-  );
+
+  // Remove the below try/catch before v10 RC
+  // tslint:disable-next-line: trailing-comma no-implicit-dependencies
+  let utils: typeof import('@angular/localize/src/tools/src/translate/source_files/source_file_utils');
```
petebacondarwin, line 699: I believe that this line will fail when you update to latest master too...

alan-agius4: That will fail only if we bump the dependencies in the repo. This will need to be done after the FW cuts a release.

alan-agius4:
Closing in favor of https://github.com/angular/angular-cli/pull/17574

PR #17574, opened 2020-04-29
fix(@angular-devkit/build-angular): the path to `source_file_utils` changed in v10.0.0

petebacondarwin:


packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -511,11 +511,11 @@ function createReplacePlugin(replacements: [string, string][]): PluginObj {
 
 const USE_LOCALIZE_PLUGINS = false;
 
-async function createI18nPlugins(
+export async function createI18nPlugins(
   locale: string,
   translation: unknown | undefined,
-  missingTranslation: 'error' | 'warning' | 'ignore',
-  localeDataContent: string | undefined,
+  missingTranslation?: 'error' | 'warning' | 'ignore',
```
alan-agius4, line 517: `missingTranslation: 'error' | 'warning' | 'ignore' = 'ignore',`

And remove all the checks from above.

petebacondarwin: I made it a required param and ensured that a suitable choice is made in all callers.

alan-agius4:
Please add merge label when commit is fixed.

PR #17594, opened 2020-05-01
test: pin moment in e2e tests

alan-agius4:
Moment 2.25.0 typings are broken and results in red ci, for the time being we use a previous version of moment

See: https://github.com/moment/moment/issues/5486


alan-agius4:
Closing as it’s fixing in moment 2.25.1

PR #17597, opened 2020-05-01
refactor(@schematics/angular): remove unused buildDefaultPath

SchnWalter:
The `@schematics/angular` package contains 2 functions that do the same thing, one of them is private and unused. The functions are both called `buildDefaultPath()`.


packages/schematics/angular/utility/project.ts
```diff
@@ -1,48 +0,0 @@
...
-
-// TODO(hans): change this any to unknown when google3 supports TypeScript 3.0.
-// tslint:disable-next-line:no-any
-export function isWorkspaceSchema(workspace: any): workspace is WorkspaceSchema {
-  return !!(workspace && (workspace as WorkspaceSchema).projects);
-}
-
-// TODO(hans): change this any to unknown when google3 supports TypeScript 3.0.
-// tslint:disable-next-line:no-any
-export function isWorkspaceProject(project: any): project is WorkspaceProject {
```
SchnWalter, line 46: Looks like this file is public after all, and used by the material schematics, but it's not documented and it doesn't have a Public API guard.

I'll check if there is any chance to remove the duplicate code while still preserving this public file. 

Should this file be public?

devoto13: I would be double careful when changing these files as many projects providing schematics rely on the utilities. E.g. [ng-bootstrap](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/schematics/ng-add/index.ts#L8).

SchnWalter: Looks like the material schematics code that uses this file also requires some cleanup:

```
// TODO(jelbourn): figure out if the need for this `as any` is a bug due to two different
// incompatible `WorkspaceProject` classes in @angular-devkit
options.path = buildDefaultPath(project as any);
```

SchnWalter: I see, well, this calls for a documented Public API Guard in order to avoid the removal of various things without pushing them through a deprecation phase.

SchnWalter: Looks like we have a story for documenting the Schematics API: https://github.com/angular/angular/issues/29241

alan-agius4: These are actually private APIs which if unused we can remove.

SchnWalter: @alan-agius4, looks like the `ng add @angular/localize` schematic is also affected by this PR, but the e2e tests don't cover that package, so I've created a PR to update the schematic to the newer API, please provide feedback: https://github.com/angular/angular/pull/36897

PR #17607, opened 2020-05-02
fix(@angular-devkit/build-angular): update worker-plugin

alan-agius4:
This update contains the following fix https://github.com/GoogleChromeLabs/worker-plugin/issues/65

Closes: #17373


alan-agius4:
Caretaker note: please ignore pending status from Angular robot.

PR #17609, opened 2020-05-02
ci: disable ngcc async

alan-agius4:
Disable ngcc async mode, because the circle ci container is low on resources and causes worker failures when run in async.


alan-agius4:
Ngcc sync version 76.58s vs async version 20.86s, thus given the failure rate is more or less 50% (if not more), I think it’s worth it.

PR #17611, opened 2020-05-04
fix(@ngtools/webpack): only emit import default helper when needed

clydin:
Previously, the import default TypeScript helper was emitted for every file when in JIT mode.  This was unused code in the majority of cases.  The helper is now emitted only when needed.  For this package that would be when an Angular component decorator's resource URL properties are adjusted to support JIT execution with Webpack.


dgp1130:

PR #17624, opened 2020-05-04
build: fix //packages/angular_devkit/build_angular:build_angular_<foo>_test on OSX

gregmagolan:
These need to be run with "local" tag because puppeteer Chrome has lib files with spaces in them which Bazel does not support in runfiles. However, when tagged "local", this fixes that problem as the files with spaces are visible outside of the sandbox but it introduces the issue of the linker conflicting with the node_modules symlinks created by yarn workspaces. So in order for a test to work with "local" we need to also disable yarn workspaces.

### DIFF
--- a/packages/angular_devkit/core/BUILD.bazel
+++ b/packages/angular_devkit/core/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
+load("@build_bazel_rules_nodejs//:index.bzl", "copy_to_bin")
 load("@npm_bazel_jasmine//:index.bzl", "jasmine_node_test")
 load("//tools:defaults.bzl", "ts_library")
 
@@ -16,6 +17,17 @@ licenses(["notice"])  # MIT License
 
 # @angular-devkit/core
 
+copy_to_bin(
+    name = "core_data",
+    srcs = glob(
+        include = ["**/*.json"],
+        # NB: we need to exclude the nested node_modules that is laid out by yarn workspaces
+        exclude = ["node_modules/**"],
+    ) + [
+        "//packages/angular_devkit/core/third_party/github.com/chalk/supports-color",
+    ],
+)
+
 ts_library(
     name = "core",
     srcs = glob(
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/core/BUILD.bazel
```diff
@@ -16,6 +17,17 @@ licenses(["notice"])  # MIT License
 
 # @angular-devkit/core
 
+copy_to_bin(
```
gregmagolan, line 20: NB: this copy_to_bin is necessary because the linker is linking to bazel-out so any source files (such as `package.json`) that are required in the package need to be pushed to bazel-out with copy_to_bin. Not ideal but outside of the sandbox there is no better approach AFAIK. Side-effect is that this works on Windows which does not have a sandbox or runfiles.

gregmagolan: Without this, you'll get ```==================== Test output for //packages/angular_devkit/build_angular:build_angular_karma_test (shard 2 of 2):
internal/modules/cjs/loader.js:796
    throw err;
    ^

Error: Cannot find module '@angular-devkit/core'
Require stack:
- /private/var/tmp/_bazel_greg/dbb005a7eeb5fa8b89a9797588e36854/execroot/angular_cli/bazel-out/darwin-fastbuild/bin/packages/angular_devkit/build_angular/build_angular_karma_test.sh.runfiles/angular_cli/packages/angular_devkit/build_angular/src/karma/code-coverage_spec.js
- /private/var/tmp/_bazel_greg/dbb005a7eeb5fa8b89a9797588e36854/execroot/angular_cli/bazel-out/darwin-fastbuild/bin/packages/angular_devkit/build_angular/build_angular_karma_test.sh.runfiles/npm/node_modules/jasmine/lib/jasmine.js
```
because there is no package.json in the bazel-out folder of `@angular-devkit/core` which is linked

gregmagolan:
For reference to the spaces/runfiles issue on Bazel, here is the open issue on bazelbuild/bazel: https://github.com/bazelbuild/bazel/issues/4327 and the PR to fix it which passed OSS review but had some internal side-effects so it was dropped: https://github.com/bazelbuild/bazel/pull/9351

dgp1130:
We're going through some old PRs and came across this one. It's pretty out of date and it seems like a workaround for running certain Bazel tests on iOS. We have some inline comments about how to run these tests locally, so we do have known workarounds. Ideally they would work without requiring that knowledge, but it seems like it buts up against unsupported Bazel behavior, so it's a hard battle to win here.

Since we have these workarounds and there doesn't seem to be much need for this change right now, I think it's safe to close this PR. If we find a good strategy for handling spaces in file paths (or Bazel adds support), then we can definitely reevaluate this and see where we can improve our test infra.

PR #17625, opened 2020-05-04
fix(@angular-devkit/build-angular): downlevel and optimize locale data

clydin:
Locale data is now transformed to be compatible with the ECMAScript level of the application bundles.  The locale data is also optimized to remove comments and unnecessary whitespace.

Fixes: #17497

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
@@ -634,7 +634,7 @@ export async function inlineLocales(options: InlineOptions) {
       // If locale data is provided, load it and prepend to file
       const localeDataPath = i18n.locales[locale]?.dataPath;
       if (localeDataPath) {
-        localeDataContent = await loadLocaleData(localeDataPath, true);
+        localeDataContent = await loadLocaleData(localeDataPath, true, options.es5);
       }
     }
 
@@ -753,7 +753,7 @@ async function inlineLocalesDirect(ast: ParseResult, options: InlineOptions) {
       let localeDataSource: Source | null = null;
       const localeDataPath = i18n.locales[locale] && i18n.locales[locale].dataPath;
       if (localeDataPath) {
-        const localeDataContent = await loadLocaleData(localeDataPath, true);
+        const localeDataContent = await loadLocaleData(localeDataPath, true, options.es5);
         localeDataSource = new OriginalSource(localeDataContent, path.basename(localeDataPath));
       }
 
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Cool! So after this we could revert https://github.com/angular/angular/commit/078b0be4dc69ef4a8cdf4eb15caedde1114950c7, right?

tests/legacy-cli/e2e/tests/i18n/ivy-localize-dl-xliff2.ts
```diff
@@ -41,6 +41,10 @@ export async function executeTest() {
     await expectFileToMatch(`${outputPath}/vendor-es5.js`, '.ng.common.locales');
     await expectFileToMatch(`${outputPath}/vendor-es2015.js`, '.ng.common.locales');
 
+    // Verify the locale data is browser compatible
+    await expectToFail(() => expectFileToMatch(`${outputPath}/vendor-es5.js`, /\bconst\b/));
+    await expectFileToMatch(`${outputPath}/vendor-es2015.js`, /\bconst\b/);
+
```
petebacondarwin, line 47: Should you also have a test that proves that the comments have been removed?

clydin:
Yes.  The referenced changes can be reverted with this in place.

tests/legacy-cli/e2e/tests/i18n/ivy-localize-dl-xliff2.ts
clydin: Added.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -870,19 +870,36 @@ function findLocalizePositions(
...
   const content = fs.readFileSync(path, 'utf8');
 
-  // NOTE: This can be removed once the locale data files are preprocessed in the framework
-  if (optimize) {
-    const result = await terserMangle(content, {
-      compress: true,
-      ecma: 5,
-    });
+  // Downlevel and optimize the data
+  const transformResult = await transformAsync(content, {
```
alan-agius4, line 878: We we don't need to downlevel, why not just call `terserMangle` instead?

https://github.com/angular/angular-cli/blob/6c8b0f131595edcc278e14a4723ff01fd4419279/packages/angular_devkit/build_angular/src/utils/process-bundle.ts#L129-L164

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -870,19 +870,36 @@ function findLocalizePositions(
...
+    inputSourceMap: false as any,
+    babelrc: false,
+    configFile: false,
+    presets: [
+      [
+        require.resolve('@babel/preset-env'),
+        {
+          bugfixes: true,
+          // IE 9 is the oldest supported browser
+          targets: es5 ? { ie: '9' } : { esmodules: true },
```
alan-agius4, line 891: Shouldn't we use `targets` derived from browserslist, similar to the downlevelling process?


https://github.com/angular/angular-cli/blob/6c8b0f131595edcc278e14a4723ff01fd4419279/packages/angular_devkit/build_angular/src/utils/process-bundle.ts#L129-L164

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
clydin: Consistency mainly but this also ensures that the files are supported regardless of the locale data source ECMAScript version.  The locale data files can now target the latest version (esnext) and still work.  terser is also unnecessary for these files since they are already generated to be minimal; comments and whitespace are really the only items that remain.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
alan-agius4: thoughts @clydin ?

clydin: I actually copied the code from there and simplified it.  Went with this approach to reduce the complexity and make the output more deterministic.  When the `es5` parameter is true, the content returned is required to be es5 compatible.  Targeting IE9 guarantees that the output code will work for all supported browsers that require ES5.  And since the content is well-known and controlled and essentially an array of numbers and strings, there's really not a tangible benefit to targeting IE11 over IE9.
On the other side of the `es5` conditional, this would be possible but the targets would need to be scrubbed of any older browsers.  However, the combination of the `bugfixes` option and the `esmodules` option causes only a minimal set of changes to code (mainly to fix engine bugs) and allows most ES2016/ES2017 features to remain that would otherwise be removed.  This part can also be revisited in the future if even newer features are introduced into the locale data but right now it doesn't appear that the output would benefit.
This approach also eliminates any odd behavior due to a broken browserslist file or defects in the browserslist/caniuse packages.

PR #17630, opened 2020-05-05
feat(@schematics/angular): update compiler options target and module settings

alan-agius4:
With this change we update the target and module settings of various compilation units.

- We replace ES5 target in protractor. Protractor runs on Node.Js which support ES2018
- For applications we now use `ES2020` instead of `ESNext` as a module to avoid unexpected changes in behaviour
- For application we now use `ES2016` as a syntax target instead of `ES2015`

This changes also adds a migration to update existing projects and also removes `module` from the Universal tsconfig as per #17352 to enable lazy loading on the server.

### DIFF
--- a/integration/angular_cli/e2e/tsconfig.json
+++ b/integration/angular_cli/e2e/tsconfig.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "../out-tsc/e2e",
     "module": "commonjs",
-    "target": "es5",
+    "target": "es2018",
     "types": [
       "jasmine",
       "jasminewd2",
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
@@ -32,7 +32,8 @@ export enum ThresholdSeverity {
 }
 
 enum DifferentialBuildType {
-  ORIGINAL = 'es2015',
+  // FIXME: this should match the actual file suffix and not hardcoded.
+  ORIGINAL = 'es2016',
   DOWNLEVEL = 'es5',
 }
 
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator_spec.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator_spec.ts
@@ -126,7 +126,7 @@ describe('bundle-calculator', () => {
         {
           name: '0',
           original: {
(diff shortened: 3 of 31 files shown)


alan-agius4:
//cc @IgorMinar

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
```diff
@@ -32,7 +32,8 @@ export enum ThresholdSeverity {
 }
 
 enum DifferentialBuildType {
-  ORIGINAL = 'es2015',
+  // FIXME: this should match the actual file suffix and not hardcoded.
```
alan-agius4, line 35: 😮

integration/angular_cli/e2e/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "../out-tsc/e2e",
     "module": "commonjs",
-    "target": "es5",
+    "target": "es2018",
```
IgorMinar, line 6: Have you verified that the oldest version of node that we support is compatible with full es2018? I think it might be but I'm not sure.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator_spec.ts
```diff
@@ -126,7 +126,7 @@ describe('bundle-calculator', () => {
         {
           name: '0',
           original: {
-            filename: 'foo-es2015.js',
+            filename: 'foo-es2016.js',
```
IgorMinar, line 129: Oh uh. I didn't realize that we will start renaming the files in dist/ as well. Hmm. This file rename will break a lot of stuff in angular/angular (e.g  size tracking) and invalidate docs guides.

Let's back off on the es2016 change and keep only the modules change.

In v11 we can create a proper plan for target=es2016 or maybe even higher (zones!).

alan-agius4: I think we might want to name the files something more generic maybe instead of the syntax version.

integration/angular_cli/e2e/tsconfig.json
alan-agius4: The oldest Node version that we support is `10.13.x`, it seems that the only thing which is not supported is https://tc39.es/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-ownpropertykeys

See: https://node.green/#ES2018

alan-agius4:
New PR https://github.com/angular/angular-cli/pull/17637

PR #17632, opened 2020-05-05
[PATCH] fix(@schematics/angular): don't add `@angular/language-service` in new Angular projects

alan-agius4:
This package is no longer required since the VS code extension is already shipped with it.


alan-agius4:
@dgp1130, please not that this will never go green due to missing CI task on patch version.

PR #17637, opened 2020-05-05
feat(@schematics/angular): update compiler options target and module settings

alan-agius4:
With this change we update the target and module settings of various compilation units.

- We replace ES5 target in protractor. Protractor runs on Node.Js which support ES2018
- For applications we now use `ES2020` instead of `ESNext` as a module to avoid unexpected changes in behaviour

This changes also adds a migration to update existing projects and also removes `module` from the Universal tsconfig as per #17352 to enable lazy loading on the server.

### DIFF
--- a/integration/angular_cli/e2e/tsconfig.json
+++ b/integration/angular_cli/e2e/tsconfig.json
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "../out-tsc/e2e",
     "module": "commonjs",
-    "target": "es5",
+    "target": "es2018",
     "types": [
       "jasmine",
       "jasminewd2",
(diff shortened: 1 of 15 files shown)


integration/angular_cli/e2e/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "../out-tsc/e2e",
     "module": "commonjs",
-    "target": "es5",
+    "target": "es2018",
```
IgorMinar, line 6: I assume that this is intentionally es2018 because our oldest node version doesn't support module=es2020. Can you please leave a comment here so that we know in the future? Thanks

alan-agius4: This is target not module 😜

PR #17641, opened 2020-05-05
ci: exclude private packages from consideration for license checking

josephperrott:


dgp1130:

PR #17644, opened 2020-05-06
refactor(@angular-devkit/build-angular): support ES5 target with ES2015 APF

clydin:


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -519,10 +518,52 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
           parser: { system: true },
         },
         {
-          test: /\.js$/,
-          // Factory files are processed by BO in the rules added in typescript.ts.
-          exclude: /(ngfactory|ngstyle)\.js$/,
-          ...buildOptimizerUseRule,
+          test: /\.m?js$/,
+          exclude: /[\/\\](?:core-js|\@babel|tslib)[\/\\]/,
```
alan-agius4, line 522: Is it intended that BO now runs on `ngfactory` and `ngstyle` files?

clydin: was curious if it worked actually.  I'll add it back in though.

PR #17645, opened 2020-05-06
feat(@angular/cli): add option to enable parallel linting operations

### LINKED ISSUE #15482: ng lint param for parallel execution for fast CI builds
# 🚀 Feature request

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [x] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
On a CI environment we often have machines with lots of cores. 
The Lint without project argument runs the jobs sequentially in one thread. 
This makes the whole process unnecessarily long. 
In our application we can get the lint time down to a third if we run them in paralel with a wait command at the end. 

A parameter '--parallel' on ng lint that runs all available parts like the parameterless version but in parallel with a wait at the end. 
For each individual job the outputs should be pushed once the job is done so the different jobs don' write across one another. 
Alternatively it could always be launched silently wihout outputs except success/failure at the end. 

Handcrafting a job that explicitly calls all individual lints and joins them with a wait.

dnagl:
Resolves #15482

Add `--parallel` option to `ng lint` command to run multiple lint operations in parallel

### DIFF
--- a/docs/documentation/lint.md
+++ b/docs/documentation/lint.md
@@ -74,3 +74,12 @@ ng lint [project]
     Output format (prose, json, stylish, verbose, pmd, msbuild, checkstyle, vso, fileslist, codeFrame).
   </p>
 </details>
+<details>
+  <summary>parallel</summary>
+  <p>
+    <code>--parallel</code>
+  </p>
+  <p>
+    Run multiple lint operations in parallel.
+  </p>
+</details>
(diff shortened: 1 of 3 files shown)


alan-agius4:
Closing since the TSLint builder is deprecated as part of the TSLint deprecation.

More info https://blog.angular.io/version-11-of-angular-now-available-74721b7952f7

PR #17648, opened 2020-05-06
ci: add 0BSD as a valid license

alan-agius4:
This is required for tslib versions >= 1.11.2. Tslib changed it's license to 0BSD which should allow projects that embed or bundle tslib to omit its license header.

See: https://opensource.google/docs/thirdparty/licenses/#unencumbered

### DIFF
--- a/scripts/validate-licenses.ts
+++ b/scripts/validate-licenses.ts
@@ -42,6 +42,7 @@ const licensesWhitelist = [
   // Have a full copyright grant. Validated by opensource team.
   'Unlicense',
   'CC0-1.0',
+  '0BSD',
 
   // Combinations.
   '(AFL-2.1 OR BSD-2-Clause)',


scripts/validate-licenses.ts
```diff
@@ -42,6 +42,7 @@ const licensesWhitelist = [
   // Have a full copyright grant. Validated by opensource team.
   'Unlicense',
   'CC0-1.0',
+  '0BSD',
```
IgorMinar, line 45: we might need to consult OSPO about this. I believe it's fine to use it, but it would be great to get a confirmation from @juliaferraioli or someone else from that part of the world.

I'll shoot an email to them now.

alan-agius4:
@IgorMinar, It’s not blocking anything important. This is needed to update tslib, but for now the update is not needed.

IgorMinar:
I emailed our open source team for their confirmation that this license is ok.

PR #17652, opened 2020-05-06
fix(@angular/cli): Do not use floating promises, we can resolve these…

timfrijtersppw:
In this PR I've fixed some floating promises in the generation of component spec files and app.component.spec.ts.

Without this fix tslint errors occur when generating components when the tslint option 'no-floating-promises' is enabled. 

Also see:
https://palantir.github.io/tslint/rules/no-floating-promises/

### DIFF
--- a/packages/schematics/angular/application/other-files/app.component.spec.ts.template
+++ b/packages/schematics/angular/application/other-files/app.component.spec.ts.template
@@ -3,8 +3,8 @@ import { RouterTestingModule } from '@angular/router/testing';<% } %>
 import { AppComponent } from './app.component';
 
 describe('AppComponent', () => {
-  beforeEach(async(() => {
-    TestBed.configureTestingModule({<% if (routing) { %>
+  beforeEach(async () => {
+    return TestBed.configureTestingModule({<% if (routing) { %>
       imports: [
         RouterTestingModule
       ],<% } %>
@@ -12,7 +12,7 @@ describe('AppComponent', () => {
         AppComponent
       ],
     }).compileComponents();
-  }));
+  });
 
   it('should create the app', () => {
     const fixture = TestBed.createComponent(AppComponent);
--- a/packages/schematics/angular/e2e/files/src/app.po.ts.template
+++ b/packages/schematics/angular/e2e/files/src/app.po.ts.template
@@ -1,11 +1,11 @@
 import { browser, by, element } from 'protractor';
 
 export class AppPage {
-  navigateTo(): Promise<unknown> {
+  async navigateTo(): Promise<unknown> {
     return browser.get(browser.baseUrl) as Promise<unknown>;
   }
 
-  getTitleText(): Promise<string> {
+  async getTitleText(): Promise<string> {
     return element(by.css('<%= rootSelector %> .content span')).getText() as Promise<string>;
   }
 }
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.spec.ts.template
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.spec.ts.template
@@ -6,12 +6,12 @@ describe('<%= classify(name) %><%= classify(type) %>', () => {
   let component: <%= classify(name) %><%= classify(type) %>;
   let fixture: ComponentFixture<<%= classify(name) %><%= classify(type) %>>;
 
-  beforeEach(async(() => {
-    TestBed.configureTestingModule({
+  beforeEach(async () => {
+    return TestBed.configureTestingModule({
       declarations: [ <%= classify(name) %><%= classify(type) %> ]
     })
     .compileComponents();
-  }));
+  });
 
   beforeEach(() => {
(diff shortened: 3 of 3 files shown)


packages/schematics/angular/application/other-files/app.component.spec.ts.template
```diff
@@ -3,16 +3,16 @@ import { RouterTestingModule } from '@angular/router/testing';<% } %>
 import { AppComponent } from './app.component';
 
 describe('AppComponent', () => {
-  beforeEach(async(() => {
+  beforeEach(async(() =>
```
clydin, line 6: Can you convert these to proper async anonymous functions instead of the custom async function call?  Please keep the blocks though to simplify user additions to these functions.

packages/schematics/angular/e2e/files/src/app.e2e-spec.ts.template
```diff
@@ -8,9 +8,9 @@ describe('workspace-project App', () => {
     page = new AppPage();
   });
 
-  it('should display welcome message', () => {
-    page.navigateTo();
-    expect(page.getTitleText()).toEqual('<%= relatedAppName %> app is running!');
+  it('should display welcome message', async () => {
```
clydin, line 11: Protractor uses a custom control flow mechanism that handles the promises.  These are technically not floating.  This is something that will eventually be addressed but needs to be coordinated with protractor and migrations need to be put in place to synchronous behavior of existing projects with new projects.
There's an open PR with some more details here: https://github.com/angular/angular-cli/pull/15472

packages/schematics/angular/application/other-files/app.component.spec.ts.template
timfrijtersppw: You mean using the native async keyword instead of the custom angular async function right?

clydin: yes

packages/schematics/angular/e2e/files/src/app.e2e-spec.ts.template
timfrijtersppw: Okay, I'll leave these as is by removing them from my commits.

timfrijtersppw:
Please note that I've never really done an open source pull request before. If there is anything missing please let me know.

packages/schematics/angular/application/other-files/app.component.spec.ts.template
timfrijtersppw: Done

packages/schematics/angular/application/other-files/app.component.spec.ts.template
```diff
@@ -3,16 +3,16 @@ import { RouterTestingModule } from '@angular/router/testing';<% } %>
 import { AppComponent } from './app.component';
 
 describe('AppComponent', () => {
-  beforeEach(async(() => {
-    TestBed.configureTestingModule({<% if (routing) { %>
+  beforeEach(async () => {
+    return TestBed.configureTestingModule({<% if (routing) { %>
```
gkalpak, line 7: Is there any point in using `async` if you are not `await`ing the promise?
I would expect either `async () => { await TestBed... }` or `() => { return TestBed... }`.

packages/schematics/angular/e2e/files/src/app.po.ts.template
```diff
@@ -1,11 +1,11 @@
 import { browser, by, element } from 'protractor';
 
 export class AppPage {
-  navigateTo(): Promise<unknown> {
+  async navigateTo(): Promise<unknown> {
```
gkalpak, line 4: FWIW, you can also use `PromiseLike<...>` here and not have to cast below.
Do you know if there are any downsides to using `PromiseLike`?

packages/schematics/angular/application/other-files/app.component.spec.ts.template
clydin: i'd probably go with `await` to make it more explicit that it is an asynchronous operation and also slightly simplify future user additions (either before or after the call).

packages/schematics/angular/e2e/files/src/app.po.ts.template
clydin: `PromiseLike` is mainly useful when dealing with alternative implementations of promises (bluebird, etc.).  The only real downside is that it won't be accepted for any consumer that expects a `Promise` (e.g., `function myFunc(p: Promise<unknown>)`).

packages/schematics/angular/application/other-files/app.component.spec.ts.template
timfrijtersppw: I think async shouldn't only be used when awaiting, returning the promise also showcases asynchronous behavior.

packages/schematics/angular/e2e/files/src/app.po.ts.template
alan-agius4: Also, no reason for `async` here since you are not using `await`.

clydin:
Thank you for the contribution.  This change has, however, been superseded by https://github.com/angular/angular-cli/pull/18243 and https://github.com/angular/angular-cli/pull/18951.  Both of which will be available in the upcoming version 11.

PR #17653, opened 2020-05-06
refactor: use node v12 types, matching WORKSPACE version

jkrems:
The WORKSPACE was updated to use node 12 in https://github.com/angular/angular-cli/pull/16976 but the type checking was still using the types for node 12. The newer typings are a bit more precise but also bring in a return type overload for `fs.stat` (due to `{bigint: true}`).

This updates existing references to node 10 to node 12, moves to the latest typings packages, selects node v12 types, and fixes type errors that arise from those changes.

For the error argument type change in `node/host.ts` I went for a union of the previous (wrong) type and the newly fixed type. Technically `err` can never be undefined but this way it's still safe to pull this code into contexts that may be using outdated versions of `@types/node`. A nice future cleanup would be remove the question mark on those `err?` arguments.


jkrems:
Does the build failure here mean that we'd need to land the new types in both the main repo and this one at the same time?

alan-agius4:
Thanks for your contribution, however this is now obsolete.

PR #17670, opened 2020-05-08
 build: vendor typescript 3.6 in build optimizer

alan-agius4:
With this change we "vendor" TypeScript 3.6 in build optimizer instead of using directly the npm package.

Reasons:
- We cannot update to a more recent version of TypeScript due to microsoft/TypeScript#38412
- Yarn workspace are not supported under Bazel. This means currently we are running unit testswhich uses a different TypeScript version at runtime.
- In TypeScript 3.9, the shape of ES2015 classes changed microsoft/TypeScript#32011, this requires some changes in the UT expects, but this again will not be correct to change now.

### DIFF
--- a/packages/schematics/angular/no_typescript_runtime_dep_spec.js
+++ b/packages/schematics/angular/no_typescript_runtime_dep_spec.js
@@ -21,9 +21,9 @@ describe('@schematics/angular javascript code', () => {
           const file = path.join(subdir, f);
           if (fs.statSync(file).isDirectory()) {
             check(file);
-          } else if (file.endsWith('.ts')) {
+          } else if (file.endsWith('.js')) {
             const content = fs.readFileSync(file, { encoding: 'utf-8' });
-            if (content.indexOf(`from 'typescript'`) >= 0) {
+            if (content.includes(`require("typescript")`) || content.includes(`require('typescript')`)) {
               fail(`${file} has a typescript import`);
             }
           }
--- a/packages/angular_devkit/build_optimizer/BUILD.bazel
+++ b/packages/angular_devkit/build_optimizer/BUILD.bazel
@@ -31,12 +31,12 @@ ts_library(
     module_name = "@angular-devkit/build-optimizer",
     module_root = "src/index.d.ts",
     deps = [
+        "//packages/angular_devkit/build_optimizer/third_party/github.com/Microsoft/TypeScript",
         "@npm//@types/node",
         "@npm//@types/webpack",
         "@npm//@types/webpack-sources",
         "@npm//source-map",
         "@npm//tslib",
-        "@npm//typescript",
     ],
 )
 
@@ -47,6 +47,7 @@ ts_library(
     # @external_begin
     deps = [
         ":build_optimizer",
+        "//packages/angular_devkit/build_optimizer/third_party/github.com/Microsoft/TypeScript",
         "//packages/angular_devkit/core",
         "@npm//source-map",
     ],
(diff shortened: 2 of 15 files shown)


packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -194,6 +195,7 @@ ts_library(
     deps = [
         ":build_angular",
         "//packages/angular_devkit/architect/testing",
+        "//packages/angular_devkit/build_optimizer/third_party/github.com/Microsoft/TypeScript",
```
alan-agius4, line 198: This doesn't seem quite right, but I was getting:

```ts
ERROR: /home/circleci/ng/packages/angular_devkit/build_angular/BUILD.bazel:183:1: Compiling TypeScript (devmode) //packages/angular_devkit/build_angular:build_angular_test_lib failed (Exit 1)
packages/angular_devkit/build_angular/src/utils/build-browser-features_spec.ts:12:30 - error TS2307: [strictDeps] transitive dependency on packages/angular_devkit/build_optimizer/third_party/github.com/Microsoft/TypeScript/lib/typescript.d.ts not allowed. Please add the BUILD target to your rule's deps.

12 import { ScriptTarget } from 'typescript';
                                ~~~~~~~~~~~~
```

@gregmagolan / @filipesilva silva have you seen this in the past when vendoring packages?

filipesilva: Unsure what that means, but I agree it doesn't look right. Looks like a bad resolution to the transitive one.

packages/schematics/angular/no_typescript_runtime_dep_spec.js
```diff
@@ -21,9 +21,9 @@ describe('@schematics/angular javascript code', () => {
           const file = path.join(subdir, f);
           if (fs.statSync(file).isDirectory()) {
             check(file);
-          } else if (file.endsWith('.ts')) {
+          } else if (file.endsWith('.js')) {
             const content = fs.readFileSync(file, { encoding: 'utf-8' });
-            if (content.indexOf(`from 'typescript'`) >= 0) {
+            if (content.includes(`require("typescript")`) || content.includes(`require('typescript')`)) {
```
filipesilva, line 26: Have you checked these imports in the generated files? Might be in UMD, although I think it shouldn't

alan-agius4: Yeah it’s a UMD.

filipesilva: Does this test actually validates that there's no typescript import on UMDs?

alan-agius4: It’s validates that typescript is not imported/required.

packages/angular_devkit/build_angular/BUILD.bazel
gregmagolan: I think the dep on `"//packages/angular_devkit/build_optimizer/third_party/github.com/Microsoft/TypeScript",` is winning over the `@npm//typescript` dep here. You have both listed and they are both resolvable by `typescript`. Intentional?

As for the strict deps warning, that is an odd one. I wonder if ts_library is not doing the right thing with `typescript.d.ts` in ```ts_library(
    name = "TypeScript",
    srcs = ["lib/typescript.d.ts"],
    data = ["lib/typescript.js"],
    module_name = "typescript",
    module_root = "lib/typescript.d.ts",
    visibility = [
        "//packages/angular_devkit/build_angular:__subpackages__",
        "//packages/angular_devkit/build_optimizer:__subpackages__",
    ],
)```?

gregmagolan: Nvrmind. I thought you were getting the error with the vendored dep there. It makes sense that you'd get it without the dep there. The transiitive module_name from a dep that depends on the vendored typescript is winning over on the import and you're getting the vendored typescript instead of the one you want. 

Since you're importing the vendored typescript with relative imports, if you remove the module_name and module_root from the its ts_library target then you won't inadvertantly resolve to it in the spec.

alan-agius4: Thanks @gregmagolan, that works like a charm :)

Issue #10172, opened 2018-04-04
Unable to generate a scoped library with v6 rc0


ocombe:
### Versions
```sh
Angular CLI: 6.0.0-rc.0
Node: 8.9.3
OS: win32 x64
Angular: 6.0.0-rc.1
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

@angular/cli: 6.0.0-rc.0
@angular-devkit/architect: 0.0.10
@angular-devkit/build-angular: 0.0.10
@angular-devkit/build-ng-packagr: 0.0.4
@angular-devkit/build-optimizer: 0.4.9
@angular-devkit/core: 0.4.9
@angular-devkit/schematics: 0.5.0
@ngtools/json-schema: 1.1.0
@ngtools/webpack: 6.0.0-beta.9
@schematics/angular: 0.5.0
@schematics/update: 0.5.0
typescript: 2.7.2
webpack: 4.1.0
```

### Repro steps
* Create a new cli project with `ng new xxx`
* `cd` into the new project
* Try to generate a library with a scoped name: `ng g library @xxx/core`

### Observed behavior
It throws the following error in the shell:
```sh
Could not find an NgModule. Use the skip-import option to skip importing in NgModule.
```

### Desired behavior
It should generate a library named `@xxx/core` in the folder `projects/xxx-core` or maybe `projects/xxx/core` or even `projects/core` with the correct package name `@xxx/core` in the package.json file.

If it's not possible, then it should throw an error message that explains that it doesn't support scoped package names.

about-code:
I suggest to generate a folder structure where the package name is a correct directory path structure - similar to how scoped packages are installed into the `node_modules` folder. 

So when generating a scoped package with `ng g library @xxx/core` where the package name in `package.json` will be `@xxx/core` CLI should generate a folder structure 

```
 |-@xxx
    |-core
       |-package.json
```

Not only would it be compatible with npm layout. It is likely to work better with VSCode's auto-import suggestions. For example consider a monorepo folder structure similar to the Angular repos. You would probably generate a library in the `packages` folder then.

```
${PROJECT_ROOT}
    |-node_modules
    |-packages
    |    |-@xxx
    |    |    |-core
    |    |       |-package.json
    |    |       |-index.ts
    |    |-@foo
    |    |    |-bar
    |    |       |-src
    |    |           |-Foo.ts
    |    |       |-package.json
    |    |       |-index.ts
[...]

PR #17690, opened 2020-05-11
refactor: let library to support “@”

Eusen:
When I execute `ng g application @apps/app1`, I get the following results:
``` text
- @app/
-       app1/
(newProjectRoot = "")
```

But !!! When I execute `ng g library @libs/lib1`, I got the result:
``` text
- libs/
-      lib1/
```

This is too awkward !!!

Please! Remove the `.replace(/^@/, '')` here, or add the `.replace(/^@/, '')` to the application!

I think it's really weird now...

### DIFF
--- a/packages/schematics/angular/library/index.ts
+++ b/packages/schematics/angular/library/index.ts
@@ -181,7 +181,7 @@ export default function (options: LibraryOptions): Rule {
     let scopeName = null;
     if (/^@.*\/.*/.test(options.name)) {
       const [scope, name] = options.name.split('/');
-      scopeName = scope.replace(/^@/, '');
+      scopeName = scope;
       options.name = name;
     }
 


splincode:
@alan-agius4 is it still active?

dgp1130:
Going through some old PRs and took a quick look at this. We looked back through history and found the original motivation for stripping `@` symbols to be [terminal autocomplete issues](https://github.com/angular/angular-cli/issues/10172#issuecomment-384445140). This still seems to exist at least with Bash on Mac, where `@` directories can't be autocompleted. The CLI also generates a path mapping in `tsconfig.json`, so it isn't quite this trivial of a change.

This is probably a case where the cure is worse than the disease. Changing this behavior now would introduce a discrepancy for all new generated libraries (which include `@` in the directory), while all old library don't include `@`. We would probably need an option for existing workspaces, but we don't feel a change like this is worth supporting such an option. The autocomplete issue is also not obvious to users and it's hard to surface this as a decision with a clear cost/benefit. If we were implementing this feature now we might be more inclined to include `@` in the directory, but the nature of having existing files which don't have the `@` changes the calculus significantly.

The workaround is also fairly straightforward, since you just need to rename the directory. Arguably someone could make a custom schematic that wrapped `ng generate` and automatically renamed the output, though I'm not sure how tricky that would be.

TL;DR: The original motivation for stripping the `@` still exists, removing it would break existing user workflows, and the workaround is a simple file rename. As a result, we think it's better to keep the status quo rather than introduce additional friction with minimal value to users.

PR #17692, opened 2020-05-11
fix(@ngtools/webpack): getCanonicalFileName should return FS compatible paths

alan-agius4:
Unlike TSC which has it's own mechanism the resolve and join paths in POSIX format, NGTSC heavily relies onNode.JS `fs` and `path` modules. This prevents `Path` usage because in Windows `path.resolve` will causes an absolute path to be resolved or joined incorrectly.  Example:  `/D/MyPath/MyProject` -> `D:/d/mypath/myproject`.

With this change we change the `getCanonicalFileName` method to return FS compatible paths.


alan-agius4:
//CC @JoostK and @petebacondarwin.

Issue #17705, opened 2020-05-12
Specify page to open (mimic webpack-dev-server "openPage" option)


vokeit-gschuster:
# 🚀 Feature request

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description
 A clear and concise description of the problem or missing capability...
webpack-dev-server has an option to specify the page to open in the browser.
This is useful to have an other starting point than the baseHref, e.g. to directly start on a subpage with further options.

### Describe the solution you'd like
 If you have a solution in mind, please describe it.
"openPage" option available in angilar.json and as CMD parameter

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?
Entering in the location bar by hand each day / time when starting development - tedious & boring!

PR is here: https://github.com/angular/angular-cli/pull/17706

alan-agius4:
Hi @vokeit-gschuster, Thanks for the feature request and your contribution.

We however, don't think that this should be provided by the Angular CLI. We are very cautions in what to add since we need to supportand maintain it. We feel that this feature doesn't add a lot of value. 

Thanks,
Alan

PR #17706, opened 2020-05-12
feat(@angular/cli): Add option "openPage"

vokeit-gschuster:
Allow specification of the URL path to open in the browser in conjunction with "open" set to true to mimic webpack-dev-server's "openPage" option


alan-agius4:
Thanks again, please see: https://github.com/angular/angular-cli/issues/17705#issuecomment-627855187

PR #17711, opened 2020-05-13
Update to TypeScript 3.9

alan-agius4:


alan-agius4:
Blocking until the FW cuts a release that supports TS 3.9. Since the current pre-release version of NGCC on NPM doesn't support TS 3.9.

packages/angular_devkit/build_angular/package.json
```diff
@@ -75,7 +75,7 @@
   },
   "peerDependencies": {
     "@angular/compiler-cli": ">=10.0.0-next.0 < 11",
-    "typescript": ">=3.6 < 3.9"
+    "typescript": ">=3.9 < 4.0"
```
clydin, line 78:
```suggestion
    "typescript": ">=3.9 < 3.10"
```
Just in case.

packages/ngtools/webpack/package.json
```diff
@@ -28,13 +28,13 @@
   },
   "peerDependencies": {
     "@angular/compiler-cli": ">=10.0.0-next.0 < 11",
-    "typescript": ">=3.6 < 3.9",
+    "typescript": ">=3.9 < 4.0",
```
clydin, line 31:
```suggestion
    "typescript": ">=3.9 < 3.10",
```
Just in case as well.

packages/angular_devkit/build_angular/package.json
alan-agius4: This is actually the same as the FW.

packages/ngtools/webpack/package.json
alan-agius4: Same as above.

packages/angular_devkit/build_angular/package.json
clydin: What happens if there is though?  I think it also sends conflicting signals regarding the need to be cautious when upgrading TS minors.

alan-agius4: Fair point, though that’s very unlikely to happen.

alan-agius4: Updated

Issue #17473, opened 2020-04-14
in angular.json buildOptimizer not allowed to be set for "serve"


freed00m:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 No, I think it is simple oversight in angular.json schema 

### Description

 

Cannot disable AOT for serve if buildOptimizer set to true.
due to

buildOptimizer is illegal option in serve

## 🔬 Minimal Reproduction

I've created a custom configuration 

```
...
      "architect": {
        "build": {
          "builder": "@angular-devkit/build-angular:browser",
          "options": {
            "outputPath": "dist/frontend",
            "index": "src/index.html",
            "main": "src/main.ts",
            "polyfills": "src/polyfills.ts",
            "tsConfig": "tsconfig.app.json",
            "aot": false,
            "assets": [
              "src/assets/favicon.ico",
              "src/assets/svg",
              "src/assets/i18n"
            ],
            "styles": ["src/styles.scss"],
            "stylePreprocessorOptions": {
              "includePaths": ["src/app/shared/", "src/assets/"]
            },
            "scripts": []
          },
          "configurations": [
...
            "demo": {
              "fileReplacements": [
              ],
              "baseHref": "./",
              "optimization": true,
              "outputHashing": "none",
              "sourceMap": true,
              "aot": true,
              "extractCss": true,
              "namedChunks": false,
              "extractLicenses": true,
              "vendorChunk": false,
              "buildOptimizer": true,
              "budgets": [
                {
                  "type": "initial",
                  "maximumWarning": "5mb",
                  "maximumError": "10mb"
                },
                {
                  "type": "anyComponentStyle",
                  "maximumWarning": "6kb",
                  "maximumError": "10kb"
                }
           }

..
```

but for serving I want to turn of AOT, which is 

I could do

```
"serve": {
          "builder": "@angular-devkit/build-angular:dev-server",
          "options": {
            "browserTarget": "frontend:build",
            "baseHref": "/app/"
          },
          "configurations": {
            "production": {
              "browserTarget": "frontend:build:production"
            },
            "demo": {
              "browserTarget": "frontend:build:demo",
              "optimization": false,
              "aot": false,
[...]

PR #17713, opened 2020-05-13
Added buildOptimizer option for the ng serve command

Heerschop:
Added the `--build-optimizer` option to the `ng serve` command. 

With this change you can use `ng serve` in a production configuration without optimization enabled, which is a lot faster. This can be convenient when doing debugging, of the production configuration. Without this change this is only possible by editing the `angular.json` file.


Heerschop:
@googlebot I signed it!

alan-agius4:
Hi @Heerschop, thanks for your contribution. But, we would not like to add this atm.

In fact we want to remove more build related options for the dev server builder.

Let’s continue the discussion here: https://github.com/angular/angular-cli/issues/17473

PR #17715, opened 2020-05-13
build: use @bazel/bazelisk rather than @bazel/bazel

josephperrott:
Migrates to using @bazel/bazelisk rather than @bazel/bazel as the
latter is being deprecated.  @bazel/bazelisk is a drop in
replacement for @bazel/bazel.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -866,41 +866,15 @@
     lodash "^4.17.13"
     to-fast-properties "^2.0.0"
 
-"@bazel/bazel-darwin_x64@2.1.0":
-  version "2.1.0"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -10646,7 +10620,6 @@ sass@1.26.5, sass@^1.23.0:
 
 "sauce-connect-proxy@https://saucelabs.com/downloads/sc-4.5.4-linux.tar.gz":
   version "0.0.0"
-  uid dc5efcd2be24ddb099a85b923d6e754754651fa8
```
clydin, line 10649: any idea why this changed?

josephperrott: No idea, I just ran `yarn install` and the `yarn.lock` updated

clydin:
@josephperrott This has conflicts with the patch branch (9.1.x).  Can you create a separate PR?

Issue #15804, opened 2019-10-10
Javascript files are mapped to TypeScript compiled code instead of the original code


aminanan:
# 🐞 Bug report

### Command (mark with an `x`)

```
- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description

Javascript files are mapped to transpiled code instead of the original code.

## 🔬 Minimal Reproduction

Create new app
<pre><code>
ng new testsourcemap
cd testsourcemap
</code></pre>
Add '"allowJs": true' in compilerOptions section of tsconfig.json
<pre><code>
{
  "compileOnSave": false,
  "compilerOptions": {
    "baseUrl": "./",
    "outDir": "./dist/out-tsc",
    "sourceMap": true,
    "declaration": false,
    "downlevelIteration": true,
    "experimentalDecorators": true,
    "module": "esnext",
    "moduleResolution": "node",
    "importHelpers": true,
    "target": "es2015",
    "typeRoots": [
      "node_modules/@types"
    ],
    "lib": [
      "es2018",
      "dom"
    ],
    "allowJs": true,
  },
  "angularCompilerOptions": {
    "fullTemplateTypeCheck": true,
    "strictInjectionParameters": true
  }
}
</code></pre>
Create file testjs.js in src/app
<pre><code>
function add(a,b){
  return a+b;
}

export {
  add
};
</code></pre>

Change app.component.ts to 
<pre><code>
import { Component } from '@angular/core';
import * as test from './testjs';

@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})
export class AppComponent {
  title = 'testsourcemap';
  
  constructor() {
    let add = test.add;
  }
}
</code></pre>

## 🔥 Exception or Error
The mapped file 'webpack:///./src/app/testjs.js' under Chrome shows the following code
<pre><code>

import * as tslib_1 from "tslib";
function add(a, b) {
    return a + b;
}
export { add };
//# sourceMappingURL=testjs.js.map
</code></pre>

## 🌍 Your Environment
<pre><code>

Angular CLI: 8.3.8
Node: 10.15.3
OS: win32 x64
Angular: 8.2.10
... animations, common, compiler, compiler-cli, core, forms
[...]

PR #17717, opened 2020-05-13
Ivy-only Webpack Compiler Plugin

clydin:
This change introduces a new Ivy-only Webpack compiler plugin.  The plugin is based on and leverages the Ivy ngtsc APIs  from within the `@angular/compiler-cli` package.  The plugin also simplifies and reduces the amount of code within the plugin by leveraging newer TypeScript features and capabilities.  The need for the virtual filesystem has also been removed.  The file replacements capability was the primary driver for the previous need for the virtual filesystem.  File replacements are now implemented using a two-pronged approach.  The first, for TypeScript, is to hook TypeScript module resolution and adjust the resolved modules based on the configured file replacements.  This is similar in behavior to TypeScript path mapping.  The second, for Webpack, is the use of the `NormalModuleReplacementPlugin` to facilitate bundling of the configured file replacements.  An advantage to this approach is that the build system (both TypeScript and Webpack) are now aware of the replacements and can operate without augmenting multiple aspects of system as was needed previously.

The plugin also introduces the use of TypeScript’s builder programs.  The current primary benefit is more accurate and simplified dependency discovery.  Further, they also provide for the introduction of incremental build support and incremental type checking.

NOTE: The deprecated string format for lazy routes is not supported by this plugin.  Dynamic imports are recommended for use with Ivy and are required when using the new plugin.

Performance and memory usage have also been improved with these changes.  Initial benchmarking indicates an average ~10% reduction in production build time and a similar ~10% reduction in peak memory usage.  Watch mode (including the development server) no longer requires the forked type checker which provides a ~30% improvement in memory usage for typical development scenarios.  The plugin is also not yet fully optimized.  The next phase of development will introduce performance tracing and support further improvements.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
@@ -26,21 +26,29 @@ function canUseIvyPlugin(wco: WebpackConfigOptions): boolean {
     return false;
   }
 
-  // Allow fallback to legacy build system via environment variable ('NG_BUILD_IVY_LEGACY=1')
-  const flag = process.env['NG_BUILD_IVY_LEGACY'];
-  if (flag !== undefined && flag !== '0' && flag.toLowerCase() !== 'false') {
-    wco.logger.warn(
-      '"NG_BUILD_IVY_LEGACY" environment variable detected. Using legacy Ivy build system.',
-    );
-
+  // Allow new ivy build system via environment variable ('NG_BUILD_IVY_EXPERIMENTAL=1')
+  // TODO: Remove this section and adjust warnings below once the Ivy plugin is the default
+  const flag = process.env['NG_BUILD_IVY_EXPERIMENTAL'];
+  if (flag === undefined || flag === '0' || flag.toLowerCase() === 'false') {
     return false;
   }
+
+  // Allow fallback to legacy build system via environment variable ('NG_BUILD_IVY_LEGACY=1')
+  // TODO: Enable this section once the Ivy plugin is the default
+  // const flag = process.env['NG_BUILD_IVY_LEGACY'];
+  // if (flag !== undefined && flag !== '0' && flag.toLowerCase() !== 'false') {
+  //   wco.logger.warn(
+  //     '"NG_BUILD_IVY_LEGACY" environment variable detected. Using legacy Ivy build system.',
+  //   );
+
+  //   return false;
+  // }
 
   // Lazy modules option uses the deprecated string format for lazy routes which is not supported
   if (wco.buildOptions.lazyModules && wco.buildOptions.lazyModules.length > 0) {
     wco.logger.warn(
-      '"lazyModules" option is deprecated and not supported by the new Ivy build system. ' +
-        'Using legacy Ivy build system.'
+      '"lazyModules" option is deprecated and not supported by the experimental Ivy build system. ' +
+        'Using original Ivy build system.'
     );
 
     return false;
@@ -48,8 +56,17 @@ function canUseIvyPlugin(wco: WebpackConfigOptions): boolean {
 
   // This pass relies on internals of the original plugin
   if (wco.buildOptions.experimentalRollupPass) {
+    wco.logger.warn(
+      'The experimental rollup pass is not supported by the experimental Ivy build system. ' +
+        'Using original Ivy build system.'
+    );
+
     return false;
   }
+
+  wco.logger.warn(
+    '"NG_BUILD_IVY_EXPERIMENTAL" environment variable detected. Using experimental Ivy build system.',
+  );
 
   return true;
 }
--- a/packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
@@ -13,9 +13,78 @@ import {
   AngularCompilerPlugin,
   AngularCompilerPluginOptions,
   NgToolsLoader,
-  PLATFORM
+  PLATFORM,
+  ivy,
 } from '@ngtools/webpack';
 import { WebpackConfigOptions, BuildOptions } from '../../utils/build-options';
+import { CompilerOptions } from '@angular/compiler-cli';
+import { RuleSetLoader } from 'webpack';
+
+function canUseIvyPlugin(wco: WebpackConfigOptions): boolean {
+  // Can only be used with Ivy
+  if (!wco.tsConfig.options.enableIvy) {
+    return false;
+  }
+
+  // Allow fallback to legacy build system via environment variable ('NG_BUILD_IVY_LEGACY=1')
+  const flag = process.env['NG_BUILD_IVY_LEGACY'];
+  if (flag !== undefined && flag !== '0' && flag.toLowerCase() !== 'false') {
+    wco.logger.warn(
+      '"NG_BUILD_IVY_LEGACY" environment variable detected. Using legacy Ivy build system.',
+    );
+
+    return false;
+  }
+
+  // Lazy modules option uses the deprecated string format for lazy routes which is not supported
+  if (wco.buildOptions.lazyModules && wco.buildOptions.lazyModules.length > 0) {
+    wco.logger.warn(
+      '"lazyModules" option is deprecated and not supported by the new Ivy build system. ' +
+        'Using legacy Ivy build system.'
+    );
+
+    return false;
+  }
+
+  // This pass relies on internals of the original plugin
+  if (wco.buildOptions.experimentalRollupPass) {
+    return false;
+  }
+
+  return true;
+}
+
+function createIvyPlugin(
+  wco: WebpackConfigOptions,
+  aot: boolean,
+  tsconfig: string,
+): ivy.AngularWebpackPlugin {
+  const { buildOptions } = wco;
+  const optimize = buildOptions.optimization.scripts;
+
+  const compilerOptions: CompilerOptions = {
+    skipTemplateCodegen: !aot,
+    sourceMap: buildOptions.sourceMap.scripts,
+  };
+
+  if (buildOptions.preserveSymlinks !== undefined) {
+    compilerOptions.preserveSymlinks = buildOptions.preserveSymlinks;
+  }
+
+  const fileReplacements: Record<string, string> = {};
+  if (buildOptions.fileReplacements) {
+    for (const replacement of buildOptions.fileReplacements) {
+      fileReplacements[replacement.replace] = replacement.with;
+    }
+  }
+
+  return new ivy.AngularWebpackPlugin({
+    tsconfig,
+    compilerOptions,
+    fileReplacements,
+    emitNgModuleScope: !optimize,
+  });
+}
 
 function _pluginOptionsOverrides(
   buildOptions: BuildOptions,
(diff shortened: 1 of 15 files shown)


packages/ngtools/webpack/src/ivy/host.ts
```diff
@@ -0,0 +1,199 @@
...
+          ngcc.processModule(name, result);
+        }
+
+        return result;
+      });
+    };
+  }
+}
+
+export function augmentHostWithReplacements(
```
devoto13, line 140: It may be a bit cleaner to add these to the `paths` property of the TS config object instead of applying extra augmentation.

clydin: The two options work on different levels of the process.  `paths` operates on module names whereas `fileReplacements` operators on the resolved file path.  There's also an ordering issue as well.  `fileReplacements` must be handled after module resolution is complete including any defined `paths`.

packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
```diff
@@ -103,40 +162,78 @@ function _createAotPlugin(
...
+    module: {
+      rules: [
+        {
+          test: useIvyOnlyPlugin ? /\.tsx?$/ : /(?:\.ngfactory\.js|\.ngstyle\.js|\.tsx?)$/,
+          use: [
+            ...buildOptimizerRules,
+            useIvyOnlyPlugin ? ivy.AngularWebpackLoaderPath : NgToolsLoader,
+          ],
+        },
+        // "allowJs" support with ivy plugin - ensures build optimizer is not run twice
```
alan-agius4, line 209: This should also solve https://github.com/angular/angular-cli/issues/15804

packages/ngtools/webpack/src/index.ts
```diff
@@ -14,3 +14,8 @@ export const NgToolsLoader = __filename;
 
 // We shouldn't need to export this, but webpack-rollup-loader uses it.
 export type { VirtualFileSystemDecorator } from './virtual_file_system_decorator';
+
+import * as ivy from './ivy';
```
alan-agius4, line 18:
```suggestion
export * as ivy from './ivy';
```

packages/ngtools/webpack/src/ivy/loader.ts
```diff
@@ -0,0 +1,77 @@
...
+  const emitFile = this._compilation[AngularPluginSymbol] as FileEmitter;
+  if (typeof emitFile !== 'function') {
+    if (this.resourcePath.endsWith('.js')) {
+      // Passthrough for JS files when no plugin is used
+      this.callback(undefined, content, map);
+
+      return;
+    }
+
+    callback(new Error('The Angular Webpack loader requires the AngularWebpackPlugin'));
```
alan-agius4, line 36:
```suggestion
    callback(new Error('The Angular Webpack loader requires the AngularWebpackPlugin.'));
```

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -0,0 +1,489 @@
...
+            diagnosticsReporter,
+            resourceLoader,
+          );
+
+      const allProgramFiles = builder
+        .getSourceFiles()
+        .map((sourceFile) => sourceFile.fileName)
+        .filter(
+          (sourceFile) =>
+            !/(\.(ngfactory|ngstyle|ngsummary|ngtypecheck)\.ts|ng_typecheck__\.ts)$/.test(
```
alan-agius4, line 259:
```suggestion
            !sourceFile.endsWith('ng_typecheck__.ts')
```

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -0,0 +1,489 @@
...
+      emitNgModuleScope: false,
+      fileReplacements: {},
+      substitutions: {},
+      directTemplateLoading: true,
+      tsconfig: 'tsconfig.json',
+      ...options,
+    };
+  }
+
+  get options() {
```
alan-agius4, line 106: NIt: return type.

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -0,0 +1,489 @@
...
+        });
+    });
+
+    let ngccProcessor: NgccProcessor | undefined;
+    const resourceLoader = new WebpackResourceLoader();
+    let previousUnused: Set<string> | undefined;
+    compiler.hooks.thisCompilation.tap(PLUGIN_NAME, (thisCompilation) => {
+      const compilation = thisCompilation as WebpackCompilation;
+
+      // Store watch mode; assume true if not present (old webpack versions)
```
alan-agius4, line 159: Should we still handle this?

packages/ngtools/webpack/src/ivy/system.ts
```diff
@@ -0,0 +1,89 @@
...
+
+// Webpack's CachedInputFileSystem uses the default directory separator in the paths it uses
+// for keys to its cache. If the keys do not match then the file watcher will not purge outdated
+// files and cause stale data to be used in the next rebuild. TypeScript always uses a `/` (POSIX)
+// directory separator internally which is also supported with Windows system APIs. However,
+// if file operations are performed with the non-default directory separator, the Webpack cache
+// will contain a key that will not be purged.
+function createToSystemPath(): (path: string) => string {
+  if (process.platform === 'win32') {
+    return (path) => path.replace(/\//g, '\\');
```
alan-agius4, line 23: Should we memoize these?

packages/ngtools/webpack/src/ivy/system.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as ts from 'typescript';
+import { InputFileSystem } from 'webpack';
+
+function shouldNotWrite(): never {
+  throw new Error('Webpack TypeScript System should not write');
```
alan-agius4, line 12:
```suggestion
  throw new Error('Webpack TypeScript System should not write.');
```

packages/ngtools/webpack/src/ivy/host.ts
```diff
@@ -0,0 +1,246 @@
...
+      };
+    } else {
+      return resolvedModule;
+    }
+  };
+
+  augmentResolveModuleNames(host, tryReplace, moduleResolutionCache);
+}
+
+export function augmentHostWithSubstitutions(
```
alan-agius4, line 173: 👀 😃

packages/ngtools/webpack/src/ivy/transformation.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  if (first.afterDeclarations) {
+    result.afterDeclarations = first.afterDeclarations.slice();
+    if (second.afterDeclarations) {
+      result.afterDeclarations = result.afterDeclarations.concat(second.afterDeclarations);
+    }
+  }
+
+  return result;
+}
```
alan-agius4, line 78: Why not?
```ts
export function mergeTransformers(
  first: ts.CustomTransformers,
  second: ts.CustomTransformers,
): ts.CustomTransformers {
  const result: ts.CustomTransformers = {};

  if (first.before && second.before) {
    result.before = [...first.before, ...second.before];
  }

  if (first.after && second.after) {
    result.after = [...first.after, ...second.after];
  }

  if (first.afterDeclarations && second.afterDeclarations) {
    result.afterDeclarations = [...first.afterDeclarations, ...second.afterDeclarations];
  }

  return result;
}
```

alan-agius4 (commented):
From the initial pass, this is really good!  I particular like the `augmentHost..` pattern, removal of RXJS, Cached FS and forked type checker. In future, I can also see this working particular well, by caching the TS build info file on disk, to speed up cold builds.

Left a couple of NITs, but other than that it should be good to go! (I'll give it another pass latter tomorrow).

Good job 👍

packages/ngtools/webpack/src/ivy/plugin.ts
clydin: The peer dependency for Webpack is still `^4.0.0` and `watchMode` was added in `4.23.0`.  Once we eventually shift to Webpack 5.0 than we can probably adjust this.
I simplified the logic though so its really not much more code.

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -0,0 +1,488 @@
...
+            diagnosticsReporter,
+            resourceLoader,
+          );
+
+      const allProgramFiles = builder
+        .getSourceFiles()
+        .map((sourceFile) => sourceFile.fileName)
+        .filter(
+          (sourceFile) =>
+            !/(\.(ngfactory|ngstyle|ngsummary|ngtypecheck)\.ts)$/.test(sourceFile) &&
```
alan-agius4, line 259: You seem to have forgotten to remove this line
```suggestion
            !/(\.(ngfactory|ngstyle|ngsummary|ngtypecheck)\.ts)$/.test(sourceFile) &&
```


ngtypecheck -> was used before
ngfactory|ngstyle|ngsummary -> are for VE

alan-agius4 (approved):
One last comment above, otherwise LGTM.

Feel free to mark it for merge when the above has been addressed.

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -0,0 +1,486 @@
...
+  }
+
+  private updateAotProgram(
+    compilerOptions: CompilerOptions,
+    rootNames: string[],
+    host: CompilerHost,
+    diagnosticsReporter: DiagnosticsReporter,
+    resourceLoader: WebpackResourceLoader,
+  ) {
+    // Create an ngtsc plugin instance used to initialize a TypeScipt host and program
```
JoostK, line 306:
```suggestion
    // Create an ngtsc plugin instance used to initialize a TypeScript host and program
```

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -0,0 +1,486 @@
...
+        break;
+      }
+    }
+
+    // Collect semantic diagnostics
+    for (const sourceFile of builder.getSourceFiles()) {
+      if (!ignoreForDiagnostics.has(sourceFile)) {
+        diagnosticsReporter(builder.getSemanticDiagnostics(sourceFile));
+      }
+    }
```
JoostK, line 358: I believe that the non-semantic diagnostics are typically gathered _before_ semantic diagnostics.

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -0,0 +1,486 @@
...
+
+    const builder = ts.createEmitAndSemanticDiagnosticsBuilderProgram(
+      program,
+      wrappedHost,
+      this.builder,
+    );
+
+    // Save for next rebuild
+    if (this.watchMode) {
+      this.ngtscNextProgram = ngtsc.getNextProgram();
```
JoostK, line 339: Requesting the next program here might be premature, as requesting type check diagnostics (`NgtscPlugin.getDiagnostics`) will create a new `ts.Program` which provides a new next program to use for the next incremental build. I verified that calling `NgtscPlugin.getNextProgram` just before calling `this.createFileEmitter` in the promise handler does indeed return a different `ts.Program`.

packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
```diff
@@ -13,9 +13,95 @@ import {
...
+  //   );
+
+  //   return false;
+  // }
+
+  // Lazy modules option uses the deprecated string format for lazy routes which is not supported
+  if (wco.buildOptions.lazyModules && wco.buildOptions.lazyModules.length > 0) {
+    wco.logger.warn(
+      '"lazyModules" option is deprecated and not supported by the experimental Ivy build system. ' +
+        'Using existing Ivy build system.'
```
JoostK, line 51: I think "existing" is confusing here, how about "original" (to avoid calling it "legacy" while it's not enabled by default). Same for the message below.

packages/ngtools/webpack/src/ivy/plugin.ts
clydin: This was originally later in the process.  As an experimental with the release of TS 4.0, I moved it to its current location and it didn't appear to cause a change in behavior (no crashes, diagnostics appear to be the same, etc.).  TS 4.0 removed an unnecessary assert statement that would previously cause an exception in this case (https://github.com/microsoft/TypeScript/blob/v3.9.2/src/compiler/program.ts#L1239).

JoostK: Yeah I also didn't observe any difference in behavior. I was expecting it to crash due to that assertion, but if it's been removed then surely it does not. I'm not sure if this causes extra memory to be retained, and what the size of a reused `ts.Program` would be. Maybe @alxhub has some more insights here.

alxhub (changes requested):
`NgTscPlugin` is not yet a finished and stable API, and we're not ready to have Tooling -> FW dependencies on it.

Rush:
Is it available as a beta API though? Can I just replace `AngularCompilerPlugin` with `NgTscPlugin`?

alxhub:
> Is it available as a beta API though? Can I just replace `AngularCompilerPlugin` with `NgTscPlugin`?

`Plugin` means two different things between these two classes.

`AngularCompilerPlugin` is a webpack plugin. `NgTscPlugin` is a plugin to `tsc_wrapped` from the Bazel build infrastructure. It _might_ be wrappable in a webpack context (as `ivy.AngularCompilerPlugin` in this PR does), but that is left as an exercise for the reader currently.

PR #17718, opened 2020-05-13
build: use @bazel/bazelisk rather than @bazel/bazel

josephperrott:
Migrates to using @bazel/bazelisk rather than @bazel/bazel as the
latter is being deprecated.  @bazel/bazelisk is a drop in
replacement for @bazel/bazel.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -33,7 +33,7 @@ load("@build_bazel_rules_nodejs//:index.bzl", "check_bazel_version", "check_rule
 check_bazel_version(
     message = """
 You no longer need to install Bazel on your machine.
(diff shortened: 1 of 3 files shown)


filipesilva:
@josephperrott this is marked for merge but it's failing on the bazel test. Are you sure you want to make this change though? Master is already on bazelisk.

josephperrott:
I can look at the failure, I originally made the `@bazel/bazelisk` PR to target both master and patch, but was told that it couldn't cherry-pick cleanly into patch.  I will have to investigate what is happening on this PR.

alan-agius4:
Closing as we are already using bazelisk

PR #17739, opened 2020-05-15
Test performance and local execution usability improvements

clydin:


packages/angular_devkit/build_angular/src/browser/specs/styles_spec.ts
```diff
@@ -305,6 +305,9 @@ describe('Browser Builder styles', () => {
       '.browserslistrc': 'IE 10',
     });
 
+    // Set to target to ES5 to avoid differential loading and unnecessary testing time
+    host.replaceInFile('tsconfig.json', '"target": "es2015"', '"target": "es5"');
```
alan-agius4, line 309: Nice!

Issue #17735, opened 2020-05-15
'let' instead of 'var' in ES5 bundles


javiermarinros:
The Angular CLI generated ES5 bundles fails in old ES5 browsers because it's using `let` instead of `var` in source code areas related to localization and i18n using Angular + Angular CLI 9.1.7

Steps to reproduce:

> ng new test
> ng add @angular/localize
> set `"localize": true` in `angular.json`
> ng build
> check generated `vendor-es5.js ` file and look for the string `let`

petebacondarwin:
This should have been fixed by https://github.com/angular/angular-cli/pull/17625

petebacondarwin:
But it only landed on `master` so will only take effect in v10.0.0...

petebacondarwin:
Perhaps you could try updating CLI to v10.0.0-next.5?

javiermarinros:
For now I'm manually replacing 'let' with 'var' before deploying to production and it works, I'll try CLI v10

alan-agius4:
Closed via https://github.com/angular/angular-cli/pull/17749

PR #17749, opened 2020-05-18
fix(@angular-devkit/build-angular): downlevel and optimize locale data

alan-agius4:
Locale data is now transformed to be compatible with the ECMAScript level of the application bundles.  The locale data is also optimized to remove comments and unnecessary whitespace.

Fixes: #17497 and Fixes: #17735

Patch version of https://github.com/angular/angular-cli/pull/17625


clydin:
@googlebot I consent.

PR #17759, opened 2020-05-19
Tslib 2.0 changes

alan-agius4:
See individual commits.

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/dedupe-module-resolve-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/dedupe-module-resolve-plugin.ts
@@ -1,0 +1,81 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+interface NormalModuleFactoryRequest {
+  request: string;
+  context: {
+    issuer: string;
+  };
+  relativePath: string;
+  path: string;
+  descriptionFileData: {
+    name: string;
+    version: string;
+  };
+  descriptionFileRoot: string;
+  descriptionFilePath: string;
+}
+
+export interface DedupeModuleResolvePluginOptions {
+  verbose?: boolean;
+}
+
+/**
+ * DedupeModuleResolvePlugin is a webpack resolver plugin which dedupes modules with the same name and versions
+ * that are laid out in different parts of the node_modules tree.
+ *
+ * This is needed because Webpack relies on package managers to hoist modules and doesn't have any deduping logic.
+ */
+export class DedupeModuleResolvePlugin {
+  modules = new Map<string, NormalModuleFactoryRequest>();
+
+  constructor(private options?: DedupeModuleResolvePluginOptions) { }
+
+  // tslint:disable-next-line: no-any
+  apply(resolver: any) {
+    resolver
+      .getHook('before-described-relative')
+      .tapPromise('DedupeModuleResolvePlugin', async (request: NormalModuleFactoryRequest) => {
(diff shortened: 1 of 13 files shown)


packages/schematics/angular/migrations/migration-collection.json
```diff
@@ -85,15 +85,20 @@
...
+    },
+    "update-workspace-dependencies": {
+      "version": "10.0.0-beta.7",
+      "factory": "./update-10/update-dependencies",
+      "description": "Workspace dependencies updates."
+    },
+    "update-libraries-tslib": {
+      "version": "10.0.0-beta.7",
+      "factory": "./update-10/update-libraries-tslib",
+      "description": "Update library projects to use tslib version 2 as a direct dependency."
```
SchnWalter, line 101: If you are a author of a private library and you control all the dependencies, it should be mentioned that you can still use `tslib` as a peer dependency if you use version 2 across the board.

alan-agius4: I personally don’t think that we should make any distinction between private and public libraries. Since private libraries can also have different private consumers when published.

In case you are using your private library in the same mono repo peerDependencies and dependencies have no effect, because the library is never installed.

Also, since tslib is coupled with the TypeScript version that the library author compiled the library with, it should not be up to the consumer to decide which version of tslib to install, but rather it is the responsible of the library author to choose the correct tslib version. Example if a library was compiled using TS 3.9 it should depend on tslib 2, while if it was compiled using TS 3.8, it should depend upon tslib 1.13.

SchnWalter: I see. Thanks for the feedback.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/dedupe-module-resolve-plugin.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export interface DedupeModuleResolvePluginOptions {
+  verbose?: boolean;
+}
+
+/**
+ * DedupeModuleResolvePlugin is a webpack resolver plugin which dedupes modules with the same name and versions
+ * that are laid out in different parts of the node_modules tree.
+ *
+ * This is needed because Webpack relias on package managers to hoist modules and doesn't have any deduping logic.
```
clydin, line 32:
```suggestion
 * This is needed because Webpack relies on package managers to hoist modules and doesn't have any deduping logic.
```

tests/legacy-cli/e2e/tests/build/material.ts
```diff
@@ -6,7 +6,7 @@ import { updateJsonFile } from '../../utils/project';
 const snapshots = require('../../ng-snapshot/package.json');
 
 export default async function () {
-  await ng('add', '@angular/material');
+  await ng('add', '@angular/material@next');
```
clydin, line 9: We probably need a conditional here that checks the CLI version being tested for `next` so that when material's `@next` is at 11 these tests don't fail for the 10.x branch

tests/legacy-cli/e2e/tests/misc/debupe-duplicate-modules.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        console.log({
+            __assign_0,
+            __assign_1,
+            __assign_2,
+        })
+    `);
+
+    const { stderr } = await ng('build', '--verbose', '--no-vendor-chunk');
+    if (!/\[DedupeModuleResolvePlugin\]:.+\/node_modules\/tslib-1-copy -> .+\/node_modules\/tslib-1/.test(stderr)) {
```
clydin, line 33: Nice test method.

alan-agius4: thanks ;)

PR #17777, opened 2020-05-21
feat(@schematics/angular): Add `index.ts`, so that IDEs can "shorten import" normally

asnowwolf:
If the source code of the library is added to the path map in `tsconfig.json`. Then IDE (such as WebStorm) will not be able to "shorten import" normally, because it only recognizes `/index.ts`, but not `/src/public-api.ts`. Adding an `/index.ts` by default can solve this issue.

Considering that many internal libraries may prefer this way, I think it might be worth being shipped by default.

Reprod: https://github.com/asnowwolf/reprod-path-mapping

### DIFF
--- a/packages/schematics/angular/library/files/index.ts
+++ b/packages/schematics/angular/library/files/index.ts
@@ -1,0 +1,1 @@
+export * from './src/public-api';


SchnWalter:
@asnowwolf, I've been using `public-api.ts` in libraries for a long time now and didn't encounter any issues. And I've been using WebStorm, PhpStorm and IntelliJ IDEA Ultimate Edition. What kind of issues do you have?

asnowwolf:
@SchnWalter 

1. Open https://github.com/asnowwolf/reprod-path-mapping with WebStorm
2. Remove `import { SharedService } from 'shared';` from `app.component.ts` and `projects/shared/index.ts`
3. Try to auto import `SharedService`. It will import `import { SharedService } from '../../projects/shared/src/lib/shared.service';` instead of `import { SharedService } from 'shared';`

What I mean is not to rename `public-api.ts` to `index.ts`, but to provide an additional `index.ts` to meet IDE conventions.

SchnWalter:
(This is strange, I can't create a thread).


You've changed you tsconfig to something that is invalid. By default the paths entry for the library points to the built package.
```
    "paths": {
      "shared": [
        "dist/shared/shared",
        "dist/shared"
      ]
    }
```

If you want to use the code directly, which you should avoid, then you need to change the paths entry to this:
```
    "paths": {
      "shared": [
        "projects/shared/src/public-api.ts",
      ]
    }
```
There is no `"projects/shared"` TS module, the alias that you have is invalid.

asnowwolf:
Yes, It works! My mistakes. Thank you!

PR #17787, opened 2020-05-25
docs: README info about running tests

juristr:
The current README is out of date as it still refers to running tests with node. Instead the `package.json` scripts have been modified to use Bazel instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,21 +116,14 @@ as expected.
 There are three different test suites which can be run locally:
 
 * Unit tests
-  * Run: `yarn test --full`
-  * Debug: `yarn debug:test --full`
+  * Run: `yarn bazel:test`
 * Large tests
-  * Run: `yarn test-large --full`
-  * Debug: `yarn debug:test-large --full`
+  * Run: `yarn bazel:integration`
 * End to end tests
   * Run: `node tests/legacy-cli/run_e2e.js`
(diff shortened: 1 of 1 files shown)


juristr:
@alan-agius4 oh I see. Will do later 👍

cexbrayat:
@juristr 👍 Awesome!

I think it would be great to also showcase how you can run only _some_ of the tests, as the full build takes forever. 
I'm using `yarn bazel test //packages/angular_devkit/build_angular:build_angular_browser_test` for example but maybe someone has an even better command than this? Is there a way to have even more granularity? When you want to make a small contribution, and run just one spec, it would be great to find how to do it in the README.

juristr:
@cexbrayat agree, unfortunately I don't either. Didn't look into it. I just had to run the tests myself and found the outdated readme, which is why I went and submitted this PR :)

Maybe @alan-agius4 knows 🙂

@alan-agius4 I'm gonna fix this PR in the coming days

alan-agius4:
You can use the standard jasmine methods `fit` and `fdescribe`.

cexbrayat:
@alan-agius4 Might be worth adding to the README, but Bazel will still download and compile everything no? `yarn bazel test //packages/angular_devkit/build_angular:build_angular_browser_test` might be worth mentionning as it "just" compile the package you try to test?

alan-agius4:
Closing as this should be already fixed on master.

PR #17793, opened 2020-05-26
fix(@schematics/angular): remove node typings from tsconfig.spec.json

alan-agius4:


packages/schematics/angular/workspace/files/package.json.template
```diff
@@ -25,8 +25,7 @@
   },
   "devDependencies": {
     "@angular/cli": "<%= '~' + version %>",
-    "@angular/compiler-cli": "<%= latestVersions.Angular %>",
-    "@types/node": "^12.11.1",<% if (!minimal) { %>
```
devoto13, line 29: It is still used by [`e2e/tsconfig.json`](https://github.com/angular/angular-cli/blob/4fedf956e63eaea033d69227a142881a519577e7/packages/schematics/angular/e2e/files/tsconfig.json.template), isn't it?

alan-agius4: You are right.

Issue #17598, opened 2020-05-01
ng generate should generate code directly in the sourceRoot


SchnWalter:
# 🚀 Feature request

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

Currently the `ng generate` command doesn't generate the code directly in the `sourceRoot` of the project. Depending on the project type, the code is generated under an `app` or `lib` sub-directory. Because of this, when you want to generate code in the sourceRoot, you need to prefix the name for the schematic with `../`.

For example, if you want to add a `core` sub-package to a library project and expose it under `@example/library/core`, you need to use this command:
```
ng generate module ../core/example-library-core --flat
```
Instead of:
```
ng generate module core/example-library-core --flat
```
which will generate the new `ExampleLibraryCoreModule` under `@example/library/lib/core`, instead of just `@example/library/core`.

The same applies to application projects, where all code is generated under the `app` sub-directory of the sourceRoot and this happens because the generate command uses [`buildDefaultPath()`](https://github.com/angular/angular-cli/blob/v9.1.4/packages/schematics/angular/utility/workspace.ts#L71-L80) to and always adds this extra `projectDirName` directory .

I know that removing the extra directory will break existing example commands, but generating code directly in the `sourceRoot` directory feels more intuitive then generating code in these "arbitrary" directories.

**NOTE:** If your library doesn't use the `lib` directory at all in order to have nice sub-package imports, you still have to provide the `../` prefix for all generate commands.

### Describe the solution you'd like

Remove the extra sub-directories added by `buildDefaultPath()`, but find a way to keep old example commands working, so that in old projects, a command like `ng g m extras` would generate the module in the same location. Maybe a new "defaultDirectory" schematics configuration could be introduced in `angular.json` in order to avoid introducing regressions in existing codebases.

### Describe alternatives you've considered

You can always use `../` to prefix the names of generated code (modules/components/etc...).

---

P.S. Currently there are 2 `buildDefaultPath()` functions that have the same role, one of them is unused, see PR https://github.com/angular/angular-cli/pull/17597 and commit https://github.com/angular/components/commit/d509ff7f26875e0208d4dcbe284846acd713da6e

kyliau:
[...]

PR #17816, opened 2020-05-28
fix(@schematics/angular): add appDir option

### LINKED ISSUE #17813: while generating appShell option appDir is ignored
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

We have an application with universal, where the app.server.module.ts file is not in app folder, when I tried to use ng generate appShell --appDir='path' it ignores it throws exception 

Specified module 'app.server.module.ts' does not exist.
Looked in the following directories:
    /src/app/app.server.module.ts
    /src/app/app-shell
    /src/app
    /src

### Description

## 🔥 Exception or Error
<pre><code>

Specified module 'app.server.module.ts' does not exist.
Looked in the following directories:
/src/app/app.server.module.ts
/src/app/app-shell
/src/app
/src
</code></pre>

## 🌍 Your Environment
<pre><code>

Angular CLI: 9.0.5
Node: 12.14.1
OS: win32 x64

Angular: 9.0.5
... platform-server, router
Ivy Workspace: Yes

@angular-devkit/architect         0.900.5
@angular-devkit/build-angular     0.900.5
@angular-devkit/build-optimizer   0.900.5
@angular-devkit/build-webpack     0.900.5
@angular-devkit/core              9.0.5
@angular-devkit/schematics        9.0.5
@ngtools/webpack                  9.0.5
@nguniversal/builders             9.0.1
@nguniversal/common               9.0.1
@nguniversal/express-engine       9.0.1
@schematics/angular               9.0.5
@schematics/update                0.900.5
rxjs                              6.5.4
typescript                        3.7.5
</code></pre>

**Anything else relevant?**

santoshyadavdev:
fixes #17813


packages/schematics/angular/app-shell/index.ts
```diff
@@ -237,7 +237,11 @@ function addServerRoutes(options: AppShellOptions): Rule {
     if (!clientServerOptions) {
       throw new SchematicsException('Server target does not contain options.');
     }
-    const modulePath = getServerModulePath(host, clientProject.sourceRoot || 'src', options.main as string);
+    const sourceRoot = clientProject.sourceRoot || 'src';
+    const serverPath = options.appDir ?
+     `${sourceRoot}/${options.appDir}` :
+    sourceRoot;
```
SchnWalter, line 243: This change needs a more thorough analysis. I think that `appDir` option should be removed - it didn't work in the first place, so just deprecate it now and remove it later, and there is no harm done.

In other schematics, [buildDefaultPath()](https://github.com/angular/angular-cli/blob/v10.0.0-rc.0/packages/schematics/angular/utility/workspace.ts#L72-L81) is used to hardcode an `app` or `lib` directory depending on the project type and it will come into conflict with this; you want other schematics to work even for projects that have used the `app-shell` schematics.

Alternatively, this could be implemented for all schematics, and have it fallback to the current behavior of `buildDefaultPath()`. I've actually submitted a similar feature suggestion, but for removing the sub-directory for all schematics, see #17598.

But I'm not sure if either of these should be implemented, we need to make sure that existing tutorials on how to generate components will continue to work in the future (including for projects that have used `app-shell`).

santoshyadav198613:
@alan-agius4 it's open for review.

alan-agius4:
@santoshyadav198613, I’ll review it tomorrow. Thanks.

packages/schematics/angular/app-shell/index_spec.ts
```diff
@@ -10,6 +10,21 @@ import { Schema as ApplicationOptions } from '../application/schema';
 import { Schema as WorkspaceOptions } from '../workspace/schema';
 import { Schema as AppShellOptions } from './schema';
 
+function createServerFile(appTree: UnitTestTree): void {
```
alan-agius4, line 13: This is not needed.

packages/schematics/angular/app-shell/schema.json
```diff
@@ -33,8 +33,7 @@
     "appDir": {
       "type": "string",
       "format": "path",
-      "description": "The name of the application directory.",
-      "default": "app"
```
alan-agius4, line 37: Please revert this.

packages/schematics/angular/app-shell/index_spec.ts
santoshyadav198613: Ohh got it, So my impression was incorrect, I will make the changes.

santoshyadav198613:
Hi @alan-agius4 ,
So let me understand this correctly, we can use appDir option in appShell only if while running universal schematics we used appDir option, is my understanding correct?

alan-agius4:
@santoshyadav198613, yes I believe so. The `appDir` is primarily used by the universal schematic. 

We might need to use the `path` option in the component schematic, but I guess this will be uncovered when creating the UT.

santoshyadav198613:
Hi @alan-agius4 , Need one help the app-shell test is failing should I simulate ngAdd for universal as done here https://github.com/angular/angular-cli/blob/4d741de8a9969b5d828cc6f7f71e8452a0181a35/packages/schematics/angular/app-shell/index_spec.ts#L160

santoshyadav198613:
Hi @alan-agius4 ,
Need one help app-shell test case fails it, I thought app-shell runs universal schematics.

alan-agius4:
@santoshyadav198613, just to let you know that I didn’t forget about this! Apologies for taking so long to look at this.

alan-agius4:
@santoshyadavdev thanks for your PR. This is now obsolete due to changes in the recent releases.

PR #17834, opened 2020-06-01
feat(@schematics/angular): add tslint no-any and typedef rules when in strict mode

### LINKED ISSUE #17821: Enforce stricter types via lint when in strict mode
# 🚀 Feature request

### Command
- [x] generate library

### Description
By default, it is not required to explicitly type functions' return value, as TypeScript can always infer the type based on the function code.

But the inference assumes that the function's code is correct. If the code change or is incorrect, the return type inference will reflect that.

While it is acceptable in applications, as it will raise an error where the function is used if the types do not match anymore, it is more dangerous in a library, as the functions are not used directly. It can lead to unwanted breaking changes in the public API.

TSLint has a rule to enforce functions' return type (and ESLint too, for the future migration):
```json
{
  "rules": {
    "typedef": [true, "call-signature"]
  }
}
```

It could be added when doing `ng g library`.

It is an easy change as:
- it won't impact applications: libraries already have a specific `tslint.json` due to `--prefix` option
- it won't impact already created libraries as it only happens when generating a new one

This is a really easy thing to add, I can take care of the PR if Angular team signal is green.

If it is considered a too big step, it could be added only if the workspace is already using TypeScript `strict` mode (following the new `--strict` option in Angular >= 9). But I think as we are talking about libraries, it should really be the default: libraries authors must be stricter to avoid compatibility issues (`strict` mode should even be the default for libraries, but it would be another discussion).

cyrilletuzi:
Fixes #17821

### DIFF
--- a/packages/schematics/angular/workspace/files/tslint.json.template
+++ b/packages/schematics/angular/workspace/files/tslint.json.template
@@ -83,7 +83,10 @@
         "named": "never"
       }
     },
-    "typedef": [true, "call-signature"],
+    "typedef": [
+      true,
+      "call-signature"
+    ],
     "typedef-whitespace": {
       "options": [
         {
--- a/packages/schematics/angular/workspace/files/tslint.json.template
+++ b/packages/schematics/angular/workspace/files/tslint.json.template
@@ -42,7 +42,8 @@
           "instance-method"
         ]
       }
-    ],
+    ],<% if (strict) { %>
+    "no-any": true,<% } %>
     "no-console": [
       true,
       "debug",
@@ -82,6 +83,7 @@
         "named": "never"
       }
     },
+    "typedef": [true, "call-signature"],
     "typedef-whitespace": {
       "options": [
         {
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/workspace/files/tslint.json.template
```diff
@@ -131,6 +131,8 @@
     "template-banana-in-box": true,
     "template-no-negated-async": true,
     "use-lifecycle-interface": true,
-    "use-pipe-transform-interface": true
+    "use-pipe-transform-interface": true<% if (strict) { %>,
+    "typedef": [true, "call-signature"],
```
alan-agius4, line 135: We should add `typedef` always.

cyrilletuzi: @alan-agius4 Done.

cyrilletuzi:
@alan-agius4 Just one question: will it be included in v10.0?

alan-agius4:
@cyrilletuzi, since it's a feature it will be available in 10.1. Since the v10 scope is closed.

cyrilletuzi:
@alan-agius4 Right, but as it do not affect current projects, and as it goes along with the other changes in v10 enhancing `--strict` option (`new new` prompt, full TS `strict`, stricter budgets, etc.), wouldn't it be nice to include it in v10.0 to promote this mode when Angular 10 stable is released (for example in the release blog post)?

alan-agius4:
So we had a chat around this, and we decided to have it part of 10.0

PR #17835, opened 2020-06-01
refactor(@schematics/angular): add `jsonc-parser` to parse and modify JSON

alan-agius4:
Currently, interaction with JSON AST is complex and requires a lot of boiler-platting. With this change we add `jsonc-parser` and add a basic helper class to greatly simplify such interaction.

This would come in handle when writing new migrations in 10.1+

**Note**: this is a private implementation and shouldn't be used by 3rd parties.


packages/schematics/angular/utility/json-file.ts
```diff
@@ -0,0 +1,82 @@
...
+  modify(jsonPath: JSONPath, value: JsonValue, getInsertionIndex?: (properties: string[]) => number): void {
+    if (!getInsertionIndex) {
+      const property = jsonPath.slice(-1)[0];
+      getInsertionIndex = properties => [...properties, property].sort().findIndex(p => p === property);
+    }
+
+    const edits = modify(
+      this.content,
+      jsonPath,
+      value === null ? undefined : value,
```
clydin, line 62: `null` is a valid JSON value.  This should probably support `undefined` as a valid type for the `value` parameter.

alan-agius4:
Ps: there are still some schematics that use the old methods, but I’ll continue the cleanup later on.

I want to get this in asap, so that I can write the 10.1 migration using this new method.

packages/schematics/angular/utility/json-file.ts
alan-agius4: Sounds like a good idea!

PR #17884, opened 2020-06-09
perf(@ngtools/webpack): only process dependent packages with NGCC

alan-agius4:
With this change we consume the change in https://github.com/angular/angular/pull/37075 where the NGCC now exposes a new option `--use-program-dependencies`, to only process packages which are part of the TypeScript program.

From the initial benchmarking the time taken for NGCC to process that dependencies needed for an `ng new` application went down by 36% from 10526ms to 6708ms.

### DIFF
--- a/packages/ngtools/webpack/src/ngcc_processor.ts
+++ b/packages/ngtools/webpack/src/ngcc_processor.ts
@@ -85,6 +85,7 @@ export class NgccProcessor {
         '--async',
         '--tsconfig', /** tsConfigPath */
         this.tsConfigPath,
+        '--use-program-dependencies',
       ],
       {
         stdio: ['inherit', process.stderr, process.stderr],


petebacondarwin:
The real concern is actually if it slows down the processing of large application because we now scan all their source files as part of the ngcc processing (even if the node_modules have all been processed already).
Does anyone have any large apps that they could test this on?

alan-agius4:
@petebacondarwin let's take this offline.

alan-agius4:
Testing it locally with a dummy 800 component files the NOOP regressed by 200ms.

alan-agius4:
In a large real world application (super-productivity) a NOOP regressed from 1129.807ms to 4685.788ms

alan-agius4:
On another application which was shared privately with the team. The timings of a noop increased from 687.110ms to 1342.186ms.

petebacondarwin:
I ran some tests on super-productivity using the directory-walker and program-based entry-point finders. See below. I would say that in large projects like this the cost of the program-based finder, is not significant in the bigger picture of the overall build. In fact the variation in the overall build on my machine was so wide that the total build time actually decreased in the "noop" case between the directory-walker and the program-based entry-point finder runs.


directory walker entry-point finder |   |   |  
-- | -- | -- | --
  | ngcc async | create program | total
cold | 79083 | 5926 | 107101
hot | 720 | 5220 | 64986
hot | 959 | 5449 | 66637
hot | 830 | 5944 | 72631
hot (average) | 836.3 | 5537.7 | 68084.7

program-based entry-point finder |   |   |  
-- | -- | -- | --
  | ngcc async | create program | total
cold | 86989 | 8079 | 113627
hot | 3443 | 5547 | 68565
hot | 3547 | 5039 | 66117
hot | 3590 | 4980 | 64235
hot (average) | 3526.7 | 5188.7 | 66305.7

  | ngcc async | create program | total
-- | -- | -- | --
cold diff | 7906 | 2153 | 6526
  | 10.00% | 36.33% | 6.09%
hot diff | 2690.3 | -349.0 | -1779.0
  | 321.68% | -6.30% | -2.61%

PR #17893, opened 2020-06-10
feat(@angular/cli): add default value for options in help

### LINKED ISSUE #17191: Show default values for each command flag
# 🚀 Feature request

Today I started a new NG9 project and first things first I have set up CLI commands, packages and so on. While doing that, I was wondering what defaults are being set by the CLI and couldn't find any reference in the docs.

Let's say `ng serve --hmr`, is HMR not enabled by default? Those are questions I ask myself and I think others that are starting with NG as well. To make that easier. We should add the default value behind every flag:

```
ng server --help
...
  --hmr
    Enable hot module replacement. (default: true)
```

At the moment my command looks like that:

```
ng serve --watch --live-reload --aot --hmr --open --progress --source-map
```

Because I'm not sure if any of them are true by default. Sure, I expect them to be true but I'm not sure. My previous projects were developed through Ionic CLI which had handled that `ng serve` under the hood.

- [x] new
- [x] build
- [x] serve
- [x] test
- [x] e2e
- [x] generate
- [x] add
- [x] update
- [x] lint
- [x] xi18n
- [x] run
- [x] config
- [ ] help
- [ ] version
- [ ] doc

### Description

A note in the docs

kgajera:
Closes #17191

### DIFF
--- a/packages/angular/cli/models/command.ts
+++ b/packages/angular/cli/models/command.ts
@@ -113,7 +113,8 @@ export abstract class Command<T extends BaseCommandOptions = BaseCommandOptions>
               : '';
           this.logger.info(`  ${colors.cyan('--' + strings.dasherize(o.name))} ${aliases}`);
           if (o.description) {
-            this.logger.info(formatDescription(o.description));
+            const defaultValue = o.default !== undefined ? ` (default: ${o.default})` : '';
+            this.logger.info(formatDescription(o.description) + defaultValue);
           }
         });
     }


packages/angular/cli/models/command.ts
```diff
@@ -115,6 +115,9 @@ export abstract class Command<T extends BaseCommandOptions = BaseCommandOptions>
           if (o.description) {
             this.logger.info(formatDescription(o.description));
           }
+          if (o.default !== undefined) {
+            this.logger.info('    Default: ' + o.default);
```
alan-agius4, line 119: Can we add this on the same line as descriptions please? Similar to what is posted in the original request?

Thanks.

kgajera: Just made this change.

dgp1130:
Going through some old PRs and we came across this one. We should definitely be more transparent about default values, but it's not quite as simple as we would hope. Users can specify a default value at many different configuration points, so printing the default value that was hard-coded in the builder is a bit misleading and could confuse users who have a project configuration that changes the default from the hard-coded value. Whatever we do here needs to take those configurations into account to be clear about the differences between them. More context in https://github.com/angular/angular-cli/issues/17191#issuecomment-858843767.

As a result, this PR isn't very actionable as is, so I'm closing this for now. Adding just the hard-coded default would just confuse people more. We're definitely open to a PR that can take a more comprehensive approach to displaying all the various defaults.

PR #17894, opened 2020-06-10
fix(@ngtools/webpack): only run ngcc sync mode during recompilation

alan-agius4:


alan-agius4:
> it seems that one thing it does is stop pushing pathMappings to ngcc - I assume because ngcc now reads the tsconfig.json file... I this correct? If so, I would put this in a separate commit.
yes

> it seems to be avoiding re-instantiating the NgccProcessor multiple times
Nope, NgccProcessor was always initialised once.

> Finally, can you provide some more information about why you would not run ngcc on the first run?

We don't need to run ngcc sync mode in the first run because we run the ngcc async mode.

petebacondarwin:
I seem to remember that we agreed that the sync-mode integration was a fail-safe catch in case the async-mode integration missed an entry-point, e.g. a local lib in a `dist` folder? Are you confident this is no longer needed? Also if we implement the `use-program-dependencies` there is a small chance that an import might be missed.

alan-agius4:
> I seem to remember that we agreed that the sync-mode integration was a fail-safe catch in case the async-mode integration missed an entry-point, e.g. a local lib in a dist folder? 

AFAIK, that was fixed when NGCC started read the tsconfig, it does seem to work since E2E's for local libraries are green.

> Also if we implement the use-program-dependencies there is a small chance that an import might be missed.
In that case, let's not land this since the gain is pretty low. (91.45500 milliseconds) on a large project

PR #17898, opened 2020-06-10
Update less library to version 3.11.3

vishnutsivan:
less@3.11.1 has dependency with acorn@6.3.0 which has vulnerability issues


vishnutsivan:
@alan-agius4 should I make new pull request? and I didnt find lock file for this project

alan-agius4:
As you prefer but you can push with force on this branch with a valid commit messages and the changes.

The lock file is in the root of the repository.

alan-agius4:
As you prefer, but you can push with force on this branch with a valid commit messages and the changes.

The commit should be something like
```
fix(@angular-devkit/build-angular): update less library to version 3.11.3

Closes:  #17899 
```

The lock file is in the root of the repository.

vishnutsivan:
@alan-agius4 yarn.lock right?

Issue #17757, opened 2020-05-19
karma-coverage-istanbul-reporter author requires payment for support and issue reporting


muuvmuuv:
# 🐞 Bug report

### Command (mark with an `x`)

- [x] new

### Is this a regression?

I don't know but I think it must have worked.

### Description

I have installed a starter through `ng new` leaved everything the default and tried to create a coverage report by enabling `karma-coverage-istanbul-reporter` with `reporters: ['progress', 'coverage-istanbul', 'kjhtml'],`. This is how the author describes the installation on his repo. Sadly this does not work and just prints

```
Statements   : Unknown% ( 0/0 )
Branches     : Unknown% ( 0/0 )
Functions    : Unknown% ( 0/0 )
Lines        : Unknown% ( 0/0 )
```

I tried to fill an issue but this author **REALLY WANTS $100/month** for that!!! This is in acceptable for an OSS project despite the fact that it is coupled to Angular. This makes me angry because new developer will see this and will get a wrong definition of what OSS is and maybe follow this behaviour. Sure, include a sponsor button and say that you don't want to work on it without receiving a _SMALL_ amount of money but then add a "help wanted" badge and ask for help or use issuehunt. I would love to help to solve this issue but nope, [my issue gets closed and locked](https://github.com/mattlewis92/karma-coverage-istanbul-reporter/issues/83) (like many others).

I suggest removing that reporter from the starter and let the author know that this isn't a behaviour Angular supports.

## 🔬 Minimal Reproduction

Fresh starter with coverage enabled.

## 🔥 Exception or Error

<pre><code>
Statements   : Unknown% ( 0/0 )
Branches     : Unknown% ( 0/0 )
Functions    : Unknown% ( 0/0 )
Lines        : Unknown% ( 0/0 )
</code></pre>

## 🌍 Your Environment

<pre><code>
Angular CLI: 9.1.6
Node: 12.16.1
OS: darwin x64

Angular: 9.1.7
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router
Ivy Workspace: Yes

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.901.6
@angular-devkit/build-angular     0.901.6
@angular-devkit/build-optimizer   0.901.6
@angular-devkit/build-webpack     0.901.6
@angular-devkit/core              9.1.6
@angular-devkit/schematics        9.1.6
@angular/cli                      9.1.6
[...]

PR #17901, opened 2020-06-10
feat(@angular-devkit/build-angular): Switch to karma-coverage

kyliau:
This commit switches coverage tooling from karma-coverage-istanbul-reporter
to karma-coverage since it's better supported.

Fixes https://github.com/angular/angular-cli/issues/17757

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
@@ -88,12 +88,32 @@ const init: any = (config: any, emitter: any, customFileHandlers: any) => {
 
   config.reporters.unshift('@angular-devkit/build-angular--event-reporter');
 
-  // When using code-coverage, auto-add coverage-istanbul.
-  config.reporters = config.reporters || [];
-  if (options.codeCoverage && config.reporters.indexOf('coverage-istanbul') === -1) {
-    config.reporters.unshift('coverage-istanbul');
+  // When using code-coverage, auto-add karma-coverage.
+  if (options.codeCoverage) {
+    config.plugins = config.plugins || [];
+    config.reporters = config.reporters || [];
+    const {plugins, reporters} = config;
+    const hasCoveragePlugin = plugins.some((p: {}) => 'reporter:coverage' in p);
+    const hasIstanbulPlugin = plugins.some((p: {}) => 'reporter:coverage-istanbul' in p);
+    const hasCoverageReporter = reporters.includes('coverage');
+    const hasIstanbulReporter = reporters.includes('coverage-istanbul');
+    if (hasCoveragePlugin && !hasCoverageReporter) {
+      reporters.push('coverage');
+    }
+    else if (hasIstanbulPlugin && !hasIstanbulReporter) {
+      // coverage-istanbul is deprecated in favor of karma-coverage
+      reporters.push('coverage-istanbul');
+    }
+    else {
+      throw new Error('karma-coverage must be installed in order to run code coverage');
+    }
+    if (hasIstanbulPlugin) {
+      logger.warn(`'karma-coverage-istanbul-reporter' usage has been deprecated since version 11.\n` +
+       `Please install 'karma-coverage' and update 'karma.conf.js.' ` +
+       'For more info, see https://github.com/karma-runner/karma-coverage/blob/master/README.md');
+    }
   }
-    
+
   // Add webpack config.
   const webpackConfig = config.buildWebpack.webpackConfig;
   const webpackMiddlewareConfig = {
@@ -188,7 +208,7 @@ const init: any = (config: any, emitter: any, customFileHandlers: any) => {
       logger.error(statsErrorsToString(json, statsConfig));
       lastCompilationHash = undefined;
       // Emit a failure build event if there are compilation errors.
-      failureCb && failureCb();
+      failureCb();
     } else if (stats.hash != lastCompilationHash) {
       // Refresh karma only when there are no webpack errors, and if the compilation changed.
       lastCompilationHash = stats.hash;
@@ -269,9 +289,9 @@ const eventReporter: any = function (this: any, baseReporterDecorator: any, conf
 
   this.onRunComplete = function (_browsers: any, results: any) {
     if (results.exitCode === 0) {
-      successCb && successCb();
+      successCb();
     } else {
-      failureCb && failureCb();
+      failureCb();
     }
   }
 
--- a/packages/angular_devkit/build_angular/src/karma/code-coverage_spec.ts
+++ b/packages/angular_devkit/build_angular/src/karma/code-coverage_spec.ts
@@ -5,9 +5,23 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+
+// tslint:disable:no-implicit-dependencies
+
 import { Architect } from '@angular-devkit/architect';
 import { normalize, virtualFs } from '@angular-devkit/core';
+import { last, tap } from 'rxjs/operators';
+import { promisify } from 'util';
 import { createArchitect, host, karmaTargetSpec } from '../test-utils';
+
+// In each of the test below we'll have to call setTimeout to wait for the coverage
+// analysis to be done. This is because karma-coverage performs the analysis
+// asynchronously but the promise that it returns is not awaited by Karma.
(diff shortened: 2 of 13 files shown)


kyliau:
Because the coverage test is run asynchronously in a "runaway" promise, I had to do two things here to make the tests work:
1. Call setTimeout to wait for coverage files to be written
2. Pass Karma results to the callback because the callback is immediately invoked even though the test hasn't actually completed. Passing the Karma results by reference allows us to retrieve the final exit code modified by karma-coverage if the coverage doesn't meet the threshold.

I'm not very satisfied with both approaches, but after discussing with @filipesilva I think this is the best compromise.
I'm totally happy to revise this PR if there are better suggestions.

packages/angular_devkit/build_angular/src/karma/code-coverage_spec.ts
```diff
@@ -7,8 +7,19 @@
  */
 import { Architect } from '@angular-devkit/architect';
 import { normalize, virtualFs } from '@angular-devkit/core';
+import { TestResults } from 'karma';  // tslint:disable-line
```
kyliau, line 10: This is a little weird, because `no-implicit-dependencies` tslint rule insists that `karma` (or rather @types/karma) is one of the dependencies in `package.json`. Is this a normal occurrence?

clydin: Ideally there should be a separate lint configuration for unit test files but the current linting setup does not support multiple configuration files.

packages/angular_devkit/build_angular/src/karma/code-coverage_spec.ts
```diff
@@ -27,12 +38,14 @@ describe('Karma Builder code coverage', () => {
...
 
+    await setTimeoutPromise(1000);
+
     const exists = host.scopedSync().exists(coverageFilePath);
-    expect(exists).toBe(true);
+    expect(exists).toBe(true, `${coverageFilePath} does not exist`);
 
     if (exists) {
       const content = virtualFs.fileBufferToString(host.scopedSync().read(coverageFilePath));
-      expect(content).toContain('polyfills.ts');
```
kyliau, line 35: karma-coverage by default does not check files that are not tested.
If users want coverage for all files, they could turn it on by specifying [`includeAllSources`](https://github.com/karma-runner/karma-coverage/blob/master/docs/configuration.md#includeallsources)

packages/angular_devkit/build_angular/src/karma/code-coverage_spec.ts
```diff
@@ -112,19 +127,17 @@ describe('Karma Builder code coverage', () => {
   }, 120000);
 
   it(`should fail when coverage is below threhold and 'emitWarning' is false`, async () => {
```
alan-agius4, line 129: Can you update the test description? `emitWarning` is no longer a valid option.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
```diff
@@ -72,8 +73,13 @@ const init: any = (config: any, emitter: any, customFileHandlers: any) => {
 
   // When using code-coverage, auto-add coverage-istanbul.
   config.reporters = config.reporters || [];
-  if (options.codeCoverage && config.reporters.indexOf('coverage-istanbul') === -1) {
-    config.reporters.push('coverage-istanbul');
+  // 'coverage-istanbul' comes from karma-coverage-istanbul-reporter package.
```
alan-agius4, line 76: We need to add the report based on the plugin which is configured in the karma.config.js. When users, have configured `karma-coverage-istanbul-reporter`, we need to add `coverage-istanbul`

Something like;
```ts
  // When using code-coverage, auto-add coverage-istanbul.
  if (options.codeCoverage) {
    // Determine which of the supported coverage reporter is being used.
    const pluginIds: string[] = [];
    config.plugins.forEach((p: object) => pluginIds.push(...Object.keys(p)));
   //  karma-coverage
    const hasKarmaCoveragePlugin = pluginIds.includes('reporter:coverage');
   //  karma-coverage-istanbul-reporter
    const hasKarmaIstanbulCoveragePlugin = pluginIds.includes('reporter:coverage-istanbul');

    // @keen - Maybe we shoudl display a warning, if we indent to deprecate this?
    if (hasKarmaIstanbulCoveragePlugin) {
      logger.warn(`Usage of 'karma-coverage-istanbul-reporter' is deprecated since version 10. Please use 'karma-coverage' instead.`)
    }

    if (hasKarmaCoveragePlugin && config.reporters.includes('coverage')) {
      config.reporters.push('coverage');
    } else if (hasKarmaIstanbulCoveragePlugin && config.reporters.includes('coverage-istanbul')) {
      config.reporters.push('coverage-istanbul');
    } else {
      // @keen - Maybe throw that we cannot determine which code coverage reporter is being used?
    }
  }

```

packages/angular_devkit/build_angular/src/karma/code-coverage_spec.ts
```diff
@@ -133,7 +146,13 @@ describe('Karma Builder code coverage', () => {
...
     const run = await architect.scheduleTarget(karmaTargetSpec, { codeCoverage: true });
-    await expectAsync(run.result).toBeResolvedTo(jasmine.objectContaining({ success: false }));
     await run.stop();
+
+    const result = await run.result;
+    // success is true because all tests passed
+    expect(result.success).toBeTrue();
+    const karmaResults: TestResults = result.karmaResults as never;
+    // exitCode is non-zero because coverage failed
+    expect(karmaResults.exitCode).toBe(1);
```
alan-agius4, line 156: With this change, it means that currently with the CLI, there is no way to exit the process with a non zero error code when the

packages/angular_devkit/build_ng_packagr/test/ng-packaged/projects/lib/karma.conf.js
```diff
@@ -16,16 +16,19 @@ module.exports = function (config) {
...
     },
-    coverageIstanbulReporter: {
+    coverageReporter: {
       dir: require('path').join(__dirname, 'coverage'),
-      reports: ['html', 'lcovonly'],
-      fixWebpackSourcePaths: true
+      subdir: '.',
+      reporters: [
+        {type: 'html'},
+        {type: 'lcov'},
```
alan-agius4, line 30: Probably we should also add `{type: 'text-summary'},` here as per the application

kyliau:
Update: even though karma-coverage 2.0.3 was released to address the issue of `onRunComplete()` return a promise, more changes are needed upstream before this could be implemented on CLI side without hacky solutions. This is currently blocked on:
1. https://github.com/karma-runner/karma-coverage/pull/420
1. https://github.com/karma-runner/karma/pull/3541

Since karma-coverage 2.0.3 no longer modifies the karma results in-place, the solution implemented in this PR is no longer applicable.

alan-agius4:
@kyliau, any chance to get in touch with one of the Karma Googler’s to prioritise the above fixes?

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
```diff
@@ -88,12 +88,30 @@ const init: any = (config: any, emitter: any, customFileHandlers: any) => {
...
+    }
+    else if (hasIstanbulPlugin && !hasIstanbulReporter) {
+      // coverage-istanbul is deprecated in favor of karma-coverage
+      reporters.push('coverage-istanbul');
+    }
+    else {
+      throw new Error('karma-coverage must be installed in order to run code coverage');
+    }
+    if (hasIstanbulPlugin) {
+      logger.warn(`'karma-coverage-istanbul-reporter' has been deprecated since version 11. Please install 'karma-coverage' instead.`);
```
alan-agius4, line 111: This is misleading Installing karma-coverage will not be enough, users will need to amend their karma config. Maybe we can mention this as well?

alan-agius4:
@googlebot I consent

alan-agius4 (approved):
LGTM, just one small NIT.

packages/angular_devkit/build_angular/src/angular-cli-files/plugins/karma.ts
kyliau: I've updated the warning to mention updating `karma.conf.js` and added a link to https://github.com/karma-runner/karma-coverage/blob/master/README.md

alan-agius4 (approved):

PR #17906, opened 2020-06-11
Add links to generated configs

alan-agius4:
See individual commits.

Blocked on: https://github.com/angular/angular/pull/37533

### DIFF
--- a/packages/schematics/angular/migrations/update-10/side-effects-package-json.ts
+++ b/packages/schematics/angular/migrations/update-10/side-effects-package-json.ts
@@ -34,7 +34,10 @@ export default function (): Rule {
         const pkgJsonContent = {
           name: strings.dasherize(projectName),
           private: true,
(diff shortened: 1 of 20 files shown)


packages/schematics/angular/migrations/update-10/side-effects-package-json.ts
```diff
@@ -34,7 +34,10 @@ export default function (): Rule {
         const pkgJsonContent = {
           name: strings.dasherize(projectName),
           private: true,
-          description: `This is a special package.json file that is not used by package managers. It is however used to tell the tools and bundlers whether the code under this directory is free of code with non-local side-effect. Any code that does have non-local side-effects can't be well optimized (tree-shaken) and will result in unnecessary increased payload size. It should be safe to set this option to 'false' for new applications, but existing code bases could be broken when built with the production config if the application code does contain non-local side-effects that the application depends on.`,
+          description_1: 'This is a special package.json file that is not used by package managers.',
```
alan-agius4, line 37: In VS code, this is unreadable because it will not be wrapped in multilines line in GitHub. Thus, I split it in multiple descriptions.

PR #17940, opened 2020-06-15
fix(@schematics/angular): disable v10 side effect package.json migration

clydin:


kyliau:
Please remove / disable this test:

https://github.com/angular/angular-cli/blob/88d0eaf885c6a055cf601b11c6190072031bf000/packages/schematics/angular/migrations/update-10/side-effects-package-json_spec.ts#L53-L58

dgp1130:
Also merged to `10.0.x` patch branch.

PR #17948, opened 2020-06-16
Update .editorconfig

Noriller:
Update .editorconfig with "quote_type = single"

Single quotes seems to be the default used, but without this config in .editorconfig it might be overridden by the default of code formatters.

Eg: Using "Prettier" on Visual Studio Code. Since CLI creates a .editorconfig, configurations to enforce single quotes are overridden by the lack of this in .editorconfig.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,7 @@ indent_size = 2
 insert_final_newline = true
 spaces_around_brackets = inside
 trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


Noriller:
@googlebot I signed it!

alan-agius4:
Closing as CI is red. 

Feel free to open a new PR when addressing the failures.

Thanks

PR #17951, opened 2020-06-16
test: skip dedupe modules stderr matching on Node 10

alan-agius4:
This is really flakey on CI Node.js 10.

### DIFF
--- a/tests/legacy-cli/e2e/tests/misc/dedupe-duplicate-modules.ts
+++ b/tests/legacy-cli/e2e/tests/misc/dedupe-duplicate-modules.ts
@@ -4,38 +4,39 @@ import { updateJsonFile } from '../../utils/project';
 import { expectToFail } from '../../utils/utils';
 
 export default async function () {
-    // Force duplicate modules
-    await updateJsonFile('package.json', json => {
-        json.dependencies = {
-            ...json.dependencies,
-            'tslib': '2.0.0',
-            'tslib-1': 'npm:tslib@1.13.0',
-            'tslib-1-copy': 'npm:tslib@1.13.0',
-        };
-    });
+  // Force duplicate modules
+  await updateJsonFile('package.json', json => {
+    json.dependencies = {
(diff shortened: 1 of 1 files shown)


tests/legacy-cli/e2e/tests/misc/dedupe-duplicate-modules.ts
```diff
@@ -4,38 +4,39 @@ import { updateJsonFile } from '../../utils/project';
...
+      })
     `);
 
-    const { stderr } = await ng('build', '--verbose', '--no-vendor-chunk', '--no-progress');
-    if (!/\[DedupeModuleResolvePlugin\]:.+tslib-1-copy -> .+tslib-1/.test(stderr)) {
-        throw new Error('Expected stderr to contain [DedupeModuleResolvePlugin] log for tslib.');
-    }
+  const { stderr } = await ng('build', '--verbose', '--no-vendor-chunk', '--no-progress');
+  const [major] = process.versions.node.split('.').map(Number);
+  if (major > 10 && !/\[DedupeModuleResolvePlugin\]:.+tslib-1-copy -> .+tslib-1/.test(stderr)) {
```
kyliau, line 34: So effectively we're skipping this test in Node 10?
Do you know why it's failing specifically in Node 10?

alan-agius4: We skip the scrapping of stderr, I didn’t manage to replicate the flake locally on Node 10.

Also, actually a flake not a constant failure

alan-agius4: Actually with this change I am seeing the failure on Node 12 as well!

kyliau: OK, so it looks like we'll have to do more investigation and figure out the root cause.

PR #17972, opened 2020-06-19
refactor(@angular-devkit/build-angular): add space after file path in budget warnings

villelahdenvuo:
Visual Studio Code editor has nice link handling that allows one to easily open a file from the terminal output, however it doesn't work if the file path ends with a dot.

Another option would be to add quotes or parentheses around the file path.

Here's a screenshot of the issue:
Screen Shot 2020-06-19 at 17 23 39


villelahdenvuo:
I can look into changing the error messages. I originally did the fix in those places since not all labels are paths, but if we change the message format it should eliminate the issue as well.

PR #17982, opened 2020-06-22
ci: fix dedupe-duplicate-modules flakes

alan-agius4:
Following #17973, which we emitted the stderr when this test failed on CI.

We see that the resolved module order sometimes changes, which is totally fine because of the async nature of the operation.

https://circleci.com/gh/angular/angular-cli/167318#tests/containers/5
```
[DedupeModuleResolvePlugin]: /tmp/angular-cli-e2e-i03vMQ/test-project/node_modules/tslib-1 -> /tmp/angular-cli-e2e-i03vMQ/test-project/node_modules/tslib-1-copy
```

With this change we handle both cases which should eliminate the flakes

### DIFF
--- a/tests/legacy-cli/e2e/tests/misc/dedupe-duplicate-modules.ts
+++ b/tests/legacy-cli/e2e/tests/misc/dedupe-duplicate-modules.ts
@@ -4,20 +4,20 @@ import { updateJsonFile } from '../../utils/project';
 import { expectToFail } from '../../utils/utils';
 
 export default async function () {
-    // Force duplicate modules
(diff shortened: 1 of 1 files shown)


dgp1130:
I saw some failures from this flake on the `10.0.x` branch. Should we also merge this PR into `10.0.x`?

alan-agius4:
@dgp1130 there was a separate PR for that https://github.com/angular/angular-cli/pull/18007

dgp1130:
Gotcha, thanks @alan-agius4!

PR #17984, opened 2020-06-22
[patch] ci: fix dedupe-duplicate-modules flakes

alan-agius4:
Following #17973, which we emitted the stderr when this test failed on CI.

We see that the resolved module order sometimes changes, which is totally fine because of the async nature of the operation.

https://circleci.com/gh/angular/angular-cli/167318#tests/containers/5
```
[DedupeModuleResolvePlugin]: /tmp/angular-cli-e2e-i03vMQ/test-project/node_modules/tslib-1 -> /tmp/angular-cli-e2e-i03vMQ/test-project/node_modules/tslib-1-copy
```

With this change we handle both cases which should eliminate the flakes

(cherry picked from commit bf599e4fac433c0b7041b5fac72d10a1677070cc)


kyliau:
Currently blocked until v10.0.0 is released

PR #18030, opened 2020-06-25
fix(@ngtools/webpack): add ngcc initial run check

clydin:
This change performs a ngcc run check to determine if an initial execution is required. If a run hash file exists that matches the current package manager lock file and the project's tsconfig, then an initial ngcc run has already been performed.
By limiting the need to spawn and execute ngcc in another process on each build, the total build time can be reduced on subsequent builds.

### DIFF
--- a/packages/ngtools/webpack/src/ngcc_processor.ts
+++ b/packages/ngtools/webpack/src/ngcc_processor.ts
@@ -8,7 +8,8 @@
 
 import { LogLevel, Logger, process as mainNgcc } from '@angular/compiler-cli/ngcc';
 import { spawnSync } from 'child_process';
-import { accessSync, constants, existsSync } from 'fs';
+import { createHash } from 'crypto';
+import { accessSync, constants, existsSync, mkdirSync, readFileSync, writeFileSync } from 'fs';
 import * as path from 'path';
 import * as ts from 'typescript';
 import { InputFileSystem } from 'webpack';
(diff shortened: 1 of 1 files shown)


alan-agius4:
@petebacondarwin, in that case ngcc sync would process that, similarly to when a new dependency is added to to a source file.

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -54,6 +55,46 @@ export class NgccProcessor {
       return;
     }
 
+    // Perform a ngcc run check to determine if an initial execution is required.
+    // If a run hash file exists that matches the current package manager lock file and the
+    // project's tsconfig, then an initial ngcc run has already been performed.
+    let skipProcessing = false;
+    let runHashFilePath: string | undefined;
+    const ngccPath = path.join(this._nodeModulesDirectory, '.ngcc');
```
petebacondarwin, line 63: Could you make this file more CLI specific? `.ngcc` feels a bit like it is stepping on someone else's namespace :-)

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -54,6 +55,46 @@ export class NgccProcessor {
...
+        .update(lockFile)
+        .update(tsconfigData)
+        .update(relativeTsconfigPath)
+        .digest('hex');
+
+      runHashFilePath = path.join(ngccPath, runHash + '.lock');
+
+      // If the run hash lock file exists, then ngcc was already run against this project state
+      if (existsSync(runHashFilePath)) {
+        skipProcessing = true;
```
petebacondarwin, line 88: Don't you also need to compare the hash with the contents of the file?

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -54,6 +55,46 @@ export class NgccProcessor {
...
+
+      // Generate a hash that represents the state of the package lock file and used tsconfig
+      const runHash = createHash('sha256')
+        .update(lockData)
+        .update(lockFile)
+        .update(tsconfigData)
+        .update(relativeTsconfigPath)
+        .digest('hex');
+
+      runHashFilePath = path.join(ngccPath, runHash + '.lock');
```
petebacondarwin, line 84: Nice idea to use the hash in the filename, which prevents race conditions on reading the file.

packages/ngtools/webpack/src/ngcc_processor.ts
alan-agius4: It’s actually a folder 😜

petebacondarwin:
I think that this could not fit well with the `--use-program-dependencies` option in ngcc, which only processes the entry-points that are reachable from the source files of the project.

Consider a source file changing to import a new dependency (which was already npm installed but not previously imported). In this case there will have been no changes to the package lock file, the tsconfig or the ngcc config, so the marker will still be valid, and therefore the async ngcc integration will not be triggered.

Of course, when the webpack build happens the sync ngcc integration will catch this and build it, but since it is sync mode, if there is any parallel ngcc process already running this would crash the build.

In the case of local development, this is the same issue as we have when running `ng serve` and a source file changing. So in that case, it would not be problematic. But I could imagine a scenario (even on a CI with node_modules cached) where you check out a new branch that has different source code files, while all the package lockfile, tsconfig and ngcc configs are unchanged. In this case ngcc might not run. If the previous run of ngcc that resulted in the marker being created had no source files (or very few) then it is possible that there are a lot of unprocessed entry-points.

PR #18105, opened 2020-07-01
Improve documentation for major release

kyliau:
This PR documents some of the learnings from the recent v10.0.0 release.
Our release doc does not sufficiently capture the process of a major release, and our tooling is insufficient to support the workflow.

There are two commits:
1. Update the release doc
    - It now reminds the caretaker to double check the versions we are about to release.
    - It adds instructions for major release.
2. Add a script to automatically retag CLI packages.

Since there isn't a test for the scripts, the best I could do is attach the command output.

```
$ yarn run -s admin dist-tag --version 10.0.0 --tag latest --verbose
npm dist-tag add '@angular/cli@10.0.0' 'latest'
npm dist-tag add '@angular/pwa@0.1000.0' 'latest'
npm dist-tag add '@angular-devkit/architect@0.1000.0' 'latest'
npm dist-tag add '@angular-devkit/architect-cli@0.1000.0' 'latest'
npm dist-tag add '@angular-devkit/build-angular@0.1000.0' 'latest'
npm dist-tag add '@angular-devkit/build-ng-packagr@0.1000.0' 'latest'
npm dist-tag add '@angular-devkit/build-optimizer@0.1000.0' 'latest'
npm dist-tag add '@angular-devkit/build-webpack@0.1000.0' 'latest'
npm dist-tag add '@angular-devkit/core@10.0.0' 'latest'
npm dist-tag add '@angular-devkit/schematics@10.0.0' 'latest'
npm dist-tag add '@angular-devkit/schematics-cli@0.1000.0' 'latest'
npm dist-tag add '@ngtools/webpack@10.0.0' 'latest'
npm dist-tag add '@schematics/angular@10.0.0' 'latest'
npm dist-tag add '@schematics/schematics@0.1000.0' 'latest'
npm dist-tag add '@schematics/update@0.1000.0' 'latest'
```

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -101,7 +101,7 @@ merge commits into LTS branches must open a pull request against the specific ba
 In general, cherry picks for LTS should only be done if it meets one of the criteria below:
 
 1. It addresses a critical security vulnerability.
-2. It fixes a breaking change in the external environment.  
+2. It fixes a breaking change in the external environment.
    For example, this could happen if one of the dependencies is deleted from NPM.
 3. It fixes a legitimate failure on CI for a particular LTS branch.
 
@@ -122,9 +122,20 @@ As commits are cherry-picked when PRs are merged, creating the release should be
 ```bash
 git commit -a -m 'release: vXX'
 git tag 'vXX'
+```
 
-# Make sure to run these commands together, as missing tags can cause CI
-# failures.
+The package versions we are about to publish are derived from the git tag that
+we just created. Double check that the versions are correct by running the
+following command.
+
+```bash
+yarn admin packages --version
+```
+
+Now push the commit and the tag to the upstream repository.
+**Make sure to run these commands together, as missing tags can cause CI failures.**
+
+```bash
 git push upstream && git push upstream --tags
 ```
 
@@ -151,7 +162,10 @@ After closing the tab, you have successfully logged in, it is time to publish.
 
 **It is a good idea to wait for CI to be green on the patch branch and tag before doing the release.**
 
-Check out the patch branch (e.g. `9.1.x`), then run:
+For the first release of a major version, follow the instructions in
+[Publishing a Major Version](#publishing-a-major-version) section.
+
+For non-major release, check out the patch branch (e.g. `9.1.x`), then run:
 ```bash
 yarn # Reload dependencies
 yarn admin publish
@@ -208,3 +222,23 @@ If you don't have the firebase CLI installed, you can install it using
 `npm install --global firebase-tools` (or use your package manager of choice).
 
 This is detailed in [`etc/cli.angular.io/README.md`](https://github.com/angular/angular-cli/blob/master/etc/cli.angular.io/README.md).
+
+## Publishing a Major Version
+
+For the first release of a major version, say `v10.0.0`, checkout the major branch
+(i.e. `10.0.x`), then run:
+
+```bash
+yarn # Reload dependencies
+yarn admin publish --tag next # a major release is always tagged as next initially
+```
+
+Confirm with downstream repositories (Components, etc) that everything is ok.
+Once the release is stable, wait for Framework to retag their packages, then
+retag the CLI packages as `latest`.
+The command below will automatically retag stable packages as well as experimental
+packages.
+
+```bash
+yarn admin dist-tag --version 10.0.0 --tag latest
+```
--- a/scripts/dist-tag.ts
+++ b/scripts/dist-tag.ts
@@ -6,7 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-// tslint:disable: no-implicit-dependencies
+// tslint:disable: no-implicit-dependencies we import @angular-devkit/core but
+// it is not in package.json, which is fine, this is just a script.
 
 import { logging } from '@angular-devkit/core';
 import { execSync } from 'child_process';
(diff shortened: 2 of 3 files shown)


docs/process/release.md
```diff
@@ -128,6 +128,14 @@ git tag 'vXX'
 git push upstream && git push upstream --tags
 ```
 
+Double check that the packages we are about to publish are versioned correctly
+by running the following command.
+Their versions are derived from the git tag that we just created.
+
+```
+yarn admin release
```
dgp1130, line 136: Does this just print out some metadata? It sounds like a very scary command to run and I'm not sure I would be comfortable executing this without fully understanding what it does.

Could we rename it, or at least document more clearly that this won't actually trigger a release? Or maybe it does and I'm just not understanding how this works.

docs/process/release.md
```diff
@@ -208,3 +219,23 @@ If you don't have the firebase CLI installed, you can install it using
 `npm install --global firebase-tools` (or use your package manager of choice).
 
 This is detailed in [`etc/cli.angular.io/README.md`](https://github.com/angular/angular-cli/blob/master/etc/cli.angular.io/README.md).
+
+## Publishing a Major Version
+
+For the first release of a major version, say `v10.0.0`, checkout the patch branch
```
dgp1130, line 225: > checkout the patch branch

Nit: I guess it is the `10.0.x` branch, but wouldn't we call this the major? Probably good to avoid name collisions with other branches.

docs/process/release.md
dgp1130, line 136: Should this happen before the commit or tag or push? I've had to revert and then recommit or retag a few times in some releases. It would be good to sanity check before we push the release commit.

dgp1130, line 136: Should we run `yarn` before this to reload dependencies like the other commands? The user likely just checked out this branch and may have different dependencies loaded.

scripts/dist-tag.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * This function adds a tag to all public packages in the CLI repo.
+ */
+export default function(args: DistTagOptions, logger: logging.Logger) {
+  const {version, tag} = args;
+  if (!version || version.startsWith('v')) {
+    throw new Error('Version must be specified in format d+.d+.d+');
+  }
+  if (version.startsWith('0')) {
+    throw new Error('Version must be "stable", with major version > 0');
```
dgp1130, line 37: If I saw this error message I'm not sure I'd understand it. My interpretation looking the code is that this is trying to disallow `0.x.x` packages. Maybe we can state this a little more clearly? The definition of "stable" is a bit vague.

The doc comment for `version` is also a bit unclear because of its use of "stable".

scripts/dist-tag.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { logging } from '@angular-devkit/core';
+import { execSync } from 'child_process';
+import { packages, stableToExperimentalVersion } from '../lib/packages';
+
+interface DistTagOptions {
+  /**
+   * Version must be specified in format d+.d+.d+ where d is a 0-9 digit.
+   * This must be a stable version with major version > 0.
+   * The script will automatically convert stable version to experimental.
```
dgp1130, line 19: > The script will automatically convert stable version to experimental.

The script will automatically convert stable version to experimental for experimental packages.

scripts/dist-tag.ts
```diff
@@ -0,0 +1,51 @@
...
+
+// tslint:disable: no-implicit-dependencies
+
+import { logging } from '@angular-devkit/core';
+import { execSync } from 'child_process';
+import { packages, stableToExperimentalVersion } from '../lib/packages';
+
+interface DistTagOptions {
+  /**
+   * Version must be specified in format d+.d+.d+ where d is a 0-9 digit.
```
dgp1130, line 17: Nit: Should give an example here so humans don't need to parse regexes. Maybe include:

> For example, `1.2.3` or `10.0.0`.

Is there any scenario where we would want to use `-next` or `-rc`? I'm guessing not, but we should explicitly state one way or the other in the doc.

scripts/dist-tag.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Tag is usually "latest" or "next", but could also be "v10-lts" for example.
+   */
+  tag: string;
+}
+
+/**
+ * This function adds a tag to all public packages in the CLI repo.
+ */
+export default function(args: DistTagOptions, logger: logging.Logger) {
```
dgp1130, line 31: Nit: I'm not too familiar with the framework, but is there any help or explanation content that would display for this subcommand?

Can we do `yarn admin dist-tag --help` or `yarn admin help dist-tag` and get anything useful? If that's not supported then consider adding `--help` as an argument. It isn't too big a deal, but if this is easy then we should make sure to add the relevant metadata to support it.

scripts/dist-tag.ts
```diff
@@ -0,0 +1,51 @@
...
+    throw new Error('Version must be "stable", with major version > 0');
+  }
+  if (!tag) {
+    throw new Error('Tag must be non-empty, for example: latest, next, v10-lts, etc');
+  }
+  const publicPackages = Object.values(packages).filter(p => !p.private);
+  for (const {name, experimental} of publicPackages) {
+    const actualVersion = experimental ? stableToExperimentalVersion(version) : version;
+    // See https://docs.npmjs.com/cli/dist-tag for documentation
+    const cmd = `npm dist-tag add ${name}@${actualVersion} ${tag}`;
```
dgp1130, line 46: Nit: I'd recommend quoting the arguments just to make sure shell escaping doesn't convert the expanded command into something we don't expect (probably fine but good to be safe).

```
const cmd = `npm dist-tag add "${name}@${actualVersion} ${tag}"`;
```

Alternatively we can use the form of `exec()` which takes the command name and array of arguments. That would avoid any shell weirdness that might be unexpected.

scripts/dist-tag.ts
```diff
@@ -0,0 +1,51 @@
...
+    logger.debug(cmd);  // print debug output by specifying --verbose
+    const output = execSync(cmd, { encoding: 'utf8' });
```
dgp1130, line 48: Nit: Consider using the async version. Really doesn't matter for such a simple script right now, but things do get more complicated over time and it's good to avoid "broken windows" early on.

docs/process/release.md
kyliau: I've created another PR to cleanup and rename `scripts/release.ts`.
https://github.com/angular/angular-cli/pull/18184

docs/process/release.md
kyliau: renamed patch branch to major branch

docs/process/release.md
kyliau: Good call. It should be commit & tag -> check versions -> push. I've updated the order.

kyliau: The `release` script is really just used to check the package versions, so no yarn install is necessary.
I've renamed `release.ts` to `versions.ts`, so this should be less confusing.

scripts/dist-tag.ts
kyliau: Removed all mentions of "stable", and provided suggestion instead.

scripts/dist-tag.ts
kyliau: Mentioned both -next and -rc.

scripts/dist-tag.ts
kyliau: added quotes, since `exec()` does not accept arguments.

scripts/dist-tag.ts
kyliau: Making this async is rather troublesome, because the "framework" doesn't really support async operations.
Since this is mostly a script, I think there is negligible difference in performance.

scripts/dist-tag.ts
kyliau: The "framework" aka devkit-admin doesn't really support help, since it's mostly a loose collection of scripts.
I've added `--help` as a command line argument since that's the easiest.

docs/process/release.md
```diff
@@ -122,9 +122,20 @@ As commits are cherry-picked when PRs are merged, creating the release should be
...
+following command.
+
+```bash
+yarn admin packages --version
+```
 
-# Make sure to run these commands together, as missing tags can cause CI
-# failures.
+Now push the commit and the tag to the upstream repository.
+# Make sure to run these commands together, as missing tags can cause CI failures.
```
dgp1130, line 136: Nit: This should probably be bolded rather than a header, not sure why it was a header already.

scripts/dist-tag.ts
```diff
@@ -0,0 +1,70 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// tslint:disable: no-implicit-dependencies
```
dgp1130, line 9: Nit: Should give a reason this lint check is disabled.

scripts/dist-tag.ts
dgp1130: Sorry, should have been more explicitly, I was referring to [`execFileSync()`](https://nodejs.org/api/child_process.html#child_process_child_process_execfilesync_file_args_options). Up to you if you want to use that instead as a "safer" version.

docs/process/release.md
kyliau: changed to bold

PR #18106, opened 2020-07-01
build: update validate-commits script to list valid types and scopes

kyliau:
The commit linter expects a certain type and/or scope in the commit message,
but if the right type/scope is not provided it doesn't tell us what to do.

Instead of this:
```
The following commit has an unknown type. You can use wip: to avoid this.
```

It now outputs this:
```
The following commit has an unknown type. It must be one of [docs, refactor, style, test, feat, fix, build, revert, ci, release]. You can use wip: to avoid this.
```

### DIFF
--- a/scripts/validate-commits.ts
+++ b/scripts/validate-commits.ts
@@ -154,25 +154,27 @@ export default function (argv: ValidateCommitsOptions, logger: logging.Logger) {
     }
 
     const [type, scope] = subject.slice(1);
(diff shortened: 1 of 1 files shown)


alan-agius4:
@kyliau with your change you actually print out the types and not the scopes. I do get confused sometimes with scope/type convention. 😜

I do agree with @dgp1130, that we should also print the valid scopes.

kyliau:
Revised the script to print out valid types and scopes.

PR #18159, opened 2020-07-06
refactor(@angular-devkit/build-angular): improve budget warning messages

villelahdenvuo:
Visual Studio Code editor has nice link handling that allows one to easily open a file from the terminal output, however it doesn't work if the file path ends with a dot.

Here's a screenshot of the issue:
image


villelahdenvuo:
@alan-agius4 I made some improvements to the PR based on your comment, but I closed the old PR by accident so here's a new one.

villelahdenvuo:
@alan-agius4 hey, just dropping a comment so that the PR won't be automatically discarded. Have a nice one!

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
```diff
@@ -289,7 +290,7 @@ class AnyCalculator extends Calculator {
       .filter(asset => !asset.name.endsWith('.map'))
       .map(asset => ({
         size: asset.size,
-        label: asset.name,
+        label: relative(process.cwd(), asset.name),
```
alan-agius4, line 293: While I like the idea of having relative paths, I think this needs to be taken a more holistic and generic approach. Currently most if not error errors/warnings are absolute.

alan-agius4:
@villelahdenvuo, thanks for the reminder.
Left a small comment about relative paths in errors.

packages/angular_devkit/build_angular/src/angular-cli-files/utilities/bundle-calculator.ts
villelahdenvuo: I can remove that and it can be solved upstream. It did feel a bit hacky since there is no base path available.

alan-agius4: I think we can do this in the CLI but at a "global" level.

Let's remove this change for the time being.

villelahdenvuo:
@alan-agius4 Should be in order now, looks the CI failed on some missing package, though.

alan-agius4:
@villelahdenvuo, there is currently an ongoing NPM issue https://status.npmjs.org/

alan-agius4:
@villelahdenvuo, just noticed that your commits are not squashed. Can you please squash them? Thanks.

villelahdenvuo:
@alan-agius4 I cleaned up the commits.

alan-agius4:
Thanks

villelahdenvuo:
Thanks guys ✨

PR #18164, opened 2020-07-06
refactor(@angular-devkit/build-angular): display translation-file dia…

petebacondarwin:
…gnostics

If no translation parsers could parse a translation-file
we displayed a generic error, which makes it difficult to
track down if there is a problem with the file.

https://github.com/angular/angular/pull/37909 introduces
a new `parser.analyze()` API that allows us to get hold
of the diagnostic messages from trying to parse the
translation-files.

This commit will render these messages if none of the
translation parsers are succcesful.

Note that in order to maintain compatibility with versions
of `@angular/localize` that do not have the `analyze()`
method, the commit includes a polyfill, which can be removed
after a release where we can guarantee that the method
will be available.


petebacondarwin:
Because of the polyfill, this PR is not actually blocked on angular/angular#37909

PR #18196, opened 2020-07-08
fix(@schematics/cli): only show prompts in TTY terminals

alan-agius4:


packages/angular_devkit/schematics_cli/bin/schematics.ts
```diff
@@ -337,6 +338,8 @@ function getUsage(): string {
       --list-schematics   List all schematics from the collection, by name. A collection name
                           should be suffixed by a colon. Example: '@schematics/schematics:'.
 
+      --interactive       Enables interactive input prompts.
```
clydin, line 341:
```suggestion
      --no-interactive       Disables interactive input prompts.
```
They are enabled by default so the flag would only be used to disable them.
Yarn's option, for instance:
```
--non-interactive                   do not show interactive prompts
```

alan-agius4: I agree, this is better

PR #18219, opened 2020-07-10
Fixes publish and dist-tag scripts.

dgp1130:
This fixes a few known bugs in `publish` and `dist-tag`, as well as a few new ones I found along the way.

Ideally this would have been at least two distinct PRs, but they touched each other when it relates to registries in a way that would have been a pain with multiple PRs.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -168,13 +168,23 @@ For the first release of a major version, follow the instructions in
 For non-major release, check out the patch branch (e.g. `9.1.x`), then run:
 ```bash
 yarn # Reload dependencies
-yarn admin publish
+yarn admin publish --tag latest
 ```
 
 If also publishing a prerelease, check out `master`, then run:
 ```bash
 yarn # Reload dependencies
 yarn admin publish --tag next
+```
+
+If also publish an LTS branch, check out that patch branch (e.g. `8.3.x`), then
+run:
+
+**Make sure to update the NPM tag for the version you are releasing!**
+
+```bash
+yarn # Reload dependencies
+yarn admin publish --tag v8-lts
 ```
 
 ### Release Notes
--- a/scripts/dist-tag.ts
+++ b/scripts/dist-tag.ts
@@ -37,7 +37,7 @@ interface DistTagOptions {
 /**
  * This function adds a tag to all public packages in the CLI repo.
  */
-export default function(args: DistTagOptions, logger: logging.Logger) {
+export default function(args: Partial<DistTagOptions>, logger: logging.Logger) {
   if (args.help) {
     logger.info(`dist-tag adds a tag to all public packages in the CLI repo.
 
--- a/scripts/dist-tag.ts
+++ b/scripts/dist-tag.ts
@@ -12,6 +12,7 @@
 import { logging } from '@angular-devkit/core';
 import { execSync } from 'child_process';
 import { packages, stableToExperimentalVersion } from '../lib/packages';
+import { wombat } from '../lib/registries';
 
 interface DistTagOptions {
   /**
@@ -24,10 +25,15 @@ interface DistTagOptions {
    * experimental packages.
    */
   version: string;
+
   /**
    * Tag is usually "latest" or "next", but could also be "v10-lts" for example.
    */
   tag: string;
+
+  /** The package registry to tag. */
+  registry: string;
+
   /**
    * If true, prints the help message.
    */
@@ -50,7 +56,7 @@ Usage:
 
     return;
   }
-  const {version, tag} = args;
+  const {version, tag, registry: registryArg} = args;
   if (!version || version.startsWith('v')) {
     throw new Error('Version must be specified in format d+.d+.d+');
   }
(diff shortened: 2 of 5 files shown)


docs/process/release.md
```diff
@@ -177,6 +177,15 @@ yarn # Reload dependencies
...
 ```
 
+If also publish an LTS branch, check out that patch branch (e.g. `8.3.x`), then
+run:
+```bash
+yarn # Reload dependencies
+yarn admin publish --tag v8-lts
+```
+
+**Make sure to update the tag for the version you are releasing!**
```
kyliau, line 187: We should probably mention this before the release steps.
Also `tag` here is overloaded, since it could be git tag or npm tag.
It's best to be explicit.

scripts/dist-tag.ts
```diff
@@ -60,11 +66,12 @@ Usage:
   if (!tag) {
     throw new Error('Tag must be non-empty, for example: latest, next, v10-lts, etc');
   }
+  const registry = registryArg ?? wombat;
```
kyliau, line 69: The wombat registry will not work in this case, I think.
@mhevery tried it once and ran into authentication errors.
It has to be the default NPM registry.

Also, since registry is non-optional, we don't need nullish-coalescing.

scripts/publish.ts
```diff
@@ -10,16 +10,39 @@ import { logging, tags } from '@angular-devkit/core';
 import { spawnSync } from 'child_process';
 import * as semver from 'semver';
 import { packages } from '../lib/packages';
+import { wombat } from '../lib/registries';
 import build from './build';
 
-
 export interface PublishArgs {
   tag?: string;
```
kyliau, line 17: make tag non-optional?

scripts/publish.ts
```diff
@@ -10,16 +10,39 @@ import { logging, tags } from '@angular-devkit/core';
...
-  branchCheck?: boolean;
-  versionCheck?: boolean;
+  branchCheck?: string; // Represents a boolean (should be "true" or "false").
```
kyliau, line 18: This is a messy situation, I think, due to the dynamic way we parse command line arguments. 
If it is specified in the command line as `--branchCheck` then it'll be a Boolean.
Personally I think keeping it as a boolean makes sense, so we don't have to guess what the possible values are.

scripts/dist-tag.ts
dgp1130: Tested this and Wombat does work with the `npm dist-tag` command.

`--registry` is optional, defaulting to Wombat, so I believe this is necessary.

scripts/publish.ts
dgp1130: Semantically `--tag` is required, but when given the `PublishArgs` type, we haven't validated that yet. We can't assume that any arguments are provided in this type.

This is a really unfortunate system, so there's not a great way of doing this. I think the best solution would be to make `PublishArgs` semantically correct (`tag` is required), but then in the default exported function accept `Partial<PublishArgs>`, that way everything is forced to be possibly `undefined` and must be manually validated at runtime.

Even that wouldn't really handle incorrectly typed values so it's not quite fool-proof (such as the `--branchCheck=false` case), but the only way to handle that would be to accept an `unknown` or a `Record<string, unknown>`... I can't think of a better solution that doesn't involve re-writing the way we process commands as a whole. WDYT?

scripts/publish.ts
dgp1130: Ah, I see. The error messages suggested using `--*Check=false`, I thought this was an existing bug where we were getting strings when we expected booleans. Now I see that minimist does support this, it's just that those messages were wrong.

After testing, it seems that `--no-*` is correctly interpreted as a boolean set to `false`. I updated the error messages (and a test) to use `--no-*` and reverted the types to booleans. (The casing is a little weird as it mixes skewer and camel case `--no-branchCheck`, but I think that's a problem for another day).

dgp1130:
@clydin, do you want to take a look at this change as well?

kyliau:
@dgp1130, please open separate patch PR for each LTS that you'd like to target. Currently active LTS are v8 and v9

tests/legacy-cli/e2e/setup/010-local-publish.ts
```diff
@@ -8,8 +8,8 @@ export default async function() {
     'admin',
     '--',
     'publish',
-    '--versionCheck=false',
-    '--branchCheck=false',
+    '--no-versionCheck',
+    '--no-branchCheck',
     '--registry=http://localhost:4873',
```
clydin, line 13: Now that `--tag` is required,  I think this file also needs to add `--tag latest` to the argument list when not a prerelease.  `--tag next` is already added in the prerelease case on line 18.

clydin:
For v8, the nullish coalescing will need to be changed. It's still on TS 3.5.

tests/legacy-cli/e2e/setup/010-local-publish.ts
dgp1130: Done.

dgp1130:
Made separate PRs for LTS branches:
9.1.x: https://github.com/angular/angular-cli/pull/18324
8.3.x: https://github.com/angular/angular-cli/pull/18323

dgp1130:
@clydin, thanks for pointing out the TS 3.5 issue on v8. Updated that PR to not use `??`.

alan-agius4:
@dgp1130, seems like the CLA bot status is stalled.

dgp1130:
@alan-agius4, strange. This PR was stuck on CLA, while the other two were stuck on 'Missing required status "ci/circleci: integration"'. I did an empty amend and force pushed it, so that should re-trigger all the status checks.

Issue #18220, opened 2020-07-10
Lint Issue: Angular CLI generate  `app-routing.module.ts` with consecutive blank lines  


kumaran-is:
# 🐞 Bug report

### Command (mark with an `x`)
`app-routing.module.ts` generated by `Angular-cli v9`  has consecutive blank lines( line 3 & 4 in below screenshot) , causing lint issue. This was not happening in `Angular-cli v8`.

![Screen Shot 2020-07-10 at 3 50 59 PM](https://user-images.githubusercontent.com/6869139/87207410-d4141880-c2d9-11ea-9d56-09fd79c4c62f.png)

with tslint.json configuration, 

"rules": {
    "no-consecutive-blank-lines":true
}

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [X] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

This was not happening in `Angular-cli v8`.

### Description

As mentioned above. 

## 🔬 Minimal Reproduction

- Install angular cli 9.1  `npm install -g @angular/cli@9`
- Create a new project `ng new my-app`  with route 
- Update your tslint.json configuration with below rule

"rules": {
    "no-consecutive-blank-lines": true
}

- Run. `ng lint`

- Lint fails with below error

/Users/work/Angular9/my-app/src/app/app-routing.module.ts:4:1
ERROR: 4:1  no-consecutive-blank-lines  Consecutive blank lines are forbidden

Lint errors found in the listed files.
npm ERR! code ELIFECYCLE
npm ERR! errno 1

## 🌍 Your Environment
Angular CLI: 9.1.11
Node: 12.13.0
OS: darwin x64

Angular: 
... 
Ivy Workspace: 

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.901.11
@angular-devkit/core         9.1.11
@angular-devkit/schematics   9.1.11
@schematics/angular          9.1.11
@schematics/update           0.901.11
rxjs                         6.5.4

sacgrover:
Please assign to me

PR #18226, opened 2020-07-11
fix(@schematics/angular): consecutive blank lines in routing.module.ts

sacgrover:
Fixes: #18220


sacgrover:
@alan-agius4 Ok, I will do that.. Actually, I made the same earlier but to make it consistent with other I modified. Because there are a blank line between module import and component import in other places.

alan-agius4:
@sacgrover, I see, in that case there should be a blank line in the logical statement and before the constant declaration.

```ts
import { Routes, RouterModule } from '@angular/router';<% if (lazyRoute) { %>

import { <%= classify(name) %>Component } from './<%= dasherize(name) %>.component';<% } %>

const routes: Routes = [<% if (lazyRoute) { %>{ path: '', component: <%= classify(name) %>Component }<% } %>];
```

sacgrover:
Changes done as per asked.

sacgrover:
> LGTM, thanks

Could You please rerun it as i don't have permissions for it?

PR #18232, opened 2020-07-12
fix(@angular-devkit/build-angular): wrap ES5 differential loading bundles

clydin:
This change ensures that classic (ES5) script's top-level function helpers do not get overwritten by other scripts top-level functions that happen to have the same name.  This is not an issue when using module script types because each module has its own scope.

Closes https://github.com/angular/angular/issues/36521


packages/angular_devkit/build_angular/src/browser/specs/differential_loading_spec.ts
```diff
@@ -10,6 +10,7 @@ import { Architect } from '@angular-devkit/architect';
 import { PathFragment } from '@angular-devkit/core';
 import { browserBuild, createArchitect, host } from '../../test-utils';
 
+// tslint:disable-next-line: no-big-function
```
petebacondarwin, line 13: Clearly you have written too many tests!

PR #18239, opened 2020-07-13
Update contributing.ejs

Shivam7-1:


alan-agius4:
It seems that you didn't run `yarn admin templates`

alan-agius4:
Addressed in https://github.com/angular/angular-cli/pull/18440

PR #18255, opened 2020-07-14
[EXPERIMENT] feat(@angular-devkit/build-angular): optimize names of internal AOT properties

clydin:


packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/common.ts
```diff
@@ -411,21 +411,31 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
+      mangle: allowMangle &&
+        buildOptions.platform !== 'server' &&
+        !differentialLoadingMode && {
+          properties: {
+            regex: /^(?:ɵfac|ɵcmp|ɵinj|ɵmod)$/,
+          },
+        },
+      nameCache: {
+        vars: { props: {} },
+        props: { props: { $ɵmod: 'm', $ɵinj: 'i', $ɵfac: 'f', $ɵcmp: 'c' } },
```
petebacondarwin, line 437: What happens if a component (say) already has a property called `c`?

alan-agius4: These changes would also brea analytics build stats

clydin: The existing static property gets overwritten by the Angular property.  This is the same as current behavior (although less likely with the more exotic name) and is really a consequence of using string property names.  To fix this we could consider using symbol names instead.  They work well as a metadata store.

petebacondarwin: Yeah but `ɵcmp` is much less likely to be used on a component than `c`...

clydin: I did mention that :)

Alternatively, we could try running terser serially and let it pick a unique name; although at the cost of build time.

alfaproject: I can see the appeal but should we really try to reinvent compression? Surely, everyone delivers static assets compressed by gzip or brotli or...

alfaproject: I see, thank you for expanding. (:

PR #18306, opened 2020-07-20
docs: fix typo in ng update --next description

vltansky:
`--next` arg description probably should be 
Use the **latest** version, including beta and RCs and not **largest**

### DIFF
--- a/packages/angular/cli/commands/update.json
+++ b/packages/angular/cli/commands/update.json
@@ -38,7 +38,7 @@
           "type": "boolean"
         },
         "next": {
-          "description": "Use the latest version, including beta and RCs.",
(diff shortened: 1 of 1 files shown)


vltansky:
@googlebot I signed it!

alan-agius4:
@vltansky, I think this was not a mistake, as latest, might get confused with latest tag.

Maybe a better word would be `prerelease.`

vltansky:
@alan-agius4 updated to `prerelease`

alan-agius4:
@vltansky thanks.

alan-agius4:
**Caretaker note**: please squash commits when merging.

PR #18316, opened 2020-07-21
fix(@schematics/angular): remove compileComponents from component test schematic

cexbrayat:
`compileComponents` is not necessary when using the CLI and just adds boilerplate code, so we can remove it from the test schematic and make it independent from `async/await` (only place we would have it in the CLI generated code, and in most Angular apps).

### DIFF
--- a/integration/angular_cli/src/app/app.component.spec.ts
+++ b/integration/angular_cli/src/app/app.component.spec.ts
@@ -2,13 +2,9 @@ import { TestBed } from '@angular/core/testing';
 import { AppComponent } from './app.component';
 
 describe('AppComponent', () => {
-  beforeEach(async () => {
-    await TestBed.configureTestingModule({
-      declarations: [
-        AppComponent
-      ],
-    }).compileComponents();
-  });
+  beforeEach(() => TestBed.configureTestingModule({
+    declarations: [AppComponent]
+  }));
 
   it('should create the app', () => {
     const fixture = TestBed.createComponent(AppComponent);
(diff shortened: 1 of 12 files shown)


cexbrayat:
@alan-agius4 Now that we are in v11, do you think we can land this PR?

alan-agius4:
Let me bring this out with the rest of the team in our next team meeting.

dgp1130:
We're planning some further investigation of the `TestBed` API and ways that we might be able to simplify it, particularly follow View Engine removal. Some better understanding there might help us remove the underlying need for `compileComponents()`.

In our discussion today, the immediate concern is around portability. If a developer runs tests via direct Karma or Jest integration, they must use `compileComponents()` in order to handle the `templateUrl` parameter. If we remove this from the generated components, it could cause confusing errors which can only be fixed by special build plugins. Jest's Angular preset already [seems to have one](https://github.com/thymikee/jest-preset-angular/blob/5cfe109a61519a879639d83d1e127781f5b11eab/src/transformers/inline-files.ts#L96-L123), though this is extra configuration that would be required to work with any external build tooling. It's more portable and far simpler from a tooling perspective to include `compileComponents()`, even if it may not be necessary for `ng test` use cases.

All that said, we'll need to revisit this once we have a better understanding of the future of `TestBed`.

dgp1130:
We've been looking through some older PRs and came across this one. Landing this largely depends on future of TestBed which we still don't have a great strategy for right now and likely won't for some time. It's unlikely this PR will be able to go anywhere without a better idea of how TestBed will evolve, and we don't have a timeline for when that will happen.

Whenever it does happen, we'll likely need to update schematics anyways and possibly change more than just `compileComponents()` so it makes to handle this as part of that change instead of removing `compileComponents()` preemptively or saving this PR until it can actually be merged. As a result, I'm closing this PR, I imagine we'll revist `compileComponents()` when the time is right and update the schematic accordingly.

PR #18317, opened 2020-07-21
fix(@schematics/schematics): generate new schematics using TypeScript 3.9

alan-agius4:


packages/schematics/schematics/schematic/files/package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "@angular-devkit/core": "^<%= coreVersion %>",
     "@angular-devkit/schematics": "^<%= schematicsVersion %>",
-    "typescript": "~3.8.2"
+    "typescript": "~3.9.2"
```
clydin, line 18: Nit: might as well increase the patch version to the latest also.

alan-agius4: Done

PR #18318, opened 2020-07-21
refactor: clean up several files

alan-agius4:
Most of these changes are requires for TS 4 update

### DIFF
--- a/packages/angular_devkit/schematics_cli/bin/schematics.ts
+++ b/packages/angular_devkit/schematics_cli/bin/schematics.ts
@@ -267,7 +267,8 @@ export async function main({
       return argv._;
     }
   });
-  delete parsedArgs._;
+
+  parsedArgs._  = [];
 
   // Add prompts.
   if (argv['interactive'] && isTTY()) {
--- a/packages/ngtools/webpack/src/transformers/make_transform.ts
+++ b/packages/ngtools/webpack/src/transformers/make_transform.ts
@@ -99,8 +99,9 @@ export function makeTransform(
 // If TS sees an empty decorator array, it will still emit a `__decorate` call.
 //    This seems to be a TS bug.
 function cleanupDecorators(node: ts.Node) {
-  if (node.decorators && node.decorators.length === 0) {
-    node.decorators = undefined;
+  if (node.decorators?.length === 0) {
+    // tslint:disable-next-line:no-any
+    (node as any).decorators = undefined;
   }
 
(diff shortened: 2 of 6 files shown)


packages/ngtools/webpack/src/transformers/make_transform.ts
```diff
@@ -99,8 +99,9 @@ export function makeTransform(
 // If TS sees an empty decorator array, it will still emit a `__decorate` call.
 //    This seems to be a TS bug.
 function cleanupDecorators(node: ts.Node) {
-  if (node.decorators && node.decorators.length === 0) {
-    node.decorators = undefined;
+  if (node.decorators?.length === 0) {
+    // tslint:disable-next-line:no-any
+    (node as any).decorators = undefined;
```
clydin, line 104: I'm curious if this workaround is still needed at all now.

packages/angular_devkit/schematics_cli/bin/schematics.ts
```diff
@@ -267,7 +267,8 @@ export async function main({
       return argv._;
     }
   });
-  delete parsedArgs._;
+
+  parsedArgs._  = [];
```
clydin, line 271: The behavior is now different.  Does this have any effect on usages of the property?

alan-agius4: This property is marked as non optional thus it cannot be deleted.

From the schematics it seems that an undefined or empty array have the same behaviour.

packages/ngtools/webpack/src/transformers/make_transform.ts
alan-agius4: Can take a look when doing the upgrade.

alan-agius4: @clydin just checked this, and it seems that the issue is still present.

Expected
```
import { __decorate } from "tslib"; import { AnotherDecorator } from 'another-lib'; export class AppComponent { constructor() { this.title = 'app'; } onEscape() { console.log('run'); } } __decorate([ AnotherDecorator() ], AppComponent.prototype, "onEscape", null); AppComponent = __decorate([], AppComponent)
```
to equal
```
import { __decorate } from "tslib"; import { AnotherDecorator } from 'another-lib'; export class AppComponent { constructor() { this.title = 'app'; } onEscape() { console.log('run'); } } __decorate([ AnotherDecorator() ], AppComponent.prototype, "onEscape", null);'.
```

PR #18319, opened 2020-07-21
Support Typescript 4

alan-agius4:


packages/schematics/schematics/blank/project-files/package.json.template
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "@angular-devkit/core": "^<%= coreVersion %>",
     "@angular-devkit/schematics": "^<%= schematicsVersion %>",
-    "typescript": "~3.9.7"
+    "typescript": "~4.0.2-insiders.20200818"
```
clydin, line 18: Is this version intentional? “schematics/schematic” contains this version as well.

alan-agius4: Oops missed that.

Michael-Ziluck:
What version of Angular would we expect to see this in? I'm guessing 10.1 but would like confirmation. I try to avoid pings but am in the process of an upgrade for work and am curious when we can jump TypeScript versions as well. Was there an issue for this work that might have more details that I'm missing?

clydin:
It will be available in 10.1.0.  The recently released prerelease version of 10.1.0 has support included.  You can try it with your project via `ng update --next @angular/cli`

alan-agius4:
No, there wasn’t any issue for this. For more information please check the
TypeScript release 4.0 release post.

We’ll be releasing this as part of 10.1

PR #18322, opened 2020-07-21
ci: update release docs to use annotated tags for releases

dgp1130:
This changes from using lightweight tags to annotated tags. The later includes author, time, and message information that is lacking from the former. It is also included in Git's `--follow-tags` option. This should allow us to push tags alongside release commits more atomically and hopefully eliminate or at least reduce the possibility of a race condition arising from CI starting on the release commit before the associated tag is pushed. Such a state causes CI failures, because they take the latest version from the most recent `v*` tag.

I tested the flow on my local fork, but I can't test a real release unfortunately. I believe our existing infrastructure should "just work", but please let me know if that doesn't appear to be the case.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -121,7 +121,7 @@ As commits are cherry-picked when PRs are merged, creating the release should be
 
 ```bash
 git commit -a -m 'release: vXX'
-git tag 'vXX'
+git tag -a 'vXX' -m 'release: tag vXX'
 ```
 
 The package versions we are about to publish are derived from the git tag that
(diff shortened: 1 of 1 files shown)


kyliau:
I think this should target patch as well so that the workflow is similar when publishing `latest` and `next`

dgp1130:
@kyliau, updated to `master & patch`, though this is just a doc change, so it doesn't really make much of a difference in the end.

alan-agius4:
@dgp1130, can you please create another PR for patch? It doesn't apply cleanly

dgp1130:
Looks like this was merged to patch.

https://github.com/angular/angular-cli/commit/274673ab7207fc015ea695c5956ed45160483be0

alan-agius4:
@dgp1130, as per our offline conversation I fixed the conflict.

PR #18323, opened 2020-07-21
Fixes publish script to require NPM tag, use Wombat registry, and correctly read check flags

dgp1130:
This is a patch PR for https://github.com/angular/angular-cli/pull/18219. The `dist-tag` changes are not present here because that script does not exist and would not be used from this branch.

I also dropped the `doc:` commit since we don't typically look at docs on the non-master branch.

Only meaningful difference with the base PR is that I replaced the nullish coalescing operators with ugly ternaries to be compatible with the older typescript version.

### DIFF
--- a/lib/registries.ts
+++ b/lib/registries.ts
@@ -1,0 +1,10 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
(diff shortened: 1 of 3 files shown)


dgp1130:
This is stuck on 'Missing required status "ci/circleci: integration", <and more>'. Is there something more I need to do or is CI trying to run checks that don't exist on this branch?

alan-agius4:
it doesn’t exists in this branch, I wonder if we should do a dump task until Angular robot can handle this.

PR #18349, opened 2020-07-23
fix(@schematics/angular): show info message when depending on `rxjs-compat`

alan-agius4:


packages/schematics/angular/migrations/update-10/update-dependencies.ts
```diff
@@ -51,14 +51,21 @@ export default function (): Rule {
...
             'This package typically does not need to be installed manually. ' +
             'If it is not being used by project code, it can be removed from the package.json.',
         );
       }
     }
+
+    if (getPackageJsonDependency(host, 'rxjs-compat')) {
+      context.logger.info(
+        `Package "rxjs-compat" found in the workspace package.json. ` +
+        'This package typically was used during migration from RxJs version 5 to 6 during the Angular 5 timeframe.\n' +
```
clydin, line 66: maybe add `and may no longer be needed.` or similar?

alan-agius4: Updated

PR #18350, opened 2020-07-23
ci: use CircleCI pipeline SHA variables for commit validation

clydin:
This also speeds up the validation checks because the GitHub API request method is no longer used.


alan-agius4:
@googlebot ping.

PR #18354, opened 2020-07-24
build: minor fixes for google3 sync

kyliau:
This commit makes a few minor changes to enable syncing the CLI repository
into google3.

1. mark pkg_tar and pkg_npm as external
2. remove dependencies (marked as comment) which are stale in google3
3. add code fence for TS files generated from JSON schema in BUILD files
    since these files are compiled and added to the g3 codebase at sync time
4. Some minor typing changes
5. Remove duplicate licenses
6. mark dependencies which are not available in g3 as external
7. Immediately type the result of JSON.parse() as required by g3 linter.
   Otherwise, the type defaults to `any`.

### DIFF
--- a/packages/angular_devkit/build_ng_packagr/BUILD.bazel
+++ b/packages/angular_devkit/build_ng_packagr/BUILD.bazel
@@ -3,11 +3,14 @@
 # Use of this source code is governed by an MIT-style license that can be
 # found in the LICENSE file at https://angular.io/license
 
-load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
 load("@npm_bazel_jasmine//:index.bzl", "jasmine_node_test")
-load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
 load("//tools:defaults.bzl", "ts_library")
 load("//tools:ts_json_schema.bzl", "ts_json_schema")
+
+# @external_begin
+load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
+load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
+# @external_end
 
 licenses(["notice"])  # MIT
 
@@ -41,6 +44,7 @@ ts_library(
     ],
 )
 
+# @external_begin
 pkg_npm(
     name = "npm_package",
     deps = [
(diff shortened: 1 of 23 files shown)


packages/angular_devkit/build_ng_packagr/BUILD.bazel
```diff
@@ -56,6 +60,7 @@ pkg_tar(
     tags = ["manual"],
 )
 
+# @external_end
```
dgp1130, line 63: Nit: Shift this up one line so it looks associated with `pkg_npm()` rather than `ts_library()`.

packages/schematics/schematics/blank/factory.ts
```diff
@@ -102,7 +102,7 @@ export default function (options: Schema): Rule {
     try {
       const packageJsonContent = tree.read('/package.json');
       if (packageJsonContent) {
-        const packageJson = JSON.parse(packageJsonContent.toString('utf-8'));
+        const packageJson = JSON.parse(packageJsonContent.toString('utf-8')) as {};
```
dgp1130, line 105: This is for the linter? I'm surprised it runs at all on third party code.

We should add a comment here to explain that we need to cast the result for the linter. Also consider using `unknown` as a safer alternative to `{}` if not too much trouble.

packages/angular_devkit/build_ng_packagr/BUILD.bazel
kyliau: This is actually formatted by buildifier, which we run as part of CI. 
Please see https://github.com/angular/angular-cli/blob/master/package.json#L23

packages/schematics/schematics/blank/factory.ts
kyliau: Sorry, to be exact, it is the compiler, not the linter. go/tsetse is built into the compiler, so the errors show up as hard errors at compile time.
I added the comments and typed it as `never` instead of `{}`. I'm surprised that bracket notation is allowed on `never` type.

packages/angular_devkit/build_ng_packagr/BUILD.bazel
kyliau: ah, it looks like buildifier is still happy after I shifted the comment. All good now.

PR #18399, opened 2020-07-30
test: ensure schematics CLI E2E tests use local package registry

clydin:


alan-agius4:
From my understanding it's because previously the .npmrc file was not in the project directory.

PR #18442, opened 2020-08-04
Typo

daniel-liuzzi:


alan-agius4:
Hi @daniel-liuzzi, 

Thanks for this. However we are in the process of deleted the Angular CLI version 1 documentation as this is no longer supported.

PR #18467, opened 2020-08-06
Improve compiler options migration error reporting

clydin:
This change provides more fine-grained warnings during the `update-module-and-target-compiler-options` migration for V10.0 in the event a TypeScript configuration file could not be updated.  This can help prevent update issues where, for instance, a tsconfig path was misspelled prior to the update.  The `JSONFile` utility class was also augmented to directly throw when creating to ensure that the `content` property is always initialized.

### DIFF
--- a/packages/schematics/angular/utility/json-file.ts
+++ b/packages/schematics/angular/utility/json-file.ts
@@ -16,7 +16,6 @@ export type JSONPath = (string | number)[];
 /** @internal */
 export class JSONFile {
   content: string;
-  error: undefined | Error;
 
   constructor(
     private readonly host: Tree,
@@ -26,7 +25,7 @@ export class JSONFile {
     if (buffer) {
       this.content = buffer.toString();
     } else {
-      this.error = new Error(`Could not read ${path}.`);
+      throw new Error(`Could not read '${path}'.`);
     }
   }
 
--- a/packages/schematics/angular/migrations/update-10/add-deprecation-rule-tslint.ts
+++ b/packages/schematics/angular/migrations/update-10/add-deprecation-rule-tslint.ts
@@ -31,8 +31,10 @@ export default function (): Rule {
     }
 
     // Update tslint config.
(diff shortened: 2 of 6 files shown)


packages/schematics/angular/utility/json-file.ts
```diff
@@ -26,7 +25,7 @@ export class JSONFile {
     if (buffer) {
       this.content = buffer.toString();
     } else {
-      this.error = new Error(`Could not read ${path}.`);
+      throw new Error(`Could not read '${path}'.`);
```
alan-agius4, line 28: Much better. :)

alan-agius4:
Changed to target to `master only` as the `JSONFile` implementation doesn't exist in the patch branch.

ChaitanyaBabar:
Hi @filipesilva , @clydin  

Thanks for the above changes in [update-module-and-target-compiler-options.ts](https://github.com/angular/angular-cli/pull/18467/commits/0c59b907f42f1353c3690e9f5c63569399d0b926).

Actually my angular migration from 9 to 10 **halted** as due to **error** on in switch case for **workspace_models_1.Builders.NgPackagr:**  in file [update-module-and-target-compiler-options.ts](https://github.com/angular/angular-cli/pull/18467/commits/0c59b907f42f1353c3690e9f5c63569399d0b926).

Can you please let me know when would be the master branch containing these above changes done [here](https://github.com/angular/angular-cli/pull/18467/commits/0c59b907f42f1353c3690e9f5c63569399d0b926) 
would be published in npm registry, because as of now the **above changes are not** present in @angular/cli version 10.0.5 in npm registry.

Thanks,
Chaitanya

alan-agius4:
Hi @ChaitanyaBabar, this will available in 10.1.0, which will be released as stable some time around end of this month. 

You can already try this by running `ng update @angular/cli --next`.

ChaitanyaBabar:
@alan-agius4 

Thanks for the update.

PR #18488, opened 2020-08-08
Fix relative paths with bundleDependencies=false

amakhrov:
When `bundleDependencies` is turned off, webpack is configured to bundle all modules imported via relative path.
It works well for import paths like `'./some/file'` but fails for imports traversing the dir structure up, like `'../some/file'`.

In most cases it would still work well, because of the additional fallback (which checks whether node can resolve this module at all). However, in rare cases this fallback fails (see https://github.com/angular/universal/issues/1792)

Closes: #18493

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/server.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/server.ts
@@ -53,7 +53,7 @@ export function getServerConfig(wco: WebpackConfigOptions): Configuration {
       ...externalDependencies,
       (context: string, request: string, callback: (error?: null, result?: string) => void) => {
         // Absolute & Relative paths are not externals
-        if (request.startsWith('./') || isAbsolute(request)) {
+        if (request.startsWith('.') || isAbsolute(request)) {
           callback();
 
           return;


amakhrov:
@googlebot I signed it!

packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/server.ts
```diff
@@ -53,7 +53,7 @@ export function getServerConfig(wco: WebpackConfigOptions): Configuration {
       ...externalDependencies,
       (context: string, request: string, callback: (error?: null, result?: string) => void) => {
         // Absolute & Relative paths are not externals
-        if (request.startsWith('./') || isAbsolute(request)) {
+        if (request.startsWith('./') || request.startsWith('../') || isAbsolute(request)) {
```
alan-agius4, line 56: Can be simplified to `request.startsWith('.')`

amakhrov: I personally find the intention expressed clearer the way it is. But I'm ok with the change if it better matches the codebase.

amakhrov: On a second thought - wouldn't it make the condition too permissive?

alan-agius4: External modules cannot start with dot, since that’s an invalid name. Therefore, all paths that start with a dot must be relative.

alan-agius4: Also, in other classes we are using the same pattern.

amakhrov:
Thanks for the super quick feedback @alan-agius4  - quite impressive!

amakhrov:
@alan-agius4 I have a followup question though.
While the PR seems to fix the specific issue I was facing, I'm wondering if there are more cases that could fail like that.

As I mentioned, despite the incomplete original check for partial path, the builds tend to work well.
The reason is that if this check returns false, the code below has an extra check - tries to resolve the path via `require.resolve`. But this resolve call is performed in the context of `server.ts` file, not in the context of the compiled module.
Hence, in my case the `"../../../utils"` path was successfully resolved! Because it matches utils from angular_devkit itself
```
 /Users/amakhrov/Projects/angular-testing/node_modules/@angular-devkit/build-angular/src/utils/index.js
```

Do you think it's possible that this check can potentially give false positives for some arbitrary user module alias, if it also happens to match some internal module from devkit?

alan-agius4:
@amakhrov, glad you hear that you had a good experience contributing to the Angular CLI. 😁

I do think that there are some rare edge cases, where using `require.resolve(request)` might cause undesired behavior. A more correct way would be to changing the resolutions paths to use the workspace root path. `require.resolve(request, { paths: [wco.root] })`.

PR #18518, opened 2020-08-13
fix(@angular-devkit/build-angular): fix detecting commonjs deps in parent deps

ms-dosx86:
For now if we import a library (that exported as a CommonJS module) as an ES6 module, but has CommonJS imports, we get a warning for the library and also for its deps.

In my case I was using `camel-case`, but got two warnings for `camel-case` and `no-case` that is used by `camel-case` internally. Here is a minimal repro https://github.com/ms-dosx86/commonjs-warnings-repro

### DIFF
--- a/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/common-js-usage-warn-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/angular-cli-files/plugins/common-js-usage-warn-plugin.ts
@@ -28,6 +28,7 @@ interface WebpackModule extends compilation.Module {
 
 interface CommonJsRequireDependencyType {
   request: string;
+  module: WebpackModule | null;
 }
 
 export interface CommonJsUsageWarnPluginOptions {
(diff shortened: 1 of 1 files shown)


alan-agius4:
Hi @ms-dosx86, 

Thanks for this, while indeed the warning for `no-case`, shouldn't be displayed because it's a transitive dependency of a commonJS module. I think this change will omit the warnings for both. 

In your case, it's correct that a warning is displayed for `camel-case`, because `camel-case` is a commonJs library.

ms-dosx86:
@alan-agius4 Thanks for the quick response! Showing a warning for `camel-case` is totally fine, I was trying to get rid of warnings of transitive deps. You said my changes wouldn't show any warnings  and the test didn't passed. Looks like you're right. But there is a test case in `common-js-warning_spec.ts` with `bootstrap` import that checks for transitive deps warning. Altho it passes the test, it doesn't seem to be true when I add `import bootstrap` to my local project. I see two warnings about `bootstrap` and `jquery`. I was trying it on the same repro project that I had attached in the description, so you can reproduce it easily. I'm not familiar with bazel, so I can't say if the tests are right.

alan-agius4:
Hi @ms-dosx86, the problem seems to be that that particular test is testing applications in JIT mode, while in your case the application is in AOT mode. 

I'll take a look at this today, and add some more tests for both AOT and JIT mode and also address the issue reported in https://github.com/angular/angular-cli/issues/18526 

Thanks a lot of your investigation and the tentative fix, it's definitely very helpful.

PR #18522, opened 2020-08-13
test: improve reliability of E2E local npm registry on CI

clydin:


alan-agius4:
@clydin can you please create a separate PR for patch as it doesn't apply cleanly?

Thanks

PR #18530, opened 2020-08-14
Move ng-packagr builder to build-angular

alan-agius4:
With this change we deprecated `@angular-devkit/ng-packagr`, and move it's contents in `@angular-devkit/build-angular`


packages/angular_devkit/build_angular/src/ng-packagr/index.ts
```diff
@@ -0,0 +1,42 @@
...
+import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
+import { resolve } from 'path';
+import { Observable, from } from 'rxjs';
+import { mapTo, switchMap } from 'rxjs/operators';
+import { Schema as NgPackagrBuilderOptions } from './schema';
+
+async function initialize(
+  options: NgPackagrBuilderOptions,
+  root: string,
+  // tslint:disable-next-line:no-implicit-dependencies
```
clydin, line 17: With the peer dependency in place, these tslint comments shouldn't be needed

alan-agius4: You are correct, updated!

PR #18535, opened 2020-08-14
refactor(@angular/cli): replace experimental workspace API with stable API

clydin:
This change removes one of the last usages of the experimental workspace API.

### DIFF
--- a/packages/angular/cli/utilities/config.ts
+++ b/packages/angular/cli/utilities/config.ts
@@ -10,18 +10,20 @@ import {
   JsonAstObject,
   JsonObject,
   JsonParseMode,
-  experimental,
-  isJsonObject,
-  normalize,
+  json,
   parseJson,
   parseJsonAst,
-  virtualFs,
+  workspaces,
 } from '@angular-devkit/core';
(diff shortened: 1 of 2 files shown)


packages/angular/cli/utilities/config.ts
```diff
@@ -134,10 +168,13 @@ export function getWorkspaceRaw(
...
-  new NodeJsSyncHost().read(normalize(configPath))
-    .subscribe(data => content = virtualFs.fileBufferToString(data));
-
+  const data = readFileSync(configPath);
+  let start = 0;
+  if (data.length > 3 && data[0] === 0xef && data[1] === 0xbb && data[2] === 0xbf) {
+    // Remove BOM
+    start = 3;
+  }
+  const content = data.toString('utf-8', start);
```
alan-agius4, line 177: Maybe `data.toString('utf-8').replace(/^\uFEFF/, '')` to remove the BOM?

clydin: Less efficient especially for large files due to the need to allocate two strings plus the regex cost.

alan-agius4: Fair enough

Issue #18534, opened 2020-08-14
The build fails due to missing file after upgrade.


TrevorKarjanis:
# 🐞 Bug report

### Command

- [ ] new
- [x] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the previous version in which this bug was not present was 10.0.1.

### Description

After running [```ng update @angular/cli --next```](https://github.com/angular/angular-cli/pull/18478#issuecomment-673626215) in order to fix issue #18040 , the build fails with the error below. It appears the migration did not update a path in ```src/tsconfig.app.json```. After fixing that, there is another error with ```tsconfig.json``` which extends a non-existent configuration, and the compiler options from base are missing. After fixing that, the build succeeds. I reproduced with a new application generated with version 10.0.1.

## 🔬 Minimal Reproduction
```bash
> ng new my-app
> ng build
> ng update @angular/cli --next
> ng build
```

## 🔥 Exception or Error

```bash
> my-app@4.0.0 build /Users/trevor/my-app
> ng build

An unhandled exception occurred: error TS5083: Cannot read file '/Users/trevor/MyApp/tsconfig.base.json'.

See "/private/var/folders/lt/77mxdc9s0bl0ff37cm33cjzh0000gn/T/ng-bqLn9R/angular-errors.log" for further details.
npm ERR! code ELIFECYCLE
npm ERR! syscall spawn
npm ERR! file sh
npm ERR! errno ENOENT
npm ERR! insight-cm@4.0.0 build: `ng build`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the insight-cm@4.0.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/trevor/.npm/_logs/2020-08-14T19_44_27_149Z-debug.log
```

```bash
> my-app@4.0.0 build /Users/trevor/my-app
> ng build

An unhandled exception occurred: error TS5083: Cannot read file '/Users/trevor/MyApp/tsconfig.json'.

See "/private/var/folders/lt/77mxdc9s0bl0ff37cm33cjzh0000gn/T/ng-K3RBoq/angular-errors.log" for further details.
[...]

PR #18540, opened 2020-08-15
fix(@schematics/angular): change `extends`  from `tsconfig.base.json` to `tsconfig.json`

alan-agius4:
Closes: #18534


Splaktar:
Based on what I observed in https://github.com/angular/angular-cli/issues/18534#issuecomment-674504479, I'm not sure that this is sufficient.

I still had comments in my `tsconfig.json` about it being a "Solution Style" tsconfig.json file and the contents of my `tsconfig.base.json` got deleted but not moved into my `tsconfig.json`.

alan-agius4:
@Splaktar likely it's because we where using the wrong JsonFile instance here: https://github.com/angular/angular-cli/pull/18540/files#diff-d1380d3bc19bb56c82d31f1222da33ecL63

PR #18541, opened 2020-08-15
refactor(@angular-devkit/schematic): replace any with unknown

santoshyadavdev:


packages/angular_devkit/schematics/src/tree/action.ts
```diff
@@ -157,13 +157,15 @@ export function isContentAction(action: Action): action is CreateFileAction | Ov
 }
 
 
-export function isAction(action: any): action is Action {  // tslint:disable-line:no-any
-  const kind = action && action.kind;
+export function isAction(action: unknown): action is Action {
```
alan-agius4, line 160: This is a breaking changing, because `unknown` is more restrictive than `any`.

JoostK: I don't think this is breaking. `unknown` is a top-type like `any`, such that anything is assignable to it. It's only within this fn's implementation that more restrictions apply when using `action` (although that's immediately undone with the speculative cast to `Action`)

alan-agius4: You are absolutely right, probably in this case it’s best to remove the cast and check add a check using typeof

`typeof action === ‘object’ && action.kind`

santoshyadavdev: Hi @alan-agius4, let me know if any changes are needed.

packages/angular_devkit/schematics/src/tree/action.ts
```diff
@@ -157,13 +157,16 @@ export function isContentAction(action: Action): action is CreateFileAction | Ov
...
+  const act = action as Action;
 
-  return action !== null
-      && typeof action.id == 'number'
-      && typeof action.path == 'string'
-      && (kind == 'c' || kind == 'o' || kind == 'r' || kind == 'd');
+  const kind = act && act.kind;
+
+  return typeof action === 'object'
+    && typeof act.hasOwnProperty('kind')
```
alan-agius4, line 166:
```suggestion
```

packages/angular_devkit/schematics/src/tree/action.ts
alan-agius4: Nit: can you use === instead of ==?

packages/angular_devkit/schematics/src/tree/action.ts
```diff
@@ -157,13 +157,16 @@ export function isContentAction(action: Action): action is CreateFileAction | Ov
...
+export function isAction(action: unknown): action is Action {
+  const act = action as Action;
 
-  return action !== null
-      && typeof action.id == 'number'
-      && typeof action.path == 'string'
-      && (kind == 'c' || kind == 'o' || kind == 'r' || kind == 'd');
+  const kind = act && act.kind;
+
+  return typeof action === 'object'
```
alan-agius4, line 165: typeof null is also an object, I think it’s better to add a check for null.

packages/angular_devkit/schematics/src/tree/action.ts
```diff
@@ -157,13 +157,16 @@ export function isContentAction(action: Action): action is CreateFileAction | Ov
 }
 
 
-export function isAction(action: any): action is Action {  // tslint:disable-line:no-any
-  const kind = action && action.kind;
+export function isAction(action: unknown): action is Action {
+  const act = action as Action;
```
alan-agius4, line 161: In all fairness I am not a big fan of this casting here, as this is hiding potential errors.

alan-agius4: Something like seems to be more "correct", but wondering if we should just stop at `hasOwnProperty` since it's the object has those 3 props it's should be an action.

```ts
export function isAction(action: unknown): action is Action {
  if (
    action
    && typeof action === 'object'
    && action.hasOwnProperty('path')
    && action.hasOwnProperty('id')
    && action.hasOwnProperty('kind')
  ) {
    const { kind, id, path } = action as { id: unknown, path: unknown, kind: unknown };
    if (
      typeof kind === 'string'
      && ['c', 'o', 'r', 'd'].includes(kind)
      && typeof id === 'number'
      && typeof path === 'string'
    ) {
      return true;
    }
  }

  return false;
}
```

//cc @clydin what do you think?

clydin: based on the usage of the function, I think we should consider deprecating it and leaving it as-is for now.  I put together a PR that removes the only use internally: https://github.com/angular/angular-cli/pull/18637

alan-agius4: SGTM

alan-agius4: @santoshyadavdev, mind opening a separate PR to deprecate the this method?

santoshyadavdev: Ahh cool @alan-agius4 will do that today.

santoshyadavdev: @alan-agius4 please see https://github.com/angular/angular-cli/pull/18727, should I close this one.

alan-agius4: Thanks, I’ll close it for you.

PR #18542, opened 2020-08-15
refactor(@angular-devkit/core): remove any types

santoshyadavdev:


packages/angular_devkit/core/src/json/schema/interface.ts
```diff
@@ -69,8 +69,7 @@ export interface SchemaValidator {
 
 export interface SchemaFormatter {
   readonly async: boolean;
-  // tslint:disable-next-line:no-any
-  validate(data: any): boolean | Observable<boolean>;
+  validate(data: unknown): boolean | Observable<boolean>;
```
clydin, line 72: This is a breaking change since the interface is public and `unknown` is more restrictive than `any`.

santoshyadavdev: Hey @clydin sorry cannot understand what I need to change? can you explain, please

clydin: Can you revert this one change? The any usage will unfortunately need to stay in this case since it may break users of the public API.  This change could potentially be made in the future but it would need to be in a major release.

etc/api/angular_devkit/core/src/_golden-api.d.ts
```diff
@@ -889,7 +889,7 @@ export interface SchemaFormat {
 
 export interface SchemaFormatter {
     readonly async: boolean;
-    validate(data: any): boolean | Observable<boolean>;
+    validate(data: unknown): boolean | Observable<boolean>;
```
santoshyadavdev, line 892: Hi @clydin what about this change, do I need to rollback this too?

PR #18555, opened 2020-08-17
refactor(@ngtools/webpack): use ES modules imports instead of `require`

alan-agius4:
With this change we update the replace_resources transformer for styles and templates to use ES6 module import syntax instead of the CommonJs require syntax.

### DIFF
--- a/packages/ngtools/webpack/src/transformers/replace_resources.ts
+++ b/packages/ngtools/webpack/src/transformers/replace_resources.ts
@@ -7,18 +7,6 @@
  */
 import * as ts from 'typescript';
 
-// emit helper for `import Name from "foo"`
-// importName is marked as an internal property but is needed for the tslib import.
-const importDefaultHelper: ts.UnscopedEmitHelper & { importName?: string } = {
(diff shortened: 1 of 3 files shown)


packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -50,20 +39,29 @@ export function replaceResources(
...
+        return sourceFile;
+      }
+
+      const updatedSourceFile = ts.visitNode(sourceFile, visitNode);
+      if (resourceImportDeclarations.length) {
+        return ts.updateSourceFileNode(updatedSourceFile, [
+          // Add resource imports
+          ...resourceImportDeclarations,
+          ...updatedSourceFile.statements,
+        ]);
```
clydin, line 52:
```suggestion
        return ts.updateSourceFileNode(
          updatedSourceFile,
          ts.setTextRange(
            ts.createNodeArray([resourceImportDeclarations, ...updatedSourceFile.statements]),
            sourceFile.statements,
          ),
        );
```

PR #18566, opened 2020-08-18
Update Bazel and rules_nodejs

alan-agius4:


.circleci/bazel.rc
```diff
@@ -12,9 +12,10 @@ test --test_tag_filters=-manual
 
 # Workaround https://github.com/bazelbuild/bazel/issues/3645
 # Bazel doesn't calculate the memory ceiling correctly when running under Docker.
-# Limit Bazel to consuming resources that fit in CircleCI "medium" class which is the default:
+# Limit Bazel to consuming resources that fit in CircleCI "xlarge" class
 # https://circleci.com/docs/2.0/configuration-reference/#resource_class
-build --local_resources=3072,2.0,1.0
+build --local_cpu_resources=8
```
alan-agius4, line 17: NB: We should fine tune the resource cases needed, maybe we can switch to Medium+/Large instead of XL. Tried to use medium, but it failed with `java.io.IOException: Stream closed`.  Thus for now, I just updated the resources to match the class being used.

PR #18597, opened 2020-08-22
fix(@schematics/schematics): run install step with sample schematics

### LINKED ISSUE #17425: Run npm i with the Sample Schematics Command
Running:

```
schematics schematic --name demo
```

Creates the demo, but we still need to run `npm i`manually on the project to get the deps.

Ayudh:
Closes #17425

### DIFF
--- a/packages/schematics/schematics/schematic/factory.ts
+++ b/packages/schematics/schematics/schematic/factory.ts
@@ -8,6 +8,8 @@
 import { Path, strings } from '@angular-devkit/core';
 import {
   Rule,
+  SchematicContext,
+  Tree,
   apply,
   mergeWith,
   move,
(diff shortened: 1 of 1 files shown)


packages/schematics/schematics/schematic/factory.ts
```diff
@@ -8,31 +8,36 @@
...
   Rule,
+  SchematicContext,
+  Tree,
   apply,
   mergeWith,
   move,
-  partitionApplyMerge,
-  template,
-  url,
+  partitionApplyMerge, template, url,
```
alan-agius4, line 16: Nit: can we keep these in a single line please? Thanks

alan-agius4 (approved):
LGTM just on small thing

packages/schematics/schematics/schematic/factory.ts
Ayudh: Yes. Updated.

PR #18610, opened 2020-08-25
refactor(@angular/cli): improve update package discovery

andreialecu:
Removes the dependency on [read-package-tree](https://www.npmjs.com/package/read-package-tree) which is closely tied to `npm`.


andreialecu:
The tests seem not to be able to find `rxjs-compat`:

```
An unhandled exception occurred: Cannot find module 'rxjs-compat'
Require stack:
- /mnt/ramdisk/assets/1.0-project-1598367923182/node_modules/rxjs/Rx.js
```

I'm not sure if this error is related to the changes in the PR. Any ideas would be appreciated.

andreialecu:
I was finally able to understand why the error occurs.

The problem is with the use of `require.resolve` to find the location of dependencies, populating this NodeJS internal cache:
https://github.com/nodejs/node/blob/dcba12895ad58275ba5b027c2c5110461dfebf66/lib/internal/modules/cjs/loader.js#L239

`rxjs@5.5.12` had the `main` entry in package.json set to `./Rx.js` (https://unpkg.com/rxjs@5.5.12/package.json)

`rxjs@6.6.2` has it set to `./index.js` (https://unpkg.com/rxjs@6.6.2/package.json)

Here's a step by step explanation:

1) The code gathers `package.json` dependencies, then visits each dependency's `package.json` to check which version is actually installed (NodeJS `packageJsonCache` gets populated, see link above). At this point `rxjs` is at version `5.5.12`
2) `yarn/npm install` is ran and `rxjs` is updated to version `6.6.2`.
3) `executeMigrations` tries to run the schematics in the context of the same process that ran step 1. `rxjs` will resolve to `rxjs/Rx.js` instead of `rxjs/index.js`, and thus the process fails

The problem is also described here: https://stackoverflow.com/questions/59865584/how-to-invalidate-cached-require-resolve-results

andreialecu:
I also included an E2E test for a `yarn workspaces` monorepo and it seems to pass (besides a warning in one of the schematics).

andreialecu:
I also included an E2E test for a `yarn workspaces` monorepo and it seems to pass (besides a warning in one of the migrations).

andreialecu:
There are still some potential problems and edge cases, which cannot be fixed just in this package.

Particularly in this migration script:
https://github.com/angular/angular/blob/master/packages/core/schematics/migrations/missing-injectable/transform.ts#L9-L12

The script above depends on `@angular/angular-cli` and `typescript` and neither of them are dependencies of `@angular/core`: https://unpkg.com/@angular/core@10.0.0/package.json

This causes problems if `yarn` installs the packages in different directories. For example I have a project where `@angular/cli` and most deps are hoisted to the top level of the monorepo, but `@angular/angular-cli` is nested inside one particular workspace. If the dependency was defined properly, `yarn` would have a better hint of where to place dependencies within the file system and this error wouldn't exist.

Here's the error it can produce:

```
An unhandled exception occurred: Cannot find module '@angular/compiler-cli/src/ngtsc/imports'
Require stack:
- /home/aandrei/projects/monorepo/node_modules/@angular/core/schematics/migrations/missing-injectable/transform.js
- /home/aandrei/projects/monorepo/node_modules/@angular/core/schematics/migrations/missing-injectable/index.js
- /home/aandrei/projects/monorepo/node_modules/@angular-devkit/schematics/tools/export-ref.js
- /home/aandrei/projects/monorepo/node_modules/@angular-devkit/schematics/tools/index.js
- /home/aandrei/projects/monorepo/node_modules/@angular/cli/utilities/json-schema.js
- /home/aandrei/projects/monorepo/node_modules/@angular/cli/models/command-runner.js
- /home/aandrei/projects/monorepo/node_modules/@angular/cli/lib/cli/index.js
- /home/aandrei/projects/monorepo/node_modules/@angular/cli/lib/init.js
- /home/aandrei/projects/monorepo/node_modules/@angular/cli/bin/ng
```

However:

```json
  "workspaces": {
    "packages": [
      "packages/*"
    ],
    "nohoist": [
      "**/@angular*",
      "**/@angular*/**",
      ...
    ]
  },
```

I could work around it by adding the above to my root `package.json` thus disabling hoisting on `@angular` related packages so they all sit together. `ng update` then runs properly. The `nohoist` entries can then be removed after running `ng update`, since `@angular` itself works properly.

The correct fix here would be for `@angular/core` to add `@angular/compiler-cli` and `typescript` to `peerDependencies` or `peerDependenciesMeta`, which would also move https://github.com/angular/angular-cli/issues/16980 further along. Unfortunately, because updates work one version at a time, I'm not sure how to fix this without going back and fixing it in various previous versions of `@angular/core`.

But even so, the changes in this PR should move this further along.

clydin:
Thank you for the extensive work on improving the update command.
Based on the issues still present after this PR in regards to full workspace support, can you adjust the change to be a refactor that improves update package discovery instead of a feature. The remaining update issues will need to be addressed as well as an analysis of any potential CLI usage and runtime concerns including documentation before yarn workspace support can be considered feature complete. However, these changes definitely provide an improvement.

andreialecu:
@clydin I have updated the PR as per your requirement.

I'd like to also point out something that I think may be a fundamental problem in how `ng update` currently works and should probably be addressed at some point:

* First of all, to illustrate, run this: https://runkit.com/andreialecu/5f4a0a45b7d56e001a469c1e
  If you comment out line 18 (`require.resolve`) the output is entirely different. Make sure you're running on Node 12. On Node 8 (the default) there's different behavior.

This migration script imports `rxjs`: https://github.com/angular/angular/blob/master/packages/core/schematics/migrations/static-queries/index.ts

As per my previous comment, `rxjs` has moved its entry point between version 5 and 6. The project being upgraded was on version 5 initially, with the `./Rx.js` entry point. Version 6 has a main of `./index.js` but also has its own `./Rx.js` file that just imports `rxjs-compat`: https://unpkg.com/rxjs@6.6.2/Rx.js. (Note that you can get the error back by replacing `resolve.sync` with `require.resolve` `packages/angular/cli/utilities/package-tree.ts`)

Now when a multi-version upgrade runs, the following can happen:
1) if a schematic for a previous version depends on a package such as `rxjs` and requires it, it will lock down its entry point (and potentially even its entire code via `require.cache`)
2) `npm/yarn install` may update `rxjs` to a future version, with different behavior
3) a different schematic may also require `rxjs` but even though `node_modules` was updated, it is not guaranteed to be able to load the latest version.

I think the only right solution would be to always start a fresh process after `npm/yarn install` and run further schematics and migrations in this fresh process. If further clarification is required, let me know.

andreialecu:
> The correct fix here would be for @angular/core to add @angular/compiler-cli and typescript to peerDependencies or peerDependenciesMeta, which would also move #16980 further along.

Further reading here: https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

clydin:
>  The correct fix here would be for @angular/core to add @angular/compiler-cli and typescript to peerDependencies or peerDependenciesMeta, which would also move #16980 further along.

This is intentional.  The framework libraries are web-only packages that cannot have additional dependencies.  Improvements to schematics and migrations are in the planning/design stages that will help to alleviate these type of issues.

merceyz:
> framework libraries are web-only packages that cannot have additional dependencies

The thing is they do have additional dependencies they just don't declare it

clydin:
That is correct but that part is the intentional aspect since they are only needed for one specific use case during development and not during runtime or outside the migration scenario of the CLI.  One of the primary goals of the changes mentioned to schematics/migrations is to remove the situation of the _hidden_ dependencies.

packages/angular/cli/lib/init.ts
```diff
@@ -20,7 +21,7 @@ const packageJson = require('../package.json');
 
 function _fromPackageJson(cwd = process.cwd()): SemVer | null {
   do {
-    const packageJsonPath = path.join(cwd, 'node_modules/@angular/cli/package.json');
+    const packageJsonPath = resolve.sync('@angular/cli/package.json', { paths: [process.cwd()] });
```
clydin, line 24: Can you remove this change? While this file could use some improvements, it is out of scope for the update related changes in this PR.

tests/legacy-cli/e2e/utils/assets.ts
```diff
@@ -4,7 +4,8 @@ import {getGlobalVariable} from './env';
 import {relative} from 'path';
 import {copyFile, writeFile} from './fs';
 import {useBuiltPackages} from './project';
-import { git, silentNpm } from './process';
+import { git, silentNpm, silentYarn } from './process';
+import { skip } from 'rxjs/operators';
```
clydin, line 8: `skip` doesn't appear to be used.

packages/angular/cli/lib/init.ts
andreialecu: @clydin so there is a bug here I introduced, thanks for catching it, it should be `paths: [cwd]` instead. 

However, this change **can be** related to this PR. Consider the following scenario:

```
/app/package.json
/app/node_modules/@angular/cli/... <- @angular/cli could be hoisted here by yarn

/app/packages/webapp/package.json <- subproject here
/app/packages/webapp/node_modules/... <- other deps, but @angular/cli not between them
```

Now, if a user runs `ng update` while being `cwd` to `/app/packages/webapp/` the logic above will fail to find the locally installed `@angular/cli` and will use the global one. This can lead to unexpected behavior.

Are you sure you want me to revert it?

tests/legacy-cli/e2e/utils/assets.ts
andreialecu: Got it. Will force push a fix for this once a decision is made on the other comment.

packages/angular/cli/lib/init.ts
clydin: Please do.  That function will not exist soon.  The problem demonstrated above will be removed as a result as well.

andreialecu:
I'm not sure what happened with the tests. They were previously running fine. Maybe it's a flake?

alan-agius4:
Master is currently red, we’re working on it 😊
PS: you have a conflict

andreialecu:
@alan-agius4 apologies, somehow I messed up to revert that file properly, hence the conflict. Should be good now I hope.

clydin:
A rebase against master should fix CI.
Otherwise I think this looks good.

clydin (approved):

PR #18635, opened 2020-08-27
feat(@schematics/angular): add migration to replace deprecated library builder

alan-agius4:
In 10.1, `@angular-devkit/build-ng-packagr` was deprecated in favor of `@angular-devkit/build-angular`. This migration updates the library build target to use the non deprecated builder.

### DIFF
--- a/packages/schematics/angular/migrations/update-11/replace-ng-packagr-builder.ts
+++ b/packages/schematics/angular/migrations/update-11/replace-ng-packagr-builder.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Rule, chain } from '@angular-devkit/schematics';
+import { NodeDependencyType, addPackageJsonDependency, removePackageJsonDependency } from '../../utility/dependencies';
+import { latestVersions } from '../../utility/latest-versions';
+import { updateWorkspace } from '../../utility/workspace';
+import { Builders } from '../../utility/workspace-models';
+
+export default function (): Rule {
+  return chain([
+    updateWorkspace(workspace => {
+      for (const [, project] of workspace.projects) {
+        for (const [, target] of project.targets) {
+          if (target.builder === Builders.DeprecatedNgPackagr) {
+            target.builder = Builders.NgPackagr;
+          }
+        }
+      }
+    }),
+    host => {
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/migrations/update-11/replace-ng-packagr-builder.ts
```diff
@@ -0,0 +1,40 @@
...
+import { NodeDependencyType, addPackageJsonDependency, removePackageJsonDependency } from '../../utility/dependencies';
+import { latestVersions } from '../../utility/latest-versions';
+import { updateWorkspace } from '../../utility/workspace';
+import { Builders, ProjectType } from '../../utility/workspace-models';
+
+export default function (): Rule {
+  return chain([
+    updateWorkspace(workspace => {
+      for (const [, project] of workspace.projects) {
+        if (project.extensions.projectType !== ProjectType.Library) {
```
clydin, line 19: projectType is technically optional.  We may want to skip this check.

packages/schematics/angular/migrations/update-11/replace-ng-packagr-builder.ts
```diff
@@ -0,0 +1,40 @@
...
+
+        for (const [, target] of project.targets) {
+          if (target.builder === Builders.DeprecatedNgPackagr) {
+            target.builder = Builders.NgPackagr;
+          }
+        }
+      }
+    }),
+    host => {
+      removePackageJsonDependency(host, '@angular-devkit/build-ng-packagr');
```
clydin, line 31: Not for this PR but we should consider turning these dependency helpers into actual rules so the anonymous function isn't needed.

packages/schematics/angular/migrations/update-11/replace-ng-packagr-builder_spec.ts
```diff
@@ -0,0 +1,72 @@
...
+          } as any,
+        },
+      },
+    },
+  };
+
+  tree.create('/angular.json', JSON.stringify(angularConfig, undefined, 2));
+}
+
+describe(`Migration to add tslib as a direct dependency in library projects`, () => {
```
clydin, line 35: copy/paste error on the string?

alan-agius4: Yes

packages/schematics/angular/migrations/update-11/replace-ng-packagr-builder.ts
alan-agius4: yeah I think that's a good idea.

PR #18641, opened 2020-08-27
feat(@angular-devkit/build-angular): ask to use a new port if in use

### LINKED ISSUE #18638: Find available port when running ng serve instead of asking the user to specify it manually
# 🚀 Feature request

- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description
When running `ng serve` for a second project it should use the next available port instead of throwing an error. Eg: 4201

If I run `ng serve` and port 4200 is being used I expect the command to be smart enough to try the next available port until successful. Eg: 4201, 4202, etc

This feature is available in all these: Stencil, React and Vue.

clydin:
The dev-server builder will now ask the user if an available port can be used when a specified port is already in use.  If the user answers in the affirmative, an available port will be used instead.  Otherwise, the serve operation will fail.  If the terminal is not interactive (for example, CI), the serve operation will fail directly in the event of an already in use port.

Closes #18638


alan-agius4 (approved):
Good work!

PR #18647, opened 2020-08-28
docs: update `fileReplacements` description

alan-agius4:


packages/angular/cli/lib/config/schema.json
```diff
@@ -725,7 +725,7 @@
               ]
             },
             "fileReplacements": {
-              "description": "Replace files with other files in the build.",
+              "description": "Replace a compilation source files with other compilation source files in the build.",
```
clydin, line 728:
```suggestion
              "description": "Replace compilation source files with other compilation source files in the build.",
```

PR #18651, opened 2020-08-28
Initial Angular schematics cleanup

clydin:
This PR starts updating older schematics to use newer utilities and helper functions with the goal to reduce complexity as well as remove older helper code.

### DIFF
--- a/packages/schematics/angular/migrations/update-7/typescript-helpers.ts
+++ b/packages/schematics/angular/migrations/update-7/typescript-helpers.ts
@@ -5,18 +5,14 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { JsonParseMode, parseJsonAst } from '@angular-devkit/core';
-import { Rule, Tree, chain } from '@angular-devkit/schematics';
+import { Rule, chain } from '@angular-devkit/schematics';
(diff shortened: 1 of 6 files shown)


packages/schematics/angular/migrations/update-7/typescript-helpers.ts
```diff
@@ -42,38 +38,27 @@ export function typeScriptHelpersRule(): Rule {
...
+    try {
+      tsConfigJson = new JSONFile(host, tsConfigPath);
+    } catch {
+      return;
     }
 
-    const importHelpers = findPropertyInAstObject(compilerOptions, 'importHelpers');
-    if (importHelpers && importHelpers.value === true) {
-      return host;
+    const compilerOptions = tsConfigJson.get(['compilerOptions']);
```
alan-agius4, line 51: This check here should be redundant with the new helpers.

alan-agius4 (approved):

alan-agius4:
Nice cleanup!

packages/schematics/angular/migrations/update-7/typescript-helpers.ts
clydin: The original skipped the addition if `compilerOptions` was not present.  There is no test for that case so I was unsure of whether it was intentional but I left the behavior for consistency.

PR #18657, opened 2020-08-29
fix(@angular-devkit/core): allow property remove with workspace API

clydin:
Removal changes were previously being improperly recorded with the wrong parent.  A property can now be directly removed by setting the value to undefined or by using the delete operator.


alan-agius4 (approved):
LGTM

PR #18661, opened 2020-08-30
refactor(@schematics/angular): change the order of node types in package.json template

bampakoa:
Re-arrange the node types npm package so that `devDependencies` are sorted in  alphabetical order

### DIFF
--- a/packages/schematics/angular/workspace/files/package.json.template
+++ b/packages/schematics/angular/workspace/files/package.json.template
@@ -24,10 +24,10 @@
   },
   "devDependencies": {
     "@angular/cli": "<%= '~' + version %>",
-    "@angular/compiler-cli": "<%= latestVersions.Angular %>",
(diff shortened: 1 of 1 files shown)


alan-agius4 (changes requested):
Thanks for this, however this changes the semantics because now `@types/jasmine` and `@types/jasminewd2` are included when `minimal` is true.

Also commit message scope is not correct as it should be `@schematics/angular`

bampakoa:
Thanks for the feedback @alan-agius4 . I have made some adjustments to the committed code along with the change in the message scope. I tried also to run the test suite with `yarn test --full` but yarn complained that it did not find the test command. Any thoughts on that?

alan-agius4:
Thanks for this. Yeah to run the unit tests of `@schematics/angular` you need to run `yarn bazel test //packages/schematics/angular:angular_test `

alan-agius4 (approved):

PR #18699, opened 2020-09-04
fix(@angular-devkit/core): fix typo for JobOutboundMessageKind.Pong

kyliau:
Marking this commit as `fix` instead of `refactor` as it's technically a
breaking change.

### DIFF
--- a/packages/angular_devkit/core/src/experimental/jobs/api.ts
+++ b/packages/angular_devkit/core/src/experimental/jobs/api.ts
@@ -131,7 +131,7 @@ export enum JobOutboundMessageKind {
   OnReady = 'c',
   Start = 's',
   End = 'e',
-  Pong = 'p',
+  Ping = 'p',
 
   // Feedback messages.
   Output = 'o',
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/core/src/experimental/jobs/api.ts
```diff
@@ -131,7 +131,7 @@ export enum JobOutboundMessageKind {
   OnReady = 'c',
   Start = 's',
   End = 'e',
-  Pong = 'p',
+  Ping = 'p',
```
alan-agius4, line 134: 😅

alan-agius4 (approved):

packages/angular_devkit/core/src/experimental/jobs/api.ts
clydin: I’m not sure that’s a typo

alan-agius4: I see, ping <-> pong

clydin: For a network ping, ICMP uses echo request (type 8) and echo reply (type 0) for the messages.

kyliau:
Closing this since "pong" here seems to imply an echo reply

PR #18723, opened 2020-09-08
 build: update postcss-loader to version 4.0

alan-agius4:


alan-agius4:
Blocked on https://github.com/webpack-contrib/postcss-loader/pull/475

clydin (approved):

PR #18727, opened 2020-09-08
refactor(@angular-devkit/schematics): deprecate isAction

santoshyadavdev:


etc/api/angular_devkit/schematics/src/_golden-api.d.ts
```diff
@@ -339,6 +339,9 @@ export declare class InvalidUpdateRecordException extends BaseException {
     constructor();
 }
 
+/**
+ * @deprecated
```
alan-agius4, line 343: Can you please add the version when it was deprecated please? (V11)

See other deprecated methods for example 😊

alan-agius4:
@santoshyadavdev, thanks for this. 

Small NIT, also the commit message is valid.

etc/api/angular_devkit/schematics/src/_golden-api.d.ts
santoshyadavdev: Let me know if looks fine now.

alan-agius4 (approved):

PR #18729, opened 2020-09-08
build: remove experimental build event protocol

kyliau:
Build event protocol was implemented as part of the Bazel experiment, but it's no longer worked on / needed by anyone so remove it for now.


alan-agius4:
Wow that was fast 😊

alan-agius4:
I think you can now remove the dependency on `@bazel/labs`.

Should we also remove BEP from other parts of the CLI? @clydin thoughts?

packages/angular_devkit/core/src/analytics/index.ts
```diff
@@ -25,7 +25,7 @@ export enum NgCliAnalyticsDimensions {
   RamInGigabytes = 3,
   NodeVersion = 4,
   NgAddCollection = 6,
-  NgBuildBuildEventLog = 7,
+  NgBuildBuildEventLog = 7, // deprecated
```
kyliau, line 28: I marked this enum as deprecated instead of deleting it because deleting would mess up the dimension IDs in https://github.com/angular/angular-cli/blob/master/docs/design/analytics.md

clydin (approved):

packages/angular_devkit/core/src/analytics/index.ts
alan-agius4, line 28: Nit
```suggestion
  /** @deprecated*/
  NgBuildBuildEventLog = 7,
```

alan-agius4 (approved):

alan-agius4 (approved):

PR #18743, opened 2020-09-09
 refactor(@angular-devkit/core): remove deprecated terminal utils

alan-agius4:
BREAKING CHANGE:

Deprecated `@angular-devkit/core` terminal API has been removed. Use 3rd party packages like chalk, colors or ansi-colors instead.

**Note:** this changes doesn't effect application developers


packages/angular_devkit/core/node/cli-logger.ts
```diff
@@ -6,7 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import { filter } from 'rxjs/operators';
-import { logging, terminal } from '../src';
+import { logging } from '../src';
+import { colors as terminalColors } from './color';
```
alan-agius4, line 10: I am actually inclined to remove colors from here and rely on consumers to pass the colors via the colors parameter when colors are desired.

clydin: Sounds good.

packages/angular/cli/utilities/color.ts
```diff
@@ -8,18 +8,16 @@
...
 
+type AnsiColors = typeof ansiColors;
+
 // Typings do not contain the function call (added in Node.js v9.9.0)
-export const supportsColor =
+const supportsColor =
   process.stdout instanceof WriteStream &&
   ((process.stdout as unknown) as { getColorDepth(): number }).getColorDepth() > 1;
 
-export function removeColor(text: string): string {
```
clydin, line 16: This was added due to the `unstyle` method from the package not working when colors were disabled.
https://github.com/doowb/ansi-colors/blob/a4794363369d7b4d1872d248fc43a12761640d8e/index.js#L38

alan-agius4:
@clydin PTAL

clydin (approved):

PR #18749, opened 2020-09-10
fix(@angular-devkit/build-angular): set logging level to verbose when using `--vebose`

alan-agius4:


clydin (approved):
nit: `--vebose` --> `--verbose` in commit message

PR #18767, opened 2020-09-11
refactor: delete unused AIO benchmark

dgp1130:
This benchmark doesn't appear to be used and should be safe to remove (it still references a commit from Angular v7).


alan-agius4 (approved):

nutza249949:
master

PR #18779, opened 2020-09-14
feat(@angular-devkit/build-angular): improve build stats output format

alan-agius4:
With this change we also remove sourcemaps from build info to align with Webpack 5 output.

**Before**
image

image

**After**
image


alan-agius4:
@clydin PTAL

clydin (approved):

PR #18783, opened 2020-09-14
Improve async handling in generated code

apeeters:
This PR fixes two tslint errors in generated projects:

- Promise returning method should be async
- Avoid floating promises

### DIFF
--- a/integration/angular_cli/e2e/src/app.e2e-spec.ts
+++ b/integration/angular_cli/e2e/src/app.e2e-spec.ts
@@ -8,9 +8,9 @@ describe('workspace-project App', () => {
     page = new AppPage();
   });
 
-  it('should display welcome message', () => {
-    page.navigateTo();
-    expect(page.getTitleText()).toEqual('integration-project app is running!');
+  it('should display welcome message', async () => {
+    await page.navigateTo();
+    await expect(await page.getTitleText()).toEqual('integration-project app is running!');
   });
 
   afterEach(async () => {
--- a/integration/angular_cli/e2e/src/app.po.ts
+++ b/integration/angular_cli/e2e/src/app.po.ts
@@ -1,11 +1,11 @@
 import { browser, by, element } from 'protractor';
 
 export class AppPage {
-  navigateTo(): Promise<unknown> {
+  async navigateTo(): Promise<unknown> {
     return browser.get(browser.baseUrl) as Promise<unknown>;
   }
 
-  getTitleText(): Promise<string> {
+  async getTitleText(): Promise<string> {
     return element(by.css('app-root .content span')).getText() as Promise<string>;
   }
 }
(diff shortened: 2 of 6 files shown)


alan-agius4:
The floating promises are intended. See: https://github.com/angular/angular-cli/pull/17652#discussion_r420878945

apeeters:
> The “floating” promises are intended. See: [#17652 (comment)](https://github.com/angular/angular-cli/pull/17652#discussion_r420878945)

Unfortunately the PR mentioned in that comment (https://github.com/angular/angular-cli/pull/15472) is closed. I can't find any documentation or plan on how this will be fixed/resolved.

alan-agius4 (changes requested):
We had a chat around this yesterday during our team meeting and doing this change shouldn't break Protractor. Therefore this should be good to go after
- Update the commit message scopes to `@schematics/angular`
- Dropping Jasmine bump commit since this seems redundant.

integration/angular_cli/e2e/src/app.e2e-spec.ts
```diff
@@ -8,9 +8,9 @@ describe('workspace-project App', () => {
     page = new AppPage();
   });
 
-  it('should display welcome message', () => {
-    page.navigateTo();
-    expect(page.getTitleText()).toEqual('integration-project app is running!');
+  it('should display welcome message', async () => {
+    await page.navigateTo();
+    await expect(await page.getTitleText()).toEqual('integration-project app is running!');
```
clydin, line 13: The outer await should not be needed here and in the other tests.

clydin (changes requested):

integration/angular_cli/e2e/src/app.e2e-spec.ts
apeeters: The `tsconfig.json.template` file in `packages/schematics/angular/e2e/files` specifies `jasminewd2` in types. The method signatures in these type definitions return Promises. So I believe this change is needed.

apeeters:
> * Update the commit message scopes to `@schematics/angular`
> * Dropping Jasmine bump commit since this seems redundant.

Fixed these remarks and rebased the PR.

integration/angular_cli/e2e/src/app.e2e-spec.ts
alan-agius4: It returns a promise because it's a generic, therefore if you `await` the return type is no longer a promise.

```ts
declare function expect<T>(actual: ArrayLike<T>): jasmine.ArrayLikeMatchers<T>;
```

clydin:
Thank you for the contribution.  This change has, however, been superseded by  https://github.com/angular/angular-cli/pull/18951.  This will be available in the upcoming version 11.

PR #18788, opened 2020-09-15
feat(@angular-devkit/build-angular): auto configure hot module replacement (HMR)

### LINKED ISSUE #17324: HMR configuration - Live Reload still enabled
Hi all,

i've just tried to setup HMR for my angular 9 application- which is working fine for changes within my SCSS files (... stylesheets are updated without page reload).

But it doesn't work out for markup/component files - changes there still trigger a reload of the whole page. 

I assume, that I've done sth. wrong, but I could not figure out what (due to this message during startup):

```
NOTICE: Hot Module Replacement (HMR) is enabled for the dev server.
The project will still live reload when HMR is enabled,
but to take advantage of HMR additional application code is required'
(not included in an Angular CLI project by default).'
See https://webpack.js.org/guides/hot-module-replacement
for information on working with HMR for Webpack.
To disable this warning use "hmrWarning: false" under "serve" options in "angular.json".
```

My configuration is done accordingly to the 'official' but deprecated docs: https://github.com/angular/angular-cli/wiki/stories-configure-hmr

angular.json:

```
[...]
          "configurations": {
            "hmr": {
              "fileReplacements": [{
                "with": "src/environments/environment.hmr.ts"
              }]
            },
            "production": {
              "fileReplacements": [{
              }],
[...]

        "serve": {
          "options": {
            "browserTarget": "PortfolioManagerWebClient:build"
          },
          "configurations": {
            "production": {
              "browserTarget": "PortfolioManagerWebClient:build:production"
            },
            "hmr": {
              "hmr": true,
              "browserTarget": "PortfolioManagerWebClient:build:hmr"
            }
          }
        },
```

package.json:

`    "hmr": "ng serve --configuration=hmr",
`

and env. file (src\environments\environment.hmr.ts)

```
export const environment = {
	production: false,
	hmr: true,
[...]
```

I couldn't figure out, that there is sth. obvious wrong. Could you advise me there? 
Thanks in advance!

alan-agius4:
With this change we configure HMR internally and therefore users which want to use basic HMR functionality will not longer be required to change their application code.

This is important because previously a lot of users missed out on HMR or reported a broken behaviour. This also gives novice/new users a better chance to appreciate HMR and Angular because of the zero effort required to use HMR.

Closes #17324

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/hmr.js
+++ b/packages/angular_devkit/build_angular/src/webpack/hmr.js
@@ -1,0 +1,212 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// TODO: change the file to TypeScript and build soley using Bazel.
+
+import { ApplicationRef, PlatformRef, ɵresetCompiledComponents } from '@angular/core';
+import { filter, take } from 'rxjs/operators';
+
+if (module['hot']) {
+  module['hot'].accept();
+  module['hot'].dispose(() => {
+    if (typeof ng === 'undefined') {
+      console.warn(`[NG HMR] Cannot find global 'ng'. Likely this is caused because scripts optimization is enabled.`);
+
+      return;
+    }
+
+    if (!ng.getInjector) {
+      // View Engine
+      return;
+    }
+
+    // Reset JIT compiled components cache
+    ɵresetCompiledComponents();
+    const appRoot = getAppRoot();
+    if (!appRoot) {
+      return;
+    }
+
+    const appRef = getApplicationRef(appRoot);
+    if (!appRef) {
+      return;
+    }
+
+    const oldInputs = document.querySelectorAll('input, textarea');
+    const oldOptions = document.querySelectorAll('option');
+
+    // Create new application
+    appRef.components
+      .forEach(cp => {
+        const element = cp.location.nativeElement;
+        const parentNode = element.parentNode;
+        parentNode.insertBefore(
+          document.createElement(element.tagName),
+          element,
+        );
+
+        parentNode.removeChild(element);
+      });
+
+    // Destroy old application, injectors, <style..., etc..
+    const platformRef = getPlatformRef(appRoot);
+    if (platformRef) {
+      platformRef.destroy();
+    }
+
+    // Restore all inputs and options
+    const bodyElement = document.body;
+    if ((oldInputs.length + oldOptions.length) === 0 || !bodyElement) {
+      return;
+    }
+
+    // Use a `MutationObserver` to wait until the app-root element has been bootstrapped.
+    // ie: when the ng-version attribute is added.
+    new MutationObserver((_mutationsList, observer) => {
+      observer.disconnect();
+
+      const newAppRoot = getAppRoot();
+      if (!newAppRoot) {
+        return;
+      }
+      
+      const newAppRef = getApplicationRef(newAppRoot);
+      if (!newAppRef) {
+        return;
+      }
+
+      // Wait until the application isStable to restore the form values
+      newAppRef.isStable
+        .pipe(
+          filter(isStable => !!isStable),
+          take(1),
+        )
+        .subscribe(() => restoreFormValues(oldInputs, oldOptions));
+    })
+    .observe(bodyElement, {
+      attributes: true,
+      subtree: true,
+      attributeFilter: ['ng-version'],
+    });
+  });
+}
+
+function getAppRoot() {
+  const appRoot = document.querySelector('[ng-version]');
+  if (!appRoot) {
+    console.warn('[NG HMR] Cannot find the application root component.');
+
+    return undefined;
+  }
+
+  return appRoot;
+}
+
+function getToken(appRoot, token) {
+  return typeof ng === 'object' && ng.getInjector(appRoot).get(token) || undefined;
+}
+
+function getApplicationRef(appRoot) {
+  const appRef = getToken(appRoot, ApplicationRef);
+  if (!appRef) {
+    console.warn(`[NG HMR] Cannot get 'ApplicationRef'.`);
+
+    return undefined;
+  }
+
+  return appRef;
+}
+
+function getPlatformRef(appRoot) {
+  const platformRef = getToken(appRoot, PlatformRef);
+  if (!platformRef) {
+    console.warn(`[NG HMR] Cannot get 'PlatformRef'.`);
+
+    return undefined;
+  }
+
+  return platformRef;
+}
+
+function dispatchEvents(element) {
+  element.dispatchEvent(new Event('input', {
+    bubbles: true,
+    cancelable: true,
+  }));
+
+  element.blur();
+
+  element.dispatchEvent(new KeyboardEvent('keyup', { key: 'Enter' }));
+}
+
+function restoreFormValues(oldInputs, oldOptions) {
+  // Restore input
+  const newInputs = document.querySelectorAll('input, textarea');
+  if (newInputs.length && newInputs.length === oldInputs.length) {
+    console.log('[NG HMR] Restoring input/textarea values.');
+    for (let index = 0; index < newInputs.length; index++) {
+      const newElement = newInputs[index];
+      const oldElement = oldInputs[index];
+
+      switch (oldElement.type) {
+        case 'button':
+        case 'image':
+        case 'submit':
+        case 'reset':
+          // These types don't need any value change.
+          continue;
+        case 'radio':
+        case 'checkbox':
+          newElement.checked = oldElement.checked;
+          break;
+        case 'color':
+        case 'date':
+        case 'datetime-local':
+        case 'email':
+        case 'file':
+        case 'hidden':
+        case 'image':
+        case 'month':
+        case 'number':
+        case 'password':
+        case 'radio':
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
```diff
@@ -37,7 +37,10 @@ export class CommonJsUsageWarnPlugin {
   // Allow the below depedency for HMR
   // tslint:disable-next-line: max-line-length
   // https://github.com/angular/angular-cli/blob/1e258317b1f6ec1e957ee3559cc3b28ba602f3ba/packages/angular_devkit/build_angular/src/dev-server/index.ts#L605-L638
-  private allowedDepedencies = new Set<string>(['webpack/hot/dev-server']);
+  private allowedDepedencies = new Set<string>([
+    '@angular-devkit/build-angular',
```
clydin, line 41: Is this needed?  The `hmr.js` has import statements so it should be considered ESM.

alan-agius4: Ah not anymore.

packages/angular_devkit/build_angular/src/webpack/hmr.js
```diff
@@ -0,0 +1,175 @@
...
+      if (!newAppRef) {
+        return;
+      }
+
+      // Wait until the application isStable to restore the form values
+      newAppRef.isStable
+        .pipe(
+          filter(isStable => !!isStable),
+          take(1),
+          tap(() => restoreFormValues(oldInputs, oldOptions)),
```
clydin, line 88: I think the `tap` operator could be dropped and the handler moved into the `subscribe` call?

alan-agius4: That can be done.

clydin (approved):

packages/angular_devkit/build_angular/src/webpack/hmr.js
```diff
@@ -0,0 +1,174 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// TODO: change the file to TypeScript and build soley using Bazel.
```
kyliau, line 9: What's stopping us from doing it today? Global typings?

packages/angular_devkit/build_angular/src/webpack/hmr.js
```diff
@@ -0,0 +1,174 @@
...
+        case 'radio':
+        case 'checkbox':
+          newElement.checked = oldElement.checked;
+          break;
+        default:
+          newElement.value = oldElement.value;
+          break;
+      }
+
+      dispatchKeyUpEvent(newElement);
```
kyliau, line 159: What's the purpose of dispatching this event?

packages/angular_devkit/build_angular/src/webpack/hmr.js
```diff
@@ -0,0 +1,174 @@
...
+}
+
+function dispatchKeyUpEvent(element) {
+  element.dispatchEvent(new KeyboardEvent('keyup', { key: 'Enter' }));
+}
+
+function restoreFormValues(oldInputs, oldOptions) {
+  // Restore input
+  const newInputs = document.querySelectorAll('input');
+  if (newInputs.length && newInputs.length === oldInputs.length) {
```
kyliau, line 143: Should we print a warning if we are not able to restore old inputs?
This is a very conservative assumption, because if newInputs.length > oldInputs.length we won't restore anything.

packages/angular_devkit/build_angular/src/webpack/hmr.js
```diff
@@ -0,0 +1,174 @@
...
+      });
+
+    // Destroy old application, injectors, <style..., etc..
+    const platformRef = getPlatformRef(appRoot);
+    if (platformRef) {
+      platformRef.destroy();
+    }
+
+    // Restore all inputs and options
+    const bodyElement = document.querySelector('body');
```
dgp1130, line 63: Can we use `document.body` here?

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: Since we have a single compilation we cannot include both node and Dom libs as it causes conflicts for instance when using `console`.

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: Yeah I’ll add that

packages/angular_devkit/build_angular/src/webpack/hmr.js
```diff
@@ -0,0 +1,174 @@
...
+    console.warn(`[NG HMR] Cannot get 'PlatformRef'.`);
+
+    return undefined;
+  }
+
+  return platformRef;
+}
+
+function dispatchKeyUpEvent(element) {
+  element.dispatchEvent(new KeyboardEvent('keyup', { key: 'Enter' }));
```
dgp1130, line 137: Why the enter key? I'm assuming this is to "flush" the restored values, but couldn't triggering an enter have other possible effects (such as submitting a `type="submit"` input)?

Would trigger a blur event be safer, or is there another reason for this?

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: The reason for this that users add event listeners on keyup.

Such as to fetch data from an API or store.

I’ll add a comment

packages/angular_devkit/build_angular/src/webpack/hmr.js
```diff
@@ -0,0 +1,174 @@
...
+      const oldElement = oldInputs[index];
+
+      switch (oldElement.type) {
+        case 'radio':
+        case 'checkbox':
+          newElement.checked = oldElement.checked;
+          break;
+        default:
+          newElement.value = oldElement.value;
+          break;
```
dgp1130, line 156: Assuming the default is has a `value` prop could fail if W3C introduces a new `<input />` type that doesn't use `value`. I'm thinking we should list out all the currently supported types explicitly and then default to `console.warn("Unsupported input type.")` or something like that. It would be safer and avoid breaking us from otherwise non-breaking platform changes.

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: In future this can be improved, maybe by matching inputs by the name attribute or something more clever.

packages/angular_devkit/build_angular/src/webpack/hmr.js
```diff
@@ -0,0 +1,174 @@
...
+  return platformRef;
+}
+
+function dispatchKeyUpEvent(element) {
+  element.dispatchEvent(new KeyboardEvent('keyup', { key: 'Enter' }));
+}
+
+function restoreFormValues(oldInputs, oldOptions) {
+  // Restore input
+  const newInputs = document.querySelectorAll('input');
```
dgp1130, line 142: What about shadow roots? I don't think this would pick up input elements hidden under a shadow root, though I don't know how commonly those are used in Angular applications.

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: Actually maybe it might work if we enable `isolatedModules`.

Will give it a try tomorrow.

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: Yes.

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: Agree

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: https://github.com/angular/angular-cli/pull/18788#discussion_r494632207

Probably should ignore type submit

packages/angular_devkit/build_angular/src/webpack/hmr.js
dgp1130: Is `keyup` the only event we expect users to listen to that might be affected? What about `blur` or `input`? Would we expect those events to trigger as well?

alan-agius4: I think we can add these two events as well.

packages/angular_devkit/build_angular/src/webpack/hmr.js
alan-agius4: I don't think it's that common, I don't want to add complexity and use-cases to support just for the sake of it. I think we can evaluate shadow root based on the community feedback.

kyliau (approved):

packages/angular_devkit/build_angular/src/webpack/hmr.js
dgp1130: I'm fine to defer it, though we may want to document that limitation somewhere.

dgp1130 (approved):

PR #18820, opened 2020-09-18
Experimental Webpack 5 Support

clydin:
This change enables experimental support for Webpack v5 in the Angular CLI. At this point, we don't recommend using Webpack v5 for production usage.

By default, the Angular CLI uses Webpack v4. To enable Webpack v5, use `yarn` as your package manager and add the following top-level property to `package.json`:
```
  "resolutions": {
    "webpack": "^5.0.0-beta.30"
  }
```

Please note that the support is experimental, and webpack v5 is still in beta. Some Angular CLI features and options are partially supported, and new prereleases of webpack v5 may contain breaking changes. You may temporarily see Webpack deprecation warnings, which we'll fix over time.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
@@ -32,6 +32,7 @@ import {
 } from 'webpack-sources';
 import { allowMangle, allowMinify, shouldBeautify } from './environment-options';
 import { I18nOptions } from './i18n-options';
+import { isWebpackFiveOrHigher } from './webpack-version';
 
 type LocalizeUtilities = typeof import('@angular/localize/src/tools/src/source_file_utils');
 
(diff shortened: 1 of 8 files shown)


alan-agius4 (commented):
👍

alan-agius4 (approved):

xiaoxiangmoe:
@clydin webpack relase [v5.0.0-rc.0](https://github.com/webpack/webpack/releases/tag/v5.0.0-rc.0) now.

tatsujb:
Will the  2020-10-10 release of webpack 5 imply a new version of angular-cli so as to fix a year-long chain-dependency deprecation issue?

Angular Cli is actually unbuildable on mac 10.15 (catalina) as it stands today. 

This, in my view, should make it into v10.1.5 and that be released on the day ( 2020-10-10 )

PR #18837, opened 2020-09-21
fix(@schematics/angular): remove IE9/10 from browserslist for new projects

clydin:
BREAKING CHANGE:
Deprecated support for IE 9-10 has been removed in Angular 11.
Only IE 9-10 support has been removed. IE 11 is officially supported in Angular 11.
Additional information can be found here: https://v10.angular.io/guide/deprecations#ie-9-10-and-mobile

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/index.ts
+++ b/packages/angular_devkit/build_angular/src/browser/index.ts
@@ -256,6 +256,19 @@ export function buildWebpackBrowser(
         `);
         }
 
+        const hasIE9 = buildBrowserFeatures.supportedBrowsers.includes('ie 9');
+        const hasIE10 = buildBrowserFeatures.supportedBrowsers.includes('ie 10');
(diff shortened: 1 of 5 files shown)


alan-agius4 (approved):

alan-agius4:
Should we do a migration to remove these from existing project?

clydin:
I think leaving existing projects is acceptable.  IE9/10 while not officially supported may still work for a given project and users may wish to keep that behavior.
I do think some form of warning and/or future health check audit may be prudent to help users optimally configure their projects.

kyliau (changes requested):
Please add this link to the commit message: https://angular.io/guide/deprecations#ie-9-10-and-mobile

kyliau (approved):
We should either do a migration or print out a warning when users upgrade to v11.
Could you please do this as a follow up? Thank you.

PR #18843, opened 2020-09-22
Give generated tests typed descriptions

boyum:
Whenever class names change, test descriptions often are overlooked, as IDE's won't change them automatically. By using Function#name, we can render the exact same strings without having to worry about changes elsewhere.

I'm a first time contributor and I most probably do things in the wrong order or in a non-angular style. Please feel free to correct me and to lead the way. ✨

This is a redo of #18835 to fix commit messages


packages/schematics/angular/class/files/__name@dasherize____type__.spec.ts.template
```diff
@@ -1,6 +1,6 @@
 import { <%= classify(name) %> } from './<%= dasherize(name) %><%= type %>';
 
-describe('<%= classify(name) %>', () => {
+describe(<%= classify(name) %>.name, () => {
```
alan-agius4, line 3: These should be wrapped in backticks, otherwise it will cause a runtime error.

alan-agius4 (commented):
Personally, I am not a big fan of this. I find that with this pattern it's makes it harder to find failling unit tests, but that is just my personal opinion.

boyum:
> Personally, I am not a big fan of this. I find that with this pattern it's makes it harder to find failling unit tests, but that is just my personal opinion.

I'm not sure if I understand you. Could you please elaborate why this would affect looking for the unit tests? Are you thinking about the case where devs search for `describe('MyComponent` to find the given test file?

Rogier-blip:
The benefit I see to this approach is the following:

When you, for example, change the name of a component the test automatically gets changed as well. This makes refactoring easier. To @boyum point, also changing the name when you change the name of a class gets overlooked often. I have seen multiple projects where many class names and test names are out of sync. 

@alan-agius4 To your point, it will still be easy to find failing tests. In the overview of the test run, you will see the same name of the failing test as you would see if the string value of the class was there. Besides that, if you do a search in your IDE for the failing test name you will still be able to find the test since the class name is still there there is just a .name after the class name.

I believe this is a high-value feature for the maintainability of the angular tests.

dgp1130:
After discussion amongst the team, our consensus is to keep generating the string literal rather than a reference to the class name.

See https://github.com/angular/angular-cli/issues/19026#issuecomment-709576265 for additional info.

PR #18845, opened 2020-09-22
fix(@angular/cli): favor dirname when resolving @schematics/angular

### LINKED ISSUE #18840: ng generate interceptor - no schematic "interceptor" found.
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

It seems like this has worked previously.

### Description

Not able to generate an interceptor from schematic in ng cli 10.1.2

## 🔬 Minimal Reproduction

run `ng g interceptor <interceptor-name>`

## 🔥 Exception or Error
<pre><code>
An unhandled exception occurred: Schematic "interceptor" not found in collection "@schematics/angular".
See "/private/var/folders/q2/xny4d6016zz_0pdpg8cjybcr0000gn/T/ng-rA8whv/angular-errors.log" for further details.
</code></pre>

## 🌍 Your Environment
<pre><code>
Angular CLI: 10.1.2
Node: 14.9.0
OS: darwin x64

Angular: 10.1.2
Ivy Workspace: Yes

@angular-devkit/architect       0.1001.2
@angular-devkit/build-angular   0.1001.2
@angular-devkit/core            8.0.0
@angular-devkit/schematics      8.0.0
@schematics/angular             8.0.0
@schematics/update              0.1001.2
rxjs                            6.5.5
typescript                      3.9.7
webpack                         4.44.2
</code></pre>

alan-agius4:
This ensures that the correct build-in version of `@schematics/angular` is resolved.

Closes #18840

### DIFF
--- a/packages/angular/cli/models/schematic-command.ts
+++ b/packages/angular/cli/models/schematic-command.ts
@@ -256,7 +256,10 @@ export abstract class SchematicCommand<
       registry: new schema.CoreSchemaRegistry(formats.standardFormats),
       resolvePaths: !!this.workspace.configFile
         // Workspace
-        ? [process.cwd(), this.workspace.root, __dirname]
+        ? this.collectionName === this.defaultCollectionName
+          // Favor __dirname for @schematics/angular to use the build-in version
+          ? [__dirname, process.cwd(), this.workspace.root]
+          : [process.cwd(), this.workspace.root, __dirname]
         // Global
         : [__dirname, process.cwd()],
     });


packages/angular/cli/models/schematic-command.ts
```diff
@@ -256,7 +256,7 @@ export abstract class SchematicCommand<
       registry: new schema.CoreSchemaRegistry(formats.standardFormats),
       resolvePaths: !!this.workspace.configFile
         // Workspace
-        ? [process.cwd(), this.workspace.root, __dirname]
+        ? [__dirname, process.cwd(), this.workspace.root]
```
clydin, line 259: This will cause very different schematic resolution especially in monorepos.  The location of invocation should take precedence locally.
The root cause of the referenced issue is really package hoisting combined with a dependency using a much older version of the built-in Angular schematics.  Unfortunately we cannot fix that so an alternative may be to special case the built-in `@schematics/angular` package.  This would break users intentionally  installing a different version of the package but is probably the least invasive fix.

clydin (approved):

PR #18847, opened 2020-09-22
fix(@ngtools/webpack): add Webpack 5 to peer dependencies

valorkin:
Very minor addition to https://github.com/angular/angular-cli/pull/18820
adding optional webpack@5 to peer dependencies of ngtools/webpack

### DIFF
--- a/packages/ngtools/webpack/package.json
+++ b/packages/ngtools/webpack/package.json
@@ -28,7 +28,7 @@
   "peerDependencies": {
     "@angular/compiler-cli": "^11.0.0 || ^11.0.0-next",
     "typescript": ">=3.9 < 4.1",
-    "webpack": "^4.0.0"
+    "webpack": "^4.0.0 || ^5.0.0-rc.0"
   },
   "devDependencies": {
     "@angular/compiler": "11.0.0-next.2",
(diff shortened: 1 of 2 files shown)


packages/ngtools/webpack/package.json
```diff
@@ -28,7 +28,7 @@
   "peerDependencies": {
     "@angular/compiler-cli": "^11.0.0 || ^11.0.0-next",
     "typescript": ">=3.9 < 4.1",
-    "webpack": "^4.0.0"
+    "webpack": "^4.0.0 || ^5.0.0"
```
alan-agius4, line 31: Webpack version 5 is not yet stable thus the semver version here is invalid.  It should be `^5.0.0-rc.0`. 

Chiming in @clydin just to confirm whether we should add this `peerDependencies` in the first place since Webpack 5 support is still experimental, and adding Webpack 5 might send out the wrong message.

alan-agius4: PS: `chore` is an invalid commit message type. You can use `fix` for this.

`fix(@ngtools/webpack): add Webpack 5 to peer dependencies`

valorkin: I know, just made it the same way as it appears in all other occurrences. I can change those to rc.0 too

valorkin: all comments applied

clydin:
However, this has been superseded by #20037.

PR #18850, opened 2020-09-22
Angular schematics cleanup - Part 3

clydin:
This PR continues updating older schematics to use newer utilities and helper functions with the goal to reduce complexity as well as remove older helper code.


alan-agius4 (approved):
Nice

Issue #18744, opened 2020-09-09
update migration to remove solution style tsconfig is not run


Splaktar:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?
No

### Description

Running `ng update @angular/cli` to update to v10.1.0 from v10.0.8 does not run the migration to remove solution style tsconfigs.

## 🔬 Minimal Reproduction

1. `ng update @angular/cli`
```shell
Your global Angular CLI version (10.1.0) is greater than your local
version (10.0.8). The local Angular CLI version is used.

To disable this warning use "ng config -g cli.warnings.versionMismatch false".
The installed local Angular CLI version is older than the latest stable version.
Installing a temporary version to perform the update.
Installing packages for tooling via npm.
Installed packages for tooling via npm.
Using package manager: 'npm'
Collecting installed dependencies...
Found 54 dependencies.
Fetching dependency metadata from registry...
    Updating package.json with dependency @angular-devkit/build-angular @ "0.1001.0" (was "0.1000.8")...
    Updating package.json with dependency @angular/cli @ "10.1.0" (was "10.0.8")...
UPDATE package.json (3764 bytes)
✔ Packages installed successfully.
splaktar:xlts.dev splaktar$ npm i
npm WARN @angular/fire@6.0.0 requires a peer of @angular/common@^9.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @angular/fire@6.0.0 requires a peer of @angular/core@^9.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @angular/fire@6.0.0 requires a peer of @angular/platform-browser@^9.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @angular/fire@6.0.0 requires a peer of @angular/platform-browser-dynamic@^9.0.0 but none is installed. You must install peer dependencies yourself.

removed 4 packages and audited 2828 packages in 12.241s

126 packages are looking for funding
  run `npm fund` for details

found 4 high severity vulnerabilities
  run `npm audit fix` to fix them, or `npm audit` for details
```
2. Notice that there are no migrations run
3. Check that `tsconfig.base.json` still exists

## 🔥 Exception or Error
None

[...]

Issue #18040, opened 2020-06-26
New "Solution Style" tsconfig.json file breaks the unused files in most Editors


destus90:
If we have an unused service, we will run into some Typescript errors in Editors since the unused file isn't really a part of any TypeScript compilation.
**VS code**
![image](https://user-images.githubusercontent.com/17944366/85844033-8963ae00-b7bb-11ea-8be1-68b98bb26262.png)
**Webstorm**
![image](https://user-images.githubusercontent.com/17944366/85844101-a13b3200-b7bb-11ea-9702-3be43650809e.png)

Shall we revert the changes for tsconfig.file?

kyliau:
Please make sure you are using the workspace version of Typescript. 
Solution-style tsconfig is not supported by typescript < v3.9.
In vscode, you can select typescript version like so:

<img width="781" alt="Screen Shot 2020-06-27 at 9 37 48 AM" src="https://user-images.githubusercontent.com/2941178/85927314-25dc8d80-b85a-11ea-80b4-ead4886fdc2f.png">

destus90:
@kyliau 
Thanks for your suggestion. However, my TypeScript version is set as a workspace version. The real problem here has been described by @alan-agius4 https://github.com/angular/angular/issues/37781#issuecomment-650573639. I have created a repo https://github.com/destus90/angular-v10-issues. It would be nice if you could investigate it (file src/app/foo.service.ts). I found out that adding unused files into tsconfig.app.json files array solves the issue, but it would pain us to manually put all unused files into it. Any further assistance would be appreciated. 
![ng10issue](https://user-images.githubusercontent.com/17944366/85938188-4a9d3780-b924-11ea-85b0-207f47f36ef9.png)

alan-agius4:
We are aware that “unused” files safe are effected as described above.

This is because we have a strict file Inclusion policy in our tsconfig, which results in un-referenced files not being part of correct program. 

We’ll be discussing this issue during our meeting this week.

cyrilletuzi:
Hi @alan-agius4, what is the status about this issue? I just discovered it while animating an Angular training and it's a real pain. Any time you create a service, as shown in the screenshot of the initial message, there is an error telling that `experimentalDecorators` should be set to `true`.

[...]

PR #18851, opened 2020-09-22
fix(@schematics/angular): remove solution style throw an error

eranshmil:
`remove-solution-style-tsconfig` migration fails when upgrading from version 10.0.* to 10.1.* with the following error:
```bash
[error] Error: Could not read 'tsconfig.json'
```

There is no issue regarding this but you can see the following comments:
https://github.com/angular/angular-cli/issues/18744#issuecomment-689790320
https://github.com/angular/angular-cli/issues/18040#issuecomment-689833338

Btw, the files are not formatted correctly by the `prettier` configuration but I decided to leave these changes out of the scope of this PR.

### DIFF
--- a/packages/schematics/angular/migrations/update-10/remove-solution-style-tsconfig.ts
+++ b/packages/schematics/angular/migrations/update-10/remove-solution-style-tsconfig.ts
@@ -35,16 +35,23 @@ export default function (): Rule {
   return (host, context) => {
     const logger = context.logger;
 
-    const files = new JSONFile(host, 'tsconfig.json').get(['files']);
-    if (!(Array.isArray(files) && files.length === 0)) {
-      logger.info('Migration has already been executed.');
+    const tsConfigExists = host.exists('tsconfig.json');
+    if (tsConfigExists) {
+      const files = new JSONFile(host, 'tsconfig.json').get(['files']);
+      if (!(Array.isArray(files) && files.length === 0)) {
+        logger.info('Migration has already been executed.');
 
-      return;
+        return;
+      }
     }
 
     if (host.exists('tsconfig.base.json')) {
-      host.overwrite('tsconfig.json', host.read('tsconfig.base.json') || '');
-      host.delete('tsconfig.base.json');
+      if (tsConfigExists) {
+        host.overwrite('tsconfig.json', host.read('tsconfig.base.json') || '');
+        host.delete('tsconfig.base.json');
+      } else {
+        host.rename('tsconfig.base.json', 'tsconfig.json');
+      }
     }
 
     // Iterate over all tsconfig files and change the extends from 'tsconfig.base.json' to 'tsconfig.json'.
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/update-10/remove-solution-style-tsconfig.ts
```diff
@@ -35,11 +35,16 @@ export default function (): Rule {
...
+      const files = new JSONFile(host, 'tsconfig.json').get(['files']);
 
-      return;
+      if (!(Array.isArray(files) && files.length === 0)) {
+        logger.info('Migration has already been executed.');
+
+        return;
+      }
+    } catch {
+      host.create('tsconfig.json', '');
```
alan-agius4, line 47: Maybe instead of create rename `tsconfig.base.json`?

```ts
    const tsConfigExists = host.exists('tsconfig.json');
    if (tsConfigExists) {
      const files = new JSONFile(host, 'tsconfig.json').get(['files']);
      if (!(Array.isArray(files) && files.length === 0)) {
        logger.info('Migration has already been executed.');

        return;
      }
    }

    if (host.exists('tsconfig.base.json')) {
      if (tsConfigExists) {
        host.overwrite('tsconfig.json', host.read('tsconfig.base.json') || '');
      } else {
        host.rename('tsconfig.base.json', 'tsconfig.json');
      }

      host.delete('tsconfig.base.json');
    }
```

packages/schematics/angular/migrations/update-10/remove-solution-style-tsconfig.ts
```diff
@@ -35,11 +35,16 @@ export default function (): Rule {
   return (host, context) => {
     const logger = context.logger;
 
-    const files = new JSONFile(host, 'tsconfig.json').get(['files']);
-    if (!(Array.isArray(files) && files.length === 0)) {
-      logger.info('Migration has already been executed.');
+    try {
```
alan-agius4, line 38: I think in this case, it would be better to check if the file exists instead of using a try/catch block.

packages/schematics/angular/migrations/update-10/remove-solution-style-tsconfig_spec.ts
```diff
@@ -70,7 +70,6 @@ describe('Migration to remove "Solution Style" tsconfig', () => {
 
     // Create tsconfigs
     const compilerOptions = { target: 'es2015' };
-    createJsonFile(tree, 'tsconfig.json', { files: [] });
```
alan-agius4, line 73: This changes the behaviour of all test cases.

I think it's best to create a separate test case.

Ex:
```
  it(`should not error when 'tsconfig.json' doesn't exist`, async () => {
    tree.delete('tsconfig.json');
    const newTree = await schematicRunner.runSchematicAsync(schematicName, {}, tree).toPromise();
    expect(readJsonFile(newTree, 'tsconfig.json')['compilerOptions']).toBeTruthy();
    expect(newTree.exists('tsconfig.base.json')).toBeFalse();
  });
```

alan-agius4 (changes requested):
Thanks for this, a couple of suggested changed above.

packages/schematics/angular/migrations/update-10/remove-solution-style-tsconfig.ts
eranshmil: I changed it to your sugesstion but moved the `host.delete` line to be after the `host.overwrite` because it throws an error after a rename.

alan-agius4 (approved):
LGTM thanks

PR #18869, opened 2020-09-24
fix(@angular-devkit/build-angular): add missing defaults to `optimization` property

alan-agius4:


clydin (approved):

alan-agius4:
Just remembed these are intentional left out because of

PR #18879, opened 2020-09-25
feat(@angular-devkit/build-angular): support package references in styles & scripts options

clydin:
The browser builder's `styles` and `scripts` options now support using a package name in the path when specifying a style or script.  This removes the need to use a relative path to the node modules directory in these options.  This provides support for Yarn PnP as well as reducing the complexity of the options especially for monorepo setups.  Relatively located files will take precedence over packages if they exist.  This precedence provides backwards compatibility with existing configurations.

Before :
`"styles": ["../node_modules/bootstrap/dist/css/bootstrap.css"]`

After:
`"styles": ["bootstrap/dist/css/bootstrap.css"]`

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/specs/styles_spec.ts
+++ b/packages/angular_devkit/build_angular/src/browser/specs/styles_spec.ts
@@ -570,13 +570,23 @@ describe('Browser Builder styles', () => {
     expect(main).toContain(`url('/base/assets/component-img-absolute.svg')`);
   }, 90000);
 
-  it(`supports bootstrap@4`, async () => {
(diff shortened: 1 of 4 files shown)


alan-agius4 (changes requested):
Can you also update please https://github.com/angular/angular-cli/blob/75e34bc17ff637b639ebce8f9abc479a5b4df0ab/packages/angular_devkit/build_angular/src/dev-server/index.ts#L579-L581

Maybe the path should be resolved in `normalizeExtraEntryPoints`?

alan-agius4 (approved):

PR #18884, opened 2020-09-25
fix(@angular-devkit/build-angular): webpack5 deprecation of module property in Dependency

valorkin:
fixes 24 occurances of `Error: module property was removed from Dependency (use compilation.moduleGraph.getModule(dependency) instead)`
in:
```bash
yarn bazel test //packages/angular_devkit/build_angular:build_angular_browser_test
yarn bazel test //packages/angular_devkit/build_angular:build_angular_browser_ve_test
```

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
@@ -1,4 +1,3 @@
-
 /**
  * @license
  * Copyright Google Inc. All Rights Reserved.
@@ -10,6 +9,7 @@
 import { isAbsolute } from 'path';
 import { Compiler, compilation } from 'webpack';
 import { addWarning } from '../../utils/webpack-diagnostics';
+import { isWebpackFiveOrHigher } from '../../utils/webpack-version';
 
 // Webpack doesn't export these so the deep imports can potentially break.
 const CommonJsRequireDependency = require('webpack/lib/dependencies/CommonJsRequireDependency');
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
```diff
@@ -100,13 +103,17 @@ export class CommonJsUsageWarnPlugin {
...
+    compilation: compilation.Compilation,
+    dependencies: WebpackModule[],
+    checkParentModules = false): boolean {
     for (const dep of dependencies) {
       if (dep instanceof CommonJsRequireDependency || dep instanceof AMDDefineDependency) {
         return true;
       }
 
-      if (checkParentModules && dep.module && this.hasCommonJsDependencies(dep.module.dependencies)) {
+      const module = getWebpackModule(compilation, dep);
```
alan-agius4, line 115: Can we avoid this when `checkParentModules` is false?

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
```diff
@@ -100,13 +103,17 @@ export class CommonJsUsageWarnPlugin {
     });
   }
 
-  private hasCommonJsDependencies(dependencies: WebpackModule[], checkParentModules = false): boolean {
+  private hasCommonJsDependencies(
+    compilation: compilation.Compilation,
```
alan-agius4, line 107: Maybe just provide the `moduleGraph` instead of the entire `compilation`.

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
```diff
@@ -10,12 +9,16 @@
...
+import { isWebpackFiveOrHigher } from '../../utils/webpack-version';
 
 // Webpack doesn't export these so the deep imports can potentially break.
 const CommonJsRequireDependency = require('webpack/lib/dependencies/CommonJsRequireDependency');
 const AMDDefineDependency = require('webpack/lib/dependencies/AMDDefineDependency');
 
 // The below is extended because there are not in the typings
+// TODO_WEBPACK_5 type Types of property 'issuer' are incompatible.
+// Type 'WebpackModule | null' is not assignable to type 'Module'
+// @ts-ignore
```
alan-agius4, line 21: Since we don't have Webpack 5 as a dev dependencies we should be able to remove this.

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
```diff
@@ -121,5 +128,23 @@ export class CommonJsUsageWarnPlugin {
       // Non-scoped request ex: lodash/isEmpty -> lodash
       : rawRequest.split('/', 1)[0];
   }
+}
+
+// TODO_WEBPACK_5 ModuleGraph type stub
```
alan-agius4, line 133: I don't see any value added by having these stub types, considering they are used  below in a single method and you're casting.

Might as well cast inline.

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
valorkin: `moduleGraph` it doesn't exist in wepback@4 as far as I know

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
valorkin: this allow's to run unit tests with webpack@5

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
valorkin: ok

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
valorkin: ok

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
valorkin: removed

packages/angular_devkit/build_angular/src/webpack/plugins/common-js-usage-warn-plugin.ts
valorkin: to expand my thought a little:
- if I provide an entire `compilation` object I can hide and inline types for webpack@5
- if I will try to provide only `moduleGraph`, invocation line will contain `as unknown as ModuleGraph`, 3 times in this file
So, I though 1st option is more clean

alan-agius4: SGTM

alan-agius4 (approved):
LGTM thnaks

PR #18893, opened 2020-09-25
fix(@angular-devkit/build-angular): add karma as an optional peer dependency

clydin:
karma is currently used by the karma builder within this package but is not represented in the dependencies.  The can lead to accidental version mismatches as well as package manager hoisting problems due to the package manager not knowing the full dependency set of the package.


alan-agius4 (approved):
I think there are also other deps such as tslint, protoactor etc.. but this is a good start!

alan-agius4 (approved):

clydin:
added protractor

alan-agius4 (approved):

PR #18894, opened 2020-09-25
refactor(@angular-devkit/build-angular): remove redundant tslint vers…

alan-agius4:
…ion check

Tslint has been added as an optional peer dependency which makes this check unnecessary since the package manager will issue a warning when an incorrect version is installed.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tslint/index.ts
+++ b/packages/angular_devkit/build_angular/src/tslint/index.ts
@@ -22,18 +22,6 @@ interface LintResult extends tslint.LintResult {
   fileNames: string[];
 }
 
-async function _loadTslint() {
-  let tslint;
-  try {
-    tslint = await import('tslint');
-  } catch {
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/tslint/index.ts
```diff
@@ -64,7 +43,7 @@ async function _run(
     throw new Error('A "project" must be specified to enable type checking.');
   }
 
-  const projectTslint = await _loadTslint();
+  const projectTslint = await import('tslint');
```
clydin, line 46: Can you wrap this in a try/catch with a builder error result returned if this fails to find tslint?  With it being an optional dependency it is a potential error case.

clydin (changes requested):

packages/angular_devkit/build_angular/src/tslint/index.ts
alan-agius4: Done

clydin (approved):

PR #18906, opened 2020-09-28
fix(@angular-devkit/build-angular): maxModules were deprecated and renamed to modulesSpace

valorkin:
maxModules were deprecated and renamed to modulesSpace

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/stats.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/stats.ts
@@ -7,6 +7,7 @@
  */
 
 import { WebpackConfigOptions } from '../../utils/build-options';
+import { isWebpackFiveOrHigher } from '../../utils/webpack-version';
 
 const webpackOutputOptions = {
   colors: true,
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):
LGTM

packages/angular_devkit/build_angular/src/webpack/configs/stats.ts
```diff
@@ -41,6 +41,12 @@ const verboseWebpackOutputOptions = {
   logging: 'verbose',
 };
 
+if (!isWebpackFiveOrHigher()) {
```
alan-agius4, line 44: Nit: the not/! here can be removed if the logic below is inverted.

```
if (isWebpackFiveOrHigher()) {
  verboseWebpackOutputOptions['modulesSpace'] = Infinity;
} else {
  verboseWebpackOutputOptions['maxModules'] = Infinity;
}
```

packages/angular_devkit/build_angular/src/webpack/configs/stats.ts
```diff
@@ -25,11 +26,10 @@ const webpackOutputOptions = {
   moduleTrace: false,
 };
 
-const verboseWebpackOutputOptions = {
+const verboseWebpackOutputOptions: {[index: string]: boolean|string|number} = {
```
alan-agius4, line 29:
```suggestion
const verboseWebpackOutputOptions: Record<string, boolean | string | number> = {
```

packages/angular_devkit/build_angular/src/webpack/configs/stats.ts
valorkin: done

packages/angular_devkit/build_angular/src/webpack/configs/stats.ts
valorkin: applied

alan-agius4 (approved):

valorkin:
thanks

PR #18914, opened 2020-09-28
fix(@angular-devkit/build-angular): fixes optimizeChunkAssets is deprecated in webpack 5

valorkin:
should fix 83 warnings `[DEP_WEBPACK_COMPILATION_OPTIMIZE_CHUNK_ASSETS] DeprecationWarning: optimizeChunkAssets is deprecated (use Compilation.hook.processAssets instead and use one of Compilation.PROCESS_ASSETS_STAGE_* as stage option)`

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
@@ -8,25 +8,49 @@
 import * as cssNano from 'cssnano';
 import { ProcessOptions, Result } from 'postcss';
 import { Compiler, compilation } from 'webpack';
-import { RawSource, SourceMapSource } from 'webpack-sources';
+import { RawSource, Source, SourceMapSource } from 'webpack-sources';
 import { addWarning } from '../../utils/webpack-diagnostics';
+import { isWebpackFiveOrHigher } from '../../utils/webpack-version';
 
 export interface OptimizeCssWebpackPluginOptions {
   sourceMap: boolean;
   test: (file: string) => boolean;
 }
 
+const PLUGIN_NAME = 'optimize-css-webpack-plugin';
+
 function hook(
   compiler: Compiler,
-  action: (
-    compilation: compilation.Compilation,
-    chunks: Iterable<compilation.Chunk>,
-  ) => Promise<void>,
+  action: (compilation: compilation.Compilation, assetsURI: string[]) => Promise<void>,
 ) {
-  compiler.hooks.compilation.tap('optimize-css-webpack-plugin', (compilation) => {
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
```diff
@@ -8,25 +8,49 @@
...
+
 function hook(
   compiler: Compiler,
-  action: (
-    compilation: compilation.Compilation,
-    chunks: Iterable<compilation.Chunk>,
-  ) => Promise<void>,
+  action: (compilation: compilation.Compilation, assetsURI: string[]) => Promise<void>,
 ) {
   compiler.hooks.compilation.tap('optimize-css-webpack-plugin', (compilation) => {
```
alan-agius4, line 26:
```suggestion
  compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
```

packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
```diff
@@ -8,25 +8,49 @@
...
-    );
+    if (isWebpackFiveOrHigher()) {
+      // webpack 5 migration "guide"
+      // https://github.com/webpack/webpack/blob/07fc554bef5930f8577f91c91a8b81791fc29746/lib/Compilation.js#L527-L532
+      // TODO_WEBPACK_5 const stage = Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE;
+      const stage = 100;
+      // tslint:disable-next-line: no-any
+      (compilation.hooks as any)
+        .processAssets.tapPromise({name: PLUGIN_NAME, stage}, (assets: Record<string, Source>) => {
+          return action(compilation, Object.keys(assets || {}));
```
alan-agius4, line 35: Based on the type above, assets cannot be null/undefined therefore `|| {}` is redundant.

valorkin: I don't trust webpack5 types, but your are right, this one is `{}` at least 
https://github.com/webpack/webpack/blob/07fc554bef5930f8577f91c91a8b81791fc29746/lib/Compilation.js#L743

packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
valorkin: applied

packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
valorkin: done

alan-agius4 (approved):
👍  thanks LGTM

Issue #18730, opened 2020-09-08
[RFC] Eliminate Render Blocking Requests


alan-agius4:
**Authors**: Alan Agius (@alan-agius4)
**Status**: Open
**Closing Date**: 2020-09-22

# Summary
We’re proposing to [eliminate render blocking requests](https://web.dev/render-blocking-resources/) loading by:

- Loading CSS files asynchronously in a single-page application.
- Inline critical CSS in Angular applications which use Angular Universal Server Side Rendering Pre-rendering, App Shell and Angular Client Side Rendered Applications.
- Inline Google Fonts and Icons.

At the moment there is no easy or streamlined way to identify, extract or inline critical CSS in Angular Universal applications. We’re proposing to offer an out-of-the-box solution with little or no configuration needed.

# Motivation

CSS files are [render-blocking](https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-blocking-css) because the browser must download and parse these files before starting to render the page. This makes CSS files a bottleneck when they are large or when having poor or limited network connectivity. Each of these files will result in a penalty on the Performance Score of your application #17966.

We can reduce this render-blocking time and at the same time improve the [first contentful paint](https://web.dev/first-contentful-paint/) (FCP) by extracting and inlining the critical CSS and loading the CSS files asynchronously.

# Proposal
## Load CSS files asynchronously

In most cases JavaScript bundles will take a longer time to download, parse than for Angular to bootstrap and start rendering the first component thus the chance of having Flash of unstyled content (FOUC) is relatively low. 

We are proposing to introduce an experimental async CSS loading use the “**media**” technique which can be opted-in/out via an option angular.json

**Before**
```html
<link rel="stylesheet" href="styles.css">
```
**After**
```html
<link rel="stylesheet" href="styles.css" media="only x" onload="this.media='all'">
<noscript><link rel="stylesheet" href="styles.css"></noscript>
```

## CSS files budgets
CSS files are great for code-sharing, but other than that they are a bottleneck to achieving great performance. Two of the main reasons for this is that they are render blocking and might contain dead-rules. 

Since a CSS file is not strictly associated with the components loaded on the page, a number of CSS declarations that get downloaded and parsed will remain unused by the view that was rendered.

Having render blocking and/or dead-rules will cause performance score penalties in Lighthouse.

We propose to add two new bundle budgets allStyle and anyStyle:

- **anyStyle**: any given external CSS files
- **allStyle**: cumulative size of all external CSS files.

[...]

PR #18926, opened 2020-09-29
feat(@angular-devkit/build-angular): enable font inlining optimizations

alan-agius4:
With this change we inline Google fonts and icons in the index html file when optimization is enabled.

**Before**
```html
<link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```
**After**
```html
<style>
  @font-face {
    font-family: 'Material Icons';
    font-style: normal;
    font-weight: 400;
    src: url(https://fonts.gstatic.com/s/materialicons/v55/flUhRq6tzZclQEJ-Vdg-IuiaDsNcIhQ8tQ.woff2) format('woff2');
  }

  .material-icons {
    font-family: 'Material Icons';
    font-weight: normal;
    font-style: normal;
    font-size: 24px;
    line-height: 1;
    letter-spacing: normal;
    text-transform: none;
    display: inline-block;
    white-space: nowrap;
    word-wrap: normal;
    direction: ltr;
  }
</style>
```

To opt-out of this feature set `optimization.fonts: false` or `optimization.fonts.inline: false` in the browser builder options.

Example:
```js
"configurations": {
  "production": {
    "fileReplacements": [
      {
      }
    ],
    "optimization": {
      "fonts": false
    },     
```

More information about the motivation for this feature can be found: https://github.com/angular/angular-cli/issues/18730

Note: internet access is required during the build for this optimization to work.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
@@ -7,39 +7,37 @@
  */
 
 import * as cacache from 'cacache';
-import { readFile } from 'fs';
+import { readFile as readFileAsync } from 'fs';
 import * as https from 'https';
+import { URL } from 'url';
 import { promisify } from 'util';
 import { findCachePath } from '../cache-path';
 import { cachingDisabled } from '../environment-options';
 import { htmlRewritingStream } from './html-rewriting-stream';
 
-const cacheFontsPath = cachingDisabled ? undefined : findCachePath('angular-build-fonts');
+const cacheFontsPath: string | undefined = cachingDisabled ? undefined : findCachePath('angular-build-fonts');
 const packageVersion = require('../../../package.json').version;
+const readFile = promisify(readFileAsync);
 
 const enum UserAgent {
-  Chrome = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko)',
-  IE = 'Mozilla/5.0 (Windows NT 10.0; Trident/7.0; rv:11.0) like Gecko',
+  Chrome = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.121 Safari/537.36',
+  IE = 'Mozilla/5.0 (Windows NT 10.0; Trident/7.0; rv:11. 0) like Gecko',
 }
 
 const SUPPORTED_PROVIDERS = [
-  'https://fonts.googleapis.com/',
+  'fonts.googleapis.com',
 ];
 
 export interface InlineFontsOptions {
-  content: string;
   minifyInlinedCSS: boolean;
-  WOFF1SupportNeeded: boolean;
+  WOFFSupportNeeded: boolean;
 }
 
 export class InlineFontsProcessor {
-  async process(options: InlineFontsOptions): Promise<string> {
-    const {
-      content,
-      minifyInlinedCSS,
-      WOFF1SupportNeeded,
-    } = options;
 
+  constructor(private options: InlineFontsOptions) { }
+
+  async process(content: string): Promise<string> {
     const hrefList: string[] = [];
 
     // Collector link tags with href
@@ -64,7 +62,7 @@ export class InlineFontsProcessor {
     await new Promise(resolve => collectorStream.on('finish', resolve));
 
     // Download stylesheets
-    const hrefsContent = await this.processHrefs(hrefList, minifyInlinedCSS, WOFF1SupportNeeded);
+    const hrefsContent = await this.processHrefs(hrefList);
     if (hrefsContent.size === 0) {
       return content;
     }
@@ -94,13 +92,13 @@ export class InlineFontsProcessor {
     return transformedContent;
   }
 
-  private async getResponse(url: string, userAgent: UserAgent): Promise<string> {
+  private async getResponse(url: URL, userAgent: UserAgent): Promise<string> {
     const key = `${packageVersion}|${url}|${userAgent}`;
 
     if (cacheFontsPath) {
       const entry = await cacache.get.info(cacheFontsPath, key);
       if (entry) {
-        return promisify(readFile)(entry.path, 'utf8');
+        return readFile(entry.path, 'utf8');
       }
     }
 
@@ -116,7 +114,7 @@ export class InlineFontsProcessor {
         res => {
           res
             .on('data', chunk => rawResponse += chunk)
-            .on('end', () => resolve(rawResponse.toString()));
+            .on('end', () => resolve(rawResponse));
         },
       )
         .on('error', e => reject(e));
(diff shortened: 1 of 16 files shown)


packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
```diff
@@ -0,0 +1,170 @@
...
+    return transformedContent;
+  }
+
+  private async getResponse(url: string, userAgent: UserAgent): Promise<string> {
+    const key = `${packageVersion}|${url}|${userAgent}`;
+
+    if (cacheFontsPath) {
+      const entry = await cacache.get.info(cacheFontsPath, key);
+      if (entry) {
+        return promisify(readFile)(entry.path, 'utf8');
```
kyliau, line 99: consider calling `promisify(readFIle)` once at the top of the file and save as a variable

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
```diff
@@ -0,0 +1,170 @@
...
+        url,
+        {
+          headers: {
+            'user-agent': userAgent,
+          },
+        },
+        res => {
+          res
+            .on('data', chunk => rawResponse += chunk)
+            .on('end', () => resolve(rawResponse.toString()));
```
kyliau, line 115: why do we call `.toString()` when it's already a string?

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
```diff
@@ -0,0 +1,170 @@
...
+
+import * as cacache from 'cacache';
+import { readFile } from 'fs';
+import * as https from 'https';
+import { promisify } from 'util';
+import { findCachePath } from '../cache-path';
+import { cachingDisabled } from '../environment-options';
+import { htmlRewritingStream } from './html-rewriting-stream';
+
+const cacheFontsPath = cachingDisabled ? undefined : findCachePath('angular-build-fonts');
```
kyliau, line 17: please add type annotation

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
```diff
@@ -0,0 +1,170 @@
...
+          normalizedHref = href.replace('http:', 'https:');
+        } else {
+          // Unsupported CSS href.
+          continue;
+        }
+      }
+
+      if (!SUPPORTED_PROVIDERS.some(url => normalizedHref.startsWith(url))) {
+        // Provider not supported.
+        continue;
```
kyliau, line 147: We should probably print out a warning here

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    return data;
+  }
+
+  private async processHrefs(hrefList: string[]): Promise<Map<string, string>> {
+    const hrefsContent = new Map<string, string>();
+
+    for (const href of hrefList) {
+      // Normalize protocols to https://
+      let normalizedHref = href;
```
kyliau, line 133: Consider using the [URL](https://nodejs.org/api/url.html#url_url) class to extract the protocol

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
```diff
@@ -0,0 +1,170 @@
...
+      let cssContent = '';
+      if (this.options.WOFFSupportNeeded) {
+        cssContent += await this.getResponse(normalizedHref, UserAgent.IE);
+      }
+      cssContent += await this.getResponse(normalizedHref, UserAgent.Chrome);
+
+      if (this.options.minifyInlinedCSS) {
+        cssContent = cssContent
+          // Comments and new lines.
+          .replace(/(\n|\/\*\s.+\s\*\/)/g, '')
```
kyliau, line 160: nit: Can we factor out the two regexes and write some tests for them?
(address in a follow-up PR is fine)

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts_spec.ts
```diff
@@ -0,0 +1,81 @@
...
+  const content = `
+  <html>
+    <head>
+      <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+    </head>
+    <body></body>
+  </html>`;
+
+  it('should inline supported fonts and icons in HTML', async () => {
+    const source = new InlineFontsProcessor({
```
kyliau, line 20: `new` here returns a Promise? @_@
Consider adding type annotation to `source`

packages/angular_devkit/build_angular/src/utils/normalize-optimization.ts
```diff
@@ -6,13 +6,28 @@
...
  */
 
-import { OptimizationClass, OptimizationUnion } from '../browser/schema';
+import { FontsClass, OptimizationClass, OptimizationUnion } from '../browser/schema';
+
+export type NormalizeOptimizationOptions = Required<Omit<OptimizationClass, 'fonts'>> & {
+  fonts: FontsClass,
+};
+
+export function normalizeOptimization(optimization: OptimizationUnion = false): NormalizeOptimizationOptions {
```
kyliau, line 15: Not exactly sure what the default value here means - isn't the `default` in the schema `true`?

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
```diff
@@ -97,7 +97,7 @@ export function setup(data: number[] | { cachePath: string; i18n: I18nOptions })
 
 async function cachePut(content: string, key: string | undefined, integrity?: string): Promise<void> {
   if (cachePath && key) {
-    await cacache.put(cachePath, key || null, content, {
+    await cacache.put(cachePath, key || null as unknown as string, content, {
```
kyliau, line 100: why is this cast needed?

kyliau (commented):
some minor comments

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
alan-agius4: You are right :)

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
alan-agius4: Added, though typically we don't add type annotations in this cases since it's inferred.

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
alan-agius4: 👍

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
alan-agius4: I actually left it separate because the new lines is rather simple. But I can split it.

There is a basic test for this here: https://github.com/angular/angular-cli/pull/18926/files#diff-3e96aa638f2afb73f78ff67569b9b185c46a32968faf2c85cf0c5e75987452d8R71-R80

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts_spec.ts
alan-agius4: It's not the new that returns a Promise, but rather the `process` method

```ts
const source = new InlineFontsProcessor({
    WOFFSupportNeeded: false,
    minifyInlinedCSS: true,
}).process(content);
```

Would it be simpler to understand if it was not chained?

```ts

const inlineFontsProcessor = new InlineFontsProcessor({
    WOFFSupportNeeded: false,
    minifyInlinedCSS: true,
});

const source = inlineFontsProcessor.process(content);
```

packages/angular_devkit/build_angular/src/utils/normalize-optimization.ts
alan-agius4: Strictly speaking the default is defined in the schematic which is `false`. But in the typings these are still marked as optional.

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
alan-agius4: Because `null` is not accepted by the types provided via @types.

packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
alan-agius4: Having unsupported providers is a valid use-case. One such example is for css files which are consumed via a CDN. Example: `https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/css/bootstrap.min.css`

packages/angular_devkit/build_angular/src/utils/process-bundle.ts
kyliau: We've already asserted that `key` is truthy on the line above, do we still need `null`?

alan-agius4: Dang you are right.

Andrew77540 (approved):

Andrew77540 (approved):

Andrew77540 (approved):

Andrew77540 (approved):

Andrew77540:
Trying the different

clydin (approved):

PR #18931, opened 2020-09-30
feat(@schematics/angular): don't generate  special `package.json` and `no-any` when using `--strict`

alan-agius4:
Following feedback from the community, which showed a high number of users found it hard to work with the `no-any` lint rule and also the `sideEffects` package.json file. With DevRel we decided to remove both features when generating a strict workspace and/or application.

### DIFF
--- a/packages/schematics/angular/application/schema.json
+++ b/packages/schematics/angular/application/schema.json
@@ -107,7 +107,7 @@
       "x-user-analytics": 15
     },
     "strict": {
-      "description": "Creates an application with stricter build optimization options.",
+      "description": "Creates an application with stricter bundle budgets settings.",
       "type": "boolean",
       "default": false
     },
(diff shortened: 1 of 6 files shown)


clydin (approved):

packages/schematics/angular/application/schema.json
```diff
@@ -107,7 +107,7 @@
       "x-user-analytics": 15
     },
     "strict": {
-      "description": "Creates an application with stricter build optimization options.",
+      "description": "Creates an application with stricter bundle budgets settings.",
```
kyliau, line 110: does this not include stricter type checking?

alan-agius4: The stricter type checking is at workspace level.

kyliau (approved):

PR #18940, opened 2020-09-30
refactor(@angular/cli): discover/load workspace on startup

clydin:
Previously, the workspace configuration file was found and loaded by individual commands potentially multiple times.  This change moves the initial workspace location discovery and loading of the workspace to the CLI startup.  It also provides the workspace to each command so that the commands can reuse the already loaded and parsed workspace configuration.
This change will serve as a base to further refactor and consolidate workspace option usage.


alan-agius4 (approved):
Nice work! 👍

PR #18941, opened 2020-09-30
fix(@angular-devkit/build-webpack): avoid deprecation warning with Webpack 5 watch mode

clydin:
With Webpack 5, passing the watch configuration option into the `webpack` factory function will cause the Webpack compiler to assume it should immediately start watching which requires a callback to be passed as well.  Since the execution of the compiler is handled later in the process, a callback at the compiler creation stage is undesirable and could result in potentially breaking changes to accomplish.


alan-agius4 (approved):
LGTM

PR #18944, opened 2020-10-01
Revert "ci: disable lint check in the update test"

dgp1130:
Re-enabling the lint check to merge once `11.0.0-next.4` is released.

I'll leave this PR in the blocked state until the version is released and https://github.com/angular/angular/pull/39070 is merged. Once those have both happened, it should be safe to merge this as well.

### DIFF
--- a/tests/legacy-cli/e2e/tests/update/update-7.0.ts
+++ b/tests/legacy-cli/e2e/tests/update/update-7.0.ts
@@ -36,8 +36,7 @@ export default async function() {
   // Run CLI commands.
   await ng('generate', 'component', 'my-comp');
   await ng('test', '--watch=false');
-  // TODO: Re-enable after v11.0.0-next.4 release.
-  // await ng('lint');
+  await ng('lint');
   await ng('e2e');
   await ng('e2e', '--prod');
 


dgp1130:
The relevant [blocking PR](angular/angular#39070) was merged, however it was reverted in https://github.com/angular/angular/pull/39082 because it is not compatible with TS 3.9 which is still used in google3. I'm looking for a solution that will be compatible with both versions and we can re-enable this test once that lands.

dgp1130:
I have a fix in https://github.com/angular/angular/pull/39102, where using `ts.createIdentifier()` is able to make a single quoted string literal in both TS 3.9 and TS 4.0. This should be compatible with google3.

dgp1130:
The fix was merged and doesn't appear to have been rolled back yet, so I think it's safe to merge this PR and re-enable the lint check.

dgp1130:
Actually, CI is still failing for this PR itself, because the fix on `angular/angular` has not been released and included in the CLI. I think we'll also need to wait for that, so I'll need to follow up next week on this.

clydin (approved):

dgp1130:
@alan-agius4 pointed out that e2e tests load FW from NPM, so we only need FW to release the migration fix, not the CLI.

This should be unblocked now.

alan-agius4 (approved):

PR #18950, opened 2020-10-01
refactor(@schematics/angular): display Stylus documentation url with HTTPS

bampakoa:
Display the url of the Stylus format in the list of stylesheet format selection with SSL


alan-agius4 (approved):
Thanks

PR #18959, opened 2020-10-02
build: configure dependabot updates

josephperrott:
configure dependabot updates to use the expected commit message type/scope


clydin (approved):

josephperrott:
@clydin can you take another look?

clydin (approved):

PR #18971, opened 2020-10-04
fix(@angular/cli): add blank line in migration commit message

drumonii:
When using the `-C` option of `update`, the commit message of the migration is missing a blank line between the subject and body. 

For example,

Previously was missing blank line after the commit subject:
```
@angular/cli migration - rename-browserslist-config
Update Browserslist configuration file name to '.browserslistrc' from deprecated 'browserslist'.
```

Now fixed:
```
@angular/cli migration - rename-browserslist-config

Update Browserslist configuration file name to '.browserslistrc' from deprecated 'browserslist'.
```


alan-agius4 (approved):
Thanks, LGTM

Issue #17900, opened 2020-06-10
Analytics reports only contains options passed via the command line


alan-agius4:
# 🐞 Bug report

### Is this a regression?
No

### Description
At the moment `reportAnalytics` is called outside of `run` this has a side effect of only sending options that are used in the command line and are non leftovers. Which results in a lot of data being not being sent to Google Analytics.

https://github.com/angular/angular-cli/blob/b630317b4bcef1089b8c1c3221c3407aded990f8/packages/angular/cli/models/command.ts#L183-L184

Reasons:
1) Most of Schematics options are leftovers example:
```
{ dryRun: true, schematic: 'c', '--': [ 'baz', '--style', 'css' ] }
``` 

2) Values defined in `angular.json` or via prompts are ignored and not sent.

3) Option parsing and merging happen during the `run`.

## 🔥 Exception or Error
None

## 🌍 Your Environment
All versions

PR #18982, opened 2020-10-06
fix(@angular/cli): collect analytics option usage from workspace config and prompts

alan-agius4:
With this change we fix two analytics collection bugs:
- We now collect the usage of options defined in the workspace config (angular.json).
- We now also collect values set via schematic prompts.

Closes: #17900

### DIFF
--- a/packages/angular/cli/models/command.ts
+++ b/packages/angular/cli/models/command.ts
@@ -23,7 +23,8 @@ export interface BaseCommandOptions {
 }
 
 export abstract class Command<T extends BaseCommandOptions = BaseCommandOptions> {
-  public allowMissingWorkspace = false;
+  protected allowMissingWorkspace = false;
+  protected useReportAnalytics = true;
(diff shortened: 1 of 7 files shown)


packages/angular/cli/models/command.ts
```diff
@@ -173,7 +173,6 @@ export abstract class Command<T extends BaseCommandOptions = BaseCommandOptions>
       return this.printJsonHelp(options);
     } else {
       const startTime = +new Date();
-      await this.reportAnalytics([this.description.name], options);
```
clydin, line 176: With this change, I think non-architect and non-schematic commands will also need to manually run `reportAnalytics`.

alan-agius4: You are right!

alan-agius4: PTAL

clydin (approved):

PR #18984, opened 2020-10-06
deprecate tslint builder and lintFix options

alan-agius4:


packages/angular_devkit/schematics/tasks/tslint-fix/executor.ts
```diff
@@ -79,7 +79,7 @@ function _listAllFiles(root: string): string[] {
   return result;
 }
 
-
+/** @deprecated Use `ng lint --fix` directly instead. */
```
alan-agius4, line 82: @kyliau can you check if this is used in Google?

packages/angular_devkit/build_angular/src/tslint/index.ts
```diff
@@ -134,6 +139,7 @@ async function _run(
 }
 
 
+/** @deprecated Deprecated as part of the TSLint deprecation. */
```
clydin, line 142: Can you add the version number for the deprecation to the comments as well?

clydin (approved):

packages/angular_devkit/build_angular/src/tslint/index.ts
alan-agius4: Yeah, even in the schema?

packages/angular_devkit/build_angular/src/tslint/index.ts
```diff
@@ -26,6 +26,11 @@ async function _run(
   options: TslintBuilderOptions,
   context: BuilderContext,
 ): Promise<BuilderOutput> {
+  context.logger.warn(
+    'The TSLint builder is deprecated and will be removed in future versions.\n' +
```
mgechev, line 30: Let's share why are we deprecating it.

packages/angular_devkit/build_angular/src/tslint/index.ts
clydin: probably just the code comments.

packages/angular_devkit/schematics/tasks/tslint-fix/executor.ts
kyliau: negative. Google doesn't use this task executor to fix code.

alan-agius4: Copy that.

packages/angular_devkit/build_angular/src/tslint/index.ts
alan-agius4: @mgechev, PTAL.

mgechev (approved):

PR #18991, opened 2020-10-07
ci: allow private packages as commit scopes

alan-agius4:
Example `fix(@angular-devkit/benchmark): bla bla`


.ng-dev/commit-message.ts
```diff
@@ -16,7 +16,5 @@ export const commitMessage: CommitMessageConfig = {
   maxLineLength: Infinity,
   minBodyLength: 0,
   minBodyLengthTypeExcludes: ['docs'],
-  scopes: [
-    ...Object.keys(packages).filter(name => !packages[name].private),
-  ],
+  scopes: Object.keys(packages),
```
clydin, line 19: Looks good but can you leave the array in place.  It will cause less code churn if/when new scopes are added.

alan-agius4: Re added

clydin (approved):

PR #18993, opened 2020-10-07
build: use @angular/dev-infra-private for bazel stamping

clydin:
Some initial preparation for building the packages with bazel. Currently unused.


josephperrott (approved):
LGTM

PR #18998, opened 2020-10-07
build: delete @angular-devkit/build-ng-packagr

kyliau:
BREAKING CHANGE:

Deprecated `@angular-devkit/build-ng-packagr` has been removed. Please use `ng-packagr` builder from `@angular-devkit/build-angular`
instead.

Note: this change doesn't effect application and library developers, as they will be migrated to use the new builder during ng update.


alan-agius4 (approved):
LGTM. Can you please update the commit message to something like:

```
BREAKING CHANGE:

Deprecated `@angular-devkit/build-ng-packagr` has been removed. Please use `ng-packagr` builder from `@angular-devkit/build-angular`
instead.

Note: this change doesn't effect application and library developers, as they will be migrated to use the new builder during ng update.
```

PR #19015, opened 2020-10-08
feat(@angular-devkit/build-angular): support custom headers in dev-server

bjarkler:
Make it possible to configure dev-server to send custom HTTP headers on
every client request. These headers can be specified as a key-value map
under the new "headers" property of the dev-server builder in
angular.json. These headers are then passed on to the webpack devserver.

An example use case for this is to enable various security features,
such as CSP and Trusted Types, both in local application development and
in integration tests, by setting appropriate HTTP headers.

This is part of an effort to add support for Trusted Types in Angular.
The ability to enforce Trusted Types during development and integration
tests is essential, as this can help detect Trusted Types violations
that might otherwise break applications when they're pushed to
production where such security features may be enforced.

This change is completely backwards compatible, and does not change
any behavior before the "headers" property is set.

Related to: https://github.com/angular/angular/pull/39222

### DIFF
--- a/packages/angular_devkit/build_angular/src/dev-server/index.ts
+++ b/packages/angular_devkit/build_angular/src/dev-server/index.ts
@@ -397,10 +397,13 @@ export function buildServerConfig(
   const servePath = buildServePath(serverOptions, browserOptions, logger);
   const { styles, scripts } = normalizeOptimization(browserOptions.optimization);
 
-  const config: WebpackDevServer.Configuration & { logLevel: string } = {
+  const config: WebpackDevServer.Configuration&{logLevel: string} = {
     host: serverOptions.host,
     port: serverOptions.port,
-    headers: { 'Access-Control-Allow-Origin': '*' },
+    headers: {
+      'Access-Control-Allow-Origin': '*',
+      ...serverOptions.headers,
+    },
     historyApiFallback: !!browserOptions.index && {
       index: `${servePath}/${getIndexOutputFile(browserOptions)}`,
       disableDotRule: true,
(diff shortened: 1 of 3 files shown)


bjarkler:
CC @engelsdamien @koto

packages/angular_devkit/build_angular/src/dev-server/schema.json
```diff
@@ -189,6 +189,16 @@
     "poll": {
       "type": "number",
       "description": "Enable and define the file watching poll time period in milliseconds."
+    },
+    "headers": {
```
IgorMinar, line 193: can you please move this section to the top, maybe under `sslCert` - that way all the serving configuration is together. thanks!

packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -397,10 +397,19 @@ export function buildServerConfig(
...
 
-  const config: WebpackDevServer.Configuration & { logLevel: string } = {
+  const headers: {[key: string]: string} = {
+    'Access-Control-Allow-Origin': '*',
+  };
+  if (serverOptions.headers) {
+    for (const [key, value] of Object.entries(serverOptions.headers)) {
+      headers[key] = value;
+    }
+  }
```
IgorMinar, line 407:
```suggestion
  const headers: {[key: string]: string} = {
    'Access-Control-Allow-Origin': '*',
    ...serverOptions.headers
  };
```

IgorMinar (approved):
LGTM with some nits. @bjarkler can you please clean this PR up? thanks!

@clydin can you please take a quick look as well? thanks

packages/angular_devkit/build_angular/src/dev-server/schema.json
bjarkler: Done.

packages/angular_devkit/build_angular/src/dev-server/index.ts
bjarkler: Neat! Applied this directly in the config map.

clydin (approved):

PR #19017, opened 2020-10-08
feat(@angular-devkit/build-optimizer): introduce ivy optimization mode

clydin:
This mode allows disabling of certain passes that are not needed by Ivy compilation but are still required by View Engine.  Currently only the prefix functions pass is affected by this mode.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -414,18 +414,39 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
         allowMinify &&
         (buildOptions.platform == 'server'
           ? {
-            ecma: terserEcma,
-            global_defs: angularGlobalDefinitions,
-            keep_fnames: true,
-          }
+              ecma: terserEcma,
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -411,18 +414,27 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
-            // See https://github.com/webpack/webpack/issues/2899#issuecomment-317425926.
-            passes: buildOptions.buildOptimizer ? 3 : 1,
-            global_defs: angularGlobalDefinitions,
-          }),
+              ecma: terserEcma,
+              pure_getters: buildOptions.buildOptimizer,
+              // PURE comments work best with 3 passes.
+              // See https://github.com/webpack/webpack/issues/2899#issuecomment-317425926.
+              passes: buildOptions.buildOptimizer ? 3 : 1,
+              global_defs: angularGlobalDefinitions,
```
alan-agius4, line 427: Thinking about it, shouldn't `global_defs` and `pure_funcs` be configured when platform is server as well?

Issue #18692, opened 2020-09-04
ng new logs a warning about entryComponents being deprecated


cexbrayat:
# 🐞 Bug report

### Command (mark with an `x`)

- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the previous version in which this bug was not present was: v10.0.x

### Description

When running `ng new`, a warning about `entryComponents` is logged

## 🔥 Exception or Error
<pre><code>
> ng new entry-comp-warning --defaults
Option "entryComponent" is deprecated: Since version 9.0.0 with Ivy, entryComponents is no longer necessary.
CREATE entry-comp-warning/README.md (1034 bytes)
CREATE entry-comp-warning/.editorconfig (274 bytes)
CREATE entry-comp-warning/.gitignore (631 bytes)
CREATE entry-comp-warning/angular.json (3662 bytes)
CREATE entry-comp-warning/package.json (1261 bytes)
...
</code></pre>

## 🌍 Your Environment
<pre><code>
Angular CLI: 10.1.0
Node: 14.6.0
OS: darwin x64

Angular:
...
Ivy Workspace:

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1001.0
@angular-devkit/core         10.1.0
@angular-devkit/schematics   10.1.0
@schematics/angular          10.1.0
@schematics/update           0.1001.0
rxjs                         6.6.2
</code></pre>

**Anything else relevant?**

`npx @angular/cli@10.0.8 new entry-comp-warning --defaults` does not raise this warning.

alan-agius4:
Ummm, this is an interesting one!

prabh-62:
`> npx @angular/cli@11.0.0-next.6 new admin-console --strict true --routing true --style css`
outputs zero warnings

`> npx @angular/cli@11.0.0-next.6 new admin-console --strict true --defaults true`
spawns few warnings

<img width="1342" alt="ng_new_11" src="https://user-images.githubusercontent.com/13006107/95597104-389b5c80-0a1c-11eb-9071-4d2a320ab556.png">

PR #19028, opened 2020-10-09
fix(@angular/cli): don't show deprecation messages for defaults when using `--defaults`

alan-agius4:
With this change we use the addUndefinedDefaults transformer post validation and don't create prompts which have a default value when options.defaults is truthy.

Closes: #18692

### DIFF
--- a/packages/angular/cli/models/schematic-command.ts
+++ b/packages/angular/cli/models/schematic-command.ts
@@ -304,53 +304,51 @@ export abstract class SchematicCommand<
       ...(await getSchematicDefaults(schematic.collection.name, schematic.name, getProjectName())),
       ...current,
     });
+
     workflow.engineHost.registerOptionsTransform(defaultOptionTransform);
 
-    if (options.defaults) {
-      workflow.registry.addPreTransform(schema.transforms.addUndefinedDefaults);
-    } else {
-      workflow.registry.addPostTransform(schema.transforms.addUndefinedDefaults);
-    }
-
+    workflow.registry.addPostTransform(schema.transforms.addUndefinedDefaults);
     workflow.registry.addSmartDefaultProvider('projectName', getProjectName);
     workflow.registry.useXDeprecatedProvider(msg => this.logger.warn(msg));
 
     if (options.interactive !== false && isTTY()) {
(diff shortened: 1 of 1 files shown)


packages/angular/cli/models/schematic-command.ts
```diff
@@ -304,20 +304,18 @@ export abstract class SchematicCommand<
...
 
+    workflow.registry.addPreTransform(schema.transforms.addUndefinedDefaults);
     workflow.registry.addSmartDefaultProvider('projectName', getProjectName);
     workflow.registry.useXDeprecatedProvider(msg => this.logger.warn(msg));
 
     if (options.interactive !== false && isTTY()) {
       workflow.registry.usePromptProvider((definitions: Array<schema.PromptDefinition>) => {
-        const questions: inquirer.QuestionCollection = definitions.map(definition => {
+        const questions: inquirer.QuestionCollection = definitions
+        .filter(definition => !options.defaults || definition.default == null)
```
clydin, line 317:
```suggestion
        .filter(definition => !options.defaults || definition.default == undefined)
```
I think this should be `undefined` since `null` is a valid default value.

packages/angular/cli/models/schematic-command.ts
```diff
@@ -304,20 +304,18 @@ export abstract class SchematicCommand<
...
-    workflow.engineHost.registerOptionsTransform(defaultOptionTransform);
 
-    if (options.defaults) {
-      workflow.registry.addPreTransform(schema.transforms.addUndefinedDefaults);
-    } else {
-      workflow.registry.addPostTransform(schema.transforms.addUndefinedDefaults);
-    }
+    workflow.engineHost.registerOptionsTransform(defaultOptionTransform);
 
+    workflow.registry.addPreTransform(schema.transforms.addUndefinedDefaults);
```
clydin, line 310: Should this be a post transform?

clydin (changes requested):

packages/angular/cli/models/schematic-command.ts
alan-agius4: Ah yeah you are right

clydin (approved):

PR #19047, opened 2020-10-11
fix: @angular/cli web worker schematic

guiseek:
web worker schematics are missing base suffix in tsconfig


alan-agius4:
Hi @guiseek,

Thanks for your contribution but `tsconfig.base.json` is no longer generated since Angular CLI 10.1. It is also removed from existing projects when updating using `ng update`.

guiseek:
> Hi @guiseek,
> 
> Thanks for your contribution but `tsconfig.base.json` is no longer generated since Angular CLI 10.1. It is also removed from existing projects when updating using `ng update`.

hmmm, i understood the reason for the confusion, some libs search with a base suffix, if not, errors appear, and if it is suffixed, some search without and it also generates an error. shouldn't that be managed by the typescript team?

but okay, thanks! if i have anything else that i can contribute, i will be back.
thanks
[]s

Issue #19058, opened 2020-10-09
Wrong order of style-tags in head-tag in a multi-app repo.


infacto:
# 🐞 bug report

It's a blocker. Please handle with higher priority.

### Affected Package
I think it's about the compiler or any other related package which handles the style-tag order.

### Description

I have to Angular project. A default build and a multi-project one.
Both has the same setup. In this case it uses Ionic. But this does not matter. I use the Angular CLI for builds.
The `angular.json` is the same. Execpt of the path of course. (And some missing e2e and unit test. Nothing related.)

Both repos works great. But on runtime, I found some weird issues. The order of style-tags are different in html-head.
The order of the multi-app one is not correct. Please read on ...

## 🔬 Minimal Reproduction

A Stackblitz example is not possible in this case. Please download the zip repos. (click on the link in the header of the table.)

**Style-Tags**
The following is a table with the order of all style-tags and a short code snippet inside. The style-tags itself have no id. So, only the code snippet can identify it. (remeber: click on the header link to download the related repo.)

| [Working Demo](https://github.com/ionic-team/ionic-framework/files/5345465/issue-demo-2.zip)<br/><small>(Click to download)</small>                                 | [Problem Demo](https://github.com/ionic-team/ionic-framework/files/5339318/issue-demo.zip)<br/><small>(Click to download)</small>                                 |
| -------------------------------------------- | -------------------------------------------- |
| **1** (data-styles) `ion-select-popover` ... | **1** (data-styles) `ion-select-popover` ... |
| **2** `html.plt-mobile ion-app` ...          | **7** `:root { --ion-color-primary` ...      |
| **3** `ion-header` ...                       | **8** `// App Global CSS` ... `html.ios` ... |
| **4** `ion-list` ...                         | **2** `html.plt-mobile ion-app` ...          |
| **5** `ion-item-group` ...                   | **3** `ion-header` ...                       |
| **6** `.sc-ion-buttons-md-h` ...             | **6** `.sc-ion-buttons-md-h` ...             |
| **7** `:root { --ion-color-primary` ...      | **4** `ion-list` ...                         |
| **8** `// App Global CSS` ... `html.ios` ... | **5** `ion-item-group` ...                   |

(In the working demo, you may see other versions of the dependency. Sorry for that. But it does not matter. You can update it if you want. It will not change / fix the problem.)

## 🔥 Exception or Error
[...]

PR #19071, opened 2020-10-14
fix(@angular/cli): resolve packages package.json from workspace directory

alan-agius4:
Unlike `require.resolve`, the `resolve` package `paths` is only used as a fallback when the package is not resolved from the `basedir`,

Previously this resulted in the temporary version of CLI being resolved here which resulted in `ng update` incorrectly stating that there are no updates.


alan-agius4:
Mentioned in https://github.com/angular/angular-cli/issues/19058#issuecomment-708197489

clydin (approved):

clydin (approved):

PR #19077, opened 2020-10-15
Support running the E2E test suite with yarn

clydin:
With this change, the E2E test suite can be run using yarn as the package manager instead of npm by using the `--yarn` command line option.


alan-agius4 (approved):
Nice 👍

PR #19088, opened 2020-10-15
refactor(@angular-devkit/build-angular): move dev-server webpack config in a separate file

alan-agius4:


packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -89,12 +90,12 @@ export function serveWebpackBrowser(
   async function setup(): Promise<{
     browserOptions: json.JsonObject & BrowserBuilderSchema;
     webpackConfig: webpack.Configuration;
-    webpackDevServerConfig: WebpackDevServer.Configuration;
     projectRoot: string;
     locale: string | undefined;
   }> {
     // Get the browser configuration from the target name.
     const rawBrowserOptions = await context.getTargetOptions(browserTarget);
+    options.port = await checkPort(options.port || 4200, options.host || 'localhost');
```
clydin, line 98: `0` is a valid value for port but is also falsy so `||` can't be used.

packages/angular_devkit/build_angular/src/dev-server/ssl_spec.ts
```diff
@@ -102,8 +102,8 @@ describe('Dev Server Builder ssl', () => {
 
     const overrides = {
       ssl: true,
-      sslKey: '../ssl/server.key',
-      sslCert: '../ssl/server.crt',
+      sslKey: 'ssl/server.key',
+      sslCert: 'ssl/server.crt',
```
clydin, line 106: Did this test not work before?

clydin (changes requested):

packages/angular_devkit/build_angular/src/dev-server/ssl_spec.ts
alan-agius4: Yeah, wrong paths

packages/angular_devkit/build_angular/src/dev-server/index.ts
alan-agius4: That is correct, though that is the current behaviour. I’ll update to use `??`

packages/angular_devkit/build_angular/src/dev-server/ssl_spec.ts
alan-agius4: Perviously, wrong paths errors where swallowed.

clydin (approved):

Issue #19116, opened 2020-10-19
[router/testing] `No base href set` errors in tests appear with router when using Debug Mode


IxquitilisSaid:
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

No. Seems to me that it's a pretty old issue.

### Description

In short, when testing an app with routing, tests will run fine unless you toggle Karma's debug mode. There all components with routing will throw the aforementioned "No base href set" error.

## 🔬 Minimal Reproduction
I've made a demo repo for it [here](https://github.com/IxquitilisSaid/karma-debug-routing-error)

```
npm i
ng test
```

Initially, all the tests will be green and succeed but if you click the `Debug` button above, you'll see the error being thrown

## 🔥 Exception or Error
<pre><code>
Error: No base href set. Please provide a value for the APP_BASE_HREF token or add a base element to the document.
    at new PathLocationStrategy (http://localhost:9876/_karma_webpack_/node_modules/@angular/common/__ivy_ngcc__/fesm2015/common.js:378:1)
    at Object.provideLocationStrategy [as useFactory] (http://localhost:9876/_karma_webpack_/node_modules/@angular/router/__ivy_ngcc__/fesm2015/router.js:5520:1)
    at Object.factory (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js:11338:1)
    at R3Injector.hydrate (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js:11249:1)
    at R3Injector.get (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js:11071:1)
    at injectInjectorOnly (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js:801:1)
    at Module.ɵɵinject (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js:805:1)
[...]

PR #19119, opened 2020-10-19
fix(@angular-devkit/build-angular): add a base href to karma debug context

IxquitilisSaid:
This commits adds a base href value in the karma context iframe used to run unit tests.

It solves #19116  where a unit test throws:

```
No base href set. Please provide a value for the APP_BASE_HREF token or add a base element to the document.
```

Even if the application is fine. This is because the index.html from Angular CLI contains a base href value, but not the Karma context iframe. So when adding a unit test with a testing module that imports a NgModule, for example AppModule, which itself imports RouterModule, the unit test used to throw an error (regression appeared in router 3.1).

That could be solved by either adding `RouterTestingModule` to the testing module, or by adding a provider `{ provide: APP_BASE_HREF, useValue: '/' }`, but required to understand the issue.

This solves the issue in a transparent way: developers won't even encounter the problem anymore.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/karma-debug.html
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/karma-debug.html
@@ -9,6 +9,7 @@ just for immediate execution, without reporting to Karma server.
 <head>
 %X_UA_COMPATIBLE%
   <title>Karma DEBUG RUNNER</title>
+  <base href="/">
   <link href="favicon.ico" rel="icon" type="image/x-icon" />
   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):
Thanks, for this @IxquitilisSaid, LGTM.

One small NIT, can you please add the PR description as part of the commit message body and add `Closes #19116` as the commit message footer.

See: https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#commit for more info.

IxquitilisSaid:
@alan-agius4 I've updated the commit message, hopefully didn't mess it up. 
Let me know if it's still wrong and I'll fix it again 😄

alan-agius4 (approved):
The commit message, looks good.

Thanks

alan-agius4:
@googlebot ping!

PR #19132, opened 2020-10-20
refactor(@angular/cli): rename `ng xi18n` to `ng extract-i18n`

alan-agius4:
`ng i18n-extract` and `ng xi18n` has been deprecated in favor of `ng extract-i18n` to have a better intuitive naming and match the architect key in `angular.json`.

### DIFF
--- a/packages/angular/cli/commands/extract-i18n-impl.ts
+++ b/packages/angular/cli/commands/extract-i18n-impl.ts
@@ -8,12 +8,12 @@
 
 import { ArchitectCommand } from '../models/architect-command';
 import { Arguments } from '../models/interface';
-import { Schema as Xi18nCommandSchema } from './xi18n';
+import { Schema as ExtractI18nCommandSchema } from './extract-i18n';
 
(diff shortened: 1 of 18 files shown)


packages/angular/cli/commands/extract-i18n-impl.ts
```diff
@@ -23,6 +23,11 @@ export class Xi18nCommand extends ArchitectCommand<Xi18nCommandSchema> {
       return 1;
     }
 
+    const commandName = process.argv[2];
+    if (['xi18n', 'i18n-extract'].includes(commandName)) {
+      this.logger.warn(`Warning: "ng ${commandName}" has been deprecated and will be removed in future major version. Please use "ng extract-i18n" instead.`);
```
clydin, line 28:
```suggestion
      this.logger.warn(`Warning: "ng ${commandName}" has been deprecated and will be removed in a future major version. Please use "ng extract-i18n" instead.`);
```

clydin (approved):

PR #19136, opened 2020-10-20
ci: schedule nightly CI for 10.1.x branch

alan-agius4:


.circleci/config.yml
```diff
@@ -420,3 +419,37 @@ workflows:
...
+            branches:
+              only:
+                - 10.1.x
+    jobs:
+      # Linux jobs
+      - setup
+      - build:
+          requires:
+            - setup
+          filters:
```
clydin, line 437: This filter can be dropped (completely actually).  The `docs-preview` branch doesn't exist anymore.

.circleci/config.yml
```diff
@@ -420,3 +419,37 @@ workflows:
...
+            branches:
+              ignore:
+              - /docs-preview/
+      - e2e-cli:
+          post-steps:
+            - store_artifacts:
```
clydin, line 443: We can probably drop this as well for nightly runs since the artifacts should be the same each time.

.circleci/config.yml
```diff
@@ -420,3 +419,37 @@ workflows:
...
+                path: /tmp/dist
+                destination: cli/new-production
+          requires:
+            - build
+      - e2e-cli:
+          name: e2e-cli-ve
+          ve: true
```
clydin, line 450: Should we do an ivy run as well?

alan-agius4: Yeah, it's the step on top of this one.

clydin (approved):

PR #19144, opened 2020-10-21
fix(@angular/cli): coerce prompt answers to requested property types

clydin:
This change converts answers from prompts into the property type requested from the schema.  This allows properties that expect a number to correctly validate when an input prompt is used.

### DIFF
--- a/packages/angular/cli/models/schematic-command.ts
+++ b/packages/angular/cli/models/schematic-command.ts
@@ -322,6 +322,32 @@ export abstract class SchematicCommand<
             const validator = definition.validator;
             if (validator) {
               question.validate = input => validator(input);
+
+              // Filter allows transformation of the value prior to validation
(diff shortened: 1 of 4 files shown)


packages/angular/cli/models/schematic-command.ts
```diff
@@ -322,6 +322,35 @@ export abstract class SchematicCommand<
...
+                  let value;
+                  switch (type) {
+                    case 'string':
+                      value = String(input);
+                      break;
+                    case 'integer':
+                    case 'number':
+                      value = Number(input);
+                      break;
+                    case 'boolean':
```
alan-agius4, line 338: What is the use case of the boolean coercion here? `false` will be coerced to `true` in this case.

```
Boolean('false') // true
```

alan-agius4 (approved):

PR #19177, opened 2020-10-22
feat(@schematics/angular): Schematic to migrate invalid interpolations

ayazhafiz:
This commit adds a migration schematic to fix interpolation-like markup
now invalidated by #39107. Previously, the angular parser would not
match markup like `{{ 1 + 2 }` or `{{ 1 + 2 }<!-- -->}` to be partial
(but malformed) interpolations, an implementation detail that some
developers may have relied on to treat interpolation-like markup as
text. #39107 changes the parser implementation to know emit errors in
such cases, because the markup looks like an interpolation but is not
well-formed. In such cases it is preferred to use an interpolation
directly or escape interpolation delimiters (e.g. `{{ '{{' }} 1 + 2
{{ '}}' }}`). This commit provides fixes for malformed cases so as to
ease migration for applications relying on the previous implementation.

### DIFF
--- a/packages/schematics/angular/migrations/update-11/fix-invalid-interpolations.ts
+++ b/packages/schematics/angular/migrations/update-11/fix-invalid-interpolations.ts
@@ -1,0 +1,85 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {extname, Path} from '@angular-devkit/core';
+import {Rule} from '@angular-devkit/schematics';
+
+const RE_INTERPOLATIONS: ReadonlyArray<[RegExp, string]> = [
+  // Matching an interpolation:
+  //
+  // {{((?:[^}'"]|'[^']*?'|"[^"]*?")*?)}
+  //        ^^^^^^                    match everything except a }, ", or '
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/migrations/update-11/fix-invalid-interpolations.ts
```diff
@@ -0,0 +1,85 @@
...
+  // The former is a little-known pattern previously used to display an
+  // iterpolation literally, but that does not longer parse. The latter would
+  // still parse.
+  [
+    /{{((?:'[^']*?'|"[^"]*?"|[^}'"])*?)}<!--.*-->}/g,
+    //                                 ^^^^^^^^^^^ match a comment b/w braces
+    `{{ '{{' }}$1{{ '}}' }}`
+  ],
+
+  // Replace "{{expr}" with "{{expr}}".
```
ayazhafiz, line 29: Hm, maybe we should do `{{expr} -> {{ '{{' }}expr{{ '}' }}` to preserve behavior

alan-agius4:
@ayazhafiz, this looks like it should be a migration in `@angular/core` and not `@angular/cli`

ayazhafiz:
That makes sense, thanks!

clydin:
Something else to consider is that these cases could be legitimate application errors that were never caught.  An alternative could be to allow the user to selectively fix the errors rather than to propagate the errors in a new form.  Maybe a language service quick fix that performs the conversion upon user request per instance?

PR #19217, opened 2020-10-26
fix(@schematics/update): update fail when using yarn 2.0 protocols

### LINKED ISSUE #19203: [Yarn2] Migrations fail when using custom protocols in `package.json`
# 🐞 Bug report

- [x] update

### Is this a regression?

No

### Description

Yarn 2 allows using custom protocols for dependencies. A current list is available here:
https://yarnpkg.com/features/protocols

It seems that when trying to use `ng update` for migrations, the process fails because it thinks it finds a malformed specifier.

I believe this should not be a hard error, but perhaps just a warning. It is safe to ignore these dependencies.

Note that this error occurs with the currently compatible Yarn2/Angular mode of using the standard `node_modules` linker. See: https://github.com/angular/angular-cli/issues/16980

Relevant section of code:

https://github.com/angular/angular-cli/blob/e22c196c32278dee78f56088613de51bbf826131/packages/schematics/update/update/index.ts#L869-L871

## 🔬 Minimal Reproduction

```
cd /tmp
ng new --skip-install ngupdate-repro
cd ngupdate-repro

yarn set version 2
yarn config set nodeLinker node-modules
yarn install

yarn add @angular/cli@latest
yarn add @angular/material@8.2.3

yarn add src@link:./src


yarn ng update @angular/material --allow-dirty
```

## 🔥 Exception or Error
```
Repository is not clean. Update changes will be mixed with pre-existing changes.
Using package manager: 'yarn'
Found 31 dependencies.
✖ Migration failed: Failed to parse dependency "src" with specifier "link:./src" from package.json. Is the specifier malformed?
  See "/tmp/ng-SLmx6n/angular-errors.log" for further details.
```

## 🌍 Your Environment
<pre><code>
ngular CLI: 10.2.0
Node: 13.7.0
OS: linux x64

Angular: 10.1.6
Ivy Workspace: Yes

@angular-devkit/architect       0.1001.7
@angular-devkit/build-angular   0.1001.7
@angular-devkit/core            10.2.0
@angular-devkit/schematics      10.2.0
@angular/cli                    10.2.0
@angular/material               8.2.3
@schematics/angular             10.2.0
@schematics/update              0.1002.0
rxjs                            6.6.3
typescript                      4.0.3
</code></pre>

alan-agius4:
Closes #19203

### DIFF
--- a/packages/schematics/update/update/index.ts
+++ b/packages/schematics/update/update/index.ts
@@ -845,10 +845,9 @@ export default function(options: UpdateSchema): Rule {
       try {
         return isPkgFromRegistry(name, specifier);
       } catch {
-        // Abort on failure because package.json is malformed.
-        throw new SchematicsException(
-          `Failed to parse dependency "${name}" with specifier "${specifier}"`
-              + ` from package.json. Is the specifier malformed?`);
+        logger.warn(`Package ${name} was not found on the registry. Skipping.`);
+
+        return false;
       }
     }));
     const packages = _buildPackageList(options, npmDeps, logger);
--- a/packages/schematics/update/update/index_spec.ts
+++ b/packages/schematics/update/update/index_spec.ts
@@ -72,6 +72,24 @@ describe('@schematics/update', () => {
     ).toPromise().then(done, done.fail);
   }, 45000);
 
+  it('should not error with yarn 2.0 protocols', async () => {
+    const tree = new UnitTestTree(new HostTree(new virtualFs.test.TestHost({
+      '/package.json': `{
+        "name": "blah",
+        "dependencies": {
+          "src": "src@link:./src",
+          "@angular-devkit-tests/update-base": "1.0.0"
+        }
+      }`,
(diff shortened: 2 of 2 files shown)


packages/schematics/update/update/index.ts
```diff
@@ -865,10 +865,7 @@ export default function(options: UpdateSchema): Rule {
       try {
         return isPkgFromRegistry(name, specifier);
       } catch {
-        // Abort on failure because package.json is malformed.
-        throw new SchematicsException(
-          `Failed to parse dependency "${name}" with specifier "${specifier}"`
-              + ` from package.json. Is the specifier malformed?`);
+        return false;
```
clydin, line 868: Could we add a warning here so that the user knows that the dependency will not be used during the version analysis?

alan-agius4: Added.

clydin (approved):

PR #19227, opened 2020-10-27
fix(@angular-devkit/build-angular): when optimizing don't wrap function arguments in parenthesis

alan-agius4:
With this change function and arrow function arguments are not wrapped in parenthesis during the optimization phase. 

`wrap_func_args` which is enabled by default in terser will wrap function arguments in parenthesis. Recently this was also changed to wrap lamdas as well:
https://github.com/terser/terser/commit/66c3a5ce66c7873bded8c6466e72027f9816c548

An increase in bundle size was observed without this change. See: https://github.com/angular/angular/pull/39432#discussion_r512345752

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/utils/process-bundle.ts
@@ -400,6 +400,7 @@ async function terserMangle(
       ascii_only: true,
       webkit: true,
       beautify: shouldBeautify,
+      wrap_func_args: false,
     },
     sourceMap:
       !!options.map &&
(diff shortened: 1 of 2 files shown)


clydin (approved):

alfaproject:
But if this sacrifices performance, shouldn't we accept the bigger bundle size? Or is the performance sacrifice negligible? j/w

alan-agius4:
@alfaproject  from https://github.com/angular/angular/pull/39432#discussion_r512556210
> Based on the following article, the “wrap_func_args” should probably be disabled either way. The option makes files slightly larger in the attempt to improve runtime performance by trying to trigger special cases in the parser. However, this apparently is not recommended in newer versions of V8 and can lead to worse performance in some situations.
> 
> Article: https://v8.dev/blog/preparser#pife

PR #19232, opened 2020-10-27
feat(@schematics/angular): add `enableI18nLegacyMessageIdFormat` defa…

petebacondarwin:
…ult to new projects

The ViewEngine message extraction generates a variety of legacy formats
for extracted message ids. These formats have a number of issues related
to whitespace handling and reliance upon information inside the original
HTML of a template. The new message format is more resilient to things
like whitespace changes, and can be generated directly from calls to
`$localize`, which allows messages in application code to have the same
id as identical messages in templates.

As a first step in migrating projects away from the legacy id format
for i18n messages, this commit updates newly generated projects to turn
off the legacy ids.

In the future the default will be flipped and this can be removed.
Eventually the legacy message id support will be removed altogether,
probably in sync with removal of ViewEngine.

### DIFF
--- a/packages/schematics/angular/workspace/files/tsconfig.json.template
+++ b/packages/schematics/angular/workspace/files/tsconfig.json.template
@@ -20,10 +20,11 @@
       "es2018",
       "dom"
     ]
-  }<% if (strict) { %>,
+  },
   "angularCompilerOptions": {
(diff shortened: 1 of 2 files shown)


petebacondarwin:
Related docs fix here https://github.com/angular/angular/pull/39453

kyliau (approved):

alan-agius4 (approved):
Thanks for this 👍

PR #19255, opened 2020-10-29
Bundle output stats fixes

alan-agius4:
See each commit.

These changes are following feedback from twitter and GDE slack channel.

Screenshot 2020-10-29 at 17 19 02

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
@@ -56,7 +56,7 @@ export function generateBundleStats(
   }
 }
 
-function generateBuildStatsTable(data: BundleStats[], colors: boolean): string {
+function generateBuildStatsTable(data: BundleStats[], colors: boolean, showTotalSize: boolean): string {
   const g = (x: string) => colors ? ansiColors.greenBright(x) : x;
   const c = (x: string) => colors ? ansiColors.cyanBright(x) : x;
   const bold = (x: string) => colors ? ansiColors.bold(x) : x;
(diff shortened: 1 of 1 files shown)


clydin (approved):

packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
```diff
@@ -21,11 +21,13 @@ export function formatSize(size: number): string {
 
   const abbreviations = ['bytes', 'kB', 'MB', 'GB'];
   const index = Math.floor(Math.log(size) / Math.log(1024));
-
-  return `${+(size / Math.pow(1024, index)).toPrecision(3)} ${abbreviations[index]}`;
+  const roundedSize = size / Math.pow(1024, index);
+  // bytes don't have a fraction
+  const fractionDigits = (roundedSize % 1) + index === 0 ? 0 : 2;
```
JoostK, line 26: The modulo seems to be a mistake, as it's always 0?

```suggestion
  const fractionDigits = index === 0 ? 0 : 2;
```

packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
```diff
@@ -124,7 +147,20 @@ function statsToString(json: any, statsConfig: any, bundleState?: BundleStats[])
...
+    if (a.stats[2] > b.stats[2]) {
+      return -1;
+    }
+
+    if (a.stats[2] < b.stats[2]) {
+      return 1;
+    }
+
+    return 0;
+  });
```
JoostK, line 161: nit: this could be a simple subtraction:

```suggestion
  const changedChunksStatsDesc = changedChunksStats.sort((a, b) => b.stats[2] - a.stats[2]);
```

Also, `sort` operates in-place so does in fact modify `changedChunksStats`, which makes `changedChunksStatsDesc` identical to `changedChunksStats`. I'd remove `changedChunksStatsDesc` to avoid confusion, or explicitly copy `changedChunksStats` before sorting.

packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
alan-agius4: The roundedSize in this case can contain decimal places, because the rounded is rounded to the nearest size.

Example
```ts
const size = 3000;
const roundedSize  = size / Math.pow(1024, 1); // 1 here is the index of KB
 // => 2.9296875

```

packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
alan-agius4: Good catch on the mutation operation.

The size here can also be a string which will be `NaN`, which actually makes the above the simplest IMHO.

packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
JoostK: Oh, that is trippy! But if `index === 0` then we must be dealing with bytes already, which are known to be integral?

packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
JoostK: Ah, I did not consider the `NaN` case, fair enough :-)

packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
alan-agius4: Ohh indeed

PR #19258, opened 2020-10-29
fix(@angular-devkit/build-angular): re-enable webpack 5 license extraction support

clydin:
With updates to the `license-webpack-plugin` and adjustments to the web worker plugin configuration, license extraction can now be used with webpack 5.  This change also removes the need to filter out the duplicate asset warning on Webpack 4 that was previously being generated.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/browser.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/browser.ts
@@ -47,6 +47,7 @@ export function getBrowserConfig(wco: WebpackConfigOptions): webpack.Configurati
       },
       perChunkOutput: false,
       outputFilename: '3rdpartylicenses.txt',
+      skipChildCompilers: true,
     }));
   }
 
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/browser.ts
```diff
@@ -47,6 +47,7 @@ export function getBrowserConfig(wco: WebpackConfigOptions): webpack.Configurati
       },
       perChunkOutput: false,
       outputFilename: '3rdpartylicenses.txt',
+      skipChildCompilers: true,
```
alan-agius4, line 50: This option might cause licenses not to be extracted
https://github.com/xz64/license-webpack-plugin/issues/83

clydin: The license plugin was also added directly to the web worker compilation (the other location where code is compiled) to allow capture of licenses.  If `perChunkOutput` is set to false in combination with child compilations (including internal ones such as in the CSS extraction), the plugin becomes incompatible with Webpack 5 since the duplicate asset warning in Webpack 4 is now an error.

Alternatively, we could only change the outputFilename to `3rdpartylicenses-[hash].txt` which will generate a unique file per compilation including each child compilation.  Downside is that potentially thousands of these will be created (one per component stylesheet) all with `css-loader` as a dependency even though they are extracted to CSS and have no code.

alan-agius4: Let’s keep it like this for now. We can iterate in the future if needs be.

alan-agius4 (approved):

PR #19260, opened 2020-10-29
fix(@angular-devkit/build-optimizer): mark rxjs add imports as having side effects

clydin:
This change prevents the build optimizer from removing the operator add imports from the rxjs package (for example, `import 'rxjs/add/operator/filter';`).  The entire rxjs package is currently marked as side effect free from within the rxjs `package.json` but the files in the add directory intentionally contain side effects.

### DIFF
--- a/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
+++ b/packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
@@ -54,6 +54,11 @@ function isKnownCoreFile(filePath: string) {
 }
 
 function isKnownSideEffectFree(filePath: string) {
+  // rxjs add imports contain intentional side effects
+  if (/[\\/]node_modules[\\/]rxjs[\\/]add[\\/]/.test(filePath)) {
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_optimizer/src/build-optimizer/build-optimizer.ts
```diff
@@ -54,6 +54,11 @@ function isKnownCoreFile(filePath: string) {
 }
 
 function isKnownSideEffectFree(filePath: string) {
+  // rxjs add imports contain intentional side effects
+  if (/[\\/]node_modules[\\/]rxjs[\\/]add[\\/]/.test(filePath)) {
```
alan-agius4, line 58: Wasn't this pattern used in RxJs 5, which is no longer supported.

clydin: Still usable in 6.x and worked until the recent fix for the size regression.
Once 7 is the minimum (and released) this logic here and in the webpack configuration can be removed.

alan-agius4 (approved):

PR #19274, opened 2020-11-02
refactor(@angular-devkit/build-angular): use custom babel preset to configure babel-loader

clydin:
A custom babel preset is introduced to centralize the configuration of babel within the package.  Application related presets and plugins are now encapsulated within the custom preset which is used via the Webpack babel-loader.  This new custom preset will also provide the integration point for the upcoming Angular linker.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -551,26 +551,10 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
                       sourceType: 'unambiguous',
                       presets: [
                         [
-                          require.resolve('@babel/preset-env'),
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):
Nice work!

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -551,26 +551,10 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
                       sourceType: 'unambiguous',
                       presets: [
                         [
-                          require.resolve('@babel/preset-env'),
+                          require.resolve('../../babel/presets/application'),
```
alan-agius4, line 554: Maybe export a const which has the value of `__filename` instead?

PR #19189, opened 2020-10-23
fix(@angular-devkit/build-angular): improve builder phase reporting

alan-agius4:
Oct-23-2020 15-20-47

image

PR #19285, opened 2020-11-02
Patch #19189

alan-agius4:
Patch #19189


alan-agius4:
Note: sass-node failure will be fixed by https://github.com/angular/angular-cli/pull/19282

filipesilva (approved):

PR #19293, opened 2020-11-03
refactor(@schematics/angular): use jsonc-parser instead of devkit parser

alan-agius4:


packages/angular/cli/commands/version-impl.ts
```diff
@@ -5,11 +5,10 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { JsonParseMode, isJsonObject, parseJson } from '@angular-devkit/core';
-import * as fs from 'fs';
 import * as path from 'path';
 import { Command } from '../models/command';
 import { colors } from '../utilities/color';
+import { JSONFile, parseJson } from '../utilities/json-file';
```
clydin, line 11:
```suggestion
import { JSONFile } from '../utilities/json-file';
```
`parseJson` doesn't appear to be used.

clydin (approved):

PR #19303, opened 2020-11-04
fix(@schematics/angular): library generation on a solution-styled project

Sherpard:
Solution Styled projects depends of `tsconfig.base.json`,  and when the new library paths has to be added, an unhandled exception is thrown (compilerOptions is not part of `tsconfig.json`)

this PR tries first to update `tsconfig.base.json`, and if it doesn't exists, fallback to the `tsconfig.json`.

### DIFF
--- a/packages/schematics/angular/library/index.ts
+++ b/packages/schematics/angular/library/index.ts
@@ -56,19 +56,25 @@ function updateJsonFile<T>(host: Tree, path: string, callback: UpdateJsonFn<T>):
 }
 
 function updateTsConfig(packageName: string, ...paths: string[]) {
-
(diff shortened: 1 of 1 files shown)


Sherpard:
@googlebot I signed it!

alan-agius4:
@Sherpard, thanks for this, but solution style tsconfig are not longer used in Angular CLI and were only supported in Angular CLI 10.0.x.

When updating to version 10.1 or later the solutions tsconfig file is automatically removed from the workspace.

Sherpard:
Sorry, @alan-agius4 I had no idea that solution style was discontinued, I'll try to migrate my projects to non solution. Style

PR #19323, opened 2020-11-09
refactor(@angular-devkit/build-angular): clean up index generation part 3

alan-agius4:
With this change we cleanup the index generation. The `IndexHtmlWebpackPlugin` now extends the base `IndexHtmlGenerator` class which makes it easier to override methods to retrieve compilation assets. This is important for the critical css extraction implementation because Critters needs to access the `assets` from the either the compilation when running in memory or the file-system.


packages/angular_devkit/build_angular/src/utils/index-file/index-html-generator.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export type IndexHtmlTransform = (content: string) => Promise<string>;
+
+export class IndexHtmlGenerator {
+  private readonly plugins: IndexHtmlGeneratorPlugin[];
+
+  constructor(readonly options: IndexHtmlGeneratorOptions) {
+    this.plugins = [
+      augmentIndexHtmlPlugin(this),
+      this.options.optimization?.fonts.inline ? inlineFontsPlugin(this) : noopPlugin(this),
```
clydin, line 46: could potentially drop the `noopPlugin` if `inlineFontsPlugin` was pushed in an `if` block.

alan-agius4: Agreed

clydin (approved):

PR #19324, opened 2020-11-09
test(@angular-devkit/build-angular): refactor platform server test to use universal schematic

alan-agius4:
Using `@nguniversal/express-engine` requires a lot of maintenance and manual updates to set the correct versions of `@angular-devkit/architect` and `'@angular-devkit/core`.  This is because `@nguniversal/express-engine` can rely on different versions of this packages from the local built versions which would cause package installations failure.


alan-agius4:
This also fixes the CI failure on the patch branch.

clydin (approved):

PR #19336, opened 2020-11-10
fix(@angular-devkit/build-angular): Downgrade Karma to 5.1.x

kyliau:
There's a bug in Karma 5.2.x that causes unit tests to fail when they
should not. It was supposed to be fixed in
https://github.com/karma-runner/karma/pull/3565 but CI still failed,
blocking the release of Karma.

In order to mitigate this, we temporarily downgrade Karma back to 5.1.x

A side-effect of this is that coverage tests that uses karma-coverage
and exceeds the threshold will not fail, because karma-coverage requires
Karma 5.2.x. We expect this to be fixed very soon.

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -13,7 +13,7 @@
     "@angular-devkit/core": "0.0.0",
     "@babel/core": "7.12.3",
     "@babel/generator": "7.12.1",
-    "@babel/plugin-transform-runtime": "7.12.1",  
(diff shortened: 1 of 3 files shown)


alan-agius4 (changes requested):
Can you please change https://github.com/angular/angular-cli/blob/c1d68c46932d5be7eb59407eb24c36a167a33ca1/packages/angular_devkit/build_angular/package.json#L83 to `~5.1.0` as otherwise existing projects will be migrated to 5.2.x.

alan-agius4 (approved):
LGTM, thanks @kyliau

PR #19343, opened 2020-11-11
refactor(@angular-devkit/architect): abstract workspace access from architect

clydin:
This change allows architect runtime implementations to have more control over how builder information and options for targets are found.  A workspace file (and accompanying definition class) is not necessarily needed now.  This also has benefits for unit testing by reducing the amount of potential setup needed per test.


alan-agius4 (approved):
LGTM

PR #19355, opened 2020-11-12
feat(@ngtools/webpack): remove `isPlatform*` calls

arturovt:
The original issue is: https://github.com/angular/universal/issues/1187

At the moment, the code that should be executed on the server is delivered to the browser, thereby increasing the size of the bundle, but this code will never be executed there.

This PR adds a transformer that replaces `isPlatform*` function calls with corresponding boolean values when the application is being built for the browser or the server.

This will allow Terser to remove unnecessary blocks of code during the optimization process.

I was wondering if it's possible to "enable" this transformer only when the application is built in production mode via `ng build --prod`.

### DIFF
--- a/packages/ngtools/webpack/src/angular_compiler_plugin.ts
+++ b/packages/ngtools/webpack/src/angular_compiler_plugin.ts
@@ -61,6 +61,7 @@ import {
 } from './transformers';
 import { collectDeepNodes } from './transformers/ast_helpers';
 import { removeIvyJitSupportCalls } from './transformers/remove-ivy-jit-support-calls';
+import { replaceIsPlatformCalls } from './transformers/replace_is_platform_calls';
 import {
   AUTO_START_ARG,
 } from './type_checker';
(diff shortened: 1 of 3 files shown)


alan-agius4:
Hi @arturovt, thanks for your contribution. In general `isPlatformBrowser` and `isPlatformServer` should really be avoided and platform specific implementations is to be preferred.

For more information see an article by @CaerusKaru, how is one of the Angular universal maintainers. https://medium.com/@caerus.karu/isplatformbrowser-is-not-your-friend-5ee52b06da89

Also, if such optimizations needs to happen @ngtools/webpack is not the right place to have them.

arturovt:
@alan-agius4 thanks for the response. I saw this post but it doesn't solve the bundle size problem. Do you have any ideas how to solve it and what would be the right place to have such optimizations outside of the ngtools/webpack?

CaerusKaru:
I'm not sure what you mean when you say it doesn't solve the bundle size problem. The entire point of the article is about how if you segment and organize your code correctly, you end up with code in separate bundles for free. That's why @alan-agius4 is saying this is unnecessary: if you follow best practices, your browser bundle will have no server code at all (because you build specific code for individual platforms, not to be reused on multiple platforms).

alan-agius4:
Indeed I don't think such optimizations are needed. If you are using `isPlatformBrowser` and `isPlatformServer` I am assuming these are guards for trivial and small logic statements and removing these calls should result in little to no size decease.

alan-agius4:
Closing as per above comments.

PR #19356, opened 2020-11-12
fix(@schematics/angular): remove duplicated Karma traces

cexbrayat:
Currently, Karma is reporting a failed test and the total twice in a bare CLI project:

```
12 11 2020 13:59:46.666:INFO [launcher]: Starting browser Chrome
✔ Browser application bundle generation complete.
12 11 2020 13:59:50.457:INFO [Chrome 86.0.4240.198 (Mac OS 10.15.7)]: Connected on socket F0ehOBWL6BYFqXfbAAAA with id 69358036
Chrome 86.0.4240.198 (Mac OS 10.15.7) AppComponent should render title FAILED
	Error: Expected 'ponyracer app is running!' to contain 'other'.
	    at <Jasmine>
	    at UserContext.<anonymous> (src/app/app.component.spec.ts:29:65)
	    at ZoneDelegate.invoke (node_modules/zone.js/dist/zone-evergreen.js:364:1)
	    at ProxyZoneSpec.push.QpwO.ProxyZoneSpec.onInvoke (node_modules/zone.js/dist/zone-testing.js:292:1)
Chrome 86.0.4240.198 (Mac OS 10.15.7): Executed 2 of 3 (1 FAILED) (0 secs / 0.231 secs)
Chrome 86.0.4240.198 (Mac OS 10.15.7) AppComponent should render title FAILED
	Error: Expected 'ponyracer app is running!' to contain 'other'.
	    at <Jasmine>
	    at UserContext.<anonymous> (src/app/app.component.spec.ts:29:65)
	    at ZoneDelegate.invoke (node_modules/zone.js/dist/zone-evergreen.js:364:1)
Chrome 86.0.4240.198 (Mac OS 10.15.7): Executed 3 of 3 (1 FAILED) (0.309 secs / 0.242 secs)
TOTAL: 1 FAILED, 2 SUCCESS
TOTAL: 1 FAILED, 2 SUCCESS
```

This is a bit annoying when you have several tests failing, and tend to confuse beginners.

This commit configures the Karma HTML reporter to suppress the duplicates (both error and success), which results in only one reporter showing the failed test and total:

```
12 11 2020 14:01:43.002:INFO [launcher]: Starting browser Chrome
✔ Browser application bundle generation complete.
12 11 2020 14:01:58.728:INFO [Chrome 86.0.4240.198 (Mac OS 10.15.7)]: Connected on socket Pc0xPggxJPdC8E_LAAAA with id 54797430
[...]

PR #19357, opened 2020-11-12
fix(@schematics/angular): document jasmine options

cexbrayat:
Currently it can be hard to find how to configure Jasmine in the Karma config.
For example, a developer might want to reproduce a randomly failing test with a specific seed.

This commit adds some information that may be helpful to beginners.

### DIFF
--- a/packages/schematics/angular/application/files/karma.conf.js.template
+++ b/packages/schematics/angular/application/files/karma.conf.js.template
@@ -13,6 +13,12 @@ module.exports = function (config) {
       require('@angular-devkit/build-angular/plugins/karma')
     ],
     client: {
+      jasmine: {
+        // you can add configuration options for Jasmine here
+        // the possible options are listed at https://jasmine.github.io/api/edge/Configuration.html
+        // for example, you can disable the random execution with `random: false`
+        // or set a specific seed with `seed: 4321`
+      },
       clearContext: false // leave Jasmine Spec Runner output visible in browser
     },
     coverageReporter: {
--- a/packages/schematics/angular/library/files/karma.conf.js.template
+++ b/packages/schematics/angular/library/files/karma.conf.js.template
@@ -13,6 +13,12 @@ module.exports = function (config) {
       require('@angular-devkit/build-angular/plugins/karma')
     ],
     client: {
+      jasmine: {
+        // you can add configuration options for Jasmine here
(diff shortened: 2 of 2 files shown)


packages/schematics/angular/application/files/karma.conf.js.template
```diff
@@ -13,6 +13,11 @@ module.exports = function (config) {
       require('@angular-devkit/build-angular/plugins/karma')
     ],
     client: {
+      jasmine: {
+        // if you run into a randomly failing test, you can find the culprit seed in the browser window when running `ng test`
+        // you can then run the same test order with `JASMINE_SEED=123 ng test`
```
alan-agius4, line 18: I don't think this command works cross-platform. 

In general, I am not sure the added benefit of this.  I don't think that when a test fails and user figure that it's because of randomization they will go and look `karma.conf.js`. I find environment variables are not that discoverable.

//cc @clydin what do you think?

cexbrayat: No worries, feel free to discard the PR.
I've seen a lot of teams struggling with this issue and it can be hard to find how to set the seed, so I thought it might be worth it :)

clydin: What about if we leave the `jasmine` object (but empty) and add a comment explaining that jasmine specific options can be placed here?  A link to the available jasmine options may be useful as well.   If there are commonly used options, they could potentially also be listed with a brief description on when they would be useful.

alan-agius4: Agreed. that would be a better approach!

cexbrayat: Sounds good. I pushed a new commit if you want to take another look.
I also updated the commit and PR title/description.

cexbrayat: @alan-agius4 I just updated the config template, but I can of course update the existing karma.conf in the project as I did for #19356 once we agree on the content

alan-agius4 (approved):
LGTM

packages/schematics/angular/application/files/karma.conf.js.template
alan-agius4: No need to add this change there. Thanks

clydin (approved):

PR #19358, opened 2020-11-12
fix(@schematics/angular): remove trailing comma in karma conf

cexbrayat:
We usually don't have trailing commas in the generated code of the CLI (and this one makes the linter/formatter angry).


alan-agius4 (approved):
LGTM

PR #19386, opened 2020-11-13
fix(@angular-devkit/build-angular): ignore hidden inputs with hmr

mhartington:
Don't query for hidden inputs when using HMR

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/hmr/hmr-accept.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/hmr/hmr-accept.ts
@@ -52,7 +52,8 @@ export default function (mod: any): void {
       return;
     }
 
-    const oldInputs = document.querySelectorAll('input, textarea');
+    // Inputs that are hidden should be ignored
+    const oldInputs = document.querySelectorAll('input:not([type="hidden"]), textarea');
     const oldOptions = document.querySelectorAll('option');
 
     // Create new application
@@ -160,8 +161,8 @@ function dispatchEvents(element: any): void {
 }
 
 function restoreFormValues(oldInputs: any[], oldOptions: any[]): void {
-  // Restore input
-  const newInputs = document.querySelectorAll('input, textarea');
+  // Restore input that are not hidden
+  const newInputs = document.querySelectorAll('input:not([type="hidden"]), textarea');
   if (newInputs.length && newInputs.length === oldInputs.length) {
     console.log('[NG HMR] Restoring input/textarea values.');
     for (let index = 0; index < newInputs.length; index++) {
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/hmr/hmr-accept.ts
```diff
@@ -52,7 +52,8 @@ export default function (mod: any): void {
       return;
     }
 
-    const oldInputs = document.querySelectorAll('input, textarea');
+    // Inputs that are hidden should be ignored
+    const oldInputs = document.querySelectorAll('input:not([type="hidden"]), textarea:not([type="hidden"])');
```
alan-agius4, line 56: Textarea's don't have a type hidden.

packages/angular_devkit/build_angular/src/webpack/plugins/hmr/hmr-accept.ts
```diff
@@ -160,8 +161,8 @@ function dispatchEvents(element: any): void {
 }
 
 function restoreFormValues(oldInputs: any[], oldOptions: any[]): void {
-  // Restore input
-  const newInputs = document.querySelectorAll('input, textarea');
+  // Restore input that are not hidden
+  const newInputs = document.querySelectorAll('input:not([type="hidden"]), textarea:not([type="hidden"])');
```
alan-agius4, line 165: Same

alan-agius4 (approved):
I'd love to understand better what is happening. I tried to take a look at the reproduction https://github.com/mhartington/hrm-ionic-test but it's empty.

alan-agius4 (changes requested):
Approved by mistake.

mhartington:
@alan-agius4 pushed the code to hmr-ionic-test


Basically what's happening is that you have the following  DOM structure 


Screen Shot 2020-11-16 at 9 20 40 AM


Where ionic components like Range and Checkbox will render a hidden input within the component. So at the time HMR grabs a snapshot of the outgoing DOM structure, it will grab those hidden inputs. which do nothing and are not something users can directly interact with


So as HMR grabs the snapshot, it could count 1 actual input, and one hidden input. But the hidden input within Ionic's process might not be available right away when the changes are applied, meaning there will be a mismatch in the number of inputs.


Screen Shot 2020-11-16 at 9 26 21 AM


When hmr-accepts attempts to restore the values, with different input lengths it will `noop` out of the whole thing. Which in most cases, makes sense. But in this case, since the inputs are purposefully hidden and are not something that users can interact with, it's safe to ignore them. 


You can see the dom changes yourself by setting a break point at `restoreFormValues` and inspecting the DOM

alan-agius4 (approved):
@mhartington, thanks for the details explanation and the PR. The change LGTM.

Can you please squash the commits to make CI green?

Thanks

mhartington:
Done!

PR #19400, opened 2020-11-16
Update tests E2E

alan-agius4:


tests/legacy-cli/e2e/tests/update/update-8.ts
```diff
@@ -0,0 +1,34 @@
...
+  // Use the packages we are building in this commit, and CI Chrome.
+  await useBuiltPackages();
+  await useCIChrome('./');
+  await useCIChrome('./e2e/');
+  await useCIDefaults('eight-project');
+  await installWorkspacePackages(false);
+
+  // Update Angular.
+  const extraUpdateArgs = await isPrereleaseCli() ? ['--next', '--force'] : [];
+  await ng('update', '@angular/core', ...extraUpdateArgs);
```
clydin, line 20: Since this is starting from version 8, I think it can do a standard update of `ng update @angular/cli @angular/core` instead of the separate migrate only part.

alan-agius4: It doesn't appear to work, because the installed version will be 11 and hence it will not run the previous updates. This is because `ng update` also checks the installed version and not just the version in the workspace package.json.

alan-agius4:
```

Running "tests/update/update-8" (6 of 6)...
Running `npm "install"` [silent]...
CWD: /private/var/folders/mp/zsgl3srd389_js72bk4_bkgh0000gn/T/angular-cli-e2e-O0cueV/assets/8.0-project-1605552811440
ENV: undefined
Running `ng "update" "@angular/cli" "@angular/core" "--next" "--force"`...
CWD: /private/var/folders/mp/zsgl3srd389_js72bk4_bkgh0000gn/T/angular-cli-e2e-O0cueV/assets/8.0-project-1605552811440
ENV: undefined
  Using package manager: 'npm'
  Found 30 dependencies.
  Package '@angular/cli' is already up to date.
```

clydin (approved):

PR #19405, opened 2020-11-16
fix(@ngtools/webpack): remove use of Webpack 5 deprecated compilationDependencies

clydin:
The `Compilation.compilationDependencies` property is now deprecated in Webpack 5.  The `Compilation.fileDependencies` property is now used instead with Webpack 5.

### DIFF
--- a/packages/ngtools/webpack/src/ivy/plugin.ts
+++ b/packages/ngtools/webpack/src/ivy/plugin.ts
@@ -20,7 +20,7 @@ import { TypeScriptPathsPlugin } from '../paths-plugin';
 import { WebpackResourceLoader } from '../resource_loader';
 import { forwardSlashPath } from '../utils';
 import { addError, addWarning } from '../webpack-diagnostics';
-import { mergeResolverMainFields } from '../webpack-version';
+import { isWebpackFiveOrHigher, mergeResolverMainFields } from '../webpack-version';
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -245,9 +245,13 @@ export class AngularWebpackPlugin {
         .filter((sourceFile) => !internalFiles?.has(sourceFile));
 
       // Ensure all program files are considered part of the compilation and will be watched
-      allProgramFiles.forEach((sourceFile) =>
-        compilation.compilationDependencies.add(sourceFile.fileName),
-      );
+      allProgramFiles.forEach((sourceFile) => {
+        if (isWebpackFiveOrHigher()) {
```
alfaproject, line 249: Wouldn't a feature check be better?
```suggestion
        if (compilation.fileDependencies) {
```

clydin: It can actually be present in both cases in certain situations.  In webpack 4 it is added as a property during the compilation process based on the contents of `compilationDependencies`.  Out of an abundance of caution, I went with this method to avoid the potential for hard to diagnose issues in the future. And once Webpack 5 is the default, it can be removed completely.  `isWebpackFiveOrHigher` is also cached so it reduces to a boolean check after the first use.

alan-agius4 (approved):

PR #19436, opened 2020-11-19
fix(@ngtools/webpack): dedupe duplicate dependencies

alan-agius4:


clydin:
Webpack handles this internally (lazily in Webpack 5).
Webpack 4: https://github.com/webpack/webpack/blob/c572c15a413ef7d086b52ccc78d9512a192954d7/lib/NormalModule.js#L305
Webpack 5: 
https://github.com/webpack/webpack/blob/1936615458352f38ff517d258b936d690a7b631f/lib/NormalModule.js#L659-L660

PR #19446, opened 2020-11-20
fix(@angular-devkit/build-angular): apply local libraries sourcemap without `vendor: true`

### LINKED ISSUE #11305: Add option to exclude certain files from sourcemap.
```
- [x] feature request
```

### Area
```
- [x] devkit
- [ ] schematics
```

### Desired functionality

I have projects where enabling sourcemaps slows down the ng build operation by a factor of 10. It would be useful to add an option to exclude certain files from source maps generation, or, even better, being able to specify only the files you want to use to generate source map.

This will allow us to specify which file we want to debug without slowing a lot the compilation time.

alan-agius4:
With this change we apply sourcemaps of libraries in a monorepo without the need to enable `vendor: true`. This is more intuitive behavior and  in line with what the users expect.

We also suppress the `Failed to parse source map from` in-actionable warning.

Closes #11305


clydin (approved):

filipesilva:
@alan-agius4 this PR does not apply cleanly to patch. Can you make a new one please?

PR #19467, opened 2020-11-23
build: update zone.js to 0.11.3

JiaLiPassion:
Update zone.js to the latest version of 0.11.3

### DIFF
--- a/package.json
+++ b/package.json
@@ -237,7 +237,7 @@
     "webpack-sources": "2.2.0",
     "webpack-subresource-integrity": "1.5.2",
     "worker-plugin": "5.0.0",
-    "zone.js": "^0.10.2"
+    "zone.js": "^0.11.3"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):
LGTM. Thanks.

integration/angular_cli/package.json
```diff
@@ -22,7 +22,7 @@
     "@angular/router": "~9.1.1",
     "rxjs": "~6.5.4",
     "tslib": "^1.10.0",
-    "zone.js": "~0.10.2"
```
clydin, line 25: This test project is still using Angular v9 which I believe doesn't support zone.js 0.11?

clydin (changes requested):
Can you split this into two separate commits so that the changelog will show that new projects use zone.js 0.11.x?

1) The current build commit
2) Changes to `packages/schematics/angular/utility/latest-versions.ts` in a `feat(@schematics/angular): ...` commit

integration/angular_cli/package.json
JiaLiPassion: I believe zone.js 0.11 still supports Angular v9, there should no breaking changes impact Angular v9.

JiaLiPassion:
@clydin , got it, I split the PR into two commits, thanks.

clydin (approved):

Issue #19461, opened 2020-11-22
Components using JSX does not trigger application rebuild when running ng serve and AOT


jsantha:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 8.3.29 (maybe related to Ivy or AOT enabled while serving?)

### Description

When an Angular component uses JSX syntax and thus is defined with `.tsx` extension instead of just `.ts` then compiler does not detect changes in Angular component related files `templateUrl` and `styleUrls`. Seems to work well when AOT is disabled.

## 🔬 Minimal Reproduction

Example application representing the issue:
https://github.com/jsantha/angular-cli-tsx-component-not-rebuilding

Reproduction:
1. Create a new Angular workspace and run `ng serve`
2. Create a new Angular component
3. Change component extension from `.ts` to `.tsx` (might need to enable jsx in tsconfig.json)
4. Make any change in component referenced files `templateUrl` or `styleUrls`
5. App does not pick up changes and not trigger rebuild

## 🌍 Your Environment
<pre><code>
Angular CLI: 11.0.2
Node: 14.15.0
OS: win32 x64

Angular:
...
Ivy Workspace:

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1100.2 (cli-only)
@angular-devkit/core         11.0.2 (cli-only)
@angular-devkit/schematics   11.0.2 (cli-only)
@schematics/angular          11.0.2 (cli-only)
@schematics/update           0.1100.2 (cli-only)

</code></pre>

**Anything else relevant?**

alan-agius4:
Hi @jsantha,

This should no longer be an issue in the upcoming Angular CLI version 11.1. You can try the pre-release using the below command. 

```
ng update @angular/cli --next
```

jsantha:
Hi @alan-agius4 

I have updated the reproduction repo to the 11.1. The problem is still here. The only thing that changed is that the stylesheet file triggers the rebuild now but changes are not reflected in the browser. Template changes are still being ignored.

https://github.com/jsantha/angular-cli-tsx-component-not-rebuilding

```
Angular CLI: 11.1.0-next.0
[...]

PR #19471, opened 2020-11-24
test(@ngtools/webpack): update rebuild E2E with component resource changes

clydin:
This adds additional testing for the basic rebuild test to include a change to an external component stylesheet and an external component template.

Closes: #19461

### DIFF
--- a/packages/ngtools/webpack/src/ivy/paths.ts
+++ b/packages/ngtools/webpack/src/ivy/paths.ts
@@ -1,0 +1,42 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import * as nodePath from 'path';
+
+const normalizationCache = new Map<string, string>();
+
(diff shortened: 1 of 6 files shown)


packages/ngtools/webpack/src/ivy/paths.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+import * as nodePath from 'path';
+
+const normalizationCache = new Map<string, string>();
+
+export function normalizePath(path: string): string {
+  let result = normalizationCache.get(path);
+
+  if (result === undefined) {
+    result = nodePath.win32.normalize(path).replace(/\\/g, nodePath.posix.sep);
```
alan-agius4, line 16: Why not just `path.normalize` for posix systems?

clydin: There's no guarantee on the input paths.  They could be a mix of directory separators; potentially doubled up as well.  The Windows normalize variant will guarantee only the default Windows directory separator.  This in turn reduces the complexity of the replace and provides a stronger guarantee on the output.

alan-agius4 (approved):

PR #19530, opened 2020-12-02
fix(@angular/cli): show deprecation warning when using camel cased arguments

alan-agius4:


clydin (approved):

cyrilletuzi:
Hi @alan-agius4, can you explain the motivation of this change?

While it may seem a small change, and a good one for consistency, changing this after years of allowing camelCased arguments will have far more consequences than you may think:

1. As an Angular teacher, wanting to teach correctly, and as it was indeed inconsistent, at one point I stopped an hour to look the details and decide which notation was supposed to be the good one. I don't remember why exactly, but my conclusion was that the good notation was the camelCased one (probably because this was and still is the notation in schematics `schema.json`).

If I jumped to this conclusion, I'm probably not the only one.

It also means to change all the documentation / course materials.

2. There are community tools using the schematics, for example [Angular schematics extension for VS Code](https://marketplace.visualstudio.com/items?itemName=cyrilletuzi.angular-schematics), used by 380 000 developers, or the NX Console.

These tools extract the commands' options from the schematics `schema.json`, where the notation is (still) camelCased, including in the official Angular CLI schematics ([example here](https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/component/schema.json) for `ng g component`).

So it will require quite some change in these tools: now we have to check if it is camelCased with a regular expression (not sure this is good for performance) to know if it must be dasherized.

kyliau:
Hi @cyrilletuzi, we understand that this is not a trivial change, and we assure you this is not a decision that was made hastily. We had a long discussion about this, and the main points we debated were consistency, usability, and flexibility.

While supporting both kebab-case and camelCase helps with flexibility, it does not enforce consistency nor clear usability. New users are often confused by the myriad of options and contradicting instructions.

As you mentioned, the community needs time to adapt to this change. That's why we are only emitting a warning at this moment. We'll give article and library authors at least a year to update their code before we actually remove support for camelCase. The soonest this could land is v13.

camelCase in `schema.json` is necessary by design, because they are converted to TypeScript interfaces, where kebab-case is difficult to use in code. However, `schema.json` is considered private implementation of a schematic.

For your extension, I believe the primary requirement for you is the ability to scrape all options supported by a particular command. Internally, we have the same need when generating documentation for CLI on angular.io, for example https://angular.io/cli/new

To do that, Angular uses `--help=json` command to retrieve all options, and parses that programmatically. This feature is documented in https://angular.io/cli/help. We are currently in the process of updating all documentation to use kebab-case, for example https://pr39944-0a4c4e1.ngbuilds.io/cli/new

Please try this feature, and let us know if it satisfies your need. If not, we are happy to discuss alternatives to make this transition as seamless as possible for you and other authors. Thank you.

cyrilletuzi:
Hi @kyliau, thanks for your answer.

> camelCase in schema.json is necessary by design

Are you confirming that in `schema.json`, it can only be camelCase? That would avoid the need and performance cost to do a check with a regular expression, as I could dasherize directly.

> To do that, Angular uses --help=json command to retrieve all options

I'll try that, but I expect it will be difficult, because I already had problems to run bash commands from inside a VS Code extension (totally unrelated to Angular). In the context of a VS Code extension, that's way easier and safer to get the `schema.json`.

> schema.json is considered private implementation of a schematic.

This worries me, because all the Angular Schematics extension for VS Code relies on this. While I understand normal users shouldn't look into this kind of files, are you saying my tool shouldn't either?

Again, related to what I said above, having only the solution to run a bash command to know the schematics options would be quite problematic in some scenarios.

PR #19535, opened 2020-12-02
doc: update package.json version note to be more visible

dgp1130:
During the release today my eyes completely skipped over the new requirement to update `package.json`. I didn't notice this requirement and was very confused when the release check failed. Changed this to a list to give more visual weight and guide readers eyes to both places that need to be modified.


josephperrott (approved):
LGTM

PR #19557, opened 2020-12-07
docs: remove when true, when false from argument descriptions

alan-agius4:
It's confusing and redundant to use `When true` and `when false` in descriptions for a CLI arguments because specifying false/true is redundant in a command line argument flag and in most cases users will not do it.

Example:
`--foo=true` is the same as `--foo`
`--foo=false` is the same as `--no-foo`

### DIFF
--- a/packages/angular/cli/commands/definitions.json
+++ b/packages/angular/cli/commands/definitions.json
@@ -21,7 +21,7 @@
           ]
         },
         "prod": {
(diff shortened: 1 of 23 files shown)


packages/angular/cli/commands/definitions.json
```diff
@@ -57,12 +57,12 @@
         "interactive": {
           "type": "boolean",
           "default": "true",
-          "description": "When false, disables interactive input prompts."
+          "description": "Enable interactive input prompts."
```
devoto13, line 60: nit: maybe be consistent on the verb form used (e.g. "enable", but "disable**s**" for the next option)?

alan-agius4: Agree

clydin (changes requested):

clydin (approved):

PR #19580, opened 2020-12-09
ci: update remaining references to old pr targeting system

josephperrott:
Completes the migration to use the new unified target label system, updating
dependabot, renovate and angular-robot to expect the new labels.

### DIFF
--- a/.github/angular-robot.yml
+++ b/.github/angular-robot.yml
@@ -18,7 +18,7 @@ merge:
 \nPlease help to unblock it by resolving these conflicts. Thanks!"
 
   # label to monitor
-  mergeLabel: "PR action: merge"
+  mergeLabel: "action: merge"
 
   # list of checks that will determine if the merge label can be added
   checks:
@@ -26,16 +26,15 @@ merge:
     noConflict: true
     # whether the PR should have all reviews completed.
     requireReviews: true
-    # list of labels that a PR needs to have, checked with a regexp (e.g. "PR target:" will work for the label "PR target: master")
+    # list of labels that a PR needs to have, checked with a regexp (e.g. "target:" will work for the label "target: major")
     requiredLabels:
-      - "PR target: *"
+      - "target: *"
       - "cla: yes"
 
     # list of labels that a PR shouldn't have, checked after the required labels with a regexp
     forbiddenLabels:
-      - "PR target: TBD"
-      - "PR action: cleanup"
-      - "PR action: review"
+      - "action: cleanup"
+      - "action: review"
       - "PR state: blocked"
       - "cla: no"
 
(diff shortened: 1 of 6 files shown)


renovate.json
```diff
@@ -6,7 +6,7 @@
   "separateMajorMinor": false,
   "prHourlyLimit": 2,
   "labels": [
-    "PR target: master-only",
+    "target: major",
```
alan-agius4, line 9: I find this confusing, because `master` can be for both a `major` or a `minor`.

Based on the [label docs](https://github.com/angular/angular/blob/master/docs/TRIAGE_AND_LABELS.md#pr-target) `target: major` should be used for breaking changes, while `target: minor` is for new features.

Since there is no 1 to 1 relation between a label and master branch. I feel the best fit is to use `minor`. Since that is how currently the workflow is on the angular/angular repo as far as I can tell. 

Example: https://github.com/angular/angular/pull/40058

Looks to me that `target: major` is more of a substitute for `PR target: master-only` and `flag: breaking change`.

kyliau (approved):
DO you know why we use both dependabot and renovate? Don't they provide the same functionality?

alan-agius4:
dependabot opens PR for transitive security fixes before renovate does AFAIK. renovate supports Bazel workspaces updates while dependabot doesn't.

alan-agius4 (approved):
LGTM

.github/dependabot.yml
```diff
@@ -9,7 +9,7 @@ updates:
       prefix: "build"
     labels:
       - "comp: build & ci"
-      - "PR target: master & patch"
+      - "target: patch"
       - "PR action: merge"
```
alan-agius4, line 13:
```suggestion
      - "action: merge"
```

renovate.json
```diff
@@ -6,7 +6,7 @@
   "separateMajorMinor": false,
   "prHourlyLimit": 2,
   "labels": [
-    "PR target: master-only",
+    "target: minor",
     "PR action: merge"
```
alan-agius4, line 10:
```suggestion
    "action: merge"
```

alan-agius4 (changes requested):
Just noticed that you missing replacing `PR action: merge` with `action: merge`

.github/angular-robot.yml
```diff
@@ -26,14 +26,13 @@ merge:
     noConflict: true
```
alan-agius4, line 26: mergeLabel: "PR action: merge" ->  mergeLabel: "action: merge"

alan-agius4 (approved):
👍

alan-agius4:
@kyliau it looks like renovate are also handling security fixes. Hence, I think we should probably disable dependabot.

PR #19585, opened 2020-12-10
fix(@schematics/angular): adjust block comment 

oleksiipotiekhin:
https://github.com/Microsoft/tslint-microsoft-contrib

ERROR: 21:1  **no-single-line-block-comment**  Replace block comment with a single-line comment

### DIFF
--- a/packages/schematics/angular/application/files/src/polyfills.ts.template
+++ b/packages/schematics/angular/application/files/src/polyfills.ts.template
@@ -18,7 +18,9 @@
  * BROWSER POLYFILLS
  */
 
-/** IE11 requires the following for NgClass support on SVG elements */
+/**
+ * IE11 requires the following for NgClass support on SVG elements
+ */
 // import 'classlist.js';  // Run `npm install --save classlist.js`.
 
 /**
--- a/packages/schematics/angular/migrations/update-8/drop-es6-polyfills_spec.ts
+++ b/packages/schematics/angular/migrations/update-8/drop-es6-polyfills_spec.ts
@@ -41,7 +41,9 @@ import 'core-js/es6/map';
 // import 'core-js/es6/weak-map';
 import 'core-js/es6/set';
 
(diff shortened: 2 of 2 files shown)


packages/schematics/angular/application/files/src/polyfills.ts.template
```diff
@@ -18,7 +18,7 @@
  * BROWSER POLYFILLS
  */
 
-/** IE11 requires the following for NgClass support on SVG elements */
+// IE11 requires the following for NgClass support on SVG elements
```
clydin, line 21: The block comment is intention to aid in indicating which lines should be uncommented and which are informational.
If you would still like to adjust this so that the custom tslint rule set works, can you change this into a multi-line block comment instead?

clydin (changes requested):

packages/schematics/angular/application/files/src/polyfills.ts.template
oleksiipotiekhin: The thing is this is a custom tslint rule set from **Microsoft**, not mine. It could be fixed easily. This is the only reason I assumed that this should be fixed by default.

clydin: The package `tslint-microsoft-contrib` is currently archived and no longer supported (more details: https://github.com/microsoft/tslint-microsoft-contrib/issues/876).

However, a changed similar to the following should allow the rule to pass without changing the comment type:
```
/** 
 * IE11 requires the following for NgClass support on SVG elements
 */
```

oleksiipotiekhin: Sure. So if you think it make sense to change it like that, I will do it. If you think this should be done individually (localy), feel free to close this MR.

alan-agius4 (approved):
@oleksiipotiekhin, thanks for this change. Can you please squash the commits? Otherwise LGTM.

clydin (approved):

Issue #19591, opened 2020-12-11
`--hmr` with `enableProdMode()` breaks HMR


andreialecu:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Not sure.

### Description

Using the new `ng serve --hmr` functionality in Angular 11 while the app was also enabling production mode results in an error while swapping in the modules, and the page falling back to a normal reload with minimal indication that this was happening.

**I was only able to see what's going on by enabling "Preserve Log" in the Chrome console.**

A warning in the console if `enableProdMode()` is called while HMR is active might be a better indication about this.

## 🔬 Minimal Reproduction

```
ng serve --hmr
```

```js
enableProdMode()
```
## 🔥 Exception or Error
<pre><code>
[HMR] Cannot apply update. Need to do a full reload!
push.dZZH.module.exports @ polyfills.js:6327
(anonymous) @ polyfills.js:1434
invoke @ polyfills.js:13461
run @ polyfills.js:13220
(anonymous) @ polyfills.js:13954
invokeTask @ polyfills.js:13496
runTask @ polyfills.js:13264
drainMicroTaskQueue @ polyfills.js:13666
Promise.then (async)
scheduleMicroTask @ polyfills.js:13649
scheduleTask @ polyfills.js:13485
scheduleTask @ polyfills.js:13307
scheduleMicroTask @ polyfills.js:13327
scheduleResolveOrReject @ polyfills.js:13944
then @ polyfills.js:14076
hotUpdateDownloaded @ runtime.js:386
hotAddUpdateChunk @ runtime.js:362
webpackHotUpdateCallback @ runtime.js:58
(anonymous) @ main.4b7dbf18777e0ec0b808.hot-update.js:1

log.js:26 [HMR] Error: Cannot enable prod mode after platform setup.
    at enableProdMode (http://localhost:3000/vendor.js:129787:15) [<root>]
</code></pre>

<img width="853" alt="Screenshot 2020-12-11 at 10 24 39" src="https://user-images.githubusercontent.com/697707/101880138-18625800-3b9b-11eb-87ab-e559a28bd7f1.png">

## 🌍 Your Environment
<pre><code>
[...]

PR #19592, opened 2020-12-11
fix(@angular-devkit/build-angular): add console error when using HMR in production mode

alan-agius4:
FIx: #19591

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/hmr/hmr-accept.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/hmr/hmr-accept.ts
@@ -7,7 +7,7 @@
  */
 
 // tslint:disable-next-line: no-implicit-dependencies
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/hmr/hmr-accept.ts
```diff
@@ -27,6 +27,12 @@ export default function (mod: any): void {
     return;
   }
 
+  if (!isDevMode()) {
+    console.error(`[NG HMR] Cannot use HMR when Angular is running in production mode. Don't call enableProdMode() to enable development mode.`);
```
clydin, line 31:
```suggestion
    console.error(`[NG HMR] Cannot use HMR when Angular is running in production mode. To prevent production mode, do not call 'enableProdMode()'.`);
```

clydin (approved):

filipesilva:
@alan-agius4 this PR does not apply cleanly to 10.2.x. Can you take a look?

Issue #19624, opened 2020-12-14
Vulnerability problem due to outdated dependency


areyes05:
A recent scan found a new vulnerability caused by an outdated dependency. The problem may cause data pollution and it has been ranked [High severity by the CVV](https://nvd.nist.gov/vuln/detail/CVE-2020-7774#vulnCurrentDescriptionTitle).

This is the dependency tree with a problem.
 
| package     | used version | current version |
| ----------- | ------------ | --------------- |
| y18n        | 4.0.0        | 5.0.5           |
| cacache     | 12.0.4       | 15.0.5          |
| pacote      | 9.5.12       | 11.1.13         |
| angular/cli | 11.0.4      | 11.0.4          |

That is, angular/cli uses Pacote 9, which uses Cacache 12 which uses y18n 4. The latter one is where the vulnerability exists.

If angular/cli updates their dependency to request Pacote11, Cacache 15 does not even use y18n as dependency so the problem will be solved for good.

As per the [guidelines](https://angular.io/guide/security#report-issues), an email has been sent to security@angular.io and this issue is created to track its resolution.

clydin:
`y18n` version 4.0.1 was released on 2020-11-30 ([changelog](https://github.com/yargs/y18n/blob/v4/CHANGELOG.md#bug-fixes)) and addresses the linked vulnerability report.  Based on the semver range used within `cacache@12.0.4` (https://github.com/npm/cacache/blob/v12.0.4/package.json#L76), the newer, corrected 4.0.1 version is compatible with `cacache@12.0.4`.  New projects will receive the updated version and existing projects can update their transitive dependencies to resolve the issue.

alan-agius4:
Closing as per above.

PR #19626, opened 2020-12-15
    fix(@angular/cli): update pacote to version 11

alan-agius4:
With this change Pacote is updated to version 11.1.13.
    
This also requires normalization of options because Pacote now passes the options to `npm-registry-fetch` which requires some options to be camelCased.
    
Partially addresses #19624

### DIFF
--- a/packages/angular/cli/utilities/package-metadata.ts
+++ b/packages/angular/cli/utilities/package-metadata.ts
@@ -57,18 +57,20 @@ export interface PackageMetadata {
   'dist-tags'?: unknown;
 }
 
-let npmrc: { [key: string]: string };
+type PackageManagerOptions = Record<string, unknown>;
+
+let npmrc: PackageManagerOptions;
 
 function ensureNpmrc(logger: logging.LoggerApi, usingYarn: boolean, verbose: boolean): void {
   if (!npmrc) {
     try {
       npmrc = readOptions(logger, false, verbose);
-    } catch {}
+    } catch { }
 
     if (usingYarn) {
       try {
         npmrc = { ...npmrc, ...readOptions(logger, true, verbose) };
-      } catch {}
+      } catch { }
     }
   }
 }
@@ -77,7 +79,7 @@ function readOptions(
   logger: logging.LoggerApi,
   yarn = false,
   showPotentials = false,
-): Record<string, string> {
+): PackageManagerOptions {
   const cwd = process.cwd();
   const baseFilename = yarn ? 'yarnrc' : 'npmrc';
   const dotFilename = '.' + baseFilename;
@@ -107,7 +109,7 @@ function readOptions(
     logger.info(`Locating potential ${baseFilename} files:`);
   }
 
-  let options: { [key: string]: string } = {};
+  const options: PackageManagerOptions = {};
   for (const location of [...defaultConfigLocations, ...projectConfigLocations]) {
     if (existsSync(location)) {
       if (showPotentials) {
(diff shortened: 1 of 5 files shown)


packages/angular/cli/utilities/package-metadata.ts
```diff
@@ -107,25 +109,46 @@ function readOptions(
...
-        delete options.cafile;
-        try {
-          options.ca = readFileSync(cafile, 'utf8').replace(/\r?\n/, '\\n');
-        } catch {}
+      // Normalize RC options that are needed by 'npm-registry-fetch'.
+      // See: https://github.com/npm/npm-registry-fetch/blob/ebddbe78a5f67118c1f7af2e02c8a22bcaf9e850/index.js#L99-L126
+      const rcConfig: PackageManagerOptions = yarn ? lockfile.parse(data) : ini.parse(data);
+      for (const [key, value] of Object.entries(rcConfig)) {
+        switch (key) {
+          case 'noproxy':
```
clydin, line 125: `no-proxy` was support as well

clydin (approved):

packages/angular/cli/utilities/package-metadata.ts
```diff
@@ -107,25 +109,48 @@ function readOptions(
...
+            options['strictSSL'] = value;
+            break;
+          case 'local-address':
+            options['localAddress'] = value;
+            break;
+          case 'cafile':
+            if (typeof value === 'string') {
+              const cafile = path.resolve(path.dirname(location), value);
+              try {
+                options['ca'] = readFileSync(cafile, 'utf8').replace(/\r?\n/, '\\n');
```
kyliau, line 146: Do you know why is removing `\r` necessary?
Also, do we need the global `g` flag for the regex? The old code doesn't use the flag, but it seems like a bug.

packages/schematics/update/update/npm.ts
```diff
@@ -55,25 +56,48 @@ function readOptions(
...
-      if (options.cafile) {
-        const cafile = path.resolve(path.dirname(location), options.cafile);
-        delete options.cafile;
-        try {
-          options.ca = readFileSync(cafile, 'utf8').replace(/\r?\n/, '\\n');
-        } catch { }
+      // Normalize RC options that are needed by 'npm-registry-fetch'.
+      // See: https://github.com/npm/npm-registry-fetch/blob/ebddbe78a5f67118c1f7af2e02c8a22bcaf9e850/index.js#L99-L126
+      const rcConfig: PackageManagerOptions = yarn ? lockfile.parse(data) : ini.parse(data);
+      for (const [key, value] of Object.entries(rcConfig)) {
```
kyliau, line 70: nit: Should probably reuse the code. Could you please add a todo?

kyliau (approved):
I guess the risk is there might be options outside the list of known values that we didn't handle?

packages/schematics/update/update/npm.ts
alan-agius4: Probably we should remove this package altogether!

packages/angular/cli/utilities/package-metadata.ts
alan-agius4: I’ll investigate.

clydin: Latest npm code for `cafile`/`ca` option: https://github.com/npm/config/blob/95d566201587931dccb71ef5ca0df826f440e191/lib/index.js#L699-L702

PR #19628, opened 2020-12-15
Angular Builders testing refactor

clydin:
This introduces a Builder test harness to facilitate testing of each of the builders within the `build-angular` package.  The test harness is intended to reduce the amount of test code necessary to test the implementations of each builder while also moving the inputs of the test into the test code itself where possible.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/tests/options/main_spec.ts
+++ b/packages/angular_devkit/build_angular/src/browser/tests/options/main_spec.ts
@@ -1,0 +1,68 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { buildWebpackBrowser } from '../../index';
+import { BASE_OPTIONS, BROWSER_BUILDER_INFO, describeBuilder } from '../setup';
+
+describeBuilder(buildWebpackBrowser, BROWSER_BUILDER_INFO, (harness) => {
+  describe('Option: "main"', () => {
+    it('uses a provided TypeScript file', async () => {
+      harness.useTarget('build', {
+        ...BASE_OPTIONS,
+        main: 'src/main.ts',
+      });
+
+      const { result } = await harness.executeOnce();
+
+      expect(result?.success).toBe(true);
+
+      harness.expectFile('dist/runtime.js').toExist();
+      harness.expectFile('dist/main.js').toExist();
+      harness.expectFile('dist/vendor.js').toExist();
+      harness.expectFile('dist/index.html').toExist();
+    });
+
+    it('uses a provided JavaScript file', async () => {
+      await harness.writeFile('src/main.js', `console.log('main');`);
+
+      harness.useTarget('build', {
+        ...BASE_OPTIONS,
+        main: 'src/main.js',
+      });
+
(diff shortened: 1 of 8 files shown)


packages/angular_devkit/build_angular/src/testing/jasmine-helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+import { readFileSync } from 'fs';
+import { host } from '../test-utils';
+import { BuilderHarness } from './builder-harness';
+
+export function describeBuilder<T>(
+  builderHandler: BuilderHandlerFn<T & json.JsonObject>,
+  options: { name?: string; schemaPath: string },
+  specDefinitions: (harness: JasmineBuilderHarness<T>) => void,
+): void {
+  const optionSchema = JSON.parse(readFileSync(options.schemaPath, 'utf8'));
```
alan-agius4, line 19: Wouldn't it be better to use `require()` so that the schema is cached and avoid an extra IO for every spec file?

I know it won't make much difference between the underlying tests still use the IO heavily.

packages/angular_devkit/build_angular/src/testing/jasmine-helpers.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export interface HarnessFileMatchers {
+  toExist(): boolean;
+  toNotExist(): boolean;
+  readonly content: jasmine.ArrayLikeMatchers<string>;
+  readonly size: jasmine.Matchers<number>;
+}
+
+export function expectFile<T>(path: string, harness: BuilderHarness<T>): HarnessFileMatchers {
```
alan-agius4, line 47: ❤️

packages/angular_devkit/build_angular/src/browser/tests/options/main_spec.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { buildWebpackBrowser } from '../../index';
+import { BASE_OPTIONS, BROWSER_BUILDER_INFO, describeBuilder } from '../setup';
+
+describeBuilder(buildWebpackBrowser, BROWSER_BUILDER_INFO, (harness) => {
+  beforeAll(() => {
+    harness.useProject('test', { root: '.', sourceRoot: 'src' });
```
alan-agius4, line 13: I think in most cases the the `harness.useProject('test', { root: '.', sourceRoot: 'src' });` will be the same for the majority of tests.

This is because, if either one of these options changes, changes need to be done in the underlying test project.

Hence, wouldn't be better that the `useProject` is not needed, or if needed it has default values?

clydin: Good point.  I’ll set those as the default.

packages/angular_devkit/build_angular/src/testing/jasmine-helpers.ts
clydin: Cached the schemas.  I generally prefer not to use `require` for data files.  We also may want to use the full JSON parser instead of `JSON.parse` in the future.

alan-agius4: Fair enough

alan-agius4 (approved):

PR #19639, opened 2020-12-16
fix(@schematics/angular): remove unnecessary scripts and dependencies

oleksiipotiekhin:
When new application is generated with `--minimal` flag it will be created without any testing frameworks, linters, etc. But `package.json` still contains broken scripts and dependencies. So `"test": "ng test"`, `"lint": "ng lint"` and `"tslint": "~6.1.0"` should be removed. By analogy with `"e2e": "ng e2e"`.

### DIFF
--- a/packages/schematics/angular/workspace/files/package.json.template
+++ b/packages/schematics/angular/workspace/files/package.json.template
@@ -4,9 +4,9 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
-    "build": "ng build",
+    "build": "ng build"<% if (!minimal) { %>,
     "test": "ng test",
-    "lint": "ng lint"
+    "lint": "ng lint"<% } %>
   },
   "private": true,
   "dependencies": {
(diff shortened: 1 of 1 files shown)


oleksiipotiekhin:
@alan-agius4

packages/schematics/angular/workspace/files/package.json.template
```diff
@@ -36,8 +36,8 @@
     "karma-jasmine": "~4.0.0",
     "karma-jasmine-html-reporter": "^1.5.0",
     "protractor": "~7.0.0",<% } %>
-    "ts-node": "~8.3.0",
-    "tslint": "~6.1.0",
+    "ts-node": "~8.3.0",<% if (!minimal) { %>
```
alan-agius4, line 39:
```diff
-    "protractor": "~7.0.0",<% } %>
-    "ts-node": "~8.3.0",<% if (!minimal) { %>
+    "protractor": "~7.0.0",
+    "ts-node": "~8.3.0",
     "tslint": "~6.1.0",<% } %>
```

`ts-node` is only required for `protractor`. Hence it can safely be removed when minimal is `true.`

alan-agius4 (changes requested):
Thanks for this.

See below comment.

alan-agius4 (approved):
LGTM

packages/schematics/angular/workspace/files/package.json.template
oleksiipotiekhin: Done. Now `ts-node` is placed inside of `if` block.

PS: commit message body was updated.

PR #19641, opened 2020-12-16
feat(@angular-devkit/build-angular): integrate Angular compiler linker

clydin:
The newly introduced library linker mode that removes the need for ngcc is now integrated into the Angular CLI's build system.  This allows libraries that are built in linker mode to be used when building an application.  Linker processing is only performed on files that contain _partial_ compiled AOT code.

NOTE: Support for the linker is currently experimental.


packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
```diff
@@ -0,0 +1,83 @@
...
+    cacheCompression: false,
+    sourceType: 'unambiguous',
+  });
+  let hasLinkerSupport: undefined | boolean;
+
+  return {
+    customOptions({ forceES5, ...loaderOptions }, { source }) {
+      const requiresLinking =
+        (source.includes('ɵɵngDeclareDirective') || source.includes('ɵɵngDeclareComponent')) &&
+        !source.includes('function ɵɵngDeclareDirective(decl)');
```
petebacondarwin, line 30: Now that https://github.com/angular/angular/pull/40137 has landed, we can use the provided function.

packages/angular_devkit/build_angular/src/babel/presets/application.ts
```diff
@@ -103,6 +105,12 @@ export default function (api: unknown, options: ApplicationPresetOptions) {
   const plugins = [];
   let needRuntimeTransform = false;
 
+  if (options.angularLinker) {
+    // Babel currently is synchronous so import cannot be used
+    const { createEs2015LinkerPlugin } = require('@angular/compiler-cli/linker/babel');
+    plugins.push(createEs2015LinkerPlugin());
```
petebacondarwin, line 111: For full external template source-map support we will need to provide a `FileSystem` and a `Logger` to this function. See https://github.com/angular/angular/pull/40237

clydin: Logger is now integrated.  The filesystem is a little more complicated to integrate.  Once the upstream PR is in place, we can do a followup PR to add full sourcemap support to the tooling integration.

tests/legacy-cli/e2e/tests/generate/library/library-consumption-linker.ts
```diff
@@ -0,0 +1,100 @@
...
+  // Enable partial compilation mode (linker) for the library
+  await updateJsonFile('projects/my-lib/tsconfig.lib.json', config => {
+    const { angularCompilerOptions = {} } = config;
+    angularCompilerOptions.compilationMode = 'partial';
+    config.angularCompilerOptions = angularCompilerOptions;
+  });
+  // Enable ivy for production as well
+  await updateJsonFile('projects/my-lib/tsconfig.lib.prod.json', config => {
+    const { angularCompilerOptions = {} } = config;
+    angularCompilerOptions.enableIvy = true;
```
alan-agius4, line 17:
```suggestion
    angularCompilerOptions.enableIvy = true;
    angularCompilerOptions.compilationMode = 'partial';
```

tests/legacy-cli/e2e/tests/generate/library/library-consumption-linker.ts
```diff
@@ -0,0 +1,100 @@
...
+import { ng } from '../../../utils/process';
+import { updateJsonFile } from '../../../utils/project';
+
+export default async function () {
+  await ng('generate', 'library', 'my-lib');
+
+  // Enable partial compilation mode (linker) for the library
+  await updateJsonFile('projects/my-lib/tsconfig.lib.json', config => {
+    const { angularCompilerOptions = {} } = config;
+    angularCompilerOptions.compilationMode = 'partial';
```
alan-agius4, line 11: The partial should only be set in the production configuration.

This is because, during development Ivy will run in full mode, to avoid the linker step for faster incremental builds. (At least that was the idea before).

```suggestion
```

alan-agius4 (changes requested):
Small NIT.

packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
```diff
@@ -0,0 +1,94 @@
...
+          this.resourcePath,
+          source,
+        );
+
+        // If the linker entry point is present then there is linker support
+        hasLinkerSupport = true;
+      } catch {
+        // Fallback for Angular versions less than 11.1.0
+        requiresLinking =
+          source.includes('ɵɵngDeclareDirective') || source.includes('ɵɵngDeclareComponent');
```
petebacondarwin, line 39:
```suggestion
        requiresLinking = false;
```

Given that there is no support for linking prior to 11.1.0

packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
```diff
@@ -0,0 +1,94 @@
...
+      }
+
+      if (/[\\\/]@angular[\\\/](?:compiler|core)/.test(this.resourcePath)) {
+        requiresLinking = false;
+      }
```
petebacondarwin, line 44: Perhaps move this up above the `try` block and stick the `try` block in any `else` clause.
This would mean that you don't need to try to incur the cost of checking the `needsLinking()` branch if we know we don't care about this file.

tests/legacy-cli/e2e/tests/generate/library/library-consumption-linker.ts
clydin: I initially did this for the test to validate it would work with both types of configuration option sets (the ivy development scenario is already tested in the existing consumption test).  While we most likely won't generate an initial library this way, it should still work if a user chooses to do so.  I'm open to changing it though.

packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
clydin: I agree and was actually considering pulling this out into a separate function.

tests/legacy-cli/e2e/tests/generate/library/library-consumption-linker.ts
alan-agius4: In that case there isn’t any difference between a library built in dev or prod mode.

Only the consumption would be different.

I am fine either way.

clydin: Updated to align with the recommended future library use cases.

alan-agius4 (approved):

packages/angular_devkit/build_angular/src/babel/presets/application.ts
```diff
@@ -98,11 +101,47 @@ function createI18nPlugins(
   return plugins;
 }
 
+function createNgtscLogger(
+  reporter: DiagnosticReporter | undefined,
+): import('@angular/compiler-cli/src/ngtsc/logging').Logger {
+  return {
+    level: 1, // Info level
```
alan-agius4, line 108: Nit: since we don't use the `info` messages shouldn't this be changed to `warning`.
https://github.com/angular/angular-cli/pull/19641/files#diff-c36474716c699264f5af41ac0f4fb12c6e0a8580e9063813b20fff48e5f43548R107

PR #19659, opened 2020-12-21
fix(@schematics/angular): replace of old default html placeholder in the template

oleksiipotiekhin:
When `--inline-template` flag is set to `true` the `app.component.html` will be generated with old html placeholder. Instead it should use new template from `app.component.html.template`.

P.S. the html placeholder was copied and only one adjustment was made. Content of `.terminal::before` pseudo-element was changed from `content: "\2022 \2022 \2022";` to `content: "\\2022 \\2022 \\2022";` because of inline template parsing.

### DIFF
--- a/packages/schematics/angular/application/other-files/app.component.ts.template
+++ b/packages/schematics/angular/application/other-files/app.component.ts.template
@@ -3,29 +3,540 @@ import { Component } from '@angular/core';
 @Component({
   selector: '<%= selector %>',<% if(inlineTemplate) { %>
   template: `
-    <!--The content below is only a placeholder and can be replaced.-->
-    <div style="text-align:center" class="content">
-      <h1>
-        Welcome to {{title}}!
-      </h1>
-      <span style="display: block">{{ title }} app is running!</span>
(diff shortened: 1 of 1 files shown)


oleksiipotiekhin:
> When `--inline-template` flag is set to `true` the `app.component.html` will be generated with old html placeholder. Instead it should use new template from `app.component.html.template`.
> 
> P.S. the html placeholder was copied and only one adjustment was made. Content of `.terminal::before` pseudo-element was changed from `content: "\2022 \2022 \2022";` to `content: "\\2022 \\2022 \\2022";` because of inline template parsing.

oleksiipotiekhin:
@alan-agius4

alan-agius4 (commented):
Thanks for this. 

I personally think that this change will introduce a lot of chum for new projects when inline templates is enabled.

That said, let me mark this for discussion with the rest of the team.

alan-agius4:
Hi @oleksiipotiekhin,

Thanks for this contribution. 

We have discussed this during our team meeting and we feel that we shouldn't include the full template when choosing `--inline-template` as the `app.component` file to grow quite a lot.

PR #19672, opened 2020-12-23
build: update dependencies of 11.0.x

matheo:
This PR contains the following updates:

latest `@babel` packages
latest `@angular` 11.0.x packages and hashes

Last night I updated my websites and `@babel/templates` was fixed to an old version
so this synces the 11.0.x branch with `@angular/*` 11.0.x too,
as there's a little void updating these when `master` gets the `next` version.


matheo:
@filipesilva can we have a patch release with updated deps?

alan-agius4:
Hi @matheo, thanks for this.

The latest Babel packages are consumed in version 11.1. Typically dependencies updates only happen in minors because of the risk of unexpected breakages.

Is there any reason why we need to update `@babel/` packages in a patch version? Does it fix an outstanding issue?

matheo:
@alan-agius4 nope, it just syncs the package-lock versions in the Angular Apps, no known issues tho

alan-agius4:
In such case it’s best not to update Babel packages in a patch version.

matheo:
ok, closing this

PR #19702, opened 2020-12-31
test(@angular-devkit/build-angular): add initial browser builder watch option tests

clydin:
This change adds internal support for providing a custom file watching mechanism to the browser (and associated) builders.  The support integrates and overrides the Webpack watch system when enabled.  This is currently intended to support builder unit testing use cases.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/tests/options/watch_spec.ts
+++ b/packages/angular_devkit/build_angular/src/browser/tests/options/watch_spec.ts
@@ -1,0 +1,103 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { concatMap, count, take, timeout } from 'rxjs/operators';
+import { buildWebpackBrowser } from '../../index';
+import { BASE_OPTIONS, BROWSER_BUILDER_INFO, describeBuilder } from '../setup';
+
+describeBuilder(buildWebpackBrowser, BROWSER_BUILDER_INFO, (harness) => {
+  describe('Option: "watch"', () => {
+    it('does not wait for file changes when false', (done) => {
+      harness.useTarget('build', {
+        ...BASE_OPTIONS,
+        watch: false,
+      });
+
+      // If the build waits then it will timeout with the custom timeout.
+      // A single build should not take more than 15 seconds.
+      let count = 0;
+      harness
+        .execute()
+        .pipe(timeout(15000))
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/browser/tests/options/watch_spec.ts
```diff
@@ -0,0 +1,76 @@
...
+        ...BASE_OPTIONS,
+        main: 'src/main.ts',
+        watch: true,
+      });
+
+      let count = 0;
+      await harness
+        .execute()
+        .pipe(
+          timeout(30000),
```
alan-agius4, line 51: Do we still need the timeout?

packages/angular_devkit/build_angular/src/browser/tests/options/watch_spec.ts
```diff
@@ -0,0 +1,76 @@
...
+            }
+          }),
+          mergeMap(async () => {
+            if (count === 1) {
+              return harness.modifyFile(
+                'src/main.ts',
+                (content) => content + 'console.log("abcd1234");',
+              );
+            }
+          }),
```
alan-agius4, line 68: Maybe the below is simpler?

```suggestion
        concatMap(async (_, index) => {
          switch (index) {
            case 0:
              harness.expectFile('dist/main.js').content.not.toContain('abcd1234');
              await harness.modifyFile('src/main.ts', (content) => content + 'console.log("abcd1234");'),
              break;
            case 1:
              harness.expectFile('dist/main.js').content.toContain('abcd1234');
              break;
          }
        }),
```

clydin: Yes, sounds good.

packages/angular_devkit/build_angular/src/browser/tests/options/watch_spec.ts
clydin: It helps in error cases such as watch not picking up the change.  With this timeout being less than the jasmine timeout, the observable will be guaranteed to be closed out.  Whereas with just the Jasmine timeout, the watch could trigger after the test fails and still execute the operators.  I was actually considering adding an automatic builder abort into the jasmine test harness to mitigate this without the extra timeouts.

alan-agius4: >  I was actually considering adding an automatic builder abort into the jasmine test harness to mitigate this without the extra timeouts.

I do like that, although that can be done in a seperate PR.

alan-agius4 (approved):
LGTM

PR #19704, opened 2021-01-01
fix(@angular-devkit/build-angular): normalize optimization in getComm…

rhutchison:
…onConfig

Ref: https://github.com/angular/angular-cli/commit/eb30a92e8ad4bf27461010b1917d1fcffd10ff31#diff-f3bcbd6f6b0756d2178dd69396750fa742093fb7f4668ab284463c0ca348b032R393

WARN TypeError: Cannot read property 'minify' of undefined
WARN     at getCommonConfig (node_modules\@angular-devkit\build-angular\src\webpack\configs\common.js:299:28)
WARN     at Object.getAngularCliParts (node_modules\@storybook\angular\dist\server\angular-cli_utils.js:83:30)
WARN     at Object.applyAngularCliWebpackConfig (node_modules\@storybook\angular\dist\server\angular-cli_config.js:134:40)
WARN     at Object.webpackFinal (node_modules\@storybook\angular\dist\server\framework-preset-angular-cli.js:10:33)
WARN     at node_modules\@storybook\core\dist\server\presets.js:259:28

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -25,7 +25,7 @@ import {
 } from 'webpack';
 import { RawSource } from 'webpack-sources';
 import { AssetPatternClass } from '../../browser/schema';
-import { BuildBrowserFeatures, maxWorkers } from '../../utils';
+import { BuildBrowserFeatures, maxWorkers, normalizeOptimization } from '../../utils';
 import { WebpackConfigOptions } from '../../utils/build-options';
 import { findCachePath } from '../../utils/cache-path';
 import {
@@ -62,11 +62,12 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
       scripts: scriptsSourceMap,
       vendor: vendorSourceMap,
     },
-    optimization: {
-      styles: stylesOptimization,
-      scripts: scriptsOptimization,
-    },
   } = buildOptions;
+
+  const {
+    styles: stylesOptimization,
+    scripts: scriptsOptimization,
+  } = normalizeOptimization(buildOptions.optimization);
 
   const extraPlugins: { apply(compiler: Compiler): void }[] = [];
   const extraRules: RuleSetRule[] = [];


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -62,12 +62,13 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
-    optimization: {
-      styles: stylesOptimization,
-      scripts: scriptsOptimization,
-    },
   } = buildOptions;
 
+  const {
+    styles: stylesOptimization,
+    scripts: scriptsOptimization,
+  } = normalizeOptimization(buildOptions.optimization);
```
alan-agius4, line 70: At this stage the `buildOptions.optimization` are already normalized.

https://github.com/angular/angular-cli/blob/97aa1fb7cc7595ca696fbe72fdf4982dd9815dfd/packages/angular_devkit/build_angular/src/utils/normalize-builder-schema.ts#L51

rhutchison: That works for angular internals, but what about 3rd party libs using public api?

https://github.com/storybookjs/storybook/blob/bf61334046da41dbf1580d95a3faae5305a0deb7/app/angular/src/server/angular-cli_utils.ts#L86

alan-agius4: These are all private APIs which shouldn’t be used by 3rd parties.

Anything which is imported using deep imports or not from `@angular-devkit/build-angular` is private.

rhutchison: @alan-agius4 normalizeBrowserSchema is also private -- if I were to work on a PR for @storybook/angular, what would be the public api I should look for?  Is there a document you can point me to?

alan-agius4: Unfortunately there isn’t a documentation for the public API.

Although all the public API methods can be found here: https://github.com/angular/angular-cli/blob/cfb368408b532e488acaa1a5a921240f0dc7dfad/packages/angular_devkit/build_angular/src/index.ts#L30

rhutchison: Thanks for feedback, is there a reason you're opposed to this fix?

alan-agius4: The reason being is that its redundant in the supported scenarios.

PR #19713, opened 2021-01-05
fix(@angular-devkit/build-angular): correctly resolve symlinked global styles entrypoints

### LINKED ISSUE #3500: Stylesheet from (npm) linked package not loaded correctly
### OS
Tested on both Windows 7 and Linux (Fedora 24)

### Version
```
$ ng --version
angular-cli: 1.0.0-beta.22-1
node: 4.6.1
os: linux x64
```
The app was created using CLI.

### Repro steps
1. Install a package containing a stylesheet (say bootstrap-sass, actually any package will do) globally via npm.
2. git clone https://github.com/twbs/bootstrap-sass.git into a working directory, cd to that working directory and perform ```npm link``` command.
3. ng new project in another separate directory, cd to the newly create project directory and perform ```npm link bootstrap-sass```. This will create a ```bootstrap-sass``` symlink in node_modules pointing to the local git repository.
4. Modify angular-cli.json to reference the stylesheet in the linked package.
```
      "styles": [
        "../node_modules/bootstrap-sass/assets/stylesheets/_bootstrap.scss",
        "styles.scss"
      ],
```
5. ```ng serve``` and observe that styles from bootstrap is not exported correctly in the index page.
6. ```ng build``` and observe that the generated styles.bundle.js does not do "exports.push()".

There was no error log, webpack declared the "bundle is now valid".

Using normal npm install approach does not have this issue. i.e. after ```npm unlink bootstrap-sass && npm install bootstrap-sass``` then all is fine.

Use case: I have a forked ```gentelella``` repository locally to fix some of the issues I found, so I cannot just simply npm install it in my project. Using npm link approach as I outlined above should  work. It seems angular-cli (or webpack, I am not sure) is being confused by symlink. As a temporary workaround, I replace the stylesheet reference in the angular-cli.json with an absolute path to the stylesheet file in the local git repo.

alan-agius4:
With this change we resolve the global stylesheet entrypoint path to use the realpath instead of the symlink path.

Fixes #3500

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -51,6 +51,11 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
           resolvedPath = require.resolve(style.input, { paths: [root] });
         } catch {}
       }
+
+      if (!buildOptions.preserveSymlinks) {
+        resolvedPath = fs.realpathSync(resolvedPath);
+      }
+
       // Add style entry points.
       if (entryPoints[style.bundleName]) {
         entryPoints[style.bundleName].push(resolvedPath);
--- a/tests/legacy-cli/e2e/tests/build/styles/symlinked-global.ts
+++ b/tests/legacy-cli/e2e/tests/build/styles/symlinked-global.ts
@@ -1,0 +1,34 @@
+import { symlinkSync } from 'fs';
+import { resolve } from 'path';
+import { expectFileToMatch, writeMultipleFiles } from '../../../utils/fs';
+import { ng } from '../../../utils/process';
+import { updateJsonFile } from '../../../utils/project';
+
+export default async function () {
+  await writeMultipleFiles({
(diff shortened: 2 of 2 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -46,11 +46,14 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
 
     normalizeExtraEntryPoints(buildOptions.styles, 'styles').forEach(style => {
       let resolvedPath = path.resolve(root, style.input);
-      if (!fs.existsSync(resolvedPath)) {
+      if (fs.existsSync(resolvedPath)) {
+        resolvedPath = fs.realpathSync(resolvedPath);
```
clydin, line 50: Can you add a check for the `preserveSymlinks` option here as well?  Always using the real path could break some use cases.

clydin (changes requested):

clydin (approved):

PR #19760, opened 2021-01-11
test(@angular-devkit/build-angular): change `extractLicenses` unit test to use build harness

alan-agius4:
This change adds a unit tests for the browser builder's extractLicenses option using the builder test harness.


clydin (commented):
Can you add one more for the not present state (tests the default value)?

alan-agius4:
@clydin PTAL.

clydin (approved):

PR #19766, opened 2021-01-11
ci: change merge strategy to rebase instead of squash

josephperrott:
Change the merge method used by merge tooling to use rebase instead of using
squash.


alan-agius4 (approved):
Thanks for this @josephperrott

PR #19790, opened 2021-01-14
fix(@ngtools/webpack): use `getDiagnosticsForFile` instead of `getDiagnostics`

alan-agius4:
With this change we replace `getDiagnostics` with `getDiagnosticsForFile`. `getDiagnostics` no longer accept a parameter and is intended to be used instead of `getDiagnostics`.

I used `OptimizeFor.WholeProgram` based on the comments in https://github.com/angular/angular/pull/40331#discussion_r555953776 which suggests that if you iterate through files one at a time the `WholeProgram` flag should be used.

For more context see: https://github.com/angular/angular/pull/40331


packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -425,7 +426,7 @@ export class AngularWebpackPlugin {
       for (const sourceFile of builder.getSourceFiles()) {
         // Collect Angular template diagnostics
         if (!ignoreForDiagnostics.has(sourceFile)) {
-          diagnosticsReporter(angularCompiler.getDiagnostics(sourceFile));
+          diagnosticsReporter(angularCompiler.getDiagnosticsForFile(sourceFile, OptimizeFor.WholeProgram));
```
petebacondarwin, line 429: Do you need a fallback to `getDiagnostics()` if the new API is not available?

alan-agius4: The Ivy plugin is only available in CLI version 11.1. I believe this doesn't work properly Angular 11.0. But I might be wrong.

@clydin would know more.

clydin: It’s unlikely but someone could use 11.0 of the compiler with 11.1 of the CLI.  So a guard and some creative casting should ideally be present here.

alan-agius4: Updated

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -425,7 +425,17 @@ export class AngularWebpackPlugin {
...
-          diagnosticsReporter(angularCompiler.getDiagnostics(sourceFile));
+          // The below check should be removed once support for compiler 11.0 is dropped.
+          // Also, the below require should be changed to an ES6 import.
+          if (angularCompiler.getDiagnosticsForFile) {
+            // @angular/compiler-cli 11.1+
+            const { OptimizeFor } = require('@angular/compiler-cli/src/ngtsc/typecheck/api');
+            diagnosticsReporter(angularCompiler.getDiagnosticsForFile(sourceFile, OptimizeFor.WholeProgram));
+          } else {
+            // @angular/compiler-cli 11.0+
+            const getDiagnostics = angularCompiler.getDiagnostics as (sourceFile: ts.SourceFile) => ts.Diagnostic[];
```
clydin, line 436: This will either need a `bind` here or a use of `call` at the invocation on the next line to maintain the object instance reference.

alan-agius4: You are right, good catch.

clydin (approved):

PR #19801, opened 2021-01-15
fix(@angular-devkit/build-angular): resolve i18n outFile with normalized format

cexbrayat:
If someone uses the `extract-i18n` command with the deprecated `--i18n-format` option,
then `options.format` is undefined and the `outFile` defaults to `messages.xlf`,
whereas the content will be using the proper format.

    ng extract-i18n --i18n-format json
    # generates messages.xlf with JSON content

As the command is deprecated I'm not sure it's worth adding a complete e2e test. Let me know if the PR needs to be improved.


alan-agius4 (approved):
LGTM. Thanks

PR #19818, opened 2021-01-19
refactor(@angular-devkit/build-angular): remove virtualfs host usage from browser/dev-server

clydin:
The virtualfs Host was only used for exist and directory checks during asset and file replacement option normalization within the browser and dev-server builders.


alan-agius4 (approved):
LGTM

PR #19821, opened 2021-01-19
fix(@angular/cli): add @angular-devkit/architect to ng-update packageGroup

minijus:
Having `@angular-devkit/architect` as a dependency and running `ng update @angular/cli` leaves it not updated.
Adding this package to `@angular/cli` `packageGroup` will solve this inconvenience.


alan-agius4 (approved):
LGTM, thanks

Issue #16392, opened 2019-12-08
ng update fails with "Cannot find module '@angular-devkit/schematics'"


rokerkony:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

Hi there, after reading a tweet from Igor to test out the new RC of ng9 I tried it on our application. The update command is not able to be finished even with flag `--force` due to the log below.

## 🔥 Exception or Error
<pre><code>
ng update @angular/core @angular/cli --next --force
The installed Angular CLI version is older than the latest pre-release version.
Installing a temporary version to perform the update.
Installing packages for tooling via npm.
Installed packages for tooling via npm.
Using package manager: 'npm'
Collecting installed dependencies...
Found 74 dependencies.
Fetching dependency metadata from registry...
                  Package "@angular/core" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/compiler-cli" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/compiler" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/common" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/animations" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/platform-browser" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/forms" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/router" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/platform-browser-dynamic" has a missing peer dependency of "tslib" @ "^1.10.0".
                  Package "@angular/service-worker" has a missing peer dependency of "tslib" @ "^1.10.0".
    Updating package.json with dependency @angular/cli @ "9.0.0-rc.5" (was "8.3.20")...
    Updating package.json with dependency @angular/core @ "9.0.0-rc.5" (was "8.2.14")...
    Updating package.json with dependency @angular/compiler-cli @ "9.0.0-rc.5" (was "8.2.14")...
    Updating package.json with dependency @angular/language-service @ "9.0.0-rc.5" (was "8.2.14")...
[...]

PR #19857, opened 2021-01-23
fix(@angular/cli): direct Angular schematic dependency requests to known versions

clydin:
This change adds logic to redirect module resolution requests for Angular schematics to ensure that the correct versions of core schematic related packages are used.  This also ensures that the runtime version of the schematics package matches the version used inside the schematic and that object instances passed into the schematic are compatible.
The current set of core schematic related packages are `@angular-devkit/*` and `@schematics/angular`.  Only first-party Angular schematics are currently affected by this change.
This change also solves issues such as #16392 which were caused by the package manager not hoisting the core schematic packages to a location the update migrations could access.

### DIFF
--- a/packages/angular/cli/models/schematic-engine-host.ts
+++ b/packages/angular/cli/models/schematic-engine-host.ts
@@ -1,0 +1,189 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { RuleFactory, SchematicsException, Tree } from '@angular-devkit/schematics';
+import { NodeModulesEngineHost } from '@angular-devkit/schematics/tools';
+import { readFileSync } from 'fs';
+import { parse as parseJson } from 'jsonc-parser';
+import { dirname, resolve } from 'path';
+import { Script } from 'vm';
+
+/**
+ * Environment variable to control schematic package redirection
+ * Default: Angular schematics only
+ */
+const schematicRedirectVariable = process.env['NG_SCHEMATIC_REDIRECT']?.toLowerCase();
+
+function shouldWrapSchematic(schematicFile: string): boolean {
+  // Check environment variable if present
+  if (schematicRedirectVariable !== undefined) {
+    switch (schematicRedirectVariable) {
+      case '0':
+      case 'false':
+      case 'off':
+      case 'none':
+        return false;
+      case 'all':
+        return true;
+    }
+  }
+
+  // Never wrap `@schematics/update` when executed directly
+  // It communicates with the update command via `global`
+  if (/[\/\\]node_modules[\/\\]@schematics[\/\\]update[\/\\]/.test(schematicFile)) {
+    return false;
+  }
+
+  // Default is only first-party Angular schematic packages
+  // Angular schematics are safe to use in the wrapped VM context
+  return /[\/\\]node_modules[\/\\]@(?:angular|schematics|nguniversal)[\/\\]/.test(schematicFile);
+}
+
+export class SchematicEngineHost extends NodeModulesEngineHost {
+  protected _resolveReferenceString(refString: string, parentPath: string) {
+    const [path, name] = refString.split('#', 2);
+    // Mimic behavior of ExportStringRef class used in default behavior
+    const fullPath = path[0] === '.' ? resolve(parentPath ?? process.cwd(), path) : path;
+
+    const schematicFile = require.resolve(fullPath, { paths: [parentPath] });
+
+    if (shouldWrapSchematic(schematicFile)) {
+      const schematicPath = dirname(schematicFile);
+
+      const moduleCache = new Map<string, unknown>();
+      const factoryInitializer = wrap(
+        schematicFile,
+        schematicPath,
+        moduleCache,
+        name || 'default',
+      ) as () => RuleFactory<{}>;
+
+      const factory = factoryInitializer();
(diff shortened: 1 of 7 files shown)


packages/angular/cli/models/schematic-engine-host.ts
```diff
@@ -0,0 +1,122 @@
...
+import { readFileSync } from 'fs';
+import { dirname } from 'path';
+import { Script } from 'vm';
+
+export class SchematicEngineHost extends NodeModulesEngineHost {
+  protected _resolveReferenceString(refString: string, parentPath: string) {
+    const [path, name] = refString.split('#', 2);
+    const schematicFile = require.resolve(path, { paths: [parentPath] });
+
+    const isAngularSchematic = /[\/\\]node_modules[\/\\]@(?:angular|schematics)[\/\\]/.test(
```
alan-agius4, line 19:
```suggestion
    const isAngularSchematic = /[\/\\]node_modules[\/\\]@(?:angular|schematics|nguniversal)[\/\\]/.test(
```

packages/angular/cli/models/schematic-engine-host.ts
```diff
@@ -0,0 +1,122 @@
...
+      // Resolve from inside the @angular/cli project
+      const packagePath = require.resolve(id);
+
+      return require(packagePath);
+    } else if (id.startsWith('.') || id.startsWith('@angular/cdk')) {
+      // Resolve from the original file
+      const modulePath = scopedRequire.resolve(id);
+
+      // Do not wrap vendored third-party packages
+      if (!modulePath.includes('@schematics/angular/third_party')) {
```
alan-agius4, line 79: Why not `startsWith`

packages/angular/cli/models/schematic-engine-host.ts
```diff
@@ -0,0 +1,122 @@
...
+  // `createRequireFromPath` is deprecated in 12+ and can be removed once 10.x support is removed
+  const scopedRequire = createRequire?.(schematicFile) || createRequireFromPath(schematicDirectory);
+
+  const customRequire = function (id: string) {
+    if (id.startsWith('@angular-devkit/') || id.startsWith('@schematics/')) {
+      // Resolve from inside the @angular/cli project
+      const packagePath = require.resolve(id);
+
+      return require(packagePath);
+    } else if (id.startsWith('.') || id.startsWith('@angular/cdk')) {
```
alan-agius4, line 74: Why the CDK schematic is handled differently?

packages/angular/cli/models/schematic-engine-host.ts
```diff
@@ -0,0 +1,122 @@
...
+  moduleCache: Map<string, unknown>,
+  exportName?: string,
+): () => unknown {
+  const { createRequire, createRequireFromPath } = require('module');
+  // Node.js 10.x does not support `createRequire` so fallback to `createRequireFromPath`
+  // `createRequireFromPath` is deprecated in 12+ and can be removed once 10.x support is removed
+  const scopedRequire = createRequire?.(schematicFile) || createRequireFromPath(schematicDirectory);
+
+  const customRequire = function (id: string) {
+    if (id.startsWith('@angular-devkit/') || id.startsWith('@schematics/')) {
```
alan-agius4, line 69: I think this logic might cause problems when using Angular packages which rely on older versions of Angular schematics packages.

Something like;
```
ng update @angular/core@8
```

Previously, users could have worked around this problem by installing the correct version of the dependency in their project, Example ` npm i @angular-devkit/schematics@8`, but now this will break, since the dependency will always be resolved from the CLI package would always be the latest.

packages/angular/cli/models/schematic-engine-host.ts
clydin: The cdk as a set of schematic utility functions that import back to the core schematic packages so they need to be redirected as well.
I'll add a comment.

packages/angular/cli/models/schematic-engine-host.ts
clydin: `@angular/core` is actually functional in this regard.  `@angular/material` is, however, broken currently in those situations (with or without this PR).  The direct package install is a potential workaround for those cases but it can also cause other breakage since the object instances given to the schematic will be from the latest versions but the functions acting on them will be from an older version which might assume different internal behavior/structure so it's not an ideal solution.
An environment variable will be available (`NG_SCHEMATIC_REDIRECT=0`) which will allow the workaround to continue for those edge cases with old updates using new CLI versions.  Updates before version 8 do not have the update bootstrap feature so the changes in this PR will not apply.
This PR also provides the base capabilities to potentially remove the need for any workarounds by providing virtual compatibility modules for the deprecated and removed imports.

clydin: Minimal shim modules for the legacy deep imports into `@schematics/angular` are now available for the old first-party schematics that previously used them.  This should remove the need for the manual package install workaround in the latest CLI updating to old package versions case.

packages/angular/cli/models/schematic-engine-host.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    // All other schematics use default behavior
+    return super._resolveReferenceString(refString, parentPath);
+  }
+}
+
+/**
+ * Minimal shim modules for legacy deep imports of `@schematics/angular`
+ */
+const legacyModules: Record<string, unknown> = {
```
alan-agius4, line 81: Good call to add a shim

alan-agius4 (approved):

Splaktar:
Thank you!

PR #19869, opened 2021-01-25
feat(@schematics/angular): augment `universal` schematics to import `platform-server` shims

gkalpak:
In angular/angular#40559, a new `@angular/platform-server/shims` entry-point was introduced, which can be used to shim the server environment with APIs normally provided by the browser (such as DOM globals).

If the project is using a version of Angular (and thus `@angular/platform-server`) that is known to include this new entry-point, the newly generated `main.server.ts` file should import it at the top in order to ensure that the shims will be available as soon as possible (before other direct or transitive imports that may rely on them). See also angular/angular#40559 for more details.

NOTE:
This `universal` schematic is the base for `app-shell`, `@nguniversal/express-engine` and `@nguniversal/hapi-engine`.

### DIFF
--- a/packages/schematics/angular/universal/files/src/__main@stripTsExtension__.ts.template
+++ b/packages/schematics/angular/universal/files/src/__main@stripTsExtension__.ts.template
@@ -1,6 +1,15 @@
+/***************************************************************************************************
+ * Initialize the server environment - for example, adding DOM built-in types to the global scope.
+ *
+ * NOTE:
+ * This import must come before any imports (direct or transitive) that rely on DOM built-ins being
+ * available, such as `@angular/elements`.
+ */
+import '@angular/platform-server/init';
+
 <% if (hasLocalizePackage) { %>/***************************************************************************************************
  * Load `$localize` onto the global scope - used if i18n tags appear in Angular templates.
-*/
+ */
 import '@angular/localize/init';
 
 <% } %>import { enableProdMode } from '@angular/core';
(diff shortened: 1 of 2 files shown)


gkalpak:
This should probably wait for angular/angular#40559 to be merged and included in a release, but I created it early to discuss/get some guidance on what else I might need to change.

packages/schematics/angular/universal/index.ts
```diff
@@ -259,6 +261,7 @@ export default function (options: UniversalOptions): Rule {
         ...options as object,
         stripTsExtension: (s: string) => s.replace(/\.ts$/, ''),
         hasLocalizePackage: !!getPackageJsonDependency(host, '@angular/localize'),
+        hasPlatformServerShims: Number(ngMajorVersion) >= 12,
```
alan-agius4, line 264: This check shouldn't be needed, since Schematics v12 are only compatible with Angular FW v12.

gkalpak: Yeah. Basically, I left it here as a placeholder for disucssing whether we want to do it in a minor version (in which case we could do a `<major>.<minor>` check).

But if we want to introduce this in a major, then we can remove this.

gkalpak: Since, we missed the train for v11.2.x (and there probably won't be another minor before 12.0.0), I am going to remove this check.

alan-agius4: SGTM

alan-agius4 (approved):
Thanks LGTM

PR #19878, opened 2021-01-26
fix(@angular-devkit/schematics): Fix merge that causes an overwrite

### LINKED ISSUE #11337: MergeStrategy seems to be completely broken (mergeWith failure)
Re-post of https://github.com/angular/devkit/issues/745

```
- [ ] feature request
```

### Area
```
- [ ] devkit
- [x] schematics
```

### Versions
Node v8.11.1
NPM 5.8.0
Mac OS High Sierra

### Repro steps
Use the following schematic
```ts
import { Rule, url, apply, mergeWith, chain, MergeStrategy } from '@angular-devkit/schematics';

export function index(): Rule {
  return chain([
    simpleMerge(),
  ]);
}

function simpleMerge(): Rule {

  const source = url('./files');

  return mergeWith(source, MergeStrategy.Overwrite);
}
```

### Desired functionality
All files should be overwritten.

I tried every MergeStrategy possible (AllowOverwriteConflict, etc.), same result.

https://github.com/angular/devkit/issues/745
https://stackoverflow.com/questions/48957132/how-to-overwrite-file-with-angular-schematics

Brocco:
This fixes #11337 to allow for merging of a tree with another when the the file already exists in the tree being merged into.

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/host-tree.ts
+++ b/packages/angular_devkit/schematics/src/tree/host-tree.ts
@@ -56,7 +56,7 @@ export class HostDirEntry implements DirEntry {
     readonly path: Path,
     protected _host: virtualFs.SyncDelegateHost,
     protected _tree: Tree,
-  ) {}
+  ) { }
 
   get subdirs(): PathFragment[] {
     return this._host.list(this.path)
@@ -180,7 +180,7 @@ export class HostTree implements Tree {
         case 'c': {
           const { path, content } = action;
 
-          if ((this._willCreate(path) || this._willOverwrite(path))) {
+          if (this._willCreate(path) || this._willOverwrite(path) || this.exists(path)) {
             const existingContent = this.read(path);
             if (existingContent && content.equals(existingContent)) {
               // Identical outcome; no action required
--- a/packages/angular_devkit/schematics/src/tree/host-tree_spec.ts
+++ b/packages/angular_devkit/schematics/src/tree/host-tree_spec.ts
@@ -7,9 +7,47 @@
  */
 import { normalize, virtualFs } from '@angular-devkit/core';
 import { FilterHostTree, HostTree } from './host-tree';
+import { MergeStrategy } from './interface';
 
 describe('HostTree', () => {
+  describe('merge', () => {
+    it('should create files from each tree', () => {
+      const tree = new HostTree();
+      tree.create('/file1', 'a');
+      const tree2 = new HostTree();
+      tree2.create('/file2', 'a');
+      tree.merge(tree2);
+      expect(tree.actions[0].kind).toEqual('c');
+      expect(tree.actions[1].kind).toEqual('c');
+    });
 
+    it('should overwrite if the file exists in one tree', () => {
+      const tree = new HostTree();
+      tree.create('/file1', 'a');
+      const tree2 = new HostTree();
+      tree2.create('/file1', 'b');
+      tree.merge(tree2, MergeStrategy.Overwrite);
+      expect(tree.actions[0].kind).toEqual('c');
+    });
+
+    it('should throw if the file exists in one tree with the correct MergeStrategy', () => {
+      const tree = new HostTree();
+      tree.create('/file1', 'a');
+      const tree2 = new HostTree();
+      tree2.create('/file1', 'b');
+      expect(() => tree.merge(tree2)).toThrow();
+    });
+
+    it('should not have a second action if the file content is the same', () => {
+      const tree = new HostTree();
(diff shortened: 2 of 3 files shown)


packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -180,7 +180,7 @@ export class HostTree implements Tree {
         case 'c': {
           const { path, content } = action;
 
-          if ((this._willCreate(path) || this._willOverwrite(path))) {
+          if ((this._willCreate(path) || this._willOverwrite(path) || this.exists(path))) {
```
Brocco, line 183: This new check determines if the file already exists versus being created or overwritten.

packages/angular_devkit/schematics/src/tree/host-tree_spec.ts
```diff
@@ -2,14 +2,43 @@
  * @license
  * Copyright Google Inc. All Rights Reserved.
  *
- * Use of this source code is governed by an MIT-style license that can be
+ * Use of this source code is governed by an ZIT-style license that can be
```
alan-agius4, line 5: Kindly revert.

Brocco: Done, should've caught that myself... Thanks!

packages/angular_devkit/schematics/src/tree/host-tree_spec.ts
```diff
@@ -7,9 +7,38 @@
...
+    });
+
+    it('should overwrite if the file exists in one tree', () => {
+      const tree = new HostTree();
+      tree.create('/file1', '');
+      const tree2 = new HostTree();
+      tree2.create('/file1', '');
+      tree.merge(tree2, MergeStrategy.Overwrite);
+      expect(tree.actions[0].kind).toEqual('c');
+      expect(tree.actions[1].kind).toEqual('o');
```
clydin, line 31: Since the file content is the same, there should only be a single create action.
Can you also add a test with different content?

packages/angular_devkit/schematics/src/tree/host-tree_spec.ts
```diff
@@ -7,9 +7,38 @@
...
+    });
 
+    it('should throw if the file exists in one tree', () => {
+      const tree = new HostTree();
+      tree.create('/file1', '');
+      const tree2 = new HostTree();
+      tree2.create('/file1', '');
+      expect(() => tree.merge(tree2)).toThrow();
```
clydin, line 39: The content needs to be different for this to throw.

clydin (changes requested):
Looks like at least one E2E test is failing as well.
`node ./tests/legacy-cli/run_e2e.js ./tests/legacy-cli/e2e/tests/generate/component/component-duplicate`

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -180,7 +180,7 @@ export class HostTree implements Tree {
         case 'c': {
           const { path, content } = action;
 
-          if ((this._willCreate(path) || this._willOverwrite(path))) {
+          if ((this._willCreate(path) || this._willOverwrite(path) || (this.exists(path) && overwriteConflictAllowed))) {
```
clydin, line 183: I don't think the `overwriteConflichAllowed` check is needed here.
Looking at the failing E2E test (https://github.com/angular/angular-cli/blob/e7e01d213e9fb2473dfec4ec531e53f4c5822fe6/tests/legacy-cli/e2e/tests/generate/component/component-duplicate.ts), it appears that it only passed because it relied on the defect that is being fixed in this PR.  The test is creating the same exact component twice so the second should be a no-op (and not fail since it doesn't do anything).  To make the test work, something would need to be changed in one of the component's files after the first generate command so that there would be a legitimate conflict.

Brocco: Good call, I reverted the update to the exists logic and updated that test to update the contents of the app component ts file.

tests/legacy-cli/e2e/tests/generate/component/component-duplicate.ts
```diff
@@ -12,5 +13,9 @@ export default function () {
           in ${output.stdout}.`);
       }
     })
+    .then(() => {
+      const filePath = 'src/app/test-component/test-component.component.ts';
+      appendToFile(filePath, '\n// new content');
```
clydin, line 18:
```suggestion
      return appendToFile(filePath, '\n// new content');
```

clydin (approved):

PR #19908, opened 2021-01-30
fix(@angular-devkit/build-angular): fix aot mode for web worker

moczix:
i setup angular app in web-worker to use angular DI powers. It all worked well until i build production build. I got error about something bootstrap of undefined. I started investigate angular sources, and found this code. 
With aot flag set to false my angularApp inside web-worker with production build doesnt have ivy metadata like "emod".
Comparing this flag to other flags about aot i didnt find out why this flag is set to false. So i made if condition like other places


moczix:
@googlebot I signed it!

moczix:
@alan-agius4 can you review? or set the target? or did i do something wrong here?

alan-agius4:
@moczix, sorry for getting back to you so late.

I am not certain that we should allow this.

alan-agius4:
Hi @moczix, 

Thanks for your contribution, however this is the intended behaviour. 

Can you please elaborate your use case a bit more to why you need Angular DI in a worker?

Thanks.

moczix:
@alan-agius4 , i made app similar to your deprecated ui worker thread. But in my case i placed in worker only business logic, and it was cool to have angular DI in worker

PR #19920, opened 2021-02-01
fix(@schematics/angular): avoid unuse imports for canLoad guard generation

billyjov:


billyjov:
@googlebot I signed it!

alan-agius4 (changes requested):
Thanks for your contribution.


I think it's cleaner if you update https://github.com/angular/angular-cli/blob/0f61827dcf605b193d7d1a83d325e230cc973596/packages/schematics/angular/guard/index.ts#L43:L46

Also, can you kindly add a unit test here https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/guard/index_spec.ts?

Thanks.

billyjov:
@alan-agius4 sure.

is there a way to deal with this bazel error under Windows?

I ran into an issue by running unit tests: 

```

WARNING: Download from https://mirror.bazel.build/nodejs.org/dist/v12.14.1/node-v12.14.1-win-x64.zip failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
--

```

then 

```
no such package '@local_config_cc//'
```

Is there a way to fix the download ?

alan-agius4:
@billyjov, have you installed [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145) which is a prerequisite?

See: https://docs.bazel.build/versions/master/install-windows.html

packages/schematics/angular/guard/index.ts
```diff
@@ -39,11 +39,18 @@ export default function (options: GuardOptions): Rule {
...
     let implementationImports = `${options.implements.join(', ')}, `;
+    const commonImports = 'ActivatedRouteSnapshot, RouterStateSnapshot, ';
+
+    implementationImports = `${implementationImports}${commonImports}`;
     // As long as we aren't in IE... ;)
     if (options.implements.includes(GuardInterface.CanLoad)) {
       implementationImports = `${implementationImports}Route, UrlSegment, `;
     }
 
+    if (options.implements.length === 1 && options.implements[0] === GuardInterface.CanLoad) {
```
alan-agius4, line 50: IMHO, I think this is a bit unnecessarily complicated. Something like the berlow, is easier to follow and cleaner.


```ts
const routerSnapshotsNameImports = ['ActivatedRouteSnapshot', 'RouterStateSnapshot'];
const routerNamedImports: string[] = [...options.implements];

if (options.implements.includes(GuardInterface.CanLoad)) {
  routerNamedImports.push(
    'Route',
    'UrlSegment',
    ...(options.implements.length > 0 ? routerSnapshotsNameImports : []),
  );
} else {
  routerNamedImports.push(...routerSnapshotsNameImports);
}

routerNamedImports.sort();
const implementationImports = routerNamedImports.join(', ');

```

Also, kindly add unit test.

billyjov: Thanks you for pointing  this out..It's much cleaner.

packages/schematics/angular/guard/files/__name@dasherize__.guard.ts.template
```diff
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { <%= implementationImports %>ActivatedRouteSnapshot, RouterStateSnapshot, UrlTree } from '@angular/router';
+import { <%= implementationImports %>UrlTree } from '@angular/router';
```
alan-agius4, line 2:
```suggestion
import { <%= implementationImports %> } from '@angular/router';
```

packages/schematics/angular/guard/index.ts
alan-agius4: Can you try to run

```
yarn bazel test //packages/schematics/angular:angular_test
```

alan-agius4: Are you getting an error when running `yarn bazel test //packages/schematics/angular:angular_test`?

billyjov: Yep..here the full output

ng-cli-tests-output

alan-agius4: Can you try to rebase and check if the issue persist?

billyjov: I did a rebase and the errors does'nt appear anymore. 
by using `yarn bazel test //packages/schematics/angular:angular_test` i get now no errors.

i was expected to see some specs output in the console but the build only show success.

i tried to simulate some tests to failed, but i always get a success build... i'm missing something ? i went through the doc but found nothing to point me to the right way.

Thanks in advance

alan-agius4: > i was expected to see some specs output in the console but the build only show success

Only failed tests will appear in the terminal.

Note, the above command will only run the schematics angular tests. Which tests are amending?

billyjov: i change some tests inside Guard `index.spec.ts` just to see how a failed tests looks like. 
e.g

just change a line like : `expect(files).toContain('/projects/bar/src/app/foo/foo.guard.ts')` 

to : `expect(files).not.toContain('/projects/bar/src/app/foo/foo.guard.ts')`

and run the above command again. The build output always show success.

alan-agius4: I see, Bazel is running not in verbose failure mode. 

You can either check the errors in the log file which is pointed out in the above screenshot or run the test command with `--verbose_failures=true`.

billyjov: Ok thanks..i will take a look later on this..i'm AFK for the next 3 weeks.

billyjov: @alan-agius4 i had to switch to a linux subsystem to get tests running.

I just added some unit tests.

packages/schematics/angular/guard/files/__name@dasherize__.guard.ts.template
```diff
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { <%= implementationImports %>ActivatedRouteSnapshot, RouterStateSnapshot, UrlTree } from '@angular/router';
+import { <%= implementationImports %>, UrlTree } from '@angular/router';
```
alan-agius4, line 2: If `UrlTree` is always used shouldn't it be part of `routerSnapshotsNameImports`? so that it's also sorted?

packages/schematics/angular/guard/index_spec.ts
```diff
@@ -93,9 +91,8 @@ describe('Guard Schematic', () => {
 
   it('should respect the implements values', async () => {
     const implementationOptions = ['CanActivate', 'CanLoad', 'CanActivateChild'];
-    const options = { ...defaultOptions, implements: implementationOptions};
-    const tree = await schematicRunner.runSchematicAsync('guard', options, appTree)
-      .toPromise();
+    const options = { ...defaultOptions, implements: implementationOptions };
```
alan-agius4, line 94: Next time please avoid done unnecessary styling changes, as it caused chum and makes reviewing harding. Thanks.

packages/schematics/angular/guard/index_spec.ts
```diff
@@ -118,4 +114,46 @@ describe('Guard Schematic', () => {
...
     expect(fileString).not.toContain('canLoad');
   });
+
+  it('should add correct imports based on CanLoad implementation', async () => {
+    const implementationOptions = ['CanLoad'];
+    const options = { ...defaultOptions, implements: implementationOptions };
+    const tree = await schematicRunner.runSchematicAsync('guard', options, appTree).toPromise();
+    const fileString = tree.readContent('/projects/bar/src/app/foo.guard.ts');
+    const expectedImports = `import { CanLoad, Route, UrlSegment, UrlTree } from '@angular/router';`;
+    const wrongOrderImports = `import { UrlTree, CanLoad, Route, UrlSegment } from '@angular/router';`;
```
alan-agius4, line 124: The test for wrong imports here is tightly is not testing anything other than that the ordering is correct, which is already testing in the `expectedImports` hence, this makes it redundant.

packages/schematics/angular/guard/files/__name@dasherize__.guard.ts.template
billyjov: Yes it should. But can we find another name for the constant, because it's no longer just about snapshot imports? or names with snapshot as suffix?

packages/schematics/angular/guard/index_spec.ts
billyjov: yes, I will do that. In this case it was just wrongly formatted. See curly braces.

packages/schematics/angular/guard/files/__name@dasherize__.guard.ts.template
billyjov: @alan-agius4  I just notice `routerSnapshotsNameImports` is only for other implementations than `canLoad`. `UrlTree` is already sorted so it does'nt matter to add this to the array before sort.

So two options:

- Keep UrlTree in the template, because it is already sorted and always comes last
- Or we add an extra array. e.g. routerCommonImports, which only contains a single UrlTree element and is added to the routerNamedImports before sorting 

Which one should we go for ?

alan-agius4: I think adding in the array is more future proof. 

See an example below;

```
 const commonRouterNameImports = ['ActivatedRouteSnapshot', 'RouterStateSnapshot', 'UrlTree'];
```

packages/schematics/angular/guard/index_spec.ts
alan-agius4: Can you please remove the redundant expects? thanks

packages/schematics/angular/guard/files/__name@dasherize__.guard.ts.template
billyjov: done

packages/schematics/angular/guard/index_spec.ts
billyjov: done

alan-agius4 (approved):

PR #19929, opened 2021-02-02
fix(@angular-devkit/build-angular): remove `0.0.0` version from supported Angular semver versions error message

alan-agius4:


alan-agius4:
@clydin PTAL

clydin (approved):

PR #19932, opened 2021-02-02
docs: replace broken link

destus90:
There is a broken link inside `ng new` section (https://angular.io/cli/new). When you click on that link, placed next to `strict` option, you are redirected to 404 (Page not found).

### DIFF
--- a/packages/schematics/angular/ng-new/schema.json
+++ b/packages/schematics/angular/ng-new/schema.json
@@ -127,7 +127,7 @@
       "x-user-analytics": 14
     },
     "strict": {
-      "description": "Creates a workspace with stricter type checking and stricter bundle budgets settings. This setting helps improve maintainability and catch bugs ahead of time. For more information, see https://angular.io/strict",
+      "description": "Creates a workspace with stricter type checking and stricter bundle budgets settings. This setting helps improve maintainability and catch bugs ahead of time. For more information, see https://angular.io/guide/strict-mode",
       "x-prompt": "Do you want to enforce stricter type checking and stricter bundle budgets in the workspace?\n  This setting helps improve maintainability and catch bugs ahead of time.\n  For more information, see https://angular.io/strict",
       "type": "boolean",
       "default": false,


alan-agius4:
Thanks for this. 

https://angular.io/strict is a redirection to https://angular.io/guide/strict-mode both of which are valid links.

I think the problem is that in the docs the redirection is not happening because both hosts are the same and therefore routing will be handled by Angular when the link is in angular.io.

Let me have a chat with the docs-infra folks before doing any changes.

packages/schematics/angular/ng-new/schema.json
```diff
@@ -127,8 +127,8 @@
       "x-user-analytics": 14
     },
     "strict": {
-      "description": "Creates a workspace with stricter type checking and stricter bundle budgets settings. This setting helps improve maintainability and catch bugs ahead of time. For more information, see https://angular.io/strict",
-      "x-prompt": "Do you want to enforce stricter type checking and stricter bundle budgets in the workspace?\n  This setting helps improve maintainability and catch bugs ahead of time.\n  For more information, see https://angular.io/strict",
+      "description": "Creates a workspace with stricter type checking and stricter bundle budgets settings. This setting helps improve maintainability and catch bugs ahead of time. For more information, see https://angular.io/guide/strict-mode",
+      "x-prompt": "Do you want to enforce stricter type checking and stricter bundle budgets in the workspace?\n  This setting helps improve maintainability and catch bugs ahead of time.\n  For more information, see https://angular.io/guide/strict-mode",
```
alan-agius4, line 131: Following a convo with @gkalpak, we do this change, but only for the "description" field.

The problem only effects angular.io, and prompts description are not displayed in AIO.

Hence, can you please revert this?



```suggestion
      "x-prompt": "Do you want to enforce stricter type checking and stricter bundle budgets in the workspace?\n  This setting helps improve maintainability and catch bugs ahead of time.\n  For more information, see https://angular.io/strict",
```

destus90: Done

alan-agius4 (approved):
Thanks for this.

PR #19940, opened 2021-02-03
fix(@angular/cli): resolve migration collections from containing package

clydin:
This change ensures that migration collections defined in a package's `ng-update` metadata are resolved from the containing package.  This avoids issues that may arise from package manager hoisting behavior which can result in the wrong migration collection being chosen.  The `--migrate-only` option already contained this resolution logic and now the full update contains it as well.


alan-agius4 (approved):
LGTM.

PR #19956, opened 2021-02-03
ci: remove integration job

alan-agius4:
This is currently redundant and hence we should disable it.


alan-agius4:
@kyliau can we please remove `ci/circleci: integration` from required steps?

kyliau (approved):

clydin (approved):

PR #19970, opened 2021-02-04
fix(@angular/cli): temporarily limit npm to version 6

clydin:
This change will display an error message if using npm 7 (or versions earlier than 6). This is a temporarily change while npm 7 usability concerns are addressed.


kyliau (approved):
LGTM, +1 to the test!

PR #19980, opened 2021-02-05
feat(@angular-devkit/build-angular): add `postcss-preset-env` with stage 3 features

alan-agius4:
With this change we add `postcss-preset-env` with stage 3 features. This stage includes support for:

 - all property
 - break properties
 - custom properties
 - font-variant property
 - gap properties
 - media query ranges

See https://preset-env.cssdb.org/features#stage-3


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -222,7 +224,12 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
           extracted,
         }),
         ...extraPostcssPlugins,
-        autoprefixer(),
+        postcssPresetEnv({
```
clydin, line 227: It looks like `postcss-preset-env` also has a `browsers` option.  Can this be set as well?  This would ideally ensure that all the Browserslist usage is synchronized.

clydin (changes requested):

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -222,7 +224,14 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
...
+          // tslint:disable-next-line: no-any
+          browsers: supportedBrowsers as any, // Typinmgs only allow a string
```
clydin, line 229:
```suggestion
          browsers: supportedBrowsers as any, // Typings only allow a string
```
small typo

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -222,7 +224,14 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
...
+          autoprefixer: {
+            overrideBrowserslist: supportedBrowsers,
```
clydin, line 231: Looks like it will automatically set this if the `browsers` option is used:
https://github.com/csstools/postcss-preset-env/blob/d7652b1e6196e8f55bf3f0aac4ac090fec7ed54e/src/postcss.js#L26

clydin (approved):

PR #19982, opened 2021-02-05
fix(@angular-devkit/build-angular): display the code-frame for localization errors

petebacondarwin:
Previously if there was an error processing the AST in the Babel plugin that does
the replacement of i18n messages, the error shown did not provide much help
in where in the code the problem appears.

Now we display a "code-frame" representation of the position in the parsed
source-code where the AST processing failed.

Partially resolves https://github.com/angular/angular/issues/40702#issuecomment-773307996


petebacondarwin:
Probably easiest to review the changes with whitespace matching turned off.

petebacondarwin:
I totally forgot about this. I see it is conflicted and I set it to WIP - perhaps because of the Windows failures...
I'll look through it tomorrow and either update it or close it. Thanks for the prod.

petebacondarwin:
This is stale and would require a `PathManipulation` interface to be provided, which would create a lot of churn in this code for little benefit. Closing.

PR #19995, opened 2021-02-07
fix(@angular-devkit/build-angular): the root Tailwind configuration f…

ahasall:
…ile is always picked

 A configuration file in the project root should take precedence over one in the workspace root, but it's not currently the case.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -166,6 +166,7 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
     const fullPath = path.join(basePath, tailwindConfigFile);
     if (fs.existsSync(fullPath)) {
       tailwindConfigPath = fullPath;
+      break;
     }
   }
   // Only load Tailwind CSS plugin if configuration file was found.


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -162,7 +162,7 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
   // A configuration file can exist in the project or workspace root
   const tailwindConfigFile = 'tailwind.config.js';
   let tailwindConfigPath;
-  for (const basePath of [wco.projectRoot, wco.root]) {
+  for (const basePath of [wco.root, wco.projectRoot]) {
     const fullPath = path.join(basePath, tailwindConfigFile);
     if (fs.existsSync(fullPath)) {
```
alan-agius4, line 167: The problem here is that there is no break statement.

alan-agius4 (changes requested):
Thanks for this, see comment above.

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
ahasall: Done!

alan-agius4 (approved):
Thanks

PR #19997, opened 2021-02-08
refactor(@angular-devkit/core): refactor NodeJsAsyncHost to use promisify

alan-agius4:


alan-agius4:
Done as part https://github.com/angular/angular-cli/pull/19998

PR #20007, opened 2021-02-09
Remove experimental rollup pass

### LINKED ISSUE #15836: --experimentalRollupPass feedback issue
Heya all, 

In https://github.com/angular/angular-cli/pull/15690 we introduced a new experimental flag, --`--experimentalRollupPass`, that attempts an advanced optimization pass. It's marked experimental because we don't expect it to work out of the box for all projects. 

If you try it and it causes your project to fail to build, or to fail at runtime, please provide us with information here so we can try to address it.

The optimization aims at improving tree-shaking and dead code elimination efficiency for applications that use code splitting and code lazy loading. In our early experiments with some of our apps we've seen significant XX% code size reduction (depending on the app and various factors).

If you do experiment with this setting please post before and after payload size numbers + the output of `ng version`. thanks!

Docs: https://v9.angular.io/cli/build

alan-agius4:
**refactor(@angular-devkit/build-angular): remove experimental rollup pass**

The experimental rollup pass has significant issues with the new Ivy webpack plugin. It also, didn't produce as well as we hoped in real world scenarios infact in many cases this caused build to fail.

REAKING CHANGE:

The experimental rollup pass `--experimental-rollup-pass` option has been removed.

Closes #15836

**feat(@schematics/angular): add migration to remove deprecated options from 'angular.json'**

**refactor(@ngtools/webpack): remove export of `VirtualFileSystemDecorator` type**

This was only needed for the rollup loader.


clydin (approved):

Splaktar:
@alan-agius4 There is a typo in the commit that might cause this breaking change to not be reflected in the CHANGELOG:

```
REAKING CHANGE:
```
Instead of 
```
BREAKING CHANGE:
```
Indeed, I don't see this breaking change in https://github.com/angular/angular-cli/releases/tag/v12.0.0-next.0 where the commit was included.

PR #20036, opened 2021-02-11
feat(@angular-devkit/build-angular): tailwindcss purge

### LINKED ISSUE #20015: Tailwind support
@clydin few things that maybe will help you ensure full TailwindCSS support, based on issues we had in `ngneat/tailwind`:

1. Tailwind uses PurgeCSS under the hood, which is enabled when `process.env.NODE_ENV` is `production`, which Angular not setting by default.
https://tailwindcss.com/docs/optimizing-for-production#removing-unused-css
https://twitter.com/Nartc1410/status/1357504574479810568
2. Tailwind dark mode directives won't work correctly in component styles with view encapsulation other than `none`, as tailwind will just try to prepend class with `.dark`, so the result code will be something like `.dark[_ng-content-**] .value[_ng-content-**]`. Also not sure what will happen with `ShadowDom`, tested only with `emulated`. In `ngneat/tailwind` we made PostCSS plugin for it: https://github.com/vltansky/postcss-ng-tailwind-in-components. Probably there are alternative solutions.
https://github.com/ngneat/tailwind/issues/54

relevant to: https://github.com/angular/angular-cli/commit/73b409881f71a8235769a345356dcde3c568d0c3

vltansky:
closes #20015

setting `tailwindConfig.purge.enabled` based on `buildOptions.optimization.styles.minify`.

Would be happy for guidance from @clydin to finish this PR if it goes in the right direction. Or feel free to close PR and implement it yourself :)

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -186,7 +186,27 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
       );
     }
     if (tailwindPackagePath) {
-      extraPostcssPlugins.push(require(tailwindPackagePath)({ config: tailwindConfigPath }));
+      const tailwindConfig = require(tailwindConfigPath);
+      if (typeof tailwindConfig !== 'function') {
+        if (!tailwindConfig.purge) {
+          tailwindConfig.purge = {};
+        }
+        if (Array.isArray(tailwindConfig.purge)) {
+          tailwindConfig.purge = {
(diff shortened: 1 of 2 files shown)


vltansky:
I've added a check to ensure that the config doesn't export a function, so we can set the `purge.enabled` option.
I think it is fine, as even official [Tailwind VSCode extension](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss) won't work if config exports a function (see https://github.com/ngneat/tailwind/pull/39)

vltansky:
I guess this PR is not relevant anymore

PR #20037, opened 2021-02-12
feat(@ngtools/webpack): support Webpack 5

clydin:
The `@ngtools/webpack` package now officially supports Webpack 5.
It is also now built against Webpack 5 types.  Webpack 4 support is temporarily maintained while the remainder of the tooling is transitioned.


packages/ngtools/webpack/src/ivy/system.ts
```diff
@@ -9,11 +9,19 @@ import * as ts from 'typescript';
 import { InputFileSystem } from 'webpack';
 import { externalizePath } from './paths';
 
+export interface InputFileSystemWithSync extends InputFileSystem {
```
alan-agius4, line 12: Consider naming this `InputFileSystemSync` similar to standard FS methods.

alan-agius4 (approved):

PR #20047, opened 2021-02-12
Switch angular.io from HTTP to HTTPS

aravindvnair99:
Since angular.io/analytics supports HTTPS, why not use that instead of HTTP?


alan-agius4 (commented):
Can you please change the commit message to 
```
docs(@angular/cli): switching angular.io URL from HTTP to HTTPS 
```

alan-agius4 (approved):
Thanks for your contribution.

PR #20076, opened 2021-02-17
fix(@angular-devkit/build-angular): improve fidelity of linker sourcemaps

clydin:
This change uses Webpack's sourcemap processing to merge sourcemaps generated from the customized babel loader used to perform the partial compilation linking in applications.
Webpack's sourcemap processing supports identity mappings which allow character to character mappings to the original source.


packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
```diff
@@ -144,5 +148,19 @@ export default custom<AngularCustomOptions>(() => {
...
+          this.resourcePath,
+          result.map,
+          source,
+          inputSourceMap as string,
+          true,
+        ).map();
+      }
+
+      return result;
+    },
```
petebacondarwin, line 164: This didn't help in my scenario. But this did:

```js
        result(result, { map: inputSourceMap }) {
            if (result.map && inputSourceMap) {
              const fs = require('fs');
              const path = require('path');
              const sourcemaps = require('@angular/compiler-cli/src/ngtsc/sourcemaps');

              const fileSystem = {
                resolve: path.resolve,
                exists: fs.existsSync,
                dirname: path.dirname,
                relative: path.relative,
                readFile: fs.readFileSync,
              };
              const loader = new sourcemaps.SourceFileLoader(fileSystem, {}, {});
              const outputSource = loader.loadSourceFile(this.resourcePath, result.code, {map: result.map, path: this.resourcePath + '.map'});
              result.map = outputSource.renderFlattenedSourceMap()
            }
            return result;
        },
```

petebacondarwin: But this is quite resource intensive and ends up running for a large number of files, not just the ones that have been linked.

PR #20084, opened 2021-02-18
Webpack 5

clydin:
WIP integration test PR for default Webpack 5 support

### DIFF
--- a/package.json
+++ b/package.json
@@ -226,7 +226,7 @@
     "typescript": "4.2.3",
     "verdaccio": "4.12.0",
     "verdaccio-auth-memory": "^10.0.0",
-    "webpack": "5.21.2",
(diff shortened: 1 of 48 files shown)


thw0rted:
Hi @clydin , over at #20359 they said to track this PR to see when official Webpack 5 support will land, so I can stop ignoring the mismatched peer-dep error in my `npm install`.  Does closing this branch mean that it's no longer being worked? Is there another PR we should be tracking?  (If closing the PR is a normal part of the workflow, no problem!)

clydin:
The elements of this PR were reorganized and became part of #20466 which has been merged.
The next prerelease of v12 (due this week) will officially use Webpack 5 and no manual installation of Webpack 5 will be required.

PR #20096, opened 2021-02-19
feat(@angular-devkit/build-angular): enable inlineCritical by default

alan-agius4:
BREAKING CHANGE:

Critical CSS inlining is now enabled by default. If you wish to turn this off set `inlineCritical` to `false`.

See: https://angular.io/guide/workspace-config#optimization-configuration

### DIFF
--- a/packages/angular/cli/lib/config/schema.json
+++ b/packages/angular/cli/lib/config/schema.json
@@ -740,7 +740,7 @@
                             "inlineCritical": {
                               "type": "boolean",
                               "description": "Extract and inline critical CSS definitions to improve first paint time.",
-                              "default": false
+                              "default": true
                             }
                           },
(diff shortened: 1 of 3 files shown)


rhutchison:
@alan-agius4 inlineCritical doesn't play well with tailwind @apply rules.  Has there been anything raised or discussed internally regarding this issue?

What I've noticed is that the css that gets pushed inline has not gone through the tailwind pre-processor.

alan-agius4:
@rhutchison, no one brought up any issues with inlining of critical CSS. That said, the inlining of critical CSS is done post processing the stylesheets.

If there is any issue please file an issue with a reproduction in the issue tracker. Thanks.

clydin (approved):

PR #20098, opened 2021-02-19
docs: fix build-system-overview graph

Jefiozie:
When I was looking at different documents I found out that graph wasn't working anymore. 
This PR should resolve this, I don't believe I've seen a issue around this.


alan-agius4 (approved):
Thanks LGTM

PR #20128, opened 2021-02-24
feat(@schematics/angular): production builds by default

### LINKED ISSUE #14471: Default to production builds
# 🚀 Feature request
I know this is tough due to it being a breaking change with the ecosystem, but I think `ng build` should do a production build by default.

```
- [x] build
```

### Description
Most of the time, developers should be building for production, and serving for development. It would save a lot of time overall if `ng build` defaulted to production mode.

During the update process to 9 (or some other version), prompt the user if they want to keep the current behavior or migrate to the new behavior, and then default all new projects with `ng new` to production by default.

None

alan-agius4:
With this change we do several changes to the `angular.json` configuration for `build` , `server` and `app-shell` targets.

- build, server and app-shell targets are configured to run production by default.
- We add a new configuration named `development` to run the mentioned builder targets in development. Ex: `ng build --configuration development`.
- When adding `universal` or `app-shell`, we generate the full set of configurations as per the `buiid` target. Previously, we only generated the `production` configuration.
- We added a helper script in `package.json` to run build in watch mode. `npm run watch` which is a shortcut for `ng build --watch --configuration development`
- We deprecated the `--prod` command line argument. This argument is confusing especially to new users, since users expect that this builds an application in production mode. This however, is only an alias for `--configuration="production"`

Closes #14471

### DIFF
--- a/packages/schematics/angular/app-shell/index.ts
+++ b/packages/schematics/angular/app-shell/index.ts
@@ -155,7 +155,7 @@ function addUniversalTarget(options: AppShellOptions): Rule {
 }
 
 function addAppShellConfigToWorkspace(options: AppShellOptions): Rule {
-  return () => {
+  return (host, context) => {
     if (!options.route) {
       throw new SchematicsException(`Route is not defined`);
     }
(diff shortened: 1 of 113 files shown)


destus90:
@alan-agius4 
Shouldn't be added a separator `--` after `npm run build` in order to pass additional options? I guess 
`npm run buid -- --watch --configuration development`

alan-agius4:
@destus90, yeah indeed, thanks for pointing that out.

alan-agius4:
@googlebot ping!

packages/schematics/angular/universal/index.ts
```diff
@@ -35,62 +36,65 @@ import { findBootstrapModuleCall, findBootstrapModulePath } from '../utility/ng-
...
+          extractLicenses: options?.extractLicenses,
+        };
+      };
+
+      const buildTarget = clientProject.targets.get('build');
+      if (buildTarget?.options) {
+        buildTarget.options.outputPath = `dist/${options.clientProject}/browser`;
+      }
+
+      const buildConfigurations = clientProject.targets.get('build')?.configurations;
```
clydin, line 72:
```suggestion
      const buildConfigurations = buildTarget?.configurations;
```

tests/legacy-cli/e2e/tests/build/lazy-load-syntax.ts
```diff
@@ -75,16 +74,17 @@ export default async function () {
   // This way we can use `ng e2e` to test JIT and `ng e2e --prod` to test AOT.
   await updateJsonFile('angular.json', json => {
     const buildTarget = json['projects'][projectName]['architect']['build'];
-    buildTarget['options']['aot'] = false;
-    buildTarget['configurations']['production'] = { aot: true };
+    buildTarget['options']['aot'] = true;
+    buildTarget['options']['aot'] = true;
```
clydin, line 78: Duplicate statement?

tests/legacy-cli/e2e/tests/build/multiple-configs.ts
```diff
@@ -65,7 +72,7 @@ export default async function () {
   await expectFileToExist('dist/test-project/main.js.map');
   await expectToFail(() => expectFileToExist('dist/test-project/vendor.js'));
   // Use three configurations and a override, and prod at the end.
-  await ng('build', '--configuration=one,two,three', '--vendor-chunk=false', '--prod');
+  await ng('build', '--configuration=one,two,three', '--vendor-chunk=false');
```
clydin, line 75: The comment above should probably be changed as well since `--prod` is not at the end anymore.

clydin (changes requested):

packages/schematics/angular/app-shell/index.ts
```diff
@@ -155,7 +155,7 @@ function addUniversalTarget(options: AppShellOptions): Rule {
 }
 
 function addAppShellConfigToWorkspace(options: AppShellOptions): Rule {
-  return () => {
+  return (host, content) => {
```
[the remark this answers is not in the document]
alan-agius4: oops

clydin (approved):

Issue #18398, opened 2020-07-29
out of memory on ng serve with i18n and targeting es5


nomeaning777:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the previous version in which this bug was not present was: 9.1.12

### Description

If the Angular project satisfies the following conditions, the memory usage of `ng serve` is excessive
 * it uses i18n with `@angular/localize`
 * it is configure to target ES5
 * it loads large object from external library.

The issue doesn't happen if I changed the target to ES2015 or disabled i18n.
`ng build` and `ng serve --no-source-map` is also not affected.

## 🔬 Minimal Reproduction
 
1. `ng add @angular/localize`
2. create a translation file and add configuration to `angular.json`.
3. set target in `tsconfig.base.json` to `"es5"` and enable IE11 in browserlist.
4. add external npm library which contains large object and use it from a component.
    * I can reproduce with `encoding-japanese` and `text-encoding`.
5. run `ng serve` and it consumes a lot of memory.

reproduction: https://github.com/nomeaning777/ng-serve-memory-issue-reproduction

## 🔥 Exception or Error

`ng serve` failed with OOM Error. 
```
$ ng serve
Compiling @angular/core : es2015 as esm2015
Compiling @angular/compiler/testing : es2015 as esm2015
Compiling @angular/animations : es2015 as esm2015
Compiling @angular/common : es2015 as esm2015
Compiling @angular/core/testing : es2015 as esm2015
Compiling @angular/animations/browser : es2015 as esm2015
Compiling @angular/platform-browser : es2015 as esm2015
Compiling @angular/common/http : es2015 as esm2015
Compiling @angular/animations/browser/testing : es2015 as esm2015
Compiling @angular/common/testing : es2015 as esm2015
Compiling @angular/router : es2015 as esm2015
Compiling @angular/forms : es2015 as esm2015
Compiling @angular/platform-browser-dynamic : es2015 as esm2015
Compiling @angular/platform-browser/testing : es2015 as esm2015
Compiling @angular/platform-browser/animations : es2015 as esm2015
Compiling @angular/common/http/testing : es2015 as esm2015
[...]

PR #20172, opened 2021-03-01
refactor(@angular-devkit/build-angular): use custom babel loader for i18n dev-server support

clydin:
The custom babel loader allows files to be conditionally processed by the i18n inlining transforms based on both file path and content.
By allowing content based checks, the entire babel parse/transform/print process can be skipped for files that do not contain localizations.

Closes: #18398


alan-agius4 (approved):
I think this might solve https://github.com/angular/angular-cli/issues/18398 can you please check?

DmitryEfimenko:
I noticed that this commit has the following tags: `v12.0.0-next.5`  `v12.0.0-next.4` `v12.0.0-next.3`

Is there a breaking change in this commit that prevents it to be released in an earlier version?

PR #20173, opened 2021-03-01
test(@angular-devkit/build-angular): improve reliability of app-shell server-worker test

clydin:
The test now waits until the express server is listening and uses a try/finally block to ensure the server is shutdown.
The test server also only now binds to localhost and lets the OS choose a port. This provides better support for running multiple tests in parallel.


alan-agius4 (approved):

alan-agius4:
Maybe this should also be merged to the patch branch?

PR #20190, opened 2021-03-03
fix(@angular-devkit/schematics): UpdateRecorder should support multiple adjacent inserts

kyubisation:
At the current time this is more of a question than a fix.
Should the update recorder allow multiple adjacent inserts? I added a test to illustrate this question.

While I do have a potential fix for this, in its current form it would switch to pure string handling and I assume the Buffer usage was done due to performance?

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/recorder_spec.ts
+++ b/packages/angular_devkit/schematics/src/tree/recorder_spec.ts
@@ -30,6 +30,18 @@ describe('UpdateRecorderBase', () => {
     expect(result.toString()).toBe('Hello beautiful World');
   });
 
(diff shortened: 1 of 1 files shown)


dgp1130:
Hey there, looks like this PR fell through the cracks and we didn't get to it until just now. This seems like more of an issue rather than a PR we can easily land (since we don't want to merge failing tests), so I opened #21110 for discussion and tracking there.

If you have a specific fix in mind, feel free to open a PR with that and we can help review and hopefully land that fix.

kyubisation:
@dgp1130 That's fine. I was more curious if this _should_ work in your eyes. Thank you for creating the related issue. 👍

PR #20194, opened 2021-03-04
 feat(@angular-devkit/architect): add implementation for defaultConfiguration

alan-agius4:
With this change, the architect can be configured to use a default configuration when it's not provided as part of the target.

Consider the below, where `defaultConfiguration` is configured to `production`. Running `ng build` will be invoked with "production" configuration.

```js
"build": {
  "defaultConfiguration": "production",
  "options": {
      ...
  },
  "configurations": {
    "production": {
        ...
    }
  }
}
```

### DIFF
--- a/tests/legacy-cli/e2e/tests/misc/target-default-configuration.ts
+++ b/tests/legacy-cli/e2e/tests/misc/target-default-configuration.ts
@@ -23,8 +23,8 @@ export default async function () {
   // Add new configuration and set "defaultConfiguration"
   await updateJsonFile('angular.json', workspace => {
     const build = workspace.projects['test-project'].architect.build;
-    build.defaultConfiguration = 'defaultConfiguration';
-    build.configurations.defaultConfiguration = {
+    build.defaultConfiguration = 'foo';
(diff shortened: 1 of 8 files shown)


tests/legacy-cli/e2e/tests/misc/target-default-configuration.ts
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  await ng('build');
+  await expectFileToExist('dist/test-project/main.js');
+  await expectFileToExist('dist/test-project/main.js.map');
+
+  // Add new configuration and set "defaultConfiguration"
+  await updateJsonFile('angular.json', workspace => {
+    const build = workspace.projects['test-project'].architect.build;
+    build.defaultConfiguration = 'defaultConfiguration';
```
clydin, line 26: Can you use a different name for the configuration (`my-default`, `foo`, etc.) ?  This may be looked at as an example and the configuration having the same name as the property may cause confusion.

clydin (approved):

PR #20205, opened 2021-03-05
Webpack styles configuration refactor and cleanup

clydin:
See individual commits


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -312,9 +211,130 @@ export function getStylesConfig(wco: WebpackConfigOptions) {
     }
   }
 
+  // Rule for all supported style types
+  const styleRule: webpack.RuleSetRule = {
+    test: /\.(?:css|scss|sass|less|styl)$/,
+    rules: [
+      // Setup processing rules for global and component styles
+      {
+        oneOf: [
```
alan-agius4, line 220: I wasn't aware that this was possible.

alan-agius4 (approved):
Nice cleanup 👍

PR #20212, opened 2021-03-07
fix (@angular-devkit/build-angular)

pavelvlodarczyk:
wraps ES5 scripts in an IIFE

Inspired by @clydin: This change ensures that classic (ES5) script's top-level function helpers do not get overwritten by other scripts top-level functions that happen to have the same name.


alan-agius4:
Hi @pavelvlodarczyk, 

Thanks for this, but Angular CLI version 9 is currently LTS and only critical and security fixes are merged. Seeing that this doesn't fall in neither of the above categories it cannot be merged.

Kindly update to a more recent version of the Angular CLI to avail from the fix, thanks.

More information about Angular releases and LTS policy can be found here https://angular.io/guide/releases

PR #20236, opened 2021-03-11
perf(@ngtools/webpack): avoid adding transitive dependencies to Webpack's dependency graph

clydin:
This change augments a TypeScript Compiler Host's resolveModuleNames function to collect dependencies of the containing file based on the module names passed to the resolveModuleNames function. This process assumes that consumers of the Compiler Host will call resolveModuleNames with modules that are actually present in a containing file.  The TypeScript compiler exhibits such behavior making this process effective at generating a set of all direct dependencies for a given source file.
This process is a workaround for gathering a TypeScript SourceFile's dependencies as there is no currently exposed public method to do so. A BuilderProgram does have a `getAllDependencies` function. However, that function returns all transitive dependencies as well which can cause excessive Webpack rebuilds especially in larger programs.

### DIFF
--- a/packages/ngtools/webpack/src/ivy/host.ts
+++ b/packages/ngtools/webpack/src/ivy/host.ts
@@ -90,6 +90,78 @@ function augmentResolveModuleNames(
   }
 }
 
+/**
+ * Augments a TypeScript Compiler Host's resolveModuleNames function to collect dependencies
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):

alan-agius4:
@clydin, can you please create a path version as it doesn't merge cleanly

Thanks.

PR #20237, opened 2021-03-11
perf(@ngtools/webpack): avoid full compiler lazy route analysis on JIT rebuilds

clydin:
The deprecated lazy route string form requires code analysis prior to each build to locate any lazy routes and connect them with Webpack.
This change uses the previous fast but less accurate AST based lazy route discovery for JIT rebuilds. The first build will still perform a full Angular compiler analysis to ensure an accurate list of all initial deprecated string form lazy routes.

### DIFF
--- a/packages/ngtools/webpack/src/ivy/plugin.ts
+++ b/packages/ngtools/webpack/src/ivy/plugin.ts
@@ -16,6 +16,7 @@ import {
   NormalModuleReplacementPlugin,
   compilation,
 } from 'webpack';
+import { findLazyRoutes } from '../lazy_routes';
 import { NgccProcessor } from '../ngcc_processor';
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -547,39 +549,61 @@ export class AngularWebpackPlugin {
...
+        return this.createFileEmitter(builder, transformers, () => []);
+      });
+      const analyzingFileEmitter: FileEmitter = async (file) => {
+        const innerFileEmitter = await pendingAnalysis;
+
+        return innerFileEmitter(file);
+      };
+
+      if (this.watchMode) {
+        this.ngtscNextProgram = angularProgram;
```
JoostK, line 580: Should this be dropped now?

PR #20241, opened 2021-03-11
fix(@schematics/angular): remove references to the prod flag

cexbrayat:
As the `--prod` is now deprecated, and `ng build` uses by default the `production` configuration,
then we should remove the references to `ng build --prod` in the generated project.


alan-agius4 (approved):

alan-agius4:
Thanks

PR #20243, opened 2021-03-11
feat(@schematics/angular): update several TypeScript compilation target (Syntax)

alan-agius4:
With this change we update the TypeScript target of server compilation unit for new projects.

- E2E target has been update from `es2018` to `es2019`
- Browser target has been update from `es2015` to `es2017`
- Server target has been update from `es2016` to `es2019`

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/specs/lazy-module_spec.ts
+++ b/packages/angular_devkit/build_angular/src/browser/specs/lazy-module_spec.ts
@@ -131,8 +131,6 @@ describe('Browser Builder lazy modules', () => {
       'src/lazy-module.ts': 'export const value = 42;',
       'src/main.ts': `import('./lazy-module');`,
     });
-    // Using `import()` in TS require targetting `esnext` modules.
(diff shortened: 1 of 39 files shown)


packages/angular_devkit/build_angular/src/browser/specs/lazy-module_spec.ts
```diff
@@ -132,7 +132,7 @@ describe('Browser Builder lazy modules', () => {
       'src/main.ts': `import('./lazy-module');`,
     });
     // Using `import()` in TS require targetting `esnext` modules.
-    host.replaceInFile('src/tsconfig.app.json', `"module": "es2015"`, `"module": "esnext"`);
+    host.replaceInFile('src/tsconfig.app.json', `"module": "es2017"`, `"module": "esnext"`);
```
clydin, line 135: This entire line shouldn't be needed anymore since new projects set the `module` option to `es2020`.
Same for the other changed line in this file below.

clydin (approved):

PR #20250, opened 2021-03-11
perf(@ngtools/webpack): use precalculated dependencies in unused file check

clydin:
This change uses the newly introduced precalculated file dependencies for each TypeScript file instead of querying TypeScript for the SourceFile's dependencies when performing the unused file check at the end of the build cycle. This change removes the need to recalculate the dependencies for each TypeScript file present in the Webpack compilation.

### DIFF
--- a/packages/ngtools/webpack/src/ivy/plugin.ts
+++ b/packages/ngtools/webpack/src/ivy/plugin.ts
@@ -137,7 +137,7 @@ export class AngularWebpackPlugin {
       // Example: module -> module__ivy_ngcc
       resolverFactoryHooks.resolveOptions
         .for('normal')
-        .tap(PLUGIN_NAME, (resolveOptions: { mainFields: string[], plugins: unknown[] }) => {
+        .tap(PLUGIN_NAME, (resolveOptions: { mainFields: string[]; plugins: unknown[] }) => {
           const originalMainFields = resolveOptions.mainFields;
           const ivyMainFields = originalMainFields.map((f) => `${f}_ivy_ngcc`);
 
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -301,6 +298,26 @@ export class AngularWebpackPlugin {
     });
   }
 
+  private markResourceUsed(
+    normalizedResourcePath: string,
+    currentUnused: Set<string>,
+    visitedResources = new Set<string>(),
```
JoostK, line 304: This function is called for all Webpack modules without a visited set, so individual modules won't realize that they, or any of its transitive deps, could have already been visited when traversing another module.

clydin: good call.  updated.

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -301,6 +299,26 @@ export class AngularWebpackPlugin {
     });
   }
 
+  private markResourceUsed(
+    normalizedResourcePath: string,
+    currentUnused: Set<string>,
+    visitedResources: Set<string>,
+  ): void {
+    if (visitedResources.has(normalizedResourcePath)) {
```
alan-agius4, line 307: Why do we need to keep track of `visitedResources` separately? We can use `currentUnused` as our single source of truth.

```ts
if (!currentUnused.has(normalizedResourcePath)) {
    return;
}

currentUnused.delete(normalizedResourcePath);

const dependencies = this.fileDependencies.get(normalizedResourcePath);
if (!dependencies) {
    return;
}

for (const dependency of dependencies) {
    this.markResourceUsed(normalizePath(dependency), currentUnused);
}
```

JoostK: I think Alan is correct; there is a risk for Webpack modules outside of the set of non-declaration source files, as cycles between those files wouldn't be detected through `currentUnused`, but this should not be a problem in practicle as `this.fileDependencies` only tracks deps between source files in the program.

clydin: I tend to be wary of multi-purpose situations since a future change of one part may unintentionally break an invariant of the second.  But in this case both are localized and very specific in purpose so I think this makes sense.

alan-agius4 (approved):

PR #20251, opened 2021-03-11
[11.2.x] perf(@ngtools/webpack): use precalculated dependencies in unused file check

clydin:
This change uses the newly introduced precalculated file dependencies for each TypeScript file instead of querying TypeScript for the SourceFile's dependencies when performing the unused file check at the end of the build cycle. This change removes the need to recalculate the dependencies for each TypeScript file present in the Webpack compilation.

### DIFF
--- a/packages/ngtools/webpack/src/ivy/plugin.ts
+++ b/packages/ngtools/webpack/src/ivy/plugin.ts
@@ -569,7 +569,6 @@ export class AngularWebpackPlugin {
         rootNames,
         compilerOptions,
         host,
-        this.ngtscNextProgram,
       );
       const angularCompiler = angularProgram.compiler;
       const pendingAnalysis = angularCompiler.analyzeAsync().then(() => {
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -304,6 +302,26 @@ export class AngularWebpackPlugin {
     });
   }
 
+  private markResourceUsed(
+    normalizedResourcePath: string,
+    currentUnused: Set<string>,
+    visitedResources: Set<string>,
+  ): void {
+    if (visitedResources.has(normalizedResourcePath)) {
```
alan-agius4, line 310: Why do we need to keep track of `visitedResources` separately? We can use `currentUnused` as our single source of truth.

```ts
if (!currentUnused.has(normalizedResourcePath)) {
    return;
}

currentUnused.delete(normalizedResourcePath);

const dependencies = this.fileDependencies.get(normalizedResourcePath);
if (!dependencies) {
    return;
}

for (const dependency of dependencies) {
    this.markResourceUsed(normalizePath(dependency), currentUnused);
}
```

alan-agius4 (approved):

PR #20267, opened 2021-03-14
perf(@angular-devkit/build-angular): remove Webpack Stats.toJson usage in analytics plugin

clydin:
Webpack's `Stats.toJson` function is an expensive operation and is recommended to be avoided where possible. In the case of the analytics plugin, the chunks and assets can be accessed directly from the Webpack compilation.


alan-agius4 (approved):
Nice

PR #20270, opened 2021-03-15
fix(@angular-devkit/schematics-cli): accept windows like paths for sc…

mzocateli:
…hematics

correctly identify schematics paths like 'C:/dir/collection.json:schematic' that used to return 'C' as collection

### DIFF
--- a/packages/angular_devkit/schematics_cli/bin/schematics.ts
+++ b/packages/angular_devkit/schematics_cli/bin/schematics.ts
@@ -38,7 +38,10 @@ function parseSchematicName(str: string | null): { collection: string, schematic
 
   let schematic = str;
   if (schematic && schematic.indexOf(':') != -1) {
-    [collection, schematic] = schematic.split(':', 2);
+    [collection, schematic] = [
+      schematic.slice(0, schematic.lastIndexOf(':')),
+      schematic.substring(schematic.lastIndexOf(':') + 1),
+    ];
   }
 
   return { collection, schematic };


mzocateli:
@alan-agius4  Am I missing something here? Is this viable or should I close it?

OBardo (approved):

packages/angular_devkit/schematics_cli/bin/schematics.ts
```diff
@@ -38,7 +38,10 @@ function parseSchematicName(str: string | null): { collection: string, schematic
 
   let schematic = str;
   if (schematic && schematic.indexOf(':') != -1) {
-    [collection, schematic] = schematic.split(':', 2);
+    [collection, schematic] = [
+      schematic.slice(0, schematic.lastIndexOf(':')),
+      schematic.substring(schematic.lastIndexOf(':') + 1),
+    ];
```
alan-agius4, line 44: Thanks for this, I think this can be simplified by changing the split to use a RegExp

```suggestion
    [collection, schematic] = schematic.split(/[^A-Z]:/, 2);
```

mzocateli: Pleased to help!
I've tried this before, it does split the string at the last colon but cuts the letter before it

So "C:/dir/collection.json:schematic" becomes ["C:/dir/collection.**jso**", "schematic"]

Maybe some other RegEx could fit better, but that's really not my expertise

OBardo: Actually, you could use a negative lookahead to search for the last match of ":"
Like this: **/(:)(?!.*\1)/**

But schematics.split(/(:)(?!.*\1)/), 2) will return ["C:/dir/collection.json", ":"]
Changing the limit to 3 we would have ["C:/dir/collection.json", ":", "schematics"] and then probably could use .filter or .splice to get rid of the ":" item

Don't know if it looks better with this RegEx

mzocateli: Did some research, I think I've came to the right RegEx: **`/:(?=[^:]+$)/`**

```suggestion
[collection, schematic] = schematic.split(/:(?=[^:]+$)/, 2)
```

mzocateli: Alright, I forgot about the `/dir/.:` case, it does not work with the regular expressions above.
Should we still try to use RegEx or settle with the `.slice()` plus `.substring()`?

alan-agius4: Let's go with the `slice` approach.

mzocateli: Ok, all set!

alan-agius4 (approved):
LGTM, thanks

PR #20274, opened 2021-03-15
fix(@schematics/angular): fix youtube icon margin

OmarioHasan:
change youtube icon style from `margin-left: 16px;` to `margin: 0 16px;`

to be like this 
image

instead of this 
image


alan-agius4 (changes requested):
Thanks for this.

Can you please fix the commit message to 

```
fix(@schematics/angular): fix youtube icon margin
```

Thanks.

alan-agius4 (approved):
Thanks

PR #20278, opened 2021-03-15
perf(@angular-devkit/build-angular): disable `showCircularDependencies` by default

alan-agius4:
Change the default of this flag to disabled by default as this is a caused of rebuild performance degradation for large projects.

BREAKING CHANGE

The server and browser builder `showCircularDependencies` option default value has been changed from `true` to `false`.


packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -290,7 +290,8 @@
     "showCircularDependencies": {
       "type": "boolean",
       "description": "Show circular dependency warnings on builds.",
-      "default": true
+      "default": false,
+      "x-deprecated": "Use a linting rule instead instead."
```
JoostK, line 294:
```suggestion
      "x-deprecated": "Use a linting rule instead."
```

packages/angular_devkit/build_angular/src/server/schema.json
```diff
@@ -187,7 +187,8 @@
     "showCircularDependencies": {
       "type": "boolean",
       "description": "Show circular dependency warnings on builds.",
-      "default": true
+      "default": false,
+      "x-deprecated": "Use a linting rule instead instead."
```
JoostK, line 191:
```suggestion
      "x-deprecated": "Use a linting rule instead."
```

alan-agius4: Was still sleeping :)

packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -290,7 +290,8 @@
     "showCircularDependencies": {
       "type": "boolean",
       "description": "Show circular dependency warnings on builds.",
-      "default": true
+      "default": false,
+      "x-deprecated": "Use a linting rule instead."
```
clydin, line 294:
```suggestion
      "x-deprecated": "The recommended method to detect circular dependencies in project code is to use a either a lint rule or other external tooling."
```

packages/angular_devkit/build_angular/src/browser/tests/options/show-circular-dependencies_spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  describe('Option: "showCircularDependencies"', () => {
+    beforeEach(async () => {
+      // Add circular dependency
+      await harness.appendToFile(
+        'src/app/app.component.ts',
+        `import { AppModule } from './app.module'; console.log(AppModule);`,
+      );
+    });
+
+    it('should show cylic dependency warning when option is set to true', async () => {
```
clydin, line 22:
```suggestion
    it('should show cyclic dependency warning when option is set to true', async () => {
```

clydin (approved):

PR #20282, opened 2021-03-15
fix(@schematics/angular): remove Native value from viewEncapsulation option

mouadennaciri:
This commit remove "Native" value from `viewEncapsulation` option in `ng new`, `ng generate application` and `ng generate component`.

Has been removed here: [#38882](https://github.com/angular/angular/pull/38882)


alan-agius4 (approved):
LGTM, thanks

PR #20294, opened 2021-03-16
feat(@angular-devkit/build-angular): expose legacy-migrate message format

crisbeto:
We recently added a new message format called `legacy-extract` that is used to migrate between legacy and canonical message IDs (see https://github.com/angular/angular/pull/41026).

These changes expose the new format in `extract-i18n`.

### DIFF
--- a/packages/angular_devkit/build_angular/src/extract-i18n/index.ts
+++ b/packages/angular_devkit/build_angular/src/extract-i18n/index.ts
@@ -13,6 +13,7 @@ import {
 import { BuildResult, runWebpack } from '@angular-devkit/build-webpack';
 import { JsonObject } from '@angular-devkit/core';
 import type { ɵParsedMessage as LocalizeMessage } from '@angular/localize';
+import type { Diagnostics } from '@angular/localize/src/tools/src/diagnostics';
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/extract-i18n/index.ts
```diff
@@ -73,8 +75,12 @@ async function getSerializer(format: Format, sourceLocale: string, basePath: str
       const { SimpleJsonTranslationSerializer } =
         await import('@angular/localize/src/tools/src/extract/translation_files/json_translation_serializer');
 
-      // tslint:disable-next-line: no-any
       return new SimpleJsonTranslationSerializer(sourceLocale);
+    case Format.LegacyMigrate:
```
crisbeto, line 79: I'm not totally sure whether this value is correct. The `./schema` import didn't work for me, but I assume it's generated from the `schema.json`?

clydin: should be able to run `yarn build` to refresh the generated `.d.ts` files for the schemas.

clydin (approved):

PR #20303, opened 2021-03-17
feat(@angular/cli): support TypeScript 4.2

alan-agius4:
BREAKING CHANGE

Drop support for TypeScript versions prior to 4.2.3

### DIFF
--- a/tests/legacy-cli/e2e/setup/500-create-project.ts
+++ b/tests/legacy-cli/e2e/setup/500-create-project.ts
@@ -34,6 +34,15 @@ export default async function() {
     await expectFileToExist(join(process.cwd(), 'test-project'));
     process.chdir('./test-project');
 
+    // Disable the TS version check to make TS updates easier.
+    // Only VE does it, but on Ivy the i18n extraction uses VE.
+    await updateJsonFile('tsconfig.json', config => {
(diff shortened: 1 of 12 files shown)


clydin (approved):

tests/legacy-cli/e2e/setup/500-create-project.ts
```diff
@@ -34,6 +34,15 @@ export default async function() {
     await expectFileToExist(join(process.cwd(), 'test-project'));
     process.chdir('./test-project');
 
+    // Disable the TS version check to make TS updates easier.
+    // Only VE does it, but on Ivy the i18n extraction uses VE.
```
kyliau, line 38: Didn't we switch the i18n extraction for Ivy projects to use the Ivy compiler already?

alan-agius4: There is a test that use the VE extraction pipeline for Ivy.

https://github.com/angular/angular-cli/blob/47373010e235657c04471c0afc5d8cab5512d952/tests/legacy-cli/e2e/tests/i18n/extract-ivy.ts

kyliau: ah I see, thanks for the clarification.

PR #20306, opened 2021-03-17
doc: update `yarn` command to freeze lockfile.

dgp1130:
In today's release, running `yarn` modified the `yarn.lock` file, which is not desireable for releases which should be as close to CI as possible. This updates the docs to freeze the lockfile (similar to `npm ci`) to avoid changing dependency verisons mid-release.

/cc Caretakers for release (@kyliau and @josephperrott).


kyliau (approved):

alan-agius4:
@googlebot ping!

PR #20311, opened 2021-03-18
build: update changelog to include perf commits

alan-agius4:


scripts/templates/changelog.ejs
```diff
@@ -55,18 +55,39 @@
...
+      if (a.type < b.type) {
+        return -1;
+      }
+
+      if (a.type > b.type) {
+        return 1;
+      }
+
+      return 0;
+    });
```
clydin, line 90:
```suggestion
    scopeCommits.sort((a, b) => a.type - b.type);
```

clydin (approved):

scripts/templates/changelog.ejs
alan-agius4: Actually this is not correct since `a.type - b.type` will always be `NaN`

clydin: ah.  `type` is not numeric.  Either `scopeCommits.sort((a, b) => a.localeCompare(b));` or `scopeCommits.sort((a,b) => a.type > b.type ? 1 : -1);`

PR #20314, opened 2021-03-18
fix(@schematics/angular): use title for svg on home page

cexbrayat:
The SVGs are using `alt` which is not valid HTML. The W3C validator throws:

    Attribute alt not allowed on element svg at this point.

This commit updates the home page to use a proper `title` for accessibility.


alan-agius4 (approved):

alan-agius4:
Thanks

PR #20326, opened 2021-03-18
refactor(@angular-devkit/build-angular): update to cssnano 5

ludofischer:
Check if the latest cssnano release candidate works fine with angular-cli


alan-agius4:
Thanks for this, let's consider this when cssnano 5 is stable.

PR #20332, opened 2021-03-19
fix(@schematics/angular): remove leftover workspace tslint config

alan-agius4:


cexbrayat (commented):
Looks good 👍

I think it is also mentioned [here](https://github.com/angular/angular-cli/blob/a09000cbe655a44c5507642dd9b56da51bbc7490/tests/legacy-cli/e2e/tests/commands/new/new-minimal.ts) Probably can be removed?

alan-agius4:
@cexbrayat, those tests are actually disabled.

clydin (approved):

PR #20348, opened 2021-03-22
perf(@ngtools/webpack): only check affected files for Angular semantic diagnostics

clydin:
This change improves the performance of incremental type checking of Angular templates by reducing the number of calls to retrieve the diagnostics.
Only the set of affected files will be queried on a rebuild. The affected set includes files TypeScript deems affected, files that
are required to be emitted by the Angular compiler, and the original file for any TTC shim file that TypeScript deems affected.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/tests/behavior/rebuild-errors_spec.ts
+++ b/packages/angular_devkit/build_angular/src/browser/tests/behavior/rebuild-errors_spec.ts
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+// tslint:disable: no-big-function
 import { logging } from '@angular-devkit/core';
 import { concatMap, count, take, timeout } from 'rxjs/operators';
 import { buildWebpackBrowser } from '../../index';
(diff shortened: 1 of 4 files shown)


packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -422,13 +421,34 @@ export class AngularWebpackPlugin {
...
+      const result = builder.getSemanticDiagnosticsOfNextAffectedFile(undefined, (sourceFile) => {
+        // If the affected file is a TTC shim, add the shim's original source file.
+        // This ensures that changes that affect TTC are typechecked even when the changes
+        // are otherwise unrelated from a TS perspective and do not result in Ivy codegen changes.
+        // For example, changing @Input property types of a directive used in another component's
+        // template.
+        if (ignoreForDiagnostics.has(sourceFile)) {
+          // This file name conversion relies on internal compiler logic and should be converted
+          // to an official method when available.
+          const originalFilename = sourceFile.fileName.replace(/\.ngtypecheck(\.tsx?)$/, '$1');
```
JoostK, line 435: The shim will always be `.ts`, even if the original was `.tsx`. So the capture group won't help here (and the reverse mapping is therefore ambiguous, although this can be resolved based on `originalSourceFile` ends up being `undefined`)

JoostK: And possibly we should only do this for files that actually end in `.ngtypecheck.ts`? As otherwise we'd add the shim source file to the list of `affectedFiles`.

packages/angular_devkit/build_angular/src/browser/tests/behavior/rebuild-errors_spec.ts
```diff
@@ -5,13 +5,267 @@
...
+        .pipe(
+          timeout(60000),
+          concatMap(async ({ result, logs }, index) => {
+            switch (index) {
+              case 0:
+                expect(result?.success).toBeTrue();
+
+                // Update directive to use a different input type for 'foo' (number -> string)
+                // Should cause a template error
+                await harness.writeFile('src/app/app.component.ts', `
```
JoostK, line 201: Changing the component itself directly affects its TTC file, so this is a trivial case.

A more challenging case would be to have an additional directive that also matches the `dir` element, which initially has a `foo: string;` field which becomes an `@Input` in an incremental update. That does not affect its TS shape (I believe), but it implicitly affects `AppComponent` which now writes the `foo` binding into both directives; one of which is incorrect. This scenario should be covered by this logic:

```ts
          // If required to emit, diagnostics may have also changed
          if (!ignoreForDiagnostics.has(sourceFile)) {
            affectedFiles.add(sourceFile);
          }
```

and I'm not sure if the current tests would fail without this logic, whereas I'd expect the more complex scenario to fail without that logic.

packages/angular_devkit/build_angular/src/browser/tests/behavior/rebuild-errors_spec.ts
```diff
@@ -5,13 +5,267 @@
...
+        .execute({ outputLogsOnFailure: false })
+        .pipe(
+          timeout(60000),
+          concatMap(async ({ result, logs }, index) => {
+            switch (index) {
+              case 0:
+                expect(result?.success).toBeTrue();
+
+                // Update directive to use a different input type for 'foo' (number -> string)
+                // Should cause a template error
```
JoostK, line 200: This comment was not adapted.

packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -468,35 +491,55 @@ export class AngularWebpackPlugin {
...
+        );
+        diagnosticsReporter(angularDiagnostics);
+        this.sourceFileCache?.updateAngularDiagnostics(affectedFile, angularDiagnostics);
+      }
+
+      // NOTE: Workaround to fix stale reuse program. Can be removed once fixed upstream.
+      // tslint:disable-next-line: no-any
+      (angularProgram as any).reuseTsProgram =
+        // tslint:disable-next-line: no-any
+        angularCompiler?.getNextProgram() || (angularCompiler as any)?.getCurrentProgram();
```
JoostK, line 541: Cool; since this is targeting major I've also changed angular/angular#41289 to just major, so this can be removed after that lands without v11 being affected.

JoostK (approved):

Issue #20349, opened 2021-03-22
When budgets are not met and statsJson: true is set, the budget error is swallowed


mattlewis92:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the previous version in which this bug was not present was: `11.2.4`

### Description

When bundle budgets are not met + statsJson: true is set, ng build fails with a cryptic error message instead of showing the budget error.

## 🔬 Minimal Reproduction
* Add budgets that don't pass
* Set `statsJson: true`
* Run `ng build --prod`

## 🔥 Exception or Error
<pre><code>✔ Browser application bundle generation complete.
ENOENT: no such file or directory, open '/path/to/repo/dist/apps/app-name/stats.json'
</code></pre>

## 🌍 Your Environment
<pre><code>
Angular CLI: 11.2.5
Node: 14.13.1
OS: darwin x64

Angular: 11.2.6
... animations, common, compiler, compiler-cli, core, elements
... forms, language-service, localize, platform-browser
... platform-browser-dynamic, router, service-worker
Ivy Workspace: <error>

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1100.6
@angular-devkit/build-angular   0.1102.5
@angular-devkit/core            11.0.6
@angular-devkit/schematics      11.2.4
@angular/cdk                    9.0.1
@angular/cli                    11.2.5
@angular/google-maps            11.0.3
@schematics/angular             11.2.4
@schematics/update              0.1102.5
ng-packagr                      11.2.3
rxjs                            6.6.3
typescript                      4.1.3
</code></pre>

**Anything else relevant?**
It was [this commit]( https://github.com/angular/angular-cli/commit/d5645675fd555e7f1afd523d4f2d42095034fc46) that caused the regression, I manually reverted to the old behaviour in my node_modules and verified that it fixed the problem.

mattlewis92:
Verified that this is now fixed in `11.2.6`, thanks for the quick turnaround @clydin! 🥳

PR #20350, opened 2021-03-22
fix(@angular-devkit/build-angular): ensure output directory is present before writing stats JSON

clydin:


alan-agius4 (approved):
Can you add a reference to https://github.com/angular/angular-cli/issues/20349 in the commit?

PR #20352, opened 2021-03-23
feat(@schematics/angular): create add schema method in ast-utils

jtneal:
The AST Utils include methods to add new values to each section of an NgModule's metadata including `declarations`, `imports`, `providers`, `exports`, `bootstrap` and `entryComponents`. However, there is no method for schemas. Schematics authors can easily use the `addSymbolToNgModuleMetadata` method to achieve this capability, but I propose to add this `addSchemaToModule` method as a minor quality of life improvement.

### DIFF
--- a/packages/schematics/angular/utility/ast-utils.ts
+++ b/packages/schematics/angular/utility/ast-utils.ts
@@ -504,6 +504,15 @@ export function addExportToModule(source: ts.SourceFile,
 }
 
 /**
+ * Custom function to insert a schema into NgModule. It also imports it.
+ */
+export function addSchemaToModule(source: ts.SourceFile,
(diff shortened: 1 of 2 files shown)


alan-agius4:
Hi @jtneal,

Thanks a lot for your contribution. 

However, all of the methods that require a deep import to use are part of the private API and are not meant to be used by 3rd parties. Any part of the private API can change without warning in any version.

Seeing that the new added method is not used internally, I don't see any reason to add this method to our private API.

Thanks again,
Alan

jtneal:
Ah, thanks for the info @alan-agius4 - I did notice `addProviderToModule` and `addBootstrapToModule` weren't used internally as well, so I figured this might be a valid use-case/contribution.

PR #20353, opened 2021-03-23
build: determine release version from package.json instead of git tag

kyliau:
Currently, the version of a release is determined by the git tag.
This PR changes the release script to determine the release version from the
`version` property in the root `package.json`.

With this change, the caretaker will have to update the version in both
`package.json` and `latest-versions.ts` **after** the release.

Release doc has also been updated.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -22,9 +22,6 @@ Alan    | Doug
 Charles | Keen
 Filipe  | Joey
 
-## Triaging Issues
-TBD
-
 ## Merging PRs
 
 The list of PRs which are currently ready to merge (approved with passing status checks) can
@@ -37,7 +34,6 @@ When ready to merge a PR, run the following command:
 ```
 yarn ng-dev pr merge <pr>
 ```
-
 
 ### Maintaining LTS branches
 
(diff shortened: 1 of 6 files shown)


alan-agius4 (approved):
LGTM

josephperrott (approved):
LGTM

docs/process/release.md
```diff
@@ -159,6 +149,12 @@ using the `--githubToken` flag. You just then have to confirm the draft.
 
 > **Tags containing `next` or `rc` should be marked as pre-release.**
 
+## Post-release Version Update
```
dgp1130, line 152: Do we want to document/expect any particular commit message? Release commits are very particular, though it is probably less important here.

dgp1130 (approved):
MTGL

Out of curiosity, what is the need to update the FW version each patch? Semver already allows latest patch, so this seems mostly unnecessary. If we didn't need to do that, we wouldn't need a release commit anymore (at least for patch releases).

docs/process/release.md
kyliau: Added commit message.
```sh
git commit package.json packages/schematics/angular/utility/latest-versions.ts -m "build: bump version to vXX"
```

kyliau:
> Out of curiosity, what is the need to update the FW version each patch? Semver already allows latest patch, so this seems mostly unnecessary. If we didn't need to do that, we wouldn't need a release commit anymore (at least for patch releases).

That's a great suggestion, we certainly don't have to update `latest-versions.ts` for a patch release.
I think we should discuss whether we want an empty release commit or none at all.
I think we should always create a release commit, because it might be confusing if a release commit is needed for new major / minor release but not for patch.

kyliau:
Added item to meeting agenda to discuss whether we should update `latest-versions.ts`. Meanwhile, will merge this PR in preparation for release tomorrow.

clydin (approved):

clydin:
@kyliau Can you open a patch target PR? There appears to be a conflict.

PR #20376, opened 2021-03-25
refactor(@angular/cli): inline @schematics/update:update schematic

clydin:
This change removes the need for the `@schematics/update` package within the Angular tooling and removes the dependency from the `@angular/cli` package.
Only the `update` schematic from the `@schematics/update` package was used and this schematic's logic will eventually be folded into the update command logic directly.


alan-agius4:
This is a great initiative!

kyliau (approved):

PR #20384, opened 2021-03-26
fix(@schematics/angular): explicitly specify ServiceWorker registration strategy

gkalpak:
This PR updates the `service-worker` schematics to explicitly specify the ServiceWorker registration strategy in the [ServiceWorkerModule.register()] call.

We still use the default strategy, so there should be no change in the behavior of the generated apps. However, it will help people find out what the default behavior is when debugging potential issues with delayed ServiceWorker registration.
(See the discussion in angular/angular#41223 for more details.)

##
In order to be able to do this (and format the code in a readable way), this PR also makes some updates to the `addSymbolToNgModuleMetadata()` function. See individual commits for details.

[1]: https://angular.io/api/service-worker/ServiceWorkerModule#register

### DIFF
--- a/packages/schematics/angular/service-worker/index.ts
+++ b/packages/schematics/angular/service-worker/index.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { join, normalize } from '@angular-devkit/core';
+import { join, normalize, tags } from '@angular-devkit/core';
 import {
   Rule,
   SchematicContext,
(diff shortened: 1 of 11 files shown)


packages/schematics/angular/service-worker/index.ts
```diff
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import { join, normalize } from '@angular-devkit/core';
+import { tags } from '@angular-devkit/core';
```
alan-agius4, line 9: Duplicate import for devit core.

alan-agius4 (changes requested):
LGTM

alan-agius4 (approved):
LGTM

PR #20397, opened 2021-03-29
fix(@angular-devkit/build-angular): remove deprecated View Engine support for i18n extraction

clydin:
BREAKING CHANGE: Removal of View Engine support from i18n extraction
With the removal of the deprecated View Engine compiler in Angular version 12 for applications, the `ng extract-i18n` command will now always use the Ivy compiler. The `--ivy` option has also been removed as Ivy-based extraction is always enabled.
The default behavior for applications is to use the Ivy compiler for building/extraction and no changes are required for these applications. For applications that have opted-out of Ivy, a warning will be shown and Ivy-based extraction will be attempted. If the extraction fails, the application may need to be updated to become Ivy compatible.


packages/angular_devkit/build_angular/src/extract-i18n/schema.json
```diff
@@ -49,7 +49,8 @@
     },
     "ivy": {
       "type": "boolean",
-      "description": "Use Ivy compiler to extract translations. The default for Ivy applications."
+      "description": "Use Ivy compiler to extract translations. The default for Ivy applications.",
+      "x-deprecated": "Option is always enabled. Angular version 12 and higher no longer support View Engine compilation for applications."
```
alan-agius4, line 53: Should we just remove anice it’s a noop and the package is technically experimental

clydin: Ok

alan-agius4 (approved):

PR #20404, opened 2021-03-30
test(@angular-devkit/build-angular): clean up tests from ViewEngine code

alan-agius4:


packages/angular_devkit/build_angular/src/browser/specs/rebuild_spec.ts
```diff
@@ -224,52 +223,6 @@ describe('Browser Builder rebuilds', () => {
       .toPromise();
   });
 
-  it('rebuilds on transitive type-only file changes', async () => {
-    if (veEnabled) {
```
clydin, line 228: This looks like an Ivy only test.  I think the `if` block can just be removed.

clydin (approved):

PR #20412, opened 2021-03-31
refactor(@angular-devkit/build-angular): remove deprecated i18n options from server and browser builder

alan-agius4:
BREAKING CHANGE:

Removal of deprecated browser and server command options.
- `i18nFile`,  use `locales` object in the project metadata instead.
- `i18nFormat`, No longer needed as the format will be determined automatically.
- `i18nLocale`, use `localize` option instead.


alan-agius4:
Blocked until https://github.com/angular/angular-cli/pull/20407 is landed.

packages/angular_devkit/build_angular/src/browser/index.ts
```diff
@@ -668,7 +668,7 @@ export function buildWebpackBrowser(
                       const { content, warnings, errors } = await indexHtmlGenerator.process({
                         baseHref: getLocaleBaseHref(i18n, locale) || options.baseHref,
                         // i18nLocale is used when Ivy is disabled
-                        lang: locale || options.i18nLocale,
+                        lang: locale || undefined,
```
clydin, line 671: Is the undefined needed here?

clydin (approved):

packages/angular_devkit/build_angular/src/browser/index.ts
alan-agius4: It is needed because when it’s locale empty string, a lang attribute is added with a an empty string.

Maybe the index generation shouldn’t generate the lang attribute if lang is an empty string.

clydin: ok.  This works for now.  We can look at that in the future.

Issue #19492, opened 2020-11-27
Unable to execute schematics with object configured via angular.json


bgotink:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, this worked in 11.0.1 / until #18945

### Description

It is impossible to start a schematic that has an option of type object where that object comes filled in via the angular.json file.

## 🔬 Minimal Reproduction

Clone https://github.com/bgotink/angular-cli-repro-19492

Run `ng g @repro/schematics:test`, get an error.

Comment/delete the default configuration from `schematics` in angular.json.

Run `ng g @repro/schematics:test` again, it'll succeed this time.

## 🔥 Exception or Error
<pre><code>$ ng g @repro/schematics:test
Cannot read property 'constructor' of null
</code></pre>

The error is thrown at

https://github.com/angular/angular-cli/blob/707e03f60da9f44bfcc98fbd7f4e9c55a65f1ff1/packages/angular_devkit/core/src/utils/object.ts#L43

## 🌍 Your Environment
<pre><code>

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 11.0.2
Node: 12.19.0
OS: darwin x64

Angular: 11.0.2
... animations, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router
Ivy Workspace: Yes

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1100.2
@angular-devkit/build-angular   0.1100.2
@angular-devkit/core            11.0.2
@angular-devkit/schematics      11.0.2
@schematics/angular             11.0.2
@schematics/update              0.1100.2
rxjs                            6.6.3
typescript                      4.0.5
</code></pre>

**Anything else relevant?**

PR #20430, opened 2021-04-01
fix(@angular-devkit/core): correctly handle null prototype in deepCopy

daiscog:
Cherry-pick of the fix already in master and 11.2.x for #19492

### DIFF
--- a/packages/angular_devkit/core/src/utils/object.ts
+++ b/packages/angular_devkit/core/src/utils/object.ts
@@ -40,7 +40,7 @@ export function deepCopy<T extends any>(value: T): T {
       return JSON.parse(valueCasted['toJSON']());
     }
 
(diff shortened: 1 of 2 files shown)


alan-agius4:
Hi @daiscog,

Thanks for this, but version 10.2 is currently in LTS and only security fixes can be merged. See https://angular.io/guide/releases#lts-fixes for more information.

Kindly update to the most recent version to avail of this fix.

daiscog:
@alan-agius4 this is really frustrating, as we cannot yet move to Angular 11 due to external dependencies.

alan-agius4:
@daiscog, I am sorry, but in this case I suggest to file an issue with these external dependencies. 

Our LTS policy is pretty clear about this.

PR #20431, opened 2021-04-01
refactor(@angular/pwa): remove `@angular-devkit/core` dependency

clydin:
The `@angular-devkit/core` dependency was only used for two path manipulation functions. Since this schematic already uses Node.js builtins, the `path` builtin can also be used to provide equivalent function without the need for an additional dependency.

### DIFF
--- a/packages/angular/pwa/pwa/index.ts
+++ b/packages/angular/pwa/pwa/index.ts
@@ -5,7 +5,6 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { join, normalize } from '@angular-devkit/core';
 import {
   Rule,
   SchematicsException,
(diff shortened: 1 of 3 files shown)


packages/angular/pwa/pwa/index.ts
```diff
@@ -117,7 +117,7 @@ export default function(options: PwaOptions): Rule {
     }
 
     // Add manifest to asset configuration
-    const assetEntry = join(normalize(project.root), 'src', 'manifest.webmanifest');
+    const assetEntry = join(project.root, project.sourceRoot ?? 'src', 'manifest.webmanifest');
```
alan-agius4, line 120: Use posix.join to retain the same behaviour on non posix file systems.

clydin: Updated.

alan-agius4:
Can you update the build file dependencies too please?

https://github.com/angular/angular-cli/blob/1fcc120b16c34d00ba9e1d4a88178fd045a33834/packages/angular/pwa/BUILD.bazel#L41

alan-agius4 (approved):

PR #20443, opened 2021-04-03
fix(@ngtools/webpack): remove Webpack plugin for deprecated ViewEngine compiler

clydin:
BREAKING CHANGE: Removal of View Engine support from application builds
With the removal of the deprecated View Engine compiler in Angular version 12 for applications, the View Engine Webpack plugin has been removed.
The Ivy-based Webpack plugin is the default used within the Angular CLI.
If using a custom standalone Webpack configuration, the removed `AngularCompilerPlugin` should be replaced with the Ivy-based `AngularWebpackPlugin`.


packages/ngtools/webpack/src/index.ts
```diff
@@ -5,11 +5,5 @@
...
  */
-
-export * from './angular_compiler_plugin';
-export * from './interfaces';
-export { ngcLoader as default } from './loader';
-
-export const NgToolsLoader = __filename;
-
+export { AngularPluginOptions, AngularWebpackPlugin, default } from './ivy';
 export * as ivy from './ivy';
```
alan-agius4, line 9: Shouldn't this export be dropped?

alan-agius4 (approved):
One NIT.

packages/ngtools/webpack/src/index.ts
clydin: Marked the namespace export as deprecated.  Also, remove its usage from the builders (in a separate commit).

PR #20466, opened 2021-04-08
Upgrade tooling to use Webpack 5

clydin:
With this change Webpack 5 is now used by the Angular tooling to build applications. Webpack 4 usage and support has been removed.
No project level configuration changes are required to take advantage of the upgraded Webpack version when using the official Angular builders.
Custom builders based on this package that use the experimental programmatic APIs may need to be updated to become compatible with Webpack 5.

BREAKING CHANGE: Webpack 5 lazy loaded file name changes
Webpack 5 generates similar but differently named files for lazy loaded JavaScript files in development configurations (when the `namedChunks` option is enabled).
For the majority of users this change should have no effect on the application and/or build process. Production builds should also not be affected as the `namedChunks` option is disabled by default in production configurations.
However, if a project's post-build process makes assumptions as to the file names then adjustments may need to be made to account for the new naming paradigm.
Such post-build processes could include custom file transformations after the build, integration into service-side frameworks, or deployment procedures.
Example development file name change: `lazy-lazy-module.js` --> `src_app_lazy_lazy_module_ts.js`

BREAKING CHANGE: Webpack 5 web worker support
Webpack 5 now includes web worker support. However, the structure of the URL within the `Worker` constructor must be in a specific format that differs from the current requirement.
Web worker usage should be updated as shown below (where `./app.worker` should be replaced with the actual worker name):
Before: `new Worker('./app.worker', ...)`
After:  `new Worker(new URL('./app.worker', import.meta.url), ...)`

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
@@ -7,8 +7,8 @@
  */
 import * as cssNano from 'cssnano';
 import { ProcessOptions, Result } from 'postcss';
-import { Compiler, compilation } from 'webpack';
-import { RawSource, Source, SourceMapSource } from 'webpack-sources';
+import { Compiler, compilation, sources } from 'webpack';
+import { Source } from 'webpack-sources';
 import { addWarning } from '../../utils/webpack-diagnostics';
 import { isWebpackFiveOrHigher } from '../../utils/webpack-version';
 
@@ -36,7 +36,7 @@ function hook(
       });
     } else {
       compilation.hooks.optimizeChunkAssets
-        .tapPromise(PLUGIN_NAME, (chunks: Iterable<compilation.Chunk>) => {
+        .tapPromise(PLUGIN_NAME, (chunks) => {
           const files: string[] = [];
           for (const chunk of chunks) {
             if (!chunk.files) {
@@ -124,7 +124,7 @@ export class OptimizeCssWebpackPlugin {
 
           let newSource;
           if (output.map) {
-            newSource = new SourceMapSource(
+            newSource = new sources.SourceMapSource(
               output.css,
               file,
               // tslint:disable-next-line: no-any
@@ -133,7 +133,7 @@ export class OptimizeCssWebpackPlugin {
               map,
             );
           } else {
-            newSource = new RawSource(output.css);
+            newSource = new sources.RawSource(output.css);
           }
 
           compilation.assets[file] = newSource;
--- a/packages/angular_devkit/build_optimizer/package.json
+++ b/packages/angular_devkit/build_optimizer/package.json
@@ -16,7 +16,7 @@
     "webpack-sources": "2.2.0"
   },
   "peerDependencies": {
-    "webpack": "^4.0.0 || ^5.20.0"
+    "webpack": "^4.0.0 || ^5.30.0"
   },
   "peerDependenciesMeta": {
     "webpack": {
@@ -25,6 +25,6 @@
   },
   "devDependencies": {
     "@types/webpack": "^4.41.22",
-    "webpack": "5.21.2"
+    "webpack": "5.30.0"
   }
 }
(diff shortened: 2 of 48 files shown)


packages/angular_devkit/build_optimizer/package.json
```diff
@@ -16,7 +16,7 @@
     "webpack-sources": "2.2.0"
   },
   "peerDependencies": {
-    "webpack": "^4.0.0 || ^5.20.0"
+    "webpack": "^4.0.0 || ^5.30.0"
```
alan-agius4, line 19: Should we still support V4 here?

packages/angular_devkit/build_webpack/package.json
```diff
@@ -12,10 +12,11 @@
   },
   "devDependencies": {
     "@angular-devkit/core": "0.0.0",
-    "node-fetch": "2.6.1"
+    "node-fetch": "2.6.1",
+    "webpack": "5.30.0"
   },
   "peerDependencies": {
-    "webpack": "^4.6.0",
+    "webpack": "^4.6.0 || ^5.30.0",
```
alan-agius4, line 19: Should we support v4 here?

packages/ngtools/webpack/package.json
```diff
@@ -27,13 +27,13 @@
   "peerDependencies": {
     "@angular/compiler-cli": "^12.0.0-next",
     "typescript": "~4.2.3",
-    "webpack": "^4.0.0 || ^5.20.0"
+    "webpack": "^4.0.0 || ^5.30.0"
```
alan-agius4, line 30: Same as previous

packages/ngtools/webpack/src/ivy/symbol.ts
```diff
@@ -15,3 +15,43 @@ export interface EmitFileResult {
 }
 
 export type FileEmitter = (file: string) => Promise<EmitFileResult | undefined>;
+
+export class FileEmitterRegistration {
+  #fileEmitter?: FileEmitter;
```
alan-agius4, line 20: 👍

alan-agius4 (approved):
Good job!

Couple of points that should probably be addressed in a seperate PR.

- clean up left over `withWebpackFourOrFive` and `isWebpackFiveOrHigher` calls especially in build-angular.
- should peer dependency on webpack 4 be dropped though-out? 
- will a migration for the web-worker syntax change be done?
- update all 3P webpack plugins and loaders, since some of them are outdated due to the lack of support for Webpack 4.

packages/angular_devkit/build_optimizer/package.json
clydin: It technically will still work with 4 (for now).  There is some additional refactoring/cleanup that can be done to leverage newer features though and once that is in 4 support can be dropped.

clydin:
Once this is in, there are a variety of refactoring and cleanup opportunities throughout the codebase by leveraging new Webpack 5 capabilities.  I tried to keep this PR close to a minimal set of changes to enable Webpack 5 to reduce the effort in getting it reviewed and merged.

alan-agius4:
> Once this is in, there are a variety of refactoring and cleanup opportunities throughout the codebase by leveraging new Webpack 5 capabilities. I tried to keep this PR close to a minimal set of changes to enable Webpack 5 to reduce the effort in getting it reviewed and merged.

Yeah that was a good plan!

packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
```diff
@@ -133,7 +133,7 @@ export class OptimizeCssWebpackPlugin {
               map,
             );
           } else {
-            newSource = new RawSource(output.css);
+            newSource = new sources.RawSource(output.css);
           }
 
           compilation.assets[file] = newSource;
```
alexander-akait, line 139: Just note, we have https://github.com/webpack-contrib/css-minimizer-webpack-plugin - cache and multi threading included

alan-agius4: Thanks for sharing, will take a look

PR #20482, opened 2021-04-09
feat(@angular-devkit/core): update schema validator

alan-agius4:
**feat(@angular-devkit/core): update schema validator**
    
With this change we update ajv to version 8
    
BREAKING CHANGE: support for JSON Schema draft-04 and draft-06 is removed. If you have schemas using the `id` keyword replace them with `$id`. For an interim period we will auto rename any top level `id` keyword to `$id`.
    
**NB**: This change only effects schematics and builders authors.

**refactor: add correct schema keys**
    
With this change
- We replace `id` with `$id`,  this no longer valid in draft-07.
- Replace all `$schemas` to `http://json-schema.org/draft-07/schema`, this is needed to "pin" the schema to `draft-07`.
    
More information about `draft-07` can be found https://json-schema.org/draft-07/json-schema-release-notes.html


packages/angular_devkit/core/src/json/schema/registry.ts
```diff
@@ -361,57 +338,32 @@ export class CoreSchemaRegistry implements SchemaRegistry {
 
       // Validate using ajv
       const result = validator.call(validationContext, data);
```
clydin, line 340: I think this can also still return a promise.  Prompts should be the scenario where this can happen.

clydin: Probably could restructure this section with an `await` on the call itself to normalize the result and then wrapped in a `try`/`catch` to normalize the different error handling mechanisms (`validator.errors` vs the exception).

alan-agius4: You are right.

clydin (approved):

PR #20487, opened 2021-04-09
Inline component CSS style processing

clydin:
Support is added to allow processing of an Angular component's inline CSS styles (`styles` decorator property).  Previously, only file stylesheets (`styleUrls` decorator property) would be processed by the Angular CLI with such tools as `postcss` or `autoprefixer`.


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -211,129 +217,150 @@ export function getStylesConfig(wco: WebpackConfigOptions): webpack.Configuratio
...
-                    paths: includePaths,
-                  },
-                },
-              },
-            ],
+        },
+      ],
+    },
+    {
+      extensions: ['styl', 'stylus'],
```
alan-agius4, line 309: Cannot find any reference to `stylus` being a using as a file extension for stylus

```suggestion
      extensions: ['styl'],
```

alan-agius4 (approved):

Issue #18844, opened 2020-09-22
Consider removing protractor as the default E2E tool


rooby:
# 🚀 Feature request

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [x] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

The default E2E tool that is provided is protractor, which appears to be effectively unmaintained and doesn't seem to have any kind of road map or any communications from the maintainers that would make me think that's going to change any time soon.

I think that there is a high chance that new users will either just use the provided tool by default and not think twice about it, or they would give extra weight to the provided solution when weighing up options (I know I have done that in the past) and then go on to possibly regret that choice (or resent the default they didn't realise they could change).

I think we should stop steering people in that direction.

### Describe the solution you'd like

I think it would be preferable to either ship with a tool that is actively maintained or to ship without a default tool at all and just provide documentation on how to configure the tool of their choice (just the Angular specific docs, they can get tool specific docs from the tool they choose).

### Describe alternatives you've considered

If I'm wrong about the maintenance status of protractor, then a possible alternative is that it becomes an actively maintained, 
 user friendly solution that's worthy of the default tool status.

Brandinga:
https://angular.io/guide/roadmap#update-our-e2e-testing-strategy

rooby:
> https://angular.io/guide/roadmap#update-our-e2e-testing-strategy

I've seen that, however that doesn't really mean anything by itself. It's too vague.
There is no timeline, no github issues to track the progress or decision making conversation, and there's no date for when that was written so you don't know if it's a recent thing and we still need to give people time to work things out or if it's been in that list for a long time and progress has stalled.

ChristopherKiss:
As a side note, there are also no docs on how to even write e2e tests or use Protractor (as apposed to their fantastic unit test section which is incredibly detailed).

I requested more info almost 2 years ago: https://github.com/angular/angular/issues/28968

Which didn't see any progress from the looks of it. Would be great to get official direction on the state of e2e with Angular.

rooby:
There still seems to have been no progress on development or documentation for the Protractor project.
The protractor release version mismatch between github and npm also hasn't been addressed.
The issues in the Protractor issue queue seem to generally go unanswered.

Please can we stop encouraging users to use this approach for E2E?

PR #20503, opened 2021-04-12
refactor(@schematics/angular): Do not include Protractor test in new project

kyliau:
Do not include Protractor for new projects since we do not have a concrete plan for Protractor yet.

Closes https://github.com/angular/angular-cli/issues/18844

### DIFF
--- a/packages/angular/cli/commands/e2e-impl.ts
+++ b/packages/angular/cli/commands/e2e-impl.ts
@@ -14,8 +14,20 @@ import { Schema as E2eCommandSchema } from './e2e';
 export class E2eCommand extends ArchitectCommand<E2eCommandSchema> {
   public readonly target = 'e2e';
   public readonly multiTarget = true;
+  public readonly missingTargetError = `
+Cannot find "e2e" target for the specified project.
 
-  public async run(options: E2eCommandSchema & Arguments) {
-    return this.runArchitectTarget(options);
+You should add a package that implements end-to-end testing capabilities.
+
+For example:
+  WebdriverIO: ng add @wdio/schematics
+
+More options will be added to the list as they become available.
+`;
+
+  async initialize(options: E2eCommandSchema & Arguments) {
+    if (!options.help) {
+      return super.initialize(options);
+    }
   }
 }
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -29,7 +29,6 @@ import {
 } from '@angular-devkit/schematics';
 import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
 import { Schema as ComponentOptions } from '../component/schema';
-import { Schema as E2eOptions } from '../e2e/schema';
 import { NodeDependencyType, addPackageJsonDependency } from '../utility/dependencies';
 import { latestVersions } from '../utility/latest-versions';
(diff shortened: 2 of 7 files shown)


packages/angular/cli/commands/e2e-impl.ts
```diff
@@ -14,8 +14,14 @@ import { Schema as E2eCommandSchema } from './e2e';
 export class E2eCommand extends ArchitectCommand<E2eCommandSchema> {
   public readonly target = 'e2e';
   public readonly multiTarget = true;
+  public readonly missingTargetError = `
+Cannot find "e2e" target for the specified project.
 
-  public async run(options: E2eCommandSchema & Arguments) {
-    return this.runArchitectTarget(options);
+You should add a package that implements e2e capabilities.`;
```
alan-agius4, line 20: You should add a package that implements end-to-end testing capabilities.

alan-agius4 (approved):
One optional NIT, otherwise LGTM

packages/angular/cli/commands/e2e-impl.ts
```diff
@@ -14,8 +14,19 @@ import { Schema as E2eCommandSchema } from './e2e';
...
   public readonly target = 'e2e';
   public readonly multiTarget = true;
+  public readonly missingTargetError = `
+Cannot find "e2e" target for the specified project.
 
-  public async run(options: E2eCommandSchema & Arguments) {
-    return this.runArchitectTarget(options);
+You should add a package that implements e2e capabilities.
+
+Here are some options that are available:
```
alan-agius4, line 22: Would be nice to keep the same format as the other builders ie: https://github.com/angular/angular-cli/blob/c3cc50e6b95fbbd814d752eec808ea0a0ef411c0/packages/angular/cli/commands/deploy-impl.ts#L18

packages/angular/cli/commands/e2e-impl.ts
```diff
@@ -14,8 +14,19 @@ import { Schema as E2eCommandSchema } from './e2e';
...
+  1. TestCafe (https://testcafe.io/documentation/402793/recipes/integrations/use-angular-cli-builder)
+  2. WebdriverIO (https://webdriver.io/docs/protractor-migration/)
```
alan-agius4, line 24: Should we mentioned this? Since it cannot be used with this command?

clydin: maybe like here: https://github.com/webdriverio/webdriverio-schematics

packages/angular/cli/commands/e2e-impl.ts
kyliau: done

packages/angular/cli/commands/e2e-impl.ts
kyliau: Mentioned `ng add` for WebdriverIO, removed TestCafe for now

packages/angular/cli/commands/e2e-impl.ts
kyliau: Removed TestCafe for now, opened https://github.com/politie/angular-testcafe/issues/63 to ask them to add schematic.
Mentioned `ng add` for WebdriverIO.

christian-bromann (approved):
👍

clydin (approved):

PR #20504, opened 2021-04-12
ci: remove integration test run under Bazel

kyliau:
Test is obsolete, and not run on CI.


alan-agius4 (changes requested):
You can also delete https://github.com/angular/angular-cli/tree/master/tools/npm_integration_test since this is only needed for this test

alan-agius4 (approved):
👍

kyliau:
> You can also delete https://github.com/angular/angular-cli/tree/master/tools/npm_integration_test since this is only needed for this test

Done!

PR #20510, opened 2021-04-13
Remove old migrations

alan-agius4:


packages/angular/cli/commands/update-impl.ts
```diff
@@ -550,6 +550,28 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
...
 
+      if (pkg.name === '@angular/cli') {
+        // Migrations for non LTS versions of Angular CLI are no longer included in @schematics/angular v12.
+        const fromMajorVersion = +node.version.split('.')[0];
+        const installedMajorVersion = +node.package.version.split('.')[0];
+        if (fromMajorVersion < 8 && (installedMajorVersion > 11 || pkg.type !== 'version')) {
+          const updateVersions: Record<number, number> = {
+            1: 6,
+            6: 8,
+            8: 9,
```
clydin, line 561:
```suggestion
            7: 8,
```

packages/angular/cli/commands/update-impl.ts
```diff
@@ -550,6 +550,28 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
         return 1;
       }
 
+      if (pkg.name === '@angular/cli') {
```
clydin, line 553: I think this check should be moved down to the next section that retrieves the manifest for the version that will be installed. `manifest.version` (the exact version that will be installed) and `node.package.version` (the exact version currently installed) can then be used within the logic.
Right above the following would probably be a good spot: https://github.com/angular/angular-cli/blob/f77aef238a4f65f4613d87562673f7d2f4343fe0/packages/angular/cli/commands/update-impl.ts#L648

clydin (changes requested):
This code should also be removed now that the migration is no longer present:
https://github.com/angular/angular-cli/blob/f77aef238a4f65f4613d87562673f7d2f4343fe0/packages/angular/cli/commands/update-impl.ts#L371-L382

clydin (approved):

PR #20514, opened 2021-04-13
feat(@angular-devkit/build-angular): support specifying stylesheet language for inline component styles

clydin:
A new build option named `inlineStyleLanguage` has been introduced that will allow a project to define the stylesheet language used in an application's inline component styles. Inline component styles are styles defined via the `styles` property within the Angular `Component` decorator. Both JIT and AOT mode are supported. However, JIT mode requires that inline styles only be string literals (compile-time partial evaluation is not supported in JIT mode). Currently supported language options are: `CSS` (default), `Sass`, `SCSS`, and `Less`. If the option is not specified, `CSS` will be used and enables existing projects to continue to function as expected.

### DIFF
--- a/packages/angular_devkit/build_angular/src/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/browser/schema.json
@@ -40,6 +40,17 @@
         "$ref": "#/definitions/extraEntryPoint"
       }
     },
+    "inlineStyleLanguage": {
+      "description": "The stylesheet language to use for the application's inline component styles.",
+      "type": "string",
+      "default": "css",
+      "enum": [
+        "css",
+        "less",
+        "sass",
+        "scss"
+      ]
+    },
     "stylePreprocessorOptions": {
       "description": "Options to pass to style preprocessors.",
       "type": "object",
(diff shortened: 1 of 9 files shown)


packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -41,6 +41,17 @@
         "$ref": "#/definitions/extraEntryPoint"
       }
     },
+    "inlineStyleLanguage": {
```
alan-agius4, line 44: Drive by comment.

I think this option needs to be added in karma and server builder as well.

packages/angular_devkit/build_angular/src/browser/schema.json
```diff
@@ -41,6 +41,17 @@
...
+      "description": "The stylesheet language to use for the application's inline component styles.",
+      "type": "string",
+      "default": "css",
+      "enum": [
```
alan-agius4, line 48: stylus is missing, is it on purpose?

alan-agius4 (approved):

PR #20515, opened 2021-04-13
Webpack 5 Cleanup - Remove redundant dependencies

clydin:
Webpack 5 now provides common utility functions and class as direct exports.  This includes getting a loader's options (`getOptions`) and the source classes from `webpack-sources`.


alan-agius4 (approved):
LGTM

PR #20518, opened 2021-04-14
refactor(@angular-devkit/build-angular): use Node.js promise fs API in service-worker augmentation

clydin:
With the minimum version of Node.js now set to v12, the promise fs API can now be leveraged within the tooling.
This change also uses `copyFile` (with copy-on-write where available) to setup the the service worker files as well a streaming APIs to generate service worker hashes. Both of which improves performance and reduces memory usage.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/service-worker.ts
+++ b/packages/angular_devkit/build_angular/src/utils/service-worker.ts
@@ -5,65 +5,56 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {
-  Path,
-  dirname,
-  getSystemPath,
-  join,
-  normalize,
-  relative,
-  tags,
-  virtualFs,
-} from '@angular-devkit/core';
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/utils/service-worker.ts
```diff
@@ -5,67 +5,57 @@
...
-
-      if (await this._host.isDirectory(item).toPromise()) {
-        await this._recursiveList(item, items);
-      } else {
-        items.push('/' + relative(normalize(this.base), item));
+  private async _recursiveList(dir: string, items: string[]): Promise<string[]> {
+    const subdirectories = [];
+    for await (const entry of await fs.opendir(dir)) {
+      if (entry.isFile()) {
+        items.push('/' + path.relative(this.base, path.join(dir, entry.name)));
```
alan-agius4, line 49: Does the SW generator handle non posix path?

clydin: apparently not. updated to use posix path variants.

alan-agius4 (approved):

PR #20532, opened 2021-04-14
feat(@schematics/angular): migrate web workers to support Webpack 5

clydin:
Webpack 5 now includes web worker support. However, the structure of the URL within the `Worker` constructor must be in a specific format.
A migration has been added for Angular v12 that will convert web workers from the old structure to the new structure.
Before: `new Worker('./app.worker', ...)`
After: `new Worker(new URL('./app.worker', import.meta.url), ...)`

### DIFF
--- a/packages/schematics/angular/migrations/update-12/update-web-workers.ts
+++ b/packages/schematics/angular/migrations/update-12/update-web-workers.ts
@@ -1,0 +1,100 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import { DirEntry, Rule, UpdateRecorder } from '@angular-devkit/schematics';
+import * as ts from '../../third_party/github.com/Microsoft/TypeScript/lib/typescript';
+
+function* visit(directory: DirEntry): IterableIterator<ts.SourceFile> {
+  for (const path of directory.subfiles) {
+    if (path.endsWith('.ts') && !path.endsWith('.d.ts')) {
+      const entry = directory.file(path);
+      if (entry) {
+        const content = entry.content;
+        if (content.includes('Worker')) {
+          const source = ts.createSourceFile(
+            entry.path,
+            // Remove UTF-8 BOM if present
+            // TypeScript expects the BOM to be stripped prior to parsing
+            content.toString().replace(/^\uFEFF/, ''),
+            ts.ScriptTarget.Latest,
+            true,
+          );
+
+          yield source;
+        }
+      }
+    }
+  }
+
+  for (const path of directory.subdirs) {
+    if (path === 'node_modules' || path.startsWith('.')) {
+      continue;
+    }
+
+    yield* visit(directory.dir(path));
+  }
+}
+
+function hasPropertyWithValue(node: ts.Expression, name: string, value: unknown): boolean {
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/migrations/update-12/update-web-workers.ts
```diff
@@ -0,0 +1,75 @@
...
+function* visit(directory: DirEntry): IterableIterator<ts.SourceFile> {
+  for (const path of directory.subfiles) {
+    if (path.endsWith('.ts') && !path.endsWith('.d.ts')) {
+      const entry = directory.file(path);
+      if (entry) {
+        const content = entry.content;
+        if (content.includes('Worker')) {
+          const source = ts.createSourceFile(
+            entry.path,
+            content.toString().replace(/^\uFEFF/, ''),
```
kyliau, line 20: Could you please add a comment to explain why replacing the byte order mark is necessary?

packages/schematics/angular/migrations/update-12/update-web-workers.ts
```diff
@@ -0,0 +1,75 @@
...
+import { DirEntry, Rule, UpdateRecorder } from '@angular-devkit/schematics';
+import * as ts from '../../third_party/github.com/Microsoft/TypeScript/lib/typescript';
+
+function* visit(directory: DirEntry): IterableIterator<ts.SourceFile> {
+  for (const path of directory.subfiles) {
+    if (path.endsWith('.ts') && !path.endsWith('.d.ts')) {
+      const entry = directory.file(path);
+      if (entry) {
+        const content = entry.content;
+        if (content.includes('Worker')) {
```
kyliau, line 17: should we check `new Worker(` instead to reduce the likelihood of processing unnecessary files?

packages/schematics/angular/migrations/update-12/update-web-workers.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+export default function (): Rule {
+  return (tree) => {
+    for (const sourceFile of visit(tree.root)) {
+      let recorder: UpdateRecorder | undefined;
+
+      ts.forEachChild(sourceFile, function analyze(node) {
+        if (
```
kyliau, line 46: Please add a comment to show the expression we are looking for: `new Worker('./app.worker', ...)`

packages/schematics/angular/migrations/update-12/update-web-workers.ts
```diff
@@ -0,0 +1,75 @@
...
+          ts.isNewExpression(node) &&
+          ts.isIdentifier(node.expression) &&
+          node.expression.text === 'Worker' &&
+          node.arguments?.length === 2 &&
```
kyliau, line 50: Is the second parameter mandatory in Webpack Worker?
In the browser version the second parameter is optional: https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker
If possible please add link to the Webpack API.

packages/schematics/angular/migrations/update-12/update-web-workers.ts
clydin: There can be an arbitrary amount of whitespace between `new` and `Worker` which would require a regular expression and negatively affect performance.  `includes` is significantly faster than using a regular expression.

packages/schematics/angular/migrations/update-12/update-web-workers.ts
clydin: done

kyliau (approved):

PR #20536, opened 2021-04-14
build: bump version to v12.0.0-rc.0

kyliau:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@angular/devkit-repo",
-  "version": "12.0.0-next.9",
+  "version": "12.0.0-rc.0",
```
josephperrott, line 3: We are not moving to `rc` yet, we should still be doing a `next` release

josephperrott (changes requested):

package.json
kyliau: This is for the release next week :)

josephperrott: Now that I have looked at the calendar... we are only doing feature freeze for one week before moving to rc.  My bad.

josephperrott (approved):
LGTM

PR #20549, opened 2021-04-15
fix(@angular-devkit/build-angular): avoid double build optimizer processing

clydin:
TypeScript files had the potential to be processed twice by the build optimizer. This did not affect the output code but could lead to longer production build times. The build optimizer is now configured in one centralized location for both TypeScript and JavaScript files. The Webpack configuration partial for TypeScript support is also reduced to one common function for both AOT and JIT as a result.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
@@ -5,7 +5,6 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { buildOptimizerLoaderPath } from '@angular-devkit/build-optimizer';
 import { getSystemPath } from '@angular-devkit/core';
 import { CompilerOptions } from '@angular/compiler-cli';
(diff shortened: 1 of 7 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
```diff
@@ -78,8 +77,13 @@ function createIvyPlugin(
   });
 }
 
-export function getNonAotConfig(wco: WebpackConfigOptions) {
-  const { tsConfigPath } = wco;
+export function getTypeScriptConfig(wco: WebpackConfigOptions) {
+  const { buildOptions, tsConfigPath } = wco;
+  const aot = !!buildOptions.aot;
+
+  if (aot) {
```
alan-agius4, line 84: We should always run this irrespective if it's AOT or JIT. Otherwise there is the risk that during JIT mode users use VE runtime instead of Ivy runtime, if the packages have not been processed by NGCC.

alan-agius4 (approved):
One little comment

alan-agius4 (approved):

PR #20552, opened 2021-04-16
Webpack 5 Refactoring and fixes

alan-agius4:
See each commit


packages/angular_devkit/build_angular/src/utils/bundle-calculator_spec.ts
```diff
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+import { StatsCompilation } from 'webpack';
```
alan-agius4, line 8: We should get back to these tests and clean them up.

packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
```diff
@@ -167,46 +149,54 @@ const init: any = (config: any, emitter: any) => {
...
+        warnings: true,
+      });
 
-    if (typeof callback === 'function') {
-      callback();
+      logger.error(statsErrorsToString(statsJson, { colors: true }));
+
+      // Notify potential listeners of the compile error.
+      emitter.emit('compile_error', {
+        errors: statsJson.errors,
```
clydin, line 170: `statsJson.errors` can be an object with a message property.  Does the `compile_error` event support that?

alan-agius4: Mapped to retain current behaviour.

clydin (approved):

PR #20556, opened 2021-04-17
fix(@angular/cli): Support XDG Base Directory Specfication

joshuachp:
Reading throw  #16034 the angular config should be placed in the `$XDG_CONFIG_HOME/angular/` directory, but if the environmental variable $XDG_CONFIG_HOME is set the first check in the `xdgConfigHome` function will only put it in `$XDG_CONFIG_HOME` instead of the `angular/` directory.

Also renamed the config file from `.angular-config.json` to `config.json`  when it is in the `~/.config/angular/` directory since it doesn't need to be hidden.

### DIFF
--- a/packages/angular/cli/utilities/config.ts
+++ b/packages/angular/cli/utilities/config.ts
@@ -47,14 +47,24 @@ function getSchemaLocation(): string {
 
 export const workspaceSchemaPath = getSchemaLocation();
 
-const configNames = [ 'angular.json', '.angular.json' ];
+const configNames = ['angular.json', '.angular.json'];
 const globalFileName = '.angular-config.json';
 
 function xdgConfigHome(home: string, configFile?: string): string {
   // https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+  const xdgConfigHome = process.env['XDG_CONFIG_HOME'] || path.join(home, '.config');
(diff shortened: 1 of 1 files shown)


joshuachp:
@googlebot I signed it!

clydin (changes requested):
It does appear this new behavior is correct based on the underlying specification.
However, since this would break existing users, both scenarios would need to be temporarily supported (two major versions is the typical timeframe). The older configuration location should also be converted to the new location to _fix_ existing cases.
Could you update the PR with the above support for existing users?

joshuachp:
I re added support for the old file locations.
Should I now automatically move the old files to the new location or is better to just print a warning to the user?

clydin:
I think a warning would suffice in this case.

packages/angular/cli/utilities/config.ts
```diff
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { json, workspaces } from '@angular-devkit/core';
+import { join, json, workspaces } from '@angular-devkit/core';
```
clydin, line 8: The `join` addition here appears unused.

joshuachp:
I added the warning using `console.warn`, since it seems overwritten in the index file by the logger.

clydin (approved):
LGTM.

PR #20565, opened 2021-04-19
Angular Webpack Compiler transformer cleanup

clydin:
See individual commits.


packages/ngtools/webpack/src/transformers/make_transform.ts
```diff
@@ -1,108 +0,0 @@
...
-import {
-  AddNodeOperation,
-  OPERATION_KIND,
-  RemoveNodeOperation,
-  ReplaceNodeOperation,
-  StandardTransform,
-  TransformOperation,
-} from './interfaces';
-
-export function makeTransform(
```
alan-agius4, line 19: 😃 😃

alan-agius4 (approved):

PR #20569, opened 2021-04-20
fix(@angular-devkit/build-angular): change several builder options defaults

alan-agius4:
BREAKING CHANGE:

A number of browser and server builder options have had their default values changed. The aim of these changes is to reduce the configuration complexity and support the new "production builds by default" initiative.

**Browser builder**
| Option                                 | Previous default value    | New default value |
|----------------------------------------|---------------------------|-------------------|
| optimization                           | false                     | true              |
| aot                                    | false                     | true              |
| buildOptimizer                         | false                     | true              |
| sourceMap                              | true                      | false             |
| extractLicenses                        | false                     | true              |
| namedChunks                            | true                      | false             |
| vendorChunk                            | true                      | false             |

**Server builder**
| Option        | Previous default value | New default value |
|---------------|------------------------|-------------------|
| optimization  | false                  | true              |
| sourceMap     | true                   | false             |


clydin (approved):
Maybe change the first line in the commit message slightly:
>A number of browser and server builder options have had their default values changed. The aim of these changes is to reduce the configuration complexity and support the new "production builds by default" initiative.

PR #20572, opened 2021-04-20
fix(@angular-devkit/build-angular): recover from CSS optimization errors

alan-agius4:


clydin (approved):

packages/angular_devkit/build_angular/src/webpack/plugins/optimize-css-webpack-plugin.ts
```diff
@@ -86,33 +86,40 @@ export class OptimizeCssWebpackPlugin {
...
+              addWarning(compilation, text);
+            }
+
+            let newSource;
+            if (output.map) {
+              newSource = new sources.SourceMapSource(
+                output.css,
+                file,
+                // tslint:disable-next-line: no-any
+                output.map.toString() as any,
```
clydin, line 109: nit: I think the casting can be dropped here

PR #20596, opened 2021-04-21
build: bump version to v11.2.11

josephperrott:


dgp1130 (commented):
I'm forgetting, do we update the `latest-versions.ts` now? The other PR for `master` only updated `package.json`. Is there a difference between the branches?

The [documentation](https://github.com/angular/angular-cli/blob/master/docs/process/release.md#post-release-version-update) is actually a little contradictory here, but I think we only update `latest-versions.ts` in the actual release commit?

josephperrott:
We update the `Angular` value in the object during the actual release commit.

The `DevkitBuid*` values have to be updated after the release occurs because they rely on the packages existing on npm during testing.

The other PR for the `12.0.x` branch does not include it because it is automatically generated now ([link](https://github.com/angular/angular-cli/blob/12.0.x/packages/schematics/angular/utility/latest-versions.ts#L17-L20))

dgp1130 (approved):
Ah gotcha, thanks for the clarification.

PR #20601, opened 2021-04-22
fix(@schematics/angular): remove jasmine-spec-reporter from default workspace

cexbrayat:
The dependency is only needed for protractor tests, so it should only be added by the e2e schematics.


alan-agius4 (commented):
Can you also move `ts-node` please?

cexbrayat:
@alan-agius4 Sure, done 👍

alan-agius4 (approved):
LGTM, thanks.

PR #20660, opened 2021-04-29
test: replace deprecated `tslint` linter with `eslint`

clydin:
`tslint` is now replaced as the linter for the repository and `eslint` is used in its place.
The eslint configuration maintains the existing behavioral rules as well as adding many recommended rules from both `eslint` and `@typescript-eslint` via their respective exported recommended configurations. By leveraging the recommended configurations, the repository's configuration has also been simplified. Development scripts and unit tests also can now have specific rule overrides which reduces the amount of manually introduced disabling comments throughout the code.


.eslintrc.json
```diff
@@ -0,0 +1,105 @@
...
+    "import/no-extraneous-dependencies": ["error", { "devDependencies": false }],
+    "import/no-unassigned-import": ["error", { "allow": ["symbol-observable"] }],
+    "import/order": [
+      "error",
+      {
+        "alphabetize": { "order": "asc" },
+        "groups": [["builtin", "external"], "parent", "sibling", "index"]
+      }
+    ],
+    "max-len": [
```
alan-agius4, line 41: Drive by comment, do we need this, since line length is enforced via prettier.

clydin: There are cases where a long line cannot be split and will be retained.  This rule complements the formatter by ensuring that such lines are within reason.  Some examples include regular expressions, strings, and certain comments (URLs, etc.).

alan-agius4 (approved):
For the record, ESLint appears to be 39.5% slower even though we now have less "complex' limiting rules.

PR #20703, opened 2021-05-05
fix(@angular/cli): propagate update's force option to package managers

clydin:
When the CLI update command's `--force` option is used, the underlying package manager will also be executed with its force option. This behavior is especially important with the advent of npm 7 which will fail installation if any peer dependency version ranges are mismatched.


lodenrogue (approved):
Any way to select one and not the other?

alan-agius4 (approved):

PR #20705, opened 2021-05-06
build: bump version to v12.0.0

kyliau:


alan-agius4:
I noticed that bumping post release in this case will break the caretaking workflow. This is because if we merge this, `11.2.0` branch will become the LTS and 12.0.0 will be promoted to the patch branch.

This, I think will cause confusion since we will diverage from the workflow that is used in Angular proper, universal and components repos.

PR #20709, opened 2021-05-06
test: several fixes to update-8 E2E test

alan-agius4:
Previously in `update-8` test we used the Angular CLI in node_modules which caused incorrect result.
    
We now don't run `ng update @angular/core --next` or `ng update @angular/core`. This is due to our bumping strategy, which causes a period were `@angular/cli@latest` (v12.0.0) `@angular/core@latest` (v11.2.x) are of different major/minor version on the local NPM server. This causes `ng update` to fail.
    
NB: `ng update @angula/cli` will still cause `@angular/core` packages to be updated.

### DIFF
--- a/tests/legacy-cli/e2e/tests/update/update-8.ts
+++ b/tests/legacy-cli/e2e/tests/update/update-8.ts
@@ -16,7 +16,7 @@ export default async function () {
     // Update Angular to 9
     await installPackage('@angular/cli@8');
     const { stdout } = await ng('update', '@angular/cli@9.x', '@angular/core@9.x');
-    if (!stdout.includes('Executing migrations of package \'@angular/cli\'')) {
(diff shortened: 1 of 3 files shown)


tests/legacy-cli/e2e/tests/update/update-8.ts
```diff
@@ -5,8 +5,6 @@ import { ng, noSilentNg } from '../../utils/process';
 import { isPrereleaseCli, useCIChrome, useCIDefaults } from '../../utils/project';
 
 export default async function () {
-  const extraUpdateArgs = await isPrereleaseCli()  || true ? ['--next', '--force'] : [];
```
alan-agius4, line 8: Opps always truthy!!!

clydin (approved):

PR #20722, opened 2021-05-08
fix(@angular-devkit/schematics): detect package manager

fresto32:
The Angular Devkit Schematics defaults to NPM when no package manager option is passed into the Package Manager Executor.

I think that we can improve on this by attempting to detect the package manager by the presence of lock files. 

Please let me know what you think and how it can be improved.


clydin:
However, this is actually handled at a higher level by the schematic runtime consumer which has more complete knowledge of the project/workspace.
The Angular CLI has logic to determine the preferred package manager which can be found here: https://github.com/angular/angular-cli/blob/master/packages/angular/cli/utilities/package-manager.ts#L34

PR #20740, opened 2021-05-11
perf(@angular-devkit/build-angular): execute dart-sass in a worker

clydin:
The dart-sass Sass implementation will now be executed in a separate worker thread. The wrapper worker implementation provides an interface that can be directly used by Webpack's `sass-loader`.
The worker implementation allows dart-sass to be executed in its synchronous mode which can be up to two times faster than its asynchronous mode. The worker thread also allows Webpack to continue other bundling tasks while the Sass stylesheets are being processed.
Initial testing indicated a ~14% build time improvement for AIO.

### DIFF
--- a/packages/angular_devkit/build_angular/src/sass/worker.ts
+++ b/packages/angular_devkit/build_angular/src/sass/worker.ts
@@ -1,0 +1,65 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { ImporterReturnType, Options, renderSync } from 'sass';
+import { MessagePort, parentPort, receiveMessageOnPort, workerData } from 'worker_threads';
+
+/**
+ * A request to render a Sass stylesheet using the supplied options.
+ */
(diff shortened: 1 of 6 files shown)


alan-agius4:
This is really looking good. 

I am project that I am testing I also saw a similar improvement. The compilation time improved by around `~15%` and in some cases even up to `18%`.

packages/angular_devkit/build_angular/src/sass/worker.ts
```diff
@@ -0,0 +1,65 @@
...
+parentPort.on('message', ({ id, hasImporter, options }: RenderRequestMessage) => {
+  try {
+    if (hasImporter) {
+      // When a custom importer function is present, the importer request must be proxied
+      // back to the main thread where it can be executed.
+      // This process must be synchronous from the perspective of dart-sass. The `Atomics`
+      // functions combined with the shared memory `importSignal` and the Node.js
+      // `receiveMessageOnPort` function are used to ensure synchronous behavior.
+      options.importer = (url, prev) => {
+        Atomics.store(importerSignal, 0, 0);
```
alan-agius4, line 50: Clever way, to make the importer sync 👍

alan-agius4 (approved):

PR #20749, opened 2021-05-13
fix(@angular-devkit/build-angular): fix error with inline styles when running extract-i18n

terencehonles:
fixes: #20750

When extracting inline styles in Angular 12 there seems to be no CSS loader specified and there will be something like the following message:

```
angular-resource://:2:4 - Error: Module parse failed: Unexpected token (2:4) You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|
>     :host {
|       display: block;
|       height: 0;
```

This change adds a `webpack.NormalModuleReplacementPlugin` in extract-i18n for `angular-resource://` handling them as the non inline styles are handled.

### DIFF
--- a/packages/angular_devkit/build_angular/src/extract-i18n/index.ts
+++ b/packages/angular_devkit/build_angular/src/extract-i18n/index.ts
@@ -237,6 +237,10 @@ export async function execute(
             /\.(css|scss|sass|styl|less)$/,
             path.join(__dirname, 'empty-export-default.js'),
           ),
+          new webpack.NormalModuleReplacementPlugin(
+            /^angular-resource:\/\//,
+            path.join(__dirname, 'empty-export-default.js'),
+          ),
         ],
       });
 


alan-agius4 (commented):
Thanks for this.

Can you please update the commit message to the following;
```
fix(@angular-devkit/build-angular): fix extracting inline styles when running extract-i18n

When extracting inline styles in Angular 12 there seems to be no CSS loader specified and there will be something like the following message:
    ```
angular-resource://:2:4 - Error: Module parse failed: Unexpected token (2:4)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
>     :host {
|       display: block;
|       height: 0;
    ```

This change adds the mime type to the `angular-resource://` url in order to allow extract-i18n to provide a `webpack.NormalModuleReplacementPlugin` regex for the style URLs to handle them as the non inline styles are handled.

fixes: #20750
```

alan-agius4 (approved):
LGTM

terencehonles:
@alan-agius4 was just doing that right now. I think the title is now a little more accurate

alan-agius4:
@terencehonles can you add the PR description and the fixes clause as part of the commit message please?

alan-agius4:
@terencehonles can you add the PR description as commit message body and the fixes clause as part of the commit footer please?

packages/angular_devkit/build_angular/src/extract-i18n/index.ts
```diff
@@ -237,6 +237,10 @@ export async function execute(
             /\.(css|scss|sass|styl|less)$/,
             path.join(__dirname, 'empty-export-default.js'),
           ),
+          new webpack.NormalModuleReplacementPlugin(
+            /^angular-resource:\/\/text\/(x-less|x-sass|x-scss|css)/,
```
clydin, line 241: Only styles use this syntax so you can actually ignore all `angular-resource` URIs.  This removes the need to modify more than one package.

terencehonles: Ok, I wasn't sure if it might be used for anything else so I figured it would be safer to pass the mime type, but I'll adjust it.

terencehonles:
> @terencehonles can you add the PR description as commit message body and the fixes clause as part of the commit footer please?

Sorry, I had thought I had (for the description) and missed the trailer, updated.

alan-agius4 (approved):
Thanks for your contribution.

clydin (approved):

PR #20756, opened 2021-05-13
perf(@angular-devkit/build-angular): enable opt-in usage of file system cache 

alan-agius4:
With this change we enable Webpack's filesystem cache, this important because `terser-webpack-plugin`, `css-minimizer-webpack-plugin` and `copy-webpack-plugin` all use Webpacks' caching API to avoid additional processing during the 2nd cold build.

This changes causes `node_modules` to be treated as immutable. Webpack will avoid hashing and timestamping them, assume the version is unique and will use it as a snapshot.

To opt-in using the experimental persistent build cache use the`NG_PERSISTENT_BUILD_CACHE` environment variable. 

```
NG_PERSISTENT_BUILD_CACHE=1 ng serve
```

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/environment-options.ts
+++ b/packages/angular_devkit/build_angular/src/utils/environment-options.ts
@@ -80,6 +80,13 @@ export const cachingBasePath = (() => {
   return cacheVariable;
 })();
 
+// Persistent build cache
+const persistentBuildCacheVariable = process.env['NG_PERSISTENT_BUILD_CACHE'];
+export const persistentBuildCacheEnabled =
+  !cachingDisabled &&
+  isPresent(persistentBuildCacheVariable) &&
+  isEnabled(persistentBuildCacheVariable);
+
 // Build profiling
 const profilingVariable = process.env['NG_BUILD_PROFILING'];
 export const profilingEnabled = isPresent(profilingVariable) && isEnabled(profilingVariable);
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -10,7 +10,13 @@ import {
   BuildOptimizerWebpackPlugin,
   buildOptimizerLoaderPath,
 } from '@angular-devkit/build-optimizer';
+import {
+  GLOBAL_DEFS_FOR_TERSER,
+  GLOBAL_DEFS_FOR_TERSER_WITH_AOT,
+  VERSION as NG_VERSION,
+} from '@angular/compiler-cli';
 import * as CopyWebpackPlugin from 'copy-webpack-plugin';
+import { createHash } from 'crypto';
 import { createWriteStream, existsSync, promises as fsPromises } from 'fs';
 import * as path from 'path';
 import { ScriptTarget } from 'typescript';
@@ -20,6 +26,7 @@ import {
   ContextReplacementPlugin,
   ProgressPlugin,
   RuleSetRule,
+  WebpackOptionsNormalized,
   debug,
 } from 'webpack';
 import { AssetPatternClass } from '../../browser/schema';
@@ -31,6 +38,7 @@ import {
   allowMinify,
   cachingDisabled,
   maxWorkers,
+  persistentBuildCacheEnabled,
   profilingEnabled,
   shouldBeautify,
 } from '../../utils/environment-options';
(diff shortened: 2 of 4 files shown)


packages/angular_devkit/build_angular/src/utils/environment-options.ts
```diff
@@ -80,6 +80,13 @@ export const cachingBasePath = (() => {
   return cacheVariable;
 })();
 
+// Persistent build cache
+const persistentBuildCacheVariable = process.env['NG_PERSISTENT_BUILD_CACHE'];
```
alan-agius4, line 84: I am open to change the name of this env variable.

packages/angular_devkit/build_angular/src/webpack/utils/stats.ts
```diff
@@ -160,6 +160,12 @@ function generateBuildStats(hash: string, time: number, colors: boolean): string
   return `Build at: ${w(new Date().toISOString())} - Hash: ${w(hash)} - Time: ${w('' + time)}ms`;
 }
 
+// We use this cache because we can have multiple builders running in the same process,
+// where each builder has different output path.
+
+// Idle, we shoud create the logging callback as a factory, but that would need a refactoring.
```
clydin, line 166:
```suggestion
// Ideally, we should create the logging callback as a factory, but that would need a refactoring.
```

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -497,3 +495,38 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
+  if (persistentBuildCacheEnabled) {
+    const packageVersion = require('../../../package.json').version;
+
+    return {
+      type: 'filesystem',
+      cacheDirectory: findCachePath('angular-webpack'),
+      maxMemoryGenerations: 1,
+      // We use the version as name, webpack our webpack condfiguration are too
+      // dynamic and shared among different build types test, build and serve.
+      // None of which are "named".
```
clydin, line 512:
```suggestion
      // We use the versions and build options as the cache name. The Webpack configurations are too
      // dynamic and shared among different build types: test, build and serve.
      // None of which are "named".
```

packages/angular_devkit/build_angular/src/utils/environment-options.ts
clydin: I'm fine with that name.

I think this should interact with `NG_BUILD_CACHE` though which should probably override.  Maybe add it as the first check in `getCacheSettings`?

clydin (changes requested):

packages/angular_devkit/build_angular/src/utils/environment-options.ts
alan-agius4: It already interacts with `NG_BUILD_CACHE`

In the below check 
```
const persistentBuildCacheVariable = process.env['NG_PERSISTENT_BUILD_CACHE'];
export const persistentBuildCacheEnabled =
  !cachingDisabled &&
  isPresent(persistentBuildCacheVariable) &&
  isEnabled(persistentBuildCacheVariable);
```

`cachingDisabled` is the value of `NG_BUILD_CACHE`.

alan-agius4: https://github.com/angular/angular-cli/blob/573160352c83e0460b816a65ceb8ad94f7ce2f3f/packages/angular_devkit/build_angular/src/utils/environment-options.ts#L69-L88

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -497,3 +495,38 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
+        .update(packageVersion)
+        .update(wco.projectRoot)
+        .update(JSON.stringify(wco.tsConfig))
+        .update(JSON.stringify(wco.buildOptions))
+        .update(supportedBrowsers.join(''))
+        .digest('base64'),
+    };
+  }
+
+  if (wco.buildOptions.watch) {
```
clydin, line 524:
```suggestion
  if (wco.buildOptions.watch && !cachingDisabled) {
```
To maintain existing `NG_BUILD_CACHE` behavior.

alan-agius4: Good catch

clydin (approved):

splincode:
@alan-agius4 hi!, what do you think about support additional property in `angularCompilerOptions`?

so that caching can be easily configured

tsconfig.json

```json5
{
  "angularCompilerOptions": {
     // ...
     persistentBuildCache: true
  }
}
```

alan-agius4:
Hi @splincode,

This is not really a compiler option, so `angular.json` would be a better place. That said, we did discuss this and for the time being we don't want to add this to the API for a number of reasons.

- Having this in an file will make it harder to opt-in/opt-out in different environments
- This is a still experimental. We want to understand the impact of this and if in v13 we should abandon, turn it on by default, or leave it as opt-in.

AFDevMike:
Can anyone confirm this will (potentially) improve `--prod / --configuration production` as well?

PR #20774, opened 2021-05-13
fix(@schematics/angular): remove --prod option from README template

acohenOT:
New apps using Angular 12 defaults `ng build` to use `--production configuration` (formerly `--prod`)


alan-agius4 (approved):

alan-agius4:
Thanks for your contribution.

PR #20780, opened 2021-05-14
fix(@angular-devkit/build-angular): enable webpack Trusted Types support

bjarkler:
As reported in #20113, webpack triggers a Trusted Types violation when lazy-loading is used. To mitigate that, enable webpack's Trusted Types module (introduced in https://github.com/webpack/webpack/pull/9856) using 'angular#bundler' as the policy name.

### DIFF
--- a/tests/legacy-cli/e2e/tests/misc/http-headers.ts
+++ b/tests/legacy-cli/e2e/tests/misc/http-headers.ts
@@ -1,0 +1,42 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { ng } from '../../utils/process';
+import { updateJsonFile } from '../../utils/project';
+
+export default async function () {
+  // This test ensures that ng e2e serves the HTTP headers that are configured
+  // in the 'headers' field of the serve options. We do this by serving the
+  // strictest possible CSP headers (default-src 'none') which blocks loading of
+  // any resources (including scripts, styles and images) and should cause ng
+  // e2e to fail with a CSP-related error, which is asserted below.
+
+  await updateJsonFile('angular.json', (json) => {
+    const serve = json['projects']['test-project']['architect']['serve'];
+    if (!serve['options']) serve['options'] = {};
+    serve['options']['headers'] = {
+      'Content-Security-Policy': "default-src 'none'",
+    };
+  });
+
+  let errorMessage = null;
+  try {
+    await ng('e2e');
+  } catch (error) {
+    errorMessage = error.message;
+  }
+
+  if (!errorMessage) {
+    throw new Error(
+      'Application loaded successfully, indicating that the CSP headers were not served.',
+    );
+  }
+  if (!errorMessage.match(/Refused to load/)) {
+    throw new Error('Expected to see CSP loading failure in error logs.');
+  }
+}
--- a/packages/angular_devkit/build_angular/src/webpack/configs/browser.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/browser.ts
@@ -87,6 +87,7 @@ export function getBrowserConfig(wco: WebpackConfigOptions): webpack.Configurati
         : 'web',
     output: {
       crossOriginLoading,
+      trustedTypes: 'angular#bundler',
     },
     optimization: {
       runtimeChunk: 'single',
--- a/tests/legacy-cli/e2e/tests/misc/trusted-types.ts
+++ b/tests/legacy-cli/e2e/tests/misc/trusted-types.ts
@@ -1,0 +1,78 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { appendToFile, prependToFile, replaceInFile, writeFile } from '../../utils/fs';
+import { ng } from '../../utils/process';
+import { updateJsonFile } from '../../utils/project';
+
+export default async function () {
+  // Add app routing.
+  // This is done automatically on a new app with --routing.
+  await prependToFile('src/app/app.module.ts', `import { RouterModule } from '@angular/router';`);
+  await replaceInFile(
+    'src/app/app.module.ts',
+    `imports: [`,
+    `imports: [ RouterModule.forRoot([]),`,
+  );
+  await appendToFile('src/app/app.component.html', '<router-outlet></router-outlet>');
+
+  // Add lazy route.
+  await ng('generate', 'module', 'lazy', '--route', 'lazy', '--module', 'app.module');
+
+  // Add lazy route e2e
+  await writeFile(
+    'e2e/src/app.e2e-spec.ts',
+    `
+     import { browser, logging, element, by } from 'protractor';
+
+     describe('workspace-project App', () => {
+       it('should display lazy route', async () => {
+         await browser.get(browser.baseUrl + '/lazy');
+         expect(await element(by.css('app-lazy p')).getText()).toEqual('lazy works!');
+       });
+
+       afterEach(async () => {
+         // Assert that there are no errors emitted from the browser
+         const logs = await browser.manage().logs().get(logging.Type.BROWSER);
+         expect(logs).not.toContain(jasmine.objectContaining({
+           level: logging.Level.SEVERE,
+         }));
+       });
+     });
+   `,
+  );
+
+  const testCases = [
+    {
+      aot: false,
+      csp: `trusted-types angular angular#unsafe-bypass angular#unsafe-jit angular#bundler; require-trusted-types-for 'script';`,
+    },
+    {
+      aot: true,
+      csp: `trusted-types angular angular#unsafe-bypass angular#bundler; require-trusted-types-for 'script';`,
+    },
+  ];
+
+  for (const { aot, csp } of testCases) {
+    await updateJsonFile('angular.json', (json) => {
+      const architect = json['projects']['test-project']['architect'];
+      architect['build']['options']['aot'] = aot;
+      if (!architect['serve']['options']) architect['serve']['options'] = {};
+      architect['serve']['options']['headers'] = {
+        'Content-Security-Policy': csp,
+      };
+    });
+
+    try {
+      await ng('e2e');
+    } catch (error) {
+      console.error(`Test case AOT ${aot} with CSP header ${csp} failed.`);
+      throw error;
+    }
+  }
+}


alan-agius4 (commented):
Thanks for this @bjarkler. 

I think there needs to be some discussion around the trusted type name, if we want to provide an opt-out from trusted types, and also E2E tests need to be added.

bjarkler:
@alan-agius4 I've added E2E tests now, could you take a quick look?

The policy name is mostly for documentation purposes. I felt `webpack` was descriptive, since this is the policy that Webpack uses internally. We could go for something like `angular#webpack` to mimic the existing policies that Angular uses, but it feels wrong to me since there isn't anything Angular-specific about this particular policy, other than that it's Angular that configures Webpack to use that particular name for its policy.

I can look into implementing an opt-out configuration setting for this. We could do something similar to Webpack, and make the setting be a string that denotes the name of the policy that Webpack should use, or the empty string to disable it. Can you point me to where it would be best to make this setting live, and what would be an appropriate name for it?

alan-agius4:
@bjarkler, I added the `needs further discussion` label so that we discuss this PR in the next team meeting.

Will get back to you when we decide the desired way forward.

packages/angular_devkit/build_angular/src/protractor/trusted_types_spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    await run.stop();
+  });
+
+  it('works in a basic application using AOT', async () => {
+    const overrides = {
+      headers: {
+        'Content-Security-Policy':
+          "trusted-types angular angular#unsafe-bypass; require-trusted-types-for 'script';",
+      },
+      devServerTarget: 'app:serve:production',
```
IgorMinar, line 44: I'm not super familiar with this testing setup so it's possible that I'm wrong, but this doesn't look right to me.

The AOT mode is on by default now, and even if it wasn't the case in this test setup (see my comment above), we'd still want to selectively turn on the AOT and use the dev mode pipeline, so that we are testing the workflow that developers would use most often.

btw we could reduce the number of tests, by using lazy imports by default. It's not strictly necessary, but it would reduce the number of combinations we need to test.

Let's double check how the setup works for these tests and ensure that we are testing these four combinations:
- JIT (devserver) + lazy imports
- AOT (devserver) + lazy imports
- JIT (prodserver) + lazy imports
- AOT (prodserver) + lazy imports

packages/angular_devkit/build_angular/src/protractor/trusted_types_spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    await run.stop();
+  });
+
+  it('works in an application using lazy-loading', async () => {
+    const main = virtualFs.fileBufferToString(host.scopedSync().read(normalize('src/main.ts')));
+
+    host.writeMultipleFiles({
+      'src/lazy-module.ts': 'export const value = 42;',
+      'src/main.ts': `import('./lazy-module');\n` + main,
```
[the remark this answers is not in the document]
IgorMinar: we could even consider using a import that is more closely aligned with common lazy import patterns in Angular apps, but I don't know how tricky that would be to mimic in this test suite. @alan-agius4 might have some recommendations.

packages/angular_devkit/build_angular/src/protractor/works_spec.ts
```diff
@@ -133,4 +133,22 @@ describe('Protractor Builder', () => {
...
+        // Very strict CSP headers that should cause the application to fail to
+        // load.
+        'Content-Security-Policy': "default-src 'none';",
+      },
+    };
+
+    const run = await architect.scheduleTarget(protractorTargetSpec, overrides);
+
+    // Assert that the application did not load successfully, meaning that the
+    // CSP headers were indeed served.
```
IgorMinar, line 149: can you please clarify why this would fail? is it because the "angular" and "webpack" CSP policy creation would blow up? Please leave a comment for future reads of this code.

IgorMinar (commented):
Thanks @bjarkler for the PR!

I'm supportive of this change, but I'd like to discuss the pros and cons of turning `trustedTypes: 'webpack'` on by default, and the implications of silently turning on the `webpack` CSP policy.

The pros I'm aware of are:
- it makes using trusted types with Angular easier as all you need to do is to configure only the Angular-specific CSP headers

The cons:
- small size penalty (do we know how big is the runtime increase?)

Other clarifications:
- does `trustedTypes: 'webpack'` automatically result in webpack's dev server adding CSP headers for the `webpack` policy? if so doesn't that mean that the CSP header configuration is different between Angular CLI (no explicit `webpack` config needed) and prod servers (you must include the `webpack` CSP policy in the header config)? 
- doesn't turning on this feature by default in webpack result in breakages to existing code?

We should consider all of these things, and then determine if it's safe to turn this feature on by default in v12.1 or if we have to wait for making it the default until a future major version.

alan-agius4:
@bjarkler, just a drive by comment because I don’t want you to waste time fixing the tests.

I think that these tests needs to be revisited for a number of reasons.

1. These tests are using the old testing harness
2.  They are actually testing the protractor builder which is deprecated. 
3. Trusted types tests should be under browser/dev-server builder. Or an actually E2E.
4. The added headers public API from the protractor builder should be removed.

alan-agius4:
Couple of other points, doubt that web-dev-server adds the necessary headers considering the current webpack-dev-server  is intended to be used with webpack 4. 
 
Also, I am not too keen about having the trusted type named `webpack` as in future if we want to replace the bundler, users will need to update their server configuration. The fact that we use webpack as a bundler is an implementation detail and it should be transparent to the users if it is replaced.

IgorMinar:
@alan-agius4 thanks for the info about the testing setup! Great insights.

With regards to the header name, we'll likely need a name that is specific to javascript loading, because the policies we have are intentionally siloed, so reusing one of the existing ones is likely not a good idea. So we either stick with `webpack` or create something like `angular#bundling` or `angular#loading`.

Also, keep in mind that it's very unlikely that if we were to adopt a new bundler in the future, this new bundler would not use ES modules, and instead would still perform unsafe operations that would require CSP policy exceptions. At this point in time, this is a webpack only thing that might even go away from webpack once ESM is used to create chunks rather than the current custom module format.

alan-agius4:
**TLTR**;
- Let's enable trusted types by default without the possibility to opt-out. (Unless @IgorMinar feels otherwise)
- Size increase is negligibly 
- Let's retain the policy name (see below)

--- 
Hi @bjarkler,

Firstly let me thank you for taking this up 😀.

Yesterday, during our weekly meeting we discussed this feature and we had a couple of unanswered questions. Today, I looked into this a bit more and for visibility sake I will post here the questions we had and answers for each of them.

> Is the implementation in Webpack guarded or we need to provide polyfills if the browser doesn't support trusted types?

The implementation to add trusted types is guarded and therefore we don't need to add any addition polyfil. Extract from the runtime file

```js
((o = { createScriptURL: (e) => e }),
 "undefined" != typeof trustedTypes && trustedTypes.createPolicy
  && (o = trustedTypes.createPolicy("my-application#webpack", o))),
  o.createScriptURL(e)
```

> What is the size impact if we enable this by default?

Negligibly, around 20 bytes 

> How should we provide this feature? Opt-in/opt-out etc..

Considering the above, I think we should always enable trusted types without the possibility of opting-out.

> Name of the policy 

We discussed this, and we prefer if the policy name is something else then `webpack` and more future proof since this name needs to be configured on the server. My preference is to go with something like `angular#lazy-bundles`, `angular#bundles`, `angular#bundles` etc // @IgorMinar / @clydin any preference?

Regarding testing, I think in this case it would be better to create an E2E, you can creating it under `tests/legacy-cli/e2e/tests/misc` maybe named `trusted-types`?

Here's is how I think this test should look like. (The below is just an idea to be considered as pseudo, feel free to do your thing!)
```ts
/**
 * @license
 * Copyright Google LLC All Rights Reserved.
 *
 */

import { appendToFile, prependToFile, readFile, replaceInFile, writeFile } from '../../utils/fs';
import { ng } from '../../utils/process';
import { updateJsonFile } from '../../utils/project';

  // Add app routing.
  // This is done automatically on a new app with --routing.
  await prependToFile('src/app/app.module.ts', `import { RouterModule } from '@angular/router';`);
  await replaceInFile(
    'src/app/app.module.ts',
    `imports: [`,
    `imports: [ RouterModule.forRoot([]),`,
  );
  await appendToFile('src/app/app.component.html', '<router-outlet></router-outlet>');

  // Add lazy route.
  await ng('generate', 'module', 'lazy', '--route', 'lazy', '--module', 'app.module');

  // Add lazy route e2e
  await writeFile(
    'e2e/src/app.e2e-spec.ts',
    `
     import { browser, logging, element, by } from 'protractor';
 
     describe('workspace-project App', () => {
       it('should display lazy route', async () => {
         await browser.get(browser.baseUrl + '/lazy');
         expect(await element(by.css('app-lazy-comp p')).getText()).toEqual('lazy-comp works!');
       });
 
       afterEach(async () => {
         // Assert that there are no errors emitted from the browser
         const logs = await browser.manage().logs().get(logging.Type.BROWSER);
         expect(logs).not.toContain(jasmine.objectContaining({
           level: logging.Level.SEVERE,
         }));
       });
     });
   `,
  );

  const testCases = [
    {
      aot: false,
      csp: `trusted-types angular angular#unsafe-bypass webpack; require-trusted-types-for 'script';`,
    },
    {
      aot: true,
      csp: `trusted-types angular angular#unsafe-bypass; require-trusted-types-for 'script';`,
    },
  ];

  for (const { aot, csp } of testCases) {
    await updateJsonFile('angular.json', (json) => {
      const architect = json['projects']['test-project']['architect'];
      architect['build']['options']['aot'] = aot;
      architect['serve']['options']['headers'] = {
        'Content-Security-Policy': csp,
      };
    });

    try {
      await ng('e2e');
    } catch (error) {
      console.error(`Test case AOT ${aot} with CSP header ${csp} failed.`);
      throw error;
    }
  }
}

```

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -457,6 +457,7 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
         }
       },
       chunkFilename: `[id]${targetInFileName}${hashFormat.chunk}.js`,
+      trustedTypes: 'webpack',
```
alan-agius4, line 460: NIT, I think it is better that this is added https://github.com/angular/angular-cli/blob/36b5040cc470cdb2c0554785a9a368ce3c226ad1/packages/angular_devkit/build_angular/src/webpack/configs/browser.ts#L89

Since, the common configuration is used for generating server and UT bundles where trusted types is not useful. Although adding it here, doesn't add any overhead.

bjarkler:
Thank you so much for taking a closer look at this @alan-agius4! I'm sorry I didn't have time yesterday to address your and @IgorMinar's comments, but I will be working on incorporating all of your suggestions into the PR today.

I have one question regarding the e2e tests. I was originally led to these tests through the documentation but when I ran `node ./tests/legacy-cli/run_e2e` I got the following error:
```
STDERR:                                                                                       
Command failed: npm "publish", "--tag", "next", "--registry", "http://localhost:4873"
STDERR:
...
npm ERR! code ENEEDAUTH
npm ERR! need auth This command requires you to be logged in.
npm ERR! need auth You need to authorize this machine using `npm adduser`

npm ERR! A complete log of this run can be found in:
npm ERR!     ~/.npm/_logs/2021-05-19T11_32_47_819Z-debug.log

TypeError: Cannot read property 'message' of undefined
    at ~/angular-tt/angular-cli/scripts/publish.ts:134:68
    at _exec (~/angular-tt/angular-cli/tests/legacy-cli/e2e/utils/process.ts:64:17)
    at Object.npm (~/angular-tt/angular-cli/tests/legacy-cli/e2e/utils/process.ts:182:12)
    at default_1 (~/angular-tt/angular-cli/tests/legacy-cli/e2e/setup/010-local-publish.ts:8:21)
    at ~/angular-tt/angular-cli/tests/legacy-cli/e2e_runner.ts:145:25
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```
This, coupled with the fact that these tests have `legacy` in the name, made me think that this was something that wasn't being used anymore. But do you know how to fix this error? Do I have to change the verdaccio configuration somehow? It doesn't seem straightforward to do this `npm adduser` authorization ahead of time since verdaccio seems to be run from a temporary directory.

Regarding the policy name I see your point about being bundler-agnostic. Given that, my vote would go towards either `angular#bundler` (since this is the policy that is used by the bundler) or `angular#loading` (since the bundler currently uses this policy to lazy-load bundles, as well as support [worker construction](https://webpack.js.org/guides/web-workers/) and calls to importScripts, all of which are in some way loading new scripts into the page).

IgorMinar:
On by default sounds good as long as we don't throw an error when the app runs on a browser with TT support but the server didn't send the webpack CSP header.

IgorMinar:
> On by default sounds good as long as we don't throw an error when the app runs on a browser with TT support but the server didn't send the webpack CSP header.

I spoke with @bjarkler and we agree that this is not a concern, so we can move forward with turning TT on in webpack by default.

bjarkler:
I believe I've updated the PR to take all your suggestions into account. In particular, @alan-agius4's e2e test already looked great so I used it with minor modifications, and moved the webpack configuration into browser.ts. I also added e2e tests to make sure that HTTP headers are served in e2e tests, since otherwise the Trusted Types e2e test will always pass.

IgorMinar (commented):
This looks much better! I love that we are using the lazy route scenario in the test. Thanks @bjarkler 

@alan-agius4 do you think it's necessary to explicitly test dev and prod pipelines or is it enough to rely on the e2e test default?

tests/legacy-cli/e2e/tests/misc/http-headers.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!error.message.match(/Refused to load/)) {
+      throw new Error('Expected to see CSP loading failure in error logs.');
+    }
+  }
+
+  if (loadedSuccessfully) {
+    throw new Error(
+      'Application loaded successfully, indicating that the CSP headers were not served.',
+    );
+  }
```
alan-agius4, line 41: I think the above can be simplified a bit.

```suggestion
  try {
    await ng('e2e');
  } catch (error) {
    loadedSuccessfully = false;
    if (!error.message.match(/Refused to load/)) {
      throw new Error('Expected to see CSP loading failure in error logs.');
    }
    
    throw new Error(
      'Application loaded successfully, indicating that the CSP headers were not served.',
    );
  }

```

alan-agius4:
@alan-agius4 do you think it's necessary to explicitly test dev and prod pipelines or is it enough to rely on the e2e test default?

I don't think that in this case terser will do anything which causes trusted type to behaviour differently.

alan-agius4 (approved):
LGTM, just on minor comment and can you change the commit message from `fix` to `feat` please?

Otherwise, this is great. Thanks a lot @bjarkler!

tests/legacy-cli/e2e/tests/misc/http-headers.ts
bjarkler: We only want to throw the last error if no exception was thrown by `ng('e2e')`. I don't think that can be done without an extra variable, but I tried to clean this up a little bit in my most recent push.

alan-agius4: `ng e2e` is only run once in this test case, so I am not following how a promise can be rejected multiple times.

alan-agius4: Nevermind!

alan-agius4 (approved):

PR #20798, opened 2021-05-14
fix(@angular/cli): clean node modules directory prior to updating

clydin:
Prior to performing the initial updated package installation during the `ng update` process, the workspace node modules directory will be removed. This cleaning increases the guarantees that the package manager will hoist packages into the correct locations and avoid peer dependency inconsistencies. This change should also help to ensure a more complete package update.

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -17,7 +17,7 @@ import { Command } from '../models/command';
 import { Arguments } from '../models/interface';
 import { SchematicEngineHost } from '../models/schematic-engine-host';
 import { colors } from '../utilities/color';
-import { runTempPackageBin } from '../utilities/install-package';
+import { installAllPackages, runTempPackageBin } from '../utilities/install-package';
 import { writeErrorToLogFile } from '../utilities/log-file';
 import { ensureCompatibleNpm, getPackageManager } from '../utilities/package-manager';
 import {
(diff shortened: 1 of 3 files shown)


mattlewis92:
Hey there, I landed on this PR as we hit the problem it's trying to solve. I just thought it was worth mentioning why this might cause a different set of problems, largely because semver never works perfectly and it's very common for another 3rd party package inside of `dependencies` to accidentally break apps in minor or patch releases (i.e. bugs got introduced, a typing got updated which breaks with your app, the package doesn't follow semver etc). My worry is that with this change, as it's not obvious that angular is essentially updating all dependencies, it will erode the great trust that the community has on angular upgrades not to break their apps. A couple ideas I had which might help to mitigate this:

1. Run `npm update` before wiping node_modules and the package-lock. This will make it very obvious to the user that other dependencies have been updated as those dependency changes will be reflected in the `package.json`
2. Before doing this, update the package.json to remove all the `~` and `^` from dependencies and then after the install, restore the package.json to how it was. That will give all the benefits of a clean install, as well as upgrade transitive dependencies (in my experience it's very rare to have a transitive dependency break the app after upgrading, normally it's something referenced in the package.json)
3. Prompt the user and let them choose if they want angular to completely reinstall node_modules, and warn them that it will upgrade every other dependency in their app which could potentially break things.

clydin:
We can drop the package lock aspect for now and revisit it in v13 since it is a larger change in behavior.

alan-agius4 (approved):

PR #20826, opened 2021-05-17
build: set up releasing using ng-dev

josephperrott:
Note: This does not update the instructions for releasing yet as we are not yet using this for releases.


clydin (approved):
Nice.
Much less complicated than I was expected.

PR #20854, opened 2021-05-19
perf(@ngtools/webpack): reduce non-watch mode TypeScript diagnostic analysis overhead

clydin:
When not in a watch mode, the analyis performed by TypeScript to improve incremental type checking can be avoided by creating an abstract builder program that only wraps the underlying TypeScript program.
Performance enhancements in the upcoming TypeScript 4.3 may remove the need for this.  However, TypeScript 4.3 is not yet released and is not yet supported. In addition, TypeScript 4.2 will continue to be supported throughout the v12 major even when TypeScript 4.3 is also supported.


alan-agius4 (approved):
LGTM

PR #20874, opened 2021-05-20
fix(angular_devkit): Update postcss dependency to 8.2.15

jcesarmobile:
I'm having problems installing "@angular-devkit/build-angular": "~0.1102.13" because it uses postcss 8.2.14, but also cssnano 5.0.2, which uses  cssnano-preset-default@5.1.0, which has a peer dependency to postcss@"^8.2.15", so npm 7 fails to install because of peer dependency mismatch

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -50,7 +50,7 @@
     "ora": "5.3.0",
     "parse5-html-rewriting-stream": "6.0.1",
     "pnp-webpack-plugin": "1.6.4",
-    "postcss": "8.2.14",
+    "postcss": "8.2.15",
     "postcss-import": "14.0.0",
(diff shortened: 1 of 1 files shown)


jcesarmobile:
@googlebot I signed it!

naderalfakesh (approved):

dogandipcin (approved):

alan-agius4 (changes requested):
Thanks for your contribution, but there needs to be some changes to be able to land this PR.

- Can you please update the commit message as per our [guidelines](https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md#commit)?
- Update the version `postcss` in the root `package.json` and update also `yarn.lock` file

jcesarmobile:
I did it from github site, so can't make those changes, sorry, just wanted to unblock all the people that can't install this package on npm 7.

PR #20881, opened 2021-05-21
feat(@angular/cli): show Node.js version support status in version command

### LINKED ISSUE #20879: Include node js version support status in `ng version`
# 🚀 Feature request

- [x] version

### Description

The `ng version` command should surface whether the version of node being used in conjunction with the projects angular dependencies is supported.

I believe we have a couple of options in the output of `ng version`:

1. Note when an unsupported version is used alongside the Node version stamp
```sh
...
Angular CLI: 12.0.0
Node: 8.2.1 (Unsupported)
Package Manager: yarn 1.22.10
OS: linux x64
...
```

2. Print a warning at the end of the output.
```sh
...
@schematics/angular               12.0.0
rxjs                              6.6.7
typescript                        4.2.4

Warning: The current version of Node (8.2.1) is not supported by Angular
```

clydin:
Unsupported versions of Node.js will now show an unsupported warning when the `ng version` command is executed.
Currently Node.js major versions 12 and 14 are considered supported and tested.

Closes #20879

### DIFF
--- a/packages/angular/cli/commands/version-impl.ts
+++ b/packages/angular/cli/commands/version-impl.ts
@@ -13,6 +13,11 @@ import { colors } from '../utilities/color';
 import { getPackageManager } from '../utilities/package-manager';
 import { Schema as VersionCommandSchema } from './version';
 
+/**
+ * Major versions of Node.js that are officially supported by Angular.
+ */
+const SUPPORTED_NODE_MAJORS = [12, 14];
+
 interface PartialPackageInfo {
   name: string;
   version: string;
@@ -29,6 +34,9 @@ export class VersionCommand extends Command<VersionCommandSchema> {
     try {
       workspacePackage = require(path.resolve(this.context.root, 'package.json'));
     } catch {}
+
+    const [nodeMajor] = process.versions.node.split('.').map((part) => Number(part));
+    const unsupportedNodeVersion = !SUPPORTED_NODE_MAJORS.includes(nodeMajor);
 
     const patterns = [
       /^@angular\/.*/,
@@ -104,7 +112,7 @@ export class VersionCommand extends Command<VersionCommandSchema> {
     this.logger.info(
       `
       Angular CLI: ${ngCliVersion}
-      Node: ${process.versions.node}
+      Node: ${process.versions.node}${unsupportedNodeVersion ? ' (Unsupported)' : ''}
       Package Manager: ${await this.getPackageManager()}
       OS: ${process.platform} ${process.arch}
 
@@ -134,6 +142,12 @@ export class VersionCommand extends Command<VersionCommandSchema> {
         .join('\n')}
     `.replace(/^ {6}/gm, ''),
     );
+
+    if (unsupportedNodeVersion) {
+      this.logger.warn(
+        `Warning: The current version of Node (${process.versions.node}) is not supported by Angular.`,
+      );
+    }
   }
 
   private getVersion(moduleName: string): string {
--- a/tests/legacy-cli/e2e/tests/misc/version.ts
+++ b/tests/legacy-cli/e2e/tests/misc/version.ts
@@ -1,7 +1,7 @@
 import { deleteFile } from '../../utils/fs';
 import { ng } from '../../utils/process';
 
-export default async function() {
+export default async function () {
   const { stdout: commandOutput } = await ng('version');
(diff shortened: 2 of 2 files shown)


josephperrott (approved):
LGTM

alan-agius4 (commented):
Instead of adding an additional place where we need to update once we change support for Node

Can we instead of `engines.node` from package JSON and if the current Node doesn't fall within the specified semver, using the semver package we issue a warning?

alan-agius4 (approved):

petebacondarwin:
If I understand him correctly, I think I agree with @alan-agius4 here. Instead of hiding this information in the `version-impl.ts` file, how about putting it in the package.json:

```
"engines": {
  "node.js": ">12...",
  "node.js:supported": "12.x||14.x",
  ...
}
```

Then it is available to developers and external tooling if they want to read or consume it, while not causing a hard-error if a different version is used.

packages/angular/cli/commands/version-impl.ts
```diff
@@ -134,6 +142,12 @@ export class VersionCommand extends Command<VersionCommandSchema> {
         .join('\n')}
     `.replace(/^ {6}/gm, ''),
     );
+
+    if (unsupportedNodeVersion) {
+      this.logger.warn(
+        `Warning: The current version of Node (${process.versions.node}) is not supported by Angular`,
```
alan-agius4, line 148: OCD NIT

```suggestion
        `Warning: The current version of Node (${process.versions.node}) is not supported by Angular.`,
```

josephperrott:
@petebacondarwin @alan-agius4 I think I read Alan's proposal a bit differently which was to use the values in `engines.node` to determine the semver matching.


I actually tried to do this yesterday after I made the feature request initially before Charles ended up creating this.  What I had was (roughly):

```ts
private isNodeSupportedForPackages(moduleNames: string[]): boolean {
  const nodeEnginesEntries = new Set<string>();
  let isUnsupportedVersion = false;
  for (const moduleName of moduleNames) {
    // Find the package in the workspace
    try {
      const packagePath = require.resolve(`${moduleName}/package.json`, {
        paths: [this.context.root],
      });
      if (packagePath) {
        const { engines } = require(packagePath) as { engines: { node: string } };
        nodeEnginesEntries.add(engines.node.replace(/(\>=?)/g, '^'));
      }
    } catch {}
  }
  nodeEnginesEntries.forEach((entry: string) => {
    isUnsupportedVersion =
      isUnsupportedVersion || !semver.satisfies(process.versions.node, entry);
  });
  return isUnsupportedVersion;
}

```

The concern/issue notably is the requirement of doing something like `engines.node.replace(/(\>=?)/g, '^')` to make our check slightly more restrictive than the actual engines node field.

alan-agius4:
> @petebacondarwin @alan-agius4 I think I read Alan's proposal a bit differently which was to use the values in `engines.node` to determine the semver matching.
> 
> I actually tried to do this yesterday after I made the feature request initially before Charles ended up creating this. What I had was (roughly):
> 
> ```ts
> private isNodeSupportedForPackages(moduleNames: string[]): boolean {
>   const nodeEnginesEntries = new Set<string>();
>   let isUnsupportedVersion = false;
>   for (const moduleName of moduleNames) {
>     // Find the package in the workspace
>     try {
>       const packagePath = require.resolve(`${moduleName}/package.json`, {
>         paths: [this.context.root],
>       });
>       if (packagePath) {
>         const { engines } = require(packagePath) as { engines: { node: string } };
>         nodeEnginesEntries.add(engines.node.replace(/(\>=?)/g, '^'));
>       }
>     } catch {}
>   }
>   nodeEnginesEntries.forEach((entry: string) => {
>     isUnsupportedVersion =
>       isUnsupportedVersion || !semver.satisfies(process.versions.node, entry);
>   });
>   return isUnsupportedVersion;
> }
> ```
> 
> The concern/issue notably is the requirement of doing something like `engines.node.replace(/(\>=?)/g, '^')` to make our check slightly more restrictive than the actual engines node field.

@josephperrott / @petebacondarwin, following a convo with @clydin on Slack, we decided that for the time being we roll with the current approach. And indeed I did mean to use the `node` property in the `engines` field and not create a custom key.

clydin:
Custom engine fields can also cause install-time errors.  We ran into that issue when `pnpm` was briefly added as an engine.

PR #20885, opened 2021-05-21
fix(@ngtools/webpack): re-emit component stylesheet assets

### LINKED ISSUE #20882: Angular 12 : Loading svg using url(../../assets/images/app_logo.svg) is not working in Scss
# 🐞 Bug report

- [ ] new
- [ ] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: Angular 10

### Description

Recently I have created a new application using angular 12 in which I am getting the following bug,

I am using SVG image as `background: url('../../assets/images/app_logo.svg') no-repeat;` in `auth.component.scss` file, When I do any changes in this file it renders it in the first load properly but when I do any changes in other files, it is not able to get the image and gives `GET http://localhost:4200/app_logo.svg 404 (Not Found)`.

During this time I am getting the following error in the console 
image

Previously it was working in angular 10.
I have checked these #18115, #18013, #18041 and #18043 related issues but not found any proper solution.

## 🔬 Minimal Reproduction

In `AppComponent` template add one div and load any svg file from `assets` in that div as background 
```
<!-- app.component.html -->
<div class="root"></div>
```
```
// app.component.scss
.root{
    background: url('../assets/images/app_logo.svg') no-repeat;
    height: 200px;
}
```
Serve your application, when you save changes in `app.component.scss` it will render the image properly. Now update anything in any other file let's take in `app.component.html`, Now when an application is reloaded it will not show the logo SVG image.

## 🔥 Exception or Error

image

## 🌍 Your Environment

<pre><code>

Angular CLI: 12.0.1
Node: 12.17.0
Package Manager: npm 6.14.4
OS: win32 x64

Angular: 12.0.1
... forms, material, platform-browser, platform-browser-dynamic
... router

@angular-devkit/architect       0.1200.1
@angular-devkit/build-angular   12.0.1
@angular-devkit/core            12.0.1
@angular-devkit/schematics      12.0.1
@schematics/angular             12.0.1
</code></pre>

**Anything else relevant?**

alan-agius4:
With this change we re-emit assets referenced in component stylesheets which where uneffected by the change that re-triggered a re-compilation.

Since we cache the the result of processed component CSS, during a re-compilation `postcss-cli-resources` plugin will not run which causes assets to be to emit. With this change we now cache the asset and re-emit them on every change.

Closes #20882

### DIFF
--- a/packages/ngtools/webpack/src/resource_loader.ts
+++ b/packages/ngtools/webpack/src/resource_loader.ts
@@ -9,7 +9,7 @@
 import { createHash } from 'crypto';
 import * as path from 'path';
 import * as vm from 'vm';
-import { Compilation, EntryPlugin, NormalModule, library, node, sources } from 'webpack';
+import { Asset, Compilation, EntryPlugin, NormalModule, library, node, sources } from 'webpack';
 import { normalizePath } from './ivy/paths';
 
 interface CompilationOutput {
@@ -25,6 +25,8 @@ export class WebpackResourceLoader {
 
   private fileCache?: Map<string, CompilationOutput>;
   private inlineCache?: Map<string, CompilationOutput>;
+  private assetCache?: Map<string, Asset>;
+
   private modifiedResources = new Set<string>();
   private outputPathCounter = 1;
 
@@ -32,6 +34,7 @@ export class WebpackResourceLoader {
     if (shouldCache) {
       this.fileCache = new Map();
       this.inlineCache = new Map();
+      this.assetCache = new Map();
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/ngtools/webpack/src/resource_loader.ts
```diff
@@ -200,6 +222,14 @@ export class WebpackResourceLoader {
 
           parent.warnings.push(...childCompilation.warnings);
           parent.errors.push(...childCompilation.errors);
+          for (const { info, name, source } of childCompilation.getAssets()) {
+            if (info.sourceFilename === undefined) {
+              throw new Error(`'${name}' asset info 'sourceFilename' is 'undefined'.`);
+            }
+
+            parent.updateAsset(name, source, info);
```
clydin, line 230: I don't think this update call is needed.  The assets are emitted with essentially the same loop right before this callback is executed.  If there is any overlap with the cached assets, there should be an error before this is updated.
https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L526-L528

clydin (approved):

PR #20913, opened 2021-05-23
perf(@angular-devkit/build-angular): load postcss-preset-env configuration once

JoostK:
Loading the postcss-preset-env plugin includes building a mapping of unsupported
browsers by feature, which is somewhat expensive. In large compilations, this mapping
would be recomputed for each postcss-loader instance, as the plugin was recreated
for each loader invocation. By extracting the plugin instance outside of the dynamic
plugin computation, this overhead is avoided.

image

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -158,6 +158,11 @@ export function getStylesConfig(wco: WebpackConfigOptions): webpack.Configuratio
   }
 
   const { supportedBrowsers } = new BuildBrowserFeatures(wco.projectRoot);
+  const postcssPresetEnvPlugin = postcssPresetEnv({
+    browsers: supportedBrowsers,
+    autoprefixer: true,
+    stage: 3,
+  });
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -158,6 +158,11 @@ export function getStylesConfig(wco: WebpackConfigOptions): webpack.Configuratio
   }
 
   const { supportedBrowsers } = new BuildBrowserFeatures(wco.projectRoot);
+  const postcssPresetEnvPlugin = postcssPresetEnv({
```
alan-agius4, line 161: You can just push to `extraPostcssPlugins`.

JoostK: I was contemplating that but went with this, as I felt `extraPostcssPlugins` was more for conditional plugins. Anyway, I just tried pushing to `extraPostcssPlugins` but then the types are incompatible; presumably because `postcssPresetEnv` actually depends on postcss 7.

alan-agius4 (approved):
Thanks!

PR #20944, opened 2021-05-25
refactor(@ngtools/webpack): use Webpack inline resource matching for inline resources

clydin:
Webpack provides a method to construct a special request string that will remap a loader chain to appear to be from a specific resource path. This requires a custom loader, however, it allows rule matching to leverage the existing file extension based approach and allows providing a path to other loaders which simplifies relative request handling.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
@@ -51,20 +51,20 @@ function createIvyPlugin(
     }
   }
 
-  let inlineStyleMimeType;
+  let inlineStyleFileExtension;
   switch (buildOptions.inlineStyleLanguage) {
     case 'less':
-      inlineStyleMimeType = 'text/x-less';
+      inlineStyleFileExtension = 'less';
       break;
     case 'sass':
-      inlineStyleMimeType = 'text/x-sass';
+      inlineStyleFileExtension = 'sass';
       break;
     case 'scss':
-      inlineStyleMimeType = 'text/x-scss';
+      inlineStyleFileExtension = 'scss';
       break;
     case 'css':
     default:
-      inlineStyleMimeType = 'text/css';
+      inlineStyleFileExtension = 'css';
       break;
   }
 
@@ -74,7 +74,7 @@ function createIvyPlugin(
     fileReplacements,
     jitMode: !aot,
     emitNgModuleScope: !optimize,
-    inlineStyleMimeType,
+    inlineStyleFileExtension,
   });
 }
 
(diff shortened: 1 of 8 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
```diff
@@ -74,7 +74,7 @@ function createIvyPlugin(
     fileReplacements,
     jitMode: !aot,
     emitNgModuleScope: !optimize,
-    inlineStyleMimeType,
+    inlineStyleFileExtension,
```
alan-agius4, line 77:
```suggestion
    buildOptions.inlineStyleLanguage ?? 'css',
```

packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
```diff
@@ -51,20 +51,20 @@ function createIvyPlugin(
     }
   }
 
-  let inlineStyleMimeType;
+  let inlineStyleFileExtension;
```
alan-agius4, line 54: Lines 64-69 can be removed.

alan-agius4, line 54:
```suggestion
  const inlineStyleFileExtension = buildOptions.inlineStyleLanguage ?? 'css';
```

packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
```diff
@@ -51,20 +51,20 @@ function createIvyPlugin(
...
     case 'scss':
-      inlineStyleMimeType = 'text/x-scss';
+      inlineStyleFileExtension = 'scss';
       break;
     case 'css':
     default:
-      inlineStyleMimeType = 'text/css';
+      inlineStyleFileExtension = 'css';
       break;
   }
```
alfaproject, line 69: Less verbose:
```ts
if (['less', 'sass', 'scss', 'css'].includes(buildOptions.inlineStyleLanguage)) {
  inlineStyleFileExtension = buildOptions.inlineStyleLanguage;
}
```

Or if `buildOptions.inlineStyleLanguage` is already validated based on the schema then you can simplify even more

packages/angular_devkit/build_angular/src/webpack/configs/typescript.ts
clydin: The current code and the various suggestions all suffer from a similar problem in that they conflate the semantic meaning of the language with the language's file extension (as used within the Webpack configuration).

I have an additional refactor on top of this one that retrieves style language information from an adjusted form of the language info structure in the style configuration file (https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts#L268).  That would in turn change this code into a lookup of language info and ensure that the file extension used matches the Webpack configuration's expected value.
If preferred, I can include that commit in this PR as well.

alan-agius4: Lets do that.

alan-agius4 (approved):

PR #20948, opened 2021-05-26
fix(@angular-devkit/build-angular): do not resolve web-workers in server builds

### LINKED ISSUE #20877: Issue with Worker & SSR (universal)
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, worked in v11

### Description

Application with a web worker requires 2 TS compilations, one for the app itself and one for the worker.  
This is supported by angular via the `webWorkerTsConfig` build target configuration.

We **exclude** the web worker file from the client build and include it in the `webWorkerTsConfig` file.  
This works fine since the IVY loader will try compiling it with both programs, the first emit wins.

With SSR server compilation this is not the case, we don't have a worker compilation, instead we need to handle the case
where there is no worker available in the environment and provide an alternative.

In V11 this worked fine, since we used webpack 4 and `worker-loader`. In server builds we just
don't include the `worker-loader` which means it will not detect the worker TS file module and will not
try to compile it.

Now, with the new Webpack 5 native worker support we no longer use `worker-loader`.

```ts
new Worker(new URL('../search.worker', import.meta.url), { name: 'searchWorker', type: 'module' });
```

Webpack will natively detect the `search.worker.ts` module and add it to the compilation, passing it to the IVY loader
which now has only 1 TS program configuration, which exclude the worker files so it will **fail** saying that the TS file is not part of the compiler configuration and we should add it to "include" or "files".

As a workaround, I managed to bypass this issue with file replacements and in server builds I just replace the file referencing the worker (`new URL('...')`) with a noop file that does not reference it.  
This is good for some scenarios but other might not fit this.  

I guess one can also go back to the `worker-loader` method and change the `new URL` syntax to a literal string.

## 🌍 Your Environment

<pre><code>
@angular-devkit/architect       0.1200.0
@angular-devkit/build-angular   12.0.0
@angular-devkit/core            12.0.0
@angular-devkit/schematics      12.0.0
@angular/flex-layout            12.0.0-beta.34
@nguniversal/common             12.0.0
@nguniversal/express-engine     12.0.0
@schematics/angular             12.0.1
ng-packagr                      12.0.0
rxjs                            6.6.3
</code></pre>

**Anything else relevant?**

alan-agius4:
Web-workers are not supported on the server and therefore we should not try to request these files from the Angular plugin.
Closes #20877

### DIFF
--- a/packages/angular_devkit/build_angular/src/server/base_spec.ts
+++ b/packages/angular_devkit/build_angular/src/server/base_spec.ts
@@ -146,4 +146,31 @@ describe('Server Builder', () => {
 
     await run.stop();
   });
+
+  it('should not try to resolve web-worker', async () => {
+    host.writeMultipleFiles({
+      'src/app/app.worker.ts': `
+        /// <reference lib="webworker" />
+  
+        const foo: string = 'hello world';
+        addEventListener('message', ({ data }) => {
+          postMessage(foo);
+        });
+      `,
+      'src/main.server.ts': `
+        if (typeof Worker !== 'undefined') {
+          const worker = new Worker(new URL('./app/app.worker', import.meta.url), { type: 'module' });
+          worker.onmessage = ({ data }) => {
+            console.log('page got message:', data);
+          };
+          worker.postMessage('hello');
+        }
+      `,
+    });
+
+    const run = await architect.scheduleTarget(target);
+    const output = (await run.result) as ServerBuilderOutput;
+    expect(output.success).toBe(true);
+    await run.stop();
+  });
 });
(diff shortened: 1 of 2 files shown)


clydin:
To disable parsing of the `new Worker`, the server Webpack config partial could be augmented with:
```
    module: {
      parser: {
        javascript: {
          worker: false,
        }
      }
    },
```
I think that would essentially mimic the prior v11 behavior.

alan-agius4:
I should have mentioned why, I did go with this approach instead of disabling the parser. I updated the PR description and commit message to reflect this.

clydin (approved):

PR #20952, opened 2021-05-26
fix(@angular-devkit/build-angular): show `--disable-host-check` warning only when not using `disableHostCheck`

### LINKED ISSUE #20951: Angular 9 `--disableHostCheck` flag has no effect when `--host` is passed (or `--configuration`)
Created a new project with Angular CLI 9.1.15 (but also tested with version 11.2.13, no difference at all):

```none
$ npm run start -- --host 0.0.0.0 

> angular-nine-template@0.0.0 start /home/rkrisztian/tmp/angular-nine-template
> ng serve "--host" "0.0.0.0"

WARNING: This is a simple server for use in testing or debugging Angular applications
locally. It hasn't been reviewed for security issues.

Binding this server to an open connection can result in compromising your application or
computer. Using a different host than the one passed to the "--host" flag might result in
websocket connection issues. You might need to use "--disableHostCheck" if that's the
case.
(...)
```

```none
$ npm run start -- --host 0.0.0.0 --disableHostCheck true 

> angular-nine-template@0.0.0 start /home/rkrisztian/tmp/angular-nine-template
> ng serve "--host" "0.0.0.0" "--disableHostCheck" "true"

WARNING: This is a simple server for use in testing or debugging Angular applications
locally. It hasn't been reviewed for security issues.

Binding this server to an open connection can result in compromising your application or
computer. Using a different host than the one passed to the "--host" flag might result in
websocket connection issues. You might need to use "--disableHostCheck" if that's the
case.
WARNING: Running a server with --disable-host-check is a security risk. See https://medium.com/webpack/webpack-dev-server-middleware-security-issues-1489d950874a for more information.
(...)
```

Why does it still warn me?

I can't do anything about it, becuase if I move the host parameter to `angular.json`, and use `--configuration` to specify the configuration I want to use, the problem is still back, and `--disableHostCheck` cannot do anything about it!

Please help, this is really frustrating. Although right now we are on Angular version 9, we do plan to upgrade all the way up to 11 (maybe to 12 later), but it takes time...

alan-agius4:
Closes #20951

### DIFF
--- a/packages/angular_devkit/build_angular/src/dev-server/index.ts
+++ b/packages/angular_devkit/build_angular/src/dev-server/index.ts
@@ -226,6 +226,7 @@ export function serveWebpackBrowser(
     }
 
     if (
+      !options.disableHostCheck &&
       options.host &&
       !/^127\.\d+\.\d+\.\d+/g.test(options.host) &&
       options.host !== 'localhost'


packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -233,17 +233,21 @@ export function serveWebpackBrowser(
...
-
-        Binding this server to an open connection can result in compromising your application or
-        computer. Using a different host than the one passed to the "--host" flag might result in
-        websocket connection issues. You might need to use "--disableHostCheck" if that's the
-        case.
       `);
+
+      if (!options.disableHostCheck) {
+        logger.warn(tags.stripIndent`
+            Binding this server to an open connection can result in compromising your application or
```
clydin, line 240: The "Binding this server...` sentence should probably stay in the warning above since it is the triggering element of the above warning.

PR #20958, opened 2021-05-26
docs: fold post-release version bump with the release commit

dgp1130:
The process has changed and post-release bumps are no longer necessary.

Also updated the doc to include a section on changing shifts with the caveat that you must be logged in to the `@angular.io` account to have edit access to the calendar.

/cc release caretakers @josephperrott @kyliau


kyliau (approved):

josephperrott (approved):
LGTM

PR #20960, opened 2021-05-27
fix(@angular-devkit/build-angular): provided earlier build feedback in console

### LINKED ISSUE #20957: Better CLI feedback for e.g. ng serve.
# 🚀 Improvement / Bug report / Feature request.

### Command

- [x] [serve](https://angular.io/cli/serve)

### Description

Just a small request: `ng serve` (and maybe all other commands) should print some text instantly after the command is executed.

demo

You can see that after enter `ng serve` there is a long waiting time without any response. Appears to be dead. Annoying.
I would prefer an initial message directly after the command was entered. Something like: 

```text
ng serve
Builds and serves your app, rebuilding on file changes.
[-] Generating browser bla bla ...
```

This text comes from the [docs](https://angular.io/cli/serve). But we could also just print: `Preparing...` or the joke of the day. 
Seriously, no response feedback on command execution feels weird and seems to be frozen. 

Maybe this should be implemented on all commands?

alan-agius4:
With this change we provide earlier feedback in the console during a build. The setup phase can take some times if the project has a large set of dependencies.

We also move several warnings in the dev-server to an earlier phase so that we don't clutter the progress indicator.

Closes #20957

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -68,6 +68,42 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
     buildOptions.differentialLoadingNeeded,
   );
 
+  if (buildOptions.progress) {
+    const spinner = new Spinner();
+    spinner.start(`Generating ${platform} application bundles (phase: setup)...`);
+
+    let previousPercentage: number | undefined;
+    extraPlugins.push(
+      new ProgressPlugin({
+        handler: (percentage: number, message: string) => {
+          if (previousPercentage === 1 && percentage !== 0) {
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -68,6 +68,42 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
     buildOptions.differentialLoadingNeeded,
   );
 
+  if (buildOptions.progress) {
+    const spinner = new Spinner();
+    spinner.start(`Generating ${platform} application bundles (phase: setup)...`);
```
alan-agius4, line 73: This is the root change.

clydin (approved):

PR #20982, opened 2021-05-28
[12.0.x] build: update Webpack to version 5.38.1 

alan-agius4:
With this change we update to `webpack` to `5.38.1`, this also updates `webpack-sources` to `2.3.0`, The latter is causing OOM errors on our CI, but it in real projects the memory usage only increased by a small fraction. Also, the OOM errors don't manifest themselves when the entire test suit is run locally.

Therefore with this change we also disable sourcemap genertation for most of the browser builds to speed up the tests and reduce memery usage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     ]
   },
   "resolutions": {
-    "**/@types/copy-webpack-plugin/webpack": "5.36.2"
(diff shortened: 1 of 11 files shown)


clydin (approved):

SirJosh1987 (commented):
English plz

PR #20999, opened 2021-05-31
Babel build optimizer

clydin:
The build optimizer passes that are still relevant for Ivy are now implemented as babel plugins.  Using babel has several advantages.  The `babel-loader` is already present within the build pipeline and many packages will already need to be processed by it.  The `babel-loader` also provides the necessary infrastructure to setup babel and link it to the Webpack build system.  This removes the need for the infrastructure code within the build optimizer loader and reduces the build optimizer to only a set of transformation plugins for babel to consume.  The babel plugin architecture also allows for less code to represent similar transformations and provides a variety of helper utilities which further reduces the amount code needed.
The passes are now implemented to be safer when transforming code.  Enum values which contain potential side effects are also no longer altered.  Enums are wrapped in less destructive manner which reduces the likelihood of incorrectly emitting the transformed enum. Class static fields which contain potential side effects are no longer wrapped in pure annotated IIFEs.  Known safe Angular static fields are also wrapped and Angular static fields which are not required in production code are also dropped.
The conversion of the passes not only reduces the amount of code to maintain but also provides a noticeable performance improvement.  Initial tests of a production build of AIO resulted in a ~10% total build time reduction.

The new babel-based build optimizer passes are now enabled when targetting ES2015 and higher. ES5 targets will currently continue to use the previous build optimizer. ES5 support will require additional refactoring of the downlevelling process as the new build optimizer passes are designed for ES2015 code.


packages/angular_devkit/build_angular/src/babel/plugins/adjust-angular-classes_spec.ts
```diff
@@ -0,0 +1,555 @@
...
+}): void {
+  const result = transform(input, {
+    configFile: false,
+    babelrc: false,
+    plugins: [[adjustAngularClasses, options]],
+  });
+  if (!result) {
+    fail('Expected babel to return a transform result.');
+  } else {
+    expect(prettier.format(result.code, { parser: 'babel' })).toEqual(
```
alan-agius4, line 32: Nifty trick!

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -301,7 +301,7 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
   }
 
   let buildOptimizerUseRule: RuleSetRule[] = [];
-  if (buildOptions.buildOptimizer) {
+  if (buildOptions.buildOptimizer && wco.scriptTarget < ScriptTarget.ES2015) {
```
alan-agius4, line 304: Seeing that all Angular libraries are published in ES2015, should we in this case enable the new BO for non TS sources and leave the old BO for TS sources?

packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
```diff
@@ -80,6 +85,21 @@ export default custom<AngularCustomOptions>(() => {
...
+      if (optimize) {
+        const angularPackage = /[\\\/]node_modules[\\\/]@angular[\\\/]/.test(this.resourcePath);
+        customOptions.optimize = {
+          // Angular packages provide additional tested side effects guarantees and can use
+          // otherwise unsafe optimizations.
+          looseEnums: angularPackage,
+          pureTopLevel: angularPackage,
+          // JavaScript modules that are marked as side effect free are considered to have
+          // no decorators that contain non-local effects.
+          wrapDecorators: !!this._module?.factoryMeta?.sideEffectFree,
```
alan-agius4, line 97: Previously we only ran BO, on known TypeScript libraries https://github.com/angular/angular-cli/blob/a41f2ed85c0f01be9a7a99d2fd40c495cd6248c3/packages/angular_devkit/build_optimizer/src/build-optimizer/webpack-plugin.ts#L28-L29

Should we retain this behaviour?

packages/angular_devkit/build_angular/src/babel/plugins/adjust-angular-classes.ts
```diff
@@ -0,0 +1,300 @@
...
+ *
+ * @param origin The starting NodePath location for analyzing siblings.
+ * @param classIdentifier The identifier node that represents the name of the class.
+ * @param allowDecorators Whether to allow decorators to be wrapped.
+ * @returns An object containing the results of the analysis.
+ */
+function analyzeClassSiblings(
+  origin: NodePath,
+  classIdentifier: types.Identifier,
+  allowDecorators: boolean,
```
alan-agius4, line 72: Maybe, a more self-explanatory name would be `wrapDecorators` or `allowWrappingDecorators`

packages/angular_devkit/build_angular/src/babel/plugins/adjust-angular-classes_spec.ts
```diff
@@ -0,0 +1,555 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { transform } from '@babel/core';
+import { default as adjustAngularClasses } from './adjust-angular-classes';
```
alan-agius4, line 10: I find the name here slightly misleading as this does more than just modify "Angular" classes. It visits all TypeScript classes and amended static properties, and wraps decorators.

packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
```diff
@@ -101,12 +121,30 @@ export default custom<AngularCustomOptions>(() => {
...
+
+        plugins.push(require('./plugins/elide-angular-metadata').default);
+        plugins.push([
+          require('./plugins/adjust-typescript-enums').default,
+          { loose: customOptions.optimize.looseEnums },
+        ]);
+        plugins.push([
+          require('./plugins/adjust-angular-classes').default,
+          { wrapDecorators: customOptions.optimize.wrapDecorators },
+        ]);
```
alan-agius4, line 138: NIT

```suggestion
        plugins.push(
          require('./plugins/elide-angular-metadata').default,
          [
            require('./plugins/adjust-typescript-enums').default,
            { loose: customOptions.optimize.looseEnums },
          ],
          [
            require('./plugins/adjust-angular-classes').default,
            { wrapDecorators: customOptions.optimize.wrapDecorators },
          ],
        );
```

packages/angular_devkit/build_angular/package.json
```diff
@@ -12,6 +12,7 @@
     "@angular-devkit/core": "0.0.0",
     "@babel/core": "7.14.3",
     "@babel/generator": "7.14.3",
+    "@babel/helper-annotate-as-pure": "7.12.13",
```
alan-agius4, line 15: Should be added in the root level package.json.

packages/angular_devkit/build_angular/package.json
```diff
@@ -68,6 +69,7 @@
     "terser-webpack-plugin": "5.1.2",
     "text-table": "0.2.0",
     "tree-kill": "1.2.2",
+    "tslib": "2.2.0",
```
alan-agius4, line 72: Let's pin the one in the root `package.json` since Bazel doesn't support workspaces.

packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -186,6 +186,7 @@ ts_library(
         "@npm//terser-webpack-plugin",
         "@npm//text-table",
         "@npm//tree-kill",
```
alan-agius4, line 188: `@npm//@babel/helper-annotate-as-pure` is missing from here, how is the test even working?

packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
clydin: That was an attempt to workaround the unsafe nature of the existing build optimizer passes.  With the increased safety and reduced number of AST mutations, the workaround can be removed.  In the event of any edge cases, additional strictness checks can be added to the new passes.  The typings field check also generates a large amount of false positives since many libraries contain typings but are not written in TypeScript.

packages/angular_devkit/build_angular/src/babel/plugins/adjust-angular-classes_spec.ts
clydin: Indeed.  the plugin started out as Angular only but since the logic was nearly identical for the non-Angular cases they were combined.  I'll change the name to account for this.

packages/angular_devkit/build_angular/BUILD.bazel
clydin: Since that package doesn't provide typings, they were added inside `@angular-devkit/build-angular` which is all Bazel needs to compile.

alan-agius4: But isn’t this needed to run the tests?

clydin: The magic of package hoisting.  I'll add it here as well though.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
alan-agius4: Bump!

alan-agius4 (approved):
LGTM, great work.

Just one final question.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
clydin: Since this is ideally intended to be temporary, I went with the approach that appeared to require the least amount of additional logic. The old plus new BO combination is also viable but would need additional conditionally added Webpack loader rules to ensure that only application code (when targeting ES5) is processed by the old BO.

PR #21025, opened 2021-06-02
test(@angular/cli): locally publish LTS tooling packages for E2E update tests

clydin:
The tooling packages needed to perform a canonical update process are now published to the local test package registry. This removes the need to switch back and forth between the npm registry and the test registry during E2E tests. It also ensures that CLI versions are used that are compatible with the Node.js version used for the tests.


alan-agius4 (approved):
Nicely done!

PR #21037, opened 2021-06-03
fix(@angular/cli): avoid shell exec when bootstrapping update command

clydin:
This change removes the usage of the shell execution when spawning the latest CLI version when bootstrapping the update command. The absolute path of the node process is now spawned which removes the need for any shell path resolution. This also removes the need to quote and/or escape command arguments which can be error-prone. Node.js and the OS will now handle any quoting and escaping automatically and without any custom logic.


alan-agius4 (approved):
Good call

PR #21046, opened 2021-06-03
feat(@ngtools/webpack): add support for custom typescript transformers

edusperoni:
This change allows developers to add custom typescript transformers to the webpack compilation, which was previously supported by AngularCompilerPlugin and its platformTransformers option.

The rationale behind this change is that before we were able to specify custom transformers either via platformTransformers or by hacking into `_transformers` (https://medium.com/joolsoftware/custom-typescript-transformers-with-angular-cli-7f4150797e05). Since that is not available anymore, some projects are no longer able to use such transformations.

### DIFF
--- a/packages/ngtools/webpack/src/ivy/plugin.ts
+++ b/packages/ngtools/webpack/src/ivy/plugin.ts
@@ -48,6 +48,7 @@ const DIAGNOSTICS_AFFECTED_THRESHOLD = 1;
 export interface AngularWebpackPluginOptions {
   tsconfig: string;
   compilerOptions?: CompilerOptions;
+  customTransformers: ts.CustomTransformers;
   fileReplacements: Record<string, string>;
   substitutions: Record<string, string>;
   directTemplateLoading: boolean;
@@ -111,6 +112,7 @@ export class AngularWebpackPlugin {
       substitutions: {},
       directTemplateLoading: true,
       tsconfig: 'tsconfig.json',
+      customTransformers: {},
       ...options,
     };
   }
@@ -497,7 +499,10 @@ export class AngularWebpackPlugin {
       }
     }
 
-    const transformers = createAotTransformers(builder, this.pluginOptions);
+    const transformers = mergeTransformers(
+      createAotTransformers(builder, this.pluginOptions),
+      this.pluginOptions.customTransformers,
+    );
 
     const getDependencies = (sourceFile: ts.SourceFile) => {
       const dependencies = [];
@@ -613,7 +618,10 @@ export class AngularWebpackPlugin {
     ];
     diagnosticsReporter(diagnostics);
 
-    const transformers = createJitTransformers(builder, this.pluginOptions);
+    const transformers = mergeTransformers(
+      createJitTransformers(builder, this.pluginOptions),
+      this.pluginOptions.customTransformers,
+    );
 
     return {
       fileEmitter: this.createFileEmitter(builder, transformers, () => []),


dgp1130:
Hi @edusperoni, thanks for drafting this PR!

This topic has come up before on the compiler side, and @alxhub left a great explanation about it here: https://github.com/angular/angular/issues/22434#issuecomment-594074902. The biggest issue seems to be that the Angular compiler analyzes code before running transformers, so even if we support this they wouldn't be usable to modify Angular code. I think all the points he mentions are still relevant and it doesn't make sense to do this at the tooling level without explicit compiler support.

As a result, we don't think we can land this PR with the current state of world. If we ever did support custom transformers, it would probably be done at the compiler level anyways, so I'm closing this PR as a result.

dgp1130:
Hi @edusperoni, what exactly is the motivation for this change? Are you trying to update NativeScript for Webpack 5 and running into this missing option? If the motivation is getting NativeScript on Webpack 5, then I would rather focus on a solution for that on supporting that instead of scoping to this particular transformer option.

We came across [this commit](https://github.com/NativeScript/NativeScript/commit/c265c3ac3f7a9b770de86fd58cb7b2cb661570d0#diff-45f4e6ff45c508724bfdf20be521f824a2ccebbb80bc79d7d9f032221bb7d4ca) which seems to imply that there's a workaround. Would that be good enough for NativeScript or is there some critical limitation you need to solve at the Angular level?

bufke:
@edusperoni I would love to know the conclusion to this attempt at supporting code sharing in NativeScript with webpack 5.

edusperoni:
@dgp1130 Sorry for the delay, I had a couple of things going last week and couldn't focus on this issue.

To give some context, just like Zone.js relies on transforming native async/await into promises to ensure zones work, NativeScript relied on the old es5 class syntax to make a class extend a native class. Now that most apps target es6+, a design decision was made to use a `@NativeClass()` decorator that uses a typescript transformer to add the required data for these classes to work. At the time, typescript transformers were supported by AngularCompilerPlugin via `platformTransformers` .

This isn't an issue with libraries because NativeScript libraries use conventional entrypoints for the "Native" code (`plugin/*.d.ts` and `plugin/*.js`) and a different entrypoint entirely for frontend flavors (`plugin/angular` is an entrypoint that follows the APF via ng-packagr). The issue is when app developers want to use `@NativeClass()` with Angular 12 (not in libraries, but in actual applications). A component can never be a NativeClass, so it never breaks the Angular compiler.

While that workaround does make the decorator work with Angular 12, I feel that it'd be better suited as an option to ensure both backwards and future compatibility, as it only affects applications (not libraries) and requires the ejection of the webpack config to use, so we expect that people who do it to know the limitations of the approach (just like it was on Angular 11 or lower). The workaround also modifies a private API that can be changed in Angular 12+x and detecting and adapting to such changes can quickly become a maintenance nightmare when the same end result can be achieved by changing concatenating the transformers to the ones already used by the AngularWebpackPlugin.

In the end, I just don't see how adding this option *back* would affect the framework negatively.


-----
@bufke I think you got confused, this issue isn't about code sharing at all. I've made a comment about this (https://github.com/NativeScript/nativescript-angular/issues/2290#issuecomment-854226427), but the gist of it is that code sharing NativeScript+Angular with Web+Angular on the same application has always been a bad idea and instead they should be different applications that can share specific things using a "monorepo" setup (with nx, for example). That thread has more information on it and you're welcome to share your thoughts there!

dgp1130:
Hi @edusperoni, we discussed this some more in our weekly CLI triage. We don't officially support custom Webpack configs and while you can depend on `@ngtools/webpack` and leverage its APIs, they are subject to change. We try to limit breaking changes because we know people do depend on them, but across major versions breaking changes are inevitable to support Angular itself.

As a result these APIs are a bit between "public" and "private". We try not to break things where possible but we don't offer any long term support guarantees for any external use case. Adding `platformTransformers` specifically is a fairly trivial code change, but accepting it is implicitly elevating it to a more official "public" API in a package which otherwise doesn't have that strong of support contract.

We could accept this PR with the caveat that `platformTransformers` also isn't explicitly supported, but that doesn't really address the motivation behind the change:

> While that workaround does make the decorator work with Angular 12, I feel that it'd be better suited as an option to ensure both backwards and future compatibility

We'd be likely to accidentally refactor and remove this in the future and break you again because the option isn't needed for Angular. The workaround you already have is just as good and has the same support contract from Angular, so accepting this PR wouldn't change that.

Our takeaway is that accepting this PR doesn't really improve anything for NativeScript and opens the possibility for others to misinterpret and misuse it, as well as placing a higher expectation on the Angular team to support the API (whether or not that expectation will actually be met).

wszgrcy:
I'd like to add method to reset transformer or handle(eg: ComponentDecoratorHandler) .
because WechatMiniProgram can't render html by angular, so I want to change component template method to support this(ompatibility support)

PR #21047, opened 2021-06-04
feat(@schematics/angular): add migration to replace deprecated `--prod`

### LINKED ISSUE #21036: Handle --prod deprecation more gracefully
# 🚀 Feature request

### Command

- [x] update

### Description

`--prod` has been deprecated for `ng build`.  However, I would think that most apps probably have build/deploy automation scripts configured in package.json, which will be affected by this change.

I would like to see `ng update` automate the change of package.json scripts so that `--prod` is automatically changed to `--configuration production`.

The alternative is disruptive and painful for the end user... Either manually updating every single project, or developing our own automated process - resulting in varying experiences.

alan-agius4:
With this change we add migration to replace the deprecated `--prod` with `--configuration production` in the scripts section of the package.json.

Closes #21036

### DIFF
--- a/packages/schematics/angular/migrations/update-12/replace-prod-flag.ts
+++ b/packages/schematics/angular/migrations/update-12/replace-prod-flag.ts
@@ -1,0 +1,30 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Rule } from '@angular-devkit/schematics';
+import { JSONFile } from '../../utility/json-file';
+
+export default function (): Rule {
+  return (tree) => {
+    const file = new JSONFile(tree, 'package.json');
+    const scripts = file.get(['scripts']);
+    if (!scripts || typeof scripts !== 'object') {
+      return;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/migrations/update-12/replace-prod-flag.ts
```diff
@@ -0,0 +1,28 @@
...
+    const file = new JSONFile(tree, 'package.json');
+    const scripts = file.get(['scripts']);
+    if (!scripts || typeof scripts !== 'object') {
+      return;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const updatedScripts = Object.entries(scripts!).map(([key, value]) => [
+      key,
+      typeof value === 'string' ? value.replace(/ --prod/g, ' --configuration production') : value,
```
JoostK, line 23: Maybe require a word boundary after `--prod`?
```suggestion
      typeof value === 'string' ? value.replace(/ --prod(?!\w)/g, ' --configuration production') : value,
```

Or `/ --prod(?=\W)/g` would also work. This avoids false positives for e.g. a `--product` flag.

SirJosh1987 (approved):

packages/schematics/angular/migrations/update-12/replace-prod-flag.ts
alan-agius4: Good catch

clydin (approved):

PR #21049, opened 2021-06-04
refactor(@angular-devkit/build-angular): remove usage of rimraf package

alan-agius4:
Use Node.JS `rmdirSync` instead.


clydin (approved):
Should be able to remove the package from the root package.json as well now.

alan-agius4:
It’s being used by a number of e2e tests

PR #21051, opened 2021-06-04
feat(@angular-devkit/schematics): extend move rule to accept array of destinations

mruffieux:
### Current behaviour

Currently, if we want to move the same source files to multiple directories, we have to repeat the rule multiple times.

For example:

```javascript
  const toPath1 = 'a/b';
  const toPath2 = 'a/c';
  const toPath3 = 'a/d';

  const sourceFiles = url(path);
  return chain([
    apply(sourceFiles, [template({ ...options, ...strings }), move(toPath1)]),
    apply(sourceFiles, [template({ ...options, ...strings }), move(toPath2)]),
    apply(sourceFiles, [template({ ...options, ...strings }), move(toPath3)]),
  ]);
```

### New behaviour

This pull request suggest to extend `move` rule to also accept an array of destination paths:

```javascript
  const toPath1 = 'a/b';
  const toPath2 = 'a/c';
  const toPath3 = 'a/d';

  const sourceFiles = url(path);
  return chain([
    apply(sourceFiles, [template({ ...options, ...strings }), move([toPath1, toPath2, toPath3])]),
  ]);
```

### Introduce breaking changes
- [ ] Yes
- [X] No

### DIFF
--- a/packages/angular_devkit/schematics/src/rules/move.ts
+++ b/packages/angular_devkit/schematics/src/rules/move.ts
@@ -10,29 +10,42 @@ import { join, normalize } from '@angular-devkit/core';
 import { Rule } from '../engine/interface';
 import { noop } from './base';
 
-export function move(from: string, to?: string): Rule {
+export function move(from: string, to?: string | string[]): Rule {
   if (to === undefined) {
     to = from;
     from = '/';
   }
 
-  const fromPath = normalize('/' + from);
-  const toPath = normalize('/' + to);
+  if (typeof to === 'string') {
+    // For simplicity, we wrap to string into an array
+    to = new Array(to);
+  }
 
-  if (fromPath === toPath) {
+  if (typeof from === 'string' && typeof to === 'string' && from === to) {
     return noop;
   }
 
   return (tree) => {
-    if (tree.exists(fromPath)) {
-      // fromPath is a file
-      tree.rename(fromPath, toPath);
-    } else {
-      // fromPath is a directory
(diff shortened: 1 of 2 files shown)


alan-agius4:
You can achieve the same result by using the `applyToSubtree`

```ts
const sourceFiles = mergeWith(
  apply(url(path), [
    template({ ...options, ...strings }),
  ]),
);

return chain([
  applyToSubtree(toPath1, [sourceFiles]),
  applyToSubtree(toPath2, [sourceFiles]),
  applyToSubtree(toPath3, [sourceFiles]),
]);
```

mruffieux:
Oh, didn't know about `applyToSubtree`. I guess that's pretty much what I did using `branch` instead of `subtree` aha. However, the issue is the same: one has to repeat the same code for different paths. Maybe, it would make sense to apply the same idea to `applyToSubtree`, e.g.:

```typescript
[...]

return chain([applyToSubtree([toPath1, toPath2, toPath3], [sourceFiles])]);
```

What do you think?

alan-agius4:
@mruffieux, the code repetition in this case is really minimal.

Increasing the API is not something that we take lightly since increase the API surface will increase our maintenance and support efforts.

Considering that the desired outcome can be achieved using a different API we should consider this feature if there is a strong community desire for it.

mruffieux:
@alan-agius4 alright then, I understand. I kinda T-DRY to the limit but anyway, I guess we can close this.

PR #21057, opened 2021-06-06
fix(@schematics/angular): added webWorkerTsConfig into test options

InSantoshMahto:
I have found one issue while working with a web worker.
I have a simple web-worker and working perfectly with `ng build ` but in the case of the `ng test` will fail.

after some time I found that  `"webWorkerTsConfig": "tsconfig.worker.json"` missing in test options.

and also check --target with ng g web-worker CLI reference.

Today I have made a simple change in angular-cli so that it will also added same config into test if it available it will change.


alan-agius4 (approved):
Thanks for your contribution.

PR #21060, opened 2021-06-07
test: remove rimraf and fs-extra dependencies

alan-agius4:


tests/legacy-cli/e2e/utils/fs.ts
```diff
@@ -1,195 +1,145 @@
-import * as fs from 'fs-extra';
```
alan-agius4, line 1: 😱  we relied on a transitive dep

tests/legacy-cli/e2e/utils/fs.ts
```diff
@@ -1,195 +1,145 @@
...
-    .then(() => new Promise((resolve, reject) => {
-      const rd = fs.createReadStream(from);
-      rd.on('error', (err: Error) => reject(err));
+export async function copyFile(from: string, to: string): Promise<void> {
+  await createDir(dirname(to));
 
-      const wr = fs.createWriteStream(to);
-      wr.on('error', (err: Error) => reject(err));
-      wr.on('close', () => resolve());
+  return new Promise<void>((resolve, reject) => {
```
clydin, line 45: This can probably be reduced to:
`return fs.copyFile(from, to, constants.COPYFILE_FICLONE);`

PR #21074, opened 2021-06-07
build: remove unused non-bazel unit test script

clydin:
All unit testing is now performed via Bazel. This removed script is also outdated and not maintained.


alan-agius4 (commented):
<del> I was under the impression that Karma tests don’t run locally on OSX.

I actually use this to Karma tests locally.

Can you confirm that we can now run Karma test with Bazel in OSX?</del>

alan-agius4:
Superseded by https://github.com/angular/angular-cli/pull/21440

PR #21082, opened 2021-06-08
perf(@ngtools/webpack): remove usage of native private class fields

alan-agius4:
TypeScript will downlevel native private class fields in form a global variable which can cause memory retention.

```js
var _fileEmitter, _registrations;
exports.FileEmitterCollection = exports.FileEmitterRegistration = exports.AngularPluginSymbol = void 0;
exports.AngularPluginSymbol = Symbol.for('@angular-devkit/build-angular[angular-compiler]');
class FileEmitterRegistration {
    constructor() {
        _fileEmitter.set(this, void 0);
    }
    update(emitter) {
        __classPrivateFieldSet(this, _fileEmitter, emitter);
    }
    emit(file) {
        if (!__classPrivateFieldGet(this, _fileEmitter)) {
            throw new Error('Emit attempted before Angular Webpack plugin initialization.');
        }
        return __classPrivateFieldGet(this, _fileEmitter).call(this, file);
    }
}
```

### DIFF
--- a/packages/ngtools/webpack/src/ivy/symbol.ts
+++ b/packages/ngtools/webpack/src/ivy/symbol.ts
@@ -18,37 +18,37 @@ export interface EmitFileResult {
 export type FileEmitter = (file: string) => Promise<EmitFileResult | undefined>;
 
 export class FileEmitterRegistration {
-  #fileEmitter?: FileEmitter;
+  private fileEmitter?: FileEmitter;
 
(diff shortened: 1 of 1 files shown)


JoostK:
> which can cause memory retention.

Do we know this is the case? I'd expect normal fields to behave the same; as long as the instance is reachable it will retain the field and that's the same for the retention in WeakMap.

PR #21086, opened 2021-06-08
Support TypeScript 4.3

clydin:
With the support of TypeScript 4.3 within the Angular framework, the CLI now supports using TypeScript 4.3 as well.  TypeScript 4.2 also continues to be supported for existing projects.
All new projects will now use TypeScript 4.3 by default.


alan-agius4 (commented):
Since we are using TS 4.3 we should update tslib to 2.2.0 or later as previous versions are not fully compatible with the output of TS 4.3.

alan-agius4 (approved):

PR #21089, opened 2021-06-09
fix(@ngtools/webpack): ensure plugin provided Webpack instance is used

clydin:
Webpack 5 provides the Webpack instance as a property on the Webpack compiler which allows Webpack plugins to leverage the same Webpack instance that was used to initiate the build. The `AngularWebpackPlugin` now will only use the provided instance to ensure that differing Webpack instances and/or versions are not used.


alan-agius4 (approved):

alan-agius4:
@clydin this doesn't apply cleanly on the patch branch. Kindly open a new PR.

PR #21099, opened 2021-06-09
refactor(@ngtools/webpack): use Webpack loader context hook to access file emitters

clydin:
The Angular Webpack loader will now access the Angular Webpack Plugin's file emitters via the loader context directly. This is instead of using the not fully public Webpack compilation property on the loader context. This change also removes the need to directly alter the Webpack compilation object which previously added an additional property.

### DIFF
--- a/packages/ngtools/webpack/src/ivy/plugin.ts
+++ b/packages/ngtools/webpack/src/ivy/plugin.ts
@@ -51,11 +51,6 @@ export interface AngularWebpackPluginOptions {
   inlineStyleFileExtension?: string;
 }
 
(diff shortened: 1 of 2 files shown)


packages/ngtools/webpack/src/ivy/plugin.ts
```diff
@@ -314,6 +305,23 @@ export class AngularWebpackPlugin {
     });
   }
 
+  private registerWithCompilation(compilation: Compilation) {
```
alan-agius4, line 308: NIT: missing return type

alan-agius4 (approved):

PR #21153, opened 2021-06-17
build: support ES2020 API usage

clydin:
With the removal of support for Node.js v10, the TypeScript library version can be updated to ES2020. Node.js v12.11+ supports the API elements of ES2020.
The compilation target must still be set to ES2019 as the language syntax elements are not fully supported until Node.js v14+.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,7 +13,7 @@
     "skipLibCheck": true,
     "strict": true,
     "target": "es2019",
-    "lib": ["es2019"],
+    "lib": ["es2020"],
     "baseUrl": "",
     "rootDirs": [".", "./dist-schema/", "./bazel-bin/"],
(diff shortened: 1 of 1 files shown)


hiepxanh:
that  nice, I really love this PR, it help us alot, I wonder is it relevant to zone.js ? because as I remember last time zone.js block angular to stay at ES2015

alan-agius4:
@hiepxanh, this change only effects the Angular CLI source code. 

Regarding `zone.js` I think you are referring to the async/await interception issue when targeting ES2017 or greater, this is no longer an issue in CLI version 12 since we now downlevel async/await. Also, important to mentioned that `target` and `libs` are also two different settings. see https://www.typescriptlang.org/docs/handbook/compiler-options.html for more information.

alan-agius4 (approved):

PR #21158, opened 2021-06-18
fix(@schematics/angular): display warning during migrations when using third-party builders

alan-agius4:


packages/schematics/angular/migrations/update-12/update-angular-config.ts
```diff
@@ -30,29 +30,39 @@ const ServerBuilderOptions: BuilderOptionsType = [
...
+    updateWorkspace((workspace) => {
+      for (const [targetName, target, projectName] of allWorkspaceTargets(workspace)) {
+        if (
+          !target.builder.startsWith('@angular-devkit/') &&
+          !target.builder.startsWith('@nguniversal/')
+        ) {
+          context.logger.warn(tags.stripIndent`
+            "${targetName}" target in "${projectName}" project is using a third-party builder.
+            You may need to adjust the options to retain the existing behavior.
+            For more information, see https://github.com/angular/angular-cli/releases/tag/v12.0.0
```
clydin, line 43:
```suggestion
            For more information, see the breaking changes section within the release notes: https://github.com/angular/angular-cli/releases/tag/v12.0.0
```

The release notes are quite long.  It would be nice if we could link directly to the breaking changes section.

clydin (approved):

packages/schematics/angular/migrations/update-12/update-angular-config.ts
alan-agius4: Yeah that would be nice, I tried using anchor tags but it doesn’t appear to work

PR #21163, opened 2021-06-18
perf(@angular-devkit/build-angular): use combination of `esbuild` and `terser` as a JavaScript optimizer

clydin:
The javascript optimization pipeline is now a two-phase process.  `esbuild` is used in the first phase to remove the majority of the unused code and shorten identifiers in each output bundle script.  `esbuild` can accomplish this in a fraction of the time that `terser` previously required.  However, `esbuild` does not yet implement all of the optimizations that `terser` performs.  As a result, `terser` is used as a second phase to further optimize and reduce the size of the output bundle scripts.  Since `terser` is operating on a smaller input size, the time required for `terser` to complete is significantly reduced.  To further improve performance when source maps are enabled, the source map merging is now performed within the optimization workers.  By offloading the source map merging, the main process is not blocked by each script’s source map output. A maximum of four (4) optimization workers are currently used and this value can be adjusted via the `NG_BUILD_MAX_WORKERS` environment variable.

Initial tests provided a ~10% total build time improvement for both a newly created v12 project and AIO.


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -316,78 +317,43 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
-            ecma: 5,
-          },
-          output: {
-            ...terserOptions.output,
+      // The `terser-webpack-plugin` will add the minified flag to the asset which will prevent
+      // additional optimizations by the next plugin.
+      const TerserPlugin = require('terser-webpack-plugin');
+      extraMinimizers.push(
+        new TerserPlugin({
+          parallel: 1,
```
alan-agius4, line 333: I am a bit concerned about setting this value to one.  This may impact negativity large applications that depend on a large number of scripts.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -316,78 +317,43 @@ export function getCommonConfig(wco: WebpackConfigOptions): Configuration {
...
-    extraMinimizers.push(
-      new TerserPlugin({
-        parallel: maxWorkers,
-        extractComments: false,
-        exclude: globalScriptsNames,
-        terserOptions,
-      }),
+    if (
+      globalScriptsNames.length > 0 &&
+      (differentialLoadingMode || wco.scriptTarget < ScriptTarget.ES2015)
```
alan-agius4, line 324: I am not following why `differentialLoadingMode || wco.scriptTarget < ScriptTarget.ES2015` is needed here. 

- Scripts bundles are not down-levelled and they can contain a mixture of syntaxes
- We are filtering what to minimize with TerserPlugin using the `includes` option.

package.json
```diff
@@ -191,6 +193,7 @@
     "parse5-html-rewriting-stream": "6.0.1",
     "pidtree": "^0.5.0",
     "pidusage": "^2.0.17",
+    "piscina": "3.1.0",
```
alan-agius4, line 196: Very nice alternative to jest considering we no longer need to fallback using forked processes.

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+                  })
+                  .then(
+                    ({ code, name, map }) => {
+                      let optimizedAsset;
+                      if (map) {
+                        optimizedAsset = new SourceMapSource(code, name, map);
+                      } else {
+                        optimizedAsset = new OriginalSource(code, name);
+                      }
+                      compilation.updateAsset(name, optimizedAsset, { minimized: true });
```
alan-agius4, line 158: It would be great if in a followup we cache these results using Webpack's caching API.

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+   * all intermediate sourcemaps.
+   */
+  sourcemap: boolean;
+
+  /**
+   * The ECMAScript version that should be used when generated output code.
+   * The value must be lower case.
+   * Examples: `es5`, `es2015`, or `es2017`
+   */
+  target: string;
```
alan-agius4, line 51:
```suggestion
  target: keyof typeof ScriptTarget;
```

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
```diff
@@ -0,0 +1,178 @@
...
+              define[key] = String(value);
+            }
+          }
+
+          // Setup the options used by all worker tasks
+          const optimizeOptions = {
+            sourcemap: this.options.sourcemap,
+            define,
+            keepNames: this.options.keepNames,
+            target: this.options.target ? Number(this.options.target.slice(2)) : 2017,
```
alan-agius4, line 121: `Number(this.options.target.slice(2))` would be `NaN` when target is `ESNext` or `Latest`.

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-worker.ts
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * The options to use when optimizing.
+   */
+  options: {
+    advanced: boolean;
+    define?: Record<string, string>;
+    keepNames: boolean;
+    removeLicenses: boolean;
+    sourcemap: boolean;
+    target: 5 | 2015 | 2017;
```
alan-agius4, line 26: NIT, because we do use 2019 for server builds.

```suggestion
    target: 5 | 2015 | 2016 | 2017 | 2018 | 2019 | 2020;
```

alan-agius4 (changes requested):
Couple of minor comments.

Great worker! Really excited about this.

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
clydin: Yes. There is already a decent amount of new code in this PR so I didn't want to add the complexity of the Webpack 5 caching initially. Adding support for caching would be a fairly straightforward incremental improvement once this PR lands.

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
clydin: Adjusted the logic to prevent those cases from happening.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
clydin: I set it to `maxWorkers` now.  My original concern was too many workers at once but it was like that before.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
clydin: This was an attempt at an optimization for the scripts.  The global scripts should only need to be ES5 when using differential loading (the scripts are shared between es5/es2015 bundles) and when targeting ES5 directly.  Otherwise they should be able to be processed by the new plugin at the specified target.  `esbuild` will error if it detects ES2015+ content when targeting ES5 so the above conditionals bypass the problem for the ES5 cases.  I think long-term the error could be beneficial since it would notify users that the scripts are not compatible with the specified output target but that would currently be a breaking change.

Alternatively, the above conditionals can be dropped and scripts can always be treated as ES5 code.  This is actually the current behavior so it may be better to retain that for now and revisit in an upcoming major release.

alan-agius4: I think also treating it as ES5 make sense.

clydin: Updated it here.

alan-agius4: Eventually it would be nice to get rid of `terser-webpack-plugin` completely.

alan-agius4 (approved):
LGTM

PR #21174, opened 2021-06-21
Reduce usage of `require` calls in `@angular/cli` package

clydin:
These set of changes reduce the number of `require` calls through the `@angular/cli` package and will aid in the transition to ESM.  This also helps remove duplicate code as well as some now unused code.

### DIFF
--- a/packages/angular/cli/lib/init.ts
+++ b/packages/angular/cli/lib/init.ts
@@ -11,6 +11,7 @@ import 'symbol-observable';
 import * as fs from 'fs';
 import * as path from 'path';
 import { SemVer } from 'semver';
+import { VERSION } from '../models/version';
 import { colors } from '../utilities/color';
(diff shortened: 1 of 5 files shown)


packages/angular/cli/lib/init.ts
```diff
@@ -8,48 +8,13 @@
...
 // symbol polyfill must go first
-import * as fs from 'fs';
+import { promises as fs } from 'fs';
 import * as path from 'path';
 import { SemVer } from 'semver';
+import { VERSION } from '../models/version';
 import { colors } from '../utilities/color';
 import { isWarningEnabled } from '../utilities/config';
 
-// Check if we need to profile this CLI run.
```
alan-agius4, line 17: Nice!

alan-agius4 (approved):

PR #21188, opened 2021-06-24
build: update to new remote instance name for RBE

josephperrott:
Update to use remote instance name, primary_instance, for RBE


alan-agius4:
Drive by comment, we also need to backport this to all LTS branches

alan-agius4 (commented):
LGTM

alan-agius4 (approved):

PR #21209, opened 2021-06-25
feat(@angular-devkit/schematics): add UpdateBuffer2 based on magic-string

### LINKED ISSUE #21110: UpdateRecorder should support multiple adjacent inserts
# 🐞 Bug report

- [X] schematics

### Is this a regression?

No

### Description

From #20190.

`UpdateRecorder` doesn't support multiple adjacent inserts.

## 🔬 Minimal Reproduction

```typescript
it('works with multiple adjacent inserts', () => {
  const buffer = Buffer.from('Hello beautiful World');
  const entry = new SimpleFileEntry(normalize('/some/path'), buffer);

  const recorder = new UpdateRecorderBase(entry);
  recorder.remove(6, 9);
  recorder.insertRight(6, 'amazing');
  recorder.insertRight(15, ' and fantastic');
  const result = recorder.apply(buffer);
  expect(result.toString()).toBe('Hello amazing and fantastic World');
});
```

**Anything else relevant?**

Best workaround is probably to use something more comprehensive like `magic-string`, but that probably shouldn't be required for something this simple.

kyubisation:
This PR adds UpdateBuffer2 which should eventually replace UpdateBuffer. UpdateBuffer2 internally uses the magic-string library.
UpdateBuffer and related symbols have been marked as deprecated.

Closes #21110

### DIFF
--- a/packages/angular_devkit/schematics/src/utility/update-buffer.ts
+++ b/packages/angular_devkit/schematics/src/utility/update-buffer.ts
@@ -7,6 +7,8 @@
  */
 
 import { BaseException } from '@angular-devkit/core';
+import MagicString from 'magic-string';
+import { updateBufferV2Enabled } from './environment-options';
 import { LinkedList } from './linked-list';
 
 export class IndexOutOfBoundException extends BaseException {
@@ -14,6 +16,7 @@ export class IndexOutOfBoundException extends BaseException {
     super(`Index ${index} outside of range [${min}, ${max}].`);
   }
 }
+/** @deprecated Since v13.0 */
 export class ContentCannotBeRemovedException extends BaseException {
   constructor() {
     super(`User tried to remove content that was marked essential.`);
@@ -26,6 +29,7 @@ export class ContentCannotBeRemovedException extends BaseException {
  * it means the content itself was deleted.
  *
  * @see UpdateBuffer
+ * @deprecated Since v13.0
  */
 export class Chunk {
   private _content: Buffer | null;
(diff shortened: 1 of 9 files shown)


kyubisation:
This is an initial implementation of the `magic-string` changes discussed in #21110. I'm open to suggestions for improvement.
Also while this is not a breaking change to the API, it does change internal behavior (see changed unit tests). Would this be considered a breaking change?

clydin:
I'll do a more thorough review soon but I wanted to address the breaking change question. Looking at the changes to the tests, this would indeed be breaking.  As-is, this would need to go into a major version (v13 being the next in Fall 2021).  To get this in sooner, both modes would need to be supported and controlled via an option.  This option could potentially be exposed via `beginUpdate` but if there were a more appropriate place that would work as well.  This would allow the current behavior to remain the default for 12.x but still allow the new behavior for those who need/prefer it.  The default could also potentially be swapped in v13+ and then followed by an eventual removal.

kyubisation:
This fix is not urgent for me, so landing it in v13 would be fine by me. I can change the deprecation notices accordingly.

As for providing it side-by-side; A possibility would be to pass a parameter through `NodeWorkflowOptions` => `BaseWorkflowOptions` => `HostTree` => `beginUpdate`. 
I'm blanking a bit on the potential names. `useMagicString` for parameter and `UpdateRecoderMagicString` for the `UpdateRecorder` implementation?

packages/angular_devkit/core/src/utils/index.ts
```diff
@@ -10,6 +10,7 @@ import * as tags from './literals';
 import * as strings from './strings';
 
 export * from './array';
+export * from './magic-string';
```
clydin, line 13: Instead of increasing the Public API surface of `@angular-devkit/core`, `magic-string` can be added as a direct dependency `@angular-devkit/schematics`.
An entry for will need to be added here as well: https://github.com/angular/angular-cli/blob/259e26979ebc712ee08fd36fb68a9576c1e02447/packages/angular_devkit/schematics/BUILD.bazel#L40-L44

clydin:
For the option name, preference would be to treat `magic-string` as an implementation detail and keep it encapsulated if possible.  Maybe a name such as `compatibilityMode` or similar (open to other ideas) with a default to true which would use the existing methodology.

kyubisation:
An alternative for the configuration may be an environment variable (e.g. NG_LEGACY_UPDATE_RECORDER, NG_SAFE_UPDATE_RECORDER or simply NG_UPDATE_RECORDER_V2), as there are already precedents like NG_PERSISTENT_BUILD_CACHE.
This would avoid the need for the propagation.

kyubisation:
@clydin I have updated the implementation. I have reverted previous changes to the existing UpdateBuffer and created an additional UpdateBuffer2, which now both extend UpdateBufferBase. I have implemented a switch based on the environment variable `NG_UPDATE_BUFFER_V2`.
I'm still very much open to suggestions for improvement.

kyubisation:
@clydin I don't know whether the change to the goldens is acceptable in this case. I can revert the changes to HostSink, if it is a problem.

kyubisation:
@clydin With 13 looming, could I ask for another review of this PR?

clydin (approved):
Sorry for the review delay.
This looks good for v13.

Can you rebase? There appears to be a conflict with `package.json`.

packages/angular_devkit/schematics/src/utility/update-buffer.ts
```diff
@@ -26,6 +29,7 @@ export class ContentCannotBeRemovedException extends BaseException {
  * it means the content itself was deleted.
  *
  * @see UpdateBuffer
+ * @deprecated Since v12.2
```
alan-agius4, line 32: Nit: deprecated comments should be updated to `@deprecated Since v13.0`

kyubisation:
@clydin No problem (in regards to the delay). I have rebased the branch.
One question; Should I switch to usage to default to UpdateBuffer2 (e.g. rename `NG_UPDATE_BUFFER_V2` to `NG_LEGACY_UPDATE_BUFFER` and adapt accordingly) or would you like to continue using UpdateBuffer for v13?

packages/angular_devkit/schematics/src/utility/update-buffer.ts
kyubisation: Thanks for the catch. Fixed accordingly.

clydin:
Let's leave the default with the existing behavior for now.  We may switch it for v13 but we would like to do some analysis regarding the magnitude of breakage before changing it.

Thank you again for the contribution.

kyubisation:
Thank you for the review and other feedback. 😃

PR #21203, opened 2021-06-25
fix(@angular-devkit/build-angular): handle `ENOENT` and `ENOTDIR` errors when deleting outputs

alan-agius4:
Closes #21202

PR #21215, opened 2021-06-26
fix(@angular-devkit/build-angular): fix node 16 breaking when deleting the output path

terencehonles:
According to `rmdirSync` [1] the `recursive` option is deprecated. The wording looks like it was changed (compared to Node LTS [2]) to not include anything about ignoring errors on paths not existing. In the future the call should be `rmSync(..., {..., force: true})`, but when testing locally adding a "force" option does not restore the previous behavior. This isn't documented to work, but since the code original code is deprecated I was wondering if the underlying functionality was just passing all the options `rmSync` on Node 16.

This change includes an existence check to skip deleting a non existent directory.

Fixes: https://github.com/angular/angular-cli/issues/21216

[1]: https://nodejs.org/api/fs.html#fs_fs_rmdirsync_path_options
[2]: https://nodejs.org/docs/latest-v14.x/api/fs.html#fs_fs_rmdirsync_path_options

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/delete-output-dir.ts
+++ b/packages/angular_devkit/build_angular/src/utils/delete-output-dir.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { rmdirSync } from 'fs';
+import { existsSync, rmdirSync } from 'fs';
 import { resolve } from 'path';
 
 /**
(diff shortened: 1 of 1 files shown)


terencehonles:
[This](https://github.com/angular/angular-cli/commit/c0106f6e3a36bf50d0c1535ddb42fbff74b0eca7#diff-1e1929a835af943574840d813513b3003e376dc809ce396d8b9ef791c76c4d09R270) looks like it also is related, but since exceptions are ignored I don't think this change is needed

alan-agius4:
Duplicate of https://github.com/angular/angular-cli/pull/21203

PR #21226, opened 2021-06-28
perf(@angular-devkit/build-angular): cache JavaScriptOptimizerPlugin results

alan-agius4:
With this change we cache JavaScriptOptimizerPlugin results in memory or on the file system based on the caching strategy used.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
@@ -8,6 +8,7 @@
 
 import Piscina from 'piscina';
 import { ScriptTarget } from 'typescript';
+import type { Compiler, sources } from 'webpack';
 import { maxWorkers } from '../../utils/environment-options';
 
 /**
@@ -74,7 +75,7 @@ export interface JavaScriptOptimizerOptions {
 export class JavaScriptOptimizerPlugin {
   constructor(public options: Partial<JavaScriptOptimizerOptions> = {}) {}
 
-  apply(compiler: import('webpack').Compiler) {
+  apply(compiler: Compiler) {
     const { OriginalSource, SourceMapSource } = compiler.webpack.sources;
 
     compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
```diff
@@ -84,19 +85,37 @@ export class JavaScriptOptimizerPlugin {
...
+                }
+
+                const { source, map } = scriptAssetSource.sourceAndMap();
                 scriptsToOptimize.push({
                   name: scriptAsset.name,
                   code: typeof source === 'string' ? source : source.toString(),
                   map,
                 });
+
+                cachedItems[name] = cacheItem;
```
clydin, line 118: I think `cacheItem` could be stored in the object literal pushed into `scriptsToOptimize` above this line.  This should allow removal of the `cachedItems` object map.

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
```diff
@@ -84,19 +85,37 @@ export class JavaScriptOptimizerPlugin {
           stage: compiler.webpack.Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_SIZE,
         },
         async (compilationAssets) => {
+          const cache = compilation.getCache('JavaScriptOptimizerPlugin');
```
clydin, line 88: Would it be worthwhile to add a check here if caching is enabled and make `cache` possibly `undefined`?  With the potential advantage of skipping the cache related calls below.  I'm not sure if there is an easy and definitive way to determine cache enablement though.

alan-agius4: I think that should be possible via `compilation.options.experiments.cache`.

alan-agius4: Actually, it appears that will other experiments options can be accessed via `compilation.options.experiments`, `cache` is not.

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
```diff
@@ -85,19 +86,42 @@ export class JavaScriptOptimizerPlugin {
         },
         async (compilationAssets) => {
           const scriptsToOptimize = [];
+          const cache = compilation.getCache('JavaScriptOptimizerPlugin');
```
clydin, line 89:
```suggestion
          const cache = !!compilation.options.cache && compilation.getCache('JavaScriptOptimizerPlugin');
```
nit: checking here should avoid the `getCache` call if not needed.

clydin (approved):

packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-plugin.ts
alan-agius4: I was going for that but I did complicate the type in Line 100.

Will try to give it another shoot.

PR #21227, opened 2021-06-28
refactor: disable esbuild worker threads

alan-agius4:
This is a workaround for `TypeError [Error]: Cannot read property 'workerPort' of undefined`.


alan-agius4:
Included the commit in https://github.com/angular/angular-cli/pull/21221

PR #21259, opened 2021-07-01
refactor(@angular-devkit/build-angular): avoid loading Webpack for differential loading sourcemaps

clydin:
The `@ampproject/remapping` package is now used for sourcemap processing instead of Webpack for differential loading and i18n processing. This dependency is already used within the recently added JavaScript optimizer refactoring and reduces the amount of code that needs to be loaded into each worker to support differential loading source maps. Testing with large output bundles (~12MB vendor bundles) resulted in an ~8% total build time improvement when using differential loading and source maps.

### DIFF
--- a/packages/angular_devkit/build_angular/BUILD.bazel
+++ b/packages/angular_devkit/build_angular/BUILD.bazel
@@ -178,7 +178,6 @@ ts_library(
         "@npm//sass",
         "@npm//sass-loader",
         "@npm//semver",
-        "@npm//source-map",
         "@npm//source-map-loader",
         "@npm//source-map-support",
         "@npm//style-loader",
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):
LGTM, one nit.

packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -178,7 +178,6 @@ ts_library(
         "@npm//sass",
         "@npm//sass-loader",
         "@npm//semver",
-        "@npm//source-map",
```
alan-agius4, line 181: Should also be removed from https://github.com/angular/angular-cli/blob/8be19dada4dc0593a95f7309790f9baf8d213ef9/packages/angular_devkit/build_angular/BUILD.bazel#L181

clydin: I think that's the line changed here unless I missed another location?

alan-agius4 (approved):

PR #21273, opened 2021-07-02
refactor(@angular-devkit/build-angular): use Webpack provided loader types

clydin:
Webpack now provides loader function type definitions. These type definitions are now used in custom loaders within the package.
This improves type safety and behavior correctness of the loaders when used with Webpack.
The direct usage of the `loader-utils` package is also reduced as the Webpack 5 loader context now provides the `getOptions` function.

### DIFF
--- a/packages/angular_devkit/build_angular/src/extract-i18n/ivy-extract-loader.ts
+++ b/packages/angular_devkit/build_angular/src/extract-i18n/ivy-extract-loader.ts
@@ -7,26 +7,23 @@
  */
 
 import { MessageExtractor } from '@angular/localize/src/tools/src/extract/extraction';
-import { getOptions } from 'loader-utils';
 import * as nodePath from 'path';
+
+// Extract loader source map parameter type since it is not exported directly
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):

packages/angular_devkit/build_angular/src/extract-i18n/ivy-extract-loader.ts
```diff
@@ -37,20 +34,22 @@ export default function localizeExtractLoader(
...
     },
     error(...args: string[]): void {
-      loaderContext.emitError(args.join(''));
+      loaderContext.emitError(new Error(args.join('')));
     },
   };
 
   let filename = loaderContext.resourcePath;
-  if (map?.file) {
+  const mapObject =
```
alan-agius4, line 48: Looks like this like is causing the issue.

clydin: Was the call to `this.getOptions`.  Accidentally left `this.getOptions(this)` which caused Webpack to treat `this` as a JSON schema.

PR #21274, opened 2021-07-03
fix(@angular-devkit/build-angular): ensure `NG_PERSISTENT_BUILD_CACHE` always creates a cache in the specified cache directory

terencehonles:
This change fixes `NG_PERSISTENT_BUILD_CACHE` sometimes creating cache entries that live outside of the cache directory by using a hex encoding rather than a base64 encoding. This error is caused because the base64 alphabet includes `/`. According to the webpack documentation [1] the default `cacheLocation` is: `path.resolve(cache.cacheDirectory, cache.name)` which means cache names with a leading `/` would remove the `cacheDirectory` altogether.

[1]: https://webpack.js.org/configuration/other-options/#cachecachelocation

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -532,7 +532,7 @@ function getCacheSettings(
         .update(JSON.stringify(wco.tsConfig))
         .update(JSON.stringify(wco.buildOptions))
         .update(supportedBrowsers.join(''))
-        .digest('base64'),
+        .digest('hex'),
     };
   }
 


terencehonles:
@alan-agius4 I'm not sure if base64 is preferred and a prefix should just be added to the cache name. Kicking off the CI to see what tests are on it.

alan-agius4 (approved):
LGTM, although we could also be `encodeURIComponent`

terencehonles:
True, but it sounds like there's not a strong reason to use base64 and it looks like it doesn't break any tests so it probably makes sense to just leave it as hex encoding. Node 16 looks like it has base64url, but I'm not sure it's worth picking the encoding based on the node version.

PR #21155, opened 2021-06-17
fix(@angular-devkit/build-angular): control linker template sourcemapping via builder sourcemap options

clydin:
This change allows the linker sourcemap behavior to be controlled by the Webpack sourcemap configuration. For example, if a vendor file is being processed and vendor sourcemaps are disabled, the linker will now skip its internal sourcemap loading and processing steps.

PR #21284, opened 2021-07-05
Revert "fix(@angular-devkit/build-angular): control linker template sourcemapping via builder sourcemap options"

### LINKED ISSUE #21271: Partially-compiled multi-entry Ivy library missing in vendor sourcemaps
# 🐞 Bug report

- [ ] new
- [x] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

I don't know, we've moved directly from Angular 10 (with libraries published in ViewEngine build) to Angular 12

### Description

We have a rather big library with multiple entry points (package.json/ng-packagr definition per module, built and published as one library). We have application with vendor sourcemaps enabled. In the vendor sourcemaps, none of our modules with any components are included. Modules containing only services or directives are included. This is true for both our usual applications, and simple minimal app generated .

Other situations:
- library compiled in full mode:  source maps included
- vendor maps disabled: source map pointing to js files included (as expected)
- 3rd party libraries published in ViewEngine version: source maps included
- published vs npm linked library: identical behaviour
- aot true/false: identical behaviour

So it seems the source maps are somehow lost the moment angular does the local (in-app) compilation of the library components' templates.

When tested with minimal library generated by ng cli, this actually works correctly, so it is caused by something in our setup (or dependencies), but at this point I am unable to figure out what (I've matched tsconfig/angular.json settings as well as I could). 

I did inspect the source maps generated in the library and as far as I can tell, they're correct (paths pointing to the correct folders etc)

## 🔬 Minimal Reproduction

As the library is big and not public, it's not easy for me to provide reproduction. I will try to see if I can duplicate the issue by expanding the sample library; but at this point I would appreciate pointers that would help us to identify the problem. I am willing to debug the compilation pipeline, if there is an obvious starting point.

## 🌍 Your Environment

<pre><code>
Angular CLI: 12.1.0
Node: 12.18.2
Package Manager: npm 6.14.11
OS: win32 x64

Angular: 12.1.1
... router

@angular-devkit/architect       0.1201.0
@angular-devkit/build-angular   12.1.0
@angular-devkit/core            12.1.0
@angular-devkit/schematics      12.1.0
@angular/cli                    12.1.0
@schematics/angular             12.1.0
ng-packagr                      12.1.0
</code></pre>

**Anything else relevant?**

No

alan-agius4:
Reverts angular/angular-cli#21155

Following a debugging and investigation with @petebacondarwin it appears that when the external template handling in the linker generates Babel AST nodes that reference the external template files which breaks Babel when it tried to flatten final source-map, which ends up no emitting any source-maps.

As an interim solution we should revert this. 

Closes #21271

### DIFF
--- a/packages/angular_devkit/build_angular/src/babel/presets/application.ts
+++ b/packages/angular_devkit/build_angular/src/babel/presets/application.ts
@@ -20,7 +20,6 @@ export interface ApplicationPresetOptions {
   angularLinker?: {
     shouldLink: boolean;
     jitMode: boolean;
-    sourcemap: boolean;
   };
 
   forceES5?: boolean;
@@ -138,7 +137,6 @@ export default function (api: unknown, options: ApplicationPresetOptions) {
     plugins.push(
       createEs2015LinkerPlugin({
         linkerJitMode: options.angularLinker.jitMode,
-        sourceMapping: options.angularLinker.sourcemap,
         logger: createNgtscLogger(options.diagnosticReporter),
         fileSystem: {
           resolve: path.resolve,
--- a/packages/angular_devkit/build_angular/src/babel/presets/application.ts
+++ b/packages/angular_devkit/build_angular/src/babel/presets/application.ts
@@ -137,6 +137,8 @@ export default function (api: unknown, options: ApplicationPresetOptions) {
     plugins.push(
       createEs2015LinkerPlugin({
         linkerJitMode: options.angularLinker.jitMode,
+        // This is a workaround until https://github.com/angular/angular/issues/42769 is fixed.
+        sourceMapping: false,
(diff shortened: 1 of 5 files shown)


alan-agius4:
Related issue in FW https://github.com/angular/angular/issues/42769

petebacondarwin (approved):

petebacondarwin:
Is it possible to create a test that demonstrated that the source-map for the library is created after this revert?

clydin (approved):

PR #21286, opened 2021-07-05
fix(@angular-devkit/build-angular): exclude `outputPath` from persistent build cache key

### LINKED ISSUE #21275: `NG_PERSISTENT_BUILD_CACHE` breaks when localizing
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

No, `NG_PERSISTENT_BUILD_CACHE` is new

### Description

When running `NG_PERSISTENT_BUILD_CACHE` if using `--localize` the cache path generates new cache keys for every run. I was trying to test out the new option as mentioned in https://github.com/angular/angular-cli/issues/20792 but I wasn't seeing any difference in performance. I had noticed a few cache keys getting generated and while I didn't notice it corresponded to the number of runs I was testing (partially due to https://github.com/angular/angular-cli/pull/21274 creating nested directories _in_ the cache directory due to base64 encoding `/` not as the prefix). Wonder what the cache key being hashed was I edited the file and logged each of the following:

https://github.com/angular/angular-cli/blob/736a5f89deaca85f487b78aec9ff66d4118ceb6a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts#L528-L535

All of the variables were the same *expect* the `wco.buildOptions`. What I noticed was the `outputPath` was the following:
* `"outputPath":/tmp/angular-cli-i18n-9lmmzl"`
* `"outputPath":"/tmp/angular-cli-i18n-LR9px9"`

I should have noticed directly from the name that `i18n` was enabled and was likely related, but looking through code for `outputPath` I noticed that the following file had a match:

https://github.com/angular/angular-cli/blob/736a5f89deaca85f487b78aec9ff66d4118ceb6a/packages/angular_devkit/build_angular/src/utils/i18n-options.ts#L264-L265

Still not paying attention to the generated output path :sweat_smile: but realizing it might be due to the localization I removed the `--localize` flag from my command and noticed that it returned to the value I expected.

However I now receive errors that I was trying to write a file to the root directory. Thinking this was because `NG_BUILD_CACHE` was not specified I did so, but then realized there was a `/` in the cache name prefix and regardless of the cache directory specified (automatic or by me) the cache was trying to be created in the root directory.

I proposed https://github.com/angular/angular-cli/pull/21274 to fix the issue not related to `--localize`, but I'm not sure if a constant output path should be swapped back in before JSON serialization or if there should be some other fix. I can include that in https://github.com/angular/angular-cli/pull/21274 to fix both if desired.

## 🔬 Minimal Reproduction

Use any app which already has localization and try to run with and without `NG_PERSISTENT_BUILD_CACHE` and observe there is no difference after the initial run (the initial speedup is due to `NG_BUILD_CACHE` not being disabled)

## 🔥 Exception or Error

There is no exception, the caching just doesn't work.

## 🌍 Your Environment

<pre><code>
> NexleafDP@1.0.0 ng /code
> ng "version"

                |___/
    

Angular CLI: 12.1.0
Node: 14.17.1
Package Manager: npm 6.14.13
OS: linux x64

Angular: 12.1.0
... forms, google-maps, language-service, localize, material

@angular-devkit/architect       0.1201.0
@angular-devkit/build-angular   12.1.0
@angular-devkit/core            12.1.0
@angular-devkit/schematics      12.1.0
@schematics/angular             12.1.0

</code></pre>

alan-agius4:
With this change we exclude `outputPath` from cache key due to i18n extraction which causes it to change on every build https://github.com/angular/angular-cli/blob/736a5f89deaca85f487b78aec9ff66d4118ceb6a/packages/angular_devkit/build_angular/src/utils/i18n-options.ts#L264-L265

Closes #21275

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -507,7 +507,14 @@ function getCacheSettings(
         .update(packageVersion)
         .update(wco.projectRoot)
         .update(JSON.stringify(wco.tsConfig))
-        .update(JSON.stringify(wco.buildOptions))
+        .update(
+          JSON.stringify({
+            ...wco.buildOptions,
+            // Needed because outputPath changes on every build when using i18n extraction
+            // https://github.com/angular/angular-cli/blob/736a5f89deaca85f487b78aec9ff66d4118ceb6a/packages/angular_devkit/build_angular/src/utils/i18n-options.ts#L264-L265
+            outputPath: undefined,
+          }),
+        )
         .update(supportedBrowsers.join(''))
         .digest('hex'),
     };


terencehonles:
LGTM, thanks @alan-agius4! Definitely looking forward to the speed boost I saw due to caching :slightly_smiling_face:. I'm assuming anything cached wouldn't be stored in the `outputPath` so it's safe to remove before and then after the run like you did.

terencehonles:
@clydin friendly ping about the review (I assume it needs to come from you not me). I'd definitely like to see this in the next patch version if possible :slightly_smiling_face:

clydin (approved):

clydin (approved):

PR #21297, opened 2021-07-06
fix(@angular/cli): handle NPM_CONFIG environment variables during ng update and ng add

mikejancar:
Some organizations are moving away from storing tokens/secrets in an npm config file in favor of environment variables that only exist for the span of a terminal session.  This commit will read those variables in when no options have been configured via a file.

From a testing perspective, I was not able to get the full unit test/e2e suites to run successfully.  The unit test suite fails immediately with a Bazel build failure. If you have any insight into what might be causing this, I'm more than happy to keep trying things.  I wasn't very successful at finding similar issues in the normal places (i.e. Stack Overflow).  

image

The full e2e suite failed at different steps in each of my attempts.  The latest one is below as an example.  If there's some common gotchas with the e2e tests, I'm willing to try again.  

image

I was able to successfully run the following tests which looked to be the most relevant to the logic I was updating.  

- tests\legacy-cli\e2e\tests\update\\*.ts
- tests\legacy-cli\e2e\tests\commands\add\\*.ts

### DIFF
--- a/packages/angular/cli/utilities/package-metadata.ts
+++ b/packages/angular/cli/utilities/package-metadata.ts
@@ -131,7 +131,7 @@ function readOptions(
     logger.info(`Locating potential ${baseFilename} files:`);
   }
 
-  const options: PackageManagerOptions = {};
+  let options: PackageManagerOptions = {};
   for (const location of [...defaultConfigLocations, ...projectConfigLocations]) {
     if (existsSync(location)) {
       if (showPotentials) {
@@ -142,58 +142,84 @@ function readOptions(
       // Normalize RC options that are needed by 'npm-registry-fetch'.
       // See: https://github.com/npm/npm-registry-fetch/blob/ebddbe78a5f67118c1f7af2e02c8a22bcaf9e850/index.js#L99-L126
       const rcConfig: PackageManagerOptions = yarn ? lockfile.parse(data) : ini.parse(data);
-      for (const [key, value] of Object.entries(rcConfig)) {
-        let substitutedValue = value;
 
-        // Substitute any environment variable references.
-        if (typeof value === 'string') {
-          substitutedValue = value.replace(/\$\{([^\}]+)\}/, (_, name) => process.env[name] || '');
-        }
+      options = normalizeOptions(rcConfig, location);
+    }
+  }
 
-        switch (key) {
-          // Unless auth options are scope with the registry url it appears that npm-registry-fetch ignores them,
-          // even though they are documented.
-          // https://github.com/npm/npm-registry-fetch/blob/8954f61d8d703e5eb7f3d93c9b40488f8b1b62ac/README.md
-          // https://github.com/npm/npm-registry-fetch/blob/8954f61d8d703e5eb7f3d93c9b40488f8b1b62ac/auth.js#L45-L91
-          case '_authToken':
-          case 'token':
-          case 'username':
-          case 'password':
-          case '_auth':
-          case 'auth':
-            options['forceAuth'] ??= {};
-            options['forceAuth'][key] = substitutedValue;
-            break;
-          case 'noproxy':
-          case 'no-proxy':
-            options['noProxy'] = substitutedValue;
-            break;
-          case 'maxsockets':
-            options['maxSockets'] = substitutedValue;
-            break;
-          case 'https-proxy':
-          case 'proxy':
-            options['proxy'] = substitutedValue;
-            break;
-          case 'strict-ssl':
-            options['strictSSL'] = substitutedValue;
-            break;
-          case 'local-address':
-            options['localAddress'] = substitutedValue;
-            break;
-          case 'cafile':
-            if (typeof substitutedValue === 'string') {
-              const cafile = path.resolve(path.dirname(location), substitutedValue);
-              try {
-                options['ca'] = readFileSync(cafile, 'utf8').replace(/\r?\n/g, '\n');
-              } catch {}
-            }
-            break;
-          default:
-            options[key] = substitutedValue;
-            break;
+  for (const [key, value] of Object.entries(process.env)) {
+    if (!value || !key.toLowerCase().startsWith('npm_config_')) {
+      continue;
+    }
+
+    const normalizedName = key
+      .substr(11)
+      .replace(/(?!^)_/g, '-') // don't replace _ at the start of the key
+      .toLowerCase();
+    options[normalizedName] = value;
+  }
+
+  options = normalizeOptions(options);
+
+  return options;
+}
+
+function normalizeOptions(
+  rawOptions: PackageManagerOptions,
+  location = process.cwd(),
+): PackageManagerOptions {
+  const options: PackageManagerOptions = {};
+
+  for (const [key, value] of Object.entries(rawOptions)) {
+    let substitutedValue = value;
+
+    // Substitute any environment variable references.
+    if (typeof value === 'string') {
+      substitutedValue = value.replace(/\$\{([^\}]+)\}/, (_, name) => process.env[name] || '');
+    }
+
+    switch (key) {
+      // Unless auth options are scope with the registry url it appears that npm-registry-fetch ignores them,
+      // even though they are documented.
+      // https://github.com/npm/npm-registry-fetch/blob/8954f61d8d703e5eb7f3d93c9b40488f8b1b62ac/README.md
+      // https://github.com/npm/npm-registry-fetch/blob/8954f61d8d703e5eb7f3d93c9b40488f8b1b62ac/auth.js#L45-L91
+      case '_authToken':
+      case 'token':
+      case 'username':
+      case 'password':
+      case '_auth':
+      case 'auth':
+        options['forceAuth'] ??= {};
+        options['forceAuth'][key] = substitutedValue;
+        break;
+      case 'noproxy':
+      case 'no-proxy':
+        options['noProxy'] = substitutedValue;
+        break;
+      case 'maxsockets':
+        options['maxSockets'] = substitutedValue;
+        break;
+      case 'https-proxy':
+      case 'proxy':
+        options['proxy'] = substitutedValue;
+        break;
+      case 'strict-ssl':
+        options['strictSSL'] = substitutedValue;
+        break;
+      case 'local-address':
+        options['localAddress'] = substitutedValue;
+        break;
+      case 'cafile':
+        if (typeof substitutedValue === 'string') {
+          const cafile = path.resolve(path.dirname(location), substitutedValue);
+          try {
+            options['ca'] = readFileSync(cafile, 'utf8').replace(/\r?\n/g, '\n');
+          } catch {}
         }
-      }
+        break;
+      default:
+        options[key] = substitutedValue;
+        break;
     }
   }
 
--- a/tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
+++ b/tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
@@ -1,0 +1,33 @@
+import { expectFileNotToExist, expectFileToExist } from '../../../utils/fs';
+import { getActivePackageManager } from '../../../utils/packages';
+import { git, ng } from '../../../utils/process';
+import {
+  createNpmConfigForAuthentication,
+  setNpmEnvVarsForAuthentication,
+} from '../../../utils/registry';
+
+export default async function () {
+  const packageManager = getActivePackageManager();
+
+  if (packageManager === 'npm') {
+    const originalEnvironment = { ...process.env };
+    try {
+      const command = ['add', '@angular/pwa', '--skip-confirmation'];
+
+      // Environment variables only
+      await expectFileNotToExist('src/manifest.webmanifest');
+      setNpmEnvVarsForAuthentication();
+      await ng(...command);
+      await expectFileToExist('src/manifest.webmanifest');
+      await git('clean', '-dxf');
+
+      // Mix of config file and env vars works
+      await expectFileNotToExist('src/manifest.webmanifest');
+      await createNpmConfigForAuthentication(false, true);
+      await ng(...command);
+      await expectFileToExist('src/manifest.webmanifest');
+    } finally {
+      process.env = originalEnvironment;
+    }
+  }
+}
--- a/tests/legacy-cli/e2e/tests/commands/add/secure-registry.ts
+++ b/tests/legacy-cli/e2e/tests/commands/add/secure-registry.ts
@@ -5,8 +5,11 @@ import { expectToFail } from '../../../utils/utils';
 
 export default async function () {
   // The environment variable has priority over the .npmrc
-  const originalRegistryVariable = process.env['NPM_CONFIG_REGISTRY'];
-  delete process.env['NPM_CONFIG_REGISTRY'];
+  let originalRegistryVariable;
+  if (process.env['NPM_CONFIG_REGISTRY']) {
+    originalRegistryVariable = process.env['NPM_CONFIG_REGISTRY'];
+    delete process.env['NPM_CONFIG_REGISTRY'];
+  }
 
   try {
     const command = ['add', '@angular/pwa', '--skip-confirmation'];
@@ -32,6 +35,8 @@ export default async function () {
     await createNpmConfigForAuthentication(true, true);
     await expectToFail(() => ng(...command));
   } finally {
-    process.env['NPM_CONFIG_REGISTRY'] = originalRegistryVariable;
+    if (originalRegistryVariable) {
+      process.env['NPM_CONFIG_REGISTRY'] = originalRegistryVariable;
+    }
   }
 }
--- a/tests/legacy-cli/e2e/utils/registry.ts
+++ b/tests/legacy-cli/e2e/utils/registry.ts
@@ -19,6 +19,10 @@ export function createNpmRegistry(withAuthentication = false): ChildProcess {
   });
 }
 
+// Token was generated using `echo -n 'testing:s3cret' | openssl base64`.
+const VALID_TOKEN = `dGVzdGluZzpzM2NyZXQ=`;
+const SECURE_REGISTRY = `//localhost:4876/`;
+
 export function createNpmConfigForAuthentication(
   /**
    * When true, the authentication token will be scoped to the registry URL.
@@ -37,9 +41,8 @@ export function createNpmConfigForAuthentication(
   /** When true, an incorrect token is used. Use this to validate authentication failures. */
   invalidToken = false,
 ): Promise<void> {
-  // Token was generated using `echo -n 'testing:s3cret' | openssl base64`.
-  const token = invalidToken ? `invalid=` : `dGVzdGluZzpzM2NyZXQ=`;
-  const registry = `//localhost:4876/`;
+  const token = invalidToken ? `invalid=` : VALID_TOKEN;
+  const registry = SECURE_REGISTRY;
 
   return writeFile(
     '.npmrc',
@@ -53,4 +56,15 @@ export function createNpmConfigForAuthentication(
         registry=http:${registry}
       `,
   );
+}
+
+export function setNpmEnvVarsForAuthentication(
+  /** When true, an incorrect token is used. Use this to validate authentication failures. */
+  invalidToken = false,
+): void {
+  const token = invalidToken ? `invalid=` : VALID_TOKEN;
+  const registry = SECURE_REGISTRY;
+
+  process.env['NPM_CONFIG_REGISTRY'] = `http:${registry}`;
+  process.env['NPM_CONFIG__AUTH'] = token;
 }
--- a/tests/legacy-cli/e2e/tests/commands/add/registry-option.ts
+++ b/tests/legacy-cli/e2e/tests/commands/add/registry-option.ts
@@ -11,8 +11,11 @@ export default async function () {
     '.npmrc': 'registry=http://127.0.0.1:9999',
   });
   // The environment variable has priority over the .npmrc
-  const originalRegistryVariable = process.env['NPM_CONFIG_REGISTRY'];
-  delete process.env['NPM_CONFIG_REGISTRY'];
+  let originalRegistryVariable;
+  if (process.env['NPM_CONFIG_REGISTRY']) {
+    originalRegistryVariable = process.env['NPM_CONFIG_REGISTRY'];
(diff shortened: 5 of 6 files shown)


mikejancar:
@googlebot I signed it!

alan-agius4 (changes requested):
Thanks for your contribution. 

I did take a quick look at this and it appears that some wrong assumptions were made.

- You can use NPMRC togather environment variables. The latter will override the former.

```ini
engine-strict = true
registry = http://foo.com
```

```
NPM_CONFIG_REGISTRY=http://bar.com npm i
```

Should result in the following
```js
{
   registry: 'http://bar.com',
   'engine-strict': true,
}
```

- Environment name variables need to be normalized `NPM_CONFIG_HTTPS_PROXY` -> `https-proxy`. The rules are trim the first 11 characters, lowercase and replace `_` with `-`.

Once the above are addressed, can you also add an E2E test, something similar to https://github.com/angular/angular-cli/blob/master/tests/legacy-cli/e2e/tests/commands/add/secure-registry.ts but instead of using an NPMRC to set the authentication,  use an environment variable?. When you are at this stage, I can help you with this.

mikejancar:
Thanks for reviewing so quickly, Alan.  I'll get started on the changes and let you know if I have any issues.

mikejancar:
Just to clarify, @alan-agius4, the use case that I'm trying to address is when there is no configuration file present.  When that occurs, we have found that the environment variables are never looked at because the substitution is only done in the existing code if a file is found.  This leads to `ng update` failing since it isn't using our private registry with the proper authentication.

The code I added will address this scenario.  Are you saying that the code shouldn't be needed at all or are you asking for a change to the logic I've added?  

I'll work on adding an E2E test in the meantime to illustrate this use case.

alan-agius4:
I am asking for changes in logic to make this code behave as NPM would. Because certain assumptions made during this change are not correct.

1) It is assumed that when using enviorment variables no config file is used. This is not correct as one can use both at the same time. As described here https://github.com/angular/angular-cli/pull/21297#pullrequestreview-701664518
2) Environment variables names needs to be normalised. Ex: `NPM_CONFIG_HTTPS_PROXY` -> `https-proxy`, See the full constraint here https://github.com/angular/angular-cli/pull/21297#pullrequestreview-701664518

mikejancar:
Thanks for the clarification.  Changes coming.

packages/angular/cli/utilities/package-metadata.ts
```diff
@@ -142,61 +142,102 @@ function readOptions(
...
-          default:
-            options[key] = substitutedValue;
-            break;
-        }
+  // eslint-disable-next-line no-console
+  console.log(`env vars: ${JSON.stringify(process.env)}`); // TODO: Remove
+  // eslint-disable-next-line no-console
+  console.log(`options post-config: ${JSON.stringify(options)}`); // TODO: Remove
+
+  for (const envVar in process.env) {
```
alan-agius4, line 155: Drive by comment, the normalization here is not totally valid for all variables.

It should be generic to handle all possible key names
```ts
for (const [key, value] in Object.entries(process.env)) {
  if (value === undefined || !key.toLowerCase().startsWith('npm_config_')) {
    continue;
  }

  const normalizedName = key
    .substr(11)
    .replace(/(?!^)_/g, '-') // don't replace _ at the start of the key
    .toLowerCase();
  options[normalizedName] = process.env[value];
}
```

mikejancar: Thank you.  Regex is not my strong suit.

mikejancar: @alan-agius4 I was able to finally clear up the E2E failures based on my changes.  I'm going to try and put my own E2E test back in and clean up the debug logging I added.  Hopefully, it should be ready for your review tomorrow some time.

mikejancar:
@alan-agius4 I believe I have everything up and running now.  Took me a while to pin down some of the inadvertent changes to the environment variables in the tests but I think I found them all.  Please review and let me know what you think.

packages/angular/cli/utilities/package-metadata.ts
```diff
@@ -142,58 +142,85 @@ function readOptions(
...
-            options[key] = substitutedValue;
-            break;
+  options = normalizeOptions(options);
+
+  return options;
+}
+
+function normalizeOptions(
+  rawOptions: PackageManagerOptions,
+  configLocation?: string,
```
alan-agius4, line 169:
```suggestion
  location = process.cwd(),
```

packages/angular/cli/utilities/package-metadata.ts
```diff
@@ -142,58 +142,85 @@ function readOptions(
...
+      case 'strict-ssl':
+        options['strictSSL'] = substitutedValue;
+        break;
+      case 'local-address':
+        options['localAddress'] = substitutedValue;
+        break;
+      case 'cafile':
+        if (typeof substitutedValue === 'string') {
+          const location = configLocation || '';
+          const cafile = path.resolve(path.dirname(location), substitutedValue);
```
alan-agius4, line 215:
```suggestion
          const cafile = path.resolve(path.dirname(location), substitutedValue);
```

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
```diff
@@ -0,0 +1,35 @@
...
+import {
+  createNpmConfigForAuthentication,
+  resetNpmEnvVars,
+  setNpmEnvVarsForAuthentication,
+} from '../../../utils/registry';
+
+export default async function () {
+  const packageManager = getActivePackageManager();
+
+  if (packageManager === 'npm') {
```
alan-agius4, line 13: We should test these with Yarn as well. Yarn does respect NPM_CONFIG.

See: https://classic.yarnpkg.com/en/docs/envvars/

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
```diff
@@ -0,0 +1,35 @@
...
+    try {
+      const command = ['add', '@angular/pwa', '--skip-confirmation'];
+
+      // Environment variables only
+      await expectFileNotToExist('src/manifest.webmanifest');
+      await writeFile('.npmrc', '');
```
alan-agius4, line 19: Doesn't seem to be needed.

```suggestion
```

tests/legacy-cli/e2e/utils/registry.ts
```diff
@@ -54,3 +57,19 @@ export function createNpmConfigForAuthentication(
...
+): void {
+  const token = invalidToken ? `invalid=` : validToken;
+  const registry = `http://localhost:4876/`;
+
+  process.env['NPM_CONFIG_REGISTRY'] = registry;
+  process.env['NPM_CONFIG__AUTH'] = token;
+}
+
+export function resetNpmEnvVars() {
+  process.env['NPM_CONFIG_REGISTRY'] = getGlobalVariable('package-registry');
```
alan-agius4, line 73:
```suggestion
await  setRegistry(true);
```

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
```diff
@@ -0,0 +1,35 @@
...
+      await expectFileToExist('src/manifest.webmanifest');
+      await git('clean', '-dxf');
+
+      // Mix of config file and env vars works
+      await expectFileNotToExist('src/manifest.webmanifest');
+      await createNpmConfigForAuthentication(false, true);
+      await ng(...command);
+      await expectFileToExist('src/manifest.webmanifest');
+    } finally {
+      await setRegistry(true);
```
alan-agius4, line 31:
```suggestion
```

tests/legacy-cli/e2e/tests/commands/add/secure-registry.ts
```diff
@@ -4,9 +4,12 @@ import { createNpmConfigForAuthentication } from '../../../utils/registry';
 import { expectToFail } from '../../../utils/utils';
 
 export default async function () {
-  // The environment variable has priority over the .npmrc
-  const originalRegistryVariable = process.env['NPM_CONFIG_REGISTRY'];
-  delete process.env['NPM_CONFIG_REGISTRY'];
+  let originalRegistryVariable;
```
alan-agius4, line 7: Why are the following changes needed?

alan-agius4 (changes requested):
Thanks for this, a couple of NITs.

packages/angular/cli/utilities/package-metadata.ts
mikejancar: Good call.  Updating now.

packages/angular/cli/utilities/package-metadata.ts
mikejancar: Updating as well.

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
mikejancar: I had added this because I was getting Yarn-specific test failures along the way.  I'll try and remove this to see if my test is in a better place now.

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
mikejancar: I wanted to try and simulate the specific use case where an *.npmrc* did not exist.  Now that I understand how these tests work a little more, I can see why this step probably isn't needed.  I'll take it out.

tests/legacy-cli/e2e/tests/commands/add/secure-registry.ts
mikejancar: When debugging the tests that were failing after my changes, I found that at least two of the tests following this pattern  - *secure-registry* and *registry-option* - ended up setting `NPM_CONFIG_REGISTRY` to a string value of "undefined".  This caused subsequent tests to fail now that the environment variables are being read in.  I figured I should update all three occurrences of this pattern to be thorough.  Only after making these changes was I able to get the entire test suite to pass.

tests/legacy-cli/e2e/utils/registry.ts
mikejancar: This was the first approach that I tried, but I don't believe the `NPM_CONFIG_REGISTRY` env var gets set during a CI run because of the way `setRegistry` is coded.  It worked when I ran the tests locally, but the tests continued to fail when I pushed my changes to the server.  I didn't want to update the logic in `setRegistry` because I figured there was a good reason it was coded that way so I isolated the change here.

```javascript
  if (isCI) {
    // Safe to set a user configuration on CI
    await npm('config', 'set', 'registry', url);
  } else {
    // Yarn does not use the environment variable so an .npmrc file is also required
    await writeFile('.npmrc', `registry=${url}`);
    process.env['NPM_CONFIG_REGISTRY'] = url;
  }
```

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
mikejancar: This is one of the changes I found was needed because subsequent tests fail due to their assumption that the *.npmrc* file is pointing to the non-authenticated verdaccio server (:4873) but it's still set to :4876 when it leaves this test.  I added this step to set the environment back to its original state prior to my test.

tests/legacy-cli/e2e/utils/registry.ts
alan-agius4: In that case, we should probably restore the original `process.env.NPM_CONFIG_REGISTRY` that was set prior of using `setNpmEnvVarsForAuthentication`.

alan-agius4: I think it might be easier to just restore the original process.env, in the test file.

tests/legacy-cli/e2e/utils/registry.ts
```diff
@@ -54,3 +57,19 @@ export function createNpmConfigForAuthentication(
       `,
   );
 }
+
+export function setNpmEnvVarsForAuthentication(
+  /** When true, an incorrect token is used. Use this to validate authentication failures. */
+  invalidToken = false,
+): void {
+  const token = invalidToken ? `invalid=` : validToken;
+  const registry = `http://localhost:4876/`;
```
alan-agius4, line 66: Can you extract into a shared const, since it's being used in multiple places
Something along `SECURE_REGISTRY`.

tests/legacy-cli/e2e/utils/registry.ts
```diff
@@ -19,6 +20,9 @@ export function createNpmRegistry(withAuthentication = false): ChildProcess {
   });
 }
 
+// Token was generated using `echo -n 'testing:s3cret' | openssl base64`.
+const validToken = `dGVzdGluZzpzM2NyZXQ=`;
```
alan-agius4, line 24: NIT.

```suggestion
const VALID_TOKEN = `dGVzdGluZzpzM2NyZXQ=`;
```

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
mikejancar: The Yarn -specific test failed again.  I'll dig in to see if I can get more detail as to why.

tests/legacy-cli/e2e/utils/registry.ts
mikejancar: Sounds good.

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
alan-agius4: It appears that Yarn has a special variable for the registry `YARN_REGISTRY`. Let me at this a bit.

tests/legacy-cli/e2e/utils/registry.ts
mikejancar: Definitely.

tests/legacy-cli/e2e/utils/registry.ts
mikejancar: Will do.

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
alan-agius4: This will require some additional changes, including infra changes.

Let me follow up in a separate PR

mikejancar: Will do.

tests/legacy-cli/e2e/tests/commands/add/secure-registry.ts
mikejancar: @alan-agius4 if you look at my latest test failure, I dumped the environment variables after all the setup scripts have completed (search for `env-vars-post-500`) and right before the *secure-registry* script tries to save the `NPM_CONFIG_REGISTRY` value.  In both cases the value doesn't exist hence the `undefined` string value is set when being put back when the test completes.  

Do you know why the registry value isn't set in the `setRegistry` function when `ci` is set to true?  That might be the root of the issue here from what I'm seeing.

alan-agius4: On CI we can use the `npm config` directly. Setting `NPM_CONFIG_REGISTRY` to `undefined` shouldn't be a problem at all. Let me try to get to the bottom of this.

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
alan-agius4: Files get cleaned prior every test run. https://github.com/angular/angular-cli/blob/4a08ec07913ecb3f8b981d6439f1bbfba86221f4/tests/legacy-cli/e2e_runner.ts#L182

tests/legacy-cli/e2e/tests/commands/add/secure-registry.ts
alan-agius4: Here's the stack 

```
[error] TypeError [ERR_INVALID_URL]: Invalid URL: undefined/undefined/@angular%2fcli
    at onParseError (internal/url.js:259:9)
    at new URL (internal/url.js:335:5)
    at regFetch (/mnt/ramdisk/e2e-main/subdirectory/subdirectory-test-project/node_modules/npm-registry-fetch/index.js:48:5)
    at RegistryFetcher.packument (/mnt/ramdisk/e2e-main/subdirectory/subdirectory-test-project/node_modules/pacote/lib/registry.js:86:15)
    at RegistryFetcher.manifest (/mnt/ramdisk/e2e-main/subdirectory/subdirectory-test-project/node_modules/pacote/lib/registry.js:117:17)
    at Object.manifest (/mnt/ramdisk/e2e-main/subdirectory/subdirectory-test-project/node_modules/pacote/lib/index.js:16:45)
    at Object.fetchPackageManifest (/mnt/ramdisk/e2e-main/subdirectory/subdirectory-test-project/node_modules/@angular/cli/utilities/package-metadata.js:215:35)
    at UpdateCommand.checkCLILatestVersion (/mnt/ramdisk/e2e-main/subdirectory/subdirectory-test-project/node_modules/@angular/cli/commands/update-impl.js:658:60)
    at UpdateCommand.run (/mnt/ramdisk/e2e-main/subdirectory/subdirectory-test-project/node_modules/@angular/cli/commands/update-impl.js:218:49)
    at async UpdateCommand.validateAndRun (/mnt/ramdisk/e2e-main/subdirectory/subdirectory-test-project/node_modules/@angular/cli/models/command.js:136:22)

```

alan-agius4: Note: the tests which is failing `tests/misc/update-git-clean-subdirectory.ts` ran prior to `tests/update/update-secure-registry.ts`.

```diff
setup/010-local-publish.ts
setup/100-global-cli.ts
setup/200-create-tmp-dir.ts
setup/300-log-environment.ts
setup/500-create-project.ts
tests/basic/e2e.ts
tests/basic/serve.ts
tests/build/barrel-file.ts
tests/build/bundle-budgets.ts
tests/build/differential-cache.ts
tests/build/json.ts
tests/build/no-sourcemap.ts
tests/build/prod-build.ts
tests/build/rebuild-replacements.ts
tests/build/styles/imports.ts
tests/build/styles/preset-env.ts
tests/build/ts-paths.ts
tests/commands/add/add.ts
tests/commands/add/registry-option.ts
tests/commands/config/config-global.ts
tests/commands/help/help-json.ts
tests/commands/unknown-configuration.ts
tests/generate/component/component-duplicate.ts
tests/generate/component/component-module-export.ts
tests/generate/directive/directive-basic.ts
tests/generate/generate-error.ts
tests/generate/help-output-no-duplicates.ts
tests/generate/library/library-consumption-ve.ts
tests/generate/pipe/pipe-basic.ts
tests/generate/service/service-basic.ts
tests/i18n/ivy-localize-dl-json.ts
tests/i18n/ivy-localize-hashes.ts
tests/i18n/ivy-localize-sourcelocale.ts
tests/misc/coverage.ts
tests/misc/forwardref-es2015.ts
tests/misc/minimal-config.ts
tests/misc/npm-audit.ts
tests/misc/target-default-configuration.ts
- tests/misc/update-git-clean-subdirectory.ts
tests/schematics_cli/basic.ts
tests/test/test-scripts.ts
+ tests/update/update-secure-registry.ts
```

alan-agius4: Ill continue trying to get to the bottom of this later on today.

packages/angular/cli/utilities/package-metadata.ts
```diff
@@ -142,58 +142,84 @@ function readOptions(
...
-        // Substitute any environment variable references.
-        if (typeof value === 'string') {
-          substitutedValue = value.replace(/\$\{([^\}]+)\}/, (_, name) => process.env[name] || '');
-        }
+      options = normalizeOptions(rcConfig, location);
+    }
+  }
+
+  for (const [key, value] of Object.entries(process.env)) {
+    if (value === undefined || !key.toLowerCase().startsWith('npm_config_')) {
```
alan-agius4, line 151:
```suggestion
    if (!value || !key.toLowerCase().startsWith('npm_config_')) {
```

tests/legacy-cli/e2e/tests/commands/add/secure-registry.ts
alan-agius4: So what is happening is that Node.JS will serialize undefined to `'undefined'`. This means that the check in https://github.com/angular/angular-cli/pull/21297/files#diff-1fe35077f343d441cac1946e85d88adf8808fa6b3a0898e85209cc7143e6b9c0R151 should be changed since to reset a variable one would need to set an empty string.

mikejancar: Same result.  A string of `undefined` will still result in a truthy value so it will still set the option from the environment variable.  I can either update my code in *package-metadata* to look for a string value of `undefined` or do the changes I proposed earlier in those tests to make sure they don't set an undefined value to `undefined` when leaving the test.  In the second test suite failure it sets the env var to `undefined` when leaving *secure-registry* which then causes *update-git-clean* to fail.

image

Those are the only two options that I see, but I'm open to other suggestions.

alan-agius4: Let just update the spec files, since checking for `"undefined"` isn't correct.

mikejancar: Sounds good.

tests/legacy-cli/e2e/tests/commands/add/npm-env-vars.ts
mikejancar: Gotcha.  Removing...

mikejancar:
Looks like we're back to a passing test suite.  Let me know if there's anything else you'd like me to do.

alan-agius4 (approved):
Thanks for your contribution.

mikejancar:
Thanks for your patience and guidance, @alan-agius4.

alan-agius4:
@mikejancar, welcome and thanks again for your contribution. FYI: I did follow up to support Yarn enviorment vars https://github.com/angular/angular-cli/pull/21378

mikejancar:
I did see that.  Nice follow up.

PR #21317, opened 2021-07-09
    fix(@angular-devkit/build-angular): `extractLicenses` didn't have an effect when using server builder

alan-agius4:


packages/angular_devkit/build_angular/src/server/tests/options/source-map_spec.ts
```diff
@@ -0,0 +1,150 @@
...
+        },
+      });
+
+      const { result } = await harness.executeOnce();
+      expect(result?.success).toBe(true);
+      harness.expectFile('dist/main.js.map').toNotExist();
+      harness.expectFile('dist/main.js').content.not.toContain('sourceMappingURL=main.js.map');
+    });
+
+    it(`should not generate scripts sourceMaps when "scripts" option is true`, async () => {
```
clydin, line 136:
```suggestion
    it(`should generate scripts sourceMaps when "scripts" option is true`, async () => {
```

clydin (approved):

PR #21323, opened 2021-07-09
perf(@angular-devkit/build-angular): use `esbuild` as a CSS optimizer for component styles

clydin:
The stylesheet optimization pipeline now uses `esbuild` for component stylesheets. The usage of `esbuild` provides noticeable build time improvements as well as, on average, smaller output sizes.
`esbuild` currently does not support stylesheet sourcemaps. However, since the Angular CLI does not support component stylesheet sourcemaps when optimizing, this lack of support is not problematic.
Global stylesheets will continue to use `cssnano` as an optimizer due to sourcemaps being required for global stylesheets even when optimizing. When `esbuild` adds stylesheet sourcemap support, global stylesheets may be transitioned to `esbuild` as well.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -277,7 +277,9 @@ export function getStylesConfig(wco: WebpackConfigOptions): webpack.Configuratio
   const extraMinimizers = [];
   if (buildOptions.optimization.styles.minify) {
     const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
-    const minimizerOptions = {
+    const esbuild = require('esbuild') as typeof import('esbuild');
+
+    const cssnanoOptions = {
       preset: [
         'default',
         {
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -299,20 +301,39 @@ export function getStylesConfig(wco: WebpackConfigOptions): webpack.Configuratio
     );
 
     extraMinimizers.push(
+      // Component styles use esbuild which is faster and generates smaller files on average.
+      // esbuild does not yet support style sourcemaps but component style sourcemaps are not
```
alan-agius4, line 305: Can we add a comment to reference https://github.com/evanw/esbuild/issues/519 to traceability?

alan-agius4 (approved):
🚀🚀

demike:
> The stylesheet optimization pipeline now uses `esbuild` for component stylesheets. The usage of `esbuild` provides noticeable build time improvements as well as, on average, smaller output sizes.
> `esbuild` currently does not support stylesheet sourcemaps. However, since the Angular CLI does not support component stylesheet sourcemaps when optimizing, this lack of support is not problematic.
> Global stylesheets will continue to use `cssnano` as an optimizer due to sourcemaps being required for global stylesheets even when optimizing. When `esbuild` adds stylesheet sourcemap support, global stylesheets may be transitioned to `esbuild` as well.

fyi: esbuild now supports css source maps.

PR #21330, opened 2021-07-12
refactor(@schematics/angular): sync schematics dependencies using Renovate

alan-agius4:
With this change we sync workspace and library schematic dependencies using Renovate. We do this to avoid having to keep these in sync by hand.

We could have used TypeScripts' `resolveJsonModule` to make the `latestVersion` object typesafe. But `ts_library` doesn't support JSON inputs.


packages/schematics/angular/utility/latest-versions/package.json
```diff
@@ -0,0 +1,20 @@
+{
```
alan-agius4, line 1: NB: this is a seperate child directory due avoid ESLint errors from the `extraneous-dependencies` rule.

packages/schematics/angular/utility/latest-versions/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "description": "Package versions used by schematics in @schematics/angular.",
+  "comment": "This file is needed so that depedencies are sycned by Renovate.",
```
clydin, line 3:
```suggestion
  "comment": "This file is needed so that depedencies are synced by Renovate.",
```

packages/schematics/angular/application/index.ts
```diff
@@ -50,7 +50,7 @@ function addDependenciesToPackageJson(options: ApplicationOptions) {
       {
         type: NodeDependencyType.Dev,
         name: 'typescript',
-        version: latestVersions.TypeScript,
+        version: latestVersions.typescript,
```
clydin, line 53: What do you think about using indexed property access for all the package.json originating properties?

clydin (approved):

PR #21344, opened 2021-07-14
refactor(@schematics/angular): use a helper function for basic generation schematics

clydin:
An internal schematics helper rule has been introduced for Angular schematics that only generate a set of project files from a set of templated files. Multiple current generation schematics contained essentially the same code which increased the sustainment burden and made refactoring and improvements more complicated.


alan-agius4 (approved):
Nice

PR #21362, opened 2021-07-18
fix(@angular-devkit/build-angular): dasherize disable-host-check suggestion

LeonEck:
Camel case arguments have been deprecated and therefore shouldn't be suggested.


alan-agius4 (approved):
Thanks for your contribution.

PR #21372, opened 2021-07-20
build: transition from ts-api-guardian to dev-infra api-golden utility

clydin:
The dev-infra tooling now directly provides public API change testing capabilities that leverage the `api-extractor` utility. These new testing capabilities are shared with framework and components.
`ts-api-guardian` has been removed as a dependency as a result.
The new public API tooling searches for nested package.json files to determine the location of secondary entrypoints. All secondary entrypoints for the CLI related packages now contain a secondary entrypoint package.json file.
The internal monorepo package discovery script was also updated to support the presence of the nested package.json files.


goldens/public-api/manage.js
```diff
@@ -0,0 +1,54 @@
...
+const BAZEL_PUBLIC_API_TARGET_QUERY_CMD =
+    `yarn -s bazel query --output label 'kind(nodejs_test, ...) intersect attr("tags", "api_guard", ...)'`
+// Bazel targets for testing Public API goldens
+process.stdout.write('Gathering all Public API targets');
+const ALL_PUBLIC_API_TESTS = exec(BAZEL_PUBLIC_API_TARGET_QUERY_CMD, {silent: true})
+                                 .trim()
+                                 .split('\n')
+                                 .map(test => test.trim());
+process.stdout.clearLine();
+process.stdout.cursorTo(0);
```
devversion, line 18: I had issues in the past with this script in the FW repo (on windows), which is why I always used a custom bash script with `xargs`. The bazel query above never returns anything:

```
Failed run: .accept
  Loading:
  Loading: 0 packages loaded
  ERROR: Skipping '.accept': no such target '//:.accept': target '.accept' not declared in package '' defined by C:/users/paul/projects/angular/BUILD.bazel
  WARNING: Target pattern parsing failed.
  ERROR: no such target '//:.accept': target '.accept' not declared in package '' defined by C:/users/paul/projects/angular/BUILD.bazel
  INFO: Elapsed time: 0.051s
```

The `ALL_PUBLIC_API_TESTS` array just was an array with an empty string as single item FWIW.

lib/packages.ts
```diff
@@ -100,23 +100,32 @@ function loadPackageJson(p: string) {
   return pkg;
 }
 
-function _findAllPackageJson(dir: string, exclude: RegExp): string[] {
-  const result: string[] = [];
-  fs.readdirSync(dir).forEach((fileName) => {
-    const p = path.join(dir, fileName);
+function* _findAllPackageJson(dir: string, exclude: RegExp): Iterable<string> {
```
devversion, line 103: Optional nit: Do you want to rename this to something more specific now? e.g. `_findPrimaryPackageJsonFiles()`

devversion (approved):
LGTM overall. Thanks for integrating the new tool in the CLI repo! 🎉

goldens/public-api/manage.js
clydin: I actually wanted to investigate adding a more generalized version of this into the dev-infra tools directly since variations of this are done in several places.

PR #21375, opened 2021-07-20
build: update ng-dev configuration for release

josephperrott:
Update the configuration for releasing via ng-dev.

### DIFF
--- a/.ng-dev/release.ts
+++ b/.ng-dev/release.ts
@@ -9,8 +9,15 @@ const npmPackages = Object.keys(releasePackages);
 /** Configuration for the `ng-dev release` command. */
 export const release: ReleaseConfig = {
   npmPackages,
-  buildPackages,
-  releaseNotes: {},
+  buildPackages: () => buildPackages(),
+  releaseNotes: {
+    groupOrder: [
+      '@angular/cli',
+      '@schematics/angular',
+      '@angular-devkit/architect-cli',
+      '@angular-devkit/schematics-cli',
+    ],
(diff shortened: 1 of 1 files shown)


alan-agius4:
@josephperrott, the package is updated on a daily bases in the master branch through Renovate.

I suggest that if you need this particular version to be available also in the patch branch you only create a PR, targeting solely patch branch.

josephperrott:
@alan-agius4 We should be keeping the `@angular/dev-infra-private` package up to date in patch as well.

alan-agius4:
@josephperrott, I can configure renovate to only update the dev-Infra package in the patch branch. The caveat of this is that with every minor/patch release we’d need to update the renovate config and change the patch branch name.

.ng-dev/release.ts
```diff
@@ -9,8 +9,15 @@ const npmPackages = Object.keys(releasePackages);
 /** Configuration for the `ng-dev release` command. */
 export const release: ReleaseConfig = {
   npmPackages,
-  buildPackages,
-  releaseNotes: {},
+  buildPackages: () => buildPackages(),
+  releaseNotes: {
+    groupOrder: [
```
alan-agius4, line 14: Shouldn’t we add all the package names here?

josephperrott: Since these are the only ones ordered, they are the only ones that need to be included.  The rest are just sorted alphabetically.

alan-agius4 (approved):

PR #21376, opened 2021-07-20
build: update renovate to update @angular/dev-infra-private as target: patch

josephperrott:
Updating the @angular/dev-infra-private packages should be done on all active
development branches.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -21,8 +21,14 @@
   "packageRules": [
     {
       "packagePatterns": ["^@angular/.*"],
+      "excludePackageNames": ["@angular/dev-infra-private"],
       "groupName": "angular",
       "pinVersions": false
+    },
+    {
+      "matchPackageNames": ["@angular/dev-infra-private"],
+      "groupName": "dev-infra",
+      "labels": ["action: merge", "target: patch"]
     },
     {
       "packagePatterns": ["^@babel/.*"],


renovate.json
```diff
@@ -21,9 +21,15 @@
...
     {
       "packagePatterns": ["^@angular/.*"],
+      "excludePackagePatterns": ["^@angular/dev-infra-private"],
       "groupName": "angular",
       "pinVersions": false
     },
+    {
+      "packagePatterns": ["^@angular/dev-infra-private"],
+      "groupName": "dev-infra",
+      "labels": ["action: merge", "target: patch"]
```
alan-agius4, line 31: I think this is not enough since likely the commit will not apply cleanly due to possible conflicts in package.json.

alan-agius4: That said, maybe we should try this approach first, and see how often we do have conflict before going with the other approach.

Thoughts?

renovate.json
```diff
@@ -21,9 +21,15 @@
   "packageRules": [
     {
       "packagePatterns": ["^@angular/.*"],
+      "excludePackagePatterns": ["^@angular/dev-infra-private"],
       "groupName": "angular",
       "pinVersions": false
     },
+    {
+      "packagePatterns": ["^@angular/dev-infra-private"],
```
alan-agius4, line 29:
```suggestion
      "matchPackageNames": ["@angular/dev-infra-private"],
```

renovate.json
```diff
@@ -21,9 +21,15 @@
   "packageRules": [
     {
       "packagePatterns": ["^@angular/.*"],
+      "excludePackagePatterns": ["^@angular/dev-infra-private"],
```
alan-agius4, line 24:
```suggestion
      "excludePackageNames": ["@angular/dev-infra-private"],
```

renovate.json
josephperrott: Lets try this first as I imagine that we don't actually run into conflicts very often because the dependencies don't often change for dev-infra.

alan-agius4: SGTM.

alan-agius4 (approved):

alan-agius4:
@josephperrott, can you create a patch PR that sync up dev-infra package? As otherwise Renovate PR will conflict.

renovate.json
alan-agius4: It appears that as I suspected this approach will not work as the commit cannot be applied cleanly 

Example of another PR https://github.com/angular/angular-cli/pull/21409

alan-agius4: Let me know if we should explore the `baseBranch` option.

PR #21390, opened 2021-07-21
fix(@angular-devkit/build-angular): allow classes with pure annotated static properties to be optimized

clydin:
When script optimizations are enabled, classes containing downlevelled static properties are wrapped in a pure annotated IIFE to allow the class to be removed if it is otherwise unused. Only properties with initializer values that do not have the potential for side effects can be safely wrapped. Previously, pure annotations were not considered when analyzing the values and this caused classes to be retained that could be considered safe for removal. To resolve this issue, pure annotations are now considered when analyzing a property's initializer value for potential side effects.

### DIFF
--- a/packages/angular_devkit/build_angular/src/babel/plugins/adjust-static-class-members.ts
+++ b/packages/angular_devkit/build_angular/src/babel/plugins/adjust-static-class-members.ts
@@ -57,6 +57,33 @@ export function getKeywords(): Iterable<string> {
 }
 
 /**
+ * Determines whether a property and its initializer value can be safely wrapped in a pure
+ * annotated IIFE. Values that may cause side effects are not considered safe to wrap.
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/babel/plugins/adjust-static-class-members.ts
```diff
@@ -56,6 +56,34 @@ export function getKeywords(): Iterable<string> {
...
+ */
+function canWrapProperty(propertyName: string, assignmentValue: NodePath): boolean {
+  if (angularStaticsToWrap.has(propertyName)) {
+    return true;
+  }
+
+  const { leadingComments } = assignmentValue.node as { leadingComments?: { value: string }[] };
+  if (
+    leadingComments &&
+    leadingComments.some(
```
alan-agius4, line 76: Can use optional chaining here.

alan-agius4 (approved):

PR #21401, opened 2021-07-22
refactor(@schematics/angular): improve automatic versions

alan-agius4:
With this change we improve the versions we set in the users project to be less confusing.  Previously, for both stable and next versions we set the versions of Angular packages to the below;

```
"@angular/animations": "~12.2.0-"
```

This might be problematic as one might not immediately notice that they might depend on a prerelease. The `-` is short character and can be missed when having a large set of dependencies.

With this can we are more explicate by adding `--next` when it's a prerelease and omit the suffix for stable releases.

### DIFF
--- a/packages/schematics/angular/utility/latest-versions.ts
+++ b/packages/schematics/angular/utility/latest-versions.ts
@@ -7,15 +7,15 @@
  */
 
 /** Retrieve the minor version for the provided version string. */
-function getEarliestMinorVersion(version: string) {
-  const versionMatching = version.match(/^(\d+)\.(\d+)\.*/);
+function getAngularEarliestMinorVersion(version: string): string {
+  const versionMatching = version.match(/^(\d+)\.(\d+)\.\d+(-\w+)?/);
 
   if (versionMatching === null) {
     throw Error('Unable to determine the minor version for the provided version');
   }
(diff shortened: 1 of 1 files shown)


packages/schematics/angular/utility/latest-versions.ts
```diff
@@ -8,14 +8,14 @@
 
 /** Retrieve the minor version for the provided version string. */
 function getEarliestMinorVersion(version: string) {
-  const versionMatching = version.match(/^(\d+)\.(\d+)\.*/);
+  const versionMatching = version.match(/^(\d+)\.(\d+)\.(?:\d+)(-\w+)?/);
```
josephperrott, line 11: Is it possible that `@schematics/angular` could move into `rc` while `@angular/core` is still in `next` resulting in a mismatch for this discovered `prerelease` segment?

josephperrott (approved):
LGTM

Just one question to check on

packages/schematics/angular/utility/latest-versions.ts
clydin, line 11: nit: the grouping around the patch number shouldn't be needed

alan-agius4: I don't think so, the FW is always released first.

PR #21429, opened 2021-07-27
refactor(@angular/cli): convert `npm-package-arg` usage from require to import

clydin:
The typings package was already present for `npm-package-arg` and the removal of require usage supports the eventual shift to an ESM output for the CLI package.

### DIFF
--- a/packages/angular/cli/commands/add-impl.ts
+++ b/packages/angular/cli/commands/add-impl.ts
@@ -8,6 +8,7 @@
 
 import { analytics, tags } from '@angular-devkit/core';
 import { NodePackageDoesNotSupportSchematics } from '@angular-devkit/schematics/tools';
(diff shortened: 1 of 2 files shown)


packages/angular/cli/commands/add-impl.ts
```diff
@@ -235,6 +231,28 @@ export class AddCommand extends SchematicCommand<AddCommandSchema> {
     return this.executeSchematic(collectionName, options['--']);
   }
 
+  private async isProjectVersionValid(packageIdentifier: npa.Result): Promise<boolean> {
```
clydin, line 234: This was refactored into a separate method to avoid the 200 line limit for the `run` method.

alan-agius4 (approved):

PR #21437, opened 2021-07-28
fix(@angular-devkit/build-angular): add priority to copy-webpack-plug…

### LINKED ISSUE #21427: Serve and build output different assets
# 🐞 Bug report

- [ ] new
- [x] build
- [x] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, the previous version in which this bug was not present was: 11.x

### Description

Due to an update of copy-webpack-plugin (from 6.2.1 to 9.0.0 in version 12.1.x) the cli now outputs different assets when serving and building an application. This only occurs, when multiple asset configuration have the same output folder and use globs matching files with equal names. The reason for this is that copy-webpack-plugin (which is used when running `ng serve`) applies all globs in parallel and stores the result into a shared array. This leads to a race condition, as the order in which the asset files are copied to the output folder depend on how fast each glob is evaluated. This is not the case with `ng build` as the copy-webpack-plugin is not used at all, but all assets are copied sequentially by a utils function.

## 🔬 Minimal Reproduction

I created a [reproduction repo](https://github.com/originalfrostig/angular-assets-rc), with two asset configurations, where the first one should be overwritten by the second one. This is only the case for `ng build` and not for `ng serve`, due to a nested folder inside `src/assets` as it increases the duration for evaluating the first glob.

## 🌍 Your Environment

<pre><code>

                |___/
    

Angular CLI: 12.1.3
Node: 14.17.3
Package Manager: yarn 1.22.5
OS: win32 x64

Angular: 12.1.3

@angular-devkit/architect       0.1201.3
@angular-devkit/build-angular   12.1.3
@angular-devkit/core            12.1.3
@angular-devkit/schematics      12.1.3
@schematics/angular             12.1.3

</code>
</pre>

**Anything else relevant?**
In my opinion this bug could be solved by utilizing the priority property, when converting `AssetPatternClass` (@angular/cli) to `Pattern` (copy-webpack-plugin). As a result the array mentioned above would be reordered before the files are copied, leading to a consistent behaviour when serving and building an application. If you agree with me this is a bug and should be fixed, I could create a PR.

originalfrostig:
Added priority property to copy-webpack-plugin patterns to ensure all assets are processed in order. 

Closes #21427

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -15,7 +15,7 @@ import {
   GLOBAL_DEFS_FOR_TERSER_WITH_AOT,
   VERSION as NG_VERSION,
 } from '@angular/compiler-cli';
-import CopyWebpackPlugin from 'copy-webpack-plugin';
+import CopyWebpackPlugin, { ObjectPattern } from 'copy-webpack-plugin';
 import { createHash } from 'crypto';
 import { createWriteStream, existsSync, promises as fsPromises } from 'fs';
 import * as path from 'path';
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):

originalfrostig:
@alan-agius4 Don't mention it. Thanks all the great work you guys are doing, I am thrilled that I could contribute a little.

PR #21440, opened 2021-07-28
remove unused non-bazel unit test script

alan-agius4:
This script is no longer used.


alan-agius4:
Will fix commit validation issue later!

clydin (approved):

PR #21484, opened 2021-08-03
feat(@angular-devkit/build-angular): add `type=module` to all scripts tags

alan-agius4:
With this change we add `type=module` to all script tags. This is now possible since IE is no longer supported.

More information about modules can be found here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/package-chunk-sort.ts
+++ b/packages/angular_devkit/build_angular/src/utils/package-chunk-sort.ts
@@ -9,36 +9,35 @@
 import { ExtraEntryPoint } from '../builders/browser/schema';
 import { normalizeExtraEntryPoints } from '../webpack/utils/helpers';
 
-export function generateEntryPoints(appConfig: {
+export type EntryPointsType = [name: string, isModule: boolean];
+
+export function generateEntryPoints(options: {
   styles: ExtraEntryPoint[];
   scripts: ExtraEntryPoint[];
-}) {
+  isHMREnabled?: boolean;
+}): EntryPointsType[] {
   // Add all styles/scripts, except lazy-loaded ones.
-  const extraEntryPoints = (
-    extraEntryPoints: ExtraEntryPoint[],
-    defaultBundleName: string,
-  ): string[] => {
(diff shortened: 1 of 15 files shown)


packages/angular_devkit/build_angular/src/utils/package-chunk-sort.ts
```diff
@@ -9,35 +9,38 @@
...
+    ['polyfills', true],
+    ...extraEntryPoints(options.styles, 'styles'),
+    ...extraEntryPoints(options.scripts, 'scripts'),
+    ['vendor', true],
+    ['main', true],
   ];
 
   const duplicates = [
-    ...new Set(entryPoints.filter((x) => entryPoints.indexOf(x) !== entryPoints.lastIndexOf(x))),
+    ...new Set(
```
clydin, line 39: Is this Set still useful here?

packages/angular_devkit/build_angular/src/dev-server/index.ts
```diff
@@ -305,17 +305,19 @@ export function serveWebpackBrowser(
     switchMap(({ browserOptions, webpackConfig, locale }) => {
       if (browserOptions.index) {
         const { scripts = [], styles = [], baseHref } = browserOptions;
-        const entrypoints = generateEntryPoints({ scripts, styles });
+        const entrypoints = generateEntryPoints({
+          scripts,
+          styles,
+          isHMREnabled: webpackConfig.devServer?.hot,
```
clydin, line 311: Can you add a comment here explain why this HMR check is currently needed?
Hopefully we can remove it in the future.

clydin (approved):

packages/angular_devkit/build_angular/src/utils/package-chunk-sort.ts
alan-agius4: I think not.

packages/angular_devkit/build_angular/src/dev-server/index.ts
alan-agius4: Sure

PR #21493, opened 2021-08-04
feat(@angular-devkit/build-angular): drop support for `karma-coverage-instanbul-reporter`

alan-agius4:
BREAKING CHANGE:

Support for `karma-coverage-instanbul-reporter` has been dropped in favor of the official karma coverage plugin `karma-coverage`.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
@@ -86,32 +86,11 @@ const init: any = (config: any, emitter: any) => {
   config.reporters.unshift('@angular-devkit/build-angular--event-reporter');
 
   // When using code-coverage, auto-add karma-coverage.
-  if (options.codeCoverage) {
-    config.plugins = config.plugins || [];
-    config.reporters = config.reporters || [];
-    const { plugins, reporters } = config;
-    const hasCoveragePlugin = plugins.some(isPlugin('karma-coverage', 'reporter:coverage'));
-    const hasIstanbulPlugin = plugins.some(
-      isPlugin('karma-coverage-istanbul-reporter', 'reporter:coverage-istanbul'),
-    );
-    const hasCoverageReporter = reporters.includes('coverage');
-    const hasIstanbulReporter = reporters.includes('coverage-istanbul');
-    if (hasCoveragePlugin && !hasCoverageReporter) {
-      reporters.push('coverage');
-    } else if (hasIstanbulPlugin && !hasIstanbulReporter) {
-      // coverage-istanbul is deprecated in favor of karma-coverage
-      reporters.push('coverage-istanbul');
-    } else if (!hasCoveragePlugin && !hasIstanbulPlugin) {
-      throw new Error('karma-coverage must be installed in order to run code coverage.');
-    }
-
-    if (hasIstanbulPlugin) {
-      logger.warn(
-        `'karma-coverage-istanbul-reporter' usage has been deprecated since version 11.\n` +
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
```diff
@@ -86,32 +86,11 @@ const init: any = (config: any, emitter: any) => {
...
-    );
-    const hasCoverageReporter = reporters.includes('coverage');
-    const hasIstanbulReporter = reporters.includes('coverage-istanbul');
-    if (hasCoveragePlugin && !hasCoverageReporter) {
-      reporters.push('coverage');
-    } else if (hasIstanbulPlugin && !hasIstanbulReporter) {
-      // coverage-istanbul is deprecated in favor of karma-coverage
-      reporters.push('coverage-istanbul');
-    } else if (!hasCoveragePlugin && !hasIstanbulPlugin) {
-      throw new Error('karma-coverage must be installed in order to run code coverage.');
```
clydin, line 105: Is this plugin check still useful?

alan-agius4: Need to check if Karma throw a meaningful error when not installed.

packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
```diff
@@ -86,32 +86,11 @@ const init: any = (config: any, emitter: any) => {
   config.reporters.unshift('@angular-devkit/build-angular--event-reporter');
 
   // When using code-coverage, auto-add karma-coverage.
-  if (options.codeCoverage) {
-    config.plugins = config.plugins || [];
-    config.reporters = config.reporters || [];
-    const { plugins, reporters } = config;
-    const hasCoveragePlugin = plugins.some(isPlugin('karma-coverage', 'reporter:coverage'));
```
devoto13, line 93: I guess `isPlugin` function definition can be cleaned up as well?

packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
devoto13: There is an error message, but it is pretty [generic](https://github.com/karma-runner/karma/blob/2b71a3c6a01757e0b5c1bf8d241b588656b62127/lib/plugin.js#L85). Probably better to have a dedicated error message in the CLI with the specific migration instructions or an update schematic as the switch requires both dependency changes and [configuration changes](https://github.com/angular/angular-cli/commit/8995e49f0bc2de3cba2f2fea87afe898f7bedf71#diff-f23d77c1515d2212caa684c6edee8c197a846c385405851ea9be3bd90f3179f9) (i.e. just installing karma-coverage is not enough).

packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
alan-agius4: Now it's needed :)

packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
alan-agius4: I think `Can not load reporter "coverage", it is not registered! Perhaps you are missing some plugin?` is good enough. 

Regarding the migration. As far as I reckon back in version 11 we decided that we'll it up to the users to migrate as it was non trivial.

clydin (approved):

PR #21497, opened 2021-08-04
Update Angular peer dependency ranges to v13 prerelease

clydin:
This sets the peer dependency ranges for Angular packages to the upcoming v13.

### DIFF
--- a/packages/schematics/angular/utility/latest-versions.ts
+++ b/packages/schematics/angular/utility/latest-versions.ts
@@ -15,7 +15,7 @@ export const latestVersions: Record<string, string> & {
   ...require('./latest-versions/package.json')['dependencies'],
 
   // As Angular CLI works with same minor versions of Angular Framework, a tilde match for the current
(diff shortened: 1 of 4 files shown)


packages/schematics/angular/utility/latest-versions.ts
```diff
@@ -6,6 +6,18 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+/** Retrieve the minor version for the provided version string. */
+function getAngularEarliestMinorVersion(version: string): string {
```
alan-agius4, line 10: Wouldn't this cause CI failures again when we change the major/minor version?

clydin: Left as hardcoded for now.
Once the build system is fully transitioned to Bazel, we can use stamping to set the version automatically at build time.

alan-agius4 (approved):

PR #21499, opened 2021-08-04
Bump version to "v12.1.5" with changelog.

dgp1130:


alan-agius4:
Today’s release should be 12.2

clydin:
Based on the change set, I think we should skip this patch release.
Also, the change log appears to be unformatted (validate check is failing).

dgp1130:
Alright, dropping this release and moving on to the `12.2.0` release.

PR #21500, opened 2021-08-04
Bump version to "v12.2.0" with changelog.

dgp1130:


dgp1130:
Unfortunately the release notes are still empty, do we expect stuff from `12.2.0-next.*` and `12.2.0-rc.*` to be in here? I can update that manually.

dgp1130:
Updated the changelog to copy over everything from `12.2.0-next.*` and `12.2.0-rc.*`. Took a while since some of the older releases use a different format, hopefully I didn't introduce any errors or miss anything.

The tool also wanted to format the file, which slightly adjusted some of the older releases, but I think it's better to do it now than later.

josephperrott (approved):
LGTM

PR #21535, opened 2021-08-09
perf(@angular-devkit/build-angular): use esbuild as a CSS optimizer for global styles

alan-agius4:
Esbuild now support CSS sourcemaps which now makes it possible to be used to optimize global CSS.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -13,13 +13,13 @@ import { ExtraEntryPoint } from '../../builders/browser/schema';
 import { SassWorkerImplementation } from '../../sass/sass-service';
 import { BuildBrowserFeatures } from '../../utils/build-browser-features';
 import { WebpackConfigOptions } from '../../utils/build-options';
-import { maxWorkers } from '../../utils/environment-options';
 import {
   AnyComponentStyleBudgetChecker,
   PostcssCliResources,
   RemoveHashPlugin,
   SuppressExtractedTextChunksWebpackPlugin,
 } from '../plugins';
+import { CssOptimizerPlugin } from '../plugins/css-optimizer-plugin';
 import {
   assetNameTemplateFactory,
   getOutputHashFormat,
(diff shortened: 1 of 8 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -264,64 +265,26 @@ export function getStylesConfig(wco: WebpackConfigOptions): webpack.Configuratio
...
-        },
+        minify: minimizeCss,
       }),
-      // Global styles use cssnano since sourcemap support is required even when minify
-      // is enabled. Once esbuild supports style sourcemaps this can be changed.
-      // esbuild stylesheet source map support issue: https://github.com/evanw/esbuild/issues/519
       new CssMinimizerPlugin({
         test: /\.css$/,
         include: globalBundlesRegExp,
         parallel: maxWorkers,
```
clydin, line 283:
```suggestion
        parallel: false,
```
No longer needed with esbuild and enabling can actually result in slower builds.

packages/angular_devkit/build_angular/src/webpack/plugins/css-optimizer.ts
```diff
@@ -0,0 +1,51 @@
...
+  const { code, warnings, map } = await esbuild.transform(input, {
+    loader: 'css',
+    minify: true,
+    sourcemap: minimizerOptions?.sourceMap && 'external',
+    sourcefile,
+  });
+
+  // Merge sourcemaps
+  let fullSourcemap;
+  if (minimizerOptions?.sourceMap) {
```
clydin, line 27: Esbuild should be able to remap the source map if provided.  Unfortunately, it can't be passed in as an option but it will pick it up if it is in the source map url comment.
Rough code:
```
          let sourceMapLine;
          if (inputMap) {
            // esbuild will automatically remap the sourcemap if provided
            sourceMapLine = '\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,';
            sourceMapLine += Buffer.from(
              typeof inputMap === 'string' ? inputMap : JSON.stringify(inputMap),
            ).toString('base64');
          }

          const { code, map, warnings } = await esbuild.transform(
            sourceMapLine ? input + sourceMapLine : input,
            {
              legalComments : 'inline',
              loader: 'css',
              minify: true,
              sourcefile,
              sourcemap: inputMap ? 'external' : false,
            },
          );
```

clydin (approved):

PR #21562, opened 2021-08-12
fix(@angular-devkit/build-angular): handle undefined entrypoints when marking async chunks

SimonPrimetzhofer:
When using custom builders, it is possible to remove entries from the webpack-configuration manually. This may eventually lead to undefined entryPoints in async-chunks.ts - this fix handles exactly this problem


clydin (changes requested):
The code change seems reasonable.
Can you add the PR description to the commit message body?  Also can you expand the commit subject to maybe something similar to `handle undefined entry points when marking async chunks`?
Otherwise looks good.

SimonPrimetzhofer:
Hello,
thanks for the reply, I have just updated the commit-message and added the description in the body.

Best regards,
Simon

clydin (approved):

PR #21590, opened 2021-08-17
docs: update release documentation regarding prerelease semver ranges

clydin:
When transitioning from prerelease to stable, several SemVer ranges for packages referenced in the `package.json` files need to be adjusted so that the unsupported prerelease versions are not mentioned in the final stable release.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -55,6 +55,13 @@ In general, cherry picks for LTS should only be done if it meets one of the crit
 
 Releasing is performed using Angular's unified release tooling. Each week, two releases are expected, `latest` and `next` on npm.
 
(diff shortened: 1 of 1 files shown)


dgp1130 (approved):
Thanks for adding this, it's good to document caveats like this for releases.

I'm a bit concerned caretakers won't read this at the time that it's relevant. Ideally, this is something that release tooling would do automatically. Can we surface this as a bug to dev infra as well (or maybe address it in the fixit) so we don't need to remember to do this during major releases?

clydin:
Being able to extend the Bazel stamping configuration during release is on the backlog for dev-infra which should be able to be used to support use case such as this.
Ideally, no manual steps should be needed for the release.

PR #21626, opened 2021-08-23
feat(@angular-devkit/build-angular): expose declarations in AngularWebpackPlugin

Kamshak:
When creating an instance of the AngularWebpackPlugin where the `CompilerOptions` contain `declarations: true` it is now possible to access the declarations emitted by the typescript compiler through the `FileEmitter`. This makes it possible to write a Webpack plugin to emit .d.ts files for files compiled through the AngularWebpackPlugin.

This is useful when using the angular-cli to bundle libraries intended to be consumed in other frameworks or when ng-packagr cannot be used.

### DIFF
--- a/packages/ngtools/webpack/src/ivy/plugin.ts
+++ b/packages/ngtools/webpack/src/ivy/plugin.ts
@@ -655,6 +655,7 @@ export class AngularWebpackPlugin {
 
       let content: string | undefined;
       let map: string | undefined;
+      let declaration: string | undefined;
       program.emit(
         sourceFile,
         (filename, data) => {
@@ -662,6 +663,8 @@ export class AngularWebpackPlugin {
             map = data;
           } else if (filename.endsWith('.js')) {
             content = data;
+          } else if (filename.endsWith('.d.ts')) {
+            declaration = data;
           }
         },
         undefined,
@@ -683,7 +686,7 @@ export class AngularWebpackPlugin {
         ...getExtraDependencies(sourceFile),
       ].map(externalizePath);
 
-      return { content, map, dependencies, hash };
+      return { content, map, declaration, dependencies, hash };
     };
   }
 }
--- a/packages/ngtools/webpack/src/ivy/symbol.ts
+++ b/packages/ngtools/webpack/src/ivy/symbol.ts
@@ -11,6 +11,7 @@ export const AngularPluginSymbol = Symbol.for('@ngtools/webpack[angular-compiler
 export interface EmitFileResult {
   content?: string;
   map?: string;
+  declaration?: string;
   dependencies: readonly string[];
   hash?: Uint8Array;
 }


alan-agius4:
Thanks for your contribution.

This plugin however, is only intended to be used to compile Angular applications.

Libraries should be bundled following the the [APF](https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview) spec, which ng-packagr does.

What’s the use-case where ng-packagr cannot be used?

Kamshak:
1) Paths imports in NX workspace: ng-packagr requires all files to be packaged to be in the root dir of the project to be built. In an nx workspace paths are however mapped using tsconfig paths. Some support for this [has been added](https://github.com/angular/angular/pull/33828) however I could not get this to work and I believe some configurations are [unsupported](https://github.com/angular/angular/issues/29361#issuecomment-474616177).

2) Custom webpack config used in libraries (for example base64 svg loaders using resourceGroups) is not possible when using ng-packagr (since it's not using webpack). We customize our webpack config in the app and use custom postcss plugins which is [not supported](https://github.com/ng-packagr/ng-packagr/issues/1418#issuecomment-725358476) in ng-packagr. We need this because we export some Components as WebComponents for usage outside of Angular and the cli/AngularCompilerPlugin seems to be the way to go for this.

3) APF seems to be mostly geared towards publishing libraries to be consumed by Angular ("The format described here uses a distinct file layout and metadata configuration that enables a package to work seamlessly under most common scenarios where Angular is used, and makes it compatible with the tooling offered by the Angular team and the community itself."). Another usecase we have is to bundle a bunch of Angular services together and export an injector that can be used in React:


Angular Side:
```ts
// In main.ts
export const getApiInjector = (baseUrl) => ɵcreateInjector(DataAccessModule, null, [
  {
    provide: API_BASE_URL,
    useValue: baseUrl,
  },
  {
    provide: DOCUMENT,
    useValue: document,
  },
  {
    provide: PLATFORM_ID,
    useValue: "browser",
  },
  {
    provide: AuthTokenProvider,
    useValue: {
      currentAuthToken$,
    },
  },
]);
export { ApiService } from '@monorepo/shared/data-access-sutter';
```

React Side:
```ts
// React app's main.tsx will import this and can use it to get access to Services

import { getApiInjector, ApiService } from '../../../dist/apps/data-access/main';
const injector = getApiInjector('https://sutter.dev.innovint.us');
const apiService: ApiService = injector.get(ApiService);

```

Mainly, using ng-packagr has the constraint that we just cannot use custom webpack configuration (loaders, postcss config, resource groups, ...) that we use to build the main app. Using the same build tool (webpack) to create a library package with webcomponents is very appealing.

With this addition to the plugin I can write a small Webpack plugin that will write the declarations file, e.g.
```
	config.plugins.push({
		apply(compiler: webpack.Compiler) {
			compiler.hooks.thisCompilation.tap('CopyDeclarationToCorrectPlace', (compilation) => {
				let angularPluginFileEmitter: FileEmitterCollection | null = null;
				compilation.compiler.webpack.NormalModule.getCompilationHooks(compilation).loader.tap(
					'CopyDeclarationToCorrectPlace',
					(loaderContext: any) => {
						if (loaderContext[AngularPluginSymbol]) {
							angularPluginFileEmitter = loaderContext[AngularPluginSymbol];
						}
					}
				);
				compilation.hooks.finishModules.tapPromise('CopyDeclarationToCorrectPlace', async () => {
					if (!angularPluginFileEmitter) {
						throw new Error('No angular file emitter found');
					}
					await angularPluginFileEmitter.emit(config.entry.main[0]).then((result: any) => {
						if (!result.declaration) {
							throw new Error('Declaration not emitted');
						}
						compilation.emitAsset('main.d.ts', new webpack.sources.RawSource(result.declaration));
					});
				});
			});
		},
	});
```

PR #21627, opened 2021-08-23
WIP:Feat/extends angular cli questionnaire

fstodulski:
### Extend AngularCLI Questionnaire
**Current behaviour:**
Currently Angular CLI asks only few questions during creating new application. 
1. Name
2. Routing
3. Styles

Rest of configuration has to be done after project is created.

Based on VueCLI I would like to extend CLI questionnaire with some additional options. In my opinion, they would speed up a little generating new Workspace.

**Expected behaviour:**
`ng new` triggers Angular CLI schematics.
User has to answer to few more questions before new workspace will be created.

1. Name (duh)
2. Would you like to have Inline Styles? (y/N) `by default it's False`
3. Would you like to have Inline Template? (y/N) `by default is' False` - this one is often used for small components. Personal choice
4. Which Change Detection Strategy would you like to use? `OnPush / Default`. Currently developer has to manually edit angular.json to include this option for newly created components
5. Prefix for selectors (app) - `by default is 'app'`, but sometimes we need to change it in case of having multiple projects + library
6. Would you like to add Angular routing? (y/N) - default one. Nothing has changed.
7. Which stylesheet format would you like to use? - default one. Nothing has changed.

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -31,7 +31,7 @@ import { relativePathToWorkspaceRoot } from '../utility/paths';
 import { validateProjectName } from '../utility/validation';
 import { getWorkspace, updateWorkspace } from '../utility/workspace';
 import { Builders, ProjectType } from '../utility/workspace-models';
-import { Schema as ApplicationOptions, Style } from './schema';
+import { Schema as ApplicationOptions, ChangeDetection, Style } from './schema';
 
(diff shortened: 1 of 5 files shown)


alan-agius4:
Thanks for your contribution. However, at this point in time we'd like to limit the number of prompts to show during `ng-new`. 

In future ideally before taking on a feature, open an issue to discuss this.

Thanks.

PR #21737, opened 2021-09-09
fix(@angular-devkit/build-angular): handle `FORCE_COLOR` when stdout is not instance of `WriteStream`

### LINKED ISSUE #21732: Allow forcing colors
Hi,

I run angular CLI tasks in my own scripts that internally use `concurrently`. This setup disables colors in the angular CLI issue.

It would be great to be able to force color support via an env var, like `FORCE_COLOR`.

Related to:
* https://github.com/angular/angular-cli/issues/6478
* https://github.com/open-cli-tools/concurrently/issues/238

alan-agius4:
In some cases, custom implementation of stdout, don't extend `WriteStream` which causes colors not to be included in the output.

Closes #21732

### DIFF
--- a/packages/angular/cli/utilities/color.ts
+++ b/packages/angular/cli/utilities/color.ts
@@ -10,7 +10,33 @@ import * as ansiColors from 'ansi-colors';
 import { WriteStream } from 'tty';
 
 type AnsiColors = typeof ansiColors;
-const supportsColor = process.stdout instanceof WriteStream && process.stdout.getColorDepth() > 1;
+
+function supportColor(): boolean {
+  if (process.env.FORCE_COLOR !== undefined) {
+    // 2 colors: FORCE_COLOR = 0 (Disables colors), depth 1
+    // 16 colors: FORCE_COLOR = 1, depth 4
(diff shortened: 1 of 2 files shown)


packages/angular/cli/utilities/color.ts
```diff
@@ -6,21 +6,50 @@
...
+ export function removeColor(text: string): string {
+   // This has been created because when colors.enabled is false unstyle doesn't work
+   // see: https://github.com/doowb/ansi-colors/blob/a4794363369d7b4d1872d248fc43a12761640d8e/index.js#L38
+   return text.replace(ansiColors.ansiRegex, '');
+ }
+ 
+ // Create a separate instance to prevent unintended global changes to the color configuration
+ // Create function is not defined in the typings. See: https://github.com/doowb/ansi-colors/pull/44
+ const colors = (ansiColors as AnsiColors & { create: () => AnsiColors }).create();
+ colors.enabled = supportsColor;
```
clydin, line 52: nit: could avoid the variable and put the call here.

clydin (approved):

PR #21739, opened 2021-09-09
fix(@angular-devkit/build-angular): add web-streams-polyfill to downlevel exclusion list

clydin:
Polyfill related packages should not be downlevelled due to the nature of their code which may need to perform feature testing or leverage native capabilities to extract browser support information necessary to properly polyfill a given browser. In the case of `web-streams-polyfill`, it leverages the `%AsyncIteratorPrototype%`, when available, to fully polyfill its stream implementations.  To access `%AsyncIteratorPrototype%`, a native async generator is needed and therefore the code present in the package cannot have this case of a native async generator downlevelled. `core-js` is also excluded for similar (and additional reasons).

Closes: #21735


alan-agius4 (approved):

filipesilva:
@clydin this PR conflicts on 12.2.x, can you create a new PR for that branch?

PR #21763, opened 2021-09-16
fix(@angular-devkit/build-angular): support WASM-based esbuild optimizer fallback

clydin:
In the event that the Angular CLI is executed on a platform that does not yet have native support for esbuild, the WASM-based variant of esbuild will now be used. If the first attempt to optimize a file fails to execute the native variant of esbuild, future executions will instead use the WASM-based variant instead which will execute regardless of the native platform. The WASM-based variant, unfortunately, can be significantly slower than the native version (some cases can be several times slower). For install time concerns regarding the esbuild post-install step, esbuild is now listed as an optional dependency which will allow the post-install step to fail but allow the full npm install to pass. This install scenario should only occur in the event that the esbuild native binary cannot be installed or is otherwise unavailable.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-worker.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-worker.ts
@@ -7,8 +7,9 @@
  */
 
 import remapping from '@ampproject/remapping';
-import { TransformFailure, transform } from 'esbuild';
+import type { TransformFailure, TransformResult } from 'esbuild';
 import { minify } from 'terser';
+import { EsbuildExecutor } from './esbuild-executor';
 
 /**
  * A request to optimize JavaScript using the supplied options.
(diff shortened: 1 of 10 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/javascript-optimizer-worker.ts
```diff
@@ -18,61 +19,76 @@ interface OptimizeRequest {
...
+    /**
+     * Specifies the target ECMAScript version for the output code.
+     */
     target: 5 | 2015 | 2016 | 2017 | 2018 | 2019 | 2020;
+    /**
+     * Controls whether esbuild should only use the WASM-variant instead of trying to
+     * use the native variant. Some platforms may not support the native-variant and
+     * this option allows one support test to be conducted prior to all the workers starting.
+     */
+    usingWasmEsbuild: boolean;
```
alan-agius4, line 56: Drive by comment. I think grammatically `use` would be a better choice here.

alan-agius4 (approved):

PR #21771, opened 2021-09-17
Support a future ESM-only `@angular/compiler-cli` package (1)

clydin:
With the Angular CLI currently being a CommonJS package, this change uses a dynamic import to load `@angular/compiler-cli` which may be ESM. CommonJS code can load ESM code via a dynamic import. Unfortunately, TypeScript will currently, unconditionally downlevel dynamic import into a require call. require calls cannot load ESM code and will result in a runtime error. To workaround this, a Function constructor is used to prevent TypeScript from changing the dynamic import. Once TypeScript provides support for keeping the dynamic import this workaround can be dropped and replaced with a standard dynamic import.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/build-options.ts
+++ b/packages/angular_devkit/build_angular/src/utils/build-options.ts
@@ -7,7 +7,7 @@
  */
 
 import { logging } from '@angular-devkit/core';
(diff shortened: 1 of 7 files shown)


devversion (approved):
🎉

alan-agius4 (approved):

PR #21778, opened 2021-09-20
refactor(@ngtools/webpack): support an ESM-only `@angular/compiler-cli` package

clydin:
With the Angular CLI currently being a CommonJS package, this change uses a dynamic import to load @angular/compiler-cli which may be ESM. CommonJS code can load ESM code via a dynamic import. Unfortunately, TypeScript will currently, unconditionally downlevel dynamic import into a require call. require calls cannot load ESM code and will result in a runtime error. To workaround this, a Function constructor is used to prevent TypeScript from changing the dynamic import. Once TypeScript provides support for keeping the dynamic import this workaround can be dropped and replaced with a standard dynamic import.
Type only static import statements for `@angular/compiler-cli` are also now used since the `@angular/compiler-cli` is used as a peer dependency and has the potential to not be present. As a result static imports should only be used for types and value imports should be dynamic so that they can be guarded in the event of package absence.

BREAKING CHANGE:
Applications directly using the `webpack-cli` and not the Angular CLI to build must set the environment variable `DISABLE_V8_COMPILE_CACHE=1`. The `@ngtools/webpack` package now uses dynamic imports to provide support for the ESM `@angular/compiler-cli` package. The `v8-compile-cache` package used by the `webpack-cli` does not currently support dynamic import expressions and will cause builds to fail if the environment variable is not specified. Applications using the Angular CLI are not affected by this limitation.


devversion (approved):

alan-agius4 (approved):

alan-agius4:
Note: there is still a require in  https://github.com/angular/angular-cli/blob/c4bce16ee8ee314146d3056a07b65b39b670041d/packages/ngtools/webpack/src/ivy/plugin.ts#L579

alan-agius4:
The E2E failure looks genuine.

clydin:
`OptimizeFor` needs to be exported from the main entry point to correct that require usage.  Once the first set of FW changes lands we can clean that up.

clydin:
The E2E failure was traced to the `webpack-cli` itself which appears to not support native dynamic import expressions.  Programmatic usage of Webpack is fully functional both within the Angular CLI and directly but the `webpack-cli` consistently crashes.

alan-agius4 (approved):

PR #21781, opened 2021-09-21
fix(@angular-devkit/build-webpack): better handle concurrent dev-servers

alan-agius4:
Webpack-dev-server doesn't handle concurrency very well. When using port 0, and 2 processes start at the same time, they end up being given the same port. The main reason for the issue is that it find a free port, put only uses at a later stage.


alan-agius4:
Needed for angular/angular#43417 and https://github.com/angular/universal/pull/2313

clydin (approved):

PR #21782, opened 2021-09-21
refactor(@schematics/angular): remove migrations for unsupported versions

alan-agius4:


tests/legacy-cli/e2e/tests/update/update-9.ts
```diff
@@ -1,30 +1,26 @@
...
   try {
-    await createProjectFromAsset('8.0-project', true, true);
+    await createProjectFromAsset('9.0-project', true, true);
 
     await setRegistry(false);
     await installWorkspacePackages();
 
-    // Update Angular to 9
-    await installPackage('@angular/cli@8');
-    const { stdout } = await ng('update', '@angular/cli@9.x', '@angular/core@9.x');
```
clydin, line 18: Could we keep this check to make sure migrations are actually running?  Can just move it down to the 10 execution.

alan-agius4: done

Issue #21013, opened 2021-06-01
Add option for globally updating @angular/cli@latest


twerske:
# 🚀 Feature request

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [x] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

When I `ng update` I frequently get the message: 

> The installed local Angular CLI version is older than the latest stable version.
> Installing a temporary version to perform the update.

Instead I'd like the option to download the latest CLI as part of the update steps (or instructions on how to do so, so that I'm not continuously updating without updating my global CLI version).

### Describe the solution you'd like

Something like:

> The installed local Angular CLI version is older than the latest stable version.
> Would you like to download the latest version of you local Angular CLI? (y/N)
> *runs npm i -g @angular/cli@latest before continuing*

### Describe alternatives you've considered

Instead of running the command in a prompt it could be included in the output of the ng update command as an additional update:

```
    We analyzed your package.json, there are some packages to update:
    
      Name                               Version                  Command to update
     --------------------------------------------------------------------------------
      @angular/cdk                       11.0.0-rc.0 -> 12.0.2    ng update @angular/cdk
      @angular/cli                       11.0.0-rc.1 -> 12.0.2    ng update @angular/cli
      @angular/core                      11.0.0-rc.1 -> 12.0.2    ng update @angular/core
      @angular/material                  11.0.0-rc.0 -> 12.0.2    ng update @angular/material
     --------------------------------------------------------------------------------
      global @angular/cli                       11.0.0-rc.1 -> 12.0.2    npm i -g @angular/cli@latest
```

PR #21784, opened 2021-09-21
Support null for options parameter from OptionTransform type

### LINKED ISSUE #21023: Passing null options to RunSchematicTask leads to CLI crash
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [x] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 11.x

### Description

 
Using `context.addTask(new RunSchematicTask(schematicName, Options))` inside a schematic, we can schedule another schematic. Until (including) Angular 11 we could simply pass `null` as second `Options` parameter.
Starting with Angular 12 the CLI will crash with the error message `Cannot read property 'project' of null`.

## 🔬 Minimal Reproduction

Run a schematic (using `ng add` or `ng g`) that itself tries to run another schematic using `context.addTask(new RunSchematicTask(schematicName, null))` (`null` is important here).

Since the `ng-add` schematic of `@da-mkay/ng-builder-electron` uses the above mentioned code you can use these commands to reproduce the issue:

    $ ng new workspace --create-application=false
    $ cd workspace
    $ ng add @da-mkay/ng-builder-electron@0.2.0

(Version 0.2.0 is important as I will release a new version in a moment)

## 🔥 Exception or Error

<pre><code>
Cannot read property 'project' of null
</code></pre>

## 🌍 Your Environment

<pre><code>
Angular CLI: 12.0.2
Node: 14.17.0
Package Manager: npm 6.14.13
OS: darwin x64

Angular: 12.0.2

@angular-devkit/architect    0.1200.2
@angular-devkit/core         12.0.2
@angular-devkit/schematics   12.0.2
@schematics/angular          12.0.2
rxjs                         6.6.7
typescript                   4.2.4
</code></pre>

**Anything else relevant?**

The issue is caused by the following line inside `SchematicCommand` class:
https://github.com/angular/angular-cli/blob/master/packages/angular/cli/models/schematic-command.ts#L247

    typeof (current as Record<string, unknown>).project === 'string'

Here, `current` is what we pass as second argument to the `RunSchematicTask` constructor (`null`). Since that parameter is typed as generic `T`, IMHO we should be able to pass `null` here. But maybe I'm wrong 😉

As a workaround I can pass `{}` instead of `null` to the `RunSchematicTask` constructor.

A solution would be to change the above line to:

    typeof (current as Record<string, unknown>)?.project === 'string'

(added `?`)

I can also create a PR for that gigantic change if needed 😋

ddubrava:
Fixes #21023

### DIFF
--- a/goldens/public-api/angular_devkit/schematics/tools/index.md
+++ b/goldens/public-api/angular_devkit/schematics/tools/index.md
@@ -117,7 +117,7 @@ export abstract class FileSystemEngineHostBase implements FileSystemEngineHost_2
     // (undocumented)
     registerContextTransform(t: ContextTransform): void;
     // (undocumented)
-    registerOptionsTransform<T extends object, R extends object>(t: OptionTransform<T, R>): void;
+    registerOptionsTransform<T extends object | null, R extends object>(t: OptionTransform<T, R>): void;
     // (undocumented)
     registerTaskExecutor<T>(factory: TaskExecutorFactory<T>, options?: T): void;
     // (undocumented)
@@ -235,7 +235,7 @@ export interface NodeWorkflowOptions {
     // (undocumented)
     force?: boolean;
     // (undocumented)
-    optionTransforms?: OptionTransform<object, object>[];
+    optionTransforms?: OptionTransform<Record<string, unknown> | null, object>[];
     // (undocumented)
     packageManager?: string;
     // (undocumented)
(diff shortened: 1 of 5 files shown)


ddubrava:
Hi! I’m interested in merging :) so if you want to close the issue (#21013), I will complete this PR (resolve conflicts, concerns)

alan-agius4:
In that case can you please fix the conflicts and afterwards resolve the failing CI steps by running the below.

```
yarn bazel run //packages/angular_devkit/schematics:schematics_api.accept
yarn ng-dev format changed
```

Thanks

alan-agius4 (approved):
LGTM, can you please squash the commits into the first two?

alan-agius4 (approved):

PR #21801, opened 2021-09-23
feat(@schematics/angular): support pipe name prefix

shaharkazaz:
like components often you want to prefix your pipes

BREAKING CHANGE: creating a new pipe will prefix the name with the project's prefix if defined


dgp1130:
The style guide does not include names, and we try to follow that as closely as we can. @jelbourn I know you're looking at updating the style guide, maybe this is a change we should consider making? If so we can revive this PR to match.

PR #21813, opened 2021-09-24
feat(@schematics/angular): componentClass.name instead of string vari…

kievsash:
**Description:**
At the moment ng new component <name> generates .spec file with describe(**'className'**, () => {}) string variable

I propose to replace it with describe(**className.name**, () => {})

image


kievsash:
@googlebot I signed it!

Issue #19026, opened 2020-10-09
Generation of test: Use Classname.name in describe of spec files instead of 'Classname'


Rogier-blip:
# 🚀 feature request
When a spec file is generated by the CLI use Classname.name instead of the 'Classname'.

### Relevant Package
@angular/cli

### Description
It would be nice if the spec files would be generated in a way that makes them adaptable to change.

If a test is generated by the CLI the initial `describe` looks like this:
```typescript
...
describe('AppComponent', () => {
...
``` 

When I change the name of the app component from `AppComponent` to `AppComponent1`. I have to change the name of the test by hand. 

### Describe the solution you'd like
The solution that I have in mind is to use the .name property of the typescript class. The describe would have the following structure. 

```typescript
...
describe(AppComponent.name, () => {
...
```
Now when the AppComponent name is changed through an IDE. The test name is automatically changed.

Cheers 🐵

alan-agius4:
There is actually a PR for this https://github.com/angular/angular-cli/pull/18843

dgp1130:
We had a discussion about this in among the team with a significant amount of bike shedding on this topic.

The main advantage of using `AppComponent.name` is that using an automated refactor to rename the class will also update the test name. Which is very convenient and I know I personally have forgotten to update `describe()` names before.

However, there are a few, less obvious downsides to consider:
* Readability: Using a simple string is more clear about the intent of labeling the `describe()` block, whereas using `Class.name` takes more mental effort to track. It is also one less layer of indirection, since developers don't have to figure out the meaning of `name`.
* `name` property is not well known: Many developers not familiar with many of the intricacies of JavaScript may not know the `name` property. To them, it looks like its accessing a static `name` property on the `AppComponent` class, which doesn't appear to be defined anywhere. It's also technically on the `Function` prototype, since classes in JavaScript are just functions, which can be unintuitive to someone not familiar with the decades of legacy JavaScript syntax and coming from a more object-oriented language. Such a developer might reasonably think this is some kind of Angular magic included via the `@Component()` decorator or injected by the Angular compiler. If TS supported a `nameof()` operator like in C#, this might be a more reasonable change.
* Language service: The TS language service is crashing for me atm, but I believe it does not show `name` in autocomplete, which can add to confusion about the property.
* Uncommon pattern: It is standard practice to just use string literals in `describe()` and `it()` blocks. While that's not universal, there is an advantage to staying with the mainstream and reducing variety in the ecosystem when it is unnecessary.
[...]

PR #21814, opened 2021-09-24
feat(@schematics/angular): componentClass.name instead of string vari…

kievsash:
**Description:**
At the moment ng new component <name> generates .spec file with describe(**'className'**, () => {}) string variable

I propose to replace it with describe(**className.name**, () => {})

image


alan-agius4:
Thanks for your contribution. However, this is not something that we'd like to do. 

Please see https://github.com/angular/angular-cli/issues/19026 for more context.

PR #21815, opened 2021-09-24
fix(@angular/cli): remove unused cli project options.

alan-agius4:
This fixes the JSON workspace schema and removed options projects `.project.cli`. Currently only `defaultCollection` is used from `projects.cli` other options are ignored.

https://github.com/angular/angular-cli/blob/9afe185fc61390efec6bfbb16a0586487fd6f118/packages/angular/cli/models/schematic-command.ts#L381

### DIFF
--- a/packages/angular/cli/lib/config/workspace-schema.json
+++ b/packages/angular/cli/lib/config/workspace-schema.json
@@ -139,7 +139,10 @@
       "type": "object",
       "properties": {
         "cli": {
-          "$ref": "#/definitions/cliOptions"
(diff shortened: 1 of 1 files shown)


clydin:
Will this cause any breakage if a developer happens to have any of the other items in the schema present in the global configuration file?

alan-agius4:
The `projects` section shouldn’t be in the global configuration, so it should cause any issue in that regard.

alan-agius4:
Any further concern her?

clydin (approved):

PR #21825, opened 2021-09-27
perf(@angular-devkit/build-angular): change webpack hashing function to `xxhash64`

alan-agius4:
`xxhash64` uses a faster WASM based hash mechanism.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/app-shell/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/app-shell/index.ts
@@ -143,7 +143,7 @@ async function _getServerModuleBundlePath(
     throw new Error(`Could not find server output directory: ${outputPath}.`);
   }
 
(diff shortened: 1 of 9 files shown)


clydin:
Is there a referenced to what the options enables and/or changes that we can track?

alan-agius4:
The docs don’t explicitly mention the options https://webpack.js.org/configuration/experiments/#experimentsfuturedefaults

alan-agius4:
Maybe we set each option manually to the future default and add an inline commit that this is the future default?

clydin:
I think that’s a good idea. For this PR I think just updating the hash function should be good since that’s pretty much the majority of the PR.
My concern is that the “futureDefaults” option appears to be fluid and may change in minors in potentially breaking ways.

alan-agius4:
@clydin PTAL.

clydin (approved):

PR #21827, opened 2021-09-27
 feat(@angular-devkit/build-angular): enable disk cache by default and provide configurable options

alan-agius4:
Persistent disk build cache is now enabled by default. A number of options have been added to allow fine tuning of the cache.

The options can be configuration in `cli.cache` section in the `angular.json` as shown below.

- `enabled`: Configure whether disk caching is enabled. Defaults to `true`
- `environment`: Configure in which environment disk cache is enabled. Valid values `ci`, `local` or `all`. Defaults to: `local`
- `path`: cache base path. Defaults to `.angular/cache`

DEPRECATED: `NG_BUILD_CACHE` environment variable option will be removed in the next major version. Configure `cli.cache` in the workspace configuration instead.

BREAKING CHANGE:  `NG_PERSISTENT_BUILD_CACHE` environment variable option no longer  have effect. Configure `cli.cache` in the workspace configuration instead.

```json
{
  "version": 1,
  "cli": {
    "cache": {
      "enabled": true,
      "path": ".cache",
      "environment": "all"
    }
  }
  ...
}
```


clydin:
I think we should consider keeping `NG_BUILD_CACHE` (at least for now).  It's very useful for testing and debugging scenarios and I believe some users are leveraging it in CI environments to adjust the path to the specifics of the project's CI setup which may differ from the local caching location (most likely left as default).

alan-agius4:
@clydin PTAL

clydin (approved):

filipesilva:
@alan-agius4 this PR now conflicts with the target branch, can you rebase please?

PR #21849, opened 2021-10-01
refactor(@schematics/angular): drop IE-only polyfills and drop versions from comment

dgp1130:
This removes polyfills only required on Internet Explorer since it is no longer supported (`web-animations-js`). Also updates the doc comment to leave specific versions unspecified, since they are already out of date and updating them now would just cause them to go out of date again in the future. Instead, users can visit the browser support guide to understand find the most up to date browser versions supported.

Not sure if this is considered a breaking change, but personally I don't view the exact contents of any generated files to be strict public API. Since this PR doesn't change the polyfills actually used in a new project, I don't think this needs to be called out as a breaking change.

### DIFF
--- a/packages/schematics/angular/application/files/src/polyfills.ts.template
+++ b/packages/schematics/angular/application/files/src/polyfills.ts.template
@@ -8,8 +8,8 @@
  *      file.
  *
  * The current setup is for so-called "evergreen" browsers; the last versions of browsers that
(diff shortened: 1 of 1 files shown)


clydin (approved):
Some additional background on the removal: https://github.com/angular/angular/issues/43035

IgorMinar (approved):
Yay!

alan-agius4 (approved):

PR #21852, opened 2021-10-01
feat(@schematics/angular): drop polyfills required only for Internet Explorer now that support has been dropped for it

dgp1130:
Removes the `classlist.js` and `web-animations-js` polyfills from any polyfills files in a workspace. Specifically, it looks for `import 'classlist.js';` and `import 'web-animations-js';`, then removes those them along with any preceeding comments (aside from the file overview and browser polyfills header comment).

Previous versions of `ng new` generated a polyfills file with commented out imports for both modules, to avoid including them by default while giving users an easy way to enable them. This migration also looks for these commented out imports and removes them along with associated descriptive comments. This has no effect on the application, but cleans up unnecessary noise from the polyfills file.

This PR was heavily inspired by a [previous migration](https://github.com/angular/angular-cli/blob/9.1.x/packages/schematics/angular/migrations/update-8/drop-es6-polyfills.ts) which happened to be very similar.

**BREAKING CHANGE:**

`classlist.js` and `web-animations-js` are removed from application polyfills and uninstalled from the package. These were only needed for compatibility with Internet Explorer, which is no longer needed now that Angular only supports evergreen browsers. See: https://angular.io/guide/browser-support.

Add the following to the polyfills file for an app to re-add these packages:

```typescript
import 'classlist.js';
import 'web-animations-js';
```

And then run:

```sh
npm install classlist.js web-animations-js --save
```

### DIFF
--- a/packages/schematics/angular/migrations/update-13/drop-ie-polyfills_spec.ts
+++ b/packages/schematics/angular/migrations/update-13/drop-ie-polyfills_spec.ts
@@ -1,0 +1,659 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { EmptyTree } from '@angular-devkit/schematics';
+import { SchematicTestRunner, UnitTestTree } from '@angular-devkit/schematics/testing';
+
+describe('Migration to remove polyfills specific to Internet Explorer', () => {
+  const schematicRunner = new SchematicTestRunner(
+    'migrations',
+    require.resolve('../migration-collection.json'),
+  );
+
+  let tree: UnitTestTree;
+
+  beforeEach(async () => {
+    tree = await schematicRunner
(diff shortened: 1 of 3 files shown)


packages/schematics/angular/migrations/update-13/drop-ie-polyfills_spec.ts
```diff
@@ -0,0 +1,538 @@
...
+  "devDependencies": {
+    "@angular/cli": "^13.0.0"
+  }
+}
+    `.trim(),
+    );
+
+    // Assert that `npm install` is scheduled.
+    const taskNames = schematicRunner.tasks.map((task) => task.name);
+    expect(taskNames).toEqual(['node-package']);
```
dgp1130, line 536: I wasn't sure how to test that the `npm install` task got scheduled, so I just looked for its name here. I'm not super happy with this though, so I'm open to suggestions for better ways to verify this.

clydin: That is  pretty much the best option currently.

PR #21853, opened 2021-10-01
feat(@angular-devkit/build-angular): favor es2020 main fields

alan-agius4:
APF version 13 will include es2020, with this change we favor es2020 entrypoints.


itea-dev:
There seems to be a contradiction between the title and the description.

> APF version 13 will include es2020, with this change we favor **es2015** entrypoints.

alan-agius4:
@itea-dev thanks for pointing that out. Updated.

itea-dev:
@alan-agius4 You're welcome. Thank you for your hard work.

clydin (approved):

PR #21854, opened 2021-10-01
feat(@angular/cli): officially support Node.js v16

clydin:
Node.js v16 will be entering an LTS state prior to the release of Angular v13 which allows Node.js v16 to be marked as officially supported by the Angular CLI. The initial bootstrapping check now adds Node.js v16 to the output message in the event of an unsupported Node.js version.
NOTE: Prior to the final v13 release, the Node.js v16 minor should be updated to the actual LTS version once available.


LayZeeDK:
Will Node.js 12.x support be dropped?

alan-agius4:
> Will Node.js 12.x support be dropped?

No, Node.JS 12 is still in LTS. https://nodejs.org/en/about/releases/

alan-agius4 (approved):

PR #21855, opened 2021-10-01
refactor: update `@angular/dev-infra-private` package

dgp1130:
This is to unblock the `12.2.8` release with https://github.com/angular/dev-infra/pull/247.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "@angular/compiler": "13.0.0-next.8",
     "@angular/compiler-cli": "13.0.0-next.8",
     "@angular/core": "13.0.0-next.8",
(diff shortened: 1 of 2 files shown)


alan-agius4:
Already being addressed here https://github.com/angular/angular-cli/pull/21851

dgp1130:
Ah, didn't realize the robot beat me to it. I'll merge that one since it's the standard update.

devversion:
On a related note: Ideally the dev-infra package would have been updated in the patch version as well. I'm not sure why in the past this didn't happen. `12.2.x` and `master` are currently diverged and have different configuration files for `.ng-dev`. 

Usually it's good keeping th build setup in sync for version branches (such as `master` or `12.2.x`).

PR #21873, opened 2021-10-05
fix(@angular-devkit/build-angular): enable custom `es2020` and `es2015` conditional exports

clydin:
By adding the `es2020` and `es2015` condition names, the build process will now use ES2020 or ES2015 specific files if a package's exports entries contain files for either of those conditions. This allows packages to provide multiple variants of the code that the bundler can then use based on the bundler's configuration. As of Angular v13, ES2020 code is preferred. However, some packages may prefer to export other variants by default but provide an `es2020`/`es2015` condition to allow other variants if requested.


packages/angular_devkit/build_angular/src/webpack/configs/browser.ts
```diff
@@ -55,6 +55,7 @@ export function getBrowserConfig(wco: WebpackConfigOptions): webpack.Configurati
     devtool: false,
     resolve: {
       mainFields: ['es2020', 'es2015', 'browser', 'module', 'main'],
+      conditionNames: ['es2020', 'es2015', '...'],
```
IgorMinar, line 58: shouldn't this also contain `esm2020` and `fesm2020` as we discussed with @devversion ?

clydin: Currently only the flat files are used by the CLI which will be enabled by the `es2020` and `es2015` conditions. With the FW (and ng-packagr generated) libraries prioritizing the `es2020` condition, the CLI will use the ES2020 flat files for APF packages.  This mimics the pre-exports behavior with the main fields on the line above. If we added `esm2020` then the non-flat files would get bundled which unfortunately results in larger application sizes (at least currently with Webpack).

PR #21881, opened 2021-10-06
docs: mark `@angular-devkit/build-optimizer` as deprecated.

dgp1130:
It's functionality has been included in `@angular-devkit/build-angular` so this package is no longer needed by the CLI and we will stop publishing the package soon.


alan-agius4:
Nit: deprecations should be included in the commit message using `DEPRECATED:`.

https://github.com/angular/angular/blob/master/CONTRIBUTING.md

dgp1130:
Thanks @alan-agius4, I was wondering about that. Updated the commit message.

PR #21953, opened 2021-10-14
refactor: consistent import from webpack style

lekhmanrus:
Explicit import from `webpack` whenever it is possible for consistency reasons.


lekhmanrus:
@clydin, thanks! 🙂

PR #22015, opened 2021-10-26
feat(@angular_devkit/core): add titlecase string util

JaimeStill:
add titlecase string util


alan-agius4:
Hi @JaimeStill, 

Thanks for your contribution.

However since this utils is not needed directly by the tooling packages. If you need this in your project it's suggested that you add it there.

PR #22051, opened 2021-11-01
fix(@angular-devkit/build-angular): use es2015 when generating server bundles

alan-agius4:
ES2020 can contain JS syntax which are not compatible with all supported Node.js versions such as optional chaining.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -300,9 +300,9 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
       symlinks: !buildOptions.preserveSymlinks,
       modules: [tsConfig.options.baseUrl || projectRoot, 'node_modules'],
       mainFields: isPlatformServer
-        ? ['es2020', 'es2015', 'module', 'main']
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -300,9 +300,9 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
       symlinks: !buildOptions.preserveSymlinks,
       modules: [tsConfig.options.baseUrl || projectRoot, 'node_modules'],
       mainFields: isPlatformServer
-        ? ['es2020', 'es2015', 'module', 'main']
+        ? ['es2015', 'module', 'main']
         : ['es2020', 'es2015', 'browser', 'module', 'main'],
-      conditionNames: ['es2020', 'es2015', '...'],
+      conditionNames: isPlatformServer ? ['node', 'es2015', '...'] : ['es2020', 'es2015', '...'],
```
clydin, line 305:
```suggestion
      conditionNames: isPlatformServer ? ['es2015', '...'] : ['es2020', 'es2015', '...'],
```
`node` should be automatically added by Webpack based on the target.

clydin (approved):

PR #22056, opened 2021-11-02
refactor(@angular-devkit/build-angular): clean up webpack configurations

alan-agius4:
With this change we remove the worker and typescript configuration.


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -258,6 +263,18 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
     });
   }
 
+  if (main || polyfills) {
+    extraRules.push({
+      test: /\.[jt]sx?$/,
```
clydin, line 268:
```suggestion
      test: /\.[cm]?[tj]sx?$/,
```
That should cover all JS and upcoming TS variations.

clydin (approved):

PR #22070, opened 2021-11-03
Bump version to "v13.0.0" with changelog.

dgp1130:


josephperrott (approved):
LGTM

dgp1130:
`CHANGELOG` includes some manual edits were I removed notes about inline critical CSS being disabled by default, because that was reverted (https://github.com/angular/angular-cli/pull/21899), so there is no change to users.

We should probably have a follow up discussion about how to handle reverts in the changelog, but a manual edit is the easiest solution right now.

PR #22071, opened 2021-11-03
docs: release notes for the v13.0.0 release

josephperrott:


dgp1130 (approved):
Is it supposed to remove all the `-next.*` and `-rc.*` changelog entries? I would think we would want to keep those?

josephperrott:
@dgp1130 Yes, they are removed as all of the commits from those entries are contained in the `13.0.0` release notes entry.  The release notes for `13.0.0` are generated from the range`13.0.0-next.0..13.0.0`

PR #22103, opened 2021-11-05
refactor: fix ng-new template pointing to Angular animations with Material logo

devversion:
Fixes that the `new new` generated app component template currently
provides a link to Angular Animations with the Angular Material logo,
according to: https://angular.io/presskit.

This commit switches the link to Angular Material with the assumption
that this was intended to point to Angular Material from the
beginning on.

tempsnip

### DIFF
--- a/packages/schematics/angular/application/other-files/app.component.html.template
+++ b/packages/schematics/angular/application/other-files/app.component.html.template
@@ -366,6 +366,12 @@
       <svg class="material-icons" xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M10 6L8.59 7.41 13.17 12l-4.58 4.59L10 18l6-6z"/></svg>
     </a>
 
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):

twerske (changes requested):
Since Angular Material is already in the above section, can we change this logo to the Animations logo from https://angular.io/presskit#animations instead?

devversion:
@twerske Updated as per offline discssion (removal of the CLI and Material icons below; and just having Angular Material above in the useful links section). Here is a screenshot and a test-app:

image

https://stackblitz.com/edit/angular-g8empx?file=src%2Fapp%2Fautocomplete-auto-active-first-option-example.html 

Note that I just took the Angular Material logo and turned it gray to match up with the other icons.

twerske (approved):
LGTM - thank you!

alan-agius4 (approved):

PR #22133, opened 2021-11-10
fix(@angular-devkit/core): add Angular CLI major version as analytics dimension

### LINKED ISSUE #22130: Telemetry collection: collect Angular major version as a separate dimension/metric
# 🚀 Feature request

### Description

I'd like to able to easily build dashboards that show Angular adoption status by major version. Currently we only collect the exact patch version which is hard to post-process within Google Analytics.

We can add a new custom dimension, or we could reuse "Browser Version" field for this purpose (we currently use "Application Version" to collect the patch version).

I've tried to group the metrics via "starts with" filters, but then I can only draw one version graph per dashboard widget. I'd like to be able to have a single diagram that contains all major versions.

alan-agius4:
With this change we replace the custom dimension 8 `AOT Enabled`, with `Angular CLI Major Version`. The motivation behind replacing this dimension is that the there is already an `aot` dimension with id 13 which serves for the same purpose.

More information to why we need a new dimension for the Angular CLI major version can be found #22130

Closes #22130

### DIFF
--- a/docs/design/analytics.md
+++ b/docs/design/analytics.md
@@ -51,7 +51,7 @@ Note: There's a limit of 20 custom dimensions.
 | 5 | `Flag: --style` | `string` |
 | 6 | `--collection` | `string` |
 | 7 | `Flag: --strict` | `boolean` |
-| 8 | `AOT Enabled` | `boolean` |
+| 8 | `Angular CLI Major Version` | `string` |
 | 9 | `Flag: --inline-style` | `boolean` |
 | 10 | `Flag: --inline-template` | `boolean` |
 | 11 | `Flag: --view-encapsulation` | `string` |
--- a/goldens/public-api/angular_devkit/core/src/index.md
+++ b/goldens/public-api/angular_devkit/core/src/index.md
@@ -1064,7 +1064,7 @@ class MultiAnalytics implements Analytics {
 // @public
 enum NgCliAnalyticsDimensions {
     // (undocumented)
-    AotEnabled = 8,
+    AngularCLIMajorVersion = 8,
     // (undocumented)
     BuildErrors = 20,
     // (undocumented)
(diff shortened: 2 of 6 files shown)


docs/design/analytics.md
```diff
@@ -51,7 +51,7 @@ Note: There's a limit of 20 custom dimensions.
 | 5 | `Flag: --style` | `string` |
 | 6 | `--collection` | `string` |
 | 7 | `Flag: --strict` | `boolean` |
-| 8 | `AOT Enabled` | `boolean` |
+| 8 | `Angular CLI Major Version` | `boolean` |
```
clydin, line 54: Should the type be a string (or a number)?

alan-agius4: Good catch

alan-agius4: @IgorMinar, would you prefer the value to be a `string` or `number`? As we are uncertain what works bests for filter and querying in GA.

IgorMinar: I'm actually not sure. It seems that Google analytics converts everything to strings for querying. Let's go with number and we can adjust.

IgorMinar:
Can we not reuse this dimension and instead create a new one or use browser version as suggested? The reason being that the dimension is already used for other purposes which then results in queries that span long time periods into the past to return garbage or unbelievable results. Thanks

alan-agius4:
@IgorMinar, so both dimension 8 and 13 are used in the reports to determine if a build is AOT or not?

We cannot add an additional dimension because we are already using the maximum allowed which is 20. With regards of using the Browser Version, this is currently used to track which OS is being used by building a fake UA string, I think in the past we found to be useful. Also this is not a custom dimension and I need to verify if GA API has some validations for this field. (Did try to look it up in the docs but got lost 😞)

IgorMinar:
> @IgorMinar, maybe we can use aid (app Id) or aiid (app installer Id)? The latter is currently not used.

Maybe that would work, but @mgechev asked me about analytics 360 which would give us more dimensions. Let's explore if we can get on that plan and resolve this without hacks. We could use additional dimensions for future metrics anyway.

alan-agius4:
Yeah I brought that up with Minko, yesterday. That would allow us to have 200 dimensions and 200 metrics. On the free version we only 20.

mgechev (approved):

PR #22166, opened 2021-11-15
feat(@schematics/angular): loosen project name validation

### LINKED ISSUE #11051: ng new <project-name> -- Using "." or "_" in"project-name" produces errors.
### Versions
Angular CLI: 6.0.5
Node: 10.2.1
OS: win32 x64
Angular:
...

@angular-devkit/architect    0.6.5
@angular-devkit/core         0.6.5
@angular-devkit/schematics   0.6.5
@schematics/angular          0.6.5
@schematics/update           0.6.5
rxjs                         6.2.0
typescript                   2.7.2

Windows 10

### Repro steps

* ng new project.1
* ng new project_1

### Observed behavior
```
<!-- Normally this includes a stack trace and some more information. -->
-------------------------------------------------------------
$ ng new project.1
Project name "project.1" is invalid.
-------------------------------------------------------------
$ ng new project_1
Schematic input does not validate against the Schema: {"name":"project_1","version":"6.0.5","newProjectRoot":"projects","skipInstall":false,"linkCli":false,"skipGit":false,"commit":null}
Errors:

  Data path ".name" should match format "html-selector".
-------------------------------------------------------------


### Desired behavior
<!--
What would like to see implemented?
What is the usecase?
What did you expect to see?
-->

* Expect "." or "_" to be valid characters in a project name as was in the case in older versions of CLI.

### PR DESCRIPTION
**Pull request by alan-agius4**
With this change we update the validation of the libraries and application projects names to fully allow characters that make a valid NPM package name. http://json.schemastore.org/package has been used as reference.

We also remove validators that are no longer needed.

Closes #11051

### DIFF
--- a/packages/schematics/angular/application/index.ts
+++ b/packages/schematics/angular/application/index.ts
@@ -11,7 +11,6 @@ import {
   MergeStrategy,
   Rule,
   SchematicContext,
-  SchematicsException,
   Tree,
   apply,
   applyTemplates,
@@ -28,7 +27,6 @@ import { Schema as ComponentOptions } from '../component/schema';
 import { NodeDependencyType, addPackageJsonDependency } from '../utility/dependencies';
 import { latestVersions } from '../utility/latest-versions';
 import { relativePathToWorkspaceRoot } from '../utility/paths';
-import { validateProjectName } from '../utility/validation';
 import { getWorkspace, updateWorkspace } from '../utility/workspace';
 import { Builders, ProjectType } from '../utility/workspace-models';
 import { Schema as ApplicationOptions, Style } from './schema';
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: packages/schematics/angular/application/index.ts [hunk 1]
```diff
@@ -264,13 +260,20 @@ export default function (options: ApplicationOptions): Rule {
     const workspace = await getWorkspace(host);
     const newProjectRoot = (workspace.extensions.newProjectRoot as string | undefined) || '';
     const isRootApp = options.projectRoot !== undefined;
+
+    // If scoped project (i.e. "@foo/bar"), convert dir to "foo/bar".
+    let folderName = options.name.startsWith('@') ? options.name.substr(1) : options.name;
+    if (/[A-Z]/.test(folderName)) {
```
clydin, line 266: Can you add test cases for the scoped and un-scoped dasherized cases? (`@foo/BarBaz`, etc.)

clydin (approved):

packages/schematics/angular/application/index.ts
alan-agius4: Added some more tests.

PR #22167, opened 2021-11-15
feat(@schematics/angular): add VS Code configurations when generating a new workspace

### LINKED ISSUE #2672: Recommended VS Code configuration
### OS?

> Windows 10, Ubuntu, Mac

I think lots of developers out there use VS Code for ng2 projects. It would be very cool to have some kind of recommended configuration on the following VS Code settings for an optimal user experience. 
- Recommended extensions for ng2 developers (i.e. tslint extention to enable style ng2 guide linting)
- Build Task Configuration to use integrated build runner (ctrl+shift+b)
- How to configure the debugger to be usable

Could be written for other ide's like webstorm to.

alan-agius4:
When generating a new workspace we now generate several VS Code configurations.

- extensions: we added config to recommand to use official VS Code extensions.
- launch: added run and debug launch scripts for `ng test` and `ng serve`
- tasks: configured `npm start` and `npm test` to better support the launch scripts

More information about each files can be found
- extensions: https://go.microsoft.com/fwlink/?linkid=827846
- launch: https://go.microsoft.com/fwlink/?linkid=830387
- tasks: https://go.microsoft.com/fwlink/?LinkId=733558

Closes #2672

### DIFF
--- a/packages/schematics/angular/workspace/files/__dot__vscode/extensions.json.template
+++ b/packages/schematics/angular/workspace/files/__dot__vscode/extensions.json.template
@@ -1,0 +1,4 @@
+{
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=827846
+  "recommendations": ["angular.ng-template"]
+}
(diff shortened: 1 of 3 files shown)


dgp1130 (approved):
LGTM. I assume we don't want to migrate existing projects since they may already have their own configuration and may not want such a configuration if they don't use VSCode.

Should we consider an optional migration for those than want to update or "canonicalize" their debugging configuration in existing projects?

alan-agius4:
> Should we consider an optional migration for those than want to update or "canonicalize" their debugging configuration in existing projects?

I don’t feel too strongly about this, but we can surly do an optional migration if we think it will be useful.

clydin (approved):

Issue #21394, opened 2021-07-22
Show gzip compression size after build


splincode:
# 🚀 Feature request

### Command (mark with an `x`)

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

```
$ ng build
- Generating browser application bundles (phase: setup)...
✔ Browser application bundle generation complete.
- Generating ES5 bundles for differential loading...
✔ Browser application bundle generation complete.
✔ ES5 bundle generation complete.
- Copying assets...
✔ Copying assets complete.
- Generating index html...
✔ Index html generation complete.
Initial Chunk Files                    | Names                |      Size
main-es5.99064527cd544b123bc9.js       | main                 | 175.56 kB
main-es2015.99064527cd544b123bc9.js    | main                 | 149.94 kB
polyfills-es5.90dd4e2aec326855ce1b.js  | polyfills-es5        |  93.01 kB
styles.0b064ae4068cf2388220.css        | styles               |   5.02 kB
runtime-es2015.a0da3e8506fcfd740c60.js | runtime              |   1.16 kB
runtime-es5.a0da3e8506fcfd740c60.js    | runtime              |   1.16 kB
| Initial ES5 Total    | 274.75 kB
| Initial ES2015 Total | 156.12 kB
Build at: 2021-07-21T10:58:01.250Z - Hash: 2398fbe5637097cd0447 - Time: 17873ms
```

### Describe the solution you'd like

```
$ ng build
- Generating browser application bundles (phase: setup)...
✔ Browser application bundle generation complete.
- Generating ES5 bundles for differential loading...
✔ Browser application bundle generation complete.
✔ ES5 bundle generation complete.
- Copying assets...
✔ Copying assets complete.
- Generating index html...
✔ Index html generation complete.
Initial Chunk Files                    | Names                |      Size                 |      GZip
main-es5.99064527cd544b123bc9.js       | main                 | 175.56 kB                 | 100.26 kB
main-es2015.99064527cd544b123bc9.js    | main                 | 149.94 kB                 | 80.14 kB
polyfills-es5.90dd4e2aec326855ce1b.js  | polyfills-es5        |  93.01 kB                 |  43.11 kB
[...]

PR #22172, opened 2021-11-16
feat(@angular-devkit/build-angular): add estimated transfer size to build output report

clydin:
When optimizations are enabled (either scripts or styles), an additional column will now be present in the output report shown in the console for an application build. This additonal column will display the estimated transfer size for each file as well as the total initial estimated transfer size for the initial files. The estimated transfer size is determined by calculating the compressed size of the file using brotli's default settings. In a development configuration (a configuration with optimizations disabled), the calculations are not performed to avoid any potential increase in rebuild speed due to the large size of unoptimized files.

```
Initial Chunk Files           | Names         |  Raw Size | Estimated Transfer Size
main.51c3048e87eac4fd.js      | main          | 132.07 kB |                38.63 kB
polyfills.df676194f15dea6d.js | polyfills     |  36.20 kB |                11.47 kB
runtime.ab1d1f80a99b92f8.js   | runtime       |   1.05 kB |               601 bytes
styles.ef46db3751d8e999.css   | styles        |   0 bytes |                       -

                              | Initial Total | 169.33 kB |                50.69 kB
```

Closes: #21394

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -31,6 +31,7 @@ import {
   ScriptsWebpackPlugin,
 } from '../plugins';
 import { ProgressPlugin } from '../plugins/progress-plugin';
+import { TransferSizePlugin } from '../plugins/transfer-size-plugin';
 import { createIvyPlugin } from '../plugins/typescript';
 import {
   assetPatterns,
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -284,6 +285,7 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
         removeLicenses: buildOptions.extractLicenses,
         advanced: buildOptions.buildOptimizer,
       }),
+      new TransferSizePlugin(),
```
alan-agius4, line 288: I think we should only add this if the it's a browser build. There is no reason why the transfer size should be displayed when building a server bundle.

alan-agius4 (approved):
LGTM, other than one comment.

PR #22182, opened 2021-11-17
ci: remove `cla/google` from required statues

alan-agius4:
The CLA tools no longer report statuses as it is being run as Github actions.


josephperrott (approved):
LGTM

PR #22200, opened 2021-11-19
fix(@angular/cli): logic which determines which temp version of the CLI is to be download during `ng update`

alan-agius4:
Previously, when using an older version of the Angular CLI, during `ng update`, we download the temporary `latest` version to run the update. The ensured that when running that the runner used to run the update contains the latest bug fixes and improvements.

This however, can be problematic in some cases. Such as when there are API breaking changes, when running a relatively old schematic with the latest CLI can cause runtime issues, especially since those schematics were never meant to be executed on a CLI X major versions in the future.

With this change, we improve the logic to determine which version of the Angular CLI should be used to run the update.

Below is a summarization of this.

- When using the `--next` command line argument, the `@next` version of the CLI will be used to run the update.
- When updating an `@angular/` or `@nguniversal/` package, the target version will be used to run the update. Example: `ng update @angular/core@12`,  the update will run on most recent patch version of `@angular/cli` of that major version `@12.2.6`.
- When updating an `@angular/` or `@nguniversal/` and no target version is specified. Example: `ng update @angular/core` the update will run on most latest version of the `@angular/cli`.
- When updating a third-party package, the most recent patch version of the installed `@angular/cli` will be used to run the update. Example if `13.0.0` is installed and `13.1.1` is available on NPM, the latter will be used.

### DIFF
--- a/packages/angular/cli/commands/update-impl.ts
+++ b/packages/angular/cli/commands/update-impl.ts
@@ -37,11 +37,6 @@ import {
 } from '../utilities/package-tree';
 import { Schema as UpdateCommandSchema } from './update';
 
-const NG_VERSION_9_POST_MSG = colors.cyan(
-  '\nYour project has been updated to Angular version 9!\n' +
-    'For more info, please see: https://v9.angular.io/guide/updating-to-version-9',
-);
-
 const UPDATE_SCHEMATIC_COLLECTION = path.join(
   __dirname,
   '../src/commands/update/schematic/collection.json',
@@ -56,6 +51,8 @@ const disableVersionCheck =
   disableVersionCheckEnv !== undefined &&
   disableVersionCheckEnv !== '0' &&
   disableVersionCheckEnv.toLowerCase() !== 'false';
+
+const ANGULAR_PACKAGES_REGEXP = /^@(?:angular|nguniversal)\//;
 
 export class UpdateCommand extends Command<UpdateCommandSchema> {
   public override readonly allowMissingWorkspace = true;
@@ -272,19 +269,26 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
   async run(options: UpdateCommandSchema & Arguments) {
     await ensureCompatibleNpm(this.context.root);
 
-    // Check if the current installed CLI version is older than the latest version.
-    if (!disableVersionCheck && (await this.checkCLILatestVersion(options.verbose, options.next))) {
-      this.logger.warn(
-        `The installed local Angular CLI version is older than the latest ${
-          options.next ? 'pre-release' : 'stable'
-        } version.\n` + 'Installing a temporary version to perform the update.',
+    // Check if the current installed CLI version is older than the latest compatible version.
+    if (!disableVersionCheck) {
+      const cliVersionToInstall = await this.checkCLIVersion(
+        options['--'],
+        options.verbose,
+        options.next,
       );
 
-      return runTempPackageBin(
-        `@angular/cli@${options.next ? 'next' : 'latest'}`,
-        this.packageManager,
-        process.argv.slice(2),
-      );
+      if (cliVersionToInstall) {
+        this.logger.warn(
+          'The installed Angular CLI version is outdated.\n' +
+            `Installing a temporary Angular CLI versioned ${cliVersionToInstall} to perform the update.`,
+        );
+
+        return runTempPackageBin(
+          `@angular/cli@${cliVersionToInstall}`,
+          this.packageManager,
+          process.argv.slice(2),
+        );
+      }
     }
 
     const logVerbose = (message: string) => {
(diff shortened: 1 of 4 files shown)


packages/angular/cli/commands/update-impl.ts
```diff
@@ -879,22 +857,63 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
...
+          verbose,
+          usingYarn: this.packageManager === PackageManager.Yarn,
+        },
+      );
+
+      return VERSION.full === version ? null : version;
+    } catch (error) {
+      if (
+        error instanceof Error &&
+        error.message.includes('No matching version found for @angular/cli')
```
clydin, line 908: Which tests fail because of this?  Would it be possible to adjust the tests so that this CI specific error cause isn't needed?  Or are there cases outside of CI where this could happen?

clydin (approved):

packages/angular/cli/commands/update-impl.ts
alan-agius4: Pretty much all the update tests failed, because with verdaccio `@angular/cli@13` will cause the above error. 

I wonder if it will work if we a caret in the version.

clydin: Ah, that's probably because the local version is `@angular/cli@13.1.0-next.2` and prereleases won't be picked up by `13`.
Maybe we could publish a stable version of 13 (or whatever the current major is) in the test setup so that the local registry can resolve the major.

packages/angular/cli/commands/update-impl.ts
```diff
@@ -879,22 +857,63 @@ export class UpdateCommand extends Command<UpdateCommandSchema> {
...
-        verbose,
-        usingYarn: this.packageManager === PackageManager.Yarn,
-      },
-    );
+  async checkCLIVersion(
+    packagesToUpdate: string[] | undefined,
+    verbose = false,
+    next = false,
+  ): Promise<string | null> {
+    let cliRunnerTargetVersion: string | number;
```
dgp1130, line 868: Could this be improved with:

```typescript
const cliRunnerTargetVersion = getRunnerTargetVersion(/* ... */);

function getRunnerTargetVersion(/* ... */): string { /* ... */ }
```

Then all that logic can be encapsulated in its own function and simplified with `return`.

dgp1130 (commented):
> When updating an `@angular/` or `@nguniversal/` and no target version is specified. Example: `ng update @angular/core` the update will run on most latest version of the `@angular/cli`.

Does `ng update @angular/core` go to latest or the next major? If I'm on v10 and run this, it should go to v11 right, since we only allow one major at a time? In that scenario, wouldn't we want to upgrade with the latest patch of the destination major (in this case, `@angular/cli@11`)?

alan-agius4:
> Does ng update @angular/core go to latest or the next major? If I'm on v10 and run this, it should go to v11 right, since we only allow one major at a time? In that scenario, wouldn't we want to upgrade with the latest patch of the destination major (in this case, @angular/cli@11)?

When no version is specified the update will use the `@latest` tag. We can improve this but, it might be confusing that updating an angular package will update to the next major version while other packages will update to `latest`.

packages/angular/cli/commands/update-impl.ts
alan-agius4: Adjusted the test to remove the need for this try/catch block.

alan-agius4:
Note: once merged, I will create PRs to backport this change to all LTS versions of the CLI.

dgp1130:
> When no version is specified the update will use the `@latest` tag. We can improve this but, it might be confusing that updating an angular package will update to the next major version while other packages will update to `latest`.

Hmm, I'm a bit torn here. `npm install @angular/core` means `@angular/core@latest`, so I wouldn't want to diverge from that. It would also be weird for `@angular/*` packages to be special-cased like you mentioned. On the other hand running with the latest CLI means that it needs to support migrating from more than the previous version, which is what we're hoping to avoid.

I just tried this with today's logic to make sure I understood and running `ng update @angular/core @angular/cli` on a v10 project fails because it tries to update to v13, which isn't allowed (though the error message doesn't actually say that and we should probably be more clear that it tried to update to `@latest`).

```
$ ng update @angular/core @angular/cli
Using package manager: 'npm'
Found 29 dependencies.
Updating multiple major versions of '@angular/core' at once is not supported. Please migrate each major version individually.
Run 'ng update @angular/core@11' in your workspace directory to update to latest '11.x' version of '@angular/core'.

For more information about the update process, see https://update.angular.io/?v=10.0-11.0
```

So I guess the decision of which CLI runner to use doesn't really matter since the update will fail anyways if `@latest` is not the next major version. Based on that, I'm ok with using the latest CLI runner as you currently have here, but this is maybe something we should look into when we eventually revamp `ng update`. I would love to see running `ng update` just go to the next major with no fuss, and this is something we'd have to tackle more comprehensively to make that happen safely.

dgp1130 (approved):

PR #22209, opened 2021-11-21
feat(@angular-devkit/build-angular): expose i18nDuplicateTranslation …

### LINKED ISSUE #22201: Expose '--i18n-duplicate-translation`option for build target
### Which @angular/* package(s) are relevant/releated to the feature request?

compiler-cli, compiler, localize

### Description

For applications that use duplicate translations build output become verbose with warnings.
It would be great to expose new `--i18n-duplicate-translation`option on `ng build` in a similar manner that it is already done with `--i18n-missing-translation`.

`translate`already supports this with `duplicateTranslation ` [option](https://github.com/angular/angular/blob/master/packages/localize/tools/src/translate/index.ts#L77), but this option is not exposed up to `ng build`.

### Proposed solution

Expose new `--i18n-duplicate-translation`option on `ng build`.

### Alternatives considered

None. Exposing what is already supported seem like a proper way.

minijus:
…option of browser and server builders

Closes #22201

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/build-options.ts
+++ b/packages/angular_devkit/build_angular/src/utils/build-options.ts
@@ -13,7 +13,7 @@ import {
   Budget,
   CrossOrigin,
   ExtraEntryPoint,
-  I18NMissingTranslation,
+  I18NTranslation,
   IndexUnion,
   InlineStyleLanguage,
   Localize,
@@ -38,7 +38,7 @@ export interface BuildOptions {
   verbose?: boolean;
   progress?: boolean;
   localize?: Localize;
-  i18nMissingTranslation?: I18NMissingTranslation;
+  i18nMissingTranslation?: I18NTranslation;
   bundleDependencies?: boolean;
   externalDependencies?: string[];
   watch?: boolean;
--- a/packages/angular_devkit/build_angular/src/utils/i18n-options.ts
+++ b/packages/angular_devkit/build_angular/src/utils/i18n-options.ts
@@ -12,7 +12,7 @@ import fs from 'fs';
 import module from 'module';
 import os from 'os';
 import path from 'path';
-import { Schema as BrowserBuilderSchema } from '../builders/browser/schema';
+import { Schema as BrowserBuilderSchema, I18NTranslation } from '../builders/browser/schema';
 import { Schema as ServerBuilderSchema } from '../builders/server/schema';
 import { readTsconfig } from '../utils/read-tsconfig';
 import { TranslationLoader, createTranslationLoader } from './load-translations';
@@ -233,6 +233,7 @@ export async function configureI18nBuild<T extends BrowserBuilderSchema | Server
         },
       },
       usedFormats,
+      buildOptions.i18nDuplicateTranslation,
     );
 
     if (usedFormats.size > 1 && tsConfig.options.enableI18nLegacyMessageIdFormat !== false) {
@@ -282,6 +283,7 @@ export function loadTranslations(
   loader: TranslationLoader,
   logger: { warn: (message: string) => void; error: (message: string) => void },
   usedFormats?: Set<string>,
+  duplicateTranslation?: I18NTranslation,
 ) {
   for (const file of desc.files) {
     const loadResult = loader(path.join(workspaceRoot, file.path));
@@ -308,9 +310,18 @@ export function loadTranslations(
       // Merge translations
       for (const [id, message] of Object.entries(loadResult.translations)) {
         if (desc.translation[id] !== undefined) {
-          logger.warn(
-            `WARNING [${file.path}]: Duplicate translations for message '${id}' when merging`,
-          );
+          const duplicateTranslationMessage = `[${file.path}]: Duplicate translations for message '${id}' when merging.`;
+          switch (duplicateTranslation) {
+            case I18NTranslation.Ignore:
+              break;
+            case I18NTranslation.Error:
+              logger.error(`ERROR ${duplicateTranslationMessage}`);
+              break;
+            case I18NTranslation.Warning:
+            default:
+              logger.warn(`WARNING ${duplicateTranslationMessage}`);
+              break;
+          }
         }
         desc.translation[id] = message;
       }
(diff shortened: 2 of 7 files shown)


packages/angular_devkit/build_angular/src/utils/build-options.ts
```diff
@@ -13,6 +13,7 @@ import {
   Budget,
   CrossOrigin,
   ExtraEntryPoint,
+  I18NDuplicateTranslation,
```
minijus, line 16: @alan-agius4 @petebacondarwin do you notice what I am missing here?
It seems that changing schema and running `yarn build` generated schema types that are missing both `I18NDuplicateTranslation` and `I18NMissingTranslation`.

alan-agius4: Since the enum values of `I18NDuplicateTranslation` and `I18NMissingTranslation` are the same. They have been merged into a single type.

```ts
    /**
     * How to handle duplicate translations for i18n.
     */
    i18nDuplicateTranslation?: I18NTranslation;
    /**
     * How to handle missing translations for i18n.
     */
    i18nMissingTranslation?: I18NTranslation;
```

minijus:
I feel that I will need some extra guidance on this one :) @alan-agius4 @petebacondarwin @alxhub your advice/feedback would be very appreciated.

Duplicate and missing translations are handled differently, and it makes sense:
* duplicate translations can be identified during loading
* missing translations can be identified only during inlining

To handle missing translations `Diagnostics` are used and there is single place which register them: https://github.com/angular/angular/blob/master/packages/localize/tools/src/source_file_utils.ts#L379

Meanwhile, duplicate translation handling  (or loading of translations itself) is duplicated (pun intended):
* There is `TranslationsLoader` in `@angular/localize` with `Diagnostics` (deprecated?) based duplicate translations handling: https://github.com/angular/angular/blob/master/packages/localize/tools/src/translate/translation_files/translation_loader.ts#L126
* and there similar [`TranslationsLoader`](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/utils/load-translations.ts#L22) in `@angular/angular-cli` with different (not based on `Diagnostics`) handling of duplicate translations: https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/utils/i18n-options.ts#L254

Couple of questions:
* Should `TranslationsLoader` from `@angular/localize` be used in CLI?
* If not, should handling of duplicate translations in `@angular/angular-cli` be based on `Diagnostics`?
* Should `I18NTranslation` (used to be `I18NMissingTranslation`) somehow point to [`DiagnosticHandlingStrategy`](https://github.com/angular/angular/blob/master/packages/localize/tools/src/diagnostics.ts#L12) instead?

alan-agius4:
Hi @minijus, 

Sorry for taking so long to get back to you. 

> Should TranslationsLoader from @angular/localize be used in CLI?

If that is possible, sure! But, I suspect it's not as straightforward.

> If not, should handling of duplicate translations in @angular/angular-cli be based on Diagnostics?

If not, I assume that we can just change the `warn` here to either be ignored or changed to error.
https://github.com/angular/angular-cli/blob/1a9da01b3a13f037b91c5502b6fa09055539bbe9/packages/angular_devkit/build_angular/src/utils/i18n-options.ts#L313

> Should I18NTranslation (used to be I18NMissingTranslation) somehow point to DiagnosticHandlingStrategy instead?

`I18NTranslation` is generated from a JSON schema and there is no good way that this can be replaced with an import to `DiagnosticHandlingStrategy`

alan-agius4:
@minijus, is this something that you still want to spear head?

minijus:
> @minijus, is this something that you still want to spear head?

@alan-agius4 - yes, absolutely. We have a custom `extract-i18n` builder which splits translations per libs/projects and when building such applications the output is trashed with warnings about duplicate tranaslations. So, I am very much interested/motivated to "fix" it. And the classic part now, "just have not found enough time during last week" :)

packages/angular_devkit/build_angular/src/utils/i18n-options.ts
```diff
@@ -308,9 +310,18 @@ export function loadTranslations(
...
         if (desc.translation[id] !== undefined) {
-          logger.warn(
-            `WARNING [${file.path}]: Duplicate translations for message '${id}' when merging`,
-          );
+          const duplicateTranslationMessage = `[${file.path}]: Duplicate translations for message '${id}' when merging`;
+          switch (duplicateTranslation) {
+            case I18NTranslation.Ignore:
+              break;
+            case I18NTranslation.Error:
+              logger.error(duplicateTranslationMessage);
```
alan-agius4, line 318:
```suggestion
              logger.error(`ERROR ${duplicateTranslationMessage}`);```

File: packages/angular_devkit/build_angular/src/utils/i18n-options.ts [hunk 3]
```diff
@@ -308,9 +310,18 @@ export function loadTranslations(
       // Merge translations
       for (const [id, message] of Object.entries(loadResult.translations)) {
         if (desc.translation[id] !== undefined) {
-          logger.warn(
-            `WARNING [${file.path}]: Duplicate translations for message '${id}' when merging`,
-          );
+          const duplicateTranslationMessage = `[${file.path}]: Duplicate translations for message '${id}' when merging`;
```
alan-agius4, line 313: NIT: 
```suggestion
          const duplicateTranslationMessage = `[${file.path}]: Duplicate translations for message '${id}' when merging.`;
```

packages/angular_devkit/build_angular/src/utils/i18n-options.ts
alan-agius4, line 318:
```suggestion
              logger.error(`ERROR ${duplicateTranslationMessage}`);
```

alan-agius4 (changes requested):
Thanks for this!!! A couple of small NITS. 

Can you please add a test? Maybe amend https://github.com/angular/angular-cli/blob/master/tests/legacy-cli/e2e/tests/i18n/ivy-localize-merging.ts? and amend the commit message to include `Closes #22201` as the footer?

alan-agius4 (approved):
LGTM, thanks

PR #22210, opened 2021-11-22
feat(@angular-devkit/build-angular): set `dir` attribute when using localization

### LINKED ISSUE #16047: i18n-locale should inject dir attribute
# 🚀 Feature request
Add dir="rtl" for rtl locales

```
- [ ] new
- [ x ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description
Angular pages are always LTR, even when specifying an rtl language in `i18n-locale` parameter

When running build with `i18n-locale` set to an RTL language such as Hebrew or Arabic, add `dir="rtl"` to html or body tags in index.html.
I know that the cli can modify the html, because `base-href` parameter does it

Manually setting the dir tag, but then what happens if I want to support different build configs for rtl, and ltr locales? I will have to manually change the dir tag before every build.

alan-agius4:
We add the `dir` (direction) HTML attribute when using localization.

Closes #16047

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
@@ -7,6 +7,7 @@
  */
 
 import { createHash } from 'crypto';
+import { loadEsmModule } from '../load-esm';
 import { htmlRewritingStream } from './html-rewriting-stream';
 
 export type LoadOutputFileFunctionType = (file: string) => Promise<string>;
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
```diff
@@ -207,3 +221,19 @@ function updateAttribute(
...
+      await loadEsmModule<typeof import('@angular/common/locales/en')>(
+        `@angular/common/locales/${lang}`,
+      )
+    ).default;
+
+    const dir = localeData[localeData.length - 2];
+
+    return isString(dir) ? dir : undefined;
+  } catch {
+    warnings.push(`Locale data for '${lang}' cannot be found.`);
```
clydin, line 237: nit: Maybe mention in the warning that the `dir` attribute cannot be set.

clydin (approved):

Issue #21967, opened 2021-10-17
Tests with code coverage fail when using Angular 12.2.10


hosuaby:
# 🐞 Bug report

### Command (mark with an `x`)
- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

While running tests with code coverage `ng test --code-coverage=true` Karma fails with error.

Version `12.2.9` doesn't have this problem.

## 🔬 Minimal Reproduction

`ng test --code-coverage=true` 

## 🔥 Exception or Error

```
./src/app/timesheet/services/timesheet.service.ts - Error: Module build failed (from ./node_modules/@jsdevtools/coverage-istanbul-loader/lib/index.js):
Error: /home/hosuaby/projects/my-timesheet/src/app/timesheet/services/timesheet.service.ts: don't know how to turn this value into a node
    at valueToNode (/home/hosuaby/projects/my-timesheet/node_modules/@babel/types/lib/converters/valueToNode.js:98:9)
    at Object.valueToNode (/home/hosuaby/projects/my-timesheet/node_modules/@babel/types/lib/converters/valueToNode.js:92:58)
    at Object.exit (/home/hosuaby/projects/my-timesheet/node_modules/istanbul-lib-instrument/dist/visitor.js:641:30)
    at PluginPass.exit (/home/hosuaby/projects/my-timesheet/node_modules/istanbul-lib-instrument/dist/instrumenter.js:104:29)
    at newFn (/home/hosuaby/projects/my-timesheet/node_modules/@babel/traverse/lib/visitors.js:177:21)
    at NodePath._call (/home/hosuaby/projects/my-timesheet/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/home/hosuaby/projects/my-timesheet/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/home/hosuaby/projects/my-timesheet/node_modules/@babel/traverse/lib/path/context.js:99:8)
    at TraversalContext.visitQueue (/home/hosuaby/projects/my-timesheet/node_modules/@babel/traverse/lib/context.js:103:16)
[...]

PR #22220, opened 2021-11-23
fix(@angular-devkit/build-angular): transform remapped sourcemap into a plain object

divdavem:
Hello,

I am opening this PR to fix one of the problems we noticed when upgrading to Angular 13 in [ng-bootstrap](https://ng-bootstrap.github.io/) (cf https://github.com/ng-bootstrap/ng-bootstrap/issues/4178)

We are using a custom webpack configuration with `@angular-builders/custom-webpack:browser` to add `coverage-istanbul-loader` but it fails with the `don't know how to turn this value into a node` error.

This similar to #21967 which was previously fixed in commit da1733cc69eeac1417e47df23389da1658850ada but the fix was lost as part of commit 0c44ab305836abab26dc81d7c962f330898213fd.
Cf also https://github.com/istanbuljs/istanbuljs/issues/646

`remapping` returns a SourceMap object and not a plain object. This causes Babel to fail with `don't know how to turn this value into a node` when invoked from `istanbul-lib-instrument` as Babel checks if the value is a plain object.

See: https://github.com/babel/babel/blob/780aa48d2a34dc55f556843074b6aed45e7eabeb/packages/babel-types/src/converters/valueToNode.ts#L115-L130

### DIFF
--- a/packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
+++ b/packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
@@ -239,10 +239,16 @@ export default custom<ApplicationPresetOptions>(() => {
         // `@ampproject/remapping` source map objects but both are compatible with Webpack.
         // This method for merging is used because it provides more accurate output
         // and is faster while using less memory.
-        result.map = remapping(
-          [result.map as SourceMapInput, inputSourceMap as SourceMapInput],
-          () => null,
-        ) as typeof result.map;
+        result.map = {
+          // Convert the SourceMap back to simple plain object.
+          // This is needed because otherwise code-coverage will fail with `don't know how to turn this value into a node`
+          // Which is thrown by Babel if it is invoked again from `istanbul-lib-instrument`.
+          // https://github.com/babel/babel/blob/780aa48d2a34dc55f556843074b6aed45e7eabeb/packages/babel-types/src/converters/valueToNode.ts#L115-L130
+          ...(remapping(
+            [result.map as SourceMapInput, inputSourceMap as SourceMapInput],
+            () => null,
+          ) as typeof result.map),
+        };
       }
 
       return result;


alan-agius4:
This was intentionally removed.

Now in the CLI we only do a single Babel pass. Previously, this workaround was needed because of the usage of `coverage-istanbul-loader` which caused additional Babel pass. 

The usage of this webpack loader has been removed in v13 and instead we use the official Istanbul Babel plugin https://github.com/istanbuljs/babel-plugin-istanbul which also has better source map fidelity and caches instrumented code which causes faster testing times.

Can you share a reproduction on how you are experiencing the error without the mentioned loader?

divdavem:
@alan-agius4 Thank you for your answer. I actually updated the description of the issue which should answer your question:

> We are using a custom webpack configuration with `@angular-builders/custom-webpack:browser` to add `coverage-istanbul-loader` but it fails with the `don't know how to turn this value into a node` error.

cf https://github.com/ng-bootstrap/ng-bootstrap/blob/master/e2e-app/coverage.webpack.js
Note that it is currently temporarily not used anymore (because of this error) cf this commit: https://github.com/ng-bootstrap/ng-bootstrap/commit/9d6e62b548a7bce799f604188a4502163a280e50

divdavem:
Note that it would be good actually to be able to enable instrumentation from an option in `angular.json` and then benefit from the improvements you did in Angular 13 (single pass transformation and improved source map fidelity) for our e2e test coverage use case (thus avoiding the use of our custom webpack configuration).

divdavem:
So the issue can be reproduced with:

```sh
git clone https://github.com/ng-bootstrap/ng-bootstrap.git
cd ng-bootstrap
# the current head commit in the master branch should already be 37c84a6241dc2465c47dc2b357c6f07313569b99 anyway
git checkout 37c84a6241dc2465c47dc2b357c6f07313569b99
git revert 9d6e62b548a7bce799f604188a4502163a280e50
yarn
ng serve e2e-app -c playwright
```

alan-agius4 (approved):
LGTM

PR #22245, opened 2021-11-29
Update index_spec.ts

kayaramazan:
letter error fixed


packages/angular_devkit/build_angular/src/builders/browser/specs/index_spec.ts
```diff
@@ -44,7 +44,7 @@ describe('Browser Builder index HTML processing', () => {
   });
 
   // todo: enable when utf16 is supported
-  xit('works with UTF16 LE BOM', async () => {
```
alan-agius4, line 47: As per comment above this is on purpose to disable the test.

PR #22247, opened 2021-11-29
fix(@ngtools/webpack): JIT mode CommonJS accessing inexistent `default` property

### LINKED ISSUE #21588: @angular-devkit/build-angular 12.2.1 unit test failures - TypeError: Cannot read property 'match' of undefined
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

No, this was introduced after upgrading from 12.1.4 to 12.2.1

### Description

After updating unit test will fail:

```
Failed: Cannot read property 'match' of undefined
        TypeError: Cannot read property 'match' of undefined
            at extractCommentsWithHash (node_modules/@angular/compiler/fesm2015/compiler.js:9555:1)
            at ShadowCss.shimCssText (node_modules/@angular/compiler/fesm2015/compiler.js:9118:1)
            at node_modules/@angular/compiler/fesm2015/compiler.js:22336:1
            at Array.map (<anonymous>)
            at compileStyles (node_modules/@angular/compiler/fesm2015/compiler.js:22335:1)
            at compileComponentFromMetadata (node_modules/@angular/compiler/fesm2015/compiler.js:21882:1)
            at CompilerFacadeImpl.compileComponentFromMeta (node_modules/@angular/compiler/fesm2015/compiler.js:22492:1)
            at CompilerFacadeImpl.compileComponent (node_modules/@angular/compiler/fesm2015/compiler.js:22482:1)
            at Function.get (node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js:27388:1)
            at getComponentDef (node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js:1108:1)
```

## 🌍 Your Environment

```
Angular CLI: 12.2.1
Node: 14.16.1
Package Manager: npm 7.12.0
OS: darwin x64

Angular: 12.2.1
... service-worker

Package                                    Version
--------------------------------------------------------------------
@angular-devkit/architect                  0.1202.1
@angular-devkit/build-angular              12.2.1
@angular-devkit/core                       12.2.1
@angular-devkit/schematics                 12.1.4
@nguniversal/express-engine                12.1.0
@nguniversal/module-map-ngfactory-loader   9.0.0-next.9
@schematics/angular                        12.1.4
rxjs                                       6.6.7
typescript                                 4.3.5
```

alan-agius4:
With this change we fix an issue in JIT mode were we try to access an inexistent `default` property of styles modules when emitted in CJS.

Also, we now emit templates modules without the default property when not targeting ES2015 or greater to normalize the how to `require` templates and styles.

Closes #21588

### DIFF
--- a/packages/ngtools/webpack/src/transformers/replace_resources.ts
+++ b/packages/ngtools/webpack/src/transformers/replace_resources.ts
@@ -146,7 +146,7 @@ function visitComponentMetadata(
   styleReplacements: ts.Expression[],
   directTemplateLoading: boolean,
   resourceImportDeclarations: ts.ImportDeclaration[],
-  moduleKind?: ts.ModuleKind,
+  moduleKind: ts.ModuleKind = ts.ModuleKind.ES2015,
   inlineStyleFileExtension?: string,
 ): ts.ObjectLiteralElementLike | undefined {
   if (!ts.isPropertyAssignment(node) || ts.isComputedPropertyName(node.name)) {
@@ -159,9 +159,10 @@ function visitComponentMetadata(
       return undefined;
 
     case 'templateUrl':
+      const loaderOptions = moduleKind < ts.ModuleKind.ES2015 ? '?esModule=false' : '';
       const url = getResourceUrl(
         node.initializer,
-        directTemplateLoading ? `!${DirectAngularResourceLoaderPath}!` : '',
+        directTemplateLoading ? `!${DirectAngularResourceLoaderPath}${loaderOptions}!` : '',
       );
       if (!url) {
         return node;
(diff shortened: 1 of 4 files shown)


packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -159,9 +159,10 @@ function visitComponentMetadata(
       return undefined;
 
     case 'templateUrl':
+      const loaderOptions = moduleKind < ts.ModuleKind.ES2015 ? '?esModule=false' : '';
```
dgp1130, line 162: Is `ts.ModuleKind` intended to be comparable like this? I'm wondering if there is a way this could break with future changes to `ts.ModuleKind`? I guess any changes will (probably) come > ES2015 and it is unlikely (though not impossible) that those could would be incompatible with ESM.

I'm wondering if it might be better to explicitly call out the specific modules where we want to use `?esModule=false`? I don't feel too strongly about this, I just noticed the `<` comparison on an enum which made me wonder if that could bite us later.

dgp1130 (approved):

packages/ngtools/webpack/src/transformers/replace_resources.ts
alan-agius4: The check is already present in https://github.com/angular/angular-cli/blob/5305e00cb8082dfca8364debf2091097e0f1f7e9/packages/ngtools/webpack/src/transformers/replace_resources.ts#L263.

If TypeScript changes the enum values, I think we can revisit this, considering we don’t support future TypeScript versions.

dgp1130: SGTM, no need to make this more complicated than necessary right now.

alan-agius4: Ps: I don’t feel strongly about not changing it either way.

PR #22252, opened 2021-11-30
Use `asset modules` and `resourceQuery`

### LINKED ISSUE #7245: "Expected 'styles' to be an array of strings." when a stylesheet is referenced in a Component's styleUrls[] and .angular-cli.json's app[0].styles[]
```
- [ ] feature request
```

### Versions.
```
λ ng --version && node -v && npm -v                         
               |___/                                        
@angular/cli: 1.2.6                                         
node: 8.1.2                                                 
os: win32 x64                                               
@angular/animations: 4.3.2                                  
@angular/common: 4.3.2                                      
@angular/compiler: 4.3.2                                    
@angular/core: 4.3.2                                        
@angular/forms: 4.3.2                                       
@angular/http: 4.3.2                                        
@angular/platform-browser: 4.3.2                            
@angular/platform-browser-dynamic: 4.3.2                    
@angular/router: 4.3.2                                      
@angular/cli: 1.2.6                                         
@angular/compiler-cli: 4.3.2                                
@angular/language-service: 4.3.2                            
v8.1.2
5.3.0
```
### Repro steps.
With an SCSS-based project, include a SCSS stylesheet file reference in both a `@Component` decorator's `styleUrls[]` array **and** `.angular-cli.json`'s `app[0].styles[]` array.

Trivial example:

#### app.component.ts
```
@Component({
  selector: 'my-app',
  template: `
    <router-outlet></router-outlet>
    <message-banner *ngIf="showMessage" [messageContainer]="messageContainer"></message-banner>
  `,
  styleUrls: ["./assets/scss/my/styles.scss"],
  providers: [AuthenticationService, AuthenticationGuard, MessagingService]
})
```

concurrently with
#### .angular-cli.json
```
"styles": [
  "./../node_modules/bootstrap/dist/css/bootstrap.min.css",
  "./../node_modules/font-awesome/scss/font-awesome.scss",
  "./assets/scss/my/loadingPage.scss",
  "./assets/scss/my/styles.scss"
],
```

```
Uncaught Error: Expected 'styles' to be an array of strings.
    at assertArrayOfStrings (compiler.es5.js:3920)
    at CompileMetadataResolver.webpackJsonp.../../../compiler/@angular/compiler.es5.js.CompileMetadataResolver.getNonNormalizedDirectiveMetadata (compiler.es5.js:15174)
    at CompileMetadataResolver.webpackJsonp.../../../compiler/@angular/compiler.es5.js.CompileMetadataResolver._getEntryComponentMetadata (compiler.es5.js:15891)
    at compiler.es5.js:15473
    at Array.map (<anonymous>)
    at CompileMetadataResolver.webpackJsonp.../../../compiler/@angular/compiler.es5.js.CompileMetadataResolver.getNgModuleMetadata (compiler.es5.js:15473)
    at JitCompiler.webpackJsonp.../../../compiler/@angular/compiler.es5.js.JitCompiler._loadModules (compiler.es5.js:26799)
    at JitCompiler.webpackJsonp.../../../compiler/@angular/compiler.es5.js.JitCompiler._compileModuleAndComponents (compiler.es5.js:26772)
    at JitCompiler.webpackJsonp.../../../compiler/@angular/compiler.es5.js.JitCompiler.compileModuleAsync (compiler.es5.js:26701)
    at PlatformRef_.webpackJsonp.../../../core/@angular/core.es5.js.PlatformRef_._bootstrapModuleWithZone (core.es5.js:4536)
```

Breaking at `compiler.es5.js#assertArrayOfStrings()` shows the errant entry in `values[]` is `{}`, along with the other stringified styles.

When the duplicate reference is removed from all Component decorators, it appears to work as expected.

### Desired functionality.
Any of:
- That the redundant style would fall out of the tree when shaken and would not produce an error, or
- A more informative error message in the console (though looking at `compiler.es5.ts#assertArrayOfStrings` the cause may be exceedingly difficult to ascertain), or
- Better [documentation on global styling](https://github.com/angular/angular-cli/wiki/stories-global-styles) indicating allowed values, caveats and restrictions. (I'd be happy to submit a PR for that).

This application structure exists because we are porting an old angular 2 beta-6 application built with `gulp` to angular 4.3.2  built with `ng`.

While these errors are almost certainly artifacts of legacy practices, I found it _exceedingly_ difficult to isolate the cause. While `ng` itself may mitigate a large portion of the instances of this issue, it remains [an often-cited error message on SO](https://stackoverflow.com/search?q=Expected+%27styles%27+to+be+an+array+of+strings), each with a different cause.

alan-agius4:
**refactor(@ngtools/webpack): remove direct angular resource loader** 
We remove the custom direct resource loader which is used for JIT when `directTemplateLoading` is enabled. Instead, use Webpack's [asset modules](https://webpack.js.org/guides/asset-modules/) which were introduced in version 5.

To the resource URL, we also add a query parameter, `ngResource`. This is used to be filter request based on a query. See https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax for more information.

**fix(@angular-devkit/build-angular): differentiate components and global styles using file query instead of filename.**
Previously, we introduced the `ngResource` query to Angular component resources we now use it with `resourceQuery` to differentiate between global and components styles, since in some cases while unlikely a file can be used as a component and global style.

Closes #7245


packages/ngtools/webpack/src/resource_loader.ts
```diff
@@ -110,17 +111,23 @@ export class WebpackResourceLoader {
...
-        ? `${containingFile}-${this.outputPathCounter}.${fileExtension}!=!${this.inlineDataLoaderPath}!${containingFile}`
-        : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          `angular-resource:${resourceType},${createHash('md5').update(data!).digest('hex')}`);
+    const getEntry = (): string => {
+      if (filePath) {
+        return `${filePath}?${NG_COMPONENT_RESOURCE_QUERY}`;
+      } else if (resourceType) {
+        return (
+          `${containingFile}-${this.outputPathCounter}.${fileExtension}` +
+          `?${NG_COMPONENT_RESOURCE_QUERY}!=!${this.inlineDataLoaderPath}!${containingFile}`
```
dgp1130, line 120: Nit: These templates are pretty complicated, so I wonder if an example in a comment might make more clear what kind of path we're generating here.

packages/ngtools/webpack/src/resource_loader.ts
```diff
@@ -110,17 +111,23 @@ export class WebpackResourceLoader {
...
+        );
+      } else if (data) {
+        // Create a special URL for reading the resource from memory
+        return `angular-resource:${resourceType},${createHash('md5').update(data).digest('hex')}`;
+      }
 
-    if (!entry) {
       throw new Error(`"filePath" or "data" must be specified.`);
```
dgp1130, line 127: Should we call out that `resourceType` is also sufficient?

packages/ngtools/webpack/src/resource_loader.ts
```diff
@@ -110,17 +111,23 @@ export class WebpackResourceLoader {
...
     }
 
-    // Create a special URL for reading the resource from memory
-    const entry =
-      filePath ||
-      (resourceType
-        ? `${containingFile}-${this.outputPathCounter}.${fileExtension}!=!${this.inlineDataLoaderPath}!${containingFile}`
-        : // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-          `angular-resource:${resourceType},${createHash('md5').update(data!).digest('hex')}`);
+    const getEntry = (): string => {
```
dgp1130, line 114: I like how you put this in a separate function to take advantage of `return`, I think it's the most elegant way of handling this kind of logic.

packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -200,9 +192,11 @@ function visitComponentMetadata(
           if (inlineStyleFileExtension) {
             const data = Buffer.from(node.text).toString('base64');
             const containingFile = node.getSourceFile().fileName;
-            url = `${containingFile}.${inlineStyleFileExtension}!=!${InlineAngularResourceLoaderPath}?data=${encodeURIComponent(
-              data,
-            )}!${containingFile}`;
+            url =
```
dgp1130, line 195: Nit: Similar to before, a comment with an example would be really useful to help communicate what this is trying to do.

packages/angular_devkit/build_angular/src/builders/browser/specs/styles_spec.ts
```diff
@@ -655,4 +655,26 @@ describe('Browser Builder styles', () => {
     result = await browserBuild(architect, host, target, { optimization: true });
     expect(await result.files['styles.css']).toContain('rgba(0,0,0,.15)');
   });
+
+  it('works when using the same css file in `styles` and `stylesUrl`', async () => {
+    host.writeMultipleFiles({
+      'src/styles.css': `
```
dgp1130, line 661: This is used as a global style by default right? Is there some way we can explicitly list this as a global file as part of this test? As is, we're implicitly depending on this default which could change and invalidate the intent of the test but probably leave it still passing.

packages/angular_devkit/build_angular/src/builders/browser/tests/options/named-chunks_spec.ts
```diff
@@ -11,7 +11,7 @@ import { BASE_OPTIONS, BROWSER_BUILDER_INFO, describeBuilder } from '../setup';
 
 const MAIN_OUTPUT = 'dist/main.js';
 const NAMED_LAZY_OUTPUT = 'dist/src_lazy-module_ts.js';
-const UNNAMED_LAZY_OUTPUT = 'dist/8.js';
+const UNNAMED_LAZY_OUTPUT = 'dist/459.js';
```
dgp1130, line 14: Just out of my own curiosity, where does this come from and why does it need to change? Is this some kind of hash of a lazy loaded module which happened to change slightly from this commit?

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -375,6 +375,12 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
           : undefined,
 
       rules: [
+        {
+          test: /\.?(svg|html)$/,
+          // Only process HTML and SVG which are known Angular component resources.
+          resourceQuery: /ngResource/,
```
dgp1130, line 381: Should we include the `?` in this query (`/?ngResource/`)? I'm wondering if this might pick up other files not intended to be resources (like `import ./routingResource.ts`). Is there some processing which makes sure that user imports don't get picked up here?

I guess the Webpack documentation doesn't include the `?`, so I'm not sure if this is a real issue. If this is intended to look like a URL parameter, then it should technically also accept `&ngResource`, but I'm guessing that's not a real problem?

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -377,14 +377,14 @@ export function getStylesConfig(wco: WebpackConfigOptions): Configuration {
             oneOf: [
               // Component styles are all styles except defined global styles
               {
-                exclude: globalStylePaths,
                 use: componentStyleLoaders,
+                resourceQuery: /ngResource/,
```
dgp1130, line 381: Same comment here and below about whether we should consider `/?ngResource/`.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
alan-agius4: I think adding the `?` might be safer.

packages/angular_devkit/build_angular/src/builders/browser/tests/options/named-chunks_spec.ts
alan-agius4: The ID generation is internal to Webpack, I think it changed due to the changes in resource paths due to the additional of the query parameter.

packages/angular_devkit/build_angular/src/builders/browser/specs/styles_spec.ts
alan-agius4: Yes, that is correct.

packages/angular_devkit/build_angular/src/builders/browser/tests/options/named-chunks_spec.ts
dgp1130: Gotcha, thanks for the clarification.

dgp1130 (approved):
LGTM, up to you if you want to wait until January for Charles' input, no rush on this I think.

PR #22258, opened 2021-11-30
build: add pkg_npm rules for remaining packages

kormide:
Before re-writing the `build.ts` script to use bazel for building, I thought a good place to start would be to add some missing rules. Three of the packages built by `build.ts` didn't have corresponding `pkg_npm` targets.

### DIFF
--- a/packages/angular/pwa/BUILD.bazel
+++ b/packages/angular/pwa/BUILD.bazel
@@ -6,6 +6,8 @@
 load("@npm//@bazel/jasmine:index.bzl", "jasmine_node_test")
 load("//tools:defaults.bzl", "ts_library")
 load("//tools:ts_json_schema.bzl", "ts_json_schema")
+load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
+load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
 
 licenses(["notice"])  # MIT
 
@@ -66,4 +68,19 @@ ts_library(
 jasmine_node_test(
     name = "pwa_test",
     srcs = [":pwa_test_lib"],
+)
+
+pkg_npm(
+    name = "npm_package",
+    deps = [
+        ":pwa",
+    ],
+)
+
+pkg_tar(
+    name = "npm_package_archive",
+    srcs = [":npm_package"],
+    extension = "tar.gz",
+    strip_prefix = "./npm_package",
+    tags = ["manual"],
 )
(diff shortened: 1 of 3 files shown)


packages/angular/pwa/BUILD.bazel
```diff
@@ -7,6 +7,11 @@ load("@npm//@bazel/jasmine:index.bzl", "jasmine_node_test")
 load("//tools:defaults.bzl", "ts_library")
 load("//tools:ts_json_schema.bzl", "ts_json_schema")
 
+# @external_begin
+load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
+load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
+# @external_end
```
kormide, line 13: I added these external to follow the existing convention, but I'm not sure what they are used for.

devversion: I was going to ask. It doesn't seem like there is any value in having these as the CLI build files are not exposed publicly anyway.

alan-agius4: These comments were added by @kyliau and were needed for a G3 sync.

devversion: interesting. we should probably re-evaluate (or if @josephperrott already did), whether this is still something we want to maintain long-term. I suspect it will complicate the Bazel migration/maintaince.

josephperrott: This is something we do not need to include as we do not use our BUILD files internally in g3, we have google3 specific ones.

josephperrott (approved):
LGTM, though we can likely remove the `@external...` blocks.

kormide:
> LGTM, though we can likely remove the `@external...` blocks.

I'll remove them.

alan-agius4 (approved):

kormide:
I removed the `@external` tags. I can remove the ones for existing `pkg_npm` rules in futures PRs as I work through the bazel migration.

packages/angular/pwa/BUILD.bazel
```diff
@@ -6,6 +6,8 @@
 load("@npm//@bazel/jasmine:index.bzl", "jasmine_node_test")
 load("//tools:defaults.bzl", "ts_library")
 load("//tools:ts_json_schema.bzl", "ts_json_schema")
+load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
```
devversion, line 9: FYI: the `pkg_tar` rule from `bazel_tools` is deprecated and does not work on Windows. https://docs.bazel.build/versions/main/be/pkg.html

Not necessarily blocking for this PR as it probably is used elsewhere in the repo, but just a heads up.

devversion (approved):
LGTM, but one comment.

packages/angular/pwa/BUILD.bazel
kormide: Noted. I'll look into changing those rules before the bazel built packages actually get used.

packages/angular/pwa/BUILD.bazel
dgp1130: Drive by comment: I'm hoping to reevaluate how we integrate the CLI internally to improve the syncing story so we can focus just on the external side here.

PR #22274, opened 2021-12-02
docs: update ng deploy suggestions

fmalcher:
This PR changes the default output of `ng deploy`.

- 1. remove `@zeit/ng-deploy`. This package seems to be unmaintained, it should therefore not be suggested in the default output. See https://github.com/vercel/ng-deploy-vercel/issues/14 or https://github.com/vercel/ng-deploy-vercel/issues/12
- 2. add the (well-maintained) `angular-cli-ghpages` by @JohannesHoppe and me to the suggestions. This can be discussed ;-)


JohannesHoppe:
+1 for `angular-cli-ghpages`! 😉

alan-agius4:
Once approved by our dev-rel, https://angular.io/guide/deployment should also be updated.

packages/angular/cli/commands/deploy-impl.ts
```diff
@@ -19,7 +19,7 @@ favorite platform.
 For example:
   ng add @angular/fire
   ng add @azure/ng-deploy
-  ng add @zeit/ng-deploy
+  ng add angular-cli-ghpages
```
mgechev, line 22: I love `angular-cli-ghpages` and use it in my personal projects! It gets tricky to feature community modules without ignoring others. Our policy is to only recommend deployment builders that are maintained by the hosting vendor. In this case, if the package was under the `@github/*` scope we'd have been able to merge the PR. For now, I'd suggest just to remove `@zeit/ng-deploy`.

mgechev (changes requested):
Left a comment inline. Thank you for keeping our list of builders up to date!

packages/angular/cli/commands/deploy-impl.ts
fmalcher: I understand this, thanks for elaborating! 🙂 I'll remove the second commit.

alan-agius4 (approved):

PR #22291, opened 2021-12-03
fix(@angular-devkit/build-angular): fallback to use language ID to set the `dir` attribute

### LINKED ISSUE #22285: `dir` attribute is not set for default locale `en-US`
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

No, this is a new feature introduced in v13.1.0-next.3

### Description

Running `ng build --localize` in a brand new CLI app generated with v13.1.0-next.3 outputs the following warning:

```
Locale data for 'en-US' cannot be found. 'dir' attribute will not be set for this locale.
```

## 🔬 Minimal Reproduction

```
npx @angular/cli@13.1.0-next.3 new dir-attribute --defaults
cd dir-attribute
ng add @angular/localize@13.1.0-next.3
ng build --localize
```

## 🔥 Exception or Error

<pre><code>
Locale data for 'en-US' cannot be found. 'dir' attribute will not be set for this locale.
</code></pre>

## 🌍 Your Environment

<pre><code>
Angular CLI: 13.1.0-next.3
Node: 14.18.1
Package Manager: npm 6.14.14
OS: darwin x64

Angular: 13.0.3

@angular-devkit/architect       0.1301.0-next.3
@angular-devkit/build-angular   13.1.0-next.3
@angular-devkit/core            13.1.0-next.3
@angular-devkit/schematics      13.1.0-next.3
@angular/cli                    13.1.0-next.3
@angular/localize               13.1.0-next.3
@schematics/angular             13.1.0-next.3
rxjs                            7.4.0
typescript                      4.4.4
</code></pre>

**Anything else relevant?**

alan-agius4:
In some cases we don't ship certain locales, or they map to files which are named only the language IDs. Example `en-US`, which it's locale data is available from `@angular/common/locales/en.mjs`

Closes #22285

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
@@ -222,11 +222,26 @@ function isString(value: unknown): value is string {
   return typeof value === 'string';
 }
 
-async function getLanguageDirection(lang: string, warnings: string[]): Promise<string | undefined> {
+async function getLanguageDirection(
+  locale: string,
+  warnings: string[],
+): Promise<string | undefined> {
+  const dir = await getLanguageDirectionFromLocales(locale);
+
+  if (!dir) {
+    warnings.push(
+      `Locale data for '${locale}' cannot be found. 'dir' attribute will not be set for this locale.`,
+    );
+  }
+
+  return dir;
+}
+
+async function getLanguageDirectionFromLocales(locale: string): Promise<string | undefined> {
   try {
     const localeData = (
       await loadEsmModule<typeof import('@angular/common/locales/en')>(
-        `@angular/common/locales/${lang}`,
+        `@angular/common/locales/${locale}`,
       )
     ).default;
 
@@ -234,8 +249,13 @@ async function getLanguageDirection(lang: string, warnings: string[]): Promise<s
 
     return isString(dir) ? dir : undefined;
   } catch {
-    warnings.push(
-      `Locale data for '${lang}' cannot be found. 'dir' attribute will not be set for this locale.`,
-    );
+    // In some cases certain locales might map to files which are named only with language id.
+    // Example: `en-US` -> `en`.
+    const [languageId] = locale.split('-', 1);
+    if (languageId !== locale) {
+      return getLanguageDirectionFromLocales(languageId);
+    }
   }
+
+  return undefined;
 }
--- a/packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html_spec.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html_spec.ts
@@ -104,6 +104,24 @@ describe('augment-index-html', () => {
       `);
   });
 
+  it(`should fallback to use language ID to set the dir attribute (en-US)`, async () => {
+    const { content, warnings } = await augmentIndexHtml({
+      ...indexGeneratorOptions,
+      lang: 'en-US',
+    });
+
+    expect(warnings).toHaveSize(0);
+    expect(content).toEqual(oneLineHtml`
+        <html lang="en-US" dir="ltr">
+          <head>
+            <base href="/">
+          </head>
+          <body>
+          </body>
+        </html>
+      `);
+  });
+
   it(`should work when lang (locale) is not provided by '@angular/common'`, async () => {
     const { content, warnings } = await augmentIndexHtml({
       ...indexGeneratorOptions,


cexbrayat (approved):

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
```diff
@@ -222,20 +222,40 @@ function isString(value: unknown): value is string {
...
       await loadEsmModule<typeof import('@angular/common/locales/en')>(
-        `@angular/common/locales/${lang}`,
+        `@angular/common/locales/${locale}`,
       )
     ).default;
 
     const dir = localeData[localeData.length - 2];
 
     return isString(dir) ? dir : undefined;
   } catch {
```
dgp1130, line 251: Have we considered any means of actually determining what the correct locale is for looking up the dir attribute? Is there any pattern to which locales have dir attributes that we could exploit or should we consider updating locale data to be more consistent or address this use case?

Maybe it's not practical right now, and I think this approach of "just try the base locale" is probably fine, I just want to make sure we've at least considered a direct approach beyond trial-and-error.

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
```diff
@@ -222,20 +222,40 @@ function isString(value: unknown): value is string {
...
-    warnings.push(
-      `Locale data for '${lang}' cannot be found. 'dir' attribute will not be set for this locale.`,
-    );
+    // In some cases certain locales might map to files which are named only with language id.
+    // Example: `en-US` -> `en`.
+    const [languageId] = locale.split('-', 1);
```
dgp1130, line 254: Will the dir attribute always be on the primary subtag? I always get confused with the localization standard, so I'm sure what applies here, but I think it is possible for [multiple dashes to be included](https://en.wikipedia.org/wiki/IETF_language_tag#:~:text=Up%20to%20three%20optional%20extended%20language%20subtags%20composed%20of%20three%20letters%20each%2C%20separated%20by%20hyphens%3B):

> Up to three optional extended language subtags composed of three letters each, separated by hyphens

Is there a possible case where the locale is something like `zh-Hans-SG`, and the dir attribute is set on `zh-Hans` rather than `zh`?

dgp1130 (approved):

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
alan-agius4: The first item is actual the language Id, which is always the first.

We are actually using this methodology in other places https://github.com/angular/angular-cli/blob/033c4415efd6bb4a073018da446643c1abbe3a40/packages/angular_devkit/build_angular/src/utils/i18n-options.ts#L198

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
alan-agius4: I don’t know what’s the reason why for example the default Angular local is `en-US` but we actually don’t ship such a locale, but rather its data is in `en`.

I did ask on slack https://angular-team.slack.com/archives/CDT06KWR2/p1638469098001200

Also, it seems that we are using the logic elsewhere. (Linked in the other comment)

dgp1130: Ah, so it's not just the dir attribute but all the locale data. My understanding of the spec is that each subtag gives more specific info about the locale, so if something is too specific, and we need to drop some subtags to get a piece of information (like `dir`), then that is reasonable fallback behavior.

Do we have a hard restriction on the locales we accept? IIRC, users can add their own, which makes me think there is a case where support a user adds some custom `xx` locale and then uses `xx-YY` and expects it to fall back to `xx`, and would need to have some kind of fallback mechanism like this regardless of what we do in `@angular/common/locales`?

alan-agius4: >. Do we have a hard restriction on the locales we accept?

No, users can create their own locales. But the CLI will not try to load the data for such locales. We only load data from `@angular/common/locales`.

dgp1130: In that case, it might be worth following up to see if there are any improvements we can make to `@angular/common/locales` to not need this kind of logic, but there might be a good reason for why it works that way today (maybe `en` is equivalent to `en-US` so it would just be a waste of bytes?). Regardless I don't think it needs to block this PR, but just a conversation that might be worth having.

alan-agius4: It does appear that en is equivalent to en-US.

https://angular-team.slack.com/archives/CDT06KWR2/p1638469098001200

PR #22292, opened 2021-12-03
refactor(@angular-devkit/build-angular): remove `NG_BUILD_PROFILING` and `DEVKIT_PROFILING` logic

### LINKED ISSUE #20336: Document NG_BUILD_PROFILING
I feel like I've discovered some great secret, and I'd like other to know about it.

Environment var NG_BUILD_PROFILING=1 tells you where the precious seconds of your life are going.

alan-agius4:
Generating a CPU profile using Node.JS `--cpu-prof` or Chrome inspector is preferred.

Closes #20336


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -133,14 +132,6 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
     }
   }
 
-  if (profilingEnabled) {
-    extraPlugins.push(
-      new debug.ProfilingPlugin({
```
dgp1130, line 138: Nit: I'm guessing there's no way to drop a dependency on this `ProfilingPlugin` at the NPM layer? Does it come directly from Webpack in a way we can't separate it?

dgp1130 (approved):
Thanks for cleaning this up! Glad we can simplify our infra and support here.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
alan-agius4: That is correct.

PR #22294, opened 2021-12-03
fix(@angular-devkit/build-angular): Sass compilation in StackBlitz webcontainers

alan-agius4:
When `process.versions.webcontainer` is a string it means that we are running in a StackBlitz webcontainer. `SassWorkerImplementation` uses `receiveMessageOnPort` Node.js `worker_thread` API to ensure sync behavior which is ~2x faster. However, it is non trivial to support this in a webcontainer and while slower we choose to use `dart-sass` which in Webpack uses the slower async path.

// cc @markwhitfeld

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -107,14 +107,16 @@ export function getStylesConfig(wco: WebpackConfigOptions): Configuration {
     );
   }
 
-  const sassImplementation = new SassWorkerImplementation();
-  extraPlugins.push({
-    apply(compiler) {
-      compiler.hooks.shutdown.tap('sass-worker', () => {
-        sassImplementation?.close();
-      });
-    },
-  });
+  const sassImplementation = getSassImplementation();
+  if (sassImplementation instanceof SassWorkerImplementation) {
+    extraPlugins.push({
+      apply(compiler) {
+        compiler.hooks.shutdown.tap('sass-worker', () => {
+          sassImplementation?.close();
+        });
+      },
+    });
+  }
 
   const assetNameTemplate = assetNameTemplateFactory(hashFormat);
 
@@ -403,4 +405,15 @@ export function getStylesConfig(wco: WebpackConfigOptions): Configuration {
     },
     plugins: extraPlugins,
   };
+}
+
+function getSassImplementation(): SassWorkerImplementation | typeof import('sass') {
+  const { webcontainer } = process.versions as unknown as Record<string, unknown>;
+
+  // When `webcontainer` is a truthy it means that we are running in a StackBlitz webcontainer.
+  // `SassWorkerImplementation` uses `receiveMessageOnPort` Node.js `worker_thread` API to ensure sync behavior which is ~2x faster.
+  // However, it is non trivial to support this in a webcontainer and while slower we choose to use `dart-sass`
+  // which in Webpack uses the slower async path.
+  // We should periodically check with StackBlitz folks (Mark Whitfeld / Dominic Elm) to determine if this workaround is still needed.
+  return webcontainer ? require('sass') : new SassWorkerImplementation();
 }


markwhitfeld (approved):
That check looks good 👍

d3lm:
Thanks so much for this quick implementation and for adding an env check here for the time being. I am actually gonna look into this very soon so we can support `receiveMessageOnPort` at least minimally. Thanks a lot!

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -404,3 +406,14 @@ export function getStylesConfig(wco: WebpackConfigOptions): Configuration {
...
+
+function getSassImplementation(): SassWorkerImplementation | typeof import('sass') {
+  const { webcontainer } = process.versions as unknown as Record<string, unknown>;
+
+  // When `webcontainer` is a string it means that we are running in a StackBlitz webcontainer.
+  // `SassWorkerImplementation` uses `receiveMessageOnPort` Node.js `worker_thread` API to ensure sync behavior which is ~2x faster.
+  // However, it is non trivial to support this in a webcontainer and while slower we choose to use `dart-sass`
+  // which in Webpack uses the slower async path.
+  // We should periodically check with StackBlitz folks (Mark Whitfeld / Dominic Elm) to determine if this workaround is still needed.
+  return typeof webcontainer === 'string' ? require('sass') : new SassWorkerImplementation();
```
dgp1130, line 418: Nit: Do we care that `webcontainer` is a `string`? Would it be safer/cleaner to just check if the property exists / is truthy?

dgp1130 (approved):
Just for context, do we only use `dart-sass` for webcontainers, or are there other usages were it is still supported? I'm just wondering if this is the major thing blocking us from moving to `node-sass`, or if there are other problems that need to be resolved anyways.

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
alan-agius4: I think we can check if it’s truthy. But  stackblitz folks suggested this way.

@markwhitfeld thoughts?

alan-agius4:
> Just for context, do we only use `dart-sass` for webcontainers, or are there other usages were it is still supported? I'm just wondering if this is the major thing blocking us from moving to `node-sass`, or if there are other problems that need to be resolved anyways.

We only support dart-sass.

dgp1130:
> We only support dart-sass.

Ah thanks, I was getting the direction of `node-sass` -> `dart-sass` backwards. Disabling the worker in webcontainers is reasonable until it has the necessary support.

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
d3lm: The value will be a commit sha and it should be a string yes. I think either works.

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -404,3 +406,14 @@ export function getStylesConfig(wco: WebpackConfigOptions): Configuration {
...
+
+function getSassImplementation(): SassWorkerImplementation | typeof import('sass') {
+  const { webcontainer } = process.versions as unknown as Record<string, unknown>;
+
+  // When `webcontainer` is a truthy it means that we are running in a StackBlitz webcontainer.
+  // `SassWorkerImplementation` uses `receiveMessageOnPort` Node.js `worker_thread` API to ensure sync behavior which is ~2x faster.
+  // However, it is non trivial to support this in a webcontainer and while slower we choose to use `dart-sass`
+  // which in Webpack uses the slower async path.
+  // We should periodically check with StackBlitz folks (Mark Whitfeld / Dominic Elm) to determine if this workaround is still needed.
+  return webcontainer ? require('sass') : new SassWorkerImplementation();
```
d3lm, line 418: Maybe we can make this more explicit and say `webcontainer != null`, WDYT?

PR #22297, opened 2021-12-03
refactor(@ngtools/webpack): use webpack API to delete assets in compilation

alan-agius4:


packages/ngtools/webpack/src/resource_loader.ts
```diff
@@ -204,8 +204,11 @@ export class WebpackResourceLoader {
         () => {
           finalContent = childCompilation.assets[outputFilePath]?.source().toString();
 
-          delete childCompilation.assets[outputFilePath];
-          delete childCompilation.assets[outputFilePath + '.map'];
+          for (const { files } of childCompilation.chunks) {
+            for (const file of files) {
+              childCompilation.deleteAsset(file);
```
alan-agius4, line 209: Yes, we want to delete all chunk assets

dgp1130 (approved):

PR #22305, opened 2021-12-06
fix(@angular-devkit/build-angular): prefer ES2015 entrypoints when application targets ES2019 or lower

### LINKED ISSUE #22270: Some unsupported code with es2015 target set

common

### Is this a regression?

No

### Description

image
This is not supported for es2015. But it's still there after compile. I'm trying to write a webpage for OBS studio. The ?. is not supported I think. And I think it's not in es2015 as well.
image
Here is the tsconfig. Set target to es2015 doesn't help. But the unsupported is gone if I set target to es5.
https://github.com/angular/angular/blob/b5ab7aff433a67cddaa55e621d17b1a1b07b57c2/packages/common/src/location/location_strategy.ts#L176-L178
Why this ?. is kept even I set target to es2015?

_No response_

_No response_

_No response_

### Anything else?

_No response_

alan-agius4:
Previously, we always consumed the ES2020 entrypoints, which caused issues in environments where the application compilation target is ES2019 or lower and ES2020 is not supported.

This is because we only downlevel code when we target ES5 or below.

- ES5 or below compilations, ES2015 entrypoints are used and their code is downlevelled to ES5.
- ES2019 or below, ES2015 entrypoints are used and no downlevelling is involved.
- ES2020 or later, ES2020 entrypoints are used.

Closes #22270

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser/specs/lazy-module_spec.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser/specs/lazy-module_spec.ts
@@ -152,7 +152,7 @@ describe('Browser Builder lazy modules', () => {
     const { files } = await browserBuild(architect, host, target);
     expect(files['src_one_ts.js']).not.toBeUndefined();
     expect(files['src_two_ts.js']).not.toBeUndefined();
-    expect(files['default-node_modules_angular_common_fesm2020_http_mjs.js']).toBeDefined();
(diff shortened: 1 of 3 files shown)


destus90:
@alan-agius4 
Is there a typo in the commit description? 
> no downloading is involved.

alan-agius4:
Yeah, that should be downlevelling

dgp1130 (approved):

Issue #13505, opened 2019-01-23
ng-add easy support for libraries


rjlopezdev:
# 🚀 Feature request

### Command (mark with an `x`)

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description

It is very common that when we generate a library, we want to add `ng-add` support, and it is tedious to have to configure it (in addition, __there is no or I have not found documentation__ to carry out the process).

### Describe the solution you'd like

Add a flag `--ng-add` to generate a library with a schematics project (with collection.json) and a default scehamtic `ng-add`.

__Example:__

`ng generate library <lib-name> --ng-add` : it generates a new library with a configuration described in __Describe the solution you'd like__

### Describe alternatives you've considered

If this is not possible, add a __step-by-step guide__ explaining how to configure it to the official documentation.

alan-agius4:
Also relates to https://github.com/ng-packagr/ng-packagr/issues/1180

mgechev:
@alan-agius4 should we keep track of this issue in `ng-packagr` or the CLI is the right place?

alan-agius4:
@mgechev, if eventually ng-packagr supports this, we'd need to do some work in the schematics.

We can either;
1) Close this and when `ng-packagr` supports this feature (if it will) we can implement the schematics from our end.
2) Not dependent on `ng-packagr` to do this. As schematics transpilation can be done outside of ng-packagr.
3) keep this for reference.

mgechev:
Thanks fort he quick follow up! Let's keep it.

alan-agius4:
Hi @rjlopezdev, 

Thanks for this feature request. However, currently this is out of scope of Angular CLI. 

Let's revisit this in the future.

PR #22311, opened 2021-12-07
feat(@schematics/angular): add a new schematic command

### LINKED ISSUE #22288: Generate and manage schematic configuration files
# 🚀 Feature request

- [X] generate

### Description

It came up in an Angular Nation discussion today that making a new schematic is really difficult and requires a lot of steps with complex documentation. I think this came up before but I wasn't able to find an existing issue quickly.

Ideally we would have something like `ng generate schematic` (or `schematics new`) which could create a new schematic. We'll probably have to workshop how to support both creating a new collection and adding to an existing collection. This would also help improve [some of the documentation](https://angular.io/guide/schematics-authoring), since users can get started with schematics much easily.

Alternative is to just document everything and make users type everything out by hand, which is how it works today.

/cc @kroeder.

kroeder:
# Introduction

This PR adds a new command `ng generate schematic` to the CLI. The primary goal is to automate the initial setup and support the creation of new schematics by letting the CLI update the corresponding JSON config files (migration.json, collection.json).

See https://angular.io/guide/schematics-for-libraries

## Current situation

Right now, developers can already set up schematics outside of an existing Angular workspace:

```
// see https://angular.io/guide/schematics-authoring#schematics-cli
npm i -g @angular-devkit/schematics-cli
schematics blank --name=hello-world
```

The command above creates a new, independent, TypeScript project with a basic schematic and a collection.json. 

```
hello-world/
├── node_modules
├── src/
│   ├── hello-world/
│   │   ├── index_spec.ts
│   │   └── index.ts
│   └── collection.json
├── .gitignore
├── .npmignore
├── package-lock.json
├── package.json
├── README.md
└── tsconfig.json
```

The most significant pain point is that it does not integrate into an existing Angular workspace which gives developers two choices:

1) Copy and paste things over to their project
2) Use a manual guide https://angular.io/guide/schematics-for-libraries

# What is this PR trying to solve?

## Goals
- Extend the Angular CLI with a new command "ng generate schematic <options>"
- Automate the initial setup for Angular workspaces
  - Install necessary dependencies
  - Create a schematics directory
  - Create a collection.json or migration.json and add their path to the package.json of the project
  - Add necessary manual build steps (https://angular.io/guide/schematics-for-libraries#building-your-schematics)
- Provide schematics for
  - Adding ng-add support (ng generate schematic --type add <...>)
  - Creating ng-generate schematics (ng generate schematic --type generate <...>)
  - Creating ng-update schematics (ng generate schematic --type update <...>)

## Non-Goals
- Add a builder that handles all necessary build steps
- Extending the existing "@angular-devkit/schematics-cli" API

# Related issues 

https://github.com/ng-packagr/ng-packagr/issues/1180 
https://github.com/angular/angular-cli/issues/13505

Closes #22288

### DIFF
--- a/packages/schematics/angular/collection.json
+++ b/packages/schematics/angular/collection.json
@@ -118,6 +118,11 @@
       "factory": "./web-worker",
       "schema": "./web-worker/schema.json",
       "description": "Create a Web Worker."
+    },
+    "schematic": {
+      "factory": "./schematic",
+      "schema": "./schematic/schema.json",
+      "description": "Create a schematic"
     }
   }
 }
--- a/packages/schematics/angular/schematic/files/__name@dasherize@if-flat__/__name@dasherize__.spec.ts.template
+++ b/packages/schematics/angular/schematic/files/__name@dasherize@if-flat__/__name@dasherize__.spec.ts.template
@@ -1,0 +1,21 @@
+import { Tree } from '@angular-devkit/schematics';
+import { SchematicTestRunner } from '@angular-devkit/schematics/testing';
+import * as path from 'path';
+
+const collectionPath = path.join(__dirname, '<%= collectionJsonPath %>');
+
+describe('<%= dasherize(name) %>', () => {
+  let runner: SchematicTestRunner;
+  let tree: Tree;
+
+  beforeEach(() => {
+    runner = new SchematicTestRunner('schematics', collectionPath);
+    tree = Tree.empty();
+  });
+
+  // todo add proper default test
+  it('noop', async () => {
+    await runner.runSchematicAsync('<%= dasherize(name) %>', {}, tree).toPromise();
+    expect(true).toBe(true);
+  });
+});
--- a/packages/schematics/angular/schematic/files/__name@dasherize@if-flat__/__name@dasherize__.ts.template
+++ b/packages/schematics/angular/schematic/files/__name@dasherize@if-flat__/__name@dasherize__.ts.template
@@ -1,0 +1,11 @@
+import { noop, Tree } from "@angular-devkit/schematics";
+
+export interface <%= classify(name) %>Options {
+  name: string;
+}
+
+export function <%= camelize(name) %>(options: <%= classify(name) %>Options) {
+  return async (host: Tree) => {
+    return noop();
+  };
+}
--- a/packages/schematics/angular/schematic/files/__name@dasherize@if-flat__/schema.json.template
+++ b/packages/schematics/angular/schematic/files/__name@dasherize@if-flat__/schema.json.template
@@ -1,0 +1,19 @@
+{
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Name",
+      "$default": {
+        "$source": "argv",
+        "index": 0
+      }
+    },
+    "path": {
+      "type": "string",
+      "format": "path",
+      "description": "",
+      "visible": false
+    }
+  },
+  "required": ["name"]
+}
--- a/packages/schematics/angular/schematic/index.ts
+++ b/packages/schematics/angular/schematic/index.ts
@@ -1,0 +1,54 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { strings } from '@angular-devkit/core';
+import { apply, applyTemplates, chain, FileOperator, filter, forEach, mergeWith, move, noop, Rule, SchematicsException, Tree, url } from '@angular-devkit/schematics';
+import { parseName } from '../utility/parse-name';
+import { buildDefaultPath, getWorkspace } from '../utility/workspace';
+import { Schema as ComponentOptions } from './schema';
+
+export default function (options: ComponentOptions): Rule {
+  return async (host: Tree) => {
+    const workspace = await getWorkspace(host);
+    const project = workspace.projects.get(options.project as string);
+
+    if (!project) {
+      throw new SchematicsException(`Project "${options.project}" does not exist.`);
+    }
+
+    if (options.path === undefined) {
+      options.path = buildDefaultPath(project);
+    }
+
+    const parsedPath = parseName(options.path as string, options.name);
+    options.name = parsedPath.name;
+    options.path = parsedPath.path;
+
+    const templateSource = apply(url('./files'), [
+      options.skipTests ? filter((path) => !path.endsWith('.spec.ts.template')) : noop(),
+      applyTemplates({
+        ...strings,
+        'if-flat': (s: string) => (options.flat ? '' : s),
+        ...options,
+      }),
+      !options.type
+        ? forEach(((file) => {
+            return file.path.includes('..')
+              ? {
+                  content: file.content,
+                  path: file.path.replace('..', '.'),
+                }
+              : file;
+          }) as FileOperator)
+        : noop(),
+      move(parsedPath.path),
+    ]);
+
+    return chain([mergeWith(templateSource)]);
+  };
+}
--- a/packages/schematics/angular/schematic/index_spec.ts
+++ b/packages/schematics/angular/schematic/index_spec.ts
@@ -1,0 +1,128 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { SchematicTestRunner, UnitTestTree } from '@angular-devkit/schematics/testing';
+import { Schema as ApplicationOptions, Style as AppStyle } from '../application/schema';
+import { Schema as WorkspaceOptions } from '../workspace/schema';
+import { Schema as SchematicOptions, Type } from './schema';
+
+describe('Ng Schematic Schematic', () => {
+  const schematicRunner = new SchematicTestRunner(
+    '@schematics/angular',
+    require.resolve('../collection.json'),
+  );
+  const defaultOptions: SchematicOptions = {
+    name: 'foo',
+    // path: 'src/app',
+    type: 'Generate' as Type,
+    skipTests: false,
+    project: 'bar',
+  };
+
+  const workspaceOptions: WorkspaceOptions = {
+    name: 'workspace',
+    newProjectRoot: 'projects',
+    version: '6.0.0',
+  };
+
+  const appOptions: ApplicationOptions = {
+    name: 'bar',
+    inlineStyle: false,
+    inlineTemplate: false,
+    routing: false,
+    style: AppStyle.Css,
+    skipTests: false,
+    skipPackageJson: false,
+  };
+  let appTree: UnitTestTree;
+  beforeEach(async () => {
+    appTree = await schematicRunner.runSchematicAsync('workspace', workspaceOptions).toPromise();
+    appTree = await schematicRunner
+      .runSchematicAsync('application', appOptions, appTree)
+      .toPromise();
+  });
+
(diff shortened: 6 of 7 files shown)


kroeder:
This is still work-in-progress. I thought I push this early on to also have a chance to ask questions or discuss things.

For instance, how am I able to run the schematic test locally?
https://github.com/angular/angular-cli/pull/22311/files#diff-11baa61e683c71b2842832e554bbf57c6becabad795fae70b6a1aede1cb9622fR1

There is no karma configuration file in this package.

@alan-agius4

alan-agius4 (commented):
HI @kroeder, 

Thanks for your contribution.

I think this needs some more thought and potential this is a good candidate for a design document.

There are a couple of points that we needs to be discussed and eventually see if this is something that we want to support in the long run. On paper this looks straight forward but there are a number of missing parts that are actually more complicated.

There are 3 main use-cases were users want to add a schematic. 
1) In an Angular Ivy library
2) In an Architect builder
3) A standalone schematic maybe to customize a component schematic to the needs or your org.

The first one is built with **ng-packagr** and in an Angular workspace is defined as a project of type library, the latter two are not part of an Angular workspace. With your feature request I am assuming that you want to tackle and add support for the first use-case, is that so?

I also remembered that there was a feature request in **ng-packagr** to be handle schematics out-of-the-box, but it turns out that this isn't a very popular request. In 3 years it got 7 upvotes. (https://github.com/ng-packagr/ng-packagr/issues/1180). In the CLI we also had a similar request https://github.com/angular/angular-cli/issues/13505 which wasn't popular with the community (0 upvotes in 1.5 years).

I think that the main reason for that it is actually hard to write the migration itself. This is  because in many cases you'd need to be familiar with TypeScript APIs and AST (Example: you'd want to add your library in an `ngModule` imports.), or you'd want to do code modes to replace deprecated APIs.

There are also some other things that we need to consider especially in a multi-project workspace. Such as how to test and build a schematic. 

In an Angular workspace the default test builder uses Karma which runs test in a browser. We don't provide an alternate test builder such as Jasmine to test schematics which are intended to run on Node.Js

Also, we don't have a plain TypeScript builder.  We can create the below helper scripts, but this is not ideal especially since in a multi library workspace such scripts will conflict.

```js
"scripts": {
    "build": "../../node_modules/.bin/tsc -p tsconfig.schematics.json",
    "copy:schemas": "cp --parents schematics/*/schema.json ../../dist/my-lib/",
    "copy:files": "cp --parents -p schematics/*/files/** ../../dist/my-lib/",
    "copy:collection": "cp schematics/collection.json ../../dist/my-lib/schematics/collection.json",
    "postbuild": "npm run copy:schemas && npm run copy:files && npm run copy:collection"
  },
```

kroeder:
Hey @alan-agius4 

Yes, I am talking about 1) as proposed by https://github.com/angular/angular-cli/issues/22288#issuecomment-987120513  - I should have added a bit more context to the PR description - I wanted to create the PR initially to have exactly those discussions

I also don't want to cover `ng build` (also proposed in the linked comment above).

I absolutely agree with you. Before I proceed I will write some things down and we can talk about the design goals of this PR - there's a lot to cover like "what happens if you run `ng generate schematic` the first time without having any schematic packages installed etc. 

I give you a ping as soon as I have written something down

kroeder:
@alan-agius4 I updated the original post

I didn't go into details, though.
Can you take a look at it and comment if you agree/disagree with the goals of this PR?

alan-agius4:
@kroeder, thanks for that. Let me go through it later and get back to you.

alan-agius4:
Hi @kroeder, 

Thanks for the updating the description. 

My main concerns are around how will the schematic be tested and build in a multi project scenario. Having NPM scripts is fine when having a single project but in a multi project scenario I my option it gets a bit messy. 

What I noticed is that in the linked documentation, testing is actually not even mentioned Also, the copying of assets can actually be done using `ng-packagr`, instead of the NPM scripts.

There is also another point to consider.  With this approach we'd need to add the schematics dependencies as library dependencies and not just as `devDependencies` in the workspace root `package.json` as otherwise the schematic might fail during runtime because of missing deps.

Overall there are a couple of things that we should iron out. (//cc @dgp1130 / @clydin  for visibility)
- Should we wait for the Schematic runtime to land before proceeding with this?
- Is this something that we really want to do and support right now? There is little interest from this in the community and it's not trivial to provide a great DX right now.
- How will the schematic be tested (Jasmine like the blank schematic generator)?
- Should we actually invest the time into something more valuable to the community and overall less complex(IMHO)? Such as a schematic to generate secondary entry-points which currently need to be done manually and often they are misconfigured https://github.com/angular/angular/issues/44414?

dgp1130:
I put down some initial thoughts below. I still don't fully understand the full implications of this so please correct and push back where I'm overstepping my bounds @alan-agius4. We can chat more on Monday and fully evaluate the context here.

> My main concerns are around how will the schematic be tested and build in a multi project scenario. Having NPM scripts is fine when having a single project but in a multi project scenario I my option it gets a bit messy.

This is a problem today, and I'm not sure that this change makes it any worse. I agree that we should have a better story around this, but I don't know that is strictly blocking for these changes? Could we limit this to single project workspaces or find some equally hacky scripts to generate which are compatible with multi-project workspaces even if not ideal?

> There is also another point to consider. With this approach we'd need to add the schematics dependencies as library dependencies and not just as `devDependencies` in the workspace root `package.json` as otherwise the schematic might fail during runtime because of missing deps.

I believe this is an existing rough edge with schematics, and hopefully the common runtime will improve this. Isn't that how schematics are deployed today? Are there any strong reasons not to add a dep on schematics besides "It's kinda weird and we hope to figure out a better way soon."?

> * Should we wait for the Schematic runtime to land before proceeding with this?

I don't think this is strictly blocked on a common schematics runtime, it's more a matter of what we're comfortable exporting as a public API for library authors to use. I don't think this FR actually changes that API (from a schematics perspective), it's just raising the visibility of the API and possibly getting more community investment in something that we may need to change.

It's a bit of a judgement call, but I don't get the impression that shipping this would make future schematic changes noticeably more difficult to land, considering that there are already libraries that use schematics in the same manner today? Curious how you feel about that point @alan-agius4, is there more I'm not seeing here?

> * Is this something that we really want to do and support right now? There is little interest from this in the community and it's not trivial to provide a great DX right now.
> * Should we actually invest the time into something more valuable to the community and overall less complex(IMHO)? Such as a schematic to generate secondary entry-points which currently need to be done manually and often they are misconfigured [Library compilation fails with   `Unable to write a reference to` compiler cli 13.0.4 angular#44414](https://github.com/angular/angular/issues/44414)?

I agree the interest is minimal atm. My (probably idealistic) feeling is that if we agree this is a feature we would like to add to the CLI and there is a community member willing to do the work (@kroeder), I'd like to do what I can to make that happen. However, I do understand that it's not always that simple. Not all projects can be driven by community members and some times just aren't compatible with current focus of a team, and even then it is still a non-trivial effort to support the implementation even in the best cases.

These are just initial thoughts, we can chat more next week and come up with more coherent reasoning and a clearer decision.

kroeder:
> My main concerns are around how will the schematic be tested and build in a multi project scenario. Having NPM scripts is fine when having a single project but in a multi project scenario I my option it gets a bit messy.

What exactly troubles you with this? I might have overseen something important

**The setup steps for building a schematic:**

- Create a tsconfig.schematics.json in `projects/my-project/`
- Add npm scripts in `projects/my-project/package.json`
- Add schematic dependencies in `projects/my-project/package.json`
  - Also add them in `<root>/package.json` in case they aren't already exist

**What you need during the build step**
- A `tsconfig.schematics.json`
  - Already part of the npm script `tsc -p tsconfig.schematics.json` 
- Path to `collection.json` and `migration.json` 
  - Read this from `projects/my-project/package.json` "schematics" or "ng-update" path
- Path to `schena.json` for each generate schematic 
  - Most likely 1:1 copyable but in case we need paths we can read that from `collection.json` for each schematic 

> What I noticed is that in the linked documentation, testing is actually not even mentioned Also, the copying of assets can actually be done using ng-packagr, instead of the NPM scripts.

That would be a nice addition but would build things out of sync, wouldn't it? If you now run `ng build <lib>` it would also copy assets but not build the actual schematics

> There is also another point to consider. With this approach we'd need to add the schematics dependencies as library dependencies and not just as devDependencies in the workspace root package.json as otherwise the schematic might fail during runtime because of missing deps.

This is correct but I did not expect this to be an issue. As @dgp1130 said, this is how it is done nowadays

All the other points are out of my scope 😀 If you decide that this is not the right time for it I can still proceed working on ngx-cli-toolkit as a first attempt 

## About "little interest from the community"

> Is this something that we really want to do and support right now? There is little interest from this in the community and it's not trivial to provide a great DX right now.

This is my personal opinion on that and I might be completely wrong but my take on this is: Creating schematics is absolutely inaccessible for Angular developers that are used to having a "one-cli-command setup" for almost everything nowadays. With updating the docs it has gotten a lot better! I started schematics at a time I had to code-surf the Material repo or other libraries that already adopted schematics. I build my own Frankensteins monster with copy/paste code until it has worked

What if people are not interested because they are already overwhelmed by how much you need to know? If you update the docs and say "Try it out! Run `ng generate schematic --type add` and you are done. A good next step would be to add common schematics use cases like you already did for `ng generate` (templates). For example "this is how you can install additional dependencies when people run `ng add <package>`

Also, the tooling world grows as well! There's for example https://github.com/TinkoffCreditSystems/ng-morph that works on top of `ts-morph`. It makes it so much easier to write update schematics for example! 

Again, forgive me if I am completely wrong. This is a highly subjective opinion but also what has motivated me to improve the current situation. I believe schematics is a great concept and needs a little push to be the next better version of itself 🙂 And since you already mentioned other schematic issues I believe this is what the Angular team already does: improving it.

dgp1130:
Had a chat with @alan-agius4 today about this. The biggest concern here is mainly about user expectations for this addition. While schematics are usable today, the tooling around developing them is not particularly mature and we don't have a great story for it right now. We'd like to improve that in a lot of areas, and what's awkward here is that an `ng generate schematic` command is probably the last piece of that puzzle. Some of the other aspects we're hoping to improve are:

* How to build schematics?
* How to test schematics?
* How to configure a library with schematics such that we (the Angular team) can support it via `ng update` and evolve over time.
* Simplifying the public API surface of schematics.
* Better documentation around schematics concepts and how to develop them.
* How to support tools that don't deploy with APF or use the CLI directly?

Improvements here will definitely need to be delivered incrementally. While we could ship `ng generate schematic` first, this sets some misleading user expectations around the rest of the schematics authoring process. Some hackier solutions might work around the build and test problems, but those become much harder to support via `ng update`, something many users might reasonably expect for an officially supported generated piece of code. If all those other features existed, then this would be a very easy win, but the current situation is a bit more nuanced.

Waiting until we have a common schematics runtime and a story around building/testing schematics would provide a much more complete authoring experience that lives up to users' expectations and could be better supported by the Angular team. I see `ng generate schematic` as an excellent ["keystone"](https://martinfowler.com/bliki/KeystoneInterface.html) to these improvements around authoring schematics. Doing this part first makes the full product story of schematics more complicated and the value add for users becomes very fuzzy.

Apologies for the conflicting signals here @kroeder, I know it can be very frustrating. We definitely like the proposal, it's just about a year or so ahead of where we are right now.

kroeder:
These are strong arguments and after reading your post I fully agree and understand this decision! 

 > How to support tools that don't deploy with APF or use the CLI directly?

That is something that confused me in `ngx-cli-toolkit` as well! I knew `ng update` can be executed in e.g. a node project. But `ng generate` cannot and depends on an angular workspace. 

> How to configure a library with schematics such that we (the Angular team) can support it via ng update and evolve over time.

Absolutely understandable. I forgot the fact that random npm scripts are not maintainable and this already increases the scope a lot

Thanks for the summary 👍

PR #22343, opened 2021-12-10
build: update schematics angular vendored TypeScript to 4.5

alan-agius4:


dgp1130 (approved):
Nit: Commit message is a bit confusing to me. Maybe `build: update @schematics/angular vendored TypeScript to 4.5`? Or `build: update vendored TypeScript for Angular schematics to v4.5`?

PR #22353, opened 2021-12-10
refactor(@angular-devkit/core): update missing workspace file error to explicitly call out `angular.json`

dgp1130:
Internally, we've gotten a lot of bugs of users running the CLI in the wrong directory and not understanding this error. Hopefully calling this out more explicitly will key users to double-check their working directory before filing a bug.

### DIFF
--- a/packages/angular_devkit/core/src/workspace/core.ts
+++ b/packages/angular_devkit/core/src/workspace/core.ts
@@ -81,7 +81,10 @@ export async function readWorkspace(
       }
     }
     if (!found) {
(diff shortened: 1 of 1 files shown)


dgp1130:
Equivalent google3 change in http://cl/415625079.

josephperrott (approved):
LGTM

alan-agius4 (approved):

alan-agius4:
Nit: `fix` likely can be used as commit message type.

PR #22360, opened 2021-12-12
feat(@schematics/angular): Added two extra params to skip generating …

gradzio:
We add the `skipComponent` and `skipService` option to `ng g lib` schematic

This allows developers to generate more customized libraries.


alan-agius4:
Hi @gradzio, Thanks for your contribution however deleting a component and service is trivial enough that it doesn't warrant extending the public API with 2 options.

PR #22367, opened 2021-12-14
docs: add `latest-versions.ts` bump to minor/major version release instructions

dgp1130:
In v13.1, we forgot to increment this value and caused an error for early adopters of `13.1.0`. CI should catch this mistake now, and these docs will serve as a reminder to bump the value before release.

/cc @josephperrott FYI, since you're the other release caretaker.


alan-agius4 (approved):
Thanks for this.

Issue #22323, opened 2021-12-09
Angular cache dir (.angular) grows and grows


jcompagner:
When building our product with Angular 13 the .angular/cache is now generated.
At a first build thats around 500MB but it will not take long and it is grown to 9GB is that normal?

What is exactly cached there and how is it reused?

alan-agius4:
Hi @jcompagner, 

There are several parts of the pipeline that are cached on disk, the main benefit of this is that future compilations become warm as the build state is restored from disk instead of having to start the build from a cold state. This helps to decrease the build time by up to 65% as a number of computations are not performed.

 There are a number of things that are cached, including but not limited to: 
- Babel transformations
- Results of the minified CSS and JS
- Build dependencies and resolutions
- Modules transformation result

The cache is expected to continue to grow, unless it is periodically purged by the users. The more code changes are performed in a repo, the larger the cache will become. You can read the [cache documentation](https://angular.io/cli/cache#clearing-the-cache) on how to the clear the cache. 

In future, we are hoping to perform automatic cleanups of stalled cache data. 

It would be interesting to know which files/directories are consuming the most disk usage. Can you please provide some metrics?

jcompagner:
currently after a clean with a first build:

.angular\cache\angular-webpack\452f48df2f16eff7f35d217be73a1e17c8ec5851

the 0.pack file is taking almost all the space 447MB (the whole cache dir id 472MB)

after i have installed 2 updates of packages and run a build the total size si already grown to 516MB
and now the 0.pack file shrink to 207MB but i have a 1.pack now also that is 266MB and a 2.pack file that is 9MB

----                 -------------         ------ ----
-a---           9-12-2021    17:22      217749179 0.pack
-a---           9-12-2021    17:22      279087529 1.pack
-a---           9-12-2021    17:22        9778010 2.pack
-a---           9-12-2021    17:22        7874557 index.pack
-a---           9-12-2021    12:50        7649466 index.pack.old


then i run one build again:

-a---           9-12-2021    17:22      217749179 0.pack
-a---           9-12-2021    17:28      278296822 1.pack
-a---           9-12-2021    17:22        9778010 2.pack
-a---           9-12-2021    17:28        4769610 3.pack
-a---           9-12-2021    17:28        7874590 index.pack
-a---           9-12-2021    17:22        7874557 index.pack.old

running one other build again:

---                 -------------         ------ ----
-a---           9-12-2021    17:22      217749179 0.pack
[...]

PR #22377, opened 2021-12-15
perf(@ngtools/webpack): reduce redundant module rebuilds when cache is restored

alan-agius4:
With this change we reduce the redundant module rebuilds when Webpack's FS cache is restored.

Previously, when the cache was restored we caused all modules to be rebuild even though their contents didn't change. This is because the file emit history wasn't persisted to disk.

This also caused the side effect that Webpack will create additional cache `pack` files due to the unnecessary module rebuilds, which ultimately causes increase of the size of the cache on disk.

Partially addresses https://github.com/angular/angular-cli/issues/22323#issuecomment-991081570

### DIFF
--- a/packages/ngtools/webpack/src/ivy/plugin.ts
+++ b/packages/ngtools/webpack/src/ivy/plugin.ts
@@ -90,12 +90,13 @@ function initializeNgccProcessor(
   return { processor, errors, warnings };
 }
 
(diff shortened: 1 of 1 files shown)


alan-agius4:
From initial benchmarks it also shows a ~600ms reduction of build time on `ng new` apps for warm builds.

dgp1130 (approved):

PR #22379, opened 2021-12-15
fix(@angular-devkit/build-angular): ability to perform DCE but retain symbol names

### LINKED ISSUE #22354: Provide an ability to perform tree-shaking but retain symbol names
# 🚀 Feature request

- [x] build

### Description

Currently it's not possible to use `ng build` to perform dead code elimination via Terser, but retain symbol names in generated JS bundle. It'd be useful to have this capability to simplify bundle size-related investigations (which often times involves looking for retained symbols).

// cc @alan-agius4

alan-agius4:
Previously, we enabled the `keepNames` esbuild when mangling was disabled, this caused dead code to be retained because of the transformations that esbuild did to the input.

Input
```js
class foo {}
```

Output
```js
var l = Object.defineProperty,
  a = (s, c) => l(s, "name", { value: c, configurable: !0 });
class foo {}
a(foo, "foo");
```

Previously we enabled the `keepNames` esbuild option when mangling was disabled, which is actually not needed to retain the name of symbols but is needed for SSR because Domino relies on the `name` property on functions and classes. 

Closes #22354

**Example of output when mangling is disabled**
https://gist.github.com/alan-agius4/6b0b50974bc85214cf1e4fb9fd0f7190

//cc @AndrewKushnir


alan-agius4:
Ummm looks like this breaks the server E2E, will need to continue investigating.

alan-agius4:
CI failures are unrelated to this change.

dgp1130 (approved):
Nit: In the commit message, "Previously we enabled the `keepNames` esbuild _option_ when mangling was disabled, ...".

The commit message sounds more like a feature to me, but I believe this is for an experimental/debug option, so I'm ok with it going into patch anyways.

alan-agius4:
Maybe I should change the type from fix to something else since this is mostly a private/debug option

PR #22402, opened 2021-12-17
fix(@angular-devkit/build-angular): disable parsing `new URL` syntax

alan-agius4:
When web-workers are enabled we allowing parsing `new URL` syntax which has the side-effect that Webpack will treat all assets as asset modules (https://webpack.js.org/guides/asset-modules/#url-assets). With this change we remove this inconsistency by disabling the `url` parsing which doesn't effect `new Worker(new URL(...))`.


hiepxanh:
hope this will be merge soon, I'm stuck after update new version :D

dgp1130 (approved):

PR #22403, opened 2021-12-17
fix(@angular-devkit/build-angular): localized bundle generation fails in watch mode

### LINKED ISSUE #22395: Localized bundle generation fails if `--watch` flag is used

localize

### Is this a regression?

Yes

### Description

Build with `--watch` option fails with error like: `✖ Localized bundle generation failed: ENOENT: no such file or directory, open '/tmp/angular-cli-i18n-oPSjKQ/polyfills.js'`. Error message changes from time to time (it complains about different directories in /tmp and two files: polyfills.js and main.js).
Full command is: `npm run build -- --configuration development --watch`

Project builds without `--watch` option without errors. Full command is: `npm run build -- --configuration development`

It used to work correctly on "^12.2.3" version of angular packages, and it broke since I've updated to "^13.1.0".

https://github.com/human33/angular-i18n-min-reproduction

```true
✖ Localized bundle generation failed: ENOENT: no such file or directory, open '/tmp/angular-cli-i18n-oPSjKQ/polyfills.js'
```

```true
Angular CLI: 13.1.2
Node: 12.22.5
Package Manager: npm 7.5.2
OS: linux x64

Angular: 13.1.1
... localize, material, platform-browser

@angular-devkit/architect       0.1301.2
@angular-devkit/build-angular   13.1.2
@angular-devkit/core            13.1.2
@angular-devkit/schematics      13.1.2
@angular/cli                    13.1.2
@schematics/angular             13.1.2
rxjs                            7.4.0
typescript                      4.5.4
```

### Anything else?

_No response_

alan-agius4:
Previously, we used to delete the temporary emitted JS and map files. However this causes a problem in watch mode, as Webpack will not re-emit these deleted files during the next incremental re-build.

With this change we now delete the entire temporary directory when the process is being terminated instead of a file by file bases.

Closes #22395

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser/index.ts
@@ -73,6 +73,12 @@ export type BrowserBuilderOutput = json.JsonObject &
     outputPath: string;
   };
 
+/**
+ * Maximum time in milliseconds for single build/rebuild
+ * This accounts for CI variability.
+ */
+export const BUILD_TIMEOUT = 30_000;
+
 async function initialize(
   options: BrowserBuilderSchema,
   context: BuilderContext,
--- a/packages/angular_devkit/build_angular/src/utils/i18n-options.ts
+++ b/packages/angular_devkit/build_angular/src/utils/i18n-options.ts
@@ -248,11 +248,12 @@ export async function configureI18nBuild<T extends BrowserBuilderSchema | Server
     const tempPath = fs.mkdtempSync(path.join(fs.realpathSync(os.tmpdir()), 'angular-cli-i18n-'));
     buildOptions.outputPath = tempPath;
 
-    // Remove temporary directory used for i18n processing
-    process.on('exit', () => {
-      try {
-        fs.rmdirSync(tempPath, { recursive: true, maxRetries: 3 });
-      } catch {}
+    process.on('exit', () => deleteTempDirectory(tempPath));
+    process.once('SIGINT', () => {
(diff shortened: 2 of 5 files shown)


packages/angular_devkit/build_angular/src/builders/browser/index.ts
```diff
@@ -73,6 +73,12 @@ export type BrowserBuilderOutput = json.JsonObject &
     outputPath: string;
   };
 
+/**
+ * Maximum time for single build/rebuild
+ * This accounts for CI variability.
+ */
+export const BUILD_TIMEOUT = 40000;
```
dgp1130, line 80: Nit: Add a unit to this variable so future readers know what it represents (I assume milliseconds?). Can add that as a var name suffix or part of the comment.

May also want to consider adding an `_` for readability. I originally read it as `4_000` but it's actually `40_000`.

packages/angular_devkit/build_angular/src/utils/i18n-options.ts
```diff
@@ -249,10 +249,35 @@ export async function configureI18nBuild<T extends BrowserBuilderSchema | Server
     buildOptions.outputPath = tempPath;
 
     // Remove temporary directory used for i18n processing
-    process.on('exit', () => {
+    const deleteTempDirectory = () => {
```
dgp1130, line 252: Nit: Can this be moved to the top-level scope (accept `tempPath` as a parameter)? Does it need to be in this closure?

dgp1130 (approved):

PR #22416, opened 2021-12-21
build: perform package.json substitutions in bazel build

kormide:
@josephperrott @gregmagolan 

This pr mimics the `package.json` substitutions that `build.ts` performs, minus substituting dependency versions for tar paths on a local build (that's coming up in the next pr). The current build will not be affected by this change as we haven't yet flipped `build.ts` to use bazel.

### DIFF
--- a/tools/defaults.bzl
+++ b/tools/defaults.bzl
@@ -4,6 +4,8 @@ load("@npm//@bazel/typescript:index.bzl", _ts_library = "ts_library")
 load("@build_bazel_rules_nodejs//:index.bzl", _pkg_npm = "pkg_npm")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("@npm//@angular/dev-infra-private/bazel:extract_js_module_output.bzl", "extract_js_module_output")
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
+load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
 
 _DEFAULT_TSCONFIG = "//:tsconfig-build.json"
 _DEFAULT_TSCONFIG_TEST = "//:tsconfig-test.json"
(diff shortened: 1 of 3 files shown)


tools/defaults.bzl
```diff
@@ -79,6 +81,47 @@ def pkg_npm(name, use_prodmode_output = False, **kwargs):
         deps = deps,
     )
 
+    # Merge package.json with root package.json and perform various substitutions to
+    # prepare it for release.
+    jq(
+        name = "basic_substitutions",
+        srcs = ["//:package.json"] + kwargs.pop("srcs"),
+        filter = """
```
devversion, line 89: Out of curiosity, why don't we use a Bash/Node script for merging the package json? That would seem a little more readable and maintainable IMO, plus it wouldn't rely on additional dependencies like `jq` binaries.

kormide: Writing a rule that invokes a custom script would've been a viable option, too. We decided to build the `jq` rule to solve the problem of json substitutions more generally, so that others can use it and perform different types of substitutions. Jq filters do require some learning, but the library is also fairly popular and widespread.

If you prefer a custom script I can do that instead. Alternatively I could throw this filter into a file to reduce verboseness—jq can load filters from files as well.

devversion: Yeah, I don't feel strongly enough and I don't think I can make a decision here, but generally I'd prefer to stick with technology that is already commonly used in e.g. this repo. Jq js awesome, but a JS script may be more flexible (in the future), and readable (plus it removes an additional dependency)

We can just leave as is, and move forward (from my side). Happy holidays!

kormide: Happy holidays!

@josephperrott how do you feel about using `jq` here versus a custom js script?

tools/defaults.bzl
```diff
@@ -79,6 +81,47 @@ def pkg_npm(name, use_prodmode_output = False, **kwargs):
...
+            # Root package.json
+            .[0] as $root
+            # Project package.json
+            | .[1] as $proj
```
devversion, line 93: Is it guaranteed that the project sub `package.json` is the second entry here? It seems like it could break when there is another JSON file at package top-level (like for migrations)

kormide: You're right, this will break if the first thing passed to `srcs` in `pkg_npm` is not `package.json`.

kormide: Updated to look through `srcs` and `deps` for package.json and explicitly use that.

kormide:
Closing in favour of https://github.com/angular/angular-cli/pull/22425.

PR #22420, opened 2021-12-22
test: run tests on multiple node versions

dymart:


packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -341,6 +341,10 @@ LARGE_SPECS = {
     },
 }
 
+# TODO: test on multiple node versions
```
jbedard, line 344: Just about any test here seems to fail when running concurrently with multiple node versions so I didn't do any of them for the `build_angular` tests.

jbedard:
Note the current test failure seems to be because the medium test run is faster then expected? See https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/builders/browser/specs/poll_spec.ts

alan-agius4:
Is this something that you still want to land?

jbedard:
Yeah we'd still like to do this one. I had I think all but one test green but have been working on https://github.com/angular/angular-cli/pull/23074 instead lately.

jbedard:
@alan-agius4 there is one TODO left here but I think this is worth landing as-is for now. WDYT?

packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -383,6 +398,9 @@ LARGE_SPECS = {
     for spec in LARGE_SPECS
 ]
 
+# TODO: test on multiple node versions
+# Some tests use specific ports and can not be run concurrently
+# Some tests fail for unknown reasons when run on multiple platforms
```
alan-agius4, line 403: These tests are probably the most important that should be ran on multiple platforms.

Do you have examples of the failures? 

Probably we can investigate this in the near future.

packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -236,15 +244,22 @@ pkg_npm(
...
-    name = "build_angular_api",
-    data = [
-        ":npm_package",
-        "//goldens:public-api",
-    ],
-    golden_dir = "angular_cli/goldens/public-api/angular_devkit/build_angular",
-    npm_package = "angular_cli/packages/angular_devkit/build_angular/npm_package",
-)
+[
+    api_golden_test_npm_package(
```
alan-agius4, line 248: I don't think we should run any of the api_golden tests on multi platform. It's not like the API golden file will change based on OS.

packages/angular_devkit/build_angular/BUILD.bazel
alan-agius4: That shouldn’t be hard to do as the CLI does support port 0. 

I can take a look at these maybe in a week.

alan-agius4: I am happy to get the setup in before that. Unless @clydin has any reservations la

packages/angular_devkit/build_angular/BUILD.bazel
jbedard: Ok these ones have been reverted.

a9c9bbd1507c30cdd3b6a01d36f2f7fbfbf60191 for reference to the original that does change these

packages/angular_devkit/build_angular/BUILD.bazel
jbedard: I'm going to work on this and push to this PR but you can merge the first commit now if you'd like

.bazelrc
```diff
@@ -148,3 +148,5 @@ try-import .bazelrc.user
 # Enable runfiles even on Windows.
 # Architect resolves output files from data files, and this isn't possible without runfile support.
 test --enable_runfiles
+
+build --experimental_remote_merkle_tree_cache
```
jbedard, line 152: This caches the analysis of bazel actions better. Because we have `exports_directories_only = False` (which `ts_library` requires?) every single file in the `node_moduels` gets traversed instead of only the package folders which creates a giant tree. This is also the reason for increasing the circleci `resource_class`.

jbedard:
The flakyness of these tests is very odd, and I think a lot of it has nothing to do with this change. Things like `@angular-devkit/core` not being found errors 🤔

alan-agius4:
> The flakyness of these tests is very odd, and I think a lot of it has nothing to do with this change. Things like `@angular-devkit/core` not being found errors 🤔

Strange as I haven’t seen any similar flakes on the main branch recently.

packages/angular_devkit/build_angular/src/builders/dev-server/tests/options/port_spec.ts
```diff
@@ -40,23 +40,9 @@ describeBuilder(serveWebpackBrowser, DEV_SERVER_BUILDER_INFO, (harness) => {
...
-        // Base options set port to zero
-        port: undefined,
-      });
-
-      const { result, response, logs } = await executeOnceAndFetch(harness, '/');
-
-      expect(result?.success).toBeTrue();
-      expect(getResultPort(result)).toBe('4200');
-      expect(await response?.text()).toContain('<title>');
+      const result = await _defaultOptions({ browserTarget: 'test:build' });
```
alan-agius4, line 43: This change bypasses the end user facing logic.


What we can do in this case to valid default port is the validate that the dev-server started on port 4200, if that fails check for the error messages that port 4200 is in use.

packages/angular_devkit/build_angular/src/builders/dev-server/index.ts
```diff
@@ -56,6 +56,19 @@ export type DevServerBuilderOutput = DevServerBuildOutput & {
   baseUrl: string;
 };
 
+export async function _defaultOptions(
```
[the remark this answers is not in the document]
alan-agius4: The method seems redundant as we the CLI supports port 0 which will cause a free port to be assigned.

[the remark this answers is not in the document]
jbedard: All I did was move it into a function though. I don't think I want to change actual behavior here, just put it in a method that can be tested on its own.

tools/defaults.bzl
```diff
@@ -162,6 +162,7 @@ def pkg_npm(name, pkg_deps = [], use_prodmode_output = False, **kwargs):
         exclude_prefixes = [
             "packages",  # Exclude compiled outputs of dependent packages
         ],
+        allow_overwrites = True,
```
jbedard, line 165: This needed to be added due to an update to `aspect_bazel_lib` that someone else must have done. I don't fully understand why the build didn't fail until this PR though?

packages/angular_devkit/build_angular/src/builders/dev-server/index.ts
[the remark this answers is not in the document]
alan-agius4: It seems that the reason why this was extracted into a method was to be used in the integration tests. But that effects the code path that the tests takes.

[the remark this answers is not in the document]
jbedard: Yeah, I can revert it if that is important. But that means tests can not run in parallel...

packages/angular_devkit/build_angular/src/builders/dev-server/tests/options/port_spec.ts
jbedard: I'll try that so we don't have to disable parallel tests 👍

tools/defaults.bzl
jbedard: I've reverted this for now to make this PR the absolute minimum number of changes.

jbedard:
@alan-agius4 we'd like to merge this as-is without actually enabling the multiple versions and just setting up the toolchains + BUILDs.

It seems like these tests are very flaky and doubling the number of them makes it fail what seems like 100% of the time. Specifically the `packages/angular_devkit/build_angular` tests which are put under the `LARGE_SPECS` variable name, the rest actually seem ok if we want to enable the rest (14/15 of them). This test even fails sometimes with no changes at all (like the PR currently is - only adding the toolchains config).

tools/toolchain_info.bzl
```diff
@@ -0,0 +1,27 @@
+# look at the toolchains registered in the workspace file with nodejs_register_toolchains
+
+# the name can be anything the user wants this is just added to the target to create unique names
+# the order will match against the order in the TOOLCHAIN_VERSION list.
+TOOLCHAINS_NAMES = [
+    "node14",
+    # TODO: enable node16 for unit tests
```
alan-agius4, line 7: Is this because of build-angular tests?

alan-agius4 (approved):

alan-agius4:
@jbedard can you please rebase? 
As I am unable to do that for you as edits by maintainers is not allowed for this PR.

tools/toolchain_info.bzl
alan-agius4: @jbedard  ^

jbedard: @alan-agius4 it seems to be passing fine now (2 builds in a row) so I've enabled it again 🤷

alan-agius4 (approved):

alan-agius4:
Since this change went in, test flakiness have increased. 

One of the main errors is 
```
java.lang.OutOfMemoryError: Java heap space
```

jbedard:
@alan-agius4 can we increase the circleci resource_class?

jbedard:
Or it might actually be better as a separate circleci step per node version? Then we're doubling the number of steps instead of number of concurrent tests. That is probably required for the [e2e tests](https://github.com/angular/angular-cli/pull/23074) so might be a good prefactor that we can do first...

alan-agius4:
We have discussed this during our meeting today and for the time being until we find out the cause of the increase in memory usage. We will like to disable testing on multiple node versions.

The main reasons for this is that we don’t have many targets which justice usage of 14Gb memory considering we also use RBE. An increase the resource class would also double the credit usage if we go with `2 X-large`.

I am curious to know which additional “steps” you thing we can share.

jbedard:
The increase in memory usage is from doubling the number of tests running concurrently in the `test` circleci job.

Today we're running the different node versions concurrently in one circleci step within the test job. We could do the same as the `e2e-test`s and use matrix params to create a separate job per node version, or create a separate step per node version within the `test` job so they run sequentially within a single job.

alan-agius4:
> The increase in memory usage is from doubling the number of tests running concurrently in the test circleci job.

But we are using RBE so they are not actually run on circle. Also I wouldn’t expect 14GB of memory to be used just to run these tests considering we don’t have many targets.

jbedard:
> But we are using RBE so they are not actually run on circle.

Right... you're sure bazel reporting those means it is the circleci machine and not the remote one?

jbedard:
> But we are using RBE so they are not actually run on circle.

Looking at the bazelrc... isn't that only configuring bazel build (not test) to run remotely?

alan-agius4:
test inherits from the build.

https://docs.bazel.build/versions/main/guide.html#option-defaults

jbedard:
If it is running remotely why would `browser-tools/install-chrome` be required? Is there any way to verify it's actually running remote? I'm not too familiar with circleci or remote exec...

alan-agius4:
It does appear that there are some tests that run locally. 

```
```

jbedard:
Right, there are a few tagged as no-remote I think due to protractor/webdriver issues: https://github.com/angular/angular-cli/blob/9171543f06641293cb97d5a801a626a889243db2/packages/angular_devkit/build_angular/BUILD.bazel#L298-L299 (and another one down a bit)

PR #22425, opened 2021-12-24
build: perform package.json substitutions in bazel build

kormide:
@josephperrott @gregmagolan 

This pr mimics the package.json substitutions that build.ts performs. The current build will not be affected by this change as we haven't yet flipped build.ts to use bazel.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -2,6 +2,8 @@
 #
 # Use of this source code is governed by an MIT-style license that can be
 # found in the LICENSE file at https://angular.io/license
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+
 package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"])
@@ -11,10 +13,24 @@ exports_files([
     "tsconfig.json",  # @external
     "tsconfig-test.json",  # @external
     "tsconfig-build.json",  # @external
+    "package.json",
 ])
 
 # Detect if the build is running under --stamp
 config_setting(
     name = "stamp",
     values = {"stamp": "true"},
+)
+
+# If set will replace dependency versions with tarballs for packages in this repo
+bool_flag(
+    name = "enable_package_json_tar_deps",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "package_json_use_tar_deps",
+    flag_values = {
+        ":enable_package_json_tar_deps": "true",
+    },
 )
--- a/packages/angular/cli/BUILD.bazel
+++ b/packages/angular/cli/BUILD.bazel
@@ -316,7 +316,14 @@ genrule(
 
 pkg_npm(
     name = "npm_package",
-    srcs = [":package.json"],
+    pkg_deps = [
+        "//packages/angular_devkit/architect:package.json",
+        "//packages/angular_devkit/build_angular:package.json",
+        "//packages/angular_devkit/build_webpack:package.json",
+        "//packages/angular_devkit/core:package.json",
+        "//packages/angular_devkit/schematics:package.json",
+        "//packages/schematics/angular:package.json",
+    ],
     deps = [
         ":README.md",
         ":angular-cli",
--- a/tools/defaults.bzl
+++ b/tools/defaults.bzl
@@ -1,9 +1,13 @@
 """Re-export of some bazel rules with repository-wide defaults."""
 
 load("@npm//@bazel/typescript:index.bzl", _ts_library = "ts_library")
-load("@build_bazel_rules_nodejs//:index.bzl", _pkg_npm = "pkg_npm")
+load("@build_bazel_rules_nodejs//:index.bzl", "copy_to_bin", _pkg_npm = "pkg_npm")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
 load("@npm//@angular/dev-infra-private/bazel:extract_js_module_output.bzl", "extract_js_module_output")
+load("@aspect_bazel_lib//lib:utils.bzl", "to_label")
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
+load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
+load("//tools:link_package_json_to_tarballs.bzl", "link_package_json_to_tarballs")
 
 _DEFAULT_TSCONFIG = "//:tsconfig-build.json"
 _DEFAULT_TSCONFIG_TEST = "//:tsconfig-test.json"
@@ -44,8 +48,20 @@ def ts_library(
         **kwargs
     )
 
-def pkg_npm(name, use_prodmode_output = False, **kwargs):
-    """Default values for pkg_npm"""
+def pkg_npm(name, pkg_deps = [], use_prodmode_output = False, **kwargs):
+    """Override of pkg_npm to produce package outputs and version substitutions conventional to the angular-cli project.
+
+    Produces a package and a tar of that package. Expects a package.json file
+    in the same folder to exist.
+
+    Args:
+        name: Name of the pkg_npm rule. '_archive.tar.gz' is appended to create the tarball.
+        pkg_deps: package.json files of dependent packages. These are used for local path substitutions when --config=local is set.
+        use_prodmode_output: False to ship ES5 devmode output, True to ship ESM output. Defaults to False.
+        **kwargs: Additional arguments passed to the real pkg_npm.
+    """
+    pkg_json = ":package.json"
+
     visibility = kwargs.pop("visibility", None)
 
     common_substitutions = dict(kwargs.pop("substitutions", {}))
@@ -79,6 +95,57 @@ def pkg_npm(name, use_prodmode_output = False, **kwargs):
         deps = deps,
     )
 
+    # Merge package.json with root package.json and perform various substitutions to
+    # prepare it for release. For jq docs, see https://stedolan.github.io/jq/manual/.
+    jq(
+        name = "basic_substitutions",
+        # Note: this jq filter relies on the order of the inputs
+        # buildifier: do not sort
+        srcs = ["//:package.json", pkg_json],
+        filter_file = "//tools:package_json_release_filter.jq",
+        args = ["--slurp"],
+        out = "substituted/package.json",
+    )
+
+    # Copy package.json files to bazel-out so we can use their bazel-out paths to determine
+    # the corresponding package npm package tar.gz path for substitutions.
+    copy_to_bin(
+        name = "package_json_copy",
+        srcs = [pkg_json],
+    )
+    pkg_deps_copies = []
+    for pkg_dep in pkg_deps:
+        pkg_label = to_label(pkg_dep)
+        if pkg_label.name != "package.json":
+            fail("ERROR: only package.json files allowed in pkg_deps of pkg_npm macro")
+        pkg_deps_copies.append("@%s//%s:package_json_copy" % (pkg_label.workspace_name, pkg_label.package))
+
+    # Substitute dependencies on other packages in this repo with tarballs.
+    link_package_json_to_tarballs(
+        name = "tar_substitutions",
+        src = "substituted/package.json",
+        pkg_deps = [":package_json_copy"] + pkg_deps_copies,
+        out = "substituted_with_tars/package.json",
+    )
+
+    # Move the generated package.json along with other deps into a directory for pkg_npm
+    # to package up because pkg_npm requires that all inputs be in the same directory.
+    copy_to_directory(
+        name = "package",
+        srcs = select({
+            # Do tar substitution if config_setting 'package_json_use_tar_deps' is true (local builds)
+            "//:package_json_use_tar_deps": [":%s_js_module_output" % name, "substituted_with_tars/package.json"],
+            "//conditions:default": [":%s_js_module_output" % name, "substituted/package.json"],
+        }),
+        replace_prefixes = {
+            "substituted_with_tars/": "",
+            "substituted/": "",
+        },
+        exclude_prefixes = [
+            "packages",  # Exclude compiled outputs of dependent packages
+        ],
+    )
+
     _pkg_npm(
         name = name,
         # We never set a `package_name` for NPM packages, neither do we enable validation.
@@ -101,7 +168,7 @@ def pkg_npm(name, use_prodmode_output = False, **kwargs):
             "//conditions:default": substitutions,
         }),
         visibility = visibility,
-        deps = [":%s_js_module_output" % name],
+        nested_packages = ["package"],
         tgz = None,
         **kwargs
     )
--- a/tools/package_json_release_filter.jq
+++ b/tools/package_json_release_filter.jq
@@ -1,0 +1,32 @@
+# Copyright Google Inc. All Rights Reserved.
+#
+# Use of this source code is governed by an MIT-style license that can be
+# found in the LICENSE file at https://angular.io/license
+#
+# This filter combines a subproject package.json with the root package.json
+# and performs substitutions to prepare it for release. It should be called
+# with the --slurp argument and be passed the root pacakge.json followed by
+# the subproject package.json.
+#
+# See jq docs for filter syntax: https://stedolan.github.io/jq/manual/.
+
+.[0] as $root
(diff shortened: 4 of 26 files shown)


packages/angular/cli/BUILD.bazel
```diff
@@ -316,7 +316,14 @@ genrule(
...
     name = "npm_package",
-    srcs = [":package.json"],
+    pkg_deps = [
+        "//packages/angular_devkit/architect:package.json",
+        "//packages/angular_devkit/build_angular:package.json",
+        "//packages/angular_devkit/build_webpack:package.json",
+        "//packages/angular_devkit/core:package.json",
+        "//packages/angular_devkit/schematics:package.json",
+        "//packages/schematics/angular:package.json",
+    ],
```
kormide, line 326: When doing tar substitution we need the dependent package.json files so that we can get the package name and change:
```
  "dependencies": {
    "@some-package: "0.0.0",
  }
```
to
```
  "dependencies": {
    "@some-package: "file:/path/to/pacakge.tar.gz",
  }
```

BUILD.bazel
```diff
@@ -11,10 +13,24 @@ exports_files([
...
+    name = "enable_package_json_tar_deps",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "package_json_use_tar_deps",
+    flag_values = {
+        ":enable_package_json_tar_deps": "true",
+    },
+)
```
kormide, line 36: This is enabled by running `yarn build:bazel --local`. Once we flip to using the bazel script it will just be `yarn build --local`.

packages/angular/cli/BUILD.bazel
kormide: Unlike release and snapshot stamping, which just needs to look for `0.0.0` and replace that.

tools/substitute_tar_deps.bzl
```diff
@@ -33,7 +33,7 @@ def substitute_tar_deps(name, src, pkg_deps, out):
                 pkg_dep,
             ],
             cmd = """
-                TAR=$$(dirname $$(readlink $(execpath {pkg_dep})))/npm_package_archive.tar.gz
+                TAR=$$(dirname $$(cd $$(dirname $(execpath {pkg_dep})) && pwd))/npm_package_archive.tar.gz
```
kormide, line 36: @gregmagolan This readlink trick to get the absolute tar path worked locally but was failing on circleci. It seems that on the ci environment the action isn't run in a sandbox, so there is no symlink from the copy of package.json in the sandbox folder to the exec root. Using this trick instead (cd'ing in to the directory and calling pwd) works on circleci but not locally because locally it will reference the sandbox directory which gets deleted.

I'm a bit confused on what to do here. Should the ci environment be executing actions in a sandbox?

tools/defaults.bzl
```diff
@@ -44,8 +48,20 @@ def ts_library(
         **kwargs
     )
 
-def pkg_npm(name, use_prodmode_output = False, **kwargs):
-    """Default values for pkg_npm"""
+def pkg_npm(name, pkg_deps = [], use_prodmode_output = False, **kwargs):
+    """Override of pkg_npm to produce package outputs and version substitutions conventional to the anguarl-cli project.
```
gregmagolan, line 52: spelling: angular-cli

tools/defaults.bzl
```diff
@@ -44,8 +48,20 @@ def ts_library(
...
+
+    Produces a package and a tar of that package. Expects a package.json file
+    in the same folder to exist.
+
+    Args:
+        name: Name of the pkg_npm rule. '_archive.tar.gz' is appended to create the tarball.
+        pkg_deps: package.json files of dependent packages.
+        use_prodmode_output: False to ship ES5 devmode output, True to ship ESM output.
```
gregmagolan, line 60: nit: Defaults to False

tools/defaults.bzl
kormide: 🙀

tools/defaults.bzl
```diff
@@ -44,8 +48,20 @@ def ts_library(
...
-    """Default values for pkg_npm"""
+def pkg_npm(name, pkg_deps = [], use_prodmode_output = False, **kwargs):
+    """Override of pkg_npm to produce package outputs and version substitutions conventional to the anguarl-cli project.
+
+    Produces a package and a tar of that package. Expects a package.json file
+    in the same folder to exist.
+
+    Args:
+        name: Name of the pkg_npm rule. '_archive.tar.gz' is appended to create the tarball.
+        pkg_deps: package.json files of dependent packages.
```
gregmagolan, line 59: "These are used for local path substitutions when --config=local is set"

tools/substitute_tar_deps.bzl
```diff
@@ -0,0 +1,81 @@
...
+#
+# Use of this source code is governed by an MIT-style license that can be
+# found in the LICENSE file at https://angular.io/license
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
+load("@aspect_bazel_lib//lib:utils.bzl", "to_label")
+
+def substitute_tar_deps(name, src, pkg_deps, out):
+    """Substitute tar paths into a package.json file for the packages it depends on.
+
+    Args:
```
gregmagolan, line 11: "src and pkg_deps must be labels in the bazel-out tree for the derived path to the npm_package_archive.tar.gz to be correct"

tools/substitute_tar_deps.bzl
```diff
@@ -0,0 +1,81 @@
+# Copyright Google Inc. All Rights Reserved.
+#
+# Use of this source code is governed by an MIT-style license that can be
+# found in the LICENSE file at https://angular.io/license
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
+load("@aspect_bazel_lib//lib:utils.bzl", "to_label")
+
+def substitute_tar_deps(name, src, pkg_deps, out):
+    """Substitute tar paths into a package.json file for the packages it depends on.
+
```
gregmagolan, line 10: "src and pkg_deps must be labels in the bazel-out tree for the derived path to the npm_package_archive.tar.gz to be correct"

tools/substitute_tar_deps.bzl
```diff
@@ -0,0 +1,83 @@
...
+        if pkg_label.package != src_pkg:
+            pkg_tar = "@%s//%s:npm_package_archive.tar.gz" % (pkg_label.workspace_name, pkg_label.package)
+            srcs.append(pkg_tar)
+
+        native.genrule(
+            name = "%s_%s_filter" % (name, i),
+            srcs = srcs,
+            cmd = """
+                TAR=$$(dirname $$({abs_path_sandbox} || {abs_path_nosandbox}))/npm_package_archive.tar.gz
+                PKGNAME=$$(cat $(execpath {pkg_name}))
```
gregmagolan, line 57: We can assert the output tree req if we check that `$TAR` contains the string `bazel-out`.

Something like,
```
if [[ "$TAR" != *bazel-out* ]]; then
  echo "ERROR: package.json passed to substitute_tar_deps must be in the output tree. You can use copy_to_bin to copy a source file to the output tree"
  exit 1
fi
```

tools/defaults.bzl
```diff
@@ -79,6 +95,73 @@ def pkg_npm(name, use_prodmode_output = False, **kwargs):
...
+            # Do tar substitution if config_setting 'package_json_use_tar_deps' is true (local builds)
+            "//:package_json_use_tar_deps": [":%s_js_module_output" % name, "substituted_with_tars/package.json"],
+            "//conditions:default": [":%s_js_module_output" % name, "substituted/package.json"],
+        }),
+        replace_prefixes = {
+            "substituted_with_tars/": "",
+            "substituted/": "",
+        },
+        exclude_prefixes = [
+            "node_modules",
```
gregmagolan, line 160: Are these `exclude_prefixes` still needed or just left over from earlier?

kormide: Still needed.

gregmagolan: Cool. Can you comment why? Not clear at first glance

gregmagolan (approved):
Very nice 🌮

tools/defaults.bzl
kormide: I was mistaken. `node_modules` doesn't need to be excluded, but `packages` does. Otherwise the packages contain compiled outputs of the other packages.

kormide: As discussed in standup, the `extract_js_module_output` includes transitive dependencies on other packages, so excluding `packages` is required. Excluding `node_modules` is no longer required due to your change to `copy_to_directory`.

josephperrott (approved):
LGTM

I think it looks good overall, but I want to have @devversion weigh in and see if he has any concerns or questions as this adds some complexity for substitution that we don't have in our other repositories.

josephperrott (changes requested):
Discussed a little bit more with @devversion and I think we both are on the same page.


We have some concerns about using `jq` in the `bzl` file to modify the `package.json`

I think it would be more readable and become testable if we were to use a simple and easier to understand nodejs script that modifies the `package.json` file.

josephperrott (approved):
LGTM

Approved to merge this as is and then follow up with additional testing and reorganization

devversion (approved):
lgtm, we should make sure to keep track of the jq follow-up.

kormide:
> lgtm, we should make sure to keep track of the jq follow-up.

The takeaway from our meeting yesterday was that we should:
* move the large jq filter into a file
* write a test to show an example of how it transforms a package.json file
* link to the jq documentation

I'll be working on this next. Let me know if there's anything else you think I should add to that list.

kormide:
Updated with the above ^.

tools/defaults.bzl
```diff
@@ -79,6 +95,55 @@ def pkg_npm(name, use_prodmode_output = False, **kwargs):
         deps = deps,
     )
 
+    # Merge package.json with root package.json and perform various substitutions to
+    # prepare it for release. For jq docs, see https://stedolan.github.io/jq/manual/.
+    jq(
+        name = "basic_substitutions",
+        srcs = ["//:package.json", pkg_json],
+        filter_file = "//tools:package_json_filter.txt",
```
kormide, line 103: Moved the filter to its own file.

tools/test/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+    args = ["--slurp"],
+    filter_file = "//tools:package_json_filter.txt",
+)
+
+# Test correctness of the filter that prepares each project's package.json file for release
+diff_test(
+    name = "package_json_filter_test",
+    file1 = "expected_package.json",
+    file2 = ":final_package_json",
+)
```
kormide, line 20: Added this test for the filter.

kormide:
Looks like I have to change the tabwidth on the json test files I added.

tools/test/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+load("@aspect_bazel_lib//lib:jq.bzl", "jq")
+
+jq(
+    name = "final_package_json",
+    # buildifier: do not sort
```
gregmagolan, line 6: nit: I'd mention that order matters here explicitly in a comment above the `# buildifier: do not sort` line since the order these are passed to `jq` is load bearing

tools/test/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
...
+    srcs = [
+        "root_package.json",
+        "project_package.json",
+    ],
+    args = [
+        "--slurp",
+    ],
+    filter_file = "//tools:package_json_filter.txt",
```
gregmagolan, line 14: nit: a more verbose name on this file would be good. something like `package_json_release_filter.jq`? I made up the `.jq` extension but it seems reasonable

tools/test/BUILD.bazel
kormide: Good call. I should also add this in it's actual usage in `defaults.bzl`. It doesn't get auto-get sorted there, but someone could change some variable names later on.

tools/defaults.bzl
```diff
@@ -79,6 +95,55 @@ def pkg_npm(name, use_prodmode_output = False, **kwargs):
         deps = deps,
     )
 
+    # Merge package.json with root package.json and perform various substitutions to
+    # prepare it for release. For jq docs, see https://stedolan.github.io/jq/manual/.
+    jq(
+        name = "basic_substitutions",
+        srcs = ["//:package.json", pkg_json],
```
gregmagolan, line 102: nit: mention that order of srcs matters here as well since it is load bearing when passing them to jq

tools/test/BUILD.bazel
gregmagolan: You read my mind. Just left a comment there as well.

tools/test/BUILD.bazel
```diff
@@ -0,0 +1,35 @@
...
+)
+
+# Test correctness of the filter that prepares each project's package.json file for release
+diff_test(
+    name = "package_json_filter_test",
+    file1 = "expected_package.json",
+    file2 = select({
+        # jq outputs CR on windows https://github.com/stedolan/jq/issues/92
+        # strip the CRs to do a correct comparison
+        "@bazel_tools//src/conditions:host_windows": ":final_package_json_cr_stripped",
```
devversion, line 25: Out of curiosity: Why use `host_windows` instead of just `:windows` which would seem more idiomatic and future-proof in terms of RBE etc?

tools/test/BUILD.bazel
```diff
@@ -0,0 +1,35 @@
...
+        "//conditions:default": ":final_package_json",
```
devversion, line 26: Also, why not strip `\r` always? that should avoid the select + another target.

tools/package_json_release_filter.jq
```diff
@@ -0,0 +1,32 @@
...
+| $proj + $root_overrides
+
+# Combine keywords from both
+| .keywords = ($root.keywords + $proj.keywords | unique)
+
+# Remove devDependencies
+| del(.devDependencies)
+
+# Add engines
++ {"engines": {"node": "^12.20.0 || ^14.15.0 || >=16.10.0", "npm": "^6.11.0 || ^7.5.6 || >=8.0.0", "yarn": ">= 1.13.0"}}
```
devversion, line 32: Should the engines really be configured here? personally that seems like a bad spot for me to have it placed. Would it make sense to have it somewhere else in a common spot for configuration of release output?

e.g. in Angular components we have a general config starlark file at the workspace top-level.

This could be a follow-up discussion, but that's just something I feel like could be improved IMO

devversion (approved):
lgtm still, great to see a test for this. I added a few minor comments but other than that it seems ready to me.

tools/test/BUILD.bazel
kormide: I wasn't aware of `:windows`. Glad you pointed it out.

tools/test/BUILD.bazel
kormide: I was thinking it would be more self documenting, but yeah I agree it would be simpler to leave out the select.

tools/package_json_release_filter.jq
kormide: It looks like `engines` is set in the root `package.json`, but it was overridden in here: https://github.com/angular/angular-cli/blob/master/lib/packages.ts#L87

What if we just keep it in the root package.json and have that field override the project package.json in the filter?

Or would you prefer stamped substitutions like in the components repo?

devversion: Good point. I think we cannot use the top-level `engines` field because that one prohobits the use of NPM, while actual consumers of the CLI should be fine to use NPM. Generally I like how we did the majority of versions within Angular Components (it's obviously not done for the engines, but I like that approach where a single place controls these common things; and it's not deeply hidden in some files).

Please feel free to leave this as is and discuss with @josephperrott and/or the CLI team.

josephperrott: Yeah, I would love to do the substitutions if possible, but the follow up makes sense to me.

With the substitutions in place, we could have the template markers placed here in the JQ process and then use the starlark to consistently place the engine values.

josephperrott (approved):
LGTM

PR #22446, opened 2022-01-03
fix(@schematics/angular): set `skipTests` flag for resolvers when using ng new --skip-tests

Lehoczky:
Currently when creating a new project with `ng new xy --skip-tests`, the `ng g` command will still generate spec files for resolvers.

This PR aims to fix this issue.


alan-agius4 (approved):
LGTM, thanks

PR #22454, opened 2022-01-08
fix(@angular-devkit/build-angular): websocket client only injected if required

egoettelmann:
After the webpack-dev-server migration to v4, the websocket client was always injected, even if not required. This caused unnecessary 'ws' requests when live-reload and hmr were disabled.

This PR modifies the building of the webpack-dev-server configuration to configure the websocket client only if either live-reload is enabled or hmr.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/dev-server.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/dev-server.ts
@@ -51,8 +51,6 @@ export async function getDevServerConfig(
     });
   }
 
-  const webSocketPath = posix.join(servePath, 'ws');
-
   return {
     plugins: extraPlugins,
     module: {
@@ -76,11 +74,6 @@ export async function getDevServerConfig(
           },
         ],
       },
-      webSocketServer: {
-        options: {
-          path: webSocketPath,
-        },
-      },
       compress: false,
       static: false,
       server: getServerConfig(root, wco.buildOptions),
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/dev-server.ts
```diff
@@ -51,7 +51,29 @@ export async function getDevServerConfig(
     });
   }
 
-  const webSocketPath = posix.join(servePath, 'ws');
+  const webSocketConfig: {
```
alan-agius4, line 54: Can you please change this, into a function, similar to the other methods? 

```ts
  function getWebSocketAndClientSettings(
    options: WebpackDevServerOptions,
    servePath: string,
  ): {
    webSocketServer?: DevServerConfiguration['webSocketServer'];
    client?: DevServerConfiguration['client'];
  } {
    const { hmr, liveReload } = options;
    if (!(hmr && liveReload)) {
      return {
        webSocketServer: false,
        client: false,
      };
    }

    const webSocketPath = posix.join(servePath, 'ws');
    return {
      webSocketServer: {
        options: {
          path: webSocketPath,
        },
      },
      client: {
        logging: 'info',
        webSocketURL: getPublicHostOptions(options, webSocketPath),
        overlay: {
          errors: true,
          warnings: false,
        },
      },
    };
  }
```

alan-agius4 (approved):
LGTM, thanks

PR #22467, opened 2022-01-10
build: use placeholder version for stamping

kormide:
@josephperrott @devversion @gregmagolan 

The first commit switches the unstamped versions to use the more canonical `0.0.0-PLACEHOLDER` and also affects the non-bazel build as it touches `package.json`s and the legacy build script.

The second commit tries to take the approach in the angular components repo for substituting versions and moves the engine substitution versions there.


gregmagolan:
This is nice in that it makes it clear what is being substituted where and also brings this repo in like with the substitution strings used in angular/angular.

A few questions @kormide,

1) Did you test the output of the legacy build system to verify it is still doing the right thing?

2) Did you run the code path with the change validate-licenses?

substitutions.bzl
```diff
@@ -0,0 +1,14 @@
+# Each placeholder is used to stamp versions during the build process, replacing the key with it's
+# value pair. These replacements occur during building of `npm_package`, primarily in `package.json`s.
+NPM_PACKAGE_SUBSTITUTIONS = {
+    # Version of the local package being built, generated via the `--workspace_status_command` flag.
+    "0.0.0-PLACEHOLDER": "{BUILD_SCM_VERSION}",
+    # Versions to declare in release `package.json`s engines field
+    "0.0.0-ENGINES-NODE": "^12.20.0 || ^14.15.0 || >=16.10.0",
```
gregmagolan, line 7: Was hoisting the versions to here so that they are more visible as opposed to being hidden in the filter file?

gregmagolan: I suspect its better to not add a new top-level file named substitutions.bzl as the substitutions are implementation details of pkg_npm; perhaps a top-level constants.bzl which defines all the constants and then use those constants in `tools/defaults.bzl`. Something like

```
RELEASE_ENGINES_NODE = "^12.20.0 || ^14.15.0 || >=16.10.0"
```

Some constants could be defined for WORKSPACE as well since there are node versions & hard-coded there.

kormide:
> This is nice in that it makes it clear what is being substituted where and also brings this repo in like with the substitution strings used in angular/angular.
> 
> A few questions @kormide,
> 
> 1. Did you test the output of the legacy build system to verify it is still doing the right thing?

Yep.
> 2. Did you run the code path with the change validate-licenses?

Yeah, I changed that in response to a ci failure.

substitutions.bzl
kormide: I added this in response to this feedback on the previous PR: https://github.com/angular/angular-cli/pull/22425#discussion_r780445465. The angular components repo has a top-level bazel file with some version substitutions.

I could make it a file with just constants. Wdyt @devversion?

gregmagolan: Oh. You borrowed the pattern from components. That makes sense. I looked in angular/angular and didn't see any top level .bzl files. Their subs are defined in tools/defaults.bzl https://github.com/angular/angular/blob/master/tools/defaults.bzl#L254

gregmagolan: I see it now https://github.com/angular/components/blob/master/packages.bzl

gregmagolan: I like `constants.bzl` as its clear what the file is meant for but that is just bike shedding at that point

scripts/build.ts
```diff
@@ -411,8 +411,8 @@ export default async function (
             } else {
               obj[depName] = `github:${pkg.snapshotRepo}#${pkg.snapshotHash}`;
             }
-          } else if ((obj[depName] as string).match(/\b0\.0\.0\b/)) {
-            obj[depName] = (obj[depName] as string).replace(/\b0\.0\.0\b/, v);
+          } else if ((obj[depName] as string).match(/\b0\.0\.0-PLACEHOLDER\b/)) {
```
josephperrott, line 414: Do we need to escape the `-` in this case?

```
\b0\.0\.0\-PLACEHOLDER\b
```

substitutions.bzl
kormide: Renamed to `constants.bzl` and moved substitution logic into `tools/defaults.bzl`.

scripts/build.ts
kormide: Nope, the `-` is treated as a literal outside of a character class.

packages/angular_devkit/schematics_cli/blank/project-files/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "<%= dasherize(name) %>",
-  "version": "0.0.0",
+  "version": "0.0.0-PLACEHOLDER",
```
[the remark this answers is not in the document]
kormide: I thought I needed this for a test to pass, but let me verify that again..

substitutions.bzl
devversion: Thank you! yeah I personally like the top-level file as this is something that folks can change without needing a deep understanding of the Bazel setup. Constants sounds good to me 👍

devversion (approved):
LGTM, aside from Joey's comment on the template file

packages/angular_devkit/schematics_cli/blank/project-files/package.json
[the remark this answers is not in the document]
kormide: You were right. It's not needed. Removed.

josephperrott (approved):
LGTM

PR #22469, opened 2022-01-11
docs: add README to `@schematics/update`

dgp1130:
This just points out that the package is deprecated and should no longer be used.

@alan-agius4, I believe this is considered a private package since it's experimental, so I didn't really give much context or justification in the README. Let me know if you think there's anything worth mentioning here.


alan-agius4 (approved):

alan-agius4:
LGTM

PR #22473, opened 2022-01-11
fix(@angular-devkit/build-angular): automatically purge stale build cache entries

### LINKED ISSUE #22323: Angular cache dir (.angular) grows and grows

alan-agius4:
With every build-angular release, previously created cache entries get stale and are no longer used. This causes the cache to keep growing as older files are not purged.

With this change we automatically purge entries that have been created with older version of build-angular and can no longer be used with the current installed version.

Closes #22323

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/server/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/server/index.ts
@@ -19,6 +19,7 @@ import { NormalizedBrowserBuilderSchema, deleteOutputDir } from '../../utils';
 import { i18nInlineEmittedFiles } from '../../utils/i18n-inlining';
 import { I18nOptions } from '../../utils/i18n-options';
 import { ensureOutputPaths } from '../../utils/output-paths';
+import { purgeStaleBuildCache } from '../../utils/purge-cache';
 import { assertCompatibleAngularVersion } from '../../utils/version';
 import { generateI18nBrowserWebpackConfigFromContext } from '../../utils/webpack-browser-config';
 import { getCommonConfig, getStylesConfig } from '../../webpack/configs';
@@ -146,6 +147,9 @@ async function initialize(
   i18n: I18nOptions;
   target: ScriptTarget;
 }> {
+  // Purge old build disk cache.
+  await purgeStaleBuildCache(context);
+
   const originalOutputPath = options.outputPath;
   const { config, i18n, target } = await generateI18nBrowserWebpackConfigFromContext(
     {
--- a/packages/angular_devkit/build_angular/src/utils/purge-cache.ts
+++ b/packages/angular_devkit/build_angular/src/utils/purge-cache.ts
@@ -1,0 +1,53 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { BuilderContext } from '@angular-devkit/architect';
+import { PathLike, existsSync, promises as fsPromises } from 'fs';
+import { join } from 'path';
+import { normalizeCacheOptions } from './normalize-cache';
+
+/** Delete stale cache directories used by previous versions of build-angular. */
+export async function purgeStaleBuildCache(context: BuilderContext): Promise<void> {
+  const projectName = context.target?.project;
+  if (!projectName) {
+    return;
+  }
+
+  const metadata = await context.getProjectMetadata(projectName);
+  const { basePath, path, enabled } = normalizeCacheOptions(metadata, context.workspaceRoot);
+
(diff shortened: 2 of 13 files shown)


packages/angular_devkit/build_angular/src/builders/server/index.ts
```diff
@@ -55,6 +56,8 @@ export function execute(
   // Check Angular version.
   assertCompatibleAngularVersion(root);
 
+  void purgeStaleBuildCache(context);
```
dgp1130, line 59: Is the `void` in order to suppress an unused `Promise` result? Is there perhaps an ignore comment we could use that would make this more clear? I can't say I've ever seen the `void` operator used by a human, so this was quite confusing at first glance.

Outside the style perspective, ignoring a `Promise` always makes me a bit uncomfortable since it's basically racing the entire program and errors don't propagate in a useful way. I get that we don't want to block a user's build on this, but I'm wondering if we should `await` it at the very end so we know it completed successfully without negatively impacting the DX. I assume purging the stale cache is probably quicker than a typical build and only happens when updating the CLI. Otherwise, I'm ok with ignoring the result, just trying to think if there's a way to avoid that pattern without hurting users.

packages/angular_devkit/build_angular/src/utils/purge-cache.ts
```diff
@@ -0,0 +1,58 @@
...
+      ) => Promise<void>;
+    };
+
+    try {
+      if (rm) {
+        await rm(entryPath, { force: true, recursive: true, maxRetries: 3 });
+      } else {
+        await rmdir(entryPath, { recursive: true, maxRetries: 3 });
+      }
+    } catch {}
```
dgp1130, line 56: I feel like I just saw this snippet in another PR, is this a common pattern in the CLI? Would it be worth factoring this out to something reusable?

dgp1130, line 56: Should we log or output the error somewhere? I'm concerned this might silently fail and we'll never know about it.

packages/angular_devkit/build_angular/src/utils/purge-cache.ts
```diff
@@ -0,0 +1,58 @@
...
+  const metadata = await context.getProjectMetadata(projectName);
+  const { basePath, path, enabled } = normalizeCacheOptions(metadata, context.workspaceRoot);
+
+  if (!enabled || !existsSync(basePath)) {
+    return;
+  }
+
+  const entries = await fsPromises.readdir(basePath);
+
+  for (const entry of entries) {
```
dgp1130, line 30: Would it be work parallelizing all the deletes instead of running them sequentially?

dgp1130 (approved):

packages/angular_devkit/build_angular/src/builders/server/index.ts
alan-agius4: I think in this case I’d favor to await the the promise

packages/angular_devkit/build_angular/src/utils/purge-cache.ts
alan-agius4: Yeah, I can do that once both PRs are merged.

packages/angular_devkit/build_angular/src/utils/purge-cache.ts
alan-agius4: That should work too.

packages/angular_devkit/build_angular/src/utils/purge-cache.ts
alan-agius4: Maybe a log message, since we don’t want this to be fatal.

packages/angular_devkit/build_angular/src/builders/server/index.ts
alan-agius4: Actually all these methods are not `async`

packages/angular_devkit/build_angular/src/utils/purge-cache.ts
alan-agius4: Actually the other PR is using the sync version.

dgp1130: Up to you whether it's worth merging them, just something I noticed looking through it.

alan-agius4: One thing to take into consideration here, is that there might be the case that 2 processes try to delete the same path at the same time which would cause an error. 

I think at this point, I prefer not to add an error log.

alan-agius4: PS: I added an E2E to cover the purging of stale cache.

PR #22491, opened 2022-01-12
build: experimental stamping and snapshot repo substitutions

kormide:
@josephperrott @devversion  @gregmagolan

The first commit performs stamping for experimental packages which get a different version format (e.g., `0.1302.0-next.1` vs `13.2.0-next.1`). Basically any package with `"experimental": true` gets the alternative placeholder. If it becomes non-experimental then the version placeholder should also change. The two should be manually kept in sync.

The second commit performs substitutions for github snapshot repos in a snapshot build similar to the non-bazel snapshot build, so it stamps versions like `github:angular/angular-devkit-architect-builds:174fe1056a9ca4987c9176f5f2d18faf0e71c92b`.

### DIFF
--- a/constants.bzl
+++ b/constants.bzl
@@ -2,3 +2,18 @@
 RELEASE_ENGINES_NODE = "^12.20.0 || ^14.15.0 || >=16.10.0"
 RELEASE_ENGINES_NPM = "^6.11.0 || ^7.5.6 || >=8.0.0"
 RELEASE_ENGINES_YARN = ">= 1.13.0"
+
+SNAPSHOT_REPOS = {
+    "@angular/cli": "angular/cli-builds",
+    "@angular/pwa": "angular/angular-pwa-builds",
+    "@angular-devkit/architect": "angular/angular-devkit-architect-builds",
+    "@angular-devkit/architect-cli": "angular/angular-devkit-architect-cli-builds",
+    "@angular-devkit/build-optimizer": "angular/angular-devkit-build-optimizer-builds",
+    "@angular-devkit/build-angular": "angular/angular-devkit-build-angular-builds",
+    "@angular-devkit/build-webpack": "angular/angular-devkit-build-webpack-builds",
+    "@angular-devkit/core": "angular/angular-devkit-core-builds",
+    "@angular-devkit/schematics": "angular/angular-devkit-schematics-builds",
+    "@angular-devkit/schematics-cli": "angular/angular-devkit-schematics-cli-builds",
+    "@ngtools/webpack": "angular/ngtools-webpack-builds",
+    "@schematics/angular": "angular/schematics-angular-builds",
+}
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -413,6 +413,11 @@ export default async function (
             }
           } else if ((obj[depName] as string).match(/\b0\.0\.0-PLACEHOLDER\b/)) {
             obj[depName] = (obj[depName] as string).replace(/\b0\.0\.0-PLACEHOLDER\b/, v);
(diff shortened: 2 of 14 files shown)


packages/angular/cli/models/version.ts
```diff
@@ -22,8 +22,6 @@ export class Version {
   }
 }
 
-// TODO: Convert this to use build-time version stamping after flipping the build script to use bazel
-// export const VERSION = new Version('0.0.0-PLACEHOLDER');
 export const VERSION = new Version(
   (
     JSON.parse(readFileSync(resolve(__dirname, '../package.json'), 'utf-8')) as { version: string }
```
kormide, line 27: Can't do this shortcut anymore now that there are two different version placeholders.

tools/snapshot_repo_filter.bzl
```diff
@@ -0,0 +1,19 @@
+# Copyright Google Inc. All Rights Reserved.
+#
+# Use of this source code is governed by an MIT-style license that can be
+# found in the LICENSE file at https://angular.io/license
+
+load("//:constants.bzl", "SNAPSHOT_REPOS")
+
+# jq filter that replaces package.json dependencies with snapshot repos
+SNAPSHOT_REPO_JQ_FILTER = _generate_snapshot_repo_filter()
```
kormide, line 9: If preferred, I could just hardcode this into a file. I thought it would be cleaner to generate it from `SNAPSHOT_REPOS`.

packages/angular/cli/models/version.ts
kormide: Unless.... will this only apply to the `cli` package? That should never get an experimental version, right?

scripts/build.ts
```diff
@@ -413,6 +413,11 @@ export default async function (
             }
           } else if ((obj[depName] as string).match(/\b0\.0\.0-PLACEHOLDER\b/)) {
             obj[depName] = (obj[depName] as string).replace(/\b0\.0\.0-PLACEHOLDER\b/, v);
+          } else if ((obj[depName] as string).match(/\b0\.0\.0-EXPERIMENTAL-PLACEHOLDER\b/)) {
+            obj[depName] = (obj[depName] as string).replace(
+              /\b0\.0\.0-EXPERIMENTAL-PLACEHOLDER\b/,
+              v,
+            );
```
kormide, line 420: Not ideal to add this extra condition here but this build script will be deleted after flipping to the bazel build script.

josephperrott (approved):
LGTM

constants.bzl
```diff
@@ -2,3 +2,19 @@
...
 RELEASE_ENGINES_NPM = "^6.11.0 || ^7.5.6 || >=8.0.0"
 RELEASE_ENGINES_YARN = ">= 1.13.0"
+
+SNAPSHOT_REPOS = {
+    "@angular/cli": "angular/cli-builds",
+    "@angular/pwa": "angular/angular-pwa-builds",
+    "@angular-devkit/architect": "angular/angular-devkit-architect-builds",
+    "@angular-devkit/architect-cli": "angular/angular-devkit-architect-cli-builds",
+    "@angular-devkit/build-optimizer": "angular/angular-devkit-build-optimizer-builds",
+    "@angular-devkit/build-ng-packagr": "angular/angular-devkit-build-ng-packagr-builds",
```
alan-agius4, line 12: Where is this referenced? This package has long been removed.

kormide: Good to know, I'll remove this entry.

https://github.com/angular/angular-cli/blob/master/.monorepo.json

Want me to remove that reference in `.monorepo.json` in this PR?

alan-agius4 (approved):
LGTM, just one small nit

constants.bzl
kormide: Added

devversion (approved):

PR #22543, opened 2022-01-22
build: use bazel for snapshot builds

kormide:
@gregmagolan @josephperrott 

Here are the snapshot repo diffs caused by flipping to the bazel build.
| Snapshot repo  | Diff |
| ------------- | ------------- |
| @angular/cli  | [diff](https://github.com/angular/cli-builds/compare/master...snapshot-test)  |
| @angular/pwa  | [diff](https://github.com/angular/angular-pwa-builds/compare/master...snapshot-test)  |
| @angular-devkit/architect | [diff](https://github.com/angular/angular-devkit-architect-builds/compare/master...snapshot-test)  |
| @angular-devkit/architect-cli | [diff](https://github.com/angular/angular-devkit-architect-cli-builds/compare/master...snapshot-test)  |
| @angular-devkit/build-angular | [diff](https://github.com/angular/angular-devkit-build-angular-builds/compare/master...snapshot-test)  |
| @angular-devkit/build-webpack | [diff](https://github.com/angular/angular-devkit-build-webpack-builds/compare/master...snapshot-test)  |
| @angular-devkit/core | [diff](https://github.com/angular/angular-devkit-core-builds/compare/master...snapshot-test)  |
| @angular-devkit/schematics | [diff](https://github.com/angular/angular-devkit-schematics-builds/compare/master...snapshot-test)  |
| @angular-devkit/schematics-cli | [diff](https://github.com/angular/angular-devkit-schematics-cli-builds/compare/master...snapshot-test)  |
| @ngtools/webpack | [diff](https://github.com/angular/ngtools-webpack-builds/compare/master...snapshot-test)  |
| @schematics/angular | [diff](https://github.com/angular/schematics-angular-builds/compare/master...snapshot-test)  |

Note that some changes in these diffs may be caused by new things pushed to `master` of the "build" repos.

[CI runs](https://app.circleci.com/pipelines/github/angular/angular-cli?branch=snapshot-test&filter=all) for the `snapshot-test` branch set up to test the ci. The build artifacts are published to the build repos under the same `snapshot-test` branch name.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -270,6 +270,7 @@ jobs:
     resource_class: medium
     steps:
       - custom_attach_workspace
+      - install_python
       - run:
           name: Decrypt Credentials
           # Note: when changing the image, you might have to re-encrypt the credentials with a
--- a/lib/packages.ts
+++ b/lib/packages.ts
@@ -152,14 +152,15 @@ const packageJsonPaths = [
   ..._findPrimaryPackageJsonFiles(path.join(__dirname, '..', 'packages'), excludeRe),
 ];
 
-function _exec(cmd: string) {
-  return execSync(cmd).toString().trim();
+function _exec(cmd: string, opts?: { cwd?: string }) {
+  return execSync(cmd, opts).toString().trim();
 }
 
 let gitShaCache: string;
 function _getSnapshotHash(_pkg: PackageInfo): string {
   if (!gitShaCache) {
-    gitShaCache = _exec('git log --format=%h -n1');
+    const opts = { cwd: __dirname }; // Ensure we call git from within this repo
+    gitShaCache = _exec('git log --format=%h -n1', opts);
   }
 
   return gitShaCache;
--- a/tools/snapshot_repo_filter.bzl
+++ b/tools/snapshot_repo_filter.bzl
@@ -8,7 +8,7 @@ load("//:constants.bzl", "SNAPSHOT_REPOS")
 def _generate_snapshot_repo_filter():
     filter = ""
     for (i, pkg_name) in enumerate(SNAPSHOT_REPOS.keys()):
-        filter += "{sep}(..|objects|select(has(\"{pkg_name}\")))[\"{pkg_name}\"] |= \"github:{snapshot_repo}:BUILD_SCM_HASH-PLACEHOLDER\"\n".format(
+        filter += "{sep}(..|objects|select(has(\"{pkg_name}\")))[\"{pkg_name}\"] |= \"github:{snapshot_repo}#BUILD_SCM_HASH-PLACEHOLDER\"\n".format(
             sep = "| " if i > 0 else "",
             pkg_name = pkg_name,
             snapshot_repo = SNAPSHOT_REPOS[pkg_name],
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@angular/compiler": "14.0.0-next.0",
     "@angular/compiler-cli": "14.0.0-next.0",
     "@angular/core": "14.0.0-next.0",
-    "@angular/dev-infra-private": "https://github.com/angular/dev-infra-private-builds.git#2024033f3123cd1beed78d43ec7269467cd9fa93",
+    "@angular/dev-infra-private": "https://github.com/angular/dev-infra-private-builds.git#3a34722a82fb0e8cd3fa00dfc5476767a35e021a",
     "@angular/forms": "14.0.0-next.0",
     "@angular/localize": "14.0.0-next.0",
     "@angular/material": "14.0.0-next.0",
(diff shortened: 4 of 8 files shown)


gregmagolan (approved):

lib/packages.ts
```diff
@@ -152,14 +152,15 @@ const packageJsonPaths = [
...
+function _exec(cmd: string, opts?: { cwd?: string }) {
+  return execSync(cmd, opts).toString().trim();
 }
 
 let gitShaCache: string;
 function _getSnapshotHash(_pkg: PackageInfo): string {
   if (!gitShaCache) {
-    gitShaCache = _exec('git log --format=%h -n1');
+    const opts = { cwd: __dirname }; // Ensure we call git from within this repo
+    gitShaCache = _exec('git log --format=%h -n1', opts);
```
kormide, line 163: In the snapshot script, there's a `process.chdir()` outside of the git repo, which causes this calculation of the hash to fail. This just makes sure it gets run inside of the git repo. This didn't occur previously because we loaded the legacy build script cached the sha value before that chdir occurred.

.circleci/config.yml
```diff
@@ -270,6 +270,7 @@ jobs:
     resource_class: medium
     steps:
       - custom_attach_workspace
+      - install_python
```
kormide, line 273: The tarring that bazel required python.

tools/snapshot_repo_filter.bzl
```diff
@@ -8,7 +8,7 @@ load("//:constants.bzl", "SNAPSHOT_REPOS")
 def _generate_snapshot_repo_filter():
     filter = ""
     for (i, pkg_name) in enumerate(SNAPSHOT_REPOS.keys()):
-        filter += "{sep}(..|objects|select(has(\"{pkg_name}\")))[\"{pkg_name}\"] |= \"github:{snapshot_repo}:BUILD_SCM_HASH-PLACEHOLDER\"\n".format(
+        filter += "{sep}(..|objects|select(has(\"{pkg_name}\")))[\"{pkg_name}\"] |= \"github:{snapshot_repo}#BUILD_SCM_HASH-PLACEHOLDER\"\n".format(
```
kormide, line 11: Bug in the github dependency format.

josephperrott (approved):
LGTM

kormide:
I'd like to run a few more tests before we mark this as ready to merge.

kormide:
I ran into an issue trying to manually install the snapshots. The bazel build stamps snapshot versions of dependencies like:
```
github:angular/angular-devkit-architect-builds#bb8bc80581417898b10bdc23ef4d9ee4295da2a9
```
It's using the stamp variable `{BUILD_SCM_COMMIT_SHA}` provided by the `ng-dev` tool.

However, legacy build stamps using an abbreviated version of the SHA1, and the snapshot script tags the commit in the build repo using the abbreviated version, so running an npm install of the snapshots fails.

I tried changing the snapshot script to tag using the longer hash, however git complains that the name is too long.

Something I could do is to change the `dev-infra` repo to output another stamp with the abbreviated hash so that we can stamp that instead. Wdyt @josephperrott?

alan-agius4:
Question why is the version `13.2.0-next` and not `14.0.0-next`?

Screenshot 2022-01-31 at 13 06 35

kormide:
>     * why is the version `13.2.0-next` and not `14.0.0-next`?

I think that's just because `master` in the build repo has a more recent snapshot that the branch I'm working off of. That should resolve itself the next time a snapshot it pushed on `snapshot-test`.

> 
>     * can we also turn off source-map generation?
Unfortunately I don't think we can with `ts_library`. @josephperrott previously ok'ed this, and it's consistent with what the `angular` repo produces.

alan-agius4:
> Unfortunately I don't think we can with ts_library. @josephperrott previously ok'ed this, and it's consistent with what the angular bazel build produces.

This is not a big of deal, although sourcemaps are a bit redundant here and kinda of extra mbs over the wire and disk in this case. Unlike for angular/angular were they are useful for app developers. 

But yeah, I don't think there is away to disable sourcemaps with `ts_library`.

kormide:
I'm now able to install the snapshot build locally. I'm getting a failure on the e2e windows tests, which I've seen in other ci runs so it appears unrelated to my changes. Is that test suite known to be flaky? https://app.circleci.com/pipelines/github/angular/angular-cli/20377/workflows/dc026fcf-1783-4d3d-9ac9-9ec9854c1eb3/jobs/286136

alan-agius4:
@kormide, looks like our Node.Js v12 CI is red.

kormide:
> @kormide, looks like our Node.Js v12 CI is red.
> 
> Will investigate tomorrow, however it’s nothing related to this change.

In that case, I think this is ready to go, unless you can think of anything else I should test first.

alan-agius4:
Do you want to wait for @gregmagolan’s review?

kormide:
> Do you want to wait for @gregmagolan’s review?

Your and @josephperrott's review is sufficient. @gregmagolan reviewed previously and is in the loop, but hes out today.

alan-agius4 (approved):
LGTM

dgp1130:
There was a trivial conflict with `package.json` / `yarn.lock` in `13.2.x`. So I merged this to `master` and did a manual cherry pick to `13.2.x` in https://github.com/angular/angular-cli/commit/491e2103bd52bc20c886f44d00fdad02c2124b55.

PR #22553, opened 2022-01-24
fix(@angular-devkit/build-webpack): correctly handle ESM webpack configurations

### LINKED ISSUE #22547: [build_webpack] doesn't work with esm
when adding [build webpack](https://github.com/angular/angular-cli/tree/master/packages/angular_devkit/build_webpack) to an esm Angular project (a project has `type: module` in package.json and `module: ES2022` in tsconfig.json) got the following error
```
An unhandled exception occurred: Must use import to load ES Module: /home/projects/node-tyenbp/webpack.config.js
require() of ES modules is not supported.
```

alan-agius4:
Previoiusly, we didn't correctly handle ESM configurations as the `import` was always downlevelled to `require` by TypeScript.

Closes #22547


eng-dibo:
nice work <3 @alan-agius4

just-jeb:
@alan-agius4 I believe [this](https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/architect/node/node-modules-architect-host.ts#L200) will require the same treatment (if any of the builders is ESM itself).  
Context: https://github.com/just-jeb/angular-builders/issues/1101#issuecomment-1019559724

dgp1130 (approved):

PR #22558, opened 2022-01-24
build: add RC releases to framework peer deps

dgp1130:
This will allow `13.2.0-rc` CLI packages like `@angular-devkit/build-angular` to depend on RC versions of FW packages like `@angular/compiler-cli`. As it stands, `next` releases are allowed, but `rc` aren't, which makes prerelease testing more difficult than it needs to be.

Related: https://github.com/angular/angular-cli/commit/4de2eb54f82c676b94f352f694ffa7268ce98d0d

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -75,11 +75,11 @@
     "esbuild": "0.14.13"
   },
   "peerDependencies": {
-    "@angular/compiler-cli": "^13.0.0 || ^13.2.0-next",
-    "@angular/localize": "^13.0.0 || ^13.2.0-next",
-    "@angular/service-worker": "^13.0.0 || ^13.2.0-next",
+    "@angular/compiler-cli": "^13.0.0 || ^13.2.0-next || ^13.2.0-rc",
+    "@angular/localize": "^13.0.0 || ^13.2.0-next || ^13.2.0-rc",
+    "@angular/service-worker": "^13.0.0 || ^13.2.0-next || ^13.2.0-rc",
     "karma": "^6.3.0",
-    "ng-packagr": "^13.0.0 || ^13.2.0-next",
+    "ng-packagr": "^13.0.0 || ^13.2.0-next || ^13.2.0-rc",
     "protractor": "^7.0.0",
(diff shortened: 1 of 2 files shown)


alan-agius4 (commented):
These should not be needed https://jubianchi.github.io/semver-check/#/^13.2.0-next/13.2.0-rc.0

dgp1130:
Hmm, looking at the [spec](https://semver.org/#spec-item-11) (not sure how authoritative this is) this actually seems like it's somewhat coincidental. I don't see `next` or `rc` actually hard coded (outside of examples) with defined semantics or ordering. It looks like these specifiers are compared lexicographically:

> Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:
> 
> 1. Identifiers consisting of only digits are compared numerically.
> 
> 2. Identifiers with letters or hyphens are compared lexically in ASCII sort order.
> 
> 3. Numeric identifiers always have lower precedence than non-numeric identifiers.
> 
> 4. A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.
> 
> Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.

So the fact that `rc` has higher priority than `next` is kind of coincidental and is pretty awkward behavior to depend upon. I'm wondering if it might be a good idea to bump these anyway as a standard process?

alan-agius4:
@dgp1130, indeed the `rc`, `next`, `beta`, `alpha` are ordered lexicographically. 

> I'm wondering if it might be a good idea to bump these anyway as a standard process?
I am not too strongly about this. 

I still think that this is redundant due to the lexicographically order, but If you feel that we should add these I am okay with that too.

dgp1130:
Maybe it's just the way I think, but I feel that it's worth doing `next || rc` for clarity in the peer dep, or else someone debugging installed versions could be very easily confused about this behavior.

We need to update the versions for `next` anyways, so this doesn't really adds any more maintenance effort, just updating the version twice right next to each other. Lets try this strategy.

PR #22571, opened 2022-01-25
     feat(@schematics/angular): update Angular dependencies to use `^` as version prefix

### LINKED ISSUE #21925: Update Angular dependencies to use `^` for semver
# 🚀 Feature request

- [X] new
- [X] update

### Description

Angular applications are generated via `ng new` to include a `package.json` with `~` characters, meaning they only automatically include patch updates to dependencies, not feature updates. This was intended to reduce the risk of breakage due to other dependencies accidentally shipping breaking changes in minor versions (which has happened several times in the past). Downside of this is that Angular deps are not updated as frequently as they could be and we have our own CI systems and integration testing to make sure that we don't introduce breaking changes into a minor versions. There's no need to apply the same requirement to Angular when we're able to more consistently and regularly test each version.

New applications should be generated with `^` for Angular dependencies and existing applications should be migrated for consistency. If an existing app pins an explicit version (such as `"@angular/core": "12.2.4"`), then we should update that as normal but leave the semver string bare so as not to change user's intent.

Should be an easy change, though we probably just missed v13. This definitely requires a major version, so the soonest we could do this is v14.

alan-agius4:
With this change we update Angular dependencies to use `^` instead of `~` as version prefix.
    
Closes #21925


packages/schematics/angular/migrations/update-14/angular-packages-version-prefix.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import { Rule } from '@angular-devkit/schematics';
+import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
+import { NodeDependencyType } from '../../utility/dependencies';
+import { JSONFile } from '../../utility/json-file';
+
+const PACKAGES_REGEXP = /^@(?:angular|nguniversal|schematics|angular-devkit)\/|^ng-packagr$/;
+
+export default function (): Rule {
```
dgp1130, line 16: A doc comment for this rule, maybe with a short example, would be really useful to communicate exactly what the purpose of the migration this.

packages/schematics/angular/migrations/update-14/angular-packages-version-prefix.ts
```diff
@@ -0,0 +1,47 @@
...
+  return (tree, context) => {
+    const json = new JSONFile(tree, '/package.json');
+    updateVersionPrefixToTilde(json, NodeDependencyType.Default);
+    updateVersionPrefixToTilde(json, NodeDependencyType.Dev);
```
dgp1130, line 20: Are there any (maybe lesser-used) packages that would be placed in peer deps or optional deps? I just want to make sure there's no need to migrate other forms of dependencies as well. Would it be safer to update those too?

packages/schematics/angular/migrations/update-14/angular-packages-version-prefix_spec.ts
```diff
@@ -0,0 +1,151 @@
...
+describe('Migration to update Angular packages version prefix to `^` instead of `~`', () => {
+  const packageJsonPath = '/package.json';
+  const schematicName = 'update-angular-packages-version-prefix';
+  const schematicRunner = new SchematicTestRunner(
+    'migrations',
+    require.resolve('../migration-collection.json'),
+  );
+
+  let tree: UnitTestTree;
+  beforeEach(() => (tree = new UnitTestTree(new EmptyTree())));
```
dgp1130, line 21: Nit: I recommend using typical braces for assignments like this rather than an expression assignment, which is generally only used in optimized builds.

dgp1130 (approved):
Looks great!

I assume the typical `ng update` version bumps already respect the user's prefix and won't clobber the `^` with a `~`? Just want to make sure it doesn't overwrite all this hard work.

dgp1130:
Also, let's hold off on merging breaking changes like this until we're sure that we aren't doing 13.3.

packages/schematics/angular/migrations/update-14/angular-packages-version-prefix_spec.ts
alan-agius4: Agreed

packages/schematics/angular/migrations/update-14/angular-packages-version-prefix.ts
alan-agius4: Not in the root of the package.json. 

There might be peer deps in the `package.json` of library projects. But these are generally added by users to specify which versions of Angular their library supports, so I don't think we should amend those.

alan-agius4:
> I assume the typical ng update version bumps already respect the user's prefix and won't clobber the ^ with a ~? Just want to make sure it doesn't overwrite all this hard work.

Yeah it does.

packages/schematics/angular/migrations/update-14/angular-packages-version-prefix.ts
dgp1130: Sounds good, just wanted to verify.

PR #22577, opened 2022-01-26
refactor(@angular-devkit/build-angular): clean find tests logic

alan-agius4:
With this change we clean up and reduce the complexity of the `findTests` logic that handles the `include` option in the karma builder.


packages/angular_devkit/build_angular/src/builders/karma/find-tests.ts
```diff
@@ -6,55 +6,85 @@
...
 }
 
-function findMatchingTests(pattern: string, cwd: string, relativePathToMain: string): string[] {
+const normalizePath = (path: string): string => path.replace(/\\/g, '/');
+
+async function findMatchingTests(
+  pattern: string,
+  workspaceRoot: string,
+  projectSourceRoot: string,
+): Promise<string[] | string> {
```
dgp1130, line 37: Can we have this always return `string[]` and just collapse a single result to `[ result ]`?

packages/angular_devkit/build_angular/src/builders/karma/find-tests.ts
```diff
@@ -6,55 +6,85 @@
...
+): Promise<string[]> {
+  const matchingTestsPromises = patterns.map((pattern) =>
+    findMatchingTests(pattern, workspaceRoot, projectSourceRoot),
+  );
+  const files = await Promise.all(matchingTestsPromises);
 
-    return files;
-  }, [] as string[]);
+  // Unique file names
+  return [...new Set(files.flatMap((f) => f))];
```
dgp1130, line 28: Could we use `flies.flat()` here? An empty `flatMap()` is a bit unintuitive in its purpose.

packages/angular_devkit/build_angular/src/builders/karma/index.ts
```diff
@@ -112,9 +112,22 @@ export function execute(
...
+        if (!projectName) {
+          throw new Error('The builder requires a target.');
+        }
+
+        const projectMetadata = await context.getProjectMetadata(projectName);
+        const projectSourceRoot = getSystemPath(
+          join(
+            normalize(context.workspaceRoot),
+            (projectMetadata.root as string | undefined) || '',
+            (projectMetadata.sourceRoot as string | undefined) || '',
```
dgp1130, line 126: Would `??` work here and better communicate the intent?

dgp1130 (approved):
Just a few coding nitpicks.

packages/angular_devkit/build_angular/src/builders/karma/find-tests.ts
alan-agius4: Good call!

PR #22580, opened 2022-01-26
Bump version to "v13.2.0" with changelog.

dgp1130:
NOTE: This is a non-standard release since it is a minor. This PR includes manual edits to `latest-versions.ts` and peer deps on the framework to drop prerelease tags and bump the versions to `13.2.0` where appropriate.


dgp1130:
Actually looks like FW did just release, so we should be good to go if CI passes.

alan-agius4 (approved):

PR #22582, opened 2022-01-27
refactor(@angular-devkit/core): remove deprecated fs, object and array APIs

alan-agius4:
BREAKING CHANGE:

The below APIs have been removed without replacement. Users should leverage other Node.js or other APIs.
- `fs` namespace
- `clean`
- `mapObject`


dgp1130 (approved):
Is the expectation that users should leverage existing Node APIs rather than depending on devkit for these functions? We might want to call that out in the changelog.

alan-agius4:
@dgp1130, actually users will need to create a similar functionality in their project. Node itself doesn't provide these helpers.

dgp1130:
Right, I just meant that users should compose existing Node APIs just like we did. I get that it's probably a suboptimal interface when all they want to know is "Is this a directory?" But that's a Node problem, not an Angular one, so it makes sense that they shouldn't rely on devkit for such a use case.

PR #22583, opened 2022-01-27
refactor(@angular-devkit/schematics): remove deprecated `analytics` property

alan-agius4:
BREAKING CHANGE:

Deprecated `analytics` property has been removed from `TypedSchematicContext` interface


dgp1130 (approved):
What exactly was the purpose of this property? What exactly should users do instead if they were using it?

alan-agius4:
@dgp1130, it’s an unused property.

PR #22584, opened 2022-01-27
refactor(@angular/cli): remove deprecated `--prod` flag

alan-agius4:
BREAKING CHANGE:

Deprecated option `--prod` has been removed from all builders. `--configuration production`/`-c production` should be used instead if the default configuration of the builder is not configured to `production`.

### DIFF
--- a/tests/legacy-cli/e2e/tests/build/lazy-load-syntax.ts
+++ b/tests/legacy-cli/e2e/tests/build/lazy-load-syntax.ts
@@ -16,29 +16,37 @@ export default async function () {
 
   // Add app routing.
   // This is done automatically on a new app with --routing.
-  await writeFile(appRoutingModulePath, `
-    import { NgModule } from '@angular/core';
-    import { Routes, RouterModule } from '@angular/router';
+  await writeFile(
+    appRoutingModulePath,
+    `
+      import { NgModule } from '@angular/core';
+      import { Routes, RouterModule } from '@angular/router';
 
-    const routes: Routes = [];
+      const routes: Routes = [];
 
-    @NgModule({
-      imports: [RouterModule.forRoot(routes)],
-      exports: [RouterModule]
-    })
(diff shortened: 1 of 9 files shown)


tests/legacy-cli/e2e/tests/build/lazy-load-syntax.ts
```diff
@@ -27,32 +29,44 @@ export default async function () {
       exports: [RouterModule]
     })
     export class AppRoutingModule { }
-  `);
-  await prependToFile('src/app/app.module.ts',
-    `import { AppRoutingModule } from './app-routing.module';`);
+  `,
+  );
```
dgp1130, line 33: Nit: If the formatter allows it, I recommended lining up indents between the starting and ending backticks (and maybe indent the text one more level), which can help readers associate the start and end tokens together.

```typescript
await writeFile(
  appRoutingModulePath,
  `
    // ...
  `,
);
```

tests/legacy-cli/e2e/tests/misc/third-party-decorators.ts
```diff
@@ -24,7 +26,7 @@ export default function () {
           async getCounter() { return element(by.xpath('/html/body/app-root/div/span')).getText(); }
         }
       `,
```
dgp1130, line 28: Nit: Similar point here, can we indent the code under the file paths? Hopefully this doesn't fight the formatter too much.

```typescript
// ...
  .then(_ => writeMultipleFiles({
    './e2e/src/app.po.ts': `
      import { browser, by, element } from 'protractor';
      // ...
    `,
    // ...
  }));
```

dgp1130 (approved):
I recommend pointing out `-c production` in the changelog as a shorthand for `--configuration production`, as I imagine many users won't want to type out that much, and it would be nice to head off that concern as best we can.

PR #22585, opened 2022-01-27
feat(@schematics/angular): update new and existing projects compilation target to `ES2020`

alan-agius4:
With this change we update the TypeScript compilation target to `ES2020` for both new and existing projects. This is because all browsers that Angular supports (https://angular.io/guide/browser-support) support `ES2020` features without the need for polyfills.

### DIFF
--- a/packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
+++ b/packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
@@ -1,0 +1,56 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Rule } from '@angular-devkit/schematics';
+import { JSONFile } from '../../utility/json-file';
+import { getWorkspace } from '../../utility/workspace';
+import { Builders } from '../../utility/workspace-models';
+
+/** Migration to update tsconfig compilation target option to es2020. */
+export default function (): Rule {
+  return async (host) => {
+    /** Builders for which the migration will run. */
+    const supportedBuilders = [Builders.Karma, Builders.NgPackagr, Builders.Browser];
+
+    /** Compilation targets values that should not be amended. */
+    const skipTargets = ['es2020', 'es2021', 'es2022', 'esnext'];
+
+    const uniqueTsConfigs = new Set(['/tsconfig.json']);
+
+    // Find all tsconfig files which are refereced by the builders.
+    const workspace = await getWorkspace(host);
+    for (const project of workspace.projects.values()) {
+      for (const target of project.targets.values()) {
+        if (!supportedBuilders.includes(target.builder as Builders)) {
+          // Unknown builder.
+          continue;
+        }
+
+        // Update all other known CLI builders that use a tsconfig.
+        const allOptions = [target.options ?? {}, ...Object.values(target.configurations ?? {})];
+        for (const opt of allOptions) {
+          if (typeof opt?.tsConfig === 'string') {
+            uniqueTsConfigs.add(opt.tsConfig);
+          }
+        }
+      }
+    }
+
+    // Modify tsconfig files
+    const targetJsonPath = ['compilerOptions', 'target'];
+    for (const tsConfigPath of uniqueTsConfigs) {
+      const json = new JSONFile(host, tsConfigPath);
+      const target = json.get(targetJsonPath);
+
+      // Update compilation target when it's current set lower than es2020.
+      if (typeof target === 'string' && !skipTargets.includes(target.toLowerCase())) {
+        json.modify(targetJsonPath, 'es2020');
+      }
+    }
+  };
+}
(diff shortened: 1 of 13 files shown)


packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const SUPPORTED_BUILDERS = [Builders.Karma, Builders.NgPackagr, Builders.Browser];
+
+/** Migration to update tsconfig compilation target option to es2020. */
+export default function (): Rule {
+  return async (host) => {
+    // Workspace level tsconfig
+    updateTargetToES2020(host, 'tsconfig.json');
+
+    // Find all tsconfig which are refereced by the builders
```
dgp1130, line 22: Typo: "Find all tsconfig *files* which..."

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
```diff
@@ -0,0 +1,56 @@
...
+    const workspace = await getWorkspace(host);
+    for (const [, project] of workspace.projects) {
+      for (const [, target] of project.targets) {
```
dgp1130, line 25: Nit: Could we do `for (const project of Object.values(workspace.projects)) { /* ... */ }`?

That would avoid the awkward destructure.

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
```diff
@@ -0,0 +1,56 @@
...
+        if (!SUPPORTED_BUILDERS.includes(target.builder as Builders)) {
+          // Unknown builder
+          continue;
+        }
+
+        // Update all other known CLI builders that use a tsconfig
+        const tsConfigs = [target.options || {}, ...Object.values(target.configurations || {})]
```
dgp1130, line 32: Nit: Would `??` better communicate the intent over `||`?

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
```diff
@@ -0,0 +1,56 @@
...
+          .filter((opt) => typeof opt?.tsConfig === 'string')
```
dgp1130, line 33: Nit: We can drop the cast by adding an `is` clause to the `filter()` result.

```typescript
const tsConfigs = [ /* ... */ ]
    .filter((opt) opt is { tsConfig: string } => typeof opt?.tsConfig === 'string')
    .map((opt) => opt.tsConfig);
```

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
```diff
@@ -0,0 +1,56 @@
...
+import { getWorkspace } from '../../utility/workspace';
+import { Builders } from '../../utility/workspace-models';
+
+const SUPPORTED_BUILDERS = [Builders.Karma, Builders.NgPackagr, Builders.Browser];
+
+/** Migration to update tsconfig compilation target option to es2020. */
+export default function (): Rule {
+  return async (host) => {
+    // Workspace level tsconfig
+    updateTargetToES2020(host, 'tsconfig.json');
```
dgp1130, line 20: Idea: I wonder if this would be simpler to have a `findTsConfigs()` function which returns `[ 'tsconfig.json', ...tsconfigsFromWorkspace ]`, and then run `updateTargetToES2020()` for all of them. That might be clearer than updating each one imperatively.

dgp1130, line 20: Question: Does Angular require that the main tsconfig is named `tsconfig.json`? Could users rename this and would it fail the migration if so? I'm thinking we just need to check if the file exists and only migrate it if so.

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+    }
+  };
+}
+
+function updateTargetToES2020(host: Tree, tsConfigPath: string) {
+  const json = new JSONFile(host, tsConfigPath);
+  const target = json.get(['compilerOptions', 'target']);
+
+  if (typeof target === 'string' && target.toLowerCase() !== 'es2020') {
```
dgp1130, line 53: Should we only update if target < ES2020? If someone was running on ES2021 or next, we shouldn't downgrade them.

dgp1130 (approved):

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
alan-agius4: Replaced it with;

```ts
    for (const project of workspace.projects.values()) {
      for (const target of project.targets.values()) {
```

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
alan-agius4: Technically they can rename it, but I don't think that anyone does because;

1) This is the tsconfig that is used by the IDE - without it the TS language service will not work correctly. 
2) This tsconfig is expected to be there by other schematics that we have, such as `ng g app`, `ng g lib` etc...

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
alan-agius4: Good call.

packages/schematics/angular/migrations/update-14/update-tsconfig-target.ts
dgp1130: Since there would be other problems from renaming the base `tsconfig.json` anyways, I'm not too concerned about this one.

PR #22587, opened 2022-01-27
docs: move developer guide into a seperate file

alan-agius4:
Currently, the `readme.md` also contains information on how to build and test the CLI which isn't useful to most users. With this change we move this information into a seperate document.


alan-agius4:
Preview https://github.com/alan-agius4/angular-cli/tree/developer.md-file

README.md
```diff
@@ -10,44 +10,137 @@
...
+  <br>
+  <img src="/docs/images/angular-cli-logo.png" alt="Angular CLI logo" width="100px" height="108px"/>
+  <br><br>
+  <i>The Angular CLI is a command-line interface tool that you use to initialize, develop, scaffold, 
+    <br>and maintain Angular applications directly from a command shell.</i>
+  <br>
+</p>
 
+<p align="center">
+  <a href="https://cli.angular.io"><strong>cli.angular.io</strong></a>
```
dgp1130, line 25: I wonder if we should point people at angular.io/cli instead? This subdomain redirects there already and I don't think we have any plans to do anything with it.

README.md
```diff
@@ -41,128 +41,6 @@ Angular DevKit.
 DevKit's goal is to provide a large set of libraries that can be used to manage, develop, deploy and
 analyze your code.
 
-# Getting Started - Local Development
```
dgp1130, line 44: We should at least link to the new `DEVELOPER.md` doc from the `README`, maybe somewhere under Contributing. That way someone who wants to make a PR can still find the doc, but people who just want to use the CLI will recognize that it isn't for them.

dgp1130 (approved):
Loving the new look! Very clean and direct.

README.md
alan-agius4: I don’t have any strong objects. Although in the angular/angular readme cli.angular.io is used.

PR #22579, opened 2022-01-26
release: update version to 14.0.0-next.0

josephperrott:

PR #22596, opened 2022-01-28
Refactor `ng update` E2E

alan-agius4:
**test(@angular/cli): remove hardcoded versions from ng-update test**
    
We improve the migrations logic to automatically run the needed migrations based on the build version of the Angular CLI. This helps us to avoid having to manually update this test when we bump the major version. Example: https://github.com/angular/angular-cli/pull/22579/files

**test(@angular/cli): update ng-update test to start from version 10**
    
With this change we update the `ng-update` test to start from version 10 as version 9 has been out of support for some time now.


alan-agius4:
@dgp1130, the files in first commit were created via `npx @angular/cli@10 new ten-project`.

tests/legacy-cli/e2e/tests/update/update-10.ts
```diff
@@ -0,0 +1,62 @@
...
+    // CLI current version.
+    const cliMajorVersion = NgCLIVersion.major;
+
+    for (let version = cliMajorProjectVersion + 1; version < cliMajorVersion; version++) {
+      // Run all the migrations until the current build major version - 1.
+      // Example: when the project is using CLI version 10 and the build CLI version is 14.
+      // We will run the following migrations:
+      // - 10 -> 11
+      // - 11 -> 12
+      // - 12 -> 13
```
dgp1130, line 30: Thanks for this comment, the example is really helpful here.

tests/legacy-cli/e2e/tests/update/update-10.ts
```diff
@@ -0,0 +1,62 @@
...
+      const { stdout } = await ng('update', `@angular/cli@${version}`, `@angular/core@${version}`);
+      if (!stdout.includes("Executing migrations of package '@angular/cli'")) {
```
dgp1130, line 32: Should we check the exit code as well or is that implicit in the `ng` command? Is there a scenario where `ng update` exits with code 0 but doesn't run the migration?

tests/legacy-cli/e2e/tests/update/update-10.ts
```diff
@@ -0,0 +1,62 @@
...
+  } finally {
+    await setRegistry(true);
+  }
+
+  // Update Angular current build
+  const extraUpdateArgs = isPrereleaseCli() ? ['--next', '--force'] : [];
+  // For the latest/next release we purposely don't add `@angular/core`.
+  // This is due to our bumping strategy, which causes a period were `@angular/cli@latest` (v12.0.0) `@angular/core@latest` (v11.2.x)
+  // are of different major/minor version on the local NPM server. This causes `ng update` to fail.
+  // NB: `ng update @angula/cli` will still cause `@angular/core` packages to be updated.
```
dgp1130, line 45: Typo: `@angular/cli`

tests/legacy-cli/e2e/tests/update/update-10.ts
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+  } finally {
+    await setRegistry(true);
+  }
+
+  // Update Angular current build
+  const extraUpdateArgs = isPrereleaseCli() ? ['--next', '--force'] : [];
+  // For the latest/next release we purposely don't add `@angular/core`.
+  // This is due to our bumping strategy, which causes a period were `@angular/cli@latest` (v12.0.0) `@angular/core@latest` (v11.2.x)
```
dgp1130, line 43: I don't fully understand this. Does this happen right after we release a major version but still have dev dependencies on FW's latest patch? If we skip `@angular/core` here, will we ever run update schematics for the upcoming major?

dgp1130 (commented):
Thanks for making this change! Glad we can remove one more version bump.

Mainly just a question about how we depend on FW for this test.

tests/legacy-cli/e2e/tests/update/update-10.ts
alan-agius4: Only when there aren’t any migrations, which is unlikely.

dgp1130: Ok, that's probably fine to leave it, though we should probably include a comment that we're verifying at least one migration was executed.

tests/legacy-cli/e2e/tests/update/update-10.ts
alan-agius4: There can be a period when we bump the master to a new major ex: 14.0.0-next.0 but the same version of the FW is not yet available on NPM. This would cause the test to fail.

alan-agius4: `ng update @angular/cli` will always update and run `@angular/core` because the later is a peer dependency of the CLI.

dgp1130: I think major releases do an immediate prerelease specifically to avoid that state. The [release checklist](https://docs.google.com/document/d/1GAyhWTTxsb28AlJZpn4yjrDZW3nT8NMv6xP5aQ9qcvA/edit?resourcekey=0-eM9Rx3k1fFjNaoTbcKEyhQ#) has FW release the major version *and* the next before CLI releases it's major and bumps to next. Does that mean this state shouldn't happen?

alan-agius4: I requested access to the document.

Ill dig more into this, but I remember it was a recurring issue, and this was a workaround without any side-effect.

alan-agius4: The problem was mainly during the RC periods, when FW do not release master as next version.

Ex: CLI master is 14.0.0-next.0, but @angular/core@next is pointing to 13.2.0-rc.0 and on NPM @angular/core@14.0.0-next.0 doesn’t exist yet. This would cause `ng update @angular/core@14` to fail.

dgp1130: Gave you access, I assumed the team already had it. Agreed that we should chat more on Monday.

alan-agius4: I updated the comment to capture as much as I can to why this needs to be here.

dgp1130: Thanks for the comment, I hadn't considered the case where FW ships a major version as stable but hasn't yet tagged it as `latest`. That addresses my concern, and installing just `@angular/cli` is a reasonable workaround.

dgp1130 (approved):

PR #22599, opened 2022-01-28
refactor(@angular-devkit/build-angular): remove deprecated `showCircularDependencies` browser and server builder option

alan-agius4:
BREAKING CHANGE:

The deprecated `showCircularDependencies` browser and server builder option has been removed. The recommended method to detect circular dependencies in project code is to use either a lint rule or other external tools.


packages/schematics/angular/migrations/update-14/update-angular-config.ts
```diff
@@ -0,0 +1,28 @@
...
+/** Migration to remove deprecated/removed builder options. */
+export default function (): Rule {
+  return updateWorkspace((workspace) => {
+    for (const project of workspace.projects.values()) {
+      for (const target of project.targets.values()) {
+        if (
+          target.builder === '@angular-devkit/build-angular:server' ||
+          target.builder === '@angular-devkit/build-angular:browser'
+        ) {
+          for (const [, options] of allTargetOptions(target)) {
```
dgp1130, line 21: Nit: Can use `.values()` / `Object.values()` to drop the destructure.

packages/schematics/angular/migrations/update-14/update-angular-config.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Rule } from '@angular-devkit/schematics';
+import { allTargetOptions, updateWorkspace } from '../../utility/workspace';
+
+/** Migration to remove deprecated/removed builder options. */
```
dgp1130, line 12: Do we expect to remove more options in this schematic? This feels more vague than it needs to be if we're just dropping `showCircularDependencies`, but maybe you have bigger plans for it?

dgp1130 (approved):

packages/schematics/angular/migrations/update-14/update-angular-config.ts
alan-agius4: I though to name it generic in case we need to remove addition options. 

I checked again and it seems we don’t have any more deprecations that we can remove in v14. 

Therefore , I’ll rename it to be more explicit.

packages/schematics/angular/migrations/update-14/update-angular-config.ts
alan-agius4: Sorry about this, old habit!

dgp1130: No problem! I don't have anything against empty destructures, just think we should use an API that directly gives what we want when it's possible.

PR #22604, opened 2022-01-28
build: add an action for handling feature requests

josephperrott:
Add feature request voting process github actions to the repo.

### DIFF
--- a/.github/workflows/feature-requests.yml
+++ b/.github/workflows/feature-requests.yml
@@ -1,0 +1,17 @@
+name: Feature request triage bot
+
+on:
+  schedule:
+    # Run at 13:00 every day
+    - cron: '0 13 * * *'
+
+jobs:
+  feature_triage:
+    # To prevent this action from running in forks, we only run it if the repository is exactly the
+    # angular/angular-cli repository.
+    if: github.repository == 'angular/angular-cli'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: angular/dev-infra/github-actions/feature-request@354871956801929457d9f08482fb81f7012ed7c8
+        with:
+          angular-robot-key: ${{ secrets.ANGULAR_ROBOT_PRIVATE_KEY }}


.github/workflows/feature-requests.yml
```diff
@@ -0,0 +1,15 @@
+name: Feature request triage bot
+
+on:
+  schedule:
+    # Run at 13:00 every day
+    - cron: '0 13 * * *'
+
+jobs:
+  feature_triage:
+    if: github.repository == 'angular/angular-cli'
```
dgp1130, line 10: Is the motivation for this to disable the workflow for forks? If so, we should probably put a comment calling that out as it isn't very obvious. Someone would reasonably thinking that this is always `true` and not consider the forking use case.

.github/workflows/feature-requests.yml
```diff
@@ -0,0 +1,15 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: angular/dev-infra/github-actions/feature-request@354871956801929457d9f08482fb81f7012ed7c8
```
dgp1130, line 13: Does this require manual version bumps whenever the bot changes? I'm wondering if there's a way to more easily depend on the latest version or at least a release tag?

dgp1130 (approved):

.github/workflows/feature-requests.yml
josephperrott: This is automatically updated by renovate.

.github/workflows/feature-requests.yml
josephperrott: Added.

.github/workflows/feature-requests.yml
dgp1130: Gotcha, if Renovate handles it, then that should be good enough.

josephperrott: It is the process used for keeping the other dev-infra based actions up to date, so we have already proved its success

PR #22612, opened 2022-01-31
fix(@angular-devkit/build-angular): ensure to use content hash as filenames hashing mechanism

alan-agius4:
Previously we used hash which resulted in a unique hash generated for every build even when the contents of the files didn't differ.

More info: https://webpack.js.org/guides/caching/#output-filenames

NB: I did try to add a unit test, but it non trivial.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/build-options.ts
+++ b/packages/angular_devkit/build_angular/src/utils/build-options.ts
@@ -16,6 +16,7 @@ import {
   IndexUnion,
   InlineStyleLanguage,
   Localize,
+  OutputHashing,
   ScriptElement,
   SourceMapClass,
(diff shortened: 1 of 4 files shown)


dgp1130 (approved):
What exactly was non-deterministic that is fixed? I don't see anything too obvious being changed, is it updating `.[hash:${length}]` -> `.[contenthash:${length}]`?

dgp1130:
I think this conflicted with another PR, I'd fix it in GitHub, but I think that will create a merge commit. Probably easier for you to fix when you get a chance @alan-agius4.

dgp1130:
Merged to patch in https://github.com/angular/angular-cli/pull/22623, so retargeting this PR to `major`.

Issue #17125, opened 2020-02-29
Allow .ts files in angular.json `scripts` to be transpiled to target ES version


denkan:
# 🚀 Feature request

### Command
- [x] build
- [x] serve

### Description
There are scenarios where we need to include a specific file to the bundle. Today, we can do this by using the `scripts` option in angular.json. But this only works for .js files. 

It would be great to be able to specify .ts files too, which then get transpiled into target ES version.  
This would e.g. allow us to include separate files which reads from environments.ts vars correctly.

**My specific scenario:**
I'm implementing [Firebase Messaging](https://firebase.google.com/docs/cloud-messaging/js/receive), which require `firebase-messaging-sw.js` in the root and contains config vars for `messagingSenderId` (which I have stored in environment.ts). 

### Describe the solution you'd like
When specifying an input with .ts extension, the builder transpiles it into .js.
```json
"scripts": [
  { "input": "src/firebase-messaging-sw.ts", "inject": false, "bundleName": "firebase-messaging-sw" }
]
```

### Describe alternatives you've considered
[This solution ](https://stackoverflow.com/questions/55609495/how-to-compile-a-ts-file-to-js-in-angular-and-make-it-accessible-in-root-when-se) suggests to transpile the .ts file before the Angular build (`tsc my-file.ts && ng build`). 
But then we won't access the correct environment.ts, as it gets determent in the Angular build...

clydin:
The `scripts` option is intended to be functionally equivalent to adding a `script` element to the index HTML file for an application.  It is primarily a convenience option to remove the need for developers to manage multiple index HTML files within a single application solely to support multiple configurations with differing `script` element requirements.

To allow for the behavior as described in the above issue, the `scripts` option would need to be changed into what would essentially be a miniature application builder.  Not only would TypeScript processing be necessary but also bundling (with the potential for ES target downleveling and other post-processing steps).  Additional script options for specifying a `tsconfig` and bundling options would be needed on a per script level as well to support the requirements of the script.  This unfortunately would be a large increase in complexity and is currently out of scope for this option.  If generating service worker code becomes common, then a standalone builder and a separate project within the Angular workspace may be the most beneficial long-term solution for this problem space.

flash-me:
Hi @denkan 

why not manually transpiling your TypeScript files with `tsc` and appropriate `tsconfig.json`?

- Create a dedicate `tsconfig.scripts.json`
- Put your required files into to `files` array in the config
[...]

PR #22613, opened 2022-01-31
feat(@angular-devkit/build-angular): validate file extensions for `scripts` and `styles` options

alan-agius4:
In some cases unexpected files may be provided which can cause to unsupported or broken behaviour.

One such use-case is users can provide TypeScript files as `scripts` input, this would not be processed by the TypeScript compiler, see: https://github.com/angular/angular-cli/issues/17125 and would cause the build to fail with a an unhelpful error message during optimization as the JS optimizers cannot parse TypeScript input.

BREAKING CHANGE:

Browser builder `script` and `styles` options input files extensions are now validated.

Valid extensions for `scripts` are:
- `.js`
- `.cjs`
- `.mjs`
- `.jsx`
- `.cjsx`
- `.mjsx`

Valid extensions for `styles` are:
- `.css`
- `.less`
- `.sass`
- `.scss`
- `.styl`

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser/schema.json
+++ b/packages/angular_devkit/build_angular/src/builders/browser/schema.json
@@ -29,7 +29,35 @@
       "type": "array",
       "default": [],
       "items": {
-        "$ref": "#/definitions/extraEntryPoint"
+        "oneOf": [
+          {
+            "type": "object",
+            "properties": {
+              "input": {
+                "type": "string",
+                "description": "The file to include.",
+                "pattern": "\\.[cm]?jsx?$"
+              },
+              "bundleName": {
+                "type": "string",
+                "pattern": "^[\\w\\-.]*$",
+                "description": "The bundle name for this extra entry point."
+              },
+              "inject": {
+                "type": "boolean",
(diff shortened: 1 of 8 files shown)


packages/angular_devkit/build_angular/src/builders/browser/schema.json
```diff
@@ -29,15 +29,71 @@
...
       "items": {
-        "$ref": "#/definitions/extraEntryPoint"
+        "oneOf": [
+          {
+            "type": "object",
+            "properties": {
+              "input": {
+                "type": "string",
+                "description": "The file to include.",
+                "pattern": "\\.[cm]?[j]sx?$"
```
dgp1130, line 39: Is this supposed to be `\\.[cm]?[jt]sx?$`? Does it allow TypeScript? If not, could we simplify `[j]` to just `j`?

Same point in a couple other places in the PR.

dgp1130 (approved):

packages/angular_devkit/build_angular/src/builders/browser/schema.json
alan-agius4: No, TypeScript files are not allowed. Updated the regexp to remove the square braces.

PR #22625, opened 2022-02-01
docs(@angular/cli): update `@angular/cli` package readme

alan-agius4:
Previously, the `@angular/cli` readme contained a lot of un-needed information for application users


dgp1130 (approved):
I'm wondering if we should simplify `packages/angular/cli/README.md` even more by just pointing to the root README. I don't think we want application devs looking at that README. It makes sense to move some of the content out of it, since people could land here accidentally, but we'll definitely forget to update this in the future and it will fall out of date.

If there's anything useful to discuss about the `@angular/cli` package specifically, I think that would go in this README, or a nearby doc with a different name so application devs don't get confused. I'm not sure that we have any useful content for such a doc though?

alan-agius4:
> I'm wondering if we should simplify `packages/angular/cli/README.md` even more by just pointing to the root README. I don't think we want application devs looking at that README. It makes sense to move some of the content out of it, since people could land here accidentally, but we'll definitely forget to update this in the future and it will fall out of date.
> 
> If there's anything useful to discuss about the `@angular/cli` package specifically, I think that would go in this README, or a nearby doc with a different name so application devs don't get confused. I'm not sure that we have any useful content for such a doc though?

Agreed and updated.

PR #22633, opened 2022-02-02
fix(@ngtools/webpack): support locating PNPM lock file during NGCC processing

### LINKED ISSUE #22632: @ngtools/webpack expects yarn or npm, but not pnpm
Hi!

`@ngtools/webpack` performs a ngcc run check to determine if an initial execution is required. It looks for `yarn.lock` or `package-lock.json` to run this check. But our project uses [pnpm](https://pnpm.io/), which has `pnpm-lock.yaml`. And `@ngtools/webpack` thinks it is first time build and runs 'ngcc'.

We are using version `12.2.14`, but I see this code still [there](https://github.com/angular/angular-cli/blob/master/packages/ngtools/webpack/src/ngcc_processor.ts): 

```js
    try {
      let lockData;
      let lockFile = 'yarn.lock';
      try {
        lockData = readFileSync(path.join(projectBasePath, lockFile));
      } catch {
        lockFile = 'package-lock.json';
        lockData = readFileSync(path.join(projectBasePath, lockFile));
      }

      let ngccConfigData;
      try {
        ngccConfigData = readFileSync(path.join(projectBasePath, 'ngcc.config.js'));
      } catch {
        ngccConfigData = '';
      }
```

alan-agius4:
Previously, we only handled Yarn and NPM lock files even though PNPM is supported as package manager.

The lock file is used to perform checks to determine if an initial execution of NGCC was already performed.

Closes #22632

### DIFF
--- a/packages/ngtools/webpack/src/ngcc_processor.ts
+++ b/packages/ngtools/webpack/src/ngcc_processor.ts
@@ -73,15 +73,6 @@ export class NgccProcessor {
     const runHashBasePath = path.join(this._nodeModulesDirectory, '.cli-ngcc');
     const projectBasePath = path.join(this._nodeModulesDirectory, '..');
     try {
-      let lockData;
-      let lockFile = 'yarn.lock';
-      try {
-        lockData = readFileSync(path.join(projectBasePath, lockFile));
-      } catch {
-        lockFile = 'package-lock.json';
-        lockData = readFileSync(path.join(projectBasePath, lockFile));
-      }
-
       let ngccConfigData;
       try {
         ngccConfigData = readFileSync(path.join(projectBasePath, 'ngcc.config.js'));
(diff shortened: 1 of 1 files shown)


packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -248,6 +240,24 @@ export class NgccProcessor {
...
+
+      try {
+        return {
+          lockFilePath,
+          lockFileData: readFileSync(lockFilePath),
+        };
+      } catch {}
+    }
+
+    throw new Error('Cannot locate a package manager lock file.');
```
dgp1130, line 259: Should we list the files we tried in this error message? Otherwise it's hard to know if the CLI is looking in the wrong place or if the file itself was misplaced.

dgp1130 (approved):

packages/ngtools/webpack/src/ngcc_processor.ts
alan-agius4: Actually this error message is never shown to the users. As this is non fatal and the exception is caught further down

dgp1130: Up to you, could be useful for local development if it's ever thrown for an unexpected reason.

PR #22634, opened 2022-02-02
Use PNPM as package manager when `pnpm-lock.yaml` exists

alan-agius4:
**refactor(@angular/cli): remove support for legacy Angular CLI version 1 configurations**
Remove references to legacy Angular CLI version 1 configurations. By now users should have been migrated to use the new configuration.

**feat(@angular/cli): use PNPM as package manager when pnpm-lock.yaml exists**
While supported, we didn't automatically try to determine if PNPM was used through the lock files like we do for other package managers.

### DIFF
--- a/packages/angular/cli/utilities/config.ts
+++ b/packages/angular/cli/utilities/config.ts
@@ -10,6 +10,7 @@ import { json, workspaces } from '@angular-devkit/core';
 import { existsSync, readFileSync, statSync, writeFileSync } from 'fs';
 import * as os from 'os';
 import * as path from 'path';
+import { PackageManager } from '../lib/config/workspace-schema';
 import { findUp } from './find-up';
 import { JSONFile, readAndParseJson } from './json-file';
 
@@ -298,18 +299,19 @@ export function getProjectByCwd(workspace: AngularWorkspace): string | null {
   return null;
 }
 
-export async function getConfiguredPackageManager(): Promise<string | null> {
-  const getPackageManager = (source: json.JsonValue | undefined): string | undefined => {
+export async function getConfiguredPackageManager(): Promise<PackageManager | null> {
+  const getPackageManager = (source: json.JsonValue | undefined): PackageManager | null => {
     if (isJsonObject(source)) {
       const value = source['packageManager'];
       if (value && typeof value === 'string') {
-        return value;
+        return value as PackageManager;
       }
     }
+
+    return null;
   };
 
-  let result: string | undefined | null;
-
+  let result: PackageManager | null = null;
   const workspace = await getWorkspace('local');
   if (workspace) {
     const project = getProjectByCwd(workspace);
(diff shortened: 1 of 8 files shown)


packages/angular/cli/utilities/package-manager.ts
```diff
@@ -6,55 +6,87 @@
...
+import { promisify } from 'util';
 import { PackageManager } from '../lib/config/workspace-schema';
 import { getConfiguredPackageManager } from './config';
 
-function supports(name: string): boolean {
+const execPromise = promisify(exec);
+async function supports(name: PackageManager): Promise<boolean> {
   try {
-    execSync(`${name} --version`, { stdio: 'ignore' });
+    await execPromise(`${name} --version`);
```
dgp1130, line 20: Nit: I suggest naming the variable you want to use the most intuitive name `exec`, and name the other one with the qualifier `execCallback`, that way future changes to this file will naturally use the correct symbol.

The counterpoint for this particular case is that `exec` is a fairly well-known Node API, so having that return a `Promise` can be counter-intuitive for some. Ultimately up to you, just a suggestion.

dgp1130, line 20: Nit: Would it be better to look to see if the program is on the path rather than executing it? It's hard to predict what the program will do and could be unnecessarily slow or lead to side effects (ie. updating the `package-lock.json`, prompting a login). We also don't know that this is actually NPM/Yarn/pnpm, it could be anything the user has configured with that name.

I don't know if `isInPathEnv('npm')` exists, but might be worth checking if that can be a better alternative. If not, this is fine as is, I admit this concern is pretty edge-casey.

packages/angular/cli/utilities/config.ts
```diff
@@ -385,30 +381,6 @@ export function migrateLegacyGlobalConfig(): boolean {
   return false;
 }
 
-// Fallback, check for packageManager in config file in v1.* global config.
-function getLegacyPackageManager(): string | null {
```
dgp1130, line 389: Was this intended to be removed in the other commit? Not that big a deal to be here, just making sure it was intentional.

dgp1130 (approved):
Thanks for cleaning up the v1 config stuff, I didn't realize we still had that lying around.

packages/angular/cli/utilities/config.ts
alan-agius4: Yeah

packages/angular/cli/utilities/package-manager.ts
alan-agius4: Let's rename it to `exec`

alan-agius4: You definitly made some good points here, although I don't think there is a Node.js similar to `isInPathEnv`.

PR #22638, opened 2022-02-02
build: flip build script to bazel

kormide:
This flips the cli build to use bazel, replacing the legacy build script with the bazel one.

Looking for careful feedback on the second commit where I stop running bazel clean before a release build. This is something I originally copied over from the angular project, but I can no longer reproduce the issue locally. I think that my use of `config_setting`s and `select`s with the release artifacts causes the build to correctly stamp artifacts without the need for a clean... but someone who understands this better should take a look.


alan-agius4:
I am seeing that the mode is release, so we want to have that for CI?

kormide:
> I am seeing that the mode is release, so we want to have that for CI?

The ci currently calls `yarn build` which defaults to a release build, so I was just preserving that. I could change it to a local or snapshot build if preferred.

kormide:
> I am seeing that the mode is release, so we want to have that for CI?
> 
> This would cause `bazel clean` to get invoked
> 
> https://github.com/angular/angular-cli/blob/82971c786f5ee160604115266968343cb015f4ed/scripts/build-bazel.ts#L125

I remove that in the second commit.

.circleci/config.yml
```diff
@@ -307,6 +323,8 @@ jobs:
           keys:
             - *cache_key_win
       - run: yarn install --frozen-lockfile --cache-folder ../.cache/yarn
+      #TEMP: test
```
josephperrott, line 326: Do we still need this line?  I imagine so if we are wanting to build windows via bazel here as well

.circleci/config.yml
```diff
@@ -228,6 +235,9 @@ jobs:
     resource_class: medium
     steps:
       - custom_attach_workspace
+      - install_python
+      - setup_bazel_rbe
+      - run: sudo cp .circleci/bazel.rc /etc/bazel.bazelrc
```
josephperrott, line 240: Can this step just be added to the `setup_bazel_rbe` command?  This would prevent us having to repeat it in all of the spots using rbe

.circleci/config.yml
kormide: I'm calling bazelisk/bazel directly to work around a strange issue that the angular project also encountered. At the moment I'm just pushing a bunch of tests to try to get the CI working for windows tests.

kormide:
I'm going to close this for now. I realized that we can't remove the legacy build script until we have the integration tests also running with bazel. Will bring this back when ready.

PR #22646, opened 2022-02-04
refactor(@angular-devkit/build-angular): remove workaround for fidelity and performance of babel sourcemaps

alan-agius4:
With this change we remove the workaround for fidelity and performance of babel sourcemaps as this is no longer needed as Babel now uses `@ampproject/remapping` to merge sourcemaps.

See https://github.com/babel/babel/pull/14209 for more context.


dgp1130 (approved):
Awesome! Glad to see this optimization got upstreamed.

PR #22670, opened 2022-02-08
fix(@angular-devkit/build-angular): block Karma from starting until build is complete

### LINKED ISSUE #22495: Chrome/karma is started before the webpack dev server is ready to serve
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

No, AFAIK, this has always been present

### Description

There is this issue in Karma: https://github.com/karma-runner/karma/issues/3745

What happens is that Angular CLI is starting karma and the webpack dev server at the same time. On large Angular projects, like ours, Angular + webpack is very slow (1-2 minutes to start `ng serve` for example, separate issue we really need to address somehow). 

What seems to occur is that chrome (or ChromeHeadless) gives up trying to load the page (it would see the problem that the server is not responding). Karma then kills the browser and restarts it (triggering issue https://github.com/karma-runner/karma/issues/3745). Eventually the webpack dev server starts responding, and the tests run. Once the tests complete, the coverage report (and potentially other reporters) fail to write their reports.

While there is a bug in Karma that needs to be fixed, it does not seem to make sense to start the test framework until the webpack dev server is ready to serve the application. When it is ready, we see the `Browser application bundle generation complete` printed to the terminal.

Ideally karma would be started after a 200 HTTP response could be obtained from the webpack dev server

## 🔬 Minimal Reproduction

Only thing that is required is to run `ng test` on an Angular project. The problem to reproduce is that the application has to have enough code to cause webpack dev server to be too slow to start before chrome gives up loading the page.

## 🔥 Exception or Error

<pre><code>

</code></pre>

## 🌍 Your Environment

<pre><code>

Angular CLI: 13.1.2
Node: 14.17.1
Package Manager: npm 6.14.13
OS: darwin x64

Angular: 13.1.1
... router

@angular-devkit/architect       0.1301.2
@angular-devkit/build-angular   13.1.2
@angular-devkit/core            13.1.2
@angular-devkit/schematics      13.1.2
@angular/cli                    13.1.2
@ngtools/webpack                13.1.2
@schematics/angular             13.1.2
ng-packagr                      13.1.2
typescript                      4.5.4
webpack                         5.65.0
</code></pre>

**Anything else relevant?**

alan-agius4:
This change is inspired by https://github.com/karma-runner/karma-chrome-launcher/issues/154#issuecomment-986661937 which blocks Karma from launching the browsers until the compilation is complete.

This is needed especially for large applications when using code-coverage where otherwise the users would have to fine-tune several Karma timeouts such as  `captureTimeout` for various environments.

Closes #22495

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
@@ -169,6 +169,12 @@ const init: any = (config: any, emitter: any) => {
   compiler.hooks.watchRun.tapAsync('karma', (_: any, callback: () => void) => handler(callback));
   compiler.hooks.run.tapAsync('karma', (_: any, callback: () => void) => handler(callback));
 
+  webpackMiddleware = webpackDevMiddleware(compiler, webpackMiddlewareConfig);
+  emitter.on('exit', (done: any) => {
+    webpackMiddleware.close();
+    done();
+  });
+
   function unblock() {
     isBlocked = false;
     blocked.forEach((cb) => cb());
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
```diff
@@ -169,29 +169,42 @@ const init: any = (config: any, emitter: any) => {
...
-      // Refresh karma only when there are no webpack errors, and if the compilation changed.
-      lastCompilationHash = stats.hash;
-      emitter.refreshFiles();
-    }
-    unblock();
-  });
+  let isFirstRun = true;
+
+  return new Promise<void>((resolve) => {
+    compiler.hooks.done.tap('karma', (stats) => {
```
dgp1130, line 188: What if Karma fails? Will this callback still be executed or do we need a separate error handler to reject the `Promise`?

dgp1130 (approved):

packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
alan-agius4: I think you meant the webpack fails, yeah this will always be executed.

PR #22695, opened 2022-02-11
build: add permissions to current github action workflows

clydin:
The currently recommended best practice for Github action workflows is to set top-level permissions to read only. And if the job uses the automatic `GITHUB_TOKEN`, fine-grained permissions for each job based on the job's requirements should also be added.
All existing workflows in the repository now have top-level read only permission blocks.
Only the `scorecard` workflow currently requires additional job level permissions and the minimum set of permissions were already present for the job.

### DIFF
--- a/.github/workflows/dev-infra.yml
+++ b/.github/workflows/dev-infra.yml
@@ -1,5 +1,9 @@
 name: DevInfra
 
+# Declare default permissions as read only.
+permissions:
+  contents: read
+
 on:
   pull_request_target:
     types: [opened, synchronize, reopened]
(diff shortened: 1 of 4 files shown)


.github/workflows/dev-infra.yml
```diff
@@ -1,12 +1,19 @@
...
+permissions: read-all
+
 on:
   pull_request_target:
     types: [opened, synchronize, reopened]
 
 jobs:
   labels:
     runs-on: ubuntu-latest
+    permissions:
```
josephperrott, line 13: We would need to verify, but I think we don't actually need to set permissions here because we don't use the `GithubAction` account token.

clydin: That appears to be correct.  Permissions have been updated to reflect this.

josephperrott (approved):
LGTM

PR #22716, opened 2022-02-16
fix(@angular-devkit/build-angular): fix development mode i18n error f…

Zuckjet:
related to https://github.com/angular/angular/issues/45066#issuecomment-1041151690


alan-agius4:
This is working as intended. Only a single locale can be localized during development.

PR #22721, opened 2022-02-16
ci: use renovate Github Action tag version pinning

clydin:
Renovate supports use hashed version pinning for individual Github actions while still following SemVer-based tags.
All workflow actions external to the Angular organization now leverage this support to ensure both that stable versions of the actions are used and that the actions are pinned to a hashed version of the tag.


alan-agius4 (approved):

josephperrott (approved):
LGTM

PR #22749, opened 2022-02-22
fix(@angular-devkit/build-angular): find `ngswConfig` as relative to the application root

josephperrott:
Previously when determining the `configPath` for building the service worker, the config path
was loaded as a relative path to location where  execution of ng build occured.  Instead the
`ngswConfig` value provided in `angular.json` should be considered relative to th application
root, as is the case with the other paths provided in the configuration.

Before this change, assuming a project root of `/project`, and a `ngswConfig` value of `ngsw-config.json`,
the application built successfully when the exuction working directory was `/project`, but if the
user invoked ng build from `/project/src` for instance, the build failed as `ngsw-config.json` does not
exist at `/project/src/ngsw-config.json`.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/service-worker.ts
+++ b/packages/angular_devkit/build_angular/src/utils/service-worker.ts
@@ -69,10 +69,11 @@ export async function augmentAppWithServiceWorker(
 ): Promise<void> {
   const distPath = getSystemPath(normalize(outputPath));
 
-  // Determine the configuration file path
-  const configPath = ngswConfigPath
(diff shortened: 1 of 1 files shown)


clydin:
Good catch on this one.
Unfortunately, the change will probably need to be a little more complicated since all the paths in the `angular.json` file should be relative the the location of the `angular.json` file (that location is also the workspace root).  The default value for `ngswConfig` is `${appRoot}/ngsw-config.json` with that value then being relative to the workspace root (assuming the `appRoot` value has not already been resolved with the workspace root).

clydin:
Superseded by #23045

PR #22767, opened 2022-02-24
refactor(@angular-devkit/build-angular): replace most custom path `resolve` usages with Node.js builtins

clydin:
During the build initialization phase, many paths are converted back and forth between multiple normalized forms. These conversions involve potentially expensive string operations. The majority of the custom path `resolve` function from `@angular-devkit/core` usages have now been removed in favor of the Node.js builtin path functions. This change reduces the need to perform additional string manipulation where possible.


alan-agius4 (approved):
Nice!

PR #22803, opened 2022-03-04
feat(@angular/cli): drop support for Node.js 12

alan-agius4:
Node.js v12 will become EOL on 2022-04-30. As a result, Angular CLI v14 will no longer support Node.js v12.

BREAKING CHANGE:

Support for Node.js v12 has been removed as it will become EOL on 2022-04-30. Please use Node.js v14.15 or later.


packages/angular_devkit/core/node/host.ts
```diff
@@ -91,24 +91,7 @@ export class NodeJsAsyncHost implements virtualFs.Host<Stats> {
     return this.isDirectory(path).pipe(
       mergeMap(async (isDirectory) => {
         if (isDirectory) {
```
clydin, line 93: With the use of only `rm`, the `isDirectory` check should also be able to be removed.

.circleci/config.yml
```diff
@@ -339,7 +339,6 @@ workflows:
             - setup
       - e2e-cli:
           name: e2e-cli
-          nodeversion: '14.15'
```
clydin, line 342: With the package engine field using 14.15 as the minimum, I think we should consider keeping this node version for the E2E tests.

clydin (approved):

dgp1130 (approved):
Thanks for cleaning this up!

packages/angular_devkit/core/node/host.ts
alan-agius4: Done

PR #22815, opened 2022-03-09
feat(@angular/cli): add `--global` option to `ng analytics` command

alan-agius4:
With this change we add a `--global` option to `ng analytics` command.

BREAKING CHANGE:

Several changes to the `ng analytics` command syntax.

- `ng analytics project <setting>` has been replaced with `ng analytics <setting>`
- `ng analytics <setting>` has been replaced with `ng analytics <setting> --global`
---

feat(@angular/cli): add `ng analytics info` command

With this change we add a subcommand to `ng analytics`. This command can be used tp display analytics gathering and reporting configuration.

Example:
```
$ ng analytics info                              
Global setting: off
Local setting: on
Effective status: disabled
```

### DIFF
--- a/packages/angular/cli/src/analytics/analytics.ts
+++ b/packages/angular/cli/src/analytics/analytics.ts
@@ -63,10 +63,11 @@ export function isPackageNameSafeForAnalytics(name: string): boolean {
 
 /**
  * Set analytics settings. This does not work if the user is not inside a project.
- * @param level Which config to use. "global" for user-level, and "local" for project-level.
+ * @param global Which config to use. "global" for user-level, and "local" for project-level.
  * @param value Either a user ID, true to generate a new User ID, or false to disable analytics.
  */
-export function setAnalyticsConfig(level: 'global' | 'local', value: string | boolean) {
+export function setAnalyticsConfig(global: boolean, value: string | boolean): void {
+  const level = global ? 'global' : 'local';
   analyticsDebug('setting %s level analytics to: %s', level, value);
   const [config, configPath] = getWorkspaceRaw(level);
   if (!config || !configPath) {
@@ -79,6 +80,8 @@ export function setAnalyticsConfig(level: 'global' | 'local', value: string | bo
     throw new Error(`Invalid config found at ${configPath}. CLI should be an object.`);
   }
 
+  console.log(`Configured ${level} analytics to "${analyticsConfigValueToHumanFormat(value)}".`);
+
   if (value === true) {
     value = uuidV4();
   }
@@ -87,6 +90,18 @@ export function setAnalyticsConfig(level: 'global' | 'local', value: string | bo
   config.save();
 
   analyticsDebug('done');
+}
+
+export function analyticsConfigValueToHumanFormat(value: unknown): 'on' | 'off' | 'not set' | 'ci' {
+  if (value === false) {
+    return 'off';
+  } else if (value === 'ci') {
+    return 'ci';
+  } else if (typeof value === 'string' || value === true) {
+    return 'on';
+  } else {
+    return 'not set';
+  }
 }
 
 /**
@@ -110,7 +125,7 @@ export async function promptGlobalAnalytics(force = false) {
       },
     ]);
 
-    setAnalyticsConfig('global', answers.analytics);
+    setAnalyticsConfig(true, answers.analytics);
 
     if (answers.analytics) {
       console.log('');
@@ -169,7 +184,7 @@ export async function promptProjectAnalytics(force = false): Promise<boolean> {
       },
     ]);
 
-    setAnalyticsConfig('local', answers.analytics);
+    setAnalyticsConfig(false, answers.analytics);
 
     if (answers.analytics) {
       console.log('');
--- a/packages/angular/cli/src/analytics/analytics.ts
+++ b/packages/angular/cli/src/analytics/analytics.ts
@@ -118,7 +118,7 @@ export async function promptGlobalAnalytics(force = false) {
         Thank you for sharing anonymous usage data. If you change your mind, the following
         command will disable this feature entirely:
 
-            ${colors.yellow('ng analytics off')}
+            ${colors.yellow('ng analytics off --global')}
       `);
       console.log('');
 
@@ -177,7 +177,7 @@ export async function promptProjectAnalytics(force = false): Promise<boolean> {
         Thank you for sharing anonymous usage data. Should you change your mind, the following
         command will disable this feature entirely:
 
-            ${colors.yellow('ng analytics project off')}
+            ${colors.yellow('ng analytics off')}
       `);
       console.log('');
 
--- a/packages/angular/cli/src/analytics/analytics.ts
+++ b/packages/angular/cli/src/analytics/analytics.ts
@@ -15,6 +15,7 @@ import { getWorkspace, getWorkspaceRaw } from '../utilities/config';
 import { isTTY } from '../utilities/tty';
 import { VERSION } from '../utilities/version';
 import { AnalyticsCollector } from './analytics-collector';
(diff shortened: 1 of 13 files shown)


packages/angular/cli/src/commands/analytics/info/cli.ts
```diff
@@ -0,0 +1,50 @@
...
+  CommandModuleImplementation,
+  Options,
+} from '../../../command-builder/command-module';
+import { getWorkspaceRaw } from '../../../utilities/config';
+
+export class AnalyticsInfoCommandModule
+  extends CommandModule
+  implements CommandModuleImplementation
+{
+  command = 'info';
```
alan-agius4, line 23: I am not uncertain about the command to output the info `$ ng analytics info` vs `$ ng analytics`. I used the former as I found it to be more explicate.

packages/angular/cli/src/analytics/analytics.ts
```diff
@@ -220,7 +167,6 @@ export async function hasGlobalAnalyticsConfiguration(): Promise<boolean> {
  */
 export async function getGlobalAnalytics(): Promise<AnalyticsCollector | undefined> {
   analyticsDebug('getGlobalAnalytics');
-  const propertyId = AnalyticsProperties.AngularCliDefault;
 
   if ('NG_CLI_ANALYTICS' in process.env) {
     if (process.env['NG_CLI_ANALYTICS'] == 'false' || process.env['NG_CLI_ANALYTICS'] == '') {
```
alan-agius4, line 172: We we expect users to set this env variable to determine that we are running in CI. Why not use `CI` instead?

clydin (approved):

packages/angular/cli/src/commands/analytics/info/cli.ts
```diff
@@ -0,0 +1,50 @@
...
+    const localSetting = localWorkspace?.get(['cli', 'analytics']);
+
+    this.context.logger.info(tags.stripIndents`
+      Global setting: ${analyticsConfigValueToHumanFormat(globalSetting)}
+      Local setting: ${
+        this.context.workspace
+          ? analyticsConfigValueToHumanFormat(localSetting)
+          : 'No local workspace configuration file.'
+      }
+      Effective setting: ${analyticsConfigValueToHumanFormat(
```
dgp1130, line 44: I imagine some users might not understand what "effective" means in this context? "Local" is also a bit overloaded as "project" in other situations. We should define these terms in `ng analytics info` help content and then maybe when we print the status, we can include a link to the AIO page for `ng analytics` which would include that content.

packages/angular/cli/src/analytics/analytics.ts
dgp1130: I agree that `CI` would be more consistent. No one is going to go out of their way to set `NG_CLI_ANALYTICS=ci`, so I feel like this value is pretty meaningless as is. We should definitely fall back to `CI` if `NG_CLI_ANALYTICS` is not set.

packages/angular/cli/src/analytics/analytics.ts
```diff
@@ -246,57 +177,18 @@ export async function getGlobalAnalytics(): Promise<AnalyticsCollector | undefin
...
-  } catch (err) {
-    analyticsDebug('Error happened during reading of analytics config: %s', err.message);
-
-    return undefined;
-  }
+  return getAnalytics('global');
 }
 
-export async function hasWorkspaceAnalyticsConfiguration(): Promise<boolean> {
+export async function hasAnalyticsConfiguration(level: 'global' | 'local'): Promise<boolean> {
```
dgp1130, line 187: Nit: Could we simplify this implementation to `return !!(await getWorkspaceAnalytics());`? Could we replace calls to `hasAnalyticsConfiguration()` to just `!!(await getWorkspaceAnalytics());` and drop the function altogether? That might make some of the callers' logic simpler if they just try to get the analytics object and check if they actually got one.

packages/angular/cli/src/commands/analytics/info/cli.ts
alan-agius4: I find `enabled` to be rather similar to `enable` which might be confusing. Actually, I'd prefer if we the turn off/on the commands where `ng analytics enable` / `ng analytics disable`.


My reasoning to use `info` was that I future we might want to add additional details such as the uuid etc... and eventually we can use the same command syntax for `ng cache`. `ng cache info`,  `ng cache on` etc...

packages/angular/cli/src/commands/analytics/long-description.md
```diff
@@ -1,10 +1,9 @@
-The value of `setting-or-project` is one of the following.
+The value of `setting` is one of the following.
 
 - `on`: Enables analytics gathering and reporting for the user.
 - `off`: Disables analytics gathering and reporting for the user.
```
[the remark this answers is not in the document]
alan-agius4: This file should have been deleted since it's redundant now as the descriptions are part of each command.

packages/angular/cli/src/analytics/analytics.ts
alan-agius4: Yes, that can be done.

packages/angular/cli/src/commands/analytics/info/cli.ts
```diff
@@ -0,0 +1,50 @@
...
+        // If global is not disabled fallback to workspace.
+        globalSetting === false ? false : localSetting ?? globalSetting,
```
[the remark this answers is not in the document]
alan-agius4: Agreeded.

packages/angular/cli/src/commands/analytics/info/cli.ts
alan-agius4: I agree, although I'd prefer to leave this for a seperate PR and add the link then.

packages/angular/cli/src/analytics/analytics.ts
```diff
@@ -89,6 +92,18 @@ export function setAnalyticsConfig(level: 'global' | 'local', value: string | bo
   analyticsDebug('done');
 }
 
+export function analyticsConfigValueToHumanFormat(value: unknown): 'on' | 'off' | 'not set' | 'ci' {
```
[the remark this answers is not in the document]
alan-agius4: We actually already print a bunch of other information on enabling the analytics in stdout. 

https://github.com/angular/angular-cli/blob/92528b1df31a40770355ead7bbe76cf3fe3ccea7/packages/angular/cli/src/analytics/analytics.ts#L116-L123

I'll update these usages too.

packages/angular/cli/src/analytics/analytics.ts
alan-agius4: SGTM

packages/angular/cli/src/analytics/analytics.ts
alan-agius4: Actually checking the logic more, we cannot use `getWorkspaceAnalytics` instead of `hasAnalyticsConfiguration` has we need to differentiate between being disabled and unset in which `getWorkspaceAnalytics` will always return undefined in this case.

packages/angular/cli/src/commands/analytics/info/cli.ts
dgp1130: I'm fine with `info`. Agreed that it would be great to be consistent across subcommands if possible.

packages/angular/cli/src/analytics/analytics.ts
dgp1130: One option might be to have `getWorkspaceAnalytics()` return a discriminated union or return `Analytics | 'disabled' | 'unset'`.

It's up to you, I'm fine with `hasWorkspaceAnalyticsConfiguration()` if that's the best approach, just suggesting an alternative to consider.

PR #22817, opened 2022-03-09
refactor(@angular/cli): always use posix separator in `longDescriptionRelativePath`

alan-agius4:
Needed for consistency that will fix a Windows CI failure


josephperrott (approved):
LGTM

PR #22841, opened 2022-03-14
build: update webpack to 5.70.0

### LINKED ISSUE #22839: HMR not working properly

core

### Is this a regression?

Yes

### Description

I cannot get HMR to work properly. It always gives me the following error in the browser console.

browser console

I noticed that for some reason on every save of any file there are two builds, which would at least explain why the file cannot be found. It looks like this in the terminal:

terminal

_No response_

_No response_

```true
Angular CLI: 13.2.6
Node: 16.14.0
Package Manager: npm 8.3.1
OS: darwin x64

Angular: 13.2.6
... forms, material, platform-browser, platform-browser-dynamic
... router
```

### Anything else?

_No response_

alan-agius4:
This brings a couple of fixes to HMR, one of which is double reloaded.
See: https://github.com/webpack/webpack/releases/tag/v5.70.0

Closes #22839

### DIFF
--- a/package.json
+++ b/package.json
@@ -218,7 +218,7 @@
     "typescript": "4.6.2",
     "verdaccio": "5.5.1",
     "verdaccio-auth-memory": "^10.0.0",
-    "webpack": "5.67.0",
+    "webpack": "5.70.0",
     "webpack-dev-middleware": "5.3.0",
     "webpack-dev-server": "4.7.3",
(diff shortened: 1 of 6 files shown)


alan-agius4:
Confirmed in https://github.com/angular/angular-cli/issues/22839#issuecomment-1067835082 that this fixed the HMR issue.

clydin (approved):

PR #22848, opened 2022-03-16
refactor(@angular/cli): remove `hasAnalyticsConfig` analytics logic

alan-agius4:
With this change we clean up further the analytics code and re-use the `getAnalytics` to determine if the config is set or not.

Also, this change inclused a refactor to the `createAnalytics` method to make it more readable.

### DIFF
--- a/packages/angular/cli/src/analytics/analytics.ts
+++ b/packages/angular/cli/src/analytics/analytics.ts
@@ -152,16 +152,21 @@ export async function promptAnalytics(global: boolean, force = false): Promise<b
 
 /**
  * Get the analytics object for the user.
+ *
+ * @returns
+ * - `AnalyticsCollector` when enabled.
+ * - `analytics.NoopAnalytics` when disabled.
+ * - `undefined` when not configured.
  */
 export async function getAnalytics(
   level: 'local' | 'global',
-): Promise<AnalyticsCollector | undefined> {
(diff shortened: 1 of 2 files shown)


packages/angular/cli/src/analytics/analytics.ts
```diff
@@ -231,33 +238,49 @@ export async function createAnalytics(
...
-  const isDisabledGlobally = (await getWorkspace('global'))?.getCli()['analytics'] === false;
-  // If in workspace and global analytics is enabled, defer to workspace level
-  if (workspace && !isDisabledGlobally) {
-    const skipAnalytics = skipPrompt || analyticsDisabled;
-    // TODO: This should honor the `no-interactive` option.
-    //       It is currently not an `ng` option but rather only an option for specific commands.
-    //       The concept of `ng`-wide options are needed to cleanly handle this.
-    if (!skipAnalytics && !(await hasAnalyticsConfig('local'))) {
-      await promptAnalytics(false);
+  // Global config take precedence over local config only for the disabled check.
```
dgp1130, line 241: Typo: Global config *takes* precedence...

dgp1130 (approved):

packages/angular/cli/src/analytics/analytics.ts
alan-agius4: Updated

PR #22851, opened 2022-03-16
feat(@angular/cli): add `ng cache` command

alan-agius4:
With this change we create a new command `ng cache` that can be used control and check the disk cache settings.

This command has 4 subcommands
 - `ng cache enable` which can be used to enable the cache.
 - `ng cache disable` which can be used to disable the cache.
 - `ng cache clean` which can be used to delete the cache from disk.
 - `ng cache info` which will print statistics and information about the cache.


alan-agius4:
I haven't created sub-commands to change the cache environment and path. As changing this is more of a one of thing which can already be done through `ng config`. We can always reconsider this if users demand friendlier commands to change these 2 settings.

packages/angular/cli/src/utilities/environment-options.ts
```diff
@@ -14,9 +14,23 @@ function isPresent(variable: string | undefined): variable is string {
   return typeof variable === 'string' && variable !== '';
 }
 
+function isEnabled(variable: string): boolean {
+  return variable === '1' || variable.toLowerCase() === 'true';
```
dgp1130, line 18: This means that `VAR=2` or `VAR=test` would be considered falsey, is that the behavior we want? I imagine it could trip some people up.

I wonder if it might more intuitive to define `isDisabled()` as `variable === '0' || variable === '' || variable.toLowerCase() === 'false'` and then make `isEnabled()` negate it? That way it would follow the truthiness semantics of JS and other languages.

Alternatively it might be a better idea to actually limit allowed inputs to `true` / `false` / `0` / `1` and hard error on anything else. That probably requires more documentation, but might give a cleaner abstraction.

packages/angular/cli/src/utilities/environment-options.ts
```diff
@@ -14,9 +14,23 @@ function isPresent(variable: string | undefined): variable is string {
...
+
+const ciVariable = process.env['CI'];
+export const isCI = isPresent(ciVariable) && isEnabled(ciVariable);
+
+const disableVersionCheckVariable = process.env['NG_DISABLE_VERSION_CHECK'];
+export const disableVersionCheck =
+  isPresent(disableVersionCheckVariable) && isEnabled(disableVersionCheckVariable);
+
+const ngDebugEnvVariable = process.env['NG_DEBUG'];
+export const ngDebug = isPresent(ngDebugEnvVariable) && isEnabled(ngDebugEnvVariable);
```
dgp1130, line 36: We seem to be doing `isPresent() && isEnabled()` a lot, would it be worth some kind of `envToBoolean('NG_DEBUG')` that just converts the input variable to a regular boolean? That sounds like the utility we actually want here.

packages/angular/cli/src/analytics/analytics.ts
```diff
@@ -12,10 +12,10 @@ import * as inquirer from 'inquirer';
 import { v4 as uuidV4 } from 'uuid';
 import { colors } from '../utilities/color';
 import { getWorkspace, getWorkspaceRaw } from '../utilities/config';
+import { analyticsDisabled, analyticsShareDisabled } from '../utilities/environment-options';
```
dgp1130, line 15: Nit: Consider moving the environment options changes into a separate refactor commit.

packages/angular/cli/src/commands/cache/clean/cli.ts
```diff
@@ -0,0 +1,37 @@
...
+import {
+  CommandModule,
+  CommandModuleImplementation,
+  CommandScope,
+} from '../../../command-builder/command-module';
+import { getCacheConfig } from '../utilities';
+
+export class CacheCleanModule extends CommandModule implements CommandModuleImplementation {
+  command = 'clean';
+  describe = 'Deletes cache from disk.';
```
dgp1130, line 20: Nit: "cache" is kind of vague, maybe "build cache" or "disk cache" would more clearly articulate what we mean here? I'm mainly concerned that devs might misinterpret this as the NPM cache in `node_modules/`.

If you do change this, don't forget to also update the other descriptions to use consistent language.

packages/angular/cli/src/commands/cache/info/cli.ts
```diff
@@ -0,0 +1,104 @@
...
+    `);
+  }
+
+  private async getSizeOfDirectory(path: string): Promise<string> {
+    const directoriesStack = [path];
+    let size = 0;
+
+    while (directoriesStack.length) {
+      const dirPath = directoriesStack.pop();
+      if (dirPath === undefined) {
```
dgp1130, line 49: Why is this necessary? Doesn't `while (directoriesStack.length)` already handle an empty stack? Is there a case where the stack is not empty but has an `undefined`?

packages/angular/cli/src/commands/cache/info/cli.ts
```diff
@@ -0,0 +1,104 @@
...
+
+      for (const entry of entries) {
+        const entryPath = join(dirPath, entry);
+        const stats = await fs.stat(entryPath);
+
+        if (stats.isDirectory()) {
+          directoriesStack.push(entryPath);
+        }
+
+        size += stats.size + stats.blksize;
```
dgp1130, line 66: Isn't `blksize` the block size? Is `stats.size` enough? Otherwise, wouldn't it be `stats.blocks * stats.blksize`?

dgp1130, line 66: Also, it looks like this is run for directories too. I guess directories do have a file size, though it's pretty trivial. Just want to make sure that was intentional.

packages/angular/cli/src/commands/cache/info/cli.ts
```diff
@@ -0,0 +1,104 @@
...
+    return this.formatSize(size);
+  }
+
+  private formatSize(size: number): string {
+    if (size <= 0) {
+      return '0 bytes';
+    }
+
+    const abbreviations = ['bytes', 'kB', 'MB', 'GB'];
+    const index = Math.floor(Math.log(size) / Math.log(1024));
```
dgp1130, line 79: Clever way of picking an index, I never thought about using a logarithm this way. I think this deserves a comment to explain that it's doing log<sub>1024</sub>(size) and why that lands on the correct order of magnitude.

packages/angular/cli/src/commands/cache/info/cli.ts
```diff
@@ -0,0 +1,104 @@
...
+    const roundedSize = size / Math.pow(1024, index);
+    // bytes don't have a fraction
+    const fractionDigits = index === 0 ? 0 : 2;
+
+    return `${roundedSize.toFixed(fractionDigits)} ${abbreviations[index]}`;
+  }
+
+  private effectiveEnabledStatus(): 'enabled' | 'disabled' {
```
dgp1130, line 87: Nit: I suggest returning a boolean and letting the caller doe the `result ? 'enabled' : 'disabled'` to decouple the view logic from the implementation a little cleaner.

That can also allow each case to just return the right value so you don't need `enabledStatus`.

packages/angular/cli/src/commands/cache/long-description.md
```diff
@@ -0,0 +1,47 @@
...
+To amend the default cache settings, add the `cli.cache` object to your [Workspace Configuration](guide/workspace-config).
+The object goes under `cli.cache` at the top level of the file, outside the `projects` sections.
+
+```json
+{
+  "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
+  "version": 1,
+  "cli": {
+    "cache": {
+      ...
```
dgp1130, line 14: Nit: I believe GitHub supports `jsonc` as a code block language which is read as JSON with comments. That would allow `// ...` to be interpreted as a real comment and avoid the ugly red syntax error.

packages/angular/cli/src/commands/cache/long-description.md
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "projects": {}
+}
+```
+
+For more information, see [cache options](guide/workspace-config#cache-options).
+
+### Cache environments
+
+By default, disk cache is only enabled for local environments.
```
dgp1130, line 25: Nit: A short definition of "environment" might be useful here. It's not clear what other environments there are, and we already use the term "environment" for dev / prod configuration.

packages/angular/cli/src/commands/cache/long-description.md
```diff
@@ -0,0 +1,47 @@
...
+The Angular CLI checks for the presence and value of the `CI` environment variable to determine in which environment it is running.
+
+</div>
+
+### Cache path
+
+By default, `.angular/cache` is used as a base directory to store cache results. To change this path, run the following command:
+
+```bash
+ng config cli.cache.path ".cache/ng"
```
dgp1130, line 46: Nit: The explanation implies that this can be changed to any path but the example lists a specific path. We should make these align by either:
1. Making the lead-in specific: "For example, to change this path to `.cache/ng`, run the following command:"
2. Making the command generic: `ng config cli.cache.path "${NEW_PATH}"`

tests/legacy-cli/e2e/tests/commands/cache/cache-info.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Should be enabled by default for local builds.
+    process.env['CI'] = '0';
+    await configureCacheOptions();
+    await execAndWaitForOutputToMatch(
+      'ng',
+      ['cache', 'info'],
+      /Effective status on current machine: enabled/,
+    );
+
+    // Should be enabled by default for CI builds.
```
dgp1130, line 17: The expectation is for this to be "disabled", should the comment be updated?

tests/legacy-cli/e2e/tests/commands/cache/cache-info.ts
```diff
@@ -0,0 +1,63 @@
...
+      'ng',
+      ['cache', 'info'],
+      /Effective status on current machine: disabled/,
+    );
+  } finally {
+    process.env['CI'] = originalCIValue;
+  }
+}
+
+async function configureCacheOptions(cacheOptions?: {
```
dgp1130, line 55: Should this clear/set the `CI` variable? I think it would be easy to forget to reset it between test cases and leak that state accidentally.

tests/legacy-cli/e2e/tests/commands/cache/cache-info.ts
```diff
@@ -0,0 +1,63 @@
...
+    process.env['CI'] = '1';
+    await configureCacheOptions({ enabled: true });
+    await execAndWaitForOutputToMatch(
+      'ng',
+      ['cache', 'info'],
+      /Effective status on current machine: disabled/,
+    );
+
+    // Effective status should be enabled when 'environment' is set to 'all' or 'ci'.
+    await configureCacheOptions({ environment: 'all' });
```
dgp1130, line 27: Should this test with `CI` set and not?

dgp1130 (commented):
> I haven't created sub-commands to change the cache environment and path. As changing this is more of a one of thing which can already be done through `ng config`. We can always reconsider this if users demand friendlier commands to change these 2 settings.

+1. I think `ng config` is fine for now, if it ever becomes more complicated than a single command, we can re-evaluate an `ng cache` subcommand.

Would @clydin's input also be helpful here?

packages/angular/cli/src/commands/cache/info/cli.ts
clydin: `pop()` has a return type of `T | undefined`.  The conditional was probably for TypeScript to not error.  Two alternatives would be a `while (true)` (downside of an extra pop call) or `const dirPath = directoriesStack.pop()!;` (downside of an eslint ignore comment).

packages/angular/cli/src/utilities/environment-options.ts
alan-agius4: This is actually the same code used in other devkit packages. So if we want to change this, I think we should re-evaluate all the packages and do it in a seperate PR.

packages/angular/cli/src/utilities/environment-options.ts
alan-agius4: Yeah that should like a good idea.

packages/angular/cli/src/commands/cache/clean/cli.ts
alan-agius4: Agreed.

packages/angular/cli/src/commands/cache/info/cli.ts
alan-agius4: Indeed the conditional is for TypeScript. 

`while (true)` would still require an eslint ignore comment. So I'll go with a non null assertion.

packages/angular/cli/src/commands/cache/info/cli.ts
alan-agius4: Yes, this is intentional.

packages/angular/cli/src/commands/cache/info/cli.ts
alan-agius4: This is actually a copy of another function in build-angular.

packages/angular/cli/src/commands/cache/info/cli.ts
alan-agius4: Yeah size should be enough here.

packages/angular/cli/src/commands/cache/long-description.md
alan-agius4: Added more info.

packages/angular/cli/src/commands/cache/long-description.md
alan-agius4: Updated.

tests/legacy-cli/e2e/tests/commands/cache/cache-info.ts
alan-agius4: Yes

tests/legacy-cli/e2e/tests/commands/cache/cache-info.ts
alan-agius4: All tests here apart from the first one test with `process.env['CI'] = '1';`

packages/angular/cli/src/utilities/environment-options.ts
dgp1130: Hmm, I agree we should be consistent at least. At minimum they should probably share this code so we don't accidentally decouple them. Let's leave it for now and we can reevalute if we want to move forward with changing it after this PR.

packages/angular/cli/src/utilities/environment-options.ts
dgp1130: Yeah good point, encoding the default value somehow would be useful. A couple other ways of doing that could be either `envToBoolean('NG_DEBUG', false /* default */)` or making `envToBoolean()` return `undefined` if not present and then using `??` like `envToBoolean('NG_DEBUG') ?? false`. Up to you exactly how you want to handle it.

packages/angular/cli/src/commands/cache/info/cli.ts
dgp1130: Ah gotcha, thanks for the clarification. Agreed that the assertion is at least clear that "This shouldn't happen, but we have to tell types that".

packages/angular/cli/src/utilities/environment-options.ts
```diff
@@ -0,0 +1,34 @@
...
+export const analyticsShareDisabled = isDisabled(analyticsShareVariable);
+
+const ciVariable = process.env['CI'];
+export const isCI = isEnabled(ciVariable);
+
+const disableVersionCheckVariable = process.env['NG_DISABLE_VERSION_CHECK'];
+export const disableVersionCheck = isEnabled(disableVersionCheckVariable);
+
+const ngDebugEnvVariable = process.env['NG_DEBUG'];
+export const ngDebug = isEnabled(ngDebugEnvVariable);
```
dgp1130, line 34: I'm fine with this approach, I agree it's a lot cleaner.

One minor nit: Consider inlining the `*Variable` values, now that they are only referenced once each.

packages/angular/cli/src/utilities/environment-options.ts
alan-agius4: I updated the usage to remove the `isPresent`, PTAL. I find it clearer to use `isEnabled` and `isDisabled`.

tests/legacy-cli/e2e/tests/commands/cache/cache-info.ts
```diff
@@ -0,0 +1,65 @@
...
+      /Effective status on current machine: enabled/,
+    );
+
+    // Effective status should be disabled when 'enabled' is set to false
+    await configureTest('1' /** envCI */, { environment: 'all', enabled: false });
+    await execAndWaitForOutputToMatch(
+      'ng',
+      ['cache', 'info'],
+      /Effective status on current machine: disabled/,
+    );
```
dgp1130, line 46: Should we include tests for `environment: 'local'` with `CI=0` and `CI=1`?

dgp1130 (approved):

packages/angular/cli/src/utilities/environment-options.ts
alan-agius4: Yeah, let me do that

tests/legacy-cli/e2e/tests/commands/cache/cache-info.ts
alan-agius4: Added

clydin (approved):
I think the current list of subcommands is good to start.  We can always add additional based on user need.

PR #22833, opened 2022-03-11
feat(@angular/cli): support TypeScript 4.6 

alan-agius4:

PR #22853, opened 2022-03-16
refactor: bump CLI version to `13.3.0-rc.0`

dgp1130:
We won't actually be releasing an RC, but this is necessary for the release tool to understand that it will do a minor release from this branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@angular/devkit-repo",
-  "version": "13.3.0-next.0",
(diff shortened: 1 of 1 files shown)


devversion (approved):
LGTM, I was asking for that in the [initial PR](https://github.com/angular/angular-cli/pull/22833#issuecomment-1065331209)

dgp1130:
Skipping CI here to start on the real 13.3.0 release, I'll run the full CI on that PR and we can handle any problems there.

PR #22860, opened 2022-03-17
feat(@angular/cli): add support for multiple schematics collections

### LINKED ISSUE #12157: Specify multiple default collections
_From @jeffbcross on October 31, 2017 22:4_

```
- [x] feature request
```

### Area
```
- [] devkit
- [x] schematics
```

### Desired functionality
It'd be nice to have multiple collections configured in my CLI project instead of a single default collection. I.e. I'd like to have a project that gives preference to `nrwl/schematics` but will fall back to `angular/schematics` if I try to generate something not provided by `nrwl/schematics`. This would make it easier to publish many focused schematics without requiring the user to specify the collection each time they're generating something new. My naive suggestion is to change the `.angular-cli.json` schema from a single `collection` value to a cascading array of `collections` like so:

```json
"defaults": {
  "collection": "nrwl/schematics"
}
```

```json
"defaults": {
  "collections": [
    "nrwl/schematics",
    "angular/schematics"
  ]
}
```

Then when calling `ng generate`, schematics would use the first collection that provides a matching schematic.

This issue seems cross-cutting between CLI and Schematics, so I thought this repository was the most appropriate place to open it, but lmk if I should open it on angular/angular-cli.

(This may seem similar to #34, but it's more focused on end user experience than schematics author experience).

_Copied from original issue: angular/devkit#249_

alan-agius4:
The `schematicsCollections` can be placed under the `cli` option in the global `.angular.json` configuration, at the root or at project level in `angular.json` .

```json
{
  "version": 1,
  "cli": {
    "schematicsCollections": [
      "@schematics/angular",
      "@angular/material"
    ]
  }
  ...
}
```

••Rationale••

When this option is not configured and a user would like to run a schematic which is not part of `@schematics/angular`,
the collection name needs to be provided to `ng generate` command in the form of `[collection-name:schematic-name]`. This make the `ng generate` command too verbose for repeated usages.

This is were `schematicsCollections` comes handle. When adding `@angular/material` to the list of `schematicsCollections`, the generate command will try to locate the schematic in the specified collections.

```
ng generate navigation
```

is be equivalent to:

```
ng generate @angular/material:navigation
```

••Conflicting schematic names••

When multiple collections have a schematic with the same name. Both `ng generate` and `ng new` will run the first schematic with the matched based on the ordering (as specified) of `schematicsCollections`.

Closes #12157


docs/specifications/schematics-collections-config.md
```diff
@@ -0,0 +1,38 @@
+# Schematics Collections (`schematicsCollections`)
+
+The `schematicsCollections` can be placed under the `cli` option in the global `.angular.json` configuration, at the root or at project level in `angular.json` .
```
alan-agius4, line 3: `schematicsCollections` or `schematicCollections`?

packages/angular/cli/src/command-builder/schematics-command-module.ts
```diff
@@ -238,36 +243,49 @@ export abstract class SchematicsCommandModule
...
+    const globalWorkspace = await getWorkspace('global');
+    const value =
+      getschematicsCollections(localWorkspace?.getCli()) ??
+      getschematicsCollections(globalWorkspace?.getCli());
+    if (value) {
+      return (this._schematicsCollections = value);
     }
 
-    return (this._defaultSchematicCollection = DEFAULT_SCHEMATICS_COLLECTION);
+    return (this._schematicsCollections = new Set([DEFAULT_SCHEMATICS_COLLECTION]));
```
dgp1130, line 288: Nit: I see that this was pre-existing, but I think we should separate out these assignments into separate statements to avoid a side-effectful return.

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -48,35 +49,27 @@ export class GenerateCommandModule
...
-    }
-
-    for (const schematicName of schematicNames) {
       const {
-        description: { schemaJson, aliases: schematicAliases, hidden: schematicHidden },
+        description: {
+          schemaJson,
+          aliases: schematicAliases,
+          hidden: schematicHidden,
+          description: schematicDescription,
```
dgp1130, line 63: Oh cool, I didn't know you could nest destructure renames like this.

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -142,12 +138,14 @@ export class GenerateCommandModule
     );
 
     const dasherizedSchematicName = strings.dasherize(schematicName);
+    const schematicsCollectionsFromConfig = await this.getschematicsCollections();
+    const collectionNames = await this.getCollectionNames();
 
     // Only add the collection name as part of the command when it's not the default collection or when it has been provided via the CLI.
     // Ex:`ng generate @schematics/angular:component`
```
dgp1130, line 145: Nit: Do we need to update this comment?

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -162,4 +160,45 @@ export class GenerateCommandModule
...
+     */
+    const schematicsToRegister = new Map<string, string>();
+    const collectionNames = await this.getCollectionNames();
+
+    for (const collectionName of collectionNames) {
+      const workflow = this.getOrCreateWorkflowForBuilder(collectionName);
+      const collection = workflow.engine.createCollection(collectionName);
+      const schematicsInCollection = collection.description.schematics;
+
+      // We cannot use `collection.listSchematicNames()` as this doesn't return hidden schematics.
```
dgp1130, line 184: Can we add a new `collection.listHiddenSchematicNames()` or `collection.listAllSchematicNames()`?

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -162,4 +160,45 @@ export class GenerateCommandModule
...
+      for (const schematicName of Object.keys(schematicsInCollection)) {
+        if (schematicsInCollection[schematicName].private) {
+          continue;
+        }
+
+        if (schematicNameFromArgs === schematicName) {
+          // No need to process all schematics since we know which one the user invoked.
+          return [[schematicName, collectionName]];
+        }
```
dgp1130, line 193: Nit: This feels like it's breaking abstraction to me. The point of this function is to get all the schematics, right? Not to find a particular schematic. I think it would be up to the caller to look for a particular schematic. If you want to short-circuit the logic, this could be rewritten as a generator, and then the caller can just `break;` when it finds what it wants. The `.sort()` call might make that impractical though.

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -162,4 +160,45 @@ export class GenerateCommandModule
...
+
+        // If a schematic with this same name is already registered skip.
+        if (!schematicsToRegister.has(schematicName)) {
+          schematicsToRegister.set(schematicName, collectionName);
+        }
+      }
+    }
+
+    return [...schematicsToRegister.entries()].sort();
```
dgp1130, line 202: `.entries()` is an array of key value pairs (arrays) right? How does `.sort()` compare each of these pairs? A quick look at [the MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description) says that `.sort()` will stringify it's inputs and compare the strings, which seems like it could lead to bugs. An explicit compare function seems like a good idea here.

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -162,4 +160,45 @@ export class GenerateCommandModule
 
     return `${commandName}${positionalArgs ? ' ' + positionalArgs : ''}`;
   }
+
+  private async getSchematicsToRegister(): Promise<
+    [schematicName: string, collectionName: string][]
```
dgp1130, line 165: Nit: It might be a cleaner abstraction to have this return a `Map<schematicName: string, collectionName: string>` and then do the `.entires().sort()` in the caller. Up to you where we should enforce ordering.

packages/angular/cli/src/commands/generate/cli.ts
alan-agius4: Or maybe 

```ts
listSchematicNames(collection: FileSystemCollectionDesc, includeHidden: boolean): string[];
```

packages/angular/cli/src/commands/generate/cli.ts
alan-agius4: The point of this method is actually get the schematics that needs to be registered, hence the name `getSchematicsToRegister`. 

IIf `schematicNameFromArgs` matches one of the schematic we only need to register the single schematic and ignore all the others, while when `schematicNameFromArgs` doesn't match any known schematic we need to register all schematics. 

I don't think a generator is the best approach here, since what we want is to get a list of schematics that we should register. 

But I did refactored this a bit more.

packages/angular/cli/src/commands/generate/cli.ts
alan-agius4: Haha, I tend to prefer the spread :) But okay!

I do agree about the sort.

packages/angular/cli/src/commands/generate/cli.ts
alan-agius4: I would agree about the abstraction that it should return the full list of schematics and the caller should filter. But in this case then the method should renamed to something else like `getListOfSchematics` etc.. which IMHO in this case isn't really needed for 2 points, this is a private method, which in case in future we need to get a full list of schematics we can always update the code later on.

That said, if you feel we should go with your approach I don't have not totally opposed to the idea.

packages/angular/cli/src/commands/generate/cli.ts
alan-agius4: Same as https://github.com/angular/angular-cli/pull/22860#discussion_r829879240

packages/angular/cli/src/commands/generate/cli.ts
alan-agius4: https://github.com/angular/angular-cli/pull/22860/commits/ff692ca72acba38015201b2a32b01580ab7a246c

packages/angular/cli/src/commands/generate/cli.ts
dgp1130: I guess `getListOfSchematics()` and registering all of them until we find the one we want makes more sense to me than `getSchematicsToRegister()` and doing all unless we find the one we want early.

I'm happy to defer to you here, it just stood out to me. I don't think it's a big enough deal to be a blocker for this PR.

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -162,4 +159,54 @@ export class GenerateCommandModule
...
+          schematicsToRegister.set(schematicName, collectionName);
+        }
+      }
+    }
+
+    // Didn't find the schematic or no schematic name was provided Ex: `ng generate --help`.
+    return Array.from(schematicsToRegister).sort((a, b) => {
+      // ignore casing
+      const nameA = a[0].toUpperCase();
+      const nameB = b[0].toUpperCase();
```
dgp1130, line 199: Nit: Could this be written as a destructure? That way it's a little more self-descriptive about what values are being compared.

```typescript
Array.from(schematicsToRegister).sort(([key1], [key2]) => /* ... */);
```

dgp1130 (approved):

packages/angular/cli/src/commands/generate/cli.ts
alan-agius4: Done

packages/angular/cli/src/commands/generate/cli.ts
alan-agius4: I added both method. The reason I kept `getSchematicsToRegister` is to reduce the size of the builder method and make it readable.

docs/specifications/schematics-collections-config.md
alan-agius4: Updated.

dgp1130 (approved):

packages/angular/cli/src/commands/generate/cli.ts
dgp1130: I like your approach here of making `getSchematicsToRegister()` wrap `getSchematics()` and adding the early return there.

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -171,4 +159,64 @@ export class GenerateCommandModule
...
+  private async *getSchematics(): AsyncGenerator<{
+    schematicName: string;
+    collectionName: string;
+  }> {
+    const seenNames = new Set<string>();
+    for (const collectionName of await this.getCollectionNames()) {
+      const workflow = this.getOrCreateWorkflowForBuilder(collectionName);
+      const collection = workflow.engine.createCollection(collectionName);
+
+      for (const schematicName of collection.listSchematicNames(true /** includeHidden */)) {
```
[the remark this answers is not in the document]
alan-agius4: The list returned by `collection.listSchematicNames` is already unique. But we want to have unique list from multiple `collection.listSchematicNames`, hence the above will not work.

packages/angular/cli/src/commands/generate/cli.ts
```diff
@@ -171,4 +159,64 @@ export class GenerateCommandModule
...
+  private async getSchematicsToRegister(): Promise<
+    [schematicName: string, collectionName: string][]
+  > {
+    const schematicsToRegister: [schematicName: string, collectionName: string][] = [];
+    const [, schematicNameFromArgs] = this.parseSchematicInfo(
+      // positional = [generate, component] or [generate]
+      this.context.args.positional[1],
+    );
+
+    for await (const { schematicName, collectionName } of await this.getSchematics()) {
```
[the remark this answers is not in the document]
alan-agius4: Good catch.

packages/angular/cli/src/commands/generate/cli.ts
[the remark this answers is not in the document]
dgp1130: Oh gotcha, NVM then.

packages/schematics/angular/migrations/update-14/replace-default-collection-option.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { JsonValue, isJsonObject } from '@angular-devkit/core';
+import { Rule } from '@angular-devkit/schematics';
+import { updateWorkspace } from '../../utility/workspace';
+
+/** Migration to raplace 'defaultCollection' option in angular.json. */
```
clydin, line 13:
```suggestion
/** Migration to replace 'defaultCollection' option in angular.json. */
```

PR #22862, opened 2022-03-18
fix(@schematics/angular): Consolidated setup with a single `beforeEach()`

PowerKiKi:
Since f463c8403547 we use native `async`/`await` and thus we do not need
to split setup in two `beforeEach()`

This would avoid confusion for newcomers as seen in https://stackoverflow.com/questions/52170948/why-component-generated-by-ng-cli-has-the-spec-file-with-2-beforeeach-method


alan-agius4 (approved):

@clydin any concern here?

Issue #22872, opened 2022-03-22
Prototype Pollution in minimist | @angular-devkit/schematics-cli v13.3.0


Shinigami92:
https://github.com/nestjs/nest-cli/issues/1579

**Blocking CI/CD in company project using `@nestjs/cli`**

Prototype Pollution in minimist

```node
> pnpm audit
┌─────────────────────┬───────────────────────────────────────────────────┐
│ high                │ Prototype Pollution in minimist                   │
├─────────────────────┼───────────────────────────────────────────────────┤
│ Package             │ minimist                                          │
├─────────────────────┼───────────────────────────────────────────────────┤
│ Vulnerable versions │ <=1.2.5                                           │
├─────────────────────┼───────────────────────────────────────────────────┤
│ Patched versions    │ <0.0.0                                            │
├─────────────────────┼───────────────────────────────────────────────────┤
│ More info           │ https://github.com/advisories/GHSA-xvch-5gv4-984h │
└─────────────────────┴───────────────────────────────────────────────────┘
1 vulnerabilities found
Severity: 1 high
```

```node
> pnpm why minimist
Legend: production dependency, optional only, dev only

dependencies:
@nestjs/apollo 10.0.7
├─┬ @nestjs/core 8.4.2 peer
│ └─┬ @nestjs/platform-express 8.4.2 peer
│   └─┬ multer 1.4.4
│     └─┬ mkdirp 0.5.5
│       └── minimist 1.2.6
└─┬ @nestjs/graphql 10.0.7 peer
  └─┬ @nestjs/core 8.4.2 peer
    └─┬ @nestjs/platform-express 8.4.2 peer
      └─┬ multer 1.4.4
        └─┬ mkdirp 0.5.5
          └── minimist 1.2.6
@nestjs/core 8.4.2
└─┬ @nestjs/platform-express 8.4.2 peer
  └─┬ multer 1.4.4
    └─┬ mkdirp 0.5.5
      └── minimist 1.2.6
@nestjs/graphql 10.0.7
└─┬ @nestjs/core 8.4.2 peer
  └─┬ @nestjs/platform-express 8.4.2 peer
    └─┬ multer 1.4.4
      └─┬ mkdirp 0.5.5
        └── minimist 1.2.6
@nestjs/platform-express 8.4.2
└─┬ multer 1.4.4
  └─┬ mkdirp 0.5.5
    └── minimist 1.2.6

devDependencies:
@nestjs/cli 8.2.4
[...]

PR #22873, opened 2022-03-22
build: update `minimist` to `1.2.6`

alan-agius4:
Closes https://github.com/angular/angular-cli/issues/22872


imdurgadas:
@alan-agius4 When will the version be released to npm that includes the above fix ?

PR #22876, opened 2022-03-22
Improve CLI bootstrapping time

alan-agius4:
With this change we reduce the time that is needed to bootstrap the CLI by ~200ms

Before
```
time ng b --help  1.65s
```

After
```
time ng b --help  1.49s
```

### DIFF
--- a/packages/angular/cli/src/analytics/analytics.ts
+++ b/packages/angular/cli/src/analytics/analytics.ts
@@ -8,7 +8,6 @@
 
 import { analytics, json, tags } from '@angular-devkit/core';
 import debug from 'debug';
-import * as inquirer from 'inquirer';
 import { v4 as uuidV4 } from 'uuid';
 import { colors } from '../utilities/color';
 import { getWorkspace, getWorkspaceRaw } from '../utilities/config';
@@ -103,7 +102,8 @@ export async function promptAnalytics(global: boolean, force = false): Promise<b
   }
 
   if (force || isTTY()) {
-    const answers = await inquirer.prompt<{ analytics: boolean }>([
+    const { prompt } = await import('inquirer');
+    const answers = await prompt<{ analytics: boolean }>([
       {
         type: 'confirm',
         name: 'analytics',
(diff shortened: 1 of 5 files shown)


packages/angular/cli/src/command-builder/command-module.ts
```diff
@@ -39,7 +38,6 @@ export interface CommandContext {
   root: string;
   workspace?: AngularWorkspace;
   logger: logging.Logger;
-  packageManager: PackageManager;
```
clydin, line 42: I kind of liked that this was always available to the commands.  how much time was spent calculating the package manager?

alan-agius4: I do like this myself quite a bit, so I put some more thoughts into it this morning. 

Currently, to get the package manager it takes `~349ms` which isn't great. There are a couple of options that we can do to minimize this.

1) Rely only lock files instead of checking for binaries which would bring down the time of the method to `~4ms`
```ts
   const packageManager = await getConfiguredPackageManager();
    if (packageManager) {
      return packageManager;
    }

    const [hasYarnLock, hasNpmLock, hasPnpmLock] = await Promise.all([
      hasLockfile(root, PackageManager.Yarn),
      hasLockfile(root, PackageManager.Npm),
      hasLockfile(root, PackageManager.Pnpm),
    ]);

    if (hasYarnLock && !hasNpmLock) {
      return PackageManager.Yarn;
    }
   
    if (hasPnpmLock && !hasNpmLock) {
      return PackageManager.Pnpm;
    }
   
    // TODO: This should eventually inform the user of ambiguous package manager usage.
    //       Potentially with a prompt to choose and optionally set as the default.
    return PackageManager.Npm;
```

I think the above should cover the majority of cases. 


2) Call `getPackageManager` on demand and potentially memorize the result. 

```ts
const context: {
     // ...
     packageManager: () => getPackageManager(root);
}
```

clydin (approved):

PR #22882, opened 2022-03-23
refactor(@angular/cli): clean up package-metadata retrieval logic

alan-agius4:
With this change we clean up the package-metadata retrieval logic and types by using public `@types/` packages.

Also, we lazily require `pacote` since this has a large set to dependencies which slows down module resolution.


packages/angular/cli/src/utilities/package-metadata.ts
```diff
@@ -48,37 +35,26 @@ export interface PackageIdentifier {
...
-    save?: NgAddSaveDepedency;
+    save?: NgAddSaveDependency;
   };
   'ng-update'?: {
-    migrations: string;
-    packageGroup: Record<string, string>;
+    migrations?: string;
+    packageGroup?: string[] | Record<string, string>;
+    packageGroupName?: string;
+    requirements?: string[] | Record<string, string>;
```
alan-agius4, line 46: I didn't even though this was a thing!!!

clydin (approved):

PR #22893, opened 2022-03-25
refactor(@angular/cli): create package manager util class

alan-agius4:
Apart from better code quality, this helps reduce the time of CLI bootstrapping time, as retrieving the package manager name is a rather expensive operator due to the number of process spawns.

The package manager name isn't always needed until we run a command and therefore in some cases we can see an improvement of around `~600ms`. Ex: `ng b --help`. From ` 1.34s` to `0.76s`.

This will be important when we eventually introduce auto complete as users will get faster loopback.


packages/angular/cli/src/commands/add/cli.ts
```diff
@@ -107,9 +105,9 @@ export class AddCommandModule
 
   // eslint-disable-next-line max-lines-per-function
   async run(options: Options<AddCommandArgs> & OtherOptions): Promise<number | void> {
-    const { root, logger, packageManager } = this.context;
+    const { logger, packageManager } = this.context;
     const { verbose, registry, collection, skipConfirmation } = options;
-    await ensureCompatibleNpm(root);
+    await packageManager.ensureCompatibleNpm();
```
clydin, line 110: nit: maybe just `ensureCompatibility` now in case we ever expand the checks?

packages/angular/cli/src/utilities/package-manager.ts
```diff
@@ -6,115 +6,297 @@
...
-        `npm version ${version} detected.` +
-          ' When using npm 7 with the Angular CLI, npm version 7.5.6 or higher is recommended.',
-      );
+    const [hasYarnLock, hasNpmLock, hasPnpmLock, hasYarn, hasPnpm, hasNpm] = await Promise.all([
+      this.hasLockfile(PackageManager.Yarn),
+      this.hasLockfile(PackageManager.Npm),
+      this.hasLockfile(PackageManager.Pnpm),
+      this.getVersion(PackageManager.Yarn),
+      this.getVersion(PackageManager.Pnpm),
+      this.getVersion(PackageManager.Npm),
```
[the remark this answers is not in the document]
alan-agius4: Actually, we will not invoke the package manager again and again when we add the memoize decorator. (If it gets approved and merged).

packages/angular/cli/src/commands/add/cli.ts
alan-agius4: Agree

PR #22900, opened 2022-03-28
fix(@angular-devkit/build-angular): allow Workers in Stackblitz

alan-agius4:
This is no longer needed as Stackblitz implemented the missing Node.Js worker APIs.//cc @d3lm from StackBlitz.


d3lm (approved):
Looks good to me 👍

PR #22910, opened 2022-03-29
refactor(@angular-devkit/core): deprecate unused exception classes

alan-agius4:
With this change we deprecate exception classes that are not used in the CLI repo.

DEPRECATED:

- `ContentHasMutatedException`, `InvalidUpdateRecordException`, `UnimplementedException` and `MergeConflictException` symbol from `@angular-devkit/core` have been deprecated in favor of the  symbol from `@angular-devkit/schematics`.
- `UnsupportedPlatformException` - A custom error exception should be created instead.

### DIFF
--- a/packages/angular_devkit/core/src/index.ts
+++ b/packages/angular_devkit/core/src/index.ts
@@ -12,7 +12,7 @@ import * as json from './json/index';
 import * as logging from './logger/index';
 import * as workspaces from './workspace';
 
-export * from './exception/exception';
+export * from './exception';
 export * from './json/index';
 export * from './utils/index';
 export * from './virtual-fs/index';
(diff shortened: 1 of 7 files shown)


packages/angular_devkit/core/src/index.ts
```diff
@@ -12,7 +12,7 @@ import * as json from './json/index';
 import * as logging from './logger/index';
 import * as workspaces from './workspace';
 
-export * from './exception/exception';
+export * from './exception/index';
```
clydin, line 15: nit: Should we get rid of the index file?  It appears to only reexport everything from `exception`.

clydin (approved):

PR #22914, opened 2022-03-30
Couple of minor clean ups

alan-agius4:
**refactor(@angular/cli): remove unused `testing` option**
This option is never set and doesn't provide any value, since typically when one wants to debug, you add a `debugger` statement where needed or use the `break on exception` option in the debugger.

**refactor(@angular-devkit/build-angular): replace `try/catch` block with `catch`**
Since this is a promise the try/catch block will not catch promise rejections which was the reason for the try/catch block.

**refactor(@angular/cli): remove `getProjectsByPath` config util**
We already have a similar method `getProjectByCwd` that does the same thing.


packages/angular/cli/src/utilities/config.ts
```diff
@@ -21,21 +21,25 @@ function isJsonObject(value: json.JsonValue | undefined): value is json.JsonObje
...
+      return fs.writeFile(path, data);
     },
     async isDirectory(path) {
       try {
-        return statSync(path).isDirectory();
+        const stats = await fs.stat(path);
+
+        return stats.isDirectory();
       } catch {
         return false;
```
dgp1130, line 35: Unrelated: This `catch` behavior seems very confusing, if we don't have permissions to access a file, then both `isFile()` and `isDirectory()` would return `false`, which isn't intuitive.

Not really related to this change anyways, but it's another cleanup to consider.

packages/angular/cli/src/utilities/config.ts
```diff
@@ -21,21 +21,25 @@ function isJsonObject(value: json.JsonValue | undefined): value is json.JsonObje
 function createWorkspaceHost(): workspaces.WorkspaceHost {
   return {
     async readFile(path) {
-      return readFileSync(path, 'utf-8');
+      return fs.readFile(path, 'utf-8');
     },
     async writeFile(path, data) {
-      writeFileSync(path, data);
+      return fs.writeFile(path, data);
```
dgp1130, line 27: Nit: `WorkspaceHost.writeFile()` is declared to return `Promise<void>`. If `fs.writeFile()` changes in the future, this could introduce a type error or start returning a value at runtime which isn't defined by the type. I suggest dropping the `return`, or updating `WorkspaceHost` to be dependent on `fs`'s real types if that is the behavior we actually want.

dgp1130 (approved):

packages/angular/cli/src/utilities/config.ts
alan-agius4: Agreed but this is currently what the WorkspaceHost expected to be implemented.

PR #22927, opened 2022-03-31
refactor(@angular/cli): remove coercion from `packages`  positional configuration

alan-agius4:
Configuration the `packages` positional argument to be an array instead of using `coerce` method.


dgp1130 (approved):
Typo in commit message: "Configuration the `packages` positional argument..."

PR #22928, opened 2022-03-31
Minimatch v3.0.5 for branch 13.3.x

therek:
It seems that in the last dev dependencies update the version of minimatch has been updated in one place and omitted in the other.

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,7 +172,7 @@
     "loader-utils": "3.2.0",
     "magic-string": "0.25.7",
     "mini-css-extract-plugin": "2.5.3",
-    "minimatch": "3.0.4",
+    "minimatch": "3.0.5",
     "minimist": "1.2.6",
     "ng-packagr": "13.1.3",
     "node-fetch": "^2.2.0",
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -42,7 +42,7 @@
     "license-webpack-plugin": "4.0.2",
     "loader-utils": "3.2.0",
     "mini-css-extract-plugin": "2.5.3",
-    "minimatch": "3.0.4",
+    "minimatch": "3.0.5",
     "open": "8.4.0",
     "ora": "5.4.1",
     "parse5-html-rewriting-stream": "6.0.1",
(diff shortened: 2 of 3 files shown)


therek:
It's been modifed here: https://github.com/angular/angular-cli/commit/7ee3a80eafb592ccffe137e23cd2df372e0c6d12

alan-agius4:
That updates  the `@types/minimatch` package which is different

therek:
Though if you'd update `minimatch` package as well it wouldn't fail my security scans. That's why I need it to be at least v3.0.5.

alan-agius4:
3.0.4 has no reported vulnerabilities. 

See: https://snyk.io/vuln/npm:minimatch

therek:
Reported as part of VULNDB-282782:

> minimatch contains a flaw in the braceExpand() function in minimatch.js that is triggered as an improper regular expression is used to match patterns for brace expansion. This may allow a context-dependent attacker to hang or slow down a Node process using the library.

Fixed in 3.0.5 (https://github.com/isaacs/minimatch/commit/a8763f4388e51956be62dc6025cec1126beeb5e6). More info: https://huntr.dev/bounties/e4e1393c-d590-4492-9f43-8be3f3321629/

alan-agius4:
In this case the vulnerability cannot be exploited as `braceExpand` is not used by the Angular CLI. Also, we don't expect that the CLI to be used in production environments were users provide malicious crafted inputs. 

That being said, if you wish to update this package there are a couple of things you need to do.
1) Sign Google CLA
2) Update the dependency in https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/package.json and https://github.com/angular/angular-cli/blob/master/package.json
3) Amend the commit message to follow our guidelines, see: https://github.com/angular/angular-cli/blob/master/CONTRIBUTING.md. Please also squash all commits.

therek:
Done all three steps. I cannot set target labels (or any other labels) so I guess that's up to you.

alan-agius4 (approved):
LGTM

NIT: can you please update the commit message to the below?
```
fix(@angular-devkit/build-angular): update `minimatch` dependency to `3.0.5`

Update build-angular's minimatch dependency to v3.0.5 as a result of VULNDB-282782. See: https://huntr.dev/bounties/e4e1393c-d590-4492-9f43-8be3f3321629/
```

therek:
Updated

PR #22929, opened 2022-03-31
refactor(@angular-devkit/build-angular): replace most custom path normalize usage with Node.js builtins

clydin:
During the build initialization phase, many paths are converted back and forth between multiple normalized forms. These conversions involve potentially expensive string operations. The majority of the custom path `normalize` function from `@angular-devkit/core` usages have now been removed in favor of the Node.js builtin path functions. This change reduces the need to perform additional string manipulation where possible.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/normalize-file-replacements.ts
+++ b/packages/angular_devkit/build_angular/src/utils/normalize-file-replacements.ts
@@ -6,8 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { BaseException, Path, getSystemPath, join, normalize } from '@angular-devkit/core';
+import { BaseException } from '@angular-devkit/core';
 import { existsSync } from 'fs';
+import * as path from 'path';
 import { FileReplacement } from '../builders/browser/schema';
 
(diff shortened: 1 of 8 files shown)


packages/angular_devkit/build_angular/src/utils/normalize-file-replacements.ts
```diff
@@ -48,26 +49,26 @@ export function normalizeFileReplacements(
...
+  let withPath: string;
   if (fileReplacement.src && fileReplacement.replaceWith) {
-    replacePath = normalize(fileReplacement.src);
-    withPath = normalize(fileReplacement.replaceWith);
+    replacePath = path.normalize(fileReplacement.src);
+    withPath = path.normalize(fileReplacement.replaceWith);
   } else if (fileReplacement.replace && fileReplacement.with) {
-    replacePath = normalize(fileReplacement.replace);
-    withPath = normalize(fileReplacement.with);
+    replacePath = path.normalize(fileReplacement.replace);
```
alan-agius4, line 60: Do we need path normalization here? Especially considered that the segments are than joined using `path.join` which does normalization as well.

PR #22979, opened 2022-04-13
refactor(@angular-devkit/core): remove `fast-json-stable-stringify` usage in memoize helper

alan-agius4:
This is not needed as inputs are typed checked to be `JsonValue`. With this change we get a step closer to remove `fast-json-stable-stringify` dependency.

### DIFF
--- a/packages/angular_devkit/core/src/experimental/jobs/strategy.ts
+++ b/packages/angular_devkit/core/src/experimental/jobs/strategy.ts
@@ -6,10 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import stableStringify from 'fast-json-stable-stringify';
 import { Observable, Subject, concat, of } from 'rxjs';
 import { finalize, ignoreElements, share, shareReplay, tap } from 'rxjs/operators';
-import { JsonValue } from '../../json';
+import { JsonObject, JsonValue, isJsonObject } from '../../json';
 import {
   JobDescription,
   JobHandler,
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/core/src/experimental/jobs/strategy.ts
```diff
@@ -126,7 +125,7 @@ export namespace strategy {
 
     return (handler, options) => {
       const newHandler = (argument: A, context: JobHandlerContext<A, I, O>) => {
-        const argumentJson = stableStringify(argument);
+        const argumentJson = JSON.stringify(argument);
```
clydin, line 128: Technically this would cause a miss for `{ a: 1, b: 2 }` and `{ b: 2, a: 1 }`.  However, this function currently is not used inside the CLI and experimental. Maybe a basic key sort first would be sufficient and capture the overwhelming majority of cases.

packages/angular_devkit/core/src/experimental/jobs/strategy.ts
```diff
@@ -126,7 +125,21 @@ export namespace strategy {
 
     return (handler, options) => {
       const newHandler = (argument: A, context: JobHandlerContext<A, I, O>) => {
-        const argumentJson = stableStringify(argument);
+        let sortedArgs: JsonValue = argument;
+
+        if (isJsonArray(argument)) {
+          sortedArgs = argument.sort();
```
clydin, line 131: Maybe drop the array part since order could be significant in an array.

clydin (approved):

PR #23006, opened 2022-04-19
Introduce Schematics Tree methods for `readText` and `readJSON`

clydin:
The schematics `Tree` now contains an additional `readText` method that supports directly
reading the content of a file as UTF-8 text. This avoids the need to manually decode a Buffer
within a schematic when text content is needed. If a file path does not exist, an exception
will be thrown. While this differs from the semantics of `read`, it helps reduce the amount
of code needed for common schematic use cases.

The schematics `Tree` now contains an additional `readJSON` method that supports directly
reading and parsing the content of a file as UTF-8 JSON with comment support. This avoids the need to manually
decode a Buffer, parse and handle JSON comments within a schematic when JSON content is needed.
If a file path does not exist, an exception will be thrown. While this differs from the semantics of `read`,
it helps reduce the amount of code needed for common schematic use cases. JSON parse errors will also result
in an exception being thrown with a message detailing the error.

### DIFF
--- a/goldens/public-api/angular_devkit/schematics/index.md
+++ b/goldens/public-api/angular_devkit/schematics/index.md
@@ -235,6 +235,8 @@ export class DelegateTree implements Tree_2 {
     // (undocumented)
     read(path: string): Buffer | null;
     // (undocumented)
+    readText(path: string): string;
+    // (undocumented)
     rename(from: string, to: string): void;
     // (undocumented)
     get root(): DirEntry;
@@ -546,6 +548,8 @@ export class HostTree implements Tree_2 {
     overwrite(path: string, content: Buffer | string): void;
     // (undocumented)
     read(path: string): Buffer | null;
+    // (undocumented)
+    readText(path: string): string;
     // (undocumented)
     rename(from: string, to: string): void;
     // (undocumented)
--- a/goldens/public-api/angular_devkit/schematics/index.md
+++ b/goldens/public-api/angular_devkit/schematics/index.md
@@ -7,6 +7,7 @@
 /// <reference types="node" />
 
 import { BaseException } from '@angular-devkit/core';
+import { JsonValue } from '@angular-devkit/core';
 import { logging } from '@angular-devkit/core';
 import { Observable } from 'rxjs';
 import { Path } from '@angular-devkit/core';
@@ -234,6 +235,8 @@ export class DelegateTree implements Tree_2 {
     overwrite(path: string, content: Buffer | string): void;
     // (undocumented)
     read(path: string): Buffer | null;
+    // (undocumented)
+    readJson(path: string): JsonValue;
     // (undocumented)
     readText(path: string): string;
     // (undocumented)
@@ -548,6 +551,8 @@ export class HostTree implements Tree_2 {
     overwrite(path: string, content: Buffer | string): void;
     // (undocumented)
     read(path: string): Buffer | null;
+    // (undocumented)
+    readJson(path: string): JsonValue;
     // (undocumented)
     readText(path: string): string;
     // (undocumented)
--- a/packages/angular_devkit/schematics/src/tree/host-tree.ts
+++ b/packages/angular_devkit/schematics/src/tree/host-tree.ts
@@ -18,6 +18,7 @@ import {
 } from '@angular-devkit/core';
 import { EMPTY, Observable } from 'rxjs';
 import { concatMap, map, mergeMap } from 'rxjs/operators';
+import { TextDecoder } from 'util';
 import {
   ContentHasMutatedException,
   FileAlreadyExistException,
@@ -162,10 +163,10 @@ export class HostTree implements Tree {
       return tree._ancestry.has(this._id);
     }
     if (tree instanceof DelegateTree) {
-      return this.isAncestorOf(((tree as unknown) as { _other: Tree })._other);
+      return this.isAncestorOf((tree as unknown as { _other: Tree })._other);
     }
     if (tree instanceof ScopedTree) {
-      return this.isAncestorOf(((tree as unknown) as { _base: Tree })._base);
+      return this.isAncestorOf((tree as unknown as { _base: Tree })._base);
     }
 
     return false;
@@ -206,9 +207,9 @@ export class HostTree implements Tree {
               throw new MergeConflictException(path);
             }
 
-            this._record.overwrite(path, (content as {}) as virtualFs.FileBuffer).subscribe();
+            this._record.overwrite(path, content as {} as virtualFs.FileBuffer).subscribe();
           } else {
-            this._record.create(path, (content as {}) as virtualFs.FileBuffer).subscribe();
+            this._record.create(path, content as {} as virtualFs.FileBuffer).subscribe();
           }
 
           return;
@@ -234,7 +235,7 @@ export class HostTree implements Tree {
           }
           // We use write here as merge validation has already been done, and we want to let
           // the CordHost do its job.
-          this._record.write(path, (content as {}) as virtualFs.FileBuffer).subscribe();
+          this._record.write(path, content as {} as virtualFs.FileBuffer).subscribe();
 
           return;
         }
@@ -289,6 +290,26 @@ export class HostTree implements Tree {
 
     return entry ? entry.content : null;
   }
+
+  readText(path: string): string {
+    const data = this.read(path);
+    if (data === null) {
+      throw new FileDoesNotExistException(path);
+    }
+
+    const decoder = new TextDecoder('utf-8', { fatal: true });
+
+    try {
+      // With the `fatal` option enabled, invalid data will throw a TypeError
+      return decoder.decode(data);
+    } catch (e) {
+      if (e instanceof TypeError) {
+        throw new Error(`Failed to decode "${path}" as UTF-8 text.`);
+      }
+      throw e;
+    }
+  }
+
   exists(path: string): boolean {
     return this._recordSync.isFile(this._normalizePath(path));
   }
@@ -337,7 +358,7 @@ export class HostTree implements Tree {
       throw new FileDoesNotExistException(p);
     }
     const c = typeof content == 'string' ? Buffer.from(content) : content;
-    this._record.overwrite(p, (c as {}) as virtualFs.FileBuffer).subscribe();
+    this._record.overwrite(p, c as {} as virtualFs.FileBuffer).subscribe();
   }
   beginUpdate(path: string): UpdateRecorder {
     const entry = this.get(path);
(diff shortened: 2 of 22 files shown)


packages/angular_devkit/schematics/src/tree/interface.ts
```diff
@@ -91,6 +91,16 @@ export interface Tree {
...
   readText(path: string): string | null;
+
+  /**
+   * Reads and parses a file from the Tree as a UTF-8 encoded JSON file.
+   * Parsing supports JSON comments and trailing commas.
+   *
+   * @param path The path of the file to read.
+   * @returns A JsonValue containing the parsed contents of the file; if not found, returns `undefined`.
+   */
+  readJSON(path: string): JsonValue | undefined;
```
dgp1130, line 102: Nit: Do we have any standard casing convention for JSON? Should this be `readJson`?

I see that `JsonValue` is lower cased, but I'm not sure if we have any public APIs which use the term JSON and should make sure we're consistent with them.

dgp1130, line 102: If this supports comments and trailing commas, should we call it `readJSONC()` or something similar? It will successfully parse inputs that aren't JSON, which makes me think that `readJSON()` is the wrong name.

packages/angular_devkit/schematics/src/tree/host-tree_spec.ts
```diff
@@ -26,6 +26,23 @@ describe('HostTree', () => {
...
 
+  describe('readJSON', () => {
+    it('returns a JSON value when reading a file that exists', () => {
+      const tree = new HostTree();
+      tree.create('/textfile1', '{ "a": true, "b": "xyz" }');
+      tree.create('/textfile2', '123');
+      tree.create('/textfile3', 'null');
+      expect(tree.readJSON('/textfile1')).toEqual({ a: true, b: 'xyz' });
+      expect(tree.readJSON('/textfile2')).toEqual(123);
+      expect(tree.readJSON('/textfile3')).toBeNull();
```
dgp1130, line 37: Nit: Should test JSON files with comments and trailing commas, since that's an explicit part of its contract.

Should also test a malformed JSON file and confirm the failure case.

packages/angular_devkit/schematics/src/tree/interface.ts
```diff
@@ -91,6 +91,16 @@ export interface Tree {
    * @returns A string containing the contents of the file; if not found, returns `null`.
    */
   readText(path: string): string | null;
+
+  /**
+   * Reads and parses a file from the Tree as a UTF-8 encoded JSON file.
+   * Parsing supports JSON comments and trailing commas.
+   *
+   * @param path The path of the file to read.
+   * @returns A JsonValue containing the parsed contents of the file; if not found, returns `undefined`.
```
dgp1130, line 100: Nit: Should document what happens when given a file with malformed JSON.

packages/angular_devkit/schematics/src/tree/host-tree_spec.ts
```diff
@@ -11,6 +11,21 @@ import { FilterHostTree, HostTree } from './host-tree';
 import { MergeStrategy } from './interface';
 
 describe('HostTree', () => {
+  describe('readText', () => {
+    it('returns text when reading a file that exists', () => {
+      const tree = new HostTree();
+      tree.create('/textfile1', 'abc');
+      tree.create('/textfile2', '123');
+      expect(tree.readText('/textfile1')).toEqual('abc');
+      expect(tree.readText('/textfile2')).toEqual('123');
```
dgp1130, line 20: Nit: Should include a test for reading a binary file and verifying the failure case.

goldens/public-api/angular_devkit/schematics/index.md
```diff
@@ -234,6 +235,8 @@ export class DelegateTree implements Tree_2 {
     // (undocumented)
     read(path: string): Buffer | null;
     // (undocumented)
```
dgp1130, line 237: Question: `readJSON` is documented, so why is it listed as `(undocumented)` here?

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -299,6 +301,19 @@ export class HostTree implements Tree {
     return data.toString();
   }
 
+  readJSON(path: string): JsonValue | undefined {
+    const content = this.readText(path);
+    if (content === null) {
+      // undefined is returned for the not found case because null is a valid JSON value
+      return undefined;
+    }
```
dgp1130, line 309: This is really awkward, but I think your approach is probably the best one.

Did you consider changing `readText()` and `read()` to return `undefined` as well? I'm guessing that's more complicated because `read()` is already public API and that would be a breaking change. Just wondering if it would be worth making that change.

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -16,6 +17,7 @@ import {
   normalize,
   virtualFs,
 } from '@angular-devkit/core';
+import { parse } from 'jsonc-parser';
```
dgp1130, line 20: Nit: Consider aliasing this to `parseJsonc` or something similar, as `parse` is very generic and not clear what it would represent when called from the tree.

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -299,6 +301,19 @@ export class HostTree implements Tree {
...
   }
 
+  readJSON(path: string): JsonValue | undefined {
+    const content = this.readText(path);
+    if (content === null) {
+      // undefined is returned for the not found case because null is a valid JSON value
+      return undefined;
+    }
+
+    return parse(content, undefined /* errors */, {
```
dgp1130, line 311: Question: What is this `errors` argument? Is that some kind of list that `jsonc-parser` appends to with any errors? Should we validate that no errors were thrown before returning? The function declaration suggests that it will be fault tolerant and return something even when errors are thrown, which seems like implementation-specific behavior we probably don't want to expose publicly.

```typescript
/**
 * Parses the given text and returns the object the JSON content represents. On invalid input, the parser tries to be as fault tolerant as possible, but still return a result.
 * Therefore always check the errors list to find out if the input was valid.
 */
export declare function parse(text: string, errors?: {error: ParseErrorCode;}[], options?: ParseOptions): any;
```

dgp1130 (changes requested):
Thanks for splitting up the commits and taking the time to update existing usages. The `readJSON()` improvements seem particularly nice.

packages/angular_devkit/schematics/src/tree/interface.ts
clydin: I think that's a distinction that may lead to more confusion than benefit in this case.  Most `.json` files that would be typically interacted with in a schematic (tsconfig.json/angular.json/eslint.json/tslint.json/etc.) support comments (package.json withstanding) but are still generally considered to be JSON files.  From looking through TypeScript's JSON parsing functionality, it also does not appear to mention JSONC either but it does support comments.
But if there is a demand for strict adherence to the JSON spec, we could always add a `strict` option to this function.

goldens/public-api/angular_devkit/schematics/index.md
clydin: looks like api extractor doesn't look at the inheritance hierarchy.

packages/angular_devkit/schematics/src/tree/interface.ts
dgp1130: I agree that many of these files are really JSONC despite having a `.json` extension. I think if someone wants to parse strict JSON, they can just do `JSON.parse(tree.readText())`, so I don't see much motivation to support that. I see this function as supporting the JSONC use case, which is more complicated to implement and requires an extra dependency. Does it make sense to label it as such if that's all we want to support here?

packages/angular_devkit/schematics/src/tree/host-tree.ts
clydin: I agree.  I don't really like this either.
I have an alternative form of readText/readJson that throws instead which i think i'll bring into this PR.  From reworking existing schematics i think this better represents the usages of `read` even though the semantics will now differ from `read`.  Code similar to the following is repeated extensively throughout schematics (which could all be reduced to a single `readText` call):
```
  const text = host.read(modulePath);
  if (text === null) {
    throw new SchematicsException(`File ${modulePath} does not exist.`);
  }
  const sourceText = text.toString('utf-8');
```

packages/angular_devkit/schematics/src/tree/host-tree.ts
clydin: Something I was considering as well.
Added a check for the errors array that will throw with a message containing the error information.

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -289,6 +289,16 @@ export class HostTree implements Tree {
...
   }
+
+  readText(path: string): string | null {
+    const data = this.read(path);
+    if (data === null) {
+      return null;
+    }
+
+    return data.toString();
+  }
```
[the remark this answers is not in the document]
clydin: From reviewing existing schematics, converting to a string happens after almost every read call.  The raw data does not appear useful for most use cases.

packages/angular_devkit/schematics/src/tree/host-tree.ts
dgp1130: +1 to the error. I wish we could use some kind of `Result<string, SchematicsException>` here, but given the current state of JS, I think the error is the best way to go here.

packages/angular_devkit/schematics/src/tree/host-tree.ts
```diff
@@ -289,6 +289,16 @@ export class HostTree implements Tree {
...
     return entry ? entry.content : null;
   }
+
+  readText(path: string): string | null {
+    const data = this.read(path);
+    if (data === null) {
+      return null;
+    }
+
+    return data.toString();
```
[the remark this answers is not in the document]
clydin: It will now throw if the data cannot be decoded successfully as a UTF-8 string by using the `TextDecoder` class with the `fatal` option.

packages/angular_devkit/schematics/src/tree/interface.ts
clydin: If JSONC was formalized then i think it would make sense.  However, it's more of a shorthand for "JSON with comments" but without formalization as to the type/format of comments.  The JSON specification (RFC 8259 section 9) does allow for a compliant parser to support extensions to the specification though.  So i think `readJson` can be interpreted as parsing commonly used variants of JSON.  The documentation comment has been updated to make the parsing extensions more explicit.

dgp1130: TIL the JSON spec allows extensions. I guess that means JSON is a superset of itself?

Based on that reading, I agree that `readJson()` is an accurate name (though I don't know how many people would actually interpret it correctly). As long as the documentation explains what extensions this includes, then I think we're ok.

PR #23012, opened 2022-04-20
feat(@ngtools/webpack): add types generation

megion:
New feature: generate types when use webpack plugin `@ngtools/webpack`. 
We can not use webpack for build library with code based on angular because `@ngtools/webpack` does not generate types.
This changes save `d.ts` files to the `declarationDir` folder.


megion:
Hi @alan-agius4 ,

Thanks for your answer.

Unfortunately ng-packager can not compile library with styles.
`import 'style.css'`

Webpack can compile library with styles but '@ngtools/webpack' don't generate types.
So now there are no tools for create angular library with styles.

PR #23024, opened 2022-04-22
feat(@schematics/angular): ng generate component --standalone

alxhub:
This commit implements generation of a standalone component.


packages/schematics/angular/component/schema.json
```diff
@@ -48,6 +48,13 @@
       "alias": "t",
       "x-user-analytics": 10
     },
+    "standalone": {
+      "description": "Generate a standalone component",
```
alan-agius4, line 52: Drive by comment, can you please add a full stop at the end of the sentence?

PR #23032, opened 2022-04-24
feat(@schematics/angular): add --standalone to ng generate

crisbeto:
Adds the `--standalone` flag when generating components, directives or pipes through `ng generate`.

### DIFF
--- a/packages/schematics/angular/component/schema.json
+++ b/packages/schematics/angular/component/schema.json
@@ -48,6 +48,12 @@
       "alias": "t",
       "x-user-analytics": 10
     },
+    "standalone": {
+      "description": "Whether the generated component is standalone.",
+      "type": "boolean",
+      "default": false,
+      "x-user-analytics": 15
+    },
     "viewEncapsulation": {
       "description": "The view encapsulation strategy to use in the new component.",
       "enum": ["Emulated", "None", "ShadowDom"],
--- a/packages/schematics/angular/utility/find-module.ts
+++ b/packages/schematics/angular/utility/find-module.ts
@@ -17,6 +17,7 @@ export interface ModuleOptions {
   skipImport?: boolean;
   moduleExt?: string;
   routingModuleExt?: string;
+  standalone?: boolean;
 }
 
 export const MODULE_EXT = '.module.ts';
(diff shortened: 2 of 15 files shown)


alfaproject:
Isn't this a duplicate of https://github.com/angular/angular-cli/pull/23024

crisbeto:
I picked this up from Alex.

packages/schematics/angular/utility/find-module.ts
```diff
@@ -26,8 +27,7 @@ export const ROUTING_MODULE_EXT = '-routing.module.ts';
  * Find the module referred by a set of options passed to the schematics.
  */
 export function findModuleFromOptions(host: Tree, options: ModuleOptions): Path | undefined {
-  // eslint-disable-next-line no-prototype-builtins
-  if (options.hasOwnProperty('skipImport') && options.skipImport) {
+  if (options.standalone || options.skipImport) {
```
crisbeto, line 30: It might make sense to get rid of the `skipImport` option at some point.

packages/schematics/angular/component/schema.json
```diff
@@ -48,6 +48,13 @@
       "alias": "t",
       "x-user-analytics": 10
     },
+    "standalone": {
+      "description": "Whether the generated component is standalone.",
+      "type": "boolean",
+      "default": false,
+      "alias": "sa",
```
clydin, line 55: POSIX compliant aliases should only be a single letter.  This is not currently enforced but will be in the future.
Since lowercase `s` is already taken, we could use capital `S`, a different letter (`a`?), or drop the alias for now.

packages/schematics/angular/component/schema.json
```diff
@@ -48,6 +48,13 @@
...
+      "x-user-analytics": 10
```
clydin, line 56: Is standalone component generation something we want to track with analytics?  If so 15 appears to be the only free slot currently.
/cc @mgechev

clydin (changes requested):

packages/schematics/angular/utility/find-module.ts
crisbeto: I was considering that, but it could cause unintended consequences if the behavior of `skipImport` changes.

crisbeto:
I've addressed the feedback.

packages/schematics/angular/component/schema.json
crisbeto: I'll skip the alias for now.

clydin (approved):
Code LGTM.
Just need @mgechev & @dgp1130 approval for the analytics.

dgp1130 (approved):
LGTM. @clydin, was slot 15 used for something else previously? I'm just wondering if there's existing data for this flag? I think I remember that we have a limited number of dimensions we can use, is that this value?

clydin:
From the doc (https://github.com/angular/angular-cli/blob/master/docs/design/analytics.md#list-of-all-dimensions), it appears that there are only 20 allowed and number 15 was the only unused value.

mgechev (approved):

PR #23038, opened 2022-04-25
build: changes for primary branch rename to `main` [11.2.x]

devversion:
Changes part of the `DIRECT` phase of the "renaming master
to main" planning doc.

Note the version of ng-dev used in this LTS branch does not properly
support custom primary branches, but this should be non-relevant here
because this commit will only land in a version branch (11.2.x)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -19,7 +19,7 @@
     "every weekend"
   ],
   "baseBranches": [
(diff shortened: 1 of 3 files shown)


josephperrott (approved):
LGTM

PR #23041, opened 2022-04-26
feat(@schematics/angular): introduce a utility subpath export for Angular rules and utilities

clydin:
The `@schematics/angular` package now contains a defined set of package `exports` including a `utility` subpath export.
A wildcard export is also temporarily defined to support transition away from existing deep-import usage.
The `@schematics/angular/utility` subpath export will contain supported utility methods used by the first-party schematics
contained within the `@schematics/angular` package and can be considered public API that will follow SemVer stability constraints.
The first group of utilities introduced in this change are used to modify the `angular.json` workspace file
within the schematics and include the `updateWorkspace` rule and `readWorkspace`/`writeWorkspace` helpers.

### DIFF
--- a/packages/schematics/angular/utility/index.ts
+++ b/packages/schematics/angular/utility/index.ts
@@ -1,0 +1,21 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// Workspace related rules and types
+export {
+  ProjectDefinition,
+  TargetDefinition,
+  WorkspaceDefinition,
+  getWorkspace as readWorkspace,
+  updateWorkspace,
+  writeWorkspace,
+} from './workspace';
+export { Builders as AngularBuilder } from './workspace-models';
+
+// Package dependency related rules and types
+export { DependencyType, addDependency } from './dependency';
--- a/packages/schematics/angular/utility/workspace.ts
+++ b/packages/schematics/angular/utility/workspace.ts
@@ -10,58 +10,107 @@ import { json, workspaces } from '@angular-devkit/core';
 import { Rule, Tree, noop } from '@angular-devkit/schematics';
 import { ProjectType } from './workspace-models';
 
-function createHost(tree: Tree): workspaces.WorkspaceHost {
-  return {
-    async readFile(path: string): Promise<string> {
-      return tree.readText(path);
-    },
-    async writeFile(path: string, data: string): Promise<void> {
-      return tree.overwrite(path, data);
-    },
-    async isDirectory(path: string): Promise<boolean> {
-      // approximate a directory check
-      return !tree.exists(path) && tree.getDir(path).subfiles.length > 0;
-    },
-    async isFile(path: string): Promise<boolean> {
-      return tree.exists(path);
-    },
-  };
+const DEFAULT_WORKSPACE_PATH = '/angular.json';
+
+// re-export the workspace definition types for convenience
+export type WorkspaceDefinition = workspaces.WorkspaceDefinition;
+export type ProjectDefinition = workspaces.ProjectDefinition;
+export type TargetDefinition = workspaces.TargetDefinition;
+
+/**
+ * A {@link workspaces.WorkspaceHost} backed by a Schematics {@link Tree} instance.
+ */
+class TreeWorkspaceHost implements workspaces.WorkspaceHost {
+  constructor(private readonly tree: Tree) {}
+
+  async readFile(path: string): Promise<string> {
+    return this.tree.readText(path);
+  }
+
+  async writeFile(path: string, data: string): Promise<void> {
+    if (this.tree.exists(path)) {
+      this.tree.overwrite(path, data);
+    } else {
+      this.tree.create(path, data);
+    }
+  }
+
+  async isDirectory(path: string): Promise<boolean> {
+    // approximate a directory check
+    return !this.tree.exists(path) && this.tree.getDir(path).subfiles.length > 0;
+  }
+
+  async isFile(path: string): Promise<boolean> {
+    return this.tree.exists(path);
+  }
 }
 
+/**
+ * Updates the workspace file (`angular.json`) found within the root of the schematic's tree.
+ * The workspace object model can be directly modified within the provided updater function
+ * with changes being written to the workspace file after the updater function returns.
+ * The spacing and overall layout of the file (including comments) will be maintained where
+ * possible when updating the file.
+ *
+ * @param updater An update function that can be used to modify the object model for the
+ * workspace. A {@link WorkspaceDefinition} is provided as the first argument to the function.
+ */
 export function updateWorkspace(
-  updater: (workspace: workspaces.WorkspaceDefinition) => void | Rule | PromiseLike<void | Rule>,
-): Rule;
-export function updateWorkspace(workspace: workspaces.WorkspaceDefinition): Rule;
-export function updateWorkspace(
-  updaterOrWorkspace:
-    | workspaces.WorkspaceDefinition
-    | ((workspace: workspaces.WorkspaceDefinition) => void | Rule | PromiseLike<void | Rule>),
+  updater: (workspace: WorkspaceDefinition) => void | Rule | PromiseLike<void | Rule>,
 ): Rule {
   return async (tree: Tree) => {
-    const host = createHost(tree);
+    const host = new TreeWorkspaceHost(tree);
 
-    if (typeof updaterOrWorkspace === 'function') {
-      const { workspace } = await workspaces.readWorkspace('/', host);
+    const { workspace } = await workspaces.readWorkspace(DEFAULT_WORKSPACE_PATH, host);
 
-      const result = await updaterOrWorkspace(workspace);
+    const result = await updater(workspace);
 
-      await workspaces.writeWorkspace(workspace, host);
+    await workspaces.writeWorkspace(workspace, host);
 
-      return result || noop;
-    } else {
-      await workspaces.writeWorkspace(updaterOrWorkspace, host);
-
-      return noop;
-    }
+    return result || noop;
   };
 }
 
-export async function getWorkspace(tree: Tree, path = '/') {
-  const host = createHost(tree);
+// TODO: This should be renamed `readWorkspace` once deep imports are restricted (already exported from `utility` with that name)
+/**
+ * Reads a workspace file (`angular.json`) from the provided {@link Tree} instance.
+ *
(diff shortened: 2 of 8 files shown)


packages/schematics/angular/utility/workspace.ts
```diff
@@ -10,60 +10,109 @@ import { json, workspaces } from '@angular-devkit/core';
...
+  }
+
+  async isDirectory(path: string): Promise<boolean> {
+    // approximate a directory check
+    return !this.tree.exists(path) && this.tree.getDir(path).subfiles.length > 0;
+  }
+
+  async isFile(path: string): Promise<boolean> {
+    return this.tree.exists(path);
+  }
```
dgp1130, line 45: Similarly here: `isFile('some/non/existant/path')` returns `false`, when I think most users would expect a non-existent file to throw or something. Not sure if this is fixable right now though.

packages/schematics/angular/utility/workspace.ts
```diff
@@ -10,60 +10,109 @@ import { json, workspaces } from '@angular-devkit/core';
...
-      await workspaces.writeWorkspace(workspace, host);
+    await workspaces.writeWorkspace(workspace, host);
 
-      return result || noop;
-    } else {
-      await workspaces.writeWorkspace(updaterOrWorkspace, host);
-
-      return noop;
-    }
+    return result || noop;
```
dgp1130, line 70: Is `updater` supposed to modify the input workspace object or return a `Rule` which updates the entire tree with a new workspace? In the first it seems like `result` is useless while the second's `writeWorkspace()` doesn't do anything?

Maybe I just don't know enough about schematics to understand why this makes sense?

packages/schematics/angular/utility/workspace_spec.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const workspace = await readWorkspace(tree, '/xyz/');
+    expect(workspace.projects.has('test')).toBeTrue();
+  });
+
+  it('reads a workspace when specifying a file path', async () => {
+    const tree = new EmptyTree();
+    tree.create('/angular.json', TEST_WORKSPACE_CONTENT);
+
+    const workspace = await readWorkspace(tree, 'angular.json');
```
dgp1130, line 40: Nit: Should use a different file path here as `readWorkspace()` could ignore this field altogether and still pass the test because it happens to match the default.

packages/schematics/angular/utility/workspace_spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  it('throws if workspace file does not exist when using the default path value', async () => {
+    const tree = new EmptyTree();
+
+    await expectAsync(readWorkspace(tree)).toBeRejectedWithError();
+  });
+
+  it('throws if workspace file does not exist when specifying a file path', async () => {
+    const tree = new EmptyTree();
+
+    await expectAsync(readWorkspace(tree, 'angular.json')).toBeRejectedWithError();
```
dgp1130, line 53: Nit: Similar point here, should use a non-default workspace path and maybe even write the file at the default path in the tree just to confirm that `readWorkspace()` _doesn't_ pick up the default path and throws an error.

packages/schematics/angular/utility/workspace_spec.ts
```diff
@@ -0,0 +1,110 @@
...
+      workspace.projects.add({
+        name: 'abc',
+        root: 'src',
+      });
+    });
+
+    // Simulate execution of the rule in a schematic
+    await rule(tree, {} as unknown as SchematicContext);
+
+    expect(tree.read('angular.json')?.toString()).toContain('"abc"');
```
dgp1130, line 94: Nit: Could we simplify this assertion (and some of the previous ones) with `readWorkspace()`?

packages/schematics/angular/utility/index.ts
```diff
@@ -0,0 +1,18 @@
+/**
```
dgp1130, line 1: I'm wondering if there is a better name we can give this subpath. Most of the stuff here is related to workspaces, so should we consider exporting at `@schematics/angular/workspace`? What do you foresee adding in the future? Should we consider other subpaths for other common use cases?

packages/schematics/angular/utility/workspace.ts
```diff
@@ -10,60 +10,109 @@ import { json, workspaces } from '@angular-devkit/core';
...
+    if (this.tree.exists(path)) {
+      this.tree.overwrite(path, data);
+    } else {
+      this.tree.create(path, data);
+    }
+  }
+
+  async isDirectory(path: string): Promise<boolean> {
+    // approximate a directory check
+    return !this.tree.exists(path) && this.tree.getDir(path).subfiles.length > 0;
```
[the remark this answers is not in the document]
clydin: The Tree behavior was influenced by Git (there's branch and merge rules as well) and mainly focuses on files; not directories.

This class isn't exposed and is only used internally to adapt the Tree into a set of file access primitives needed by the workspace reader/writer.

packages/schematics/angular/utility/workspace.ts
```diff
@@ -10,60 +10,109 @@ import { json, workspaces } from '@angular-devkit/core';
...
+export type ProjectDefinition = workspaces.ProjectDefinition;
+export type TargetDefinition = workspaces.TargetDefinition;
+
+/**
+ * A {@link workspaces.WorkspaceHost} backed by a Schematics {@link Tree} instance.
+ */
+class TreeWorkspaceHost implements workspaces.WorkspaceHost {
+  constructor(private readonly tree: Tree) {}
+
+  async readFile(path: string): Promise<string> {
```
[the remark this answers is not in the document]
clydin: Correct, the tree is synchronous.  Within the `@angular/cli`, async calls are made via Node.js' `fs/promises` instead where direct filesystem access is performed.

packages/schematics/angular/utility/workspace.ts
```diff
@@ -10,60 +10,109 @@ import { json, workspaces } from '@angular-devkit/core';
...
+
+// re-export the workspace definition types for convenience
+export type WorkspaceDefinition = workspaces.WorkspaceDefinition;
+export type ProjectDefinition = workspaces.ProjectDefinition;
+export type TargetDefinition = workspaces.TargetDefinition;
+
+/**
+ * A {@link workspaces.WorkspaceHost} backed by a Schematics {@link Tree} instance.
+ */
+class TreeWorkspaceHost implements workspaces.WorkspaceHost {
```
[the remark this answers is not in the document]
clydin: Yes, it's a file access abstraction to allow the reader/writer to function in differing environments.  Here with a Schematics Tree but in the `@angular/cli` it's via Node.js `fs` functions.

packages/schematics/angular/utility/workspace.ts
clydin: The workspace object internally tracks changes to itself and when written fine-grained updates to the underlying file are made.  This allows for comments, formatting, etc. to be maintained where possible.

The updater function can optionally return a schematic rule (or chain of rules) that will then be executed by the runtime after the updateWorkspace rule completes.  For example, this can be useful when you would want to conditionally perform some additional actions (add a dependency?) based on existing and/or changed workspace content.
I'll add that to the doc comment.

packages/schematics/angular/utility/workspace.ts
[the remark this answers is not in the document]
dgp1130: If `WorkspaceHost` isn't public then I'm not too concerned for now, but this definitely feels like a footgun. Should we document this as part of the `WorkspaceHost` interface's contract? Maybe consider changing this later (not sure how hard that would be)?

packages/schematics/angular/utility/workspace.ts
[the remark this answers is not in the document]
dgp1130: Gotcha, thanks for the clarification.

packages/schematics/angular/utility/workspace.ts
dgp1130: Those sound like two different actions for `updater()`, does it need to do both? Couldn't users just put any other rules afterwards like:

```typescript
chain([
  updateWorkspace((wksp) => wksp.extensions.push({ /* ... */ })),
  addDependency('foo'),
])
```

packages/schematics/angular/utility/index.ts
clydin: I think too many subpaths can lead to a lack of discoverability.  It can also lead to a large amount of import statements in schematics for only a small amount of imported values per module.  Many of these rules are frequently used together.  The dependency related rules and the eventual AST helpers will also be available here as well.

packages/schematics/angular/utility/workspace.ts
clydin: `isFile`/`isDirectory` are not expected to throw.  If the path does not reference the expected item, it should return `false`.  The `fs` based implementation also behaves in this manner.  We can review and/or adjust the workspace related APIs moving forward though.

packages/schematics/angular/utility/workspace.ts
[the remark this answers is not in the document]
clydin: I think when we consider future Schematics runtime improvements, the Tree would be a good candidate for potential adjustments in areas like the above; especially regarding directories versus files.

packages/schematics/angular/utility/workspace.ts
clydin: If adding the dependency doesn't depend on the content of the workspace, then yes.
Here's an example from one of the old migrations that leverages the functionality:
https://github.com/angular/angular-cli/blob/6723a3c91efc52a48cdc22a1681928dbf7cd1944/packages/schematics/angular/migrations/update-9/ivy-libraries.ts#L23-L60

packages/schematics/angular/utility/workspace.ts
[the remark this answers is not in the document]
dgp1130: Sounds good, happy to defer for now.

packages/schematics/angular/utility/workspace.ts
dgp1130: +1

packages/schematics/angular/utility/workspace.ts
dgp1130: I see, I feel like we could use `readWorkspace()` before the `chain()` to some extent, but I could see some use cases that might be awkward with that. I don't want to hold this up, but something to consider in the future.

packages/schematics/angular/utility/index.ts
dgp1130: Sounds reasonable, I mainly just wanted to make sure we were considering alternatives as well.

dgp1130 (approved):

PR #23045, opened 2022-04-27
fix(@angular-devkit/build-angular): correctly resolve custom service worker configuration file

clydin:
Paths within the `angular.json` file should be relative to the location of the `angular.json` file.
The `ngswConfigPath` option was incorrectly using the current working directory for a base path when
a relative configuration path was specified. Most of the time this would work as a build command usually
is executed from the root of the workspace. However, this may not always be the case and for those cases
the actual workspace root is now used to resolve the full path for the service worker configuration file.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/app-shell/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/app-shell/index.ts
@@ -115,6 +115,7 @@ async function _renderUniversal(
     if (browserOptions.serviceWorker) {
       await augmentAppWithServiceWorker(
         projectRoot,
+        root,
         outputPath,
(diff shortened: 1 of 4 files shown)


clydin:
Supersedes #22749

dgp1130 (approved):
Should we list this as a breaking change? I get that it's really a bugfix, but if it's going into v14 anyways, we may as well label it as such.

I see that it's labeled as `target: minor`, where are you intending to merge this? Right now, I thought `target: major` would go into v14 while `target: patch` would do v14 and v13.3?

clydin:
Originally intended patch but the code differs due to a refactoring so a different PR for patch will be needed to avoid conflicts.

PR #23055, opened 2022-04-28
build: update node to m1 compatible v16

jbedard:
This sure would be nice :/

npm also updated to v7 due to: https://github.com/npm/npm/issues/3623 although that doesn't seem to fix it


jbedard:
I've switched to a different laptop so I don't need this atm, but it would be nice if using an m1 was possible. I'll close for now though since I don't think I'll do this anytime soon.

PR #23059, opened 2022-04-29
feat(@angular-devkit/schematics): disable package script execution by default in `NodePackageInstallTask`

clydin:
In an effort to improve supply chain security, the `NodePackageInstallTask` will now use the package
manager's `--ignore-scripts` option by default. Without the option, all direct and transitive
dependencies would have their scripts executed during the task's package manager installation operation.
The change only affects the package manager behavior controlled by the Schematics `NodePackageInstallTask`.

First-party Angular schematics do not currently require any direct or transitive dependency
`install`/`postinstall` scripts to execute. Only two dependencies within a v14.0 new project would
potentially be affected by this: `nice-napi` (transitive from `piscina`) and `esbuild`. The `nice-napi`
functionality of `piscina` is unused within the Angular CLI with no plans to use it in the future.
Even if it was used, the `install` script runs `node-gyp-build` which would only have an effect
(based on the current version 1.0.2) on platforms that are not Windows, darwin-x64, or linux-x64.
In the event this functionality is eventually used, the Angular CLI could be setup to automatically execute
this particular script for unsupported platforms. For `esbuild`, the `postinstall` functionality
performs an optional native binary bootstrap optimization but would only be performed if not
using Windows or Yarn. As such, it would not be performed for many users regardless of the change in
this commit. If noticeable performance regressions on platforms where the optimization was previously
performed are reported, the script could also be setup to be automatically run by the Angular CLI during
project creation and/or first build.

BREAKING CHANGE: Schematics `NodePackageInstallTask` will not execute package scripts by default
The `NodePackageInstallTask` will now use the package manager's `--ignore-scripts` option by default.
The `--ignore-scripts` option will prevent package scripts from executing automatically during an install.
If a schematic installs packages that need their `install`/`postinstall` scripts to be executed, the
`NodePackageInstallTask` now contains an `allowScripts` boolean option which can be enabled to provide the
previous behavior for that individual task. As with previous behavior, the `allowScripts` option will
prevent the individual task's usage of the `--ignore-scripts` option but will not override the package
manager's existing configuration.

### DIFF
--- a/packages/angular_devkit/schematics/tasks/package-manager/install-task.ts
+++ b/packages/angular_devkit/schematics/tasks/package-manager/install-task.ts
@@ -15,11 +15,13 @@ interface NodePackageInstallTaskOptions {
   workingDirectory?: string;
   quiet?: boolean;
   hideOutput?: boolean;
+  allowScripts?: boolean;
 }
 
 export class NodePackageInstallTask implements TaskConfigurationGenerator<NodePackageTaskOptions> {
   quiet = true;
   hideOutput = true;
+  allowScripts = false;
   workingDirectory?: string;
   packageManager?: string;
   packageName?: string;
@@ -45,6 +47,9 @@ export class NodePackageInstallTask implements TaskConfigurationGenerator<NodePa
       if (options.packageName != undefined) {
         this.packageName = options.packageName;
       }
+      if (options.allowScripts !== undefined) {
+        this.allowScripts = options.allowScripts;
+      }
     }
   }
 
@@ -58,6 +63,7 @@ export class NodePackageInstallTask implements TaskConfigurationGenerator<NodePa
         workingDirectory: this.workingDirectory,
         packageManager: this.packageManager,
         packageName: this.packageName,
+        allowScripts: this.allowScripts,
       },
     };
   }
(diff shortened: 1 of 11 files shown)


packages/angular_devkit/schematics/tasks/package-manager/install-task.ts
```diff
@@ -45,6 +47,9 @@ export class NodePackageInstallTask implements TaskConfigurationGenerator<NodePa
       if (options.packageName != undefined) {
         this.packageName = options.packageName;
       }
+      if (options.allowScripts != undefined) {
```
dgp1130, line 50: Nit: Should these be using `!==`? Probably not an issue when comparing with `undefined`, but good to use it consistently.

dgp1130 (approved):
Typo: `nice-napi` in the commit message.

Is there anything we should consider doing to make sure future changes to transitive dependencies don't introduce something which requires a postinstall?

dgp1130:
@twerske / @marktechson - I wonder if this would be worth mentioning the v14 blog post / release announcement as one way we're improving security for Angular applications?

clydin:
>Is there anything we should consider doing to make sure future changes to transitive dependencies don't introduce something which requires a postinstall?

Added a new E2E test that checks for install scripts within a new project's dependencies and fails if any unexpected packages are detected.

PR #23060, opened 2022-04-29
feat(@angular/cli): prompt to set up autocompletion and add `ng completion init`

### LINKED ISSUE #23003: Add a default prompt to run `ng completion`
# 🚀 Feature request

Prompt on any CLI without ng completion that hasn't been prompted before to run `ng completion` by default.

### Description

`ng completion` is the autocomplete feature I've always wanted, to improve CLI discoverability, prevent typos, and in general provide a better developer experience.

Adding this as a default prompt to CLI v14 users who have not run it yet helps make this improvement more built-in to the CLI instead of needing to advertise it and make it discoverable itself. This will greatly increase the number of developers who benefit from autocompletion.

Prompt CLI users to add completion and autocomplete to their `.bashrc`, `.bash_profile`, `.zshrc` or `.zsh_profile` by default (with user consent).

Angular team internal discussion here: [Thread](https://angular-team.slack.com/archives/C46U16D4Z/p1650310374177409)
Related PR here: https://github.com/angular/angular-cli/pull/22967

dgp1130:
This updates the CLI to set up autocompletion for users automatically. Running `ng completion` now finds the relevant `~/.bashrc`, `~/.zshrc`, etc. file and adds the command to load Angular autocompletion for you, so users don't have to understand how to do it. `ng completion script` prints the actual shell script which needs to be evaluated in the shell (previously this was done by `ng completion` directly).

This also includes a prompt which is displayed to users on first command execution. After set up or if the user refuses the prompt, the CLI stores this state in the global workspace configuration and avoids prompting again. Users are expected to use `ng completion` if they want to set up autocompletion after previously refusing the prompt. This hopefully provides a good middle ground where users who might not be aware of the feature can discover it without being too annoying to users who have custom shell setups or don't want autocompletion.

The prompt is also skipped when running on CI, in a non-interactive terminal, when running `ng completion` or `ng update` commands, or if a run configuration for the current shell already has `ng completion script` in it (implying the user already set this up).

End-to-end tests cover most of the flows, though there are a couple error conditions which are impractical to test. This required some modifications to test infrastructure to support passing stdin and environment variables in the right places. I also had to rewrite analytics tests to be more stable and not break from adding a new prompt.

Closes #23003.

### DIFF
--- a/packages/angular/cli/src/command-builder/command-module.ts
+++ b/packages/angular/cli/src/command-builder/command-module.ts
@@ -125,7 +125,10 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
     }
 
     // Set up autocompletion if appropriate.
-    const autocompletionExitCode = await considerSettingUpAutocompletion(this.context.logger);
+    const autocompletionExitCode = await considerSettingUpAutocompletion(
+      this.commandName,
+      this.context.logger,
+    );
     if (autocompletionExitCode !== undefined) {
       process.exitCode = autocompletionExitCode;
 
--- a/packages/angular/cli/src/command-builder/command-module.ts
+++ b/packages/angular/cli/src/command-builder/command-module.ts
@@ -19,6 +19,7 @@ import {
 } from 'yargs';
 import { Parser as yargsParser } from 'yargs/helpers';
 import { createAnalytics } from '../analytics/analytics';
+import { considerSettingUpAutocompletion } from '../utilities/completion';
 import { AngularWorkspace } from '../utilities/config';
 import { memoize } from '../utilities/memoize';
 import { PackageManagerUtils } from '../utilities/package-manager';
@@ -121,6 +122,14 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
     const camelCasedOptions: Record<string, unknown> = {};
     for (const [key, value] of Object.entries(options)) {
       camelCasedOptions[yargsParser.camelCase(key)] = value;
+    }
+
+    // Set up autocompletion if appropriate.
+    const autocompletionExitCode = await considerSettingUpAutocompletion(this.context.logger);
+    if (autocompletionExitCode !== undefined) {
+      process.exitCode = autocompletionExitCode;
+
+      return;
     }
 
     // Gather and report analytics.
--- a/packages/angular/cli/src/commands/completion/cli.ts
+++ b/packages/angular/cli/src/commands/completion/cli.ts
@@ -8,12 +8,50 @@
 
 import { join } from 'path';
 import yargs, { Argv } from 'yargs';
-import { CommandModule, CommandModuleImplementation } from '../../command-builder/command-module';
+import {
+  CommandModule,
+  CommandModuleImplementation,
+  CommandScope,
+} from '../../command-builder/command-module';
+import { addCommandModuleToYargs } from '../../command-builder/utilities/command';
+import { colors } from '../../utilities/color';
+import { initializeAutocomplete } from '../../utilities/completion';
 
 export class CompletionCommandModule extends CommandModule implements CommandModuleImplementation {
   command = 'completion';
-  describe = 'Generate a bash and zsh real-time type-ahead autocompletion script.';
+  describe = 'Set up Angular CLI autocompletion for your terminal.';
   longDescriptionPath = join(__dirname, 'long-description.md');
+
+  builder(localYargs: Argv): Argv {
+    return addCommandModuleToYargs(localYargs, CompletionScriptCommandModule, this.context);
+  }
+
+  async run(): Promise<number> {
+    let rcFile: string;
+    try {
+      rcFile = await initializeAutocomplete();
+    } catch (err) {
+      this.context.logger.error(err.message);
+
+      return 1;
+    }
+
+    this.context.logger.info(
+      `
+Appended \`source <(ng completion script)\` to \`${rcFile}\`. Restart your terminal or run the following to autocomplete \`ng\` commands:
+
+    ${colors.yellow('source <(ng completion script)')}
+      `.trim(),
+    );
+
+    return 0;
+  }
+}
+
+class CompletionScriptCommandModule extends CommandModule implements CommandModuleImplementation {
+  command = 'script';
+  describe = 'Generate a bash and zsh real-time type-ahead autocompletion script.';
+  longDescriptionPath = undefined;
 
   builder(localYargs: Argv): Argv {
     return localYargs;
--- a/packages/angular/cli/src/utilities/completion.ts
+++ b/packages/angular/cli/src/utilities/completion.ts
@@ -6,13 +6,23 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { logging } from '@angular-devkit/core';
+import { json, logging } from '@angular-devkit/core';
 import { promises as fs } from 'fs';
 import * as path from 'path';
 import { env } from 'process';
 import { colors } from '../utilities/color';
+import { getWorkspace } from '../utilities/config';
 import { forceAutocomplete } from '../utilities/environment-options';
 import { isTTY } from '../utilities/tty';
+
+/** Interface for the autocompletion configuration stored in the global workspace. */
+interface CompletionConfig {
+  /**
+   * Whether or not the user has been prompted to set up autocompletion. If `true`, should *not*
+   * prompt them again.
+   */
+  prompted?: boolean;
+}
 
 /**
  * Checks if it is appropriate to prompt the user to setup autocompletion. If not, does nothing. If
@@ -24,14 +34,27 @@ export async function considerSettingUpAutocompletion(
   logger: logging.Logger,
 ): Promise<number | undefined> {
   // Check if we should prompt the user to setup autocompletion.
-  if (!(await shouldPromptForAutocompletionSetup())) {
-    return undefined; // Already set up, nothing to do.
+  const completionConfig = await getCompletionConfig();
+  if (!(await shouldPromptForAutocompletionSetup(completionConfig))) {
+    return undefined; // Already set up or prompted previously, nothing to do.
   }
 
   // Prompt the user and record their response.
   const shouldSetupAutocompletion = await promptForAutocompletion();
   if (!shouldSetupAutocompletion) {
-    return undefined; // User rejected the prompt and doesn't want autocompletion.
+    // User rejected the prompt and doesn't want autocompletion.
+    logger.info(
+      `
+Ok, you won't be prompted again. Should you change your mind, the following command will set up autocompletion for you:
+
+    ${colors.yellow(`ng completion`)}
+    `.trim(),
+    );
+
+    // Save configuration to remember that the user was prompted and avoid prompting again.
+    await setCompletionConfig({ ...completionConfig, prompted: true });
+
+    return undefined;
   }
 
   // User accepted the prompt, set up autocompletion.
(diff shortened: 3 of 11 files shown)


packages/angular/cli/src/utilities/completion.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  wksp.extensions['cli'] ??= {};
+  const cli = wksp.extensions['cli'];
+  if (!json.isJsonObject(cli)) {
+    throw new Error(
+      `Invalid config found at ${wksp.filePath}. \`extensions.cli\` should be an object.`,
+    );
+  }
+  cli.completion ??= config as json.JsonObject;
```
alan-agius4, line 104: This doesn't  appear to be in the JSON schema. 

This needs to be added https://github.com/angular/angular-cli/blob/1c634cd327e5a850553b258aa2d5e6a6b2c75c65/packages/angular/cli/src/commands/config/cli.ts#L100 and also in the workspace schema and https://github.com/angular/angular-cli/blob/main/packages/angular/cli/lib/config/workspace-schema.json


Note: at the moment options can be both at global and local level, currently there is no schema for global configs only.

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -0,0 +1,237 @@
...
+  await wksp.save();
+}
+
+async function shouldPromptForAutocompletionSetup(config?: CompletionConfig): Promise<boolean> {
+  // Force whether or not to prompt for autocomplete to give an easy path for e2e testing to skip.
+  const forceAutocomplete = env['NG_FORCE_AUTOCOMPLETE'];
```
alan-agius4, line 110: Can you please move env variables to packages/angular/cli/src/utilities/environment-options.ts?

tests/legacy-cli/e2e/tests/misc/completion.ts
```diff
@@ -1,6 +1,10 @@
 import { execAndWaitForOutputToMatch } from '../../utils/process';
 
 export default async function () {
+  // TODO: `execAndWaitForOutputToMatch()` would resolve when the process completed, even if the
```
alan-agius4, line 4: good catch. I'll do a seperate PR.

alan-agius4:
Does using `source <(ng completion)` regress the auto completion times in anyway? My concern here is that the CLI would need to be be bootstrapped twice, which I think might slow the auto completion flow. 

Maybe paste the completion script instead?

packages/angular/cli/src/command-builder/command-module.ts
```diff
@@ -129,6 +130,14 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
       await this.reportAnalytics(camelCasedOptions);
     }
 
+    // Set up autocompletion if appropriate.
+    const autocompletionExitCode = await considerSettingUpAutocompletion(this.context.logger);
```
alan-agius4, line 134: Maybe placing this in the command-runner prior to running the command is better. Also, we might want to disable the auto completion prompt for certain commands, example, `ng update`.

packages/angular/cli/src/commands/completion/cli.ts
```diff
@@ -8,18 +8,57 @@
...
   run(): void {
     yargs.showCompletionScript();
   }
 }
+
+class CompletionInitCommandModule extends CommandModule implements CommandModuleImplementation {
+  command = 'init';
+  describe = 'Set up Angular CLI autocompletion for your terminal.';
+  longDescriptionPath = undefined;
+  static override scope = CommandScope.Both;
```
alan-agius4, line 38:
```suggestion
```

packages/angular/cli/src/commands/completion/cli.ts
```diff
@@ -8,18 +8,57 @@
...
+    return addCommandModuleToYargs(localYargs, CompletionInitCommandModule, this.context);
   }
 
   run(): void {
     yargs.showCompletionScript();
   }
 }
+
+class CompletionInitCommandModule extends CommandModule implements CommandModuleImplementation {
+  command = 'init';
```
alan-agius4, line 35: Maybe instead of having a sub command, the prompt should be in the `completion` command?

alan-agius4:
Using `<(ng completion)` will slow down the terminal spawning by around 300ms (The time need for `ng completion` to run).

Maybe we should consider adding directly the script instead?

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -0,0 +1,237 @@
...
+    // Failed to set up autocompeletion, log the error and abort.
+    logger.error(err.message);
+
+    return 1;
+  }
+
+  // Notify the user autocompletion was set up successfully.
+  logger.info(
+    `
+Appended \`source <(ng completion)\` to \`${rcFile}\`. Restart your terminal or run the following to autocomplete \`ng commands\`:
```
alan-agius4, line 73: `or run the following to autocomplete \`ng commands\` not really sure if this is really helpful. I personally find it confusing to ask users to run a command that doesn’t exists.

packages/angular/cli/src/utilities/completion.ts
alan-agius4, line 104: Since this is a method that sets the config, should this line be `cli.completion = config as json.JsonObject;`?


Otherwise might risk not being set if the property is an empty object literal for example.

dgp1130: Thanks for pointing that out, added `cli.completion.prompted` (I'm not sure if we expect `cli.completion` to do anything)?

packages/angular/cli/src/commands/completion/cli.ts
dgp1130: Maybe I'm not understanding, but I thought the purpose of this `scope` was whether to allow `ng completion init` to run inside a project, outside a project, or both? I think we would want `ng completion init` to work in any context right? It doesn't actually need to be related to a particular project?

packages/angular/cli/src/commands/completion/cli.ts
dgp1130: I see `ng completion` as responsible for generating the shell configuration for auto completion, while `ng completion init` is for setting up your `.bashrc` to use `ng completion`. Since these are two different actions, I think they should be different commands.

Considering that `ng completion` is something that users won't generally be calling themselves, I could see changing `ng completion` to initialize the setup by appending `source <(ng completion --get-shell-config)` or something if we want to view that part as more of an implementation detail. I could go either way whether `--get-shell-config` should be a flag on `ng completion` or a subcommand.

packages/angular/cli/src/utilities/completion.ts
dgp1130: Yes you're right, good catch.

dgp1130: I'm also wondering if we should have a more centralized workspace definition with some runtime component so we don't need to hard-code these paths? It would be very easy to forget them in the future.

packages/angular/cli/src/utilities/completion.ts
dgp1130: Oh whoops, I mean to say "`ng` commands" but misplaced the close tilde.

dgp1130:
> Using `<(ng completion)` will slow down the terminal spawning by around 300ms (The time need for `ng completion` to run).
> 
> Maybe we should consider adding directly the script instead?

I was wondering about this too. My thought process was that having users inline Yargs output would be leaking an implementation detail. It becomes harder to change in the future (whether by us or Yargs itself), is less clear about its purpose, and harder to detect if it is already setup (need to search for Yargs output which more tightly couples us to it).

Where did you get the 300ms number from? Is that some general understanding you have about the CLI startup time or something you've directly measured?

packages/angular/cli/src/command-builder/command-module.ts
dgp1130: I saw that for analytics, but didn't fully understand it. Thinking now, is that intended to avoid:

1. User runs `ng update` on 13.3.x.
2. `@angular/cli@13.3.x` updates itself to `14.0.x`.
3. `@angular/cli@14.0.x` runs `ng update`.
4. `ng update` prompts to set up autocompletion.

Instead we would rather prompt on the next `ng build` / `ng test` right? I updated it to ignore `ng update`, just want to make sure I understand the motivation.

I also skipped the prompt for `ng completion` commands, since that would be redundant. Calling `ng completion` means you already have a setup, and `ng completion init` means you _want_ a setup, so a prompt is unnecessary in both cases.

packages/angular/cli/src/commands/completion/cli.ts
alan-agius4: By default it's both, that why it is redundant.

packages/angular/cli/src/command-builder/command-module.ts
alan-agius4: Yeah, that is indeed the motivation behind that.

alan-agius4:
> > Using `<(ng completion)` will slow down the terminal spawning by around 300ms (The time need for `ng completion` to run).
> > Maybe we should consider adding directly the script instead?
> 
> I was wondering about this too. My thought process was that having users inline Yargs output would be leaking an implementation detail. It becomes harder to change in the future (whether by us or Yargs itself), is less clear about its purpose, and harder to detect if it is already setup (need to search for Yargs output which more tightly couples us to it).
> 
> Where did you get the 300ms number from? Is that some general understanding you have about the CLI startup time or something you've directly measured?

This is something I measured, and is the time taken to execute the `ng completion` command.

packages/angular/cli/src/commands/completion/cli.ts
alan-agius4: I like the `get-shell-config` concept. Maybe since users will not need to use this command/option it can also be hidden both from help and docs (aio).

packages/angular/cli/src/commands/completion/cli.ts
dgp1130: Oh I see. Removed.

packages/angular/cli/src/commands/completion/cli.ts
dgp1130: I think I like `ng completion` being the one that users run to set up autocompletion and the `ng completion some-flag-or-command` being the one that is added to shell configs, since users care less about that. I'm don't feel very strongly about `some-flag-or-command` given that it still needs to be public API anyways just for people who customize their `.bashrc` setups heavily. I think I still slightly prefer a subcommand rather than a flag since it has a completely different use case and different outputs which would be a lot to change with a single flag. As for the subcommand name `get-completion-script` is a little long and redundant, but it's also not something that users regularly type, so we don't need to overthink it too much.

I guess where I'm landing is:
* `ng completion` modifies your `.bashrc` to set up autocompletion for you.
* `ng completion get-completion-script` prints the Yargs setup.

dgp1130: @clydin suggested `ng completion script`, which I like since we don't need to repeat "completion" (and the "get" isn't that valuable either).

dgp1130:
> This is something I measured, and is the time taken to execute the `ng completion` command.

Hmm, 300ms startup isn't great, but I'm also wary of exposing that shell configuration too broadly. I'm thinking we start with `source <(ng completion)` and see how people react. We can consider other improvements to startup time where possible and reevaluate if there's significant push back, we can always change it later (though updating existing `.bashrc` files might be tricky).

I'm also thinking that power users are likely willing to inline `ng completion` output even if it's not strict public API, they could do that now if they want. It could change in a patch version, but that is 1) probably uncommon and 2) easy to fix when it happens. I imagine some folks will do that and it's probably fine, just not something we want to recommend unless they understand the consequences of that choice.

clydin (approved):

PR #23066, opened 2022-05-02
Completion tests

alan-agius4:


tests/legacy-cli/e2e/tests/misc/completion.ts
```diff
@@ -2,49 +2,61 @@ import { execAndWaitForOutputToMatch } from '../../utils/process';
 
 export default async function () {
```
dgp1130, line 3: Nit: Should these commits be squashed or swapped in ordering? As is, if the second commit got rolled back, CI would fail at just the first commit.

dgp1130 (approved):
Thanks for putting this together Alan, I was going to look into it later this week.

I don't want you to be too concerned with trying to land this PR, I'm happy to clone it and make a separate PR so you don't have to be worried about it. Just let me know and I'll take it off your hands.

dgp1130 (approved):

PR #23067, opened 2022-05-02
feat(@schematics/angular): introduce `addDependency` rule to utilities

clydin:
An `addDependency` schematics rule has been added to the newly introduced `utility` subpath export for the `@schematics/angular` package.
This new rule allows for adding a package as a dependency to a `package.json`. By default the `package.json` located
at the schematic's root will be used. The `packageJsonPath` option can be used to explicitly specify
a `package.json` in different location. The type of the dependency can also be specified instead
of the default of the `dependencies` section by using the `type` option for either development or peer dependencies.

Will be added to the export path in #23041

### DIFF
--- a/packages/schematics/angular/utility/dependency.ts
+++ b/packages/schematics/angular/utility/dependency.ts
@@ -1,0 +1,97 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Rule, SchematicContext } from '@angular-devkit/schematics';
+import { NodePackageInstallTask } from '@angular-devkit/schematics/tasks';
+import * as path from 'path';
+
+const installTasks = new WeakMap<SchematicContext, Set<string>>();
+
+interface MinimalPackageManifest {
+  dependencies?: Record<string, string>;
+  devDependencies?: Record<string, string>;
+  peerDependencies?: Record<string, string>;
+}
+
+/**
+ * An enum used to specify the type of a dependency found within a package manifest
+ * file (`package.json`).
+ */
+export enum DependencyType {
+  Default = 'dependencies',
+  Dev = 'devDependencies',
+  Peer = 'peerDependencies',
+}
+
+/**
+ * Adds a package as a dependency to a `package.json`. By default the `package.json` located
+ * at the schematic's root will be used. The `manifestPath` option can be used to explicitly specify
+ * a `package.json` in different location. The type of the dependency can also be specified instead
+ * of the default of the `dependencies` section by using the `type` option for either `devDependencies`
+ * or `peerDependencies`.
+ *
+ * When using this rule, {@link NodePackageInstallTask} does not need to be included directly by
+ * a schematic. A package manager install task will be automatically scheduled as needed.
+ *
+ * @param name The name of the package to add.
+ * @param specifier The package specifier for the package to add. Typically a SemVer range.
+ * @param options An optional object that can contain the `type` of the dependency
+ * and/or a path (`packageJsonPath`) of a manifest file (`package.json`) to modify.
+ * @returns A Schematics {@link Rule}
+ */
+export function addDependency(
+  name: string,
+  specifier: string,
+  options: {
+    /**
+     * The type of the dependency determines the section of the `package.json` to which the
+     * dependency will be added. Defaults to {@link DependencyType.Default} (`dependencies`).
+     */
+    type?: DependencyType;
+    /**
+     * The path of the package manifest file (`package.json`) that will be modified.
+     * Defaults to `/package.json`.
+     */
+    packageJsonPath?: string;
+  } = {},
+): Rule {
+  const { type = DependencyType.Default, packageJsonPath = '/package.json' } = options;
+
+  return (tree, context) => {
+    const manifest = tree.readJson(packageJsonPath) as MinimalPackageManifest;
+    const dependencySection = manifest[type];
+
+    if (!dependencySection) {
+      // Section is not present. The dependency can be added to a new object literal for the section.
+      manifest[type] = { [name]: specifier };
+    } else if (dependencySection[name] === specifier) {
+      // Already present with same specifier
+      return;
+    } else if (dependencySection[name]) {
+      // Already present but different specifier
+      throw new Error(`Package dependency "${name}" already exists with a different specifier.`);
+    } else {
+      // Add new dependency in alphabetical order
+      const entries = Object.entries(dependencySection);
+      entries.push([name, specifier]);
+      entries.sort((a, b) => a[0].localeCompare(b[0]));
+      manifest[type] = Object.fromEntries(entries);
+    }
+
+    tree.overwrite(packageJsonPath, JSON.stringify(manifest, null, 2));
+
+    const installPaths = installTasks.get(context) ?? new Set<string>();
+    if (!installPaths.has(packageJsonPath)) {
+      context.addTask(
+        new NodePackageInstallTask({ workingDirectory: path.dirname(packageJsonPath) }),
+      );
+      installPaths.add(packageJsonPath);
+      installTasks.set(context, installPaths);
+    }
+  };
+}
--- a/packages/schematics/angular/utility/dependency_spec.ts
+++ b/packages/schematics/angular/utility/dependency_spec.ts
@@ -1,0 +1,277 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {
+  EmptyTree,
+  Rule,
+  SchematicContext,
+  TaskConfigurationGenerator,
+  Tree,
+  callRule,
+  chain,
+} from '@angular-devkit/schematics';
+import { DependencyType, addDependency } from './dependency';
+
+async function testRule(rule: Rule, tree: Tree): Promise<TaskConfigurationGenerator[]> {
+  const tasks: TaskConfigurationGenerator[] = [];
+  const context = {
+    addTask(task: TaskConfigurationGenerator) {
+      tasks.push(task);
+    },
+  };
+
+  await callRule(rule, tree, context as unknown as SchematicContext).toPromise();
+
+  return tasks;
+}
+
+describe('addDependency', () => {
+  it('adds a package to "dependencies" by default', async () => {
+    const tree = new EmptyTree();
+    tree.create(
+      '/package.json',
+      JSON.stringify({
+        dependencies: {},
+      }),
+    );
+
+    const rule = addDependency('@angular/core', '^14.0.0');
+
+    await testRule(rule, tree);
+
+    expect(tree.readJson('/package.json')).toEqual({
+      dependencies: { '@angular/core': '^14.0.0' },
+    });
+  });
+
+  it('throws if a package is already present with a different specifier', async () => {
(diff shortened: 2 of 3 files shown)


packages/schematics/angular/utility/dependency.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @param name The name of the package to add.
+ * @param version The version specifier for the package to add.
+ * @param options An optional object that can contain the `type` of the dependency
+ * and/or a path (`manifestPath`) of a manifest file (`package.json`) to modify.
+ * @returns A Schematics {@link Rule}
+ */
+export function addDependency(
+  name: string,
+  version: string,
+  options: { type?: 'dev' | 'peer'; manifestPath?: string } = {},
```
dgp1130, line 40: Nit: I recommend *not* calling this `manifestPath` as `package.json` is not typically thought of as "manifest" and might be confused with other things like web app manifests or browser extension manifests.

Unfortunately there's no good generic name for `package.json`, so I suggest either `packageJson` or `packageJsonPath`. Not very elegant, but I think that would be more clear.

packages/schematics/angular/utility/dependency.ts
```diff
@@ -0,0 +1,78 @@
...
+  dependencies?: Record<string, string>;
+  devDependencies?: Record<string, string>;
+  peerDependencies?: Record<string, string>;
+}
+
+/**
+ * Adds a package as a dependency to a `package.json`. By default the `package.json` located
+ * at the schematic's root will be used. The `manifestPath` option can be used to explicitly specify
+ * a `package.json` in different location. The type of the dependency can also be specified instead
+ * of the default of the `dependencies` section by using the `type` option with either `dev` or `peer`
```
dgp1130, line 25: Nit: What about optional variants of dependencies? Do we want to support those?

packages/schematics/angular/utility/dependency.ts
dgp1130, line 40: Nit: Consider updating `type` to be `'dependencies' | 'devDependencies' | 'peerDependencies'` so users can explicitly specify `dependencies` if desired (makes scripting easier if they don't need to hard-code an exception for the default case) and simplify some of the logic below. An enum might also be a good way of typing this value.

dgp1130, line 40: Nit: I also suggest trying to come up with a more specific name than `type`, which is very vague and means many different things in many different circumstances.. Maybe `dependencyType`? That's [what Yarn calls them](https://classic.yarnpkg.com/en/docs/dependency-types) at least.

packages/schematics/angular/utility/dependency.ts
```diff
@@ -0,0 +1,78 @@
...
+  options: { type?: 'dev' | 'peer'; manifestPath?: string } = {},
+): Rule {
+  const { type, manifestPath = '/package.json' } = options;
+
+  return (tree, context) => {
+    const manifest = tree.readJson(manifestPath) as MinimalPackageManifest;
+
+    const sectionName: keyof MinimalPackageManifest = type ? `${type}Dependencies` : 'dependencies';
+    const dependencySection: Record<string, string> = (manifest[sectionName] ??=
+      Object.create(null));
```
dgp1130, line 49: Question: What's the need for `Object.create(null)` here? Couldn't we just use `{}`? If this is necessary, we should have a comment explaining why.

packages/schematics/angular/utility/dependency.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    if (dependencySection[name] === version) {
+      // Already present
+      return;
+    } else if (dependencySection[name]) {
+      // Update existing version
+      dependencySection[name] = version;
```
dgp1130, line 56: Question: Could this downgrade the version as well? Do we want to allow that?

packages/schematics/angular/utility/dependency.ts
```diff
@@ -0,0 +1,78 @@
...
+ * at the schematic's root will be used. The `manifestPath` option can be used to explicitly specify
+ * a `package.json` in different location. The type of the dependency can also be specified instead
+ * of the default of the `dependencies` section by using the `type` option with either `dev` or `peer`
+ * values.
+ *
+ * When using this rule, {@link NodePackageInstallTask} does not need to be included directly by
+ * a schematic. A package manager install task will be automatically scheduled as needed.
+ *
+ * @param name The name of the package to add.
+ * @param version The version specifier for the package to add.
```
dgp1130, line 32: Nit: Document that this needs to be a semver string.

packages/schematics/angular/utility/dependency_spec.ts
```diff
@@ -0,0 +1,263 @@
...
+        null,
+        2,
+      ),
+    );
+
+    const rule = addDependency('@angular/core', '^14.0.0');
+
+    await testRule(rule, tree);
+
+    expect(tree.readText('/package.json')).toBe(
```
dgp1130, line 87: Nit: Could we get better error messages from parsing the JSON and comparing the structured data?

```typescript
expect(tree.readJson('/package.json')).toEqual({
  dependencies: jasmine.objectInOrderOrSomething({
    '@angular/common': '^14.0.0',
    '@angular/core': '^14.0.0',
    '@angular/router': '^14.0.0',
  }),
});
```

I'm not sure if that asserts ordering, or if there's a `jasmine.*` utility which would do that. If not, we could do:

```typescript
expect(Object.entries(tree.readJson('/package.json').dependencies)).toEqual([
  [ '@angular/common', '^14.0.0' ],
  [ '@angular/core', '^14.0.0' ],
  [ '@angular/router', '^14.0.0' ],
]);
```

packages/schematics/angular/utility/dependency_spec.ts
```diff
@@ -0,0 +1,263 @@
...
+  it('schedules a package install task', async () => {
+    const tree = new EmptyTree();
+    tree.create('/package.json', JSON.stringify({}));
+
+    const rule = addDependency('@angular/core', '^14.0.0');
+
+    const tasks = await testRule(rule, tree);
+
+    // Should be an array with a single element with workingDirectory set
+    expect(tasks).toEqual([jasmine.objectContaining({ workingDirectory: '/' })]);
```
dgp1130, line 179: Is there any indication that this is the `NodePackageInstallTask`? Could we do an `instanceof` check to make sure the right task got scheduled? As is, it's hard to be sure that `workingDirectory` actually means we made the right task.

packages/schematics/angular/utility/dependency.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const installPaths = installTasks.get(context);
+    const needsInstall = !installPaths?.includes(manifestPath);
+    if (needsInstall) {
+      context.addTask(new NodePackageInstallTask({ workingDirectory: path.dirname(manifestPath) }));
+      if (!installPaths) {
+        installTasks.set(context, [manifestPath]);
+      } else {
+        installPaths.push(manifestPath);
+      }
```
dgp1130, line 75: Nit: Recommending making `installTasks` a `WeakMap<SchematicsContext, Set<string>>`, since order doesn't matter and we don't want duplicates that should be a more accurate type. It also means this section can be simplified to:

```typescript
const installPaths = installTasks.get(context) ?? new Set();
if (installPaths.has(manifestPath)) {
  context.addTask(new NodePackageInstallTask(/* ... */));
}
installPaths.add(manifestPath);
installTasks.set(context, installPaths);
```

packages/schematics/angular/utility/dependency.ts
clydin: Yarn calls them manifest files (https://yarnpkg.com/configuration/manifest) but we can go with `packageJsonPath`.

packages/schematics/angular/utility/dependency.ts
clydin: Considered them as well but they are not particularly common and could always be added in the future if there is demand.  There would also be optional peer dependencies but those need extra handling since it would need the addition of a `peerDependenciesMeta ` section with the package name.

packages/schematics/angular/utility/dependency.ts
clydin: Another option would be to throw if already present and maybe a flag to allow the behavior.

packages/schematics/angular/utility/dependency.ts
clydin: Can technically be any valid package specifier.  Updated the name and description to `specifier` instead.

packages/schematics/angular/utility/dependency.ts
dgp1130: +1 to deferring optional stuff until there's demand.

packages/schematics/angular/utility/dependency.ts
dgp1130: Interesting, I don't think I've ever heard of `package.json` files referred to as manifests, though I don't normally use Yarn.

clydin: The libraries (e.g., https://github.com/npm/pacote#readme) used by `npm` also use the term but from a quick look the `npm` documentation doesn't appear to mention it.

packages/schematics/angular/utility/dependency.ts
dgp1130: I'll defer to you on the right way to handle this, I don't want to make it more complicated than we need to. Throwing on downgrade or multiple `addDependency()` of the same package might be the most straightforward solution.

packages/schematics/angular/utility/dependency.ts
```diff
@@ -0,0 +1,97 @@
...
+    } else if (dependencySection[name] === specifier) {
+      // Already present with same specifier
+      return;
+    } else if (dependencySection[name]) {
+      // Already present but different specifier
+      throw new Error(`Package dependency "${name}" already exists with a different specifier.`);
+    } else {
+      // Add new dependency in alphabetical order
+      const entries = Object.entries(dependencySection);
+      entries.push([name, specifier]);
```
dgp1130, line 81: TIL that `Object.entries()` returns an actual `Array` and not an `Iterable` or other useless Array-like object.

packages/schematics/angular/utility/dependency_spec.ts
```diff
@@ -0,0 +1,259 @@
...
+    tree.create('/abc/package.json', JSON.stringify({}));
+
+    const rule = addDependency('@angular/core', '^14.0.0', {
+      packageJsonPath: '/abc/package.json',
+    });
+
+    const tasks = await testRule(rule, tree);
+
+    // Should be an array with a single element with working directory set
+    expect(tasks).toEqual([jasmine.objectContaining({ workingDirectory: '/abc' })]);
```
dgp1130, line 185: Nit: I'm a little worried looking for `workingDirectory` might be too brittle. We could check for `instanceof NodePackageInstallTask` for all of these, but internal changes to that task could still break these tests.

As an alternative, I wonder if we should consider:

```typescript
expect(tasks).toEqual([ new NodePackageInstallTask({ workingDirectory: '/abc' }) ]);
```

Or we could spy on the constructor and return some stubbed object, but my experience with Jasmine has taught me that is more difficult than it should be.

I'm not convinced that's totally better, since it assumes `new NodePackageInstallTask()` is pure / hermetic, so I'll leave that up to you as a suggestion.

dgp1130 (approved):
Great stuff @clydin, glad to get this in!

packages/schematics/angular/utility/dependency_spec.ts
clydin: Switch to testing the output of `toConfiguration` on the tasks array elements which is only an object literal and contains the relevant information the needs to be present for the actual task anyway.  This avoids the entire `instanceof` checking situation.

dgp1130 (approved):

PR #23074, opened 2022-05-03
run e2e legacy tests under bazel

jbedard:
This runs the legacy-cli/e2e tests via bazel using the bazel built npm packages from `packages/**`.

After all the pre-factors it's mainly just build config and a few code changes to use the bazel environment vars if present. This can be simplified if the non-bazel versions are dropped in the future. Currently all the CI jobs are prefixed with `bazel-`.

**Build targets / API:**

There are 9 bazel targets invoking the e2e tests with different configurations:
* `//tests/legacy-cli:e2e.{npm,yarn,esbuild,saucelabs}`: configured like CI today
* `//tests/legacy-cli:e2e.snapshot.{npm,yarn,esbuild,saucelabs}`: ^ using snapshots
* `//tests/legacy-cli:e2e`: nothing pre-configured, does not run by default or on CI, is designed to be run and configured manually with bazel cli args.

With any of these you can:
* `bazel test` to run as a bazel test
* `bazel run` to do the equivalent of `--debug` and remain running on failure
* `--test_filters="..."` which essentially does the e2e_runner `--glob="..."`
* `--test_arg="..."` to specify additional e2e_runner args, `--glob` would work here too

**Sharding**
Today e2e tests have `--shard` and `--nb-shards` cli params which circleci sets based on the `parallelism` config.
Now those cli args are defaulted based on the bazel environment variables set by `nodejs_test(shard_count)` and/or the `E2E_SHARD_TOTAL` + `E2E_SHARD` environment variables I added. This way both bazel and circleci sharding can be combined into one (bazel shards via processes on one circleci environment, circleci sharding in isolated environments).

**Prefactors**:
* https://github.com/angular/angular-cli/pull/23130
* https://github.com/angular/angular-cli/pull/23114
* https://github.com/angular/angular-cli/pull/23115
* https://github.com/angular/angular-cli/pull/23245
* https://github.com/angular/angular-cli/pull/23313
* https://github.com/angular/angular-cli/pull/23326
* https://github.com/angular/angular-cli/pull/23372
* https://github.com/angular/angular-cli/pull/23640
* https://github.com/angular/angular-cli/pull/23541
* https://github.com/angular/angular-cli/pull/23648
* https://github.com/angular/angular-cli/pull/23649
* https://github.com/angular/angular-cli/pull/23687
* https://github.com/angular/angular-cli/pull/23694
* https://github.com/angular/angular-cli/pull/23711
* https://github.com/angular/angular-cli/pull/23712
* https://github.com/angular/angular-cli/pull/23686
* https://github.com/angular/angular-cli/pull/23767
* https://github.com/angular/angular-cli/pull/23714
* https://github.com/angular/angular-cli/pull/23753
* https://github.com/angular/angular-cli/pull/23902
* https://github.com/angular/angular-cli/pull/23903
* https://github.com/angular/angular-cli/pull/23926

### DIFF
--- a/.circleci/dynamic_config.yml
+++ b/.circleci/dynamic_config.yml
@@ -48,7 +48,9 @@ var_6: &only_pull_requests
       only:
         - /pull\/\d+/
 
+# All e2e test suites
 var_7: &all_e2e_subsets ['npm', 'esbuild', 'yarn']
+var_8: &all_e2e_build_types ['e2e', 'snapshot']
 
 # Executor Definitions
 # https://circleci.com/docs/2.0/reusing-config/#authoring-reusable-executors
@@ -63,10 +65,20 @@ executors:
     working_directory: ~/ng
     resource_class: small
 
+  bazel-executor:
+    parameters:
+      nodeversion:
+        type: string
+        default: *default_nodeversion
+    docker:
+      - image: cimg/node:<< parameters.nodeversion >>-browsers
+    working_directory: ~/ng
+    resource_class: xlarge
+
   windows-executor:
     # Same as https://circleci.com/orbs/registry/orb/circleci/windows, but named.
     working_directory: ~/ng
-    resource_class: windows.medium
+    resource_class: windows.large
     shell: powershell.exe -ExecutionPolicy Bypass
     machine:
       # Contents of this image:
@@ -116,7 +128,7 @@ commands:
       - initialize_env
       - run: nvm install 16.13
       - run: nvm use 16.13
-      - run: npm install -g yarn@1.22.10
+      - run: npm install -g yarn@1.22.10 @bazel/bazelisk@${BAZELISK_VERSION}
       - run: node --version
       - run: yarn --version
 
@@ -126,6 +138,7 @@ commands:
         type: env_var_name
         default: CIRCLE_PROJECT_REPONAME
     steps:
+      - run: sudo cp .circleci/bazel.rc /etc/bazel.bazelrc
       - devinfra/setup-bazel-remote-exec:
           bazelrc: ./.bazelrc.user
 
@@ -269,23 +282,24 @@ jobs:
           paths:
             - dist/_*.tgz
 
-  build-bazel-e2e:
-    executor: action-executor
-    resource_class: medium
+  bazel-build:
+    executor: bazel-executor
     steps:
       - custom_attach_workspace
-      - run: yarn bazel build //tests/legacy-cli/...
+      - setup_bazel_rbe
+      - run:
+          name: Bazel Build Packages
+          command: yarn bazel build //...
+      - fail_fast
 
-  unit-test:
-    executor: action-executor
-    resource_class: xlarge
+  bazel-test:
+    executor: bazel-executor
     parameters:
       nodeversion:
         type: string
         default: *default_nodeversion_major
     steps:
       - custom_attach_workspace
-      - browser-tools/install-chrome
       - setup_bazel_rbe
       - run: sudo cp .circleci/bazel.rc /etc/bazel.bazelrc
       - when:
@@ -309,6 +323,59 @@ jobs:
                 # This timeout provides time for the actual tests to timeout and report status
                 # instead of CircleCI stopping the job without test failure information.
                 no_output_timeout: 40m
+      - fail_fast
+
+  bazel-e2e-tests:
+    executor: bazel-executor
+    parallelism: 8
+    parameters:
+      build_type:
+        type: enum
+        enum: *all_e2e_build_types
+        default: 'e2e'
+      subset:
+        type: enum
+        enum: *all_e2e_subsets
+        default: 'npm'
+    steps:
+      - custom_attach_workspace
+      - initialize_env
+      - setup_bazel_rbe
+      - run: mkdir /mnt/ramdisk/e2e
+      - run:
+          name: Test << parameters.build_type >> << parameters.subset >>
+          command: yarn bazel test --define=E2E_TEMP=/mnt/ramdisk/e2e --define=E2E_SHARD_TOTAL=${CIRCLE_NODE_TOTAL} --define=E2E_SHARD_INDEX=${CIRCLE_NODE_INDEX} --config=<< parameters.build_type >> //tests/legacy-cli:e2e.<< parameters.subset >>
+          no_output_timeout: 40m
+      - store_artifacts:
+          path: dist/testlogs/tests/legacy-cli/e2e.<< parameters.subset >>
+      - store_test_results:
+          path: dist/testlogs/tests/legacy-cli/e2e.<< parameters.subset >>
+      - fail_fast
+
+  bazel-test-browsers:
+    executor: bazel-executor
+    steps:
+      - custom_attach_workspace
+      - initialize_env
+      - setup_bazel_rbe
+      - run:
+          name: Initialize Saucelabs
+          command: setSecretVar SAUCE_ACCESS_KEY $(echo $SAUCE_ACCESS_KEY | rev)
+      - run:
+          name: Start Saucelabs Tunnel
+          command: ./scripts/saucelabs/start-tunnel.sh
+          background: true
+        # Waits for the Saucelabs tunnel to be ready. This ensures that we don't run tests
+        # too early without Saucelabs not being ready.
+      - run: ./scripts/saucelabs/wait-for-tunnel.sh
+      - run:
+          name: E2E Saucelabs Tests
+          command: yarn bazel test --config=saucelabs //tests/legacy-cli:e2e.saucelabs
+      - run: ./scripts/saucelabs/stop-tunnel.sh
+      - store_artifacts:
+          path: dist/testlogs/tests/legacy-cli/e2e.saucelabs
+      - store_test_results:
+          path: dist/testlogs/tests/legacy-cli/e2e.saucelabs
       - fail_fast
 
   snapshot_publish:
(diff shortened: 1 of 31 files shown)


alan-agius4:
Couple of questions 
- How is it possible to run a single test or using `--glob` option? 
- How to debug a single test IE the `--debug` option? 

Both of these are paramount for day to day tasks.

jbedard:
I've updated to add debugging via `bazel run` (instead of `bazel test`) and filtering with `bazel test --test_filters=...`. Both just set env vars that the e2e_runner reads as default values for the runner cli params.

jbedard:
Just working on it atm but...
* tests will be specified using the bazel `--test_filter` (will just get passed along as the `--glob` value)
* debug will be enabled with `bazel run` but not `bazel test`

.circleci/dynamic_config.yml
```diff
@@ -72,10 +74,20 @@ executors:
...
+    docker:
+      - image: cimg/node:<< parameters.nodeversion >>
+    working_directory: ~/ng
+    resource_class: 2xlarge
+
   windows-executor:
     # Same as https://circleci.com/orbs/registry/orb/circleci/windows, but named.
     working_directory: ~/ng
-    resource_class: windows.medium
+    resource_class: windows.2xlarge
```
alan-agius4, line 90: This will increase the credit usage drastically  500 credits/min compared previously 40 credits/min

https://circleci.com/product/features/resource-classes/#dev-windows

.circleci/dynamic_config.yml
```diff
@@ -72,10 +74,20 @@ executors:
...
 
+  bazel-executor:
+    parameters:
+      nodeversion:
+        type: string
+        default: *default_nodeversion
+    docker:
+      - image: cimg/node:<< parameters.nodeversion >>
+    working_directory: ~/ng
+    resource_class: 2xlarge
```
alan-agius4, line 85: Do we really need to use **2xlarge**?

.circleci/dynamic_config.yml
```diff
@@ -283,16 +296,23 @@ jobs:
...
       - run: yarn build
+      - persist_to_workspace:
+          root: *workspace_location
+          paths:
+            - dist/_*.tgz
 
-  build-bazel-e2e:
-    executor: action-executor
+  bazel-build:
+    executor: bazel-executor
```
alan-agius4, line 305: This would cause usage `2xlarge` resource_class class which is not great.I believe here large should be enough.

.circleci/dynamic_config.yml
```diff
@@ -283,16 +296,23 @@ jobs:
...
+      - run:
+          name: Bazel Build Packages
+          command: yarn bazel build //...
+      - fail_fast
 
-  unit-test:
-    executor: test-executor
-    resource_class: xlarge
+  bazel-test:
+    executor: bazel-executor
```
alan-agius4, line 315: Same as above.

.circleci/dynamic_config.yml
```diff
@@ -325,6 +345,60 @@ jobs:
...
+        type: enum
+        enum: *all_e2e_build_types
+        default: 'e2e'
+      subset:
+        type: enum
+        enum: *all_e2e_subsets
+        default: 'npm'
+    steps:
+      - custom_attach_workspace
+      - browser-tools/install-chrome
```
alan-agius4, line 362: Why is chrome needed?

.circleci/win-ram-disk.ps1
```diff
@@ -27,4 +27,4 @@ if (-not (Test-Path -Path $aimContents)) {
 
 # Setup RAM disk mount. Same parameters as ImDisk
 # See: https://support.circleci.com/hc/en-us/articles/4411520952091-Create-a-windows-RAM-disk
-./aim/cli/x64/aim_ll.exe -a -s 5G -m X: -p "/fs:ntfs /q /y"
+./aim/cli/x64/aim_ll.exe -a -s 12G -m X: -p "/fs:ntfs /q /y"
```
alan-agius4, line 30: Why such an increase in disk space usage for projects?

docs/DEVELOPER.md
```diff
@@ -84,6 +84,7 @@ You can find more info about debugging [tests with Bazel in the docs.](https://g
 
 - Run: `node tests/legacy-cli/run_e2e.js`
 - Run a subset of the tests: `node tests/legacy-cli/run_e2e.js tests/legacy-cli/e2e/tests/i18n/ivy-localize-*`
+- Run on a custom set of npm packages (tar files): `node tests/legacy-cli/run_e2e.js --package _angular_cli.tgz _angular_create.tgz dist/*.tgz ...`
```
alan-agius4, line 87: Would be helpful to write the equivalent of `node tests/legacy-cli/run_e2e.js tests/legacy-cli/e2e/tests/i18n/ivy-localize-*`.

.circleci/dynamic_config.yml
alan-agius4: //cc @josephperrott.

jbedard: I'll try decreasing it again. This was partially to understand why things were flaky or failing which I think was from running 8+ tests concurrently on a single machine. Now it's running 6 in bazel spread over 4 CI shards (the `parallelism` param).

.circleci/win-ram-disk.ps1
jbedard: bazel is running 6 tests concurrently.

I could try decreasing this to the minimum and/or reducing the concurrency if preffered?

alan-agius4: I am okay with this.

.circleci/dynamic_config.yml
jbedard: Without it I'm getting https://app.circleci.com/pipelines/github/angular/angular-cli/26617/workflows/8a85cca7-f1fb-4d5a-8698-83e522f485d4/jobs/352420/steps?invite=true#step-106-1703

... which might just required propper config of chrome within bazel. I can look into it more. It would be great to drop this step which requires 20-30s of CI time.

jbedard: This is also used in the unit tests (bazel) and non-bazel e2e tests, so maybe it should stay?

.circleci/dynamic_config.yml
```diff
@@ -48,7 +48,9 @@ var_6: &only_pull_requests
       only:
         - /pull\/\d+/
 
-var_7: &all_e2e_subsets ['npm', 'esbuild', 'yarn']
+# All e2e test suites
+var_7_subsets: &all_e2e_subsets ['npm', 'esbuild', 'yarn']
+var_7_build_types: &all_e2e_build_types ['e2e', 'snapshot']
```
josephperrott, line 53:
```suggestion
var_7: &all_e2e_subsets ['npm', 'esbuild', 'yarn']
var_8: &all_e2e_build_types ['e2e', 'snapshot']
```

Since the key in this pair isn't actually used by yaml, we typical just count rather than providing any "context" in the naming scheme.

.circleci/dynamic_config.yml
josephperrott: As I understand this it should not be necessary because we provide the browsers via bazel

BUILD.bazel
```diff
@@ -16,6 +17,14 @@ exports_files([
     "package.json",
 ])
 
+# Files required by e2e tests
+copy_to_bin(
+    name = "dot-files",
```
josephperrott, line 22: This is not a dot file right?  Were others included at one point?

jbedard: I assume it was at one point and has evolved over time. I'll change it 👍

josephperrott (approved):
LGTM though still other comments that appear to be unaddressed.

.circleci/dynamic_config.yml
jbedard: This was switched to just `xlarge` and I think we confirmed upgrading the resource_class was ok so I'm going to resolve this for now.

.circleci/dynamic_config.yml
jbedard: Same as above re: resource_class

jbedard:
I think those were all addressed offline and not in the comments here, just no one had pressed resolve.

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -182,42 +187,96 @@ export function useCIDefaults(projectName = 'test-project'): Promise<void> {
   });
 }
 
+const KARMA_CONF_DEFAULT = `
```
alan-agius4, line 190: Ideally we don't create this file as this would cause the the default CLI `ng test` code path not to be covered by the tests at all.

Maybe add the ChromeHeadlessNoSandbox as a custom browser in the builder?

alan-agius4 (approved):
One comment otherwise LGTM. Can you also squash please?

tests/legacy-cli/e2e/utils/project.ts
jbedard: Any examples for doing it in the builder like that?

I agree ideally this would not be here since it is not the standard user workflow that we want to be testing.

alan-agius4: It can be added here the custom browser in:
 https://github.com/angular/angular-cli/blob/df9674bd361b46b1275eed98e503f8897e701af5/packages/angular_devkit/build_angular/src/builders/karma/index.ts#L198

You can than use that browser by using the browsers option which can be set either in the CLI or angular.json 
```
ng test --browsers=ChromeHeadlessNoSandbox
```

jbedard: I got this working by adding the `ChromeHeadlessNoSandbox` to the karma builder. Is that what you meant? Adding it to the actual shipped code...?

alan-agius4 (approved):

tests/legacy-cli/e2e/utils/project.ts
alan-agius4: Yes.

PR #23075, opened 2022-05-03
fix(@angular/cli): improve error message for Windows autocompletion use cases

dgp1130:
Autocompletion tests [were failing on Windows](https://app.circleci.com/pipelines/github/angular/angular-cli/22152/workflows/8d883186-c144-4924-8a14-088e32d7c584/jobs/300321) due to a missing `$HOME` variable, presumably Windows wasn't inheriting the `$HOME` variable like Linux does?

Also fixing a semi-related typo I happened to notice around the same time.

Not sure exactly how to test this on Windows before merging, just hope for the best I guess?

### DIFF
--- a/packages/angular/cli/src/utilities/completion.ts
+++ b/packages/angular/cli/src/utilities/completion.ts
@@ -193,7 +193,9 @@ export async function initializeAutocomplete(): Promise<string> {
   if (!shell) {
     throw new Error(
       '`$SHELL` environment variable not set. Angular CLI autocompletion only supports Bash or' +
-        ' Zsh.',
+        " Zsh. If you're on Windows, Cmd and Powershell don't support command autocompletion," +
+        ' but Git Bash or Windows Subsystem for Linux should work, so please try again in one of' +
+        ' those environments.',
     );
   }
   const home = env['HOME'];
@@ -234,7 +236,7 @@ export async function initializeAutocomplete(): Promise<string> {
   return rcFile;
 }
 
-/** Returns an ordered list of possibile candidates of RC files used by the given shell. */
+/** Returns an ordered list of possible candidates of RC files used by the given shell. */
 function getShellRunCommandCandidates(shell: string, home: string): string[] | undefined {
   if (shell.toLowerCase().includes('bash')) {
     return ['.bashrc', '.bash_profile', '.profile'].map((file) => path.join(home, file));
--- a/tests/legacy-cli/e2e/tests/misc/completion-prompt.ts
+++ b/tests/legacy-cli/e2e/tests/misc/completion-prompt.ts
@@ -19,6 +19,13 @@ const DEFAULT_ENV = Object.freeze({
 });
 
 export default async function () {
+  // Windows Cmd and Powershell do not support autocompletion. Run a different set of tests to
+  // confirm autocompletion skips the prompt appropriately.
+  if (process.platform === 'win32') {
+    await windowsTests();
+    return;
+  }
+
   // Sets up autocompletion after user accepts a prompt from any command.
   await mockHome(async (home) => {
     const bashrc = path.join(home, '.bashrc');
@@ -42,7 +49,7 @@ export default async function () {
     const bashrcContents = await fs.readFile(bashrc, 'utf-8');
     if (!bashrcContents.includes('source <(ng completion script)')) {
       throw new Error(
-        'Autocompletion was *not* added to `~/.bashrc` after accepting the setup' + ' prompt.',
+        'Autocompletion was *not* added to `~/.bashrc` after accepting the setup prompt.',
       );
     }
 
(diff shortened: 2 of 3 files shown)


packages/angular/cli/src/utilities/completion.ts
```diff
@@ -188,6 +193,11 @@ will modify configuration files in your home directory.)
  * @return The full path of the configuration file modified.
  */
 export async function initializeAutocomplete(): Promise<string> {
+  // Windows doesn't support autocompletion, give a clear error message.
+  if (process.platform === 'win32') {
+    throw new Error(`Windows does not support shell autocompletion.`);
```
alan-agius4, line 198: While I haven't tried it, I don't think this is correct. Auto complete should work on Windows as long as the user has bash installed, example through git bash.

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -116,6 +116,11 @@ async function shouldPromptForAutocompletionSetup(
     return forceAutocomplete;
   }
 
+  // Don't prompt on Windows because it doesn't support autocompletion.
+  if (process.platform === 'win32') {
```
alan-agius4, line 120: Same as before.

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -116,6 +116,11 @@ async function shouldPromptForAutocompletionSetup(
...
+    return false;
+  }
+
```
alan-agius4, line 123: Nit: probably we should use the same logic as yargs to determine if the shell is bash or zsh.

https://github.com/yargs/yargs/blob/db354232705623bbcd8fad362f6a4d6d59650be5/lib/completion.ts#L44-L47

packages/angular/cli/src/utilities/completion.ts
clydin: I'm not sure about git bash but WSL 1/2 should have a `platform` of `linux` so that will be covered with that check.

alan-agius4: I am referring to usages outside of WSL. 

On native Windows, zsh and bash can still be installed and available and autocompletion is still possible using;

Example using;
- cygwin
- cmder
- git bash

Considering a large set of our users are on Windows OS, I think we should support this case better. Alternatively, for the time being we can say that the automated setup doesn't support Windows but explain how this can be done by modifying the zsh and bash profiles manually.

clydin: We could also check the `OSTYPE` environment variable when on Windows.  I believe it should be `msys` or `cygwin` for those cases but would need to confirm.

dgp1130: Yeah, I was going to check Git Bash today, as I suspect that will be `win32`. I forgot about Cygwin, but I could test that too. Cmder I thought was just a terminal which still used `cmd` under the hood and would be `win32` (it's not a Linux emulator IIRC)? WSL should definitely be detected as Linux.

I'm wondering if a better approach might be to check `$SHELL` for `cmd.exe` or `powershell.exe` and print an error message on that case? Not sure if there are other Windows shells which might come up there.

dgp1130: Just checked my personal machine and `process.platform` + `$SHELL` for each environment is:

* Cmd has platform `win32` and `$SHELL` is unset - should fail with a useful error message.
* Powershell has platform `win32` and `$SHELL` is unset - should fail with a useful error message.
* WSL2 has platform `linux` and shell `/usr/bin/zsh` - "just works".
* Git Bash has platform `win32` and shell `C:\Program files\Git\usr\bin\bash.exe` - NVM autocompletion works, so I suspect we should too, but I wasn't able to get it to use a local CLI build to confirm.
* Cmder has platform `win32` and `$SHELL` is unset, but I do see `$CMDER_SHELL="cmd"` - I suspect this shouldn't work and we should fail with a useful error message.
* Cygwin couldn't install Node in a reasonable time frame, so I gave up there -  I suspect this is similar to Git Bash and probably falls into that case.

All the unsupported systems have an unset `$SHELL`, so I think that is the best way of detecting the failure. Unfortunately there is no good way to disambiguate between the environments (Powershell doesn't seem to have any good environment variable to identify itself for instance). I'm inclined to fail with a more generic message on missing `$SHELL` along the lines of:

> `$SHELL` environment variable not set. Only Bash and Zsh are supported. If you're on Windows, Cmd and Powershell don't support command autocompletion, but Git Bash or Windows Subsystem for Linux should work, so make sure to run the command from there.
> 
> https://angular.io/cli/some-doc-about-autocompletion

(Sidebar: We should have a more directed/"goal-oriented" doc for setting up autocompletion which can explain the supported shells and link to relevant pieces.)

tests/legacy-cli/e2e/tests/misc/completion-prompt.ts
```diff
@@ -19,6 +19,11 @@ const DEFAULT_ENV = Object.freeze({
 });
 
 export default async function () {
+  if (process.platform === 'win32') {
+    await windowsTests();
+    return; // Windows should never prompt for autocompletion.
```
clydin, line 24: Would this fail if the test was run on Windows with git-bash?
CircleCI is setup to use powershell so CI would be fine. maybe expand the comment to note that.

clydin (approved):

tests/legacy-cli/e2e/tests/misc/completion-prompt.ts
dgp1130: Yes, I think this would fail for Git Bash, I wasn't too concerned since CI isn't doing that.

It kinda sucks since any contributors using Git Bash will run the wrong set of tests and possibly get false negatives. I'm not too worried about it right now though, and I think we can figure that out when someone actually tries to do it, resolves all the other issues they'd run into with Git Bash in our setup (no idea how it plays with Bazel for instance), and can more easily find the right detection mechanism.

clydin (approved):

PR #23077, opened 2022-05-04
build: cleanup references to old `master` branch

devversion:
Cleans up all referneces to the old `master` branch.


clydin (approved):

dgp1130:
Merged to just `main` for now. This has conflicts with `13.3.x`, so I think we'll need a manual cherry-pick if we want to make the change there.

PR #23079, opened 2022-05-04
build: run prettier on legacy-cli e2e files

jbedard:
I'm working on switching these to run via bazel and getting significant diffs when editing the files. Are we ok with a single commit formatting all these? For now it's just the legacy-cli which I hope is all the files I'll be editing when switching to bazel...

### DIFF
--- a/tests/legacy-cli/e2e/setup/100-global-cli.ts
+++ b/tests/legacy-cli/e2e/setup/100-global-cli.ts
@@ -1,7 +1,7 @@
 import { getGlobalVariable } from '../utils/env';
 import { exec, silentNpm } from '../utils/process';
 
-export default async function() {
+export default async function () {
   const argv = getGlobalVariable('argv');
   if (argv.noglobal) {
     return;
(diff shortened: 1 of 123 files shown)


clydin (approved):

dgp1130 (approved):
LGTM. I'm fine with a one-time format like this, but @clydin should we start enforcing the formatter for legacy tests? It seems like we haven't been doing that so far.

I didn't realize we have so many tests which still use Promises, we should consider updating those to `async/await` next time we touch any of them.

jbedard:
I think the pre-commit runs the formatter? I was assuming it was added at some point but never run on these files, because now when I commit any of them they get formatted...

Also note I've only formatted the e2e legacy-cli files which I hope is all #23085 touches, there's plenty more outside that directory.

clydin:
As mentioned, the commit hooks will run the formatter on changed files.  CI will also verify the formatting for all changed files in the PR.  When `prettier` was introduced essentially all files except those in the `tests` directory were formatted.

dgp1130:
Ok, so we just didn't do the one-time format when we first enabled in the formatter. As long as we're consistent moving forward, I have no concerns.

jbedard:
Thanks!

PR #23085, opened 2022-05-04
 build: replace minimist with yargs-parser 

jbedard:
`minimist` is not in package.json. This causes errors when building e2e tests with bazel which enforces this. Removing it was simple to reduce dependencies instead of adding more...

### DIFF
--- a/bin/devkit-admin
+++ b/bin/devkit-admin
@@ -17,10 +17,10 @@
 require('../lib/bootstrap-local');
 
 
-const minimist = require('minimist');
+const yargsParser = require('yargs-parser');
 const path = require('path');
 
-const args = minimist(process.argv.slice(2), {
+const args = yargsParser(process.argv.slice(2), {
   boolean: ['verbose']
 });
 const scriptName = args._.shift();


package.json
```diff
@@ -169,6 +171,7 @@
     "magic-string": "0.26.1",
     "mini-css-extract-plugin": "2.6.0",
     "minimatch": "5.0.1",
+    "minimist": "^1.2.6",
```
alan-agius4, line 174: Can we change the usages of `minimist` with `yargs-parser`? It should be pretty trivial.

package.json
```diff
@@ -189,6 +192,7 @@
     "puppeteer": "13.7.0",
     "quicktype-core": "6.0.69",
     "regenerator-runtime": "0.13.9",
+    "request": "^2.88.2",
```
alan-agius4, line 195: Should not be needed once https://github.com/angular/angular-cli/pull/23086/files#diff-7c1a5afe62ea7b34798fd2016b3cf81eecfbbf41ebaa2e35f7c884509349372e is merged.

package.json
jbedard: I'll take a quick look but I'd rather stick with the minimal changes and just get this done asap

jbedard: Well that was even easier then I thought it would be 👍

clydin (approved):

PR #23086, opened 2022-05-05
test: remove unused http request helper utility

clydin:
With the use of the `node-fetch`, the http E2E helper utility module is now unused within the E2E tests and can be removed.
This also removes usage of the `request` package which is not an actual dependency of the project.


alan-agius4 (dismissed):

alan-agius4 (dismissed):

jbedard:
Can we merge https://github.com/angular/angular-cli/pull/23079 first? 🙏

alan-agius4 (approved):

PR #23095, opened 2022-05-06
fix(@schematics/angular): remove extra space in standalone imports

cexbrayat:
There is an extra space after the new `imports` decorator option in standalone components, whereas there are none after all the other existing options.


alan-agius4 (approved):

alan-agius4:
Thanks for this.

cexbrayat:
@alan-agius4 Not the greatest contribution of all time, but the linters over the world will be happier 😉

PR #23096, opened 2022-05-06
ci: disable redundant Angular bot size check

clydin:
The E2E test suite contains a production build test (`build/prod-build`) that performs
a size comparison check to ensure a production build of a newly generated application
stays within 10% of a predefined value (currently 124,000 bytes).
As a result of this preexisting check, the need to perform another check that involves
the added complexity of Github status hooks and API calls is currently not needed.

### DIFF
--- a/.github/angular-robot.yml
+++ b/.github/angular-robot.yml
@@ -93,6 +93,8 @@ triage:
 
 # Size checking
 size:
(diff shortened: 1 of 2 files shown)


josephperrott (approved):
LGTM

PR #23106, opened 2022-05-09
fix(@angular-devkit/build-angular): properly handle locally-built APF v14 libraries

devversion:
Locally-built APF v14 libraries should be resolved properly. Webpack
currently does not resolve them (in e.g. `dist/`) because the local
distribution folders are not marked as module roots, causing Webpack
to never hit the `resolve-in-package` with the proper description file (of
the locally built-package) and therefore skipping package exports resolution
and breaking secondary entry-points from being resolved properly (when bundling).

We fix this by determining local module/package roots through the
tsconfig. Ideally we could have leveraged information from other targets
in the project but the builder does not have access to that, and
leveraging tsconfig `paths` might also make this more flexible.

### DIFF
--- a/packages/ngtools/webpack/src/paths-plugin.ts
+++ b/packages/ngtools/webpack/src/paths-plugin.ts
@@ -8,6 +8,7 @@
 
 import * as path from 'path';
 import { CompilerOptions } from 'typescript';
+
 import type { Configuration } from 'webpack';
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
@@ -15,6 +16,9 @@ export interface TypeScriptPathsPluginOptions extends Pick<CompilerOptions, 'pat
 
 // Extract Resolver type from Webpack types since it is not directly exported
 type Resolver = Exclude<Exclude<Configuration['resolve'], undefined>['resolver'], undefined>;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type DoResolveValue = any;
 
 interface PathPattern {
   starIndex: number;
@@ -154,8 +158,29 @@ export class TypeScriptPathsPlugin {
         // For example, if the first one resolves, any others are not needed and do not need
         // to be created.
         const replacements = findReplacements(originalRequest, this.patterns);
+        const basePath = this.baseUrl ?? '';
 
-        const tryResolve = () => {
+        const attemptResolveRequest = (request: DoResolveValue): Promise<DoResolveValue | null> => {
+          return new Promise((resolve, reject) => {
+            resolver.doResolve(
+              target,
+              request,
+              '',
+              resolveContext,
+              (error: Error | null, result: DoResolveValue) => {
+                if (error) {
+                  reject(error);
+                } else if (result) {
+                  resolve(result);
+                } else {
+                  resolve(null);
+                }
+              },
+            );
+          });
+        };
+
+        const tryNextReplacement = () => {
           const next = replacements.next();
           if (next.done) {
             callback();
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/utils/local-module-roots-tsconfig.ts
```diff
@@ -0,0 +1,94 @@
...
+  for (const [key, mappings] of Object.entries(tsConfig.options.paths)) {
+    for (const mapping of mappings) {
+      let target = resolve(basePath, normalize(mapping));
+
+      // Omit the trailing segments when a wildcard is used.
+      if (key[key.length - 1] === '*') {
+        target = target.slice(0, target.lastIndexOf('*')) as Path;
+      }
+
+      const packageRoot = await findPackageRoot(normalize(target), basePath);
```
devversion, line 42: I think the `baseUrl` is not the best "boundary" here, so I should likely remove that. Ideally limit it to the workspace path

packages/ngtools/webpack/src/paths-plugin.ts
```diff
@@ -155,39 +159,61 @@ export class TypeScriptPathsPlugin {
...
             return;
           }
 
-          const potentialRequest = {
+          // Resolution in the target location, preserving the original request.
+          // This will work with the `resolve-in-package` resolution hook, supporting
+          // package exports for e.g. locally-built APF libraries.
+          const potentialRequestAsPackage = {
+            ...request,
+            path: path.resolve(this.baseUrl ?? '', next.value),
```
alan-agius4, line 195: Nit: you can store `path.resolve(this.baseUrl ?? '', next.value),` in a const sine you are accessing it multiple times

tests/legacy-cli/e2e/tests/build/library-with-demo-app.ts
```diff
@@ -0,0 +1,87 @@
...
+  await updateJsonFile('tsconfig.json', (json) => {
+    json.compilerOptions.paths = {
+      'mylib': ['dist/mylib'],
+      'mylib/secondary': ['dist/mylib/secondary'],
+      'mylib/another': ['dist/mylib/another'],
+    };
+    return json;
+  });
+
+  await ng('build', 'mylib');
```
alan-agius4, line 63: No need to rebuild the library here

tests/legacy-cli/e2e/tests/build/library-with-demo-app.ts
```diff
@@ -0,0 +1,87 @@
...
+  await ng('build');
+
+  // Scenario #2 where we don't use wildcard path mappings.
+  await updateJsonFile('tsconfig.json', (json) => {
+    json.compilerOptions.paths = {
+      'mylib': ['dist/mylib'],
+      'mylib/secondary': ['dist/mylib/secondary'],
+      'mylib/another': ['dist/mylib/another'],
+    };
+    return json;
```
alan-agius4, line 60: Redundant return.

alan-agius4 (dismissed):
Couple of NITs, otherwise LGTM

alan-agius4 (dismissed):

alan-agius4 (approved):

alan-agius4 (approved):
LGTM, thanks for this @devversion!

packages/ngtools/webpack/src/paths-plugin.ts
```diff
@@ -154,40 +158,73 @@ export class TypeScriptPathsPlugin {
...
           const next = replacements.next();
           if (next.done) {
             callback();
 
             return;
           }
 
-          const potentialRequest = {
+          const targetPath = path.resolve(basePath, next.value);
+          const canBeModule = path.extname(targetPath) !== '';
```
alan-agius4, line 192: should this be
```suggestion
          const canBeModule = path.extname(targetPath) === '';
```

devversion: As discussed offline, the tests worked because the TS file actually exist and just when package resolution would come first, Webpack would transpile the file and the `module-resolution` spec would pass unexpectedly. If file resolution comes first (like it does now), then Webpack will complain about the TS file being outside of the sources/compilation.

alan-agius4 (dismissed):

packages/ngtools/webpack/src/paths-plugin.ts
```diff
@@ -154,40 +158,75 @@ export class TypeScriptPathsPlugin {
...
-              }
-            },
-          );
+          // Resolution in the original callee location, but with the updated request
+          // to point to the mapped target location.
+          const potentialRequestAsFile = {
+            ...request,
+            request: targetPath,
+            typescriptPathMapped: true,
+          };
```
clydin, line 211: nit: Could probably reduce the amount of code needed by creating another generator function (`createReplacementRequests` ?) that produces the request objects to try. Each replacement would either yield one or two values depending on the extension check.
Then this could be setup to be `const requests = createReplacementRequests(replacements);` and then `requests` would be used instead of `replacements`.

packages/ngtools/webpack/src/paths-plugin.ts
```diff
@@ -154,40 +158,75 @@ export class TypeScriptPathsPlugin {
...
+
+          let resultPromise = attemptResolveRequest(potentialRequestAsFile);
+
+          // If the request can be a module, we configure the resolution to try package/module
+          // resolution if the file resolution errored or did not have a result.
+          if (canBeModule) {
+            resultPromise = resultPromise
+              .then((result) => result ?? attemptResolveRequest(potentialRequestAsPackage))
+              .catch(() => attemptResolveRequest(potentialRequestAsPackage));
```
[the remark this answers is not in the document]
devversion: Good question, can we be sure about this? We can likely just see how it goes. The module resolution case should be covered by the integration test.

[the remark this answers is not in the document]
clydin: Documentation on creating custom plugins is unfortunately limited but from a review of the `enhanced-resolve` package, `null` appears to be the result value for a not found condition.  Combined with this previously being setup that way, i think we can be confident in that behavior.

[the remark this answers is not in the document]
devversion: Seems to work, updated. good idea!

packages/ngtools/webpack/src/paths-plugin.ts
devversion: Given you declaring this as a `nit`, I assume we can clean this up in a follow-up. I don't want to make significant further changes to this PR which already took up a surprising amount of time (due to exploring various approaches, reversing to understand enhanced-resolve etc.)

clydin (approved):

alan-agius4 (approved):

clydin (approved):

PR #23110, opened 2022-05-10
fix(@schematics/angular): don't add path mapping to old entrypoint file

alan-agius4:
With the APF changes in version 14, the dts entrypoint file has been changed to `index.d.ts`, this results in TypeScript being able to resolve the type definition file without the need of the additional path to the flat module file dts.


devversion (approved):

clydin (approved):
Nice improvement

PR #23114, opened 2022-05-10
test: use random ports for local verdaccio npm servers

jbedard:
Best viewed [ignoring whitespace](https://github.com/angular/angular-cli/pull/23114/files?w=1).

When working on the bazel e2e tests (#23074) we might run tests in parallel such as bazel + non-bazel, npm + yarn etc.

Unfortunately the verdaccio `--listen` only sets the primary port and not the fallback which we need to also set, so we need to do the `.replace` thing :(

### DIFF
--- a/tests/legacy-cli/e2e/utils/registry.ts
+++ b/tests/legacy-cli/e2e/utils/registry.ts
@@ -1,17 +1,24 @@
-import { ChildProcess, spawn } from 'child_process';
-import { copyFileSync, mkdtempSync, realpathSync } from 'fs';
+import { spawn } from 'child_process';
+import { mkdtempSync, realpathSync } from 'fs';
 import { tmpdir } from 'os';
 import { join } from 'path';
-import { writeFile } from './fs';
+import { getGlobalVariable } from './env';
+import { writeFile, readFile } from './fs';
 
-export function createNpmRegistry(withAuthentication = false): ChildProcess {
+export async function createNpmRegistry(
+  port: number,
+  httpsPort: number,
+  withAuthentication = false,
+) {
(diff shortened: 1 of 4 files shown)


tests/legacy-cli/e2e/utils/registry.ts
```diff
@@ -1,17 +1,24 @@
...
   // Setup local package registry
   const registryPath = mkdtempSync(join(realpathSync(tmpdir()), 'angular-cli-e2e-registry-'));
 
-  copyFileSync(
+  let configContent = await readFile(
     join(__dirname, '../../', withAuthentication ? 'verdaccio_auth.yaml' : 'verdaccio.yaml'),
-    join(registryPath, 'verdaccio.yaml'),
   );
+  configContent = configContent.replace(/\$\{HTTP_PORT\}/g, String(port));
+  configContent = configContent.replace(/\$\{HTTPS_PORT\}/g, String(httpsPort));
```
alan-agius4, line 20: NIT 
```suggestion
 configContent = configContent
  .replace(/\$\{HTTP_PORT\}/g, String(port))
  .replace(/\$\{HTTPS_PORT\}/g, String(httpsPort));
```

alan-agius4 (approved):
LGTM, thanks

tests/legacy-cli/e2e/utils/registry.ts
jbedard: 👍 but looks like it's too late...

alan-agius4: @jbedard, feel free to include this suggested change in on of your upcoming PRs.

PR #23130, opened 2022-05-12
test: fix ts compiler error

jbedard:


jbedard:
I've added a commit adding the BUILDs for the legacy-cli where all the compiler errors are and added the compilation of these to circleci. With all the changes landing keeping this PR up to date with the fixes without the actual bazel build (https://github.com/angular/angular-cli/pull/23074) is quite the hassle. Adding the actual bazel build to CI ensures this PR is actually up to date and correct.

While the bazel build of legacy-cli is unnecessary right now I think it's worth avoiding compilation errors. When these tests are moved to bazel (https://github.com/angular/angular-cli/pull/23074) this will be required anyway, which hopefully will be finished soon 🤞

tests/legacy-cli/e2e/tests/build/platform-server.ts
```diff
@@ -12,11 +12,13 @@ export default async function () {
 
   const isSnapshotBuild = getGlobalVariable('argv')['ng-snapshots'];
   if (isSnapshotBuild) {
-    const packagesToInstall = [];
+    const packagesToInstall: string[] = [];
     await updateJsonFile('package.json', (packageJson) => {
       const dependencies = packageJson['dependencies'];
       // Iterate over all of the packages to update them to the snapshot version.
-      for (const [name, version] of Object.entries(snapshots.dependencies)) {
+      for (const [name, version] of Object.entries(
```
clydin, line 19: nit: I think you could use `Object.entries<string>(` here instead of the argument cast.

jbedard:
I've reverted the addition of `webdriver-manager` to the package.json since #23245 switches to the one within protractor anyway (`node_modules/protractor/bin/webdriver-manager`). Since it's only used in a `require.resolve` it's a runtime dependency anyway and doesn't isn't required to compile.

clydin (approved):
LGTM.
Some of these changes would fail linting checks but these files are not linted yet and would need other cleanup prior to that anyway.

jbedard:
I can maybe do linting next but really need type checking before anything else :+1:

PR #23131, opened 2022-05-13
Remove out-of-date names in documentation

dgp1130:
Analytics approvers and caretaker pairings are out of date. Updated to not list explicit names at all so this doesn't require updating as consistently.

FYI @mgechev as current DevRel lead.
FYI @jelbourn as current UberTL who is now dropped from explicit analytics approvals in favor of the tooling TL (I assume you're fine with me and Minko making analytics calls).


jelbourn (approved):
LGTM

clydin (approved):

PR #23145, opened 2022-05-17
fix(@angular/cli): skip prompt or warn when setting up autocompletion without a global CLI install

### LINKED ISSUE #23135: Problems installing ng14 cli completion on macOS
# 🐞 Bug report

- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?
No, a new feature

 

### Description
I went to play with 14rc1 but since I don't have ng installed globally, I issued the following command `npx @angular/cli new ng14`
(from the WebStorm embedded terminal).

Once the new cli got installed, I was asked if I want the cli completion to be enabled. Answered yes.
The result in the terminal was something like: `source<(ng completion script)`

The completion did not work OOTB or even in a new terminal tab, decided to restart.
After the restart saw this in the reopened terminal:

image 2022-05-14 20-11-29

Then I realized the embedded terminal in WebStorm might have not been elevated so I repeated the entire `ng new` process from macOS Terminal. Unfortunately, I couldn't force the completion prompt to appear again. Now `ng new` proceeds now straight to the routing question. Don't really know how to mitigate this.

## 🔥 Exception or Error

`ng g c my-comp [tab]` doesn't produce any output as expected, just a void macOS sound

## 🌍 Your Environment

I don't have ng installed globally so `ng version` results in ng command not found.

<pre><code>

macOS Monterey 12.3.1
nvm
node -v 18.1

</code></pre>

**Anything else relevant?**

dgp1130:
If the user does not have a global install of the Angular CLI, the autocompletion prompt is skipped and `ng completion` emits a warning. The reasoning for this is that `source <(ng completion script)` won't work without `ng` on the `$PATH`, which is only really viable with a global install. Local executions like `git clone ... && npm install && npm start` or ephemeral executions like `npx @angular/cli` don't benefit from autocompletion and unnecessarily impede users.

A global install of the Angular CLI is detected by running `which -a ng`, which appears to be a cross-platform means of listing all `ng` commands on the `$PATH`. We then look over all binaries in the list and exclude anything which is a directo child of a `node_modules/.bin/` directory. These include local executions and `npx`, so the only remaining locations should be global installs (`/usr/bin/ng`, NVM, etc.).

The tests are a little awkward since `ng` is installed globally by helper functions before tests start. These tests uninstall the global CLI and install a local, project-specific version to verify behavior, before restoring the global version. Hypothetically this could be emulated by manipulating the `$PATH` variable, but `which` needs to be available (so we can't clobber the whole `$PATH`) and `node` exists in the same directory as the global `ng` command (so we can't remove that directory anyways). There's also no good way of testing the case where `which` fails to run.

Closes #23135.

/cc @alan-agius4

### DIFF
--- a/packages/angular/cli/src/commands/completion/cli.ts
+++ b/packages/angular/cli/src/commands/completion/cli.ts
@@ -8,14 +8,10 @@
 
 import { join } from 'path';
 import yargs, { Argv } from 'yargs';
-import {
-  CommandModule,
-  CommandModuleImplementation,
-  CommandScope,
-} from '../../command-builder/command-module';
+import { CommandModule, CommandModuleImplementation } from '../../command-builder/command-module';
 import { addCommandModuleToYargs } from '../../command-builder/utilities/command';
 import { colors } from '../../utilities/color';
-import { initializeAutocomplete } from '../../utilities/completion';
+import { hasGlobalCliInstall, initializeAutocomplete } from '../../utilities/completion';
 
 export class CompletionCommandModule extends CommandModule implements CommandModuleImplementation {
   command = 'completion';
@@ -43,6 +39,16 @@ Appended \`source <(ng completion script)\` to \`${rcFile}\`. Restart your termi
     ${colors.yellow('source <(ng completion script)')}
       `.trim(),
     );
+
+    if ((await hasGlobalCliInstall()) === false) {
+      this.context.logger.warn(
+        'Setup completed successfully, but there does not seem to be a global install of the' +
+          ' Angular CLI. For autocompletion to work, the CLI will need to be on your `$PATH`, which' +
+          ' is typically done with the `-g` flag in `npm install -g @angular/cli`.' +
+          '\n\n' +
+          'For more information, see https://angular.io/cli/completion#global-install',
+      );
+    }
 
     return 0;
   }
--- a/packages/angular/cli/src/utilities/completion.ts
+++ b/packages/angular/cli/src/utilities/completion.ts
@@ -7,6 +7,7 @@
  */
 
 import { json, logging } from '@angular-devkit/core';
+import { execFile } from 'child_process';
 import { promises as fs } from 'fs';
 import * as path from 'path';
 import { env } from 'process';
@@ -78,6 +79,16 @@ Appended \`source <(ng completion script)\` to \`${rcFile}\`. Restart your termi
     `.trim(),
   );
 
+  if ((await hasGlobalCliInstall()) === false) {
+    logger.warn(
+      'Setup completed successfully, but there does not seem to be a global install of the' +
+        ' Angular CLI. For autocompletion to work, the CLI will need to be on your `$PATH`, which' +
+        ' is typically done with the `-g` flag in `npm install -g @angular/cli`.' +
+        '\n\n' +
+        'For more information, see https://angular.io/cli/completion#global-install',
+    );
+  }
+
   // Save configuration to remember that the user was prompted.
   await setCompletionConfig({ ...completionConfig, prompted: true });
 
@@ -145,6 +156,12 @@ async function shouldPromptForAutocompletionSetup(
   const rcFiles = getShellRunCommandCandidates(shell, home);
   if (!rcFiles) {
     return false; // Unknown shell.
+  }
+
+  // Don't prompt if the user is missing a global CLI install. Autocompletion won't work after setup
+  // anyway and could be annoying for users running one-off commands via `npx` or using `npm start`.
+  if ((await hasGlobalCliInstall()) === false) {
+    return false;
   }
 
   // Check each RC file if they already use `ng completion script` in any capacity and don't prompt.
@@ -244,5 +261,57 @@ function getShellRunCommandCandidates(shell: string, home: string): string[] | u
     return ['.zshrc', '.zsh_profile', '.profile'].map((file) => path.join(home, file));
   } else {
     return undefined;
+  }
+}
+
+/**
+ * Returns whether the user has a global CLI install or `undefined` if this can't be determined.
+ * Execution from `npx` is *not* considered a global CLI install.
+ *
+ * This does *not* mean the current execution is from a global CLI install, only that a global
+ * install exists on the system.
+ */
+export async function hasGlobalCliInstall(): Promise<boolean | undefined> {
+  // List all binaries with the `ng` name on the user's `$PATH`.
+  const proc = execFile('which', ['-a', 'ng']);
+  let stdout = '';
+  proc.stdout?.addListener('data', (content) => {
+    stdout += content;
+  });
+  const exitCode = await new Promise<number | null>((resolve) => {
+    proc.addListener('exit', (exitCode) => {
+      resolve(exitCode);
+    });
+  });
+
+  switch (exitCode) {
+    case 0:
+      // Successfully listed all `ng` binaries on the `$PATH`. Look for at least one line which is a
+      // global install. We can't easily identify global installs, but local installs are typically
+      // placed in `node_modules/.bin` by NPM / Yarn. `npx` also currently caches files at
+      // `~/.npm/_npx/*/node_modules/.bin/`, so the same logic applies.
+      const lines = stdout.split('\n').filter((line) => line !== '');
+      const hasGlobalInstall = lines.some((line) => {
+        // A binary is a local install if it is a direct child of a `node_modules/.bin/` directory.
+        const parent = path.parse(path.parse(line).dir);
+        const grandparent = path.parse(parent.dir);
+        const localInstall = grandparent.base === 'node_modules' && parent.base === '.bin';
+
+        return !localInstall;
+      });
+
+      return hasGlobalInstall;
+    case 1:
+      // No instances of `ng` on the user's `$PATH`.
+      return false;
+    case null:
+      // `which` was killed by a signal and did not exit gracefully. Maybe it hung or something else
+      // went very wrong, so treat this as inconclusive.
+      return undefined;
+    default:
+      // `which` returns exit code 2 if an invalid option is specified and `-a` doesn't appear to be
+      // supported on all systems. Other exit codes mean unknown errors occurred. Can't tell whether
+      // CLI is globally installed, so treat this as inconclusive.
+      return undefined;
   }
 }
--- a/tests/legacy-cli/e2e/tests/misc/completion-prompt.ts
+++ b/tests/legacy-cli/e2e/tests/misc/completion-prompt.ts
@@ -2,7 +2,13 @@ import { promises as fs } from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import { env } from 'process';
-import { execAndCaptureError, execWithEnv } from '../../utils/process';
+import { getGlobalVariable } from '../../utils/env';
+import {
+  execAndCaptureError,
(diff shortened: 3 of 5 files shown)


packages/angular/cli/src/utilities/completion.ts
```diff
@@ -246,3 +270,38 @@ function getShellRunCommandCandidates(shell: string, home: string): string[] | u
...
+/**
+ * Returns whether the user has a global CLI install or `undefined` if this can't be determined.
+ * Execution from `npx` is *not* considered a global CLI install.
+ *
+ * This does *not* mean the current execution is from a global CLI install, only that a global
+ * install exists on the system.
+ */
+export async function hasGlobalCliInstall(): Promise<boolean | undefined> {
+  // List all binaries with the `ng` name on the user's `$PATH`.
+  const proc = execFile('which', ['-a', 'ng']);
```
alan-agius4, line 283: Wouldn't it simpler to just use `exec`, this also ensures cross-platform compatibility.


```ts
function hasGlobalAngularCLI(): Promise<boolean> {
  return exec(`ng help`, {
    stdio: "ignore",
  })
    .then(() => true)
    .catch(() => false);
}

```

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -72,12 +73,29 @@ Ok, you won't be prompted again. Should you change your mind, the following comm
...
+  if ((await hasGlobalCliInstall()) === false) {
+    logger.warn(
+      `
+${tags.oneLine`
+  Setup completed successfully, but there does not seem to be a global install of the Angular CLI.
+  For autocompletion to work, the CLI will need to be on your \`$PATH\`, which is typically done
+  with the \`-g\` flag in \`npm install -g @angular/cli\`.
+`}
+
+See https://angular.io/cli/completion#global-install for more info.
```
alan-agius4, line 94: I believe that information CTAs are all in the below format.

```suggestion
For more information, see https://angular.io/cli/completion#global-install
```

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -72,12 +73,29 @@ Ok, you won't be prompted again. Should you change your mind, the following comm
...
+`}
 
     ${colors.yellow(`source <(ng completion script)`)}
     `.trim(),
   );
 
+  if ((await hasGlobalCliInstall()) === false) {
+    logger.warn(
+      `
+${tags.oneLine`
```
alan-agius4, line 88: I think this would cause too much clutter and make it hard to read if it was in a single line.

Maybe better have it multiline?

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -72,12 +73,29 @@ Ok, you won't be prompted again. Should you change your mind, the following comm
   // Notify the user autocompletion was set up successfully.
   logger.info(
     `
-Appended \`source <(ng completion script)\` to \`${rcFile}\`. Restart your terminal or run the following to autocomplete \`ng\` commands:
+${tags.oneLine`
```
alan-agius4, line 76: Same as previous.

packages/angular/cli/src/commands/completion/cli.ts
```diff
@@ -38,12 +35,29 @@ export class CompletionCommandModule extends CommandModule implements CommandMod
 
     this.context.logger.info(
       `
-Appended \`source <(ng completion script)\` to \`${rcFile}\`. Restart your terminal or run the following to autocomplete \`ng\` commands:
+${tags.oneLine`
```
alan-agius4, line 38: Same.

packages/angular/cli/src/commands/completion/cli.ts
```diff
@@ -38,12 +35,29 @@ export class CompletionCommandModule extends CommandModule implements CommandMod
...
+`}
 
     ${colors.yellow('source <(ng completion script)')}
       `.trim(),
     );
 
+    if ((await hasGlobalCliInstall()) === false) {
+      this.context.logger.warn(
+        `
+${tags.oneLine`
```
alan-agius4, line 50: same.

packages/angular/cli/src/commands/completion/cli.ts
```diff
@@ -38,12 +35,29 @@ export class CompletionCommandModule extends CommandModule implements CommandMod
...
+  Setup completed successfully, but there does not seem to be a global install of the Angular CLI.
+  For autocompletion to work, the CLI will need to be on your \`$PATH\`, which is typically done
+  with the \`-g\` flag in \`npm install -g @angular/cli\`.
+`}
+
+See https://angular.io/cli/completion#global-install for more info.
+        `.trim(),
```
alan-agius4, line 57: same as before.

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -72,12 +73,29 @@ Ok, you won't be prompted again. Should you change your mind, the following comm
...
+  Appended \`source <(ng completion script)\` to \`${rcFile}\`. Restart your terminal or run the
+  following to autocomplete \`ng\` commands:
+`}
 
     ${colors.yellow(`source <(ng completion script)`)}
     `.trim(),
```
alan-agius4, line 82: same as before.

tests/legacy-cli/e2e/tests/misc/completion-prompt.ts
```diff
@@ -368,6 +376,54 @@ source <(ng completion script)
...
+      ...DEFAULT_ENV,
+      SHELL: '/bin/bash',
+      HOME: home,
+    });
+  });
+
+  // Does *not* prompt when a global CLI install is missing from the system.
+  await mockHome(async (home) => {
+    try {
+      // Temporarily uninstall the global CLI binary from the system.
```
alan-agius4, line 395: I think we are move away from the global installations altogether. At least based on one of Bazel PRs I've seen recently.

//cc @clydin might have more context.

packages/angular/cli/src/utilities/completion.ts
dgp1130: I don't think that would be sufficient because we need to ignore any local installs of `ng`. Since NPM scripts add the local `node_modules/.bin/` to the `$PATH`, your example would return `true` without ever looking beyond the local scope.

`which -a` is necessary to list *all* the CLI binaries on the `$PATH`, whereby we can exclude local ones and determine if any global installs exist.

packages/angular/cli/src/utilities/completion.ts
dgp1130: Chatted about this offline a bit. The goal here was to avoid introducing `\n` into the string for long paragraphs of text since it forces line wrapping in a terminal we don't know the full width of. I was hoping `tags.oneLine` would make that a little easier, but trying again with concatenated strings seems a bit more readable, so I went with that approach.

tests/legacy-cli/e2e/tests/misc/completion-prompt.ts
dgp1130: Yeah, I was thinking about that too. I'm hoping we're moving away from "always use the global CLI for tests" and doing something a little more local.

Unfortunately for this test, the behavior it's testing is reliant on finding a global CLI on the system, and I don't really see a way around that. Within a Bazel sandbox, we could try to put a CLI in particular place and load it into the `$PATH`, but I'm not sure how involved that will be. We could also fake things and just `touch $SOME_PATH_DIR/ng`, since we don't actually need to execute it in that situation.

The test now is a bit worse since the global CLI already exists and we have to remove it to test things correctly. If Bazel e2e tests don't rely on invoking the CLI via the global `$PATH`, then I think this test becomes a lot easier to write since we just need to add a fake global CLI rather than remove an existing real global CLI.

packages/angular/cli/src/utilities/completion.ts
dgp1130: Also updated this to properly parse `which` output as I realized the test was inadvertently relying on `which` failing due to the local CLI not being on the path, should be fixed now.

alan-agius4: > NPM scripts add the local `node_modules/.bin/` to the $PATH

AFAIK, this is old NPM behaviour prior to NPM 6 prior to NPX, and local `node_modules/.bin/` is no longer added to `$PATH`. 

I did try the above snippet, and locally I did work as expected, when `ng` is only installed locally it returned `false`, while when it is installed globally it return `true`.

```
$ which ng
ng not found

$ ls node_modules/.bin/ng
node_modules/.bin/ng

$ node --eval "require('child_process').execSync('ng help', { stdio: 'ignore' })"
child_process.js:866
    throw err;
    ^

Error: Command failed: ng help
    at checkExecSyncError (child_process.js:790:11)
    at Object.execSync (child_process.js:863:15)
    at [eval]:1:26
    at Script.runInThisContext (vm.js:134:12)
    at Object.runInThisContext (vm.js:310:38)
    at internal/process/execution.js:81:19
    at [eval]-wrapper:6:22
    at evalScript (internal/process/execution.js:80:60)
    at internal/main/eval_string.js:27:3 {
  status: 127,
  signal: null,
  output: [ null, null, null ],
  pid: 96183,
  stdout: null,
  stderr: null
}

$ npm i @angular/cli@next -g

added 206 packages, and audited 207 packages in 6s

25 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

$  node --eval "require('child_process').execSync('ng help', { stdio: 'ignore' })"
```

I am not too strongly, about using `which`, but the code is more involved compared to executing `ng` directly.

dgp1130: I don't think your test case of directly running `node --eval "require('child_process').execSync('ng help', { stdio: 'inherit' })"` in a project is accurate, Node doesn't modify the `$PATH` in this situation. But NPM does for NPM scripts, so if you do the same thing in script and execute that, a local CLI will successfully execute.

```jsonc
{
  "scripts": {
    "path-test": "node --eval \"require('child_process').execSync('ng help', { stdio: 'inherit' })\""
  },
  // ...
}
```

```
douglasparker@basilisk (~/Source/ng-new)
[22-05-18 10:13:48]$ node --eval "require('child_process').execSync('ng help', { stdio: 'inherit' })"
/bin/sh: line 1: ng: command not found

# Snip...
douglasparker@basilisk (~/Source/ng-new)
[22-05-18 10:13:51]$ npm run path-test

> ng-new@0.0.0 path-test
> node --eval "require('child_process').execSync('ng help', { stdio: 'inherit' })"

ng <command>

Commands:
  ng add <collection>            Adds support for an external library to your project.
  ng analytics                   Configures the gathering of Angular CLI usage metrics. See https://angular.io/cli/usage-analytics-gathering
  ng build [project]             Compiles an Angular application or library into an output directory named dist/ at the given output path.                                           [aliases: b]

# Snip...
```

Since the CLI can be run in this script context (such as `npm run build` or `npm start`), we have to assume any CLI on the `$PATH` could originate from a local install, meaning we need `which -a` to filter them out.

dgp1130: I'll merge this now just so we can get in for the RC today, but I'm open to simplifying this if we can find a way that doesn't rely on `which -a`.

PR #23164, opened 2022-05-18
fix(@angular/cli): correctly handle `--search` option in `ng doc`

gkalpak:
When the `--search` option is passed to `ng doc`, it should open a URL that does a search across the whole website (instead of the default behavior of limiting the search to API docs).
Fix the `ng doc` command to implement the iteded behavior (previously, it did the opposite).


alan-agius4 (approved):
LGTM

alan-agius4:
@gkalpak, this doesn't apply cleanly on the patch branch. Hence I will change the target to rc.

PR #23172, opened 2022-05-18
build: add caretaker configuration to ng-dev config

josephperrott:
Add the caretaker configuration to set up being able to run both the carataker
check and the handoff commands.

The caretaker handoff command will operate using the angular-cli-caretaker group
which has already been seeded with the current information.

Caretakers can update the current caretakers in the angular-cli-caretaker group, granting them the necessary access for caretaking using the `ng-dev caretaker handoff` command.  Additionally, the caretaker can check the status of the repository and merge queues using `ng-dev caretaker check`.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -17,6 +17,15 @@ Each shift consists of two caretakers. The primary caretaker is responsible for
 merging PRs to `main` and patch whereas the secondary caretaker is responsible
 for the release.
 
+At the end of each caretaker's rotation, they should peform a handoff in which they provide
+information to the next caretaker about the current state of the repository and update the
+access group to now include the next caretaker and their secondary. To perform this update
+to the access group, the caretaker can run:
+
+```bash
+$ yarn ng-dev caretaker handoff
+```
+
 ## Merging PRs
 
 The list of PRs which are currently ready to merge (approved with passing status checks) can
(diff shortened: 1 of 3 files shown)


clydin (dismissed):

dgp1130:
Should this also use `target: rc`? Otherwise it will be dropped from the 14.0.x branch but be present in 13.3.x and 14.1.x?

dgp1130:
@clydin pointed out to me that `target: patch` also merges to RC during this period, so that should be fine. Apologies for the noise there.

josephperrott:
@dgp1130 Updated as requested.

docs/process/release.md
```diff
@@ -17,6 +17,15 @@ Each shift consists of two caretakers. The primary caretaker is responsible for
 merging PRs to `main` and patch whereas the secondary caretaker is responsible
 for the release.
 
+At the end of each caretaker's rotation, they should peform a handoff in which they provide
+information to the next caretaker about the current state of the repository and update the
+access group to now include the next caretaker and their secondary. To perform this update
+to the access group, the caretaker can run:
```
dgp1130, line 23: We should probably link to the caretaker rotation here, but I'm moving it this week anyways, so I'll follow up with a separate PR for that.

dgp1130 (approved):
Thanks!

PR #23179, opened 2022-05-19
build: do not depend on test build order

jbedard:
Was there another reason to do this that tests aren't covering?

### DIFF
--- a/tests/legacy-cli/e2e_runner.ts
+++ b/tests/legacy-cli/e2e_runner.ts
@@ -1,7 +1,5 @@
-// This may seem awkward but we're using Logger in our e2e. At this point the unit tests
-// have run already so it should be "safe", teehee.
-import { logging } from '@angular-devkit/core';
-import { createConsoleLogger } from '@angular-devkit/core/node';
+import { logging } from '../../packages/angular_devkit/core/src';
+import { createConsoleLogger } from '../../packages/angular_devkit/core/node';
 import * as colors from 'ansi-colors';
 import glob from 'glob';
 import yargsParser from 'yargs-parser';


alan-agius4 (approved):

tests/legacy-cli/e2e_runner.ts
```diff
@@ -1,7 +1,5 @@
-// This may seem awkward but we're using Logger in our e2e. At this point the unit tests
```
clydin, line 1: That comment is outdated.  The tests will work on a fresh git clone with yarn install.

jbedard: I think you're correct that this comment is wrong and this can remain as-is. Maybe initially I just had missing dependencies when converting e2e tests to bazel so I just switched to relative paths 🤷 😅 

Should I update the PR and just remove the comment? Or feel free to close it.

clydin: It's good to merge; maybe update the message.  Technically probably better to not rely on unbuilt packages like that anyway.

jbedard: I think it only works do to the `lib/bootstrap-local.js` which compiles it at runtime and resolves the path using the `tsconfig-test.json`? Otherwise the comment would be true and it would depend on the package being built first?

jbedard: I've updated the commit message but I'm not sure if it is 100% correct?

clydin: that works.

clydin (approved):

PR #23185, opened 2022-05-20
fix(@angular-devkit/build-angular): downlevel libraries based on the browserslist configurations

### LINKED ISSUE #23126: Different handling of EcmaScript version specific features in JS and TS code
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

I don't know.

### Description

I have a problem with Logical OR assignments:
When I use a Logical OR assignment in a TypeScript file and compile it with Angular Cli it is converted from `a ||= b` to `a || (a = b)` up to ES2020. With ES2021 it stays unchanged. 
If such a Logical OR assignment is used in a JS lib (in my case pdfjs-dist) it is converted until ES5, but from ES6/ES2015 on it stays unchanged. 

ES5
grafik

ES2015
grafik

The browserlist setting seem's to have no impact. 

## 🔬 Minimal Reproduction
Create a plain Angular project, install a lib that contains a logical or assignment like pdfjs-dist, play around with the target property in tsconfig.json. 

## 🔥 Exception or Error
Generated TS and JS code differs in usage of EcmaScript features. We detected it, because we have used an older Chrome version (83) for testing, which suddenly failed, because the `||=` was unkown. 

## 🌍 Your Environment

<pre><code>
Angular CLI: 13.3.5
Node: 16.14.2
Package Manager: npm 8.6.0
OS: win32 x64

Angular: 13.3.7

@angular-devkit/architect       0.1303.5
@angular-devkit/build-angular   13.3.5
@angular-devkit/core            13.3.5
@angular-devkit/schematics      13.3.5
@angular/cli                    13.3.5
@schematics/angular             13.3.5
rxjs                            7.5.5
typescript                      4.6.4
</code></pre>

**Anything else relevant?**

alan-agius4:
There is no standard for library authors to ship their library in different ES versions, which can result in vendor libraries to ship ES features which are not supported by one or more browsers that the user's application supports.

With this change, we will be downlevelling libraries based on the list of supported browsers which is configured in the browserslist configuration. Previously, we only downlevelled libraries when targeting ES 5.

The TypeScript target option will not effect how the libraries get downlevelled.

Closes #23126

### DIFF
--- a/packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
+++ b/packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
@@ -72,18 +72,26 @@ export default custom<ApplicationPresetOptions>(() => {
 
   return {
     async customOptions(options, { source, map }) {
-      const { i18n, scriptTarget, aot, optimize, instrumentCode, ...rawOptions } =
-        options as AngularBabelLoaderOptions;
+      const {
+        i18n,
+        scriptTarget,
+        aot,
+        optimize,
+        instrumentCode,
+        supportedBrowsers,
+        ...rawOptions
+      } = options as AngularBabelLoaderOptions;
 
       // Must process file if plugins are added
       let shouldProcess = Array.isArray(rawOptions.plugins) && rawOptions.plugins.length > 0;
 
       const customOptions: ApplicationPresetOptions = {
         forceAsyncTransformation: false,
-        forceES5: false,
+        forcePresetEnv: false,
         angularLinker: undefined,
         i18n: undefined,
         instrumentCode: undefined,
+        supportedBrowsers,
       };
 
       // Analyze file for linking
@@ -107,19 +115,34 @@ export default custom<ApplicationPresetOptions>(() => {
 
       // Analyze for ES target processing
       const esTarget = scriptTarget as ScriptTarget | undefined;
-      if (esTarget !== undefined) {
-        if (esTarget < ScriptTarget.ES2015) {
-          customOptions.forceES5 = true;
-        } else if (esTarget >= ScriptTarget.ES2017 || /\.[cm]?js$/.test(this.resourcePath)) {
-          // Application code (TS files) will only contain native async if target is ES2017+.
-          // However, third-party libraries can regardless of the target option.
-          // APF packages with code in [f]esm2015 directories is downlevelled to ES2015 and
-          // will not have native async.
-          customOptions.forceAsyncTransformation =
-            !/[\\/][_f]?esm2015[\\/]/.test(this.resourcePath) && source.includes('async');
-        }
-        shouldProcess ||= customOptions.forceAsyncTransformation || customOptions.forceES5 || false;
+      const isJsFile = /\.[cm]?js$/.test(this.resourcePath);
+
+      // The below should be dropped when we no longer support ES5 TypeScript output.
+      if (esTarget === ScriptTarget.ES5) {
+        // This is needed because when target is ES5 we change the TypeScript target to ES2015
+        // because it simplifies build-optimization passes.
+        // @see https://github.com/angular/angular-cli/blob/22af6520834171d01413d4c7e4a9f13fb752252e/packages/angular_devkit/build_angular/src/webpack/plugins/typescript.ts#L51-L56
+        customOptions.forcePresetEnv = true;
+        // Comparable behavior to tsconfig target of ES5
+        customOptions.supportedBrowsers = ['IE 9'];
+      } else if (isJsFile) {
+        // Applications code ES version can be controlled using TypeScript's `target` option.
+        // However, this doesn't effect libraries and hence we use preset-env to downlevel ES fetaures
+        // based on the supported browsers in browserlist.
+        customOptions.forcePresetEnv = true;
       }
+
+      if ((esTarget !== undefined && esTarget >= ScriptTarget.ES2017) || isJsFile) {
+        // Application code (TS files) will only contain native async if target is ES2017+.
+        // However, third-party libraries can regardless of the target option.
+        // APF packages with code in [f]esm2015 directories is downlevelled to ES2015 and
+        // will not have native async.
+        customOptions.forceAsyncTransformation =
+          !/[\\/][_f]?esm2015[\\/]/.test(this.resourcePath) && source.includes('async');
+      }
+
+      shouldProcess ||=
+        customOptions.forceAsyncTransformation || customOptions.forcePresetEnv || false;
 
       // Analyze for i18n inlining
       if (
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
```diff
@@ -107,18 +115,35 @@ export default custom<ApplicationPresetOptions>(() => {
 
       // Analyze for ES target processing
       const esTarget = scriptTarget as ScriptTarget | undefined;
-      if (esTarget !== undefined) {
-        if (esTarget < ScriptTarget.ES2015) {
-          customOptions.forceES5 = true;
-        } else if (esTarget >= ScriptTarget.ES2017 || /\.[cm]?js$/.test(this.resourcePath)) {
+
+      // The below should be dropped when we no longer support ES5 TypeScript output.
+      if (esTarget === ScriptTarget.ES5) {
```
alan-agius4, line 120: This is to retain the same behaviour, but eventually we should start ignoring the TypeScript target and rely on the browserslist configuration. Although that does seem like a breaking behaviour therefore probably that must happen in v15?

dgp1130: To make sure I understand, is the behavior here that when setting `target: "es5"`, we effectively ignore browserslist and always downlevel everything to ES5? Does that change existing application code, or only libraries because they are now downeleveled too?

I had a quick conversation with the rest of the leads and several parts of Angular no longer work on IE (some components and some security / sanitization stuff), but there are still aspects which could work and I think we're unwilling to actively break such use cases just yet.

We can probably emit a warning in this case (whether you want to do that in this PR or another one) to be clear that this isn't supported and the user is probably doing something they shouldn't be.

clydin: This maintains the existing behavior of downlevelling to ES5 compatible code for *both* application code and libraries when the TS target is set to `es5`.  The browserslist will continue to be ignored in this case.

dgp1130: Ah right, we were already doing that for ES5. So now we're doing it for all language levels, it's just that normally it's dependent on browserslist, but setting `target: "es5"` overrides that. I agree we should get rid of that override eventually, but I'm fine waiting for v15 to make that happen.

packages/angular_devkit/build_angular/src/builders/browser/tests/behavior/typescript-target_spec.ts
```diff
@@ -198,5 +199,31 @@ describeBuilder(buildWebpackBrowser, BROWSER_BUILDER_INFO, (harness) => {
...
+        return JSON.stringify(tsconfig);
+      });
+
+      harness.useTarget('build', {
+        ...BASE_OPTIONS,
+      });
+
+      const { result, logs } = await harness.executeOnce();
+      expect(result?.success).toBeTrue();
+      expect(logs).toContain(
```
dgp1130, line 220: Nit: Should this include an assertion that the deprecation warning is only emitted once? Seems like you had to put in some extra effort to limit it, so it would be good to test that behavior as well.

dgp1130 (approved):
Commit message nit: "fix(@angular-devkit/build-angular): **downlevel** libraries based on the browserlist configuration"

packages/angular_devkit/build_angular/src/builders/browser/tests/behavior/typescript-target_spec.ts
alan-agius4: Will do.

clydin (approved):

terencehonles:
I'm a bit confused after this change.

I was hoping it would just work, but it doesn't seem like my browser list is being respected. I have it defined in the `package.json` and I'm not sure if that might be the issue.

I previously had the tsconfig set to es5 to output for legacy browsers, and I had a check to make sure the built bundle was es5 to catch any libraries that snuck in which is what this change was meant to address.

After upgrading to Angular 14 the migration changed my tsconfig output setting, and since I had previously seen this change I figured it was fine and would just work. However, the es5 check is now complaining and the only way to make sure the build is correct is to set the target back to es5 which triggers the warning that es5 output is deprecated.

Is there some configuration option I am missing? I can obviously set the target back to es5 but I'm not sure why there was a migration to change it if this breaks now.

Thanks in advance and I can open an issue if needed.

alan-agius4:
Indeed ES5 output support is deprecated since Angular no longer supports legacy browsers such as IE 11 which require ES5 syntax. I am not really sure what you mean by `the es5 check is now complaining`.

Without more information such as a minimal reproduction it's hard wether or not there is an issue.

terencehonles:
@alan-agius4 I may want to just open an issue if more details are needed than this, but the ES5 check I am referring to is using https://github.com/acornjs/acorn after the build step to check if any libraries had slipped in non ES5 features since libraries weren't previously downleveled before this change. Previously our tsconfig output was set to es5, but after updating to Angular 14 and migrating it changed to es2020.

Based on the title of this PR, and what I thought the feature would do, is that the tsconfig output could now be set to es2020 which was previously used internally and all generated output would respect the browserslist configuration. I don't need to strictly check for es5 and I can remove acorn from our build process, but I _did_ notice that the output generated was still using features that were too new for our browserslist configuration and I'm wondering if I'm either not using an option I should be doing or if I'm misunderstanding what's supposed to happen and I should keep the tsconfig output at es5 until we can remove our need for that downleveling.

alan-agius4:
Hi @terencehonles, 

During the update to version 14 we do indeed update the target to ES2020, this is because all browsers that Angular supports ES2020 features out of the box without the need of any downlevelling or polyfilling features.
Out of curiosity, why do you have requirement to ship ES5? ES5 is larger in terms of bundle size.

 Prior to this change downlevelling of libraries was only done when targetting ES5.
The downlevelling only happens for 3rd party libraries and doesn't effect the TypeScript output. If the TypeScript target is configured to ES2020 it will remain as such in the final bundle. Downlevelling on TypeScript sources is only done when targeting ES5 and this is done because various parts of the build only support modern ES syntax.

Moving forward in another major version we want to change this and "ignore" the TypeScript target and downlevel ECMA features/syntax based on the configured browserslist configuration even on TypeScript output.

alan-agius4:
Hi @terencehonles, 

During the update to version 14 we do indeed update the target to ES2020, this is because all browsers that Angular supports ES2020 features out of the box without the need of any downlevelling or polyfilling features.
Out of curiosity, why do you have requirement to ship ES5? ES5 is larger in terms of bundle size.

Prior to this change downlevelling of libraries was only done when targetting ES5. Now downlevelling for libraries always happen. That said this change doesn't effect the TypeScript output when targetting ES2015 or later. Downlevelling of TypeScript sources is only done when targeting ES5 and this is done because various parts of the build only support modern ES syntax.

Moving forward in another major version we want to change this and "ignore" the TypeScript target and downlevel ECMA features/syntax based on the configured browserslist configuration even on TypeScript output.

PR #23198, opened 2022-05-23
fix(@angular/cli): handle format `path` in invoked schematics

alan-agius4:
With this change we change the how we handle `"format": "path"` schematic property option. Previously, in version 14 this format was only being applied on parent schematics that were registered as commands. With this change we apply the formatter on all options, even those who are called programmatically using APIs like `externalSchematic`.

### DIFF
--- a/packages/angular/cli/src/command-builder/schematics-command-module.ts
+++ b/packages/angular/cli/src/command-builder/schematics-command-module.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { schema, tags } from '@angular-devkit/core';
+import { normalize as devkitNormalize, isJsonObject, schema, tags } from '@angular-devkit/core';
 import { Collection, UnsuccessfulWorkflowExecution, formats } from '@angular-devkit/schematics';
 import {
   FileSystemCollectionDescription,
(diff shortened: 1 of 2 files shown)


packages/angular/cli/src/command-builder/schematics-command-module.ts
```diff
@@ -150,6 +152,32 @@ export abstract class SchematicsCommandModule
...
+        }
+
+        const property = properties[key];
+        if (!property || !isJsonObject(property)) {
+          continue;
+        }
+
+        if (property['format'] === 'path') {
+          (options as Record<string, unknown>)[key] = workingDir;
+        }
```
clydin, line 178: It looks like the logic previously set the default to the workingDir for only properties named `path` but this overrides the value with the workDir for all properties with a `path` format.  I think we might want to keep the `path` name condition to prevent accidentally breaking third-party schematics that could be using the `path` format for other purposes.

We should probably add in a Schematics specific `workingDirectory` concept (on the context maybe?) so we can avoid this whole workaround in the future.

alan-agius4: https://github.com/angular/angular-cli/pull/23212 is a different approach to this change.

PR #23208, opened 2022-05-24
ci: remove nightly CI run for 10.2.x branch

alan-agius4:
We disable the nightly CI runs for this branch as version 10 is no longer in LTS.


alan-agius4:
@dgp1130, since these are no longer supported. Can we merge this regardless of the failing CI step?

dgp1130:
@alan-agius4, yeah I don't think we care about CI anymore here.

PR #23212, opened 2022-05-24
fix(@angular/cli): populate path with working directory in nested schematics

alan-agius4:
With this change we change the how we handle `"format": "path"` schematic property option. We replace the formatter in favour of a `SmartDefaultProvider`, which ensures that nested schematics can access the `workingDirectory`.

### DIFF
--- a/packages/angular/cli/src/command-builder/command-module.ts
+++ b/packages/angular/cli/src/command-builder/command-module.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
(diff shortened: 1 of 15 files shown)


packages/schematics/angular/universal/schema.json
```diff
@@ -34,6 +34,9 @@
     "rootModuleFileName": {
       "type": "string",
       "format": "path",
+      "$default": {
+        "$source": "workingDirectory"
+      },
```
clydin, line 39: This property already has a default value defined.

PR #23220, opened 2022-05-25
docs: correct `completion.gif` image path

alan-agius4:
Assets paths in AIO need to be relative. This also ensures that the base href element is respected if changed. 

https://app.circleci.com/pipelines/github/angular/angular/46755/workflows/ea228e6e-62ab-4d7f-9764-2c1056ad0366/jobs/1173152


clydin (approved):

dgp1130:
Darn it, I knew I was going to screw this up when I had to split that PR. Thanks for fixing this!

Issue #22048, opened 2021-10-30
Incompatibility with yarn v3


kaname-png:
### Which @angular/* package(s) are the source of the bug?

Don't known / other

### Is this a regression?

Yes

### Description

When I try to use the "ng add" tool in any package I get that error always in yarn v3.

### Please provide the exception or error you saw

```true
PS app> ng add apollo-angular
ℹ Using package manager: yarn
✔ Found compatible package version: apollo-angular@2.6.0.
✔ Package information loaded.
 
The package apollo-angular@2.6.0 will be installed and executed.
Would you like to proceed? Yes
Unknown Syntax Error: Command not found; did you mean one of:

  0. yarn add [--json] [-E,--exact] [-T,--tilde] [-C,--caret] [-D,--dev] [-P,--peer] [-O,--optional] [--prefer-dev] [-i,--interactive] [--cached] [--mode #0] ...
  1. yarn add [--json] [-E,--exact] [-T,--tilde] [-C,--caret] [-D,--dev] [-P,--peer] [-O,--optional] [--prefer-dev] [-i,--interactive] [--cached] [--mode #0] ...

While running add apollo-angular@2.6.0 --silent
✖ Package install failed, see above.
```

### Please provide the environment you discovered this bug in

```true
Angular CLI: 12.2.7
Node: 16.11.0 (Unsupported)
Package Manager: yarn 3.0.2
OS: win32 x64

Angular: 12.2.8
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1202.7
@angular-devkit/build-angular   12.2.7
@angular-devkit/core            12.2.7
@angular-devkit/schematics      12.2.7
@angular/cdk                    12.2.12
@angular/cli                    12.2.7
@angular/flex-layout            12.0.0-beta.35
@angular/localize               12.2.12
@schematics/angular             12.2.7
rxjs                            6.6.7
typescript                      4.3.5
```

### Anything else?

I think it is the same with yarn v2 as the --silent flag no longer exists since yarn v2.

clydin:
Can you try manually running `yarn add apollo-angular@2.6.0 --silent` to see if that command successfully completes?

kaname-png:
@clydin same error

![image](https://user-images.githubusercontent.com/56084970/139705090-eb125824-23f1-447e-8a89-711adeb1a192.png)

[...]

PR #23242, opened 2022-05-27
fix(@angular/cli): support silent package installs with Yarn 2+

clydin:
Yarn 2 and higher no longer support the `--silent` flag.
The Angular CLI will now spawn package managers with the stdout ignored when silent
mode is needed and only show stderr when the process exits unsuccessfully. This
provides the necessary functionality for the CLI without relying on package manager
options.


alan-agius4 (approved):
This should resolve https://github.com/angular/angular-cli/issues/22048

Also, this change should also be done in https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/schematics/tasks/package-manager/install-task.ts

alan-agius4 (approved):

PR #23245, opened 2022-05-28
run e2e tests in isolated child processes

jbedard:
~~Note I'd like to merge https://github.com/angular/angular-cli/pull/23130 first.~~ DONE

This is to stop mixing the angular-cli test-runner environment with the cli environments being tested. Essentially launching each test in a subprocess to a) not mix with the test-runner PATH, npm modules etc, b) prevent environment (vars, cwd, ...) side effects from one test effecting the next tests or the test-runner.

Prefactor commits (can be merged on their own ahead of time instead of squashing):
* ~~Wait for `killAllProcesses` to actually complete before proceeding. Killing a procedure is async but was not being await-ed. This creates race conditions but once tests are run in a subprocess those are far more likely (the subprocess dies before killing subprocesses completes).~~ EDIT: was done in https://github.com/angular/angular-cli/commit/f70557afc11bccf549dd37d82484104865fa3fb4
* Refactor the main e2e_runner.ts into separate methods, async/await instead of `.then`.
* Distinguish test setup vs tests better (in preparation for tests being launched in subprocesses, also distinguishes setup vs test better in the stdout).

Main change:
* A specific version of npm + yarn are installed alongside the angular-cli npm package being tested. This will be required in bazel where nothing outside the sandbox (global node/npm/yarn) is accessible. This also sets specific version numbers which could be dynamic in the future.
* Tests run in a sandboxed subprocess environment using those specific npm + yarn + cli packages.
* Tests no longer have access to the node_modules or environment of the test-runner, only the modules installed within that sandboxed environment. So things like `puppeteer` now need to be installed within the sandboxed test environment (previously it was using the one in the git repo node_modules). The `protractor` used in the tests is now the correct one in the test project node_modules (not the one the cli git repo node_modules), so `webdriver-manager` must now run on that test project and not only the git repo.

Overview:
* launch "setup" in the main process to setup the test environment (no real changes from before)
* launch each "initializer" in subprocesses (some of the previous "setup" scripts are now "initializers" which run in the subprocess, but no big changes)
* launch each test in the subprocesses


tests/legacy-cli/e2e_runner.ts
```diff
@@ -115,141 +116,162 @@ console.log(testsToRun.join('\n'));
...
+
+    const projectsRoot = getGlobalVariable('projects-root');
+
+    // Enter the projects root as the default cwd for test initializers
+    process.chdir(projectsRoot);
+
+    for (const [initializerIndex, initializer] of allInitializers.entries()) {
+      printSetupHeader(initializer, initializerIndex, allInitializers.length, 'initializer');
+
+      await runInitializer((lastTestRun = initializer));
```
gregmagolan, line 150: nit: comment that `runInitializer` runs `ng init` and creates `test-project`

tests/legacy-cli/e2e_runner.ts
```diff
@@ -115,141 +116,162 @@ console.log(testsToRun.join('\n'));
...
+  }
+
+  const currentFileName = relativeName.replace(/\.ts$/, '');
 
-      process.exit(1);
-    },
+  return { absoluteName, currentFileName };
+}
+
+async function runSetup(relativeName: string) {
```
gregmagolan, line 199: nit: add comment on this function. "how is this different than runInitializer?"

tests/legacy-cli/e2e_runner.ts
```diff
@@ -115,141 +116,162 @@ console.log(testsToRun.join('\n'));
...
+  const start = +new Date();
+
+  const module = require(absoluteName);
+
+  await (typeof module === 'function' ? module : module.default)();
+
+  printFooter(currentFileName, start);
+}
+
+async function runInitializer(relativeName: string) {
```
gregmagolan, line 210: nit: add comment on this function. "how is this different than runSetup?"

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -39,9 +39,29 @@ export async function prepareProjectForE2e(name) {
 
   console.log(`Project ${name} created... Installing packages.`);
   await installWorkspacePackages();
+  await installPackage('puppeteer');
```
gregmagolan, line 42: pin this to an exact version;

This can be done in `projects-root/node_modules` so that all "test-projects" can access it via node_modules resolution.

tests/legacy-cli/e2e/utils/test_process.ts
```diff
@@ -0,0 +1,20 @@
+const testScript: string = process.argv[2];
+const args = process.argv.slice(2).map((a) => (a === 'true' || a === 'false' ? a === 'true' : a));
```
gregmagolan, line 2: nit: this line can be removed now

tests/legacy-cli/e2e/initialize/500-create-project.ts
```diff
@@ -22,6 +22,10 @@ export default async function () {
     // Ensure local test registry is used when outside a project
     await setNPMConfigRegistry(true);
 
+    // Install puppeteer in the parent directory of the test projects for use
+    // by the CLI within any test project
+    await installPackage('puppeteer@14.1.1');
```
alan-agius4, line 27: NIT: would be great if we can use the same version of `package.json`. This would keep it in sync as otherwise I feel that we'll forget to update this.

tests/legacy-cli/e2e/setup/100-global-cli.ts
```diff
@@ -1,5 +1,8 @@
 import { getGlobalVariable } from '../utils/env';
-import { exec, silentNpm } from '../utils/process';
+import { silentNpm } from '../utils/process';
+
+const NPM_VERSION = '7.5.6';
```
alan-agius4, line 4: Should we use a more recent version? On the main branch we use `7.24.0`

tests/legacy-cli/e2e/initialize/500-create-project.ts
alan-agius4: SGTM.

tests/legacy-cli/e2e/tests/packages/webpack/test-app.ts
```diff
@@ -30,6 +30,4 @@ export default async function (skipCleaning: () => void) {
     'DISABLE_V8_COMPILE_CACHE': '1',
   });
   await expectFileToMatch('dist/app.main.js', 'AppModule');
-
-  skipCleaning();
```
jbedard, line 34: This didn't seem to effect tests at all? This was the only use case of the "skip cleaning" and removing it didn't cause any tests to start failing.

tests/legacy-cli/e2e/utils/process.ts
```diff
@@ -299,22 +300,21 @@ export function silentNpm(
       {
         silent: true,
         cwd: (options as { cwd?: string } | undefined)?.cwd,
-        env: extractNpmEnv(),
```
jbedard, line 302: None of these are needed anymore because the child process has the correct env passed in already, see `launchTestProcess`.

alan-agius4 (approved):

tests/legacy-cli/e2e/tests/basic/e2e.ts
```diff
@@ -59,13 +59,6 @@ export default function () {
       // Should run side-by-side with `ng serve`
       .then(() => execAndWaitForOutputToMatch('ng', ['serve'], / Compiled successfully./))
       .then(() => ng('e2e', 'test-project', '--dev-server-target='))
-      // Should fail without updated webdriver
```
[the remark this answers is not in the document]
alan-agius4: Yeah, this is because the webdriver manager is already update when running https://github.com/angular/angular-cli/pull/23245/files#diff-6f3e609e1aead9e976e142745edcbcf23cbe30c3f61a400146fed645dfd751b2R47-R63

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -42,6 +42,26 @@ export async function prepareProjectForE2e(name) {
 
   await ng('generate', 'e2e', '--related-app-name', name);
 
+  // Initialize welenium webdrivers.
```
alan-agius4, line 45:
```suggestion
  // Initialize selenium webdriver.
```

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -42,6 +42,26 @@ export async function prepareProjectForE2e(name) {
...
+    'node',
+    'node_modules/protractor/bin/webdriver-manager',
+    'update',
+    '--standalone',
+    'false',
+    '--gecko',
+    'false',
+    '--versions.chrome',
+    '101.0.4951.41',
+  );
```
alan-agius4, line 58: Nit 
```suggestion
  const webdriverCommand = exec(
    'node_modules/protractor/bin/webdriver-manager',
    'update',
    '--standalone',
    'false',
    '--gecko',
    'false',
    '--versions.chrome',
    '101.0.4951.41',
  );
```

tests/legacy-cli/e2e/tests/basic/e2e.ts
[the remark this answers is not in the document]
jbedard: Wasn't it before as well? The postinstall in the root package.json was running it and previously the tests were consuming that. Why is it different now that it's using the webdriver within the test project?

tests/legacy-cli/e2e/utils/project.ts
jbedard: Then repeat that twice? I did the `.bind` so I can execute it twice without having to repeat the args twice). Could also do `() => exec(...)` instead of `bind` which I guess might be the preferred style now (with es6/ts, unlike 10y ago)?

I'm running it twice because the `webdriver-manager` fails 100% of the time when first run locally, then passes the second time... so I ran it twice here as well otherwise I can't run the tests locally at all. Previously when this was run with the yarn postinstall I had to run yarn twice before I could run the tests, but now that the tests themselves do it I can't run them at all without running it twice here within the tests...

clydin (approved):

tests/legacy-cli/e2e/tests/basic/e2e.ts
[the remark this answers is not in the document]
alan-agius4: That's being for E2E's the local package would be used as such the repository package was not being consumed.

alan-agius4 (approved):

clydin:
This appears to be causing failures for the full Windows CI job and the snapshots job:
https://app.circleci.com/pipelines/github/angular/angular-cli/23169/workflows/2aa69b28-122e-468f-9220-f7a36f6ae56c

jbedard:
```
Error: Running "cmd.exe /c git clean -dxf" returned error. 1...
warning: failed to remove .angular/cache/14.1.0-next.0/ng-packagr/content-v2/sha512/03/ce/bb4940cc77a94bd0f18e4752d30a7c2895fb086c8b48ae6258a7f2628e0c348bd1e42161f20f7efcbd9b6ff60dda64a1c9710bc3cb530e2831201a1e63d2: Filename too long
warning: failed to remove .angular/cache/14.1.0-next.0/ng-packagr/content-v2/sha512/18/eb/2ce695e9c32252f3fde4a56bcabc8bc319ed487f0afbc5322d41c5c8c9bee6079f160ca19345a84baa8a46e9339d1c1fae76e00465bebf52bf13434791db: Filename too long
warning: failed to remove .angular/cache/14.1.0-next.0/ng-packagr/content-v2/sha512/41/50/af2048670b4a59c036628768bde838d380c8e46d771b0f0893b8ba0af3d05b0de0ec1606973bb4bd8672cd6751d1db6e39d3671240d45c409b844fddae41: Filename too long
...
```

Any ideas here? Should that folder be gitignored and not be cleaned anyway?

clydin:
https://github.com/angular/angular-cli/pull/23311 should fix the snapshot job. Windows will require more investigation though. 

These are the jobs that are only run on main (Windows has a subset run on PRs):
* `e2e-cli-node-16` https://github.com/angular/angular-cli/blob/9c26e2850c75f5af5919fc6af72be423622eda4a/.circleci/config.yml#L373
* `e2e-cli-ng-snapshots` https://github.com/angular/angular-cli/blob/9c26e2850c75f5af5919fc6af72be423622eda4a/.circleci/config.yml#L365-L369
*  `e2e-cli-win` https://github.com/angular/angular-cli/blob/9c26e2850c75f5af5919fc6af72be423622eda4a/.circleci/config.yml#L328-L332

alan-agius4:
I briefly took a look from my mobile, what is strange is that the files not being removed are ng-packagr cache files, which strangely isn’t even run in the failings tests. Have them been left overs from previous tests?

I also noticed that some other tests are failing with task kill related errors
```
Error: Command failed: taskkill /pid 1892 /T /F
ERROR: The process "1892" not found.
```

PR #23248, opened 2022-05-30
fix(@angular-devkit/core): address Node.js `DEP0128` warning

alan-agius4:
Modules that have an invalid `main` entry as `index.js` file in the top level directory will resolve the `index.js` file. That is deprecated and is going to throw an error in future Node.js versions.

See: https://nodejs.org/api/deprecations.html#DEP0128


alan-agius4:
Blocked until merge freeze is lifted.

PR #23259, opened 2022-06-01
fix(@angular/cli): don't try to setup completion in CI environments

alan-agius4:
Auto completion isn't useful in CI environments and in some cases commands are run in parallel which causes race conditions.

Example: https://angular-team.slack.com/archives/C46U16D4Z/p1654089933910829


alan-agius4:
Marked as blocked until merge freeze is lifted.

dgp1130:
I thought I already accounted for turning this off in CI? I think `isTTY()` checks for the `CI` environment variable:
https://github.com/angular/angular-cli/blob/8da926966e9f414ceecf60b89acd475ce1b55fc5/packages/angular/cli/src/utilities/tty.ts#L21

There's even an existing test for this:
https://github.com/angular/angular-cli/blob/94228620e5f7d1f134431c745b9c247064a737b7/tests/legacy-cli/e2e/tests/misc/completion-prompt.ts#L283-L294

alan-agius4:
@dgp1130 good catch, so there must be something else going on.

Issue #23289, opened 2022-06-03
Angular CLI 14: running `ng test --watch=false` on Angular 13 or 12 project does not end process when complete


Blackbaud-PaulCrowder:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, globally installing Angular CLI 13 worked with both Angular 13 and Angular 12 projects when running `ng test --watch=false`.

### Description

Installing Angular CLI 14 globally and running `ng test --watch=false` on an Angular 13 or Angular 12 application results in the process never terminating. This causes CI builds to hang indefinitely on any step that runs Angular unit tests using this command.

## 🔬 Minimal Reproduction

- Run `npm i @angular/cli@13 -g` to Install Angular CLI 13.
- Run `ng new` to create a new Angular application. I chose to add Angular routing and selected SCSS for the stylesheet format. Selecting SCSS seems to be required to reproduce the issue, but adding routing does not.
- Run `ng test --watch=false` and observe the tests running, then the process terminating after completion.
- Run `npm i @angular/cli@14 -g` to install Angular CLI 14.
- Run `ng test --watch=false`. The message `Your global Angular CLI version (14.0.0) is greater than your local version (13.3.7). The local Angular CLI version is used.` will be displayed. Observe the tests running, then the process continuing to run even after the tests complete.

## 🔥 Exception or Error

<pre><code>N/A</code></pre>

## 🌍 Your Environment

macOS 12.4
node 14.17.6
npm 6.14.15

### Global `ng version`
<pre><code>Angular CLI: 14.0.0
Node: 14.17.6
Package Manager: npm 6.14.15 
OS: darwin x64

Angular: 
... 

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1400.0 (cli-only)
@angular-devkit/core         14.0.0 (cli-only)
@angular-devkit/schematics   14.0.0 (cli-only)
@schematics/angular          14.0.0 (cli-only)</code></pre>

### `ng version` inside application directory
<pre><code>Angular CLI: 13.3.7
Node: 14.17.6
Package Manager: npm 6.14.15
OS: darwin x64

Angular: 13.3.11
[...]

PR #23298, opened 2022-06-07
fix(@angular/cli): ensure full process exit with older local CLI versions

clydin:


alan-agius4 (approved):
NIT: can you update the commit to close `#23289`?

PR #23304, opened 2022-06-07
fix(@angular/cli): don't redirect `@schematics/angular` package

alan-agius4:
When a schematic has a direct dependency on `@schematics/angular` it should be favored instead of the one relative to the Angular CLI.

### DIFF
--- a/packages/angular/cli/src/command-builder/utilities/schematic-engine-host.ts
+++ b/packages/angular/cli/src/command-builder/utilities/schematic-engine-host.ts
@@ -146,10 +146,10 @@ function wrap(
 
       return builtinModule;
     } else if (id.startsWith('@angular-devkit/') || id.startsWith('@schematics/')) {
-      // Files should not redirect `@angular/core` and instead use the direct
-      // dependency if available. This allows old major version migrations to continue to function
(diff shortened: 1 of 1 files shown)


packages/angular/cli/src/command-builder/utilities/schematic-engine-host.ts
```diff
@@ -146,10 +146,10 @@ function wrap(
...
       return builtinModule;
     } else if (id.startsWith('@angular-devkit/') || id.startsWith('@schematics/')) {
-      // Files should not redirect `@angular/core` and instead use the direct
-      // dependency if available. This allows old major version migrations to continue to function
-      // even though the latest major version may have breaking changes in `@angular/core`.
-      if (id.startsWith('@angular-devkit/core')) {
+      // Files should not redirect `@angular-devkit/core` and `@schematics/angular`.
+      // Instead use the direct dependency if available. This allows old major version migrations and schematics (ex: `ng add @angular/pwa`)
+      // to continue to function even though the latest major version may have breaking changes.
+      if (id.startsWith('@angular-devkit/core') || id.startsWith('@schematics/angular')) {
```
alan-agius4, line 152: I feel that this should be true for all packages and not just these 2. Thoughts @clydin?

PR #23309, opened 2022-06-07
ci: disable outdated Angular robot merge and triage behavior

clydin:
The issue triage behavior is no longer used within this repository.
For the merge behavior, only the label status checks are potentially
beneficial but they are already checked via the `ng-dev` tooling
when merging.


josephperrott (approved):
LGTM

PR #23315, opened 2022-06-08
Windows E2E CI fixes

alan-agius4:
See each commit. 

Green run https://app.circleci.com/pipelines/github/angular/angular-cli/23222/workflows/8afc60f4-ed0b-4158-b48f-0904e8855b76/jobs/308980/parallel-runs/7?filterBy=ALL

FYI: @jbedard

### DIFF
--- a/tests/legacy-cli/e2e/utils/process.ts
+++ b/tests/legacy-cli/e2e/utils/process.ts
@@ -62,7 +62,6 @@ function _exec(options: ExecOptions, cmd: string, args: string[]): Promise<Proce
   }
 
   const childProcess = child_process.spawn(cmd, args, spawnOptions);
-  // @ts-ignore
   childProcess.stdout!.on('data', (data: Buffer) => {
     stdout += data.toString('utf-8');
     if (options.silent) {
@@ -74,7 +73,7 @@ function _exec(options: ExecOptions, cmd: string, args: string[]): Promise<Proce
       .filter((line) => line !== '')
       .forEach((line) => console.log('  ' + line));
   });
-  // @ts-ignore
+
   childProcess.stderr!.on('data', (data: Buffer) => {
     stderr += data.toString('utf-8');
     if (options.silent) {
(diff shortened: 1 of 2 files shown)


tests/legacy-cli/e2e/utils/process.ts
```diff
@@ -197,22 +196,31 @@ export function waitForAnyProcessOutputToMatch(
...
-            }
-          });
-        }),
-    ),
-  );
+  const processesToKill: Promise<void>[] = [];
+
+  while (_processes.length) {
+    const childProc = _processes.pop();
+    if (!childProc) {
```
jbedard, line 203: When would `childProc` not be set? I think it always would be since nothing in this loop is async...?

jbedard (approved):

jbedard:
Thanks for looking into and fixing both of those!

tests/legacy-cli/e2e/utils/process.ts
alan-agius4: `pop` can technically return undefined when there are no items in the array.

jbedard: Everything is synchronous until the `await` at the end of the method though so the `_processes.length` check is still valid.

Doesn't hurt, but I think this `if` is unnecessary 🤷

tests/legacy-cli/e2e/utils/process.ts
```diff
@@ -197,22 +196,31 @@ export function waitForAnyProcessOutputToMatch(
...
+            reject(err);
+          } else {
+            resolve();
+          }
+        });
+      }),
+    );
+  }
 
-  _processes = [];
```
jbedard, line 215: Since this was after the `await` I guess that means processes could have been added to this array during the `await` and then we'd clear it here and lose those added processes. Was that the cause of one of the issues? (in addition to the `'not found'`)

PR #23318, opened 2022-06-08
fix(@angular-devkit/build-angular): log modified and removed files when using the `verbose` option

alan-agius4:
With this change we print out the modified and removed files when running a build in verbose mode. This can be useful to debug builds that rebuilds multiple times without an apparent file change.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser/tests/options/verbose_spec.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser/tests/options/verbose_spec.ts
@@ -1,0 +1,106 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/builders/browser/tests/options/verbose_spec.ts
```diff
@@ -0,0 +1,106 @@
...
+const VERBOSE_LOG_TEXT = 'LOG from webpack.';
+
+describeBuilder(buildWebpackBrowser, BROWSER_BUILDER_INFO, (harness) => {
+  describe('Option: "verbose"', () => {
+    beforeEach(async () => {
+      // Application code is not needed for verbose output
+      await harness.writeFile('src/main.ts', '');
+    });
+
+    it('should not include verbose logs when true', async () => {
```
clydin, line 24:
```suggestion
    it('should include verbose logs when true', async () => {
```

clydin (approved):

PR #23321, opened 2022-06-08
fix(@angular/cli): correctly print package manager name when an install is needed

dgp1130:
I was testing if we supported a warning here today and noticed that it was printing "Try installing with '[object Object] install'". Easy fix at least.

### DIFF
--- a/tests/legacy-cli/e2e/tests/commands/builder-not-found.ts
+++ b/tests/legacy-cli/e2e/tests/commands/builder-not-found.ts
@@ -1,5 +1,5 @@
 import { moveFile } from '../../utils/fs';
-import { installPackage, uninstallPackage } from '../../utils/packages';
+import { getActivePackageManager, installPackage, uninstallPackage } from '../../utils/packages';
 import { execAndWaitForOutputToMatch, ng } from '../../utils/process';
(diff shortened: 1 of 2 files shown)


tests/legacy-cli/e2e/tests/commands/builder-not-found.ts
```diff
@@ -14,7 +14,11 @@ export default async function () {
       /Could not find the '@angular-devkit\/build-angular:browser' builder's node package\./,
     );
     await expectToFail(() =>
-      execAndWaitForOutputToMatch('ng', ['build'], /Node packages may not be installed\./),
+      execAndWaitForOutputToMatch(
+        'ng',
+        ['build'],
+        /Node packages may not be installed\. Try installing with 'npm install'\./,
```
clydin, line 20: The package manager command will change if the test is run in a project using a different package manager (e.g., yarn).
`getActivePackageManager` can be used to get the name of the package manager used in the test.

dgp1130: Yeah, I figured CI would fail with this but wasn't sure how to get the current package manager. Updated.

clydin (approved):

PR #23326, opened 2022-06-09
test: do not depend on a global npm binary

jbedard:
When sandboxed in bazel there is no accessible npm and it must be installed manually some way or another. The npm shipped with node is also not accessible (see https://github.com/bazelbuild/rules_nodejs/pull/3467).

This ~~yarn~~ npm is for the test-runner to install the packages being tested, not for the tests themselves which use a custom cli/npm/yarn within the test process.

Best attempt at using yarn: https://github.com/angular/angular-cli/commit/975f92a33b989b05cb917e099d045dbc8d0a47e4

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,6 +171,7 @@
     "minimatch": "5.1.0",
     "ng-packagr": "14.0.0",
     "node-fetch": "^2.2.0",
+    "npm": "^8.11.0",
     "npm-package-arg": "9.0.2",
     "open": "8.4.0",
     "ora": "5.4.1",
--- a/tests/legacy-cli/e2e/setup/010-local-publish.ts
+++ b/tests/legacy-cli/e2e/setup/010-local-publish.ts
@@ -1,11 +1,10 @@
 import { getGlobalVariable } from '../utils/env';
-import { execWithEnv, extractNpmEnv } from '../utils/process';
+import { globalNpm, extractNpmEnv } from '../utils/process';
 import { isPrereleaseCli } from '../utils/project';
 
 export default async function () {
   const testRegistry: string = getGlobalVariable('package-registry');
-  await execWithEnv(
-    'npm',
+  await globalNpm(
     [
       'run',
       'admin',
--- a/tests/legacy-cli/e2e/setup/100-global-cli.ts
+++ b/tests/legacy-cli/e2e/setup/100-global-cli.ts
@@ -1,5 +1,5 @@
 import { getGlobalVariable } from '../utils/env';
-import { silentNpm } from '../utils/process';
+import { globalNpm } from '../utils/process';
 
 const NPM_VERSION = '7.24.0';
 const YARN_VERSION = '1.22.18';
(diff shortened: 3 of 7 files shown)


alan-agius4 (approved):

package.json
```diff
@@ -171,6 +171,7 @@
     "minimatch": "5.1.0",
     "ng-packagr": "14.0.0",
     "node-fetch": "^2.2.0",
+    "npm": "^8.11.0",
```
clydin, line 174: Would it be possible to use the in-repo version of yarn instead?  This would avoid adding another dependency.  yarn is also the package manager used for other development related tasks and configured for use by Bazel.

jbedard: Yeah I think we can do that.

This is just what is used to install the packages being tested so I don't think npm/yarn or the version really matters. Those packages being tested are the cli + yarn + npm and those are the versions which matter (yes, we are discussing the npm/yarn that installs cli+npm+yarn 😨).

jbedard: @clydin that seems to cause issues with some of the tests which use npm, and now npm and yarn are sharing the same directory which contains the globals? So they have the same globals directory but different global DBs I think?

```
  Error: Running "npm install --global @angular/cli --registry=http://localhost:34837" returned error. 1...
  STDOUT:
  STDERR:
  npm ERR! code EEXIST
  npm ERR! path /mnt/ramdisk/e2e-main/npm-global/bin/ng
  npm ERR! EEXIST: file already exists
  npm ERR! File exists: /mnt/ramdisk/e2e-main/npm-global/bin/ng
  npm ERR! Remove the existing file and try again, or run npm
  npm ERR! with --force to overwrite files recklessly.
```

Fixing that will probably require more changes that I'd rather not do right now. Are we ok sticking with the my original commit (https://github.com/angular/angular-cli/commit/46a029c947bc8a10995401e823a145ed3f91a8c5)?

jbedard: Ok this actually had an easy and obvious fix... the global npm and yarn directories/databases just need to be separate. See [diff](https://github.com/angular/angular-cli/compare/9d506c9b723db962ca682146b888c20a6d3de00c..98a3eab99c403f4923cbb6e6745cdebc7a4f1ad9)

jbedard: Using yarn also requires using yarn for all installations of the packages being tested (cli, yarn, npm): [diff1](https://github.com/angular/angular-cli/compare/1b393e8003fe2f6bf377e9a48772c26ac04c0310..33824e29875ec0ea48890af3fb91408438c12928), [diff2](https://github.com/angular/angular-cli/compare/33824e29875ec0ea48890af3fb91408438c12928..701d249b7c8be15020c997ee85dda817a3e81d95)

jbedard: Now yarn seems to be having other issues. I've seen an issue publishing saying there's a version conflict (the tested cli package depends on something that doesn't exist?), that might be related to this other one:
```
  npm ERR! code E404
  npm ERR! 404 Not Found - GET http://localhost:62353/pend - no such package available
  npm ERR! 404 
  npm ERR! 404  'pend@~1.2.0' is not in this registry.
  npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
  npm ERR! 404 
  npm ERR! 404 Note that you can also install from a
  npm ERR! 404 tarball, folder, http url, or git url.
```
?

Are we ok going back to my original change keeping things as-is using npm? It just means installing the extra `npm` package...

jbedard: Here's the commit trying to switch to yarn: https://github.com/angular/angular-cli/pull/23326/commits/975f92a33b989b05cb917e099d045dbc8d0a47e4

I'm going to switch this PR back to npm and we can maybe try yarn again in the future.

alan-agius4:
@jbedard, can you please rebase?

jbedard:
Rebased

alan-agius4 (approved):

PR #23340, opened 2022-06-09
refactor(@angular/cli): assert catch clause variable type before usage

clydin:
Prepares the `@angular/cli` package for the eventual change of enabling the
TypeScript `useUnknownInCatchVariables` option. This option provides additional
code safety by ensuring that the catch clause variable is the proper type before
attempting to access its properties. Similar changes will be needed in the other
packages in the repository prior to enabling `useUnknownInCatchVariables`.


alan-agius4 (approved):
👍

PR #23345, opened 2022-06-10
test: reduce disk-cache E2E tests duration

alan-agius4:
We don't need to include all applications code to verify disk cache existence.

This reduces the test duration of `disk-cache.ts` by a whopping `~65s` since this tests runs `ng build` 5 times and `disk-cache-purge.ts` by `~13s`.

These timings were gathered using gLinux and is expected that the gain is more significant on Circle CI Linux and Windows executors


dgp1130 (approved):
Wow, 65 seconds is pretty significant. Great find!

PR #23347, opened 2022-06-10
docs: update install command

matzar:
Update the install command to be compliant with npm v8 and it does not throw the warning:
> npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Get started with Angular CLI, learn the fundamentals and explore advanced topics
 Install the Angular CLI globally:
 
 ```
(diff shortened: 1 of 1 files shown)


alan-agius4 (commented):
I think we should still keep the old version since NPM 6 is installed by default on Node 14.

alan-agius4:
@matzar thanks for your contribution. 


However as pointed above NPM 6 is still supported and the default in Node 14. Hence, let's re-consider this change in the future.

PR #23349, opened 2022-06-10
fix(@angular/cli): disable version check when running `ng completion` commands

dgp1130:
Running autocompletion with `14.0.1` installed globally in a `14.0.0` project logs the version warning in a very annoying fashion:

```
$ ng verYour global Angular CLI version (14.0.1) is greater than your local version (14.0.0). The local Angular CLI version is used.

sion
```

This adds an exception for `ng completion` commands to avoid this edge case. This doesn't include any tests, as I'm not aware of a great way to test version mismatching behavior. I did try it locally and found that it worked as expected. Happy to add a test if someone can point me to the right method / location.

### DIFF
--- a/packages/angular/cli/lib/init.ts
+++ b/packages/angular/cli/lib/init.ts
@@ -82,7 +82,7 @@ let forceExit = false;
       // If using the update command and the global version is greater, use the newer update command
       // This allows improvements in update to be used in older versions that do not have bootstrapping
       if (
-        process.argv[2] === 'update' &&
+        (process.argv[2] === 'update' || process.argv[2] === 'completion') &&
         cli.VERSION &&
         cli.VERSION.major - globalVersion.major <= 1
       ) {


alan-agius4 (changes requested):
When `NG_DISABLE_VERSION_CHECK` is true, we'll force the usage of the invoked CLI which in this case would be the global one and in some cases we might not want that, example to get the latest fixes from the project version.

https://github.com/angular/angular-cli/blob/789d78fa9143ee8d84a3778889a50726e2090d7e/packages/angular/cli/lib/init.ts#L36-L42

Instead we should change the below logic;
https://github.com/angular/angular-cli/blob/789d78fa9143ee8d84a3778889a50726e2090d7e/packages/angular/cli/lib/init.ts#L81-L100

Adding tests for this might be tricky especially now, moving our E2E to use Bazel where global installations are not permitted.

dgp1130:
Not sure I fully understand the reasoning here, but updated to use the branch in `lib.ts` when running `ng completion`, which I think is what you were going for here @alan-agius4.

packages/angular/cli/lib/init.ts
```diff
@@ -82,7 +82,7 @@ let forceExit = false;
       // If using the update command and the global version is greater, use the newer update command
       // This allows improvements in update to be used in older versions that do not have bootstrapping
       if (
-        process.argv[2] === 'update' &&
+        (process.argv[2] === 'update' || process.argv[2] === 'completion') &&
```
alan-agius4, line 85: While my suggested change was indeed to the logic in here. This change in logic isn't quite right.

This would cause the global CLI to be used if major version is <=1 from the local version. n general we should always use the local CLI, update is an edge case where we do this to avoid downloading a temporary Angular CLI to run the update. 

We should skip the entire `isGlobalGreater` when the command name is `completion`.

```diff
-    let isGlobalGreater = false;
     try {
-      isGlobalGreater = !!localVersion && globalVersion.compare(localVersion) > 0;
-    } catch (error) {
-      // eslint-disable-next-line  no-console
-      console.error('Version mismatch check skipped. Unable to compare local version: ' + error);
-    }
+      const isGlobalGreater = !!localVersion && globalVersion.compare(localVersion) > 0;
+      const rawCommandName = process.argv[2];
 
-    if (isGlobalGreater) {
-      // If using the update command and the global version is greater, use the newer update command
-      // This allows improvements in update to be used in older versions that do not have bootstrapping
-      if (
-        process.argv[2] === 'update' &&
-        cli.VERSION &&
-        cli.VERSION.major - globalVersion.major <= 1
-      ) {
-        cli = await import('./cli');
-      } else if (await isWarningEnabled('versionMismatch')) {
-        // Otherwise, use local version and warn if global is newer than local
-        const warning =
-          `Your global Angular CLI version (${globalVersion}) is greater than your local ` +
-          `version (${localVersion}). The local Angular CLI version is used.\n\n` +
-          'To disable this warning use "ng config -g cli.warnings.versionMismatch false".';
+      // When using the completion command and the global version is greater,
+      // don't show the warning as otherwise this will break completion.
+      if (isGlobalGreater && rawCommandName !== 'completion') {
+        // If using the update command and the global version is greater, use the newer update command
+        // This allows improvements in update to be used in older versions that do not have bootstrapping.
+        if (
+          rawCommandName === 'update' &&
+          cli.VERSION &&
+          cli.VERSION.major - globalVersion.major <= 1
+        ) {
+          cli = await import('./cli');
+        } else if (await isWarningEnabled('versionMismatch')) {
+          // Otherwise, use local version and warn if global is newer than local
+          const warning =
+            `Your global Angular CLI version (${globalVersion}) is greater than your local ` +
+            `version (${localVersion}). The local Angular CLI version is used.\n\n` +
+            'To disable this warning use "ng config -g cli.warnings.versionMismatch false".';
 
-        // eslint-disable-next-line  no-console
-        console.error(colors.yellow(warning));
+          // eslint-disable-next-line  no-console
+          console.error(colors.yellow(warning));
+        }
       }
+    } catch (error) {
+      // eslint-disable-next-line  no-console
+      console.error('Version mismatch check skipped. Unable to compare local version: ' + error);
     }
```

dgp1130: Thanks Alan, TBH I'm still a bit lost in the logic for this file and didn't get a chance to look into it too much today to really understand your solution. Would it be easier for you to just take it over given that you already have a full patch there and clearly understand it better than I do?

alan-agius4: Sure, I can take over.

alan-agius4:
Closing in favor of https://github.com/angular/angular-cli/pull/23375

PR #23375, opened 2022-06-15
fix(@angular/cli): disable version check when running `ng completion` commands

alan-agius4:
Running autocompletion with `14.0.1` installed globally in a `14.0.0` project logs the version warning in a very annoying fashion:

```
$ ng verYour global Angular CLI version (14.0.1) is greater than your local version (14.0.0). The local Angular CLI version is used.

sion
```

This adds an exception for `ng completion` commands to avoid this edge case.

### DIFF
--- a/packages/angular/cli/lib/init.ts
+++ b/packages/angular/cli/lib/init.ts
@@ -78,11 +78,13 @@ let forceExit = false;
       console.error('Version mismatch check skipped. Unable to compare local version: ' + error);
     }
 
-    if (isGlobalGreater) {
+    const rawCommandName = process.argv[2];
+    // When using the completion command, don't show the warning as otherwise this will break completion.
+    if (isGlobalGreater && rawCommandName !== 'completion') {
       // If using the update command and the global version is greater, use the newer update command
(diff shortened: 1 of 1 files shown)


packages/angular/cli/lib/init.ts
```diff
@@ -78,11 +78,14 @@ let forceExit = false;
       console.error('Version mismatch check skipped. Unable to compare local version: ' + error);
     }
 
-    if (isGlobalGreater) {
+    const rawCommandName = process.argv[2];
+    // When using the completion command and the global version is greater,
+    // don't show the warning as otherwise this will break completion.
+    if (isGlobalGreater && rawCommandName !== 'completion') {
```
dgp1130, line 84: This comment doesn't align with the if statement (should be `if (!(isGlobalGreater && rawCommandName === 'completion')` or `if (!isGlobalGreater || rawCommandName !== 'completion')`).

I think we don't want to change the existing behavior, so I think this comment should read "When using the completion command, don't show the warning as otherwise this will break completion." Is that what you're intending here?

dgp1130 (approved):

PR #23390, opened 2022-06-17
    feat(@angular-devkit/build-angular): add `externalDependencies` to the esbuild browser builder

### LINKED ISSUE #23322: support for providing esbuild options
# 🚀 Feature request

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

Regarding the new experimental `@angular-devkit/build-angular:browser-esbuild` builder, it would be great if you could expose a way to set esbuild options via configuration, specifically the `external` option is what I'm looking for at the moment to ensure externalized code (consumed/shared at runtime via ESM from a CDN) is not included in the Angular bundle (we want to use bare module specifiers and import maps).

For context, we are using `@angular/elements` combined with the new esbuild builder to share a custom element across applications, and consume shared ESM dependencies in both the application and custom element. We use typings from an npm package locally for the bare module specifier, but `importmap` at runtime to route it to the CDN.

I know this builder is experimental still, but I figured you would be interested in understanding how we're attempting to use it.

I would expect to provide these options via `angular.json`. Something along the lines of an `esbuildOptions` property that will get passed directly down to esbuild.

I've modified the `node_modules/@angular-devkit/build-angular/src/builders/browser-esbuild/esbuild.js` file locally for testing purposes to get by for now which works, but otherwise I'll need to fork this repo to produce my own build that supports this until it's implemented in the main package. If this isn't something on your radar any time soon please let me know and I could create a PR when it comes time for us needing to put more priority on this.

Also, if you know of an alternative workaround to pass the config for the short term please let me know!

alan-agius4:
This commit add a new `externalDependencies` option to the experimental browser builder.
    
Dependencies listed in this option will not be included in the final bundle, instead the user would need to provide them at runtime using import maps or another method.

Closes #23322

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
@@ -1,0 +1,542 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "title": "Esbuild browser schema for Build Facade.",
+  "description": "Browser target options",
+  "type": "object",
+  "properties": {
+    "assets": {
+      "type": "array",
+      "description": "List of static application assets.",
+      "default": [],
+      "items": {
+        "$ref": "#/definitions/assetPattern"
+      }
+    },
+    "main": {
+      "type": "string",
+      "description": "The full path for the main entry point to the app, relative to the current workspace."
+    },
+    "polyfills": {
+      "type": "string",
+      "description": "The full path for the polyfills file, relative to the current workspace."
+    },
+    "tsConfig": {
+      "type": "string",
+      "description": "The full path for the TypeScript configuration file, relative to the current workspace."
+    },
+    "scripts": {
+      "description": "Global scripts to be included in the build.",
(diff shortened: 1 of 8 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
```diff
@@ -0,0 +1,542 @@
+{
```
alan-agius4, line 1: Note: while technically we could use `allOf` ex:
```js
{
  "allOf": [
    {
      "properties": {
        "externalDependencies": {...}
    },
    { "$ref": "../browser/schema.json" }
  ]
}
```

quicktype doesn't merge this properly. See: https://github.com/quicktype/quicktype/issues/1840

clydin (approved):

PR #23393, opened 2022-06-17
build: update `ng-dev` setup to work with new ESM version

devversion:
The `ng-dev` tool is now strict ESM. Config loading needs to be updated
to also emit `.mjs` with `ts-node` (this cannot happen dynamically as
ESM Node loading cannot be patched by ts-node at runtime).


package.json
```diff
@@ -24,6 +24,7 @@
     "build:bazel": "node ./bin/devkit-admin build-bazel",
     "build-tsc": "tsc -p tsconfig.json",
     "lint": "eslint --cache --max-warnings=0 \"**/*.ts\"",
+    "ng-dev": "TS_NODE_PROJECT=$PWD/.ng-dev/tsconfig.json node_modules/@angular/dev-infra-private/ng-dev/bundles/cli.mjs",
```
devversion, line 27: Note: We cannot dynamically set the tsconfig as ts-node/esm needs to be registered before runtime. The explicit tsconfig is needed for `ts-node` to load the `.ng-dev` config as ESM

.ng-dev/tsconfig.json
```diff
@@ -1,8 +1,10 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "module": "Node16",
+    "moduleResolution": "Node16",
     "noEmit": true
   },
-  "include": ["**/*.ts"],
+  "include": ["**/*.mts"],
```
devversion, line 8: Additional note: `.mts` avoids relying on the scoping `package.json` file and this seems to be more reliable for the language service and it also makes it very explicit that this is supposed to run as ESM.

(I plan on doing the same for `ts_library` in the future -- Bazel will not be happy with scoping `package.json` files)

package.json
alan-agius4: You sure about that? I believe that if you change the shebang of `node_modules/@angular/dev-infra-private/ng-dev/bundles/cli.mjs` to `#!/usr/bin/env ts-node-esm` it should work.

devversion: @alan-agius4 It's already doing that, notice how we don't run `ts-node` here. The deciding part is that we would need to tell `ts-node` about the `tsconfig` before runtime as well (hence the environment variable)

alan-agius4: Oh I didn't notice that.

alan-agius4 (approved):

alan-agius4:
Merged to minor as it doesn’t apply cleanly

PR #23404, opened 2022-06-20
fix(@angular/cli): prevent schema validation error

t-bowersox:
Add `cliOptions.completion` to workspace-schema.json.

Resolve an error found using `ng-config` to disable version mismatch:

```
$ ng config -g cli.warnings.versionMismatch false
Your global Angular CLI version (14.0.2) is greater than your local version (14.0.1). The local Angular CLI version is used.

Error: Schema validation failed with the following errors:
  Data path "/cli" must NOT have additional properties(completion).
```

---

Please note that I was unable to run the test suite locally on my M1 Mac. After running `yarn bazel test //packages/...`, I received the following error that I was unable to resolve (I tried following the error message guidance to no avail):

```
ERROR: no such package '@nodejs_darwin_arm64//': No nodejs is available for darwin_arm64 at version 14.17.1
    Consider upgrading by setting node_version in a call to node_repositories in WORKSPACE.
    Note that Node 16.x is the minimum published for Apple Silicon (M1 Macs)
```

However, I did test a build locally following the [developer guide instructions](https://github.com/angular/angular-cli/blob/main/docs/DEVELOPER.md#building-and-installing-the-cli) and `ng config -g cli.warnings.versionMismatch false` completed without an error.

### DIFF
--- a/packages/angular/cli/lib/config/workspace-schema.json
+++ b/packages/angular/cli/lib/config/workspace-schema.json
@@ -67,7 +67,8 @@
               "description": "Show a warning when the global version is newer than the local one.",
               "type": "boolean"
             }
-          }
+          },
+          "additionalProperties": false
         },
         "analytics": {
           "type": ["boolean", "string"],
@@ -86,7 +87,8 @@
               "type": "string",
               "format": "uuid"
             }
-          }
+          },
+          "additionalProperties": false
         },
         "cache": {
           "description": "Control disk cache.",
@@ -105,7 +107,74 @@
               "description": "Cache base path.",
               "type": "string"
             }
+          },
+          "additionalProperties": false
+        }
+      },
+      "additionalProperties": false
+    },
+    "cliGlobalOptions": {
+      "type": "object",
+      "properties": {
+        "defaultCollection": {
+          "description": "The default schematics collection to use.",
+          "type": "string",
+          "x-deprecated": "Use 'schematicCollections' instead."
+        },
+        "schematicCollections": {
+          "type": "array",
+          "description": "The list of schematic collections to use.",
+          "items": {
+            "type": "string",
+            "uniqueItems": true
           }
+        },
+        "packageManager": {
+          "description": "Specify which package manager tool to use.",
+          "type": "string",
+          "enum": ["npm", "cnpm", "yarn", "pnpm"]
+        },
+        "warnings": {
+          "description": "Control CLI specific console warnings",
+          "type": "object",
+          "properties": {
+            "versionMismatch": {
+              "description": "Show a warning when the global version is newer than the local one.",
+              "type": "boolean"
+            }
+          },
+          "additionalProperties": false
+        },
+        "analytics": {
+          "type": ["boolean", "string"],
+          "description": "Share anonymous usage data with the Angular Team at Google."
+        },
+        "analyticsSharing": {
+          "type": "object",
+          "properties": {
+            "tracking": {
+              "description": "Analytics sharing info tracking ID.",
+              "type": "string",
+              "pattern": "^(GA|UA)?-\\d+-\\d+$"
+            },
+            "uuid": {
+              "description": "Analytics sharing info universally unique identifier.",
+              "type": "string",
+              "format": "uuid"
+            }
+          },
+          "additionalProperties": false
+        },
+        "completion": {
+          "type": "object",
+          "description": "Angular CLI completion settings.",
+          "properties": {
+            "prompted": {
+              "type": "boolean",
+              "description": "Whether the user has been prompted to add completion command prompt."
+            }
+          },
+          "additionalProperties": false
         }
       },
       "additionalProperties": false
@@ -552,14 +621,13 @@
       "type": "object",
       "properties": {
         "$schema": {
-          "type": "string",
-          "format": "uri"
+          "type": "string"
         },
         "version": {
           "$ref": "#/definitions/fileVersion"
         },
         "cli": {
-          "$ref": "#/definitions/cliOptions"
+          "$ref": "#/definitions/cliGlobalOptions"
         },
         "schematics": {
           "$ref": "#/definitions/schematicOptions"
--- a/packages/angular/cli/src/commands/config/cli.ts
+++ b/packages/angular/cli/src/commands/config/cli.ts
@@ -97,27 +97,6 @@ export class ConfigCommandModule
       throw new CommandModuleError('Invalid Path.');
     }
 
-    const validGlobalCliPaths = new Set<string>([
-      'cli.warnings.versionMismatch',
-      'cli.defaultCollection',
-      'cli.schematicCollections',
-      'cli.packageManager',
-
-      'cli.analytics',
-      'cli.analyticsSharing.tracking',
-      'cli.analyticsSharing.uuid',
-
-      'cli.completion.prompted',
-    ]);
-
-    if (
-      options.global &&
-      !options.jsonPath.startsWith('schematics.') &&
-      !validGlobalCliPaths.has(options.jsonPath)
-    ) {
-      throw new CommandModuleError('Invalid Path.');
-    }
-
     const [config, configPath] = await getWorkspaceRaw(options.global ? 'global' : 'local');
     const { logger } = this.context;
 
@@ -140,7 +119,7 @@ export class ConfigCommandModule
       return 1;
     }
 
-    await validateWorkspace(parseJson(config.content));
+    await validateWorkspace(parseJson(config.content), options.global ?? false);
 
     config.save();
 
--- a/packages/angular/cli/src/utilities/config.ts
+++ b/packages/angular/cli/src/utilities/config.ts
@@ -240,11 +240,20 @@ export async function getWorkspaceRaw(
   return [new JSONFile(configPath), configPath];
 }
 
-export async function validateWorkspace(data: json.JsonObject): Promise<void> {
-  const schema = readAndParseJson(workspaceSchemaPath) as json.schema.JsonSchema;
+export async function validateWorkspace(data: json.JsonObject, isGlobal: boolean): Promise<void> {
+  const schema = readAndParseJson(workspaceSchemaPath);
+
+  // We should eventually have a dedicated global config schema and use that to validate.
+  const schemaToValidate: json.schema.JsonSchema = isGlobal
+    ? {
+        '$ref': '#/definitions/global',
+        definitions: schema['definitions'],
+      }
+    : schema;
+
   const { formats } = await import('@angular-devkit/schematics');
   const registry = new json.schema.CoreSchemaRegistry(formats.standardFormats);
-  const validator = await registry.compile(schema).toPromise();
+  const validator = await registry.compile(schemaToValidate).toPromise();
 
   const { success, errors } = await validator(data).toPromise();
   if (!success) {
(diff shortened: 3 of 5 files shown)


packages/angular/cli/lib/config/workspace-schema.json
```diff
@@ -106,6 +106,15 @@
               "type": "string"
             }
           }
+        },
+        "completion": {
```
alan-agius4, line 110: `completion` is only valid in global context. This need some more thought.

t-bowersox: @alan-agius4 -- I gave this another look based on your feedback. Not sure if this is what you had in mind, but I added a check inside `ConfigCommandModule`'s `set()` method that will throw an "Invalid Path" error if a user tries to set `cli.completion.prompted` without the `--global` option.

packages/angular/cli/lib/config/workspace-schema.json
```diff
@@ -106,6 +106,15 @@
...
+          "description": "Enable shell command completions.",
```
alan-agius4, line 111: Nit: can you improve the descriptions?

- completion description: Angular CLI completion settings.
- prompted description: Whether the user has been prompted to add completion command prompt.

packages/angular/cli/lib/config/workspace-schema.json
alan-agius4: @t-bowersox, thanks for taking a look at this. 

I noticed that we have an unused `global` definition in the schema. 

https://github.com/angular/angular-cli/blob/9fafb2e1251f6790392a1f9c84854086c85d2191/packages/angular/cli/lib/config/workspace-schema.json#L551-L568

What we can do is to use this definition and validate against that. 

I'll summarize what needs to be done, but don't feel obligated to take this. I can take over if you don't wish to continue working on this.

- update the the `cli` property to include the `completion` option. This can be done using `anyOf`.
- Remove `format` from `$schema` property.
- Revert changes in https://github.com/angular/angular-cli/pull/23404/files#diff-2196c0c0d0c3badafa233d0e2e43d633e8059652f1021f4c1605f025cccfd6e8 and remove all logic related to `validGlobalCliPaths`.
- Update [validateWorkspace](https://github.com/angular/angular-cli/blob/9fafb2e1251f6790392a1f9c84854086c85d2191/packages/angular/cli/src/utilities/config.ts#L243-L253) to validate against the global definitions, something like.

```ts
export async function validateWorkspace(data: json.JsonObject, isGlobal: boolean): Promise<void> {
  const schema = readAndParseJson(workspaceSchemaPath);
  // We should eventually have a dedicate global config schema and use that to validate.
  const schemaToValidate = isGlobal
    ? {
        oneOf: [{ '$ref': '#/definitions/global' }],
        definitions: schema['definitions'],
      }
    : schema;
```
- Add an E2E test in [tests/legacy-cli/e2e/tests/commands/config](https://github.com/angular/angular-cli/blob/main/tests/legacy-cli/e2e/tests/commands/config)

alan-agius4 (changes requested):
Thanks for this. Kindly see comments.

t-bowersox:
@alan-agius4 -- I made most of the changes that you requested. The only exception was for:

> Revert changes in https://github.com/angular/angular-cli/pull/23404/files#diff-2196c0c0d0c3badafa233d0e2e43d633e8059652f1021f4c1605f025cccfd6e8 and remove all logic related to validGlobalCliPaths.

Although I did revert my previous changes, I had to leave the `validGlobalCliPaths` logic in place. Here's what happened:

The schema validation successfully prevents `cli.completion.prompted` from being get/set without the `--global` flag. (I added e2e tests for those scenarios.)

However, I found that I could also set any arbitrary property I wanted on the global config (like `cli.completion.foobar`, etc.).

At first, I tried using `"additionalProperties": false` on the `global.cli` schema, but that didn't work. It seemed to conflict with the `"additionalProperties": false` property of the `cliOptions` schema. This resulted in validation errors that prevented setting any global options.

I then tried the workaround suggested in the [JSON Schema docs](https://json-schema.org/understanding-json-schema/reference/object.html?highlight=additionalproperties#extending-closed-schemas). That essentially consisted of:

- A `cliBaseOptions` schema without the `"additionalProperties": false` property.
- A `cliOptions` schema that extended `cliBaseOptions`, but with the `"additionalProperties": false` property set.
- A `global.cli` schema that extended `cliBaseOptions`, added the `completion.prompted` property, and also had the `"additionalProperties": false` property set.

Although that resolved the errors and I was unable to get/set `cli.completion.prompted` without the `--global` flag, it _still_ didn't prevent setting arbitrary global properties.

I am at a loss as to why this was happening. Since I couldn't figure out a way to prevent this with only schema validation, I decided to leave that extra validation logic in place. I'm happy to take that out, though, if you have any suggestions.

alan-agius4:
> @alan-agius4 -- I made most of the changes that you requested. The only exception was for:
> 
> > Revert changes in https://github.com/angular/angular-cli/pull/23404/files#diff-2196c0c0d0c3badafa233d0e2e43d633e8059652f1021f4c1605f025cccfd6e8 and remove all logic related to validGlobalCliPaths.
> 
> Although I did revert my previous changes, I had to leave the `validGlobalCliPaths` logic in place. Here's what happened:
> 
> The schema validation successfully prevents `cli.completion.prompted` from being get/set without the `--global` flag. (I added e2e tests for those scenarios.)
> 
> However, I found that I could also set any arbitrary property I wanted on the global config (like `cli.completion.foobar`, etc.).
> 
> At first, I tried using `"additionalProperties": false` on the `global.cli` schema, but that didn't work. It seemed to conflict with the `"additionalProperties": false` property of the `cliOptions` schema. This resulted in validation errors that prevented setting any global options.
> 
> I then tried the workaround suggested in the [JSON Schema docs](https://json-schema.org/understanding-json-schema/reference/object.html?highlight=additionalproperties#extending-closed-schemas). That essentially consisted of:
> 
> * A `cliBaseOptions` schema without the `"additionalProperties": false` property.
> * A `cliOptions` schema that extended `cliBaseOptions`, but with the `"additionalProperties": false` property set.
> * A `global.cli` schema that extended `cliBaseOptions`, added the `completion.prompted` property, and also had the `"additionalProperties": false` property set.
> 
> Although that resolved the errors and I was unable to get/set `cli.completion.prompted` without the `--global` flag, it _still_ didn't prevent setting arbitrary global properties.
> 
> I am at a loss as to why this was happening. Since I couldn't figure out a way to prevent this with only schema validation, I decided to leave that extra validation logic in place. I'm happy to take that out, though, if you have any suggestions.

There is more info about this https://github.com/ajv-validator/ajv/issues/1496. We could use `unevaluatedProperties` but that would require some changes in `@angular-devkit/core`. I can potentially land those changes before hand. That being said, likely the most simple approach would be just to copy the common properties of `cliOptions` in `cliGlobalOptions`. These options rarely change. 

The `cliOptions` contain a property named `cache` which is not a valid global option, therefore copying seems to be the best way forward.

tests/legacy-cli/e2e/tests/commands/config/config-global-completion.ts
```diff
@@ -0,0 +1,18 @@
+import { homedir } from 'os';
+import * as path from 'path';
+import { deleteFile, expectFileToExist } from '../../../utils/fs';
+import { ng } from '../../../utils/process';
+import { expectToFail } from '../../../utils/utils';
+
+export default async function () {
+  await expectToFail(() => ng('config', 'cli.completion.prompted', 'true'));
```
alan-agius4, line 8: Can you please validate the failure, something like?

```suggestion
  {
    const { message } = await expectToFail(() =>
      silentNg('config', 'cli.completion.prompted', 'true'),
    );

    if (!message.includes('Data path "/cli" must NOT have additional properties(completion).')) {
      throw new Error('Should have failued with must NOT have additional properties(completion).');
    }
  }
```

tests/legacy-cli/e2e/tests/commands/config/config-global-completion.ts
```diff
@@ -0,0 +1,18 @@
...
+  await expectToFail(() => ng('config', 'cli.completion.prompted'));
+
+  await ng('config', '--global', 'cli.completion.prompted', 'true');
+  const { stdout } = await ng('config', '--global', 'cli.completion.prompted');
+  if (!stdout.includes('true')) {
+    throw new Error(`Expected "true", received "${JSON.stringify(stdout)}".`);
+  }
```
alan-agius4, line 15: Can you please add  a test that checks that no additional properties can be added

Example: 
```ts
const { message } = await expectToFail(() => silentNg('config', '--global', 'cli.completion.invalid', 'true'));

const { message } = await expectToFail(() => silentNg('config', '--global', 'cli.cache.enabled','true'));
```

packages/angular/cli/lib/config/workspace-schema.json
```diff
@@ -552,14 +567,16 @@
...
-          "type": "string",
-          "format": "uri"
+          "type": "string"
         },
         "version": {
           "$ref": "#/definitions/fileVersion"
         },
         "cli": {
-          "$ref": "#/definitions/cliOptions"
+          "anyOf": [
```
alan-agius4, line 576: Please see https://github.com/angular/angular-cli/pull/23404#issuecomment-1162843225

alan-agius4 (changes requested):
Thanks for this.

I know that working with schemas sometimes can be not that fun.

t-bowersox:
@alan-agius4 -- the latest changes are ready for your review.

It turns out that the issue wasn't just setting invalid top-level properties (like `cli.foobar`) -- it was also that invalid sub-properties could be set (like `cli.completion.foobar`). This didn't just affect the `cliGlobalOptions` schema either -- I discovered I could do that for `cliOptions`, too. For example, the CLI allowed me to run `ng config cli.warnings.zzzz true` without an error.

For the global configs, having a separate `cliGlobalOptions` config did help with the former issue, since it could now have a top-level `additionalProperties: false`.

To solve the latter issue, though, I added `additionalProperties: false` recursively for both `cliOptions` and `cliGlobalOptions`. For example, changing this:

```json
{
  "warnings": {
    "description": "Control CLI specific console warnings",
    "type": "object",
    "properties": {
      "versionMismatch": {
        "description": "Show a warning when the global version is newer than the local one.",
        "type": "boolean"
      }
    }
  }
}
```

to this:

```json
{
  "warnings": {
    "description": "Control CLI specific console warnings",
    "type": "object",
    "properties": {
      "versionMismatch": {
        "description": "Show a warning when the global version is newer than the local one.",
        "type": "boolean"
      }
    },
    "additionalProperties": false
  }
}
```

It turns out there wasn't actually a test that was checking if an invalid sub-property could be set. Although there was a test for:

```typescript
await expectToFail(() => ng('config', 'cli.warnings.zzzz'));
```

There wasn't one for trying to set `cli.warnings.zzzz`. I went ahead and added one to the config-set.ts test file to catch any regressions in the future.

My own personal testing worked out as expected for both valid and invalid properties, too, so I think this is shaping up well now. Happy to work on any additional changes you think are needed, though. Thanks for your help and feedback so far!

alan-agius4 (approved):
LGTM. 

Can you kindly squash the commits?

t-bowersox:
All set, @alan-agius4.

alan-agius4 (approved):
Thanks for your contibution.

PR #23416, opened 2022-06-22
test: improve resilience of E2E postinstall check to package hoisting

clydin:
False positive package paths are now checked relative to their containing `node_modules` directory.
This avoids issues checking the paths when the package manager changes its hoisting strategy for
one of the affected paths.


dgp1130 (approved):
Thanks for the quick fix!

PR #23422, opened 2022-06-23
build: avoid use of ts-node wrapper script as it breaks child process forking

devversion:
`ts-node` spawns a custom node launcher script and has a brotli-base64
encoded configuration that will point to the user-provided entry-point.

This breaks child process forking as we wouldn't be able to specify our
custom build worker and basically `ng-dev` would run inside `ng-dev`
again.


alan-agius4 (approved):

dgp1130:
`14.0.x`-specific patch in #23424.

PR #23424, opened 2022-06-23
build: avoid use of ts-node wrapper script as it breaks child process forking

dgp1130:
`ts-node` spawns a custom node launcher script and has a brotli-base64
encoded configuration that will point to the user-provided entry-point.

This breaks child process forking as we wouldn't be able to specify our
custom build worker and basically `ng-dev` would run inside `ng-dev`
again.

(cherry picked from commit 83b51ec0174fdeb5109b1a06397bf0c8185bd439)

This is a `14.0.x`-specific patch of #23422. Just ran `yarn` to resolve lockfile conflicts.

/cc @devversion


devversion (approved):
Thx Doug!

Issue #23414, opened 2022-06-22
ng new ignores --collection for checking schematic options


jaschwartzkopf:
# 🐞 Bug report

### Command (mark with an `x`)

- [x ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?
Yes, before you could use --collection to run a custom schematic for ng new.

### Description

The schematic in --collection is ran, but options are based on @schematics/angular:ng-new instead of --colleciton.

The issue is how collectionNameFromArgs is extracted from the options:
https://github.com/angular/angular-cli/blob/main/packages/angular/cli/src/commands/new/cli.ts#L48
```
const {
      options: { collectionNameFromArgs },
    } = this.context.args;
```

If this line is changed to:
```
const {
      options: { collection: collectionNameFromArgs },
    } = this.context.args;
```

The options from --collection are used correctly.

## 🔬 Minimal Reproduction
Try to use ng new --collection with a collection that has arguments not in @schematics/angular:ng-new

## 🔥 Exception or Error

Only options for @shematics/angular:ng-new are recognized.

```
Error: Unknown arguments: ngrx, mock-api, snap-oauth, mock-oauth
```

## 🌍 Your Environment

<pre><code>

```
$ ng version

     _                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/


Angular CLI: 14.0.1
Node: 16.13.1
Package Manager: yarn 1.22.17
OS: win32 x64

Angular:
...

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1400.1 (cli-only)
@angular-devkit/core         14.0.1 (cli-only)
@angular-devkit/schematics   14.0.1 (cli-only)
@schematics/angular          14.0.1 (cli-only)

</code></pre>
```

alan-agius4:
@jaschwartzkopf, your investigation looks good. Want to try to land a PR?

[...]

PR #23471, opened 2022-06-30
fix(@angular/cli): correctly handle --collection option in ng g #23414

Lipata:
related #23414
using PR #23440


alan-agius4:
Closing as `ng generate` doesn't have a collection option.

Lipata:
Hi, @alan-agius4. My mistake relating the problem I want to fix with --collection property. The problem I think the CLI is having is the following:

"ng g @custom-schematics:schematic-name --custom-schematic-property=value" is throwing 

`Error: Unknown arguments: custom-schematic-property`

Issue #21310, opened 2021-07-08
Angular should barf with the actual problem if project's root is missing.


EvanCarroll:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ X ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

No, not that I know

### Description

If you have more than two projects, and any project lacks `root` you will get the following error on `ng generate`

> [error] TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined

## 🔬 Minimal Reproduction

```json
{
  "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
  "version": 1,
  "newProjectRoot": "./projects",
  "defaultProject": "demo-app",
  "cli": {
    "packageManager": "npm"
  },
  "projects": {
    "demo-app": {},
    "bar": { root: "bar" }
  }
}
```

Note, either `demo-app` and `bar` must lack `root`, or **both** must lack `root`. If only one entry exists and it omits `root` then `ng g` will work as desired.

## 🔥 Exception or Error

An error not about and internal use of pathing.

## 🌍 Your Environment

```text
Angular CLI: 12.1.1
Node: 14.17.3
Package Manager: <error>
OS: linux x64

Angular: 11.2.14
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1102.14
@angular-devkit/build-angular   0.1102.14
@angular-devkit/core            11.2.14
@angular-devkit/schematics      12.1.1
@angular/cli                    12.1.1
@schematics/angular             12.1.1
ng-packagr                      11.2.4
rxjs                            6.5.5
typescript                      4.0.8
```

PR #23473, opened 2022-06-30
fix(@angular-devkit/core): provide actionable error when a workspace project has missing `root` property

alan-agius4:
The `root` property is required in a workspace project. Now we issue an actionable error message when this is missing.

Closes: #21310


alan-agius4:
@clydin PTAL

clydin (approved):

packages/angular_devkit/core/src/workspace/json/reader.ts
```diff
@@ -60,8 +60,10 @@ export async function readJsonWorkspace(
       // TODO: Diagnostic reporting support
       throw new Error(message);
     },
-    warn(_message, _node) {
+    warn(message, _node) {
       // TODO: Diagnostic reporting support
+      // eslint-disable-next-line no-console
+      console.warn(message);
```
clydin, line 66: nit: Does the node provide location info? line/column info may be useful for fixing the underlying problem.

PR #23478, opened 2022-06-30
build: add bazel config that disabled sharding and flaky re-runs

alan-agius4:
This is useful when running tests locally as otherwise 50 workers are spawned when running the `//packages/angular_devkit/build_angular:build_angular_browser_test` target even when having an `fit` on a single test. This also disabled re-runs on flaky tests.

//cc @devversion, not sure if there is a better way.


clydin (approved):

devversion (commented):
It's also worth noting that the sharing is enabled locally in COMP, FW and dev-infra repositories because we also want tests to be fast to be fast there.

Should this maybe just go into `--config=debug` as well? or `--config=no-sharding`?

PR #23485, opened 2022-07-01
Schematics class name validation

### LINKED ISSUE #12868: Prevent numbers in module/pipe names
```
- [ ] feature request
```

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [X] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Versions

Angular CLI: 6.2.1
Node: 8.11.2
OS: darwin x64
Angular: 6.1.7

@angular-devkit/architect       0.8.1
@angular-devkit/build-angular   0.8.1
@angular-devkit/build-webpack   0.8.1
@angular-devkit/core            0.8.1
@angular-devkit/schematics      0.8.1
@angular/cdk                    6.4.7
@angular/cli                    6.2.1
@angular/material               6.4.7
@schematics/angular             0.8.1
@schematics/update              0.8.1
rxjs                            6.3.2
typescript                      2.9.2
webpack                         4.19.0

### Repro steps
```
ng generate pipe 18decimals
```

Everything performs smoothly. The pipe is created, but referencing modules with numbers in name creates problems, like here, referencing names with numbers.

```
import { 18decimalsPipe } from './18decimals.pipe';

describe('18decimalsPipe', () => {
  it('create an instance', () => {
    const pipe = new 18decimalsPipe();
    expect(pipe).toBeTruthy();
  });
});
```

### Desired functionality
The cli should disallow names with numbers to prevent issues.

### LINKED ISSUE #13824: Improve name validation with component schematic (and others)
# 🚀 Feature request

```
- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [X] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] xi18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc
```

### Description

Using dots in component name when using component schematic (might also happen with other schematics) should fail. I.e. `ng generate component a.b.x` should fail.

I noticed that people who are new to Angular tend to use the schematic like this for the first time:
```sh
ng generate component hello-world.component
```
which produces something funky like this:
```ts
export class HelloWorld.ComponentComponent
```

Use some whitelist regex to check the component's name before.

By the way, why does the component's selector allow dots in the regular expression below?
https://github.com/angular/angular-cli/blob/eca5dc322f11360b82309c31ed6a716a894fc051/packages/schematics/angular/utility/validation.ts#L20

```sh
ng generate component hello-world.component
echo D'OH!
git reset --hard
git clean -d -f
ng generate component hello-world
```

Anyway, I am in for a pull request.

alan-agius4:
**fix(@angular-devkit/core): `classify` string util should concat string without using a `.`**

`.` is not a valid character in ES6 class names.

Prior to this change `foo.module` before used to be incorrectly classified to `Foo.Module` instead of `FooModule`.

Closes #13824

----

**fix(@schematics/angular): prevent numbers from class names**
    
With this change we prevent creating classes with invalid characters.
    
Closes #12868

---

Combined into a single PR as they are dependent on each other.

### DIFF
--- a/packages/angular_devkit/core/src/utils/strings.ts
+++ b/packages/angular_devkit/core/src/utils/strings.ts
@@ -74,13 +74,14 @@ export function camelize(str: string): string {
 /**
  Returns the UpperCamelCase form of a string.
 
+ @example
  ```javascript
  'innerHTML'.classify();          // 'InnerHTML'
  'action_name'.classify();        // 'ActionName'
  'css-class-name'.classify();     // 'CssClassName'
  'my favorite items'.classify();  // 'MyFavoriteItems'
+ 'app.component'.classify();      // 'AppComponent'
  ```
-
  @method classify
  @param {String} str the string to classify
  @return {String} the classified string
@@ -89,7 +90,7 @@ export function classify(str: string): string {
   return str
     .split('.')
     .map((part) => capitalize(camelize(part)))
-    .join('.');
+    .join('');
 }
 
 /**
--- a/packages/schematics/angular/utility/validation.ts
+++ b/packages/schematics/angular/utility/validation.ts
@@ -12,8 +12,17 @@ import { SchematicsException } from '@angular-devkit/schematics';
 // When adding a dash the segment after the dash must also start with a letter.
 export const htmlSelectorRe = /^[a-zA-Z][.0-9a-zA-Z]*(:?-[a-zA-Z][.0-9a-zA-Z]*)*$/;
 
+// See: https://github.com/tc39/proposal-regexp-unicode-property-escapes/blob/fe6d07fad74cd0192d154966baa1e95e7cda78a1/README.md#other-examples
+const ecmaIdentifierNameRegExp = /^(?:[$_\p{ID_Start}])(?:[$_\u200C\u200D\p{ID_Continue}])*$/u;
+
 export function validateHtmlSelector(selector: string): void {
   if (selector && !htmlSelectorRe.test(selector)) {
-    throw new SchematicsException(`Selector (${selector}) is invalid.`);
+    throw new SchematicsException(`Selector "${selector}" is invalid.`);
+  }
+}
+
+export function validateClassName(className: string): void {
+  if (!ecmaIdentifierNameRegExp.test(className)) {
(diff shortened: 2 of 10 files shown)


packages/schematics/angular/utility/validation.ts
```diff
@@ -11,9 +11,16 @@ import { SchematicsException } from '@angular-devkit/schematics';
 // Must start with a letter, and must contain only alphanumeric characters or dashes.
 // When adding a dash the segment after the dash must also start with a letter.
 export const htmlSelectorRe = /^[a-zA-Z][.0-9a-zA-Z]*(:?-[a-zA-Z][.0-9a-zA-Z]*)*$/;
+const classNameRegExp = /^[a-zA-Z_]\w*$/;
```
clydin, line 14: `$` is allowed as well as some unicode characters.
There's actually a nice example of testing for a valid ECMAScript identifier in the regexp unicode property proposal: https://github.com/tc39/proposal-regexp-unicode-property-escapes/blob/fe6d07fad74cd0192d154966baa1e95e7cda78a1/README.md#other-examples

packages/angular_devkit/core/src/utils/strings.ts
```diff
@@ -89,7 +90,7 @@ export function classify(str: string): string {
   return str
     .split('.')
     .map((part) => capitalize(camelize(part)))
-    .join('.');
+    .join('');
```
clydin, line 93: I'm wondering if this would be a breaking change since it is public and it could be expected to operate in the previous manner.

clydin (changes requested):

packages/angular_devkit/core/src/utils/strings.ts
alan-agius4: Umm… not sure show to classify this. Maybe we should discuss it during our next sync.

clydin (approved):

PR #23487, opened 2022-07-01
fix(@ngtools/webpack): do not run ngcc when `node_modules` does not exist

JoostK:
Prior to this change the CLI would fail with an error when ngcc could not
be run due to a missing `node_modules` directory, as is the case with Yarn
Plug'n'Play for example. The workaround was to create an empty `node_modules`
directory, but this is inconvenient and shouldn't make any difference.

This commit removes the error and simply skips ngcc processing.

### DIFF
--- a/packages/ngtools/webpack/src/ngcc_processor.ts
+++ b/packages/ngtools/webpack/src/ngcc_processor.ts
@@ -32,7 +32,7 @@ type ResolverWithOptions = ReturnType<Compiler['resolverFactory']['get']>;
 export class NgccProcessor {
   private _processedModules = new Set<string>();
   private _logger: NgccLogger;
-  private _nodeModulesDirectory: string;
+  private _nodeModulesDirectory: string | null;
 
   constructor(
     private readonly compilerNgcc: typeof import('@angular/compiler-cli/ngcc'),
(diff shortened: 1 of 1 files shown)


JoostK:
Suggestions on how/where to add a test are welcome!

JoostK:
I will also change this in ng-packagr once this has been reviewed/merged here.

packages/ngtools/webpack/src/ngcc_processor.ts
```diff
@@ -189,7 +193,10 @@ export class NgccProcessor {
 
     const timeLabel = `NgccProcessor.processModule.ngcc.process+${moduleName}`;
     time(timeLabel);
-    this.compilerNgcc.process({
+
+    // Use a void expression to suppress @typescript-eslint/no-floating-promises as this calls
+    // ngcc synchronously.
+    void this.compilerNgcc.process({
```
alan-agius4, line 199: This method doesn’t return a promise.

JoostK: ESLint gets confused because of the conditional return type (based on the `async: true` option). I'm not even sure why this is an issue now, does ESLint only report errors for touched files?

alan-agius4: No, it reports errors on all files, which is even more strange.

JoostK: Hmmm, I just ran `yarn lint` locally after reverting this change and there was no error then. So now it's even more weird why [the initial CI run](https://app.circleci.com/pipelines/github/angular/angular-cli/23830/workflows/2a1cc71b-a214-46b9-b6f2-bdeaec517b4b/jobs/314149) errored out on this. I can drop the commit if you'd prefer.

JoostK: Actually, I do get it now. There was a compile error in that commit where the options object included a `null` type for `basePath`, which is not accepted. So it didn't type check and the return type couldn't actually be narrowed to be non-Promise. So I'm dropping this change :-)

clydin (approved):

PR #23492, opened 2022-07-04
feat(@angular/create): add support for `yarn create` and `npm init`

### LINKED ISSUE #10339: Add more support to yarn create
Yarn supports add "create-*" in global _env_ and update automagically: https://yarnpkg.com/docs/cli/create/
Why don't create an alias? like ```yarn create angular-app ng-app```<br>
I use a command similar in npm with npx
```bash
npx @angular/cli new ng-app --minimal
```

### LINKED ISSUE #14292: Support npm init @angular shortcut
# 🚀 Feature request

### Description
npm supports a convention for quick setup of projects via `npm init @scopename` - see https://docs.npmjs.com/cli/init

Running the `npm init @angular` command should be a shortcut for the `ng new ...` process.

alan-agius4:
With this change we add support to scaffold  an Angular worksapce using `yarn create @angular` and `npm init @angular`. These shortcuts support all of the `ng-new` options.

Closes #10339 and closes #14292


alan-agius4:
Points to discuss:

1) should we have a snapshot build for the new package?
2) should it be `yarn create angular` or  `yarn create @angular`?

brandonroberts:
Seconded for `yarn create angular`

chihab:
Would the result be the same as doing `npx @angular/cli new --defaults`?

alan-agius4:
> Would the result be the same as doing `npx @angular/cli new ...`?

Yes

packages/angular/create/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
+# Copyright Google Inc. All Rights Reserved.
+#
+# Use of this source code is governed by an MIT-style license that can be
+# found in the LICENSE file at https://angular.io/license
+
+load("//tools:defaults.bzl", "pkg_npm", "ts_library")
+
+licenses(["notice"])  # MIT
+
+package(default_visibility = ["//visibility:public"])
```
dgp1130, line 10: Nit: Put `//visibility:public` on the targets that actually need to be public (for example `:license` probably doesn't need to be public).

packages/angular/create/README.md
```diff
@@ -0,0 +1,19 @@
...
+# Create an Angular CLI workspace
+
+Scaffold an Angular CLI workspace without needing to install the Angular CLI globally. All of the [ng new](https://angular.io/cli/new) options and features are supported.
+
+# Usage
+
+NPM
+
+```
+npm init @angular@latest [project-name] -- [...options]
```
dgp1130, line 12: Nit: Is `@latest` necessary?

dgp1130 (approved):
I was going to chat about this in one of the triage meetings, but since we've skipped a couple times I'll just respond here.

> 1. should we have a snapshot build for the new package?

I don't feel too strongly about this. Any particular reasons we do / don't want a snapshot build? I get that it probably wouldn't be very useful in an unpublished context, not sure if `npm init` or `yarn create` have a means of pulling an unpublished package for testing.

> 2. should it be `yarn create angular` or  `yarn create @angular`? We do own `create-angular`.

My feeling is that `@angular` is more "obviously secure". Most of our packages are under `@angular/*` so users should feel more confident that `npm init @angular` is a safe operation than `npm init angular` would. I don't think there's any guarantee that `angular` and `create-angular` are owned by the same group, so I would be much more hesitant to type `npm init angular` (fortunately we do own that and it is safe, but there's no way for a user to know that without looking).

We should definitely update `npm init angular` to at least fail with a clear message and suggest using `npm init @angular` since that is a very easy mistake to make.

packages/angular/create/README.md
alan-agius4: Yes, for NPM this is needed as otherwise if this command was previously used, the cached binary will be used.

packages/angular/create/BUILD.bazel
alan-agius4: Good call. We should update this through out.

alan-agius4:
> My feeling is that @angular is more "obviously secure". Most of our packages are under @angular/* so users should feel more confident that npm init @angular is a safe operation than npm init angular would. I don't think there's any guarantee that angular and create-angular are owned by the same group, so I would be much more hesitant to type npm init angular (fortunately we do own that and it is safe, but there's no way for a user to know that without looking).

We are on the same page here.

> We should definitely update npm init angular to at least fail with a clear message and suggest using npm init @angular since that is a very easy mistake to make.

I can do a quick script and just publish to NPM repo manually.

PR #23529, opened 2022-07-08
feat(@angular-devkit/build-angular): add Sass file support to experimental esbuild-based builder

clydin:
This change adds support for using Sass stylesheets within an application built with the experimental
esbuild-based browser application builder. Global stylesheets (`styles` build option) and component
stylesheets (`@Component({ styleUrls: [...], ...})`) with Sass can now be used.
The `stylePreprocessorOptions.includePaths` option is also available for Sass stylesheets.
Both the default format (`.scss`) and the indented format (`.sass`) are supported.
Inline component stylesheet support is not yet available with the esbuild-based builder.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/sass-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/sass-plugin.ts
@@ -1,0 +1,59 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import type { Plugin, PluginBuild } from 'esbuild';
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/sass-plugin.ts
```diff
@@ -0,0 +1,59 @@
...
+
+      build.onStart(() => {
+        sass = new SassWorkerImplementation();
+      });
+
+      build.onEnd(() => {
+        sass?.close();
+      });
+
+      build.onLoad({ filter: /\.s[ac]ss$/ }, async (args) => {
```
alan-agius4, line 27: Can you maybe add a comment that the current implementation would not watch files when there are errors during the first run.  Possible this could be tacked through some custom importer.

alan-agius4 (approved):
LGTM

PR #23530, opened 2022-07-08
test: update live-reload test to reduce flakes 

alan-agius4:
- assign dynamic port to proxy server
- remove RXJS `finalize` in favor of `finally`

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/dev-server/specs/live-reload_spec.ts
+++ b/packages/angular_devkit/build_angular/src/builders/dev-server/specs/live-reload_spec.ts
@@ -9,9 +9,11 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import { Architect, BuilderRun } from '@angular-devkit/architect';
 import { tags } from '@angular-devkit/core';
+import { createServer } from 'http';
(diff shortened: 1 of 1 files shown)


alan-agius4:
Tried several runs and didn't see any flakes.

packages/angular_devkit/build_angular/src/builders/dev-server/specs/live-reload_spec.ts
```diff
@@ -22,10 +24,20 @@ interface ProxyInstance {
...
+      const port = (server.address() as AddressInfo).port;
+      server.close((e) => (e ? reject(e) : resolve(port)));
+    });
+    server.once('error', (e) => server.close(() => reject(e)));
+    server.listen();
+  });
+}
 
+async function createProxy(target: string, secure: boolean, ws = true): Promise<ProxyInstance> {
+  const proxyPort = await findFreePort();
```
clydin, line 40: Can this be zero and then use `(server.address() as AddressInfo).port` on the proxy server below to get the port to use in the test?

clydin: That's unfortunate.

PR #23536, opened 2022-07-11
fix(@angular/cli): remove deprecation warning of `no` prefixed schema options

alan-agius4:
Prefixing options with `no` in schema definitions appears to be used more widely than initially thought.

The `noOption` can also be provided in the `angular.json` which makes this property definition less ambiguous, since boolean options don't get prefixed with `no` in the JSON config. Therefore, in order to reduce the community changes we remove the deprecation warning for such options and change the interim solution to permanent one.

None-the-less, it's still recommended that options are defined without the `no` prefix.


clydin (approved):

JohannesHoppe:
🙏

fmalcher:
The fix was released in Angular CLI 14.0.6 almost two weeks ago. I think it's time to cut a new version here! @JohannesHoppe

PR #23537, opened 2022-07-11
fix(@angular-devkit/build-angular): no script type module attribute o…

johnny-mh:
…n below es5 target build

With this browsers that don't support [type=module] will works properly.

----

The `script[type=module]` attribute makes browser freeze on script loading that does not support.

But es5 bundles are can work properly despite of unsupported that feature.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser/index.ts
@@ -300,6 +300,7 @@ export function buildWebpackBrowser(
                   const entrypoints = generateEntryPoints({
                     scripts: options.scripts ?? [],
                     styles: options.styles ?? [],
+                    scriptType: config.output?.scriptType,
                   });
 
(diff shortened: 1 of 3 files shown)


alan-agius4:
All of the browsers that Angular supports, support `module` scripts. 
Please see: https://angular.io/guide/browser-support

johnny-mh:
Most customer of our service using browser that doesn't support `module` scripts yet.

I looking for way to serving apps to them. I spent many time to solving it. but can't find another way. (I can modify bundle js directly but it is tricky and not safe 🤔)

alan-agius4:
Hi @johnny-mh, in that case it's best to use Angular 12 which is the last version of Angular that provided "Differential Loading" which provides the automatic building and conditional loading of ES5 and ES2015+ scripts for the application code. This would also not penalize users users using modern browser.

You can read more about this here: https://v12.angular.io/guide/deployment#differential-loading

PR #23541, opened 2022-07-12
test: use random e2e test ports

jbedard:


packages/schematics/angular/e2e/files/protractor.conf.js.template
```diff
@@ -17,7 +17,7 @@ exports.config = {
   },
   directConnect: true,
   SELENIUM_PROMISE_MANAGER: false,
-  baseUrl: 'http://localhost:4200/',
+  baseUrl: 'http://localhost:0/',
```
alan-agius4, line 20: I highly doubt this is valid.

Probably this works because the baseUrl is set in the protoactor builder. And thus this line can probably be dropped. 

Same as the others.

alan-agius4: I think you can leave it as it was since in this case port 0 in the baseUrl is technically invalid

jbedard: @alan-agius4 are you sure this is invalid? See https://github.com/angular/angular-cli/commit/6db7038c1bb07604348fbe804534561c0a6f7b7a

I'm looking at this again because I'm getting port-in-use errors again when running e2e tests concurrently...

jbedard: (although I should only be changing it in the e2e tests, not in this schematics file)

alan-agius4 (changes requested):
Changes other than that in https://github.com/angular/angular-cli/blob/c74e33eef1bbb4fc73e204520f57bba3574891dc/packages/angular_devkit/build_angular/src/builders/dev-server/specs/works_spec.ts should be reverted. For E2E the port should be set in

https://github.com/angular/angular-cli/blob/5677bee4e7b5bb0bee40209870c09fde7118c02c/tests/legacy-cli/e2e/utils/project.ts#L170

jbedard:
I don't think that catches all cases but looks like it fixes the ones actually using e2e so is probably enough 👍

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -165,9 +165,11 @@ export function useCIDefaults(projectName = 'test-project') {
     const appTargets = project.targets || project.architect;
     appTargets.build.options.progress = false;
     appTargets.test.options.progress = false;
-    // Disable auto-updating webdriver in e2e.
     if (appTargets.e2e) {
+      // Disable auto-updating webdriver in e2e.
       appTargets.e2e.options.webdriverUpdate = false;
+      // Use a random port in e2e.
+      appTargets.e2e.options.port = 0;
```
jbedard, line 172: @alan-agius4 do you think this would also be the ideal place to set `appTargets.serve.options.port`? I just had a test running `ng serve` fail in another PR due to 4200 being in use...

jbedard: Note there's various `ng serve --port=0` throughout the tests already but a few are missing such as
```
rebuild-deps-type-check.ts
rebuild-types.ts
```

jbedard: I've added another commit but feel free to only merge the first if you'd like.

alan-agius4: Yes we should.

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -165,9 +165,16 @@ export function useCIDefaults(projectName = 'test-project') {
...
+    }
+
+    if (appTargets.serve) {
+      // Use a random port in serve.
+      appTargets.serve.options.port = 0;
```
alan-agius4, line 177:
```suggestion
      appTargets.serve.options ?? = {};
      appTargets.serve.options.port = 0;
```

jbedard: You can tell I did that in a hurry 😬

alan-agius4 (approved):
👍

PR #23542, opened 2022-07-12
ci: enable full E2E suit test on Windows for PRs

alan-agius4:
This commits changes the CI workflow to run all the complete E2E test suit on Windows for PRs.

The also increase the parallelism for Windows tests to 16 to reduce the time taken for this step and removes the requirement for the `build` step before running `e2e-cli-win`.

Based on the current timings, this change will not increase PR CI times.


josephperrott (approved):
LGTM

PR #23545, opened 2022-07-12
feat(@angular/cli): Identify third-party sources in sourcemaps

victorporof:
This PR includes a webpack plugin which adds a field to source maps that identifies which sources are vendored or runtime-injected (aka third-party) sources. These will be consumed by Chrome DevTools to automatically ignore-list sources.

When vendor source map processing is enabled, this is interpreted as the developer intending to debug third-party code; in this case, the feature is disabled.

Signed-off-by: Victor Porof <victorporof@chromium.org>


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -434,12 +438,26 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
...
         },
       },
     },
-    plugins: [new NamedChunksPlugin(), new DedupeModuleResolvePlugin({ verbose }), ...extraPlugins],
+    plugins: [
+      new NamedChunksPlugin(),
+      new DedupeModuleResolvePlugin({ verbose }),
+      new DevToolsIgnorePlugin({
+        vendors: !vendorSourceMap &&
+          !!vendorChunk && {
```
alan-agius4, line 451: In case `vendorChunking` is not enabled, I think we should still enable this.  When vendorChunk is disabled vendors are still included, just in another bundle.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+  vendors?: false | { test: RegExp };
+  runtime?: false | { test: RegExp };
+}
+/**
+ * This plugin adds a field to source maps that identifies which sources are
+ * vendored or runtime-injected (aka third-party) sources. These are consumed by
+ * Chrome DevTools to automatically ignore-list sources.
+ */
+export class DevToolsIgnorePlugin {
+  options: Options;
```
alan-agius4, line 27: Redundant. 

```suggestion
```

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  constructor(options: Options) {
+    this.options = options;
```
alan-agius4, line 30:
```suggestion
```

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+            const map = asset.map() as Object & {
+              sources: string[];
+              [IGNORE_LIST]: number[];
+            };
+            if (!map) {
+              continue;
+            }
+
+            map[IGNORE_LIST] = Object.entries(map.sources)
+              .filter(([, source]) => source.match(vendorsTest) || source.match(runtimeTest))
```
alan-agius4, line 61:
```suggestion
              .filter(([, source]) => vendorsTest.test(source) || runtimeTest.test(source))
```

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
victorporof: We can do `constructor(private options: Options) {}` otherwise tsc will complain. What do you think?

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
victorporof: Ok!

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tap(
+        {
+          name: PLUGIN_NAME,
+          stage: Compilation.PROCESS_ASSETS_STAGE_DEV_TOOLING,
+          additionalAssets: true,
+        },
+        (assets) => {
+          if (!this.options.vendors || !this.options.runtime) {
```
alan-agius4, line 44: Should this be `&&` instead of `||`? IE: if either of them is truthy we should not exit.

victorporof: You're right, I was just looking at this as well. This can be either or, will make the change.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -434,12 +438,26 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
...
+            test: VENDORS_TEST,
           },
         },
       },
     },
-    plugins: [new NamedChunksPlugin(), new DedupeModuleResolvePlugin({ verbose }), ...extraPlugins],
+    plugins: [
+      new NamedChunksPlugin(),
+      new DedupeModuleResolvePlugin({ verbose }),
+      new DevToolsIgnorePlugin({
```
alan-agius4, line 449: I feel it would be better to just not add the plugin when `vendorSourceMap` is truthy, Instead of the options.

Ex: https://github.com/angular/angular-cli/blob/56320245a126481816a8ccfb0cc248a00be49561/packages/angular_devkit/build_angular/src/webpack/configs/common.ts#L222-L228

victorporof: Makes sense! Might gain us a bit of a speed bump if the tapping isn't zero cost.

alan-agius4:
Do we know what is the build time impact of adding this plugin in a barebones angular project?

victorporof:
> Do you know what is the build time impact of adding this plugin in a barebones angular project?

I don't. How can I benchmark this? I've also noticed there are some benchmark tests in the repo, would those catch any regressions?

packages/angular_devkit/build_angular/src/builders/browser/specs/vendor-source-map_spec.ts
```diff
@@ -10,6 +10,10 @@ import { Architect } from '@angular-devkit/architect';
 import * as path from 'path';
```
alan-agius4, line 10: NIT: commit message scope should be `@angular-devkit/build-angular`

packages/angular_devkit/build_angular/src/builders/browser/specs/vendor-source-map_spec.ts
```diff
@@ -50,3 +54,90 @@ describe('Browser Builder external source map', () => {
     expect(hasTsSourcePaths).toBe(false, `vendor.js.map not should have '.ts' extentions`);
   });
 });
+
+describe('Identifying third-party code in source maps', () => {
```
alan-agius4, line 58: It's prefer if new tests were created using the new test harness. https://github.com/angular/angular-cli/tree/56320245a126481816a8ccfb0cc248a00be49561/packages/angular_devkit/build_angular/src/builders/browser/tests/behavior

But I guess we can migrate this as a whole eventually.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
alan-agius4: SGTM.

alan-agius4:
> I don't. How can I benchmark this? I've also noticed there are some benchmark tests in the repo, would those catch any regressions?

We actually don't have any benchmark tests. There is a bench-marker but you'd need to create and run the tests yourself. 

Let me benchmark this myself and get back to you.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
victorporof: Done.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  apply(compiler: Compiler) {
+    const { SourceMapSource } = compiler.webpack.sources;
+
+    compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
+      compilation.hooks.processAssets.tap(
+        {
+          name: PLUGIN_NAME,
+          stage: Compilation.PROCESS_ASSETS_STAGE_DEV_TOOLING,
+          additionalAssets: true,
```
alan-agius4, line 37: Out of curiosity, do we need the `additionalAssets` option?

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -101,6 +105,19 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
   // determine hashing format
   const hashFormat = getOutputHashFormat(buildOptions.outputHashing);
 
+  if (!vendorSourceMap) {
```
alan-agius4, line 108: Can you actually put this under https://github.com/angular/angular-cli/blob/c4401882d6e46ed79a2190501500dd06000d2182/packages/angular_devkit/build_angular/src/webpack/configs/common.ts#L200 so that this plugin is only enabled when sourcemaps are enabled?

victorporof: Done.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
victorporof: Yes, but I won't be able to tell you exactly why. In my testing, the source map assets are received only when this is true.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,74 @@
...
+            const map = asset.map() as Object & SourceMap;
+            if (!map) {
+              continue;
+            }
+
+            map[IGNORE_LIST] = Object.entries(map.sources)
+              .filter(([, source]) => vendorsTest?.test(source) || runtimeTest?.test(source))
+              .map(([index]) => +index);
+
+            compilation.assets[name] = new SourceMapSource(asset.source(), name, map);
```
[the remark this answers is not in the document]
victorporof: That's reasonable! I've taken inspiration from the terser plugin, and I assume many other plugins go the subscript update route as well, so if webpack decides to change this behavior there's going to be some pain 😅 . Will update.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
victorporof: I'm referring to the assets for which `asset.map()` actually returns something.

victorporof: Oh wait, no, this was backwards. Without additional assets, the following are processed:
```
main.js
polyfills.js
styles.css
runtime.js
vendor.js
```

With additional assets, the following extra are processed:
```
main.js.map
polyfills.js.map
styles.css.map
runtime.js.map
vendor.js.map
```

So the plugin should also work _without_ `additionalAssets: true`.

victorporof: Removed this, works as expected.

alan-agius4:
From my benchmark it appears that the plugin itself took about 500ms on a barebones project and ~2.8s on a [real world application](https://github.com/johannesjo/super-productivity)

Overall however, the build times didn't increase that much due to the async nature of the plugin.

victorporof:
Squashed.

victorporof:
> From my benchmark it appears that the plugin itself took about 500ms on a barebones project and ~2.8s on a [real world application](https://github.com/johannesjo/super-productivity)
> 
> Overall however, the build times didn't increase that much due to the async nature of the plugin.

We've found a way to reduce this from 2.8s to 200ms in the real world project by modifying the raw source maps directly instead of getting to them through the generated assets. Alan will push a commit.

alan-agius4 (changes requested):
@victorporof, overall I am happy with this following the perf improvements that we did. Which in a real world app the processing took ~200ms were previously it took 2.9s

@clydin / @dgp1130 might want to take a look as well.

alan-agius4 (approved):
Oops wrong radiobox.

alan-agius4 (approved):
LGTM

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+            const mapContent = asset.source().toString();
+            if (!mapContent) {
+              continue;
+            }
+
+            const map = JSON.parse(mapContent) as SourceMap;
+
+            map[IGNORE_LIST] = Object.entries(map.sources)
+              .filter(([, source]) => vendorsTest?.test(source) || runtimeTest?.test(source))
+              .map(([index]) => +index);
```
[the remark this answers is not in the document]
clydin: To avoid the conversion, this could be changed to something similar to:
```
map[IGNORE_LIST] = map.sources
        .map((source, index) => [source, index] as const)
        .filter(([source]) => vendorsTest?.test(source) || runtimeTest?.test(source))
        .map(([, index]) => index);
```
Other alternatives could be using `reduce` or a `for` loop.

packages/angular_devkit/build_angular/src/builders/browser/specs/vendor-source-map_spec.ts
```diff
@@ -10,6 +10,10 @@ import { Architect } from '@angular-devkit/architect';
 import * as path from 'path';
 import { browserBuild, createArchitect, host } from '../../../testing/test-utils';
 
+// Following the naming conventions from
+// https://sourcemaps.info/spec.html#h.ghqpj1ytqjbm
+const IGNORE_LIST = 'x_google_devtoolsIgnore';
```
bmeurer, line 15: Somehow I think `devtoolsIgnore` reads a bit weird. Maybe `x_google_ignoreList` instead?

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+
+// Following the naming conventions from
+// https://sourcemaps.info/spec.html#h.ghqpj1ytqjbm
+const IGNORE_LIST = 'x_google_devtoolsIgnore';
+
+const PLUGIN_NAME = 'devtools-ignore-plugin';
+
+interface Options {
+  vendors?: false | { test: RegExp };
+  runtime?: false | { test: RegExp };
```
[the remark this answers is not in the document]
victorporof: We could, but it won't follow the coding style used across other parts of the CLI codebase where `!!foo && value` is used to specify `value` depending on `foo`.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
```diff
@@ -0,0 +1,83 @@
...
+        },
+        (assets) => {
+          let vendorsTest: RegExp | null = null;
+          let runtimeTest: RegExp | null = null;
+          if (this.options.vendors) {
+            vendorsTest = this.options.vendors.test;
+          }
+          if (this.options.runtime) {
+            runtimeTest = this.options.runtime.test;
+          }
```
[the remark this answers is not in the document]
victorporof: Yup, happy to inline all of this.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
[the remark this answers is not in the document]
victorporof: @dgp1130 Object.entries gives string keys for arrays. Hence the `+` to convert to number. `Number(index)` is also fine.

packages/angular_devkit/build_angular/src/builders/browser/specs/vendor-source-map_spec.ts
victorporof: Happy with `x_google_ignoreList` as well!

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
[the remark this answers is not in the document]
victorporof: As discussed, I'll remove the `test` options altogether, and just keep the `runtime` as a boolean flag.

packages/angular_devkit/build_angular/src/webpack/plugins/devtools-ignore-plugin.ts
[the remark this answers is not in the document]
victorporof: Switched to a for loop.

packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -190,6 +194,19 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
       include.push(/css$/);
     }
 
+    if (!vendorSourceMap) {
```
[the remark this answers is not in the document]
victorporof: Fwiw, I don't feel strongly about this at all and happy with either decision.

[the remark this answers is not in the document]
alan-agius4: I think what @dgp1130 is suggesting makes sense. 

By the way do devtools currently support the ignore list? I am not seeing any differenced in the devtools with our without this change.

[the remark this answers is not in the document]
victorporof: @alan-agius4 We're in the process of building this feature. We expect to land the parsing of `x_google_ignoreList` and the automatic ignore-listing feature soon.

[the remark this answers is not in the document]
dgp1130: This probably depends on a bit on what the DevTools UI actually looks like and how easy it is to discover and flip "just my code" as needed. Let's remove this condition for now and always include this data. If we find users are getting confused by it, then I think we can either treat that as a UI bug in DevTools or revisit this decision so `vendorSourceMaps` can be used as a CLI-level opt-out.

victorporof:
Now generating the ignore-list even when vendored source maps processing is enabled.

dgp1130 (approved):
Thanks! Looking forward to trying this out once it's implemented in DevTools!

PR #23550, opened 2022-07-13
test: remove unnecessary `extract-i18n` command from i18n setup

alan-agius4:
This helps to cut down on the time taken i18n tests take to run.


dgp1130:
Had a merge conflict with patch/rc. Can you make separate PRs for those if you want to get it in there? We can just put this in RC and let patch fall out of usage if you don't want to make two PRs.

PR #23564, opened 2022-07-14
fix(@angular-devkit/core): log name of invalid extension too

mandarini:
When logging the warning `Project extension with invalid name found.`, include the name of the project extension, to facilitate debugging.


alan-agius4 (approved):
Thanks

jeffwhelpley:
Hey, question on this @alan-agius4 @mandarini. I am getting 2 `Project extension with invalid name found` warnings when the names of my projects in angular.json are `contact` and `ghphone`. I know the warning is from these despite not having it log out the name because when I remove one of them, I get one warning instead of two.

So, why are these names invalid? Is this a bug or are project names supposed to follow some format?

alan-agius4:
@jeffwhelpley, that should be valid. Can you please open an issue with a repro? Thanks.

PR #23566, opened 2022-07-14
fix(@angular-devkit/core): add i18n as valid project extension

voria:
When parsing the angular.json file, the 'i18n' key used to configure localization in a localized project should be treated as a special project extension key.

This fixes the warning "Project extension with invalid name found." printed when starting/building the project.


alan-agius4 (changes requested):
Can you please update the commit subject to `fix(@angular-devkit/core): add i18n as valid project extension`?

voria:
Commit subject updated.

alan-agius4:
Looks like Circle CI was not picking up this PR. Had to close/open.

alan-agius4 (approved):
LGTM. thanks

PR #23594, opened 2022-07-18
refactor(@angular-devkit/architect-cli): remove rxjs as a dependency

clydin:
`rxjs` was only a dependency for the `tap` operator which can be replaced
by accessing the result object returned from awaiting the builder output
`Observable` which is already converted to a Promise.


alan-agius4 (commented):
You can also remove
 #https://github.com/angular/angular-cli/blob/2b4d54a813482b7e4f52848dd34afb85410dd30b/packages/angular_devkit/architect_cli/BUILD.bazel#L27

alan-agius4 (approved):

PR #23600, opened 2022-07-19
ci: create e2e matrices

alan-agius4:
With this commit, we create different matrices for our e2e tests. This reduces the duration of the workflow by ~7mins

More information: https://circleci.com/docs/using-matrix-jobs

### DIFF
--- a/.circleci/dynamic_config.yml
+++ b/.circleci/dynamic_config.yml
@@ -37,6 +37,8 @@ var_5: &only_release_branches
         - main
         - /\d+\.\d+\.x/
 
+var_6: &all_e2e_subsets ['npm', 'esbuild', 'yarn']
+
 # Executor Definitions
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

Though I am curious how this reduces the overall test time if we are still just running all of the items that we have in subsets now.

alan-agius4:
At the moment we run all the subsets in the same job. Sequential to each other.

clydin (approved):

clydin:
The Github requires status checks will need to be updated.  Currently `ci/circleci: e2e-cli` is required but no longer exists with this PR.  i think `e2e-cli-npm` and `e2e-cli-yarn` are good candidates to replace it.

alan-agius4:
Probably `e2e-cli-npm` or both to cover the entire suit.

alan-agius4:
@dgp1130 can you take care of this please?

PR #23624, opened 2022-07-22
feat(@angular-devkit/build-angular): switch to use Sass modern API

alan-agius4:
Sass modern API provides faster compilations times when used in an async manner.

|Application compilation duration | Sass API and Compiler|
|-- | --|
|60852ms | dart-sass legacy sync API|
|52666ms | dart-sass modern API|

Note: https://github.com/johannesjo/super-productivity was used for benchmarking.

Prior art: http://docs/document/d/1CvEceWMpBoEBd8SfvksGMdVHxaZMH93b0EGS3XbR3_Q?resourcekey=0-vFm-xMspT65FZLIyX7xWFQ

BREAKING CHANGE:

- Deprecated support for tilde import has been removed. Please update the imports by removing the `~`.

Before
```scss
@import "~font-awesome/scss/font-awesome";
```

After
```scss
@import "font-awesome/scss/font-awesome";
```

- By default the CLI will use Sass modern API, While not recommended, users can still opt to use legacy API by setting `NG_BUILD_LEGACY_SASS=1`.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -9,11 +9,12 @@
 import * as fs from 'fs';
 import MiniCssExtractPlugin from 'mini-css-extract-plugin';
 import * as path from 'path';
-import type { Configuration, RuleSetUseItem } from 'webpack';
+import { Configuration, RuleSetUseItem } from 'webpack';
 import { StyleElement } from '../../builders/browser/schema';
 import { SassWorkerImplementation } from '../../sass/sass-service';
+import { SassLegacyWorkerImplementation } from '../../sass/sass-service-legacy';
 import { WebpackConfigOptions } from '../../utils/build-options';
-import { addWarning } from '../../utils/webpack-diagnostics';
+import { useLegacySass } from '../../utils/environment-options';
 import {
   AnyComponentStyleBudgetChecker,
   PostcssCliResources,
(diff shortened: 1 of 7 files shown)


alan-agius4:
Opt in or out?

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -419,3 +373,69 @@ function getTailwindConfigPath({ projectRoot, root }: WebpackConfigOptions): str
...
+        // See: https://github.com/webpack-contrib/sass-loader/blob/997f3eb41d86dd00d5fa49c395a1aeb41573108c/src/utils.js#L642-L651
+        webpackImporter: false,
+        sassOptions: {
+          loadPaths: [
+            ...includePaths,
+            // Needed to resolve node packages and retain the same behaviour of with the legacy API as sass-loader resolves
+            // scss also from the cwd and project root.
+            // See: https://github.com/webpack-contrib/sass-loader/blob/997f3eb41d86dd00d5fa49c395a1aeb41573108c/src/utils.js#L307
+            projectRoot,
+            path.join(root, 'node_modules'),
```
clydin, line 400: This workaround won't support package `exports` fields nor Yarn PnP.  We'll need to do a followup with a fix for this.  Probably a custom Sass importer that tries node resolution first from the project root.

clydin (approved):
Should we add a note to the breaking change description about the environment variable?

packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
alan-agius4: Yes, I already have some ideas for that.

PR #23626, opened 2022-07-22
build: remove dependency on font-awesome

alan-agius4:
This dependency is only used to valid that node packages are resolved correctly.

With this change we mock the structure of font-awesome to avoid having to install it.


clydin (approved):

dgp1130:
This had a merge conflict in patch, so I retargeted it to minor and made https://github.com/angular/angular-cli/pull/23681/ with a patch-specific PR.

PR #23643, opened 2022-07-27
build: update dev-infra packages and account for build-tooling split from `ng-dev`

devversion:
See individual commits


alan-agius4 (approved):
LGTM, 🎉

yarn.lock
```diff
@@ -122,6 +122,42 @@
     "@angular/core" "^13.0.0 || ^14.0.0-0"
     reflect-metadata "^0.1.13"
 
+"@angular/build-tooling@https://github.com/angular/dev-infra-private-build-tooling-builds.git#1addc303bef0b6acc0dca0961e9e642629f3a5cd":
+  version "0.0.0-fa61d03a603e04af2b66f3598f1af01da1e1cfb1"
+  resolved "https://github.com/angular/dev-infra-private-build-tooling-builds.git#1addc303bef0b6acc0dca0961e9e642629f3a5cd"
+  dependencies:
+    "@angular-devkit/build-angular" "14.1.0-rc.3"
```
alan-agius4, line 129: Why does `dev-infra-private-build-tooling-builds` have a dependency on build-angular?

devversion: Because build-tooling exposes a rule for bundling Angular apps closely to how CLI works. i.e. ESBuild -> Babel plugins from build-angular -> Terser (just not webpack)

devversion: We can't make any of these dependencies optional or peer deps (we had that in the past) because of how Bazel and NPM dependencies work.

alan-agius4: That is unfortunate as build-angular has a lot of dependencies.

devversion: 1. Yeah, maybe in the future we need to revise this. Also this may change when we ship stuff with the new `rules_js`

devversion: This isn't something new in this PR, but a good point you bring up. Definitely was annoying and non-ideal to me as well

alan-agius4: I am aware of that, just wanted to point it out 🙂

alan-agius4:
Could not merge pull request into the following branches due to merge conflicts: 14.1.x.

yarn.lock
devversion: yeah, we should revisit this soon

PR #23649, opened 2022-07-27
test: move tempdir creation to util method

jbedard:
To put the login in a common location. This will allow easily using the bazel temp dir when converting to bazel.

### DIFF
--- a/tests/legacy-cli/e2e/utils/utils.ts
+++ b/tests/legacy-cli/e2e/utils/utils.ts
@@ -1,3 +1,7 @@
+import { mkdtemp, realpath } from 'fs/promises';
+import { tmpdir } from 'os';
+import path from 'path';
+
 export function expectToFail(fn: () => Promise<any>, errorMessage?: string): Promise<any> {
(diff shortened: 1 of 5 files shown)


tests/legacy-cli/e2e/utils/utils.ts
```diff
@@ -18,3 +22,7 @@ export function wait(msecs: number): Promise<void> {
     setTimeout(resolve, msecs);
   });
 }
+
+export async function mktempd(prefix: string) {
+  return await realpath(await mkdtemp(path.join(tmpdir(), prefix)));
+}
```
alan-agius4, line 28: NIT

```suggestion
export async function mktempd(prefix: string): Promise<string> {
  return realpath(await mkdtemp(path.join(tmpdir(), prefix)));
}
```

alan-agius4 (approved):
LGTM, just one NIT

alan-agius4:
@jbedard, can you please squash?

alan-agius4 (approved):

PR #23650, opened 2022-07-28
test: add shard and test type/name information to test logging

jbedard:
Log info that I found useful.

### DIFF
--- a/tests/legacy-cli/e2e_runner.ts
+++ b/tests/legacy-cli/e2e_runner.ts
@@ -123,11 +123,19 @@ const tests = allTests.filter((name) => {
 const testsToRun = tests.filter((name, i) => shardId === null || i % nbShards == shardId);
 
 if (testsToRun.length === 0) {
-  console.log(`No tests would be ran, aborting.`);
-  process.exit(1);
+  if (shardId !== null && tests.length >= shardId ? 1 : 0) {
+    console.log(`No tests to run on shard ${shardId}, exiting.`);
+    process.exit(0);
+  } else {
+    console.log(`No tests would be ran, aborting.`);
+    process.exit(1);
+  }
 }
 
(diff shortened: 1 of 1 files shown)


tests/legacy-cli/e2e_runner.ts
```diff
@@ -123,11 +123,19 @@ const tests = allTests.filter((name) => {
 const testsToRun = tests.filter((name, i) => shardId === null || i % nbShards == shardId);
 
 if (testsToRun.length === 0) {
-  console.log(`No tests would be ran, aborting.`);
-  process.exit(1);
+  if (shardId !== null && tests.length >= shardId ? 1 : 0) {
```
jbedard, line 126: I guess this is actually a fix... if there are more shards then tests it shouldn't fail.

tests/legacy-cli/e2e_runner.ts
```diff
@@ -138,6 +146,8 @@ if (testsToRun.length == allTests.length) {
   console.log(`Running ${testsToRun.length} tests (${allTests.length} total)`);
 }
 
+console.log('Tests: \n  ', testsToRun.join('\n  '));
```
alan-agius4, line 149: I see what you are trying to do here, but actually this causes incorrect indent for the first item.

Screenshot 2022-07-28 at 09 25 28
Two ways can can address this

```ts
console.log('Tests:', testsToRun.join('\n  '));  
```
or
```ts
console.log(['Tests:', ...testsToRun].join('\n '));
```

alan-agius4 (approved):

tests/legacy-cli/e2e_runner.ts
jbedard: Good catch 👍

alan-agius4 (approved):

PR #23673, opened 2022-08-02
fix(@angular-devkit/schematics): Export 'rename' rule

tvsbrent:
- Export the 'rename' rule.
  - The other rules in this directory are exported, so I assumed this one should
    be exported as well.

### DIFF
--- a/goldens/public-api/angular_devkit/schematics/index.md
+++ b/goldens/public-api/angular_devkit/schematics/index.md
@@ -699,6 +699,9 @@ export interface RandomOptions {
 }
 
 // @public (undocumented)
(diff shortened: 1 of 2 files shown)


alan-agius4:
@tvsbrent, thanks for your contribution and apologies for taking so long to get to this PR.

We have discussed this during our triage meeting and we decided that we should remove this Rule although. 

Do you want to create a PR for that?

tvsbrent:
Thanks for the feedback! I can do that... although I've got some schematics I'd want to update first. 😄  I was a bad person and pulled a reference to rule directly, since it wasn't exported.

(Meaning, I did `import { rename } from '@angular-devkit/schematics/src/rules/rename.js';` in my code.)

PR #23686, opened 2022-08-04
test: remove ci specific logic

jbedard:
This seems unnecessary.

All the tests where the removed `if` is truthy seem to work locally and CI is green.

### DIFF
--- a/tests/legacy-cli/e2e_runner.ts
+++ b/tests/legacy-cli/e2e_runner.ts
@@ -160,7 +160,6 @@ if (testsToRun.length == allTests.length) {
 console.log(['Tests:', ...testsToRun].join('\n '));
 
 setGlobalVariable('argv', argv);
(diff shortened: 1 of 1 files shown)


tests/legacy-cli/e2e/utils/assets.ts
```diff
@@ -51,10 +51,6 @@ export async function createProjectFromAsset(
   process.chdir(dir);
   if (!useNpmPackages) {
     await useBuiltPackages();
-    if (!getGlobalVariable('ci')) {
```
jbedard, line 54: fwiw the registry is controlled using the environment variables. I think those variables even override this file so this was probably never doing anything?

jbedard:
Looks like other changes have made this PR essentially just removing an unused variable now 🤷

alan-agius4 (approved):

PR #23694, opened 2022-08-04
test: improve test error message logs

jbedard:


tests/legacy-cli/e2e/utils/process.ts
```diff
@@ -96,39 +96,47 @@ function _exec(options: ExecOptions, cmd: string, args: string[]): Promise<Proce
   _processes.push(childProcess);
 
   // Create the error here so the stack shows who called this function.
```
jbedard, line 98: Seems like this wasn't actually done before? Creating the `error` with the stack trace at this point is the primary change. Otherwise the stack trace is always just the child_process exit/error stack which is useless.

tests/legacy-cli/e2e/utils/process.ts
```diff
@@ -96,39 +96,47 @@ function _exec(options: ExecOptions, cmd: string, args: string[]): Promise<Proce
...
+  try {
+    throw new Error();
+  } catch (e) {
+    error = e as Error;
+  }
+
+  // Return log info about the current process status
+  function envDump() {
+    return [
+      `ENV:${JSON.stringify(spawnOptions.env, null, 2)}`,
```
jbedard, line 109: This also changed but is a bit hard to see. It now outputs the env of the process being spawned, not the parent process - `spawnOptions.env` isntead of `process.env`. I think that is far more useful and probably the original intention?

tests/legacy-cli/e2e/utils/process.ts
```diff
@@ -96,39 +96,47 @@ function _exec(options: ExecOptions, cmd: string, args: string[]): Promise<Proce
   _processes.push(childProcess);
 
   // Create the error here so the stack shows who called this function.
+  let error: Error;
```
alan-agius4, line 99: Wouldn't the below do the same?

```suggestion
  const error = new Error();
```

jbedard: Looks like you're correct. I always thought the stack was where the error was thrown, but it's actually just where it was constructed?

tests/legacy-cli/e2e/utils/process.ts
```diff
@@ -96,39 +96,42 @@ function _exec(options: ExecOptions, cmd: string, args: string[]): Promise<Proce
...
-    childProcess.on('exit', (error: any) => {
+    childProcess.on('exit', (code: number) => {
       _processes = _processes.filter((p) => p !== childProcess);
 
       if (options.waitForMatch && !matched) {
-        error = `Output didn't match '${options.waitForMatch}'.`;
+        reject(
+          `Process output didn't match - "${cmd} ${args.join(' ')}": '${
+            options.waitForMatch
+          }': ${code}...\n\n${envDump()}\n`,
```
jbedard, line 120: Note both the `code` (formerly `error`) and the `envDump()` were not here previously.

tests/legacy-cli/e2e/utils/process.ts
alan-agius4: That is correct. Can also change the `catch` to which is more readable. 

```ts
  }).catch((err) => {
    // Create the error here so the stack shows who called this function.
    return Promise.reject(new Error(err));
  });
```

alan-agius4 (approved):

tests/legacy-cli/e2e/utils/process.ts
jbedard: If you do the `new Error` in the catch callback doesn't that mean the stack trace will be just the callback?

alan-agius4: right

PR #23706, opened 2022-08-08
fix(@angular-devkit/build-angular): process stylesheet resources from url tokens with esbuild browser builder

clydin:
Stylesheet url tokens (`url(....)`) will now be processed when using the esbuild-based experimental browser
application builder. The paths will be resolved via the bundler's resolution system and then loaded
via the bundler's `file` loader. The functionality is implemented using an esbuild plugin to allow for all
file types to be supported without the need to manually specify each current and future file extension within
the build configuration.
The `externalDependencies` option also applies to the referenced resources. This allows for resource paths
specified with the option to remain unprocessed within the application output. This is useful if the relative
path for the resource does not exist on disk but will be available when the application is deployed.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/css-resource-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/css-resource-plugin.ts
@@ -1,0 +1,62 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import type { Plugin, PluginBuild } from 'esbuild';
+import { readFile } from 'fs/promises';
+
+/**
+ * Symbol marker used to indicate CSS resource resolution is being attempted.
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/css-resource-plugin.ts
```diff
@@ -0,0 +1,62 @@
...
+ * and types to be supported without needing to manually specify all extensions
+ * within the build configuration.
+ *
+ * @returns An esbuild {@link Plugin} instance.
+ */
+export function createCssResourcePlugin(): Plugin {
+  return {
+    name: 'angular-css-resource',
+    setup(build: PluginBuild): void {
+      build.onResolve({ filter: /.*/ }, async (args) => {
```
alan-agius4, line 30: - Should absolute paths be excluded?

clydin: Absolute URLs  and data URIs are automatically handled by esbuild during the resolution process.

alan-agius4 (approved):

PR #23709, opened 2022-08-09
build: disable UT testing multiple node versions

alan-agius4:
With this change we disable running UTs on multiple node.js versions. This is mainly an interim solution until we find the root cause of the memory increase and OOM errors during bazel test.

During this change I also noticed that the toolchain was configured incorrectly. Setting the toolchain to use Node.js 16 causes the App-shell tests to fail when used under Bazel. See: https://app.circleci.com/pipelines/github/angular/angular-cli/25009/workflows/4ea86ec8-d1f7-4c91-8810-76f10f8e878c/jobs/325982

### DIFF
--- a/tools/toolchain_info.bzl
+++ b/tools/toolchain_info.bzl
@@ -4,7 +4,8 @@
 # the order will match against the order in the TOOLCHAIN_VERSION list.
 TOOLCHAINS_NAMES = [
     "node14",
-    "node16",
+    # TODO enable one we know more why there is a memory usage increase and app-shell tests work with Node.js 16.
+    # "node16",
 ]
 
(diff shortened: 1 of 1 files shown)


alan-agius4:
//cc @jbedard

clydin (approved):

jbedard:
FWIW this issue seems similar? https://github.com/nrwl/nx/issues/10537

This is an option for now disabling the failing test and splitting the node 14 vs 16 into separate CI steps: https://github.com/angular/angular-cli/pull/23714

alan-agius4:
> FWIW this issue seems similar? https://github.com/nrwl/nx/issues/10537

It is similar, but the unlike the issue, the actually builder works when run outside of Bazel.

PR #23714, opened 2022-08-09
test: unit test multiple node versions in separate jobs

jbedard:
This re-enables the multiple node versions/toolchains but runs them as separate circleci jobs to not overload the executors. Also disabling the node16 version of the one test that seems to crash with node16.

### DIFF
--- a/.circleci/dynamic_config.yml
+++ b/.circleci/dynamic_config.yml
@@ -25,6 +25,10 @@ parameters:
 var_1: &cache_key v1-angular_devkit-14.19-{{ checksum "yarn.lock" }}
 var_1_win: &cache_key_win v1-angular_devkit-win-16.10-{{ checksum "yarn.lock" }}
 var_3: &default_nodeversion '14.19'
+var_3_major: &default_nodeversion_major '14'
+# The major version of node toolchains. See tools/toolchain_info.bzl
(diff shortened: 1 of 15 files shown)


.circleci/dynamic_config.yml
```diff
@@ -291,19 +295,39 @@ jobs:
...
-          # This timeout provides time for the actual tests to timeout and report status
-          # instead of CircleCI stopping the job without test failure information.
-          no_output_timeout: 40m
+      - when:
+          # The default nodeversion runs all *excluding* other versions
+          condition:
+            equal: [*default_nodeversion_major, << parameters.nodeversion >>]
+          steps:
+            - run:
+                command: yarn bazel test --test_tag_filters=-node16,-node<< parameters.nodeversion >>-broken //packages/...
```
jbedard, line 316: This ugly condition-x + condition-not-x is to ensure no tests are forgotten.

When running the `default_nodeversion_major` it runs all tests *excluding* other versions.
When running other version it runs all tests *including* that specific version.

If the `node{{nodeversion}}` tag is ever forgotten it will be included in the first case. Where if we always do the *including* (and got rid of this ugly condition) then forgetting a tag would cause the test to never run.

josephperrott (approved):
LGTM

alan-agius4 (approved):

PR #23718, opened 2022-08-10
fix(@angular/cli): add builders and schematic names as page titles in collected analytics

alan-agius4:
With this commit the builder and schematic names are added as page title to page events.

Also, we address a bug where during a watch or error in some cases analytics where not flushed. Examples when the builder has a watch mode.

Screenshot 2022-08-10 at 11 25 11

### DIFF
--- a/packages/angular/cli/src/command-builder/command-module.ts
+++ b/packages/angular/cli/src/command-builder/command-module.ts
@@ -140,8 +140,11 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
 
     // Gather and report analytics.
     const analytics = await this.getAnalytics();
+    let stopPeriodicFlushes: (() => Promise<void>) | undefined;
+
     if (this.shouldReportAnalytics) {
       await this.reportAnalytics(camelCasedOptions);
+      stopPeriodicFlushes = this.periodicAnalyticsFlush(analytics);
     }
 
     let exitCode: number | void | undefined;
@@ -151,7 +154,6 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
       exitCode = await this.run(camelCasedOptions as Options<T> & OtherOptions);
       const endTime = Date.now();
       analytics.timing(this.commandName, 'duration', endTime - startTime);
-      await analytics.flush();
     } catch (e) {
       if (e instanceof schema.SchemaValidationException) {
         this.context.logger.fatal(`Error: ${e.message}`);
@@ -160,6 +162,8 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
         throw e;
       }
     } finally {
+      await stopPeriodicFlushes?.();
+
       if (typeof exitCode === 'number' && exitCode > 0) {
         process.exitCode = exitCode;
       }
@@ -170,6 +174,7 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
     options: (Options<T> & OtherOptions) | OtherOptions,
     paths: string[] = [],
     dimensions: (boolean | number | string)[] = [],
+    title?: string,
   ): Promise<void> {
     for (const [name, ua] of this.optionsWithAnalytics) {
       const value = options[name];
@@ -183,6 +188,7 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
     analytics.pageview('/command/' + [this.commandName, ...paths].join('/'), {
       dimensions,
       metrics: [],
+      title,
     });
   }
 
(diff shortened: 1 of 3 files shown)


clydin (approved):

packages/angular/cli/src/command-builder/command-module.ts
```diff
@@ -139,9 +139,17 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
...
+    let analyticsFlushInterval: NodeJS.Timeout | undefined;
     const analytics = await this.getAnalytics();
+
     if (this.shouldReportAnalytics) {
       await this.reportAnalytics(camelCasedOptions);
+
+      // Flush on an interval (if the event loop is waiting).
+      analyticsFlushInterval = setInterval(() => {
+        analyticsFlushPromise = analyticsFlushPromise.then(() => analytics.flush());
+      }, 2000);
```
dgp1130, line 152: I forget the exact edge cases of `setInterval()`, but is there a risk multiple tasks get queued and executed at the same time? What about if a report to analytics is very slow, or we're hung on a long synchronous task?

https://developer.mozilla.org/en-US/docs/Web/API/setInterval#ensure_that_execution_duration_is_shorter_than_interval_frequency

packages/angular/cli/src/command-builder/command-module.ts
```diff
@@ -139,9 +139,17 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
...
+    let analyticsFlushPromise = Promise.resolve();
+    let analyticsFlushInterval: NodeJS.Timeout | undefined;
     const analytics = await this.getAnalytics();
+
     if (this.shouldReportAnalytics) {
       await this.reportAnalytics(camelCasedOptions);
+
+      // Flush on an interval (if the event loop is waiting).
+      analyticsFlushInterval = setInterval(() => {
+        analyticsFlushPromise = analyticsFlushPromise.then(() => analytics.flush());
```
dgp1130, line 151: I assume this is the fix for watch-mode analytics. Would this be a better feature in the analytics module rather than in the command builder?

packages/angular/cli/src/command-builder/command-module.ts
```diff
@@ -160,6 +167,10 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
         throw e;
       }
     } finally {
+      // Flush one last time.
+      clearInterval(analyticsFlushInterval);
+      await analyticsFlushPromise.then(() => analytics.flush());
```
dgp1130, line 172: This looks a lot like an Observable pattern. I'm wondering if we should consider using that or maybe an async generator to manage the `setInterval()` / `clearInterval()` part in a more ergonomic way?

packages/angular/cli/src/command-builder/command-module.ts
alan-agius4: That should not be a problem in this case as calling `flush` multiple times before the http request is done will not trigger the same events.

See: https://github.com/angular/angular-cli/blob/05a0b28b2f5b1a67d5c6cb6a1e1cce64c06e2329/packages/angular/cli/src/analytics/analytics-collector.ts#L158-L161

dgp1130:
Currently figuring out the privacy approval process for this change, so don't merge this just yet.

packages/angular/cli/src/command-builder/command-module.ts
alan-agius4: This technically would need to be an API change as analytics API is in `@angular-devkit/core`.  https://github.com/angular/angular-cli/tree/main/packages/angular_devkit/core/src/analytics

I don't however see a completing reason why analytics API should expose a period flush method considering this is quite a specific case where a flush need to happen periodicly.

packages/angular/cli/src/command-builder/command-module.ts
alan-agius4: I updated this but I avoid to use RxJs as it's not a dependency and in this case while it was possible to use  generators it does make it more complex. 

Therefore I went with a simple method that returns a method which can be called to run the teardown logic.

packages/angular/cli/src/command-builder/command-module.ts
```diff
@@ -140,8 +140,11 @@ export abstract class CommandModule<T extends {} = {}> implements CommandModuleI
 
     // Gather and report analytics.
     const analytics = await this.getAnalytics();
+    let periodicAnalyticsFlush: (() => Promise<void>) | undefined;
```
dgp1130, line 143: Nit: Name this variable more clearly after what it will do, like `stopPeriodicFlushes`, so this makes more sense when it is called.

dgp1130 (commented):
LGTM, just waiting on privacy before approving.

dgp1130:
I just realized we also list out the data collected on AIO: https://angular.io/analytics#what-is-collected. We should update that as well when this is ready to land.

dgp1130:
This is fully privacy approved, so we're good to merge this in and release: http://launch/4185851.

Don't forget to update https://angular.io/analytics#what-is-collected, whether in this PR or another one.

alan-agius4:
> This is fully privacy approved, so we're good to merge this in and release: http://launch/4185851.
> 
> Don't forget to update https://angular.io/analytics#what-is-collected, whether in this PR or another one.

Awesome. Has to be a separate PR as it's a different repo.

sleepynano (approved):

sleepynano (approved):
packages/angular/cli/src/command-builder/schematics-command-module.ts

PR #23737, opened 2022-08-12
ci: reduce windows parallelism from 16 to 14

alan-agius4:
Since reducing the `resource_class` we started seeing flakes similar to https://app.circleci.com/pipelines/github/angular/angular-cli/25049/workflows/e451d67c-a44b-49d8-919b-928bf98c3473/jobs/326520

Therefore, we try to reduce parallelism to see if this flakiness improves.


alan-agius4:
Doesn't appear to solve the issue.

PR #23750, opened 2022-08-16
feat(@angular-devkit/schematics): update zone.js to 0.11.8 and add async tagging API support

JiaLiPassion:
1. Update to `zone.js` 0.11.8
2. `import 'zone.js/plugins/async-stack-tagging` in `development` mode
to support `async tagging API` for better debugging experience.

For more information, please refer to this PR https://github.com/angular/angular/pull/46693


alan-agius4:
Can you please change the scope to `@schematics/angular`? 

Thanks

JiaLiPassion:
@alan-agius4 , sure, updated, thank you.

packages/schematics/angular/application/files/src/environments/environment.ts.template
```diff
@@ -14,3 +14,4 @@ export const environment = {
  * on performance if an error is thrown.
  */
 // import 'zone.js/plugins/zone-error';  // Included with Angular CLI.
+import 'zone.js/plugins/async-stack-tagging'; // Included with Angular CLI to support async stack tagging API.
```
alan-agius4, line 17: Question: Should this be enabled by default or commented out like the above line and users can opt-in? Does it have any side effects of having this enabled for production?

JiaLiPassion: @alan-agius4 
- I believe this one should be included in the `development` mode by default since this need to integrate with `Angular Devtools`.
- in production mode, if we enable it, it will have a little impact to the performance, since it will add a new `parent Zone` to the `ngZone`, just like `TaskTackingZone`.
@mgechev , could you confirm my understanding is correct? Thank you.

alan-agius4: At the moment there is no way to conditionally include polyfills based on if it’s development/production. 

I think a better approach would be to add zone-error and async-stack-tagging internally in the builder so that we can conditionally add this based on if the build is optimized not.

mgechev: We'd like to enable it by default in development so folks can get complete stack traces and not use it in production. Async stack tagging (AST) is independent from Angular DevTools and we use it to concatenate async stack traces so that we can give the entire picture to developers.

AST is very similar to Zone's long stack traces with the difference that they work with debugging tools out of the box.

JiaLiPassion:
This is handled by https://github.com/angular/angular/pull/47416.

PR #23752, opened 2022-08-16
fix(@angular-devkit/build-angular): fail ng test if compilation

### LINKED ISSUE #23751: ng test exits with success even if test code does not compile
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 13.2.2

### Description

A regression introduced in v13.2.3 of the CLI has been allowing test code that doesn't even compile to appear as a passing job in our CICD system.

The problem can be traced to [this change](https://github.com/angular/angular-cli/pull/22670/files#r801922504) which includes an unconditional promise resolve following compilation, no matter if there are compilation errors. I think I understand the intent here, but I believe the `singleRun` scenario was overlooked. When running in watch mode, it's true that you always want Karma to start up and the exit code of the command is irrelevant, since the command runs until the user forces it to stop. But within CICD systems `singleRun` is the norm and the exit code is very important. We do not want karma to start if the compilation fails and the command should exit with a non zero code.

I think all that is required is to add a reject handler and call that one instead it there are errors `&& singleRun`. I have verified this fix locally and will open a PR with precisely this change against this issue.

## 🔬 Minimal Reproduction

Create an Angular project using ng cli 13.2.3 or above.

Create one unit test. Include a compilation error in the test.

Ensure the karma option `failOnEmptyTestSuite` is `false`.

Run `ng test --project=<project-name> --watch=false`

Observe that karma starts despite the compilation error and the command exits with success (0 code).

This is all captured in the following reproduction repo: https://github.com/vigie/false-positive-repo-app

## 🔥 Exception or Error

<pre><code>

</code></pre>

## 🌍 Your Environment

<pre><code>
                |___/

Angular CLI: 13.3.8
Node: 14.20.0
Package Manager: npm 6.14.17
OS: linux x64

Angular: 13.3.11
... platform-server, router

@angular-devkit/architect       0.1303.8
@angular-devkit/build-angular   13.3.8
@angular-devkit/core            13.3.8
@angular-devkit/schematics      13.3.8
@angular/cdk                    13.3.9
@angular/cli                    13.3.8
@schematics/angular             13.3.8
ng-packagr                      13.3.1
typescript                      4.6.3
webpack                         5.73.0
</code></pre>

**Anything else relevant?**

As mentioned, this was introduced at v13.2.3 and as far as I can tell still exists in latest v14. The fix should be backported and released in v13 also.

vigie:
…errors exist and in singleRun mode

A bug was introduced in v13.2.3
(ref https://github.com/angular/angular-cli/pull/22670/files)
that makes the `ng test` command exit with success when run in
`singleRun` mode, even in the presence of compilation errors.

This change fixes the issue by calling karma `init`'s `reject` handler
if compilation errors are present and running in `singleRun` mode.

The effect is karma does not start and the browser does not launch,
instead the process simply terminates in error. This behavior is very
important to have for CICD systems.

Fixes #23751

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
@@ -161,7 +161,7 @@ const init: any = (config: any, emitter: any) => {
   let lastCompilationHash: string | undefined;
   let isFirstRun = true;
 
-  return new Promise<void>((resolve) => {
+  return new Promise<void>((resolve, reject) => {
     compiler.hooks.done.tap('karma', (stats) => {
       if (isFirstRun) {
         // This is needed to block Karma from launching browsers before Webpack writes the assets in memory.
@@ -169,7 +169,9 @@ const init: any = (config: any, emitter: any) => {
         // https://github.com/karma-runner/karma-chrome-launcher/issues/154#issuecomment-986661937
         // https://github.com/angular/angular-cli/issues/22495
         isFirstRun = false;
-        resolve();
+        // In singleRun mode, if there are compilation errors, the process should terminate 
+        // in error at this point. In watch mode, the process is kept alive regardless.
+        config.singleRun && stats.hasErrors() ? reject() : resolve();
       }
 
       if (stats.hasErrors()) {


alan-agius4:
Hi @vigie, 

Thanks for taking a stab at this. 

Unfortunately, this doesn't solve the issue as it keep the process hanging. 

Feel free to open a new PR, if you find a fix for this. 

Thanks.

PR #23753, opened 2022-08-16
test: run e2e tests on pre-compiled packages

jbedard:
The NPM packages to test must be specified via --package instead of invoking the build from within the e2e tests

This means the packages must be built before tested. This will align with bazel which will pass the bazel built (and cached) tgz files. The packages being tested could be the real npm packages before being published, although CI will probably test snapshot-like packages.


alan-agius4:
NIT: in general it would be great that the PR description is part of the commit message. As it add more context when checking the commit history.

jbedard:
I've updated the commit message a bit (and rebased). Let me know if there's other information you think should be in there.

scripts/build.ts
```diff
@@ -73,7 +73,6 @@ function _tar(out: string, dir: string) {
       strict: true,
       portable: true,
       cwd: dir,
-      prefix: 'package/',
```
jbedard, line 76: I don't understand this one... this puts everything within a `package` directory in the tar, but the real npm packages do not have that directory. Why is this here and why does it work when the real npm packages don't have it?

It was added [to fix a windows issue](https://github.com/angular/angular-cli/commit/f5fe1b60d42b3aa7359b592b5e917a3b41b11398)? But the public packages (such as [cli](https://unpkg.com/browse/@angular/cli@14.0.0/) don't have the directory?

tests/legacy-cli/e2e/utils/tar.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Extract and return the contents of a single file out of a tar file.
+ *
+ * @param tarball the tar file to extract from
+ * @param filePath the path of the file to extract
+ * @returns the Buffer of file or an error on fs/tar error or file not found
+ */
+export async function extractFile(tarball: string, filePath: string): Promise<Buffer> {
+  return new Promise((resolve, reject) => {
+    let chunks: Buffer[] | null = null;
```
josephperrott, line 22: Is there a reason that the chunks need defined in this block?  Since its only referenced inside of the `onentry` closure, you should be able to just define it there correct?

josephperrott (approved):
LGTM

tests/legacy-cli/e2e/utils/tar.ts
jbedard: It was probably just me learning the fs/parse API. Updated 👍

josephperrott (approved):
LGTM

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -95,15 +94,17 @@ export async function prepareProjectForE2e(name: string) {
...
+  const packages = getGlobalVariable('package-tars');
+
   return updateJsonFile('package.json', (json) => {
     json['dependencies'] ??= {};
     json['devDependencies'] ??= {};
 
     for (const packageName of Object.keys(packages)) {
       if (packageName in json['dependencies']) {
-        json['dependencies'][packageName] = packages[packageName].version;
+        json['dependencies'][packageName] = packages[packageName];
```
alan-agius4, line 105: Since we have access to the version in `findPackageTars` can we keep using the version instead?

tests/legacy-cli/e2e_runner.ts
```diff
@@ -49,6 +53,7 @@ const argv = yargsParser(process.argv.slice(2), {
   ],
   string: ['devkit', 'glob', 'ignore', 'reuse', 'ng-tag', 'tmpdir', 'ng-version'],
   number: ['nb-shards', 'shard'],
+  array: ['package'],
```
alan-agius4, line 56: For the time being can we default this to `./dist/_*.tgz`? The command itself is already a bit long and having to write `--package=/dist/_*.tgz` multiple times a day seems a bit unnecessary.

For context here's an example of the command now.
```
node tests/legacy-cli/run_e2e.js tests/legacy-cli/e2e/tests/build/library/lib-consumption-full-aot.ts --package=/dist/_*.tgz
```

alan-agius4 (changes requested):
Can you please update the developers guide? 

https://github.com/angular/angular-cli/blob/main/docs/DEVELOPER.md#end-to-end-tests

tests/legacy-cli/e2e/utils/project.ts
```diff
@@ -220,8 +221,12 @@ export async function useCIChrome(projectDir: string = ''): Promise<void> {
   }
 }
 
-export const NgCLIVersion = new SemVer(packages['@angular/cli'].version);
```
jbedard, line 223: Since `packages` is now computed from the package tars passed on the CLI there's no guarantee that it has been put into the "global variable"s when this file is first loading, so I've moved it into a method.

Previously in this PR I had this doing `require(../../../....../package.json).version` instead of reading it from the loaded packages. However a) that is not the same as before this PR and b) the git repo version might not be the same as the packages being tested.

jbedard:
FWIW `CODE_OF_CONDUCT.md` is currently causing the CI "validation" step to fail.

jbedard:
Updated and squashed, see [diff](https://github.com/angular/angular-cli/compare/3cd0a6a..7b4e421ffd4846a84f5513483bfe2c796d94ae4e)

alan-agius4 (approved):

docs/DEVELOPER.md
```diff
@@ -84,6 +84,7 @@ You can find more info about debugging [tests with Bazel in the docs.](https://g
 
 - Run: `node tests/legacy-cli/run_e2e.js`
```
alan-agius4, line 85: Can you add a line here that now running `yarn build` needs to be ran prior to `node tests/legacy-cli/run_e2e.js` as the latter will not build the packages.

jbedard: Updated

alan-agius4 (approved):

PR #23764, opened 2022-08-18
build: update to TypeScript 4.8 RC

crisbeto:
Updates to the TypeScript 4.8 RC and adds some code to account for a breaking change where the decorators and modifiers of an AST node have been combined into a single array.


packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -76,6 +65,75 @@ export function replaceResources(
...
+  moduleKind: ts.ModuleKind | undefined,
+  inlineStyleFileExtension: string | undefined,
+): ts.ClassDeclaration {
+  let decorators: ts.Decorator[] | undefined;
+  let modifiers: ts.Modifier[] | undefined;
+
+  if (IS_TS_48) {
+    node.modifiers?.forEach((modifier) => {
+      if (ts.isDecorator(modifier)) {
+        decorators = decorators || [];
```
dgp1130, line 90: Nit: Would `decorators ??= []` be a more modern syntax for this?

packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -76,6 +65,75 @@ export function replaceResources(
...
+      resourceImportDeclarations,
+      moduleKind,
+      inlineStyleFileExtension,
+    ),
+  );
+
+  return IS_TS_48
+    ? nodeFactory.updateClassDeclaration(
+        node,
+        [...decorators, ...(modifiers || [])],
```
dgp1130, line 120: Nit: I think `modifiers ?? []` would work here and be a little more intuitive?

packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -327,3 +385,14 @@ function getDecoratorOrigin(
...
+
+/** Checks if the current version of TypeScript is after the specified major/minor versions. */
+function isAfterVersion(targetMajor: number, targetMinor: number): boolean {
+  const [major, minor] = ts.versionMajorMinor.split('.').map((part) => parseInt(part));
+
+  if (major < targetMajor) {
+    return false;
+  }
+
+  return major === targetMajor ? minor >= targetMinor : true;
```
dgp1130, line 397: Nit: I recommend leaving this logic in the if/else for symmetry. Otherwise the `major > targetMajor` case is a bit obfuscated.

```typescript
if (major < targetMajor) {
  return false;
} else if (major > targetMajor) {
  return true;
} else {
  return minor => targetMinor;
}
```

dgp1130 (approved):

crisbeto:
The feedback has been addressed.

PR #23768, opened 2022-08-19
build: use same node version in bazel as CI

jbedard:
Versions should probably align?


jbedard:
Actually there are version mismatches all over the place so I'm going to just close this. I originally did it while debugging an odd error wondering if it was something to do with different version numbers locally vs CI etc...

PR #23770, opened 2022-08-19
build: prevent busting bazel analysis cache between build and test

jbedard:
The first+last in the diff prevent busting the analysis cache between build+test. The others are minor improvements or will be required for e2e tests.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,5 +1,5 @@
 # Disable NG CLI TTY mode
-test --action_env=NG_FORCE_TTY=false
+build --action_env=NG_FORCE_TTY=false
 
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -146,6 +152,9 @@ build:remote --google_default_credentials
 # These settings are required for rules_nodejs
 ###############################
 
+# Fixes use of npm paths with spaces such as some within the puppeteer module
+build --experimental_inprocess_symlink_creation
```
alan-agius4, line 156: 👍

alan-agius4 (approved):

PR #23787, opened 2022-08-24
fix(@angular-devkit/build-angular): issue warning when using deprecated tilde imports

alan-agius4:
With this change we add a warning when using deprecated tilde/`~` imports.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
@@ -9,7 +9,7 @@
 import * as fs from 'fs';
 import MiniCssExtractPlugin from 'mini-css-extract-plugin';
 import * as path from 'path';
-import { Configuration, RuleSetUseItem } from 'webpack';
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -306,6 +324,15 @@ export function getStylesConfig(wco: WebpackConfigOptions): Configuration {
             implementation: sassImplementation,
             sourceMap: true,
             sassOptions: {
+              importer: (url: string, from: string) => {
+                if (url.charAt(0) === '~') {
+                  sassTildeUsageMessage.add(
+                    `'${from}' imports '${url}' with a tilda. Usage of '~' in imports is deprecated.`,
```
JoostK, line 330:
```suggestion
                    `'${from}' imports '${url}' with a tilde. Usage of '~' in imports is deprecated.`,
```

;)

clydin (approved):

clydin (approved):

PR #23789, opened 2022-08-24
build: allow no-remote-exec targets to be cached on CI

jbedard:


.bazelrc
```diff
@@ -51,7 +50,7 @@ test --incompatible_strict_action_env
 build --experimental_remote_merkle_tree_cache
 
 # Ensure that tags applied in BUILDs propagate to actions
-build --incompatible_allow_tags_propagation
+build --experimental_allow_tags_propagation
```
jbedard, line 53: This was renamed quite some time ago. `incompatible_allow_tags_propagation` still works but will be removed eventually

.bazelrc
```diff
@@ -34,8 +34,7 @@ build --symlink_prefix=dist/
 build --nowatchfs
 
 # Turn off legacy external runfiles
-run --nolegacy_external_runfiles
-test --nolegacy_external_runfiles
+build --nolegacy_external_runfiles
```
jbedard, line 37: Only doing this for run/test meant the analysis cache was broken after not doing it for build.

.bazelrc
```diff
@@ -136,9 +135,6 @@ build:remote --extra_execution_platforms=@npm//@angular/build-tooling/bazel/remo
 build:remote --host_platform=@npm//@angular/build-tooling/bazel/remote-execution:platform_with_network
 build:remote --platforms=@npm//@angular/build-tooling/bazel/remote-execution:platform_with_network
 
-# Set remote caching settings
-build:remote --remote_accept_cached=true
```
jbedard, line 140: This is the default. IMO only setting it (to the default) on `--config=remote` is confusing...

.circleci/dynamic_config.yml
```diff
@@ -136,14 +136,6 @@ commands:
             # cause decryption failures based on the openssl version. https://stackoverflow.com/a/39641378/4317734
             openssl aes-256-cbc -d -in .circleci/gcp_token -md md5 -k "${<< parameters.key >>}" -out /home/circleci/.gcp_credentials;
             sudo bash -c "echo -e 'build --google_credentials=/home/circleci/.gcp_credentials' >> .bazelrc.user";
-            # Upload/don't upload local results to cache based on environment
-            if [[ -n "{$CIRCLE_PULL_REQUEST}" ]]; then
-              sudo bash -c "echo -e 'build:remote --remote_upload_local_results=false\n' >> .bazelrc.user";
```
jbedard, line 141: Why was this done? Disabling remote caching for targets where remote execution is disabled...

Normally this would only be done for things like local dev to read but not write to the remote cache. I thought maybe it was to avoid polluting the cache server with random PRs but most bazel targets (all but 3 or 4) get executed remote and are cached despite this flag.

But this is what was causing a few of the unit tests and all the e2e tests (e2e being a [WIP](https://github.com/angular/angular-cli/pull/23074)) to never be cached.

josephperrott (approved):
LGTM

.bazelrc
devversion: I wonder if we still need it— we don't use it in other repos AFAIK. Nothing to worry here though, just bring it up.

.bazelrc
devversion: Should it just be set for all of the three commands? run/test & build?

devversion (approved):
LGTM, just one comment about the legacy externa runfiles

.bazelrc
jbedard: It's a new/experimental feature that will hopefully eventually become the default. I'm not sure if it's required for anything in this repo but I've started using it.

.bazelrc
jbedard: run + test inherit from build, so this change sit to be set 100% of the time

See "Option precedence:" in https://bazel.build/run/bazelrc

devversion: Ohh, I didn't know that and always specified explicitly. Nice finding

.bazelrc
devversion: Okay, not opposed to keeping it here, just wanted to see if if we actively rely on it

PR #23790, opened 2022-08-25
fix(@angular-devkit/build-angular): correct esbuild builder global stylesheet sourcemap URL

clydin:
The sourcemap URL in the output CSS files for global stylesheets is now correctly updated to
reflect the name of the global stylesheet output file and not the internal `stdin` virtual
file name.


alan-agius4 (approved):

dgp1130:
This has a merge conflict with 14.1.x. Given that it's going away (hopefully) today, I'll switch to `target: rc` and land the change in `14.2.0`.

PR #23794, opened 2022-08-25
build: update ng-dev version

josephperrott:
update to the latest version of ng-dev

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "@angular/forms": "14.2.0-rc.0",
     "@angular/localize": "14.2.0-rc.0",
     "@angular/material": "14.2.0-rc.0",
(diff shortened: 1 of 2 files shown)


dgp1130 (approved):

dgp1130:
This has conflicts with `14.1.x` and `14.2.x`. Switching to `target: minor` to land this on the `main` branch.

* I'll make a manual cherry-pick Pr for `14.2.x`, since that will become the new patch branch.
* I don't think we care about `14.1.x` since that's (hopefully) going away today, so I'll skip that for now. I'm assuming this update isn't necessary for a final patch release? If so, I can make a cherry-pick for that too.

josephperrott:
Works for me, since its really just about making the release process works, it will only run on the `main` branch anyway

dgp1130:
Cherry pick to `14.2.x` in #23795.

PR #23801, opened 2022-08-26
fix(@schematics/angular): generate new workspaces using TypeScript 4.8

alan-agius4:


alan-agius4:
The failures are caused by usage of `getMutableClone` in the compiler. 
@crisbeto replaced most of the usages however this was only merged in `main`.  https://github.com/angular/angular/pull/47167

PR #23807, opened 2022-08-26
refactor(@angular-devkit/build-angular): remove deprecated usage of `canParse`

alan-agius4:
All parsers now implement `parser.analyze`.


packages/angular_devkit/build_angular/src/utils/load-translations.ts
```diff
@@ -26,9 +26,11 @@ export async function createTranslationLoader(): Promise<TranslationLoader> {
     const content = fs.readFileSync(path, 'utf8');
     const unusedParsers = new Map();
     for (const [format, parser] of Object.entries(parsers)) {
-      const analysis = analyze(parser, path, content);
+      const analysis = parser.analyze(path, content);
       if (analysis.canParse) {
-        const { locale, translations } = parser.parse(path, content, analysis.hint);
+        // Types don't overlap here so we need to use any.
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
clydin, line 32: Is this an issue with the types on the `@angular/localize` side?
Seems like the result of a parser's analysis should work for the same parser's parse function.

clydin (approved):

packages/angular_devkit/build_angular/src/utils/load-translations.ts
alan-agius4: Not really, it’s more of a combination that the parsers are in array and TypeScript doesn’t infer the type correctly.

PR #23811, opened 2022-08-26
fix(@schematics/angular): update ng-packagr version to `^14.2.0`

dgp1130:
This is one-half of the fix in https://github.com/angular/angular-cli/pull/23801 in case we need to decouple this fix from the TS 4.8 update.


alan-agius4:
Can you change the target to the patch branch as on main it needs to be updated to v15?

dgp1130:
@alan-agius4, good point. I was gonna ask about that in your PR but just forgot. Thanks for catching it.

alan-agius4 (approved):

PR #23812, opened 2022-08-26
Bump version to "v14.2.1" with changelog.

dgp1130:


clydin (approved):

alan-agius4:
Shouldn’t https://github.com/angular/angular-cli/pull/23805 be included regardless?

dgp1130:
@alan-agius4, I thought that was only for TS4.8 support? Do we need that to fix the `ng new` `-next` problem?

alan-agius4:
Technically ^14.0.0-next , ^14.0.0 and ^14.2.0 will install the same version.

For TS 4.8, the min version however would need to be 14.2.

dgp1130:
Sure, I'm deferring all the TS4.8 work for today. So I think we can skip that PR for now and come back to it after `ng-conf` right?

PR #23820, opened 2022-08-30
refactor(@schematics/angular): remove deprecated appDir option

alan-agius4:
BREAKING CHANGE: Removed unused`appDir` option from Universal schematic


clydin (approved):
Can you add a brief description to the breaking change notice that mentions the option can safely be removed if present since it no longer has an effect?

PR #23822, opened 2022-08-30
refactor(@angular-devkit/build-angular): remove old `bundleDependencies` logic

alan-agius4:


clydin:
The string form of the option appears to be still present in the schema.  Should that be removed as well?
https://github.com/angular/angular-cli/blob/a0f3f66da052f705648fe4b02ab58a943fc4eea4/packages/angular_devkit/build_angular/src/builders/server/schema.json#L191-L194

alan-agius4:
Good catch

clydin (approved):

PR #23853, opened 2022-09-02
test: enable bazel e2e windows ci job

jbedard:
Extension of https://github.com/angular/angular-cli/pull/23074 to get running on windows while keeping that one green.


jbedard:
Moving to https://github.com/angular/angular-cli/pull/23074 now that it's passing

Issue #23844, opened 2022-09-01
[ng serve]: webpack-dev-middleware Error: ENOENT: no such file or directory


ramoncarreras:
### Command

serve

### Is this a regression?

- [X] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

14.1.3

### Description

After upgrading to angular 14.2.0 and serving the APP locally, I get the following error:

```
[webpack-dev-middleware] Error: ENOENT: no such file or directory, readdir '<project-directory>/test-app/dist/test-app/'
    at createError (<project-directory>/test-app/node_modules/memfs/lib/volume.js:128:17)
    at Volume.readdirBase (<project-directory>/test-app/node_modules/memfs/lib/volume.js:1415:19)
    at Immediate.<anonymous> (<project-directory>/test-app/node_modules/memfs/lib/volume.js:695:33)
    at processImmediate (node:internal/timers:466:21) {
  code: 'ENOENT'
}
```

### Minimal Reproduction

I created a minimal reproduction in the following repository:

https://github.com/ramoncarreras/angular-test-app

I have tried to minimize the issue as I could, sorry in advance (also see each commit).

In the README.md of the above repository there is a small description of the steps to follow.

The issue appears when I add the localize and the PWA official angular packages.

### Exception or Error

```text
[webpack-dev-middleware] Error: ENOENT: no such file or directory, readdir '<project-directory>/test-app/dist/test-app/'
    at createError (<project-directory>/test-app/node_modules/memfs/lib/volume.js:128:17)
    at Volume.readdirBase (<project-directory>/test-app/node_modules/memfs/lib/volume.js:1415:19)
    at Immediate.<anonymous> (<project-directory>/test-app/node_modules/memfs/lib/volume.js:695:33)
    at processImmediate (node:internal/timers:466:21) {
  code: 'ENOENT'
}
```
```


### Your Environment

```text
_                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 14.2.1
Node: 16.14.0
Package Manager: npm 8.5.3 
OS: darwin x64

Angular: 
... 

[...]

PR #23858, opened 2022-09-05
fix(@angular-devkit/build-angular): change service worker errors to compilation errors

alan-agius4:
Previously, when there was an error during a build that had service workers enabled, the dev-server crashed as the promise was rejected instead of emitting a compilation error.

With this change we update the logic so that any errors during the SW augmentation phase are changed to a compilation error and also update the logic so that when there are compilation errors we don't try to generate a SW.

See: https://github.com/angular/angular-cli/issues/23844#issuecomment-1236311405

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/service-worker-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/service-worker-plugin.ts
@@ -20,8 +20,15 @@ export class ServiceWorkerPlugin {
   constructor(private readonly options: ServiceWorkerPluginOptions) {}
 
   apply(compiler: Compiler) {
-    compiler.hooks.done.tapPromise('angular-service-worker', async ({ compilation }) => {
+    compiler.hooks.done.tapPromise('angular-service-worker', async (stats) => {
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/plugins/service-worker-plugin.ts
```diff
@@ -31,17 +37,25 @@ export class ServiceWorkerPlugin {
...
+          baseHref,
+          ngswConfigPath,
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (compiler.inputFileSystem as any).promises,
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (compiler.outputFileSystem as any).promises,
+        );
+      } catch (error) {
+        compilation.errors.push(
+          new WebpackError(
```
clydin, line 54:
```suggestion
          new compilation.compiler.webpack.WebpackError(
```
Removes the need to change the import and also guarantees the same Webpack instance that is executing the plugin.

clydin (approved):

PR #23866, opened 2022-09-07
feat(@angular-devkit/build-angular): export `@angular/platform-server` symbols in server bundle

alan-agius4:
This commit adds an internal file to export needed symbols from `@angular/platform-server` when building a server bundle. This is needed.  This is needed so that DI tokens can be referenced and set at runtime outside of the bundle.

Also, it adds a migration to remove these exports from the users files as otherwise an export collision would occur due to the same symbol being exported multiple times.

### DIFF
--- a/packages/schematics/angular/migrations/update-15/remove-platform-server-exports_spec.ts
+++ b/packages/schematics/angular/migrations/update-15/remove-platform-server-exports_spec.ts
@@ -1,0 +1,85 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { EmptyTree } from '@angular-devkit/schematics';
+import { SchematicTestRunner } from '@angular-devkit/schematics/testing';
+
+describe('Migration to delete platform-server exports', () => {
+  const schematicName = 'remove-platform-server-exports';
+
+  const schematicRunner = new SchematicTestRunner(
+    'migrations',
+    require.resolve('../migration-collection.json'),
+  );
+
+  let tree: EmptyTree;
+
+  beforeEach(() => {
+    tree = new EmptyTree();
+  });
+
+  const testTypeScriptFilePath = './test.ts';
+
+  describe(`Migration to remove '@angular/platform-server' exports`, () => {
+    it(`should delete '@angular/platform-server' export when 'renderModule' is the only exported symbol`, async () => {
+      tree.create(
+        testTypeScriptFilePath,
+        `
+          import { Path, join } from '@angular-devkit/core';
+          export { renderModule } from '@angular/platform-server';
+        `,
+      );
+
+      const newTree = await schematicRunner.runSchematicAsync(schematicName, {}, tree).toPromise();
+      const content = newTree.readText(testTypeScriptFilePath);
+      expect(content).not.toContain('@angular/platform-server');
+      expect(content).toContain(`import { Path, join } from '@angular-devkit/core';`);
+    });
+
+    it(`should delete only 'renderModule' when there are additional exports`, async () => {
+      tree.create(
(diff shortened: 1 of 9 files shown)


packages/angular_devkit/build_angular/src/builders/server/platform-server-exports.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+export * from '@angular/platform-server';
```
alan-agius4, line 9: @AndrewKushnir, maybe we can be more selective here and export what we need.

AndrewKushnir: I like the idea of being more conservative and just export what we need. At least, we can do that from the very beginning and change this later if needed (if this becomes an issue and we end up updating this constantly). If this is a viable solution overall, I can update https://github.com/angular/angular/pull/47103 to add a new token that we re-export here.

Meanwhile, would it be possible to export our "rendering" functions from here, such as `renderModule` and `renderApplication` (for standalone use-cases), so we don't need changes to the `main.server.ts` to proxy the `renderApplication` to make it available to NgUniversal?

alan-agius4: Yeah that is viable. 

To export `renderModule` we'd need to do a migration to remove it from the `main.server.ts` exports to avoid colliding exports. 

For the `renderApplication`, that is correct given that users **will never** need to set `platformProviders` themselves. If there are cases were users would need to set such providers than a "proxy" is the the only way forward. 

// cc @clydin to see if he has any objections to this change.

clydin: sounds good.  I definitely prefer the minimal approach to start.

alan-agius4:
Blocked on https://github.com/angular/angular-cli/pull/23907

alan-agius4:
Webpack has a bug which does not export symbols from multiple entries

https://github.com/webpack/webpack/issues/15936

packages/schematics/angular/migrations/update-15/remove-platform-server-exports_spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    tree = new UnitTestTree(new EmptyTree());
+  });
+
+  const testTypeScriptFilePath = './test.ts';
+
+  describe('Migration to import() style lazy routes', () => {
+    beforeEach(async () => {
+      tree = new UnitTestTree(new EmptyTree());
+      tree.create('/package.json', JSON.stringify({}));
+    });
```
clydin, line 32: This looks like a copy/paste leftover?

packages/schematics/angular/migrations/update-15/remove-platform-server-exports_spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    it(`should not delete 'renderModule' when it's exported from another module`, async () => {
+      tree.create(
+        testTypeScriptFilePath,
+        `
+          export { renderModule } from '@angular/core';
+        `,
+      );
+
+      const newTree = await schematicRunner.runSchematicAsync(schematicName, {}, tree).toPromise();
+      const content = newTree.readContent(testTypeScriptFilePath);
```
clydin, line 73: nit: `Tree` has `readText` now so `readContent` isn't needed anymore.  Could probably also drop `UnitTestTree` completely in the spec as well.

clydin (approved):

packages/schematics/angular/migrations/update-15/remove-platform-server-exports_spec.ts
alan-agius4: Yes

PR #23881, opened 2022-09-08
build: use pkg_npm(tgz) instead of extra pkg_tar() rule

jbedard:


jbedard:
Turns out this is buggy on windows...

PR #23903, opened 2022-09-12
test: catch and log all errors include test setup

jbedard:
I had a case of something else above throwing and it was hidden due to this.

### DIFF
--- a/tests/legacy-cli/e2e_runner.ts
+++ b/tests/legacy-cli/e2e_runner.ts
@@ -181,6 +181,8 @@ Promise.all([findFreePort(), findFreePort()])
       } else {
         console.log(colors.green('Done.'));
       }
+
+      process.exitCode = 0;
     } catch (err) {
       if (err instanceof Error) {
         console.log('\n');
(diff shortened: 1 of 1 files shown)


tests/legacy-cli/e2e_runner.ts
```diff
@@ -175,13 +175,22 @@ Promise.all([findFreePort(), findFreePort()])
...
+  .then(
+    () => {
       if (shardId !== null) {
         console.log(colors.green(`Done shard ${shardId} of ${nbShards}.`));
       } else {
         console.log(colors.green('Done.'));
       }
-    } catch (err) {
+
+      process.exit(0);
```
alan-agius4, line 191: Why setting the exit code is needed? (If needed it would be better to use `process.exitCode` instead.)

tests/legacy-cli/e2e_runner.ts
```diff
@@ -175,13 +175,22 @@ Promise.all([findFreePort(), findFreePort()])
       await runSteps(runSetup, allSetups, 'setup');
       await runSteps(runInitializer, allInitializers, 'initializer');
       await runSteps(runTest, testsToRun, 'test');
-
+    } finally {
+      registryProcess.kill();
```
alan-agius4, line 179: This would cause the NPM registry to be terminated as well when debug is enabled.

jbedard: Good point 👍

alan-agius4 (approved):
Can you please squash the fixups? Thanks

jbedard:
Squashed 👍

PR #23910, opened 2022-09-14
ci: remove manually specifying GITHUB_TOKEN in scorecard action

clydin:
The `ossf/scorecard-action` action will automatically use the builtin GITHUB_TOKEN as needed.


josephperrott (approved):
LGTM

PR #23916, opened 2022-09-15
ci: adjust workflow permissions for OpenSSF Scorecard results publishing

clydin:
Version 2 of the `openssf/scorecard-action` GitHub action requires the `id-token: write` permission to publish results.


josephperrott (approved):
LGTM

PR #23925, opened 2022-09-16
build: fix yarn ng-dev bin on windows

jbedard:
This doesn't seem to work on windows (CI or Windows 11)? 🤷


alan-agius4:
This is expected as GitBash is required for Windows. 

See: https://github.com/angular/angular/issues/46858 and https://github.com/angular/angular/pull/46859#discussion_r922683816

PR #23926, opened 2022-09-17
test: enable git core.longpaths option

jbedard:
If things like the temp dir, project root etc have long paths (such as custom paths passed on the cli, or bazel runfiles/temp directories) then this is required.

### DIFF
--- a/.circleci/dynamic_config.yml
+++ b/.circleci/dynamic_config.yml
@@ -129,7 +129,6 @@ commands:
       - run: npm install -g yarn@1.22.10
       - run: node --version
       - run: yarn --version
-      - run: git config --global core.longpaths true
 
   setup_bazel_rbe:
     parameters:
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):
Can you please remove https://github.com/angular/angular-cli/blob/91133fc2445d83d0f9a8cbfae820f54adf56277c/.circleci/dynamic_config.yml#L132?

jbedard:
You don't think that might be necessary for other git actions? Or was it only for git being launched within the e2e tests?

alan-agius4:
> You don't think that might be necessary for other git actions? Or was it only for git being launched within the e2e tests?

It was the only one.

jbedard:
I guess as little as possible should be CI specific, updated 👍

alan-agius4 (approved):

PR #23932, opened 2022-09-19
test: remove unused E2E size tracking artifact generation test

clydin:
These files are no longer used to perform any size tracking and are also no longer stored as artifacts within CircleCI.


alan-agius4 (approved):
LGTM

PR #23939, opened 2022-09-21
refactor(@angular-devkit/build-angular): simplify how specs are found and loaded

### LINKED ISSUE #22531: Running tests with code coverage doesn't include files outside the sourceRoot directory
# 🐞 Bug report

- [ ] new
- [ ] build
- [ ] serve
- [x] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: Any time before PR #11974

### Description

## 🔬 Minimal Reproduction

1. ng new test-app
2. create packages/test-package outside of src folder
3. add there simple test.component.ts and test.component.spec.ts
4. Modify test.ts file and add context of 'packages directory
replace
```javascript
const context = require.context('./', true, /\.spec\.ts$/);
context.keys().map(context1);
```
with
```javascript
const context1 = require.context('../packages', true, /\.spec\.ts$/);
const context2 = require.context('./', true, /\.spec\.ts$/);

context1.keys().map(context1);
context2.keys().map(context2);
```
5. Add "../packages/**/*.spec.ts", to tsconfig.spec.json in "include" section

6. Run `ng test --code-coverage` and open coverage/index.html

## 🔥 Exception or Error

No error but the files outside the directory are not included in the coverage report.

**Anything else relevant?**

This was reported previously on issue #12474. They simply dismissed it as saying only files under `sourceRoot` are listed on the coverage report. Having the intended behavior be where tests from outside `sourceRoot` are ran but not included in the coverage report seems counterintuitive. They should either both be ran and included in the coverage report or not included in either place.

### LINKED ISSUE #23751: ng test exits with success even if test code does not compile

alan-agius4:
Prior to this change specs where found and loaded using Webpack's `require.context` API. The `require.context` is found in the users project `test.ts`. This resulted in a complex and hacky setup especially to filter tests when the `include` builder option is provided, were we had to amend the `test.ts` in memory.

With this change we find all the specs files and add them as part of the main entrypoint.

Closes #23751 and closes #22531


packages/angular_devkit/build_angular/src/builders/karma/index.ts
```diff
@@ -110,45 +106,22 @@ export function execute(
...
-            loader: SingleTestTransformLoader,
-            options: {
-              files,
-              logger: context.logger,
-            },
-          },
-        });
-      }
+      webpackConfig.plugins ??= [];
+      webpackConfig.plugins?.push(
```
clydin, line 118: nit: `plugins` should always be defined via the previous statement

packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
```diff
@@ -124,10 +124,10 @@ const init: any = (config: any, emitter: any) => {
 
       logger.error(statsErrorsToString(statsJson, { colors: true }));
 
-      // Notify potential listeners of the compile error.
-      emitter.emit('compile_error', {
-        errors: statsJson.errors?.map((e) => e.message),
-      });
+      if (config.singleRun) {
+        // Notify potential listeners of the compile error.
+        emitter.emit('load_error');
```
clydin, line 129: Why the change to the event name here?

packages/angular_devkit/build_angular/src/builders/karma/find-tests-plugin.ts
```diff
@@ -6,15 +6,76 @@
...
+
+export class FindTestsPlugin {
+  private compilation: Compilation | undefined;
+
+  constructor(private options: FindTestsPluginOptions) {}
+
+  apply(compiler: Compiler): void {
+    const { include = ['**/*.spec.ts'], projectSourceRoot, workspaceRoot } = this.options;
+
+    compiler.hooks.environment.tap(PLUGIN_NAME, () => {
```
clydin, line 38: Is this hook needed?  Could the compiler options be modified directly in the apply function?

packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
alan-agius4: load_error is needed as otherwise in some cases karma will jot exit when there is a webpack error as karma doesn’t listen to compilation_error.

packages/angular_devkit/build_angular/src/builders/karma/find-tests-plugin.ts
alan-agius4: I believe so, will update.

packages/angular_devkit/build_angular/src/builders/karma/index.ts
alan-agius4: You are right 👍

clydin (approved):

vigie:
Is it possible to backport this to the 14 branch so we can consume the fix to https://github.com/angular/angular-cli/issues/23751?

PR #23947, opened 2022-09-23
feat(@angular-devkit/build-angular): remove `bundleDependencies` from server builder

### LINKED ISSUE #23905: bundleDependencies option does not support false in angular 13/14.
### Command

build

### Is this a regression?

angular 12

### Description

Since angular 13 shipped with ESM package format only, if we build the project with bundleDependencies=false, it will report an error like this:
> require() of ES Module /Users/kame/code/web/node_modules/@angular/core/fesm2015/core.mjs not supported.

I know we can set bundleDependencies=true to make the build work. But I found ssr render performance with all dependencies bundled is worse than unbundled in previous version(angular 12, not sure why). 

### Minimal Reproduction

Here is the minimal repo to reproduce the issue: https://github.com/ganatan/angular-ssr
Steps to reproduce:
1. Clone the repo
2. Change "bundleDependencies" option to false for server build in angular.json file
3. Run command `npm run build:ssr`
4. Run `npm run serve:ssr`

### Exception or Error

```text
node:internal/modules/cjs/loader:979
    throw new ERR_REQUIRE_ESM(filename, true);
    ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/kame/code/github/angular-ssr/node_modules/@angular/common/fesm2015/common.mjs not supported.
Instead change the require of /Users/kame.chen/code/github/angular-ssr/node_modules/@angular/common/fesm2015/common.mjs to a dynamic import() which is available in all CommonJS modules.
    at Object.8260 (/Users/kame.chen/code/github/angular-ssr/dist/angular-starter/server/main.js:517:18)
    at __webpack_require__ (/Users/kame.chen/code/github/angular-ssr/dist/angular-starter/server/main.js:671:41)
    at /Users/kame.chen/code/github/angular-ssr/dist/angular-starter/server/main.js:814:73
    at /Users/kame.chen/code/github/angular-ssr/dist/angular-starter/server/main.js:870:3
    at Object.<anonymous> (/Users/kame.chen/code/github/angular-ssr/dist/angular-starter/server/main.js:875:12) {
  code: 'ERR_REQUIRE_ESM'
```

### Your Environment

```text
Angular CLI: 14.1.3
Node: 16.15.0
Package Manager: npm 8.5.5
OS: darwin x64

Angular: 14.1.3
... router, service-worker

@angular-devkit/architect       0.1401.3
@angular-devkit/build-angular   14.1.3
@angular-devkit/core            14.1.3
@angular-devkit/schematics      14.1.3
@nguniversal/builders           14.1.0
@nguniversal/express-engine     14.1.0
@schematics/angular             14.1.3
rxjs                            7.5.6
typescript                      4.7.4
```

### Anything else relevant?

_No response_

alan-agius4:
This commit removes the usages of `bundleDependencies` which does not correctly work as webpack will use `require` to import ESM module since we configure the server bundle to be outputted in CJS. Migrating fully to ESM is also currently not viable due to the lack of support from Domino.
Even if full ESM was possible, using this option would have resulted in a runtime overhead as Angular libraries would be linked during runtime instead of compile time.

BREAKING CHANGE:
The server builder `bundleDependencies` option has been removed. This option was used pre Ivy. Currently, using this option is unlikely to produce working server bundles.

The `externalDependencies` option can be used instead to exclude specific node_module packages from the final bundle.

Closes #23905


packages/schematics/angular/migrations/migration-collection.json
```diff
@@ -14,6 +14,11 @@
       "version": "15.0.0",
       "factory": "./update-15/update-typescript-target",
       "description": "Update TypeScript compiler `target` and set `useDefineForClassFields`. These changes are for IDE purposes as TypeScript compiler options `target` and `useDefineForClassFields` are set to `ES2022` and `false` respectively by the Angular CLI. To control ECMA version and features use the Browerslist configuration."
+    },
+    "update-workspace-config": {
+      "version": "13.0.0",
```
clydin, line 19:
```suggestion
      "version": "15.0.0",
```

clydin (approved):

packages/schematics/angular/migrations/migration-collection.json
alan-agius4: 😅

PR #23951, opened 2022-09-23
feat(@angular-devkit/build-angular): add `ng-server-context` when using app-shell builder

alan-agius4:
With this change we configure the app-shell builder to set the `ɵSERVER_CONTEXT` private provider.

FYI @AndrewKushnir


AndrewKushnir (approved):
@alan-agius4 thanks for adding this 👍

clydin (approved):

PR #23955, opened 2022-09-26
feat(@schematics/angular): add migration to remove require calls from karma builder main file

alan-agius4:
With the recent changes in build-angular the `require.context` calls have become unneeded.

BREAKING CHANGE: `require.context` are no longer parsed. Webpack specific features are not supported nor guaranteed to work in the future.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -342,8 +342,7 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
       strictExportPresence: true,
       parser: {
         javascript: {
-          // TODO(alanagius): disable the below once we have a migration to remove `require.context` from test.ts file in users projects.
-          requireContext: true,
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -342,8 +342,7 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
       strictExportPresence: true,
       parser: {
         javascript: {
-          // TODO(alanagius): disable the below once we have a migration to remove `require.context` from test.ts file in users projects.
-          requireContext: true,
+          requireContext: false,
```
clydin, line 345: While this is not an officially supported feature, we might want to add a breaking change notice to let users know that this Webpack specific feature will definitely not work anymore. (And maybe reiterate that Webpack specific features are not guaranteed to work in the future.)

alan-agius4: Let’s do that.

PR #23962, opened 2022-09-27
CI build-bazel-e2e fix

alan-agius4:
See each commit.

### DIFF
--- a/.circleci/dynamic_config.yml
+++ b/.circleci/dynamic_config.yml
@@ -271,6 +271,7 @@ jobs:
 
   build-bazel-e2e:
     executor: action-executor
+    resource_class: medium
     steps:
       - custom_attach_workspace
       - run: yarn bazel build //tests/legacy-cli/...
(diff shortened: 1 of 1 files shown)


devversion (approved):

devversion:
One thing I noticed: Usually Bazel has problems determining the memory/CPU ceilings and we explicitly set them in the .circleci/bazelrc but here the CI bazelrc config is not wired up in all Bazel jobs (as it seems): https://github.com/angular/angular-cli/blob/main/.circleci/dynamic_config.yml#L298

alan-agius4:
Good point, although the values there are for an xlarge class
https://github.com/angular/angular-cli/blob/308e3a017f876bfc727e68803bfbce11e9d3396e/.circleci/bazel.rc#L10-L15

devversion:
Yes, but Bazel might incorrectly compute the ceiling for the classes currently used (e.g. medium)

alan-agius4:
Yes you right.

PR #23989, opened 2022-09-30
feat(@angular-devkit/build-angular): add initial watch support to esbuild-based builder

clydin:
The experimental esbuild-based browser application builder now contains initial support for
watching input files and rebuilding the application via the `--watch` option. This initial
implemention is not yet optimized for incremental rebuilds and will perform a full rebuild
upon detection of a change. Incremental rebuild support will be added in followup changes
and will significantly improve the rebuild speed.
The `chokidar` npm package is used to perform the file watching which allows for native file-
system event based watching. Polling is also support via the `--poll` option for environments
that require it.

### DIFF
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -27,6 +27,7 @@
     "babel-plugin-istanbul": "6.1.1",
     "browserslist": "^4.9.1",
     "cacache": "16.1.3",
+    "chokidar": "3.5.3",
     "copy-webpack-plugin": "11.0.0",
     "critters": "0.0.16",
     "css-loader": "6.7.1",
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
@@ -28,40 +28,14 @@ import { logExperimentalWarnings } from './experimental-warnings';
 import { normalizeOptions } from './options';
 import { Schema as BrowserBuilderOptions, SourceMapClass } from './schema';
 import { bundleStylesheetText } from './stylesheets';
+import { createWatcher } from './watcher';
 
-/**
- * Main execution function for the esbuild-based application builder.
- * The options are compatible with the Webpack-based builder.
- * @param options The browser builder options to use when setting up the application build
- * @param context The Architect builder context object
- * @returns A promise with the builder result output
- */
-// eslint-disable-next-line max-lines-per-function
-export async function buildEsbuildBrowser(
+async function execute(
   options: BrowserBuilderOptions,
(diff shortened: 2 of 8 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
```diff
@@ -410,4 +369,87 @@ async function bundleGlobalStylesheets(
...
+  // Setup a watcher
+  const watcher = createWatcher({
+    polling: typeof initialOptions.poll === 'number',
+    interval: initialOptions.poll,
+    // Ignore the output path to avoid infinite rebuild cycles
+    ignored: [normalizedOptions.outputPath],
+  });
+
+  // Temporarily watch the entire workspace
+  watcher.add(context.workspaceRoot);
```
alan-agius4, line 437: Shouldn't this be the project root?

packages/angular_devkit/build_angular/package.json
```diff
@@ -27,6 +27,7 @@
     "babel-plugin-istanbul": "6.1.1",
     "browserslist": "^4.9.1",
     "cacache": "16.1.3",
+    "chokidar": "3.5.3",
```
alan-agius4, line 30: Can you please add this to the main package.json as well?

packages/angular_devkit/build_angular/src/builders/browser-esbuild/watcher.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+export interface BuildWatcher extends AsyncIterableIterator<ChangedFiles> {
+  add(paths: string | string[]): void;
+  remove(paths: string | string[]): void;
+  close(): Promise<void>;
+}
+
+export function createWatcher(options?: {
```
alan-agius4, line 33: Cool usage of async iterators.

packages/angular_devkit/build_angular/package.json
clydin: Already there but not pinned (i'll fix that).
https://github.com/angular/angular-cli/blob/6f8142ec7e5e9f8b074411d8ff2efa8e6bbad979/package.json#L139

packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
clydin: It is temporarily overly broad right now but I wanted to ensure that any relevant changes are picked up initially.
My logic for workspace root was that it would capture package changes as well as tsconfig changes but the project root would work for essentially all other cases and avoid unrelated generated projects from trigger a rebuild.
This will become more fine-grained once the dependencies are propagated from the build steps either way.

alan-agius4 (approved):

PR #24008, opened 2022-10-04
feat(@schematics/angular): remove `ngOnInit` from component template

### LINKED ISSUE #23678: Add flag to ng generate component that skips adding ngOnInit() interface.
# 🚀 Feature request

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

Generating components with `ng g c component-name` always generates a component implementing `ngOnInit()` interface.

Many components do not benefit from implementing such interface. A good example might be a dumb component, that uses mostly `@Input()` and `@Output()`.

Right now there is no way to override this behaviour so if you want to generate a dumb component, you need to:

1. remove `ngOnInit()` method, 
2. remove `implements OnInit` clause from the component class,
3. remove the OnInit import from '@angular/core';

This has to be repeated for every component generated, so it gets old really fast 🤷 

Ideally, there should be an option to generate a component that does not implement the `ngOnInit()` hook.

Maybe something like this could work:
```ng g c component-name --skip-on-init``` or ```ng g c component-name --skip-lifecycle```

It would match the naming of flags such as `--skip-tests` or `--skip-selector`.

alan-agius4:
This commit removes the `constructor` and `ngOnInit` as most of the users prefer to add this manually if needed.

Closes #23678


Brocco:
I like this change!

clydin (approved):

PR #24019, opened 2022-10-05
refactor(@angular/cli) Address issues and enable strictNullChecks.

gunan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Currently, builds happen without strictNullChecks enabled in TS compiler. This change enables that.

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular_devkit/core/src/logger/logger.ts
+++ b/packages/angular_devkit/core/src/logger/logger.ts
@@ -161,7 +161,8 @@ export class Logger extends Observable<LogEntry> implements LoggerApi {
       arguments as unknown as Parameters<Observable<LogEntry>['subscribe']>,
     );
   }
-  override forEach(next: (value: LogEntry) => void, PromiseCtor?: typeof Promise): Promise<void> {
-    return this._observable.forEach(next, PromiseCtor);
+
+  override forEach(next: (value: LogEntry) => void, promiseCtor?: typeof Promise): Promise<void> {
+    return this._observable.forEach(next, promiseCtor);
   }
 }
--- a/goldens/public-api/angular_devkit/core/index.md
+++ b/goldens/public-api/angular_devkit/core/index.md
@@ -616,7 +616,7 @@ class Logger extends Observable<LogEntry> implements LoggerApi {
     // (undocumented)
     fatal(message: string, metadata?: JsonObject): void;
     // (undocumented)
-    forEach(next: (value: LogEntry) => void, PromiseCtor?: typeof Promise): Promise<void>;
+    forEach(next: (value: LogEntry) => void, promiseCtor?: typeof Promise): Promise<void>;
     // (undocumented)
     info(message: string, metadata?: JsonObject): void;
     // (undocumented)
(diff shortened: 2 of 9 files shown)


tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "rootDir": ".",
     "skipLibCheck": true,
     "strict": true,
+    "strictNullChecks": true,
```
alfaproject, line 17: This is redundant. `strict` already includes this flag

gunan: Done

packages/angular_devkit/core/src/logger/logger.ts
```diff
@@ -161,7 +161,8 @@ export class Logger extends Observable<LogEntry> implements LoggerApi {
       arguments as unknown as Parameters<Observable<LogEntry>['subscribe']>,
     );
   }
-  override forEach(next: (value: LogEntry) => void, PromiseCtor?: typeof Promise): Promise<void> {
-    return this._observable.forEach(next, PromiseCtor);
+
+  override forEach(next: (value: LogEntry) => void, promiseCtor?: typeof Promise): Promise<void> {
+    return this._observable.forEach(next, promiseCtor as PromiseConstructor);
```
alan-agius4, line 166: The casting here is redundant. 

```suggestion
    return this._observable.forEach(next, promiseCtor);
```

alan-agius4 (changes requested):
Thanks for your contribution.

Couple of points mainly related to the commit message 
- Please remove the commit message scope as your are amending multiple packages and not just the Angular CLI. 
- is incorrect as it states that it is enabling StrictNullChecks which is not the case as strictNullChecks are already enabled.
- The type should be a `refactor` as it doesn't provide a `fix`

Something like the below:
```
refactor: remove add implicate types for better code readability
```

packages/angular_devkit/core/src/logger/logger.ts
gunan: Actually, it is an optional value, and when `strictNullChecks` is enabled, compiler throws an error for this line when nothing is done. I initially added a non-null assertion `!` but linter did not like it, so I added this cast.
I can take an alternative approach if you have a suggestion.

gunan: Done, thank you for your guidance. Please take a look.

gunan:
Commit message updated. Please take a look

packages/angular_devkit/core/src/logger/logger.ts
alan-agius4: I am a bit confused by your statement as strictNullChecks are already enabled and the build currently succeeds.

alan-agius4: Let’s do that.I am good with the other changes though. So if you want to revert this I will LGTM.

gunan: I reverted the cast, but I kept the variable naming change, as it was confusing. `PromiseCtor` looked more like a type rather than a local variable. Is that OK?

alan-agius4: Yes, I totally agree with you

alan-agius4 (approved):
LGTM

alan-agius4 (approved):
LGTM

PR #24026, opened 2022-10-06
feat(@angular/cli): add support for Node.js version 18

alan-agius4:
Pacote version 14 does requires `14.17.x` or `16.13.x`.

BREAKING CHANGE: The Angular CLI  no longer supports `16.10.x`, `16.11.x` and `16.12.x`. Current supported versions of Node.js are `14.20.x`, `16.13.x` and `18.10.x`.


alfaproject:
Oh man, there's a huge problem with Jest for node 16.11+ so this means we won't be able to upgrade any time soon ):

Context: https://github.com/facebook/jest/issues/11956

alan-agius4:
Going through the issue and linked issues in the Node.js repo https://github.com/nodejs/node/issues/44211 and https://github.com/nodejs/node/issues/42080 it doesn't seem like it's something that is being tacked by the Node.js.

alfaproject:
Yes, exactly. It's been a bit of a mess for months and we haven't been able to update to higher versions of node yet. I guess if the worst comes to worst we will just have to pay for xtra xtra xtra large CI runners. I was kind of hoping that we could stay another year on 16.10 if none of our dependencies had the need to update but it looks like that won't happen, meh

clydin (approved):
The Jest situation is unfortunate.  However, Node.js 16.10 is over a year old now and 16.13 is technically the first v16 LTS release.  More and more packages will start using newer APIs and this will most likely accelerate once Node.js v14 is EOL in several months.
From reading through the linked Jest issue, there does appear to be several PRs with workarounds to some of the problems being encountered.  Hopefully a new release of Jest can mitigate the issue.
@alfaproject Thank you for bringing up the Jest issue.  The team will continue to monitor the issue and potentially evaluate if anything can be done to help mitigate the issue.

PR #24028, opened 2022-10-06
fix(@angular/cli): skip downloading temp CLI when running `ng update` without package names

### LINKED ISSUE #24024: Useless installation of temporary CLI
### Command

update

### Is this a regression?

_No response_

### Description

I'm not sure if this should be a bug or a feature, it's more of an annoyance.

Every week when I check with 'ng update' it first complains that the CLI is outdated and starts installing a temporary CLI version, which takes a while. Then it tells me there is a new CLI version and I have to install it again with 'ng update @angular/cli', this time for real. Next time I do a 'ng update' it quickly shows me, without the delay.

Is it really necessary to install a new temporary cli version every week ?

Or maybe if there is a new one, can't we have question or a parameter to install the new CLI version (automatically), instead of installing it temporarily (and thus twice)? Just like when you type 'ng new' you also get a question.

### Minimal Reproduction

> ng update

The installed Angular CLI version is outdated.
Installing a temporary Angular CLI versioned 14.2.5 to perform the update.
√ Packages successfully installed.

### Exception or Error

_No response_

### Your Environment

```text
Angular CLI: 14.2.5
Node: 16.16.0
Package Manager: npm 8.3.0
OS: win32 x64
```

### Anything else relevant?

_No response_

alan-agius4:
In the case when `ng update` is ran without a package name, an update is not be performed. In this case using the current installed version of the CLI is good enough.

Closes #24024

### DIFF
--- a/packages/angular/cli/src/utilities/package-metadata.ts
+++ b/packages/angular/cli/src/utilities/package-metadata.ts
@@ -292,17 +292,10 @@ export async function getNpmPackageJson(
   const { usingYarn = false, verbose = false, registry } = options;
   ensureNpmrc(logger, usingYarn, verbose);
   const { packument } = await import('pacote');
-  const resultPromise = packument(packageName, {
+  const response = packument(packageName, {
     fullMetadata: true,
     ...npmrc,
     ...(registry ? { registry } : {}),
-  });
-
-  // TODO: find some way to test this
-  const response = resultPromise.catch((err) => {
-    logger.warn(err.message || err);
-
-    return { requestedName: packageName };
   });
 
   npmPackageJsonCache.set(packageName, response);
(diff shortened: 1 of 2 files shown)


packages/angular/cli/src/utilities/package-metadata.ts
```diff
@@ -292,19 +292,12 @@ export async function getNpmPackageJson(
...
   ensureNpmrc(logger, usingYarn, verbose);
   const { packument } = await import('pacote');
-  const resultPromise = packument(packageName, {
+  const response = packument(packageName, {
     fullMetadata: true,
     ...npmrc,
     ...(registry ? { registry } : {}),
   });
 
-  // TODO: find some way to test this
```
alan-agius4, line 301: Not sure why this was here @clydin maybe you remember? This needs to be removed as otherwise certain errors are not fatal. Example: https://app.circleci.com/pipelines/github/angular/angular-cli/26850/workflows/10782a5e-36c3-4de6-ae85-0ebcfe456e4e/jobs/356140/parallel-runs/0/steps/0-105

The test previously failed because when trying to download the temp version of the CLI. This check is bypassed since pacote is not used.

clydin (approved):

packages/angular/cli/src/utilities/package-metadata.ts
clydin: May have been leftover from before the package names were validated to be registry names.

PR #24032, opened 2022-10-07
Auto include `@angular/platform-server/init` and `@angular/localize/init`

alan-agius4:
**feat(@angular-devkit/build-angular): auto include `@angular/platform-server/init` during server builds**
    
This changes removes the need to import `@angular/platform-server/init` in the `main.server.ts` instead we now include this as an entry-point when we are bundling as server bundle.

--- 
**feat(@angular-devkit/build-angular): auto include `@angular/localize/init` when found in `types`**
    
With this change we remove the need to include the `@angular/localize/init` package as polyfills in `polyfills.ts`, `angular.json` and `main.server.ts`.


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -95,14 +95,30 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
     extraPlugins.push(new ProgressPlugin(platform));
   }
 
+  const localizePackageInitEntryPoint = '@angular/localize/init';
+  if (tsConfig.options.types?.includes(localizePackageInitEntryPoint)) {
```
clydin, line 99: Could we have this check both `@angular/localize/init` and `@angular/localize`? I'd like to expose the global types in the later as well (or eventually instead) since its possible to use `$localize` without the `init` when build localization is enabled.

clydin (approved):

PR #24035, opened 2022-10-07
refactor(@angular-devkit/schematics): remove `UpdateBuffer` and rename `UpdateBuffer2` to `UpdateBuffer`

kyubisation:
This PR removes the internally built `UpdateBuffer` and renames `UpdateBuffer2` (based on magic-string) to `UpdateBuffer`. This should have little to no impact for consumers.

BREAKING CHANGE:
The depracated `UpdateBuffer` has been removed and `UpdateBuffer2` is renamed to `UpdateBuffer`. With this change the related and deprecated symbols `ContentCannotBeRemovedException` and `Chunk` have also been removed.

## PR Checklist

Please check to confirm your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular-cli/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

The deprecated UpdateBuffer instance is being used per default.

Issue Number: N/A

## What is the new behavior?

The magic-string based UpdateBuffer2 is being used per default.

## Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

## Other information

See #21209 and #21110

PR #24044, opened 2022-10-10
Revert "refactor(@angular-devkit/schematics): remove `UpdateBuffer` and rename `UpdateBuffer2` to `UpdateBuffer`"

alan-agius4:
Reverts angular/angular-cli#24035

Reverting as this breaks Angular material schematics in a number of ways.

See: https://angular-team.slack.com/archives/C46U16D4Z/p1665219000524689

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/recorder_spec.ts
+++ b/packages/angular_devkit/schematics/src/tree/recorder_spec.ts
@@ -7,6 +7,7 @@
  */
 
 import { normalize } from '@angular-devkit/core';
+import { UpdateBuffer2, UpdateBufferBase } from '../utility/update-buffer';
 import { SimpleFileEntry } from './entry';
 import { UpdateRecorderBase, UpdateRecorderBom } from './recorder';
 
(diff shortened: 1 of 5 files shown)


kyubisation:
@alan-agius4 Could I have some additional insight into how it breaks Angular Material? If at all possible I would like to try to fix the issue.

alan-agius4:
We tried to fix this before doing the revert and ultimately we deemed it too risky to do this for version 15. 

We do plan to remove do the breaking change to remove  `UpdateBuffer` in version 16 where it would not be necessary to fix the [legacy components material migrations](https://github.com/angular/components/tree/a0ec4becacf840fe553099699121d2bde0d6e585/src/material/schematics/ng-update/migrations/legacy-components-v15). 

This migration is rather tricky as it's keeping track of the indices of the content when using postcss. This is very fragile and works due to a bug in the deprecated `UpdateBuffer`.

An easier approach would be for post css to on the entire stylesheet and the schematic would replace the entire css. This way there would not be the need to keep the indices synced between postcss and the schematic host taking into consideration the length of the content modified/deleted.

alan-agius4:
> Thank you for the explanation. I can easily live with the UpdateBuffer change being postponed to 16. Should I re-raise the PR for 16 or will the Angular team take care of it?

Feel free to open a PR for this right now, and I'll just put it in the list for the next major version.

PR #24071, opened 2022-10-12
fix(@angular-devkit/build-angular): account for package.json exports fields with CSS import statements

clydin:
The `postcss-imports` package was previously used to support `@import` within CSS files. Unfortunately, the package does not account for `package.json` exports fields. This prevents imports defined within that field from working when used within a build.  The `css-loader` package does provide this functionality and is now used to provide support for CSS `@import` instead of `postcss-imports`. This change does not affect preprocessors that provide their own import behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -185,7 +185,6 @@
     "piscina": "3.2.0",
     "popper.js": "^1.14.1",
     "postcss": "8.4.18",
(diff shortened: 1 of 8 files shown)


alan-agius4 (approved):
This is great!

PR #24086, opened 2022-10-14
test(@angular/cli): disable analytics gathering during CLI E2E

alan-agius4:
Currently, analytics is enabled which causes a lot of noise in our analytics report.


alan-agius4:
``

PR #24088, opened 2022-10-14
fix(@schematics/angular): add `@angular/localize` as type when localize package is installed

alan-agius4:
The `@angular/localize` entrypoint now exposes the global `$localize` method type.

See: https://github.com/angular/angular/pull/47763


alan-agius4:
Blocked on FW release 15.0.0-next.7.

PR #24146, opened 2022-10-26
perf(@angular-devkit/build-angular): initialize Angular shims for first rebuild in esbuild builder

clydin:
When in watch mode with the experimental esbuild-based browser builder, the Angular compiler's internal type checking shims are now initialized earlier in the initial build. This allows the TypeScript builder program to become aware of the existence of these files during the initial build instead of during the first rebuild. As a result, the TypeScript builder program will no longer assume all the shim files and associated project files have been affected and required to be emitted and type checked again. This change causes the first rebuild to be similar in performance to subsequent rebuilds.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/compiler-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/compiler-plugin.ts
@@ -253,6 +253,7 @@ export function createCompilerPlugin(
       const babelDataCache = new Map<string, Uint8Array>();
       const diagnosticCache = new WeakMap<ts.SourceFile, ts.Diagnostic[]>();
 
+      // eslint-disable-next-line max-lines-per-function
       build.onStart(async () => {
         const result: OnStartResult = {
           warnings: setupWarnings,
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/compiler-plugin.ts
```diff
@@ -346,7 +347,30 @@ export function createCompilerPlugin(
...
+        if (!previousBuilder && pluginOptions.sourceFileCache) {
+          profileSync('NG_INITIALIZE_SHIMS', () => {
+            for (const sourceFile of angularProgram.getTsProgram().getSourceFiles()) {
+              // The file cannot be a declaration file to ensure the shim generation is triggered.
+              if (sourceFile.isDeclarationFile) {
+                continue;
+              }
+              angularCompiler.getDiagnosticsForFile(sourceFile, OptimizeFor.WholeProgram);
+              if (angularProgram.getTsProgram() !== angularCompiler.getCurrentProgram()) {
+                // Sucessful if the two programs are no longer the same
```
alan-agius4, line 366:
```suggestion
                // Successful if the two programs are no longer the same
```

PR #24152, opened 2022-10-27
fix(@angular/cli): exclude `@angular/localize@<10.0.0` from ng add pa…

alan-agius4:
…ckage discovery

`@angular/localize@<10.0.0` has no peer dependencies. This can cause `ng add` to pick these versions of the package if the newer versions.

See: https://app.circleci.com/pipelines/github/angular/angular-cli/27402/workflows/faa64532-541a-4bea-b599-3c53afe42019/jobs/364822
```
Test Process error Error: Process exit error - "ng add @angular/localize --skip-confirmation": 1...

STDOUT:

STDERR:
npm version 7.4.0 detected. When using npm 7 with the Angular CLI, npm version 7.5.6 or higher is recommended.
- Determining package manager...
ℹ Using package manager: npm
- Searching for compatible package version...
✔ Found compatible package version: @angular/localize@10.0.0-next.7.
- Loading package information from registry...
✔ Package information loaded.
- Installing packages...
✔ Packages successfully installed.
NOT SUPPORTED: keyword "id", use "$id" for schema ID
```

### DIFF
--- a/packages/angular/cli/src/commands/add/cli.ts
+++ b/packages/angular/cli/src/commands/add/cli.ts
@@ -49,7 +49,7 @@ interface AddCommandArgs extends SchematicsCommandArgs {
  * The key is a package name and the value is a SemVer range of versions to exclude.
  */
 const packageVersionExclusions: Record<string, string | Range> = {
-  // @angular/localize@9.x and earlier versions of @angular/localize@10.x prereleases do not have peer dependencies setup.
+  // @angular/localize@9.x and earlier versions as well as @angular/localize@10.0 prereleases do not have peer dependencies setup.
   '@angular/localize': '<10.0.0',
   // @angular/material@7.x versions have unbounded peer dependency ranges (>=7.0.0).
   '@angular/material': '7.x',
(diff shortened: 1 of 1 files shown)


packages/angular/cli/src/commands/add/cli.ts
```diff
@@ -48,10 +48,10 @@ interface AddCommandArgs extends SchematicsCommandArgs {
  * when attempting to find a compatible version for a package.
  * The key is a package name and the value is a SemVer range of versions to exclude.
  */
-const packageVersionExclusions: Record<string, string | undefined> = {
-  // @angular/localize@9.x versions do not have peer dependencies setup
-  '@angular/localize': '9.x',
-  // @angular/material@7.x versions have unbounded peer dependency ranges (>=7.0.0)
+const packageVersionExclusions: Record<string, string | Range> = {
+  // @angular/localize@9.x and earlier versions of @angular/localize@10.x prereleases do not have peer dependencies setup.
```
clydin, line 52:
```suggestion
  // @angular/localize@9.x and earlier versions as well as @angular/localize@10.0 prereleases do not have peer dependencies setup.
```

clydin (approved):

PR #24167, opened 2022-11-01
fix(@schematics/angular): guard schematics should include all guards (CanMatch)

atscott:
The `CanMatch` guard was added in v14.1 but not added to the list of possible interfaces to implement.
This commit adds `CanMatch` to the list of possible interfaces to implement. It has the same type imports as the CanLoad interface.

To reviewers: Should this be a feat instead?


alan-agius4 (approved):

alan-agius4:
`fix` is okay in this case.

PR #24170, opened 2022-11-02
Add functional router guards and resolvers

atscott:
### Commit 1
feat(@schematics/angular): Add schematics for generating functional router guards

Functional guards were introduced in the Angular router in v14.2.
This commit adds the ability to generate functional router guards by
specifying `--type` instead of `--implements`. These guards are also
accompanied by a test that includes a helper function for executing the
guard in the `TestBed` environment so that any `inject` calls in the
guard will work properly.

### Commit 2
feat(@schematics/angular): Add schematics for generating functional router resolvers


packages/schematics/angular/guard/schema.json
```diff
@@ -49,9 +49,13 @@
       "items": {
         "enum": ["CanActivate", "CanActivateChild", "CanDeactivate", "CanLoad", "CanMatch"],
         "type": "string"
```
alan-agius4, line 51: Should implements be deprecated?
If so add an option.
`"x-deprecated: "Use '--type' instead.........."`

packages/schematics/angular/guard/schema.json
```diff
@@ -49,9 +49,13 @@
...
-      },
-      "default": ["CanActivate"],
```
alan-agius4, line 53: This looks like a breaking change.

packages/schematics/angular/resolver/schema.json
```diff
@@ -25,6 +25,11 @@
       "description": "When true (the default), creates the new files at the top level of the current project.",
       "default": true
     },
+    "functional": {
+      "type": "boolean",
+      "description": "When true, creates the resolver as a `ResolveFn`.",
+      "default": true
```
alan-agius4, line 31: This also looks like a breaking change as there is a change in behaviour.

alan-agius4 (commented):
I am concerned that this is happing outside of a major version. As there are 2 changes in behaviour here which technically are breaking changes.

atscott:
> I am concerned that this is happing outside of a major version. As there are 2 changes in behaviour here which technically are breaking changes.

I don't know what the CLI team considers breaking, so you'll have to help determine what we're allowed to do here. I assumed, maybe incorrectly, that keeping all the options the same but changing the defaults would be fine.

packages/schematics/angular/guard/schema.json
atscott: Since this has a default but `implements` and `type` cannot be used together, how should this be handled? Should the logic just change to use `type` when both are provided instead of throwing an error?

packages/schematics/angular/resolver/schema.json
atscott: The changes to the defaults were done in separate commits. If those need to be considered breaking, I'm happy to drop those commits. That said, it will be a bit weird for a bit in the CLI that it will generate deprecated code by default.

packages/schematics/angular/guard/schema.json
atscott: @dgp1130 suggested implementing the default in the `index.ts`. I think that works, but unresolve this thread if not.

packages/schematics/angular/guard/files/__name@dasherize__.guard.spec.ts.template
```diff
@@ -1,3 +1,4 @@
+<% if (implements.length > 0) { %>
```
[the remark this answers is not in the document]
alan-agius4: What you an do is create two directories which holds the respective templates. 

Ex:
```
packages/schematics/angular/guard/implements-files/
packages/schematics/angular/guard/type-files/
```

Amend https://github.com/angular/angular-cli/blob/78482b30ede41cbddfbefd80428e4eddb945ea73/packages/schematics/angular/utility/generate-from-files.ts#L35 to take a path argument to provide a way to override https://github.com/angular/angular-cli/blob/78482b30ede41cbddfbefd80428e4eddb945ea73/packages/schematics/angular/utility/generate-from-files.ts#L50

packages/schematics/angular/resolver/schema.json
```diff
@@ -25,6 +25,11 @@
       "description": "When true (the default), creates the new files at the top level of the current project.",
       "default": true
     },
+    "functional": {
+      "type": "boolean",
+      "description": "When true, creates the resolver as a `ResolveFn`.",
```
alan-agius4, line 30: NIT: we are trying to avoid when true and when false. As this can be confusing when used as a CLI flag

```suggestion
      "description": "Creates the resolver as a `ResolveFn`.",
```

packages/schematics/angular/guard/schema.json
alan-agius4: Looks like this has not been done.

packages/schematics/angular/guard/index.ts
```diff
@@ -7,14 +7,35 @@
...
 import { Rule, SchematicsException } from '@angular-devkit/schematics';
+
 import { generateFromFiles } from '../utility/generate-from-files';
+
 import { Implement as GuardInterface, Schema as GuardOptions } from './schema';
 
 export default function (options: GuardOptions): Rule {
-  if (!options.implements) {
-    throw new SchematicsException('Option "implements" is required.');
+  if (!options.implements && !options.guardType) {
```
alan-agius4, line 16: If non is set we should fallback to `CanActivate` to avoid the breaking change. As with this change `implements` becomes as required option which previously was not.

packages/schematics/angular/guard/index.ts
```diff
@@ -7,14 +7,35 @@
...
+  if (options.implements && options.implements.length > 0 && options.guardType) {
+    throw new SchematicsException('Options "implements" and "guardType" cannot be used together.');
+  }
+
+  return options.guardType ? functionalGuardRule(options) : classBasedGuardRule(options);
+}
+
+function functionalGuardRule(options: GuardOptions): Rule {
+  if (!options.guardType) {
+    throw new SchematicsException('Option "implements" or "guardType" is required.');
```
alan-agius4, line 28: Copy/paste mistake?

packages/schematics/angular/guard/index.ts
```diff
@@ -7,14 +7,35 @@
...
   }
+  const guardType = options.guardType.replace(/^can/, 'Can') + 'Fn';
```
alan-agius4, line 30: Out of curiosity why not use set the value of the enum as `Can**Fn` directly?

packages/schematics/angular/guard/index.ts
```diff
@@ -7,14 +7,35 @@
...
 
+  return generateFromFiles(options, { guardType });
+}
+
+function classBasedGuardRule(options: GuardOptions): Rule {
+  if (!options.implements || options.implements.length === 0) {
+    options.implements = [GuardInterface.CanActivate];
```
alan-agius4, line 37: This is never triggered 

https://github.com/angular/angular-cli/pull/24170/files#r1013666129

packages/schematics/angular/guard/schema.json
```diff
@@ -50,8 +50,12 @@
         "enum": ["CanActivate", "CanActivateChild", "CanDeactivate", "CanLoad", "CanMatch"],
         "type": "string"
       },
-      "default": ["CanActivate"],
       "x-prompt": "Which interfaces would you like to implement?"
```
alan-agius4, line 53: I think the prompt in this case should be removed. Since otherwise there wouldn't be a way to use `guardType`.

packages/schematics/angular/guard/schema.json
atscott: I didn't deprecate it in this PR but it's in the one that updates the defaults. That way, this change is purely a feature addition rather than combining a feature, a deprecation, and an update to the default values.

packages/schematics/angular/resolver/schema.json
atscott: 👍 Sounds good!

packages/schematics/angular/guard/index.ts
atscott: I wanted to map the names to the `Route` properties because it's less typing and less to remember.

packages/schematics/angular/guard/index.ts
atscott: This change didn't update the schema for `implements`. I don't think this would be a breaking change here.

atscott: Oh, you're right. It did :)

packages/schematics/angular/guard/index.ts
fmalcher: My unsolicited 2 cents:
We discussed this for our book recently and differentiated between the *type of the guard* and the *TypeScript type of the function*. A "CanActivate guard" is the concept while `CanActivateFn` is the TS type for it (or even `CanActivate` for classes).

I think this distinction between concept and technical type is important. `guardType` describes the concept.

tl;dr; Agree with @atscott :smile:

packages/schematics/angular/guard/index.ts
atscott: Typescript isn't quite able to make this work. I've removed the helper functions.

packages/schematics/angular/guard/schema.json
atscott: I'm assuming there's not a way to make the prompt conditional? I feel like the lack of a prompt is a step down in UX.

packages/schematics/angular/guard/files/__name@dasherize__.guard.spec.ts.template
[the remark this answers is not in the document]
atscott: Done

packages/schematics/angular/BUILD.bazel
```diff
@@ -47,6 +47,10 @@ ts_library(
             # Also exclude templated files.
             "*/files/**/*.ts",
             "*/other-files/**/*.ts",
+            "*/implements_files/**/*",
```
alan-agius4, line 50: NIT: can we change this to dashes to match the other folder style?

packages/schematics/angular/guard/schema.json
alan-agius4: Correct. we don't support conditional prompts.

alan-agius4 (approved):
LGTM, one little nit.

packages/schematics/angular/guard/schema.json
atscott: So I tried this out and I don't think this is going to work. The experience now is that `ng g guard myGuard` will just generate a `canActivate` guard automatically. Most users will not know about the `implements` flag and will not know how to generate a guard of another type. We'll have to come up with another way to make this work.

atscott: Okay, I updated the approach to use the `--functional` flag and throw an error if there is more than 1 value for `implements` when using it. The names of the existing `implements` options are the interfaces, and while this isn't ideal, I don't think it's really an issue.

PTAL

PR #24182, opened 2022-11-03
Update defaults for guards and resolvers to be functional guards

atscott:
first two commits are part of https://github.com/angular/angular-cli/pull/24170. 

The two commits specific to this are the default updates, which may be considered a breaking change.

### DIFF
--- a/packages/schematics/angular/guard/schema.json
+++ b/packages/schematics/angular/guard/schema.json
@@ -50,12 +50,13 @@
         "enum": ["CanActivate", "CanActivateChild", "CanDeactivate", "CanLoad", "CanMatch"],
         "type": "string"
       },
-      "x-prompt": "Which interfaces would you like to implement?"
+      "x-deprecated": "Use '--guardType' instead."
     },
     "guardType": {
       "type": "string",
       "description": "Specifies type of guard to generate.",
-      "enum": ["canActivate", "canActivateChild", "canDeactivate", "canLoad", "canMatch"]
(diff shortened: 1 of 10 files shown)


packages/schematics/angular/guard/schema.json
```diff
@@ -50,8 +50,13 @@
...
       },
-      "default": ["CanActivate"],
-      "x-prompt": "Which interfaces would you like to implement?"
+      "x-deprecated": "Use '--guardType' instead."
+    },
+    "guardType": {
+      "type": "string",
+      "description": "Specifies type of guard to generate.",
+      "enum": ["canActivate", "canActivateChild", "canDeactivate", "canLoad", "canMatch"],
+      "x-prompt": "Which type of guard would you like to generate?"
```
alan-agius4, line 59: Should this have a default?

alan-agius4:
Hi @atscott, 

We did discuss this PR during our last triage meeting and while we do not guarantee that the generated code does not change between versions. This is too big of a change to be done outside of a major. 

We suggest a call out in the deprecation note so users can opt-in and use the non deprecated code template by default

```
ng config schematics.@schematics/angular:guard.functional true
ng config schematics.@schematics/angular:resolver.functional true
```

Feel free to close the PR or rebase it and make the necessary changes to eventually land in v16. I also have a separate PR to add an aliased for the `implements` flag https://github.com/angular/angular-cli/pull/24216

alan-agius4:
Closing in favor of https://github.com/angular/angular-cli/pull/24585

PR #24185, opened 2022-11-04
fix(@angular-devkit/build-angular): rebase Sass url() values when using esbuild-based builder

clydin:
When using the experimental esbuild-based browser application builder with Sass, paths contained in `url()` CSS functions will now be rebased to reflect their final location relative to the output of the Sass compiler. This allows the bundler to locate and process any resource files such as images or fonts used within the stylesheet.
The implementation of the rebasing functionality uses a series of custom Sass importers to transform the input stylesheets (from import or use rules) when they are loaded from the file system. However, for the Sass compiler to use the load functionality of an importer, the `canonicalize` function (used to resolve an import) of the same importer must return a resolved path to the requested stylesheet. To support this requirement, all stylesheet resolution has been implemented in three new importers: relative, module, and load path. An additional benefit of these resolvers is that they have been implemented to minimize the volume of filesystem calls and also cache directory entry information (currently per request) to further reduce file system access. The reduction in filesystem calls provides a small performance benefit for projects containing a large amount of Sass file imports.


alan-agius4 (approved):
LGTM

PR #24200, opened 2022-11-08
fix(@angular-devkit/build-angular): warn when components styles sourcemaps are not generated when styles optimization is enabled

### LINKED ISSUE #22834: CSS sourcemap files aren't produced
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 
I don't believe so, but I'm not for sure on that.

### Description

 
When building with `angular.json` => `{ sourcemaps: true, optimization: true }` (more specifically `optimization.styles.minify: true`), CSS sourcemap comments in `main.*.js` point to a non-existent `*.component.css.map` file:
```
/*sourceMappingURL=footer.component.css.map*/
```
The corresponding `footer.component.css.map` file is not created by the build.

Here is an example of an example app containing a FooterComponent built with four different `optimization` build settings in `angular.json` (but sourcemaps turned on in all four cases):

image

No `footer.component.css.map` file is ever created. When the app source code is viewed in Chrome Dev Tools, it tries and fails to load the non-existant CSS sourcemaps.

## 🔬 Minimal Reproduction

Build using `angular.json`:
```
{
  "sourcemaps": true,
  "optimization": true
}
```

## 🌍 Your Environment

<pre><code>
Angular CLI: 13.2.6
Node: 16.14.0
Package Manager: npm 8.3.1
OS: win32 x64

Angular: 13.2.6
... forms, language-service, material, platform-browser

@angular-devkit/architect       0.1302.6
@angular-devkit/build-angular   13.2.6
@angular-devkit/core            13.2.6
@angular-devkit/schematics      13.2.6
@schematics/angular             13.2.6
rxjs                            7.5.5
typescript                      4.5.5
</code></pre>

alan-agius4:
With this change we add a warning to inform the users that sourcemaps are not generated when both styles sourcemaps and optimization are enabled. This is because component style sourcemaps are inline which would drastically increase the bundle size.

Closes #22834


packages/angular_devkit/build_angular/src/webpack/configs/styles.ts
```diff
@@ -148,6 +148,12 @@ export function getStylesConfig(wco: WebpackConfigOptions): Configuration {
     !buildOptions.sourceMap.hidden
   );
 
+  if (!componentsSourceMap && cssSourceMap) {
+    logger.warn(
+      'Components styles sourcemaps are not generated when styles optimization is enabled.',
```
clydin, line 153: Based on the logic for `componentsSourceMap`, enabling hidden source maps will also trigger this behavior. Maybe add that to the warning message also?

clydin (approved):

clydin (approved):

PR #24202, opened 2022-11-08
build: remove unused dependency `minimatch` from `@angular-devkit/build-angular`

clydin:
`minimatch` was unused within the `@angular-devkit/build-angular` package. It is still used in a repository level development script and has been kept in the root `package.json`.


alan-agius4 (approved):
Good catch

PR #24209, opened 2022-11-09
fix(@angular/cli): respect registry in RC when running update through yarn

alan-agius4:
This commit fixes an issue where when `ng update` was ran using `yarn` (`yarn ng update`) the registry was always being overridden to `https://registry.yarnpkg.com`.

This is because yarn will set the `npm_config_registry` env variable to `https://registry.yarnpkg.com` even when an RC file is present with a different repository.

reported by @JoostK and @SanderElias over slack.

### DIFF
--- a/packages/angular/cli/src/utilities/package-metadata.ts
+++ b/packages/angular/cli/src/utilities/package-metadata.ts
@@ -139,6 +139,18 @@ function readOptions(
       continue;
     }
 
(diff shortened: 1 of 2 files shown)


SanderElias:
@alan-agius4 I'm glad you found it!

clydin (approved):

PR #24260, opened 2022-11-17
ci: disable nightly ci for version 12

alan-agius4:
This version is no longer in LTS as such we no longer need to test it.


alan-agius4:
Please ignore failure since this branch is no longer going to be released.

PR #24266, opened 2022-11-17
perf(@angular-devkit/build-angular): use worker pool for JavaScript transforms in esbuild builder

clydin:
When using the experimental esbuild-based browser application builder, the JavaScript transformation steps of the build process will now be performed within a worker pool to allow for the steps to be executed in parallel when possible. This also moves the steps off of the main thread which provides more time for the build orchestration and esbuild integration code to execute.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/javascript-transformer.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/javascript-transformer.ts
@@ -1,0 +1,91 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import Piscina from 'piscina';
+
+/**
+ * Transformation options that should apply to all transformed files and data.
+ */
+export interface JavaScriptTransformerOptions {
+  sourcemap: boolean;
+  thirdPartySourcemaps?: boolean;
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/javascript-transformer.ts
```diff
@@ -0,0 +1,91 @@
...
+   * @returns A promise that resolves to a UTF-8 encoded Uint8Array containing the result.
+   */
+  async transformData(filename: string, data: string, skipLinker: boolean): Promise<Uint8Array> {
+    // Perform a quick test to determine if the data needs any transformations.
+    // This allows directly returning the data without the worker communication overhead.
+    let forceAsyncTransformation;
+    if (skipLinker && !this.options.advancedOptimizations) {
+      // If the linker is being skipped and no optimizations are needed, only async transformation is left.
+      // This checks for async generator functions. All other async transformation is handled by esbuild.
+      forceAsyncTransformation = data.includes('async') && /async\s+function\s*\*/.test(data);
```
alan-agius4, line 67: This will not catch async arrow functions which still need to be downlevelled.

clydin: Arrow functions cannot be async generators so we can avoid having to test for them.
ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function*#syntax

alan-agius4 (approved):

PR #24357, opened 2022-12-01
fix(@angular-devkit/build-angular): downlevel class properties when targeting Safari <=v15

### LINKED ISSUE #24355: Angular apps not working in Safari v15 due to `ReferenceError`

Don't known / other

### Is this a regression?

No

### Description

When we open angular.io website in safari browser in mac then it giving blank page and if we do refresh that page than also it's giving blank page and when we open inspect element from developer open in safari browser and do refresh that page then it open perfectly so it's an issue I have provided screenshot.

- This when I open angular.io link in Safari browser in Mac.
Screenshot 2022-11-30 at 6 37 28 PM

- This when I open inspect element of safari browser and still website is not working.
Screenshot 2022-11-30 at 6 39 15 PM

- And at last this after opening Inspect element I refresh the page then website is working properly.
Screenshot 2022-11-30 at 6 39 27 PM

https://angular.io

_No response_

_No response_

### Anything else?

_No response_

devversion:
The Angular compiler is dependent on static fields being attached to
user-defined classes. e.g. `static ecmp = defineComponent`.

These static fields sometimes rely on variables from outside of the
class. e.g. the Angular compiler generates constants for content
projection that are then accessed in the static field initializer.

Surprisingly such access to these variables may break in Safari <=v15
when a page is loaded without devtools open. The bug (already solved in
v16 of Safari)- is very subtle, hard to re-reproduce but basically
variable scope tracking is broken. This bug is triggered by additional
parenthesis in the initializer expression. See:
https://bugs.webkit.org/show_bug.cgi?id=236843.

The TypeScript compiler may generate such additional parenthesis when
it tries to adjust the `this` context when invoking methods, such as for
defining animations in the `ecmp` definition.

More details can be found here:
https://github.com/angular/angular-cli/issues/24355#issuecomment-1333477033

To ensure Angular applications are not subject to this bug when
targeting Safari <=v15. v15 Safari, both for iOS and Mac is still part of
the default CLI browserslist with `last 2 Safari majors` (at time of
writing).

Note that it is important that the Babel plugin properly handles the
downleveling of static block-defined members. TypeScript will transform
static fields, like `static ecmp` into `static { this.ecmp = X }` when
`useDefineForClassFields = false` (which is the case for CLI apps). The
class properties plugin from Babel seems to handle this in an acceptable
way. Unlike actual static fields, Babel will not use helpers like
`defineProperty` for such extracted static blocks though. e.g.

See repro: https://gist.github.com/devversion/dec0dea26e348c509921bf62079b60be

```js
class Test {
  x = true;

  static b = true;
  static {
    this.a = true;
  }
}

// into

class X {
  constructor() {
    _defineProperty(this, "x", true);
  }
}
_defineProperty(X, "b", true);
X.a = true;
```

note that in practice TypeScript with `useDefineForClassFields = false`
will put non-static members into the constructor as normal assignments
regardless- so there would be no change by the Babel plugin.

Fixes #24355

### DIFF
--- a/packages/angular_devkit/build_angular/src/babel/presets/application.ts
+++ b/packages/angular_devkit/build_angular/src/babel/presets/application.ts
@@ -14,8 +14,25 @@ import type {
   makeLocalePlugin,
 } from '@angular/localize/tools';
 import { strict as assert } from 'assert';
+import browserslist from 'browserslist';
 import * as fs from 'fs';
 import * as path from 'path';
+
+/**
+ * List of browsers which are affected by a WebKit bug where class field
+ * initializers might have incorrect variable scopes.
+ *
+ * See: https://github.com/angular/angular-cli/issues/24355#issuecomment-1333477033
+ * See: https://github.com/WebKit/WebKit/commit/e8788a34b3d5f5b4edd7ff6450b80936bff396f2
+ */
+const safariClassFieldScopeBugBrowsers = new Set(
+  browserslist([
+    // Safari <15 is technically not supported via https://angular.io/guide/browser-support,
+    // but we apply the workaround if forcibly selected.
+    'Safari <=15',
+    'iOS <=15',
+  ]),
+);
 
 export type DiagnosticReporter = (type: 'error' | 'warning' | 'info', message: string) => void;
 
@@ -45,7 +62,6 @@ export interface ApplicationPresetOptions {
     linkerPluginCreator: typeof import('@angular/compiler-cli/linker/babel').createEs2015LinkerPlugin;
   };
 
-  forcePresetEnv?: boolean;
   forceAsyncTransformation?: boolean;
   instrumentCode?: {
     includedBasePath: string;
@@ -171,13 +187,26 @@ export default function (api: unknown, options: ApplicationPresetOptions) {
     );
   }
 
-  if (options.forcePresetEnv) {
+  // Applications code ES version can be controlled using TypeScript's `target` option.
+  // However, this doesn't effect libraries and hence we use preset-env to downlevel ES features
+  // based on the supported browsers in browserslist.
+  if (options.supportedBrowsers) {
+    const includePlugins: string[] = [];
+
+    // If a Safari browser affected by the class field scope bug is selected, we
+    // downlevel class properties by ensuring the class properties Babel plugin
+    // is always included- regardless of the preset-env targets.
+    if (options.supportedBrowsers.some((b) => safariClassFieldScopeBugBrowsers.has(b))) {
+      includePlugins.push('@babel/plugin-proposal-class-properties');
+    }
+
     presets.push([
       require('@babel/preset-env').default,
       {
         bugfixes: true,
         modules: false,
         targets: options.supportedBrowsers,
+        include: includePlugins,
         exclude: ['transform-typeof-symbol'],
       },
     ]);
--- a/tests/legacy-cli/e2e/tests/misc/safari-15-class-properties.ts
+++ b/tests/legacy-cli/e2e/tests/misc/safari-15-class-properties.ts
@@ -1,0 +1,56 @@
+import { expectFileToExist, readFile, writeFile } from '../../utils/fs';
+import { ng } from '../../utils/process';
+import { updateJsonFile } from '../../utils/project';
+
+const unexpectedStaticFieldErrorMessage =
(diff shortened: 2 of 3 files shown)


packages/angular_devkit/build_angular/src/babel/presets/application.ts
```diff
@@ -172,12 +182,23 @@ export default function (api: unknown, options: ApplicationPresetOptions) {
   }
 
   if (options.forcePresetEnv) {
+    const selectedBrowsers = browserslist(options.supportedBrowsers);
```
alan-agius4, line 185: SupporteredBrowsers is already a parsed browserslist query.

packages/angular_devkit/build_angular/src/babel/presets/application.ts
```diff
@@ -16,6 +16,16 @@ import type {
...
+import browserslist from 'browserslist';
+
+/**
+ * List of browsers which are affected by a WebKit bug where class field
+ * initializers might have incorrect variable scopes.
+ *
+ * See: https://github.com/angular/angular-cli/issues/24355#issuecomment-1333477033
+ * See: https://github.com/WebKit/WebKit/commit/e8788a34b3d5f5b4edd7ff6450b80936bff396f2
+ */
+const safariClassFieldScopeBugBrowsers = browserslist(['Safari <15', 'iOS <15']);
```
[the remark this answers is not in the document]
devversion: +1 to a set. I think keeping as a constant makes it clear that this is not something config-dependent

packages/angular_devkit/build_angular/src/babel/presets/application.ts
```diff
@@ -14,9 +14,26 @@ import type {
...
+ *
+ * See: https://github.com/angular/angular-cli/issues/24355#issuecomment-1333477033
+ * See: https://github.com/WebKit/WebKit/commit/e8788a34b3d5f5b4edd7ff6450b80936bff396f2
+ */
+const safariClassFieldScopeBugBrowsers = new Set(
+  browserslist([
+    // Safari <15 is technically not supported via https://angular.io/guide/browser-support,
+    // but we apply the workaround if forcibly selected.
+    'Safari <=15',
+    'iOS <=15',
```
devversion, line 33: FYI: Discussed with @clydin that we should just apply for v14 too. The range can be that loose just to be safe. older versions will have the plugin enabled anyway.

packages/angular_devkit/build_angular/src/babel/presets/application.ts
```diff
@@ -45,7 +62,6 @@ export interface ApplicationPresetOptions {
     linkerPluginCreator: typeof import('@angular/compiler-cli/linker/babel').createEs2015LinkerPlugin;
   };
 
-  forcePresetEnv?: boolean;
```
devversion, line 48: Note: I removed this so that we just rely on `supportedBrowsers` for better type safety (otherwise it may be undefined type-wise). I don't know what the reason for this extra option is.

tests/legacy-cli/e2e/tests/misc/safari-15-class-properties.ts
```diff
@@ -0,0 +1,43 @@
...
+      optimization: false,
+      outputHashing: 'none',
+    };
+  });
+
+  // Matches two types of static fields that indicate that the Safari bug
+  // may still occur. With the workaround this should not appear in bundles.
+  //   - static { this.ecmp = bla }
+  //   - static #_ = this.ecmp = bla
+  const staticIndicatorRegex = /static\s+(\{|#[_\d]+\s+=)/;
```
devversion, line 20: This is a sanity test. Obviously it's not validating using AST, nor validating Safari. Should be helpful regardless though.

packages/angular_devkit/build_angular/src/babel/presets/application.ts
clydin: It's leftover from when it needed to downlevel to ES5 based on the typescript target.  It doesn't really serve a purpose now that everything is based on the supported browsers.

clydin (approved):
LGTM.
Just two suggestions:
* Can you also add an additional test case for an explicit `safari <=15`?
* Also can you add a comment that the default test case (no explicit browserslist) will need to be remove once Safari 17 is released to prevent this from failing?

devversion:
@clydin Updated.

PR #24363, opened 2022-12-01
fix(@angular-devkit/build-angular): downlevel class fields with Safari <= v15 for esbuild

clydin:
To provide a workaround for a Safari bug involving class fields and variable scoping,
the esbuild-based browser application builder will now downlevel class fields if Safari
(desktop or iOS) v15.x or earlier is within the target browsers for an application.
This is an esbuild variant of the fix for the Webpack-based builder. For more details
regarding the issue, please see: https://github.com/angular/angular-cli/pull/24357

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
@@ -253,19 +253,7 @@ function createCodeBundleOptions(
     entryNames: outputNames.bundles,
     assetNames: outputNames.media,
     target,
-    supported: {
-      // Native async/await is not supported with Zone.js. Disabling support here will cause
-      // esbuild to downlevel async/await and for await...of to a Zone.js supported form. However, esbuild
-      // does not currently support downleveling async generators. Instead babel is used within the JS/TS
-      // loader to perform the downlevel transformation.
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
```diff
@@ -314,6 +302,56 @@ function createCodeBundleOptions(
...
+    // is identical since it would be set to false by esbuild if present as a target.
+    if (majorVersion === 14 || majorVersion === 15) {
+      safariStaticFieldScopeBug = true;
+      break;
+    }
+  }
+  // If static field support cannot be used set to false; otherwise leave undefined to allow
+  // esbuild to use `target` to determine support.
+  if (safariStaticFieldScopeBug) {
+    supported['class-static-field'] = false;
```
devversion, line 349: do you want to downlevel normal class members too? As we discussed, likely shouldn't be hit at all since TS moves them into constructor assignments with `useDefineForClaasFields = false`, but IIUC the bug also surfaces for normal class members?

clydin: We might as well be comprehensive here.  `useDefineForClassFields` as true (or the default for es2022+) is the ideal once libraries fully support.

alan-agius4 (approved):

PR #24407, opened 2022-12-09
Allow extending generated webpack config of karma builder

DanielRose:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The `@angular-devkit/build-angular:karma` builder generates a webpack configuration, and runs webpack (and karma) with that configuration. Currently that configuration cannot be changed by the user, even if the user provides a webpack configuration in their karma config.

## What is the new behavior?

The user's webpack configuration is merged together with the generated configuration, using `webpack-merge`.

- [ ] Yes
- [x] No

## Other information

This was previously possible, but the commit 04f5dfe5e2fc97bcf0641f1d2f8edfc59dae57b2 replaced  `Object.assign(webpackConfig, config.webpack)` with a spread operator, which means that `webpackConfig` was not changed anymore.

Since the generated configuration uses functions (and the user's configuration could as well), `webpack-merge` is used to merge the configurations. Since `webpack-merge` will throw an error when passing undefined values, the code first checks that a value to-be-merged exists.

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/plugins/karma/karma.ts
@@ -12,6 +12,7 @@ import * as http from 'http';
 import * as path from 'path';
 import webpack from 'webpack';
 import webpackDevMiddleware from 'webpack-dev-middleware';
+import { merge as webpackMerge } from 'webpack-merge';
 
 import { statsErrorsToString } from '../../utils/stats';
 import { createConsoleLogger } from '@angular-devkit/core/node';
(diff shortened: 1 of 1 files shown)


DanielRose:
A custom webpack configuration is needed to allow tests touching monaco-editor, due to https://github.com/microsoft/monaco-editor/issues/886

DanielRose:
I'm not sure if this should be considered a new feature or a bugfix for an (unintended?) change in 04f5dfe5e2fc97bcf0641f1d2f8edfc59dae57b2

I couldn't find the documentation to the karma builder. If there is one, I could also add some documentation to describe this.

alan-agius4:
Hi @DanielRose, I am surprised that this worked before. It was definitely was unintended.

The current way to extend the Webpack configuration it by using the transform API through a custom builder https://github.com/angular/angular-cli/blob/afe9feaa45913cbebe7f22c678d693d96f38584a/packages/angular_devkit/build_angular/src/builders/karma/index.ts#L34 

It important to point our that by extending the Webpack confirmation you are opt-ing out of the supported path. 

While we allow extending the CLI using the a of mentioned API, we don’t not provide support for any issues caused by extending the Webpack configure. Webpack specific features are not supported by the Angular CLI and their usage might break in any version without warning.

DanielRose:
So I should build my own Angular CLI builder, which calls the karma builder internally and adds a webpackConfigurationTransformer?

alan-agius4:
@DanielRose, no need to build your own builder. There are already a number of builders which have been created by the community. Here are a few:
- https://www.npmjs.com/package/@angular-builders/custom-webpack
- https://github.com/manfredsteyer/ngx-build-plus

PR #24458, opened 2022-12-20
feat(@schematics/angular): add configuration files generation schematic

### LINKED ISSUE #24294: Expose the built-in karma config.
### Command

test

### Description

Angular 15 hides the cli config files. This is admirable, but this means that tools depending on this config cannot work. For example, StrykerJS needs this configuration. See https://github.com/stryker-mutator/stryker-js/issues/3871

I would love for the cli to expose this built-in config, similar to how the [vue cli exposes webpack config as a file](https://cli.vuejs.org/guide/webpack.html#using-resolved-config-as-a-file).

An added benefit is that users of Angular 15 can expand their custom karma config upon it instead of needing to copy it entirely and make adjustments.

_No response_

alan-agius4:
This commits add a schematic to generate Karma and Browserlist files which since version 15 are no longer generated by default. This schematic should be used to generate these files when further customisation is needed.

Usage
```
ng generate config karma
ng generate config browserlist
```

Closes #24294

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/karma/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/karma/index.ts
@@ -96,7 +96,7 @@ export function execute(
 
       const karmaOptions: KarmaConfigOptions = options.karmaConfig
         ? {}
-        : getBuiltInKarmaConfig(karma, context.workspaceRoot, projectName);
+        : getBuiltInKarmaConfig(context.workspaceRoot, projectName);
 
       karmaOptions.singleRun = singleRun;
 
@@ -186,7 +186,6 @@ export function execute(
 }
 
 function getBuiltInKarmaConfig(
-  karma: typeof import('karma'),
   workspaceRoot: string,
   projectName: string,
 ): ConfigOptions & Record<string, unknown> {
@@ -197,6 +196,7 @@ function getBuiltInKarmaConfig(
 
   const workspaceRootRequire = createRequire(workspaceRoot + '/');
 
+  // Any changes to the config here need to be synced to: packages/schematics/angular/config/files/karma.conf.js.template
   return {
     basePath: '',
     frameworks: ['jasmine', '@angular-devkit/build-angular'],
--- a/packages/angular_devkit/build_angular/src/builders/karma/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/karma/index.ts
@@ -219,10 +219,6 @@ function getBuiltInKarmaConfig(
       reporters: [{ type: 'html' }, { type: 'text-summary' }],
     },
     reporters: ['progress', 'kjhtml'],
-    port: 9876,
-    colors: true,
-    logLevel: karma.constants.LOG_INFO,
-    autoWatch: true,
     browsers: ['Chrome'],
     restartOnFileChange: true,
   };
--- a/packages/schematics/angular/config/files/karma.conf.js.template
+++ b/packages/schematics/angular/config/files/karma.conf.js.template
@@ -1,0 +1,39 @@
+// Karma configuration file, see link for more information
+// https://karma-runner.github.io/1.0/config/configuration-file.html
+
+module.exports = function (config) {
+  config.set({
+    basePath: '',
+    frameworks: ['jasmine', '@angular-devkit/build-angular'],
+    plugins: [
+      require('karma-jasmine'),
+      require('karma-chrome-launcher'),
+      require('karma-jasmine-html-reporter'),
+      require('karma-coverage'),
+      require('@angular-devkit/build-angular/plugins/karma')
+    ],
+    client: {
+      jasmine: {
+        // you can add configuration options for Jasmine here
+        // the possible options are listed at https://jasmine.github.io/api/edge/Configuration.html
+        // for example, you can disable the random execution with `random: false`
+        // or set a specific seed with `seed: 4321`
+      },
+      clearContext: false // leave Jasmine Spec Runner output visible in browser
+    },
+    jasmineHtmlReporter: {
+      suppressAll: true // removes the duplicated traces
+    },
+    coverageReporter: {
+      dir: require('path').join(__dirname, '<%= relativePathToWorkspaceRoot %>/coverage/<%= folderName %>'),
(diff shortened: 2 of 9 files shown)


packages/schematics/angular/config/files/karma.conf.js.template
```diff
@@ -0,0 +1,39 @@
+// Karma configuration file, see link for more information
+// https://karma-runner.github.io/1.0/config/configuration-file.html
+
+module.exports = function (config) {
```
nicojs, line 4: This seems to be a duplicate of `/build_angular/src/builders/karma/index.ts`. How are these kept in sync?

nicojs:
As far as I can see, this doesn't close #24294. Tools like the IntelliJ karma plugin and StrykerJS are still unable to auto-discover the config. 🤷‍♀️

packages/schematics/angular/config/files/karma.conf.js.template
alan-agius4: This config file is rarely updated. 

The file however will need to be kept in sync manually since `@schematics/angular` does not have a dependency on `@angular-devkit/build-angular` and vice-versa.

alan-agius4:
> As far as I can see, this doesn't close #24294. Tools like the IntelliJ karma plugin and StrykerJS are still unable to auto-discover the config. 🤷‍♀️

How would these plugins auto discover the config if present in node_modules? You would still need to create a `karma.conf.js` in the project.

nicojs:
With my change, a tool could use `require('@angular-devkit/build-angular/src/builders/karma/karma.conf.js')` to import the karma config and use that to configure karma from their own config file.

packages/angular_devkit/build_angular/src/builders/karma/index.ts
```diff
@@ -219,10 +219,6 @@ function getBuiltInKarmaConfig(
       reporters: [{ type: 'html' }, { type: 'text-summary' }],
     },
     reporters: ['progress', 'kjhtml'],
-    port: 9876,
-    colors: true,
-    logLevel: karma.constants.LOG_INFO,
-    autoWatch: true,
```
clydin, line 225: Are these options not needed anymore?

packages/schematics/angular/config/files/karma.conf.js.template
clydin: The duplication is unfortunate but needed to avoid coupling the two packages currently.
Can we add a comment to the inline config in the builder about updating this template if any changes are made?

packages/schematics/angular/config/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    const project = workspace.projects.get(options.project);
+    if (!project) {
+      throw new SchematicsException(`Project name "${options.project}" doesn't not exist.`);
+    }
+
+    const testTarget = project.targets.get('test');
+    if (!testTarget) {
+      throw new SchematicsException(
+        `No "test" target found for project "${options.project}".` +
+          ' A "test" target is required to generate environment files.',
```
clydin, line 65:
```suggestion
          ' A "test" target is required to generate a karma configuration.',
```

packages/schematics/angular/config/schema.json
```diff
@@ -0,0 +1,28 @@
...
+      }
+    },
+    "type": {
+      "type": "string",
+      "description": "Specifies which type of configuration file to create.",
+      "enum": ["karma", "browserslist"],
+      "items": {
+        "enum": ["karma", "browserslist"],
+        "type": "string"
+      },
```
clydin, line 23: `items` shouldn't be needed here. It should only apply to array types.

clydin (approved):

packages/schematics/angular/config/schema.json
```diff
@@ -0,0 +1,28 @@
...
+  "properties": {
+    "project": {
+      "type": "string",
+      "description": "The name of the project.",
+      "$default": {
+        "$source": "projectName"
+      }
+    },
+    "type": {
+      "type": "string",
```
clydin, line 17: nit: Maybe also add the following to allow specifying the config type directly:
```
      "$default": {
        "$source": "argv",
        "index": 0
      },
```
Example: `ng g config karma`

alan-agius4: Agreed.

packages/angular_devkit/build_angular/src/builders/karma/index.ts
alan-agius4: No as these are the defaults of Karma.

packages/schematics/angular/config/files/karma.conf.js.template
alan-agius4: Added

alan-agius4:
> Could we at least export a factory method for configuring karma?

You can use the builder API to transform the built in option https://github.com/angular/angular-cli/blob/fbfbca59a4e1cb0ead2f8069ab6a9f678c8e03b7/packages/angular_devkit/build_angular/src/builders/karma/index.ts#L78

A karma config on its own without the builder is not that useful as karma requires the output of the webpack builder which is done when running `ng test` in addition some of the karma plugins requires to run with an Angular CLI context.

nicojs:
Ah, I understand the confusion. Please let me elaborate.

StrykerJS (and the IntelliJ plugin) discover whether the current project is an Angular or a regular Karma project. If it is an Angular project, **it uses `ng test --karma-config=/path/to/stryker/karma.conf.js` to start karma**. This makes sure that webpack and such are correctly started. 

The karma config file ("/path/to/stryker/karma.conf.js") provided is a custom karma config file crafted in StrykerJS. From that config file, we load your local karma.conf.js. Since Angular  15, this file is no longer generated; thus, we need a new way to load the angular karma config. I was hoping something like this: https://github.com/angular/angular-cli/pull/24458#issuecomment-1359490555


This is the part of the StrykerJS code that uses `ng test`: 
https://github.com/stryker-mutator/stryker-js/blob/8bf885f490de6625466a011dc79cf04d53aeb698/packages/karma-runner/src/starters/angular-starter.ts#L30-L51

PR #24460, opened 2022-12-20
feat(@angular-devkit/build-angular): implement stats-json option for esbuild builder

clydin:
When using the experimental esbuild-based browser application builder, the `--stats-json` option can now be used to create an esbuild metafile named `stats.json` in the output directory of the application. The metafile contents will contain information about the application JavaScript, global stylesheets, and the component stylesheets used by the Angular compiler. While the `--stats-json` option controls the output of the file onto the filesystem, the metafile data is internally always created to support the future integration of the bundle budget and console build stat output capabilities.
The metafile format follows the structure of the esbuild metafile format. Information regarding the file format can be found here: https://esbuild.github.io/api/#metafile

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
@@ -7,7 +7,7 @@
  */
 
 import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
-import type { BuildInvalidate, BuildOptions, OutputFile } from 'esbuild';
+import type { BuildInvalidate, BuildOptions, Metafile, OutputFile } from 'esbuild';
 import assert from 'node:assert';
 import * as fs from 'node:fs/promises';
 import * as path from 'node:path';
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
```diff
@@ -258,6 +268,7 @@ function createCodeBundleOptions(
     mainFields: ['es2020', 'browser', 'module', 'main'],
     conditions: ['es2020', 'es2015', 'module'],
     resolveExtensions: ['.ts', '.tsx', '.mjs', '.js'],
+    metafile: true,
```
alan-agius4, line 271: Should this be false, when stats are disabled?

clydin: The info will be used for the console stat output as well as bundle budgets so we'll always need to generate the data internally.  It is only written to disk if the stats option is enabled though.

alan-agius4 (approved):

PR #24483, opened 2023-01-04
build: use bazel to perform release builds

clydin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

When performing a release via the dev-infra `ng-dev` tooling, the release builds for the packages that will be published are now performed using bazel. Prior to this, the release builds were performed using a custom build script that programmatically invoked TypeScript APIs. The Bazel build and discovery process for the releasable packages is performed by a script that is based on the scripts from components and framework repositories. Several small modifications were performed to match the behavior and structure of the cli repository:
* Use of `packages` as the source root in the bazel query
* Use of `pkg_npm` rule in the bazel query
* Partial transition to native Node.js `fs` APIs instead of `shelljs`
* Directory creation per package when copying output (supports multiple package scopes)
* Copying of archives (tgz) for each package

The snapshot and local build capabilities are not modified as part of this change but will be merged in a followup as part of a larger transition to use bazel throughout the package build process.

- [ ] Yes
- [X] No

### DIFF
--- a/scripts/build-packages-dist.mts
+++ b/scripts/build-packages-dist.mts
@@ -1,0 +1,168 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * Script that builds the release output of all packages which have the "release-package
+ * Bazel tag set. The script builds all those packages and copies the release output to the
+ * distribution folder within the project.
+ */
+
+import { BuiltPackage } from '@angular/ng-dev';
+import { execSync } from 'node:child_process';
+import { chmodSync, copyFileSync, mkdirSync, rmSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import sh from 'shelljs';
+
+/** Name of the Bazel tag that will be used to find release package targets. */
+const releaseTargetTag = 'release-package';
+
+/** Path to the project directory. */
+const projectDir = join(dirname(fileURLToPath(import.meta.url)), '../');
+
+/** Command that runs Bazel. */
+const bazelCmd = process.env.BAZEL || `yarn -s bazel`;
+
+/** Command that queries Bazel for all release package targets. */
+const queryPackagesCmd =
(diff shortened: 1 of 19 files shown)


scripts/build-packages-dist.mts
```diff
@@ -0,0 +1,155 @@
...
+  console.log('######################################');
+
+  // List of targets to build. e.g. "packages/angular/cli:npm_package"
+  const targets = exec(queryPackagesCmd, true)
+    .split(/\r?\n/)
+    .map((target) => target + '_archive');
+  const packageNames = getPackageNamesOfTargets(targets);
+  const bazelBinPath = exec(`${bazelCmd} info bazel-bin`, true);
+  const getBazelOutputPath = (pkgName: string) =>
+    join(bazelBinPath, 'packages', pkgName, 'npm_package');
```
devversion, line 68: I think this would just directly point to the archive right? It's a little odd because you are only requesting the archives, but also copy the `npm_package` tree artifact/folder too. Maybe both targets should be explicitly requested and we have two functions like `getOutputNpmPackagePath()` and `getOutputNpmPackageArchivePath()`

scripts/build-packages-dist.mts
```diff
@@ -0,0 +1,155 @@
...
+  });
+}
+
+/**
+ * Gets the package names of the specified Bazel targets.
+ * e.g. //packages/angular/cli:npm_package = angular/cli
+ */
+function getPackageNamesOfTargets(targets: string[]): string[] {
+  return targets.map((targetName) => {
+    const matches = targetName.match(/\/\/packages\/(.*):npm_package/);
```
devversion, line 128: I believe this just works currently because `npm_package_archive` matches too. May want to clarify the comment above, or be explicit `:npm_package(_archive)?$`

devversion (approved):
LGTM overall. Nice!

scripts/build-packages-dist.mts
clydin: Updated to build both targets and added separate variables for each output type (directory/archive).
I'd like to move towards just using the generated archives in the future to reduce the amount of files/directories that need to be copied and managed.

devversion: Yup, agreed!

PR #24487, opened 2023-01-05
ci: use snapshot deploy secret directly

devversion:
Instead of encoding the GitHub snapshot robot access token in a file that needs to be decoded by an ambiguously named secret called `KEY`, the token can be directly stored in the secrets of CircleCI given easier maintenance and the same risk.

The leakage of the secret still means the GitHub token file could be decoded. We are switching to a similar model as in the components repo, which also simplifies key rotations etc.


josephperrott (approved):
LGTM

PR #24522, opened 2023-01-11
feat(@angular-devkit/build-angular): support CommonJS dependency checking in esbuild

clydin:
When using the experimental esbuild-based browser application builder, input files
for the build will now be checked to determine if they are non-ESM modules. This
behavior is comparable to the existing behavior within the default Webpack-based browser.
Warnings will now be issued for any non-ESM modules (for example, CommonJS or UMD) when
script optimizations are enabled (typically production builds). ESM files can be tree-
shaken and otherwise optimized in ways that CommonJS files cannot which allows for
more optimized and smaller output bundle files.
If any allowed dependencies are provided via the `allowedCommonJsDependencies`
option, both the direct import and any deep imports of the dependency will be ignored
during the checks and no diagnostic will be generated for the dependency.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/commonjs-checker.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/commonjs-checker.ts
@@ -1,0 +1,138 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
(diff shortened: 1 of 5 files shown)


alan-agius4 (approved):
LGTM.

It's impressive how simply the implementation is, compared to the Webpack one.

PR #24527, opened 2023-01-11
refactor: only collect Node version if it's a standard version string

dgp1130:
Any other kinds of Node version are collected as "other" to avoid pulling in an unbounded user input.

Issue Number: http://b/264671793

### DIFF
--- a/packages/angular/cli/src/analytics/analytics-collector.ts
+++ b/packages/angular/cli/src/analytics/analytics-collector.ts
@@ -10,6 +10,7 @@ import { randomUUID } from 'crypto';
 import * as https from 'https';
 import * as os from 'os';
 import * as querystring from 'querystring';
+import * as semver from 'semver';
 import type { CommandContext } from '../command-builder/command-module';
 import { ngDebug } from '../utilities/environment-options';
 import { assertIsError } from '../utilities/error';
@@ -52,8 +53,7 @@ export class AnalyticsCollector {
 
     this.requestParameterStringified = querystring.stringify(requestParameters);
 
-    // Remove the `v` at the beginning.
-    const nodeVersion = process.version.substring(1);
+    const parsedVersion = semver.parse(process.version);
     const packageManagerVersion = context.packageManager.version;
 
     this.userParameters = {
@@ -62,8 +62,10 @@ export class AnalyticsCollector {
       [UserCustomDimension.OsArchitecture]: os.arch(),
       // While User ID is being collected by GA, this is not visible in reports/for filtering.
       [UserCustomDimension.UserId]: userId,
-      [UserCustomDimension.NodeVersion]: nodeVersion,
-      [UserCustomDimension.NodeMajorVersion]: +nodeVersion.split('.', 1)[0],
+      [UserCustomDimension.NodeVersion]: parsedVersion
+        ? `${parsedVersion.major}.${parsedVersion.minor}.${parsedVersion.patch}`
+        : 'other',
+      [UserCustomDimension.NodeMajorVersion]: parsedVersion?.major,
       [UserCustomDimension.PackageManager]: context.packageManager.name,
       [UserCustomDimension.PackageManagerVersion]: packageManagerVersion,
       [UserCustomDimension.PackageManagerMajorVersion]: packageManagerVersion


packages/angular/cli/src/analytics/analytics-collector.ts
```diff
@@ -52,8 +52,7 @@ export class AnalyticsCollector {
 
     this.requestParameterStringified = querystring.stringify(requestParameters);
 
-    // Remove the `v` at the beginning.
-    const nodeVersion = process.version.substring(1);
+    const nodeVersion = parseVersion(process.version);
```
alan-agius4, line 55:
```suggestion
    const nodeVersion = semver.clean(process.version);
```

packages/angular/cli/src/analytics/analytics-collector.ts
```diff
@@ -62,8 +61,8 @@ export class AnalyticsCollector {
       [UserCustomDimension.OsArchitecture]: os.arch(),
       // While User ID is being collected by GA, this is not visible in reports/for filtering.
       [UserCustomDimension.UserId]: userId,
-      [UserCustomDimension.NodeVersion]: nodeVersion,
-      [UserCustomDimension.NodeMajorVersion]: +nodeVersion.split('.', 1)[0],
+      [UserCustomDimension.NodeVersion]: nodeVersion ?? 'other',
```
alan-agius4, line 64:
```suggestion
      [UserCustomDimension.NodeVersion]: nodeVersion,
```

packages/angular/cli/src/analytics/analytics-collector.ts
```diff
@@ -62,8 +61,8 @@ export class AnalyticsCollector {
...
+      [UserCustomDimension.NodeMajorVersion]: nodeVersion ? +nodeVersion.split('.', 1)[0] : 'other',
```
alan-agius4, line 65: This parameter can only have a value as a number or undefined.

packages/angular/cli/src/analytics/analytics-collector.ts
dgp1130: Thanks for pointing that out, I wasn't aware of `semver.clean()`. Though I think that still leaves in the user-defined pre-release? https://www.npmjs.com/package/semver indicates that `semver.clean('2.1.5-test') === '2.1.5-test'`, so there's still an unbounded string there we need to strip.

alan-agius4: Your `parseVersion` also keeps the prerelease string. 

Couple of ways you can remove the prerelease.

```ts
const version = '2.1.5-test';

// Using clean
let newVersion = semver.clean(version).split('-', 1)[0];

// Using parse
const [major, minor, patch] = semver.parse(version);
let newVersion = `${major}.${minor}.${patch}`
```

The latter can also be used to populate `UserCustomDimension.NodeMajorVersion`

dgp1130: Ah whoops, I got so focused on the leading `v` I forgot about the prerelease :sweat_smile:. I like your `semver.parse()` approach, updated to use that.

packages/angular/cli/src/analytics/analytics-collector.ts
dgp1130: Updated to be `undefined` if we don't have a valid version.

packages/angular/cli/src/analytics/analytics-collector.ts
```diff
@@ -52,8 +53,7 @@ export class AnalyticsCollector {
 
     this.requestParameterStringified = querystring.stringify(requestParameters);
 
-    // Remove the `v` at the beginning.
-    const nodeVersion = process.version.substring(1);
+    const parsedVersion = semver.parse(process.version.slice(1) /* drop leading 'v' */);
```
alan-agius4, line 56: The `semver.parse` will clean up any leading `v`.

```suggestion
    const parsedVersion = semver.parse(process.version);
```

alan-agius4 (approved):
LGTM.


One nit.

packages/angular/cli/src/analytics/analytics-collector.ts
alan-agius4, line 56: The `semver.parse` will clean up any leading `v`.

```suggestion
    const parsedVersion = semver.parse(process.version);
```

PR #24565, opened 2023-01-18
feat(@angular/cli): log number of files update during `ng update`

### LINKED ISSUE #24488: Tell users what migrations were actually performed instead of what was considered
### Command

update

### Description

Currently `ng update` does  a good job of telling you which migrations the CLI is attempting to apply, but you don't know which migrations actually _were_  applied.

I'd love it if the migrations reports / logs actually logged what changes were made.  Example console output before:

```
UPDATE angular.json (3617 bytes)
  Migration completed.

▸ Remove 'showCircularDependencies' option from browser and server builders.
  Migration completed.

▸ Replace 'defaultCollection' option in workspace configuration with 'schematicCollections'.
  Migration completed.

▸ Update Angular packages 'dependencies' and 'devDependencies' version prefix to '^' instead of '~'.

```

Example after:
```
UPDATE angular.json (3617 bytes)
  Migration completed - no changes made.

▸ Remove 'showCircularDependencies' option from browser and server builders.
  Migration completed - no changes made.

▸ Replace 'defaultCollection' option in workspace configuration with 'schematicCollections'.
  Migration completed - 3 updates made.

▸ Update Angular packages 'dependencies' and 'devDependencies' version prefix to '^' instead of '~'.
Migration completed - 15 updates made.

```

_No response_

alan-agius4:
This commit updates `ng update` to include the number of files updated when a migration is completed.

Closes #24488

### DIFF
--- a/packages/angular/cli/src/commands/update/cli.ts
+++ b/packages/angular/cli/src/commands/update/cli.ts
@@ -387,16 +387,29 @@ export class UpdateCommandModule extends CommandModule<UpdateCommandArgs> {
         logger.info('  ' + description.join('.\n  '));
       }
 
-      const result = await this.executeSchematic(
+      const { success, files } = await this.executeSchematic(
         workflow,
         migration.collection.name,
         migration.name,
       );
-      if (!result.success) {
+      if (!success) {
         return 1;
       }
 
-      logger.info('  Migration completed.');
+      let modifiedFilesText: string;
+      switch (files.size) {
+        case 0:
+          modifiedFilesText = 'No changes made';
+          break;
+        case 1:
+          modifiedFilesText = '1 file modified';
+          break;
+        default:
+          modifiedFilesText = `${files.size} files modified`;
+          break;
+      }
+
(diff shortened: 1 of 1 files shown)


packages/angular/cli/src/commands/update/cli.ts
```diff
@@ -387,16 +387,16 @@ export class UpdateCommandModule extends CommandModule<UpdateCommandArgs> {
...
         migration.collection.name,
         migration.name,
       );
-      if (!result.success) {
+      if (!success) {
         return 1;
       }
 
-      logger.info('  Migration completed.');
+      logger.info(`  Migration completed (${files.size} files modified).`);
```
clydin, line 399:
```suggestion
      logger.info(`  Migration completed (${files.size} file(s) modified).`);
```
Could optionally make it smarter and add the `s` to file if the size is greater than 1.

Also, I think we should special case zero with a more direct message that no changes were required.

clydin (approved):

PR #24635, opened 2023-01-31
feat(@angular-devkit/build-angular): support JIT compilation with esbuild

clydin:
When using the experimental esbuild-based browser application builder, the `aot` build option can now be set to `false` to enable JIT compilation mode. The JIT mode compilation operates in a similar fashion to the Webpack-based builder in JIT mode. All external Component stylesheet and template references are converted to static import statements and then the content is bundled as text. All inline styles are also processed in this way as well to support inline style languages such as Sass. This approach also has the advantage of minimizing the processing necessary during rebuilds. In JIT watch mode, TypeScript code does not need to be reprocessed if only an external stylesheet or template is changed.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/compiler-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/compiler-plugin.ts
@@ -21,6 +21,8 @@ import * as path from 'node:path';
 import { pathToFileURL } from 'node:url';
 import ts from 'typescript';
 import { maxWorkers } from '../../utils/environment-options';
+import { JitCompilation } from './angular/jit-compilation';
+import { setupJitPluginCallbacks } from './angular/jit-plugin-callbacks';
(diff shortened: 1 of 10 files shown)


packages/angular_devkit/build_angular/src/builders/browser-esbuild/compiler-plugin.ts
```diff
@@ -291,7 +287,11 @@ export function createCompilerPlugin(
         };
 
         // Create new compilation if first build; otherwise, use existing for rebuilds
-        compilation ??= new AngularCompilation();
+        if (pluginOptions.jit) {
+          compilation ??= new JitCompilation();
+        } else {
+          compilation ??= new AngularCompilation();
```
clydin, line 293: NOTE: `AngularCompilation` will be renamed to `AotCompilation` in a follow-up PR.

alan-agius4 (approved):

PR #24662, opened 2023-02-05
fix(@angular/cli): error if Angular compiler is used in a schematic

crisbeto:
When a schematic is executed, it is wrapped in a custom Node context. This context doesn't expose the same set of global variables as the default one. This can lead to an error if a schematic is importing the Angular compiler and the app is using i18n, because the `TextEncoder` isn't exposed through the custom context (see https://github.com/angular/angular/issues/48940).

These changes add the `TextEncoder` to the context.

Fixes https://github.com/angular/angular/issues/48940.

### DIFF
--- a/packages/angular/cli/src/command-builder/utilities/schematic-engine-host.ts
+++ b/packages/angular/cli/src/command-builder/utilities/schematic-engine-host.ts
@@ -12,6 +12,7 @@ import { readFileSync } from 'fs';
 import { parse as parseJson } from 'jsonc-parser';
 import { createRequire } from 'module';
 import { dirname, resolve } from 'path';
+import { TextEncoder } from 'util';
 import { Script } from 'vm';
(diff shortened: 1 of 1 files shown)


packages/angular/cli/src/command-builder/utilities/schematic-engine-host.ts
```diff
@@ -12,6 +12,7 @@ import { readFileSync } from 'fs';
 import { parse as parseJson } from 'jsonc-parser';
 import { createRequire } from 'module';
 import { dirname, resolve } from 'path';
+import { TextEncoder } from 'util';
```
crisbeto, line 15: I used an `import` for this, even though it's available globally, because for some reason the global variable wasn't in the typings.

alan-agius4 (approved):

PR #24666, opened 2023-02-06
fix(@schematics/angular): remove bootstrapping wrapping in universal schematic

alan-agius4:
The bootstrapping wrapping is no longer necessary as of 15.1.3. See: https://github.com/angular/angular/pull/48868

### DIFF
--- a/packages/schematics/angular/universal/index.ts
+++ b/packages/schematics/angular/universal/index.ts
@@ -29,7 +29,7 @@ import {
   getPackageJsonDependency,
 } from '../utility/dependencies';
 import { latestVersions } from '../utility/latest-versions';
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/utility/ng-ast-utils.ts
```diff
@@ -12,7 +12,7 @@ import { dirname } from 'path';
 import * as ts from '../third_party/github.com/Microsoft/TypeScript/lib/typescript';
 import { findNode, getSourceNodes } from '../utility/ast-utils';
 
-export function findBootstrapModuleCall(host: Tree, mainPath: string): ts.CallExpression | null {
```
clydin, line 15: While this is technically not public API, it is still accessible.  We may want to leave it exported for now and drop it in v16 just to avoid any potential breakage.

clydin (approved):

PR #24692, opened 2023-02-09
fix(@angular-devkit/build-angular): use babel default export helper in build optimizer

### LINKED ISSUE #24688: Build failure for default exported modules
### Command

build

### Is this a regression?

15.1.4

### Description

I'm using the [router unwrap feature](https://blog.angular.io/angular-v15-is-now-available-df7be7f2f4c8#:~:text=Router%20unwraps%20default%20imports) of Angular 15 when loading lazy loaded modules in the router module.
In version 15.1.4 it worked, but since version 15.1.5 the build fails for the lazy loaded modules.

### Minimal Reproduction

This worked before:

```
import { RouterModule, Routes } from '@angular/router';

const routes: Routes = [
  {
    path: 'lazy-loaded',
    loadChildren: () => import('./lazy-loaded/lazy-loaded.module')
  }
];

@NgModule({
  imports: [RouterModule.forRoot(routes)],
  exports: [RouterModule]
})
export class AppRoutingModule {}
```

```

@NgModule({})
export default class LazyLoadedModule {}
```

But now it breaks, unless I change it back to the "old" way:

```
import { RouterModule, Routes } from '@angular/router';

const routes: Routes = [
  {
    path: 'lazy-loaded',
    loadChildren: () => import('./lazy-loaded/lazy-loaded.module').then(m => m.LazyLoadedModule)
  }
];

@NgModule({
  imports: [RouterModule.forRoot(routes)],
  exports: [RouterModule]
})
export class AppRoutingModule {}
```

```

@NgModule({})
export class LazyLoadedModule {}
```

### Exception or Error

```text
./src/app/lazy-loaded/lazy-loaded.module.ts - Error: Module build failed (from ./node_modules/@angular-devkit/build-angular/src/babel/webpack-loader.js):
Error: /ng-app/src/app/lazy-loaded/lazy-loaded.module.ts: We don't know what to do with this node type. We were previously a Statement but we can't fit in here?
```

### Your Environment

```text
Angular CLI: 15.1.5
Node: 16.19.0
Package Manager: npm 8.19.3
OS: win32 x64

Angular: 15.1.4
... language-service, localize, material, platform-browser

@angular-devkit/architect       0.1501.5
@angular-devkit/build-angular   15.1.5
@angular-devkit/core            15.1.5
@angular-devkit/schematics      15.1.5
@angular/cli                    15.1.5
@schematics/angular             15.1.5
rxjs                            7.8.0
typescript                      4.9.5
```

### Anything else relevant?

_No response_

clydin:
Within the build optimizer's static member optimization pass, a class that is directly default exported must be split into two statements: the class declaration and the default export. This is because the pass can wrap classes in a pure annotated IIFE which results in a variable declaration replacement and variable declarations can not be directly default exported. Previously, the pass did this splitting manually but this was causing later babel plugins to fail. In addition to updating the AST in this case, scoping information also needed to be updated. To support this, a babel helper package is now used that handles the details of the statement split operation.

Fixes #24688

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "@babel/core": "7.20.12",
     "@babel/generator": "7.20.14",
     "@babel/helper-annotate-as-pure": "7.18.6",
+    "@babel/helper-split-export-declaration": "7.18.6",
     "@babel/plugin-proposal-async-generator-functions": "7.20.7",
     "@babel/plugin-transform-async-to-generator": "7.20.7",
     "@babel/plugin-transform-runtime": "7.19.6",
--- a/packages/angular_devkit/build_angular/BUILD.bazel
+++ b/packages/angular_devkit/build_angular/BUILD.bazel
@@ -111,6 +111,7 @@ ts_library(
         "@npm//@babel/core",
         "@npm//@babel/generator",
         "@npm//@babel/helper-annotate-as-pure",
+        "@npm//@babel/helper-split-export-declaration",
         "@npm//@babel/plugin-proposal-async-generator-functions",
         "@npm//@babel/plugin-transform-async-to-generator",
         "@npm//@babel/plugin-transform-runtime",
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -13,6 +13,7 @@
     "@babel/core": "7.20.12",
     "@babel/generator": "7.20.14",
     "@babel/helper-annotate-as-pure": "7.18.6",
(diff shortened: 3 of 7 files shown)


alan-agius4 (changes requested):
`@babel/helper-split-export-declaration` should be added as a `dep` https://github.com/angular/angular-cli/blob/77aa04ceadb5be15aceae89a3d4780ff5657f1c6/packages/angular_devkit/build_angular/BUILD.bazel#L98

alan-agius4 (approved):

PR #24737, opened 2023-02-16
fix(@angular-devkit/build-angular): allow empty scripts to be optimized

### LINKED ISSUE #24736: Error: Terser failed for unknown reason if scripts file is empty
### Command

build, serve

### Is this a regression?

11.0.7

### Description

Hi, I have upgraded angular from `11.0.7` to `13.3.10` and now Terser plugin throws an error if I have `scripts` in my build config and they produce an empty Terser build. It happens **only** with `"aot": true`

### Minimal Reproduction

```
 "targets": {
    "build": {
      "executor": "@angular-devkit/build-angular:browser",
      "options": {
        .... 
        "aot": true,
        "scripts": [
          {
            "input": "apps/app/src/monitoring.js",
            "inject": false,
            "bundleName": "monitoring"
          }
        ]
      },
```

`monitoring.js` looks like this:
```
// some commented js code
```

Command to run
```
> nx run app:build
```

### Exception or Error

```text
✔ Browser application bundle generation complete.

Error: Optimization error [monitoring.js]: Error: Terser failed for unknown reason.
    at optimizeWithTerser (/Users/cakeinpanic/projects/app/node_modules/@angular-devkit/build-angular/src/webpack/plugins/javascript-optimizer-worker.js:137:15)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async default_1 (/Users/cakeinpanic/projects/app/node_modules/@angular-devkit/build-angular/src/webpack/plugins/javascript-optimizer-worker.js:29:26)
    at async /Users/cakeinpanic/projects/app/node_modules/piscina/dist/src/worker.js:141:26
```

### Your Environment

```text
"@angular/animations": "13.3.12",
    "@angular/cdk": "13.3.9",
    "@angular/common": "13.3.12",
    "@angular/compiler": "13.3.12",
    "@angular/core": "13.3.12",
    "@angular/forms": "13.3.12",
    "@angular/platform-browser": "13.3.12",
    "@angular/platform-browser-dynamic": "13.3.12",
    "@angular/router": "13.3.12",
    "@angular-devkit/build-angular": "13.3.10",
    "@angular/cli": "13.3.10",
    "@angular/compiler-cli": "13.3.12",
```

### Anything else relevant?

1. If I change`monitoring.js`content to 🔽 all build well
```
console.log(333)
// some commented js code
```
2. If I remove `"aot": true`, all build well

I assume that the issue is that Terser produces an *empty file* and can't handle this

clydin:
When using the internal JavaScript optimizer plugin for Webpack with an empty script file provided via the `scripts` option, the build would fail. This was because of a safety check that was checking whether the terser result was falsy. Since an empty string is considered falsy, the build considered the result to be an error. The safety check now will only trigger if the terser result is not a string value to avoid this case.

Fixes #24736

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser/tests/options/scripts_spec.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser/tests/options/scripts_spec.ts
@@ -27,6 +27,27 @@ describeBuilder(buildWebpackBrowser, BROWSER_BUILDER_INFO, (harness) => {
       expect(result?.success).toBe(true);
     });
 
+    it('processes an empty script when optimizing', async () => {
+      await harness.writeFile('src/test-script-a.js', '');
+
+      harness.useTarget('build', {
+        ...BASE_OPTIONS,
+        optimization: {
+          scripts: true,
+        },
+        scripts: ['src/test-script-a.js'],
+      });
+
+      const { result } = await harness.executeOnce();
+
+      expect(result?.success).toBe(true);
+
+      harness.expectFile('dist/scripts.js').toExist();
+      harness
+        .expectFile('dist/index.html')
+        .content.toContain('<script src="scripts.js" defer></script>');
+    });
+
     describe('shorthand syntax', () => {
       it('processes a single script into a single output', async () => {
         await harness.writeFile('src/test-script-a.js', 'console.log("a");');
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/builders/browser/tests/options/scripts_spec.ts
```diff
@@ -27,6 +27,27 @@ describeBuilder(buildWebpackBrowser, BROWSER_BUILDER_INFO, (harness) => {
       expect(result?.success).toBe(true);
     });
 
+    fit('processes an empty script when optimizing', async () => {
```
alan-agius4, line 30: Oops

alan-agius4 (approved):

PR #24757, opened 2023-02-22
build: update angular dependencies to version 15.2.x

alan-agius4:


alan-agius4:
This will need a re-run once FW releases the stable versions.

dgp1130 (approved):

PR #24772, opened 2023-02-23
build: add pre-release checks for dependency on framework

devversion:
Adds an automatic check that runs before cutting releases. This will help avoiding issues where
we forget to update peer dependencies or the `latest-versions.ts` file.

Screenshot 2023-02-24 at 14 31 26

### DIFF
--- a/scripts/release-checks/dependency-ranges/index.mts
+++ b/scripts/release-checks/dependency-ranges/index.mts
@@ -1,0 +1,54 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import semver from 'semver';
+import { Log, bold, ReleasePrecheckError } from '@angular/ng-dev';
+import { checkPeerDependencies } from './peer-deps-check.mjs';
+import { checkSchematicsAngularLatestVersion } from './latest-versions-check.mjs';
+import { PackageMap } from '../../../lib/packages.js';
+
+/** Environment variable that can be used to skip this pre-check. */
+const skipEnvVar = 'SKIP_DEPENDENCY_RANGE_PRECHECK';
+
+/**
+ * Ensures that dependency ranges are properly updated before publishing
+ * a new version. This check includes:
+ *
+ *   - checking of `latest-versions.ts` of `@schematics/angular`.
+ *   - checking of peer dependencies in `@angular-devkit/build-angular`.
+ *
+ * @throws {ReleasePrecheckError} If validation fails.
+ */
+export async function assertValidDependencyRanges(
+  newVersion: semver.SemVer,
+  allPackages: PackageMap,
+) {
+  if (process.env[skipEnvVar] === '1') {
+    return;
+  }
+
+  const failures: string[] = [
+    ...(await checkPeerDependencies(newVersion, allPackages)),
+    ...(await checkSchematicsAngularLatestVersion(newVersion)),
+  ];
+
+  if (failures.length !== 0) {
+    Log.error('Discovered errors when validating dependency ranges.');
+
+    for (const f of failures) {
+      Log.error(`  - ${bold(f)}`);
+    }
+
+    Log.warn();
+    Log.warn('Please fix these failures before publishing a new release.');
+    Log.warn(`These checks can be forcibly ignored by setting: ${skipEnvVar}=1`);
+    Log.warn();
+
+    throw new ReleasePrecheckError();
+  }
+}
--- a/scripts/release-checks/dependency-ranges/latest-versions-check.mts
+++ b/scripts/release-checks/dependency-ranges/latest-versions-check.mts
@@ -1,0 +1,37 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import semver from 'semver';
+
+export async function checkSchematicsAngularLatestVersion(
+  newVersion: semver.SemVer,
+): Promise<string[]> {
+  const {
+    default: { latestVersions },
+  } = await import('../../../packages/schematics/angular/utility/latest-versions.js');
+
+  const keysToCheck = ['ng-packagr', 'Angular'];
+  const { major, minor } = newVersion;
+  const isPrerelease = !!newVersion.prerelease[0];
+  const failures: string[] = [];
+
+  let expectedFwDep = `^${major}.${minor}.0`;
+  if (isPrerelease) {
+    expectedFwDep = `^${major}.${minor}.0-next.0`;
+  }
+
+  for (const key of keysToCheck) {
+    if (latestVersions[key] !== expectedFwDep) {
+      failures.push(
+        `latest-versions: Invalid dependency range for "${key}". Expected: ${expectedFwDep}`,
+      );
+    }
+  }
+
+  return failures;
+}
--- a/scripts/release-checks/dependency-ranges/peer-deps-check.mts
+++ b/scripts/release-checks/dependency-ranges/peer-deps-check.mts
@@ -1,0 +1,74 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import path from 'path';
+import url from 'url';
+import semver from 'semver';
+import { PackageMap } from '../../../lib/packages.js';
+
+/** Path to the current directory. */
+const currentDir = path.dirname(url.fileURLToPath(import.meta.url));
+
+/** Path to the project directory. */
+const projectDir = path.join(currentDir, '../../../');
+
+/** Describes a parsed `package.json` file. */
+interface PackageJson {
+  name?: string;
+  peerDependencies?: Record<string, string>;
+}
+
+export async function checkPeerDependencies(
+  newVersion: semver.SemVer,
+  allPackages: PackageMap,
+): Promise<string[]> {
+  const { major, minor } = newVersion;
+  const isPrerelease = !!newVersion.prerelease[0];
+  const isMajor = minor === 0;
+
+  let expectedFwPeerDep = `^${major}.0.0`;
+  if (isMajor && isPrerelease) {
+    expectedFwPeerDep = `^${major}.0.0-next.0`;
+  } else if (isPrerelease) {
+    expectedFwPeerDep = `^${major}.0.0 || ^${major}.${minor}.0-next.0`;
+  }
+
+  const failures: string[] = [];
+  for (const pkgInfo of Object.values(allPackages)) {
+    failures.push(...checkPackage(pkgInfo.packageJson, expectedFwPeerDep));
+  }
+
+  return failures;
+}
+
+/** Checks the given package and collects errors for the peer dependency ranges. */
+function checkPackage(pkgJson: PackageJson, expectedFwPeerDep: string): string[] {
+  if (pkgJson.peerDependencies === undefined) {
+    return [];
+  }
+
+  const failures: string[] = [];
+
+  for (const [depName, range] of Object.entries(pkgJson.peerDependencies)) {
+    // Even though `ng-packagr` might not strictly follow the same release schedules
+    // like official Angular packages, we generally expect it to match. It's better
+    // flagging it than silently passing pre-checks. The caretaker can always forcibly
+    // ignore this check.
+    if (!depName.startsWith('@angular/') && depName !== 'ng-packagr') {
+      continue;
+    }
+
+    if (range !== expectedFwPeerDep) {
+      failures.push(
+        `${pkgJson.name}: Unexpected peer dependency range for "${depName}". Expected: ${expectedFwPeerDep}`,
+      );
+    }
+  }
+
+  return failures;
+}
--- a/.ng-dev/release.mts
+++ b/.ng-dev/release.mts
@@ -1,3 +1,4 @@
+import semver from 'semver';
 import { ReleaseConfig } from '@angular/ng-dev';
 import packages from '../lib/packages.js';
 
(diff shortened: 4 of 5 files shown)


scripts/release-checks/assert-valid-peer-dependency.mts
```diff
@@ -0,0 +1,54 @@
...
+  let expectedFwDep: string;
+  if (isMajor) {
+    if (isPrerelease) {
+      expectedFwDep = `^${major}.0.0-next.0`;
+    } else {
+      expectedFwDep = `^${major}.0.0 || ^${major}.${newVersion.minor + 1}.0-next.0`;
+    }
+  } else {
+    expectedFwDep = `^${major}.0.0  || ^${newVersion.major}.${newVersion.minor + 1}.0-next.0`;
+  }
```
dgp1130, line 41: I think we also need to include the RC phase in this. I think it just needs to be `|| ^x.x.x-rc.0` alongside `-next` releases?

dgp1130, line 41: Patch releases should just be `^${major}.0.0` right?

scripts/release-checks/assert-valid-peer-dependency.mts
```diff
@@ -0,0 +1,54 @@
...
+  const ngDevkitBuildAngularPkgJson = await loadPackageJson(
+    path.join(projectDir, 'packages/angular_devkit/build_angular/package.json'),
+  );
+
+  let expectedFwDep: string;
+  if (isMajor) {
+    if (isPrerelease) {
+      expectedFwDep = `^${major}.0.0-next.0`;
+    } else {
+      expectedFwDep = `^${major}.0.0 || ^${major}.${newVersion.minor + 1}.0-next.0`;
```
dgp1130, line 37: Do we normally include `-next` in non-prerelease versions? For example, I thought CLI `15.2.0` would only have a peer dep on FW `^15.0.0`?

scripts/release-checks/assert-valid-peer-dependency.mts
```diff
@@ -0,0 +1,54 @@
...
+    }
+  } else {
+    expectedFwDep = `^${major}.0.0  || ^${newVersion.major}.${newVersion.minor + 1}.0-next.0`;
+  }
+
+  // TODO: check all angular packages.
+  // TODO: check `latest-versions.ts`.
```
dgp1130, line 44: https://github.com/angular/angular-cli/blob/711214f210f10e4881b1bd419def9920e95a78ba/packages/schematics/angular/utility/latest-versions.ts contains the instruction we typically use and where they need to be updated.

I think we also need to check `ng-packagr` based on the last PR? https://github.com/angular/angular-cli/pull/24757

dgp1130 (commented):
Disclaimer: I get confused by these versions all the time, so @alan-agius4 and @clydin are definitely the authority on what we should set when.

General question: Does this work when we need to bump peer dependencies before the release PR? That means we would have peer deps updated to the version we're _about_ to release, while the repository is still using the _previous_ version number. Is that a problem here?

scripts/release-checks/assert-valid-peer-dependency.mts
alan-agius4: Correct on a stable version we do not add add any pre-release support.

scripts/release-checks/assert-valid-peer-dependency.mts
```diff
@@ -0,0 +1,54 @@
...
+
+  let expectedFwDep: string;
+  if (isMajor) {
+    if (isPrerelease) {
+      expectedFwDep = `^${major}.0.0-next.0`;
+    } else {
+      expectedFwDep = `^${major}.0.0 || ^${major}.${newVersion.minor + 1}.0-next.0`;
+    }
+  } else {
+    expectedFwDep = `^${major}.0.0  || ^${newVersion.major}.${newVersion.minor + 1}.0-next.0`;
```
alan-agius4, line 40: `^${newVersion.major}.${newVersion.minor + 1}.0-next.0` should only been done if if't a prerelease. I mean technically I does not hurt, but I will cause NPM peer deps errors more verbose.

scripts/release-checks/assert-valid-peer-dependency.mts
alan-agius4: `-next` also includes `-rc`.

alan-agius4: Yes

scripts/release-checks/assert-valid-peer-dependency.mts
alan-agius4: I am not 100% certain about ng-packagr, just because currently it is a in different repo and releases are cut by myself. 

While I do cut an ng-packagr release prior to cutting the CLI. This might not always happen in the future. as I might be unavailable for the day either sick (Which technically I can still do a release if needed), be on vacation or have no internet access. At which point this check would cause the CLI release to be blocked.

scripts/release-checks/assert-valid-peer-dependency.mts
devversion: You need to be careful with adding too many variations because otherwise it will become make the release process more cumbersome.

scripts/release-checks/assert-valid-peer-dependency.mts
devversion: Discussed offline. We'll check `ng-packagr` but have an escape hatch.

devversion:
@dgp1130 The checks run _only_ as part of the release tool before the version bump is proposed as a PR. The checks depend on what the new version will be (based on the selected release action). Hence the checks cannot run on CI, but can run as a release pre-check. Should not cause any problems if you bump them before. That's actually expected!

scripts/release-checks/dependency-ranges/build-angular-peer-deps-check.mts
```diff
@@ -0,0 +1,54 @@
...
+const currentDir = path.dirname(url.fileURLToPath(import.meta.url));
+
+/** Path to the project directory. */
+const projectDir = path.join(currentDir, '../../../');
+
+export async function checkBuildAngularPeerDependencies(
+  newVersion: semver.SemVer,
+): Promise<string[]> {
+  const pkgJson = await loadPackageJson(
+    path.join(projectDir, 'packages/angular_devkit/build_angular/package.json'),
```
alan-agius4, line 24: `packages/ngtools/webpack/package.json` also have a peer dep that need to be checked. 

Instead of listing the packages here, I suggest to use the default export from `'lib/packages.js'

```
import packages from '../lib/packages.js';
```

The above, has already the loaded package.json files, and paths.

scripts/release-checks/dependency-ranges/latest-versions-check.mts
```diff
@@ -0,0 +1,34 @@
...
+  if (isPrerelease) {
+    expectedFwDep = `^${major}.${minor}.0-next.0`;
+  }
+
+  if (latestVersions.Angular !== expectedFwDep) {
+    failures.push(
+      `latest-versions: Invalid dependency range on the framework. Expected: ${expectedFwDep}`,
+    );
+  }
+
```
alan-agius4, line 32: You also need to add the check for ng-packagr here.

scripts/release-checks/dependency-ranges/build-angular-peer-deps-check.mts
```diff
@@ -0,0 +1,54 @@
...
+    // like official Angular packages, we generally expect it to match. It's better
+    // flagging it than silently passing pre-checks. The caretaker can always forcibly
+    // ignore this check.
+    if (!depName.startsWith('@angular/') && depName !== 'ng-packagr') {
+      continue;
+    }
+
+    if (range !== expectedFwPeerDep) {
+      failures.push(
+        `${pkgJson.name}: Unexpected peer dependency for "${depName}". Expected: ${expectedFwPeerDep}`,
```
alan-agius4, line 49: Would be helpful to indicate which package.json caused the failure.

scripts/release-checks/dependency-ranges/build-angular-peer-deps-check.mts
devversion: Changed to checking _all_ packages now, using `lib/packages.js`

scripts/release-checks/dependency-ranges/build-angular-peer-deps-check.mts
devversion: I decided to not do this because the package name should be sufficient. Feel free to add this in the future if it ends up being confusing.

alan-agius4 (approved):
LGTM, thanks for this 🚀

alan-agius4 (approved):

scripts/release-checks/dependency-ranges/peer-deps-check.mts
```diff
@@ -0,0 +1,74 @@
...
+import path from 'path';
+import url from 'url';
+import semver from 'semver';
+import { PackageMap } from '../../../lib/packages.js';
+
+/** Path to the current directory. */
+const currentDir = path.dirname(url.fileURLToPath(import.meta.url));
+
+/** Path to the project directory. */
+const projectDir = path.join(currentDir, '../../../');
```
dgp1130, line 18: Consider: Could we resolve the `package.json` file to find the project root instead of manually maintaining `../` paths? Would be easy to move this file and forget to update it. Not sure if there's an easy way to get the path of the current `package.json` without walking the directory tree, but something to consider.

scripts/release-checks/dependency-ranges/peer-deps-check.mts
```diff
@@ -0,0 +1,74 @@
...
+  if (isMajor && isPrerelease) {
+    expectedFwPeerDep = `^${major}.0.0-next.0`;
+  } else if (isPrerelease) {
+    expectedFwPeerDep = `^${major}.0.0 || ^${major}.${minor}.0-next.0`;
+  }
+
+  const failures: string[] = [];
+  for (const pkgInfo of Object.values(allPackages)) {
+    failures.push(...checkPackage(pkgInfo.packageJson, expectedFwPeerDep));
+  }
```
dgp1130, line 44: Consider:

```typescript
const failures = Object.values(allPackages)
    .flatMap((pkgInfo) => checkPackage(pkgInfo.packageJson, expectedFwPeerDep));
```

scripts/release-checks/dependency-ranges/peer-deps-check.mts
```diff
@@ -0,0 +1,74 @@
...
+  return failures;
+}
+
+/** Checks the given package and collects errors for the peer dependency ranges. */
+function checkPackage(pkgJson: PackageJson, expectedFwPeerDep: string): string[] {
+  if (pkgJson.peerDependencies === undefined) {
+    return [];
+  }
+
+  const failures: string[] = [];
```
dgp1130, line 55: One trick I've used in the past for stuff like this to avoid manual array management is yielding error messages like:

```typescript
function checkPackage(pkgJson: PackageJson, expectedFwPeerDep: string): Generator<string, void, void> {
  for (const [depName, range] of Object.entries(pkgJson.peerDependencies)) {
    if (range !== expectedFwPeerDep) {
      yield `${pkgJson.name}: Unexpected peer dependency range for "${depName}". Expected: ${expectedFwPeerDep}`;
    }
  }
}
```

It a less common pattern I know, so ultimately up to you. Just one thing to think about.

scripts/release-checks/dependency-ranges/index.mts
```diff
@@ -0,0 +1,54 @@
...
+ *   - checking of `latest-versions.ts` of `@schematics/angular`.
+ *   - checking of peer dependencies in `@angular-devkit/build-angular`.
+ *
+ * @throws {ReleasePrecheckError} If validation fails.
+ */
+export async function assertValidDependencyRanges(
+  newVersion: semver.SemVer,
+  allPackages: PackageMap,
+) {
+  if (process.env[skipEnvVar] === '1') {
```
dgp1130, line 31: Nit: I could see someone trying `SKIP_DEPENDENCY_RANGE_PRECHECK="true"` and being confused when that doesn't work. I wonder if we should skip if the variable is truthy or at least throw an error that the only valid value is `1` or unset?

dgp1130 (commented):
Thanks @devversion! Looks great, mostly just some optional suggestions for improvements.

My one real concern is testing, can we include any tests for these checks? I know that can be a bit tricky in this context, but I'm worried the first time we really run this will be on release day, which is the worst possible time to try and fix it. Some tests would at least make sure they catch all the common mistakes and that we don't accidentally break it in the future.

scripts/release-checks/dependency-ranges/peer-deps-check.mts
devversion: That's the common pattern we follow in other repos too. You would need to adjust relative imports as well, so not something new here 😆

scripts/release-checks/dependency-ranges/peer-deps-check.mts
devversion: Fair enough. I think it comes down to personal preference. Please feel free to make this change if you feel like it

scripts/release-checks/dependency-ranges/peer-deps-check.mts
devversion: great suggestion, but see comment above.

scripts/release-checks/dependency-ranges/index.mts
devversion: Honestly, I don't know it's worth doing that. the error is pretty clear about what needs to be used, and it's also only the experienced caretakers running the release process.

devversion:
@dgp1130 yeah, pre-release checks are by nature not trivially testable because for CI/testing it would be unclear what the "next" version would be. We could have unit tests for e.g. possible semver computations etc, but testing the full integration is non-trivial. e.g. might need a virtual patched FS to modify package json files etc. What is testing the tests? 😆

I agree with the sentiment in general, but not sure if we can realistically do anything about it/ or if it's worth it- I think it's already better than not having anything, like before. There might be a way to improve this, but that's another larger story.

Also note that this check is never blocking. On release day it could always be disabled.

dgp1130:
Agreed that gets complicated, and the skip is a good opt-out in this particular case which derisks it a lot.

Sorry I didn't realize this merged in before my last comments, definitely nothing important enough to require a follow-up PR. :sweat_smile:

PR #24779, opened 2023-02-25
feat(@schematics/angular): factor out logic from getRouterModuleDeclaration

samali0996:
Can now get module declarations of modules other than RouterModule. This is useful for adding a custom declaration to module using schematics. Checking the spec files will show a use case for this new feature, now the getModuleDeclarationByName function can be used to get the module declaration of a custom module, in this case ngxsModule.

## PR Checklist

- [ x] The commit message follows our guidelines: https://github.com/angular/angular-cli/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [x ] Tests for the changes have been added (for bug fixes / features)
- [x ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes
- [ ] CI related changes

The logic of getting module declaration was tied up in getRouterModuleDeclaration. This limits us from using that function to get other Modules.
On my personal project I am working on schematics that add new Ngxs States, including adding them to the NgxsModule in the NgModule.

Issue Number: N/A

## What is the new behavior?
These modifications will allow me to use the angular schematics utility functions to assist me with my personal project.

- [ ] Yes
- [x ] No

### DIFF
--- a/packages/schematics/angular/utility/ast-utils.ts
+++ b/packages/schematics/angular/utility/ast-utils.ts
@@ -554,9 +554,15 @@ export function isImported(
 }
 
 /**
- * Returns the RouterModule declaration from NgModule metadata, if any.
+ * Returns the module declaration from NgModule metadata, by name, if exists.
+ * @param source the source object
+ * @param moduleName name of the module declaration to get
+ * @return module declaration
  */
-export function getRouterModuleDeclaration(source: ts.SourceFile): ts.Expression | undefined {
+export function getModuleDeclarationByName(
+  source: ts.SourceFile,
+  moduleName: string,
+): ts.Expression | undefined {
   const result = getDecoratorMetadata(source, 'NgModule', '@angular/core');
   const node = result[0];
   if (!node || !ts.isObjectLiteralExpression(node)) {
@@ -578,7 +584,14 @@ export function getRouterModuleDeclaration(source: ts.SourceFile): ts.Expression
 
   return arrLiteral.elements
     .filter((el) => el.kind === ts.SyntaxKind.CallExpression)
-    .find((el) => (el as ts.Identifier).getText().startsWith('RouterModule'));
+    .find((el) => (el as ts.Identifier).getText().startsWith(moduleName));
+}
+
+/**
+ * Returns the RouterModule declaration from NgModule metadata, if exists.
+ */
+export function getRouterModuleDeclaration(source: ts.SourceFile): ts.Expression | undefined {
+  return getModuleDeclarationByName(source, 'RouterModule');
 }
 
 /**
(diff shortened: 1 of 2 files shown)


alan-agius4:
The AST utils are considered part of the private API.

Anything which is not exported from `@schematics/angular/utility` directly and require a deep import is private.

samali0996:
> The AST utils are considered part of the private API.
> 
> Anything which is not exported from `@schematics/angular/utility` directly and require a deep import is private.

Hi Alan, thanks for the response,
Does this mean we cannot merge this pull request in? I was thinking this was a good refactor since `getRouterModuleDeclaration` is a very big block of code, and breaking it off into smaller functions follows good code practice.

Are there rules towards adding new private API functions? It should be safer then to create a new function since it's considered private?

Thanks!

samali0996:
Further justification:

I am writing a schematic that adds a newly generated NgxsState to a specified NgModule, and I found myself needing to use a lot of the same logic as the `addRouteDeclarationToModule`, but with parameters differing from `Router` module.

There are many helpful utility functions that I am already using, such as 
```
import {
 buildDefaultPath,
  getWorkspace,
} from '@schematics/angular/utility/workspace';
```
But I am unable to use the logic used by `addRouteDeclarationToModule` because that's all baked into one function geared towards Router.

This change refactors and creates a new function `getModuleDeclarationByName` which allows its use in a similar style to other functions in this file, such as `getDecoratorMetadata`, `addDeclarationToModule`,

samali0996:
This is a first step I'm working on towards decoupling the logic of `addRouteDeclarationToModule` from a specific implementation (in this case using Router as the only parameter). I want to follow these patterns and enable this code to do the same kind of `action` with custom Modules (such as in my personal case NgxsModule)

I see value in this decoupling because these same patterns can be used to perform the same kind of actions to other Modules, not just Router, and enable users using `schematics` in their own projects, and potentially the angular-cli team, to better create the schematics they are working on.

For example if a user runs a generate ngxs schematic, the schematic can also add the State to the NgModule of choice.

alan-agius4:
Hi, privates APIs are added when needed, and removed once no longer needed in the internal schematics. Please be aware that also `@schematics/angular/utility/workspace` is also a private APIs and can change without warning even in patch versions. 

For reading and updating to the workspace please use the exports from methods exported from `@schematics/angular/utility` which is the **only** entry-point considered as public.

We definitely, want to add more utilities to the public API, but this requires a deeper analysis and design. 

If you really want to use a private utility it would be better to copy it in your project.

samali0996:
Hi Alan,

I understand your point, thanks for explaining.

I will take a look at the publicly available APIs, and will close this PR.

Will also take a look at schematic-related opened tickets in my free time and see if there’s anything I can work on!

PR #24847, opened 2023-03-13
feat(@angular-devkit/build-angular): support standalone app-shell generation

alan-agius4:
This commit adds support for generating an app-shell for a standalone application.

The `main.server.ts`, will need to export a bootstrapping function that returns a `Promise<ApplicationRef>`.

Example
```ts
export default () => bootstrapApplication(AppComponent, {
  providers: [
    importProvidersFrom(ServerModule),
    provideRouter([{ path: 'shell', component: AppShellComponent }]),
  ],
});
```

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/app-shell/render-worker.ts
+++ b/packages/angular_devkit/build_angular/src/builders/app-shell/render-worker.ts
@@ -6,8 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import type { Type } from '@angular/core';
-import type * as platformServer from '@angular/platform-server';
+import type { ApplicationRef, StaticProvider, Type } from '@angular/core';
+import type { renderApplication, renderModule, ɵSERVER_CONTEXT } from '@angular/platform-server';
 import assert from 'node:assert';
 import { workerData } from 'node:worker_threads';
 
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/builders/app-shell/render-worker.ts
```diff
@@ -43,29 +60,44 @@ interface RenderRequest {
...
+    AppServerModule,
+    renderModule,
+    renderApplication,
+    default: bootstrapAppFn,
+  } = (await import(serverBundlePath)) as ServerBundleExports;
 
-  assert(renderModule, `renderModule was not exported from: ${serverBundlePath}.`);
-  assert(AppServerModule, `AppServerModule was not exported from: ${serverBundlePath}.`);
   assert(ɵSERVER_CONTEXT, `ɵSERVER_CONTEXT was not exported from: ${serverBundlePath}.`);
+  assert(renderModule, `renderModule was not exported from: ${serverBundlePath}.`);
```
clydin, line 72: Is `renderModule` needed if `renderApplication` is present?

alan-agius4: Technically no, although it’s always exported from our builder.

But, thinking about it probably it best to move the asserts in the if blocks.

clydin (approved):

PR #24849, opened 2023-03-13
fix(@angular-devkit/build-angular): set public class fields as properties

alan-agius4:
Configure Babel to use `setPublicClassFields: true`. As when shipping Angular packages without `"useDefineForClassFields": false` will increase the bundle size of an ng-new app by ~2Kb when `useDefineForClassFields`  is not false. due to the additional `_defineProperty`, which will be added on every class property.

See: https://babeljs.io/docs/babel-plugin-proposal-class-properties

### DIFF
--- a/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
+++ b/packages/angular_devkit/build_angular/src/webpack/configs/common.ts
@@ -411,6 +411,13 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
             {
               loader: require.resolve('../../babel/webpack-loader'),
               options: {
+                assumptions: {
+                  // Use `setPublicClassFields: true` to avoid decrease bundle sizes
+                  // when targetting ES2022+ with `useDefineForClassFields: true`.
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/webpack/configs/common.ts
```diff
@@ -411,6 +411,13 @@ export async function getCommonConfig(wco: WebpackConfigOptions): Promise<Config
             {
               loader: require.resolve('../../babel/webpack-loader'),
               options: {
+                assumptions: {
+                  // Use `setPublicClassFields: true` to avoid decrease bundle sizes
+                  // when targetting ES2022+ with `useDefineForClassFields: true`.
+                  // This is because ervery property of the class will otherwise be wrapped in a `_defineProperty`.
+                  // This is not needed for TypeScript as TS itself will emit the right decleration of class properties.
```
clydin, line 418:
```suggestion
                  // This is not needed for TypeScript as TS itself will emit the right declaration of class properties.
```

clydin (approved):

PR #24879, opened 2023-03-17
build: update to TypeScript 5 and drop support for TypeScript 4.8

crisbeto:
Bumps up the dev dependency to TS 5 and narrows the required version to exclude 4.8.

BREAKING CHANGE:
* TypeScript 4.8 is no longer supported.


packages/angular_devkit/build_angular/package.json
```diff
@@ -79,7 +79,7 @@
     "ng-packagr": "^16.0.0-next.0",
     "protractor": "^7.0.0",
     "tailwindcss": "^2.0.0 || ^3.0.0",
-    "typescript": ">=4.8.2 <5.1"
+    "typescript": ">=4.9.3 <5.1"
```
alan-agius4, line 82: Should we also drop support for 4.9 in 3P since 5.0 is stable and this is a major version?

crisbeto: Because of the decorator changes and a lot of compiler APIs bring removed, I decided to keep support for 4.9 around a bit longer.

alan-agius4 (approved):

PR #24880, opened 2023-03-19
feat(@angular-devkit/build-angular): add CSP support for inline styles

crisbeto:
Companion change to https://github.com/angular/angular/pull/49444. Adds an HTML processor that finds the `ngCspNonce` attribute and copies its value to any inline `style` tags in the HTML. The processor runs late in the processing pipeline in order to pick up any `style` tag that might've been added by other processors (e.g. critical CSS).

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/index-file/style-nonce.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/style-nonce.ts
@@ -1,0 +1,62 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { htmlRewritingStream } from './html-rewriting-stream';
+
+/**
+ * Pattern matching the name of the Angular nonce attribute. Note that this is
+ * case-insensitive, because HTML attribute names are case-insensitive as well.
+ */
+const NONCE_ATTR_PATTERN = /ngCspNonce/i;
+
+/**
+ * Finds the `ngCspNonce` value and copies it to all inline `<style>` tags.
+ * @param html Markup that should be processed.
+ */
+export async function addStyleNonce(html: string): Promise<string> {
+  const nonce = await findNonce(html);
+
+  if (!nonce) {
+    return html;
+  }
+
+  const { rewriter, transformedContent } = await htmlRewritingStream(html);
+
+  rewriter.on('startTag', (tag) => {
+    if (tag.tagName === 'style' && !tag.attrs.some((attr) => attr.name === 'nonce')) {
+      tag.attrs.push({ name: 'nonce', value: nonce });
+    }
+
+    rewriter.emitStartTag(tag);
+  });
+
+  return transformedContent();
+}
+
+/** Finds the Angular nonce in an HTML string. */
+async function findNonce(html: string): Promise<string | null> {
+  // Inexpensive check to avoid parsing the HTML when we're sure there's no nonce.
+  if (!NONCE_ATTR_PATTERN.test(html)) {
+    return null;
+  }
+
+  const { rewriter, transformedContent } = await htmlRewritingStream(html);
+  let nonce: string | null = null;
+
+  rewriter.on('startTag', (tag) => {
+    const nonceAttr = tag.attrs.find((attr) => NONCE_ATTR_PATTERN.test(attr.name));
+    if (nonceAttr?.value) {
+      nonce = nonceAttr.value;
+      rewriter.stop(); // Stop parsing since we've found the nonce.
+    }
+  });
+
+  await transformedContent();
+
+  return nonce;
+}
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/utils/index-file/style-nonce.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { htmlRewritingStream } from './html-rewriting-stream';
+
+/**
+ * Pattern matching the name of the Angular nonce attribute. Note that the it is
+ * case-insensitive, because HTML attribute names are case-insesitive as well.
+ */
```
alan-agius4, line 14:
```suggestion
/**
 * Pattern matching the name of the Angular nonce attribute. Note that this is
 * case-insensitive, because HTML attribute names are case-insensitive as well.
 */
```

packages/angular_devkit/build_angular/src/utils/index-file/style-nonce.ts
```diff
@@ -0,0 +1,62 @@
...
+const NONCE_ATTR_PATTERN = /ngCspNonce/i;
```
alan-agius4, line 15:
```suggestion
const NONCE_ATTR_PATTERN = /^ngCspNonce$/i;
```

packages/angular_devkit/build_angular/src/utils/index-file/style-nonce.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  rewriter.on('startTag', (tag) => {
+    if (tag.tagName === 'style' && !tag.attrs.some((attr) => attr.name === 'nonce')) {
+      tag.attrs.push({ name: 'nonce', value: nonce });
+    }
+
+    rewriter.emitStartTag(tag);
+  });
+
+  return await transformedContent();
```
alan-agius4, line 38: Redundant await. 

```suggestion
  return transformedContent();
```

packages/angular_devkit/build_angular/src/utils/index-file/style-nonce.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (!NONCE_ATTR_PATTERN.test(html)) {
+    return null;
+  }
+
+  const { rewriter, transformedContent } = await htmlRewritingStream(html);
+  let nonce: string | null = null;
+
+  rewriter.on('startTag', (tag) => {
+    const nonceAttr = tag.attrs.find((attr) => NONCE_ATTR_PATTERN.test(attr.name));
+    if (nonceAttr && nonceAttr.value) {
```
alan-agius4, line 53: NIT
```suggestion
    if (nonceAttr?.value) {
```

alan-agius4 (changes requested):
I am not certain that creating a new plugin is the right direction. 

The plugin addresses only one use case (CSR), but it does not add nonce to the app-shell, which uses extended Critters directly. See: https://github.com/angular/angular-cli/blob/5a171ddff66ff366089616736baf7545fe44f570/packages/angular_devkit/build_angular/src/builders/app-shell/index.ts#L95-L107 Extending this class and doing the logic in here IMHO would be better as we’ll share the same logic and it would be easier to port this change to Universal https://github.com/angular/universal/blob/main/modules/common/tools/src/inline-css-processor.ts. 

I was under the impression, that we were going to add nonce support in Critters, as adding `nonce` to the `style` elements  is also not sufficient to make the page CSP complaint, crtitters adds `onload` events to `link` elements, as such we will need to modify critters to output this is a CSP complaint way.

Currently critters will change 
```html
<link rel="stylesheet" href="styles.css”>
```
to 
```html
<link rel="stylesheet" href="styles.css" media="print" onload="this.media='all'">
```

Now, we need to amend the above and make it CSP complaint when nonce is provided (Or we can do the same logic irrespective if CSP is enabled.)
```html
<link rel="stylesheet" href=“styles.css” ngAsyncStyle>
<script nonce=“..">
  document.querySelectorAll('link[ngAsyncStyle]’)
      .forEach(s => s.addEventListener(‘load’, () => s.media = ‘all'));
</script>
```

All the above, would be complex to do with this sort of plugin created, but should be more doable using by extended Critters.

packages/angular_devkit/build_angular/src/utils/index-file/style-nonce.ts
crisbeto: This isn't what we want, because the pattern is also used to check the entire HTML so we know not to parse if the nonce attribute isn't present.

crisbeto (commented):
The change I was planning to make to Critters was to allow for custom attributes to be specified and we can achieve the same without touching Critters at all using the changes in this PR. I can bake this logic into the Critters plugin, but I think that it makes sense to apply it to all inline `style` tags, not just the ones generated by Critters.

I think that the `link` use case would require a much larger change to Critters so for now I would rather focus only on the `style` tags.

alan-agius4:
> I can bake this logic into the Critters plugin, but I think that it makes sense to apply it to all inline style tags, not just the ones generated by Critters.

Fair enough.

crisbeto:
Comments have been addressed.

alan-agius4 (approved):

PR #24901, opened 2023-03-24
fix(@schematics/angular): remove compileComponents from component test schematic

cexbrayat:
Note: This is an updated version of #18316 after discussing with @atscott who thought we are now in a place where it could land

`compileComponents` is not necessary when using the CLI (as the templates are inlined) and just adds boilerplate code. So we can remove it from the test schematic and make it independent from `async/await` (only place we would have it in the CLI generated code, and in most Angular apps).

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Tests are generating with `compileComponent` calls that aren't needed with the CLI

## What is the new behavior?

The boilerplate is reduced to only keep what is really necessary

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/karma/tests/options/assets_spec.ts
+++ b/packages/angular_devkit/build_angular/src/builders/karma/tests/options/assets_spec.ts
@@ -23,7 +23,7 @@ describeBuilder(execute, KARMA_BUILDER_INFO, (harness) => {
             import { NgModule } from '@angular/core';
             import { HttpClientModule } from '@angular/common/http';
             import { AppComponent } from './app.component';
-    
+
             @NgModule({
               declarations: [
                 AppComponent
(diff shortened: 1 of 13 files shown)


dgp1130:
Apologies for blocking this PR the first time, I'm on board with the change now at least. :sweat_smile:

Just to note somewhere for future reference, removing `.compileComponents()` does make the generated schematic slightly less portable since it won't fetch `templateUrl` and `styleUrl` resources during testing. That's fine for CLI tests and `jest-preset-angular` (AFAICT) because they include transforms in the build process to inline such resources at build time (the compiler rewrites `templateUrl: './template.ng.html'` to `template: '<div>...</div>'`), so they don't need to be loaded at runtime in the test environment.

Aside from being an optimization, I think it's actually required regardless of `.compileComponents()`. `templateUrl` and `styleUrls` are both file-relative, but at runtime which file is requesting a given resource is lost. So there's no way to know what directory a reference to `./template.ng.html` actually wants. If we had design this to look like:

```typescript
@Component({
  templateUrl: new URL('./template.ng.html', import.meta.url),
  // ...
})
class MyComponent {}
```

Then this information would be kept at runtime and we could resolve resources correctly. Without it, I think there's enough ambiguity here to say that Angular just doesn't work without the inline resources transform. Since `compileComponents()` is only needed in situations where users don't have the inline resources transform and that use case is basically impossible anyways, then there's not much value in generating `.compileComponents()`.

cexbrayat:
@dgp1130 No hard feelings 😉. Thank you for considering it this time, and for the detailed explanation!

PR #24906, opened 2023-03-24
ci: disable windows job on PRs

alan-agius4:
Currently CLI's windows testing accounts for ~40% of all of our credit usage. This job is more of a sanity check and running only on release branches should suffice.


josephperrott (approved):
LGTM

PR #24920, opened 2023-03-28
feat(@schematics/angular): update app-shell schematic to support standalone applications     

alan-agius4:
This commit adds support to run `ng generate app-shell` in standalone applications.


Brocco:
LGTM

PR #24951, opened 2023-04-04
feat(@schematics/angular): add support to add service worker to standalone application

alan-agius4:
This commit adds support to generate a service worker in a standalone application.


alan-agius4:
**Note:** CI failure is unrelated.

packages/schematics/angular/private/standalone.ts
```diff
@@ -163,14 +163,15 @@ export function addModuleImportToStandaloneBootstrap(
  * @param functionName Name of the function that should be called.
  * @param importPath Path from which to import the function.
  * @param args Arguments to use when calling the function.
+ * @returns The file path that the provider was added too.
```
clydin, line 166:
```suggestion
 * @returns The file path that the provider was added to.
```

clydin (approved):

PR #25013, opened 2023-04-13
perf(@angular/cli): register CLI commands lazily

alan-agius4:
Currently there is a lot of overhead coming from requiring external modules when registering commands such as `ng update` and `ng add`. This is because these commands do not lazily require all the modules causes the resolution of unneeded packages to be part of the critical path.

With this change we "require” only the command that we we need to execute, which reduce the number of node modules resolutions in the critical path.


packages/angular/cli/src/command-builder/command-runner.ts
```diff
@@ -168,3 +137,23 @@ export async function runCommand(args: string[], logger: logging.Logger): Promis
...
+  if (commandName in RootCommands) {
+    commands.push(RootCommands[commandName as CommandNames]);
+  } else if (commandName in RootCommandsAliases) {
+    commands.push(RootCommandsAliases[commandName]);
+  } else {
+    // Unknown command, register every possible command.
+    Object.values(RootCommands).forEach((c) => commands.push(c));
+  }
+
+  return Promise.all(commands.map((command) => command.factory()));
```
clydin, line 158: nit: Could potentially put the `.then((m) => m.default)` here to avoid repeating it in the factory fields.

clydin (approved):
Nice.

PR #25019, opened 2023-04-13
ci: update CI `.bazelrc` to better support CI systems.

dgp1130:
A number of small changes with comments here. The main one is that removing `--noshow_progress` will make Bazel print progress updates which tells Circle CI that test execution is still ongoing and avoids test failures due to lack of output.

The other notable trade-offs being made here are:
1. [`--keep_going`](https://bazel.build/docs/user-manual#keep-going) means `bazel test` will run all tests report all failures, which could be slower and noisier than reporting just the first failure but is more complete.
2. [`--build_tests_only`](https://bazel.build/docs/user-manual#build-tests-only) means only tests and binaries will be built, so any library targets not tested could have build errors hidden. Any such important targets should either be tested or use Skylib's `build_test()`.

The UI options mostly came from [this talk](https://www.youtube.com/watch?v=j332WfhNAFg).

### DIFF
--- a/.circleci/bazel.common.rc
+++ b/.circleci/bazel.common.rc
@@ -4,8 +4,15 @@
 # Echo all the configuration settings and their source
 build --announce_rc
 
-# Don't be spammy in the logs
-build --noshow_progress
+# Print extra information for build failures to help with debugging.
+build --verbose_failures
+
+# Show progress so CI doesn't appear to be stuck, but rate limit to avoid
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):
LGTM

devversion (approved):
LGTM. I'm not 100% sure I like build tests only, but that's fine :)

PR #25030, opened 2023-04-17
Various Esbuild improvements

dgp1130:
## PR Checklist

## PR Type

- [X] Feature

Issue Number: N/A

## What is the new behavior?

* `browser-esbuild` options are now able to differ slightly from the existing `browser` builder. This is to support adding additional functionality not currently supported by the `browser` builder, staying compatible with the existing builder is still a primary goal of the project.
* `browser-esbuild` now supports an additional `entryPoints` option in its `InfrastructureSettings`.
    * This required an update to builder test harnesses in order to effectively test.
* `main` and `entryPoints` now support absolute file paths. Absolute files paths are always output with the same basename in the root of the output directory.
* `browser-esbuild` now allows `index` to be optional. It already supported `index` not being set, this just updates the schema.
* `browser-esbuild` now supports an `outExtension` option to configure the output file extension.

- [X] No


dgp1130:
The formatter seems to be conflicting with the linter here? Linter complains that an error message exceeds the line length, but if I break it into two lines, the formatter collapses them back on to one. Not sure how to handle that?

alan-agius4:
> The formatter seems to be conflicting with the linter here? Linter complains that an error message exceeds the line length, but if I break it into two lines, the formatter collapses them back on to one. Not sure how to handle that?

You'd need to split the message into multiple strings.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
```diff
@@ -13,9 +13,16 @@
       }
```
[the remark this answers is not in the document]
dgp1130: Moved the new `entryPoints` option into `InfrastructureSettings` of `browser-esbuild`, so it's out of the schema. I opted to leave the optional index in the schema as @clydin felt that was at least somewhat appropriate.

Unfortunately this got tricky to test because Architect doesn't have knowledge of this extra option, and I need to use the builder harness to get a valid context. I ended up adding just enough support for the builder harness to pass through this extra `InfrastructureSettings` (open to better naming suggestions) option into the builder. The actual production builder APIs still don't support passing through this value as that got more complicated and affected more of the public API which I would rather stay decoupled from here. Since we're calling `browser-esbuild` outside of Architect anyways, that functionality isn't actually needed yet.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
```diff
@@ -397,6 +402,10 @@
             }
           },
           "required": ["input"]
+        },
+        {
+          "type": "null",
```
clydin, line 407:
```suggestion
          "const": false,
```
Will allow for `--no-index` on the command line.

clydin (approved):

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
dgp1130: Done, thanks for reminding me.

dgp1130:
I also added another commit to support absolute file paths in entry points. Hopefully nothing too objectionable there, can you take another quick look @clydin?

goldens/public-api/angular_devkit/architect/index.md
```diff
@@ -49,8 +49,8 @@ export interface BuilderContext {
 }
 
 // @public
-export interface BuilderHandlerFn<A> {
-    (input: A, context: BuilderContext): BuilderOutputLike;
+export interface BuilderHandlerFn<Options, InfrastructureSettings = {}> {
```
alan-agius4, line 52: While, I understand why we need `infrastructureSettings` since we want to override certain "unknown" options when running the builder.

Example
```ts
  const browserTargetRun = await context.scheduleTarget(browserTarget);
```

I think there are a couple of things that we need to understand and agree on here. 
- The new test builder will be coupled with the implementation of the browser builder. In this case the CLI is opinionated that users **cannot** use any other builder if you want to use the new test runner, or if they want, the new builder would need to support the same options that the in-build Angular CLI builder supports. 
- This will increase improbabilities which custom builders and also make it almost impossible to use the new test builder with another browser builder, let's say NX.

IMO, I don't think we should couple this, and currently the test builder is not coupled with the browser builder API. Example, currently users can use esbuild/rspack/webpack to build their application and use webpack to run their tests through karma.

I don't want to block this PR, but I feel like this is a discussion that we need to have, as if we don't want to highly couple ourselves our browser builder, which would make it harder for users to use 3P builders etc.

If we are going with pre-bundling, I do believe that we should not invoke the builder through architect, but rather invoke the internal esbuild method, and the reason for this is as by using the architect builder directly we are actually increasing complexity, coupling the browser and test builder for just re-using a couple of options.  

**Option to re-use**
- `fileReplacements`- which I am iffy if we should add this in the test builder.
- `inlineStyleLanguage` - if we want to support component styles
- `stylePreprocessorOptions`- if we want to support component styles
- `externalDependencies` - not sure if this is needed for testing purposes, as this will likely cause problems.

**Options to override**
- `sourceMap` - this options also depends on the code-coverage value as when this is set to true sourcemaps also need to be true
- `watch` - this is based on if the test runner is running in watch mode.
- `aot` - tests always in JIT
- `index` - not needed
- `dist`
- `main` - overridden to entrypoints.
- `tsConfig` - tsconfig.spec.ts
- `progress` - based on the test builder config
- `serviceWorker` - always disabled
- `scripts` - not needed for tests
- `styles` - not needed for tests
- `webWorkerTsConfig` - not supported for tests
- `budgets` - disabled for tests
- `assets` - not needed for tests.
- `polyfills` - are different for test

Eventually more and more options will need to be overridden.

alan-agius4 (changes requested):
It looks like there is a potential breaking change and I left some feedback comments.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
```diff
@@ -216,6 +216,11 @@
       "type": "string",
       "description": "The full path for the new output directory, relative to the current workspace.\nBy default, writes output to a folder named dist/ in the current project."
     },
+    "outExtension": {
```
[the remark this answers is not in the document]
alan-agius4: Following todays design review this should not be necessary as the option can be provided directly when run the bundle function.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
```diff
@@ -14,7 +14,7 @@
     },
     "main": {
       "type": "string",
-      "description": "The full path for the main entry point to the app, relative to the current workspace."
+      "description": "The full path for the main entry point to the app. May be a relative or absolute path. If given a relative path, it is resolved relative to the current workspace."
```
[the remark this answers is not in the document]
dgp1130: The motivation here is to support injecting code by the build process. In particular for testing, we need to inject `TestBed.initTestEnvironment()` which will not be within the user's workspace. I agree in general, users should not bundle code from outside the workspace, but this is one case where we effectively need to inject a file from `node_modules/...` which may not be in the workspace (ex. Yarn PnP).

I could probably limit this to `entryPoints` which would be restricted to programmatic invocations of ESBuild-er, that way `main` could not be abused in this manner and hopefully limit "bad" usages in the wild, though it would also be confusing that `main` and `entryPoints` resolve files differently. In that case, I would probably want to ban absolute paths altogether from `main` instead of treating them like relative paths.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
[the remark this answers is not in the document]
dgp1130: Agree that we could probably always generate `*.mjs` files. I expect many users won't like the "aesthetics" of it, but that's a pretty minor complaint. Another option is to move this flag to `InfrastructureSettings` to make it a little more "private", as I don't think it actually needs to be in the public schema. WDYT @clydin?

packages/angular_devkit/build_angular/src/builders/browser-esbuild/options.ts
```diff
@@ -20,6 +20,20 @@ import { Schema as BrowserBuilderOptions, OutputHashing } from './schema';
 
 export type NormalizedBrowserOptions = Awaited<ReturnType<typeof normalizeOptions>>;
 
+/** Internal infrastructure settings hidden from builder schema but available when invoked programmatically. */
+export interface InfrastructureSettings {
+  /** Whether or not to write output files to disk. */
+  write?: boolean;
+
+  /**
```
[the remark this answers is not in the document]
dgp1130: To follow up here, we discussed this somewhat in the design review today and generally agree that invoking the builder outside Architect is both required and undesirable as it produces a direct dependency on `browser-esbuild` specifically which should not be necessary. We decided that this is ok to start with and we should look into decoupling test builders from `browser-esbuild` later as we determine the right way of implementing them in Architect.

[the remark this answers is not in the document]
alan-agius4: Didn't we agree that since Architect is limiting we we should not using it and instead use the same pattern as the dev-server? 

https://github.com/angular/angular-cli/blob/866263b21e71d03800ae91033e951f9407ad0547/packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts#L61

Nit: In general, I am not a big fan of adding not infra options such as entrypoint to the infrastructure options, but for now it's okay.  

I think where I'd like the builder to head is that there is a method which is exposed irrespective of the browser-builder options something along this lines. Where the latter would be invoked from "certain" builders that need custom logic not needed in browser builder. 

```ts
async function* buildEsbuildBrowser(
  userOptions: BrowserBuilderOptions,
  context: BuilderContext,
  infrastructureSettings?: {
    write?: boolean;
  },
): AsyncIterable<
  BuilderOutput & {
    outputFiles?: OutputFile[];
    assetFiles?: { source: string; destination: string }[];
  }
> {

  return build(userOptions, context, infrastructureSettings)
}

interface BuildOptions extends BrowserBuilderOptions {
  entryPoints: Record<string, string>,
  fileExtension: 'mjs'|'js',
  virtualModules: Record<string, string>,
}

export async function* buildEsbuildBrowser(
  options: NormalizedBrowserOptions,
  context: BuilderContext,
  infrastructureSettings?: {
    write?: boolean;
  },
): AsyncIterable<
  BuilderOutput & {
    outputFiles?: OutputFile[];
    assetFiles?: { source: string; destination: string }[];
  }
> {

  //

}
```

packages/angular_devkit/build_angular/src/builders/browser-esbuild/options.ts
```diff
@@ -20,6 +20,20 @@ import { Schema as BrowserBuilderOptions, OutputHashing } from './schema';
...
+   * Entry points to use for the compilation. Incompatible with `main`, which must not be provided. May be relative or absolute paths.
+   * If given a relative path, it is resolved relative to the current workspace and will generate an output at the same relative location
+   * in the output directory. If given an absolute path, the output will be generated in the root of the output directory with the same base
+   * name.
+   */
+  entryPoints?: Set<string>;
```
alan-agius4, line 34: I am okay with this for now as a workaround, but I am not a big fan of additional non infra related option in the infra settings.

packages/angular_devkit/architect/src/api.ts
```diff
@@ -300,16 +300,22 @@ async function handleAsyncIterator<T>(
 /**
  * A builder handler function. The function signature passed to `createBuilder()`.
  */
-export interface BuilderHandlerFn<A> {
+export interface BuilderHandlerFn<Options, InfrastructureSettings = {}> {
```
alan-agius4, line 303: Following yesterdays discussion, we should not invoke the builder using architect, instead we should use the same pattern like we currently use in the dev-server. Thus, I think that changes the fact that we don't need to add this a public API.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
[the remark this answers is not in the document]
alan-agius4: > The motivation here is to support injecting code by the build process. In particular for testing, we need to inject TestBed.initTestEnvironment() which will not be within the user's workspace. I agree in general, users should not bundle code from outside the workspace, but this is one case where we effectively need to inject a file from node_modules/... which may not be in the workspace (ex. Yarn PnP).

I think for the time being as a temporary solution, but we definitly need to better approach for the in built `test.ts`. While we could get away with bundling a file in `node_module` I have doubts that it will solve Yarn PNP causes as this could cause module resolutions errors, because esbuild will try to resolve `@angular/` from in `@angular-devkit/build-angular` which is incorrect as it should be resolved from the users workspace root. 

Eventually, we should explore virtual modules or something like we currently have for [webpack](https://github.com/angular/angular-cli/blob/866263b21e71d03800ae91033e951f9407ad0547/packages/angular_devkit/build_angular/src/builders/karma/index.ts#L242-L260), as something similar as this will also be needed for SSR in `16.1`.

> I would probably want to ban absolute paths altogether from main instead of treating them like relative paths.

I agree with this, but in this case, I think we need to do this for all builders options as otherwise it will be confused that in some cases we treating absolute like relative paths and sometimes not. Also, this sounds like a thing we'd want to do in a major with a migration due to the breaking nature of it.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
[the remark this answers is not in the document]
alan-agius4: My 2 cents would be to use always `.mjs` for simplicity. I definitely don't think that we should expose to the users. I see that this option can become confusing to us pretty soon in (16.1) when we want esbuilder to generate both `.js` and `.mjs` for the same run. (`.mjs` for the server and `.js` for the browser).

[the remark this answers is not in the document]
clydin: I'd prefer we stick with `.js` since its the "standard" file extension for JavaScript files.  `.mjs` is a Node.js invention and is otherwise unrelated to the ECMAScript specification and browsers in general.
Changing the file extension default could have unknown effects on downstream tools and behavior (secondary processing, deployment, server configurations, mime-type entry setup, etc.).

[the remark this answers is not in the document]
clydin: For an additional option, the build system could potentially be setup to not need to output `.mjs` files on the server. A `package.json` file containing the following will cause Node.js to execute ESM `.js` files appropriately:
```
{
  "type": "module"
}
```
This could be placed in the output directory of server/test code as an auto-generated asset as needed.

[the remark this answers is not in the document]
alan-agius4: Good points @clydin

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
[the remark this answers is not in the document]
dgp1130: As mentioned in another comment, I dropped all the changes to `InfrastructureSettings` and handled them via a separate `InternalOptions` type exposed from a new `buildEsbuildBrowserInternal()` entry point.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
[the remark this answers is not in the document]
dgp1130: Moved `outExtension` to `InternalOptions` so it is hidden from public API, but left `*.js` as the default. `package.json` is definitely another viable option, though since it's also a Node-ism I'm not sure that's strictly better.

Hopefully this should be good enough for now.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/options.ts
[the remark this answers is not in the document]
dgp1130: Chatted offline to make sure I was understanding Alan's point here. The plan is to directly invoke ESBuild-er outside of Architect like Alan suggested. The concern here is that `InfrastructureSettings` is not really intended to represent internal-only options like I was using it for.

Instead, I refactored things to create a new `buildEsbuildBrowserInternal()` (I'm open to better naming suggestions) function as a separate entry point. This function also includes `entryPoints` and `outExtension` as internal options without putting them in `InfrastructureSettings` or requiring Architect to pass through that option. With that, I was able to revert all the Architect harness changes and simplify things there.

Hopefully this should address your concerns, but let me know if I'm still not quite understanding it.

packages/angular_devkit/architect/src/api.ts
dgp1130: Yep, reverted all the Architect changes in favor of using a `buildEsbuildBrowserInternal()` entry point with the extra options so we don't need to change `InfrastructureSettings`.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/options.ts
dgp1130: Moved to `InternalOptions` distinct from `InfrastructureSettings`.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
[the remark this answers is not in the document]
dgp1130: Agreed that virtual modules would make this better and there is a potential package resolution issue with `pnpm`. This should be good enough for now and we can update it to use virtual modules once we have a better idea of what that would look like.

For now, I limited absolute paths to just the new internal `entryPoints` option. `main` is mostly unaffected (only new behavior is that it throws when an empty string is passed in). That should limit the blast radius of this particular change until we can leverage virtual modules here.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/options.ts
```diff
@@ -204,6 +221,75 @@ export async function normalizeOptions(
...
+        throw new Error(
+          `\`${existingEntryPointPath}\` and \`${entryPointPath}\` both output to the same location \`${entryPointName}\`.` +
+            ' Rename or move one of the files to fix the conflict.',
+        );
+      }
+
+      entryPointPaths.set(entryPointName, entryPointPath);
+    }
+
+    return Object.fromEntries(entryPointPaths.entries());
```
alan-agius4, line 289: NIT: using a `Map` here is redundant. Consider using a key-value pair instead (same time complexity). 
This will also remove the need to convert the map to a key-value pair before returing.

```suggestion
    const entryPointPaths: Record<string, string> = {};
    for (const entryPoint of entryPoints) {
      const parsedEntryPoint = path.parse(entryPoint);

      // Use the input file path without an extension as the "name" of the entry point dictating its output location.
      // Relative entry points are generated at the same relative path in the output directory.
      // Absolute entry points are always generated with the same file name in the root of the output directory. This includes absolute
      // paths pointing at files actually within the workspace root.
      const entryPointName = path.isAbsolute(entryPoint)
        ? parsedEntryPoint.name
        : path.join(parsedEntryPoint.dir, parsedEntryPoint.name);

      // Get the full file path to the entry point input.
      const entryPointPath = path.isAbsolute(entryPoint)
        ? entryPoint
        : path.join(workspaceRoot, entryPoint);

      // Check for conflicts with previous entry points.
      const existingEntryPointPath = entryPointPaths[entryPointName];
      if (existingEntryPointPath) {
        throw new Error(
          `\`${existingEntryPointPath}\` and \`${entryPointPath}\` both output to the same location \`${entryPointName}\`.` +
            ' Rename or move one of the files to fix the conflict.',
        );
      }

      entryPointPaths[entryPointName] = entryPointPath;
    }

    return entryPointPaths;
```

packages/angular_devkit/build_angular/src/builders/browser-esbuild/tests/options/entry-points_spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  let tempDir!: string;
+
+  beforeEach(async () => {
+    tempDir = await fs.mkdtemp(path.join(tmpdir(), 'angular-cli-e2e-browser-esbuild-main-spec-'));
+  });
+
+  afterEach(async () => {
+    await fs.rm(tempDir, { recursive: true });
+  });
+
```
alan-agius4, line 25:
```suggestion
```

packages/angular_devkit/build_angular/src/builders/browser-esbuild/tests/options/entry-points_spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  describe('Option: "entryPoints"', () => {
+    it('provides multiple entry points', async () => {
+      await harness.writeFiles({
```
alan-agius4, line 28: Can you format a little bit?

packages/angular_devkit/build_angular/src/builders/browser-esbuild/tests/options/entry-points_spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    it('throws when `main` is omitted and an empty `entryPoints` Set is provided', async () => {
+      harness.useTarget('build', {
+        ...BASE_OPTIONS,
+        main: undefined,
+        entryPoints: new Set(),
+      });
+
+      const { result, error } = await harness.executeOnce();
+      expect(result?.success).toBeUndefined();
```
alan-agius4, line 61:
```suggestion
      expect(result).toBeUndefined();
```

packages/angular_devkit/build_angular/src/builders/browser-esbuild/tests/options/entry-points_spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    it('throws when provided with a `main` option', async () => {
+      harness.useTarget('build', {
+        ...BASE_OPTIONS,
+        main: 'src/main.ts',
+        entryPoints: new Set(['src/entry.ts']),
+      });
+
+      const { result, error } = await harness.executeOnce();
+      expect(result?.success).toBeUndefined();
```
alan-agius4, line 74:
```suggestion
      expect(result).toBeUndefined();
```

packages/angular_devkit/build_angular/src/builders/browser-esbuild/tests/options/entry-points_spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      await fs.writeFile(entry2, `console.log('entry2');`);
+
+      harness.useTarget('build', {
+        ...BASE_OPTIONS,
+        main: undefined,
+        entryPoints: new Set([entry1, entry2]),
+      });
+
+      const { result, error } = await harness.executeOnce();
+      expect(result?.success).toBeUndefined();
```
alan-agius4, line 113:
```suggestion
      expect(result).toBeUndefined();
```

alan-agius4 (approved):
LGTM, just a couple of NITs.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
```diff
@@ -595,6 +597,28 @@ export async function* buildEsbuildBrowser(
     outputFiles?: OutputFile[];
     assetFiles?: { source: string; destination: string }[];
   }
+> {
+  yield* buildEsbuildBrowserInternal(userOptions, context, infrastructureSettings);
```
clydin, line 601: nit: Since this function is only passing through the internal one, it could just be a normal function that directly returns the iterator created by `buildEsbuildBrowserInternal`.

clydin (approved):

packages/angular_devkit/build_angular/src/builders/browser-esbuild/options.ts
dgp1130: Good point, usually I prefer `Map`, but since we ultimately need to return an object, there's not much point in it here.

packages/angular_devkit/build_angular/src/builders/browser-esbuild/tests/options/entry-points_spec.ts
dgp1130: Updated to be a little closer to how this looks in other files, hopefully that's what you meant here?

packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
dgp1130: Good point, simplified to `return buildEsbuildBrowserInternal(/* ... */);`.

PR #25055, opened 2023-04-24
Initial experimental Jest support

dgp1130:
## PR Checklist

## PR Type

- [X] Feature

## What is the new behavior?

This PR adds initial experimental [Jest](https://jestjs.io/) support. This is _experimental_ and not ready for production use yet. For now, it it sufficient to pass the tests generated by the `ng new` application and we will continue to expand on this design going forward.

Enable the Jest runner by updating the `test` builder in your `angular.json` to be:

```jsonc
{
  "projects": {
    "my-app": {
      "architect": {
        // ...
        "test": {
          "builder": "@angular-devkit/build-angular:jest"
        }
      }
    }
  }
}
```

The current approach is to pre-build application tests with the `browser-esbuild` Angular builder under the hood, and then run Jest on the JavaScript outputs. We'll continue to experiment with this architecture and others to see what the best path forward is in terms of performance, maintenance, and developer ergonomics.

- [X] No

### DIFF
--- a/packages/angular/cli/lib/config/workspace-schema.json
+++ b/packages/angular/cli/lib/config/workspace-schema.json
@@ -360,6 +360,7 @@
                       "@angular-devkit/build-angular:dev-server",
                       "@angular-devkit/build-angular:extract-i18n",
                       "@angular-devkit/build-angular:karma",
+                      "@angular-devkit/build-angular:jest",
                       "@angular-devkit/build-angular:protractor",
                       "@angular-devkit/build-angular:server",
                       "@angular-devkit/build-angular:ng-packagr"
@@ -512,6 +513,28 @@
                   "type": "object",
                   "additionalProperties": {
                     "$ref": "../../../../angular_devkit/build_angular/src/builders/karma/schema.json"
+                  }
+                }
+              }
+            },
+            {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "builder": {
+                  "const": "@angular-devkit/build-angular:jest"
+                },
+                "defaultConfiguration": {
+                  "type": "string",
+                  "description": "A default named configuration to use when a target configuration is not provided."
+                },
+                "options": {
+                  "$ref": "../../../../angular_devkit/build_angular/src/builders/jest/schema.json"
+                },
+                "configurations": {
+                  "type": "object",
+                  "additionalProperties": {
+                    "$ref": "../../../../angular_devkit/build_angular/src/builders/jest/schema.json"
                   }
                 }
               }
--- a/packages/angular_devkit/build_angular/BUILD.bazel
+++ b/packages/angular_devkit/build_angular/BUILD.bazel
@@ -39,6 +39,11 @@ ts_json_schema(
 )
 
 ts_json_schema(
+    name = "jest_schema",
+    src = "src/builders/jest/schema.json",
+)
+
+ts_json_schema(
     name = "karma_schema",
     src = "src/builders/karma/schema.json",
 )
@@ -79,6 +84,7 @@ ts_library(
         "//packages/angular_devkit/build_angular:src/builders/browser/schema.ts",
         "//packages/angular_devkit/build_angular:src/builders/dev-server/schema.ts",
         "//packages/angular_devkit/build_angular:src/builders/extract-i18n/schema.ts",
+        "//packages/angular_devkit/build_angular:src/builders/jest/schema.ts",
         "//packages/angular_devkit/build_angular:src/builders/karma/schema.ts",
         "//packages/angular_devkit/build_angular:src/builders/ng-packagr/schema.ts",
         "//packages/angular_devkit/build_angular:src/builders/protractor/schema.ts",
@@ -267,7 +273,7 @@ ts_library(
             "src/**/tests/*.ts",
         ],
         exclude = [
-            "src/testing/**/*_spec.ts",
+            "src/**/*_spec.ts",
         ],
     ),
     data = glob(["test/**/*"]),
@@ -345,6 +351,12 @@ LARGE_SPECS = {
         "shards": 10,
         "extra_deps": [
             "@npm//buffer",
+        ],
+    },
+    "jest": {
+        "extra_deps": [
+            "@npm//glob",
+            "@npm//@types/glob",
         ],
     },
 }
--- a/packages/angular_devkit/build_angular/BUILD.bazel
+++ b/packages/angular_devkit/build_angular/BUILD.bazel
@@ -96,6 +96,7 @@ ts_library(
             "builders.json",
             "src/**/schema.json",
             "src/**/*.js",
+            "src/**/*.mjs",
             "src/**/*.html",
         ],
     ),
--- a/packages/angular_devkit/build_angular/src/builders/jest/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/jest/index.ts
@@ -7,12 +7,18 @@
  */
 
 import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
+import { execFile as execFileCb } from 'child_process';
+import * as path from 'path';
+import { promisify } from 'util';
+import { colors } from '../../utils/color';
 import { buildEsbuildBrowserInternal } from '../browser-esbuild';
 import { BrowserEsbuildOptions } from '../browser-esbuild/options';
 import { OutputHashing } from '../browser-esbuild/schema';
 import { normalizeOptions } from './options';
 import { Schema as JestBuilderSchema } from './schema';
 import { findTestFiles } from './test-files';
+
+const execFile = promisify(execFileCb);
 
 /** Main execution function for the Jest builder. */
 export default createBuilder(
@@ -22,11 +28,35 @@ export default createBuilder(
     );
 
     const options = normalizeOptions(schema);
-    const testFiles = await findTestFiles(options, context.workspaceRoot);
     const testOut = 'dist/test-out'; // TODO(dgp1130): Hide in temp directory.
 
+    // Verify Jest installation and get the path to it's binary.
+    // We need to `node_modules/.bin/jest`, but there is no means to resolve that directly. Fortunately Jest's `package.json` exports the
+    // same file at `bin/jest`, so we can just resolve that instead.
+    const jest = resolveModule('jest/bin/jest');
+    if (!jest) {
+      return {
+        success: false,
+        // TODO(dgp1130): Display a more accurate message for non-NPM users.
+        error:
+          'Jest is not installed, most likely you need to run `npm install jest --save-dev` in your project.',
+      };
+    }
+
+    // Verify that JSDom is installed in the project.
+    const environment = resolveModule('jest-environment-jsdom');
+    if (!environment) {
+      return {
+        success: false,
+        // TODO(dgp1130): Display a more accurate message for non-NPM users.
+        error:
+          '`jest-environment-jsdom` is not installed. Install it with `npm install jest-environment-jsdom --save-dev`.',
+      };
+    }
+
     // Build all the test files.
-    return await build(context, {
+    const testFiles = await findTestFiles(options, context.workspaceRoot);
+    const buildResult = await build(context, {
       entryPoints: testFiles,
       tsConfig: options.tsConfig,
       polyfills: options.polyfills,
@@ -44,6 +74,53 @@ export default createBuilder(
         vendor: false,
       },
     });
+    if (!buildResult.success) {
+      return buildResult;
+    }
+
+    // Execute Jest on the built output directory.
+    const jestProc = execFile('node', [
+      '--experimental-vm-modules',
+      jest,
+
+      `--rootDir=${testOut}`,
+      '--testEnvironment=jsdom',
+
+      // TODO(dgp1130): Enable cache once we have a mechanism for properly clearing / disabling it.
+      '--no-cache',
+
+      // Run basically all files in the output directory, any excluded files were already dropped by the build.
+      `--testMatch=${path.join('<rootDir>', '**', '*.mjs')}`,
+
+      // Load polyfills before each test, and don't run them directly as a test.
+      `--setupFilesAfterEnv=${path.join('<rootDir>', 'polyfills.mjs')}`,
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'polyfills\\.mjs')}`,
+
+      // Skip shared chunks, as they are not entry points to tests.
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'chunk-.*\\.mjs')}`,
+
+      // Optionally enable color.
+      ...(colors.enabled ? ['--colors'] : []),
+    ]);
+
+    // Stream test output to the terminal.
+    jestProc.child.stdout?.on('data', (chunk) => {
+      context.logger.info(chunk);
+    });
+    jestProc.child.stderr?.on('data', (chunk) => {
+      // Write to stderr directly instead of `context.logger.error(chunk)` because the logger will overwrite Jest's coloring information.
+      process.stderr.write(chunk);
+    });
+
+    try {
+      await jestProc;
+    } catch (error) {
+      // No need to propagate error message, already piped to terminal output.
+      // TODO(dgp1130): Handle process spawning failures.
+      return { success: false };
+    }
+
+    return { success: true };
   },
 );
 
@@ -62,5 +139,14 @@ async function build(
       success: false,
       error: (err as Error).message,
     };
+  }
+}
+
+/** Safely resolves the given Node module string. */
+function resolveModule(module: string): string | undefined {
+  try {
+    return require.resolve(module);
+  } catch {
+    return undefined;
   }
 }
--- a/packages/angular_devkit/build_angular/src/builders/jest/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/jest/index.ts
@@ -1,0 +1,66 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
+import { buildEsbuildBrowserInternal } from '../browser-esbuild';
+import { BrowserEsbuildOptions } from '../browser-esbuild/options';
+import { OutputHashing } from '../browser-esbuild/schema';
+import { normalizeOptions } from './options';
+import { Schema as JestBuilderSchema } from './schema';
+import { findTestFiles } from './test-files';
+
+/** Main execution function for the Jest builder. */
+export default createBuilder(
+  async (schema: JestBuilderSchema, context: BuilderContext): Promise<BuilderOutput> => {
+    context.logger.warn(
+      'NOTE: The Jest builder is currently EXPERIMENTAL and not ready for production use.',
+    );
+
+    const options = normalizeOptions(schema);
+    const testFiles = await findTestFiles(options, context.workspaceRoot);
+    const testOut = 'dist/test-out'; // TODO(dgp1130): Hide in temp directory.
+
+    // Build all the test files.
+    return await build(context, {
+      entryPoints: testFiles,
+      tsConfig: options.tsConfig,
+      polyfills: options.polyfills,
+      outputPath: testOut,
+      aot: false,
+      index: null,
+      outputHashing: OutputHashing.None,
+      outExtension: 'mjs', // Force native ESM.
+      commonChunk: false,
+      optimization: false,
+      buildOptimizer: false,
+      sourceMap: {
+        scripts: true,
+        styles: false,
+        vendor: false,
+      },
+    });
+  },
+);
+
+async function build(
+  context: BuilderContext,
+  options: BrowserEsbuildOptions,
+): Promise<BuilderOutput> {
+  try {
+    for await (const _ of buildEsbuildBrowserInternal(options, context)) {
+      // Nothing to do for each event, just wait for the whole build.
+    }
+
+    return { success: true };
+  } catch (err) {
+    return {
+      success: false,
+      error: (err as Error).message,
+    };
+  }
+}
--- a/packages/angular_devkit/build_angular/src/builders/jest/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/jest/index.ts
@@ -56,10 +56,13 @@ export default createBuilder(
 
     // Build all the test files.
     const testFiles = await findTestFiles(options, context.workspaceRoot);
+    const jestGlobal = path.join(__dirname, 'jest-global.mjs');
+    const initTestBed = path.join(__dirname, 'init-test-bed.mjs');
     const buildResult = await build(context, {
-      entryPoints: testFiles,
+      // Build all the test files and also the `jest-global` and `init-test-bed` scripts.
+      entryPoints: new Set([...testFiles, jestGlobal, initTestBed]),
       tsConfig: options.tsConfig,
-      polyfills: options.polyfills,
+      polyfills: options.polyfills ?? ['zone.js', 'zone.js/testing'],
       outputPath: testOut,
       aot: false,
       index: null,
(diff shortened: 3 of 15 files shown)


packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      // Skip shared chunks, as they are not entry points to tests.
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'chunk-.*\\.mjs')}`,
+    ]);
+
+    // Stream test output to the terminal.
+    jestProc.child.stdout?.on('data', (chunk) => {
+      context.logger.info(chunk);
+    });
+    jestProc.child.stderr?.on('data', (chunk) => {
+      context.logger.info(chunk);
```
alan-agius4, line 117: Nit: This will cause stderr to be redirected to stdout
```suggestion
      context.logger.error(chunk);
```

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    });
+
+    try {
+      await jestProc;
+    } catch (error) {
+      // No need to propagate error message, already piped to terminal output.
```
alan-agius4, line 123: Nit: this is not always the cases. There are cases that when an error is throw during process creation.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+        styles: false,
+        vendor: true,
+      },
+    });
+    if (!buildResult.success) {
+      return buildResult;
+    }
+
+    // Execute Jest on the built output directory.
+    const jestProc = execFile('node', [
```
alan-agius4, line 82: I believe using `spawn` is a better here.

From my understanding `execFile` will cause the data to be buffered once the processes closes, this might cause users to think that the tests hanged as they don't get any feedback until all tests have executed. Unlike spawn where there is a continues data flow.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      `--setupFilesAfterEnv=${path.join('<rootDir>', 'jest-global.mjs')}`,
+      `--setupFilesAfterEnv=${path.join('<rootDir>', 'polyfills.mjs')}`,
+      `--setupFilesAfterEnv=${path.join('<rootDir>', 'init-test-bed.mjs')}`,
+
+      // Don't run any infrastructure files as tests, they are manually loaded where needed.
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'jest-global\\.mjs')}`,
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'polyfills\\.mjs')}`,
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'init-test-bed\\.mjs')}`,
+
+      // Skip shared chunks, as they are not entry points to tests.
```
alan-agius4, line 108: This should not be needed since jest will not pick up anything which does not match the [`testMatch`](https://jestjs.io/docs/configuration#testmatch-arraystring)

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    const jestProc = execFile('node', [
+      '--experimental-vm-modules',
+      jest,
+
+      `--rootDir=${testOut}`,
+      '--testEnvironment=jsdom',
+      '--color',
+
+      // Run basically all files in the output directory, any excluded files were already dropped by the build.
+      `--testMatch=${path.join('<rootDir>', '**', '*.mjs')}`,
```
alan-agius4, line 91: We should only run unit tests.

```suggestion
      `--testMatch='<rootDir>/**/*.spec.mjs'`,
```

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      // the environment for fake async to work correctly.
+      // Third, we initialize `TestBed`. This is dependent on fake async being set up correctly beforehand.
+      `--setupFilesAfterEnv=${path.join('<rootDir>', 'jest-global.mjs')}`,
+      `--setupFilesAfterEnv=${path.join('<rootDir>', 'polyfills.mjs')}`,
+      `--setupFilesAfterEnv=${path.join('<rootDir>', 'init-test-bed.mjs')}`,
+
+      // Don't run any infrastructure files as tests, they are manually loaded where needed.
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'jest-global\\.mjs')}`,
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'polyfills\\.mjs')}`,
+      `--testPathIgnorePatterns=${path.join('<rootDir>', 'init-test-bed\\.mjs')}`,
```
alan-agius4, line 106: The `path.join`'s here are redundant. Also, there is no need to create a regexp pattern for `testPathIgnorePatterns` as these can simple be paths.

I also suggest that values are wrapped in quotes to avoid possible issues when there are special characters or spaces in the paths. 

```suggestion
      // Third, we initialize `TestBed`. This is dependent on fake async being set up correctly beforehand.
      `--setupFilesAfterEnv='<rootDir>/jest-global.mjs'`,
      `--setupFilesAfterEnv='<rootDir>/polyfills.mjs'`,
      `--setupFilesAfterEnv='<rootDir>/init-test-bed.mjs'`,

      // Don't run any infrastructure files as tests, they are manually loaded where needed.
      `--testPathIgnorePatterns='<rootDir>/jest-global.mjs'`,
      `--testPathIgnorePatterns='<rootDir>/polyfills.mjs'`,
      `--testPathIgnorePatterns='<rootDir>/init-test-bed.mjs'`,
```

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    // Execute Jest on the built output directory.
+    const jestProc = execFile('node', [
+      '--experimental-vm-modules',
+      jest,
+
+      `--rootDir=${testOut}`,
+      '--testEnvironment=jsdom',
+      '--color',
```
alan-agius4, line 88: I think you meant `colors` here. 

I also think that we should set this option based on https://github.com/angular/angular-cli/blob/81768965c45cf2efb9044bf5333fd804f68eef6c/packages/angular_devkit/build_angular/src/utils/color.ts#L47 to avoid printing color on non TTY terminals.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    });
+    if (!buildResult.success) {
+      return buildResult;
+    }
+
+    // Execute Jest on the built output directory.
+    const jestProc = execFile('node', [
+      '--experimental-vm-modules',
+      jest,
+
```
alan-agius4, line 85: Should we disable the cache since we are not providing an option to clear it?

packages/angular_devkit/build_angular/src/builders/jest/schema.json
```diff
@@ -0,0 +1,48 @@
...
+      "items": {
+        "type": "string"
+      },
+      "default": [],
+      "description": "Globs of files to exclude, relative to the project root."
+    },
+    "tsConfig": {
+      "type": "string",
+      "description": "The name of the TypeScript configuration file.",
+      "default": "tsconfig.spec.json"
```
alan-agius4, line 26:
```suggestion
      "description": "The name of the TypeScript configuration file."
```

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      };
+    }
+
+    // Build all the test files.
+    const testFiles = await findTestFiles(options, context.workspaceRoot);
+    const jestGlobal = path.join(__dirname, 'jest-global.mjs');
+    const initTestBed = path.join(__dirname, 'init-test-bed.mjs');
+    const buildResult = await build(context, {
+      // Build all the test files and also the `jest-global` and `init-test-bed` scripts.
+      entryPoints: new Set(Array.from(testFiles).concat([jestGlobal, initTestBed])),
```
alan-agius4, line 60: NIT: personal preference as I find it cleaner and easier to read.

```suggestion
      entryPoints: new Set([...testFiles, jestGlobal, initTestBed]),
```

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    return { success: true };
+  } catch (err) {
+    return {
+      success: false,
+      error: (err as Error).message,
+    };
+  }
+}
```
clydin, line 147: Unused code?

packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -267,7 +274,7 @@ ts_library(
             "src/**/tests/*.ts",
         ],
         exclude = [
-            "src/testing/**/*_spec.ts",
+            "src/**/*_spec.ts",
```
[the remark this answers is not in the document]
dgp1130: It's actually necessary because otherwise the Jest test sources get pulled into multiple `ts_library()` targets and conflict with each other during analysis.

packages/angular_devkit/build_angular/src/builders/jest/schema.json
```diff
@@ -0,0 +1,48 @@
...
+          "default": []
+        },
+        {
+          "type": "string",
+          "description": "The full path for the polyfills file, relative to the current workspace or a module specifier. Example: 'zone.js'."
+        }
+      ]
+    }
+  },
+  "additionalProperties": false
```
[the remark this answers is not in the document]
dgp1130: `tsConfig` defaults to `tsconfig.spec.json` which aligns with the default generated file structure. In the upcoming v16 blog post we just use `"builder": "@angular-devkit/build-angular:jest"` with no options, so I'm inclined to keep that at least for the initial release and we can make some of these options required later if we want to.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+/** Main execution function for the Jest builder. */
+export default createBuilder(
+  async (schema: JestBuilderSchema, context: BuilderContext): Promise<BuilderOutput> => {
+    context.logger.warn(
+      'NOTE: The Jest builder is currently EXPERIMENTAL and not ready for production use.',
+    );
+
+    const options = normalizeOptions(schema);
+    const testOut = 'dist/test-out'; // TODO(dgp1130): Hide in temp directory.
```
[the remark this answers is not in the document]
dgp1130: Agreed, my longer-term thought is to generate these files in a temporary directory outside the project and then delete it afterwards to make this less observable. For now it's useful for debugging purposes and since we don't have a `--debug` flag yet this is easiest way to introspect what's going on at runtime. Once we have better debug support, I agree we should hide the directory altogether.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      index: null,
+      outputHashing: OutputHashing.None,
+      outExtension: 'mjs', // Force native ESM.
+      commonChunk: false,
+      optimization: false,
+      buildOptimizer: false,
+      sourceMap: {
+        scripts: true,
+        styles: false,
+        vendor: true,
```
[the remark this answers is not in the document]
dgp1130: I think I basically just trying to enable sourcemaps, not sure if these specific options are the best approach. In retrospect, I wonder if `scripts` is referring to third-party scripts configured in `angular.json`? That would always be empty right now anyways. @clydin is this the right way to enable sourcemaps for now?

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: Yeah, I didn't like that either, but printing to stderr colors all the inputs red and clobbers Jest own coloring. Not sure if there's a better way to print to stderr while retaining the output style?

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: That is a good point, what do you think would be the right behavior here? If we just print stderr again, then it duplicates the test output which we already streamed out. I don't think there's an easy way to say "if the error is a process creation issue then print stderr"? Is that something `spawn` would be able to give?

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: This gets tricky because the user can specify custom include globs using `*.(mc)?ts` extensions and I wanted to avoid transforming globs to `*.mjs` if it can be helped. I think it's easier to blocklist the infrastructure files we _know_ we don't have to run than to try to transform a user-provided glob.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: Since it's running all the `**/*.mjs` files, we need to explicitly blocklist any files we shouldn't run.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: This is called to trigger the actual build. I just wanted to encapsulate the `for await` and `try {} catch {}` since  those get verbose when inlined.

packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -38,6 +38,11 @@ ts_json_schema(
     src = "src/builders/extract-i18n/schema.json",
```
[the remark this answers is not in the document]
dgp1130: Interesting, I didn't know we had this. Updated.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+      'NOTE: The Jest builder is currently EXPERIMENTAL and not ready for production use.',
+    );
+
+    const options = normalizeOptions(schema);
+    const testOut = 'dist/test-out'; // TODO(dgp1130): Hide in temp directory.
+
+    // Verify Jest installation and get the path to it's binary.
+    // We need to `node_modules/.bin/jest`, but there is no means to resolve that directly. Fortunately Jest's `package.json` exports the
+    // same file at `bin/jest`, so we can just resolve that instead.
+    const jest = require.resolve('jest/bin/jest');
```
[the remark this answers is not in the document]
dgp1130: Oh right, this was leftover from when I used `which` as that returned `undefined`. Updated.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (!jest) {
+      return {
+        success: false,
+        error:
+          'Jest is not installed, most likely you need to run `npm install jest --save-dev` in your project.',
```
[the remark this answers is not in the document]
dgp1130: Chatted briefly about this and we don't see to have out-of-the-box utilities for this kind of thing just yet. I left a `TODO` for now and we can come back to that later.

packages/angular_devkit/build_angular/src/builders/jest/schema.json
```diff
@@ -0,0 +1,48 @@
...
+  "description": "Jest target options",
+  "type": "object",
+  "properties": {
+    "include": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "default": ["**/*.spec.ts"],
+      "description": "Globs of files to include, relative to project root. \nThere are 2 special cases:\n - when a path to directory is provided, all spec files ending \".spec.@(ts|tsx)\" will be included\n - when a path to a file is provided, and a matching spec file exists it will be included instead."
```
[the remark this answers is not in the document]
dgp1130: Good catch. I legitimately don't know where that string came from, I don't think I wrote it...

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: I had a feeling those `path.join()`'s were unnecessary, but wasn't too sure if Windows would throw a wrench into it. `--testPathIgnorePatterns` is treated as a regex/glob, so the `\.` is needed to be specific about the file. https://jestjs.io/docs/cli#--testpathignorepatternsregexarray

The quotes is a good point. I think we _should_ be skipping shell expansion here, but good to be safe.

alan-agius4: Based on the tests https://github.com/jestjs/jest/blob/268afca708199c0e64ef26f35995907faf4454ff/e2e/multi-project-config-root/package.json the values can be string paths.

packages/angular_devkit/build_angular/src/builders/jest/test-files.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Finds all test files in the project.
+ *
+ * @param options The builder options describing where to find tests.
+ * @param workspaceRoot The path to the root directory of the workspace.
+ * @param glob A promisified implementation of the `glob` module. Only intended for
+ *     testing purposes.
+ * @returns A set of all test files in the project.
+ */
+export async function findTestFiles(
```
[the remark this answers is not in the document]
dgp1130: This is a good idea, and I hadn't considered it earlier. Thinking about it now however, I'm inclined _not_ to reuse Karma's test discovery logic for a few reasons:

1. Direct 1:1 compatibility between Jest and Karma is currently not a goal.
2. I personally think some of the behavior in `findTests` is not intuitive or necessarily helpful. Directory inclusion and `.spec` injection both sound like features I would not want to include in the first place unless compatibility was an objective.
3. Karma's implementation hard-codes `.spec` as a convention in both of the above features, even if projects are actually using a different convention like `.test`.
4. The behavior of `include` is dependent on the actual file system layout which I think is a bit of an anti-pattern for a feature like this.
5. The `test-files.ts` file is currently a bit coupled to Webpack. It doesn't look that hard to separate out but we generally can't assume a dependency on Webpack or Karma exists at all and need to make sure the import graph aligns with that expectation.

I'm certainly open to revisiting if users really do like those extra features or get tripped up by the difference with the existing Karma builder, but I don't think it's strictly necessary for this initial experiment.

packages/angular_devkit/build_angular/src/builders/jest/tests/test-files_spec.ts
```diff
@@ -0,0 +1,172 @@
+/**
```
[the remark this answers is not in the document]
dgp1130: I took a stab at this but unfortunately wasn't able to get it fully working. There were a number of weird issues I encountered from a random typing error in unchanged code (I suspect installing Jest bumped the `@babel/core` version, but the lockfile diff doesn't indicate that), to Jest suddenly wanting a config file, and the builder harness silently failing to write any files.

I agree we should probably restructure these tests to fit that model, but I think I'll have to follow up after this PR.

packages/angular/cli/lib/config/workspace-schema.json
```diff
@@ -516,6 +516,28 @@
                 }
               }
             },
+            {
```
alan-agius4, line 519: You missed to add `@angular-devkit/build-angular:jest` in https://github.com/angular/angular-cli/blob/b492080059e792913485161779261fdaf5f3b0ea/packages/angular/cli/lib/config/workspace-schema.json#L363

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,164 @@
...
+  } catch (err) {
+    return {
+      success: false,
+      error: (err as Error).message,
+    };
+  }
+}
+
+/** Safely resolves the given Node module string. */
+function resolve(module: string): string | undefined {
```
alan-agius4, line 158: Nit: I suggest to rename it to something more specific like `resolveModule` to avoid conflict with path `resolve`.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    // Build all the test files.
+    const testFiles = await findTestFiles(options, context.workspaceRoot);
+    const jestGlobal = path.join(__dirname, 'jest-global.mjs');
+    const initTestBed = path.join(__dirname, 'init-test-bed.mjs');
+    const buildResult = await build(context, {
+      // Build all the test files and also the `jest-global` and `init-test-bed` scripts.
+      entryPoints: new Set(Array.from(testFiles).concat([jestGlobal, initTestBed])),
+      tsConfig: options.tsConfig,
+      polyfills: options.polyfills ?? ['zone.js', 'zone.js/testing'],
```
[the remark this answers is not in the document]
alan-agius4: This similar to the other comment about "tsConfig". We should have consistence between builders. I am okay with adding a default as long as we know that other builders will follow suit or we are going to delete the default from this builder. Also, in this case the default is in code, unlike the one for `tsConfig`, this is not ideal in documentation/help pages it will show that by default no polyfills are included which is not the case.

Therefore, if we are adding a default this should be in the schema.json and not in code.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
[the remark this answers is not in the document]
alan-agius4: This is the right approach to enable sourcemaps. But in this case probably `vendor` should be turned off to avoid extra processing.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
alan-agius4: There is no good way right now when using the logger. You'd need to use `process.stderr` directly.

packages/angular_devkit/build_angular/src/builders/jest/test-files.ts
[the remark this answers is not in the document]
alan-agius4: Thinking about it a bit more, the `include` karma functionality was mainly intended to match Jest features, but with Jest this might be a bit redundant, since Jest can already handle most of this features such as running all tests in a specific directory.

packages/angular_devkit/build_angular/BUILD.bazel
```diff
@@ -347,6 +354,12 @@ LARGE_SPECS = {
             "@npm//buffer",
         ],
     },
+    "jest": {
+        "extra_deps": [
+            "@npm//glob",
```
alan-agius4, line 359: I am surprised this worked. But should it `jest` and `jest-environment-jsdom` be added here as well?

alan-agius4 (commented):
@dgp1130, please also squash the commits as in the CLI we do not use fixup.

packages/angular_devkit/build_angular/BUILD.bazel
dgp1130: The unit tests don't actually run Jest because they are testing our code around it (evaluating the `include`  and `exclude` globs for instance). The e2e tests run `npm install jest jest-environment-jsdom` during their setup, so it's outside the Bazel layer.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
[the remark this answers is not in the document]
dgp1130: After some discussion we decided to update this to align with other builders, so `polyfills` defaults to `[]` (as specified by the schema, so docs should be accurate). Users need to add `['zone.js', 'zone.js/testing']` if they are using Zones.

We should probably have a separate conversation about whether this should be the default, but decision should probably be made for _all_ official builders, not just Jest.

packages/angular_devkit/build_angular/src/builders/jest/test-files.ts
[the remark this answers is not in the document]
dgp1130: Interesting, I wasn't aware of that history. For now, let's keep `include` functionality simple, and we can always expand it later if there's enough desire for it.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: Updated to write to stderr with a comment explaining why `context.logger.error` is undesirable.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: For the initial experiment we probably don't need to worry about it too much. I left a `TODO` for now to make sure we don't forget this later.

packages/angular/cli/lib/config/workspace-schema.json
dgp1130: Ah, good catch. Updated.

clydin (approved):

PR #25065, opened 2023-04-25
fix(@angular-devkit/build-angular): avoid double sourcemap comments with esbuild dev-server

clydin:
When using the esbuild-based browser application builder with the dev-server builder, sourcemap URL comments were unintentionally doubled due to vite adding an additional inline sourcemap comment when processing the JavaScript files for the application. To avoid this, JavaScript files now have any sourcemap URL comments removed prior to being processed by vite. The sourcemap content is already passed directly to vite and allows the sourcemaps to be processed and provided to the browser as needed without the previously existing sourcemap URL comment. The removal is done without modifying the esbuild-based builder's options to avoid assuming the builder used with the dev-server has sourcemap options that allow hidden sourcemap generation.


alan-agius4 (approved):
LGTM

PR #25070, opened 2023-04-26
fix(@angular-devkit/build-angular): provide option to run build-optimizer on server bundles

alan-agius4:
This commit adds the option to run build-optimizer on the server bundles.

This is needed as in some cases, there can be restrictions on the size of server bundles that can be executed. One of these cases is CloudFlare workers
 which by default does not work with an `ng-new` application due to the size of the
 bundle mainly due to the retention of the `@angular/compiler`.

//cc @AndrewKushnir

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/server/schema.json
+++ b/packages/angular_devkit/build_angular/src/builders/server/schema.json
@@ -189,6 +189,11 @@
       "description": "Extract all licenses in a separate file, in the case of production builds only.",
       "default": true
     },
+    "buildOptimizer": {
+      "type": "boolean",
+      "description": "Enables advanced build optimizations.",
+      "default": true
+    },
     "namedChunks": {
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/builders/server/schema.json
```diff
@@ -189,6 +189,11 @@
       "description": "Extract all licenses in a separate file, in the case of production builds only.",
       "default": true
     },
+    "buildOptimizer": {
+      "type": "boolean",
+      "description": "Enables advanced build optimizations.",
+      "default": false
```
alan-agius4, line 195: For consistence should we set this to `true`? 

//cc @clydin

clydin: i think so.

clydin (approved):

PR #25083, opened 2023-04-27
fix(@angular-devkit/build-angular): exclude `@angular/platform-server/init` from unsafe optimizations

alan-agius4:
`@angular/platform-server/init` entry-point is side-effectful and thus it's not safe to perform advanced optimizations.

### DIFF
--- a/packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
+++ b/packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
@@ -154,12 +154,15 @@ export default custom<ApplicationPresetOptions>(() => {
       }
 
       if (optimize) {
-        const angularPackage = /[\\/]node_modules[\\/]@angular[\\/]/.test(this.resourcePath);
+        // @angular/platform-server/init entry-point has side-effects.
+        const safeAngularPackage =
+          /[\\/]node_modules[\\/]@angular[\\/]/.test(this.resourcePath) &&
+          !/@angular[\\/]platform-server[\\/]f?esm2022[\\/]init/.test(this.resourcePath);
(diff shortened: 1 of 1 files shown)


clydin (approved):

packages/angular_devkit/build_angular/src/babel/webpack-loader.ts
```diff
@@ -154,12 +154,15 @@ export default custom<ApplicationPresetOptions>(() => {
       }
 
       if (optimize) {
-        const angularPackage = /[\\/]node_modules[\\/]@angular[\\/]/.test(this.resourcePath);
+        // @angular/platform-server/init entry-point has side-effects.
```
devversion, line 157: what about other side-effect entry-points? we might be adding arbitrary ones. Can't we rely on the `sideEffects` field in package.json? We also have `localize/init` IIRC (but yes- that one is usually not in user code I think)

clydin: Once we move this into the actual FW packaging and bundling process, these checks can all be removed.
Also, adding any side effect entry points should be done with care since import statement usage then can become order dependent.  As well as also causing deoptimizations in other aspects of the build pipeline.

alan-agius4: `@angular/localize/init` is not in users code indeed.

devversion: Yes exactly, just need to be careful in adding new entry-points. I don't think the current solution is ideal (and I'm not aware of the concrete plans going forward)

PR #25085, opened 2023-04-27
refactor: replace `node` with `process.execPath`

alan-agius4:
This is ensure that the same node executable is used to spawn Jest


alan-agius4:
Green windows run https://app.circleci.com/jobs/github/angular/angular-cli/415684

PR #25126, opened 2023-05-03
docs: update release doc to be more clear about peer dependency updates and timing

dgp1130:
I mainly rewrote the section about modifying dependencies after a major/minor release, however I also refactored the other sections to more clearly specify the timing of each action and share a single list of all files which need to be updated so duplicates don't fall out of date.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -67,11 +67,9 @@ In general, cherry picks for LTS should only be done if it meets one of the crit
 
 # Release
 
-Releasing is performed using Angular's unified release tooling. Each week, two releases are expected, `latest` and `next` on npm.
-
-**For a minor OR major release:**
-
-After FW releases `-rc.0` for an upcoming minor/major version, update the corresponding version in:
+Releasing is performed using Angular's unified release tooling. Each week, two releases are expected, `latest` and `next` on npm. For major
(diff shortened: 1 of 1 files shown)


docs/process/release.md
```diff
@@ -67,38 +67,36 @@ In general, cherry picks for LTS should only be done if it meets one of the crit
...
 - [`@angular/pwa`](/packages/angular/pwa/package.json)
 - [`@angular-devkit/build-angular`](/packages/angular_devkit/build_angular/package.json)
 - [`@ngtools/webpack`](/packages/ngtools/webpack/package.json)
 
-The same needs to be done for a `-next.0` release, and needs to be done for both minor _and_ major
-releases.
+**BEFORE a minor OR major CLI release:**
+
+After FW releases `-rc.0` for an upcoming minor/major version, update the corresponding version in the above files to match `-rc.0`.
+
```
alan-agius4, line 83: This is actually not correct. We should not need to update to `-RC.0`. If we want to do so, we'd need to update the release tool checks.

dgp1130: Right, I remember this was a challenge. I would like to find a path to do this just so users' `package.json` files reflect the RC state, but we can follow up separately on that. Removed this line for now.

alan-agius4 (approved):

clydin (approved):

PR #25145, opened 2023-05-05
fix(@angular-devkit/build-angular): fix index option const value for browser-esbuild

leosvelperez:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The `index` option of the `browser-esbuild` builder is defined to accept one of a `string`, an `object`, or a const `"false"`. Note the const value it's not `false` but a string value. Because of that, there's no way to disable the index generation:

- If we set the option to `"false"` we get an error because it matches both the `string` and the const:
	```
	Error: Schema validation failed with the following errors:
    Data path "/index" must be object.
    Data path "/index" must match exactly one schema in oneOf.
    ```
- If we set the option to `false` we get an error because it doesn't match any of the possible definitions:
	```
	Error: Schema validation failed with the following errors:
    Data path "/index" must be string.
    Data path "/index" must be object.
    Data path "/index" must be equal to constant.
    Data path "/index" must match exactly one schema in oneOf.
    ```
- If we run the build with `ng build --no-index` (as one of the goals of the change https://github.com/angular/angular-cli/commit/0d4a40fd6eea6e60f5fc17ddb4ff74e58086669d) we get the same as the previous error:
    ```
	Error: Schema validation failed with the following errors:
    Data path "/index" must be string.
    Data path "/index" must be object.
    Data path "/index" must be equal to constant.
    Data path "/index" must match exactly one schema in oneOf.
    ```

Issue Number: N/A

## What is the new behavior?

The `index` option of the `browser-esbuild` builder is defined to accept one of a `string`, and `object` or a const `false` (boolean). This allows users to set the option to `false` in their `angular.json` or to run the build with `--no-index`.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
+++ b/packages/angular_devkit/build_angular/src/builders/browser-esbuild/schema.json
@@ -399,7 +399,7 @@
           "required": ["input"]
         },
         {
-          "const": "false",
+          "const": false,
           "description": "Does not generate an `index.html` file."
         }
       ]


alan-agius4 (approved):
LGTM

PR #25149, opened 2023-05-05
fix(@angular-devkit/build-angular): convert dev-server glob proxy entries for esbuild builder

### LINKED ISSUE #25148: Proxy is not working when using esbuild
### Command

serve

### Is this a regression?

_No response_

### Description

When running an application using the esbuild the proxy config is ignored. After changing the builder back to the default, the proxy gets hit. 

### Minimal Reproduction

1. Create a new app in Angular 16 using `ng new proxy-test --standalone`
2. Change the builder option in `angular.json` to `@angular-devkit/build-angular:browser-esbuild`
3. Add a `proxyConfig` to the serve configuration
4. Run the proxy and the application
5. Endpoints using the defined proxy path are not received in the proxy.

Expected behavior: the proxy configured in the proxy.conf.json should be called when the defined endpoint is called.

Only after changing the builder option back to the default (`@angular-devkit/build-angular:browser`) is the proxy used.

### Exception or Error

```text
Nothing
```

### Your Environment

```text
Angular CLI: 16.0.0
Node: 16.14.0
Package Manager: npm 8.3.1
OS: linux x64

Angular: 16.0.0

@angular-devkit/architect       0.1600.0
@angular-devkit/build-angular   16.0.0
@angular-devkit/core            16.0.0
@angular-devkit/schematics      16.0.0
@schematics/angular             16.0.0
typescript                      5.0.4
```

### Anything else relevant?

_No response_

clydin:
When using the esbuild-based browser application builder with the development server, an underlying Vite server is used. The Vite server currently does not support glob-based entries for the proxy configuration. They must either be prefix strings or regular expressions. The Webpack-based development server, however, does support globs. To remove the need to have different proxy configuration files for the two servers, the entries will now be normalized to regular expressions when using the Vite server. This allows existing proxy configurations to work without modification.

Fixes #25148


packages/angular_devkit/build_angular/src/builders/dev-server/load-proxy-config.ts
```diff
@@ -6,10 +6,12 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import isGlob from 'is-glob';
```
alan-agius4, line 9: Instead of adding an additional dependency, the `glob` library `hasMagic` method can be used to check is the input is glob.

alan-agius4 (approved):

PR #25180, opened 2023-05-09
feat(@angular-devkit/build-angular): allow forcing esbuild builder with dev-server

clydin:
To allow lower overhead trial of the developer preview of the esbuild-based builder system, the development server now has an option to force the usage of the esbuild-based build system while still retaining the default Webpack-based build system for the `build` command. The `forceEsbuild`/`--force-esbuild` option can be added to the `angular.json` options for the `serve` target or used on the command line, respectively. The `browser-esbuild` builder will be used to build the application using the options specified by the server configuration's `browserTarget` option. Unsupported build options will be ignored. If using a third-party builder, a warning will be issued but the build will still be attempted. Third-party builder usage in this context is considered unsupported and may result in unexpected behavior or build failures.

### DIFF
--- a/goldens/public-api/angular_devkit/build_angular/index.md
+++ b/goldens/public-api/angular_devkit/build_angular/index.md
@@ -110,6 +110,7 @@ export interface DevServerBuilderOptions {
     allowedHosts?: string[];
     browserTarget: string;
     disableHostCheck?: boolean;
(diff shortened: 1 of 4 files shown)


clydin:
Yes, the option allows it to be set in the configuration once so all team members for a project can leverage the same development experience without changes to their development workflow.

Issue #25198, opened 2023-05-12
esbuild: Error when declaring relative path global scripts without starting them by "./" on Windows.


wartab:
### Command

build, serve

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

This issue only happens on Windows, I couldn't reproduce this on Linux.

When importing global scripts with relative paths that don't start with "./" in angular.json , the commands "ng build" and "ng serve" fail with an error.

### Minimal Reproduction

- ng new test-project
- npm i marked
- Update angular.json to include:
"node_modules/marked/marked.min.js" making sure it's not "./node_modules/marked/marked.min.js"
- npm i

Changing it to 
"./node_modules/marked/marked.min.js" gets rid of the error.

### Exception or Error

```text
X [ERROR] Could not resolve "node_modules/marked/marked.min.js" [plugin angular-global-scripts]
```

### Your Environment

```text
_                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/


Angular CLI: 16.0.1
Node: 16.18.0
Package Manager: npm 9.6.4
OS: win32 x64

Angular: 16.0.1
... animations, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1600.1
@angular-devkit/build-angular   16.0.1
@angular-devkit/core            16.0.1
@angular-devkit/schematics      16.0.1
@schematics/angular             16.0.1
rxjs                            7.8.1
typescript                      5.0.4
```

### Anything else relevant?

Windows only issue.

PR #25204, opened 2023-05-12
fix(@angular-devkit/build-angular): normalize Vite dev-server Windows asset paths

clydin:
When using the esbuild-based browser application builder with the Vite-based development server on Windows, source asset paths were previously not normalized prior to being included in request URLs. This could result in invalid asset request URLs due to invalid path segment separators.


alan-agius4 (approved):

alan-agius4:
Does this solve https://github.com/angular/angular-cli/issues/25198?

PR #25247, opened 2023-05-19
ci: add initial Github Action lint workflow

clydin:
The eslint CI action has now been moved to a Github Action.
The check also now adds annotations to the PR code when eslint generates an error
or warning. Additionally, the eslint cache is saved and reused which provides improved
runtimes for the CI action.


alan-agius4 (approved):
Nice

PR #25253, opened 2023-05-20
fix(@schematics/angular): allow dash in selector before a number

### LINKED ISSUE #25164: generators: allow dash `-` in selector before a number
### Command

generate

### Description

We use jira. Tickets always follow the schema `<PROJECT>-<TICKETNUMBER>` so e.g. `Task-123`. If we do an a/b test, we like to postfix the files that mirror the control group with the ticket, so e.g `feature-task-123.component.ts`. But the cli right now doesn't allow this.

See `ng g c TodoItemTask123` neither the files nor the selector have the `-`. So `task-123` is actually `task123`:
CleanShot 2023-05-08 at 13 48 53@2x

So what I'd like to do would be one of these:
```
ng g c TodoItemTask-123
ng g c TodoItemTask123 --selector app-todo-item-task-123
```

but both commands get rejected with `Selector "app-todo-item-task-123" is invalid`:
CleanShot 2023-05-08 at 13 46 29@2x

I'd like if `ng g c TodoItemTask-123` was allowed and generates files like:
```
todo-item-task-123.component.ts
              ^ having the dash
```

Renaming all the files and updating the NgModule by hand afterwards.

sumitparakh:
Closes #25164

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Does not allow dash in selector before a number. Which means, following command would throw error: -
`ng g c TodoItemTask-123`

Issue Number: https://github.com/angular/angular-cli/issues/25164

## What is the new behavior?

Command like this will work: -
`ng g c TodoItemTask-123`
image

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/schematics/angular/utility/validation.ts
+++ b/packages/schematics/angular/utility/validation.ts
@@ -10,7 +10,8 @@ import { SchematicsException } from '@angular-devkit/schematics';
 
 // Must start with a letter, and must contain only alphanumeric characters or dashes.
 // When adding a dash the segment after the dash must also start with a letter.
-export const htmlSelectorRe = /^[a-zA-Z][.0-9a-zA-Z]*(:?-[a-zA-Z][.0-9a-zA-Z]*)*$/;
+export const htmlSelectorRe =
+  /^[a-zA-Z][.0-9a-zA-Z]*((:?-[0-9]+)*|(:?-[a-zA-Z][.0-9a-zA-Z]*(:?-[0-9]+)*)*)$/;
 
 // See: https://github.com/tc39/proposal-regexp-unicode-property-escapes/blob/fe6d07fad74cd0192d154966baa1e95e7cda78a1/README.md#other-examples
 const ecmaIdentifierNameRegExp = /^(?:[$_\p{ID_Start}])(?:[$_\u200C\u200D\p{ID_Continue}])*$/u;
--- a/packages/schematics/angular/component/index_spec.ts
+++ b/packages/schematics/angular/component/index_spec.ts
@@ -208,6 +208,30 @@ describe('Component Schematic', () => {
     ).toBeRejectedWithError('Selector "app-1-one" is invalid.');
   });
 
+  it('should allow dash in selector before a number', async () => {
+    const options = { ...defaultOptions, name: 'one-1' };
+
+    const tree = await schematicRunner.runSchematic('component', options, appTree);
+    const content = tree.readContent('/projects/bar/src/app/one-1/one-1.component.ts');
+    expect(content).toMatch(/selector: 'app-one-1'/);
+  });
+
+  it('should allow dash in selector before a number and with a custom prefix', async () => {
+    const options = { ...defaultOptions, name: 'one-1', prefix: 'pre' };
+
+    const tree = await schematicRunner.runSchematic('component', options, appTree);
+    const content = tree.readContent('/projects/bar/src/app/one-1/one-1.component.ts');
+    expect(content).toMatch(/selector: 'pre-one-1'/);
+  });
+
+  it('should allow dash in selector before a number and without a prefix', async () => {
+    const options = { ...defaultOptions, name: 'one-2', selector: 'one-2' };
+
+    const tree = await schematicRunner.runSchematic('component', options, appTree);
+    const content = tree.readContent('/projects/bar/src/app/one-2/one-2.component.ts');
+    expect(content).toMatch(/selector: 'one-2'/);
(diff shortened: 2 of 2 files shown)


alan-agius4 (changes requested):
Thanks for your contribution. 

Can you please replace the E2E with a Unit test, similar to: https://github.com/angular/angular-cli/blob/e037dc9675e7ef999ec5655fc3ca88e2444e74e6/packages/schematics/angular/application/index_spec.ts#L512-L517

Also, kindly update the commit message to 
```
fix(@schematics/angular): allow dash in selector before a number

Closes #25164
```

sumitparakh:
Hi @alan-agius4 , I have made the suggested changes but the setup step failed with `500 internal server error`. Does not seem to be related to my code.

sumitparakh:
> Can you please clean you the commit message description and remove
> 
> ```
> fix(@angular/cli): fix regex
> 
> fix(@angular/cli): updated test case
> 
> fix(@angular/cli): fix "too big tests" error in ci
> 
> added a new file to test dash with number case
> ```
> 
> if you do want to add a description it should be before the `closes` and it should describe the change. Ex:
> 
> ```
> 
> fix(@schematics/angular): allow dash in selector before a number
> 
> This commit updates the validator regexp to allow a dash before a number.
> 
> Closes #25164
> ```

Done @alan-agius4. :)

packages/schematics/angular/utility/validation.ts
```diff
@@ -10,7 +10,7 @@ import { SchematicsException } from '@angular-devkit/schematics';
 
 // Must start with a letter, and must contain only alphanumeric characters or dashes.
 // When adding a dash the segment after the dash must also start with a letter.
-export const htmlSelectorRe = /^[a-zA-Z][.0-9a-zA-Z]*(:?-[a-zA-Z][.0-9a-zA-Z]*)*$/;
+export const htmlSelectorRe = /^[a-zA-Z][.0-9a-zA-Z]*(:?-[a-zA-Z][.0-9a-zA-Z]*(:?-[0-9]+)*)*$/;
```
alan-agius4, line 13: In some cases this regexp will fail to match

Example:
```
one-1
```

Maybe we can use the same logic as https://github.com/angular/angular-cli/blob/b66d94e5949102b2235b89cb75866973986ea30a/packages/angular_devkit/schematics/src/formats/html-selector.ts#L17C1-L44

//cc @clydin do you recall why we support upper characters as based on the spec it seems that these are not valid. https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name

sumitparakh: @alan-agius4 Found another problem. If we allow `one-1` like selectors then it will also verify `app-1-one` as valid and in that case, the component class name becomes like this: -

```
export class 1OneComponent {
}
```

This is invalid class name. Isn't it?

So, I might have to add some conditions that regexp doesn't allow `1-one` if the component is being created with a prefix which will ultimately become `app-1-one`.

packages/schematics/angular/utility/validation.ts
```diff
@@ -10,7 +10,8 @@ import { SchematicsException } from '@angular-devkit/schematics';
 
 // Must start with a letter, and must contain only alphanumeric characters or dashes.
 // When adding a dash the segment after the dash must also start with a letter.
-export const htmlSelectorRe = /^[a-zA-Z][.0-9a-zA-Z]*(:?-[a-zA-Z][.0-9a-zA-Z]*)*$/;
+export const htmlSelectorRe =
+  /^[a-zA-Z][.0-9a-zA-Z]*((:?-[0-9]+)*|(:?-[a-zA-Z][.0-9a-zA-Z]*(:?-[0-9]+)*)*)$/;
```
sumitparakh, line 14: Regexp used [here](https://github.com/angular/angular-cli/blob/b66d94e5949102b2235b89cb75866973986ea30a/packages/angular_devkit/schematics/src/formats/html-selector.ts#L17C1-L44) doesn't validate html selector correctly. For eg., it considers `foo-` as a valid selector which is wrong, right? And it also allows underscore. 

I have modified regexp in validation.ts file itself to cover selectors like `one-1` but if we still want to use the logic at `html-selector.ts` then we might have to make some changes in that logic.

sumitparakh:
Hi @alan-agius4, I have made some changes. Could you please review this?

alan-agius4 (approved):
LGTM

PR #25255, opened 2023-05-22
build: update to TypeScript 5.1

crisbeto:
Bumps up the dev dependencies to TypeScript 5.1-beta and expands the allowed version ranges to prepare for the final release.

### DIFF
--- a/packages/angular_devkit/build_angular/test/hello-world-app/tsconfig.json
+++ b/packages/angular_devkit/build_angular/test/hello-world-app/tsconfig.json
@@ -8,12 +8,10 @@
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "skipLibCheck": true,
     "target": "es2022",
     "module": "es2022",
     "useDefineForClassFields": false,
-    "typeRoots": [
-      "node_modules/@types"
-    ],
     "lib": [
       "es2022",
       "dom"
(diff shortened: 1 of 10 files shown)


alan-agius4 (approved):

packages/angular_devkit/build_angular/test/hello-world-app/tsconfig.json
```diff
@@ -8,12 +8,10 @@
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "skipLibCheck": true,
```
crisbeto, line 11: @alan-agius4 heads-up for this one: I had to disable it, because there seems to be some sort of mismatch between the types in the Bazel bin and what is used in the rest of the project.

packages/angular_devkit/build_angular/test/hello-world-app/tsconfig.json
```diff
@@ -8,12 +8,10 @@
...
     "target": "es2022",
     "module": "es2022",
     "useDefineForClassFields": false,
-    "typeRoots": [
```
crisbeto, line 14: In TS 5.1 the presence of `typeRoots` means that only those specific directories will be checked. Previously TS was also going up the file tree. I had to remove this, because the type root doesn't contain types for `@anguar/localize`.

PR #25258, opened 2023-05-22
refactor(@angular-devkit/core): use package exports field for secondary entries

clydin:
The secondary entry point (`node`) within the `@angular-devkit/core` package is now defined by an exports entry in the main package.json file for the package. This removes the need for the additional nested package.json file in a subdirectory. The exports field also contains backward compatible entries to allow for continued deep imports into the package. These compatibility entries will, however, be removed in a future major version. Deep imported files are not considered part of the public API of the package.


packages/angular_devkit/core/package.json
```diff
@@ -7,6 +7,23 @@
...
+      "default": "./node/index.js",
+      "types": "./node/index.d.ts"
+    },
+    "./node/testing": {
+      "default": "./node/testing/index.js",
+      "types": "./node/testing/index.d.ts"
+    },
+    "./package.json": "./package.json",
+    "./*": "./*.js",
+    "./*.js": "./*.js"
```
JoostK, line 25: Are these evaluated in order, or by longest prefix matching? In case of the former these need to be swapped.

packages/angular_devkit/core/package.json
```diff
@@ -7,6 +7,23 @@
   "keywords": [
     "core"
   ],
+  "exports": {
+    ".": {
+      "default": "./src/index.js",
+      "types": "./src/index.d.ts"
```
JoostK, line 13: The [docs for conditional exports](https://nodejs.org/api/packages.html#conditional-exports) specifies that `default` should always come last.

PR #25259, opened 2023-05-22
build: remove unused internal benchmarking tool

clydin:
This benchmarking tool is unpublished and not used within the repository. It is also largely unmaintained outside of periodic dependency updates.


alan-agius4 (commented):
We can now removal several unused deps. 

```
pidtree
pidusage
@types/pidusage
```

PR #25279, opened 2023-05-26
feat(@angular-devkit/build-angular): support autoprefixer/tailwind CSS with Less/Sass in esbuild builder

clydin:
When using the esbuild-based browser application builder, Sass and Less stylesheets will now be post-processed with autoprefixer and/or Tailwind CSS when applicable. CSS stylesheets were already processed by these tools. Autoprefixer is queried based on the configured browserslist to determine if any processing is required and is not added to the build pipeline if no transformations are required. Likewise for Tailwind, if no Tailwind configuration file is present, Tailwind CSS will also not be added to the build pipeline. If both autoprefixer and Tailwind are not required, `postcss` (the tool used to post-process the stylesheets) itself is not added to the build pipeline. This removes the potential for unneeded build time overhead for projects that do not require these post-processing steps.
The default browserslist currently does require the use of autoprefixer based on autoprefixer's prefix analysis.


packages/angular_devkit/build_angular/src/builders/browser-esbuild/index.ts
```diff
@@ -32,7 +32,7 @@ import { extractLicenses } from './license-extractor';
 import { BrowserEsbuildOptions, NormalizedBrowserOptions, normalizeOptions } from './options';
 import { Schema as BrowserBuilderOptions } from './schema';
 import { createSourcemapIngorelistPlugin } from './sourcemap-ignorelist-plugin';
-import { shutdownSassWorkerPool } from './stylesheets/sass-plugin';
+import { shutdownSassWorkerPool } from './stylesheets/sass-language';
```
alan-agius4, line 35: +1 for the rename

alan-agius4 (approved):

PR #25297, opened 2023-05-31
refactor(@angular-devkit/build-angular): use fast-glob for file searching support

clydin:
The file searching within the build system (both Webpack and esbuild) now use the `fast-glob` package for globbing which provides a small performance improvement. Since the assets option in particular is within the critical path of the buil pipeline, the performance benefit from the switch will be most prevalent in asset heavy projects. As an example, the Angular Material documentation site saw the asset discovery time reduced by over half with the switch. `fast-glob` is also the package used by Vite which provides additional benefit by ensuring that the Angular CLI behavior matches that of the newly integrated Vite development server.


alan-agius4 (approved):
Nice 🚀

PR #25309, opened 2023-06-02
APIs for authoring bootstrap-agnostic ng add schematics

crisbeto:
These changes add the necessary APIs that library authors can use to create `ng add` schematics that don't need to account for whether a project is NgModule-based or standalone. They're split up across a few commits to make reviewing easier. High-level description of the commits:

1. `refactor(@schematics/angular): allow for imports to be inserted under an alias` - Expands the `insertImport` utility to allow for aliases to be specified. This is necessary to avoid naming conflicts when automatically adding imports. I've also added some tests for `insertImport` since it didn't have any test coverage.
2. `refactor(@schematics/angular): add utility to find top-level identifiers` - Adds a utility that can be used to find whether a file has a top-level identifier with a specific name. Useful when trying to avoid import conflicts.
3. `feat(@schematics/angular): add bootstrap-agnostic utilities for writing ng-add schematics` - This is the commit that adds the two new `addRootImport` and `addRootProvider` utilities that library authors can use.
4. `refactor(@schematics/angular): deprecate private standalone utilities` - Marks the existing private standalone utilities as deprecated in favor of the new ones.

### DIFF
--- a/packages/schematics/angular/utility/ast-utils_spec.ts
+++ b/packages/schematics/angular/utility/ast-utils_spec.ts
@@ -19,6 +19,7 @@ import {
   addSymbolToNgModuleMetadata,
   findNodes,
   insertAfterLastOccurrence,
+  insertImport,
 } from './ast-utils';
 
 function getTsSource(path: string, content: string): ts.SourceFile {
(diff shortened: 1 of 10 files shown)


packages/schematics/angular/utility/ast-utils_spec.ts
```diff
@@ -685,4 +687,164 @@ describe('ast utils', () => {
       });
     });
   });
+
+  describe('insertImport', () => {
```
crisbeto, line 691: These are tests mostly for existing functionality. I added them since we had no coverage.

packages/schematics/angular/utility/standalone/app_config.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/** App config that was resolved to its source node. */
+export interface ResolvedAppConfig {
+  /** Tree-relative path of the file containing the app config. */
+  filePath: string;
+
+  /** Node defining the app config. */
+  node: ts.ObjectLiteralExpression;
+}
+
```
crisbeto, line 22: There's a bit of duplication here with the utilities we exposed for Components. The private utilities will be deleted once Components has been switched over.

alan-agius4 (approved):

PR #25390, opened 2023-06-16
build: update build-tooling package and enable Node.js 18 tests

alan-agius4:
This commit updates the build-tooling package which contains fixes for RBE on Node.js 18. It also enables the Node.js 18 test suit.


packages/angular_devkit/build_webpack/src/webpack-dev-server/index_spec.ts
```diff
@@ -54,7 +54,7 @@ describe('Dev Server Builder', () => {
     const output = (await run.result) as DevServerBuildOutput;
     expect(output.success).toBe(true);
 
-    const response = await fetch(`http://${output.address}:${output.port}/bundle.js`);
+    const response = await fetch(`http://localhost:${output.port}/bundle.js`);
```
alan-agius4, line 57: This is needed as on Node 18 this can be an IP v6 example `::1`.

devversion (approved):

PR #25435, opened 2023-06-23
fix(@angular-devkit/build-angular): use all style language watch files in esbuild builder

clydin:
A recent change to better support Tailwind CSS in watch mode unintentionally caused part of the watch files list for stylesheets to be ignored when Tailwind and/or autoprefixer were required to be executed. This resulted in rebuilds occurring but all stylesheet changes were not fully propagated to the development server. This has now been corrected.


alan-agius4 (approved):

princemaple:
Tested 16.1.2, problem still there. Neither updating css files directly (irrelevant to tailwind) nor adding tailwind css classes that weren't previously used during first compilation to an html element made any difference to the appearance of the page.

PR #25454, opened 2023-06-27
refactor(@angular-devkit/build-angular): use built-in async generator downleveling in esbuild

clydin:
As of esbuild v0.18.8, async generators can be downleveled when not supported. This improvement allows for the removal of the babel plugins to perform the async generator downleveling that were previously used. This can provide a performance benefit for cases of async generator usage in application and third-party library code.


alan-agius4 (approved):
Nice!

PR #25456, opened 2023-06-27
feat(@angular-devkit/build-angular): add initial support for server bundle generation using esbuild

alan-agius4:
This commit adds initial support to generate the server bundle using esbuild as the underlying bundler.


packages/angular_devkit/build_angular/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -29,11 +29,12 @@ import {
   resetCumulativeDurations,
 } from '../profiling';
 import { BundleStylesheetOptions, bundleComponentStylesheet } from '../stylesheets/bundle-options';
-import { AngularCompilation } from './angular-compilation';
 import { AngularHostOptions } from './angular-host';
-import { AotCompilation } from './aot-compilation';
+import { AngularCompilation } from './compilation/angular-compilation';
+import { AotCompilation } from './compilation/aot-compilation';
+import { JitCompilation } from './compilation/jit-compilation';
+import { NoopCompilation } from './compilation/noop-compilation';
```
clydin, line 36: nit: If we are going to move them all, i think a barrel import would be useful to reduce the amount of import statements here

clydin (approved):

PR #25481, opened 2023-06-30
fix(@angular-devkit/build-angular): normalize paths in loader cache with esbuild

clydin:
When using the esbuild-based application builder, some plugins may return watch file lists that contain POSIX paths on Windows systems. This can cause the file watcher to not correctly invalidate files that need to be processed during a rebuild. All watch files are now normalized prior to being added to the in-memory cache to avoid this problem.


marcellkiss:
When is it supposed to get released?

dgp1130:
Normally we release on Wednesdays, but I forgot to do it due to 4th of July in the U.S. Just released today so it should be included in 16.1.4.

marcellkiss:
Great, thanks!

PR #25489, opened 2023-07-05
feat(@angular-devkit/build-angular): add pre-rendering (SSG) and App-shell support generation to application builder     

alan-agius4:
This commit introduces experimental support to pre-render (SSG) and app-shell generation to the new application builder.
    
- `appShell`: option which can have a value of `true` or `false` has been added to support generating an app-shell.
- `prerender`: option which can have a value of `true`, `false` or an object with the below listed properties can be used to static render pages;
  - `routes`: Array of routes to render.
  - `discoverRoutes`: Whether the builder should statically discover routes.
  - `routesFile`: The path to a file containing routes separated by newlines.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/schema.json
+++ b/packages/angular_devkit/build_angular/src/builders/application/schema.json
@@ -353,6 +353,7 @@
         },
         {
           "const": false,
+          "type": "boolean",
           "description": "Does not generate a service worker configuration."
         }
       ]
@@ -384,6 +385,7 @@
         },
         {
           "const": false,
+          "type": "boolean",
           "description": "Does not generate an `index.html` file."
         }
       ]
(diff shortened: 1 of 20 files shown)


packages/angular_devkit/build_angular/src/utils/ssg/render.ts
```diff
@@ -0,0 +1,143 @@
...
+      case '.mjs': // Contains the server runnable application code.
+      case '.css': // Global styles for critical CSS inlining.
+        outputFilesForWorker[path] = text;
+        break;
+    }
+  }
+
+  const renderWorker = new Piscina({
+    filename: require.resolve('./render-worker'),
+    maxThreads: allRoutes.size > maxThreads ? maxThreads : 1,
```
clydin, line 53:
```suggestion
    maxThreads: Math.min(allRoutes.size, maxThreads),
```
Or is it intended to only use one thread if less than maxThreads?

alan-agius4: The intention was indeed to use a single worker if it’s less than maxThreads.

packages/angular_devkit/build_angular/src/builders/application/schema.json
```diff
@@ -418,6 +420,46 @@
...
       },
       "default": []
+    },
+    "ssg": {
+      "description": "Prerender (SSG) pages of your application during build time.",
+      "default": false,
+      "oneOf": [
+        {
+          "type": "boolean",
+          "description": "Disable SSG."
```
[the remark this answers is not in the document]
alan-agius4: Yes, in which case the default options will be applied.

clydin (approved):

PR #25499, opened 2023-07-06
 feat(@angular-devkit/build-angular): add `ssr` option in application builder

alan-agius4:
This commit adds an `ssr` option to the application builder, this can be either a `boolean` or an `object` with an `entryPoint` property.

In the future, server bundles will only be emitted when the ssr option is truthy, as unlike SSR, SSG and AppShell do not require the server bundles to be written to disk.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/schema.json
+++ b/packages/angular_devkit/build_angular/src/builders/application/schema.json
@@ -452,6 +452,26 @@
         }
       ]
     },
+    "ssr": {
+      "description": "Server side render (SSR) pages of your application during runtime.",
+      "default": false,
+      "oneOf": [
+        {
+          "type": "boolean",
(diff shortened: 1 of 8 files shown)


packages/angular_devkit/build_angular/src/builders/application/schema.json
```diff
@@ -452,6 +452,26 @@
...
+        {
+          "type": "boolean",
+          "description": "Enable the server bundles to be written to disk."
+        },
+        {
+          "type": "object",
+          "properties": {
+            "entry": {
+              "type": "string",
+              "description": "The server entry-point that when executed will spawn the server."
```
clydin, line 468:
```suggestion
              "description": "The server entry-point that when executed will spawn the web server."
```
Just to make it a little more explicit as to the options purpose

clydin (approved):

PR #25503, opened 2023-07-06
build: update prettier to v3.0.0

clydin:
Tests that used prettier for assertions were updated to reflect the new prettier promise-based API.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,6 +1,5 @@
 {
   "printWidth": 100,
   "quoteProps": "preserve",
-  "singleQuote": true,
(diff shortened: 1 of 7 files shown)


.prettierrc
```diff
@@ -1,6 +1,5 @@
 {
   "printWidth": 100,
   "quoteProps": "preserve",
-  "singleQuote": true,
-  "trailingComma": "all"
```
clydin, line 5: This value is now the default

alan-agius4 (approved):

Issue #25189, opened 2023-05-11
@angular-devkit/build-angular:jest creates invalid test suite


rafagsiqueira:
### Command

test

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

Running tests using ng test <project name> with the new @angular-devkit/build-angular:jest builder results always in at least one failed test suite:
 ```
FAIL  dist/test-out/StandardController-QMDYSHA5.mjs
  ● Test suite failed to run

    Your test suite must contain at least one test.

      at onResult (../../node_modules/@jest/core/build/TestScheduler.js:133:18)
      at ../../node_modules/@jest/core/build/TestScheduler.js:254:19
      at ../../node_modules/emittery/index.js:363:13
          at Array.map (<anonymous>)
      at Emittery.emit (../../node_modules/emittery/index.js:361:23)
```
Perhaps is just a matter of adding an extra testPathIgnorePattern similar to this one?
https://github.com/angular/angular-cli/blob/35d239e9376277e50cee8c83b2e7a90626ec941d/packages/angular_devkit/build_angular/src/builders/jest/index.ts#L114

### Minimal Reproduction

ng test <project name>

### Your Environment

```text
Angular CLI: 16.0.0
Node: 18.16.0
Package Manager: yarn 3.5.0
OS: darwin x64

Angular: 16.0.1
... animations, common, compiler, compiler-cli, core, forms
... localize, platform-browser, platform-browser-dynamic, router
... service-worker

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1600.0
@angular-devkit/build-angular   16.0.0
@angular-devkit/core            16.0.0
@angular-devkit/schematics      16.0.0-rc.1
@angular/cdk                    16.0.0
@angular/cli                    16.0.0
@angular/flex-layout            15.0.0-beta.42
@angular/material               16.0.0
@angular/pwa                    16.0.0
@schematics/angular             16.0.0
ng-packagr                      16.0.1
rxjs                            7.8.1
typescript                      5.0.4
```

dgp1130:
This is interesting, where does `StandardController` come from? Is that a file name in your application?

I was under the impression ESBuild always generated shared chunks in `chunk-*.mjs` files, but maybe that's not strictly accurate.

rafagsiqueira:
It's not. It was generated by the ESBuild.

dplarina:
[...]

PR #25511, opened 2023-07-07
fix(@angular-devkit/build-angular): Set chunk names explicitly

gultyayev:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, chunk names are determined by ESBuild. Some libraries may resolve to a name that is not in a format of `chunk.hash.mjs`. This causes Jest to run over the lib files and fail as it contains no test suite.

Fixes: #25189

## What is the new behavior?

The change sets chunk name pattern explicitly so that all `chunk`s are exluded by pattern from running tests over.

- [ ] Yes
- [x] No

## Other information

I didn't figure out how this could be covered by tests. I did run a local build & installation and verified on a sample project that `@azure/msal-angular` no longer creates a "StandardController" file. Instead, it's named as `chunk...` and hence is excluded from tests.

### DIFF
--- a/tests/legacy-cli/e2e/tests/basic/rebuild.ts
+++ b/tests/legacy-cli/e2e/tests/basic/rebuild.ts
@@ -1,13 +1,13 @@
-import { waitForAnyProcessOutputToMatch, silentNg } from '../../utils/process';
-import { writeFile, writeMultipleFiles } from '../../utils/fs';
 import fetch from 'node-fetch';
-import { ngServe } from '../../utils/project';
 import { getGlobalVariable } from '../../utils/env';
+import { writeFile, writeMultipleFiles } from '../../utils/fs';
+import { silentNg, waitForAnyProcessOutputToMatch } from '../../utils/process';
+import { ngServe } from '../../utils/project';
 
 export default async function () {
   const esbuild = getGlobalVariable('argv')['esbuild'];
   const validBundleRegEx = esbuild ? /complete\./ : /Compiled successfully\./;
-  const lazyBundleRegEx = esbuild ? /lazy\.module/ : /lazy_module_ts\.js/;
+  const lazyBundleRegEx = esbuild ? /chunk-/ : /lazy_module_ts\.js/;
 
   const port = await ngServe();
   // Add a lazy module.
@@ -17,6 +17,7 @@ export default async function () {
   // We need to use Promise.all to ensure we are waiting for the rebuild just before we write
   // the file, otherwise rebuilds can be too fast and fail CI.
   // Count the bundles.
+  // Verify that a new chunk was created.
   await Promise.all([
     waitForAnyProcessOutputToMatch(lazyBundleRegEx),
     writeFile(
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/application-code-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/application-code-bundle.ts
@@ -251,6 +251,7 @@ function getEsBuildCommonOptions(options: NormalizedApplicationBuildOptions): Bu
     outExtension: outExtension ? { '.js': `.${outExtension}` } : undefined,
     sourcemap: sourcemapOptions.scripts && (sourcemapOptions.hidden ? 'external' : true),
     splitting: true,
+    chunkNames: 'chunk-[hash]',
     tsconfig,
     external: externalDependencies,
     write: false,


mullinsmikey:
This seems like a good idea, but can you please provide some context?

gultyayev:
> This seems like a good idea, but can you please provide some context?

Mostly, the issue comes down to ESBuild generating chunk names as `chunk.hash.mjs` in 99% of the cases. Sometimes, however it generates a chunk with a different name. For me it was `StandardController.mjs` for imported `@azure/msal-angular` library (v3.0.0-alpha.0).

When running tests with Jest, the builder is using pattern `**/*.mjs` to literally pick up all the mjs files. It does have exclusion for `chunk` files. Because of that, when running tests with Jest it fails, because `StandardController.mjs` is included to run tests over and it doesn't have any.

First idea was to simply change the pattern to `.spec.mjs`, but as it was highlighted by @dgp1130 users may change the tests name pattern. Therefore, it's easier to make all chunks follow the name pattern.

For build code it doesn't make any difference. For tests it makes sure that all chunks have the same name pattern and are excluded from the tests.

dgp1130 (changes requested):
Thanks @gultyayev for looking into this, I haven't had a chance to dig in yet. I think this is probably a reasonable solution, though it looks like one of the tests is failing, as it is expecting the app to [print the lazy loaded file's name](https://github.com/angular/angular-cli/blob/a96b1b1f78f87baf9d839c7cbafe6bd0301e7717/tests/legacy-cli/e2e/tests/basic/rebuild.ts#L10) which is no longer true. Maybe we could search for a more generic "rebuild done" string like "Application bundle generation complete"?

Also making this change in `builder-context.ts` means that it will apply to all ESBuild executions, not just Jest. I think that's _probably_ ok, but I wonder if there might be some unintended consequences from it. @clydin, do you have any concerns with that?

gultyayev:
> Also making this change in builder-context.ts means that it will apply to all ESBuild executions, not just Jest.

TBH I can't think of any possible shortcomings from this. The names are always generated and hence cannot be relied on. Furthermore, current behavior is to create chunks with `chunk-hsh.mjs` name. So, technically we are fixing those limited cases when the name of the chunk is different.

It's also possible to pass this change for Jest builder only. I didn't locate the place though. Not sure if it would make much sense to do.

alan-agius4 (commented):
Instead of setting the option in the builder context which only contains infrastructure related options. It would be better to follow the existing pattern for similar options and set it in https://github.com/angular/angular-cli/blob/4a8edff07fd616ef4abc5290c7f0ec9c1e8569d2/packages/angular_devkit/build_angular/src/tools/esbuild/application-code-bundle.ts#L183 or even better https://github.com/angular/angular-cli/blob/4a8edff07fd616ef4abc5290c7f0ec9c1e8569d2/packages/angular_devkit/build_angular/src/builders/application/options.ts#L92

gultyayev:
> Instead of setting the option in the builder context which only contains infrastructure related options. It would be better to follow the existing pattern for similar options and set it in
> 
> https://github.com/angular/angular-cli/blob/4a8edff07fd616ef4abc5290c7f0ec9c1e8569d2/packages/angular_devkit/build_angular/src/tools/esbuild/application-code-bundle.ts#L183
> 
> or even better
> https://github.com/angular/angular-cli/blob/4a8edff07fd616ef4abc5290c7f0ec9c1e8569d2/packages/angular_devkit/build_angular/src/builders/application/options.ts#L92

Did the first suggested change. The 2nd one doesn't work.

gultyayev:
One concern that may be is that chunk names become obscure. Current `ng build` outputs lazy loaded chunks as `component-name.component-[hash].js`. With the change it becomes just `chunk-[hash]` which may be undesirable.

tests/legacy-cli/e2e/tests/basic/rebuild.ts
```diff
@@ -1,13 +1,13 @@
...
 import { getGlobalVariable } from '../../utils/env';
+import { writeFile, writeMultipleFiles } from '../../utils/fs';
+import { silentNg, waitForAnyProcessOutputToMatch } from '../../utils/process';
+import { ngServe } from '../../utils/project';
 
 export default async function () {
   const esbuild = getGlobalVariable('argv')['esbuild'];
   const validBundleRegEx = esbuild ? /complete\./ : /Compiled successfully\./;
-  const lazyBundleRegEx = esbuild ? /lazy\.module/ : /lazy_module_ts\.js/;
+  const bundleGeneratedRegEx = esbuild ? /bundle generation complete/ : /Compiled successfully/;
```
clydin, line 10: The original intent for this was to ensure a new chunk was being created for the added lazy module.
Can you bring back the name-based check but with `chunk` instead for the esbuild case?

dgp1130:
@gultyayev, thanks for your patience. Using `chunk-[hash]` everywhere should be fine. It does make chunks a little harder to debug, however there is no general 1:1 correlation between chunks and source files, so preserving the name is maybe setting a bad expectation in the first place.

Just the one comment from @clydin about making sure a chunk is created on the rebuild and this should be good to go.

clydin (approved):

dgp1130 (approved):

dgp1130:
Thanks! This looks good to merge.

gultyayev:
Thank you for guidance.

PR #25534, opened 2023-07-12
perf(@angular-devkit/build-angular): inject Sass import/use directive importer information when resolving

clydin:
To correctly resolve a package based import reference in a Sass file with pnpm or Yarn PnP, the importer file path must be known. Unfortunately, the Sass compiler does not provided the importer file to import plugins. Previously to workaround this issue, all previously resolved stylesheets were tried as the importer path. This allowed the stylesheets to be resolved but it also could cause a potentially large increase in build time due to the amount of previous stylesheets that would need to be tried. To avoid the performance impact and to also provide more accurate information regarding the importer file, a lexer is now used to extract import information for a stylesheet and inject the importer file path into the specifier. This information is then extracted from the import specifier during the Sass resolution process and allows the underlying package resolution access to a viable location to resolve the package for all package managers. This information is currently limited to specifiers referencing the `@angular` and `@material` package scopes but a comprehensive pre-resolution process may be added in the future.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/sass/lexer.ts
+++ b/packages/angular_devkit/build_angular/src/tools/sass/lexer.ts
@@ -1,0 +1,267 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// TODO: Combine everything into a single pass lexer
+
+/**
+ * Determines if a unicode code point is a CSS whitespace character.
+ * @param code The unicode code point to test.
+ * @returns true, if the code point is CSS whitespace; false, otherwise.
+ */
+function isWhitespace(code: number): boolean {
+  // Based on https://www.w3.org/TR/css-syntax-3/#whitespace
+  switch (code) {
+    case 0x0009: // tab
+    case 0x0020: // space
+    case 0x000a: // line feed
+    case 0x000c: // form feed
+    case 0x000d: // carriage return
+      return true;
+    default:
+      return false;
+  }
+}
+
+/**
+ * Scans a CSS or Sass file and locates all valid url function values as defined by the
+ * syntax specification.
+ * @param contents A string containing a CSS or Sass file to scan.
+ * @returns An iterable that yields each CSS url function value found.
+ */
+export function* findUrls(
+  contents: string,
+): Iterable<{ start: number; end: number; value: string }> {
+  let pos = 0;
+  let width = 1;
+  let current = -1;
+  const next = () => {
+    pos += width;
+    current = contents.codePointAt(pos) ?? -1;
+    width = current > 0xffff ? 2 : 1;
+
+    return current;
+  };
+
+  // Based on https://www.w3.org/TR/css-syntax-3/#consume-ident-like-token
+  while ((pos = contents.indexOf('url(', pos)) !== -1) {
+    // Set to position of the (
+    pos += 3;
+    width = 1;
+
+    // Consume all leading whitespace
+    while (isWhitespace(next())) {
+      /* empty */
+    }
+
+    // Initialize URL state
+    const url = { start: pos, end: -1, value: '' };
+    let complete = false;
+
+    // If " or ', then consume the value as a string
+    if (current === 0x0022 || current === 0x0027) {
+      const ending = current;
+      // Based on https://www.w3.org/TR/css-syntax-3/#consume-string-token
+      while (!complete) {
+        switch (next()) {
+          case -1: // EOF
+            return;
+          case 0x000a: // line feed
+          case 0x000c: // form feed
+          case 0x000d: // carriage return
+            // Invalid
+            complete = true;
+            break;
+          case 0x005c: // \ -- character escape
+            // If not EOF or newline, add the character after the escape
+            switch (next()) {
+              case -1:
+                return;
+              case 0x000a: // line feed
+              case 0x000c: // form feed
+              case 0x000d: // carriage return
+                // Skip when inside a string
+                break;
+              default:
+                // TODO: Handle hex escape codes
+                url.value += String.fromCodePoint(current);
+                break;
+            }
+            break;
+          case ending:
+            // Full string position should include the quotes for replacement
+            url.end = pos + 1;
+            complete = true;
+            yield url;
+            break;
+          default:
+            url.value += String.fromCodePoint(current);
+            break;
+        }
+      }
+
+      next();
+      continue;
+    }
+
+    // Based on https://www.w3.org/TR/css-syntax-3/#consume-url-token
+    while (!complete) {
+      switch (current) {
+        case -1: // EOF
+          return;
+        case 0x0022: // "
+        case 0x0027: // '
+        case 0x0028: // (
+          // Invalid
+          complete = true;
+          break;
+        case 0x0029: // )
+          // URL is valid and complete
+          url.end = pos;
+          complete = true;
+          break;
+        case 0x005c: // \ -- character escape
+          // If not EOF or newline, add the character after the escape
+          switch (next()) {
+            case -1: // EOF
+              return;
+            case 0x000a: // line feed
+            case 0x000c: // form feed
+            case 0x000d: // carriage return
+              // Invalid
+              complete = true;
+              break;
+            default:
+              // TODO: Handle hex escape codes
+              url.value += String.fromCodePoint(current);
+              break;
+          }
+          break;
+        default:
+          if (isWhitespace(current)) {
+            while (isWhitespace(next())) {
+              /* empty */
+            }
+            // Unescaped whitespace is only valid before the closing )
+            if (current === 0x0029) {
+              // URL is valid
+              url.end = pos;
+            }
+            complete = true;
+          } else {
+            // Add the character to the url value
+            url.value += String.fromCodePoint(current);
+          }
+          break;
+      }
+      next();
+    }
+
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
```diff
@@ -74,6 +83,26 @@ abstract class UrlRebasingImporter implements Importer<'sync'> {
       updatedContents.update(start, end, rebasedUrl);
     }
 
+    for (const { start, end, value } of findImports(contents)) {
+      // Currently only provide directory information for known/common packages:
+      // * `@material/`
+      // * `@angular/`
+      // Comprehensive pre-resolution support may be added in the future.
+      if (!value.startsWith('@angular/') && !value.startsWith('@material/')) {
```
dgp1130, line 91: Question: Any particular reason for limiting to `@angular/` and `@material/`? Just staying safe in case it breaks something?

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
```diff
@@ -74,6 +83,26 @@ abstract class UrlRebasingImporter implements Importer<'sync'> {
       updatedContents.update(start, end, rebasedUrl);
     }
 
+    for (const { start, end, value } of findImports(contents)) {
```
dgp1130, line 86: Suggestion: Rename `value` to something more meaningful. Maybe `importSpecifier`?

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
```diff
@@ -444,20 +315,41 @@ export class ModuleUrlRebasingImporter extends RelativeUrlRebasingImporter {
...
+      specifier: string,
+      options: { fromImport: boolean; resolveDir?: string },
+    ) => URL | null,
   ) {
     super(entryDirectory, directoryCache, rebaseSourceMaps);
   }
 
   override canonicalize(url: string, options: { fromImport: boolean }): URL | null {
-    if (url.startsWith('file://')) {
+    if (url[0] === 'f' && url.startsWith('file://')) {
```
dgp1130, line 329: Question: What's the need for `url[0] === 'f'`? Is that a performance optimization? Is `startsWith` that slow that it makes a difference?

I also see this pattern used a couple times. Should we make a separate `fastStartsWith` function which reads `substring[0]` so we don't have to duplicate the first character and can comment explaining why it's necessary to do this?

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
```diff
@@ -444,20 +315,41 @@ export class ModuleUrlRebasingImporter extends RelativeUrlRebasingImporter {
...
 
-    const result = this.finder(url, options);
+    let resolveDir;
+    if (url[0] === 'n' && url.startsWith('ng-package;')) {
+      const values = url.split(';', 3);
+      resolveDir = values[1];
+      url = values[2];
+    }
+
+    let result: URL | undefined | null = this.moduleCache.get(url)?.resolved;
```
dgp1130, line 340: The module cache seems kind of unrelated to this rebaser/canonicalizer. I wonder if we could pull that out to a slightly higher level of abstraction and maybe cache before/after calling `canonicalize()` to separate concerns a little cleaner?

That would probably apply the cache to more than just `ng-package;` specifiers, but maybe that's a good thing? Not sure how viable this is, just a thought which stood out to me.

packages/angular_devkit/build_angular/src/tools/sass/lexer.ts
```diff
@@ -0,0 +1,267 @@
+/**
```
dgp1130, line 1: I'm mostly ignoring the lexer since it doesn't look like it changed much. If there's anything in particular I should be looking at, let me know. I recommend moving this refactor into a different commit so the changes to the lexer which support this Sass change show up in the diff more clearly.

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
```diff
@@ -74,6 +83,26 @@ abstract class UrlRebasingImporter implements Importer<'sync'> {
...
+      if (!value.startsWith('@angular/') && !value.startsWith('@material/')) {
+        continue;
+      }
+
+      const resolvePath = stylesheetDirectory + ';' + value;
+
+      // Normalize path separators and escape characters
+      // https://developer.mozilla.org/en-US/docs/Web/CSS/url#syntax
+      const resolvedUrl =
+        'ng-package;' + resolvePath.replace(/\\/g, '/').replace(/[()\s'"]/g, '\\$&');
```
dgp1130, line 100: Question: Would this break if the user named a file/project `ng-package`? I know it's unlikely, but not exactly implausible. Would something like `__ng-package__` be a little safer?

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
```diff
@@ -444,20 +315,41 @@ export class ModuleUrlRebasingImporter extends RelativeUrlRebasingImporter {
...
 
   override canonicalize(url: string, options: { fromImport: boolean }): URL | null {
-    if (url.startsWith('file://')) {
+    if (url[0] === 'f' && url.startsWith('file://')) {
       return super.canonicalize(url, options);
     }
 
-    const result = this.finder(url, options);
+    let resolveDir;
+    if (url[0] === 'n' && url.startsWith('ng-package;')) {
```
dgp1130, line 334: The two `ng-package` usages, where we create the fake module specifier and then unpack it later, are kind of far apart and not clearly connected.

Suggestion: Make a pair of functions to pack/unpack the data in this string format, possibly with a constant for `ng-package` to be clear that we're defining a special string format and using it in two places which are tied together. This will make it more obvious that updating one also requires updating the other.

dgp1130 (commented):
Do we have any existing unit tests for this functionality we should be updating as well? Or is the existing integration test suite sufficient for this change, given that it's a performance optimization which shouldn't observably change anything?

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
clydin: it is a micro-optimization. dropped it for now and can re-evaluate in the future with more profiling.

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
clydin: Currently it would apply to all non-relative resolution attempts.  Since we don't control the execution of `canonicalize` it needs to be done inside. `canonicalize` is called by the Sass compiler itself and is akin to a onResolve hook.

I dropped the cache from this PR.  It does result in less of a performance improvement overall here but it can be introduced in a separate change. I wanted to eventually make the cache work across all active Sass workers anyway.

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
```diff
@@ -22,6 +23,38 @@ export interface DirectoryEntry {
...
+/**
+ * A prefix that is added to import and use directive specifiers that should be resolved
+ * as modules and that will contain added resolve directory information.
+ *
+ * This functionality is used to workaround the Sass limitation that it does not provide the
+ * importer file to custom resolution plugins.
+ */
+const MODULE_RESOLUTION_PREFIX = '__NG_PACKAGE__';
+
+function packModuleSpecifier(specifier: string, resolveDir: string): string {
```
dgp1130, line 35: Suggestion: An interface for the unpacked format would allow you to accept/return the same type rather than disparite values.

```typescript
interface ModuleSpecifier {
  specifier: string;
  resolveDir?: string;
}
```

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
```diff
@@ -22,6 +23,38 @@ export interface DirectoryEntry {
...
+
+  // Normalize path separators and escape characters
+  // https://developer.mozilla.org/en-US/docs/Web/CSS/url#syntax
+  const normalizedPacked = packed.replace(/\\/g, '/').replace(/[()\s'"]/g, '\\$&');
+
+  return normalizedPacked;
+}
+
+function unpackModuleSpecifier(specifier: string): { specifier: string; resolveDir?: string } {
+  if (specifier.startsWith(`${MODULE_RESOLUTION_PREFIX};`)) {
```
dgp1130, line 46: Nit: The "interesting" part of the function is the `specifier.split(';', 3)`. I suggest inverting the `if` statement to early return `{ specifier }` in the uninteresting case so the rest of the function can focus on the interesting one.

dgp1130 (approved):

packages/angular_devkit/build_angular/src/tools/sass/rebasing-importer.ts
clydin: Safer for now and avoids some more complex logic for determining if it is relative or an actual module reference.  CSS/Sass don't require `./` or `../` for relative paths. So for general support, a bare specifier would need to be checked for relative path viability first. Technically a project could have a relative `@angular` directory but that's unlikely and easy to check/remedy (add `./` to the specifier).

dgp1130: Makes sense. Might be useful to include some of that in the comment so we know the constraints if/when we change or open up the package restriction.

PR #25537, opened 2023-07-13
refactor(@angular-devkit/build-angular): configure AOT compiler to skip NgModule scope metadata emit in esbuild

clydin:
The esbuild plugin used within the esbuild-based browser application builder will now use the newly introduced internal `supportJitMode` AOT compiler option to disable the emit of NgModule scope metadata functions within the output code. This removes the need to perform an additional transformation of the AOT compiler generated code to immediately remove the metadata. The Angular CLI neither previously nor currently supports hybrid AOT/JIT mode. In the future if this support situation changes, additional behavior may be added control the internal `supportJitMode` option.


dgp1130 (approved):
> The Angular CLI previously nor currently

Should this be "The Angular CLI **neither** previously nor currently"?

PR #25538, opened 2023-07-13
fix(@angular-devkit/build-angular): correctly wrap CommonJS exported enums when optimizing

### LINKED ISSUE #25524: Enum transpiled from TypeScript 5.1 are imported as `undefined`
### Command

build, serve

### Is this a regression?

_No response_

#### Context

TypeScript 5.1 change the way they transpile `enum` to JavaScript

[TS 5.0](https://www.typescriptlang.org/play?module=1&ts=5.0.4#code/KYDwDg9gTgLgBMAdgVwLZwNIEtEBM4DeAUHKXAGoCGANssHALxwBEAbjXc0QL5EDGERAGcI1YADpqEAOYAKbHgCUQA)

`./Enum_TS_5_0.js`

```js
"use strict";
exports.Enum_TS_5_0 = void 0;
var Enum_TS_5_0;
(function (Enum_TS_5_0) {
  Enum_TS_5_0["Value"] = "Value";
})((Enum_TS_5_0 = exports.Enum_TS_5_0 || (exports.Enum_TS_5_0 = {})));
```

[TS 5.1](https://www.typescriptlang.org/play?module=1&ts=5.1.6#code/KYDwDg9gTgLgBMAdgVwLZwNIEtEBM4DeAUHKXAGoCGANssHALxwBEAbjXc0QL5FA)

`./Enum_TS_5_1.js`

```js
"use strict";
exports.Enum_TS_5_1 = void 0;
var Enum_TS_5_1;
(function (Enum_TS_5_1) {
  Enum_TS_5_1["Value"] = "Value";
})(Enum_TS_5_1 || (exports.Enum_TS_5_1 = Enum_TS_5_1 = {}));
```

#### Problem

Enable `"buildOptimizer": true` in you `angular.json` file will break the way enum is loaded in your application.

```ts
import { Enum_TS_5_0 } from "./Enum_TS_5_0";
import { Enum_TS_5_1 } from "./Enum_TS_5_1";

console.log("Enum_TS_5_0", Enum_TS_5_0);
console.log("Enum_TS_5_1", Enum_TS_5_1);
```

```console
Enum_TS_5_0 { "Value": "Value" }
Enum_TS_5_1 undefined
```

If the flag is set to `false` the problem does not occurs.

### Minimal Reproduction

Here is a minimal repository to reproduce the issue
https://github.com/ghoullier/angular-typescript-5-1-enum-issue

### Exception or Error

_No response_

### Your Environment

```text
                |___/
    

Angular CLI: 16.1.4
Node: 18.16.1
Package Manager: yarn 3.6.1
OS: darwin arm64

Angular: 16.1.4

@angular-devkit/architect       0.1601.4
@angular-devkit/build-angular   16.1.4
@angular-devkit/core            16.1.4
@angular-devkit/schematics      16.1.4
@schematics/angular             16.1.4
typescript                      5.1.6
```

### Anything else relevant?

Problem only occurs when the module is `CommonJS`

clydin:
When optimizing a CommonJS exported enum, the build optimizer enum wrapping pass was previously dropping the `exports` object assignment from the enum wrapper function call expression. This would not occur with application code but is possible with library code that was built with TypeScript and shipped as CommonJS.

Assuming the following TypeScript enum:
```
 export enum ChangeDetectionStrategy {
   OnPush = 0,
   Default = 1,
 }
```

TypeScript 5.1 will emit an exported enum for CommonJS as follows:
```
exports.ChangeDetectionStrategy = void 0;
var ChangeDetectionStrategy;
(function (ChangeDetectionStrategy) {
    ChangeDetectionStrategy[ChangeDetectionStrategy["OnPush"] = 0] = "OnPush";
    ChangeDetectionStrategy[ChangeDetectionStrategy["Default"] = 1] = "Default";
})(ChangeDetectionStrategy || (exports.ChangeDetectionStrategy = ChangeDetectionStrategy = {}));
```

The build optimizer would previously transform this into:
```
exports.ChangeDetectionStrategy = void 0;
var ChangeDetectionStrategy = /*#__PURE__*/ (() => {
  ChangeDetectionStrategy = ChangeDetectionStrategy || {};
  ChangeDetectionStrategy[(ChangeDetectionStrategy["OnPush"] = 5)] = "OnPush";
  ChangeDetectionStrategy[(ChangeDetectionStrategy["Default"] = 8)] = "Default";
  return ChangeDetectionStrategy;
})();
```
But this has a defect wherein the `exports` assignment is dropped. To rectify this situation, the build optimizer will now transform the code into:
```
exports.ChangeDetectionStrategy = void 0;
var ChangeDetectionStrategy = /*#__PURE__*/ (function (ChangeDetectionStrategy) {
  ChangeDetectionStrategy[(ChangeDetectionStrategy["OnPush"] = 0)] = "OnPush";
  ChangeDetectionStrategy[(ChangeDetectionStrategy["Default"] = 1)] = "Default";
  return ChangeDetectionStrategy;
})(ChangeDetectionStrategy || (exports.ChangeDetectionStrategy = ChangeDetectionStrategy = {}))
```
This retains the `exports` assignment.

Closes #25524

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/babel/plugins/adjust-typescript-enums_spec.ts
+++ b/packages/angular_devkit/build_angular/src/tools/babel/plugins/adjust-typescript-enums_spec.ts
@@ -36,6 +36,15 @@ function testCase({
   };
 }
 
+// The majority of these test cases are based off TypeScript emitted enum code for the FW's
+// `ChangedDetectionStrategy` enum.
+// https://github.com/angular/angular/blob/55d412c5b1b0ba9b03174f7ad9907961fcafa970/packages/core/src/change_detection/constants.ts#L18
+// ```
+//  export enum ChangeDetectionStrategy {
+//    OnPush = 0,
+//    Default = 1,
+//  }
+// ```
 describe('adjust-typescript-enums Babel plugin', () => {
   it(
     'wraps unexported TypeScript enums',
@@ -48,12 +57,11 @@ describe('adjust-typescript-enums Babel plugin', () => {
         })(ChangeDetectionStrategy || (ChangeDetectionStrategy = {}));
       `,
       expected: `
-        var ChangeDetectionStrategy = /*#__PURE__*/ (() => {
-          ChangeDetectionStrategy = ChangeDetectionStrategy || {};
+        var ChangeDetectionStrategy = /*#__PURE__*/ (function (ChangeDetectionStrategy) {
           ChangeDetectionStrategy[(ChangeDetectionStrategy["OnPush"] = 0)] = "OnPush";
           ChangeDetectionStrategy[(ChangeDetectionStrategy["Default"] = 1)] = "Default";
           return ChangeDetectionStrategy;
-        })();
+        })(ChangeDetectionStrategy || {});
       `,
     }),
   );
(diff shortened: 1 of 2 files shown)


dgp1130 (approved):
A before/after of the actual TS code in the commit message might be helpful to communicate what exactly this is trying to trying to change. The issue has some of that context, but it would be good to replicate here as well.

Separately from this PR, it might be helpful to document each of these Babel transforms with examples of what they are actually transforming. The test cases kind of have this information, but it would be good to get this written down in a more human readable fashion.

packages/angular_devkit/build_angular/src/tools/babel/plugins/adjust-typescript-enums_spec.ts
```diff
@@ -69,12 +68,45 @@ describe('adjust-typescript-enums Babel plugin', () => {
...
           ChangeDetectionStrategy[(ChangeDetectionStrategy["Default"] = 1)] = "Default";
           return ChangeDetectionStrategy;
-        })();
+        })(ChangeDetectionStrategy || {});
+      `,
+    }),
+  );
+
+  it(
+    'does not wrap exported TypeScript enums from CommonJS (<5.1)',
```
dgp1130, line 81: Question: Isn't this a regression? Doesn't it result in enums being less tree shakable than today?

clydin: This case was never actually wrapped.  But it was also not broken by the transformation unlike the TS 5.1+ case.
So no regression but still suboptimal behavior for this one particular case.  However, this older TS form will become increasingly less common over time.

dgp1130: Ok, that sounds reasonable. Agree we don't need to overoptimize for CommonJS.

clydin:
Added more background to the commit message.

More general documentation for these would definitely be good to have.
Although, I am hoping that TypeScript itself could emit enums in this form in the future and remove the need for this transform.

PR #25548, opened 2023-07-17
perf(@angular-devkit/build-angular): use in-memory Sass module resolution cache

clydin:
When using Sass files with module import references in the esbuild-based browser application builder, the module resolution attempts will now be cached in memory. This caching is only local to each entry Sass file currently. However, this may be expanded to encompass all Sass entries within a build in the future. In addition to caching the entire resolution attempt, individual package root resolution is also cached for deep import attempts. This is useful for packages (such as `@material/*`) which deep import to multiple different files in the same package.
With this change combined with the previous lexer change (#25534), package managers (pnpm & Yarn PnP) that require workarounds for functioning resolution will now perform at an equivalent level to other package managers.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
@@ -65,6 +65,33 @@ export const SassStylesheetLanguage = Object.freeze<StylesheetLanguage>({
   },
 });
 
+function parsePackageName(url: string): { packageName: string; readonly pathSegments: string[] } {
+  const parts = url.split('/');
+  const hasScope = parts.length >= 2 && parts[0].startsWith('@');
+  const [nameOrScope, nameOrFirstPath, ...pathPart] = parts;
+  const packageName = hasScope ? `${nameOrScope}/${nameOrFirstPath}` : nameOrScope;
+
+  return {
+    packageName,
+    get pathSegments() {
+      return !hasScope && nameOrFirstPath ? [nameOrFirstPath, ...pathPart] : pathPart;
+    },
+  };
+}
+
+class Cache<K, V> extends Map<K, V> {
+  async getOrCreate(key: K, creator: () => V | Promise<V>): Promise<V> {
+    let value = this.get(key);
+
+    if (value === undefined) {
+      value = await creator();
+      this.set(key, value);
+    }
+
+    return value;
+  }
+}
+
 async function compileString(
   data: string,
   filePath: string,
@@ -81,6 +108,15 @@ async function compileString(
     sassWorkerPool = new sassService.SassWorkerImplementation(true);
   }
 
+  // Cache is currently local to individual compile requests.
+  // Caching follows Sass behavior where a given url will always resolve to the same value
+  // regardless of its importer's path.
+  // A null value indicates that the cached resolution attempt failed to find a location and
+  // later stage resolution should be attempted. This avoids potentially expensive repeat
+  // failing resolution attempts.
+  const resolutionCache = new Cache<string, URL | null>();
+  const packageRootCache = new Cache<string, string | null>();
+
   const warnings: PartialMessage[] = [];
   try {
     const { css, sourceMap, loadedUrls } = await sassWorkerPool.compileStringAsync(data, {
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
```diff
@@ -97,31 +103,49 @@ async function compileString(
...
+              // Caching package root locations is particularly beneficial for `@material/*` packages
+              // which extensively use deep imports.
+              let packageRoot = packageRootCache.get(packageName);
+              if (packageRoot === undefined) {
+                const packageResult = await resolveUrl(packageName + '/package.json', options);
+                if (packageResult.path) {
+                  packageRoot = dirname(packageResult.path);
+                  packageRootCache.set(packageName, packageRoot);
+                }
+              }
```
dgp1130, line 133: Suggestion: `const packageRoot = resolvePackageRoot(packageName);` would help separate concerns and reduce the mutability which readers need to follow.

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
```diff
@@ -97,31 +103,49 @@ async function compileString(
...
             const result = await resolveUrl(url, options);
             if (result.path) {
-              return pathToFileURL(result.path);
+              resolved = pathToFileURL(result.path);
+            } else {
+              // Check for package deep imports
+              const parts = url.split('/');
+              const hasScope = parts.length >= 2 && parts[0].startsWith('@');
+              const [nameOrScope, nameOrFirstPath, ...pathPart] = parts;
+              const packageName = hasScope ? `${nameOrScope}/${nameOrFirstPath}` : nameOrScope;
```
dgp1130, line 122: Suggestion: Refactoring to a `parsePackageName` function would be useful to isolate the specifics of package parsing. Not sure if that's feasible with the following `join`, but it would help a lot with readability if we can reduce the responsibility here.

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
```diff
@@ -97,31 +103,49 @@ async function compileString(
...
+                  packageRoot = dirname(packageResult.path);
+                  packageRootCache.set(packageName, packageRoot);
+                }
+              }
+
+              if (packageRoot) {
+                resolved = pathToFileURL(
+                  join(
+                    packageRoot,
+                    !hasScope && nameOrFirstPath ? nameOrFirstPath : '',
```
dgp1130, line 139: Question: Doesn't this drop the `@angular` part of the package name? Maybe I'm missing something, but I feel like it would generate `path/to/node_modules/material/button/button.scss`?

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
```diff
@@ -81,6 +81,12 @@ async function compileString(
     sassWorkerPool = new sassService.SassWorkerImplementation(true);
   }
 
+  // Cache is currently local to individual compile requests.
+  // Caching follows Sass behavior where a given url will always resolve to the same value
+  // regardless of its importer's path.
+  const resolutionCache = new Map<string, URL | null>();
```
dgp1130, line 87: Question: Is there a need for a distinct `null` in this `Map` as distinct from `undefined` (not in the `Map`)?

What does it _mean_ to have a URL resolve to `null` as opposed to `undefined`? Is it for the error case where the URL can't be resolved?

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
```diff
@@ -97,31 +103,49 @@ async function compileString(
             url,
             options: FileImporterWithRequestContextOptions,
           ): Promise<URL | null> => {
+            let resolved = resolutionCache.get(url);
```
dgp1130, line 106: Suggestion: Similar to the comment thread in the previous PR, I wonder if we could separate out the caching logic to a different layer. Something like decorating or memoizing the `findFileUrl` function with the resolution cache? That would allow readers to think of this implementation as two functions, one which resolves URLs and another which caches them. `if (result.path)` can turn from a mutation to an early return which would make it much easier to follow IMHO.

Could arguably apply the same thing to `packageRootCache`, though that usage is much more localized so I'm less concerned about that one.

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
```diff
@@ -97,31 +103,49 @@ async function compileString(
...
+              let packageRoot = packageRootCache.get(packageName);
+              if (packageRoot === undefined) {
+                const packageResult = await resolveUrl(packageName + '/package.json', options);
+                if (packageResult.path) {
+                  packageRoot = dirname(packageResult.path);
+                  packageRootCache.set(packageName, packageRoot);
+                }
+              }
+
+              if (packageRoot) {
```
dgp1130, line 135: Question: Under what circumstance would we not have a `packageRoot` for an associated `packageName`? Would that be if the user typos a Sass import and specifies a non-existent package?

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
clydin: Returning `null` is not actually an error at this stage.  It just means this importer plugin could not resolve the given input. A later plugin could then successfully resolve it. Configured Sass load paths, for instance, are always tried last. Within that context, if the same input is provided again within a compilation then there's no need to attempt the full resolution when it can instead skip it and let the later stages handle it right away.

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
clydin: Could be a typo in which case there will be an error generated at the end of the full resolution process.  It could also not be a package and is handled by a later resolution stage.

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
clydin: In the scoped case of `@angular/material/button` (assuming an actual deep import and not an exports entry):
`packageRoot`: `path/to/node_modules/@angular/material`
`hasScope`: `true`
`nameOrFirstPath`: unused but contains `material`
`pathPart`: `['button']`
which becomes `join('path/to/node_modules/@angular/material', '', 'button')`

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
dgp1130: Ok, so we're caching the not found case. Can you include a comment on the `resolutionCache` declaration to write this down in the code for future readers?

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts
dgp1130: Does Sass require `./` for relative imports or not (I'm guessing not)? If it does require `./`, we should probably check for that explicitly to limit this logic to real packages. If it doesn't then I get that we need this check and should probably include a comment mentioning that we're covering the non-package case.

clydin: It does not unfortunately. A bare specifier could potentially be relative.

PR #25577, opened 2023-07-24
fix(@angular-devkit/build-angular): correctly load dev server assets with vite 4.4.0+

clydin:
The underlying development server (Vite) for the application build system updated
its static file serving dependencies which resulted in the `@fs` special file URLs
supported by Vite to have a different format.  Previously, the paths were encoded
using `encodeURIComponent` but are now use `encodeURI`. The development server
integration with Vite will now use the matching encoding to allow build defined
assets to be correctly served.

### DIFF
--- a/tests/legacy-cli/e2e/tests/basic/serve.ts
+++ b/tests/legacy-cli/e2e/tests/basic/serve.ts
@@ -18,9 +18,15 @@ export default async function () {
 }
 
 async function verifyResponse(port: number): Promise<void> {
-  const response = await fetch(`http://localhost:${port}/`);
+  const indexResponse = await fetch(`http://localhost:${port}/`);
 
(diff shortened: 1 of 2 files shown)


tests/legacy-cli/e2e/tests/basic/serve.ts
```diff
@@ -23,4 +23,10 @@ async function verifyResponse(port: number): Promise<void> {
   if (!/<app-root><\/app-root>/.test(await response.text())) {
     throw new Error('Response does not match expected value.');
   }
+
+  const response2 = await fetch(`http://localhost:${port}/favicon.ico`);
```
dgp1130, line 27: Nit: `assetResponse` would be a slightly more descriptive name.

dgp1130 (approved):

PR #25584, opened 2023-07-26
fix(@angular-devkit/build-angular): error during critical CSS inlining for external stylesheets

### LINKED ISSUE #25419: Error encountered when updating @angular-devkit/build-angular to Version 16.1.0
### Command

build

### Is this a regression?

16.0.0

### Description

Upon updating @angular-devkit/build-angular to Version 16.1.0, an error occurs during index HTML generation. The specific error message is "document.documentElement.setAttribute is not a function."

Upon investigation, I have identified the root cause of the issue. It appears that enabling the optimization option, specifically setting inlineCritical to True under styles, triggers the error. However, if this option is disabled, the build process completes successfully.

### Minimal Reproduction

Command used:
npx nx run *project*:build:production

Package use:
@angular-devkit/build-angular 16.1.0

### Exception or Error

```text
√ Browser application bundle generation complete.
√ Copying assets complete.
× Index html generation failed.
document.documentElement.setAttribute is not a function
 ——————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————
 >  NX   Ran target build for project *projectname* (58s)
 
    ×    1/1 failed
    √    0/1 succeeded [0 read from cache]
```

### Your Environment

```text
Angular CLI: 16.1.0
Node: 18.10.0
Package Manager: npm 8.19.2
OS: win32 x64

Angular: 16.1.1
... router

@angular-devkit/architect       0.1601.0
@angular-devkit/build-angular   16.1.0
@angular-devkit/core            16.1.0
@angular-devkit/schematics      16.1.0
@angular/cli                    16.1.0
@schematics/angular             16.1.0
typescript                      5.1.3
```

### Anything else relevant?

_No response_

crisbeto:
These changes revert the performance improvement from https://github.com/angular/angular-cli/pull/24925/commits/e88aea65bdb66fa4e9c6ad580ff733642e1fb24c and add a test. The problem with the previous approach is that it assumes that a `link` tag processed by Critters will be preceded by a `style` tag that was inserted by Critters. This assumption might not necessarily be true if Critters is unable to resolve the content of the linked styles. Furthermore, I'm not sure if we need to optimize this code path to begin with, because it only does a shallow traversal of the `head` tag which is generally fast and is unlikely to have many direct descendants.

Fixes #25419.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/index-file/inline-critical-css.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/inline-critical-css.ts
@@ -146,7 +146,15 @@ class CrittersExtended extends Critters {
         this.conditionallyInsertCspLoadingScript(document, cspNonce);
       }
 
-      link.prev?.setAttribute('nonce', cspNonce);
+      // Ideally we would hook in at the time Critters inserts the `style` tags, but there isn't
+      // a way of doing that at the moment so we fall back to doing it any time a `link` tag is
+      // inserted. We mitigate it by only iterating the direct children of the `<head>` which
+      // should be pretty shallow.
+      document.head.children.forEach((child) => {
+        if (child.tagName === 'style' && !child.hasAttribute('nonce')) {
+          child.setAttribute('nonce', cspNonce);
+        }
+      });
     }
 
     return returnValue;
--- a/packages/angular_devkit/build_angular/src/utils/index-file/inline-critical-css_spec.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/inline-critical-css_spec.ts
@@ -133,4 +133,34 @@ describe('InlineCriticalCssProcessor', () => {
     html { color: white; }
     </style>`);
   });
+
+  it('should not modify the document for external stylesheets', async () => {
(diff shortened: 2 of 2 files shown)


packages/angular_devkit/build_angular/src/utils/index-file/inline-critical-css.ts
```diff
@@ -146,7 +146,15 @@ class CrittersExtended extends Critters {
         this.conditionallyInsertCspLoadingScript(document, cspNonce);
       }
 
-      link.prev?.setAttribute('nonce', cspNonce);
```
crisbeto, line 149: We could've kept this and checked that `link.prev.tagName === 'style'` but that'll be incorrect for a markup like this:

```
<head>
  <style></style>
  <link rel="stylesheet" href="https://some/external/style.css">
</head>
```

vlio20:
Hi @crisbeto , in what version this fix will be available?

crisbeto:
It should've been out in 16.1.6.

PR #25622, opened 2023-08-02
ci: reduce bazel test shards for browser builder test

clydin:
Reduce the number of test shards for the browser builder from 50 to 10. 50 is more than needed and can cause excessive resource usage.
This fixes the OOM errors on CI for the test job.


josephperrott (approved):
LGTM

PR #25666, opened 2023-08-10
fix(@angular-devkit/build-angular): encode Sass package resolve directories in importer URLs

clydin:
When using the new developer preview application build system, Sass import/use usage that specifies a package is adjusted to contain the resolve directory to workaround Sass import plugin limitations. This resolve directory is now encoded to prevent the new specifier from looking like a URL with a scheme to the Sass compiler. This can occur on Windows when a drive letter is present (`C:\`).


alan-agius4 (approved):

dgp1130:
I'm still following up with fixing the `ng-dev` tool for merging PRs to patch. In the mean time, this one only merged to `main`. I created #25674 to apply the same change to `16.2.x`.

Issue #22871, opened 2022-03-22
extract-i18n does not extract message from files loaded only server side


manzonif:
# 🐞 Bug report

### Command (mark with an `x`)

- [ ] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [x ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....

### Description
The extraction tool does not process files that do not have a direct dependency on the app.module, or, generally, on the browser side.
In my case an abstract service is replaced by two versions of the service, one browser side, the other server side. The server-side one is ignored during the extraction process.

## 🔬 Minimal Reproduction

I made a simple [repro app](https://github.com/manzonif/ng-extract-i18n-server) that reproduce the problem.
Just run extract-i18n to see that message in [server-service.service.ts](https://github.com/manzonif/ng-extract-i18n-server/blob/master/src/app/server-service.service.ts) is ignored.
## 🔥 Exception or Error

<pre><code>

</code></pre>

## 🌍 Your Environment

<pre><code>

Angular: 12.0.1
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, language-service, material, platform-browser
... platform-browser-dynamic, platform-server, router
... service-worker

Package                           Version
-----------------------------------------------------------
@angular-devkit/architect         0.1200.5
@angular-devkit/build-angular     12.0.5
@angular-devkit/build-optimizer   0.1200.5
@angular-devkit/core              12.0.1
@angular-devkit/schematics        12.0.1
@angular/flex-layout              12.0.0-beta.34
@angular/localize                 12.0.5
@nguniversal/builders             12.1.3
@nguniversal/common               12.0.0
@nguniversal/express-engine       12.0.0
@schematics/angular               12.0.1
rxjs                              7.4.0
typescript                        4.2.4
</code></pre>

**Anything else relevant?**

alan-agius4:
What is happening here is that `extract-i18n` requires a browser builder target (build) which uses the `tsconfig.app.json`. This tsconfig doesn't include the `main.server.ts` as entry-point. 

At the moment, extraction for server target is not a supported use case and hence I am marking this as a feature request.

manzonif:
I understand, thanks.
Do you think there is some work around to allow for that extraction?

[...]

PR #25672, opened 2023-08-11
feat(@angular-devkit/build-angular): initial i18n extraction support for application builder

clydin:
The `ng extract-i18n` command now supports using either the developer preview esbuild-based browser or application builders. Support for the existing Webpack-based build system has been maintained. The extraction process will now build the application based on the build target defined builder in the case of either `@angular-devkit/build-angular:browser-esbuild` and `@angular-devkit/build-angular:application`. In the case of the application builder, SSR output code generation is disabled to prevent duplicate messages for the same underlying source code.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/extract-i18n/application-extraction.ts
+++ b/packages/angular_devkit/build_angular/src/builders/extract-i18n/application-extraction.ts
@@ -1,0 +1,167 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import type { ɵParsedMessage as LocalizeMessage } from '@angular/localize';
+import type { MessageExtractor } from '@angular/localize/tools';
+import type { BuilderContext, BuilderOutput } from '@angular-devkit/architect';
+import assert from 'node:assert';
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/builders/extract-i18n/application-extraction.ts
```diff
@@ -0,0 +1,164 @@
...
+  // Setup the build options for the application based on the browserTarget option
+  const buildOptions = (await context.validateOptions(
+    await context.getTargetOptions(options.browserTarget),
+    builderName,
+  )) as unknown as ApplicationBuilderInternalOptions;
+  buildOptions.optimization = false;
+  buildOptions.sourceMap = { scripts: true, vendor: true };
+  if (buildOptions.ssr) {
+    buildOptions.ssr = false;
+  }
```
alan-agius4, line 41: You would also need to disable prerendering and app-shell as disabling ssr will only disable 
server side rendering but not the server build.

That said, in future we need to improve this, as there might be server only components that require translation such as the app-shell. See: https://github.com/angular/angular-cli/issues/22871

```suggestion
 buildOptions.ssr = false;
 buildOptions.appShell = false;
 buildOptions.prerender = false;
```

alan-agius4 (commented):
Just one thing.

alan-agius4 (approved):

PR #25709, opened 2023-08-24
feat(@ngtools/webpack): add automated preconnects for image domains

atcastle:
This PR adds new functionality to identify image domains used in [NgOptimizedImage](https://angular.io/api/common/NgOptimizedImage) loaders, and automatically add `link rel="preload"` tags to the index.html file for those domains, if the end-user isn't already managing their own preconnects. I designed this change with some substantial help from @alan-agius4 and @clydin. ATTN: @kara @AndrewKushnir 

### Technical details
How domains are identified:
Image domains are identified during the Typescript transformation process, where application Typescript files are already being traversed as an AST multiple times. This PR adds a new, non-modifying transformer, which conducts a narrow traversal of the AST to find image domains in the code. The traversal looks like this:
1) Starting at the root of the file, do a one-layer traversal, looking for a ClassDeclaration with name 'appModule'.
2) If found, look for the property assignment for `ɵinj`
3) If found, continue traversal to find the `providers` array
4) Visit all elements of the providers array and determine if they match one of two patterns:
  4a) Invocations of a [built-in CDN loader function](https://angular.io/guide/image-directive#built-in-loaders). The domain is extracted from the argument.
  4b) A manually-defined provider for IMAGE_LOADER with a [custom loader function](https://angular.io/guide/image-directive#custom-loaders). The function body is traversed and a regex is used on any string literals to find domain strings.

Found domains are added to a Set, which is exported from `@ngtools/webpack`, and imported cross-package into `packages/angular_devkit/build_angular/src/builders/browser/index.ts`. From there, the domains are used during index.html generation to add preload tags.

### Performance impacts
I've designed the changes in this PR so that they should have minimal effect on build times. The AST traversal is narrow in scope, and will quit early if it doesn't encounter the specific pattern it's looking for.

### Testing 
Unit tests are provided for both the new transform (`find_image_domains.ts`) and for the modifications to `augment-index-html.ts`. This feature really wants E2E testing (as it includes changes at both ends of the build process), but I couldn't find a good example of similar E2E tests for Ivy code in angular-cli. I'd be interested in feedback on this, and will be happy to add E2E tests with some guidance. 

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
@@ -40,6 +40,7 @@ export interface AugmentIndexHtmlOptions {
   /** Used to set the document default locale */
   lang?: string;
   hints?: { url: string; mode: string; as?: string }[];
+  imageDomains?: string[];
 }
 
 export interface FileInfo {
@@ -53,10 +54,21 @@ export interface FileInfo {
  * after processing several configurations in order to build different sets of
  * bundles for differential serving.
  */
+// eslint-disable-next-line max-lines-per-function
 export async function augmentIndexHtml(
   params: AugmentIndexHtmlOptions,
 ): Promise<{ content: string; warnings: string[]; errors: string[] }> {
-  const { loadOutputFile, files, entrypoints, sri, deployUrl = '', lang, baseHref, html } = params;
+  const {
+    loadOutputFile,
+    files,
+    entrypoints,
+    sri,
+    deployUrl = '',
+    lang,
+    baseHref,
+    html,
+    imageDomains,
+  } = params;
 
   const warnings: string[] = [];
   const errors: string[] = [];
@@ -175,6 +187,7 @@ export async function augmentIndexHtml(
   const dir = lang ? await getLanguageDirection(lang, warnings) : undefined;
   const { rewriter, transformedContent } = await htmlRewritingStream(html);
   const baseTagExists = html.includes('<base');
+  const foundPreconnects = new Set<string>();
 
   rewriter
     .on('startTag', (tag) => {
@@ -204,6 +217,13 @@ export async function augmentIndexHtml(
             updateAttribute(tag, 'href', baseHref);
           }
           break;
+        case 'link':
+          if (readAttribute(tag, 'rel') === 'preconnect') {
+            const href = readAttribute(tag, 'href');
+            if (href) {
+              foundPreconnects.add(href);
+            }
+          }
       }
 
       rewriter.emitStartTag(tag);
@@ -214,7 +234,13 @@ export async function augmentIndexHtml(
           for (const linkTag of linkTags) {
             rewriter.emitRaw(linkTag);
           }
-
+          if (imageDomains) {
+            for (const imageDomain of imageDomains) {
+              if (!foundPreconnects.has(imageDomain)) {
+                rewriter.emitRaw(`<link rel="preconnect" href="${imageDomain}" data-ngimg>`);
+              }
+            }
+          }
           linkTags = [];
           break;
         case 'body':
@@ -263,6 +289,15 @@ function updateAttribute(
   } else {
     tag.attrs[index] = newValue;
   }
+}
+
+function readAttribute(
+  tag: { attrs: { name: string; value: string }[] },
+  name: string,
+): string | undefined {
+  const targetAttr = tag.attrs.find((attr) => attr.name === name);
+
+  return targetAttr ? targetAttr.value : undefined;
 }
 
 function isString(value: unknown): value is string {
--- a/packages/ngtools/webpack/src/transformers/find_image_domains.ts
+++ b/packages/ngtools/webpack/src/transformers/find_image_domains.ts
@@ -1,0 +1,126 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as ts from 'typescript';
+
+const TARGET_TEXT = '@NgModule';
+const BUILTIN_LOADERS = new Set([
+  'provideCloudflareLoader',
+  'provideCloudinaryLoader',
+  'provideImageKitLoader',
+  'provideImgixLoader',
+]);
+const URL_REGEX = /(https?:\/\/[^/]*)\//g;
+
+export function findImageDomains(imageDomains: Set<string>): ts.TransformerFactory<ts.SourceFile> {
+  return (context: ts.TransformationContext) => {
+    return (sourceFile: ts.SourceFile) => {
+      const isBuiltinImageLoader = (node: ts.CallExpression): Boolean => {
+        return BUILTIN_LOADERS.has(node.expression.getText());
+      };
+
+      const findDomainString = (node: ts.Node) => {
+        if (
+          ts.isStringLiteral(node) ||
+          ts.isTemplateHead(node) ||
+          ts.isTemplateMiddle(node) ||
+          ts.isTemplateTail(node)
+        ) {
+          const domain = node.text.match(URL_REGEX);
+          if (domain && domain[0]) {
+            imageDomains.add(domain[0]);
+
+            return node;
+          }
+        }
+        ts.visitEachChild(node, findDomainString, context);
+
+        return node;
+      };
+
+      function isImageProviderKey(property: ts.ObjectLiteralElementLike): boolean {
+        return (
+          ts.isPropertyAssignment(property) &&
+          property.name.getText() === 'provide' &&
+          property.initializer.getText() === 'IMAGE_LOADER'
+        );
+      }
+
+      function isImageProviderValue(property: ts.ObjectLiteralElementLike): boolean {
+        return ts.isPropertyAssignment(property) && property.name.getText() === 'useValue';
+      }
+
+      function checkForDomain(node: ts.ObjectLiteralExpression) {
+        if (node.properties.find(isImageProviderKey)) {
+          const value = node.properties.find(isImageProviderValue);
+          if (value && ts.isPropertyAssignment(value)) {
+            if (
+              ts.isArrowFunction(value.initializer) ||
+              ts.isFunctionExpression(value.initializer)
+            ) {
+              ts.visitEachChild(node, findDomainString, context);
+            }
+          }
+        }
+      }
+
+      function findImageLoaders(node: ts.Node) {
+        if (ts.isCallExpression(node)) {
+          if (isBuiltinImageLoader(node)) {
+            const firstArg = node.arguments[0];
+            if (ts.isStringLiteralLike(firstArg)) {
+              imageDomains.add(firstArg.text);
+            }
+          }
+        } else if (ts.isObjectLiteralExpression(node)) {
+          checkForDomain(node);
+        }
+
+        return node;
+      }
+
+      function findPropertyAssignment(node: ts.Node) {
+        if (ts.isPropertyAssignment(node)) {
+          if (ts.isIdentifier(node.name) && node.name.escapedText === 'providers') {
+            ts.visitEachChild(node.initializer, findImageLoaders, context);
+          }
+        }
+
+        return node;
+      }
+
+      function findPropertyDeclaration(node: ts.Node) {
+        if (
+          ts.isPropertyDeclaration(node) &&
+          ts.isIdentifier(node.name) &&
+          node.name.escapedText === 'ɵinj' &&
+          node.initializer &&
+          ts.isCallExpression(node.initializer) &&
+          node.initializer.arguments[0]
+        ) {
+          ts.visitEachChild(node.initializer.arguments[0], findPropertyAssignment, context);
+        }
+
+        return node;
+      }
+
+      // Continue traversal if node is ClassDeclaration and has name "AppModule"
+      function findClassDeclaration(node: ts.Node) {
+        if (ts.isClassDeclaration(node)) {
+          ts.visitEachChild(node, findPropertyDeclaration, context);
+        }
+
+        return node;
+      }
+
+      ts.visitEachChild(sourceFile, findClassDeclaration, context);
+
+      return sourceFile;
+    };
+  };
+}
--- a/packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
+++ b/packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
@@ -1,0 +1,179 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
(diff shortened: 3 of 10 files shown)


packages/ngtools/webpack/src/transformers/find_image_domains.ts
```diff
@@ -0,0 +1,138 @@
...
+import * as ts from 'typescript';
+
+const TARGET_TEXT = '@NgModule';
+const BUILTIN_LOADERS = new Set([
+  'provideCloudflareLoader',
+  'provideCloudinaryLoader',
+  'provideImageKitLoader',
+  'provideImgixLoader',
+]);
+const URL_REGEX = /(https?:\/\/[^/]*)\//g;
```
atcastle, line 18: This regex is sort of a placeholder. I'll be updating this with a better one shortly, but I didn't want that to hold up discussion of the rest of the changes.

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
```diff
@@ -171,10 +183,17 @@ export async function augmentIndexHtml(
       linkTags.push(`<link ${attrs.join(' ')}>`);
     }
   }
+  const domainLinkTags: string[] = [];
+  if (imageDomains) {
+    for (const imageDomain of imageDomains) {
+      domainLinkTags.push(`<link rel="preconnect" href="${imageDomain}">`);
```
kara, line 189: Let's add a "data-" attribute here for tracking?

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
```diff
@@ -204,6 +223,10 @@ export async function augmentIndexHtml(
             updateAttribute(tag, 'href', baseHref);
           }
           break;
+        case 'link':
+          if (!foundPreconnect && readAttribute(tag, 'rel') === 'preconnect') {
```
kara, line 227: Am I reading this right that we don't generate preconnects to the image domain if any preconnect tags are found?  It seems fairly common for there to be unrelated preconnects, e.g. for fonts or GA. Should we have a check more specific to the image domains?

packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
```diff
@@ -0,0 +1,153 @@
...
+            NgOptimizedImage] }); })();
+    `;
+}
+
+describe('@ngtools/webpack transformers', () => {
+  describe('find_image_domains', () => {
+    it('should find a domain when a built-in loader is used with a string-literal-like argument', () => {
+      // Intentionally inconsistent use of quote styles in this data structure:
+      const builtInLoaders: Array<[string, string]> = [
+        ['provideCloudflareLoader("www.cloudflareUrl.com")', 'www.cloudflareUrl.com'],
```
kara, line 68: Super nit: the capitalization of "Url" in the URL feels a bit odd - lower case it?

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
```diff
@@ -40,6 +40,7 @@ export interface AugmentIndexHtmlOptions {
   /** Used to set the document default locale */
   lang?: string;
   hints?: { url: string; mode: string; as?: string }[];
+  imageDomains?: string[];
```
clydin, line 43: You should be able to use the `hints` option to add the preconnects:
``
hints: [{ mode: 'preconnect', url: 'https://example.com' }, ...],
``

packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
```diff
@@ -0,0 +1,153 @@
...
+      for (const loader of builtInLoaders) {
+        const input = inputTemplate(loader[0]);
+        const result = Array.from(findDomains(input));
+        expect(result.length).toBe(1);
+        expect(result[0]).toBe(loader[1]);
+      }
+    });
+
+    it('should not find a domain when a built-in loader is used with a variable', () => {
+      const input = inputTemplate(`provideCloudflareLoader(myImageCDN)`);
```
kara, line 85: Thought for docs: we should probably add a note that recommends using a raw string to allow preconnect generation to work

packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    it('should not find a domain when a built-in loader is used with an expression', () => {
+      const input = inputTemplate(
+        `provideCloudflareLoader("https://www." + (dev ? "dev." : "") + "cloudinaryUrl.net")`,
+      );
+      const result = Array.from(findDomains(input));
+      expect(result.length).toBe(0);
+    });
+
+    it('should not find a domain when a built-in loader is used with an template literal', () => {
```
kara, line 98:
```suggestion
    it('should not find a domain when a built-in loader is used with a template literal', () => {
```

packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
```diff
@@ -0,0 +1,153 @@
...
+      const result = Array.from(findDomains(input));
+      expect(result.length).toBe(0);
+    });
+
+    it('should not find a domain in a function that is not a built-in loader', () => {
+      const input = inputTemplate('provideNotARealLoader("https://www.foo.com")');
+      const result = Array.from(findDomains(input));
+      expect(result.length).toBe(0);
+    });
+
```
kara, line 111: It would be good to have a test where there are multiple providers (maybe 3 with the image one in the middle?)

kara (commented):
This is looking great!

packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
atcastle: Good call. I added one at the end with multiple providers/provider types

packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
atcastle: 👍

packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
atcastle: Yeah camelcase for URLs looks weird. Changed 'em

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
atcastle: Yeah, I did that on purpose originally because I was a little worried about missing almost-duplicates, like if they added a preconnect for `"www.example.com/"` and the preconnect generator adds one for `"www.example.com"`. But after thinking about it a little more, I think that's kind of a corner case, and the failure case of having two preconnects for the same domain in the index.html looks a little weird but wouldn't actually cause any harm.

I went back and added a per-domain check, instead of an overall check.

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
atcastle: Done!

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
atcastle: There's a little extra logic for image domains beyond just generic preconnects. I just pushed a change to this PR where we're [checking the image domains vs any preconnects](https://github.com/angular/angular-cli/pull/25709/files#diff-c8f27b3539a3f2924e95325c9d6105de9f37ff3b73cbcd7ca15223d461335e43R238) that the end user has manually included in the index.html file, to make sure not to add duplicates. That would be harder to do if the image domains were just mixed in with the rest of the preconnect hints. Do you think it's still worth trying to avoid using the extra data sctructure?

kara (approved):
LGTM

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
alfaproject: Cant the hints themselves be deduplicated in a generic way?

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
```diff
@@ -171,10 +183,12 @@ export async function augmentIndexHtml(
       linkTags.push(`<link ${attrs.join(' ')}>`);
     }
   }
+  const domainLinkTags: string[] = [];
```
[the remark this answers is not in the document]
atcastle: Nope, removed.

packages/ngtools/webpack/src/transformers/find_image_domains_spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  transformTypescript(input, [transformer], program, compilerHost);
+
+  return domains;
+}
+
+function inputTemplate(provider: string) {
+  /* eslint-disable max-len */
+  return tags.stripIndent`
+    export class AppModule {
```
AndrewKushnir, line 35: I'd like to propose that we add a test for standalone bootstrap pattern as well (though the `providers` array structure would be the same), so that we make sure it works for standalone cases. An input can look something like this:

```
bootstrapApplication(AppComponent, {
  providers: [ /* the format here is similar to what you have in other tests*/ ]
});
```

packages/ngtools/webpack/src/transformers/find_image_domains.ts
```diff
@@ -0,0 +1,138 @@
...
+      }
+
+      function findPropertyDeclaration(node: ts.Node) {
+        if (
+          ts.isPropertyDeclaration(node) &&
+          ts.isIdentifier(node.name) &&
+          node.name.escapedText === 'ɵinj'
+        ) {
+          if (node.initializer) {
+            if (ts.isCallExpression(node.initializer)) {
```
clydin, line 109: nit: To reduce nesting, the two inner if statements could be added to the outer if

packages/ngtools/webpack/src/transformers/find_image_domains.ts
```diff
@@ -0,0 +1,138 @@
...
+              ts.visitEachChild(node.initializer.arguments[0], findPropertyAssignment, context);
```
clydin, line 110: nit: There _should_ always be a first argument from the AOT compiler but it may be useful to assert here just in case something goes wrong.

packages/ngtools/webpack/src/transformers/find_image_domains.ts
atcastle: Yep--there was originally some more logic in there, I think. Condensed it down to one big expression.

packages/ngtools/webpack/src/transformers/find_image_domains.ts
atcastle: Sure, added it as another condition in the if, just to be safe.

packages/ngtools/webpack/src/transformers/find_image_domains.ts
```diff
@@ -0,0 +1,138 @@
...
+  'provideImageKitLoader',
+  'provideImgixLoader',
+]);
+const URL_REGEX = /(https?:\/\/[^/]*)\//g;
+
+export function findImageDomains(
+  getTypeChecker: () => ts.TypeChecker,
+  imageDomains: Set<string>,
+): ts.TransformerFactory<ts.SourceFile> {
+  const typeChecker = getTypeChecker();
```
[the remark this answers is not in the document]
atcastle: Nope, not any more. Removed.

packages/ngtools/webpack/src/transformers/find_image_domains.ts
```diff
@@ -0,0 +1,138 @@
...
+        return node;
+      }
+
+      function sourceFileVisitor(node: ts.SourceFile) {
+        ts.visitEachChild(node, findClassDeclaration, context);
+
+        return node;
+      }
+
+      ts.visitNode(sourceFile, sourceFileVisitor);
```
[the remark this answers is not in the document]
atcastle: Good catch, done.

packages/ngtools/webpack/src/transformers/find_image_domains.ts
```diff
@@ -0,0 +1,138 @@
...
+            }
+          }
+        }
+
+        return node;
+      }
+
+      // Continue traversal if node is ClassDeclaration and has name "AppModule"
+      function findClassDeclaration(node: ts.Node) {
+        if (ts.isClassDeclaration(node) && node.name?.escapedText === 'AppModule') {
```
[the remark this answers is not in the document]
atcastle: Ah, thanks for pointing this out. I tried to keep this as specific as possible to limit traversal (for performance), but I guess here it just needs to be all ClassDeclaration nodes. Since most files won't have _that_ many top-level class declarations, and the traversal should fail at the next step for any non-appmodule classes, it shouldn't cause too much additional work.

packages/angular_devkit/build_angular/src/utils/index-file/augment-index-html.ts
```diff
@@ -214,7 +234,13 @@ export async function augmentIndexHtml(
           for (const linkTag of linkTags) {
             rewriter.emitRaw(linkTag);
           }
-
+          if (imageDomains) {
+            for (const imageDomain of imageDomains) {
+              if (!foundPreconnects.has(imageDomain)) {
```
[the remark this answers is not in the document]
atcastle: Oh, maybe I misunderstood. You're saying that they've already got a crossorigin preconnect to another domain (because they're going to fetch a script or something), but they're using that same domain to load images. So in that case there would need to be both a crossorigin and a non-crossorigin preconnect link for the same domain?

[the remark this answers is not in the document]
clydin: Yes, that was my concern. Although in the image domain context this is probably more of an edge case.  Maybe a comment regarding the situation would be sufficient here.  The crossorigin attribute is one of the more confusing attributes for link elements.
If we implemented generic deduplication in the future, we can factor crossorigin into the logic at that time.

clydin (approved):

kara:
@alan-agius4 @clydin Does minor target mean this feature is going into v17? Or is there a minor planned before that?

alan-agius4:
@kara, this will go in v17 as there is no minor planned

AndrewKushnir (approved):

AndrewKushnir:
@alan-agius4 @clydin do we need a presubmit or can we just proceed with the merge?

clydin:
no presubmit needed.
I'll label it for merge.

alan-agius4:
Thanks @atcastle for this change.

PR #25712, opened 2023-08-28
ci: move Windows CI job to GHA

clydin:
The Windows E2E tests are now executed with a Github Action job via an additional entry in the existing E2E job's matrix. The CircleCI configuration previously needed to execute the job with CircleCI has also been removed.


alan-agius4 (approved):
Can you also delete https://github.com/angular/angular-cli/blob/f056487c094e528cc6aa8854d35ec64ccbc406f4/.circleci/win-ram-disk.ps1 ?

josephperrott (approved):
LGTM

PR #25713, opened 2023-08-28
build: update Angular version to v17 prerelease

clydin:
Package dependency versions and the version used to generate a new project have been updated to use `17.0.0-next.1`.


package.json
```diff
@@ -59,22 +59,22 @@
   },
   "devDependencies": {
     "@ampproject/remapping": "2.2.1",
-    "@angular/animations": "16.1.8",
+    "@angular/animations": "17.0.0-next.0",
```
crisbeto, line 62: Drive-by comment: can we bump this to `next.1` which has support for TS 5.2?

alan-agius4 (approved):
Will merge to unblock CI, but can you do a followup to also update https://github.com/angular/angular-cli/blob/981f80e5a9946f4a7c967abe3d3b00821543eab1/packages/angular/ssr/package.json#L16? 

Thanks.

PR #25771, opened 2023-09-01
ci: move snapshot E2E tests to GHA

clydin:
The E2E snapshot tests have now been migrated to use Github Actions instead
of CircleCI. The CircleCI jobs and workflow configuration for the test execution
have also been removed. The CircleCI dynamic configuration aspects are still
present to minimize the changeset complexity and will be removed in a future
change.
The snapshot tests are guarded by a new analyze job which checks for a change
to the `tests/legacy-cli/e2e/ng-snapshot/package.json` file which is used to
store the snapshot commit SHAs for the relevant Angular packages used in tests.
This job based guard approach allows the entire set of snapshot E2E jobs to be
skipped when not required. Additionally, a single skipped PR status entry is
created.


josephperrott (approved):
LGTM

PR #25776, opened 2023-09-04
Account for new compiler output in custom tranformers

crisbeto:
There's some new output coming from the Angular compiler. These changes update the various custom transformers to account for it. It includes:
1. Emitting calls to `setClassMetadataAsync` instead of `setClassMetadata` for components that use the new `defer` block syntax.
2. Using arrow functions instead of function expressions for IIFEs. This isn't merged yet, but is being explored in https://github.com/angular/angular/pull/51637.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/babel/plugins/elide-angular-metadata.ts
+++ b/packages/angular_devkit/build_angular/src/tools/babel/plugins/elide-angular-metadata.ts
@@ -58,13 +58,13 @@ export default function (): PluginObj {
         }
 
         // The metadata function is always emitted inside a function expression
-        if (!path.getFunctionParent()?.isFunctionExpression()) {
-          return;
-        }
+        const parent = path.getFunctionParent();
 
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_angular/src/tools/babel/plugins/elide-angular-metadata.ts
```diff
@@ -41,31 +47,58 @@ export default function (): PluginObj {
...
         ) {
-          return;
-        }
+          // The metadata function is always emitted inside a function expression
+          const parent = path.getFunctionParent();
 
-        // The metadata function is always emitted inside a function expression
-        if (!path.getFunctionParent()?.isFunctionExpression()) {
-          return;
+          if (parent?.isFunctionExpression() || parent?.isArrowFunctionExpression()) {
```
alan-agius4, line 59: NIT

```suggestion
          if (parent && (parent.isFunctionExpression() || parent.isArrowFunctionExpression())) {
```

PR #25799, opened 2023-09-07
feat(@schematics/angular): update` ng new` to use the esbuild based application builder by default

alan-agius4:
This commit updates the `ng generate application` to use the esbuild `application` builder. This also updates the schematics to support both `browser` and `application` builders.

BREAKING CHANGE: `rootModuleClassName`, `rootModuleFileName` and `main` options have been removed from the public `pwa` and `app-shell` schematics.


PowerKiKi:
Would you mind explaining the relationship between the new-ish `@angular-devkit/build-angular:browser-esbuild`, and the brand new `@angular-devkit/build-angular:application` ?

I was under the impression that `@angular-devkit/build-angular:browser-esbuild` would eventually replace the old `@angular-devkit/build-angular:browser`. But that now that `@angular-devkit/build-angular:application` has been released in developer preview, I am a bit confused. Does `@angular-devkit/build-angular:application`  somehow uses `@angular-devkit/build-angular:browser-esbuild` under the hood ? or will `@angular-devkit/build-angular:browser-esbuild` be replaced by `@angular-devkit/build-angular:application` even before being getting promoted as generally available ?

alan-agius4:
Under the hood the `@angular-devkit/build-angular:browser-esbuild` uses `@angular-devkit/build-angular:application`, The `browser-esbuild` is a compatibility layer as it is used to normalize several options to be compatible with the `browser` builder to allow users to switch back and forth between an esbuild and webpack based builders without changing much in builder configuration.

The `application` builder will replace the `browser-esbuild`. This builder has additional functionality which is not exposed via the `browser-esbuild` builder such capabilities to generate an app-shell, ssr and prerendering application. Hence, this will replace and the remove the need for the `prerender`, `server` and `app-shell` builders.

packages/schematics/angular/environments/index.ts
```diff
@@ -72,15 +72,18 @@ function* generateConfigurationEnvironments(
   sourceRoot: string,
   projectName: string,
 ): Iterable<Rule> {
-  if (!buildTarget.builder.startsWith(AngularBuilder.Browser)) {
+  if (
+    buildTarget.builder !== AngularBuilder.Browser &&
+    buildTarget.builder !== AngularBuilder.Application
```
clydin, line 77: Should probably put the `browser-esbuild` in here too so that we avoid false positives.

packages/schematics/angular/app-shell/index.ts
```diff
@@ -249,25 +248,14 @@ function addServerRoutes(options: AppShellOptions): Rule {
...
+    if (!project) {
+      throw new SchematicsException(`Invalid project name (${options.project})`);
     }
-    const modulePath = getServerModulePath(
-      host,
-      clientProject.sourceRoot || 'src',
-      options.main as string,
-    );
+
+    const modulePath = getServerModulePath(host, project.sourceRoot || 'src', 'main.server.ts');
```
clydin, line 256: nit: move `main.server.ts` to a constant

packages/angular/ssr/schematics/ng-add/index.ts
```diff
@@ -254,7 +303,7 @@ function routingInitialNavigationRule(options: UniversalOptions): Rule {
 }
 
 function addDependencies(): Rule {
-  return (_host: Tree) => {
+  return () => {
```
clydin, line 306: nit: can remove this and just return the chain directly if the tree parameter isn't used

clydin (approved):

PR #25812, opened 2023-09-09
Account for string literals in styles and the new styleUrls

crisbeto:
Updates the JIT transformers to account for the upcoming changes from https://github.com/angular/angular/pull/51715.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/angular/jit-resource-transformer.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/angular/jit-resource-transformer.ts
@@ -143,75 +143,90 @@ function visitComponentMetadata(
   switch (node.name.text) {
     case 'templateUrl':
       // Only analyze string literals
-      if (
-        !ts.isStringLiteral(node.initializer) &&
-        !ts.isNoSubstitutionTemplateLiteral(node.initializer)
-      ) {
+      if (!ts.isStringLiteralLike(node.initializer)) {
         return node;
       }
 
-      const url = node.initializer.text;
-      if (!url) {
-        return node;
-      }
-
-      return nodeFactory.updatePropertyAssignment(
-        node,
-        nodeFactory.createIdentifier('template'),
-        createResourceImport(
-          nodeFactory,
-          generateJitFileUri(url, 'template'),
-          resourceImportDeclarations,
-        ),
-      );
+      return node.initializer.text.length === 0
+        ? node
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/angular/jit-resource-transformer.ts
```diff
@@ -143,82 +143,93 @@ function visitComponentMetadata(
   switch (node.name.text) {
     case 'templateUrl':
       // Only analyze string literals
-      if (
```
crisbeto, line 146: Note: I couldn't figure out where this logic is tested so I didn't add any tests. I can add some if somebody can point me in the right direction.

alan-agius4: You can add a test similar to https://github.com/angular/angular-cli/blob/734a46e99ff94d5fffd93f223e4b2b7646790f78/packages/angular_devkit/build_angular/src/builders/application/tests/behavior/component-stylesheets_spec.ts or an E2E https://github.com/angular/angular-cli/tree/main/tests/legacy-cli/e2e/tests/build

crisbeto: Hmm okay, it looks like I can't write a test there, because the change hasn't landed in a release yet so writing a component with the new properties results in a compilation error.

crisbeto: Actually even if this compiled, it wouldn't be testing the behavior we're looking for, because the transformer only runs on JIT code. As far as I can tell, these tests are running AOT compilation.

alan-agius4: > Actually even if this compiled, it wouldn't be testing the behavior we're looking for, because the transformer only runs on JIT code. As far as I can tell, these tests are running AOT compilation.

You can switch between AOT and JIT by providing the `aot` as an option in https://github.com/angular/angular-cli/blob/734a46e99ff94d5fffd93f223e4b2b7646790f78/packages/angular_devkit/build_angular/src/builders/application/tests/behavior/component-stylesheets_spec.ts#L20

Let's land it for now, and follow up with a unit test in a separate PR once the changes in the FW are released on NPM.

PR #25846, opened 2023-09-14
refactor(@angular-devkit/build-angular): remove usage of terser constants in esbuild

clydin:
The terser build time constant import from the `@angular/compiler-cli` package is no longer used in the esbuild-based builder. The constants present are already defined and conditionally added within the build configuration itself. This not only provides more flexibility but also removes the need to import the package early in the process. The import is also an expensive import due to it needing TypeScript and being ESM that needs to be dynamically imported via a function helper to work around current ESM/TypeScript/CommonJS limitations.


alan-agius4 (approved):
LGTM

PR #25847, opened 2023-09-14
ci: use `head_ref` for concurrency group

alan-agius4:
Use the head_ref value for the concurrency group on `ci` rather than the ref, so that the main ref is not used and cancelling across pull requests and the main branch

Backport https://github.com/angular/angular/pull/51784 to the CLI.


josephperrott (approved):
LGTM

PR #25856, opened 2023-09-15
feat(@angular-devkit/build-angular): support standalone apps route discovery during prerendering

alan-agius4:
This fixes an issue were routes could not be discovered automatically in a standalone application.

This is a total overhaul of the route extraction process as instead of using `guess-parser` NPM package, we now use the Angular Router. This enables a number of exciting possibilities for the future which were not possible before.

# How it works?
The application is bootstrapped and through DI injection we get the injector and router config instance and recursively build the routes tree.


packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
```diff
@@ -0,0 +1,102 @@
+/**
```
alan-agius4, line 1: @AndrewKushnir and @atscott would appreciate if you could take a look at the changes in this file.

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    if (redirectTo !== undefined) {
+      // TODO: handle `redirectTo`.
+      continue;
+    }
+
+    if (/[:*]/.test(path)) {
+      // TODO: handle parameterized routes population.
+      continue;
+    }
```
AndrewKushnir, line 42: I'm wondering if we should collect those routes somewhere in a separate array and output at the end like this (may be if a developer specify a special flag like `--verbose`?):

```
...
Successfully prerendered 1546 routes!

The following routes were skipped because they contain routes with dynamic parameters:

* category/:cid/product/:pid
* product/:id

Learn more about prerendering for such routes at: https://angular.io/guides/prerendering#routes-with-params

The following routes were skipped because they contain redirects:

* path/with/redirect
* another/path/with/redirect

Learn more about prerendering for such routes at: https://angular.io/guides/prerendering#routes-with-redirects
```

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    // Extract all the routes from the config.
+    yield* getRoutesFromRouterConfig(router.config, routerConfigLoader, injector);
+  } finally {
+    platformRef.destroy();
+  }
+}
+
+function isBootstrapFn(value: unknown): value is () => Promise<ApplicationRef> {
+  // We can differentiate between a module and a bootstrap function by reading `cmp`:
```
AndrewKushnir, line 96:
```suggestion
  // We can differentiate between a module and a bootstrap function by reading compiler-generated `ɵmod` static property:
```

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
```diff
@@ -0,0 +1,102 @@
...
+import {
+  INITIAL_CONFIG,
+  ɵINTERNAL_SERVER_PLATFORM_PROVIDERS as INTERNAL_SERVER_PLATFORM_PROVIDERS,
+} from '@angular/platform-server';
+import { Route, Router, ɵROUTER_PROVIDERS } from '@angular/router';
+import { first } from 'rxjs/operators'; // Import from `/operators` to support rxjs 6 which is still supported by the Framework.
+
+// TODO(alanagius): replace the below once `RouterConfigLoader` is privately exported from `@angular/router`.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const RouterConfigLoader = ɵROUTER_PROVIDERS[5] as any;
```
AndrewKushnir, line 19: I'd propose to resolve that asap, since relying on an index would be very unstable. @atscott should we consider exposing a helper method that would take a `Route` as an input and return a list of loaded child `Route`s? That'd probably be the minimal possible API extension to support this. WDYT?

AndrewKushnir (approved):
@alan-agius4 fantastic work! 🎉

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
alan-agius4: I was planning to add a private export to `RouterConfigLoader` when this was approved.

AndrewKushnir: We can do that too, I was just wondering if there is a smaller API surface that we can expose (either publicly or via a private symbol), so that it's easier to maintain later in the Router code.

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
```diff
@@ -0,0 +1,102 @@
...
+    let applicationRef: ApplicationRef;
+    if (isBootstrapFn(bootstrapAppFnOrModule)) {
+      applicationRef = await bootstrapAppFnOrModule();
+    } else {
+      const moduleRef = await platformRef.bootstrapModule(bootstrapAppFnOrModule);
+      applicationRef = moduleRef.injector.get(ApplicationRef);
+    }
+
+    // Wait until the application is stable.
+    await applicationRef.isStable.pipe(first((isStable) => isStable)).toPromise();
```
clydin, line 82: In the future, it would be nice if we had a promise-based API alternative here. Something like `untilStable()` or similar. No piping and also no operator imports then.

packages/angular_devkit/build_angular/src/utils/routes-extractor/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
+# Copyright Google Inc. All Rights Reserved.
+#
+# Use of this source code is governed by an MIT-style license that can be
+# found in the LICENSE file at https://angular.io/license
+
+load("//tools:defaults.bzl", "ts_library")
+
+# NOTE This is built as ESM as this is included in the users server bundle.
```
clydin, line 8: Is this using a different tsconfig to all ESM output?  The main tsconfig is set to `commonjs` currently.

clydin (approved):

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
alan-agius4: I like this.

packages/angular_devkit/build_angular/src/utils/routes-extractor/BUILD.bazel
alan-agius4: Under bazel, the "module" and "target" are not configured in the tsconfig.

[is](https://github.com/angular/angular-cli/blob/main/tools/defaults.bzl#L54-L89)

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
alan-agius4: Yeah, we definitely

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
atscott: Yea, there's certainly a smaller API surface that should be exposed here. There's a _lot_ going on in the config loader including coordinating failed loads, duplicate loads with the preloader, lazy components, dev mode errors, and event emits. There are really only two bits in there you care about: executing the load callback and mapping that result to returned routes (and the injector if it was a lazy module).

atscott (approved):
Agreed with @AndrewKushnir on the config loader code sharing. I'll see what I can do for that.

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
alan-agius4: SGTM, let’s merge as is and get back to this when the API is available.

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (/[:*]/.test(path)) {
+      // TODO: handle parameterized routes population.
+      yield { route: currentRoutePath, success: false, redirect: false };
+      continue;
+    }
+
+    yield { route: currentRoutePath, success: true, redirect: false };
+
+    if (children?.length || loadChildren) {
+      yield* getRoutesFromRouterConfig(
```
atscott, line 57: BTW, as you recurse through child routes, you might need to update the injector. For now, I don't think this will matter so much but technically if a route has an injector, you should use that as the injector when loading child routes.

PR #25859, opened 2023-09-15
feat(@angular-devkit/build-angular): support i18n inlining with esbuild-based builder

clydin:
When using the esbuild-based application build system through either the `application` or `browser-esbuild` builder, the `localize` option will now allow inlining project defined localizations. The process to configure and enable the i18n system is the same as with the Webpack-based `browser` builder. The implementation uses a similar approach to the `browser` builder in which the application is built once and then post-processed for each active locale. In addition to inlining translations, the locale identifier is injected and the locale specific data is added to the applications. Currently, this implementation adds all the locale specific data to each application during the initial building. While this may cause a small increase in the polyfills bundle (locale data is very small in size), it has a benefit of faster builds and a significantly less complicated build process. Additional size optimizations to the data itself are also being considered to even further reduce impact. Also, with the eventual shift towards the standard `Intl` web APIs, the need for the locale data will become obsolete in addition to the build time code necessary to add it to the application.
While build capabilities are functional, there are several areas which have not yet been fully implemented but will be in future changes. These include console progress information, efficient watch support, and app-shell/service worker support.


packages/angular_devkit/build_angular/src/tools/esbuild/i18n-inliner-worker.ts
```diff
@@ -0,0 +1,156 @@
...
+ */
+async function loadLocalizeTools(): Promise<LocalizeUtilityModule> {
+  if (localizeToolsModule !== undefined) {
+    return localizeToolsModule;
+  }
+
+  // Load ESM `@angular/localize/tools` using the TypeScript dynamic import workaround.
+  // Once TypeScript provides support for keeping the dynamic import this workaround can be
+  // changed to a direct dynamic import.
+  return loadEsmModule('@angular/localize/tools');
```
JoostK, line 76: Drive-by:

```suggestion
  return localizeToolsModule = loadEsmModule('@angular/localize/tools');
```

packages/angular_devkit/build_angular/src/builders/application/execute-build.ts
```diff
@@ -154,14 +161,18 @@ export async function executeBuild(
   let indexContentOutputNoCssInlining: string | undefined;
 
   // Generate index HTML file
-  if (indexHtmlOptions) {
+  // If localization is enabled, index generation is handled in the inlining process.
+  // NOTE: Localization with SSR is not currently supported.
```
alan-agius4, line 165: By not supported you mean in this PR, but will be before v17 is fully supported?

Right now, i18n is possible and support when using the server builder, it's just that there is some server configuration/workarounds that users need to do support this for their production server.

packages/angular_devkit/build_angular/src/builders/application/index.ts
```diff
@@ -42,6 +42,23 @@ export async function* buildApplicationInternal(
   }
 
   const normalizedOptions = await normalizeOptions(context, projectName, options);
+
+  // Warn about prerender/ssr not yet supporting localize
+  if (
```
alan-agius4, line 47: Are you planning to get back to this into a separate PR? SSG and AppShell support i18n is a fully supported feature right now with the webpack based builders.

packages/angular_devkit/build_angular/src/tools/esbuild/application-code-bundle.ts
```diff
@@ -245,6 +283,19 @@ function getEsBuildCommonOptions(options: NormalizedApplicationBuildOptions): Bu
     jit,
   } = options;
 
+  // Ensure unique hashes for i18n translation changes when using post-process inlining.
```
alan-agius4, line 286: I am assuming that this is done so that the hash in the filename is also updated. (Maybe add this to the comment to make it more clear of the intention to force a unique hash).

Is there any other reason?

clydin: That's correct. There's custom code in the Webpack builder to do something similar.
I'll add a comment.

packages/angular_devkit/build_angular/src/builders/application/index.ts
clydin: Yes in a followup.  app shell support needs some additional refactoring and the E2E test needs a fairly substantial rework.

alan-agius4 (approved):

PR #25860, opened 2023-09-15
Initial Web Test Runner Implementation

dgp1130:
This adds a new `@angular-devkit/build-angular:web-test-runner` builder which invokes Web Test Runner to execute tests. The initial implementation is still experimental and lacks a lot of features, but it is sufficient to run tests generated by `ng new`.

There's no docs or setup instructions yet, however the main path to trying this out is to replace `@angular-devkit/build-angular:karma` with `@angular-devkit/build-angular:web-test-runner` in `angular.json`. Since Web Test Runner is intended to be a drop-in replacement for Karma, options should mostly align. Currently only a couple options are actually supported right now, but more will be added over time.

The commit messages go more in detail on the implementation. In particular, Web Test Runner integration is done via a custom HTML page which explicitly loads the user's JavaScript in a very particular order. This allows our custom Jasmine integration, sets up Zone.js to patch those symbols correctly, configures `TestBed`, and then runs the user's tests.

Currently some unrelated unit tests appear to be failing due to Web Test Runner pulling in `@types/mocha` which is conflicting with `@types/jasmine`. I'm looking into the proper fix for that.

### DIFF
--- a/packages/angular/cli/lib/config/workspace-schema.json
+++ b/packages/angular/cli/lib/config/workspace-schema.json
@@ -364,6 +364,7 @@
                       "@angular-devkit/build-angular:ng-packagr",
                       "@angular-devkit/build-angular:prerender",
                       "@angular-devkit/build-angular:jest",
+                      "@angular-devkit/build-angular:web-test-runner",
                       "@angular-devkit/build-angular:protractor",
                       "@angular-devkit/build-angular:server",
                       "@angular-devkit/build-angular:ssr-dev-server"
@@ -560,6 +561,28 @@
                   "type": "object",
                   "additionalProperties": {
                     "$ref": "../../../../angular_devkit/build_angular/src/builders/jest/schema.json"
+                  }
+                }
+              }
+            },
+            {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "builder": {
+                  "const": "@angular-devkit/build-angular:web-test-runner"
+                },
+                "defaultConfiguration": {
+                  "type": "string",
+                  "description": "A default named configuration to use when a target configuration is not provided."
+                },
+                "options": {
+                  "$ref": "../../../../angular_devkit/build_angular/src/builders/web-test-runner/schema.json"
+                },
+                "configurations": {
+                  "type": "object",
+                  "additionalProperties": {
+                    "$ref": "../../../../angular_devkit/build_angular/src/builders/web-test-runner/schema.json"
                   }
                 }
               }
--- a/packages/angular_devkit/build_angular/src/builders/web-test-runner/builder-status-warnings.ts
+++ b/packages/angular_devkit/build_angular/src/builders/web-test-runner/builder-status-warnings.ts
@@ -1,0 +1,49 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { BuilderContext } from '@angular-devkit/architect';
+import { Schema as WtrBuilderOptions } from './schema';
+
+const UNSUPPORTED_OPTIONS: Array<keyof WtrBuilderOptions> = [
+  'main',
+  'assets',
+  'scripts',
+  'styles',
+  'inlineStyleLanguage',
+  'stylePreprocessorOptions',
+  'sourceMap',
+  'progress',
+  'poll',
+  'preserveSymlinks',
+  'browsers',
+  'codeCoverage',
+  'codeCoverageExclude',
+  'fileReplacements',
+  'webWorkerTsConfig',
+  'watch',
+];
+
+/** Logs a warning for any unsupported options specified. */
+export function logBuilderStatusWarnings(options: WtrBuilderOptions, ctx: BuilderContext) {
+  // Validate supported options
+  for (const unsupportedOption of UNSUPPORTED_OPTIONS) {
+    const value = (options as unknown as WtrBuilderOptions)[unsupportedOption];
+
+    if (value === undefined || value === false) {
+      continue;
+    }
+    if (Array.isArray(value) && value.length === 0) {
+      continue;
+    }
+    if (typeof value === 'object' && Object.keys(value).length === 0) {
+      continue;
+    }
+
+    ctx.logger.warn(`The '${unsupportedOption}' option is not yet supported by this builder.`);
+  }
+}
--- a/packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
@@ -14,6 +14,7 @@ import path from 'node:path';
 import { findTestFiles } from '../../utils/test-files';
 import { buildApplicationInternal } from '../application';
 import { OutputHashing } from '../browser-esbuild/schema';
+import { logBuilderStatusWarnings } from './builder-status-warnings';
 import { WtrBuilderOptions, normalizeOptions } from './options';
 import { Schema } from './schema';
 
@@ -22,6 +23,7 @@ export default createBuilder(
     ctx.logger.warn(
       'NOTE: The Web Test Runner builder is currently EXPERIMENTAL and not ready for production use.',
     );
+    logBuilderStatusWarnings(schema, ctx);
 
     // Dynamic import `@web/test-runner` from the user's workspace. As an optional peer dep, it may not be installed
     // and may not be resolvable from `@angular-devkit/build-angular`.
--- a/packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
@@ -1,0 +1,194 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
+import type * as WebTestRunner from '@web/test-runner';
+import { promises as fs } from 'node:fs';
+import { createRequire } from 'node:module';
+import path from 'node:path';
+import { findTestFiles } from '../../utils/test-files';
+import { buildApplicationInternal } from '../application';
+import { OutputHashing } from '../browser-esbuild/schema';
+import { WtrBuilderOptions, normalizeOptions } from './options';
+import { Schema } from './schema';
+
+export default createBuilder(
+  async (schema: Schema, ctx: BuilderContext): Promise<BuilderOutput> => {
+    ctx.logger.warn(
+      'NOTE: The Web Test Runner builder is currently EXPERIMENTAL and not ready for production use.',
+    );
+
+    // Dynamic import `@web/test-runner` from the user's workspace. As an optional peer dep, it may not be installed
+    // and may not be resolvable from `@angular-devkit/build-angular`.
+    const require = createRequire(`${ctx.workspaceRoot}/`);
+    let wtr: typeof WebTestRunner;
+    try {
+      wtr = require('@web/test-runner');
+    } catch {
+      return {
+        success: false,
+        // TODO(dgp1130): Display a more accurate message for non-NPM users.
+        error:
+          'Web Test Runner is not installed, most likely you need to run `npm install @web/test-runner --save-dev` in your project.',
+      };
+    }
+
+    const options = normalizeOptions(schema);
+    const testDir = 'dist/test-out';
+
+    // Parallelize startup work.
+    const [testFiles] = await Promise.all([
+      // Glob for files to test.
+      findTestFiles(options.include, options.exclude, ctx.workspaceRoot).then((files) =>
+        Array.from(files).map((file) => path.relative(process.cwd(), file)),
+      ),
+      // Clean build output path.
+      fs.rm(testDir, { recursive: true, force: true }),
+    ]);
+
+    // Build the tests and abort on any build failure.
+    const buildOutput = await buildTests(testFiles, testDir, options, ctx);
+    if (!buildOutput.success) {
+      return buildOutput;
+    }
+
+    // Run the built tests.
+    return await runTests(wtr, `${testDir}/browser`, options);
+  },
+);
+
+/** Build all the given test files and write the result to the given output path. */
+async function buildTests(
+  testFiles: string[],
+  outputPath: string,
+  options: WtrBuilderOptions,
+  ctx: BuilderContext,
+): Promise<BuilderOutput> {
+  const entryPoints = new Set([
+    ...testFiles,
+    'jasmine-core/lib/jasmine-core/jasmine.js',
+    '@angular-devkit/build-angular/src/builders/web-test-runner/jasmine_runner.js',
+  ]);
+
+  // Extract `zone.js/testing` to a separate entry point because it needs to be loaded after Jasmine.
+  const [polyfills, hasZoneTesting] = extractZoneTesting(options.polyfills);
+  if (hasZoneTesting) {
+    entryPoints.add('zone.js/testing');
+  }
+
+  // Build tests with `application` builder, using test files as entry points.
+  // Also bundle in Jasmine and the Jasmine runner script, which need to share chunked dependencies.
+  const buildOutput = await first(
+    buildApplicationInternal(
+      {
+        entryPoints,
+        tsConfig: options.tsConfig,
+        outputPath,
+        aot: false,
+        index: false,
+        outputHashing: OutputHashing.None,
+        optimization: false,
+        externalDependencies: [
+          // Resolved by `@web/test-runner` at runtime with dynamically generated code.
+          '@web/test-runner-core',
+        ],
+        sourceMap: {
+          scripts: true,
+          styles: true,
+          vendor: true,
+        },
+        polyfills,
+      },
+      ctx,
+    ),
+  );
+
+  return buildOutput;
+}
+
+function extractZoneTesting(
+  polyfills: readonly string[],
+): [polyfills: string[], hasZoneTesting: boolean] {
+  const polyfillsWithoutZoneTesting = polyfills.filter(
+    (polyfill) => polyfill !== 'zone.js/testing',
+  );
+  const hasZoneTesting = polyfills.length !== polyfillsWithoutZoneTesting.length;
+
+  return [polyfillsWithoutZoneTesting, hasZoneTesting];
+}
+
+/** Run Web Test Runner on the given directory of bundled JavaScript tests. */
+async function runTests(
+  wtr: typeof WebTestRunner,
+  testDir: string,
+  options: WtrBuilderOptions,
+): Promise<BuilderOutput> {
+  const testPagePath = path.resolve(__dirname, 'test_page.html');
+  const testPage = await fs.readFile(testPagePath, 'utf8');
+
+  const runner = await wtr.startTestRunner({
+    config: {
+      rootDir: testDir,
+      files: [
+        `${testDir}/**/*.js`,
+        `!${testDir}/polyfills.js`,
+        `!${testDir}/chunk-*.js`,
+        `!${testDir}/jasmine.js`,
+        `!${testDir}/jasmine_runner.js`,
+        `!${testDir}/testing.js`, // `zone.js/testing`
+      ],
+      testFramework: {
+        config: {
+          defaultTimeoutInterval: 5_000,
+        },
+      },
+      nodeResolve: true,
+      port: 9876,
+      watch: options.watch ?? false,
+
+      testRunnerHtml: (_testFramework, _config) => testPage,
+    },
+    readCliArgs: false,
+    readFileConfig: false,
+    autoExitProcess: false,
+  });
+  if (!runner) {
+    throw new Error('Failed to start Web Test Runner.');
+  }
+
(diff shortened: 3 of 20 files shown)


dgp1130:
Sent https://github.com/modernweb-dev/web/pull/2523 to fix the `@types/mocha` issue in Web Test Runner.

GerritErpenstein:
> I'm waiting for this PR to land. I'm wondering, what's blocking this? Is there any estimate on if it'll be part of some Angular 17 version or will it come with 18? I would like to try this out, as an Angular app that I work on has a lot of it's testing blocked by this PR. If there's an external library where things can be iterated or a way to add this specific package from git, please let me know

I am also eagerly awaiting this feature. What I have done is to compile Angular myself and integrate it temporarily (see [developer docs](https://github.com/angular/angular-cli/blob/main/docs/DEVELOPER.md) ). In my experience so far, the integration of Web Test Runner works perfectly (big thumbs-up to @dgp1130). However, so far there are no configuration options (see [here](https://github.com/angular/angular-cli/pull/25860/files#diff-604f9ee9ce7f263aa3415fc1417e03c9020cffaa04f7e0eb825368679023d348R121-R144)), which is not a big problem for me at the moment.

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
```diff
@@ -0,0 +1,291 @@
...
+            "uniqueItems": true
+          },
+          "default": []
+        },
+        {
+          "type": "string",
+          "description": "The full path for the polyfills file, relative to the current workspace or a module specifier. Example: 'zone.js'."
+        }
+      ]
+    },
```
alan-agius4, line 32:
```suggestion
    "polyfills": {
      "description": "Polyfills to be included in the build.",
        {
          "type": "array",
          "description": "A list of polyfills to include in the build. Can be a full path for a file, relative to the current workspace or module specifier. Example: 'zone.js'.",
          "items": {
            "type": "string",
            "uniqueItems": true
          },
          "default": []
        }
    },
```

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,199 @@
...
+);
+
+/** Build all the given test files and write the result to the given output path. */
+async function buildTests(
+  testFiles: string[],
+  outputPath: string,
+  options: WtrBuilderOptions,
+  ctx: BuilderContext,
+): Promise<BuilderOutput> {
+  const jasmine = path.relative(
```
alan-agius4, line 75: Avoid using `process.cwd()` as this is not guaranteed to be the workspace root, it can be several level down, example `src/app/feature` or when `ng test` is ran into a sub-project it will be `projects/sub-project`.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,199 @@
...
+
+export default createBuilder(
+  async (schema: Schema, ctx: BuilderContext): Promise<BuilderOutput> => {
+    ctx.logger.warn(
+      'NOTE: The Web Test Runner builder is currently EXPERIMENTAL and not ready for production use.',
+    );
+    logBuilderStatusWarnings(schema, ctx);
+
+    const options = normalizeOptions(schema);
+    const testDir = 'dist/test-out';
```
alan-agius4, line 31: This is okay for now, but we should move away from writting spec files to disk and inside wire up WRT to read the files from in-memory similar to how we handle Karma and Vite.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,199 @@
...
+  testFiles: string[],
+  outputPath: string,
+  options: WtrBuilderOptions,
+  ctx: BuilderContext,
+): Promise<BuilderOutput> {
+  const jasmine = path.relative(
+    process.cwd(),
+    require.resolve('jasmine-core/lib/jasmine-core/jasmine.js'),
+  );
+  const jasmineRunner = path.relative(process.cwd(), path.join(__dirname, 'jasmine_runner.js'));
```
alan-agius4, line 79: Same as above.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  // Build tests with `application` builder, using test files as entry points.
+  // Also bundle in Jasmine and the Jasmine runner script, which need to share chunked dependencies.
+  const buildOutput = await first(
```
alan-agius4, line 83: This seems like an unwanted behaviour as this would cause watch not to function.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,199 @@
...
+  // Wait for the tests to complete and stop the runner.
+  const passed = (await once(runner, 'finished')) as boolean;
+  await runner.stop();
+
+  // No need to return error messages because Web Test Runner already printed them to the console.
+  return { success: passed };
+}
+
+/** Returns a list of polyfills copied from the given polyfills but with `zone.js` and `zone.js/testing` added if necessary. */
+function withZonePolyfills(inputPolyfills: string[]): string[] {
```
alan-agius4, line 158: I don't think we should add polyfills when users did not add them explicitly.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,199 @@
...
+import { logBuilderStatusWarnings } from './builder-status-warnings';
+import { WtrBuilderOptions, normalizeOptions } from './options';
+import { Schema } from './schema';
+
+export default createBuilder(
+  async (schema: Schema, ctx: BuilderContext): Promise<BuilderOutput> => {
+    ctx.logger.warn(
+      'NOTE: The Web Test Runner builder is currently EXPERIMENTAL and not ready for production use.',
+    );
+    logBuilderStatusWarnings(schema, ctx);
```
alan-agius4, line 28: This is not needed, since the builder is not webpack based.

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
```diff
@@ -0,0 +1,291 @@
+{
```
alan-agius4, line 1: Can you please keep up the schema, it looks like very few options are being used here.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,199 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
+// eslint-disable-next-line import/no-duplicates
+import type * as WebTestRunner from '@web/test-runner';
+// eslint-disable-next-line import/no-duplicates
+import type { EventEmitter } from '@web/test-runner';
+import { promises as fs } from 'fs';
```
alan-agius4, line 14:
```suggestion
import fs from 'node:fs/promises';
```

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,199 @@
...
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
+// eslint-disable-next-line import/no-duplicates
+import type * as WebTestRunner from '@web/test-runner';
```
alan-agius4, line 11: Consider using `typeof import('@web/test-runner')` instead.

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
```diff
@@ -0,0 +1,291 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "title": "Web Test Runner Target",
+  "description": "Web Test Runner target options for Build Facade.",
+  "type": "object",
+  "properties": {
+    "main": {
```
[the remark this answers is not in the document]
dgp1130: The options were copied from Karma's schema, but not all of them are implemented yet and the WTR builder will warn when an unsupported option is used. I'll add more support over time.

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
dgp1130: Karma supports a string case, so I think we would want WTR to support that as well?

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
dgp1130: Good point. Just passing `jasmine-core/...` as an entry point to `application` builder doesn't work and seems to be treated as a relative file path.

```
✘ [ERROR] Could not resolve "/usr/local/google/home/douglasparker/Temp/angular-cli-e2e-YT0qj2/e2e-test/test-project/jasmine-core/lib/jasmine-core/jasmine.js"
```

Using your suggested `require.resolve` approach does work though. Updated to that.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
dgp1130: Agreed, I'll need to look into whether and how to pass through a virtual file system to Web Test Runner.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
dgp1130: Watch mode is out of scope for this PR, but will definitely add support for it later.

packages/angular_devkit/build_angular/src/builders/web-test-runner/builder-status-warnings.ts
```diff
@@ -0,0 +1,48 @@
...
+  'sourceMap',
+  'progress',
+  'poll',
+  'preserveSymlinks',
+  'browsers',
+  'codeCoverage',
+  'codeCoverageExclude',
+  'fileReplacements',
+  'webWorkerTsConfig',
+];
```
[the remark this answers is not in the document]
dgp1130: Added to the list.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
dgp1130: Curious to understand what the benefit is there? Is it just to remove the duplicate import and lint exception? I don't feel that strongly, just wondering if there's an advantage to this I'm not seeing.

I did realize we can use `WebTestRunner.EventEmitter` and drop the second import, so I consolidated those two imports together and dropped the lint comment.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
dgp1130: I dropped `withZonePolyfills` altogether, so we should just directly respect the user's provided `polyfills` array.

You pointed me at [Karma's implementation](https://github.com/angular/angular-cli/blob/7a50df5c046b65bb196408dd2c4771d5deccd034/packages/angular_devkit/build_angular/src/tools/webpack/configs/common.ts#L133-L161) separately and it feels like this is effectively moving a `zone.js/testing` polyfill to the end of the start of the main test bundle.

I updated this PR to make a new entry point for `zone.js/testing` and load that after Jasmine. I believe that should align with Karma's file ordering behavior. I had to make a small edit to `application` builder to allow bare module specifiers in entry points because `require.resolve('zone.js/testing')` was giving me the UMD module. Side effect of that is that I can take your original suggestion in the other comment for using `jasmine-core/...` and `.../jasmine_runner.js` as entry points directly and drop their `require.resolve()` calls.

I need to double check that this doesn't break Jest along the way, since that was the existing usage of this, but otherwise I think this looks good.

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
alan-agius4: It comes down if we want to have the same options or not. IE: we want users to switch back and forth from karma.

Since, this is a new builder, we can update and remove unneeded options, and simplify certain options. Similar to how we did for the application builder were one of the changes was to not support the string format.

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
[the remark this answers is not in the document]
alan-agius4: Is the intention here to allow users to switch back and forth to karma/simplify manual transition?

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
alan-agius4: Yes, it's most to avoid the duplicate import and lint exception

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,194 @@
...
+    const testDir = 'dist/test-out';
+
+    // Parallelize startup work.
+    const [testFiles, wtr] = await Promise.all([
+      // Glob for files to test.
+      findTestFiles(options.include, options.exclude, ctx.workspaceRoot).then((files) =>
+        Array.from(files).map((file) => path.relative(process.cwd(), file)),
+      ),
+      // Dynamic import `@web/test-runner` because as an optional peer dep, it may not be installed.
+      import('@web/test-runner').catch(() => undefined),
```
alan-agius4, line 37: Since this is installed by the users using `import('@web/test-runner')` will have an incorrect context as it will be loaded relative to this file instead of the users workspace. 

```js
 const workspaceRootRequire = createRequire(workspaceRoot + '/');
 workspaceRootRequire('@web/test-runner');
```

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,194 @@
...
+    // Build the tests and abort on any build failure.
+    const buildOutput = await buildTests(testFiles, testDir, options, ctx);
+    if (!buildOutput.success) {
+      return buildOutput;
+    }
+
+    // Run the built tests.
+    const testOutput = await runTests(wtr, `${testDir}/browser`, options);
+
+    return testOutput;
```
alan-agius4, line 61:
```suggestion
    return runTests(wtr, `${testDir}/browser`, options);
```

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,194 @@
...
+        externalDependencies: [
+          // Resolved by `@web/test-runner` at runtime with dynamically generated code.
+          '@web/test-runner-core',
+        ],
+        sourceMap: {
+          scripts: true,
+          styles: true,
+          vendor: true,
+        },
+        polyfills: polyfills,
```
alan-agius4, line 105: NIT
```suggestion
        polyfills,
```

alan-agius4 (commented):
Can you also please update https://github.com/angular/angular-cli/blob/6999c2e55edd245030c3017bbc410dfe956cdd29/packages/angular/cli/lib/config/workspace-schema.json#L346-L410?

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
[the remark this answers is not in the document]
dgp1130: More or less. We'll want this to be as close to Karma as we can. It may not be as perfectly straightforward as "update `@angular-devkit/build-angular:karma` -> `@angular-devkit/build-angular:web-test-runner`. It may not be exactly 1:1 and we'll have to decide on divergences as they arise, but I think this is a good starting point at least.

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
dgp1130: We'll need to make those decisions as we start looking at specific options. I'm definitely open to changing and cleaning things up here, however supporting a schematic migration from Karma to WTR is in scope here, so whatever changes we do make will need to be automatable. I don't think we need to worry too much about users switching _back_ to Karma, but we do want to make sure the migration to WTR is as smooth as we can make it.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
dgp1130: Good call. I'm curious why the trailing slash? You clearly added that intentionally and it does appear to be load-bearing in my testing, though I couldn't find any documentation on it. Does it treat that file path part as file and not a directory, or something like that?

dgp1130:
> Can you also please update
> 
> https://github.com/angular/angular-cli/blob/6999c2e55edd245030c3017bbc410dfe956cdd29/packages/angular/cli/lib/config/workspace-schema.json#L346-L410
> 
> ?

Ah totally forgot about that again. Updated.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
alan-agius4: Yeah correct, it is because the `createRequire` first parameter is a filepath.

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
alan-agius4: This migration should remain unaffected as long as the implementations are incorporated later.

Since this is an experimental phase and we won't be generating projects using this version, I'm comfortable leaving it unchanged for now. However, users opting for this builder might face some challenges as options will undergo changes in the future.

Therefore, I see two potential paths forward:
1. Remove all presently unused options and introduce them gradually as their implementations are integrated, aligning them with the application builder options.
2. Maintain the current setup and revise or remove the options before finalizing this builder. This, naturally, implies more disruptions for users already using WTR.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
```diff
@@ -0,0 +1,196 @@
...
+  for await (const value of generator) {
+    return value;
+  }
+
+  throw new Error('Expected generator to emit at least once.');
+}
+
+/** Listens for a single emission of an event and returns the value emitted. */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function once<Map extends Record<string, any>, EventKey extends string & keyof Map>(
```
clydin, line 185: Is this because wtr doesn't expose an actual EventEmitter?

packages/angular_devkit/build_angular/src/builders/web-test-runner/schema.json
dgp1130: I'm not too concerned about disrupting experimental users, that is the contract we have for experimental. I was mainly following the `browser-esbuild` precedent of having compatible options with warnings for unsupported ones. I think this will evolve over time during the experimental period, and as long as we nail it down prior to stable / dev preview, I'm not too concerned about the churn.

packages/angular_devkit/build_angular/src/builders/web-test-runner/index.ts
dgp1130: Interesting, I wasn't aware of a built-in API for this.

Web Test Runner seems to be using that under the hood, but doesn't expose it. Their `EventEmitter` does not extend or implement Node's `EventEmitter`: https://github.com/modernweb-dev/web/blob/245123e1b92fe0d468fc7ae1e25663f52e775bea/packages/test-runner-core/src/utils/EventEmitter.ts

Those utilities seem nice, and _maybe_ it's compatible enough for them to work, but I'm hesitant to rely on that if we can avoid it.

This is somewhat temporary anyways, once we get to a real watch mode, `once` will go away entirely.

packages/angular_devkit/build_angular/src/builders/web-test-runner/options.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Schema as WtrBuilderSchema } from './schema';
+
+/**
+ * Options supported for the Jest builder. The schema is an approximate
```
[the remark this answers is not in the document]
dgp1130: Ah, caught my copy-paste. Updated.

clydin (approved):

packages/angular/cli/lib/config/workspace-schema.json
```diff
@@ -362,6 +362,7 @@
                       "@angular-devkit/build-angular:extract-i18n",
                       "@angular-devkit/build-angular:karma",
                       "@angular-devkit/build-angular:jest",
+                      "@angular-devkit/build-angular:web-test-runner",
```
alan-agius4, line 365: You missing adding the configuration. Example
https://github.com/angular/angular-cli/blob/08b86d25850d43d6fa5c1549f53f9c9dee89712e/packages/angular/cli/lib/config/workspace-schema.json#L544-L565

alan-agius4 (approved):

PR #25869, opened 2023-09-20
refactor(@angular-devkit/build-angular): use `ɵloadChildren` helper from router package

alan-agius4:
This commit updates the routes extractor to use the newly exported private `ɵloadChildren` method from the router to executes a `route.loadChildren` callback and return an array of child routes.

See: https://github.com/angular/angular/pull/51818

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
+++ b/packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
@@ -6,18 +6,20 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { ApplicationRef, Injector, Type, createPlatformFactory, platformCore } from '@angular/core';
+import {
+  ApplicationRef,
+  Compiler,
+  Injector,
+  Type,
+  createPlatformFactory,
+  platformCore,
+} from '@angular/core';
 import {
   INITIAL_CONFIG,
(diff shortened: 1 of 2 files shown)


alan-agius4:
CI is red as we are waiting for an Angular release.

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
```diff
@@ -53,13 +55,21 @@ async function* getRoutesFromRouterConfig(
...
+
+    if (loadChildren) {
+      const loadedChildRoutes = await loadChildrenHelper(
+        route,
+        compiler,
+        parentInjector,
+      ).toPromise();
+
+      if (loadedChildRoutes) {
+        const { routes: childRoutes, injector = parentInjector } = loadedChildRoutes;
```
atscott, line 70: the returned injector might be `undefined` here

atscott (approved):

packages/angular_devkit/build_angular/src/utils/routes-extractor/extractor.ts
alan-agius4: Yeah, I am
Aware of that, and that’s why it defaults to the `parentInjector`

atscott: Oh, I just was totally unfamiliar with the syntax 😅

PR #25884, opened 2023-09-21
feat(@schematics/angular): enable standalone by default in new applications

alan-agius4:
This commit update the schematics to generate a standalone application by default.


clydin (approved):
Nice job cleaning up all those tests.

PR #25889, opened 2023-09-22
 feat(@schematics/angular): enable hydration when adding SSR, SSG or AppShell     

alan-agius4:
This commits updates the internal server schematic that is used by SSR and AppShell schematics to include `provideClientHydration` in the list of providers, which causes hydration to be enabled for application.
    
For more information about `provideClientHydration`, see: https://angular.io/api/platform-browser/provideClientHydration

FYI @AndrewKushnir


alan-agius4:
Please ignore the first 2 commits as this is on top of https://github.com/angular/angular-cli/pull/25884

clydin (approved):

PR #25899, opened 2023-09-25
feat(@schematics/angular): add `ng new --ssr`

alan-agius4:
This commit enabled users to opt-in adding SSR and SSG to their application during the `ng new` experience. This can be done either by using the `--ssr` option or answer `Yes` when prompted.


AndrewKushnir (approved):
👍

PR #25916, opened 2023-09-28
refactor(@angular-devkit/build-angular): use `whenStable` private API in route extractor

alan-agius4:
This commit refactors the route extractor to use the newly `whenStable` API.


AndrewKushnir (approved):
👍

PR #25926, opened 2023-09-29
test(@angular/cli): remove Safari 15 test

alan-agius4:
Safari v15 is longer included in the default browserlist configuration of CLI apps.


devversion (approved):

alfaproject:
@alan-agius4 should the name of the test file be updated or even the file removed altogether? It still mentions Safari 15

alan-agius4:
Safari 15 is still support but not by default. Users needs to opt-in by adding it in their browserslist config.

Issue #25924, opened 2023-09-29
Component ID generation collision detected is thrown after first change when running default angular installation


eneajaho:
### Command

serve

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

Component ID generation collision detected is thrown after first change when running default angular installation

### Minimal Reproduction

1. npm create @angular@next 
2. Choose ssr to be enabled
3. ng serve
4. change app component html -> no warning
5. change app component html again -> warning is shown in console

![CleanShot 2023-09-29 at 11 56 57@2x](https://github.com/angular/angular-cli/assets/25394362/0167ebf0-9f7a-49a3-b9cf-99b2437b087a)

### Your Environment

```text
Angular CLI: 17.0.0-next.6
Node: 18.17.0
Package Manager: npm 9.6.7
OS: darwin arm64

Angular: 17.0.0-next.6
... animations, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, platform-server
... router, ssr

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1700.0-next.6
@angular-devkit/build-angular   17.0.0-next.6
@angular-devkit/core            17.0.0-next.6
@angular-devkit/schematics      17.0.0-next.6
@schematics/angular             17.0.0-next.6
rxjs                            7.8.1
typescript                      5.2.2
zone.js                         0.14.0
```

PR #25927, opened 2023-09-29
fix(@angular-devkit/build-angular): clean up internal Angular state during rendering SSR

alan-agius4:
This commit clean ups the compiled components state when the build is being executed in watch mode. This is required as otherwise during development `Component ID generation collision detected` are displayed on the server.

Closes: #25924


AndrewKushnir (approved):
👍

PR #25930, opened 2023-09-30
build: re-enable Material tests

crisbeto:
Re-enables Material tests that had to be disabled during the TS 5.2 update.


crisbeto:
I suspect this shows a breakage introduced in https://github.com/crisbeto/material2/commit/6d4051797150721b35858786ce41c53b98870718. It didn't show up since the breakage came up while the test was disabled.

alan-agius4:
@crisbeto looks like https://github.com/angular/angular-cli/blob/a9c80e00d7d93459f5efdb7b2de2127ca9468d7a/tests/legacy-cli/e2e/tests/build/material.ts needs to be updated since ng-new apps are not standalone by default.

crisbeto:
@alan-agius4 it's fixed now.

alan-agius4 (approved):

PR #25938, opened 2023-10-03
feat(@angular-devkit/build-angular): support i18n with service worker and app-shell with esbuild builders

clydin:
When using the esbuild-based application build system through the `application` builder, the `localize` option will now allow inlining project defined localizations when using the app shell, prerendering, and service worker features. Previously a warning was issued and these features were disabled when the `localize` option was enabled.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/application/index.ts
@@ -43,20 +43,12 @@ export async function* buildApplicationInternal(
 
   const normalizedOptions = await normalizeOptions(context, projectName, options);
 
-  // Warn about prerender/ssr not yet supporting localize
-  if (
-    normalizedOptions.i18nOptions.shouldInline &&
-    (normalizedOptions.prerenderOptions ||
-      normalizedOptions.ssrOptions ||
-      normalizedOptions.appShellOptions)
-  ) {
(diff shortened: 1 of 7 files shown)


packages/angular_devkit/build_angular/src/builders/application/index.ts
```diff
@@ -43,20 +43,12 @@ export async function* buildApplicationInternal(
...
   const normalizedOptions = await normalizeOptions(context, projectName, options);
 
-  // Warn about prerender/ssr not yet supporting localize
-  if (
-    normalizedOptions.i18nOptions.shouldInline &&
-    (normalizedOptions.prerenderOptions ||
-      normalizedOptions.ssrOptions ||
-      normalizedOptions.appShellOptions)
-  ) {
+  // Warn about SSR not yet supporting localize
```
alan-agius4, line 46: I am not fully understanding why this is not supported. To prerender, the i189n files are already inlined. Thus, this should also work no?

alan-agius4 (approved):
LGTM, just one question about SSR.

packages/angular_devkit/build_angular/src/builders/application/index.ts
clydin: The inlining should work but wanted to wait for the application builder output path updates first to avoid making any incorrect assumptions in the inlining code.

PR #25958, opened 2023-10-05
refactor(@angular-devkit/build-angular): remove hardcoded Node.js version in application builder

alan-agius4:
This commit removed the hard coded Node.js version in application builder server config and instead passes the Angular CLI supported Node.js versions that are currently stamped using Bazel.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/utils.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/utils.ts
@@ -14,6 +14,7 @@ import fs from 'node:fs/promises';
 import path from 'node:path';
 import { promisify } from 'node:util';
 import { brotliCompress } from 'node:zlib';
+import { coerce } from 'semver';
 import { Spinner } from '../../utils/spinner';
 import { BundleStats, generateBuildStatsTable } from '../webpack/utils/stats';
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/utils.ts
```diff
@@ -290,3 +291,18 @@ export function transformSupportedBrowsersToTargets(supportedBrowsers: string[])
...
+const SUPPORTED_NODE_VERSIONS = '0.0.0-ENGINES-NODE';
+
+/**
+ * Transform supported Node.js versions to esbuild target.
+ * @see https://esbuild.github.io/api/#target
+ */
+export function getSupportedNodeTargets(): string[] {
+  if (SUPPORTED_NODE_VERSIONS.charAt(0) === '0') {
+    // Unlike `pkg_npm`, `ts_library` which is used to run unit tests does not support substitutions.
+    return [];
```
clydin, line 304: Do unit tests need a node version here?

clydin: maybe use `node18.13` as a safe default?

clydin (approved):

PR #25967, opened 2023-10-06
feat(@angular-devkit/build-angular): add initial support for bundle budgets to esbuild builders

clydin:
The bundle budget functionality (`budgets` option) is not available when using the esbuild-based builders (`browser-esbuild`/`application`). The existing option format from the Webpack-based builder can continue to be used. All budget types except `anyComponentStyle` are implemented. Any usage of the `anyComponentStyle` type will be ignored when calculating budget failures. This type will be implemented in a future change.


alan-agius4 (approved):
LGTM.

PR #25976, opened 2023-10-06
feat(@angular-devkit/build-angular): support component style budgets in esbuild builders

clydin:
The `anyComponentStyle` budget type is now supported when using bundle budgets in the esbuild-based builders (`browser-esbuild`/`application`). The configuration and behavior are intended to match that of the Webpack-based  builder (`browser`). With this addition, the bundle budget feature is now at feature parity with the Webpack-based builder. The implementation of this budget type requires less code than the Webpack implementation due to the component stylesheet information being present in the application's output metadata. This removes the need for a custom budget plugin to extract the stylesheet size information during the build.


alan-agius4 (approved):
So much simpler with Esbuild!

Issue #25556, opened 2023-07-18
Expose internal application builder function


bgotink:
### Command

build

### Description

The new experimental `buildApplicationInternal` function has a very flexible API. Its support for passing multiple entry points is especially interesting.

Scenario: I'm building a regular old website (MPA), documentation for a component library. There are over 100 components, with each multiple examples that I'd like to run live on the site.
The documentation site is not an Angular SPA, but set up via a "markdown -> static site" generator ([mdbook](https://rust-lang.github.io/mdBook/)).

### Describe the solution you'd like

Expose the internal application builder function, allowing me to build the entire documentation with multiple entry-points—each with its own bootstrap—in one go.

```ts
export default createBuilder((input, context) =>
	buildApplicationInternal( // <-- using internal import here
		{
			watch: false,
			progress: false,

			index: false,
			entryPoints: collectEntryPoints(context),

			// more options
		},
		context,
	),
);
```

### Describe alternatives you've considered

Past solutions for our documentation:

1. Make the documentation a regular angular app. That's how we started out, turns out no one likes writing documentation like this.
2. Copy the aio setup from angular/angular and make our own "markdown pages -> angular app" pipeline. Tried that, but the setup of aio is&hellip; not the easiest to understand or maintain if you're not used to it.
3. Use storybook. That's where we're currently at.

Alternatives to exposing the internal application builder function:

4. Build every page into a separate angular app. This makes the build prohibitively slow.
5. Build a single angular app with bootstraps for every documentation page. This would either load all code for all pages for any page, or it would require quite a lot of shenanigans to ensure we don't load all that superfluous javascript over a shabby internal corporate proxy server.
6. Make the multiple entry points available in the existing experimental application builder's schema. For me this solution would work equally well, but I believe the use case (running angular in an MPA) to be so different from what a regular angular app looks like that adding the option will only lead to confusion for developers who just want to build SPAs.
7. Make a separate "MPA application" builder. This puts the onus on the angular team to properly support building Angular MPAs, which I don't expect to align with the goals of the framework.

bgotink:
@clydin @alan-agius4 Thanks for exposing the `buildApplication` function!

If I read the changes in #26011 correctly, I believe the interface of this function doesn't currently support multiple entryPoints, which would be necessary for my use case.
Is there interest on your side to expose this part of the API?

Issue #24163, opened 2022-10-31
Allow transforming ESBuild configuration (@angular-devkit/build-angular:browser-esbuild) and align interface with the rest of builders


arturovt:
### Command

build

### Description

The `@angular-devkit/build-angular:browser-esbuild` builder is missing the ability to transform the ESBuild configuration as it's done in `@angular-devkit/build-angular:browser`. The `executeBrowserBuilder` function from the `@angular-devkit/build-angular:browser` builder accepts the `transforms` object as an optional argument at the end with the `webpackConfiguration` property. This allows many packages, such as `@angular-builders/custom-webpack`, to extend the Webpack configuration and then call `executeBrowserBuilder` at the end.

### Describe the solution you'd like

The `buildEsbuildBrowser` function could have a third `transforms` argument with properties like `esbuildConfiguration` (like `transforms.webpackConfiguration`) and have a type of `ExecutionTransformer<esbuild.BuildOptions>`. This will basically pass the base configuration into the transformer and allow customizing it, then pass the configuration back to `bundleCode -> esbuild.build`.

That would align the builder interface with the rest of builders (`:browser`, `:server`, `:karma`, and `extract-i18n`).

clydin:
Due to the experimental status of the builder and the potential for internal changes, we are not currently exposing a similar set of transformation options.  However, as the builder moves towards a stable status, we will be reevaluating what types of customization options will be provided.
To help us better provide a flexible set of options, would it be possible to provide more details as to your project use cases in this area?

arturovt:
@clydin I'm currently using the `@angular-builders/custom-webpack` package, which allows custom Webpack configuration and merges it with the base one. It uses `transform.webpackConfiguration` internally to do that. I have custom Webpack loaders that do some code transformations. I'd want to do the same for the ESBuild builder (by using ESBuild plugins). We could add a new 3rd party package, `@angular-builders/custom-esbuild`, that would allow proving custom ESBuild configuration and merging it with the base one. We're currently unable to implement that package because the `browser-esbuild` builder functionality is a bit restricted and we're not able to hook in it (like it's possible to do with `:browser` or `:server` builder). A `transform` object will also align the builder interface with other existing builders. I thought ESBuild builder should've had this option since it's already done in other builders, and it's a standard.

alan-agius4:
[...]

PR #26011, opened 2023-10-11
refactor(@angular-devkit/build-angular): export application builder for experimental programmatic usage

clydin:
Similar to the existing Webpack-based `browser` builder, the new `application` builder is also exported from the `@angular-devkit/build-angular` package for use programmatically. As is the case for the existing builder JavaScript exports from the package, the new export (`buildApplication`) is also considered experimental and does not provide the support nor semver guarantees that the builders have when used via `angular.json` configuration. The usage of the `plugins` parameter of the `buildApplication` allows adding esbuild compatible plugins to the end of the plugin list for the main application code bundling. However, usage of the parameter may result in unexpected application output and/or build failures. Stable and supported methods for build process extension are being evaluated for a future release.

Closes https://github.com/angular/angular-cli/issues/25556 and closes https://github.com/angular/angular-cli/issues/24163


alan-agius4 (approved):
LGTM

PR #26016, opened 2023-10-11
fix(@angular-devkit/build-angular): limit concurrent output file writes with application builder

clydin:
When using localization with the application builder for a large amount of locales, the number of files that may need to be written to disk can become large. This may be problematic on certain operating systems depending on the open file process limits. To avoid approaching the limit, the number of concurrent writes is now limited to 64.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/utils.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/utils.ts
@@ -175,40 +175,67 @@ export function getFeatureSupport(target: string[]): BuildOptions['supported'] {
   return supported;
 }
 
+const MAX_CONCURRENT_WRITES = 64;
+
 export async function writeResultFiles(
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/utils.ts
```diff
@@ -175,25 +175,39 @@ export function getFeatureSupport(target: string[]): BuildOptions['supported'] {
...
+
+    const actions = [];
+    while (fileIndex < groupMax) {
+      actions.push(writeOutputFile(outputFiles[fileIndex++]));
+    }
+
+    await Promise.all(actions);
+  }
 
   if (assetFiles?.length) {
```
alan-agius4, line 212: Shouldn’t the same thing be done for assets?

alan-agius4 (approved):

PR #26066, opened 2023-10-19
     perf(@angular-devkit/build-angular): cache polyfills virtual module result

alan-agius4:
This commit adds caching to the polyfills virtual modules to avoid module resolutions on rebuilds which can take up to 3 seconds in some cases. Whilst these are done async, these can still slow down the build slightly due to IO limits.


alan-agius4:
Note: this PR is on top of https://github.com/angular/angular-cli/pull/26045

PR #26082, opened 2023-10-20
fix(@angular/cli): remove the need to specify `--migrate-only` when `--name` is used during `ng update`

alan-agius4:
This commit updates the behaviour of `ng update --migrate-only` to remove the need for `--migrate-only` option to be specified. `--migrate-only` will be set internally.

Before
```
ng update @angular/cli --migrate-only --name=migration-name
```

Now
```
ng update @angular/cli --name=migration-name
```

### DIFF
--- a/packages/angular/cli/src/commands/update/cli.ts
+++ b/packages/angular/cli/src/commands/update/cli.ts
@@ -107,23 +107,21 @@ export default class UpdateCommandModule extends CommandModule<UpdateCommandArgs
       })
       .option('name', {
         description:
-          'The name of the migration to run. ' +
-          `Only available with a single package being updated, and only with 'migrate-only' option.`,
+          'The name of the migration to run. Only available when a single package is updated.',
         type: 'string',
-        implies: ['migrate-only'],
         conflicts: ['to', 'from'],
       })
       .option('from', {
         description:
           'Version from which to migrate from. ' +
-          `Only available with a single package being updated, and only with 'migrate-only'.`,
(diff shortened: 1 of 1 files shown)


alan-agius4:
As discussed offline @dgp1130.

packages/angular/cli/src/commands/update/cli.ts
```diff
@@ -107,10 +107,8 @@ export default class UpdateCommandModule extends CommandModule<UpdateCommandArgs
       })
       .option('name', {
         description:
-          'The name of the migration to run. ' +
-          `Only available with a single package being updated, and only with 'migrate-only' option.`,
+          'The name of the migration to run. Only available with a single package being updated.',
```
dgp1130, line 110: Suggestion: "Only available when a single package is updated."

dgp1130 (approved):
Typo in commit message, I think you meant:

> This commit updates the behaviour of `ng update --name=some-migration` to remove the need for `--migrate-only` option to be specified. `--migrate-only` will not be set internally.

Also shouldn't be be "`--migration-only` **will** be set internally"? Or am I misunderstanding the message?

alan-agius4:
> Typo in commit message, I think you meant:
> 
> > This commit updates the behaviour of `ng update --name=some-migration` to remove the need for `--migrate-only` option to be specified. `--migrate-only` will not be set internally.
> 
> Also shouldn't be be "`--migration-only` **will** be set internally"? Or am I misunderstanding the message?

Yes you are right

PR #26084, opened 2023-10-21
perf(@angular-devkit/build-angular): conditionally add Angular compiler plugin to polyfills bundling

clydin:
When using the esbuild-based builders (`application`/`browser-esbuild`), the Angular compiler plugin is now only added to the polyfills bundler configuration if TypeScript files are found in the `polyfills` build option. This is not the case for a default project. The Angular compiler plugin is used to provide type-checking diagnostics for any TypeScript files present within the build. However, if there are no TypeScript files to process, there is no need to use it for the polyfills processing.


alan-agius4 (approved):
Nice!

PR #26107, opened 2023-10-23
fix(@angular-devkit/build-angular): automatically include known packages in vite prebundling

clydin:
When using the Vite-based development server, the application build step already contains the list of known packages that would need to be prebundled. This information can be passed to Vite directly to avoid Vite needing to perform discovery on every output file that will be requested. This also avoids the Vite server behavior where Vite forces a reload of the page when it discovers a new dependency. This behavior can result in lost state during lazy loading of a route.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
+++ b/packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
@@ -115,6 +115,10 @@ export async function* serveWithVite(
   let hadError = false;
   const generatedFiles = new Map<string, OutputFileRecord>();
   const assetFiles = new Map<string, string>();
+  const externalMetadata: { implicit: string[]; explicit: string[] } = {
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
```diff
@@ -336,7 +349,7 @@ export async function setupServer(
   outputFiles: Map<string, OutputFileRecord>,
   assets: Map<string, string>,
   preserveSymlinks: boolean | undefined,
-  prebundleExclude: string[] | undefined,
+  externalMetdata: { implicit: string[]; explicit: string[] },
```
alan-agius4, line 352: Typo

```suggestion
  externalMetadata: { implicit: string[]; explicit: string[] },
```

alan-agius4 (approved):
LGTM, just a typo

Issue #25648, opened 2023-08-07
bug: running "external" schematics in yarn pnp workspace


michaelfaith:
I'm finally getting around to moving a library project that uses `@schematics/angular` to use PnP, and discovered an issue with some code in the `NodeModulesEngineHost` within `@angular-devkit/schematics`.  From my code, within a unit test, we're using the `SchematicTestRunner`'s `runExternalSchematic` function to generate a test project.  Running that function from within a pnp environment is throwing the following error:

```bash
 ENOTDIR: not a directory, lstat '/node_modules/@schematics/angular/collection.json/package.json'

      21005 |
      21006 | function makeError$1(code, message) {
    > 21007 |   return Object.assign(new Error(`${code}: ${message}`), { code });
            |                        ^
      21008 | }
      21009 | function EBUSY(message) {
      21010 |   return makeError$1(`EBUSY`, message);

      at makeError$1 (../../.pnp.cjs:21007:24)
      at ENOTDIR (../../.pnp.cjs:21025:10)
      at ZipFS.resolveFilename (../../.pnp.cjs:22607:15)
      at ZipFS.realpathSync (../../.pnp.cjs:22398:28)
      at ../../.pnp.cjs:23704:100
      at ../../.pnp.cjs:24200:60
      at ZipOpenFS.getZipSync (../../.pnp.cjs:24329:14)
      at ZipOpenFS.makeCallSync (../../.pnp.cjs:24200:17)
      at ZipOpenFS.realpathSync (../../.pnp.cjs:23696:17)
      at VirtualFS.realpathSync (../../.pnp.cjs:23446:26)
      at PosixFS.realpathSync (../../.pnp.cjs:23197:41)
      at NodePathFS.realpathSync (../../.pnp.cjs:23197:41)
      at maybeRealpathSync (../../.yarn/cache/resolve-patch-6d2631dc64-66cc788f13.zip/node_modules/resolve/lib/sync.js:52:16)
      at resolveSync (../../.yarn/cache/resolve-patch-6d2631dc64-66cc788f13.zip/node_modules/resolve/lib/sync.js:97:25)
      at NodeModulesTestEngineHost.resolve (../angular_devkit/schematics/tools/node-module-engine-host.ts:61:39)
      at NodeModulesTestEngineHost.resolve (../angular_devkit/schematics/tools/node-module-engine-host.ts:68:29)
      at createTestCaseSetup (schematics/test-utils.ts:117:40)

```

[...]

PR #26109, opened 2023-10-23
fix(@angular-devkit/schematics): running external schematics with yarn pnp

michaelfaith:
This change addresses an issue encountered when running external schematics from a yarn pnp workspace.  The function used to resolve a collection json using node used recursion in a way that it effectively walked itself into an exception. Then if the exception is the type it expected, it would keep going.  This was flawed in that yarn with pnp throws a different type of error when it failed to load the mis-constructed collection path (e.g. `/node_modules/@schematics/angular/collection.json/package.json`).  `ENOTDIR` instead of `MODULE_NOT_FOUND`.

This process of intentionally / knowingly walking into an exception seems problematic in
general.  So, I addressed it by first checking if the `schematics` entry in the package
is a relative path.  If it is, then don't construct the collection path from that.  
If entry is not relative, then assume it's pointing at another package and we need
to recurse to get to the actual collection path.

I've tested this in both yarn pnp and non-pnp environments.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #25648

- [ ] Yes
- [x] No

Fixes: #25648

### DIFF
--- a/packages/angular_devkit/schematics/tools/node-module-engine-host.ts
+++ b/packages/angular_devkit/schematics/tools/node-module-engine-host.ts
@@ -33,18 +33,7 @@ export class NodeModulesEngineHost extends FileSystemEngineHostBase {
     super();
   }
 
-  private resolve(name: string, requester?: string, references = new Set<string>()): string {
-    if (requester) {
-      if (references.has(requester)) {
-        references.add(requester);
-        throw new Error(
-          'Circular schematic reference detected: ' + JSON.stringify(Array.from(references)),
-        );
-      } else {
-        references.add(requester);
-      }
-    }
-
+  private resolve(name: string, requester?: string): string {
     const relativeBase = requester ? dirname(requester) : process.cwd();
     let collectionPath: string | undefined = undefined;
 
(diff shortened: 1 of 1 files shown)


clydin:
The recursion is used to facilitate the use of package references within the `schematics` field of which the referenced package name could then contain a `schematics` field in its `package.json` that then needs to be resolved. This scenario unfortunately is lacking test coverage currently.
The relative path case appears well covered within this change but the package reference case with `schematics` fields look s like it would fail here. This may lead to breakage of some schematics packages.

michaelfaith:
@clydin thanks for the feedback.  Do you happen to have an example of a package that has their collection declared in that way?  Just so I understand the use case correctly.  Or just as snipped of what that `package.json` entry might look like?

michaelfaith:
I went ahead and added the recursion back in, but after the check for relative path.  Let me know if you feel this addresses the package use-case.

clydin (approved):
LGTM.

Thank you for taking the time to improve this area of the code.

PR #26143, opened 2023-10-26
fix(@angular-devkit/build-angular): handle HTTP requests to assets during SSG in dev-server

alan-agius4:
This commit fixes an issue were during SSG/SSR in the dev-server http requests to assets causes the page rendering to fail.


packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
```diff
@@ -528,9 +528,11 @@ export async function setupServer(
               }
 
               transformIndexHtmlAndAddHeaders(url, rawHtml, res, next, async (html) => {
+                const protocol = serverOptions.ssl ? 'https' : 'http';
```
alan-agius4, line 531: Note: `req.protocol` does not exit.

PR #26164, opened 2023-10-27
Fix empty `styles` string and update component schematic to use single `styles` or `styleUrl` property

dgp1130:
An array is no longer required in v17 and we always generate a single style or style URL anyways, so we may as well drop the array.

This actually revealed a bug in the Webpack compilation where an empty `styles: ''` string would break PostCSS / the Webpack loader. Fixed this by just skipping resource inlining for an empty string, since there's no work to do anyways.

### DIFF
--- a/packages/ngtools/webpack/src/transformers/replace_resources.ts
+++ b/packages/ngtools/webpack/src/transformers/replace_resources.ts
@@ -236,6 +236,11 @@ function transformInlineStyleLiteral(
     return node;
   }
 
+  // Don't transform empty strings as PostCSS will choke on them. No work to do anyways.
+  if (node.text === '') {
+    return node;
+  }
+
   if (!isInlineStyle) {
     const url = getResourceUrl(node);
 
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):

packages/ngtools/webpack/src/transformers/replace_resources.ts
```diff
@@ -236,6 +236,11 @@ function transformInlineStyleLiteral(
     return node;
   }
 
+  // Don't transform empty strings as PostCSS will choke on them. No work to do anyways.
+  if (node.text === '') {
```
alfaproject, line 240: Should the input be trimmed first before the check?

dgp1130: I believe this is only a problem if the file is completely empty, I think a file containing only whitespace should be ok.

alfaproject: Yeah, I guess just an optimization to not even use style processing since there's no styles in that case. Probably not worth it anyway

PR #26186, opened 2023-10-31
fix(@angular-devkit/build-angular): Windows Node.js 20 prerendering failure

alan-agius4:
On Node.js 20 prerendering failed on Windows with `An unhandled exception occurred: No handler function exported` error. This appears to be caused by transforming Piscina CJS bundles using the `JavaScriptTransformer`. interestingly, this does not effect other OS like Linux and Mac.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
+++ b/packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
@@ -71,7 +71,7 @@ export function resolve(
 export async function load(url: string, context: { format?: string | null }, nextLoad: Function) {
   const { format } = context;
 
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
```diff
@@ -69,7 +69,10 @@ export function resolve(
 }
 
 export async function load(url: string, context: { format?: string | null }, nextLoad: Function) {
-  if (isFileProtocol(url)) {
+  const { format } = context;
+
+  // CommonJs modules require not transformations and are not in memory.
```
clydin, line 74:
```suggestion
  // CommonJs modules require no transformations and are not in memory.
```

clydin (approved):

PR #26193, opened 2023-11-01
fix(@angular-devkit/build-angular): fail build on non bundling error when using the esbuild based builders

alan-agius4:
This change fixes an issue were non bundling related errors that happen during prerendering, index generation and bundle budget failures did not cause the build to terminate with an error.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/execute-build.ts
+++ b/packages/angular_devkit/build_angular/src/builders/application/execute-build.ts
@@ -28,7 +28,7 @@ import {
   logMessages,
   transformSupportedBrowsersToTargets,
 } from '../../tools/esbuild/utils';
-import { checkBudgets } from '../../utils/bundle-calculator';
+import { BudgetCalculatorResult, checkBudgets } from '../../utils/bundle-calculator';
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/builders/application/execute-build.ts
```diff
@@ -225,53 +211,75 @@ export async function executeBuild(
...
-      i18nOptions.hasDefinedSourceLocale ? i18nOptions.sourceLocale : undefined,
-    );
-
-    errors = result.errors;
-    warnings = result.warnings;
-    prerenderedRoutes = result.prerenderedRoutes;
-    executionResult.outputFiles.push(...result.additionalOutputFiles);
-    executionResult.assetFiles.push(...result.additionalAssets);
-  }
+  if (executionResult.errors.length === 0) {
```
clydin, line 214: Should be able to drop this conditional. Bundle budget errors won't cause anything else to break and might as well get a comprehensive list of errors when possible.
If there are bundling errors which are typically fatal to the rest of the steps, the build already exits early.

clydin (approved):

PR #26197, opened 2023-11-01
fix(ngtools/webpack): let static scan to find image domains in standlone components

atcastle:
Currently, the static scanning done by the image domain scanner catches domains provided in files following the AppModule pattern, but misses domains when using the standalone component module. This PR adjusts the scan path to fix that. The actual logic governing the detection of domains and inserting them into the index.html file is unchanged, this fix simply tells the scanner to also look in one additional place in files.

Specifically, the new scanning route looks for the `static ɵinj = /*@__PURE__*/ i0.ɵɵdefineInjector(...)` invocation, which exists in post-transform standalone component files, and contains the providers.

As with the existing scanning, this new code path is very specific, looking at a constant number of nodes, and should not effect build time. CC: @clydin @AndrewKushnir @kara

### DIFF
--- a/packages/ngtools/webpack/src/transformers/find_image_domains.ts
+++ b/packages/ngtools/webpack/src/transformers/find_image_domains.ts
@@ -84,7 +84,7 @@ export function findImageDomains(imageDomains: Set<string>): ts.TransformerFacto
         return node;
       }
 
-      function findPropertyAssignment(node: ts.Node) {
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):
LGTM

alan-agius4:
@atcastle, this does not apply cleanly on 16.2

PR #26205, opened 2023-11-02
refactor(@angular/ssr): use `join` instead of `resolve`

alan-agius4:
With this change we replace `resolve` with `join` to avoid potential path traversal vulnerability.

More context about the reasoning behind this change can be found in https://buganizer.corp.google.com/issues/299878755#comment26

//cc @securityMB, @AndrewKushnir

### DIFF
--- a/packages/angular/ssr/src/common-engine.ts
+++ b/packages/angular/ssr/src/common-engine.ts
@@ -9,7 +9,7 @@
 import { ApplicationRef, StaticProvider, Type } from '@angular/core';
 import { renderApplication, renderModule, ɵSERVER_CONTEXT } from '@angular/platform-server';
 import * as fs from 'node:fs';
-import { dirname, resolve } from 'node:path';
+import { dirname, join, normalize, resolve } from 'node:path';
 import { URL } from 'node:url';
 import { InlineCriticalCssProcessor, InlineCriticalCssResult } from './inline-css-processor';
 import {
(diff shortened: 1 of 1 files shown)


clydin (approved):

packages/angular/ssr/src/common-engine.ts
```diff
@@ -117,32 +117,34 @@ export class CommonEngine {
...
+    // Do not use `resolve` here as otherwise it can lead to path traversal vulnerability.
+    // See: https://portswigger.net/web-security/file-path-traversal
+    const pagePath = join(publicPath, pathname, 'index.html');
+
+    if (this.pageIsSSG.get(pagePath)) {
+      // Serve pre-rendered page.
+      return fs.promises.readFile(pagePath, 'utf-8');
+    }
+
+    if (!pagePath.startsWith(normalize(publicPath))) {
```
securityMB, line 130: `pagePath` should also be normalized so that we compare two normalized paths (otherwise, there might be false negatives if symbolic links are involved).

securityMB (commented):
I left one comment with a nit. Otherwise, LGTM.

packages/angular/ssr/src/common-engine.ts
alan-agius4: `pagePath` is already normalized as it is constructed using `join` which always return a normalized path

securityMB: You're right, sorry, I missed that part of `join`. So LGTM 😀

PR #26211, opened 2023-11-02
refactor(@angular-devkit/build-angular): cache compiled load ESM file helper

clydin:
The dynamically compiled ESM import helper is now cached to prevent the need to recompile the helper function everytime a load ESM helper call is made. This helper is currently used to workaround dynamic import limitations with the TypeScript compilation output. Once the build process is updated, it will no longer be required.

### DIFF
--- a/packages/angular/cli/src/utilities/load-esm.ts
+++ b/packages/angular/cli/src/utilities/load-esm.ts
@@ -7,6 +7,11 @@
  */
 
 /**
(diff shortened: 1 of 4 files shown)


alan-agius4 (commented):
There are a couple of places where similar function is defined. 

Should this be done throughout?

clydin:
yes. i'll take a look at the others.

alan-agius4 (approved):

PR #26227, opened 2023-11-03
fix(@angular-devkit/build-angular): avoid in-memory prerendering ESM loader errors

clydin:
The in-memory ESM loader hooks have been adjusted to avoid potential errors when
resolving and loading the bundled server code during prerendering. These errors
could result in hard to diagnose build failures. E2E testing via the deprecated
protractor builder, would silently exit, for instance. To ensure on disk files
including node modules are resolved from the workspace root, a virtual file root
is used for all in memory files. This path does not actually exist but is used to
overlay the in memory files with the actual filesystem for resolution purposes.
A custom URL schema (such as `memory://`) cannot be used for the resolve output
because the in-memory files may use `import.meta.url` in ways that assume a file
URL. `createRequire` from the Node.js `module` builtin is one example of this usage.

Passing Windows CI run: https://github.com/angular/angular-cli/actions/runs/6750057482/job/18352859851?pr=26227

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
+++ b/packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
@@ -6,7 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { join, relative } from 'node:path';
+import assert from 'node:assert';
+import { randomUUID } from 'node:crypto';
+import { join } from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { fileURLToPath } from 'url';
 import { JavaScriptTransformer } from '../../../tools/esbuild/javascript-transformer';
@@ -17,14 +19,14 @@ import { callInitializeIfNeeded } from './node-18-utils';
  * @see: https://nodejs.org/api/esm.html#loaders for more information about loaders.
  */
 
+const MEMORY_URL_SCHEME = 'memory://';
+
 export interface ESMInMemoryFileLoaderWorkerData {
   outputFiles: Record<string, string>;
   workspaceRoot: string;
 }
 
-const TRANSFORMED_FILES: Record<string, string> = {};
-const CHUNKS_REGEXP = /file:\/\/\/((?:main|render-utils)\.server|chunk-\w+)\.mjs/;
-let workspaceRootFile: string;
+let memoryVirtualRootUrl: string;
 let outputFiles: Record<string, string>;
 
 const javascriptTransformer = new JavaScriptTransformer(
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
```diff
@@ -47,49 +56,93 @@ export function resolve(
...
-    if (source === undefined) {
-      source = TRANSFORMED_FILES[filePath] = Buffer.from(
-        await javascriptTransformer.transformFile(filePath),
-      ).toString('utf-8');
-    }
+  // Only module files potentially require transformation. Angular libraries that would
+  // need linking are ESM only.
+  if (format === 'module' && isFileProtocol(url)) {
+    const filePath = fileURLToPath(url);
+    const source = await javascriptTransformer.transformFile(filePath);
```
alan-agius4, line 139: OOC, should we keep caching this?

packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
```diff
@@ -38,7 +40,14 @@ const javascriptTransformer = new JavaScriptTransformer(
...
 
 export function initialize(data: ESMInMemoryFileLoaderWorkerData) {
-  workspaceRootFile = pathToFileURL(join(data.workspaceRoot, 'index.mjs')).href;
+  // This path does not actually exist but is used to overlay the in memory files with the
+  // actual filesystem for resolution purposes.
+  // A custom URL schema (such as `memory://`) cannot be used for the resolve output because
+  // the in-memory files may use `import.meta.url` in ways that assume a file URL.
+  // `createRequire` is one example of this usage.
+  memoryVirtualRootUrl = pathToFileURL(
+    join(data.workspaceRoot, `.angular/prerender-root/${randomUUID()}/`),
```
alan-agius4, line 49: Why do we need `randomUUID`?


```suggestion
    join(data.workspaceRoot, '.angular/prerender-root', randomUUID()),
```

alan-agius4 (approved):
LGTM, couple of NITs

packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
clydin: Each ES module is only ever loaded once so the cache won’t get hit.

packages/angular_devkit/build_angular/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
clydin: Guarantees a unique URL for the internal ES module map.

PR #26231, opened 2023-11-04
perf(@angular-devkit/build-angular): start SSR dependencies optimization before the first request

alan-agius4:
Vite will only start dependency optimization of SSR modules when the first request comes in.

In some cases, this causes a long waiting time. To mitigate this, we call `ssrLoadModule` to initiate this process before the first request.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
+++ b/packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
@@ -216,8 +216,20 @@ export async function* serveWithVite(
       );
 
       server = await createServer(serverConfiguration);
-
       await server.listen();
+
+      if (browserOptions.ssr) {
+        /**
+         * Vite will only start dependency optimization of SSR modules when the first request comes in.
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
```diff
@@ -216,8 +216,15 @@ export async function* serveWithVite(
...
 
       server = await createServer(serverConfiguration);
-
       await server.listen();
+
+      if (browserOptions.ssr) {
+        // Vite will only start dependency optimization of SSR modules when the first request comes in.
+        // In some cases, this causes a long waiting time. To mitigate this, we call `ssrLoadModule` to
+        // initiate this process before the first request.
+        void server.ssrLoadModule('<deps-caller>').catch(() => {});
```
clydin, line 225: Can you add a note here that this will intentionally fail from the unknown module argument but there is no other way currently to initiate the ssr dep optimizer?

clydin (approved):

PR #26265, opened 2023-11-07
fix(@angular-devkit/build-angular): disable Worker wait loop for TS/NG parallel compilation in web containers

clydin:
When using the `application` builder, a parallel TS/NG compilation is used that is run inside a Node.js Worker.
This Worker by default uses an Atomics-based wait loop to improve performance while waiting for messages. This
loop relies on the synchronous API `receiveMessageOnPort`. While this works well in Node.js, the web container
execution environment does not currently support passing transferable objects via `receiveMessageOnPort`. Attempting
to do so will cause a serialization error and a failed application build. To avoid this problem, the wait loop
optimization is disabled when the web container execution environment is detected. This change is only needed
for the TS/NG compilation as no other parallel operation within the build system currently uses `receiveMessageOnPort`
with transferable objects.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/angular/compilation/parallel-compilation.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/angular/compilation/parallel-compilation.ts
@@ -36,6 +36,9 @@ export class ParallelCompilation extends AngularCompilation {
       minThreads: 1,
       maxThreads: 1,
       idleTimeout: Infinity,
+      // Web containers do not support transferable objects with receiveOnMessagePort which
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):

d3lm:
Just to add some extra color here: I have tested the build with parallel compilation enabled and atomic loop disabled and it seems to work well in Chrome for both initial build and re-compilation. In Firefox the initial build also seems to work now, but it still gets stuck on re-compilation, but not always. Sometimes it worked. I am still trying to understand what's going on here. But at least the initial build now seems to work in Firefox as well.

PR #26314, opened 2023-11-10
fix(@angular-devkit/build-angular): do not process server entry-point when running `ng serve`

alan-agius4:
The server entry-point is not used by vite.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
+++ b/packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
@@ -68,11 +68,13 @@ export async function* serveWithVite(
     builderName,
   )) as json.JsonObject & BrowserBuilderOptions;
 
-  if (browserOptions.prerender) {
+  if (browserOptions.prerender || browserOptions.ssr) {
     // Disable prerendering if enabled and force SSR.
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
```diff
@@ -68,12 +68,14 @@ export async function* serveWithVite(
...
 
-  if (browserOptions.prerender) {
+  if (browserOptions.prerender || browserOptions.ssr) {
     // Disable prerendering if enabled and force SSR.
     // This is so instead of prerendering all the routes for every change, the page is "prerendered" when it is requested.
-    browserOptions.ssr = true;
     browserOptions.prerender = false;
 
+    // Avoid bundling and processing the ssr entry-point as this is not used by the dev-server.
+    browserOptions.ssr = true;
```
clydin, line 77: Should this be `false`?

alan-agius4: `ssr` should be true when either `prerender` or `ssr` are truthy to enable SSR for the dev-server. If this is set to false, both SSR and prerendering will be disabled which will cause no request based SSR to happen in the dev-server.

clydin (approved):

PR #26436, opened 2023-11-20
fix(@angular-devkit/build-angular): ensure service worker hashes index HTML file for application builder

clydin:
The generated HTML files for the application were previously not included in the hash calculations for the service worker. This has now been corrected and the index file for an application will not be included when service worker support is enabled via the `serviceWorker` option.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/execute-post-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/builders/application/execute-post-bundle.ts
@@ -144,7 +144,8 @@ export async function executePostBundleSteps(
         workspaceRoot,
         serviceWorker,
         options.baseHref || '/',
-        outputFiles,
+        // Ensure additional files recently added are used
+        [...outputFiles, ...additionalOutputFiles],
         assetFiles,
       );
       additionalOutputFiles.push(


alan-agius4 (approved):

packages/angular_devkit/build_angular/src/builders/application/execute-post-bundle.ts
```diff
@@ -144,7 +144,10 @@ export async function executePostBundleSteps(
         workspaceRoot,
         serviceWorker,
         options.baseHref || '/',
-        outputFiles,
+        // Ensure additional files recently added are used and also only browser files
+        [...outputFiles, ...additionalOutputFiles].filter(
```
alan-agius4, line 148: Actually this will cause media files not to be included.

https://github.com/angular/angular-cli/blob/b9b0fb9580f39d4731eb845da279a0fd16f40804/packages/angular_devkit/build_angular/src/utils/service-worker.ts#L69-L86

You can potentially just remove the `filter` from here since it's already handled at a latter stage.

alan-agius4 (changes requested):
See comment

alan-agius4 (approved):

Issue #26432, opened 2023-11-20
Special comments /*! are removed from the styles.css


CarlosTorrecillas:
### Which @angular/* package(s) are the source of the bug?

compiler-cli

### Is this a regression?

No

### Description

I'm not sure if that's a regression, but having a brand new Angular 17 application created from the `ng new` command is not keeping the comments written in the styles.scss even if they are done in the "special" way of `/*!`

If you inspect the styles.scss you will see two sections of special comments that should be respected after build. When you perform a prod build you will see they are not.

### Please provide a link to a minimal reproduction of the bug

https://github.com/CarlosTorrecillas/angular17

### Please provide the exception or error you saw

```true
There are no exceptions. If you run "ng build" and go to the dist folder and inspect the styles.XXX.css file generated you will see the CSS that does not contain the special comment that should have.
```

### Please provide the environment you discovered this bug in (run `ng version`)

```true
_                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/


Angular CLI: 17.0.1
Node: 18.13.0
Package Manager: npm 8.19.3
OS: win32 x64

Angular: 17.0.3
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, platform-server
... router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1700.1
@angular-devkit/build-angular   17.0.1
@angular-devkit/core            17.0.1
@angular-devkit/schematics      17.0.1
@angular/cli                    17.0.1
@angular/ssr                    17.0.1
@schematics/angular             17.0.1
rxjs                            7.8.1
typescript                      5.2.2
zone.js                         0.14.2
```

### Anything else?

That has happened to me with a single file or multiple (css) files included in the angular.json project. I can also see that on recently migrated Angular solutions from v16 to v17 and on brand new A17 applications. I would need to test it on A16 applications to verify if that's a regression or not but what I can see, if I don't use the target browser application to compile is that when I have multiple files, I do see some comments respected, but only in one file. The rest of the files are removed.

dgp1130:
[...]

PR #26444, opened 2023-11-21
feat(@angular-devkit/build-angular): add option to retain CSS special comments in global styles

alan-agius4:
Prior to this change special CSS comments `/*! comment */` were being removed during minification when using the application builder. This caused tools that ran post build that rely on such comments such as purgeCSS and critters not to function properly.

We now provide a `removeSpecialComments` option to enable retention of these comments in global CSS files.

Usage example:
```json
{
  "projects": {
    "my-app": {
      "architect": {
        "build": {
          "configurations": {
            "production": {
              "optimization": {
                "styles": {
                  "removeSpecialComments": false
                }
              }
            }
          }
        }
      }
    }
  }
}
```

Closes: #26432


CarlosTorrecillas:
Thanks @alan-agius4, as always, providing an outstanding turnaround !

CarlosTorrecillas:
Hi guys, just for our information, is this planned to be merged at some point soon? No rush at all, it is because this is holding us back from release the upgrade to Angular 17. Thanks!

alan-agius4:
Had a chat with @clydin earlier last week and we decided to go with an additional option. The PR has been updated to add this option.

This should be available as part of 17.1.0.

clydin (approved):

packages/angular_devkit/build_angular/src/builders/application/tests/options/optimization-remove-special-comments_spec.ts
```diff
@@ -0,0 +1,79 @@
...
+        },
+      });
+
+      const { result } = await harness.executeOnce();
+      expect(result?.success).toBeTrue();
+
+      harness.expectFile('dist/browser/styles.css').content.not.toContain('important-comment');
+    });
+
+    it(`should not retain special comments when 'removeSpecialComments' is set not set`, async () => {
```
clydin, line 63:
```suggestion
    it(`should not retain special comments when 'removeSpecialComments' is not set`, async () => {
```

PR #26453, opened 2023-11-22
fix(@angular-devkit/build-angular): avoid native realpath in application builder

clydin:
When the `preserveSymlinks` option is false (the default), the tsconfig path was passed through realpath to ensure that the TypeScript source files were resolved correctly. However, the promise version of the Node.js API was previously used. This version used `realpath.native` internally but the native version has numerous behavior problems on Windows systems. This includes potential case conversion which can result in differing cases for files within the build system and in turn failed path comparison checks. The synchronous version is now used which has a JavaScript implementation that does not suffer from these problems.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/angular/compiler-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/angular/compiler-plugin.ts
@@ -16,7 +16,7 @@ import type {
   PluginBuild,
 } from 'esbuild';
 import assert from 'node:assert';
-import { realpath } from 'node:fs/promises';
+import { realpathSync } from 'node:fs';
 import * as path from 'node:path';
 import { maxWorkers } from '../../../utils/environment-options';
 import { JavaScriptTransformer } from '../javascript-transformer';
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -61,8 +61,11 @@ export function createCompilerPlugin(
       if (!preserveSymlinks) {
         // Use the real path of the tsconfig if not preserving symlinks.
         // This ensures the TS source file paths are based on the real path of the configuration.
+        // NOTE: promises.realpath should not be used here since it uses realpath.native which
+        // can cause case conversion and other undesirable behavior on Windows systems.
+        // ref: https://github.com/nodejs/node/issues/7726
         try {
-          tsconfigPath = await realpath(tsconfigPath);
+          tsconfigPath = realpathSync(tsconfigPath);
```
alan-agius4, line 68: NIT: This should no longer be needed now, since tsconfig path is already resolved using the workspace realpath, which is part of the options normalisation.

clydin: Left it there just to be safe.
I'll also be doing a round of path normalization cleanup where we can address this as well.

alan-agius4: Tested again, and this is indeed no longer required after https://github.com/angular/angular-cli/pull/26453/commits/23b89fb1de5edda16055c5fc65d2d9256ca0d27d

alan-agius4 (approved):

PR #26458, opened 2023-11-22
fix(@angular-devkit/build-angular): treeshake unused class that use custom decorators

alan-agius4:
This changes enables wrapping classes in side-effect free modules that make use of custom decorators when using the esbuild based builders so that when such classes are unused they can be treeshaken.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/angular/compiler-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/angular/compiler-plugin.ts
@@ -358,10 +358,12 @@ export function createCompilerPlugin(
           };
         } else if (typeof contents === 'string') {
           // A string indicates untransformed output from the TS/NG compiler
+          const sideEffects = await hasSideEffects(request);
           contents = await javascriptTransformer.transformData(
             request,
             contents,
             true /* skipLinker */,
+            sideEffects,
           );
 
           // Store as the returned Uint8Array to allow caching the fully transformed code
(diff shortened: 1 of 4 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -358,10 +358,12 @@ export function createCompilerPlugin(
           };
         } else if (typeof contents === 'string') {
           // A string indicates untransformed output from the TS/NG compiler
+          const sideEffects = await hasSideEffects(args.path);
```
clydin, line 361:
```suggestion
          const sideEffects = await hasSideEffects(request);
```
The file path may have been changed by a `fileReplacements` entry.

packages/angular_devkit/build_angular/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -430,6 +434,22 @@ export function createCompilerPlugin(
...
+       * Checks if the file has side-effects when `advancedOptimizations` is enabled.
+       */
+      async function hasSideEffects(path: string): Promise<boolean | undefined> {
+        if (!pluginOptions.advancedOptimizations) {
+          return undefined;
+        }
+
+        const { sideEffects } = await build.resolve(path, {
+          kind: 'import-statement',
+          resolveDir: build.initialOptions.absWorkingDir ?? '',
```
clydin, line 448: nit: The path has been resolved to a full path at this point so it shouldn't need the `resolveDir` option.

clydin (approved):

PR #26545, opened 2023-12-01
fix(@angular-devkit/build-angular): correctly set `ngDevMode` and `ngJitMode` when using vite

alan-agius4:
Prior to this change `ngDevMode` and `ngJitMode` were not correctly being set when using the vite base dev-server.


packages/angular_devkit/build_angular/src/builders/dev-server/vite-server.ts
```diff
@@ -12,28 +12,40 @@ import type { json, logging } from '@angular-devkit/core';
 import type { Plugin } from 'esbuild';
```
alan-agius4, line 12: We could also go with another approach, where we inject the `define` as globals in the polyfills file, through the banner option, however there might be cases were there are no polyfils.

Issue #21279, opened 2021-07-03
Enable more linter checks in tsconfig.json by default 


prabh-62:
# 🚀 Feature request

### Command (mark with an `x`)

- [x] new
- [ ] build
- [ ] serve
- [ ] test
- [ ] e2e
- [x] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Description

```shell
❯ npx @angular/cli@12 new web-apps --create-application false
❯ cd web-apps && yarn ng g app admin-console --routing --style scss
```
Currently, base `tsconfig.json` only contains `strict`, `noImplicitReturns` and `noFallthroughCasesInSwitch` flags
```jsonc
{
    "compileOnSave": false,
    "compilerOptions": {
      ...
      // all strict checks
      "strict": true,
      //  Linter checks
      "noImplicitReturns": true,
      "noFallthroughCasesInSwitch": true,
      ...
    }
}
```

### Describe the solution you'd like
Enable new linter flags exposed by typescript in recent versions by defalt
```jsonc
{
  "compileOnSave": false,
  "compilerOptions": {
    ...
    // all strict checks
    "strict": true,
    //  Linter checks
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
	// TS 4.1
    "noUncheckedIndexedAccess": true,
	// TS 4.2
    "noPropertyAccessFromIndexSignature": true,
	// TS 4.3
    "noImplicitOverride": true,
	// TS 4.4
    "useUnknownInCatchVariables": true,
    "exactOptionalPropertyTypes": true,
    ...
  }
}
```
<img width="910" alt="angular_strict_tsconfig_ts_4_4" src="https://user-images.githubusercontent.com/13006107/124361786-da905400-dbfe-11eb-9be2-f0e6dafc4f57.png">

Relevant blog posts:
- [Announcing TypeScript 4.1](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/)
	- `--noUncheckedIndexedAccess`	
- [Announcing TypeScript 4.2](https://devblogs.microsoft.com/typescript/announcing-typescript-4-2/)
	- `--noPropertyAccessFromIndexSignature`
- [Announcing TypeScript 4.3](https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/)
	- `--noImplicitOverride`
[...]

PR #26547, opened 2023-12-01
feat(@schematics/angular): enable TypeScript `exactOptionalPropertyTypes` in new strict workspaces

alan-agius4:
TypeScripts recommanded enabling `exactOptionalPropertyTypes`. From https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes

With `exactOptionalPropertyTypes` enabled, TypeScript applies stricter rules around how it handles properties on type or interfaces which have a `?` prefix.

For example, this interface declares that there is a property which can be one of two strings: ‘dark’ or ‘light’ or it should not be in the object.
```ts
interface UserDefaults {
  // The absence of a value represents 'system'
  colorThemeOverride?: "dark" | "light";
}
```

Without this flag enabled, there are three values which you can set `colorThemeOverride` to be: “dark”, “light” and undefined.

Setting the value to undefined will allow most JavaScript runtime checks for the existence to fail, which is effectively falsy. However, this isn’t quite accurate; `colorThemeOverride: undefined` is not the same as `colorThemeOverride` not being defined. For example, "colorThemeOverride" in settings would have different behavior with undefined as the key compared to not being defined.

`exactOptionalPropertyTypes` makes TypeScript truly enforce the definition provided as an optional property.

```ts
const settings = getUserSettings();
settings.colorThemeOverride = "dark";
settings.colorThemeOverride = "light";

// But not:
settings.colorThemeOverride = undefined;
// Type 'undefined' is not assignable to type '"dark" | "light"' with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the type of the target.
```

### DIFF
--- a/packages/schematics/angular/workspace/files/tsconfig.json.template
+++ b/packages/schematics/angular/workspace/files/tsconfig.json.template
@@ -3,6 +3,7 @@
   "compileOnSave": false,
   "compilerOptions": {
     "outDir": "./dist/out-tsc",<% if (strict) { %>
+    "exactOptionalPropertyTypes": true,
     "forceConsistentCasingInFileNames": true,
     "strict": true,
     "noImplicitOverride": true,


dgp1130:
Worth pointing out that we discussed this option previously and decided against adding it: https://github.com/angular/angular-cli/issues/21279#issuecomment-876665399

Most of the same arguments still hold, I don't think the functionality has changed since then. At that time I believe this option wasn't explicitly recommended. Even today it isn't used by default in `tsc --init`. Internally at Google [the flag was evaluated](http://g/typescript-users/HeyVLbdgn-M/LoHr8BATAQAJ) and rejected due to a significant number of breakages with most not reflecting real coding errors. For new Angular projects, risk of breakage isn't too much of a concern, but I do think the fact the signal to noise ratio of the flag didn't justify its usage internally is relevant to this decision.

I do agree this flag makes things slightly more correct, and `settings.colorThemeOverride = undefined` is ever so slightly different from `settings = {}` due to the presence of the `colorThemeOverride` property (`'colorThemeOverride' in settings` returns different values). Just out of curiosity, I checked and `delete settings.colorThemeOverride` is allowed, which seems like the correct replacement for that statement which maintains expected presence semantics.

I'm still a little worried about educating developers unfamiliar with this specific option. The difference between `?` and `| undefined` is quite nuanced, non-obvious, and usually doesn't matter. If `tsc --init` enabled this flag, I'd be more inclined to align Angular to match. At the moment I'm probably leaning away from enabling it unless someone feels strongly in favor of it?

alan-agius4:
I totally forgot about that discussion. Agreed. Let’s reconsider in the future.

PR #26556, opened 2023-12-01
fix(@angular-devkit/build-angular): normalize asset source locations in Vite-based development server

clydin:
The Vite-based development server uses an allow list to permit access to configured assets. This list is checked internally to Vite by uses its normalized path form. To ensure that assets provided by the build are checked correctly on all platforms, the asset list is now normalized with Vite's path normalization prior to being used.


tests/legacy-cli/e2e/tests/commands/serve/assets.ts
```diff
@@ -0,0 +1,23 @@
+import assert from 'assert';
```
alan-agius4, line 1: I am not totally following the test, it's named `assets` but it's testing only for `favicon.ico`. Which is covered by other tests. 

https://github.com/angular/angular-cli/blob/e2f92ab957e797d8085616fcdea0b73344e614af/tests/legacy-cli/e2e/tests/basic/serve.ts#L22

It seems that the tests itself, is mostly testing that the index is always server on not found pages?

PR #26588, opened 2023-12-05
refactor: update rxjs imports

alan-agius4:
Remove `rxjs/operators` legacy import


packages/angular_devkit/schematics/tools/schema-option-transform.ts
```diff
@@ -7,8 +7,7 @@
  */
 
 import { deepCopy, schema } from '@angular-devkit/core';
-import { Observable, from, of as observableOf } from 'rxjs';
-import { first, map, mergeMap } from 'rxjs/operators';
+import { Observable, first, from, map, mergeMap, of as observableOf } from 'rxjs';
```
clydin, line 10: I think you'll need to keep this in place to support syncing.

clydin (approved):

PR #26613, opened 2023-12-07
fix(@angular-devkit/build-angular): resolve correct outdir for esbuild

edbzn:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Esbuild options `outdir` is currently using the `workspaceRoot` value instead of the `outputPath`, it is problematic for instance when we want to create a plugin where we need the actual outdir.  

Issue Number: N/A

## What is the new behavior?

Now esbuild build options are using the real output path of the application instead of the workspace root. 

- [ ] Yes
- [x] No

## Other information

Before writing tests, I wanted to see if this change could be accepted.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/application-code-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/application-code-bundle.ts
@@ -10,7 +10,7 @@ import type { BuildOptions } from 'esbuild';
 import assert from 'node:assert';
 import { createHash } from 'node:crypto';
 import { readFile } from 'node:fs/promises';
-import { extname, join } from 'node:path';
+import { extname, join, resolve } from 'node:path';
 import type { NormalizedApplicationBuildOptions } from '../../builders/application/options';
 import { allowMangle } from '../../utils/environment-options';
 import { createCompilerPlugin } from './angular/compiler-plugin';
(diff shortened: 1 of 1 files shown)


edbzn:
@alan-agius4 I'm curious about this, could you elaborate on why it is by design? For the context, I'm using Nx and we can define custom esbuild plugins using the browser-esbuild executor that delegates the work to the Angular builder. These plugins can access options passed to esbuild, and we expect to have access to the output path somewhere, but it is not the case we don't have the information as the workspace root value is used instead.

alan-agius4:
During an Angular builder invocation, several esbuild compilations occur. These result in output directed to both /browser and /server, all outside the esbuild context since esbuild itself doesn't write files to disk. The workspace root is utilized, ensuring esbuild outputs files without a directory.

Example"

- "main.js"
- "media/img.png"
- 
OOC, why do you need the output path in a plugin?

alan-agius4:
I am not sure how this plugin would work, as esbuild doesn't easily allow adding assets which are not referenced in code. Writing the files directly using Node.js FS methods would result in the "asset" not being present during development when using Vite.

edbzn:
Good point, I missed the dev server issue, thanks for the hint.

PR #26621, opened 2023-12-08
feat(@angular-devkit/build-angular): inline Google and Adobe fonts located in stylesheets

### LINKED ISSUE #23054: Support font-inlining for global styles
# 🚀 Feature request

- [x] build

### Description

Currently, font inlining only works for font that are declared in the `index.html`.
 
```html
<link href="https://fonts.googleapis.com/css2?family=Open+Sans:wght@300&display=swap" rel="stylesheet">
```

However, fonts can also be imported in CSS files using the `@import` rule. 

```css
@import url("https://fonts.googleapis.com/css2?family=Open+Sans:wght@300&display=swap");
``` 

We should probably handle both cases.

alan-agius4:
`@import url()` to Google and Adobe fonts that are located in global and component CSS will now be inlined when using the esbuild based builders.

Input
```css
@import url(https://fonts.googleapis.com/css?family=Roboto:300,400,500);
```

Output
```css
/* latin */
@font-face {
  font-family: 'Roboto';
  font-style: normal;
  font-weight: 500;
  src: url(https://fonts.gstatic.com/s/roboto/v30/KFOlCnqEu92Fr1MmEU9fBBc4AMP6lQ.woff2) format('woff2');
  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+0304, U+0308, U+0329, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
}
```

Closes #23054

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/css-inline-fonts-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/css-inline-fonts-plugin.ts
@@ -1,0 +1,77 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import type { Plugin, PluginBuild } from 'esbuild';
+import { InlineFontsProcessor } from '../../../utils/index-file/inline-fonts';
+import { NormalizedCachedOptions } from '../../../utils/normalize-cache';
+import { LoadResultCache, createCachedLoad } from '../load-result-cache';
+
+/**
+ * Options for the createCssInlineFontsPlugin
+ * @see createCssInlineFontsPlugin
+ */
+export interface CssInlineFontsPluginOptions {
+  /** Disk cache normalized options */
+  cacheOptions?: NormalizedCachedOptions;
+  /** Load results cache. */
+  cache?: LoadResultCache;
+}
+
+/**
+ * Creates an esbuild {@link Plugin} that inlines fonts imported via import-rule.
+ * within the build configuration.
+ */
+export function createCssInlineFontsPlugin({
+  cache,
+  cacheOptions,
+}: CssInlineFontsPluginOptions): Plugin {
+  return {
+    name: 'angular-css-inline-fonts-plugin',
+    setup(build: PluginBuild): void {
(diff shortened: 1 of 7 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/css-inline-fonts-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+
+      build.onResolve({ filter: /fonts\.googleapis\.com|use\.typekit\.net/ }, (args) => {
+        // Only attempt to resolve import-rule tokens which only exist inside CSS.
+        if (args.kind !== 'import-rule') {
+          return null;
+        }
+
+        if (!inlineFontsProcessor.canInlineRequest(args.path)) {
+          return null;
+        }
```
clydin, line 46: This could throw if the path is not actually a URL or an invalid URL.  Can you add a catch here and probably return null?

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/css-inline-fonts-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+        return {
+          path: args.path,
+          namespace: 'css-inline-fonts',
+        };
+      });
+
+      build.onLoad(
+        { filter: /./, namespace: 'css-inline-fonts' },
+        createCachedLoad(cache, async (args) => ({
+          contents: await inlineFontsProcessor.processURL(args.path),
```
clydin, line 57: Can this throw if there's a network problem?
Not sure what the behavior should be in that case. I suppose it would need to error at this point.

clydin (approved):

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/css-inline-fonts-plugin.ts
alan-agius4: Updated so that it never throws.

packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/css-inline-fonts-plugin.ts
alan-agius4: updated.

PR #26623, opened 2023-12-08
fix(@angular-devkit/build-angular): inlining of fonts results in jagged fonts for Windows users

### LINKED ISSUE #22248: Automatic inlining of fonts results in jagged fonts for Windows users
# 🐞 Bug report

- [ ] new
- [x] build
- [ ] serve
- [ ] test
- [ ] e2e
- [ ] generate
- [ ] add
- [ ] update
- [ ] lint
- [ ] extract-i18n
- [ ] run
- [ ] config
- [ ] help
- [ ] version
- [ ] doc

### Is this a regression?

Yes, automatic inlining of fonts was introduced in Angular 11 (defaulted to on)

### Description

Some Google fonts are rendering poorly for Windows users when automatic font inlining is turned on

## 🔬 Minimal Reproduction

Start up a new angular project, and add google open sans font into your `index.html` like so:

```
  <link rel="preconnect" href="https://fonts.googleapis.com">
  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
  <link href="https://fonts.googleapis.com/css2?family=Open+Sans&display=swap" rel="stylesheet">
```
Set the font-family of text to 'Open Sans' and font size to 12px.
Build the project so that the font inlining happens, and serve the static bundle locally. 
One quick example to do this is to run `npx http-server dist/{PROJECT_NAME_HERE}`
Notice that the rendered fonts look jagged and poor aliased.

image

If I turn off font-inlining and rebuild the font renders better:

image

## 🌍 Your Environment

<pre><code>
Angular CLI: 12.2.13
Node: 15.3.0 (Unsupported)
Package Manager: npm 7.0.14
OS: win32 x64

Angular: 12.2.13

@angular-devkit/architect       0.1202.13
@angular-devkit/build-angular   12.2.13
@angular-devkit/core            12.2.13
@angular-devkit/schematics      12.2.13
@schematics/angular             12.2.13
</code></pre>

**Anything else relevant?**

[Looking at how the build-time font inlining works](https://github.com/angular/angular-cli/blob/7ff8c5350ea2e49574dd659adae02215957d2685/packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts#L185), it's making a request at build time with a Mac OSX user-agent which returns a different set of fonts than if I hit the Google Fonts endpoint with a Windows user-agent. In my testing when I replaced the inlined styles with fonts inlined with a Windows user agent the fonts rendered fine.

alan-agius4:
We now replace the user agent string used to perform font inlining with a Windows one. This is because Google fonts will including hinting in fonts for Windows. Hinting is a technique used with Windows files to improve appearance however results in 20-50% larger file sizes, however this will make the fonts display correctly on all platforms.

More information about this can be found in:
- https://fonts.google.com/knowledge/glossary/hinting
- https://glyphsapp.com/learn/hinting-manual-truetype-hinting
- http://google3/java/com/google/fonts/css/OpenSansWebFontsCssBuilder.java?l=22

Note: this has been discussed with Katie Hempenius from the Chrome Aurora team.

Closes #22248

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/inline-fonts.ts
@@ -211,8 +211,17 @@ export class InlineFontsProcessor {
           {
             agent,
             headers: {
+              /**
+               * Always use a Windows UA. This is because Google fonts will including hinting in fonts for Windows.
+               * Hinting is a technique used with Windows files to improve appearance however
+               * results in 20-50% larger file sizes.
+               *
+               * @see http://google3/java/com/google/fonts/css/OpenSansWebFontsCssBuilder.java?l=22
+               * @see https://fonts.google.com/knowledge/glossary/hinting (short)
+               * @see https://glyphsapp.com/learn/hinting-manual-truetype-hinting (deep dive)
+               */
               'user-agent':
-                'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.121 Safari/537.36',
+                'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36',
             },
           },
           (res) => {


dgp1130 (approved):
Definitely a good fix.

I am a little worried that Google Fonts may change over time. It's possible that Windows/Mac/Linux fonts completely diverge with no "least common denominator" shared between them. If so, asking for Windows fonts would be just as wrong as asking for Mac fonts.

I feel like the better approach would be to ask for platform-independent fonts. Maybe Google Fonts API doesn't support that today (though I'm curious what happens if you delete `User-Agent` altogether). Might be worth a follow up conversation just to make this a little more durable to changes in font infrastructure.

Issue #26632, opened 2023-12-11
Config option for `application` builder to set the exact path to output files, without hardcoded `browser` path


e-oz:
### Command

build, new

### Description

Please create an option to set the exact output path for "browser" and "server" files for `application` builder, to override the default `browser` path and avoid issues, described in #26304.

Would also be great to have some option in `ng new` to set this path during application/workspace initialization.

### Describe the solution you'd like

As an idea: add `clientFilesFolder` field and give it the default value `'browser'`, so users could re-define it to `''` to get the desired behavior. Maybe `serverFilesFolder` as well (or some better name).

### Describe alternatives you've considered

As we found in #26627, the only existing alternative is a temporary solution that will be removed.

Issue #26057, opened 2023-10-18
[browser-esbuild] add ability to configure processed assets output folder name


MarcoGlauser:
### Command

build

### Description

The esbuild builder will hash asset filenames that are imported from css and put them in the `/media` folder. 
```css
.test {
    background: url('assets/images/next-arrow.svg') center no-repeat;
}
```
Will be transformed into the following:
```css
.test {
    background: url('./media/next-arrow.CZQC2LXJ.svg') center no-repeat;
}
```

While this is a nice feature for cache busting, it clashes with our django server that also serves a `/media` folder. 
For the sake of not breaking existing urls, it would be nice, if there was a way to configure this folder name.
We tried the `resourcesOutputPath` option but that doesn't seem to have any effect. 

### Describe the solution you'd like

It would be great if there was a way to configure the name of the processed asset output folder.

### Describe alternatives you've considered

We can change the media folder for django. However, that would break existing links and I think that the folder name should not be hardcoded in the first place.

I-A-S:
Agreed. Ability to customize this would be a good thing. I can look into implementing this if there's enough interest and the core team approves.

For the time being, if using an custom built Angular-cli is an option for you, start looking at these files and modify as needed. If you don't want to hardcode a new name even for the time being, you can look into adding to angular.json schema as well.

Line 122 of packages/angular_devkit/build_angular/src/builders/application/options.ts

Line 148 of packages/schematics/angular/service-worker/index.ts

Note these are just where angular hardcodes where to output but you might also want to update where it consumes/expects those resources to be too.

Cheers

dgp1130:
We discussed this within the team recently. Normally asset URLs are just left alone because they can be used anywhere in any application and we can't reliably rename them across all usages. In this case the image was copied specifically because it is a file imported from CSS. It's likely the file actually exists at both the asset URL and the `media/` path.

You can opt out of this transformation by adding `assets/*` to `externalDependencies` in the `angular.json` builder. https://github.com/angular/angular-cli/issues/26093#issuecomment-1777401822 This also means the URL with not include the file hash, which may or may not desirable for your use case. This is probably the easiest workaround for your particular use case.

We did briefly debate whether assets should always be considered external, given that we have to assume there is some reference to their path which can't be renamed by the build. However we decided not to because:
1. If asset paths are automatically external, there is no easy to way to disable that behavior. We'd need some other `assetsDefaultToExternal: false` flag, which we'd rather not add.
[...]

PR #26675, opened 2023-12-14
feat(@angular-devkit/build-angular): allow customization of output locations

alan-agius4:
This update introduces the ability for users to define the locations for storing `media`, `browser`, and `server` files.

You can achieve this by utilizing the extended `outputPath` option.
```json
{
  "projects": {
    "my-app": {
      "architect": {
        "build": {
          "options": {
            "outputPath": {
              "base": "dist/my-app",
              "browser": "",
              "server": "node-server",
              "media": "resources"
            }
          }
        }
      }
    }
  }
}
```

While not recommended, choosing to set the `browser` option empty will result in files being output directly under the specified `base` path. It's important to note that this action will generate certain files like `stats.json` and `prerendered-routes.json` that aren't intended for deployment in this directory.

**Validation rules:**
- `browser` and `server` are relative to the configuration set in the `base` option.
- When SSR is enabled, `browser` cannot be set to an empty string, and cannot be the same as `server`.
- `media` is relative to the value specified in the `browser` option.
- `media` cannot be set to an empty string.
- `browser`, `media`, or `server` cannot contain slashes.

Closes: #26632 and closes: #26057

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/build-action.ts
+++ b/packages/angular_devkit/build_angular/src/builders/application/build-action.ts
@@ -17,17 +17,18 @@ import { withNoProgress, withSpinner, writeResultFiles } from '../../tools/esbui
 import { deleteOutputDir } from '../../utils/delete-output-dir';
 import { shouldWatchRoot } from '../../utils/environment-options';
 import { NormalizedCachedOptions } from '../../utils/normalize-cache';
+import { NormalizedOutputOptions } from './options';
 
 export async function* runEsBuildBuildAction(
   action: (rebuildState?: RebuildState) => ExecutionResult | Promise<ExecutionResult>,
   options: {
     workspaceRoot: string;
     projectRoot: string;
-    outputPath: string;
+    outputOptions: NormalizedOutputOptions;
     logger: logging.LoggerApi;
     cacheOptions: NormalizedCachedOptions;
-    writeToFileSystem?: boolean;
-    writeToFileSystemFilter?: (file: BuildOutputFile) => boolean;
+    writeToFileSystem: boolean;
+    writeToFileSystemFilter: ((file: BuildOutputFile) => boolean) | undefined;
     watch?: boolean;
     verbose?: boolean;
     progress?: boolean;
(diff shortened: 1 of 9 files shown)


JoostK:
Should there be a check that `browser` and `server` are not the same?

alan-agius4:
@JoostK, good call out! Updated.

packages/angular_devkit/build_angular/src/builders/application/build-action.ts
```diff
@@ -54,7 +55,7 @@ export async function* runEsBuildBuildAction(
   } = options;
 
   if (deleteOutputPath && writeToFileSystem) {
-    await deleteOutputDir(workspaceRoot, outputPath, ['browser', 'server']);
+    await deleteOutputDir(workspaceRoot, outputOptions);
```
clydin, line 58:
```suggestion
    await deleteOutputDir(workspaceRoot, outputOptions.base, [outputOptions.browser, outputOptions.server]);
```
This would remove the need to change the `deleteOutputDir` file and remove the circular dependencies.

alan-agius4: This is now done internally in the `deleteOutputDir` method.

alan-agius4: Updated

clydin (approved):

e-oz:
Why “media” can not be an empty string? I just want the “assets” folder to be copied into the “assets” folder, what is so wrong with that? It was working for 16 versions and now it is wrong?

alan-agius4:
@e-oz, `assets` and `media` are two unrelated features. 

- `assets`, configured via the `assets` option, consist of files untouched by the build system; they are simply copied as they are.
- `media` are files processed by the build system, commonly referenced in CSS files, like background images and font files.

e-oz:
@alan-agius4 that might be true until we reference files from `assets` in our CSS files. I understand that some people want cache busting for such files, but I don't understand why it's not allowed to opt out.

Anyway, thank you for implementing this, it is much better now ❤️

alan-agius4:
Referencing assets in CSS will result in duplicate emissions — one in the designated assets directory and another in media or the root, with the webpack-based builder.

For proper processing by the build system relative referenced resources in CSS, it's advisable not to store these assets directly in the "assets" directory. Consider reorganising their placement to avoid such duplication.

PR #26681, opened 2023-12-15
feat(@schematics/angular): update SSR and application builder migration schematics to work with new `outputPath`

alan-agius4:
In #26675 we introduced a long-form variant of `outputPath`, this commit updates the application builder migration and ssr schematics to handle this change.


packages/schematics/angular/migrations/update-17/use-application-builder.ts
```diff
@@ -14,7 +14,7 @@ import {
   chain,
   externalSchematic,
 } from '@angular-devkit/schematics';
-import { dirname } from 'node:path';
+import { posix } from 'node:path';
```
clydin, line 17: nit: you can do `node:path/posix` now

packages/schematics/angular/migrations/update-17/use-application-builder_spec.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    const { outputPath, resourcesOutputPath } = app.architect['build'].options;
+    expect(outputPath).toEqual({
+      base: 'dist/project',
+      media: 'resources',
+    });
+    expect(resourcesOutputPath).toBeUndefined();
+  });
+
+  it(`should renove 'browser' portion from 'outputPath'`, async () => {
```
clydin, line 84:
```suggestion
  it(`should remove 'browser' portion from 'outputPath'`, async () => {
```

packages/schematics/angular/migrations/update-17/use-application-builder.ts
```diff
@@ -68,8 +68,30 @@ export default function (): Rule {
...
-          options['outputPath'] = options['outputPath']?.replace(/\/browser\/?$/, '');
+        let outputPath = options['outputPath'];
+        if (typeof outputPath === 'string') {
+          if (!/\/browser\/?$/.test(outputPath)) {
+            // TODO: add prompt.
+            context.logger.warn(
+              `The output location of the browser build have been updated from "${outputPath}" to ` +
+                `"${join(outputPath, 'browser')}". ` +
+                'You might need to adjust your deployment pipeline or, as an alternative, ' +
+                'set outputPath.browser to "" in order to maintain the previous functionality.',
```
clydin, line 79: Since we don't have a prompt right now, should we add the long-form variant in this case to make the configuration adjustment easier?

packages/schematics/angular/migrations/update-17/use-application-builder.ts
```diff
@@ -68,8 +68,30 @@ export default function (): Rule {
...
+            );
+          } else {
+            outputPath = outputPath.replace(/\/browser\/?$/, '');
+          }
+
+          options['outputPath'] = outputPath;
+          if (typeof options['resourcesOutputPath'] === 'string') {
+            const media = options['resourcesOutputPath'].replaceAll('/', '');
```
clydin, line 87: Were `media`, `media/`, `/media`, and `/media/` considered equivalent for `resourcesOutputPath`?

I think this would also treat `med/ia` as being equivalent to `media`.  Although i think that's unlikely to occur.

packages/schematics/angular/migrations/update-17/use-application-builder.ts
alan-agius4: SGTM

packages/schematics/angular/migrations/update-17/use-application-builder.ts
alan-agius4: Yes, all the above were considered the same.

packages/schematics/angular/migrations/update-17/use-application-builder.ts
```diff
@@ -68,8 +68,30 @@ export default function (): Rule {
...
+            if (media && media !== 'media') {
```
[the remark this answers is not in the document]
alan-agius4: I don't think the dot form should be handled. As whilst it works it's unintended.

clydin (approved):

PR #26710, opened 2023-12-19
fix(@angular-devkit/build-angular): add missing tailwind `@screen` directive in matcher

### LINKED ISSUE #26709: Tailwind @screen Directive does not work with Application builder but previously worked with Webpack builder
### Command

build, serve

### Is this a regression?

_No response_

### Description

_I create a new issue based on my previous one https://github.com/angular/angular-cli/issues/26644#issuecomment-1857633589 since it's a different topic now._

I've reproduced the issue successfully with the old, but still valid directive `@screen` (even if not documented). If this isn't supposed to work the ESBuild, then I wonder what the upgrade path (from `@screen` to `@media screen (xy)`) would be. It seems like nothing is mentioned anywhere?

See https://github.com/mirobo/nx-angular-tailwind/tree/main and (maybe you need to run `npm install` it with legacy peer deps). Then `npm run start` (it starts both variants).

Resize the window to see colors of the h1 with "TAILWIND SCREENS TEST" change (the variant with ESBuild won't work)

I see that both executors do something with postcss and autoprefixer internally (which are both added as dependencies in this test project), but only the Webpack executor seems to have a "sass-loader" [webpack/configs/styles.ts](https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/tools/webpack/configs/styles.ts#L226C36-L226C47) while in [esbuild..sass-language.ts](https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/sass-language.ts) and [esbuild..stylesheet-plugin-factory.ts](https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts) I didn't find it.

The styles.css produced by both bundlers is pretty much identical: https://github.com/mirobo/nx-angular-tailwind/tree/main/style-css-output

image

In the generated main.js I see that some processing was not done:

**ESBuild**

image

**Webpack**

image

### Minimal Reproduction

- clone https://github.com/mirobo/nx-angular-tailwind/tree/main
- `npm install`  (maybe run it with legacy peer deps). 
- `npm run start` (it starts both variants, one with Application builder and one with Webpack).
- Resize the window to see how the header `TAILWIND SCREENS TEST` changes color (in case of Webpack) and how it doesn't change color (in case of Application/ESBuild builder)

### Exception or Error

_No response_

### Your Environment

```text
Angular CLI: 17.0.0
Node: 18.14.0
Package Manager: npm 9.3.1
OS: win32 x64

Angular: 17.0.5
... router

@angular-devkit/architect       0.1700.5
@angular-devkit/build-angular   17.0.5
@angular-devkit/core            17.0.5
@angular-devkit/schematics      17.0.5
@angular/cli                    17.0.0
@schematics/angular             17.0.5
ng-packagr                      17.0.2
rxjs                            7.8.0
typescript                      5.2.2
zone.js                         0.14.2
```

### Anything else relevant?

_No response_

alan-agius4:
`@screen` is not documented in tailwind documentation as it is not a recommanded option, however it still works and they don't have plans to remove it.

https://github.com/tailwindlabs/tailwindcss/discussions/7516

Closes #26709

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
@@ -54,7 +54,15 @@ export interface StylesheetPluginOptions {
  *
  * Based on https://tailwindcss.com/docs/functions-and-directives
  */
-const TAILWIND_KEYWORDS = ['@tailwind', '@layer', '@apply', '@config', 'theme(', 'screen('];
+const TAILWIND_KEYWORDS = [
+  '@tailwind',
+  '@layer',
+  '@apply',
+  '@config',
+  'theme(',
+  'screen(',
+  '@screen', // Undocumented in version 3, see: https://github.com/tailwindlabs/tailwindcss/discussions/7516.
+];
 
 export interface StylesheetLanguage {
   name: string;


clydin (approved):

mirobo:
Thank you very much @alan-agius4 ! 🤩🤗

PR #26720, opened 2023-12-20
test: always use package manager force option in version add command E2E

clydin:
The goal of the `commands/add/version-specifier` test is to check the behavior of the `ng add` command itself and not that of the package manager. To do so requires adding a package using a variety of different version specifiers. Some of which a package manager may consider incompatible with the hosting project. However, this is not relevant to the selection logic of the `ng add` command which should select the version as designed. To verify it does so within the Angular CLI, the installed version needs to be checked. As a result the package manager must install the package.


alan-agius4 (approved):
Nice!

PR #26766, opened 2023-12-29
fix(@angular-devkit/schematics): replace template line endings with platform specific

### LINKED ISSUE #26764: Avoid "LF will be replaced by CRLF the next time Git touches it" warning on Windows
### Command

new

### Description

I know the `LF will be replaced by CRLF the next time Git touches it` is a `warning` message that can safely be ignored. Some Angular newbies might feel confused sometimes. Is it possible to avoid this message been shown to the user?

I think this could be configurable for default newline characters. It can be detected easily for user's `core.autocrlf` setting of Git. So it's possible that change the newline of all the generated text file from LF to CRLF when needed.

Workaround 1: Avoid `git init` after `ng new` and do it by user itself

```sh
ng new demo1 --skip-git
cd demo1
git init -b main
git config --local core.autocrlf false
git add . && git commit -m "Initial commit"
```

Workaround 2: Change `core.autocrlf` setting of Git to `false`

```sh
git config --global core.autocrlf false
```

alan-agius4:
Currently, when using `ng new` on Windows, users will get a number of `LF will be replaced by CRLF the next time Git touches it` warnings.

This commit, replaces the line endings in templates to be platform specific.

Closes #26764

### DIFF
--- a/packages/angular_devkit/schematics/src/rules/template.ts
+++ b/packages/angular_devkit/schematics/src/rules/template.ts
@@ -7,7 +7,7 @@
  */
 
 import { BaseException, normalize, template as templateImpl } from '@angular-devkit/core';
-import { TextDecoder } from 'util';
+import { EOL } from 'node:os';
 import { FileOperator, Rule } from '../engine/interface';
 import { FileEntry } from '../tree/interface';
 import { chain, composeFileOperators, forEach, when } from './base';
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/schematics/src/rules/template.ts
```diff
@@ -7,7 +7,8 @@
  */
 
 import { BaseException, normalize, template as templateImpl } from '@angular-devkit/core';
-import { TextDecoder } from 'util';
+import { EOL } from 'node:os';
+import { TextDecoder } from 'node:util';
```
clydin, line 11: `TextDecoder` is a global object so this can be dropped completely.

clydin (approved):

PR #26651, opened 2023-12-12
refactor(@angular-devkit/build-angular): unify proxy config loading i…

m4s7:
…n dev-server and ssr-dev-server

This improvement harmonizes the proxy configuration loading mechanisms between the dev-server and ssr-dev-server. Previously, these servers used different methods for loading the proxy configuration, leading to inconsistencies. Notably, the ssr-dev-server was limited to loading configurations only in JSON format. This change enables the ssr-dev-server to support various configuration formats, aligning it with the dev-server's capabilities and impunify proxy config loading in dev-server and ssr-dev-server

## PR Checklist

Please check to confirm your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular-cli/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

The two dev-servers use different methods for loading the proxy configuration, leading to inconsistencies. Notably, the ssr-dev-server was limited to loading configurations only in JSON format.

Issue Number: N/A

## What is the new behavior?

The proxy configuration loading mechanism is now harmonized between the dev-server and ssr-dev-server. The ssr-dev-server now supports various configuration formats, aligning it with the dev-server's capabilities.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #26801, opened 2024-01-05
fix(@angular-devkit/build-angular): fix support for proxy.conf.js fil…

m4s7:
…es in ssr-dev-server

This fix activates normalization in loadProxyConfiguration for the ssr-dev-server to support formats beyond JSON. This enhancement follows up on the unified proxy configuration loading introduced in PR #26651. Added a test to confirm js config file functionality.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

While the JS format proxy configuration is successfully loaded, it is not interpreted correctly due to the lack of normalization.

Issue Number: #26651

## What is the new behavior?

The interpretation of the JS format proxy configuration is now correct and the regular expressions generated through the normalization process are accurately passed as filters to the proxy.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/ssr-dev-server/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/ssr-dev-server/index.ts
@@ -375,24 +375,32 @@ function getSslConfig(
   return ssl;
 }
 
+function fixWildcardBehavior(regex: string): string {
+  // This fixes an issue with the normalized wildcard regex and the proxy middleware
+
+  return regex.replace(
+    '\\/(?!\\.)(?=.)[^/]*?\\/?$', // Matches paths with one segment
+    '(?:\\/(?!\\.)(?:(?:(?!(?:^|\\/)\\.).)*?)|$)$', // Matches paths with more than one segment
+  );
+}
+
 async function getProxyConfig(root: string, proxyConfig: string): Promise<MiddlewareHandler[]> {
-  const proxy = await loadProxyConfiguration(root, proxyConfig, false);
+  const proxy = await loadProxyConfiguration(root, proxyConfig, true);
   const createdProxies = [];
   const { createProxyMiddleware } = await import('http-proxy-middleware');
+
   for (const [key, context] of Object.entries(proxy)) {
-    if (typeof key === 'string') {
+    if (key.startsWith('^') && key.endsWith('$')) {
       createdProxies.push(
         createProxyMiddleware(
-          key.replace(/^\*$/, '**').replace(/\/\*$/, ''),
+          (pathname) => !!pathname.match(fixWildcardBehavior(key)),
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/builders/ssr-dev-server/specs/proxy_spec.ts
```diff
@@ -106,4 +106,49 @@ describe('Serve SSR Builder', () => {
...
+        response.end();
+      }
+    });
+
+    try {
+      await new Promise<void>((resolve) => proxyServer.listen(0, '127.0.0.1', resolve));
+      const proxyAddress = proxyServer.address() as import('net').AddressInfo;
+
+      host.writeMultipleFiles({
+        'proxy.config.ts': `
```
alan-agius4, line 126: TypeScript files are not supported. In this case it works because the contents are plain JavaScript.

packages/angular_devkit/build_angular/src/builders/ssr-dev-server/index.ts
```diff
@@ -376,23 +376,22 @@ function getSslConfig(
...
 
 async function getProxyConfig(root: string, proxyConfig: string): Promise<MiddlewareHandler[]> {
-  const proxy = await loadProxyConfiguration(root, proxyConfig, false);
+  const proxy = await loadProxyConfiguration(root, proxyConfig, true);
   const createdProxies = [];
   const { createProxyMiddleware } = await import('http-proxy-middleware');
+
   for (const [key, context] of Object.entries(proxy)) {
-    if (typeof key === 'string') {
+    if (key.startsWith('^') && key.endsWith('$')) {
```
alan-agius4, line 384: This seems like it is a breaking change, which is not desired. 

In the below code, the proxy key is modified to behave the same way as webpack-dev-server seeing that this builder is intended to be used with webpack based builders.

Example, with webpack-dev-server,  `/api/*` incorrectly matches `/api/foo/bar` whilst with vite it will not.

packages/angular_devkit/build_angular/src/builders/ssr-dev-server/specs/proxy_spec.ts
m4s7: You are right, I'll change it to js file.

packages/angular_devkit/build_angular/src/builders/ssr-dev-server/index.ts
m4s7: I have created a function with which the generated regex is adapted as before.
The test has also been adapted accordingly to check the behaviour with more than one path segment.

binarybro:
Can we get this merged? We need our proxy to be implemented in code, as we need to resolve properties from the environment.

PR #26815, opened 2024-01-08
feat(@angular-devkit/build-angular): provide default and abbreviated build target support for dev-server and extract-i18n

clydin:
The `buildTarget` options for both the `dev-server` and `extract-i8n` builders now have default values that reflect the recommended and new project generated values. The defaults are as follows where `<current-project>` is the name of the project where the `dev-server` or `extract-i18n` builder target is located:
* `dev-server` --> `<current-project>:build:development`
* `extract-i18n` --> `<current-project>:build`

Additionally, abbreviated target specifiers are now supported for these options. This allows target specifiers such as `::production` which would expand to `<current-project>:build:production` for either builder. Abbreviated target specifiers are only supported for the `buildTarget` option in the `dev-server` and `extract-i18n` builders.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/dev-server/options.ts
+++ b/packages/angular_devkit/build_angular/src/builders/dev-server/options.ts
@@ -34,8 +34,9 @@ export async function normalizeOptions(
 
   const cacheOptions = normalizeCacheOptions(projectMetadata, workspaceRoot);
 
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const buildTarget = targetFromTargetString(options.buildTarget ?? options.browserTarget!);
+  // Target specifier defaults to the current project's build target using a development configuration
+  const buildTargetSpecifier = options.buildTarget ?? options.browserTarget ?? `::development`;
+  const buildTarget = targetFromTargetString(buildTargetSpecifier, projectName, 'build');
 
   // Initial options to keep
   const {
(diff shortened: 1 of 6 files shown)


packages/angular_devkit/build_angular/src/builders/extract-i18n/options.ts
```diff
@@ -33,8 +33,16 @@ export async function normalizeOptions(
   const projectMetadata = await context.getProjectMetadata(projectName);
   const projectRoot = path.join(workspaceRoot, (projectMetadata.root as string | undefined) ?? '');
 
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const buildTarget = targetFromTargetString(options.buildTarget ?? options.browserTarget!);
+  // Target specifier defaults to the current project's build target with no specified configuration
+  const buildTargetSpecifier = options.buildTarget ?? options.browserTarget ?? `:build`;
```
alan-agius4, line 37: Why not provide the default in the schema so that it will also be documented?

packages/angular_devkit/build_angular/src/builders/dev-server/options.ts
```diff
@@ -34,8 +34,16 @@ export async function normalizeOptions(
 
   const cacheOptions = normalizeCacheOptions(projectMetadata, workspaceRoot);
 
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  const buildTarget = targetFromTargetString(options.buildTarget ?? options.browserTarget!);
+  // Target specifier defaults to the current project's build target using a development configuration
+  const buildTargetSpecifier = options.buildTarget ?? options.browserTarget ?? `::development`;
```
alan-agius4, line 38: Same.

packages/angular_devkit/build_angular/src/builders/extract-i18n/options.ts
clydin: It breaks the handling of the `builderTarget` and deprecated `browserTarget` handling. `builderTarget` will always have a value causing `browserTarget` to never be used. Once we remove `browserTarget` we can definitely do that.

alan-agius4 (approved):

Issue #25434, opened 2023-06-23
Exposing the built-in jest config


danny12321:
### Command

config

### Description

 Expose the jest config used by angular so third party apps can use it. For example: [StrykerJS](https://github.com/stryker-mutator/stryker-js) needs this config to run mutation testing on angular project.

### Describe the solution you'd like

It would be nice to have something simular as `npx ng generate config karma` but than for jest.

### Describe alternatives you've considered

I tried reverse engineer the config but it was quite complex and may be prone to change in the future which increases the potential of errors with new releases.

dgp1130:
> it was quite complex and may be prone to change in the future which increases the potential of errors with new releases.

This is exactly why the Jest config is not exposed right now. We've talked about this a good amount on the team to understand the trade-offs being made here. The challenge is that Angular CLI has to make a lot of assumptions about Jest's setup and configuration to provide a reliable experience and for us as developers to effectively reason about what your Jest execution actually looks like and how we can improve it. This is easy to do when we control that configuration and can specify it exactly. When we start allowing user-specific configurations, it becomes very easy for users to break that configuration and run into confusing edge cases.

One of the primary goals of the Angular CLI is to solve these kinds of tooling and configuration problems so users don't have to deal with them as much as possible. We typically try to enable common web development use cases directly and provide a holistic experience as much as we can, rather than implementing a minimal core and asking users to fill in the blanks. To that end, one area we're experimenting with here is trying to see how far we can go without _needing_ to expose the Jest configuration. **Is it possible for Angular CLI to support most common testing use cases through other means without having to expose the Jest config?**

As one example, I expect some users might think "I want to modify the Jest configuration so I can measure code coverage". That's certainly one approach, but code coverage is a pretty common task I would much rather just support code coverage in Angular CLI directly and say "Just use that". Let Angular CLI handle the Jest configuration for you and solve the problem more directly. We don't yet have support for code coverage in the current Jest implementation, but it is something we're hoping to add following this mindset.

Jest support is still experimental, and I think the answer to the above question remains to be seen. What I think would be helpful here is understanding everyone's _motivations_ for modifying the Jest configuration. Mutation testing is one such use case. I'm not terribly familiar with it and will need to do more research to make sure this kind of approach would be the right way of integrating it. But I would love to hear from the rest of the community about what problems you all are trying to solve. If we can collect enough use cases here, I think it will greatly help to answer the above question and understand whether or not exposing the Jest config is truly the best path forward for the Angular ecosystem.

[...]

PR #26852, opened 2024-01-11
fix: ignore custom Jest configuration file and warn if one is found

dgp1130:
Refs #25434.

Currently with Jest we're experimenting with whether or not Angular CLI can meet user needs without exposing the underlying configuration file. See https://github.com/angular/angular-cli/issues/25434#issuecomment-1610298235 for more context.

If hiding the Jest configuration is a challenge for you, please consider if an alternative approach _without_ hooking into the Jest config is reasonable. Otherwise please post your use case to https://github.com/angular/angular-cli/issues/25434 so we can evaluate the best path forward for this feature.

Jest was incorrectly picking up custom configurations, even though they exist outside Jest's root directory. This commit fixes that behavior so Jest does not see user configurations and does not apply them. Ideally we would throw an error if a Jest configuration is found to be more clear that it isn't doing what developers expect. However they may use a Jest config to run other projects in the same workspace outside of Angular CLI and we don't want to prevent that. A warning seems like the best trade off of notifying users that a custom config isn't supported while also not preventing unrelated Jest usage in the same workspace.

I've noticed a few issues where users were using a custom config either causing a problem or attempting to solve one. I want to be up front during the experimentation process what our current stance is for users playing around with Jest so we can collect feedback on how it impacts users earlier.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/jest/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/jest/index.ts
@@ -7,9 +7,10 @@
  */
 
 import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
-import { execFile as execFileCb } from 'child_process';
-import * as path from 'path';
-import { promisify } from 'util';
+import { execFile as execFileCb } from 'node:child_process';
+import * as fs from 'node:fs/promises';
+import * as path from 'node:path';
+import { promisify } from 'node:util';
 import { colors } from '../../utils/color';
 import { findTestFiles } from '../../utils/test-files';
 import { buildApplicationInternal } from '../application';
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -8,6 +8,7 @@
 
 import { BuilderContext, BuilderOutput, createBuilder } from '@angular-devkit/architect';
 import { execFile as execFileCb } from 'child_process';
+import { promises as fs } from 'fs';
```
clydin, line 11:
```suggestion
import fs from 'node:fs/promises';
```

packages/angular_devkit/build_angular/src/builders/jest/index.ts
```diff
@@ -85,6 +103,7 @@ export default createBuilder(
       jest,
 
       `--rootDir="${path.join(testOut, 'browser')}"`,
+      `--config=${path.join(__dirname, 'jest.config.mjs')}`,
```
clydin, line 106: I assume there is no way to prevent Jest from searching for a config?

clydin (approved):

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: Done. Out of curiosity, is there a meaningful difference between `{ promises as fs }` and `fs/promises` or is this just a style / consistency thing? Just wondering if there's some reasoning I'm not aware of.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
dgp1130: Not to my knowledge. I tried `--config /dev/null` but Jest rejects that because it doesn't have the right file extension (.js, .ts, .json, etc.)

Eventually we'll probably need to supply our own config anyways, just to customize certain options not available on the CLI, so I don't think there's much benefit in omitting the file altogether.

packages/angular_devkit/build_angular/src/builders/jest/index.ts
clydin: Effectively the same outcome for this case.  It provides more benefit when using named imports such as:
```
import { readFile } from 'node:fs/promises';
```

alan-agius4 (approved):

PR #26931, opened 2024-01-23
fix(@angular-devkit/build-angular): return 404 for assets that are not found

### LINKED ISSUE #26917: 200 now returned for missing assets instead of 404
### Command

serve

### Is this a regression?

16

### Description

I've updated a project from 16.2.11 to 17.1.0

When an attempt to load a missing asset is made the response is a 200 containing index.html instead of the expected 404.

This happens when using `ng serve` to run development locally, but i'm also seeing the same issue when deploying my production build to a staging server.

### Minimal Reproduction

https://github.com/jamesbrobb/v17-missing-asset-issue

### Exception or Error

```text
200 returned instead of 404 when asset does not exist
```

### Your Environment

```text
Angular CLI: 17.1.0
Node: 20.10.0
Package Manager: npm 10.2.3
OS: darwin x64

Angular: 17.1.0

@angular-devkit/architect       0.1701.0
@angular-devkit/build-angular   17.1.0
@angular-devkit/core            17.1.0
@angular-devkit/schematics      17.1.0
@angular/cdk                    17.0.5
@angular/material               17.0.5
@schematics/angular             17.1.0
typescript                      5.2.2
zone.js                         0.14.3
```

### Anything else relevant?

_No response_

alan-agius4:
This commit updates the vite dev-server to return 404 for assets and files that are not found.

Closes #26917

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/vite/angular-memory-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/tools/vite/angular-memory-plugin.ts
@@ -90,6 +90,7 @@ export function createAngularMemoryPlugin(options: AngularMemoryPluginOptions): 
         map: mapContents && Buffer.from(mapContents).toString('utf-8'),
       };
     },
+    // eslint-disable-next-line max-lines-per-function
     configureServer(server) {
       const originalssrTransform = server.ssrTransform;
       server.ssrTransform = async (code, map, url, originalCode) => {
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/tools/vite/angular-memory-plugin.ts
```diff
@@ -306,3 +314,42 @@ function pathnameWithoutBasePath(url: string, basePath: string): string {
...
+      req.headers.accept.includes('*/*')
+    )
+  ) {
+    next();
+
+    return;
+  }
+
+  req.url = '/index.html';
+  next();
```
clydin, line 344:
```suggestion
  if (
    (req.method === 'GET' || req.method === 'HEAD') &&
    (!req.url || !lookupMimeTypeFromRequest(req.url)) &&
     (!req.headers.accept ||
      req.headers.accept.includes('text/html') ||
      req.headers.accept.includes('text/*') ||
      req.headers.accept.includes('*/*')
    )
  ) {
      req.url = '/index.html';
  }

  next();
```
Consolidates to one `next()` and removes the extra `return`.
Also handles `text/*` which supports `text/html`

clydin (approved):

PR #26941, opened 2024-01-24
fix(@angular-devkit/build-angular): handle handle load event for multiple stylesheets and CSP nonces

### LINKED ISSUE #26932: Angular does not always load CSS bundles properly on sites with multiple stylesheets and CSP nonces enabled

common

### Is this a regression?

Yes

### Description

I recently enabled CSP with nonces for one of our web projects. After that, I started noticing that some styling wasn't always properly loaded. For example, I noticed some styling for our dialogs was sometimes missing. The screenshot below shows a test page with and without the issue:

issue

I noticed that "media=print" was not replaced with "media=all" for the styles.css bundle in the incorrect version. It happened frequently, but not all the time. So it seemed to be some sort of timing issue. Eventually, I narrowed it down to three conditions that need to be met:

1. CSP nonce is set (e.g. <app-root ngCspNonce="[nonce]">)
2. Part of the styling has to be loaded from an external stylesheet, in our case a .css file provided by a CDN. The second part of the styling is contained in a local bundle (e.g. styles.css) and injected automatically by Angular.
3. The external stylesheet has to take at least 50ms to load.

This might sound quite rare. But this issue is now surfacing frequently across seven web projects we run. I suspect many developers will stylesheets provided by external sources, amended by local stylesheets, and combined with CSP nonces.

It isn't easy to reproduce the issue because it requires a server that adds the correct CSP headers. I set up a temporary site at [https://test-cspissue.theliberatorsimproveyourteam.com](https://test-cspissue.theliberatorsimproveyourteam.com) to demonstrate the issue. The source code for the Angular site is available at [https://github.com/christiaanverwijs/angular-csp-issue](https://github.com/christiaanverwijs/angular-csp-issue). To reproduce:

1. In Chrome, use the Developers Tools. Go to "network" and "throttle" to "3G Fast" (this is only to slow the load of the CDN stylesheet to at least 50ms). Also, disable the local cache. Alternatively, just reload the page a few times in a new tab or window (CTRL+F5/F5) and the issue will show up.
2. Load [https://test-cspissue.theliberatorsimproveyourteam.com](https://test-cspissue.theliberatorsimproveyourteam.com)
3. Click 'Open dialog'
4. If the issue is present, the dialog box will appear immediately below the button, and not centered on a darker background (as it should be by the node_modules/@angular/material/prebuilt-themes/indigo-pink.css that we include in the local bundle).
5. If the issue is not present, the dialog box appears nicely in the middle of the page.

Note that this isn't an issue with Material Design, but with Angular. The issue just surfaced most visibly first with a Material Design component.

https://github.com/christiaanverwijs/angular-csp-issue

```true
Angular CLI: 17.1.0
Node: 18.17.1
Package Manager: npm 9.6.7
OS: win32 x64

Angular: 17.1.0
... forms, language-service, material, material-moment-adapter

@angular-devkit/architect       0.1701.0
@angular-devkit/build-angular   17.1.0
@angular-devkit/core            17.1.0
@angular-devkit/schematics      17.1.0
@schematics/angular             17.1.0
typescript                      5.3.3
zone.js                         0.14.3
```

### Anything else?

I found that I could circumvent the issue by defining the "styles" as follows in angular.json. Not ideal, but it seems to work:

**UPDATE JAN 23**: As it happens, the solution below doesn't consistently resolve the issue on a production environment with a full-blown webapp. 

```
                  "styles": [
                     {
                        // the first injected bundle isn't always loaded. so I added a dummy one. 
                        "input": "src/less/styles.less",
                        "bundleName": "dummy"
                     },
                     {
                        "input": "node_modules/@angular/material/prebuilt-themes/indigo-pink.css",
                        "bundleName": "material"
                     },
                     {
                        "input": "src/less/styles.less",
                        "bundleName": "app"
                     }
                  ],
```

My hunch is that the script that is inserted into index.html by Angular when CSP nonces are active executes too quickly. 

```
<script nonce="">(() => {
  const children = document.head.children;
  function onLoad() {this.media = this.getAttribute('ngCspMedia');}
  for (let i = 0; i < children.length; i++) {
    const child = children[i];
    child.hasAttribute('ngCspMedia') && child.addEventListener('load', onLoad);
  }
})();</script>

```

alan-agius4:
The `load` event for each stylesheet may not always be triggered by Google Chrome's handling. Refer to: https://crbug.com/1521256

This results in the media attribute persistently being set to print, leading to distorted styles in the UI. To address this issue, we substitute the onload logic by replacing `link.addEventListener('load', ...` with `document.documentElement.addEventListener('load', ...` and filtering for link tags.

Closes #26932

### DIFF
--- a/packages/angular/ssr/src/inline-css-processor.ts
+++ b/packages/angular/ssr/src/inline-css-processor.ts
@@ -21,23 +21,34 @@ const CSP_MEDIA_ATTR = 'ngCspMedia';
 
 /**
  * Script text used to change the media value of the link tags.
+ *
+ * NOTE:
+ * We do not use `document.querySelectorAll('link').forEach((s) => s.addEventListener('load', ...)`
+ * because this does not always fire on Chome.
+ * See: https://github.com/angular/angular-cli/issues/26932 and https://crbug.com/1521256
  */
 const LINK_LOAD_SCRIPT_CONTENT = [
-  `(() => {`,
-  // Save the `children` in a variable since they're a live DOM node collection.
-  // We iterate over the direct descendants, instead of going through a `querySelectorAll`,
-  // because we know that the tags will be directly inside the `head`.
-  `  const children = document.head.children;`,
-  // Declare `onLoad` outside the loop to avoid leaking memory.
-  // Can't be an arrow function, because we need `this` to refer to the DOM node.
-  `  function onLoad() {this.media = this.getAttribute('${CSP_MEDIA_ATTR}');}`,
-  // Has to use a plain for loop, because some browsers don't support
-  // `forEach` on `children` which is a `HTMLCollection`.
-  `  for (let i = 0; i < children.length; i++) {`,
-  `    const child = children[i];`,
-  `    child.hasAttribute('${CSP_MEDIA_ATTR}') && child.addEventListener('load', onLoad);`,
-  `  }`,
-  `})();`,
+  '(() => {',
+  `  const CSP_MEDIA_ATTR = '${CSP_MEDIA_ATTR}';`,
+  '  const documentElement = document.documentElement;',
+  '  const listener = (e) => {',
+  '    const target = e.target;',
+  `    if (!target || target.tagName !== 'LINK' || !target.hasAttribute(CSP_MEDIA_ATTR)) {`,
+  '     return;',
+  '    }',
+
+  '    target.media = target.getAttribute(CSP_MEDIA_ATTR);',
+  '    target.removeAttribute(CSP_MEDIA_ATTR);',
+
+  // Remove onload listener when there are no longer styles that need to be loaded.
+  '    if (!document.head.querySelector(`link[${CSP_MEDIA_ATTR}]`)) {',
+  `      documentElement.removeEventListener('load', listener);`,
+  '    }',
+  '  };',
+
+  //  We use an event with capturing (the true parameter) because load events don't bubble.
+  `  documentElement.addEventListener('load', listener, true);`,
+  '})();',
 ].join('\n');
 
 export interface InlineCriticalCssProcessOptions {
@@ -62,6 +73,7 @@ interface PartialHTMLElement {
   hasAttribute(name: string): boolean;
   removeAttribute(name: string): void;
   appendChild(child: PartialHTMLElement): void;
+  insertBefore(newNode: PartialHTMLElement, referenceNode?: PartialHTMLElement): void;
   remove(): void;
   name: string;
   textContent: string;
@@ -164,7 +176,7 @@ class CrittersExtended extends Critters {
         // `addEventListener` to apply the media query instead.
         link.removeAttribute('onload');
         link.setAttribute(CSP_MEDIA_ATTR, crittersMedia[1]);
-        this.conditionallyInsertCspLoadingScript(document, cspNonce);
+        this.conditionallyInsertCspLoadingScript(document, cspNonce, link);
       }
 
       // Ideally we would hook in at the time Critters inserts the `style` tags, but there isn't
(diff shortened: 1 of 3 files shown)


packages/angular/ssr/src/inline-css-processor.ts
```diff
@@ -21,23 +21,33 @@ const CSP_MEDIA_ATTR = 'ngCspMedia';
...
+  '    target.media = target.getAttribute(CSP_MEDIA_ATTR);',
+  '    target.removeAttribute(CSP_MEDIA_ATTR);',
+
+  // Remove onload listener when there are no longer styles that need to be loaded.
+  '    if (!document.head.querySelector(`link[${CSP_MEDIA_ATTR}]`)) {',
+  `      documentElement.removeEventListener('load', listener);`,
+  '    }',
+  '  };',
+
+  `  documentElement.addEventListener('load', listener, true);`,
```
crisbeto, line 49: May be worth commenting here that we use an event with capturing (the `true` parameter) because `load` events don't bubble.

packages/angular/ssr/src/inline-css-processor.ts
```diff
@@ -21,23 +21,33 @@ const CSP_MEDIA_ATTR = 'ngCspMedia';
...
+  '    const target = e.target;',
+  `    if (target.tagName !== 'LINK' || !target.hasAttribute(CSP_MEDIA_ATTR)) {`,
+  '     return;',
+  '    }',
+
+  '    target.media = target.getAttribute(CSP_MEDIA_ATTR);',
+  '    target.removeAttribute(CSP_MEDIA_ATTR);',
+
+  // Remove onload listener when there are no longer styles that need to be loaded.
+  '    if (!document.head.querySelector(`link[${CSP_MEDIA_ATTR}]`)) {',
```
crisbeto, line 44: Should this not have the interpolation around `CSP_MEDIA_ATTR`? There's already a constant with the actual name.

packages/angular/ssr/src/inline-css-processor.ts
```diff
@@ -21,23 +21,33 @@ const CSP_MEDIA_ATTR = 'ngCspMedia';
...
-  `    const child = children[i];`,
-  `    child.hasAttribute('${CSP_MEDIA_ATTR}') && child.addEventListener('load', onLoad);`,
-  `  }`,
-  `})();`,
+  '(() => {',
+  `  const CSP_MEDIA_ATTR = '${CSP_MEDIA_ATTR}';`,
+  '  const documentElement = document.documentElement;',
+  '  const listener = (e) => {',
+  '    const target = e.target;',
+  `    if (target.tagName !== 'LINK' || !target.hasAttribute(CSP_MEDIA_ATTR)) {`,
```
crisbeto, line 36: AFAIK the `target` can technically be `null`.

packages/angular/ssr/src/inline-css-processor.ts
alan-agius4: The interpolation is for the generated content. 

```js
(() => {
  const CSP_MEDIA_ATTR = "ngCspMedia";
  const documentElement = document.documentElement;
  const listener = (e) => {
    const target = e.target;
    if (
      !target ||
      target.tagName !== "LINK" ||
      !target.hasAttribute(CSP_MEDIA_ATTR)
    ) {
      return;
    }
    target.media = target.getAttribute(CSP_MEDIA_ATTR);
    target.removeAttribute(CSP_MEDIA_ATTR);
    if (!document.head.querySelector(`link[${CSP_MEDIA_ATTR}]`)) {
      documentElement.removeEventListener("load", listener);
    }
  };
  documentElement.addEventListener("load", listener, true);
})();

```

crisbeto (approved):

PR #26995, opened 2024-01-30
feat(@angular-devkit/build-angular): add JSON build logs when using the application builder

alan-agius4:
This change implements the capability to display JSON build logs in the terminal instead of a format readable by humans. This is particularly useful for hosting providers, as it allows them to effortlessly access the necessary information without having to parse the JSON configuration.

To enable this output, set the `NG_BUILD_LOGS_JSON=1` environment variable. Additionally, warnings, errors, and logs are automatically colorized when the standard output is a WritableStream. You can disable the colors by using the `FORCE_COLOR=0` environment variable.

```
FORCE_COLOR=0 NG_BUILD_LOGS_JSON=1 ng b
{
  "errors": [],
  "warnings": [],
  "outputPaths": {
    "base": "file:///usr/local/test/home//test-project/dist/test-project",
    "browser": "file:///usr/local/test/home//test-project/dist/test-project/browser",
    "server": "file:///usr/local/test/home//test-project/dist/test-project/server"
  },
  "serverEntryFile": "file:///usr/local/test/home//test-project/dist/test-project/server/server.mjs",
  "prerenderedRoutes": [
    "/"
  ]
}
```

```
NG_BUILD_LOGS_JSON=1 ng b
{
  "errors": [],
  "warnings": [],
  "outputPaths": {
    "base": "file:///usr/local/test/home//test-project/dist/test-project",
    "browser": "file:///usr/local/test/home//test-project/dist/test-project/browser",
    "server": "file:///usr/local/test/home//test-project/dist/test-project/server"
  },
  "serverEntryFile": "file:///usr/local/test/home//test-project/dist/test-project/server/server.mjs",
  "prerenderedRoutes": [
    "/"
  ]
}
```

PR #27004, opened 2024-01-31
fix(@angular-devkit/build-angular): add output location in build stats

alan-agius4:
This can be used by users to determine where the output path is located without needing to read the angular.json.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/application/index.ts
@@ -87,8 +87,15 @@ export async function* buildApplicationInternal(
       const result = await executeBuild(normalizedOptions, context, rebuildState);
 
       const buildTime = Number(process.hrtime.bigint() - startTime) / 10 ** 9;
-      const status = result.errors.length > 0 ? 'failed' : 'complete';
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/builders/application/index.ts
```diff
@@ -87,8 +87,15 @@ export async function* buildApplicationInternal(
       const result = await executeBuild(normalizedOptions, context, rebuildState);
 
       const buildTime = Number(process.hrtime.bigint() - startTime) / 10 ** 9;
-      const status = result.errors.length > 0 ? 'failed' : 'complete';
-      logger.info(`Application bundle generation ${status}. [${buildTime.toFixed(3)} seconds]`);
+      const hasError = result.errors.length > 0;
+
+      if (writeToFileSystem && !hasError) {
+        logger.info(`Output location: ${normalizedOptions.outputOptions.base}\n`);
```
alan-agius4, line 93: @clydin, what do you prefer the path to be here absolute or relative to the CWD?

alan-agius4:
Merged into https://github.com/angular/angular-cli/pull/26995

PR #27006, opened 2024-01-31
feat(@schematics/angular): replace `assets` with `public` directory

alan-agius4:
The `assets` directory is confusing for the users and commonly users place "assets" which are not meant to be copied but instead processed by the build system. This causes some files both bundled and copied.

With this change we rename the `assets` directory to `public` and also move the `favicon.ico` inside this newly created directory.


packages/angular/pwa/pwa/index.ts
```diff
@@ -146,11 +146,15 @@ export default function (options: PwaOptions): Rule {
     const { title, ...swOptions } = options;
 
     await writeWorkspace(host, workspace);
+    let assetsDir = posix.join(sourcePath, 'assets');
+    if (!host.exists(assetsDir)) {
+      assetsDir = posix.join(sourcePath, 'public');
+    }
```
markgoho, line 152: maybe there's a style guide that doesn't prefer ternaries? but if not:

```suggestion
    const assetsPath = host.exists(assetsDir)
      ? 'assets'
      : 'public';
    const assetsDir = posix.join(sourcePath, assetsPath);
```

alan-agius4: This will not work, since you are using `assetsDir` prior to being defined.

clydin (approved):

HCenggel:
This isn't a favorable change. Angular has been around for almost a decade now, and during this time, I've worked on numerous projects that rely on it. I'd prefer not to undergo this change.

HCenggel:
This is not a good change. Angular has been around for almost 10 years, and during this decade, I have been involved in maintaining many projects. I hope that this change will not be implemented. In my projects, there are times when files in the assets need to be bundled and copied. Some projects require bundling and copying of JSON, image, and video files, making assets essential.

HCenggel:
This is not a good change.

PR #27010, opened 2024-01-31
refactor(@angular-devkit/build-angular): allow internal externalPackages option to exclude packages

clydin:
To support future use cases, the internal `externalPackages` application builder option now also accepts an object form with an `exclude` array field in addition to the existing boolean value. The `exclude` capability allows for specifying individual packages that should always be bundled when using the external packages functionality. Currently the external packages functionality is only used by the Vite-based development server.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/external-packages-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/external-packages-plugin.ts
@@ -18,25 +18,31 @@ const EXTERNAL_PACKAGE_RESOLUTION = Symbol('EXTERNAL_PACKAGE_RESOLUTION');
  *
  * @returns An esbuild plugin.
  */
-export function createExternalPackagesPlugin(): Plugin {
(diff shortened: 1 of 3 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/external-packages-plugin.ts
```diff
@@ -18,28 +18,35 @@ const EXTERNAL_PACKAGE_RESOLUTION = Symbol('EXTERNAL_PACKAGE_RESOLUTION');
...
   return {
     name: 'angular-external-packages',
     setup(build) {
-      // Safe to use native packages external option if no loader options present
+      // Safe to use native packages external option if no loader options or exclusions present
       if (
-        build.initialOptions.loader === undefined ||
-        Object.keys(build.initialOptions.loader).length === 0
+        !exclusions &&
+        (build.initialOptions.loader === undefined ||
```
alan-agius4, line 30: Nit: consider combining the multiple `Object.keys`

alan-agius4 (approved):

PR #27018, opened 2024-02-01
feat(@angular-devkit/build-angular): provide option to allow automatically cleaning the terminal screen during rebuilds

### LINKED ISSUE #25699: ng build/serve --watch should have a switch to clear the screen before each rebuild
### Command

build, serve

### Description

Reviving #21559, as I'd like to see the same thing implemented, to always have a clear view of what is actually an error and what has already been addressed. Without this it's very confusing to scroll back and forth to see where the last batch of errors are starting, and what was the output from the last build.

Description of #21559 below:

When I'm refactoring large code-base i revised folowing process:

i have ng serve running
i do some changes in bulk without save
save all
i stop the process
clear the console
then start the ng serve again
I'm doing this to see in isolation what I break, fixed what are new issues etc. So I can focus only on what is relevant

it would be nice to have a switch to run ng serve which would clear console everytime it detects change

Currently I'm doing it manually - stopping the process, clearing console and starting ng serve again

alan-agius4:
When setting `"clearScreen": true` to the appliction builder during rebuilds the terminimal screen will be cleaned.

```json
{
  "projects": {
    "my-app": {
      "projectType": "application",
      "root": "",
      "sourceRoot": "src",
      "architect": {
        "build": {
          "options": {
            "clearScreen": true
          }
        }
      }
    }
  }
}
```

Closes #25699

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/build-action.ts
+++ b/packages/angular_devkit/build_angular/src/builders/application/build-action.ts
@@ -19,6 +19,20 @@ import { shouldWatchRoot } from '../../utils/environment-options';
 import { NormalizedCachedOptions } from '../../utils/normalize-cache';
 import { NormalizedOutputOptions } from './options';
 
+// Watch workspace for package manager changes
+const packageWatchFiles = [
+  // manifest can affect module resolution
+  'package.json',
+  // npm lock file
+  'package-lock.json',
+  // pnpm lock file
+  'pnpm-lock.yaml',
+  // yarn lock file including Yarn PnP manifest files (https://yarnpkg.com/advanced/pnp-spec/)
+  'yarn.lock',
+  '.pnp.cjs',
+  '.pnp.data.json',
+];
+
 export async function* runEsBuildBuildAction(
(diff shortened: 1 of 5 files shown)


packages/angular_devkit/build_angular/src/builders/application/build-action.ts
```diff
@@ -164,6 +167,10 @@ export async function* runEsBuildBuildAction(
         break;
       }
 
+      if (clearScreen) {
+        clearTerminal();
```
clydin, line 171:
```suggestion
        console.clear();
```
It essentially does what this call is doing.

packages/angular_devkit/build_angular/src/builders/application/schema.json
```diff
@@ -133,6 +133,11 @@
       },
       "default": []
     },
+    "clearScreen": {
+      "type": "boolean",
+      "default": false,
+      "description": "Automatically clearing the terminal screen during rebuilds."
```
clydin, line 139:
```suggestion
      "description": "Automatically clear the terminal screen during rebuilds."
```

clydin (approved):

PR #27021, opened 2024-02-01
feat(@angular-devkit/build-angular): allow control of Vite-based development server prebundling

clydin:
Previously, the Vite-based development server that is automatically used with the `application` and `browser-esbuild` builders would always use prebundling if the Angular CLI caching was enabled. The development server now has a specific `prebundle` option to allow more control over prebundling while still allowing other forms of caching within the Angular CLI. The `prebundle` option can be a boolean value of `true` or `false` that will enable or disable prebundling, respectively. Additionally, the option also has an object long-form. This long-form enables prebundling and currently contains one property named `exclude`. The `exclude` property supports cases where a package should not be prebundled and rather should be bundled directly into the application code. These cases are not common but can happen based on project specific requirements. If the `prebundle` option is enabled when using the `browser` builder or any other Webpack-based builder, it will be ignored as the Webpack-based development server does not contain such functionality.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/dev-server/schema.json
+++ b/packages/angular_devkit/build_angular/src/builders/dev-server/schema.json
@@ -106,6 +106,24 @@
       "type": "boolean",
       "description": "Force the development server to use the 'browser-esbuild' builder when building. This is a developer preview option for the esbuild-based build system.",
       "default": false
+    },
+    "prebundle": {
+      "description": "Enable and control the Vite-based development server's prebundling capabilities. To enable prebundling, the Angular CLI cache must also be enabled. This option has no effect when using the 'browser' or other Webpack-based builders.",
+      "oneOf": [
+        { "type": "boolean" },
+        {
+          "type": "object",
+          "properties": {
+            "exclude": {
+              "description": "List of package imports that should not be prebundled by the development server. The packages will be bundled into the application code itself.",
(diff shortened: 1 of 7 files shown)


packages/angular_devkit/build_angular/src/builders/dev-server/schema.json
```diff
@@ -106,6 +106,24 @@
       "type": "boolean",
       "description": "Force the development server to use the 'browser-esbuild' builder when building. This is a developer preview option for the esbuild-based build system.",
       "default": false
+    },
+    "prebundle": {
```
alan-agius4, line 110: Can you mention that it only works when caching is enabled?

alan-agius4, line 110: Should this have a default of true?

clydin: Could go either way on the default.  It's enabled by default when using the application builder but its not used with the browser builder.

I'll add the caching info.

clydin: I added some usage warnings that rely on the raw option value being undefined to check whether the option was actually specified. To allow that to work there cannot be a default in the schema.

alan-agius4 (approved):

PR #27047, opened 2024-02-07
feat(@angular-devkit/build-angular): add define build option to application builder

clydin:
The `application` builder now supports a new option named `define`. This option allows global identifiers present in the code to be replaced with another value at build time. This is similar to the behavior of Webpack's `DefinePlugin` which was previously used with some custom Webpack configurations that used third-party builders. The option has similar capabilities to the `esbuild` option of the same name. The documentation for that option can be found here: https://esbuild.github.io/api/#define 
The command line capabilities of the Angular CLI option are not yet implemented and will added in a future change.

The option within the `angular.json` configuration file is of the form of an object. The keys of the object represent the global identifier to replace and the values of the object represent the corresponding replacement value for the identifier. An example is as follows:

```
"define": {
    "SOME_CONSTANT": "5",
    "ANOTHER": "'this is a string literal'"
}
```

All replacement values are defined as strings within the configuration file. If the replacement is intended to be an actual string literal, it should be enclosed in single quote marks. This allows the flexibility of using any valid JSON type as well as a different identifier as a replacement.

Additionally, TypeScript needs to be aware of the module type for the import to prevent type-checking
errors during the build. This can be accomplished with an additional type definition file within the
application source code (`src/types.d.ts`, for example) with the following or similar content:
```
declare const SOME_CONSTANT: number;
declare const ANOTHER: string;
```
The default project configuration is already setup to use any type definition files present in the
project source directories. If the TypeScript configuration for the project has been altered, the
tsconfig may need to be adjusted to reference this newly added type definition file.

An important caveat to the option is that it does not function when used with values contained within Angular metadata such as a Component or Directive decorator. This limitation was present with previous third-party builder usage as well.

### DIFF
--- a/goldens/public-api/angular_devkit/build_angular/index.md
+++ b/goldens/public-api/angular_devkit/build_angular/index.md
@@ -32,6 +32,9 @@ export interface ApplicationBuilderOptions {
     budgets?: Budget_2[];
     clearScreen?: boolean;
     crossOrigin?: CrossOrigin_2;
+    define?: {
+        [key: string]: string;
+    };
     deleteOutputPath?: boolean;
     externalDependencies?: string[];
     extractLicenses?: boolean;
--- a/packages/angular_devkit/build_angular/src/builders/application/options.ts
+++ b/packages/angular_devkit/build_angular/src/builders/application/options.ts
@@ -295,6 +295,7 @@ export async function normalizeOptions(
     budgets,
     deployUrl,
     clearScreen,
+    define,
   } = options;
 
   // Return all the normalized options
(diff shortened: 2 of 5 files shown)


alan-agius4 (approved):

goldens/public-api/angular_devkit/build_angular/index.md
```diff
@@ -32,6 +32,9 @@ export interface ApplicationBuilderOptions {
     budgets?: Budget_2[];
     clearScreen?: boolean;
     crossOrigin?: CrossOrigin_2;
+    define?: {
+        [key: string]: string;
+    };
```
sod, line 37: This is awesome. Thanks a lot. A real usecase is e.g. if you use sentry, you can [omit parts of the library with this](https://docs.sentry.io/platforms/javascript/configuration/tree-shaking/#list-of-flags) like `__SENTRY_DEBUG__: false`

PR #27083, opened 2024-02-12
refactor(@angular-devkit/schematics-cli): rename schematic template files with a template extension

clydin:
Files that contain template instructions should include a `.template` extension to prevent them from being treated as TypeScript/JSON/etc. files. Not doing so can result in parse errors or other failures when the files are incorrectly processed by other tools.


alan-agius4:
The downside of this is that renovate will not be able to pick these up. If we are fine with that we should clean up https://github.com/angular/angular-cli/blob/ae51ed674f6756feb32a92bce33e78b4077ea371/renovate.json#L61-L76

clydin:
I can put those `package.json` files back for now.  The TS files were the main issue here since they are syntactically invalid.

alan-agius4 (approved):

PR #27114, opened 2024-02-16
feat(@angular-devkit/build-angular): add `deployUrl` to application builder

### LINKED ISSUE #26411: Add `--deploy-url` to `application` builder
I am referencing following response from @dgp1130, where is stated that "team does not expect that any existing use cases to suddenly become unsupported". From v17 there is no possibility to reference static assets with absolute url. Using base hrefs is not proper replacement because relative links must stay relative to root when users wants to open links in new tab (ctrl + click) and also for non javascript consumers of server side rendered pages. Combinations with APP_BASE_HREF are not applicable for this use case. Due to this we believe this is regression for no good reason.

              We are still planning to remove `deployUrl` eventually largely because of all the confusion it causes. The exact version it gets dropped is still TBH, but I can confidently say it will *not* be removed in v14, so this shouldn't be an immediate problem for anyone. The blog post in particular should hopefully help address common use cases which have used `deployUrl` in the past and how we recommend moving forward for each of them. We're not expecting any existing use cases to suddenly become unsupported, just that the right approach for you will probably vary by your particular requirements.

_Originally posted by @dgp1130 in https://github.com/angular/angular-cli/issues/22113#issuecomment-1099723870_

   
**Updated 23.2.2024 with minimal reproduction:**
https://github.com/TheTomasJ/angular-router-example

alan-agius4:
This commit adds the deployUrl option ot the application builder.

Closes #26411

### DIFF
--- a/goldens/public-api/angular_devkit/build_angular/index.md
+++ b/goldens/public-api/angular_devkit/build_angular/index.md
@@ -36,6 +36,7 @@ export interface ApplicationBuilderOptions {
         [key: string]: string;
     };
     deleteOutputPath?: boolean;
+    deployUrl?: string;
     externalDependencies?: string[];
     extractLicenses?: boolean;
     fileReplacements?: FileReplacement_2[];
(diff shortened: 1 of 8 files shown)


dgp1130 (commented):
I don't have any objections or strong feelings here. The two questions we should probably answer before merging:

@clydin, how do you feel about undeprecating `deployUrl` in `browser` / `browser-esbuild` for the time being?

Should we give this a more descriptive name for `application` builder (with `deployUrl` as an alias to whatever the new name is)? Potential options here would be `resourceUrl` or `publicPath`. Maybe something else?

clydin (approved):

PR #27120, opened 2024-02-16
fix(@angular-devkit/build-angular): avoid build performance hit when using custom loader

sod:
loaders in esbuild aren't really custom code. You can only choose from the loaders that esbuild provides. So I doubt they require this `externalPackagesPlugin` which comes at a pretty steep cost of 1 second per 1500 `import` statements your project has on every build and watch rebuild.

fixes https://github.com/angular/angular-cli/issues/27116
followup to https://github.com/angular/angular-cli/pull/26923

## PR Checklist

Please check to confirm your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular-cli/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #27146, opened 2024-02-22
fix(@angular-devkit/build-angular): improve rebuild performance with loaders and prebundling

clydin:
Processing overhead has been reduce during resolution when the fallback path is enabled for prebundling due to either prebundling exclusions or custom loader options being present. The resolution logic now follows the same behavior as the underlying bundler's external package option and avoids the need for an additional bundler resolution roundtrip. While the overhead of the plugin itself is still present, the execution cost is now greatly reduced. Further optimizations to conditionally include the plugin based on code analysis heuristics may be considered in the future.

Blocked on a separate refactoring to allow typescript path mapping entries to be added to the exclude list of the plugin.

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/esbuild/external-packages-plugin.ts
+++ b/packages/angular_devkit/build_angular/src/tools/esbuild/external-packages-plugin.ts
@@ -7,9 +7,7 @@
  */
 
 import type { Plugin } from 'esbuild';
-import { extname } from 'node:path';
-
-const EXTERNAL_PACKAGE_RESOLUTION = Symbol('EXTERNAL_PACKAGE_RESOLUTION');
+import { extname, isAbsolute } from 'node:path';
 
 /**
  * Creates a plugin that marks any resolved path as external if it is within a node modules directory.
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/tools/esbuild/external-packages-plugin.ts
```diff
@@ -34,49 +32,31 @@ export function createExternalPackagesPlugin(options?: { exclude?: string[] }):
         return;
       }
 
-      const loaderFileExtensions = new Set(loaderOptionKeys);
+      const loaderFileExtensions = loaderOptionKeys && new Set(loaderOptionKeys);
 
-      // Only attempt resolve of non-relative and non-absolute paths
+      // esbuild's external packages option considers any package-like path as external
+      // which would exclude paths that are absolute or explicitly relative
       build.onResolve({ filter: /^[^./]/ }, async (args) => {
```
sod, line 39: Could this https://github.com/evanw/esbuild/pull/3667 help to exclude this callback to run already in esbuild? Like 

```ts
const loaders = loaderFileExtensions ? `(Array.from(loaderFileExtensions).join('|'))$` : '';
const libraries = '^(@angular|rxjs)\/)`';
const exclude = loaders ? `(${libraries}|${loaders})` : libraries;

build.onResolve({ filter: /^[^./]/, exclude }, async (args) => {
```

alan-agius4: No particularly for this use-case.

sod: Seems like my esbuild PR is not happening. At least I got no feedback for 3 months. No urgency on my part, as we patch angular cli via `yarn patch` to apply https://github.com/angular/angular-cli/pull/27120 ourself.

PR #27198, opened 2024-03-04
refactor(@angular-devkit/build-angular): use async pipeline helper in HTML rewriting stream

clydin:
The Node.js async `pipeline` helper function reduces the amount of infrastructure code needed to pipe HTML content through the parse5 transform stream.

### DIFF
--- a/packages/angular_devkit/build_angular/src/utils/index-file/html-rewriting-stream.ts
+++ b/packages/angular_devkit/build_angular/src/utils/index-file/html-rewriting-stream.ts
@@ -6,7 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { Readable, Writable } from 'stream';
+import { Readable } from 'node:stream';
+import { pipeline } from 'node:stream/promises';
 import { loadEsmModule } from '../load-esm';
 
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_angular/src/utils/index-file/html-rewriting-stream.ts
```diff
@@ -6,7 +6,8 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { Readable, Writable } from 'stream';
+import { Readable } from 'stream';
```
alan-agius4, line 9: nit

```suggestion
import { Readable } from 'node:stream';
```

packages/angular_devkit/build_angular/src/utils/index-file/html-rewriting-stream.ts
```diff
@@ -6,7 +6,8 @@
...
+import { pipeline } from 'stream/promises';
```
alan-agius4, line 10: Nit 
```suggestion
import { pipeline } from 'node:stream/promises';
```

alan-agius4 (approved):
Nice cleanup!

alan-agius4 (approved):

PR #27270, opened 2024-03-13
build: clean up build scripts

alan-agius4:
Several cleanups to build scripts


lib/packages.mts
```diff
@@ -0,0 +1,45 @@
+/**
```
alan-agius4, line 1: For the time being I had to create 2 files. `lib/packages.mts` and `lib/packages.ts`. They have the same content, but are needed because of ts-node, and local-bootstrap.

PR #27364, opened 2024-03-26
docs: replace links links to aio with links to adev

JeanMeche:
Currently those are handled by a custom link remapper (in dev-infra) which should be removed.


packages/angular/cli/src/commands/cache/long-description.md
```diff
@@ -18,7 +18,7 @@ The object goes under `cli.cache` at the top level of the file, outside the `pro
 }
 ```
 
-For more information, see [cache options](guide/workspace-config#cache-options).
+For more information, see [cache options](reference/configs/workspace-config#cache-options).
```
alan-agius4, line 21: Do these anchor tags still work?

packages/angular/cli/src/commands/completion/cli.ts
```diff
@@ -51,7 +51,7 @@ Appended \`source <(ng completion script)\` to \`${rcFile}\`. Restart your termi
           ' Angular CLI. For autocompletion to work, the CLI will need to be on your `$PATH`, which' +
           ' is typically done with the `-g` flag in `npm install -g @angular/cli`.' +
           '\n\n' +
-          'For more information, see https://angular.io/cli/completion#global-install',
+          'For more information, see https://angular.dev/cli/completion#global-install',
```
alan-agius4, line 54: The anchor tag doesn't seem to be working here.

packages/angular/cli/src/utilities/completion.ts
```diff
@@ -88,7 +88,7 @@ Appended \`source <(ng completion script)\` to \`${rcFile}\`. Restart your termi
         ' Angular CLI. For autocompletion to work, the CLI will need to be on your `$PATH`, which' +
         ' is typically done with the `-g` flag in `npm install -g @angular/cli`.' +
         '\n\n' +
-        'For more information, see https://angular.io/cli/completion#global-install',
+        'For more information, see https://angular.dev/cli/completion#global-install',
```
alan-agius4, line 91: This doesn't seem to be pointing to the right section of the doc. (Wrong anchor?)

alan-agius4 (commented):
Please ensure to update https://github.com/angular/angular-cli/blob/main/packages/angular/cli/src/commands/doc/cli.ts to replace AIO with ADEV. It's currently awkward that the main documentation functionality still refers to AIO while the actual documentation points to ADEV.

packages/angular/cli/src/utilities/completion.ts
JeanMeche: There is no anchor (yet) in CLI docs. I'm planning to make that change on the dev-infra part.

JeanMeche:
I was aware about the `doc` command pointing to the AIO search. 
We don't have a search page on ADEV and already reached out to Emma for the alternative we should consider.

alan-agius4:
> I was aware about the `doc` command pointing to the AIO search. We don't have a search page on ADEV and already reached out to Emma for the alternative we should consider.

Can we have an issue in angular/angular tracking this? I am also unsure if we should should just drop the command. @clydin & @dgp1130 thoughts about `ng doc`?

dgp1130:
@alan-agius4, does `ng doc` need to point to the search page? It could just go to the home page or something. Search would just be one more click, so not the end of the world if we don't have a dedicated search page just yet.

I'm personally skeptical of the value of `ng doc`, as I struggle to see any use cases where that would be particularly helpful. I can't imagine many users would discover `ng doc` without previously discovering angular.io or angular.dev. A better approach IMHO would be for `--help` to link to CLI reference or guides for specific commands, which feels much more useful and discoverable to me. In the absence of such a solution though, I don't know that I care enough to drop `ng doc`, it doesn't seem like it's particularly challenging to maintain. I'm fine to defer to @alan-agius4 or @clydin here, I don't think I feel that strongly any direction.

alan-agius4:
The `ng doc` command essentially handles the search for you. For instance, typing `ng doc router` will redirect you to a search page with results related to the router. In my opinion, having `ng doc` only redirect to the documentation site doesn't add much usefulness, not that currently it’s particularly useful. I'm personally fine with removing this command.

JeanMeche:
I'm also on the side of also removing it.

alan-agius4:
@JeanMeche can you please fix the failing test?

alan-agius4 (approved):

PR #27480, opened 2024-04-16
feat(@angular-devkit/build-angular): inject event-dispatcher in SSR HTML page

alan-agius4:
This commit add support in the Angular CLI to inject the event-dispatcher script when using the application builder.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/application/execute-post-bundle.ts
+++ b/packages/angular_devkit/build_angular/src/builders/application/execute-post-bundle.ts
@@ -52,7 +52,6 @@ export async function executePostBundleSteps(
     indexHtmlOptions,
     optimizationOptions,
     sourcemapOptions,
-    ssrOptions,
     prerenderOptions,
     appShellOptions,
     workspaceRoot,
@@ -64,7 +63,7 @@ export async function executePostBundleSteps(
    *
    * NOTE: we don't perform critical CSS inlining as this will be done during server rendering.
    */
-  let indexContentOutputNoCssInlining: string | undefined;
+  let ssrIndexContent: string | undefined;
 
   // When using prerender/app-shell the index HTML file can be regenerated.
   // Thus, we use a Map so that we do not generate 2 files with the same filename.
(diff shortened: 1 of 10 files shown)


alan-agius4:
Blocked on https://github.com/angular/angular/pull/55361 and a FW release.

packages/angular_devkit/build_angular/src/utils/index-file/add-js-action-contract.ts
```diff
@@ -0,0 +1,31 @@
...
+import { readFile } from 'node:fs/promises';
+import { htmlRewritingStream } from './html-rewriting-stream';
+
+let jsActionContractScript: string;
+
+export async function addJsActionContract(html: string): Promise<string> {
+  const { rewriter, transformedContent } = await htmlRewritingStream(html);
+
+  jsActionContractScript ??=
+    '<script type="text/javascript" name="ng-jsa-contract">' +
```
alan-agius4, line 18: @AndrewKushnir, I used a name here do the possibility that an ID would conflict if there are 2 NG apps running on the same, and unfortunately at this point we don't have access to the API ID.

alan-agius4: @clydin mentioned that JSA in 3P has been renamed to event-delegation. Should we change this to something else instead of JSA?

AndrewKushnir: @alan-agius4 this script should only be present on a page once (irrespective of the number of apps on a page), so I think using an "id" attribute should be ok.

> Should we change this to something else instead of JSA?

We can call it "ng-event-contract" instead.

AndrewKushnir: > this script should only be present on a page once (irrespective of the number of apps on a page)

Just to confirm: would that be the case with the change in this PR?

AndrewKushnir: Also cc @iteriani for visibility and context.

alan-agius4: > Just to confirm: would that be the case with the change in this PR?

Yes

AndrewKushnir (approved):

clydin (approved):

PR #27485, opened 2024-04-17
feat(@angular/build): introduce new official build system package

clydin:
The `@angular/build` package has been introduced to house the esbuild/Vite-based build system that was newly introduced as stable in v17. The existing `@angular-devkit/build-angular` package will continue to contain the Webpack-based build system and to ensure backwards compatibility it will also provide aliases for the esbuild/Vite-based build system. The changes here are the first stage of the builder transition and moves the `application` builder into the new package. The application builder can now be accessed when this package is installed directly via `@angular/build:application`.

No changes are required for existing projects.


packages/angular/cli/lib/config/workspace-schema.json
```diff
@@ -389,6 +389,28 @@
               "additionalProperties": false,
               "required": ["builder"]
             },
+            {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "builder": {
+                  "const": "@angular/build:application"
```
alan-agius4, line 397: You need to add the new builder in https://github.com/angular/angular-cli/blob/b50ad71503af3c9d34be596686206b3e16450b32/packages/angular/cli/lib/config/workspace-schema.json#L357

packages/angular/build/BUILD.bazel
```diff
@@ -0,0 +1,115 @@
+load("//tools:defaults.bzl", "pkg_npm", "ts_library")
```
alan-agius4, line 1: It seems that there is no test infra for this package, hence none of the tests are executed.

alan-agius4 (approved):
LGTM

PR #27525, opened 2024-04-24
refactor(@angular/build): remove Webpack-specific options from Vite-only dev-server

clydin:
The `dev-server` build within the `@angular/build` package no longer uses Webpack. As a result, the Webpack related options have been removed from the builder schema. These options were not previously used by the Vite-based development server with the exception of `browserTarget` which was removed due to it being deprecated on the `@angular-devkit/build-angular` version.


packages/angular/build/src/builders/dev-server/schema.json
```diff
@@ -102,11 +79,6 @@
       "type": "number",
       "description": "Enable and define the file watching poll time period in milliseconds."
     },
-    "forceEsbuild": {
-      "type": "boolean",
-      "description": "Force the development server to use the 'browser-esbuild' builder when building. This is a developer preview option for the esbuild-based build system.",
-      "default": false
-    },
     "prebundle": {
```
alan-agius4, line 82: Can you update the description here to remove the webpack part?

alan-agius4 (approved):

PR #27561, opened 2024-04-29
refactor(@angular/build): move postcss stylesheet processor to an optional peer dependency

clydin:
The Postcss stylesheet processor is now an optional peer dependency of the newly introduced `@angular/build` build system package. This change removes the need for the package to have a direct dependency on the `postcss` package and allows some version customization of the optional `postcss` package if needed. Postcss plugins (including Tailwind) are still fully supported but usage will now require installing the `postcss` package within the Angular workspace. An error with instructions to install the package will be generated during a build if the package is not present. This change only affects direct usage of the new `@angular/build` package. The `@angular-devkit/build-angular` package which is currently used for all existing projects will continue to contain and directly depend on the `postcss` package.

No changes are required for existing applications.

### DIFF
--- a/packages/angular/build/package.json
+++ b/packages/angular/build/package.json
@@ -50,6 +50,7 @@
     "@angular/platform-server": "^18.0.0 || ^18.0.0-next.0",
     "@angular/service-worker": "^18.0.0 || ^18.0.0-next.0",
     "less": "^4.2.0",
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):
LGTM, we’d also need to add this as direct dependency when running the application builder migration

PR #27572, opened 2024-05-01
build: replace `@angular/build` version with archive in local builds

clydin:
When creating a local development build of the repository via `yarn build --local`, the package version for `@angular/build` within `@angular-devkit/build-angular` will now correctly be replaced with the path to the locally built archive of `@angular/build`.


alan-agius4 (approved):

alan-agius4:
I suspect that https://github.com/angular/angular-cli/commit/eb8f1faa4cf5712c0ada23e3d12e6033eb416473 will no longer be needed after this change.

PR #27635, opened 2024-05-13
perf(@angular/build): add persistent caching of JavaScript transformations

clydin:
When caching is enabled for the Angular CLI, the JavaScript transformer within the application build system will now perform persistent caching of its output. This allows for improved warm build times by removing the need to run the Angular linker and build optimizer steps against unchanged third-party code within each build. This does not affect hot rebuilds that would take place during watch mode since the outputs are already cached within memory. The on-disk storage of the cached data is handled by the `lmdb` package which provides fast key/value storage and built-in off-thread compression. This package is currently used by such projects as Gatsby and parcel for their respective caching subsystems.

Warm production build of a new 18.0.0-rc.2 project with disabled caching:
```
Application bundle generation complete. [3.698 seconds]
```
With enabled caching:
```
Application bundle generation complete. [2.277 seconds]
```

### DIFF
--- a/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
+++ b/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
@@ -19,6 +19,7 @@ import assert from 'node:assert';
 import * as path from 'node:path';
 import { maxWorkers, useTypeChecking } from '../../../utils/environment-options';
 import { JavaScriptTransformer } from '../javascript-transformer';
+import { LmbdCacheStore } from '../lmdb-cache-store';
 import { LoadResultCache, createCachedLoad } from '../load-result-cache';
 import { logCumulativeDurations, profileAsync, resetCumulativeDurations } from '../profiling';
 import { BundleStylesheetOptions } from '../stylesheets/bundle-options';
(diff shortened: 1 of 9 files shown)


alan-agius4 (approved):

packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -62,7 +63,17 @@ export function createCompilerPlugin(
       const preserveSymlinks = build.initialOptions.preserveSymlinks;
 
       // Initialize a worker pool for JavaScript transformations
-      const javascriptTransformer = new JavaScriptTransformer(pluginOptions, maxWorkers);
+      let cacheStore;
+      if (pluginOptions.sourceFileCache?.persistentCachePath) {
+        cacheStore = new LmbdCacheStore(
+          pluginOptions.sourceFileCache.persistentCachePath + '/angular-compiler.db',
```
sod, line 69: I guess to make the 5 windows users happy, this could be:
```ts
path.join(pluginOptions.sourceFileCache.persistentCachePath, 'angular-compiler.db')
```

clydin: Windows actually supports both path separators; but, yes it could.

PR #27687, opened 2024-05-22
fix(@angular/build): only import persistent cache store with active caching

clydin:
To avoid loading and initializing the underlying persistent cache store when caching is disabled, the lmdb-based cache will now only be imported when caching is enabled.


d3lm (commented):
👍

alan-agius4 (approved):

Issue #27702, opened 2024-05-23
"Invalid self-closing element in index HTML file" error when using SVG in said index.html file since Angular 18


wartab:
### Command

build, serve

### Is this a regression?

- [X] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

17.x.x

### Description

We're getting an issue when including an SVG tag (which expects valid XML) directly in our index.html file.

### Minimal Reproduction

```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
    <base href="/">
    <meta
            name="viewport"
            content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
    <link rel="icon" type="image/x-icon" href="favicon.ico">
    <style type="text/css">
        .main-loading-screen img {
            width: 100px;
            height: auto;
        }

        .main-loading-screen svg {
            animation-name: spin-loader;
            font-size: 6em;
            animation-iteration-count: infinite;
            animation-duration: 1s;
            animation-timing-function: steps(8);
        }

        @keyframes spin-loader {
            0% {
                transform: rotate(0deg);
            }

            100% {
                transform: rotate(360deg);
            }
        }
    </style>
</head>
<body>
<app>
    <div class="loading-screen main-loading-screen">
        <svg xmlns="http://www.w3.org/2000/svg" height="1em" viewBox="0 0 512 512">
            <defs>
                <style>.fa-secondary {
                    fill: currentColor;
                    opacity: 0.4;
                }
                .fa-primary {
                    fill: currentColor;
                }
                </style>
            </defs>
            <path class="fa-primary"
                  d="M256 0c-17.7 0-32 14.3-32 32V96c0 17.7 14.3 32 32 32s32-14.3 32-32V32c0-17.7-14.3-32-32-32zM0 256c0 17.7 14.3 32 32 32H96c17.7 0 32-14.3 32-32s-14.3-32-32-32H32c-17.7 0-32 14.3-32 32zM120.2 75C107.7 62.5 87.5 62.5 75 75s-12.5 32.8 0 45.3l45.3 45.3c12.5 12.5 32.8 12.5 45.3 0s12.5-32.8 0-45.3L120.2 75z"/>
            <path class="fa-secondary"
[...]

PR #27703, opened 2024-05-23
fix(@angular/build): support valid self-closing SVG tags in HTML index file

alan-agius4:
Prior to this change, the Angular build process incorrectly flagged self-closing SVG tags (e.g., `<path />`) as invalid HTML. This commit rectifies this issue by explicitly allowing self-closing syntax for SVG elements embedded within the HTML index file.

Closes: #27702


clydin (approved):

wartab:
It seems unlikely that people will actually do this, but MathML also contains some self-closing tags such as mspace. Not sure if that needs to be supported

Issue #27674, opened 2024-05-22
Node.js v22.2 causes ng build hangs up


lacolaco:
### Command

build

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

When the Node.js version is v22.2.0 (released on 2024-05-15), the `ng build` command hangs up. Node.js v22.1.0 doesn't.
Angular CLI is not supporting that version of Node.js officially, but on Firebase App Hosting, the build environment is using Node.js v22.2 today by default. So it is an easy situation for developers to get stumped.

My workaround is to set `engines.node` field in `package.json` as `"20"`. App Hosting builder can determine the Node.js version to install. 

### Minimal Reproduction

- set Node.js version to v22.2.0
- `ng new` on `@angular/cli@latest` (v17.3.7)
- `ng build`
- print `Building...` and never finished.

### Your Environment

```text
Angular CLI: 17.3.7
Node: 22.2.0 (Unsupported)
Package Manager: pnpm 8.15.6
OS: darwin arm64

Angular: 17.3.9
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, platform-server
... router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1703.7
@angular-devkit/build-angular   17.3.7
@angular-devkit/core            17.3.7
@angular-devkit/schematics      17.3.7
@angular/cli                    17.3.7
@angular/ssr                    17.3.7
@schematics/angular             17.3.7
rxjs                            7.8.1
typescript                      5.4.5
zone.js                         0.14.6
    
Warning: The current version of Node (22.2.0) is not supported by Angular.
```

JoostK:
Does this repro using Angular 18 RCs?

lacolaco:
@JoostK Yes, it reproduces the same situation even after `ng update @angular/core @angular/cli --next`

```
Angular CLI: 18.0.0-rc.3
Node: 22.2.0
Package Manager: pnpm 8.15.6
OS: darwin arm64

Angular: 18.0.0-rc.3
... animations, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, platform-server
... router, ssr

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1800.0-rc.3
@angular-devkit/build-angular   18.0.0-rc.3
@angular-devkit/core            18.0.0-rc.3
[...]

PR #27721, opened 2024-05-27
     fix(@angular/build): address prerendering in-memory ESM resolution in Node.js 22.2.0 and later

alan-agius4:
Node.js 22.2.0 introduced a breaking change affecting custom ESM resolution.
    
For more context, see: [Node.js issue #53097](https://github.com/nodejs/node/issues/53097)
    
Closes: #27674

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -95,15 +95,15 @@ nodejs_register_toolchains(
     name = "node22",
     # The below can be removed once @rules_nodejs/nodejs is updated to latest which contains https://github.com/bazelbuild/rules_nodejs/pull/3701
     node_repositories = {
-        "22.0.0-darwin_arm64": ("node-v22.0.0-darwin-arm64.tar.gz", "node-v22.0.0-darwin-arm64", "ea96d349cfaa67aa87ceeaa3e5b52c9167f7ac302fd8d1ff162d0785e9dc0785"),
-        "22.0.0-darwin_amd64": ("node-v22.0.0-darwin-x64.tar.gz", "node-v22.0.0-darwin-x64", "422a3887ff5418f0a4552d89cf99346ab8ab51bb5d384660baa88b8444d2c111"),
-        "22.0.0-linux_arm64": ("node-v22.0.0-linux-arm64.tar.xz", "node-v22.0.0-linux-arm64", "83711d29cbe46375bdffab5419f3d831892e24294169272f6c39edc364556241"),
-        "22.0.0-linux_ppc64le": ("node-v22.0.0-linux-ppc64le.tar.xz", "node-v22.0.0-linux-ppc64le", "2b3fb8707a79243bfb3131312b86716ddc3855bce21bb168095b6b916798e5e9"),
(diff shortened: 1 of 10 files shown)


alan-agius4:
Blocking as there are discussion about reverting the Node.js breaking change https://github.com/nodejs/node/issues/53182

GeoffreyBooth:
@alan-agius4 Do you mind explaining the approach taken in this PR? It seems like you refactored from using worker threads to using child processes?

alan-agius4:
@GeoffreyBooth, the refactoring in this PR involves the main process forking a child process and applying the `--loader` flag. The child process will then be responsible for creating the thread workers.


Before
```
[main process] -> [thread workers --loader]
```
Now
```
[main process] -> [forked process --loader] -> [thread workers]
```

GeoffreyBooth:
> the refactoring in this PR involves the main process forking a child process and applying the `--import` flag. The child process will then be responsible for creating the thread workers.

Thanks. Do you have a way to measure the performance impact of using child processes instead of workers?

So you understand the context, we’re wondering about the necessity of supporting different hooks for different threads, as doing so adds a large amount of complexity to a feature that’s already quite complex.

alan-agius4:
@GeoffreyBooth, I can run some tests and provide you with some numbers tomorrow, as I am based in Europe. However, the performance impact will vary depending on the application's size and complexity. In a simple "hello world" application, the performance regression is negligible. But as the application grows, the additional IPC layer introduced by the forked process might become a bottleneck, especially when transferring large amounts of data between the main process and the thread worker through the forked process and vice-versa.

GeoffreyBooth:
> However, the performance impact will vary depending on the application's size and complexity.

Of course, I was just hoping to get a general idea of the difference for an app of nontrivial size.

Once you're using child processes, is there a reason to use worker threads?

alan-agius4:
> Once you're using child processes, is there a reason to use worker threads?

Yes, we use worker threads to prerender an N number of routes in parallel. Additionally, we need to render the application in isolation to ensure that `globalThis` and the global scopes of the forked process remain unaltered by the rendering process, which could otherwise be affected and modified by running the application.

> Thanks. Do you have a way to measure the performance impact of using child processes instead of workers?

Actually, this highlighted a bug in the above implementation, which I haven't had time to fix yet. The issue is that in big apps we are sending a large amount of data from the forked process to the main process via a single message, causing an `Invalid string length` error. I suspect that there will be an extra latency introduced by the additional IPC layer in the forked process. Previously, the worker process sent data directly to the main process. However, now the data flows through an intermediate forked process, adding complexity to the communication.

GeoffreyBooth:
> Yes, we use worker threads to prerender an N number of routes in parallel.

Right, but why not use child processes for this purpose? What advantages do worker threads provide? (Sincere question.)

Along those lines, why spawn X number of child processes which then spawn Y number of worker threads? Why not just fork Y number of child processes directly?

I was assuming that threads would be faster and the only question was how much, but then I did some searching and found this: https://github.com/orgs/nodejs/discussions/44264. I’m curious to see if your test case shows worker threads to be faster, once you resolve the issues on this branch. If this branch _is_ faster, would you want to keep this new implementation even if worker threads became an option again? If not, why not? (All sincere questions 😄)

Issue #27726, opened 2024-05-27
No source of warning on SCSS files


Delagen:
### Command

build, serve

### Is this a regression?

- [X] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

Webpack Builder

### Description

When using new ESBuild based builder cannot find source of warning in console

### For component styles
```
[WARNING] <warning text> [plugin angular-sass]
```

### For global styles
```
[WARNING] <warning text> [plugin angular-sass]
    angular:styles/global:styles:1:8:
      1 │ @import 'src/styles.scss';
        ╵         ~~~~~~~~~~~~~~~~~
```

### For component styles
```
./src/.../some.component.scss?ngResource - Warning: Module Warning (from ./node_modules/sass-loader/dist/cjs.js):
<warning text>
<source file path> <line>:<pos>  context function()
... <all stack>
src/.../some.component.scss <line>:<pos>  root stylesheet
```

### For global styles
```
./src/styles.scss - Warning: Module Warning (from ./node_modules/sass-loader/dist/cjs.js):
<warning text>
<source file path> <line>:<pos> context function()
... <all stack>
src/styles.scss <line>:<pos> root stylesheet
```

### Minimal Reproduction

Any project with specified build system

place anywhere in component stylesheet file or imported file into global stylesheet
```
@warn 'some message';
```

Build

### Your Environment

```text
_                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 17.3.7
Node: 20.13.1
Package Manager: npm 10.8.0
OS: win32 x64

Angular: 17.3.9
... animations, cdk, common, core, elements, forms, localize
... platform-browser, platform-browser-dynamic, platform-server
... router, service-worker

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1703.7
@angular-devkit/build-angular   17.3.7
@angular-devkit/core            17.3.7
@angular-devkit/schematics      17.3.7
@angular/cli                    17.3.7
@angular/pwa                    17.3.7
@schematics/angular             17.3.7
rxjs                            7.8.1
typescript                      5.4.5
[...]

PR #27732, opened 2024-05-28
fix(@angular/build): print Sass `@warn` location

alan-agius4:
When using the `@warn` directive, the `span` entry in the warning object is often undefined. Instead, the `stack` property is populated.

```js
{
  "warnings": [
    {
      "deprecation": false,
      "deprecationType": null,
      "span": undefined,
      "stack": "projects/foo/src/app/app.component.scss 1:1  root stylesheet\n",
      "message": "some message"
    }
  ]
}
```

### Before
```
▲ [WARNING] some message [plugin angular-sass]
```

### Now
```
▲ [WARNING] some message [plugin angular-sass]

  projects/foo/src/app/app.component.scss 1:1  root stylesheet
```

Closes: #27726

### DIFF
--- a/packages/angular/build/src/tools/esbuild/stylesheets/sass-language.ts
+++ b/packages/angular/build/src/tools/esbuild/stylesheets/sass-language.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import type { OnLoadResult, PartialMessage, ResolveResult } from 'esbuild';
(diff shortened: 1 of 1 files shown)


clydin (approved):

alan-agius4:
> This seems like a bug on the Sass side. If it can generate a stack trace, it knows the location or is it always at `1:1` in this case? Maybe add a comment that this is a workaround for a lack of a warning location?

It knows the location.

PR #27752, opened 2024-05-30
perf(@angular/build): use direct transpilation with isolated modules

clydin:
When using the application builder and the TypeScript `isolatedModules` option is enabled and script sourcemaps are disabled, TypeScript code will be transpiled via the bundler instead of the current behavior of using TypeScript. The use of the `isolatedModules` option ensures that TypeScript code can be safely transpiled without the need for the type-checker. This mode of operation has several advantages. The bundler (esbuild in this case) will know have knowledge of the TypeScript code constructs, such as enums, and can optimize the output code based on that knowledge including inlining both const and regular enums where possible. Additionally, this allows for the removal of the babel-based optimization passes for all TypeScript code. These passes are still present for all JavaScript code such as from third-party libraries/packages. These advantages lead to an improvement in build time, especially in production configurations. To ensure optimal output code size in this setup, the `useDefineForClassFields` TypeScript option should either be removed or set to `true` which enables ECMAScript standard compliant behavior.

Initial testing reduced a warm production build of a newly generated project from ~2.3 seconds to ~2.0 seconds.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/angular/compilation/aot-compilation.ts
+++ b/packages/angular/build/src/tools/esbuild/angular/compilation/aot-compilation.ts
@@ -199,9 +199,12 @@ export class AotCompilation extends AngularCompilation {
 
   emitAffectedFiles(): Iterable<EmitFileResult> {
     assert(this.#state, 'Angular compilation must be initialized prior to emitting files.');
-    const { angularCompiler, compilerHost, typeScriptProgram, webWorkerTransform } = this.#state;
-    const buildInfoFilename =
-      typeScriptProgram.getCompilerOptions().tsBuildInfoFile ?? '.tsbuildinfo';
+    const { affectedFiles, angularCompiler, compilerHost, typeScriptProgram, webWorkerTransform } =
+      this.#state;
+    const compilerOptions = typeScriptProgram.getCompilerOptions();
+    const buildInfoFilename = compilerOptions.tsBuildInfoFile ?? '.tsbuildinfo';
+    const useTypeScriptTranspilation =
+      !compilerOptions.isolatedModules || !!compilerOptions.sourceMap;
 
     const emittedFiles = new Map<ts.SourceFile, EmitFileResult>();
     const writeFileCallback: ts.WriteFileCallback = (filename, contents, _a, _b, sourceFiles) => {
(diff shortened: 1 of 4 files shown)


packages/angular/build/src/tools/esbuild/angular/compilation/aot-compilation.ts
```diff
@@ -228,11 +234,33 @@ export class AotCompilation extends AngularCompilation {
...
+          undefined,
+          transformers,
+        )
+      ) {
+        // Empty
+      }
+    } else if (compilerOptions.tsBuildInfoFile) {
+      // Manually get the builder state for the persistent cache
+      // The TypeScript API currently embeds this behavior inside the program emit
+      // via emitNextAffectedFile but that also pplies all internal transforms.
```
eneajaho, line 253:
```suggestion
      // via emitNextAffectedFile but that also applies all internal transforms.
```

packages/angular/build/src/tools/esbuild/angular/compilation/aot-compilation.ts
```diff
@@ -245,11 +269,37 @@ export class AotCompilation extends AngularCompilation {
...
+      }
+
+      if (useTypeScriptTranspilation) {
+        typeScriptProgram.emit(sourceFile, writeFileCallback, undefined, undefined, transformers);
         continue;
       }
 
-      typeScriptProgram.emit(sourceFile, writeFileCallback, undefined, undefined, transformers);
+      // When not using TypeScript transpilation, directly apply only Angular specific transformations
+      const transformResult = ts.transform(
```
JoostK, line 285: It would be nice if we'd know here if the compiler has any transforms for this source file pending, otherwise all of this work is essentially a noop. This is known in `TraitCompiler` (its `filesWithoutTraits` field, or inversely when the source file is not present in `fileToClasses`).

I also wonder how using the printer affects source mapping, as esbuild will now map into transformed TS and there's no mapping back to the original source file, as the printer is incapable of generating source maps TMK.

clydin: Definitely opportunities to further enhance with some potential adjustments to the AOT compiler.

Sourcemaps are the main reason this is still draft.  First stage will probably be limiting this to when sourcemaps are disabled which is most production builds. Production builds are also where this change provides the most benefit.
After that, exploring sourcemap production from the TS side would be the next step.  TS does export the needed elements (`createSourceMapGenerator`, for instance) from the JS but are not present in the typings.

PR #27758, opened 2024-05-31
fix(@angular/build): disable JS transformer persistent cache on web containers

clydin:
The backing store for the persistent caching of the JavaScript transformers is not currently supported on web containers. To allow other potential forms of caching, the JavaScript transformer caching is now selectively disabled when executed on a web container. This allows for development server prebundling to be used on web containers if needed.


alan-agius4 (approved):

clydin:

PR #27763, opened 2024-06-02
fix(@angular/build): watch all related files during a Sass error

clydin:
When `dart-sass` returns an error, the provided location span may not be the exact location of the error but can instead be the location of the `@error` call within a Sass function. The generated stack trace string does contain all files related to the error including the usage site in the case of the function `@error` case. To ensure all related files for the error are watched and allow for a rebuild that may correct the error, the stack trace is parsed and found files are added to the watch list. Unfortunately, `dart-sass` does not provide the list directly. If that changes in the future, the stack trace parsing could be removed.


alan-agius4 (approved):
Looks good to me. However, I wonder if we could use the the `importers` instead of the stack trace. This would ensure that all files up to the point of the exception are added to the watched list.

Basically in the importers we keep track of the resolved/containing file.

dgp1130:

Issue #27766, opened 2024-06-03
A silent error is thrown in `AddCommandModule.builder`, when passing a version to the command `ng add @my/lib@<version>`


Platonn:
### Command

add

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

## Error
An exception is thrown during the method [createCollection()](https://github.com/angular/angular-cli/blob/ea843ebfeaa2e1a232bf199b624f7084e35399b9/packages/angular/cli/src/commands/add/cli.ts#L98), when the `collectionName` contains a `@<version>`, e.g. `ng add @my/lib@1.2.3`.

## Negative consequences
The consequence is that the custom schematics' schema is not applied for parsing, i.e. the method [addSchemaOptionsToCommand()](https://github.com/angular/angular-cli/blob/ea843ebfeaa2e1a232bf199b624f7084e35399b9/packages/angular/cli/src/commands/add/cli.ts#L101) is skipped, becasue of the thrown exception

And btw. the later this exception is ignored [here](https://github.com/angular/angular-cli/blob/ea843ebfeaa2e1a232bf199b624f7084e35399b9/packages/angular/cli/src/commands/add/cli.ts#L102-L107). 

## Why it happens (internal Angular CLI code)

The error happens becasue eventually [this line](https://github.com/angular/angular-cli/blob/ea843ebfeaa2e1a232bf199b624f7084e35399b9/packages/angular_devkit/schematics/tools/node-module-engine-host.ts#L62) throws an error, becasue there is no such a directory like `node_modules/@my/lib@1.2.3`, but only `node_modules/@my/lib`

Because an error is thrown, a custom schema.json is not applied for parsing the arguments then.

### Idea of a fix:
Before passing the `collectionName` as an argument to the `createCollection()` method, we could first extract any versions from it, e.g. by passing it through `npa from 'npm-package-arg'`, e.g.:

```ts
private async getCollectionName(): Promise<string> {
    const [, collection] = this.context.args.positional;

    const packageIdentifier = npa(collection);
    return packageIdentifier.name;
  }
```

## Additional notes
- ❌ `ng add @my/lib@1.2.3` - causes error
- ❌ `ng add @my/lib@` (with just `@`) - also causes error
- ✅  `ng add @my/lib` (nothing after the package's name)  - doesn't cause error

### Minimal Reproduction

1. Create a fresh angular 17 app
[...]

PR #27767, opened 2024-06-03
fix(@angular/cli): add `schema.json` options to parsed command, also when a version is passed to `ng add <package>@<version>`

Platonn:
This commit fixes the method `AddCommandModule.getCollectionName()`, so it now returns only the package name, but remove the specified version (if present).

Previously, a `@<version>` specified in the const `collectionName` was causing a (silenced) error during the invocation of `workflow.engine.createCollection(collectionName)`, which lead to skipping eventually the invocation of the method `this.addSchemaOptionsToCommand()` in the try/catch block.

fixes https://github.com/angular/angular-cli/issues/27766

## PR Checklist

- [x] N/A Docs have been added / updated (for bug fixes / features)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/angular/angular-cli/issues/27766

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/cli/src/commands/add/cli.ts
+++ b/packages/angular/cli/src/commands/add/cli.ts
@@ -346,7 +346,17 @@ export default class AddCommandModule
   }
 
   private async getCollectionName(): Promise<string> {
-    const [, collectionName] = this.context.args.positional;
+    let [, collectionName] = this.context.args.positional;
+
+    // The CLI argument may specify also a version, like `ng add @my/lib@13.0.0`,
+    // but here we need only the name of the package, like `@my/lib`
+    try {
+      const packageIdentifier = npa(collectionName);
+      collectionName = packageIdentifier.name ?? collectionName;
+    } catch (e) {
+      assertIsError(e);
+      this.context.logger.error(e.message);
+    }
 
     return collectionName;
   }


packages/angular/cli/src/commands/add/cli.ts
```diff
@@ -346,7 +346,17 @@ export default class AddCommandModule
   }
 
   private async getCollectionName(): Promise<string> {
-    const [, collectionName] = this.context.args.positional;
+    let [, collectionName] = this.context.args.positional;
+
+    // The CLI argument may specify also a version, like `ng add @my/lib@13.0.0`,
+    // but here we need only the name of the package, like `@my/lib`
+    try {
+      const packageIdentifier = npa(collectionName);
```
alan-agius4, line 354: We don't need to use NPA for this. This can be done by using `lastIndexOf`

```ts
const versionIdx = collectionName.lastIndexOf('@');
if (versionIdx > 0) {
    collectionName = collectionName.slice(0, versionIdx)
}
```

Platonn: Many thanks for your review @alan-agius4 ! 🙌 

Agreed, we don't need NPM. I've ushed changes now with a code comment on disregarding the edge-case of npm-aliases (that use `@` separator for other purposes).

All that said, IMHO the solution with NPA was cleaner - no custom code to maintain possibly in the future; no need to add code comments to explain assumptions and disregard of edge cases). Also I was inspired by the usage of NPA in the `run()` method. The runtime cost of executing NPA is low, because we call the method `getCollectionName` only once. So I wouldn't mind reverting to the original solution with NPA, if you like it more. Please let me know :) 

Source: https://docs.npmjs.com/cli/v8/commands/npm-install

alan-agius4: I am okay with using NPA, however it's worth noting that aliases will not work even if we use NPA.

Platonn: Thanks for your reply. I've reverted to NPA solution now.

alan-agius4 (approved):

alan-agius4 (approved):
Thanks for your contribution.

dgp1130:

PR #27772, opened 2024-06-04
ci: use Github action-based dependency license checking

clydin:
Dependency know leverages the Github dependency review action instead of the previous custom solution. The action uses the Github dependency API to analyze dependencies. This not only should provide more accurate results but also lower the maintenance costs for the license checking. The initial allowed licenses list is based on the previous checker list
with licenses that are no longer used removed.

Action Documentation: https://github.com/actions/dependency-review-action


package.json
```diff
@@ -193,7 +192,6 @@
     "source-map": "0.7.4",
     "source-map-loader": "5.0.0",
     "source-map-support": "0.5.21",
-    "spdx-satisfies": "^5.0.0",
```
alan-agius4, line 196: Can you also remove the typings packages?

.github/dependency-review-config.yml
```diff
@@ -0,0 +1,12 @@
+vulnerability_check: false
+allow_licenses:
```
josephperrott, line 2: Can we preserve the commenting that we have in place previously? I think its helpful to see the categorization of the licenses

josephperrott (approved):
LGTM

Once we fully prove this out, we should centralize this in dev-infra and have it used everywhere, but thats a bigger follow up

PR #27798, opened 2024-06-07
fix(@angular/build): issue warning when auto adding `@angular/localize/init`

alan-agius4:
This commit introduces a warning for when the application builders automatically add the @angular/localize/init polyfill. The current approach has a drawback: the localize polyfill will always be included if it is found in a monorepo, even if an application does not use i18n.

To address this, we will issue a warning to inform users about this behavior and encourage them to explicitly add the polyfill to their polyfills configuration.

Additionally, this commit fixes an issue where the polyfill was not removed when using the build-time Angular i18n.


clydin (approved):

alan-agius4:

PR #27800, opened 2024-06-07
fix(@angular/build): generate module preloads next to script elements in index HTML

clydin:
The `modulepreload` link elements for initial scripts are now generated next to the actual script elements that require the referenced preloaded scripts. This better represents the fetch priorities to the browser and also allows easier visual discovery of the relevant script related elements inside the index HTML.


alan-agius4 (approved):

clydin:

PR #27807, opened 2024-06-10
fix(@angular/build): Initiate PostCSS only once

### LINKED ISSUE #27804: postcss in Angular 18+ version running twice

compiler-cli, compiler

### Is this a regression?

Yes

### Description

In Angular 18+ versions, the PostCSS configuration appears to execute twice during both development and production builds. This behavior can cause issues, particularly for users employing plugins such as postcss-rtlcss. In Angular 17 and earlier versions, this issue did not occur.

**Steps to Reproduce:**

1. Create a postcss.config.json file with the postcss-rtlcss plugin configured.
2. Run the Angular application in development mode or build it for production.

**Expected Behavior:**

PostCSS should execute once, correctly applying the postcss-rtlcss plugin and preserving specific CSS comments for RTL processing (e.g., /*!rtl:begin:ignore*/).

**Observed Behavior:**

PostCSS runs twice:

1. The first execution correctly applies the RTL CSS transformations and respects the ignore comments, producing CSS compatible with both LTR and RTL styles.
2. The second execution occurs after the comments have been removed, causing postcss-rtlcss to apply transformations again, leading to unintended CSS changes where comments were intended to prevent modifications.

**Impact:**

This issue causes CSS styles to be incorrectly processed, particularly in areas where ignore comments are used to maintain certain styles. The double processing results in unexpected changes to the CSS, breaking the intended styling for both LTR and RTL directions.

**Request:**

Please investigate and resolve the issue of PostCSS running twice in Angular 18+ to ensure consistent and expected CSS processing. This will help maintain compatibility with plugins like postcss-rtlcss and preserve the integrity of the styling configurations.

https://github.com/oarshad/angular-postcss-issue

_No response_

```true
Angular Cli: 18.0.3
```

### Anything else?

_No response_

alan-agius4:
Previously, PostCSS was initialized three times (once for each preprocessor), resulting in plugins being applied multiple times to each instance. This issue occured due to a race condition triggered by multiple esbuild plugins.

Fixes #27804

### DIFF
--- a/packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
+++ b/packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
@@ -96,67 +96,21 @@ export interface StylesheetLanguage {
 const postcssProcessors = new Map<string, WeakRef<PostcssProcessor>>();
 
 export class StylesheetPluginFactory {
-  private postcssProcessor?: PostcssProcessor;
-
   constructor(
     private readonly options: StylesheetPluginOptions,
     private readonly cache?: LoadResultCache,
   ) {}
 
   create(language: Readonly<StylesheetLanguage>): Plugin {
+    const { cache, options, setupPostcss } = this;
+
     // Return a noop plugin if no load actions are required
-    if (
-      !language.process &&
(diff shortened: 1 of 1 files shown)


clydin (approved):

clydin:

PR #27818, opened 2024-06-11
refactor: replace custom promise-based setTimeout with native Node.js setTimeout

alan-agius4:
In this commit, the custom implementation of a promise-based setTimeout function has been replaced with the native Node.js setTimeout function, which now returns a promise.


clydin (approved):

alan-agius4:

PR #27828, opened 2024-06-12
refactor(@angular/cli): remove dependency on `resolve`

alan-agius4:
Instead of using the `resolve` npm package, we avoid populating `Module._pathCache` when resolving package JSON files.

### DIFF
--- a/packages/angular/cli/src/utilities/package-tree.ts
+++ b/packages/angular/cli/src/utilities/package-tree.ts
@@ -6,9 +6,10 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import * as fs from 'fs';
-import { dirname, join } from 'path';
-import * as resolve from 'resolve';
+import assert from 'node:assert';
+import { readFile } from 'node:fs/promises';
+import Module, { createRequire } from 'node:module';
+import { dirname, join } from 'node:path';
 import { NgAddSaveDependency } from './package-metadata';
 
 interface PackageJson {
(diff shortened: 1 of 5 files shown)


packages/angular/cli/src/utilities/package-tree.ts
```diff
@@ -44,20 +44,25 @@ export interface PackageTreeNode {
...
+    require.cache = {};
+    const workspaceRequire = createRequire(`${workspaceDir}/`);
 
-    return packageJsonPath;
+    return workspaceRequire.resolve(`${packageName}/package.json`);
   } catch {
     return undefined;
+  } finally {
+    require.cache = originalRequireCache;
   }
```
clydin, line 66: Something similar to the following should work for resolve (`Module` imported from `node:module`):
```
  assert(
    '_pathCache' in Module && Module._pathCache && typeof Module._pathCache === 'object',
    'Node.js Module does not contain _pathCache',
  );

  // Avoid resolve cache problems, see: https://github.com/angular/angular-cli/pull/18610#issuecomment-681980185
  const originalPathCache = Module._pathCache;
  try {
    Module._pathCache = {};
    const workspaceRequire = createRequire(`${workspaceDir}/`);
    
    return workspaceRequire.resolve(`${packageName}/package.json`);
  } catch {
    return undefined;
  } finally {
    Module._pathCache = originalPathCache;
  }
```

alan-agius4: Looked a bit more and it looks like it will not solve https://github.com/angular/angular-cli/pull/18610#issuecomment-681980185

PR #27829, opened 2024-06-12
build: update E2E setup to use NPM 10 to publish packages

alan-agius4:
This commit updates the e2e setup to use the latest version of NPM to publish packages.

### DIFF
--- a/.github/dependency-review-config.yml
+++ b/.github/dependency-review-config.yml
@@ -2,6 +2,7 @@ vulnerability_check: false
 allow_licenses:
   - '0BSD'
   - 'Apache-2.0'
+  - 'Artistic-2.0'
   - 'BlueOak-1.0.0'
   - 'BSD-2-Clause'
   - 'BSD-3-Clause'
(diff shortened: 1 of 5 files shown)


.github/dependency-review-config.yml
```diff
@@ -1,6 +1,7 @@
 vulnerability_check: false
 allow_licenses:
   - '0BSD'
+  - 'Artistic-2.0'
   - 'Apache-2.0'
```
clydin, line 5:
```suggestion
  - 'Apache-2.0'
  - 'Artistic-2.0'
```

tests/legacy-cli/e2e/utils/registry.ts
```diff
@@ -17,16 +17,19 @@ export async function createNpmRegistry(
...
-    },
-  );
+  const verdaccioServer = fork(require.resolve('verdaccio/bin/verdaccio'), ['-c', configPath]);
+  await new Promise<void>((done) => {
+    verdaccioServer.on('message', (msg: { verdaccio_started: boolean }) => {
+      if (msg.verdaccio_started) {
+        done();
+      }
+    });
+  });
```
clydin, line 30: Should be able to do something like the following (`on` is from `node:events`):
```
for await (const event of on(verdacioServer, 'message')) {
    if (event.verdaccio_started) {
        break;
    }
}
```

Could even add a timeout as well: `{ signal: AbortSignal.timeout(30_000) }` for the third parameter of `on`

clydin (approved):

PR #27835, opened 2024-06-13
test: use Node test runner for E2E tests

dgp1130:
This updates the E2E test runner to invoke test programs with `node --test`. This allows us to use the `assert` API for expressive assertions and print failures with more details.

Unfortunately, it seems that the test runner and the test file are actually executed as two separate processes. This means that we run `node --test` for _each_ test and print out N reports each containing exactly 1 test. This isn't ideal, but it's good enough for using assertions. We can also explore using `node --test` for the test runner just to print a single test report with all the tests listed. This might require running test code in process with the test runner.

`node --test` is a little awkward in that it seems to treat all arguments as file paths to test, meaning there's no easy way to pass arguments into the tests. Since we're actually invoking our own test runner, we need to pass through the actual file being tested, but this is not easy to do with `node --test`. I feel like `node --test test_runner.js -- file_under_test.js` _should_ work, but actually doesn't. Instead, I opted to pass the file through as an environment variable. It's inelegant, but good enough for now.

I also had to separate out running a _test_ from running an arbitrary process, which were previously using the same code path and conflicting with each other.

Example output for a test failure: https://gist.github.com/dgp1130/e795d5d540ec98de7aac02a363f56998

### DIFF
--- a/tests/legacy-cli/e2e/tests/web-test-runner/basic.ts
+++ b/tests/legacy-cli/e2e/tests/web-test-runner/basic.ts
@@ -1,15 +1,11 @@
+import assert from 'node:assert';
 import { noSilentNg } from '../../utils/process';
 import { applyWtrBuilder } from '../../utils/web-test-runner';
 
 export default async function () {
-  // Temporary disable this test until we fix it.
-  return;
-
   await applyWtrBuilder();
 
   const { stderr } = await noSilentNg('test');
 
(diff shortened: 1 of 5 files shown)


dgp1130:
This doesn't really work easily unfortunately. Given enough time it could probably be made to work, but I don't think I have the mental capacity to push it through outside the fixit.

Issue #27833, opened 2024-06-12
Worker and global style files do not respect `sourceMap.hidden`


csvn:
### Command

build

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

When setting `build.options.sourceMap.hidden` to `true`, there is two cases I've seen where `//# sourceMappingURL=` is still added to the bottom of the file: any scripts files added as worker, and for style files specified in `build.options.styles`. I cannot find a way to make all emitted files respect the `hidden` flag for sourcemaps.

### Minimal Reproduction

1. Generate a new Angular project:
   ```bash
   npx @angular/cli@latest new --style=css --ssr=false ng-repro
   ```
2. Add `angular.json` config to use hidden sourcemaps:
   ```jsonc
   "build": {
     "builder": "@angular-devkit/build-angular:application",
       "options": {
         "sourceMap": {
           "hidden": true,
           "scripts": true,
           "styles": true
         },
         // ...
   ```
3. Add a worker file at `src/main.worker.ts`:
   ```ts
   globalThis.postMessage('Hello! I am a worker!');
   ```
4. Import the worker in `main.ts`, by adding this at the bottom:
   ```ts
   const worker = new Worker(new URL('./main.worker', import.meta.url));
   worker.addEventListener('message', e => console.log(e.data));
   ```
5. Run the build step:
   ```bash
   npm run build
   ```

By checking `styles-<hash>.css` and `worker-<hash>.js` in the `dist/<project>/browser` folder, we can see that at the bottom of both files there is a `//# sourceMappingURL=` mapping, and thus the sourcemap is not hidden.

### Your Environment

```text
Angular CLI: 18.0.3
Node: 22.2.0
Package Manager: npm 10.7.0
OS: win32 x64

Angular: 18.0.2
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1800.3
@angular-devkit/build-angular   18.0.3
@angular-devkit/core            18.0.3
@angular-devkit/schematics      18.0.3
@angular/cli                    18.0.3
@schematics/angular             18.0.3
rxjs                            7.8.1
typescript                      5.4.5
zone.js                         0.14.7
```

PR #27838, opened 2024-06-13
fix(@angular/build): do not reference sourcemaps in web workers and global stylesheet bundles when hidden setting is enabled

alan-agius4:
This commit addresses an issue where sourcemaps in web workers and global stylesheets were referenced despite the hidden setting being enabled.

Closes: #27833


clydin (approved):

csvn:
Awesome, that was really quick. Thanks!🎉

PR #27847, opened 2024-06-13
test: use the same process for the test runner and test files

dgp1130:
Instead of creating a subprocess to execute each test file, they are directly required into the main test runner. This allows us to use `node:test` to identify each test and create a single report at the end of execution.


tests/legacy-cli/e2e/tests/web-test-runner/basic.ts
```diff
@@ -1,12 +1,29 @@
+import assert from 'node:assert';
```
alan-agius4, line 1: Drive by comment, one reason why a new process is used is so that changes in `process.env` do not effect other tests.

dgp1130:
This doesn't really work easily (CI is lying and incorrectly passing everything). Given enough time it could probably be made to work, but I don't think I have the mental capacity to push it through outside the fixit.

cjihrig:
I'm not sure if it helps you, but I believe [Astro is running `node:test` in a single process](https://github.com/withastro/astro/blob/main/scripts/cmd/test.js). Maybe they have something you could use quickly without too much work.

PR #27868, opened 2024-06-17
ci: remove ci testing push trigger

josephperrott:
Remove ci testing pushing trigger as it no longer will work in the future and is unneccessary


clydin (approved):

alan-agius4:

PR #27875, opened 2024-06-18
fix(@angular/build): add CSP nonce to script with src tags

### LINKED ISSUE #27874: ngCspNonce should add the nonce value to the script tags loading the main bundles

compiler-cli

### Description

`ngCspNonce` is a great addition and almost solves a lot of the CSP difficulties I'm having but I think is missing some aspects to simplify the process of writing a CSP.

CSP level 3 introduces [strict-dynamic](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#strict-dynamic) for `script-src`. This avoids needing a whitelist and favours using a nonce (or hash) on scripts. The great part about it is that this "trust" given by a nonce also propagates to scripts that are loaded by the root script. This is particularly important for myself since we load a tracking script which in turn loads a bunch of other tracking services.

This _almost_ works. The problem is that you can't use `strict-dynamic` in conjunction with `'self'`. Angular currently does not add the nonce from `ngCspNonce` to the runtime/main bundles included at the bottom of the `index.html`. So, without a nonce or allowing "self" the browser blocks these.

image

### Proposed solution

Along with the other inline styles/scripts that `ngCspNonce` will add the nonce placeholder to, also add this nonce to the bundles.

### Alternatives considered

An alternate solution is to avoid `strict-dynamic` use whitelisting of domains but for an application with more complex scripts it is a lifesaver. [Google's CSP evaluator](https://csp-evaluator.withgoogle.com/) utilises `strict-dynamic` in their "sample safe policy".

alan-agius4:
Prior to this change, script tags with the `src` attribute were not being assigned a CSP nonce during the build process. This is useful strict-dynamic is a Content Security Policy (CSP) directive that simplifies the management of dynamically loaded scripts while maintaining a high level of security. It allows scripts that are initially trusted (through a nonce or hash) to load other scripts without additional restrictions.

Closes #27874

### DIFF
--- a/packages/angular/build/src/utils/index-file/nonce.ts
+++ b/packages/angular/build/src/utils/index-file/nonce.ts
@@ -29,8 +29,7 @@ export async function addNonce(html: string): Promise<string> {
 
   rewriter.on('startTag', (tag) => {
     if (
-      (tag.tagName === 'style' ||
-        (tag.tagName === 'script' && !tag.attrs.some((attr) => attr.name === 'src'))) &&
+      (tag.tagName === 'style' || tag.tagName === 'script') &&
       !tag.attrs.some((attr) => attr.name === 'nonce')
     ) {
       tag.attrs.push({ name: 'nonce', value: nonce });
--- a/packages/angular/build/src/utils/index-file/nonce_spec.ts
+++ b/packages/angular/build/src/utils/index-file/nonce_spec.ts
@@ -74,22 +74,22 @@ describe('addNonce', () => {
     expect(result).toContain('<style nonce="{% nonce %}">.a {color: red;}</style>');
   });
 
-  it('should to all inline script tags', async () => {
+  it('should to all script tags', async () => {
     const result = await addNonce(`
       <html>
         <head>
         </head>
         <body>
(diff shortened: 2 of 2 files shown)


clydin (approved):

packages/angular/build/src/utils/index-file/nonce.ts
sagartalaviya91, line 1: I want this change on Angular version 16. I can't upgrade to latest version now. Is there any way to do this fix in Version 16?

sagartalaviya91:
I want this change on Angular version 16. I can't upgrade to latest version now. Is there any way to do this fix in Version 16?

packages/angular/build/src/utils/index-file/nonce.ts
alan-agius4: Version 16 is in LTS phase and only received security fixes.

For all other fixes please update to the latest version.

Please see: https://angular.dev/reference/releases#actively-supported-versions

sagartalaviya91: In which version this fix will be available?

PR #27876, opened 2024-06-18
fix(@angular/build): use Node.js available parallelism for default worker count

clydin:
The Node.js `os.availableParallelism` API now provides more accurate values when used within containers such as many CI environments. This provides a better default when using resource limited CI setups. The value is still set to a maximum default of four. However, the `NG_BUILD_MAX_WORKERS` environment variable can be used to set the value to an explicit value if required.


alan-agius4 (approved):

alan-agius4:

PR #27891, opened 2024-06-20
refactor(@angular/build): include Angular compiler plugin in private API

clydin:
Add `createCompilerPlugin` function to the private export of the package. Note that these are not considered part of the public API and are intended for use only with the `@angular-devkit/build-angular` package.


alan-agius4 (approved):

alan-agius4:

PR #27894, opened 2024-06-21
build: update to TypeScript 5.5 stable

crisbeto:
Updates the repo to the stable version of TypeScript 5.5. I'm holding off on updating the new projects to it due to https://github.com/angular/angular/pull/56358.

### DIFF
--- a/package.json
+++ b/package.json
@@ -197,7 +197,7 @@
     "tree-kill": "1.2.2",
     "ts-node": "^10.9.1",
     "tslib": "2.6.3",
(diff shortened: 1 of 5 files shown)


alan-agius4 (approved):

crisbeto:
I mentioned it in the commit message: I'll hold off on switching new projects to it for now until we can land https://github.com/angular/angular/pull/56358. Without it there's an edge case where users can see a very cryptic TS error.

PR #27900, opened 2024-06-21
fix(@angular/build): show JavaScript cache store initialization warning

clydin:
If the persistent cache store for the JavaScript transformations fails to initialize, a warning will now be shown to better explain the outcome and to allow the build to continue. The build will still complete without the cache but may be slower to finish.


alan-agius4 (approved):

clydin:

PR #27944, opened 2024-06-26
refactor(@angular/build): improve BuildOutputFile property access

clydin:
The `BuildOutputFile` type's helper functions have been adjusted to cache commonly accessed property values to avoid potentially expensive repeat processing. This includes encoding/decoding UTF-8 content and calculating hash values for the output file content. A size property has also been added to allow consumers to more directly determine the byte size of the output file. The size property is currently unused but will be leveraged in forthcoming updates to bundle budgets and console info logging.


alan-agius4 (approved):

clydin:

PR #27947, opened 2024-06-27
fix(@angular/build): correctly name entry points to match budgets

alan-agius4:
This commit addresses an issue where some lazy entry points were not name correctly to align with specified budgets.

Closes: #27936

### DIFF
--- a/packages/angular/build/src/builders/application/tests/options/bundle-budgets_spec.ts
+++ b/packages/angular/build/src/builders/application/tests/options/bundle-budgets_spec.ts
@@ -93,6 +93,39 @@ describeBuilder(buildApplication, APPLICATION_BUILDER_INFO, (harness) => {
       );
     });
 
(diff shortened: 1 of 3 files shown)


clydin (approved):

packages/angular/build/src/builders/application/tests/options/bundle-budgets_spec.ts
```diff
@@ -93,6 +93,39 @@ describeBuilder(buildApplication, APPLICATION_BUILDER_INFO, (harness) => {
       );
     });
 
+    it(`should not warn when none injected style is not within the baseline threshold`, async () => {
```
clydin, line 96:
```suggestion
    it(`should not warn when non-injected style is not within the baseline threshold`, async () => {
```

alan-agius4:

PR #27950, opened 2024-06-27
refactor(@angular/build): use build output files directly in stats and budgets

clydin:
The bundle budget calculators and the console build stats output are now calculated directly from the build output file information instead of the esbuild metafile where possible. This provides a more generic method of accessing the information and can more accurately account for post-processing steps that may alter the output files. The metafile is still used for component style budgets and lazy chunk name information.


alan-agius4 (approved):

clydin:

Issue #27715, opened 2024-05-25
Creating unnecessary excessive chunks


ChristopherPHolder:
### Command

build

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

**Disclaimer**

This is not really a bug, as even if it worked differently with `webpack`, for `esbuild` this is normal. However, since the default bundler is now `esbuild` some would consider this a regression.

## Required chunks references in dynamic imported code get split in to separate chunks

When splitting code, esbuild does not seems to properly take into account what is already statically imported (thus "required") and does not need to be split into a separate chunk.

To illustrate this lets consider an example where we have the following 5 standalone components:

- `AppComponent` (The root component which is not lazy loaded)
- `Feature1Component` (A feature component lazy loaded via a dynamic import in the router)
- `Feature21Component` (A feature component lazy loaded via a dynamic import in the router)
- `Ui1Component` (A UI component imported statically, like a button)
- `Ui2Component` (A UI component imported statically, like a button)

### Chunks without Shared Components

If the `AppComponent` has a `RouterOutlet` which has separate paths `Feature1Component` and `Feature21Component` it will generate a chunk for the code shared between these components (lets call it Common Chunk). 

In this case the build output is: 

| Initial chunk files             | Names               | 
| --- | --- | 
| chunk-YIKH6UD4.js               | -  ("common")  | 
| main-V6UA4PHT.js                | main                |

| Lazy chunk files                | Names               |
| --- | --- | 
| feature-1.component-AQEBZKLJ.js | feature-1-component |
| feature-2.component-EH7EGWEP.js | feature-2-component |

And if `Feature1Component` imports  `Ui1Component` and  `Feature21Component` imports  `Ui2Component` then code from `Ui1Component` will be bundle in `Feature1Component` and the code from `Ui2Component` will be bundle in `Feature2Component`.

### Chunks with Shared Components

However, if the `AppComponent` also imports  `Ui1Component` and  `Ui2Component` this code is now shared and will be split into separate chunks.

Thus the build output will now be:

| Initial chunk files             | Names               | 
| --- | --- | 
| chunk-YIKH6UD4.js               | -  ("common")  | 
| main-V6UA4PHT.js                | main                |
| chunk-WKA4CZR7.js               | -  ("ui-1-component") |
| chunk-XCJUHD2B.js               | - ("ui-2-component") |

| Lazy chunk files                | Names               |
| --- | --- | 
[...]

PR #27953, opened 2024-06-27
refactor(@angular/build): add experimental chunk optimizer for production application builds

clydin:
An experimental chunk optimizer is now available for initial usage. To enable the optimization, script optimization must be enabled as well as an environment variable `NG_BUILD_OPTIMIZE_CHUNKS=1`. This build step uses `rollup` internally to process the build files directly in memory. The main bundling performs all resolution, bundling, and tree-shaking of the application. The chunk optimizer step then only needs to access the in-memory built files and does not need to perform any disk access or module resolution. This allows the step to be performed fairly quickly but it does add time to the overall production build. The `NG_BUILD_DEBUG_PERF=1` environment variable can be used to view how long the step takes within a build via the `OPTIMIZE_CHUNKS` entry. In the future, this optimization step may be automatically enabled based on initial file entry count and size. There are several current known issues:
1) Bundle budgets for named lazy chunks may not work as expected.
2) The console output may not show names (files will be present) for lazy chunk files.
3) The stats file (`--stats-json` option) will not exactly reflect the final written application files. This is similar to the current behavior of the `browser` builder with Webpack's stat file.

### DIFF
--- a/tests/legacy-cli/e2e/tests/build/chunk-optimizer.ts
+++ b/tests/legacy-cli/e2e/tests/build/chunk-optimizer.ts
@@ -1,0 +1,19 @@
+import assert from 'node:assert/strict';
+import { readFile } from 'node:fs/promises';
+import { execWithEnv } from '../../utils/process';
+
+/**
+ * AOT builds with chunk optimizer should contain generated component definitions.
+ * This is currently testing that the generated code is propagating through the
+ * chunk optimization step.
+ */
+export default async function () {
+  await execWithEnv('ng', ['build', '--output-hashing=none'], {
+    ...process.env,
+    NG_BUILD_OPTIMIZE_CHUNKS: '1',
+    NG_BUILD_MANGLE: '0',
+  });
+
+  const content = await readFile('dist/test-project/browser/main.js', 'utf-8');
+  assert.match(content, /\\u0275\\u0275defineComponent/);
+}
--- a/package.json
+++ b/package.json
@@ -180,7 +180,7 @@
     "puppeteer": "18.2.1",
     "quicktype-core": "23.0.170",
     "resolve-url-loader": "5.0.0",
-    "rollup": "~4.18.0",
+    "rollup": "4.18.0",
     "rollup-plugin-sourcemaps": "^0.6.0",
     "rxjs": "7.8.1",
     "sass": "1.77.6",
--- a/packages/angular/build/BUILD.bazel
+++ b/packages/angular/build/BUILD.bazel
@@ -89,6 +89,7 @@ ts_library(
         "@npm//picomatch",
         "@npm//piscina",
         "@npm//postcss",
+        "@npm//rollup",
         "@npm//sass",
         "@npm//semver",
         "@npm//tslib",
--- a/packages/angular/build/package.json
+++ b/packages/angular/build/package.json
@@ -38,6 +38,7 @@
     "parse5-html-rewriting-stream": "7.0.0",
     "picomatch": "4.0.2",
     "piscina": "4.6.1",
+    "rollup": "4.18.0",
     "sass": "1.77.6",
     "semver": "7.6.2",
     "undici": "6.19.2",
--- a/packages/angular/build/src/builders/application/chunk-optimizer.ts
+++ b/packages/angular/build/src/builders/application/chunk-optimizer.ts
@@ -1,0 +1,211 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import assert from 'node:assert';
+import { rollup } from 'rollup';
+import {
+  BuildOutputFile,
+  BuildOutputFileType,
+  BundleContextResult,
+  InitialFileRecord,
+} from '../../tools/esbuild/bundler-context';
+import { createOutputFile } from '../../tools/esbuild/utils';
+import { assertIsError } from '../../utils/error';
+
+export async function optimizeChunks(
+  original: BundleContextResult,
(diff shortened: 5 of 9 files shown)


ChristopherPHolder:
https://github.com/angular/angular-cli/issues/27715

Would love to help with this issue, I have a functioning implementation on a large project and would love to cooperate.

alan-agius4 (approved):
NIT: can you pin https://github.com/clydin/angular-cli/blob/8e7bb8342a130b9460fd27493d2de825d76b437a/package.json#L183?

tests/legacy-cli/e2e/tests/build/chunk-optimizer.ts
```diff
@@ -0,0 +1,16 @@
+import assert from 'node:assert/strict';
+import { readFile } from 'node:fs/promises';
+import { execWithEnv } from '../../utils/process';
+
+/**
+ * AOT builds with chunk optimizer should contain generated component factories
+ */
+export default async function () {
+  await execWithEnv('ng', ['build', '--aot=true', '--configuration=development'], {
```
alan-agius4, line 9: Isn't chunk optimizer disabled in this? Since script optimisations are disabled during development builds?

https://github.com/angular/angular-cli/pull/27953/files#diff-725bf17eb0a0bed74c9aaca5f4c4ff2b2239a664a4baac4b29a43d961069adf7R70

clydin:

clydin:
@ChristopherPHolder Thank you for offering to help.  This will be available for use in the release next week.  If you (or anyone else interested) have the time, please try it with your application and let us know how well it works for your project.  If you are interested in trying it out sooner, there are also snapshot builds available:
If using `@angular/build:application`, `github:angular/angular-build-builds#8d314a1`.
If using `@angular-devkit/build-angular:application`, `github:angular/angular-devkit-build-angular-builds#8d314a1`.
Depending on the package manager used, you may need to use the `--force` option with the snapshot builds.

Please provide feedback in the mentioned issue (https://github.com/angular/angular-cli/issues/27715).

ChristopherPHolder:
Absolutely, thanks, will try it out and update the issue. 

Are you planing on specifying the number of initial chunks? I ask because of the environmental variable. 

We are if not this means we end up with one large chunk. 

And when using manual chunks rollup does not take into account circular dependencies which means the strategy you use needs to prevent that.

Ideally we could choose the number of maximum chunks and try to distribute the bundle in the rest as evenly as possible.

dimeloper:
@ChristopherPHolder did you manage to try it out? if yes did it work for you? 

We are facing the same issue at our application (excessive amount of chunks), and I tried the rollup workaround mentioned in the #27715 issue, however this caused some unexpected side effects to our app and we had to revert it.

ChristopherPHolder:
@dimeloper no, not yet but i have a similar implementation in a mono repo filled with enterprise apps, so i know the general solution works. 

> We are facing the same issue at our application (excessive amount of chunks), and I tried the rollup workaround mentioned in the #27715 issue, however this caused some unexpected side effects to our app and we had to revert it.

When we implemented it, we also faced a bunch of issues, circular deps, broken sourcemaps, invalid import transformations, etc.

What is the issue you are facing? Maybe i got a solution for you? I will also happily publish my solution but since @clydin is working on it internally and that would allow us to handle the build transformation in memory, that seems like a nicer solution

dimeloper:
@ChristopherPHolder while the esbuild production app runs fine, after running the rollup command to reduce the amount of chunks, the generated app throws error like the following in random places, that prevent specific components from being loaded:

```
Error: no ngModuleRef constructor found
    at chunk-UKEZH6HP-Djqkv2-u.js:246:4618
    at Generator.next (<anonymous>)
    at n (chunk-UKEZH6HP-Djqkv2-u.js:1:1248)
    at k.invoke (zone.js:368:26)
    at Object.onInvoke (chunk-UKEZH6HP-Djqkv2-u.js:9:52548)
    at k.invoke (zone.js:367:52)
    at se.run (zone.js:130:43)
    at zone.js:1260:36
    at k.invokeTask (zone.js:403:31)
    at Object.onInvokeTask (chunk-UKEZH6HP-Djqkv2-u.js:9:52359)
    at k.invokeTask (zone.js:402:36)
    at se.runTask (zone.js:174:47)
    at E (zone.js:582:35)
```

ChristopherPHolder:
@dimeloper That issue was mostly related to circular deps in out case. 

Did you simply run the rollup cli on it? or did you add minChunkSize ?

Unfortunately minChunkSize will not work or at lease did not for us

dimeloper:
@ChristopherPHolder thank you for your input. Yes I tried this with `minChunkSize` since we had to reduce the amount of generated chunks. May I ask how did you made it work without the `minChunkSize` option?

dimeloper:
@clydin I have one question, so today I decided to upgrade our project to `18.1-rc` and now I am trying to use the optimisation you introduced without success, even though I set the following environment variable, as per your instructions: `NG_BUILD_OPTIMIZE_CHUNKS=1`.

We are on an NX project though (version `19.5.0-beta.0`), and therefore we use the `@nx/angular:application` builder. I suppose your change is not available there (yet)?

dimeloper:
@clydin not sure if you are still following this issue, today I managed to define the necessary environment variable (btw if anyone is using NX, [here](https://nx.dev/recipes/angular/use-environment-variables-in-angular) is how you can do it), however I'm still getting the same (excessive) amount of chunks (I also debugged the chunk optimizer and went through some breakpoints to make sure it was being called while it was building). 

- Before switching to the application builder:  326 chunks in total
- Application builder, before enabling the chunk optimizer:  510 chunks in total
- Application builder, after enabling the chunk optimizer:  510 chunks in total

My guess would be that we would need to inject some rollup parameters to that `optimizeChunks` method for our case. If you have any further guidance or if you think I'm missing anything, let me know.

PR #27958, opened 2024-06-28
fix(@angular/build): redirect to path with trailing slash for asset directories

### LINKED ISSUE #27949: serve: serving static files using vite differs from the previous builder
### Command

serve

### Is this a regression?

_No response_

### Description

We are migrating our project to Angular's new build system. It is the preferred option for new projects and we think migrating to it is a good idea. But we have found a small issue with how the static assets are served compared to the previous `browser` builder.

Our application requires a bunch of static files to work, and one of the html files is accessed using the parent folder name without a trailing slash: `localhost:4200/oauth2redirect`. Using the previous builder this was loading the `oauth2redirect/index.html` file.

After migrating to the new `application` builder, this same URI throws the following error:
```
Error: NG04002: Cannot match any routes. URL Segment: 'oauth2redirect'
```

This can be fixed by adding a trailing slash to the URI (`localhost:4200/oauth2redirect/`). But this would require updating our OAuth redirect uri. And since this is something already supported by other tools and platforms (Google Cloud Buckets for example), I think it may be a good idea to support it here as well.

### Minimal Reproduction

Add some static `index.html` to an angular project:

```
src/
  app/...
  static/
    oauth2redirect/
      index.html
```

```jsonc
// angular.json
"assets": [
  {
    "glob": "**/*",
    "input": "src/static",
    "output": "/"
  }
]
```

Serve the app
```sh
ng serve
```

And try to load the file by using its parent folder.

All three URIs work with previous `browser` builder. And the first one fails with the new `application`  builder.
```sh
localhost:4200/oauth2redirect # => error
localhost:4200/oauth2redirect/ # => loads the file
localhost:4200/oauth2redirect/index.hmtl # => loads the file
```

### Exception or Error

_No response_

### Your Environment

```text
Angular CLI: 17.3.8
Node: 18.20.3
Package Manager: npm 10.8.1
OS: darwin arm64

Angular: 17.3.11
... language-service, localize, platform-browser

@angular-devkit/architect       0.1703.8
@angular-devkit/build-angular   17.3.8
@angular-devkit/core            17.3.8
@angular-devkit/schematics      17.3.8
@angular/cdk                    17.3.10
@angular/cli                    17.3.8
@angular/material               17.3.10
@schematics/angular             17.3.8
typescript                      5.3.3
zone.js                         0.14.4
```

### Anything else relevant?

_No response_

alan-agius4:
Prior to this commit, accessing a static asset directory without a trailing slash resulted in a 404 error. With this change, we now redirect to the path with a trailing slash, aligning with the behavior of express static.

Closes #27949

### DIFF
--- a/packages/angular/build/src/builders/dev-server/tests/behavior/build-assets_spec.ts
+++ b/packages/angular/build/src/builders/dev-server/tests/behavior/build-assets_spec.ts
@@ -11,7 +11,7 @@ import { executeOnceAndFetch } from '../execute-fetch';
 import { describeServeBuilder } from '../jasmine-helpers';
 import { BASE_OPTIONS, DEV_SERVER_BUILDER_INFO } from '../setup';
 
-describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupTarget, isVite) => {
+describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupTarget) => {
   const javascriptFileContent =
     "import {foo} from 'unresolved'; /* a comment */const foo = `bar`;\n\n\n";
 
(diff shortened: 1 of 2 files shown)


clydin (approved):

alan-agius4:

PR #27961, opened 2024-06-28
refactor(@angular/build): move Vite middlewares into separate files

alan-agius4:
As the number of middlewares has increased over time, this commit enhances code health by relocating them into individual files.

### DIFF
--- a/packages/angular/build/src/tools/vite/angular-memory-plugin.ts
+++ b/packages/angular/build/src/tools/vite/angular-memory-plugin.ts
@@ -7,28 +7,30 @@
  */
 
 import remapping, { SourceMapInput } from '@ampproject/remapping';
-import { lookup as lookupMimeType } from 'mrmime';
 import assert from 'node:assert';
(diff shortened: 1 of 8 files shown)


packages/angular/build/src/tools/vite/angular-memory-plugin.ts
```diff
@@ -7,28 +7,31 @@
...
 import type { Connect, Plugin } from 'vite';
-import { renderPage } from '../../utils/server-rendering/render-page';
+import {
+  angularHtmlFallbackMiddleware,
+  createAngularAssetsMiddleware,
+  createAngularIndexHtmlMiddleware,
+  createAngularSSRMiddleware,
+} from './middlewares';
+
+export type AngularMemoryOutputFiles = Map<string, { contents: Uint8Array; servable: boolean }>;
```
clydin, line 21: nit: Moving this type to `utils.ts` should remove the circular references

clydin (approved):

Issue #27167, opened 2024-02-27
building failed sometimes with Can't find stylesheet to import


robertIsaac:
### Command

build

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

about 5% of our pipeline fail with this error
```
✘ [ERROR] Can't find stylesheet to import.
   ╷
27 │ @use '@material/ripple/ripple';
   │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
  node_modules/@material/button/_button-ripple.scss 27:1                             @use
  node_modules/@material/button/_button-shared-theme.scss 42:1                       @use
  node_modules/@material/button/_button-base.scss 35:1                               @use
  @material/button/_button.scss 30:1                                                 @use
  node_modules/@angular/material/button/_button-theme.scss 2:1                       @use
  node_modules/@angular/material/core/density/private/_all-density.scss 2:1          @forward
  @angular/_index.scss 20:1                                                          @use
  packages/web/src/theme/palette.scss 3:1                                            @import
  theme.scss 1:9                                                                     @import
  packages/web/src/app/grid-layout/grid-controls/scroll-controls.component.scss 1:9  root stylesheet [plugin angular-sass]
```

it happened only one with `ng serve` and it was very difficult to resolve, my colleague had to uninstall node and install it again
previously he tried to delete the project and clone it again, remove yarn cache but neither work

### Minimal Reproduction

it's impossible to reproduce, since even rerunning the same one succeed
![image](https://github.com/angular/angular-cli/assets/14242054/6bd82111-74f1-45b2-acaf-550837c2b52f)
it's not code error

this happens even when we were using webpack (we are using the application builder which uses esbuild), I noticed it in versions 15 and 17 of Angular

### Exception or Error

```text
2024-02-27T12:22:15.0367412Z - Building...
2024-02-27T12:24:41.4014964Z [31m✘ [41;31m[[41;97mERROR[41;31m][0m [1mCan't find stylesheet to import.
2024-02-27T12:24:41.4015914Z    ╷
2024-02-27T12:24:41.4016209Z 27 │ @use '@material/ripple/ripple';
2024-02-27T12:24:41.4016714Z    │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-02-27T12:24:41.4016908Z    ╵
2024-02-27T12:24:41.4017254Z   node_modules/@material/button/_button-ripple.scss 27:1                             @use
[...]

PR #27969, opened 2024-07-01
fix(@angular/build): reduce the number of max workers to available CPUs minus one

alan-agius4:
This commit reduces the maximum number of workers to the available CPUs minus 1. This adjustment ensures that some resources are left for the main thread, preventing it from being starved of CPU cycles.

For more context see: https://github.com/angular/angular-cli/issues/27167#issuecomment-2199987874


clydin (approved):

alan-agius4:

PR #27972, opened 2024-07-01
fix(@angular/build): allow top-level await in zoneless applications

clydin:
With the potential future introduction of features such as WASM/ES module integration and the experimental chunk optimizer, the need to restrict top-level await usage is no longer needed and a blocker for new features. As such, top-level await will now be available if an application has been configured to be zoneless.
Zoneless Angular is currently experimental and more details can be found here: https://angular.dev/guide/experimental/zoneless


alan-agius4 (approved):

clydin:

PR #27973, opened 2024-07-02
feat(@angular/build): support WASM/ES Module integration proposal

clydin:
Application builds will now support the direct import of WASM files. The behavior follows the WebAssembly/ES module integration proposal. The usage of this feature requires the ability to use native async/await and top-level await. Due to this requirement, applications must be zoneless to use this new feature. Applications that use Zone.js are currently incompatible and an error will be generated if the feature is used in a Zone.js application. Manual setup of a WASM file is, however, possible in a Zone.js application if WASM usage is required. Further details for manual setup can be found here: https://developer.mozilla.org/en-US/docs/WebAssembly/Loading_and_running

The following is a brief example of using a WASM file in the new feature with the integration proposal behavior:
```
import { multiply } from './example.wasm';

console.log(multiply(4, 5));
```

NOTE: TypeScript will not automatically understand the types for WASM files. Type definition files will need to be created for each WASM file to allow for an error-free build. These type definition files are specific to each individual WASM file and will either need to be manually created or provided by library authors.

The feature relies on an active proposal which may change as it progresses through the standardization process. This may result in behavioral differences between versions.
Proposal Details: https://github.com/WebAssembly/esm-integration/tree/main/proposals/esm-integration

For more information regarding zoneless applications, you can visit https://angular.dev/guide/experimental/zoneless

### DIFF
--- a/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
+++ b/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
@@ -23,6 +23,7 @@ import { createRxjsEsmResolutionPlugin } from './rxjs-esm-resolution-plugin';
 import { createSourcemapIgnorelistPlugin } from './sourcemap-ignorelist-plugin';
 import { getFeatureSupport, isZonelessApp } from './utils';
 import { createVirtualModulePlugin } from './virtual-module-plugin';
+import { createWasmPlugin } from './wasm-plugin';
 
 export function createBrowserCodeBundleOptions(
   options: NormalizedApplicationBuildOptions,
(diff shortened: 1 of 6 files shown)


packages/angular/build/src/tools/esbuild/application-code-bundle.ts
```diff
@@ -48,8 +51,9 @@ export function createBrowserCodeBundleOptions(
     entryNames: outputNames.bundles,
     entryPoints,
     target,
-    supported: getFeatureSupport(target, isZonelessApp(polyfills)),
+    supported: getFeatureSupport(target, zoneless),
     plugins: [
+      createWasmPlugin({ allowAsync: zoneless, cache: sourceFileCache?.loadResultCache }),
```
alan-agius4, line 56: Shouldn't this be also done in the `createServerCodeBundle` it believe it will work for Node.js

alan-agius4 (approved):

clydin:

PR #27977, opened 2024-07-02
fix(@angular-devkit/build-angular): rollback terser to `5.29.2`

### LINKED ISSUE #27866: bug(Ripple): error "Cannot set properties of undefined (setting 'FADING_IN')"
### Is this a regression?

17.3

### Description

An error is thrown for production builds when following conditions are met:

1. The angular appplication makes uses of material (for example MatDialog) but does not make use of MatButton (or MatRipple directly)
2. The webpack builder is used (`@angular-devkit/build-angular:browser`)
3. It happens only when building in production mode (not in development mode)

The following error is thrown in JS console, and the app is broken:
``` bash
TypeError: Cannot set properties of undefined (setting 'FADING_IN')
```

Note: since I could not reproduce this with the esbuild builder (`@angular-devkit/build-angular:application`), this could very well be an issue related to the angular-cli instead of material itself. Since I am not 100% sure, I am reporting it here.

### Reproduction

StackBlitz link: https://stackblitz.com/edit/sjwtcm-g8jwcl?file=src%2Fexample%2Fdialog-data-example.ts
Steps to reproduce:
1. Open the StackBlitz link
2. Check JS console to see the error

### Expected Behavior

No errors, app works.

### Actual Behavior

App is broken, an error is visible in the JS console:
``` bash
TypeError: Cannot set properties of undefined (setting 'FADING_IN')
```

### Environment

- Angular: 18.0.3
- CDK/Material: 18.0.3
- Browser(s): Google Chrome
- Operating System (e.g. Windows, macOS, Ubuntu): Windows

alan-agius4:
More information in https://github.com/terser/terser/issues/1539

Closes #27866

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,7 +192,7 @@
     "source-map-support": "0.5.21",
     "symbol-observable": "4.0.0",
     "tar": "^6.1.6",
-    "terser": "5.31.1",
+    "terser": "5.29.2",
     "tree-kill": "1.2.2",
     "ts-node": "^10.9.1",
     "tslib": "2.6.3",
--- a/packages/angular_devkit/build_angular/package.json
+++ b/packages/angular_devkit/build_angular/package.json
@@ -58,7 +58,7 @@
     "semver": "7.6.2",
     "source-map-loader": "5.0.0",
     "source-map-support": "0.5.21",
-    "terser": "5.31.1",
+    "terser": "5.29.2",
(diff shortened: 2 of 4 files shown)


alan-agius4:

ahaller-billtrust:
Hi @alan-agius4 sorry to bother. Do you know when we can expect this fix to be released?

PR #27992, opened 2024-07-03
fix(@angular/build): check inlineSourceMap option with isolated modules optimization

clydin:
The isolated modules transpilation option that was recently introduced is currently disabled when source maps are enabled. TypeScript has two mutually exclusive options that can be used to check for source map usage. Both of these options (`sourceMap` and `inlineSourceMap`) are now checked to determine if the direct transpilation optimization can be used.


alan-agius4 (approved):

clydin:

PR #27993, opened 2024-07-03
refactor(@angular/cli): provide a default extract-i18n target for applications

clydin:
The `extract-i18n` command will now use a default project target and builder name if the target entry is not explicitly defined. This allows the removal of additional configuration from an `angular.json` file. If the target is already present than it will take priority over any default builder behavior. Infrastructure is also now present to allow other architect commands to have similar functionality in the future.


alan-agius4 (approved):

alan-agius4:

PR #28011, opened 2024-07-08
refactor(@angular/cli): provide default serve target for applications

clydin:
The `serve` command will now use a default project target and builder name if the target entry is not explicitly defined. This allows the removal of additional configuration from an `angular.json` file. If the target is already present than it will take priority over any default builder behavior. The default logic will use the appropriate development server builder for officially supported packages (`@angular/build`/`@angular-devkit/build-angular`). The `dev-server` builder from these packages will currently assume a `development` configuration is present within the `build` target. The default behavior may be expanded in the future to support more arbitrary `build` target configurations. If there is third-party package usage within the `build` target, the CLI will attempt to discover a `dev-server` builder within the same package used by the `build` target. If none is found, no default will be added and the `serve` command will issue an error when no explicit target is present.

### DIFF
--- a/goldens/public-api/angular/build/index.api.md
+++ b/goldens/public-api/angular/build/index.api.md
@@ -109,7 +109,7 @@ export enum BuildOutputFileType {
 // @public
 export type DevServerBuilderOptions = {
     allowedHosts?: AllowedHosts;
-    buildTarget: string;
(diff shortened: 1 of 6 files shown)


zygarios:
What about this PR? It's cool, and i hope it will be released with 19 :(

PR #28012, opened 2024-07-08
fix(@angular/build): read WASM file from script location on Node.js

clydin:
When using a WASM file on Node.js via SSR/SSG/etc. the path for the `readFile` call will now be based on the location of the script using the WASM file instead of the current working directory.
This change also adds a general Node.js WASM E2E test via prerendering.


alan-agius4 (approved):

alan-agius4:

PR #28029, opened 2024-07-10
refactor(@angular/build): use `listr2` for spinner and color support

clydin:
The `ansi-color` and `ora` dependencies have been replaced with `listr2`. This package provides both color and spinner capabilities and further reduces the dependency count. This also aligns the dependencies with the `@angular/cli` package with already uses `listr2`. The spinner also now will not overwrite console output that happens to be written while the spinner is active. Instead, the output will be written after the spinner task is complete.


alan-agius4 (approved):
LGTM, do you also plan to make this change in build-angular?

Issue #26575, opened 2023-12-04
Raw import of .ts files as text with `build-angular:application`


santam85:
### Command

build, serve

### Description

A similar issue (https://github.com/angular/angular-cli/issues/26368) has been opened and resolved already with the `loader` config options for custom extensions/types.

Unfortunately the implemented solution still lacks the ability to perform raw imports of `.ts` files. This is vital for some use-cases where documenting the examples shown in an application (IE: technical docs for a library) is achieved by including the plain source files themselves as part of the text content rendered in the app.

### Describe the solution you'd like

The loader prefix used in webpack to force a particular behaviour was quite a neat solution. Anything done when defining the import file path (ie: `import textContent from 'raw:./file.ts'`) is preferable over a generic project metadata config, imho.

### Describe alternatives you've considered

The only possible solution at the moment seems to maintain a copy of the source file with a `.txt` extension.

mb21:
An alternative syntax that [vite is using](https://vitejs.dev/guide/assets#importing-asset-as-string) would be:

    import textContent from './file?raw'

mb21:
@clydin that TypeScript issue you linked to talks about `with { type: "css" }` which would import some CSS and assign it type `CSSStyleSheet`. As I understood the OP, this issue is about importing a file's contents and simply assign it type `string`?

clydin:
That is one specific (and common) use case involving the need for TypeScript to provide additional syntax to allow import attribute aware type definitions.

alan-agius4:
Thanks for reporting this issue. Luckily, it has already been fixed in one of the recent releases. Please update to the most recent version to resolve the problem.

If the problem persists in your application after upgrading, please open a new issue, provide a simple repository reproducing the problem, and describe the difference between the expected and current behavior. You can use `ng new repro-app` to create a new project where you reproduce the problem.

santam85:
@alan-agius4 thanks for looping back on this. Can you provide some reference link on which of the 3 options discussed in the issue is now supported and from which version this became available?

alan-agius4:
In Angular CLI 20.x, the standard based approach using [import attributes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with) (i.e. `import textContent from './file' with { loader: 'text' };`).  This is now possible without using `ts-ignore`, thanks to TypeScript's new import attribute-aware type definitions.

PR #28040, opened 2024-07-11
feat(@angular/build): support import attribute based loader configuration

clydin:
When using the `application` builder, a `loader` import attribute is now available for use with import statements and expressions. The presence of the import attribute takes precedence over all other loading behavior including JS/TS and any `loader` build option values. This allows per file control over loading behavior. For general loading for all files of an otherwise unsupported file type, the `loader` build option is recommended.

For the import attribute, the following loader values are supported:
* `text` - inlines the content as a string
* `binary` - inlines the content as a Uint8Array
* `file` - emits the file and provides the runtime location of the file

Unfortunately, at this time, TypeScript does not support type definitions that are based on import attribute values. The use of `@ts-expect-error` or the use of individual type definition files (assuming the file is only imported with the same loader attribute) is currently required.

Additionally, the TypeScript `module` option must be set to `esnext` to allow TypeScript to successfully build the application code.

As an example, an SVG file can be imported as text via:
```
// @ts-expect-error TypeScript cannot provide types based on attributes yet
import contents from './some-file.svg' with { loader: 'text' };
```

When using the development server and a file that is referenced from a Node.js package with a loader attribute, the package must be excluded from prebundling via the development server `prebundle` option. This does not apply to relative file references.

This feature provides the bundling aspects of #26575. TypeScript type definition support would provide the full feature set for that issue but is currently blocked on TypeScript upstream.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/loader-import-attribute-plugin.ts
+++ b/packages/angular/build/src/tools/esbuild/loader-import-attribute-plugin.ts
@@ -1,0 +1,44 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import type { Loader, Plugin } from 'esbuild';
+import { readFile } from 'node:fs/promises';
+
+const SUPPORTED_LOADERS: Loader[] = ['binary', 'file', 'text'];
+
+export function createLoaderImportAttributePlugin(): Plugin {
+  return {
+    name: 'angular-loader-import-attributes',
+    setup(build) {
+      build.onLoad({ filter: /./ }, async (args) => {
+        const loader = args.with['loader'] as Loader | undefined;
+        if (!loader) {
+          return undefined;
+        }
+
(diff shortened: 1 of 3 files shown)


packages/angular/build/src/tools/esbuild/loader-import-attribute-plugin.ts
```diff
@@ -0,0 +1,44 @@
...
+import { readFile } from 'node:fs/promises';
+
+const SUPPORTED_LOADERS: Loader[] = ['binary', 'file', 'text'];
+
+export function createLoaderImportAttributePlugin(): Plugin {
+  return {
+    name: 'angular-loader-import-attributes',
+    setup(build) {
+      build.onLoad({ filter: /./ }, async (args) => {
+        const loader = args.with['loader'] as Loader | undefined;
```
JoostK, line 19: I wonder if there's a feature request for esbuild to have the ability to configure a plugin filter based on an import attribute.

clydin: Discussed here: https://github.com/evanw/esbuild/issues/3612#issuecomment-1912486781
So it's possible it will be available in the future and we can then transition to that when ready.

alan-agius4 (approved):

PR #28060, opened 2024-07-16
refactor: rename internal vite plugin name.

alan-agius4:
Follow convention.

### DIFF
--- a/packages/angular/build/src/tools/vite/id-prefix-plugin.ts
+++ b/packages/angular/build/src/tools/vite/id-prefix-plugin.ts
@@ -17,36 +17,38 @@ const escapeRegexSpecialChars = (inputString: string): string => {
   return inputString.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
 };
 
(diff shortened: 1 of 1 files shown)


packages/angular/build/src/tools/vite/id-prefix-plugin.ts
```diff
@@ -17,36 +17,38 @@ const escapeRegexSpecialChars = (inputString: string): string => {
...
+        return;
+      }
+
+      const escapedExternals = externals.map(escapeRegexSpecialChars);
+      const prefixedExternalRegex = new RegExp(
+        `${VITE_ID_PREFIX}(${escapedExternals.join('|')})`,
+        'g',
+      );
+
+      // @ts-expect-error: Property 'pusxh' does not exist on type 'readonly Plugin<any>[]'
```
JeanMeche, line 36:
```suggestion
      // @ts-expect-error: Property 'push' does not exist on type 'readonly Plugin<any>[]'
```

clydin (approved):

PR #28069, opened 2024-07-18
fix(@angular/build): serve HTML files directly

### LINKED ISSUE #28063: ng serve does not allow us access to raw 'index.csr.html' when using SSR
### Command

serve

### Is this a regression?

_No response_

### Description

If I go to `http://localhost:4200/`, I get a transformed html response which includes my lazy-loaded router-component for the `''`path. This is as expected during SSR.

However,
If I go to `http://localhost:4200/index.html`, I expect to get an html response which is exactly the index.html on disk, without the ssr transformation. But the dev server sends me a "404 Not Found" instead.

Turns out that, for SSR, angular creates 2 index.html - `index.csr.html` and `index.server.html`. I am interested in the csr version, because I want to precache it.

However, the angular dev server's middleware prevents me from accessing that file.

image

This is a screenshot of me debugging the dev-server. The request was for `/index.html`

In line 25, the middleware tries to search the outputFiles for 'index.html'. The problem is that the outputFile don't have 'index.html'. They have 'index.csr.html' and 'index.server.html'. As a result, `rawHtml` is going to be always be `undefined`, and the middleware is  going to go to line 27 and skip the middleware.

If I try to go for `/index.csr.html`, then the middleware will skip out before it even reaches Line 25, due to the `if` condition on Line 12
Line 12 says that any request that are not '/' or '/index.html' will be ignored. 

So we are perpetually denied being served the raw `index.html` while serving in SSR

### Minimal Reproduction

Create a new angular app
Add SSR to it as per the documentation
Add a lazy loaded component for the path `''` with the selector `test-element` and an empty template, 
Remove the contents of app.component's template and add a router outlet to app.component.html
Do other task necessary to set up routing
Serve the app with ng serve

Go to `http://localhost:4200` and check the response that came on first load. You should see the element `<test-element></test-element>` in the html. This is ok

Go to `http://localhost:4200/index.html`. This should result in 404 Not Found. This is ok.

Go to `http://localhost:4200/index.csr.html`. This should result in 404 Not Found. This is not ok. It should have returned html without the test-element in it.

### Exception or Error

_No response_

### Your Environment

```text
Angular CLI: 18.1.0
Node: 20.13.1
Package Manager: pnpm 9.4.0
OS: linux x64

Angular: 18.1.0
... forms, language-service, material, material-luxon-adapter
... router, ssr

@angular-devkit/architect       0.1702.3
@angular-devkit/build-angular   18.1.0
@angular-devkit/core            18.1.0
@angular-devkit/schematics      18.1.0
@schematics/angular             18.1.0
ng-packagr                      18.1.0
typescript                      5.4.4
```

### Anything else relevant?

https://github.com/angular/angular-cli/blob/f5c250ab48e22c5aff31d8ebd35fb9a88e380fd7/packages/angular/build/src/tools/vite/middlewares/index-html-middleware.ts#L30-L42

This is the part that likely needs to be changed. 
- change line 31 to allow  `index.csr.html`
- change line 37 to search for `index.csr.html` or `index.html`  as per the actual request

alan-agius4:
Ensure direct requests to HTML files result in them being served.

Closes #28063

### DIFF
--- a/packages/angular/build/src/tools/vite/middlewares/index-html-middleware.ts
+++ b/packages/angular/build/src/tools/vite/middlewares/index-html-middleware.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
+import { extname } from 'node:path';
 import type { Connect, ViteDevServer } from 'vite';
 import {
   AngularMemoryOutputFiles,
@@ -28,13 +29,14 @@ export function createAngularIndexHtmlMiddleware(
     // Parse the incoming request.
     // The base of the URL is unused but required to parse the URL.
     const pathname = pathnameWithoutBasePath(req.url, server.config.base);
-    if (pathname !== '/' && pathname !== '/index.html') {
+    const extension = extname(pathname);
+    if (extension !== '.html') {
       next();
 
       return;
     }
 
-    const rawHtml = outputFiles.get('/index.html')?.contents;
+    const rawHtml = outputFiles.get(pathname)?.contents;
     if (!rawHtml) {
       next();
 


packages/angular/build/src/tools/vite/middlewares/index-html-middleware.ts
```diff
@@ -28,13 +29,14 @@ export function createAngularIndexHtmlMiddleware(
     // Parse the incoming request.
     // The base of the URL is unused but required to parse the URL.
     const pathname = pathnameWithoutBasePath(req.url, server.config.base);
-    if (pathname !== '/' && pathname !== '/index.html') {
```
alan-agius4, line 31: This part is already handled in https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/tools/vite/middlewares/html-fallback-middleware.ts

clydin (approved):

clydin:

PR #28072, opened 2024-07-18
fix(@angular/build): prevent redirection loop

### LINKED ISSUE #28071: serve: trailing slash added to static files does not take query string into account
### Command

serve

### Is this a regression?

_No response_

### Description

This is more or less a continuation of #27949. I was testing the fix for that issue in Angular 18.1.0 and found that it is still not working. And since I cannot reopen that issue I'm writing this new one.

Whenever I try to access a static file without a trailing slash, angular returns a redirect to the same path with a trailing backslash. 
```
Get: /oauth2redirect

Status: 301
location: /oauth2redirect/
```

**But if the path contains a query string, then the trailing backslash is added to the end of the query string**:
```
Get: /oauth2redirect?code=xxx&scope=yyy&state=zzz

Status: 301
location: /oauth2redirect?code=xxx&scope=yyy&state=zzz/
```

So I end up in an infinite redirect loop that ends with a `ERR_TOO_MANY_REDIRECTS` error and the following URI:
```
/oauth2redirect?code=xxx&scope=yyy&state=zzz///////////////////////////////////////////////////////////////////////
```

### Minimal Reproduction

The same used in #27949. Just add some query string parameter to the static file.

### Exception or Error

_No response_

### Your Environment

```text
Angular CLI: 18.1.0
Node: 18.20.3
Package Manager: npm 10.8.1
OS: darwin arm64

Angular: 18.1.0
... forms, language-service, localize, material
... service-worker

@angular-devkit/architect       0.1801.0
@angular-devkit/build-angular   18.1.0
@angular-devkit/core            18.1.0
@angular-devkit/schematics      18.1.0
@schematics/angular             18.1.0
typescript                      5.5.3
zone.js                         0.14.7
```

### Anything else relevant?

Pinging @alan-agius4 as you are already familiar with the issue.

alan-agius4:
Added handling to ensure trailing slashes on static files do not disregard query strings.

Closes #28071

### DIFF
--- a/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
+++ b/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
@@ -87,7 +87,8 @@ export function createAngularAssetsMiddleware(
     if (!pathnameHasTrailingSlash) {
       for (const assetPath of assets.keys()) {
         if (pathname === assetPath.substring(0, assetPath.lastIndexOf('/'))) {
-          const location = req.url + '/';
+          const { pathname, search, hash } = new URL(req.url, 'http://localhost');
+          const location = [pathname, '/', search, hash].join('');
 
           res.statusCode = 301;
           res.setHeader('Content-Type', 'text/html');


clydin (approved):

alan-agius4:

PR #28076, opened 2024-07-18
refactor(@angular/build): provide structured application builder result types

clydin:
The application builder now provides structured output types to its internal consumers. The architect builders themselves and the programmatic API is not changed. These output result types allow for the development server to receive additional information regarding the build and update the active browser appropriately. This functionality is not yet implemented but the additional result types provide the base infrastructure to enable future features. The result types also allow for reduced complexity inside other builders such as i18n extraction and the browser compatibility builder. The usage is not yet fully optimized and will be refined in future changes.


alan-agius4 (approved):
Nice! great job.

PR #28080, opened 2024-07-18
fix(@angular/build): account for HTML base HREF for dev-server externals

### LINKED ISSUE #28077: Explicit externals not working in the Vite Dev Server when BaseHref is set
### Command

serve

### Is this a regression?

18.0

### Description

External dependencies set in angular.json used to be rewritten to /@id/dependency when running the dev server prior to Angular 18.1.

Since version 18.1 the /@id/ part of the dependency was removed from external dependencies, which is helpful but doesn't take in account that vite will rewrite urls to match the BaseHref option and removes any trailing slash added to the BaseHref.

### Minimal Reproduction

- Create a new application with `ng new`
- Install any package with npm and use it in the AppComponent
- The package to "externalDependencies" in angular.json (ex: `eruda`)
- Add a baseHref with a trailing slash to angular.json (ex: `/test/`)
- Run ng serve and notice that Angular has failed to load due to vite trying to load `/testeruda` and failing

### Exception or Error

_No response_

### Your Environment

```text
                |___/
    

Angular CLI: 18.1.1
Node: 18.19.1
Package Manager: npm 10.2.4
OS: win32 x64

Angular: 18.1.1

@angular-devkit/architect       0.1801.1
@angular-devkit/build-angular   18.1.1
@angular-devkit/core            18.1.1
@angular-devkit/schematics      18.1.1
@schematics/angular             18.1.1
typescript                      5.5.3
```

### Anything else relevant?

_No response_

clydin:
When adjusting URLs to support explicit external dependencies when using Vite, the workaround will now account for the presence of a base HREF value within the specifier. Vite will automatically add the base HREF as a prefix to the path when specified. This previously resulted in invalid specifiers due to the partial removal of the full Vite specific `@id` path prefix.

Closes #28077

### DIFF
--- a/packages/angular/build/src/builders/dev-server/tests/behavior/build-external-dependencies_spec.ts
+++ b/packages/angular/build/src/builders/dev-server/tests/behavior/build-external-dependencies_spec.ts
@@ -45,5 +45,43 @@ describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupT
       expect(text).toContain(`import { BehaviorSubject } from "rxjs";`);
       expect(text).toContain(`import { map } from "rxjs/operators";`);
     });
+
+    it('respects import specifiers when using baseHref with trailing slash', async () => {
+      setupTarget(harness, {
+        externalDependencies: ['rxjs', 'rxjs/operators'],
+        baseHref: '/test/',
+      });
+
+      harness.useTarget('serve', {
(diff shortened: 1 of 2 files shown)


alan-agius4 (approved):

kfrederix:
@clydin kudos for fixing this so fast! (and sorry for the bug, totally missed that case...)

PR #28114, opened 2024-07-26
fix(@angular/build): add CSP `nonce` attribute to script tags when inline critical CSS is disabled

### LINKED ISSUE #28102: Angular 18.1.1  is not adding CSP nonce attribute to script tags generated during build(main.ts,polyfills.js etc...)
### Command

build

### Is this a regression?

_No response_

### Description

I want to implement strict CSP policy for my project, so replaced unsafe-eval and unsafe-inline with nonce-dynamicnonce.
But Angular version 18.1.1 is not adding nonce to script tags of main.js, polyfills.js etc. 
Because of that it is considering this scripts tags as unsafe and browser is blocking its execution.
In below stackblitz I used 18.1.1 and added nonce using CSP_NONCE and ngCspNonce. But it is not adding nonce to script tags. Please inspect and check in Dom.
Stackblitz URL: https://stackblitz.com/edit/stackblitz-starters-wr9a9h?file=src%2Fapp%2Fapp.config.ts

As per below merged Issue, It should work in 18.1.1. But it is not working.
https://github.com/angular/angular-cli/pull/27875

Kindly check!!
Thanks in advanced!

### Minimal Reproduction

1) apply nonce using ngCspNonce and CSP_NONCE provider
2) using dev tools check script tags  in DOM.
3) It is not adding nonce to script tags of main.js,polyfills.js etc..

### Exception or Error

_No response_

### Your Environment

```text
18.1.1
```

### Anything else relevant?

_No response_

alan-agius4:
Prior to this change when inline critical CSS is disabled CSP `nonce` was not added to script tags.

Closes #28102

### DIFF
--- a/packages/angular/build/src/builders/application/tests/behavior/csp-nonce_spec.ts
+++ b/packages/angular/build/src/builders/application/tests/behavior/csp-nonce_spec.ts
@@ -1,0 +1,35 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { buildApplication } from '../../index';
+import { APPLICATION_BUILDER_INFO, BASE_OPTIONS, describeBuilder } from '../setup';
+
+describeBuilder(buildApplication, APPLICATION_BUILDER_INFO, (harness) => {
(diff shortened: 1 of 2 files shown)


dgp1130 (approved):
Is there any easy opportunity for a test to ensure that the nonce is set in this scenario?

alan-agius4:

PR #28115, opened 2024-07-26
fix(@angular/build): resolve error with `extract-i18n` builder for libraries

### LINKED ISSUE #28109: Path[1] argument must be of type string. Received undefined
### Command

extract-i18n

### Is this a regression?

_No response_

### Description

I have been trying to export the i18n information from a part of our code, previously this code was in our project as components.
Now I want to create a library out of it and the actual building and using it in the project as a package works but when exporting with `ng extract-i18n --project <projectNameHere>` I keep getting the error from the title: 

> An unhandled exception occurred: The "paths[1]" argument must be of type string. Received undefined

In debugging the offending line and file I've found that the actual object used in this process has the `path` variable but it's an object.

Changing this to an array or just a string does not resolve the issue and even breaks the actual build of the library.

I've reproduced this in a brand new project and still the same error.

### Minimal Reproduction

Create a new workspace, add a library and try to extract localization info

- ng new company.Angular.Components --no-create-application
- cd .\company.Angular.Components\  
- ng generate library @company/angular-lib-forms 
- ng add @angular/localize
- ng extract-i18n --project @company/angular-lib-forms

### Exception or Error

```text
[error] TypeError [ERR_INVALID_ARG_TYPE]: The "paths[1]" argument must be of type string. Received undefined
    at Object.resolve (node:path:171:9)

    at getCommonConfig (C:\Users\<username>\Desktop\Projects\<projectname>\node_modules\@angular-devkit\build-angular\src\tools\webpack\configs\common.js:275:24)

    at async Promise.all (index 2)

    at async generateWebpackConfig (C:\Users\<username>\Desktop\Projects\<projectname>\node_modules\@angular-devkit\build-angular\src\utils\webpack-browser-config.js:64:22)

    at async generateBrowserWebpackConfigFromContext (C:\Users\<username>\Desktop\Projects\<projectname>\node_modules\@angular-devkit\build-angular\src\utils\webpack-browser-config.js:119:20)

    at async extractMessages (C:\Users\<username>\Desktop\Projects\<projectname>\node_modules\@angular-devkit\build-angular\src\builders\extract-i18n\webpack-extraction.js:52:24)

    at async execute (C:\Users\<username>\Desktop\Projects\<projectname>\node_modules\@angular-devkit\build-angular\src\builders\extract-i18n\builder.js:84:28)
```

### Your Environment

```text
Angular CLI: 18.1.2
Node: 20.15.0
Package Manager: npm 10.7.0
OS: win32 x64

Angular: 18.1.1

@angular-devkit/architect       0.1801.2
@angular-devkit/build-angular   18.1.2
@angular-devkit/core            18.1.2
@angular-devkit/schematics      18.1.2
@angular/cli                    18.1.2
@schematics/angular             18.1.2
ng-packagr                      18.1.0
```

### Anything else relevant?

_No response_

alan-agius4:
The `extract-i18n` builder is only intended to be used with application projects.

Closes #28109

### DIFF
--- a/packages/angular/build/src/builders/extract-i18n/builder.ts
+++ b/packages/angular/build/src/builders/extract-i18n/builder.ts
@@ -32,6 +32,18 @@ export async function execute(
     return { success: false };
   }
 
+  const { projectType } = (await context.getProjectMetadata(projectName)) as {
+    projectType?: string;
+  };
+  if (projectType !== 'application') {
+    context.logger.error(
+      `Tried to extract from ${projectName} with 'projectType' ${projectType}, which is not supported.` +
+        ` The 'extract-i18n' builder can only extract from applications.`,
+    );
+
+    return { success: false };
+  }
+
   // Check Angular version.
   assertCompatibleAngularVersion(context.workspaceRoot);
 
--- a/packages/angular_devkit/build_angular/src/builders/extract-i18n/builder.ts
+++ b/packages/angular_devkit/build_angular/src/builders/extract-i18n/builder.ts
@@ -35,6 +35,18 @@ export async function execute(
     return { success: false };
   }
 
+  const { projectType } = (await context.getProjectMetadata(projectName)) as {
+    projectType?: string;
+  };
+  if (projectType !== 'application') {
+    context.logger.error(
+      `Tried to extract from ${projectName} with 'projectType' ${projectType}, which is not supported.` +
+        ` The 'extract-i18n' builder can only extract from applications.`,
+    );
+
+    return { success: false };
+  }
+
   // Check Angular version.
   assertCompatibleAngularVersion(context.workspaceRoot);
 


packages/angular/build/src/builders/extract-i18n/builder.ts
```diff
@@ -32,6 +32,17 @@ export async function execute(
     return { success: false };
   }
 
+  const { projectType } = (await context.getProjectMetadata(projectName)) as {
+    projectType?: string;
+  };
+  if (projectType !== 'application') {
+    context.logger.error(
+      `The 'extract-i18n' builder requires the project type to be an application.`,
```
dgp1130, line 40: I think a more descriptive error message would be useful to make sure we're clear about what project we're referring to and what we mean by "project type" in this context. How about something like:

> Tried to extract from `${projectName}` with `projectType` `${projectType}`, which is not supported. The `extract-i18n` builder can only extract from applications.

This way we're calling out the affected project (there could be multiple) and referencing the relevant `projectType` property (which is going to print `library` in this case).

dgp1130 (approved):
Should we have a test somewhere that the builder emits this error as expected?

alan-agius4:

PR #28116, opened 2024-07-26
fix(@angular/build): prevent build failures with remote CSS imports when Tailwind is configured

### LINKED ISSUE #28113: esbuild: Using Tailwind breaks stylesheet processing for CSS @imports with URLs ending in .css
### Command

build

### Is this a regression?

_No response_

### Description

When Tailwind is configured in a project (maybe just PostCSS), any `@import url()` statement with a URL ending in `.css` (such as `@import url(https://fonts.googleapis.com/earlyaccess/notokufiarabic.css);` causes the build to fail, as the url is deemed a local file by the language filters passed to [`StyleSheetPluginFactory.create()`](https://github.com/angular/angular-cli/blob/048df34ccb7d4ef67b5877e126c74c0db65b879c/packages/angular/build/src/tools/esbuild/stylesheets/bundle-options.ts#L59-L61). The `create` method seems to register an esbuild `onLoad` handler that will [read the path as a local file](https://github.com/angular/angular-cli/blob/048df34ccb7d4ef67b5877e126c74c0db65b879c/packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts#L167-L170)

Any URLs ending in `.less`, `.scss` or `.sass` are also affected.

In our case we can't easily remove `.css` from the imported URL, as it's defined by an imported third party CSS file.

### Minimal Reproduction

1. Create a new project with `ng new`
2. Install Tailwind by following the [Tailwind + Angular installation guide](https://tailwindcss.com/docs/guides/angular)
3. Add the following to `src/styles.css`: `@import url(https://fonts.googleapis.com/earlyaccess/notokufiarabic.css);`
4. Run `npm run build`
5. The build fails, as the URL is attemped to be read as a local file

When no Tailwind config file is present, the build runs fine:
1. Remove `tailwind.config.js` and the `@tailwind` lines from the CSS file, while keeping the `@import url()`
2. Run `npm run build`
3. The build succeeds

### Exception or Error

```text
ENOENT: no such file or directory, open 'https://fonts.googleapis.com/earlyaccess/notokufiarabic.css' [plugin angular-css]
```

### Your Environment

```text
Angular CLI: 18.1.2
Node: 20.14.0
Package Manager: npm 10.7.0
OS: darwin arm64

Angular: 18.1.2

@angular-devkit/architect       0.1801.2
@angular-devkit/build-angular   18.1.2
@angular-devkit/core            18.1.2
@angular-devkit/schematics      18.1.2
@schematics/angular             18.1.2
```

### Anything else relevant?

_No response_

alan-agius4:
This addresses a bug where `@import url()` statements with remote CSS files (ending in .css) caused build errors when Tailwind was present. The issue arised from incorrect handling of remote URLs by the stylesheet plugin, which treated them as local files. This fix ensures proper handling of remote CSS imports.

Closes #28113

### DIFF
--- a/packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
+++ b/packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
@@ -165,7 +165,7 @@ export class StylesheetPluginFactory {
 
         // Add a load callback to support files from disk
         build.onLoad(
-          { filter: language.fileFilter },
+          { filter: language.fileFilter, namespace: 'file' },
           createCachedLoad(cache, async (args) => {
             const data = await readFile(args.path, 'utf-8');
 
--- a/packages/angular/build/src/tools/esbuild/stylesheets/bundle-options.ts
+++ b/packages/angular/build/src/tools/esbuild/stylesheets/bundle-options.ts
@@ -63,7 +63,7 @@ export function createStylesheetBundleOptions(
   ];
 
   if (options.inlineFonts) {
-    plugins.push(createCssInlineFontsPlugin({ cache, cacheOptions: options.cacheOptions }));
+    plugins.unshift(createCssInlineFontsPlugin({ cache, cacheOptions: options.cacheOptions }));
   }
 
   return {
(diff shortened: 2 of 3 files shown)


packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
```diff
@@ -165,7 +165,7 @@ export class StylesheetPluginFactory {
 
         // Add a load callback to support files from disk
         build.onLoad(
-          { filter: language.fileFilter },
+          { filter: language.fileFilter, namespace: 'file' },
```
alan-agius4, line 168: This is not required, but I think it's worth having.

clydin (approved):

alan-agius4:

PR #28153, opened 2024-08-02
refactor(@angular-devkit/build-angular): only write test related files in WTR builder

clydin:
With the structured build results available, the WTR builder can now more easily write only the needed files to a temporary location. This reduces the need to assume build directory output structure and reduces the amount of potential filesystem calls. The temporary files are also now written into a UUID subdirectory within the existing `dist/test-out` location. This allows for multiple projects to be tested concurrently without overwriting each other.
The Jest builder has also been updated with similar changes.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/jest/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/jest/index.ts
@@ -6,18 +6,17 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):

alan-agius4:
@clydin, looks like you have 2 identical commits.

clydin:
One is for WTR and one is for Jest

PR #28161, opened 2024-08-05
fix(@angular/build): allow explicitly disabling TypeScript incremental mode

clydin:
If the TypeScript `incremental` option is explicitly set to `false`, the `application` builder will no longer attempt to enable and use incremental compilation mode via stored TypeScript build file information. This prevents an potential build time error that would otherwise occur due to the `tsBuildInfoFile` option being set without the `incremental` option. Behavior remains the same if the option is not present or set to `true`.


alan-agius4 (approved):

clydin:

PR #28173, opened 2024-08-07
fix(@angular/build): lazy load Node.js inspector for dev server

### LINKED ISSUE #28172: `ng build` throws on `import inspector from 'node:inspector';` on gentoo default node.js install
### Command

build, serve

### Is this a regression?

No

### Description

One of our linux enjoyers couldn't build via angular cli due a node.js error: `ERR_INSPECTOR_NOT_AVAILABLE`.

Turns out the `import inspector from 'node:inspector'` from https://github.com/angular/angular-cli/blob/39b00dbb050991f058bc38f45cbda96960c1f6a2/packages/angular/build/src/builders/dev-server/vite-server.ts#L13 fails, as gentoo linux doesn't include the inspector by default. You have to explicitly opt-in via `USE inspector` (https://wiki.gentoo.org/wiki/Node.js / https://packages.gentoo.org/useflags/inspector)

Would be lovely if you could lazyload the `'node:inspector'` package on demand, given that it's a super niche usecase (ssr + inspect enabled)

Feel free to reject if this sounds ridiculous. He already recompiled node.js with injector included by now.

clydin:
The Node.js inspector will now only be imported if SSR is enabled and the `inspect` option is used. This prevents potential failures when a custom Node.js binary build is used that has not enabled inspector usage. While this is not common, loading the inspector on demand is a minimal change that also avoids loading code that will not be used for the majority of development server usage.

Closes #28172

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -10,7 +10,6 @@ import type { BuilderContext } from '@angular-devkit/architect';
 import type { Plugin } from 'esbuild';
 import assert from 'node:assert';
 import { readFile } from 'node:fs/promises';
-import inspector from 'node:inspector';
 import { builtinModules } from 'node:module';
 import { join } from 'node:path';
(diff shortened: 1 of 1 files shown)


clydin:

PR #28183, opened 2024-08-09
feat(@angular/ssr): dynamic route resolution using Angular router

alan-agius4:
This enhancement eliminates the dependency on file extensions for server-side rendering (SSR) route handling, leveraging Angular's router configuration for more dynamic and flexible route determination. Additionally, configured redirectTo routes now correctly respond with a 302 redirect status.

The new router uses a radix tree for storing routes. This data structure allows for efficient prefix-based lookups and insertions, which is particularly crucial when dealing with nested and parameterized routes.

This change also lays the groundwork for potential future server-side routing configurations, further enhancing the capabilities of Angular's SSR functionality.

### DIFF
--- a/packages/angular/ssr/src/assets.ts
+++ b/packages/angular/ssr/src/assets.ts
@@ -1,0 +1,47 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { AngularAppManifest } from './manifest';
+
+/**
+ * Manages server-side assets.
+ */
+export class ServerAssets {
+  /**
+   * Creates an instance of ServerAsset.
+   *
+   * @param manifest - The manifest containing the server assets.
+   */
+  constructor(private readonly manifest: AngularAppManifest) {}
+
+  /**
+   * Retrieves the content of a server-side asset using its path.
+   *
+   * @param path - The path to the server asset.
+   * @returns A promise that resolves to the asset content as a string.
+   * @throws Error If the asset path is not found in the manifest, an error is thrown.
+   */
+  async getServerAsset(path: string): Promise<string> {
+    const asset = this.manifest.assets[path];
+    if (!asset) {
+      throw new Error(`Server asset '${path}' does not exist.`);
+    }
+
+    return asset();
+  }
+
+  /**
+   * Retrieves and caches the content of 'index.server.html'.
+   *
+   * @returns A promise that resolves to the content of 'index.server.html'.
+   * @throws Error If there is an issue retrieving the asset.
+   */
+  getIndexServerHtml(): Promise<string> {
+    return this.getServerAsset('index.server.html');
+  }
+}
--- a/packages/angular/ssr/src/routes/ng-routes.ts
+++ b/packages/angular/ssr/src/routes/ng-routes.ts
@@ -1,0 +1,239 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { APP_BASE_HREF, PlatformLocation } from '@angular/common';
+import {
+  ApplicationRef,
+  Compiler,
+  Injector,
+  createPlatformFactory,
+  platformCore,
+  ɵwhenStable as whenStable,
+  ɵConsole,
+  ɵresetCompiledComponents,
+} from '@angular/core';
+import {
+  INITIAL_CONFIG,
+  ɵINTERNAL_SERVER_PLATFORM_PROVIDERS as INTERNAL_SERVER_PLATFORM_PROVIDERS,
+} from '@angular/platform-server';
+import { Route, Router, ɵloadChildren as loadChildrenHelper } from '@angular/router';
+import { Console } from '../console';
+import { AngularBootstrap, isNgModule } from '../utils/ng';
+import { joinUrlParts } from '../utils/url';
+
+/**
+ * Result of extracting routes from an Angular application.
+ */
+interface AngularRouterConfigResult {
+  /**
+   * The base URL for the application.
+   * This is the base href that is used for resolving relative paths within the application.
+   */
+  baseHref: string;
+
+  /**
+   * An async iterator that yields `RouteResult` objects.
+   *
+   * Each `RouteResult` represents a route and its associated information, such as the path
+   * and any potential redirection target.
+   */
+  routes: AsyncIterableIterator<RouteResult>;
+}
+
+/**
+ * Represents the result of processing a route.
+ */
+interface RouteResult {
+  /**
+   * The resolved path of the route.
+   *
+   * This string represents the complete URL path for the route after it has been
+   * resolved, including any parent routes or path segments that have been joined.
+   */
+  route: string;
+
+  /**
+   * The target path for route redirection, if applicable.
+   *
+   * If this route has a `redirectTo` property in the configuration, this field will
+   * contain the full resolved URL path that the route should redirect to.
+   */
+  redirectTo?: string;
+}
+
+/**
+ * Recursively traverses the Angular router configuration to retrieve routes.
+ *
+ * Iterates through the router configuration, yielding each route along with its potential
+ * redirection or error status. Handles nested routes and lazy-loaded child routes.
+ *
+ * @param options - An object containing the parameters for traversing routes.
+ * @returns An async iterator yielding `RouteResult` objects.
+ */
+async function* traverseRoutesConfig(options: {
+  /** The array of route configurations to process. */
+  routes: Route[];
+  /** The Angular compiler used to compile route modules. */
+  compiler: Compiler;
+  /** The parent injector for lazy-loaded modules. */
+  parentInjector: Injector;
+  /** The parent route path to prefix child routes. */
+  parentRoute: string;
+}): AsyncIterableIterator<RouteResult> {
+  const { routes, compiler, parentInjector, parentRoute } = options;
+
+  for (const route of routes) {
+    const { path = '', redirectTo, loadChildren, children } = route;
+    const currentRoutePath = joinUrlParts(parentRoute, path);
+
+    yield {
+      route: currentRoutePath,
+      redirectTo:
+        typeof redirectTo === 'string'
+          ? resolveRedirectTo(currentRoutePath, redirectTo)
+          : undefined,
+    };
+
+    if (children?.length) {
+      // Recursively process child routes.
+      yield* traverseRoutesConfig({
+        routes: children,
+        compiler,
+        parentInjector,
+        parentRoute: currentRoutePath,
+      });
+    }
+
+    if (loadChildren) {
+      // Load and process lazy-loaded child routes.
+      const loadedChildRoutes = await loadChildrenHelper(
+        route,
+        compiler,
+        parentInjector,
+      ).toPromise();
+
+      if (loadedChildRoutes) {
+        const { routes: childRoutes, injector = parentInjector } = loadedChildRoutes;
+        yield* traverseRoutesConfig({
+          routes: childRoutes,
+          compiler,
+          parentInjector: injector,
+          parentRoute: currentRoutePath,
+        });
+      }
+    }
+  }
+}
+
+/**
+ * Resolves the `redirectTo` property for a given route.
+ *
+ * This function processes the `redirectTo` property to ensure that it correctly
+ * resolves relative to the current route path. If `redirectTo` is an absolute path,
+ * it is returned as is. If it is a relative path, it is resolved based on the current route path.
+ *
+ * @param routePath - The current route path.
+ * @param redirectTo - The target path for redirection.
+ * @returns The resolved redirect path as a string.
+ */
+function resolveRedirectTo(routePath: string, redirectTo: string): string {
+  if (redirectTo[0] === '/') {
+    // If the redirectTo path is absolute, return it as is.
+    return redirectTo;
+  }
+
+  // Resolve relative redirectTo based on the current route path.
+  const segments = routePath.split('/');
+  segments.pop(); // Remove the last segment to make it relative.
+
+  return joinUrlParts(...segments, redirectTo);
+}
+
+/**
+ * Retrieves routes from the given Angular application.
+ *
+ * This function initializes an Angular platform, bootstraps the application or module,
+ * and retrieves routes from the Angular router configuration. It handles both module-based
+ * and function-based bootstrapping. It yields the resulting routes as `RouteResult` objects.
+ *
+ * @param bootstrap - A function that returns a promise resolving to an `ApplicationRef` or an Angular module to bootstrap.
+ * @param document - The initial HTML document used for server-side rendering.
+ * This document is necessary to render the application on the server.
+ * @param url - The URL for server-side rendering. The URL is used to configure `ServerPlatformLocation`. This configuration is crucial
+ * for ensuring that API requests for relative paths succeed, which is essential for accurate route extraction.
+ * See:
+ *  - https://github.com/angular/angular/blob/d608b857c689d17a7ffa33bbb510301014d24a17/packages/platform-server/src/location.ts#L51
+ *  - https://github.com/angular/angular/blob/6882cc7d9eed26d3caeedca027452367ba25f2b9/packages/platform-server/src/http.ts#L44
+ * @returns A promise that resolves to an object of type `AngularRouterConfigResult`.
+ */
+export async function getRoutesFromAngularRouterConfig(
+  bootstrap: AngularBootstrap,
+  document: string,
+  url: URL,
+): Promise<AngularRouterConfigResult> {
+  // Need to clean up GENERATED_COMP_IDS map in `@angular/core`.
+  // Otherwise an incorrect component ID generation collision detected warning will be displayed in development.
+  // See: https://github.com/angular/angular-cli/issues/25924
+  ɵresetCompiledComponents();
+
+  const { protocol, host } = url;
+
+  // Create and initialize the Angular platform for server-side rendering.
+  const platformRef = createPlatformFactory(platformCore, 'server', [
+    {
+      provide: INITIAL_CONFIG,
+      useValue: { document, url: `${protocol}//${host}/` },
+    },
+    {
+      provide: ɵConsole,
+      useFactory: () => new Console(),
+    },
+    ...INTERNAL_SERVER_PLATFORM_PROVIDERS,
+  ])();
+
+  try {
+    let applicationRef: ApplicationRef;
+
+    if (isNgModule(bootstrap)) {
+      const moduleRef = await platformRef.bootstrapModule(bootstrap);
+      applicationRef = moduleRef.injector.get(ApplicationRef);
+    } else {
+      applicationRef = await bootstrap();
+    }
+
+    // Wait until the application is stable.
+    await whenStable(applicationRef);
+
+    const injector = applicationRef.injector;
+    const router = injector.get(Router);
+
+    const baseHref =
+      injector.get(APP_BASE_HREF, null, { optional: true }) ??
+      injector.get(PlatformLocation).getBaseHrefFromDOM();
+
+    if (router.config.length === 0) {
+      // No routes found in the configuration.
+      return { baseHref, routes: (async function* () {})() };
+    } else {
+      const compiler = injector.get(Compiler);
+
+      // Retrieve all routes from the Angular router configuration.
+      return {
+        baseHref,
+        routes: traverseRoutesConfig({
+          routes: router.config,
+          compiler,
+          parentInjector: injector,
+          parentRoute: '',
+        }),
+      };
+    }
+  } finally {
+    platformRef.destroy();
+  }
+}
--- a/packages/angular/ssr/src/routes/route-tree.ts
+++ b/packages/angular/ssr/src/routes/route-tree.ts
@@ -1,0 +1,296 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { stripTrailingSlash } from '../utils/url';
+
+/**
+ * Represents the serialized format of a route tree as an array of node metadata objects.
+ * Each entry in the array corresponds to a specific node's metadata within the route tree.
+ */
+export type SerializableRouteTreeNode = ReadonlyArray<RouteTreeNodeMetadata>;
+
+/**
+ * Represents metadata for a route tree node, excluding the 'route' path segment.
+ */
+export type RouteTreeNodeMetadataWithoutRoute = Omit<RouteTreeNodeMetadata, 'route'>;
+
+/**
+ * Describes metadata associated with a node in the route tree.
+ * This metadata includes information such as the route path and optional redirect instructions.
+ */
+export interface RouteTreeNodeMetadata {
+  /**
+   * Optional redirect path associated with this node.
+   * This defines where to redirect if this route is matched.
+   */
+  redirectTo?: string;
+
+  /**
+   * The route path for this node.
+   *
+   * A "route" is a URL path or pattern that is used to navigate to different parts of a web application.
+   * It is made up of one or more segments separated by slashes `/`. For instance, in the URL `/products/details/42`,
+   * the full route is `/products/details/42`, with segments `products`, `details`, and `42`.
+   *
+   * Routes define how URLs map to views or components in an application. Each route segment contributes to
+   * the overall path that determines which view or component is displayed.
+   *
+   * - **Static Routes**: These routes have fixed segments. For example, `/about` or `/contact`.
+   * - **Parameterized Routes**: These include dynamic segments that act as placeholders, such as `/users/:id`,
+   *   where `:id` could be any user ID.
+   *
+   * In the context of `RouteTreeNodeMetadata`, the `route` property represents the complete path that this node
+   * in the route tree corresponds to. This path is used to determine how a specific URL in the browser maps to the
+   * structure and content of the application.
+   */
+  route: string;
+}
+
+/**
+ * Represents a node within the route tree structure.
+ * Each node corresponds to a route segment and may have associated metadata and child nodes.
+ */
+interface RouteTreeNode {
+  /**
+   * The segment value associated with this node.
+   * A segment is a single part of a route path, typically delimited by slashes (`/`).
+   * For example, in the route `/users/:id/profile`, the segments are `users`, `:id`, and `profile`.
+   * Segments can also be wildcards (`*`), which match any segment in that position of the route.
+   */
+  segment: string;
+
+  /**
+   * The index indicating the order in which the route was inserted into the tree.
+   * This index helps determine the priority of routes during matching, with lower indexes
+   * indicating earlier inserted routes.
+   */
+  insertionIndex: number;
+
+  /**
+   * A map of child nodes, keyed by their corresponding route segment or wildcard.
+   */
+  children: Map<string, RouteTreeNode>;
+
+  /**
+   * Optional metadata associated with this node, providing additional information such as redirects.
+   */
+  metadata?: RouteTreeNodeMetadata;
+}
+
+/**
+ * A route tree implementation that supports efficient route matching, including support for wildcard routes.
+ * This structure is useful for organizing and retrieving routes in a hierarchical manner,
+ * enabling complex routing scenarios with nested paths.
+ */
+export class RouteTree {
+  /**
+   * The root node of the route tree.
+   * All routes are stored and accessed relative to this root node.
+   */
+  private readonly root = this.createEmptyRouteTreeNode('');
+
+  /**
+   * A counter that tracks the order of route insertion.
+   * This ensures that routes are matched in the order they were defined,
+   * with earlier routes taking precedence.
+   */
+  private insertionIndexCounter = 0;
+
+  /**
+   * Inserts a new route into the route tree.
+   * The route is broken down into segments, and each segment is added to the tree.
+   * Parameterized segments (e.g., :id) are normalized to wildcards (*) for matching purposes.
+   *
+   * @param route - The route path to insert into the tree.
+   * @param metadata - Metadata associated with the route, excluding the route path itself.
+   */
+  insert(route: string, metadata: RouteTreeNodeMetadataWithoutRoute): void {
+    let node = this.root;
+    const normalizedRoute = stripTrailingSlash(route);
+    const segments = normalizedRoute.split('/');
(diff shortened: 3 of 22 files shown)


packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -0,0 +1,183 @@
...
+ * Initializes an Angular platform, bootstraps the application or module,
+ * and retrieves routes from the Angular router configuration. Handles both module-based and
+ * function-based bootstrapping. Yields the resulting routes as `RouteResult` objects.
+ *
+ * @param bootstrap - A function that returns a promise resolving to an `ApplicationRef` or an Angular module to bootstrap.
+ * @param document - The initial HTML document used for server-side rendering.
+ * @param url - The URL for server-side rendering.
+ * @returns An async iterator yielding `RouteResult` objects.
+ */
+export async function* getRoutesFromAngularRouterConfig(
```
AndrewKushnir, line 62: Do we use a similar logic when extracting routes for SSG purposes? If that's the case, I'm wondering if we can eventually share the logic to avoid duplication and simplify maintenance.

packages/angular/ssr/src/utils/url.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (part[0] === '/') {
+      start = 1;
+    }
+
+    let end: -1 | undefined;
+    if (part[part.length - 1] === '/') {
+      end = -1;
+    }
+
+    const normalizePart = start || end ? part.slice(start, end) : part;
```
AndrewKushnir, line 61: nit:
```suggestion
    const normalizePart = start ?? end ? part.slice(start, end) : part;
```

AndrewKushnir (approved):
@alan-agius4 looks great, thanks 👍

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    return bestMatchMetadata;
+  }
+
+  /**
+   * Converts the route tree into a nested object representation.
+   * @returns A nested object where each key is a route segment and each value is an object containing
+   *          metadata and possibly further nested route segments.
+   */
+  toObject(): RouteNodeObjectRepresentation {
```
dgp1130, line 164: Question: What is the use case for `toObject` and `fromObject`? They seem to only be used in tests.

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+  /** The route segment for this node. */
+  route: string;
+}
+
+/**
+ * A node in a route tree.
+ */
+interface RouteTreeNode {
+  /** The value associated with this node (a route segment or wildcard). */
+  segment: string;
```
dgp1130, line 38: Question: How is `segment` different from `metadata.route`?

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+      node: RouteTreeNode,
+      currentPath: string,
+      currentObj: RouteNodeObjectRepresentation,
+    ) => {
+      if (node.metadata) {
+        currentObj.metadata = node.metadata;
+      }
+
+      if (node.children) {
+        for (const [segment, childNode] of node.children) {
```
dgp1130, line 180: TIL: I always thought you needed `node.children.entries()`, I didn't realize `Map` was directly iterable like this!

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+   */
+  toObject(): RouteNodeObjectRepresentation {
+    const result: RouteNodeObjectRepresentation = {};
+
+    /**
+     * Recursively traverses the route tree to build the nested object.
+     */
+    const traverse = (
+      node: RouteTreeNode,
+      currentPath: string,
```
dgp1130, line 172: Nit: Is this used for anything? Looks like we build up the path string recursively but don't do anything with it.

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+  static fromObject(nestedObject: RouteNodeObjectRepresentation): RouteTree {
+    const tree = new RouteTree();
+
+    /**
+     * Recursively traverses the nested object to insert routes into the tree.
+     */
+    const traverse = (currentObj: RouteNodeObjectRepresentation, currentPath: string) => {
+      if (currentObj.metadata) {
+        // Insert the current path with its metadata into the route tree
+        tree.insert(currentPath, currentObj.metadata);
```
dgp1130, line 208: Question: How does this work with ordering? `tree.insert` is sensitive to ordering right? Is `RouteTree.fromObject(tree.toObject())` guaranteed to preserve ordering and have the same match behavior?

Even if that does work, wouldn't this imply that `RouteTree.fromObject` is sensitive to the ordering of properties in `children: { /* ... */ }`? That feels brittle to me and easy to not understand when using this API.

packages/angular/ssr/src/routes/router.ts
```diff
@@ -0,0 +1,95 @@
...
+  requiresRoutesToBeBuilt = true;
+
+  /**
+   * Creates an instance of the Router.
+   *
+   * @param manifest - An instance of `AngularAppManifest` that contains route information.
+   * The manifest provides routing configuration and other details necessary for route management.
+   */
+  constructor(private manifest: AngularAppManifest) {
+    if (manifest.routes) {
```
dgp1130, line 34: Question: Why do we have this branch? Why is it that sometimes the manifest will have routes and sometimes it won't?

I typically try to lean away from having "init" functions like `buildRoutes` which are easy to forget to call or call out of order (ex. accidentally calling `match` before `buildRoutes` is done). I'm wondering if there's a way to refactor this so the `ServerRouter` always has access to a list of routes (or at least a consistent way to get them).

packages/angular/ssr/src/utils/ng.ts
```diff
@@ -28,13 +29,20 @@ import { renderApplication, renderModule } from '@angular/platform-server';
 export function renderAngular(
   html: string,
   bootstrap: Type<unknown> | (() => Promise<ApplicationRef>),
-  url: string,
+  url: URL,
```
dgp1130, line 32: Consider: Call this `unnormalizedUrl` and then switch `normalizedUrl` -> `url`. That way the default "url" concept is the normalized version. This makes it harder for future changes to this function to accidentally use the unnormalized version.

packages/angular/ssr/src/utils/url.ts
```diff
@@ -0,0 +1,97 @@
...
+ * This function takes multiple URL segments, normalizes them by removing leading
+ * and trailing slashes where appropriate, and then joins them into a single URL.
+ *
+ * @param parts - The parts of the URL to join. Each part can be a string with or without slashes.
+ * @returns The joined URL string, with normalized slashes.
+ *
+ * @example
+ * ```js
+ * joinUrlParts('path/', '/to/resource'); // '/path/to/resource'
+ * joinUrlParts('/path/', 'to/resource'); // '/path/to/resource'
```
dgp1130, line 38: Question: Do we need to care about cases like `joinUrlParts('unnecessary////slashes')`?

packages/angular/ssr/src/utils/url.ts
```diff
@@ -0,0 +1,97 @@
...
+ * ```
+ */
+export function joinUrlParts(...parts: string[]): string {
+  // Initialize an array with an empty string to always add a leading slash
+  const normalizeParts: string[] = [''];
+
+  for (const part of parts) {
```
dgp1130, line 45: Consider: I think a map and filter might be appropriate here?

```typescript
const normalizedParts = parts.map(() => { /* ... */ })
    .filter((part) => part !== '');
```

The need for the filter is kind of unfortunate and makes this less convenient than it should be, something to think about for you.

packages/angular/ssr/src/utils/url.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @param url - The URL object to process.
+ * @returns A new URL object with `/index.html` removed from the path, if it was present.
+ *
+ * @example
+ * ```typescript
+ * const originalUrl = new URL('http://www.example.com/page/index.html');
+ * const cleanedUrl = stripIndexHtmlFromURL(originalUrl);
+ * console.log(cleanedUrl.href); // Output: 'http://www.example.com/page'
+ * ```
+ */
```
dgp1130, line 86: I appreciate all the detailed JSDoc explanations for these functions with clear examples. It's very helpful for maintainability here!

packages/angular/ssr/src/utils/url.ts
```diff
@@ -0,0 +1,97 @@
...
+export function stripIndexHtmlFromURL(url: URL): URL {
+  if (url.pathname.endsWith('/index.html')) {
+    const modifiedURL = new URL(url);
+    // Remove '/index.html' from the pathname
+    modifiedURL.pathname = modifiedURL.pathname.slice(0, -11);
```
dgp1130, line 91: Suggestion: The `-11` is kind of magical. I think taking the length of a literal would be more clear?

```typescript
modifiedURL.pathname = modifiedURL.pathname.slice(0, -'/index.html'.length);
```

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+   * @returns A nested object where each key is a route segment and each value is an object containing
+   *          metadata and possibly further nested route segments.
+   */
+  toObject(): RouteNodeObjectRepresentation {
+    const result: RouteNodeObjectRepresentation = {};
+
+    /**
+     * Recursively traverses the route tree to build the nested object.
+     */
+    const traverse = (
```
dgp1130, line 170: Consider: Make a generic `traverse` or `walk` function which just returns a generator which iterates over all the nodes in the tree. Then `toObject` can use that to iterate and focus on building the object representation.

This would allow better decoupling between the two problems of 1) walking the `RouteTree` and 2) building the object representation.

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+   *                     metadata and possibly further nested route segments.
+   * @returns A RouteTree constructed from the nested object.
+   */
+  static fromObject(nestedObject: RouteNodeObjectRepresentation): RouteTree {
+    const tree = new RouteTree();
+
+    /**
+     * Recursively traverses the nested object to insert routes into the tree.
+     */
+    const traverse = (currentObj: RouteNodeObjectRepresentation, currentPath: string) => {
```
dgp1130, line 205: Consider: You could also make a separate `walk` function for `RouteNodeObjectRepresentation` for better decoupling here too.

packages/angular/ssr/src/routes/ng-routes.ts
alan-agius4: Yes, that is a goal.

packages/angular/ssr/src/assets.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(private manifest: AngularAppManifest) {}
+
+  /**
+   * Retrieves the content of a server-side asset using its path.
+   *
+   * @param path - The path to the server asset.
+   * @returns A promise that resolves to the asset content as a string.
+   * @throws Error If the asset path is not found in the manifest, an error is thrown.
+   */
+  async getServerAsset(path: string): Promise<string> {
```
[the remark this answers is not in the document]
alan-agius4: Server assets are limited to CSS and HTML.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    const baseHref =
+      injector.get(APP_BASE_HREF, null, { optional: true }) ??
+      injector.get(PlatformLocation).getBaseHrefFromDOM();
+
+    if (router.config.length === 0) {
+      // No routes found in the configuration.
+      yield { route: baseHref };
+    } else {
+      const compiler = injector.get(Compiler);
```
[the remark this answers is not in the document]
alan-agius4: This is not the compiler package but the Compiler class in core.
https://github.com/angular/angular/blob/37bf7dee9917498f47bfd93d5d0bf9ce37f1877c/packages/core/src/linker/compiler.ts#L53

The router is on this https://github.com/angular/angular/blob/37bf7dee9917498f47bfd93d5d0bf9ce37f1877c/packages/router/src/router_config_loader.ts#L47

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -0,0 +1,183 @@
...
+  routes: Route[],
+  compiler: Compiler,
+  parentInjector: Injector,
+  parentRoute: string,
+  baseHref: string,
+): AsyncIterableIterator<RouteResult> {
+  for (const route of routes) {
+    const { path, redirectTo, loadChildren, children } = route;
+
+    if (path === undefined) {
```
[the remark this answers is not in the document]
alan-agius4: Undefined is actually threaded as an empty string. So updated the implementation.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -0,0 +1,183 @@
...
+      yield* traverseRoutesConfig(children, compiler, parentInjector, currentRoutePath, baseHref);
+    }
+
+    if (loadChildren) {
+      // Load and process lazy-loaded child routes.
+      const loadedChildRoutes = await loadChildrenHelper(
+        route,
+        compiler,
+        parentInjector,
+      ).toPromise();
```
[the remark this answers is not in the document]
alan-agius4: Since Angular still supports version RxJs 6, we need to use `toPromise`, `firstValueFrom` is only available from RxJs version 7.

It might be worth discussing with the framework team to determine when we plan to drop support for RxJs 6, especially since it is not used in G3

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -0,0 +1,183 @@
...
+    };
+
+    if (children?.length) {
+      // Recursively process child routes.
+      yield* traverseRoutesConfig(children, compiler, parentInjector, currentRoutePath, baseHref);
+    }
+
+    if (loadChildren) {
+      // Load and process lazy-loaded child routes.
+      const loadedChildRoutes = await loadChildrenHelper(
```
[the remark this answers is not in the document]
alan-agius4: I don’t think it would make a significant difference. The `loadChildrenHelper` function operates on the same thread and does not involve HTTP calls or other asynchronous operations that would require context switching. In this case, parallelizing might actually hurt performance rather than improve it.

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+      // Parameterized route segment, replace it with a wildcard
+      const normalizedSegment = segment[0] === ':' ? '*' : segment;
+      let childNode = node.children?.get(normalizedSegment);
+
+      if (!childNode) {
+        childNode = {
+          segment: normalizedSegment,
+          insertionIndex: -1,
+        };
+        node.children ??= new Map();
```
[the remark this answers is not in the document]
alan-agius4: Originally, my goal was to avoid creating potentially hundreds of redundant `Map` instances in the leaf nodes for large applications, especially when considering i18n.

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+        node.children.set(normalizedSegment, childNode);
```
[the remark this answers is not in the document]
alan-agius4: In `/foo/*/bar` and `/foo/*/baz` the segments `bar` and `baz` are different:

```
/foo/*/bar    //  http://something.com/foo/category/bar
/foo/*/baz    //  http://something.com/foo/product/baz
```

On the other hand,  `/foo/:bar` is the same as `/foo/:baz` the Angular router does not currently handle this as a rejection. Instead, it simply fails to match the second route:

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,224 @@
...
+     * Recursively traverses the route tree to find the best match.
+     */
+    const traverse = (currentNode: RouteTreeNode, remainingSegments: string[]): void => {
+      const { insertionIndex, metadata, children } = currentNode;
+      if (!remainingSegments.length) {
+        // Reached a leaf node, check if it's a better match
+        if (metadata && insertionIndex < lowestIndex) {
+          bestMatchMetadata = metadata;
+          lowestIndex = insertionIndex;
+        }
```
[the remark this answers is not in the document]
alan-agius4: I am a bit confused on how any of the above would return the match based on the order it was inserted.

packages/angular/ssr/src/routes/route-tree.ts
alan-agius4: These will eventually be used to create, store, and restore the routes from the manifest. This approach is designed to avoid building the routes on the first request, which can be critical for performance in edge workers or serverless deployments. Without this, the application would need to be bootstrapped and rendered multiple times for each request, impacting performance.

packages/angular/ssr/src/routes/route-tree.ts
alan-agius4: The `metadata.route` represents the full route up to the current node, provided it is a terminal route, while `segment` indicates the current route segment.

For example:

```json
{
  "segment": "en",
  "child": {
    "segment": "foo",
    "metadata": {
      "route": "en/foo"
    },
    "child": {
      "segment": "*",
      "metadata": {
        "route": "en/foo/:id"
      }
    }
  }
}
```

packages/angular/ssr/src/routes/route-tree.ts
alan-agius4: > Question: How does ordering affect this process? Since `tree.insert` depends on ordering, does `RouteTree.fromObject(tree.toObject())` reliably maintain that order and ensure the same matching behavior?

The test results suggest that order is preserved.

> However, if this is the case, doesn't it imply that `RouteTree.fromObject` depends on the order of properties in children: { /* ... */ }? This seems fragile and could be confusing when using the API.

Yes, the function does rely on order, but isn't order also crucial in router configuration? In both cases, the sequence matters.

Would it be clearer to change the structure from:
```js
children: {
  '': {
    children: {
      first: {
        metadata: { route: '/first' },
      },
      '*': {
        metadata: { route: '/:id' },
      },
      second: {
        metadata: { route: '/second' },
      },
    },
  },
},
```

to 

```js
children: [
  {
    segment: '',
    children: [
      {
        segment: 'first',
        metadata: { route: '/first' },
      },
      {
        segment: '*',
        metadata: { route: '/:id' },
      },
      {
        segment: 'second',
        metadata: { route: '/second' },
      },
    ],
  },
]

```

Or perhaps a flattened version of the above?

```js
routes: [
  {
    path: '/something/first',
    metadata: { route: '/something/first' },
  },
  {
    path: '/something/[:id]',
    metadata: { route: '/something/[:id]' },
  },
  {
    path: '/something/second',
    metadata: { route: '/something/second' },
  },
]

```

packages/angular/ssr/src/utils/ng.ts
alan-agius4: Refactored a bit as technically the input URL is normalized.

packages/angular/ssr/src/utils/url.ts
alan-agius4: I don't think there is every a possibility of having that.

packages/angular/ssr/src/routes/route-tree.ts
alan-agius4: Good call, updated.

packages/angular/ssr/src/routes/route-tree.ts
alan-agius4: Let me think about this following what we decide in https://github.com/angular/angular-cli/pull/28183#discussion_r1712408994

packages/angular/ssr/src/routes/router.ts
alan-agius4: Certainly! Here's a more concise and focused rewrite:

---

During development, `manifest.routes` is `undefined` and the routing table is generated dynamically on the first request. This speeds up the edit-refresh cycle by eliminating the need for prebuilding.

In production, however, the routing table is prebuilt and included in the manifest. While dynamic generation is convenient for development, it's less efficient in production, especially in edge or serverless environments. Prebuilding ensures better performance by preventing the need to bootstrap and render the application multiple times per request.

I understand your concern, but since these are private APIs, the risk of "forgetting to call something" isn't a significant issue.

packages/angular/ssr/src/routes/ng-routes.ts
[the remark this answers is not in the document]
dgp1130: I had a feeling RxJS 6 was the reason. We can ask about the current status but I suspect nothing is meaningfully changed for now.

packages/angular/ssr/src/routes/ng-routes.ts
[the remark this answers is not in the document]
dgp1130: Ok, I'm fine to leave this as is until we end up with performance data which says otherwise.

packages/angular/ssr/src/routes/route-tree.ts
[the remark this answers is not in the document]
dgp1130: I personally wouldn't worry too much about optimizing at this stage until we have a sense of where the performance bottlenecks are, but I'll defer to you on which improvements are likely to be significant enough wins to be worth landing up front.

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,251 @@
...
+   * This allows for easy iteration over the tree's nodes.
+   * @returns A generator that yields the metadata of each node in the tree.
+   */
+  private *traverse(): Generator<RouteTreeNodeMetadata> {
+    function* traverseNode(node: RouteTreeNode): Generator<RouteTreeNodeMetadata> {
+      if (node.metadata) {
+        yield node.metadata;
+      }
+
+      for (const [, childNode] of node.children) {
```
dgp1130, line 200: Consider: If you're ignoring the key anyways, maybe iterate over `node.children.values()`?

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,251 @@
...
+        yield* traverseNode(childNode);
+      }
+    }
+
+    yield* traverseNode(this.root);
```
dgp1130, line 205: Consider: You could use define this function with the node as a parameter which defaults to `this.root` and avoids the extra function definition altogether.

```typescript
private *traverse(node: RouteTreeNode = this.route): Generator<RouteTreeNodeMetadata> {
  if (node.metadata) yield node.metadata;

  for (const [, childNode] of node.children) {
    yield* this.traverseNode(childNode);
  }
}
```

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  static fromObject(value: RouteTreeObjectRepresentation): RouteTree {
+    const tree = new RouteTree();
+
+    for (const { route, ...metadata } of value) {
+      tree.insert(route, metadata);
+    }
+
+    return tree;
+  }
```
dgp1130, line 173: Love how much simpler `traverse` and `traverseByRoute` make these functions. It's so much more intuitive to factor out that iteration logic.

packages/angular/ssr/src/routes/route-tree.ts
alan-agius4: We need to make this serializable, as it will be stored in the manifest, which is generated during the build process. The manifest cannot include non-native import symbols.

Initializing a tree from a non-serializable format would add a bit more complexity.

packages/angular/ssr/src/routes/route-tree.ts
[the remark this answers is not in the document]
alan-agius4: Updated.

packages/angular/ssr/src/routes/route-tree.ts
alan-agius4: Indeed, parameterized routes can get lost, so we could consider adding an additional field to store those parameters. However, using `metadata.route` is slightly more convenient since it eliminates the need to reconstruct the route path.

packages/angular/ssr/src/routes/router.ts
alan-agius4: The URL is needed because Angular requires a URL to configure `ServerPlatformLocation` this is needed as otherwise API requests for relative requests would fail which would cause extraction to fail.

https://github.com/angular/angular/blob/d608b857c689d17a7ffa33bbb510301014d24a17/packages/platform-server/src/location.ts#L51

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -0,0 +1,228 @@
...
+  /** The array of route configurations to process. */
+  routes: Route[];
+  /** The Angular compiler used to compile route modules. */
+  compiler: Compiler;
+  /** The parent injector for lazy-loaded modules. */
+  parentInjector: Injector;
+  /** The parent route path to prefix child routes. */
+  parentRoute: string;
+  /** The base URL of the application used to resolve relative paths. */
+  baseHref: string;
```
atscott, line 73: Is it necessary to carry the baseHref around and attach it to all the route results? Can't the base href logic just be applied to all results where you use them and the `RouteResults` are all just "relative to the app base"?

packages/angular/ssr/src/utils/url.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (part[0] === '/') {
+      start = 1;
+    }
+
+    let end: -1 | undefined;
+    if (part[part.length - 1] === '/') {
+      end = -1;
+    }
+
+    const normalizePart = (start ?? end) ? part.slice(start, end) : part;
```
atscott, line 61:
```suggestion
    let normalizedPart = part;
    if (part[0] === '/') {
      normalizedPart = part.substring(1);
    }
    if (part[part.length - 1] === '/') {
      normalizedPart = part.slice(0, -1);
    }
```
This was a bit hard to follow. I'd suggest just updating the normalized part as you go rather than saving start and end.

packages/angular/ssr/src/utils/url.ts
```diff
@@ -0,0 +1,97 @@
...
+ * joinUrlParts('path/', '/to/resource'); // '/path/to/resource'
+ * joinUrlParts('/path/', 'to/resource'); // '/path/to/resource'
+ * ```
+ */
+export function joinUrlParts(...parts: string[]): string {
+  // Initialize an array with an empty string to always add a leading slash
+  const normalizeParts: string[] = [''];
+
+  for (const part of parts) {
+    if (!part) {
```
atscott, line 46:
```suggestion
    if (part === '') {
```

atscott (approved):

packages/angular/ssr/src/routes/ng-routes.ts
alan-agius4: Great idea. I think it should be doubled. I'll give it a try tomorrow.

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -173,79 +183,122 @@ export class RouteTree {
...
+  /**
+   * Determines if the candidate node is prioritized based on its insertion index compared to the current best match node.
+   * A node with a lower insertion index is prioritized as it was defined earlier.
+   *
+   * @param currentBestMatchNode - The current best match node.
+   * @param candidateNode - The node being evaluated for higher priority based on insertion index.
+   * @returns `true` if the candidate node is prioritized (i.e., has a lower insertion index)
+   * compared to the current best match node, otherwise `false`.
+   */
+  private isPrioritizedByInsertionIndex(
```
dgp1130, line 281: Consider: Rather than treating the current best match specially, consider treating the parameters symmetrically and just returning the best one. If you check for `null` as well you might be able to simplify the usage to just:

```typescript
bestMatchNode = firstByInsertionIndex(bestMatchNode, node.children.get(segment));
bestMatchNode = firstByInsertionIndex(bestMatchNode, node.children.get('*'));
bestMatchNode = firstByInsertionIndex(bestMatchNode, node.children.get('**'));
```

Alternatively you could write this like a compare function, sort an array of the candidates with it, and then pick the first one.

packages/angular/ssr/src/routes/router.ts
```diff
@@ -6,86 +6,89 @@
...
 export class ServerRouter {
-  /** The route tree used for matching URLs to route metadata. */
-  private routeTree: RouteTree | undefined;
+  /**
+   * Creates an instance of the `ServerRouter`.
+   *
+   * @param routeTree - An instance of `RouteTree` that holds the routing information.
+   * The `RouteTree` is used to match request URLs to the appropriate route metadata.
+   */
+  constructor(private readonly routeTree: RouteTree) {}
```
dgp1130, line 28: Nit: Should the constructor be private so consumers go through `ServerRouter.from` instead?

packages/angular/ssr/src/routes/router.ts
dgp1130: Thanks for adding `ServerRouter.from`, I think that greatly simplifies the contract for this class.

Are we just using the `url` from the first request? Do we build different routes based on different URL inputs? I imagine we don't really care about the URL at this stage, would it make more sense to just hard-code a `/` path?

packages/angular/ssr/src/routes/router.ts
```diff
@@ -6,86 +6,89 @@
...
+    if (ServerRouter.#extractionPromise) {
+      return ServerRouter.#extractionPromise;
     }
 
     // Create and store a new promise for the build process.
     // This prevents concurrent builds by re-using the same promise.
-    this.buildPromise = (async () => {
-      this.routeTree = new RouteTree();
-
+    ServerRouter.#extractionPromise = (async () => {
```
dgp1130, line 62: Consider: You can drop the `if (ServerRouter.#extractionPromise)` with a nullish coalescing assignment.

```typescript
ServerRouter.#extractionPromise ??= (async () => { /* ... */})();
return ServerRouter.#extractionPromise;
```

dgp1130 (approved):
Thanks for landing these improvements, looks great!

packages/angular/ssr/src/routes/route-tree.ts
alan-agius4: Good idea

packages/angular/ssr/src/routes/router.ts
alan-agius4: We are actually only using the protocol and host from the url as that is what is needed in this case.

https://github.com/angular/angular-cli/blob/396e83a454199763abd61b70960cfd976f370a04/packages/angular/ssr/src/routes/ng-routes.ts#L175
https://github.com/angular/angular-cli/blob/396e83a454199763abd61b70960cfd976f370a04/packages/angular/ssr/src/routes/ng-routes.ts#L181

PR #28202, opened 2024-08-14
build: remove Bazel typings workaround

alan-agius4:
The workaround is no longer necessary since the bug has been resolved upstream.


alan-agius4:

PR #28215, opened 2024-08-15
fix(@angular/build): improve error message when an unhandled exception occurs during prerendering

### LINKED ISSUE #28212: Report which route failed during pre-rendering.
### Description

While build a pre-rendered app, I'm greeted with the following error : 

```
Exception:
UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<ne>".
ERROR
```

This error makes it hard to investigate the cause as it only happens during pre-rendering (dev mode serve is fine). 

It'd be great to have some details, like for example which route is throwing while being pre-rendered.

alan-agius4:
This change enhances the error messaging when an unhandled exception occurs during the prerendering process. The updated error message provides more context and clarity.

**Previous Behavior**

```
ng b
An unhandled exception occurred: Some error!!!
See "/tmp/ng-S2ABKF/angular-errors.log" for further details.
```

**Updated Behavior:**
```
ng b
Browser bundles
Initial chunk files     | Names               |  Raw size | Estimated transfer size
main-AFPIPGGK.js        | main                | 218.00 kB |                59.48 kB
polyfills-Z2GOM3BN.js   | polyfills           |  35.82 kB |                11.80 kB
styles-5INURTSO.css     | styles              |   0 bytes |                 0 bytes

                        | Initial total       | 253.82 kB |                71.28 kB

Server bundles
Initial chunk files     | Names               |  Raw size
server.mjs              | server              |   1.11 MB |
chunk-HZL5H5M5.mjs      | -                   | 526.77 kB |
polyfills.server.mjs    | polyfills.server    | 269.91 kB |
chunk-GFWAPST7.mjs      | -                   |  19.16 kB |
chunk-5XUXGTUW.mjs      | -                   |   2.55 kB |
render-utils.server.mjs | render-utils.server |   1.46 kB |
main.server.mjs         | main.server         | 149 bytes |

Lazy chunk files        | Names               |  Raw size
chunk-7YC4RJ5P.mjs      | xhr2                |  12.08 kB |

Prerendered 1 static route.
Application bundle generation failed. [4.923 seconds]

✘ [ERROR] An error occurred while prerendering route '/'.

Error: Some error!!!
    at render (node_modules/@angular/build/src/utils/server-rendering/render-worker.js:20:20)
    at /angular-cli/abc/node_modules/piscina/dist/worker.js:146:32
```

Closes #28212

### DIFF
--- a/packages/angular/build/src/utils/server-rendering/prerender.ts
+++ b/packages/angular/build/src/utils/server-rendering/prerender.ts
@@ -81,7 +81,11 @@ export async function prerenderPages(
   }
 
   // Get routes to prerender
-  const { routes: allRoutes, warnings: routesWarnings } = await getAllRoutes(
+  const {
+    routes: allRoutes,
+    warnings: routesWarnings,
+    errors: routesErrors,
+  } = await getAllRoutes(
     workspaceRoot,
     outputFilesForWorker,
     assetsReversed,
@@ -92,11 +96,15 @@ export async function prerenderPages(
     verbose,
   );
 
+  if (routesErrors?.length) {
+    errors.push(...routesErrors);
+  }
+
   if (routesWarnings?.length) {
     warnings.push(...routesWarnings);
   }
 
-  if (allRoutes.size < 1) {
+  if (allRoutes.size < 1 || errors.length > 0) {
     return {
       errors,
       warnings,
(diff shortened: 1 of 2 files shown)


packages/angular/build/src/utils/server-rendering/prerender.ts
```diff
@@ -275,8 +288,12 @@ async function getAllRoutes(
     recordTiming: false,
   });
 
+  const errors: string[] = [];
   const { routes: extractedRoutes, warnings }: RoutersExtractorWorkerResult = await renderWorker
     .run({})
+    .catch((err) => {
+      errors.push(`An error occurred while extracting routes.\n\n${err.stack}`);
```
alan-agius4, line 295: In this case, I opted to display the stack trace instead of just the error message. This can help users identify the exact part of their application that is causing the error.

Alternatively, I can modify this behavior to only display the stack trace when verbose mode is enabled.

clydin (approved):

alan-agius4:

PR #28228, opened 2024-08-19
refactor(@angular/ssr): bundle Critters

alan-agius4:
This commit bundles the Critters library to ensure compatibility with Nodeless environments. Additionally, all licenses for bundled libraries, including Critters, are now included in the package. This helps maintain compliance with open-source license requirements.


dgp1130:
I can see that a decent chunk of this code was just moved around, so I tried not to comment too much on it. Feel free to ignore any pre-existing issues which should be out of scope for this PR.

packages/angular/ssr/third_party/critters/BUILD.bazel
```diff
@@ -0,0 +1,56 @@
...
+# Filter out esbuild metadata files and only copy the necessary files
+genrule(
+    name = "bundled_critters_files",
+    srcs = [
+        ":bundled_critters",
+    ],
+    outs = [
+        "index.js",
+        "index.js.map",
+        "THIRD_PARTY_LICENSES.txt",
```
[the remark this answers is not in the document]
alan-agius4: Wouldn’t a single golden test handle both cases?

packages/angular/ssr/src/common-engine/inline-css-processor.ts
```diff
@@ -6,258 +6,24 @@
...
-  constructor(protected readonly options: InlineCriticalCssProcessorOptions) {}
-
-  async process(
-    html: string,
-    options: InlineCriticalCssProcessOptions,
-  ): Promise<InlineCriticalCssResult> {
-    const critters = new CrittersExtended({ ...this.options, ...options }, this.resourceCache);
-    const content = await critters.process(html);
+  async process(html: string, outputPath: string | undefined): Promise<string> {
+    const critters = new InlineCriticalCssProcessor(outputPath);
```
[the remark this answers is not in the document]
alan-agius4: The output path is not available in the constructor.

We could technically create this when on first call to render.

packages/angular/ssr/third_party/critters/BUILD.bazel
```diff
@@ -0,0 +1,56 @@
...
+    ],
+    cmd = """
+    mkdir -p $(@D)
```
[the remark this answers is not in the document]
alan-agius4: Will take a look, tomorrow.

packages/angular/ssr/src/utils/inline-critical-css.ts
```diff
@@ -0,0 +1,214 @@
...
+      // `embedLinkedStylesheetOverride` will have to be updated.
+      preload: 'media',
+      noscriptFallback: true,
+      inlineFonts: true,
+    });
+
+    // We can't use inheritance to override `embedLinkedStylesheet`, because it's not declared in
+    // the `Critters` .d.ts which means that we can't call the `super` implementation. TS doesn't
+    // allow for `super` to be cast to a different type.
+    this.initialEmbedLinkedStylesheet = this.embedLinkedStylesheet;
```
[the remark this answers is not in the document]
alan-agius4: Updated, although I had to do it in another way since Critters is exported as a default class, and type merging isn't supported in that scenario.

packages/angular/ssr/test/npm_package/BUILD.bazel
```diff
@@ -0,0 +1,35 @@
...
+    name = "test",
+    srcs = [":unit_test_lib"],
+    data = [
+        "THIRD_PARTY_LICENSES.txt.golden",
+        "//packages/angular/ssr:npm_package",
+        "@npm//diff",
+    ],
+)
+
+nodejs_binary(
```
dgp1130, line 25: Suggestion: Would Skylib's `diff_test` be a more out-of-the-box solution here?

https://github.com/bazelbuild/bazel-skylib/blob/main/docs/diff_test_doc.md

packages/angular/ssr/src/utils/inline-critical-css.ts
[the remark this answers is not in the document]
dgp1130: Ah, that's unfortunate, but extending a custom base class isn't too bad. It would be helpful to still include a comment explaining why we need this declaration merge.

packages/angular/ssr/test/npm_package/THIRD_PARTY_LICENSES.txt.golden
```diff
@@ -0,0 +1,350 @@
...
+--------------------------------------------------------------------------------
+Package: chalk
+License: "MIT"
+
+
+--------------------------------------------------------------------------------
+Package: color-convert
+License: "MIT"
+
+Copyright (c) 2011-2016 Heather Arthur <fayearthur@gmail.com>
```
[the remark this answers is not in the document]
alan-agius4: I think we can update this in the future if needed.

packages/angular/ssr/test/npm_package/BUILD.bazel
alan-agius4: I did try to use it, but it always fails with a strange diff error that I am not understanding

```
==================== Test output for //packages/angular/ssr/test/npm_package:critters_license_test:
174c174
< 
---
>  
177c177
< 
---
>  
FAIL: files "packages/angular/ssr/test/npm_package/THIRD_PARTY_LICENSES.txt.golden" and "packages/angular/ssr/third_party/critters/THIRD_PARTY_LICENSES.txt" differ.
```

alan-agius4: Ahh never mind it was spaces.

packages/angular/ssr/test/npm_package/BUILD.bazel
```diff
@@ -0,0 +1,52 @@
...
+    name = "critters_license_test.accept",
+    testonly = True,
+    data = [
+        "THIRD_PARTY_LICENSES.txt.golden",
+        ":unit_test_lib",
+        "//packages/angular/ssr:npm_package",
+    ],
+    entry_point = ":critters-license-test-accept.ts",
```
alan-agius4, line 51: For content I did try to use `copy_file`, but I don’t think there is a way to copy the file to the source tree.

alan-agius4: Actually I probably managed to find a way to do it, by creating a custom macro

https://github.com/bazelbuild/bazel-skylib/blob/5c071b5006bb9799981d04d74a28bdee2f000d4a/docs/private/stardoc_with_diff_test.bzl#L64

packages/angular/ssr/package.json
```diff
@@ -29,7 +29,6 @@
     "@angular/platform-server": "19.0.0-next.1",
     "@angular/router": "19.0.0-next.1",
     "@bazel/runfiles": "^5.8.1",
```
dgp1130, line 31: Question: Would `@bazel/runfiles` make more sense in the root `package.json`? I'm kinda surprised it's not there already.

packages/angular/ssr/test/npm_package/BUILD.bazel
```diff
@@ -8,28 +9,50 @@ ts_library(
...
+    file1 = ":THIRD_PARTY_LICENSES.txt.golden",
+    file2 = ":critters_license_file",
+)
+
+write_file(
+    name = "critters_license_test.accept",
+    out = "critters_license_file_accept.sh",
+    content =
+        [
+            "#!/usr/bin/env bash",
```
dgp1130, line 52: Neat trick with `write_file`, didn't know it could create a binary like that.

One limitation here is that this will only work for Unix developers, not Windows devs. It's probably not a huge deal, especially since anyone could manually copy over the text, even if that is a bit annoying.

Maybe once we get on `@aspect_rules_js` it might be worth making a more reusable and cross-platform `golden_test` macro. I thought `@aspect_bazel_lib` had something here, but it seems like [their `diff_test`](https://docs.aspect.build/rulesets/aspect_bazel_lib/docs/diff_test/) isn't much smarter than Skylib's.

dgp1130 (approved):

packages/angular/ssr/package.json
alan-agius4: It is there: https://github.com/angular/angular-cli/pull/28228/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R86

packages/angular/ssr/test/npm_package/BUILD.bazel
alan-agius4: We could create a macro that creates a CMD instead a bash for Windows.

But as mentioned there are workaround, and likely the update of the golden will be done by one of team which at the moment use Linux or Mac.

PR #28230, opened 2024-08-19
refactor(@angular/ssr): expose private APIs for build system integration and refactor app management

alan-agius4:
- Exposed several utility functions as private APIs to support the integration with the build system.
- Removed `isDevMode` and caching logic from `AngularAppEngine`. This was necessary to better handle updates when using Vite. Instead, `AngularServerApp` is now treated as a singleton to simplify management.
- Switched asset storage from an `Object` to a `Map` in the manifest for improved efficiency and consistency.

This refactor sets the groundwork for seamless wiring with the build system.

### DIFF
--- a/packages/angular/ssr/private_export.ts
+++ b/packages/angular/ssr/private_export.ts
@@ -1,0 +1,18 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+export { ServerRenderContext as ɵServerRenderContext } from './src/render';
+export { getRoutesFromAngularRouterConfig as ɵgetRoutesFromAngularRouterConfig } from './src/routes/ng-routes';
+export {
+  getOrCreateAngularServerApp as ɵgetOrCreateAngularServerApp,
+  destroyAngularServerApp as ɵdestroyAngularServerApp,
+} from './src/app';
+export {
+  setAngularAppManifest as ɵsetAngularAppManifest,
+  setAngularAppEngineManifest as ɵsetAngularAppEngineManifest,
+} from './src/manifest';
--- a/packages/angular/ssr/src/app.ts
+++ b/packages/angular/ssr/src/app.ts
@@ -13,47 +13,22 @@ import { ServerRenderContext, render } from './render';
 import { ServerRouter } from './routes/router';
 
 /**
- * Configuration options for initializing a `AngularServerApp` instance.
- */
-export interface AngularServerAppOptions {
-  /**
-   * Indicates whether the application is in development mode.
-   *
-   * When set to `true`, the application runs in development mode with additional debugging features.
-   */
-  isDevMode?: boolean;
-
-  /**
-   * Optional hooks for customizing the server application's behavior.
-   */
-  hooks?: Hooks;
-}
-
-/**
  * Represents a locale-specific Angular server application managed by the server application engine.
  *
  * The `AngularServerApp` class handles server-side rendering and asset management for a specific locale.
  */
 export class AngularServerApp {
   /**
+   * Hooks for extending or modifying the behavior of the server application.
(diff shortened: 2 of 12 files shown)


packages/angular/ssr/src/app.ts
```diff
@@ -113,3 +97,24 @@ export class AngularServerApp {
...
+  return (angularServerApp ??= new AngularServerApp(options));
+}
+
+/**
+ * Resets the instance of `AngularServerApp` to undefined, effectively
+ * clearing the reference. Use this to recreate the instance.
+ */
+export function resetAngularServerApp(): void {
+  angularServerApp = undefined;
+}
```
AndrewKushnir, line 120: nit: is there a need for any cleanup once reset happens? Another quick nit: may be call this function `destroyAngularServerApp` (so it pairs with "create" a bit better)?

AndrewKushnir (approved):
LGTM 👍

packages/angular/ssr/src/app.ts
alan-agius4: I am fine with destroy, no there is no additional cleanup required once reset happens.

packages/angular/ssr/src/app.ts
```diff
@@ -113,3 +97,27 @@ export class AngularServerApp {
...
+let angularServerApp: AngularServerApp | undefined;
+
+/**
+ * Creates an instance of `AngularServerApp` if it does not already exist.
+ * If an instance exists, it will return the existing one.
+ *
+ * @param options - (Optional) The configuration options for the Angular server app.
+ * @returns The created or existing instance of `AngularServerApp`.
+ */
+export function createAngularServerApp(options?: AngularServerAppOptions): AngularServerApp {
```
dgp1130, line 110: Nit: `create` strongly implies that this function always creates an `AngularServerApp`, when it is actually reused most of the time. Maybe `getAngularServerApps` or `getOrCreateAngularServerApp` might be more clear?

Related: Is it confusing that `options` is ignored and might diverge from a previous invocation?

```typescript
const app1 = createAngularServerApp({ hooks: [someHook] });
const app2 = createAngularServerApp({ hooks: [someOtherHook] }); // `someOtherHook` is actually unused since we'll reuse `app1`.
```

packages/angular/ssr/test/assets_spec.ts
```diff
@@ -15,10 +15,10 @@ describe('ServerAsset', () => {
...
     assetManager = new ServerAssets({
       bootstrap: undefined as never,
-      assets: {
-        'index.server.html': async () => '<html>Index</html>',
-        'index.other.html': async () => '<html>Other</html>',
-      },
+      assets: new Map([
+        ['index.server.html', async () => '<html>Index</html>'],
+        ['index.other.html', async () => '<html>Other</html>'],
+      ]),
```
dgp1130, line 21: Consider: `new Map(Object.entries({ /* ... */ }))` can retain the `key: 'value'` literal structure.

dgp1130 (approved):

packages/angular/ssr/private_export.ts
```diff
@@ -0,0 +1,18 @@
...
+export { ServerRenderContext as ɵServerRenderContext } from './src/render';
+export { getRoutesFromAngularRouterConfig as ɵgetRoutesFromAngularRouterConfig } from './src/routes/ng-routes';
+export {
+  createAngularServerApp as ɵcreateAngularServerApp,
+  destroyAngularServerApp as ɵdestroyAngularServerApp,
+} from './src/app';
+export {
+  setAngularAppManifest as ɵsetAngularAppManifest,
+  setAngularAppEngineManifest as ɵsetAngularAppEngineManifest,
+} from './src/manifest';
```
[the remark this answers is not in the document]
alan-agius4: Yes, as these otherwise they will become “public”.

Example: https://github.com/angular/angular/blob/main/packages/common/src/private_export.ts

packages/angular/ssr/src/app.ts
alan-agius4: Maybe in the case the options differ we should create a new instance?

PR #28231, opened 2024-08-19
refactor(@angular/build): support dev server direct component style serving

clydin:
The Vite-based development server now provides support for serving individual component stylesheets both with and without emulated view encapsulation. This capability is not yet used by the Angular runtime code. The ability to use external stylesheets instead of bundling the style content is an enabling capability primarily for automatic component style HMR features. Additionally, it has potential future benefits for development mode deferred style processing which may reduce the initial build time when using the development server. The application build itself also does not yet generate external stylesheets.

### DIFF
--- a/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
+++ b/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
@@ -9,16 +9,20 @@
 import { lookup as lookupMimeType } from 'mrmime';
 import { extname } from 'node:path';
 import type { Connect, ViteDevServer } from 'vite';
+import { loadEsmModule } from '../../../utils/load-esm';
 import {
   AngularMemoryOutputFiles,
   appendServerConfiguredHeaders,
   pathnameWithoutBasePath,
 } from '../utils';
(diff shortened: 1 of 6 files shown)


packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
```diff
@@ -69,13 +73,52 @@ export function createAngularAssetsMiddleware(
...
+            if (usedIds === undefined) {
+              usedComponentStyles.set(pathname, [componentId]);
+            } else {
+              usedIds.push(componentId);
+            }
+            // Shim the stylesheet if a component ID is provided
+            if (componentId.length > 0) {
+              // Validate component ID
+              // TODO: Determine if this regex is too strict for the APP_ID segment
+              if (/[.-A-Za-z0-9]+-c\d{9}$/.test(componentId)) {
```
alan-agius4, line 92: Technically for the app id portion it can be anything but I think this should be good enough. Consider adding `_` though to the regexp as that is differently used.

alan-agius4 (approved):

PR #28237, opened 2024-08-20
fix(@angular/build): support reading on-disk files during i18n extraction

clydin:
If an application has JavaScript files that are sourced directly from disk, the extraction would previously fail due to the i18n extractor only able to access the in-memory generated JavaScript files. The extractor can now access both memory and disk-based JavaScript files.


alan-agius4 (approved):

alan-agius4:

PR #28238, opened 2024-08-20
fix(@angular-devkit/build-angular): remove outdated browser-esbuild option warning

clydin:
The `resourcesOutputPath` option from the browser builder is supported as of 18.2. The unsupported warning is now removed. The warning logic has also been consolidated now that there are only several warnings left.


alan-agius4 (approved):

alan-agius4:

PR #28252, opened 2024-08-21
feat(@angular/ssr): improve handling of aborted requests in `AngularServerApp`

alan-agius4:
Introduce support for handling request signal abortions in the `AngularServerApp`. This is particularly useful in the development server integration where a 30-second timeout is enforced for requests/responses.


packages/angular/ssr/src/app.ts
```diff
@@ -52,7 +63,44 @@ export class AngularServerApp {
...
+
+  /**
+   * Creates a promise that rejects when the request is aborted.
+   *
+   * @param request - The HTTP request to monitor for abortion.
+   * @returns A promise that never resolves but rejects with an `AbortError` if the request is aborted.
+   */
+  private createAbortPromise(request: Request): Promise<never> {
+    return new Promise<never>((_, reject) => {
+      request.signal.onabort = function () {
```
dgp1130, line 85: Suggestion: Prefer `addEventListener('abort')` so as not to clobber any existing callback on `onabort`. You can also use `once` for slightly better memory management (probably doesn't matter much in practice here).

```typescript
request.signal.addEventListener('abort', () => {
  // ...
}, {once: true});
```

packages/angular/ssr/src/app.ts
```diff
@@ -73,7 +121,60 @@ export class AngularServerApp {
...
+    const { manifest, hooks, assets } = this;
+
+    let html = await assets.getIndexServerHtml();
+    // Skip extra microtask if there are no pre hooks.
+    if (hooks.has('html:transform:pre')) {
+      html = await hooks.run('html:transform:pre', { html });
+    }
+
+    return new Response(
+      await renderAngular(html, manifest.bootstrap(), new URL(request.url), platformProviders),
```
dgp1130, line 175: Question: Can we pass through `request.signal` anywhere in here to abort these operations? Otherwise we're somewhat wasting effort doing rendering work for a request which has already been canceled.

* I think `getIndexServerHtml` is ultimately a dynamic import, so we probably can't abort that.
* We could pass it through to the hook callbacks if we want to add it to the public API there.
* `renderAngular` is probably the most important operation to abort. Can we maybe just stop waiting for stability?

If it's tricky to support this, an easy workaround might just be `if (request.signal.aborted) throw new DOMException('AbortError');` to avoid starting rendering at all. Not as comprehensive, but it's at least something.

alan-agius4: Let talk about this during our sync.

PR #28254, opened 2024-08-21
build: mark server tests as flaky

alan-agius4:
These server tests frequently fail due to timeouts. Despite efforts, the root cause has not yet been identified, so they are being marked as flaky to prevent blocking the pipeline while further investigation continues.


alan-agius4:

PR #28265, opened 2024-08-22
build: clean up redundant `exclude` field from `tsconfig-build.json`

alan-agius4:
The exclude field has been removed from `tsconfig-build.json` as it is redundant. This configuration is used by Bazel, which already specifies the input files explicitly, making the exclude field unnecessary. This change simplifies the configuration without impacting the build process.


clydin (approved):

alan-agius4:

PR #28268, opened 2024-08-22
feat(@schematics/angular): add option to export component as default

### LINKED ISSUE #25023: Feature: add default export option when using a schematic to generate a component
### Command

generate

### Description

When I want to generate a component to use as a lazy loaded component with `loadComponent` I can use the schematic

```sh
ng generate component about --standalone
```

```ts
```ts
import { Component } from '@angular/core';
import { CommonModule } from '@angular/common';

@Component({
  selector: 'app-about',
  standalone: true,
  imports: [CommonModule],
  templateUrl: './about.component.html',
  styleUrls: ['./about.component.css']
})
export class AboutComponent {

}
```

Then when I go to use the component in the route config, I immediately get an error because its not a default export

```ts
[
  { path: 'about', loadComponent: () => import('./about/about.component') }
]
```

Add an option to generate a component with the class as a default export

```sh
ng generate component about --standalone (--export-default or --default-export)
```

```ts
import { Component } from '@angular/core';
import { CommonModule } from '@angular/common';

@Component({
  selector: 'app-about',
  standalone: true,
  imports: [CommonModule],
  templateUrl: './about.component.html',
  styleUrls: ['./about.component.css']
})
export default class AboutComponent {

}
```

Manually update the component after generation

aparzi:
added option for export class in default mode

Closes #25023

## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #25023

## What is the new behavior?

added option for export class in default mode. Example `ng g c foo --export-default=true`. This option will generate a component with default keyword: ... export default class ...

- [ ] Yes
- [X] No

### DIFF
--- a/packages/schematics/angular/component/index_spec.ts
+++ b/packages/schematics/angular/component/index_spec.ts
@@ -496,4 +496,20 @@ describe('Component Schematic', () => {
       await expectAsync(schematicRunner.runSchematic('component', options, appTree)).toBeRejected();
     });
   });
+
+  it('should export the component as default when exportDefault is true', async () => {
+    const options = { ...defaultOptions, exportDefault: true };
+
+    const tree = await schematicRunner.runSchematic('component', options, appTree);
+    const tsContent = tree.readContent('/projects/bar/src/app/foo/foo.component.ts');
+    expect(tsContent).toContain('export default class FooComponent');
+  });
+
+  it('should export the component as a named export when exportDefault is false', async () => {
+    const options = { ...defaultOptions, exportDefault: false };
+
+    const tree = await schematicRunner.runSchematic('component', options, appTree);
+    const tsContent = tree.readContent('/projects/bar/src/app/foo/foo.component.ts');
+    expect(tsContent).toContain('export class FooComponent');
+  });
 });
--- a/packages/schematics/angular/component/schema.json
+++ b/packages/schematics/angular/component/schema.json
@@ -130,6 +130,11 @@
       "type": "boolean",
       "default": false,
       "description": "The declaring NgModule exports this component."
+    },
+    "exportDefault": {
+      "type": "boolean",
+      "default": false,
+      "description": "Use default export for the component instead of a named export."
     }
   },
   "required": ["name", "project"]
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
@@ -19,6 +19,6 @@ import { <% if(changeDetection !== 'Default') { %>ChangeDetectionStrategy, <% }%
   encapsulation: ViewEncapsulation.<%= viewEncapsulation %><% } if (changeDetection !== 'Default') { %>,
   changeDetection: ChangeDetectionStrategy.<%= changeDetection %><% } %>
 })
-export class <%= classify(name) %><%= classify(type) %> {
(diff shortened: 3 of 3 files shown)


packages/schematics/angular/component/index_spec.ts
```diff
@@ -496,4 +496,20 @@ describe('Component Schematic', () => {
...
+    expect(tsContent).toMatch('export default class');
+  });
+
+  it('should use named export', async () => {
+    const options = { ...defaultOptions, exportDefault: false };
+
+    const tree = await schematicRunner.runSchematic('component', options, appTree);
+    const tsContent = tree.readContent('/projects/bar/src/app/foo/foo.component.ts');
+    expect(tsContent).toMatch('export class');
+  });
```
alan-agius4, line 514:
```suggestion
  it('should export the component as default when exportDefault is true', async () => {
    const options = { ...defaultOptions, exportDefault: true };

    const tree = await schematicRunner.runSchematic('component', options, appTree);
    const tsContent = tree.readContent('/projects/bar/src/app/foo/foo.component.ts');
    expect(tsContent).toContain('export default class FooComponent');
  });

  it('should export the component as a named export when exportDefault is false', async () => {
    const options = { ...defaultOptions, exportDefault: false };

    const tree = await schematicRunner.runSchematic('component', options, appTree);
    const tsContent = tree.readContent('/projects/bar/src/app/foo/foo.component.ts');
    expect(tsContent).toContain('export class FooComponent');
  });
```

packages/schematics/angular/component/schema.json
```diff
@@ -130,6 +130,11 @@
       "type": "boolean",
       "default": false,
       "description": "The declaring NgModule exports this component."
+    },
+    "exportDefault": {
+      "type": "boolean",
+      "default": false,
+      "description": "Create a component with default export."
```
alan-agius4, line 137:
```suggestion
      "description": "Use default export for the component instead of a named export."
```

packages/schematics/angular/component/schema.json
```diff
@@ -130,6 +130,11 @@
       "type": "boolean",
       "default": false,
       "description": "The declaring NgModule exports this component."
+    },
+    "exportDefault": {
```
alan-agius4, line 134: Maybe we should name this `exportAsDefault` since it clearly communicates that the component will be exported as a default export.

@clydin/@dgp1130 thoughts?

alan-agius4 (changes requested):
Certainly! Here’s the revised message:

---

Thank you for this. I have a couple of minor suggestions.

The current commit message is unclear and does not accurately describe the feature.

I recommend using:

```
feat(@schematics/angular): add option to export component as default

Introduces option `--export-default` to control whether
the generated component uses a default export instead of a named export.

Closes: #25023
```

packages/schematics/angular/component/schema.json
clydin: While more explicit in regards to intention, I think `--export-as-default` is fairly verbose from a usability perspective.  I do not have a strong opinion either way but would probably prefer `--export-default`.

alan-agius4: In that case let’s go with `--export-default`.

aparzi:
Hi @alan-agius4,
I fixed the suggestions you pointed out to me.

Thanks.

alan-agius4:
@aparzi, it looks like you still have the initial commit which is invalid, can you please squash and remove it?

thanks

aparzi:
@alan-agius4 I fixed the commits everything should be ok. Please recheck the changes as a last check but everything should be ok.

Thanks.

alan-agius4 (approved):

PR #28279, opened 2024-08-26
feat(@angular/ssr): export `AngularAppEngine` as public API

alan-agius4:
Added `AngularAppEngine` to the public API of `@angular/ssr`, allowing users to access it directly for enhanced server-side rendering functionality.

### DIFF
--- a/packages/angular/ssr/public_api.ts
+++ b/packages/angular/ssr/public_api.ts
@@ -7,3 +7,9 @@
  */
 
 export * from './private_export';
+
+export {
+  type AngularServerAppManager,
+  getOrCreateAngularAppEngine,
+  destroyAngularAppEngine,
+} from './src/app-engine';
--- a/goldens/circular-deps/packages.json
+++ b/goldens/circular-deps/packages.json
@@ -23,10 +23,5 @@
   [
     "packages/angular/cli/src/analytics/analytics.ts",
     "packages/angular/cli/src/command-builder/command-module.ts"
-  ],
-  [
-    "packages/angular/ssr/src/app.ts",
-    "packages/angular/ssr/src/assets.ts",
-    "packages/angular/ssr/src/manifest.ts"
   ]
 ]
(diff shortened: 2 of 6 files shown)


packages/angular/ssr/public_api.ts
```diff
@@ -12,4 +12,6 @@ export {
   type CommonEngineOptions,
 } from './src/common-engine/common-engine';
 
+export { AngularAppEngine } from './src/app-engine';
```
alan-agius4, line 15: @AndrewKushnir, I recall you previously suggested that, instead of exporting `AngularAppEngine`, we might consider exporting a creator function like `createAngularAppEngine`. Do you think that approach would be necessary?

AndrewKushnir: As an option, we can also mimic the [`createPlatform`](https://angular.dev/api/core/createPlatform) and [`getPlatform`](https://angular.dev/api/core/getPlatform) APIs:

- the "create" function would only create an instance if no instance exists already and throw an error otherwise
- the "get" function would return the current instance of the engine or `null` if no engine is created yet

WDYT?

alan-agius4: I've updated the pattern to align with the conventions used for `ServerApp`. However, I'm considering whether the return type should be modified to an `AngularAppEngine` interface, ensuring that at least its public methods are included in the API golden.

alan-agius4: IE: https://github.com/angular/angular-cli/pull/28279/commits/588200adf9f22011f879c58180ab1ea76a4ec8a9

AndrewKushnir (approved):
LGTM, thanks Alan 👍 

Just as an idea: we can consider marking those APIs as "developer preview" for now.

alan-agius4:
> Just as an idea: we can consider marking those APIs as "developer preview" for now.

Added.

PR #28303, opened 2024-08-28
feat(@schematics/angular): use TypeScript module preserve option for new projects

clydin:
Newly generated projects will now use the `preserve` value for the `module` option within the TypeScript configuration for the workspace (`tsconfig.json`). This value was added in TypeScript 5.4 and is intended to model the behavior of modern bundlers such as used in the default `application` builder. This option value also has the advantage of automatically enabling `esModuleInterop` and setting `moduleResolution` to `bundler` which are the currently generated values for new projects. This allows explicit use of these options to be removed from the generated file. The `resolveJsonModule` option is also enabled with `preserve` which removes the need for developers to manually add it if JSON files are imported. JSON file imports are fully support including unused property treeshaking with named imports in the `application` builder.
Additional details on the option can be found here: 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#support-for-require-calls-in---moduleresolution-bundler-and---module-preserve


clydin:
There appears to be a TypeScript issue related to `preserve` and `importHelpers`.
This change is blocked pending a resolution.
Ref: https://github.com/microsoft/TypeScript/issues/59789

alan-agius4 (approved):

PR #28311, opened 2024-08-29
build: update to TypeScript 5.6 RC

crisbeto:
Updates the repo to the TypeScript 5.6 RC and narrows down the version to drop support for 5.4, in line with the compiler.

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/action.ts
+++ b/packages/angular_devkit/schematics/src/tree/action.ts
@@ -150,7 +150,7 @@ export class ActionList implements Iterable<Action> {
   get length() {
     return this._actions.length;
   }
-  [Symbol.iterator]() {
+  [Symbol.iterator](): IterableIterator<Action> {
     return this._actions[Symbol.iterator]();
   }
 }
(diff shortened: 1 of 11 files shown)


packages/angular_devkit/schematics/src/tree/action.ts
```diff
@@ -150,7 +150,7 @@ export class ActionList implements Iterable<Action> {
   get length() {
     return this._actions.length;
   }
-  [Symbol.iterator]() {
+  [Symbol.iterator](): IterableIterator<Action> {
```
crisbeto, line 153: This needed an explicit return type or the API extractor would crash.

.ng-dev/format.mjs
```diff
@@ -5,7 +5,12 @@
  */
 export const format = {
   'prettier': {
-    matchers: ['**/*.{mts,cts,ts,mjs,cjs,js,json,yml,yaml,md}'],
+    matchers: [
+      '**/*.{mts,cts,ts,mjs,cjs,js,json,yml,yaml,md}',
+
+      // Don't expect vendored files to be formatted.
+      '!packages/schematics/angular/third_party/**',
```
alan-agius4, line 12: Use `.prettierignore` instead.

https://github.com/angular/angular-cli/blob/5226c535bd94bb6c57aa6305e1acef0a70095478/.prettierignore#L15

alan-agius4 (approved):
One small thing, otherwise looks good.

crisbeto:
Feedback has been addressed.

Issue #28271, opened 2024-08-22
Unit tests produce the error "Some of your tests did a full page reload!" after Chrome update


philmayfield:
### Which @angular/* package(s) are the source of the bug?

core

### Is this a regression?

No

### Description

I don't believe is an Angular issue, but I'm creating this report for clarity and to increase visibility.

Since updating to Chrome ~`128.x` running unit tests produce an error in spite of all tests passing.  To reproduce you can simply create a new project and run `ng test`.

### Please provide the exception or error you saw

```true
Chrome Headless 128.0.0.0 (Mac OS 10.15.7) ERROR
  Some of your tests did a full page reload!
Chrome Headless 128.0.0.0 (Mac OS 10.15.7): Executed 4129 of 4129 SUCCESS (18.521 secs / 17.638 secs)
```

### Please provide the environment you discovered this bug in (run `ng version`)

```true
Angular CLI: 17.3.8
Node: 18.20.4
Package Manager: npm 10.7.0
OS: darwin arm64

Angular: 17.3.11
... animations, common, compiler, compiler-cli, core, forms
... language-service, localize, platform-browser
... platform-browser-dynamic, router

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.1703.8
@angular-devkit/build-angular      17.3.8
@angular-devkit/core               17.3.8
@angular-devkit/schematics         17.3.8
@angular-devkit/schematics-cli     17.3.8
@angular/cdk                       17.3.10
@angular/cli                       17.3.8
@angular/flex-layout               15.0.0-beta.42
@angular/material                  17.3.10
@angular/material-moment-adapter   17.3.10
@schematics/angular                17.3.8
ng-packagr                         17.3.0
rxjs                               7.8.1
typescript                         5.3.3
zone.js                            0.14.7
```

### Anything else?

I tested it out on new Angular 17 and 18 projects using node 18 and 20 and get the same results for all of them.  I was able to verify that upon reverting to Chrome `127.x` the error is not produced.

OPuder:
The same thing happened here after updating Chrome to v128.0.0.0.

![Bildschirmfoto 2024-08-23 um 15 45 57](https://github.com/user-attachments/assets/03671bdc-2d23-49a6-95e2-84561f5d3405)


![Bildschirmfoto 2024-08-23 um 15 44 15](https://github.com/user-attachments/assets/683a706f-49eb-4077-a403-e7b5a535fc6a)

[...]

PR #28315, opened 2024-08-29
fix(@angular-devkit/build-angular): clear context in Karma by default for single run executions

dgp1130:
This works around https://github.com/angular/angular-cli/issues/28271.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/karma/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/karma/index.ts
@@ -98,6 +98,12 @@ export function execute(
 
       karmaOptions.singleRun = singleRun;
 
+      // Workaround https://github.com/angular/angular-cli/issues/28271, by clearing context by default
+      // for single run executions. Not clearing context for multi-run (watched) builds allows the
+      // Jasmine Spec Runner to be visible in the browser after test execution.
+      karmaOptions.client ??= {};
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/builders/karma/index.ts
```diff
@@ -146,6 +146,14 @@ export function execute(
         logger: context.logger,
       };
 
+      // Workaround https://github.com/angular/angular-cli/issues/28271, by clearing context by default
+      // for single run executions.
+      if (karmaOptions.client?.clearContext === undefined) {
```
alan-agius4, line 151: We already set have a `singleRun` variable in the 99
https://github.com/angular/angular-cli/blob/e6c2c4f67b1ca5da546c3818eba035dd336a76a9/packages/angular_devkit/build_angular/src/builders/karma/index.ts#L99

Maybe since this is quite coupled with that option, move it under it?

Also you can simply this to
```ts
karmaOptions.client ??= {};
// leave Jasmine Spec Runner output visible in browser
karmaOptions.client.clearContext ??= singleRun ?? false;
```

alan-agius4 (approved):
LGTM, just one minor NIT

alan-agius4 (approved):

dgp1130:

PR #28329, opened 2024-08-30
build: update `.prettierignore` to remove incorrect exclude pattern

alan-agius4:
The pattern `third_party/github.com/` does not match any files. It should be replaced with `/packages/schematics/angular/third_party/`, which is already included in the list.


clydin (approved):
There used to be something in that location but has since been removed.

PR #28338, opened 2024-09-03
feat(@angular/ssr): add `getHeaders` method to `AngularAppEngine` and `AngularNodeAppEngine` for handling pages static headers

alan-agius4:
**feat(@angular/ssr): Add `getHeaders` Method to `AngularAppEngine` and `AngularNodeAppEngine` for handling pages static headers**

This commit introduces a new `getHeaders` method to the `AngularAppEngine` and `AngularNodeAppEngine` classes, designed to facilitate the retrieval of HTTP headers based on URL pathnames for statically generated (SSG) pages.

```typescript
const angularAppEngine = new AngularNodeAppEngine();

app.use(express.static('dist/browser', {
  setHeaders: (res, path) => {
    // Retrieve headers for the current request
    const headers = angularAppEngine.getHeaders(res.req);

    // Apply the retrieved headers to the response
    for (const { key, value } of headers) {
      res.setHeader(key, value);
    }
  }
}));
```

In this example, the `getHeaders` method is used within an Express application to dynamically set HTTP headers for static files. This ensures that appropriate headers are applied based on the specific request, enhancing the handling of SSG pages.

---

**feat(@angular/ssr): introduce `AngularNodeAppEngine` API for Node.js integration**

Expose a variant of `AngularAppEngine` tailored for Node.js. These additions streamline the process by minimizing boilerplate code and eliminate the need for users to call `createWebRequestFromNodeRequest` before invoking the render method.

### DIFF
--- a/packages/angular/ssr/node/src/app-engine.ts
+++ b/packages/angular/ssr/node/src/app-engine.ts
@@ -1,0 +1,95 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { ɵAngularAppEngine as AngularAppEngine } from '@angular/ssr';
+import type { IncomingMessage } from 'http';
+import { createWebRequestFromNodeRequest } from './request';
+
+/**
+ * Angular server application engine.
+ * Manages Angular server applications (including localized ones) and handles rendering requests.
+
+ * @developerPreview
+ */
+export interface AngularNodeServerAppManager {
+  /**
+   * Renders an HTTP response based on the incoming request using the Angular server application.
+   *
+   * The method processes the incoming request, determines the appropriate route, and prepares the
+   * rendering context to generate a response. If the request URL corresponds to a static file (excluding `/index.html`),
+   * the method returns `null`.
+   *
+   * Example: A request to `https://www.example.com/page/index.html` will render the Angular route
+   * associated with `https://www.example.com/page`.
+   *
+   * @param request - The incoming HTTP request object to be rendered. It can be a `Request` or `IncomingMessage` object.
+   * @param requestContext - Optional additional context for the request, such as metadata or custom settings.
+   * @returns A promise that resolves to a `Response` object, or `null` if the request URL is for a static file
+   * (e.g., `./logo.png`) rather than an application route.
+   */
+  render(request: Request | IncomingMessage, requestContext?: unknown): Promise<Response | null>;
+}
+
+/**
+ * Angular server application engine.
+ * Manages Angular server applications (including localized ones), handles rendering requests,
+ * and optionally transforms index HTML before rendering.
+ */
+export class AngularNodeAppEngine extends AngularAppEngine implements AngularNodeServerAppManager {
+  /**
+   * Renders an HTTP response based on the incoming request using the Angular server application.
+   *
+   * The method processes the incoming request, determines the appropriate route, and prepares the
+   * rendering context to generate a response. If the request URL corresponds to a static file (excluding `/index.html`),
+   * the method returns `null`.
+   *
+   * Example: A request to `https://www.example.com/page/index.html` will render the Angular route
+   * associated with `https://www.example.com/page`.
+   *
+   * @param request - The incoming HTTP request object to be rendered. It can be a `Request` or `IncomingMessage` object.
+   * @param requestContext - Optional additional context for the request, such as metadata or custom settings.
+   * @returns A promise that resolves to a `Response` object, or `null` if the request URL is for a static file
+   * (e.g., `./logo.png`) rather than an application route.
+   */
+  override render(
+    request: Request | IncomingMessage,
+    requestContext?: unknown,
+  ): Promise<Response | null> {
+    const webReq = request instanceof Request ? request : createWebRequestFromNodeRequest(request);
+
+    return super.render(webReq, requestContext);
+  }
+}
+
+let angularNodeAppEngine: AngularNodeAppEngine | undefined;
+
+/**
+ * Retrieves the existing `AngularNodeAppEngine` instance or creates a new one if it doesn't exist.
+ *
+ * This method ensures that a single instance of `AngularNodeAppEngine` is used throughout the application's lifecycle,
+ * promoting efficient resource management. If an instance does not exist, it will be instantiated upon the first call.
+ *
+ * @developerPreview
+ * @returns The existing or newly created instance of `AngularNodeAppEngine`.
+ */
+export function getOrCreateAngularNodeAppEngine(): AngularNodeServerAppManager {
+  return (angularNodeAppEngine ??= new AngularNodeAppEngine());
+}
+
+/**
+ * Destroys the current `AngularNodeAppEngine` instance and releases any associated resources.
+ *
+ * This method resets the reference to the `AngularNodeAppEngine` instance to `undefined`, allowing for the creation
+ * of a new instance on subsequent calls to `getOrCreateAngularNodeAppEngine()`. This is typically used when
+ * reinitializing the server environment or refreshing the application state.
+ *
+ * @developerPreview
+ */
+export function destroyAngularNodeAppEngine(): void {
+  angularNodeAppEngine = undefined;
+}
--- a/packages/angular/ssr/node/src/app-engine.ts
+++ b/packages/angular/ssr/node/src/app-engine.ts
@@ -6,8 +6,8 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import { ɵAngularAppEngine as AngularAppEngine } from '@angular/ssr';
-import type { IncomingMessage } from 'http';
+import { destroyAngularAppEngine, getOrCreateAngularAppEngine } from '@angular/ssr';
+import type { IncomingMessage } from 'node:http';
 import { createWebRequestFromNodeRequest } from './request';
 
 /**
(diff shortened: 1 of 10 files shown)


packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -0,0 +1,95 @@
...
+   */
+  render(request: Request | IncomingMessage, requestContext?: unknown): Promise<Response | null>;
+}
+
+/**
+ * Angular server application engine.
+ * Manages Angular server applications (including localized ones), handles rendering requests,
+ * and optionally transforms index HTML before rendering.
+ */
+export class AngularNodeAppEngine extends AngularAppEngine implements AngularNodeServerAppManager {
```
dgp1130, line 75: Question: Do we / should we try to prevent users from manually constructing an `AngularNodeAppEngine`?

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -0,0 +1,95 @@
...
+   * Example: A request to `https://www.example.com/page/index.html` will render the Angular route
+   * associated with `https://www.example.com/page`.
+   *
+   * @param request - The incoming HTTP request object to be rendered. It can be a `Request` or `IncomingMessage` object.
+   * @param requestContext - Optional additional context for the request, such as metadata or custom settings.
+   * @returns A promise that resolves to a `Response` object, or `null` if the request URL is for a static file
+   * (e.g., `./logo.png`) rather than an application route.
+   */
+  override render(
+    request: Request | IncomingMessage,
```
dgp1130, line 77: Question: Is there a particular reason we want to support `Request` as an input? If you're handling `Request` objects, you can just use `AngularAppEngine` directly right?

Is this so we can leverage `extends AngularAppEngine` here by making the `request` parameter compatible? Should `AngularNodeAppEngine` extend `AngularAppEngine` like this if they have different inputs?

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -0,0 +1,95 @@
...
+    requestContext?: unknown,
+  ): Promise<Response | null> {
+    const webReq = request instanceof Request ? request : createWebRequestFromNodeRequest(request);
+
+    return super.render(webReq, requestContext);
```
dgp1130: Consider: Is it possible to include any unit tests for this? Can we effectively test this with the singleton approach?

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+}
+
+let angularNodeAppEngine: AngularNodeAppEngine | undefined;
```
dgp1130, line 98: Question: I'm a little worried by the amount of singletons we've been introducing. I get that this is somewhat necessary to support injecting the manifest data and maybe it's unavoidable, but I'm wondering if there's any strategies we can use to at least minimize the usage of singletons. I assume this needs to be a singleton because `AngularAppEngine` is.

Does the CLI need knowledge of this many different symbols to inject the correct data? Is there a way we can consolidate onto a different abstraction which the CLI interacts with that doesn't require all the underlying pieces to be singletons as well?

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -41,28 +73,25 @@ export interface AngularNodeServerAppManager {
...
     request: Request | IncomingMessage,
     requestContext?: unknown,
   ): Promise<Response | null> {
-    const webReq = request instanceof Request ? request : createWebRequestFromNodeRequest(request);
+    return super.render(this.toWebRequest(request), requestContext);
+  }
+
+  override getHeaders(
+    request: Request | IncomingMessage,
+  ): Readonly<Record<string, string>> | undefined {
```
dgp1130, line 85: Consider: Can we drop the `| undefined` case here and just return `{}` in this case?

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -41,28 +73,25 @@ export interface AngularNodeServerAppManager {
...
+    if (this.manifest.staticPathsHeaders.size === 0) {
+      return undefined;
+    }
+
+    return super.getHeaders(this.toWebRequest(request));
```
dgp1130, line 90: Optimization: Is it worth caching the `toWebRequest` result? I expect basically every network request would call `getHeaders` and then `render`, meaning the request gets transformed twice. Not sure how expensive that is in practice or how much we care.

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -33,6 +33,38 @@ export interface AngularNodeServerAppManager {
...
+   *
+   *     // Set the retrieved headers on the response
+   *     for (const [key, value] of Object.entries(headers)) {
+   *       res.setHeader(key, value);
+   *     }
+   *   }
+   * }));
+   * ```
+   */
+  getHeaders(request: Request | IncomingMessage): Readonly<Record<string, string>> | undefined;
```
dgp1130, line 67: Consider: I assume this is `Readonly` because we're returning the existing instance from the manifest. Would it be worth cloning that object just to protect it? `readonly` in TS is not great at protecting data, it's trivially easy to cast out of, and I don't think a lot of developers respect it. Is this a strong enough guarantee that it won't be modified?

dgp1130, line 67: Consider: Should we use a `Map` here to reduce the possibility of prototype pollution bugs?

dgp1130, line 67: Question: Is the idea that this is for purely static cases where `render` is never called (ex. `express.static`)? If so, should the name and documentation of this function be clear that:
1. This is limited to prerendered routes _only_.
2. This does not need to be called for non-prerendering use cases. (I assume `render` would automatically set SSR headers when components inject `RESPONSE` and set a header.)

dgp1130 (commented):
Mostly looks good, lots of little comments just to make sure I understand the use case and how this fits into the architecture.

packages/angular/ssr/node/src/app-engine.ts
alan-agius4: `AngularAppEngine` and `AngularNodeAppEngine` don't strictly need to be singletons, but creating multiple instances wouldn't be ideal because there is only one manifest corresponding to a single app.

The `getOrCreateX` and `destroyX` methods were implemented to clearly convey this to users. 

The CLI will only interact directly with the `AngularAppEngine` class.

// @AndrewKushnir, these methods were created based on his feedback.

packages/angular/ssr/node/src/app-engine.ts
alan-agius4: Correct, `Request` was added as a type to support extension.

A potentially alternative approach might look like this:

```typescript
class AngularNodeAppEngine {
  private angularAppEngine = new AngularAppEngine();

  async render(request: IncomingMessage): Promise<Response> {
    return this.angularAppEngine.toWebRequest(request);
  }
}
```

WDYT?

packages/angular/ssr/node/src/app-engine.ts
dgp1130: Is there a significant risk of developers using multiple `AngularAppEngine` classes? For the most part, users are going to copy whatever we scaffold, and any cloud integrations would likely take the effort to reuse something like this. We could potentially also do something like `new AngularAppEngine({ manifest: globalThis.ngSsrManifest })` to manually link this to the manifest singleton rather than forcing it to be a singleton too (maybe it just defaults to that).

packages/angular/ssr/node/src/app-engine.ts
dgp1130: A composition approach LGTM. If these classes don't need to extend each other, I suspect it would be more flexible and maintainable to keep their class hierarchies distinct.

packages/angular/ssr/node/src/app-engine.ts
AndrewKushnir: > Is there a significant risk of developers using multiple AngularAppEngine classes?

I don't think the risk is high, but when it happens, what would be the behavior? There should also be a way to retrieve a reference to a singleton instance itself, in case it was created in a different part of the code (potentially not owned by an app developer). We can potentially come up with different APIs to represent that if the existing ones don't look ergonomic.

packages/angular/ssr/src/app-engine.ts
```diff
@@ -120,6 +130,23 @@ export class AngularAppEngine implements AngularServerAppManager {
...
+   * @returns A record of HTTP headers or `undefined` if there are no headers for the given pathname.
+   */
+  getHeaders(request: Request): Readonly<Record<string, string>> | undefined {
+    if (this.manifest.staticPathsHeaders.size === 0) {
+      return undefined;
+    }
+
+    const { pathname } = stripIndexHtmlFromURL(new URL(request.url));
+
+    return this.manifest.staticPathsHeaders.get(pathname);
```
AndrewKushnir, line 148: Should we consider always returning a map? That can help avoid the need to make extra checks in apps code.

```suggestion
  getHeaders(request: Request): Readonly<Record<string, string>> {
    if (this.manifest.staticPathsHeaders.size === 0) {
      return new Map();
    }

    const { pathname } = stripIndexHtmlFromURL(new URL(request.url));

    return this.manifest.staticPathsHeaders.get(pathname) ?? new Map();
```

packages/angular/ssr/node/src/app-engine.ts
alan-agius4: The `getHeaders` function is only invoked for static files, so the `render` function won’t be executed. That said, converting a request should not be costly.

packages/angular/ssr/node/src/app-engine.ts
alan-agius4: Yes, so users use the `getOrCreate` method

packages/angular/ssr/node/src/app-engine.ts
alan-agius4: > I don't think the risk is high, but when it happens, what would be the behavior? 

Potentially the impact would be some very minor performance impacts.

> In case it was created in a different part of the code (potentially not owned by an app developer).

I'm not entirely clear on the use-case you're describing.

I'm not particularly attached to any specific approach, but we should definitely decide on the desired structure of the public API.

packages/angular/ssr/node/src/app-engine.ts
alan-agius4: > Question: Is the idea that this is for purely static cases where render is never called (ex. express.static)? 

Yes. Updated the docs.

packages/angular/ssr/node/src/app-engine.ts
alan-agius4: What exactly would you like to add unit tests for?

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -0,0 +1,108 @@
...
+   *
+   * @param request - The incoming HTTP request object to be rendered.
+   * @param requestContext - Optional additional context for the request, such as metadata or custom settings.
+   * @returns A promise that resolves to a `Response` object, or `null` if the request URL is for a static file
+   * (e.g., `./logo.png`) rather than an application route.
+   */
+  render(request: IncomingMessage, requestContext?: unknown): Promise<Response | null>;
+
+  /**
+ * Retrieves HTTP headers for a request associated with statically generated (SSG) pages,
```
AndrewKushnir, line 38: nit: I think this line (and lines below) require 1 extra space.

dgp1130 (approved):

AndrewKushnir (approved):

PR #28346, opened 2024-09-04
feat(@angular/ssr): add server routing configuration API

alan-agius4:
This commit introduces a new server routing configuration API, as discussed in RFC https://github.com/angular/angular/discussions/56785. The new API provides several enhancements:

```ts
const serverRoutes: ServerRoute[] = [
  {
    path: '/error',
    renderMode: RenderMode.Server,
    status: 404,
    headers: {
      'Cache-Control': 'no-cache'
    }
  }
];
```

```ts
const serverRoutes: ServerRoute[] = [
  {
    path: '/product/:id',
    renderMode: RenderMode.Prerender,
    async getPrerenderParams() {
      const dataService = inject(ProductService);
      const ids = await dataService.getIds(); // Assuming this returns ['1', '2', '3']
      return ids.map(id => ({ id })); // Generates paths like: [{ id: '1' }, { id: '2' }, { id: '3' }]
    }
  }
];
```

```ts
const serverRoutes: ServerRoute[] = [
  {
    path: '/product/:id',
    renderMode: RenderMode.Prerender,
    fallback: PrerenderFallback.Server, // Can be Server, Client, or None
    async getPrerenderParams() {
    }
  }
];
```

```ts
const serverRoutes: ServerRoute[] = [
  {
    path: '/product/:id',
    renderMode: RenderMode.Server,
  },
  {
    path: '/error',
    renderMode: RenderMode.Client,
  },
  {
    path: '/**',
    renderMode: RenderMode.Prerender,
  },
];
```

These additions aim to provide greater flexibility and control over server-side rendering configurations and prerendering behaviors.


packages/angular/ssr/public_api.ts
```diff
@@ -13,3 +13,15 @@ export {
...
+
+export {
+  type SSGFallback,
+  type RenderMode,
+  type ServerRoute,
+  type ServerRouteAppShell,
+  type ServerRouteCSR,
+  type ServerRouteSSG,
+  type ServerRouteSSR,
+  type ServerRouteCommon,
```
alan-agius4, line 25: I decided to export all of these so they become part of the API goldens. Otherwise, they wouldn't be included in the golden set, which increases the risk of unintentionally introducing breaking changes.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -40,92 +63,97 @@ interface AngularRouterConfigResult {
...
   compiler: Compiler;
-  /** The parent injector for lazy-loaded modules. */
   parentInjector: Injector;
-  /** The parent route path to prefix child routes. */
   parentRoute: string;
-}): AsyncIterableIterator<RouteResult> {
-  const { routes, compiler, parentInjector, parentRoute } = options;
-
+  serverConfigRouteTree: RouteTree<ServerConfigRouteTreeAdditionaMetadata> | undefined;
+  invokeGetPrenderPaths: boolean;
```
eneajaho, line 105:
```suggestion
  invokeGetPrerenderPaths: boolean;
```

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -139,12 +167,71 @@ async function* traverseRoutesConfig(options: {
...
+ *
+ * @param metadata - The metadata associated with the route tree node.
+ * @param parentInjector - The dependency injection container for the parent route.
+ * @param invokeGetPrenderPaths - A flag indicating whether to invoke the `getPrenderPaths` function.
+ * @returns An async iterable iterator that yields route tree node metadata for each SSG path.
+ */
+async function* handleSSGRoute(
+  metadata: ServerConfigRouteTreeNodeMetadata,
+  parentInjector: Injector,
+  invokeGetPrenderPaths: boolean,
```
eneajaho, line 190:
```suggestion
  invokeGetPrerenderPaths: boolean,
```

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * @example
+   * ```typescript
+   * export const serverRouteConfig: ServerRoutes[] = [
+   *   {
+   *     path: '/product/:id',
+   *     remderMode: RenderMode.SSG,
+   *     async getPrenderPaths() {
+   *       const dataService = inject(ProductService);
+   *       const ids = await productService.getIds(); // Assuming this returns ['1', '2', '3']
```
eneajaho, line 105:
```suggestion
   *       const productService = inject(ProductService);
   *       const ids = await productService.getIds(); // Assuming this returns ['1', '2', '3']
```

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * A server route that uses Static Site Generation (SSG) mode.
+ */
+export interface ServerRouteSSG extends Omit<ServerRouteCommon, 'status'> {
+  /** Specifies that the route uses Static Site Generation (SSG) mode. */
+  renderMode: RenderMode.SSG;
+
+  /**
+   * Optional fallback strategy to use if the SSG path is not pre-rendered.
+   * Defaults to `SSGFallback.SSR` if not provided.
```
dgp1130, line 83: Discussion: This is handling the case where a request matches an SSG route with a parameter not returned by `getPrerenderPaths` right?

I think we've discussed before what the right default fallback behavior is but I don't know that we came to any particular consensus. I want to at least consider whether `SSGFallback.None` would be a better default. My reasoning is that I expect most users don't think through this particular edge case or ask themselves whether or not it is ok to allow users to request arbitrary data in an SSG route at runtime. They might not expect that route to ever render on the server.

There are two potential counter points against `SSGFallback.None` I can think of:

First, the devserver ignores `RenderMode.SSG`, does not call `getPrerenderPaths`, and instead renders those routes like `SSR` on demand right? If so, then `SSGFallback.None` actually diverges the runtime behavior between the devserver (route is SSR'd because devserver can't know it's not provided by `getPrerenderPaths`) and production server (doesn't match, likely falls back to 404).

Second, this same case also exists with CSR. If the user visits any page and then performs a client-side navigation to a missing SSG route, the app router doesn't know the route is invalid and will attempt to render it. This diverges the initial load behavior for `SSGFallback.None` routes which don't match (likely 404) on the server but would attempt to render on the client. Client-side and server-side routing are already a bit diverged anyways, so this isn't the end of the world, but I do suspect it is a good idea to at least make sure they agree on what route ultimately gets rendered.

While this problem does technically exist with CSR, it's pretty difficult for users to encounter unless they find a link which actually redirects them to an SSG route with an invalid param. I think you'd need to inject some code in the browser console to get it to do a client-side navigation to such a route without an existing link you can click on. For SSR, this is a bit easier to walk into, given that users can type an invalid SSG param in the browser URL directly. Maybe that's insignificant enough for us to not worry about, but there is a use case difference with SSR which potentially makes this a bigger problem than it has been in the past.

I'm a bit torn between `SSGFallback.None` and `SSGFallback.SSR` as a result. I feel the former is more intuitive to me as a developer, but the latter is more internally consistent. I'm curious to hear what others think? /cc @AndrewKushnir @atscott

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -0,0 +1,146 @@
...
+   *
+   * @returns A Promise resolving to an array where each element is an object with string keys (representing URL parameter names)
+   * and string values (representing the corresponding values for those parameters in the route path).
+   *
+   * @example
+   * ```typescript
+   * export const serverRouteConfig: ServerRoutes[] = [
+   *   {
+   *     path: '/product/:id',
+   *     remderMode: RenderMode.SSG,
```
dgp1130, line 98: Typo: `remderMode`

acastre (approved):

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -0,0 +1,146 @@
...
+   *       const productService = inject(ProductService);
+   *       const ids = await productService.getIds(); // Assuming this returns ['1', '2', '3']
+   *
+   *       return ids.map(id => ({ id })); // Generates paths like: [{ id: '1' }, { id: '2' }, { id: '3' }]
+   *     },
+   *   },
+   * ];
+   * ```
+   */
+  getPrerenderPaths?: () => Promise<Record<string, string>[]>;
```
dgp1130, line 109: Consider: Should this support `Iterable` / `AsyncIterable`? That would allow arrays to work but also potentially allow us to stream rendering (potentially as a future feature, doesn't necessarily need to be in scope for this PR).

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -228,24 +318,36 @@ export async function getRoutesFromAngularRouterConfig(
...
 
+      const serverRoutesConfig = injector.get(SERVER_ROUTES_CONFIG, null, { optional: true });
+      let serverConfigRouteTree: RouteTree<ServerConfigRouteTreeAdditionaMetadata> | undefined;
+
+      if (serverRoutesConfig) {
+        serverConfigRouteTree = new RouteTree();
+        for (const { path, ...metadata } of serverRoutesConfig) {
+          serverConfigRouteTree.insert(path, metadata);
+        }
+      }
```
dgp1130, line 334: Consider: Split this out into another function creating the `RouteTree`. This way it's not generally mutable throughout the function.

```typescript
const serverConfigRouteTree = buildRouteTree(serverRoutesConfig);
```

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -0,0 +1,146 @@
...
+export interface ServerRouteSSR extends ServerRouteCommon {
+  /** Specifies that the route uses Server-Side Rendering (SSR) mode. */
+  renderMode: RenderMode.SSR;
+}
+
+/**
+ * Server route configuration.
+ * @developerPreview
+ */
+export type ServerRoute = ServerRouteAppShell | ServerRouteCSR | ServerRouteSSG | ServerRouteSSR;
```
dgp1130, line 124: Great use of an ADT here, very helpful to get good autocompletion.

packages/angular/ssr/src/routes/route-config.ts
dgp1130, line 109: Question: I don't suppose it would be possible to make `getPrerenderPaths` type dependent on `path` such that it could infer / validate all the required params are given?

I [made an attempt](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgEoHsCukA8AFOMAC2QgA9IQATAZ2RrClAHMA+ZAbwChlfkAHQkQBcyAsQDcPPswhg8UaBGrRxRGgAoAlKIJQ4AWxo4AIlHT8AMhDhUWAZQA2cGkRwA5dAEkQMaPiFWIKkAXy4uMABPfhQzC2tbB2dXAOJSCmVaekYWdgBeMSF0yiyAAwB6ABIOUD8oNAgGENLkAH4GhmRREAgAN2gpCOiUPUNjNWLMugYmEDZkAonyErpS4Wra6EL9AxCqmt8t1EawZuleduOEdCgqAJ2AGmzZ+YAybbGcY4ZWc67CtLLKbIUrVGYsPYbQ71b6nUp-PhtD5GL4nX6IxG6IpA6irdYHOrIs4YklIq43O6jAxPcFzdGkzGcEKDcrlZAAFSIwDoAHcbgBrGgAOiGMQ5JwWyOMcSsNjscycLjcAHJysIYOh0OVIAw1QAjOBQbUnfVwABeyuC4VZHK5dCo6EaIGVYFaXGuIE65mwjUlAG0-twGYJiKJVerNcbdcIDUadWBTRaHgiZHIFEoVFA1DRREGGXwNegw4Xlcn83xY2HY6WUySDWaq+aawyQmW+K2uABdeiEbkwYC+jA+nC0th+zsSIA) at this, but it seems that making two properties of an object dependent on each other is not as straightforward as I expected. Maybe I'm missing something?

This might be too cute to be worth doing anyways, but something we should at least consider.

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -49,13 +50,30 @@ export interface RouteTreeNodeMetadata {
...
+  renderMode?: RenderMode;
 }
 
 /**
  * Represents a node within the route tree structure.
  * Each node corresponds to a route segment and may have associated metadata and child nodes.
+ * The `additionalMetadata` type parameter allows for extending the node metadata with custom data.
  */
-interface RouteTreeNode {
+interface RouteTreeNode<additionalMetadata> {
```
dgp1130, line 76: Style: Should this be PascalCased (`AdditionalMetadata`)?

dgp1130, line 76: Suggestion: Should we do `AdditionalMetadata extends {}` (or `Record<string, unknown>`)? We probably wouldn't want a string or boolean here.

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * A server route that uses Static Site Generation (SSG) mode.
+ */
+export interface ServerRouteSSG extends Omit<ServerRouteCommon, 'status'> {
+  /** Specifies that the route uses Static Site Generation (SSG) mode. */
+  renderMode: RenderMode.SSG;
+
+  /**
+   * Optional fallback strategy to use if the SSG path is not pre-rendered.
```
dgp1130, line 82: Nit: I think a little more detail describing this case would be useful for developers who don't already know what it's trying to describe.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -27,7 +28,29 @@ import { Console } from '../console';
...
+
+/**
+ * An set of HTTP status codes that are considered valid for redirect responses.
+ */
+const VALID_REDIRECT_RESPONSE_CODES = new Set([301, 302, 303, 307, 308]);
+
+/**
+ * Additional metadata for a server configuration route tree.
+ */
+type ServerConfigRouteTreeAdditionaMetadata = Record<string, unknown> & Partial<ServerRoute>;
```
dgp1130, line 47: Typo: `AdditionaMetadata`

dgp1130, line 47: Question: What data are we actually putting into this untyped "additional metadata" value? Is there a stronger abstraction we could define for this or potentially wrap the existing metadata types rather than extending them like this?

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -272,26 +374,30 @@ export async function getRoutesFromAngularRouterConfig(
...
  * If not provided, the default manifest is retrieved using `getAngularAppManifest()`.
- *
+ * @param invokeGetPrerenderPaths - A boolean flag indicating whether to invoke `getPrerenderPaths` for parameterized SSG routes
+ * to handle prerendering paths. Defaults to `false`.
  * @returns A promise that resolves to a populated `RouteTree` containing all extracted routes from the Angular application.
  */
 export async function extractRoutesAndCreateRouteTree(
   url: URL,
   manifest: AngularAppManifest = getAngularAppManifest(),
+  invokeGetPrerenderPaths = false,
```
dgp1130, line 384: Question: When would we choose to set `invokeGetPrerenderPaths = false`? Is that for the devserver use case which uses SSR for SSG routes? If so, I'm wondering if this should be a slightly more generic `allowPrerendering` which controls both:
1. Whether an SSG route is rendered as SSG or SSR.
2. Whether `getPrerenderPaths` is invoked.

Is that a better abstraction here? Is 1. even an issue in this context?

packages/angular/ssr/test/routes/ng-routes_spec.ts
```diff
@@ -0,0 +1,166 @@
...
+            path: 'user/:id/role/:role',
+            renderMode: RenderMode.SSG,
+            async getPrerenderPaths() {
+              return [
+                { id: 'joe', role: 'admin' },
+                { id: 'jane', role: 'writer' },
+              ];
+            },
+          },
+          { path: '/user/:id/role/:role', renderMode: RenderMode.SSG, fallback: SSGFallback.SSR },
```
dgp1130, line 65: Question: Does `fallback` need to be in its own route? Can we not move `fallback: SSGFallback.SSR` to the first route?

packages/angular/ssr/test/routes/router_spec.ts
```diff
@@ -6,8 +6,13 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
+/* eslint-disable import/no-unassigned-import */
+import '@angular/compiler';
```
dgp1130, line 10: Nit: Can you add a comment about the side effect we're relying on here? I assume this is needed for the JIT compiler, but I'm surprised there's a load-bearing top-level side effect we can depend on in this context.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -139,12 +167,71 @@ async function* traverseRoutesConfig(options: {
...
+
+  if (invokeGetPrerenderPaths && currentRoutePath.includes(':')) {
+    if (!getPrerenderPaths) {
+      throw new Error(
+        `The Angular route '${currentRoutePath}' includes parameters, but 'getPrerenderPaths' is missing. ` +
+          `Please define 'getPrerenderPaths' in your server routing configuration.`,
+      );
+    }
+
+    const parameters = await runInInjectionContext(parentInjector, () => getPrerenderPaths());
```
dgp1130, line 206: Question: How does `runInInjectionContext` handle async callbacks? `getPrerenderPaths()` might return a `Promise`. What happens for:

```typescript
function getPrerenderPaths() {
  await someAsyncWork();
  const service = inject(SomeService); // Can we inject after an `await`?
}
```

dgp1130 (commented):
Great stuff here @alan-agius4! Glad to see this all coming together. :grin:

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -27,7 +28,29 @@ import { Console } from '../console';
...
 import { AngularBootstrap, isNgModule } from '../utils/ng';
 import { joinUrlParts } from '../utils/url';
-import { RouteTree } from './route-tree';
+import { RenderMode, SERVER_ROUTES_CONFIG, SSGFallback, ServerRoute } from './route-config';
+import { RouteTree, RouteTreeNodeMetadata } from './route-tree';
+
+/**
+ * Regular expression to match segments preceded by a colon in a string.
+ */
+const URL_PARAMETER_REGEXP = /:([^/]+)(?=\/|$)/g;
```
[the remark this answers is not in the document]
alan-agius4: You are right, the lookahead is not necessary in this case.

packages/angular/ssr/src/routes/ng-routes.ts
alan-agius4: Removing `Record<string, unknown>`

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -40,92 +63,97 @@ interface AngularRouterConfigResult {
...
+      }
+      metadata = { ...metadata, route: currentRoutePath };
+    } else {
+      metadata = {
+        route: currentRoutePath,
+      };
+    }
+
+    // Handle redirects
+    if (typeof redirectTo === 'string') {
```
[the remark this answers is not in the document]
alan-agius4: This is technically valid, for example, when the app is entirely static. 

Redirects handling for SSG will come in a follow up though.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -139,12 +167,71 @@ async function* traverseRoutesConfig(options: {
...
+ * @returns An async iterable iterator that yields route tree node metadata for each SSG path.
+ */
+async function* handleSSGRoute(
+  metadata: ServerConfigRouteTreeNodeMetadata,
+  parentInjector: Injector,
+  invokeGetPrerenderPaths: boolean,
+): AsyncIterableIterator<RouteTreeNodeMetadata> {
+  if (metadata.renderMode !== RenderMode.SSG) {
+    return;
+  }
```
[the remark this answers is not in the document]
alan-agius4: Updated

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -139,12 +167,71 @@ async function* traverseRoutesConfig(options: {
...
+
+  const { route: currentRoutePath, getPrerenderPaths, fallback, ...meta } = metadata;
+
+  if (invokeGetPrerenderPaths && currentRoutePath.includes(':')) {
```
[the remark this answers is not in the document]
alan-agius4: Updated.

packages/angular/ssr/src/routes/ng-routes.ts
alan-agius4: The `inject` needs to happen before any async work.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -139,12 +167,71 @@ async function* traverseRoutesConfig(options: {
...
+      );
+    }
+
+    const parameters = await runInInjectionContext(parentInjector, () => getPrerenderPaths());
+
+    for (const params of parameters) {
+      const routeWithResolvedParams = currentRoutePath.replace(URL_PARAMETER_REGEXP, (match) => {
+        const parameterName = match.slice(1);
+        const value = params[parameterName];
+        if (value === undefined) {
```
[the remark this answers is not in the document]
alan-agius4: Updated to throw an error when the value is not a string

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -139,12 +167,71 @@ async function* traverseRoutesConfig(options: {
...
+          throw new Error(
+            `The 'getPrerenderPaths' function returned an undefined value for parameter '${parameterName}' in route '${currentRoutePath}'.`,
```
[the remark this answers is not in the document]
alan-agius4: We can throw an error, but it is worth failings builds for unused parameters?

packages/angular/ssr/src/routes/ng-routes.ts
alan-agius4: `invokeGetPrerenderPaths` should be `false` in almost all the case, the only time this needs to be `true` is when the this method is being invoked to generate the manifest. 

I prefer the parameter name to be called `invokeGetPrerenderPaths` as it's clear and `allowPrerendering` can be misleading.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -139,12 +167,71 @@ async function* traverseRoutesConfig(options: {
...
+    return;
+  }
+
+  const { route: currentRoutePath, getPrerenderPaths, fallback, ...meta } = metadata;
+
+  if (invokeGetPrerenderPaths && currentRoutePath.includes(':')) {
+    if (!getPrerenderPaths) {
+      throw new Error(
+        `The Angular route '${currentRoutePath}' includes parameters, but 'getPrerenderPaths' is missing. ` +
+          `Please define 'getPrerenderPaths' in your server routing configuration.`,
```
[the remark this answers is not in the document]
alan-agius4: I assume that in 99% of the case if the route has a parameter, it needs to be there.

packages/angular/ssr/test/routes/ng-routes_spec.ts
alan-agius4: We can move it into the route, but this would require special fallback rendering during the `render` phase. By adding an extra route, we can bypass this requirement and handle the fallback as a standard route instead.

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: I'm not entirely clear on the point about streams here. We could include `Iterable` or `AsyncIterable`, but I don't think `AsyncIterable` is necessary in this context. It's also worth noting that, as of now, none of the Angular APIs use `AsyncIterable` as a type.

alan-agius4: I'll experiment a bit tomorrow. However, in your case, the issue is caused by `satisfies Route<string>[];`, which broadens the types.

packages/angular/ssr/src/routes/ng-routes.ts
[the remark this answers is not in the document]
dgp1130: I was mainly thinking this could be indicative of a bug where a user has a route like `/product/:id` and accidentally returns `{ di: 1234 }`. But I think we should already throw if the user doesn't return `id` in that case, so maybe it's not too big a deal? You'd need to add a new parameter and then forget to update the route which is a bit harder mistake to make.

Up to you if it's worth throwing for.

packages/angular/ssr/src/routes/route-config.ts
dgp1130: My thinking was that this would leave open the door to potentially parallelizing the rendering process. For example:

```typescript
async function getPrerenderPaths(): AsyncIterable<Record<string, string>> {
  yield { foo: 'bar' };
  await someTime();
  yield { foo: 'baz' };
}
```

Angular can theoretically start prerendering the `{ foo: 'bar' }` case while waiting for `someTime`. Using `yield` can also just generally be a nice way of authoring this function, even if it doesn't parallelize anything.

packages/angular/ssr/test/routes/ng-routes_spec.ts
dgp1130: I don't think I'm following. How does having this as a separate route make the implementation easier?

From the developer's perspective, it seems weird to have a second route with a `fallback` option (isn't the second route itself the fallback?). How is this different from just using `renderMode: RenderMode.SSR` on the second route? Should it be an error that the user defined `RenderMode.SSG` and a route with two params (`:id` / `:role`), but is missing a `getPrerenderParams` implementation?

packages/angular/ssr/src/routes/ng-routes.ts
dgp1130: Is there any precedent for Angular APIs which are naturally async but run in an injection context? This feels like an easy tripping hazard.

I guess the workaround would be to keep a reference to the injector (or inject before the `await` like you suggested)?

```typescript
async function getPrerenderPaths() {
  const injector = inject(Injector);
  const result = await someAsyncWork();
  if (result === something) {
    injector.inject(SomethingElse);
  }
}
```

/cc @AndrewKushnir is this something we should be concerned about?

packages/angular/ssr/src/app.ts
```diff
@@ -70,18 +79,30 @@ export class AngularServerApp {
...
+  /**
+   * Renders a page based on the provided URL via server-side rendering and returns the corresponding HTTP response.
+   * The rendering process can be interrupted by an abort signal, where the first resolved promise (either from the abort
+   * or the render process) will dictate the outcome.
+   *
+   * @param url - The full URL to be processed and rendered by the server.
+   * @param signal - (Optional) An `AbortSignal` object that allows for the cancellation of the rendering process.
+   * @returns A promise that resolves to the generated HTTP response object, or `null` if no matching route is found.
+   */
+  renderStatic(url: URL, signal?: AbortSignal): Promise<Response | null> {
```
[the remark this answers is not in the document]
alan-agius4: This is a private API, and I chose not to use `prerender` for several reasons:
- The term `prerender` can be misinterpreted as something that should be executed before `render`.
- `renderStatic` clearly indicates that it renders a static version of the page without being tied to SSG usage. For example, it can also be utilized to generate app shells.

IE: `renderStatic` can be called to generate SSG and app-shell pages.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -26,7 +27,29 @@ import { Console } from '../console';
...
 import { AngularBootstrap, isNgModule } from '../utils/ng';
 import { joinUrlParts } from '../utils/url';
-import { RouteTree } from './route-tree';
+import { PrerenderFallback, RenderMode, SERVER_ROUTES_CONFIG, ServerRoute } from './route-config';
+import { RouteTree, RouteTreeNodeMetadata } from './route-tree';
+
+/**
+ * Regular expression to match segments preceded by a colon in a string.
+ */
+const URL_PARAMETER_REGEXP = /:([^/]+)/g;
```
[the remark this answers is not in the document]
alan-agius4: updated

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -138,12 +158,96 @@ async function* traverseRoutesConfig(options: {
...
+): AsyncIterableIterator<RouteTreeNodeMetadata> {
+  if (metadata.renderMode !== RenderMode.Prerender) {
+    throw new Error(
+      `'handleSSGRoute' was called for a route which rendering mode is not prerender.`,
+    );
+  }
+
+  const { route: currentRoutePath, getPrerenderParams, fallback, ...meta } = metadata;
+
+  if (invokeGetPrerenderParams && currentRoutePath.includes('/:')) {
```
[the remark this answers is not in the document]
alan-agius4: >Question: What about /foo:bar routes? Should we generate /fooSomeBarData? Should we error or require the colon to be escaped?

Currently the Angular router treats `/foo:bar` as a valid route thus we should do the same.


> Also should this reuse URL_PARAMETER_REGEXP and just look for any match?

Yes, good idea.

packages/angular/ssr/test/routes/ng-routes_spec.ts
alan-agius4: Ohh I now understand the questions. I made a mistake in the UT.

packages/angular/ssr/src/routes/ng-routes.ts
alan-agius4: Both guards and interceptors run in an injection context.

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: I agree that this is a good approach to authoring the function. However, since most users are likely to use `fetch` or `HttpClient` to populate data for the prerender paths, `AsyncIterable` might be somewhat redundant.

It's also important to note that the route tree is created before prerendering starts.

I’m open to adding `AsyncIterable` as a type, but at this stage, I don’t see a strong enough reason to expand the API. We can definitely reconsider and add it as a feature in the future.

alan-agius4: [This](https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true&noUnusedParameters=true&exactOptionalPropertyTypes=true&noPropertyAccessFromIndexSignature=true#code/JYOwLgpgTgZghgYwgAgEoHsCukA8AFOMAC2QgA9IQATAZ2RrClAHMAaZPKaCa6A4mgD5kAbwBQyScgAOhIgC4OcgNwSpzCGE7deUfkRoAKAJSLO6ALbAaEfFy679Q1QF8xYsAE9pKfQHVgYiwwABkIOCoWAGUAGzgaInw5UgoeWnpGFmEAXjVJfRTKdIADAHoAEhFQGGg0CAYXYuQAfjqGZDMVdy8fJWICKDgLTWgaJOJCtLoGJhBmHLy+knIiumL5SuragaGXCqqQGqg2sEbF1pFkAG0AaWRQJUGLAF1FGZZkF2QAMiWd4cgUDGqHqYEEi06ExWU2QxUq7zme02h1qIIaxXOfzgTxGQJwaLBEKWk2oaw2ByOj12GKkLVE1zuD3+rwys2YnyJIhcqjEMEwIAQYGA6BAyAQXEIEAw2FsBWhpNZWUMiygwQgimluH07H0-1xY38gSIwTCEWicQS4yIghtYlMaDVVp1cj1gINcgCQWwpsic1i8US+htwnEtK4YEwUFFqplrncCBF7RjkDo2Wui3F4UgmoghlDtKksmIigARKV5DB0OhSimwOWAEbYmughtwABeJdYi1p8U8AuQGi09jSfDkRmMom7Bck4cjovz08XyEr6FLK87U6XkkbUFLO43W8Xjbbe-bB8Pn1US5cXdpLmMt6kmclObzm5kclL5eAVGbDHP069v2g7aA4VCjgIJiTheyCzlG0EwVIP5rlWAGISsgzmNIigAOQ4cgpSlMgVDoMgIDoGApBQKqUDvneV6Ljeiz3o+kjPtmapvkuRYKMgZYVlWf5gGhPY0H2CADpooEjnoY5QQuW5wfOdGLshfHrqxW7cu+TF3g+YjPMg8RiomYCqEAA) is what I managed to achieve

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: I don't have a strong preference for the default setting, but I believe SSR is a good default. Next.js also uses a similar "blocking" default, where the page is rendered on the server upon the initial request and then cached for subsequent visits.

If SSR is chosen as a fallback, the same Angular routing rules and guards that apply to CSR are maintained. So, if a product page is accessible via SSR, it will also be accessible via CSR.

packages/angular/ssr/src/app.ts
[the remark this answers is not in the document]
dgp1130: Sounds reasonable, just as long as it's an intentional choice.

packages/angular/ssr/src/routes/route-config.ts
dgp1130: I'm ok with deferring this, that's potentially a feature we could add in a minor anyways, so I don't think we need `Iterable` / `AsyncIterable` to insulate from a future breaking change. Just a potential future optimization we can consider.

dgp1130: Interesting, I was hoping this could be done without changing the syntax but I don't see a better way than what you have.

In terms of value, `ng serve` would not run `getPrerenderPaths` right? If a developer doesn't go out of their way to check that `ng build` passes, they might not notice such an issue until CI runs. I can see some value in getting type checking here.

That said, I'm not totally sure it's worth a no-op `createRoute` function. Is this idea worth considering in the design or something we should just drop for now and come back to later if/when the cost/benefit is more clear?

dgp1130:
Thanks for working through this @alan-agius4! My only remaining concern is the choice of default value for `PrerenderFallback`, the rest I'm happy to defer to you / @AndrewKushnir. Maybe we should have a separate conversation on that point?

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -17,30 +17,40 @@ export enum RenderMode {
...
+  /**
+   * Fallback to Client-Side Rendering (CSR) if the pre-rendered path is not available.
+   * This strategy allows the page to be rendered on the client side.
+   */
+  Client,
 
-  /** No fallback; Angular will not handle the response if the path is not pre-rendered. */
+  /**
+   * No fallback; if the path is not pre-rendered, the server will not handle the request.
+   * This means the application will not provide any response for paths that are not pre-rendered.
```
AndrewKushnir, line 50: It's a bit unclear whether that means 404/500 error codes or something else. What would be the behavior in this case if there was no prerendered HTML page found for this request?

AndrewKushnir (commented):
LGTM, just a few minor comments, mostly related to error message text improvements.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -138,12 +158,101 @@ async function* traverseRoutesConfig(options: {
...
+ * @returns An async iterable iterator that yields route tree node metadata for each SSG path.
+ */
+async function* handleSSGRoute(
+  metadata: ServerConfigRouteTreeNodeMetadata,
+  parentInjector: Injector,
+  invokeGetPrerenderParams: boolean,
+): AsyncIterableIterator<RouteTreeNodeMetadata> {
+  if (metadata.renderMode !== RenderMode.Prerender) {
+    throw new Error(
+      `'handleSSGRoute' was called for a route which rendering mode is not prerender.`,
```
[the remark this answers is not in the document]
alan-agius4: This error is not internal purposes only. This will never be disabled to the users.

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: Currently, setting it to 'None' means Angular will not handle the request and it will instead fallback to the next middleware. 

While we could be more explicit by returning a 404 error ourselves, we currently lack a way to define and serve a custom error page.

Maybe for now, we could send a page with some `404 not found text` and 404 status code?

AndrewKushnir: Discussed offline, the behavior makes sense, we'll update the docs to make it more clear.

AndrewKushnir (approved):

packages/angular/ssr/src/app.ts
```diff
@@ -6,25 +6,34 @@
...
+ * - `RenderMode.Client` maps to an empty string `''` (Client-Side Rendering, no server context needed).
  */
-export enum ServerRenderContext {
-  SSR = 'ssr',
-  SSG = 'ssg',
-  AppShell = 'app-shell',
-}
+const SERVER_CONTEXT_VALUE: Record<RenderMode, string> = {
+  [RenderMode.Prerender]: 'ssg',
+  [RenderMode.Server]: 'ssr',
```
dgp1130, line 33: Consider: If we've renamed `RenderMode` to use `Prerender` and `Server`, should we update these string values as well?

packages/angular/ssr/src/app.ts
```diff
@@ -6,25 +6,34 @@
...
+  [RenderMode.AppShell]: 'app-shell',
+  [RenderMode.Client]: '',
```
dgp1130, line 35: Consider: Should we not provide a `SERVER_CONTEXT` provider if we're doing CSR? If we do want to provide this, any reason not to use `csr` / `Client`?

dgp1130 (approved):

packages/angular/ssr/src/app.ts
alan-agius4: No, as these are the values that Aurora track.

packages/angular/ssr/src/app.ts
alan-agius4: For client this is a noop, its value is never used

PR #28362, opened 2024-09-06
feat(@angular/cli): handle string key/value pairs, e.g. --define

jkrems:
## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features): https://github.com/angular/angular/pull/57699

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The values of `define` can only be updated by changing the Angular config file.

Issue Number: https://github.com/angular/angular-cli/issues/4318 (arguably)

## What is the new behavior?

Defines can now be introduced or changed by passing them to the CLI, e.g.:

```
ng build --define "BUILD_TIME=$(date)"
```

- [ ] Yes
- [x] No

## Other information

The PR also moves `addSchemaOptionsToCommand` to be colocated with `parseJsonSchemaToOptions`. The two functions are tightly coupled and it makes unit testing the schema-to-yargs flow a lot easier.

### DIFF
--- a/packages/angular/cli/BUILD.bazel
+++ b/packages/angular/cli/BUILD.bazel
@@ -147,6 +147,7 @@ ts_library(
         "//packages/angular_devkit/schematics",
         "//packages/angular_devkit/schematics/testing",
         "@npm//@types/semver",
+        "@npm//@types/yargs",
     ],
 )
 
--- a/packages/angular/cli/src/command-builder/utilities/json-schema.ts
+++ b/packages/angular/cli/src/command-builder/utilities/json-schema.ts
@@ -6,8 +6,8 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import { json } from '@angular-devkit/core';
-import yargs from 'yargs';
+import { json, strings } from '@angular-devkit/core';
+import yargs, { Arguments, Argv, PositionalOptions, Options as YargsOptions } from 'yargs';
 
 /**
  * An option description.
@@ -43,6 +43,55 @@ export interface Option extends yargs.Options {
    * If this is falsey, do not report this option.
    */
   userAnalytics?: string;
+
+  /**
+   * Type of the values in a key/value pair field.
+   */
+  itemValueType?: 'string';
+}
+
+function coerceToStringMap(
+  dashedName: string,
+  value: (string | undefined)[],
+): Record<string, string> | Promise<never> {
+  const stringMap: Record<string, string> = {};
+  for (const pair of value) {
+    // This happens when the flag isn't passed at all.
+    if (pair === undefined) {
+      continue;
+    }
+
+    const eqIdx = pair.indexOf('=');
+    if (eqIdx === -1) {
+      // TODO: Remove workaround once yargs properly handles thrown errors from coerce.
+      // Right now these sometimes end up as uncaught exceptions instead of proper validation
+      // errors with usage output.
+      return Promise.reject(
+        new Error(
+          `Invalid value for argument: ${dashedName}, Given: '${pair}', Expected key=value pair`,
+        ),
+      );
+    }
+    const key = pair.slice(0, eqIdx);
+    const value = pair.slice(eqIdx + 1);
+    stringMap[key] = value;
+  }
+
+  return stringMap;
+}
+
+function isStringMap(node: json.JsonObject): boolean {
+  // Exclude fields with more specific kinds of properties.
+  if (node.properties || node.patternProperties) {
+    return false;
+  }
+
+  // Restrict to additionalProperties with string values.
+  return (
+    json.isJsonObject(node.additionalProperties) &&
+    !node.additionalProperties.enum &&
+    node.additionalProperties.type === 'string'
+  );
 }
 
 export async function parseJsonSchemaToOptions(
@@ -106,10 +155,13 @@ export async function parseJsonSchemaToOptions(
 
           return false;
 
+        case 'object':
+          return isStringMap(current);
+
         default:
           return false;
       }
-    }) as ('string' | 'number' | 'boolean' | 'array')[];
+    }) as ('string' | 'number' | 'boolean' | 'array' | 'object')[];
 
     if (types.length == 0) {
       // This means it's not usable on the command line. e.g. an Object.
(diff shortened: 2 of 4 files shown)


jkrems:
Unfortunately it seems like this is currently still not properly triggering usage errors when invalid values are passed. Need to dig a bit more.

packages/angular/cli/BUILD.bazel
```diff
@@ -147,6 +147,7 @@ ts_library(
         "//packages/angular_devkit/schematics",
```
alan-agius4, line 147: Drive by comment. I have not actually checked the code.
But have we considered allowing an array of strings as an input and normalizing the options in the builder?

IE: 
In https://github.com/angular/angular-cli/blob/d87822c1cef972026e7ea52be7728e7076478911/packages/angular/build/src/builders/application/schema.json#L223
```
    "define": {
      "description": "Defines global identifiers that will be replaced with a specified constant value when found in any JavaScript or TypeScript code including libraries. The value will be used directly. String values must be put in quotes. Identifiers within Angular metadata such as Component Decorators will not be replaced.",
      "oneOf": [
        {
          "description": "Defines global identifiers",
          "type": "array",
          "items": {
            "type": "string",
            "pattern": "^.+=.+$"
          },
          "default": []
        },
        {
          "type": "object",
          "additionalProperties": {
            "type": "string"
          }
        }
      ]
    },
```

And in https://github.com/angular/angular-cli/blob/d87822c1cef972026e7ea52be7728e7076478911/packages/angular/build/src/builders/application/options.ts#L218

```ts
  let define: any;
  if (Array.isArray(options.define)) {
    define = {};
    for (const keyValue of options.define) {
      const [key, value] = keyValue.split('=')
      define[key] = value
    }
  } else {
    define = options.define;
  }

```

jkrems: I assume the question there is if we would want the alternate syntax in the config file as well? I tend to be hesitant when it comes to creating that kind of branching value space since it can make both documentation and automation harder down the line. Case in point: It may introduce `.split` bugs if some other code tries to interpret the key/value pairs in the array and doesn't consider string values that contain `=` inside of them.

packages/angular/cli/src/command-builder/utilities/json-schema.ts
```diff
@@ -43,6 +43,54 @@ export interface Option extends yargs.Options {
...
+  const stringMap: Record<string, string> = {};
+  for (const pair of value) {
+    // This happens when the flag isn't passed at all.
+    if (pair === undefined) {
+      continue;
+    }
+
+    const eqIdx = pair.indexOf('=');
+    if (eqIdx === -1) {
+      throw new Error(
```
alan-agius4, line 63: NIT:

There is a bug in Yargs' `coerce` function when an exception is thrown. According to the documentation (https://yargs.js.org/docs/#api-reference-coercekey-fn), the exception should be treated as a validation error, but it is currently being thrown as an unhandled error.

**Workaround:** You can address this by returning the error wrapped in a `Promise`:

```javascript
return Promise.reject(`Invalid value for argument: ${dashedName}, Given: '${pair}', Expected key=value pair`);
```

Alternatively, you can throw the error and mark the method as `async`:

```javascript
throw `Invalid value for argument: ${dashedName}, Given: '${pair}', Expected key=value pair`;
```

alan-agius4 (approved):
LGTM, just a couple of NITs.

packages/angular/cli/src/command-builder/utilities/json-schema.ts
jkrems: Thanks! I thought I had tested this but I definitely saw this behavior with an earlier revision of this change. Doing the reject workaround.

Do you have an upstream issue reference for this? I only found an issue that was marked as fixed (https://github.com/yargs/yargs/issues/755).

packages/angular/cli/src/command-builder/utilities/json-schema.ts
```diff
@@ -43,6 +43,54 @@ export interface Option extends yargs.Options {
...
    */
   userAnalytics?: string;
+
+  /**
+   * Type of the values in a key/value pair field.
+   */
+  itemValueType?: 'string';
+}
+
+function coerceToStringMap(dashedName: string, value: (string | undefined)[]) {
```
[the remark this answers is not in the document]
jkrems: Good idea, done!

packages/angular/cli/src/command-builder/utilities/json-schema.ts
```diff
@@ -43,6 +43,62 @@ export interface Option extends yargs.Options {
...
+  /**
+   * Type of the values in a key/value pair field.
+   */
+  itemValueType?: 'string';
+}
+
+function coerceToStringMap(
+  dashedName: string,
+  value: (string | undefined)[],
+): Record<string, string> | Promise<Record<string, string>> {
```
alan-agius4, line 56: Consider 

```suggestion
): Record<string, string> | Promise<never> {
```

packages/angular/cli/src/command-builder/utilities/json-schema.ts
```diff
@@ -43,6 +43,62 @@ export interface Option extends yargs.Options {
...
+    }
+    const key = pair.slice(0, eqIdx);
+    const value = pair.slice(eqIdx + 1);
+    stringMap[key] = value;
+  }
+
+  return stringMap;
+}
+
+function isStringMap(node: json.JsonObject) {
```
alan-agius4, line 83:
```suggestion
function isStringMap(node: json.JsonObject): boolean {
```

packages/angular/cli/src/command-builder/utilities/json-schema.ts
jkrems: I'll try to see if I can find a quick way to repro but that's realistically independent of this change.

alan-agius4 (approved):
LGTM, two minor optional nits.

packages/angular/cli/src/command-builder/utilities/json-schema.ts
alan-agius4: Might also want to consider combining the multiple logical statements.

jkrems: Grouped into "exclude other kinds of objects" and "define desired kind of objects".

jkrems:
Thanks for the review!

alan-agius4 (approved):

PR #28386, opened 2024-09-10
refactor: Add `lines-around-comment` rule

alan-agius4:
I often struggle with spacing around block comments, so I've decided to add the `lines-around-comment` lint rule to help manage this.

For more details, see the https://eslint.style/rules/js/lines-around-comment

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,8 +33,24 @@
     "project": "tsconfig.json",
     "sourceType": "module"
   },
-  "plugins": ["eslint-plugin-import", "header", "@typescript-eslint"],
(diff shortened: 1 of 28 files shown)


dgp1130 (approved):
Thanks for this! It's a minor thing but it definitely helps with readability for me.

alan-agius4:
@dgp1130, unfortunately it’s not working as I am getting `TypeError: Cannot read properties of null (reading 'type')`

Will try to find the root cause and open an issue upstream.

alan-agius4:
Looks like this setting conflicts with Prettier in `Interfaces`. Will take another stab at this tomorrow.

PR #28404, opened 2024-09-12
refactor(ssr): use `ReadonlyMap` instead of `Readonly` mapped type

JoostK:
The `Readonly` mapped type only introduces the `readonly` modifier for each property, but `Map` has mutating methods like `clear`, `delete` and `set` that would still remain usable. This commit switches those usages over to the `ReadonlyMap` type which doesn't have any of the mutating methods.

### DIFF
--- a/goldens/public-api/angular/ssr/index.api.md
+++ b/goldens/public-api/angular/ssr/index.api.md
@@ -6,7 +6,7 @@
 
 // @public
 export class AngularAppEngine {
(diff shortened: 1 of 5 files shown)


JoostK:
> LGTM, can you please change the commit message scope to `@angular/ssr`?

Yep, done, I was waiting to see if all other tests were green :smile:

alan-agius4 (approved):

PR #28405, opened 2024-09-12
fix(@angular/ssr): manage unhandled errors in zoneless applications

alan-agius4:
Implement the `attachNodeGlobalErrorHandlers` function to handle 'unhandledRejection' and 'uncaughtException' events in Node.js. This function logs errors to the console, preventing unhandled errors from crashing the server. It is particularly useful for zoneless apps, ensuring error handling without relying on zones.

Closes https://github.com/angular/angular/issues/58123


packages/angular/ssr/node/src/errors.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+/**
+ * Attaches listeners to the Node.js process to capture and handle unhandled rejections and uncaught exceptions.
+ * Captured errors are logged to the console. This function logs errors to the console, preventing unhandled errors
+ * from crashing the server. It is particularly useful for Zoneless apps, ensuring error handling without relying on Zone.js.
+ *
+ * @developerPreview
+ */
+export function attachNodeErrorHandlers(): void {
```
AndrewKushnir, line 16: nit:
```suggestion
export function attachNodeGlobalErrorHandlers(): void {
```

AndrewKushnir (approved):

packages/angular/ssr/node/src/errors.ts
```diff
@@ -0,0 +1,31 @@
...
+    return;
+  }
+
+  gThis.ngAttachNodeGlobalErrorHandlersCalled = true;
+
+  process
+    // eslint-disable-next-line no-console
+    .on('unhandledRejection', (error) => console.error('unhandledRejection', error))
+    // eslint-disable-next-line no-console
+    .on('uncaughtException', (error) => console.error('uncaughtException', error));
```
jkrems, line 30: It's worth noting that this leaves the process in an undefined state which _may_ cause resource leaks and other unexpected behavior. See also: https://nodejs.org/api/process.html#warning-using-uncaughtexception-correctly

Could this behavior be made optional and come with "buyer beware"?

atscott: +1 - discussions with @alxhub and @pkozlowski-opensource before going on leave were around _not_ doing this and needing the framework to handle application errors instead. 

This handler makes a choice about errors coming from more than just the framework/application.

alan-agius4: I totally agree, I wasn’t aware that things change since we last talked.

I definitely agree that the FW should handle errors.

alxhub: Agree as well - in general I would favor letting a process crash and restart if an error escapes the individual request scope. At that point, you can't really guarantee it can safely continue processing requests.

I wonder if there's potential for some kind of graceful shutdown though, where an uncaught error doesn't terminate the process immediately, but stops accepting new requests and gives some timeout for existing ones to shut down.

jkrems: Happy to chat more about options here. Unfortunately, there is (or at least used to be) no easy way to stop accepting new requests (vs. accepting new connections). IIRC the best strategy was to check on the end of requests if there's other concurrent requests on the same connection and using that to separate connections to proxies/load balancers.

alxhub: Should Angular servers implement `/healthz`? :D

jkrems: It's a really interesting open question because it says a lot about what this server is. Actually curious about what (if anything) other backend-for-frontend-in-a-box solutions do for this. Wouldn't be surprised if the answer is "not much" since it doesn't matter for most apps if a couple of concurrent requests end up with preventable 500s when something goes wrong..?

alan-agius4: Following the discussions we had recently. I re-opened the PR.

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -22,6 +23,10 @@ import { createWebRequestFromNodeRequest } from './request';
 export class AngularNodeAppEngine {
   private readonly angularAppEngine = new AngularAppEngine();
 
+  constructor() {
+    attachNodeGlobalErrorHandlers();
```
alan-agius4, line 27: As an alterative, we can export this a public API, and let the users invoke it manually in the `server.ts`

atscott (approved):

PR #28429, opened 2024-09-17
refactor(@angular-devkit/core): fix up internal typings

crisbeto:
Resolves some type errors that showed up internally.


alan-agius4:
Looks like the API Extractor doesn't like the new Symbol
```
Analysis will use the bundled TypeScript version 5.4.2
Internal Error: Unable to follow symbol for "MapIterator"
```

JeanMeche:
Fwiw we dropped support for 5.4 in v19 angular/angular#57577

crisbeto:
Yeah I'll look for a different solution since the latest version of the API extractor is still on 5.4.

crisbeto:
I've worked around it using a resolution until the API extractor gets updated to TS 5.6.

clydin (approved):

PR #28432, opened 2024-09-17
perf(@angular/cli): enable Node.js compile code cache when available

clydin:
The Angular CLI will now enable the Node.js compile cache when available for use. Node.js v22.8 and higher currently provide support for this feature. The compile cache stores the v8 intermediate forms of JavaScript code for the Angular CLI itself. This provides a speed up to initialization on subsequent uses the Angular CLI. The Node.js cache is stored in a temporary directory in a globally accessible location so that all Node.js instances of a compatible version can share the cache. The code cache can be disabled if preferred via `NODE_DISABLE_COMPILE_CACHE=1`.

Based on initial profiling, this change provides an ~6% production build time improvement for a newly generated project once the cache is available.

```
Benchmark 1: NODE_DISABLE_COMPILE_CACHE=1 node ./node_modules/.bin/ng build
  Time (mean ± σ):      2.617 s ±  0.016 s    [User: 3.795 s, System: 1.284 s]
  Range (min … max):    2.597 s …  2.640 s    10 runs

Benchmark 2: node ./node_modules/.bin/ng build
  Time (mean ± σ):      2.475 s ±  0.017 s    [User: 3.555 s, System: 1.354 s]
  Range (min … max):    2.454 s …  2.510 s    10 runs

Summary
  node ./node_modules/.bin/ng build ran
    1.06 ± 0.01 times faster than NODE_DISABLE_COMPILE_CACHE=1 node ./node_modules/.bin/ng build
```

### DIFF
--- a/packages/angular/build/src/utils/worker-pool.ts
+++ b/packages/angular/build/src/utils/worker-pool.ts
@@ -1,0 +1,44 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { getCompileCacheDir } from 'node:module';
+import { Piscina } from 'piscina';
+
+export type WorkerPoolOptions = ConstructorParameters<typeof Piscina>[0];
+
+export class WorkerPool extends Piscina {
+  constructor(options: WorkerPoolOptions) {
+    const piscinaOptions: WorkerPoolOptions = {
+      minThreads: 1,
+      idleTimeout: 1000,
+      // Web containers do not support transferable objects with receiveOnMessagePort which
+      // is used when the Atomics based wait loop is enable.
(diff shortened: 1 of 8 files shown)


alan-agius4 (approved):

packages/angular/build/src/utils/worker-pool.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { getCompileCacheDir } from 'node:module';
+import { Piscina } from 'piscina';
+
+export type WorkerPoolOptions = ConstructorParameters<typeof Piscina>[0];
+
+export class WorkerPool extends Piscina {
```
alfaproject, line 14: Would a build function that builds a Piscina instance this way be simpler?

clydin: It will be further abstracted in the future.

PR #28433, opened 2024-09-17
fix(@angular/build): support HTTP HEAD requests for virtual output files

clydin:
When using the development server, HTTP HEAD requests will now correctly respond for the virtual output files generated from the Angular build system.  Previously Vite only handled GET requests for the files. While HEAD requests are not common in development workflows, it can be needed in more complex cases with additional servers/proxies/etc. during development.


alan-agius4 (approved):

clydin:

PR #28456, opened 2024-09-19
fix(@angular/build): allow missing HTML file request to fallback to index

clydin:
If a HTTP request is made to the development server that explicitly requests an HTML file (i.e., `/abc.html`), the development server will now attempt to fallback to the root `index.html` file if the requested HTML file does not exist. Since this may indicate a defect or other application misconfiguration such as a missing asset, a warning will also be issued in the console during development to notify the developer that something may be wrong.


alan-agius4 (approved):

clydin:

PR #28463, opened 2024-09-20
feat(@angular/build): utilize `ssr.entry` in Vite dev-server when available

alan-agius4:
**feat(@angular/build): utilize `ssr.entry` in Vite dev-server when available**
When `ssr.entry` (server.ts) is defined, Vite will now use it in the dev-server. This allows API and routes defined in `server.ts` to be accessible during development. This feature requires the new `@angular/ssr` APIs, which are currently in developer preview.

**feat(@angular/ssr): add `createRequestHandler` and `createNodeRequestHandler `utilities**

Introduced the `createRequestHandler` and `createNodeRequestHandler` utilities to expose middleware functions from the `server.ts` entry point for use with Vite.
This provides flexibility in integrating different server frameworks, including Express, Hono, and Fastify, with Angular SSR.

Examples:

**Express**
```ts
export default createNodeRequestHandler(app);
```

**Nest.js**
```ts
const app = await NestFactory.create(AppModule);
export default createNodeRequestHandler(app);
```

**Hono**
```ts
const app = new Hono();
export default createRequestHandler(app.fetch);
```

**Fastify**
```ts
export default createNodeRequestHandler(async (req, res) => {
  await app.ready();
  app.server.emit('request', req, res);
});
```


clydin (approved):
We should consider consolidating some of the E2E test logic here at some point.

Otherwise, LGTM.

PR #28469, opened 2024-09-23
feat(@schematics/angular): integrate `withEventReplay()` in `provideClientHydration` for new SSR apps

alan-agius4:
Event Replay is now stable (see https://github.com/angular/angular/pull/57895). This commit adds `withEventReplay()` to the `provideClientHydration()` call in newly created Angular applications with SSR.


AndrewKushnir (approved):
@alan-agius4 thanks for the help! 👍

PR #28471, opened 2024-09-23
 feat(@angular/build): utilize `ssr.entry` during prerendering to enable access to local API routes     

alan-agius4:
The server.ts file is now utilized during prerendering, allowing access to locally defined API routes for improved data fetching and rendering.

### DIFF
--- a/packages/angular/build/src/utils/server-rendering/launch-server.ts
+++ b/packages/angular/build/src/utils/server-rendering/launch-server.ts
@@ -1,0 +1,64 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import assert from 'node:assert';
+import { createServer } from 'node:http';
+import { loadEsmModule } from '../load-esm';
+import { loadEsmModuleFromMemory } from './load-esm-from-memory';
+import { isSsrNodeRequestHandler, isSsrRequestHandler } from './utils';
+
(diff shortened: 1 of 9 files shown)


packages/angular/build/src/utils/server-rendering/launch-server.ts
```diff
@@ -0,0 +1,63 @@
...
+    })().catch((e) => {
+      res.statusCode = 500;
+      res.end('Internal Server Error.');
+      throw e;
+    });
+  });
+
+  server.unref();
+
+  await new Promise<void>((resolve) => server.listen(0, resolve));
```
clydin, line 56: This should probably specify `localhost` for the host option here.  It defaults to all interfaces otherwise.

packages/angular/build/src/utils/server-rendering/launch-server.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (!isSsrNodeRequestHandler(handler) && !isSsrRequestHandler(handler)) {
+    return DEFAULT_URL;
+  }
+
+  const server = createServer((req, res) => {
+    (async () => {
+      // handle request
+      if (isSsrNodeRequestHandler(handler)) {
+        await handler(req, res, (e) => {
+          throw e;
```
clydin, line 36: How does the error thrown here propagate?  Does it affect the whole dev-server?

clydin (approved):

packages/angular/build/src/utils/server-rendering/launch-server.ts
alan-agius4: Update to log the error instead.

PR #28472, opened 2024-09-23
test: unlink symlink in ng add base E2E test

clydin:
Instead of attempting to delete all files within the directory, the created symlink can instead be removed on its own. This should prevent odd Windows failures in CI.


clydin:

PR #28501, opened 2024-09-26
perf(@angular/ssr): cache resolved entry-points

alan-agius4:
Improve request throughput by 4.34% and reduce latency by caching resolved entry-points.


AndrewKushnir (approved):
👍

PR #28506, opened 2024-09-26
perf(@angular/ssr): prevent potential stampede in entry-points cache

jkrems:
If multiple concurrent requests hit `getEntryPointExports`, all of
them would previously see the cache miss for entry point. With this
change, only the first request will and the others can leverage the
cache.

This can be important when instances are added to a pool under high
traffic.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/angular/ssr/src/app-engine.ts
+++ b/packages/angular/ssr/src/app-engine.ts
@@ -50,7 +50,7 @@ export class AngularAppEngine {
   /**
    * A cache that holds entry points, keyed by their potential locale string.
    */
-  private readonly entryPointsCache = new Map<string, EntryPointExports>();
+  private readonly entryPointsCache = new Map<string, Promise<EntryPointExports>>();
 
   /**
    * Renders a response for the given HTTP request using the server application.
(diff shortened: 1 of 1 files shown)


jkrems (commented):
Feel free to close as "not worth it". I'm just having flashbacks to issues with cache stampedes during bootup, so I'm a bit overly sensitive about it. :)

packages/angular/ssr/src/app-engine.ts
```diff
@@ -124,7 +127,9 @@ export class AngularAppEngine {
       return undefined;
     }
 
-    const entryPointExports = await entryPoint();
+    const eventualExports = entryPoint();
```
jkrems, line 130: There's _technically_ still a race condition here if `entryPoint()` causes another call to `getEntryPointExports`. Not familiar enough with the code flow to know if that's a real concern, I suspect not.

alan-agius4: Tbh, it's not that much of a concern, although I am not against this improvement. I suggested a couple of things on slack, let me know what you think otherwise looks good.

packages/angular/ssr/src/app-engine.ts
```diff
@@ -124,7 +124,7 @@ export class AngularAppEngine {
       return undefined;
     }
 
-    const entryPointExports = await entryPoint();
+    const entryPointExports = entryPoint();
```
alan-agius4, line 127: NIT: the `async` is not redundant in `getEntryPointExports`.

alan-agius4 (approved):
LGTM, one tiny nit.

packages/angular/ssr/src/app-engine.ts
jkrems: Good point

alan-agius4 (approved):

PR #28507, opened 2024-09-26
refactor(@angular-devkit/build-angular): remove implicit localize polyfill

jkrems:
In the v19 application builder, the localize polyfill should not be added implicitly.

See: https://github.com/angular/angular-cli/commit/d6a34034d7489c09753090642ade4c606cd98ece

### What Changes

1. Test: Reference `./src/polyfills.ts` in the base options so we have an easy way to inject the polyfill based on which builder is used.
2. Test: Make the `setupTarget` function async so we can write to the polyfills file as part of the setup.
3. Builder: Remove the code in the application-builder based Karma support to inject localize.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/karma/tests/options/exclude_spec.ts
+++ b/packages/angular_devkit/build_angular/src/builders/karma/tests/options/exclude_spec.ts
@@ -11,8 +11,8 @@ import { BASE_OPTIONS, KARMA_BUILDER_INFO, describeKarmaBuilder } from '../setup
 
 describeKarmaBuilder(execute, KARMA_BUILDER_INFO, (harness, setupTarget) => {
   describe('Option: "exclude"', () => {
-    beforeEach(() => {
-      setupTarget(harness);
+    beforeEach(async () => {
+      await setupTarget(harness);
     });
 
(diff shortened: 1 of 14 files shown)


packages/angular_devkit/build_angular/src/builders/karma/tests/options/exclude_spec.ts
```diff
@@ -54,7 +54,7 @@ describeKarmaBuilder(execute, KARMA_BUILDER_INFO, (harness, setupTarget) => {
       expect(result?.success).toBeTrue();
     });
 
-    it(`should exclude spec that matches the 'exclude' pattern prefixed with a slash`, async () => {
+    fit(`should exclude spec that matches the 'exclude' pattern prefixed with a slash`, async () => {
```
clydin, line 57:
```suggestion
    it(`should exclude spec that matches the 'exclude' pattern prefixed with a slash`, async () => {
```

jkrems: Argh, forgot that I did it on two levels. Thanks!

clydin (approved):

PR #28509, opened 2024-09-26
feat(@angular-devkit/build-angular): karma-coverage w/ app builder

jkrems:
## Tasks to Cover (🥁)

- [x] Determine list/filtering of files to instrument for coverage.
- [x] Actually instrument those files. Importantly, using a recent version of `istanbul-lib-instrument` and not the one that `karma-coverage` depends on.
- [x] Preserve source maps so that coverage is mapped back to source files.

### Implementation Notes

* The additions in `packages/angular/build/src/tools/babel/plugins/*` are 1:1 copies of files already present in `packages/angular_devkit/build_angular/src/tools/babel/plugins/*`. I couldn't find a clean way to prevent that. Either @angular/build have to expose the plugin in `package.json#exports` or it would have to depend on @angular-devkit/build-angular. Both seemed undesirable.
* The logic for excluding files from coverage is lifted from the [equivalent webpack config](https://github.com/angular/angular-cli/blob/ecfb2b261356946d5f4a653f90c0b78db4ef519c/packages/angular_devkit/build_angular/src/tools/webpack/configs/common.ts#L424-L429). Given the state of Karma, I didn't make an effort to refactor the code to share the logic - I assume we don't expect many changes here. That would definitely be possible with limited effort though.

### Historical Notes

Approaches that _didn't_ work:

- Instrument entire output chunks. `karma-coverage` really doesn't want to filter data from coverage data after the fact. Not to mention that it could be fairly expensive to send that data back to the karma process if it's then just discarded.
- Implied above: Use `karma-coverage`'s built-in preprocessor. The outdated instrumentation code it's using doesn't handle async function correctly and gives bad function coverage data because of it.

Things that _could've_ worked:

- Apply source maps while instrumenting instead of while post-processing the coverage data. But that would've been swimming against the stream.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
+++ b/packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
@@ -21,6 +21,7 @@ interface JavaScriptTransformRequest {
   skipLinker?: boolean;
   sideEffects?: boolean;
   jit: boolean;
+  instrumentForCoverage?: boolean;
 }
 
 const textDecoder = new TextDecoder();
@@ -64,8 +65,13 @@ async function transformWithBabel(
   const { default: importAttributePlugin } = await import('@babel/plugin-syntax-import-attributes');
   const plugins: PluginItem[] = [importAttributePlugin];
 
-  // Lazy load the linker plugin only when linking is required
+  if (options.instrumentForCoverage) {
+    const { default: coveragePlugin } = await import('../babel/plugins/add-code-coverage.js');
+    plugins.push(coveragePlugin);
+  }
+
   if (shouldLink) {
+    // Lazy load the linker plugin only when linking is required
     const linkerPlugin = await createLinkerPlugin(options);
     plugins.push(linkerPlugin);
   }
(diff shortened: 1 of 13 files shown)


packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
```diff
@@ -64,8 +65,15 @@ async function transformWithBabel(
   const { default: importAttributePlugin } = await import('@babel/plugin-syntax-import-attributes');
   const plugins: PluginItem[] = [importAttributePlugin];
 
-  // Lazy load the linker plugin only when linking is required
+  if (options.instrumentForCoverage) {
+    const { default: coveragePlugin } = await loadEsmModule<
+      typeof import('../babel/plugins/add-code-coverage.js')
+    >('../tools/babel/plugins/add-code-coverage.js');
```
clydin, line 71: A direct dynamic import should work here.  Everything in the package is compiled to CJS currently so there shouldn't be any interop issues.

clydin (approved):

packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
jkrems: Makes sense, thanks!

PR #28515, opened 2024-09-27
fix(@angular/build): incomplete string escaping or encoding

alan-agius4:
See: https://github.com/angular/angular-cli/security/code-scanning/76


packages/angular/build/src/utils/server-rendering/manifest.ts
```diff
@@ -149,9 +139,7 @@ export function generateAngularServerAppManifest(
       file.path === INDEX_HTML_CSR ||
       (inlineCriticalCss && file.path.endsWith('.css'))
     ) {
-      serverAssetsContent.push(
-        `['${file.path}', async () => ${escapeUnsafeChars(JSON.stringify(file.text))}]`,
-      );
+      serverAssetsContent.push(`['${file.path}', async () => \`${escapeUnsafeChars(file.text)}\`]`);
```
alan-agius4, line 142: I think that using string literals here makes the other replacements unnecessary. However, the code scanner does not run on pull requests.

clydin (approved):

PR #28519, opened 2024-09-27
feat(@angular-devkit/build-angular): karma+esbuild+watch

jkrems:
This introduces support for `--watch` when using the application builder. It's tested as far as the relevant test case is concerned. But I wouldn't be surprised if there's still some rough corners.

~~**NOTE:** Currently includes the commit from https://github.com/angular/angular-cli/pull/28518 because the two touch some of the same code.~~ Rebased.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
+++ b/packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
@@ -8,6 +8,7 @@
 
 import { BuildOutputFileType } from '@angular/build';
 import {
+  ApplicationBuilderInternalOptions,
   ResultFile,
   ResultKind,
   buildApplicationInternal,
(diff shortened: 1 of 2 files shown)


jkrems:
I think this was the last blocker in terms of "passes all tests". I'll likely try it out once these PRs are part of the next version to see if there's anything else "obviously" wrong with it.

packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
```diff
@@ -33,6 +34,76 @@ class ApplicationBuildError extends Error {
...
+          },
+          context,
+        )) {
+          if (buildOutput.kind === ResultKind.Failure) {
+            subscriber.next({ success: false, message: 'Build failed' });
+          } else if (
+            buildOutput.kind === ResultKind.Incremental ||
+            buildOutput.kind === ResultKind.Full
+          ) {
+            this.emitter.refreshFiles();
```
clydin, line 75: The result files will need to be written here first so that the new test run can use them.
The integration test happens to pass now due to the syntax error that is introduced and the build failing before the tests run.

jkrems: Updated to write test files before calling `refreshFiles` - thanks for catching that!

clydin (approved):

PR #28531, opened 2024-09-30
feat(@angular/build): set development/production condition

jkrems:
Ensures that conditional exports/imports work as intended in `ng serve` and `ng build`:

1. For optimized builds, the "production" condition is enabled.
2. For non-optimized builds, the "development" condition is enabled.

This allows quick conditional logic like the following to work as expected:

```ts
import {verboseLogging} from '#logger';
```

The file can be switched in the `imports` field in `package.json`:

```js
// ...
  "imports": {
    "#logger": {
      "development": "./src/logging/debug.ts",
      "default": "./src/logging/noop.ts"
    }
  },
```

Something similar already (and continues to) work for SSR vs browser code.
The difference can be expressed using the "browser" condition:

```js
// ...
  "imports": {
    "#crashReporter": {
      "browser": "./src/crash/sendBeacon.ts",
      "default": "./src/crash/writeCrashDumpFile.ts"
    }
  },
```

References:
* https://esbuild.github.io/api/#how-conditions-work
* https://vitejs.dev/guide/env-and-mode

### DIFF
--- a/modules/testing/builder/src/dev_prod_mode.ts
+++ b/modules/testing/builder/src/dev_prod_mode.ts
@@ -1,0 +1,54 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { BuilderHarness } from './builder-harness';
+
+export const GOOD_TARGET = './src/good.js';
+export const BAD_TARGET = './src/bad.js';
+
+/** Setup project for use of conditional imports. */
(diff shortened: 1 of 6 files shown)


jkrems:
The changes themselves should be in fairly good shape but leaving as draft until the next tooling meeting.

brandonroberts:
Will there be a condition for the SSR vs client build also?

jkrems:
> Will there be a condition for the SSR vs client build also?

I think that should already exist. I've used the "browser" condition successfully which comes with esbuild/Vite natively.

jkrems:
I removed the code for `import.meta.env.DEV` since that API still needs more discussion. But I did add a test to verify that the "browser" condition works as expected.

clydin (approved):

PR #28554, opened 2024-10-02
fix(@angular/build): disable dev-server websocket when live reload is disabled

clydin:
When live reload is disabled (`"liveReload": false`/`no-live-reload`) within the development server, the Vite websocket server will no longer be initialized. Additionally, the client code will not be loaded by the browser. This allows the development server to be used in test scenarios that would prefer to more fully represent the production fielded configuration such as E2E testing or other forms of browser-based testing.

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -612,6 +612,8 @@ export async function setupServer(
       host: serverOptions.host,
       open: serverOptions.open,
       headers: serverOptions.headers,
+      // Disable the websocket if live reload is disabled (false/undefined are the only valid values)
+      ws: serverOptions.liveReload === false ? false : undefined,
       proxy,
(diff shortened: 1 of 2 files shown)


packages/angular/build/src/builders/dev-server/vite-server.ts
```diff
@@ -612,6 +612,8 @@ export async function setupServer(
       host: serverOptions.host,
       open: serverOptions.open,
       headers: serverOptions.headers,
+      // Disable the websocket if live reload is disabled (false/undefined are the only valid values)
+      ws: serverOptions.liveReload === false ? false : undefined,
```
alan-agius4, line 616: Shouldn't the check be updated to also account for `serverOptions.hmr` ie

`serverOptions.liveReload === false && serverOptions.hmr === false`

alan-agius4 (approved):

alan-agius4:

PR #28571, opened 2024-10-04
feat(@schematics/angular): update app-shell and ssr schematics to adopt new Server Rendering API

alan-agius4:
This commit revises the app-shell and server schematics to incorporate the new Server Rendering API, along with the integration of server-side routes.

BREAKING CHANGE: The app-shell schematic is no longer compatible with Webpack-based builders.


dgp1130 (approved):
LGTM, though I'm not super familiar with schematics. Maybe @clydin has more useful feedback to give here?

clydin (approved):

PR #28598, opened 2024-10-08
feat(@angular/build): introduce `ssr.experimentalPlatform` option

alan-agius4:
This commit introduces a new option called `experimentalPlatform` to the Angular SSR configuration. 

The `experimentalPlatform` option allows developers to specify the target platform for the server bundle, enabling the generation of platform-neutral bundles suitable for deployment in environments like edge workers and other serverless platforms that do not rely on Node.js APIs.

This change enhances the portability of Angular SSR applications and expands their deployment possibilities.

**Note:** that this feature does not include polyfills for Node.js modules and is experimental, subject to future changes.


alan-agius4:
**Note:**  The test failure occurs due to the current test structure relying on https://github.com/angular/angular-cli/pull/28571. Once that PR is merged, the test will pass successfully.

tests/legacy-cli/e2e/tests/build/server-rendering/server-routes-output-mode-server-platform-neutral.ts
```diff
@@ -0,0 +1,133 @@
...
+  const filesDoNotExist: string[] = ['csr/index.html', 'ssr/index.html', 'redirect/index.html'];
+  for (const filePath of filesDoNotExist) {
+    const file = join('dist/test-project/browser', filePath);
+    assert.equal(existsSync(file), false, `Expected '${file}' to not exist.`);
+  }
+
+  const port = await findFreePort();
+  await execAndWaitForOutputToMatch(
+    'npx',
+    ['-y', 'listhen', './dist/test-project/server/server.mjs', `--port=${port}`],
```
dgp1130, line 126: Question: Should `listhen` be in our `devDependencies`? That way our `bazel test` doesn't need network access to download it dynamically and pin the dependency insulate us from breaking changes.

dgp1130 (approved):

tests/legacy-cli/e2e/tests/build/server-rendering/server-routes-output-mode-server-platform-neutral.ts
alan-agius4: Since it’s used only in a single test I preferred to use npx. I can add a version specifier though.

clydin (approved):

PR #28630, opened 2024-10-14
fix(@angular/build): avoid race condition in sass importer

hborchardt:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #27167

Builds under load sometimes fail with "Can't find stylesheet to import" errors.

I was able to reproduce this error in my CI pipeline consistently, by running 4 `npx ng build` builds in parallel on an 8 core VPS. It also occurs sometimes during regular builds/`ng test`, especially when other concurrent jobs put load on the system.

By adding lots and lots of logging, I found the root cause, which is a race condition when the sass worker manages to call `findFileUrl` again in between `Atomics.store(..., 0, 1)` and `Atomics.notify(..., 0)` invocations - I left more details as a comment.

## What is the new behavior?

The race no longer causes failing builds. After being notified, the worker checks again that the signal is not set to 0, and waits for another `notify` if necessary, which will be the correct one.

- [ ] Yes
- [x] No

## Other information
I verified that the second `Atomics.wait(..., 0, 0)` actually does something meaningful, by logging the return value. Under load, I observed two instances where the return value was `"ok"`, indicating that the signal was 0 at the time of the second `wait`, and the build would therefore have aborted without the change.

### DIFF
--- a/packages/angular/build/src/tools/sass/worker.ts
+++ b/packages/angular/build/src/tools/sass/worker.ts
@@ -106,6 +106,25 @@ export default async function renderSassStylesheet(
               containingUrl: containingUrl ? fileURLToPath(containingUrl) : null,
             },
           });
+          // Wait for the main thread to set the signal to 1 and notify, which tells
+          // us that a message can be received on the port.
+          // If the main thread is fast, the signal will already be set to 1, and no
+          // sleep/notify is necessary.
+          // However, there can be a race condition here:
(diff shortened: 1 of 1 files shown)


clydin (approved):

jkrems:

PR #28631, opened 2024-10-14
refactor(@angular/build): add component update middleware to development server

clydin:
An additional development server middleware has been added that responds to client `/@ng/component` requests from the Angular framework for hot component template updates. These client requests are made by the framework after being triggered by the development server sending a `angular:component-update` WebSocket event. The Angular compiler's new internal `_enableHmr` option will emit template replacement and reloading code that uses this new development server update middleware. Within the development server, the component update build result will be used to indicate that an event should be sent to active clients. The build system does not yet generate component update results.

### DIFF
--- a/packages/angular/build/src/tools/vite/middlewares/component-middleware.ts
+++ b/packages/angular/build/src/tools/vite/middlewares/component-middleware.ts
@@ -1,0 +1,42 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import type { Connect } from 'vite';
+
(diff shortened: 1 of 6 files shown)


packages/angular/build/src/tools/vite/middlewares/component-middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+import type { Connect } from 'vite';
+
+const ANGULAR_COMPONENT_PREFIX = '/@ng/component';
+
+export function createAngularComponentMiddleware(
+  templateUpdates: ReadonlyMap<string, string>,
+): Connect.NextHandleFunction {
+  return function angularAssetsMiddleware(req, res, next) {
```
alan-agius4, line 16:
```suggestion
  return function angularComponentMiddleware(req, res, next) {
```

alan-agius4 (approved):
LGTM, just one NIT

PR #28632, opened 2024-10-14
Remove Protractor Builder

dgp1130:
This renames the Protractor builder and its `e2e` generator schematic to mark them private and remove them from Angular's public API. This will give us an opportunity to clean up internal usages and then remove the builder altogether in a future release. Since this rename makes Protractor non-public API, it may be completely deleted in a future patch release.

Protractor was [marked end-of-life in August 2023](https://protractortest.org/). Projects still relying on Protractor should consider migrating to another E2E testing framework, several support solid migration paths from Protractor.

Additional resources:
* https://angular.dev/tools/cli/end-to-end
* https://blog.angular.dev/the-state-of-end-to-end-testing-with-angular-d175f751cb9c

### DIFF
--- a/packages/angular_devkit/core/src/workspace/json/test/angular.json
+++ b/packages/angular_devkit/core/src/workspace/json/test/angular.json
@@ -111,7 +111,7 @@
       "prefix": "",
       "architect": {
         "e2e": {
-          "builder": "@angular-devkit/build-angular:protractor",
+          "builder": "@angular-devkit/build-angular:private-protractor",
           "options": {
             "protractorConfig": "e2e/protractor.conf.js",
             "devServerTarget": "my-app:serve"
--- a/packages/schematics/angular/collection.json
+++ b/packages/schematics/angular/collection.json
@@ -23,10 +23,10 @@
       "schema": "./application/schema.json",
       "description": "Create an Angular application."
     },
-    "e2e": {
+    "private-e2e": {
       "factory": "./e2e",
       "schema": "./e2e/schema.json",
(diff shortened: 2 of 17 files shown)


packages/schematics/angular/collection.json
```diff
@@ -23,10 +23,10 @@
       "schema": "./application/schema.json",
       "description": "Create an Angular application."
     },
-    "e2e": {
```
alan-agius4, line 26: Note: this is already private as it’s marked as hidden

dgp1130: Ah thanks for pointing that out. I think it's still worth renaming this just to identify usage on it which needs to be removed. When we delete Protractor fully, that will likely include this schematic too.

dgp1130:
I patched this change into the FW and Components repos. https://github.com/angular/angular/pull/58254 should be enough to update FW. Components actually doesn't seem to be affected at all. All their Protractor usage comes through Bazel, they don't invoke it through Angular CLI.

Once this lands, we should do another `next` release and then I'll change https://github.com/angular/angular/pull/58254 to include this update of Angular CLI with those Protractor fixes to avoid any internal breakages from this change.

packages/angular_devkit/core/src/workspace/json/test/angular.json
```diff
@@ -111,7 +111,7 @@
       "prefix": "",
```
alan-agius4, line 111: Any changes with `core` can be reverted as this are just test file.

alan-agius4 (approved):

PR #32, opened 2015-08-17
feat(test generated projets): setup unit testing structure powered by…

cironunes:
… Karma

Closes #10

To do

- [ ] Make the `$ ng test` command run karma

PR #28639, opened 2024-10-15
refactor(@angular/build): Auto-CSP support as an index file transformation.

aaronshim:
Auto-CSP is a feature to rewrite the `<script>` tags in a index.html file to either hash their contents or rewrite them as a dynamic loader script that can be hashed. These hashes will be placed in a CSP inside a `<meta>` tag inside the `<head>` of the document to ensure that the scripts running on the page are those known during the compile-time of the client-side rendered application.

This is an application of the recommended [Strict CSP](https://web.dev/articles/strict-csp#choose-hash) approach that is also recommended by the [OWASP CSP Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#strict-csp) and inspired by https://github.com/google/strict-csp. The approach of having build tooling auto-generate a hash-based CSP addresses the main concern around hash-based CSPs that generating and updating the hashes manually is difficult and error-prone.

## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [X] Build related changes
- [ ] CI related changes

There is currently no logic to process the `index.html` to auto-generate hash-based CSPs.

## What is the new behavior?

Adding logic to process the `index.html` to auto-generate hash-based CSPs.

- [ ] Yes
- [X] No

### DIFF
--- a/packages/angular/build/src/utils/index-file/auto-csp.ts
+++ b/packages/angular/build/src/utils/index-file/auto-csp.ts
@@ -1,0 +1,299 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import * as crypto from 'node:crypto';
+import { StartTag, htmlRewritingStream } from './html-rewriting-stream';
+
+/**
+ * The hash function to use for hash directives to use in the CSP.
+ */
+const HASH_FUNCTION = 'sha256';
+
+/**
+ * Store the appropriate attributes of a sourced script tag to generate the loader script.
+ */
+interface SrcScriptTag {
+  src: string;
+  type?: string;
+  async: boolean;
+  defer: boolean;
+}
+
+/**
+ * Get the specified attribute or return undefined if the tag doesn't have that attribute.
+ *
+ * @param tag StartTag of the <script>
+ * @returns
+ */
+function getScriptAttributeValue(tag: StartTag, attrName: string): string | undefined {
+  return tag.attrs.find((attr) => attr.name === attrName)?.value;
+}
+
+/**
+ * Checks whether a particular string is a MIME type associated with JavaScript, according to
+ * https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types#textjavascript
+ *
+ * @param mimeType a string that may be a MIME type
+ * @returns whether the string is a MIME type that is associated with JavaScript
+ */
+function isJavascriptMimeType(mimeType: string): boolean {
+  return mimeType.split(';')[0] === 'text/javascript';
+}
+
+/**
+ * Which of the type attributes on the script tag we should try passing along
+ * based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type
+ * @param scriptType the `type` attribute on the `<script>` tag under question
+ * @returns whether to add the script tag to the dynamically loaded script tag
+ */
+function shouldDynamicallyLoadScriptTagBasedOnType(scriptType: string | undefined): boolean {
+  return (
+    scriptType === undefined ||
+    scriptType === '' ||
+    scriptType === 'module' ||
+    isJavascriptMimeType(scriptType)
+  );
+}
+
+/**
+ * Calculates a CSP compatible hash of an inline script.
+ * @param scriptText Text between opening and closing script tag. Has to
+ *     include whitespaces and newlines!
+ * @returns The hash of the text formatted appropriately for CSP.
+ */
+export function hashTextContent(scriptText: string): string {
+  const hash = crypto.createHash(HASH_FUNCTION).update(scriptText, 'utf-8').digest('base64');
+
+  return `'${HASH_FUNCTION}-${hash}'`;
+}
+
+/**
+ * Finds all `<script>` tags and creates a dynamic script loading block for consecutive `<script>` with `src` attributes.
+ * Hashes all scripts, both inline and generated dynamic script loading blocks.
+ * Inserts a `<meta>` tag at the end of the `<head>` of the document with the generated hash-based CSP.
+ *
+ * @param html Markup that should be processed.
+ * @returns The transformed HTML that contains the `<meta>` tag CSP and dynamic loader scripts.
+ */
+export async function autoCsp(html: string): Promise<string> {
+  const { rewriter, transformedContent } = await htmlRewritingStream(html);
+
+  let openedScriptTag: StartTag | undefined = undefined;
+  let scriptContent: SrcScriptTag[] = [];
+  const hashes: string[] = [];
+
+  /**
+   * Generates the dynamic loading script and puts it in the rewriter and adds the hash of the dynamic
+   * loader script to the collection of hashes to add to the <meta> tag CSP.
+   */
+  function emitLoaderScript() {
+    const loaderScript = createLoaderScript(scriptContent);
+    hashes.push(hashTextContent(loaderScript));
+    rewriter.emitRaw(`<script>${loaderScript}</script>`);
+    scriptContent = [];
+  }
+
+  rewriter.on('startTag', (tag, html) => {
+    if (tag.tagName === 'script') {
+      openedScriptTag = tag;
+      const src = getScriptAttributeValue(tag, 'src');
+
+      if (src) {
+        // If there are any interesting attributes, note them down.
+        const scriptType = getScriptAttributeValue(tag, 'type');
+        if (shouldDynamicallyLoadScriptTagBasedOnType(scriptType)) {
+          scriptContent.push({
+            src: src,
+            type: scriptType,
+            async: getScriptAttributeValue(tag, 'async') !== undefined,
+            defer: getScriptAttributeValue(tag, 'defer') !== undefined,
+          });
+
+          return; // Skip writing my script tag until we've read it all.
+        }
+      }
+    }
+    // We are encountering the first start tag that's not <script src="..."> after a string of
+    // consecutive <script src="...">. <script> tags without a src attribute will also end a chain
+    // of src attributes that can be loaded in a single loader script, so those will end up here.
+    //
+    // The first place when we can determine this to be the case is
+    // during the first opening tag that's not <script src="...">, where we need to insert the
+    // dynamic loader script before continuing on with writing the rest of the tags.
+    // (One edge case is where there are no more opening tags after the last <script src="..."> is
+    // closed, but this case is handled below with the final </body> tag.)
+    if (scriptContent.length > 0) {
+      emitLoaderScript();
+    }
+    rewriter.emitStartTag(tag);
+  });
+
+  rewriter.on('text', (tag, html) => {
+    if (openedScriptTag && !getScriptAttributeValue(openedScriptTag, 'src')) {
+      hashes.push(hashTextContent(html));
+    }
+    rewriter.emitText(tag);
+  });
+
+  rewriter.on('endTag', (tag, html) => {
+    if (openedScriptTag && tag.tagName === 'script') {
+      const src = getScriptAttributeValue(openedScriptTag, 'src');
+      const scriptType = getScriptAttributeValue(openedScriptTag, 'type');
+      openedScriptTag = undefined;
+      // Return early to avoid writing the closing </script> tag if it's a part of the
+      // dynamic loader script.
+      if (src && shouldDynamicallyLoadScriptTagBasedOnType(scriptType)) {
+        return;
+      }
+    }
+
+    if (tag.tagName === 'body' || tag.tagName === 'html') {
+      // Write the loader script if a string of <script>s were the last opening tag of the document.
+      if (scriptContent.length > 0) {
+        emitLoaderScript();
+      }
+    }
+    rewriter.emitEndTag(tag);
+  });
+
+  const rewritten = await transformedContent();
+
+  // Second pass to add the header
+  const secondPass = await htmlRewritingStream(rewritten);
+  secondPass.rewriter.on('startTag', (tag, _) => {
+    secondPass.rewriter.emitStartTag(tag);
+    if (tag.tagName === 'head') {
+      // See what hashes we came up with!
+      secondPass.rewriter.emitRaw(
+        `<meta http-equiv="Content-Security-Policy" content="${getStrictCsp(hashes)}">`,
+      );
+    }
+  });
+
+  return secondPass.transformedContent();
+}
+
+/**
+ * Returns a strict Content Security Policy for mitigating XSS.
+ * For more details read csp.withgoogle.com.
+ * If you modify this CSP, make sure it has not become trivially bypassable by
+ * checking the policy using csp-evaluator.withgoogle.com.
+ *
+ * @param hashes A list of sha-256 hashes of trusted inline scripts.
+ * @param enableTrustedTypes If Trusted Types should be enabled for scripts.
+ * @param enableBrowserFallbacks If fallbacks for older browsers should be
+ *   added. This is will not weaken the policy as modern browsers will ignore
+ *   the fallbacks.
+ * @param enableUnsafeEval If you cannot remove all uses of eval(), you can
+ *   still set a strict CSP, but you will have to use the 'unsafe-eval'
+ *   keyword which will make your policy slightly less secure.
+ */
+function getStrictCsp(
+  hashes?: string[],
+  // default CSP options
+  cspOptions: {
+    enableBrowserFallbacks?: boolean;
+    enableTrustedTypes?: boolean;
+    enableUnsafeEval?: boolean;
+  } = {
+    enableBrowserFallbacks: true,
+    enableTrustedTypes: false,
+    enableUnsafeEval: false,
+  },
+): string {
+  hashes = hashes || [];
+  const strictCspTemplate: Record<string, string[]> = {
+    // 'strict-dynamic' allows hashed scripts to create new scripts.
+    'script-src': [`'strict-dynamic'`, ...hashes],
+    // Restricts `object-src` to disable dangerous plugins like Flash.
+    'object-src': [`'none'`],
+    // Restricts `base-uri` to block the injection of `<base>` tags. This
+    // prevents attackers from changing the locations of scripts loaded from
+    // relative URLs.
+    'base-uri': [`'self'`],
+  };
+
+  // Adds fallbacks for browsers not compatible to CSP3 and CSP2.
+  // These fallbacks are ignored by modern browsers in presence of hashes,
+  // and 'strict-dynamic'.
+  if (cspOptions.enableBrowserFallbacks) {
+    // Fallback for Safari. All modern browsers supporting strict-dynamic will
+    // ignore the 'https:' fallback.
+    strictCspTemplate['script-src'].push('https:');
+    // 'unsafe-inline' is only ignored in presence of a hash or nonce.
+    if (hashes.length > 0) {
+      strictCspTemplate['script-src'].push(`'unsafe-inline'`);
+    }
+  }
+
+  // If enabled, dangerous DOM sinks will only accept typed objects instead of
+  // strings.
+  if (cspOptions.enableTrustedTypes) {
+    strictCspTemplate['require-trusted-types-for'] = ['script'];
+  }
+
+  // If enabled, `eval()`-calls will be allowed, making the policy slightly
+  // less secure.
+  if (cspOptions.enableUnsafeEval) {
+    strictCspTemplate['script-src'].push(`'unsafe-eval'`);
+  }
+
+  return Object.entries(strictCspTemplate)
+    .map(([directive, values]) => {
+      return `${directive} ${values.join(' ')};`;
+    })
+    .join('');
+}
+
+/**
+ * Returns JS code for dynamically loading sourced (external) scripts.
+ * @param srcList A list of paths for scripts that should be loaded.
+ */
+function createLoaderScript(srcList: SrcScriptTag[], enableTrustedTypes = false): string {
+  if (!srcList.length) {
+    throw new Error('Cannot create a loader script with no scripts to load.');
+  }
+  const srcListFormatted = srcList
+    .map((s) => {
+      // URI encoding means value can't escape string, JS, or HTML context.
+      const srcAttr = encodeURI(s.src).replaceAll("'", "\\'");
+      // Can only be 'module' or a JS MIME type or an empty string.
+      const typeAttr = s.type ? "'" + s.type + "'" : undefined;
+      const asyncAttr = s.async ? 'true' : 'false';
+      const deferAttr = s.defer ? 'true' : 'false';
+
+      return `['${srcAttr}', ${typeAttr}, ${asyncAttr}, ${deferAttr}]`;
+    })
+    .join();
+
+  return enableTrustedTypes
+    ? `
+  var scripts = [${srcListFormatted}];
+  var policy = self.trustedTypes && self.trustedTypes.createPolicy ?
+    self.trustedTypes.createPolicy('angular#auto-csp', {createScriptURL: function(u) {
+      return scripts.includes(u) ? u : null;
+    }}) : { createScriptURL: function(u) { return u; } };
+  scripts.forEach(function(scriptUrl) {
+    var s = document.createElement('script');
+    s.src = policy.createScriptURL(scriptUrl[0]);
+    s.type = scriptUrl[1];
+    s.async = !!scriptUrl[2];
+    s.defer = !!scriptUrl[3];
+    document.body.appendChild(s);
+  });\n`
+    : `
+  var scripts = [${srcListFormatted}];
+  scripts.forEach(function(scriptUrl) {
+    var s = document.createElement('script');
+    s.src = scriptUrl[0];
+    s.type = scriptUrl[1];
+    s.async = !!scriptUrl[2];
+    s.defer = !!scriptUrl[3];
+    document.body.appendChild(s);
+  });\n`;
+}
--- a/packages/angular/build/src/utils/index-file/auto-csp_spec.ts
+++ b/packages/angular/build/src/utils/index-file/auto-csp_spec.ts
@@ -1,0 +1,173 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { autoCsp, hashTextContent } from './auto-csp';
+
+// Utility function to grab the meta tag CSPs from the HTML response.
+const getCsps = (html: string) => {
+  return Array.from(
+    html.matchAll(/<meta http-equiv="Content-Security-Policy" content="([^"]*)">/g),
+  ).map((m) => m[1]); // Only capture group.
+};
+
+const ONE_HASH_CSP =
+  /script-src 'strict-dynamic' 'sha256-[^']+' https: 'unsafe-inline';object-src 'none';base-uri 'self';/;
+
+const FOUR_HASH_CSP =
+  /script-src 'strict-dynamic' (?:'sha256-[^']+' ){4}https: 'unsafe-inline';object-src 'none';base-uri 'self';/;
+
+describe('auto-csp', () => {
+  it('should rewrite a single inline script', async () => {
+    const result = await autoCsp(`
+      <html>
+        <head>
+        </head>
+        <body>
+          <script>console.log('foo');</script>
+          <div>Some text </div>
+        </body>
+      </html>
+    `);
+
+    const csps = getCsps(result);
+    expect(csps.length).toBe(1);
+    expect(csps[0]).toMatch(ONE_HASH_CSP);
+    expect(csps[0]).toContain(hashTextContent("console.log('foo');"));
+  });
+
+  it('should rewrite a single source script', async () => {
+    const result = await autoCsp(`
+      <html>
+        <head>
+        </head>
+        <body>
+          <script src="./main.js"></script>
+          <div>Some text </div>
+        </body>
+      </html>
+    `);
+
+    const csps = getCsps(result);
+    expect(csps.length).toBe(1);
+    expect(csps[0]).toMatch(ONE_HASH_CSP);
+    expect(result).toContain(`var scripts = [['./main.js', undefined, false, false]];`);
+  });
+
+  it('should rewrite a single source script in place', async () => {
+    const result = await autoCsp(`
+      <html>
+        <head>
+        </head>
+        <body>
+          <div>Some text</div>
+          <script src="./main.js"></script>
+        </body>
+      </html>
+    `);
+
+    const csps = getCsps(result);
+    expect(csps.length).toBe(1);
+    expect(csps[0]).toMatch(ONE_HASH_CSP);
+    // Our loader script appears after the HTML text content.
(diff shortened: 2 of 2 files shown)


packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+    enableTrustedTypes?: boolean;
+    enableUnsafeEval?: boolean;
+  } = {
+    enableBrowserFallbacks: true,
+    enableTrustedTypes: false,
+    enableUnsafeEval: false,
+  },
+): string {
+  hashes = hashes || [];
+  let strictCspTemplate = {
```
dgp1130, line 169: Consider: I usually recommend `Map<string, T>` over `Record<string, T>` for anything with dynamically assigned keys to mitigate potential prototype pollution issues.

All the keys are string literals, so I don't think it's too important here, up to you if it's worth the effort.

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * Returns JS code for dynamically loading sourced (external) scripts.
+ * @param srcList A list of paths for scripts that should be loaded.
+ */
+function createLoaderScript(
+  srcList: SrcScriptTag[],
+  enableTrustedTypes = false,
+): string | undefined {
+  if (!srcList.length) {
+    return undefined;
```
dgp1130, line 224: Suggestion: Throw an error instead of returning `undefined`.

It's a bit confusing to have a `createLoaderScript` function that sometimes doesn't create a loader script. We already check `if (scriptContent)` above, so I think we should always have items in the list. If we ever don't have items we might not be enabling auto CSP as expected. I suspect it would be better to fail up front rather than no-op in that situation.

Removing `| undefined` from the type also makes usage of this function a little simpler.

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * The hash function to use for hash directives to use in the CSP.
+ */
+const HASH_FUNCTION = 'sha256';
+
+/**
+ * Store the appropriate attributes of a sourced script tag to generate the loader script.
+ */
+interface SrcScriptTag {
+  scriptType: 'src';
```
dgp1130, line 22: Question: Do we / will we need to check this type anywhere? Should we drop it?

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+ */
+function createLoaderScript(
+  srcList: SrcScriptTag[],
+  enableTrustedTypes = false,
+): string | undefined {
+  if (!srcList.length) {
+    return undefined;
+  }
+  const srcListFormatted = srcList
+    .map((s) => `['${s.src}',${s.type ? "'" + s.type + "'" : undefined}, ${s.async}]`)
```
dgp1130, line 227: Nit: Missing space between `'${s.src}',${s.type ...}`

dgp1130, line 227: Question: Do we need to escape `src` / `type` / `async`? Those could contain `"><script src="...">` which we probably wouldn't want (admittedly auto CSP should solve the script injection issue and exploiting this would require compromising source code / a dependency, but still something to potentially mitigate).

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+  });\n`
+    : `
+  var scripts = [${srcListFormatted}];
+  scripts.forEach(function(scriptUrl) {
+    var s = document.createElement('script');
+    s.src = scriptUrl[0];
+    s.type = scriptUrl[1];
+    s.async = !!scriptUrl[2];
+    document.body.appendChild(s);
+  });\n`;
```
dgp1130, line 251: Consider: You should be fine to use modern features like `const`, `() => {}`, `for-of`, destructuring, etc. Angular already limits support to browsers with those features. Not sure if you were avoiding those for compatibility or performance reasons.

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
```diff
@@ -0,0 +1,119 @@
...
+        </head>
+        <body>
+          <script>console.log('foo');</script>
+          <div>Some text </div>
+        </body>
+      </html>
+    `);
+
+    expect(result).toContain(
+      `<meta http-equiv="Content-Security-Policy" content="script-src 'strict-dynamic' 'sha256-1kOLrDKT3TBiHLcnxiGsc7HF/lyVJKLhoZDSn0UwCfo=' https: 'unsafe-inline';object-src 'none';base-uri 'self';">`,
```
dgp1130, line 25: Suggestion: Rather than hard-coding the hashes which don't mean anything to humans anyways, could we replace this with something like:

```typescript
export(result).toContain(`... '${hash('console.log(\'foo\');')}' ...`);
```

I get that this runs the risk that the `hash` function is wrong, but I think that's outweighed by making the tests easier to read and avoid hard-coding or having to update hashes manually.

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
```diff
@@ -0,0 +1,119 @@
...
+        </body>
+      </html>
+    `);
+
+    expect(result).toContain(
+      `<meta http-equiv="Content-Security-Policy" content="script-src 'strict-dynamic' 'sha256-vkcRYaaRLTkuunplxAyjuivyXpK+pbbEfJB92l8u+aY=' https: 'unsafe-inline';object-src 'none';base-uri 'self';">`,
+    );
+    expect(result).toContain(
+      `var scripts = [['./main1.js',undefined, false],['./main2.js',undefined, false],['./main3.js','module', false]];`,
+    );
```
dgp1130, line 87: Consider: Maybe verify that we only have one `<script>` tag in the output? That way we ensure they were batched together as expected.

dgp1130 (commented):
Thanks for putting this together! Very excited to see this feature land.

I left a bunch of readability / refactoring suggestions, but none of them are really critical and what you have is pretty reasonable so feel free to decline any which aren't worth applying.

For the commit message, we should use `refactor` here since it doesn't actually expose any functionality yet. When you get to the "keystone" commit which exposes the relevant option for end users, that's where you would put `feat` to be included in the changelog.

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Store the appropriate attributes of a sourced script tag to generate the loader script.
+ */
+interface SrcScriptTag {
+  scriptType: 'src';
+  src: string;
+  type?: string;
+  async?: boolean;
+}
```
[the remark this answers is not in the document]
aaronshim: Admittedly, I haven't done a lot of testing with the `defer` attribute, but for now, we can add it to the list of attributes that we want to pass along-- that way, worst case scenario is that we are left with some vestigial code.

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+function getScriptAttributeValue(tag: StartTag, attrName: string): string | undefined {
+  return tag.attrs.find((attr) => attr.name === attrName)?.value;
+}
+
+/**
+ * Calculates a CSP compatible hash of an inline script.
+ * @param scriptText Text between opening and closing script tag. Has to
+ *     include whitespaces and newlines!
+ */
+function hashInlineScript(scriptText: string): string {
```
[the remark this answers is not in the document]
aaronshim: We meant "inline script" as in a script tag with text content for us to hash-- renaming!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+  const hash = crypto.createHash(HASH_FUNCTION).update(scriptText, 'utf-8').digest('base64');
+  return `'${HASH_FUNCTION}-${hash}'`;
+}
+
+/**
+ * Finds all `<script>` tags and creates a dynamic script loading block for consecutive `<script>` with `src` attributes.
+ * Hashes all scripts, both inline and generated dynamic script loading blocks.
+ * Inserts a `<meta>` tag at the end of the `<head>` of the document with the generated hash-based CSP.
+ *
+ * @param html Markup that should be processed.
```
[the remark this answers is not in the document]
aaronshim: Added!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+ */
+export async function autoCsp(html: string): Promise<string> {
+  const { rewriter, transformedContent } = await htmlRewritingStream(html);
+
+  let openedScriptTag: StartTag | undefined = undefined;
+  let scriptContent: SrcScriptTag[] | undefined = undefined;
```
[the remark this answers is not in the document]
aaronshim: Done!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+    rewriter.emitText(tag);
+  });
+
+  rewriter.on('endTag', (tag, html) => {
+    if (tag.tagName === 'script') {
+      const src = getScriptAttributeValue(openedScriptTag!, 'src');
+      openedScriptTag = undefined;
+
+      if (src) {
```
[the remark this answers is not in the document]
aaronshim: Added, thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+        scriptContent.push({
+          scriptType: 'src',
+          src: src,
+          type: getScriptAttributeValue(tag, 'type'),
+          async: !!getScriptAttributeValue(tag, 'async'),
+        });
+        return; // Skip writing my script tag until we've read it all...
+      }
+    }
+    // We are encountering the first start tag that's not <script src="..."> after a string of those.
```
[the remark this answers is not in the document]
aaronshim: Added, thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+    if (tag.tagName === 'body' || tag.tagName === 'html') {
+      // Write the loader script if a string of <script>s were the last opening tag of the document.
+      if (scriptContent) {
+        const loaderScript = createLoaderScript(scriptContent);
+        if (loaderScript) {
+          hashes.push(hashInlineScript(loaderScript));
+          rewriter.emitRaw(`<script>${loaderScript}</script>`);
+        }
+        scriptContent = undefined;
+      }
```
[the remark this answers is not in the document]
aaronshim: We'll opt for the first approach, and if the logic still looks weird, we can try the second later, thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+    rewriter.emitEndTag(tag);
+  });
+
+  const rewritten = await transformedContent();
+
+  // Second pass to add the header
+  const secondPass = await htmlRewritingStream(rewritten);
+  secondPass.rewriter.on('endTag', (tag, _) => {
```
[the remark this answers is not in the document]
aaronshim: Sure, I think that's easy enough to change for now!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+        `<meta http-equiv="Content-Security-Policy" content="${getStrictCsp(hashes)}">`,
+      );
+    }
+    secondPass.rewriter.emitEndTag(tag);
+  });
+  return secondPass.transformedContent();
+}
+
+/**
+ * Returns a strict Content Security Policy for mittigating XSS.
```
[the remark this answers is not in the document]
aaronshim: Done!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+  }
+
+  // If enabled, dangerous DOM sinks will only accept typed objects instead of
+  // strings.
+  if (cspOptions.enableTrustedTypes) {
+    strictCspTemplate = {
+      ...strictCspTemplate,
+      ...{ 'require-trusted-types-for': [`'script'`] },
+    };
```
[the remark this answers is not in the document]
aaronshim: Changed, thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,252 @@
...
+  // and 'strict-dynamic'.
+  if (cspOptions.enableBrowserFallbacks) {
+    // Fallback for Safari. All modern browsers supporting strict-dynamic will
+    // ignore the 'https:' fallback.
+    strictCspTemplate['script-src'].push('https:');
+    // 'unsafe-inline' is only ignored in presence of a hash or nonce.
+    if (hashes.length > 0) {
+      strictCspTemplate['script-src'].push(`'unsafe-inline'`);
+    }
+  }
```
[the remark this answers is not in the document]
aaronshim: For what it's worth, this is largely from https://github.com/google/strict-csp/blob/main/strict-csp/index.ts

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: Done, thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: Done.

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: I'm not quite sure what you mean-- do you mean getting rid of this type definition entirely?

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: Done

aaronshim: Attempted escaping without a dependency-- is it likely to break some behavior, in your opinion?

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
aaronshim: Done!

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
aaronshim: Tested!

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: I'm not quite sure I'm following-- this is for `strictCspTemplate`?

packages/angular/build/src/utils/index-file/auto-csp.ts
dgp1130: Hmm, this seems like a surprisingly tricky problem. We're trying to move a string from an HTML context to a JS context without altering its content, yet the two languages have different escaping rules. I don't think `encodeURI` is the right mechanism, it doesn't drop quotes.

```typescript
encodeURI("foo'bar"); // foo'bar
```

There's two sides to this. First, I think we need to escape the JS side (special characters in a single quoted string literal) such as `/foo'bar\'baz`:
1.  `'`, since that started the string literal and is the only token which can end it.
2. `\`, since that can escape other characters which might not have been intended to be escaped.

```typescript
function escape(str: string): string {
  // Convert `\` -> `\\` and `'` -> `\'`.
  // Probably can't use `replaceAll` just yet, I think you need a regex match with the 'g' flag?
  return str.replaceAll('\\', '\\\\').replaceAll("'", "\\'");
}

const srcListFormatted = srcList.map((s) => `['${escape(s.src)}', ...]`);
// ...
```

Second, we need to handle HTML entities such as `<script src="/foo&amp;bar">` which requests `/foo&bar`. I wonder if the HTML parser does this automatically. If you give it `<script src="/foo&amp;bar">`, does it give you a JS string containing `/foo&bar`? I feel like that would be the correct behavior for this parser and would generate a JS string literal of `'/foo&bar'` in the loader which seems like what we want?

Tests for all these cases would be super useful to make sure the HTML string and the generated JS string are equivalent. The ones which come to mind are:
* `/one'two\'three\\'four\\\'five` - Check quotes and backslashes.
* `/one&two&amp;three&amp;amp;four` - Check HTML entities.

There are other HTML entities like `#32;`, but I think if the named entities work we're probably fine? Not sure if there would be value in a test case there. Any other edge case scenarios we need to handle here?

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,322 @@
...
+ */
+function emitLoaderScript(
+  scriptContent: SrcScriptTag[],
+  hashes: string[],
+  rewriter: RewritingStream,
+) {
+  const loaderScript = createLoaderScript(scriptContent);
+  hashes.push(hashScriptText(loaderScript));
+  rewriter.emitRaw(`<script>${loaderScript}</script>`);
+}
```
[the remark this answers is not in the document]
aaronshim: You're so right, we are in JS so we can use all the language niceties like closure capture! 🎉

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,322 @@
...
+  let openedScriptTag: StartTag | undefined = undefined;
+  let scriptContent: SrcScriptTag[] = [];
+  let hashes: string[] = [];
+
+  rewriter.on('startTag', (tag, html) => {
+    if (tag.tagName === 'script') {
+      openedScriptTag = tag;
+      const src = getScriptAttributeValue(tag, 'src');
+
+      if (src) {
```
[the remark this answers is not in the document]
aaronshim: Perhaps I am misunderstanding-- but if we do `if (!src) return;` here, they won't get to later on:

```js
if (scriptContent.length > 0) {
  emitLoaderScript();
}
rewriter.emitStartTag(tag);
```

which we want script tags that *do not* have a src to do? (Emit a loader script based on the consecutive `<script src="...">` seen so far and then continue writing as normal)

Sorry, the logic is really tricky here and involves falling through various if statements-- but this is the rewritten version because before in the version where I had more if's and else-if's to denote each of the subcases distinctly, there was a lot of repetition.

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,322 @@
...
+            src: src,
+            type: scriptType,
+            async: !(getScriptAttributeValue(tag, 'async') === undefined),
+            defer: !(getScriptAttributeValue(tag, 'defer') === undefined),
+          });
+          return; // Skip writing my script tag until we've read it all.
+        }
+      }
+    }
+    // We are encountering the first start tag that's not <script src="..."> after a string of
```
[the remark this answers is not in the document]
aaronshim: I think we want the loader script to be written before a `<script>console.log(...)</script>`, if the previous tags were a string of `<script src=...>`, right?

This is captured in the unit test that does

```html
<script>console.log('foo');</script>
<script src="./main.js"></script>
<script src="./main2.js"></script>
<script>console.log('bar');</script>
<script src="./main3.js"></script>
<script src="./main4.js"></script>
<div>Some text </div>
```

Which I thought should be

```html
<script>console.log('foo');</script>
<script>...Loader script for main.js and main2.js</script>
<script>console.log('bar');</script>
<script>...Loader script for main3.js and main4.js</script>
<div>Some text </div>
```

As per a script with a malformed type (i.e. `<script type="not/js" ...>`), the current behavior is "leave it alone", since it's not guaranteed to run anyway according to the spec? If we want to be more forgiving, we can always rewrite it into a loader script if it also contains a `src` attribute, but I thought at least leaving it alone was preferred to silently stripping it from the HTML. (Regardless, we need the logic to fall through to `rewriter.emitStartTag` as written to preserve this behavior.)

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,322 @@
...
+    if (scriptContent.length > 0) {
+      emitLoaderScript(scriptContent, hashes, rewriter);
+      scriptContent = [];
+    }
+    rewriter.emitStartTag(tag);
+  });
+
+  rewriter.on('text', (tag, html) => {
+    if (openedScriptTag && !getScriptAttributeValue(openedScriptTag, 'src')) {
+      hashes.push(hashScriptText(html));
```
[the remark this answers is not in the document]
aaronshim: Done, thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: I was doing so to be extra sure about correctness since our reference implementation (which we have been using for the testing and feasibility explorations so far) has been doing-- is the modern syntax something that you feel strongly about swapping out for the initial release or is there time to follow up in a later fix? Thanks!

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
```diff
@@ -0,0 +1,124 @@
...
+        </head>
+        <body>
+          <script src="./main.js"></script>
+          <div>Some text </div>
+        </body>
+      </html>
+    `);
+
+    expect(result).toContain(
+      `<meta http-equiv="Content-Security-Policy" content="script-src 'strict-dynamic' 'sha256-cfa69N/DhgtxzDzIHo+IFj9KPigQLDJgb6ZGZa3g5Cs=' https: 'unsafe-inline';object-src 'none';base-uri 'self';">`,
```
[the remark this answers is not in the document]
aaronshim: Done, thanks for the suggestion!

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: So, first of all, I really am enjoying this conversation, but for the sake of the threat model of whether our dynamic loader script would cause insecurities-- I don't think our CSP is meant to be a defense against *malicious developers*, especially with this flow, since all of the malicious payloads has to be around at compile time. (We imagine that if they have influence over the `src` attribute of a script tag in code that appears at compile time, they don't need to use this pathway to do a lot more harm.)

That being said, I think it's a great goal to make sure that developers can't inadvertently break our loader script! I don't think we need to worry about `.type` since that's fixed to one of the values allowed by `shouldDynamicallyLoadScriptTagBasedOnType`, but `.src` can be any free-form value.

I think we can stop the `'` case breaking out of our JS strings and rendering the JS unusable (by `replaceAll`ing the single-quotes) but I think there are other trivial JS escapes, like `</script>` appearing anywhere in the script to immediately exit the JS context back into the HTML context-- is this another case worth covering for? (Is there a developer that would set `<script src='./</script>'>` legitimately?) If so, I think `encodeURI` might be encoding `<` and `>`?

I added all of these as test cases. Thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
dgp1130: What you have is fine, if you'd rather stick with that to stay aligned with your prior testing I'm ok with that. I just wasn't sure why you were using legacy syntax and wanted to point out that's not a hard constraint here.

+1 to fixing later if you'd rather go down that approach.

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,311 @@
...
+ * @param srcList A list of paths for scripts that should be loaded.
+ */
+function createLoaderScript(srcList: SrcScriptTag[], enableTrustedTypes = false): string {
+  if (!srcList.length) {
+    throw new Error('Cannot create a loader script with no scripts to load.');
+  }
+  const srcListFormatted = srcList
+    .map(
+      (s) =>
+        `['${encodeURI(s.src).replaceAll("'", "\\'")}', ${s.type ? "'" + encodeURI(s.type) + "'" : undefined}, ${s.async ? 'true' : 'false'}, ${s.defer ? 'true' : 'false'}]`,
```
dgp1130, line 283: Consider: `type` isn't a URI and it's limited to known values. I feel like we don't need `encodeURI` for it?

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    `);
+
+    const csps = getCsps(result);
+    expect(csps.length).toBe(1);
+    expect(csps[0]).toMatch(ONE_HASH_CSP);
+    // &amp; encodes correctly
+    expect(result).toContain(`'/foo&bar'`);
+    // Impossible to escape a string and create invalid loader JS with a '
+    // (Quotes and backslashes work)
+    expect(result).toContain(`'/one\\'two%5C\\'three%5C%5C\\'four%5C%5C%5C\\'five'`);
```
dgp1130, line 129: Question: Why do we want to `encodeURI` and convert `\` to `%5C`? If I do `<script src="/foo\bar">` and `<script src="/foo%5Cbar">` the browser seems to request two different URLs. I feel like we'd want to just propagate the `\` character?

dgp1130 (approved):
Thanks for working through these comments! I don't have any significant concerns at this point, @clydin can you take a quick look too?

packages/angular/build/src/utils/index-file/auto-csp.ts
dgp1130: Ooh, `</script>` is a really tricky one...

Yeah, I'm less concerned from the security angle, if an attacker compromised source code they could do a lot more damage than this particular attribute. It's not even that vulnerable to a compromised dependency since the `index.html` is a hard-coded static file, so even then we're very unlikely to parse dependency code here.

I'm mainly focused on correctness here. The user provided a `<script src="...">` string, and that string should be unchanged in the final `<script src="...">` we append to the DOM. It's probably a bit overly pedantic, I can't imagine any legitimate use cases for `</script>` in a URL for example and maybe it's not worth stressing about that. I could see URLs with a quote or `&amp;` though and would want to stay consistent. We can always fix up any missed edge cases later though, so I wouldn't get hung up on the specifics too much more than I already made you. :sweat_smile:

dgp1130:
FYI: For a review, the commits you have are probably fine, but once this is approved and ready to submit, you'll need to squash/amend them to match the commit message conventions so we can merge it. Just a heads up.

aaronshim:
> FYI: For a review, the commits you have are probably fine, but once this is approved and ready to submit, you'll need to squash/amend them to match the commit message conventions so we can merge it. Just a heads up.

For sure! I realized that the GitHub code review interface was hiding comments as "outdated" and such when I amended and force pushed on the same commit, so I decided to stop doing that until we are ready to merge!

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: Agreed!

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
aaronshim: That is... certainly unfortunate, I am perplexed why this is the case? I was under the impression that `encodeURI` was spec-compliant to only encode characters that would be treated the same.

What are our options here? If we move away from encodeURI, I am not sure what clean options we have that are along the lines of "escape everything... except for %5C".

aaronshim: I think we are going to live with backslashes being `%5C` for now, as it is RFC 3986 compliant.

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,314 @@
...
+      const src = getScriptAttributeValue(tag, 'src');
+
+      if (src) {
+        // If there are any interesting attributes, note them down.
+        const scriptType = getScriptAttributeValue(tag, 'type');
+        if (shouldDynamicallyLoadScriptTagBasedOnType(scriptType)) {
+          scriptContent.push({
+            src: src,
+            type: scriptType,
+            async: !(getScriptAttributeValue(tag, 'async') === undefined),
```
alan-agius4, line 137:
```suggestion
            async: getScriptAttributeValue(tag, 'async') !== undefined,
```

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,314 @@
...
+            defer: !(getScriptAttributeValue(tag, 'defer') === undefined),
```
alan-agius4, line 138:
```suggestion
            defer: getScriptAttributeValue(tag, 'defer') !== undefined,
```

package.json
```diff
@@ -174,6 +174,7 @@
     "ora": "5.4.1",
     "pacote": "19.0.0",
     "parse5-html-rewriting-stream": "7.0.0",
+    "parse5-sax-parser": "7.0.0",
```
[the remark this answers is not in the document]
clydin: Actually, opened up #28666 so once in you can just import from `./html-rewriting-stream`

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
dgp1130: I'm ok with `%5C`, I'm inclined to go with that for now and revisit if we discover any problems.

package.json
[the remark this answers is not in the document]
aaronshim: Done, thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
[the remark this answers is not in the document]
aaronshim: Comment was updated to call out that we stop on `<script>` tags with no `src`! Thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim: We can look into streamlining the loader script in future bugfixes, thanks!

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
aaronshim: Keeping the escapaing and test cases as-is for now! Thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -0,0 +1,322 @@
...
+  'text/javascript1.1',
+  'text/javascript1.2',
+  'text/javascript1.3',
+  'text/javascript1.4',
+  'text/javascript1.5',
+  'text/jscript',
+  'text/livescript',
+  'text/x-ecmascript',
+  'text/x-javascript',
+]);
```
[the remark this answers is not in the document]
aaronshim: Refactored to remove all the legacy JS MIME types and only check for `text/javascript`.

dgp1130 (approved):

clydin (approved):

PR #28642, opened 2024-10-15
fix(@schematics/angular): explicitly set standalone:false

jkrems:
Since the new default for standalone is `true`, any place that previously omitted the value to get an implicit `false` now needs to set it to `false` explicitly.

See: https://github.com/angular/angular/pull/58169


jkrems:
Updated to apply the explicit `false` consistently for component, directive, and pipe - matching the default flip in core.

tests/legacy-cli/e2e/tests/build/prerender/discover-routes-ngmodule.ts
```diff
@@ -86,7 +86,7 @@ export default async function () {
   const componentNames: string[] = ['one', 'two-child-one', 'two-child-two'];
 
   for (const componentName of componentNames) {
-    await ng('generate', 'component', componentName, '--project', projectName);
+    await ng('generate', 'component', componentName, '--project', projectName, '--no-standalone');
```
alan-agius4, line 89: Is this really needed? When a component is generated in an ngModule based app, the component is generated in non standalone mode.

https://github.com/angular/angular-cli/blob/3db1d81397b8d0b883ab9f7b9e6a64303f7072d2/packages/schematics/angular/application/index.ts#L192

jkrems: Let's try it! I made this change before I actually updated the schematics, so very possible it wasn't actually needed.

jkrems:
Included the angular update commit in the PR again to ensure the post-upgrade state is tested.

jkrems:
🍏 💚 🟢 !

alan-agius4 (approved):
LGTM

PR #28646, opened 2024-10-16
feat(@angular/build): enable component stylesheet hot replacement by default

clydin:
When using the `application` builder (default for new projects) with the development server, component style only changes will now automatically replace the styles within the running application without a full reload of the page. No application code changes are necessary and both file-based (`styleUrl`/`styleUrls`) and inline (`styles`) component styles are supported. Within a component template, `<style>` elements and `<link rel="stylesheet">` elements with relative `href` attributes are also supported.

If any issues are encountered or it is preferred to not hot replace component styles, the `NG_HMR_CSTYLES=0` environment variable can be used to disable the feature. Setting the `liveReload` option to false will also disable all updates.

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -137,8 +137,13 @@ export async function* serveWithVite(
     process.setSourceMapsEnabled(true);
   }
 
-  // TODO: Enable by default once full support across CLI and FW is integrated
-  browserOptions.externalRuntimeStyles = useComponentStyleHmr;
+  // Enable to support component style hot reloading (`NG_HMR_CSTYLES=0` can be used to disable)
+  browserOptions.externalRuntimeStyles = !!serverOptions.liveReload && useComponentStyleHmr;
+  if (browserOptions.externalRuntimeStyles) {
+    // Preload the @angular/compiler package to avoid first stylesheet request delays.
+    // Once @angular/build is native ESM, this should be re-evaluated.
+    void loadEsmModule('@angular/compiler');
+  }
 
   // Setup the prebundling transformer that will be shared across Vite prebundling requests
   const prebundleTransformer = new JavaScriptTransformer(
@@ -166,7 +171,7 @@ export async function* serveWithVite(
     explicitBrowser: [],
     explicitServer: [],
   };
-  const usedComponentStyles = new Map<string, string[]>();
+  const usedComponentStyles = new Map<string, Set<string>>();
   const templateUpdates = new Map<string, string>();
 
   // Add cleanup logic via a builder teardown.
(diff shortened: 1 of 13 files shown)


packages/angular/build/src/builders/dev-server/vite-server.ts
```diff
@@ -137,8 +137,13 @@ export async function* serveWithVite(
...
   }
 
-  // TODO: Enable by default once full support across CLI and FW is integrated
-  browserOptions.externalRuntimeStyles = useComponentStyleHmr;
+  // Enable to support component style hot reloading (`NG_HMR_CSTYLES=0` can be used to disable)
+  browserOptions.externalRuntimeStyles = !!serverOptions.liveReload && useComponentStyleHmr;
+  if (browserOptions.externalRuntimeStyles) {
+    // Preload the @angular/compiler package to avoid first stylesheet request delays.
+    // Once @angular/build is native ESM, this should be re-evaluated.
+    await loadEsmModule('@angular/compiler');
```
JoostK, line 145: Is there any harm in not awaiting this, just using it to prime the module cache? It should have an exception handler attached to avoid unhandled promise exceptions, but other than that I think it's fine to ignore this result (although tsetse may disagree)

packages/angular/build/src/builders/dev-server/vite-server.ts
```diff
@@ -137,8 +137,13 @@ export async function* serveWithVite(
     process.setSourceMapsEnabled(true);
   }
 
-  // TODO: Enable by default once full support across CLI and FW is integrated
-  browserOptions.externalRuntimeStyles = useComponentStyleHmr;
+  // Enable to support component style hot reloading (`NG_HMR_CSTYLES=0` can be used to disable)
+  browserOptions.externalRuntimeStyles = !!serverOptions.liveReload && useComponentStyleHmr;
```
dgp1130, line 141: Consider: Prefer `Boolean(...)` over `!!` to be more clear about your intent.

packages/angular/build/src/builders/dev-server/vite-server.ts
clydin: shouldn't be any

packages/angular/build/src/builders/dev-server/vite-server.ts
```diff
@@ -470,7 +475,7 @@ async function handleUpdate(
           // are not typically reused across components.
           const componentIds = usedComponentStyles.get(filePath);
           if (componentIds) {
-            return componentIds.map((id) => ({
+            return [...componentIds].map((id) => ({
```
dgp1130, line 478: Consider: Prefer `Array.from` to be more clear about your intent.

https://techhub.social/@develwithoutacause/111819710786873448

packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
```diff
@@ -16,7 +16,7 @@ export function createAngularAssetsMiddleware(
   server: ViteDevServer,
   assets: Map<string, string>,
   outputFiles: AngularMemoryOutputFiles,
-  usedComponentStyles: Map<string, string[]>,
+  usedComponentStyles: Map<string, Set<string>>,
```
dgp1130, line 19: Definitely a good use of `Set`, underappreciated API IMHO.

dgp1130 (approved):
LGTM, just some minor readability suggestions. Up to you if they're worth changing.

Issue #28656, opened 2024-10-17
Invalid component stylesheet ID request


cexbrayat:
### Command

serve

### Is this a regression?

- [x] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

19.0.0-next.10

### Description

When creating a new project with CLI v19.0.0-next.11, the application doesn't have proper styling and we can see an error `Invalid component stylesheet ID request: ng-c1946662535`

![Image](https://github.com/user-attachments/assets/4b89a775-66a8-4970-99c1-8f1b2dceba15)

### Minimal Reproduction

npx @angular/cli@next new v19-next-11 --defaults --no-routing
ng serve
Open the browser ad see the error logged in terminal running `ng serve`

### Exception or Error

Invalid component stylesheet ID request: ng-c1946662535

### Your Environment

Angular CLI: 19.0.0-next.11
Node: 20.13.1
Package Manager: npm 9.7.1
OS: darwin arm64

Angular: 19.0.0-next.10
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1900.0-next.11
@angular-devkit/build-angular   19.0.0-next.11
@angular-devkit/core            19.0.0-next.11
@angular-devkit/schematics      19.0.0-next.11
@angular/cli                    19.0.0-next.11
@schematics/angular             19.0.0-next.11
rxjs                            7.8.1
typescript                      5.6.3
zone.js                         0.15.0

PR #28658, opened 2024-10-17
fix(@angular/build): relax constraints on external stylesheet component id

clydin:
The number of digits in the component identifier that is generated at runtime can vary. The check for a valid identifier now accounts for this case.

Closes: #28656


alan-agius4 (approved):

spock123:
I guess this will only come in the next release in a weeks time..

PR #28663, opened 2024-10-17
feat(@angular/build): Auto-CSP support as a part of angular.json schema

aaronshim:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [X] Build related changes
- [ ] CI related changes

There are no angular.json configurations to add the auto-csp to the index.html transformation.

Issue Number: N/A

## What is the new behavior?

There are angular.json configurations to add the auto-csp to the index.html transformation.

- [ ] Yes
- [X] No

## Other information

This PR also contains content from https://github.com/angular/angular-cli/pull/28639. Please ignore the files changed in that previous PR and instead on

- packages/angular/build/src/utils/index-file/index-html-generator.ts
- packages/angular/build/src/tools/esbuild/index-html-generator.ts
- packages/angular/build/src/builders/application/schema.json
- packages/angular/build/src/builders/application/options.ts

### DIFF
--- a/packages/angular/build/src/builders/application/schema.json
+++ b/packages/angular/build/src/builders/application/schema.json
@@ -37,6 +37,33 @@
       "type": "string",
       "description": "Customize the base path for the URLs of resources in 'index.html' and component stylesheets. This option is only necessary for specific deployment scenarios, such as with Angular Elements or when utilizing different CDN locations."
     },
+    "security": {
+      "description": "Security features to protect against XSS and other common attacks",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "autoCsp": {
+          "description": "Enables automatic generation of a hash-based Strict Content Security Policy (https://web.dev/articles/strict-csp#choose-hash) based on scripts in index.html. Will default to true once we are out of experimental/preview phases.",
+          "default": false,
+          "oneOf": [
+            {
+              "type": "object",
+              "properties": {
+                "unsafeEval": {
+                  "type": "boolean",
+                  "description": "Include the `unsafe-eval` directive (https://web.dev/articles/strict-csp#remove-eval) in the auto-CSP. Please only enable this if you are absolutely sure that you need to, as allowing calls to eval will weaken the XSS defenses provided by the auto-CSP.",
+                  "default": false
+                }
+              },
+              "additionalProperties": false
+            },
+            {
+              "type": "boolean"
+            }
+          ]
+        }
+      }
+    },
     "scripts": {
       "description": "Global scripts to be included in the build.",
       "type": "array",
--- a/packages/angular/build/src/utils/index-file/index-html-generator.ts
+++ b/packages/angular/build/src/utils/index-file/index-html-generator.ts
@@ -12,6 +12,7 @@ import { NormalizedCachedOptions } from '../normalize-cache';
 import { NormalizedOptimizationOptions } from '../normalize-optimization';
 import { addEventDispatchContract } from './add-event-dispatch-contract';
 import { CrossOriginValue, Entrypoint, FileInfo, augmentIndexHtml } from './augment-index-html';
+import { autoCsp } from './auto-csp';
 import { InlineCriticalCssProcessor } from './inline-critical-css';
 import { InlineFontsProcessor } from './inline-fonts';
 import { addNgcmAttribute } from './ngcm-attribute';
@@ -32,6 +33,10 @@ export interface IndexHtmlGeneratorProcessOptions {
   hints?: { url: string; mode: HintMode; as?: string }[];
 }
 
+export interface AutoCspOptions {
+  unsafeEval: boolean;
+}
+
 export interface IndexHtmlGeneratorOptions {
   indexPath: string;
   deployUrl?: string;
@@ -43,6 +48,7 @@ export interface IndexHtmlGeneratorOptions {
   cache?: NormalizedCachedOptions;
   imageDomains?: string[];
   generateDedicatedSSRContent?: boolean;
+  autoCsp?: AutoCspOptions;
 }
 
 export type IndexHtmlTransform = (content: string) => Promise<string>;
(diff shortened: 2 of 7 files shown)


packages/angular/build/src/utils/index-file/index-html-generator.ts
```diff
@@ -86,6 +92,11 @@ export class IndexHtmlGenerator {
       this.csrPlugins.push(addNgcmAttributePlugin());
       this.ssrPlugins.push(addEventDispatchContractPlugin(), addNoncePlugin());
     }
+
+    // Auto-CSP (as the last step)
+    if (options.autoCsp) {
+      this.csrPlugins.push(autoCspPlugin(options.autoCsp.unsafeEval));
```
dgp1130, line 98: @alan-agius4, do we also want to push this to `this.ssrPlugins`? I think that uses a different copy of the `index.html` page which might need the same transformation, but I'm not 100% on the architecture here.

packages/angular/build/src/builders/application/schema.json
```diff
@@ -37,6 +37,33 @@
       "type": "string",
       "description": "Customize the base path for the URLs of resources in 'index.html' and component stylesheets. This option is only necessary for specific deployment scenarios, such as with Angular Elements or when utilizing different CDN locations."
     },
+    "security": {
+      "description": "Security features to protect against XSS and other common attacks",
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "autoCsp": {
+          "description": "Enables auto-CSP generation. Will default to true once we are out of experimental/preview phases.",
```
dgp1130, line 46: Suggestion: Some users might not immediately know what "CSP" means in this context, so the description should probably spell out the term and give a little more details. Maybe say "Enables automatic generation of a Content Security Policy based on scripts in `index.html`." If there's a good MDN or Chrome Developers link explaining what CSP is, that would also be good to link here.

Once the documentation for this feature exists on angular.dev, it would be great to update this description to point to that documentation and help users understand exactly what "auto-CSP generation" actually means.

packages/angular/build/src/builders/application/schema.json
```diff
@@ -37,6 +37,33 @@
       "type": "string",
       "description": "Customize the base path for the URLs of resources in 'index.html' and component stylesheets. This option is only necessary for specific deployment scenarios, such as with Angular Elements or when utilizing different CDN locations."
     },
+    "security": {
```
dgp1130, line 40: Is `security` too broad a grouping? I'm not entirely sure what else we'd put in here. I'm wondering if we should drop this object and just have `autoCsp` as a top-level option? WDYT @alan-agius4, @clydin?

packages/angular/build/src/builders/application/schema.json
```diff
@@ -37,6 +37,33 @@
...
+        "autoCsp": {
+          "description": "Enables auto-CSP generation. Will default to true once we are out of experimental/preview phases.",
+          "default": false,
+          "oneOf": [
+            {
+              "type": "object",
+              "properties": {
+                "unsafeEval": {
+                  "type": "boolean",
+                  "description": "Include the `unsafe-eval` directive in the auto-CSP. Please only enable this if you are absolutely sure that you need to, as allowing calls to eval will weaken the XSS defenses provided by the auto-CSP.",
```
dgp1130, line 54: Consider: If there's a "canonical" link which provides a good explanation of `unsafe-eval`, we should link to that here for users not familiar with the directive.

dgp1130 (commented):
No major concerns on my end here, just want to make sure we agree the option name and that it's wired up correctly for SSR users.

packages/angular/build/src/utils/index-file/index-html-generator.ts
alan-agius4: Yeah, the SSR does use a different copy. During SSR, there are additional script tags added and removed. 

However, it seems that this will not work for SSR, as this happens because render and during SSR rendering there are extra scripts that are added/removed.

packages/angular/build/src/builders/application/schema.json
aaronshim: For now, I think we only have `autoCsp` to add to this top-level object, but this is a logical location for grouping other secure-by-default options that we may pursue in the future. If you feel strongly about not having this nesting for this release, I suppose we can go to a top-level option. Also always happy to chat about what other secure-by-default features we can add beyond auto-CSP!

packages/angular/build/src/utils/index-file/index-html-generator.ts
aaronshim: So, to clarify, we do not have concerns around leaving the plugin as-is and only applying to `csrPlugins`, correct?

packages/angular/build/src/builders/application/schema.json
aaronshim: Updated the description and added a link to web.dev that describes our approach-- IIRC the MDN documentation around strict hash-based CSPs are WIP right now, but we'd be super happy to swap to an angular.dev or a MDN source once we have a canonical source there to point to. Thanks!

packages/angular/build/src/builders/application/schema.json
aaronshim: Added our web.dev source for describing what and why `unsafe-eval` is needed here. Thanks!

packages/angular/build/src/utils/index-file/index-html-generator.ts
dgp1130: Yeah that's fine. I'd love to find a way to make this SSR compatible in the future, but that can be a future feature.

Should we throw an error somewhere if someone configures `autoCsp` *and* enables SSR? We can also do that in a follow up PR, doesn't need to be this one.

packages/angular/build/src/builders/application/schema.json
dgp1130: We can keep `security` just to leave the door open for future features in this space. Always happy to discuss making Angular more secure-by-default!

aaronshim: Excited to work further with y'all's! Thanks!

dgp1130 (approved):

packages/angular/build/src/utils/index-file/index-html-generator.ts
dgp1130: Seems reasonable to me.

dgp1130 (approved):

alan-agius4 (approved):

dgp1130 (approved):

dgp1130 (approved):

PR #28672, opened 2024-10-20
refactor(@angular/build): add internal support for generating template update functions

clydin:
The internal AOT Angular compilation processing can now generate the newly introduced template update functions during development server rebuilds of template only file changes. These template update functions are not yet used but provide the infrastructure to enable template hot replacement in a future change.

### DIFF
--- a/packages/angular/build/src/tools/angular/compilation/aot-compilation.ts
+++ b/packages/angular/build/src/tools/angular/compilation/aot-compilation.ts
@@ -8,6 +8,7 @@
 
 import type ng from '@angular/compiler-cli';
 import assert from 'node:assert';
+import { relative } from 'node:path';
 import ts from 'typescript';
 import { profileAsync, profileSync } from '../../esbuild/profiling';
 import {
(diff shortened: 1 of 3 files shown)


packages/angular/build/src/tools/angular/compilation/aot-compilation.ts
```diff
@@ -385,3 +422,16 @@ function findAffectedFiles(
 
   return affectedFiles;
 }
+
+function hasOnlyTemplates(modifiedFiles: Set<string>): boolean {
+  for (const file of modifiedFiles) {
+    const lowerFile = file.toLowerCase();
+    if (lowerFile.endsWith('.html') || lowerFile.endsWith('.svg')) {
```
alan-agius4, line 429: NIT: 

```suggestion
    const ext = extname(file).toLowerCase();
    if (ext === '.html' || ext === '.svg') {
```

Or also

```js
 return [...modifiedFiles].every(file => /\.(html|svg)$/i.test(file));
```

alan-agius4 (approved):
LGTM, just one optional NIT

PR #28678, opened 2024-10-21
refactor(@angular/ssr): export tokens from the `/tokens` entry point

### LINKED ISSUE #28670: Server code goes to browser bundle on v19
### Command

build

### Is this a regression?

_No response_

### Description

Hi. I've noticed some strange code coming to prod build on v19 (+outputMode server + ssr enabled + prerender disabled) during lighthouse test:
 
1) Some kind of server-only error:
Image

2) Postcss?
Image

3) Unvenv with a bunch of polyfills for node's fs?

Image

Correct me if i am wrong, but should this all (50+kb unminified) code come to browser build? Is this hydration-related packages?

### Minimal Reproduction

ng build

### Exception or Error

_No response_

### Your Environment

latest v19

### Anything else relevant?

_No response_

alan-agius4:
This commit relocates the DI tokens to a dedicated entry point. This change ensures that we do not depend on tree-shaking to eliminate server code from client bundles.

Closes #28670


packages/angular/ssr/tokens/BUILD.bazel
```diff
@@ -0,0 +1,19 @@
...
+
+ts_library(
+    name = "tokens",
+    srcs = glob(
+        [
+            "*.ts",
+            "src/**/*.ts",
+        ],
+    ),
+    module_name = "@angular/ssr/tokens",
```
alan-agius4, line 13: Any other ideas for as an entry-point name?

dgp1130: `@angular/ssr/tokens` seems reasonable to me.

dgp1130 (approved):
Is the goal to support users doing:

```typescript
import { REQUEST } from '@angular/ssr/tokens';

inject(REQUEST, { optional: true });
```

And be able to run that on the client without a risk of retaining server code?

I can see some users having encountering friction from having to import the non-default entry point, but IDE auto-imports should mostly solve that problem, and reducing the risk of server code in the client bundle seems worth the trade-off.

alan-agius4:
> Is the goal to support users doing:
> 
> ```ts
> import { REQUEST } from '@angular/ssr/tokens';
> 
> inject(REQUEST, { optional: true });
> ```
> 
> And be able to run that on the client without a risk of retaining server code?
> 
> I can see some users having encountering friction from having to import the non-default entry point, but IDE auto-imports should mostly solve that problem, and reducing the risk of server code in the client bundle seems worth the trade-off.

Correct.

PR #28682, opened 2024-10-22
refactor: remove workarounds for Safari 14 and 15

alan-agius4:
Both of these versions are no longer supported by Angular and thus these workers are no longer required.


devversion:
I was leaning towards not doing this but waiting a bit longer because:

- the issues caused by the Safari bug are out of this world; and super difficult to track; I remember debugging and trying to figure out why the lexical scope of some code randomly was broken
- its low effort to keep this a bit longer
- even though those versions aren't officially supporter anymore, a few teams may still build for it

alan-agius4:
I’m not very strongly about this, but we haven’t officially supported Safari 15 for quite a while now, and currently, we support the two most recent major versions of Safari, which are 17 and 18. 

I agree that keeping this code is low maintenance, though.

//cc @dgp1130 thoughts?

PR #28701, opened 2024-10-23
test: e2e tests for autoCsp

aaronshim:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/angular/angular-cli/blob/main/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe: E2E tests

No e2e tests for https://github.com/angular/angular-cli/pull/28663

Issue Number: N/A

## What is the new behavior?

e2e tests for https://github.com/angular/angular-cli/pull/28663

- [ ] Yes
- [X] No

### DIFF
--- a/tests/legacy-cli/e2e/tests/build/auto-csp.ts
+++ b/tests/legacy-cli/e2e/tests/build/auto-csp.ts
@@ -1,0 +1,136 @@
+import assert from 'node:assert';
+import { getGlobalVariable } from '../../utils/env';
+import { expectFileToMatch, writeMultipleFiles } from '../../utils/fs';
+import { findFreePort } from '../../utils/network';
+import { execAndWaitForOutputToMatch, ng } from '../../utils/process';
+import { updateJsonFile } from '../../utils/project';
+
+const CSP_META_TAG = /<meta http-equiv="Content-Security-Policy"/;
+
+export default async function () {
+  assert(
+    getGlobalVariable('argv')['esbuild'],
+    'This test should not be called in the Webpack suite.',
+  );
+
+  // Turn on auto-CSP
+  await updateJsonFile('angular.json', (json) => {
+    const build = json['projects']['test-project']['architect']['build'];
+    build.options = {
+      ...build.options,
+      security: { autoCsp: true },
+    };
+  });
+
+  await writeMultipleFiles({
+    'serve.js': `
+    const express = require('express');
+    const path = require('path');
+
+    const app = express();
+    const PORT = process.env.PORT || 3000;
+
+    app.use(express.static(path.join(__dirname, 'dist/test-project/browser')));
+
+    app.listen(PORT, () => {
+      console.log('Node Express server listening on ' + PORT);
+    });
+    `,
+    'public/script1.js': `
+    const externalScriptCreated = 1337;
+    console.warn('First External Script: ' + inlineScriptBodyCreated);
+    `,
+    'public/script2.js': `console.warn('Second External Script: ' + externalScriptCreated);`,
+    'src/index.html': `
+      <!doctype html>
+      <html lang="en">
+      <head>
+        <meta charset="utf-8">
+        <base href="/">
+        <script>
+          const inlineScriptHeadCreated = 1339;
+          console.warn("Inline Script Head");
+        </script>
+      </head>
+      <body>
+        <app-root></app-root>
+        
+        <script>
+          const inlineScriptBodyCreated = 1338;
+          console.warn("Inline Script Body: " + inlineScriptHeadCreated);
+        </script>
+        <script src='script1.js'></script>
+        <script src='script2.js'></script>
+        </body>
+      </html>
+    `,
+    'e2e/src/app.e2e-spec.ts': `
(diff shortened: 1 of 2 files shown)


tests/legacy-cli/e2e/tests/build/auto-csp.ts
```diff
@@ -0,0 +1,131 @@
...
+      <!doctype html>
+      <html lang="en">
+      <head>
+        <meta charset="utf-8">
+        <base href="/">
+        <script>console.log("Inline Script Head");</script>
+      </head>
+      <body>
+        <app-root></app-root>
+        <script>console.log("Inline Script Body");</script>
```
dgp1130, line 58: Should we include an external script tag here too to make sure that works? You can probably put a JS file in the `src/public/` directory which will be included in the output so the browser can load it successfully.

tests/legacy-cli/e2e/tests/build/auto-csp.ts
```diff
@@ -0,0 +1,131 @@
...
+          // ).not.toBeNull();
+
+          // Test the contents.
+          expect(await element(by.css('h1')).getText()).toMatch('Hello');
+
+          // Make sure the server styles got replaced by client side ones.
+          expect(
+            await element(by.css('style[ng-app-id="ng"]')).isPresent()
+          ).toBeFalsy();
+          expect(await element(by.css('style')).getText()).toMatch('');
```
dgp1130, line 104: I think we can drop these CSS checks?

tests/legacy-cli/e2e/tests/build/auto-csp.ts
```diff
@@ -0,0 +1,131 @@
+import assert from 'node:assert';
+import { getGlobalVariable } from '../../utils/env';
+import { expectFileToMatch, rimraf, writeMultipleFiles } from '../../utils/fs';
+import { findFreePort } from '../../utils/network';
+import { installWorkspacePackages } from '../../utils/packages';
+import { execAndWaitForOutputToMatch, ng } from '../../utils/process';
+import { updateJsonFile, updateServerFileForWebpack, useSha } from '../../utils/project';
+
+const MULTI_HASH_CSP =
+  /script-src 'strict-dynamic' (?:'sha256-[^']+' )+https: 'unsafe-inline';object-src 'none';base-uri 'self';/;
```
dgp1130, line 10: At this level of abstraction I wouldn't assert on the policy this precisely, that's more of a unit test level assertion. Instead I'd just check:
1. That a CSP policy was added (regex match for `<meta http-equiv="Content-Security-Policy"`), don't care what the policy is.
2. That the browser did not error from that policy.

Optionally it might also be good to verify that the scripts actually executed (maybe just check that they printed to the console?). That way we know we aren't asserting too soon before the scripts have been checked, I think there's a low risk of a false positive there though, so probably not too important.

dgp1130 (approved):
Just a few minor suggestions, hopefully @clydin can take a quick look once he's back just to make sure we're restricting the test to esbuild correctly.

tests/legacy-cli/e2e/tests/build/auto-csp.ts
```diff
@@ -0,0 +1,131 @@
...
+    getGlobalVariable('argv')['esbuild'],
+    'This test should not be called in the Webpack suite.',
+  );
+  // forcibly remove in case another test doesn't clean itself up
+  // await rimraf('node_modules/@angular/ssr');
+  // We don't need SSR yet but eventually for when Auto-CSP will also have a SSR implementation.
+  // await ng('add', '@angular/ssr', '--skip-confirmation', '--skip-install');
+
+  await useSha();
+  await installWorkspacePackages();
```
clydin, line 23: This can also be commented out for now since it is only need for the above `@angular/ssr` installation.

tests/legacy-cli/e2e/tests/build/auto-csp.ts
clydin: I think verifying that there are no errors should cover the script execution cases but to ensure i think it would be good to check the console output for the added script logging (`Inline Script...`) cases as well.

aaronshim: Updated to just check for `const CSP_META_TAG = /<meta http-equiv="Content-Security-Policy"/;` and to check that all four scripts ran in order (two inline and two sourced) with globally set variables from previously run scripts.

dgp1130 (approved):
LGTM, I wish we had a better mechanism for serving the built page and running the browser, but given the current state of our e2e infra I don't think we have a better option which doesn't pull on a lot of other complexity. Any concerns here @clydin?

clydin (approved):

jkrems:
I assume `target: patch` or `target: minor` are both equally fine for this change?

dgp1130:
Should be `patch`, this is released on 19.0.0.

jkrems:

PR #28705, opened 2024-10-24
refactor(@schematics/angular): remove redundant `standalone: true` from templates

alan-agius4:
This is no longer needed.

### DIFF
--- a/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
+++ b/packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
@@ -2,7 +2,6 @@ import { <% if(changeDetection !== 'Default') { %>ChangeDetectionStrategy, <% }%
 
 @Component({<% if(!skipSelector) {%>
   selector: '<%= selector %>',<%}%><% if(standalone) {%>
-  standalone: true,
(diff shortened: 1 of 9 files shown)


packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
```diff
@@ -1,9 +1,7 @@
 import { <% if(changeDetection !== 'Default') { %>ChangeDetectionStrategy, <% }%>Component<% if(!!viewEncapsulation) { %>, ViewEncapsulation<% }%> } from '@angular/core';
 
 @Component({<% if(!skipSelector) {%>
-  selector: '<%= selector %>',<%}%><% if(standalone) {%>
-  standalone: true,
-  imports: [],<%} else { %>
```
clydin, line 6: nit: Should we leave the `imports` field?

clydin (approved):

packages/schematics/angular/component/files/__name@dasherize@if-flat__/__name@dasherize__.__type@dasherize__.ts.template
alan-agius4: I left it as it was there and in most of the cases, users would need to import something.

alan-agius4:

PR #28714, opened 2024-10-25
Output mode warning changed

alan-agius4:
**fix(@angular/build): simplify disabling server features with `--no-server` via command line**

When `outputMode` is configured, `--prerender` and `--app-shell` become no-ops, making server feature disabling difficult without modifying the configuration.

This commit introduces the `--no-server` option, which can be used with `--output-mode static` to fully disable all server features.

```
ng build --output-mode static --no-server
```

 **fix(@angular/build): add warning when `--prerendering` or `--app-shell` are no-ops**
    
Both options are ineffective when used with `outputMode`, so a warning is now issued.

### DIFF
--- a/packages/angular/build/src/builders/application/options.ts
+++ b/packages/angular/build/src/builders/application/options.ts
@@ -220,17 +220,17 @@ export async function normalizeOptions(
       options.ssr = false;
     }
 
-    if (options.prerender) {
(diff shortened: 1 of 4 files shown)


clydin (approved):

alan-agius4:

PR #28718, opened 2024-10-25
fix(@angular/ssr): ensure wildcard RenderMode is applied when no Angular routes are defined

alan-agius4:
This fix addresses a bug where, in the absence of defined Angular routes, the RenderMode was not correctly applied based on the wildcard setting.

### DIFF
--- a/packages/angular/ssr/src/routes/ng-routes.ts
+++ b/packages/angular/ssr/src/routes/ng-routes.ts
@@ -411,26 +411,26 @@ export async function getRoutesFromAngularRouterConfig(
       injector.get(APP_BASE_HREF, null, { optional: true }) ??
       injector.get(PlatformLocation).getBaseHrefFromDOM();
 
+    const compiler = injector.get(Compiler);
+
+    const serverRoutesConfig = injector.get(SERVER_ROUTES_CONFIG, null, { optional: true });
+    let serverConfigRouteTree: RouteTree<ServerConfigRouteTreeAdditionalMetadata> | undefined;
+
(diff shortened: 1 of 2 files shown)


packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -478,7 +478,12 @@ export async function getRoutesFromAngularRouterConfig(
         }
       }
     } else {
-      routesResults.push({ route: '', renderMode: RenderMode.Prerender });
+      const renderMode = serverConfigRouteTree?.match('')?.renderMode ?? RenderMode.Prerender;
```
alan-agius4, line 481: @dgp1130, I'd prefer if in this case it was a hard error both the old and new API. But this would require us to generate an additional component and add route when doing `ng-new`. 

Is this something that we'll be willing to do?

dgp1130: This is the case where `export const SERVER_ROUTES = [];`? You think that should be a hard error?

I can see where you're coming from. This has bugged me in the past because we do the same thing for client-side routes and it means the `<router-outlet>` is a somewhat confusing no-op in `ng new` and it's not clear how to add a route, which leads to devs using `component` instead of `loadComponent` and deoptimizing their apps.

Could we generate `{ path: '/**', renderMode: RenderMode.Prerender }` by default in SSR? Would that avoid the need for a defined route? Or do we need to add a route in the app router to get this to work?

dgp1130 (approved):

alan-agius4:

PR #28719, opened 2024-10-25
refactor(@angular/build): Pin the supported browsers in the default browserslist config

JeanMeche:
In order to deterministic list of supported browser for a given version, we now pin the versions in the browserlists config.

This could fix issue where like in #28707, a minor of Angular broke support for Safari 16, where is previously worked (because the definition of "last 2" got updated to drop Safari 16 with that bump). 

This would also address a demand on the fw side angular/angular#54195.

### DIFF
--- a/packages/angular/build/src/utils/supported-browsers.ts
+++ b/packages/angular/build/src/utils/supported-browsers.ts
@@ -12,13 +12,14 @@ export function getSupportedBrowsers(
   projectRoot: string,
   logger: { warn(message: string): void },
 ): string[] {
+  // This list should match the last 2 versions of the browsers we support at the release.
   browserslist.defaults = [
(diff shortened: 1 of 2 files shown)


alan-agius4:
Thanks for this but this would require a wider discussion and certainly the blessing of the framework team, thus let me mark this for further discussion.

JeanMeche:
SGTM, will bring this up in the next fw sync.

dgp1130:
We started a conversation about this in Tooling and agree that pinning the version is likely a good thing to do, however there are broader challenges with Angular's current browser support policy which likely should be changed as well. I started a separate conversation about potentially revising that policy and depending on how that goes, we can consider updating the CLI's definition to match in this PR.

PR #28731, opened 2024-10-28
refactor(@angular/build): use ETag headers for development external component styles

clydin:
When using the development server, Angular component stylesheet requests that use hot replacement (default for v19) will now use the `ETag` header to remove the need to reprocess and resend encapsulated component styles if the styles have not changed since last usage. The ETag value used is a combination of the style content and the encapsulation component identifier.


clydin:

PR #28732, opened 2024-10-28
refactor(@angular/build): use ETags for dev-server resource requests

clydin:
When using the development server with the application build system, resource files managed by the build system will now using ETag headers to avoid resending content that has not changed since the last request. This includes such content as global stylesheet files, image/font files referenced in stylesheets, and other files managed by the bundler.


clydin:

PR #28737, opened 2024-10-29
fix(@angular/ssr): enable serving of prerendered pages in the App Engine

alan-agius4:
This commit implements the capability for the App Engine to serve prerendered pages directly. Previously, we relied on frameworks like Express for this functionality, which resulted in inconsistent redirects for directories where in some cases a trailing slash was added to the route.

**Note:** This change applies only when using the new SSR APIs. When using the `CommonEngine`, a 3rd party static serve middleware is still required.

### DIFF
--- a/goldens/public-api/angular/ssr/index.api.md
+++ b/goldens/public-api/angular/ssr/index.api.md
@@ -8,8 +8,7 @@ import { EnvironmentProviders } from '@angular/core';
 
 // @public
 export class AngularAppEngine {
-    getPrerenderHeaders(request: Request): ReadonlyMap<string, string>;
-    render(request: Request, requestContext?: unknown): Promise<Response | null>;
+    handle(request: Request, requestContext?: unknown): Promise<Response | null>;
     static ɵhooks: Hooks;
 }
 
--- a/packages/angular/ssr/node/src/app-engine.ts
+++ b/packages/angular/ssr/node/src/app-engine.ts
@@ -24,49 +24,19 @@ export class AngularNodeAppEngine {
   private readonly angularAppEngine = new AngularAppEngine();
 
   /**
-   * Renders an HTTP response based on the incoming request using the Angular server application.
+   * Handles an incoming HTTP request by serving prerendered content, performing server-side rendering,
+   * or delivering a static file for client-side rendered routes based on the `RenderMode` setting.
    *
-   * The method processes the incoming request, determines the appropriate route, and prepares the
-   * rendering context to generate a response. If the request URL corresponds to a static file (excluding `/index.html`),
-   * the method returns `null`.
+   * @param request - The HTTP request to handle.
+   * @param requestContext - Optional context for rendering, such as metadata associated with the request.
+   * @returns A promise that resolves to the resulting HTTP response object, or `null` if no matching Angular route is found.
    *
-   * Example: A request to `https://www.example.com/page/index.html` will render the Angular route
-   * associated with `https://www.example.com/page`.
-   *
-   * @param request - The incoming HTTP request object to be rendered.
-   * @param requestContext - Optional additional context for the request, such as metadata or custom settings.
-   * @returns A promise that resolves to a `Response` object, or `null` if the request URL is for a static file
-   * (e.g., `./logo.png`) rather than an application route.
+   * @note A request to `https://www.example.com/page/index.html` will serve or render the Angular route
+   * corresponding to `https://www.example.com/page`.
    */
-  render(request: IncomingMessage, requestContext?: unknown): Promise<Response | null> {
-    return this.angularAppEngine.render(createWebRequestFromNodeRequest(request), requestContext);
-  }
+  async handle(request: IncomingMessage, requestContext?: unknown): Promise<Response | null> {
+    const webRequest = createWebRequestFromNodeRequest(request);
 
-  /**
-   * Retrieves HTTP headers for a request associated with statically generated (SSG) pages,
-   * based on the URL pathname.
-   *
-   * @param request - The incoming request object.
-   * @returns A `Map` containing the HTTP headers as key-value pairs.
-   * @note This function should be used exclusively for retrieving headers of SSG pages.
-   * @example
-   * ```typescript
-   * const angularAppEngine = new AngularNodeAppEngine();
-   *
-   * app.use(express.static('dist/browser', {
-   *   setHeaders: (res, path) => {
-   *     // Retrieve headers for the current request
-   *     const headers = angularAppEngine.getPrerenderHeaders(res.req);
-   *
-   *     // Apply the retrieved headers to the response
-   *     for (const [key, value] of headers) {
-   *       res.setHeader(key, value);
-   *     }
-   *   }
-     }));
-  * ```
-  */
-  getPrerenderHeaders(request: IncomingMessage): ReadonlyMap<string, string> {
-    return this.angularAppEngine.getPrerenderHeaders(createWebRequestFromNodeRequest(request));
+    return this.angularAppEngine.handle(webRequest, requestContext);
   }
 }
--- a/packages/angular/ssr/src/app.ts
+++ b/packages/angular/ssr/src/app.ts
@@ -12,11 +12,19 @@ import { ServerAssets } from './assets';
 import { Hooks } from './hooks';
 import { getAngularAppManifest } from './manifest';
 import { RenderMode } from './routes/route-config';
+import { RouteTreeNodeMetadata } from './routes/route-tree';
 import { ServerRouter } from './routes/router';
 import { sha256 } from './utils/crypto';
 import { InlineCriticalCssProcessor } from './utils/inline-critical-css';
 import { LRUCache } from './utils/lru-cache';
 import { AngularBootstrap, renderAngular } from './utils/ng';
+import { joinUrlParts, stripIndexHtmlFromURL, stripLeadingSlash } from './utils/url';
+
+/**
+ * The default maximum age in seconds.
+ * Represents the total number of seconds in a 365-day period.
+ */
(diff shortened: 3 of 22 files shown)


goldens/public-api/angular/ssr/index.api.md
```diff
@@ -8,8 +8,9 @@ import { EnvironmentProviders } from '@angular/core';
 
 // @public
 export class AngularAppEngine {
```
alan-agius4, line 10: Happy to change the names here.

dgp1130: I think I'm fine with `AngularAppEngine`. Sometimes it won't be rendering Angular, but I think it's close enough to work.

goldens/public-api/angular/ssr/index.api.md
```diff
@@ -8,8 +8,9 @@ import { EnvironmentProviders } from '@angular/core';
...
-    getPrerenderHeaders(request: Request): ReadonlyMap<string, string>;
+    process(request: Request, requestContext?: unknown): Promise<Response | null>;
     render(request: Request, requestContext?: unknown): Promise<Response | null>;
+    serve(request: Request): Promise<Response | null>;
```
dgp1130, line 13: Question: Is there still value in having separate methods for these operations? Should we just have a single `serve` method which does everything?

I'm a bit worried this API might be too complicated to use effectively and users might forget to call one of these steps or do them out of order. If we do need multiple methods here, we should at least document how we expect them to be used together.

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -43,30 +43,37 @@ export class AngularNodeAppEngine {
...
+   * @returns A promise that resolves to the HTTP response object resulting from the request handling,
+   * or null if no matching content is found.
+   */
+  async process(request: IncomingMessage, requestContext?: unknown): Promise<Response | null> {
+    const webRequest = createWebRequestFromNodeRequest(request);
+
+    return (
+      (await this.angularAppEngine.serve(webRequest)) ??
+      (await this.angularAppEngine.render(webRequest, requestContext))
+    );
```
dgp1130: Consider: Is there a reason not to delegate to `this.angularAppEngine.process`? Would that reduce duplication in the logic?

packages/angular/ssr/src/app.ts
```diff
@@ -124,6 +131,66 @@ export class AngularServerApp {
...
+   * This method attempts to match the incoming request URL with a prerendered route in the application.
+   * If the route is found and is configured for prerendering, the corresponding static asset is retrieved
+   * and returned as a `Response` object with the appropriate HTTP headers.
+   *
+   * @param request - The incoming HTTP request for a potential static page.
+   * @returns A promise that resolves to a `Response` object containing the prerendered page content
+   * if available, or `null` if the request does not match a prerendered route or asset.
+   */
+  async serve(request: Request): Promise<Response | null> {
+    const url = stripIndexHtmlFromURL(new URL(request.url));
```
dgp1130: Question: Does this imply that we're serving `/foo.html` (and presumably the app router would perform a client-side nav to `/foo`)?

Would it be better to 301/302 redirect to `/foo`?

packages/angular/ssr/src/app.ts
```diff
@@ -124,6 +131,66 @@ export class AngularServerApp {
...
+
+    const matchedRoute = this.router.match(new URL(request.url));
+    if (!matchedRoute) {
+      return null;
+    }
+
+    const { headers, renderMode } = matchedRoute;
+    if (renderMode !== RenderMode.Prerender) {
+      return null;
+    }
```
dgp1130, line 204: Question: Just for my curiosity, what is the need to check the router? Is `hasServerAsset` not sufficient to tells us whether a prerendered route exists and should be served?

packages/angular/ssr/src/app.ts
```diff
@@ -124,6 +131,66 @@ export class AngularServerApp {
...
+
+    // TODO(alanagius): handle etags
+
+    const content = await this.assets.getServerAsset(assetPath);
+
+    return new Response(content, {
+      headers: {
+        'Content-Type': 'text/html;charset=UTF-8',
+        // 30 days in seconds
+        'Cache-Control': `max-age=${DEFAULT_MAX_AGE}`,
```
dgp1130, line 220: Consider: Is 30 days too long for a default cache? If I make a change to my app and redeploy, existing users may not receive that change for a month by default? That feels high to me. I'm thinking something on the order of hours might be a better default policy. What do other frameworks/providers do here?

dgp1130, line 220: Consider: Do we need this to be configurable? I expect cache time is something users will want to control.

I know we've talked about a more comprehensive API for working with caches in the future, but is there anything that's worth doing for this PR?

packages/angular/ssr/src/assets.ts
```diff
@@ -35,6 +35,16 @@ export class ServerAssets {
...
 
+  /**
+   * Checks if a specific server-side asset exists.
+   *
+   * @param path - The path to the server asset.
+   * @returns A boolean indicating whether the asset exists.
+   */
+  hasServerAsset(path: string): boolean {
+    return this.manifest.assets.has(path);
+  }
```
dgp1130, line 46: Consider: Could we update `getServerAsset` to return `Promise<string> | null` and drop `hasServerAsset`? I can't imagine many situations we'd want to check if an asset exists but not immediately read it.

dgp1130 (commented):
Left a few comments, the two main concerns are whether we can simplify the API surface of `AngularAppEngine` and how we should let devs interact with the caching policy.

goldens/public-api/angular/ssr/index.api.md
alan-agius4: I mean that methods names :)

packages/angular/ssr/src/app.ts
alan-agius4: We could offer an option to configure this globally, right now users do have the option to set this per route.

packages/angular/ssr/src/app.ts
alan-agius4: No, the stripping of the `index.html` is simply done for matching the route. No redirections or change in URL is involved here.

packages/angular/ssr/src/app.ts
alan-agius4: Great question! This is because some files in the server assets, like CSS files needed for critical CSS inlining, shouldn't be served by the Angular App Engine. Instead, these files should be handled by other middleware to ensure they’re processed appropriately.

We should only handle known Angular routes.

packages/angular/ssr/src/app.ts
alan-agius4: Our current default is set to 365 days. With the addition of etags (which I'll address in a follow-up), a longer period should not be an issue. I looked into other frameworks like Next and Nuxt, and they also appear to set this to 365 days.

goldens/public-api/angular/ssr/index.api.md
alan-agius4: Here’s a revised version of your text:

---

- **`render`**: handles server-side rendering (SSR) requests.  
- **`serve`**: handles static site generation (SSG) requests.  
- **`process`**: manages both SSR and SSG requests.  

I’m considering renaming these methods or potentially consolidating them into a single method that addresses both SSG and SSR. My intention was to provide users with more control over request handling, but having a single method might simplify the process.

packages/angular/ssr/src/assets.ts
alan-agius4: I considered this, but I am not too keen on this because we'd and up having to check for null and when `null` throw an error in a couple of other places.

packages/angular/ssr/src/app.ts
dgp1130: Gotcha, makes sense!

packages/angular/ssr/src/app.ts
dgp1130: Oh wow, I didn't realize they cached that long. You're right that ETags should solve that problem, so maybe we don't need to worry about it too much here.

dgp1130: If it's settable per-route, then that's probably good enough. Especially if we add ETag support, then there's likely not much motivation to shorten the caching period. We're probably fine as-is then.

packages/angular/ssr/src/assets.ts
dgp1130: Ultimately up to you, I don't feel too strongly, just something to think about.

goldens/public-api/angular/ssr/index.api.md
dgp1130: My suggestion is to narrow this down to just `process` unless we have a need to expose the other two pieces. I feel like they just provide an opportunity for a caller to misuse them. We can always re-add them in the future if we find a need.

The one use case I can think of is that some apps might want to disable SSG serving because they expect a CDN layer in front on this server to do that and want a hard-error from Angular if that was misconfigured. That's a pretty specific use case though and might make more sense as an option to `process` than exposing a separate `render` method?

dgp1130 (approved):
I gave some suggestions for the `process`, `serve`, `render` naming, but ultimately I'm happy to defer to you on what you think is best here.

goldens/public-api/angular/ssr/index.api.md
alan-agius4: Yes, that could be a fairly specific use case. As for naming, do you have any preference? `.render` feels a bit misleading, especially since it also deals with returning the pre-rendered page.

I'm leaning towards `.process`, `.handle`, or `.handleRequest`." any preference?

Issue #28730, opened 2024-10-28
Karma builder doesn't watch files with "builderMode: application"


cexbrayat:
### Command

test

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

Using the `"builderMode": "application",` option in the Karam builder breaks `ng test` watch capabilities

### Minimal Reproduction

Create a new project with `npx @angular/cli@19.0.0-next.13 new karma-watch --defaults --no-routing`
Add `"builderMode": "application",` to `angular.json` in the `karma`options.
Launch `ng test`
Update any file of the project like `app.component.ts` or `app.component.spec.ts`: the tests don't run again.

### Your Environment

Angular CLI: 19.0.0-next.13
Node: 20.13.1
Package Manager: npm 9.7.1
OS: darwin arm64

Angular: 19.0.0-next.11
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1900.0-next.13
@angular-devkit/build-angular   19.0.0-next.13
@angular-devkit/core            19.0.0-next.13
@angular-devkit/schematics      19.0.0-next.13
@angular/cli                    19.0.0-next.13
@schematics/angular             19.0.0-next.13
rxjs                            7.8.1
typescript                      5.6.3
zone.js                         0.15.0

jkrems:
I can repro:

```console
-- Watches for changes:
$ ng test --builder-mode=application --watch

-- Doesn't watch for changes:
$ ng test --builder-mode=application --watch

-- Watches for changes:
$ ng test --builder-mode=browser
```

So it looks like the `--watch` default is different.

jkrems:
Thanks for trying this out & the great bug details. Verified that the following fixes the bug: https://github.com/angular/angular-cli/pull/28739

jkrems:
The fix should be included in this week's `next` release (e.g. ~tomorrow).

PR #28739, opened 2024-10-29
fix(@angular-devkit/build-angular): fix --watch regression in karma

jkrems:
Outside of single-run mode, the karma test provider was expected to watch for changes. In the application builder branch, we only handled the case of an explicit `--watch` though. This meant that the karma runner was kept running but didn't see any file changes.

Fixes https://github.com/angular/angular-cli/issues/28730


clydin (approved):

jkrems:

PR #28745, opened 2024-10-30
refactor(@angular/build): directly bundle external component file-based style changes

clydin:
When using the development server with the application builder, file-based component styles will now be bundled during the main builder execution instead of within the application code bundling step. This allows for these styles to be processed independently from any code bundling steps and will support future changes that will allow the builder to completely skip code bundling if only file-based component stylesheets are changed.


alan-agius4 (approved):

clydin:

PR #28747, opened 2024-10-30
refactor(@angular/build): allow forcing bundler context rebundle

clydin:
A bundle call for a `BundlerContext` can now force a rebundling in cases where the cached version should not be used. This is currently not leveraged within the code but will be used for improvements to typescript based context rebuilds in the future.


alan-agius4 (approved):

clydin:

PR #28748, opened 2024-10-30
fix(@angular/build): set `ngServerMode` during vite prebundling

alan-agius4:
This commit sets the `ngServerMode` in the prebundled deps.


jkrems (approved):

alan-agius4:

PR #28757, opened 2024-10-31
refactor(@angular/ssr): add route configs to public API

alan-agius4:
These interfaces needs to be exported to be included in adev.


dgp1130 (approved):

alan-agius4:

PR #28758, opened 2024-10-31
fix(@angular/build): Exclude known `--import` from execArgv when spawning workers

alan-agius4:
This change prevents Bazel compilation issues in the ADEV build process.


clydin (approved):

alan-agius4:

PR #28761, opened 2024-10-31
refactor(@angular/build): allow experimental template update generation via environment variable

clydin:
The experimental Angular component template hot replacement capabilities will be initially controlled via an environment variable. Setting `NG_HMR_TEMPLATES=1` with the development server that is using the application builder will generate the runtime event code to support hot replacement of an individual component template. The build system itself does not yet generate the component update build results needed to trigger the runtime events. The environment variable is currently intended to support integration of the remaining code to fully implement the feature.


alan-agius4 (approved):

clydin:

PR #28763, opened 2024-11-01
refactor(@angular/ssr): remove duplicate code and streamline functionality

alan-agius4:
This commit cleans up duplicate code left from the previous implementations of process, serve, and render. Additionally, prerender serve now exclusively handles HEAD and GET requests, aligning with updated handling requirements. The private `renderStatic` method has been removed in favor of the `handle` method for improved maintainability.

### DIFF
--- a/packages/angular/ssr/src/app-engine.ts
+++ b/packages/angular/ssr/src/app-engine.ts
@@ -6,11 +6,10 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import type { AngularServerApp } from './app';
+import type { AngularServerApp, getOrCreateAngularServerApp } from './app';
 import { Hooks } from './hooks';
 import { getPotentialLocaleIdFromUrl } from './i18n';
 import { EntryPointExports, getAngularAppEngineManifest } from './manifest';
-import { stripIndexHtmlFromURL, stripTrailingSlash } from './utils/url';
 
 /**
  * Angular server application engine.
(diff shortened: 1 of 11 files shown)


packages/angular/ssr/src/app-engine.ts
```diff
@@ -24,23 +23,23 @@ import { stripIndexHtmlFromURL, stripTrailingSlash } from './utils/url';
...
-   * provide extensibility points for the application lifecycle.
+   * A flag to enable or disable the rendering of prerendered routes.
+   *
+   * Typically used during development to avoid prerendering all routes ahead of time,
+   * allowing them to be rendered on the fly as requested.
    *
    * @private
    */
-  static ɵhooks = /* #__PURE__*/ new Hooks();
+  static ɵallowStaticRouteRender = false;
```
dgp1130, line 33: Consider: Should this default to `true`? If someone naiively uses this class without any extra context, they would likely expect static rendering to work. It's only specific dev use cases where a user would want to go out of their way to disable static rendering.

packages/angular/ssr/src/app.ts
```diff
@@ -51,9 +51,21 @@ const SERVER_CONTEXT_VALUE: Record<RenderMode, string> = {
...
 export class AngularServerApp {
+  /**
+   * A flag to enable or disable the rendering of prerendered routes.
+   *
+   * Typically used during development to avoid prerendering all routes ahead of time,
+   * allowing them to be rendered on the fly as requested.
+   *
+   * @private
+   */
+  allowStaticRouteRender = false;
```
dgp1130, line 62: Consider: Could we make this a `readonly` constructor parameter? We probably wouldn't want this to change mid execution. Maybe it could also be directly marked `private` instead of needing `@private`?

packages/angular/ssr/src/app.ts
```diff
@@ -134,45 +131,45 @@ export class AngularServerApp {
...
+          'abort',
+          () => {
+            const abortError = new Error(
+              `Request for: ${request.url} was aborted.\n${request.signal.reason}`,
+            );
+            abortError.name = 'AbortError';
+            reject(abortError);
+          },
+          { once: true },
+        );
```
dgp1130, line 146: Consider: Move this to a helper function which creates a `Promise` from an `AbortSignal`.

dgp1130 (approved):
Sorry for the delay on this, just a few minor suggestions.

packages/angular/ssr/src/app-engine.ts
alan-agius4: This is actually to allow static pages to be renderer, not served. 

Only in the dev-server and builder should be done.

alan-agius4:

PR #28768, opened 2024-11-01
fix(@angular/cli): recommend optional application update migration during v19 update

clydin:
The optional application builder migration will now default to enabled and recommended during the update process (`ng update`) for v19. The migration is still optional and can be unselected if preferred when updating.


clydin:

PR #28774, opened 2024-11-02
refactor(@angular/build): split SSR server assets into separate chunks

alan-agius4:
This commit refactors the build process for server-side rendering (SSR) by dividing server assets into separate, importable chunks rather than bundling them into a single output file


dgp1130 (approved):
Any tests we can add to make sure imported assets are readable / paths line up correctly?

alan-agius4:
The already existing tests are dependent on these files and hence they are a validation that these work.

alan-agius4:

PR #28780, opened 2024-11-04
fix(@angular/build): workaround Vite CSS ShadowDOM hot replacement

clydin:
When using the development server with the application builder (default for new projects), Angular components using ShadowDOM view encapsulation will now cause a full page reload. This ensures that these components styles are correctly updated during watch mode. Vite's CSS hot replacement client code currently does not support searching and replacing `<link>` elements inside shadow roots. When support is available within Vite, an HMR based update for ShadowDOM components can be supported as other view encapsulation modes are now.

### DIFF
--- a/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
+++ b/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
@@ -15,7 +15,7 @@ export function createAngularAssetsMiddleware(
   server: ViteDevServer,
   assets: Map<string, string>,
   outputFiles: AngularMemoryOutputFiles,
-  usedComponentStyles: Map<string, Set<string>>,
+  usedComponentStyles: Map<string, Set<string | boolean>>,
   encapsulateStyle: (style: Uint8Array, componentId: string) => string,
 ): Connect.NextHandleFunction {
   return function angularAssetsMiddleware(req, res, next) {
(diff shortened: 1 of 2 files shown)


packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
```diff
@@ -76,14 +76,19 @@ export function createAngularAssetsMiddleware(
...
-            // Record the component style usage for HMR updates
+            // Track if the component uses ShadowDOM encapsulation (3 = ViewEncapsulation.ShadowDom)
+            const shadow = searchParams.get('e') === '3';
+
+            // Record the component style usage for HMR updates (true = shadow; false = none; string = emulated)
             const usedIds = usedComponentStyles.get(pathname);
+            const trackingId = componentId || shadow;
             if (usedIds === undefined) {
-              usedComponentStyles.set(pathname, new Set([componentId]));
+              usedComponentStyles.set(pathname, new Set([trackingId]));
```
alan-agius4, line 89: It might be clearer if we structure the data differently rather than using true/false/string, as a object to make it easier to follow:

```typescript
{
    encapsulation?: string;
    componentId?: string;
}
```

I can foresee that, in the future, this can be a bit confusing whlist debugging that as component ids we store booleans

alan-agius4 (approved):
LGTM. one NIT.

packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
clydin: I'm hoping we can revert this workaround sooner rather than later. Post v19 i may also look into patching the Vite client code to add support for shadow DOM with a potential followup Vite PR.

alan-agius4:

PR #28782, opened 2024-11-04
refactor(@angular/build): integrate template update function generation with builder results

clydin:
When using the development server with the application builder (default for new projects), experimental support for component template hot replacement is now available for use. To enable support, the `NG_HMR_TEMPLATES=1` environment variable must present when executing the development server (for example, `NG_HMR_TEMPLATES=1 ng serve`). Once support has become stable, template hot replacement will be enabled by default.


alan-agius4:

PR #28784, opened 2024-11-05
fix(@angular/build): always clear dev-server error overlay on non-error result

clydin:
Now that additional result object types can be returned from the build system, the error overlay must be cleared on all non-error results. The introduction of component update results and eventual incremental results will now properly clear the error overlay after a successful build.


alan-agius4 (approved):

alan-agius4:

PR #28787, opened 2024-11-05
fix(@angular/ssr): add validation to prevent use of `provideServerRoutesConfig` in browser context

alan-agius4:
Introduced an error check to ensure that 'provideServerRoutesConfig' is not utilized in the browser part of the application. This helps avoid unintended behavior.


alan-agius4:

PR #28792, opened 2024-11-05
refactor(@angular/ssr): Add `RequestHandlerFunction` and `NodeRequestHandlerFunction` to public API

alan-agius4:
These additions are necessary to ensure their inclusion in adev.


alan-agius4:

Issue #28755, opened 2024-10-31
Karma builder logs a lot of file changes with "builderMode: application"


cexbrayat:
### Command

test

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

When running `ng test` in watch mode, updating a file triggers a re-run of the tests as expected, but sometimes logs a lot of file changes.

### Minimal Reproduction

Create a new project with `npx @angular/cli@19.0.0-rc.0 new karma-assets --defaults --no-routing`.
Add `"builderMode": "application",` to angular.json in the karmaoptions.
Launch `ng test`
Clear the template `app.component.html` and check the logs.

### Exception or Error

31 10 2024 07:31:27.111:INFO [filelist]: Changed file "/Users/cedric/Code/temp/karma-assets/dist/test-out/c45dfc31-7cf6-461e-b12e-14642f1f29ac/app.component.spec.js.map".
31 10 2024 07:31:27.112:INFO [filelist]: Changed file "/Users/cedric/Code/temp/karma-assets/dist/test-out/c45dfc31-7cf6-461e-b12e-14642f1f29ac/styles.css.map".
31 10 2024 07:31:27.112:INFO [filelist]: Changed file "/Users/cedric/Code/temp/karma-assets/dist/test-out/c45dfc31-7cf6-461e-b12e-14642f1f29ac/styles.css".
31 10 2024 07:31:27.113:INFO [filelist]: Changed file "/Users/cedric/Code/temp/karma-assets/dist/test-out/c45dfc31-7cf6-461e-b12e-14642f1f29ac/init_test_bed.js".
31 10 2024 07:31:27.113:INFO [filelist]: Changed file "/Users/cedric/Code/temp/karma-assets/dist/test-out/c45dfc31-7cf6-461e-b12e-14642f1f29ac/app.component.spec.js".
31 10 2024 07:31:27.113:INFO [filelist]: Changed file "/Users/cedric/Code/temp/karma-assets/dist/test-out/c45dfc31-7cf6-461e-b12e-14642f1f29ac/app.component.css.map".
31 10 2024 07:31:27.114:INFO [filelist]: Changed file "/Users/cedric/Code/temp/karma-assets/dist/test-out/c45dfc31-7cf6-461e-b12e-14642f1f29ac/polyfills.js".
[...]

PR #28794, opened 2024-11-05
fix(@angular-devkit/build-angular): remove double-watch in karma

jkrems:
The Karma file watching was racing with the file writes done by the application builder. Since we already tell Karma when to reun via `.refeshFiles()`, disabling Karma's own file watcher should make things more reliable.

This allows removing a weird special-case in the test case and removes the noisy "File chaned" logs generated by Karma.

Fixes https://github.com/angular/angular-cli/issues/28755

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
+++ b/packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
@@ -9,6 +9,7 @@
 import { BuildOutputFileType } from '@angular/build';
 import {
   ApplicationBuilderInternalOptions,
+  Result,
   ResultFile,
   ResultKind,
   buildApplicationInternal,
@@ -42,6 +43,7 @@ class ApplicationBuildError extends Error {
 function injectKarmaReporter(
   context: BuilderContext,
   buildOptions: BuildOptions,
+  buildIterator: AsyncIterator<Result>,
   karmaConfig: Config & ConfigOptions,
   subscriber: Subscriber<BuilderOutput>,
 ) {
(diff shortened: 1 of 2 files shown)


clydin (approved):

jkrems:
Increased the debounce window in the test which shouldn't fundamentally slow the test down. The race between Karma starting to run, the watching build, and the browser running tests may not be worth fixing since it likely requires larger refactoring. Which may end up ugly unless we change Karma as well (and I'm pretty sure we don't want to do that at this point).

jkrems:
The end result is a bit more complex now, mostly to remove a race condition that caused the test instability:

1. The initial build completes.
2. During the initial _watching_ build, a file is modified.
3. The file change is never picked up - there are no further builds and the test cannot make progress.

This is likely not a super common case for real users but correctness-wise, it's likely better to set up the file watching at the very beginning. Doing so requires reusing the same `buildApplicationInternal` call for the entire test run which makes for ugly code (manual iterator loop instead of a neat `for await ... of`).

packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
```diff
@@ -64,13 +66,15 @@ function injectKarmaReporter(
...
-          context,
-        )) {
+        // This is effectively "for await of but skip what's already consumed".
+        let isDone = false; // to mark the loop condition as "not constant".
+        while (!isDone) {
+          const { done, value: buildOutput } = await buildIterator.next();
+          if (done) {
+            isDone = true;
+            break;
+          }
```
clydin, line 76: Can this `for..await` the buildIterator here or is the manual iterator usage required?

clydin (approved):

packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
jkrems: Unfortunately `for await..of` needs an `AsyncIterable` because it wants to create a fresh (and reset) `AsyncIterator` itself. From what I can gather, we can't (easily) use the nicer syntax here.

jkrems:

Issue #28754, opened 2024-10-31
Karma builder ignores assets directory with "builderMode: application"


cexbrayat:
### Command

test

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

The karma builder has an option to define a assets directories.
By default, it uses: `{"glob": "**/*", "input": "public" }`.

When using `"builderMode: application"`, this option is ignored and the Karma server throws 404 for the assets used in tested components.

### Minimal Reproduction

Generate a new project with `npx @angular/cli@19.0.0-rc.0 new karma-assets --defaults --no-routing`.
Add `"builderMode": "application",` to angular.json in the karma options.
Rename `public/favicon.ico`  to `public/favicon.png`.
Replace the `app.component.html` template with `<img src="favicon.png" alt="Angular Logo" />`.
Replace the `app.component.spec.ts` with:

```ts
import { TestBed } from '@angular/core/testing';
import { AppComponent } from './app.component';

describe('AppComponent', () => {
  beforeEach(() => TestBed.configureTestingModule({}));

  it('should create the app', () => {
    const fixture = TestBed.createComponent(AppComponent);
    fixture.detectChanges();
    expect(fixture.componentInstance).toBeTruthy();
  });
});
```

Launch `ng test`

### Exception or Error

31 10 2024 07:20:02.816:WARN [karma]: No captured browser, open http://localhost:9876/
31 10 2024 07:20:02.825:INFO [karma-server]: Karma v6.4.4 server started at http://localhost:9876/
31 10 2024 07:20:02.825:INFO [launcher]: Launching browsers ChromeHeadless with concurrency unlimited
31 10 2024 07:20:02.827:INFO [launcher]: Starting browser ChromeHeadless
31 10 2024 07:20:03.544:INFO [Chrome Headless 130.0.0.0 (Mac OS 10.15.7)]: Connected on socket V7qJDTy8680sTot9AAAB with id 30066879
31 10 2024 07:20:03.699:WARN [web-server]: 404: /favicon.png

### Your Environment

Angular CLI: 19.0.0-rc.0
Node: 20.13.1
Package Manager: npm 9.7.1
OS: darwin arm64

Angular: 19.0.0-rc.0
... animations, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1900.0-rc.0
[...]

PR #28795, opened 2024-11-05
fix(@angular-devkit/build-angular): serve assets

jkrems:
Turns out the test for asset serving didn't actually test asset serving because of a missing `whenStable()` - it always succeeded, even with random filenames. This change adds a custom middleware to serve our static files since it seemed easier (and more predictable) than to try to map the `assets` concept to Karma's `files`+`proxy` logic.

One side effect of adding the middleware is that we don't actually need the Karma `files` anymore with the exception of any files that we'd want to include on page load. I don't touch that part here because it will be cleaner to change it when also addressing the naming collisions between spec files.

Fixes https://github.com/angular/angular-cli/issues/28754

~~P.S.: Currently the first commit is from a different PR (https://github.com/angular/angular-cli/pull/28794), I'm planning to rebase after landing it.~~


clydin (approved):

jkrems:

Issue #28756, opened 2024-10-31
Karma builder can't handle multiple spec files with same name with "builderMode: application"


cexbrayat:
### Command

test

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

This issue is similar to https://github.com/angular/angular-cli/issues/26810

When a project has several spec files with the same name, the esbuilder outputs them to the same location and crashes.

### Minimal Reproduction

Generate a new project with `npx @angular/cli@19.0.0-rc.0 new karma-collision --defaults --no-routing`.
Add `"builderMode": "application"`, to angular.json in the karma options.
Add a new component with `ng g c other/app`
Run `ng test`

### Exception or Error

An unhandled exception occurred: `/Users/cedric/Code/temp/karma-collision/src/app/app.component.spec.ts` and `/Users/cedric/Code/temp/karma-collision/src/app/other/app/app.component.spec.ts` both output to the same location `app.component.spec`. Rename or move one of the files to fix the conflict.
See "/private/var/folders/fl/sx8_jbfd1gg3gg04jb0lr_n00000gn/T/ng-m92j7I/angular-errors.log" for further details.

### Your Environment

Angular CLI: 19.0.0-rc.0
Node: 20.13.1
Package Manager: npm 9.7.1
OS: darwin arm64

Angular: 19.0.0-rc.0
... animations, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1900.0-rc.0
@angular-devkit/build-angular   19.0.0-rc.0
@angular-devkit/core            19.0.0-rc.0
@angular-devkit/schematics      19.0.0-rc.0
@schematics/angular             19.0.0-rc.0
rxjs                            7.8.1
typescript                      5.6.3
zone.js                         0.15.0

PR #28797, opened 2024-11-05
fix(@angular-devkit/build-angular): handle basename collisions

jkrems:
Generates unique names for each test entry point and also creates a nice & consistent name for the main ("init test bed") chunk. All specs are now `spec-*.js` files.

Fixes https://github.com/angular/angular-cli/issues/28756

~~P.S.: Current based on https://github.com/angular/angular-cli/pull/28795, will rebase post-merge.~~

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
+++ b/packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
@@ -260,7 +260,7 @@ async function collectEntrypoints(
   options: KarmaBuilderOptions,
   context: BuilderContext,
   projectSourceRoot: string,
-): Promise<Set<string>> {
+): Promise<Map<string, string>> {
   // Glob for files to test.
   const testFiles = await findTests(
     options.include ?? [],
@@ -269,7 +269,28 @@ async function collectEntrypoints(
     projectSourceRoot,
   );
 
-  return new Set(testFiles);
+  const seen = new Set<string>();
+
+  return new Map(
+    Array.from(testFiles, (testFile) => {
+      const relativePath = path
+        .relative(
+          testFile.startsWith(projectSourceRoot) ? projectSourceRoot : context.workspaceRoot,
+          testFile,
+        )
+        .replace(/^[./]+/, '_')
+        .replace(/\//g, '-');
+      let uniqueName = `spec-${path.basename(relativePath, path.extname(relativePath))}`;
+      let suffix = 2;
+      while (seen.has(uniqueName)) {
+        uniqueName = `${relativePath}-${suffix}`;
+        ++suffix;
+      }
+      seen.add(uniqueName);
+
+      return [uniqueName, testFile];
+    }),
+  );
 }
 
 async function initializeApplication(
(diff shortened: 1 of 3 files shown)


jkrems:
Adding @dgp1130 as another reviewer in case he has thoughts on the changes to the `entryPoints` option (based on Chat discussion with @clydin).

packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
```diff
@@ -356,25 +376,16 @@ async function initializeApplication(
     { pattern: `${outputPath}/polyfills.js`, type: 'module', watched: false },
     // Serve global setup script.
     { pattern: `${outputPath}/${mainName}.js`, type: 'module', watched: false },
-    // Serve all source maps.
-    { pattern: `${outputPath}/*.map`, included: false, watched: false },
```
dgp1130, line 360: Question: Are we intentionally removing sourcemaps here?

dgp1130 (commented):
The use of a `Map` to explicitly define output entry points seems generally reasonable. But I'm wondering if we should allow `Set` at all? I think this was already an issue for the Jest and WTR builders which I hadn't gotten to fixing yet and it makes me think we should never use the `Set` case as its implemented today or we'll potentially run into the same bug. I'm thinking we should either:

1. Remove `Set` from the parameter type altogether and require a `Map`. All callers need to manually construct this `Map`.
2. Update `normalizeEntryPoints` to find and deduplicate basenames in the set (`foo.spec.js` becomes `foo.spec-2.js`) so the caller doesn't need to worry about this.

I'd probably lean towards 2., given that Jest and WTR are likely to duplicate the work done in the Karma builder, and if we go with 2. they likely wouldn't need to worry about this given that none of the test runners really care about the output file name too much.

jkrems:
A potential issue with 2 is that the test runners (or, at least Karma) need to be able to find the entrypoints afterwards to load them. Which makes me lean towards refactoring the Jest/WTR builders to use the Map as well.

Either way, happy to do a potential follow-up if necessary to ensure that we only have one or the other at the end of this. Agreed that we should pick one data structure, not both.

dgp1130 (approved):
> A potential issue with 2 is that the test runners (or, at least Karma) need to be able to find the entrypoints afterwards to load them. Which makes me lean towards refactoring the Jest/WTR builders to use the Map as well.

I did have a similar thought, but I feel it's manageable if we either maintain the relevant extension (`.spec.js`), though that might be a little tricky without codifying that particular extension into `application` builder, or we could potentially dump these into a subdirectory like `dist/entry-points/...` where the exact file name doesn't matter because all the entry points are there instead of `dist/chunk-*.js`. Not sure exactly which of these would be easier, but it's a potential option.

I'm fine to tackle that as a follow up, it would definitely be great to have a common solution for all three test runners.

clydin (approved):

jkrems:
> I'm fine to tackle that as a follow up, it would definitely be great to have a common solution for all three test runners.

FWIW, that is a general theme for the application builder support for Karma. I've been lifting code (e.g. `writeTestOutputs`) from the other runners and it would likely be nice to refactor to make sure we don't have unnecessary duplication.

packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts
jkrems: Originally: Yes. I assumed that serving static files using the Karma middleware would "just work" now. But Karma is certainly... creative and the one thing I can say with certainty is that it doesn't _just work_. So, given how annoying it is to test source maps in particular, let me back out these changes.

jkrems:

PR #28799, opened 2024-11-06
fix(@angular/ssr): initialize the DI tokens with `null` to avoid requiring them to be set to optional

alan-agius4:
This commit initializes the `REQUEST`, `RESPONSE_INIT`, and `REQUEST_CONTEXT` DI tokens with `null` values by default.

Previously, these tokens lacked default values, which meant that developers had to explicitly mark them as optional when injecting them in constructors. This was necessary to avoid errors when these tokens were not provided, particularly in scenarios like client-side rendering or during development.

By initializing these tokens with `null`, we eliminate the need for developers to mark them as optional, simplifying the code and improving the developer experience.

### DIFF
--- a/goldens/public-api/angular/ssr/tokens/index.api.md
+++ b/goldens/public-api/angular/ssr/tokens/index.api.md
@@ -7,13 +7,13 @@
 import { InjectionToken } from '@angular/core';
 
 // @public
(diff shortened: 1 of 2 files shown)


dgp1130 (approved):

alan-agius4:

PR #28800, opened 2024-11-06
fix(@angular/build): skip wildcard routes from being listed as prerendered routes

alan-agius4:
This fix ensures that the wildcard routes are not included in the prerendered list, improving the accuracy of prerendered route generation.


clydin (approved):

alan-agius4:

PR #28814, opened 2024-11-07
build: update `rollup-license-plugin` to `3.0.1`

alan-agius4:
There is no need for the local patch as the issue was fixed upstream


clydin (approved):

alan-agius4:

PR #28816, opened 2024-11-08
fix(@angular/build): avoid overwriting inline style bundling additional results

clydin:
A component may contain multiple inline styles that will originate from the same containing file. The content of the processed style is sent directly to the Angular compiler. However, any additional result files are collected and emitted in the application output. In most cases, this worked as expected as inline styles rarely had resource references that would overwrite each other. However, the potential is present for later inline styles for a component to overwrite these output files. To avoid this potential problem, the internal identifier now accounts for both the class name and order of the inline styles. This ensures that each inline style retains a unique additional results entry.


alan-agius4:

PR #28818, opened 2024-11-08
refactor(ssr): remove `RenderMode.AppShell` in favor of new configuration option

alan-agius4:
This commit removes the `RenderMode.AppShell` option. Instead, a new configuration parameter, `{ appShellRoute: 'shell' }`, is introduced to the `provideServerRoutesConfig` method.

```ts
provideServerRoutesConfig(serverRoutes, { appShellRoute: 'shell' })
```

### DIFF
--- a/packages/angular/ssr/public_api.ts
+++ b/packages/angular/ssr/public_api.ts
@@ -14,9 +14,9 @@ export { createRequestHandler, type RequestHandlerFunction } from './src/handler
 export {
   type PrerenderFallback,
   type ServerRoute,
+  type ServerRoutesConfigOptions,
   provideServerRoutesConfig,
   RenderMode,
-  type ServerRouteAppShell,
   type ServerRouteClient,
   type ServerRoutePrerender,
   type ServerRoutePrerenderWithParams,
--- a/packages/angular/ssr/src/routes/route-config.ts
+++ b/packages/angular/ssr/src/routes/route-config.ts
@@ -15,9 +15,6 @@ import { EnvironmentProviders, InjectionToken, makeEnvironmentProviders } from '
  * @developerPreview
  */
 export enum RenderMode {
-  /** AppShell rendering mode, typically used for pre-rendered shells of the application. */
-  AppShell,
-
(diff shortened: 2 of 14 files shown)


packages/angular/ssr/public_api.ts
```diff
@@ -14,12 +16,14 @@ export { createRequestHandler, type RequestHandlerFunction } from './src/handler
...
   RenderMode,
-  type ServerRouteAppShell,
   type ServerRouteClient,
   type ServerRoutePrerender,
   type ServerRoutePrerenderWithParams,
   type ServerRouteServer,
   type ServerRouteCommon,
 } from './src/routes/route-config';
+
+provideZoneChangeDetection();
```
dgp1130, line 29: Question: How does Zone relate to this app shell refactoring? Should this maybe be in a different commit? Should Zones be more opt-in to make future Zoneless support easier?

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -15,17 +15,14 @@ import { EnvironmentProviders, InjectionToken, makeEnvironmentProviders } from '
...
-  Server,
+  Server = 1,
 
   /** Client-Side Rendering (CSR) mode, where content is rendered on the client side in the browser. */
-  Client,
+  Client = 2,
 
   /** Static Site Generation (SSG) mode, where content is pre-rendered at build time and served as static files. */
-  Prerender,
+  Prerender = 3,
```
dgp1130, line 25: Question: Do we need to maintain theses indexes? Should we re-index on 0?

dgp1130 (approved):

packages/angular/ssr/public_api.ts
alan-agius4: That’s a mistake.

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: We could

alan-agius4:

PR #28845, opened 2024-11-12
fix(@angular/ssr): allow using function with optional request context argument

pieh:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using `createRequestHandler` on Netlify edge function compatible request handler function I am forced to use `@ts-expect-error` directive currently:

```ts
import { AngularAppEngine, createRequestHandler } from '@angular/ssr'
const angularAppEngine = new AngularAppEngine()

// @ts-expect-error - createRequestHandler expects a function with single Request argument and doesn't allow context argument
export const reqHandler = createRequestHandler(async (request: Request, context: any) => {
  const result = await angularAppEngine.handle(request, context)
  return result || new Response('Not found', { status: 404 })
})
```

Issue Number: N/A

## What is the new behavior?

Don't need to use `@ts-expect-error` directive anymore.

```ts
import { AngularAppEngine, createRequestHandler } from '@angular/ssr'
const angularAppEngine = new AngularAppEngine()

export const reqHandler = createRequestHandler(async (request: Request, context: any) => {
  const result = await angularAppEngine.handle(request, context)
  return result || new Response('Not found', { status: 404 })
})
```

- [ ] Yes
- [x] No

## Other information

From what I understand applying metadata done by `createRequestHandler` is used to recognize in dev mode that exported handle can be used to handle requests and falling back to "default" when it's not (and ignoring any customizations in `server.ts` file in dev mode because of it), so there doesn't seem to exist workaround the current type while still having access to second parameter?


packages/angular/ssr/src/handler.ts
```diff
@@ -16,6 +16,8 @@
  */
 export type RequestHandlerFunction = (
   request: Request,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ...ignoredArgs: any[]
```
alan-agius4, line 20: Instead of using `any` here, can we instead use generics to infer the type? 

Example

```ts
export type RequestHandlerFunction<T = unknown> = (
  request: Request,
  requestContext?: T,
) => Promise<Response | null> | null | Response;

export function createRequestHandler<T = unknown>(
  handler: RequestHandlerFunction<T>,
): RequestHandlerFunction<T> {

}
```

pieh:
I found alternative that doesn't require type modification - if anyone is interested it's something like this (attempted to simplify it in case other would face similar issue to solve):

```ts
import { AsyncLocalStorage } from 'node:async_hooks'

import { AngularAppEngine, createRequestHandler } from '@angular/ssr'
import type { Context } from '@netlify/edge-functions'

const angularAppEngine = new AngularAppEngine()
const ContextStorage = new AsyncLocalStorage<Context>()

function requestHandlerImpl(request: Request) {
  const context = ContextStorage.getStore()
  // context will be undefined in dev, but should be defined in prod

  return angularAppEngine.handle(request, context)
}

// this will be production deployment entry point - it sets context on the async local storage and we can access it later in common handler implementation without passing it as function argument
export default function prodHandler(request: Request, context: Context) {
  return ContextStorage.run(context, () => requestHandlerImpl(request))
}

// dev
export const reqHandler = createRequestHandler(requestHandlerImpl)
```

So I'll close this PR as given that there is pretty portable solution that can be adjusted to work with any provider - attempting to adjust the types for specific signature is unnecessary risk to take and would be annoying to deal with in the future if another incompatible signature would need to be handled.

PR #28854, opened 2024-11-13
feat: add aot option to karma

yjaaidi:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Tests run in JIT only.

## What is the new behavior?

Tests can run in AOT, using the new `aot` option.

- [ ] Yes
- [X] No


packages/angular_devkit/build_angular/src/builders/karma/tests/options/aot_spec.ts
```diff
@@ -0,0 +1,47 @@
...
+          import { Component } from '@angular/core';
+          
+          describe('Hello', () => {
+            it('should *not* contain jit instructions', () => {
+              @Component({
+                template: 'Hello',
+              })
+              class Hello {}
+
+              expect((Hello as any).ɵcmp.template.toString()).not.toContain('jit');
```
yjaaidi, line 29: This is fragile and sensitive to any structural changes in JIT. It could easily cause false negatives.
What would be a better way to achieve this detection?

packages/angular_devkit/build_angular/src/builders/karma/tests/options/aot_spec.ts
```diff
@@ -0,0 +1,47 @@
...
+            });
+          });
+`,
+      });
+
+      harness.useTarget('test', {
+        ...BASE_OPTIONS,
+        aot: true,
+        /** Cf. {@link ../builder-mode_spec.ts} */
+        polyfills: ['zone.js', '@angular/localize/init', 'zone.js/testing'],
```
yjaaidi, line 39: Should we factorize this in `BASE_OPTIONS`? or create a `BASE_POLYFILLS`? or an object mother: `optionsMother.base().withPolyfills()`?

jkrems:
Sorry for the lack of review so far but I just want to quickly say that I'm really excited to see this option added. `aot` all the things! 🎉  - *goes back to working through final fixes for the rc candidate*

yjaaidi:
No rush at all!

packages/angular_devkit/build_angular/src/builders/karma/tests/options/aot_spec.ts
jkrems: `describeKarmaBuilder` already runs the builder in both modes automatically. You shouldn't need any override but `aot: true` in here if everything works as expected. The builder mode spec is special because it explicitly tests that the mode can be changed by explicitly setting this. But it seems wasteful to rerun this test case twice in application builder mode (which I think is what's currently happening with this file).

packages/angular_devkit/build_angular/src/builders/karma/tests/options/aot_spec.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { execute } from '../../index';
+import { BASE_OPTIONS, KARMA_BUILDER_INFO, describeKarmaBuilder } from '../setup';
+import { BuilderMode } from '../../schema';
+
+describeKarmaBuilder(execute, KARMA_BUILDER_INFO, (harness, setupTarget) => {
```
jkrems, line 13: If you want to restrict this to application builder mode only, you might need to add the 3rd argument for this callback (`isApplicationBuilder`) and skip/disable the test if that boolean is false.

packages/angular_devkit/build_angular/src/builders/karma/tests/options/aot_spec.ts
jkrems: I think as long as we have a second test case that tests the reverse, it's fine..? That way we can make sure that "jit" appears in one branch and doesn't appear in the other, with `aot: true` vs `aot: false` as the only difference. That seems like an okay state to be in for now.

jkrems (commented):
LGTM - sorry for the late review!

yjaaidi:
Thx! No problem at all.

jkrems (approved):

jkrems:
Sorry for the delay here. Approved the change but there's a tiny conflict and I don't think I can resolve it without push access to the branch. Can you undo my merge commit (e.g. `git reset --hard 48d0a96b53f9464f334248999408aa1c3c597bbf`) and then rebase?

packages/angular_devkit/build_angular/src/builders/karma/tests/options/aot_spec.ts
yjaaidi: I've added the jit test to make sure jit symbols are there.

jkrems:
Thanks for the updated test! Can you rebase instead of doing the merge?

Also, given that the base version landed (https://github.com/angular/angular-cli/pull/29608), I think the commit message could be reworded to be `test:`.

yjaaidi:
> Thanks for the updated test! Can you rebase instead of doing the merge?
> 
> Also, given that the base version landed (#29608), I think the commit message could be reworded to be `test:`.

Hey @jkrems, sorry for the delay.
I've updated the test, and squash rebased.

jkrems (approved):

PR #28859, opened 2024-11-14
build: update ng-dev to fix release tool

jkrems:
Includes: https://github.com/angular/dev-infra/pull/2446

Equivalent to https://github.com/angular/angular/pull/58651 in the framework repo.


jkrems:
Actually, I can just land https://github.com/angular/angular-cli/pull/28856 instead.

Issue #28870, opened 2024-11-15
Unmet peer dependency issue with angular-cli and @listr2/prompt-adapter-inquirer


arobinson:
### Command

other

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

```
├─┬ @angular/cli 19.0.0-rc.2
│ └─┬ @listr2/prompt-adapter-inquirer 2.0.17
│   └── ✕ unmet peer @inquirer/prompts@">= 3 < 7": found 7.1.0
```
RC2 is using 2.0.17 of `@listr2/prompt-adapter-inquirer` but using an incompatible version of `@inquirer/prompts`

### Minimal Reproduction

```
pnpm i -g @angular/cli@next
 WARN  1 deprecated subdependencies found: puppeteer@13.7.0
Packages: +109 -61
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-------------------------------------------------------------
Progress: resolved 500, reused 500, downloaded 0, added 109, done

/Users/andrew/.local/share/pnpm/global/5:
- @angular/cli 18.0.0
+ @angular/cli 19.0.0-rc.2

 WARN  Issues with peer dependencies found
.
└─┬ @angular/cli 19.0.0-rc.2
  └─┬ @listr2/prompt-adapter-inquirer 2.0.17
    └── ✕ unmet peer @inquirer/prompts@">= 3 < 7": found 7.1.0
```

### Exception or Error

```text
WARN  Issues with peer dependencies found
```

### Your Environment

```text
_                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/


Angular CLI: 19.0.0-rc.2
Node: 22.11.0
Package Manager: npm 10.9.0
OS: darwin arm64

Angular:
...

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1900.0-rc.2
@angular-devkit/core         19.0.0-rc.2
@angular-devkit/schematics   19.0.0-rc.2
@schematics/angular          19.0.0-rc.2
```

jkrems:
[...]

PR #28872, opened 2024-11-15
build: fix peer dep warning for listr2

jkrems:
The previous version specified `@inquirer/prompts < 7` which caused a warning about invalid peer deps. The only difference between 2.0.17 and 2.0.18 is that this range was expanded to `< 8`.

See: https://github.com/listr2/listr2/issues/730

Fixes: https://github.com/angular/angular-cli/issues/28870

Commit history: https://github.com/listr2/listr2/commits/@listr2%2fprompt-adapter-inquirer@2.0.18/packages/prompt-adapter-inquirer

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "@discoveryjs/json-ext": "0.6.3",
     "@inquirer/confirm": "5.0.2",
     "@inquirer/prompts": "7.1.0",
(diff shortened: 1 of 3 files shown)


jkrems:
I think this would be good to include so that 19.0.0 doesn't greet people with a peer dep error/warning as the first impression.

dgp1130 (approved):

jkrems:

PR #28873, opened 2024-11-15
fix(@angular-devkit/build-angular): bring back style tags in browser builder

jkrems:


dgp1130:
Thanks for looking into this! Seems like a tricky issue.

I feel like this comment is particularly telling: https://github.com/angular/angular/blob/92f30a749d676a290f5e173760ca29f0ff85ba8c/packages/compiler-cli/src/ngtsc/annotations/component/src/resources.ts#L435-L436

"If the preload worked" seems to mean "If it returned a `Promise`", which seems wrong to me? The timing of the function shouldn't indicate whether or not it was successful. I wonder if we should change that file to parse the template regardless of whether `readResource` returned `undefined` or `Promise<undefined>`. WDYT?

I'm not sure I'm seeing the "preload didn't work" cause in `readResource`, but if necessary, maybe we need to return whether or not it was successful independently of whether it returned a `undefined` or `Promise<undefined>`?

If this is sufficient to get the test to pass, I'm probably ok to merge it with `Promise.resolve(content)` as the fix just to avoid blocking v19, but we should definitely get a look from Charles on Monday before we merge it.

jkrems:
If nothing else, I believe this kind of "undefined and Promise<undefined> have two different meanings" design deserves a closer look. I don't think there's a reason for this kind of API without much of a hint about the semantics of the values. The change here is meant to be least disruptive to reduce risk for v19. I hope we can fix this properly in the coming weeks even if we land it.

clydin (approved):

jkrems:

PR #28876, opened 2024-11-15
fix: use default import for `express`

dgp1130:
Now that we're using `esModuleInterop`, a default import is required.


dgp1130:
Targeting RC for now, this might make sense for v18 projects or earlier, but we can backport later as needed.

jkrems (approved):

jkrems:

PR #28878, opened 2024-11-16
fix(@schematics/angular): don't show server routing prompt when using `browser` builder

dgp1130:
The new routing APIs don't support `browser` builder, but calling `ng add @angular/ssr` with a `browser` builder would still prompt the user to add them. If the user said "Yes", it would actually ignore that answer and not enable the new APIs.

With this change, `ng add @angular/ssr` when using `browser` builder does not show the prompt and assumes the answer is "No". It also throws an error if the user runs `ng add @angular/ssr --server-routing`.

I'm not aware of a built-in prompting mechanism in schematics beyond `x-prompt`, which can't be used here, so instead I just called Inquirer directly. Unfortunately testing the prompt is a little awkward, as Inquirier does not provide useful APIs in this space. I evaluated `@inquirer/testing`, but ultimately decided that was more intended for testing custom Inquirer prompts, not mocking usage of standard prompts. Schematics APIs do not provide a useful way to inject additional data like a mock, so instead I had to do this through a `setPrompterForTestOnly` function. I'm not a huge fan of it, but I don't see a more straightforward way of solving the problem.

### DIFF
--- a/packages/angular/ssr/schematics/ng-add/index_spec.ts
+++ b/packages/angular/ssr/schematics/ng-add/index_spec.ts
@@ -14,6 +14,7 @@ import { join } from 'node:path';
 describe('@angular/ssr ng-add schematic', () => {
   const defaultOptions = {
     project: 'test-app',
+    serverRouting: false,
   };
 
   const schematicRunner = new SchematicTestRunner(
--- a/packages/schematics/angular/BUILD.bazel
+++ b/packages/schematics/angular/BUILD.bazel
@@ -84,6 +84,7 @@ ts_library(
         "//packages/angular_devkit/schematics",
         "//packages/angular_devkit/schematics/tasks",
         "//packages/schematics/angular/third_party/github.com/Microsoft/TypeScript",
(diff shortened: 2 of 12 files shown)


jkrems:
Looks like some test(s) hang with interactive prompts right now?

```
Running test "tests/build/wasm-esm" (4 of 6 [2:24] (75/123))...
Running `ng "build"` [silent]...
CWD: /tmp/angular-cli-e2e-XXXXXXAvFCLu/e2e-test/test-project
Running `ng "e2e"` [silent]...
CWD: /tmp/angular-cli-e2e-XXXXXXAvFCLu/e2e-test/test-project
Running `ng "add" "@angular/ssr" "--skip-confirmation"`...
CWD: /tmp/angular-cli-e2e-XXXXXXAvFCLu/e2e-test/test-project
  Skipping installation: Package already installed
  ? Would you like to use the Server Routing and App Engine APIs (Developer
  Preview) for this server application? (y/N)
-- Test timed out at 2024-11-16 08:40:36 UTC --
```

PR #28879, opened 2024-11-16
fix(@angular-devkit/build-angular): fix hanging terminal when `browser-sync` is not installed

dgp1130:
Running the SSR dev server when `browser-sync` is not installed would print the error, but then build the browser and server targets, then hang and never return control to the user until they manually Ctrl+C. This change skips building at all if `browser-sync` is not installed, immediately returning control to the user.

This is a simple workaround, but there are two deeper bugs which would benefit from investigation:
1.  Figure out why NPM sometimes doesn't install `browser-sync`. It was happening inconsistently for me when running `ng add @angular/ssr`.
2.  Figure out why Architect does not cancel/await targets still executing when a builder completes.

I'm not sure if this fix should be backported to older versions, but we can worry about that after the v19 launch.


alan-agius4 (approved):

dgp1130:

PR #28886, opened 2024-11-18
fix(@angular/build): fully disable component style HMR in JIT mode

clydin:
The component stylesheet HMR functionality requires build-time analysis of each component by the AOT compiler to provide the needed information to identify initial styles and detect individual changes to each style. Part of the style HMR rebuild logic was unintentionally enabled in JIT mode. The initial load of the application operated correctly but subsequent changes to file-based stylesheets were delayed by one rebuild cycle. To avoid this misalignment, all component stylesheet HMR functionality is now disabled when in JIT mode.


dgp1130 (approved):

dgp1130:

PR #28907, opened 2024-11-20
fix(@angular/ssr): handle baseHref that start with `./`

alan-agius4:
Updated function to support handling `baseHref` starting with './' path correctly.


dgp1130 (approved):

alan-agius4:

PR #28926, opened 2024-11-21
fix(@angular/build): use `sha256` instead of `sha-256` as hash algorithm name

clydin:
Stackblitz appears to fail when attempting to use `crypto.createHash` with an algorithm value of `sha-256`. Since Node.js supports both the hyphenated and unhyphenated values, the later is now used to avoid issues when running on Stackblitz.


alan-agius4 (approved):

clydin:

PR #28928, opened 2024-11-21
ci: set up perf workflow job and create initial workflow

josephperrott:
Set up the initial workflow to build upon.


clydin (approved):

.ng-dev/dx-perf-workflows.yml
```diff
@@ -0,0 +1,7 @@
+workflows:
+  build-adev:
```
clydin, line 2:
```suggestion
  build-cli:
```

PR #28932, opened 2024-11-22
fix(@schematics/angular): skip SSR routing prompt in webcontainer

dgp1130:
Apparently `inquirer` requires `async_hooks` which isn't supported in webcontainers, therefore prompting the user fails. Instead we always fall back to the default option.

See: https://github.com/SBoudrias/Inquirer.js/issues/1426


clydin (approved):

clydin:

PR #28935, opened 2024-11-22
refactor(@angular/ssr): replace `:params` with `*`

alan-agius4:
The SSR router relies on wildcard matching rather than params. This commit refactors the stored routes by removing the `:params`.


dgp1130 (approved):

dgp1130:

Issue #28938, opened 2024-11-22
Support tailwind v4


heddendorp:
### Command

build

### Description

Tailwind v4 is changing the installation and configuration. It would be great to update the cli to make sure v4 is supported.

### Describe the solution you'd like

The cli should support the new postcss plugin https://tailwindcss.com/docs/v4-beta#installing-with-post-css
Also detection probably has to be changed as v4 no longer uses js to configure tailwind but instead css variables https://tailwindcss.com/docs/v4-beta#installing-with-post-css
I guess the best detection at this point is having tailwindcss with a version >4 installed and ideally the cli also makes sure @tailwindcss/postcss is installed.

### Describe alternatives you've considered

Tailwind also offers a vite plugin, but to my understanding, postcss should be easier to support

clydin:
Custom postcss configurations are possible with the `application` build system (default for all new projects).  The `@tailwindcss/postcss` plugin can be used directly by creating a `postcss.config.json`/``.postcssrc.json` file and including the plugin.

As tailwindcss v4 moves towards a stable release, we are also evaluating direct inclusion of the optimized pipeline present in the Vite plugin to avoid the need for postcss if preferred.

pumano:
@clydin looks like `vite` plugin should be more performant than `postcss`

dcheglakov:
@clydin I’ve tried following your approach with the `postcss.config.json` file and enabling Tailwind CSS via PostCSS like this:
```json
{
  "plugins": {
    "@tailwindcss/postcss": {}
  }
}
```
for my Angular 16 project. Unfortunately, it didn’t resolve the issue for me. Maybe there’s something specific to Angular 16 that I might be missing? Thank you.

clydin:
Custom postcss configuration is only supported with the `application` build system in v17.2 or higher.

JoepKockelkorn:
I can confirm the `postcss.config.json` approach works. I did have to install tailwind and its postcss plugin using `--legacy-peer-deps` due to conflicts.

Char2sGu:
> Custom postcss configurations are possible with the `application` build system (default for all new projects).

Should this be documented? I would never know about this if I didn't accidentally clicked into this issue. @clydin

sheikalthaf:
![Image](https://github.com/user-attachments/assets/11982f69-10e9-4fc7-b5a8-8fd1be1f30f6)

Tailwind v4 is getting release next week and we need to support it from angular is good.

kyjus25:
[...]

PR #28939, opened 2024-11-22
feat: support tailwind v4

heddendorp:
## PR Checklist

- [x ] The commit message follows our guidelines: https://github.com/angular/angular-cli/blob/main/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes
- [ ] CI related changes

The current integration if for tailwind V3 only.

Issue Number: #28938

## What is the new behavior?

Detect if tailwind v4 is used as indicated by the presence of the `@tailwindcss/postcss` package. If so, check that also `tailwindcss` is installed. If both are present, Add the tailwind postCSS plugin.

- [ ] Yes
- [x ] No

### DIFF
--- a/packages/angular_devkit/build_angular/src/tools/webpack/configs/styles.ts
+++ b/packages/angular_devkit/build_angular/src/tools/webpack/configs/styles.ts
@@ -96,6 +96,23 @@ export async function getStylesConfig(wco: WebpackConfigOptions): Promise<Config
       extraPostcssPlugins.push(require(tailwindPackagePath)({ config: tailwindConfigPath }));
     }
   }
+  // Check for tailwind V4
+  // Tailwind V4 uses the new `@tailwindcss/postcss` package
+  try {
+    const tailwindPostCSSPackagePath = require.resolve('@tailwindcss/postcss', { paths: [root] });
+    try {
+      require.resolve('tailwindcss', { paths: [root] });
+      // Tailwind V4 is installed, add the plugin
+      extraPostcssPlugins.push(require(tailwindPostCSSPackagePath)());
+    } catch {
+      logger.warn(
+        `The '@tailwindcss/postcss' package is installed but the 'tailwindcss' package is not installed.` +
+          ` To enable Tailwind CSS, please install the 'tailwindcss' package.`,
+      );
+    }
+  } catch {
+    // Tailwind V4 is not installed, no further action is needed
+  }
 
   const autoprefixer: typeof import('autoprefixer') = require('autoprefixer');
 


packages/angular_devkit/build_angular/src/tools/webpack/configs/styles.ts
```diff
@@ -96,6 +96,23 @@ export async function getStylesConfig(wco: WebpackConfigOptions): Promise<Config
       extraPostcssPlugins.push(require(tailwindPackagePath)({ config: tailwindConfigPath }));
     }
   }
+  // Check for tailwind V4
```
jkrems, line 99: Should this be in the previous `if` block to prevent accidentally enabling tailwind because of transitive deps?

heddendorp: Tailwind 4 doesn't have a co fig file anymore
It's all in css now

pumano:
would be good to have same support for esbuild builder toom not only for webpack

heddendorp:
> would be good to have same support for esbuild builder too, not only for webpack

Agreed, I found what seems to be the correct place here https://github.com/angular/angular-cli/pull/28939/files#diff-c31d390b8683db112fcc30e2ee984d893bdce6bd8e6d7d70c43ecc410a626a3dR489
But I am unsure how to add it correctly for the application builder.

sheikalthaf:
image

@clydin  tailwind v4 is getting released next week, it is better have support from day 1

DominicWrege:
Can't wait to get this feature

Kludgy4:
The largest change in Tailwind v4 is the extra spe e  e   d from their new build system, which we can't fully appreciate with postcss mandated. I'd say it's essential this support allows us to avoid postcss if desired

Eirmas:
+1 bump

krasi:
Bump! It's not possible to use Angular with Storybook and Tailwindcss v4 until we get this merged.

this part is the crucial one, if anyone else struggles with the same:
```
const tailwindPostCSSPackagePath = require.resolve('@tailwindcss/postcss', { paths: [root] });
```

utpaulBS23:
Which angular version will we get that tailwindcss 4 supports @jkrems? Waiting for your reply

hybrist:
Sorry for the delay here! Few things as frustrating as having a PR linger without a clear signal. There were some misunderstandings around the details of the situation which led to a bit of a bystander effect. To sum things up:

* Tailwind v4 works with the application builder in both v19 and v20, using a custom postcss JSON config file.
* Storybook currently requires the use of the _browser_ builder (webpack) for Angular components.
* We have landed https://github.com/angular/angular-cli/pull/30592 which backports support for plugins in a postcss config file to the browser builder. This should go out with the next minor version of v20, so 20.1.0.

At that point, the documented setup for the application builder will also work when using the browser builder. This should hopefully reduce the amount of conflicting instructions across the builders.

It's worth noting that in any case, this is all based on the postcss plugin for tailwind. If there's interest in an integration with Tailwind's lightning css flow (especially if it comes with concrete estimates or even data on the impact on an Angular app), that might be worth as a separate issue.

I'm closing this PR since I believe https://github.com/angular/angular-cli/pull/30592 ensures that apps can make use of tailwind v4 even when they need the browser builder.

PR #28961, opened 2024-11-25
refactor(@angular/build): maintain previous compiler plugin factory signature

clydin:
To minimize downstream breakage for consumers of the private API, the `createCompilerPlugin` function signature has been adjusted to maintain compatibility with v18.


alan-agius4 (approved):

clydin:

Issue #28985, opened 2024-11-27
bug(*): Issue with `ng add @angular/material` in Angular CLI


samir-dahal:
### Is this a regression?

- [x] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

17.3.5

### Description

Dear Angular Team,

I encountered an issue while using the Angular CLI to integrate `@angular/material` into a new project. Here are the steps I followed and the problem I observed:

1. **Environment**:
   - Angular project created using the Angular CLI.
   - Angular CLI version: [18.2.12].
   - Node.js version: [22.11.0].

2. **Issue**:
   - I ran the command `ng add @angular/material` in the terminal.
   - The installation was marked as successful, but the necessary project files (e.g., `angular.json`, `app.config.ts, styles.css`, etc) were not modified.
   - Additionally, the following error was displayed:
     ```
     Cannot find module '\node_modules\@angular\cli\node_modules\@schematics\angular\private\components.js'
     ```
I also checked the node_modules folder but didn't find any folder named "private"

### Reproduction

StackBlitz link: Sorry, I am working on my local machine, so cannot provide the link here.
Steps to reproduce:
1. Create a new project using angular CLI
2. Add the @angular/material using `ng add @angular/material`

### Expected Behavior

Should properly install the @angular/material package.

### Actual Behavior

Error after installing the package.
```
     Cannot find module '\node_modules\@angular\cli\node_modules\@schematics\angular\private\components.js'
```

### Environment

- Angular: 18.2.12
- CDK/Material: 19.0.0
- Browser(s): Microsoft Edge 131.0.2903.63 (Official build) (64-bit) 
- Operating System (e.g. Windows, macOS, Ubuntu):

alan-agius4:
The problem appears that you have an outdated version of `@angular/material` installed.

json-derulo:
@alan-agius4 I think you misunderstood the issue. When creating a fresh Angular 18 non-standalone workspace **without Material**, after running `ng add @angular/material` the wrong version of Angular Material is chosen, which probably leads to the described issue.

alan-agius4:
Sorry I misunderstood the issue.

deerbuster:
this was not fixed.  I have the issue today.  I force cleared my NPM cache, and create a fresh angular 18.2.12 app, ng add @angular/material, and it chose material v17 again. "found compatible package version: @angular/material@17.0.0"

JeanMeche:
[...]

PR #28986, opened 2024-11-27
fix(@angular/cli): correctly select package versions in descending order during `ng add`

alan-agius4:
When using the `ng add` command, the package version selection logic was not correctly selected based on the available versions in desc order. This could lead to selecting an unintended version of the package.

Closes: #28985


clydin (approved):

alan-agius4:

PR #28992, opened 2024-11-28
build: update to TypeScript 5.7

crisbeto:
Updates the repo to use TypeScript 5.7.


packages/schematics/angular/utility/latest-versions/package.json
```diff
@@ -17,13 +17,13 @@
     "karma-jasmine": "~5.1.0",
     "karma": "~6.4.0",
     "less": "^4.2.0",
-    "ng-packagr": "^19.0.0",
+    "ng-packagr": "^19.1.0-next.1",
```
alan-agius4, line 20: This should be `.0` otherwise the prerelease check will fail.

```suggestion
    "ng-packagr": "^19.1.0-next.0",
```

alan-agius4 (changes requested):
These peer dependencies are not valid. 

The string value need to be an **exact** match to `^19.0.0 || ^19.1.0-next.0`, any other value is not valid and will cause the release checks to fail.

I have a PR to bump the packages https://github.com/angular/angular-cli/pull/29007

crisbeto:
This won't allow any other 19.0.x versions though (e.g. 19.0.1). I think the right thing to do is to require only 19.1.x in the next minor.

alan-agius4:
Sorry I was meant to be `^19.0.0 || ^19.1.0-next.0`

packages/angular_devkit/build_angular/test/hello-world-lib/projects/lib/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "lib",
```
alan-agius4, line 2: Kindly revert this, this is a test file.

packages/schematics/angular/utility/latest-versions/package.json
```diff
@@ -17,13 +17,13 @@
     "karma-jasmine": "~5.1.0",
     "karma": "~6.4.0",
     "less": "^4.2.0",
-    "ng-packagr": "^19.1.0-next.0",
+    "ng-packagr": "^19.1.0-next.2",
```
alan-agius4, line 20: This needs to be `.0`. 

All framework and ng-packagr versions must be `^19.1.0-next.0` during the prerelease phase any other value is not allowed. 

```suggestion
    "ng-packagr": "^19.1.0-next.0",
```

alan-agius4 (commented):
Looks like there are a couple of places where typescript has not been updated.

```
packages/angular_devkit/schematics_cli/blank/project-files/package.json
packages/angular_devkit/schematics_cli/schematic/files/package.json
```

crisbeto:
Should be ready to go now.

alan-agius4 (approved):

PR #29011, opened 2024-12-02
feat(@angular/build): add support for customizing URL segments with i18n

### LINKED ISSUE #16997: Allow custom "outputPath" directory for each locale
# 🚀 Feature request

- [x] build

### Description
New version of Angular CLI allows us to use the new --localize option to generate different locales. This is great as it significantly reduces the build time for apps with many languages. However, as far as I can tell, the name of the output directory for each language is set to the name of the locale.
So if I have this locale defined:

`        "locales": {
          "foo-bar": {
            "translation": "src/i18n/messages.en-US.xlf"
          }
        }`

the output directory for this locale would be `/foo-bar`.

I would like to be able to specify outputPath per locale, ideally by just defining "outputPath" in each locale which would then be appended to existing build options outputPath.

Alternative is to just change name of the locale to the one I want. But this then creates a problem where LOCALE_ID is also changed. And I want my LOCALE_ID to be, for example en-US and my output directory of build to be /en.

### LINKED ISSUE #28967: Set locale baseHref to '/' is not working with AngularNodeAppEngine
### Command

build

### Is this a regression?

_No response_

### Description

I have an application where English is the default locale and I don't want the prefix '/en' in the URL. With CommonEngine it was possible to set the baseHref in angular.json to '/' to achieve this.

```json
      "i18n": {
        "sourceLocale": {
          "code": "en",
          "baseHref": "/"
        },
        "locales": {
          "de": "src/messages/messages.de.xlf",
          "es": "src/messages/messages.es.xlf"
        }
      }
```

Now I have migrated this application to Angular 19 and noticed that this does not seem to be possible anymore with AngularNodeAppEngine as the default locale 'en' is not recognized. The build is successful and all other locales work fine, except 'en'. Even the root path works, because the potentialLocale is '' (empty string) in this case, but no child routes can be found (at least not with ssr). Is this behavior expected? If so, are there still any workarounds to disable the locale prefix for a specific locale?

In a (maybe) naive approach, I tried adding the following line just before returning the entryPointExport in [app-engine.ts#L146](https://github.com/angular/angular-cli/blame/7b0aa3dec56298bb364516ac189aa32accc7dcc2/packages/angular/ssr/src/app-engine.ts#L146) and it worked, so it seems to be supported everywhere except in this particular location and therefore I really hope it's just a bug and I don't have to revert the migration.
```javascript
if (!entryPoints.has(potentialLocale) && entryPoints.has("")) {
    return this.getEntryPointExports("");
}
```

Any help is appreciated, because everything else is already migrated.

### Minimal Reproduction

Repo: https://github.com/eufri/angular-ssr-issue

Basically I have created a new app
`ng new angular-ssr-issue --ssr --minimal --package-manager=yarn --routing --server-routing --skip-tests --ssr`

Added localize package
`ng add @angular/localize`

And then added some test locales + messages and a child component +route

### Exception or Error

```text

```

### Your Environment

```text
Angular: 19.0.0
... localize, platform-browser, platform-browser-dynamic
... platform-server, router

@angular-devkit/architect       0.1900.1
@angular-devkit/build-angular   19.0.1
@angular-devkit/core            19.0.1
@angular-devkit/schematics      19.0.1
@angular/cli                    19.0.1
@angular/ssr                    19.0.1
@schematics/angular             19.0.1
```

### Anything else relevant?

_No response_

alan-agius4:
Previously, the `baseHref` option under each locale allowed for generating a unique base href for specific locales. However, users were still required to handle file organization manually, and `baseHref` appeared to be primarily designed for this purpose.

This commit introduces a new `subPath` option, which simplifies the i18n process, particularly in static site generation (SSG) and server-side rendering (SSR). When the `subPath` option is used, the `baseHref` is ignored. Instead, the `subPath` serves as both the base href and the name of the directory containing the localized version of the app.

Below is an example configuration showcasing the use of `subPath`:

```json
"i18n": {
  "sourceLocale": {
    "code": "en-US",
    "subPath": ""
  },
  "locales": {
    "fr-BE": {
      "subPath": "fr",
      "translation": "src/i18n/messages.fr-BE.xlf"
    },
    "de-BE": {
      "subPath": "de",
      "translation": "src/i18n/messages.de-BE.xlf"
    }
  }
}
```

The following tree structure demonstrates how the `subPath` organizes localized build output:
```
dist/
├── app/
│   └── browser/  # Default locale, accessible at `/`
│       ├── fr/  # Locale for `fr-BE`, accessible at `/fr`
│       └── de/  # Locale for `de-BE`, accessible at `/de`
```

DEPRECATED: The `baseHref` option under `i18n.locales` and `i18n.sourceLocale` in `angular.json` is deprecated in favor of `subPath`.

The `subPath` defines the URL segment for the locale, serving as both the HTML base HREF and the directory name for output. By default, if not specified, `subPath` will use the locale code.

Closes #16997 and closes #28967

### DIFF
--- a/packages/angular/build/src/utils/i18n-options.ts
+++ b/packages/angular/build/src/utils/i18n-options.ts
@@ -18,6 +18,7 @@ export interface LocaleDescription {
   translation?: Record<string, unknown>;
   dataPath?: string;
   baseHref?: string;
+  subPath: string;
 }
 
 export interface I18nOptions {
@@ -54,19 +55,31 @@ function normalizeTranslationFileOption(
 
 function ensureObject(value: unknown, name: string): asserts value is Record<string, unknown> {
   if (!value || typeof value !== 'object' || Array.isArray(value)) {
-    throw new Error(`Project ${name} field is malformed. Expected an object.`);
+    throw new Error(`Project field '${name}' is malformed. Expected an object.`);
   }
 }
 
 function ensureString(value: unknown, name: string): asserts value is string {
   if (typeof value !== 'string') {
-    throw new Error(`Project ${name} field is malformed. Expected a string.`);
+    throw new Error(`Project field '${name}' is malformed. Expected a string.`);
   }
 }
 
+function ensureValidsubPath(value: unknown, name: string): asserts value is string {
+  ensureString(value, name);
+
+  if (!/^[\w-]*$/.test(value)) {
+    throw new Error(
+      `Project field '${name}' is invalid. It can only contain letters, numbers, hyphens, and underscores.`,
+    );
+  }
+}
 export function createI18nOptions(
   projectMetadata: { i18n?: unknown },
   inline?: boolean | string[],
+  logger?: {
+    warn(message: string): void;
+  },
 ): I18nOptions {
   const { i18n: metadata = {} } = projectMetadata;
 
@@ -82,21 +95,40 @@ export function createI18nOptions(
     },
   };
 
-  let rawSourceLocale;
-  let rawSourceLocaleBaseHref;
+  let rawSourceLocale: string | undefined;
+  let rawSourceLocaleBaseHref: string | undefined;
+  let rawsubPath: string | undefined;
   if (typeof metadata.sourceLocale === 'string') {
     rawSourceLocale = metadata.sourceLocale;
   } else if (metadata.sourceLocale !== undefined) {
-    ensureObject(metadata.sourceLocale, 'i18n sourceLocale');
+    ensureObject(metadata.sourceLocale, 'i18n.sourceLocale');
 
     if (metadata.sourceLocale.code !== undefined) {
-      ensureString(metadata.sourceLocale.code, 'i18n sourceLocale code');
+      ensureString(metadata.sourceLocale.code, 'i18n.sourceLocale.code');
       rawSourceLocale = metadata.sourceLocale.code;
     }
 
     if (metadata.sourceLocale.baseHref !== undefined) {
-      ensureString(metadata.sourceLocale.baseHref, 'i18n sourceLocale baseHref');
+      ensureString(metadata.sourceLocale.baseHref, 'i18n.sourceLocale.baseHref');
+      logger?.warn(
+        `The 'baseHref' field under 'i18n.sourceLocale' is deprecated and will be removed in future versions. ` +
+          `Please use 'subPath' instead.\nNote: 'subPath' defines the URL segment for the locale, acting ` +
+          `as both the HTML base HREF and the directory name for output.\nBy default, ` +
+          `if not specified, 'subPath' uses the locale code.`,
+      );
+
       rawSourceLocaleBaseHref = metadata.sourceLocale.baseHref;
+    }
+
+    if (metadata.sourceLocale.subPath !== undefined) {
+      ensureValidsubPath(metadata.sourceLocale.subPath, 'i18n.sourceLocale.subPath');
+      rawsubPath = metadata.sourceLocale.subPath;
+    }
+
+    if (rawsubPath !== undefined && rawSourceLocaleBaseHref !== undefined) {
+      throw new Error(
+        `'i18n.sourceLocale.subPath' and 'i18n.sourceLocale.baseHref' cannot be used together.`,
+      );
     }
   }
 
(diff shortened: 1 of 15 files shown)


packages/angular/build/src/utils/i18n-options.ts
```diff
@@ -18,6 +18,7 @@ export interface LocaleDescription {
   translation?: Record<string, unknown>;
   dataPath?: string;
   baseHref?: string;
```
alan-agius4, line 20: Should be deprecate `baseHref` from the application builder?

dgp1130: +1 to deprecating.

packages/angular/build/src/utils/i18n-options.ts
```diff
@@ -64,6 +65,16 @@ function ensureString(value: unknown, name: string): asserts value is string {
   }
 }
 
+function ensureValidateUrlSegment(value: unknown, name: string): asserts value is string {
```
dgp1130, line 68: Style: `ensureValidate...` feels a bit redundant. Should this be `ensureValidUrlSegment` or `validateUrlSegment`?

packages/angular/build/src/utils/i18n-options.ts
```diff
@@ -64,6 +65,16 @@ function ensureString(value: unknown, name: string): asserts value is string {
...
+  ensureString(value, name);
+
+  if (!/^[\w-]*$/.test(value)) {
+    throw new Error(
+      `Project ${name} field is malformed. Expected to match pattern: '/^[\\w-]*$/'.`,
```
dgp1130, line 73: Suggestion: Can we describe this verbally rather than just referencing a regex? Newer developers aren't necessarily going to be familiar with regex or understand what pattern this is trying to express.

packages/angular/build/src/utils/i18n-options.ts
```diff
@@ -108,21 +129,35 @@ export function createI18nOptions(
...
         }
+
+        if ('urlSegment' in options) {
+          ensureString(options.urlSegment, `i18n locales ${locale} urlSegment`);
+          urlSegment = options.urlSegment;
+        }
+
+        if (urlSegment !== undefined && baseHref !== undefined) {
+          throw new Error(
+            `i18n locales ${locale} urlSegment and baseHref cannot be used togather.`,
```
dgp1130, line 158: Typo: "togather"

packages/angular/build/src/utils/i18n-options.ts
dgp1130, line 73: Question: `name` is "i18n sourceLocale urlSegment" below right? Isn't that an "option", not a "project"?

packages/angular/build/src/utils/i18n-options.ts
```diff
@@ -98,6 +110,15 @@ export function createI18nOptions(
       ensureString(metadata.sourceLocale.baseHref, 'i18n sourceLocale baseHref');
       rawSourceLocaleBaseHref = metadata.sourceLocale.baseHref;
     }
+
+    if (metadata.sourceLocale.urlSegment !== undefined) {
+      ensureValidateUrlSegment(metadata.sourceLocale.urlSegment, 'i18n sourceLocale urlSegment');
```
dgp1130, line 115: Consider: "i18n sourceLocale urlSegment" is an `angular.json` option right? Would `i18n.sourceLocale.urlSegment` be more clear about the context?

It seems like this format is pre-existing in this file, so maybe we already have a convention here, I'm just not sure I would understand this formatting.

packages/angular/build/src/utils/i18n-options.ts
```diff
@@ -98,6 +110,15 @@ export function createI18nOptions(
...
+      rawUrlSegment = metadata.sourceLocale.urlSegment;
+    }
+
+    if (rawUrlSegment !== undefined && rawSourceLocaleBaseHref !== undefined) {
+      throw new Error(`i18n sourceLocale urlSegment and baseHref cannot be used togather.`);
```
dgp1130, line 120: Typo: "togather"

dgp1130 (approved):
This looks great! Mostly just some small suggestions to make sure the error cases are clear and understandable to users.

My one thought is whether there is a more clear name than `urlSegment` we could use? In my mind that really doesn't describe what this will do and without any further context I'd have no idea what to use it for.

"segment" to me also implies it is exactly one segment between slashes, but I'm not sure that's a hard constraint? Users can write `urlSegment: 'foo/bar'` right? That would look like two segments to me.

Not really sure of an alternative, but we could consider something like `localeBase`, `localeRoot`, `root` (maybe we can rely on the context that this option is only available with a locale), `directory`, `outputDir`, `subpath`, `localeBaseHref` / `baseHref` (it does say this serves as "baseHref" in the JSON doc), ...

I'll defer to you on the exact naming, I just want to make sure we're considering alternatives and thinking critically about what this concept represents to users.

packages/angular/build/src/utils/i18n-options.ts
alan-agius4: Improved all the error messages

packages/angular/build/src/utils/i18n-options.ts
alan-agius4: Updates

packages/angular/build/src/utils/i18n-options.ts
```diff
@@ -82,22 +95,41 @@ export function createI18nOptions(
...
+          `Please use 'subPath' instead.\nNote: 'subPath' defines the URL segment for the locale, acting ` +
+          `as both the HTML base HREF and the directory name for output.\nBy default, ` +
+          `if not specified, 'subPath' uses the locale code.`,
+      );
+
       rawSourceLocaleBaseHref = metadata.sourceLocale.baseHref;
     }
+
+    if (metadata.sourceLocale.subPath !== undefined) {
+      ensureValidsubPath(metadata.sourceLocale.subPath, 'i18n.sourceLocale.subPath');
```
clydin, line 124: We should also validate that each sub path is unique.  Otherwise the files will overwrite each other in the output. Maybe track used subPath to locale and then generate an error based on that.

clydin: Actually probably need to do that after this since a configured sub path could collide with a default value as well.

alan-agius4: Yes good idea

clydin (approved):
One option validation comment, otherwise LGTM.

alan-agius4:
19.1, which is currently in prerelease.

hheexx:
Hi @alan-agius4 ,

looks like this feature was not present in builds before RC.
Now it's inside so I took a look.

I added to angular.json like this:
image


Slashes made a problem and generated manifest was:
export default {
  basePath: '/',
  supportedLocales: {
  "en": "en",
  "sr": "",
  "ru": "/ru/"
},
  entryPoints: {
    'en': () => import('./en/main.server.mjs'),
    '': () => import('./main.server.mjs'),
    '/ru/': () => import('.//ru//main.server.mjs')
  },
};



Russian was not working.
After removing the slashes everything is OK but maybe you want to check why is there different handling between en and ru.

alan-agius4:
@hheexx, it looks like the validation function was not being called correctly. https://github.com/angular/angular-cli/pull/29328

PR #29017, opened 2024-12-02
refactor(@angular/build): add initial component HMR source file analysis

clydin:
When component template HMR support is enabled (`NG_HMR_TEMPLATES=1`), TypeScript file changes will now be analyzed to determine if Angular component metadata has changed and if the changes can support a hot replacement. Any other changes to a TypeScript file will cause a full page reload to avoid inconsistent state between the code and running application. The analysis currently has an upper limit of 32 modified files at one time to prevent a large of amount of analysis to be performed which may take longer than a full rebuild. This value may be adjusted based on feedback. Component template HMR is currently experimental and may not support all template modifications. Both inline and file-based templates are now supported. However, rebuild times have not yet been optimized.


packages/angular/build/src/tools/angular/compilation/hmr-candidates.ts
```diff
@@ -0,0 +1,316 @@
...
+  firstSource: ts.SourceFile,
+  secondRange: ts.ReadonlyTextRange | undefined,
+  secondSource: ts.SourceFile,
+): boolean {
+  // Check matching undefined values
+  if (firstRange === undefined) {
+    return secondRange === undefined;
+  } else if (secondRange === undefined) {
+    return firstRange === undefined;
+  }
```
alan-agius4, line 297:
```suggestion
 // Check matching undefined values
  if (!firstRange || !secondRange) {
    return firstRange === secondRange;
  }
```

alan-agius4 (approved):

Issue #29012, opened 2024-12-02
Development and Production differs on entry '/'


alexandresebrao:
### Command

serve

### Is this a regression?

- [x] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

19.0.0

### Description

On development mode I'm not having any problem when using AppNodeEngine

But when in production when accessing the root is not working the same way.

On my test I'm having a problem with angularNodeAppEngine

### Minimal Reproduction

The example uses fastify

```
    server.get('*', async (req, reply) => {
       const user = {username: 'test'}
        try {
            const response = await angularNodeAppEngine.handle(req.raw, {
                user
            })
            if (response) {
                console.log(" i'm here ")
                await writeResponseToNodeResponse(response, reply.raw)
            } else {
                reply.callNotFound()
            }
        } catch (error) {
            reply.send(error)
        }
    })
```

If i hit any route like
/profile

The console will say 'im here'

And everything will work fine.

But with /

It will not

With this my canActivate wont receive the REQUEST_CONTEXT of the user and will redirect him to the login page, instead of the homepage
app.server.route.ts

```import { RenderMode, ServerRoute } from '@angular/ssr'

export const serverRoutes: ServerRoute[] = [
    {
        path: '**',
        renderMode: RenderMode.Server
    }
]```


### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 19.0.2
Node: 22.11.0
Package Manager: pnpm 9.13.2
OS: linux x64

Angular: 19.0.1
... animations, cdk, common, compiler, compiler-cli, core, forms
... localize, material, material-luxon-adapter, platform-browser
... platform-browser-dynamic, platform-server, router
... service-worker

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1900.2
@angular-devkit/core         19.0.1
@angular-devkit/schematics   19.0.1
@angular/build               19.0.2
@angular/cli                 19.0.2
@angular/ssr                 19.0.2
@schematics/angular          19.0.1
rxjs                         7.8.1
typescript                   5.6.3
zone.js                      0.15.0
```

### Anything else relevant?

I didn't make the mini repo, but if its really needed I will send as soon as possible, any other info I will be pretty happy to share.

alan-agius4:
Hi @alexandresebrao, I tried this locally and I was not able to replicate the mentioned behaviour. Can you setup a minimal repro please?

[...]

PR #29024, opened 2024-12-03
fix(@angular/build): ensure correct handling of `index.output` for SSR

alan-agius4:
Previously, the index file was not being renamed correctly when using server-side rendering (SSR).

Closes: #29012


clydin (approved):

alan-agius4:

PR #29027, opened 2024-12-03
refactor(@angular/ssr): replace `Map` with `Record` in SSR manifest

alan-agius4:
Replaced `Map` with `Record` in SSR manifest to simplify structure and improve testing/setup.

### DIFF
--- a/packages/angular/build/src/utils/server-rendering/manifest.ts
+++ b/packages/angular/build/src/utils/server-rendering/manifest.ts
@@ -55,7 +55,7 @@ export function generateAngularServerAppEngineManifest(
   i18nOptions: NormalizedApplicationBuildOptions['i18nOptions'],
   baseHref: string | undefined,
 ): string {
-  const entryPointsContent: string[] = [];
+  const entryPoints: Record<string, string> = {};
 
   if (i18nOptions.shouldInline) {
     for (const locale of i18nOptions.inlineLocales) {
@@ -69,18 +69,22 @@ export function generateAngularServerAppEngineManifest(
       const end = localeWithBaseHref[localeWithBaseHref.length - 1] === '/' ? -1 : undefined;
       localeWithBaseHref = localeWithBaseHref.slice(start, end);
 
-      entryPointsContent.push(`['${localeWithBaseHref}', () => import('${importPath}')]`);
+      entryPoints[localeWithBaseHref] = `() => import('${importPath}')`;
     }
   } else {
-    entryPointsContent.push(`['', () => import('./${MAIN_SERVER_OUTPUT_FILENAME}')]`);
+    entryPoints[''] = `() => import('./${MAIN_SERVER_OUTPUT_FILENAME}')`;
   }
 
   const manifestContent = `
 export default {
   basePath: '${baseHref ?? '/'}',
-  entryPoints: new Map([${entryPointsContent.join(', \n')}]),
+  entryPoints: {
+    ${Object.entries(entryPoints)
+      .map(([key, value]) => `'${key}': ${value}`)
+      .join(',\n    ')}
+  },
 };
-  `;
+`;
 
   return manifestContent;
 }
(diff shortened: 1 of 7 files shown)


packages/angular/ssr/src/manifest.ts
```diff
@@ -123,8 +147,11 @@ let angularAppManifest: AngularAppManifest | undefined;
  *
  * @param manifest - The manifest object to set for the Angular application.
  */
-export function setAngularAppManifest(manifest: AngularAppManifest): void {
-  angularAppManifest = manifest;
+export function setAngularAppManifest(manifest: AngularAppManifestConfig): void {
+  angularAppManifest = {
+    ...manifest,
+    assets: new Map(Object.entries(manifest.assets)),
```
alan-agius4, line 153: @dgp1130, I’m wondering if using a `Map` here is excessive, and if we should just stick with a plain `Record`. I understand the decision to avoid prototype pollution, but it feels like an overly cautious approach.

We are only using the this to read data, with no need for iteration or modification. Given this, I believe a `Record` is a better choice, since there's no need for the features that a `Map` offers, such as key ordering or methods like `set` or `delete`.

alfaproject: `Object.create(null)` creates a object without prototype, if that's the only concern, right?

packages/angular/build/src/utils/server-rendering/manifest.ts
```diff
@@ -69,18 +69,22 @@ export function generateAngularServerAppEngineManifest(
...
+    entryPoints[''] = `() => import('./${MAIN_SERVER_OUTPUT_FILENAME}')`;
   }
 
   const manifestContent = `
 export default {
   basePath: '${baseHref ?? '/'}',
-  entryPoints: new Map([${entryPointsContent.join(', \n')}]),
+  entryPoints: {
+    ${Object.entries(entryPoints)
+      .map(([key, value]) => `'${key}': ${value}`)
```
dgp1130, line 83: Nit: I probably should have noticed this the first time, but do we need to escape `key` if it contains a single quote? I guess it would be pretty rare for an entry point to contain a single quote in the file name. Are we comfortable relying on that assumption?

packages/angular/ssr/src/manifest.ts
```diff
@@ -10,7 +10,20 @@ import type { SerializableRouteTreeNode } from './routes/route-tree';
...
+ */
+type TransformRecordWithReadonlyMap<T> = {
+  [K in keyof T]: T[K] extends (...args: unknown[]) => unknown // Check if it's a function
+    ? T[K] // Leave function properties untouched
+    : T[K] extends Readonly<Record<string, unknown>> // Check if it's a Readonly Record
+      ? T[K] extends Readonly<Record<infer RK, infer RV>>
+        ? ReadonlyMap<RK, RV> // Convert Readonly Record to ReadonlyMap
+        : never
+      : T[K]; // Leave other properties unchanged
+};
```
dgp1130, line 23:
```suggestion
type TransformRecordWithReadonlyMap<T> = {
  [K in keyof T]: T[K] extends (...args: unknown[]) => unknown // Check if it's a function
    ? T[K] // Leave function properties untouched
    : T[K] extends Readonly<Record<infer RK, infer RV>>
        ? ReadonlyMap<RK, RV> // Convert Readonly Record to ReadonlyMap
        : T[K]; // Leave other properties unchanged
};
```

Suggestion: Might be moot if we drop the `Map`, but can we get rid of the double `extends Readonly<...>` check? I think we can `infer` at the same time?

packages/angular/ssr/src/manifest.ts
dgp1130: I'm ok with a `Record`. My suggestion is more to prefer `Map` where possible, but when there are compelling reasons to keep the `Record` (like inlining in a manifest), I think it's perfectly reasonable to consider.

dgp1130 (approved):

packages/angular/build/src/utils/server-rendering/manifest.ts
alan-agius4: In this case the is the locale, thus cannot contain non valid letters.

packages/angular/ssr/src/manifest.ts
alan-agius4: I tried this, but it does not work as expected

alan-agius4:

Issue #29029, opened 2024-12-03
deployUrl is added to absolute paths in link rel=preload tags


ihsnow:
### Command

build

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

deployUrl is added to absolute paths in link rel=preload tags

### Minimal Reproduction

Please check:

- angular 17: https://stackblitz.com/github/ihsnow/deploy-example-17/
- angular 19: https://stackblitz.com/github/ihsnow/deploy-example-19/

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 17.3.11
Node: 20.15.0
Package Manager: npm 10.8.1
OS: darwin arm64

Angular: 17.3.12
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1703.11
@angular-devkit/build-angular   17.3.11
@angular-devkit/core            17.3.11
@angular-devkit/schematics      17.3.11
@angular/cli                    17.3.11
@schematics/angular             17.3.11
rxjs                            7.8.1
typescript                      5.4.5
zone.js                         0.14.10
```

PR #29032, opened 2024-12-03
fix(@angular/build): avoid deploy URL usage on absolute preload links

clydin:
The `deployUrl` option was unintentionally being prepended to preload links with absolute URLs within the generated index HTML for an appplication. This has now been corrected and absolute URLs will not be altered when a deploy URL is configured.

Closes: #29029


alan-agius4 (approved):

alan-agius4:

Issue #29054, opened 2024-12-04
Auto-CSP generates invalid loader scripts for some <script> tags


aaronshim:
### Command

build

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

Auto-CSP relies on rewriting scripts of the form `<script src="..."></script>` into an inline loader script that will dynamically re-create these tags and add them to the DOM.

There are some `<script src="...">` configurations that generate inline loader scripts that will not run on the transformed page, such as

- Having your `<script src="...">` tag appear in the `<head>` rather than the `<body>` of the document.
- Having your `<script src="...">` tag interpreted as something that will create a `type="undefined"` attribute on the generated node added to the DOM

### Minimal Reproduction

The `index.html` should have the following:

```
<!doctype html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <base href="/">
  <script src='script1.js'></script>
</head>
<body>
  <app-root></app-root>  
  <script src='script2.js'></script>
  </body>
</html>
```

### Exception or Error

```text
`script1.js` and `script2.js` do not run.
```

### Your Environment

```text
_                      _                 ____ _     ___
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 19.0.3
Node: 20.15.1
Package Manager: npm 10.7.0
OS: linux x64

Angular: 
... 

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1900.3 (cli-only)
@angular-devkit/core         19.0.3 (cli-only)
@angular-devkit/schematics   19.0.3 (cli-only)
@schematics/angular          19.0.3 (cli-only)
```

aaronshim:
Proposing #29055 to fix this issue.

alan-agius4:
Closed via https://github.com/angular/angular-cli/pull/29055

PR #29055, opened 2024-12-04
fix(@angular/build): Fixing auto-csp edge cases

aaronshim:
- <script> is the last tag before </head> close
- .appendChild is called before </head> (because document.body is undefined then)
- <script> tags with a src attribute and no specified type attribute should not write <script type="undefined" ...>

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Auto-CSP creates non-loading loader scripts

Issue Number: #29054 

## What is the new behavior?

Auto-CSP generates loading loader scripts

- [ ] Yes
- [X] No


packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -288,7 +301,7 @@ function createLoaderScript(srcList: SrcScriptTag[], enableTrustedTypes = false)
     s.type = scriptUrl[1];
     s.async = !!scriptUrl[2];
     s.defer = !!scriptUrl[3];
-    document.body.appendChild(s);
+    document.${inBody ? 'body' : 'head'}.appendChild(s);
```
dgp1130, line 304: Consider: What if we updated this to support either case like:

```typescript
(document.body ?? document.head).appendChild(s);
```

This way we don't need to track which element we're in? That's probably a slight bundle size regression, but I feel like this should be insignificant enough to not matter? There's probably bigger opportunities here if we wanted to optimize anyways.

For that matter, does it make difference to insert in the head or the body as long as that element exists? Could we just always insert tags in to `head`, even if they originated from the body? Would that cause problems?

packages/angular/build/src/utils/index-file/auto-csp_spec.ts
```diff
@@ -160,14 +166,40 @@ describe('auto-csp', () => {
...
     );
     // Exactly 4 scripts should be left.
     expect(Array.from(result.matchAll(/<script>/g)).length).toEqual(4);
   });
+
+  it('should write a loader script that appends to head', async () => {
+    const result = await autoCsp(`
+      <html>
+        <head>
+        <script src="./head.js"></script>
```
dgp1130, line 184: Nit: Indenting.

packages/angular/build/src/utils/index-file/auto-csp.ts
aaronshim:
```
That's probably a slight bundle size regression, but I feel like this should be insignificant enough to not matter?
```

The nit is that the extra characters for ` ?? document.head)` will be included in every loader script, which can scale to the number of `<script src="..."></script>` nodes in cases like:

```
<script src="..."></script>
<script>console.log("hi");</script>
<script src="..."></script>
<script>console.log("hi");</script>
<script src="..."></script>
...
```

since there is one loader script per contiguous block of `<script>` tags. Depending on how many `<script src="..."></script>` tags there are, maybe it isn't a big deal.

```
For that matter, does it make difference to insert in the head or the body as long as that element exists? Could we just always insert tags in to head, even if they originated from the body?
```

For this question, I think (as we found out with the execution order for nodes added through `document.appendChild(...)`) the execution order depends on the time at which the node was inserted into the document rather than where it was inserted into the document-- but one advantage of correctly appending to the appropriate `head` or `body` is that when the Chrome Dev Tools is used to look at the final markup, the loader scripts insert `<script src=...>` tags into where they *would've* been in the DOM, which be a small benefit aesthetically. Do you think the added complexity of adding the scripts to where they would've been in the DOM is worth it for this aesthetic benefit? Thanks!

aaronshim: As discussed, we will stop tracking `inBody` and instead go with `document.lastElementChild` to make it even more exact where the `<script>` tag is re-inserted into the DOM.

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -258,7 +267,11 @@ function getStrictCsp(
  * Returns JS code for dynamically loading sourced (external) scripts.
  * @param srcList A list of paths for scripts that should be loaded.
  */
-function createLoaderScript(srcList: SrcScriptTag[], enableTrustedTypes = false): string {
+function createLoaderScript(
+  srcList: SrcScriptTag[],
+  enableTrustedTypes = false,
+  inBody = true,
```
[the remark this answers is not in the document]
aaronshim: No longer tracking `inBody`-- thanks!

packages/angular/build/src/utils/index-file/auto-csp.ts
```diff
@@ -258,7 +267,11 @@ function getStrictCsp(
  * Returns JS code for dynamically loading sourced (external) scripts.
  * @param srcList A list of paths for scripts that should be loaded.
```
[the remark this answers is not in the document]
aaronshim: No longer tracking `inBody`-- thanks!

dgp1130 (approved):
Looks great, thanks!

dgp1130:

PR #29062, opened 2024-12-05
ci: update dev-infra sha to use latest Bazel keys

alan-agius4:
This commit consumes the latest version of dev-infra to fix CI.

See: angular/dev-infra@40b2cbd


josephperrott (approved):
LGTM

Issue #29003, opened 2024-12-01
Define, webpack transforms and esbuild plugins are not supported in v19 Karma Buillder


chihab:
### Command

test

### Is this a regression?

- [x] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

v18

### Description

There is no way to pass in webpack transforms or esbuild plugins or define to the internal Angular builders used in Karma.

### Browser builder
When using webpack (browser) builder, webpack transforms are ignored:
https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/builders/karma/browser_builder.ts#L37

I guess they should be passed in as a third parameter to [initializeBrowser](https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/builders/karma/browser_builder.ts#L114)

as it used to be passed in in v18 here https://github.com/angular/angular-cli/blob/18.2.x/packages/angular_devkit/build_angular/src/builders/karma/index.ts#L87

### Application builder
When using the newly supported esbuild builder in Karma, esbuild `plugins` and `define` are not supported 
Define option is not supported in Karma builder (neither exposed in the builder schema nor in the builder api)
https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/builders/karma/application_builder.ts#L354

Regarding the application builder, I am not sure whether it is a feature request or a bug as the application builder support is new. 

Question: why are we using an internal builder with a closed configuration rather than using  the one defined in angular.json by the user?

### Browser builder

```ts
return executeKarmaBuilder(options, context, {
  webpackConfiguration: (webpackConfig) => {
    console.log("webpackConfig", webpackConfig);
    return webpackConfig;
  },
});
```

### Application builder

```json
"test": {
  "builder": "@angular-devkit/build-angular:karma",
  "options": {
    "polyfills": ["zone.js", "zone.js/testing"],
    "tsConfig": "tsconfig.spec.json",
    "assets": [
      {
        "glob": "**/*",
        "input": "public"
      }
    ],
    "define": {
      "ENV_MESSAGE": "'Hello from Angular 19'"
    },
    "styles": ["src/styles.css"],
    "scripts": []
  }
}
```

```
> ng-app19@0.0.0 test
> ng test

Error: Schema validation failed with the following errors:
  Data path "" must NOT have additional properties(define).
```

### Your Environment

```text
Angular CLI: 19.0.0
Node: 20.13.1
[...]

PR #29066, opened 2024-12-05
fix(@angular-devkit/build-angular): webpack config transform for karma

jkrems:
This PR is 10% passing an argument that got dropped during refactoring and 90% finding a way to test the behavior.

I'm making up a new concept for the harness (`additionalExecuteArguments`) which is pretty ugly but I couldn't think of a cleaner way of doing it. There might be a way to capture rest arguments of the `execute` function signature inside of the harness to make the cast go away. But I'm not 100% convinced that the more complicated types leaking throughout the harness code would really be worth it for this edge case.

See: https://github.com/angular/angular-cli/issues/29003


clydin (approved):

jkrems:

PR #29073, opened 2024-12-06
fix(@angular/ssr):  include `Content-Language` header when locale is set

alan-agius4:
The server now includes the `Content-Language` HTTP header in responses whenever a locale is explicitly set.


dgp1130 (approved):
Thanks for adding this! Glad we can align more closely with the spec.

alan-agius4:

Issue #29092, opened 2024-12-09
SSR build prerender broken if configuration has externalDependencies options


klerick:
### Command

build

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

Dear Angular Team,

We are experiencing an issue with external dependencies when using Angular 19. In our project, we utilize importmap to load external dependencies and have a custom registry hook for resolving these dependencies on our SSR server. However, we encounter an error when attempting to build the project in version 19.

Causes of the Issue:

1. Dependency Filtering:
   - This part of the code (https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/builders/application/execute-build.ts#L160) only retains Node.js dependencies for the server, leading to issues with external dependencies.

2. Prerendering:
   - Even though we do not have prerender routes, this section (https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/builders/application/execute-post-bundle.ts#L133) attempts to call the prerender hook, resulting in an error.
   - We are unable to set partialSSRBuild as it is an internal variable, and appShellOptions and prerenderOptions are set by default when outputMode is specified.

3. Custom Hook:
   - Custom load hook (https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts) does not load dependencies as expected.

### Minimal Reproduction

https://stackblitz.com/edit/stackblitz-starters-iudtwvic?file=package.json

npm run build.

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 19.0.4
Node: 20.13.1
Package Manager: npm 10.5.2
OS: darwin arm64

Angular: 19.0.3
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... platform-server, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1900.4
@angular-devkit/build-angular   19.0.4
@angular-devkit/core            19.0.4
@angular-devkit/schematics      19.0.4
@angular/cdk                    19.0.2
@angular/cli                    19.0.4
@angular/ssr                    19.0.4
@schematics/angular             19.0.4
rxjs                            7.8.1
typescript                      5.6.3
zone.js                         0.15.0
```

### Possible Solutions

1. Expose partialSSRBuild Configuration:
[...]

PR #29094, opened 2024-12-09
fix(@angular/build): handle external `@angular/` packages during SSR

alan-agius4:
This commit introduces `ngServerMode` to ensure proper handling of external `@angular/` packages when they are used as externals during server-side rendering (SSR).

Closes: #29092

### DIFF
--- a/packages/angular/build/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
+++ b/packages/angular/build/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
@@ -14,6 +14,11 @@ import { fileURLToPath } from 'url';
 import { JavaScriptTransformer } from '../../../tools/esbuild/javascript-transformer';
 
 /**
+ * @note For some unknown reason, setting `globalThis.ngServerMode = true` does not work when using ESM loader hooks.
+ */
+const NG_SERVER_MODE_INIT_BYTES = new TextEncoder().encode('var ngServerMode=true;');
+
+/**
  * Node.js ESM loader to redirect imports to in memory files.
  * @see: https://nodejs.org/api/esm.html#loaders for more information about loaders.
  */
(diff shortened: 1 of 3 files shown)


packages/angular/build/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
```diff
@@ -133,7 +138,12 @@ export async function load(url: string, context: { format?: string | null }, nex
   // need linking are ESM only.
   if (format === 'module' && isFileProtocol(url)) {
     const filePath = fileURLToPath(url);
-    const source = await javascriptTransformer.transformFile(filePath);
+    let source = await javascriptTransformer.transformFile(filePath);
+
+    if (filePath.includes('@angular/')) {
```
klerick, line 143: What happens if the name does not include '@angular/'?

alan-agius4: It will be handled by the standard Node.js resolution, this is the same as what happens during `ng serve`.

packages/angular/build/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
```diff
@@ -133,7 +138,12 @@ export async function load(url: string, context: { format?: string | null }, nex
...
+      // Prepend 'var ngServerMode=true;' to the source.
+      source = new Uint8Array([...NG_SERVER_MODE_INIT_BYTES, ...source]);
```
clydin, line 145:
```suggestion
      source = Buffer.concat([NG_SERVER_MODE_INIT_BYTES, source]);
```

clydin (approved):

alan-agius4:

PR #29101, opened 2024-12-10
fix(@angular/ssr): disable component bootstrapping during route extraction

### LINKED ISSUE #29085: Route extraction shouldnt load the whole app in dev server
I dont want to be that type of guy that just reopens a issue. But i only saw @alan-agius4 answers on that topic and i would kindly ask other angular engineers about their opinion and if its something that can be solved differently. Maybe without @alan-agius4 indirect implication that it would "only" be [solveable with file based](https://github.com/angular/angular-cli/issues/28995#issuecomment-2510789006) routing.

There are no docs on route extraction, but my understanding now is that all routes are computed in advance. I'm not deeper into angular code than the public_api, but I can't imagine that a full application load (full lifecycle or template execution) are necessary to figure this out. What does this have to do with route extraction?

[proof of concept](https://stackblitz.com/~/github.com/lennybakkalian/angular-bug?file=src/app/app.component.ts)
In this example, it would only make sense for SSG if the REQUEST token is not available on the server. (but according to @alan-agius4 its "[bad practice](https://github.com/angular/angular-cli/issues/28727#issuecomment-2441443287)" to think that the `REQUEST` token is available on the server with disabled prerender) - Just think about that for a second.

Sorry if I'm a bit salty, but it took me a really long time to figure out that (imo this bug) is due to the angular dev server.

Related to:
https://github.com/angular/angular-cli/issues/28995
https://github.com/angular/angular-cli/issues/28727#issuecomment-2441443287
https://github.com/angular/angular-cli/issues/29084 (my previous ticket)
[Angular Discord - My thread](https://discord.com/channels/748677963142135818/1315303097664409691)
[Angular Discord - Another thread](https://discord.com/channels/748677963142135818/1308766801919869009)
and probably many more

alan-agius4:
This commit disables component bootstrapping during route extraction to prevent invoking the AppComponent and its lifecycle hooks.

Closes #29085

### DIFF
--- a/packages/angular/ssr/src/routes/ng-routes.ts
+++ b/packages/angular/ssr/src/routes/ng-routes.ts
@@ -7,7 +7,16 @@
  */
 
 import { APP_BASE_HREF, PlatformLocation } from '@angular/common';
-import { ApplicationRef, Compiler, Injector, runInInjectionContext, ɵConsole } from '@angular/core';
+import {
+  APP_INITIALIZER,
+  ApplicationRef,
+  Compiler,
+  ComponentRef,
+  Injector,
+  inject,
+  runInInjectionContext,
+  ɵConsole,
+} from '@angular/core';
 import { INITIAL_CONFIG, platformServer } from '@angular/platform-server';
 import {
   Route as AngularRoute,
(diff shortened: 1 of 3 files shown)


alan-agius4:

PR #29109, opened 2024-12-11
feat(@schematics/angular): use signal in app component

cexbrayat:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?

Now that signals are recommended, this updates the generated `AppComponent` to use a signal for the `title` field.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/schematics/angular/application/files/common-files/src/app/app.html.template
+++ b/packages/schematics/angular/application/files/common-files/src/app/app.html.template
@@ -225,7 +225,7 @@
           <clipPath id="a"><path fill="#fff" d="M0 0h982v239H0z" /></clipPath>
         </defs>
       </svg>
-      <h1>Hello, {{ title }}</h1>
+      <h1>Hello, {{ title() }}</h1>
       <p>Congratulations! Your app is running. 🎉</p>
     </div>
     <div class="divider" role="separator" aria-label="Divider"></div>
--- a/packages/schematics/angular/application/files/module-files/src/app/app.ts.template
+++ b/packages/schematics/angular/application/files/module-files/src/app/app.ts.template
@@ -1,9 +1,9 @@
-import { Component } from '@angular/core';
+import { Component, signal } from '@angular/core';
 
 @Component({
   selector: '<%= selector %>',<% if(inlineTemplate) { %>
   template: `
-    <h1>Welcome to {{title}}!</h1>
+    <h1>Welcome to {{ title() }}!</h1>
 
     <% if (routing) {
      %><router-outlet /><%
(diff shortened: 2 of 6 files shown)


packages/schematics/angular/application/files/module-files/src/app/app.component.ts.template
```diff
@@ -15,5 +15,5 @@ import { Component } from '@angular/core';
   styleUrl: './app.component.<%= style %>'<% } %>
 })
 export class AppComponent {
-  title = '<%= name %>';
+  title = signal('<%= name %>');
```
alan-agius4, line 18: Is it logical for this to be considered a signal, given that it's just a static string?
PS: In case we go forward with this we need to update of a couple of other templates.

alan-agius4: I am not too strongly about this, but I wonder if it would send the wrong "signal", that anything referenced in the template needs to be a `signal`.

I'll defer to @alxhub. An alternative for correctness maybe I could see this title having the `readonly` modifier added.

dgp1130: Looking through some older PRs and chatted with @alxhub about this. It's definitely not great that this generates a binding which naturally requires Zone.js to update whenever `title` changes, which isn't a good practice to set right now. The options here are probably:
1. Use `title = signal('...');` to allow it to change in a reactive fashion.
2. Use `readonly title = '...';` to signify that the value should _not_ change.
3. Inline the title in the template directly, remove the binding altogether.

Part of the goal here is for devs to have something to play with when they aren't familiar with Angular and just experimenting. So 3., doesn't work because we'd want to have _some_ kind of binding just to display it. 2. also isn't terribly interesting, and if the value is `readonly`, it begs the question of why it shouldn't be inlined in the template in the first place.

So Alex and I aligned on 1. as the best path forward here.

Small nit: I'd do `protected readonly title = signal('...');` to signify that `this.title = ...` is an error and to hint that template binding doesn't need to be public.

> PS: In case we go forward with this we need to update of a couple of other templates.

@alan-agius4, what other templates are you thinking of? Do we generate many bindings in others?

cexbrayat: Thanks for you feedback @dgp1130 , that's exactly what I think as well.
I'll update the PR to use `protected readonly title = signal(...)` 👍

cexbrayat:
@alan-agius4 @dgp1130 PR is ready for review with your feedback

alan-agius4 (approved):
LGTM

dgp1130:
Thanks for the contribution, apologies for the long delay on this!

PR #29123, opened 2024-12-12
refactor(@angular/build): only load chunk optimizer if enabled

clydin:
The experimental chunk optimizer is now only imported if it is enabled via its environment variable (`NG_BUILD_OPTIMIZE_CHUNKS=1`). This prevents the loading of the rollup package as well as any transitive dependencies when these packages will not be used by the build.


jkrems:

PR #29148, opened 2024-12-16
ci: sync package locks information for aspect with renovate

alan-agius4:
This is needed as otherwise the `npm_translate_lock` file will be out of sync


josephperrott (approved):
LGTM

PR #29150, opened 2024-12-16
ci: ignore changes to `.yarn/releases`

alan-agius4:
There is an unidentified issue causing the Yarn binary to be altered, resulting in packages not installing correctly and leading to failures in the process.


josephperrott (approved):
LGTM

PR #29152, opened 2024-12-16
ci: disable `pnpm-lock.yaml` update via renovate

alan-agius4:
This commit tries to display `pnpm-lock.yaml` update using renovate and instead it uses `yarn bazel run @npm2//:sync`.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -14,7 +14,7 @@
       "yarn install --frozen-lockfile --non-interactive",
       "yarn bazel run @npm2//:sync"
     ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -31,6 +31,7 @@
     "tests/legacy-cli/e2e/ng-snapshot/package.json",
     ".github/workflows/**/*.yml"
   ],
+  "ignorePaths": ["pnpm-lock.yaml"],
```
josephperrott, line 34: Should this be done instead with adding this into the package rules?

```
{
  "matchPackageManagers": ["pnpm"],
  "enabled": false
},
```

PR #29159, opened 2024-12-17
Renovate fixes

alan-agius4:
**ci: remove `ignorePaths` paths for `pnpm-lock.yaml`**
This does not work as expected because the paths are expected to be package files.

**build: update `engines` field in `package.json`**
This commit removes the `yarn` engine field. The `packageManager` version was previously removed, causing Renovate to pick up an outdated Yarn version. Additionally, Yarn Berry does not use this field, rendering it unnecessary.

**build: workaround for `ERROR: An error occurred during the fetch of repository 'npm2'`**

In some cases `yarn bazel run @npm2//:sync` will fail in the first run see: https://github.com/aspect-build/rules_js/issues/1445 as a workaround we run a build before.

```
ERROR: An error occurred during the fetch of repository 'npm2':
   Traceback (most recent call last):
        File "/usr/local/google/home/alanagius/.cache/bazel/_bazel_alanagius/2fa837e4c5ce941f68b762e5f8e7dc4d/external/aspect_rules_js/npm/private/npm_translate_lock.bzl", line 112, column 21, in _npm_translate_lock_impl
                fail(msg)
Error in fail:

INFO:


josephperrott (approved):
LGTM

PR #29162, opened 2024-12-17
fix(@angular/ssr): ensure correct `Location` header for redirects behind a proxy

### LINKED ISSUE #29151: Angular server redirects to incorrect Location when serving the app behind a proxy
### Command

serve

### Is this a regression?

_No response_

### Description

When deploying Angular ssr server on services like Cloud Run or Firebase App Hosting where the application is served behind a frontend proxy, the Location header is incorrectly constructed, and redirect fails.

### Minimal Reproduction

app.routes.ts

```
import { Routes } from '@angular/router';

export const routes: Routes = [
  { path: 'blog', loadChildren: () => import('./blog/blog-routing.module').then(m => m.routes) },
  { path: '**', redirectTo: '/blog' },
];
```

```bash
$ ng build && node dist/testme/server/server.mjs
Node Express server listening on http://localhost:4000
```

This set up works fine in localhost:

```bash
$ curl -v localhost:4000
* Host localhost:4000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:4000...
* Connected to localhost (::1) port 4000
> GET / HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 302 Found
< X-Powered-By: Express
< location: http://localhost:4000/blog
< Date: Mon, 16 Dec 2024 18:13:02 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< Content-Length: 0
<
* Connection #0 to host localhost left intact
```

But when connecting to a server behind a frontend proxy, the `Location` header is wrong

```bash
~ curl -v https://ng-redirect-bug--danielylee-90.us-central1.hosted.app

* Host ng-redirect-bug--danielylee-90.us-central1.hosted.app:443 was resolved.
...
> GET / HTTP/2
> Host: ng-redirect-bug--danielylee-90.us-central1.hosted.app
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/2 302
< alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
< alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
< content-type: text/html
< date: Mon, 16 Dec 2024 18:14:33 GMT
< location: https://ng-redirect-bug--danielylee-90.us-central1.hosted.app:8080/blog
< server: envoy
< x-cloud-trace-context: c08777b57d0762a74338d2597fedc290;o=1
< x-powered-by: Express
< content-length: 0
< server-timing: l2gfet4t7; dur=65
< via: 1.1 google
< cdn-cache-status: miss
<
* Connection #0 to host ng-redirect-bug--danielylee-90.us-central1.hosted.app left intact
```

Note the header `location: https://ng-redirect-bug--danielylee-90.us-central1.hosted.app:8080/blog`

Specifically, the port is set to be `8080` which is the port used on the server, but the port used in the frontend proxy is actually at `443`.

### Exception or Error

```text

```

### Your Environment

```bash
$ ng version

                |___/


Angular CLI: 19.0.1
Node: 18.19.0
Package Manager: npm 10.2.3
OS: darwin arm64

Angular: 19.0.0
... router

@angular-devkit/architect       0.1702.3
@angular-devkit/build-angular   19.0.1
@angular-devkit/core            17.2.3
@angular-devkit/schematics      19.0.1
@angular/cli                    19.0.1
@angular/ssr                    19.0.1
@schematics/angular             19.0.1
```

### Anything else relevant?

The relevant code seems to be at

https://github.com/angular/angular-cli/blob/e4ff5c9d4aa544e14a03f4f95a29f6e78e85d7a5/packages/angular/ssr/node/src/request.ts#L88

alan-agius4:
Previously, when the application was served behind a proxy, server-side redirects generated an incorrect Location header, causing navigation issues. This fix updates `createRequestUrl` to use the port from the Host header, ensuring accurate in proxy environments. Additionally, the Location header now only contains the pathname, improving compliance with redirect handling in such setups.

Closes #29151

### DIFF
--- a/packages/angular/ssr/node/src/request.ts
+++ b/packages/angular/ssr/node/src/request.ts
@@ -83,18 +83,40 @@ function createRequestUrl(nodeRequest: IncomingMessage | Http2ServerRequest): UR
     originalUrl,
   } = nodeRequest as IncomingMessage & { originalUrl?: string };
   const protocol =
-    headers['x-forwarded-proto'] ?? ('encrypted' in socket && socket.encrypted ? 'https' : 'http');
-  const hostname = headers['x-forwarded-host'] ?? headers.host ?? headers[':authority'];
-  const port = headers['x-forwarded-port'] ?? socket.localPort;
+    getFirstHeaderValue(headers['x-forwarded-proto']) ??
+    ('encrypted' in socket && socket.encrypted ? 'https' : 'http');
+  const hostname =
+    getFirstHeaderValue(headers['x-forwarded-host']) ?? headers.host ?? headers[':authority'];
 
   if (Array.isArray(hostname)) {
     throw new Error('host value cannot be an array.');
   }
 
   let hostnameWithPort = hostname;
-  if (port && !hostname?.includes(':')) {
-    hostnameWithPort += `:${port}`;
+  if (!hostname?.includes(':')) {
+    const port = getFirstHeaderValue(headers['x-forwarded-port']);
+    if (port) {
+      hostnameWithPort += `:${port}`;
+    }
   }
 
   return new URL(originalUrl ?? url, `${protocol}://${hostnameWithPort}`);
+}
+
+/**
+ * Extracts the first value from a multi-value header string.
+ *
+ * @param value - A string or an array of strings representing the header values.
+ *                           If it's a string, values are expected to be comma-separated.
+ * @returns The first trimmed value from the multi-value header, or `undefined` if the input is invalid or empty.
+ *
+ * @example
+ * ```typescript
+ * getFirstHeaderValue("value1, value2, value3"); // "value1"
+ * getFirstHeaderValue(["value1", "value2"]); // "value1"
+ * getFirstHeaderValue(undefined); // undefined
+ * ```
+ */
+function getFirstHeaderValue(value: string | string[] | undefined): string | undefined {
+  return value?.toString().split(',', 1)[0]?.trim();
 }
--- a/packages/angular/ssr/src/app.ts
+++ b/packages/angular/ssr/src/app.ts
@@ -161,14 +161,15 @@ export class AngularServerApp {
 
     const { redirectTo, status, renderMode } = matchedRoute;
     if (redirectTo !== undefined) {
-      return Response.redirect(
-        new URL(buildPathWithParams(redirectTo, url.pathname), url),
+      return new Response(null, {
         // Note: The status code is validated during route extraction.
         // 302 Found is used by default for redirections
         // See: https://developer.mozilla.org/en-US/docs/Web/API/Response/redirect_static#status
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        (status as any) ?? 302,
-      );
+        status: status ?? 302,
+        headers: {
+          'Location': buildPathWithParams(redirectTo, url.pathname),
+        },
+      });
     }
 
     if (renderMode === RenderMode.Prerender) {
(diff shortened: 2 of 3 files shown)


packages/angular/ssr/node/src/request.ts
```diff
@@ -83,18 +83,42 @@ function createRequestUrl(nodeRequest: IncomingMessage | Http2ServerRequest): UR
...
+ * @returns The first trimmed value from the multi-value header, or `undefined` if the input is invalid or empty.
+ *
+ * @example
+ * ```typescript
+ * getFirstForwardedValue("value1, value2, value3"); // "value1"
+ * getFirstForwardedValue(["value1", "value2"]); // "value1"
+ * getFirstForwardedValue(undefined); // undefined
+ * ```
+ */
+function getFirstForwardedValue(
```
dgp1130, line 120: Nit: I don't think this actually has anything to do with forwarded headers, could we simplify this to just `getFirstHeaderValue`?

packages/angular/ssr/node/src/request.ts
```diff
@@ -83,18 +83,42 @@ function createRequestUrl(nodeRequest: IncomingMessage | Http2ServerRequest): UR
...
 
   let hostnameWithPort = hostname;
-  if (port && !hostname?.includes(':')) {
-    hostnameWithPort += `:${port}`;
+  if (!hostname?.includes(':')) {
+    const port = getFirstForwardedValue(headers['x-forwarded-port']);
+    if (port) {
+      hostnameWithPort += `:${port}`;
+    }
   }
```
dgp1130, line 101: Question: Do we care about the case where two ports are specified like:

```
X-Forwarded-Host: example.com:1234
X-Forwarded-Port: 4321
```

Seems like here we'll favor `1234`. Is that intentional or are we calling this unspecified behavior?

packages/angular/ssr/src/app.ts
```diff
@@ -161,14 +161,15 @@ export class AngularServerApp {
...
+      return new Response(null, {
         // Note: The status code is validated during route extraction.
         // 302 Found is used by default for redirections
         // See: https://developer.mozilla.org/en-US/docs/Web/API/Response/redirect_static#status
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        (status as any) ?? 302,
-      );
+        status: status ?? 302,
+        headers: {
+          'Location': buildPathWithParams(redirectTo, url.pathname),
```
dgp1130, line 170: Question: If we're manually constructing a redirect, any particular reason to still use the forwarded hostname/port? Why not just `/home`?

dgp1130 (approved):

packages/angular/ssr/src/app.ts
dgp1130: Oh rereading the PR it seems like this is relative now.

packages/angular/ssr/node/src/request.ts
alan-agius4: From what I understand the `X-Forwarded-Host` should not have a `port`. Unlike the `host` header.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

dgp1130: Oh I see, this is intended for the `Host: example.com:1234` case. Seems like it will also catch the `X-Forwarded-Host: example.com:1234` case, but I'm fine calling that out of scope if it's not expected for that header.

I was going to point out that "host" as a term typically does not include a port, but since this is a de-facto standard I'm not sure what to expect from `X-Forwarded-Host`.

dgp1130:

PR #29164, opened 2024-12-17
refactor(@angular/ssr): mark `LINK_LOAD_SCRIPT_CONTENT` as `@PURE`

alan-agius4:
This const is pure.


dgp1130 (approved):
How exactly does marking this pure help us? Does it allow this string to be pruned from bundles when unused or something like that?

dgp1130:

PR #29168, opened 2024-12-17
ci: restore `pnpm-lock.yaml` that was updated by renovate

alan-agius4:
Frce this fill to be generated by the tooling


josephperrott (approved):
LGTM

Issue #29190, opened 2024-12-19
Angular 19.0.6 breaks karma/jasmine on our build server by trying to load jasmine.css as a script module


hamfastgamgee:
### Command

test

### Is this a regression?

- [x] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

19.0.5

### Description

After the 19.0.6 release (and presumably 19.1.0-next.2, although I have not tested that), our build server cannot execute tests appropriately because the new AngularPolyfillsPlugin executed to fix jasmine.clock turns the <link> tag that is supposed to load jasmine.css into a <script type="module"> tag, which is clearly inappropriate for CSS.  Interestingly, "ng test" executes fine on my local workstation (although the styles in karma if you load it up in the browser are broken) even with this change, but on our build server this results in the following error:

Chrome Headless 131.0.0.0 (Windows 10) ERROR
Uncaught Error loading node_modules/jasmine-core/lib/jasmine-core/jasmine.css
   at http://localhost:9876/context.html:255:289

Rolling back to a pinned 19.0.5 works; obviously the bug that was addressed would be broken, but we don't use jasmine.clock in our tests.  As far as I can tell, adding some additional logic around line 151 to ignore CSS files in application_builder.ts should do the trick.

### Minimal Reproduction

Run "ng test" and observe the source code generated in the karma server, noting that the CSS file is improperly referenced as <script type="module">.

(I do not know why our build server fails here and my workstation succeeds, even running the same commands on the same Chrome version.)

### Exception or Error

```text
Chrome Headless 131.0.0.0 (Windows 10) ERROR
Uncaught Error loading node_modules/jasmine-core/lib/jasmine-core/jasmine.css
   at http://localhost:9876/context.html:255:289
```

### Your Environment

```text
_                      _                 ____ _     ___ 
    / \   _ __   __ _ _   _| | __ _ _ __     / ___| |   |_ _|
   / △ \ | '_ \ / _` | | | | |/ _` | '__|   | |   | |    | |
  / ___ \| | | | (_| | |_| | | (_| | |      | |___| |___ | |
 /_/   \_\_| |_|\__, |\__,_|_|\__,_|_|       \____|_____|___|
                |___/
    

Angular CLI: 19.0.6
Node: 20.11.1
Package Manager: npm 10.9.2
OS: win32 x64

Angular: 19.0.5
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
[...]

PR #29193, opened 2024-12-19
fix(@angular-devkit/build-angular): preserve css type for jasmine.css

jkrems:
We should only force the type for files that we know are JavaScript. Otherwise we risk breaking the magic type detection done by Karma.

The previous code broke for `jasmine.css`.

Fixes https://github.com/angular/angular-cli/issues/29190


clydin (approved):

clydin:

PR #29194, opened 2024-12-20
fix(@angular/build): mitigate JS transformer worker `execArgv` errors

clydin:
Node.js workers will currently fail to initialize if the `execArgv` option is used and it contains v8 specific options. This is currently problematic for the JS transformer worker because it contains a workaround to remove the SSR `--import` argument that is used to add a loader hook for SSR purposes. The filtering of the argument and subsequent use of the `execArgv` array had the potential to pass custom Node.js options to the worker and cause it to fail. These options can be passed by developers on the command line when invoking the Angular CLI.
To mitigate this problem, the `execArgv` option is now only filtered and used if the SSR import argument is present in the array. Otherwise, no value is passed which allows the default Node.js behavior to be used. While this does not fully solve the problem for all projects, it does remove the problem from non-SSR projects.


alan-agius4 (approved):

alan-agius4 (approved):

clydin:

PR #29200, opened 2024-12-20
build: set a mnemonic for template expand and file write

josephperrott:
Set the mnemonic to determine if it allows us to reduce the number of RBE queries executed


clydin (approved):

clydin:

PR #29204, opened 2024-12-20
build: migrate `@angular-devkit/build-webpack` to `ts_project`

clydin:
The `@angular-devkit/build-webpack` package has been migrated to the `rules_js` ts_project rule.

### DIFF
--- a/packages/angular_devkit/build_webpack/BUILD.bazel
+++ b/packages/angular_devkit/build_webpack/BUILD.bazel
@@ -5,7 +5,8 @@
 
 load("@npm//@angular/build-tooling/bazel/api-golden:index.bzl", "api_golden_test_npm_package")
 load("@npm//@bazel/jasmine:index.bzl", "jasmine_node_test")
-load("//tools:defaults.bzl", "pkg_npm", "ts_library")
+load("//tools:defaults.bzl", "pkg_npm")
+load("//tools:interop.bzl", "ts_project")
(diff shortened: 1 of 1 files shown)


packages/angular_devkit/build_webpack/BUILD.bazel
```diff
@@ -66,34 +67,32 @@ ts_library(
...
+    interop_deps = [
         "//packages/angular_devkit/core",
         "//packages/angular_devkit/core/node",
         "//packages/ngtools/webpack",
-        "@npm//@angular/common",
-        "@npm//@angular/compiler",
-        "@npm//@angular/compiler-cli",
-        "@npm//@angular/core",
-        "@npm//@angular/platform-browser",
-        "@npm//@angular/platform-browser-dynamic",
```
devversion, line 82: do you know why these moved from `ts_project` deps to the `data` deps below? would love to avoid this in general

clydin: These are used by the generated CLI project used inside the tests.  Treating them as data for the tests is probably more accurate then where they were before since they are not needed for building the actual unit tests.

devversion: I see, that makes sense then

devversion (approved):

clydin:

PR #29224, opened 2024-12-24
build: migrate `@angular/build` to `ts_project`

devversion:
This commit migrates `@angular/build` to `ts_project`.

Note: The `tsconfig.json` can be cleaned up in a follow-up once we have standardized on `index.ts`


josephperrott (approved):
LGTM

PR #29252, opened 2025-01-06
fix(@angular/build): handle relative URLs when constructing new URLs during server fetch

### LINKED ISSUE #29236: SSG allow loading static assets
### Command

build

### Is this a regression?

17

### Description

I am using [transloco](https://github.com/jsverse/transloco) for localization.
When building (used to be prerendering) now the build process tries to render the pages.
In doing so, it hits this loader, which loads a static asset (the language in question)
```ts
@Injectable({providedIn: 'root'})
export class HttpLoader implements TranslocoLoader {
  private http = inject(HttpClient);

  getTranslation(langPath: string): Observable<Translation> {
    return this.http.get<Translation>(`assets/i18n/${langPath}.json`);
  }
}
```
However, the HTTP request fails, since there seems to not be an http server involved in the SSG process.

If I run this in SSR, it also fails, but that is because it can't handle the path, and requires a full URL (http://localhost:4000/assets/...)

If I accept not having a server, I need to have a similar behavior to what I do during testing:
```ts
    if (isPlatformServer(this.platformId)) {
      const path = require('path');
      const filePath = path.join(__dirname, '..', '..', '..', assetPath);
      console.error('filePath', filePath);
      const fs = require('fs');

      const data = fs.readFileSync(filePath, 'utf8');
      return of(JSON.parse(data));
    }
```

But this relies on node modules, and the builder errors:
>   The package "fs" wasn't found on the file system but is built into node. Are you trying to bundle for node? You can use "platform: 'node'" to do that, which will remove this error.

Therefore, my issues are:

1. SSG should have some sort of way to load static assets. 
2. SSR should not need the full URL. somehow, it should understand that any path requested is under the same host as the REQUEST.

### Minimal Reproduction

1. Make a new app, with ssr and ssg (`"prerender": true`, "ssr": {"entry": "src/server.ts"}`)
2. under `src/assets` add `something.json`
3. in the `AppComponent`'s `ngOnInit` function, make `this.httpClient.get('/assets/something.json').subscribe()`
4. run `ng build --configuration=production`

### Exception or Error

```text
- SSG should have a way to load static assets
- SSR should be able to load assets using HTTP
```

### Your Environment

```text
Angular CLI: 19.0.6
Node: 22.0.0
Package Manager: npm 10.5.1
OS: darwin arm64

Angular: 19.0.5
... router, service-worker

@angular/cdk                    19.0.4
@angular/material               19.0.4
@angular/ssr                    19.0.6
```

### Anything else relevant?

In the old prerendering, it used to work.

alan-agius4:
Ensures proper handling of relative URLs to prevent errors in server-side fetch operations.

Closes #29236

### DIFF
--- a/packages/angular/build/src/utils/server-rendering/fetch-patch.ts
+++ b/packages/angular/build/src/utils/server-rendering/fetch-patch.ts
@@ -28,9 +28,9 @@ export function patchFetchToLoadInMemoryAssets(baseURL: URL): void {
     if (input instanceof URL) {
       url = input;
     } else if (typeof input === 'string') {
-      url = new URL(input);
+      url = new URL(input, baseURL);
     } else if (typeof input === 'object' && 'url' in input) {
-      url = new URL(input.url);
+      url = new URL(input.url, baseURL);
     } else {
       return originalFetch(input, init);
     }


clydin (approved):

alan-agius4:

PR #29261, opened 2025-01-06
ci: run less jobs per executor for RBE

josephperrott:
Use less jobs per bazel run to reduce the rate at which we perform API calls for RBE


clydin (approved):

jkrems:

PR #29270, opened 2025-01-07
feat(@angular/build): add ng-packagr builder to the package 

clydin:
To support migration to the `@angular/build` package which contains the
`application` builder that is used by all new projects, the `ng-packagr`
builder used to build Angular libraries is also now available within this
package. This removes the need for projects that are using the application
builder but also would like to build a library from having to install the
Webpack related `@angular-devkit/build-angular` package.  This can result
in a significant reduction in overall Node.js packages installed within the
project.

### DIFF
--- a/goldens/public-api/angular/build/index.api.md
+++ b/goldens/public-api/angular/build/index.api.md
@@ -150,12 +150,23 @@ export function executeDevServerBuilder(options: DevServerBuilderOptions, contex
 export function executeExtractI18nBuilder(options: ExtractI18nBuilderOptions, context: BuilderContext, extensions?: ApplicationBuilderExtensions): Promise<BuilderOutput>;
 
 // @public
+export function executeNgPackagrBuilder(options: NgPackagrBuilderOptions, context: BuilderContext): AsyncIterableIterator<BuilderOutput>;
+
+// @public
 export interface ExtractI18nBuilderOptions {
     buildTarget?: string;
     format?: Format;
     outFile?: string;
     outputPath?: string;
     progress?: boolean;
+}
+
+// @public
+export interface NgPackagrBuilderOptions {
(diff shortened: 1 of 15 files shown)


pweyrich:
Can this be called the library builder (@angular/build:library) instead?
In my opinion, this would perfectly complement the application builder.

clydin:
There will be a library builder in the future but it will effectively be rewrite that will leverage the compilation internals of the application builder.
The goal with this PR is to remove the need to install a large amount of transitive Webpack related dependencies just to build a library.

pweyrich:
> There will be a library builder in the future but it will effectively be rewrite that will leverage the compilation internals of the application builder.
> The goal with this PR is to remove the need to install a large amount of transitive Webpack related dependencies just to build a library.

Got it. Thanks for elaborating on this.
Looking forward!

alan-agius4 (approved):
LGTM, just a couple of questions.

1. Do you intend to remove the implementation from build-angular?
2. Should we move the spec files?
3. Should we export the builder from https://github.com/angular/angular-cli/blob/328854c99b706d78cbb0440102a71fe91eb3a481/packages/angular/build/src/index.ts?

jkrems:
Is this ready to be merged for today's rc?

clydin:
> LGTM, just a couple of questions.
> 
> 1. Do you intend to remove the implementation from build-angular?
> 2. Should we move the spec files?
> 3. Should we export the builder from https://github.com/angular/angular-cli/blob/328854c99b706d78cbb0440102a71fe91eb3a481/packages/angular/build/src/index.ts?

1. Yes, I plan on doing this in a followup refactor PR.
2. Same as above since the existing spec file is using the old test setup.
3. Added to this PR.

PR #29279, opened 2025-01-08
fix(@angular/build): pass `define` option defined in application builder to Vite prebundling

### LINKED ISSUE #29278: Define build option does not define values for prebundling
### Command

serve

### Is this a regression?

_No response_

### Description

The define build option is not passed on for prebundling https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/tools/vite/utils.ts#L101 and so library code that relies on a global variable being set cannot be prebundled.

### Minimal Reproduction

Pre-bundle a library that relies on a global.

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 19.0.6
Node: 20.12.1
Package Manager: npm 10.5.0
OS: darwin arm64

Angular: 19.0.5

```

### Anything else relevant?

_No response_

alan-agius4:
This update ensures that the `define` option is correctly passed to Vite during the prebundling phase of the application builder, improving compatibility and optimization of the build process.

Closes #29278

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -422,6 +422,7 @@ export async function* serveWithVite(
         componentStyles,
         templateUpdates,
         browserOptions.loader as EsbuildLoaderOption | undefined,
+        browserOptions.define,
         extensions?.middleware,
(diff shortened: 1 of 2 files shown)


alan-agius4:

PR #29280, opened 2025-01-08
build: resolve issue with uncompiled `@angular/build/private` being included

alan-agius4:
Reference: https://github.com/angular/angular-build-builds/commit/3735297d21181b5885cf9a3625cb554c46efb8b3


devversion:
I'm currently investigating that as well, not sure yet

PR #29283, opened 2025-01-08
build: update `@angular/ng-dev` to support `rulesJsInteropMode`

alan-agius4:
The current version does not support `rulesJsInteropMode` which causes the build to fail.


devversion (approved):

devversion:
Thank you ❤️  (weird that CI didn't fail)

alan-agius4:
It does on the patch branch https://github.com/angular/angular-cli/tree/19.0.x

PR #29286, opened 2025-01-08
fix(@angular/ssr): throw error when using route matchers

alan-agius4:
Route matchers are not currently supported in Angular SSR. This commit ensures an error is issued when a route matcher is detected.


clydin (approved):

denisyilmaz:
@alan-agius4 I was using `urlMatcher` in Angular 19.0 without any issues in our SSR project. Since 19.1 the application is broken. Is there a reason why the application needs to break instead of a soft error log?

alan-agius4:
@denisyilmaz, I replied on https://github.com/angular/angular-cli/issues/29384

PR #29288, opened 2025-01-08
ci: prevent `NpmPackageExtract` results from being placed remote cache

josephperrott:
Placing all of the node_modules files into remote cache is too much


devversion (approved):

jkrems:

PR #29290, opened 2025-01-08
build: never use interop targets for ts compilations

devversion:
* Removes `interop_deps` from the `ts_project` interop macro.
* Keeps `_rjs` suffix for now as we still need the interop targets for e.g. `jasmine_node_test` and the `rules_nodejs` linker.

In follow-ups we can remove the suffix, and interop layer.


josephperrott (approved):
LGTM

clydin (approved):

alan-agius4:

PR #29292, opened 2025-01-08
build: enable rulesJsInteropMode for releases

josephperrott:


alan-agius4:
This has already been added https://github.com/angular/angular-cli/blob/88431b7564d6757898744597a67fcdc178413128/.ng-dev/release.mjs#L35

PR #29295, opened 2025-01-08
Update next branch to reflect new release-train "v19.2.0-next.0".

jkrems:
The previous "next" release-train has moved into the release-candidate phase. This PR updates the next branch to the subsequent release-train.

Also this PR cherry-picks the changelog for v19.1.0-rc.0 into the main branch so that the changelog is up to date.


dgp1130 (approved):

alan-agius4 (changes requested):
You need to run `yarn bazel run @npm2//:sync`, we should do this as part of ng-dev as well. 

//cc @devversion

jkrems:
> You need to run `yarn bazel run @npm2//:sync`, and commit the generated hash file.
> 
> We should do this as part of ng-dev as well.
> 
> //cc @devversion

Done!

alan-agius4 (approved):

PR #29306, opened 2025-01-09
fix(@angular-devkit/core): handle Windows drive letter case insensitivity in path functions

### LINKED ISSUE #27029: @angular-devkit/core normalize+relative function return unexpected
### Command

other

### Is this a regression?

_No response_

### Description

Windows drive letters may not be case sensitive?

but use normalize windows path ,C://=>/C/ ,c://=>/c/

### Minimal Reproduction

```ts
import {
  normalize,
  relative,
    let a1 = `E://abc`;
    let a2 = `e://abc/action`;
    let c = relative(normalize(a1), normalize(a2));
    c;
```

and relative return 
image
If I compare directly using the Node API, the return result is normal

```ts
    let xxx=path.relative(a1,a2)

```

### Exception or Error

_No response_

### Your Environment

```text
Angular CLI: 17.0.1
Node: 18.18.0
Package Manager: npm 9.8.1
OS: win32 x64

Angular:
...

@angular-devkit/architect    0.1700.1 (cli-only)
@angular-devkit/core         17.0.1 (cli-only)
@angular-devkit/schematics   17.0.1 (cli-only)
@schematics/angular          17.0.1 (cli-only)
```

### Anything else relevant?

_No response_

alan-agius4:
This update ensures that path-related functions in account for the case-insensitivity of drive letters on Windows systems. By addressing this inconsistency, the functionality becomes more robust and aligned with Windows filesystem behavior.

Closes #27029

### DIFF
--- a/packages/angular_devkit/core/src/virtual-fs/path.ts
+++ b/packages/angular_devkit/core/src/virtual-fs/path.ts
@@ -233,7 +233,7 @@ export function noCacheNormalize(path: string): Path {
   // Match absolute windows path.
   const original = path;
   if (path.match(/^[A-Z]:[/\\]/i)) {
-    path = '\\' + path[0] + '\\' + path.slice(3);
+    path = '\\' + path[0].toUpperCase() + '\\' + path.slice(3);
   }
 
   // We convert Windows paths as well here.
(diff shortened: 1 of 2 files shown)


clydin (approved):

alan-agius4:
The changes were merged into the following branches: main, 19.0.x, 19.1.x

PR #29312, opened 2025-01-09
build: migrate `@angular-devkit/architect` to `npm_package`

devversion:
Migrates the `@angular-devkit/architect` package to the `rules_js` npm package rule, consuming the direct `rules_ts` output JS files.

Notably, substitution is FAR different than what it used to be with `rules_nodejs`, so we needed some extra work to leverage `make_template` for substitutions in `package.json` files. **Keep in mind** that for now, this does not apply to any other files; so we only substitute in the `package.json`, but not in e.g. `.js` files as before. We will follow-up on this.

The other jq merging/filtering for snapshot or tar references in `package.json` files is kept as is, and is temporarily duplicated. This is acceptable as the migration should be pretty smooth and quick.


tools/defaults2.bzl
```diff
@@ -0,0 +1,120 @@
...
+def jasmine_test(**kwargs):
+    _jasmine_test(
+        node_modules = "//:root_modules",
+        **kwargs
+    )
+
+def ts_project(**kwargs):
+    _ts_project(**kwargs)
+
+def npm_package(
```
josephperrott, line 22: This is complicated enough that it might warrant going into its own bzl file? A lot happens here for the npm package.

josephperrott (approved):
LGTM

clydin (approved):

tools/defaults2.bzl
```diff
@@ -0,0 +1,8 @@
+load("//tools:interop.bzl", _ts_project = "ts_project")
```
alan-agius4, line 1: NIT: maybe a more meaningful filename? (`defaults-interop`) etc..

devversion: Yeah, we'll rename that in a follow-up. Probably `bazel/ts_project_interop.bzl`

alan-agius4:

PR #29318, opened 2025-01-10
Update more packages to `npm_package`

devversion:
See individual commits

### DIFF
--- a/packages/angular/build/BUILD.bazel
+++ b/packages/angular/build/BUILD.bazel
@@ -229,6 +229,10 @@ npm_package(
     pkg_deps = [
         "//packages/angular_devkit/architect:package.json",
     ],
+    stamp_files = [
+        "src/tools/esbuild/utils.js",
+        "src/utils/normalize-cache.js",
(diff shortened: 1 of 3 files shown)


packages/angular/build/BUILD.bazel
```diff
@@ -28,6 +27,18 @@ ts_json_schema(
     src = "src/builders/ng-packagr/schema.json",
 )
 
+JSON_FILES = glob(
+    include = [
+        "src/**/schema.json",
+        "src/**/*.js",
```
alan-agius4, line 33: The `JSON_FILES` here is misleading since the glob is for a number of file types.

alan-agius4 (approved):
LGTM, one minor NIT

packages/angular/build/BUILD.bazel
devversion: :+1: I've named the other constant also `ASSET_FILES`. Will do the same here

clydin (approved):

PR #29325, opened 2025-01-10
build: update `_exec` to handle timeouts gracefully when `waitForMatch` does not match within 60 seconds.

alan-agius4:
Previously, if `waitForMatch` failed to find a match, the CI process could hang indefinitely. This commit updates the method to ensure it fails after a 60-second timeout, preventing such blocks.

Example: https://github.com/angular/angular-cli/actions/runs/12709516529/job/35428835418?pr=29318


clydin (approved):

alan-agius4:
The changes were merged into the following branches: main, 19.0.x, 19.1.x

PR #29340, opened 2025-01-13
refactor(@angular/build): move Angular compilation creation to build execution

clydin:
The creation of the internal Angular compilation used by the internal esbuild plugin is now created within the main execution phase of the build process. This is causes no behavior changes but provides the infrastructure for several addition refactoring improvements including performance improvements of the HMR process. The private API for the `createCompilerPlugin` has kept its function signature and currently is not affected by this change.


alan-agius4 (approved):

clydin:

PR #29349, opened 2025-01-14
Migrate first `jasmine_node_test` to new RJS variant

devversion:
See individual commits (review per-commit ideally)


pnpm-lock.yaml
```diff
@@ -12909,7 +12942,6 @@ packages:
   /source-map@0.7.4:
     resolution: {integrity: sha512-l3BikUxvPOcn5E74dZiq5BGsTb5yEwhaTSzccU6t4sDOH8NWJCstKO5QT2CvtFoK6F0saL7p9xHAqHOlCPJygA==}
     engines: {node: '>= 8'}
-    dev: true
```
devversion, line 12912: These dependencies lost their `dev` status because `packages/angular_devkit/core` requires them as non-dev. Perfectly fine!

**but** this brings up a good future opportunity in ensuring the root `package.json` matches the dependencies in the "shipped package.json". cc. @clydin. I assume this _would_ work if we could use pnpm for publishing as it would automatically e.g. replace `workspace:XXX` with the proper versions.

packages/angular_devkit/architect/BUILD.bazel
```diff
@@ -66,10 +68,9 @@ ts_project(
     data = JSON_FILES,
     module_name = "@angular-devkit/architect",
     deps = [
+        ":node_modules/@angular-devkit/core",
         "//:node_modules/@types/node",
         "//:node_modules/rxjs",
-        "//packages/angular_devkit/core:core_rjs",
```
devversion, line 71: This nicely prevents the cross-relative imports (like we decided in MUC!). It also fixes/improves some `.d.ts` output generation where TS currently emits e.g. `import('../core')`

packages/angular/build/src/builders/application/index.ts
```diff
@@ -259,4 +260,7 @@ function generateFullPath(
   return fullFilePath;
 }
 
-export default createBuilder(buildApplication);
+const builder: Builder<ApplicationBuilderOptions & json.JsonObject> =
```
clydin, line 263:
```suggestion
const builder: Builder<ApplicationBuilderOptions & import('@angular-devkit/core').json.JsonObject> =
```
I'd like to minimize the core package usage as much as possible here.  I think this is as far as possible right now. Assuming it works with the other changes.

clydin (approved):

packages/angular/build/src/builders/application/index.ts
alan-agius4: @devversion maybe we should switch to `Type` instead of `Interface` which would avoid the `& json.JsonObject`.

devversion: It seems to be used throughout the repository, within the same builders. Are we good keeping as is and just leaving this to a general cleanup? (seems more work, even with Alan's idea; which seems to work; I'm not sure of the consequences and there would still be quite some cleanup work then)

clydin: I'm fine with some TODOs here to cleanup later in the migration process.

devversion: Discussed offline. TODO not necessary, but we'll track this in a separate issue as cleanup.

Issue #28834, opened 2024-11-11
`ng update` breaks on "private" packages


SanderElias:
### Command

update

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

I'm unsure if this is a regression.

### Minimal Reproduction

To reproduce:
```bash
# use an "old" version to get started
npx @angular/cli@18.1 new DemoPackageIssue
cd DemoPackageIssue
# add a package from the deno std library (not really private..)
npx jsr add @std/toml
# try to upgrade:
npx ng update
```

### Exception or Error

```text
Using package manager: npm
Collecting installed dependencies...
Found 28 dependencies.
✖ Migration failed: 404 Not Found - GET https://registry.npmjs.org/@std%2ftoml - Not found
  See "/tmp/ng-f4tUqx/angular-errors.log" for further details.
```

### Your Environment

```text
I didn't test it with anything older as 18.0, but I suspect _all_ versions have this issue. 
At least 19-rc.1 still has the problem
```

eluchsinger:
The issue also affects us. Trying to upgrade from v19 to v20, the migration fails with a 404 not found, because it tries to locate our local package in the npmjs registry. Not sure what use case there is for this, but a way to skip this check would be nice.

We're using the [turborepo way](https://turborepo.com/docs/crafting-your-repository/managing-dependencies) to link local packges.

> Migration failed: 404 Not Found - GET https://registry.npmjs.org/@repo%2ftypescript-config - Not found

tsteuwer-accesso:
This is also affecting us. We use something along the lines of this in our .npmrc:

```
registry=https://{ouraccountid}.d.codeartifact.us-east-1.amazonaws.com/npm/npm-registry/
//{ouraccountid}.codeartifact.us-east-1.amazonaws.com/npm/npm-registry/:_authToken=${codeartifacttoken}
```

mumenthalers:
affected as well.

passee:
this issue was opened more than a year ago and I dont see any progress on it.
whats the problem here and why does ng update for example not acknowledge a .npmrc file?

Edit:
the issue was, that i was in a wrong directory. while `npm i` still worked because it seems like to check the parent folder too, `ng update` didnt see the npmrc file

john8329:
`ng update` never worked when there's packages in private repositories, namespaced or not. .npmrc is ignored. Just remove the packages as a workaround and re-add them later.

ChristofFritz:
[...]

PR #29358, opened 2025-01-15
fix(@angular/cli): handles not found packages for ng-update

allout58:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #28834 

## What is the new behavior?

Packages in private or unknown repositories that cannot be found are skipped when trying to run `ng update`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/cli/src/commands/update/schematic/index_spec.ts
+++ b/packages/angular/cli/src/commands/update/schematic/index_spec.ts
@@ -71,6 +71,28 @@ describe('@schematics/update', () => {
     );
   }, 45000);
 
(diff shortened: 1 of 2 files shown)


alan-agius4:
Hi @allout58, thanks for the update! Unfortunately, this fix isn't correct, as it interferes with real cases where the error needs to be propagated. In fact, some of our E2E tests are failing because of this change.

Instead of using the package name, the value should be parsed and used.

PR #29368, opened 2025-01-16
build: replace hardcoded Angular and ng-packagr peer dependencies with Bazel stamping

alan-agius4:
Refactor the build configuration to use Bazel stamping for Angular and ng-packagr dependencies, eliminating the need for hardcoded peer dependencies.

### DIFF
--- a/constants.bzl
+++ b/constants.bzl
@@ -3,6 +3,11 @@ RELEASE_ENGINES_NODE = "^18.19.1 || ^20.11.1 || >=22.0.0"
 RELEASE_ENGINES_NPM = "^6.11.0 || ^7.5.6 || >=8.0.0"
 RELEASE_ENGINES_YARN = ">= 1.13.0"
 
+NG_PACKAGR_VERSION = "^19.1.0-next.0"
+ANGULAR_FW_VERSION = "^19.1.0-next.0"
+ANGULAR_FW_PEER_DEP = "^19.0.0 || ^19.1.0-next.0"
+NG_PACKAGR_PEER_DEP = "^19.0.0 || ^19.1.0-next.0"
+
 SNAPSHOT_REPOS = {
     "@angular/cli": "angular/cli-builds",
     "@angular/pwa": "angular/angular-pwa-builds",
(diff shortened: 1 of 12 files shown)


packages/angular_devkit/build_angular/package.json
```diff
@@ -69,17 +69,17 @@
...
   },
   "peerDependencies": {
-    "@angular/compiler-cli": "^19.0.0 || ^19.1.0-next.0",
-    "@angular/localize": "^19.0.0 || ^19.1.0-next.0",
-    "@angular/platform-server": "^19.0.0 || ^19.1.0-next.0",
-    "@angular/service-worker": "^19.0.0 || ^19.1.0-next.0",
+    "@angular/compiler-cli": "0.0.0-ANGULAR-PEER-DEP",
+    "@angular/localize": "0.0.0-ANGULAR-PEER-DEP",
+    "@angular/platform-server": "0.0.0-ANGULAR-PEER-DEP",
+    "@angular/service-worker": "0.0.0-ANGULAR-PEER-DEP",
```
devversion, line 75: This looks great, but might be worth seeing how this plays out in the pnpm workspace. No idea yet.

alan-agius4: I think since this is just setting peer deps we should be good.

constants.bzl
```diff
@@ -3,6 +3,11 @@ RELEASE_ENGINES_NODE = "^18.19.1 || ^20.11.1 || >=22.0.0"
 RELEASE_ENGINES_NPM = "^6.11.0 || ^7.5.6 || >=8.0.0"
 RELEASE_ENGINES_YARN = ">= 1.13.0"
 
+NG_PACKAGR_VERSION = "^19.1.0-next.0"
+ANGULAR_VERSION = "^19.1.0-next.0"
```
dgp1130, line 7: Question: What exactly does `ANGULAR_VERSION` represent? The `@angular/core` / FW version used for direct dependencies? I wonder if a more specific name like `ANGULAR_FW_VERSION` might be helpful to communicate its purpose.

dgp1130 (approved):
This is great, thanks @alan-agius4! I've wanted something like this for a while now, glad to see it's possible with `@aspect_rules_js`!

PR #29371, opened 2025-01-16
Migrate more jasmine tests

devversion:
See individual commits


alan-agius4 (approved):

alan-agius4:
@devversion I changed the target to `minor` as it does not apply cleanly due to the change in the aspect lock file.

alan-agius4 (approved):

PR #29373, opened 2025-01-16
fix(@angular/build): perform incremental background file updates with component updates

clydin:
When HMR is enabled, a change to a lazy route component template may alter the names of lazy chunks. These lazy chunks must be updated within the development server so that any later non-HMR based update will have access to these chunks. To support this, the incremental results from the build process can now emit background file updates that will provide output file changes without triggering client updates.

### DIFF
--- a/packages/angular/build/src/builders/application/build-action.ts
+++ b/packages/angular/build/src/builders/application/build-action.ts
@@ -264,6 +264,8 @@ function* emitOutputResults(
     const incrementalResult: IncrementalResult = {
       kind: ResultKind.Incremental,
       warnings: warnings as ResultMessage[],
+      // These files need to be updated in the dev server but should not signal any updates
+      background: hasTemplateUpdates,
       added: [],
       removed: [],
       modified: [],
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):
LGTM, I am no longer able to replicate the problem with this change.


NIT: as a followup we should add a test that verifies when accessing `main.js` vite does not crash.

alan-agius4:
NIT: as a followup we should add a test that verifies when accessing `main.js` vite does not crash.

dgp1130:

p1n5u:
@alan-agius4 do you know when the next `19.1.x` release will be with this fix? I am currently trying to upgrade my application from `19.0.x` to `19.1.3` but this bug prevents `ng serve` to work as soon as i use dynamic imports inside of my code (not only when lazy loading routes).

alan-agius4:
@p1n5u, this update was released last week. I believe you may be encountering a different issue.

p1n5u:
@alan-agius4 you are right. sorry for the noise. will try to create a minimal reproduction example and open an issue.

PR #29374, opened 2025-01-16
build: exclude `pnpm` and `dependenciesMeta` fields from published NPM packages

alan-agius4:
Currently, the `pnpm` and `dependenciesMeta` fields from the root `package.json` are inadvertently copied to the `package.json` of published packages.

For example, see: https://github.com/angular/angular-devkit-schematics-builds/blob/dc4d5e0d523e3449963fa8e6efa32aee3e194e0f/package.json#L42-L52


devversion (approved):

alan-agius4:

PR #29387, opened 2025-01-17
fix(@angular/build): reset component updates on dev-server index request

clydin:
An index request from a client browser indicates a full page reload of the application should occur. In this case, the latest full output of the application should be sent to the client which would contain all separate component and incremental updates that have occurred since the last full update from the build system. To provide for this request, the individual component updates that were previously queued are now cleared to avoid reprocessing already present updates within the application output files.


alan-agius4 (approved):

clydin:

PR #29402, opened 2025-01-19
fix(@angular/build): allow asset changes to reload page on incremental updates

clydin:
Changes to asset files were previously not considered when determining if an incremental build result could be background update and not cause a page reload. Asset modifications will not cause a page reload in incremental update cases.
Additionally, a superfluous component update result will no longer be emitted in this case. The required page reload will account for the component updates.

Review Note: Enable hide whitespace for a more readable diff


alan-agius4 (approved):

alan-agius4:

PR #29405, opened 2025-01-20
fix(@angular/build): perform full reload for templates with `$localize` usage

clydin:
Component template updates that contain `$localize` usage currently may result in incorrect updates at runtime. To avoid this problem, any template update module that contains `$localize` will now perform a full rebuild. Support for component updates of these templates will be added in a future release.


crisbeto (approved):

alan-agius4 (approved):

alan-agius4:

PR #29422, opened 2025-01-20
fix(@angular/build): only issue invalid i18n config error for duplicate `subPaths` with inlined locales

### LINKED ISSUE #29396: Do Not Deprecate baseHref in i18n Configuration

localize

### Description:
The recent deprecation of the `baseHref` field in the `i18n.sourceLocale` configuration causes significant issues for projects that rely on a **shared base URL** for multiple localized builds.

While the `subPath` feature is a welcome addition for structuring locale builds, it is limiting for developers who:

1. Want locale-specific builds in subdirectories (`subPath`) without altering the `<base>` tag in `index.html`.
2. Depend on a static `baseHref` for all builds due to server or application constraints.

### Current Behavior:
With `subPath` enabled:

Angular modifies the `<base href>` tag in `index.html` to include the locale-specific subpath (e.g., `/zh-hans/`), and there is no built-in way to override this behavior.
Without `baseHref`:

Developers are forced to implement error-prone and maintenance-heavy post-build scripts to reset the `<base>` tag to their desired value.
### Expected Behavior:
Projects with `subPath` configured should still have the option to:

Use `baseHref` as a static value across all locale builds.
Separate the functionality of `subPath` (output directory) from the `<base>` tag in `index.html`.

### Proposed solution

Instead of fully deprecating `baseHref`, it could coexist with `subPath` to give developers more flexibility. Suggested behavior:

1. `baseHref` **takes precedence if defined**. This ensures that:
    - The `<base>` tag in `index.html` uses `baseHref` for all builds.
    - subPath is only used for structuring the output directory, not modifying `index.html`.
 
2. If `baseHref` is not specified, the current `subPath` behavior remains unchanged.

### Alternatives considered

Developers are forced to implement error-prone and maintenance-heavy post-build scripts to reset the <base> tag to their desired value.

### LINKED ISSUE #29398: Invalid i18n configuration: Locales 'ru' and 'en' cannot have the same subPath: ''.
### Command

build

### Is this a regression?

_No response_

### Description

Hey.
With v19.1 it's not possible to build with this config:

```json
"sourceLocale": {
      "code": "ru",
      "subPath": ""
    },
    "locales": {
      "en": {
        "translation": "apps/frontend/src/locale/messages.en.xlf",
        "subPath": ""
      }
    },
"configurations": {
        "production-en": {
          "localize": ["en"]
        },
        "production": {
          "localize": ["ru"],
}
```

How about moving non-unique subPath check inside build configuration, but not global?

This should fail:

```json
"configurations": {
 "production": {
          "localize": ["ru", "en"],
}
}
```

This should not:

```json
"configurations": {
 "production": {
          "localize": ["ru"],
}
}
```

### Minimal Reproduction

ng build

### Exception or Error

```text

```

### Your Environment

```text
v19.1
```

### Anything else relevant?

_No response_

alan-agius4:
The i18n configuration validation was incorrectly flagging errors for identical `subPaths` when both locales were not inlined within the same build. This was due to the validation not properly accounting for the inlining of locales.

This commit fixes this issue by ensuring that the validation only checks for duplicate `subPaths` when the locales are inlined.

Closes #29398

### DIFF
--- a/packages/angular/build/src/utils/i18n-options.ts
+++ b/packages/angular/build/src/utils/i18n-options.ts
@@ -193,8 +193,24 @@ export function createI18nOptions(
     }
   }
 
-  // Check that subPaths are unique.
-  const localesData = Object.entries(i18n.locales);
+  if (inline === true) {
+    i18n.inlineLocales.add(i18n.sourceLocale);
+    Object.keys(i18n.locales).forEach((locale) => i18n.inlineLocales.add(locale));
+  } else if (inline) {
+    for (const locale of inline) {
+      if (!i18n.locales[locale] && i18n.sourceLocale !== locale) {
+        throw new Error(`Requested locale '${locale}' is not defined for the project.`);
+      }
+
+      i18n.inlineLocales.add(locale);
+    }
+  }
+
+  // Check that subPaths are unique only the locales that we are inlining.
+  const localesData = Object.entries(i18n.locales).filter(([locale]) =>
+    i18n.inlineLocales.has(locale),
+  );
+
   for (let i = 0; i < localesData.length; i++) {
     const [localeA, { subPath: subPathA }] = localesData[i];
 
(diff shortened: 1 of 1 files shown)


zip-fa:
> Although route matchers are not supported with the SSR Routing API, they still function with the legacy `CommonEngine`. Therefore, no error should be issued in this case.
> 
> Closes #29396 and closes #29398

my 5c: commit description belongs to another commit which was already merged

clydin (approved):

alan-agius4:

Issue #29111, opened 2024-12-11
Support for Multi-Domain Locale Detection in Angular SSR


alan-agius4:
### Command

add

### Description

Currently, Angular SSR does not provide built-in functionality to detect the domain in a multi-domain setup and serve the appropriate localized version of the application. This feature would allow Angular SSR to dynamically render pages with the correct locale based on the domain.

**Use Case**:
   - Applications serving different localized versions (e.g., `example.com`, `fr.example.com`, `de.example.com`) need SSR to serve the appropriate locale.  
   - Example:  
     - `example.com` → English content  
     - `fr.example.com` → French content  
     - `de.example.com` → German content  

### Describe the solution you'd like

Allow SSR to use the detected domain to select the corresponding locale configuration. Which simplifies multi-language deployment without relying on path-based locales (e.g., `/en`, `/fr`).

jnizet:
Just my two cents: I had to do that for a backend (not using Angular SSR), and I found it rather elegant to do it by relying on the Accept-Language header. 

The reverse-proxy in front of the backend would be configured to set the appropriate Accept-Language header value based on the domain. This limits the configuration that needs to be done on the backend side. 

And relying on the Accept-Language header could also be a good thing for simpler deployments where there is a single URL to serve the app, and where the Accept-Language header coming from the browser is used to decide which locale to use.

autowp:
@jnizet but how to remove locale from path? AppEngine still redirects to /en-GB

I think the only current default pattern of locale detection must be a part of framework, but framework may provide interface to override ServerApp resolving and request handling (e.g. redirects) strategy

alan-agius4:
@autowp that could be handled by a reverse proxy as well.

Example NGINX has [`proxy_redirect`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect).

**Note:** I haven’t personally tried this to validate that this approach would for once the CSR app takes over.

autowp:
> [@autowp](https://github.com/autowp) that could be handled by a reverse proxy as well.
> 
> Example NGINX has [`proxy_redirect`](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect).
> 
> **Note:** I haven’t personally tried this to validate that this approach would for once the CSR app takes over.

Was tried to workaround with intermediate request processing using [vhost](https://www.npmjs.com/package/vhost): add `/{locale}/...` prefix and `Accept-Language` header 

```typescript
[...]

PR #29429, opened 2025-01-21
fix(@angular/build): replace deprecation of `i18n.baseHref` with a warning

### LINKED ISSUE #29396: Do Not Deprecate baseHref in i18n Configuration

alan-agius4:
In certain scenarios, users build applications with the same `baseHref` when using i18n, primarily for deploying localized applications across multiple domains. To address this, we are removing the deprecation of `i18n.baseHref` and will revisit potential options as part of https://github.com/angular/angular-cli/issues/29111

Instead of deprecating `i18n.baseHref`, we now issue a warning when it is used with SSR, as this may lead to undefined behavior.

Closes #29396

### DIFF
--- a/packages/angular/build/src/builders/application/options.ts
+++ b/packages/angular/build/src/builders/application/options.ts
@@ -174,7 +174,7 @@ export async function normalizeOptions(
   const i18nOptions: I18nOptions & {
     duplicateTranslationBehavior?: I18NTranslation;
     missingTranslationBehavior?: I18NTranslation;
-  } = createI18nOptions(projectMetadata, options.localize, context.logger);
+  } = createI18nOptions(projectMetadata, options.localize, context.logger, !!options.ssr);
   i18nOptions.duplicateTranslationBehavior = options.i18nDuplicateTranslation;
   i18nOptions.missingTranslationBehavior = options.i18nMissingTranslation;
   if (options.forceI18nFlatOutput) {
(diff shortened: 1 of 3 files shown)


clydin (approved):

alan-agius4:

PR #29431, opened 2025-01-21
refactor: reduce dependencies on `@angular-devkit/core`

alan-agius4:
By converting schemas from TypeScript `interfaces` to `types`, we can minimize the reliance on `json.JsonObject`. This approach avoids the error "Type 'Schema' does not satisfy the constraint 'JsonObject'" caused by the missing index signature for type 'string' in 'Schema'.


alan-agius4:

PR #29433, opened 2025-01-21
Migrate more tests to `rules_js`

devversion:
This is likely the second last PR to migrate tests to `rules_js`. One left after this one.


tsconfig.json
```diff
@@ -33,7 +33,6 @@
     "dist-schema/**",
     "**/node_modules/**/*",
     "**/third_party/**/*",
-    "packages/angular_devkit/schematics_cli/schematic/files/**/*",
-    "packages/angular_devkit/*/test/**/*"
```
alan-agius4, line 37: Why was the last entry removed? It was meant to exclude the `test` project. Also, I noticed that `modules/testing/builder/projects/**/*` is missing, which would otherwise result in linting the `test` project assets.

devversion: It caused `ts_project` to not compile tests inside `test/`. Are you saying this was just here for linting?

alan-agius4: The `ts_project` should exclude tests in the `test/` directory from compilation, as they are "assets" for the actual tests.

devversion: Done. please check again

PR #29437, opened 2025-01-21
refactor(@angular/build): allow browser client to log console message with dev-server

clydin:
The development server now supports a WebSocket event named `angular:log`. This event allows the browser client to send log messages back to the development server. Currently this is unused by the client and Angular runtime. But is intended to be used in the future for such cases as propagating error messages back to the development server console. This event is considered internal and should not be relied upon by external code.


alan-agius4:

PR #29442, opened 2025-01-22
refactor(@angular/ssr): update `getPathSegments` to use `stripTrailingSlash` instead of `filter(Boolean)`

alan-agius4:
This change modifies the `getPathSegments` function to use `stripTrailingSlash`, providing a more consistent and reliable way to handle trailing slashes in paths. This update also resolves issues causing CI failures.


crisbeto (approved):

alan-agius4:

PR #29461, opened 2025-01-23
fix(@angular/ssr): enhance dynamic route matching for better performance and accuracy

### LINKED ISSUE #29452: Routes beginning with Dynamic Parameters break Angular Prerendering and SSR
### Command

build, serve

### Is this a regression?

_No response_

### Description

Angular provides examples on how to use prerendering for parametrized routes like `post/:id`. However, in our project, we have routes with dynamic parameters at the beginning of the route path, such as `:seoAlias/:category/:offerId`.

This specific route structure causes issues in `Client`, `Server`, and `Prerender` modes, breaking other routes when serving the app in development mode or building the application.

### Minimal Reproduction

We have created a simple reproduction repository:
[https://github.com/robertsine/car-shop](https://github.com/robertsine/car-shop)

Inside the file `app.routes.server.ts` there are several route variants. By uncommenting them, you can reproduce the issue. Notably, only `offers/:slug` works well. Other server routes are breaking entire app, when serving app in dev mode or serving build.

We tried to use the `**` path as a _hack_ to prerender our parameterized routes. However, in development mode, it does not work. In build mode, it creates prerendered pages. For note, we must ensure unique parameters across the entire route configuration to avoid hitting several paths with same parameter name. However, it generates fallback content inside the index files.

### Exception or Error

```text
Error: Error(s) occurred while extracting routes:
- The '' server route does not match any routes defined in the Angular routing configuration (typically provided as a part of the 'provideRouter' call). Please make sure that the mentioned server route is present in the Angular routing configuration.
```

### Your Environment

```text
Angular CLI: 19.0.6
Node: 22.13.1
Package Manager: npm 9.5.1
OS: win32 x64

Angular: 19.0.5
... router

@angular/ssr                    19.0.6
```

### Anything else relevant?

_No response_

alan-agius4:
Updated route matching logic to prioritize closest matches, improving the accuracy of dynamic route resolution. Also we optimized performance by eliminating unnecessary recursive checks, reducing overhead during route matching.

Closes #29452

### DIFF
--- a/packages/angular/ssr/src/routes/ng-routes.ts
+++ b/packages/angular/ssr/src/routes/ng-routes.ts
@@ -199,6 +199,7 @@ async function* traverseRoutesConfig(options: {
       if (metadata.renderMode === RenderMode.Prerender) {
         // Handle SSG routes
         yield* handleSSGRoute(
+          serverConfigRouteTree,
           typeof redirectTo === 'string' ? redirectTo : undefined,
           metadata,
           parentInjector,
@@ -301,6 +302,7 @@ function appendPreloadToMetadata(
  * Handles SSG (Static Site Generation) routes by invoking `getPrerenderParams` and yielding
  * all parameterized paths, returning any errors encountered.
  *
+ * @param serverConfigRouteTree - The tree representing the server's routing setup.
  * @param redirectTo - Optional path to redirect to, if specified.
  * @param metadata - The metadata associated with the route tree node.
  * @param parentInjector - The dependency injection container for the parent route.
@@ -309,6 +311,7 @@ function appendPreloadToMetadata(
  * @returns An async iterable iterator that yields route tree node metadata for each SSG path or errors.
  */
 async function* handleSSGRoute(
+  serverConfigRouteTree: RouteTree<ServerConfigRouteTreeAdditionalMetadata> | undefined,
   redirectTo: string | undefined,
   metadata: ServerConfigRouteTreeNodeMetadata,
   parentInjector: Injector,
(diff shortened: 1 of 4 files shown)


packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -458,6 +458,13 @@ function buildServerConfigRouteTree({ routes, appShellRoute }: ServerRoutesConfi
       continue;
     }
 
+    if (path.includes('**') && 'getPrerenderParams' in metadata) {
+      errors.push(
+        `Invalid '${path}' route configuration: 'getPrerenderParams' cannot be used with a '**' route.`,
```
dgp1130, line 463: Question: Is this true for a single `*` too? Should we ban that as well?

packages/angular/ssr/src/routes/route-tree.ts
```diff
@@ -128,29 +114,28 @@ export class RouteTree<AdditionalMetadata extends Record<string, unknown> = {}>
...
 
-    for (const segment of segments) {
-      // Replace parameterized segments (e.g., :id) with a wildcard (*) for matching
-      const normalizedSegment = segment[0] === ':' ? '*' : segment;
-      let childNode = node.children.get(normalizedSegment);
+    for (let index = 0; index < segments.length; index++) {
+      let segment = segments[index];
+      if (segment[0] === ':') {
+        // Replace parameterized segments (e.g., :id) with a wildcard (*) for matching
+        segments[index] = segment = '*';
```
dgp1130, line 122: Suggestion: Personally I find this kind of mutation hard to follow and the double assignment makes it particularly tricky to notice that `segment` is updated too.

Could we do this transform separately first? It would also allow us to drop the manual iteration with `index`.

```typescript
const segs = segments.map((segment) => segment.startsWith(':') ? '*' : segment);
for (const segment of segs) {
  // ...
}
```

(Should use a better name than `segs`, but hopefully that gets the point across.)

dgp1130 (approved):
It seems like there's a few small breaking changes here. I know this feature is still dev preview, so we can probably ship this and I'm happy to defer to your judgement. I just want to make sure we're thinking about the impact here and making that decision intentionally.

packages/angular/ssr/src/routes/ng-routes.ts
alan-agius4: Updated

alan-agius4:

PR #29471, opened 2025-01-24
fix(@angular/build): update vite to version 6.0.11

clydin:
Version update from 6.0.7 to address advisory https://github.com/vitejs/vite/security/advisories/GHSA-vg6x-rcgg-rjx6

Vite version 6.0.9+, which is now used by the Angular CLI with the `application`/`browser-esbuild` builders, contains a potentially breaking change for some development setups. Examples of such setups include those that use reverse proxies or custom host names during development. The change within a patch release was made by Vite to address a security vulnerability. For projects that directly access the development server via `localhost`, no changes should be needed. However, some development setups may now need to adjust the `allowedHosts` development server option. This option can include an array of host names that are allowed to communicate with the development server. The option sets the corresponding Vite option within the Angular CLI. For more information on the option and its specific behavior, please see the Vite documentation located here:
https://vite.dev/config/server-options.html#server-allowedhosts

The following is an example of the configuration option allowing `example.com`:
```
"serve": {
      "options": {
        "allowedHosts": ["example.com"]
      },
```

### DIFF
--- a/packages/angular/build/src/builders/dev-server/schema.json
+++ b/packages/angular/build/src/builders/dev-server/schema.json
@@ -36,6 +36,23 @@
       "type": "string",
       "description": "SSL certificate to use for serving HTTPS."
     },
+    "allowedHosts": {
+      "description": "The hosts that can access the development server. This option sets the Vite option of the same name. For further details: https://vite.dev/config/server-options.html#server-allowedhosts",
+      "default": [],
+      "oneOf": [
+        {
+          "type": "array",
+          "description": "List of hosts that are allowed to access the development server.",
+          "items": {
+            "type": "string"
+          }
+        },
+        {
+          "type": "boolean",
+          "description": "Indicates that all hosts are allowed. This is not recommended and a security risk."
+        }
+      ]
+    },
     "headers": {
       "type": "object",
       "description": "Custom HTTP headers to be added to all responses.",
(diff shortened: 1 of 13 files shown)


packages/angular/build/src/builders/dev-server/schema.json
```diff
@@ -36,6 +36,23 @@
...
+      "oneOf": [
+        {
+          "type": "array",
+          "description": "List of hosts that are allowed to access the development server.",
+          "items": {
+            "type": "string"
+          }
+        },
+        {
+          "type": "boolean",
```
alan-agius4, line 51: From what I am seeing `false` is not an intended value. Thus, consider using `const`.

```suggestion
          "const": true,
          "type": "boolean",
```

packages/angular/build/src/builders/dev-server/tests/execute-fetch.ts
```diff
@@ -41,3 +42,49 @@ export async function executeOnceAndFetch<T>(
...
+ * @returns
+ */
+export async function executeOnceAndGet<T>(
+  harness: BuilderHarness<T>,
+  url: string,
+  options?: Partial<BuilderHarnessExecutionOptions> & { request?: RequestOptions },
+): Promise<BuilderHarnessExecutionResult & { response?: IncomingMessage; content?: string }> {
+  return lastValueFrom(
+    harness.execute().pipe(
+      timeout(30000),
```
alan-agius4, line 65: Nit, easier to parse :) 
```suggestion
      timeout(30_000),
```

alan-agius4 (approved):
LGTM, just a couple of nits.

PR #29473, opened 2025-01-24
fix(@angular/ssr): rename `provideServerRoutesConfig` to `provideServerRouting`

alan-agius4:
This commit renames `provideServerRoutesConfig` to `provideServerRouting` and updates the second parameter to use the `ServerRoutes` features.

This change improves alignment with the framework's API conventions and the way features are integrated.

### Example Usage:
Before:
```typescript
provideServerRoutesConfig(serverRoutes, { appShellRoute: 'shell' })
```

After:
```typescript
provideServerRouting(serverRoutes, withAppShell(AppShellComponent))
```

### DIFF
--- a/packages/angular/ssr/src/routes/route-config.ts
+++ b/packages/angular/ssr/src/routes/route-config.ts
@@ -6,11 +6,43 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import { EnvironmentProviders, InjectionToken, makeEnvironmentProviders } from '@angular/core';
+import {
+  EnvironmentProviders,
+  InjectionToken,
+  Provider,
+  Type,
+  makeEnvironmentProviders,
+} from '@angular/core';
+import { type DefaultExport, ROUTES, type Route } from '@angular/router';
+
+/**
+ * The internal path used for the app shell route.
+ * @internal
+ */
+const APP_SHELL_ROUTE = 'ng-app-shell';
+
+/**
+ * Identifies a particular kind of `ServerRoutesFeatureKind`.
+ * @see {@link ServerRoutesFeature}
+ * @developerPreview
+ */
+enum ServerRoutesFeatureKind {
+  AppShell,
+}
+
+/**
+ * Helper type to represent a server routes feature.
+ * @see {@link ServerRoutesFeatureKind}
+ * @developerPreview
+ */
+interface ServerRoutesFeature<FeatureKind extends ServerRoutesFeatureKind> {
+  ɵkind: FeatureKind;
+  ɵproviders: Provider[];
+}
 
 /**
  * Different rendering modes for server routes.
- * @see {@link provideServerRoutesConfig}
+ * @see {@link provideServerRouting}
  * @see {@link ServerRoute}
  * @developerPreview
  */
@@ -148,7 +180,7 @@ export interface ServerRouteServer extends ServerRouteCommon {
 
 /**
  * Server route configuration.
- * @see {@link provideServerRoutesConfig}
+ * @see {@link provideServerRouting}
  * @developerPreview
  */
 export type ServerRoute =
@@ -163,8 +195,9 @@ export type ServerRoute =
  * This interface defines the optional settings available for configuring server routes
  * in the server-side environment, such as specifying a path to the app shell route.
  *
- * @see {@link provideServerRoutesConfig}
- * @developerPreview
+ *
+ * @see {@link provideServerRouting}
+ * @deprecated use `provideServerRouting`. This will be removed in version 20.
  */
 
 export interface ServerRoutesConfigOptions {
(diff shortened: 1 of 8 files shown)


packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -223,3 +262,78 @@ export function provideServerRoutesConfig(
...
+ * @see {@link ServerRoute}
+ * @see {@link withAppShell}
+ * @developerPreview
+ */
+export function provideServerRouting(
+  routes: ServerRoute[],
+  ...features: ServerRoutesFeature<ServerRoutesFeatureKind>[]
+): EnvironmentProviders {
+  const config: ServerRoutesConfig = { routes };
+  const hasAppShell = features.some((f) => f.ɵkind === ServerRoutesFeatureKind.AppShell);
```
dgp1130, line 285: Question: Do we need to check the kind here or could we just look for a route containing the app shell route?

Related question: Does framework typically implement this `withFeature` pattern using an object with `kind` to disambiguate the relevant feature? I always assumed it just included a bunch of providers and let the runtime optionally inject and call those providers if they are present. Basically, I'm wondering if there's a path to make `withAppShell` just return a list of providers rather than needing a special `kind` enum.

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -223,3 +262,78 @@ export function provideServerRoutesConfig(
...
+ * patterns, such as service workers, to deliver a seamless user experience.
+ *
+ * @param component The Angular component to render for the app shell route.
+ * @returns A server routes feature configuration for the app shell.
+ *
+ * @see {@link provideServerRouting}
+ * @see {@link https://angular.dev/ecosystem/service-workers/app-shell | App shell pattern on Angular.dev}
+ */
+export function withAppShell(
+  component: Type<unknown>,
```
dgp1130, line 322: Question: Is there a use case for lazy loading the app shell? Is it worth supporting `() => Type<unknown>` or does the app shell always need to be synchronously loaded?

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -223,3 +262,78 @@ export function provideServerRoutesConfig(
...
+  return {
+    ɵkind: ServerRoutesFeatureKind.AppShell,
+    ɵproviders: [
+      {
+        provide: ROUTES,
+        useValue: [
+          {
+            path: APP_SHELL_ROUTE,
+            component,
+          },
```
dgp1130, line 333: Question: Is there a use case for configuring any of these router options? What if I want to specify specific providers, title, etc.?

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: > Do we need to check the kind here or could we just look for a route containing the app shell route?

The route config will not contain the app-shell route, the route will be added during runtime.

> Does framework typically implement this withFeature pattern using an object with kind to disambiguate the relevant feature

Yes, this is the pattern used in the framework https://github.com/angular/angular/blob/fd7ee47bf702029e481ad40362684f0a01ef2ec0/packages/router/src/provide_router.ts#L110-L123, https://github.com/angular/angular/blob/main/packages/common/http/src/provider.ts and all others. I think the goal here, is to actually make sure users to not add these providers directly in their config example `providers: [withFeature()]`.

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: We could add an additional parameter to `withAppShell` if needed, but I have not seen anyone providing additional options so far.

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: No, it can also be async.

packages/angular/ssr/src/routes/route-config.ts
dgp1130: I was thinking about this more from the perspective of the abstraction we're creating. By writing `withAppShell(AppShellComponent)` the "app shell abstraction" is effectively a component, not a route.

By contrast, if we did something like:

```typescript
withAppShell({
  component: AppShellComponent,
  providers: [ /* ... */ ],
})
```

This would lift the abstraction to be a route, not a component (maybe `path` is omitted because we always use the same static path). Of course there might be other options which don't make sense to set (`canActivate` seems weird in this context).

If we think a component is a better abstraction than a route here, then your approach LGTM, we can expose additional random options as needed. I mainly want to make sure we're considering the route abstraction and whether that's more appropriate or not in this context.

packages/angular/ssr/src/routes/route-config.ts
dgp1130: Interesting, I get we don't want users to write `providers: [withFeature]`. I was envisioning something like:

```typescript
interface Feature {
  providers: Provider[];
}
```

And then we'd just append the providers without worrying about a `kind`.

I'll defer to you whether the `kind` is worth keeping, seems like it was useful for the app router at least.

dgp1130 (approved):

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: The `kind` is helpful for filtering providers and performing additional actions based on its value.

alan-agius4:

PR #29491, opened 2025-01-27
refactor(@angular/ssr): drain node stream

alexfriesen:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

writeResponseToNodeResponse will stream an response without draining.

Issue Number: N/A

## What is the new behavior?

Node.js streams should be drained to ensure efficient and uninterrupted data flow. This PR adds draining logic to `writeResponseToNodeResponse`

- [ ] Yes
- [x] No

## Other information

I'm not sure if all checks will pass. Bazel is a weird alien to me

### DIFF
--- a/packages/angular/ssr/node/src/response.ts
+++ b/packages/angular/ssr/node/src/response.ts
@@ -71,7 +71,10 @@ export async function writeResponseToNodeResponse(
         break;
       }
 
-      (destination as ServerResponse).write(value);
+      const canContinue = (destination as ServerResponse).write(value);
+      if (!canContinue) {
+        await new Promise<void>((resolve) => destination.once('drain', resolve));
+      }
     }
   } catch {
     destination.end('Internal server error.');


packages/angular/ssr/node/src/response.ts
```diff
@@ -52,27 +53,49 @@ export async function writeResponseToNodeResponse(
...
-      if (done) {
-        destination.end();
-        break;
+      if (error) {
+        destination.destroy(error);
       }
+    }
 
-      (destination as ServerResponse).write(value);
+    function handleStream({
```
alan-agius4: The new logic seems a bit hard to follow. In my opinion, it could be simplified significantly by using `async/await` instead of relying on callbacks. Additionally, the use of `handleStream` could be avoided, which would make the flow cleaner and easier to understand.


```ts
destination.on('close', handleClose);
destination.on('error', handleError);

try {
  while (true) {
    const { done, value } = await reader.read();
    if (done) {
      destination.end();
      break;
    }

    if (!(destination as ServerResponse).write(value)) {
      await new Promise<void>((resolve) => destination.once('drain', resolve));
    }
  }
} catch (error) {
  handleError(error);
} finally {
  handleClose();
}
```

packages/angular/ssr/node/src/response.ts
```diff
@@ -52,27 +53,49 @@ export async function writeResponseToNodeResponse(
...
     }
+
+    destination.on('close', cancelStream);
+    destination.on('error', cancelStream);
+    reader.read().then(handleStream, cancelStream);
+
+    // Return a promise that resolves when the stream is closed
+    return reader.closed.finally(() => {
+      // cleanup listeners
+      (destination as ServerResponse).off('close', cancelStream);
```
alan-agius4, line 98: Type casting here is redundant.

alan-agius4 (changes requested):
Thanks for your contribution. 

- Please squash the commits.
- In the commit message body, can you please context of the change. This is important for future users that are checking the diff to see what certain code has been done etc..

alan-agius4 (approved):

alan-agius4:

PR #29510, opened 2025-01-28
refactor(@angular/build): allow component update invalidation from client

clydin:
If HMR is enabled, a component update has the potential to be unsupported at runtime or may cause an exception. While build time analysis attempts to verify that an update is possible, there could be cases that are as of yet unknown. For those cases, the runtime can now signal this information back to the development server which will clear the errant component update and trigger a full page reload. This action will be logged to the development server console along with an optional message from the client.

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -457,6 +457,41 @@ export async function* serveWithVite(
         }
       });
 
+      // Setup component HMR invalidation
+      // Invalidation occurs when the runtime cannot update a component
+      server.hot.on(
+        'angular:invalidate',
+        (data: { id: string; message?: string; error?: boolean }) => {
+          if (typeof data?.id !== 'string') {
+            context.logger.warn(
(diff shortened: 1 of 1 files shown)


packages/angular/build/src/builders/dev-server/vite-server.ts
```diff
@@ -457,6 +457,41 @@ export async function* serveWithVite(
...
+            );
+          } else {
+            context.logger.warn(
+              `Component update unsupported${data.message ? `: ${data.message}` : '.'}`,
+            );
+          }
+
+          server?.ws.send({
+            type: 'full-reload',
+            path: '*',
```
crisbeto, line 489: Does this mean that the same page will be reloaded or will it go back to the root?

crisbeto (approved):

packages/angular/build/src/builders/dev-server/vite-server.ts
alan-agius4: Same page.

clydin:

PR #29518, opened 2025-01-29
fix(@angular/ssr): redirect to locale pathname instead of full URL

### LINKED ISSUE #29514: Angular AppEngine i18 redirect includes full domain
### Command

build

### Is this a regression?

_No response_

### Description

While serving a i18n using AngularNodeAppEngine, the base path i18n redirect uses the full URL to do the 302 redirect to i18n subpaths.

https://github.com/angular/angular-cli/blob/main/packages/angular/ssr/src/app-engine.ts#L120

If the application is not served directly on the real domain (ex: using a cdn or cloud run), the redirect issued is redirecting to the bad domain.

```
browser : example.com  ----> cdn/reverse proxy  ---> app.cloud.run.domain
                                                                |
                                                           redirect 302 to
   browser            <----------------------------  app.cloud.run.domain/de/
       |                                                         
handle redirect       ------------------------------> app.cloud.run.domain/de/
```

Only redirecting to `/de/` path  instead of full url should be enough.

### Minimal Reproduction

Create a ssr i18n app 
Host it on google cloud run (url : `app.cloud.run.domain`)
Point to it using any cdn/reverse proxy (url: `example.com`)
Navigate to naked domain `example.com`
Browser is redirected to `app.cloud.run.domain/de`

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 19.1.4
Node: 22.6.0
Package Manager: npm 10.8.2
OS: linux x64

Angular: 19.1.3
... localize, platform-browser, platform-browser-dynamic
... platform-server, router, service-worker

@angular-devkit/architect       0.1901.4
@angular-devkit/build-angular   19.1.4
@angular-devkit/core            19.1.4
@angular-devkit/schematics      19.1.4
@angular/cli                    19.1.4
@angular/ssr                    19.1.4
@schematics/angular             19.1.4
```

### Anything else relevant?

_No response_

alan-agius4:
When redirecting to the preferred locale, the previous implementation used the full URL for the 302 redirect to i18n subpaths based on the user's preferred locale. This update ensures that the redirect now uses the locale-specific pathname instead of the full URL.

Closes #29514

### DIFF
--- a/packages/angular/ssr/src/app-engine.ts
+++ b/packages/angular/ssr/src/app-engine.ts
@@ -97,8 +97,8 @@ export class AngularAppEngine {
     const { basePath, supportedLocales } = this.manifest;
 
     // If the request is not for the base path, it's not our responsibility to handle it.
-    const url = new URL(request.url);
-    if (url.pathname !== basePath) {
+    const { pathname } = new URL(request.url);
+    if (pathname !== basePath) {
       return null;
     }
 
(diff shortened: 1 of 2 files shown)


devversion (approved):

alan-agius4:

Issue #29539, opened 2025-01-29
`AngularNodeAppEngine` has different unknown path routing behavior than `CommonEngine`


griest024:
### Command

serve

### Is this a regression?

- [x] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

18

### Description

When the routing configuration contains a nested wildcard route and a redirect target and an non-nested wildcard route after the nested route that redirects to the nested route (see below), `CommonEngine` will use the nested route while `AngularNodeAppEngine` will use the non-nested route.

```ts
@Component({selector: 'test', template: 'test'})
class TestComponent {}

@Component({selector: 'not-found', template: '404 not found'})
class NotFoundComponent {}

export const routes: Routes = [
  {
    path: '',
    children: [
      // AngularNodeAppEngine will end up following the redirect and rendering this route
      {
        path: 'not-found',
        component: NotFoundComponent
      },
      // CommonEngine will just simply render this route
      {
        path: '**',
        component: TestComponent
      },
    ]
  },
  // AngularNodeAppEngine chooses this route
  {
    path: '**',
    redirectTo: 'not-found'
  },
]
```

Is this intended behavior? It seems that the vite dev server follows the behavior of `AngularNodeAppEngine` and is blocking us from upgrading to v19.

Something odd is that `AngularNodeAppEngine` will render `/` as `test` but `/asdfasdf` as `404 not found`. `CommonEngine` renders `test` for both of these paths.

Some debugging found that https://github.com/angular/angular-cli/blob/249f9dc26caa4f3fe3cf9799588955cd27f51791/packages/angular/ssr/src/routes/ng-routes.ts#L606 is likely to blame. It flattens the route tree in a way that the nested wildcard route gets overridden by the non-nested one during the insert into the final route tree https://github.com/angular/angular-cli/blob/249f9dc26caa4f3fe3cf9799588955cd27f51791/packages/angular/ssr/src/routes/ng-routes.ts#L722

### Minimal Reproduction

`AngularNodeAppEngine` - https://github.com/griest024/ng19-server-routing
`CommonEngine` - https://github.com/griest024/ng19-server-routing/tree/common-engine

1. run `npx ng b && npm run serve:ssr:19-dev-server-routes`
2. navigate to `/asdfasdfasdf` or any other route that is not `/` or `/not-found`
3. See that the rendered component is different.

### Exception or Error

```text

```

### Your Environment

```text
_                      _                 ____ _     ___
[...]

PR #29543, opened 2025-01-30
fix(@angular/ssr): prioritize the first matching route over subsequent ones

alan-agius4:
Ensures that the SSR router gives precedence to the first matching route, addressing the issue where later conflicting routes.

This change prevents the incorrect prioritization of routes and ensures the intended route is matched first, aligning routing behavior.

Closes: #29539


dgp1130 (approved):

jkrems:

alfaproject:
Should we log a warning if there are multiple routes matched, maybe under a flag?

PR #29545, opened 2025-01-30
build: re-enable latest version checks during release

alan-agius4:
This commit restores the latest version checks, which are now fixed by using the stamped versions.

### DIFF
--- a/scripts/release-checks/dependency-ranges/latest-versions-check.mts
+++ b/scripts/release-checks/dependency-ranges/latest-versions-check.mts
@@ -6,30 +6,33 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import { readFile } from 'node:fs/promises';
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
 import semver from 'semver';
 
-export async function checkSchematicsAngularLatestVersion(
-  newVersion: semver.SemVer,
-): Promise<string[]> {
-  const { dependencies } = JSON.parse(
-    await readFile('./packages/schematics/angular/utility/latest-versions/package.json', 'utf-8'),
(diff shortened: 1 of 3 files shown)


scripts/release-checks/dependency-ranges/latest-versions-check.mts
```diff
@@ -6,30 +6,31 @@
...
-  newVersion: semver.SemVer,
-): Promise<string[]> {
-  const { dependencies } = JSON.parse(
-    await readFile('./packages/schematics/angular/utility/latest-versions/package.json', 'utf-8'),
-  );
+export function checkSchematicsAngularLatestVersion(newVersion: semver.SemVer): string[] {
+  // Root of the Angular CLI project.
+  const root = fileURLToPath(new URL('../../../', import.meta.url));
+  const rootRequire = createRequire(root);
+  const { latestVersions } = rootRequire('./dist/@schematics/angular/utility/latest-versions.js');
```
devversion, line 17: Is `dist/` guaranteed to be existent here? Seems like an assumption that is potentially brittle. Also the `createRequire` is not super great :(

alan-agius4: Yes, the NPM packages are always built in the `dist` https://github.com/angular/angular-cli/blob/c0d20e05c3470ff69576227d03cb913b1ab1337f/scripts/build-packages-dist.mts#L37


Regarding `createRequire`, I agree that using a `RegExp` could work. However, I lean toward this approach since it minimizes the risk of false positives or negatives.

devversion: Ideally we'd somehow pass the packages and their artifact paths to the pre=release checks, but that's not possible right now, so LGTM

PR #29550, opened 2025-01-31
refactor(@angular/ssr): simplify preload append logic in metadata

alan-agius4:
Replace `filter` and `map` with a `for...of` loop to improve readability in the preload append logic within metadata.


clydin (approved):

alan-agius4:

PR #29552, opened 2025-01-31
ci: configure codeql locally within the repo to allow for customization

josephperrott:
Moving to the configuration being in the repo allows us to specify which specific rules are run in analysis.


alan-agius4 (approved):

alan-agius4:

PR #29569, opened 2025-02-04
fix(@angular/build): avoid pre-transform errors with Vite pre-bundling

clydin:
Vite 6.0 change the option location of the `preTransformRequests` to the `dev` section of the Vite configuration. While the previous `server` section option of the same name is still present, it currently does not change behavior when configured.


alan-agius4 (approved):

alan-agius4:

PR #29570, opened 2025-02-04
fix(@angular/build): always provide Vite client helpers with development server

### LINKED ISSUE #29556: --live-reload=false results in error: The requested module '/@vite/client' does not provide an export named 'injectQuery'

Don't known / other

### Is this a regression?

Yes

### Description

When certain third-party dependencies are used in conjunction with `--live-reload=false`, the application fails to load in Angular v19:

```
Uncaught SyntaxError: The requested module '/@vite/client' does not provide an export named 'injectQuery' (at chunk-QRJ2NFPW.js?v=4363c128:1:10)
```

Removing `--live-reload=false` from the `ng serve` command results in the application working again. It also worked in Angular v18 (with `--live-reload=false`).

I have found this issue with the `@arcgis/core` dependency.

To replicate:

`npm i @arcgis/core`

Then in `app.component.ts` or `main.ts`, add an import for a file from the library (add a `console.log` to prevent dependency from being optimized away):

```
import FeatureLayer from '@arcgis/core/layers/FeatureLayer'

console.log(FeatureLayer)
```

Then run the app using:

`ng serve --live-reload=false --host=127.0.0.1`

Repro of this issue with `@arcgis/core` is available here:

https://stackblitz.com/edit/stackblitz-starters-vmdnvkkn?file=src%2Fmain.ts

Another user reported this error with the pdf.js library on this StackOverflow thread:

https://stackoverflow.com/questions/79391578/angular-19-and-vite-throwing-error-about-named-export-injectquery

StackBlitz demo for the pdfjs dependency (from above thread):

https://stackblitz.com/edit/node-c5u2pyx6?file=package.json

https://stackblitz.com/edit/stackblitz-starters-vmdnvkkn?file=src%2Fmain.ts

```true
Uncaught SyntaxError: The requested module '/@vite/client' does not provide an export named 'injectQuery' (at chunk-QMHRUMYA.js:1:10)
```

```true
Angular CLI: 19.1.5
Node: 22.13.1
Package Manager: npm 10.9.2
OS: win32 x64

Angular: 19.1.4

```

### Anything else?

Inspecting `@vite/client` file via dev tools shows that it is empty:

Image

clydin:
In addition to the WebSocket code, the Vite client module contains helper functions which may be injected into modules at request time. These helpers are required for certain behavior to function. Previously, when `--no-live-reload` was used, these helpers may not have been available which led to runtime errors. These runtime errors will no longer occur. However, the browser console will now log that the Vite client cannot connect to the development server WebSocket. This is expected in this case since live reload functionality was disabled and the server side is intentionally not available.

Closes #29556

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -854,7 +854,6 @@ export async function setupServer(
         outputFiles,
         templateUpdates,
         external: externalMetadata.explicitBrowser,
-        skipViteClient: serverOptions.liveReload === false && serverOptions.hmr === false,
       }),
     ],
     // Browser only optimizeDeps. (This does not run for SSR dependencies).


packages/angular/build/src/builders/dev-server/vite-server.ts
```diff
@@ -854,7 +854,6 @@ export async function setupServer(
         outputFiles,
```
alan-agius4, line 854: Can you also please remove from `skipViteClient` from packages/angular/build/src/tools/vite/plugins/angular-memory-plugin.ts?

clydin: adjusted the option name to better reflect its new behavior.

dgp1130:

PR #29582, opened 2025-02-06
refactor(@angular/build): also add `server.preTransformRequests`

alan-agius4:
https://github.com/angular/angular-cli/commit/5c1360179cec2f0fad6b2adb4a8e4d6930738976 moved `preTransformRequests` from the `server` to `dev` section. But vite, still uses the `server` section in such cases https://github.com/vitejs/vite/blob/bcdb51a1ac082f4e8ed6f820787d6745dfaa972d/packages/vite/src/node/server/index.ts#L673 and https://github.com/vitejs/vite/blob/bcdb51a1ac082f4e8ed6f820787d6745dfaa972d/packages/vite/src/node/server/middlewares/indexHtml.ts#L475

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -734,6 +734,7 @@ function updateResultRecord(
   }
 }
 
+// eslint-disable-next-line max-lines-per-function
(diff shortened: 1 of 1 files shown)


jkrems (approved):

alan-agius4:

PR #29598, opened 2025-02-10
fix(@angular/cli): correctly parse and resolve relative schematic collection names on Windows

### LINKED ISSUE #29559: [CLI][Schematics] Override Angular Schematics issue - Collection "D" cannot be resolved. - Windows
### Command

generate

### Is this a regression?

_No response_

### Description

If you try to override standard Angular Schematics with a custom one by the use of schematicCollections option in angular.json file like for instance:

`  "cli": {
    "analytics": false,
    "schematicCollections": [
      "./schematics",
      "@schematics/angular"
    ]
  }`

And create your custom schematics (index.ts):

`import {
  externalSchematic,
  Rule,
  SchematicContext,
  Tree
import { Schema as ComponentOptions } from '@schematics/angular/component/schema';

export function generateComponent(options: ComponentOptions): Rule {
  console.log('Custom schematics for generate component triggered...');
  return (_tree: Tree, _context: SchematicContext) => {
    return externalSchematic(
      "@schematics/angular",
      "component",
      options
    );
  };
}`

It will end up with an error message like this one:

> Collection "D" cannot be resolved.

Where D refers to the drive letter of the project folder location. 

### Minimal Reproduction

Please use the attached repo to reproduce the issue:
[schematics-test.zip](https://github.com/user-attachments/files/18633785/schematics-test.zip)

Run npm install and then try to generate a component using:
`ng g c`
command.

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 19.1.5
Node: 22.13.0
Package Manager: npm 8.6.0
OS: win32 x64

Angular: 19.1.4

```

### Anything else relevant?

The error is caused by parseSchematicInfo method defined in SchematicsCommandModule in file:
@angular\cli\src\command-builder\schematics-command-module.js
Modifying the function as follows:
` parseSchematicInfo(schematic) {
  if (schematic?.includes(':')) {
    const segments = schematic.split(':');
    const [collectionName, schematicName] = [segments.slice(0, -1).join(':'), segments.at(-1)];
    return [collectionName, schematicName];
  }
  return [undefined, schematic];
}`
Seems to fix the problem.

The issue is also present in v18.

alan-agius4:
Previously, the logic incorrectly extracted the drive letter as the collection name when the schematic collection was specified as relative on Windows. This fix ensures that relative paths are parsed and resolved correctly, preventing the drive letter from being mistakenly treated as the collection name.

Closes #29559

### DIFF
--- a/packages/angular/cli/src/command-builder/schematics-command-module.ts
+++ b/packages/angular/cli/src/command-builder/schematics-command-module.ts
@@ -13,7 +13,7 @@ import {
   FileSystemSchematicDescription,
   NodeWorkflow,
 } from '@angular-devkit/schematics/tools';
-import { relative, resolve } from 'path';
+import { relative } from 'path';
 import { Argv } from 'yargs';
 import { isPackageNameSafeForAnalytics } from '../analytics/analytics';
 import { EventCustomDimension } from '../analytics/analytics-parameters';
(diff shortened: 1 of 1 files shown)


clydin (approved):

alan-agius4:

PR #29606, opened 2025-02-11
perf(@angular/build): cache translated i18n bundles for faster builds

alan-agius4:
When disk caching is enabled, translated i18n bundles are stored on disk, improving performance and speeding up both incremental and non-incremental builds.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/i18n-inliner.ts
+++ b/packages/angular/build/src/tools/esbuild/i18n-inliner.ts
@@ -7,8 +7,11 @@
  */
 
 import assert from 'node:assert';
+import { createHash } from 'node:crypto';
+import { extname, join } from 'node:path';
 import { WorkerPool } from '../../utils/worker-pool';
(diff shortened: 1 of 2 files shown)


packages/angular/build/src/tools/esbuild/i18n-inliner.ts
```diff
@@ -113,19 +123,54 @@ export class I18nInliner {
...
+      }
+
+      const fileResult = cacheResultPromise.then(async (cachedResult) => {
+        if (cachedResult) {
+          return cachedResult;
+        }
+
+        const result = await this.#workerPool.run({ filename, locale, translation });
+        if (this.#cache && cacheKey) {
+          // Failure to settung the value should not fail the transform
```
clydin, line 166:
```suggestion
          // Failure to set the value should not fail the transform
```

clydin (approved):

alan-agius4:

PR #29607, opened 2025-02-11
fix(@angular/build): configure Vite CORS option

### LINKED ISSUE #29549: Access-Control-Allow-Origin header missing in dev server for assets
### Command

serve

### Is this a regression?

@angular-devkit/build-angular 19.1.4

### Description

Starting with Angular 19.1.5, when running the development server using `ng serve`, the `Access-Control-Allow-Origin` header is no longer included in responses when accessing assets. In version 19.1.4, this header was still being sent with the value '*'.

This affects cross-origin requests to static assets served by the development server.

### Minimal Reproduction

Repository for minimal reproduction:  
[[Link to Git Repository]](https://github.com/vkennke/angular-cors)

Steps to reproduce:

1. Clone repository and install dependencies:
```bash
git clone https://github.com/vkennke/angular-cors.git
cd angular-cors
npm install
```

2. Start dev server:
```bash
ng serve
```

3. Access http://localhost:4200/assets/test.txt and check response headers

Expected behavior:
Response headers include Access-Control-Allow-Origin: *

Actual behavior:
Access-Control-Allow-Origin header is missing from response

To verify this is a regression:
1. Edit package.json to set @angular-devkit/build-angular to 19.1.4:
```json
"@angular-devkit/build-angular": "19.1.4"
```
2. Repeat steps 2-3
3. Observe that the Access-Control-Allow-Origin: * header is present

### Exception or Error

```text
No error is thrown, but the missing header prevents cross-origin requests from succeeding.
```

### Your Environment

```text
Angular CLI: 19.1.5
Node: 20.18.0
Package Manager: npm 10.8.2
OS: win32 x64

Angular: 19.1.4

```

### Anything else relevant?

_No response_

alan-agius4:
Vite's `allowedHosts` option does not enable CORS; instead, it allows the dev server to respond to requests with a matching hostname (e.g., http://example.com/main.js). It only verifies that the request’s hostname is on the allowed list. However, this does not consider the `origin` in the case of a CORS request.

This commit updates Vite's configuration to enable CORS.

Closes #29549

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -841,6 +841,9 @@ export async function setupServer(
           ? (proxy ?? {})
           : proxy,
       cors: {
+        // This will add the header `Access-Control-Allow-Origin: http://example.com`,
+        // where `http://example.com` is the requesting origin.
+        origin: true,
         // Allow preflight requests to be proxied.
         preflightContinue: true,
       },
--- a/packages/angular/build/src/builders/dev-server/schema.json
+++ b/packages/angular/build/src/builders/dev-server/schema.json
@@ -37,12 +37,12 @@
       "description": "SSL certificate to use for serving HTTPS."
     },
     "allowedHosts": {
-      "description": "The hosts that can access the development server. This option sets the Vite option of the same name. For further details: https://vite.dev/config/server-options.html#server-allowedhosts",
+      "description": "The hosts that the development server will respond to. This option sets the Vite option of the same name. For further details: https://vite.dev/config/server-options.html#server-allowedhosts",
       "default": [],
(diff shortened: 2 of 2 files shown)


packages/angular/build/src/builders/dev-server/vite-server.ts
```diff
@@ -841,6 +841,8 @@ export async function setupServer(
           ? (proxy ?? {})
           : proxy,
       cors: {
+        // This will include `Access-Control-Allow-Origin: http://{{Origin}}`
```
AndreKoepke, line 844: Will the header sticked then to `http://`? Or should the comment more looks like this:

```suggestion
        // This will include `Access-Control-Allow-Origin: {{Origin}}`
```

dgp1130 (approved):

PR #29618, opened 2025-02-12
fix(@schematics/angular): remove animations module from ng new app

crisbeto:
Previously the animations module was added to the `ng new` app, because `platform-server` was using it. That's no longer the case as of https://github.com/angular/angular/pull/59762 so these changes remove the dependency.

### DIFF
--- a/tests/legacy-cli/e2e/tests/build/server-rendering/server-routes-output-mode-server-platform-neutral.ts
+++ b/tests/legacy-cli/e2e/tests/build/server-rendering/server-routes-output-mode-server-platform-neutral.ts
@@ -69,10 +69,6 @@ export default async function () {
         },
       ];
     `,
-    'src/app/app.config.ts': `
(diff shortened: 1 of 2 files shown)


tests/legacy-cli/e2e/tests/build/server-rendering/server-routes-output-mode-server-platform-neutral.ts
```diff
@@ -69,10 +69,6 @@ export default async function () {
         },
       ];
     `,
-    'src/app/app.config.ts': `
-      import { provideAnimationsAsync } from '@angular/platform-browser/animations/async';
```
crisbeto, line 73: I dropped this from the test since it seemed irrelevant. Another option was to install animations in the test, but I was having a hard time installing a version that would satisfy the peer dependencies.

alan-agius4 (approved):

PR #29635, opened 2025-02-13
refactor(@angular/cli): handle undefined `ng add` collection name

alan-agius4:
Fixes an issue where JSON help extraction fails if the `ng add` collection name is undefined.


alan-agius4:

Issue #29185, opened 2024-12-19
Karma builder: support `sass` in `stylePreprocessorOptions` with "builderMode: application"


cexbrayat:
### Command

test

### Description

It is currently impossible to define the `sass` option in `stylePreprocessorOptions` for the Karma builder, which makes it impossible to silence Sass deprecations with "builderMode: application."

```json
"test": {
  "builder": "@angular-devkit/build-angular:karma",
  "options": {
  "builderMode": "application",
  "stylePreprocessorOptions": {
    "sass": {
      "silenceDeprecations": ["import"]
    }
  },
```

throws:
```
Error: Schema validation failed with the following errors:
  Data path "/stylePreprocessorOptions" must NOT have additional properties(sass).
```

### Describe the solution you'd like

`stylePreprocessorOptions` allows to set `includePaths` but it would be great to handle `sass` options as well, like the `application` builder does.

alan-agius4:
This is similar to https://github.com/angular/angular-cli/issues/28809, @jkrems up to you if you want to merge these 2 issues.

bst-celum:
The same is true for the `@angular-devkit/build-angular:browser` builder which we are using through the `@angular-builders/custom-webpack` library. (https://github.com/just-jeb/angular-builders/tree/master/packages/custom-webpack#custom-webpack-browser)

@alan-agius4 is the recommendation in general to switch to `build-angular:application`, also for applications which will only run in the browser? Is there documentation for this switch (explanations, things-to-consider, etc.) - it would be very helpful in understanding why/what and how we can migrate.

TokenR1ng:
As stated in another issue, our project needs this feature in the @angular-devkit/build-angular:browser as well. Our project currently cannot migrate to the recommended application builder (which has this flag) and this small extension would really be helpful so that we can at least stay up-to-date with Angular versions. We need the stylePreprocessorOptions#sass option to suppress a lot of warnings comming from Bootstrap. Bootstrap itself will probably never fix this.

Issue #28809, opened 2024-11-07
Feature request: loader option in new esbuild karma builder


stefankoserkov:
### Command

test

### Description

The improved builder @angular-devkit/build-angular:karma which uses esbuild is missing the "loader" option. It would be really useful to have the option to require files in tests.

### Describe the solution you'd like

```json
"test": {
          "builder": "@angular-devkit/build-angular:karma",
          "options": {
            "main": "src/test.ts",
            "polyfills": [
              "zone.js",
              "zone.js/testing"
            ],
            "tsConfig": "tsconfig.spec.json",
            "assets": [
              {
                "glob": "**/*",
                "input": "public"
              }
            ],
            "styles": [
              "src/styles.css"
            ],
            "scripts": [],
            "builderMode": "application",
            **"loader": {
              ".html": "text"
            }**
          }
        }
```

jkrems:
@stefankoserkov Would it be useful if that option would be used from what's already configured for the prod or dev builder? Or would there be cases where it should only apply to the test runner?

stefankoserkov:
@jkrems I'd like it to be an option on the karma builder, to have better separation, but I'll take the other option as well. 

If you are interested the case is that .html files will be required in tests to test something relating to iframes, which will not be used in prod/dev builds. But in our particular case I don't foresee side effects of this loader, if left in prod/dev build configs.

Thank you for the fast response!

jkrems:
For a limited use in tests, you might also be able to use the import attribute syntax, e.g.:

```
import myTestHtml from './testdata/test.html' with { loader: 'text' };
```

If nothing else, this might be a workaround until we can more properly support `loader` in the Karma setup.

stefankoserkov:
I like this workaround it's very cutting edge. However, for enterprise purposes I think it kind of deviates from the Angular path, I don't think we can afford to use module: esnext and declare .html files as modules. For now, I will try to gather votes, and stick to webpack. 

Thank you for the suggestion, I learnt something new.

jeandat:
Is it not strange to not support the same set of properties than in the build target? How are we supposed to migrate to the new karma builder if we cannot build the app? The new builder is useless for me. I have to revert back to webpack (only thing that forces me to still use webpack, can't wait to remove it). 😓😭

jlchmura:
We also had to revert back to the old webpack-based karma builder.  We can't use the new karma target without the loaders option.

jeandat:
[...]

Issue #28321, opened 2024-08-29
Using externalDependencies for scss url() is incompatible with karma


pmoleri:
### Command

build, serve, test

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

I'm trying to use urls relative to the baseHref:
```
"builder": "@angular-devkit/build-angular:application",
"options": {
  "baseHref": "/testbase/",
  "externalDependencies": ["assets/*"],
  ...
}
```

then referencing assets like:
```scss
background-image: url('assets/lockup_angular.svg');
```

it works as expected when running the project but it fails on `ng test`.

I couldn't find any workaround, except using urls relative to the .scss files and getting them handled by the build.

### Minimal Reproduction

1. Create an Angular project with scss, no SSR.
1. Configure the angular.json build with the following options:
    ```
    "builder": "@angular-devkit/build-angular:application",
    "options": {
      "baseHref": "/testbase/",
      "externalDependencies": ["assets/*"],
      ...
    }
    ```
1. ng generate component Test
1. In the scss add
    ```scss
    .test {
      // use path 
      background-image: url('assets/lockup_angular.svg');
    }
    ```
1. run `npm test`

Full example: https://github.com/pmoleri/angular-basehref-scss/blob/master/src/app/very/deep/folder/my.component/my.component.scss

### Exception or Error

```text
Can't resolve './assets/lockup_angular.svg' [postcss-loader]
```

### Your Environment

```text
Angular CLI: 18.2.1
Node: 20.11.1
Package Manager: npm 10.2.4
OS: win32 x64

Angular: 18.2.2
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1802.1
@angular-devkit/build-angular   18.2.1
@angular-devkit/core            18.2.1
@angular-devkit/schematics      18.2.1
@angular/cli                    18.2.1
@schematics/angular             18.2.1
rxjs                            7.8.1
typescript                      5.5.4
zone.js                         0.14.10
```

### Anything else relevant?

I tried all this combinations:

```scss
.relative-to-file {
  // ✔ in @angular-devkit/build-angular:application
  // ✔ in @angular-devkit/build-angular:karma
  // it's inconvenient compared to the other alternatives
  background-image: url('../../../../../assets/lockup_angular.svg');
}

.relative-to-project {
[...]

PR #29640, opened 2025-02-13
feat(@angular/build): add application builder karma testing to package

clydin:
An `application` only variant of the `karma` builder found within the `@angular-devkit/build-angular` package is now available within the `@angular/build` package as `@angular/build:karma`. This builder will only use the `application` builder found within `@angular/build` and does not provide the `builderMode` option as `application` would be the only valid value. Testing behavior is effectively equivalent to using the `@angular-devkit/build-angular:karma` builder with the `builderMode` option set to `application`. However, several options have been adjusted:
* `builderMode` was removed
* `fileReplacements` legacy structure (`src`/`replaceWith`) removed
* `polyfills` only accepts an array of strings
* `loader` has been added
* `define` has been added
* `externalDependencies` has been added
* `sass` sub-option for `stylePreprocessorOptions` has been added

From an implementation perspective, the test finding logic and the application builder integration have been moved to the `@angular/build` package and reused by the `@angular-devkit/build-angular` package via the private export of `@angular/build`.

Closes: #28809
Closes: #28321
Closes: #29185


alan-agius4:
The new builder should be added https://github.com/angular/angular-cli/blob/main/packages/angular/cli/lib/config/workspace-schema.json

jkrems (approved):

PR #29647, opened 2025-02-14
test: reduce flakiness of `reuse-dep-optimization-cache` e2e test

alan-agius4:
This commit reduces test flakiness by modifying the logic to perform the fetch only after the app is bundled.


clydin (approved):

alan-agius4:

PR #29673, opened 2025-02-19
build: temporarily ignore renovate updates for @angular/build-tooling

clydin:
The latest version of the DevInfra internal `@angular/build-tooling` package contains breaking changes that need to be addressed separately before the version can be updated.


josephperrott (approved):
LGTM

PR #29674, opened 2025-02-19
build: fix strict_deps for angular-cli test lib

josephperrott:


packages/angular/cli/BUILD.bazel
```diff
@@ -107,15 +107,14 @@ ts_project(
             "node_modules/**",
         ],
     ),
-    # TODO: Fix strict_deps failure
-    ignore_strict_deps = True,
     deps = [
         ":angular-cli_rjs",
+        ":node_modules/@angular-devkit/core",
```
josephperrott, line 112: Would we prefer to move from the `ts_project` dependency here to the npm package one?

Or should the clean up be to migrate to relative paths? If I understand/remember correctly, we want to use the npm package here so that we don't cross the npm package boundary with relative paths.

If there were imports from `@angular/cli` those should be relative import paths instead though.

devversion: This makes a lot of sense. It's just a leftover because apparently no runtime test broke. My goal is to restrict/fix this and make it safe by leveraging Bazel visibility rules better

devversion (approved):

PR #29684, opened 2025-02-20
Fix misc strict deps throughout repo

josephperrott:


packages/angular_devkit/schematics/BUILD.bazel
```diff
@@ -39,11 +39,9 @@ ts_project(
     name = "schematics_test_lib",
     testonly = True,
     srcs = glob(["src/**/*_spec.ts"]),
-    # TODO: Fix strict_deps failure
-    ignore_strict_deps = True,
     deps = [
         ":node_modules/@angular-devkit/core",
-        ":schematics",
+        ":schematics_rjs",
```
alfaproject, line 44: Sorry for the unrelated drive by comment, but just curious what does `rjs` stand for?

josephperrott: the `_rjs` suffixed package exists as part of our temporary interop as we migrate from one bazel build toolchain to another (from rules_nodejs to rules_js).  It is the actual `ts_project` rule target we ultimately are migrating to.

You can see its definition in [tools/interop.bzl](https://github.com/angular/angular-cli/blob/68ef24d3d7f56dfd8f491b69dc7d882f2b9c860b/tools/interop.bzl)

alfaproject: I see, thanks!

packages/schematics/angular/config/index.ts
```diff
@@ -17,9 +17,12 @@ import {
   strings,
   url,
 } from '@angular-devkit/schematics';
-import { AngularBuilder, readWorkspace, updateWorkspace } from '@schematics/angular/utility';
 import { posix as path } from 'node:path';
 import { relativePathToWorkspaceRoot } from '../utility/paths';
+// TODO: Determine if getWorkspace still needs to be imported as readWorkspace.
+import { getWorkspace as readWorkspace, updateWorkspace } from '../utility/workspace';
+// TODO: Determine if Builders still needs to be imported as AngularBuilder.
```
devversion, line 24: out of curiosity: are these TODOs necessary or can we just drop them? seems fine to import that way

devversion (approved):

packages/schematics/angular/config/index.ts
josephperrott: I left them because I didn't want to make the decision if this is how they should be imported.  They are renamed on export to the public API.  They are unnecessary renames as far as I can tell in the inner workings of the library, but its not my area to know for sure.

devversion: Right, but I worry nobody will act on these TODOs, so I would propose removing the TODOs and let review decide?

josephperrott:
@devversion PTAL as I added the rest of the strict deps fixes into this PR as well.

devversion (approved):
Yay! nice work

clydin (approved):

PR #29687, opened 2025-02-20
fix(@schematics/angular): add additional checks for application builder usage

clydin:
The Angular schematics have been updated to perform additional checks for both the `@angular-devkit/build-angular:application` and `@angular/build:application` builders. This ensures that the schematics generate the appropriate output for each of these build systems in addition to the Webpack-based `browser` builder.


packages/angular/pwa/pwa/index_spec.ts
```diff
@@ -169,4 +169,52 @@ describe('PWA Schematic', () => {
       expect(swFlag).toBeTrue();
     });
   });
+
+  describe('@angular-devkit/build-angular:application builder', () => {
+    function convertBuilderToLegacyBrowser(): void {
```
dgp1130, line 174: Nit: This name seems wrong? I assume "LegacyBrowser" would refer to `@angular-devkit/build-angular:browser`. Should it be "LegacyApplication" or "ApplicationFromDevkit"?

Alternatively you could just inline this into `beforeEach` and avoid the naming problem altogether.

packages/angular/pwa/pwa/index_spec.ts
```diff
@@ -169,4 +169,52 @@ describe('PWA Schematic', () => {
...
+      const configText = tree.readContent('/angular.json');
+      const config = JSON.parse(configText);
+      const swFlag = config.projects.bar.architect.build.configurations.production.serviceWorker;
+
+      expect(swFlag).toBe('projects/bar/ngsw-config.json');
+    });
+  });
+
+  describe('@angular/build:application builder', () => {
+    function convertBuilderToLegacyBrowser(): void {
```
dgp1130, line 198: Same here.

packages/schematics/angular/utility/workspace-models.ts
```diff
@@ -26,11 +26,14 @@ export enum Builders {
...
+  BuildKarma = '@angular/build:karma',
   TsLint = '@angular-devkit/build-angular:tslint',
   NgPackagr = '@angular-devkit/build-angular:ng-packagr',
   BuildNgPackagr = '@angular/build:ng-packagr',
   DevServer = '@angular-devkit/build-angular:dev-server',
+  BuildDevServer = '@angular/build:dev-server',
   ExtractI18n = '@angular-devkit/build-angular:extract-i18n',
+  BuildExtractI18n = '@angular/build:extract-i18n',
   Protractor = '@angular-devkit/build-angular:private-protractor',
   BuildApplication = '@angular/build:application',
```
dgp1130, line 38: Consider: I see that `X` and `BuildX` is an existing convention, but I wonder if it would be worth renaming the existing ones to `DevkitX`, `XDevkit`, or something similar to be more clear about the distinction between them.

Alternatively we could potentially drop the `BuildX` prefix with that being the "default" option we want to use, with `DevkitX` being the legacy versions.

If we do make this change, it's probably worth a separate PR, but something to think about.

dgp1130 (approved):

packages/schematics/angular/utility/workspace-models.ts
clydin: While these are "internal", they have been and probably still are used by third-party custom schematics.  This makes more comprehensive changes problematic.  I do like the `Devkit` variation though.

clydin:

PR #29689, opened 2025-02-20
build: update postcss to v8.5.3

clydin:
A test for the browser builder needed to be updated to reflect a different order in the sources array within the generated source map.


packages/angular_devkit/build_angular/src/builders/browser/specs/vendor-source-map_spec.ts
```diff
@@ -106,9 +106,9 @@ describe('Identifying third-party code in source maps', () => {
     // All sources in the main map are first-party.
     expect(mainMap.sources.filter((_, i) => !mainMap[IGNORE_LIST].includes(i))).toEqual([
       './src/app/app.component.ts',
-      './src/app/app.component.css',
       './src/app/app.module.ts',
       './src/main.ts',
+      './src/app/app.component.css',
     ]);
```
dgp1130, line 112: Suggestion: If we don't care about the order, should we do a set comparison or some other order-irrelevant matcher?

This way future updates won't break this test.

dgp1130 (approved):

PR #29707, opened 2025-02-25
fix(@angular/build): always disable JSON stats with dev-server

clydin:
When using the development server, the `statsJson` option will now unconditionally be disabled. The output JSON file is not accessible with the server and the analysis/generation of the JSON file may increase the rebuild time. Additionally, the JSON file changes during a rebuild may unexpectedly cause component HMR fallback to a full reload due to non-component file changes in the build output.

### DIFF
--- a/packages/angular/build/src/builders/dev-server/vite-server.ts
+++ b/packages/angular/build/src/builders/dev-server/vite-server.ts
@@ -117,6 +117,16 @@ export async function* serveWithVite(
     autoCsp: false,
   };
 
+  // Disable JSON build stats.
+  // These are not accessible with the dev server and can cause HMR fallbacks.
+  if (browserOptions.statsJson === true) {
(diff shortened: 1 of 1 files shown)


packages/angular/build/src/builders/dev-server/vite-server.ts
```diff
@@ -117,6 +117,10 @@ export async function* serveWithVite(
     autoCsp: false,
   };
 
+  // Disable JSON build stats.
+  // These are not accessible with the dev server and can cause HMR fallbacks.
+  browserOptions.statsJson = false;
```
dgp1130, line 122: Question: Is it worth logging that we're ignoring `statsJson`? I could see some devs getting a bit confused that they enable it but nothing happens.

dgp1130 (approved):

PR #29709, opened 2025-02-25
build: update vite to v6.2.0

clydin:


dgp1130 (approved):
Is it worth waiting until after the 19.2.0 release? At minimum we should cut another RC for testing.

PR #29722, opened 2025-02-27
fix(@angular/build): handle undefined `less` stylesheet sourcemap values

clydin:
There can be cases where the `less` stylesheet preprocessor will return an undefined value for a sourcemap even though sourcemaps have been enabled. A check is now performed to handle these cases to prevent potential crashes when processing `less` styles.


jkrems (approved):

clydin:

PR #29723, opened 2025-02-27
fix(@angular/build): allow component HMR with a service worker

clydin:
When the service worker option is enabled, each code update will cause the `ngsw.json` configuration file to change due to the different hashes of the updated output. This is problematic for component HMR usage with the development server since the full reload fallback logic would be triggered due to the changed `ngsw.json` file.  To avoid this problem, the `ngsw.json` configuration file is now special cased within the fallback logic.


alan-agius4 (approved):

alan-agius4:

PR #29740, opened 2025-03-03
refactor(@angular/build): remove `file:` prefix from JIT component resource URLs in sourcemaps

alan-agius4:
Previously, JIT component resource URLs in sourcemaps included a `file:` prefix (e.g., `file:src/app/app.component.html`). This change removes the `file:` prefix to ensure cleaner source mappings.


clydin (approved):

alan-agius4:

PR #29747, opened 2025-03-04
fix(@angular-devkit/architect): improve error message when configuration is mising

### LINKED ISSUE #29742: Unclear error message when a configuration is not present for a project
### Command

build

### Is this a regression?

_No response_

### Description

This is a similar issue as reported in #29226.

When running a target, if a configuration isn't present, an error message is presented, as seen in the [node architect host here](https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/architect/node/node-modules-architect-host.ts#L93). This is helpful, but especially when dealing with complex nested projects, the error message is not very helpful, as the project/target in question is not indicated.

This can be improved by changing that line as follows:

```
if (!targetDefinition.configurations?.[configuration]) {
  throw new Error(`Configuration '${configuration}' for target '${target}' in project '${project}' is not set in the workspace.`);
}
```

### Minimal Reproduction

Attempt to run any comment for a project:target with a configuration that does not exist.  The project and target are not indicated in the error message.

### Exception or Error

```text

```

### Your Environment

```text
n/a
```

### Anything else relevant?

_No response_

alan-agius4:
This commit makes the error message more actionable.

Closes #29742


clydin (approved):

alan-agius4:

PR #29749, opened 2025-03-04
test: only test ng update for supported versions

alan-agius4:
Removes tests for versions 15 and 16 which are no longer in LTS.


clydin (approved):

alan-agius4:

PR #29750, opened 2025-03-04
[19.2.x] build: update dev-infra and rework windows native testing

devversion:
As part of go/ng:windows-dev-future, we are changing how our infrastructure supports Windows build & testing. Clearly:

- we will still support contributors on Windows, and we believe we will be improving and streamlining the experience here
- we will continue testing the Angular CLI for our Windows users. We are aware of the many Windows users using the `ng` CLI.

What is changing? We are no longer actively working towards a Bazel infrastructure that supports native Windows building and testing. There are currently two ways to contribute to Angular on Windows. That is via WSL, or via e.g. native Windows cmd.exe, with Git Bash on top. We acknowledge that the latter worked sometimes, but we also realize it very often breaks as nobody on our team uses, verifies it, and it introduces extra complexity because Bazel on Windows is quite disconnected from Linux/Mac (e.g. no sandboxing). Going forward, to improve our team's effectiveness, and improve our stability guarantees for Windows (and Windows contributors), we are actively discouraging the use of Git Bash for contributing to Angular; but instead ask for WSL to be used. I can speak as one of the few long-term team members that have worked on Windows (without WSL) most of my time, that WSL is great and the contributing experience is much smoother and also easier to "guide". It's a positive change because we won't be suggesting "two ways to contribute on Windows", where in reality one is very brittle and can break at any time!


josephperrott (approved):
LGTM

PR #29751, opened 2025-03-04
perf(@angular/ssr): optimize response times by introducing header flushing

alan-agius4:
Improves SSR performance by streaming the response before inlining critical CSS. This allows for earlier header flushing, reducing time to first byte (TTFB) and improving perceived load times.


jkrems (approved):
:shipit:

PR #29753, opened 2025-03-04
fix(@angular/build): exclude all entrypoints of a library from prebundling

### LINKED ISSUE #29170: Exclude all entrypoints of a library from prebundling
### Command

serve

### Is this a regression?

_No response_

### Background: 

I am currently developing a library (my-library) using ng-packagr with multiple secondary entry points. I also have another repo (my-application) that depends on the library. For setting up a local development environment, I am using `npm link` to link `dist/lib-a` to the `my-application` repo, running `ng watch` in `lib-a` and `ng serve` in `my-application` when performing development. 

I knew that in order to make `ng serve` detect the code changes from the linked lib-a, I need to do following: 
- In angular.json, specify `architect.build.configurations.development.preserveSymlink: true`
- In tsconfig.json, define the paths property (define prebundle.exclude: "lib-a" in angular.json is not helpful, cause it doesnt support glob pattern to match all of my library's sub-entries): 
```
"paths": {
      "lib-a": ["./node_modules/lib-a"],
      "lib-a/*": ["./node_modules/lib-a/*"],
    },
```

In my-library (lib-a), I have these 2 sub-entry points, component-y and service-x. Basically the component-y depends on service-x to show some data that is getting from the token that injected in service-x. In my-application (app.config.ts), I will provide the value of the token. 

service-x
```
export const X_TOKEN = new InjectionToken('X_TOKEN');

@Injectable({ providedIn: 'root' })
export class ServiceX {
    myXToken = inject(X_TOKEN);
} 

```

component-y
```
@Component({
    changeDetection: ChangeDetectionStrategy.OnPush,
    selector: "component-y",
    template: "<p>lib-a-component works! {{ myServiceX.myXToken}}</p>"
})
export class ComponentY {
    protected myServiceX = inject(ServiceX);
} 
```

In my-application (app.config.ts)
```
export const appConfig: ApplicationConfig = {
  providers: [
    provideZoneChangeDetection({ eventCoalescing: true }), 
    { provide: X_TOKEN, useValue: 'OTHER_VALUE'},
    provideRouter(routes)
  ]
};
```

### Actual Issue
Everything is working fine when it is being built or serve in `my-library`. 
**_However, in `my-application`, the behavior is not working as I expected when running with `ng serve`._** 
It produces following error when I navigate to the page that using the component-y: 

```
ERROR NullInjectorError: R3InjectorError(Environment Injector)[_ServiceX -> InjectionToken X_TOKEN -> InjectionToken X_TOKEN]: 
  NullInjectorError: No provider for InjectionToken X_TOKEN!
```

Observations: 
- If I remove the paths property in tsconfig, the null injector error is not observed in the next following `ng serve`.
- I also tested the built files that generated from ng build (development/production) in my-application with angular-http-server. Everything is working fine. 
- The above issue happens in Angular 18 as well. 

### Expected behavior:  
Running ng serve with npm link library should behave the same as running ng serve with library that installed using npm install.
I suspect this is a bug because I cant reproduce this in the codes that generated in ng build (preserveSymlink is specified also). 

### Other Questions: 
1. Can we just define prebundle.exclude: ["lib-a"] instead of ["lib-a/component-y", "lib-a/service-x"] ? Currently, if there is any code changes done in the sub-entries, it wont detect as a change if we define as ["lib-a"]. As an alternative, I am doing it now in tsconfig instead.
2. Is it valid to provide the X_TOKEN value in app.config, where the token will be injected in a root service (ServiceX)? I assume provide token in app.config is the same as providedIn: 'root' and it should not cause any race-condition issue that causes the root service not able to inject the token.

Thanks and appreciate your help on this.  

### Minimal Reproduction

my-library: 
https://github.com/kjteh/my-library-npm-link

my-application: 
https://github.com/kjteh/my-application-npm-link

Steps: 
1. In my-library-npm-link, run `npm run watch `
2. cd to dist/lib-a, run `npm link`
3. In my-application-npm-link, run `npm link lib-a --save`
4. Run `ng serve` in my-application-npm-link
5. Go to localhost:4200/pagez 
6. NullInjector error is shown in the console. 

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 19.0.5
Node: 18.20.4
Package Manager: npm 10.7.0
OS: win32 x64

Angular: 19.0.4

@angular-devkit/architect       0.1900.5
@angular-devkit/build-angular   19.0.5
@angular-devkit/core            19.0.5
@angular-devkit/schematics      19.0.5
@angular/cli                    19.0.5
@schematics/angular             19.0.5
```

### Anything else relevant?

_No response_

alan-agius4:
The configuration now ensures that when a package is listed for exclusion, all paths within that package including sub-paths like `@foo/bar/baz`  are marked as external and not prebundled by the development server.

For example, specifying `@foo/bar` in the exclude list will prevent the development server from bundling any files from the `@foo/bar` package, including its sub-paths such as `@foo/bar/baz`.

This aligns with esbuild external option behaviour https://esbuild.github.io/api/#external

Closes #29170

### DIFF
--- a/packages/angular/build/src/builders/application/execute-build.ts
+++ b/packages/angular/build/src/builders/application/execute-build.ts
@@ -163,21 +163,49 @@ export async function executeBuild(
   // Analyze external imports if external options are enabled
   if (options.externalPackages || bundlingResult.externalConfiguration) {
     const {
-      externalConfiguration,
-      externalImports: { browser, server },
+      externalConfiguration = [],
+      externalImports: { browser = [], server = [] },
     } = bundlingResult;
-    const implicitBrowser = browser ? [...browser] : [];
-    const implicitServer = server ? [...server] : [];
-    // TODO: Implement wildcard externalConfiguration filtering
-    executionResult.setExternalMetadata(
-      externalConfiguration
-        ? implicitBrowser.filter((value) => !externalConfiguration.includes(value))
-        : implicitBrowser,
-      externalConfiguration
-        ? implicitServer.filter((value) => !externalConfiguration.includes(value))
-        : implicitServer,
-      externalConfiguration,
-    );
+    // Similar to esbuild, --external:@foo/bar automatically implies --external:@foo/bar/*,
+    // which matches import paths like @foo/bar/baz.
+    // This means all paths within the @foo/bar package are also marked as external.
+    const exclusionsPrefixes = externalConfiguration.map((exclusion) => exclusion + '/');
+    const exclusions = new Set(externalConfiguration);
+    const explicitExternal = new Set<string>();
+
+    const isExplicitExternal = (dep: string): boolean => {
+      if (exclusions.has(dep)) {
+        return true;
+      }
+
+      for (const prefix of exclusionsPrefixes) {
+        if (dep.startsWith(prefix)) {
+          return true;
+        }
+      }
+
+      return false;
+    };
+
+    const implicitBrowser: string[] = [];
+    for (const dep of browser) {
+      if (isExplicitExternal(dep)) {
+        explicitExternal.add(dep);
+      } else {
+        implicitBrowser.push(dep);
+      }
+    }
+
+    const implicitServer: string[] = [];
+    for (const dep of server) {
+      if (isExplicitExternal(dep)) {
+        explicitExternal.add(dep);
+      } else {
+        implicitServer.push(dep);
+      }
+    }
+
+    executionResult.setExternalMetadata(implicitBrowser, implicitServer, [...explicitExternal]);
   }
 
   const { metafile, initialFiles, outputFiles } = bundlingResult;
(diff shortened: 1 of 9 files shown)


packages/angular/build/src/builders/application/execute-build.ts
```diff
@@ -163,21 +163,49 @@ export async function executeBuild(
...
-      externalConfiguration,
-    );
+    // Similar to esbuild, --external:@foo/bar automatically implies --external:@foo/bar/*,
+    // which matches import paths like @foo/bar/baz.
+    // This means all paths within the @foo/bar package are also marked as external.
+    const exclusionsPrefixes = externalConfiguration.map((exclusion) => exclusion + '/');
+    const exclusions = new Set(externalConfiguration);
+    const explicitExternal = new Set<string>();
+
+    const isExplicateExternal = (dep: string): boolean => {
```
clydin, line 176:
```suggestion
    const isExplicitExternal = (dep: string): boolean => {
```

clydin (approved):

PR #29757, opened 2025-03-05
feat(@angular/build): integrate Chrome automatic  workspace folders

alan-agius4:
This commit integrates automatic Chrome DevTools workspace folder connection into the vite dev-server process, leveraging the experimental feature available in Chrome Canary, as described in the Chrome DevTools documentation https://chromium.googlesource.com/devtools/devtools-frontend/+/main/docs/ecosystem/automatic_workspace_folders.md


packages/angular/build/src/tools/vite/middlewares/chrome-devtools-middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+    } catch {
+      const devtoolsConfigJson = {
+        workspace: {
+          root: projectRoot,
+          uuid: randomUUID(),
+        },
+      };
+
+      devtoolsConfig = JSON.stringify(devtoolsConfigJson, undefined, 2);
+      writeFileSync(devtoolsConfigPath, devtoolsConfig);
```
clydin, line 42: Maybe add a try catch around this as well?  Failing to cache it shouldn't be fatal.
Also, is the containing directory guaranteed to be present?

clydin (approved):

packages/angular/build/src/tools/vite/middlewares/chrome-devtools-middleware.ts
```diff
@@ -0,0 +1,51 @@
...
+  return function chromeDevtoolsMiddleware(req, res, next) {
+    if (req.url !== CHROME_DEVTOOLS_ROUTE) {
+      next();
+
+      return;
+    }
+
+    // We store the UUID and re-use it to ensure Chrome does not repeatedly ask for permissions when restarting the dev server.
+    try {
+      devtoolsConfig ??= readFileSync(devtoolsConfigPath, 'utf-8');
```
bmeurer, line 32: Maybe I'm missing something, but what happens if you relocate the project folder? That doesn't seem to trigger deletion of the cached file.

alan-agius4: That is correct, we could add a check here to validate that the path is the same otherwise we could re-generate the entry.

PR #29760, opened 2025-03-05
test: update `ng-add` version specifier test to bypass npm peer dependency issues

alan-agius4:
NPM's peer dependency resolution has known inconsistencies, which can cause issues when testing `ng-add`.


clydin (approved):

alan-agius4:

PR #29779, opened 2025-03-07
Further clean-up `rules_nodejs` `npm` workspace and remove `yarn.lock` 

devversion:
See individual commits


alan-agius4:
Note: we need to update the renovate dev-infra action to install PNPM.

devversion:
>  Note: we need to update the renovate dev-infra action to install PNPM.

I think @josephperrott took care of this already.

josephperrott:
Yes, the renovate action should not have `pnpm` installed in its environment as of last thursday.

.github/shared-actions/windows-bazel-test/action.yml
```diff
@@ -27,17 +27,21 @@ runs:
       with:
         wsl_firewall_interface: 'vEthernet (WSL (Hyper-V firewall))'
 
+    - name: Installing pnpm (in WSL)
```
josephperrott, line 30: Can this be done using `pnpm/action-setup`?

.gitignore
```diff
@@ -4,14 +4,6 @@ test-project-host-*
 dist/
 dist-schema/
 
-# Yarn
```
josephperrott, line 7: These should stay for now because if you go back to an LTS branch these will end up existing. If we want to create a TODO for their removal we can probably do it in about a year.

packages/angular/ssr/node/test/BUILD.bazel
```diff
@@ -7,16 +7,13 @@ ts_project(
     srcs = glob(["**/*_spec.ts"]),
     deps = [
         "//:node_modules/@types/node",
-        "//packages/angular/ssr/node:node_rjs",
+        "//packages/angular/ssr/node",
     ],
 )
 
 # TODO: Clean this up when this repo runs ESM consistently.
```
josephperrott, line 14: Is this cleaned up now then? Or is this TODO about something else.

josephperrott (changes requested):
`.yarn/install-state.gz` Should not be installed

.github/shared-actions/windows-bazel-test/action.yml
devversion: No, because this is inside WSL

packages/angular/ssr/node/test/BUILD.bazel
devversion: This is about executing ESM in the repo. Unrelated to the migration.

josephperrott (approved):
LGTM after the requested changes are addressed.

clydin (approved):
As mentioned, `.yarn/install-state.gz` should be removed. Also, `patch-package` should be removed as well since it's no longer needed.
Otherwise, LGTM.

devversion:
Addressed feedback. Also note: I will improve/clean-up more of the documentation in a follow-up!

alan-agius4 (approved):
LGTM

PR #29790, opened 2025-03-10
refactor(@angular/build): remove esbuild sourcemap workarounds

alan-agius4:
These are no longer requires since 0.25.1


clydin (approved):

alan-agius4:

PR #29800, opened 2025-03-11
fix(@angular/cli): record analytics for nested schematics

alan-agius4:
Prior to this commit, analytics were not recorded for nested schematics. This caused certain data to be incomplete. For example, when running `ng new` and selecting "yes" for SSR in the prompt, this choice was not recorded because the prompt exists within the `application` schematic.


clydin (approved):

alan-agius4:

PR #29828, opened 2025-03-12
refactor(@angular/build): define an internal ngHmrMode value

clydin:
An `ngHmrMode` boolean value will now be defined within application bundles. The value is based on the development server's `hmr` option with `true` when HMR is enabled and `false` when disabled. For all application builds, the value will be `false`. `ngHmrMode` is similar in behavior to `ngServeMode` or `ngDevMode`. It will not be present in the output code unless referenced and in those cases only the final boolean value will be present if not optimized out of the final code.
The value is not considered part of the public API and may change in the future.


alan-agius4 (approved):

clydin:

PR #29831, opened 2025-03-13
build: avoid WSL interop flake when preparing Windows CI environment

devversion:
For our native Windows tests (verifying `ng` CLI outside WSL), we currently build the test binaries inside WSL, leveraging bazel remote execution.

The scripts involved in moving the built binaries to outside WSL rarely flake due to a WSL interop bug when we invoke Windows commands from inside WSL (which is an untypical thing to do; but a needed trick given some WSL limitation).

This commit tries to stabilize/avoid this flake by retrying if we recognize this

### DIFF
--- a/scripts/windows-testing/convert-symlinks.mjs
+++ b/scripts/windows-testing/convert-symlinks.mjs
@@ -119,10 +119,22 @@ async function transformDir(p) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):

devversion:

PR #29846, opened 2025-03-13
feat(@angular/ssr): stabilize `AngularNodeAppEngine`, `AngularAppEngine`, and `provideServerRouting` APIs

alan-agius4:
**feat(@angular/ssr): stabilize `AngularNodeAppEngine`, `AngularAppEngine`, and `provideServerRouting` APIs**

This commit promotes the `AngularNodeAppEngine`, `AngularAppEngine`, and `provideServerRouting` APIs from dev preview to stable. These APIs enhance server-side rendering (SSR) capabilities in Angular applications, improving routing and server integration for better performance and reliability.

----

**feat(@schematics/angular): remove `--server-routing` option**

Server-side rendering (SSR) will always enable server routing when using the application builder.

BREAKING CHANGE: `--server-routing` option has been removed from several schematics. Server routing will be used when using the application builder.


dgp1130 (approved):
Would it be worth leaving `--no-server-routing` as an option (while defaulting to enabling server routing) for some period of time? Or do you feel ready to make that the only supported option for `application`  builder?

alan-agius4:
> Would it be worth leaving `--no-server-routing` as an option (while defaulting to enabling server routing) for some period of time? Or do you feel ready to make that the only supported option for `application` builder?

I’d definitely want to avoid new users using the `CommonEngine`. The reason why that is still around and not deprecated is because of the webpack based builders.

PR #29847, opened 2025-03-13
docs(@angular-devkit/build-angular): update `forceEsbuild` option description

alan-agius4:
Esbuild builder is stable and is no longer in developer preview.


alan-agius4:

PR #29856, opened 2025-03-14
docs: correct bazel e2e commands

alan-agius4:
These command are outdated and no longer work


devversion (approved):

alan-agius4:

PR #29857, opened 2025-03-14
build: move build-angular/build-webpack specific dependencies out of root

clydin:
With the migration to `rules_js`, package specific dependencies now only need to be referenced with the source `package.json` for each specific package. Most of the `@angular-devkit/build-angular` and `@angular-devkit/build-webpack` specific dependencies have now been moved. This is not exhaustive and further changes will continue to move additional dependencies.


devversion (commented):
so great 🎉

PR #29876, opened 2025-03-18
build: improve debugging mode of e2e test runner

devversion:
Currently the `while` loop may either be causing the process to be really stuck/hanging, or it somehow causes Node to exit.

This change makes the logic more robust and less CPU consuming.


alan-agius4 (approved):

alan-agius4:

PR #29881, opened 2025-03-19
build: update development Node.js version to 22.14.0 and auto-update

clydin:
The version of Node.js used for development and CI jobs is now set to the active LTS version's latest minor and patch. Renovate is also now configured to update the minor and patch versions once a week.


renovate.json
```diff
@@ -20,9 +20,20 @@
...
     "packages/**/package.json",
     "tests/legacy-cli/e2e/ng-snapshot/package.json",
-    ".github/workflows/**/*.yml"
+    ".github/workflows/**/*.yml",
+    ".nvmrc"
   ],
   "packageRules": [
+    {
+      "matchDepNames": ["node"],
+      "schedule": ["before 4:00am on monday"],
```
alan-agius4, line 29: Do we need schedule this on a monday? Can't be leave to regular scheduling?

alan-agius4 (approved):

PR #29882, opened 2025-03-19
fix(@schematics/angular): replace `@angular/platform-browser-dynamic` with `@angular/platform-browser`

alan-agius4:
The Angular CLI explicitly adds `@angular/compiler` as a polyfill when JIT mode is enabled which makes using  `@angular/platform-browser-dynamic` redundant and causes the compiler to be imported twice.


devversion (approved):

clydin:
Should the dependency be removed from the package.json as well?

alan-agius4:
We are still using it for testing, can be moved once https://github.com/angular/angular/pull/60453 is landed.

alan-agius4:

PR #29888, opened 2025-03-19
fix(@angular/build): ensure errors for missing component resources

### LINKED ISSUE #29887: No compilation errors for non-existing css files referred in component's styleUrl/styleUrls

Don't known / other

### Is this a regression?

Yes

### Description

When using @angular-devkit/build-angular:application in new Angular there seems to be no check for existence of css files referred from the component.

To replicate, create empty project using the latest Angular, then edit `src/app/app.component.ts` and change .css file path to non-existing file or any string.

After that both `ng serve` and `ng build` still run successfully without any complains/warnings.

Unit tests `ng test` is failing (probably because it's still using webpack)

_No response_

```true

```

```true
Angular CLI: 19.2.3
Node: 18.20.6
Package Manager: npm 10.8.2
OS: linux x64

Angular: 19.2.2

@angular-devkit/architect       0.1902.3
@angular-devkit/build-angular   19.2.3
@angular-devkit/core            19.2.3
@angular-devkit/schematics      19.2.3
@angular/cli                    19.2.3
@schematics/angular             19.2.3
rxjs                            7.8.2
```

### Anything else?

_No response_

clydin:
Compile time errors will now always be generated when an HTML template and/or CSS component stylesheet are not present on disk. The Angular compiler expects a return value of `null` from the `resourceNameToFileName` function to indicate that a resource path does not exist. JIT mode previously and continues to generate errors as expected.

Closes #29887

### DIFF
--- a/packages/angular/build/src/builders/application/tests/behavior/component-stylesheets_spec.ts
+++ b/packages/angular/build/src/builders/application/tests/behavior/component-stylesheets_spec.ts
@@ -44,5 +44,44 @@ describeBuilder(buildApplication, APPLICATION_BUILDER_INFO, (harness) => {
 
       harness.expectFile('dist/browser/main.js').content.not.toContain('variables');
     });
+
+    it('should generater an error for a missing stylesheet with AOT', async () => {
+      await harness.modifyFile('src/app/app.component.ts', (content) => {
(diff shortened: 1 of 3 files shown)


alan-agius4 (approved):

clydin:

PR #29895, opened 2025-03-19
build: move rxjs dependency out of root

clydin:
With the migration to `rules_js`, package specific dependencies now only need to be referenced with the source `package.json` for each specific package. The `rxjs` dependency has now been moved to usage locations instead of the root of the repository.


devversion (approved):
clean 🎉

PR #29898, opened 2025-03-19
feat(@angular/build): support a default outputPath option for applications

clydin:
The application builder will now use a default output path when the `outputPath` option is not specified either in the `angular.json` configuration or via the command line.  The default used will be `dist/<project_name>` and be relative to the workspace root. This value is the typical default for new projects. Projects may continue to customize the output path via the option if needed. Existing project behavior and configuration will not be changed.

### DIFF
--- a/packages/angular/build/src/builders/application/options.ts
+++ b/packages/angular/build/src/builders/application/options.ts
@@ -295,7 +295,7 @@ export async function normalizeOptions(
     };
   }
 
-  const outputPath = options.outputPath;
+  const outputPath = options.outputPath ?? path.join(workspaceRoot, 'dist', projectName);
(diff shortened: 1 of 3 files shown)


packages/angular/build/src/builders/application/options.ts
```diff
@@ -295,7 +295,7 @@ export async function normalizeOptions(
     };
   }
 
-  const outputPath = options.outputPath;
+  const outputPath = options.outputPath ?? path.join(projectRoot, 'dist', projectName);
```
alan-agius4, line 298:
```suggestion
  const outputPath = options.outputPath ?? path.join(workspaceRoot, 'dist', projectName);
```

alan-agius4 (approved):

PR #29923, opened 2025-03-25
fix(@angular/build): normalize karma asset paths before lookup

clydin:
When handling a request for a configured asset with the application- based karma unit testing, the asset URL is now normalized to the executing platform's path format before looking up the asset. This is required due to the build file paths being based on the underlying operating system's paths which may not align with a URL's path separator such as when using Windows.


alan-agius4 (approved):

alan-agius4:

PR #29933, opened 2025-03-25
refactor(@angular/build): use newer Node.js `cp` API for asset copying

clydin:
When using Node.js v22 (minimum of v22.11 for v22 with Angular v20), the application build system will now use the Node.js `cp` filesystem API instead of the `copyFile` API. This newer API provides equivalent functionality while also preserving timestamps for copied assets.  Additionally, it supports potential future internal refactorings to support full direct directory copying. Projects built with Node.js v20 will continue to use the `copyFile` API.

### DIFF
--- a/packages/angular/build/src/builders/application/index.ts
+++ b/packages/angular/build/src/builders/application/index.ts
@@ -28,6 +28,8 @@ import {
 import { Result, ResultKind } from './results';
 import { Schema as ApplicationBuilderOptions } from './schema';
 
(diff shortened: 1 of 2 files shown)


alfaproject:
Node 20 won't be supported anymore for Angular 20? Thought only 18 was dropped

clydin:
22.11 is the minimum version of Node.js v22 that can be used with Angular v20.  The `cp` API was stabilized in Node.js 22.3.

alan-agius4 (approved):

PR #29970, opened 2025-03-29
refactor: ensure application js is output as utf8

JeanMeche:


alan-agius4:
Hi @JeanMeche, we output ASCII as it’s more performant.

 See: https://v8.dev/blog/scanner#conclusion:~:text=As%20a%20developer%20you%20can%20further,to%20avoid%20non%2DASCII%20identifiers%20where%20possible

jkrems:
Is this in reference to:

> The easiest way to do so is by minifying your source code, stripping out unnecessary whitespace, **and to avoid non-ASCII identifiers where possible.**

I would've read this as:

"The identifier `a` (latin a) looks the same length as `α` (greek alpha) but the scanner needs to look at one instead of two bytes to parse `a`. You should prefer ASCII identifiers to reduce the number of source bytes."

Turning `ɵ` (2 UTF8 bytes) into `\u0275` (6 bytes) sounds like it would make things slower rather than faster? Or are escape sequences faster than reading the literal characters? Intuitively, escape sequences should require more parsing because they involve number parsing, string-to-number, followed by collapsing that number back into a different string. Potentially also more allocation things because it can't just "slice off" the identifier name from the source string.

jkrems:
Interesting! Thanks for the details. Forgot the latin1 of it all. :)

PR #29979, opened 2025-03-31
refactor(@angular/build): provide a pre-link resolution condition for APF

clydin:
To support future APF changes that may include pre-linked code specific to the Angular version used to create the package, an additional package resolution condition will now be automatically added to the build process for applications. A condition in the form of `angular:linked-<version>`, where `<version>` is the full package version of `@angular/core`, will now be present during application builds. This change in combination with APF changes will allow matching packages to avoid the linking step during a build. Non-version matching packages will continue to function and use the existing linking transformation process.

### DIFF
--- a/packages/angular/build/src/builders/application/options.ts
+++ b/packages/angular/build/src/builders/application/options.ts
@@ -9,7 +9,7 @@
 import type { BuilderContext } from '@angular-devkit/architect';
 import type { Plugin } from 'esbuild';
 import { realpathSync } from 'node:fs';
-import { access, constants } from 'node:fs/promises';
+import { access, constants, readFile } from 'node:fs/promises';
 import { createRequire } from 'node:module';
 import path from 'node:path';
 import { normalizeAssetPatterns, normalizeOptimization, normalizeSourceMaps } from '../../utils';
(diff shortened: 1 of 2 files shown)


packages/angular/build/src/builders/application/options.ts
```diff
@@ -706,3 +707,22 @@ function normalizeExternals(value: string[] | undefined): string[] | undefined {
 
   return [...new Set(value.map((d) => (d.endsWith('/*') ? d.slice(0, -2) : d)))];
 }
+
+async function findFrameworkVersion(projectRoot: string): Promise<string> {
```
alan-agius4, line 711: Any reason why not using the `VERSION` const exported from `@angular/core` directly?

packages/angular/build/src/tools/esbuild/application-code-bundle.ts
```diff
@@ -570,6 +571,12 @@ function getEsBuildCommonOptions(options: NormalizedApplicationBuildOptions): Bu
     'es2015',
     'es2020',
   ];
+  // JIT requires additional metadata not included in the pre-linked code
+  if (!jit) {
```
devversion, line 575: The linker does support preserving JIT code if I remember correctly. Should the APF code run with that mode?

devversion (approved):
🎉

packages/angular/build/src/tools/esbuild/application-code-bundle.ts
clydin: JIT is currently setup to link at runtime with the CLI so i would avoid the extra code for the common AOT case.

Issue #29978, opened 2025-03-31
[CLI][Schematics] Override Angular Schematics issue - Collection "[Schematics folder path]" cannot be resolved. - Windows


SlawomiKussowski:
### Command

generate

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### The previous version in which this bug was not present was

Unknown

### Description

This is similar issue to https://github.com/angular/angular-cli/issues/29559.
The behavior has been fixed but only in ng generate command is executed in the root folder.

If you try to override standard Angular Schematics with a custom one by the use of schematicCollections option in angular.json file like for instance:

```JSON
 "cli": { "analytics": false, 
"schematicCollections": [
   "./schematics", 
  "@schematics/angular" ]
 }
```
And create your custom schematics (index.ts):
```TS
import {
  externalSchematic,
  Rule,
  SchematicContext,
  Tree
} from '@angular-devkit/schematics';
import { Schema as ComponentOptions } from '@schematics/angular/component/schema';

export function generateComponent(options: ComponentOptions): Rule {
  console.log('options', structuredClone(options));
  return (_tree: Tree, _context: SchematicContext) => {
    return externalSchematic(
      "@schematics/angular",
      "component",
      options
    );
  };
}
```

And then go to any subfolder.
It will end up with an error message like this one:
Collection "D:\\projects\\schematics-test_v19\\src\\app\\schematics" cannot be resolved.

Please use the attached repo to reproduce the issue:

[schematics-test_v19.zip](https://github.com/user-attachments/files/19540005/schematics-test_v19.zip)

### Minimal Reproduction

Run npm install and then go to a subfolder, for instance:
cd .\src\app\
and try to generate a component using:
ng g c
command.

### Exception or Error

```text
Collection "D:\\projects\\schematics-test_v19\\src\\app\\schematics" cannot be resolved.
```

### Your Environment

```text
Angular CLI: 19.2.5
Node: 22.13.0
Package Manager: npm 8.6.0
OS: win32 x64

Angular: 19.2.4
... animations, common, compiler, compiler-cli, core, forms
... platform-browser, platform-browser-dynamic, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1902.5
@angular-devkit/build-angular   19.2.5
@angular-devkit/core            19.2.5
@angular-devkit/schematics      19.2.5
@angular/cli                    19.2.5
@schematics/angular             19.2.5
rxjs                            7.8.1
[...]

PR #29986, opened 2025-04-01
fix(@angular-devkit/schematics): properly resolve relative schematics when executed from a nested directory

alan-agius4:
Ensure that relative schematic paths are correctly resolved when invoked from within a nested directory.

Closes: #29978


clydin (approved):

alan-agius4:

PR #29989, opened 2025-04-01
refactor(@schematics/angular): update templates to use express version 5

alan-agius4:


alan-agius4:
The lint failure is due to `  pnpm-lock.yaml » cookie-signature@1.2.2 – License: LicenseRef-scancode-unknown-license-reference AND MIT` but checking the package it appears to be only `MIT` https://app.unpkg.com/cookie-signature@1.2.2/files/package.json

clydin (approved):

PR #29990, opened 2025-04-01
build: setup `rules_esbuild` in preparation for `devinfra` update

devversion:
The shared dev-infra repository will depend on `rules_esbuild` going forward, so we need to install it.

Unblocks https://github.com/angular/angular-cli/pull/29984


jkrems:

PR #30021, opened 2025-04-03
fix(@angular/build): include component test metadata in development builds

clydin:
To support the usage of AOT with unit tests, development builds of applications will now include injected calls to the internal Angular `setClassMetadata` function. These calls add metadata to each component that can be leveraged by the `TestBed` to override component information.


alan-agius4 (approved):

clydin:

PR #30028, opened 2025-04-04
fix(@schematics/angular): infer app component name and path in server schematic

crisbeto:
Currently the `server` schematic assumes that the app component is called `App` and it's places in `./app/app`. This will fail if the user renamed it or moved it to a different file.

These changes add a utility function to resolve the component name and path from the source the source code, and they use the new function to produce a more accurate result.

### DIFF
--- a/packages/schematics/angular/server/files/application-builder/standalone-src/main.server.ts.template
+++ b/packages/schematics/angular/server/files/application-builder/standalone-src/main.server.ts.template
@@ -1,7 +1,7 @@
 import { bootstrapApplication } from '@angular/platform-browser';
-import { App } from './app/app';
+import { <%= appComponentName %> } from '<%= appComponentPath %>';
 import { config } from './app/app.config.server';
 
-const bootstrap = () => bootstrapApplication(App, config);
+const bootstrap = () => bootstrapApplication(<%= appComponentName %>, config);
 
 export default bootstrap;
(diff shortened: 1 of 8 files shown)


packages/schematics/angular/server/files/application-builder/standalone-src/main.server.ts.template
```diff
@@ -1,7 +1,7 @@
 import { bootstrapApplication } from '@angular/platform-browser';
-import { App } from './app/app';
```
crisbeto, line 2: Not sure if this is the only place we need to change. I see that there are 3 more `main.server.ts.template` files.

alan-agius4: It needs to be also replaced in;
- https://github.com/angular/angular-cli/blob/597af7f35d83b98fe8fc03e8f239ae2049b9651c/packages/schematics/angular/server/files/application-builder/ngmodule-src/app/app.module.server.ts.template#L3
- https://github.com/angular/angular-cli/blob/597af7f35d83b98fe8fc03e8f239ae2049b9651c/packages/schematics/angular/server/files/server-builder/ngmodule-src/app/app.module.server.ts.template#L5
- https://github.com/angular/angular-cli/blob/597af7f35d83b98fe8fc03e8f239ae2049b9651c/packages/schematics/angular/server/files/server-builder/standalone-src/main.server.ts.template#L2

crisbeto: Done.

alan-agius4 (approved):
LGTM

PR #30034, opened 2025-04-04
Remove "experimental" from zoneless and add prompt for new apps

atscott:
See individual commits.

### DIFF
--- a/packages/angular/ssr/test/testing-utils.ts
+++ b/packages/angular/ssr/test/testing-utils.ts
@@ -11,7 +11,7 @@ import {
   EnvironmentProviders,
   Provider,
   Type,
-  provideExperimentalZonelessChangeDetection,
+  provideZonelessChangeDetection,
 } from '@angular/core';
 import { bootstrapApplication } from '@angular/platform-browser';
 import { RouterOutlet, Routes, provideRouter } from '@angular/router';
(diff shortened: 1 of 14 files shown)


alfaproject:
OOC is there a world where we can have both zones and zoneless in the same project? Asking for the sake of incremental migrations

atscott:
> OOC is there a world where we can have both zones and zoneless in the same project? Asking for the sake of incremental migrations

No, not for that definition of “incremental”. But you can still incrementally migrate components to be Zoneless compatible. Everything Zoneless also works with zones. Everything in the angular components repo was migrated incrementally to be Zoneless compatible prior to the experimental release in v18.

alfaproject:
Alright, so having only OnPush components, and signals, basically, right? I suppose you'll publish some tips/guides

atscott:
https://angular.dev/guide/experimental/zoneless#requirements-for-zoneless-compatibility

OnPush, no NgZone.on[Stable/MicrotaskEmpty]. Test with the exhaustive checkNoChanges. It’s surprisingly straightforward. There should be minimal toil for apps written with OnPush. For those that aren’t, that step alone is 90% of the work and 99% of the difficulty.

alfaproject:
Yup, awesome, thanks!

jkrems:
Sorry for that, I should've landed this before the other lockfile update. Can you back out the pnpm-lock.yaml change and/or regenerate?

atscott:
@jkrems Done

PR #30036, opened 2025-04-05
Update `.browserslistrc` to use Baseline

dgp1130:
This changes the default `.browserslistrc` file to one generated from a Baseline "widely available" date using `baseline-browser-mapping`, codifying the policy change from https://github.com/angular/angular/pull/60754 into the Angular CLI.

This effectively pins the `.browserslistrc` file to a hard-coded Baseline date, preventing it from automatically updating based on `caniuse-lite` versions managed in user workspaces, which should make Angular's tooling much more stable. Prior to this, expressions like "last 2 Chrome versions" were relative to the installed `caniuse-lite` version and could update at any time. By pinning `.browserslistrc`, we're ensuring that supported browsers only change during Angular major version updates, not arbitrary `caniuse-lite` updates.

Angular CLI now warns if users define a custom `.browserslistrc` which includes browsers beyond Angular's support policy, however we do not error or prevent it, given that Angular is pretty stable as a framework and generally does not _require_ newly released web features, so targeting older browser versions is very likely to work. However we still warn in this case because it is explicitly outside Angular's support policy and is not _guaranteed_ to work.

`baseline-browser-mapping` is used as a dev dependency which generates a `.browserslistrc` that is vendored in the NPM package. This reduces proliferation of the dependency where not needed. Shout out to @tonypconway for building `baseline-browser-mapping`, which makes managing this configuration significantly easier.

See individual commits for notes on the implementation.

https://github.com/ng-packagr/ng-packagr/pull/3034 makes a similar update for `ng-packagr`.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -145,3 +145,34 @@ will block the next weekly release.
     accept the invite for the new package.
 
 Once Wombat accepts the invite, regular automated releases should work as expected.
+
+## Updating Browser Support
+
+Angular's browser support is defined by a [Baseline](https://web.dev/baseline)
+"widely available" date. Before a new major version is released, this should be
+updated to approximately the current date.
+
+A few weeks before a major (around feature freeze):
+
+1.  Update `BASELINE_DATE` in
+    [`/constants.bzl`](/constants.bzl) to the end of the most recent month.
+    - For example, if it is currently May 12th, set `baselineThreshold` to April
+      30th.
+    - Picking a date at the end of a month makes it easier to cross-reference
+      Angular's support with other tools (like MDN) which state Baseline support
+      using month specificity.
+    - You can view the generated `browserlist` configuration with:
+      ```shell
+      bazel build //packages/angular/build:angular_browserslist
+      cat dist/bin/packages/angular/build/.browserslistrc
+      ```
+    - Commit and merge the change, no other alterations or automation is
+      necessary in the CLI repo.
+2.  Update
+    [`/.browserslistrc`](https://github.com/ng-packagr/ng-packagr/tree/main/.browserslistrc)
+    in the `ng-packagr` repo.
+    - Use the generated configuration from above.
+3.  Update
+    [`angular.dev` documentation](https://github.com/angular/angular/tree/main/adev/src/content/reference/versions.md#browser-support)
+    to specify the date used and link to [browsersl.ist](https://browsersl.ist)
+    with the generated configuration.
--- a/packages/angular/build/src/utils/supported-browsers.ts
+++ b/packages/angular/build/src/utils/supported-browsers.ts
@@ -12,15 +12,16 @@ export function getSupportedBrowsers(
   projectRoot: string,
   logger: { warn(message: string): void },
 ): string[] {
-  browserslist.defaults = [
-    'last 2 Chrome versions',
-    'last 1 Firefox version',
-    'last 2 Edge major versions',
-    'last 2 Safari major versions',
-    'last 2 iOS major versions',
-    'last 2 Android major versions',
-    'Firefox ESR',
-  ];
+  // Read the browserslist configuration containing Angular's browser support policy.
+  const angularBrowserslist = browserslist(undefined, {
+    path: require.resolve('../../.browserslistrc'),
+  });
+
+  // Use Angular's configuration as the default.
+  browserslist.defaults = angularBrowserslist;
+
+  // Get the minimum set of browser versions supported by Angular.
+  const minimumBrowsers = new Set(angularBrowserslist);
 
   // Get browsers from config or default.
   const browsersFromConfigOrDefault = new Set(browserslist(undefined, { path: projectRoot }));
@@ -28,19 +29,32 @@ export function getSupportedBrowsers(
   // Get browsers that support ES6 modules.
   const browsersThatSupportEs6 = new Set(browserslist('supports es6-module'));
 
+  const nonEs6Browsers: string[] = [];
   const unsupportedBrowsers: string[] = [];
   for (const browser of browsersFromConfigOrDefault) {
     if (!browsersThatSupportEs6.has(browser)) {
+      // Any browser which does not support ES6 is explicitly ignored, as Angular will not build successfully.
       browsersFromConfigOrDefault.delete(browser);
+      nonEs6Browsers.push(browser);
+    } else if (!minimumBrowsers.has(browser)) {
+      // Any other unsupported browser we will attempt to use, but provide no support for.
       unsupportedBrowsers.push(browser);
     }
+  }
+
+  if (nonEs6Browsers.length) {
+    logger.warn(
+      `One or more browsers which are configured in the project's Browserslist configuration ` +
+        'will be ignored as ES5 output is not supported by the Angular CLI.\n' +
+        `Ignored browsers:\n${nonEs6Browsers.join(', ')}`,
+    );
   }
 
   if (unsupportedBrowsers.length) {
     logger.warn(
       `One or more browsers which are configured in the project's Browserslist configuration ` +
-        'will be ignored as ES5 output is not supported by the Angular CLI.\n' +
-        `Ignored browsers: ${unsupportedBrowsers.join(', ')}`,
+        "fall outside Angular's browser support for this version.\n" +
+        `Unsupported browsers:\n${unsupportedBrowsers.join(', ')}`,
     );
   }
 
--- a/packages/schematics/angular/BUILD.bazel
+++ b/packages/schematics/angular/BUILD.bazel
@@ -4,8 +4,10 @@
 # found in the LICENSE file at https://angular.dev/license
 
 load("@npm2//:defs.bzl", "npm_link_all_packages")
+load("//:constants.bzl", "BASELINE_DATE")
 load("//tools:defaults.bzl", "copy_to_bin", "jasmine_test", "npm_package", "ts_project")
 load("//tools:ts_json_schema.bzl", "ts_json_schema")
+load("//tools/baseline_browserslist:baseline_browserslist.bzl", "baseline_browserslist")
 
 licenses(["notice"])
 
@@ -42,11 +44,18 @@ copy_to_bin(
     srcs = glob(["**/schema.json"]),
 )
 
+baseline_browserslist(
+    name = "angular_browserslist",
+    out = "config/.browserslistrc",
+    baseline = BASELINE_DATE,
+)
+
 RUNTIME_ASSETS = [
     "collection.json",
     "migrations/migration-collection.json",
     "package.json",
     "utility/latest-versions/package.json",
+    ":angular_browserslist",
 ] + glob(
     include = [
         "*/schema.json",
--- a/packages/schematics/angular/config/index.ts
+++ b/packages/schematics/angular/config/index.ts
@@ -17,6 +17,7 @@ import {
   strings,
   url,
 } from '@angular-devkit/schematics';
+import { readFile } from 'node:fs/promises';
 import { posix as path } from 'node:path';
 import { relativePathToWorkspaceRoot } from '../utility/paths';
 import { getWorkspace as readWorkspace, updateWorkspace } from '../utility/workspace';
@@ -42,10 +43,13 @@ function addBrowserslistConfig(options: ConfigOptions): Rule {
       throw new SchematicsException(`Project name "${options.project}" doesn't not exist.`);
     }
 
+    // Read Angular's default vendored `.browserslistrc` file.
+    const config = await readFile(path.join(__dirname, '.browserslistrc'), 'utf8');
+
     return mergeWith(
       apply(url('./files'), [
         filter((p) => p.endsWith('.browserslistrc.template')),
-        applyTemplates({}),
+        applyTemplates({ config }),
         move(project.root),
       ]),
     );
--- a/tools/baseline_browserslist/package.json
+++ b/tools/baseline_browserslist/package.json
@@ -1,0 +1,6 @@
+{
+  "type": "module",
+  "devDependencies": {
+    "baseline-browser-mapping": "^2.2.0"
+  }
+}
--- a/.bazelignore
+++ b/.bazelignore
@@ -17,4 +17,5 @@ packages/ngtools/webpack/node_modules
 packages/schematics/angular/node_modules
 modules/testing/builder/node_modules
 tests/node_modules
+tools/baseline_browserslist/node_modules
 tools/legacy-rnjs/node_modules
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -169,6 +169,7 @@ npm_translate_lock(
         "//packages/ngtools/webpack:package.json",
         "//packages/schematics/angular:package.json",
         "//tests:package.json",
+        "//tools/baseline_browserslist:package.json",
     ],
     lifecycle_hooks_envs = {
         # TODO: Standardize browser management for `rules_js`
--- a/constants.bzl
+++ b/constants.bzl
@@ -8,6 +8,13 @@ ANGULAR_FW_VERSION = "^20.0.0-next.0"
 ANGULAR_FW_PEER_DEP = "^20.0.0 || ^20.0.0-next.0"
 NG_PACKAGR_PEER_DEP = "^20.0.0 || ^20.0.0-next.0"
 
+# Baseline widely-available date in `YYYY-MM-DD` format which defines Angular's
+# browser support. This date serves as the source of truth for the Angular CLI's
+# default browser set used to determine what downleveling is necessary.
+#
+# See: https://web.dev/baseline
+BASELINE_DATE = "2025-03-31"
+
 SNAPSHOT_REPOS = {
     "@angular/cli": "angular/cli-builds",
     "@angular/pwa": "angular/angular-pwa-builds",
--- a/modules/testing/builder/projects/hello-world-app/.browserslistrc
+++ b/modules/testing/builder/projects/hello-world-app/.browserslistrc
@@ -1,4 +1,0 @@
-# We want to run tests large with ever green browser so that 
-# we never trigger differential loading as this will slow down the tests.
-
-last 2 Chrome versions
--- a/packages/angular/build/BUILD.bazel
+++ b/packages/angular/build/BUILD.bazel
@@ -1,7 +1,9 @@
 load("@devinfra//bazel/api-golden:index_rjs.bzl", "api_golden_test_npm_package")
 load("@npm2//:defs.bzl", "npm_link_all_packages")
+load("//:constants.bzl", "BASELINE_DATE")
 load("//tools:defaults.bzl", "copy_to_bin", "jasmine_test", "npm_package", "ts_project")
 load("//tools:ts_json_schema.bzl", "ts_json_schema")
+load("//tools/baseline_browserslist:baseline_browserslist.bzl", "baseline_browserslist")
 
 licenses(["notice"])
 
@@ -39,6 +41,12 @@ copy_to_bin(
     srcs = glob(["**/schema.json"]),
 )
 
+baseline_browserslist(
+    name = "angular_browserslist",
+    out = ".browserslistrc",
+    baseline = BASELINE_DATE,
+)
+
 RUNTIME_ASSETS = glob(
     include = [
         "src/**/schema.json",
@@ -49,6 +57,7 @@ RUNTIME_ASSETS = glob(
 ) + [
     "builders.json",
     "package.json",
+    ":angular_browserslist",
 ]
 
 ts_project(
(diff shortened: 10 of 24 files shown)


docs/process/release.md
```diff
@@ -145,3 +145,24 @@ will block the next weekly release.
     accept the invite for the new package.
 
 Once Wombat accepts the invite, regular automated releases should work as expected.
+
```
alan-agius4, line 148: Overall, this looks good. But I’m wondering—what was the reasoning behind integrating Browserslist generation using `bl2bl` in the build pipeline, rather than simply extending the Browserslist configuration to achieve the same result?

**.browserslistrc**  
```
extends browserslist-config-baseline/2025
```

See: https://github.com/browserslist/browserslist?tab=readme-ov-file#popular-configs

alan-agius4 (commented):
Overall, this looks good to me although I am curious why we used this approach instead of extended the browserslist configuration.

Note: we also need to do the same in ng-packagr.

patches/bl2bl.patch
```diff
@@ -0,0 +1,21 @@
...
+   // Behaviour varies depending on whether userBrowserslistrc=true or false
++  const originallyHadBrowserslistrc = Boolean(packageJSON["browserslist"]);
+   if (!bl2blConfig.useBrowserslistrc) {
+     // if false, add baselineVersions to packageJSON object for later
+     packageJSON["browserslist"] = browserslistOutput;
+@@ -149,7 +150,7 @@ import("baseline-browser-mapping").then((bbm) => {
+   }
+ 
+   // Whatever happens, update package.json as long as the packageJSON object isn't null
+-  if (packageJSON != null) {
```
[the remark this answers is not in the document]
dgp1130: I'm planning to follow up with a PR to upstream this fix at minimum.

package.json
```diff
@@ -173,9 +174,18 @@
...
       }
+    },
+    "patchedDependencies": {
+      "bl2bl": "patches/bl2bl.patch"
     }
   },
   "resolutions": {
     "typescript": "5.8.2"
+  },
+  "bl2bl": {
```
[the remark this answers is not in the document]
dgp1130: Ah sorry, local testing with the tool kept inserting this, must've missed it before pushing.

packages/angular/build/src/browserslist/package.json
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
dgp1130: Yeah, it's definitely awkward in a Bazel context. It makes more sense in a typical web project, it's just that's not really Angular CLI.

If we continue to use `bl2bl`, I'd prefer to move to something like `bl2bl --widely-available-date YYYY-MM-DD > $(OUT)` and remove the `package.json` altogether, but we'd need to build alignment on the right mental model for `bl2bl`.

Thanks for the Renovate callout, updated our configuration.

packages/schematics/angular/BUILD.bazel
```diff
@@ -42,11 +42,20 @@ copy_to_bin(
     srcs = glob(["**/schema.json"]),
 )
 
+# Copy the generated browserslist file into this package.
+genrule(
+    name = "angular_browserslist",
+    srcs = ["//packages/angular/build:angular_browserslist"],
+    outs = [".browserslistrc"],
```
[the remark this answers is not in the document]
dgp1130: Generated it under `packages/schematics/angular/config/.browserslistrc`, which I think is what you meant?

docs/process/release.md
```diff
@@ -145,3 +145,24 @@ will block the next weekly release.
...
+1.  Update `baselineThreshold` in
+    [`browserslist/package.json`](/packages/angular/build/src/browserslist/package.json) to the
+    end of the most recent month.
+    - For example, if it is currently May 12th, set `baselineThreshold` to April 30th.
+    - Picking a date at the end of a month makes it easier to cross-reference Angular's support
+      with other tools (like MDN) which state Baseline support using month specificity.
+    - Commit and merge the change, no other alterations or automation is necessary in the CLI
+      repo.
+2.  Update
+    [`angular.dev` documentation](https://github.com/angular/angular/tree/main/adev/src/content/reference/versions.md#browser-support)
```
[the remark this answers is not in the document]
dgp1130: I updated https://github.com/angular/angular/pull/60754 to remove the hard-coded documentation, so this won't need to be refreshed for every major version.

packages/schematics/angular/BUILD.bazel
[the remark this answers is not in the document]
alan-agius4: Yes

alan-agius4:
We could adopt `browserslist-config-baseline`, but the main challenge is configuring its options, particularly `widelyAvailableOnDate`. In a monorepo, you might have multiple projects with different `.browserslistrc` files but only one `package.json`. Configuration is resolved based on the current working directory, not the browserslist config location itself, which complicates setup. Relying on CWD isn’t ideal, as it might point deep within a project rather than the monorepo root.

Looking at the code, both `bl2bl` and `browserslist-config-baseline` use `baseline-browser-mapping`. Ultimately, we only need the output of `baseline-browser-mapping` plus a small transformation.

```ts
import bbm from 'baseline-browser-mapping';

const browsers = {
  chrome: "Chrome",
  chrome_android: "ChromeAndroid",
  edge: "Edge",
  firefox: "Firefox",
  firefox_android: "FirefoxAndroid",
  safari: "Safari",
  safari_ios: "iOS",
};

const baselineBrowserslistConfig = bbm.getCompatibleVersions({
    widelyAvailableOnDate: '2025-01-01',
    includeDownstreamBrowsers: false,
    logConfigToConsole: false,
  })
  .filter((version) => browsers[version.browser])
  .map((version) => `${browsers[version.browser]} >= ${version.version}`);
```

This approach can be used with `ng build` by calling the API directly.

For `ng generate config`, we could follow the same strategy. While @dgp1130 potential concerns about module resolution, it's worth noting that `@schematics/angular` uses a custom resolution strategy that resolves modules from the repo root, so issues are unlikely.

We could also use the above to generate a static configuration. However, it might not be worth the extra complexities in the build process.

tonypconway:
If the end goal is absolutely a static `.browserslistrc` file, then yes, you could very easily import `baseline-browser-mapping` directly and do the transform as above.  A few things I'd like to point out:

1. If you're importing as an ES module, you can access the function directly, i.e.
  ```javascript
  import { getCompatibleVersions } from `baseline-browser-mapping`;
  ```
  We have to use `@http-toolkit/esm` to shim commonjs behaviour for `browserslist-config-baseline` because `browserslist` doesn't support ES modules for the `extends` functionality.
2. Is there a reason you're excluding downstream browsers?  As @notpushkin mentioned in conversation with @thesmiler , the coverage you see in browsersl.ist will be stronger if you include them, especially for anyone building in China where QQ and UC are very popular.  Samsung Internet is also widely used in a lot of markets.  If you decide to include them, don't forget to map the MDN `browser-compat-data` names for them to the `browserslist` names, [as we've done here](https://github.com/web-platform-dx/browserslist-config-baseline/blob/7b74014ab70a7ad8a6204edbaff307a2f433afdf/scripts/get-baseline-versions.js#L5-L18).
3. `logConfigToConsole` is a `browserslist-config-baseline` option, not `baseline-browser-mapping`, so you can leave it out if you're talking to `baseline-browser-mapping` directly.
4. Is there a reason you need to generate a `.browserslistrc` file, rather than e.g. using the `extends` syntax?  Could you instead expose `browserslist-config-baseline` in the generated `package.json` directly so that users know they can change it, and give them a warning if they adopt something other than your default?  The next release of `browserslist-config-baseline` will include warnings if the user has conflicting configuration options in the same `package.json`.

tonypconway:
FYI, we've just released the latest version of `browserslist-config-baseline` which supports the `widelyAvailableOnDate` config option.  Appreciate generating your own `.browserslistrc` file via `baseline-browser-mapping` may be the right option for you, but if you want to expose the option to end users, it's there.

dgp1130:
> Looking at the code, both `bl2bl` and `browserslist-config-baseline` use `baseline-browser-mapping`. Ultimately, we only need the output of `baseline-browser-mapping` plus a small transformation.

I started looking into `baseline-browser-mapping` instead of `bl2bl`. I think that can work, but for some reason I'm hitting a weird Bazel dependency issue when trying to load it in a tool. I'll need to do more investigation to see what's going on there.

> Is there a reason you're excluding downstream browsers?

My reasoning for this is because AIUI (feel free to correct me), a strict interpretation of Baseline does _not_ cover downstream browsers. Therefore Angular's support policy would not cover them and I'd like the implementation here to be as close to the stated policy as reasonably possible.

I'm certainly open to including those or having the CLI slightly diverge from stated policy if there's a compelling reason to do so. My immediate thinking is that it doesn't actually matter for the CLI if we're assuming that downstream browsers are functionally equivalent to their upstream versions (not sure how entirely true that is, but likely _mostly_ true?). If so, then adding QQ or UC wouldn't actually change the behavior of Angular CLI's build process, since an equivalent upstream browser would already be present.

> Is there a reason you need to generate a `.browserslistrc` file, rather than e.g. using the `extends` syntax?

This is a good point, I like the idea of `ng generate config browserslist` generating `extends browserslist-config-baseline` since that's semantically what Angular CLI is doing (pinned to a Baseline widely available date) and I'd like the generated config to be as close as we can get to what the CLI is doing by default.

Updated the schematic to generate this. I ended up dropping `.browserslistrc` altogether and moving it into `package.json`. Personally I'm not a huge fan of configuring things in `package.json` and would prefer a separate `.browserslistrc` file, but I opted to go this direction because we need to configure `browserslist-config-baseline` in the `package.json` as well, and I think keeping those two configs as close as possible is desirable.

That also means that this config needs to install `browserslist-config-baseline`, which makes it a little less reliable (`npm install` likes to run into conflicts). But there's a pretty small list of dependencies involved here and I think it's unlikely to be a problem in practice.

packages/schematics/angular/config/index.ts
```diff
@@ -35,20 +41,42 @@ export default function (options: ConfigOptions): Rule {
...
+    // Read Angular's default vendored `.browserslistrc` file.
+    const config = JSON.parse(
+      await fs.readFile(path.join(__dirname, 'baseline/package.json'), 'utf8'),
+    );
+    const widelyAvailableOnDate = config.bl2bl.baselineThreshold as string;
+
+    const pkgJson = new JSONFile(host, 'package.json');
+    pkgJson.modify(
+      ['browserslist'],
+      ['extends browserslist-config-baseline'],
```
alan-agius4, line 60: As mentioned in my [previous comment](https://github.com/angular/angular-cli/pull/30036#issuecomment-2785573211), this approach isn't ideal for an Angular monorepo. It can lead to inconsistent behavior, as the configuration may not always be detected due to changes in the current working directory during project builds.

dgp1130:
Resolved the build issues and updated to use `baseline-browser-mapping` instead of `bl2bl` as that works a bit cleaner with Bazel.

packages/schematics/angular/config/index.ts
dgp1130: Oh I think I understand. So you're saying we can't generate a dependency on `browserslist-config-baseline` because it requires `package.json` to configure it's date and in a monorepo, individual projects might not have their own independent `package.json` files.

I imagine ideally we'd want to generate a `.browserslistrc` file with syntax like `extends browserslist-config-baseline({ widelyAvailableDate: '2025-03-31' })`, but without that kind of support from `browserslist` we don't have a ton of options, hence the need for a `package.json` configuration.

The alternative I can think of would be to generate a separate `browserslist-config-baseline.json` file with that configuration which the package accepts, this way it can be decoupled from `package.json` and we can freely generate that file in the project source directory without breaking anything else. But it feels kind of heavyweight to create a new file just for this.

I'm thinking fow now, I can revert back to using the `.browserslistrc` configuration we generated at build time and not using `browserslist-config-baseline` at all. Then, if you and/or `browserslist` are able to find a more ergonomic way of defining the Baseline date, we can potentially update Angular CLI to generate that instead. @tonypconway, WDYT?

alan-agius4: No, it’s not just that—the configuration for `browserslist-config-baseline` is resolved using the current working directory (`process.cwd()`), as shown https://github.com/web-platform-dx/browserslist-config-baseline/blob/c741ccb7694b2dc39f5f467d653b16092afa22a1/scripts/get-baseline-versions.js#L44. This means the build output can vary depending on the CWD from which the project is built from and creating a nested `package.json` won't resolve this inconsistency.

**Example:**

```
$ ng build foo                # Uses the package.json from the monorepo root
$ cd project/foo && ng build  # Fails to pick up the root package.json
```

packages/angular/build/tools/baseline_browserslist/generate_browserslist.mts
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { getCompatibleVersions } from 'baseline-browser-mapping';
+
+// Map `baseline-browser-mapping` browsers to `browserslist` browsers.
+const browsers = new Map(
```
alan-agius4, line 12: OOC: why not use a Record? seems like there is no real reason why we need a Map here.

packages/angular/build/tools/baseline_browserslist/generate_browserslist.mts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export function generateBrowserslist(date: string): string {
+  // Generate a `browserslist` configuration.
+  return getCompatibleVersions({
+    widelyAvailableOnDate: date,
+    includeDownstreamBrowsers: false,
+  })
+    .map(({ browser, version }) => ({ browser: browsers.get(browser), version }))
+    .filter(({ browser }) => Boolean(browser))
+    .map(({ browser, version }) => `${browser} >= ${version}`)
```
alan-agius4, line 39: NIT: the initial `.map` can be removed.

```suggestion
    .filter(({ browser }) => browsers.has(browsers))
    .map(({ browser, version }) => `${browsers.has(browser)} >= ${version}`)
```

packages/angular/build/src/utils/supported-browsers.ts
```diff
@@ -7,40 +7,55 @@
  */
 
 import browserslist from 'browserslist';
+import * as path from 'node:path';
```
alan-agius4, line 10: NIT: 
```suggestion
import { join } from 'node:path';
```

packages/angular/build/tools/baseline_browserslist/baseline_browserslist.bzl
```diff
@@ -0,0 +1,26 @@
+"""Generates a `browserslist` configuration from a Baseline date."""
```
alan-agius4, line 1: From what I’ve seen, we typically don’t create nested tools directories. In my experience, other repos also follow this patter tools are placed in the root-level tools directory. Also, since this might be needed in `@schematics/angular`, it could make more sense to move it there instead.

Example: https://github.com/angular/angular-cli/tree/main/tools

//cc @devversion maybe this nested approach is now preferred

packages/angular/build/tools/baseline_browserslist/tsconfig-build.json
```diff
@@ -0,0 +1,6 @@
+{
```
alan-agius4, line 1: Rename the file to `tsconfig.json` so that the IDE can also pick it up as otherwise the config in the root tsconfig are applied when opening the nested files.

packages/angular/build/tools/baseline_browserslist/BUILD.bazel
```diff
@@ -0,0 +1,53 @@
...
+)
+
+ts_project(
+    name = "generate_browserslist_test_lib",
+    testonly = True,
+    srcs = ["generate_browserslist_spec.mts"],
+    tsconfig = "//:test-tsconfig-esm",
+    deps = [
+        ":generate_browserslist",
+        "//:node_modules/@types/jasmine",
```
alan-agius4, line 37: Is this really needed?

alan-agius4, line 37: Redundant

```suggestion
```

packages/angular/build/src/utils/supported-browsers.ts
```diff
@@ -7,40 +7,55 @@
...
-    'last 1 Firefox version',
-    'last 2 Edge major versions',
-    'last 2 Safari major versions',
-    'last 2 iOS major versions',
-    'last 2 Android major versions',
-    'Firefox ESR',
-  ];
+  // Read the browserslist configuration containing Angular's browser support policy.
+  const angularBrowserslist = browserslist(undefined, {
+    path: path.join(require.resolve('../../.browserslistrc')),
```
alan-agius4, line 18:
```suggestion
    path: require.resolve('../../.browserslistrc'),
```

packages/schematics/angular/config/index.ts
dgp1130: Ah yeah, even resolving `package.json` wouldn't work if it's done relative to the CWD, since that directory might not be the project root. I feel like the "correct" answer here would be for `browserslist-config-baseline` to resolve the `package.json` relative to the `.browserslistrc` file which has `extends browserslist-config-baseline`, but I'm not sure if that's possible with `browserslist` itself.

For now, I think I'll revert to inlining browser versions and avoid `browserslist-config-baseline` altogether. If we're able to find a way to make this work in a monorepo use case, I'd be open to switching over to that.

packages/angular/build/tools/baseline_browserslist/baseline_browserslist.bzl
dgp1130: Yeah, I went back and forth with whether to put this in `@angular/build` or in the greater workspace.

I think what I ultimately landed on was that the Baseline date and its generated `browserslist` are Angular-wide configurations not specific to any package (hence why `BASELINE_DATE` ended up in `/constants.bzl` now). So I think it follows to do the same for the tool. I like the idea of a tool-specific `package.json` file. Moved to `//tools/baseline_browserslist/...`.

I opted to just have each package run the tool themselves rather than find a single location for generated `.browserslist` and manually copy it into each package.

packages/angular/build/tools/baseline_browserslist/BUILD.bazel
dgp1130: Huh, why _isn't_ this necessary? I don't see anything which implicitly adds `//:node_modules/@types/jasmine` as a dependency...

dgp1130:
Strict deps test failures seem to be a bug in the upstream infra: https://github.com/angular/dev-infra/pull/2738.

packages/angular/build/tools/baseline_browserslist/BUILD.bazel
devversion: it's the tsconfig

packages/angular/build/tools/baseline_browserslist/BUILD.bazel
```diff
@@ -0,0 +1,53 @@
...
+load("//tools:defaults2.bzl", "jasmine_test", "ts_project")
+
+js_binary(
+    name = "baseline_browserslist",
+    data = [":baseline_browserslist_lib"],
+    entry_point = "index.mjs",
+    visibility = ["//packages/angular/build:__subpackages__"],
+)
+
+ts_project(
```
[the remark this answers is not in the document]
alan-agius4: I think it's a bit of overkill to create two separate projects for just two files especially when one file depends on the other.

packages/schematics/angular/config/index.ts
```diff
@@ -17,6 +17,7 @@ import {
   strings,
   url,
 } from '@angular-devkit/schematics';
+import { promises as fs } from 'node:fs';
```
alan-agius4, line 20: NIT: 

```suggestion
import { readFile } from 'node:fs/promises';
```

alan-agius4 (approved):
Thanks for this @dgp1130, happy to see that we finally got there. Good work!

Just two optional NITs. 

- Reduce ts_projects
- Node.js Promise imports.

tools/baseline_browserslist/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@angular/baseline-browserslist",
+  "version": "0.0.0-PLACEHOLDER",
```
alan-agius4, line 3: This is not a shippable package. These fields are redundant and confusing.

```suggestion
```

packages/angular/build/tools/baseline_browserslist/BUILD.bazel
dgp1130: Interesting. Personally I think the `ts_project` _should_ have a dependency on Jasmine given that it contains the actual `describe`, `it`, etc. which actually relies on Jasmine typings. I feel like leaking the `tsconfig` deps into its targets is a strict deps violation.

Not super important for something as simple as this, just an observation.

dgp1130:
> * Remove extra fields from package.json

Done.

> * Reduce ts_projects

Done.

> * Node.js Promise imports.

Done.

> * Update renovate config to include the new package.json

I think it's already covered by the [`**/package.json` glob](https://github.com/angular/angular-cli/blob/95d16dc52113a1d5f67c95a5f6d82e5e937f299c/renovate.json#L21). I don't see anything that would exclude `tools/baseline_browserslist/package.json`. Let me know if I'm misreading this.

alan-agius4 (approved):

PR #30052, opened 2025-04-07
fix(@schematics/angular): default component templates to not use `.ng.html` extension

clydin:
To align with recent changes to the updated style guide, Angular v20 will now generate component templates without a `.ng.html` file extension and instead use the previous `.html` by default. The `--ng-html` option is still available for use if the `.ng.html` extension is preferred in projects.


packages/schematics/angular/component/schema.json
```diff
@@ -137,7 +137,7 @@
     },
     "ngHtml": {
       "type": "boolean",
-      "default": true,
+      "default": false,
       "description": "Generate component template files with an '.ng.html' file extension instead of '.html'. The '.ng.html' file extension is recommended by the Angular style guide."
```
JoostK, line 141: This sentence may also need updating

alan-agius4 (approved):

PR #30072, opened 2025-04-09
fix(@angular/ssr): support `getPrerenderParams` for wildcard routes

### LINKED ISSUE #30035: Cannot pre-render wildcard routes when using `outputMode: 'server'`
### Command

build

### Is this a regression?

_No response_

### Description

I have a few wildcard routes in my app. There are a large number of routes which will match these wildcard routes and they are all listed in a route file as described in [the docs](https://angular.dev/guide/prerendering#prerendering-parameterized-routes).
```json
"outputMode": "server",
"prerender": {
	"routesFile": "renderableRoutes.txt",
	"discoverRoutes": true
},
```
However when I build, I get the warning `The "prerender" option is not considered when "outputMode" is specified.` and the routes specified in `renderableRoutes.txt` are not prerendered. Removing `outputMode` causes the routes to be prerendered, but I get the error `Error: Angular app engine manifest is not set. Please ensure you are using the '@angular/build:application' builder to build your server application.` when running the server.

### Minimal Reproduction

- Add `routesFile` to `angular.json`
- Remove `outputMode` from `angular.json`
- Build
- Serve SSR

### Exception or Error

```text
Error: Angular app engine manifest is not set. Please ensure you are using the '@angular/build:application' builder to build your server application
```

### Your Environment

```text
Angular CLI: 19.2.6
Node: 18.20.7
Package Manager: npm 10.8.2
OS: linux x64

Angular: 19.2.5
... animations, common, compiler, compiler-cli, core, elements
... service-worker

@angular-devkit/architect       0.1902.6
@angular-devkit/build-angular   19.2.6
@angular-devkit/core            19.2.6
@angular-devkit/schematics      19.2.6
@angular/cdk                    19.2.8
@angular/cli                    19.2.6
@angular/pwa                    19.2.6
@angular/ssr                    19.2.6
@schematics/angular             19.2.6
ng-packagr                      19.0.1
```

### Anything else relevant?

I am using the `@angular-devkit/build-angular:application` builder.

alan-agius4:
Handle `getPrerenderParams` return values when used with wildcard route paths, including support for combined routes like `/product/:id/**`.
Supports returning an array of path segments (e.g., `['category', '123']`) for `**` routes and dynamic segments combined with catch-all routes.

This enables more flexible prerendering configurations in server routes, including handling specific paths such as `/product/1/laptop/123`.

Example:
```ts
{
  path: '/product/:id/**',
  renderMode: RenderMode.Prerender,
  async getPrerenderParams() {
    return [
      { id: '1', '**': 'laptop/123' },
      { id: '2', '**': 'laptop/456' }
    ];
  }
}
```

Closes #30035

### DIFF
--- a/packages/angular/ssr/src/routes/ng-routes.ts
+++ b/packages/angular/ssr/src/routes/ng-routes.ts
@@ -47,6 +47,12 @@ interface Route extends AngularRoute {
 const MODULE_PRELOAD_MAX = 10;
 
 /**
+ * Regular expression to match a catch-all route pattern in a URL path,
+ * specifically one that ends with '/**'.
+ */
+const CATCH_ALL_REGEXP = /\/(\*\*)$/;
+
+/**
  * Regular expression to match segments preceded by a colon in a string.
  */
 const URL_PARAMETER_REGEXP = /(?<!\\):([^/]+)/g;
@@ -391,7 +397,11 @@ async function* handleSSGRoute(
     meta.redirectTo = resolveRedirectTo(currentRoutePath, redirectTo);
   }
 
-  if (!URL_PARAMETER_REGEXP.test(currentRoutePath)) {
+  const isCatchAllRoute = CATCH_ALL_REGEXP.test(currentRoutePath);
+  if (
+    (isCatchAllRoute && !getPrerenderParams) ||
+    (!isCatchAllRoute && !URL_PARAMETER_REGEXP.test(currentRoutePath))
+  ) {
     // Route has no parameters
     yield {
       ...meta,
@@ -415,7 +425,9 @@ async function* handleSSGRoute(
 
     if (serverConfigRouteTree) {
       // Automatically resolve dynamic parameters for nested routes.
-      const catchAllRoutePath = joinUrlParts(currentRoutePath, '**');
+      const catchAllRoutePath = isCatchAllRoute
+        ? currentRoutePath
+        : joinUrlParts(currentRoutePath, '**');
       const match = serverConfigRouteTree.match(catchAllRoutePath);
       if (match && match.renderMode === RenderMode.Prerender && !('getPrerenderParams' in match)) {
         serverConfigRouteTree.insert(catchAllRoutePath, {
@@ -429,20 +441,10 @@ async function* handleSSGRoute(
     const parameters = await runInInjectionContext(parentInjector, () => getPrerenderParams());
     try {
       for (const params of parameters) {
-        const routeWithResolvedParams = currentRoutePath.replace(URL_PARAMETER_REGEXP, (match) => {
-          const parameterName = match.slice(1);
-          const value = params[parameterName];
-          if (typeof value !== 'string') {
-            throw new Error(
-              `The 'getPrerenderParams' function defined for the '${stripLeadingSlash(currentRoutePath)}' route ` +
-                `returned a non-string value for parameter '${parameterName}'. ` +
-                `Please make sure the 'getPrerenderParams' function returns values for all parameters ` +
-                'specified in this route.',
-            );
-          }
-
-          return value;
-        });
+        const replacer = handlePrerenderParamsReplacement(params, currentRoutePath);
+        const routeWithResolvedParams = currentRoutePath
+          .replace(URL_PARAMETER_REGEXP, replacer)
+          .replace(CATCH_ALL_REGEXP, replacer);
 
         yield {
           ...meta,
@@ -471,6 +473,34 @@ async function* handleSSGRoute(
       renderMode: fallback === PrerenderFallback.Client ? RenderMode.Client : RenderMode.Server,
     };
   }
+}
+
+/**
+ * Creates a replacer function used for substituting parameter placeholders in a route path
+ * with their corresponding values provided in the `params` object.
+ *
+ * @param params - An object mapping parameter names to their string values.
+ * @param currentRoutePath - The current route path, used for constructing error messages.
+ * @returns A function that replaces a matched parameter placeholder (e.g., ':id') with its corresponding value.
+ */
+function handlePrerenderParamsReplacement(
+  params: Record<string, string>,
+  currentRoutePath: string,
+): (substring: string, ...args: unknown[]) => string {
+  return (match) => {
+    const parameterName = match.slice(1);
+    const value = params[parameterName];
+    if (typeof value !== 'string') {
+      throw new Error(
+        `The 'getPrerenderParams' function defined for the '${stripLeadingSlash(currentRoutePath)}' route ` +
+          `returned a non-string value for parameter '${parameterName}'. ` +
+          `Please make sure the 'getPrerenderParams' function returns values for all parameters ` +
+          'specified in this route.',
+      );
+    }
+
+    return parameterName === '**' ? `/${value}` : value;
+  };
 }
 
 /**
@@ -530,9 +560,9 @@ function buildServerConfigRouteTree({ routes, appShellRoute }: ServerRoutesConfi
       continue;
     }
 
-    if (path.includes('*') && 'getPrerenderParams' in metadata) {
+    if ('getPrerenderParams' in metadata && (path.includes('/*/') || path.endsWith('/*'))) {
       errors.push(
-        `Invalid '${path}' route configuration: 'getPrerenderParams' cannot be used with a '*' or '**' route.`,
+        `Invalid '${path}' route configuration: 'getPrerenderParams' cannot be used with a '*' route.`,
       );
       continue;
     }
--- a/packages/angular/ssr/src/routes/route-config.ts
+++ b/packages/angular/ssr/src/routes/route-config.ts
@@ -142,6 +142,10 @@ export interface ServerRoutePrerenderWithParams extends Omit<ServerRoutePrerende
    * A function that returns a Promise resolving to an array of objects, each representing a route path with URL parameters.
    * This function runs in the injector context, allowing access to Angular services and dependencies.
    *
+   * It also works for catch-all routes (e.g., `/**`), where the parameter name will be `**` and the return value will be
+   * the segments of the path, such as `/foo/bar`. These routes can also be combined, e.g., `/product/:id/**`,
+   * where both a parameterized segment (`:id`) and a catch-all segment (`**`) can be used together to handle more complex paths.
+   *
    * @returns A Promise resolving to an array where each element is an object with string keys (representing URL parameter names)
    * and string values (representing the corresponding values for those parameters in the route path).
    *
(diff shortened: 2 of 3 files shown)


goldens/public-api/angular/ssr/index.api.md
```diff
@@ -63,7 +63,7 @@ export interface ServerRoutePrerender extends Omit<ServerRouteCommon, 'status'>
 // @public
 export interface ServerRoutePrerenderWithParams extends Omit<ServerRoutePrerender, 'fallback'> {
     fallback?: PrerenderFallback;
-    getPrerenderParams: () => Promise<Record<string, string>[]>;
+    getPrerenderParams: () => Promise<(string[] | Record<string, string>)[]>;
```
alan-agius4, line 66: Another option would be to add a different option something like `getPrerenderPath` or `getPrerenderSegments`

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -429,20 +435,38 @@ async function* handleSSGRoute(
...
+          throw new Error(
+            `The 'getPrerenderParams' function for the '${stripLeadingSlash(currentRoutePath)}' ` +
+              `route returned an object '${JSON.stringify(params)}', which is not valid for parameterized routes.`,
+          );
+        }
 
-          return value;
-        });
+        const routeWithResolvedParams = isParamsArray
+          ? currentRoutePath.replace('**', params.join('/'))
```
dgp1130, line 455: Consider: What if I render `['foo/bar']`?

I remember we had a discussion on this previously, do we need to escape `/` characters returned by `getPrerenderParams`? Or did we decide not to do that?

goldens/public-api/angular/ssr/index.api.md
dgp1130: IMHO, reusing `getPrerenderParams` is reasonable. I think the intent behind this function is still the same whether it's filling in `:foo` or `**`. Though I'm a little worried that supporting the two in an _xor_ fashion might be confusing for developers. If we can find a reasonable way to support `:foo/**`, then using a single function LGTM.

packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -530,9 +554,9 @@ function buildServerConfigRouteTree({ routes, appShellRoute }: ServerRoutesConfi
       continue;
     }
 
-    if (path.includes('*') && 'getPrerenderParams' in metadata) {
+    if ('getPrerenderParams' in metadata && !path.endsWith('/**') && path.includes('*')) {
```
dgp1130, line 557: Bug: Does this handle `/foo/*/bar/**` correctly? I think it would pass but should emit an error?

packages/angular/ssr/test/routes/ng-routes_spec.ts
```diff
@@ -296,6 +276,44 @@ describe('extractRoutesAndCreateRouteTree', () => {
...
+      expect(routeTree.toObject()).toEqual([
+        { route: '/home', renderMode: RenderMode.Server },
+        { route: '/user/joe/role/admin', renderMode: RenderMode.Prerender },
+        {
+          route: '/user/jane/role/writer',
+          renderMode: RenderMode.Prerender,
+        },
+        { route: '/user/**', renderMode: RenderMode.Server },
+      ]);
+    });
```
dgp1130, line 315: Consider: Should we test the error cases as well?

dgp1130 (commented):
What about a path like `'/product/:first/**/:second'`? Not sure if we allow segments after `**` (reading this PR, seems like no?), but presumably we at least allow them before? Is this feature limited to either `/:first` *xor* `/**`?

I feel like we could support both with something like:

```typescript
{
  path: '/product/:first/**/:second',
  renderMode: RenderMode.Prerender,
  async getPrerenderParams() {
    return [
      {'first': 'foo', __spread__: ['category', '1'], 'second': 'bar'},
      {'first': 'bar', __spread__: ['category', '2'], 'second': 'foo'},
    ];
  }
}
```

Not sure exactly the right way of identifying the spread array though. Maybe a Symbol imported from `@angular/ssr`?

Is there a better way of mixing `**` and `:foo` syntax in the same path? Is `:foo/**` a rare enough pattern that we just don't care at this stage?

alan-agius4:
> What about a path like `'/product/:first/**/:second'`? Not sure if we allow segments after `**` (reading this PR, seems like no?), but presumably we at least allow them before? Is this feature limited to either `/:first` _xor_ `/**`?
> 
> I feel like we could support both with something like:
> 
> ```ts
> {
>   path: '/product/:first/**/:second',
>   renderMode: RenderMode.Prerender,
>   async getPrerenderParams() {
>     return [
>       {'first': 'foo', __spread__: ['category', '1'], 'second': 'bar'},
>       {'first': 'bar', __spread__: ['category', '2'], 'second': 'foo'},
>     ];
>   }
> }
> ```
> 
> Not sure exactly the right way of identifying the spread array though. Maybe a Symbol imported from `@angular/ssr`?
> 
> Is there a better way of mixing `**` and `:foo` syntax in the same path? Is `:foo/**` a rare enough pattern that we just don't care at this stage?

`/product/:first/**/:second` is not a valid Angular router route. Also, we did not allow this before.

packages/angular/ssr/test/routes/ng-routes_spec.ts
alan-agius4: Added

dgp1130:
> Thinking a bit more, `/product/:first/**` would be a valid route, but in that case, the route can also be defined as as `/product/**`.

What about `/product/:first/title/**`? I don't think we can generally assume you can always collapse the `:params` into `**`?

packages/angular/ssr/src/routes/ng-routes.ts
alan-agius4: No we are not escaping any character. `foo/bar` will be rendered it's a valid URL.

alan-agius4:
> > Thinking a bit more, `/product/:first/**` would be a valid route, but in that case, the route can also be defined as as `/product/**`.
> 
> What about `/product/:first/owner/**`? I don't think we can generally assume you can always collapse the `:params` into `**`?

Yes, this is not handled at the moment, maybe the simplest would be to allow something like

```ts
{
  path: '/product/:first/**',
  renderMode: RenderMode.Prerender,
  async getPrerenderParams() {
    return [
      {'first': 'foo', '**': 'category/1', 'second': 'bar'},
    ];
  }
}
```

dgp1130:
Would it be worth getting some wider feedback from the team (Andrew or Jan in particular)? Maybe a quick one-pager would be enough to get some quick thoughts on the approach.

I'm open to `{'**': 'foo/bar'}`, but also can see that being a little awkward. Presumably `/foo/:**/bar` is not allowed as a path so there's no ambiguity (if not, it probably should be banned)?

jkrems:
I like `**` as a key because it's easy to remember. More generally, I'm in favor of an API that is consistently either a `Record` or an array, especially because of the mix case. Replacing `/prod/:id/**` with `/prod/**` removes the route param as a readable property on the activated route (or from component inputs if enabled), so it's not just an "free" change.

Afaict there's no pre-existing name for the value captured from the wildcard.

alan-agius4:
@atscott, any thoughts?

dgp1130:
@AndrewKushnir might also have some quick thoughts here.

AndrewKushnir:
> @AndrewKushnir might also have some quick thoughts here.

> Thinking a bit more, /product/:first/** would be a valid route, but in that case, the route can also be defined as as /product/**.

I'm wondering what are the use-cases for this feature? What would be "hiding" behind the `**` symbol: would we only have static paths or dynamic segments (like `:first`). My thinking is that allowing dynamic segments might be confusing, since in the `getPrerenderParams()` function we'd need to return a set of values for keys that are not present in the URL, so as a developer I'd need to reconcile this with corresponding routes in the main routing file (also it'd be a challenge to keep things in sync).

Potentially, we can apply the following rules:

- `**` can only be present at the end of the URL pattern
- `**` can only be capturing static segments (i.e. a pattern like `products/**` is not valid for `products/:id`), so there is always a direct mapping between dynamic segments and corresponding values returned from the `getPrerenderParams()` function

However, I'm not sure if that would be too limiting for the use-cases that this feature is improving.

alan-agius4:
> I'm wondering what are the use-cases for this feature? 

One example is when catch-all routes need to be prerendered. For instance, see this issue: https://github.com/angular/angular-cli/issues/30035. In such cases, parameterized routes can't be used effectively because the nesting depth is arbitrary.

As for the rules you mentioned, those sound good to me, and I don't see them as restrictive.

alan-agius4:
@dgp1130 PTAL

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -158,6 +162,16 @@ export interface ServerRoutePrerenderWithParams extends Omit<ServerRoutePrerende
    *       return ids.map(id => ({ id })); // Generates paths like: [{ id: '1' }, { id: '2' }, { id: '3' }]
    *     },
    *   },
+   *   {
+   *     path: '/product/:id/**',
+   *     renderMode: RenderMode.Prerender,
+   *     async getPrerenderParams() {
+   *       return [
+   *         { id: '1', '**': '/laptop/123' },
+   *         { id: '2', '**': '/laptop/456' }
```
dgp1130, line 171: Nit: Is the leading slash necessary? If not, I'd probably leave it off given that this is replacing a `**` which is already preceded by a slash.

packages/angular/ssr/src/routes/route-config.ts
```diff
@@ -158,6 +162,16 @@ export interface ServerRoutePrerenderWithParams extends Omit<ServerRoutePrerende
...
+   *       ]; // Generates paths like: [{ id: '1', '**': '/laptop/123' }, { id: '2', '**': '/laptop/456' }]
```
dgp1130, line 172: Typo: Did you mean:

```suggestion
   *       ]; // Generates paths like: `/product/1/laptop/123`, `/product/2/laptop/456`
```

packages/angular/ssr/test/routes/ng-routes_spec.ts
```diff
@@ -296,6 +276,44 @@ describe('extractRoutesAndCreateRouteTree', () => {
       ]);
     });
 
+    it('should resolve catch all routes for SSG and not add a fallback route if fallback is Server', async () => {
```
dgp1130, line 279: Question: Why shouldn't this add a fallback route if the fallback is `Server`? Looking at the assertion, it looks to me like it does have a fallback based on `{ route: '/user/*/**', renderMode: RenderMode.Server }`?

dgp1130 (approved):

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: Not it is not

packages/angular/ssr/src/routes/route-config.ts
alan-agius4: yes

packages/angular/ssr/test/routes/ng-routes_spec.ts
alan-agius4: Typo

PR #30074, opened 2025-04-09
feat(@schematics/angular): generate applications using TypeScript project references

clydin:
When generating a project (either via `ng new` or `ng generate application`), the created TypeScript configuration files (`tsconfig.app.json`/`tsconfig.spec.json`) will be setup as composite projects and added as project references to in the root `tsconfig.json`.  This transforms the root `tsconfig.json` into a "solution" style configuration. This allows IDEs to more accurately discover and provide type information for the varying types of files (test, application, etc.) within each project. The Angular build process is otherwise unaffected by these changes.


alan-agius4 (approved):
LGTM, we should do the same for the library schematic :)

PR #30092, opened 2025-04-11
fix(@angular/build): include `module` value check when adding custom conditions

alan-agius4:
Previously, when setting `module: preserve`—which internally sets `moduleResolution` to `bundler` in TypeScript—custom conditions were not applied. This change ensures that custom conditions are also added when `module` is set to `preserve`.


jkrems (approved):

clydin (approved):

dgp1130:

PR #30097, opened 2025-04-14
fix(@angular/build): prevent nested CSS in components

alan-agius4:
The Angular encapsulation currently does not support CSS nesting syntax, which can lead to run-time errors or unexpected behavior when such styles are used in component stylesheets. This change ensures that nested CSS rules flattened to maintain compatibility with the compiler.

For more context, see: https://github.com/angular/angular/issues/58996


jkrems (approved):

alan-agius4:

PR #30107, opened 2025-04-15
refactor(@angular/build): use `ngHmrMode` define for Vite prebundling

clydin:
The internal `ngHmrMode` define value was previously available to application code and third-party modules that were not prebundled. To ensure that the value is propagated to all relevant code served by the development server, the `ngHmrMode` define is now used during the Vite prebundling phase. This allows any prebundled dependencies to also leverage the `ngHmrMode` value. The framework may, for instance, use the value to provide additional development time diagnostics.


alan-agius4 (approved):

clydin:

PR #30113, opened 2025-04-16
fix(@angular/build): pass `preserveSymlinks` option to Karma esbuild builder

alan-agius4:
Ensure that several previously omitted options are correctly passed to the Karma esbuild builder, improving consistency and expected behavior.


clydin:

PR #30115, opened 2025-04-16
refactor: minor code changes in karma `AngularAssetsMiddleware`

alan-agius4:
Early exits and reduce optional chaining.

### DIFF
--- a/packages/angular/build/src/builders/karma/application_builder.ts
+++ b/packages/angular/build/src/builders/karma/application_builder.ts
@@ -67,32 +67,30 @@ class AngularAssetsMiddleware {
   ) {}
 
   handle(req: IncomingMessage, res: ServerResponse, next: (err?: unknown) => unknown) {
-    let err = null;
-    try {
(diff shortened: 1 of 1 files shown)


jkrems (approved):
This does now have the potential to call next multiple times in case of errors. But that might be okay..?

(The code invoked by next() may throw which then could get caught by this function and trigger a 2nd call to next().)

alan-agius4:
> This does now have the potential to call next multiple times in case of errors. But that might be okay..?
> 
> (The code invoked by next() may throw which then could get caught by this function and trigger a 2nd call to next().)

Yes, that should be fine though, what was the intent of the try/catch was it to catch error from `this.serveFile` or handle invalid `new URL`?

jkrems:
> Yes, that should be fine though, what was the intent of the try/catch was it to catch error from this.serveFile or handle invalid new URL?

That's a good point! Since this entirely synchronous, we could likely depend on the framework to catch errors and remove the try/catch.

1979Allen:
_dud$

PR #30130, opened 2025-04-18
feat(@angular/build): add experimental vitest unit-testing support

clydin:
When using the application build system via the `@angular/build` package (default for new projects starting in v20), a new experimental unit-test builder is available that initially uses vitest. This experimental system is intended to provide support for investigation of future unit testing efforts within the Angular CLI. As this is experimental, no SemVer guarantees are provided, the API and behavior may change, and there may be unexpected behavior. Available test runners may be added or removed as well.

The setup is somewhat different than the previous unit-testing builders. It uses a similar mechanism to that of the `dev-server` and requires a `buildTarget` option. This allows the code building aspects of the unit-testing process to leverage pre-existing option values that are already defined for development. If differing option values are required for testing, an additional build target configuration specifically for testing can be used.

The current vitest support has multiple caveats including but not limited to:

* No watch support
* `jsdom` based testing only (`jsdom` must be installed in the project)
* Custom vitest configuration is not supported

An example configuration that would replace the `test` target for a project is as follows:
```
"test": {
    "builder": "@angular/build:unit-test",
    "options": {
        "tsConfig": "tsconfig.spec.json",
        "buildTarget": "::development",
        "runner": "vitest"
    }
}
```

### DIFF
--- a/packages/angular/build/builders.json
+++ b/packages/angular/build/builders.json
@@ -24,6 +24,11 @@
       "implementation": "./src/builders/ng-packagr/index",
       "schema": "./src/builders/ng-packagr/schema.json",
       "description": "Build a library with ng-packagr."
+    },
+    "unit-test": {
+      "implementation": "./src/builders/unit-test",
+      "schema": "./src/builders/unit-test/schema.json",
+      "description": "[EXPERIMENTAL] Run application unit tests."
     }
   }
 }
--- a/packages/angular/build/src/builders/unit-test/builder.ts
+++ b/packages/angular/build/src/builders/unit-test/builder.ts
@@ -1,0 +1,171 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import type { BuilderContext, BuilderOutput } from '@angular-devkit/architect';
+import assert from 'node:assert';
+import { randomUUID } from 'node:crypto';
+import path from 'node:path';
+import { createVirtualModulePlugin } from '../../tools/esbuild/virtual-module-plugin';
+import { loadEsmModule } from '../../utils/load-esm';
+import { buildApplicationInternal } from '../application';
+import type {
+  ApplicationBuilderExtensions,
+  ApplicationBuilderInternalOptions,
+} from '../application/options';
+import { ResultKind } from '../application/results';
+import { OutputHashing } from '../application/schema';
+import { writeTestFiles } from '../karma/application_builder';
+import { findTests, getTestEntrypoints } from '../karma/find-tests';
+import { normalizeOptions } from './options';
+import type { Schema as UnitTestOptions } from './schema';
+
+export type { UnitTestOptions };
+
+/**
+ * @experimental Direct usage of this function is considered experimental.
+ */
+export async function* execute(
+  options: UnitTestOptions,
+  context: BuilderContext,
+  extensions: ApplicationBuilderExtensions = {},
+): AsyncIterable<BuilderOutput> {
+  // Determine project name from builder context target
+  const projectName = context.target?.project;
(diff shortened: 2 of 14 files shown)


packages/angular/build/builders.json
```diff
@@ -24,6 +24,11 @@
       "implementation": "./src/builders/ng-packagr/index",
       "schema": "./src/builders/ng-packagr/schema.json",
       "description": "Build a library with ng-packagr."
+    },
+    "unit-test": {
```
alan-agius4, line 28: Please also update https://github.com/angular/angular-cli/blob/main/packages/angular/cli/lib/config/workspace-schema.json

packages/angular/build/src/builders/unit-test/schema.json
```diff
@@ -0,0 +1,82 @@
...
+        "type": "string"
+      },
+      "default": [],
+      "description": "Globs of files to exclude, relative to the project root."
+    },
+    "watch": {
+      "type": "boolean",
+      "description": "Run build when files change."
+    },
+    "browsers": {
```
alan-agius4, line 45: Looks like this is unused?

packages/angular/build/src/builders/unit-test/schema.json
```diff
@@ -0,0 +1,82 @@
...
+      "description": "Override which browsers tests are run against. Set to `false` to not use any browser.",
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "A comma seperate list of browsers to run tests against."
```
alan-agius4, line 50: Instead of comma separated, I'd prefer an array

packages/angular/build/src/builders/unit-test/schema.json
```diff
@@ -0,0 +1,82 @@
...
+    "tsConfig": {
+      "type": "string",
+      "description": "The name of the TypeScript configuration file."
+    },
+    "runner": {
+      "type": "string",
+      "description": "The name of the test runner to use for test execution.",
+      "enum": ["vitest"]
+    },
+    "runnerConfig": {
```
alan-agius4, line 21: Do we intend to expose the config? As currently this option is unused.

packages/angular/build/src/builders/unit-test/builder.ts
```diff
@@ -0,0 +1,157 @@
...
+      continue;
+    } else if (result.kind !== ResultKind.Full) {
+      assert.fail('A full build result is required from the application builder.');
+    }
+
+    assert(result.files, 'Builder did not provide result files.');
+
+    await writeTestFiles(result.files, outputPath);
+
+    const setupFiles = ['init-testbed.js'];
```
alan-agius4, line 126: Consider using the virtual module.

packages/angular/build/src/builders/unit-test/builder.ts
```diff
@@ -0,0 +1,157 @@
...
+  if (testFiles.length === 0) {
+    context.logger.error('No tests found.');
+
+    return { success: false };
+  }
+
+  const entryPoints = getTestEntrypoints(testFiles, { projectSourceRoot, workspaceRoot });
+  entryPoints.set(
+    'init-testbed',
+    path.join(__dirname, '..', 'karma', 'polyfills', 'init_test_bed.js'),
```
alan-agius4, line 76: This file has been removed.

alan-agius4 (changes requested):
Overall looks good, just some minor things.

packages/angular/build/src/builders/unit-test/schema.json
clydin: Removed for now.

packages/angular/build/src/builders/unit-test/schema.json
clydin: Removed the option for now.

alan-agius4 (approved):

alan-agius4 (approved):

PR #30136, opened 2025-04-21
build: use npm_package instead of js_library for the schematics npm package

josephperrott:


alan-agius4:
@josephperrott, is this something you are still working on or can this be closed?

josephperrott:
Lets close it for now and we can revisit during our larger cleanups it it makes sense.

PR #30145, opened 2025-04-22
fix(@angular/build): use virtual module for Karma TestBed initialization

clydin:
Using a virtual module for the Karma-based unit testing setup allows for the Angular packages used within the TestBed initialization to resolve to the same dependencies used within the application and test code. This avoids the potential for multiple copies of `@angular/core` within the test output which can result in TestBed execution failure due to the module level variable usage within TestBed.


jkrems (approved):

alan-agius4 (approved):

alan-agius4:
Note: I tried this and this will not solve the tests when run with `--preserve-symlinks=true`, however I am going to merge this as it's a cleaner approach.

PR #30148, opened 2025-04-22
feat(@schematics/angular): generate libraries using TypeScript project references

clydin:
When generating a project (via `ng generate library`), the created TypeScript configuration files (`tsconfig.lib.json`/`tsconfig.spec.json`) will be setup as composite projects and added as project references in the root `tsconfig.json`.  This transforms the root `tsconfig.json` into a "solution" style configuration. This allows IDEs to more accurately discover and provide type information for the varying types of files (test, library, etc.) within each project. The Angular build process is otherwise unaffected by these changes.


packages/schematics/angular/library/files/tsconfig.lib.json.template
```diff
@@ -9,6 +9,9 @@
     "inlineSources": true,
     "types": []
   },
+  "include": [
+    "**/*.ts"
```
alan-agius4, line 13: More explicate and allows users to add secondary entry-points without needing to update this tsconfig.

```suggestion
    "src/**/*.ts"
```

packages/schematics/angular/library/files/tsconfig.spec.json.template
```diff
@@ -9,7 +9,6 @@
     ]
   },
   "include": [
-    "**/*.spec.ts",
-    "**/*.d.ts"
+    "**/*.ts"
```
alan-agius4, line 12:
```suggestion
    "src/**/*.ts"
```

alan-agius4 (approved):
LGTM, just a nit

alan-agius4 (approved):

PR #30192, opened 2025-04-28
fix(@schematics/angular): remove explicit index option from new applications

clydin:
With the index option now defaulting to `<project_source_root>/index.html` for the `application` build system, the explicit value present in new applications is no longer required. This removal further reduces the size of the `angular.json` file for new projects.


alan-agius4 (approved):

clydin:

PR #30208, opened 2025-04-30
refactor(@angular-devkit/core): additional isolated declarations cleanup

clydin:
To eventually support the use of the TypeScript `isolatedDeclarations` option, additional type adjustments have been made to code within the `@angular-devkit/core` package. While this is not yet comprehensive, it significantly reduces the amount of build errors with the `isolatedDeclarations` option enabled.

### DIFF
--- a/packages/angular_devkit/core/src/virtual-fs/host/memory.ts
+++ b/packages/angular_devkit/core/src/virtual-fs/host/memory.ts
@@ -36,15 +36,16 @@ import {
 
 export interface SimpleMemoryHostStats {
   readonly content: FileBuffer | null;
+  inspect(): string;
 }
 
 export class SimpleMemoryHost implements Host<{}> {
-  protected _cache = new Map<Path, Stats<SimpleMemoryHostStats>>();
+  protected _cache: Map<Path, Stats<SimpleMemoryHostStats>> = new Map();
   private _watchers = new Map<Path, [HostWatchOptions, Subject<HostWatchEvent>][]>();
 
-  protected _newDirStats() {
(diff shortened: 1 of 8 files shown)


packages/angular_devkit/core/src/virtual-fs/host/memory.ts
```diff
@@ -36,15 +36,16 @@ import {
 
 export interface SimpleMemoryHostStats {
   readonly content: FileBuffer | null;
+  inspect(): string;
 }
 
 export class SimpleMemoryHost implements Host<{}> {
-  protected _cache = new Map<Path, Stats<SimpleMemoryHostStats>>();
+  protected _cache: Map<Path, Stats<SimpleMemoryHostStats>> = new Map();
   private _watchers = new Map<Path, [HostWatchOptions, Subject<HostWatchEvent>][]>();
```
jkrems, line 44: I assume this wasn't updated because it's private and won't show up in typings? Should we favor style consistency over the strict requirements (no strong preference either way, just wondering)?

jkrems (approved):

packages/angular_devkit/core/src/virtual-fs/host/memory.ts
clydin: Yes, since that is private it's not an issue from an isolated declarations perspective.
Good question, though. For this PR, I went through each error and fixed individually. Since that was not an error, it was not modified.  In general, I do favor consistency. However, we would probably need some form of lint check to both catch all these cases and prevent backsliding.  This may be a good followup task.

PR #30226, opened 2025-05-01
refactor(@angular-devkit/core): don't use concrete type as interface

jkrems:
`Set` isn't a stable interface, it's a concrete type that changes over time (new spec versions may add methods, older ones may lack methods). There shouldn't be a huge downside to just declaring this class as "its own thing".

Follow-up to https://github.com/angular/angular-cli/pull/30208 which breaks on google3 import because the `Set` types are subtly different because of different TS configs.


clydin (approved):

jkrems:

PR #30238, opened 2025-05-03
build: improve stability of windows jobs further

devversion:
With the recent commit to rework the convert symlink script, we
addressed the high flakiness of module resolution errors due to
broken/missing symlinks. Though, this commit introduces a new source of
flakiness that we tried to avoid in the past— running too many WSL exe's
at the same time. We fix this by reducing the batch size and adding
another retry for safety.

In addition, the patch branch uses an older NodeJS version where
`fs.cp` seems to have issues when the source files are readonly— so we
fix that by just invoking the shell's `cp` command. That one works.


crisbeto (approved):

devversion:

PR #30240, opened 2025-05-04
fix(@angular/build): allow a default application `browser` option

clydin:
The application build system's `browser` option is now considered optional. If not present, the value will be an `main.ts` file within the configured project source root (`sourceRoot`). This change allows the removal of the `browser` option from any project that uses the default generated value.


clydin:

PR #30248, opened 2025-05-05
fix(@angular/build) allow component HMR for templates with i18n

### LINKED ISSUE #29457: Tracking: Support i18n component updates (HMR)
Templates that contain i18n and/or manual `$localize` usage should support component updates via the development server HMR.

If possible, translation file changes should be considered as well. Although this may need to be implemented as a separate feature.

clydin:
When using the development server with the `application` build system and HMR has been enabled (default), component templates with i18n are now eligible to be hot reloaded. If translations exist within the template, they will also be translated assuming a matching translation is available. Changing the content of an i18n block may result in a missing translation warning/error. The development server continues to only support a single enabled locale.

Closes #29457


clydin:

PR #30257, opened 2025-05-06
fix(@angular/cli): add Node.js 24 as supported version

alan-agius4:
Node.js 24 has been released https://nodejs.org/en/about/previous-releases


clydin (approved):

alan-agius4:

PR #30269, opened 2025-05-07
fix(@angular-devkit/core): use more precise iterator typings

jkrems:
Since this class isn't a "real" `Set`, we can't promise that our iterators are true `SetIterator<T>`s.

Follow-up to https://github.com/angular/angular-cli/pull/30226.


clydin:

PR #30271, opened 2025-05-07
fix(@angular-devkit/build-angular): update esbuild to 0.25.1

raiseandfall:
Addresses [https://github.com/advisories/GHSA-67mh-4wv8-2f99](https://github.com/advisories/GHSA-67mh-4wv8-2f99)


alan-agius4:
Hi @raiseandfall, thanks for your contribution. However, this change carries some risk, as it upgrades esbuild across several major versions. It's also important to note that the reported vulnerability does not affect Angular CLI, since the esbuild dev server is not used.

PR #30276, opened 2025-05-08
fix(@angular/build): allow TestBed provider configuration with vitest unit-testing

clydin:
The experimental `unit-test` builder with the `vitest` runner configured can now specify an Angular providers file via the `providersFile` option. This allows the TestBed initialization to use the providers defined in this file for all executed tests. The contents of the providers file should include a default export with an array of one or more Angular providers. As an example:
```
import { provideZonelessChangeDetection } from '@angular/core';

export default [
  provideZonelessChangeDetection(),
];

```

### DIFF
--- a/packages/angular/build/src/builders/unit-test/builder.ts
+++ b/packages/angular/build/src/builders/unit-test/builder.ts
@@ -142,15 +142,28 @@ export async function* execute(
     loadContent: async () => {
       const contents: string[] = [
         // Initialize the Angular testing environment
+        `import { NgModule } from '@angular/core';`,
         `import { getTestBed, ɵgetCleanupHook as getCleanupHook } from '@angular/core/testing';`,
         `import { BrowserTestingModule, platformBrowserTesting } from '@angular/platform-browser/testing';`,
         `import { beforeEach, afterEach } from 'vitest';`,
+        '',
+        normalizedOptions.providersFile
+          ? `import providers from './${path
+              .relative(projectSourceRoot, normalizedOptions.providersFile)
+              .replace(/.[mc]?ts$/, '')
+              .replace(/\\/g, '/')}'`
+          : 'const providers = [];',
         '',
         // Same as https://github.com/angular/angular/blob/05a03d3f975771bb59c7eefd37c01fa127ee2229/packages/core/testing/src/test_hooks.ts#L21-L29
         `beforeEach(getCleanupHook(false));`,
         `afterEach(getCleanupHook(true));`,
         '',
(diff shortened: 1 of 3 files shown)


jkrems (approved):
That's a really neat solution for this!

packages/angular/build/src/builders/unit-test/builder.ts
```diff
@@ -142,15 +142,25 @@ export async function* execute(
     loadContent: async () => {
       const contents: string[] = [
         // Initialize the Angular testing environment
+        `import { NgModule } from '@angular/core';`,
         `import { getTestBed, ɵgetCleanupHook as getCleanupHook } from '@angular/core/testing';`,
         `import { BrowserTestingModule, platformBrowserTesting } from '@angular/platform-browser/testing';`,
         `import { beforeEach, afterEach } from 'vitest';`,
         '',
+        normalizedOptions.providersFile
+          ? `import providers from './${path.relative(projectSourceRoot, normalizedOptions.providersFile)}'`
```
alan-agius4, line 151: A couple of notes:

1. `path.relative` will return backslashes on Windows, which will break the import paths.
2. The file extension (.ts) is retained. Can this cause issues? (I guess not as this is not typechecked).
3. Instead of using `projectSourceRoot`, it probably makes more sense to use `workspaceRoot`, since paths in `angular.json` are all relative to the workspace root.

We handle a similar case here for reference:
https://github.com/angular/angular-cli/blob/505c8368fcec85c34bf4561638eb30c8b6285bf8/packages/angular/build/src/tools/esbuild/application-code-bundle.ts#L748-L755

packages/angular/build/src/builders/unit-test/schema.json
```diff
@@ -65,6 +65,11 @@
       "items": {
         "type": "string"
       }
+    },
+    "providersFile": {
+      "type": "string",
+      "description": "TypeScript file that exports an array of Angular providers to use during test execution. The array must be a default export.",
+      "minLength": 1
```
alan-agius4, line 72: is `"minLength": 1` intentional?

alan-agius4 (changes requested):
See feedback

packages/angular/build/src/builders/unit-test/schema.json
clydin: yes. attempts to prevent an empty string as a value.

packages/angular/build/src/builders/unit-test/builder.ts
clydin: Added the extra cleanup logic from the linked code.

alan-agius4 (approved):

clydin:

PR #30293, opened 2025-05-12
fix(@angular/build): show unit-test error for missing vitest browser package

clydin:
When using the experimental `unit-test` builder with the `vitest` runner and the `browsers` option, an error will now be reported if the `@vitest/browser` package is not found. Previously, the tests would attempt to execute and produce a runtime error and stack trace. Now an explicit error with instructions to install the package will be printed to the console prior to building and executing the tests.


alan-agius4 (approved):

clydin:

PR #30297, opened 2025-05-13
fix(@angular/build): provide vitest globals in unit-test builder

clydin:
The experimental `unit-test` builder with the `vitest` runner now provides the test APIs also as globals. This increases the potential compatibility for existing karma/jasmine based tests. The initial unit tests within a newly generated application, for instance, now require no changes to execute with the new builder.


alan-agius4 (approved):

clydin:

PR #30299, opened 2025-05-13
fix(@angular/build): setup unit-test polyfills before TestBed init

clydin:
To ensure that Zone.js is fully setup and patched prior to the use of TestBed, the setup file order for the `vitest` runner of the `unit-test` builder has been swapped. This is particularly relevant for `fakeAsync` which has a module level statement that attempts to capture the `Zone` instance. If Zone.js is not setup at that point, fakeAsync will fail to function in tests.


alan-agius4 (approved):

alan-agius4 (approved):

clydin:

PR #30307, opened 2025-05-14
fix(@schematics/angular): support using default browser option when not present

clydin:
The Angular schematics that require knowledge of the main entry point of an application will now use the default value for the `application` builder. The default value as per the `application` build system is `<project_source_root>/main.ts`.


clydin:

PR #30318, opened 2025-05-15
fix(@angular/build): use global unit-test hooks during TestBed init

clydin:
To provide improved Zone.js support with the experimental `unit-test` builder using the `vitest` runner, the TestBed initialization now uses the global `beforeEach`/`afterEach` test hooks. The global hooks can be patched by Zone.js. The vitest globals are always available when using the `unit-test` builder.


jkrems (approved):

clydin:

PR #30320, opened 2025-05-15
fix(@angular/build): provide direct debugging support for unit test builder

clydin:
When using the experimental `unit-test` builder with the `vitest` runner, a `debug` option is now available which configures the underlying runner to setup and use the Node Inspector to debug tests. This reduces the complexity to support debugging unit tests via a command execution of `ng test --debug`. This will cause the test process to initialize and then wait for a debugger to connect prior to executing the tests.


jkrems (approved):

clydin:

PR #30323, opened 2025-05-15
fix(@angular/build): enable unit-test builder watch outside CI

clydin:
When using the experimental `unit-test` builder, the `watch` option will now default to true when there is an interactive terminal and not on CI. This more closely aligns with the behavior of the existing `karma` builder and that of vitest itself. CI is determined by the presence of a "truthy" `CI` environment variable.


jkrems (approved):

clydin:

PR #30329, opened 2025-05-16
refactor: disable chrome sandbox as it no longer works in CI

alan-agius4:
This commit disables the chrome sandbox


devversion (approved):
ty!

alan-agius4:

PR #30343, opened 2025-05-19
fix(@angular/build): exclude only source test files with unit-test vitest support

### LINKED ISSUE #30333: Vitest code coverage not properly reported
### Command

test

### Is this a regression?

_No response_

### Description

I am trying out the new experimental vitest builder, which works great for me in general! I just found an issue with the coverage reporting. After installing `@vitest/coverage-v8` and running `ng test --code-coverage` with the vitest builder, the coverage is not properly reported. The displayed coverage table appears empty:

Image

The XML, JSON and HTML reports located in `dist/test-out/**/coverage` are also empty. As there is actually something tested in `app.ts`, I would have expected something to appear in the coverage report.

### Minimal Reproduction

1. Generate a fresh project with `npx@angular/cli@next new` and switch to the unit-test builder, or check out the code from the reproduction repo: https://github.com/json-derulo/ng20-vitest
2. Run `ng test --code-coverage`
3. Take a look at the console output

### Exception or Error

```text
No exception or error visible, but the coverage is not reported
```

### Your Environment

```text
Angular CLI: 20.0.0-rc.2
Node: 22.15.1
Package Manager: pnpm 9.15.6
OS: darwin x64

Angular: 20.0.0-rc.1
... platform-browser, router

@angular-devkit/architect    0.2000.0-rc.2 (cli-only)
@angular-devkit/core         20.0.0-rc.2 (cli-only)
@angular-devkit/schematics   20.0.0-rc.2 (cli-only)
@angular/build               20.0.0-rc.2
@angular/cli                 20.0.0-rc.2
@schematics/angular          20.0.0-rc.2 (cli-only)
```

### Anything else relevant?

I also tried to configure the `development` configuration as the build target for the unit tests, but the same thing happens.

clydin:
When using the experimental `unit-test` builder with the `vitest` runner and code coverage, the automatic coverage excludes are now adjust to not include the bundled test files. This previously prevent accurate coverage analysis. Vite does this as a convenience but is problematic for the bundling strategy employed by the builder's test setup. To workaround this, the excludes are adjusted here to only automatically exclude the TypeScript source test files.

Closes #30333

### DIFF
--- a/packages/angular/build/src/builders/unit-test/builder.ts
+++ b/packages/angular/build/src/builders/unit-test/builder.ts
@@ -230,11 +230,25 @@ export async function* execute(
           reporters: normalizedOptions.reporters ?? ['default'],
           coverage: {
             enabled: normalizedOptions.codeCoverage,
-            exclude: normalizedOptions.codeCoverageExclude,
             excludeAfterRemap: true,
           },
           ...debugOptions,
         },
+        plugins: [
+          {
+            name: 'angular-coverage-exclude',
+            configureVitest(context) {
+              // Adjust coverage excludes to not include the otherwise automatically inserted included unit tests.
+              // Vite does this as a convenience but is problematic for the bundling strategy employed by the
+              // builder's test setup. To workaround this, the excludes are adjusted here to only automatically
+              // exclude the TypeScript source test files.
+              context.project.config.coverage.exclude = [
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):

alan-agius4:

PR #30349, opened 2025-05-20
fix(@angular/cli): update min Node.js support to 20.19, 22.12, and 24.0

alan-agius4:
Bumps the minimum Node.js support for `@angular/cli`. This update ensures compatibility with a growing number of upstream dependencies, such as Undici and Pacote, that are adopting newer Node.js versions. These versions where selected as `require(esm)` is enabled by default. This change also prepares us for future package releases that will likely rely on this behavior.


clydin (approved):

alan-agius4:

PR #30351, opened 2025-05-20
fix(@angular/build): ensure disabled vitest config loading

clydin:
With the experimental `unit-test` builder with the `vitest` runner, the underlying vitest execution will no longer attempt to load external configuration files. The current internal vitest usage does not support integration of external configuration files.


alan-agius4 (approved):

alan-agius4:

PR #30363, opened 2025-05-21
refactor(@angular/build): check that AOT is enabled in HMR logic

alan-agius4:
This commit refactors the HMR logic within `@angular/build` to explicitly check for AOT.


alan-agius4:

PR #30366, opened 2025-05-21
fix(@angular/build): include full metadata for AOT unit-testing

clydin:
Include class metadata and JIT information in code built for development and/or test usage. All non-optimized builds will now contain this metadata (`optimization.scripts` is false). The Angular TestBed APIs require additional metadata for the Angular aspects of the application such as Components, Modules, Pipes, etc.
TestBed may also leverage JIT capabilities during testing (e.g., `overrideComponent`). To support all these TestBed usage scenarios both the underlying `setClassMetadata` and `setNgModuleScope` class augmentation functions are now emitted when test metadata inclusion is enabled.


clydin:

PR #30372, opened 2025-05-22
refactor(@angular/build): update `ensureWorkerPool` to do an early exit

alan-agius4:
The avoid unnecessary logic to be executed for every call to `transformFile`.


clydin:

PR #30375, opened 2025-05-22
perf(@angular/build): directly check code for Angular partial linking

clydin:
To more completely avoid having to load any Angular compiler linking code when analyzing for code transformations during a build, the declaration function checks are now performed directly. This is in contrast to the previous behavior that required importing the linker code to use a helper function included in the `@angular/compiler-cli` package. Avoiding the need to load and parse a potentially large amount of code to perform the check can improve performance for the cases where no linking is required. To ensure the check stays up to date and while the list of function names rarely changes, the linker must always use the existing function name prefix (`ɵɵngDeclare`) for any declaration functions.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
+++ b/packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
@@ -27,6 +27,14 @@ interface JavaScriptTransformRequest {
 const textDecoder = new TextDecoder();
 const textEncoder = new TextEncoder();
 
+/**
+ * The function name prefix for all Angular partial compilation functions.
+ * Used to determine if linking of a JavaScript file is required.
+ * If any additional declarations are added or otherwise changed in the linker,
+ * the names MUST begin with this prefix.
+ */
+const LINKER_DECLARATION_PREFIX = 'ɵɵngDeclare';
+
 export default async function transformJavaScript(
   request: JavaScriptTransformRequest,
 ): Promise<unknown> {
@@ -45,11 +53,6 @@ export default async function transformJavaScript(
 let linkerPluginCreator:
   | typeof import('@angular/compiler-cli/linker/babel').createEs2015LinkerPlugin
   | undefined;
-
-/**
- * Cached instance of the compiler-cli linker's needsLinking function.
- */
-let needsLinking: typeof import('@angular/compiler-cli/linker').needsLinking | undefined;
 
 async function transformWithBabel(
   filename: string,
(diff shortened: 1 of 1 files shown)


packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
```diff
@@ -27,6 +27,25 @@ interface JavaScriptTransformRequest {
...
 const textEncoder = new TextEncoder();
 
+/**
+ * The current list of Angular partial compilation declaration function names.
+ * Used to determine if linking of a JavaScript file is required.
+ * If any additional declarations are added or otherwise changed, this list
+ * MUST be updated.
+ * The component declaration is first as it is the most likely to be present.
+ */
+const LINKER_DECLARATION_NAMES = [
```
JoostK, line 37: WDYT of reducing this to a single search for just `ɵɵngDeclare`, which I expect is sufficiently unique to avoid false positives and reduces the maintenance/syncing cost of having the check in the CLI.

packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
```diff
@@ -125,17 +128,10 @@ async function requiresLinking(path: string, source: string): Promise<boolean> {
...
-    // Once TypeScript provides support for keeping the dynamic import this workaround can be
-    // changed to a direct dynamic import.
-    const linkerModule = await loadEsmModule<typeof import('@angular/compiler-cli/linker')>(
-      '@angular/compiler-cli/linker',
-    );
-    needsLinking = linkerModule.needsLinking;
-  }
-
-  return needsLinking(path, source);
+  // Check if the source code includes one of the declaration functions.
```
alan-agius4, line 131: NOT: Can we also add a faster check to check that the file extension is `.mjs` ? All Angular libraries are `.mjs`. This could also simplify the regexp for core and compiler.

clydin: Third party angular libraries may not be though. I don't think APF mandates the file extension.

alan-agius4: Libraries that are compiled with ng-packagr is though.

clydin: Most probably use ng-packagr but i would be hesitant to assume all do.  i think it may be good to do as a separate change in case it causes unintentional breakage.

clydin: Also, maybe APF should mandate something like `.partial.mjs` for a more encompassing check?

alan-agius4 (approved):

packages/angular/build/src/tools/esbuild/javascript-transformer-worker.ts
devversion: IMO APF should always be partial compilation output, but since that is not necessarily the case for 1st party libraries, I could see adding a field to the `package.json`?

clydin: Walking up the file system to find a relevant `package.json` for each file could get expensive.
At a minimum something like `.partial.mjs` could provide a definitive signal that it needs to be processed and no additional check would be needed.  Other files could fallback to the existing checks.

devversion: Ah, I thought it would be easy to determine the package root, and have some caching.

In either case, I think it would be easy enough to introduce the `.partial` suffix

PR #30387, opened 2025-05-26
build: preserve `peerDependenciesMeta` in snapshots

alan-agius4:
Addresses an issue where publishing snapshot builds would replace the value with the package SHAs in `peerDependenciesMeta`.

See: https://github.com/angular/angular-build-builds/blob/d4d0f1ca0932bd4fe2df4cc2da10d132789dcd8f/package.json#L89


alan-agius4:
> My script was simpler in total I think, but maybe it didn't work?

Yes, I didn't work. But I had another look and it's was due to a typo. It works now :)

PR #30415, opened 2025-05-29
fix(@schematics/angular): remove karma config devkit package usages during application migration

### LINKED ISSUE #30413: after upgrade to 20 tests fail
### Command

update

### Is this a regression?

19

### Description

1. After running `ng update @angular/core@20 @angular/cli@20` 
2. Then `npm run test` I get error

```
 Error: Cannot find module '@angular-devkit/build-angular/plugins/karma'
```

### Minimal Reproduction

npm run test

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 20.0.0
Node: 22.15.1
Package Manager: npm 10.9.2
OS: win32 x64

Angular: 20.0.0
... animations, build, cli, common, compiler, compiler-cli, core
... forms, language-service, localize, platform-browser

@angular-devkit/architect    0.1902.13
@angular-devkit/core         19.2.13
@angular-devkit/schematics   19.2.13
@schematics/angular          20.0.0
ng-packagr                   20.0.0
```

### Anything else relevant?

I'm using karma.conf.js `frameworks: ['jasmine', '@angular-devkit/build-angular'],
`

clydin:
When performing an `ng update` to Angular v20, the application migration will detect the usage of the `karma` builder's `karmaConfig` option and attempt to remove usages of the no longer needed `@angular-devkit/build-angular` karma framework and plugin usage. While the karma framework usage will be specifically ignored when using the `@angular/build:karma` builder, the plugin usage leverages a direct `require` within the configuration file. Regardless of the ability of the builder to ignore, neither usage is needed with `@angular/build:karma` and removing the code aligns with what would be generated by `ng generate config karma`.

Closes #30413

### DIFF
--- a/packages/schematics/angular/migrations/use-application-builder/migration.ts
+++ b/packages/schematics/angular/migrations/use-application-builder/migration.ts
@@ -232,6 +232,8 @@ function updateProjects(tree: Tree, context: SchematicContext) {
 
     // Use @angular/build directly if there is no devkit package usage
     if (!hasAngularDevkitUsage) {
+      const karmaConfigFiles = new Set<string>();
+
       for (const [, target] of allWorkspaceTargets(workspace)) {
         switch (target.builder) {
           case Builders.Application:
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/migrations/use-application-builder/migration.ts
```diff
@@ -292,6 +300,35 @@ function updateProjects(tree: Tree, context: SchematicContext) {
...
+
+      for (const karmaConfigFile of karmaConfigFiles) {
+        if (!tree.exists(karmaConfigFile)) {
+          continue;
+        }
+
+        try {
+          const originalKarmaConfigText = tree.readText(karmaConfigFile);
+          const updatedKarmaConfigText = originalKarmaConfigText
+            .replaceAll(`'@angular-devkit/build-angular'`, '')
```
jkrems, line 312: This one might be worth documenting inline, it's not quite clear why removing this string literal is a safe modification (it might very well be because of the likely use in the array literal but would be good to call out as a comment).

jkrems (approved):

packages/schematics/angular/migrations/use-application-builder/migration.ts
clydin: I'd be fine dropping that specific replacement especially since we automatically handle that case.

I considered an AST based transform here but the complexity increased significantly.  We can reevaluate if needed based one feedback though.

clydin:

PR #30422, opened 2025-05-30
feat(@angular/build): add dataurl, base64 loaders

### LINKED ISSUE #30391: Add support for a data url file loader with application/esbuild builders
### Command

build

### Description

This describes 4 loader schemes: text, binary, file, and empty.

My use-case is to switch up the contents of the favicon, which I do by updating the href of the relevant `<link rel="icon">`. This is simplest way to do that is with a data url (updating it with a real URL would cause it to have to be loaded separately, which works, but requires an extra request).

With the webpack loader, this is easily achieved with e.g. `import favicon from "!!url-loader!favicon.ico";` which produces a string like `data:image/x-icon;base64,....`.

Add a "data-url" (or "url" or "data") loader with the following import parameters:
* `mime: string` (defaults to mime type based on extension)
* `plain: boolean` (defaults to false, controls whether base64 is enabled)

So e.g. one could do `import contents from 'foo.svg' with {loader: 'data-url', mime: 'foo/bar', plain: true}`

Perhaps better names for the parameters could be found, e.g. the mime type isn't technically a mime type and can take additional parameters, e.g. `text/plain;charset=utf-8` and so on.

Without this, the options seem to be to do this computation at runtime, e.g.:
* `URL.createObjectURL(new Blob([uint8array], {type: mimetype}))`
* `"data:" + mimetype + ";base64," + btoa(uint8array).replace(/\+/g, '-').replace(/\//g, '_')` (untested, but something like that should work)

This works, but why not have something in the framework that does the thing you want for you, and you don't have to do these fixups each time one wants to import such an asset.

imirkin:
This passes through the existing esbuild functionality to support dataurl and base64 loaders, in addition to the pre-existing text, binary, and file loaders.

Both the dataurl and base64 loaders return strings with the relevant content. There is no way to control whether the dataurl method uses base64 or plain text encoding, and no way to control the mime type it detects.

Fixes #30391

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes
- [ ] CI related changes

Currently only the text/binary/file esbuild loaders are usable.

Issue Number: #30391

## What is the new behavior?

This adds the dataurl/base64 loaders as well.

- [ ] Yes
- [x] No

## Other information

It wasn't clear to me how to cause the documentation at e.g. https://angular.dev/tools/cli/build-system-migration#file-extension-loader-customization to be updated. I updated the comments in schema.json.

### DIFF
--- a/packages/angular/build/src/builders/application/options.ts
+++ b/packages/angular/build/src/builders/application/options.ts
@@ -206,13 +206,22 @@ export async function normalizeOptions(
     }
   }
 
-  let loaderExtensions: Record<string, 'text' | 'binary' | 'file'> | undefined;
+  let loaderExtensions:
+    | Record<string, 'text' | 'binary' | 'file' | 'dataurl' | 'base64'>
+    | undefined;
   if (options.loader) {
     for (const [extension, value] of Object.entries(options.loader)) {
       if (extension[0] !== '.' || /\.[cm]?[jt]sx?$/.test(extension)) {
         continue;
       }
-      if (value !== 'text' && value !== 'binary' && value !== 'file' && value !== 'empty') {
+      if (
+        value !== 'text' &&
+        value !== 'binary' &&
+        value !== 'file' &&
+        value !== 'dataurl' &&
+        value !== 'base64' &&
+        value !== 'empty'
+      ) {
         continue;
       }
       loaderExtensions ??= {};
--- a/packages/angular/build/src/builders/application/schema.json
+++ b/packages/angular/build/src/builders/application/schema.json
@@ -279,10 +279,10 @@
       ]
     },
     "loader": {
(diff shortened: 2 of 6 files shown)


imirkin:
Let me know if you want me to fix the lint issue. Prettier wants some IMO questionable formatting:

```ts
  let loaderExtensions:
    | Record<string, 'text' | 'binary' | 'file' | 'dataurl' | 'base64'>
    | undefined;
```

and

```ts
      if (
        value !== 'text' &&
        value !== 'binary' &&
        value !== 'file' &&
        value !== 'dataurl' &&
        value !== 'base64' &&
        value !== 'empty'
      ) {
        continue;
      }
```

But if that's the project style, happy to include.

dgp1130:
Thanks for the contribution, great addition to the CLI!

PR #30457, opened 2025-06-04
fix(@angular/cli): improve Node.js version check and error messages

alan-agius4:
Enhances the Node.js version validation to include Node.js 22.12


alan-agius4:

PR #30470, opened 2025-06-05
build: improve globbing of jasmine test files

devversion:
This doesn't seem to surface here, but in the framework and in the COMP repo, the glob wasn't catching all files because it wasn't escaped— so didn't actually get picked up by Jasmine, but was expanded by Bash.

This improves the rule and makes this logic more future-proof.

### DIFF
--- a/tools/defaults.bzl
+++ b/tools/defaults.bzl
@@ -65,9 +65,10 @@ def jasmine_test(data = [], args = [], **kwargs):
         chdir = native.package_name(),
         args = [
             "--require=%s/node_modules/source-map-support/register.js" % relative_to_root,
-            "**/*spec.js",
-            "**/*spec.mjs",
-            "**/*spec.cjs",
+            # Escape so that the `js_binary` launcher triggers Bash expansion.
+            "'**/*+(.|_)spec.js'",
+            "'**/*+(.|_)spec.mjs'",
+            "'**/*+(.|_)spec.cjs'",
         ] + args,
         data = data + ["//:node_modules/source-map-support"],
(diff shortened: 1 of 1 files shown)


tools/defaults.bzl
```diff
@@ -65,9 +65,10 @@ def jasmine_test(data = [], args = [], **kwargs):
         chdir = native.package_name(),
         args = [
             "--require=%s/node_modules/source-map-support/register.js" % relative_to_root,
-            "**/*spec.js",
-            "**/*spec.mjs",
-            "**/*spec.cjs",
+            # Escape so that the `js_binary` launcher triggers Bash expansion.
```
alan-agius4, line 68: NIT: should the below be `*_spec.js`, `*_spec.cjs` and `*_spec.mjs` to be more future proof and avoid picking up accidentally files that end up with the word `spec`.

devversion: I think, we would prefer being safe, capturing more files vs. skipping some file. e.g. when someone uses `.spec` accidentally.

We could consider just duplicating the patterns in the future, but I think this is okay to leave as is for now (until we see some accidental files being picked up and causing issues)

alan-agius4: I am fine with that, but you can also do something like `*[.-_]spec.js` (If skylark supports this syntax)

devversion: I looked more and it's actually a `node-glob` that's now properly parsed— given that we escape properly now 😄 Updated, thanks!

dgp1130:

PR #30471, opened 2025-06-05
fix(@schematics/angular): correctly detect modules using new file extension format

alan-agius4:
Previously, some internal methods failed to locate modules when using the `-module.ts` filename pattern instead of the legacy `.module.ts`. This update ensures proper detection with the new format.


clydin (approved):

alan-agius4:

PR #30473, opened 2025-06-05
build: simplify unit test target names

alan-agius4:
Renamed unit test targets for easier execution without needing to inspect build files.


clydin (approved):

alan-agius4:

PR #30536, opened 2025-06-13
fix(@angular/build): include custom bundle name scripts with karma

clydin:
When using the `karma` builder with custom "scripts" option entries that contain the `bundleName` suboption, the output files for these scripts will now correctly be injected into the test index HTML content. Previously only the default `scripts.js` output file would be used.


jkrems (approved):

clydin:

PR #30552, opened 2025-06-18
fix(@angular/build): increase worker idle timeout

### LINKED ISSUE #30549: CI crash after upgrading to angular 20 `assert.strictEqual(workerInfo.currentUsage(), 0)`
### Command

build

### Is this a regression?

19.2.2

### Description

After upgrading to Angular 20, our CI starts crashing when building the application for production:

```
❯ Building...
/home/runner/_work/clickup_frontend/clickup_frontend/node_modules/.pnpm/piscina@5.0.0/node_modules/piscina/src/index.ts:447
        assert.strictEqual(workerInfo.currentUsage(), 0);
               ^
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
1 !== 0
    at Timeout._onTimeout (/home/runner/_work/clickup_frontend/clickup_frontend/node_modules/.pnpm/piscina@5.0.0/node_modules/piscina/src/index.ts:447:16)
    at listOnTimeout (node:internal/timers:588:17)
    at processTimers (node:internal/timers:523:7) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: 1,
  expected: 0,
  operator: 'strictEqual'
}
```

The assertion is being thrown from the piscina dependency here:
https://github.com/piscinajs/piscina/blob/fe781d66040eb87446692a3f7056c9c2c7b84a4a/src/index.ts#L447 which was added in 4.9.0 by https://github.com/piscinajs/piscina/pull/590 (angular 19 uses `4.8.0` which was why this is not a problem until now)

We patched our version of the angular CLI to bump the `idleTimeout` here to be `10_000`, and that seems to have resolved it:

https://github.com/angular/angular-cli/blob/2672f6ec17de6e05b19acda0e0b09a6715c9f83f/packages/angular/build/src/utils/worker-pool.ts#L18

Would you accept a PR to apply that patch? (or maybe it could be `Infinity` to disable this check completely? I'm unsure what the consequences of that change would be though)

I assume the issue is because our application source code is so large that it takes longer than 1s for the worker to initialise, and that timeout is hit.

### Minimal Reproduction

A minimal repro is not possible as:
* The problem does not always occur (presumably because the worker sometimes initialises under the current 1s idle timeout)
* The problem only occurs on resource constrained hardware (it cannot be reproduced locally on an M3 macbook, and only within github actions)
* The problem only seems to affect large applications (100mb+ of source code). Our smaller applications do not suffer with this problem

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI `20.0.2` + framework `20.0.3`
```

### Anything else relevant?

_No response_

alan-agius4:
This is a workaround for https://github.com/piscinajs/piscina/issues/816

Closes #30549

### DIFF
--- a/packages/angular/build/src/utils/worker-pool.ts
+++ b/packages/angular/build/src/utils/worker-pool.ts
@@ -15,7 +15,8 @@ export class WorkerPool extends Piscina {
   constructor(options: WorkerPoolOptions) {
     const piscinaOptions: WorkerPoolOptions = {
       minThreads: 1,
-      idleTimeout: 1000,
+      // Workaround for https://github.com/piscinajs/piscina/issues/816
+      idleTimeout: 10_000,
       // Web containers do not support transferable objects with receiveOnMessagePort which
       // is used when the Atomics based wait loop is enable.
       atomics: process.versions.webcontainer ? 'disabled' : 'sync',


alan-agius4:

PR #30571, opened 2025-06-20
test(@angular/build): reset project metadata before each run

alan-agius4:
Prevents test flakiness caused by `useProject` not cleaning up metadata between runs.


alan-agius4:

Issue #30548, opened 2025-06-17
Add Default Prettier Configuration for Angular Templates


Mikastark:
### Description

Hello Angular Team,

Since Angular v20, component files no longer use the `.component` suffix by default. As a result, associated template files (e.g., my-feature.html) are no longer automatically recognized by Prettier as Angular templates. This leads Prettier to fall back to its default HTML parser, which doesn't handle Angular-specific syntax correctly.

### Proposed solution

To improve the developer experience and avoid formatting issues, it would be helpful for new Angular projects to include a default Prettier configuration specifying the Angular parser for .html files.

Suggested addition to package.json:

```json
"prettier": {
  "overrides": [
    {
      "files": "*.html",
      "options": {
        "parser": "angular"
      }
    }
  ]
}
```

This configuration ensures consistent and correct formatting of Angular templates out of the box, without requiring manual setup from the developer. It aligns with Angular’s philosophy of providing strong defaults for common tooling.

Thanks for considering this improvement!

### Alternatives considered

Instead of `package.json`, we could consider having a `.prettierrc` file to set prettier configuration.

vmasek:
Example of the same control flow formatted by prettier (3.6.0)

with `.component` suffix:

![Image](https://github.com/user-attachments/assets/a164d33a-94fc-4f19-9487-cfe0823d46d4)

and without the suffix:

![Image](https://github.com/user-attachments/assets/0ffda4fa-48d4-45ae-b12b-9c2f703b8d25)

Mikastark:
Another simple and effective solution would be to set the ngHtml option to true by default for the @schematics/angular:component schematic in angular.json, like so:

```json
"schematics": {
  "@schematics/angular:component": {
    "ngHtml": true
  }
}
```

This change would suffix all component template files with .ng.html, enabling Prettier to automatically apply its Angular parser.

The main challenge is that Prettier currently cannot distinguish Angular templates from regular HTML files unless they follow a specific naming convention like .ng.html.

hybrist:
> This change would suffix all component template files with `.ng.html`, enabling Prettier to automatically apply its Angular parser.

I only skimmed the prettier code and didn't try anything out but from what I've seen `.ng.html` has no special handling in prettier today. Or is that setup somewhere else?

hybrist:
We're adding the `prettier` section to new `package.json` files starting with the next release of v20. Thanks for bringing this up! This is definitely a sharp edge with the new filenames that we'll take another look at.

[...]

PR #30590, opened 2025-06-24
fix(@schematics/angular): add missing prettier config

hybrist:
The current style guide no longer enforces the use of the template file extension `.component.html`. This means that prettier won't auto-detect the proper parser for these files anymore.

To ensure template formatting works out-of-the-box, we're adding a prettier config to newly created projects.

Fixes https://github.com/angular/angular-cli/issues/30548

### DIFF
--- a/packages/schematics/angular/workspace/files/package.json.template
+++ b/packages/schematics/angular/workspace/files/package.json.template
@@ -8,6 +8,16 @@
     "watch": "ng build --watch --configuration development"<% if (!minimal) { %>,
     "test": "ng test"<% } %>
   },
+  "prettier": {
+    "overrides": [
+      {
+        "files": "*.html",
+        "options": {
+          "parser": "angular"
+        }
+      }
+    ]
+  },
   "private": true,
   "dependencies": {
     "@angular/common": "<%= latestVersions.Angular %>",
--- a/packages/schematics/angular/workspace/index_spec.ts
+++ b/packages/schematics/angular/workspace/index_spec.ts
@@ -133,4 +133,10 @@ describe('Workspace Schematic', () => {
     const { tasks } = parseJson(tree.readContent('.vscode/tasks.json').toString());
     expect(tasks).not.toContain(jasmine.objectContaining({ type: 'npm', script: 'test' }));
   });
+
+  it('should include prettier config overrides for Angular templates', async () => {
+    const tree = await schematicRunner.runSchematic('workspace', defaultOptions);
+    const pkg = JSON.parse(tree.readContent('/package.json'));
+    expect(pkg.prettier).withContext('package.json#prettier is present').toBeTruthy();
+  });
 });


packages/schematics/angular/workspace/files/package.json.template
```diff
@@ -8,6 +8,16 @@
     "watch": "ng build --watch --configuration development"<% if (!minimal) { %>,
     "test": "ng test"<% } %>
   },
+  "prettier": {
```
alan-agius4, line 11: In general, I am not a fan of having tools configured in the package manifest. I’d rather have the additional config file which is clearer.

Thoughts @clydin & @dgp1130 ?

packages/schematics/angular/workspace/files/package.json.template
```diff
@@ -8,6 +8,16 @@
...
+    "overrides": [
+      {
+        "files": "*.html",
```
alan-agius4, line 14: This will pick up the index.html file, I am assuming the angular parser can format a standard full HTML 5 document right?

packages/schematics/angular/workspace/files/package.json.template
hybrist: My thinking here was that creating a dedicated file may be more confusing for people who don't actively use `prettier`. It feels a bit less boilerplate-y. But I don't have super strong opinions either way.

packages/schematics/angular/workspace/files/package.json.template
hybrist: Yes, the assumption is that anything not actively using angular syntax would just be formatted like normal HTML.

packages/schematics/angular/workspace/files/package.json.template
alan-agius4: That’s a valid point.

alan-agius4 (approved):

hybrist:

packages/schematics/angular/workspace/files/package.json.template
robertIsaac: actually this will be super confusing for anyone using prettier since creating a prettierrc file will have zero affect (prettier doesn't merge configurations but only take one; and the one inside the package.json has the highest order)
and I'm talking from experience because I faced it before myself as in one repo someone configured prettier in the package.json and I was trying to add one plugin and it wasn't working and it took me forever to figure out why

hybrist: Is there an issue in prettier about that behavior? That seems worth fixing independent of this change.

robertIsaac: it's how it works
I can open an issue there and see where it goes, but I don't think they will change this behavior since it can lead to many repos being broken

hybrist: Sorry, I wasn't talking about changing the behavior. I was talking about silently ignoring a config file that's clearly "meant" to be used (it's called prettierrc after all) and not warning the user about it. Printing a warning should hopefully not break anything and would remove this confusion?

robertIsaac: okay I will create an issue now

robertIsaac: https://github.com/prettier/prettier/issues/17683

PR #30591, opened 2025-06-24
feat(@angular-devkit/build-angular): provide partial custom postcss configuration support

clydin:
The `browser` builder now supports providing a custom postcss configuration similar to the functionality available in the `@angular/build:application` builder. A postcss configuration file named either `postcss.config.json` or `.postcssrc.json` will be discovered in order from the project root or workspace root. Only plugins can be configured via the file. All other fields will be ignored. Additionally the plugins will be placed after any internal build system plugins but before `autoprefixer`. When using a custom postcss configuration, the automatic tailwind integration will be disabled. The Tailwind postcss plugin should be added to the configuration file in this scenario.


clydin:
Superseded by #30592

PR #30592, opened 2025-06-24
feat(@angular-devkit/build-angular): partial custom postcss config support

jkrems:
The `browser` builder now supports providing a custom postcss configuration similar to the functionality available in the `@angular/build:application` builder. A postcss configuration file named either `postcss.config.json` or `.postcssrc.json` will be discovered in order from the project root or workspace root. Only plugins can be configured via the file. All other fields will be ignored. Additionally the plugins will be placed after any internal build system plugins but before `autoprefixer`. When using a custom postcss configuration, the automatic tailwind integration will be disabled. The Tailwind postcss plugin should be added to the configuration file in this scenario.


packages/angular_devkit/build_angular/src/tools/webpack/configs/styles.ts
```diff
@@ -75,25 +79,40 @@ export async function getStylesConfig(wco: WebpackConfigOptions): Promise<Config
...
-        `Tailwind CSS configuration file found (${relativeTailwindConfigPath})` +
-          ` but the 'tailwindcss' package is not installed.` +
-          ` To enable Tailwind CSS, please install the 'tailwindcss' package.`,
-      );
+  const searchDirectories = await generateSearchDirectories([projectRoot, root]);
+  const postcssConfig = await loadPostcssConfiguration(searchDirectories);
+
+  if (postcssConfig) {
+    for (const [pluginName, pluginOptions] of postcssConfig.plugins) {
+      const resolvedPlugin = require.resolve(pluginName, { paths: [root] });
```
hybrist, line 87: This `require.resolve` was added to the original code to ensure the plugins are resolved relative to the project.

PR #30601, opened 2025-06-25
feat(@angular/cli): add initial MCP server implementation

clydin:
An initial experimental implementation of a Model Context Protocol (MCP) server is now available as a command within the Angular CLI. The server is a `stdio` based server intended to run locally by a host via the `ng mcp` command. Currently, the server provides one resource and one tool. The resource is for the best practices system instructions found https://angular.dev/ai/develop-with-ai. The tool allows for listing the Angular projects within the Angular workspace. These are preliminary and subject to change. Additional functionality may be added in the future to provide additional capabilities for IDE host integrations.


packages/angular/cli/package.json
```diff
@@ -37,7 +38,8 @@
     "pacote": "21.0.0",
     "resolve": "1.22.10",
     "semver": "7.7.2",
-    "yargs": "18.0.0"
+    "yargs": "18.0.0",
+    "zod": "3.25.67"
```
alan-agius4: Seems to be unused.

packages/angular/cli/src/commands/mcp/cli.ts
```diff
@@ -0,0 +1,50 @@
...
+import { CommandModule, CommandModuleImplementation } from '../../command-builder/command-module';
+import { isTTY } from '../../utilities/tty';
+import { createMcpServer } from './mcp-server';
+
+const INTERACTIVE_MESSAGE = `
+To start using the Angular CLI MCP Server, add this configuration to your host:
+
+{
+  "mcpServers": {
+    "angular-cli": {
```
alan-agius4, line 20: Why not `ng mcp`?

clydin: it's a host specific identifier that in some cases can be used in dot-notation.

alan-agius4: I meant the command  instead of `npx @angular/cli mcp`

clydin: It seems to be a common pattern in the configuration.  Plus it bypasses global installation issues.  The command needs to be available globally.  We can adjust and iterate on the best approach based on feedback though.

PR #30611, opened 2025-06-26
WHAT IFFFFFFF THOUGH

jkrems:
Capturing some quick numbers. Each one is an incremental test run after modifying `packages/ngtools/webpack/src/transformers/elide_imports_spec.ts` (modify describe string literal).

* Current test runner: ~8.5s (`time pnpm bazel test //packages/ngtools/webpack:test`)
* Non-interactive vitest: ~4s (`time CI=1 pnpm --filter=@ngtools/webpack exec vitest`)
* Interactive vitest: ~2s (`pnpm --filter=@ngtools/webpack exec vitest`, self-reported time in the UI)


alan-agius4:
Since this was a POC, let's close this for now.

PR #30619, opened 2025-06-27
build: use more shards for application builder tests

jkrems:
We've added more tests (yay!) but at least when running on RBE, this seems to have made things less stable.

I've tried to run the tests a few times with `--config=remote --nocache_test_results` and this appears to make them pass consistently. There may be a more perfect integer but x2 should give us stable CI again (at least for this issue).


clydin (approved):

hybrist:

PR #30632, opened 2025-07-01
fix(@angular/build): proxy karma request from `/` to `/base`

### LINKED ISSUE #30627: Scripts are not loaded properly in v20
### Command

test

### Is this a regression?

19

### Description

I was in the middle of migration from v19 to v20 for our ui-kit library. It went OK but then I tried to run unit tests and some of them failed because they're trying to load a 3rd party script. So basically this simplified piece of code was throwing an error

```ts
loadScript() {
    return new Observable<void>(obs => {
      const script = document.createElement("script");
      script.src = "test.js";

      script.onload = () => {
        obs.next();
        obs.complete();
      }

      script.onerror = () => {
        obs.error();
        obs.complete();
      }
      document.body.appendChild(script);
    })
}
```

We have a config for the script file in `angular.json`. Here it is (I just renamed it for the sake of simplicity)

```json
"scripts": [
              {
                "input": "src/test.js",
                "bundleName": "test",
                "inject": false
              }
  ]
```

This works fine in v19 but doesn't in v20.

### Minimal Reproduction

I created two repos with a unit test which tries to load a script
- v19 https://github.com/ms-dosx86/test-v19 (passes)
- v20 https://github.com/ms-dosx86/test-v20 (doesn't pass)

Just run `ng test`.

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 20.0.4
Node: 22.12.0
Package Manager: npm 10.9.0
OS: darwin arm64

Angular: 20.0.5
... platform-browser, router

@angular-devkit/architect    0.2000.4
@angular-devkit/core         20.0.4
@angular-devkit/schematics   20.0.4
@angular/build               20.0.4
@angular/cli                 20.0.4
@schematics/angular          20.0.4
```

### Anything else relevant?

_No response_

alan-agius4:
This commit fixes an issue where Karma requests were not being correctly proxied, leading to 404 errors for assets and other files during test execution in Angular projects. The fix ensures that all Karma requests from the root (`/`) are correctly forwarded to the `/base` directory, where Karma serves the files.

Karma will always serve files from `/base`. This change aligns the testing environment with this fundamental behavior, resolving pathing issues and ensuring that tests can locate their dependencies reliably.

Closes #30627

### DIFF
--- a/packages/angular/build/src/builders/karma/index.ts
+++ b/packages/angular/build/src/builders/karma/index.ts
@@ -99,6 +99,7 @@ function getBuiltInKarmaConfig(
   // Any changes to the config here need to be synced to: packages/schematics/angular/config/files/karma.conf.js.template
   return {
     basePath: '',
+    rootUrl: '/',
     frameworks: ['jasmine'],
     plugins: [
       'karma-jasmine',
@@ -106,6 +107,9 @@ function getBuiltInKarmaConfig(
       'karma-jasmine-html-reporter',
       'karma-coverage',
     ].map((p) => workspaceRootRequire(p)),
+    proxies: {
+      '/': '/base/',
+    },
     jasmineHtmlReporter: {
       suppressAll: true, // removes the duplicated traces
     },


packages/angular/build/src/builders/karma/index.ts
```diff
@@ -99,13 +99,17 @@ function getBuiltInKarmaConfig(
   // Any changes to the config here need to be synced to: packages/schematics/angular/config/files/karma.conf.js.template
```
hybrist, line 99: Does this apply? Do we need to sync this?

alan-agius4: No, as the build-angular package uses it’s how karma .html files, where the base href is hardcoded

alan-agius4:

PR #30645, opened 2025-07-03
fix(@angular/build): correctly remap Angular diagnostics

crisbeto:
The esbuild builder has some logic that re-map diagnostics coming from the compiler, depending on their `source` and `code`. Currently this is incorrect, because it assumed that a value of `ngtsc` for `source` always means that the error is from the Angular compiler, but what it actually means is that it comes from an Angular template diagnostics. Furthermore, we can't rely on a `source` always being defined.

These changes align the logic to a similar one we already have in the compiler where we assume the diagnostic comes from Angular if it starts with `-99`.


alan-agius4 (approved):

alan-agius4:

PR #30649, opened 2025-07-03
build: update bazel commands to use pnpm

alan-agius4:
`yarn` is no longer used.


alan-agius4:

PR #30654, opened 2025-07-04
feat(@angular/build): allow enabling Bazel sandbox plugin with esbuild

sbarfurth:
## PR Checklist

- [ ] ~Docs have been added / updated (for bug fixes / features)~

*User facing documentation is intentionally omitted as this will be considered internal for now.*

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

When trying to integrate the `esbuild` based builder into Bazel with `rules_js` we found that it will incorrectly follow symlinks out of the sandbox. This is because Node.js based tooling runs in the output tree to support canonical JS project directory structures. The output tree will contain symlinks outside of the sandbox. Node tooling will generally follow these symlinks, which violates the rules of Bazel sandboxing. This can manifest in a wide variety of errors. One example we encountered with Angular compilation is that the symlinked browser entry point (e.g. `main.ts`) is outside of the range of `tsconfig.json` when the compiler follows the symlink.

Issue Number: N/A

## What is the new behavior?

Setting the new `ENABLE_BAZEL_SANDBOX_PLUGIN` environment variable to `true` or `1` during a build with the `esbuild` based builder will now inject a special plugin to make `esbuild` compatible with Bazel builds in the output tree.

- [ ] Yes
- [x] No

## Other information

The plugin itself was originally written in https://github.com/aspect-build/rules_esbuild. The version container in this commit is a fork of https://github.com/aspect-build/rules_esbuild/blob/e4e49d3354cbf7087c47ac9c5f2e6fe7f5e398d3/esbuild/private/plugins/bazel-sandbox.js. I've adapted the JS file to TypeScript and made no further changes.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
+++ b/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
@@ -30,6 +30,7 @@ import { createAngularLocaleDataPlugin } from './i18n-locale-plugin';
 import type { LoadResultCache } from './load-result-cache';
 import { createLoaderImportAttributePlugin } from './loader-import-attribute-plugin';
 import { createRxjsEsmResolutionPlugin } from './rxjs-esm-resolution-plugin';
+import { createBazelSandboxPlugin } from './sandbox-plugin-bazel';
 import { createServerBundleMetadata } from './server-bundle-metadata-plugin';
 import { createSourcemapIgnorelistPlugin } from './sourcemap-ignorelist-plugin';
 import { SERVER_GENERATED_EXTERNALS, getFeatureSupport, isZonelessApp } from './utils';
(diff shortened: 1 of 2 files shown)


devversion:
For additional context: This will also be useful / needed for our usages of CLI inside Bazel for e.g. angular.dev

sbarfurth:
I've had to change the original plugin to try for a `.js` file is a `.ts` file was resolved but doesn't exist in the sandbox. One example where this happens is if a `.ts` file is pre-compiled with something like `ts_project`. The sandbox then only gets the output of `ts_project`, but the `ts_project` rule itself depended on the `.ts` file. This means the `.ts` file will be present outside the sandbox, but it won't be linked into it. The resolution will resolve the `.ts` file outside the sandbox, but the path mapped back into the sandbox won't exist. Simply rewriting to `.js` is sufficient for this case and seems sensible enough.

sbarfurth:
Should I make any further changes here? Happy to help get this merged if it's something I can support.

hybrist:
@clydin I assume this is something you're still looking at / want to look at?

sbarfurth:
I've updated the logic once more to allow remapping into the runfiles tree when requested. I've found this to be needed when executing the Architect from a `js_test`/`js_binary` where the sources are in the runfiles tree.

alan-agius4:
Is this still need in version 22.2.1 following: https://github.com/angular/angular-cli/commit/261dbb37cbe01492240c4cedc644663b15a4296a?

sbarfurth:
@alan-agius4 I've found that not to work for me. I tried it instead of my patch and my build went back to failing with path errors.

alan-agius4:
@clydin can you take a look at the above.

TurboTea3:
I can confirm, tested post merge [20.2.1](https://github.com/angular/angular-cli/releases/tag/20.2.1). The bug still persists. This is becoming quite problematic. Currently using a scrappy [esbuild overwrite package](https://www.npmjs.com/package/@angular-builders/custom-esbuild) to patch.

Would really appreciate a fix for this!

sbarfurth:
We are using this patch on a medium-large Angular app with `aspect_rules_js`, RBE builders and local sandboxing. It definitely works, it's just annoying to keep the patch updated at the moment.

TurboTea3:
Same for us! Having problems with testing, because we cannot patch the underlying bazel sandbox escape. Would be really nice to have an official fix and switch back to the official build tools.

PR #30664, opened 2025-07-07
fix(@angular/build): support extra test setup files with unit-test vitest runner

clydin:
When using the experimental unit-test builder with Vitest, a new `setupFiles` option is now available. This option is similar to the Vitest option in that it allows for setup and configuration prior to each test. The `setupFiles` are executed after any application polyfills as well as after the TestBed initialization. If custom TestBed initialization is needed (this is not typical), The TestBed environment can first be reset in a setupFile and then initialized as needed. Note that resetting the TestBed environment in this way will cause the `providersFile` to no longer add any providers to the tests and any custom providers would need to be manually added during initialization.


clydin:

PR #30669, opened 2025-07-08
Stablize CI tests

alan-agius4:
See each commit

### DIFF
--- a/modules/testing/builder/src/test-utils.ts
+++ b/modules/testing/builder/src/test-utils.ts
@@ -24,8 +24,8 @@ import {
 import path from 'node:path';
 import { firstValueFrom } from 'rxjs';
 
-// Default timeout for large specs is 2.5 minutes.
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;
+// Default timeout for large specs is 60s.
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 60_000;
 
 export const workspaceRoot = join(normalize(__dirname), `../projects/hello-world-app/`);
 export const host = new TestProjectHost(workspaceRoot);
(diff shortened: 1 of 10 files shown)


modules/testing/builder/src/test-utils.ts
```diff
@@ -24,8 +24,8 @@ import {
 import path from 'node:path';
 import { firstValueFrom } from 'rxjs';
 
-// Default timeout for large specs is 2.5 minutes.
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;
+// Default timeout for large specs is 60s.
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 60_000;
```
hybrist, line 28: ❤️  Always great to see these _decrease_!

packages/angular/build/src/builders/application/tests/options/output-path_spec.ts
```diff
@@ -168,33 +124,19 @@ describeBuilder(buildApplication, APPLICATION_BUILDER_INFO, (harness) => {
...
-        it(`should emit server bundles in 'node-server' directory`, async () => {
-          const { result } = await harness.executeOnce();
-          expect(result?.success).toBeTrue();
-
           harness.expectFile('dist/node-server/server.mjs').toExist();
         });
       });
 
       describe(`'browser' is set to 'public'`, () => {
-        beforeEach(() => {
```
hybrist, line 197: I assume this effectively the same as `s/beforeEach/beforeAll/`? Should we preserve some of the `it` descriptions using something like `expect(...).withContext()`? Or are the different tested aspects mostly self-evident?

alan-agius4: I think they are self evident.

alan-agius4:

PR #30673, opened 2025-07-08
test: increase timeout for ssr dev server tests

alan-agius4:
These old tests take longer to run


alan-agius4:

PR #30705, opened 2025-07-11
refactor(@angular/cli): add a documentation search tool to MCP server

clydin:
An additional MCP tool is now available with the `ng mcp` stdio MCP server that supports querying the `angular.dev` documentation. This uses the same algolia based search indexing that the documentation website uses. Rate limiting has been implemented with the MCP tool that may be adjusted based on feedback. The tool returns one or more URLs and titles for relevant documentation for a given query. Content of these search results are currently not fetched but rather this action is deferred to the host to determine which items are most relevant and should be retrieved from the documentation website.


clydin:

PR #30717, opened 2025-07-15
refactor(@angular/cli): provide a find examples MCP server tool

clydin:
The built-in stdio MCP server (`ng mcp`) for the Angular CLI now includes a tool that can find Angular code usage examples. The code examples are indexed and stored locally within the Angular CLI install. This removes the need for network requests to find the examples. It also ensures that the examples are relevant for the version of Angular currently being used. This tool requires Node.js 22.16 or higher. Lower versions will not have this specific tool available for use.


packages/angular/cli/src/commands/mcp/tools/examples.ts
```diff
@@ -0,0 +1,173 @@
...
+    async ({ query }) => {
+      if (!db || !queryStatement) {
+        suppressSqliteWarning();
+
+        const { DatabaseSync } = await import('node:sqlite');
+        db = new DatabaseSync(exampleDatabasePath, { readOnly: true });
+        queryStatement = db.prepare('SELECT * from examples WHERE examples MATCH ? ORDER BY rank;');
+      }
+
+      const sanitizedQuery = sanitizeSearchQuery(query);
```
devversion: Is there no library we can use for sanitization, or do we just avoid another dependency? I'd prefer us not having to maintain this.

clydin: It's particular to the Sqlite FTS5 query microsyntax.  It's mainly about wrapping query substrings to avoid unintentionally using some of the more esoteric query features like column filters via `-`.  The entire query string is properly escaped due to the prepared statement usage.

devversion: Is there no library for this? It feels like logic that can be hard to revisit in the future

clydin: no. It's not SQL. it's a custom microsyntax for the FTS5 extension that's being used as a string literal in the parent SQL query.

PR #30718, opened 2025-07-15
feat(@schematics/angular): Applications are zoneless by default

atscott:
This change updates applications to omit the ZoneJS dependency by default. It depends on https://github.com/angular/angular/pull/62655, which allows us to also exclude the `provideZonelessChangeDetection` provider.


packages/angular_devkit/build_angular/src/builders/jest/init-test-bed.mjs
```diff
@@ -9,10 +9,16 @@
...
 
+import { provideZoneChangeDetection, NgModule } from '@angular/core';
 import { getTestBed } from '@angular/core/testing';
 import { BrowserTestingModule, platformBrowserTesting } from '@angular/platform-browser/testing';
 
-getTestBed().initTestEnvironment(BrowserTestingModule, platformBrowserTesting(), {
+@NgModule({
+  providers: [typeof window.Zone !== 'undefined' ? provideZoneChangeDetection() : []],
+})
+class TestModule {}
```
atscott, line 19: Do I need to avoid the `@NgModule` here and use `compileNgModuleDefs` instead like in the karma/browser_builder? 

Same question for `karma/application_builder.ts` `web-test-runner/jasmine_runner.js`

alan-agius4: I think these are fine when using esbuild.

packages/schematics/angular/ng-new/schema.json
```diff
@@ -141,7 +141,8 @@
     },
     "zoneless": {
       "description": "Create an initial application that does not utilize `zone.js`.",
-      "type": "boolean"
+      "type": "boolean",
+      "default": true
```
alan-agius4: The default should not be set here and instead, it should only be set in the application schematic.

packages/schematics/angular/application/schema.json
```diff
@@ -126,7 +126,7 @@
       "description": "Generate an application that does not use `zone.js`.",
       "x-prompt": "Do you want to create a 'zoneless' application without zone.js?",
```
alan-agius4, line 127: Are we keeping the prompt?

tests/legacy-cli/e2e/initialize/500-create-project.ts
```diff
@@ -20,7 +20,7 @@ export default async function () {
     // Ensure local test registry is used when outside a project
     await setNPMConfigRegistry(true);
 
-    await ng('new', 'test-project', '--skip-install');
+    await ng('new', 'test-project', '--skip-install', '--no-zoneless');
```
alan-agius4, line 23: In a followup we should likely try to have most of the our tests test the default behavior (IE: zoneless).

alan-agius4 (approved):
LGTM

PR #30719, opened 2025-07-16
refactor(@schematics/angular): fix layout of pill button

JeanMeche:
In the case of a long project name, the button would wrap. This commit prevents that.


hybrist:

PR #30723, opened 2025-07-16
refactor(@angular/build): provide private factory API for internal compilation infrastructure

clydin:
The `createAngularCompilation` factory function is now available from the `private` entry point for the `@angular/build` package. All items exposed from this entry point are not subject to SemVer guarantees and may change between releases.


dgp1130 (approved):

clydin:

PR #30725, opened 2025-07-16
refactor(@angular/cli): add a get best practices guide MCP tool

clydin:
The Angular CLI's stdio-based MCP server now contains a tool to get an Angular best practices guide. This is in addition to a resource with the same content. The tool provides a description that strongly encourages the use of this tool and its content when performing Angular related tasks. This is useful in cases where MCP resource usage is not available or the resource would need to manually be added as context for specific use cases.

### DIFF
--- a/packages/angular/cli/src/commands/mcp/tools/best-practices.ts
+++ b/packages/angular/cli/src/commands/mcp/tools/best-practices.ts
@@ -1,0 +1,45 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { readFile } from 'node:fs/promises';
+import path from 'node:path';
+
+export function registerBestPracticesTool(server: McpServer): void {
+  server.registerTool(
+    'get_best_practices',
+    {
+      title: 'Get Angular Coding Best Practices Guide',
+      description:
+        'You **MUST** use this tool to retrieve the Angular Best Practices Guide ' +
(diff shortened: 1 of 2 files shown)


packages/angular/cli/src/commands/mcp/tools/best-practices.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { readFile } from 'node:fs/promises';
+import path from 'node:path';
+
+export async function registerBestPracticesTool(server: McpServer): Promise<void> {
```
hybrist: Does this function need to be async? It looks like it's just passing an async function down in its current form.

geromegrignon:
> The Angular CLI's stdio-based MCP server now contains a tool to get an Angular best practices guide. This is in addition to a resource with the same content. The tool provides a description that strongly encourages the use of this tool and its content when performing Angular related tasks. This is useful in cases where MCP resource usage is not available or the resource would need to manually be added as context for specific use cases.

If the `MCP resource usage` refers to the compatibility matrix(https://modelcontextprotocol.io/clients#feature-support-matrix), from my attemps at doing about the same with Cursor, a content with `type: "resource"` exposed by a tool is still considered as a resource, resulting in a `Content type "resource" not supported` message

clydin:
> If the MCP resource usage refers to the compatibility matrix(https://modelcontextprotocol.io/clients#feature-support-matrix), from my attemps at doing about the same with Cursor, a content with type: "resource" exposed by a tool is still considered as a resource, resulting in a Content type "resource" not supported message

It was referring to UI mechanisms to add resources as context.  However, to be safe the content is now text.

clydin:

PR #30728, opened 2025-07-16
test: use valid theme for material schematic

jkrems:
The option "custom" isn't really meant to be a value pass via the `--theme` option. This broke our tests when it stopped being accepted properly by the schematic.


hybrist:

PR #30740, opened 2025-07-18
build: update to TypeScript 5.9 beta

crisbeto:
Updates the repo to the TypeScript 5.9 beta and widens the peer dependency ranges.

### DIFF
--- a/packages/angular/build/src/builders/application/tests/behavior/rebuild-errors_spec.ts
+++ b/packages/angular/build/src/builders/application/tests/behavior/rebuild-errors_spec.ts
@@ -149,12 +149,12 @@ describeBuilder(buildApplication, APPLICATION_BUILDER_INFO, (harness) => {
         [
           async () => {
             // Add invalid block syntax
-            await harness.appendToFile('src/app/app.component.html', '@one');
+            await harness.appendToFile('src/app/app.component.html', '@if-one');
           },
           async ({ logs }) => {
             expect(logs).toContain(
               jasmine.objectContaining<logging.LogEntry>({
-                message: jasmine.stringContaining('@one'),
+                message: jasmine.stringContaining('@if-one'),
               }),
             );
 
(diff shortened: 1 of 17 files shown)


packages/angular/build/src/builders/application/tests/behavior/rebuild-errors_spec.ts
```diff
@@ -149,12 +149,12 @@ describeBuilder(buildApplication, APPLICATION_BUILDER_INFO, (harness) => {
         [
           async () => {
             // Add invalid block syntax
-            await harness.appendToFile('src/app/app.component.html', '@one');
+            await harness.appendToFile('src/app/app.component.html', '@if-one');
```
crisbeto, line 152: I had to change this test, because in the latest version Angular is a bit more permissive about parsing unsupported blocks.

packages/angular/cli/src/command-builder/utilities/schematic-workflow.ts
```diff
@@ -38,11 +38,15 @@ export function subscribeToWorkflow(
         );
         break;
       case 'update':
-        logs.push(`${colors.cyan('UPDATE')} ${eventPath} (${event.content.length} bytes)`);
+        logs.push(
+          `${colors.cyan('UPDATE')} ${eventPath} (${(event.content as unknown as Buffer).length} bytes)`,
```
clydin, line 43: This does not seem ideal from a type safety perspective.
Is there some information/context on the changes needed here?

crisbeto: So as far as I could tell, the types didn't really change but TS is being better at enforcing them. I did start to update the types, but it ended up spanning a ton of files so decided to keep the change simple for now.

clydin: Ok. makes sense.
Can you add some TODOs in there so we can clean this up later on?

clydin (approved):

packages/angular/cli/src/command-builder/utilities/schematic-workflow.ts
crisbeto: Done, I've added TODOs on all those `as unknown as` cases.

PR #30746, opened 2025-07-21
fix(@angular/cli): `define` option is being included multiple times in the JSON help

### LINKED ISSUE #30710: `define` option is being included multiple times in the help section
### Command

test

### Is this a regression?

_No response_

### Description

The define option is being included multiple times in the JSON help output example: https://github.com/angular/cli-builds/blob/4e015ce6eef338f64ce413547ce046a35ab50a01/help/test.json#L42-L51 which causes the option to be displayed multiple times in angular.dev, see: https://angular.dev/cli/test

### Minimal Reproduction

n/a

### Exception or Error

```text

```

### Your Environment

```text
n/a
```

### Anything else relevant?

Possibly this could be caused by the option being registered multiple times in yargs.

alan-agius4:
This commit addresses an issue where the `define` option was being included multiple times in the JSON help.

Closes #30710


alan-agius4:

PR #30753, opened 2025-07-21
refactor(@angular/cli): improve description of MCP documentation search tool

clydin:
The MCP tool description for the Angular CLI's documentation search tool has been expanded to include more context regarding its use and result format. When no results are found, an explicit text response is also now generated to indicate this case.


clydin:

PR #30763, opened 2025-07-22
feat(@angular-devkit/schematics): add schematics to generate ai context files.

JeanMeche:
* `ng generate config ai` to prompt support tools.
* `ng generate config ai --tool=gemini` to specify the tool.

Supported ai tools: gemini, claude, copilot, windsurf, cursor.


packages/schematics/angular/config/schema.json
```diff
@@ -16,12 +16,21 @@
...
         "index": 0
       }
+    },
+    "tool": {
+      "type": "string",
+      "description": "Specifies the AI tool to configure when type is 'ai'.",
+      "enum": ["gemini", "copilot", "claude", "cursor", "windsurf"],
+      "x-prompt": {
+        "message": "Which AI tool would you like to configure?",
+        "when": "type === 'ai'"
```
alan-agius4: Conditional prompts are not supported.

packages/schematics/angular/config/index.ts
```diff
@@ -103,3 +106,53 @@ function addKarmaConfig(options: ConfigOptions): Rule {
...
+  }
+
+  return async (host) => {
+    const workspace = await readWorkspace(host);
+    const project = workspace.projects.get(options.project);
+    if (!project) {
+      throw new SchematicsException(`Project name "${options.project}" doesn't not exist.`);
+    }
+
+    // Keep this file in sync with the one presented here https://angular.dev/ai/develop-with-ai
```
alan-agius4: We should do something automatted or document it on the FW side, since when there is a change there, a PR needs to be opened here.

packages/angular_devkit/schematics/src/rules/rename.ts
```diff
@@ -0,0 +1,18 @@
+/**
```
alan-agius4: Lets not expose new public APIs. (See below comments)

packages/angular_devkit/schematics/src/rules/move.ts
```diff
@@ -10,6 +10,8 @@ import { join, normalize } from '@angular-devkit/core';
 import { Rule } from '../engine/interface';
 import { noop } from './base';
 
+export function move(from: string, to: string): Rule;
```
alan-agius4, line 13: Can be reverted. (See previous comments)

packages/schematics/angular/config/index.ts
```diff
@@ -103,3 +106,53 @@ function addKarmaConfig(options: ConfigOptions): Rule {
...
+    // The file is located in the framework repo at adev/src/context/best-practices.md
+    const templateName = 'best-practices.md';
+
+    return mergeWith(
+      apply(url('./files'), [
+        filter((p) => p.endsWith(`${templateName}.template`)),
+        applyTemplates({}),
+        move(filePath),
+        rename(filePath, templateName, fileName),
```
alan-agius4: The rename and filtering can be removed, instead using applyTemplates to change the name.

```suggestion
        applyTemplates({ rulesName, directoryName }),
        move(filePath),
```

```
---files 
    -- __directoryName__
           --- __rulesName__.md.template
````

File: packages/schematics/angular/config/index.ts [hunk 6]
```diff
@@ -103,3 +106,53 @@ function addKarmaConfig(options: ConfigOptions): Rule {
...
+    case 'cursor':
+      fileName = 'cursor.md';
+      filePath = path.join('.cursor', 'rules');
+      break;
+    case 'windsurf':
+      fileName = 'guidelines.md';
+      filePath = path.join('.windsurf', 'rules');
+      break;
+    default:
+      throw new SchematicsException(`Unsupported AI tool: ${options.tool}`);
```
alan-agius4: This is redundant as the validation is handed already during the schema validation.

packages/schematics/angular/config/schema.json
```diff
@@ -16,12 +16,20 @@
...
+    "tool": {
+      "type": "string",
+      "description": "Specifies the AI tool to configure when type is 'ai'.",
+      "enum": ["gemini", "copilot", "claude", "cursor", "windsurf"],
+      "x-prompt": {
+        "message": "Which AI tool would you like to configure?"
+      }
     }
   },
   "required": ["project", "type"]
```
alan-agius4: Add a validation that tool is required when config is AI

```
  "allOf": [
    {
      "if": {
        "properties": {
          "type": {
            "const": "ai"
          }
        }
      },
      "then": {
        "required": [
          "tool"
        ]
      }
    }
  ]
```

packages/schematics/angular/config/schema.json
```diff
@@ -16,12 +16,20 @@
...
       "$default": {
         "$source": "argv",
         "index": 0
       }
+    },
+    "tool": {
+      "type": "string",
+      "description": "Specifies the AI tool to configure when type is 'ai'.",
+      "enum": ["gemini", "copilot", "claude", "cursor", "windsurf"],
+      "x-prompt": {
```
alan-agius4: This prompt will be shown for all other schematics which is not desired. I suggest we delete this.

packages/schematics/angular/config/index.ts
```diff
@@ -103,3 +105,49 @@ function addKarmaConfig(options: ConfigOptions): Rule {
...
+    if (!project) {
+      throw new SchematicsException(`Project name "${options.project}" doesn't not exist.`);
+    }
+
+    return mergeWith(
+      apply(url('./files'), [
+        // Keep this file in sync with the one presented here https://angular.dev/ai/develop-with-ai
+        // The file is located in the framework repo at adev/src/context/best-practices.md
+        filter((p) => p.endsWith(`__rulesName__.template`)),
+        applyTemplates({ rulesName: fileName, directoryName: filePath }),
```
dgp1130: Question: Do we need different frontmatter for each of these files, just to apply by default or other configuration to set?

I was expecting these files to be slightly different while containing the same core content.

JeanMeche: You're right, afaik only the cursor file stands out with an additional frontmatter.

packages/schematics/angular/config/index.ts
JeanMeche: Let's handle that as a follow-up when we have a better picture what we need to sync (we might have multiple schematics).

packages/schematics/angular/config/schema.json
```diff
@@ -16,12 +16,17 @@
...
       "x-prompt": "Which type of configuration file would you like to create?",
       "$default": {
         "$source": "argv",
         "index": 0
       }
+    },
+    "tool": {
+      "type": "string",
+      "description": "Specifies the AI tool to configure when type is 'ai'.",
+      "enum": ["gemini", "copilot", "claude", "cursor", "windsurf", "all"]
```
dgp1130: Consider: Should we throw if `--tool` is provided for `karma` or `browserslist` cases? Any way to express that in the schema directly?

packages/schematics/angular/config/index.ts
```diff
@@ -103,3 +106,77 @@ function addKarmaConfig(options: ConfigOptions): Rule {
...
+interface ContextFileInfo {
+  rulesName: string;
+  directory: string;
+  frontmatter?: string;
+}
+
+function addAiContextFile(options: ConfigOptions): Rule {
+  const files: ContextFileInfo[] = [];
+
+  const geminiFile: ContextFileInfo = { rulesName: 'GEMINI.md', directory: '.gemini' };
```
dgp1130: Consider: Would it make more sense to invert this relationship and make templates for `GEMINI.md`, `CLAUDE.md`, etc. which just compose a `<%= bestPractices %>` option? Would that make it easier to configure future differences in the files?

dgp1130 (approved):

packages/schematics/angular/config/index.ts
JeanMeche: I've been thinking about this too. We also need to accont the sync strategy between fw/adev and the cli repo. 
I'll keep this in mind in case this is a favorable refactoring.

packages/schematics/angular/config/schema.json
JeanMeche: addressed

packages/schematics/angular/config/files/__rulesName__.template
clydin, line 1: This appears out of date.  Please update with the content from https://github.com/angular/angular-cli/blob/main/packages/angular/cli/src/commands/mcp/instructions/best-practices.md

packages/schematics/angular/config/index.ts
clydin: Can these constants be moved to a module level constant object literal?
That would also remove the need for the switch and also allow for easier expansion/adjustment of tools.

packages/schematics/angular/config/schema.json
```diff
@@ -16,13 +16,36 @@
...
       "description": "Specifies the type of configuration file to generate.",
-      "enum": ["karma", "browserslist"],
+      "enum": ["karma", "browserslist", "ai"],
       "x-prompt": "Which type of configuration file would you like to create?",
       "$default": {
         "$source": "argv",
         "index": 0
       }
+    },
+    "tool": {
```
clydin: Should this be a multi-select prompt?
We could still default to having them all selected. Would need to change the type to a string array though.  But it's probably unlikely that everyone would want all available tools when not specifying the option.  Especially since the option may not be known to even exist.

JeanMeche: As mentionned by @alan-agius4 [here](https://github.com/angular/angular-cli/pull/30763#pullrequestreview-3046403497), the prompt would show for the others options. Or do you have an alternative in mind ?

packages/schematics/angular/ng-new/index_spec.ts
```diff
@@ -103,4 +103,13 @@ describe('Ng New Schematic', () => {
...
     expect(cli.packageManager).toBe('npm');
   });
+
+  it('should add ai config file when aiConfig is set', async () => {
+    const options = { ...defaultOptions, aiConfig: 'gemini' };
+
+    const tree = await schematicRunner.runSchematic('ng-new', options);
+    const files = tree.files;
+    expect(files).toContain('/bar/.gemini/GEMINI.md');
+    expect(tree.readContent('/bar/.gemini/GEMINI.md')).toMatch(/^You are an expert in TypeScript/);
```
JeanMeche: @dgp1130 Is this the location we would want or the workspace root ?

packages/schematics/angular/ng-new/index.ts
```diff
@@ -60,11 +61,18 @@ export default function (options: NgNewOptions): Rule {
     zoneless: options.zoneless,
   };
 
+  const configOptions: ConfigOptions = {
+    project: options.name,
+    type: ConfigType.Ai,
+    tool: options.aiConfig as unknown as AiTool,
```
JeanMeche: I'm not sure why I couldn't get the enums to match. It works fine for the packageManager above for example. (Yes I also tried with removing "none" from the enum here in ng-new).

packages/schematics/angular/ng-new/index_spec.ts
dgp1130: I think `bar` is the workspace root in this context and is probably right, unless @clydin sees something off.

lacolaco:
How does it work if there is already a rule file in the project? Can it merge them into the single file?

JeanMeche:
@lacolaco Currently it would fail with a "merge" conflict. We plan to address the "update" of those files in a follow-up PR.

packages/schematics/angular/ng-new/index.ts
```diff
@@ -60,11 +61,17 @@ export default function (options: NgNewOptions): Rule {
     zoneless: options.zoneless,
   };
 
+  const configOptions: ConfigOptions = {
+    project: options.name,
+    tool: [options.aiConfig as unknown as AiTool],
```
JeanMeche: Not exactly sure why this type assertion was necessary.

packages/schematics/angular/ng-new/schema.json
```diff
@@ -144,6 +144,12 @@
       "x-prompt": "Do you want to create a 'zoneless' application without zone.js (Developer Preview)?",
       "type": "boolean",
       "default": false
+    },
+    "aiConfig": {
+      "description": "Create an AI configuration file for the project. This file is used to improve the outputs of AI tools by following the best practices.",
+      "default": "none",
```
alan-agius4: Is the “none”, truly needed? Can’t we just set a default to “null” and not add it in the enum or omit the default.

packages/schematics/angular/ng-new/index.ts
alan-agius4: This is due to the “none” value.

packages/schematics/angular/collection.json
```diff
@@ -131,6 +131,11 @@
       "factory": "./config",
       "schema": "./config/schema.json",
       "description": "Generates a configuration file."
+    },
+    "ai-config": {
+      "factory": "./ai-config",
+      "schema": "./ai-config/schema.json",
+      "description": "Generates an ai tool configuration file."
```
alan-agius4:
```suggestion
      "description": "Generates an AI tool configuration file."
```

packages/schematics/angular/ng-new/index.ts
JeanMeche: Even without it, the types mismatch. 🤔

packages/schematics/angular/ng-new/schema.json
```diff
@@ -144,6 +144,16 @@
...
+    },
+    "aiConfig": {
+      "type": "array",
+      "default": [],
+      "uniqueItems": true,
+      "description": "Create an AI configuration file for the project. This file is used to improve the outputs of AI tools by following the best practices.",
+      "items": {
+        "type": "string",
+        "enum": ["gemini", "copilot", "claude", "cursor", "windsurf"]
+      }
```
dgp1130, line 153: Question: Do we need `x-prompt` here too to show up as a prompt for just `ng new`?

dgp1130 (approved):

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  frontmatter: `---\ncontext: true\npriority: high\nscope: project\n---`,
+};
+
+const AI_TOOLS = {
+  'gemini': geminiFile,
+  'claude': claudeFile,
+  'copilot': copilotFile,
+  'cursor': cursorFile,
+  'windsurf': windsurfFile,
+};
```
alan-agius4: NIT, but I think it makes the code more readable.

```suggestion
const AI_TOOLS = {
  gemini: { 
    rulesName: 'GEMINI.md', 
    directory: '.gemini' 
  },
  claude: { 
    rulesName: 'CLAUDE.md', 
    directory: '.claude' 
  },
  copilot: {
    rulesName: 'copilot-instructions.md',
    directory: '.github',
  },
  windsurf: {
    rulesName: 'guidelines.md',
    directory: path.join('.windsurf', 'rules'),
  },
  // Cursor file has a front matter section.
  cursor: {
    rulesName: 'cursor.mdc',
    directory: path.join('.cursor', 'rules'),
    frontmatter: `---\ncontext: true\npriority: high\nscope: project\n---`,
  },
};
```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  return addAiContextFile(options);
+}
+
+interface ContextFileInfo {
+  rulesName: string;
+  directory: string;
+  frontmatter?: string;
+}
+
+function addAiContextFile(options: ConfigOptions): Rule {
```
alan-agius4:
```suggestion
export default function(options: ConfigOptions): Rule {
```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,91 @@
...
+  'claude': claudeFile,
+  'copilot': copilotFile,
+  'cursor': cursorFile,
+  'windsurf': windsurfFile,
+};
+
+export default function (options: ConfigOptions): Rule {
+  return addAiContextFile(options);
+}
+
```
alan-agius4:
```suggestion
```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    (selectedTool: keyof typeof AI_TOOLS) => AI_TOOLS[selectedTool],
+  );
+
+  return async (host) => {
+    const workspace = await readWorkspace(host);
+    const project = workspace.projects.get(options.project);
+    if (!project) {
+      throw new SchematicsException(`Project name "${options.project}" doesn't not exist.`);
+    }
+
```
alan-agius4: This workspace logic seems unused.

```suggestion
  return (host) => {

```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    const rules = files.map(({ rulesName, directory, frontmatter }) =>
+      mergeWith(
+        apply(url('./files'), [
+          // Keep only the single source template
+          filter((p) => p.endsWith('__rulesName__.template')),
+          applyTemplates({
+            ...strings,
+            rulesName,
+            frontmatter: frontmatter ?? '',
```
alan-agius4: NIT: 

```suggestion
    const rules = files.map(({ rulesName, directory, frontmatter = '' }) =>
      mergeWith(
        apply(url('./files'), [
          // Keep only the single source template
          filter((p) => p.endsWith('__rulesName__.template')),
          applyTemplates({
            ...strings,
            rulesName,
            frontmatter,
```

packages/schematics/angular/ai-config/index_spec.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  it('should create multiple files when multiple tools are selected', async () => {
+    const tree = await runConfigSchematic([
+      ConfigTool.Gemini,
+      ConfigTool.Copilot,
+      ConfigTool.Cursor,
+    ]);
+    expect(tree.readContent('.gemini/GEMINI.md').length).toBeGreaterThan(0);
+    expect(tree.readContent('.github/copilot-instructions.md').length).toBeGreaterThan(0);
+    expect(tree.readContent('.cursor/rules/cursor.mdc').length).toBeGreaterThan(0);
```
alan-agius4:
```suggestion
    expect(tree.exists('.gemini/GEMINI.md')).toBeTrue();
    expect(tree.exists('.github/copilot-instructions.md')).toBeTrue();
    expect(tree.exists('.cursor/rules/cursor.mdc')).toBeTrue();
```

packages/schematics/angular/ai-config/files/__rulesName__.template
```diff
@@ -0,0 +1,54 @@
...
+
+## Angular Best Practices
+
+- Always use standalone components over NgModules
+- Must NOT set `standalone: true` inside Angular decorators. It's the default.
+- Use signals for state management
+- Implement lazy loading for feature routes
+- Do NOT use the `@HostBinding` and `@HostListener` decorators. Put host bindings inside the `host` object of the `@Component` or `@Directive` decorator instead
+- Use `NgOptimizedImage` for all static images.
+  - `NgOptimizedImage` does not work for inline base64 images.
```
alan-agius4, line 19:
```suggestion
- `NgOptimizedImage` does not work for inline base64 images.
```

packages/schematics/angular/ai-config/index_spec.ts
```diff
@@ -0,0 +1,97 @@
...
+  beforeEach(async () => {
+    const workspaceTree = await schematicRunner.runSchematic('workspace', workspaceOptions);
+    applicationTree = await schematicRunner.runSchematic(
+      'application',
+      defaultAppOptions,
+      workspaceTree,
+    );
+  });
+
+  it('should create a GEMINI.MD file', async () => {
```
alan-agius4, line 34: In most of the cases here, we can replace `tree.readContent` with `tree.exists` as we are not checking for text that is generated.

packages/schematics/angular/ai-config/schema.json
```diff
@@ -0,0 +1,29 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "SchematicsAngularConfig",
+  "title": "Angular Config File Options Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "description": "Generates configuration files for your project. These files control various aspects of your project's build process, testing, and browser compatibility. This schematic helps you create or update essential configuration files with ease.",
+  "properties": {
+    "project": {
```
alan-agius4: This option seems to be redundant, as the config is not generated for a project but for a workspace.

packages/schematics/angular/ai-config/schema.json
```diff
@@ -0,0 +1,29 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "SchematicsAngularConfig",
+  "title": "Angular Config File Options Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "description": "Generates configuration files for your project. These files control various aspects of your project's build process, testing, and browser compatibility. This schematic helps you create or update essential configuration files with ease.",
```
alan-agius4: The description seems wrong.

packages/schematics/angular/ng-new/index.ts
```diff
@@ -60,11 +61,19 @@ export default function (options: NgNewOptions): Rule {
     zoneless: options.zoneless,
   };
 
+  const configOptions: ConfigOptions | undefined = options.aiConfig?.length
+    ? {
+        project: options.name,
+        tool: options.aiConfig as unknown as AiTool[],
```
alan-agius4: casting should not longer be needed.

```suggestion
        tool: options.aiConfig,
```

packages/schematics/angular/ng-new/schema.json
alan-agius4: If we want to show the prompt here, we'd need to remove the default and run the schematic directly and not conditionally.

alan-agius4 (changes requested):
See above, and commit message has the wrong scope as it should be "@schematics/angular" and description.

packages/schematics/angular/ng-new/index.ts
JeanMeche:
```
    packages/schematics/angular/ng-new/index.ts:64:9 - error TS2322: Type '{ tool: AiConfig[]; } | undefined' is not assignable to type 'Schema | undefined'.
      Type '{ tool: AiConfig[]; }' is not assignable to type 'Schema'.
        Types of property 'tool' are incompatible.
          Type 'AiConfig[]' is not assignable to type 'Tool[]'.
            Type 'AiConfig' is not assignable to type 'Tool'.
    
    64   const configOptions: ConfigOptions | undefined = options.aiConfig?.length
```

packages/schematics/angular/ng-new/index.ts
```diff
@@ -60,11 +61,20 @@ export default function (options: NgNewOptions): Rule {
...
+    : undefined;
+
   return chain([
     mergeWith(
       apply(empty(), [
         schematic('workspace', workspaceOptions),
         options.createApplication ? schematic('application', applicationOptions) : noop,
+        options.aiConfig && !options.aiConfig.includes(AiConfig.None) && configOptions
+          ? schematic('ai-config', configOptions)
+          : noop,
```
JeanMeche: I'm not sure why, but with this chain, the AI prompt comes 3rd (after zoneless, but before CSS/SSR).

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  applyTemplates,
+  chain,
+  filter,
+  mergeWith,
+  move,
+  strings,
+  url,
+} from '@angular-devkit/schematics';
+import { posix as path } from 'node:path';
+import { getWorkspace as readWorkspace } from '../utility/workspace';
```
alan-agius4: Redundant import
```suggestion
```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,81 @@
...
+import { Schema as ConfigOptions, Tool } from './schema';
+
+const AI_TOOLS = {
```
alan-agius4: Add a type here to avoid having to add types in code below;

```suggestion
const AI_TOOLS: { [key in Tool]: ContextFileInfo } = {
```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  const selectedTools: Tool[] = options.tool;
+  const files: ContextFileInfo[] = selectedTools.map(
+    (selectedTool: keyof typeof AI_TOOLS) => AI_TOOLS[selectedTool],
+  );
+
+  const rules = files.map(({ rulesName, directory, frontmatter }) =>
+    mergeWith(
+      apply(url('./files'), [
+        // Keep only the single source template
+        filter((p) => p.endsWith('__rulesName__.template')),
```
alan-agius4: Looks like there is no reason for this. 

```suggestion

```

packages/schematics/angular/ai-config/index_spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    const tree = await runConfigSchematic([ConfigTool.Gemini]);
+    tree.exists('.gemini/GEMINI.md');
```
alan-agius4: Missing expect

packages/schematics/angular/ai-config/schema.json
```diff
@@ -0,0 +1,22 @@
...
+  "title": "Angular Config File Options Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "description": "Generates AI configuration files for Angular projects. This schematic creates configuration files that help AI tools follow Angular best practices, improving the quality of AI-generated code and suggestions.",
+  "properties": {
+    "tool": {
+      "type": "array",
+      "uniqueItems": true,
+      "x-prompt": "What AI tools do you want to configure with Angular best practices? https://angular.dev/ai/develop-with-ai",
+      "description": "The AI tool for which the configuration file is being generated. This can include tools like Gemini, Copilot, Claude, Cursor, or Windsurf.",
```
alan-agius4:
```suggestion
   "description": "Specifies which AI tools to generate configuration files for. These file are used to improve the outputs of AI tools by following the best practices.",
```

The later part is a bit redundant as the values will be displayed next to description and the wording is "singular"

packages/schematics/angular/ng-new/index_spec.ts
```diff
@@ -103,4 +103,16 @@ describe('Ng New Schematic', () => {
...
+  it('should add ai config file when aiConfig is set', async () => {
+    const options = { ...defaultOptions, aiConfig: ['gemini', 'claude'] };
+
+    const tree = await schematicRunner.runSchematic('ng-new', options);
+    const files = tree.files;
+    expect(files).toContain('/bar/.gemini/GEMINI.md');
+    expect(files).toContain('/bar/.claude/CLAUDE.md');
+
+    expect(tree.readContent('/bar/.gemini/GEMINI.md')).toMatch(/^You are an expert in TypeScript/);
+    expect(tree.readContent('/bar/.claude/CLAUDE.md')).toMatch(/^You are an expert in TypeScript/);
```
alan-agius4: We do not really care about the contents of the file in this test.

```suggestion
```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,81 @@
...
+interface ContextFileInfo {
+  rulesName: string;
+  directory: string;
+  frontmatter?: string;
+}
+
+export default function (options: ConfigOptions): Rule {
+  const selectedTools: Tool[] = options.tool;
+  const files: ContextFileInfo[] = selectedTools.map(
+    (selectedTool: keyof typeof AI_TOOLS) => AI_TOOLS[selectedTool],
```
alan-agius4: This type here are redundant, and is due to `AI_TOOLS` not being typed.

```suggestion
    (selectedTool) => AI_TOOLS[selectedTool],
```

packages/schematics/angular/ai-config/schema.json
```diff
@@ -0,0 +1,22 @@
...
+  "$id": "SchematicsAngularConfig",
+  "title": "Angular Config File Options Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "description": "Generates AI configuration files for Angular projects. This schematic creates configuration files that help AI tools follow Angular best practices, improving the quality of AI-generated code and suggestions.",
+  "properties": {
+    "tool": {
+      "type": "array",
+      "uniqueItems": true,
+      "x-prompt": "What AI tools do you want to configure with Angular best practices? https://angular.dev/ai/develop-with-ai",
```
alan-agius4: I think `Which` is more grammatically correct in this case.

```suggestion
      "x-prompt": "Which AI tools do you want to configure with Angular best practices? https://angular.dev/ai/develop-with-ai",
```

packages/schematics/angular/ng-new/index.ts
```diff
@@ -60,11 +61,20 @@ export default function (options: NgNewOptions): Rule {
     zoneless: options.zoneless,
   };
 
+  const configOptions: ConfigOptions | undefined = options.aiConfig?.length
```
alan-agius4: Update this statement and remove in the below line;

```suggestion
  const configOptions: ConfigOptions | undefined = options.aiConfig && !options.aiConfig.includes(AiConfig.None)
```

packages/schematics/angular/ng-new/index.ts
```diff
@@ -60,11 +61,20 @@ export default function (options: NgNewOptions): Rule {
...
+        tool: options.aiConfig as unknown as AiTool[],
+      }
+    : undefined;
+
   return chain([
     mergeWith(
       apply(empty(), [
         schematic('workspace', workspaceOptions),
         options.createApplication ? schematic('application', applicationOptions) : noop,
+        options.aiConfig && !options.aiConfig.includes(AiConfig.None) && configOptions
```
alan-agius4: Redundant checks

```suggestion
        configOptions
```

packages/schematics/angular/ng-new/index.ts
alan-agius4: I think I mentioned in a previous comment that if a prompt was to be added we should re-use the prompt and avoid the noop. 

The prompt should not be added in the ng-new schema, but instead invoke the `ai-config` schematic and in turns will invoke it's prompt. 

What is happening here is that since the ai prompt is defined in the ng-new schema it will be prioritized over all the other prompts, stylesheet, SSR, and zoneless etc...

We should not make this prompt prioritised over the others.

alan-agius4: This is why in my previous comment I mention to re-use the prompt in ai-config schematic and remove the noop chain and not add a new one in ng-new. 

Any prompt defined in the ng-new schema will take precedence over the prompts defined in the child schematics.

This means that the AI prompt, will take presence over the stylesheet, ssr and zoneless questions.

packages/schematics/angular/ng-new/schema.json
```diff
@@ -144,6 +144,16 @@
       "x-prompt": "Do you want to create a 'zoneless' application without zone.js (Developer Preview)?",
       "type": "boolean",
       "default": false
+    },
+    "aiConfig": {
+      "type": "array",
+      "uniqueItems": true,
+      "description": "Create an AI configuration file for the project. This file is used to improve the outputs of AI tools by following the best practices.",
```
alan-agius4:
```suggestion
      "description": "Specifies which AI tools to generate configuration files for. These file are used to improve the outputs of AI tools by following the best practices.",
```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+interface ContextFileInfo {
+  rulesName: string;
+  directory: string;
+  frontmatter?: string;
+}
+
+export default function (options: ConfigOptions): Rule {
+  const selectedTools: Tool[] = options.tool;
+  const files: ContextFileInfo[] = selectedTools.map((selectedTool) => AI_TOOLS[selectedTool]);
```
alan-agius4: I think, if `none` is passed as a value, it will cause the schematic to error.

Probable this should exist early if `none` is passed and also maybe add a validation in the schema that does not allow `none` and some other value.


```
  "if": {
    "properties": {
      "aiConfig": {
        "contains": {
          "const": "none"
        }
      }
    },
    "required": [
      "aiConfig"
    ]
  },
  "then": {
    "properties": {
      "aiConfig": {
        "maxItems": 1
      }
    }
},
```

Would be good to add a test.

packages/schematics/angular/ng-new/index.ts
```diff
@@ -60,11 +61,16 @@ export default function (options: NgNewOptions): Rule {
...
+  const configOptions: AiConfigOptions | false =
+    options.aiConfig?.includes(AiConfig.None) !== true &&
+    (options.aiConfig as unknown as AiConfigOptions);
+
   return chain([
     mergeWith(
       apply(empty(), [
         schematic('workspace', workspaceOptions),
         options.createApplication ? schematic('application', applicationOptions) : noop,
+        configOptions ? schematic('ai-config', configOptions) : noop,
```
alan-agius4: Reduce duplicate logic, move the none to the ai-config schematic and always call the schematic

```suggestion
  return chain([
    mergeWith(
      apply(empty(), [
        schematic('workspace', workspaceOptions),
        options.createApplication ? schematic('application', applicationOptions) : noop,
        schematic('ai-config', { tool: options.aiConfig }),
```

packages/schematics/angular/ai-config/schema.json
```diff
@@ -0,0 +1,22 @@
...
+  "$id": "SchematicsAngularConfig",
+  "title": "Angular Config File Options Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "description": "Generates AI configuration files for Angular projects. This schematic creates configuration files that help AI tools follow Angular best practices, improving the quality of AI-generated code and suggestions.",
+  "properties": {
+    "tool": {
+      "type": "array",
+      "uniqueItems": true,
+      "x-prompt": "Which AI tools do you want to configure with Angular best practices? https://angular.dev/ai/develop-with-ai",
```
alan-agius4, line 13: NIT: As a follow up can we improve the prompt to include some more details

```
  "x-prompt": {
        "message": "Which AI tools do you want to configure with Angular best practices? https://angular.dev/ai/develop-with-ai",
        "type": "list",
        "items": [
          {
            "value": "none",
            "label": "None"
          },
          {
            "value": "copilot",
            "label": "Copilot powered IDEs     [ https://https://code.visualstudio.com/docs/copilot/copilot-customization#_custom-instructions ]"
          },
          {
            "value": "cursor",
            "label": "Cursor                   [ https://docs.cursor.com/en/context/rules                                                      ]"
          }
        ]
       ...
      },
```

alan-agius4 (commented):
Mostly NITs, but I think there is a bug with the "none" value.

alan-agius4 (changes requested):
There are still a couple of mistakes in the commit message (fixed below)


```
* `ng generate ai-config` to prompt support tools.
* `ng generate ai-config --tool=gemini` to specify the tool.
* `ng new` will prompt to config AI tools.
* `ng new --ai-config=gemini` to create a new project with AI configuration.

Supported AI tools: gemini, claude, copilot, windsurf, jetbrains, cursor.
```

packages/schematics/angular/ng-new/index.ts
```diff
@@ -22,9 +22,10 @@ import {
   NodePackageInstallTask,
   RepositoryInitializerTask,
 } from '@angular-devkit/schematics/tasks';
+import { Schema as AiConfigOptions } from '../ai-config/schema';
```
alan-agius4: Can be removed with the changed above.
```suggestion

```

packages/schematics/angular/ai-config/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  filter,
+  mergeWith,
+  move,
+  strings,
+  url,
+} from '@angular-devkit/schematics';
+import { posix as path } from 'node:path';
+import { Schema as ConfigOptions, Tool } from './schema';
+
+const AI_TOOLS: { [key in Tool]: ContextFileInfo } = {
```
alan-agius4: The build is failing because `none` is not added here. But I think it makes more sense to exclude it.
```suggestion
const AI_TOOLS: { [key in Exclude<Tool, Tool.None>]: ContextFileInfo } = {
```

packages/schematics/angular/ai-config/files/__rulesName__.template
```diff
@@ -0,0 +1,54 @@
...
+## Services
+
+- Design services around a single responsibility
+- Use the `providedIn: 'root'` option for singleton services
+- Use the `inject()` function instead of constructor injection
+
+## Common pitfalls
+
+- Control flow (`@if`):
+  - You cannot use `as` expressions in `@else if (...)`. E.g. invalid code: `@else if (bla(); as x)`.
```
alan-agius4: There is a PR in the FW to remove this: https://github.com/angular/angular/pull/63132

```suggestion
```

packages/schematics/angular/ai-config/files/__rulesName__.template
```diff
@@ -0,0 +1,54 @@
...
+## Components
+
+- Keep components small and focused on a single responsibility
+- Use `input()` and `output()` functions instead of decorators
+- Use `computed()` for derived state
+- Set `changeDetection: ChangeDetectionStrategy.OnPush` in `@Component` decorator
+- Prefer inline templates for small components
+- Prefer Reactive forms instead of Template-driven ones
+- Do NOT use `ngClass`, use `class` bindings instead
+- DO NOT use `ngStyle`, use `style` bindings instead
```
alan-agius4: NIT: can be as a followup

```suggestion
- Do NOT use `ngStyle`, use `style` bindings instead
```

packages/schematics/angular/ai-config/schema.json
JeanMeche: Will take care of it in a follow up.

packages/schematics/angular/ng-new/index.ts
```diff
@@ -65,6 +65,9 @@ export default function (options: NgNewOptions): Rule {
       apply(empty(), [
         schematic('workspace', workspaceOptions),
         options.createApplication ? schematic('application', applicationOptions) : noop,
+        schematic('ai-config', {
+          tool: options.aiConfig?.length ? options.aiConfig : [AiConfig.None],
```
alan-agius4, line 69: This is not right as this will cause the prompt never to be displayed.

```suggestion
          tool: options.aiConfig,
```

packages/schematics/angular/ng-new/index.ts
```diff
@@ -24,7 +24,7 @@ import {
 } from '@angular-devkit/schematics/tasks';
 import { Schema as ApplicationOptions } from '../application/schema';
 import { Schema as WorkspaceOptions } from '../workspace/schema';
-import { Schema as NgNewOptions } from './schema';
+import { AiConfig, Schema as NgNewOptions } from './schema';
```
alan-agius4:
```suggestion
import { Schema as NgNewOptions } from './schema';
```

packages/schematics/angular/ng-new/index.ts
alan-agius4: This will cause the prompt not to be ever displayed as a value is always provided.

```suggestion
        tool: options.aiConfig?.length ? options.aiConfig : undefined,
```

packages/schematics/angular/ai-config/schema.json
```diff
@@ -0,0 +1,39 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "SchematicsAngularConfig",
+  "title": "Angular Config File Options Schema",
+  "type": "object",
+  "additionalProperties": false,
+  "description": "Generates AI configuration files for Angular projects. This schematic creates configuration files that help AI tools follow Angular best practices, improving the quality of AI-generated code and suggestions.",
+  "properties": {
+    "tool": {
+      "type": "array",
+      "uniqueItems": true,
```
alan-agius4, line 11: It occuried to me that this will fail in non TTY env as it does not have a default.

```
"default": "none",
```

alan-agius4 (changes requested):
2 comments to fix the prompt, as I doubt this works right now.

alan-agius4 (approved):

PR #30766, opened 2025-07-22
test(@angular/build): add unit tests for unit-test builder options

clydin:
Adds a suite of disabled unit tests for the options of the unit-test builder. These tests cover the functionality of each option and will be enabled in a future commit.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/tests/options/browsers_spec.ts
+++ b/packages/angular/build/src/builders/unit-test/tests/options/browsers_spec.ts
@@ -1,0 +1,75 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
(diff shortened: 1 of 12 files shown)


packages/angular/build/src/builders/unit-test/tests/options/browsers_spec.ts
```diff
@@ -0,0 +1,75 @@
...
+import { execute } from '../../index';
+import {
+  BASE_OPTIONS,
+  describeBuilder,
+  UNIT_TEST_BUILDER_INFO,
+  setupApplicationTarget,
+} from '../setup';
+
+describeBuilder(execute, UNIT_TEST_BUILDER_INFO, (harness) => {
+  xdescribe('Option: "browsers"', () => {
```
hybrist, line 18: I think `pending()` worked here without failing the test suite? IIRC that's what I used for the disabled tests for the karma-on-application-builder feature.

jkrems (approved):

PR #30772, opened 2025-07-23
Bump version to "v20.1.2" with changelog.

jkrems:


hybrist:
The Changelog lint may be failing. I'll look into a follow-up PR to ensure the the `ng-dev` versions are in sync between the different active branches.

PR #30781, opened 2025-07-24
fix(@angular/build): update `vite` to `7.0.6`

### LINKED ISSUE #30731: @angular/build 20.1.0 totally breaks console logs line numbers in browser devtools
### Command

serve

### Is this a regression?

_No response_

### Description

Hello,

after upgrading to minor version of @angular/build, whenever in my app I use console.log, in console it shows totally random line of code, for example i put console.log in my main.ts file on line 6 and in chrome devtool console, there is line 15 for example, before I used 20.0.3 version and that worked totally fine. Sometimes also happens that certain console logs are in console written as totally another file, for example I console.log something in my Directive1 and inside console log is shown in Directive2 on different line.

Took me several hours before i found that problem is definitely somewhere inside minor update of @angular/build/....

downgrade to @angular/buils@20.0.3 works as expected

### Minimal Reproduction

just run any angular project with several services and you will see, that if you console.log in some service, lets say in constructor, in devtool console, you will see different line or even file

### Exception or Error

```text

```

### Your Environment

```text
@angular-devkit/architect    0.2001.1
@angular-devkit/core         20.1.1
@angular-devkit/schematics   20.1.1
@angular/build               20.1.1
@angular/cdk                 20.1.2
@angular/google-maps         20.1.2
@angular/material            20.1.2
@schematics/angular          20.1.1
```

### Anything else relevant?

_No response_

alan-agius4:
This fixes an issue with sourcemaps

Closes #30731

### DIFF
--- a/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
+++ b/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
@@ -7,6 +7,9 @@
  */
 
 import { lookup as lookupMimeType } from 'mrmime';
+import { createHash } from 'node:crypto';
+import { readFileSync } from 'node:fs';
+import type { ServerResponse } from 'node:http';
 import { extname } from 'node:path';
 import type { Connect, ViteDevServer } from 'vite';
 import { AngularMemoryOutputFiles, AngularOutputAssets, pathnameWithoutBasePath } from '../utils';
@@ -16,6 +19,8 @@ export interface ComponentStyleRecord {
   used?: Set<string>;
   reload?: boolean;
 }
+
+const JS_TS_REGEXP = /\.[cm]?[tj]sx?$/;
 
 export function createAngularAssetsMiddleware(
   server: ViteDevServer,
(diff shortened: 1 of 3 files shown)


packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts
```diff
@@ -7,6 +7,9 @@
  */
 
 import { lookup as lookupMimeType } from 'mrmime';
+import { createHash } from 'node:crypto';
```
hybrist, line 10: This seems unrelated to the version bump / upstream fix for the source map issue. At least https://github.com/angular/angular-cli/pull/30770 appears to pass CI without further changes. Can you add some context to the PR description?

alan-agius4: This is because it was needed to bump to vite 7.0.5.

See: https://github.com/angular/angular-cli/pull/30758

hybrist: Gotcha, so effectively this is a revert. Thanks!

PR #30803, opened 2025-07-30
test(@angular/cli): add initial e2e test for MCP server tool registration

clydin:
Introduces the first end-to-end test for the MCP server. The new test, `registers-tools.ts`, verifies that the MCP server correctly initializes and registers its default set of tools. This provides a foundation for more comprehensive e2e testing of the MCP server and its features.

### DIFF
--- a/tests/legacy-cli/e2e/tests/mcp/registers-tools.ts
+++ b/tests/legacy-cli/e2e/tests/mcp/registers-tools.ts
@@ -1,0 +1,47 @@
+import { chdir } from 'process';
+import { exec, ProcessOutput, silentNpm } from '../../utils/process';
+import assert from 'node:assert/strict';
+
+const MCP_INSPECTOR_PACKAGE_NAME = '@modelcontextprotocol/inspector-cli';
+const MCP_INSPECTOR_PACKAGE_VERSION = '0.16.2';
+const MCP_INSPECTOR_COMMAND_NAME = 'mcp-inspector-cli';
+
+async function runInspector(...args: string[]): Promise<ProcessOutput> {
+  const result = await exec(
+    MCP_INSPECTOR_COMMAND_NAME,
+    '--cli',
+    'npx',
+    '--no',
+    '@angular/cli',
+    'mcp',
+    ...args,
+  );
+
+  return result;
+}
+
+export default async function () {
+  await silentNpm(
+    'install',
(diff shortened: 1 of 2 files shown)


tests/legacy-cli/e2e/tests/mcp/registers-tools.ts
```diff
@@ -0,0 +1,41 @@
...
+
+async function runInspector(...args: string[]): Promise<ProcessOutput> {
+  const result = await exec(
+    MCP_INSPECTOR_COMMAND_NAME,
+    '--cli',
+    'npx',
+    '--no',
+    '@angular/cli',
+    'mcp',
+    ...args,
```
dgp1130, line 17: Question: How is `npx` pulling in the locally built version of the CLI? Is this going through Verdaccio through an `.npmrc` configuration or something like that?

I'm wondering if there's a more direct implementation through something like: `mcp-inspector-cli --cli node_modules/.bin/ng mcp ...args`?

tests/legacy-cli/e2e/tests/mcp/registers-tools.ts
```diff
@@ -0,0 +1,41 @@
...
+  );
+
+  return result;
+}
+
+export default async function () {
+  await silentNpm('install', '--ignore-scripts', '-g', MCP_INSPECTOR_PACKAGE_NAME);
```
dgp1130: Consider: Should this be more strongly versioned? What happens when `@modelcontextprotocol/inspector-cli` releases a new major version?

dgp1130 (approved):

tests/legacy-cli/e2e/tests/mcp/registers-tools.ts
clydin: Done.

tests/legacy-cli/e2e/tests/mcp/registers-tools.ts
clydin: `npx` uses the locally installed version if present, then the global.  If neither are present, it then tries to download from the registry.  However, the `--no` option prevents that last step.  The package should always be present so the failure with the `--no` option effectively acts as an assertion for the test setup.
The arguments are based on the recommended usage/configuration as well to better represent real world use.

clydin:

PR #30825, opened 2025-08-01
refactor(@angular/cli): centralize environment variable parsing

clydin:
Refactors the environment options utility to improve maintainability and clarity. This change introduces a centralized `parseTristate` function to handle all boolean-like parsing from environment variables, removing the previous `isEnabled`, `isDisabled`, and `optional` helpers.


packages/angular/cli/src/utilities/environment-options.ts
```diff
@@ -6,28 +6,63 @@
...
+  const value = variable.toLowerCase();
+  if (TRUTHY_VALUES.has(value)) {
+    return true;
+  }
+  if (FALSY_VALUES.has(value)) {
+    return false;
+  }
+
+  return undefined;
 }
```
dgp1130, line 44: Consider: Could we remove the "tristate" concept by just providing a default value like `parseEnvBoolean(process.env['FOO'], true /* default */)`?

packages/angular/cli/src/utilities/environment-options.ts
```diff
@@ -6,28 +6,63 @@
...
 
-export const analyticsDisabled = isDisabled(process.env['NG_CLI_ANALYTICS']);
-export const isCI = isEnabled(process.env['CI']);
-export const disableVersionCheck = isEnabled(process.env['NG_DISABLE_VERSION_CHECK']);
-export const ngDebug = isEnabled(process.env['NG_DEBUG']);
-export const forceAutocomplete = optional(process.env['NG_FORCE_AUTOCOMPLETE']);
+/** Disables all analytics reporting when the `NG_CLI_ANALYTICS` environment variable is set to '0' or 'false'. */
+export const analyticsDisabled = parseTristate(process.env['NG_CLI_ANALYTICS']) === false;
```
dgp1130, line 47: Consider: Would it make more sense to frame these in the positive and then fall back to a default value with `??` rather than check equality with a specific boolean? I feel `analyticsEnabled = parseTristate(...) ?? true` would be more intuitive to me at least.

clydin: This one is more of "force analytics disabled" and analytics being enabled involves further logic. There is potential to further refactor the usages and adjust the naming but i wanted to avoid usage changes where possible here.

packages/angular/cli/src/utilities/environment-options.ts
clydin: Some of them don't have constant defaults.  The third "not present" state is relevant downstream for them.

PR #30833, opened 2025-08-04
refactor(@schematics/angular): add Karma configuration analyzer and comparer

clydin:
Introduces a new utility for analyzing and comparing Karma configuration files. This is a foundational step for migrating from Karma to other testing frameworks like Jest and Vitest, as it allows the migration schematic to understand the user's existing setup.

The new `analyzeKarmaConfig` function uses TypeScript's AST parser to safely extract settings from a `karma.conf.js` file. It can identify common patterns, including `require` calls, and flags configurations that are too complex for static analysis.

The `compareKarmaConfigs` function provides the ability to diff a user's Karma configuration against a default template. This will be used to determine which custom settings need to be migrated.

Known limitations of the analyzer:
- It does not resolve variables or complex expressions. Any value that is not a literal (string, number, boolean, array, object) or a direct `require` call will be marked as an unsupported value.
- It does not support Karma configuration files that use ES Modules (import/export syntax).

### DIFF
--- a/packages/schematics/angular/migrations/karma/karma-config-analyzer.ts
+++ b/packages/schematics/angular/migrations/karma/karma-config-analyzer.ts
@@ -1,0 +1,159 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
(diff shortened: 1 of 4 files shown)


canterwilliam:
I don't know

PR #30840, opened 2025-08-06
fix(@angular/cli): cache MCP best practices content and add tool annotations

clydin:
Caches the content of the best practices instructions to avoid redundant file reads on subsequent uses of the tool. This can provide a minor performance improvement in cases where the tool is used multiple times during a single CLI process.

Also, annotations have been added to the text to provide additional context for the assistant and allow for more tailored display of the information.


devversion (approved):

clydin:

PR #30843, opened 2025-08-06
fix(@angular/build): correct JS/TS file paths when running under Bazel

clydin:
The compiler plugin now rewrites Bazel file paths if they were resolved outside of the execution root.
This behavior is only used if Bazel environment variables are present.


clydin:

- main: 261dbb37cbe01492240c4cedc644663b15a4296a
- 20.2.x: 06a6ddc102f5dc9018ec982f6e4cf56259cc4b52

PR #30845, opened 2025-08-06
feat(@angular/cli): add modernize tool to the MCP server

amishne:


packages/angular/cli/src/commands/mcp/tools/modernize.ts
```diff
@@ -0,0 +1,192 @@
...
+    const structuredContent = {
+      instructions: allInstructions.length ? allInstructions : undefined,
+    };
+
+    return {
+      content: [{ type: 'text' as const, text: JSON.stringify(structuredContent) }],
+      structuredContent,
+    };
+  } catch (e) {
+    const message = e instanceof Error ? e.message : 'An unknown error occurred.';
```
clydin: The SDK appears to handle this automatically.  I don't think the try/catch block is needed here.

packages/angular/cli/src/commands/mcp/tools/modernize.ts
```diff
@@ -0,0 +1,192 @@
...
+            type: 'text' as const,
+            text: JSON.stringify({
+              instructions,
+            }),
+          },
+        ],
+        structuredContent: {
+          instructions,
+        },
+      };
```
clydin: This logic appears repeated in this function. Maybe consider refactoring this function into something similar to `findInstructions` that returns an array of instructions. Then handle MCP output structuring in the tool handler?

packages/angular/cli/src/commands/mcp/tools/modernize.ts
```diff
@@ -0,0 +1,192 @@
...
+import { z } from 'zod';
+
+interface Transformation {
+  name: string;
+  description: string;
+  documentationUrl: string;
+  instructions?: string;
+}
+
+const TRANSFORMATIONS: Array<Transformation> = [
```
clydin, line 19: This should be good to test and experiment.  Long term we probably want to try to pull the information from `@angular/core` directly at runtime.  This would avoid needing to keep this information synchronized.

clydin (approved):

packages/angular/cli/src/commands/mcp/tools/modernize.ts
amishne: Done

packages/angular/cli/src/commands/mcp/tools/modernize.ts
amishne: Done

packages/angular/cli/src/commands/mcp/tools/modernize.ts
amishne: Ack, though long term I feel we probably also want to suggest things beyond just the schematics, so it won't necessarily ever align.

PR #30858, opened 2025-08-08
refactor(@angular/cli): declaratively register MCP server tools

clydin:
Changes to a declarative approach for registering tools with the MCP server.
Previously, each tool was registered imperatively. This change refactors each tool to export a `ToolDeclaration` object, which encapsulates its name, description, schema, and factory function. A new central `registerTools` function now iterates over these declarations, simplifying the server setup and ensuring a consistent registration process.
This approach improves maintainability, readability, and type safety by co-locating all aspects of a tool's definition.

Introduces a `--read-only` flag to the `ng mcp` command. When this flag is present, the MCP server will only register tools that are explicitly marked as read-only.
This provides a way for host applications to connect to the Angular CLI MCP server with a restricted set of capabilities, ensuring that no tools capable of modifying the user's workspace are exposed.

Introduces a `--local-only` flag to the `ng mcp` command. When this flag is present, the MCP server will only register tools that are explicitly marked as `isLocalOnly`.
This provides a way for host applications to connect to the Angular CLI MCP server in an offline or restricted environment, ensuring that no tools that require internet access are exposed.


packages/angular/cli/src/commands/mcp/tools/modernize.ts
```diff
@@ -125,38 +125,34 @@ export async function runModernization(input: ModernizeInput) {
...
+    'them with the latest features, such as the new built-in control flow syntax.\n\n' +
+    '* When the user asks for a specific transformation: When the transformation list is populated, ' +
+    'these specific ones will be ran on the inputs.\n' +
+    '</Use Cases>\n' +
+    '<Transformations>\n' +
+    TRANSFORMATIONS.map((t) => `* ${t.name}: ${t.description}`).join('\n') +
+    '\n</Transformations>\n',
+  annotations: {
+    readOnlyHint: true,
+  },
```
devversion: should this also have the explicit `isLocalOnly` etc?

clydin: good catch. yes. it should

PR #30868, opened 2025-08-11
refactor(@angular/cli): add a get Zoneless/OnPush MCP tool

atscott:
This change adds a tool that identifies the next steps for migrating a project file or directory to Zoneless or OnPush. The priorities for migration are:

1. Find and report any unsupported uses of ZoneJS APIs. These are easily verifiable. If they exist, they need to be removed and the tool provides suggested replacements
2. Provide requirements for OnPush compatibility for any files with a Component. It is suggested to use an explicit `ChangeDetectionStrategy.Default` until it can be verified the migration is complete. The tool skips any components with explicit change detection strategy definitions. This is required since we have no way of statically verifying a component is compatible with OnPush, so we need some way to indicate the tool should move on from a component
3. When nothing is identified in the above two steps, move on to test files and suggest migrating those to use zoneless. This is the best method to verify that components are zoneless compatible.


packages/angular/cli/src/commands/mcp/tools/onpush-zoneless-migration/zoneless-migration.ts
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { RequestHandlerExtra } from '@modelcontextprotocol/sdk/shared/protocol';
+import { ServerNotification, ServerRequest } from '@modelcontextprotocol/sdk/types';
+import { glob } from 'fast-glob';
+import * as fs from 'fs';
+import ts from 'typescript';
```
clydin: If the MCP server is run outside a project, `typescript` won't be available.  This is appears to be the source for some of the CI failures.  Probably would need to dynamically import within the tool itself.

packages/angular/cli/src/commands/mcp/tools/onpush-zoneless-migration/zoneless-migration.ts
```diff
@@ -0,0 +1,180 @@
...
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { RequestHandlerExtra } from '@modelcontextprotocol/sdk/shared/protocol';
+import { ServerNotification, ServerRequest } from '@modelcontextprotocol/sdk/types';
+import { glob } from 'fast-glob';
```
clydin: You can use `import { glob } from 'node:fs/promises';` here since this is experimental (Node.js v22+ has support). Return value is slightly different but doesn't require an extra dependency.

packages/angular/cli/src/commands/mcp/tools/onpush-zoneless-migration/zoneless-migration.ts
```diff
@@ -0,0 +1,180 @@
...
+import * as fs from 'fs';
```
clydin:
```suggestion
import * as fs from 'node:fs';
```

packages/angular/cli/src/commands/mcp/tools/onpush-zoneless-migration/migrate_test_file.ts
```diff
@@ -0,0 +1,82 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { glob } from 'fast-glob';
+import * as fs from 'fs';
```
clydin:
```suggestion
import * as fs from 'node:fs';
```

packages/angular/cli/src/commands/mcp/tools/onpush-zoneless-migration/migrate_test_file.ts
```diff
@@ -0,0 +1,82 @@
...
+import { dirname, join } from 'path';
```
clydin:
```suggestion
import { dirname, join } from 'node:path';
```

packages/angular/cli/src/commands/mcp/tools/onpush-zoneless-migration/ts_utils.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import * as fs from 'fs';
```
clydin:
```suggestion
import * as fs from 'node:fs';
```

dgp1130:

- main: e8111a97247e77ac8f4cd4564064698ce080852e
- 20.2.x: 1a1826cc99e05ca11abe478a8181050e9f35d814

PR #30882, opened 2025-08-12
Update Prettier defaults to enforce single quotes and 100 character line width

suhasrk:
## PR Checklist

Please check to confirm your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/angular/angular-cli/blob/main/CONTRIBUTING.md#-commit-message-guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Currently, the prettier block in package.json does not fully align with Angular’s recommended defaults as defined in .editorconfig.
This can lead to inconsistent formatting across projects, as common preferences like "singleQuote": true and "printWidth": 100 are not enforced in Prettier, even though .editorconfig specifies similar rules.

Issue Number: #30744

## What is the new behavior?

The prettier block in package.json has been updated to include Angular’s recommended defaults (singleQuote: true, printWidth: 100), ensuring consistency between .editorconfig and Prettier formatting across all projects.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

This change avoids manual intervention in minimal projects while preserving the intended HTML formatting benefits for standard projects.

PR #30884, opened 2025-08-12
fix(@schematics/angular): add extra prettier config

### LINKED ISSUE #30744: Unexpected Interaction Between Prettier Override and TypeScript Formatting in --minimal Template
### Command

new

### Is this a regression?

20.0.3

### Description

After `@angular/cli@20.0.4`, the following Prettier configuration was introduced in package.json via [PR #30590](https://github.com/angular/angular-cli/pull/30590), related to [issue #30548](https://github.com/angular/angular-cli/issues/30548) and commit [2316fe2](https://github.com/angular/angular-cli/commit/2316fe2):

```json
"prettier": {
    "overrides": [
      {
        "files": "*.html",
        "options": {
          "parser": "angular"
        }
      }
    ]
  }
```

This works well in standard Angular templates, which include an `.editorconfig` file in the root folder. However, in the `--minimal` template, there is no `.editorconfig` file, yet the same Prettier configuration is still present in `package.json`.

I’ve noticed that in `--minimal` projects, this Prettier override seems to cause TypeScript files to format using double quotes, even when my Prettier settings specify `singleQuote: true`. As soon as I remove the overrides block, Prettier respects the single quote rule again in `.ts` files.

This seems unintended, as the override is only targeting `*.html` files.

Image

### Minimal Reproduction

To reproduce the issue, install the latest version of Angular CLI and create a project using:

```bash
ng new <project-name> --minimal
```

Then open any `.ts` file and try saving it. You’ll observe that all single quotes are replaced with double quotes, as demonstrated in the GIF above.

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 20.1.1
Node: 22.14.0
Package Manager: npm 11.4.2
OS: darwin arm64

Angular:
...

@angular-devkit/architect    0.2001.1 (cli-only)
@angular-devkit/core         20.1.1 (cli-only)
@angular-devkit/schematics   20.1.1 (cli-only)
@schematics/angular          20.1.1 (cli-only)


Prettier Version 11.0.0
```

### Anything else relevant?

I tested with an older version of Angular CLI, specifically v20.0.3 (before commit [2316fe2](https://github.com/angular/angular-cli/commit/2316fe2)), and the issue did not occur there.

As a workaround, I added an additional overrides entry in `package.json` to explicitly force Prettier to use single quotes for `.ts` files:
```json
"prettier": {
    "overrides": [
      {
        "files": "*.html",
        "options": {
          "parser": "angular"
        }
      },
      {
        "files": "*.ts",
        "options": {
          "singleQuote": true,
          "jsxSingleQuote": true
        }
      }
    ]
  }
```

Image

This seems to restore the expected formatting behavior for TypeScript files, though I’m not certain whether this workaround could introduce unintended side effects. Further testing would be helpful.

jkrems:
*Rebase & squash of https://github.com/angular/angular-cli/pull/30882.*

Closes #30744

### DIFF
--- a/packages/schematics/angular/workspace/files/package.json.template
+++ b/packages/schematics/angular/workspace/files/package.json.template
@@ -9,6 +9,8 @@
     "test": "ng test"<% } %>
   },
   "prettier": {
+    "printWidth": 100,
+    "singleQuote": true,
     "overrides": [
       {
         "files": "*.html",


alan-agius4:

suhasrk:
Thank you all for the support and guidance throughout the PR process. I’m excited to contribute more regularly and help improve the project!
@jkrems @alan-agius4

PR #30889, opened 2025-08-13
fix(@schematics/angular): correct configure the `typeSeparator` in the library schematic

### LINKED ISSUE #30886: Issue with export paths when library generated using angular cli
### Command

generate

### Is this a regression?

^19

### Description

When a library is generated using `ng generate library`, there is an  Issue/bug with export path of the library module in `public-api.ts`.

### Minimal Reproduction

**To reproduce this issue:**

1. Go to this [stackblitz project](https://stackblitz.com/edit/stackblitz-starters-psnlkruq)
2. Open new terminal and run `ng generate library lib-gen-test --standalone=false`
3.  After library is successfully created, go to `public-api.ts` file of the new library and notice the issue with export module statement.

**New library name: "lib-gen-test, you should see a lib-gen-test.module.ts file created in the library.**

**Expectation:**
I should see export * from './lib/lib-gen-test.module'; in public-api.ts file.

**Current Behavior/Issue:**
I see export * from './lib/lib-gen-test-module'; in public-api.ts file.

**Instead of dot(.) i am seeing a dash(-) in module path.**

### Exception or Error

```text

```

### Your Environment

```text
@angular-devkit/architect    0.2000.0
@angular-devkit/core         20.0.0
@angular-devkit/schematics   20.0.0
@schematics/angular          20.0.0
typescript                   5.8.2
zone.js                      0.15.0
```

### Anything else relevant?

_No response_

alan-agius4:
Prior to this the `typeSeparator` could have been inherited from the workspace setting which would cause the file references to be incorrect.

Closes #30886

### DIFF
--- a/packages/schematics/angular/library/index.ts
+++ b/packages/schematics/angular/library/index.ts
@@ -188,6 +188,9 @@ export default function (options: LibraryOptions): Rule {
             flat: true,
             path: sourceDir,
             project: packageName,
+            // Explicitly set the `typeSeparator` this also ensures that the generated files are valid even if the `module` schematic
+            // inherits its `typeSeparator` from the workspace.
+            typeSeparator: '-',
           }),
       schematic('component', {
         name: options.name,


alan-agius4:

PR #30905, opened 2025-08-13
fix(@schematics/angular): allow AI config prompt to be skipped without selecting a value

alan-agius4:
With this change, we allow the AI config prompt to be skipped by pressing "ENTER" without selecting "None" and it will default to "None".


alan-agius4:

PR #30908, opened 2025-08-13
refactor(@angular/build): remove manual Node.js types for getCompileCacheDir

clydin:
The manual type declaration is no longer needed with newer versions of the Node.js types (`@types/node`).


alan-agius4:

PR #30909, opened 2025-08-13
refactor(@angular/build): align environment option parsing with CLI package

clydin:
The environment options in the build package are updated to align with the tristate parsing logic and environment variable checks found in the CLI package. This change introduces a `parseTristate` function to handle 'true', 'false', and undefined values from environment variables, providing more consistent behavior across the two packages.

### DIFF
--- a/packages/angular/build/src/utils/environment-options.ts
+++ b/packages/angular/build/src/utils/environment-options.ts
@@ -8,22 +8,47 @@
 
 import { availableParallelism } from 'node:os';
 
-function isDisabled(variable: string): boolean {
-  return variable === '0' || variable.toLowerCase() === 'false';
-}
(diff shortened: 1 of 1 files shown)


packages/angular/build/src/utils/environment-options.ts
```diff
@@ -8,22 +8,47 @@
...
+
+  const value = variable.toLowerCase();
+  if (TRUTHY_VALUES.has(value)) {
+    return true;
+  }
+  if (FALSY_VALUES.has(value)) {
+    return false;
+  }
+
+  return undefined;
```
hybrist, line 46: Is it worth adding a TODO here and/or to log a warning? Or is this mostly unreachable because other things would intercept `--flag=treu` already?

PR #30915, opened 2025-08-14
build: update `dev-infra` action SHA

alan-agius4:
This reduces the post cleanup step timings


alan-agius4:

PR #30920, opened 2025-08-14
test: update e2e update test to use angular 18

alan-agius4:
Angular 17 is no longer supported.


alan-agius4:

PR #30926, opened 2025-08-15
fix(@angular/cli): add choices to command line parser when type is array and has an enum

alan-agius4:
This commit enhances the command-line parser in `@angular/cli` by adding support for displaying available choices when a command option is of type `array` and has an `enum`.


jkrems (approved):

alan-agius4:

PR #30930, opened 2025-08-15
fix(@schematics/angular): improve AI config prompt wording

alan-agius4:
This improves the prompt wording and adds more context.


alan-agius4:

PR #30931, opened 2025-08-15
refactor(@angular/cli): add AI config tool to GA tracking

alan-agius4:
This commit adds the AI config tool option to GA tracking. Since this option is an array we had to work around a limitation that custom dimensions cannot have an array as value.


alan-agius4:
**Note:** Blocking until we get privacy approval.

dgp1130 (commented):
Implementation LGTM, but I'll hold off on approval until we go through the privacy process.

PR #30932, opened 2025-08-15
build: improve retry exec logic

alan-agius4:
Prior to this change, there was a flaw in its handling of the retry logic, specifically within the `setTimeout` block. The function had a recursive promise resolution problem and an incorrect handling of the rejected state.


clydin (approved):

alan-agius4:

PR #30936, opened 2025-08-15
test(@angular/build): add initial unit-test vitest runner E2E tests

clydin:
Two E2E tests have been added to test initial capabilities of the experimental unit-test builder with the vitest runner. This also ensures that the generated tests are compatible with both the karma/jasmine-based runner and the vitest runner.


alan-agius4 (approved):

alan-agius4:

PR #30943, opened 2025-08-18
fix(@angular/cli): apply default to array types

alan-agius4:
This commit fixes an issue where the `default` option was not being applied to `array` type options in yargs.

This seemingly minor change required refactoring in some tests, which revealed that a `.coerce` validation was incorrectly throwing an error on failure. The validation logic was moved to a `.check` to ensure proper error handling and prevent unexpected failures.


alan-agius4:

PR #30954, opened 2025-08-19
build: move verdaccio into `tests/package.json`

alan-agius4:
These dependencies are only needed for e2e tests


packages/angular/ssr/third_party/beasties/BUILD.bazel
```diff
@@ -27,9 +27,9 @@ rollup.rollup(
         "//:node_modules/@rollup/plugin-alias",
         "//:node_modules/@rollup/plugin-commonjs",
         "//:node_modules/@rollup/plugin-node-resolve",
-        "//:node_modules/beasties",
         "//:node_modules/rollup-license-plugin",
```
alan-agius4: Tried to moved these 2 but did not manage to get it working for now.

PR #30961, opened 2025-08-19
ci: update `MODULE.bazel.lock` when updates packages

alan-agius4:
Renovate does yet handle `MODULE.bazel.lock`. See: https://github.com/renovatebot/renovate/issues/25557

We do this as a workaround for that.


josephperrott (approved):
LGTM

PR #30971, opened 2025-08-20
ci: enable Renovate on non-base branches

alan-agius4:
This change allows Renovate to run on all branches, not just the base branch. Note that it will only update cross-repo, Bazel, and GitHub Actions dependencies on these branches.

The `postUpgradeTasks` has also been removed as it has been moved into the preset. See: https://github.com/angular/dev-infra/pull/2978


alan-agius4:
Blocking until https://github.com/angular/dev-infra/pull/2978 lands.

PR #31005, opened 2025-08-21
ci: set `E2E_SHARD_TOTAL` as env variable

alan-agius4:
Whlist this defaults to 1 when not supplied, it's cleaner that this is 1. Also, this is not an action input.


alan-agius4:

- main: 290ac55c31e86109334ccd1993bc8f4a0999722a
- 20.2.x: f9424244ce16142defd9da33a9f90a42005fab8a

Issue #30987, opened 2025-08-20
ng new fails when 'none' selected for AI best practices


davidfiala:
### Which @angular/* package(s) are the source of the bug?

core, compiler-cli

### Is this a regression?

No

### Description

When creating a new project and using the space bar to select the AI best practices to generate, such as moving the indicator down to 'Cursor', pressing space, and then pressing enter to confirm, `ng new` fails. See below:

```
ng new web
✔ Which stylesheet format would you like to use? Sass (Indented) [ https://sass-lang.com/documentation/syntax#the-indented-syntax ]
✔ Do you want to enable Server-Side Rendering (SSR) and Static Site Generation (SSG/Prerendering)? No
✔ Do you want to create a 'zoneless' application without zone.js? No
? Which AI tools do you want to configure with Angular best practices? https://angular.dev/ai/develop-with-ai
 ◉ None
 ◯ Claude                 [ https://docs.anthropic.com/en/docs/claude-code/memory                                 ]
❯◉ Cursor                 [ https://docs.cursor.com/en/context/rules                                              ]
 ◯ Gemini                 [ https://ai.google.dev/gemini-api/docs                                                 ]
 ◯ GitHub Copilot         [ https://code.visualstudio.com/docs/copilot/copilot-customization#_custom-instructions ]
 ◯ JetBrains AI Assistant [ https://www.jetbrains.com/help/junie/customize-guidelines.html                        ]
 ◯ Windsurf               [ https://docs.windsurf.com/windsurf/cascade/memories#rules                             ]
```

Above I highlighted Cursor, pressed `Space` to select, then pressed `Enter` to confirm.

```
ng new web
✔ Which stylesheet format would you like to use? Sass (Indented) [ https://sass-lang.com/documentation/syntax#the-indented-syntax ]
✔ Do you want to enable Server-Side Rendering (SSR) and Static Site Generation (SSG/Prerendering)? No
✔ Do you want to create a 'zoneless' application without zone.js? No
✔ Which AI tools do you want to configure with Angular best practices? https://angular.dev/ai/develop-with-ai None, Cursor                 [ https://docs.cursor.com/en/context/rules                                              ]
Schematic input does not validate against the Schema: {"tool":["none","cursor"]}
Errors:

  Data path "/tool" must NOT have more than 1 items.
```

The problem appears to be that `None` remains selected.

[...]

PR #31023, opened 2025-08-24
fix(@schematics/angular): ensure 'none' is excluded when other aiConfig tools are selected

suhasrk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes

When running ng new, selecting None together with another AI tool (e.g., Cursor, Gemini) causes the CLI to pass both values into the ai-config schematic.
This produces a validation error:

`Schematic input does not validate against the Schema: {"tool":["none","cursor"]}
Errors:
  Data path "/tool" must NOT have more than 1 items.`

Issue Number: #30987

## What is the new behavior?

- None is now treated as mutually exclusive.
- If None is selected along with any other AI tool, it is automatically removed before schema validation.
- This prevents validation errors and aligns with user expectations.

- [ ] Yes
- [x] No

## Other information

This update improves the UX of ng new by ensuring that None cannot conflict with other AI tool selections.

### DIFF
--- a/packages/schematics/angular/ng-new/index.ts
+++ b/packages/schematics/angular/ng-new/index.ts
@@ -60,6 +60,10 @@ export default function (options: NgNewOptions): Rule {
     zoneless: options.zoneless,
   };
 
+  if (options.aiConfig?.includes('none') && options.aiConfig.length > 1) {
+    options.aiConfig = options.aiConfig.filter(t => t !== 'none');
+  }
+
   return chain([
     mergeWith(
       apply(empty(), [


suhasrk:
Hi maintainers, this PR is ready for review. Could you please assign reviewers?
@comewalk @abraham @zed-0xff @jesperronn

suhasrk:
> This change isn't enough, The schema itself currently forbids multiple selection that includes `None`:
> 
> https://github.com/angular/angular-cli/blob/cc05242aa8354f3280b5a70eb1f9c0e9d85f1408/packages/schematics/angular/ai-config/schema.json#L54-L69

Thanks for looking into the fix, @JeanMeche !
I have a different approach to resolve this issue:
1. Remove the if/then block that restricts "none" to max 1 item in packages/schematics/angular/ai-config/schema.json
2. Updating logic in packages/schematics/angular/ng-new/index.ts as follows
```ts
if (options.aiTools?.includes('none')) {
options.aiTools = ['none'];
} else {
options.aiTools = options.aiTools?.filter(t => t !== 'none');
}
```
With this change, we only need to handle "none" in one place, and the behavior will be:
- If "none" is selected, ignore all other tools.
- If other tools are selected, drop "none".

Please let me know your thoughts on this. 
Thanks.

alan-agius4:
@suhasrk 
The `if/then` in the schema block should be removed and instead of the logic shown, please just ignore any `none` values.

suhasrk:
Hello @alan-agius4 , thank you for your suggestion — it was really helpful.
I’ve applied the changes as per your recommendation. Could you please review them?

alan-agius4:
@suhasrk, these changes are still not fully correct as the filtering should not happen in the `ng-new` schematic. I also noticed that the default is not correctly being set as it should be an array. 

I went ahead and implemented the fix here: https://github.com/angular/angular-cli/pull/31027

Regardless of this thanks for your contribution.

suhasrk:
Absolutely no worries @alan-agius4  — thanks for addressing this and for the detailed feedback. I’ll look into contributing on other issues in the meantime.

PR #31027, opened 2025-08-25
fix(@schematics/angular): prevent AI config schematic from failing when 'none' and other AI tools are selected

alan-agius4:
When using 'ng new' with the interactive AI tool prompt, the 'none' option remains selected even after a user selects another tool. This results in a validation error because the schematic receives an array with both 'none' and the selected tool.

This fix filters out the 'none' value from the `tool` option if any other tool is also present, allowing the schematic to proceed.

Closes: #30987


alan-agius4:

- main: aed26c38803a465842ff128c3f81bd6984e1fe3d
- 20.2.x: e93919dea7df55a3aac2fa5c93c4560c50a2d749

PR #31038, opened 2025-08-26
ci: removing WSL for a simpler, more performant approach

alan-agius4:
We've been using a complex workflow with WSL to run our E2E tests on CI. This method, which involves building on Linux and running on Windows, has led to several issues, including:

- A complex and cumbersome setup.
- Performance overhead due to cross-OS file operations.
- Flakiness caused by symlink conversion problems.

This change overhauls the E2E test workflow to a much simpler and more stable two-step process: building natively on Linux and running natively on Windows.

This new approach significantly simplifies our CI setup by eliminating the [dev-infra wsl](https://github.com/angular/dev-infra/blob/main/github-actions/setup-wsl/action.yml) action and the tricky symlink conversions, leading to better performance and fewer test flakes.


alan-agius4:

- main: d2edcb43b722207f637b764816f02943a78a7ad1
- 20.2.x: ecc0198c226bc523e76a9c969628b319bbbca34f

PR #31040, opened 2025-08-26
build: update rolldown to v1.0.0-beta.34

clydin:


jsaguet:
Hi @clydin, sorry to tag you on a closed PR but the chunk optimizer seems broken since 20.2.0. (number of chunks stays the same) 
It may be related to the missing option `advancedChunks: { minSize: 8192 },` that you added here.
If that's the case, would it be possible to backport it to the 20.2 branch please ?

PR #31046, opened 2025-08-26
Improve existing project Tailwind CSS schematic integration

clydin:
The tailwind schematic's stylesheet injection logic has been updated to support a wider range of project configurations. Previously, it assumed a `styles.css` file was always present, which could cause issues in projects using CSS preprocessors like Sass/SCSS.
Instead of unconditionally creating a new `.postcssrc.json` file, the schematic now searches for `postcss.config.json` or `.postcssrc.json` in both the workspace and project roots. If an existing configuration file is found, it is updated with the required `@tailwindcss/postcss` plugin. A new configuration file is only created if one does not already exist.

### DIFF
--- a/packages/schematics/angular/tailwind/index.ts
+++ b/packages/schematics/angular/tailwind/index.ts
@@ -26,10 +26,12 @@ import {
   addDependency,
   updateWorkspace,
 } from '../utility';
+import { JSONFile } from '../utility/json-file';
 import { latestVersions } from '../utility/latest-versions';
 import { createProjectSchematic } from '../utility/project';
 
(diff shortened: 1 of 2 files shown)


packages/schematics/angular/tailwind/index.ts
```diff
@@ -17,53 +17,113 @@ import {
   strings,
   url,
 } from '@angular-devkit/schematics';
-import { DependencyType, ExistingBehavior, addDependency } from '../utility';
+import assert from 'node:assert';
+import { join } from 'path';
```
alan-agius4:
```suggestion
import { join } from 'node:path/posix';
```

alan-agius4 (approved):

PR #31049, opened 2025-08-26
perf(@schematics/angular): optimize AST traversal utilities

clydin:
Improves the performance of AST utility functions by reducing redundant work and memory allocations.

- The `findNodes` function is now memoized with a `WeakMap`-based cache. This prevents repeated traversals of the entire source file AST when searching for nodes of the same kind, which is a common pattern.
- The `getSourceNodes` function has been refactored to use a single array instead of two, reducing memory overhead and avoiding costly `shift()` operations.

### DIFF
--- a/packages/schematics/angular/utility/ast-utils.ts
+++ b/packages/schematics/angular/utility/ast-utils.ts
@@ -91,6 +91,11 @@ export function insertImport(
   );
 }
 
+const findNodesCache = new WeakMap<
+  ts.SourceFile,
+  Map<ts.SyntaxKind | ((node: ts.Node) => boolean), ts.Node[]>
+>();
+
 /**
  * Find all nodes from the AST in the subtree of node of SyntaxKind kind.
  * @param node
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):

packages/schematics/angular/utility/ast-utils.ts
```diff
@@ -168,20 +190,13 @@ export function findNodes<T extends ts.Node>(
...
-  while (nodes.length > 0) {
-    const node = nodes.shift();
 
-    if (node) {
-      result.push(node);
-      if (node.getChildCount(sourceFile) >= 0) {
-        nodes.unshift(...node.getChildren());
-      }
-    }
+  for (let i = 0; i < nodes.length; i++) {
```
hybrist, line 195: This might be worth a quick comment. I had to read this twice to see that it was safe (e.g. "note: nodes.length changes inside of the loop but we only append to the end" or something like that. Just because _usually_ this kind of loop would be a fixed bound.

PR #31050, opened 2025-08-27
fix(@angular/build): ensure karma polyfills reporter factory returns a value

### LINKED ISSUE #31039: Karma reporter:angular-polyfills factory does not return a reporter which breaks karma in some situations

Don't known / other

### Is this a regression?

Yes

### Description

When migrating/upgrading from angular 19 to 20 the update tool migrated the tests from `@angular-devkit/build-angular` to `@angular/build`

This means the webpack plugin is not used anymore, but a new angular-polyfills plugin is now used. Which is defined in `@angular/build` package. It's good to understand that karma uses the di library for it's dependency injection. The `AngularPolyfillsPlugin` is defined in such a way. The type is set to `factory` and a function is provided, which is run by karma. All is fine so far, however karma expects the factory function to return a reporter. This is currently not happening. It just returns undefined.

When using the `karma test explorer` plugin for `Visual Code` karma will load all defined reporters in the MultiReporter defined in `multi.js` in the karma project class. This will access the adapters property of every reporter and push to it.

The problem is, the factory function of the  `angular-polyfills` reporter returns undefined. Therefor kamra crashes on startup if run with multiple reporters, which it does in case of the `Karma Test Explorer` extension. If the factory function would return `{ adapters:[] }` all works fine and `Karma Test Explorer` is all happy again, because now karma does not crash when executed. 

However I am not an expert at all in karma reporters and the di library. So it is very well possible, some other solution, change is more appropriate to fix this issue. And I am fully aware karma is deprecated, but because there is no stable, non experimental replacement available just yet, I would really really appreciate it, if this plugin get fixed so karma will work again in scenarios other as `ng test`.

_No response_

```true

```

```true
Angular CLI: 20.1.6
Node: 22.16.0
Package Manager: npm 11.5.2
OS: win32 x64

Angular: 20.1.7

@angular-devkit/architect    0.2001.6
@angular-devkit/core         20.1.6
@angular-devkit/schematics   20.1.6
@angular/build               20.2.0
@angular/cdk                 20.2.0
@angular/cli                 20.1.6
@schematics/angular          20.1.6
```

### Anything else?

_No response_

clydin:
The factory function for the `reporter:angular--polyfills` Karma plugin did not return a value, which violates Karma's plugin contract. This could cause Karma to crash when multiple reporters were in use, particularly with tools that use Karma's multi-reporter functionality.

This change modifies the factory to return an object with an empty `adapters` array (`{ adapters: [] }`). This satisfies Karma's plugin requirements and ensures compatibility with the multi-reporter, preventing the crash.

Fixes #31039

### DIFF
--- a/packages/angular/build/src/builders/karma/application_builder.ts
+++ b/packages/angular/build/src/builders/karma/application_builder.ts
@@ -187,6 +187,9 @@ class AngularPolyfillsPlugin {
             included: true,
             watched: false,
           });
+
+          // Karma needs a return value for a factory and Karma's multi-reporter expects an `adapters` array
+          return { adapters: [] };
         }, AngularPolyfillsPlugin),
       ],
     };


clydin:

- main: 4b49a207a1de27b82416c6225a59bc10f48bdcbc
- 20.2.x: b6816b0cbaf1262d7015b9d7f7fb425f53995947

PR #31053, opened 2025-08-27
feat(@schematics/angular): support different file name style guides in `ng new`

### LINKED ISSUE #30594: CLI: [command new] add option to enable previous suffix generation on create

compiler-cli

### The problem:

**Starting in Angular v20, by default Angular CLI will not generate suffixes**

As written in https://blog.angular.dev/announcing-angular-v20-b5c9c06cf301

>Starting in Angular v20, by default Angular CLI will not generate suffixes for your components, directives, services, and pipes. For existing projects, ng update will enable suffix generation by updating your angular.json. To enable suffix generation in new projects, use the following schematic configuration:

```
{
  "projects": {
    "app": {
      ...
      "schematics": {
        "@schematics/angular:component": { "type": "component" },
        "@schematics/angular:directive": { "type": "directive" },
        "@schematics/angular:service": { "type": "service" },
        "@schematics/angular:guard": { "typeSeparator": "." },
        "@schematics/angular:interceptor": { "typeSeparator": "." },
        "@schematics/angular:module": { "typeSeparator": "." },
        "@schematics/angular:pipe": { "typeSeparator": "." },
        "@schematics/angular:resolver": { "typeSeparator": "." }
      },
  ...
}
```

### The solution:

To add an option to the new command, to be able to create a new application with suffixes:

`ng new angular-app --with-suffix`

This command will automatically add **schematics** to the `angular.json` and creates e.g. `app.component.ts` instead off `app.ts`

### Alternatives considered

The only way to achive this, add manually the schematics and rename the auto generated files manually.
This is time consuming.

clydin:
Introduces the ability to configure the file naming convention for generated files directly within the `ng new` schematic. This allows users to create new workspaces that adhere to the 2016 style guide conventions, as an alternative to the default 2025 style guide. For more information, see the Angular Style Guide (https://angular.dev/style-guide).

When a user runs `ng new --file-name-style-guide 2016`:
- The `ng-new` schematic passes the style guide option down to the `application` sub-schematic.
- The `application` schematic configures the `schematics` section of the new `angular.json` to use the 2016 naming conventions for future `ng generate` commands.
- The `application` schematic generates the initial application files with the appropriate suffixes (e.g., `app.component.ts`).

Fixes #30594


giacomo:
@clydin @dgp1130 

just in case this are my scheme settings that i use:
```
"schematics": {
                    "type": "component",
                    "style": "scss",
                    "standalone": false
                },
                "@schematics/angular:directive": {
                    "type": "directive",
                    "standalone": false
                },
                "@schematics/angular:service": {
                    "type": "service"
                },
                "@schematics/angular:guard": {
                    "typeSeparator": "."
                },
                "@schematics/angular:interceptor": {
                    "typeSeparator": "."
                },
                "@schematics/angular:module": {
                    "typeSeparator": "."
                },
                "@schematics/angular:pipe": {
                    "typeSeparator": ".",
                    "standalone": false
                },
                "@schematics/angular:resolver": {
                    "typeSeparator": "."
                }
```

packages/schematics/angular/component/schema.json
```diff
@@ -95,6 +95,11 @@
       "type": "string",
       "description": "Append a custom type to the component's filename. For example, if you set the type to `container`, the file will be named `my-component.container.ts`."
     },
+    "addTypeToClassName": {
```
alan-agius4, line 98: From @dgp1130 [comment](https://github.com/angular/angular-cli/issues/30594#issuecomment-3025454681) I get that we should not do this part. 

> -file-name-style-guide 2016 would use the 2016 style guide to determine file names. In practice, this would look like:
> 
> Generates files with the old suffixes: app.component.{ts,html,css}, app.directive.ts, app.service.ts, ...
> Generates an angular.json like the one in the OP to configure future ng generate calls to match this convention.
> Use dots as separators rather than dashes (foo.guard.ts rather than foo-guard.ts).
> Does not generate a Prettier config in the package.json, as we can rely on the default behavior of Prettier for *.component.html files.
> Context here is that using app.html broke Prettier by default, requiring https://github.com/angular/angular-cli/pull/30590 in the package.json, but this workaround shouldn't be needed for projects using --file-name-style-guide 2016.
> Does not change symbol naming (still using UserProfiler over UserProfileComponent).
> This is useful to align with selectorless templates in the future to avoid uselessly repeating Component everywhere.

clydin: By default (and previously always), the `type` is added to both the filename and the symbol name.  To maintain previous before for when a type is specified and to also support the proposal above, an additional option is needed. For the 2016 style guide case, the `addTypeToClassName` is set to false which prevents the symbol names from being modified when `type` is specified for the file names.

alan-agius4 (approved):

PR #31063, opened 2025-08-28
fix(@angular/cli): don't set a default for array options when length is 0

alan-agius4:
This change prevents the CLI from setting a default value for array options that have a length of zero. Previously, this would result in an empty array being added  as a default, which isn't the intended behavior.


alan-agius4:

- main: 2c498d2b87c13a63bef2a9be2ca4f087c72c6b8a
- 20.2.x: a793bbc473dfaddf3fe6ed15805dc4fc84f52865

PR #31101, opened 2025-09-02
fix(@angular/build): avoid extra tick in SSR dev-server builds

alan-agius4:
In SSR applications, an unnecessary event loop tick during server startup could lead to an incorrect platform being initialized.

This change introduces an `ngJitMode` define, which is set to `false` during AOT builds. This allows for the JIT-specific code paths to not be followed, preventing the async operations that caused the extra tick. This ensures that the server platform is correctly and synchronously initialized.


jkrems (approved):

alan-agius4:

- main: 7b0f69798f061d5500620828cf304e05d667199f
- 20.2.x: 5c2abffea6cf3f672ee256a944dba56dd257665b

PR #31112, opened 2025-09-03
fix(@angular/ssr): adopt `bootstrapServerApplication` for streamlined  server bootstrap

alan-agius4:
This commit updates the SSR implementation to use the new `bootstrapServerApplication` function. This simplifies the server-side rendering (SSR) bootstrap process for standalone applications by making it more explicit and removing the need for a custom `main.server.ts` file that wraps `bootstrapApplication`.

Key changes:
- The SSR engine is updated to leverage the new bootstrap function.
- The `ng add @angular/ssr` schematic is updated to generate code using `bootstrapServerApplication`.
- Internal utilities and tests are updated to align with this new approach.

### DIFF
--- a/packages/angular/ssr/src/routes/ng-routes.ts
+++ b/packages/angular/ssr/src/routes/ng-routes.ts
@@ -634,7 +634,7 @@ export async function getRoutesFromAngularRouterConfig(
       const moduleRef = await platformRef.bootstrapModule(bootstrap);
       applicationRef = moduleRef.injector.get(ApplicationRef);
     } else {
-      applicationRef = await bootstrap();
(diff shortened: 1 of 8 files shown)


jkrems (approved):

hybrist:
Question: Do we also need to change `CommonEngineOptions#bootstrap`? https://github.com/angular/angular-cli/blob/2037b912b2f78eb4469d8671fbca8c43f06cd2ff/packages/angular/ssr/node/src/common-engine/common-engine.ts#L26

I'm not sure if our tests currently cover that API but it _looks_ like it should fail because there's no way to return an app ref from a `options.bootstrap` callback if we don't pass the new context?

alan-agius4:
Yes

PR #31119, opened 2025-09-04
fix(@angular/build): resolve PostCSS plugins relative to config file

alan-agius4:
Previously, PostCSS plugins were resolved from the location of `@angular/build`. This caused issues when the PostCSS configuration file was located in a subdirectory and used plugins not hoisted to the root `node_modules`, as plugins would not be found.

This change updates the PostCSS configuration loading to resolve plugins relative to the directory containing the PostCSS configuration file. This ensures that plugins are correctly located regardless of the configuration files location.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
+++ b/packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
@@ -9,7 +9,8 @@
 import type { OnLoadResult, Plugin, PluginBuild } from 'esbuild';
 import assert from 'node:assert';
 import { readFile } from 'node:fs/promises';
-import { extname } from 'node:path';
+import { createRequire } from 'node:module';
+import { dirname, extname } from 'node:path';
 import type { Options } from 'sass';
 import { glob } from 'tinyglobby';
 import { assertIsError } from '../../../utils/error';
@@ -69,7 +70,7 @@ export interface StylesheetPluginOptions {
    * initialized and used for every stylesheet. This overrides the tailwind integration
    * and any tailwind usage must be manually configured in the custom postcss usage.
    */
-  postcssConfiguration?: PostcssConfiguration;
+  postcssConfiguration?: { config: PostcssConfiguration; configPath: string };
 
   /**
    * Optional Options for configuring Sass behavior.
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@angular/devkit-repo",
-  "version": "21.0.0-next.2",
+  "version": "21.0.0-next.12",
```
clydin:
```suggestion
  "version": "21.0.0-next.2",
```

packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
```diff
@@ -215,14 +215,16 @@ export class StylesheetPluginFactory {
...
+      const postCssInstanceKey = JSON.stringify(config);
       let postcssProcessor = postcssProcessors.get(postCssInstanceKey)?.deref();
 
       if (!postcssProcessor) {
         postcss ??= (await import('postcss')).default;
         postcssProcessor = postcss();
-        for (const [pluginName, pluginOptions] of options.postcssConfiguration.plugins) {
-          const { default: plugin } = await import(pluginName);
+        for (const [pluginName, pluginOptions] of config.plugins) {
+          const resolvedPlugin = require.resolve(pluginName, { paths: [dirname(configPath)] });
```
clydin: Can you use `createRequire` from `node:module` here?
This avoids the dependence on on the CommonJS global `require` for future ESM support.

clydin (approved):

packages/angular/build/src/tools/esbuild/stylesheets/stylesheet-plugin-factory.ts
```diff
@@ -215,14 +216,18 @@ export class StylesheetPluginFactory {
...
+      const postCssInstanceKey = JSON.stringify(config);
       let postcssProcessor = postcssProcessors.get(postCssInstanceKey)?.deref();
 
       if (!postcssProcessor) {
         postcss ??= (await import('postcss')).default;
         postcssProcessor = postcss();
-        for (const [pluginName, pluginOptions] of options.postcssConfiguration.plugins) {
-          const { default: plugin } = await import(pluginName);
+
+        const postCssPluginRequire = createRequire(dirname(configPath) + '/');
```
th0r, line 227: @alan-agius4 this change brakes importing of plugins that are ES modules as in this case the `plugin` below won't be a function, but an `{default: <function>, __esModule: true}` object and it won't pass a check below.

PR #31144, opened 2025-09-09
fix(@angular-devkit/build-angular): avoid extra tick in SSR builds

alan-agius4:
In SSR applications, an unnecessary event loop tick during server startup could lead to an incorrect platform being initialized.

This change introduces an `ngJitMode` define, which is set to `false` during AOT builds. This allows for the JIT-specific code paths to not be followed, preventing the async operations that caused the extra tick. This ensures that the server platform is correctly and synchronously initialized.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/server/index.ts
+++ b/packages/angular_devkit/build_angular/src/builders/server/index.ts
@@ -219,6 +219,7 @@ async function initialize(
           {
             plugins: [
               new webpack.DefinePlugin({
+                'ngJitMode': false,
                 'ngServerMode': true,
               }),
             ],


packages/angular_devkit/build_angular/src/builders/server/index.ts
```diff
@@ -219,6 +219,7 @@ async function initialize(
           {
             plugins: [
               new webpack.DefinePlugin({
+                'ngJitMode': false,
```
alan-agius4, line 222: JIT is not supported with the webpack builder.

jkrems (approved):

alan-agius4:

- main: 9749ec687800c1bbeae4b75550dee3608bbe6823
- 20.2.x: 1a789087344aa94d061839122e6a63efbfc9c905

PR #31173, opened 2025-09-10
refactor(@angular/cli): Add instructions to component zoneless migrat…

atscott:
…ion to retain NgZone.run

I observed that it chose to do this in one test.  These should not be removed for libraries that support Zone applications and also should not be removed for applications until the zoneless migration is entirely complete.


alan-agius4:

- main: 47d0668b824827a22fde90c0f3bc23f30ec6eb5f
- 20.3.x: 4e51032e3e474d7e7a82a493098162a1dd15bb75

PR #31174, opened 2025-09-10
test(@angular/cli): add CI validation for MCP example code

clydin:
This commit introduces a new CI check to validate the correctness of code snippets and markdown structure within the `find_examples` markdown files.

To support this, the previous the `example_db_generator.js` script has been refactored and consolidated into a single, unified, mode-driven script: `process_examples.mjs`. This improves maintainability and ensures consistent parsing and validation logic.

The new script provides two new validation modes for CI:
- `validate-structure`: Parses all examples to validate their front matter and ensure they adhere to the required markdown heading structure.
- `validate-code`: Uses a pre-built Angular application harness to perform a full `ng build` on all example code, ensuring compilation and template correctness.


tools/example-validation-harness/package.json
```diff
@@ -0,0 +1,21 @@
+{
```
alan-agius4, line 1: NIT: should this has it's own pnpm lock and workspace file, seeing that there are a seperate `pnpm i` in the script?

.github/workflows/pr.yml
```diff
@@ -59,6 +59,8 @@ jobs:
         uses: angular/dev-infra/github-actions/linting/licenses@5043638fd8529765b375831a4679b9013141b326
       - name: Check tooling setup
         run: pnpm check-tooling-setup
+      - name: Validate MCP Example Markdown
```
alan-agius4, line 62: Nit: consider moving this to before the license checks so that it's always executed even when there are license failures, this is useful for automated dependency bumps. Potentially move `Check Package Licenses` as the last step

tools/example-validation-harness/angular.json
```diff
@@ -0,0 +1,28 @@
+{
```
alan-agius4, line 1: Instead of creating a application and commit it, should we use the current setup we have to generate an application on the fly? This is similar to other validations and it would make it easier to keep the app in code and deps in sync. 

See: https://github.com/angular/angular-cli/blob/a195db338465f8b01d5e14b723bea350edd98b56/scripts/json-help.mts#L34-L35

tools/process_examples.mjs
```diff
@@ -0,0 +1,389 @@
...
+  const mode = values.mode;
+
+  console.log(`Running example processor in '${mode}' mode.`);
+
+  try {
+    switch (mode) {
+      case 'validate-structure':
+        runValidateStructure(examplesPath);
+        break;
+      case 'validate-code':
```
alan-agius4, line 370: Where is this mode being used?

tools/example-validation-harness/angular.json
alan-agius4: Thinking more about this, instead of added another CI step, why not simply add a bazel test?

```diff
diff --git a/packages/angular/cli/lib/examples/BUILD.bazel b/packages/angular/cli/lib/examples/BUILD.bazel
new file mode 100644
index 000000000..b1f9df0c5
--- /dev/null
+++ b/packages/angular/cli/lib/examples/BUILD.bazel
@@ -0,0 +1,16 @@
+
+filegroup(
+    name = "examples",
+    srcs = glob( ["**/*.md"]),
+)
+
+sh_test(
+    name = "test",
+    srcs = ["//tools:ng_example_db"],
+    data = [":examples", "//tools:ng_example_db"],
+    args = [
+       "$(locations :examples)",
+        "--mode validate-structure",
+    ],
+)
diff --git a/tools/process_examples.mjs b/tools/process_examples.mjs
index 143f8d95a..2939bb9ea 100755
--- a/tools/process_examples.mjs
+++ b/tools/process_examples.mjs
@@ -162,14 +162,12 @@ function parseExampleFile(filePath) {
 }
 
 // --- Mode Implementations ---
-
 function runValidateStructure(examplesPath) {
   console.log('Validating markdown structure of all example files...');
-  const exampleFiles = globSync('**/*.md', { cwd: examplesPath });
-  for (const file of exampleFiles) {
-    parseExampleFile(join(examplesPath, file));
+  for (const file of examplesPath) {
+    parseExampleFile(file);
   }
-  console.log(`Successfully validated structure of ${exampleFiles.length} files.`);
+  console.log(`Successfully validated structure of ${examplesPath.length} files.`);
 }
 
 function runGenerateDb(examplesPath, outputPath) {
@@ -357,7 +355,8 @@ function main() {
       },
     },
   });
-  const examplesPath = positionals[0] ?? '.';
+
+  const examplesPath = positionals ?? '.';
   const mode = values.mode;
 
   console.log(`Running example processor in '${mode}' mode.`);
@@ -371,7 +370,7 @@ function main() {
         runValidateCode(examplesPath);
         break;
       case 'generate-db':
-        runGenerateDb(examplesPath, values.output);
+        runGenerateDb(examplesPath[0], values.output);
         break;
       default:
         throw new Error(

```

In a follow up we could also update `cli_example_db` to run the filegroup and runfiles.

IMO: I think it easier in terms of maintainability and future proofing to have all tests run with a single command and under bazel, instead of having multiple commands

tools/process_examples.mjs
```diff
@@ -0,0 +1,389 @@
...
+      experimental ? 1 : 0,
+      readFileSync(example.sourcePath, 'utf-8'),
+    );
+  }
+  db.exec('END TRANSACTION');
+  db.close();
+  console.log(`Successfully generated database with ${parsedExamplesForDb.length} examples.`);
+}
+
+function runValidateCode(examplesPath) {
```
alan-agius4, line 268: You we absolutely need to build the code using the CLI? Can't we just do it inside bazel using ng_project

```bazel
load("@rules_angular//src/ng_project:index.bzl", "ng_project")
package(default_visibility = ["//visibility:public"])

load("//tools:ts_json_schema.bzl", "ts_gen")

ts_extractor(
    name = "if-block-sources",
    src = "if-block.md",
)

ng_project(
    name = "if-block-example",
    srcs = [
      "if-block.ts",
    ],
    testonly = False,
    tsconfig= "//:test-tsconfig",
    deps = [
        "//:node_modules/@angular/common",
        "//:node_modules/@angular/compiler",
        "//:node_modules/@angular/compiler-cli",
        "//:node_modules/@angular/core",
        "//:node_modules/@angular/platform-browser",
        "//:node_modules/@angular/router",
    ],
)

```

`ts_extractor` should be simply the code in here to extract the TS from md and output a TS. 

Happy to help with this if you need.

PR #31185, opened 2025-09-11
refactor(@angular/build): decouple test discovery from runner execution

clydin:
This commit refactors the unit test builder to centralize test file discovery, preventing redundant operations in the test runner's execution phase.

The `TestRunner` API has been updated to pass the test entry point mappings from the initial build options phase directly to the executor. This removes the need for the Vitest executor to re-discover tests, simplifying its logic and adhering to the DRY principle.

The JSDoc comments for the runner API have also been updated to reflect these changes.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/builder.ts
+++ b/packages/angular/build/src/builders/unit-test/builder.ts
@@ -144,9 +144,8 @@ export async function* execute(
   const normalizedOptions = await normalizeOptions(context, projectName, options);
   const runner = await loadTestRunner(normalizedOptions.runnerName);
 
-  await using executor = await runner.createExecutor(context, normalizedOptions);
-
   if (runner.isStandalone) {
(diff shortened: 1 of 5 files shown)


packages/angular/build/src/builders/unit-test/builder.ts
```diff
@@ -174,9 +173,16 @@ export async function* execute(
   }
 
   // Get runner-specific build options from the hook
-  const { buildOptions: runnerBuildOptions, virtualFiles } = await runner.getBuildOptions(
+  const {
```
alan-agius4, line 176: Wouldn't this cause a bug where new spec files files will not discover unless the the process is terminated?

clydin: That's already the case here.  Watching for new tests is one of my next tasks.

PR #31220, opened 2025-09-16
refactor: use native browserslist 'baseline' query

alan-agius4:
With the release of browserslist v4.26.0, the 'baseline' query is now natively supported. This removes the need for a custom tool to generate the browserslist configuration.

See: https://github.com/browserslist/browserslist/pull/903


tonypconway:
Thanks for picking this up Alan, I was going to message you and @dgp1130 today to let you know it was live but it looks like you're all over it!

Small nit for accuracy, the version of browserslist is 4.26.0 not 4.24.0.  Worth updating in the description for https://github.com/ng-packagr/ng-packagr/pull/3122 as well.

packages/angular/build/src/utils/supported-browsers.ts
```diff
@@ -8,14 +8,14 @@
 
 import browserslist from 'browserslist';
 
+const BASELINE_DATE = 'BASELINE-DATE-PLACEHOLDER';
```
dgp1130, line 12: Suggestion: A comment that this will be replaced by `npm_package` might be helpful, otherwise this looks like dead code.

packages/schematics/angular/config/index.ts
```diff
@@ -37,13 +36,10 @@ export default createProjectSchematic<ConfigOptions>((options, { project }) => {
...
 
 async function addBrowserslistConfig(projectRoot: string): Promise<Rule> {
-  // Read Angular's default vendored `.browserslistrc` file.
-  const config = await readFile(path.join(__dirname, '.browserslistrc'), 'utf8');
-
   return mergeWith(
     apply(url('./files'), [
       filter((p) => p.endsWith('.browserslistrc.template')),
-      applyTemplates({ config }),
+      applyTemplates({ baselineDate: 'BASELINE-DATE-PLACEHOLDER' }),
```
dgp1130, line 43: Suggestion: Same comment point here.

dgp1130 (approved):
Thanks for taking this on @alan-agius4, looks great!

PR #31230, opened 2025-09-17
build: disable eslint `await-thenable`

alan-agius4:
This was recently introduced and there are a number of intentional parts that our code does not conform.


clydin (approved):

alan-agius4:
The changes were merged into the following branches: main, 20.3.x

PR #31231, opened 2025-09-17
fix(@angular/build): update `vite` to `6.3.6`

alan-agius4:
This fixes https://github.com/advisories/GHSA-g4jq-h2w9-997c


hybrist:
Note during release prep: The fixed version of Vite 5.x (5.4.20) is covered by the range on the v18 branch (~5.4).

PR #31233, opened 2025-09-17
build: enable minimumReleaseAge to mitigate dependency chain attacks

alan-agius4:
This change configures pnpm's `minimumReleaseAge` setting to 1 day (1440 minutes). This is a security measure to mitigate dependency chain attacks, where malicious actors publish a new version of a dependency with malicious code and then trick users into updating to it before it can be discovered and reported.

By delaying the adoption of new releases, we reduce the window of opportunity for such attacks. The list of excluded packages contains trusted and frequently updated dependencies from the Angular team, which are considered safe to use without this delay.


devversion (approved):

alan-agius4:
The changes were merged into the following branches: main, 20.3.x

PR #31259, opened 2025-09-21
refactor(@angular/build): remove deprecated signature usage

crisbeto:
Fixes usages of a deprecated signature of `createImportClause`.


packages/angular/build/src/tools/angular/transformers/jit-bootstrap-transformer.ts
```diff
@@ -32,7 +32,7 @@ export function replaceBootstrap(
             bootstrapImport = nodeFactory.createImportDeclaration(
               undefined,
               nodeFactory.createImportClause(
-                false,
```
crisbeto, line 35: We also need to make similar changes in the framework, but we can't until we've switched to TS 5.9 internally. As far as I can tell, this code isn't synced into g3.

PR #31268, opened 2025-09-22
build: update rolldown to v1.0.0-beta.38

clydin:


packages/angular/build/src/builders/application/chunk-optimizer.ts
```diff
@@ -252,7 +252,8 @@ export async function optimizeChunks(
     });
 
     const result = await bundle.generate({
-      minify: { mangle: false, compress: false, removeWhitespace: true },
+      minify: { mangle: false, compress: false },
+      advancedChunks: { minSize: 8192 },
```
sod, line 256: Did this work for you? When I do `NG_BUILD_OPTIMIZE_CHUNKS=true ng build` with `@angular/build@20.3.3`, which contains this PR, then I see:

```
[INVALID_OPTION] Warning: Advanced chunks options (minSize) specified without groups. These options have no effect without groups - you should either add groups to use advanced chunking or remove these options.
```

And the amount of chunks seem to be the same as without this setting.

PR #31278, opened 2025-09-23
fix(@angular/ssr): avoid retaining rendered HTML in memory post-request

### LINKED ISSUE #31277: SSR rendered HTML is retained in memory after each request

core

### Is this a regression?

Yes

### Description

There appears to be a memory leak in our Angular application. After each server-side rendered request, the generated HTML string is not being properly garbage-collected and remains in the server's memory. This causes the Node.js server's memory consumption to grow linearly with each page refresh, eventually leading to a crash.

The issue is reproducible in a minimal application that uses TransferState to pass a large array of data from the server to the client.

https://github.com/alaa-kh-fe/memory-test

git clone https://github.com/alaa-kh-fe/memory-test.git
cd memory-test
npm install
ng build --configuration=production --prerender=false
node --inspect dist/memory-test/server/server.mjs

https://github.com/alaa-kh-fe/memory-test

```true
Open Google Chrome and navigate to chrome://inspect. Connect to the running Node.js process.
Go to the Memory tab and take a heap snapshot.
In a separate browser tab, open http://localhost:4000/memory-01 and refresh the page 5-10 times.
Return to the inspector and take a second heap snapshot.
Compare the two snapshots. You will see multiple instances of the application's <!DOCTYPE html>... string retained in memory, each corresponding to a page request.
```

Image

```true
Angular version: 19.2.0
Node.js version: 22.16.0
Operating System: Windows
```

### Anything else?

_No response_

alan-agius4:
The previous implementation for server-side rendering could lead to memory leaks where the rendered HTML content was not properly garbage-collected after the HTTP request was fulfilled.

This was caused by inefficiencies in the critical CSS caching and how the response stream was handled. This commit addresses the issue by:

1. Refactoring the `criticalCssLRUCache` to store `Uint8Array` directly and use a more robust caching strategy.
2. Using the request URL as the cache key to prevent storing multiple cache entries for the same resource.
3. Ensuring the response stream controller enqueues the final HTML as a `Uint8Array`, avoiding unnecessary string conversions and dangling references.

These changes prevent the SSR response from being retained in memory, improving the stability and performance of the server.

Closes #31277

### DIFF
--- a/packages/angular/ssr/src/app.ts
+++ b/packages/angular/ssr/src/app.ts
@@ -143,13 +143,15 @@ export class AngularServerApp {
   private readonly textDecoder = new TextEncoder();
 
   /**
-   * Cache for storing critical CSS for pages.
-   * Stores a maximum of MAX_INLINE_CSS_CACHE_ENTRIES entries.
+   * A cache that stores critical CSS to avoid re-processing for every request, improving performance.
+   * This cache uses a Least Recently Used (LRU) eviction policy.
    *
-   * Uses an LRU (Least Recently Used) eviction policy, meaning that when the cache is full,
-   * the least recently accessed page's critical CSS will be removed to make space for new entries.
+   * @see {@link MAX_INLINE_CSS_CACHE_ENTRIES} for the maximum number of entries this cache can hold.
    */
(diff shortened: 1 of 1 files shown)


alan-agius4:

- main: afa273849d0e0e62a7df2236d818bf7800c3ad13
- 20.3.x: a4c9a2007ab3e33b2c97fa63f0df8f8662427031

Issue #31270, opened 2025-09-22
Injection Tokens are not tree-shaken because the constructor is considered impure by esbuild


eblocha:
### Which @angular/* package(s) are the source of the bug?

core

### Is this a regression?

No

### Description

If an injection token is defined as `const TOKEN = new InjectionToken('desc')`, then it will not be tree-shaken from the production bundle, because esbuild considers the constructor to have side-effects.

This means that if it uses a factory function, the items that factory function references are also included.

This can cause large services to be included from a library even if they are unused because they are the default implementation of a token.

This can be mitigated by marking the token definition with `/* @__PURE__ */` like so:

```ts
const TOKEN = /* @__PURE__ */ new InjectionToken('desc')
```

I believe this is something that should be fixed in either @angular/core or @angular/build, by marking the InjectionToken constructor as pure: https://esbuild.github.io/api/#pure.

### Please provide the exception or error you saw

```true

```

### Please provide the environment you discovered this bug in (run `ng version`)

```true
Angular CLI: 19.2.16
Node: 24.1.0 (Unsupported)
Package Manager: npm 11.3.0
OS: darwin arm64

Angular: 19.2.15
... common, compiler, compiler-cli, core, elements, forms
... platform-browser, platform-browser-dynamic, router

Package                      Version
------------------------------------------------------
@angular-devkit/architect    0.1902.16
@angular-devkit/core         19.2.16
@angular-devkit/schematics   19.2.16
@angular/build               19.2.16
@angular/cli                 19.2.16
@schematics/angular          19.2.16
ng-packagr                   19.2.2
rxjs                         7.8.1
typescript                   5.8.3
zone.js                      0.15.0
```

alan-agius4:
`@angular/build` includes logic to make these tree-shakable by adding the `/* @__PURE__ */` comment. Could you provide a reproduction case where they are not being tree-shaken?

eblocha:
I was able to reproduce it by setting up a new application with `ng new`, then changing main.ts to the following:

```ts
import { bootstrapApplication } from '@angular/platform-browser';
import { appConfig } from './app/app.config';
import { App } from './app/app';
import { InjectionToken } from '@angular/core';

const _TOKEN = new InjectionToken('retained');

bootstrapApplication(App, appConfig)
  .catch((err) => console.error(err));
```

After running `ng build`, the main chunk contains the token (I formatted the file):

[...]

PR #31286, opened 2025-09-24
fix(@angular/build): mark InjectionToken as pure for improved treeshaking

alan-agius4:
`new InjectionToken(...)` is not considered pure by default by build tools, which prevents it from being tree-shaken when unused. This can lead to unused services being included in production bundles if they are provided as a default value for a token.

This change marks `new InjectionToken(...)` calls as pure. The `InjectionToken` constructor is side-effect free; its only purpose is to create a token for the dependency injection system. Marking it as pure is safe and  allows build tools like esbuild to tree-shake unused tokens and their dependencies.

Closes https://github.com/angular/angular-cli/issues/31270

### DIFF
--- a/packages/angular/build/src/tools/babel/plugins/pure-toplevel-functions.ts
+++ b/packages/angular/build/src/tools/babel/plugins/pure-toplevel-functions.ts
@@ -6,12 +6,17 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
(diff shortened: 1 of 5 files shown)


alan-agius4:

- main: acd785afc956efad56b03402085ff94855b9fcc6
- 20.3.x: bc6b631146c719a337c937e95c7cc5ebca29254b

PR #31336, opened 2025-09-29
fix(@schematics/angular): use bracket notation for `process.env['pm_id']`

alan-agius4:
The property `pm_id` on `process.env` comes from an index signature, so it must be accessed with bracket notation (`['pm_id']`) to avoid TypeScript errors. This change corrects the access to `process.env['pm_id']` in the SSR server template.

`pm_id` is an environment variable that is set by PN2.


clydin (approved):

clydin:

- main: 57075a31ad8f95a82304fe8533b3bca828d8da42
- 20.3.x: 465436c9fa21173befe5e39b61afb7f29435c2aa

PR #31343, opened 2025-09-30
feat(@schematics/angular): add AGENTS.md support to ai-config schematic

clydin:
This change introduces support for generating an `AGENTS.md` file via the `ng generate ai-config` schematic. The `AGENTS.md` standard is an emerging convention for providing instructions to AI agents that operate within a codebase.

By adding this option, users can easily create a root-level `AGENTS.md` file, helping to configure AI agents with project-specific guidelines and best practices.

The implementation includes:
- Updating the schematic's `schema.json` to include `agents` as a selectable tool.
- Extending the logic in `index.ts` to handle the creation and placement of the file.
- Adding a unit test to ensure the file is generated correctly.

### DIFF
--- a/packages/schematics/angular/ai-config/schema.json
+++ b/packages/schematics/angular/ai-config/schema.json
@@ -19,6 +19,10 @@
             "label": "None"
           },
           {
+            "value": "agents",
+            "label": "Agents.md      [ https://agents.md/                                               ]"
+          },
+          {
             "value": "claude",
             "label": "Claude         [ https://docs.anthropic.com/en/docs/claude-code/memory            ]"
           },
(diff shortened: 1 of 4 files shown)


packages/schematics/angular/ai-config/schema.json
```diff
@@ -47,7 +51,7 @@
       "description": "Specifies which AI tools to generate configuration files for. These file are used to improve the outputs of AI tools by following the best practices.",
       "items": {
         "type": "string",
-        "enum": ["none", "gemini", "copilot", "claude", "cursor", "jetbrains", "windsurf"]
+        "enum": ["none", "gemini", "copilot", "claude", "cursor", "jetbrains", "windsurf", "agents"]
```
alan-agius4, line 54: This new value should also be added in the ng-new schematic.

PR #31344, opened 2025-09-30
feat(@angular/cli): overhaul `ng version` command output

clydin:
This commit completely revamps the `ng version` command's output for a more modern and readable experience.

The key improvements include:
- A single, unified table for all packages, removing the separate Angular section.
- A polished table format using box-drawing characters and consistent padding.
- An improved header section with aligned, bolded, and more formal labels.
- The use of color to highlight version numbers and other key information, improving scannability.
- Version lookup errors are now highlighted in red for better visibility.

The header generation logic is also refactored to be more maintainable by removing repeated label strings.


alan-agius4 (approved):
Nice!!!

PR #31358, opened 2025-10-01
refactor(@angular/build): switch back to rollup for chunk optimizer

clydin:
Rollup currently appears to be more effective at optimizing total output chunk count. While this is investigated further, the experimental chunk optimizer will be switched back to using rollup.

### DIFF
--- a/packages/angular/build/src/builders/application/chunk-optimizer.ts
+++ b/packages/angular/build/src/builders/application/chunk-optimizer.ts
@@ -19,7 +19,7 @@
 
 import type { Message, Metafile } from 'esbuild';
 import assert from 'node:assert';
-import { type OutputAsset, type OutputChunk, rolldown } from 'rolldown';
+import { type OutputAsset, type OutputChunk, rollup } from 'rollup';
 import {
   BuildOutputFile,
   BuildOutputFileType,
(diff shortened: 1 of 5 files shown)


TheAlexLichter:
Possibly related: https://github.com/rolldown/rolldown/issues/5726

Generally it would be interesting to see what the difference between rolldown & rollup was, ideally with tested code so we can have a look 👍 
With the `advancedChunks` API, there should be a lot of room for optimization that can be applied at the code base level (if you'd expose that option to the user).

packages/angular/build/src/builders/application/chunk-optimizer.ts
```diff
@@ -252,8 +252,7 @@ export async function optimizeChunks(
     });
 
     const result = await bundle.generate({
-      minify: { mangle: false, compress: false },
-      advancedChunks: { minSize: 8192 },
```
sapphi-red, line 256: This option won't have effect unless `group` is specified. There should be a warning about it (https://github.com/rolldown/rolldown/pull/5009).
A correct option would be like
```js
advancedChunks: { groups: [{ name: 'chunks' }], minSize: 8192 }
```
(you would probably want to add `maxSize` as well, otherwise a huge chunk will be generated)

PR #31359, opened 2025-10-01
test: avoid forcing color on Windows

clydin:
Forcing the the use of color in the console can lead to inconsistent behavior between platforms and result in potentially confusing output on terminals that do not support color.


scripts/windows-testing/parallel-executor.mjs
```diff
@@ -30,7 +30,6 @@ async function main() {
           PATH: process.env.PATH,
           E2E_SHARD_TOTAL: process.env.E2E_SHARD_TOTAL,
           E2E_SHARD_INDEX: process.env.E2E_SHARD_INDEX,
-          FORCE_COLOR: '3',
```
alan-agius4, line 33: Maybe `0` to forcefully disable it and have consistency between local and CI runs.

alan-agius4 (approved):

PR #31474, opened 2025-10-15
fix(@angular/ssr): prevent malicious URL from overriding host

### LINKED ISSUE #31464: SSR server side request forgery
### Command

other

### Is this a regression?

_No response_

### Description

When using Angular server side rendering you can trick Angular in thinking the request is for a different domain.

**Root cause**
The room cause seems to be in the createRequestUrl function that glues together the protocol, host, port and quest path (url or originalUrl) using the native URL class constructor. https://github.com/angular/angular-cli/blob/8f0f6a5f113ffc9e81d99eeeba71f8054e2d3686/packages/angular/ssr/node/src/request.ts#L104

```
return new URL(originalUrl ?? url, `${protocol}://${hostnameWithPort}`);
```

When originalUrl or url looks like a schema relative url ('//something') it will ignore the second base argument. Zo in our example it would convert to return new URL('//something', 'http://localhost:4200') which according to the specs resolves to http://something

**Attack scenario**
As Angular thinks (using the PlatformLocator or DOCUMENT token location.href) it is handling a domain that can be defined in the URL path. An attacker can let the server execute requests to a server of their choosing if relative requests are used, possibly injecting content.

Executing a request to http://localhost:4200//domain-from-attacker.com/some-path will in the above example execute a request to http://domain-from-attacker.com

### Minimal Reproduction

1. Create a new Angular SSR project using ng new --ssr  and activate RenderMode.Server. See the example repo at https://github.com/meDavidNS/angular-ssr

2. For illustration purpose inject the DOCUMENT token and the httpClient to make a relative request.
```
  protected readonly httpClient = inject(HttpClient)
  protected readonly doc = inject(DOCUMENT)
  protected readonly loc = inject(PlatformLocation)
  ngOnInit() {
    console.log({
      'HOSTNAME': this.loc.hostname,
      'LOCATION': this.doc.location.href,
    })
    this.httpClient.get('assets/some.json').subscribe(console.log)
  }

```
3. Request a page that starts the path with a double slash, like http://localhost:4200//something, this will trick Angular in thinking it is handling a request from http://something/ instead of http://localhost:4200. In the above example it will execute the http request to http://something/assets/some.json instead of http://localhost:4200/assets/some.json

### Your Environment

```text
Angular CLI: 20.3.5
Node: 22.14.0
Package Manager: npm 10.9.2
OS: darwin arm64
    

Angular: 20.3.4

```

### Anything else relevant?

_No response_

alan-agius4:
A request with a specially crafted URL starting with a double slash (e.g., `//example.com`) could cause the server-side rendering logic to interpret the request as being for a different host. This is due to the behavior of the `URL` constructor when a protocol-relative URL is passed as the first argument.

This vulnerability could be exploited to make the server execute requests to a malicious domain when relative paths are used within the application (e.g., via `HttpClient`), potentially leading to content injection or other security risks.

The fix ensures that the request URL is always constructed as a full URL string, including the protocol and host, before being passed to the `URL` constructor. This prevents the host from being overridden by the path.

Closes #31464

### DIFF
--- a/packages/angular/ssr/node/src/request.ts
+++ b/packages/angular/ssr/node/src/request.ts
@@ -76,7 +76,7 @@ function createRequestHeaders(nodeHeaders: IncomingHttpHeaders): Headers {
  * @param nodeRequest - The Node.js `IncomingMessage` or `Http2ServerRequest` object to extract URL information from.
  * @returns A `URL` object representing the request URL.
  */
-function createRequestUrl(nodeRequest: IncomingMessage | Http2ServerRequest): URL {
+export function createRequestUrl(nodeRequest: IncomingMessage | Http2ServerRequest): URL {
   const {
     headers,
     socket,
@@ -101,7 +101,7 @@ function createRequestUrl(nodeRequest: IncomingMessage | Http2ServerRequest): UR
     }
   }
 
-  return new URL(originalUrl ?? url, `${protocol}://${hostnameWithPort}`);
+  return new URL(`${protocol}://${hostnameWithPort}${originalUrl ?? url}`);
 }
 
 /**
--- a/packages/angular/ssr/node/test/request_spec.ts
+++ b/packages/angular/ssr/node/test/request_spec.ts
@@ -1,0 +1,158 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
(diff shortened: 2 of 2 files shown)


hybrist:

- main: 7be6c8f0e2883c85546eb1691c91fa7d4aefc0d3
- 20.3.x: 5271547c80662de10cb3bcb648779a83f6efedfb

PR #31476, opened 2025-10-15
fix(@angular/ssr): correctly handle routes with matrix parameters

### LINKED ISSUE #31457: Angular SSR Setup denies accepting matrix params

`@angular/ssr`

### Is this a regression?

Yes

### Description

Angular SSR seems not to accept matrix params in URLs.

Reproduction: App with SSR; then add Matrix params to the URL through the use of angulars `routerLink` like `<a [routerLink]="['.', {foo:'bar'}]">...</a>` - when clicked in running angular app it will work as expected since it runs only in the browser. 
when trying to request this new url with matrix params from the running ssr server it will fail because of those matrix params.

`/home` --> works
`/home;foo=bar` --> not working

https://stackblitz.com/edit/angular-ssr-matrix-params-denial

```true
`404 - Not Found`

-- if there is a redirect configured with `**` in the angular `Routes` it would return a 302 to given route.
```

```true
Angular CLI: 20.3.5
Node: 20.19.1
Package Manager: npm 10.8.2
OS: linux x64
    

Angular: 20.3.4
... common, compiler, compiler-cli, core, platform-browser
... platform-server, router

```

### Anything else?

When changing the route definition from using `path` to `matcher` function which does essentially the same as the path it works as expected.

```ts
{
  {
    // path: 'home',
    matcher: ([s0]) => (s0?.path === 'home' ? { consumed: [s0] } : null),
    loadChildren: () => import('./routes/home/home-routes')
  }
}
```

alan-agius4:
This commit introduces a change to strip matrix parameters from the URL before route matching in Angular SSR. Previously, URLs containing matrix parameters would fail to match their corresponding routes.

A new `stripMatrixParams` utility function has been added to remove these parameters, ensuring that route matching is not affected by their presence.

Closes #31457

### DIFF
--- a/packages/angular/ssr/src/routes/router.ts
+++ b/packages/angular/ssr/src/routes/router.ts
@@ -7,7 +7,7 @@
  */
 
 import { AngularAppManifest } from '../manifest';
-import { stripIndexHtmlFromURL } from '../utils/url';
+import { stripIndexHtmlFromURL, stripMatrixParams } from '../utils/url';
 import { extractRoutesAndCreateRouteTree } from './ng-routes';
 import { RouteTree, RouteTreeNodeMetadata } from './route-tree';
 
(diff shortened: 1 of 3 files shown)


atscott (approved):

alan-agius4:

- main: 85c18b4ead77c6c090b9f5c63e9e034e58369152
- 20.3.x: 3a28fb6a13061215b881c49232db979fc3c2f641

PR #31497, opened 2025-10-16
fix(@angular/ssr): ensure server-side navigation triggers a redirect

### LINKED ISSUE #31482: Server side router navigation with changes to query parameters not sending a redirect response
### Command

run

### Is this a regression?

_No response_

### Description

When pushing a new navigation on server side only with:
```ts
await this.router.navigate([], {
  queryParams: {filter: 'test'},
  queryParamsHandling: 'merge', // remove to replace all query params by provided
  relativeTo: route
});
```
This does not result in a redirect response being sent to the browser, just a regular 200.

It seems like, in the [check for whether the url change](https://github.com/angular/angular-cli/blob/main/packages/angular/ssr/src/utils/ng.ts#L113), the query parameters on finalUrl and initialUrl of the lastSuccessfulNavigation object are the same, which they shouldn't!

### Minimal Reproduction

Reproduction code: https://github.com/WizzardMaker/angular-ssr-navigate-bug

Steps to reproduce in the project:
1. Navigate to e.g. ´/?filter=a´
2. See that server response is of status code 200, not any other redirect code

In short what the project does:
A route component triggers a router navigation in it's constructor like this:
```ts
if(isPlatformBrowser(inject(PLATFORM_ID))) return;

this.pendingTasks.run(async () => {
  await this.router.navigate([], {
    queryParams: {filter: 'test'},
    queryParamsHandling: 'merge', // remove to replace all query params by provided
    relativeTo: route
  })
})
```

### Exception or Error

```text

```

### Your Environment

```text
                |___/
    


Angular CLI: 20.3.5
Node: 24.9.0
Package Manager: npm 11.6.0
OS: darwin arm64
    

Angular: 20.3.4

```

### Anything else relevant?

This happens only on server side rendering on a standard node server and on ng serve. I don't quite know whether this behaviour changed between versions

alan-agius4:
When a navigation occurs on the server-side, such as using `router.navigate`, and the final URL is different from the initial URL that was requested, the server should respond with a redirect.

Previously, the initial URL was being read from `router.lastSuccessfulNavigation.initialUrl`, which could be incorrect in scenarios involving server-side navigations, causing the comparison with the final URL to fail and preventing the redirect.

This change ensures that the initial URL requested by the browser is used for the comparison, correctly triggering a redirect when necessary.

Closes #31482

### DIFF
--- a/packages/angular/ssr/test/app_spec.ts
+++ b/packages/angular/ssr/test/app_spec.ts
@@ -11,7 +11,8 @@
 import '@angular/compiler';
 /* eslint-enable import/no-unassigned-import */
 
-import { Component } from '@angular/core';
+import { Component, inject } from '@angular/core';
+import { CanActivateFn, Router } from '@angular/router';
 import { AngularServerApp } from '../src/app';
 import { RenderMode } from '../src/routes/route-config';
 import { setAngularAppTestingManifest } from './testing-utils';
@@ -26,6 +27,31 @@ describe('AngularServerApp', () => {
     })
     class HomeComponent {}
 
+    @Component({
+      selector: 'app-redirect',
+    })
+    class RedirectComponent {
+      constructor() {
+        void inject(Router).navigate([], {
+          queryParams: { filter: 'test' },
+        });
+      }
+    }
+
+    const queryParamAdderGuard: CanActivateFn = (_route, state) => {
+      const urlTree = inject(Router).parseUrl(state.url);
+
+      if (urlTree.queryParamMap.has('filter')) {
+        return true;
+      }
+
+      urlTree.queryParams = {
+        filter: 'test',
+      };
+
+      return urlTree;
+    };
+
     setAngularAppTestingManifest(
       [
         { path: 'home', component: HomeComponent },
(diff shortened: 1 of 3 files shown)


packages/angular/ssr/test/app_spec.ts
```diff
@@ -26,14 +27,32 @@ describe('AngularServerApp', () => {
...
+    @Component({
+      selector: 'app-redirect',
+    })
+    class RedirectComponent {
+      constructor() {
+        const router = inject(Router);
+        const pendingTasks = inject(PendingTasks);
+
+        pendingTasks.run(() =>
+          router.navigate([], {
```
atscott: Can we maybe do this in a guard instead? As discussed, I don’t love that this causes the navigation to not complete successfully.

packages/angular/ssr/test/app_spec.ts
```diff
@@ -26,14 +27,32 @@ describe('AngularServerApp', () => {
...
 
+    @Component({
+      selector: 'app-redirect',
+    })
+    class RedirectComponent {
+      constructor() {
+        const router = inject(Router);
+        const pendingTasks = inject(PendingTasks);
+
+        pendingTasks.run(() =>
```
atscott: Nit: you don’t need the pending task. Router navigations are already tracked internally.

packages/angular/ssr/test/app_spec.ts
alan-agius4: I tried to keep this spec case as similar to the reported issue as possible.

atscott: Yea, I get that. Can you double check that this actually does also happen (as in it doesn’t work as a redirect) for navigations in guards? By far, that’s more common so at the very least, you’d want to cover that case too. This test is good on its own, it’s just that I wish it already worked correctly with the old code and didn’t need this PR

alan-agius4: Without this change, within a guard the navigation change does not work as a redirect.

atscott (approved):

PR #31499, opened 2025-10-16
feat(@angular/cli): add style language detection to list_projects tool

clydin:
This commit enhances the `list_projects` MCP tool by adding a `styleLanguage` field to the project output. This provides critical context for an AI model, enabling it to generate components with the correct stylesheet format (e.g., `.scss`, `.css`) without needing to parse additional configuration files. This makes the AI more autonomous and its code generation more accurate.

The detection logic uses a prioritized heuristic to determine the most likely style language:
1.  Checks for a project-specific schematic setting in `angular.json`.
2.  Checks for a workspace-level schematic setting.
3.  Infers from the `build` target's `inlineStyleLanguage` option.
4.  Infers from file extensions in the `build` target's `styles` array.
5.  As a future-proof fallback, checks for the existence of an implicit `styles.{ext}` file in the project's source root.

### DIFF
--- a/packages/angular/cli/src/commands/mcp/tools/projects.ts
+++ b/packages/angular/cli/src/commands/mcp/tools/projects.ts
@@ -15,6 +15,26 @@ import { AngularWorkspace } from '../../../utilities/config';
 import { assertIsError } from '../../../utilities/error';
 import { McpToolContext, declareTool } from './tool-registry';
 
+// Single source of truth for what constitutes a valid style language.
+const styleLanguageSchema = z.enum(['css', 'scss', 'sass', 'less']);
+type StyleLanguage = z.infer<typeof styleLanguageSchema>;
+const VALID_STYLE_LANGUAGES = styleLanguageSchema.options;
+
+// Explicitly ordered for the file system search heuristic.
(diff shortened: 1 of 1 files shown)


packages/angular/cli/src/commands/mcp/tools/projects.ts
```diff
@@ -15,6 +15,26 @@ import { AngularWorkspace } from '../../../utilities/config';
 import { assertIsError } from '../../../utilities/error';
 import { McpToolContext, declareTool } from './tool-registry';
 
+// Single source of truth for what constitutes a valid style language.
+const styleLanguageSchema = z.enum(['css', 'scss', 'sass', 'less']);
```
hybrist, line 19: Should we include tailwind here already or would that depend on support in the app creation flow first?

jkrems (approved):

packages/angular/cli/src/commands/mcp/tools/projects.ts
clydin: Was going to handle tailwind separate since the heuristics are more complicated and i'm undecided on whether it should be represented here or in another form.

PR #31502, opened 2025-10-16
refactor(@angular/ssr): replace incorrect check

alan-agius4:
This should be undefined and not null.


alan-agius4:
> Do we just want to make that `null`-ish (`!= null`) or would that be incorrect?

It can never be null

alan-agius4:
-

hybrist:
Was this closed on purpose?

PR #31503, opened 2025-10-16
refactor(@angular/ssr): replace logic to determine the original URL

alan-agius4:
Using `currentNavigation` might not work in all cases as this will be null when the navigate has already completed.

### DIFF
--- a/packages/angular/ssr/src/utils/ng.ts
+++ b/packages/angular/ssr/src/utils/ng.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import { APP_BASE_HREF, PlatformLocation } from '@angular/common';
+import { LocationStrategy } from '@angular/common';
 import {
   ApplicationRef,
   type PlatformRef,
@@ -21,7 +21,7 @@ import {
   platformServer,
   ɵrenderInternal as renderInternal,
 } from '@angular/platform-server';
-import { ActivatedRoute, Router } from '@angular/router';
+import { ActivatedRoute, Router, UrlSerializer } from '@angular/router';
 import { Console } from '../console';
 import { joinUrlParts, stripIndexHtmlFromURL } from './url';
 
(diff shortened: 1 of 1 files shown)


packages/angular/ssr/src/utils/ng.ts
```diff
@@ -96,31 +96,30 @@ export async function renderAngular(
...
-      if (initialUrl !== finalUrlStringified) {
-        const baseHref =
-          envInjector.get(APP_BASE_HREF, null, { optional: true }) ??
-          envInjector.get(PlatformLocation).getBaseHrefFromDOM();
+      const baseHref =
+        envInjector.get(APP_BASE_HREF, null, { optional: true }) ??
+        envInjector.get(PlatformLocation).getBaseHrefFromDOM();
+      const finalUrl = lastSuccessfulNavigation.finalUrl;
+      const finalUrlSerialized = envInjector.get(UrlSerializer).serialize(finalUrl);
+      const finalUrlWithBaseHref = joinUrlParts(baseHref, finalUrlSerialized);
```
atscott: This bit is a lot of what `Location.prepareExternalUrl` does by combining the base with the internal application path. Should we just use that? 

```
const finalUrlSerialized = envInjector.get(UrlSerializer).serialize(finalUrl);
const finalHref = new URL(Location.prepareExternalUrl(finalUrlSerialized), new URL(urlToRender).origin).href;
```

Below, the `initialUrl` comparison can't work if the base is actually a full URL and not just a path. So you'd likely want to compare the above href to `new URL(urlToRender).href`

alan-agius4: Let's do that.

atscott (approved):
Hope this works! 🤞  The URL construction part always confuses me...

alan-agius4:
@atscott, the unit tests says it works :) thanks for this.

alan-agius4:
Looks like the last change causes some E2Es to fail. I’ll take a look tomorrow.

PR #31510, opened 2025-10-17
fix(@angular/build): allow disabling sourcemaps in the dev server

### LINKED ISSUE #31331: ng serve with esbuild/Vite ignores sourceMap configuration, injecting large inline source maps into all files
### Command

serve

### Is this a regression?

_No response_

### Description

The development server (ng serve), which uses Vite/esbuild, consistently injects large, inline Base64 source maps into every JavaScript file it serves.

This behavior occurs even when source maps are explicitly disabled in angular.json and/or tsconfig.json. The most problematic aspect is that this transformation is also applied to pre-built, third-party packages from node_modules.

This unnecessarily inflates file sizes (often by 2-3x), which slows down page loads in development and makes debugging in browser DevTools more difficult due to the massive file content. The core issue is that the configuration to disable this behavior is not being respected. 

### Minimal Reproduction

1.  Create a new Angular project:
    ```bash
    ng new sourcemap-repro
    cd sourcemap-repro
    ```
2.  In `angular.json`, modify the configuration for the development server to explicitly disable all source maps. Under `projects.sourcemap-repro.architect.build.configurations.development`, set the `sourceMap` option:
    ```json
    "sourceMap": false
    ```
    *(Alternatively, using a more granular object like `{"scripts": false, "styles": false, "vendor": false}` also fails to prevent the issue.)*

3.  Run the development server:
    ```bash
    ng serve
    ```
4.  Open the application in a browser and launch the developer tools (e.g., Chrome DevTools).
5.  Navigate to the "Network" tab, refresh the page, and inspect any `.js` file (e.g., `main.js`, `polyfills.js`, or a chunk from `@angular/core`).
6.  View the "Response" or "Source" for that file.

**Expected Behavior:**
The served JavaScript files should **not** contain an inline source map. The file should end without a `//# sourceMappingURL=...` comment.

**Actual Behavior:**
Every JavaScript file has a large inline Base64 source map appended to it, directly contradicting the configuration in `angular.json`.

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 19.2.14
Node: 20.19.0
Package Manager: npm 10.8.2
OS: darwin arm64

Angular: undefined
...

@angular-devkit/architect       0.1902.15
@angular-devkit/build-angular   19.2.14
@angular-devkit/core            19.2.14
@angular-devkit/schematics      19.2.14
@angular/cli                    19.2.14
@angular/compiler               19.2.14
@angular/compiler-cli           19.2.14
@angular/language-service       19.2.14
@schematics/angular             18.2.11
ng-packagr                      18.2.1
typescript                      5.7.2
```

### Anything else relevant?

Image

alan-agius4:
This change introduces the ability to disable sourcemaps when running the development server.

A new Vite plugin has been added to remove sourcemap comments from generated assets. The dev server's configuration has been updated to conditionally apply this plugin and control CSS sourcemap generation based on the project's sourceMap setting.

Closes #31331

### DIFF
--- a/packages/angular/build/src/tools/vite/plugins/remove-sourcemaps.ts
+++ b/packages/angular/build/src/tools/vite/plugins/remove-sourcemaps.ts
@@ -1,0 +1,19 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import type { Plugin } from 'vite';
+
+export const removeSourceMapsPlugin: Plugin = {
+  name: 'vite:angular-remove-sourcemaps',
+  transform(code) {
+    return {
+      code,
+      map: { mappings: '' },
+    };
+  },
+};
--- a/packages/angular/build/src/builders/dev-server/tests/behavior/build-source-map_spec.ts
+++ b/packages/angular/build/src/builders/dev-server/tests/behavior/build-source-map_spec.ts
@@ -1,0 +1,49 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { executeDevServer } from '../../index';
+import { executeOnceAndFetch } from '../execute-fetch';
+import { describeServeBuilder } from '../jasmine-helpers';
+import { BASE_OPTIONS, DEV_SERVER_BUILDER_INFO } from '../setup';
+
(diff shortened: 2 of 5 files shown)


packages/angular/build/src/tools/vite/plugins/remove-sourcemaps.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import type { Plugin } from 'vite';
+
+export const removeSourceMapsPlugin: Plugin = {
```
AlonMiz, line 11: isn't "remove sourcemaps" wasteful? meaning the bundler already generated and the cpu/memory resources are wasted. WDYT?

PR #31511, opened 2025-10-17
refactor(@angular/ssr): simplify redirect URL determination

alan-agius4:
The logic to determine the final URL for server-side rendering redirects is simplified. Previously, it used LocationStrategy and UrlSerializer to construct the final URL. This is replaced by using PlatformLocation to directly get the pathname, search, and hash.

This change removes unnecessary complexity and dependencies, making the code easier to understand and maintain.

### DIFF
--- a/packages/angular/ssr/src/utils/ng.ts
+++ b/packages/angular/ssr/src/utils/ng.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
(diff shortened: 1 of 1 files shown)


alan-agius4:

- main: 663908564f2036e86ed82781c230959e3c6fe597
- 20.3.x: 8c6b3e70a8bd468779690c1d43c7a6304ec12239

PR #31539, opened 2025-10-21
refactor: replace `loadEsmModule` with dynamic `import()`

alan-agius4:
The custom `loadEsmModule` helper function was a workaround to prevent TypeScript from downleveling dynamic imports to `require()` calls, which would fail to load ESM modules.

With modern TypeScript and Node.js versions, this workaround is no longer necessary. This commit removes the `loadEsmModule` utility and replaces all its usages with standard dynamic `import()` expressions.

This simplifies the codebase and relies on the native module loading capabilities of the environment.

### DIFF
--- a/packages/angular/build/src/tools/angular/compilation/angular-compilation.ts
+++ b/packages/angular/build/src/tools/angular/compilation/angular-compilation.ts
@@ -9,7 +9,6 @@
 import type ng from '@angular/compiler-cli';
 import type { PartialMessage } from 'esbuild';
 import type ts from 'typescript';
-import { loadEsmModule } from '../../../utils/load-esm';
 import { convertTypeScriptDiagnostic } from '../../esbuild/angular/diagnostics';
 import { profileAsync, profileSync } from '../../esbuild/profiling';
 import type { AngularHostOptions } from '../angular-host';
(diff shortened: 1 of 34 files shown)


packages/angular/build/src/tools/angular/compilation/angular-compilation.ts
```diff
@@ -35,8 +34,7 @@ export abstract class AngularCompilation {
   static async loadCompilerCli(): Promise<typeof ng> {
     // This uses a wrapped dynamic import to load `@angular/compiler-cli` which is ESM.
     // Once TypeScript provides support for retaining dynamic imports this workaround can be dropped.
```
clydin: This comment can be removed as well

packages/angular/build/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
```diff
@@ -137,7 +129,7 @@ export async function load(url: string, context: { format?: string | null }, nex
   // need linking are ESM only.
   if (format === 'module' && isFileProtocol(url)) {
     const filePath = fileURLToPath(url);
-    let source = await javascriptTransformer.transformFile(filePath);
```
clydin, line 140: Is the transform not needed here anymore?

packages/angular/build/src/utils/load-proxy-config.ts
```diff
@@ -49,33 +47,8 @@ export async function loadProxyConfiguration(
...
-      break;
     default:
-      // The file could be either CommonJS or ESM.
-      // CommonJS is tried first then ESM if loading fails.
-      try {
-        proxyConfiguration = require(proxyPath);
-        break;
-      } catch (e) {
-        assertIsError(e);
-        if (e.code === 'ERR_REQUIRE_ESM' || e.code === 'ERR_REQUIRE_ASYNC_MODULE') {
```
clydin, line 69: We should still handle the async case here (top-level await in the config file or its dependencies). Unfortunately, we would still need the workaround for this one case.

packages/angular/build/src/utils/server-rendering/esm-in-memory-loader/loader-hooks.ts
alan-agius4: No

packages/angular/build/src/utils/load-proxy-config.ts
alan-agius4: Done.

PR #31545, opened 2025-10-21
fix(@angular-devkit/schematics): respect `--force` option when schematic contains `host.create`

### LINKED ISSUE #30578: bug(theme-color): Generator theme-color flags --force and --defaults do nothing
### Is this a regression?

_No response_

### Description

This is our theme-color generator command, running it twice results in "already exists" even with `--force`, also `--defaults` seem to have no effect here.

```
pnpm ng generate @angular/material:theme-color --force --defaults --directory="/libs/app-theme/vendor/material/" --primary-color="#0641e6" --tertiary-color="#994aff" --neutral-color="#313138" --error-color="#eb5757" --secondary-color="#009096" --neutral-variant-color="#b2b2b8"

Path "/libs/app-theme/vendor/material/_theme-colors.scss" already exist.
```

### Reproduction

StackBlitz link: https://stackblitz.com/edit/components-issue-starter-waftjbtm?file=package.json
Steps to reproduce:
1. `npm run issue`

Btw this starter is on NG18!

### Expected Behavior

Should respect all flags, even default to avoid providing all color flags.

### Actual Behavior

Some flags are ignored.

### Environment

- Angular: See repl
- CDK/Material:
- Browser(s):
- Operating System (e.g. Windows, macOS, Ubuntu):

alan-agius4:
Removes the `FileAlreadyExistException` check within the `create` method of `HostTree`. This change allows a schematic to create a file even if one already exists at the same path, effectively overwriting it.

This provides more flexibility for schematic authors, particularly in scenarios where files need to be replaced or updated unconditionally. It is intended to be used with schematics that have a `force` or `overwrite` option.

Closes #30578

### DIFF
--- a/packages/angular_devkit/schematics/src/tree/host-tree.ts
+++ b/packages/angular_devkit/schematics/src/tree/host-tree.ts
@@ -20,7 +20,6 @@ import {
 import { ParseError, parse as jsoncParse, printParseErrorCode } from 'jsonc-parser';
 import {
   ContentHasMutatedException,
-  FileAlreadyExistException,
   FileDoesNotExistException,
   InvalidUpdateRecordException,
   MergeConflictException,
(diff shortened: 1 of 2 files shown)


clydin (approved):

alan-agius4:

- main: 18bf8e7b3b36b968d064299e5c557942c0ed7ec0
- 20.3.x: a31533cf492048f62a41b9c09e53779269ee172d

PR #31558, opened 2025-10-22
refactor(@angular/cli): use command default for unspecified target builders

clydin:
If a target definition from a project's configuration does not contain a `builder` field and the CLI command provides a default, the default will now be used for the command execution. This is in addition to the existing case where the command will use a synthetic target definition for commands that provide default builders.


jkrems (approved):
Is there a test case worth adding for this?

alan-agius4 (approved):

alan-agius4:

- main: 744ae4e81d0eb415088b7965688fcfea2760b9c8
- 21.0.x: e1e2cb94e66f3611fc65489c69416216fcc388da

PR #31603, opened 2025-10-27
fix(@angular/build): allow custom runner configuration file for unit-test

### LINKED ISSUE #31392: @angular/build:unit-test: Configuration option for the time watermark of the progress reporters similar to the coverageWatermarks option
### Command

test

### Description

It would be great to have an option in angular.json to set the time watermark of the progress reporters.

This option would behave similar to the new coverageWatermark option in angular.json that allows us to change the color coding for high, medium, and low coverage.

Currently the reporter time threshold is hard-coded to 300 ms with the Vitest unit-test. If a test runs slower than 300 ms, the test suite is highlighted in yellow and the test is shown and highlighted in yellow.

There is currently no option to configure that hard-coded value.

This could either be a new option of the unit-test configuration like

```
test": {
  "builder": "@angular/build:unit-test",
  "options": {
    ...
    "reporterWatermark ": 1000,
    ...
}
```

or an option within the "reporters" tuple.

This specific example would set the reporter watermark to 1000 ms and only tests that run slower than 1000 ms would be highlighted in yellow.

Internally, this option would set Vitest's [slowTestThreshold ](https://vitest.dev/guide/cli.html#slowtestthreshold) option.

_No response_

clydin:
Adds a new `runnerConfig` option to the `unit-test` system to provide more control over the test runner's configuration file. This option is runner-agnostic and enhances flexibility for both Vitest and Karma.

The option accepts a boolean or a string path:
- `true`: Automatically searches for a default config file (`karma.conf.js` or `vitest.config.ts`). If not found, the runner will use its internal default configuration.
- `false` (default): Disables the use of any external config file.
- `path/to/config`: Uses the specified configuration file.

For Vitest, the loaded configuration is deep-merged with the system's programmatic config. This allows users to add or override advanced options (like `test.coverage`) while preserving the essential in-memory integration. The system's settings take precedence in case of conflicts.

Please note that while the file is loaded, the Angular team does not provide direct support for its specific contents or any third-party plugins used within it.

Informational messages are now logged to indicate which configuration file is being used, improving transparency.

Closes #31392

### DIFF
--- a/packages/angular/build/src/builders/unit-test/schema.json
+++ b/packages/angular/build/src/builders/unit-test/schema.json
@@ -19,6 +19,11 @@
       "default": "vitest",
       "enum": ["karma", "vitest"]
     },
+    "runnerConfig": {
+      "type": ["boolean", "string"],
+      "description": "Specifies the configuration file for the selected test runner. If a string is provided, it will be used as the path to the configuration file. If `true`, the builder will search for a default configuration file (e.g., `vitest.config.ts` or `karma.conf.js`). If `false`, no external configuration file will be used.\\nFor Vitest, this enables advanced options and the use of custom plugins. Please note that while the file is loaded, the Angular team does not provide direct support for its specific contents or any third-party plugins used within it.",
+      "default": false
+    },
     "browsers": {
       "description": "Specifies the browsers to use for test execution. When not specified, tests are run in a Node.js environment using jsdom. For both Vitest and Karma, browser names ending with 'Headless' (e.g., 'ChromeHeadless') will enable headless mode.",
       "type": "array",
--- a/goldens/public-api/angular/build/index.api.md
+++ b/goldens/public-api/angular/build/index.api.md
@@ -235,6 +235,7 @@ export type UnitTestBuilderOptions = {
     providersFile?: string;
     reporters?: SchemaReporter[];
     runner?: Runner;
+    runnerConfig?: RunnerConfig;
     setupFiles?: string[];
     tsConfig?: string;
     ui?: boolean;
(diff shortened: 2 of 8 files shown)


yjaaidi:
Nice one @clydin, thanks!

Just one thing. 
While I think that the feature is crucial and I know that a standalone `vitest.config.ts` requires more effort and is not the current goal, I am afraid that such a config file, especially when using the conventional name (`vitest.config.ts`) will actually confuse the integration with tools such as the IDE plugins as it’s a partial configuration. 

To avoid user and tool confusion, what about naming it differently (e.g. `vitest-overrides.config.ts` or something)
Or what about exposing a `defineVitestConfig` function that strips the options that can't be overridden from the typing.
I can already imagine issues such as:
- I've changed this option in `vitest.config.ts` and when I run the tests using `vitest`, the options work but the tests involving Angular components do not work, and when I use the builder, the option seems to be ignored, but that other option is not ignored.

clydin:
This is opt-in and the name can be configured to an arbitrary value. If there is need to avoid it being picked up by other tools (some actually may want it picked up), it can be named something different.  Also of relevance, this is more of a "base" rather than an override.  Options specified on command execution take precedence which follows the behavior of Vitest.

packages/angular/build/src/builders/unit-test/schema.json
```diff
@@ -19,6 +19,11 @@
       "default": "vitest",
       "enum": ["karma", "vitest"]
     },
+    "runnerConfig": {
+      "type": ["boolean", "string"],
```
alan-agius4, line 23: Note: Not sure how to handle this, but having `type` both boolean or string will make `runnerConfig` config not possible to be used from the CLI. As options cannot be either `boolean` or `string`.

yjaaidi:
Agreed, something like a `base` suffix would be a better fit. 

I am actually concerned about the default name. Newcomers and AI-generated code will more likely pick `vitest.config.ts` and be surprised of the behavior.

clydin:
Initial goal here was consistency as the Angular CLI has used the existing `karma.conf.js` default file name.
However, we can evaluate a different default but it would require implementing manual file discovery that would match vitest's.

clydin:

- main: 20079ed73dc138db3f22aea20eeaa4b0f8b46c52
- 21.0.x: 542d528683cc0e51fd5a55ed6dbf43168f7a51a5

PR #31621, opened 2025-10-28
fix(@angular/build): introduce vitest-base.config for test configuration

clydin:
When using the Vitest runner, a standard `vitest.config.js` file can cause conflicts with other tools or IDE extensions that might assume it represents a complete, standalone configuration. However, the builder uses this file only as a *base* configuration, which is then merged with its own internal setup.

To avoid this ambiguity, the builder now searches for a `vitest-base.config.(js|ts|...etc)` file when the `runnerConfig` option is set to `true`. This makes the intent clear that the file provides a base configuration specifically for the Angular CLI builder.

The search order is as follows:
1. Project Root
2. Workspace Root

If no `vitest-base.config.*` file is found, the builder proceeds with its default in-memory configuration. Vitest's default behavior of searching for `vitest.config.*` is explicitly disabled in this mode to ensure predictable and consistent test execution.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/configuration.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/configuration.ts
@@ -1,0 +1,55 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+/**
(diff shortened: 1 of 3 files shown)


yjaaidi:
🙏 thank you Charles!

clydin:

- main: 0fe6db9481c72fe356b3fac75362eb94ad8dd458
- 21.0.x: bab5806c281fd4cdd63b7969e691d703ed1e7680

PR #31622, opened 2025-10-29
refactor(@angular/build): normalize unit test coverage options for consistency

clydin:
Previously, the unit test builder's `coverage` option was conditional. It was only defined if `coverage: true` was set, and it consolidated several top-level `coverage*` properties from the schema. This required downstream consumers, like the Karma and Vitest executors, to handle a potentially undefined object and led to complex and sometimes incorrect option merging, especially in the Vitest runner.

This commit refactors the option normalization logic to ensure the `coverage` object is *always* defined. A new `enabled` flag, controlled by the main `coverage` option, is now part of this object.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/options.ts
+++ b/packages/angular/build/src/builders/unit-test/options.ts
@@ -96,22 +96,21 @@ export async function normalizeOptions(
     exclude: options.exclude,
     filter,
     runnerName: runner ?? 'vitest',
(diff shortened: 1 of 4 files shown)


clydin:

- main: 4cb985c077ec25fee11a7d581611f8bd43a36dfa
- 21.0.x: b39d0c1054fbb88a1f2358648cf47612577a9fe7

PR #31633, opened 2025-10-29
fix(@angular/build): add `define` option to dev-server

### LINKED ISSUE #28905: ng serve does not support the "define" argument
### Command

serve

### Description

I want to pass environment variables into the application during serve and build. In Angular 19, this works for build: `ng build --define="GGCC_TARGET='hosted'"`, while this doesn't for serve: `ng serve --define="GGCC_TARGET='hosted'"`, with the following error:
```
Error: Unknown argument: define
```

The `--define` argument works for both serve and build commands.

The custom "@ngx-env/builder:application" builder works, but I'd prefer to keep the amount of extra libraries to a minimum. Given how that builder is merely a thin wrapper around "@angular/build:application", I don't see why this couldn't work as I expect out of the box.

clydin:
This change introduces a `define` option to the dev-server builder, allowing users to specify global constants that will be replaced at build time. This provides a convenient way to configure environment-specific variables directly from the dev-server configuration.

The dev-server `define` option is merged with any `define` options present in the underlying build target, with the dev-server's values taking precedence in case of conflicts.

To support testing of this feature, a new `modifyTarget` method has been added to the builder test harness. This helper simplifies the modification of a target's options within a test case. Comprehensive end-to-end tests have been added to verify the replacement, merging, and override behaviors.

Closes #28905

### DIFF
--- a/goldens/public-api/angular/build/index.api.md
+++ b/goldens/public-api/angular/build/index.api.md
@@ -112,6 +112,9 @@ export enum BuildOutputFileType {
 export type DevServerBuilderOptions = {
     allowedHosts?: AllowedHosts;
     buildTarget: string;
+    define?: {
+        [key: string]: string;
+    };
     headers?: {
         [key: string]: string;
     };


clydin:

- main: 42115349d460e7c0a1fdedf8896c0bd511c8c2c7
- 21.0.x: 530d9270e87786594dd8d1956b0806a28650db25

PR #31667, opened 2025-10-31
Add build/serve MCP tools

amishne:
New MCP experimental tools:

* "build" tool: A new MCP tool that allows for one-off, non-watched builds of Angular projects, similar to
      `ng build`.
* "serve" tools: A group of new MCP tools to manage the development server:
  * "start_devserver": Starts the Angular development server as a background process.
  * "stop_devserver": Stops a running development server.
  * "wait_for_devserver_build": Waits for a build to complete and reports the logs.

Notes:

* Implementation mostly follows the design doc for build/serve/test workflow.
* The currently-alive devservers are stored as part of the MCP context, and so are maintained across tool invocations.
* Build progress detection relies on hardcoding expected strings to match - see "dev-server.ts".
* Beyond the unit tests, also tested manually in Gemini CLI.
* A bunch of tools, included these new ones, just return basic unweighted structured content, so added a utility function in a "utils.ts" file for that functionality.

### DIFF
--- a/packages/angular/cli/src/commands/mcp/dev-server.ts
+++ b/packages/angular/cli/src/commands/mcp/dev-server.ts
@@ -1,0 +1,148 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { ChildProcess } from 'child_process';
+import { Host } from './host';
+
+// Log messages that we want to catch to identify the build status.
+
+const BUILD_SUCCEEDED_MESSAGE = 'Application bundle generation complete.';
+const BUILD_FAILED_MESSAGE = 'Application bundle generation failed.';
+const WAITING_FOR_CHANGES_MESSAGE = 'Watch mode enabled. Watching for file changes...';
+const CHANGES_DETECTED_START_MESSAGE = '❯ Changes detected. Rebuilding...';
+const CHANGES_DETECTED_SUCCESS_MESSAGE = '✔ Changes detected. Rebuilding...';
+
+const BUILD_START_MESSAGES = [CHANGES_DETECTED_START_MESSAGE];
+const BUILD_END_MESSAGES = [
+  BUILD_SUCCEEDED_MESSAGE,
+  BUILD_FAILED_MESSAGE,
+  WAITING_FOR_CHANGES_MESSAGE,
+  CHANGES_DETECTED_SUCCESS_MESSAGE,
+];
+
+export type BuildStatus = 'success' | 'failure' | 'unknown';
+
+/**
+ * An Angular development server managed by the MCP server.
+ */
+export interface DevServer {
+  /**
+   * Launches the dev server and returns immediately.
+   *
+   * Throws if this server is already running.
+   */
+  start(): void;
+
+  /**
+   * If the dev server is running, stops it.
+   */
+  stop(): void;
+
+  /**
+   * Gets all the server logs so far (stdout + stderr).
+   */
+  getServerLogs(): string[];
+
+  /**
+   * Gets all the server logs from the latest build.
+   */
+  getMostRecentBuild(): { status: BuildStatus; logs: string[] };
+
+  /**
+   * Whether the dev server is currently being built, or is awaiting further changes.
+   */
+  isBuilding(): boolean;
+
+  /**
+   * `ng serve` port to use.
+   */
+  port: number;
+}
+
+export function devServerKey(project?: string) {
+  return project ?? '<default>';
+}
+
+/**
+ * A local Angular development server managed by the MCP server.
+ */
+export class LocalDevServer implements DevServer {
+  readonly host: Host;
+  readonly port: number;
+  readonly project?: string;
+
+  private devServerProcess: ChildProcess | null = null;
+  private serverLogs: string[] = [];
+  private buildInProgress = false;
+  private latestBuildLogStartIndex?: number = undefined;
+  private latestBuildStatus: BuildStatus = 'unknown';
+
+  constructor({ host, port, project }: { host: Host; port: number; project?: string }) {
+    this.host = host;
+    this.project = project;
+    this.port = port;
+  }
+
+  start() {
+    if (this.devServerProcess) {
+      throw Error('Dev server already started.');
+    }
+
+    const args = ['serve'];
+    if (this.project) {
+      args.push(this.project);
+    }
+
+    args.push(`--port=${this.port}`);
+
+    this.devServerProcess = this.host.spawn('ng', args, { stdio: 'pipe' });
+    this.devServerProcess.stdout?.on('data', (data) => {
+      this.addLog(data.toString());
+    });
+    this.devServerProcess.stderr?.on('data', (data) => {
+      this.addLog(data.toString());
+    });
+    this.devServerProcess.stderr?.on('close', () => {
+      this.stop();
+    });
+    this.buildInProgress = true;
+  }
+
+  private addLog(log: string) {
+    this.serverLogs.push(log);
+
+    if (BUILD_START_MESSAGES.some((message) => log.startsWith(message))) {
+      this.buildInProgress = true;
+      this.latestBuildLogStartIndex = this.serverLogs.length - 1;
+    } else if (BUILD_END_MESSAGES.some((message) => log.startsWith(message))) {
+      this.buildInProgress = false;
+      // We consider everything except a specific failure message to be a success.
+      this.latestBuildStatus = log.startsWith(BUILD_FAILED_MESSAGE) ? 'failure' : 'success';
+    }
+  }
+
+  stop() {
+    this.devServerProcess?.kill();
+    this.devServerProcess = null;
+  }
+
+  getServerLogs(): string[] {
+    return [...this.serverLogs];
+  }
+
+  getMostRecentBuild() {
+    return {
+      status: this.latestBuildStatus,
+      logs: this.serverLogs.slice(this.latestBuildLogStartIndex),
+    };
+  }
+
+  isBuilding() {
+    return this.buildInProgress;
+  }
+}
--- a/packages/angular/cli/src/commands/mcp/mcp-server.ts
+++ b/packages/angular/cli/src/commands/mcp/mcp-server.ts
@@ -10,15 +10,25 @@ import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import path from 'node:path';
 import type { AngularWorkspace } from '../../utilities/config';
 import { VERSION } from '../../utilities/version';
+import { DevServer } from './dev-server';
 import { registerInstructionsResource } from './resources/instructions';
 import { AI_TUTOR_TOOL } from './tools/ai-tutor';
 import { BEST_PRACTICES_TOOL } from './tools/best-practices';
+import { BUILD_TOOL } from './tools/build';
+import { START_DEVSERVER_TOOL } from './tools/devserver/start-devserver';
+import { STOP_DEVSERVER_TOOL } from './tools/devserver/stop-devserver';
+import { WAIT_FOR_DEVSERVER_BUILD_TOOL } from './tools/devserver/wait-for-devserver-build';
 import { DOC_SEARCH_TOOL } from './tools/doc-search';
 import { FIND_EXAMPLE_TOOL } from './tools/examples';
 import { MODERNIZE_TOOL } from './tools/modernize';
 import { ZONELESS_MIGRATION_TOOL } from './tools/onpush-zoneless-migration/zoneless-migration';
 import { LIST_PROJECTS_TOOL } from './tools/projects';
 import { AnyMcpToolDeclaration, registerTools } from './tools/tool-registry';
+
+/**
+ * Tools to manage devservers. Should be bundled together, then added to experimental or stable as a group.
+ */
+const SERVE_TOOLS = [START_DEVSERVER_TOOL, STOP_DEVSERVER_TOOL, WAIT_FOR_DEVSERVER_BUILD_TOOL];
 
 /**
  * The set of tools that are enabled by default for the MCP server.
@@ -37,7 +47,7 @@ const STABLE_TOOLS = [
  * The set of tools that are available but not enabled by default.
  * These tools are considered experimental and may have limitations.
  */
-export const EXPERIMENTAL_TOOLS = [MODERNIZE_TOOL] as const;
+export const EXPERIMENTAL_TOOLS = [BUILD_TOOL, MODERNIZE_TOOL, ...SERVE_TOOLS] as const;
 
 export async function createMcpServer(
   options: {
@@ -104,6 +114,7 @@ equivalent actions.
       workspace: options.workspace,
       logger,
       exampleDatabasePath: path.join(__dirname, '../../../lib/code-examples.db'),
+      devServers: new Map<string, DevServer>(),
     },
     toolDeclarations,
   );
--- a/packages/angular/cli/src/commands/mcp/tools/build.ts
+++ b/packages/angular/cli/src/commands/mcp/tools/build.ts
@@ -1,0 +1,111 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { z } from 'zod';
+import { CommandError, Host, LocalWorkspaceHost } from '../host';
+import { createStructuredContentOutput } from '../utils';
+import { McpToolDeclaration, declareTool } from './tool-registry';
+
+const DEFAULT_CONFIGURATION = 'development';
+
+const buildStatusSchema = z.enum(['success', 'failure']);
+type BuildStatus = z.infer<typeof buildStatusSchema>;
+
+const buildToolInputSchema = z.object({
+  project: z
+    .string()
+    .optional()
+    .describe(
+      'Which project to build in a monorepo context. If not provided, builds the default project.',
+    ),
+  configuration: z
+    .string()
+    .optional()
+    .describe('Which build configuration to use. Defaults to "development".'),
+});
+
+export type BuildToolInput = z.infer<typeof buildToolInputSchema>;
+
+const buildToolOutputSchema = z.object({
+  status: buildStatusSchema.describe('Build status.'),
+  logs: z.array(z.string()).optional().describe('Output logs from `ng build`.'),
+  path: z.string().optional().describe('The output location for the build, if successful.'),
+});
+
+export type BuildToolOutput = z.infer<typeof buildToolOutputSchema>;
+
+export async function runBuild(input: BuildToolInput, host: Host) {
+  // Build "ng"'s command line.
+  const args = ['build'];
+  if (input.project) {
+    args.push(input.project);
+  }
+  args.push('-c', input.configuration ?? DEFAULT_CONFIGURATION);
+
+  let status: BuildStatus = 'success';
+  let logs: string[] = [];
+  let outputPath: string | undefined;
+
+  try {
+    logs = (await host.runCommand('ng', args)).logs;
+  } catch (e) {
+    status = 'failure';
+    if (e instanceof CommandError) {
+      logs = e.logs;
+    } else if (e instanceof Error) {
+      logs = [e.message];
+    } else {
+      logs = [String(e)];
+    }
+  }
+
+  for (const line of logs) {
+    const match = line.match(/Output location: (.*)/);
+    if (match) {
+      outputPath = match[1].trim();
+      break;
+    }
+  }
+
+  const structuredContent: BuildToolOutput = {
+    status,
+    logs,
+    path: outputPath,
+  };
+
+  return createStructuredContentOutput(structuredContent);
+}
+
+export const BUILD_TOOL: McpToolDeclaration<
+  typeof buildToolInputSchema.shape,
+  typeof buildToolOutputSchema.shape
+> = declareTool({
+  name: 'build',
+  title: 'Build Tool',
+  description: `
+<Purpose>
+Perform a one-off, non-watched build using "ng build". Use this tool whenever the user wants to build an Angular project; this is similar to
+"ng build", but the tool is smarter about using the right configuration and collecting the output logs.
+</Purpose>
+<Use Cases>
+* Building an Angular project and getting build logs back.
+</Use Cases>
+<Operational Notes>
+* This tool runs "ng build" so it expects to run within an Angular workspace.
+* If you want a watched build which updates as files are changed, use "start_devserver" instead, which also serves the app.
+* You can provide a project instead of building the root one. The "list_projects" MCP tool could be used to obtain the list of projects.
+* This tool defaults to a development environment while a regular "ng build" defaults to a production environment. An unexpected build
+  failure might suggest the project is not configured for the requested environment.
+</Operational Notes>
+`,
+  isReadOnly: false,
+  isLocalOnly: true,
+  inputSchema: buildToolInputSchema.shape,
+  outputSchema: buildToolOutputSchema.shape,
+  factory: () => (input) => runBuild(input, LocalWorkspaceHost),
+});
--- a/packages/angular/cli/src/commands/mcp/tools/build_spec.ts
+++ b/packages/angular/cli/src/commands/mcp/tools/build_spec.ts
@@ -1,0 +1,99 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { CommandError, Host } from '../host';
+import { MockHost } from '../testing/mock-host';
+import { runBuild } from './build';
+
+describe('Build Tool', () => {
+  let mockHost: MockHost;
+
+  beforeEach(() => {
+    mockHost = {
+      runCommand: jasmine.createSpy<Host['runCommand']>('runCommand').and.resolveTo({ logs: [] }),
+      stat: jasmine.createSpy<Host['stat']>('stat'),
+      existsSync: jasmine.createSpy<Host['existsSync']>('existsSync'),
+    } as Partial<MockHost> as MockHost;
+  });
+
+  it('should construct the command correctly with default configuration', async () => {
+    await runBuild({}, mockHost);
+    expect(mockHost.runCommand).toHaveBeenCalledWith('ng', ['build', '-c', 'development']);
+  });
+
+  it('should construct the command correctly with a specified project', async () => {
+    await runBuild({ project: 'another-app' }, mockHost);
+    expect(mockHost.runCommand).toHaveBeenCalledWith('ng', [
+      'build',
+      'another-app',
+      '-c',
+      'development',
+    ]);
+  });
+
+  it('should construct the command correctly for a custom configuration', async () => {
+    await runBuild({ configuration: 'myconfig' }, mockHost);
+    expect(mockHost.runCommand).toHaveBeenCalledWith('ng', ['build', '-c', 'myconfig']);
+  });
+
+  it('should handle a successful build and extract the output path and logs', async () => {
+    const buildLogs = [
+      'Build successful!',
+      'Some other log lines...',
+      'some warning',
+      'Output location: dist/my-app',
+    ];
+    mockHost.runCommand.and.resolveTo({
+      logs: buildLogs,
+    });
+
+    const { structuredContent } = await runBuild({ project: 'my-app' }, mockHost);
+
+    expect(mockHost.runCommand).toHaveBeenCalledWith('ng', [
+      'build',
+      'my-app',
+      '-c',
+      'development',
+    ]);
+    expect(structuredContent.status).toBe('success');
+    expect(structuredContent.logs).toEqual(buildLogs);
+    expect(structuredContent.path).toBe('dist/my-app');
+  });
+
+  it('should handle a failed build and capture logs', async () => {
+    const buildLogs = ['Some output before the crash.', 'Error: Something went wrong!'];
+    const error = new CommandError('Build failed', buildLogs, 1);
+    mockHost.runCommand.and.rejectWith(error);
+
+    const { structuredContent } = await runBuild(
+      { project: 'my-failed-app', configuration: 'production' },
+      mockHost,
+    );
+
+    expect(mockHost.runCommand).toHaveBeenCalledWith('ng', [
+      'build',
+      'my-failed-app',
+      '-c',
+      'production',
+    ]);
+    expect(structuredContent.status).toBe('failure');
+    expect(structuredContent.logs).toEqual(buildLogs);
+    expect(structuredContent.path).toBeUndefined();
+  });
+
+  it('should handle builds where the output path is not found in logs', async () => {
+    const buildLogs = ["Some logs that don't match any output path."];
+    mockHost.runCommand.and.resolveTo({ logs: buildLogs });
+
+    const { structuredContent } = await runBuild({}, mockHost);
+
+    expect(structuredContent.status).toBe('success');
+    expect(structuredContent.logs).toEqual(buildLogs);
+    expect(structuredContent.path).toBeUndefined();
+  });
+});
--- a/packages/angular/cli/src/commands/mcp/tools/modernize_spec.ts
+++ b/packages/angular/cli/src/commands/mcp/tools/modernize_spec.ts
@@ -10,12 +10,13 @@ import { Stats } from 'fs';
 import { mkdir, mkdtemp, rm, writeFile } from 'fs/promises';
 import { tmpdir } from 'os';
 import { join } from 'path';
-import * as host from '../host';
+import { CommandError } from '../host';
+import { MockHost } from '../testing/mock-host';
 import { ModernizeOutput, runModernization } from './modernize';
 
 describe('Modernize Tool', () => {
   let projectDir: string;
-  let mockHost: host.Host;
+  let mockHost: MockHost;
 
   beforeEach(async () => {
     // Create a temporary directory and a fake angular.json to satisfy the tool's project root search.
@@ -28,7 +29,7 @@ describe('Modernize Tool', () => {
       existsSync: jasmine.createSpy('existsSync').and.callFake((p: string) => {
         return p === join(projectDir, 'angular.json');
       }),
-    };
+    } as Partial<MockHost> as MockHost;
   });
 
   afterEach(async () => {
@@ -105,7 +106,7 @@ describe('Modernize Tool', () => {
       ['generate', '@angular/core:self-closing-tag', '--path', '.'],
       { cwd: projectDir },
     );
-    expect(structuredContent?.stderr).toBeUndefined();
+    expect(structuredContent?.logs).toBeUndefined();
     expect(structuredContent?.instructions).toEqual(
       jasmine.arrayWithExactContents([
         'Migration control-flow on directory . completed successfully.',
@@ -149,7 +150,7 @@ describe('Modernize Tool', () => {
       ['generate', '@angular/core:self-closing-tag', '--path', 'subfolder2'],
       { cwd: projectDir },
     );
-    expect(structuredContent?.stderr).toBeUndefined();
+    expect(structuredContent?.logs).toBeUndefined();
     expect(structuredContent?.instructions).toEqual(
       jasmine.arrayWithExactContents([
         'Migration control-flow on directory subfolder1 completed successfully.',
@@ -161,7 +162,7 @@ describe('Modernize Tool', () => {
   });
 
   it('should return an error if angular.json is not found', async () => {
-    (mockHost.existsSync as jasmine.Spy).and.returnValue(false);
+    mockHost.existsSync.and.returnValue(false);
 
     const { structuredContent } = (await runModernization(
       {
@@ -179,8 +180,8 @@ describe('Modernize Tool', () => {
 
   it('should report errors from transformations', async () => {
     // Simulate a failed execution
-    (mockHost.runCommand as jasmine.Spy).and.rejectWith(
-      new host.CommandError('Command failed with error', 'stdout', 'stderr', 1),
+    mockHost.runCommand.and.rejectWith(
+      new CommandError('Command failed with error', ['some logs'], 1),
     );
 
     const { structuredContent } = (await runModernization(
@@ -196,9 +197,7 @@ describe('Modernize Tool', () => {
       ['generate', '@angular/core:self-closing-tag', '--path', '.'],
       { cwd: projectDir },
     );
-    expect(structuredContent?.stdout).toContain('stdout');
-    expect(structuredContent?.stderr).toContain('stderr');
-    expect(structuredContent?.stderr).toContain('Command failed with error');
+    expect(structuredContent?.logs).toEqual(['some logs', 'Command failed with error']);
     expect(structuredContent?.instructions).toEqual([
       'Migration self-closing-tag on directory . failed.',
     ]);
--- a/packages/angular/cli/src/commands/mcp/utils.ts
+++ b/packages/angular/cli/src/commands/mcp/utils.ts
@@ -1,0 +1,24 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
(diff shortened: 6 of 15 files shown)


packages/angular/cli/src/commands/mcp/mcp-server.ts
```diff
@@ -27,6 +28,7 @@ import { AnyMcpToolDeclaration, registerTools } from './tools/tool-registry';
 const STABLE_TOOLS = [
   AI_TUTOR_TOOL,
   BEST_PRACTICES_TOOL,
+  BUILD_TOOL,
```
dgp1130: Question: Are we sure we want to start as stable? Is it worth going through some amount of experimental first?

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+  project: z
+    .string()
+    .optional()
+    .describe(
+      'Which project to build in a monorepo context. If not provided, builds the top-level project.',
+    ),
+  configuration: z
+    .enum(Object.keys(CONFIGURATIONS) as [string, ...string[]])
+    .optional()
+    .describe('Which build configuration to use. Defaults to "development".'),
```
dgp1130, line 29: Consider: `development` and `production` are the two default generated configurations, but users can in theory add others. Would it be more flexible (with minimal effort) to just make this an optional string value which defaults to `development` and then unconditionally pass it into the `-c` argument? This way it would work with any configuration.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+const buildToolOutputSchema = z.object({
+  status: z.enum(BUILD_STATUSES).describe('Build status.'),
+  logs: z.string().describe('Output logs from `ng build`.'),
+  path: z.string().optional().describe('The output path the build project was written into.'),
+});
+
+export type BuildToolOutput = z.infer<typeof buildToolOutputSchema>;
+
+export function runBuild(input: BuildToolInput) {
+  const configurationName = input.configuration ?? 'development';
```
dgp1130: Consider: Do we need to validate `input.configuration` as `'development' | 'production'`? I guess this problem goes away if we allow any arbitrary string.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+  if (configuration.args) {
+    command.push(configuration.args);
+  }
+
+  let status: BuildStatus = 'success';
+  let logs = '';
+  let outputPath: string | undefined;
+
+  try {
+    logs = execSync(command.join(' ')).toString();
```
dgp1130: Consider: Should we use the async promise version so as not to hang other queued tasks? Could the same MCP server receive multiple tool calls at the same time? Can multiple agents/prompts connect to the same MCP server?

Even if we can assume one blocking tool call at a time, we might have some background async work to do, so maybe it's worth being async anyways?

Also, I'll mention since it's not very intuitive: The way to do async with `child_process` is to feed it through `promisify` like this: https://github.com/dgp1130/rules_prerender/blob/d6a53dcfd96229bcf134b7b8c3d2c0e2921d6c1a/common/testing/binary.mts

dgp1130: Consider: I typically recommend `execFile` over `exec` (whether the sync or async variant) because it removes shell escaping. Using `exec` you need to properly quote arguments and run the risk of a malicious actor injecting some unexpected environment variables or something like that. `execFile` is much safer in that regard.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+  } catch (e) {
+    status = 'failure';
+    if (e instanceof Error) {
+      logs = e.message;
+      if ('stdout' in e) {
+        logs += `\nSTDOUT:\n${(e as { stdout: Buffer }).stdout.toString()}`;
+      }
+      if ('stderr' in e) {
+        logs += `\nSTDERR:\n${(e as { stderr: Buffer }).stderr.toString()}`;
+      }
```
dgp1130: Observation: This separates the `stdout` / `stderr` logs which can lose a bit of context. I think that's probably ok given that the result is going to an AI anyways and I think it's unlikely for a message in one channel to depend on having previously printing something to the other, but just wanted to note this explicitly.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  const structuredContent: BuildToolOutput = {
+    status,
+    logs,
+    path: outputPath,
+  };
+
+  return {
+    content: [{ type: 'text' as const, text: JSON.stringify(structuredContent) }],
```
dgp1130: Consider: Should we make this JSON "pretty" with `JSON.stringify(structuredContent, null, 2)`?

The AI probably doesn't care, and maybe the extra tokens aren't worth it, but a human inspecting the output might be helped by it. If not, then maybe we should consider some kind of "debug" mode in the future where we can enable stuff like this.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+<Use Cases>
+* Building the Angular project and getting build logs back.
+</Use Cases>
+<Operational Notes>
+* This tool runs "ng build" so it expects to run within an Angular workspace.
+* You can provide a project instead of building the root one. The
+  "list_projects" MCP tool could be used to obtain the list of projects.
+</Operational Notes>
+`,
+  isReadOnly: false,
```
dgp1130, line 106: Question: Just for my benefit, does "read only" in this context mean "does not write to disk"?

Builds don't modify source code, though I can see that writing the build output to `dist/` is altering the file system. I wonder if we could support read only mode by not writing the output files and instead just returning error logs?

Probably not very important for this PR, just wondering if that's a potential improvement we can consider.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+const CONFIGURATIONS = {
+  development: {
+    args: '-c development',
+  },
+  production: {
+    args: '',
+  },
+};
+
+const BUILD_STATUSES = ['success', 'failure'] as const;
```
dgp1130: Consider: Should this be an `enum` so we can reference it like `BuildStatus.Success` rather than using string literals later on?

packages/angular/cli/src/commands/mcp/tools/build_spec.ts
```diff
@@ -0,0 +1,85 @@
...
+});
+
+describe('Build Tool', () => {
+  beforeEach(() => {
+    mockedExecSync.calls.reset();
+  });
+
+  it('should handle a successful build and extract the output path', () => {
+    const buildLogs =
+      'Build successful!\nSome other log lines...\nOutput location: dist/my-cool-app';
```
dgp1130: Consider: I'm a little worried that changes to `ng build` logging could unintentionally break this. Ideally we would have at least one test which _actually_ runs `ng build` to verify.

However there's maybe some simpler solution of just finding wherever `Output location:` is in source code and adding a comment that this is a load-bearing detail for the MCP server. Hopefully if we ever changed that we would at least notice the comment.

Alternatively, I wonder if it's worth tweaking the build output to make this easier. Could we (maybe gated on a private environment variable) adjust `ng build` output with something more explicit like:

```
BUILD COMPLETE: {
  "outputLocation": "dist/my-cool-app"
}
```

And then parse it in a slightly more structured fashion?

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+  let status: BuildStatus = 'success';
+  let logs = '';
+  let outputPath: string | undefined;
+
+  try {
+    logs = execSync(command.join(' ')).toString();
+    const match = logs.match(/Output location: (.*)/);
+    if (match) {
+      outputPath = match[1].trim();
+    }
```
dgp1130: Consider: Is there anything we can do on the else case here? If the build succeeds but we don't find the output location, then that feels like an internal bug in Angular CLI. Should we print or log somewhere that this happened? Is it worth throwing or crashing intentionally just to make this painful enough for us or users to notice and file a bug?

/cc @clydin in case he has any thoughts.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+  } catch (e) {
+    status = 'failure';
+    if (e instanceof Error) {
+      logs = e.message;
+      if ('stdout' in e) {
+        logs += `\nSTDOUT:\n${(e as { stdout: Buffer }).stdout.toString()}`;
```
dgp1130: Consider: Should we still check for `outputLocation` here? Most likely the build failed before writing to disk, but it at least hypothetically could have failed after that. Maybe it's unlikely enough for us to not worry about, but I just want to point this out directly.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -0,0 +1,118 @@
...
+
+const BUILD_STATUSES = ['success', 'failure'] as const;
+type BuildStatus = (typeof BUILD_STATUSES)[number];
+
+const buildToolInputSchema = z.object({
+  project: z
+    .string()
+    .optional()
+    .describe(
+      'Which project to build in a monorepo context. If not provided, builds the top-level project.',
```
dgp1130: Nit: I'm not sure we've ever used the term `top-level` in this context before. I think it would just be the "default" project? @clydin, how do we refer to this elsewhere?

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        devServerProcess.kill('SIGTERM');
+        devServerProcess = null;
+
+        return {
+          structuredContent: {
+            message: 'Development server stopped.',
+            logs: serverLogs,
+          },
+        };
```
dgp1130: Question: Why does this return `structuredContent` but the `build` tool returns `text`? Is there a meaningful difference between these two?

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+      case 'start_devserver':
+        if (devServerProcess) {
+          return {
+            structuredContent: {
+              message: 'Development server is already running.',
+            },
+          };
+        }
+
+        const args = ['serve'];
```
dgp1130: Consider: Should we spawn with `--port 0`? And also emit the port to the output?

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+          structuredContent: {
+            message: 'Development server started.',
+          },
+        };
+
+      case 'stop_devserver':
+        if (!devServerProcess) {
+          return {
+            structuredContent: {
+              message: 'Development server is not running.',
```
dgp1130: Nit: I think "Development server *was* not running." might be a little clearer that this was effectively a no-op.

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+            },
+          };
+        }
+
+        devServerProcess.kill('SIGTERM');
+        devServerProcess = null;
```
dgp1130: Consider: Should we wait for the process to die? I don't think there's a strict timeout for how long it takes a process to respond to a `SIGTERM` signal, so it might still be holding its port and serving traffic for an unknown length of time. I'd be curious to know what Angular CLI actually _does_ under the hood in response to a `SIGTERM` signal.

Related: If we do wait, is there a timeout after which we assume the process is hung and should pass `SIGKILL`?

We can maybe defer this until it becomes a meaningful problem, just more thinking out loud about what the proper semantics are for killing a process like this.

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+            logs: serverLogs,
+          },
+        };
+    }
+  };
+}
+
+export const SERVE_TOOL = declareTool({
+  name: 'serve',
+  title: 'Serve Tool',
```
dgp1130: Question: I'm curious about the motivation to use a single tool with an enum parameter rather than two distinct tools? Do we see a benefit to writing it this way?

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { ChildProcess, spawn } from 'child_process';
+import { z } from 'zod';
+import { McpToolContext, declareTool } from './tool-registry';
+
+let devServerProcess: ChildProcess | null = null;
```
dgp1130: Suggestion: I think we want to track this as a `Map<string /* project name */, ChildProcess>` so we can work with multiple projects at the same time.

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+* **Starting the Server:** Use the 'start_devserver' command to begin serving the application. The tool will return immediately while the server runs in the background.
+* **Stopping the Server:** Use the 'stop_devserver' command to terminate the running development server and retrieve the logs.
+</Use Cases>
+<Operational Notes>
+* This tool manages a single, shared development server instance.
+* 'start_devserver' is asynchronous. Subsequent commands can be run while the server is active.
+* 'stop_devserver' should be called to gracefully shut down the server and access the full log output.
+</Operational Notes>
+`,
+  isReadOnly: false,
```
dgp1130: Question: Similar to `build`, I wonder if it's possible to make this work in a read only environment? Does `ng serve` emit content to the file system and does it _have_ to? Does opening a socket count as a write?

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+          serverLogs.push(data.toString());
+        });
+
+        devServerProcess.on('close', () => {
+          devServerProcess = null;
+        });
+
+        return {
+          structuredContent: {
+            message: 'Development server started.',
```
dgp1130: Consider: Just spawning the process isn't quite enough, we need to wait for it to be serving traffic given that startup might take some time (not sure if `ng serve` waits for the first build before binding to a port).

We probably need to wait for some kind of "development server started" output or send a "health check" request to the port if we know it in advance.

packages/angular/cli/src/commands/mcp/tools/build.ts
dgp1130, line 1: Consider: I had some thoughts on usage of `execSync` here, but it seems like a later commit pivots to `host` as the mechanism for running a subprocess. Would it make sense to just squash these two commits together? Is there value in having this intermediate state with `execSync`?

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -39,59 +39,61 @@ export type BuildToolInput = z.infer<typeof buildToolInputSchema>;
...
-      }
-      if ('stderr' in e) {
-        logs += `\nSTDERR:\n${(e as { stderr: Buffer }).stderr.toString()}`;
-      }
+    if (e instanceof CommandError) {
+      stdout = e.stdout;
+      stderr = e.stderr;
+    } else if (e instanceof Error) {
+      stderr = e.message;
     }
```
dgp1130, line 72: Consider: Is there anything we can do for the case where `e` is not an `Error`? Maybe just `stderr = String(e)` as a best effort approach?

packages/angular/cli/src/commands/mcp/tools/build_spec.ts
```diff
@@ -6,80 +6,89 @@
...
       'Build successful!\nSome other log lines...\nOutput location: dist/my-cool-app';
-    mockedExecSync.and.returnValue(Buffer.from(buildLogs));
-
-    const result = runBuild({ project: 'my-cool-app' });
-
-    expect(mockedExecSync).toHaveBeenCalledWith('ng build my-cool-app -c development');
-    expect(result.structuredContent.status).toBe('success');
-    expect(result.structuredContent.logs).toBe(buildLogs);
-    expect(result.structuredContent.path).toBe('dist/my-cool-app');
+    (mockHost.runCommand as jasmine.Spy).and.resolveTo({
```
dgp1130: Consider: It should be possible to type `mockHost` properties as `jasmine.Spy` themselves (if they're not already spies) and then you can drop these casts.

Typically I do something like: `jasmine.createSpy<typeof Host['prototype']['runCommand']>('runCommand')` to make sure the type matches.

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -0,0 +1,124 @@
...
+        devServerProcess.kill('SIGTERM');
+        devServerProcess = null;
+
+        return {
+          structuredContent: {
+            message: 'Development server stopped.',
+            logs: serverLogs,
+          },
+        };
+    }
```
dgp1130: Consider: Should we have a `default` case which emits an error?

packages/angular/cli/src/commands/mcp/tools/serve.ts
```diff
@@ -120,5 +124,7 @@ Manages the Angular development server ("ng serve"). It allows you to start and
   isLocalOnly: true,
   inputSchema: serveToolInputSchema.shape,
   outputSchema: serveToolOutputSchema.shape,
-  factory: serveToolFactory,
+  factory: () => (input) => {
+    return runServeTool(input, LocalWorkspaceHost, { devServerProcess: null, serverLogs: [] });
```
dgp1130: Question: How exactly is `context` preserved across invocations? I would expect this to be stored inside the factory closure, but that doesn't seem to be how it works. Is this an MCP concept which it is saving for us somehow?

packages/angular/cli/src/commands/mcp/tools/serve.ts
dgp1130, line 1: Consider: The history would likely be easier to follow if this commit came _before_ the serve tool. This way we have `Host` interface which supports `spawn` and can squash the `serve.ts` changes into that commit and immediately use it.

The way I typically architect larger PRs like this is to have the earlier commits create basic primitives which are unused (outside of tests) and then the later commits depend on and combine those primitives to solve the bigger problem I'm aiming for. This way to avoid intermediate states with minimal meaning to overall problem.

packages/angular/cli/src/commands/mcp/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+/**
+ * Returns simple structured content output from an MCP tool.
+ *
+ * Returns a structure with both `content` and `structuredContent` for maximum compatibility.
+ * @param structuredContent
+ * @returns
+ */
+function createStructureContentOutput<OutputType>(structuredContent: OutputType) {
```
dgp1130: Typo: Should this be **Structured** instead of **Structure**?

packages/angular/cli/src/commands/mcp/dev-server.ts
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+import { ChildProcess } from 'child_process';
+import { Host } from './host';
+
+const BUILD_SUCCEEDED_MESSAGE = 'Application bundle generation complete.';
+const BUILD_FAILED_MESSAGE = 'Application bundle generation failed.';
+const WAITING_FOR_CHANGES_MESSAGE = 'Watch mode enabled. Watching for file changes...';
+const CHANGES_DETECTED_START_MESSAGE = '❯ Changes detected. Rebuilding...';
+const CHANGES_DETECTED_SUCCESS_MESSAGE = '✔ Changes detected. Rebuilding...';
```
dgp1130, line 18: Consider: Similar to my point on `ng build`, would there be value in changing the output of `ng serve` (possibly behind an environment variable) to make it easier to consume? We probably don't want to tweak the general output, but maybe if we added something like:

```
MCP MESSAGE: {"status": "started"}
// Some other logs...
MCP MESSAGE: {"status": "finished", "result": "success"}

MCP MESSAGE: {"status": "started"}
// Some other logs...
MCP MESSAGE: {"status": "finished", "result": "error"}
```

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
```diff
@@ -0,0 +1,83 @@
...
+      message: `Development server for project '${projectKey}' is already running.`,
+    });
+  }
+
+  const devServer = new LocalDevServer({ host, project: input.project, port: input.port });
+  devServer.start();
+
+  if (!context.devServers) {
+    context.devServers = new Map<string, DevServer>();
+  }
```
dgp1130: Nit: Move this up first so the rest of the code can assume `context.devServers` exists.

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export type StartDevserverToolInput = z.infer<typeof startDevserverToolInputSchema>;
+
+const startDevserverToolOutputSchema = z.object({
+  message: z.string().describe('A message indicating the result of the operation.'),
+});
+
+export type StartDevserverToolOutput = z.infer<typeof startDevserverToolOutputSchema>;
+
+export const DEFAULT_PROJECT_KEY = '<default>';
```
dgp1130: Bug: I think the default project still has a name in the `angular.json`, and we might want to use that to avoid duplicating devservers. For example if I ask:

```
Start a devserver. (AI uses the default project with the "<default>" key)
Start a devserver for `my-foo`. (AI uses `my-foo` as the project name, which might be the same project as the default)
```

Ideally we would probably look up the actual name of the default project in the `angular.json` and use that as the key. We can potentially fix this in a future PR and is unlikely to be a major issue, but just something to consider.

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
```diff
@@ -0,0 +1,83 @@
...
+import { McpToolContext, McpToolDeclaration, declareTool } from './tool-registry';
+
+const startDevserverToolInputSchema = z.object({
+  project: z
+    .string()
+    .optional()
+    .describe(
+      'Which project to serve in a monorepo context. If not provided, serves the default project.',
+    ),
+  port: z.number().optional().describe('Port to listen on.'),
```
dgp1130: Nit: Document default behavior.

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
```diff
@@ -0,0 +1,83 @@
...
+});
+
+export type StartDevserverToolInput = z.infer<typeof startDevserverToolInputSchema>;
+
+const startDevserverToolOutputSchema = z.object({
+  message: z.string().describe('A message indicating the result of the operation.'),
```
dgp1130: Consider: Should we output the port as a separate option?

packages/angular/cli/src/commands/mcp/dev-server.ts
```diff
@@ -0,0 +1,135 @@
...
+      throw Error('Dev server already started.');
+    }
+
+    const args = ['serve'];
+    if (this.project) {
+      args.push(this.project);
+    }
+    if (this.port) {
+      args.push(`--port=${this.port}`);
+    }
```
dgp1130: Bug: Don't we need to default to `0` or else it will use `4200`?

packages/angular/cli/src/commands/mcp/tools/wait-for-devserver-build.ts
```diff
@@ -0,0 +1,103 @@
...
+  project: z
+    .string()
+    .optional()
+    .describe(
+      'Which project to wait for in a monorepo context. If not provided, waits for the default project server.',
+    ),
+  timeout: z
+    .number()
+    .optional()
+    .default(30000)
```
dgp1130: Question: @clydin, I'm curious to know what you think a good timeout would be here. I wonder if 30s might be too short such that some larger projects will hit this 100% of the time and the AI might not know to configure it.

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
dgp1130, line 1: Consider: Would there be value in grouping the `start-devserver.ts`, `stop-devserver.ts`, and `wait-for-devserver-build.ts` files into a single `devserver/` subdirectory? I mention this because they share important state in the devserver map in context, which could be tricky to follow. Colocating that logic as much as possible would be helpful for future debugging.

packages/angular/cli/src/commands/mcp/mcp-server.ts
amishne: It's experimental in the final code. Sorry, still new at this GitHub thing, I intended the change to be reviewed as a single change because I had some layering here, I probably should have squashed it.

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Done

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Changed it to unify all logs as a list of strings. Made it consistent across `host`, `modernize` and `build`.

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Obsolete (they now all go through the same function - and it does make it pretty)

packages/angular/cli/src/commands/mcp/tools/build_spec.ts
amishne: But then I need to actually type a `MockHost`, right? Added it under a separate `testing` folder, take a look.

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: The context is created once when we register our tools (`createMcpServer` in `mcp-servers.ts`). It's kept alive as long as the MCP server is.

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Discussed offline, this whole thing should have been squashed.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/dev-server.ts
amishne: Discussed there as well. I like the environment variable approach, but I prefer to do that in a follow-up

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
amishne: Good point, I'll follow up. What's the standard procedure to ensure it's not lost, TODO in code, or open a bug?

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
amishne: We can wait until we have a use case

packages/angular/cli/src/commands/mcp/dev-server.ts
amishne: Port isn't optional, so removed the `if`

packages/angular/cli/src/commands/mcp/tools/wait-for-devserver-build.ts
amishne: Waiting for feedback from @clydin but changing to 3 minutes for now.

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
amishne: Done

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Done

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Changed to "default".

packages/angular/cli/src/commands/mcp/tools/build_spec.ts
amishne: There's apparently a "json logging" mode that returns a structured object with location data: https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/tools/esbuild/utils.ts#L414 but looking at its content, with things such as "color" suggests it's used for a specific purpose.

I like the environment variable approach. We could use it for `serve` as well, to emit specific strings that won't easily break.

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Might as well. Moved it outside the try block.

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Done

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: We're currently just using [process.kill](https://nodejs.org/api/child_process.html#subprocesskillsignal). By default it sends `SIGTERM`, is there another signal we should be sending instead?

In any case, I'm not sure what benefits we gain if we wait for it to die. If it fails to die for some reason it's indeed a problem but it could also make the agent hang so I'm not sure waiting is an improvement.

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Obsolete

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Same comment as there

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: To clarify - you're saying "start" should also include the first "wait" call built-in? Right now the tool's documentation encourages the agent to call the "wait" tool after the "start" tool, but it's indeed not guaranteed.

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Squashed

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: I think both views are valid, but I do lean on saying that these should indeed be "read-only" especially since our documentation says "tools that do not make changes to the project". Thoughts @clydin ? It would mean changing both build and serve to be read-only, even though they write to the file system, because they don't modify the project.

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Using vanilla enum is tricky because of how it's consumed by zod. But I've changed it to be consistent with other tools.

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: It's mostly here as a graceful error message for a case I don't expect to occur. Not sure about what th ebest behavior here, but I was hesitant to entirely fail over something that's not technically blocking. e.g. if we *don't* make the location detection better, and the string breaks, then at least this flow still works. But yes it means an end-to-end test would be useful to catch this.

packages/angular/cli/src/commands/mcp/tools/serve.ts
dgp1130: AIUI, `SIGTERM` is politely asking a process to clean itself up and terminate gracefully, it receives an event which it should respond to by eventually calling `process.exit()`. Meanwhile `SIGKILL` instructs the operating system to forcibly shut down the process whether that process likes it or not.

I think `SIGTERM` is fine, it's more that the process isn't guaranteed to die if it's misbehaving (whether intentionally or not). I think my main concern is that an agent might call `stop_devserver`, and receive a successful result, but then the devserver might still be running and could lead to confusion for the agent.

The risk is probably low and we don't need to overcomplicate things at this stage, I'm fine ignoring the issue until we have more signal of where it can actually be problematic, more just pointing out that nothing here right now currently guarantees that the devserver is actually dead when `stop_devserver` completes.

packages/angular/cli/src/commands/mcp/tools/serve.ts
dgp1130: Hmm, this lifecycle is a little tricky. The critical postcondition I would expect is that when `start_devserver` completes, a subsequent HTTP request to `localhost:${port}` should respond (even if that response is "build in progress").

If we only wait for the _process_ to spawn, the server might not have bound the port yet or be ready to serve traffic. We can wait for a build, though if the  build fails we'd still want the devserver to start (not sure what `ng serve` behavior is in this respect) and we would likely want to propagate build results to the AI agents.

I guess the question here is whether it is possible and worth separating the concepts of "starting a devserver" and "running the first build". If we can, I suspect separating them would be useful. This would mean the `start_devserver` command always starts a devserver whether or not a build succeeds, and its the agent's responsibility to separate wait for a build / check results. If we can't separate these two use cases (whether because Vite or `ng serve` don't work that way), then the simplest answer is likely to have `start_devserver` also implicitly `wait_for_devserver_build`and return results as well.

packages/angular/cli/src/commands/mcp/tools/start-devserver.ts
dgp1130: I think everyone's a little different. I prefer issues as I find those more discoverable and actionable than `TODO` comments, but there's no real standard process.

packages/angular/cli/src/commands/mcp/tools/build.ts
```diff
@@ -11,84 +11,68 @@ import { CommandError, Host, LocalWorkspaceHost } from '../host';
...
-
   // Build "ng"'s command line.
   const args = ['build'];
   if (input.project) {
     args.push(input.project);
   }
-  if (configuration.args) {
-    args.push(configuration.args);
-  }
+  args.push(`-c ${input.configuration ?? DEFAULT_CONFIGURATION}`);
```
dgp1130: Bug: I think we want `-c` and `input.configuration` to be two different arguments?

```typescript
args.push('-c', input.configuration ?? DEFAULT_CONFIGURATION);
```

packages/angular/cli/src/commands/mcp/tools/build.ts
dgp1130: I think this might have gotten dropped?

packages/angular/cli/src/commands/mcp/tools/wait-for-devserver-build.ts
```diff
@@ -25,7 +25,7 @@ const waitForDevserverBuildToolInputSchema = z.object({
     ),
   timeout: z
     .number()
-    .default(30000)
+    .default(180000)
```
dgp1130: Suggestion: Use an [underscore separator](https://v8.dev/features/numeric-separators) to make the zeroes a little easier to read. I also suggest including units somewhere like:

```typescript
z.default(180_000 /* ms */)
```

packages/angular/cli/src/commands/mcp/tools/modernize_spec.ts
```diff
@@ -28,7 +29,7 @@ describe('Modernize Tool', () => {
       existsSync: jasmine.createSpy('existsSync').and.callFake((p: string) => {
         return p === join(projectDir, 'angular.json');
       }),
-    } as Partial<Host> as Host;
+    } as Partial<MockHost> as MockHost;
```
dgp1130, line 32: Consider: Is it worth using `MockHost` here?

dgp1130 (approved):
Don't forget to squash or switch to `fixup!` commits before merging, so "Address review comments" doesn't end up in our permanent history (it's fine for a PR, just don't want it being merged to `main`).

packages/angular/cli/src/commands/mcp/tools/wait-for-devserver-build.ts
clydin: We can leave it at 3 minutes for now and adjust as we perform more testing on various size projects.

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Discussed with @clydin, we're treating filesystem changes as non-read-only even if they don't change source files, but we need to take into consideration that `ng serve` builds in memory and not into disk. So, conclusion: `build` will remain `isReadOnly: false` while the devserver tools will be `isReadOnly: true`.

packages/angular/cli/src/commands/mcp/tools/build_spec.ts
amishne: Resolving for now; I'll implement the env variable approach in a follow-up.

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: Ignoring the issue for now :-) but I added this under "operational notes" in the tool's description, to make it explicit for the agent.

packages/angular/cli/src/commands/mcp/tools/serve.ts
amishne: `ng serve` still starts up and watches the workspace even if the initial build is a failure. For now I'll just amend the description of the tool to make it clear to the agent what the different roles of `start_devserver` and `wait_for_devserver_build` are and the calling sequence... we'll change it later if it becomes confusing.

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: Not sure what was going on with this, it's definitely there now

packages/angular/cli/src/commands/mcp/tools/build.ts
amishne: This was intentional since it's supported like this as well, but I'll change it.

packages/angular/cli/src/commands/mcp/tools/modernize_spec.ts
amishne: It does save on a cast and we already have it, plus it's more consistent this way.

packages/angular/cli/src/commands/mcp/tools/wait-for-devserver-build.ts
amishne: Done.

alan-agius4:

- main: c3c9ac5067275461e2d8caefba81ac9701949776

PR #31676, opened 2025-11-03
refactor(@angular/build): improve vitest `runnerConfig` support via plugin

clydin:
Refactors the Vitest unit test runner to use a dedicated plugin for merging builder-defined configurations with user-provided configurations from a `runnerConfig` file.

Previously, the configuration merging logic was handled directly in the executor, which had limitations and did not always correctly apply user overrides. By moving this logic into a Vitest plugin, we leverage Vitest's intended extension mechanism, ensuring a more robust and predictable merge of configurations. This significantly improves the ability for users to customize their test setup.

Adds a new test suite to verify that common custom configurations, such as custom reporters, file exclusions, option overrides, and environment settings, are correctly applied from a `vitest.config.ts` file.


alan-agius4:

- main: ae1b00db177c99a1dd217fccfa1fec302a8b69ca
- 21.0.x: 5dd4f8f04c874ad6fd1f43356276f1fafbe8a692

PR #31685, opened 2025-11-04
fix(@angular/ssr): improve route matching for wildcard routes

### LINKED ISSUE #31666: Problems and errors with enforcing SSR using ServerRoutes

platform-server

### Is this a regression?

No

### Description

I'm trying to get our SSR application to work with application-builder and the new `AngularNodeAppEngine`.

In our use case we only want to do SSR, no SSG. Also, we do not want to make use of any `ServerRoutes` features, as anyway all routes should be SSR, and redirects or status codes will already be handled by the application logic.

In order to enforce all routes being SSR, I had to explicitly configure the app server config with:
```typescript
provideServerRendering(
  withRoutes([
    {
      path: '**',
      renderMode: RenderMode.Server
    }
  ])
)
```
## Issue 1:
There's no global switch to simply turn off SSG and I'm forced to add `withRoutes` with a catch-all route, even though I do not want to use any of the related features. 

Hence it would be very useful to introduce a `withNoRoutes()` option, which also should disable the initial route evaluation/extraction while running a dev server - which is simply wasted computation time in our case.

As I do not want SSG, I already configured `"outputMode": "server"` in angular.json. I'm wondering why this setting isn't enough to turn off SSG completely, so that all pages will be SSR by default, not requiring `withRoutes()` configured.

## Issue 2:
Our routing setup includes a lot of lazy-load child routes. If one of the child routes has a matcher applied, then I get the below error:
> Error: Error(s) occurred while extracting routes:
> The route 'test/**' has a defined matcher but does not match any route in the server routing configuration. Please ensure this route is added to the server routing configuration.

My (simplified) routing setup is as follows:

App routes:
```typescript
{
  path: 'test',
  loadChildren: () => import('@my-app/test/feature').then(m => m.testFeatureRoutes)
},
```
Test feature routes:
```typescript
  {
    path: '',
    pathMatch: 'full',
    component: TestComponent
  },
  {
    matcher: () => null, <-- this guy here causes the error
    component: TestComponent
  },
```

Am I really forced now to add for all such routes a dedicated entry like `test/**` to ServerRoutes? In cases with very dynamic routing it seems to me also impossible or at least cumbersome to add all required sub-route catch-alls. Also, it's actually quite bad DX if one is bothered with having to solve errors for features he/she doesn't want to use anyway. 

So, to me the both issues I described are tightly coupled: 
If I would be able to turn off server routing completely, I wouldn't have to solve the problems with the catch-alls when using matchers.

Looking forward to some answers explaining to me what can be done to get around the above mentioned obstacles.

-

```true
See above
```

```true
Angular CLI: 20.3.5
Node: 22.18.0
Package Manager: npm 10.9.3
OS: linux x64
    

Angular: 20.3.4
... platform-server, router

---------------------------------------
@angular-devkit/architect       0.2003.5
@angular-devkit/build-angular   20.3.5
@angular-devkit/core            20.3.5
@angular-devkit/schematics      20.3.5
@angular/cdk                    20.2.8
@angular/cli                    20.3.5
@angular/google-maps            20.2.8
@angular/ssr                    20.3.5
@angular/youtube-player         20.2.8
@schematics/angular             20.3.5
ng-packagr                      20.3.0
rxjs                            7.8.2
typescript                      5.9.3
zone.js                         0.15.1
```

### Anything else?

_No response_

alan-agius4:
Enhance the route traversal logic to correctly identify and process wildcard route matchers (e.g., '**'). This ensures that routes with matchers are properly marked as present in the client router and handled according to their render mode.

closes #31666

### DIFF
--- a/packages/angular/ssr/src/routes/ng-routes.ts
+++ b/packages/angular/ssr/src/routes/ng-routes.ts
@@ -251,15 +251,22 @@ async function* traverseRoutesConfig(options: {
     const currentRoutePath = joinUrlParts(parentRoute, path);
 
     if (matcher && serverConfigRouteTree) {
-      let foundMatch = false;
+      const matches: (RouteTreeNodeMetadata & ServerConfigRouteTreeAdditionalMetadata)[] = [];
       for (const matchedMetaData of serverConfigRouteTree.traverse()) {
-        if (!matchedMetaData.route.startsWith(currentRoutePath)) {
-          continue;
+        if (matchedMetaData.route.startsWith(currentRoutePath)) {
+          matches.push(matchedMetaData);
         }
+      }
 
-        foundMatch = true;
+      if (!matches.length) {
+        const matchedMetaData = serverConfigRouteTree.match(currentRoutePath);
+        if (matchedMetaData) {
+          matches.push(matchedMetaData);
+        }
+      }
+
+      for (const matchedMetaData of matches) {
         matchedMetaData.presentInClientRouter = true;
-
         if (matchedMetaData.renderMode === RenderMode.Prerender) {
           yield {
             error:
@@ -282,7 +289,7 @@ async function* traverseRoutesConfig(options: {
         });
       }
 
-      if (!foundMatch) {
+      if (!matches.length) {
         yield {
           error:
             `The route '${stripLeadingSlash(currentRoutePath)}' has a defined matcher but does not ` +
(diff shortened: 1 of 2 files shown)


alan-agius4:

- main: 4eb22bc2e6bb90aa8d9d617e864e7976da3feb5a
- 20.3.x: 683697ebc5e129d2b17bded9e4ff318d598e0bd3
- 21.0.x: 4a3cfdfce75b7cedf64e26d85b0a92ec92ddd12d

PR #31692, opened 2025-11-04
fix(@schematics/angular): add `addImports` option to jasmine-vitest schematic

clydin:
Introduces a new `--add-imports` boolean option to the Jasmine to Vitest refactoring schematic.

When this option is set to `true`, the schematic will automatically add a single, consolidated `import { ... } from 'vitest';` statement at the top of the transformed file for any Vitest APIs that are used.

Key changes:
- Type-only imports (e.g., `Mock`, `MockedObject`) are now always added to ensure the refactored code remains type-correct.
- Value imports (e.g., `vi`, `describe`, `it`, `expect`) are only added when `--add-imports` is `true`.
- The import generation logic correctly handles all scenarios, including type-only imports (`import type {...}`), value-only imports, and combined imports with inline `type` keywords.
- A new test suite has been added to validate the import generation logic under various conditions.

### DIFF
--- a/packages/schematics/angular/refactor/jasmine-vitest/index.ts
+++ b/packages/schematics/angular/refactor/jasmine-vitest/index.ts
@@ -119,7 +119,9 @@ export default function (options: Schema): Rule {
     for (const file of files) {
       reporter.incrementScannedFiles();
       const content = tree.readText(file);
-      const newContent = transformJasmineToVitest(file, content, reporter);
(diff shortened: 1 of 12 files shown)


hybrist (approved):

clydin:

- main: cc132e9524a136575685961b2ef934eb0aca5c60
- 21.0.x: f89750b27866c307da546fe4f33da980693ca5c1

PR #31720, opened 2025-11-07
fix(@schematics/angular): correct `tsconfig.spec.json` include for spec files

alan-agius4:
Updates the  files in both application and library schematics to specifically include  instead of . This ensures that only test specification files are processed for testing, preventing unintended inclusion of other TypeScript files.


alan-agius4:

- main: 5b0afb324d6fa0acfc4f53394f728f9ffb65dc63
- 20.3.x: c854a719bb3a8d92fe42c8fba4b0b1789081b21c
- 21.0.x: 02f579f6e7e490744deea5a193b0751ce31262db

PR #31721, opened 2025-11-07
fix(@angular/build): builder removes @angular/localize when external packages are not inlined

roma-claudio:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`@angular/localize` scripts are always removed when inline transformations are enabled, even though not all the scripts/dependencies might have been transformed

Issue Number: N/A

## What is the new behavior?

`@angular/localize` scripts will only be removed when inline transformations are enabled and all the scripts/dependencies are actually transformed

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
+++ b/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
@@ -654,7 +654,7 @@ function getEsBuildCommonPolyfillsOptions(
   tryToResolvePolyfillsAsRelative: boolean,
   loadResultCache: LoadResultCache | undefined,
 ): BuildOptions | undefined {
-  const { jit, workspaceRoot, i18nOptions } = options;
+  const { jit, workspaceRoot, i18nOptions, externalPackages } = options;
 
   const buildOptions = getEsBuildCommonOptions(options);
   buildOptions.splitting = false;
@@ -671,8 +671,10 @@ function getEsBuildCommonPolyfillsOptions(
   // Locale data should go first so that project provided polyfill code can augment if needed.
   let needLocaleDataPlugin = false;
   if (i18nOptions.shouldInline) {
-    // Remove localize polyfill as this is not needed for build time i18n.
-    polyfills = polyfills.filter((path) => !path.startsWith('@angular/localize'));
+    if (!externalPackages) {
+      // Remove localize polyfill when i18n inline transformation have been applied to all the packages.
+      polyfills = polyfills.filter((path) => !path.startsWith('@angular/localize'));
+    }
 
     // Add locale data for all active locales
     // TODO: Inject each individually within the inlining process itself


alan-agius4:
I'm a bit confused by the description of this change. `i18n` inlining is controlled by the `localize` option and happens regardless of the `optimization` setting.

Currently, `@angular/localize` is already removed only when inlining is required https://github.com/roma-claudio/angular-cli/blob/c3c9ac5067275461e2d8caefba81ac9701949776/packages/angular/build/src/tools/esbuild/application-code-bundle.ts#L674-L675 as otherwise it will be required as a runtime dependency irrespective of the `optimization` option value.

roma-claudio:
Hi @alan-agius4 

Apologies for the misleading reference, at first I thought that I'd check if there are Vite's pre-bundles dependencies (which are not inlined) by looking at the optimized script presence.

Now let me clarify the issue that I've found:

The _i18n_ inliner processes `executionResult.outputFiles` which from what I've seen do not include Vite's pre-bundled dependencies.
If the polyfills localize is removed, those dependencies which contains `$localize` will fail to execute.

Ideally, the localize script should only be removed if there are no pre-bundled dependencies containing `$localize`. 
I'm looking for a way to check for this condition before removing or not the import 💭

What do you think?

alan-agius4:
Oh I now understand the problem, but for that case, using the optimization option is not the correct approach. In addition as a workaround you could also exclude these deps from bring pre-bundled. Using the dev-server `prebundle.exclude` option.

packages/angular/build/src/tools/esbuild/application-code-bundle.ts
```diff
@@ -671,8 +671,12 @@ function getEsBuildCommonPolyfillsOptions(
...
   let needLocaleDataPlugin = false;
   if (i18nOptions.shouldInline) {
-    // Remove localize polyfill as this is not needed for build time i18n.
-    polyfills = polyfills.filter((path) => !path.startsWith('@angular/localize'));
+    const isOptimizedBuild = options.optimizationOptions.scripts;
+    if (isOptimizedBuild) {
+      // Remove localize polyfill where script optimizations are enabled
+      // and therefore, i18n inlining transform has occurred.
+      polyfills = polyfills.filter((path) => !path.startsWith('@angular/localize'));
+    }
```
alan-agius4: The correct option is `externalPackages`

```suggestion
    if (!externalPackages) {
      // Please add comment explaining why. 
      polyfills = polyfills.filter((path) => !path.startsWith('@angular/localize'));
    }
```

Add `externalPackages` in the deconstructed object in L 659

roma-claudio:
Thanks @alan-agius4 

Code updated 👍🏻

packages/angular/build/src/tools/esbuild/application-code-bundle.ts
```diff
@@ -671,8 +671,11 @@ function getEsBuildCommonPolyfillsOptions(
   // Locale data should go first so that project provided polyfill code can augment if needed.
   let needLocaleDataPlugin = false;
   if (i18nOptions.shouldInline) {
-    // Remove localize polyfill as this is not needed for build time i18n.
-    polyfills = polyfills.filter((path) => !path.startsWith('@angular/localize'));
+    const isOptimizedBuild = options.optimizationOptions.scripts;
```
alan-agius4:
```suggestion
```

alan-agius4 (approved):

alan-agius4 (approved):
LGTM, thanks for this.

alan-agius4:

- main: 38c22001330eaf2b37b8a0a04fdda50666458664
- 20.3.x: b3908f68ed2f05cee56cbf0d9895dcfc3dc0ac25
- 21.0.x: f82b77e475e270dd0d23ce5d6b445c0a8aa8599a

PR #31722, opened 2025-11-07
fix(@angular/build): simplify SSL handling for `ng serve` with SSR 

### LINKED ISSUE #31710: HttpClient request fails in development during SSR with SSL enabled
### Command

serve

### Is this a regression?

_No response_

### Description

When using `ng serve` with SSR and an SSL certificate the HttpClient request fails during SSR

GitHub repo: https://github.com/brandonroberts/angular-ssl-assets

## Setup

1. Install dependencies

```sh
npm install
```

2. Run application

```sh
npm start
```

3. Navigate to `https://localhost:4430` and notice the error in the command line console, and the flash of data on the page.

## Workaround

The workaround is to run `NODE_TLS_REJECT_UNAUTHORIZED=0 ng serve` 

1. Run application

```sh
npm run start:cert
```

2. Navigate to `https://localhost:4430` and notice the Node warning in the command line console, but no flash of data on the page.

### Exception or Error

```text
ERROR HttpErrorResponse {
  headers: _HttpHeaders {
    headers: Map(0) {},
    normalizedNames: Map(0) {},
    lazyInit: undefined,
    lazyUpdate: null
  },
  status: 0,
  statusText: 'Unknown Error',
  url: 'https://localhost:4430/assets/data.json',
  ok: false,
  type: undefined,
  redirected: undefined,
  name: 'HttpErrorResponse',
  message: 'Http failure response for https://localhost:4430/assets/data.json: 0 undefined',
  error: TypeError: fetch failed
      at node:internal/deps/undici/undici:13510:13
      at _ZoneDelegate.invoke (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:336:158)
      at ZoneImpl.run (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:105:35)
      at eval (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:1040:30)
      at _ZoneDelegate.invokeTask (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:362:171)
      at ZoneImpl.runTask (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:143:37)
      at drainMicroTaskQueue (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:522:23)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
    [cause]: Error: unable to verify the first certificate
        at TLSSocket.onConnectSecure (node:_tls_wrap:1679:34)
        at _ZoneDelegate.invokeTask (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:362:171)
        at ZoneImpl.runTask (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:143:37)
        at invokeTask (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:439:26)
        at TLSSocket.ZoneTask.invoke (/Users/computer/projects/angular-ssl-assets/.angular/cache/20.3.9/angular-ssl-assets/vite/deps_ssr/zone__js_node.js:428:38)
        at TLSSocket.emit (node:events:518:28)
        at TLSSocket._finishInit (node:_tls_wrap:1078:8)
        at ssl.onhandshakedone (node:_tls_wrap:864:12) {
      code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
    }
  }
}
```

### Your Environment

```text
Angular: 20.3.9
... build, cli, common, compiler, compiler-cli, core, forms
... platform-browser, platform-server, router, ssr

@angular-devkit/architect    0.2003.9
@angular-devkit/core         20.3.9
@angular-devkit/schematics   20.3.9
@schematics/angular          20.3.9
```

### Anything else relevant?

Its a self-signed certificate for local development using `mkcert localhost`.

alan-agius4:
This commit simplifies the handling of self-signed SSL certificates for `ng serve` with SSR. Previously, tests and potentially users had to set `NODE_TLS_REJECT_UNAUTHORIZED` or `NODE_EXTRA_CA_CERTS` to bypass certificate validation issues with self-signed certificates.

Closes #31710

### DIFF
--- a/packages/angular/build/src/tools/vite/plugins/ssr-ssl-plugin.ts
+++ b/packages/angular/build/src/tools/vite/plugins/ssr-ssl-plugin.ts
@@ -16,10 +16,10 @@ export function createAngularServerSideSSLPlugin(): Plugin {
     async configureServer({ config, httpServer }) {
       const {
         ssr,
-        server: { https: cert },
+        server: { https },
       } = config;
 
-      if (!ssr || !cert) {
+      if (!ssr || !https?.cert) {
         return;
       }
 
@@ -27,6 +27,7 @@ export function createAngularServerSideSSLPlugin(): Plugin {
       // See: https://nodejs.org/api/tls.html#tlssetdefaultcacertificatescerts
       const { getGlobalDispatcher, setGlobalDispatcher, Agent } = await import('undici');
       const originalDispatcher = getGlobalDispatcher();
+      const { cert } = https;
       const certificates = Array.isArray(cert) ? cert : [cert];
 
       setGlobalDispatcher(


packages/angular/build/src/tools/vite/plugins/ssr-ssl-plugin.ts
```diff
@@ -0,0 +1,62 @@
...
+        return;
+      }
+
+      const { getGlobalDispatcher, setGlobalDispatcher, Agent } = await import('undici');
+      const originalDispatcher = getGlobalDispatcher();
+
+      setGlobalDispatcher(
+        new Agent({
+          connect: {
+            ca: [...getCerts(https.key), ...getCerts(https.cert)].join('\n'),
```
hybrist: I _think_ `https.key` is the private key which means it shouldn't have to be added here since it's about trusting "other party's" certs. It likely doesn't break things but better not to pass the private keys around even if they are dev-only.

Also, does this _replace_ the trusted root certs? https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions Maybe we'd need to also explicitly include https://nodejs.org/api/tls.html#tlsgetcacertificatestype?

hybrist (changes requested):

packages/angular/build/src/tools/vite/plugins/ssr-ssl-plugin.ts
alan-agius4: You are right, we should remove the `key`. This is an alternate approach of using `setDefaultCACertificates` as currently this is not available on all Node.js versions that we support. I'll add a comment.

hybrist: Should we at least add https://nodejs.org/api/tls.html#tlsrootcertificates? Or am I missing something about where we retain the certificates necessary for calling https APIs from other origins?

alan-agius4: Yes, we should add `rootCertificates`

alan-agius4:

- main: ad99e00ad7edd17e369777c8d38b4137ea736121

PR #31729, opened 2025-11-09
fix(@angular/build): enhance Vitest config merging and validation

clydin:
This change improves how user-defined Vitest configurations work with the Angular CLI's unit test builder.

The builder now checks for and handles specific options in `vitest-base.config.ts`. It detects the `test.projects` option, logs a warning, and removes it to prevent conflicts. The `test.include` option is handled in a similar way, ensuring the builder's test discovery is used. Any `test.setupFiles` from `vitest-base.config.ts` are now added to the CLI's setup files, supporting both single string and array formats. User-defined Vite plugins from `vitest-base.config.ts` are also combined with the builder's plugins, with a filter to prevent duplicating internal CLI plugins.

These updates give users more flexibility to customize their Vitest setup while keeping the Angular CLI's test builder predictable.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/plugins.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/plugins.ts
@@ -40,7 +40,7 @@ interface VitestConfigPluginOptions {
   projectSourceRoot: string;
   reporters?: string[] | [string, object][];
   setupFiles: string[];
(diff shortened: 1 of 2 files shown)


clydin:

- main: 0aab1158d63ad13f1dc02c40608bac2f68bdfc93
- 21.0.x: ae35543af7f5b3a5328c39fd4617d61b48067357

PR #31731, opened 2025-11-09
fix(@angular/build): ensure TestBed setup is robust in non-isolated Vitest

clydin:
This commit refines the Angular TestBed initialization within the Vitest runner to ensure correct behavior in non-isolated test environments (`isolate: false`).

Previously, in a non-isolated setup, the TestBed initialization logic in the virtual setup file could be problematic if the file was evaluated multiple times. This could lead to re-registering `beforeEach`/`afterEach` hooks or attempting to re-initialize the TestBed, causing errors or unexpected test behavior.

This change introduces a globally unique symbol guard (`Symbol.for('@angular/cli/testbed-setup')`) that ensures the entire TestBed setup block (including hook registration and `initTestEnvironment`) is executed only once per test run. This prevents redundant executions and potential errors.

Finally, the `isolate` option in Vitest is now explicitly defaulted to `false` when not specified by the user, aligning with the traditional Karma/Jasmine experience.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
@@ -49,17 +49,29 @@ function createTestBedInitVirtualFile(
     import { getTestBed, ɵgetCleanupHook as getCleanupHook } from '@angular/core/testing';
     import { BrowserTestingModule, platformBrowserTesting } from '@angular/platform-browser/testing';
     ${providersImport}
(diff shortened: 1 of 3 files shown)


yjaaidi:
> the isolate option in Vitest is now explicitly defaulted to false

🚀 awesome! Thanks Charles!

alan-agius4 (approved):

alan-agius4:

- main: 377780a1c2878f1028ae7a74f8c5387863a27974
- 21.0.x: 41b12509a9db8bca637e0c67d21301a75774129c

delagen:
@alan-agius4 Seems no changes

image

PR #31752, opened 2025-11-11
fix(@angular/build): ensure `ɵgetOrCreateAngularServerApp`  is always defined after errors

alan-agius4:
Addresses an issue where  could become `ɵgetOrCreateAngularServerApp` undefined after an error, leading to subsequent rendering failures.

This change modifies the HMR process to include a timestamp when loading. This ensures the server application is always re-evaluated, preventing stale application states.


clydin (approved):

alan-agius4:
The changes were merged into the following branches: main, 21.0.x

PR #31770, opened 2025-11-12
build: fix lock file

hybrist:
It looks like this lockfile got broken (somehow) in https://github.com/angular/angular-cli/pull/31761 or while landing the PR. I'm not sure how that was possible since CI is passing for the PR but not after landing on main..?


hybrist:
(Main is currently red without the regenerated lock file.)

clydin (approved):

PR #31781, opened 2025-11-13
fix(@angular/build): enhance Vitest dependency externalization and pre-bundling

clydin:
This commit refactors the dependency handling for the Vitest runner to improve flexibility and correctness, especially for browser-based tests.

The previously hardcoded list of external Angular packages has been removed. Instead, a new `externalPackages: true` option is passed to the application builder, allowing for more dynamic and configurable externalization of packages. This prevents packages from `node_modules` from being bundled into the test entry points.

Note: For cases where a dependency should not be pre-bundled, the `optimizeDeps.exclude` option can be used within a custom runner configuration file.


alan-agius4 (approved):

clydin:

- main: 6e0f5da9fceeb359e2e476328a17ef78c43391db
- 21.0.x: fec106b60553394aab51d713e5437a713085089b

PR #31782, opened 2025-11-13
refactor: address issue with bazel resolution of peer dependencies (main)

alan-agius4:
Key changes include:

- **Removal of PNPM Hook:** The `.pnpmfile.cjs` file, which previously converted peer dependencies into
  direct dependencies, has been deleted. This workaround is no longer necessary with the adoption of dynamic
  imports. The `MODULE.bazel` file has been updated to remove references to it.

- **Dynamic Imports**
  - In the `build_webpack` package, static `import` statements for `webpack` and `webpack-dev-server` are
    replaced with dynamic `import()` calls. This defers their loading until they are actively required during
    the process.
  - In the `ssr-dev-server`, `browser-sync` is now loaded using `createRequire` relative to the workspace
    root, improving resolution reliability.

- **Build Configuration:** Associated `BUILD.bazel` files have been updated to remove dependency declarations that are now obsolete due to the new lazy-loading approach.

### DIFF
--- a/goldens/public-api/angular_devkit/build_webpack/index.api.md
+++ b/goldens/public-api/angular_devkit/build_webpack/index.api.md
@@ -7,8 +7,8 @@
 import { BuilderContext } from '@angular-devkit/architect';
 import { BuilderOutput } from '@angular-devkit/architect';
 import { Observable } from 'rxjs';
-import webpack from 'webpack';
-import WebpackDevServer from 'webpack-dev-server';
+import type webpack from 'webpack';
+import type WebpackDevServer from 'webpack-dev-server';
 
 // @public (undocumented)
 export type BuildResult = BuilderOutput & {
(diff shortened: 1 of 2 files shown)


packages/angular_devkit/build_angular/src/builders/ssr-dev-server/index.ts
```diff
@@ -64,7 +65,7 @@ export function execute(
 ): Observable<SSRDevServerBuilderOutput> {
   let browserSync: typeof import('browser-sync');
   try {
-    browserSync = require('browser-sync');
+    browserSync = createRequire(context.workspaceRoot + '/')('browser-sync');
```
alan-agius4, line 68: `browser-sync` is installed in the project (app), and thus the resolution should happen from the workspaceRoot not from this package.

packages/angular_devkit/build_webpack/src/builders/webpack/index.ts
```diff
@@ -57,7 +57,7 @@ export function runWebpack(
         return of(result);
       }
     } else {
-      return of(webpack(c));
+      return from(import('webpack').then((mod) => mod.default(c)));
```
alan-agius4, line 60: When a factory is not provided we should need to try to resolve webpack from this directory. This will fail when using the peerDeps of build-angular due to how modules are hoisted.

devversion (approved):

alan-agius4:

- main: f826968d438e82f44217e09e60b9339190099dfd

Issue #31732, opened 2025-11-09
`unit-test` builder setup files in vitest and browser mode do not get the right instance of `TestBed`


yjaaidi:
### Command

test

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

When using unit test builder + vitest + browser mode, adding the following setup file

```ts
import { getTestBed } from '@angular/core/testing';

getTestBed().configureTestingModule({});
```

causes the following error:

```
Error: Need to call TestBed.initTestEnvironment() first
```

After dumping virtual files, I do not see the setup file. I am afraid that there might be a duplicate instance of `TestBed`, one in the chunk produced by Angular CLI and one directly loaded from `node_modules` by vitest.

I think that the setup files should be added to the entrypoints here: https://github.com/angular/angular-cli/blob/0aab1158d63ad13f1dc02c40608bac2f68bdfc93/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts#L99
It would be also nice to make sure `sequence.setupFiles` is set to `list`.

### Minimal Reproduction

- add `src/test-setup.ts` with:
```ts
import { getTestBed } from '@angular/core/testing';

getTestBed().configureTestingModule({});
```
- add it to setupFiles option: `"setupFiles": ["./src/test-setup.ts"]`
- configure browsers: `"browsers": ["ChromiumHeadless"]`
- run tests

### Exception or Error

```text
Error: Need to call TestBed.initTestEnvironment() first
```

### Your Environment

```text
Angular CLI       : 21.1.0-next.0+sha-0aab115
Angular           : 21.0.0-rc.1
Node.js           : 24.8.0
Package Manager   : pnpm 10.17.0
Operating System  : darwin arm64

┌───────────────────────────┬───────────────────────────┬───────────────────────────┐
│ Package                   │ Installed Version         │ Requested Version         │
├───────────────────────────┼───────────────────────────┼───────────────────────────┤
│ @angular/build            │ 21.0.0-rc.1               │ ^21.0.0-rc.1              │
│ @angular/cli              │ 21.1.0-next.0+sha-0aab115 │ github:angular/cli-builds │
│ @angular/common           │ 21.0.0-rc.1               │ ^21.0.0-rc.1              │
│ @angular/compiler         │ 21.0.0-rc.1               │ ^21.0.0-rc.1              │
│ @angular/compiler-cli     │ 21.0.0-rc.1               │ ^21.0.0-rc.1              │
[...]

PR #31791, opened 2025-11-13
fix(@angular/build): fix setup files duplicate modules

yjaaidi:
This includes setup files in the initial build and avoids lazy discovery and thus module duplicates.
Module duplicates can break many things such as dependency injection.

⚠️ this only handles `setupFiles` directly provided to the builder.
Handling `setupFiles` from the Vitest config requires substantial additional effort. 

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #31732 

- [ ] Yes
- [X] No

### DIFF
--- a/packages/angular/build/src/builders/unit-test/test-discovery.ts
+++ b/packages/angular/build/src/builders/unit-test/test-discovery.ts
@@ -73,7 +73,7 @@ export async function findTests(
   return [...resolvedTestFiles];
 }
 
-interface TestEntrypointsOptions {
+interface EntrypointsOptions {
   projectSourceRoot: string;
   workspaceRoot: string;
   removeTestExtension?: boolean;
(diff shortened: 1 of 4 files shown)


yjaaidi:
I tried to keep this the less intrusive possible but I wonder if the right fix wouldn't involve changes to `TestRunner` API.
Feel free to either pick this a quick fix and revert it after the release for a proper fix, or just see it as a spike and throw it away 😊

packages/angular/build/src/builders/unit-test/test-discovery.ts
```diff
@@ -89,15 +89,49 @@ interface TestEntrypointsOptions {
...
+    prefix: 'spec',
+  });
+}
+
+/**
+ * @param setupFiles An array of absolute paths to setup files.
+ * @param options Configuration options for generating entry points.
+ * @returns A map where keys are the generated unique bundle names and values are the original file paths.
+ */
+export function getSetupEntrypoints(
```
yjaaidi, line 107: I'm not happy with this.
I would rather have a single `getEntrypoints` that returns both test files and setup files then we would have to forward the setup file names to the executor somehow.

packages/angular/build/src/builders/unit-test/test-discovery.ts
```diff
@@ -89,15 +89,49 @@ interface TestEntrypointsOptions {
...
+  setupFiles: string[],
+  { projectSourceRoot, workspaceRoot, removeTestExtension }: EntrypointsOptions,
+): Map<string, string> {
+  return getEntrypoints(setupFiles, {
+    projectSourceRoot,
+    workspaceRoot,
+    removeTestExtension,
+    prefix: 'setup',
```
yjaaidi, line 115: This produces bundles named `setup-test-setup` which is weird.
An empty string would be better but collisions could happen with a setup file called `spec-setup...` or something.

alan-agius4:
Closing as this original issue has been solved via https://github.com/angular/angular-cli/pull/32428

PR #31796, opened 2025-11-13
test: stabilize Vitest snapshot E2E test on Windows 

clydin:
This commit introduces several changes to fix a non-deterministic E2E test for Vitest snapshots that was failing on Windows.

The primary fix is to standardize the line endings of the test spec file to LF (`\n`) before running the test. This works around an issue where Vitest can miscalculate line counts when encountering mixed line endings (CRLF and LF), which is common in Windows environments.

Additionally, the snapshot content assertion has been updated to match Vitest's current format, which includes the test suite name. The file modification logic was also refactored to use `replaceInFile` for better reliability.


hybrist:

- main: 8bde4a3c5004c45572a17e7b7798297517b4fe97
- 21.0.x: 8a53e2d5a46ea6a63ab61e4d997088b6ffe148e9

PR #31797, opened 2025-11-13
fix(@angular/build): remove explicit test isolation configuration

clydin:
This commit removes the explicit `isolate` option from the Vitest configuration within the Angular CLI builder.

Previously, the `isolate` option was explicitly set to `false` for DOM emulation and browser mode. However, due to observed non-deterministic failures and the need for consistent isolation behavior across all environments, it's more robust to allow Vitest to manage test isolation by its own default mechanisms. This change ensures that Vitest's internal defaults for isolation are respected, which are generally designed for optimal reliability.


clydin:

- main: 138649ee61a3e3f2390a448630782e75b2d69277
- 21.0.x: 6576bb5985c18dca7cecd9509939c2a78bf9758a

PR #31814, opened 2025-11-14
fix(@angular/build): ensure TestBed cleanup hooks are always registered

clydin:
This commit moves the `beforeEach` and `afterEach` hook registrations for TestBed cleanup to be outside the global setup guard.

In environments where test isolation is not strictly enforced (when `isolate: false`), the setup file can be executed multiple times in different contexts. The global guard would previously prevent the hooks from being re-registered, leading to inconsistent test behavior and potential state leakage between tests.

This change ensures the cleanup hooks are always registered, improving the reliability and predictability of tests across all execution environments.


hybrist (approved):

clydin:

- main: 79e52ccf7b785cc8877f878ab20636a30eb2c5c4
- 21.0.x: 0830f4fb549e2c45b1ef752dd42f002a1347d7c8

PR #31821, opened 2025-11-14
fix(@angular/cli): support string | boolean union types in CLI options

clydin:
This commit enhances the CLI's command builder to correctly handle JSON schema properties that define a union type of `string` and `boolean`. Previously, only the first type in a union was considered, preventing options like `--runner-config` from being used as both a boolean flag and a string value.


alan-agius4:
Will false in the following case `—option=false` be parsed as a Boolean or string?

would be great to add a couple of unit tests.

PR #31840, opened 2025-11-17
fix(@angular/build): configure Vitest cache to use Angular cache

alan-agius4:
Configures Vitest to use the Angular cache directory instead of its default location within `node_modules/.vite`. This change ensures that Vitest's cache is managed consistently with other Angular build artifacts, avoiding unnecessary writes to `node_modules`.


alan-agius4:

- main: c241dd8f1b8885a752335e3fad5f9a5a56d2d1ae
- 21.0.x: a11dd31f0c80a189e7dcb3172c89a731cfc34702

PR #31870, opened 2025-11-19
build: update FW and `ng-packagr` dependencies for major release

dgp1130:


dgp1130:
@alan-agius4 good call out. No `target: *` label goes just to `21.0.x` right? Do I just switch the PR target from `main` to `21.0.x` directly?

alan-agius4:
Yes, you can change the PR target directly, although likely it will pick up wrong commits. 

The `merge` label is correct.

PR #31878, opened 2025-11-19
Several improvements for Jasmine to Vitest refactor schematic

clydin:
Changes include:
* Enhances the jasmine-vitest schematic by adding mock names to `vi.fn()` instances created from `jasmine.createSpyObj`. This improves debugging in Vitest by providing meaningful names for mock functions, making test failures easier to understand and trace.
* Adds a safety check to the `done` callback transformation in the `jasmine-vitest` schematic. Previously, if a `done` callback was used in an unhandled way (e.g., passed as an argument to a helper function), the schematic would remove the `done` parameter but leave the usage, causing runtime errors.
* Improves the transformation of `jasmine.arrayWithExactContents` in the `jasmine-vitest` schematic. While Vitest does not have a direct equivalent for multiset equality, the schematic previously transformed this into a check for length and subset containment, which could lead to false positives (e.g., `[1, 1]` matches `[1, 2]`). Now, the schematic attaches a TODO comment to the generated code, explicitly warning the user that the transformation relies on `expect.arrayContaining` (a subset check) and advising them to verify strict content matching manually.

### DIFF
--- a/packages/schematics/angular/refactor/jasmine-vitest/transformers/jasmine-matcher_spec.ts
+++ b/packages/schematics/angular/refactor/jasmine-vitest/transformers/jasmine-matcher_spec.ts
@@ -225,19 +225,25 @@ expect(mySpyObj).toHaveSpyInteractions();`,
       {
         description: 'should transform toEqual(jasmine.arrayWithExactContents()) into two calls',
         input: `expect(myArray).toEqual(jasmine.arrayWithExactContents(['a', 'b']));`,
+        /* eslint-disable max-len */
         expected: `
+          // TODO: vitest-migration: Verify this matches strict array content (multiset equality). Vitest's arrayContaining is a subset check.
           expect(myArray).toHaveLength(2);
           expect(myArray).toEqual(expect.arrayContaining(['a', 'b']));
         `,
+        /* eslint-enable max-len */
       },
       {
         description:
           'should transform toEqual(jasmine.arrayWithExactContents()) with asymmetric matchers',
         input: `expect(myArray).toEqual(jasmine.arrayWithExactContents([jasmine.any(Number), 'a']));`,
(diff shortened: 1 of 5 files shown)


clydin:

- main: 7c6a643f94a9b6abaf6578e9e98eb45da3bc17b2
- 21.0.x: 145de4a584ce8f72746704547282299306d9bafb

packages/schematics/angular/refactor/jasmine-vitest/transformers/jasmine-matcher_spec.ts
```diff
@@ -225,19 +225,25 @@ expect(mySpyObj).toHaveSpyInteractions();`,
       {
         description: 'should transform toEqual(jasmine.arrayWithExactContents()) into two calls',
         input: `expect(myArray).toEqual(jasmine.arrayWithExactContents(['a', 'b']));`,
+        /* eslint-disable max-len */
         expected: `
+          // TODO: vitest-migration: Verify this matches strict array content (multiset equality). Vitest's arrayContaining is a subset check.
           expect(myArray).toHaveLength(2);
           expect(myArray).toEqual(expect.arrayContaining(['a', 'b']));
```
alfaproject, line 232: This is how you do the same in Jest/Vitest:

```suggestion
          expect(myArray).toEqual(['a', 'b']);
```

clydin: That's not equivalent. The following passes but would fail with that transformation:
`expect([1, 2]).toEqual(jasmine.arrayWithExactContents([2, 1]))`

alfaproject: I see, yeah that makes sense

PR #31883, opened 2025-11-20
fix(@angular/ssr): prevent redirect loop with encoded query parameters

### LINKED ISSUE #31881: SSR url normalization causes 302 redirects
### Command

serve

### Is this a regression?

_No response_

### Description

With the introduction of #31497, we noticed higher CPU usage and looked at the logs. We noticed a lot of 302 redirects happening now that will cause 2 requests to be served instead of one. 
Angular SSR is already slow, taking more than a second to answer requests. With the introduction of the redirects, the ssr server now has to compute two requests and end user latency doubles. 

It looks like the internal url normalization of angular will issue redirects, where they're not necessary.

### Minimal Reproduction
```
This url
/?email=xyz%40xyz.com
Will redirect to
/?email=xyz@xyz
```

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 20.3.10
Node: 24.11.0
Package Manager: npm 11.6.1
OS: linux x64
    

Angular: 20.3.12
... platform-server, router

-----------------------------------------
@angular-devkit/architect         0.2003.10
@angular-devkit/core              20.3.10
@angular-devkit/schematics        20.3.10
@angular/build                    20.3.10
@angular/cdk                      20.2.13
@angular/cli                      20.3.10
@angular/material                 20.2.13
@angular/material-luxon-adapter   20.2.13
@angular/ssr                      20.3.10
@angular/youtube-player           20.2.13
@schematics/angular               20.3.10
typescript                        5.9.3
zone.js                           0.15.1
```

### Anything else relevant?

_No response_

alan-agius4:
Previously, encoded query parameters caused a mismatch between the requested URL and the reconstructed URL, leading to an a redirect loop. This change ensures both URLs are decoded before comparison.

Closes #31881

### DIFF
--- a/packages/angular/ssr/src/utils/ng.ts
+++ b/packages/angular/ssr/src/utils/ng.ts
@@ -107,13 +107,15 @@ export async function renderAngular(
 
     if (!routerIsProvided) {
       hasNavigationError = false;
-    } else if (lastSuccessfulNavigation) {
+    } else if (lastSuccessfulNavigation?.finalUrl) {
       hasNavigationError = false;
-      const { pathname, search, hash } = envInjector.get(PlatformLocation);
-      const finalUrl = [stripTrailingSlash(pathname), search, hash].join('');
 
-      if (urlToRender.href !== new URL(finalUrl, urlToRender.origin).href) {
-        redirectTo = finalUrl;
+      const { pathname, search, hash } = envInjector.get(PlatformLocation);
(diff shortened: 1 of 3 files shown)


atscott (approved):

alan-agius4:

- main: 61a027dba7f3a569c1ac936c34956b8f96fd102a
- 21.0.x: 3cac0188271175e12cc238c6610b542f3ae14db3

PR #31933, opened 2025-11-24
fix(@angular/build): ensure external SSR middleware handles requests outside baseHref

### LINKED ISSUE #31896: Angular SSR: Express routes are only prefixed in development with baseHref set
### Command

serve

### Is this a regression?

_No response_

### Description

When setting the `baseHref`, all dev server requests are required to be prefixed, even if I have routes in my `server.ts` that do not need the baseHref, such as health check endpoints.

Then when building the application the endpoint works without the baseHref, causing a mismatch of behavior between dev and build.

### Minimal Reproduction

https://github.com/brandonroberts/angular-ssr-base-href

1. Clone the repo
2. Run npm install
3. Run npm start
4. Visit `http://localhost:4200/ping`
5. Note the error message that the URL should be `http://localhost:4200/custom/ping`
6. Stop the dev server
7. Run npm run build
8. Run `npm run serve:ssr:angular-ssr-base-href`
9. Visit `http://localhost:4000/ping`
10. Note the route displays correctly

### Exception or Error

```text

```

### Your Environment

```text
Angular: 20.3.12

@angular-devkit/architect    0.2003.10
@angular-devkit/core         20.3.10
@angular-devkit/schematics   20.3.10
@angular/build               20.3.10
@angular/cli                 20.3.10
@angular/ssr                 20.3.10
@schematics/angular          20.3.10
```

### Anything else relevant?

Tested with Angular v21 also

alessiopelliccione:
Move the `ExternalSsrMiddleware` registration to execute before Vite's internal middlewares. This fixes an issue where custom server routes (like `/health` or `/api`) were unreachable in development mode if the application used a specific `baseHref`, as Vite would filter them out before the SSR middleware could process them.

Closes #31896

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The `ExternalSsrMiddleware` is currently registered via the returned callback in `configureServer`, placing it after Vite's internal middlewares in the stack.
Consequently, if a `baseHref` is configured (e.g., `/app/`), Vite intercepts requests matching the root or other paths before they reach the custom SSR server.
This prevents custom handlers in `server.ts` (such as `/ping` or `/health`) from processing requests that fall outside the application's base path.

This is causing issues like #31896 

## What is the new behavior?
The `ExternalSsrMiddleware` is registered immediately within the `configureServer` `hook`, ensuring it executes before Vite's internal middlewares. This allows the custom SSR server to intercept and handle requests first, enabling support for routes defined outside the configured `baseHref` while allowing unhandled requests to fall through to Vite.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
+++ b/packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
@@ -105,15 +105,21 @@ export function createAngularSetupMiddlewaresPlugin(
 
       extensionMiddleware?.forEach((middleware) => server.middlewares.use(middleware));
 
+      // For ExternalSsrMiddleware, install the middleware early to ensure custom middleware
+      // in server.ts can handle requests outside the base path (e.g., /ping when baseHref=/custom/)
+      // This allows fallthrough for custom routes that don't start with the configured baseHref
+      if (ssrMode === ServerSsrMode.ExternalSsrMiddleware) {
+        server.middlewares.use(
+          await createAngularSsrExternalMiddleware(server, indexHtmlTransformer),
+        );
+      }
+
       // Returning a function, installs middleware after the main transform middleware but
       // before the built-in HTML middleware
       // eslint-disable-next-line @typescript-eslint/no-misused-promises
       return async () => {
+        // ExternalSsrMiddleware was already installed above, skip here
         if (ssrMode === ServerSsrMode.ExternalSsrMiddleware) {
-          server.middlewares.use(
-            await createAngularSsrExternalMiddleware(server, indexHtmlTransformer),
-          );
-
           return;
         }
 


alan-agius4:
Hi @alessiopelliccione, thanks for your contribution. We need a slightly different approach for this as the current implementation breaks some e2e tests.

alessiopelliccione:
> Hi @alessiopelliccione, thanks for your contribution. We need a slightly different approach for this as the current implementation breaks some e2e tests.

Hi @alan-agius4,

The SSR middleware was incorrectly intercepting Vite's internal requests after the recent reordering. This was causing the failures of E2E tests.

I added locally a check to bypass the middleware for paths like `/@...` and `/__vite...`, which restores correct dev server behavior and fixes the failing E2E pipeline.

I pushed this update https://github.com/alessiopelliccione/angular-cli/commit/309a096c46dd27d87c3561608bf92cfea7a499bf

Tell me if it should be ok now, and I will squash the two commits.

alan-agius4:
Hi @alessiopelliccione, I doubt that it will be enough as with your change all requests will be passed via the middleware, include those for assets in memory, this will cause assets that exist to return a 404 when using hono/fastify.

Example:
```
$ curl localhost:4200/main.js
```

You can try adding a new e2e

**tests/legacy-cli/e2e/tests/vite/ssr-serve-base-href.ts**
```ts
import assert from 'node:assert';
import { writeMultipleFiles } from '../../utils/fs';
import { ng, silentNg } from '../../utils/process';
import { installWorkspacePackages, uninstallPackage } from '../../utils/packages';
import { ngServe, updateJsonFile, useSha } from '../../utils/project';
import { getGlobalVariable } from '../../utils/env';

  assert(
    getGlobalVariable('argv')['esbuild'],
    'This test should not be called in the Webpack suite.',
  );

  // Forcibly remove in case another test doesn't clean itself up.
  await uninstallPackage('@angular/ssr');
  await ng('add', '@angular/ssr', '--skip-confirmation', '--skip-install');
  await useSha();
  await updateJsonFile('angular.json', (workspaceJson) => {
    workspaceJson.projects['test-project'].architect.build.options.baseHref = '/base/';
  });

  await writeMultipleFiles({
    'src/app/app.routes.ts': `
      import { Routes } from '@angular/router';
      import { Home } from './home/home';

      export const routes: Routes = [
        { path: 'home', component: Home }
      ];
    `,
    'src/server.ts': `
      import { AngularNodeAppEngine, writeResponseToNodeResponse, isMainModule, createNodeRequestHandler } from '@angular/ssr/node';
      import express from 'express';

      export function app(): express.Express {
        const server = express();
        const browserDistFolder = join(import.meta.dirname, '../browser');
        const angularNodeAppEngine = new AngularNodeAppEngine();

        server.use('/api/{*splat}', (req, res) => {
          res.json({ hello: 'foo' })
        });

        server.use(express.static(browserDistFolder, {
          maxAge: '1y',
          index: 'index.html'
        }));

        server.use(async(req, res, next) => {
          const response = await angularNodeAppEngine.handle(req);
          if (response) {
            writeResponseToNodeResponse(response, res);
          } else {
            next();
          }
        });

        return server;
      }

      const server = app();
      export const reqHandler = createNodeRequestHandler(server);
  `,
  });

  await silentNg('generate', 'component', 'home');

  const port = await ngServe();

  await validateResponse('/base/main.js', /bootstrapApplication/);
  await validateResponse('/base/home', /home works/);
  await validateResponse('/api/test', /foo/);

  await validate404Response('/base/api/test');
  await validate404Response('/home');
  await validate404Response('/main.js');

  function validate404Response(pathname: string): Promise<void> {
    return validateResponse(pathname, /Cannot GET/, 404);
  }

  async function validateResponse(pathname: string, match: RegExp, status = 200): Promise<void> {
    const response = await fetch(new URL(pathname, `http://localhost:${port}`));
    assert.equal(response.status, status, `Response status for ${pathname} does not match`);

    const text = await response.text();
    assert.match(text, match, `Response body for ${pathname} does not match`);
  }
}

```

And you can run it via
```
pnpm bazel run //tests/legacy-cli:e2e_node22 --config=e2e --test_arg="--glob=tests/vite/ssr-serve-base-href.ts" --test_arg="--esbuild"
```

Or ran and debug one of the failing tests via:
```
pnpm bazel run //tests/legacy-cli:e2e_node22 --config=e2e --test_arg="--glob=tests/vite/ssr-entry-hono.ts" --test_arg="--esbuild"
```

PR #31947, opened 2025-11-25
refactor(@angular/cli): add generic package manifest resolver

clydin:
Extends the package manager abstraction to resolve manifests for various specifier types beyond registry packages. This is crucial for supporting packages installed from local directories, file archives, and git repositories.

A new `getManifest` method is introduced on the `PackageManager` service. This method acts as a unified entry point for manifest resolution:
- It uses `npm-package-arg` to parse the package specifier and determine its type (registry, directory, file, git, etc.).
- For local directory specifiers, it reads the `package.json` directly from the filesystem for maximum efficiency.
- For other non-registry specifiers (git repos, remote tarballs, local tarballs), it uses the `acquireTempPackage` flow to safely fetch and extract the package.
- When acquiring a package, lifecycle scripts are disabled for improved security and performance.
- To reliably determine the package name for non-registry types, the manifest resolver now inspects the temporary `package.json` after the package manager has installed the dependency.

The existing `getPackageManifest` method has been renamed to `getRegistryManifest` to more accurately reflect its purpose.

### DIFF
--- a/packages/angular/cli/src/package-managers/host.ts
+++ b/packages/angular/cli/src/package-managers/host.ts
@@ -15,7 +15,7 @@
 
 import { spawn } from 'node:child_process';
 import { Stats } from 'node:fs';
-import { mkdtemp, readdir, rm, stat, writeFile } from 'node:fs/promises';
+import { mkdtemp, readFile, readdir, rm, stat, writeFile } from 'node:fs/promises';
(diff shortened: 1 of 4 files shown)


alan-agius4:

- main: dabe101d019fbc7b7254b214dcad30caac6071f1

PR #31984, opened 2025-12-01
fix(@schematics/angular): remove `experimentalDecorators` from new pr…

JeanMeche:
…ojects

`experimentalDecorators: true` was required to use constructor DI. With recommendation now to use `inject` for DI, new projects don't need the experimental decorators anymore.


alan-agius4:
NIT: you can now delete `tests/legacy-cli/e2e/tests/build/ts-standard-decorators.ts`

PR #31988, opened 2025-12-01
fix(@angular/build): add filename truncation to test discovery

### LINKED ISSUE #31968: Issue: ENAMETOOLONG during `@angular/build:karma` test builds on macOS
### Command

test

### Description

Hi team 👋,

I am encountering an **ENAMETOOLONG** error when running tests with the `@angular/build:karma` builder.  
This happens because Angular generates extremely long output filenames for test source maps, which include the **full absolute path** of the component under test.

On macOS (and some CI environments), the filesystem limits paths to **255 characters**, and the generated spec map paths exceed this limit.

When the project structure is nested (common in enterprise monorepos), the combined absolute path of the repo + project + feature modules + filename becomes extremely long (300–390+ chars).

This makes the test runner fail with `ENAMETOOLONG` before execution even starts.

### Impact
- Test builds cannot run on macOS  
- CI pipelines fail  
- Refactoring large monorepos becomes extremely difficult  
- Developers must manually shorten folder structures even when logically correct  

---

### **Option 1 — Shorten or hash test output paths**
Emit something like:

instead of embedding the entire directory path.

### **Option 2 — Add a configuration flag (e.g. `"shortPath": true`)**
Something like:

```json
"test": {
  "builder": "@angular/build:karma",
  "options": {
    "shortPath": true
  }
}

clydin:
Adds truncation logic to the `generateNameFromPath` function in the unit test discovery process. This prevents issues with filesystems that have filename length limits.

When a generated test entrypoint name exceeds 128 characters, it is truncated. The truncated name consists of a prefix of the original name, an 8-character SHA256 hash of the full path, and a suffix of the original name, ensuring uniqueness while keeping the name readable.

Unit tests are included to verify the truncation behavior for long filenames.

Closes #31968

### DIFF
--- a/packages/angular/build/src/builders/unit-test/test-discovery.ts
+++ b/packages/angular/build/src/builders/unit-test/test-discovery.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
+import { createHash } from 'node:crypto';
 import { type PathLike, constants, promises as fs } from 'node:fs';
 import os from 'node:os';
 import { basename, dirname, extname, isAbsolute, join, relative } from 'node:path';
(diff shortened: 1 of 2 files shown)


alan-agius4:

- main: 63c3e3f6406d345e777ca18bfad7d6d701e5c4ea
- 21.0.x: f8a1939fdf5abbc1de439d288d9357d4f92a72a9

PR #32029, opened 2025-12-04
refactor(@angular/cli): implement structured error handling in package manager API

clydin:
This commit introduces a structured error handling system into the package manager abstraction. Previously, the system treated most command failures generically, often assuming a package was simply "not found." This could mask the true root cause of issues such as network failures, registry authentication errors, or unexpected changes in a package manager's output.

The new implementation enhances diagnostics and provides clearer, more actionable error messages by:
- Introducing a standardized `ErrorInfo` interface to represent parsed error details.
- Adding dedicated parsers to extract structured error information (JSON or text-based) from the stderr of npm, yarn, pnpm, and bun.
- Making the core execution logic stricter. It now re-throws the original `PackageManagerError` if a command fails for an unknown reason, preventing silent failures.
- Abstracting the "package not found" check into the `PackageManagerDescriptor`, removing hardcoded logic from the execution layer.
- Improving the `yarn-classic` parser to use verbose logs, allowing it to accurately detect and report specific HTTP status codes (e.g., 401, 403, 404, 500).

### DIFF
--- a/packages/angular/cli/src/package-managers/error.ts
+++ b/packages/angular/cli/src/package-managers/error.ts
@@ -35,3 +35,18 @@ export class PackageManagerError extends Error {
     super(message);
   }
 }
+
+/**
+ * Represents structured information about an error returned by a package manager command.
+ * This is a data interface, not an `Error` subclass.
+ */
+export interface ErrorInfo {
+  /** A specific error code (e.g. 'E404', 'EACCES'). */
+  readonly code: string;
+
+  /** A short, human-readable summary of the error. */
+  readonly summary: string;
+
+  /** An optional, detailed description of the error. */
+  readonly detail?: string;
+}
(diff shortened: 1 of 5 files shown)


packages/angular/cli/src/package-managers/error.ts
```diff
@@ -35,3 +35,18 @@ export class PackageManagerError extends Error {
...
   }
 }
+
+/**
+ * Represents structured information about an error returned by a package manager command.
+ * This is a data interface, not an `Error` subclass.
+ */
+export interface ErrorInfo {
+  /** A specific error code (e.g. 'E404', 'EACCES'). */
+  readonly code: string;
```
hybrist, line 45: Is it worth to make this narrower than `string` to make it clearer what _kind_ of error codes to expect?

clydin: They don't seem to be particularly consistent, unfortunately.  Sometimes its a number, an HTTP status code, a Node.js error code, or even a package manager specific one.

clydin:

- main: 69b2be851321ae06001f02135b5a39790b3543db

PR #32033, opened 2025-12-05
fix(@angular/cli): correctly spawn package managers on Windows in new abstraction

clydin:
When running E2E tests on Windows, `spawn` was failing to find package manager executables like `npm` because it does not automatically resolve `.cmd` or `.bat` extensions when `shell: false` is used.

Good CI run: https://github.com/angular/angular-cli/actions/runs/19951164463/job/57211327517#step:5:428


alan-agius4:

- main: 7ab5c0b0a1c637f3e0adb71486e5e7e8716561e4

PR #32040, opened 2025-12-05
fix(@angular/build): ensure locale base href retains leading slash

### LINKED ISSUE #32030: Leading slash added to baseHref
### Command

build

### Is this a regression?

20.3.11

### Description

**Expected behavior**

The value of the `href` attribute of the `<base>` tag in the built `index.html` file should start with what's defined in the `baseHref` property of the builder.

**Actual behavior**

A leading `/` is added when the `baseHref` value does not start with one.

**Impact**

We package our built apps inside docker images that process the `index.html` file on startup to replace certain strings matching `${...}` with values from environment variables.  This allows us to deliver portable docker images that can be deployed under different paths in different environments by simply setting the `BASE_HREF` environment variable.

So when building our app, we set `baseHref` in the build config to a string like `${BASE_HREF}`.

With Angular 20.3.11, this produced the following tag in `index.html` (with a localized build):

```html
<base href="${BASE_HREF}/en/"/>
```

Since Angular 20.3.13, we now get a leading slash added:

```html
<base href="/${BASE_HREF}/en/"/>
```

For deployments where our `BASE_HREF` value is an empty string, this results in the base href value of `//en/` which is a protocol-relative URL and so all of our resources fail to load.

### Minimal Reproduction

Create an Angular app with 20.3.13, set baseHref to a value that does not start with `/`, enable localization, build.

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI: 20.3.13
Node: 22.14.0
Package Manager: npm 10.9.2
OS: darwin arm64
    

Angular: 20.3.15
... language-service, localize, platform-browser

@angular-devkit/architect    0.2003.13
@angular-devkit/core         20.3.13
@angular-devkit/schematics   20.3.13
@angular/build               20.3.13
@angular/cdk                 20.2.14
@angular/cli                 20.3.13
@schematics/angular          20.3.13
rxjs                         7.8.1
```

### Anything else relevant?

_No response_

alan-agius4:
When baseHref is provided, getLocaleBaseHref was incorrectly adding a leading slash from the joined path. This commit removes calls `stripLeadingSlash` to ensure the leading slash is removed when appropriate.

Closes #32030

### DIFF
--- a/packages/angular/build/src/builders/application/options.ts
+++ b/packages/angular/build/src/builders/application/options.ts
@@ -25,7 +25,7 @@ import {
   loadPostcssConfiguration,
 } from '../../utils/postcss-configuration';
 import { getProjectRootPaths, normalizeDirectoryPath } from '../../utils/project-metadata';
-import { addTrailingSlash, joinUrlParts } from '../../utils/url';
+import { addTrailingSlash, joinUrlParts, stripLeadingSlash } from '../../utils/url';
 import {
   Schema as ApplicationBuilderOptions,
   ExperimentalPlatform,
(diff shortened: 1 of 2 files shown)


clydin (approved):

alan-agius4:

- main: 38b16ea0108c48835dc0d81863eca84f7b8cea6e
- 21.0.x: 0fc7d576e53f45601fdbeb95f4a853ebceae4fad

PR #32056, opened 2025-12-06
Tailored error message for custom host

### LINKED ISSUE #32028: Serve Angular Application with ngrok
### Describe the problem that you experienced

I was using ngrok to serve an Angular app in a remote link but without the configuration in the Angular.json I get an error related to vite configurations.

### Enter the URL of the topic with the problem

_No response_

### Describe what you were looking for in the documentation

I don't know if there's a section in the documentation dedicated to this scenario. If so, could we consider adding it? Eventually I could open a PR

### Describe the actions that led you to experience the problem

_No response_

### Describe what you want to experience that would fix the problem

The error message generated, in my opinion, is inconsistent with the change to be made, because a vite.config.js file is typically missing. The solution is to edit angular.json by adding the URL generated by ngrok to the allowedHosts array.

### Add a screenshot if that helps illustrate the problem

_No response_

### If this problem caused an exception or error, please paste it here

```true
Blocked request. This host ("2759795c2071.ngrok-free.app") is not allowed.
To allow this host, add "2759795c2071.ngrok-free.app" to `server.allowedHosts` in vite.config.js.
```

### If the problem is browser-specific, please specify the device, OS, browser, and version

```true

```

### Provide any additional information here in as much as detail as you can

```true

```

aparzi:
Add custom middleware for to present an Angular-tailored message

Fix #32028

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The error message thrown when the Angular app is served via a custom hostname is incorrect/misleading

Issue Number: #32028

## What is the new behavior?

Add custom middleware for to present an Angular-tailored message when angular served by ngrok

- [ ] Yes
- [X] No

### DIFF
--- a/packages/angular/build/src/builders/dev-server/tests/options/allowed-hosts_spec.ts
+++ b/packages/angular/build/src/builders/dev-server/tests/options/allowed-hosts_spec.ts
@@ -34,7 +34,7 @@ describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupT
 
       expect(result?.success).toBeTrue();
       expect(response?.statusCode).toBe(403);
-      expect(response && text(response)).toContain('angular.json');
+      expect(response && (await text(response))).toContain('angular.json');
     });
 
     it('does not allow an invalid host when option is an empty array', async () => {
@@ -49,7 +49,7 @@ describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupT
 
       expect(result?.success).toBeTrue();
       expect(response?.statusCode).toBe(403);
-      expect(response && text(response)).toContain('angular.json');
+      expect(response && (await text(response))).toContain('angular.json');
     });
 
     it('allows a host when specified in the option', async () => {


packages/angular/build/src/tools/vite/middlewares/host-check-middleware.ts
```diff
@@ -0,0 +1,116 @@
+/**
```
alan-agius4, line 1: This seems to change the behaviour for a number of valid cases. 

Example: 
- `http://foo.localhost` will now be marked as invalid
-  When adding allowed hosts, all sub-domains will also be marked as valid, but with this change they will be marked as invalid. 
- Accessing the app with an external IP, will now be marked as invalid.


My suggestion would be that instead of recreating the middleware, just replace the response.

```ts

const middlewareStack = server.middlewares.stack;
const idx = middlewareStack.findIndex(({handle}) =>
  handle.name.startsWith('hostValidationMiddleware'),
);

if (idx !== -1) {
  const middleware = middlewareStack[idx];
  const originalHandle = middleware.handle;
  middleware.handle = function angularHostValidationMiddleware(req, res, next) {
    originalHandle(
      req,
      {
        writeHead: res.writeHead.bind(res),
        end: () => res.end('Error message.'),
      },
      next,
    );
  };
}

```

packages/angular/build/src/tools/vite/middlewares/host-check-middleware.ts
```diff
@@ -0,0 +1,116 @@
...
+      h1{font-size:1.5rem;margin-bottom:.75rem}
+      p{margin:.5rem 0}
+      .muted{color:#6b7280}
+      pre{background:#f9fafb;border:1px solid #e5e7eb;padding:.75rem;border-radius:.5rem;overflow:auto}
+    </style>
+  </head>
+  <body>
+    <div class="box">
+      <h1>Blocked request. This host ("${hostname}") is not allowed.</h1>
+      <p>The Angular development server only responds to local hosts by default.</p>
```
alan-agius4: This is not actually fully true, as you can get a response via the external IP address.

```suggestion
```

packages/angular/build/src/tools/vite/middlewares/host-check-middleware.ts
```diff
@@ -0,0 +1,116 @@
...
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Blocked request</title>
+    <style>
+      body{font-family:system-ui,-apple-system,Segoe UI,Roboto,Ubuntu,Cantarell,Noto Sans,sans-serif;line-height:1.4;margin:2rem;color:#1f2937}
+      code{background:#f3f4f6;padding:.15rem .35rem;border-radius:.25rem}
+      .box{max-width:760px;margin:0 auto}
+      h1{font-size:1.5rem;margin-bottom:.75rem}
+      p{margin:.5rem 0}
+      .muted{color:#6b7280}
```
alan-agius4: Unused
```suggestion

```

packages/angular/build/src/tools/vite/middlewares/host-check-middleware.ts
```diff
@@ -0,0 +1,116 @@
...
+      pre{background:#f9fafb;border:1px solid #e5e7eb;padding:.75rem;border-radius:.5rem;overflow:auto}
+    </style>
+  </head>
+  <body>
+    <div class="box">
```
alan-agius4: Optional NIT

```suggestion
    <main>
```

packages/angular/build/src/tools/vite/middlewares/host-check-middleware.ts
aparzi: I deleted the custom middleware and updated the error message trying to do as you suggested

packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
```diff
@@ -55,6 +55,57 @@ interface AngularSetupMiddlewaresPluginOptions {
...
+
+  // Remove port if present (e.g., example.com:4200)
+  const idx = hostHeader.lastIndexOf(':');
+  if (idx > -1 && hostHeader.indexOf(']') === -1) {
+    // Skip IPv6 addresses that include ':' within brackets
+    return hostHeader.slice(0, idx).toLowerCase();
+  }
+
+  return hostHeader.toLowerCase();
+}
```
alan-agius4: This cannot be IPV6 if the hostname is IPV6 it will not error.

I suggest to remove the function altogether and do

```suggestion
const hostname = req.headers.host?.split(':')[0].toLowerCase() ?? ''
```

packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
```diff
@@ -109,6 +160,88 @@ export function createAngularSetupMiddlewaresPlugin(
...
+        const middlewareStack = mws.stack;
+
+        // Helper to locate Vite's host validation middleware safely
+        const findHostValidationEntry = (
+          stack: readonly MiddlewareStackEntry[],
+        ): MiddlewareStackEntry | undefined =>
+          stack.find(
+            ({ handle }) =>
+              typeof handle === 'function' &&
+              typeof handle.name === 'string' &&
```
alan-agius4: redundant check

```suggestion
  
```

packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
```diff
@@ -109,6 +160,88 @@ export function createAngularSetupMiddlewaresPlugin(
...
+          route?: string;
+          handle?: (Connect.NextHandleFunction & { name?: string }) | undefined;
+        };
+
+        // Vite/Connect do not expose a typed stack, cast once to a precise structural type.
+        const mws = server.middlewares as unknown as { stack: MiddlewareStackEntry[] };
+        const middlewareStack = mws.stack;
+
+        // Helper to locate Vite's host validation middleware safely
+        const findHostValidationEntry = (
```
alan-agius4: NIT: you can just chain `middlewareStack.find` IMHO it's cleaner and more readable.

packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
```diff
@@ -109,6 +160,88 @@ export function createAngularSetupMiddlewaresPlugin(
...
+          stack.find(
+            ({ handle }) =>
+              typeof handle === 'function' &&
+              typeof handle.name === 'string' &&
+              handle.name.startsWith('hostValidationMiddleware'),
+          );
+
+        const entry = findHostValidationEntry(middlewareStack);
+
+        if (entry && typeof entry.handle === 'function') {
```
alan-agius4:
```suggestion
        if (typeof entry?.handle === 'function') {
```

packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
```diff
@@ -55,6 +55,57 @@ interface AngularSetupMiddlewaresPluginOptions {
...
+  const idx = hostHeader.lastIndexOf(':');
+  if (idx > -1 && hostHeader.indexOf(']') === -1) {
+    // Skip IPv6 addresses that include ':' within brackets
+    return hostHeader.slice(0, idx).toLowerCase();
+  }
+
+  return hostHeader.toLowerCase();
+}
+
+function html403(hostname: string): string {
```
alan-agius4: Optional NIT: Would it be possible to move this in a new file? (See other middlewares)

packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
```diff
@@ -109,6 +160,88 @@ export function createAngularSetupMiddlewaresPlugin(
       // before the built-in HTML middleware
       // eslint-disable-next-line @typescript-eslint/no-misused-promises
       return async () => {
+        type MiddlewareStackEntry = {
```
alan-agius4: I am a bit confused by the need for all this logic. Can you explain a bit why these are needed? I checkout out you PR locally and removed all the patching for `writeHead` and `write` etc.. and achieved the same thing.

```ts
const entry = server.middlewares.stack.find(
  ({ handle }) =>
    typeof handle === 'function' && handle.name.startsWith('hostValidationMiddleware'),
);

if (typeof entry?.handle === 'function') {
  const originalHandle = entry.handle as Connect.NextHandleFunction;

  entry.handle = function angularHostValidationMiddleware(
    req: IncomingMessage,
    res: ServerResponse,
    next: (err?: unknown) => void,
  ) {
    originalHandle(
      req,
      // The middleware only uses this part of the response.
      // If it happens to change and requires more. It will be catched by tests.
      {
        writeHead: res.writeHead.bind(res),
        end: () => {
          const hostname = req.headers.host?.toLowerCase().split(':')[0] ?? '';
          res.end(html403(hostname));
        },
      } as ServerResponse,
      next,
    );
  };
}
```

packages/angular/build/src/builders/dev-server/tests/options/allowed-hosts_spec.ts
```diff
@@ -27,12 +27,13 @@ describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupT
...
       });
 
-      const { result, response } = await executeOnceAndGet(harness, '/', {
+      const { result, response, content } = await executeOnceAndGet(harness, '/', {
         request: { headers: FETCH_HEADERS },
       });
 
       expect(result?.success).toBeTrue();
       expect(response?.statusCode).toBe(403);
+      expect(content).toContain('angular.json');
```
alan-agius4:
```suggestion

import { text } from 'node:stream/consumers';


expect(response && text(response)).toContain('angular.json');
```

packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
aparzi: Is there a specific middleware that I could use as an example? I found `assets-middleware.ts`, but the HTML code is inside.

Thanks

alan-agius4: I meant move the 2 functions inside a seperate files.

example: https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/tools/vite/middlewares/headers-middleware.ts

packages/angular/build/src/tools/vite/plugins/setup-middlewares-plugin.ts
aparzi: This way, it appears that the response header's content type is text/plain, so I don't see the HTML formatting.
From what I can gather, I need to set the headers, but I get this error:
`Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client`

Because the response is already returned, I've complicated the logic.

alan-agius4: In that case, you can simply update the `writeHead`
```ts
import assert from 'node:assert';

writeHead: (code, headers = {}) => {
  assert(typeof headers === 'object' && !Array.isArray(headers), 'headers must be a object');
  headers['content-type'] = 'text/html';
  res.writeHead(code, headers);
},
```

or even 

```ts
writeHead: (code) => {
   res.writeHead(code, { "content-type": 'text/html'})
},
```

hybrist:

- main: 30b5d81b4adafca32c94672a39574daa2e3fc8b7
- 21.0.x: 778b4cffc03e7c137940e3b8c89f290fd226cf17

PR #32066, opened 2025-12-08
feat(@angular/cli): add "test" and "e2e" MCP tools, and some test cleanup

amishne:
1. Add "test" and "e2e" MCP tools, with their tests.
2. Add an "all" experimental tool group so it's easy to enable them all at once.
3. Make the default project use the actual project name instead of the "<default>" string.
4. Test cleanup, including centralizing test-only utilities into a new file.

In addition to the unit tests, the test tool was also tested manually installed into Gemini-CLI.

### DIFF
--- a/packages/angular/cli/src/commands/mcp/testing/test-utils.ts
+++ b/packages/angular/cli/src/commands/mcp/testing/test-utils.ts
@@ -1,0 +1,91 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import { workspaces } from '@angular-devkit/core';
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { AngularWorkspace } from '../../../utilities/config';
+import { type Devserver } from '../devserver';
+import { Host } from '../host';
+import { McpToolContext } from '../tools/tool-registry';
+import { MockHost } from './mock-host';
+
+/**
+ * Creates a mock implementation of the Host interface for testing purposes.
+ * Each method is a Jasmine spy that can be configured.
+ */
+export function createMockHost(): MockHost {
+  return {
+    runCommand: jasmine.createSpy<Host['runCommand']>('runCommand').and.resolveTo({ logs: [] }),
+    stat: jasmine.createSpy<Host['stat']>('stat'),
+    existsSync: jasmine.createSpy<Host['existsSync']>('existsSync'),
+    spawn: jasmine.createSpy<Host['spawn']>('spawn'),
+    getAvailablePort: jasmine
+      .createSpy<Host['getAvailablePort']>('getAvailablePort')
+      .and.resolveTo(0),
+  } as unknown as MockHost;
+}
+
+/**
+ * Options for configuring the mock MCP tool context.
+ */
+export interface MockContextOptions {
+  /** An optional pre-configured mock host. If not provided, a default mock host will be created. */
+  host?: MockHost;
+
+  /** Initial set of projects to populate the mock workspace with. */
+  projects?: Record<string, workspaces.ProjectDefinition>;
+}
+
+/**
+ * Creates a comprehensive mock for the McpToolContext, including a mock Host,
+ * an AngularWorkspace, and a ProjectDefinitionCollection. This simplifies testing
+ * MCP tools by providing a consistent and configurable testing environment.
+ * @param options Configuration options for the mock context.
+ * @returns An object containing the mock host, context, projects collection, and workspace instance.
+ */
+export function createMockContext(options: MockContextOptions = {}): {
+  host: MockHost;
+  context: McpToolContext;
+  projects: workspaces.ProjectDefinitionCollection;
+  workspace: AngularWorkspace;
+} {
+  const host = options.host ?? createMockHost();
+  const projects = new workspaces.ProjectDefinitionCollection(options.projects);
+  const workspace = new AngularWorkspace({ projects, extensions: {} }, '/test/angular.json');
+
+  const context: McpToolContext = {
+    server: {} as unknown as McpServer,
+    workspace,
+    logger: { warn: () => {} },
+    devservers: new Map<string, Devserver>(),
+    host,
+  };
+
+  return { host, context, projects, workspace };
+}
+
+/**
+ * Adds a project to the provided mock ProjectDefinitionCollection.
+ * This is a helper function to easily populate a mock Angular workspace.
+ * @param projects The ProjectDefinitionCollection to add the project to.
+ * @param name The name of the project.
+ * @param targets A record of target definitions for the project (e.g., build, test, e2e).
+ * @param root The root path of the project, relative to the workspace root. Defaults to `projects/${name}`.
+ */
+export function addProjectToWorkspace(
+  projects: workspaces.ProjectDefinitionCollection,
+  name: string,
+  targets: Record<string, workspaces.TargetDefinition> = {},
+  root = `projects/${name}`,
+) {
+  projects.set(name, {
+    root,
+    extensions: {},
+    targets: new workspaces.TargetDefinitionCollection(targets),
+  });
+}
--- a/packages/angular/cli/src/commands/mcp/tools/devserver/devserver-stop.ts
+++ b/packages/angular/cli/src/commands/mcp/tools/devserver/devserver-stop.ts
@@ -7,8 +7,7 @@
  */
 
 import { z } from 'zod';
-import { devserverKey } from '../../devserver';
-import { createStructuredContentOutput } from '../../utils';
+import { createStructuredContentOutput, getDefaultProjectName } from '../../utils';
 import { type McpToolContext, type McpToolDeclaration, declareTool } from '../tool-registry';
 
 const devserverStopToolInputSchema = z.object({
@@ -30,21 +29,41 @@ const devserverStopToolOutputSchema = z.object({
 export type DevserverStopToolOutput = z.infer<typeof devserverStopToolOutputSchema>;
 
 export function stopDevserver(input: DevserverStopToolInput, context: McpToolContext) {
-  const projectKey = devserverKey(input.project);
-  const devServer = context.devservers.get(projectKey);
+  if (context.devservers.size === 0) {
+    return createStructuredContentOutput({
+      message: ['No development servers are currently running.'],
+      logs: undefined,
+    });
+  }
+
+  let projectName = input.project ?? getDefaultProjectName(context);
+
+  if (!projectName) {
+    // This should not happen. But if there's just a single running devserver, stop it.
+    if (context.devservers.size === 1) {
+      projectName = Array.from(context.devservers.keys())[0];
+    } else {
+      return createStructuredContentOutput({
+        message: ['Project name not provided, and no default project found.'],
+        logs: undefined,
+      });
+    }
+  }
+
+  const devServer = context.devservers.get(projectName);
 
   if (!devServer) {
     return createStructuredContentOutput({
-      message: `Development server for project '${projectKey}' was not running.`,
+      message: `Development server for project '${projectName}' was not running.`,
       logs: undefined,
     });
   }
 
   devServer.stop();
-  context.devservers.delete(projectKey);
+  context.devservers.delete(projectName);
 
   return createStructuredContentOutput({
-    message: `Development server for project '${projectKey}' stopped.`,
+    message: `Development server for project '${projectName}' stopped.`,
     logs: devServer.getServerLogs(),
   });
 }
--- a/packages/angular/cli/src/commands/mcp/tools/e2e.ts
+++ b/packages/angular/cli/src/commands/mcp/tools/e2e.ts
@@ -8,7 +8,12 @@
 
 import { z } from 'zod';
 import { CommandError, type Host, LocalWorkspaceHost } from '../host';
-import { createStructuredContentOutput } from '../utils';
+import {
+  createStructuredContentOutput,
+  getCommandErrorLogs,
+  getDefaultProjectName,
+  getProject,
+} from '../utils';
 import { type McpToolContext, type McpToolDeclaration, declareTool } from './tool-registry';
 
 const e2eStatusSchema = z.enum(['success', 'failure']);
@@ -33,32 +38,19 @@ const e2eToolOutputSchema = z.object({
 export type E2eToolOutput = z.infer<typeof e2eToolOutputSchema>;
 
 export async function runE2e(input: E2eToolInput, host: Host, context: McpToolContext) {
-  const projectName = input.project;
+  const projectName = input.project ?? getDefaultProjectName(context);
 
-  if (context.workspace) {
-    let targetProject;
-    const projects = context.workspace.projects;
-
-    if (projectName) {
-      targetProject = projects.get(projectName);
-    } else {
-      // Try to find default project
-      const defaultProjectName = context.workspace.extensions['defaultProject'] as
-        | string
-        | undefined;
-      if (defaultProjectName) {
-        targetProject = projects.get(defaultProjectName);
-      } else if (projects.size === 1) {
-        targetProject = Array.from(projects.values())[0];
-      }
-    }
-
+  if (context.workspace && projectName) {
+    // Verify that if a project can be found, it has an e2e testing already set up.
+    const targetProject = getProject(context, projectName);
     if (targetProject) {
       if (!targetProject.targets.has('e2e')) {
         return createStructuredContentOutput({
           status: 'failure',
           logs: [
-            `No e2e target is defined for project '${projectName ?? 'default'}'. Please setup e2e testing first.`,
+            `No e2e target is defined for project '${projectName}'. Please set up e2e testing` +
+              ' first by calling `ng e2e` in an interactive console.' +
+              ' See https://angular.dev/tools/cli/end-to-end.',
           ],
         });
       }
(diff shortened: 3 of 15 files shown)


packages/angular/cli/src/commands/mcp/tools/e2e.ts
```diff
@@ -0,0 +1,121 @@
...
+        targetProject = Array.from(projects.values())[0];
+      }
+    }
+
+    if (targetProject) {
+      if (!targetProject.targets.has('e2e')) {
+        return createStructuredContentOutput({
+          status: 'failure',
+          logs: [
+            `No e2e target is defined for project '${projectName ?? 'default'}'. Please setup e2e testing first.`,
```
dgp1130: Consider: We should either link to [documentation](https://angular.dev/tools/cli/end-to-end) or call out direct usage of the `ng e2e` command as a mechanism to configure these tests.

Also typo: I think "setup" is a noun, "set up" would be using it as a verb.

packages/angular/cli/src/commands/mcp/tools/e2e.ts
```diff
@@ -0,0 +1,121 @@
...
+  } catch (e) {
+    status = 'failure';
+    if (e instanceof CommandError) {
+      logs = e.logs;
+    } else if (e instanceof Error) {
+      logs = [e.message];
+    } else {
+      logs = [String(e)];
+    }
+  }
```
dgp1130, line 74: Consider: Can this be refactored to a shared primitive somewhere? Seems like it would overlap pretty heavily with other executions of the CLI in the MCP server.

packages/angular/cli/src/commands/mcp/tools/e2e.ts
```diff
@@ -0,0 +1,121 @@
...
+  const args = ['e2e'];
+  if (input.project) {
+    args.push(input.project);
+  }
+
+  let status: E2eStatus = 'success';
+  let logs: string[] = [];
+
+  try {
+    logs = (await host.runCommand('ng', args)).logs;
```
dgp1130, line 70: Question: I probably should have asked this on an earlier PR (maybe I did and forgot), but how is `ng` resolved? Is that looking for a global installation on the `$PATH`? Is it resolving within the project's `node_modules` somehow? I'm wondering if we want to call `process.argv[0]` to reuse the existing Angular CLI binary rather than attempting to find it again in the environment?

packages/angular/cli/src/commands/mcp/tools/e2e_spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    addProject('my-app', { build: { builder: 'mock-builder' } });
+
+    const { structuredContent } = await runE2e({ project: 'my-app' }, mockHost, mockContext);
+
+    expect(structuredContent.status).toBe('failure');
+    expect(structuredContent.logs?.[0]).toContain("No e2e target is defined for project 'my-app'");
+    expect(mockHost.runCommand).not.toHaveBeenCalled();
+  });
+
+  it('should error if default project does not have e2e target and no project specified', async () => {
```
dgp1130: Suggestion: A test which executes `ng e2e` successfully with the default project. I only see the error case here.

packages/angular/cli/src/commands/mcp/tools/test.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export async function runTest(input: TestToolInput, host: Host) {
+  // Build "ng"'s command line.
+  const args = ['test'];
+  if (input.project) {
+    args.push(input.project);
+  }
+
+  // This is ran by the agent so we want a non-watched, headless test.
+  args.push('--browsers', 'ChromeHeadless');
```
dgp1130, line 42: Consider: What if someone doesn't have Chrome installed and instead uses Firefox or Safari?

@clydin is there a way to trigger "headless" without specifying a specific browser? The alternative I can think of is setting the `CI=true` environment variable.

packages/angular/cli/src/commands/mcp/tools/e2e.ts
```diff
@@ -0,0 +1,121 @@
...
+        | undefined;
+      if (defaultProjectName) {
+        targetProject = projects.get(defaultProjectName);
+      } else if (projects.size === 1) {
+        targetProject = Array.from(projects.values())[0];
+      }
+    }
+
+    if (targetProject) {
+      if (!targetProject.targets.has('e2e')) {
```
dgp1130, line 47: Consider: Refactor to some kind of `validateE2eAlreadySetUp` helper since we shouldn't need any of this data to actually run E2E tests and can simplify `runE2e`.

packages/angular/cli/src/commands/mcp/tools/test.ts
```diff
@@ -0,0 +1,97 @@
...
+  description: `
+<Purpose>
+Perform a one-off, non-watched unit test execution with ng test.
+</Purpose>
+<Use Cases>
+* Running unit tests for the project.
+* Verifying code changes with tests.
+</Use Cases>
+<Operational Notes>
+* This tool runs "ng test" with "--watch false".
```
dgp1130: Consider: This line also feels like an implementation detail. Does the AI agent care if it's going through the MCP server to run these tests?

packages/angular/cli/src/commands/mcp/tools/e2e.ts
```diff
@@ -0,0 +1,121 @@
...
+  title: 'E2E Tool',
+  description: `
+<Purpose>
+Perform an end-to-end test with ng e2e.
+</Purpose>
+<Use Cases>
+* Running end-to-end tests for the project.
+</Use Cases>
+<Operational Notes>
+* This tool runs "ng e2e".
```
dgp1130: Consider: This line feels like an implementation detail. Does the AI agent care if it's going through the MCP server?

packages/angular/cli/src/commands/mcp/tools/e2e.ts
```diff
@@ -0,0 +1,121 @@
...
+> = declareTool({
+  name: 'e2e',
+  title: 'E2E Tool',
+  description: `
+<Purpose>
+Perform an end-to-end test with ng e2e.
+</Purpose>
+<Use Cases>
+* Running end-to-end tests for the project.
+</Use Cases>
```
dgp1130, line 97: Consider: Is there any more context we can give to use cases which would be useful to agents? I'm not sure how helpful declaring the `e2e` tool as "Running end-to-end tests for the project" really is (OTOH maybe it is useful to reinforce agent behavior?). Is there something higher-level we want to say about validating code changes or some distinction between unit and e2e tests?

packages/angular/cli/src/commands/mcp/tools/devserver/devserver-stop.ts
```diff
@@ -30,21 +29,41 @@ const devserverStopToolOutputSchema = z.object({
...
+    return createStructuredContentOutput({
+      message: ['No development servers are currently running.'],
+      logs: undefined,
+    });
+  }
+
+  let projectName = input.project ?? getDefaultProjectName(context);
+
+  if (!projectName) {
+    // This should not happen. But if there's just a single running devserver, stop it.
```
dgp1130, line 42: Consider: If this shouldn't happen, do we want to just `throw` and escalate the issue? I'd worry a bit that this might mask a legitimate bug elsewhere in the program.

Based on the error message below, it seems weird that it would be ignored if there happens to be just one devserver running.

packages/angular/cli/src/commands/mcp/utils.ts
```diff
@@ -52,3 +54,44 @@ export function findAngularJsonDir(startDir: string, host = LocalWorkspaceHost):
...
+
+  const defaultProjectName = context.workspace?.extensions['defaultProject'] as string | undefined;
+  if (defaultProjectName) {
+    return defaultProjectName;
+  }
+
+  // No default project defined? This might still be salvageable if only a single project exists.
+  if (projects.size === 1) {
+    return Array.from(projects.keys())[0];
+  }
```
dgp1130, line 94: Consider: What is the behavior of `ng build` / `ng test` / etc. if `defaultProject` isn't configured but only one project exists? I think it would be good to align with that behavior, even if it is to emit an error.

packages/angular/cli/src/commands/mcp/utils.ts
dgp1130, line 1: Consider: Add tests for these utilities directly. It can potentially reduce the need for some test coverage in the MCP tools as well.

packages/angular/cli/src/commands/mcp/testing/test-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+export function createMockHost(): MockHost {
+  return {
+    runCommand: jasmine.createSpy<Host['runCommand']>('runCommand').and.resolveTo({ logs: [] }),
+    stat: jasmine.createSpy<Host['stat']>('stat'),
+    existsSync: jasmine.createSpy<Host['existsSync']>('existsSync'),
+    spawn: jasmine.createSpy<Host['spawn']>('spawn'),
+    getAvailablePort: jasmine
+      .createSpy<Host['getAvailablePort']>('getAvailablePort')
+      .and.resolveTo(0),
```
dgp1130, line 29: Consider: Could we put the auto-incrementing port number implementation here and potentially reuse it across all affected tests?

packages/angular/cli/src/commands/mcp/tools/e2e_spec.ts
```diff
@@ -20,44 +21,27 @@ describe('E2E Tool', () => {
...
-
-    mockWorkspace = new AngularWorkspace(mockWorkspaceDefinition, '/test/angular.json');
-    mockContext = {
-      workspace: mockWorkspace,
-    } as McpToolContext;
+    const mock = createMockContext();
+    mockHost = mock.host;
+    mockContext = mock.context;
+    mockProjects = mock.projects;
+    mockWorkspace = mock.workspace;
```
dgp1130, line 28: Consider: If we can make this mock setup easier (potentially with the fake implementation proposed in a separate comment) we can move this out of the `beforeEach` and into each test, so each one can get a unique environment for its needs rather than attempting to share and then retroactively configure a base environment.

packages/angular/cli/src/commands/mcp/testing/test-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+import { type Devserver } from '../devserver';
+import { Host } from '../host';
+import { McpToolContext } from '../tools/tool-registry';
+import { MockHost } from './mock-host';
+
+/**
+ * Creates a mock implementation of the Host interface for testing purposes.
+ * Each method is a Jasmine spy that can be configured.
+ */
+export function createMockHost(): MockHost {
```
dgp1130, line 21: Consider: Is there a path to a "fake" host instead of a "mock" host? Could we provide fake implementations for `stat` and `existsSync` if we accept a virtual file system (or create a real one in a temp directly from inputs)?

Faking implementation of `runCommand` and `spawn` is likely a bit harder, but we could either delegate the command to an input function or actually run a real subprocess to get more accurate error behavior. I'm thinking something along the lines of:

```typescript
import * as fs from 'fs/promises';

class FakeHost implements Host {
  private nextPort = 0;

  private constructor(private readonly cmdImpls: Record<string, (...args: string[]) => CommandResult>, private readonly root: string) {}

  static async from(cmdImpls: Record<string, (...args: string[]) => CommandResult>, fs: Record<string, string | Buffer> = {}): Promise<FakeHost> {
    const tempDir = await fs.mkdtemp();

    for (const [path, content] of Object.entries(fs)) {
      await fs.writeFile(path, content);
    }

    return new FakeHost(cmdImpls, tempDir);
  }

  stat() { /* Check in `root` */ }
  existsSync() { /* Check in `root` */ }

  getAvailablePort(): number {
    return this.nextPort++;
  }

  runCommand(binary: string, args: string[]): /* ... */ {
    return this.cmdImpls[binary](args);
  }
  spawn() { /* Similar to `runCommand`... */ }

  await dispose(): Promise<void> {
    await fs.rmdir(this.root);
  }
}
```

I'm thinking this can avoid the need to define spies unique to each test and increase overall test confidence. If we can make this specific to each test as suggested in a separate comment, it can potentially remove the need to mutate the environment after the `beforeEach` within a test.

Downside is that we do need to explicitly call `dispose` to clean up the temp directory if we're using a real file system. If we can run these tests on Node 24, we can potentially get away with [`using` and `Symbol.dispose`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/using) which would help.

```typescript
it('does a thing', () => {
  await using host = await FakeHost.from(...);
  await runE2e(host, ...);
  // Host implicitly disposed at end of scope.
});
```

packages/angular/cli/src/commands/mcp/tools/e2e_spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    expect(structuredContent.logs?.[0]).toContain("No e2e target is defined for project 'default'");
+    expect(mockHost.runCommand).not.toHaveBeenCalled();
+  });
+
+  it('should proceed if no workspace context is available (fallback)', async () => {
+    // If context.workspace is undefined, it should try to run ng e2e (which might fail or prompt, but tool runs it)
+    const noWorkspaceContext = {} as McpToolContext;
+    await runE2e({}, mockHost, noWorkspaceContext);
+    expect(mockHost.runCommand).toHaveBeenCalledWith('ng', ['e2e']);
+  });
```
dgp1130, line 69: Question: This is the case where we run the MCP server outside of a specific Angular CLI project right? I wonder if we should just fail in that case rather than attempting to run an `ng e2e` command which will almost certainly fail.

Thoughts @clydin?

packages/angular/cli/src/commands/mcp/tools/e2e.ts
amishne: Done

packages/angular/cli/src/commands/mcp/tools/e2e.ts
amishne: Done.

packages/angular/cli/src/commands/mcp/tools/e2e.ts
amishne: I remembered it was discussed and we decided relying on the env is a good approach (e.g. if you installed a local one it would use it). Reusing `process.argv[0]` might be even better... but in any case, deferring this to a future change, since it's not specific to the tools in this PR.

packages/angular/cli/src/commands/mcp/tools/e2e_spec.ts
amishne: Done

packages/angular/cli/src/commands/mcp/tools/test.ts
amishne: That's an important comment, but for now I've just updated the tool's description to let the agent know that chrome is required.

packages/angular/cli/src/commands/mcp/tools/e2e.ts
amishne: The `createStructuredContentOutput` call really belongs here anyway so moving this somewhere else will in practice only move very few lines around... I did add a brief comment to explain this section.

packages/angular/cli/src/commands/mcp/tools/test.ts
amishne: I removed the `--watch` part, but here and in `test.ts`, left the basic "this tool uses X". The reason is that I want it to be clear to the agent that it doesn't have to run `ng test` or `ng e2e` independently - two command lines it would likely be familiar too if it has Angular context.

packages/angular/cli/src/commands/mcp/tools/e2e.ts
amishne: Same as above. Reworded "runs" to "uses", though.

packages/angular/cli/src/commands/mcp/tools/e2e.ts
amishne: Good idea, expanded the use case section to give me details.

packages/angular/cli/src/commands/mcp/tools/devserver/devserver-stop.ts
amishne: I wasn't sure of this myself. In general I think it's preferred to throw when something weird is going on than to hide it, but in this specific case of trying to stop a running server, I prefer to err on the side of stopping it.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: Fair point. I think I'll defer it to the next PR, in which I want to focus on the workspace/project/version aspect.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: Added test file. The only test file it creates coverage overlap with is e2e_spec, but I kinda like the tests there as well since e2e has its own workspace logic.

packages/angular/cli/src/commands/mcp/testing/test-utils.ts
amishne: Only one test currently looks at this, I prefer to not introduce more here until we need it more widely.

packages/angular/cli/src/commands/mcp/tools/e2e_spec.ts
amishne: The new utils_spec.ts kinda works this way, but the way the other tests are written I think they're fine in their current sharing shape.

packages/angular/cli/src/commands/mcp/testing/test-utils.ts
amishne: I'm not that fond of creating actual temp files or folders here, especially since the the tests are intended to test how our local code behaves, not how `ng` responds to it. I think the right approach to improve confidence would be to add end-to-end tests to this, that actually invoke the real `ng` on a real toy app and verify the results. I prefer to do that in a follow-up though, especially after the workspace/project lookup change.

hybrist:

- main: be31e7caf9c550323485a23a91353c4ee59d47ff

PR #32075, opened 2025-12-09
ci: remove single value os matrix field from E2E jobs

clydin:
Since the `os` matrix field only ever has a single value, there is no need to include it in the matrix itself. This also shortens the display name by removing the otherwise static value from each matrix job entry.


alan-agius4 (approved):

clydin:

- main: 4cdc66bae261064e47966b9fa11090ab76752034
- 21.0.x: 735a15b497692b0819e5ad0435d554732034540c

PR #32082, opened 2025-12-10
fix(#31973): Added JavaScriptTransformer close to dispose method

### LINKED ISSUE #31973: Heap corruption error on windows when createCompilerPlugin is called before serveWithVite
### Command

build, serve

### Is this a regression?

v19.x

### Description

Ok so this is a complex one, I've already lost a couple of days on this. 

Native federation is using the "createCompilerPlugin" to create reusable artifacts (read dependencies) to be used in the Angular application. Now we know this is kind of an "own risk" scenario, especially since it's in the private folder of the @angular/build package but I am slowly losing it.

Some context 
When we do "ng serve" on our own builder, _which is basically a wrapper around the default angular build_, we use ESbuild to build some packages that we want to re-use across Angular micro frontends. The code can be found in our [esbuild-adapter.ts](https://github.com/angular-architects/module-federation-plugin/blob/main/libs/native-federation/src/utils/angular-esbuild-adapter.ts) and our [builder.ts](https://github.com/angular-architects/module-federation-plugin/blob/main/libs/native-federation/src/builders/build/builder.ts). The interesting part is that we don't get an error when we remove the [AngularCompilerPlugin](https://github.com/angular-architects/module-federation-plugin/blob/c77a7d387cd4d453d790b7b457332615120b0278/libs/native-federation/src/utils/angular-esbuild-adapter.ts#L297C1-L300C9) which we use to pre-build our artifacts. 

The error
Since Angular 20, we have the issue that during the [serveWithVite](https://github.com/angular-architects/module-federation-plugin/blob/c77a7d387cd4d453d790b7b457332615120b0278/libs/native-federation/src/builders/build/builder.ts#L361C7-L361C20) and to be precise in [the line where vite is imported.](https://github.com/angular/angular-cli/blob/0b439685dda66ce190afa2da425f02b8e5c94147/packages/angular/build/src/builders/dev-server/vite/index.ts#L184), an heap error occurs. We thought it could be ESbuild since it's also used in the vite package, but all our tries have been without result so far. 

[Whe have an issue with all the context](https://github.com/angular-architects/module-federation-plugin/issues/970). _Note:_ that all assumptions so far were wrong.  Even after ng21, the issue still persists. I thought it was a race condition [when closing/disposing the compiler-plugin](https://github.com/angular-architects/module-federation-plugin/issues/970#issuecomment-3552223609) but this is not the case. 

Now I know micro frontends and using the private part of the @angular/build is for our own risk, but I would never expect a heap error. Does anyone know how this could happen or what could cause this?  Final note: It is a race condition so it happens every now and then, about 40% of the ng serve builds. It happens only on windows and you can recognize the error by the build/serve just stopping without any error. 

All help is welcome, thanks a bunch in advance.

### Minimal Reproduction

We have a reproducable repo: https://github.com/berkon/ng-serve-crash-reproduction (it also crashes when you update to ng21). 

just npm ci and npm start. Even updating to ng21 doesn't help

### Exception or Error

```text
Process finished with exit code -1073740940 (0xC0000374)
```

### Your Environment

```text
Just windows 10+, reest can be seen in the repository
```

### Anything else relevant?

_No response_

Aukevanoost:
This change allows for awaiting the disposed plugin and added the missing close method of the JavaScript transformer to the onDispose. When the plugin is run multiple times on different projects/bundles without proper closing it will result in a node heap corruption. This allows us to cleanup the build before we start a new one.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

The current behavior doesn't cleanup the JavascriptTransformer and also doesn't await the promises. I know that ESbuild is not yet processing promises in the onDispose callback but I am working on that! https://github.com/evanw/esbuild/issues/4356

 In the meantime ESbuild will just treat the callback as a void so no functional changes. 

Closes #31973. 

## What is the new behavior?

For Angular no functional changes other than the clean disposal of the JavascriptTransformer to avoid Node heap corruptions. 

- [ ] Yes
- [x] No

## Other information

For us this means that we can create a wrapper around the plugin. This wrapper  creates a promise from the callback and awaits that promise after an ESbuild dispose. This allows us to fully cleanup the build after it's finished and thus avoids unwanted memory leaks.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
+++ b/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
@@ -588,11 +588,14 @@ export function createCompilerPlugin(
         logCumulativeDurations();
       });
 
-      build.onDispose(() => {
-        sharedTSCompilationState?.dispose();
-        void compilation.close?.();
-        void cacheStore?.close();
-      });
+      build.onDispose(
+        () =>
+          void Promise.all(
+            [compilation?.close?.(), cacheStore?.close(), javascriptTransformer.close()].filter(
+              Boolean,
+            ),
+          ).then(() => sharedTSCompilationState?.dispose()),
+      );
 
       /**
        * Checks if the file has side-effects when `advancedOptimizations` is enabled.
--- a/packages/angular/build/src/private.ts
+++ b/packages/angular/build/src/private.ts
@@ -59,6 +59,7 @@ export function createCompilerPlugin(
     ),
   );
 }
+export { getSharedCompilationStateCleanup } from './tools/esbuild/angular/compilation-state';
 
 export type { AngularCompilation } from './tools/angular/compilation';
 export { DiagnosticModes } from './tools/angular/compilation';
--- a/packages/angular/build/src/tools/esbuild/angular/compilation-state.ts
+++ b/packages/angular/build/src/tools/esbuild/angular/compilation-state.ts
@@ -10,6 +10,17 @@ export class SharedTSCompilationState {
   #pendingCompilation = true;
   #resolveCompilationReady: ((value: boolean) => void) | undefined;
   #compilationReadyPromise: Promise<boolean> | undefined;
+
+  /**
+   * ESbuild doesnt allow for awaiting the cleanup of plugins, therefore this
+   * allows consumers of the compiler-plugin to await the disposal of the plugin
+   * after ESbuild dispose function was called.
+   */
+  #disposeCompilation: (() => void) | undefined;
+  awaitCompilationDisposed: Promise<void> = new Promise((resolve) => {
+    this.#disposeCompilation = resolve;
+  });
+
   #hasErrors = true;
 
   get waitUntilReady(): Promise<boolean> {
@@ -37,6 +48,7 @@ export class SharedTSCompilationState {
 
   dispose(): void {
     this.markAsReady(true);
+    this.#disposeCompilation?.();
     globalSharedCompilationState = undefined;
   }
 }
@@ -45,4 +57,8 @@ let globalSharedCompilationState: SharedTSCompilationState | undefined;
 
 export function getSharedCompilationState(): SharedTSCompilationState {
   return (globalSharedCompilationState ??= new SharedTSCompilationState());
+}
+
+export function getSharedCompilationStateCleanup() {
+  return globalSharedCompilationState?.awaitCompilationDisposed ?? Promise.resolve();
 }


packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -588,11 +588,14 @@ export function createCompilerPlugin(
...
-        void cacheStore?.close();
-      });
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
+      build.onDispose(() =>
+        Promise.all(
+          [compilation?.close?.(), cacheStore?.close(), javascriptTransformer?.close()].filter(
+            (p): p is Promise<void> => p !== undefined,
+          ),
+        ).then(() => sharedTSCompilationState?.dispose()),
+      );
```
alan-agius4:
```suggestion
        void Promise.all(
          [compilation?.close?.(), cacheStore?.close(), javascriptTransformer.close()].filter(
            Boolean,
          ),
        ).then(() => sharedTSCompilationState?.dispose());
```

packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -588,11 +588,14 @@ export function createCompilerPlugin(
         logCumulativeDurations();
       });
 
-      build.onDispose(() => {
-        sharedTSCompilationState?.dispose();
-        void compilation.close?.();
-        void cacheStore?.close();
-      });
+      // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
alan-agius4:
```suggestion

```

alan-agius4 (changes requested):
Please update the commit msg to the below (as per guidelines) and squash the commits.

```
fix(@angular/build): added JavaScriptTransformer close to dispose method

This change allows for awaiting the disposed plugin and added the
missing close method of the JavaScript transformer to the onDispose.
When the plugin is run multiple times on different projects/bundles
without proper closing it will result in a node heap corruption. This
allows us to cleanup the build before we start a new one.

Closes #31973
```

packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
Aukevanoost: Unfortunately the eslint-disable-next-line is required here, otherwise it will break the linter. I am aware that I am bypassing a linter rule here but it's not functionally breaking the current behavior or risking undefined behavior.

Aukevanoost:
It should be good now. Let me know if I'm missing something.

packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
alan-agius4: With the suggested code change, lint will pass.

Aukevanoost:
@alan-agius4 

I think I found a robust fix that doesn't break the ESlint rules, but only exposes an Promise. Let me know what you think!

packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -588,11 +588,14 @@ export function createCompilerPlugin(
         logCumulativeDurations();
       });
 
-      build.onDispose(() => {
-        sharedTSCompilationState?.dispose();
-        void compilation.close?.();
-        void cacheStore?.close();
-      });
+      build.onDispose(
```
alan-agius4, line 591: Looking at this more, isn't the below enough to your issue? Is the getSharedCompilationStateCleanup truly needed?
build.onDispose(() => {
   sharedTSCompilationState?.dispose();
   void compilation.close?.();
   void cacheStore?.close();
   void javascriptTransformer.close();
});
The compilation, cacheStore and javascriptTransformer are local variables and scoped to a single plugin. The only shared instance is sharedTSCompilationState which it's clean up is sync. I am still failing to understand how the `sharedTSCompilationState` is causing the heap error,  as this is used primarily for `NoopCompilations` which in your case you are not using. 

My point here is that dispose can be fire and forget and it should not be blocking to start a new compilation, as long as sharedTSCompilationState.dispose is sync.

Aukevanoost: Hey, good question! 

While yes, they're scoped to 1 plugin. Their workers and state aren't. When you run this plugin sequentially in a single builder with different options/contexts, it will result in a heap corruption since the shared state is being accessed before the workers had time to close down. My fix awaits all workers and state to be disposed before the "i am disposed" promise is resolved, guaranteeing that the node process is stable again. All while still returning a void.

Aukevanoost: Hi great question. 

While the properties are local, their workers and state aren't. The `sharedTSCompilationState` is not the issue but the solution. The `compilation` and `javascriptTransformer` contain state and workers that crash the node:process on sequential ESbuilds. This is caused by the fact that the new build/compilation run is using workers and state that hasn't been cleaned up yet in the previous build, causing the heap corruption. The only way to fix this is to wait for all the workers and processes to be cleaned up before starting the next ESbuild run. The exposed `getSharedCompilationStateCleanup` function allows me to do this.

alan-agius4: > The `sharedTSCompilationState` is not the issue but the solution.

But in your plugin you are not even using `sharedTSCompilationState` since none of the compilations mode is noop. 


> While the properties are local, their workers and state aren't .... The compilation and javascriptTransformer contain state and workers that crash the node:process on sequential ESbuilds. This is caused by the fact that the new build/compilation run is using workers and state that hasn't been cleaned up yet in the previous build

I'm confused by this. What non local state are you your referring too? Both the compilation and the transformer operate with local state; they do not rely on a share state. Therefore, creating multiple instances of `createCompilerPlugin()` will not result in a single, shared pool of workers or compilation environment. Each invocation remains completely isolated and initializes its own separate resources. We do this in the Angular CLI where each `ng build` creates multiple instances of `createCompilerPlugin`.

packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -588,11 +588,14 @@ export function createCompilerPlugin(
...
+        () =>
+          void Promise.all(
+            [compilation?.close?.(), cacheStore?.close(), javascriptTransformer.close()].filter(
```
alan-agius4, line 594: If you do not mind, I might create (or you can if you wish too), to actually call  include just `void javascriptTransformer.close();` so that it's included in today's release.

Aukevanoost: I can create a separate PR for this if you don't mind!

alan-agius4: Sure, go ahead.

packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
Aukevanoost: While you are correct about the sharedTSCompilationState only used in noop mode, it is still being initialized regardless of the compilation type: https://github.com/angular/angular-cli/blob/c4af1932c3e8e94022237ea8e56adb19bf338978/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts#L146C1-L146C64 therefore it is (only) being used for the clean disposal signal. 

About the scoped workers and state, while I still can't proof exactly where it's happening, awaiting this disposal does prevent the heap error. I'll have to dig deeper to find out exactly which part is causing this.

alan-agius4: >  While you are correct about the sharedTSCompilationState only used in noop mode, it is still being initialized regardless of the compilation type: https://github.com/angular/angular-cli/blob/c4af1932c3e8e94022237ea8e56adb19bf338978/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts#L146C1-L146C64 therefore it is (only) being used for the clean disposal signal

That seems like an oversight.

Aukevanoost:
Closing this issue for now until there is an update on my part.

PR #32102, opened 2025-12-11
fix(@angular/build): conditionally manage Vitest UI option

clydin:
This change improves the handling of the Vitest UI option in the unit test builder. The `ui` option is now automatically disabled when running in a CI environment to prevent issues with automated pipelines. Additionally, the `ui` property is only included in the Vitest configuration if it's explicitly a boolean, making the option handling more robust.

Removing the default value for `ui` from the schema allows the option to be overridden by custom runner configuration files when not explicitly defined in `angular.json` or via the command line.


hybrist:

- main: 7c7e6a6142a9d294e04c612463449d2a4360e692
- 21.0.x: 4dd3c1a324c8e90808cc1c5febf65c8fa49dd3b9

PR #32124, opened 2025-12-15
feat(@angular/cli): add examples for @angular/aria

tjshiu:
Adds a comprehensive set of nine new examples for the `@angular/aria` package. These examples demonstrate the usage of directives for common accessibility patterns, including accordions, comboboxes, menus, tabs, toolbars, and trees.


alan-agius4:
Further to a chat with @clydin we shouldn't get this in. As he is working getting the support for that inside components repo itself.

PR #32137, opened 2025-12-16
fix(@angular/build): allow non-prefixed requests when using SSR and base href

### LINKED ISSUE #31896: Angular SSR: Express routes are only prefixed in development with baseHref set

alan-agius4:
When using SSR with a configured baseHref, the Vite dev server's default base middleware would intercept and reject requests that did not start with the base path. This created a mismatch with production behavior, where such requests (e.g., health check endpoints or custom routes defined in server.ts) could be handled correctly without the base prefix.

This change patches the Vite base middleware to allow requests that do not match the base path to proceed, enabling them to be handled by subsequent middlewares like the Angular SSR middleware.

Fixes #31896

### DIFF
--- a/packages/angular/build/src/tools/vite/middlewares/base-middleware.ts
+++ b/packages/angular/build/src/tools/vite/middlewares/base-middleware.ts
@@ -1,0 +1,56 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+import type { IncomingMessage, ServerResponse } from 'node:http';
+import type { Connect } from 'vite';
+import { addLeadingSlash } from '../../../utils/url';
+
+/**
+ * Patches the Vite base middleware to correctly handle the Angular application's base href.
+ * This is necessary because Vite's default base middleware might not align with Angular's
+ * expected path handling when using SSR, especially when a base href is configured.
+ *
+ * @param middlewares The Connect server instance containing the middleware stack.
+ * @param base The base URL path to be handled by the middleware.
+ */
+export function patchBaseMiddleware(middlewares: Connect.Server, base: string): void {
+  const entry = middlewares.stack.find(
+    ({ handle }) => typeof handle === 'function' && handle.name.startsWith('viteBaseMiddleware'),
+  );
+
+  if (typeof entry?.handle !== 'function') {
+    return;
+  }
+
+  entry.handle = function angularBaseMiddleware(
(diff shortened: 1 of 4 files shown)


packages/angular/build/src/tools/vite/middlewares/base-middleware.ts
```diff
@@ -0,0 +1,56 @@
...
+      return next();
+    }
+
+    const { pathname, hash, search } = new URL(url, 'http://localhost');
+    if (pathname === '/' || pathname === '/index.html') {
+      res.writeHead(302, { Location: `${base}${search}${hash}` });
+      res.end();
+
+      return;
+    }
```
clydin, line 52: Would the existing fallback middleware handle this case?

alan-agius4: Yes, the existing middleware handle this https://github.com/vitejs/vite/blob/e81c183f8c8ccaf7774ef0d0ee125bf63dbf30b4/packages/vite/src/node/server/middlewares/base.ts#L29-L36

The problem with the existing middleware is anything below https://github.com/vitejs/vite/blob/e81c183f8c8ccaf7774ef0d0ee125bf63dbf30b4/packages/vite/src/node/server/middlewares/base.ts#L38

clydin (approved):

alan-agius4:

- main: f39f7ee9529113f7c75d0e0e3ffa628fed9ce92f
- 21.0.x: 570ce8d3eeb280eeb6dca6ba54593c9325674741

PR #32088, opened 2025-12-10
build: use Node.js built-in TypeScript support for dev-infra scripts

clydin:
This change removes the use of `ts-node` to execute both `ng-dev` and the repo dev scripts. Due to the use of enums imports from `@angular/cli` package code for the analytics validation, the `admin` package script currently requires the `--experimental-transform-types` option.

PR #32171, opened 2025-12-18
build: restore `ng-dev` script

dgp1130:
This got dropped in https://github.com/angular/angular-cli/pull/32088/, but is needed because the release process calls `pnpm run ng-dev`, which requires a script to exist. Meanwhile `pnpm ng-dev` does not.


dgp1130:

- main: 70bc0f6d286910de43f03658872da00b282b2b55

PR #32174, opened 2025-12-18
Bump version to "v21.1.0-next.3" with changelog.

dgp1130:


dgp1130:
Ignoring the bun failure which feels unrelated and this is a prerelease anyways, will have to take a look later to see what might be going wrong here.

PR #32206, opened 2025-12-29
fix(@angular/cli): update yarn berry package manager configuration

clydin:
This commit updates the package manager descriptor for modern Yarn (Berry) to align with its current CLI options and configuration mechanisms:

- Removes `noLockfileFlag` as Yarn Berry does not support a direct `--no-lockfile` flag for adding packages.
- Updates `ignoreScriptsFlag` to use `--mode=skip-build`, which is the modern equivalent for skipping build scripts during installation.
- Changes `getRegistryOptions` to use `YARN_NPM_REGISTRY_SERVER` environment variable, which is the correct way to configure the registry in Yarn Berry.

### DIFF
--- a/packages/angular/cli/src/package-managers/parsers.ts
+++ b/packages/angular/cli/src/package-managers/parsers.ts
@@ -254,7 +254,9 @@ export function parseNpmLikeManifest(stdout: string, logger?: Logger): PackageMa
     return null;
   }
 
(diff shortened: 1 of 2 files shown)


clydin:

- main: 316fca8626d51b28ea8cd840f3815b7c6dfcfffa

PR #32235, opened 2026-01-07
feat(@angular/build): disable TestBed teardown during debugging in Vitest

### LINKED ISSUE #31733: `unit-test` builder should allow users to disable teardown for debugging reasons
### Command

test

### Description

It would be nice to provide a way to disable teardown from the CLI.

This is useful in headed browser mode to interact with the mounted component in it's final state after the test ends.

Maybe teardown could be disabled when `--debug` is set or maybe also provide a different boolean option such as `teardown` when using headed browser mode.

Once this is fixed https://github.com/angular/angular-cli/issues/31732, one can add a setup file with the following content:

```ts
TestBed.configureTestingModule({teardown: {destroyAfterEach: false}});
```

This could be configured dynamically at user's convenience.

clydin:
When running unit tests in debug mode, it is beneficial to keep the component's DOM state intact after the test completes for inspection. This change disables the automatic TestBed teardown (destroyAfterEach) when the 'debug' option is enabled.

Closes #31733

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
@@ -17,6 +17,7 @@ import { RunnerOptions } from '../api';
 function createTestBedInitVirtualFile(
   providersFile: string | undefined,
   projectSourceRoot: string,
+  teardown: boolean,
   polyfills: string[] = [],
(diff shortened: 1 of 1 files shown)


clydin:

- main: 87175f9dcdb7349dc8701fa1d5cf61c1b8542d42

PR #32237, opened 2026-01-07
fix(@angular/build): ensure correct project targeting during Vitest debugging

### LINKED ISSUE #31652: `unit-test` builder with `vitest` runner and `--debug` option fails
### Command

test

### Is this a regression?

Just before https://github.com/angular/angular-cli/commit/cf47d15efb4ad6ff68fc64be6d60df158c53baeb

### Description

When using `--debug` option with Vitest, it produces the following error: `AggregateError: Failed to initialize projects. There were errors during projects setup. See below for more details`.

This happens because Vitest checks that there is a chromium browser in the instances list: https://github.com/vitest-dev/vitest/blob/9e24a59f2c9dbd44439e38968c6c6eb216ebd50a/packages/vitest/src/node/config/resolveConfig.ts#L850C17-L850C37
but during this check it verifies that the project is not filtered out.

Problem is that the `project` option is set to the value of `projectName` which does not match the instance name `${projectName} (chromium)`.

It would be also nice to fail gracefully when debug is set and there is no chromium browser.

_(By the way, the error is truncated when using the CLI. In plain Vitest, the aggregated errors are logged)_

### Minimal Reproduction

`ng test --debug`

### Exception or Error

```text
AggregateError: Failed to initialize projects. There were errors during projects setup. See below for more details
```

### Your Environment

Build of cf47d15efb4ad6ff68fc64be6d60df158c53baeb.

### Anything else relevant?

_No response_

clydin:
This addresses issue #31652 where debugging failed to target the correct project instance. When running browser tests, Vitest appends the browser name to the project identifier. This change updates the CLI to match that naming convention when initiating a debug session, ensuring the correct project is selected.

Fixes #31652

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/executor.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/executor.ts
@@ -6,9 +6,9 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
-import type { BuilderOutput } from '@angular-devkit/architect';
+import type { BuilderContext, BuilderOutput } from '@angular-devkit/architect';
 import assert from 'node:assert';
-import path, { join } from 'node:path';
+import path from 'node:path';
 import type { Vitest } from 'vitest/node';
 import {
   DevServerExternalResultMetadata,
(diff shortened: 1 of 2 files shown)


clydin:

- main: 32adc3a757a1e75cf8d44a227c57f7947053ca8c
- 21.0.x: 2b9be3a7cadc444e67d629c2a74c54f0c65f4a54

PR #32263, opened 2026-01-10
fix(@angular/ssr): handle platform destruction during rendering

arturovt:
Fixes crash when platform/app destroys itself during the bootstrapping and stabilization phase. Previously, the code would call `applicationRef.injector` without checking if the platform was destroyed, resulting in: "Error: Injector has already been destroyed"

This can occur when:
- Component constructor calls `inject(PlatformRef).destroy()`
- AbortSignal triggers during request handling
- APP_INITIALIZER rejects and causes cleanup
- Custom guard/resolver logic destroys the platform

Solution: Check `applicationRef.destroyed` after `whenStable()` and return error state instead of accessing destroyed injector.

Test: Added test case that destroys app in component constructor to verify graceful handling of this edge case.

### DIFF
--- a/packages/angular/ssr/src/utils/ng.ts
+++ b/packages/angular/ssr/src/utils/ng.ts
@@ -59,7 +59,10 @@ export async function renderAngular(
   url: URL,
   platformProviders: StaticProvider[],
   serverContext: string,
-): Promise<{ hasNavigationError: boolean; redirectTo?: string; content: () => Promise<string> }> {
+): Promise<
+  | { hasNavigationError: true }
+  | { hasNavigationError: boolean; redirectTo?: string; content: () => Promise<string> }
+> {
   // A request to `http://www.example.com/page/index.html` will render the Angular route corresponding to `http://www.example.com/page`.
   const urlToRender = stripIndexHtmlFromURL(url);
   const platformRef = platformServer([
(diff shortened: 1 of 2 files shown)


packages/angular/ssr/src/utils/ng.ts
```diff
@@ -100,6 +100,19 @@ export async function renderAngular(
...
+    // valid case. We should not assume the `applicationRef` is not in destroyed state.
+    // Calling `envInjector.get` would throw `NG0205: Injector has already been destroyed`.
+    if (applicationRef.destroyed) {
+      return {
+        hasNavigationError,
+        redirectTo: undefined,
+        // TODO(alanagius): let's think on the content?
+        // or should we use `Promise.resolve('...')`?
+        content: () => Promise.reject(new Error('Application was destroyed during bootstrapping')),
+      };
```
alan-agius4: You could change the return type of this function to 

```ts
): Promise<
  | { hasNavigationError: boolean; redirectTo?: string; content: () => Promise<string> }
  | { hasNavigationError: true }
> {
```

and change the return to 
```suggestion
      return {
        hasNavigationError: true
     };
```

arturovt: Thanks, updated.

dgp1130:

- main: fd1a2091d20709781438e1c588a79e8c93cf5183
- 21.0.x: 730ae6609b847802124a5c6e12c77522af54b967
- 21.1.x: 8d99cfeb09a2c4bf8d12f453dc9ac942103cdffd

PR #32276, opened 2026-01-12
refactor(@angular/cli): remove resolve dependency and use native require.resolve

clydin:
This commit removes the `resolve` dependency from `@angular/cli` and replaces its usage with native `require.resolve` via `createRequire`. The `findPackageJson` utility is now only used in migration-only scenarios where the Node.js module caching behavior that prompted the original use of `resolve` is no longer a concern. This reduces the package dependency footprint.


packages/angular/cli/src/commands/update/cli.ts
```diff
@@ -697,3 +696,14 @@ async function readPackageManifest(manifestPath: string): Promise<PackageManifes
     return undefined;
   }
 }
+
+function findPackageJson(workspaceDir: string, packageName: string): string | undefined {
+  try {
+    const projectRequire = createRequire(path.join(workspaceDir, 'package.json'));
+    const packageJsonPath = projectRequire.resolve(`${packageName}/package.json`);
```
hybrist, line 703: There's a small gotcha for this which is that `exports` fields _can_ mess with resolving `package.json`. But 99.9% of the time, an Angular app shouldn't have an `exports` field (and 90% of the remaining time, people should've added `package.json` to their exports).

alan-agius4 (approved):

packages/angular/cli/src/commands/update/cli.ts
clydin: Packages that contain Angular migrations must have a resolvable package.json so in this specific usage we can luckily avoid that situation.

clydin:

- main: e6b62a1f2f9b7f21fb438b8c254266e4e4a8e75d

PR #32279, opened 2026-01-13
refactor(@angular/cli): remove old package manager utilities

alan-agius4:
This removes the old implementation of the package-manager.

### DIFF
--- a/packages/angular/cli/src/package-managers/package-manager.ts
+++ b/packages/angular/cli/src/package-managers/package-manager.ts
@@ -552,6 +552,8 @@ export class PackageManager {
         // If the cache directory doesn't exist, create it.
         await this.host.mkdir(tempDirectory, { recursive: true });
       }
+
+      return tempDirectory;
     }
 
     const tempOptions = ['node_modules'];


packages/angular/cli/src/commands/add/cli.ts
```diff
@@ -334,13 +331,9 @@ export default class AddCommandModule
...
     }
 
-    context.packageManager = await createPackageManager({
-      cwd: this.context.root,
-      logger: this.context.logger,
-      dryRun: context.dryRun,
-      tempDirectory,
-    });
-    task.output = `Using package manager: ${color.dim(context.packageManager.name)}`;
+    this.context.packageManager.temporaryDirectory = tempDirectory;
```
clydin: Probably should move the temp directory discovery logic to the command runner as well so that other commands can take advantage of it too.

packages/angular/cli/src/package-managers/package-manager.ts
```diff
@@ -329,14 +333,34 @@ export class PackageManager {
...
+
+    return this.#version;
+  }
 
-    return stdout.trim();
+  /**
+   * Gets the version of the package manager binary.
+   * @returns The version of the package manager binary.
+   */
+  private getVersion(): string {
```
clydin: I think we can use the async version throughout now.  This also allows it to go through the host for debug logging.

packages/angular/cli/src/command-builder/command-runner.ts
```diff
@@ -163,3 +177,58 @@ async function getCommandsToRegister(
...
+
+/**
+ * Gets the configured package manager by checking package.json, or the local and global angular.json files.
+ *
+ * @param root The root directory of the workspace.
+ * @param localWorkspace The local workspace.
+ * @param globalWorkspace The global workspace.
+ * @returns The package manager name.
+ */
+async function getConfiguredPackageManager(
```
clydin, line 192: Nice.

packages/angular/cli/src/package-managers/package-manager.ts
alan-agius4: Done

packages/angular/cli/src/package-managers/package-manager.ts
```diff
@@ -544,6 +542,29 @@ export class PackageManager {
...
+    const { tempDirectory } = this.options;
+
+    if (tempDirectory && !relative(this.cwd, tempDirectory).startsWith('..')) {
+      try {
+        await this.host.stat(tempDirectory);
+      } catch {
+        // If the cache directory doesn't exist, create it.
+        await this.host.mkdir(tempDirectory, { recursive: true });
+      }
+    }
```
clydin, line 557:
```suggestion
      return tempDirectory;
    }
```

alan-agius4:

- main: f7123612bc7d910477dcd3a90042115364f5a7fc

PR #32290, opened 2026-01-14
refactor(@angular/cli): decouple `AnalyticsCollector` from `CommandContext`

clydin:
The `AnalyticsCollector` class is made more modular and easier to test by removing its direct dependency on the `CommandContext`. Instead of accepting the entire context object, the constructor now receives only the specific dependencies it requires: the logger, user ID, and package manager information.
Also removes a circular reference:
`packages/angular/cli/src/analytics/analytics-collector.ts` → `packages/angular/cli/src/command-builder/command-module.ts`

### DIFF
--- a/goldens/circular-deps/packages.json
+++ b/goldens/circular-deps/packages.json
@@ -32,10 +32,6 @@
     "packages/angular/build/src/utils/server-rendering/manifest.ts"
   ],
   [
-    "packages/angular/cli/src/analytics/analytics-collector.ts",
-    "packages/angular/cli/src/command-builder/command-module.ts"
-  ],
-  [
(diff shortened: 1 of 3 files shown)


goldens/circular-deps/packages.json
```diff
@@ -31,10 +31,6 @@
     "packages/angular/build/src/tools/esbuild/utils.ts",
     "packages/angular/build/src/utils/server-rendering/manifest.ts"
   ],
-  [
```
hybrist, line 34: ❤️

hybrist (approved):

clydin:

- main: 630584ebbbbdedf0ac7b8768302af7ea87b1703f

PR #32294, opened 2026-01-14
feat(@angular/cli): standardize MCP tools around workspace/project options

amishne:
- Many tools now have `workspace` and `project` options that make them better suited to monorepo scenarios.
- Many errors now reported as exceptions instead of regular messages. They all have informative error messages.
- The above now uses shared implementation across tools.
- Tests are now simplified and more consistent.

Tested manually with Gemini-CLI, both in an Angular directory and in a non-Angular parent directory.

### DIFF
--- a/packages/angular/cli/src/commands/mcp/tools/devserver/devserver-wait-for-build.ts
+++ b/packages/angular/cli/src/commands/mcp/tools/devserver/devserver-wait-for-build.ts
@@ -7,7 +7,10 @@
  */
 
 import { z } from 'zod';
-import { createStructuredContentOutput, getDefaultProjectName } from '../../utils';
+import { Devserver, createDevServerNotFoundError, getDevserverKey } from '../../devserver';
+import { workspaceAndProjectOptions } from '../../shared-options';
+import { createStructuredContentOutput } from '../../utils';
+import { resolveWorkspaceAndProject } from '../../workspace-utils';
 import { type McpToolContext, type McpToolDeclaration, declareTool } from '../tool-registry';
 
 /**
@@ -21,12 +24,7 @@ export const WATCH_DELAY = 1000;
 const DEFAULT_TIMEOUT = 180_000; // In milliseconds
 
 const devserverWaitForBuildToolInputSchema = z.object({
-  project: z
-    .string()
-    .optional()
-    .describe(
-      'Which project to wait for in a monorepo context. If not provided, waits for the default project server.',
-    ),
+  ...workspaceAndProjectOptions,
   timeout: z
     .number()
     .default(DEFAULT_TIMEOUT)
@@ -39,7 +37,7 @@ export type DevserverWaitForBuildToolInput = z.infer<typeof devserverWaitForBuil
 
 const devserverWaitForBuildToolOutputSchema = z.object({
   status: z
-    .enum(['success', 'failure', 'unknown', 'timeout', 'no_devserver_found'])
+    .enum(['success', 'failure', 'unknown', 'timeout'])
     .describe(
       "The status of the build if it's complete, or a status indicating why the wait operation failed.",
     ),
@@ -59,39 +57,26 @@ export async function waitForDevserverBuild(
   input: DevserverWaitForBuildToolInput,
   context: McpToolContext,
 ) {
-  if (context.devservers.size === 0) {
-    return createStructuredContentOutput({
-      status: 'no_devserver_found',
-      logs: undefined,
-    });
+  const { workspacePath, projectName } = await resolveWorkspaceAndProject({
+    host: context.host,
+    workspacePathInput: input.workspace,
+    projectNameInput: input.project,
+    mcpWorkspace: context.workspace,
+  });
+  const key = getDevserverKey(workspacePath, projectName);
+  const devserver = context.devservers.get(key);
+
+  if (!devserver) {
+    throw createDevServerNotFoundError(context.devservers);
   }
 
-  let projectName = input.project ?? getDefaultProjectName(context);
+  return performWait(devserver, input.timeout);
+}
 
-  if (!projectName) {
-    // This should not happen. But if there's just a single running devserver, wait for it.
-    if (context.devservers.size === 1) {
-      projectName = Array.from(context.devservers.keys())[0];
-    } else {
-      return createStructuredContentOutput({
-        status: 'no_devserver_found',
-        logs: undefined,
-      });
-    }
-  }
-
-  const devServer = context.devservers.get(projectName);
-
-  if (!devServer) {
-    return createStructuredContentOutput<DevserverWaitForBuildToolOutput>({
-      status: 'no_devserver_found',
-      logs: undefined,
-    });
-  }
-
-  const deadline = Date.now() + input.timeout;
+async function performWait(devserver: Devserver, timeout: number) {
+  const deadline = Date.now() + timeout;
   await wait(WATCH_DELAY);
-  while (devServer.isBuilding()) {
+  while (devserver.isBuilding()) {
     if (Date.now() > deadline) {
       return createStructuredContentOutput<DevserverWaitForBuildToolOutput>({
         status: 'timeout',
@@ -102,7 +87,7 @@ export async function waitForDevserverBuild(
   }
 
   return createStructuredContentOutput<DevserverWaitForBuildToolOutput>({
-    ...devServer.getMostRecentBuild(),
+    ...devserver.getMostRecentBuild(),
   });
 }
 
@@ -123,14 +108,11 @@ recent build.
   tool or command. When it retuns you'll get build logs back **and** you'll know the user's devserver is up-to-date with the latest changes.
 </Use Cases>
 <Operational Notes>
-* This tool expects that a dev server was launched on the same project with the "devserver.start" tool, otherwise a "no_devserver_found"
-  status will be returned.
+* This tool expects that a dev server was launched on the same project with the "devserver.start" tool, otherwise the tool will fail.
 * This tool will block until the build is complete or the timeout is reached. If you expect a long build process, consider increasing the
   timeout. Timeouts on initial run (right after "devserver.start" calls) or after a big change are not necessarily indicative of an error.
 * If you encountered a timeout and it might be reasonable, just call this tool again.
 * If the dev server is not building, it will return quickly, with the logs from the last build.
-* A 'no_devserver_found' status can indicate the underlying server was stopped for some reason. Try first to call the "devserver.start"
-  tool again, before giving up.
 </Operational Notes>
 `,
   isReadOnly: true,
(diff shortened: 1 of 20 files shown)


packages/angular/cli/src/commands/mcp/tools/devserver/devserver-wait-for-build.ts
```diff
@@ -59,39 +60,26 @@ export async function waitForDevserverBuild(
...
-    return createStructuredContentOutput<DevserverWaitForBuildToolOutput>({
-      status: 'no_devserver_found',
-      logs: undefined,
-    });
-  }
+  return performWait(devserver, input.timeout);
+}
 
-  const deadline = Date.now() + input.timeout;
+async function performWait(devserver: Devserver, timeout: number) {
```
dgp1130, line 76: Suggestion: Use an explicit return type. Otherwise, TS will implicitly union all the `return` statement types and wouldn't prevent from accidentally returning the wrong type.

https://techhub.social/@develwithoutacause/111842243141199473

packages/angular/cli/src/commands/mcp/tools/devserver/devserver_spec.ts
```diff
@@ -182,4 +190,20 @@ describe('Serve Tools', () => {
...
+    // Start app-one
+    await startDevserver({ project: 'app-one' }, mockContext);
+
+    // Try to stop app-two (which is not running)
+    try {
+      await stopDevserver({ project: 'app-two' }, mockContext);
+      fail('Should have thrown');
+    } catch (e) {
+      expect((e as Error).message).toContain('Dev server not found. Currently running servers:');
+      expect((e as Error).message).toContain("- Project 'app-one' in workspace path '/test'");
```
dgp1130, line 206: Consider: Should this message contain the port number too?

packages/angular/cli/src/commands/mcp/utils.ts
```diff
@@ -110,3 +111,139 @@ export function getCommandErrorLogs(e: unknown): string[] {
...
+
+export function createNoProjectResolvedError(workspacePath: string): Error {
+  return new Error(
+    `No project name provided and no default project found in workspace path ${workspacePath}. ` +
+      'Please provide a project name or set a default project in angular.json. ' +
+      "You can use 'list_projects' to find available projects.",
+  );
+}
+
+export function createDevServerNotFoundError(
```
dgp1130: Consider: Might be better to leave this in `devserver-stop.ts` since it specifically relates to devserver functionality.

packages/angular/cli/src/commands/mcp/utils.ts
```diff
@@ -110,3 +111,139 @@ export function getCommandErrorLogs(e: unknown): string[] {
...
+  );
+}
+
+export function createNoProjectResolvedError(workspacePath: string): Error {
+  return new Error(
+    `No project name provided and no default project found in workspace path ${workspacePath}. ` +
+      'Please provide a project name or set a default project in angular.json. ' +
+      "You can use 'list_projects' to find available projects.",
+  );
+}
```
dgp1130: Consider: Do these helpers need to be exported? Will they ever realistically be called outside this file if `resolveWorkspaceAndProject` is doing most of the work?

Going a step further, if these errors are only created in one place, would it be more clear to just inline them at their individual call sites?

packages/angular/cli/src/commands/mcp/utils.ts
```diff
@@ -110,3 +111,139 @@ export function getCommandErrorLogs(e: unknown): string[] {
...
+ * Resolves workspace and project for tools to operate on.
+ *
+ * If `workspacePathInput` is absent, uses the MCP's configured workspace. If none is configured, use the
+ * current directory as the workspace.
+ * If `projectNameInput` is absent, uses the default project in the workspace.
+ */
+export async function resolveWorkspaceAndProject({
+  host,
+  workspacePathInput,
+  projectNameInput,
```
dgp1130: Suggestion: I don't think we need the `Input` suffix here, I'd just call these `workspacePath` and `projectName`.

If the goal is to have a local variable with the same name, you can do that by aliasing in the destructure without altering the public API of the function.

```typescript
function func({foo: bar}: {foo: string}): string {
    const foo = bar;
    return foo;
}

func({foo: 'test'});
```

packages/angular/cli/src/commands/mcp/utils.ts
```diff
@@ -110,3 +111,139 @@ export function getCommandErrorLogs(e: unknown): string[] {
...
+}
+
+/**
+ * Resolves workspace and project for tools to operate on.
+ *
+ * If `workspacePathInput` is absent, uses the MCP's configured workspace. If none is configured, use the
+ * current directory as the workspace.
+ * If `projectNameInput` is absent, uses the default project in the workspace.
+ */
+export async function resolveWorkspaceAndProject({
```
dgp1130: Suggestion: I'm a bit wary of this `utils` file getting a bit long and have a nebulous purpose. Could we split it up into separate (or at least rename to a well-defined) domain-specific file(s)? Something like `projects.ts` or `workspaces.ts`? Could also combined `shared-options.ts` into the same file.

packages/angular/cli/src/commands/mcp/utils.ts
```diff
@@ -110,3 +111,139 @@ export function getCommandErrorLogs(e: unknown): string[] {
...
+}): Promise<{
+  workspace: AngularWorkspace;
+  workspacePath: string;
+  projectName: string;
+}> {
+  let workspacePath: string;
+  let workspace: AngularWorkspace;
+
+  if (workspacePathInput) {
+    if (!host.existsSync(workspacePathInput)) {
```
dgp1130: Consider: Can/should we use `async` APIs here for improved parallelism? Not sure how much of a concern multiple MCP requests would be.

packages/angular/cli/src/commands/mcp/utils.ts
dgp1130: Consider: Can we split this implementation into separate `resolveWorkspace` and `resolveProject` functions? I think this would make the internals easier to follow as we can leverage early returns rather than stateful local variables.

packages/angular/cli/src/commands/mcp/utils.ts
dgp1130: Consider: Do we really care about this as a distinct error case from `${workspacePathInput}/angular.json` as checked below? I'm not sure this is worth the maintenance effort / performance cost of an additional `stat` call.

packages/angular/cli/src/commands/mcp/utils.ts
```diff
@@ -110,3 +111,139 @@ export function getCommandErrorLogs(e: unknown): string[] {
...
+ * If `workspacePathInput` is absent, uses the MCP's configured workspace. If none is configured, use the
+ * current directory as the workspace.
+ * If `projectNameInput` is absent, uses the default project in the workspace.
+ */
+export async function resolveWorkspaceAndProject({
+  host,
+  workspacePathInput,
+  projectNameInput,
+  mcpWorkspace,
+  workspaceLoader = AngularWorkspace.load,
```
dgp1130: Consider: Is this option just for testing purposes? Can we drop it from the public API or at least make it more clearly not for general usage?

I'm thinking you can either directly depend on `AngularWorkspace.load` and then spy on it in the test, or define this as a test only parameter like:

```typescript
function resolveWorkspaceAndProject(
  {/* regular options */}: {/* ... */},
  testOnlyOptions: {
    workspaceLoader = AngularWorkspace.load,
  } = {},
): Promise<{ /* ... */}>;
```

dgp1130 (approved):
Just some minor suggestions, I'm not sure how much of the `resolveWorkspaceAndProject` implementation was reused so I might be repeating previous feedback. Feel free to ignore / defer anything we've already discussed or which can be followed up on later.

packages/angular/cli/src/commands/mcp/tools/devserver/devserver-wait-for-build.ts
amishne: Getting the factory function types is actually trickier than it seems and is cross-cutting across all tools (including ones not included in this change) so I don't want to include it in this PR.

packages/angular/cli/src/commands/mcp/tools/devserver/devserver_spec.ts
amishne: These errors are worded around how the agent should correct its argument for the next call, the port is not needed for that. Also presumably the agent should have all the ports already reported from all the start commands.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: It's used for both `stop` and `wait_for_build` so I wanted it in a central place. I did move it to `devserver.ts` though.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: Inlined them.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: I wanted to separate them because the `input` varieties are 1-to-1 with the function input, so for example they could be missing, while the non-input versions are non-nullable results of looking the inputs up.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: Split into a separate `workspace-utils.ts` file, now only some small stuff remained in `utils.ts`.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: I don't think it's critical here, and if anything I feel like if this becomes noticeable, some sort of caching would be a better direction.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: My original implementation was split! But then I realized (1) practically all tools want them both together, and (2) it would lead to duplicate lines of code running. So batched them together for now.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: I wanted very precise error messages for the agent. I think it also helps clarify that the correct input is a workspace path and not an angular.json path. Ack regarding the cost but I don't think the added check is significant here.

packages/angular/cli/src/commands/mcp/utils.ts
amishne: Done, now spying on it.

alan-agius4:

- main: 98a24d0401f36f484dc9c4d8b0f5284ffa524f19

PR #32308, opened 2026-01-15
refactor(@angular/cli): optimize package manager version detection

clydin:
This commit introduces a "hot cache" mechanism for the package manager version. When the package manager factory determines the version (e.g., to distinguish between Yarn and Yarn Classic), it now passes that discovered version to the `PackageManager` instance.

The `PackageManager` class has been updated to:
1. Accept an optional `version` in its constructor options.
2. Cache the version internally after the first fetch.
3. Use the cached/injected version in `getVersion()` to avoid redundant shell commands.

### DIFF
--- a/packages/angular/cli/src/package-managers/factory.ts
+++ b/packages/angular/cli/src/package-managers/factory.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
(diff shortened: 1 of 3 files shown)


clydin:

- main: 284c47d9d99e54858e92e9fd6cf0afb37660b641

PR #32330, opened 2026-01-20
refactor: remove deprecated compiler option

crisbeto:
Removes usages of the `baseUrl` compiler option which is deprecated and will start throwing an error in TypeScript 6.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/browser/specs/tsconfig-paths_spec.ts
+++ b/packages/angular_devkit/build_angular/src/builders/browser/specs/tsconfig-paths_spec.ts
@@ -21,16 +21,20 @@ describe('Browser Builder tsconfig paths', () => {
 
   it('works', async () => {
     host.replaceInFile('src/app/app.module.ts', './app.component', '@root/app/app.component');
+
+    // TODO(crisbeto): the `baseUrl` here will trigger a deprecation error in TS6. We may
+    // have to opt out of it for this test since Webpack seems to depend on the `baseUrl`.
     host.replaceInFile(
       'tsconfig.json',
-      /"baseUrl": ".\/",/,
+      /"compilerOptions": {/,
       `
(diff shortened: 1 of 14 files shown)


packages/angular_devkit/build_angular/src/builders/browser/specs/tsconfig-paths_spec.ts
```diff
@@ -43,23 +47,27 @@ describe('Browser Builder tsconfig paths', () => {
       'src/app/shared/meaning.ts': 'export var meaning = 42;',
       'src/app/shared/index.ts': `export * from './meaning'`,
     });
+
+    // TODO(crisbeto): the `baseUrl` here will trigger a deprecation error in TS6. We may
+    // have to opt out of it for this test since Webpack seems to depend on the `baseUrl`.
```
crisbeto, line 52: The context here is that it _should_ work without the `baseUrl` since the default has been `./` for a while, but it doesn't for some reason. I suspect something in Webpack might be depending on it. Once TS6 is out, we can opt out these tests from the deprecation error.

packages/angular_devkit/build_angular/src/builders/karma/tests/options/code-coverage_spec.ts
```diff
@@ -67,10 +67,10 @@ describeKarmaBuilder(execute, KARMA_BUILDER_INFO, (harness, setupTarget) => {
 
     it(`should collect coverage from paths in 'sourceRoot'`, async () => {
```
alan-agius4, line 68: This changes the test. Coverage from `node_modules` is always ignored.
In this case, we want to ensure that coverage for local libraries is not included (IE: not included in `node_module`).

crisbeto:
I'll close this one for now and will deal with it when bumping to TS6.

PR #32349, opened 2026-01-22
fix(@angular/build): support TypeScript isolatedDeclarations option

cexbrayat:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Recently, the Angular team said [on a panel](https://www.youtube.com/watch?v=OpprrzvyqEc&t=445s)  that `isolatedDeclarations` can make a huge difference in build times, but we currently can't test it in the CLI.

Angular CLI forcefully set both declaration and composite to false, which prevents the use of isolatedDeclarations as TypeScript requires at least one of these options to be true when isolatedDeclarations is set (TS5069 error):

```
✘ [ERROR] TS5069: Option 'isolatedDeclarations' cannot be specified without specifying option 'declaration' or option 'composite'. [plugin angular-compiler]
```

## What is the new behavior?

This change allows the use of TypeScript's isolatedDeclarations compiler option by conditionally preserving the declaration and composite options when isolatedDeclarations is enabled.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
+++ b/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
@@ -733,7 +733,9 @@ function createCompilerOptionsTransformer(
     return {
       ...compilerOptions,
       noEmitOnError: false,
-      composite: false,
+      // TypeScript requires either declaration or composite to be true when isolatedDeclarations is set
+      declaration: compilerOptions.isolatedDeclarations ? compilerOptions.declaration : false,
+      composite: compilerOptions.isolatedDeclarations ? compilerOptions.composite : false,
       inlineSources: !!pluginOptions.sourcemap,
       inlineSourceMap: !!pluginOptions.sourcemap,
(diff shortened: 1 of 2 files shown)


cexbrayat:
@clydin @alan-agius4 In your opinion, can `isolatedDeclarations` speed up builds with the CLI?

packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts
```diff
@@ -733,7 +733,9 @@ function createCompilerOptionsTransformer(
     return {
       ...compilerOptions,
       noEmitOnError: false,
-      composite: false,
+      // TypeScript requires either declaration or composite to be true when isolatedDeclarations is set
```
alan-agius4, line 736: This will most likely slow down the build since we'd be generating declarations we don't really need for an app. Also, our current setup does handles composite projects yet.

PR #32357, opened 2026-01-23
fix(@angular/build): support merging coverage thresholds with Vitest runnerConfig

clydin:
The Vitest unit test builder now correctly merges coverage thresholds and watermarks from a user's Vitest configuration with CLI-provided options. Previously, providing any CLI thresholds would completely replace the configuration's thresholds. Now, partial CLI thresholds are merged, allowing users to override specific metrics while keeping others from their config.

This change also ensures that the builder correctly reports failure when Vitest coverage thresholds are not met by monitoring `process.exitCode`.


alan-agius4:

- main: 1f1b21dab4fdec7cdbd6cd619f6e95708fd6484c
- 21.1.x: 2d30639d3c5a0eb3a1f40ec4cd8fe157f28f19f5

PR #32358, opened 2026-01-23
fix(@angular/build): loosen Vitest dependency checks when runnerConfig is used

clydin:
When a Vitest configuration file is provided via the `runnerConfig` option, the builder now only validates that the `vitest` package itself is installed. Checks for specific browser providers, DOM environments (jsdom/happy-dom), and coverage providers are skipped, as the custom configuration may handle these or use alternative setups that the builder cannot predict.


alan-agius4:

- main: 98ffb68f608ee48c14e67cd6a3dec3e33582c5ba
- 21.1.x: 80911af673365af3bcb86760bebb4200967ca433

PR #32360, opened 2026-01-24
refactor(@angular/build): avoid secondary import wrapper in Vitest unless coverage is enabled

clydin:
This commit optimizes the Vitest test runner by removing the secondary import wrapper for test entry points when code coverage is not enabled. The wrapper is only necessary to support coverage exclusion of the test files themselves.

By capturing the resolved Vitest configuration within the `configureVitest` hook, the plugin can now determine if coverage is enabled and conditionally apply the wrapper. This simplifies the module graph for standard test runs.


alan-agius4:

- main: 0b4982720e111bf5029bcf97f7e0ce2658c42d43

PR #32385, opened 2026-01-28
fix(@angular/cli): renamed files by their new path in the schematic workflow

alan-agius4:
Prior to this change, the old filename was being tracked.


alan-agius4:

- main: e5d3b942e944afce199da98e6bddc938b82515e8

PR #32400, opened 2026-01-30
fix(@schematics/angular): warn when production configuration is missing for service worker

### LINKED ISSUE #32399: bug(@angular/pwa): ng add @angular/pwa fails to update angular.json and register service worker
### Command

add

### Is this a regression?

20

### Description

After running ng add @angular/pwa, the application fails to register the service worker. Specifically, the browser attempts to fetch ngsw-worker.js but receives a 404 Not Found error. Investigation shows that the angular.json file was either not updated with the serviceWorker: true flag or the build process is not generating the required service worker files in the output directory.

The issue is resolved only after manually editing angular.json. This indicates the schematic is failing to write the necessary configuration to the workspace file even though it reports success.

Image

### Expected Behavior
The ng add schematic should correctly configure angular.json so that ng build generates ngsw-worker.js and ngsw.json in the production output. The service worker should be successfully registered and visible in chrome://serviceworker-internals/.

Environment
Angular: 21.0.8
Angular CLI:  21.0.5
Node: 24.12.0
OS: darwin arm64

### Minimal Reproduction

    1. Run ng add @angular/pwa --project your-project-name.
    2. Run a production build: ng build --configuration production.
    3. Serve the dist/ folder using a static server (e.g., http-server).
    4. Open the browser and check the Network Tab.
    5. Observed Error: A request for /ngsw-worker.js fails with a 404 error. 

Image

### Exception or Error

```text
Always getting Network error 404 while trying to request ngsw-worker.js file.
```

### Your Environment

```text
Angular CLI       : 21.0.5
Angular           : 21.0.8
Node.js           : 24.12.0
Package Manager   : pnpm 10.28.1
Operating System  : darwin arm64
```

### Anything else relevant?

Usin Pnpm Package menager
Issue happens woth http-server and default ng server

alan-agius4:
When adding service worker support to a project, the schematic attempts to add the `serviceWorker` option to the `production` build configuration. If the `production` configuration is missing, the schematic would previously do nothing and not inform the user.

This change adds a warning message to the console when the `production` configuration is not found, making it clear to the user why the `serviceWorker` option was not added.

Fixes #32399

### DIFF
--- a/packages/schematics/angular/service-worker/index.ts
+++ b/packages/schematics/angular/service-worker/index.ts
@@ -106,7 +106,7 @@ function getTsSourceFile(host: Tree, path: string): ts.SourceFile {
 }
 
 const serviceWorkerSchematic: RuleFactory<ServiceWorkerOptions> = createProjectSchematic(
-  async (options, { project, workspace, tree, logger }) => {
+  async (options, { project, workspace, tree, context: { logger } }) => {
     if (project.extensions.projectType !== 'application') {
       throw new SchematicsException(`Service worker requires a project type of "application".`);
     }


clydin (approved):

alan-agius4:

- main: 51fc77828a33fdf35051b7e18d79ad43f90cba1d
- 21.1.x: a18196a1096e5eb69cf64102943781d34c4389bf

PR #32417, opened 2026-02-03
refactor(@angular/cli): implement lazy validation for package manager

clydin:
This changes the package manager initialization to only throw errors when the binary is actually required for an operation. This allows CLI commands that do not depend on the package manager binary to function even if the configured package manager is missing.

### DIFF
--- a/packages/angular/cli/src/package-managers/package-manager.ts
+++ b/packages/angular/cli/src/package-managers/package-manager.ts
@@ -71,6 +71,12 @@ export interface PackageManagerOptions {
    * instead of running the version command.
    */
   version?: string;
+
+  /**
+   * An error that occurred during the initialization of the package manager.
+   * If provided, this error will be thrown when attempting to execute any command.
+   */
+  initializationError?: Error;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/angular/cli/src/package-managers/package-manager.ts
```diff
@@ -113,6 +121,18 @@ export class PackageManager {
     return this.descriptor.binary;
   }
 
+  /**
+   * Ensures that the package manager is installed and available in the PATH.
+   * If it is not, this method will throw an error with instructions on how to install it.
+   *
+   * @throws {Error} If the package manager is not installed.
+   */
+  async ensureInstalled(): Promise<void> {
```
alan-agius4:
```suggestion
  ensureInstalled(): void {
```

packages/angular/cli/src/package-managers/package-manager.ts
```diff
@@ -142,6 +162,8 @@ export class PackageManager {
     args: readonly string[],
     options: { timeout?: number; registry?: string; cwd?: string } = {},
   ): Promise<{ stdout: string; stderr: string }> {
+    await this.ensureInstalled();
```
alan-agius4:
```suggestion
    this.ensureInstalled();
```

alan-agius4 (approved):
LGTM, just a NIT

clydin:

- main: fe4c310049f8998e7daeeb2b8b94c1f4170c15e6

PR #32418, opened 2026-02-03
perf(@angular/cli): optimize package manager discovery with stat-based probing

clydin:
This refactors the package manager discovery logic to use 'stat' instead of 'readdir', significantly improving performance in directories with many files. It also simplifies the internal host abstraction by removing the now unused 'readdir' method.


clydin:

- main: bc363af8bc40f117a4e35ec9eb7eedf69f5b5b37

PR #32425, opened 2026-02-04
build: configure Jasmine to allow duplicate describe block names

alan-agius4:
This is needed as there are duplicate describe calls for the same builder, which is not allowed in Jasmine 6.


alan-agius4:

- main: 346973af499cd4bd23ebf3c8b37b498e51e0c1b8
- 21.1.x: d5a8314ab0cc6aaadd730cc4e795426ff98f73a1

PR #32427, opened 2026-02-04
feat(@angular/build): add headless option to unit-test builder

### LINKED ISSUE #31655: `unit-test` builder could align with Vitest's `headless` option
### Command

test

### Description

It would be nice if `unit-test` builder could take a `headless` option.
Separating browsers list and the `headless` option should provide a better DX.
For instance, one could run `ng test --headless` instead of using the `--browsers` option and providing the list of browser with the `Headless` suffix.

_No response_

The curent workaround is to use `CI=1 ng test`.

clydin:
This adds a new 'headless' option to the unit-test builder. When set to true, it forces all configured browsers to run in headless mode.

This option allows users to force headless execution in non-CI environments (where it is already the default) or to explicitly control the mode regardless of the browser name.

Informational messages are now logged when the option is redundant (e.g., all browsers are already headless), ignored (e.g., using the preview provider), or irrelevant (e.g., no browsers configured).

The 'karma' runner does not support this option and will log a warning if it is used.

Closes #31655

### DIFF
--- a/goldens/public-api/angular/build/index.api.md
+++ b/goldens/public-api/angular/build/index.api.md
@@ -231,6 +231,7 @@ export type UnitTestBuilderOptions = {
     dumpVirtualFiles?: boolean;
     exclude?: string[];
     filter?: string;
+    headless?: boolean;
     include?: string[];
     listTests?: boolean;
     outputFile?: string;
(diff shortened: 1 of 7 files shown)


clydin:

- main: ece30f2359c2dc794b0c9272447f623a121e88b0

PR #32442, opened 2026-02-05
fix(@angular/build): correctly resolve absolute setup file paths in Vitest

clydin:
The `resolveId` hook in the Vitest runner plugins now correctly handles absolute paths. Previously, the logic would blindly join the base directory with the file ID, which caused issues when the ID was already a fully qualified absolute path within the workspace. The updated logic now checks if the ID is absolute and located within the base directory before deciding whether to join paths, ensuring compatibility with both absolute paths and Vitest's short-form root-relative paths.


clydin:

- main: 8ae7f59e6f988489fda8c1346e3d2c3768d7a5f0
- 21.1.x: 7a9dd6b47e2191862c64355b10abaeead189759f

PR #32443, opened 2026-02-05
fix(@angular/build): explicitly fail when using Vitest runtime mocking

### LINKED ISSUE #31609: `unit-test` builder with `vitest` runner does not support `vi.mock()`
### Command

test

### Description

When using the new `unit-test` builder with Vitest runner, some Vitest features such as the module graph or module "mocking" with `vi.mock()` (https://vitest.dev/guide/mocking/modules.html) do not work.

This is due to the fact that the builder is building the tests and then providing the built chunks to Vitest so Vitest is not aware of the real module graph.

Luckily, Angular developers have the luxury of dependency injection that avoids relying on test framework for replacing things with test doubles.
That said, there are some use cases, specially when relying on 3rd party libraries that might require module "mocking".
For instance, AFAIK, the main approach to achieve Shallow testing in AOT without altering production code is to use this feature (cf. https://marmicode.io/blog/angular-template-code-coverage-and-future-proof-testing#:~:text=Shallow%20Testing%20is%20More%20Challenging)

Let vitest resolve modules 😇

In the mean time, it would be nice to monkey patch `vi.mock` and make it throw an error indicating that the feature is not supported.

clydin:
The Angular unit-test builder pre-bundles tests, which prevents Vitest's runtime mocking features (like `vi.mock`) from working correctly as they rely on module graph manipulation that occurs before bundling. To prevent confusion and silent failures, a patch is now injected that causes `vi.mock` and related methods to throw a descriptive error explaining the limitation and suggesting the use of Angular TestBed for mocking instead.

Closes #31609

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
@@ -113,6 +113,7 @@ export async function getVitestBuildOptions(
   }
 
   entryPoints.set('init-testbed', 'angular:test-bed-init');
+  entryPoints.set('vitest-mock-patch', 'angular:vitest-mock-patch');
 
   // The 'vitest' package is always external for testing purposes
   const externalDependencies = ['vitest'];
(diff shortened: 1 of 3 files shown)


clydin:

- main: fd5cb28c8082417288a896b89bde659bb0dc92e2

PR #32444, opened 2026-02-05
test: remove global Protractor/E2E setup from E2E utilities

clydin:
Since most E2E tests have been migrated away from Protractor, this change removes the automatic generation of the 'private-e2e' schematic and its associated configuration from the global 'prepareProjectForE2e' and 'useCIChrome' utilities. The setup is now explicitly handled within the 'browsers' test which still requires it, resulting in cleaner and faster initialization for the majority of the E2E suite.


alan-agius4 (approved):

clydin:

- main: c5c0ac4bb550f1e9fa856a1f247fbb3881bbb33e

PR #32468, opened 2026-02-10
test: execute 'protractor' binary directly in 'browsers' E2E test

clydin:
The 'browsers' E2E test now executes the 'protractor' binary directly from 'node_modules' instead of using the 'protractor' Angular CLI builder. This further reduces the test's dependency on legacy CLI infrastructure and removes the need to manually configure an 'e2e' architect target in 'angular.json'.


clydin:

- main: a391f04c77d0dbacb2bb6961be7bf34cf19ac7b6

Issue #32469, opened 2026-02-10
Allow setting preferred Vitest UI color scheme for Playwright


csvn:
### Command

test

### Description

The Vitest UI has an option to toggle Light/Dark mode on or off. But when using `@vitest/browser-playwright` then the users preference is always set to `light`, and does not respect OS preference. The big drawback here is that even if the UI button is toggled, it's reset the next time tests are started since it's not preserved. There's a way to configure this, but it requires a `vitest.config.js`: https://github.com/vitest-dev/vitest/issues/8947#issuecomment-3696478503

### Describe the solution you'd like

It would be great if Angular CLI could either allow configuring this default value or set it so it uses OS preference, when using the `"test.options.browser"` field.

- Angular CLI configures [`contextOptions.colorScheme` to be `null`](https://playwright.dev/docs/api/class-browser#browser-new-context-option-color-scheme) and thus `prefers-color-scheme` to be OS default
- An `angular.json` option to configure a new `uiTheme` to select light/dark/auto values

### Describe alternatives you've considered

Using a custom `vitest.config.js` configuration to manually set `playwright()` plugin options. But it feels a bit overkill for this A11y feature.

PR #32470, opened 2026-02-10
feat(@angular/build): run vitest browser with playwright with OS theme

csvn:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #32469

Using Vitest with `@vitest/browser-playwright` will open browser with `light` mode every time when running tests, even if theme toggle is used.

image

## What is the new behavior?

Using Vitest with `@vitest/browser-playwright` will open browser using `prefers-color-scheme` from OS over `light` default in Playwright.

- [x] Yes
- [ ] No

Setting the value like this affects the entire browser, and not just the Vitest UI. If any components with styles runs expecting `light` but getting `dark` via `prefers-color-scheme`, then tests may fail if tests rely on `light` being the default.

An alternative approach could be to only modify the `vueuse-color-scheme` for `localhost:<port>` via [`contextOptions.storageState`](https://playwright.dev/docs/api/class-browser#browser-new-context-option-storage-state) instead.

Another option would be to create an `angular.json` option for `@angular/build:unit-test` such that it's possible to opt-in to changing this default.

## Other information

I can create a test for this after feedback on which option would be best for this; current PR code, using `storageState`, or adding a `option` config value. 

Would it be best to modify the existing playwright test, or add a new one, when testing this? (https://github.com/angular/angular-cli/blob/main/tests/e2e/tests/vitest/browser-playwright.ts#L19)

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/browser-provider.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/browser-provider.ts
@@ -99,6 +99,17 @@ export async function setupBrowserConfiguration(
             launchOptions: {
               executablePath: process.env.CHROME_BIN,
             },
+            contextOptions: {
+              // Enables `prefer-color-scheme` for Vitest browser instead of `light`
+              colorScheme: null,
+            },
+          });
+        } else if (providerName === 'playwright') {
+          provider = providerFactory({
+            contextOptions: {
+              // Enables `prefer-color-scheme` for Vitest browser instead of `light`
+              colorScheme: null,
+            },
           });
         } else {
           provider = providerFactory();
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/browser-provider_spec.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/browser-provider_spec.ts
@@ -28,7 +28,7 @@ describe('setupBrowserConfiguration', () => {
     );
     await writeFile(
       join(playwrightPkgPath, 'index.js'),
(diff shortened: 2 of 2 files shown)


clydin:
The current approach appears reasonable.  For a test, you can copy the linked one above into a new file and augment as needed.

csvn:
@clydin I tried to copying `browser-no-globals.ts` and make a test that would use `matchMedia` to check if OS setting is respected (below), but I could not figure out how to run the test with a certain OS preference. There's sadly no way to check `(prefers-color-scheme: auto)`, only `light` or `dark` can be checked.

```ts
  await writeFile(
    'src/app/app.spec.ts',
    `
    import { test, expect } from 'vitest';

    test('should pass', () => {
      const mm = matchMedia('(prefers-color-scheme: dark)');
      expect(mm.matches).toBe(true);
    });
  `,
  );
```

For now, I added a test that checks that `contextOptions.colorScheme: null` is passed to `@vitest/browser-playwright`. I hope this looks alright.

clydin:
There's a formatting issue for `packages/angular/build/src/builders/unit-test/runners/vitest/browser-provider_spec.ts`.
Otherwise this looks good.

csvn:
> There's a formatting issue for `packages/angular/build/src/builders/unit-test/runners/vitest/browser-provider_spec.ts`. Otherwise this looks good.

Oops, sorry! 😅 I ran formatting and pushed the formatting changes.

clydin:

- main: cad7a7c0ff3778f04820a99ad0aa9d74f1067fd5

PR #32491, opened 2026-02-13
fix(@angular/ssr): introduce DI token to signal route discovery process

### LINKED ISSUE #32474: Allow detecting or skipping APP_INITIALIZERs during build-time route-extraction phase, with `outputMode: server`
### Command

build

### Description

When building an SSR app with `outputMode: 'server'`, Angular runs a "route extraction" phase to discover all routes for the manifest. The problem is this phase bootstraps our app and runs _all_ `APP_INITIALIZER`s - including ones that make HTTP calls, etc. which we don't want to run during the build but only during and actual live SSR or actual Prerendering.

These initializers make sense during:

- Live SSR (handling actual user requests)
- Prerendering (generating static HTML at build time) - if only we want to do the prerendering (which is not always the case when running the build!)

But during the build (and route extraction), we don't need (or want) to hit our APIs.

Side note: The route extraction phase is executed, even when `serverRoutes` have no route with `Prerender` mode.

**What we tried:**

We dug into the internals and found that `ɵSERVER_CONTEXT` from `@angular/platform-server` is the only way to tell these phases apart:

- Route extraction: `null`
- Prerendering: `'ssg'`
- Live SSR: `'ssr'`

But since it's a private API (the `ɵ` prefix), we can't rely on it. We're building an open-source Angular meta-framework for creating ecommerce sites. If we used Angular private API and Angular suddenly changes the private API anytime, we don't want our framework to break for our customers.

We also considered the public tokens `REQUEST`, `RESPONSE_INIT` and `REQUEST_CONTEXT`, but they are all `null` in both route extraction AND prerendering, so they can't help us distinguish between the two, in case we'd like to run an actual prerendering sometimes.

__Option A: Let us detect the context__

Expose a public token like `SERVER_RENDER_CONTEXT` (or make `ɵSERVER_CONTEXT` public) with values like:

- `'route-extraction'` - build-time route discovery
- `'ssg'` - prerendering
- `'ssr'` - live server rendering

Then we could do:

```typescript
provideAppInitializer(() => {
  const context = inject(SERVER_RENDER_CONTEXT, { optional: true });
  if (context === 'route-extraction') {
    return; // Skip
  }
  // Do the work for actual SSG/SSR
})
```

__Option B: Skip APP_INITIALIZERs during route extraction__

Have Angular skip running `APP_INITIALIZER`s entirely during route extraction, since:

- Route extraction only needs the Router config
- It already sets `_ENABLE_ROOT_COMPONENT_BOOTSTRAP: false` internally
- Not running initializers would be consistent with that "lightweight" approach

Though it might be risky - some apps might have APP_INITIALIZERS that configure routes dynamically.

1. __Using `ɵSERVER_CONTEXT`__ - Works but it's private API, can't use it in a our meta-framework we distribute

2. __Checking `REQUEST` token__ - Both route extraction and prerendering have `REQUEST: null`, so no way to tell them apart 🙁

alan-agius4:
A new DI token, `IS_DISCOVERING_ROUTES`, is introduced to provide a clear signal for when the application is operating in route discovery mode. This token is provided with the value `true` within the route extraction providers. Other services and components can inject this token to conditionally alter their behavior, for instance, to disable functionality that is not required or could interfere with the route discovery process.

Closes #32474

### DIFF
--- a/packages/angular/ssr/src/routes/ng-routes.ts
+++ b/packages/angular/ssr/src/routes/ng-routes.ts
@@ -11,6 +11,7 @@ import {
   ApplicationRef,
   Compiler,
   EnvironmentInjector,
+  InjectionToken,
   Injector,
   createEnvironmentInjector,
   runInInjectionContext,
@@ -23,6 +24,7 @@ import {
   Router,
   ɵloadChildren as loadChildrenHelper,
 } from '@angular/router';
+
 import { ServerAssets } from '../assets';
 import { Console } from '../console';
 import { AngularAppManifest, getAngularAppManifest } from '../manifest';
@@ -38,6 +40,22 @@ import {
   ServerRoutesConfig,
 } from './route-config';
 import { RouteTree, RouteTreeNodeMetadata } from './route-tree';
+
+/**
+ * A DI token that indicates whether the application is in the process of discovering routes.
+ *
+ * This token is provided with the value `true` when route discovery is active, allowing other
+ * parts of the application to conditionally execute logic. For example, it can be used to
+ * disable features or behaviors that are not necessary or might interfere with the route
+ * discovery process.
+ */
+export const IS_DISCOVERING_ROUTES = new InjectionToken<boolean>(
+  typeof ngDevMode === 'undefined' || ngDevMode ? 'IS_DISCOVERING_ROUTES' : '',
+  {
+    providedIn: 'platform',
+    factory: () => false,
+  },
+);
 
 interface Route extends AngularRoute {
   ɵentryName?: string;
@@ -622,6 +640,10 @@ export async function getRoutesFromAngularRouterConfig(
     {
       provide: ɵENABLE_ROOT_COMPONENT_BOOTSTRAP,
       useValue: false,
+    },
+    {
+      provide: IS_DISCOVERING_ROUTES,
+      useValue: true,
     },
   ]);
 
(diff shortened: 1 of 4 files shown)


packages/angular/ssr/src/routes/ng-routes.ts
```diff
@@ -39,6 +41,16 @@ import {
...
 
+/**
+ * A DI token that indicates whether the application is in the process of discovering routes.
+ *
+ * This token is provided with the value `true` when route discovery is active, allowing other
+ * parts of the application to conditionally execute logic. For example, it can be used to
+ * disable features or behaviors that are not necessary or might interfere with the route
+ * discovery process.
+ */
+export const DURING_ROUTE_DISCOVERY = new InjectionToken<boolean>('DURING_ROUTE_DISCOVERY');
```
eneajaho:
```suggestion
export const DURING_ROUTE_DISCOVERY = new InjectionToken<boolean>('DURING_ROUTE_DISCOVERY', {
factory: () => false
});
```

Should this have a default value? As this will run in browser mode too I guess ?

alan-agius4: Ohhh. Good catch.

dgp1130: Consider: The naming feels a little awkward to me to start with "during". I wonder if `IS_DISCOVERING_ROUTES` or `IS_BOOTSTRAPPING_FOR_ROUTES` or something like that might be more clear as a boolean value? Up to you in the end.

Platonn:
Hi Team, since the PR was approved last week, is there a chance we could squeeze it into 21.2.0?

alan-agius4:

- main: 0f41585a7c2372e31c4e68e0c7442692b75a3dae
- 21.2.x: dee3717b3faae9ea75d0a5e53c925f915949b8d0

PR #32492, opened 2026-02-13
feat(@angular/build): support runtime Zone.js detection in Vitest unit test runner

clydin:
This commit improves how the Vitest unit test runner handles Zone.js and its testing polyfills. Previously, the inclusion of provideZoneChangeDetection and zone.js/testing was determined solely by a build-time check for zone.js in the polyfills array.

Now, the runner supports three strategies for zone.js/testing inclusion:
- none: If zone.js is not installed in the project.
- static: If zone.js is explicitly included in the polyfills build option.
- dynamic: If zone.js is installed but not explicitly in polyfills. This uses a runtime check and dynamic import to load testing support if Zone is present.

Additionally, TestBed initialization now dynamically provides ZoneChangeDetection based on the runtime presence of Zone.js, better supporting zoneless applications and implicit Zone.js loading scenarios.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
@@ -6,11 +6,12 @@
  * found in the LICENSE file at https://angular.dev/license
  */
 
+import { createRequire } from 'node:module';
 import path from 'node:path';
 import { toPosixPath } from '../../../../utils/path';
 import type { ApplicationBuilderInternalOptions } from '../../../application/options';
 import { OutputHashing } from '../../../application/schema';
-import { NormalizedUnitTestBuilderOptions, injectTestingPolyfills } from '../../options';
+import { NormalizedUnitTestBuilderOptions } from '../../options';
 import { findTests, getTestEntrypoints } from '../../test-discovery';
 import { RunnerOptions } from '../api';
 
@@ -18,9 +19,8 @@ function createTestBedInitVirtualFile(
   providersFile: string | undefined,
   projectSourceRoot: string,
   teardown: boolean,
-  polyfills: string[] = [],
+  zoneTestingStrategy: 'none' | 'static' | 'dynamic',
 ): string {
-  const usesZoneJS = polyfills.includes('zone.js');
   let providersImport = 'const providers = [];';
   if (providersFile) {
     const relativePath = path.relative(projectSourceRoot, providersFile);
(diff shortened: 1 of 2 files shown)


packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
```diff
@@ -145,13 +161,30 @@ export async function getVitestBuildOptions(
...
+
+  if (isZoneJsInstalled) {
+    if (buildOptions.polyfills?.includes('zone.js/testing')) {
+      zoneTestingStrategy = 'none';
+    } else if (buildOptions.polyfills?.includes('zone.js')) {
+      zoneTestingStrategy = 'static';
+    } else {
+      zoneTestingStrategy = 'dynamic';
+    }
+  }
```
alan-agius4: In this case `zone.js` is listed in the polyfills, it seems redundant to check if it's installed. 

```suggestion
  // Inject the zone.js testing polyfill if Zone.js is installed.
  let zoneTestingStrategy: 'none' | 'static' | 'dynamic';
  if (buildOptions.polyfills?.includes('zone.js/testing')) {
    zoneTestingStrategy = 'none';
  } else if (buildOptions.polyfills?.includes('zone.js')) {
    zoneTestingStrategy = 'static';
  } else {
    try {
      const projectRequire = createRequire(path.join(projectSourceRoot, 'package.json'));
      projectRequire.resolve('zone.js');
      zoneTestingStrategy = 'dynamic';
    } catch {
      zoneTestingStrategy = 'none';
    }
  }
```

packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
```diff
@@ -145,13 +161,30 @@ export async function getVitestBuildOptions(
     externalDependencies,
   };
 
-  buildOptions.polyfills = injectTestingPolyfills(buildOptions.polyfills);
+  // Inject the zone.js testing polyfill if Zone.js is installed.
+  let zoneTestingStrategy: 'none' | 'static' | 'dynamic' = 'none';
+  let isZoneJsInstalled = false;
+  try {
+    const projectRequire = createRequire(path.join(projectSourceRoot, 'package.json'));
```
alan-agius4: Is this by intention to use the `projectSourceRoot` instead of the `workspaceRoot`?

clydin: Intentional for potential monorepo scenarios where the dependencies may be inside the project itself.

clydin:

- main: 414320d02a090b7b36720051569cba7563bf9ac6

PR #32499, opened 2026-02-17
fix(@angular/ssr): validate host headers to prevent header-based SSRF [main]

alan-agius4:
This change introduces strict validation for `Host`, `X-Forwarded-Host`, `X-Forwarded-Proto`, and `X-Forwarded-Port` headers in the Angular SSR request handling pipeline, including `CommonEngine` and `AngularAppEngine`.

Previously, the application engine constructed the base URL for server-side rendering using these headers without validation. This could allow an attacker to manipulate the headers to steer relative `HttpClient` requests to arbitrary internal or external hosts (SSRF).

With this change:
- The `Host` and `X-Forwarded-Host` headers are validated against a strict allowlist.
- `localhost` and loopback addresses (e.g., `127.0.0.1`, `[::1]`) are allowed by default.
- `X-Forwarded-Port` must be numeric.
- `X-Forwarded-Proto` must be `http` or `https`.
- Requests with invalid or disallowed headers will now be rejected with a `400 Bad Request` status code.

BREAKING CHANGE:

Server-side requests will now fail with a `400 Bad Request` error if the `Host` header does not match a customized allowlist (or localhost).

**AngularAppEngine Users:**
To resolve this, you must configure the `allowedHosts` option in your `angular.json` to include all domain names where your application is deployed.

Example configuration in `angular.json`:

```json
"architect": {
  "build": {
    "options": {
      "security": {
        "allowedHosts": ["example.com", "*.trusted-example.com"]
      }
    }
  }
}
```

**CommonEngine Users:**
If you are using `CommonEngine`,  you must now provide the `allowedHosts` option when initializing or rendering your application.

Example:
```typescript
const commonEngine = new CommonEngine({
  allowedHosts: [“example.com”, “*.trusted-example.com"]
});
```

### DIFF
--- a/packages/angular/build/src/builders/application/schema.json
+++ b/packages/angular/build/src/builders/application/schema.json
@@ -53,7 +53,7 @@
       "additionalProperties": false,
       "properties": {
         "allowedHosts": {
-          "description": "A list of hosts that are allowed to access the server-side application.",
+          "description": "A list of hostnames that are allowed to access the server-side application. For more information, see https://angular.dev/guide/ssr#configuring-allowed-hosts.",
           "type": "array",
           "uniqueItems": true,
           "items": {
--- a/packages/angular/build/src/utils/server-rendering/manifest.ts
+++ b/packages/angular/build/src/utils/server-rendering/manifest.ts
@@ -86,11 +86,7 @@ export function generateAngularServerAppEngineManifest(
   const manifestContent = `
 export default {
   basePath: '${basePath}',
-  allowedHosts: ${JSON.stringify(
-    allowedHosts.map((host) => host.replace(/^www\./i, '')),
-    undefined,
-    2,
-  )},
+  allowedHosts: ${JSON.stringify(allowedHosts, undefined, 2)},
   supportedLocales: ${JSON.stringify(supportedLocales, undefined, 2)},
   entryPoints: {
     ${Object.entries(entryPoints)
--- a/packages/angular/ssr/node/src/common-engine/common-engine.ts
+++ b/packages/angular/ssr/node/src/common-engine/common-engine.ts
@@ -12,6 +12,7 @@ import { renderApplication, renderModule, ɵSERVER_CONTEXT } from '@angular/plat
 import * as fs from 'node:fs';
 import { dirname, join, normalize, resolve } from 'node:path';
 import { URL } from 'node:url';
+import { isHostAllowed } from '../../../src/utils/validation';
 import { attachNodeGlobalErrorHandlers } from '../errors';
 import { CommonEngineInlineCriticalCssProcessor } from './inline-css-processor';
 import {
@@ -19,11 +20,6 @@ import {
   printPerformanceLogs,
   runMethodAndMeasurePerf,
 } from './peformance-profiler';
-
-/**
- * Regular expression to match and remove the `www.` prefix from hostnames.
- */
-const WWW_HOST_REGEX = /^www\./i;
 
 const SSG_MARKER_REGEXP = /ng-server-context=["']\w*\|?ssg\|?\w*["']/;
 
@@ -75,13 +71,7 @@ export class CommonEngine {
   private readonly allowedHosts: ReadonlySet<string>;
 
   constructor(private options: CommonEngineOptions) {
-    this.allowedHosts = new Set([
-      ...options.allowedHosts.map((host) => host.replace(WWW_HOST_REGEX, '')),
-      'localhost',
-      '127.0.0.1',
-      '::1',
-      '[::1]',
-    ]);
+    this.allowedHosts = new Set(options.allowedHosts);
 
     attachNodeGlobalErrorHandlers();
   }
@@ -128,27 +118,12 @@ export class CommonEngine {
       throw new Error(`URL "${url}" is invalid.`);
     }
 
-    const hostname = new URL(url).hostname.replace(WWW_HOST_REGEX, '');
-
-    if (this.allowedHosts.has(hostname)) {
-      return;
+    const hostname = new URL(url).hostname;
+    if (!isHostAllowed(hostname, this.allowedHosts)) {
+      throw new Error(
+        `Host ${hostname} is not allowed. Please provide a list of allowed hosts in the "allowedHosts" option.`,
+      );
     }
-
-    // Support wildcard hostnames.
-    for (const allowedHost of this.allowedHosts) {
-      if (!allowedHost.startsWith('*.')) {
-        continue;
-      }
-
-      const domain = allowedHost.slice(1);
-      if (hostname.endsWith(domain)) {
-        return;
-      }
-    }
-
-    throw new Error(
-      `Host ${hostname} is not allowed. Please provide a list of allowed hosts in the "allowedHosts" option.`,
-    );
   }
 
   private inlineCriticalCss(html: string, opts: CommonEngineRenderOptions): Promise<string> {
--- a/packages/angular/ssr/src/utils/validation.ts
+++ b/packages/angular/ssr/src/utils/validation.ts
@@ -17,11 +17,6 @@ const VALID_PORT_REGEX = /^\d+$/;
 const VALID_PROTO_REGEX = /^https?$/i;
 
 /**
- * Regular expression to match and remove the `www.` prefix from hostnames.
- */
-const WWW_HOST_REGEX = /^www\./i;
-
-/**
  * Regular expression to match path separators.
  */
 const PATH_SEPARATOR_REGEX = /[/\\]/;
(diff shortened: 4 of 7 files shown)


packages/angular/build/src/builders/application/schema.json
```diff
@@ -52,6 +52,14 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "allowedHosts": {
+          "description": "A list of hosts that are allowed to access the server-side application.",
```
dgp1130: Consider: Could we be more clear about the definition of `host`? Maybe indicating that it is the list of allowed `Host` request headers which will be accepted?

packages/angular/build/src/utils/server-rendering/manifest.ts
```diff
@@ -84,6 +86,11 @@ export function generateAngularServerAppEngineManifest(
   const manifestContent = `
 export default {
   basePath: '${basePath}',
+  allowedHosts: ${JSON.stringify(
+    allowedHosts.map((host) => host.replace(/^www\./i, '')),
```
dgp1130: Question: What's the motivation for stripping `www.`? Is that subdomain actually treated in a special way?

dgp1130: Question: Do we also need to include loopback addresses like we have in `CommonEngine`, or is that already handled separately?

packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -102,6 +123,34 @@ export class CommonEngine {
...
+    }
+
+    // Support wildcard hostnames.
+    for (const allowedHost of this.allowedHosts) {
+      if (!allowedHost.startsWith('*.')) {
+        continue;
+      }
+
+      const domain = allowedHost.slice(1);
+      if (hostname.endsWith(domain)) {
```
dgp1130: Consider: Should we document / state somewhere that `foo.*.example.test` is not supported?

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,156 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.dev/license
+ */
+
+/**
+ * Regular expression to validate that the port is a numeric value.
+ */
+const VALID_PORT_REGEX = /^\d+$/;
```
dgp1130, line 12: Nit: Do we care to limit to [0, 65,535]? Seems like less of a parsing/security issue if the port is too large but still numeric, but I also would expect the server to reject `X-Forwarded-Port: 99999999999` at some layer.

packages/angular/ssr/src/utils/headers.ts
```diff
@@ -0,0 +1,156 @@
...
+ * It ensures that the hostnames match the allowed hosts and that ports and protocols adhere to expected formats.
+ *
+ * @param request - The incoming `Request` object containing the headers to validate.
+ * @param allowedHosts - A set of allowed hostnames.
+ * @throws Error if any of the validated headers contain invalid values.
+ */
+export function validateHeaders(request: Request, allowedHosts: ReadonlySet<string>): void {
+  const headers = request.headers;
+  validateHost('x-forwarded-host', headers, allowedHosts);
+  validateHost('host', headers, allowedHosts);
```
dgp1130: Question: Do we need to assert that `Host` exists? Can we assume the server runtime already validated that part? Is there a risk we receive a request with neither and silently skip validation?

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,156 @@
...
+ */
+function validateHost(
+  headerName: string,
+  headers: Headers,
+  allowedHosts: ReadonlySet<string>,
+): void {
+  const value = getFirstHeaderValue(headers.get(headerName))?.replace(WWW_HOST_REGEX, '');
+  if (!value) {
+    return;
+  }
```
dgp1130, line 105: Consider: I'm a little worried something like `Host: www.` might trick this into being an empty string and causing it to skip validation? Would it be safer to check if `headers` contains `headerName` and skip processing only when the header is explicitly missing? Then we can follow up with stripping `www.` after.

packages/angular/ssr/src/utils/headers.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  const { hostname } = new URL(url);
+  if (
+    // Check the provided allowed hosts first.
+    allowedHosts.has(hostname) ||
+    checkWildcardHostnames(hostname, allowedHosts) ||
+    // Check the default allowed hosts last this is the fallback and should be rarely if ever used in production.
+    DEFAULT_ALLOWED_HOSTS.has(hostname) ||
+    checkWildcardHostnames(hostname, DEFAULT_ALLOWED_HOSTS)
+  ) {
```
dgp1130: Consider: Refactor this into `hostIsAllowed(hostname, allowedHosts)` which covers all four cases.

packages/angular/ssr/src/utils/headers.ts
```diff
@@ -0,0 +1,156 @@
...
+    return;
+  }
+
+  let errorMessage = `Header "${headerName}" with value "${value}" is not allowed.`;
+  if (typeof ngDevMode === 'undefined' || ngDevMode) {
+    errorMessage +=
+      '\n\nAction Required: Update your "angular.json" to include this hostname. ' +
+      'Path: "projects.[project-name].architect.build.options.security.allowedHosts".';
```
dgp1130: Suggestion: This is implicitly stating that `value` is an acceptable host, but it may be correctly blocked as malicious and we wouldn't want to guide users to blindly allowing any hosts which try to connect.

We should suggest this only if `value` is an expected host. Maybe something like:

> Header "`${headerName}`" with value "`${value}`" is not allowed.
>
> Possible Action Required: If your server is intended to be hosted at "`${value}`", update your "angular.json" to include this hostname. Path: "projects.[project-name].architect.build.options.security.allowedHosts". If your server is not hosted at "`${value}`", then this error can be safely ignored.

Also can we link to any docs for `allowedHosts` once that lands (maybe after this PR)?

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/**
+ * Checks if the hostname matches any of the wildcard hostnames in the allowlist.
+ * @param hostname - The hostname to check.
+ * @param allowedHosts - A set of allowed hostnames.
+ * @returns `true` if the hostname matches any of the wildcard hostnames, `false` otherwise.
+ */
+function checkWildcardHostnames(hostname: string, allowedHosts: ReadonlySet<string>): boolean {
+  for (const allowedHost of allowedHosts) {
```
dgp1130, line 155: Nitpicky perf thing: We're iterating over all of `allowedHosts` for every request. We could at least find all the wildcard hosts at startup and though only iterate through those for each request.

Probably small enough work being done here to be insignificant, up to you if it's worth the effort.

packages/angular/ssr/src/utils/validation.ts
dgp1130, line 1: Consider: Is there anything we can do to unify the two header validation implementations? I'm a bit worried we end up forgetting one as we make changes to the other.

Simplest answer is probably a link in `@fileoverview`, but that's easy to miss. Anything worth doing now to mitigate that risk?

dgp1130 (approved):
Left a handful of suggestions. Don't worry about backporting anything not critical to your other PRs, I'm less concerned about maintainability, documentation, etc. aspects there, as long as `main` is in a good state going forward.

Also feel free to skip or come back to the less important comments in future PRs, no need to block this on more than we need to.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: We do not really care about the numeric value here, we just want to ensure that there are no path injections.

packages/angular/build/src/utils/server-rendering/manifest.ts
alan-agius4: Actually let me remove this.

alan-agius4: It's already handled. (check the constructor)

packages/angular/ssr/node/src/common-engine/common-engine.ts
alan-agius4: Maybe in the docs.

packages/angular/ssr/src/utils/headers.ts
alan-agius4: What you mean by the host exists?

packages/angular/ssr/src/utils/validation.ts
alan-agius4: I feel like this list of hosts would not be that large that this sort of micro optimization would have any benefit from.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: Good catch.

packages/angular/ssr/src/utils/headers.ts
dgp1130: I'm asking if there is a case where both `req.headers.host` and `req.headers['x-forwarded-host']` are both `undefined`? If that happened, we would just skip any validation.

I think `host` is the one required header in HTTP, but I'm not sure at what layer that's enforced when we have to account for malformed requests.

alan-agius4:
Superseded by https://github.com/angular/angular-cli/pull/32516

PR #32516, opened 2026-02-19
fix(@angular/ssr): validate host headers to prevent header-based SSRF

alan-agius4:
This change introduces strict validation for `Host`, `X-Forwarded-Host`, `X-Forwarded-Proto`, and `X-Forwarded-Port` headers in the Angular SSR request handling pipeline, including `CommonEngine` and `AngularAppEngine`.

Previously, the application engine constructed the base URL for server-side rendering using these headers without validation. This could allow an attacker to manipulate the headers to steer relative `HttpClient` requests to arbitrary internal or external hosts (SSRF).

The validation rules are:

- `Host` and `X-Forwarded-Host` headers are validated against an authorized allowlist. 
- `Host` and `X-Forwarded-Host` headers cannot contain path separators.
- `X-Forwarded-Port` header must be numeric.
- `X-Forwarded-Proto` header must be `http` or `https`.

Requests with invalid or disallowed headers will now log an error and fallback to Client-Side Rendering (CSR). In a future major version, these requests will be rejected with a 400 Bad Request.

Note: Most cloud providers and CDNs already validate these headers before the request reaches the text application, but this change adds an essential layer of defense-in-depth.

**AngularAppEngine Users:**
To exclude safe hosts from validation, configure the `allowedHosts` option in `angular.json` to include all domain names where your application is deployed.

Example configuration in `angular.json`:

```json
"architect": {
  "build": {
    "options": {
      "security": {
        "allowedHosts": ["example.com", "*.trusted-example.com"]
      }
    }
  }
}
```

or

```ts
const appEngine = new AngularAppEngine({
  allowedHosts: ["example.com", "*.trusted-example.com"]
})

const appEngine = new AngularNodeAppEngine({
  allowedHosts: ["example.com", "*.trusted-example.com"]
})
```

**CommonEngine Users:**
If you are using `CommonEngine`, you must now provide the `allowedHosts` option when initializing or rendering your application.

Example:
```typescript
const commonEngine = new CommonEngine({
  allowedHosts: ["example.com", "*.trusted-example.com"]
});
```

The application also respects `NG_ALLOWED_HOSTS` (comma-separated list) and `HOSTNAME` environment variables for authorizing hosts.

### DIFF
--- a/packages/angular/ssr/node/src/common-engine/common-engine.ts
+++ b/packages/angular/ssr/node/src/common-engine/common-engine.ts
@@ -99,7 +99,7 @@ export class CommonEngine {
             'Please provide a list of allowed hosts in the "allowedHosts" option in the "CommonEngine" constructor.',
           isAllowedHostConfigured
             ? ''
-            : '\nFallbacking to client side rendering. This will become a 400 Bad Request in a future major version.',
+            : '\nFalling back to client side rendering. This will become a 400 Bad Request in a future major version.',
         );
 
         if (!isAllowedHostConfigured) {
--- a/packages/angular/ssr/src/app-engine.ts
+++ b/packages/angular/ssr/src/app-engine.ts
@@ -270,7 +270,7 @@ export class AngularAppEngine {
         errorMessage +
         (isAllowedHostConfigured
           ? ''
-          : '\nFallbacking to client side rendering. This will become a 400 Bad Request in a future major version.') +
+          : '\nFalling back to client side rendering. This will become a 400 Bad Request in a future major version.') +
         '\n\nFor more information, see https://angular.dev/best-practices/security#preventing-server-side-request-forgery-ssrf',
     );
 
--- a/packages/angular/ssr/src/utils/validation.ts
+++ b/packages/angular/ssr/src/utils/validation.ts
@@ -93,30 +93,86 @@ export function cloneRequestAndPatchHeaders(
   });
 
   const headers = clonedReq.headers;
-  const originalHeadersGet = headers.get;
 
+  const originalGet = headers.get;
   // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-  (headers.get as typeof originalHeadersGet) = (name: string): string | null => {
-    const value = originalHeadersGet.call(headers, name);
+  (headers.get as typeof originalGet) = function (name) {
+    const value = originalGet.call(headers, name);
     if (!value) {
       return value;
     }
 
-    const key = name.toLowerCase();
-    if (HOST_HEADERS_TO_VALIDATE.has(key)) {
-      try {
-        verifyHostAllowed(key, value, allowedHosts);
-      } catch (error) {
-        onError(error as Error);
-
-        return null;
-      }
-    }
+    validateHeader(name, value, allowedHosts, onError);
 
     return value;
   };
 
+  const originalValues = headers.values;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+  (headers.values as typeof originalValues) = function () {
+    for (const name of HOST_HEADERS_TO_VALIDATE) {
+      validateHeader(name, originalGet.call(headers, name), allowedHosts, onError);
+    }
+
+    return originalValues.call(headers);
+  };
+
+  const originalEntries = headers.entries;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+  (headers.entries as typeof originalEntries) = function () {
+    const iterator = originalEntries.call(headers);
+
+    return {
+      next() {
+        const result = iterator.next();
+        if (!result.done) {
+          const [key, value] = result.value;
+          validateHeader(key, value, allowedHosts, onError);
+        }
+
+        return result;
+      },
+      [Symbol.iterator]() {
+        return this;
+      },
+    };
+  };
+
+  // Ensure for...of loops use the new patched entries
+  (headers[Symbol.iterator] as typeof originalEntries) = headers.entries;
+
   return { request: clonedReq, onError: onErrorPromise };
+}
+
+/**
+ * Validates a specific header value against the allowed hosts.
+ * @param name - The name of the header to validate.
+ * @param value - The value of the header to validate.
+ * @param allowedHosts - A set of allowed hostnames.
+ * @param onError - A callback function to call if the header value is invalid.
+ * @throws Error if the header value is invalid.
+ */
+function validateHeader(
+  name: string,
+  value: string | null,
+  allowedHosts: ReadonlySet<string>,
+  onError: (value: Error) => void,
+): void {
+  if (!value) {
+    return;
+  }
+
+  if (!HOST_HEADERS_TO_VALIDATE.has(name.toLowerCase())) {
+    return;
+  }
+
+  try {
+    verifyHostAllowed(name, value, allowedHosts);
+  } catch (error) {
+    onError(error as Error);
+
+    throw error;
+  }
 }
 
 /**
--- a/packages/angular/ssr/BUILD.bazel
+++ b/packages/angular/ssr/BUILD.bazel
@@ -20,7 +20,7 @@ ts_project(
     ),
     args = [
         "--lib",
-        "dom,es2022",
+        "dom.iterable,dom,es2022",
     ],
     data = [
         "//packages/angular/ssr/third_party/beasties:beasties_bundled",
--- a/packages/angular/ssr/test/utils/validation_spec.ts
+++ b/packages/angular/ssr/test/utils/validation_spec.ts
@@ -135,7 +135,9 @@ describe('Validation Utils', () => {
       });
       const { request: secured, onError } = cloneRequestAndPatchHeaders(req, allowedHosts);
 
-      expect(secured.headers.get('host')).toBeNull();
+      expect(() => secured.headers.get('host')).toThrowError(
+        'Header "host" with value "evil.com" is not allowed.',
+      );
       await expectAsync(onError).toBeResolvedTo(
         jasmine.objectContaining({
           message: jasmine.stringMatching('Header "host" with value "evil.com" is not allowed'),
@@ -157,7 +159,9 @@ describe('Validation Utils', () => {
       });
       const { request: secured, onError } = cloneRequestAndPatchHeaders(req, allowedHosts);
 
-      expect(secured.headers.get('x-forwarded-host')).toBeNull();
+      expect(() => secured.headers.get('x-forwarded-host')).toThrowError(
+        'Header "x-forwarded-host" with value "evil.com" is not allowed.',
+      );
       await expectAsync(onError).toBeResolvedTo(
         jasmine.objectContaining({
           message: jasmine.stringMatching(
@@ -174,6 +178,63 @@ describe('Validation Utils', () => {
       const { request: secured } = cloneRequestAndPatchHeaders(req, allowedHosts);
 
       expect(secured.headers.get('accept')).toBe('application/json');
+    });
+
+    it('should validate headers when iterating with entries()', async () => {
+      const req = new Request('http://example.com', {
+        headers: { 'host': 'evil.com' },
+      });
+      const { request: secured, onError } = cloneRequestAndPatchHeaders(req, allowedHosts);
+
+      expect(() => {
+        for (const _ of secured.headers.entries()) {
+          // access the header to trigger the validation
+        }
+      }).toThrowError('Header "host" with value "evil.com" is not allowed.');
+
+      await expectAsync(onError).toBeResolvedTo(
+        jasmine.objectContaining({
+          message: jasmine.stringMatching('Header "host" with value "evil.com" is not allowed.'),
+        }),
+      );
+    });
+
+    it('should validate headers when iterating with values()', async () => {
+      const req = new Request('http://example.com', {
+        headers: { 'host': 'evil.com' },
+      });
+      const { request: secured, onError } = cloneRequestAndPatchHeaders(req, allowedHosts);
+
+      expect(() => {
+        for (const _ of secured.headers.values()) {
+          // access the header to trigger the validation
+        }
+      }).toThrowError('Header "host" with value "evil.com" is not allowed.');
+
+      await expectAsync(onError).toBeResolvedTo(
+        jasmine.objectContaining({
+          message: jasmine.stringMatching('Header "host" with value "evil.com" is not allowed.'),
+        }),
+      );
+    });
+
+    it('should validate headers when iterating with for...of', async () => {
+      const req = new Request('http://example.com', {
+        headers: { 'host': 'evil.com' },
+      });
+      const { request: secured, onError } = cloneRequestAndPatchHeaders(req, allowedHosts);
+
+      expect(() => {
+        for (const _ of secured.headers) {
+          // access the header to trigger the validation
+        }
+      }).toThrowError('Header "host" with value "evil.com" is not allowed.');
+
+      await expectAsync(onError).toBeResolvedTo(
+        jasmine.objectContaining({
+          message: jasmine.stringMatching('Header "host" with value "evil.com" is not allowed.'),
+        }),
+      );
     });
   });
 });


packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -186,6 +212,34 @@ export class CommonEngine {
 
     return doc;
   }
+
+  /**
+   * Resolves the allowed hosts from the provided options and environment variables.
+   * @param options Options for the common engine.
+   * @returns A set of allowed hosts.
+   */
+  private resolveAllowedHosts(options: CommonEngineOptions | undefined): ReadonlySet<string> {
```
dgp1130: Nit: I recommend making this `static` or a standalone function rather than calling a method in a constructor, when the class might not be fully initialized, as this can lead to unexpected behavior when accessing `this.` properties which haven't been initialized yet.

packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -74,6 +81,25 @@ export class CommonEngine {
...
+
+    if (url && URL.canParse(url)) {
+      const urlObj = new URL(url);
+      try {
+        validateUrl(urlObj, this.allowedHosts);
+      } catch (error) {
+        let document = opts.document;
+        if (!document && opts.documentFilePath) {
+          document = await this.getDocument(opts.documentFilePath);
+        }
```
dgp1130: Question: What's the purpose of reading the `document` here? Why do we need this?

packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -186,6 +212,34 @@ export class CommonEngine {
...
+        if (hostTrimmed) {
+          allowedHosts.add(hostTrimmed);
+        }
+      }
+    }
+
+    const envHostName = processEnv['HOSTNAME']?.trim();
+    if (envHostName) {
+      allowedHosts.add(envHostName);
+    }
```
dgp1130: Question: Is there any standard/precedent for `${HOSTNAME}`, or are you just making up a name here? Are we aware of any deployment infra which provides this name already?

packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -186,6 +212,34 @@ export class CommonEngine {
...
+    const envNgAllowedHosts = processEnv['NG_ALLOWED_HOSTS'];
+    if (envNgAllowedHosts) {
+      const hosts = envNgAllowedHosts.split(',');
+      for (const host of hosts) {
+        const hostTrimmed = host.trim();
+        if (hostTrimmed) {
+          allowedHosts.add(hostTrimmed);
+        }
+      }
+    }
```
dgp1130: Consider: Do you want to tackle this now, as something we'll probably have to support for a few versions, or leave it off and require `allowedHosts` to be specified in the build options in the short term until we come back to this in v22+?

I agree this is probably a good direction, but I'm wondering if it might benefit from more discussion, and I don't think we _need_ to have it right now. Would it make sense to hold off before committing to a new public API here?

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -21,32 +27,86 @@ import { createWebRequestFromNodeRequest } from './request';
...
+        if (hostTrimmed) {
+          allowedHosts.push(hostTrimmed);
+        }
+      }
+    }
+
+    const envHostName = processEnv['HOSTNAME']?.trim();
+    if (envHostName) {
+      allowedHosts.push(envHostName);
+    }
```
dgp1130: Suggestion: Can we consolidate and share the logic for resolving allowed hosts somewhere? They're both used in `@angular/ssr`, so I feel like we should be able to move this into a separate file.

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,224 @@
...
+    get(target, prop) {
+      if (prop !== 'headers') {
+        const value = Reflect.get(target, prop);
+
+        return typeof value === 'function' ? value.bind(target) : value;
+      }
+
+      const headersTarget = target.headers;
+
+      return new Proxy(headersTarget, {
```
dgp1130: Consider: I know there's nothing we can do here which will be _fully_ transparent and unobservable, but I'm wondering if there might be less churn if we overwrote specifically the `req.headers['host' | 'x-forwarded-host']` properties with a getter? Something like:

```typescript
for (const header of ['host', 'x-forwarded-host']) {
  const value = req.headers[header];
  delete req.headers[header];
  Object.defineProperty(req.headers, header, {
    get: () => {
      verifyHostAllowed(value, allowedHosts);
      return value;
    },
  });
}
```

Would that potentially be less likely to break things like `instanceof` / `constructor` checks and/or be more performant? Might also be less likely to break in more unique situations as `Proxy` is a pretty complex API and I'm less than certain that we're covering all the edge cases here.

Related: Even if we do stick with `Proxy`, do we need two of them? Could we make a proxy for `headers` and just assign it to `Request.prototype.headers`?

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,224 @@
...
+          if (headersProperty === 'get') {
+            return (name: string) => {
+              const value = headersTarget.get(name);
+              if (!value) {
+                return value;
+              }
+
+              const key = name.toLowerCase();
+              if (HOST_HEADERS_TO_VALIDATE.has(key)) {
+                verifyHostAllowed(key, value, allowedHosts);
```
dgp1130: Observation: I could see this failing some apps who iterate through headers like:

```typescript
for (const [name, value] of Object.entries(req.headers)) {
  if (!name.startsWith('x-myapp-*')) continue;

  doSomethingWith(name, value);
}
```

In this case, they're not _really_ using `host` or `x-forwarded-host`, but they will read the values and probably trigger the CSR deopt. I don't think there's anything we can meaningfully do about this, but pointing out the case of iterating through all the headers which isn't too uncommon.

packages/angular/ssr/src/app-engine.ts
```diff
@@ -57,6 +73,22 @@ export class AngularAppEngine {
    */
   private readonly entryPointsCache = new Map<string, Promise<EntryPointExports>>();
 
+  /**
+   * Creates a new instance of the Angular server application engine.
+   * @param options Options for the Angular server application engine.
+   */
+  constructor(options?: AngularAppEngineOptions) {
+    const allowedHosts = new Set(this.manifest.allowedHosts);
```
dgp1130: Consider: Can we merge these in the constructor like `new Set([...this.manifest.allowedHosts, ...options?.allowedHosts ?? []])`?

packages/angular/ssr/src/app-engine.ts
```diff
@@ -67,10 +99,43 @@ export class AngularAppEngine {
...
+* 
+* For more information see: https://angular.dev/best-practices/security#preventing-server-side-request-forgery-ssrf
+*/
   async handle(request: Request, requestContext?: unknown): Promise<Response | null> {
-    const serverApp = await this.getAngularServerAppForRequest(request);
+    const allowedHost = this.allowedHosts;
+    request = secureRequest(request, this.allowedHosts);
 
+    try {
+      validateRequest(request, allowedHost);
```
dgp1130, line 111: Question: Won't `validateRequest` read the `host` and `x-forwarded-host` headers, triggering the deopt? Shouldn't we validate before securing the request?

packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -74,6 +81,25 @@ export class CommonEngine {
...
+        validateUrl(urlObj, this.allowedHosts);
+      } catch (error) {
+        let document = opts.document;
+        if (!document && opts.documentFilePath) {
+          document = await this.getDocument(opts.documentFilePath);
+        }
+        // eslint-disable-next-line no-console
+        console.error(
+          `ERROR: Host ${urlObj.hostname} is not allowed. Please provide a list of allowed hosts in the "allowedHosts" option.`,
+          'Fallbacking to client side rendering. This will become a 400 Bad Request in a future major version.\n',
```
AndrewKushnir: nit: a minor typo + a proposal to include the current URL into the error message, so that it's easier to identify which pages are affected.
```suggestion
          'Falling back to client side rendering for ${location}. This will become a 400 Bad Request in a future major version.\n',
```

packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -74,6 +81,25 @@ export class CommonEngine {
...
+      try {
+        validateUrl(urlObj, this.allowedHosts);
+      } catch (error) {
+        let document = opts.document;
+        if (!document && opts.documentFilePath) {
+          document = await this.getDocument(opts.documentFilePath);
+        }
+        // eslint-disable-next-line no-console
+        console.error(
+          `ERROR: Host ${urlObj.hostname} is not allowed. Please provide a list of allowed hosts in the "allowedHosts" option.`,
```
AndrewKushnir: nit: can we also refer to docs here? It's a bit unclear from the error message alone where the `allowedHosts` option is located.

packages/angular/ssr/node/src/app-engine.ts
```diff
@@ -21,32 +27,86 @@ import { createWebRequestFromNodeRequest } from './request';
...
-   * @param request - The incoming HTTP request (`IncomingMessage` or `Http2ServerRequest`).
+   * @param request - The incoming HTTP request (`IncomingMessage`, `Http2ServerRequest` or `Request`).
    * @param requestContext - Optional context for rendering, such as metadata associated with the request.
    * @returns A promise that resolves to the resulting HTTP response object, or `null` if no matching Angular route is found.
    *
    * @remarks A request to `https://www.example.com/page/index.html` will serve or render the Angular route
    * corresponding to `https://www.example.com/page`.
+   * 
+   * @remarks
+* To prevent potential Server-Side Request Forgery (SSRF), this function verifies the hostname
```
AndrewKushnir: nit: formatting looks incorrect.

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,224 @@
...
+  for (const headerName of HOST_HEADERS_TO_VALIDATE) {
+    const headerValue = getFirstHeaderValue(headers.get(headerName));
+
+    // Reject any hostname containing path separators - they're invalid.
+    if (headerValue && PATH_SEPARATOR_REGEX.test(headerValue)) {
+      throw new Error(`Header "${headerName}" contains path separators which is not allowed.`);
+    }
+  }
+
+  const xForwardedPort = getFirstHeaderValue(headers.get('x-forwarded-port'));
```
AndrewKushnir, line 247: Quick question: do we normalize headers and lowercase the names earlier in the code?

packages/angular/ssr/src/app-engine.ts
```diff
@@ -67,10 +99,43 @@ export class AngularAppEngine {
    *
    * @remarks A request to `https://www.example.com/page/index.html` will serve or render the Angular route
    * corresponding to `https://www.example.com/page`.
-   */
+   *
+* @remarks
```
AndrewKushnir: nit: formatting looks incorrect.

packages/angular/ssr/src/app-engine.ts
```diff
@@ -67,10 +99,43 @@ export class AngularAppEngine {
...
+    try {
+      validateRequest(request, allowedHost);
+    } catch (error) {
+      const msg = error instanceof Error ? error.message : error;
+
+      // eslint-disable-next-line no-console
+      console.error(
+        `ERROR: Bad Request ("${request.url}").\n` +
+          msg +
+          '\nFallbacking to client side rendering. This will become a 400 Bad Request in a future major version.',
```
AndrewKushnir: nit: typo + it'd be great to add info about the current URL (similar to one of the comments above).
```suggestion
          '\nFalling back to client side rendering. This will become a 400 Bad Request in a future major version.',
```

packages/angular/ssr/node/src/common-engine/common-engine.ts
alan-agius4: Document in this is case is the index.html which we need to serve it to render CSR.

packages/angular/ssr/node/src/common-engine/common-engine.ts
alan-agius4: Removed.

packages/angular/ssr/node/src/common-engine/common-engine.ts
alan-agius4: My reason for this is that you'd be able to provide the `allowedHost` at runtime where using env variables is not an option. For instance in worker context.

If you feel we should hold this off, I am fine with this. Please let me know.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: No we do not normalize headers names, the `headers.get()` is case insensitive.

packages/angular/ssr/src/app-engine.ts
alan-agius4: Updated.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: Updated to avoid using proxy

packages/angular/ssr/src/utils/validation.ts
alan-agius4: They will actually not trigger the CSR deopt. in this case they will not be validated.

packages/angular/ssr/src/app-engine.ts
alan-agius4: yes

alan-agius4:
> Left some general suggestions, but nothing major. Feel free to ignore or come back to any of these later rather than treating them as blocking this PR. The two main things to confirm are really just:

> Should we commit to ${HOSTNAME} / ${NG_ALLOWED_HOSTS} right now, or come back later?

I removed `${HOSTNAME}`, I think `${NG_ALLOWED_HOSTS}` can be useful to support this dynamically in environments that support Node.js

> Should we error when allowedHosts is configured and we receive a request for a disallowed host, to be up front about a misconfiguration rather than silently deoptimizing?

Updated the logic to do this.

packages/angular/ssr/node/src/common-engine/common-engine.ts
dgp1130: I'm ok either way and happy to defer to you on the decision. I'm just highlighting the option to defer "providing `allowedHosts` at runtime" to v22 if you don't feel it's necessary here. Ultimately up to you, no objection to landing it now if you think that's better.

packages/angular/ssr/src/utils/validation.ts
dgp1130: Why wouldn't this be validated? Wouldn't reading `req.headers['host']` (even through `Object.entries` or another iterator) trigger validation and then deopt?

packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -74,6 +85,40 @@ export class CommonEngine {
...
+        validateUrl(urlObj, this.allowedHosts);
+      } catch (error) {
+        const isAllowedHostConfigured = this.allowedHosts.size > 0;
+        // eslint-disable-next-line no-console
+        console.error(
+          `ERROR: ${(error as Error).message}` +
+            'Please provide a list of allowed hosts in the "allowedHosts" option in the "CommonEngine" constructor.',
+          isAllowedHostConfigured
+            ? ''
+            : '\nFallbacking to client side rendering. This will become a 400 Bad Request in a future major version.',
```
dgp1130: Typo: "Falling back"

packages/angular/ssr/node/src/environment-options.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @returns An array of allowed hosts.
+ */
+export function getAllowedHostsFromEnv(): ReadonlyArray<string> {
+  const allowedHosts: string[] = [];
+  const envNgAllowedHosts = process.env['NG_ALLOWED_HOSTS'];
+  if (!envNgAllowedHosts) {
+    return allowedHosts;
+  }
+
+  const hosts = envNgAllowedHosts.split(',');
```
dgp1130, line 20: Consider: Is it worth validating that the provided names are valid hostnames (letters, numbers, dashes only I think)?

Less of a security thing and more just DX to check for typos.

packages/angular/ssr/src/app-engine.ts
```diff
@@ -67,12 +91,38 @@ export class AngularAppEngine {
...
+    const allowedHost = this.allowedHosts;
 
+    try {
+      validateRequest(request, allowedHost);
+    } catch (error) {
+      return this.handleValidationError(error as Error, request);
+    }
+
+    // Clone request with patched headers to prevent unallowed host header access.
+    const { request: securedRequest, onError: onHeaderValidationError } =
```
dgp1130, line 117: Consider: Rename `request` to `insecureRequest` and then rename `securedRequest` to `request`.

This way avoids future changes accidentally using `request` when they really wanted to use the secured one.

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,200 @@
...
+}
+
+/**
+ * Clones a request and patches the `get` method of the request headers to validate the host headers.
+ * @param request - The request to validate.
+ * @param allowedHosts - A set of allowed hostnames.
+ * @returns An object containing the cloned request and a promise that resolves to an error
+ * if any of the validated headers contain invalid values.
+ */
+export function cloneRequestAndPatchHeaders(
```
dgp1130, line 82: Consider: The functionality purist it me likes the idea of cloning the request, but I'm wondering if there's a risk of the original request leaking into user code somehow in a way which might circumvent this check? Are there any server frameworks which offer a kind of `getActiveRequest()` function which might retrieve the original? Would it be safer to mutate the original object?

Maybe this is fine as a temporary mitigation since we'll eventually move to a hard 400 error up front, so it might not be too big a deal in the short term.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: Because `.entries` retrieves the values from an internal list 

https://github.com/nodejs/undici/blob/5bdd32250090887070ca35d2f1c7750c306828d0/lib/web/fetch/headers.js#L318

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    const key = name.toLowerCase();
+    if (HOST_HEADERS_TO_VALIDATE.has(key)) {
+      try {
+        verifyHostAllowed(key, value, allowedHosts);
+      } catch (error) {
+        onError(error as Error);
+
+        return null;
```
dgp1130: Consider: Would it be safer to throw in user space rather than letting them assume a given header doesn't exist?

We probably still need `onError` as we wouldn't want the user to catch and suppress this error, but I'm thinking user code might behave strangely if we pretend the header doesn't exist at all.

packages/angular/ssr/src/utils/validation.ts
dgp1130: Oh... does that imply that we should patch these iteration functions? `entries`, `entriesList`, `rawValues`, `toSortedArray`, `Symbol.iterator`? Otherwise you could circumvent this check a basic `req.headers.entriesList().find(([header]) => header === 'host')`.

(That example is obviously intentionally working around the safety check, but anything just iterating all headers has the same risk.)

VenkatKwest:
Thanks for the update! I actually discovered this while researching X-Forwarded-Prefix issue and spent a few days doing a deep dive before reporting it to the Google OSS VRP. It was closed there as an internal duplicate, and since the recent commits didn't seem to reference this specific behavior, I thought it was worth opening here.
Just out of curiosity, was this vulnerability identified recently, or has it been on the team's radar for a while?

alan-agius4:
It was reported via the Google OSS program and indeed there were multiple reports for the same issue hence it was deemed duplicate.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: done

packages/angular/ssr/src/app-engine.ts
alan-agius4: I prefer to to rename `request` to something else, as that is the public API.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: Yes, updated.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: I am not certain about this to be honest.

alan-agius4:

- main: 6cf621c905de4815b1e12964fb909876ca6e3ea3
- 21.2.x: f78f38827861cf01471e30fcf81ef14d6265a0f2

Issue #32501, opened 2026-02-18
Open Redirect via Unsanitized X-Forwarded-Prefix Header, Chainable to Web Cache Poisoning


VenkatKwest:
### Command

add, build, other

### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

`joinUrlParts()` in `packages/angular/ssr/src/utils/url.ts` only strips **one** leading slash from URL parts. When the `X-Forwarded-Prefix` header contains multiple leading slashes (e.g., `///evil.com`), the function produces a **protocol-relative URL** (`//evil.com/home`) that browsers interpret as an external redirect.

**What happens:**
1. Angular SSR app has a redirect route: `{ path: 'redirect', redirectTo: 'home' }`
2. Request arrives with header: `X-Forwarded-Prefix: ///evil.com`
3. `app.ts` line 197 calls `joinUrlParts('///evil.com', 'home')`
4. `joinUrlParts` strips only the first `/` → `"//evil.com"` remains
5. Final output: `"//evil.com/home"` → set as `Location` header
6. Browser follows `//evil.com/home` as a protocol-relative redirect → **navigates to `evil.com`**

**Root cause** (`url.ts` line 106):
```typescript
// Current code — only strips ONE leading slash:
if (part[0] === '/') {
  normalizedPart = normalizedPart.slice(1);
}
```

**Suggested fix:**
```typescript
// Strip ALL leading slashes:
while (normalizedPart[0] === '/') {
  normalizedPart = normalizedPart.slice(1);
}
```

This affects all code paths that use `joinUrlParts()` with external input:
- `app.ts:197` — redirect routes with `X-Forwarded-Prefix`
- `ng.ts:215` — SSR navigation redirect URL construction
- `app-engine.ts:116` — i18n locale redirects

### Minimal Reproduction

**Step 1:** Create a new Angular SSR app
```bash
ng new vuln-test-app --ssr
cd vuln-test-app
```

**Step 2:** Add a redirect route in `src/app/app.routes.ts`:
```typescript
import { Routes } from '@angular/router';
import { Component } from '@angular/core';

@Component({ selector: 'app-home', template: '<h1>Home</h1>' })
export class Home {}

export const routes: Routes = [
  { path: 'home', component: Home },
  { path: 'redirect', redirectTo: 'home', pathMatch: 'full' }
];
```

**Step 3:** Build and start the SSR server:
```bash
ng build
node dist/vuln-test-app/server/server.mjs
```

[...]

PR #32521, opened 2026-02-20
 fix(@angular/ssr): prevent open redirect via X-Forwarded-Prefix header 

alan-agius4:
This change addresses a security vulnerability where `joinUrlParts()` in `packages/angular/ssr/src/utils/url.ts` only stripped one leading slash from URL parts.

When the `X-Forwarded-Prefix` header contains multiple leading slashes (e.g., `///evil.com`), the function previously produced a protocol-relative URL (e.g., `//evil.com/home`). If the application issues a redirect (e.g., via a generic redirect route), the browser interprets this 'Location' header as an external redirect to `https://evil.com/home`.

This vulnerability poses a risk as open redirects can be used in phishing attacks. Additionally, since the redirect response may lack `Cache-Control` headers, intermediate CDNs could cache the poisoned redirect, serving it to other users.

This commit fixes the issue by:
1. Updating `joinUrlParts` to internally strip *all* leading and trailing slashes from URL segments, preventing the formation of protocol-relative URLs from malicious input.
2. Adding strict validation for the `X-Forwarded-Prefix` header to immediately reject requests with values starting with multiple slashes.

Closes https://github.com/angular/angular-cli/issues/32501

### DIFF
--- a/packages/angular/ssr/src/utils/url.ts
+++ b/packages/angular/ssr/src/utils/url.ts
@@ -95,26 +95,32 @@ export function addTrailingSlash(url: string): string {
  * ```
  */
 export function joinUrlParts(...parts: string[]): string {
-  const normalizeParts: string[] = [];
+  const normalizedParts: string[] = [];
+
   for (const part of parts) {
     if (part === '') {
       // Skip any empty parts
       continue;
     }
 
-    let normalizedPart = part;
-    if (part[0] === '/') {
-      normalizedPart = normalizedPart.slice(1);
+    let start = 0;
+    let end = part.length;
+
+    // Use "Pointers" to avoid intermediate slices
+    while (start < end && part[start] === '/') {
+      start++;
     }
-    if (part.at(-1) === '/') {
-      normalizedPart = normalizedPart.slice(0, -1);
+
+    while (end > start && part[end - 1] === '/') {
+      end--;
     }
-    if (normalizedPart !== '') {
-      normalizeParts.push(normalizedPart);
+
+    if (start < end) {
+      normalizedParts.push(part.slice(start, end));
     }
   }
 
-  return addLeadingSlash(normalizeParts.join('/'));
+  return addLeadingSlash(normalizedParts.join('/'));
 }
 
 /**
--- a/packages/angular/ssr/src/utils/validation.ts
+++ b/packages/angular/ssr/src/utils/validation.ts
@@ -27,6 +27,11 @@ const VALID_PROTO_REGEX = /^https?$/i;
 const VALID_HOST_REGEX = /^[a-z0-9.:-]+$/i;
 
 /**
+ * Regular expression to validate that the prefix is valid.
+ */
+const INVALID_PREFIX_REGEX = /^[/\\]{2}|(?:^|[/\\])\.\.?(?:[/\\]|$)/;
+
+/**
  * Extracts the first value from a multi-value header string.
  *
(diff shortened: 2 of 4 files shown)


packages/angular/ssr/src/utils/validation.ts
```diff
@@ -0,0 +1,213 @@
...
+    throw new Error('Header "x-forwarded-port" must be a numeric value.');
+  }
+
+  const xForwardedProto = getFirstHeaderValue(headers.get('x-forwarded-proto'));
+  if (xForwardedProto && !VALID_PROTO_REGEX.test(xForwardedProto)) {
+    throw new Error('Header "x-forwarded-proto" must be either "http" or "https".');
+  }
+
+  const xForwardedPrefix = getFirstHeaderValue(headers.get('x-forwarded-prefix'));
+  if (xForwardedPrefix?.startsWith('//')) {
```
securityMB: this is not sufficient (see my comment in internal doc)

alan-agius4:
@dgp1130 & @AndrewKushnir please ignore the first commit as this is from https://github.com/angular/angular-cli/pull/32516. 

@josephperrott, I added you by mistake as a reviewer, and I cannot remove you.

packages/angular/ssr/src/utils/url.ts
```diff
@@ -95,26 +95,32 @@ export function addTrailingSlash(url: string): string {
  * ```
  */
 export function joinUrlParts(...parts: string[]): string {
```
dgp1130, line 97: Consider: Could a regex be simpler? Something like:

```typescript
const regex = new Regex('^/*(?<part>.*?)/*$');
const normalizedParts = parts.map((part) => regex.exec(part).groups['part']);
return addLeadingSlash(normalizedParts.join('/'));
```

packages/angular/ssr/src/utils/validation.ts
```diff
@@ -205,4 +210,9 @@ function validateHeaders(request: Request, allowedHosts: ReadonlySet<string>): v
   if (xForwardedProto && !VALID_PROTO_REGEX.test(xForwardedProto)) {
     throw new Error('Header "x-forwarded-proto" must be either "http" or "https".');
   }
+
+  const xForwardedPrefix = getFirstHeaderValue(headers.get('x-forwarded-prefix'));
+  if (xForwardedPrefix && INVALID_PREFIX_REGEX.test(xForwardedPrefix)) {
+    throw new Error('Header "x-forwarded-prefix" must not start with multiple "/" or "\\".');
```
dgp1130: Question: Is there anything we can / should do for `X-Forwarded-Prefix: ../`? Is there any risk there?

packages/angular/ssr/src/utils/url.ts
alan-agius4: This is quite a hot function, whilst I didn’t benchmark it I assume the above is much slower.

packages/angular/ssr/src/utils/validation.ts
alan-agius4: Umm.. yes we could either disallow path traversing or add another allow list.

AndrewKushnir: +1 to disable the pattern that Doug mentioned + also `./` one.

alan-agius4:

- main: f086eccc36d10cf01c426e35864bc32e1e292323
- 21.2.x: cf5a72d33c1753c121f71498c0c415a0f0864492

PR #32524, opened 2026-02-20
fix(@angular/cli): validate package manager version using `semver.valid` and throw an error if invalid

alan-agius4:
Prior to this change an `An unhandled exception occurred: Invalid Version: ` was being thrown.

### DIFF
--- a/packages/angular/cli/src/package-managers/package-manager.ts
+++ b/packages/angular/cli/src/package-managers/package-manager.ts
@@ -371,7 +371,7 @@ export class PackageManager {
     this.#version = stdout.trim();
 
     if (!valid(this.#version)) {
-      throw new Error(`Invalid semver version for ${this.name}: ${this.#version}`);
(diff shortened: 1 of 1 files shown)


packages/angular/cli/src/package-managers/package-manager.ts
```diff
@@ -370,6 +370,10 @@ export class PackageManager {
     const { stdout } = await this.#run(this.descriptor.versionCommand);
     this.#version = stdout.trim();
 
+    if (!valid(this.#version)) {
+      throw new Error(`Invalid semver version for ${this.name}: ${this.#version}`);
```
clydin: Maybe also put the version in quotes to better articulate empty string cases?

clydin (approved):

alan-agius4:

- main: 9f8f40ed1d81090700fe3194b4b3c220511f1fa9
- 21.2.x: cdd26bb66d8ab334f76323c2b5cae1aa8ce815f6

PR #32527, opened 2026-02-20
refactor(@angular/build): Expose codeBundleCache to buildApplicationInternal

Aukevanoost:
First off, thanks for having a look at this!

This PR exposes the `codeBundleCache` used in executeBuild part of the `buildApplicationInternal` to allow for performant multi-builder setups (because they can share the TS cache).

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Right now the `SourceFileCache` is initialized locally in the `executeBuild()` function. 

Issue Number: N/A

## What is the new behavior?

The `SourceFileCache` is now part of the `InternalOptions`, allowing tools that use this builder to (optionally) provide a shared cache that can be re-used over multiple builds. As fallback it will create the default cache object.

- [ ] Yes
- [x] No

## Other information

We (native-federation) need this feature to speed up our builder. This way we can share the cache between the ESBuild plugin and main builder which decreases our build time drastically:

If interested, find the changes on our end here (it's a proof-of-concept): https://github.com/native-federation/angular-adapter/pull/4/changes#diff-6508a5b483eb2d55fd58361931d864832b5ba20178c796d07be21f9e8ac327ae

### DIFF
--- a/packages/angular/build/src/utils/normalize-cache.ts
+++ b/packages/angular/build/src/utils/normalize-cache.ts
@@ -42,6 +42,7 @@ function hasCacheMetadata(value: unknown): value is { cli: { cache: CacheMetadat
 export function normalizeCacheOptions(
   projectMetadata: unknown,
   worspaceRoot: string,
+  projectName?: string,
 ): NormalizedCachedOptions {
   const cacheMetadata = hasCacheMetadata(projectMetadata) ? projectMetadata.cli.cache : {};
 
@@ -67,9 +68,11 @@ export function normalizeCacheOptions(
 
   const cacheBasePath = resolve(worspaceRoot, path);
 
+  const cachePath = join(cacheBasePath, VERSION, projectName ?? '');
+
   return {
     enabled: cacheEnabled,
     basePath: cacheBasePath,
-    path: join(cacheBasePath, VERSION),
+    path: cachePath,
   };
 }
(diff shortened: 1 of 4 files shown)


packages/angular/build/src/utils/normalize-cache.ts
```diff
@@ -67,9 +68,13 @@ export function normalizeCacheOptions(
 
   const cacheBasePath = resolve(worspaceRoot, path);
 
+  const cachePath = projectName
+    ? join(cacheBasePath, VERSION, projectName)
+    : join(cacheBasePath, VERSION);
```
alan-agius4: NIT
```suggestion
  const cachePath = join(cacheBasePath, VERSION, projectName ?? '');
```

Aukevanoost: I've incorporated your suggested refactor in the commit!

Aukevanoost:
Hi all, 

I had a small question. Is this feature/fix something that could (or is) considered for ng22? Are there any potential downsides for adding this extra option? For tools that rely on the Angular builder this could be a huge decrease in build time. I'm happy to discuss the details. 

Also, I realized later that adding a predefined cache is not very productive if the cache is being purged right before starting the builder.[[1](https://github.com/angular/angular-cli/blob/67c391a18ac417660dd540bd54ecc3482a94eeef/packages/angular/build/src/builders/application/index.ts#L45)]  Would it make sense to skip purging the old cache if a codeBundleCache is provided? 

Thanks

PR #32554, opened 2026-02-24
fix(@angular/cli): handle `oneOf` when converting schema to yargs options

terencehonles:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Options such as `--allowed-hosts` are defined as `oneOf([array<string>, boolean])` but the array option is not chosen, even though it is first, because the JSONSchema conversion only supports an array at the root of the option schema.

Issue Number: N/A

## What is the new behavior?

This change fixes JSONSchemas where `oneOf` is placed at the root of the option schema rather than in an array's `items`. This allows an array to be passed via the command-line, but additional types to be represented via configuration.

- [x] Yes
- [ ] No

This will affect any commands for each of the options that use `oneOf` at the root of their schemas. This change does not change the precedence of schema definition, so if there are any changes and they are not wanted then it is just uncovering a bug where the preferred type should have been ordered first.

### DIFF
--- a/packages/angular/cli/src/command-builder/utilities/json-schema.ts
+++ b/packages/angular/cli/src/command-builder/utilities/json-schema.ts
@@ -130,21 +130,22 @@ function isStringMap(node: json.JsonObject): boolean {
   );
 }
 
-const SUPPORTED_PRIMITIVE_TYPES = new Set(['boolean', 'number', 'string']);
+const SUPPORTED_PRIMITIVE_TYPES = new Set(['boolean', 'number', 'string'] as const);
+type SupportedPrimitiveType = Parameters<typeof SUPPORTED_PRIMITIVE_TYPES.add>[0];
 
 /**
  * Checks if a string is a supported primitive type.
  * @param value The string to check.
  * @returns `true` if the string is a supported primitive type, otherwise `false`.
  */
-function isSupportedPrimitiveType(value: string): boolean {
-  return SUPPORTED_PRIMITIVE_TYPES.has(value);
+function isSupportedPrimitiveType(value: string): value is SupportedPrimitiveType {
+  return SUPPORTED_PRIMITIVE_TYPES.has(value as any);
 }
 
 /**
  * Recursively checks if a JSON schema for an array's items is a supported primitive type.
  * It supports `oneOf` and `anyOf` keywords.
- * @param schema The JSON schema for the array's items.
+ * @param schema The JSON schema to check.
  * @returns `true` if the schema is a supported primitive type, otherwise `false`.
  */
 function isSupportedArrayItemSchema(schema: json.JsonObject): boolean {
@@ -154,6 +155,10 @@ function isSupportedArrayItemSchema(schema: json.JsonObject): boolean {
 
   if (json.isJsonArray(schema.enum)) {
     return true;
+  }
+
+  if (isJsonObject(schema.items)) {
+    return isSupportedArrayItemSchema(schema.items);
   }
 
   if (json.isJsonArray(schema.items)) {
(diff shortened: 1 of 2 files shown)


alan-agius4:
This is a breaking change `--no-allowed-hosts` will become non-functional.

terencehonles:
I forgot `--no-allowed-hosts` would be one of the accepted forms, but I think it's probably unlikely that allowed hosts are configured but you want to turn that off via the command line. You may be able to use `--allowed-hosts ""` to fake the same thing, as "" shouldn't match anything.

I looked into how yargs does its parsing and it _almost_ works (it doesn't actually are about the option definition). Using `--no-oneOfAtRoot` I expected it to return `false`, but it returns `[false]` which would fail the validation.

terencehonles:
The change I pushed should handle `--no-allowed-hosts` as yargs will try to guess you want a boolean in this case, and since the type is not declared as an array, it will return a single boolean value instead of the `[false]` I was seeing before. It's possible the recursive function I added should just be a default to `string`, but this will try to find things like an array of numbers (which I just added a test for)

packages/angular/cli/src/command-builder/utilities/json-schema.ts
```diff
@@ -130,21 +130,22 @@ function isStringMap(node: json.JsonObject): boolean {
   );
 }
 
-const SUPPORTED_PRIMITIVE_TYPES = new Set(['boolean', 'number', 'string']);
+const SUPPORTED_PRIMITIVE_TYPES = new Set(['boolean', 'number', 'string'] as const);
+type SupportedPrimitiveType = Parameters<typeof SUPPORTED_PRIMITIVE_TYPES.add>[0];
```
alan-agius4, line 134: Consider the below, which IMO it's slightly easier to read.

```suggestion
const types = ['boolean', 'number', 'string'] as const;
const SUPPORTED_PRIMITIVE_TYPES: ReadonlySet<string> = new Set(types);
type SupportedPrimitiveType = (typeof types)[number];
```

packages/angular/cli/src/command-builder/utilities/json-schema.ts
```diff
@@ -130,21 +130,22 @@ function isStringMap(node: json.JsonObject): boolean {
...
 
 /**
  * Checks if a string is a supported primitive type.
  * @param value The string to check.
  * @returns `true` if the string is a supported primitive type, otherwise `false`.
  */
-function isSupportedPrimitiveType(value: string): boolean {
-  return SUPPORTED_PRIMITIVE_TYPES.has(value);
+function isSupportedPrimitiveType(value: string): value is SupportedPrimitiveType {
+  return SUPPORTED_PRIMITIVE_TYPES.has(value as any);
```
alan-agius4, line 142: Remove the casting, this is not needed if you change the set type.

```suggestion
  return SUPPORTED_PRIMITIVE_TYPES.has(value);
```

packages/angular/cli/src/command-builder/utilities/json-schema.ts
```diff
@@ -377,9 +416,13 @@ export async function parseJsonSchemaToOptions(
             type: 'array',
             itemValueType: 'string',
           }
-        : {
-            type,
-          }),
+        : type === 'array'
```
alan-agius4, line 419: This isn't right as it cause the `array` type to be replaced by a primitive and forces yargs to parse it as a non array. Example `--flag foo bar` will now become invalid.

alan-agius4 (changes requested):
See comments

terencehonles:
Sorry, I've not had time to tend to this and can revisit this later

PR #32579, opened 2026-02-25
fix(@angular/ssr): remove CSR fallback for invalid hosts

alan-agius4:
Previously, when a request contained an unrecognized host header, the server would fallback to serving the client-side application (CSR) as a temporary migration path.

This commit removes this fallback behavior. Requests with invalid or unrecognized host headers will now strictly return a 400 Bad Request response.

BREAKING CHANGE: The server no longer falls back to Client-Side Rendering (CSR) when a request fails host validation. Requests with unrecognized  headers will now return a 400 Bad Request status code. Users must ensure all valid hosts are correctly configured in the  option.

### DIFF
--- a/packages/angular/ssr/node/src/common-engine/common-engine.ts
+++ b/packages/angular/ssr/node/src/common-engine/common-engine.ts
@@ -92,28 +92,11 @@ export class CommonEngine {
       try {
         validateUrl(urlObj, this.allowedHosts);
       } catch (error) {
-        const isAllowedHostConfigured = this.allowedHosts.size > 0;
         // eslint-disable-next-line no-console
         console.error(
           `ERROR: ${(error as Error).message}` +
             'Please provide a list of allowed hosts in the "allowedHosts" option in the "CommonEngine" constructor.',
-          isAllowedHostConfigured
-            ? ''
-            : '\nFalling back to client side rendering. This will become a 400 Bad Request in a future major version.',
         );
-
-        if (!isAllowedHostConfigured) {
-          // Fallback to CSR to avoid a breaking change.
-          // TODO(alanagius): Return a 400 and remove this fallback in the next major version (v22).
-          let document = opts.document;
-          if (!document && opts.documentFilePath) {
-            document = opts.document ?? (await this.getDocument(opts.documentFilePath));
-          }
-
-          if (document) {
-            return document;
-          }
-        }
 
         throw error;
       }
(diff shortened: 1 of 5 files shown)


packages/angular/ssr/node/src/common-engine/common-engine.ts
```diff
@@ -92,29 +92,12 @@ export class CommonEngine {
       try {
         validateUrl(urlObj, this.allowedHosts);
       } catch (error) {
-        const isAllowedHostConfigured = this.allowedHosts.size > 0;
         // eslint-disable-next-line no-console
         console.error(
           `ERROR: ${(error as Error).message}` +
             'Please provide a list of allowed hosts in the "allowedHosts" option in the "CommonEngine" constructor.',
```
dgp1130, line 98: Consider: Do we still want a separate log statement since we have an unconditional`throw` below? I feel like this would result in double-logging this error under more circumstances.

alan-agius4: In this case the `throw` will be passed to the express error which will not print the message in the console.

alan-agius4:
> Validate allowedHosts at startup (maybe limited to production builds?), so the server fails to even start without it. That would hopefully trigger a CI error in most environments and be visible locally.

This isn't feasible because `allowedHosts` can be configured through both the options object and environment variables and sense a build failure both during `ng build` and `ng serve` is not a possibility.

>  Add some kind of `allowEmptyHostsInDev` option and enable it automatically in ng new, but then cause a type error for `allowedHosts: undefined | [], allowEmptyHostsInDev: false | undefined. This way upgrading apps would get a compile-error that a host needs to be specified while not blocking new apps from dev. Then once we're sure all existing apps have a host listed, we can drop allowEmptyHostsInDev in the next major.

Again a build error is not feasible here due to the same thing as described above, in dev-server, we pass the Vite's `allowedHost` or the provided host that is provided to vite to the SSR engine. A hostname (defaults to localhost) is always needed for the dev-server so the hostname is known. One thing we do not handle is `0.0.0.0` and connecting to the dev-server with external IPs. The dev-server already has an opton `--allowed-host` which accepts a boolean and disables host checking. I think potentially we could leverage that. I'll try to followup.

alan-agius4:

- main: 27cd355619aad140dfc221c6bd161f4a981e0f3b

PR #32612, opened 2026-02-26
refactor(@angular/build): add debug logging to Vitest executor

clydin:
This commit introduces a multi-level debug logging system within the Vitest test runner's executor. This logging is intended to aid in troubleshooting internal issues and provide deeper insights into the test execution process.

Logging is controlled by the `NG_TEST_LOG` environment variable, which accepts numeric levels (0 for off, 1 for debug, 2 for verbose). Crucial lifecycle events and state changes in the `VitestExecutor` are now logged, such as constructor instantiation, build result processing (full and incremental), test re-run specifications, Vitest initialization details (including browser configuration, external configuration path, and included files), and executor disposal. These logs are guarded by the `logLevel` to prevent any performance penalties during normal execution. Furthermore, specific areas like incremental run file mappings and external dependency metadata are logged at a verbose level.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/executor.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/executor.ts
@@ -27,6 +27,11 @@ import { setupBrowserConfiguration } from './browser-provider';
 import { findVitestBaseConfig } from './configuration';
 import { createVitestConfigPlugin, createVitestPlugins } from './plugins';
 
+enum DebugLogLevel {
+  Info = 1,
+  Verbose = 2,
+}
+
 export class VitestExecutor implements TestExecutor {
   private vitest: Vitest | undefined;
   private normalizePath: ((id: string) => string) | undefined;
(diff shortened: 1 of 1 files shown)


packages/angular/build/src/builders/unit-test/runners/vitest/executor.ts
```diff
@@ -40,6 +40,7 @@ export class VitestExecutor implements TestExecutor {
     explicitBrowser: [],
     explicitServer: [],
   };
+  private readonly logLevel: number;
```
alan-agius4: Can we switch from numbers to an enum?

Using numbers increases the cognitive load when parsing the file, as one has to mentally map numbers like 1 or 2 to their true meanings and an enum would make the logic self-documenting

clydin:

- main: 2887d84986329c929bea68e80f58510f49be7f6f
- 21.2.x: ad2c14e521142479991534f8f31416dbc3f8acd3

PR #32635, opened 2026-03-02
fix(@angular/build): bundle polyfills to preserve execution order in dev server

### LINKED ISSUE #32632: Polyfills are improperly inlined and execute in the wrong order
### Command

serve

### Is this a regression?

19.2.0

### Description

I encountered this bug while migrating a project with zone flags set but I'm aware that zoneless is a thing now so maybe less people will have hit it as a result. However, this bug can be generalized to any polyfill scripts requiring a certain order. Given the following setup:

```json
// angular.json
"polyfills": [
  "src/zone-flags.js",
  "zone.js"
],
```

```js
// src/zone-flags.js
globalThis.__Zone_disable_ZoneAwarePromise = true;
```

When serving the project and inspecting the source, this is what you'd see in Angular 19:

```js
// src/zone-flags.js
globalThis.__Zone_disable_ZoneAwarePromise = true;

// node_modules/.pnpm/zone.js@0.16.1/node_modules/zone.js/fesm2015/zone.js
var __defProp = Object.defineProperty;
...
```

Starting in Angular 20, you'd see this instead:

```js
// src/zone-flags.js
globalThis.__Zone_disable_ZoneAwarePromise = true;

// angular:polyfills:angular:polyfills
import "/@fs/D:/New folder/org2/.angular/cache/21.2.0/org/vite/deps/zone__js.js?v=e512eaef";
```

For anyone unfamiliar with ESM import behavior, they are always the first to execute regardless of statements that come before them. Ergo, the zone.js script will run before the flags are set, thus rendering the flags useless and breaking the polyfill ordering.

Note that this only affects the dev server; when building, the polyfills are compiled into one script. I have not checked if this change in the dev server extends beyond the polyfill scripts, i.e. imports being inlined in application code. I'm aware that the dev server is powered by Vite but I do not know enough about this implementation to file a bug on Angular's behalf.

### Minimal Reproduction

This won't reproduce on StackBlitz so you'll have to run the commands yourself.

1. ng new --zoneless false
2. Create new file zone-flags.js
3. Add zone-flags.js before zone.js in angular.json polyfills
4. ng serve
5. Check if Promise is patched in browser F12 console

### Exception or Error

```text

```

### Your Environment

```text
Angular CLI       : 21.2.0
Node.js           : 24.14.0
Package Manager   : npm 11.11.0
Operating System  : win32 x64
```

### Anything else relevant?

_No response_

alan-agius4:
When using the development server with ESM, polyfills were built as a mix of bundled local files and external package imports. Due to ESM hoisting behavior, all top-level import statements are executed before any other code in the module. This caused polyfills that modify global state (like zone-flags.js) to run after the polyfills they were intended to configure (like zone.js), rendering the flags ineffective.

This change ensures that all polyfills are bundled together into the single polyfills.js file when using the development server, matching the behavior of the production build and preserving the intended execution order.

Closes #32632

### DIFF
--- a/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
+++ b/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
@@ -654,9 +654,13 @@ function getEsBuildCommonPolyfillsOptions(
   tryToResolvePolyfillsAsRelative: boolean,
   loadResultCache: LoadResultCache | undefined,
 ): BuildOptions | undefined {
-  const { jit, workspaceRoot, i18nOptions, externalPackages } = options;
+  const { jit, workspaceRoot, i18nOptions } = options;
 
-  const buildOptions = getEsBuildCommonOptions(options);
+  const buildOptions = getEsBuildCommonOptions({
+    ...options,
+    externalPackages: false,
+  });
+  buildOptions.packages = 'bundle';
   buildOptions.splitting = false;
   buildOptions.plugins ??= [];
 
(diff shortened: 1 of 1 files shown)


clydin (approved):

alan-agius4:

- main: ec289c36baa5a1df780406fe7c54b846302ad877
- 21.2.x: 6ad8608636ad48ae140cc7299a32e0358c761fcc

Viktor-Ivliev:
This change led to excessive RAM usage and slowed down the vitest tests by nearly half. 
https://github.com/angular/angular-cli/issues/32759

PR #32638, opened 2026-03-02
fix(@angular/cli): correct dev dependency detection logic in `ng add`

### LINKED ISSUE #32630: bug(schematics): adding package to devDependencies
### Is this a regression?

_No response_

### Description

`ng add @angular/material` adds package `@angular/material` in devDependencies section of `package.json`

### Reproduction

Steps to reproduce:
1. Create a new angular app
2. Run `ng add @angular/material`

### Expected Behavior

I would expect @angular/material to be added to `dependencies` section only, not in `devDependencies`

### Actual Behavior

@angular/material is being added to both sections `dependencies` and `devDependencies`

### Environment

- Angular: 21.2.0
- CDK/Material: 21.2.0
- Browser(s): -
- Operating System (e.g. Windows, macOS, Ubuntu): macOS

alan-agius4:
Previously, the logic for determining whether to install a package as a dev dependency in `ng add` was using a negative check (`!== 'dependencies'`). This has been changed to an explicit check (`=== 'devDependencies'`) to ensure the same behaviour as previous versions.

Closes #32630

### DIFF
--- a/packages/angular/cli/src/commands/add/cli.ts
+++ b/packages/angular/cli/src/commands/add/cli.ts
@@ -583,7 +583,7 @@ export default class AddCommandModule
         await packageManager.add(
           packageIdentifier.toString(),
           'none',
-          savePackage !== 'dependencies',
+          savePackage === 'devDependencies',
           false,
           true,
           {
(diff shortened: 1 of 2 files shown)


alan-agius4:

- main: 038a9317ed6411500210a2b771ca574452a5e93c
- 21.2.x: ae4c28d0083d948489f4ba38c571b7f955400226

PR #32646, opened 2026-03-02
refactor(@ngtools/webpack): support optional baseUrl with tsconfig paths

clydin:
The Webpack plugin now ensures a `baseUrl` is set when `paths` are present, using the `pathsBasePath` as a fallback. This makes the `baseUrl` property optional in `tsconfig.json` files when `paths` are used.

This commit also refactors several tests to remove their reliance on a pre-existing `baseUrl`. The tests now programmatically add `paths` to the `tsconfig.json` to avoid brittle string replacements. The test project's base `tsconfig.json` has also been updated to remove the now optional `baseUrl`.

### DIFF
--- a/modules/testing/builder/projects/hello-world-app/tsconfig.json
+++ b/modules/testing/builder/projects/hello-world-app/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "compileOnSave": false,
   "compilerOptions": {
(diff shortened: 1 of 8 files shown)


clydin:

- main: b47078fadf82e9502d508263d7579014d0044166

PR #32654, opened 2026-03-03
build: update to TypeScript 6

crisbeto:
Updates the repo to TypeScript 6 and resolves the resulting test failures.

### DIFF
--- a/packages/angular_devkit/build_angular/src/builders/ssr-dev-server/specs/proxy_spec.ts
+++ b/packages/angular_devkit/build_angular/src/builders/ssr-dev-server/specs/proxy_spec.ts
@@ -11,7 +11,6 @@ import { Architect } from '@angular-devkit/architect';
 import * as browserSync from 'browser-sync';
 import * as http from 'node:http';
 import { createArchitect, host } from '../../../testing/test-utils';
-import { SSRDevServerBuilderOutput } from '../index';
 
 describe('Serve SSR Builder - Proxy', () => {
   const target = { project: 'app', target: 'serve-ssr' };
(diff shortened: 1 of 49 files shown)


packages/angular/build/src/builders/application/tests/options/allowed-common-js-dependencies_spec.ts
```diff
@@ -134,6 +134,9 @@ describeBuilder(buildApplication, APPLICATION_BUILDER_INFO, (harness) => {
         tsconfig.compilerOptions.paths = {
           '@app/*': ['./src/app/*'],
         };
+        // TODO: TS reports that a deprecation error because `baseUrl` is used,
+        // but we don't have it in our tsconfig. It may be coming from a dependency.
+        tsconfig.compilerOptions.ignoreDeprecations = '6.0';
```
crisbeto: I couldn't quite figure out where the `baseUrl` for tests like this one was coming from. As far as I could tell we weren't setting it explicitly.

alan-agius4: I tried this test locally and removing `tsconfig.compilerOptions.ignoreDeprecations` passes.

packages/angular_devkit/build_angular/src/builders/ssr-dev-server/specs/proxy_spec.ts
```diff
@@ -30,10 +29,16 @@ describe('Serve SSR Builder - Proxy', () => {
     await host.initialize().toPromise();
     architect = (await createArchitect(host.root())).architect;
 
+    host.replaceInFile(
```
alan-agius4: Update the tsconfig at source instead of adding the types for every spec.

https://github.com/angular/angular-cli/blob/843ca01c78141b1e72ce592a9f970669bd03a5e7/modules/testing/builder/projects/hello-world-app/src/tsconfig.server.json#L6

crisbeto:
Should be ready to go.

clydin:

- main: 1d7fceb55c4333176a33a0794288cabe68527c0f

Issue #32641, opened 2026-03-02
Allow using `vi.mock` (and related) functions in tests


MillerSvt:
### Which @angular/* package(s) are relevant/related to the feature request?

core

### Description

Currently we cannot use `vi.mock` function for automocking component/directives/services/pipes/etc...

```
Error: The "vi.mock" and related methods are not supported with the Angular unit-test system. Please use Angular TestBed for mocking.
```

TestBed is not enough for mocking. It can't replace component/directives/services/pipes/...etc implementation with `vi.fn()` stubs. It can't mock other modules with functions/constants/...etc.

### Proposed solution

Implement something like [ng-automocks-jest](https://www.npmjs.com/package/ng-automocks-jest) for vitest.

I've implemented deep-automocking infrastructure for `jest.mock()`, it can stub any angular entities, both metadata and implementation.

### Alternatives considered

Currently we have no choice, but stay with slow and inefficient `jest` + `jest-preset-angular`.

HerrDerb:
I did run into the same issue. I did mock an import (`import { jwtDecode } from 'jwt-decode';`) for test which now does not work anymore.

```
vi.mock('jwt-decode', () => ({
  jwtDecode: vi.fn(),
}));


it(...){
  //...
    (jwtDecode as ReturnType<typeof vi.fn>).mockReturnValue({
      realm_access: {
        roles: ['admin', 'user'],
      },
    });
  //...

}
```

gcoadour:
Related to https://github.com/angular/angular-cli/issues/31609

MillerSvt:
@alan-agius4 let's continue here.

> Instead TestBed.overrideProvider or TestBed.configureTestingModule should be used as is the supported way

TestBed is not enough to mocking. As described here, it cannot mock implementation. It can only override metadata. Also TestBed cannot mock platform providers, but only root providers.

If you want to stay framework agnostic, please allow us to add esbuild plugin to implement this thing by third-party plugin. We need to intercept `vi.mock` and related calls at compile-time, and transform it for tests bundle. I can create `ng-automocks-vitest` plugin and implement this mocks on my own.

MillerSvt:
Can I try to make pull request for this to push it forward?

PR #32655, opened 2026-03-03
fix(@angular/build): conditionally allow `vi.mock` for non-relative imports

clydin:
Previously, `vi.mock` and related Vitest mocking APIs were completely disabled, causing errors even for valid use cases like mocking Node.js built-in modules or installed packages.

This commit refactors the `vitest-mock-patch` to:
- Allow `vi.mock` for non-relative module paths (e.g., 'fs', '@angular/core').
- Continue to block `vi.mock` for relative module paths (e.g., './my-local-module') to enforce the use of Angular TestBed for local dependency mocking.

### DIFF
--- a/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
+++ b/packages/angular/build/src/builders/unit-test/runners/vitest/build-options.ts
@@ -186,13 +186,32 @@ export async function getVitestBuildOptions(
 
   const mockPatchContents = `
     import { vi } from 'vitest';
+
     const error = new Error(
-    'The "vi.mock" and related methods are not supported with the Angular unit-test system. Please use Angular TestBed for mocking.');
-    vi.mock = () => { throw error; };
-    vi.doMock = () => { throw error; };
-    vi.importMock = () => { throw error; };
-    vi.unmock = () => { throw error; };
-    vi.doUnmock = () => { throw error; };
+      'The "vi.mock" and related methods are not supported for relative imports with the Angular unit-test system. ' +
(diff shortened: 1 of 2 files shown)


MillerSvt:
I don't understand how this fix can resolve #32641. In this issue we asking to allow any modules to use inside `vi.mock`, including local modules, including modules containing angular entities (components, services, directives, ...etc).

This fix is not related to #32641. And I think we should develop more common solution to avoid breaking changes in future.

So, please, do not merge this pull request until the team makes decision on #32641 issue.

alan-agius4:
@MillerSvt, while this PR doesn’t implement the "universal mocking" requested in #32641, it resolves a specific regression for mocking non relative modules.

It is important to note that mocking local Angular entities (services, components, directives, etc.) via `vi.mock` or similar module-level interception has never been officially supported by any of the Angular CLI builders, whether based on Webpack or Esbuild.

Regarding the broader issue, mocking at the module level (like `vi.mock`) is generally discouraged in Angular because it bypasses the compiler's metadata processing and is highly brittle and we want to avoid having test runner specific mocking logic.

Instead `TestBed.overrideProvider` or `TestBed.configureTestingModule` should be used as is the supported way to swap services or components and are test runner agnostic. It ensures your tests remain stable, type-safe, and compatible with future CLI changes.

clydin:

- main: 108fe0735f360d956cfdff61907682a2ea4138e3
- 21.2.x: d1739737564fbcc3e4c5a6c3369046cccf0f6120

PR #32677, opened 2026-03-04
refactor: move `baseUrl` fallback logic to `paths-plugin.ts`

alan-agius4:
This ensure that we do not set `baseUrl` internally which would trigger a deprecation warning.


alan-agius4:

- main: 634066654994b938948e856f2420f439c7d5b6d4

PR #160, opened 2016-01-30
fix(libs): 3rd party library install fixes

jkuri:
Closes https://github.com/angular/angular-cli/issues/152 and various other issues installing and injecting 3rd party libs.

PR #32678, opened 2026-03-04
fix: prevent command injection via shell metacharacters on Windows

### LINKED ISSUE #32509: Command Injection via `child_process.spawn` with `shell: true` on Windows in Angular CLI
### Command

add, new

### Is this a regression?

_No response_

### Description

Two files in Angular CLI use a deprecated `spawn()` pattern on Windows — arguments are concatenated into a single string instead of being passed as an array. This causes shell metacharacters (`&`, `|`, `>`) to be interpreted by `cmd.exe`, leading to **unintended OS command injection**.

Node.js has explicitly deprecated this pattern as **DEP0190**. The Linux/macOS code path already does it correctly.

---

**File 1: `packages/angular/cli/src/package-managers/host.ts` (Lines 135-136)**

```typescript
const childProcess = isWin32
  ? spawn(`${command} ${args.join(' ')}`, spawnOptions) // Windows: concatenated string
  : spawn(command, args, spawnOptions); // Linux: proper array
```

When `ng add` receives a package specifier with shell metacharacters (e.g., `lodash@1.0.0 & echo injected`), `npm-package-arg` parses it as a valid range specifier. The `& echo injected` portion flows unsanitized into `args`, gets concatenated, and `cmd.exe` executes it as a separate command.

---

**File 2: `packages/angular_devkit/schematics/tasks/repo-init/executor.ts` (Line 44)**

```typescript
spawn(`git ${args.join(' ')}`, spawnOptions); // Same deprecated pattern
```

The git commit message passed via `ng new --commit.message="..."` flows into this `spawn()` call. A crafted message like `" & echo PWNED > pwned.txt & rem "` breaks out of the git command and executes arbitrary commands.

The commit message is injected at line 85 without sanitization:

```typescript
await execute(['commit', `-m "${message}"`]);
```

---

**Impact:** On Windows, both code paths allow **OS command injection** — arbitrary commands execute with the privileges of the user running Angular CLI. This is a behavioral inconsistency with Linux/macOS, where args are passed as an array and shell metacharacters are treated as literal characters.

### Minimal Reproduction

**Reproduction 1: `ng add` (host.ts)**

```bash
ng new test-app
cd test-app
ng add "lodash@1.0.0 & echo INJECTED" --skip-confirmation
```

**Expected:** The entire string `lodash@1.0.0 & echo INJECTED` is treated as a single argument (same as Linux).
**Actual (Windows):** `cmd.exe` splits at `&` and runs `echo INJECTED` as a separate command.

---

**Reproduction 2: `ng new` (executor.ts)**

```bash
ng new malicious-app --skip-install --commit.message="\" & echo PWNED > pwned.txt & rem \""
```

**Expected:** The message is passed as a literal string to `git commit -m`.
**Actual (Windows):** `cmd.exe` executes `echo PWNED > pwned.txt` — a file `pwned.txt` is created in the working directory, confirming OS command injection.

### Exception or Error

```text
No exception. The injected commands execute silently alongside the intended operation. This is what makes the behavior dangerous — there is no visible error to alert the user.

Example for `ng add`:


✔ Determining Package Manager
  › Using package manager: npm
✔ Loading package information     ← injected command runs here silently
✔ Installing package
```

### Your Environment

```text
Angular CLI: 21.1.4
Node: 25.6.0
Package Manager: npm 11.3.0
OS: Windows 11

Angular: 21.1.4

@angular-devkit/architect    0.2101.4
@angular-devkit/build-angular 21.1.4
@angular-devkit/core         21.1.4
@angular-devkit/schematics   21.1.4
@angular/cli                 21.1.4
@schematics/angular          21.1.4
```

### Anything else relevant?

**Suggested fix for both files — use the same spawn pattern as Linux/macOS:**

File 1 (`host.ts`):

```diff
- const childProcess = isWin32
-   ? spawn(`${command} ${args.join(' ')}`, spawnOptions)
-   : spawn(command, args, spawnOptions);
+ const childProcess = spawn(command, args, spawnOptions);
```

File 2 (`executor.ts`):

```diff
- spawn(`git ${args.join(' ')}`, spawnOptions)
+ spawn('git', args, spawnOptions)
```

And for the commit message argument:

```diff
- await execute(['commit', `-m "${message}"`]);
+ await execute(['commit', '-m', message]);
```

#### Real-World Attack Scenarios

**1. Malicious npm package README / Installation instructions**`

**2. Open-source project contributor exploitation**

**3. CI/CD pipeline poisoning**

**References:**

- Node.js DEP0190: https://nodejs.org/api/deprecations.html#DEP0190
- CWE-78 (OS Command Injection): https://cwe.mitre.org/data/definitions/78.html
- CVE-2024-27980 (related Node.js spawn vulnerability): https://nodejs.org/en/blog/vulnerability/april-2024-security-releases

VenkatKwest:
## Description

This PR fixes command injection vulnerabilities on Windows in two files:

### 1. `executor.ts` (git spawn)
- Removed `shell: true` — git is a native `.exe` and doesn't need it
- Switched from `spawn(\`git ${args.join(' ')}\`)` to `spawn('git', args)`
- Separated `-m` flag from commit message: `['commit', '-m', message]` instead of `['commit', \`-m "${message}"\`]`

### 2. `host.ts` (package manager spawn)
- Package managers (npm/yarn/pnpm) are `.cmd` scripts on Windows and need a shell
- Instead of using `shell: true` with unsanitized argument concatenation, we now invoke `cmd.exe /d /s /c` directly with properly escaped arguments
- Escape logic is based on [cross-spawn](https://github.com/moxystudio/node-cross-spawn)'s battle-tested approach, inlined to avoid adding a dependency
- Uses `windowsVerbatimArguments: true` to prevent Node.js from re-escaping

### Testing
Verified on Windows 11 (Node v25.6.0):
- npm/git commands work correctly with the new approach
- Injection payloads (`& echo INJECTED`, `& echo PWNED > file`) are blocked
- Old `shell: true` pattern confirmed vulnerable (control test)

Fixes #32509

### DIFF
--- a/packages/angular_devkit/schematics/tasks/repo-init/executor.ts
+++ b/packages/angular_devkit/schematics/tasks/repo-init/executor.ts
@@ -29,7 +29,6 @@ export default function (
       const errorStream = ignoreErrorStream ? 'ignore' : process.stderr;
       const spawnOptions: SpawnOptions = {
         stdio: [process.stdin, outputStream, errorStream],
-        shell: true,
         cwd: path.join(rootDirectory, options.workingDirectory || ''),
         env: {
           ...process.env,
@@ -41,7 +40,7 @@ export default function (
       };
 
       return new Promise<void>((resolve, reject) => {
-        spawn(`git ${args.join(' ')}`, spawnOptions).on('close', (code: number) => {
+        spawn('git', args, spawnOptions).on('close', (code: number) => {
           if (code === 0) {
             resolve();
           } else {
@@ -82,7 +81,7 @@ export default function (
       if (options.commit) {
         const message = options.message || 'initial commit';
 
-        await execute(['commit', `-m "${message}"`]);
+        await execute(['commit', '-m', message]);
       }
 
       context.logger.info('Successfully initialized git.');


packages/angular/cli/src/package-managers/host.ts
```diff
@@ -20,6 +20,42 @@ import { platform, tmpdir } from 'node:os';
...
+  arg = `${arg}`;
+
+  // Sequence of backslashes followed by a double quote:
+  // double up all the backslashes and escape the double quote
+  arg = arg.replace(/(?=(\\+?)?)\1"/g, '$1$1\\"');
+
+  // Sequence of backslashes followed by the end of the string
+  // (which will become a double quote later):
+  // double up all the backslashes
+  arg = arg.replace(/(?=(\\+?)?)\1$/, '$1$1');
```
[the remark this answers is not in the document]
VenkatKwest: Thanks for the review! The regex is actually a character-for-character copy from cross-spawn's current `lib/util/escape.js` ([source](https://github.com/moxystudio/node-cross-spawn/blob/master/lib/util/escape.js#L23)).

The `(?=(\\+?)?)\\1` pattern looks unusual, but it's intentional — it was introduced in [cross-spawn PR #160](https://github.com/moxystudio/node-cross-spawn/pull/160) to replace a simpler `(\\*)` pattern that was vulnerable to ReDoS (catastrophic backtracking on crafted input). The current form uses a non-backtracking lookahead to safely match sequences of backslashes.

packages/angular/cli/src/package-managers/host.ts
```diff
@@ -20,6 +20,42 @@ import { platform, tmpdir } from 'node:os';
...
+
+/**
+ * Escapes an argument for safe use in cmd.exe.
+ * Based on the algorithm from cross-spawn (https://github.com/moxystudio/node-cross-spawn)
+ * and https://qntm.org/cmd
+ */
+function escapeArgForCmd(arg: string): string {
+  // Convert to string
+  arg = `${arg}`;
+
```
alan-agius4: This seems redundant since the input is a string. 

```suggestion
```

packages/angular/cli/src/package-managers/host.ts
```diff
@@ -20,6 +20,42 @@ import { platform, tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { PackageManagerError } from './error';
```
alan-agius4, line 22: The below code is literally a copy and paste from cross-spawn.  Please add links or re-write and mention that is was inspires from. Otherwise this cannot be merged.

packages/angular/cli/src/package-managers/host.ts
```diff
@@ -20,6 +20,42 @@ import { platform, tmpdir } from 'node:os';
...
+const metaCharsRegExp = /([()\][%!^"`<>&|;, *?])/g;
+
+/** Escapes a command name for safe use in cmd.exe. */
+function escapeCommandForCmd(cmd: string): string {
+  return cmd.replace(metaCharsRegExp, '^$1');
+}
+
+/**
+ * Escapes an argument for safe use in cmd.exe.
+ * Based on the algorithm from cross-spawn (https://github.com/moxystudio/node-cross-spawn)
```
alan-agius4: This isn't based but copied, please either re-write it or provide proper credits with source links.

packages/angular/cli/src/package-managers/host.ts
```diff
@@ -20,6 +20,42 @@ import { platform, tmpdir } from 'node:os';
...
+ * and https://qntm.org/cmd
+ */
+function escapeArgForCmd(arg: string): string {
```
alan-agius4: NIT, avoid multiple re-assigns.

```ts
function escapeArgForCmd(arg: string): string {
    const processed = arg
    // Escape backslashes preceding a double quote:
    // Every backslash must be doubled, and the quote itself escaped with a backslash.
    .replace(/(?=(\\+?)?)\1"/g, '$1$1\\"')

    // Handle trailing backslashes at the end of the string:
    // Since the string will be wrapped in double quotes, trailing backslashes 
    // must be doubled to prevent them from accidentally escaping the closing quote.
    .replace(/(?=(\\+?)?)\1$/, "$1$1");

  // Wrap the processed string in quotes and escape shell-specific 
  // meta characters using the caret (^) symbol for cmd.exe compatibility.
  return `"${processed}"`.replace(metaCharsRegExp, "^$1");
}
```

alan-agius4 (changes requested):
I had a chat with @clydin and actually this is not a vulnerability at all as the commit message cannot be passed via the command line and thus command injection cannot be triggered in this case.

Hence, kindly revert all the changes `host.ts`, can leave the cleanup in the `executor.ts`

VenkatKwest:
Thanks for the review, @alan-agius4! I've addressed all your feedback:

1. **Removed the redundant `arg = \`${arg}\``** — since the parameter is already typed as `string`.
2. **Added proper attribution with source links** — the docblock now clearly states it's adapted from cross-spawn's `lib/util/escape.js` and links to the authoritative Microsoft documentation for the escaping algorithm it uses.
3. **Refactored to avoid multiple re-assignments** — using a chained `.replace()` approach as you suggested.

VenkatKwest:
https://github.com/user-attachments/assets/080d4fea-04d5-49a8-8f97-2253bfa8bb4e



https://github.com/user-attachments/assets/1ad5ca72-3eab-4e45-84a3-a64149ff34a8


Thanks @alan-agius4 for the review and for checking with @clydin. I appreciate the time you both took.

You're **correct about `executor.ts`** — I verified that `--commit.message` is rejected by the CLI's yargs parser (`Unknown argument: commit.message`), and `--commit` is only registered as a boolean. The commit message in `executor.ts` is not user-controlled via the CLI. I'm happy to keep that as a defense-in-depth cleanup only.

However, I respectfully disagree about `host.ts`. **The `host.ts` vulnerability is real and exploitable.** I have two confirmed PoCs where `calc.exe` opens on Windows:

---

#### PoC 1: Package specifier injection via `ng add`

```bash
ng add "lodash@1.0.0 & calc" --skip-confirmation
```

**Result:** `calc.exe` opens. The injected command executes silently alongside `npm install`.

**Why it works — full stack trace:**

**IMPORTANT NOTE:** STACK TRACE MIGHT BE WRONG BUT ABLE TO REPRODUCE THE VULN

```
1. User runs: ng add "lodash@1.0.0 & calc" --skip-confirmation

2. add/cli.ts:164 → npa("lodash@1.0.0 & calc")
   └─ npa splits at '@' → name="lodash", spec="1.0.0 & calc"
   └─ npa.js:462 → semver.validRange("1.0.0 & calc") returns truthy
   └─ npa classifies as type: "range"
      (BYPASSES the encodeURIComponent check on line 468,
       which only runs for type "tag")
   └─ Returns: { name: "lodash", type: "range", fetchSpec: "1.0.0 & calc" }

3. add/cli.ts → loadPackageInfoTask()
   └─ packageManager.getManifest(packageIdentifier)
   └─ package-manager.ts:494 → case 'range': getRegistryMetadata("lodash")
   └─ Fetches metadata, finds version "1.0.0" satisfies range
   └─ Returns manifest for lodash@1.0.0 ✓

4. add/cli.ts → installPackageTask()
   └─ packageManager.add("lodash@1.0.0 & calc", ...)
   └─ package-manager.ts:326 → args = ["add", "lodash@1.0.0 & calc", "--save-exact"]
   └─ package-manager.ts:327 → this.#run(args, { registry })

5. package-manager.ts:196 → host.runCommand("npm", ["add", "lodash@1.0.0 & calc", ...])

```

The key bypass is in `npm-package-arg` (npa). When a user provides `lodash@1.0.0 & calc`, npa splits at `@` and passes `1.0.0 & calc` to `semver.validRange()`. Semver is permissive — it ignores the trailing garbage `& calc` and returns a valid range. Because npa classifies this as type `"range"` (not `"tag"`), the `encodeURIComponent` validation on [npa.js line 468](https://github.com/npm/npm-package-arg/blob/main/lib/npa.js#L468) is never executed. The unsanitized `& calc` flows all the way to `host.ts`.

---

#### PoC 2: Registry flag injection via `ng add --registry`

```bash
ng add @angular/localize --registry "http://registry.npmjs.org & calc" --skip-confirmation
```

**Result:** `calc.exe` opens 3 times (once per npm command invoked during install).

**Why it works — full stack trace:**

```
1. User runs: ng add @angular/localize --registry "http://registry.npmjs.org & calc"

2. add/cli.ts:118 → yargs parses --registry as a string option (no validation)
   └─ options.registry = "http://registry.npmjs.org & calc"

3. add/cli.ts:555 → installPackageTask() destructures { registry } from options
   └─ add/cli.ts:583-590 → packageManager.add(packageIdentifier, ..., { registry })
   └─ package-manager.ts:327 → this.#run(args, options)  (options includes { registry })
   └─ package-manager.ts:161 → #run(args, { registry: "http://registry.npmjs.org & calc" })

4. package-manager.ts:172 → descriptor.getRegistryOptions(registry)
   └─ Returns: { args: ["--registry", "http://registry.npmjs.org & calc"] }
   └─ package-manager.ts:180 → finalArgs.push("--registry", "http://registry.npmjs.org & calc")

5. package-manager.ts:196 → host.runCommand("npm", [..., "--registry", "http://registry.npmjs.org & calc"])

6. host.ts:182 → shellCommand = "npm view ... --registry http://registry.npmjs.org & calc"
```

---

**The `host.ts` fix is a critical security fix**, not just a cleanup. I agree that `executor.ts` is defense-in-depth only.

**Tested on:** Angular CLI 21.1.4, Node.js v25.6.0, Windows 11, npm 11.3.0

Happy to adjust the PR scope based on this. Let me know how you'd like to proceed.

alan-agius4 (approved):

alan-agius4:

- main: f1ed0257abad0f5772647911d0a2683e3e6b3eb8
- 21.2.x: 7d3844ad633b1f9b80e76014a52848d761b63510

PR #32704, opened 2026-03-05
fix(@angular/cli): conditionally quote package names when adding dependencies based on host requirements

alan-agius4:
Complex range specifiers that include shell special characters (e.g., '>', '<') can be misinterpreted when not quoted. This change ensures that version ranges are enclosed in quotes when needed to prevent such issues.


clydin (approved):

alan-agius4:

- main: 598a690a0baea5cd54af5cea38e673c33605a627
- 21.2.x: 8447d913280a8fa09a842d11193ce77527d0f7a6

PR #32705, opened 2026-03-05
feat(@angular/build): add process.env.PORT support to the dev server

### LINKED ISSUE #32236: Support process.env.PORT for `ng serve`
### Command

serve

### Description

This is supported by nextjs: https://nextjs.org/docs/app/api-reference/cli/next#next-dev-options

The use case is for the firebase emulator where we want to serve both nextjs and angular apps at specific ports determined by our emulator.

If ng serve just supported process.env.PORT then the implementation is simple and everything works as expected.

Currently, without PORT support in ng serve, we have some imperfect hack where we:
- add `--port <>` if we detect a start command that is `ng serve` directly
- but we don't support this if the user is using `npm run dev` as the start command which points to some underlying command that could be anything and could include its own --port specification which doesn't match the port the emulator expects. (https://github.com/firebase/firebase-tools/issues/9699)

(Filing this for tracking purposes only and to link to discussions in issues and PRs. We're also going to better document the caveats above in our own user docs)

_No response_

_No response_

hawkgs:
The change allows setting the dev server port via the `PORT` environment variable.

BREAKING CHANGE: The `@angular/build:dev-server (ng serve)` now assigns the highest priority to the `PORT` environment variable. This value will override any port configurations specified in `angular.json` or via the `--port` command-line flag. This includes the default port 4200.

Since the `port` option is handled by the dev server options JSON schema from what I get, we can only overwrite the already set port (whether that's using the default `4200` or the user-specified `--port`). This means that `process.env.PORT` has the highest priority when it comes to port selection unless the port is already in use. I am not sure if this the desired result, precisely, but the alternative would be to drop `port` from the schema and handle it in the builder execution code instead, if I read the things correctly, which is not an option, I guess.

Resolves #32236

### DIFF
--- a/packages/angular/build/src/builders/dev-server/options.ts
+++ b/packages/angular/build/src/builders/dev-server/options.ts
@@ -86,10 +86,20 @@ export async function normalizeOptions(
     }
   }
 
+  let port = options.port ?? 4200;
+  // Overwrite port, if process.env.PORT is available.
+  if (process.env.PORT) {
+    const envPort = Number(process.env.PORT);
+
+    if (!isNaN(envPort)) {
+      port = envPort;
+      logger.info(`Environment variable "PORT" detected. Using port ${envPort}.`);
+    }
+  }
+
   // Initial options to keep
   const {
     host,
-    port,
     poll,
     open,
     verbose,
@@ -111,7 +121,7 @@ export async function normalizeOptions(
   return {
     buildTarget,
     host: host ?? 'localhost',
-    port: port ?? 4200,
+    port,
     poll,
     open,
     verbose,
--- a/packages/angular/build/src/builders/dev-server/tests/options/port_spec.ts
+++ b/packages/angular/build/src/builders/dev-server/tests/options/port_spec.ts
@@ -42,7 +42,7 @@ describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupT
         port: undefined,
       });
 
-      const { result, response, logs } = await executeOnceAndFetch(harness, '/');
+      const { result, response } = await executeOnceAndFetch(harness, '/');
 
       expect(result?.success).toBeTrue();
       expect(getResultPort(result)).toBe('4200');
@@ -55,7 +55,7 @@ describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupT
         port: 0,
       });
 
-      const { result, response, logs } = await executeOnceAndFetch(harness, '/');
+      const { result, response } = await executeOnceAndFetch(harness, '/');
 
       expect(result?.success).toBeTrue();
       const port = getResultPort(result);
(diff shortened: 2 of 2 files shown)


packages/angular/build/src/builders/dev-server/tests/options/port_spec.ts
```diff
@@ -73,11 +73,30 @@ describeServeBuilder(executeDevServer, DEV_SERVER_BUILDER_INFO, (harness, setupT
...
+      const { result, response } = await executeOnceAndFetch(harness, '/');
+
+      expect(result?.success).toBeTrue();
+      expect(getResultPort(result)).toBe('4201');
+      expect(await response?.text()).toContain('<title>');
+
+      expect(1).toEqual(2);
+
+      delete process.env.PORT;
+    });
```
alan-agius4: Wrap this in a try/finally block so that the clean-up is always performed as otherwise if the test fails it will mess up the other tests.

```suggestion
try {
  process.env.PORT = '4201';
  const { result, response } = await executeOnceAndFetch(harness, '/');

  expect(result?.success).toBeTrue();
  expect(getResultPort(result)).toBe('4201');
  expect(await response?.text()).toContain('<title>');
} finally { 
  delete process.env.PORT;
}
```

alan-agius4 (changes requested):
- Let's not do this in build-angular, as no new features are added there.
- Update the option in https://github.com/angular/angular-cli/blob/4d345b5146b0df9e2ac8d31891f7557b0eb198b4/packages/angular/build/src/builders/dev-server/options.ts#L28
- `Handle issue: ` is not a valid keyword. Please choose Closes, Fixes, Resolves. See: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
- Since this can be a breaking change, please mention it in the commit message using `BREAKING CHANGE:`
- Add the PR description as part of the commit message body.

hawkgs:
Done. The commit already had a message body but I updated it with the breaking change comment.

packages/angular/build/src/builders/dev-server/options.ts
```diff
@@ -86,10 +86,20 @@ export async function normalizeOptions(
...
   }
 
+  let port = options.port ?? 4200;
+  // Overwrite port, if process.env.PORT is available.
+  if (process.env.PORT) {
+    const envPort = Number(process.env.PORT);
+
+    if (!isNaN(envPort)) {
+      port = envPort;
+      logger.info('process.env.PORT detected. Port set to ' + envPort);
```
alan-agius4: NIT: 

```suggestion
      logger.info(`Environment variable "PORT" detected. Using port ${envPort}.`);
```

clydin:

- main: fe720cab64bbc8bcc2db583188e32ad938e63a23

PR #32712, opened 2026-03-06
fix(@angular/build): normalize line endings for CSP hash generation

### LINKED ISSUE #32709: AutoCsp generates invalid hash for inline script in index.html if CRLF
### Command

build

### Is this a regression?

_No response_

### Description

When autoCsp reads from an index.html with CRLF line endings, it generates hashes based on the content with the CRLF line endings. However, the transformed index.html will always be written with LF line endings instead. As a result, the inline scripts become violations because they can no longer match with their hash.

### Minimal Reproduction

The following example uses CRLF line endings for its index.html; StackBlitz doesn't have any indicator for this. You can use VS Code to modify line endings and overwrite the file instead.

https://stackblitz.com/edit/cukg9skb?file=src%2Findex.html

Note that because autoCsp only applies to builds, the static built files are served instead of via ng serve, meaning you should rebuild to apply any changes.

### Exception or Error

```text
Executing inline script violates the following Content Security Policy directive 'script-src 'strict-dynamic' 'sha256-DseH/kfLLMMDrt5Fv7yG84W/vhVyzaUlY2uyB+yl+rU=' 'sha256-ku3oFQ8DDR81xC6Gz32A3bU2PKxH12YBBTzRvVuNLKM=' https: 'unsafe-inline''. Note that 'unsafe-inline' is ignored if either a hash or nonce value is present in the source list. The action has been blocked.
```

### Your Environment

```text
Angular CLI       : 21.2.0
Angular           : 21.2.0
Node.js           : 20.19.1
Package Manager   : npm 10.8.2
Operating System  : linux x64
```

### Anything else relevant?

_No response_

alan-agius4:
When `autoCsp` reads an `index.html` with CRLF line endings, it generates hashes based on the CRLF content. However, the transformed file is always written with LF line endings, causing CSP violations.

This commit ensures that script content line endings are normalized to LF before hashing to match the output file.

Closes #32709

### DIFF
--- a/packages/angular/build/src/utils/index-file/auto-csp.ts
+++ b/packages/angular/build/src/utils/index-file/auto-csp.ts
@@ -52,12 +52,7 @@ function isJavascriptMimeType(mimeType: string): boolean {
  * @returns whether to add the script tag to the dynamically loaded script tag
  */
 function shouldDynamicallyLoadScriptTagBasedOnType(scriptType: string | undefined): boolean {
-  return (
-    scriptType === undefined ||
-    scriptType === '' ||
-    scriptType === 'module' ||
-    isJavascriptMimeType(scriptType)
-  );
+  return !scriptType || scriptType === 'module' || isJavascriptMimeType(scriptType);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


clydin:

- main: 6324133c282f5e04ee6e9e46fc5f387cbbefad8e
- 21.2.x: 8dec0c62ba40af339f4fd0fa34f20cbed545cd71

PR #32717, opened 2026-03-06
perf(@angular/cli): cache root manifest and resolve restricted package exports in ng add

clydin:
This change enhances the `ng add` command's performance by caching the root project manifest (`package.json`) to avoid redundant disk reads and JSON parsing during peer dependency conflict checks.

Additionally, it improves the robustness of `package.json` resolution for installed packages. Previously, resolving `package.json` could fail if a third-party package used the Node.js `"exports"` field without explicitly exporting its `package.json`. The CLI now correctly handles this by falling back to resolving the package's entry point and traversing upwards to find the manifest.


alan-agius4 (approved):

clydin:

- main: a39a33128fb56e9c65ea89e06c4f127252d3b220

PR #32722, opened 2026-03-09
fix(@angular/cli): preserve exact version in ng add when requested

clydin:
The `ng add` command would previously always add a caret (`^`) prefix to the resolved package version, even if the user explicitly requested an exact version. This change ensures that the exact version is preserved in the package identifier if requested, while maintaining the caret prefix for general requests to ensure consistent behavior with package managers.


alan-agius4 (approved):

clydin:

- main: b5fb457e157fa78b61565eaba6b88e9f80b3b288
- 21.2.x: d2f209823a524a6effde4910017547675c7a6166

PR #32737, opened 2026-03-10
Eager Change Detection

alan-agius4:


packages/schematics/angular/component/schema.json
```diff
@@ -65,9 +65,9 @@
     },
     "changeDetection": {
       "description": "Configures the change detection strategy for the component.",
-      "enum": ["Default", "OnPush"],
```
clydin, line 68: Should we leave `Default` here until its removed on the framework side?

clydin (approved):

packages/schematics/angular/component/schema.json
alan-agius4: I removed it, cause the FW already did a migration to move all apps from using Default to Eager.

alan-agius4:

- main: 93c3eb8fb2a0d531f18779152b0a62e9b73dbb23

PR #32752, opened 2026-03-11
fix(@schematics/angular): do not add changeDetection in default template

cexbrayat:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

As the new default is `Eager`, the CLI was adding `changeDetection: ChangeDetectionStrategy.Eager` in every component

- [ ] Yes
- [x] No


alan-agius4 (changes requested):
This is intended as `ChangeDetectionStrategy.Eager` needs to be explicitly set. We should not add `ChangeDetectionStrategy.OnPush` which is now the default instead.

Maybe we should also update the schematics default to `OnPush` to align with the FW default.

cexbrayat:
@alan-agius4 Thanks for your feedback, I'll close this one and opened https://github.com/angular/angular-cli/pull/32753 instead which switches the default to OnPush and don't add it to the decorator when that's the case.

PR #32761, opened 2026-03-12
fix(@angular/build): only use external packages for polyfills when no local files are present

clydin:
The polyfills bundle now conditionally employs external package resolution only if no local files are detected in the polyfills array. If local files are present, all polyfills are bundled together to ensure the import execution order is correctly preserved between local and package-based entries.

### DIFF
--- a/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
+++ b/packages/angular/build/src/tools/esbuild/application-code-bundle.ts
@@ -654,17 +654,18 @@ function getEsBuildCommonPolyfillsOptions(
   tryToResolvePolyfillsAsRelative: boolean,
   loadResultCache: LoadResultCache | undefined,
 ): BuildOptions | undefined {
-  const { jit, workspaceRoot, i18nOptions } = options;
+  const { jit, workspaceRoot, i18nOptions, externalPackages } = options;
+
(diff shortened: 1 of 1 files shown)


alan-agius4 (approved):

packages/angular/build/src/tools/esbuild/application-code-bundle.ts
```diff
@@ -749,3 +750,18 @@ function getEsBuildCommonPolyfillsOptions(
...
+ * @param path The polyfill path to check.
+ * @returns true if the path is a local file; false otherwise.
+ */
+function isLocalFile(path: string): boolean {
+  if (path === 'zone.js' || path.startsWith('zone.js/')) {
+    return false;
+  }
+
+  return path.startsWith('.') || /\.[mc]?[jt]sx?$/.test(path);
+}
```
alan-agius4, line 767: NIT:

```suggestion
function isLocalFile(path: string): boolean {
  if (path.startsWith('zone.js')) {
    return false;
  }

  return path[0] === '.' || /\.[mc]?[jt]sx?$/.test(path);
}
```

clydin:

- main: 4643a8a3b3e2a3bcf7baae9f812ae8a9ef10ebad
- 21.2.x: d009aa1ec7411b67b61b81003eb6181cde6f306f

PR #32765, opened 2026-03-13
fix(@angular/build): alias createRequire banner import to avoid duplicate binding

arturovt:
ESBuild processes banner content as raw text outside of its module graph, so it cannot deduplicate or rename banner imports the way it does for user imports. If user code already imports `createRequire` from `node:module`, the injected banner produces a duplicate binding that causes a runtime error.

Alias the banner import to `__ngCreateRequire` to avoid colliding with any `createRequire` binding that esbuild may emit from bundled user code.


alan-agius4 (approved):
LGTM

dgp1130:

- main: 9b33e1781328d3b42665145bf580fb8e06c8ad2b
- 21.2.x: 75fa94cad26b0947e687ef94d50653cb7651d18c

PR #32771, opened 2026-03-16
fix(@angular/ssr): disallow x-forwarded-prefix starting with a backslash

alan-agius4:
Updated the INVALID_PREFIX_REGEX to ensure that prefixes starting with a backslash are considered invalid. Previously, only multiple slashes or dot segments were explicitly disallowed at the start.

Also updated the associated validation error message and unit tests to reflect this change.

### DIFF
--- a/packages/angular/ssr/src/utils/redirect.ts
+++ b/packages/angular/ssr/src/utils/redirect.ts
@@ -27,9 +27,14 @@ export function isValidRedirectResponseCode(code: number): boolean {
  * @param location - The URL to which the response should redirect.
  * @param status - The HTTP status code for the redirection. Defaults to 302 (Found).
  *                 See: https://developer.mozilla.org/en-US/docs/Web/API/Response/redirect_static#status
+ * @param headers - Additional headers to include in the response.
  * @returns A `Response` object representing the HTTP redirect.
  */
-export function createRedirectResponse(location: string, status = 302): Response {
+export function createRedirectResponse(
+  location: string,
+  status = 302,
+  headers?: Record<string, string>,
+): Response {
   if (ngDevMode && !isValidRedirectResponseCode(status)) {
     throw new Error(
       `Invalid redirect status code: ${status}. ` +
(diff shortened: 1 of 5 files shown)


packages/angular/ssr/src/utils/validation.ts
```diff
@@ -29,7 +29,7 @@ const VALID_HOST_REGEX = /^[a-z0-9.:-]+$/i;
 /**
  * Regular expression to validate that the prefix is valid.
  */
-const INVALID_PREFIX_REGEX = /^[/\\]{2}|(?:^|[/\\])\.\.?(?:[/\\]|$)/;
+const INVALID_PREFIX_REGEX = /^\\|^\/[/\\]|(?:^|[/\\])\.\.?(?:[/\\]|$)/;
```
dgp1130: Nit: Parens might make the order of operations clearer, `/(^\\|^\/)[/\\].../`

packages/angular/ssr/src/utils/redirect.ts
```diff
@@ -40,7 +45,9 @@ export function createRedirectResponse(location: string, status = 302): Response
   return new Response(null, {
     status,
     headers: {
+      ...headers,
       'Location': location,
+      'Vary': 'X-Forwarded-Prefix',
```
dgp1130: Consider: What if someone tries to set `Vary: foo`, seems like we'd override with our `Vary: X-Forwarded-Prefix`. Should we merge that to `Vary: X-Forwarded-Prefix, foo`?

packages/angular/ssr/src/utils/redirect.ts
```diff
@@ -40,7 +45,9 @@ export function createRedirectResponse(location: string, status = 302): Response
   return new Response(null, {
     status,
     headers: {
+      ...headers,
       'Location': location,
```
dgp1130: Nit: Should we warn or error (probably limited to `ngDevMode`) if the user specifies a `Location` header given that we'll ignore it?

dgp1130 (approved):

alan-agius4:

- main: 998b8298e3106c95d77ced8090ab815365c462c7
- 21.2.x: 84385411d4542d60d635aea9063c1fd751deb607

VenkatKwest:
Hi @alan-agius4 

Thanks for providing this patch. I reported this specific bypass of the CVE[https://github.com/angular/angular-cli/security/advisories/GHSA-xh43-g2fq-wjrj] in a Google OSS VRP report last month[https://issuetracker.google.com/issues/487861221], so I’m glad to see the validation for backslashes is now officially merged.

PR #32782, opened 2026-03-17
fix(@angular/cli): use parsed package name for migrate-only updates

alan-agius4:
This change updates the update command to use the parsed package name from `npm-package-arg` instead of the raw input string when performing a 'migrate-only' update. This ensures that package identifiers are correctly and consistently handled.


clydin (approved):

alan-agius4:

- main: c9f07db8fcb0b3f8400fbbf0e131be7f9857a987
- 21.2.x: 1505164bb2703254a2b25a76c7b3a1ff2fd76a85

